PR #1868, opened 2021-05-06
Recognize AWS Lambda, Heroku Dyno, and Azure App Service environments

DentonGentry:

PR #2123, opened 2021-06-14
hostinfo: capitalization of AWS

DentonGentry:
Missed one comment from https://github.com/tailscale/tailscale/pull/1868
should be isAWSLambda not isAwsLambda


DentonGentry:
Planning to just submit when tests pass.

Issue #2079, opened 2021-06-08
tstest/integration: TestAddPingRequest flakes


bradfitz:
The new `TestAddPingRequest` flakes sometimes, e.g. with:

```
$ go test -v -race  -count=5 ./tstest/integration/
...
--- PASS: TestNodeAddressIPFields (9.85s)
=== CONT  TestAddPingRequest
    integration_test.go:266: didn't get PingRequest from tailscaled
    stuntest.go:64: STUN server shutdown
--- FAIL: TestAddPingRequest (21.67s)
```

@simenghe, can you take a look?

bradfitz:
@simenghe, see the https://github.com/tailscale/tailscale/commit/0affcd4e12c0a2413712548a954166221d155298 that I just pushed, including its commit message with details of how I was debugging.

That might help you see what's happening.

I can spend some more time on it tomorrow probably.

PR #2124, opened 2021-06-14
tstest/integration: disable TestAddPingRequest

DentonGentry:
Failing often now, we don't want people to get used to
routinely ignoring test failures.

Can be re-enabled when
https://github.com/tailscale/tailscale/issues/2079
is resolved.

### DIFF
--- a/tstest/integration/integration_test.go
+++ b/tstest/integration/integration_test.go
@@ -230,6 +230,7 @@ func TestNodeAddressIPFields(t *testing.T) {
 }
 
 func TestAddPingRequest(t *testing.T) {
+	t.Skip("flaky in CI, tailscale/issues/2079")
 	t.Parallel()
 	bins := BuildTestBinaries(t)
 


tstest/integration/integration_test.go
```diff
@@ -230,6 +230,7 @@ func TestNodeAddressIPFields(t *testing.T) {
 }
 
 func TestAddPingRequest(t *testing.T) {
+	t.Skip() // TODO: https://github.com/tailscale/tailscale/issues/2079
```
bradfitz, line 233: LGTM. Usually you put the reason in the Skip message, though:

Screen Shot 2021-06-14 at 7 42 08 PM

bradfitz (approved):

bradfitz:
Also, commit message prefix should be `tstest/integration: `

tstest/integration/integration_test.go
DentonGentry: Done.

PR #2125, opened 2021-06-14
derp: log highest bandwidth ip

JulianKnodt:
Adds metrics for reporting which ips are using the most bandwidth, which would hopefully give
insight into the locations using the most derp bandwidth. I wasn't sure what N to make it, so I
left it as 100, but probably makes sense to scale up further depending on how many users there
are, possibly dynamically in the future. It should be cheap to increase the number of items tracked, as it uses a heap with a limited size.


derp/derp_server.go
```diff
@@ -1398,3 +1419,26 @@ func writePublicKey(bw *bufio.Writer, key *key.Public) error {
 	}
 	return nil
 }
+
+type ipBandwidth struct {
+	ip        string
```
bradfitz, line 1424: netaddr.IP ?

derp/derp_server.go
```diff
@@ -1398,3 +1419,26 @@ func writePublicKey(bw *bufio.Writer, key *key.Public) error {
...
+	bandwidth int64
```
bradfitz, line 1425: bandwidth implies a rate. is this just `bytesTransferred` (or `bytes`)? is this sent or received? or both?

derp/derp_server.go
```diff
@@ -510,6 +515,7 @@ func (c *sclient) run(ctx context.Context) error {
 			}
 			return fmt.Errorf("client %x: readFrameHeader: %w", c.key, err)
 		}
+		go c.s.clientBandwidth.Map.Add(c.remoteAddr, int64(fl))
```
bradfitz, line 518: we can't have this map growing unbounded forever. also: remoteAddr changes not just on IP but also on port.

derp/derp_server.go
```diff
@@ -1323,6 +1329,21 @@ func (s *Server) ExpVar() expvar.Var {
 	m.Set("average_queue_duration_ms", expvar.Func(func() interface{} {
 		return math.Float64frombits(atomic.LoadUint64(s.avgQueueDuration))
 	}))
+	m.Set("top_100_bandwidth_ips", expvar.Func(func() interface{} {
```
bradfitz, line 1332: I think this is going to be too expensive to do on every metrics scrape.

I think we need a per-client counter and also a global low watermark of what score would put us in the top 100, and then do an O(1) addition to/from the leaderboard set as needed, without O(n) work like this

derp/derp_server.go
JulianKnodt: will update to bytes sent

derp/derp_server.go
JulianKnodt: I can make this into an actual IP, but it seems a bit odd that it's not an IP on the `sclient`

derp/derp_server.go
JulianKnodt: I added a timer so that whenever anything new is added to the map, it removes items that weren't added in a certain time.

derp/derp_server.go
JulianKnodt: changed to bytesSent

derp/derp_server.go
```diff
@@ -1398,3 +1418,107 @@ func writePublicKey(bw *bufio.Writer, key *key.Public) error {
...
+
+	index int
+}
+
+func (i ipBytesSent) value(maxTime time.Duration) float64 {
+	decay := math.Max(0, float64(maxTime-time.Since(i.lastSeen))/float64(maxTime))
+	return float64(i.bytesSent) * decay
+}
+
+type ipBytesSentPriorityQueue struct {
```
bradfitz, line 1435: I still think this is going to be too slow to be in the data path.

You don't want to do a mutex or a map lookup.

I'd just increment a local value on `sclient`, and only add/remove it from a leaderboard set if the value passes a certain value

derp/derp_server.go
```diff
@@ -510,6 +517,14 @@ func (c *sclient) run(ctx context.Context) error {
 			}
 			return fmt.Errorf("client %x: readFrameHeader: %w", c.key, err)
 		}
+		go func() {
+			addr, err := netaddr.ParseIPPort(c.remoteAddr)
```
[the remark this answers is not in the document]
JulianKnodt: yep, should I change the sclient to keep a IPPort instead?

derp/derp_server.go
JulianKnodt: hm makes sense, that's why I put it in a goroutine so that it wouldn't block the main data path but it's pretty costly still

JulianKnodt:
Benchstat:
```
name                      old time/op    new time/op    delta
SendRecv/msgsize=10-8       8.95µs ± 5%    8.22µs ±20%   ~     (p=0.421 n=5+5)
SendRecv/msgsize=100-8      9.10µs ± 1%    9.37µs ±21%   ~     (p=1.000 n=4+5)
SendRecv/msgsize=1000-8     10.4µs ± 6%    10.1µs ±19%   ~     (p=0.310 n=5+5)
SendRecv/msgsize=10000-8    28.5µs ± 3%    27.0µs ±14%   ~     (p=0.151 n=5+5)

name                      old speed      new speed      delta
SendRecv/msgsize=10-8     1.12MB/s ± 4%  1.23MB/s ±23%   ~     (p=0.341 n=5+5)
SendRecv/msgsize=100-8    10.6MB/s ±13%  10.9MB/s ±24%   ~     (p=0.841 n=5+5)
SendRecv/msgsize=1000-8   96.4MB/s ± 6%  99.5MB/s ±17%   ~     (p=0.310 n=5+5)
SendRecv/msgsize=10000-8   351MB/s ± 3%   373MB/s ±15%   ~     (p=0.151 n=5+5)

name                      old alloc/op   new alloc/op   delta
SendRecv/msgsize=10-8         112B ± 0%      112B ± 0%   ~     (all equal)
SendRecv/msgsize=100-8        208B ± 0%      207B ± 1%   ~     (p=0.333 n=4+5)
SendRecv/msgsize=1000-8     1.12kB ± 0%    1.12kB ± 0%   ~     (p=0.333 n=4+5)
SendRecv/msgsize=10000-8    19.9kB ± 1%    19.8kB ± 1%   ~     (p=0.690 n=5+5)

name                      old allocs/op  new allocs/op  delta
SendRecv/msgsize=10-8         3.00 ± 0%      2.60 ±23%   ~     (p=0.333 n=4+5)
SendRecv/msgsize=100-8        3.00 ± 0%      2.60 ±23%   ~     (p=0.333 n=4+5)
SendRecv/msgsize=1000-8       3.00 ± 0%      2.60 ±23%   ~     (p=0.444 n=5+5)
SendRecv/msgsize=10000-8      3.00 ± 0%      3.00 ± 0%   ~     (all equal)
```

JulianKnodt:
Closed in favor of a different implementation which maps IP --> public Key for logging with unix tools

JulianKnodt:
Closed in favor of a different implementation which maps IP --> public Key for logging with unix tools
#2167

Issue #1235, opened 2021-01-31
Add support for automatically getting TLS certificates (using LetsEncrypt dns-01 challenges, etc)


normanr:
Many new web APIs (eg: geolocation, sensors, http/2, etc) require a TLS certificate/[secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts/features_restricted_to_secure_contexts) in order to function. It is not currently possible to get a TLS certificate for the full Magic DNS name of a tailscale node from services like Let's Encypt because they're not exposed to the public internet in any way.

There is one possible option though: [dns-01 challenges](https://letsencrypt.org/docs/challenge-types/#dns-01-challenge) require that the certificate requester can control the dns TXT records for the hostname they are requesting a certificate for. So for this to work it would require a new DNS server exposed to the public internet (with beta.tailscale.net delegated to it), and a way for the TXT records for _acme-challenge.[hostname].[user-domain].beta.tailscale.net to be updated by the ACME client running on the tailscale node.

Many ACME clients have a plugin based system for actually doing the DNS record update. The most basic just runs a script (which could just invoke the tailscale cli appropriately). It would be nice to support one of the "generic" update methods, either acme-dns (which is just RESTful HTTP API), or [dynamic dns updates](https://tools.ietf.org/html/rfc2136) (more complex). They could use 100.100.100.100 as their update endpoint.

Internal to tailscale, it would mean being able to store an extra string per-node (multiple strings if #1196 is to be supported), that the public DNS server can put into the answer for the TXT query. The data doesn't _have_ to be stored centrally (it could be kept locally in tailscaled, and fetched at run time from the central server), but I suspect that it may be easier to push it as close as possible to the DNS server as quickly as possible, as the data is typically only read once, and then it is removed. It's also not secret data, nor does it need to be persisted to disk.

The default renewal period for these certificates is 90 days, so this would be a very low QPS service (even lower than derp traffic I would guess).

The alternative challenges all require that the certificate issuer can establish TCP connections to the hostname requested in the certificate. This could be done by sharing the host with the certificate issuer (they'd have to run a tailscale deployment in order to support this), or allowing certificate issuers to connect to the host via the public internet (which I assume is also highly unlikely to ever happen).

Other alternatives include non-Automatic Certificate Management (often requires payment), or solutions involving services like ngrok / localhost.run.

bradfitz:
[...]

PR #2135, opened 2021-06-15
tailcfg: add DNSConfig.SetDNSDomains

bradfitz:
Updates #1235

Signed-off-by: Brad Fitzpatrick <bradfitz@tailscale.com>

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -43,7 +43,8 @@ import (
 //    18: 2021-04-19: MapResponse.Node nil means unchanged (all fields now omitempty)
 //    19: 2021-04-21: MapResponse.Debug.SleepSeconds
 //    20: 2021-06-11: MapResponse.LastSeen used even less (https://github.com/tailscale/tailscale/issues/2107)
-const CurrentMapRequestVersion = 20
+//    21: 2021-06-15: added MapResponse.DNSConfig.CertDomains
+const CurrentMapRequestVersion = 21
 
 type StableID string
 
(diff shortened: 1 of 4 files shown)


bradfitz:
Thoughts?

bradfitz:
Don't want it top-level in MapResponse as I don't want to add new minnification code to control for it. DNSConfig is already dup-suppressed.

tailcfg/tailcfg.go
```diff
@@ -873,6 +874,10 @@ type DNSConfig struct {
 
 	// PerDomain is not set by the control server, and does nothing.
 	PerDomain bool `json:",omitempty"`
+
+	// SetDNSDomains are the set of DNS names that are permitted to
```
danderson, line 878: I'm not sure about the field name. SetDNSDomains sounds like an imperative command for the client to set some DNS domains, not "these are the domains you can choose to request stuff on". We also don't guarantee that you can set whatever records you want on these names, it's just a hint that "if you're trying to do LE, these are the ones that are likely to succeed".

I don't have better ideas off the top of my head, the DNS record-setting stuff in tailcfg is so generic that it's hard to come up with reasonable names :/

danderson (approved):

tailcfg/tailcfg.go
bradfitz: ValidDNSCertDomain ?

bradfitz: ValidCertDomains?

normanr: What's the motivation for this field? Does it get set by the tailscale administrator to limit which machines may set TXT records, or does it limit which "sub-names" are allowed (eg: `_acme-challenge`), or is it both? Should it allow wildcards, eg: `_acme-challenge.*` (to allow any machine to use LE), or `*.some.awesome.machine` (which can set a TXT record for _any_ sub-name)

bradfitz: @normanr, we haven't decided what subdomain structure we're going to use for LetsEncrypt certs. There are various concerns about customer hostnames/emails leaking to CT logs if we do it with our current MagicDNS names. So we might make this opt-in with separate hostnames. This is giving us a place to advertise that policy from the server to the client later, so we can update it server side and not require client updates.

bradfitz: Decision from @danderson: just `CertDomains`.

PR #2139, opened 2021-06-15
net/dns: work around mysterious LXC resolv.conf.

### LINKED ISSUE #2136: Split DNS not working on a systemd-resolvconf Ubuntu system
Using tailscale version 1.9.164 on Ubuntu 20.04, with https://login.tailscale.com/admin/dns of:
Magic DNS enabled
Split DNS: ts-dns.test 100.120.55.47
Global DNS: 100.92.61.18 with Override local enabled

resolvctl shows 100.100.100.100 as configured:

```
$ resolvectl
Global
       LLMNR setting: no
MulticastDNS setting: no
  DNSOverTLS setting: no
      DNSSEC setting: no
    DNSSEC supported: no
  Current DNS Server: 10.1.10.1
         DNS Servers: 10.1.10.1
          DNSSEC NTA: 10.in-addr.arpa
                      16.172.in-addr.arpa
                      168.192.in-addr.arpa
                      17.172.in-addr.arpa
                      18.172.in-addr.arpa
                      19.172.in-addr.arpa
                      20.172.in-addr.arpa
                      21.172.in-addr.arpa
                      22.172.in-addr.arpa
                      23.172.in-addr.arpa
                      24.172.in-addr.arpa
                      25.172.in-addr.arpa
                      26.172.in-addr.arpa
                      27.172.in-addr.arpa
                      28.172.in-addr.arpa
                      29.172.in-addr.arpa
                      30.172.in-addr.arpa
                      31.172.in-addr.arpa
                      corp
                      d.f.ip6.arpa
                      home
                      internal
                      intranet
                      lan
                      local
                      private
                      test

Link 6 (tailscale0)
      Current Scopes: DNS
DefaultRoute setting: yes
       LLMNR setting: no
MulticastDNS setting: no
  DNSOverTLS setting: no
      DNSSEC setting: no
    DNSSEC supported: no
         DNS Servers: 100.100.100.100
          DNS Domain: geekhold.com.beta.tailscale.net
                      ts-dns.test
                      ~.

Link 4 (docker0)
      Current Scopes: none
DefaultRoute setting: no
       LLMNR setting: yes
MulticastDNS setting: no
  DNSOverTLS setting: no
      DNSSEC setting: no
    DNSSEC supported: no

Link 2 (eth0)
      Current Scopes: DNS
DefaultRoute setting: yes
       LLMNR setting: yes
MulticastDNS setting: no
  DNSOverTLS setting: no
      DNSSEC setting: no
    DNSSEC supported: no
         DNS Servers: 10.1.10.1
          DNS Domain: localdomain
```

Note the tilde-dot output in DNS Domains for tailscale0 really is there in the output. I don't know why.

constantine.ts-dns.test should be handled by the split DNS server at 100.120.55.47, but fails:
```
$ dig constantine.ts-dns.test

; <<>> DiG 9.16.1-Ubuntu <<>> constantine.ts-dns.test
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 7904
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;constantine.ts-dns.test.       IN      A

;; AUTHORITY SECTION:
test.                   10800   IN      SOA     localhost. nobody.invalid. 1 3600 1200 604800 10800

;; Query time: 0 msec
;; SERVER: 10.1.10.1#53(10.1.10.1)
;; WHEN: Tue Jun 15 12:59:11 PDT 2021
;; MSG SIZE  rcvd: 111
```

I think this happens because /etc/resolv.conf doesn't point to either Magic DNS nor a 127.0.0.1:53 resolver to handle split DNS, it just points to the eth0 nameserver:

```
$ cat /etc/resolv.conf
# systemd-resolved
nameserver 10.1.10.1
```

danderson:
It appears that Proxmox/LXC can install a resolv.conf that mentions
systemd-resolved by name in its header comment, but actually has nothing
at all to do with systemd-resolved. On such setups, we incorrectly choose
to configure resolved even though it has no control over the "true" DNS
configuration.

Work around this by only using resolved if resolv.conf mentions resolved
in its header comment, _and_ has a single resolver pointing to 127.0.0.53.

Strictly speaking, this can break a valid configuration of resolved, where
resolv.conf is symlinked to a resolved-managed file that just lists upstream
DNS servers directly. However, as far as I know, this configuration never
occurs in the wild, and that "upstreams directly" resolv.conf is meant
for consumption by the likes of dnsmasq and other things that don't examine
/etc/resolv.conf directly. This change still works for the "normal" way
in which resolved is deployed as the system's DNS resolver.

Fixes #2136.

### DIFF
--- a/net/dns/manager_linux.go
+++ b/net/dns/manager_linux.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/godbus/dbus/v5"
+	"inet.af/netaddr"
 	"tailscale.com/types/logger"
 	"tailscale.com/util/cmpver"
 )
@@ -50,6 +51,15 @@ func NewOSConfigurator(logf logger.Logf, interfaceName string) (ret OSConfigurat
 	switch resolvOwner(bs) {
 	case "systemd-resolved":
 		dbg("rc", "resolved")
+		// Some systems, for reasons known only to them, have a
+		// resolv.conf that has the word "systemd-resolved" in its
+		// header, but doesn't actually point to resolved. We mustn't
+		// try to program resolved in that case.
+		// https://github.com/tailscale/tailscale/issues/2136
+		if err := resolvedIsActuallyResolver(); err != nil {
+			dbg("resolved", "not-in-use")
+			return newDirectManager()
+		}
 		if err := dbusPing("org.freedesktop.resolve1", "/org/freedesktop/resolve1"); err != nil {
 			dbg("resolved", "no")
 			return newDirectManager()
@@ -180,6 +190,17 @@ func nmIsUsingResolved() error {
 	}
 	if mode != "systemd-resolved" {
 		return errors.New("NetworkManager is not using systemd-resolved for DNS")
+	}
+	return nil
+}
+
+func resolvedIsActuallyResolver() error {
+	cfg, err := readResolvConf()
+	if err != nil {
+		return err
+	}
+	if len(cfg.Nameservers) != 1 || cfg.Nameservers[0] != netaddr.IPv4(127, 0, 0, 53) {
+		return errors.New("resolv.conf doesn't point to systemd-resolved")
 	}
 	return nil
 }


DentonGentry (approved):

net/dns/manager_linux.go
```diff
@@ -50,6 +51,15 @@ func NewOSConfigurator(logf logger.Logf, interfaceName string) (ret OSConfigurat
 	switch resolvOwner(bs) {
 	case "systemd-resolved":
 		dbg("rc", "resolved")
+		// Some systems, for reasons known only to them, have a
```
bradfitz, line 54: I think it's because LXC takes its host resolv.conf and bind mounts it into the container. The host probably has systemd-resolved.

Maybe detect running-in-LXC instead? We have that code elsewhere in the tree.

danderson: The format of the resolv.conf doesn't match anything written by systemd-resolved though. It's as if LXC had made up a config, and then for the lulz just added `# systemd-resolved` at the top of the file.

I'd rather not do something special for LXC, because when appropriately configured our existing logic will do the right thing. This is just the first time I've seen a resolv.conf that claims to belong to systemd-resolved, but was actually plopped down by something completely other with no explanation.

DentonGentry: I think I probably did this: my notes for a different LXC include installing Tailscale and experimenting with getting DNS settings to work. I had a problem with Proxmox overwriting /etc/resolv.conf.

If I edited /etc/resolv.conf in the wrong window and changed the Ubuntu host's resolver config, then that would explain why the LXC I was trying to work with didn't do what I expected.

danderson: Hmm. So this may be too isolated an issue to warrant changing our code... That said, the additional check is harmless IMO, in that it won't start doing the wrong thing on configs that used to work, and may do better in a few more edge cases. On that basis, I think I'm going to merge.

PR #2141, opened 2021-06-16
fix a wrong check

lewgun:


bradfitz:
Thanks! But this could use a test update too.

bradfitz:
Timeout on DCO & test, so fixed it in f35b8c3ead116771268d49cbc8b47c481c59d7db with thanks to @lewgun!

Issue #1527, opened 2021-03-17
Add an option to allow LAN access when using an exit node.


Flameeyes:
**Describe the bug**

When selecting an exit node, the new routing table (52 in my case) does not include any of the local subnets, making it impossible to reach them, and reflecting the traffic due to them back onto `tailscale0`.

This pretty much means a node that selected an exit node becomes invisible to all the local clients. This can be solved by adding the local subnets back to the newly created routing table.

Now, according to the `ip rule` output, it loosk like this is meant to be addressed by `fwmark 0x80000`:

```
rule
0:      from all lookup local
5210:   from all fwmark 0x80000 lookup main
5230:   from all fwmark 0x80000 lookup default
5250:   from all fwmark 0x80000 unreachable
5270:   from all lookup 52
32766:  from all lookup main
32767:  from all lookup default
```

That forces a lookup on otable main, that includes the local routes.
But as far as I can tell nothing is setting that mark:

```
flameeyes@brentcube:~$ sudo iptables -L -v -n
Chain INPUT (policy ACCEPT 2330 packets, 241K bytes)
 pkts bytes target     prot opt in     out     source               destination
 2333  241K ts-input   all  --  *      *       0.0.0.0/0            0.0.0.0/0

Chain FORWARD (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination
 1939  131K ts-forward  all  --  *      *       0.0.0.0/0            0.0.0.0/0

Chain OUTPUT (policy ACCEPT 3807 packets, 523K bytes)
 pkts bytes target     prot opt in     out     source               destination

Chain ts-forward (1 references)
 pkts bytes target     prot opt in     out     source               destination
    0     0 MARK       all  --  tailscale0 *       0.0.0.0/0            0.0.0.0/0            MARK set 0x40000
    0     0 ACCEPT     all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match 0x40000
    0     0 DROP       all  --  *      tailscale0  100.64.0.0/10        0.0.0.0/0
 1939  131K ACCEPT     all  --  *      tailscale0  0.0.0.0/0            0.0.0.0/0

Chain ts-input (1 references)
 pkts bytes target     prot opt in     out     source               destination
    0     0 ACCEPT     all  --  lo     *       100.121.107.31       0.0.0.0/0
    0     0 RETURN     all  --  !tailscale0 *       100.115.92.0/23      0.0.0.0/0
[...]

PR #2146, opened 2021-06-16
wgengine/router_windows: support toggling local lan access when using exit nodes.

maisem:
Note: This requires reconfiguring the firewall while it's enabled to allow the local routes through, the way we do this is by passing in a json encoded list of routes to the firewall process via stdin.

Updates #1527


crawshaw (approved):
LGTM. I'm going to try running it here, and if it works for a while I'll merge it and do an unstable release.

PR #2151, opened 2021-06-16
derp: rate limit how often same-key clients can kick each other off server

bradfitz:


derp/derp_server.go
```diff
@@ -472,7 +495,15 @@ func (s *Server) accept(nc Conn, brw *bufio.ReadWriter, remoteAddr string, connN
...
 	}
 
-	s.registerClient(c)
+	for {
+		ok, d := s.registerClient(c)
+		if ok {
+			break
+		}
+		s.clientsReplaceSleeping.Add(1)
+		time.Sleep(d)
```
sailorfrag, line 504: should it sleep d + (something small) or is the rate limiter immune to scheduler quantization type issues where it wakes up only to discover it needs to sleep a bit longer?

bradfitz: I don't think anything else is needed. Never seen other x/time/rate users do that. And if there's a thundering herd, it's of a relatively small number of clients per herd. (and usually it's just one, as two nodes fighting is most common)

crawshaw (approved):
I think that will work! Going to stick with manual testing?

bradfitz:
I added tests which found bugs and existing problems, yay. Converting to a draft for a bit.

PR #2113, opened 2021-06-11
net/portmapper: Add upnp

JulianKnodt:
Before adding the portmapper for UPnP, I found that it takes a significant amt of time (2s), to confirm that UPnP is present. The current probing method does not allow for any backoff, or easy reuse of clients. This attempts to refactor it into a separate `Prober`, which can track state. I'll also probably add some more to use it elsewhere, but wanted to put it up to see if the idea generally makes sense.

Output from prober:
```
=== RUN   TestClientProber
    probe_test.go:25: Got: {PCP:false PMP:false UPnP:true}, <nil>
--- PASS: TestClientProber (3.00s)
```

Output from `probe`
```
=== RUN   TestClientProbe
    portmapper_test.go:40: Got: {PCP:false PMP:false UPnP:false}, <nil>
    portmapper_test.go:40: Got: {PCP:false PMP:false UPnP:false}, <nil>
--- PASS: TestClientProbe (0.62s)
=== RUN   TestClientProbeThenMap
    portmapper_test.go:51: Probe: {PCP:false PMP:false UPnP:false}, <nil>
    portmapper_test.go:53: CreateOrGetMapping: invalid IPPort, no NAT mapping available: no port mapping services were found
--- PASS: TestClientProbeThenMap (0.26s)

```

Addresses #682 (Not really sure if it _fixes_ it though)

PR #2156, opened 2021-06-17
net/portmapper: make portmapping periodic

JulianKnodt:
In order to prepare for the UPnP change, this allows for stateful probing, which is necessary
for reusing clients, and possibly adding backoff for slower routers.

It is a subset of the changes in #2113, and should probably be merged before them so that change is smaller.

This change is necessary because it will be convenient to hold state in between different probes. Specifically, for UPnP it appears we time out quite frequently, and it may be the case on some slower networks that timeouts may occur for PMP. In order to slow down future probes, the `Prober` should hold some state such as the previous attempt duration, and give future probes longer timeouts.

### DIFF
--- a/net/portmapper/portmapper.go
+++ b/net/portmapper/portmapper.go
@@ -62,6 +62,8 @@ type Client struct {
 
 	localPort  uint16
 	pmpMapping *pmpMapping // non-nil if we have a PMP mapping
+
+	*Prober // non-nil once the prober has started
 }
 
 // HaveMapping reports whether we have a current valid mapping.
--- a/net/portmapper/portmapper.go
+++ b/net/portmapper/portmapper.go
@@ -63,7 +63,7 @@ type Client struct {
 	localPort  uint16
 	pmpMapping *pmpMapping // non-nil if we have a PMP mapping
 
-	*Prober // non-nil once the prober has started
+	prober *Prober // non-nil once probe has been called.
 }
 
 // HaveMapping reports whether we have a current valid mapping.
@@ -131,6 +131,9 @@ func (c *Client) Close() error {
 	if c.closed {
 		return nil
 	}
+	if c.prober != nil {
+		c.prober.Close()
+	}
 	c.closed = true
 	c.invalidateMappingsLocked(true)
 	// TODO: close some future ever-listening UDP socket(s),
@@ -179,6 +182,24 @@ func (c *Client) invalidateMappingsLocked(releaseOld bool) {
 	c.pmpPubIPTime = time.Time{}
 	c.pcpSawTime = time.Time{}
 	c.uPnPSawTime = time.Time{}
+}
+
+// Probe will assess the network for the presence of portmapping services.
+func (c *Client) Probe(ctx context.Context) (ProbeResult, error) {
+	c.mu.Lock()
+	if c.prober == nil {
+		c.initProberLocked(ctx)
+	}
+	c.mu.Unlock()
+	return c.prober.Complete()
+}
+
+func (c *Client) StartProbing() {
+	c.mu.Lock()
+	if c.prober == nil {
+		c.initProberLocked(context.Background())
+	}
+	c.mu.Unlock()
 }
 
 func (c *Client) sawPMPRecently() bool {
@@ -435,19 +456,25 @@ func parsePMPResponse(pkt []byte) (res pmpResponse, ok bool) {
 	return res, true
 }
 
+// ProbeResults indicates which services are present after a probe.
+// The presense of services may change over time, so it represents the presense
+// of these items at a given time.
 type ProbeResult struct {
 	PCP  bool
 	PMP  bool
 	UPnP bool
 }
 
+// oldProbe is the old API for probing, retained in order to ensure back-compatibility.
+// It's currently used in TestProberEquivalent.
+//
 // Probe returns a summary of which port mapping services are
 // available on the network.
 //
 // If a probe has run recently and there haven't been any network changes since,
 // the returned result might be server from the Client's cache, without
 // sending any network traffic.
-func (c *Client) Probe(ctx context.Context) (res ProbeResult, err error) {
+func (c *Client) oldProbe(ctx context.Context) (res ProbeResult, err error) {
 	gw, myIP, ok := c.gatewayAndSelfIP()
 	if !ok {
 		return res, ErrGatewayNotFound
--- a/net/portmapper/probe.go
+++ b/net/portmapper/probe.go
@@ -1,0 +1,332 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package portmapper
+
+import (
+	"context"
+	"net"
+	"sync"
+	"time"
+
+	"go4.org/mem"
+	"inet.af/netaddr"
+	"tailscale.com/net/netns"
+)
+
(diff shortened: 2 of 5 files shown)


net/portmapper/portmapper.go
```diff
@@ -62,6 +62,8 @@ type Client struct {
 
 	localPort  uint16
 	pmpMapping *pmpMapping // non-nil if we have a PMP mapping
+
+	*Prober // non-nil once the prober has started
```
bradfitz, line 66: did you mean to embed this? mean to export it? is it guarded by the mutex? looks like it is by position in the struct, but the caller doesn't know that and doesn't need to obey, as the mutex is unexported (and shouldn't be exported)

net/portmapper/probe.go
```diff
@@ -0,0 +1,332 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package portmapper
```
bradfitz, line 4: blank line above so the copyright isn't a package comment

net/portmapper/probe.go
```diff
@@ -0,0 +1,332 @@
...
+				if upnp_ctx.Err() == context.DeadlineExceeded {
+					err = nil
+				}
+			}
+			cancel()
+			t := time.NewTimer(trustServiceStillAvailableDuration * 3 / 4)
+			p.UPnP.Set(hasUPnP, err)
+
+			select {
+			case should_pause := <-pause:
```
bradfitz, line 104: underscore

net/portmapper/probe.go
```diff
@@ -0,0 +1,332 @@
...
+				}
+			case <-t.C: // break through and retry the connection
+			}
+		}
+	}()
+
+	go func() {
+		defer p.UPnP.Set(false, nil)
+		for {
+			upnp_ctx, cancel := context.WithTimeout(ctx, portMapServiceTimeout)
```
bradfitz, line 87: underscore

net/portmapper/probe.go
```diff
@@ -0,0 +1,332 @@
...
+
+	hasPCP, errPCP := p.PCP.PresentBlock()
+	res.PCP = hasPCP
+	if err == nil {
+		err = errPCP
+	}
+	return
+}
+
+type ProbeSubResult struct {
```
bradfitz, line 168: docs

net/portmapper/probe.go
```diff
@@ -0,0 +1,332 @@
...
+	// If this probe has finished, regardless of success or failure
+	completed bool
+
+	// whether or not this feature is present
+	present bool
+	// most recent error
+	err error
+
+	// time we last saw it to be available.
+	sawTime time.Time
```
bradfitz, line 179: what are all these guarded by?

net/portmapper/probe.go
```diff
@@ -0,0 +1,332 @@
...
+
+func NewProbeSubResult() *ProbeSubResult {
+	return &ProbeSubResult{
+		cond: &sync.Cond{
+			L: &sync.Mutex{},
+		},
+	}
+}
+
+// PresentBlock blocks until the probe completes, then returns the result.
```
bradfitz, line 190: what's the result? what sort of errors can occur?

net/portmapper/probe.go
JulianKnodt: they're guarded by the lock inside of the `Cond`, what's the best way to handle that?

net/portmapper/portmapper.go
JulianKnodt: hmmm yea I was hoping to set it once then make it so that it's never unset, so a lock wouldn't be needed. I changed it so it's guarded by the mutex in the `Probe` function, `NewProber -> newProber`.

net/portmapper/probe.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}()
+}
+
+// Stop gracefully turns the Prober off, completing the current probes before exiting.
+//
+// Calling stop Close multiple times will have no additional effects.
+func (p *Prober) Close() { p.stop() }
+
+// CurrentStatus returns the current results of the prober, regardless of whether they have
```
bradfitz, line 62: what's the error value on incomplete? it's not clear how callers should handle its errors. is it an error to be incomplete? is it an error to not have 1 or more services?

syncs/syncs.go
```diff
@@ -124,3 +126,60 @@ func (s Semaphore) TryAcquire() bool {
...
+	return WaitableResult{
+		cond: &sync.Cond{
+			L: &sync.Mutex{},
+		},
+	}
+}
+
+// Complete blocks until an operation completes, then returns true if it was a success.
+// Otherwise, it returns returns false, with a possible error.
+func (wr *WaitableResult) Complete() (bool, error) {
```
bradfitz, line 156: s/Complete/Get/ ?

And `Current` can be `Peek` ?

syncs/syncs.go
```diff
@@ -124,3 +126,60 @@ func (s Semaphore) TryAcquire() bool {
...
+// not it has completed, as well as the completion time of the operation.
+func (wr *WaitableResult) Current() (time.Time, bool, error) {
+	wr.cond.L.Lock()
+	defer wr.cond.L.Unlock()
+	return wr.sawTime, wr.result, wr.err
+}
+
+// Set should be called when an operation has completed. It will unblock any items waiting for
+// the completed operation, and overwrite previous the results of previous operations.
+func (wr *WaitableResult) Set(result bool, err error) {
```
bradfitz, line 175: maybe just `Set(interface{})` and the client can determine error-ness by its type?

net/portmapper/probe.go
```diff
@@ -0,0 +1,104 @@
...
+
+	_, hasPCP, errPCP := p.PCP.Current()
+	res.PCP = hasPCP
+	if err == nil {
+		err = errPCP
+	}
+	return res, err
+}
+
+// StatusBlock blocks the caller until probing all services has completed, regardless of success
```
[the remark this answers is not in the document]
JulianKnodt: Ah I just didn't mention the return value, but the blocking behaviour. Will include mention of the return value

net/portmapper/probe.go
```diff
@@ -0,0 +1,104 @@
...
+
+// newProber will start a prober if it does not exist on the given portmapping client.
+// Should be called when the client's lock is held.
+func (c *Client) newProber(ctx context.Context) {
+	stop := false
+	p := &Prober{
+		PMP:  syncs.NewWaitableResult(),
+		PCP:  syncs.NewWaitableResult(),
+		UPnP: syncs.NewWaitableResult(),
+		stop: func() { stop = true },
```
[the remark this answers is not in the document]
JulianKnodt: I think it's fine for this to be racey, that did occur to me while writing, but since it's only written to and read from 1 spot, we're fine with it running for 1 more iteration. either way, I've changed it to an AtomicBool so it won't have any racey-ness

syncs/syncs.go
JulianKnodt: Hmmmm I think if this is used in more places having an interface makes more sense, but I think for now adding it will be additional bloat that won't be useful. I can see why it'd be good to have this in the longer term tho

bradfitz:
Is this one still applicable?

JulianKnodt:
Hmmmmmm, I don't think it's strictly necessary with the change in magicsock, so I'll just close it. I think it could still have some use, but not needed

PR #2158, opened 2021-06-17
Adding TestTwoNodeConnectivity, tests connections between two nodes using SOCKS5 proxy

simenghe:
Added test that starts up two nodes, which will also start up a tcp listener using `--socks5-server=localhost:0` for each tailscaled.
Test tries to parse the two addresses from the logs, and test communicating between the two listeners.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,3 +47,5 @@ require (
 	inet.af/wf v0.0.0-20210516214145-a5343001b756
 	rsc.io/goversion v1.2.0
 )
+
+replace golang.zx2c4.com/wireguard => /home/simenghe/gohack/golang.zx2c4.com/wireguard
--- a/tstest/integration/integration_test.go
+++ b/tstest/integration/integration_test.go
@@ -290,6 +290,7 @@ func TestAddPingRequest(t *testing.T) {
 }
 
 func TestTwoNodeConnectivity(t *testing.T) {
+	t.Parallel()
 	bins := BuildTestBinaries(t)
 	env := newTestEnv(t, bins)
 	defer env.Close()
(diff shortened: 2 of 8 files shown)


tstest/integration/integration_test.go
```diff
@@ -287,6 +291,154 @@ func TestAddPingRequest(t *testing.T) {
...
+
+		defer func() {
+			d1.MustCleanShutdown(t)
+			d2.MustCleanShutdown(t)
+			d1.Kill()
+			d2.Kill()
+		}()
+
+		const sub = `SOCK-TEST : `
+		t.Log("Starting Test!")
```
bradfitz, line 329: remove. especially the exclamation mark. I promise it'll be less exciting when we're debugging a test failure in this code later. :)

tstest/integration/integration_test.go
```diff
@@ -287,6 +291,154 @@ func TestAddPingRequest(t *testing.T) {
...
+			conn, err := l.Accept()
+			if err != nil {
+				t.Error(err)
+			}
+			defer conn.Close()
+			_, err = conn.Write([]byte("TestString"))
+			if err != nil {
+				t.Error(err)
+			}
+			t.Logf("Wrote string from %v to %v\n", conn.LocalAddr(), conn.RemoteAddr())
```
bradfitz, line 404: no need for \n in log strings

tstest/integration/integration_test.go
```diff
@@ -405,12 +557,27 @@ func (d *Daemon) MustCleanShutdown(t testing.TB) {
...
-	cmd := exec.Command(n.env.Binaries.Daemon,
-		"--tun=userspace-networking",
-		"--state="+n.stateFile,
-		"--socket="+n.sockFile,
-	)
+// Additional flags to tailscaled can be passed through specialArgs
+func (n *testNode) StartDaemon(t testing.TB, specialArgs ...string) *Daemon {
+	var cmd *exec.Cmd
+	// Support additional flags in starting up tailscaled
+	if len(specialArgs) > 0 {
```
bradfitz, line 564: no need for two duplicated blocks here. It's safe to do the `flags = append(flags, specialArgs...)` if specialArgs is length 0 (including nil)

tstest/integration/integration_test.go
```diff
@@ -287,6 +291,154 @@ func TestAddPingRequest(t *testing.T) {
...
+		// n2IP := n2.AwaitIP(t)
+
+		defer func() {
+			d1.MustCleanShutdown(t)
+			d2.MustCleanShutdown(t)
+			d1.Kill()
+			d2.Kill()
+		}()
+
+		const sub = `SOCK-TEST : `
```
bradfitz, line 328: rather than searching for substrings in the whole output, let's instead register a hook with the logcatcher that we can process per line.

I'll send a change for that part to make this sort of parsing easier in the future

bradfitz:
Sent https://github.com/tailscale/tailscale/pull/2267. Could you review that, and then rebase atop it?

bradfitz:
Please smush this into one commit and mark the review as ready when it's ready and ping me.

simenghe:
Currently need to speed up the process, as it is taking too long to dial, which can cause flakes

simenghe:
> Currently I need to speed up the test, as it is taking too long to dial, which can cause flakes

Update: Read

> Currently I need to speed up the test, as it is taking too long to dial, which can cause flakes

Update: Read function is taking a long time on the select statement in tstun/wrapper. : 
```go
	select {
	case <-t.closed:
		return 0, io.EOF
	case err := <-t.errors:
		return 0, err
	case pkt := <-t.outbound:
		n = copy(buf[offset:], pkt)
		// t.buffer has a fixed location in memory,
		// so this is the easiest way to tell when it has been consumed.
		// &pkt[0] can be used because empty packets do not reach t.outbound.
		if &pkt[0] == &t.buffer[PacketStartOffset] {
			t.bufferConsumed <- struct{}{}
		} else {
			// If the packet is not from t.buffer, then it is an injected packet.
			wasInjectedPacket = true
		}
	}
```

go.mod
```diff
@@ -47,3 +47,5 @@ require (
 	inet.af/wf v0.0.0-20210516214145-a5343001b756
 	rsc.io/goversion v1.2.0
 )
+
+replace golang.zx2c4.com/wireguard => /home/simenghe/gohack/golang.zx2c4.com/wireguard
```
simenghe, line 51: to be removed

PR #2162, opened 2021-06-17
Support GRE L2/L3 tunneling protocol.

tinoue:
This PR supports GRE tunneling protocol over tailscale network.

### DIFF
--- a/net/packet/gre4.go
+++ b/net/packet/gre4.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
--- a/net/packet/gre4.go
+++ b/net/packet/gre4.go
@@ -1,0 +1,44 @@
+// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package packet
+
+import (
+	"tailscale.com/types/ipproto"
+)
+
+// greHeaderLength is the size of the GRE packet header, not including
+// the outer IP header.
+const greHeaderLength = 4
+
+// GRE4Header is an IPv4 header.
+type GRE4Header struct {
+	IP4Header
+}
+
+// Len implements Header.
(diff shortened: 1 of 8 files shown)


net/packet/gre4.go
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
```
Xe, line 1:
```suggestion
// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
```

net/packet/gre6.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
```
Xe, line 1:
```suggestion
// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
```

DentonGentry:
Could you add Signoff lines to the commit descriptions?
https://github.com/tailscale/tailscale/pull/2162/checks?check_run_id=2880344338

wgengine/filter/tailcfg.go
```diff
@@ -18,6 +18,7 @@ var defaultProtos = []ipproto.Proto{
 	ipproto.UDP,
 	ipproto.ICMPv4,
 	ipproto.ICMPv6,
+	ipproto.GRE,
```
bradfitz, line 21: we don't want to add GRE to the defaultProtos. It needs to be opt-in for users, like we did for SCTP.

tinoue: @bradfitz 
Thank you for your comment.
To make GRE as opt-in option is nice idea.

I checked about SCTP and it seems to be allowed by port based ACL.
The problem is GRE doesn't have any port - like ICMP - and current ACL doesn't have protocol based rule.
Do you have any idea ?

amosbird:
Is there any plan to merge this PR?

tinoue:
@amosbird Thanks for reminding me about this PR. Closed becuase the latest tailscale already supports any type of protocol.

amosbird:
@tinoue Great. Starting from which version does tailscale have this ability?

tinoue:
@amosbird  From 1.2.0 according to https://github.com/tailscale/tailscale/issues/2163#issuecomment-997146054
You can specify any protocol byTailscale ACL rule (see proto field): https://tailscale.com/kb/1018/acls/#acl-rules

PR #2164, opened 2021-06-17
net/interfaces: IPv4 link local if nothing better.

DentonGentry:
The only connectivity an AWS Lambda container has is an IPv4 link-local
169.254.x.x address using NAT:
12: vtarget_1@if11: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500
                    qdisc noqueue state UP group default qlen 1000
     link/ether 7e:1c:3f:00:00:00 brd ff:ff:ff:ff:ff:ff link-netnsid 1
     inet 169.254.79.1/32 scope global vtarget_1

If there are no other IPv4/v6 addresses available, and we are running
in AWS Lambda, allow IPv4 169.254.x.x addresses to be used.

----

Similarly, a Google Cloud Run container's only connectivity is
a Unique Local Address fddf:3978:feb1:d745::c001/128.
If there are no other addresses available then allow IPv6
Unique Local Addresses to be used.
We actually did this in an earlier release, but now refactor it to
work the same way as the IPv4 link-local support is being done.


net/interfaces/interfaces.go
```diff
@@ -407,10 +437,10 @@ func GetState() (*State, error) {
 			return
 		}
 		for _, pfx := range pfxs {
-			if pfx.IP().IsLoopback() || pfx.IP().IsLinkLocalUnicast() {
```
bradfitz, line 410: This means the value of State.HaveV4 still changes on non-Lambda environments.

Maybe we need an isUsableV4 (or just fold both into isUsableIP) and do a host.GetEnvType in that func

bradfitz:
@DentonGentry, I submitted the field/func renames separately and rebased this PR (to new commit 52d78ed09d2a7439867ebbdf0a70845c2c9451dd) to make this easier to review.

net/interfaces/interfaces.go
```diff
@@ -117,17 +121,43 @@ func LocalAddresses() (regular, loopback []netaddr.IP, err error) {
 				if tsaddr.IsTailscaleIP(ip) {
 					continue
 				}
-				if ip.IsLinkLocalUnicast() {
-					continue
-				}
 				if ip.IsLoopback() || ifcIsLoopback {
 					loopback = append(loopback, ip)
+				} else if ip.Is4() && ip.IsLinkLocalUnicast() &&
+					hostinfo.GetEnvType() == hostinfo.AWSLambda {
```
bradfitz, line 127: let's put this down on line 152 instead. 

Also, right now LinkLocalUnicast IPv6 can make their way down to regular6 it looks like. 

So instead I'd make this block only about LinkLocalUnicast:

```go
    } else if ip.IsLinkLocalUnicast() {
          if ip.Is4() {
             linklocal4 = append(linklocal4, ip)
          } 
    } ...
```

Then you're just populating info about addresses found and all the policy of what to return as a function of the environment can be down below in one place, around line 152.

net/interfaces/interfaces.go
```diff
@@ -539,7 +573,8 @@ func anyInterestingIP(pfxs []netaddr.IPPrefix) bool {
 // should log in interfaces.State logging. We don't need to show
 // localhost or link-local addresses.
 func isInterestingIP(ip netaddr.IP) bool {
-	if ip.IsLoopback() || ip.IsLinkLocalUnicast() {
+	if ip.IsLoopback() ||
+		(ip.Is4() && ip.IsLinkLocalUnicast() && hostinfo.GetEnvType() == hostinfo.AWSLambda) {
```
bradfitz, line 577: This part of the change seems unnecessary for 1.10. It only controls the interface formatting, and only makes it print less. Removing this doesn't prevent any lambda things from working, right?

net/interfaces/interfaces.go
DentonGentry: Added isUsableV4

net/interfaces/interfaces.go
DentonGentry: Removed for 1.10.

net/interfaces/interfaces.go
DentonGentry: Done.

net/interfaces/interfaces.go
```diff
@@ -117,17 +121,44 @@ func LocalAddresses() (regular, loopback []netaddr.IP, err error) {
...
 	}
+	if len(regular4) == 0 && len(regular6) == 0 {
+		// if we have no usable IP addresses then be willing to accept
+		// addresses we otherwise wouldn't, like:
+		//   + 169.254.x.x (AWS Lambda uses NAT with these)
+		//   + IPv6 ULA (Google Cloud Run uses these with address translation)
+		if hostinfo.GetEnvType() == hostinfo.AWSLambda {
+			regular4 = linklocal4
+		}
+		regular6 = ula6
```
bradfitz, line 159: want to make this conditional on EnvType being Cloud Run, to be conservative?

bradfitz (approved):
LGTM.

Could you update the first line of the commit message as I'd done in https://github.com/tailscale/tailscale/commit/52d78ed09d2a7439867ebbdf0a70845c2c9451dd to add an initial verb (and remove the trailing period)?

net/interfaces/interfaces.go
DentonGentry: We could, however the current 1.8 release allows ULA unconditionally: https://github.com/tailscale/tailscale/blob/e1571ed7e0e84aed16f267be24fe408150109af6/net/interfaces/interfaces.go#L508

We can limit the use of ULA to specific container environments if we want.

DentonGentry:
Going to merge this, to be in 1.10. The IPv6 ULA support is the same as 1.8, albeit refactored. The IPv4 link-local support is new, and limited to AWS Lambda.

PR #2167, opened 2021-06-18
derp: add debug traffic handler

JulianKnodt:
This adds an endpoint for logging traffic from the DERP servers. It reports how many bytes each client of the DERP server has sent/received, as well as their public key. By hitting `/debug/traffic` it will stream the results as JSON back every N seconds, if there were a given number of bytes sent.


derp/derp_server.go
```diff
@@ -450,6 +461,9 @@ func (s *Server) accept(nc Conn, brw *bufio.ReadWriter, remoteAddr string, connN
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	var remoteIPPort netaddr.IPPort
+	remoteIPPort, _ = netaddr.ParseIPPort(remoteAddr)
```
bradfitz, line 465: or just := and ditch `var` line above

derp/derp_server.go
```diff
@@ -913,6 +931,10 @@ type sclient struct {
 	// Owned by sender, not thread-safe.
 	bw *bufio.Writer
 
+	bytesSent *uint64    // accessed atomically
```
bradfitz, line 935: I don't think we'll need these two fields if we go the route of reading `/proc/net/tcp6` instead with a new HTTP debug handler that streams JSON stats.

derp/derp_server.go
```diff
@@ -913,6 +932,10 @@ type sclient struct {
...
+	bytesRecv *uint64    // accessed atomically
+	remoteIP  netaddr.IP // zero if remoteAddr is not ip:port.
```
bradfitz, line 937: move this up to the "static after construction" block, after `canMesh`

derp/derp_server.go
```diff
@@ -141,6 +142,8 @@ type Server struct {
 	// because it includes intra-region forwarded packets as the
 	// src.
 	sentTo map[key.Public]map[key.Public]int64 // src => dst => dst's latest sclient.connNum
+
+	ipMap map[netaddr.IP]key.Public
```
bradfitz, line 146: You'll need to map from `netaddr.IPPort` ... the IP alone isn't unique on the internet. Many users behind a NAT share an IP but have different source ports.

bradfitz: Also, I'd name this more about what it represent and not how (don't need to put "map" in the name).

Like `addrKey` or `keyOfAddr`

derp/derp_server.go
```diff
@@ -141,6 +142,11 @@ type Server struct {
 	// because it includes intra-region forwarded packets as the
 	// src.
 	sentTo map[key.Public]map[key.Public]int64 // src => dst => dst's latest sclient.connNum
+
+	addrOfKey map[netaddr.IPPort]key.Public
```
bradfitz, line 146: If you're going to use "Of" in the name, a `map[A]B` is named `bOfA`, so `keyOfAddr`. The alternative is `addrKey` which can be read like "addr's key".

derp/derp_server.go
```diff
@@ -141,6 +142,11 @@ type Server struct {
...
+
+	statMu       sync.Mutex
+	lastStatTime time.Time
```
bradfitz, line 149: what is this? doc.

derp/derp_server.go
```diff
@@ -141,6 +142,11 @@ type Server struct {
 	// because it includes intra-region forwarded packets as the
 	// src.
 	sentTo map[key.Public]map[key.Public]int64 // src => dst => dst's latest sclient.connNum
+
+	addrOfKey map[netaddr.IPPort]key.Public
+
+	statMu       sync.Mutex
```
bradfitz, line 148: in general, only have one mutex on a type. otherwise you need to aggressively document & follow lock ordering rules if any caller code ever needs to update/read multiple things under different mutex. Otherwise you can deadlock.

derp/derp_server.go
```diff
@@ -343,6 +350,9 @@ func (s *Server) registerClient(c *sclient) {
 	if _, ok := s.clientsMesh[c.key]; !ok {
 		s.clientsMesh[c.key] = nil // just for varz of total users in cluster
 	}
+	if !c.remoteIPPort.IsZero() {
```
bradfitz, line 353: you could also just skip the IsZero check. not the end of the world if you add/remove useless/redundant keys under the key's zero value. Nobody will look it up by that, so it's harmless.

derp/derp_server.go
```diff
@@ -1398,3 +1416,13 @@ func writePublicKey(bw *bufio.Writer, key *key.Public) error {
 	}
 	return nil
 }
+
+func (s *Server) Stat() {
```
bradfitz, line 1420: I don't think you'll need this. All the state can be kept in local variables of the long-running /stat ServeHTTP call.

derp/derphttp/derphttp_server.go
```diff
@@ -20,6 +20,11 @@ const fastStartHeader = "Derp-Fast-Start"
 
 func Handler(s *derp.Server) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodGet && r.URL.Path == "/stat" {
```
bradfitz, line 23: put this in cmd/derper/derper.go's `func debugHandler` instead.

Make it be, say, `/debug/traffic`, and have the few lines in derper.go's just be like:

```go
                if r.RequestURI == "/debug/traffic" {
                       s.ServeDebugTraffic(w, r)
                       return
                }
```

Then add a new that new `ServeDebugTraffic` to derp/derp_server.go (not derphttp_server.go).

derp/derp_server.go
```diff
@@ -1398,3 +1413,14 @@ func writePublicKey(bw *bufio.Writer, key *key.Public) error {
 	}
 	return nil
 }
+
+func (s *Server) ServeDebugTraffic(w http.ResponseWriter, r *http.Request) error {
```
bradfitz, line 1417: remove the error return. then it's just a normal http.Handler.

derp/derp_server.go
```diff
@@ -1398,3 +1413,14 @@ func writePublicKey(bw *bufio.Writer, key *key.Public) error {
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if time.Since(s.lastStatTime) < 5*time.Minute {
+		return nil
+	}
+	s.lastStatTime = time.Now()
+	// TODO log stats here
+  return nil
```
bradfitz, line 1425: You should make your editor gofmt-on-save. (Where your gofmt command is actually goimports)

derp/derp_server.go
```diff
@@ -1398,3 +1413,14 @@ func writePublicKey(bw *bufio.Writer, key *key.Public) error {
 	}
 	return nil
 }
+
+func (s *Server) ServeDebugTraffic(w http.ResponseWriter, r *http.Request) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if time.Since(s.lastStatTime) < 5*time.Minute {
```
bradfitz, line 1420: I don't think we'll need this. it should stream regardless, with updates every N seconds (perhaps configurable over a query param)

derp/derphttp/derphttp_server.go
```diff
@@ -20,6 +20,10 @@ const fastStartHeader = "Derp-Fast-Start"
 
 func Handler(s *derp.Server) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodGet && r.RequestURI == "/debug/traffic" {
```
bradfitz, line 23: Delete. This will only be in cmd/derper.

Also, this will never be routed, as this handler is installed at path /derp, so /debug/* can't get here.

Also, install the route unconditional on method. The handler itself can check the method and fail on bogus ones.

derp/derp_server.go
JulianKnodt: hm maybe, I tend to avoid that and I'll always run fmt before any final changes

derp/derp_server.go
```diff
@@ -1398,3 +1411,83 @@ func writePublicKey(bw *bufio.Writer, key *key.Public) error {
...
+			fmt.Fprintf(w, "ss failed: %v", err)
+			return
+		}
+		newState := parseSSOutput(string(output))
+		for k, next := range newState {
+			prev := prevState[k]
+			if prev.Sent < next.Sent || prev.Recv < next.Recv {
+				if publicKey, ok := s.keyOfAddr.Load(k); ok {
+					pkey := publicKey.(key.Public)
+					if _, err := fmt.Fprintf(w, "%v, %v", next, pkey); err != nil {
```
bradfitz, line 1478: write some JSON object out instead.

And export it with docs.

derp/derp_server.go
```diff
@@ -1398,3 +1411,83 @@ func writePublicKey(bw *bufio.Writer, key *key.Public) error {
...
+			}
+		}
+		newState[src] = stat
+	}
+	return newState
+}
+
+func (s *Server) ServeDebugTraffic(w http.ResponseWriter, r *http.Request) {
+	prevState := map[netaddr.IPPort]bytesSentRecv{}
+	for {
```
bradfitz, line 1466: for r.Context().Err() == nil {

PR #2183, opened 2021-06-21
Create logger for derp bandwidth

JulianKnodt:
This is a small logger which reports the number of bytes received and sent between derp users.
It currently doesn't do anything fancy, it just keeps a running log of all the processes and
prints the top n every 2 seconds.

I also added a change for reporting deltas and not the full amt of bytes written.


JulianKnodt:
I'm also not entirely sure how to move these metrics into grafana, is there a convenient client/library to do that?

PR #2185, opened 2021-06-21
tailcfg: Add user derps field to derpmap

JulianKnodt:
Adds a field to derpmaps which will allow for users to specify their own DERP nodes outside of
those specified by tailscale's regions. Currently this field doesn't do anything, but will be populated later.

### DIFF
--- a/tailcfg/derpmap.go
+++ b/tailcfg/derpmap.go
@@ -14,6 +14,9 @@ type DERPMap struct {
 	//
 	// The numbers are not necessarily contiguous.
 	Regions map[int]*DERPRegion
+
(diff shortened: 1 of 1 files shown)


bradfitz:
This isn't how I'd imagined we'd implement this.

Notably, this would require client updates. Doing it all server-side wouldn't. (Letting users effectively add new regions, reserving certain high region IDs as user-specified)

JulianKnodt:
Since this DERP map is populated server-side, I think it's necessary to add this field here so that when the map is returned from the server it can already be filled. Control currently uses this type, so I wanted to modify it to change what control returns

bradfitz:
Control will return the existing type populated with new regions. No type changes are needed.

Issue #2187, opened 2021-06-22
magicsock: use multiple UDP sockets


bradfitz:
Tailscale usually defaults to port 41641, which people generally like for documentation and ease-of-recognition reasons.

But sometimes dinky NAT devices get overwhelmed when even as many as small handful of clients are all trying to use that port.

Then magicsock probes can report no UDP connectivity.

Magicsock could instead, when a fixed port is requested, listen also on `":0"` and fall back to using it only when the fixed port seems broken.

crawshaw:
Potential fast fix for this customer: port hint from control protocol. We send them an unstable build and see if a random port hint fixes it. Thoughts?

bradfitz:
That's still a bit of work in magicsock even for that. We don't currently have a way to change the magicsock.Options.Port at runtime.

crawshaw:
Ah. So I guess the fast experiment is to ship them an unstable that by default uses a random port.

For the real fix, do we want to make this a control setting? The only reason we use a fixed port is because some admins want to watch the traffic around their network, so that opens up the possibility that we could _default_ to a random port, and let admins choose a fixed port if they want.

bradfitz:
> Ah. So I guess the fast experiment is to ship them an unstable that by default uses a random port.

Doing it by default for everybody? That risks breaking people who are currently depending on the fixed port.

e.g. https://tailscale.com/kb/1082/firewall-ports/ already says to open up 41641.

I think the right fix is to try to use 41641 and fall back to random if it doesn't work.

crawshaw:
I think the fast experiment is to ship this one customer a binary that does it properly and see what happens.

For the right fix, how do you detect 41641 doesn't work? Right now we see two separate nodes, both connected, reporting they are on the endpoint <public IP>:41641. One of them even reports another endpoint on the same NATted IP with a different port. The breakage is an upstream router, so magicsock is going to have to receive some information from some network prober or the control plane.

crawshaw:
Presumably in the linked case, if the upstream NAT simply never let packets through 41641 for a client, then it would fall back to DERP and everything would be fine, if slow. Instead we see unreachable machines, suggesting the NAT let traffic through for a short amount of time then changed its mind when another tailscale client behind the NAT caused it to reset.

My guess is they are bouncing around in the NAT's tiny routing memory very quickly.

crawshaw:
[...]

PR #2222, opened 2021-06-22
tailcfg: add Debug.RandomizeClientPort

bradfitz:
Not yet used.

Updates #2187

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -1055,6 +1055,11 @@ type Debug struct {
 	// The client can (and should) limit the value (such as 5
 	// minutes).
 	SleepSeconds float64 `json:",omitempty"`
+
+	// RandomizeClientPort is whether magicsock should UDP bind to
+	// :0 to get a random local port, ignoring any configured
+	// fixed port.
+	RandomizeClientPort bool `json:",omitempty"`
 }
 
 func (k MachineKey) String() string                   { return fmt.Sprintf("mkey:%x", k[:]) }


tailcfg/tailcfg.go
```diff
@@ -1055,6 +1055,11 @@ type Debug struct {
 	// The client can (and should) limit the value (such as 5
 	// minutes).
 	SleepSeconds float64 `json:",omitempty"`
+
+	// RandomizeClientPort is whether magicsock should UDP bind to
+	// :0 to get a random local port, ignoring any configured
```
crawshaw, line 1060: Right now tailscaled has a `--port` which defaults to 0, meaning "we pick", and we pick 41641. This overrides it so we pick a random port. What if a user sets `--port 9999`? Does this override that?

bradfitz: Yes, that's why I wrote "ignoring any configured"

bradfitz: > Right now tailscaled has a --port which defaults to 0, meaning "we pick", and we pick 41641

No, the systemd config we use passes --port=41641 (see https://github.com/tailscale/tailscale/blob/main/cmd/tailscaled/tailscaled.service and https://github.com/tailscale/tailscale/blob/main/cmd/tailscaled/tailscaled.defaults)

crawshaw: OK, then this isn't a feature we would want to use widely, definitely not be default for new users. But maybe that's fine because we want to do the double-port thing in the future or work around these NATs some other way?

crawshaw (approved):

PR #2223, opened 2021-06-22
magicsock, etc: randomize local client port if server requests we do

crawshaw:


syncs/syncs.go
```diff
@@ -83,6 +83,17 @@ func (b *AtomicBool) Get() bool {
 	return atomic.LoadInt32((*int32)(b)) != 0
 }
 
+// AtomicUint16 is an atomic uint16.
+type AtomicUint16 int32
```
bradfitz, line 87: I'd just make this AtomicInt32 and let the caller shove it into a uint16

wgengine/magicsock/magicsock.go
```diff
@@ -2097,6 +2103,15 @@ func (c *Conn) SetNetworkUp(up bool) {
 	}
 }
 
+// SetPreferredPort sets the connection's preferred local port.
+func (c *Conn) SetPreferredPort(port uint16) {
+	if curPort := c.port.Get(); curPort == port {
```
bradfitz, line 2108: can drop curPort and just `if c.port.Get() == port`

wgengine/userspace.go
```diff
@@ -794,7 +794,9 @@ func (e *userspaceEngine) Reconfig(cfg *wgcfg.Config, routerCfg *router.Config,
 	if err := e.magicConn.SetPrivateKey(wgkey.Private(cfg.PrivateKey)); err != nil {
 		e.logf("wgengine: Reconfig: SetPrivateKey: %v", err)
 	}
-	e.magicConn.UpdatePeers(peerSet)
+	if cfg.Debug != nil && cfg.Debug.RandomizeClientPort {
```
bradfitz, line 797: this doesn't ever restore it.

how about you set/unset an AtomicBool on magicsock instead and do the `if bool { port = 0 }` later? Then this can just Set(true)  and Set(false)

wgengine/wgcfg/config.go
```diff
@@ -25,6 +25,7 @@ type Config struct {
 	MTU        uint16
 	DNS        []netaddr.IP
 	Peers      []Peer
+	Debug      *tailcfg.Debug
```
bradfitz, line 28: How about just a bool instead? But seems a little weird putting this on wgcfg.Config.

crawshaw: PTAL

crawshaw:
PTAL, now with test!

wgengine/userspace.go
```diff
@@ -794,7 +801,7 @@ func (e *userspaceEngine) Reconfig(cfg *wgcfg.Config, routerCfg *router.Config,
 	if err := e.magicConn.SetPrivateKey(wgkey.Private(cfg.PrivateKey)); err != nil {
 		e.logf("wgengine: Reconfig: SetPrivateKey: %v", err)
 	}
-	e.magicConn.UpdatePeers(peerSet)
```
bradfitz, line 797: where did this go?

wgengine/magicsock/magicsock.go
```diff
@@ -2097,6 +2103,32 @@ func (c *Conn) SetNetworkUp(up bool) {
...
 }
 
+// SetPreferredPort sets the connection's preferred local port.
+func (c *Conn) SetPreferredPort(port uint16) {
+	if uint16(c.port.Get()) == port {
+		return
+	}
+	c.port.Set(uint32(port))
+
+	if err := c.bindSocket(&c.pconn4, "udp4", true); err != nil {
```
bradfitz, line 2113: why the pconn4 before the the portMapper.SetLocalPort and then the pconn6? something subtle happening?

wgengine/magicsock/magicsock.go
```diff
@@ -2097,6 +2103,32 @@ func (c *Conn) SetNetworkUp(up bool) {
...
+		c.logf("magicsock: SetPreferredPort IPv4 failed: %w", err)
+		return
+	}
+	c.portMapper.SetLocalPort(c.LocalPort())
+	if err := c.bindSocket(&c.pconn6, "udp6", true); err != nil {
+		c.logf("magicsock: SetPreferredPort ignoring IPv6 bind failure: %v", err)
+	}
+
+	c.mu.Lock()
+	c.closeAllDerpLocked("setpreferredport")
```
bradfitz, line 2123: why does closing DERP matter?

wgengine/userspace.go
crawshaw: oops. restored. Worrying that no tests failed.

wgengine/magicsock/magicsock.go
crawshaw: This code was matching Rebind. I factored it out to make that clear.

I believe the idea is: we treat rebinding IPv4 as vital, and rebinding IPv6 as nice-to-have. So as soon as bindSocket works for udp4, we change the local conn to let STUN do its thing, then spend some time trying to get udp6 to optionally work.

wgengine/magicsock/magicsock.go
crawshaw: Removed.

crawshaw:
PTAL

bradfitz (approved):
LGTM. I could do without all the new bool parameters and random boolean literals at call sites, but I'll hold my nose for now.

crawshaw:
I added a type for the boolean

bradfitz (commented):
Thanks :)

Could make its concrete type be uint8 instead of bool so it can't be used in boolean context! 😛

Issue #1264, opened 2021-02-04
Let users add/replace DERP servers for their network


bradfitz:
Tailscale (the company) runs DERP relays around the world, such that there's always one near users, but various users/organizations have requested the ability to modify the DERP map for their network, either:

* augmenting our DERP map with DERP nodes that they run (using the open source https://pkg.go.dev/tailscale.com/cmd/derper server)
* removing certain DERP servers from our set (e.g. remove some in certain countries, for regulatory reasons), including removing all but certain geos
* some combination of both: like removing all ours, and only using the ones the user supplied.

An easy way to this is to add some fields to the ACL JSON object (which we could rebrand as the "tailnet policy config" or "Tailnet ACL and policy config").

(We've discussed this in various forums, but creating this public tracking bug)

crawshaw:
Should the DERP server be part of tailscaled? A user could enable a DERP server in the admin panel (and we could surface test results for "is this machine publicly reachable?" that all DERP servers should meet).

bradfitz:
Yeah, that could be cool!

talway:
Derp Relays can be packaged and installed directly or optionally

talway:
Is there any new progress? The network connection speed is too slow now. I hope this function can be realized as soon as possible
@bradfitz

bradfitz:
Any progress would be posted here. We have many things we'd like to do. This, while somewhat easy and fun, isn't super high on the list. Many more critical or impactful projects are above it. That prioritization could change at a moment's notice, though, particularly if a customer requires it. We'll post updates here when something is available.

hronro:
This is the feature I want most.

And is there a suggestion about how to contribute to this? And then I can see if I can contribute to this in my spare time.

talway:
Is there any new progress?

talway:
> 这是我最想要的功能。
> 
> 是否有关于如何为此做出贡献的建议？然后，我可以看到，如果我可以贡献这一点，在我的业余时间。

Can you compile standalone server software?

DentonGentry:
This issue is not being actively worked on at this time.

It seems like you're enthused about the possibility of adding a DERP server. Could I ask where you'd want it to be? Have you noticed long latency reaching one of the existing DERP servers?

talway:
Yes, we need fast connection speed, which you cannot meet at present, so we hope to integrate and package this function.

talway:
Could you please speed up the implementation of this function or simply put it online? Thank you

[...]

PR #2225, opened 2021-06-22
tailcfg/derpmap: add flag to omit ts's derps

JulianKnodt:
This adds a flag to derp maps which specifies that default Tailscale DERP servers should not be
used. If true and there are entries in this map, it indicates that the entries in this map
should take precedent and not hit any of tailscale's DERP servers.

This change is backwards compatible, as the default behavior should be false.

Updates #1264

### DIFF
--- a/tailcfg/derpmap.go
+++ b/tailcfg/derpmap.go
@@ -14,6 +14,10 @@ type DERPMap struct {
 	//
 	// The numbers are not necessarily contiguous.
 	Regions map[int]*DERPRegion
+
+	// OmitDefaultRegions specifies to not use Tailscale's DERP servers, and only use those
+	// specified in this DERPMap. If there are none set outside of the defaults, this is a noop.
+	OmitDefaultRegions bool `json:"omitDefaultRegions,omitempty"`
 }
 
 /// RegionIDs returns the sorted region IDs.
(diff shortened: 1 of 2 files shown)


tailcfg/derpmap.go
```diff
@@ -13,7 +13,11 @@ type DERPMap struct {
 	// It's keyed by the DERPRegion.RegionID.
 	//
 	// The numbers are not necessarily contiguous.
-	Regions map[int]*DERPRegion
+	Regions map[int]*DERPRegion `json:"regions"`
+
+	// OmitDefaultRegions specifies to not use Tailscale's DERP servers, and only use those
+	// specified in this DERPMap. If there are none set outside of the defaults, this is a noop.
+	OmitDefaultRegions bool `json:"omitDefaultRegions"`
```
bradfitz, line 20: Add `,omitempty` here; so we don't start sending this field to clients.

tailcfg/derpmap.go
```diff
@@ -13,7 +13,11 @@ type DERPMap struct {
 	// It's keyed by the DERPRegion.RegionID.
 	//
 	// The numbers are not necessarily contiguous.
-	Regions map[int]*DERPRegion
+	Regions map[int]*DERPRegion `json:"regions"`
```
bradfitz, line 16: I think you can probably drop this change. By default Go matches either case on parse, so we can still have people write `regions` if that's prettier. But I'm worried about changing what we send to people if we change this.

bradfitz:
In the commit message, add line:

```
Updates #1264
```

(generally, try to cross-reference everything as much as possible)

bradfitz:
Otherwise LGTM

PR #2226, opened 2021-06-23
net/dns{,/resolver}: refactor DNS forwarder, send out of right link on macOS/iOS

### LINKED ISSUE #2224: Split DNS configuration is pointing at a DNS server hosted on Tailscale.
DNS/split DNS  is failing after upgrading to 1.8.7 on macOS/IOS.

Failed to resolve the split Domain. 

Workaround: Disable the "Override Local DNS " option from the admin console. (please verify another node of windows/android/Linux/etc while you change the DNS config)

Front logo [Front conversations](https://app.frontapp.com/open/top_3ajwx)

bradfitz:
Fixes #2224
Fixes tailscale/corp#2045

### DIFF
--- a/net/dns/resolver/forwarder.go
+++ b/net/dns/resolver/forwarder.go
@@ -9,40 +9,29 @@ import (
 	"context"
 	"encoding/binary"
 	"errors"
-	"fmt"
 	"hash/crc32"
+	"io"
 	"math/rand"
 	"net"
 	"sync"
-	"syscall"
 	"time"
 
 	dns "golang.org/x/net/dns/dnsmessage"
 	"inet.af/netaddr"
-	"tailscale.com/logtail/backoff"
 	"tailscale.com/types/logger"
 	"tailscale.com/util/dnsname"
+	"tailscale.com/wgengine/monitor"
 )
 
 // headerBytes is the number of bytes in a DNS message header.
 const headerBytes = 12
 
-// connCount is the number of UDP connections to use for forwarding.
-const connCount = 32
-
 const (
-	// cleanupInterval is the interval between purged of timed-out entries from txMap.
-	cleanupInterval = 30 * time.Second
 	// responseTimeout is the maximal amount of time to wait for a DNS response.
 	responseTimeout = 5 * time.Second
 )
 
 var errNoUpstreams = errors.New("upstream nameservers not set")
-
-type forwardingRecord struct {
-	src       netaddr.IPPort
-	createdAt time.Time
-}
 
 // txid identifies a DNS transaction.
 //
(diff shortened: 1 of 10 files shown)


bradfitz:
Not yet ready for review. `cmd/tailscaled` compiles, but I haven't even run the tests locally, much less tested it by hand. (the original hacked up version with the sockopt worked with manual testing until I started to plumb it and arrived here)

net/dns/resolver/macios_ext.go
```diff
@@ -0,0 +1,51 @@
...
+		})
+		if err != nil {
+			return err
+		}
+		return sockErr
+	}
+	return nil
+}
+
+func bindIf(fd uintptr, network, address string, ifIndex int) error {
```
bradfitz, line 42: I suppose I should unify this with the peerapi code that does the same.

wgengine/userspace.go
```diff
@@ -1362,3 +1363,18 @@ func (p closeOnErrorPool) closeAllIfError(errp *error) {
...
+
+// fwdDNSlinkSelector is userspaceEngine's resolver.ForwardLinkSelector, to pick
+// which network interface to send DNS queries out of.
+type fwdDNSlinkSelector struct {
+	ue      *userspaceEngine
+	tunName string
+}
+
+func (ls fwdDNSlinkSelector) PickLink(ip netaddr.IP) (linkName string) {
+	n, err := ls.ue.peerForIP(ip)
```
bradfitz, line 1375: This is likely wrong for exit nodes. I don't think peerForIP handles that sufficiently.

net/dns/resolver/forwarder.go
```diff
@@ -283,225 +243,60 @@ func (f *forwarder) forward(query packet) error {
...
-			c.mu.Lock()
-			// It's possible that multiple senders discovered simultaneously
-			// that the network is unreachable. Avoid reconnecting multiple times:
-			// Only reconnect if the current connection is the one that we
-			// discovered to be problematic.
-			if c.conn == conn {
-				backOff(err)
-				c.reconnectLocked()
+	ctx, cancel := context.WithTimeout(f.ctx, responseTimeout)
+	defer cancel()
```
crawshaw, line 252: Not sure how pedantic you want to be in this code, but it looks like all the resolver sends that haven't returned get cleaned up by context cancellation after this method, `forward` has returned, meaning there are for some amount of time, some dangling goroutines. Might make writing a test that does goroutine accounting difficult.

net/dns/resolver/forwarder.go
```diff
@@ -283,225 +243,60 @@ func (f *forwarder) forward(query packet) error {
...
+			return ctx.Err()
+		case f.responses <- packet{v, query.addr}:
+			return nil
 		}
-		if errors.Is(err, net.ErrClosed) {
-			// We intentionally closed this connection.
-			// It has been replaced by a new connection. Try again.
-			continue
+	case <-ctx.Done():
+		closeOnCtxDone.Close()
```
crawshaw, line 289: I believe it's theoretically possible for a send goroutine to add something to closeOnCtxDone after this call to close, if the context cancellation came after a send goroutine executed lc.ListenPacket but before it executed closeOnCtxDone.Add(conn).

Maybe make closePool track its closed state and close anything Added immediately if closed?

crawshaw (approved):

net/dns/resolver/tsdns.go
```diff
@@ -291,53 +290,34 @@ func (r *Resolver) resolveLocal(domain dnsname.FQDN, typ dns.Type) (netaddr.IP,
...
 
-	var pkt packet
-	for {
-		select {
-		case <-r.closed:
+	out, err := r.respond(pkt.bs)
+	if err == errNotOurName {
+		err = r.forwarder.forward(pkt)
+		if err == nil {
+			// forward will send response into r.responses, nothing to do.
```
sailorfrag, line 308: Would it make sense for forward to also send errors to r.errors?

net/dns/resolver/tsdns_test.go
```diff
@@ -211,8 +214,12 @@ func TestRDNSNameToIPv6(t *testing.T) {
 	}
 }
 
+func newResolver(t testing.TB) *Resolver {
+	return New(t.Logf, nil /* no link monitor */, nil /* no link selector */)
```
sailorfrag, line 218: Should there be a test that does set up a link selector?

net/dns/resolver/forwarder.go
```diff
@@ -100,178 +89,149 @@ func getTxID(packet []byte) txid {
...
-		// At most one nameserver will return a response:
-		// the first one to do so will delete txid from the map.
-		if !found {
-			f.mu.Unlock()
-			continue
+	// The 1 extra byte is to detect packet truncation.
+	out := make([]byte, maxResponseBytes+1)
+	n, _, err := conn.ReadFrom(out)
+	if err != nil {
+		if err := ctx.Err(); err != nil {
```
sailorfrag, line 186: Should it check `ctx.Err()` even when `ReadFrom` doesn't return an error?

net/dns/resolver/forwarder.go
```diff
@@ -100,178 +89,149 @@ func getTxID(packet []byte) txid {
...
-
-		record, found := f.txMap[txid]
-		// At most one nameserver will return a response:
-		// the first one to do so will delete txid from the map.
-		if !found {
-			f.mu.Unlock()
-			continue
+	// The 1 extra byte is to detect packet truncation.
+	out := make([]byte, maxResponseBytes+1)
+	n, _, err := conn.ReadFrom(out)
```
sailorfrag, line 184: Should there be a `conn.SetDeadline()` so that the `WriteTo`/`ReadFrom` will abort at the same time as the context's timeout?

net/dns/resolver/forwarder.go
bradfitz: discussed in teamflow; should be sufficient. any tests counting goroutines needs a tiny retry loop as is, and all goroutines under this end quickly on cancelation.

net/dns/resolver/forwarder.go
bradfitz: Done.

net/dns/resolver/forwarder.go
bradfitz: I'd rather not pay that cost in the common case. It's not much cost but it's also misleading, as the caller is already selecting on both our success and the context being closed anyway. This check here is about returning the most appropriate error cause and not so much about checking whether the context was closed.

net/dns/resolver/forwarder.go
bradfitz: The closeOnCtxDone handles that, and also handles other reasons a context might be done (e.g. explicit context.WithCancel CancelFunc being called).

net/dns/resolver/tsdns.go
bradfitz: I kept it the same as it was before. Had to draw the cleanup line somewhere. :)

net/dns/resolver/forwarder.go
sailorfrag: Ohh, right, it'll return on something else calling Close. OK, nevermind.

net/dns/resolver/macios_ext.go
bradfitz: Done. Moved to net/netns too, and unified.

net/dns/resolver/tsdns_test.go
bradfitz: Done.

net/dns/resolver/macios_ext.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"errors"
+	"net"
+
+	"tailscale.com/net/netns"
+	"tailscale.com/wgengine/monitor"
+)
+
+func init() {
+	initListenConfig = initListenConfigNetworkExtension
```
danderson, line 18: Why do it like this rather than a file per platform like the rest of our code? No strong feeling from me.

net/dns/resolver/tsdns.go
```diff
@@ -27,10 +28,19 @@ import (
 // truncation in a platform-agnostic way.
 const maxResponseBytes = 4095
 
-// queueSize is the maximal number of DNS requests that can await polling.
+// maxActiveQueries returns the maximal number of DNS requests that be running.
```
danderson, line 31: missing a word ("can" ?)

danderson (approved):

net/dns/resolver/forwarder.go
```diff
@@ -100,178 +89,149 @@ func getTxID(packet []byte) txid {
...
+	var linkName string
+	if f.linkSel != nil {
+		linkName = f.linkSel.PickLink(dst.IP())
+	}
 
-		out = out[:n]
-		txid := getTxID(out)
+	// Make a new connection.
+	var lc net.ListenConfig
+	if initListenConfig != nil && linkName != "" {
```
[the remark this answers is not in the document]
bradfitz: which part? reading initListenConfig? that read is done ages after the package initialization `func init` work.

net/dns/resolver/macios_ext.go
bradfitz: you mean making it a `func` rather than a func variable?

two reasons: can't check for `nil` at runtime if it's a func, and can't replace it in tests (as I did for Adrian in the latest rev)

net/dns/resolver/tsdns.go
bradfitz: hah, yeah. Although I do like "that be runnin'".

PR #2227, opened 2021-06-23
cmd/cloner: support maps with clone ptrs

JulianKnodt:
In order to clone DERPMaps, it was necessary to extend the cloner so that it supports
nested pointers inside of maps which are also cloneable. This also adds cloning for DERPRegions
and DERPNodes because they are on DERPMap's maps.


crawshaw (approved):

bradfitz:
LGTM

Issue #2109, opened 2021-06-11
net/portmapper: UPnP not detected properly


JulianKnodt:
I have UPnP enabled on my home router, but the current `net/portmapper` does not detect it. I initially thought this was due to timeout, as I was able to properly detect it using an [external library](https://github.com/huin/goupnp), but it takes significantly longer than the given timeout. I double checked on `main` with 10 seconds timeout, but this also does not properly detect UPnP's presence.

Issue #682, opened 2020-08-13
UPnP support


danderson:
We should support UPnP. This is a dumping ground of notes for how to speak UPnP right.

- Send UDP SSDP discovery request to :1900 on the default gateway. One of the responses will be for ServiceType `urn:schemas-upnp-org:device:InternetGatewayDevice:1` and will point to the service description URL, e.g. http://192.168.1.1:5000/Public_UPNP_gatedesc.xml .
- Read service description XML. Look for either WANIPConnection (preferentially) or WANPPPConnection (fallback, older devices). Remember its service type and control URL.
- Things you should technically do, to verify that the gateway actually has an internet connection - but you can also just skip that and handle errors in portmap requests.
  - Hit the Connection's control URL with a GetStatus SOAP request. Proceed with port mapping if state==1 (IGD connected). state==2 (not connected) or 3 (unknown) will error out if you ask for a port mapping (maybe just try anyway and handle the error instead of discovering conn status?)
  - Hit the Connection's control URL with a GetExternalIP request. If no external IP is returned, don't ask for a port mapping.
- Hit the Connection's control URL with a AddAnyPortMapping request, with params that are either the port we would like, or the port we were last allocated, and some short-ish lease. We'll get back a public ip:port, which we should feed to control.
  - AddAnyPortMapping is only in UPnP IGD 2.0. If it's not present (we can check the WAN{PPP,IP}Connection schema URL to find out), we have to use AddPortMapping instead. We have to pick random ports and repeatedly ask until we find a free one, unlike AddAnyPortMapping which will pick a free port if the one we request is infeasible.
- Keep hitting the Connection with GetListOfPortMappings, to check that our portmap is still there, and eagerly refresh ahead of schedule if it vanishes (e.g. gateway reboots).
  - Technically I think we can also subscribe to multicast announcements to discover state changes, but ugh multicast, ugh HTTP over UDP, etc. Given that active discovery will fall back on failure, a periodic re-query seems better? Need to balance mapping correctness with battery usage on phone.
- Refresh the mapping with AddAnyPortMapping/AddPortMapping as needed before the lease expires. UPnP suggests a 1h lease time, but in this age of very mobile devices that roam networks, something shorter feels more polite?...
[...]

PR #2232, opened 2021-06-23
net/portmapper: add upnp port mapping

JulianKnodt:
This is a subset of my original PR (#2113) which had too many design changes.
I've distilled most of it into just the changes necessary for portmapping.

The code from the fork is a bit messy, and I'm not sure whether I should go and clean it up.


net/portmapper/portmapper.go
```diff
@@ -60,15 +60,29 @@ type Client struct {
...
 	uPnPSawTime time.Time // time we last saw UPnP was available
 
-	localPort  uint16
-	pmpMapping *pmpMapping // non-nil if we have a PMP mapping
+	localPort uint16
+
+	mapping mapping // non-nil if we have a mapping
+}
+
+// Mapping represents a created port-mapping over some protocol.  It specifies a lease duration,
```
bradfitz, line 68: lowercase "mapping"

net/portmapper/portmapper.go
```diff
@@ -60,15 +60,29 @@ type Client struct {
...
+// how to release the mapping, and whether the map is still valid.
+type mapping interface {
+	// isCurrent indicates whether a mapping is valid now, since most mappings have a lease.
```
bradfitz, line 71: "reports whether" (Go doc convention for bool funcs)

net/portmapper/portmapper.go
```diff
@@ -60,15 +60,29 @@ type Client struct {
...
+	isCurrent() bool
+	// release will attempt to unmap the established port mapping.
```
bradfitz, line 73: "will attempt" and no return value implies this is async? Clarify. Does this block?

net/portmapper/upnp.go
```diff
@@ -0,0 +1,203 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package portmapper
```
bradfitz, line 4: blank line above, otherwise the copyright becomes the package comment

net/portmapper/upnp.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+func (u *upnpMapping) isCurrent() bool                { return u.useUntil.After(time.Now()) }
+func (u *upnpMapping) validUntil() time.Time          { return u.useUntil }
+func (u *upnpMapping) externalIPPort() netaddr.IPPort { return u.external }
+func (u *upnpMapping) release() {
+	u.client.DeletePortMapping(context.Background(), "", u.external.Port(), "udp")
+}
+
+// UPnP Client is an interface over the multiple different clients exported by goupnp,
```
bradfitz, line 34: // upnpClient is ...

net/portmapper/portmapper.go
```diff
@@ -60,15 +60,29 @@ type Client struct {
...
-	localPort  uint16
-	pmpMapping *pmpMapping // non-nil if we have a PMP mapping
+	localPort uint16
+
+	mapping mapping // non-nil if we have a mapping
+}
+
+// Mapping represents a created port-mapping over some protocol.  It specifies a lease duration,
+// how to release the mapping, and whether the map is still valid.
+type mapping interface {
```
bradfitz, line 70: What are the concurrency rules for implementations? Looks like they're not safe for use by concurrent goroutines?

net/portmapper/upnp.go
```diff
@@ -0,0 +1,203 @@
...
+	internal netaddr.IPPort
+	useUntil time.Time
+	client   upnpClient
+}
+
+func (u *upnpMapping) isCurrent() bool                { return u.useUntil.After(time.Now()) }
+func (u *upnpMapping) validUntil() time.Time          { return u.useUntil }
+func (u *upnpMapping) externalIPPort() netaddr.IPPort { return u.external }
+func (u *upnpMapping) release() {
+	u.client.DeletePortMapping(context.Background(), "", u.external.Port(), "udp")
```
bradfitz, line 31: oh, so this _can_ block. In that case release should take a context, rather than assuming background.

net/portmapper/upnp.go
```diff
@@ -0,0 +1,203 @@
...
+	AddPortMapping(
+		ctx context.Context,
+		remoteHost string,
+		externalPort uint16,
+		protocol string,
+		internalPort uint16,
+		internalClient string,
+		enabled bool,
+		portMappingDescription string,
+		leaseDuration uint32,
```
bradfitz, line 46: duration in what? Go code should use time.Duration in general. But if we must use something else, it needs to say the unit.

bradfitz (commented):
commit message: lowercase "add" verb after colon

net/portmapper/upnp.go
```diff
@@ -0,0 +1,208 @@
...
+		return err
+	})
+	var ppp1Clients []*internetgateway2.WANPPPConnection1
+	tasks.Go(func() error {
+		var err error
+		ppp1Clients, _, err = internetgateway2.NewWANPPPConnection1Clients()
+		return err
+	})
+
+	err := tasks.Wait()
```
bradfitz, line 124: This blocks potentially forever, as your three tasks don't use `ctx` (or `errgroup.WithContext`'s returned context, which you assign to `_`)

net/portmapper/upnp.go
```diff
@@ -0,0 +1,208 @@
...
+// Adapted from https://github.com/huin/goupnp/blob/master/GUIDE.md.
+func getUPnPClient(ctx context.Context) (upnpClient, error) {
+	tasks, _ := errgroup.WithContext(ctx)
+	// Attempt to connect over the multiple available connection types concurrently,
+	// returning the fastest.
+	var ip1Clients []*internetgateway2.WANIPConnection1
+	tasks.Go(func() error {
+		var err error
+		ip1Clients, _, err = internetgateway2.NewWANIPConnection1Clients()
+		return err
```
bradfitz, line 109: do you want the failure of any one of these three to cancel everything? As written, that's what it kinda looks like you're trying to do. (You're not using errgroup.Group in quite the normal way, though, so these returned errors aren't actually canceling anyting. But once you start using the ctx, then you'd want to return nil here if it could just be some sort of "nothing exists" error... I don't know what it returns)

net/portmapper/upnp.go
```diff
@@ -0,0 +1,208 @@
...
+	case len(ppp1Clients) > 0:
+		return ppp1Clients[0], nil
+	default:
+		// Didn't get any outputs, report if there was an error or nil if
+		// just no clients.
+		return nil, err
+	}
+}
+
+// getUPnPPortMapping will attempt to create a port-mapping over the UPnP protocol. On success,
```
bradfitz, line 140: "attempts to" ... looks like it does it right away, not in the future ("will")

net/portmapper/upnp.go
```diff
@@ -0,0 +1,208 @@
...
+func (u *upnpMapping) validUntil() time.Time          { return u.useUntil }
+func (u *upnpMapping) externalIPPort() netaddr.IPPort { return u.external }
+func (u *upnpMapping) release(ctx context.Context) {
+	u.client.DeletePortMapping(ctx, "", u.external.Port(), "udp")
+}
+
+// upnpClient is an interface over the multiple different clients exported by goupnp,
+// exposing the functions we need for portmapping. They are auto-generated from XML-specs.
+type upnpClient interface {
+	AddPortMapping(
```
bradfitz, line 41: can you document these params, perhaps as same-line `//` comments, with example values and documentation of special values?

net/portmapper/upnp.go
```diff
@@ -0,0 +1,208 @@
...
+		ctx context.Context,
+		remoteHost string,
+		externalPort uint16,
+		protocol string,
+		internalPort uint16,
+		internalClient string,
+		enabled bool,
+		portMappingDescription string,
+		// Lease duration in seconds exposed by goupnp.
```
bradfitz, line 50: what does 0 mean? anything?

net/portmapper/portmapper.go
```diff
@@ -167,11 +189,11 @@ func (c *Client) gatewayAndSelfIP() (gw, myIP netaddr.IP, ok bool) {
 }
 
 func (c *Client) invalidateMappingsLocked(releaseOld bool) {
-	if c.pmpMapping != nil {
+	if c.mapping != nil {
 		if releaseOld {
-			c.pmpMapping.release()
+			c.mapping.release(context.Background())
```
bradfitz, line 194: do we want some timeout on the context here?

bradfitz (commented):
I'll take another look tonight

net/portmapper/portmapper.go
JulianKnodt: I think we don't particularly care if this fails, maybe I should just fire this off in a goroutine with some deadline?

net/portmapper/upnp.go
JulianKnodt: Yea, none of these `New*Clients` calls actually takes a context so it's a bit hard to handle properly. I was hoping to make it so that if all of them error, it returns any of the errors. If there is at least one success, just take that client and ignore other errors.

net/portmapper/upnp.go
JulianKnodt: I found the spec really confusing here, but I documented what I think it means.

net/portmapper/upnp.go
JulianKnodt: hm I've tried to fix this but I suspect what I did is kind of sketchy

net/portmapper/upnp.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// upnpClient is an interface over the multiple different clients exported by goupnp,
+// exposing the functions we need for portmapping. They are auto-generated from XML-specs.
+type upnpClient interface {
+	AddPortMapping(
+		ctx context.Context,
+
+		// remoteHost is the remote device sending packets to this device, in the format of x.x.x.x,
+		// where any host, aka a wildcard, is the empty string: "".
```
bradfitz, line 45: // The empty string means any host out on the internet can send packets in.

net/portmapper/upnp.go
```diff
@@ -0,0 +1,237 @@
...
+		remoteHost string,
+
+		// externalPort is the exposed port of this port mapping. Visible during NAT operations.
```
bradfitz, line 48: Does 0 mean the router (NAT device) selects automatically? That's what it means for NAT-PMP/PCP.

net/portmapper/upnp.go
bradfitz: Ultimately my concern is that `internetgateway2.NewWANIPConnection1Clients()` etc don't take a context. How long can they run?

net/portmapper/upnp.go
JulianKnodt: I think in general yes, updated the docs.

JulianKnodt:
I can just add a context to them in our fork, let me do that

JulianKnodt:
It ultimately boils down to here where the Connection deadline is being set:
https://github.com/tailscale/goupnp/blob/39c5a55db683675f3329e7d8f13c8094592b3be9/httpu/httpu.go#L99

net/portmapper/upnp.go
JulianKnodt: See comments on main thread: I've added a context to these calls, and previously they had a timeout set manually inside of the library.

bradfitz:
I looked at https://github.com/tailscale/goupnp/commit/39c5a55db683675f3329e7d8f13c8094592b3be9 but it looks like you still need to update https://github.com/tailscale/goupnp/blob/39c5a55db683675f3329e7d8f13c8094592b3be9/httpu/httpu.go#L99 to handle contexts that are canceled without hitting the deadline?

Typically you have a goroutine waiting on the Done() channel and then setting the conn's Deadline in the past to abort their reads.

JulianKnodt:
Updated to have an additional goroutine blocked on `<-ctx.Done()`

bradfitz:
https://github.com/tailscale/goupnp/commit/45e85838ea14cf3d11ac1a08550272c849562770 isn't right so maybe we can do code review on that repo too?

JulianKnodt:
sure, I've just been pushing to it directly since I'm the only one using it currently

JulianKnodt:
Merged and updated the other repo, this should be good then?

bradfitz:
Let me know when the test failures are fixed.

net/portmapper/upnp.go
```diff
@@ -0,0 +1,261 @@
...
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/tailscale/goupnp/dcps/internetgateway2"
+	"golang.org/x/sync/errgroup"
+	"inet.af/netaddr"
+)
+
+// DisableUPnP indicates whether to attempt UPnP mapping.
```
bradfitz, line 20: does this need to be exported?

net/portmapper/upnp.go
```diff
@@ -0,0 +1,261 @@
...
+		// greater than 0. From the spec, it appears if it is set to 0, it will switch to using
+		// 604800 seconds, but not sure why this is desired. The recommended time is 3600 seconds.
+		leaseDurationSec uint32,
+	) (err error)
+
+	DeletePortMapping(ctx context.Context, remoteHost string, externalPort uint16, protocol string) error
+	GetExternalIPAddress(ctx context.Context) (externalIPAddress string, err error)
+}
+
+const tsPortMappingDesc = "tailscale-portmap"
```
bradfitz, line 89: document what this is? this gets sent to routers as the name of the mapping?

net/portmapper/portmapper.go
```diff
@@ -66,15 +65,36 @@ type Client struct {
...
+//
+// After a mapping is created, it should be immutable, and thus reads should be safe across
+// concurrent goroutines.
+type mapping interface {
+	// isCurrent reports whether a mapping is valid now, since most mappings have a lease.
+	// isCurrent() bool
+	// release will attempt to unmap the established port mapping. Will block until completion,
+	// but can be called asynchronously. release should be idempotent, and thus even if called
+	// multiple times should not cause additional side-effects.
+	release(context.Context)
```
bradfitz, line 84: why's this one unexported but the rest are capitalized?

Normally you make one thing unexported as a way to prevent people from outside the package from implementing it, like:

https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/testing/testing.go;l=629

(see that `private()` method)

But that's not what's happening here. Seems like this should be exported?

net/portmapper/portmapper.go
```diff
@@ -66,15 +65,36 @@ type Client struct {
...
+	mapping mapping // non-nil if we have a mapping
+}
+
+// mapping represents a created port-mapping over some protocol.  It specifies a lease duration,
+// how to release the mapping, and whether the map is still valid.
+//
+// After a mapping is created, it should be immutable, and thus reads should be safe across
+// concurrent goroutines.
+type mapping interface {
+	// isCurrent reports whether a mapping is valid now, since most mappings have a lease.
```
bradfitz, line 79: delete? isCurrent is old?

net/portmapper/portmapper.go
```diff
@@ -66,15 +65,36 @@ type Client struct {
...
+	// isCurrent() bool
+	// release will attempt to unmap the established port mapping. Will block until completion,
```
bradfitz, line 81: Complete sentences: "It blocks until ..."

net/portmapper/upnp.go
```diff
@@ -0,0 +1,261 @@
...
+
+// getUPnPPortMapping attempts to create a port-mapping over the UPnP protocol. On success,
+// it will return the externally exposed IP and port. Otherwise, it will return a zeroed IP and
+// port and an error.
+func (c *Client) getUPnPPortMapping(
+	ctx context.Context,
+	gw netaddr.IP,
+	internal netaddr.IPPort,
+	prevPort uint16,
+) (external netaddr.IPPort, err error) {
```
bradfitz, line 202: maybe return an `ok bool` instead of err if all your errors are going to be `NoMappingError{ErrNoPortMappingServices}` ?  That error would look better at the call site probably anyway.

net/portmapper/upnp.go
```diff
@@ -0,0 +1,261 @@
...
+// now.
+// Adapted from https://github.com/huin/goupnp/blob/master/GUIDE.md.
+func getUPnPClient(ctx context.Context, gw netaddr.IP) (client upnpClient, err error) {
+	ctx, cancel := context.WithTimeout(ctx, 250*time.Millisecond)
+	defer cancel()
+	tasks, _ := errgroup.WithContext(ctx)
+	// Attempt to connect over the multiple available connection types concurrently,
+	// returning the fastest.
+
+	// XXX: this url seems super brittle? maybe discovery is better but this is faster
```
bradfitz, line 139: no XXX submitted. `TODO(you): ....`

net/portmapper/upnp.go
```diff
@@ -0,0 +1,261 @@
...
+	u, err := url.Parse(fmt.Sprintf("http://%s:5000/rootDesc.xml", gw))
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO need to add a context to each of these function calls
```
bradfitz, line 145: already done it looks like?

net/portmapper/upnp.go
```diff
@@ -0,0 +1,261 @@
...
+	})
+	var ppp1Clients []*internetgateway2.WANPPPConnection1
+	tasks.Go(func() error {
+		var err error
+		ppp1Clients, err = internetgateway2.NewWANPPPConnection1ClientsByURL(ctx, u)
+		errs[2] = err
+		return nil
+	})
+
+	err = tasks.Wait()
```
bradfitz, line 169: do you want to wait for all of them, or just the first successful one?

net/portmapper/upnp.go
```diff
@@ -0,0 +1,261 @@
...
+	return internalPort, err
+}
+
+// getUPnPClients gets a client for interfacing with UPnP, ignoring the underlying protocol for
+// now.
+// Adapted from https://github.com/huin/goupnp/blob/master/GUIDE.md.
+func getUPnPClient(ctx context.Context, gw netaddr.IP) (client upnpClient, err error) {
+	ctx, cancel := context.WithTimeout(ctx, 250*time.Millisecond)
+	defer cancel()
+	tasks, _ := errgroup.WithContext(ctx)
```
bradfitz, line 135: you're throwing away the derived context, but I guess you're always returning nil.

But at this point I'm not seeing what errgroup gives you. I'd need to understand the concurrency goal first.

bradfitz (commented):
Lookin' close.

net/portmapper/upnp.go
JulianKnodt: I was thinking that we want to be able to shut it down from Control if UPnP explodes. Currently it's not used elsewhere though

net/portmapper/portmapper.go
JulianKnodt: to be honest, was super lazy here and just made the others exported so it wouldn't conflict with field names. I can export it though

net/portmapper/portmapper.go
JulianKnodt: isCurrent is old

net/portmapper/upnp.go
JulianKnodt: At this point, there's not too much difference since they respect the context, but I can switch to just the first one

net/portmapper/upnp.go
JulianKnodt: I think just getting the first result is the best, so I've removed errgroup

bradfitz:
Squash back down to one commit and clean up the commit message. It doesn't need to reference past work.

And reference issue #682. (and `Updates #2109` too?)

I'm concerned about the loss of probing, though. This means stats will start to get skewed if we release 1.12.0 without adding it back, which is a real concern.

Can we add the probing back in this change, even if it's just the old code?

net/portmapper/upnp.go
bradfitz: let's unexported for now.

we should definitely have a safety knob to let us disable this later but we should make it match.

see tailcfg.go's type Debug and:

```go
                        setControlAtomic(&controlUseDERPRoute, resp.Debug.DERPRoute)
                        setControlAtomic(&controlTrimWGConfig, resp.Debug.TrimWGConfig)
```

control/controlclient/direct.go
```diff
@@ -38,6 +38,7 @@ import (
 	"tailscale.com/net/dnsfallback"
 	"tailscale.com/net/interfaces"
 	"tailscale.com/net/netns"
+	"tailscale.com/net/portmapper"
```
bradfitz, line 41: This is quite the dependency edge. Now the controlclient depends on upnp libraries, for instance.

What about a new small leaf package for these bools?

Then both controlclient and portmapper can depend on that third package?

net/portmapper/portmapper.go
```diff
@@ -344,7 +366,12 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
 	}
 	if c.lastProbe.After(now.Add(-5*time.Second)) && !haveRecentPMP {
 		c.mu.Unlock()
-		return netaddr.IPPort{}, NoMappingError{ErrNoPortMappingServices}
+		// fallback to UPnP portmapping
+		if mapping, ok := c.getUPnPPortMapping(ctx, gw, internalAddr, prevPort); ok {
+			return mapping, nil
+		} else {
```
bradfitz, line 372: drop the else and outdent on next lines. and line 415.

net/portmapper/portmapper.go
```diff
@@ -547,16 +590,11 @@ func (c *Client) Probe(ctx context.Context) (res ProbeResult, err error) {
 	} else {
 		uc.WriteTo(pcpAnnounceRequest(myIP), pcpAddr)
 	}
-	if c.sawUPnPRecently() {
```
bradfitz, line 550: this "saw UPnP recently" thing was a nice optimization we're losing now, though?

JulianKnodt: hm I guess now that I've updated the probe it's fine to add this back, I ripped it out because the old prober didn't work

control/controlclient/direct.go
```diff
@@ -41,6 +41,7 @@ import (
 	"tailscale.com/net/tlsdial"
 	"tailscale.com/net/tshttpproxy"
 	"tailscale.com/tailcfg"
+	"tailscale.com/tailcfg/controlopts"
```
bradfitz, line 44: under `tailcfg/` is a bit weird.

maybe `tailscale.com/control/controlknob` ?

control/controlknobs/controlopts.go
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
```
bradfitz, line 1: 2021

bradfitz: rename file to controlknobs.go

control/controlknobs/controlopts.go
```diff
@@ -0,0 +1,39 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package controlknobs
```
bradfitz, line 5: needs package doc.

https://github.com/golang/go/wiki/CodeReviewComments#package-comments

control/controlclient/direct.go
```diff
@@ -806,6 +807,7 @@ func (c *Direct) sendMapRequest(ctx context.Context, maxPolls int, cb func(*netm
 			}
 			setControlAtomic(&controlUseDERPRoute, resp.Debug.DERPRoute)
 			setControlAtomic(&controlTrimWGConfig, resp.Debug.TrimWGConfig)
+			controlknobs.SetDisableUPnP(resp.Debug.DisableUPnP)
```
bradfitz, line 810: we should also set this false if there is no resp.Debug.

that's a bug with the existing ones too probably.

move all 3 lines before line 801 and make them be:

     controlknobs.SetDisableUPnP(resp.Debug != nil && resp.Debug.DisableUPnP)

tailcfg/tailcfg.go
```diff
@@ -1060,6 +1060,13 @@ type Debug struct {
 	// :0 to get a random local port, ignoring any configured
 	// fixed port.
 	RandomizeClientPort bool `json:",omitempty"`
+
+	/// UPnPDisabled is whether the client will attempt to perform a UPnP portmapping.
```
bradfitz, line 1064: this is backwards.

And maybe as well make it match the naming and doc wording of `DisableSubnetsIfPAC` above.

Oh, the var IS named `Disable` first. But the doc subject doesn't match. Just fix the docs.

tstest/integration/integration_test.go
```diff
@@ -47,6 +47,7 @@ var (
 var mainError atomic.Value // of error
 
 func TestMain(m *testing.M) {
+	os.Setenv("TS_DISABLE_UPNP", "true")
```
bradfitz, line 50: comment why.

bradfitz (approved):
Anyway, looks good in general otherwise after the latest comments.

Feel free to submit once addressed.

PR #2235, opened 2021-06-23
net/isoping: add isoping package

simenghe:

Slowly taking local isoping work into the oss repo.


net/isoping/isoping_test.go
```diff
@@ -0,0 +1,91 @@
...
+// Currently only need 6 digits for the testing.
+func sigDigs(x float64, digs int) float64 {
+	return math.Round(x*math.Pow10(digs)) / math.Pow10(digs)
+}
+
+// TestOnepass_stddev tests if the function receives the same answer as in
+// the C implementation of this function.
+func TestOnepass_stddev(t *testing.T) {
+	t.Parallel()
+
```
simenghe, line 24: use slice and iterate here instead for testing!

net/isoping/isoping.go
```diff
@@ -0,0 +1,203 @@
...
+
+	srv.ListenAddr = addr
+	srv.Conn, err = net.ListenUDP("udp", addr)
+	if err != nil {
+		log.Printf("%v\n", err)
+		return
+	}
+}
+
+// initVars initializes a lot of the necessary variables for the calculation
```
DentonGentry, line 148: This is a case where I think the Go code should diverge more from C. C only added named struct initializers in C99:
```
struct Isoping{.nextTxId=1, .nextRxId=0, ...
```

For maximal portability with older C, important back when this C code was originally written, you couldn't use named initializers and added functions like this instead.

Since initVars is called from only one place, I'd recommend instead putting the field initialization into the struct when it is created.

DentonGentry (commented):
I'm partway through, sending the comments so far.

net/isoping/isoping.go
```diff
@@ -0,0 +1,203 @@
...
+	srv.latTx, srv.latTxMin, srv.latTxMax = 0, 0x7fffffff, 0
+	srv.latTxCount, srv.latTxMin, srv.latTxVarSum = 0, 0, 0
+
+	srv.latRx, srv.latRxMin, srv.latRxMax = 0, 0x7fffffff, 0
+	srv.latRxCount, srv.latRxMin, srv.latRxVarSum = 0, 0, 0
+}
+
+// generateInitialPacket generates the inital packet Tx
+func (srv *Isoping) generateInitialPacket() (*bytes.Buffer, error) {
+	srv.Tx.Magic = MAGIC
```
DentonGentry, line 177: Idiomatic Go would put most of these static initializations into the struct creation as named fields.

net/isoping/isoping.go
```diff
@@ -0,0 +1,203 @@
...
+	srv.Tx.First_ack = uint32(srv.nextTxackIndex)
+
+	// Setup the Tx to be sent from either server of client
+	buf := new(bytes.Buffer)
+	return buf, binary.Write(buf, binary.BigEndian, srv.Tx)
+}
+
+// Start starts the Isoping instance, if an address is passed a client is started
+// with that address, otherwise a server will start.
+func (srv *Isoping) Start(address ...string) {
```
DentonGentry, line 196: I think this should probably be:
```
func (srv *Isoping) Start(address []string) {
```

instead of varargs.

net/isoping/isoping_test.go
DentonGentry: Go would generally call that a table-driven test, and yes that would be a good idea for this test case.

net/isoping/isoping_test.go
```diff
@@ -0,0 +1,91 @@
...
+	if answer != expected {
+		t.Errorf("got %v, expected %v", answer, expected)
+	}
+	if answer2 != expected2 {
+		t.Errorf("got %v, expected %v", answer2, expected2)
+	}
+}
+
+// TestUstimeCast tests if casting was correct
+// sanity check, probably will be removed for redundancy
```
DentonGentry, line 38: Recommend not including the "probably will be removed..." because invariably, three years from now someone will read that and wonder if they should delete it.

"Too many tests" is rarely an actual problem that a codebase has.

net/isoping/isoping_test.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// TestValidInitialPacket will send a packet via UDP, and check if it matches
+// The size and the Magic number field that needs to be equal.
+// This mocks the initial packet sent in Isoping.
+func TestValidInitialPacket(t *testing.T) {
+	t.Parallel()
+
+	client := Isoping{IsServer: false}
+	client.Start("[::]:4948")
```
DentonGentry, line 56: In a test it is probably fine to use IPv6 specifically, though it would fail on a machine which deliberately disabled IPv6.

I'd recommend using ":4948" though.

net/isoping/isoping_test.go
```diff
@@ -0,0 +1,91 @@
...
+	err = binary.Read(buffer, binary.BigEndian, &server.Rx)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if got != binary.Size(server.Rx) || server.Rx.Magic != MAGIC {
+		t.Error("received Rx is not proper")
+	}
+
+	t.Logf("Proper Packet received from %v\n", rxaddr)
```
DentonGentry, line 90: Don't need to log successes in tests. People rapidly become blind to it and ignore everything it prints.

net/isoping/isoping.go
simenghe: I will make a constructor function for this!

net/isoping/isoping.go
```diff
@@ -0,0 +1,244 @@
...
+// Incremental standard deviation calculation, without needing to know the
+// mean in advance.  See:
+// http://mathcentral.uregina.ca/QQ/database/QQ.09.02/carlos1.html
+func onePassStddev(sumsq, sum, count int64) float64 {
+	numer := (count * sumsq) - (sum * sum)
+	denom := count * (count - 1)
+	return math.Sqrt(DIV(numer, denom))
+}
+
+// UsecMonoTimeNow returns the monotonic number of milliseconds since the program started.
```
DentonGentry, line 86: Comment says milliseconds but code returns microseconds. I think the code is right and the comment is wrong.

net/isoping/isoping.go
```diff
@@ -0,0 +1,244 @@
...
+func (srv *Isoping) UsecMonoTimeNow() uint64 {
+	tn := time.Since(srv.ClockStartTime)
+	return uint64(tn.Microseconds())
+}
+
+// UsecMonoTime returns the monotonic number of milliseconds since the program started, as a uint32.
```
DentonGentry, line 92: Comment says milliseconds

net/isoping/isoping_test.go
```diff
@@ -0,0 +1,102 @@
...
+	tests := []struct {
+		name  string
+		input []int64
+		out   float64
+	}{
+
+		{
+			name:  "basic1",
+			input: []int64{12, 2, 3},
+			out:   2.309401,
```
DentonGentry, line 33: ...that is what the C version returns? Huh. That isn't the standard deviation, but ok.

net/isoping/isoping_test.go
```diff
@@ -0,0 +1,102 @@
...
+		t.Errorf("expected %v, got : %v", expected, uint32(num))
+	}
+}
+
+// TestValidInitialPacket will send a packet via UDP, and check if it matches
+// The size and the Magic number field that needs to be equal.
+// This mocks the initial packet sent in Isoping.
+func TestValidInitialPacket(t *testing.T) {
+	client := NewInstance()
+	client.Start([]string{":4948"})
```
DentonGentry, line 69: Binding to a specific port is likely to fail on a busy CI system which is running lots of tests in parallel. Most tests bind to port 0, which means it will pick a random ephemeral port and every once in a while one of them will pick port 4948.

If you add a parameter to initServer for the port number, so you can pass in 4948 for production and 0 for tests, then:
- fetch the port number it actually used with srv.Conn.Addr().(UDPAddr).Port
- pass that port number to client.Start()

Then the test will be robust to having port 4948 be unavailable.

bradfitz:
Closing for now.

PR #2239, opened 2021-06-24
net/dns/resolver: clamp EDNS size

### LINKED ISSUE #2066: dns forwarder: clamp EDNS response size
When forwarding DNS queries, if the EDNS max response size field is larger than our forwarder supports, we should modify the query to reflect our maximum response size. Smaller values are OK (or non-EDNS queries), and we shouldn't change them because the client might be unable to handle larger responses.

sailorfrag:
This change (subject to some limitations) looks for the EDNS OPT record
in queries and responses, clamping the size field to fit within our DNS
receive buffer. If the size field is smaller than the DNS receive buffer
then it is left unchanged.

I think we will eventually need to transition to fully processing the
DNS queries to handle all situations, but this should cover the most
common case.

Mostly fixes #2066

### DIFF
--- a/net/dns/resolver/forwarder.go
+++ b/net/dns/resolver/forwarder.go
@@ -88,6 +88,57 @@ func getTxID(packet []byte) txid {
 	return (txid(hash) << 32) | txid(dnsid)
 }
 
+// clampEDNSSize attempts to limit the maximum EDNS response size. This is not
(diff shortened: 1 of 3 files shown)


sailorfrag:
Note: this is intended for 1.12, so I will hold off merging until 1.10 is branched.

danderson (approved):

PR #2241, opened 2021-06-24
wgengine/filter: add a debug flag for filter logs

Xe:
This uses a debug envvar to optionally disable filter logging rate
limits by setting the environment variable
TS_DEBUG_FILTER_RATE_LIMIT_LOGS to "all", and if it matches,
the code will effectively disable the limits on the log rate by
setting the limit to 1 millisecond. This should make sure that all
filter logs will be captured.

### DIFF
--- a/wgengine/filter/filter.go
+++ b/wgengine/filter/filter.go
@@ -7,6 +7,7 @@ package filter
 
 import (
 	"fmt"
+	"os"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


danderson (approved):

bradfitz:
The "spam_my_logs" is unprecedented, though. I thought we had an existing pattern we could follow.

bradfitz:
Oh, yes:

```
types/logger/logger.go:var disableRateLimit = os.Getenv("TS_DEBUG_LOG_RATE") == "all"
```

Xe:
I changed it to "all".

PR #2242, opened 2021-06-24
derp,cmd/derper: allow server to verify clients 

JulianKnodt:
This adds a flag to the DERP server which specifies to verify clients through a local
tailscaled. It is opt-in, so should not affect existing clients, and is mainly intended for
users who want to run their own DERP servers. It assumes there is a local tailscaled running and
will attempt to hit it for peer status information.

Updates #1264


cmd/derper/derper.go
```diff
@@ -122,6 +123,8 @@ func main() {
 	letsEncrypt := tsweb.IsProd443(*addr)
 
 	s := derp.NewServer(key.Private(cfg.PrivateKey), log.Printf)
+	// TODO feature flag this? Or is it ok to be insta-stable?
```
bradfitz, line 126: it's fine. can remove the comment.

Also, add the `Updates #nnn` line to the commit message.

derp/derp_server.go
```diff
@@ -770,7 +782,16 @@ func (c *sclient) requestMeshUpdate() {
...
+		return nil
+	}
+	status, err := tailscale.Status(context.TODO())
+	if err != nil {
+		return fmt.Errorf("Failed to query local tailscaled status: %v", err)
+	}
+	if _, exists := status.Peer[clientKey]; !exists {
+		return fmt.Errorf("client not in set of peers")
+	}
 	// TODO(bradfitz): ... and at what rate.
```
bradfitz, line 795: this comment makes less sense now without the context of the now-deleted line.

```
// TODO(bradfitz): add policy for configurable bandwidth rate per client?
```

derp/derp_server.go
```diff
@@ -770,7 +782,16 @@ func (c *sclient) requestMeshUpdate() {
 }
 
 func (s *Server) verifyClient(clientKey key.Public, info *clientInfo) error {
-	// TODO(crawshaw): implement policy constraints on who can use the DERP server
+	if !s.verifyClients {
+		return nil
+	}
+	status, err := tailscale.Status(context.TODO())
+	if err != nil {
+		return fmt.Errorf("Failed to query local tailscaled status: %v", err)
```
bradfitz, line 790: failed (lowercase)

bradfitz: and %w for wrapped errors in fmt.Errorf

derp/derp_server.go
```diff
@@ -770,7 +782,16 @@ func (c *sclient) requestMeshUpdate() {
...
+	}
+	if _, exists := status.Peer[clientKey]; !exists {
+		return fmt.Errorf("client not in set of peers")
```
bradfitz, line 793: include %v clientKey

bradfitz (approved):

PR #2251, opened 2021-06-25
net/dns: resolveConfExists reading the wrong error

crawshaw:
noideadog.gif

I don't know if this is right, but I think the wrong err was being read before.

I am also unsure how to test this in its current state, but I'm working on that separately!


bradfitz:
Change looks correct, but leaving for @danderson to consider fallout.

sailorfrag:
I agree the change looks correct

PR #2252, opened 2021-06-25
net/dns: set WSL resolv.conf

### LINKED ISSUE #775: MagicDNS domains don't get registered as expected in WSL1 /etc/resolv.conf
Version: 1.1.424 (Windows)

Steps:
- Install WSL1 (Windows Subsystem for Linux, version 1. I haven't tested version 2 but I suspect it might be the same)
- Enable MagicDNS
- From the cmd.exe prompt, try `ping <node1>` (a valid nodename in MagicDNS)
- Notice that the lookup works as expected
- Enter a WSL1 console
- Try `ping <node1>`

Expected:
- The name lookup works like it did in cmd.exe

Actual:
- No such hostname

In /etc/resolv.conf, I found this:
```
# This file was automatically generated by WSL. To stop automatic generation of 
this file, add the following entry to /etc/wsl.conf:
# [network]
# generateResolvConf = false
nameserver <my router's DNS server>
nameserver 100.100.100.100
nameserver fec0:0:0:ffff::1
search <my router's DNS domain> <the first search domain set in the tailscale admin panel> ts.tailscale.com
```

The `search` list should also contain the MagicDNS domain. Adding it by hand makes the name lookup work, but slowly.

The <my router's DNS server> entry also should not be present. Commenting it out fixes the slowness. If that's not possible, another option might be to have the global tailscale.net domain respond with NS records pointing at 100.100.100.100 I guess? But that wouldn't work well for non-tailscale.net magicdns domains that we want to support eventually.

crawshaw:
Tested on WSL2, should also work on WSL1.

Fixes #775


net/dns/ini.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"regexp"
+	"strings"
+)
+
+// parseIni parses a basic .ini file, used for wsl.conf.
+func parseIni(data string) map[string]map[string]string {
+	sectionRE := regexp.MustCompile(`^\[([^]]+)\]\s*`)
+	kvRE := regexp.MustCompile(`^\s*(\w*)\s*=\s*([^#]*)\s*`)
```
bradfitz, line 15: `\w+`, no?

net/dns/manager_windows.go
```diff
@@ -296,6 +302,18 @@ func (m windowsManager) SetDNS(cfg OSConfig) error {
...
 	}()
 
+	// On initial setup of WSL, the restart caused by --shutdown is slow,
+	// so we do it out-of-line.
+	go func() {
+		if m.wslManager != nil {
+			if err := m.wslManager.SetDNS(cfg); err != nil {
+				m.logf("WSL SetDNS: %v", err) // continue
+			} else {
+				m.logf("WSL SetDNS was successful")
```
bradfitz, line 312: SetDNS: success

net/dns/direct.go
```diff
@@ -324,3 +324,55 @@ func (m directManager) Close() error {
...
+	tmpName := fmt.Sprintf("%s.%x.tmp", filename, randBytes[:])
+	defer fs.Remove(tmpName)
+
+	if err := fs.WriteFile(tmpName, data, perm); err != nil {
+		return fmt.Errorf("atomicWriteFile: %w", err)
+	}
+	return fs.Rename(tmpName, filename)
+}
+
+// pinholeFS is a high-level file system abstraction designed just for use
```
bradfitz, line 343: I don't get the name. What's pinhole?

net/dns/direct.go
```diff
@@ -324,3 +324,55 @@ func (m directManager) Close() error {
...
+	Stat(name string) (isRegular bool, err error)
+	Rename(oldName, newName string) error
+	Remove(name string) error
+	ReadFile(name string) ([]byte, error)
+	WriteFile(name string, contents []byte, perm os.FileMode) error
+}
+
+// directFS is a pinholeFS implemented directly on the OS.
+type directFS struct {
+	prefix string // file path prefix; used for testing
```
bradfitz, line 355: so in tests, it ends in a slash?

net/dns/wsl_windows.go
```diff
@@ -0,0 +1,215 @@
...
+
+const wslConf = "/etc/wsl.conf"
+const wslConfSection = `# added by tailscale
+[network]
+generateResolvConf = false
+`
+
+// setWSLConf attempts to disable generateResolvConf in each WSL2 linux.
+// If any are changed, it reports true.
+func (wm wslManager) setWSLConf() (changed bool) {
```
bradfitz, line 140: these should be pointer receivers, as pointers are used everywhere else, and it contains a map.

bradfitz (commented):
Can you clarify in the commit message whether this is WSL1, WSL2, or both?

net/dns/direct.go
crawshaw: I couldn't think of a name for this. It's an FS, but not a standard open/read/write/close Unix API. It felt like looking through a pinhole. Any better names?

net/dns/direct.go
crawshaw: no because all the passed in names are absolute. clarified in comment.

net/dns/ini.go
```diff
@@ -0,0 +1,29 @@
...
+package dns
+
+import (
+	"regexp"
+	"strings"
+)
+
+// parseIni parses a basic .ini file, used for wsl.conf.
+func parseIni(data string) map[string]map[string]string {
+	sectionRE := regexp.MustCompile(`^\[([^]]+)\]\s*`)
```
sailorfrag, line 14: I think the trailing `\s*` for both REs can be removed with no ill effects.

net/dns/ini.go
```diff
@@ -0,0 +1,29 @@
...
+	kvRE := regexp.MustCompile(`^\s*(\w+)\s*=\s*([^#]*)\s*`)
```
sailorfrag, line 15: `;` can also be used for comments.

This actually looks pretty good: https://github.com/zieckey/goini

net/dns/wsl_windows.go
```diff
@@ -0,0 +1,215 @@
...
+	for distro, m := range wm.managers {
+		if err := m.SetDNS(cfg); err != nil {
+			wm.logf("WSL(%q) SetDNS: %v", distro, err)
+		}
+	}
+	return nil
+}
+
+const wslConf = "/etc/wsl.conf"
+const wslConfSection = `# added by tailscale
```
sailorfrag, line 133: hmm, it would probably be good to delete this on uninstall, although that might be tricky if it's edited after; deleting the `generateResolvConf` line while leaving the section name is probably required

net/dns/wsl_windows.go
```diff
@@ -0,0 +1,215 @@
...
+	if ee, _ := err.(*exec.ExitError); ee != nil {
+		if ee.ExitCode() == 1 {
+			return false, os.ErrNotExist
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+func (fs wslFS) Rename(oldName, newName string) error { return fs.cmd("mv", oldName, newName).Run() }
```
sailorfrag, line 183: How much do you want to worry about malicious filenames? `"mv", "--", oldName, newName` might be a bit safer. Does it need shell quoting? I guess for now the names are always fixed strings so it's not too scary.

net/dns/ini.go
crawshaw: Hmm. Two problems, goini expects: 1. comments start at the beginning of the line, 2. wsl.conf doesn't seem to be entirely standard?

This successfully turns off WSL resolve.conf generation here:

```
[network]
generateResolvConf = false # a comment
```

(but would be parsed incorrectly by goini I believe)

This does *not* turn off WSL resolv.conf generation on my machine:

```
[network]
generateResolvConf = false ; a comment
```

I would be happy to replace this code, but I don't think goini will quite work.

net/dns/wsl_windows.go
crawshaw: It would, but it's very problematic. To touch this file, you need to call `wsl.exe --shutdown` and restart it. On my brand-new machine, that's a multiple second operation that could be disruptive. I'd like to do it zero times.

Open to ideas here, but doing this regularly doesn't seem like a great move.

net/dns/wsl_windows.go
crawshaw: Good point, I've `--`'ed all of these.

No shell quoting is needed because I'm deliberately avoiding the shell. `wsl.exe -e` directly executes a command without the shell. That's why I'm doing the slightly janky thing of using `tee` instead of a shell pipe to write a file.

All of this is defense-in-depth, we only use file names we hard code, but it seemed worth a bit of work.

net/dns/direct.go
crawshaw: Took another attempt at the name, wholeFileFS, as you can only work with whole files.

crawshaw:
PTAL

net/dns/wsl_windows.go
sailorfrag: If tagged with a special comment, maybe the `generateResolvConf` line could be deleted by the NSIS uninstaller? That should be sufficiently infrequent.

net/dns/ini.go
sailorfrag: I think ini files aren't completely standard. Alas. This domain specific parser is fine, I suppose, since all we really need is to detect what a particular boolean value in a particular section is set to.

I didn't catch that goini requires comments be at the start of the line. That's annoying.

net/dns/wsl_windows.go
crawshaw: That would be infrequent enough for me, though it's a bit of work to make that happen. Would you be fine if I file a bug and do that later?

I suspect it isn't super important because it appears the nature of the generated resolv.conf has shifted over time at MS. The resolv.conf they generate points at a stub resolver on a fixed IP. Nothing I do to my computer's DNS settings change its contents. So I suspect the current behavior of this commit, restoring the backup file, will be adequate.

(That said, we should try and restore the backup on uninstall too.)

sailorfrag (approved):

net/dns/wsl_windows.go
sailorfrag: That sounds like a plan

PR #2253, opened 2021-06-25
cmd/tailscale/netcheck,ipnlocal: add local derpmap check

JulianKnodt:
After allowing for custom DERP maps, it's convenient to be able to see their latency in
netcheck. This adds a query to the local tailscaled for the current DERPMap.

Updates #1264


ipn/ipnlocal/local.go
```diff
@@ -2752,3 +2752,10 @@ func (b *LocalBackend) PeerDialControlFunc() func(network, address string, c sys
 	}
 	return nil
 }
+
+// DERPMap returns the current DERPMap in use.
+func (b *LocalBackend) DERPMap() *tailcfg.DERPMap {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.netMap.DERPMap
```
bradfitz, line 2760: b.netMap might be nil, in which case you should return nil

ipn/ipnlocal/local.go
```diff
@@ -2752,3 +2752,10 @@ func (b *LocalBackend) PeerDialControlFunc() func(network, address string, c sys
 	}
 	return nil
 }
+
+// DERPMap returns the current DERPMap in use.
```
bradfitz, line 2756: , or nil if not connected.

cmd/tailscale/cli/netcheck.go
```diff
@@ -59,7 +60,14 @@ func runNetcheck(ctx context.Context, args []string) error {
...
 	}
 
-	dm := derpmap.Prod()
+	ctx, cancel := context.WithTimeout(ctx, 500*time.Millisecond)
+	defer cancel()
+
+	dm, err := tailscale.CurrentDERPMap(ctx)
+	if err != nil || dm == nil {
+		// TODO should this log something?
+		dm = derpmap.Prod()
```
bradfitz, line 69: I'd like to entirely remove the derpmap.Prod from the open source tree and fall back to fetching it over the network from login.tailscale.com instead.

We might need to add a new server handler for that first.

JulianKnodt:
I still think it's a good idea to have a local copy of the DERP map, because some tests require it, and requiring that there be an internet connection for tests seem like it'll cause flakiness and annoyance

net/dnsfallback/dnsfallback.go
```diff
@@ -30,7 +30,11 @@ func Lookup(ctx context.Context, host string) ([]netaddr.IP, error) {
 		ip      netaddr.IP
 	}
 
-	dm := derpmap.Prod()
+	dm, err := tailscale.ProdDERPMap(ctx, &http.Client{})
+	if err != nil {
+		// TODO I still think it's nice to be able to be able to fallback on a static thing here.
```
bradfitz, line 35: Yeah, crap, I forgot about the DNS fallback mechanism.

Depending on DNS working breaks the whole purpose of this mechanism to get people up and running when their DNS is broken.

Hmm, let me think on this. Maybe we need to keep a cache of the last known derp map in /var/lib/tailscale as a json file that we can use instead of hitting tailscaled (this code is used by tailscaled when starting, so using the `client/tailscale` package within tailscaled doesn't really work)

bradfitz:
I just pushed 10d7c2583ca5615145fc53d634d22cebd617f31b to make `net/dnsfallback` no longer use `derpmap.Prod`.

That should unblock this, but you'll need to rebase.

wgengine/magicsock/magicsock_test.go
```diff
@@ -396,7 +396,13 @@ func TestPickDERPFallback(t *testing.T) {
 	tstest.ResourceCheck(t)
 
 	c := newConn()
-	c.derpMap = derpmap.Prod()
+	httpc := &http.Client{}
+	dm, err := tailscale.ProdDERPMap(context.Background(), httpc)
```
JulianKnodt, line 400: Should I update this to use something local? Having tests rely on the internet is also annoying in a way.

bradfitz: This test doesn't really depend on the internet. It uses derpmap.Prod just as a non-zero set of things to test the random distribution over.

You can synthesize a derpmap here in the test. Or write some function which returns a derpmap with `n` regions.

JulianKnodt: ah so it doesn't matter that it's an actual DERPMap, I'll probably just create a random one then

cmd/tailscaled/debug.go
```diff
@@ -131,7 +131,10 @@ func getURL(ctx context.Context, urlStr string) error {
 }
 
 func checkDerp(ctx context.Context, derpRegion string) error {
-	dmap := derpmap.Prod()
+	dmap, err := tailscale.ProdDERPMap(ctx, &http.Client{})
```
bradfitz, line 134: I'd just inline the fetch-from-network here. Use `ipn.DefaultControlURL` instead of hard-coding the hostname again.

JulianKnodt: er, do you mean inline all the code from ProdDERPMap? Is there any particular reason to do that?

JulianKnodt: Let me know if this is sufficient, there are two near identical copies of the same query in the codebase, but if that's fine then should be good

tstest/integration/testcontrol/testcontrol.go
```diff
@@ -629,7 +627,10 @@ func (s *Server) MapResponse(req *tailcfg.MapRequest) (res *tailcfg.MapResponse,
 	}
 	derpMap := s.DERPMap
 	if derpMap == nil {
-		derpMap = prodDERPMap
+		derpMap, err = tailscale.ProdDERPMap(context.Background(), &http.Client{})
```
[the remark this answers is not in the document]
JulianKnodt: seems to just work, will remove

client/tailscale/tailscale.go
```diff
@@ -278,3 +279,20 @@ func SetDNS(ctx context.Context, name, value string) error {
...
+	if err != nil {
+		return nil, err
+	}
+	if err = json.Unmarshal(res, &derpMap); err != nil {
+		return nil, fmt.Errorf("invalid derp map json: %w", err)
+	}
+	return &derpMap, nil
+}
+
+/*
```
bradfitz, line 297: delete

cmd/tailscale/cli/netcheck.go
```diff
@@ -59,7 +63,13 @@ func runNetcheck(ctx context.Context, args []string) error {
 		fmt.Fprintln(os.Stderr, "# Warning: this JSON format is not yet considered a stable interface")
 	}
 
-	dm := derpmap.Prod()
+	dm, err := tailscale.CurrentDERPMap(ctx)
+	if err != nil || dm == nil {
+		dm, err = prodDERPMap(ctx, &http.Client{})
```
bradfitz, line 68: http.DefaultClient

cmd/tailscale/cli/netcheck.go
```diff
@@ -176,3 +186,28 @@ func portMapping(r *netcheck.Report) string {
...
+	if err != nil {
+		return nil, fmt.Errorf("create prodDERPMap request: %w", err)
+	}
+	req = req.WithContext(ctx)
+	res, err := httpc.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("fetch prodDERPMap failed: %w", err)
+	}
+	defer res.Body.Close()
+	b, err := ioutil.ReadAll(io.LimitReader(res.Body, 1<<16))
```
bradfitz, line 201: Go style: only use shifts of <<10, <<20, <<30, etc (KiB, MIB, GiB, ...).

But 64KiB is kinda small. I'd just go 1<<20 (1MiB)

cmd/tailscale/cli/netcheck.go
```diff
@@ -176,3 +186,28 @@ func portMapping(r *netcheck.Report) string {
...
+	if err != nil {
+		return nil, fmt.Errorf("fetch prodDERPMap failed: %w", err)
+	}
+	if res.StatusCode != 200 {
+		return nil, fmt.Errorf("fetch prodDERPMap: %d: %s", res.StatusCode, string(b))
```
bradfitz, line 206: " ... %v: %s", res.Status, b)

res.Status includes the reason phrase (so use %v instead of %d)

and with %s, you don't need `string([]byte)` ... %s will format a `[]byte` as a string

cmd/tailscaled/debug.go
```diff
@@ -131,7 +133,27 @@ func getURL(ctx context.Context, urlStr string) error {
 }
 
 func checkDerp(ctx context.Context, derpRegion string) error {
-	dmap := derpmap.Prod()
+	req, err := http.NewRequest("GET", ipn.DefaultControlURL+"/derpmap/default", nil)
```
bradfitz, line 136: same comments

bradfitz:
but not cmd/ts/ ... cmd/tailscale

bradfitz:
and the package isn't netcheck. that goes after the colon if anywhere

bradfitz:
cmd/tailscale: make netcheck use active DERP map, delete static copy

PR #2261, opened 2021-06-28
control/noise: implement the 2021 control protocol base transport.

danderson:


danderson:
Current state of draft: switched to Noise IK per discussion. Client and Server functions can talk to each other successfully, and I _believe_ they implement Noise IK. Still TODO:

- [ ] Add a test against a reference implementation from noise explorer, to verify that I did in fact implement Noise IK.
- [ ] Get the nettest.TestConn test to pass, I'm currently violating some of the contract obligations of net.Conn.
- [ ] More efficient buffer use/reuse in noise.Conn
- [ ] Review the implementation of noise.Conn.Write: as implemented, each Write maps to (at least) one Noise packet. This can leak some information about the data being exchanged based on the size and timing of packets. Maybe add some buffering/padding/nagle-ish delaying to paper over that? Do we care?
- [ ] Probably rename the package control/base or somesuch, rather than center the fact that it's Noise specifically.
- [ ] Implementation review, verify that it matches the design doc and that it's implemented safely.

danderson:
@josharian as ready for review as it'll be, for the Go/performance/general code bits. Most of the open questions in the PR description are about the implementation security, but obviously feel free to review those bits as well :)

josharian:
>  Probably rename the package control/base or somesuch, rather than center the fact that it's Noise specifically.

I'd vote for a layering: package noise that specifically implements vanilla noise connection, with package control/base built atop that (if necessary, it might not be).

danderson:
Effectively, our control base protocol is (currently) pure Noise IK, so this could be net/noise. The API uses tailscale-specific wording (Client/Server instead of initiator/responder, machineKey/controlKey instead of staticKey/remoteStaticKey), and embeds a very few number of policy decisions on top of plain Noise IK (handshake frames not allowed to carry payload, no prologue, max transport frame size smaller than the absolute maximum to limit buffer sizes).

My feeling is that it's not worth splitting into a "completely generic Noise IK" package and a thin wrapper that reimplements those policy decisions - especially as those policy decisions have some security implications, so making it a little tougher/scarier to alter them is beneficial IMHO.

danderson:
I think this is ready for review. @FiloSottile annoyingly I can't add you as a reviewer in github's metadata, so, ping :p 

@josharian for Go stuff.

josharian (commented):
notes from live code review

control/noise/handshake.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+// MixDH updates s.ck and s.k with the result of X25519(priv, pub).
+//
+// MixDH corresponds to MixKey(X25519(...))) in the spec. Implementing
+// it as a single function allows for strongly-typed arguments that
+// reduce the risk of error in the caller (e.g. invoking X25519 with
+// two private keys, or two public keys), and thus producing the wrong
+// calculation.
+func (s *symmetricState) MixDH(priv key.Private, pub key.Public) error {
```
FiloSottile, line 314:
```suggestion
func (s *symmetricState) MixDH(priv PrivateKey, pub PublicKey) error {
```

control/noise/handshake.go
```diff
@@ -0,0 +1,430 @@
...
+	// TODO(danderson): check that this operation is correct. The docs
+	// for X25519 say that the 2nd arg must be either Basepoint or the
+	// output of another X25519 call.
+	//
+	// I think this is correct, because pub is the result of a
+	// ScalarBaseMult on the private key, and our private key
+	// generation code clamps keys to avoid low order points. I
```
FiloSottile, line 321: `ScalarBaseMult` is identical to `X25519(..., Basepoint)`, and they both clamp. You don't need to do clamping in your own code with either.

The code is indeed correct.

control/noise/handshake.go
```diff
@@ -0,0 +1,430 @@
...
+// EncryptAndHash encrypts plaintext into ciphertext (which must be
+// the correct size to hold the encrypted plaintext) using the current
+// s.k, mixes the ciphertext into s.h, and returns the ciphertext.
+func (s *symmetricState) EncryptAndHash(ciphertext, plaintext []byte) {
+	if s.n == invalidNonce {
+		// Noise in general permits writing "ciphertext" without a
+		// key, but in IK it cannot happen.
+		panic("attempted encryption with uninitialized key")
+	}
+	if len(ciphertext) != len(plaintext)+poly1305.TagSize {
```
FiloSottile, line 350:
```suggestion
	if len(ciphertext) != len(plaintext)+chp.Overhead {
```

and elsewhere, the poly1305 package is deprecated.

control/noise/handshake.go
```diff
@@ -0,0 +1,430 @@
...
+	if _, err := conn.Write(init[:]); err != nil {
+		return nil, fmt.Errorf("writing initiation: %w", err)
+	}
+
+	// Read in the payload and look for errors/protocol violations from the server.
+	var resp responseMessage
+	if _, err := io.ReadFull(conn, resp.Header()); err != nil {
+		return nil, fmt.Errorf("reading response header: %w", err)
+	}
+	if resp.Version() != protocolVersion {
```
FiloSottile, line 103: I did not specify the control's message to have a version field because the version is not negotiated: the client offers only one, and if the server doesn't like it it will send an error.

Having it on the wire is just another thing to check.

control/noise/handshake.go
```diff
@@ -0,0 +1,430 @@
...
+// the current s.k. If decryption is successful, it mixes the
+// ciphertext into s.h.
+func (s *symmetricState) DecryptAndHash(plaintext, ciphertext []byte) error {
+	if s.n == invalidNonce {
+		// Noise in general permits "ciphertext" without a key, but in
+		// IK it cannot happen.
+		panic("attempted encryption with uninitialized key")
+	}
+	if len(ciphertext) != len(plaintext)+poly1305.TagSize {
+		panic("plaintext is wrong size for given ciphertext")
```
FiloSottile, line 372: I know this is impossible when using the message types, but I am a little worried about DecryptAndHash being later used on attacker-controlled buffers, leading to the panic. In general, when you can return an error, I learned to do that instead and save myself a Denial of Service. :)

control/noise/key.go
```diff
@@ -0,0 +1,26 @@
...
+// might use curve25519 for.
+//
+// Furthermore, the implementation in types/key does some work that is
+// unnecessary for machine keys, and results in a harder to follow
+// implementation. In particular, machine keys do not need to be
+// clamped per the curve25519 spec because they're only used with the
+// X25519 operation, and the X25519 operation defines its own clamping
+// and sanity checking logic. Thus, these keys must be used only with
+// this Noise protocol implementation, and the easiest way to ensure
+// that is a different type.
```
FiloSottile, line 20: As noted in a comment above, you don't need clamping with the other curve25519 APIs either.

types/key/key.go
```diff
@@ -28,6 +28,17 @@ func NewPrivate() Private {
...
+	// cofactor of Curve25519). This produces a scalar that doesn't
+	// have any unsafe properties when doing operations like
+	// ScalarMult.
+	//
+	// See
+	// https://web.archive.org/web/20210228105330/https://neilmadden.blog/2020/05/28/whats-the-curve25519-clamping-all-about/
+	// for a more in-depth explanation of the constraints that led to
+	// this clamping requirement.
 	p[0] &= 248
 	p[31] = (p[31] & 127) | 64
```
FiloSottile, line 43: This is correct, but also unnecessary, as ScalarMult does it for you. 32 random bytes are a fine Private value.

FiloSottile (commented):
Reviewed control/noise/handshake.go so far.

PR #2262, opened 2021-06-28
net/dns/resolver: fmt item

JulianKnodt:
This has been bothering me for a while, but everytime I run format from the root directory
it also formats this file. I didn't want to add it to my other PRs but it's annoying to have to
revert it every time.

I don't really think this needs much review if at all, so merge it whenever


danderson (approved):

sailorfrag:
Weird, I was pretty sure I had gofmt-before-commit enabled.

I should verify that `make check` properly catches that.

PR #2276, opened 2021-06-29
cmd/tailscale: add debug flag to dump derp map

### LINKED ISSUE #2249: cmd/tailscale: 'tailscale netcheck' should use current DERP map, if available
As part of #1264, the `cmd/tailscale netcheck` subcommand should be updated to prefer to use the currently-running tailscaled's active DERP map, instead of the baked-in fallback one (that's currently only used for `netcheck`). Perhaps we even remove it, and if tailscaled is not running, we ask the default control plane server for one.

JulianKnodt:
This adds a flag in tailscale debug for dumping the derp map to stdout.

Fixes #2249


cmd/tailscale/cli/debug.go
```diff
@@ -87,6 +90,16 @@ func runDebug(ctx context.Context, args []string) error {
 		os.Stdout.Write(goroutines)
 		return nil
 	}
+	if debugArgs.derpMap {
+		dm, err := prodDERPMap(ctx, http.DefaultClient)
```
bradfitz, line 94: for this command, I don't think we should silently switch between the tailscaled derp map and the server's one.

Let's only use client/tailscale.CurrentDERPMap here. On error, it can log.Fatal some message with the URL to `ipn.DefaultControlURL/derpmap/default` or something.

bradfitz (approved):

PR #2279, opened 2021-06-29
tstest/integration: add newTestNodePair

simenghe:
newTestNodePair starts two nodes but the two nodes share a statefile.
it simulates 2 nodes with 1 machine locally as described in: the unofficial co-op guide

### DIFF
--- a/tstest/integration/integration_test.go
+++ b/tstest/integration/integration_test.go
@@ -198,6 +198,64 @@ func TestTwoNodes(t *testing.T) {
 	d2.MustCleanShutdown(t)
 }
 
+// TestTwoNodesCommonState starts up two nodes with common statefile
+// simulates 2 nodes with 1 machine locally.
+// https://www.notion.so/tailscale/Unofficial-Coop-Guide-f202e5d4f12144478f7a478e50dec82b
+func TestTwoNodesCommonState(t *testing.T) {
+	t.Parallel()
+	bins := BuildTestBinaries(t)
+
+	env := newTestEnv(t, bins)
+	defer env.Close()
+
+	// Create two nodes:
+	n1, n2 := newTestNodePair(t, env)
+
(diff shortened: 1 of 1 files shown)


tstest/integration/integration_test.go
```diff
@@ -198,6 +198,64 @@ func TestTwoNodes(t *testing.T) {
 	d2.MustCleanShutdown(t)
 }
 
+// TestTwoNodesCommonState starts up two nodes with common statefile
+// simulates 2 nodes with 1 machine locally.
+// https://www.notion.so/tailscale/Unofficial-Coop-Guide-f202e5d4f12144478f7a478e50dec82b
```
bradfitz, line 203: let's not have URLs to private things in open source

tstest/integration/integration_test.go
```diff
@@ -382,6 +440,29 @@ func newTestNode(t *testing.T, env *testEnv) *testNode {
 	}
 }
 
+// newTestNodePair allocates a temp directory for each new test node.
+// both nodes share a common statefile, to simulate 2 nodes with 1 machine locally.
```
bradfitz, line 444: I don't see how this is correct. Nodes need their own state files.

tstest/integration/integration_test.go
```diff
@@ -382,6 +440,29 @@ func newTestNode(t *testing.T, env *testEnv) *testNode {
...
+// The node is not started automatically.
+func newTestNodePair(t *testing.T, env *testEnv) (n1 *testNode, n2 *testNode) {
+	commonDir := t.TempDir()
+	n1Dir := t.TempDir()
+	n2Dir := t.TempDir()
+	t.Logf("Common shared state directory : %v", commonDir)
```
bradfitz, line 450: no spaces before the colons

PR #2283, opened 2021-06-30
derp: add optional rate limiting for DERP

JulianKnodt:
Add rate limiting for DERP servers. This adds a naive hierarchical model for rate limiting by
having a server rate limiter, and a per client limiter. The per-client limiter is intended so
that a single client cannot hog all the server bandwidth, but is hopefully bursty enough to
allow for efficient use. The per-server limiter is intended to keep costs of running DERP down.

### DIFF
--- a/derp/derp_server.go
+++ b/derp/derp_server.go
@@ -47,6 +47,22 @@ import (
 )
 
 var debug, _ = strconv.ParseBool(os.Getenv("DERP_DEBUG_LOGS"))
+var rateLimit, _ = strconv.ParseBool(os.Getenv("DERP_RATE_LIMIT"))
+
+// serverTotalAllowance is the bytes allowed to be transferred across this DERP server
+// not including disco messages within allowanceDuration. To be very specific, this is the total
+// transfer limit, and not the rate.
+const serverTotalAllowanceKB = 1 << 40
+
+// clientAllowance is the total number of allowed bytes to be transferred by a single client
+// within total allowanceDuration.
+const clientAllowanceKB = 32 << 20
+
+// allowanceDuration expresses the period over which allowance will be consumed.
+// For now allowances are over the period of a month, or 30 days worth of time.
+// The rate for the server is serverTotalAllowanceKB kilobytes every allowanceDuration seconds,
+// and the rate for the client is clientAllowanceKB kilobytes every allowanceDuration seconds.
+const allowanceDuration = time.Hour * 24 * 30
 
 // verboseDropKeys is the set of destination public keys that should
 // verbosely log whenever DERP drops a packet.
@@ -133,6 +149,10 @@ type Server struct {
 	// known peer in the network, as specified by a running tailscaled's client's local api.
 	verifyClients bool
 
+	// limiter limits data packets sent through this DERP server. If it is nil,
+	// don't rate limit.
+	limiter *rate.Limiter
+
 	mu       sync.Mutex
 	closed   bool
 	netConns map[Conn]chan struct{} // chan is closed when conn closes
@@ -212,6 +232,9 @@ func NewServer(privateKey key.Private, logf logger.Logf) *Server {
 	}
 	s.packetsDroppedTypeDisco = s.packetsDroppedType.Get("disco")
 	s.packetsDroppedTypeOther = s.packetsDroppedType.Get("other")
+	if rateLimit {
+		s.SetRateLimit()
+	}
 	return s
 }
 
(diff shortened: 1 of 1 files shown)


derp/derp_server.go
```diff
@@ -47,6 +48,20 @@ import (
 
 var debug, _ = strconv.ParseBool(os.Getenv("DERP_DEBUG_LOGS"))
 
+// serverTotalAllowanceKB is the number of allowed KB to be transferred across this DERP server
```
bradfitz, line 51: Is this a total transfer limit or a bandwidth RATE limit? why is this in terms of KB and not in terms of bytes? Why not 1<<40?

derp/derp_server.go
```diff
@@ -47,6 +48,20 @@ import (
...
+// not including disco messages. 1 << 30 KB is about a terabyte of data.
+var serverTotalAllowanceKB = 1 << 30
+
+// clientAllowanceKB is the total number of allowed KB to be transferred by a single client
+// within total allowanceDuration. It is about 32 MB.
+var clientAllowanceKB = 1 << 15
```
bradfitz, line 57: 32 << 20

derp/derp_server.go
```diff
@@ -198,6 +216,11 @@ func NewServer(privateKey key.Private, logf logger.Logf) *Server {
 		sentTo:               map[key.Public]map[key.Public]int64{},
 		avgQueueDuration:     new(uint64),
 		keyOfAddr:            map[netaddr.IPPort]key.Public{},
+
+		// TODO 1<<25 kb burst is kind of arbitrary, pick something better?
+		limiter: rate.NewLimiter(
+			rate.Every(allowanceDuration/time.Duration(serverTotalAllowanceKB)), 1<<25,
```
bradfitz, line 222: only use shifts of 10, 20, 30, 40

derp/derp_server.go
```diff
@@ -508,6 +540,18 @@ func (s *Server) accept(nc Conn, brw *bufio.ReadWriter, remoteAddr string, connN
 	return c.run(ctx)
 }
 
+// toKB converts bytes to kilobytes, rounding up if necessary.
```
bradfitz, line 543: why are kilobytes the unit used everywhere?

derp/derp_server.go
```diff
@@ -508,6 +540,18 @@ func (s *Server) accept(nc Conn, brw *bufio.ReadWriter, remoteAddr string, connN
...
+func toKB(bytes uint32) uint32 { return (bytes + (1 << 10) - 1) >> 10 }
+
+func (c *sclient) rateLimit(kb uint32) {
+	// TODO do we want to have some timeout for things here?
+	ctx := context.TODO()
+	if c.limiter != nil {
```
bradfitz, line 549: you documented that if this is nil, rate limit isn't enabled, but you are on line 552 below, which is an added mutex that all our prod traffic would be hitting even when don't want it (we rate-limit at the OS level)

derp/derp_server.go
JulianKnodt: I was thinking that bytes would be too granular, maybe just eager optimization on my part.

derp/derp_server.go
JulianKnodt: eh just an arbitrary choice, will switch to bytes

derp/derp_server.go
JulianKnodt: Hm, there are two limiters here, one on the server and one on the clients. I had the server rate limiting always be on, but I moved it to a separate function and documented the nil behavior.

derp/derp_server.go
JulianKnodt: 1 << 40 overflows int on 32 bit architectures, which might be one argument for KB

bradfitz: Don't use an int then. :)

See:
https://blog.golang.org/constants 
https://tour.golang.org/basics/16

https://golang.org/ref/spec#Constants ...

> An untyped constant has a default type which is the type to which the constant is implicitly converted in contexts where a typed value is required, for instance, in a short variable declaration such as i := 0 where there is no explicit type. The default type of an untyped constant is bool, rune, int, float64, complex128 or string respectively, depending on whether it is a boolean, rune, integer, floating-point, complex, or string constant.

bradfitz:
What bug is this for? What's the context?

JulianKnodt:
Created tailscale/corp#2281. A new feature for DERPs, so I should probably add feature flags for it

JulianKnodt:
Closing, probably as more things unfold in the future this will get reopened & the design changed.

PR #2284, opened 2021-06-30
net/dns: avoid running wsl.exe as LocalSystem

crawshaw:
It doesn't work. It needs to run as the user.

	https://github.com/microsoft/WSL/issues/4803

The mechanism for doing this was extracted from:

	https://web.archive.org/web/20101009012531/http://blogs.msdn.com/b/winsdk/archive/2009/07/14/launching-an-interactive-process-from-windows-service-in-windows-vista-and-later.aspx

### DIFF
--- a/net/dns/wsl_windows.go
+++ b/net/dns/wsl_windows.go
@@ -9,20 +9,21 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"os/user"
 	"strings"
 	"syscall"
 	"unicode/utf16"
 
+	"golang.org/x/sys/windows"
 	"tailscale.com/types/logger"
+	"tailscale.com/util/winutil"
 )
 
 // wslDistros reports the names of the installed WSL2 linux distributions.
-func wslDistros(logf logger.Logf) []string {
-	cmd := exec.Command("wsl.exe", "-l")
-	cmd.SysProcAttr = &syscall.SysProcAttr{HideWindow: true}
-	b, err := cmd.CombinedOutput()
+func wslDistros() ([]string, error) {
+	b, err := wslCombinedOutput(exec.Command("wsl.exe", "-l"))
 	if err != nil {
-		return nil
+		return nil, fmt.Errorf("%v: %q", err, string(b))
 	}
 
 	// The first line of output is a WSL header. E.g.
@@ -42,16 +43,14 @@ func wslDistros(logf logger.Logf) []string {
 	if bytes.HasPrefix(b, []byte("W\x00i\x00n\x00d\x00o\x00w\x00s\x00")) {
 		output, err = decodeUTF16(b)
 		if err != nil {
-			logf("failed to decode wsl.exe -l output %q: %v", b, err)
-			return nil
+			return nil, fmt.Errorf("failed to decode wsl.exe -l output %q: %v", b, err)
 		}
 	} else {
 		output = string(b)
 	}
-	fmt.Printf("wslDistros: %q\n", output)
 	lines := strings.Split(output, "\n")
 	if len(lines) < 1 {
-		return nil
+		return nil, nil
 	}
 	lines = lines[1:] // drop "Windows Subsystem For Linux" header
 
@@ -62,10 +61,9 @@ func wslDistros(logf logger.Logf) []string {
 		if name == "" {
 			continue
 		}
-		fmt.Printf("wslDistros: name=%q\n", name)
 		distros = append(distros, name)
 	}
-	return distros
+	return distros, nil
 }
 
 func decodeUTF16(b []byte) (string, error) {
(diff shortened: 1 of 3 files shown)


bradfitz:
I'll review when back at computer tonight.

net/dns/wsl_windows.go
```diff
@@ -197,21 +198,75 @@ func (fs wslFS) WriteFile(name string, contents []byte, perm os.FileMode) error
...
+	cmd.Stdout = buf
+	cmd.Stderr = buf
+	err := wslRun(cmd)
+	return buf.Bytes(), err
+}
+
+func wslRun(cmd *exec.Cmd) (err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("wslRun(%v): %v", cmd.Args, err)
```
bradfitz, line 229: %w for err

net/dns/wsl_windows.go
```diff
@@ -197,21 +198,75 @@ func (fs wslFS) WriteFile(name string, contents []byte, perm os.FileMode) error
...
+	procWTSGetActiveConsoleSessionId = kernel32.NewProc("WTSGetActiveConsoleSessionId")
+)
+
+// wtsGetActiveConsoleSessionID Returns the session ID of the console session.
+//
+// https://msdn.microsoft.com/en-us/library/aa383835(VS.85).aspx
+// DWORD WTSGetActiveConsoleSessionId();
+func wtsGetActiveConsoleSessionID() (sessionID uint32, err error) {
+	r1, _, e1 := procWTSGetActiveConsoleSessionId.Call()
+	if r1 == 0xFFFFFFFF {
```
bradfitz, line 268: check e1 too? Oh, I guess API docs are pretty clear that this is enough.

net/dns/wsl_windows.go
```diff
@@ -197,21 +198,75 @@ func (fs wslFS) WriteFile(name string, contents []byte, perm os.FileMode) error
...
+var (
+	kernel32                         = syscall.NewLazyDLL("kernel32.dll")
+	procWTSGetActiveConsoleSessionId = kernel32.NewProc("WTSGetActiveConsoleSessionId")
+)
+
+// wtsGetActiveConsoleSessionID Returns the session ID of the console session.
+//
+// https://msdn.microsoft.com/en-us/library/aa383835(VS.85).aspx
+// DWORD WTSGetActiveConsoleSessionId();
+func wtsGetActiveConsoleSessionID() (sessionID uint32, err error) {
```
bradfitz, line 266: move this to util/winutil? could even do the //sys thing to auto-generate the Go wrapper. (e.g. see https://github.com/tailscale/corp/commit/8a4f261c489d842049f8d8a9cdd3d1571cb0c428 for using the syscall generator)

Or really, we can add it to x/sys/windows tonight too.

net/dns/wsl_windows.go
```diff
@@ -197,21 +198,75 @@ func (fs wslFS) WriteFile(name string, contents []byte, perm os.FileMode) error
...
+	}
+	return cmd.Run()
+}
+
+var (
+	kernel32                         = syscall.NewLazyDLL("kernel32.dll")
+	procWTSGetActiveConsoleSessionId = kernel32.NewProc("WTSGetActiveConsoleSessionId")
+)
+
+// wtsGetActiveConsoleSessionID Returns the session ID of the console session.
```
bradfitz, line 262: lowercase "returns"

net/dns/wsl_windows.go
```diff
@@ -9,19 +9,20 @@ import (
...
+	"golang.org/x/sys/windows"
 	"tailscale.com/types/logger"
 )
 
 // wslDistros reports the names of the installed WSL2 linux distributions.
 func wslDistros(logf logger.Logf) []string {
-	cmd := exec.Command("wsl.exe", "-l")
-	cmd.SysProcAttr = &syscall.SysProcAttr{HideWindow: true}
-	b, err := cmd.CombinedOutput()
+	b, err := wslCombinedOutput(exec.Command("wsl.exe", "-l"))
```
bradfitz, line 23: I don't think this is going to work in practice: the machine first boots up and nobody's logged in yet, so WTSGetActiveConsoleSessionId will return an error, and then dns.NewOSConfigurator will never run again.

net/dns/wsl_windows.go
crawshaw: I tried putting this in x/sys/windows using the `//sys` wrapper but it generates the wrong code:

```
func WTSGetActiveConsoleSessionId() (sessionID uint32, err error) {
       r0, _, e1 := syscall.Syscall(procWTSGetActiveConsoleSessionId.Addr(), 0, 0, 0, 0)
       sessionID = uint32(r0)
       if sessionID == 0 {
               err = errnoErr(e1)
       }
       return
}
```

The documentation linked above says 'If there is no session attached to the physical console, (for example, if the physical console session is in the process of being attached or detached), this function returns 0xFFFFFFFF.'

So I guess I'll move it to winutil.

net/dns/wsl_windows.go
crawshaw: The Proc.Call method is bonkers:

_The returned error is always non-nil, constructed from the result of GetLastError. Callers must inspect the primary return value to decide whether an error occurred (according to the semantics of the specific function being called) before consulting the error._

https://pkg.go.dev/golang.org/x/sys/windows#Proc.Call

net/dns/wsl_windows.go
crawshaw: I don't know if anyone cares about x/sys having the correct semantics baked in or not, so I spent an hour figuring out how to get gerrit out of mothballs and sent https://go-review.googlesource.com/c/sys/+/331909, figuring I can do the 0xffffff check myself.

net/dns/wsl_windows.go
```diff
@@ -197,21 +203,58 @@ func (fs wslFS) WriteFile(name string, contents []byte, perm os.FileMode) error
...
+		if err != nil {
+			err = fmt.Errorf("wslRun(%v): %w", cmd.Args, err)
+		}
+	}()
+
+	var token syscall.Token
+	if u, err := user.Current(); err == nil && u.Name == "SYSTEM" {
+		// We need to switch user to run wsl.exe.
+		// https://github.com/microsoft/WSL/issues/4803
+		sessionID, err := winutil.WTSGetActiveConsoleSessionID()
```
sailorfrag, line 242: So I think this works fine when the tray icon is running because that implies an active session. Does it still function properly with unattended mode? I suppose it doesn't necessarily need to run at startup with unattended mode because it should already have configured everything during the initial login to configure unattended mode.

sailorfrag (approved):

net/dns/wsl_windows.go
crawshaw: I don't expect it to do anything particularly useful in unattended mode. Similarly if your win10 is configured to not auto-login, it doesn't do anything useful on boot.

I did test that on reboot, if your win10 drops you at a "type a PIN" screen, this can successfully configure wsl.exe on boot.

PR #2297, opened 2021-07-01
net/tstun: de-select Wrapper

josharian:
Every TUN Read went through several multi-case selects.
We know from past experience with wireguard-go that these are slow
and cause scheduler churn.

This PR removes multi-case selects from hot code.

There are two commits; the first is trivial.

This is an intricate and dangerous change.
I would be happy to co-review it live,
if that is helpful.

### DIFF
--- a/net/tstun/wrap.go
+++ b/net/tstun/wrap.go
@@ -57,6 +57,13 @@ var parsedPacketPool = sync.Pool{New: func() interface{} { return new(packet.Par
 type FilterFunc func(*packet.Parsed, *Wrapper) filter.Response
 
 // Wrapper augments a tun.Device with packet filtering and injection.
+//
+// Wrapper is on the hot path for packets flowing through Tailscale:
+// Wrapper.Read and Wrapper.Write execute for every single packet.
+// Wrapper is thus written with an eye towards performance.
+// In particular, it could be made simpler and cleaner by using
+// multi-channel selects to handle shutdown, but those are
+// unfortuately not suitable for performance reasons.
 type Wrapper struct {
 	logf logger.Logf
 	// tdev is the underlying Wrapper device.
@@ -71,13 +78,22 @@ type Wrapper struct {
 	// buffer stores the oldest unconsumed packet from tdev.
 	// It is made a static buffer in order to avoid allocations.
 	buffer [maxBufferSize]byte
-	// bufferConsumed synchronizes access to buffer (shared by Read and poll).
-	bufferConsumed chan struct{}
-
-	// closed signals poll (by closing) when the device is closed.
-	closed chan struct{}
-	// errors is the error queue populated by poll.
-	errors chan error
+	// bufferC coordinates access to buffer between Read and poll.
+	// Read and poll take turns using buffer: poll fills it and Read empties it.
+	// A nil buffer passed over bufferC tells poll to exit.
+	//
+	// bufferC must be buffered. See the comment in Wrap.
+	//
+	// bufferC is a chan []byte instead of a chan struct{} for two reasons.
+	// (1) []byte can be nil, which provides a useful sentinel value.
+	// (2) It is a step towards slightly decoupling Read and poll,
+	// thereby allowing us to use multiple buffers.
+	// Letting poll use multiple buffers will let us issue syscalls sooner,
+	// which is important to performance, as syscalls are the
+	// single slowest thing that Tailscale does, by a large margin.
+	bufferC chan []byte
+	// done signals (by closing) when the device is closed.
+	done chan struct{}
 	// outbound is the queue by which packets leave the TUN device.
 	//
 	// The directions are relative to the network, not the device:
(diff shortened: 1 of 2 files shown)


net/tstun/wrap.go
```diff
@@ -71,13 +71,12 @@ type Wrapper struct {
...
 	// It is made a static buffer in order to avoid allocations.
 	buffer [maxBufferSize]byte
-	// bufferConsumed synchronizes access to buffer (shared by Read and poll).
-	bufferConsumed chan struct{}
-
-	// closed signals poll (by closing) when the device is closed.
-	closed chan struct{}
-	// errors is the error queue populated by poll.
-	errors chan error
+	// bufferC coordinates re-usable buffers between Read and poll.
```
danderson, line 74: "between" suggests to me that bidirectional communication happens. Where does the buffer go after `poll` fills it? How does it get there?

Perhaps rephrase "between Read and poll" to "from Read to poll". If the return path is also interesting and relevant to how this channel functions, also explain it here.

danderson: I think I would like the docs to specify that this is a buffered channel, because it being buffered is crucial to the semantics. But I'm unsure of myself, because we don't point it out explicitly anywhere else. Your thoughts welcome.

net/tstun/wrap.go
```diff
@@ -235,40 +260,29 @@ func (t *Wrapper) Name() (string, error) {
...
-		}
-
+		// In principle, read errors are not fatal (but wireguard-go disagrees).
+		n, err := t.tdev.Read(buf, PacketStartOffset)
 		// Wireguard will skip an empty read,
 		// so we might as well do it here to avoid the send through t.outbound.
-		if n == 0 {
-			t.bufferConsumed <- struct{}{}
+		if n == 0 && err == nil {
+			t.bufferC <- buf
```
danderson, line 280: Confused by how this works. poll both reads from and writes to this channel. Is it just talking to itself? Why doesn't it just hold onto the buf?

net/tstun/wrap.go
```diff
@@ -235,40 +260,29 @@ func (t *Wrapper) Name() (string, error) {
...
 			continue
 		}
-
-		select {
-		case <-t.closed:
-			return
-		case t.outbound <- t.buffer[PacketStartOffset : PacketStartOffset+n]:
-			// continue
+		t.outbound <- tunReadResult{
+			data: buf[PacketStartOffset : PacketStartOffset+n],
```
danderson, line 284: Okay, in this instance, buf doesn't get returned, so presumably something else on the far side of outbound is returning it for reuse. It wasn't clear to me at all from the struct docs what this buffer does, I think that docstring needs a bit more attention.

net/tstun/wrap.go
```diff
@@ -325,26 +339,17 @@ func (t *Wrapper) IdleDuration() time.Duration {
...
-		// so this is the easiest way to tell when it has been consumed.
-		// &pkt[0] can be used because empty packets do not reach t.outbound.
-		if &pkt[0] == &t.buffer[PacketStartOffset] {
-			t.bufferConsumed <- struct{}{}
-		} else {
-			// If the packet is not from t.buffer, then it is an injected packet.
-			wasInjectedPacket = true
-		}
+	}
+	res := <-t.outbound
```
danderson, line 345: Ah, okay. poll and Read share one packet buffer, which circulates t.outbound -> Read -> t.bufferC -> poll -> (loop).

Mildly strange to me that the buffer in question is at t.buffer, where both routines can access it, and yet we're passing it around via this channel, rather than use the channel as a simple struct{} signalling chan. I suppose to signal close, we need it to be at least a bistate value, so chan bool where true means "you're free to use t.buffer again" and false means "closed". Am I missing something?

I think the way it's done here is also okay, but it feels like passing []byte around when it's actually just t.buffer being shared obscures the flow a little bit.

danderson (approved):
Seems good. Sprinkled a few thoughts, and have one extra overall thought: there should be a comment somewhere explaining that this is all deliberate and what the performance goal was, so that future-me doesn't look at this and go "oh I could clean up that code by using a multi-channel select!"

Let's review together one more time to make sure, but I believe what you've implemented makes sense.

net/tstun/wrap.go
josharian: Yes, it is talking to itself. I've added a new commit changing it to an inner loop. (Will push soon.) Do you think that that is clearer?

net/tstun/wrap.go
```diff
@@ -71,13 +78,22 @@ type Wrapper struct {
...
+	// bufferC is a chan []byte instead of a chan struct{} for two reasons.
+	// (1) []byte can be nil, which provides a useful sentinel value.
+	// (2) It is a step towards slightly decoupling Read and poll,
+	// thereby allowing us to use multiple buffers.
+	// Letting poll use multiple buffers will let us issue syscalls sooner,
+	// which is important to performance, as syscalls are the
+	// single slowest thing that Tailscale does, by a large margin.
+	bufferC chan []byte
+	// done signals (by closing) when the device is closed.
+	done chan struct{}
```
bradfitz, line 96: <-chan struct{} ?  or if it's writable, what are the rules on who can close/when?

net/tstun/wrap.go
```diff
@@ -160,14 +188,28 @@ func (t *Wrapper) SetDestIPActivityFuncs(m map[netaddr.IP]func()) {
...
 	var err error
 	t.closeOnce.Do(func() {
-		// Other channels need not be closed: poll will exit gracefully after this.
-		close(t.closed)
+		close(t.done)
+		// Each channel is buffered enough to guarantee that sends will not block.
+		// Signal poll to stop.
+		t.bufferC <- nil
+		// Signal Read to stop.
+		t.outbound <- tunReadResult{err: io.EOF}
```
bradfitz, line 196: can this block forever if there's no reader? say here why not. :)

josharian:
Abandoning this in favor of https://github.com/tailscale/tailscale/pull/2306

PR #2298, opened 2021-07-01
net/netns: support !CAP_NET_ADMIN

DentonGentry:
Support operation in container environments which lack CAP_NET_ADMIN, like AWS App Runner, AWS Lightsail, or Azure App Service.

This PR also makes netns handle localhost, which means tests no longer need special handling. Tests can make connections to localhost and have them work.


net/netns/netns_linux.go
```diff
@@ -26,19 +26,49 @@ import (
...
+	if err != nil || sockErr != nil {
+		return false
+	}
+
+	return true
+}
+
+// useSocketMark reports whether SO_MARK works.
 // If it doesn't, we have to use SO_BINDTODEVICE on our sockets instead.
-func ipRuleAvailable() bool {
```
danderson, line 37: Why replace `ipRuleAvailable` ? If it's not available, regardless of whether we can set socket marks, marked packets won't route correctly.

DentonGentry: Because it is not sufficient: Azure and AWS App Runner run containers which do not have CAP_NET_ADMIN. They *do* have `ip rule` and can print the table of existing rules, but attempting to use SO_MARK fails with EPERM. So on those platforms we decide to use SO_MARK, which fails, and then Tailscale cannot make any data connections.

If we try to use SO_MARK as proposed here, then we detect that SO_MARK doesn't work and we use SO_BINDTODEVICE instead. SO_BINDTODEVICE does work, even without CAP_NET_ADMIN.

DentonGentry:
This PR unbreaks netns on the GitHub CI and on people's desktops when running as non-root.

A number of the tests turn out to only pass with broken netns.

net/netns/netns_linux.go
danderson: Sorry, I meant: SO_MARK by itself is not sufficient, so the check needs to be additive. If `ip rule` isn't there, we can't add the plumbing that makes the marks work in the first place. Checking both SGTM

danderson (approved):

danderson:
Do the tests that fail with broken netns do so deliberately? Or needs investigating?

DentonGentry:
Unfortunately the test failures are real: netns now works in the GitHub CI runner, by using SO_BINDTODEVICE, which breaks a few of the tests. I'm working on fixing them.

DentonGentry:
There has been a significant change since the review: addition of a TS_BIND_IFC environment variable, needed to let the integration test work. Otherwise, netns now figures out to bind to the default route interface like eth0, but the integration test needs to connect via lo.

bradfitz:
I'll try to review tonight when kids permit. Your last comment scares me a bit: integration tests shouldn't need hand holding. Then they become less of integration tests.

hostinfo/hostinfo.go
```diff
@@ -41,6 +44,9 @@ func GetEnvType() EnvType {
 }
 
 func getEnvType() EnvType {
+	if inTestCase() {
```
bradfitz, line 47: probably worth a comment that this should be first in the list. Otherwise it'd be easy to see this failing when running Go binary tests under, say, knative, which is something I've done before.

net/netns/netns_linux.go
```diff
@@ -26,19 +28,50 @@ import (
...
-// ipRuleOnce is the sync.Once & cached value for ipRuleAvailable.
-var ipRuleOnce struct {
+// socketMarkWorksOnce is the sync.Once & cached value for useSocketMark.
+var socketMarkWorksOnce struct {
 	sync.Once
 	v bool
 }
 
-// ipRuleAvailable reports whether the 'ip rule' command works.
+// socketMarkWorks returns whether SO_MARK works. If unsure, it returns true as the
```
bradfitz, line 37: can you add `// unsure; assume yes` comments on the unsure exit paths?

tstest/integration/integration_test.go
```diff
@@ -493,6 +493,7 @@ func (n *testNode) StartDaemon(t testing.TB) *Daemon {
 		"TS_LOG_TARGET="+n.env.LogCatcherServer.URL,
 		"HTTP_PROXY="+n.env.TrafficTrapServer.URL,
 		"HTTPS_PROXY="+n.env.TrafficTrapServer.URL,
+		"TS_NETNS_BIND_IFC=lo",
```
bradfitz, line 496: I really dislike this sort of thing. I don't want to be holding the integration test's hand. I fear this will bite us in both directions in the future: break things it shouldn't (binding to lo when we don't want that), and hide broken routing in tailscaled by us telling it what to do.

The netns stuff is used by:

* magicsock UDP conns
* DERP TCP dials
* control TCP dials

Which one(s) are the problem?

If it's just magicsock, can we instead have magicsock use a different socket for localhost peers? We've been wanting to do that for a long time anyway (go multi-socket) and this would be a nice baby step into that.

net/netns/netns_linux.go
```diff
@@ -98,6 +136,11 @@ func bindToDevice(fd uintptr) error {
 		// a default route anyway, it doesn't matter.
 		ifc = "lo"
 	}
+	bind_specific_ifc := os.Getenv("TS_NETNS_BIND_IFC")
```
bradfitz, line 139: underscores: https://golang.org/doc/effective_go#mixed-caps

DentonGentry:
For tstest/integration we need localhost connections for control and magicsock UDP probing to work. I've removed TS_NETNS_BIND_IFC and started on getting those to work, going to convert PR to a draft for a while.

hostinfo/hostinfo.go
DentonGentry: Done

net/netns/netns_linux.go
DentonGentry: Done.

net/netns/netns_linux.go
DentonGentry: Removed TS_NETNS_BIND_IFC altogether.

control/controlclient/direct.go
```diff
@@ -154,13 +155,18 @@ func NewDirect(opts Options) (*Direct, error) {
 			UseLastGood:      true,
 			LookupIPFallback: dnsfallback.Lookup,
 		}
-		dialer := netns.NewDialer()
+		var context func(context.Context, string, string) (net.Conn, error)
```
bradfitz, line 158: I'd move all this stuff to netns itself, like we already have for other platforms:

```
bradfitz@tsdev:~/src/tailscale.com/net/netns$ git grep -i localhost
netns_darwin_tailscaled.go:             // Don't bind to an interface for localhost connections.
netns_windows.go:               // Don't bind to an interface for localhost connections,
```

Then this file is unchanged.

net/netns/netns.go
```diff
@@ -66,3 +67,21 @@ type Dialer interface {
...
+		var err error
+		host, _, err = net.SplitHostPort(host)
+		if err != nil {
+			return false
+		}
+	}
+	if host == "localhost" || host == "localhost6" {
+		return true
+	}
+	ip := net.ParseIP(host)
```
bradfitz, line 82: netaddr.ParseIP doesn't allocate.

DentonGentry: Cool. Done.

control/controlclient/direct.go
DentonGentry: Done.

net/netns/netns.go
```diff
@@ -66,3 +68,20 @@ type Dialer interface {
 	Dial(network, address string) (net.Conn, error)
 	DialContext(ctx context.Context, network, address string) (net.Conn, error)
 }
+
+func IsLocalhost(host string) bool {
```
bradfitz, line 72: does this need to be exported? looks like not?

If it does, though, then the bar for docs is raised and this needs proper docs.

net/netns/netns_test.go
```diff
@@ -40,3 +40,40 @@ func TestDial(t *testing.T) {
...
+		{"localhost6 with port", "localhost6:1", true},
+		{"ip6-localhost", "ip6-localhost", true},
+		{"ip6-localhost with port", "ip6-localhost:1", true},
+		{"ip6-loopback", "ip6-loopback", true},
+		{"ip6-loopback with port", "ip6-loopback:1", true},
+	}
+
+	for _, test := range tests {
+		if got := IsLocalhost(test.host); got != test.want {
+			t.Errorf("IsLocalhost(%s) = %v, want %v", test.name, got, test.want)
```
bradfitz, line 76: minor personal preference, but I'd use %q for the string instead of %s, as it'd look like a normal call with a string literal in the output

net/netns/netns.go
```diff
@@ -66,3 +68,20 @@ type Dialer interface {
...
+	if err == nil {
+		host = name
+	} // error means the string didn't contain a port number, so ignore it.
+
+	if host == "localhost" || host == "localhost6" || host == "ip6-loopback" || host == "ip6-localhost" {
+		return true
+	}
+
+	ip, err := netaddr.ParseIP(host)
+	if err == nil && ip.IsLoopback() {
```
bradfitz, line 83: could just `return err == nil && ip.IsLoopback()`

But the method call is well defined even on the ip zero value, so you could also just write:

```go
     ip, _ := netaddr.ParseIP(host)
     return ip.IsLoopback()
```

net/netns/netns.go
```diff
@@ -66,3 +68,20 @@ type Dialer interface {
 	Dial(network, address string) (net.Conn, error)
 	DialContext(ctx context.Context, network, address string) (net.Conn, error)
 }
+
+func IsLocalhost(host string) bool {
+	name, _, err := net.SplitHostPort(host)
+	if err == nil {
+		host = name
```
bradfitz, line 75: rather than give is variable different meanings at different times, maybe call the func arg `addr string` (which matches what it's named at call sites, from net.Dial), and then have a separate `host` variable inside the func that you assign to in two places, either from the split or from addr on split failure.

net/netns/netns.go
```diff
@@ -66,3 +68,20 @@ type Dialer interface {
...
+	} // error means the string didn't contain a port number, so ignore it.
+
+	if host == "localhost" || host == "localhost6" || host == "ip6-loopback" || host == "ip6-localhost" {
```
bradfitz, line 78: I've never seen these latter three. Where are they from? (distro, spec?) Maybe a little comment?

bradfitz (commented):
Lookin' good. @danderson, any comments?

net/netns/netns.go
DentonGentry: Unexported it.

net/netns/netns_test.go
DentonGentry: Done.

net/netns/netns.go
DentonGentry: Simplified.

net/netns/netns.go
DentonGentry: Done.

net/netns/netns.go
DentonGentry: Documented.

DentonGentry:
@danderson I plan to submit this tomorrow, if you have any final comments.

vysecurity:
How do we get this latest version ?

DentonGentry:
This support is in the unstable 1.11.x builds: https://pkgs.tailscale.com/unstable/
Unstable means it has not yet gone through a full test cycle, though we are quite close to the 1.12 release so the testing is mostly done.

If you want to use the 1.11.x builds you'll need to follow the instructions on https://pkgs.tailscale.com/unstable/
to join the unstable track.

This support will also be in the 1.12.0 stable release, which will go to everyone.

PR #2306, opened 2021-07-01
net/tstun: remove multi-case selects from hot code

josharian:
Every TUN Read went through several multi-case selects.
We know from past experience with wireguard-go that these are slow
and cause scheduler churn.

This PR removes multi-case selects from hot code.

There are three commits; the first contains the substantive change.
The other two are simplifications.

### DIFF
--- a/net/tstun/wrap.go
+++ b/net/tstun/wrap.go
@@ -354,15 +354,13 @@ func (t *Wrapper) Read(buf []byte, offset int) (int, error) {
 	defer allowSendOnClosedChannel() // for send to t.bufferConsumed
 	pkt := res.data
 	n := copy(buf[offset:], pkt)
-	wasInjectedPacket := false
-	// t.buffer has a fixed location in memory,
-	// so this is the easiest way to tell when it has been consumed.
+	// t.buffer has a fixed location in memory.
+	// If the packet is not from t.buffer, then it is an injected packet.
 	// &pkt[0] can be used because empty packets do not reach t.outbound.
-	if &pkt[0] == &t.buffer[PacketStartOffset] {
+	isInjectedPacket := &pkt[0] != &t.buffer[PacketStartOffset]
+	if !isInjectedPacket {
+		// We are done with t.buffer. Let poll re-use it.
 		t.bufferConsumed <- struct{}{}
-	} else {
-		// If the packet is not from t.buffer, then it is an injected packet.
-		wasInjectedPacket = true
 	}
 
 	p := parsedPacketPool.Get().(*packet.Parsed)
@@ -375,13 +373,8 @@ func (t *Wrapper) Read(buf []byte, offset int) (int, error) {
 		}
 	}
 
-	// For injected packets, we return early to bypass filtering.
-	if wasInjectedPacket {
-		t.noteActivity()
-		return n, nil
-	}
-
-	if !t.disableFilter {
+	// Do not filter injected packets.
+	if !isInjectedPacket && !t.disableFilter {
 		response := t.filterOut(p)
 		if response != filter.Accept {
 			// Wireguard considers read errors fatal; pretend nothing was read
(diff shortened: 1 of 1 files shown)


josharian:
Many thanks to @danderson for reviewing my initial mess, @bradfitz for pushing back on that mess, and @crawshaw for suggesting a better path forward.

net/tstun/wrap.go
```diff
@@ -230,46 +243,47 @@ func (t *Wrapper) Name() (string, error) {
 	return t.tdev.Name()
 }
 
+// allowSendOnClosedChannel suppressed panics due to sending on a closed channel.
```
bradfitz, line 246: suppresses

net/tstun/wrap.go
```diff
@@ -230,46 +243,47 @@ func (t *Wrapper) Name() (string, error) {
...
-		}
-
-		select {
-		case <-t.closed:
-			return
-		case t.outbound <- t.buffer[PacketStartOffset : PacketStartOffset+n]:
-			// continue
+			n, err = t.tdev.Read(t.buffer[:], PacketStartOffset)
 		}
+		t.outbound <- tunReadResult{data: t.buffer[PacketStartOffset : PacketStartOffset+n], err: err}
```
bradfitz, line 286: can you annotate which sends on closed channels (like here) are okay, to cross-reference the horror more?

net/tstun/wrap.go
```diff
@@ -88,7 +85,7 @@ type Wrapper struct {
 	//
 	// Empty reads are skipped by Wireguard, so it is always legal
 	// to discard an empty packet instead of sending it through t.outbound.
-	outbound chan []byte
+	outbound chan tunReadResult
```
bradfitz, line 88: document whether/when this channel is closed. (never?)

net/tstun/wrap.go
```diff
@@ -325,26 +339,21 @@ func (t *Wrapper) IdleDuration() time.Duration {
 }
 
 func (t *Wrapper) Read(buf []byte, offset int) (int, error) {
-	var n int
-
+	res := <-t.outbound
```
bradfitz, line 342: how does a close unblock this?

net/tstun/wrap.go
josharian: I annotated them at the `defer` sites. Let me know whether that seems like enough on another read.

net/tstun/wrap.go
josharian: Done; same for bufferConsumed.

net/tstun/wrap.go
josharian: It didn't, and that mattered. Now it does.

josharian:
PTAL

net/tstun/wrap.go
```diff
@@ -160,14 +173,24 @@ func (t *Wrapper) SetDestIPActivityFuncs(m map[netaddr.IP]func()) {
...
+		close(t.outbound)
 		err = t.tdev.Close()
 	})
 	return err
 }
 
+// isClosed reports whether t is closed.
+func (t *Wrapper) isClosed() bool {
+	select {
+	case <-t.closed:
```
crawshaw, line 187: Is this as efficient as reading/writing an atomic? Maybe the loop that uses this is rare enough it doesn’t matter?

crawshaw (approved):

net/tstun/wrap.go
josharian: It's very efficient. The compiler converts it to a call to runtime.selectnbrecv. That calls runtime.chanrecv, which does (for this input):

* two comparisons
* three atomic loads
* a call to runtime.typedmemclr, which clears exactly 0 bytes

So it's not zero cost (a couple of function calls, a few atomic loads), but it's cheap enough to be effectively free compared to everything else around it. Most relevant, there is no path in which the scheduler is involved; it is limited to only memory reads and comparisons.

josharian: It's also not as good as it could be: https://github.com/golang/go/issues/47023

PR #2320, opened 2021-07-03
util/deephash: simplify hex encode

crawshaw:
With https://go-review.googlesource.com/c/go/+/332689,

Before:
	BenchmarkHash-24                   62504             19250 ns/op            2297 B/op         58 allocs/op
	BenchmarkHashMapAcyclic-24         39852             30119 ns/op            2532 B/op        202 allocs/op

After:
	BenchmarkHash-24                   62440             19271 ns/op            2298 B/op         58 allocs/op
	BenchmarkHashMapAcyclic-24         39592             30307 ns/op            2532 B/op        202 allocs/op

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -26,13 +26,7 @@ func calcHash(v interface{}) string {
 	scratch := make([]byte, 0, 128)
 	printTo(b, v, scratch)
 	b.Flush()
-	scratch = h.Sum(scratch[:0])
-	// The first sha256.Size bytes contain the hash.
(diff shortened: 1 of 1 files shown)


crawshaw:
I'm not convinced saving the extra alloc is worth the upstream commit. I wanted to put in code my case that reducing aliasing improves readability.

bradfitz (approved):
I'm fine with this if you get that std change into our Go tree.

bradfitz:
I've sent https://github.com/tailscale/tailscale/pull/2327 instead which, in the process of redoing the API for use for the control server, makes this PR needed.

PR #2327, opened 2021-07-05
util/deephash: export a Hash func for use by the control plane

bradfitz:
```
Hash-6              69.4µs ± 6%    68.4µs ± 4%     ~     (p=0.286 n=9+9)
HashMapAcyclic-6     115µs ± 5%     115µs ± 4%     ~     (p=1.000 n=10+10)

Hash-6              2.29kB ± 0%    1.88kB ± 0%  -18.13%  (p=0.000 n=10+10)
HashMapAcyclic-6    2.53kB ± 0%    2.53kB ± 0%     ~     (all equal)

Hash-6                58.0 ± 0%      54.0 ± 0%   -6.90%  (p=0.000 n=10+10)
HashMapAcyclic-6       202 ± 0%       202 ± 0%     ~     (all equal)
```

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -20,29 +20,67 @@ import (
 	"sync"
 )
 
-func calcHash(v interface{}) string {
-	h := sha256.New()
-	b := bufio.NewWriterSize(h, h.BlockSize())
-	scratch := make([]byte, 0, 128)
-	printTo(b, v, scratch)
-	b.Flush()
-	scratch = h.Sum(scratch[:0])
-	// The first sha256.Size bytes contain the hash.
-	// Hex-encode that into the next sha256.Size*2 bytes.
-	src := scratch[:sha256.Size]
-	dst := scratch[sha256.Size:cap(scratch)]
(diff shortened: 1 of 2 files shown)


bradfitz:
Note that this is the API that the control plane server wants (it works in terms of SHA-256 byte arrays, not strings)

bradfitz:
Just got done benchmarking its use in control: reduces overall CPU like 15%, and that's before some cheap wins we can do in deephash to do better, like calling String on NodeKey and MachineKey:

Screen Shot 2021-07-04 at 9 44 54 PM

util/deephash/deephash.go
```diff
@@ -20,29 +20,70 @@ import (
...
+	h       hash.Hash
+	bw      *bufio.Writer
+	scratch []byte
+}
+
+// newHasher initializes a new hasher, for use by hasherPool.
+func newHasher() *hasher {
+	h := &hasher{
+		h:       sha256.New(),
+		scratch: make([]byte, 0, 128),
```
bradfitz, line 35: oh I should inline this into hasher as an array

PR #2328, opened 2021-07-05
tstest/integration/vms: Oracle Linux Support

Xe:
Oracle Linux[1] is a CentOS fork. It is not very special. I am adding it
to the integration jungle.

[1]: https://www.oracle.com/linux/


crawshaw (approved):
Do we have an issue to add an Oracle Linux distribution to pkgs.tailscale.com? It seems we don't have one yet.

PR #2329, opened 2021-07-05
ipnlocal: allow windows to use guest VMs/containers while using an exit node

crawshaw:
Could be extended in the future to support Docker?


maisem (approved):

maisem:
> Could be extended in the future to support Docker?

Most likely, just need to investigate the container ip range.

PR #2332, opened 2021-07-05
tstest/integration/vms: make first end to end test

Xe:
This makes sure `tailscale status` and `tailscale ping` works. It also
switches goexpect to use a batch instead of manually banging out each
line, which makes the tests so much easier to read.

### DIFF
--- a/tstest/integration/vms/vms_test.go
+++ b/tstest/integration/vms/vms_test.go
@@ -627,6 +627,8 @@ func TestVMIntegrationEndToEnd(t *testing.T) {
 	ramsem := semaphore.NewWeighted(int64(*vmRamLimit))
 	bins := integration.BuildTestBinaries(t)
 
+	makeTestNode(t, bins, loginServer)
+
 	t.Run("do", func(t *testing.T) {
 		for n, distro := range distros {
 			n, distro := n, distro
@@ -718,9 +720,9 @@ func testDistro(t *testing.T, loginServer string, d Distro, signer ssh.Signer, i
 		expect.Verbose(true),
 		expect.VerboseWriter(logger.FuncWriter(t.Logf)),
 
-		// // NOTE(Xe): if you get a timeout, uncomment this line to have the raw
-		// output be sent to the test log quicker.
-		//expect.Tee(nopWriteCloser{logger.FuncWriter(t.Logf)}),
+		// // NOTE(Xe): if you get a timeout, uncomment this region to have the raw
+		// // output be sent to the test log quicker.
+		// expect.Tee(nopWriteCloser{logger.FuncWriter(t.Logf)}),
 	)
 	if err != nil {
 		t.Fatalf("%d: can't register a shell session: %v", port, err)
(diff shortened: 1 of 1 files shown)


Xe:
Seems to be oddly flaky. Will investigate and fix by tomorrow. I wonder if it's me assuming things wrongly about how the shell config in each distro works.

Xe:
AHA! I found the reason for the flakiness. I assumed that the way the expect library worked would that it would be line by line, like TCL expect. However sometimes the shell prompt gets drawn quick enough that it's in the previous testing batch. It actually works fine!

tstest/integration/vms/vms_test.go
```diff
@@ -729,35 +731,48 @@ func testDistro(t *testing.T, loginServer string, d Distro, signer ssh.Signer, i
...
-	err = e.Send(fmt.Sprintf("tailscale up --login-server %s\n", loginServer))
-	if err != nil {
-		t.Fatalf("%d: can't send tailscale up command: %v", port, err)
-	}
-	_, _, err = e.Expect(regexp.MustCompile(`Success.`), timeout)
+		batch = append(batch, &expect.BSnd{S: "systemctl start tailscaled.service\n"})
+	}
+
+	batch = append(batch,
+		&expect.BExp{R: `(\#)`},
```
crawshaw, line 752: You can remove some boilerplate here with:

```
batch = append(batch, []*expect.BExp{
        {R: `(\#)`},
        {S: `...`},
}...)
```

tstest/integration/vms/vms_test.go
```diff
@@ -874,6 +889,57 @@ func TestDeriveBindhost(t *testing.T) {
...
+	cmd.Env = append(os.Environ(), "NOTIFY_SOCKET="+filepath.Join(dir, "notify_socket"))
+
+	err := cmd.Start()
+	if err != nil {
+		t.Fatalf("can't start tailscaled: %v", err)
+	}
+
+	t.Cleanup(func() {
+		cmd.Process.Kill()
+		cmd.Process.Kill()
```
crawshaw, line 911: Why is this here twice?

tstest/integration/vms/vms_test.go
Xe: They are different types for sending and receiving.

crawshaw: Oh so they are. I misread it as using different fields on the same type.

tstest/integration/vms/vms_test.go
Xe: Overly defensive programming! I'm gonna prune it.

tstest/integration/vms/vms_test.go
```diff
@@ -874,6 +889,56 @@ func TestDeriveBindhost(t *testing.T) {
...
+	)
+
+	cmd.Env = append(os.Environ(), "NOTIFY_SOCKET="+filepath.Join(dir, "notify_socket"))
+
+	err := cmd.Start()
+	if err != nil {
+		t.Fatalf("can't start tailscaled: %v", err)
+	}
+
+	t.Cleanup(func() {
```
crawshaw, line 909: nit: `t.Cleanup(cmd.Process.Kill)`

crawshaw (approved):

tstest/integration/vms/vms_test.go
crawshaw: oh nevermind, Kill probably returns an error.

Xe: I can't do that, Kill returns an error because the kernel can deny a kill for permissions reasons. In this case I am assuming that this will work because this is a child process and the tests aren't running with the kernel capability needed to change user ids.

PR #2336, opened 2021-07-06
util/deephash: optimize hashing of byte arrays, reduce allocs in Hash

bradfitz:
```
Hash-6               173µs ± 4%     101µs ± 3%   -41.69%  (p=0.000 n=10+9)
HashMapAcyclic-6     101µs ± 5%     105µs ± 3%    +3.52%  (p=0.001 n=9+10)
TailcfgNode-6       29.4µs ± 2%    16.4µs ± 3%   -44.25%  (p=0.000 n=8+10)

Hash-6              3.60kB ± 0%    1.13kB ± 0%   -68.70%  (p=0.000 n=10+10)
HashMapAcyclic-6    2.53kB ± 0%    2.53kB ± 0%      ~     (p=0.137 n=10+8)
TailcfgNode-6         528B ± 0%        0B       -100.00%  (p=0.000 n=10+10)

Hash-6                84.0 ± 0%      40.0 ± 0%   -52.38%  (p=0.000 n=10+10)
HashMapAcyclic-6       202 ± 0%       202 ± 0%      ~     (all equal)
TailcfgNode-6         11.0 ± 0%       0.0       -100.00%  (p=0.000 n=10+10)
```

Updates tailscale/corp#2130

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -12,6 +12,7 @@ package deephash
 import (
 	"bufio"
 	"crypto/sha256"
+	"encoding/binary"
 	"encoding/hex"
 	"fmt"
 	"hash"
@@ -21,12 +22,14 @@ import (
 	"sync"
 )
 
+const scratchSize = 128
+
 // hasher is reusable state for hashing a value.
 // Get one via hasherPool.
 type hasher struct {
 	h       hash.Hash
 	bw      *bufio.Writer
-	scratch [128]byte
+	scratch [scratchSize]byte
 	visited map[uintptr]bool
 }
 
@@ -56,8 +59,13 @@ func (h *hasher) Hash(v interface{}) (hash [sha256.Size]byte) {
 	h.h.Reset()
 	h.print(reflect.ValueOf(v))
 	h.bw.Flush()
-	h.h.Sum(hash[:0])
-	return hash
+	// Sum into scratch & copy out, as hash.Hash is an interface
+	// so the slice necessarily escapes, and there's no sha256
+	// concrete type exported and we don't want the 'hash' result
+	// parameter to escape to the heap:
+	h.h.Sum(h.scratch[:0])
+	copy(hash[:], h.scratch[:])
+	return
 }
 
 var hasherPool = &sync.Pool{
@@ -107,6 +115,16 @@ type appenderTo interface {
 	AppendTo([]byte) []byte
 }
 
+func (h *hasher) uint(i uint64) {
+	binary.BigEndian.PutUint64(h.scratch[:8], i)
+	h.bw.Write(h.scratch[:8])
+}
+
+func (h *hasher) int(i int) {
+	binary.BigEndian.PutUint64(h.scratch[:8], uint64(i))
+	h.bw.Write(h.scratch[:8])
+}
+
 // print hashes v into w.
 // It reports whether it was able to do so without hitting a cycle.
 func (h *hasher) print(v reflect.Value) (acyclic bool) {
@@ -140,31 +158,40 @@ func (h *hasher) print(v reflect.Value) (acyclic bool) {
 		return h.print(v.Elem())
 	case reflect.Struct:
 		acyclic = true
-		w.WriteString("struct{\n")
+		w.WriteString("struct")
+		h.int(v.NumField())
 		for i, n := 0, v.NumField(); i < n; i++ {
-			fmt.Fprintf(w, " [%d]: ", i)
+			h.int(i)
 			if !h.print(v.Field(i)) {
 				acyclic = false
 			}
-			w.WriteString("\n")
 		}
-		w.WriteString("}\n")
 		return acyclic
 	case reflect.Slice, reflect.Array:
+		vLen := v.Len()
+		if v.Kind() == reflect.Slice {
+			h.int(vLen)
+		}
 		if v.Type().Elem().Kind() == reflect.Uint8 && v.CanInterface() {
-			fmt.Fprintf(w, "%q", v.Interface())
+			if vLen > 0 && vLen <= scratchSize {
+				// If it fits in scratch, avoid the Interface allocation.
+				// It seems tempting to do this for all sizes, doing
+				// scratchSize bytes at a time, but reflect.Slice seems
+				// to allocate, so it's not a win.
+				n := reflect.Copy(reflect.ValueOf(&h.scratch).Elem(), v)
+				w.Write(h.scratch[:n])
+				return true
+			}
+			fmt.Fprintf(w, "%s", v.Interface())
 			return true
 		}
-		fmt.Fprintf(w, "[%d]{\n", v.Len())
 		acyclic = true
-		for i, ln := 0, v.Len(); i < ln; i++ {
-			fmt.Fprintf(w, " [%d]: ", i)
+		for i := 0; i < vLen; i++ {
+			h.int(i)
 			if !h.print(v.Index(i)) {
 				acyclic = false
 			}
-			w.WriteString("\n")
 		}
-		w.WriteString("}\n")
 		return acyclic
 	case reflect.Interface:
 		return h.print(v.Elem())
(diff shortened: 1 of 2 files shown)


util/deephash/deephash.go
```diff
@@ -136,7 +137,21 @@ func print(w *bufio.Writer, v reflect.Value, visited map[uintptr]bool, scratch [
...
 	case reflect.Slice, reflect.Array:
 		if v.Type().Elem().Kind() == reflect.Uint8 && v.CanInterface() {
-			fmt.Fprintf(w, "%q", v.Interface())
+			if v.CanAddr() {
+				Len := v.Len()
+				w.Write(strconv.AppendInt(scratch[:0], int64(Len), 10))
+				w.WriteByte(':')
+				if Len > 0 {
+					var s []byte
+					sh := (*reflect.SliceHeader)(unsafe.Pointer(&s))
```
crawshaw, line 146: this is `unsafe.Slice` in the latest Go?

util/deephash/deephash.go
```diff
@@ -136,7 +137,21 @@ func print(w *bufio.Writer, v reflect.Value, visited map[uintptr]bool, scratch [
 		return acyclic
 	case reflect.Slice, reflect.Array:
 		if v.Type().Elem().Kind() == reflect.Uint8 && v.CanInterface() {
-			fmt.Fprintf(w, "%q", v.Interface())
+			if v.CanAddr() {
+				Len := v.Len()
+				w.Write(strconv.AppendInt(scratch[:0], int64(Len), 10))
+				w.WriteByte(':')
```
sailorfrag, line 143: would it be better to append the `:` to the output of `strconv.AppendInt` and then write the resulting slice, rather than a big write followed by a small write, or is that not going to matter?

For that matter, if we're worried about performance, would it be better to write the length with `encoding/binary` and skip the colon?

util/deephash/deephash.go
```diff
@@ -136,7 +137,21 @@ func print(w *bufio.Writer, v reflect.Value, visited map[uintptr]bool, scratch [
...
+				if Len > 0 {
+					var s []byte
+					sh := (*reflect.SliceHeader)(unsafe.Pointer(&s))
+					sh.Data = v.Addr().Pointer()
+					sh.Len = Len
+					sh.Cap = Len
+					w.Write(s) // unsafe slice escapes, but w is trusted *bufio.Writer
```
crawshaw, line 150: This is all to handle arrays? I assume for the slice case you could call `v.Bytes()`?

Could we make an argument upstream that `v.Bytes()` should work on uint8 arrays too?

crawshaw (commented):
Unfortunate to drop unsafe in the middle of a package that has turned out to be tricky. Isolate it into its own package? `reflectbytes.Write(io.Reader, reflect.Value)` or maybe `reflectextra.Bytes(v) []byte`

util/deephash/deephash.go
```diff
@@ -136,7 +137,21 @@ func print(w *bufio.Writer, v reflect.Value, visited map[uintptr]bool, scratch [
...
+				}
+			} else {
+				fmt.Fprintf(w, "%q", v.Interface())
```
sailorfrag, line 153: So the old way of printing with `%q` provides some protection against intentional collisions because `"` acts as a delimiter that is escaped if it appears inside the byte array. The preceding length may still provide that same protection but I'm less convinced about it. How important is this to worry about?

util/deephash/deephash.go
bradfitz: Changed to using binary throughout.

(removing the human-readable "deepprint" printable output vestiges)

util/deephash/deephash.go
bradfitz: It's a valid concern. I thought I'd addressed it, but I just made it more explicit in the latest round. All variable length items have a length first now.

util/deephash/deephash.go
bradfitz: Yes. But can't use it quite yet.

util/deephash/deephash.go
bradfitz: I forgot about Bytes. And it would almost work on Arrays too (using reflect.Value.Slice), but Value.Slice seems to allocate, killing the whole plan.

bradfitz:
PTAL

crawshaw:
What about my comment about isolating the unsafe to its own package?

Does the `reflect.Value.Bytes` suggestion deserve an upstream issue?

bradfitz:
> What about my comment about isolating the unsafe to its own package?

Doesn't seem worth it. This is already a pretty low-level reflect-y package. I can't think of a good API boundary between the two.

Plus maybe we can find a safe way to to do it later.

> Does the reflect.Value.Bytes suggestion deserve an upstream issue?

Sure: https://github.com/golang/go/issues/47066

crawshaw:
Can you help me understand why using unsafe here is worth it? The micro benchmark doesn’t explain to me the value.

josharian:
What type is this for? Can we make it implement AppendTo instead?

I went through this fight a bunch of different ways and eventually gave up, b/c I didn't want to use unsafe.

bradfitz:
> Can you help me understand why using unsafe here is worth it? The micro benchmark doesn’t explain to me the value.

The context is replacing control's JSON MapResponse minifier current JSON checksumming mechanism with deephash. So deephash would be seeing many more types it's never been optimized for before, including everything under `tailcfg.MapResponse`, notably lots of `tailcfg.Node` (which include a few array types that are currently unnecessarily costly)

bradfitz:
I forgot to cross-reference https://github.com/tailscale/corp/issues/2130

bradfitz:
> What type is this for? Can we make it implement AppendTo instead?

I'm not a huge fan of the AppendTo hack.

> I went through this fight a bunch of different ways and eventually gave up, b/c I didn't want to use unsafe.

I'll try a bit more to do this without unsafe, even if it's not quite as fast. But it's easy to do better than what we do now.

bradfitz:
I found a way to do this without unsafe, at least for all arrays <= the scratch space size, using reflect.Copy.

Updating this PR after some other cleanups & fixes.

bradfitz:
PTAL; now without unsafe.

util/deephash/deephash.go
```diff
@@ -140,31 +153,36 @@ func (h *hasher) print(v reflect.Value) (acyclic bool) {
...
 		return acyclic
 	case reflect.Slice, reflect.Array:
+		Len := v.Len()
+		if v.Kind() == reflect.Slice {
+			h.int(Len)
+		}
 		if v.Type().Elem().Kind() == reflect.Uint8 && v.CanInterface() {
-			fmt.Fprintf(w, "%q", v.Interface())
+			if Len > 0 && Len <= scratchSize {
+				n := reflect.Copy(reflect.ValueOf(&h.scratch).Elem(), v)
```
crawshaw, line 172: Does ValueOf allocate? If not, yay, the compiler is getting better.

util/deephash/deephash.go
```diff
@@ -140,31 +153,36 @@ func (h *hasher) print(v reflect.Value) (acyclic bool) {
...
+			h.int(i)
 			if !h.print(v.Field(i)) {
 				acyclic = false
 			}
-			w.WriteString("\n")
 		}
-		w.WriteString("}\n")
 		return acyclic
 	case reflect.Slice, reflect.Array:
+		Len := v.Len()
```
crawshaw, line 166: vLen?

crawshaw (approved):

util/deephash/deephash.go
```diff
@@ -140,31 +153,36 @@ func (h *hasher) print(v reflect.Value) (acyclic bool) {
...
+		if v.Kind() == reflect.Slice {
+			h.int(Len)
+		}
 		if v.Type().Elem().Kind() == reflect.Uint8 && v.CanInterface() {
-			fmt.Fprintf(w, "%q", v.Interface())
+			if Len > 0 && Len <= scratchSize {
```
josharian, line 171: Change this to a loop, so it works for all array lengths?

josharian (approved):

util/deephash/deephash.go
bradfitz: It hasn't for a long time, as long as you don't allocate into the interface. Hence the &+Elem.

util/deephash/deephash.go
bradfitz: Added a comment about why that's not a win.

sailorfrag:
Magic!

PR #2343, opened 2021-07-07
Bradfitz/deephash methods

bradfitz:


bradfitz:
Something's wrong with the second commit but I'm not seeing it, so going to bed. (It's full of WIP debugging junk trying to figure out why one of the tests if failing)

bradfitz:
I was almost asleep and realized what it was. Brains. Not clearing visited on reuse I bet.

PR #2344, opened 2021-07-07
build_docker.sh: use build_dist.sh to inject version information

pmirshad:
`version/version.sh` was removed in commit 5088af68. Use `build_dist.sh shellvars` to provide version information instead.


danderson (approved):

danderson:
Thanks!

PR #2349, opened 2021-07-07
tstest/integration/vms: bidirectional SSH tests

Xe:
This does two things:

1. Rewrites the tests so that we get a log of what individual tests failed at the end of a test run.
2. Adds a test that runs an SSH server via the tester tailscale node and then has the VMs connect to that over Tailscale.
3. Adds a test that makes SSH connections to each VM being tested and ensures that `tailscale ip` works.

This tests outgoing TCP connections.

### DIFF
--- a/tstest/integration/vms/vms_test.go
+++ b/tstest/integration/vms/vms_test.go
@@ -7,6 +7,7 @@
 package vms
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
@@ -37,6 +38,7 @@ import (
 	expect "github.com/google/goexpect"
 	"github.com/pkg/sftp"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/net/proxy"
 	"golang.org/x/sync/semaphore"
 	"inet.af/netaddr"
 	"tailscale.com/net/interfaces"
@@ -62,6 +64,15 @@ var (
 		return result
 	}()
 )
+
+type Harness struct {
+	testerDialer   proxy.Dialer
+	testerDir      string
+	bins           *integration.Binaries
+	signer         ssh.Signer
+	cs             *testcontrol.Server
+	loginServerURL string
+}
 
 type Distro struct {
 	name           string // amazon-linux
@@ -627,7 +638,14 @@ func TestVMIntegrationEndToEnd(t *testing.T) {
 	ramsem := semaphore.NewWeighted(int64(*vmRamLimit))
 	bins := integration.BuildTestBinaries(t)
 
-	makeTestNode(t, bins, loginServer)
+	h := &Harness{
+		bins:           bins,
+		signer:         signer,
+		loginServerURL: loginServer,
+		cs:             cs,
+	}
+
+	h.makeTestNode(t, bins, loginServer)
 
 	t.Run("do", func(t *testing.T) {
 		for n, distro := range distros {
@@ -670,13 +688,17 @@ func TestVMIntegrationEndToEnd(t *testing.T) {
 					}
 				})
 
-				testDistro(t, loginServer, distro, signer, ipm, bins)
+				h.testDistro(t, distro, ipm)
 			})
 		}
 	})
 }
 
-func testDistro(t *testing.T, loginServer string, d Distro, signer ssh.Signer, ipm ipMapping, bins *integration.Binaries) {
+func (h Harness) testDistro(t *testing.T, d Distro, ipm ipMapping) {
+	signer := h.signer
+	bins := h.bins
+	loginServer := h.loginServerURL
+
 	t.Helper()
 	port := ipm.port
 	hostport := fmt.Sprintf("127.0.0.1:%d", port)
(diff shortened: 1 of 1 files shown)


tstest/integration/vms/ssh_test.go
```diff
@@ -0,0 +1,133 @@
...
+	signer, err := ssh.ParsePrivateKey(privateKey)
+	if err != nil {
+		t.Fatalf("can't parse private key: %v", err)
+	}
+
+	addr := mkSSHServer(t, signer, "::1")
+	t.Logf("connecting to %s", addr)
+	host, port, _ := net.SplitHostPort(addr)
+
+	cmd := exec.Command("ssh",
```
Xe, line 110: XXX: I tried to use go's SSH library for this but it just wouldn't work. The way my SSH server works is that it just spews stuff and kills the session afterwards. This is apparently in violation of how SSH servers are supposed to normally work, but the goal here is to get the text spewed back so it doesn't really matter.

tstest/integration/vms/vms_test.go
```diff
@@ -716,6 +716,68 @@ func testDistro(t *testing.T, loginServer string, d Distro, signer ssh.Signer, i
...
+		})
+	})
+
+	t.Run("ssh-server", func(t *testing.T) {
+		sshServer := mkSSHServer(t, signer, "::")
+		_, port, _ := net.SplitHostPort(sshServer)
+		runTestCommands(t, timeout, cli, []expect.Batcher{
+			&expect.BSnd{
+				S: fmt.Sprintf(
+					"ssh -o StrictHostKeyChecking=no -o GlobalKnownHostsFile=/dev/null -o UserKnownHostsFile=/dev/null %s -p %s\n",
```
Xe, line 771: XXX: You may ask "Why SSH?". It's guaranteed to be on every one of these VMs and I know it works over TCP. This is a case of the most common denominator.

crawshaw: Explanations about why code is a certain way belong in comments. Future readers of the code will be similarly perplexed.

tstest/integration/vms/vms_test.go
```diff
@@ -716,6 +716,68 @@ func testDistro(t *testing.T, loginServer string, d Distro, signer ssh.Signer, i
...
+	t.Run("ping-ipv4", func(t *testing.T) {
+		runTestCommands(t, timeout, cli, []expect.Batcher{
+			&expect.BSnd{S: "tailscale ping -c 1 100.64.0.1\n"},
+			&expect.BExp{R: `pong from.*\(100.64.0.1\)`},
+			&expect.BSnd{S: "ping -c 1 100.64.0.1\n"},
+			&expect.BExp{R: `bytes`},
+		})
+	})
+
+	t.Run("ssh-server", func(t *testing.T) {
```
crawshaw, line 765: It looks like every test invokes ssh now. What does this test add?

tstest/integration/vms/ssh_test.go
crawshaw: comments like this need to be in the code.

tstest/integration/vms/ssh_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"path/filepath"
+	"testing"
+	"time"
+
+	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/terminal"
+	"tailscale.com/types/logger"
+)
+
+func mkSSHServer(t *testing.T, hostKey ssh.Signer, bindhost string) string {
```
crawshaw, line 25: This is a surprising choice to me. ssh is a pain to use without a lot of setup. Do some of these systems ship without netcat or curl?

Xe: Netcat is flag-inconsistent across distros and not in all of the base images. We do add curl for other reasons in the test stack, however SSH more closely represents how people use Tailscale.

tstest/integration/vms/vms_test.go
Xe: https://github.com/tailscale/tailscale/pull/2349#discussion_r665620161

tstest/integration/vms/vms_test.go
Xe: I added a comment above the test

tstest/integration/vms/ssh_test.go
Xe: Added in a comment

tstest/integration/vms/ssh_test.go
crawshaw: I don't understand what running an ssh server in this unusual mode, moving text in a way that, according to your comment isn't the way ssh is meant to be used, tests tailscale in any useful way that an HTTP POST from curl wouldn't. They are both, from Tailscale's perspective, typical TCP sockets.

If you want to submit this to get on with working on some other part of this system that's fine, but add a `TODO: replace with something simpler like curl` to the top.

crawshaw:
Is this ready for review?

tstest/integration/vms/ssh_test.go
Xe: I made it use HTTP instead.

crawshaw (approved):

PR #2351, opened 2021-07-07
tstest/integration: help bust cmd/go's test caching

bradfitz:
It was caching too aggressively, as it didn't see our deps due to our
running "go install tailscaled" as a child process.


tstest/integration/integration_test.go
```diff
@@ -35,6 +35,18 @@ import (
 	"tailscale.com/tstest"
 	"tailscale.com/tstest/integration/testcontrol"
 	"tailscale.com/types/logger"
+
+	// And depend on a bunch of tailscaled innards, for Go's test caching.
+	// Otherwise cmd/go never sees that we depend on these packages'
+	// transitively deps when we run "go install tailscaled" in a child
```
josharian, line 41: transitive

josharian (approved):
how do we keep this up to date?

can we make cmd/tailscaled a three-line wrapper around the real cmd/tailscaled (`func main() { tailscaled.Main() }` and then blank import those guts here instead of this long list?

bradfitz:
We split it once, in b81bd8025b71f35295201d18011855c1cfae630e, and then reverted it in dad10fee9cf5cf9d2c6a7565213921892c05bae1.

It's pretty unfortunate having to do that.

I'd rather auto-generate a package with the same deps with go generate and write a test that it's in sync.

bradfitz:
Added a "go generate" command. No test (yet), though. PTAL?

crawshaw (approved):

josharian:
LGTM, thanks

josharian:
Instead of a local test, a CI check might be easier. Clone the repo, run `go generate ./...`, fail if dirty. As a bonus, ensures that we write down exactly how to get all the tools required to run our `go generate`. WDYT?

PR #2352, opened 2021-07-07
net/tstun: fix data races

josharian:
To remove some multi-case selects, we intentionally allowed
sends on closed channels (cc23049cd286f137536d011b7840b3be5fdc8cb4).

However, we also introduced concurrent sends and closes,
which is a data race.

This commit fixes the data race. The mutexes here are uncontended,
and thus very cheap.

### DIFF
--- a/net/tstun/wrap.go
+++ b/net/tstun/wrap.go
@@ -71,6 +71,9 @@ type Wrapper struct {
 	// buffer stores the oldest unconsumed packet from tdev.
 	// It is made a static buffer in order to avoid allocations.
 	buffer [maxBufferSize]byte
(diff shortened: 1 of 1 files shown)


josharian:
The sadness grows.

crawshaw (approved):
If we had generics, I would say we could improve readability by creating a closableChan object with a send method wrapping this panic logic.

As it is, we would have to have two of them for the different channel types. But maybe it's worth it anyway? It would let us move these semantics to their own file, document just one type, and put a big TODO: use parameterized types for this at the top.

josharian:
I'd feel more motivated to do that if we had generics. As it is, two's company...if (god forbid) we ever have a crowd, I'll do that.

PR #2353, opened 2021-07-08
net/uring: add go ntohs

JulianKnodt:
Instead of calling out to C for ntohs, just implement it in Go for effiency.

### DIFF
--- a/net/uring/io_uring_linux.go
+++ b/net/uring/io_uring_linux.go
@@ -21,6 +21,7 @@ import (
 	"golang.zx2c4.com/wireguard/device"
 	"golang.zx2c4.com/wireguard/tun"
 	"inet.af/netaddr"
+	"tailscale.com/util/endian"
 )
 
 const bufferSize = device.MaxSegmentSize
@@ -172,10 +173,10 @@ func (u *UDPConn) ReadFromNetaddr(buf []byte) (int, netaddr.IPPort, error) {
 	// TODO: native go endianness conversion routines so we don't have to call ntohl, etc.
 	if u.is4 {
 		ip = netaddr.IPFrom4(*(*[4]byte)((unsafe.Pointer)((&r.sa.sin_addr.s_addr))))
-		port = uint16(C.ntohs(r.sa.sin_port))
+		port = endian.Ntoh16(uint16(r.sa.sin_port))
 	} else {
 		ip = netaddr.IPFrom16(*(*[16]byte)((unsafe.Pointer)((&r.sa6.sin6_addr))))
-		port = uint16(C.ntohs(r.sa6.sin6_port))
+		port = endian.Ntoh16(uint16(r.sa6.sin6_port))
 	}
 	ipp := netaddr.IPPortFrom(ip, port)
 	rbuf := sliceOf(r.buf, n)
@@ -289,16 +290,15 @@ func (u *UDPConn) WriteTo(p []byte, addr net.Addr) (n int, err error) {
 	copy(rbuf, p)
 
 	if u.is4 {
-		// TODO: native go endianness conversion routines so we don't have to call ntohl, etc.
 		ipu32 := binary.BigEndian.Uint32(udpAddr.IP)
-		r.sa.sin_addr.s_addr = C.htonl(C.uint32_t(ipu32))
-		r.sa.sin_port = C.htons(C.uint16_t(udpAddr.Port))
+		r.sa.sin_addr.s_addr = C.uint32_t(endian.Hton32(ipu32))
+		r.sa.sin_port = C.uint16_t(endian.Hton16(uint16(udpAddr.Port)))
 		r.sa.sin_family = C.AF_INET
 	} else {
 		dst := (*[16]byte)((unsafe.Pointer)(&r.sa6.sin6_addr))
 		src := (*[16]byte)((unsafe.Pointer)(&udpAddr.IP[0]))
 		*dst = *src
-		r.sa6.sin6_port = C.htons(C.uint16_t(udpAddr.Port))
+		r.sa6.sin6_port = C.uint16_t(endian.Hton16(uint16(udpAddr.Port)))
 		r.sa6.sin6_family = C.AF_INET6
 	}
 	C.submit_sendmsg_request(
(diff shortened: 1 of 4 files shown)


net/uring/encoding.go
```diff
@@ -0,0 +1,51 @@
...
+		nativeEndian = binary.LittleEndian
+	case [2]byte{0xAB, 0xCD}:
+		nativeEndian = binary.BigEndian
+	default:
+		panic("Could not determine native endianness.")
+	}
+}
+
+// TODO if we want to not make these into runtime checks, split this into 3 separate files with
+// build tags.
```
josharian, line 26: Yes, build tags please.

net/uring/encoding.go
```diff
@@ -0,0 +1,51 @@
...
+	} else {
+		return bits.Reverse16(v)
+	}
+}
+
+func htons(v uint16) uint16 {
+	if nativeEndian == binary.BigEndian {
+		return v
+	} else {
+		return bits.Reverse16(v)
```
josharian, line 41: ReverseBytes.

This needs tests.

net/uring/netenc/big.go
```diff
@@ -0,0 +1,14 @@
+// +build mips mips64 ppc64 s390x
+
+package netenc
+
+// TODO how should I capitalize this
```
josharian, line 5: Good question. The standard library (strong) uses Atoi, which suggests Ntoh. We could also do NetworkToHost, which is nice and explicit. I think either would be OK. On a second look, we should probably use Ntoh16 instead of Ntohs, though, following math/bits and encoding/binary naming conventions.

net/uring/netenc/unknown.go
```diff
@@ -0,0 +1,14 @@
+// +build !mips,!mips64,!ppc64,!s390x,!amd64,!386,!arm,!arm64,!mipsle,!mips64le,!ppc64le,!riscv64,!wasm
+
+package netenc
+
+import "C"
+
+// Ntohs converts network into native/host order.
+func Ntohs(v uint16) uint16 { return uint16(C.ntohs(C.uint16_t(v))) }
```
josharian, line 8: I anticipate that we're going to attempt to get rid of the cgo dependency sooner rather than later. Let's omit this file entirely. If Go adds a new platform, compilation will fail, and we'll know to add it.

net/uring/netenc/encoding_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+func TestNtohs(t *testing.T) {
+	v := [2]byte{0xCD, 0xAB}
+	big := binary.BigEndian.Uint16(v[:])
+	if 0xABCD != Ntohs(big) {
+		t.Errorf("ntohs failed, want %v, got %v", 0xABCD, Ntohs(big))
+	}
+}
+
+func toBytes32(v uint32) [4]byte {
```
josharian, line 16: This is fine, but out of curiosity, why not use encoding/binary here?

net/uring/netenc/encoding_test.go
```diff
@@ -0,0 +1,54 @@
...
+	out := [4]byte{}
```
josharian, line 17: This is typically spelled `var out [4]byte`.

Or simply

```go
return [4]byte{
byte(v & 0xFF),
// ...
// ...
// ...
}
```

josharian (commented):
You accidentally committed a swap file. :)

net/uring/netenc/encoding_test.go
JulianKnodt: oh I was thinking that this should work with the native endianness (I might be wrong), but binary encoding requires passing either little or big

josharian: Silly me. The only way to do native endianness is with unsafe. So these tests are going to need to use unsafe. (That's OK; unsafe in tests is a different matter than unsafe not in tests.)

Something like (untested):

```go
func TestHton16(t *testing.T) {
  h := 0xABCD
  n := Hton16(h)
  got := *(*[4]byte)(unsafe.Pointer(&n))
  want := [4]byte{0xA, 0xB, 0xC, 0xD}
  if want != got {
    qt.Assert(want, qt.Equals, got)
  }
}
```

net/uring/netenc/big.go
```diff
@@ -0,0 +1,12 @@
+// +build mips mips64 ppc64 s390x
+
+package netenc
+
+// Ntohs converts network order into native/host.
```
josharian, line 5: Update the doc strings with the new function names.

josharian (approved):
LGTM, once the requested changes are incorporated.

bradfitz:
Don't we already have this code 2 other places? This looks like it's just adding a third.

JulianKnodt:
where else is this?

bradfitz:
See `util/endian.Native`

bradfitz:
The other is `github.com/josharian/native`, as used by `github.com/mdlayher/netlink`. (as reported by cmd/tailscaled/depaware.txt)

JulianKnodt:
I could just stick these functions inside of that package, I wasn't sure where to put them in the first place

josharian:
Oops, thanks @bradfitz, I forgot we had a copy in our tree. @JulianKnodt that looks like a fine home for them. I'm still not convinced the tests are doing the right thing; I think you need unsafe or to use endian.Native to actually test.

JulianKnodt:
@josharian they have unsafe in them, so I think they should be correct

josharian:
Ah, I see. I think TestNtoh16 is still missing an unsafe conversion.

JulianKnodt:
ah yea, that one I just left in as plain because it's converting from network to host, so it just needs to convert from big endian

josharian:
> ah yea, that one I just left without unsafe because it's converting from network to host, so it just needs to convert from big endian and that can be manually compared against the expected number

That test will currently fail on a big-endian system.

```
	v := [2]byte{0xCD, 0xAB}
	big := binary.BigEndian.Uint16(v[:])
```

after this, big == 0xCDAB, always.

It needs an unsafe conversion (or a native endian conversion), like the others.

JulianKnodt:
I think it should finally be good

josharian:
Thanks!

PR #2360, opened 2021-07-08
tstest/integration/vms: use an in-process logcatcher

Xe:
This adapts the existing in-process logcatcher from tstest/integration
into a public type and uses it on the side of testcontrol. This also
fixes a bug in the Alpine Linux OpenRC unit that makes every value in
`/etc/default/tailscaled` exported into tailscaled's environment, a-la
systemd [Service].EnviromentFile.

### DIFF
--- a/tstest/integration/vms/vms_test.go
+++ b/tstest/integration/vms/vms_test.go
@@ -126,7 +126,7 @@ func TestDownloadImages(t *testing.T) {
 
 			t.Parallel()
 
-			fetchDistro(t, distro, bins)
+			(Harness{bins: bins}).fetchDistro(t, distro)
 		})
 	}
 }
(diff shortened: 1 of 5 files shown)


tstest/integration/vms/vms_test.go
```diff
@@ -242,7 +242,7 @@ func fetchFromS3(t *testing.T, fout *os.File, d Distro) bool {
 
 // fetchDistro fetches a distribution from the internet if it doesn't already exist locally. It
 // also validates the sha256 sum from a known good hash.
-func fetchDistro(t *testing.T, resultDistro Distro, bins *integration.Binaries) string {
+func (h Harness) fetchDistro(t *testing.T, resultDistro Distro) string {
```
crawshaw, line 245: Should these be pointer-recievers, i.e. *Harness?

crawshaw (approved):

tstest/integration/vms/vms_test.go
Xe: They aren't mutating the values of the Harness struct and this is non-hot code. I'd prefer to take the "immutability" over speedhacking a test with a pointer receiver.

PR #2361, opened 2021-07-08
tstest/integration/vms: end-to-end UDP test

Xe:
This tests incoming and outgoing UDP traffic. It would test incoming UDP
traffic however our socks server doesn't seem to allow for connecting to
destinations over UDP. When the socks server gets that support the
incoming test should pass without issue.

### DIFF
--- a/tstest/integration/vms/udp_tester.go
+++ b/tstest/integration/vms/udp_tester.go
@@ -1,0 +1,80 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build ignore
+
+// Command udp_tester exists because all of these distros being tested don't
+// have a consistent tool for doing UDP traffic. This is a very hacked up tool
+// that does that UDP traffic so these tests can be done.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"net"
+	"os"
+)
+
+var (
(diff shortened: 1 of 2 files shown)


Xe:
test run: https://github.com/tailscale/tailscale/pull/2361/checks?check_run_id=3022172199

tstest/integration/vms/udp_tester.go
```diff
@@ -0,0 +1,77 @@
...
+	"net"
+	"os"
+)
+
+var (
+	client = flag.String("client", "", "host:port to connect to for sending UDP")
+	server = flag.String("server", "", "host:port to bind to for receiving UDP")
+)
+
+func main() {
```
crawshaw, line 23: `nc -u` isn't portable enough?

crawshaw (approved):

tstest/integration/vms/udp_tester.go
Xe: that implies that netcat is on all of the base images and the `netcat` package in the repos is the same netcat. This is not the case and I don't want to have to pick apart the differences if I can avoid it.

Xe: I'll add a comment though

crawshaw:
FYI: my general review process is "approve with comments" means if you follow the requests in the comments, go ahead and merge. If a comment has a ? on it, it's by its nature just a request for information, so no need to block merging on it. We haven't fully standardized this across the company, it's on my TODO list.

Xe:
I'm aware, I just wanted to wait for CI to pass :)

PR #2363, opened 2021-07-08
wgengine/magicsock: make portmap non-blocking

JulianKnodt:
Since portmapping may be expensive to perform for UPnP, make it asynchronous in magicsock so
that it's fine that it may take longer. This is an even smaller subset of changes to get UPnP working.
Previously I had hoped to make the prober itself asynchronous, but instead it might make more sense to leave it synchronous and not block in the places where performance matters a lot.

**Important Note**

I can't tell if this causes any behavior differences since I don't have any portmapping on my network other than UPnP

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -635,6 +635,35 @@ func (c *Conn) setEndpoints(endpoints []tailcfg.Endpoint) (changed bool) {
 	return true
 }
 
+// addLatePortMapEndpoint adds a portmapping endpoint to the endpoint set if it could not
+// be created/gotten during determineEndpoints.
+func (c *Conn) addLatePortMapEndpoint(addr netaddr.IPPort) (added bool) {
+	if addr.IsZero() {
+		return false
+	}
+	ep := tailcfg.Endpoint{Addr: addr, Type: tailcfg.EndpointPortmapped}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.closed || c.privateKey.IsZero() {
+		return false
+	}
+
+	if c.derpMap == nil && !inTest() {
+		c.logf("[v1] magicsock: ignoring pre-DERP map, STUN-less port map update: %v", ep)
+		return false
+	}
+	for _, v := range c.lastEndpoints {
+		if v == ep {
+			return false
+		}
+	}
+	c.lastEndpoints = append(c.lastEndpoints, ep)
+	c.logEndpointChange(c.lastEndpoints)
+	c.epFunc(c.lastEndpoints)
+	return true
+}
+
 // setNetInfoHavePortMap updates NetInfo.HavePortMap to true.
 func (c *Conn) setNetInfoHavePortMap() {
 	c.mu.Lock()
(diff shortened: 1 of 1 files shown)


wgengine/magicsock/magicsock.go
```diff
@@ -635,6 +635,27 @@ func (c *Conn) setEndpoints(endpoints []tailcfg.Endpoint) (changed bool) {
...
+		c.logf("[v1] magicsock: ignoring pre-DERP map, STUN-less port map update: %v", ep)
+		return false
+	}
+	for _, v := range c.lastEndpoints {
+		if v == ep {
+			return false
+		}
+	}
+	c.lastEndpoints = append(c.lastEndpoints, ep)
+	c.logEndpointChange([]tailcfg.Endpoint{ep})
```
bradfitz, line 654: This has to be the full set of endpoints. Likewise with epFunc.

wgengine/magicsock/magicsock.go
```diff
@@ -635,6 +635,27 @@ func (c *Conn) setEndpoints(endpoints []tailcfg.Endpoint) (changed bool) {
 	return true
 }
 
+// addLatePortMapEndpoint adds a portmapping endpoint to the endpoint set if it could not
+// be created/gotten during determineEndpoints.
+func (c *Conn) addLatePortMapEndpoint(addr netaddr.IPPort) (added bool) {
+	ep := tailcfg.Endpoint{Addr: addr, Type: tailcfg.EndpointPortmapped}
+	c.mu.Lock()
```
bradfitz, line 642: after acquiring the lock, you need to double check various invariants about the state of the world. Are we even still running, for instance? We might be closed or stopped.

wgengine/magicsock/magicsock.go
```diff
@@ -985,6 +1006,16 @@ func (c *Conn) determineEndpoints(ctx context.Context) ([]tailcfg.Endpoint, erro
 		return nil, err
 	}
 
+	portmap := make(chan netaddr.IPPort, 1)
+	go func() {
```
bradfitz, line 1010: Debugging is easier when goroutines are all named. Move this to a method on Conn and then `go c.getPortMapping` ?

wgengine/magicsock/magicsock.go
```diff
@@ -985,6 +1006,16 @@ func (c *Conn) determineEndpoints(ctx context.Context) ([]tailcfg.Endpoint, erro
...
+		if ext, err := c.portMapper.CreateOrGetMapping(ctx); err == nil {
+			portmap <- ext
+		} else if !portmapper.IsNoMappingError(err) {
+			c.logf("portmapper: %v", err)
+		}
+		close(portmap)
```
bradfitz, line 1016: maybe just close it in the failure path? It's a little redundant in normal operation.

wgengine/magicsock/magicsock.go
```diff
@@ -1051,6 +1075,22 @@ func (c *Conn) determineEndpoints(ctx context.Context) ([]tailcfg.Endpoint, erro
...
+	case ext, ok := <-portmap:
+		if ok {
+			addAddr(ext, tailcfg.EndpointPortmapped)
+			c.setNetInfoHavePortMap()
+		}
+	default:
+		// fire off a goroutine which will update netinfo later
+		go func() {
+			if ext, ok := <-portmap; ok {
+				c.setNetInfoHavePortMap()
```
bradfitz, line 1088: Hm, I think both of these lines can trigger MapRequest updates. Ideally they'd ride along together to the control server.

wgengine/magicsock/magicsock.go
JulianKnodt: I added a check for if it's closed, but are there other invariants to check?

wgengine/magicsock/magicsock.go
JulianKnodt: This is a bit confusing, but having it there is necessary if there was a `NoMappingError`, otherwise the other branch won't be hit.

wgengine/magicsock/magicsock.go
JulianKnodt: Wouldn't this have been happening previously as well? I'm not sure I'm able to change that at this level since they're both behind function pointers, although I could add an extra flag to only invoke it once or something?

JulianKnodt:
Closed in favor of https://github.com/tailscale/tailscale/pull/2375

PR #2366, opened 2021-07-08
net/uring: add probing capability

JulianKnodt:
Adds the ability to probe for various capabilities. It will not call into C unless
necessary. It also allocates one probe per call to new capability, which may be expensive, so in
theory they could be reused instead.

### DIFF
--- a/net/uring/capability_linux.go
+++ b/net/uring/capability_linux.go
@@ -1,0 +1,96 @@
+package uring
+
+// #cgo CFLAGS: -I${SRCDIR}/liburing/src/include
+// #cgo LDFLAGS: -L${SRCDIR}/liburing/src/ -luring
+// #include "io_uring.c"
+import "C"
+
+import (
+	"sync"
+	"syscall"
+	"unsafe"
+)
+
+type op = int
+
+// https://unixism.net/loti/tutorial/probe_liburing.html
+const (
+	opNop op = iota
+	opReadv
+	opWritev
+	opFsync
+	opReadFixed
+	opWriteFixed
+	opPollAdd
+	opPollRemove
+	opSyncFileRange
+	opSendMsg
+	opRecvMsg
+	opTimeout
+	opTimeoutRemove
+	opAccept
+	opAsyncCancel
+	opLinkTimeout
+	opConnect
+	opFAllocate
+	opOpenAt
+	opClose
+	opFilesUpdate
+	opStatx
+	opRead
+	opWrite
+	opFAdvise
+	opMAdvise
+	opSend
+	opRecv
+	opOpenAt2
+	opEPollControl
+	opSplice
+	opProvideBuffers
+	opRemoveBuffers
+	opCount
+)
+
+var (
+	// probed ensures that capabilities is only populated once.
+	probed sync.Once
+	// capabilities is the list of available capabilities on this system
+	capabilities map[op]bool
+	// probingSupported is whether probing is available on this system.
+	probingSupported bool
+)
+
+// checkCapability returns whether probing is supported, and if and only if probing is
+// supported, whether the operation is also supported.
+// If probing is not supported, it may still be the case that the operation can be used.
+//
+// To test for a specific operation if there is no probing, probably running it and seeing if
+// the output is non-nil should be enough.
+func checkCapability(o op) (probingSupported bool, opOk bool) {
+	probed.Do(func() {
+		probe, err := C.io_uring_get_probe()
+		if probe == nil || err != nil {
+			probingSupported = false
+			return
+		}
+		capabilities = map[op]bool{}
+		defer C.free(unsafe.Pointer(probe))
+		for i := 0; i < opCount; i++ {
+			capabilities[i] = C.io_uring_opcode_supported(probe, C.int(i)) != 0
+		}
+	})
+
+	if !probingSupported {
+		return false, false
+	}
+	return true, capabilities[o]
+}
+
+// uringOnSystem will check if it is possible to use io_uring syscalls on the system.
+func uringOnSystem() bool {
+	probe, err := C.io_uring_get_probe()
+	if err == nil && probe != nil {
+		C.free(unsafe.Pointer(probe))
+	}
+	return err != syscall.ENOSYS
+}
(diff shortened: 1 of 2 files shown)


net/uring/all.go
```diff
@@ -9,3 +9,42 @@ import "runtime"
 func Available() bool {
 	return runtime.GOOS == "linux"
 }
+
+type IORingOp = int
```
josharian, line 13: Let's put this (and all the other capability probing code) in a file called capability_linux.go. Other platforms don't care (doesn't need to be in all.go), and it's a nice, simple, standalone bit of code.

Can we call it just `op`? Then the type is `uring.op`, which seems appropriate. Also, unexpected is better.

net/uring/all.go
```diff
@@ -9,3 +9,42 @@ import "runtime"
...
+
+//https://unixism.net/loti/tutorial/probe_liburing.html
+const (
+	IORING_OP_NOP IORingOp = iota
```
josharian, line 17: Let's spell these in the Go way, not the C way:

opNop
opReadv

etc.

net/uring/all.go
```diff
@@ -9,3 +9,42 @@ import "runtime"
 func Available() bool {
 	return runtime.GOOS == "linux"
 }
+
+type IORingOp = int
+
+//https://unixism.net/loti/tutorial/probe_liburing.html
```
josharian, line 15: space between `//` and comment, please

net/uring/io_uring.c
```diff
@@ -160,3 +160,23 @@ static go_completion_result completion(struct io_uring *ring, int block) {
     io_uring_cqe_seen(ring, cqe);
     return res;
 }
+
+typedef struct {
+  int probingNotSupported;
```
josharian, line 165: This is C, so use snake_case instead of camelCase.

Use positive semantics, so this should be `probing_supported`.

net/uring/io_uring.c
```diff
@@ -160,3 +160,23 @@ static go_completion_result completion(struct io_uring *ring, int block) {
...
+  int supported;
```
josharian, line 166: Call this `capability_supported`, or the like.

net/uring/io_uring.c
```diff
@@ -160,3 +160,23 @@ static go_completion_result completion(struct io_uring *ring, int block) {
...
+} probeCheck;
```
josharian, line 167: `probe_result`?

net/uring/io_uring.c
```diff
@@ -160,3 +160,23 @@ static go_completion_result completion(struct io_uring *ring, int block) {
...
+
+
+// has_capability takes the index of a capability, and returns whether the operation is
+// supported, and if probing itself is supported.
+static probeCheck has_capability(int i) {
+    probeCheck pc = {0,0};
+    struct io_uring_probe *probe = io_uring_get_probe();
```
josharian, line 174: You can probably do this entirely in Go/cgo, by calling `C.io_uring_get_probe` and `C.io_uring_opcode_supported`. That'd avoid needing to define a C struct etc.

The borderline between "write a C helper function" and "call individual C functions as needed" is a bit unclear. But for this case--performance doesn't matter, just two calls, need to return multiple values--it seems to me to be pretty clear.

net/uring/io_uring_linux.go
```diff
@@ -602,3 +602,35 @@ func (t *TUN) Close() error {
...
+	probingNotSupported bool              = false
+)
+
+// checkCapability returns whether probing is supported, and if and only if probing is
+// supported, whether the operation is also supported.
+// If probing is not supported, it may still be the case that the operation can be used.
+//
+// To test for a specific operation if there is no probing, probably running it and seeing if
+// the output is non-nil should be enough.
+func checkCapability(op IORingOp) (probingSupported bool, opOk bool) {
```
josharian, line 619: I'm thinking about how to incorporate this into our current setup. What happens if probingSupported is false? Should we proceed or not? It appears that this style of probing isn't supported on the kernel I'm developing on (`5.4.0-66-generic`), and yet our code works. That says to me that probing this way isn't what we need, unfortunately. We might still want to do this in the future, to optionally use different ops, but our minimalist set appears to be enough.

So we need to find some different way to probe for io_uring support. Maybe we should start with making some unimportant io_uring syscall and checking whether it returns ENOSYS. Maybe later it'll turn out we also need to check the kernel version to require some bug fixes.

net/uring/io_uring_linux.go
```diff
@@ -602,3 +602,35 @@ func (t *TUN) Close() error {
 	}
 	return err2
 }
+
+var (
+	mu sync.Mutex
```
josharian, line 607: Given that these probes are pretty cheap, I'd be tempted to use a sync.Once and populate the entire capabilities map. Then the function can return a nil capabilities map if probing is unsupported and a regular (re-usable) capabilities map otherwise.

josharian (commented):
Please read through all the comments before working on this more--looks like we need a change of direction. I left the other comments in case we want to keep this code, and for pedagogical purposes.

net/uring/io_uring_linux.go
JulianKnodt: I'm still not returning a map, and instead just populating a singleton and indexing into it later. Hopefully that's good enough

net/uring/io_uring_linux.go
JulianKnodt: For a specific syscall, probably it makes sense to just try it out and see if it fails either by returning null, or ENOSYS. That being said, if this is available this would be faster to check, so prefer probing, then populating the map with syscall attempts?

josharian: Let's build the API to match our specific needs.

Our current need is to decide whether we can use io_uring; that's a binary decision. Probably the right way to make that decision--since probing is not available early enough--is to choose a single representative syscall and check for ENOSYS.

josharian:
Oops. I renamed the branch and apparently that closed this PR. Sorry about that. The new branch is josh/io_uring (https://github.com/tailscale/tailscale/pull/2370)

PR #2368, opened 2021-07-08
wgengine/magicsock: fix latent data race in test

josharian:
logBufWriter had no serialization.
It just so happens that none of its users currently ever log concurrently.
Make it safe for concurrent use.


josharian:
Ping @bradfitz. Should be an easy rubber stamp?

bradfitz (approved):

bradfitz:
> Ping @bradfitz. Should be an easy rubber stamp?

Inbox disaster, as usual.

PR #2375, opened 2021-07-09
wgengine/magicsock: make portmapping async

bradfitz:
Making portmapping creation async so we can afford to make it slower (for UPnP) without stalling other parts of magicsock.

### DIFF
--- a/net/portmapper/portmapper.go
+++ b/net/portmapper/portmapper.go
@@ -44,8 +44,14 @@ const trustServiceStillAvailableDuration = 10 * time.Minute
 type Client struct {
 	logf         logger.Logf
 	ipAndGateway func() (gw, ip netaddr.IP, ok bool)
+	onChange     func() // or nil
 
 	mu sync.Mutex // guards following, and all fields thereof
+
+	// runningCreate is whether we're currently working on creating
+	// a port mapping (whether GetCachedMappingOrStartCreatingOne kicked
+	// off a createMapping goroutine).
+	runningCreate bool
 
 	lastMyIP netaddr.IP
 	lastGW   netaddr.IP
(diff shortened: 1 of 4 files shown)


net/portmapper/portmapper.go
```diff
@@ -241,12 +251,52 @@ var (
...
+// If the background goroutine ends up creating one, the onChange hook registered with the
+// NewClient constructor (if any) will fire.
+func (c *Client) GetCachedMappingOrStartCreatingOne() (external netaddr.IPPort, ok bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// Do we have an existing mapping that's valid?
+	now := time.Now()
+	if m := c.pmpMapping; m != nil {
+		if now.Before(m.useUntil) {
```
JulianKnodt, line 265: I think the buffer time of 30 seconds is already added into useUntil here, but just wanted to double check if that's true.

net/portmapper/portmapper.go
```diff
@@ -241,12 +251,52 @@ var (
...
+
+	defer func() {
+		c.mu.Lock()
+		defer c.mu.Unlock()
+		c.runningCreate = false
+	}()
+
+	if _, err := c.createOrGetMapping(ctx); err == nil && c.onChange != nil {
+		go c.onChange()
+	}
```
JulianKnodt, line 291: Do we want to handle error cases differently? If there are no portmapping services will this still get run/continually get run? The previous behavior was to keep running so doesn't matter a ton but just wanted to mention it

We probably also want to log the error here if it's not a `NoMappingError`

JulianKnodt (approved):

net/portmapper/portmapper.go
bradfitz: oh, yeah, this is a bug. I'm not renewing the mapping in the period between useUntil (which is until half the expiry)

net/portmapper/portmapper.go
bradfitz: Done.

bradfitz:
PTAL

PR #2376, opened 2021-07-09
tstest/integration/testcontrol: ipv6 support

Xe:
And a bunch of IPv6 tests!

### DIFF
--- a/tstest/integration/testcontrol/testcontrol.go
+++ b/tstest/integration/testcontrol/testcontrol.go
@@ -28,6 +28,7 @@ import (
 	"github.com/klauspost/compress/zstd"
 	"golang.org/x/crypto/nacl/box"
 	"inet.af/netaddr"
+	"tailscale.com/net/tsaddr"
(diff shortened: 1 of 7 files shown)


tstest/integration/testcontrol/testcontrol.go
```diff
@@ -504,6 +509,16 @@ func (s *Server) UpdateNode(n *tailcfg.Node) (peersToUpdate []tailcfg.NodeID) {
 	return peersToUpdate
 }
 
+func stringInSlice(haystack []string, needle string) bool {
```
Xe, line 512: TODO(Xe): remove

crawshaw (commented):
For future reference, please move code around in one PR, then modify code in a subsequent PR. (Or at the minimum, different commits within one PR.) As it is, what could have been a 10 minute code review will take me much longer than that to read. I'll take a look this afternoon.

Xe:
I can refactor out the framework changes from the test changes

PR #2378, opened 2021-07-09
tailcfg: break DERPNode.DERPTestPort into DERPPort & InsecureForTests

bradfitz:
The DERPTestPort int meant two things before: which port to use, and
whether to disable TLS verification. Users would like to set the port
without disabling TLS, so break it into two options.

Updates #1264


tailcfg/derpmap.go
```diff
@@ -130,10 +130,15 @@ type DERPNode struct {
...
-	DERPTestPort int `json:",omitempty"`
+	// DERPPort optionally provides an alternate TLS port number
+	// for the DERP HTTPS server.
+	//
+	// If zero, 443 is used.
+	DERPPort int `json:",omitempty"`
+
+	// InsecureForTests is used by unit tests to disable TLS verification.
+	// It should not be set by users.
+	InsecureForTests bool `json:",omitempty"`
```
JulianKnodt, line 141: maybe I'm misunderstanding something, but why does this need to be exported at all if it shouldn't be set by users? Would it be more convenient to have it unexported and have a getter?

JulianKnodt (approved):

tailcfg/derpmap.go
bradfitz: We want to set it in tests easily.

It's only recently that users could use this. We can enforce it in policy.Parse that users can't set it.

Issue #1912, opened 2021-05-11
wgengine/wgcfg: optimize FromUAPI


josharian:
Lots of allocations, generally heavy-weight, and we use it a lot.

DentonGentry:
Yesterday in https://github.com/tailscale/tailscale/pull/5451 we said,
_"The idea at the time was that UAPI was slow but we could probably make it faster, which we did, over a number of rounds. But I think we made it about as fast as we can."_

Closing this as completed.

PR #2405, opened 2021-07-13
wgcfg: add benchmark for FromUAPI

JulianKnodt:
Adds a benchmark for FromUAPI in wgcfg.
Updates #1912.

I was also going to see if there were any easy perf gains here, but it doesn't appear that there's anything I can easily change. I should note there is one small change for `"replace_allowed_ips"` which wasn't handled before, and I'm not sure what indicates.


wgengine/wgcfg/parser_test.go
```diff
@@ -53,3 +58,47 @@ func TestParseEndpoint(t *testing.T) {
...
+
+	buf := new(bytes.Buffer)
+	w := bufio.NewWriter(buf)
+	if err := cfg1.ToUAPI(w, &Config{}); err != nil {
+		b.Fatal(err)
+	}
+	w.Flush()
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		_, err := FromUAPI(buf)
```
bradfitz, line 99: you're not benchmarking what you think you are.

You're measuring an empty parser:

https://play.golang.org/p/8hxjn_F5Jaj

Screen Shot 2021-07-12 at 10 10 53 PM

JulianKnodt: I'm pretty sure I have an actual cfg, I printed out the value

bradfitz: edit3: 0% sure :)

To convince yourself, add:

```go
        b.ReportAllocs()
        for i := 0; i < b.N; i++ {
                if buf.Len() == 0 {
                        b.Fatal("empty")
                }
                _, err := FromUAPI(buf)
                if err != nil {
                        b.Errorf("failed from UAPI: %v", err)
                }
        }
```

JulianKnodt:
Ah ok now there are easy perf gains:
```
name        old time/op    new time/op    delta
FromUAPI-8    4.51µs ± 2%    3.47µs ± 8%  -23.15%  (p=0.008 n=5+5)

name        old alloc/op   new alloc/op   delta
FromUAPI-8    7.18kB ± 0%    6.64kB ± 0%   -7.57%  (p=0.008 n=5+5)

name        old allocs/op  new allocs/op  delta
FromUAPI-8      53.0 ± 0%      36.0 ± 0%  -32.08%  (p=0.008 n=5+5)
```

wgengine/wgcfg/parser_test.go
```diff
@@ -53,3 +58,49 @@ func TestParseEndpoint(t *testing.T) {
...
+	}
+	w.Flush()
+	r := bytes.NewReader(buf.Bytes())
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		_, err := FromUAPI(r)
+		if err != nil {
+			b.Errorf("failed from UAPI: %v", err)
+		}
+		r.Seek(0, 0)
```
bradfitz, line 104: `0, io.SeekStart` for readability. and maybe put it before line 100 so it's more obvious and close to its use

wgengine/wgcfg/parser.go
```diff
@@ -83,12 +91,10 @@ func FromUAPI(r io.Reader) (*Config, error) {
 		if line == "" {
 			continue
 		}
-		parts := strings.Split(line, "=")
-		if len(parts) != 2 {
-			return nil, fmt.Errorf("failed to parse line %q, found %d =-separated parts, want 2", line, len(parts))
+		key, value, ok := stringsCut(line, "=")
```
bradfitz, line 94: better is to not work with strings at all. The scanner.Text() on line 90 allocates on each line.

But then line 98 goes back to working in []byte.

Ideally everything would work in terms of mem.RO (https://pkg.go.dev/go4.org/mem#RO)

bradfitz (approved):

wgengine/wgcfg/parser.go
JulianKnodt: I'll try this out before merging

JulianKnodt:
Latest change has a lot of perf improvements
```
name        old time/op    new time/op    delta
FromUAPI-8    4.51µs ± 2%    2.84µs ± 1%  -37.08%  (p=0.008 n=5+5)

name        old alloc/op   new alloc/op   delta
FromUAPI-8    7.18kB ± 0%    5.54kB ± 0%  -22.94%  (p=0.008 n=5+5)

name        old allocs/op  new allocs/op  delta
FromUAPI-8      53.0 ± 0%      13.0 ± 0%  -75.47%  (p=0.008 n=5+5)
```
Probably worth another review since there's a lot of changes, but if you don't think it's necessary it's good to merge now

wgengine/wgcfg/parser.go
```diff
@@ -56,17 +57,13 @@ func parseEndpoint(s string) (host string, port uint16, err error) {
...
 }
 
-func parseKeyHex(s string) (*wgkey.Key, error) {
-	k, err := hex.DecodeString(s)
-	if err != nil {
-		return nil, &ParseError{"Invalid key: " + err.Error(), s}
-	}
-	if len(k) != wgkey.Size {
-		return nil, &ParseError{"Keys must decode to exactly 32 bytes", s}
+// stringsCut is strings.Cut from proposed https://github.com/golang/go/issues/46336.
```
josharian: It seems to me that we should add a Cut method to mem.RO.

bradfitz, line 60: memROCut is ...

wgengine/wgcfg/parser.go
```diff
@@ -79,25 +76,23 @@ func FromUAPI(r io.Reader) (*Config, error) {
...
+		line := mem.B(scanner.Bytes())
+		if line.Len() == 0 {
 			continue
 		}
-		parts := strings.Split(line, "=")
-		if len(parts) != 2 {
-			return nil, fmt.Errorf("failed to parse line %q, found %d =-separated parts, want 2", line, len(parts))
+		key, value, ok := memROCut(line, '=')
+		if !ok {
+			return nil, fmt.Errorf("failed to cut line %v on =", line)
```
josharian, line 86: Does mem.RO print nicely? (I don't know offhand.) Do we want `%q`?

wgengine/wgcfg/parser.go
```diff
@@ -122,62 +117,70 @@ func FromUAPI(r io.Reader) (*Config, error) {
...
+	} else {
 		return fmt.Errorf("unexpected IpcGetOperation key: %v", key)
 	}
 	return nil
 }
 
-func (cfg *Config) handlePublicKeyLine(value string) (*Peer, error) {
-	k, err := parseKeyHex(value)
-	if err != nil {
+func (cfg *Config) handlePublicKeyLine(value mem.RO, valueBytes []byte) (*Peer, error) {
```
bradfitz, line 145: value mem.RO is never used?

wgengine/wgcfg/parser.go
```diff
@@ -122,62 +117,70 @@ func FromUAPI(r io.Reader) (*Config, error) {
...
 }
 
-func (cfg *Config) handlePeerLine(peer *Peer, key, value string, valueBytes []byte) error {
-	switch key {
-	case "endpoint":
-		err := json.Unmarshal(valueBytes, &peer.Endpoints)
-		if err != nil {
+func (cfg *Config) handlePeerLine(peer *Peer, key, value mem.RO, valueBytes []byte) error {
+	if key.EqualString("endpoint") {
+		if err := json.NewDecoder(mem.NewReader(value)).Decode(&peer.Endpoints); err != nil {
```
bradfitz, line 156: I think making that NewReader is worse than just using valueBytes like before

JulianKnodt: I experimented with both and didn't notice too much of a difference, will switch to using valueBytes

bradfitz: Because your benchmark never exercises this line of code.

Your input string is:

```
private_key=a87242cf17a8acc459b444e1fac8bb5e848784b97e41adf3ad00d34f43327350
public_key=f7b7f5c3aa2b5a9ec7363dc1753d5872fae63c4ec9da3591c56334d64b2d7a1b
protocol_version=1
replace_allowed_ips=true
allowed_ip=10.0.0.1/32
public_key=f7b7f5c3aa2b5a9ec7363dc1753d5872fae63c4ec9da3591c56334d64b2d7a1b
protocol_version=1
replace_allowed_ips=true
allowed_ip=10.0.0.1/32
public_key=f7b7f5c3aa2b5a9ec7363dc1753d5872fae63c4ec9da3591c56334d64b2d7a1b
protocol_version=1
replace_allowed_ips=true
allowed_ip=10.0.0.1/32
public_key=f7b7f5c3aa2b5a9ec7363dc1753d5872fae63c4ec9da3591c56334d64b2d7a1b
protocol_version=1
replace_allowed_ips=true
allowed_ip=10.0.0.1/32
```

It doesn't contain `endpoint`.

JulianKnodt: Added and now the benchmarks reflect that

wgengine/wgcfg/parser.go
```diff
@@ -122,62 +118,70 @@ func FromUAPI(r io.Reader) (*Config, error) {
...
+		return &ParseError{"Invalid key: " + err.Error(), string(s)}
+	}
+	if n != wgkey.Size {
+		return &ParseError{"Keys must decode to exactly 32 bytes", string(s)}
+	}
+	return nil
+}
+
+func (cfg *Config) handleDeviceLine(key, value mem.RO, valueBytes []byte) error {
+	if key.EqualString("private_key") {
```
josharian, line 133: This is much less readable than the switch statement. Is this performance critical?

josharian (commented):
I keep reviewing and the code keeps changing under me. Please let me know when it's ready for another look.

wgengine/wgcfg/parser.go
JulianKnodt: mem.RO doesn't handle `%q`, but I can convert it into a `string` there cause then we no longer care about perf

wgengine/wgcfg/parser.go
JulianKnodt: Makes sense, hopefully that won't block this PR tho

wgengine/wgcfg/parser.go
JulianKnodt: It's not possible to get a string here without allocating, so for memory it's pretty important. I can convert it into a switch statement, but it's not really better

josharian: Right, but do these allocations matter in the bigger picture? If not, readable code is better.

JulianKnodt: Allocating for this using a string copy does seem to be on the hot path and is actually a significant portion of the runtime, so I'm going to leave it as this.

I do understand that this is a lot less readable, but I think it's worth the cost.

josharian: > Allocating for this using a string copy does seem to be on the hot path and is actually a significant portion of the runtime, so I'm going to leave it as this.

A significant portion of the runtime for what? For this benchmark specifically, or for something measurable in the client code? To put it another way, why does it matter that this benchmark be fast?

I'm all for making things fast in general. But when there's a significant readability regression, the need for the performance optimizations to matter gets more important.

As an example, the deephash optimizations (which had an awful bug--the dangers of optimization!) were important to keeping the hello server stable under load, and enabled deephash to be used in control.

JulianKnodt: Mmm this is fair, and I probably don't have as much context on where `FromUAPI` is used as I should to know whether the benchmark I wrote is representative of the common use case in prod. That being said, it appears no matter what path the code takes through, it will hit one of the two switch statements which were changed into `EqualString`, and so however the API is used, as long as it's not empty (in which case I think the point of optimizing would've been moot in general), it will strictly increase the performance.

I hear you on the readability issue, but I personally lean towards always optimizing speed so I am a bit biased. Whether this optimization is _necessary_ is a good question.

To measure impact, would it be possible to look at pprof outputs from usage?

josharian:
I'm not sure this should have been merged. We have gone from fairly clear, and very similar to upstream code, to code that is a lot less nice.

Aside from microbenchmark improvements, what do we gain from this? I don't recall seeing these show up in pprof anywhere.

JulianKnodt:
If you feel strongly about it, then I can revert it, also since I realize this is around the release timing. That being said, I think all of the changes in this PR are strictly positive and adhere closely to the original structure but with some type modifications, and the change in readability is worth having it.

wgengine/wgcfg/parser.go
josharian: > To measure impact, would it be possible to look at pprof outputs from usage?

I have spent a fair amount of time looking at pprof output from the client. I don't think that this code is at all important to performance right now, at least right now.

> I personally lean towards always optimizing speed so I am a bit biased.

I share that bias. But where performance really doesn't matter, and when that performance actively conflicts with readability, I prefer readability.

bradfitz:
Before reverting, let's look at hello's profiles.

Sorry, I'd assumed this was motivated by something in particular.

bradfitz:
Looking at pprof --alloc_space ...

```
      12MB 0.00021%   100%  1960.54MB 0.035%  tailscale.com/wgengine/wgcfg.FromUAPI
```

And focusing on UAPI:

Screen Shot 2021-07-13 at 2 41 13 PM

josharian:
Wow, hello allocates a lot, if 1.91GB is only 0.035% of its allocations.

Anyway, 0.035% doesn't seem like enough to be worth optimizing for.

Of course, as a percentage of allocs, that'll go up once the package net optimizations to avoid per-packet allocations are in. But I still suspect it'll still be fairly irrelevant overall; I suspect that at that point, we won't be allocating enough for the allocations to matter. Happy to be proved wrong, though.

PR #2412, opened 2021-07-13
derp: allow self node when verifying clients

### LINKED ISSUE #2408: Lots of `rejected` logs in DERP server with `verify-clients` on
I have a server that runs `tailscaled` as a node and `derper` as a DERP server.

When I have `-verify-clients` option turned on, a lot of these lines being printed in the system log:

```
derp: 127.0.0.1:35730: client xxxxxxxxxxx rejected: client [11 22 33 44 55 66 ] not in set of peers
```

`xxxxxx` is the server's public key (runs `tailscaled` and `derper`), as confirmed with `tailscale status --json`.

Also another question, how can I check to see if the DERP server is running properly?

maddie:
Fixes #2408

### DIFF
--- a/derp/derp_server.go
+++ b/derp/derp_server.go
@@ -797,6 +797,9 @@ func (s *Server) verifyClient(clientKey key.Public, info *clientInfo) error {
 	if err != nil {
 		return fmt.Errorf("failed to query local tailscaled status: %w", err)
 	}
+	if clientKey == status.Self.PublicKey {
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


bradfitz:
@maddie, we require the "Signed-Off-By" line in git commits.

See https://developercertificate.org/.

You'll want to run `git commit --amend -s` to add it to the commit, then GitHub's DCO check will be happy.

maddie:
@bradfitz Done, thanks for your reminder!

PR #2415, opened 2021-07-13
cmd/tailscale: log web listen addr

syscll:
When running `tailscale web` I saw no output and was unsure which address the server was listening on, so I added a log to display this.

Signed-off-by: Dan Bond <danbond@protonmail.com>

### DIFF
--- a/cmd/tailscale/cli/web.go
+++ b/cmd/tailscale/cli/web.go
@@ -15,6 +15,7 @@ import (
 	"html/template"
 	"io/ioutil"
 	"log"
+	"net"
 	"net/http"
 	"net/http/cgi"
 	"net/url"
@@ -94,8 +95,18 @@ func runWeb(ctx context.Context, args []string) error {
 		}
 		return nil
 	}
-	log.Printf("web server running on: %s\n", webArgs.listen)
+
+	log.Printf("web server running on: %s", urlOfListenAddr(webArgs.listen))
 	return http.ListenAndServe(webArgs.listen, http.HandlerFunc(webHandler))
+}
+
+// urlOfListenAddr parses a given listen address into a formatted URL
+func urlOfListenAddr(addr string) string {
+	host, port, _ := net.SplitHostPort(addr)
+	if host == "" {
+		host = "127.0.0.1"
+	}
+	return fmt.Sprintf("http://%s", net.JoinHostPort(host, port))
 }
 
 // authorize returns the name of the user accessing the web UI after verifying
--- a/cmd/tailscale/cli/web.go
+++ b/cmd/tailscale/cli/web.go
@@ -94,6 +94,7 @@ func runWeb(ctx context.Context, args []string) error {
 		}
 		return nil
 	}
+	log.Printf("web server running on: %s\n", webArgs.listen)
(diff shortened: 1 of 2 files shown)


cmd/tailscale/cli/web.go
```diff
@@ -94,6 +94,7 @@ func runWeb(ctx context.Context, args []string) error {
 		}
 		return nil
 	}
+	log.Printf("web server running on: %s\n", webArgs.listen)
```
bradfitz, line 97: don't need `\n` with the `log` package. also, you cool with this printing e.g. "web server listening on :8080" and not a URL or IP?

Xe, line 97: Assuming this compiles, try this?
```suggestion
	host, port, _ := net.SplitHostPort(webArgs.listen)
	if host == "" {
		host = "127.0.0.1"
	}
	log.Printf("web server running on: http://%s", net.JoinHostPort(host, port))
```

bradfitz: or ideally pulled out into a `func urlOfListenAddr`

syscll: I was ok just printing the port, but I think it does make more sense to also print the IP as per @Xe's suggestion.

syscll:
Extracted into it's own function and added some simple tests.

cmd/tailscale/cli/web_test.go
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package cli
+
+import "testing"
+
+func TestUrlOfListenAddr(t *testing.T) {
+	t.Parallel()
```
bradfitz, line 10: overkill for something so short

cmd/tailscale/cli/web_test.go
```diff
@@ -0,0 +1,42 @@
...
+			addr:     "127.0.0.2:8088",
+			expected: "http://127.0.0.2:8088",
+		},
+		"TestIPv6": {
+			addr:     "[::1]:8088",
+			expected: "http://[::1]:8088",
+		},
+	}
+
+	for name, test := range testTable {
```
bradfitz, line 34: the random map enumeration is distracting in tests.

Instead:

```go
     tests := []struct{
           name string
           in, want string
      }{
    ....  
     }
```

(note: `want` instead of `expected`)

Then:

```go
     for _, tt := range tests {
          t.Run(tt.name, ....
```

cmd/tailscale/cli/web_test.go
```diff
@@ -0,0 +1,42 @@
...
+		t.Run(name, func(t *testing.T) {
+			url := urlOfListenAddr(test.addr)
+			if url != test.expected {
+				t.Errorf("expected url: '%s', got: '%s'", test.expected, url)
```
bradfitz, line 38: %q instead of '%s'

syscll:
Appreciate the feedback. Have addressed your comments :)

cmd/tailscale/cli/web_test.go
syscll: TIL

PR #2416, opened 2021-07-13
tstest/integration/vms: delete a log file i was using to debug the tests

Xe:


tstest/integration/vms/.gitignore
```diff
@@ -0,0 +1 @@
+test.log
```
bradfitz, line 1: I'd rather not have this here. You can define this globally for yourself, or just keep your scratch files out of the tree.

Xe: fair. i edited the gitignore to ignore itself locally

crawshaw (approved):

PR #2417, opened 2021-07-13
ipn/ipnlocal: save prefs to disk on UpdatePrefs

### LINKED ISSUE #2321: After system crash, tailscale came back up with an old or truncated --advertise-routes config
We’re running tailscale 1.10.0 on a Debian 10.9 system that is advertising a total of 15 routes (/32s on the loopback interface) using `--advertise-routes`.

This system suffered a power outage and came back with four of the most recently added routes missing. I noticed these routes where down after the outage and then saw a notice on the Tailscale web UI that the host had stopped advertising the routes. Sure enough, the `tailscale up` output of the existing config omitted these four routes.

I’m not sure whether the config reverted to a (several weeks) old version that had fewer routes, or if the list of routes was truncated to some maximum size. Regardless, this seems to be an issue with the state persistence mechanism.

bradfitz:
Regression from 6d10655dc3887f1a161015514a8555c175802b4d, which added
UpdatePrefs but didn't write it out to disk.

I'd planned on adding tests to state_test.go which is why I'd earlier
added 46896a93118d0eecbacb4255da2df0349da9b409 to prepare for making
such persistence tests easier to write, but turns out state_test.go
didn't even test UpdatePrefs, so I'm staying out of there.
Instead, this is tested using integration tests.

Fixes #2321

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -757,6 +757,12 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 		newPrefs := opts.UpdatePrefs
 		newPrefs.Persist = b.prefs.Persist
 		b.prefs = newPrefs
+
+		if opts.StateKey != "" {
+			if err := b.store.WriteState(opts.StateKey, b.prefs.ToBytes()); err != nil {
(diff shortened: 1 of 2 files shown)


crawshaw (approved):
Nice test!

PR #2422, opened 2021-07-13
wgengine/magicsock: add debug envvar to disable UDP

josharian:
This would have been useful during debugging yesterday.

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -83,6 +83,8 @@ var (
 	// on mobile devices, lowers the shutdown interval, and logs more
 	// verbosely about idle measurements.
 	debugReSTUNStopOnIdle, _ = strconv.ParseBool(os.Getenv("TS_DEBUG_RESTUN_STOP_ON_IDLE"))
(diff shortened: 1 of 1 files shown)


wgengine/magicsock/magicsock.go
```diff
@@ -83,6 +83,8 @@ var (
 	// on mobile devices, lowers the shutdown interval, and logs more
 	// verbosely about idle measurements.
 	debugReSTUNStopOnIdle, _ = strconv.ParseBool(os.Getenv("TS_DEBUG_RESTUN_STOP_ON_IDLE"))
+	// debugDisableUDP disables the use of UDP, forcing all peer communication over DERP.
+	debugDisableUDP, _ = strconv.ParseBool(os.Getenv("TS_DEBUG_DISABLE_UDP"))
```
bradfitz, line 87: This makes it sound like it's all UDP but it's just magicsock as implementation. There's also DNS and netstack forwarding. Rename environment var?

josharian: PTAL

bradfitz (approved):

PR #2423, opened 2021-07-14
optimize throughput a bit

josharian:
- net/tstun: buffer outbound channel
- net/tstun: eliminate per-packet time.Now calls

### DIFF
--- a/net/tstun/wrap.go
+++ b/net/tstun/wrap.go
@@ -153,7 +153,7 @@ func Wrap(logf logger.Logf, tdev tun.Device) *Wrapper {
 		// a goroutine should not block when setting it, even with no listeners.
 		bufferConsumed: make(chan struct{}, 1),
 		closed:         make(chan struct{}),
-		outbound:       make(chan tunReadResult),
+		outbound:       make(chan tunReadResult, 1),
 		eventsUpDown:   make(chan tun.Event),
 		eventsOther:    make(chan tun.Event),
 		// TODO(dmytro): (highly rate-limited) hexdumps should happen on unknown packets.
--- a/net/tstun/wrap.go
+++ b/net/tstun/wrap.go
@@ -18,6 +18,7 @@ import (
 	"golang.zx2c4.com/wireguard/tun"
 	"inet.af/netaddr"
 	"tailscale.com/net/packet"
+	"tailscale.com/syncs"
 	"tailscale.com/types/ipproto"
 	"tailscale.com/types/logger"
 	"tailscale.com/wgengine/filter"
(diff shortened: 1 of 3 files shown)


josharian:
Review per-commit.

These are "incremental benefit, medium risk", so we may want to delay merging until next dev cycle. Open to input here.

bradfitz (commented):
Sure this is real and not your VM sucking at time?

crawshaw (approved):
This looks right. Do you want to describe your throughput test? I could try it out tomorrow on a couple different machines.

net/tstun/wrap.go
```diff
@@ -64,7 +65,8 @@ type Wrapper struct {
 
 	closeOnce sync.Once
 
-	lastActivityAtomic int64 // unix seconds of last send or receive
+	idleDuration syncs.AtomicInt64 // approximate number of seconds since last activity
```
bradfitz, line 68: call it `idleSeconds` instead of duration? a duration is also an int64 and suggests nanoseconds kinda.

net/tstun/wrap.go
```diff
@@ -162,6 +164,8 @@ func Wrap(logf logger.Logf, tdev tun.Device) *Wrapper {
 
 	go tun.poll()
 	go tun.pumpEvents()
+	tun.idleDuration.Set(time.Now().Unix())
+	go tun.trackIdleDuration()
```
bradfitz, line 168: not a big fan of extra goroutines sticking around forever, especially on iOS

bradfitz (commented):
I remain very skeptical.

crawshaw:
Alternative: let's ditch time.Now, rip the vDSO code out of the runtime, put it in an xtime package, and get ourselves the right clock. On linux that means `clock_gettime(CLOCK_MONOTONIC_COARSE)`, which runs in ~7ns! https://news.ycombinator.com/item?id=18529546.

macOS 10.12 adds `mach_continuous_time` which looks like it gets us something similar, that does the right thing when the device sleeps.

josharian:
> Sure this is real and not your VM sucking at time?

On my mac, time.BenchmarkTime reports ~45ns/op. On the linux VM, ~63ns/op. So it's marginally slower, but not by much. And I'm sure people run Tailscale on VMs.

> not a big fan of extra goroutines sticking around forever, especially on iOS

Will fix.

> Alternative: let's ditch time.Now

That sounds great, but let me see whether I can put together something simpler here.

josharian:
> not a big fan of extra goroutines sticking around forever, especially on iOS

I have a fix for this implemented and documented, but it's not exactly simple. I'm going to try the "roll our own cheap time call" approach instead...

josharian:
Abandoning; prefer #2486 instead.

PR #2427, opened 2021-07-14
tstest/integration/vms: deflake the tailscale_status test

Xe:
This test used to try to run this only once, but this variant of the
test attempts to run `tailscale status` up to 6 times in a loop with
exponential backoff.

This fixes the flakiness found in previous instances of this test.

### DIFF
--- a/tstest/integration/vms/vms_test.go
+++ b/tstest/integration/vms/vms_test.go
@@ -449,11 +449,32 @@ func (h Harness) testDistro(t *testing.T, d Distro, ipm ipMapping) {
 	})
 
 	t.Run("tailscale status", func(t *testing.T) {
-		runTestCommands(t, timeout, cli, []expect.Batcher{
-			&expect.BSnd{S: "sleep 5 && tailscale status\n"},
-			&expect.BExp{R: `100.64.0.1`},
-			&expect.BExp{R: `(\#)`},
-		})
+		dur := 100 * time.Millisecond
+		var outp []byte
+		var err error
+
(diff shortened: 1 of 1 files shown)


tstest/integration/vms/vms_test.go
```diff
@@ -449,11 +449,28 @@ func (h Harness) testDistro(t *testing.T, d Distro, ipm ipMapping) {
...
-			&expect.BSnd{S: "sleep 5 && tailscale status\n"},
-			&expect.BExp{R: `100.64.0.1`},
-			&expect.BExp{R: `(\#)`},
-		})
+		dur := 100 * time.Millisecond
+		var outp []byte
+		var err error
+
+		for count := 0; count < 10; count++ {
+			time.Sleep(dur)
```
crawshaw, line 457: put the sleep at the end of the loop so in the happy path no sleep is required

tstest/integration/vms/vms_test.go
```diff
@@ -449,11 +449,28 @@ func (h Harness) testDistro(t *testing.T, d Distro, ipm ipMapping) {
...
-		runTestCommands(t, timeout, cli, []expect.Batcher{
-			&expect.BSnd{S: "sleep 5 && tailscale status\n"},
-			&expect.BExp{R: `100.64.0.1`},
-			&expect.BExp{R: `(\#)`},
-		})
+		dur := 100 * time.Millisecond
+		var outp []byte
+		var err error
+
+		for count := 0; count < 10; count++ {
```
crawshaw, line 456: document what this loop is for. what error is reported that requires a retry?

crawshaw (approved):

PR #2429, opened 2021-07-14
cmd/derper: add default -c arguments

JulianKnodt:
This adds some convenient defaults for -c, so that user-provided DERPs require less command line
flags.

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -58,7 +58,12 @@ func loadConfig() config {
 		return config{PrivateKey: mustNewKey()}
 	}
 	if *configPath == "" {
-		log.Fatalf("derper: -c <config path> not specified")
+		if os.Getuid() == 0 {
+			*configPath = "/var/lib/derper/derper.key"
+		} else {
+			*configPath = "derper.key"
+		}
+		log.Printf("no config path specified; using %s", *configPath)
 	}
 	b, err := ioutil.ReadFile(*configPath)
 	switch {


cmd/derper/derper.go
```diff
@@ -58,7 +58,11 @@ func loadConfig() config {
 		return config{PrivateKey: mustNewKey()}
 	}
 	if *configPath == "" {
-		log.Fatalf("derper: -c <config path> not specified")
+		if os.Geteuid() == 0 {
+			*configPath = "/var/lib/derper/derper.key"
+		} else {
+			*configPath = "derper.key"
+		}
 	}
```
bradfitz, line 66:
```go
log.Printf("no config path specified; using %s", *configPath)
```

cmd/derper/derper.go
```diff
@@ -58,7 +58,11 @@ func loadConfig() config {
 		return config{PrivateKey: mustNewKey()}
 	}
 	if *configPath == "" {
-		log.Fatalf("derper: -c <config path> not specified")
+		if os.Geteuid() == 0 {
```
bradfitz, line 61: does the `e` part matter? I think we usually just test `Getuid`.

bradfitz (approved):

cmd/derper/derper.go
JulianKnodt: not in particular, just saw it on SO so used the e

PR #2433, opened 2021-07-15
tstest/integration/vms: attempt to send more UDP packets

Xe:
Apparently this test was flaking because I critically misunderstood how
the kernel buffers UDP packets for senders. I'm trying to send more UDP
packets and will see if that helps.

### DIFF
--- a/tstest/integration/vms/vms_test.go
+++ b/tstest/integration/vms/vms_test.go
@@ -602,25 +602,40 @@ func (h Harness) testDistro(t *testing.T, d Distro, ipm ipMapping) {
 		}
 		defer ln.Close()
 
-		sess, err := cli.NewSession()
-		if err != nil {
-			t.Fatalf("can't open session: %v", err)
-		}
-		defer sess.Close()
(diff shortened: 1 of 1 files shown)


tstest/integration/vms/vms_test.go
```diff
@@ -602,25 +602,40 @@ func (h Harness) testDistro(t *testing.T, d Distro, ipm ipMapping) {
...
-		}
-		defer sess.Close()
+		ctx, cancel := context.WithCancel(context.Background())
+		t.Cleanup(cancel)
 
-		sess.Stdin = strings.NewReader("hi")
-		sess.Stdout = logger.FuncWriter(t.Logf)
-		sess.Stderr = logger.FuncWriter(t.Logf)
+		go func() {
+			for i := 0; i < 10; i++ {
```
crawshaw, line 609: I'm not sure it's worth putting a limit on the send here. You may as well make this `for {` and keep sending UDP packets forever, until you see one or the test times out.

crawshaw (approved):

Issue #2056, opened 2021-06-06
Configure DNS using DNS-over-TLS or DNS-over-HTTPS


ItalyPaleAle:
On iOS, you can only use one VPN app at a time. The problem is that apps like 1.1.1.1 that offer secure DNS-over-HTTPS are implemented as VPN apps, so we always have to choose whether to use Tailscale (and access our private nodes) or 1.1.1.1 (and prevent our ISP from spying on us via DNS then selling that data to advertisers).

Would it be possible in the Tailscale iOS app to be able to configure what DNS server is used as fallback when split DNS is enabled? Ideally with DoH/DoT support too :) 

Note: this request is fairly different from https://github.com/tailscale/tailscale/issues/74 . The latter is about global DNS, in which the Tailscale DNS server makes upstream requests using DoT/DoH. In this case, the request is for the client itself to be configured to make split DNS requests using DoT/DoH when split DNS is being used.

<img src="https://frontapp.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_39zrl)

DentonGentry:
Seems feasible. Looking at client support:
+ MacOS Big Sur can [configure DNS-over-TLS and DNS-over-HTTPS from a Network Extension](https://developer.apple.com/documentation/networkextension/dns_settings). I don't see a way to do so from a System Extension, which we're trying to use for a beta channel version of the app, nor from [Tailscaled on MacOS](https://github.com/tailscale/tailscale/wiki/Tailscaled-on-macOS).
+ Recent iOS versions are similar, DoT and DoH can be configured using the same APIs as MacOS
+ Linux systemd can support DNS-over-TLS in opportunistic mode where it just tries TLS to connect to an upstream DNS server. It looks like it will do this based on settings in /etc/systemd/resolved.conf
+ Windows implements opportunistic DNS-over-HTTPS if HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Dnscache\Parameters has a EnableAutoDoh key set to 2.
+ Android implements DNS-over-TLS systemwide, though I don't see a way to turn it on from the Tailscale app. We also need to come up with some resolution for #915, that the VPN framework and Private DNS appear to be incompatible.

[...]

Issue #915, opened 2020-11-11
Tailscale breaks DNS-over-TLS on Android


michaelclarkcuadrado:
**Describe the bug**
If a custom DNS-over-TLS ("Private DNS") server is configured, enabling tailscale will stop all DNS requests and show a "couldn't connect" in the android settings app.

**To Reproduce**
1. Disable the tailscale VPN. 
2. In Android setting's "WiFi and network" section, set the private dns to a specific server, not "automatic". I tested with values of both "dns.quad9.net" and "dns.nextdns.io". 
3. Observe that dns works as expected.
4. Activate tailscale.
5. Observe that dns requests fail, and that private dns' settings are labelled "couldn't connect"

Note that the "automatic" setting, for Google's DNS-over-TLS server, works correctly.

**Expected behavior**
DNS-over-TLS servers should be accessible even when running tailscale.

**Screenshots**
![Screenshot_20201111-121933](https://user-images.githubusercontent.com/15159826/98860683-e84b5c00-2418-11eb-9e0c-01ff545a6bf0.jpg)

**Version information:**
 - Device: OnePlus Phone
 - OS: Android
 - OS version: 10.
 - Tailscale version: version 1.2.2-t76c2982d8-g39dfd8495

**Additional context**

bradfitz:
Can you also share your Tailscale IP address?

michaelclarkcuadrado:
The tailscale IP of the device is 100.73.211.12

normanr:
I see the same thing when trying to use dns.google on 100.66.241.34.

If I use socat to ssl listen on port 853 (and tcp connect to port 53) on the dns server I configured in Tailscale, then Automatic mode seems to use it. Manual mode works too as long as I change my preferred Private DNS server name to match the certificate I'm using.

Just a plain forwarding of port 853 to the upstream dns server also works in Automatic mode, but not with a provided hostname. Could it be because Tailscale forces all DNS traffic to 100.100.100.100 which isn't listening on port 853?

normanr:
note: I don't think that "Automatic" mode uses Google's DNS-over-TLS server, but rather it checks if the currently configured DNS server supports DNS-over-TLS and uses it if it's available.

If you want to use Google's DNS-over-TLS server then I think you have to explicitly enter dns.google as your Private DNS provider.

DentonGentry:
[...]

Issue #74, opened 2020-02-18
Relax IP validation in admin screens


martinbaillie:
It would be nice if the admin console allowed for storing more than just an IP address for DNS, e.g. DNS over TLS: 1.1.1.1@853

<img width="377" alt="Screen Shot 2020-02-18 at 9 26 56 pm" src="https://user-images.githubusercontent.com/613740/74727693-a9c3af00-5295-11ea-9a40-243772f0e3f7.png">

danderson:
Right now we implement this DNS policy by reconfiguring the local resolver on Windows, Mac, Unixen, and iOS. Unless they all support DoT/DoH natively, we can't really make this change.

Likely the fix for this is going to be running our own DNS proxy in tailscaled, and doing "fancy" lookups there instead of delegating that to the OS. So I think this is a fine idea, but there's a bunch of stuff we need to do first.

danderson:
Aspirationally adding to Tailscale 1.8 milestone. I suspect I won't have time to get to it, but while I'm in this code is the right time to think on it at least.

DentonGentry:
In 1.12, when 100.100.100.100 needs to forward a lookup to a well-known global resolver like CloudFlare or dns.google, it automatically use DNS-over-HTTPS so the query doesn't go over the open Internet in the clear.

bradfitz:
@Saklad5, that Cloudflare test page has 4 checks. Which check are you talking about?

> Is there a technical reason that Tailscale's DNS layer can't be disabled entirely? 

You can. There's a checkbox. Or CLI: `tailscale up --accept-dns=false`

Saklad5:
I am specifically referring to the "Secure DNS" checkbox. When Tailscale's VPN is enabled, that check fails with the following note:
> We detected you’re using 1.1.1.1 (a secure DNS resolver) but not over a secure connection.

When the Tailscale VPN is disabled, the check is passed. This is what I'd expect, as I have a DNS [configuration profile](https://github.com/paulmillr/encrypted-dns/blob/master/signed/cloudflare-tls.mobileconfig) active at all times to use DNS-over-TLS.

None of the options Tailscale provides, including whether DNS settings are used or enforced, seems to impact this. Bizarrely, leaving it unenforced causes iOS to use the DNS servers provided by the current network, ignoring both Tailscale _and_ my configuration profile.

bradfitz:
> We detected you’re using 1.1.1.1 (a secure DNS resolver) but not over a secure connection.

[...]

PR #2435, opened 2021-07-15
net/dns/resolver: upgrade forwarded MagicDNS queries to DoH when IP known

bradfitz:
Recognize Cloudflare, Google, Quad9 which are by far the
majority of upstream DNS servers that people use.

RELNOTE=MagicDNS now uses DNS-over-HTTPS when querying popular upstream resolvers,
so DNS queries aren't sent in the clear over the Internet.

Updates #915 (might fix it?)
Updates #988 (gets us closer, if it fixes Android)
Updates #74 (not yet configurable, but progress)
Updates #2056 (not yet configurable, dup of #74?)

### DIFF
--- a/net/dns/resolver/doh_test.go
+++ b/net/dns/resolver/doh_test.go
@@ -1,0 +1,83 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package resolver
+
+import (
+	"context"
+	"flag"
+	"net/http"
+	"testing"
+
+	"golang.org/x/net/dns/dnsmessage"
+)
+
+var testDoH = flag.Bool("test-doh", false, "do real DoH tests against the network")
+
+const someDNSID = 123 // something non-zero as a test; in violation of spec's SHOULD of 0
+
+func someDNSQuestion(t testing.TB) []byte {
+	b := dnsmessage.NewBuilder(nil, dnsmessage.Header{
+		OpCode:           0, // query
+		RecursionDesired: true,
+		ID:               someDNSID,
+	})
+	b.StartQuestions() // err
+	b.Question(dnsmessage.Question{
+		Name:  dnsmessage.MustNewName("tailscale.com."),
+		Type:  dnsmessage.TypeA,
+		Class: dnsmessage.ClassINET,
+	})
+	msg, err := b.Finish()
+	if err != nil {
+		t.Fatal(err)
+	}
+	return msg
+}
+
+func TestDoH(t *testing.T) {
+	if !*testDoH {
+		t.Skip("skipping manual test without --test-doh flag")
+	}
+	if len(knownDoH) == 0 {
+		t.Fatal("no known DoH")
+	}
+
+	f := new(forwarder)
+
+	for ip := range knownDoH {
+		t.Run(ip.String(), func(t *testing.T) {
+			urlBase, c, ok := f.getDoHClient(ip)
+			if !ok {
+				t.Fatal("expected DoH")
+			}
+			res, err := f.sendDoH(context.Background(), urlBase, c, someDNSQuestion(t))
+			if err != nil {
+				t.Fatal(err)
+			}
+			c.Transport.(*http.Transport).CloseIdleConnections()
+
+			var p dnsmessage.Parser
+			h, err := p.Start(res)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if h.ID != someDNSID {
+				t.Errorf("response DNS ID = %v; want %v", h.ID, someDNSID)
+			}
+
+			p.SkipAllQuestions()
+			aa, err := p.AllAnswers()
+			if err != nil {
+				t.Fatal(err)
+			}
+			if len(aa) == 0 {
+				t.Fatal("no answers")
+			}
+			for _, r := range aa {
+				t.Logf("got: %v", r.GoString())
+			}
+		})
+	}
+}
--- a/net/dns/resolver/forwarder.go
+++ b/net/dns/resolver/forwarder.go
@@ -9,15 +9,20 @@ import (
 	"context"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"hash/crc32"
 	"io"
+	"io/ioutil"
 	"math/rand"
 	"net"
+	"net/http"
+	"strings"
 	"sync"
 	"time"
 
 	dns "golang.org/x/net/dns/dnsmessage"
 	"inet.af/netaddr"
+	"tailscale.com/net/netns"
 	"tailscale.com/types/logger"
 	"tailscale.com/util/dnsname"
 	"tailscale.com/wgengine/monitor"
@@ -29,6 +34,11 @@ const headerBytes = 12
 const (
 	// responseTimeout is the maximal amount of time to wait for a DNS response.
 	responseTimeout = 5 * time.Second
+
+	// dohTransportTimeout is how long to keep idle HTTP
+	// connections open to DNS-over-HTTPs servers. This is pretty
+	// arbitrary.
+	dohTransportTimeout = 30 * time.Second
 )
 
 var errNoUpstreams = errors.New("upstream nameservers not set")
@@ -158,6 +168,8 @@ type forwarder struct {
 
 	mu sync.Mutex // guards following
 
+	dohClient map[netaddr.IP]*http.Client
+
 	// routes are per-suffix resolvers to use, with
 	// the most specific routes first.
 	routes []route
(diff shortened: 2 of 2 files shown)


net/dns/resolver/doh_test.go
```diff
@@ -0,0 +1,76 @@
...
+	msg, err := b.Finish()
+	if err != nil {
+		t.Fatal(err)
+	}
+	return msg
+}
+
+func TestDoH(t *testing.T) {
+	if !*testDoH {
+		t.Skip("skipping manual test without --test-doh flag")
```
crawshaw, line 38: No local DoH service? :)

net/dns/resolver/forwarder.go
```diff
@@ -210,6 +218,57 @@ func (f *forwarder) packetListener(ip netaddr.IP) (packetListener, error) {
...
+	if c, ok := f.dohClient[ip]; ok {
+		return urlBase, c, true
+	}
+	if f.dohClient == nil {
+		f.dohClient = map[netaddr.IP]*http.Client{}
+	}
+	nsDialer := netns.NewDialer()
+	c = &http.Client{
+		Transport: &http.Transport{
+			IdleConnTimeout: 30 * time.Second,
```
crawshaw, line 237: Motivation? Feels short.

crawshaw (approved):
Is adding DoH targets via the control plane future work?

net/dns/resolver/doh_test.go
bradfitz: yeah sorry

net/dns/resolver/forwarder.go
bradfitz: happy to hear principled alternatives :)

Moved it to a const with docs and declaration of arbitrariness.

net/dns/resolver/forwarder.go
```diff
@@ -432,3 +501,48 @@ func (p *closePool) Close() error {
 	}
 	return nil
 }
+
+var knownDoH = map[netaddr.IP]string{}
```
sailorfrag, line 505: Rather than baking these into the client, I wonder if it would be better for control to pass along "DNS IP", "DoH IP" and "DoH name". This is probably fine for now, maybe we can transition to that later.

net/dns/resolver/forwarder.go
crawshaw: My instinct says: I ~never change DNS servers. I don’t resolve very often, but when I do I want the latency to be low. Which suggests: no timeout.

net/dns/resolver/forwarder.go
```diff
@@ -220,9 +284,14 @@ func (f *forwarder) packetListener(ip netaddr.IP) (packetListener, error) {
 // iOS and we want the number of pending DNS lookups to be bursty
 // without too much associated goroutine/memory cost.
 func (f *forwarder) send(ctx context.Context, txidOut txid, closeOnCtxDone *closePool, packet []byte, dst netaddr.IPPort) ([]byte, error) {
-	// TODO(bradfitz): if dst.IP is 8.8.8.8 or 8.8.4.4 or 1.1.1.1, etc, or
-	// something dynamically probed earlier to support DoH or DoT,
-	// do that here instead.
+	// Upgrade known DNS IPs to DoH (DNS-over-HTTPs).
+	if urlBase, dc, ok := f.getDoHClient(dst.IP()); ok {
+		res, err := f.sendDoH(ctx, urlBase, dc, packet)
```
sailorfrag, line 289: Before sending to DoH, the ID should be replaced with zero; the response should probably have the correct ID substituted back in.

> In order to maximize HTTP cache friendliness, DoH clients using media
> formats that include the ID field from the DNS message header, such
>   as "application/dns-message", SHOULD use a DNS ID of 0 in every DNS
>   request.  HTTP correlates the request and response, thus eliminating
>   the need for the ID in a media type such as "application/dns-
>   message".  The use of a varying DNS ID can cause semantically
>   equivalent DNS queries to be cached separately.

net/dns/resolver/forwarder.go
bradfitz: Agree. The wire protocol supports this but the plumbing can come later. First I want to see just one part working.

net/dns/resolver/forwarder.go
bradfitz: I'm going to punt on this for now, as it's just a SHOULD and I'd rather not parse & re-pack the DNS queries yet.

I tested that it doesn't matter with this, though:

```diff
diff --git a/net/dns/resolver/doh_test.go b/net/dns/resolver/doh_test.go
index 978d76f1..e283f32e 100644
--- a/net/dns/resolver/doh_test.go
+++ b/net/dns/resolver/doh_test.go
@@ -15,10 +15,13 @@ import (
 
 var testDoH = flag.Bool("test-doh", false, "do real DoH tests against the network")
 
+const someDNSID = 123 // something non-zero as a test; in violation of spec's SHOULD of 0
+
 func someDNSQuestion(t testing.TB) []byte {
        b := dnsmessage.NewBuilder(nil, dnsmessage.Header{
                OpCode:           0, // query
                RecursionDesired: true,
+               ID:               someDNSID,
        })
        b.StartQuestions() // err
        b.Question(dnsmessage.Question{
@@ -56,9 +59,13 @@ func TestDoH(t *testing.T) {
                        c.Transport.(*http.Transport).CloseIdleConnections()
 
                        var p dnsmessage.Parser
-                       if _, err := p.Start(res); err != nil {
+                       h, err := p.Start(res)
+                       if err != nil {
                                t.Fatal(err)
                        }
+                       if h.ID != someDNSID {
+                               t.Errorf("response DNS ID = %v; want %v", h.ID, someDNSID)
+                       }
 
                        p.SkipAllQuestions()
                        aa, err := p.AllAnswers()
```

net/dns/resolver/forwarder.go
bradfitz: Added named const & docs.

net/dns/resolver/forwarder.go
```diff
@@ -210,6 +218,57 @@ func (f *forwarder) packetListener(ip netaddr.IP) (packetListener, error) {
...
+		},
+	}
+	f.dohClient[ip] = c
+	return urlBase, c, true
+}
+
+const dohType = "application/dns-message"
+
+func (f *forwarder) sendDoH(ctx context.Context, urlBase string, c *http.Client, packet []byte) ([]byte, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", urlBase+"?dns="+base64.RawURLEncoding.EncodeToString(packet), nil)
```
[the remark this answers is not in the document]
bradfitz: Done.

bradfitz:
> Is adding DoH targets via the control plane future work?

Yeah. That requires more plumbing. I linked a bunch of those bugs in the commit message now.

net/dns/resolver/forwarder.go
```diff
@@ -210,6 +222,57 @@ func (f *forwarder) packetListener(ip netaddr.IP) (packetListener, error) {
...
+}
+
+const dohType = "application/dns-message"
+
+func (f *forwarder) sendDoH(ctx context.Context, urlBase string, c *http.Client, packet []byte) ([]byte, error) {
+	req, err := http.NewRequestWithContext(ctx, "POST", urlBase, bytes.NewReader(packet))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", dohType)
```
sailorfrag, line 261: I think setting both Accept and Content-Type would be good

net/dns/resolver/forwarder.go
sailorfrag: Okay, fair. It's in a section talking about improving HTTP cache behavior, which POST obliterates anyway so that's probably fine.

sailorfrag (approved):

net/dns/resolver/forwarder.go
bradfitz: I see it included in examples (https://datatracker.ietf.org/doc/html/rfc8484#page-5) but looks like the Accept one is pretty optional, though:

> The DoH client SHOULD include an HTTP Accept request header field to
>  indicate what type of content can be understood in response.
>  Irrespective of the value of the Accept request header field, the
> client MUST be prepared to process "application/dns-message" 

(and empirically isn't required by any provider)

Worth the bandwidth? With HTTP/2 it's not much, but iOS doesn't use HTTP/2 yet so it's a bit more.

bradfitz: Updated code with comment:

```patch
diff --git a/net/dns/resolver/forwarder.go b/net/dns/resolver/forwarder.go
index 8a15c099..a3184f8c 100644
--- a/net/dns/resolver/forwarder.go
+++ b/net/dns/resolver/forwarder.go
@@ -259,6 +259,12 @@ func (f *forwarder) sendDoH(ctx context.Context, urlBase string, c *http.Client,
                return nil, err
        }
        req.Header.Set("Content-Type", dohType)
+       // Note: we don't currently set the Accept header (which is
+       // only a SHOULD in the spec) as iOS doesn't use HTTP/2 and
+       // we'd rather save a few bytes on outgoing requests when
+       // empirically no provider cares about the Accept header's
+       // absence.
+
        hres, err := c.Do(req)
        if err != nil {
                return nil, err
```

bradfitz:
> Okay, fair. It's in a section talking about improving HTTP cache behavior, which POST obliterates anyway so
> that's probably fine.

The whole cache part of the spec is super weird. There aren't going to be rando cache intermediates here and Google et al aren't using rando off-the-shelf HTTP caches in front of their DNS servers. Any caching they're doing is part of the application layer after it's parsed the HTTP framing.

sreeram-boyapati:
Is it possible to support nextdns.io? Some of the power users have custom blocklists and adblock lists. DNS-Over-HTTPS with magicDNS and nextDNS nameservers would be a great addition.

DentonGentry:
Could you file that as a feature request in https://github.com/tailscale/tailscale/issues ?

PR #2437, opened 2021-07-15
tstest/integration/vms: use one testcontrol instance per VM

Xe:
This paves the way for future MagicDNS tests.

### DIFF
--- a/tstest/integration/vms/harness_test.go
+++ b/tstest/integration/vms/harness_test.go
@@ -10,14 +10,19 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"log"
 	"net"
+	"net/http"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
+	"strconv"
+	"sync"
 	"testing"
 	"time"
 
-	"github.com/gliderlabs/ssh"
+	"golang.org/x/crypto/ssh"
 	"golang.org/x/net/proxy"
 	"inet.af/netaddr"
 	"tailscale.com/tstest/integration"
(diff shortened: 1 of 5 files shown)


bradfitz:
What's the background here? I think we should only have 1 control server for all VMs under test.

Xe:
The background here is that we will need to modify things like magic DNS settings in tests and the easiest way to do that would be to have one testcontrol instance and the like per VM being tested. Because these vms are being tested in parallel, this means that modifying the testcontrol state per VM would make it harder for us to isolate testcontrol state to each node being tested.

crawshaw:
Parallel testing of VMs means maintaining different configs (DNS, netmap
add/remove, etc) simultaneously. It seemed like a lot of extra machinery to
build into testcontrol compared with running multiple.

bradfitz:
Oh, this is for testing VMs in isolation? That's not super interesting, is it?

I thought the VM-based testing world was moving towards interop testing, sending packets between VMs. In that case they should be part of the same control world.

Xe:
Yes, the vm interop testing is going to need additional work (and likely additional vms) spun up to individual control worlds. I'm still figuring it out.

crawshaw:
Individual VMs are still interesting. We currently have zero tests that we
configure the OS DNS correctly.

tstest/integration/vms/harness_test.go
```diff
@@ -28,10 +33,102 @@ type Harness struct {
...
+
+	lc := &integration.LogCatcher{}
+	if *verboseLogcatcher {
+		lc.UseLogf(t.Logf)
+	}
+	mux.Handle("/c/", lc)
+
+	// This handler will let the virtual machines tell the host information about that VM.
+	// This is used to maintain a list of port->IP address mappings that are known to be
+	// working. This allows later steps to connect over SSH. This returns no response to
```
sailorfrag, line 78: Should it return a HTTP 204 then?

sailorfrag (approved):

tstest/integration/vms/harness_test.go
Xe: I don't know how all of the versions of curl being used interact with HTTP 204 to feel safe in using it.

PR #2439, opened 2021-07-16
tailcfg: add IP and Types field to PingRequest

simenghe:

Additions to the PingRequest struct that will be used when initiating TSMP pings from Control.
An IP field is added for the IP that we are trying to Ping (TSMP)
A Types field is added for future specification of what types of Pings we want to run.

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -1303,3 +1303,59 @@ func (c *Direct) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) error {
 
 	return nil
 }
+
+// tsmpPing sends a Ping to pr.IP, and sends an http request back to pr.URL
+// with ping response data.
+func tsmpPing(logf logger.Logf, c *http.Client, pr *tailcfg.PingRequest, pinger Pinger) error {
+	var err error
+	if pr.URL == "" {
+		return errors.New("invalid PingRequest with no URL")
+	}
+	if pr.IP.IsZero() {
+		return fmt.Errorf("PingRequest with no proper IP got %v", pr.IP)
+	}
+	if !strings.Contains(pr.Types, "TSMP") {
+		return fmt.Errorf("PingRequest with no TSMP in Types, got : %v", pr.Types)
+	}
+
+	now := time.Now()
+	pinger.Ping(pr.IP, true, func(res *ipnstate.PingResult) {
+		// Currently does not check for error since we just return if it fails.
+		err = postPingResult(now, logf, c, pr, res)
+	})
+	return err
+}
+
+func postPingResult(now time.Time, logf logger.Logf, c *http.Client, pr *tailcfg.PingRequest, res *ipnstate.PingResult) error {
+	if res.Err != "" {
+		return errors.New(res.Err)
+	}
+	duration := time.Since(now)
+	if pr.Log {
+		logf("TSMP ping to %v completed in %v seconds. pinger.Ping took %v seconds", pr.IP, res.LatencySeconds, duration.Seconds())
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+
+	jsonPingRes, err := json.Marshal(res)
+	if err != nil {
+		return err
+	}
+	// Send the results of the Ping, back to control URL.
+	req, err := http.NewRequestWithContext(ctx, "POST", pr.URL, bytes.NewBuffer(jsonPingRes))
+	if err != nil {
+		return fmt.Errorf("http.NewRequestWithContext(%q): %w", pr.URL, err)
+	}
+	if pr.Log {
+		logf("tsmpPing: sending ping results to %v ...", pr.URL)
+	}
+	t0 := time.Now()
+	_, err = c.Do(req)
+	d := time.Since(t0).Round(time.Millisecond)
+	if err != nil {
+		return fmt.Errorf("tsmpPing error: %w to %v (after %v)", err, pr.URL, d)
+	} else if pr.Log {
+		logf("tsmpPing complete to %v (after %v)", pr.URL, d)
+	}
+	return nil
+}
--- a/control/controlclient/direct_test.go
+++ b/control/controlclient/direct_test.go
@@ -6,9 +6,13 @@ package controlclient
 
 import (
 	"encoding/json"
+	"net/http"
+	"net/http/httptest"
 	"testing"
+	"time"
 
 	"inet.af/netaddr"
+	"tailscale.com/ipn/ipnstate"
 	"tailscale.com/tailcfg"
 	"tailscale.com/types/wgkey"
 )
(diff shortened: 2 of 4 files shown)


tailcfg/tailcfg.go
```diff
@@ -920,6 +920,12 @@ type PingRequest struct {
 	// Log is whether to log about this ping in the success case.
 	// For failure cases, the client will log regardless.
 	Log bool `json:",omitempty"`
+
+	// Types is the types of ping that is initiated. Can be TSMP, ICMP or DISCO
```
bradfitz, line 924: singular or plural? if plural, how? comma-separated, or?

Also, DISCO isn't an acronym. And missing final period.

tailcfg/tailcfg.go
```diff
@@ -920,6 +920,12 @@ type PingRequest struct {
...
+	Types string
+
+	// IP is the IP that we will attempt to Ping
```
bradfitz, line 927: lowercase ping. trailing period.

Also, no "we".  Maybe: "IP is the ping target."

But clarify when this is required. Because for old-style PingRequest, only URL is required.

control/controlclient/direct.go
```diff
@@ -1303,3 +1303,52 @@ func (c *Direct) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) error {
...
+			return
+		}
+		// Send the results of the Ping, back to control URL.
+		req, err := http.NewRequestWithContext(ctx, "POST", pr.URL, bytes.NewBuffer(jsonPingRes))
+		if err != nil {
+			logf("http.NewRequestWithContext(%q): %v", pr.URL, err)
+			return
+		}
+		if pr.Log {
+			logf("tsmpPing: sending ping to %v ...", pr.URL)
```
sailorfrag, line 1342: should that be pr.IP, or should the log be "sending ping results to %v"?

simenghe: changed to ping results to %v

tailcfg/tailcfg.go
```diff
@@ -926,6 +926,14 @@ type PingRequest struct {
 	// Log is whether to log about this ping in the success case.
 	// For failure cases, the client will log regardless.
 	Log bool `json:",omitempty"`
+
+	// Types is the types of ping that is initiated. Can be TSMP, ICMP or disco.
+	// Types will be comma separated, such as TSMP, disco.
```
bradfitz, line 931: such "TSMP,disco" I assume you mean? I don't think we want to permit spaces in there.

tailcfg/tailcfg.go
```diff
@@ -926,6 +926,14 @@ type PingRequest struct {
 	// Log is whether to log about this ping in the success case.
 	// For failure cases, the client will log regardless.
 	Log bool `json:",omitempty"`
+
```
bradfitz, line 929: these new docs don't explain the relationship between the URL-only are-you-there pings and the pings with a type+IP.

simenghe: updated comments at the top of PingRequest

control/controlclient/direct.go
```diff
@@ -1303,3 +1303,61 @@ func (c *Direct) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) error {
...
+func tsmpPing(logf logger.Logf, c *http.Client, pr *tailcfg.PingRequest, pinger Pinger) {
+	if pr.URL == "" {
+		logf("invalid PingRequest with no URL")
+		return
+	}
+	if pr.IP.IsZero() {
+		logf("PingRequest with no proper IP got %v", pr.IP)
+		return
+	}
+	if !strings.Contains(pr.Types, "TSMP") {
```
sailorfrag, line 1318: if `pr.Types` is a comma separated list, then if `TSMP` is a substring of some other valid entry, this will still match. Better would be to split on commas and then scan the resulting slice for exact match. This works for fine now (with the valid options just being `TSMP` and `disco`), but is a possible surprise later on.

... would it be easier just to make pr.Types a `[]string`? @bradfitz would there be any unfortunate consequences to doing that?

control/controlclient/direct.go
```diff
@@ -1303,3 +1303,61 @@ func (c *Direct) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) error {
 
 	return nil
 }
+
+// tsmpPing sends a Ping to pr.IP, and sends an http request back to pr.URL
+// with ping response data.
+func tsmpPing(logf logger.Logf, c *http.Client, pr *tailcfg.PingRequest, pinger Pinger) {
+	if pr.URL == "" {
+		logf("invalid PingRequest with no URL")
```
sailorfrag, line 1311: Consider returning these failure states as errors, for the caller to log

control/controlclient/direct.go
```diff
@@ -1303,3 +1303,61 @@ func (c *Direct) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) error {
...
+	now := time.Now()
+	pinger.Ping(pr.IP, true, func(res *ipnstate.PingResult) {
+		// Currently does not check for error since we just return if it fails.
+		postPingResult(now, logf, c, pr, res)
+	})
+}
+
+func postPingResult(now time.Time, logf logger.Logf, c *http.Client, pr *tailcfg.PingRequest, res *ipnstate.PingResult) error {
+	if res.Err != "" {
+		logf(res.Err)
```
sailorfrag, line 1332: If returning an error, don't log it here, log it in the caller

control/controlclient/direct_test.go
```diff
@@ -103,3 +105,46 @@ func TestNewHostinfo(t *testing.T) {
...
+
+	now := time.Now()
+	res := &ipnstate.PingResult{
+		IP:       "123.456.7890",
+		Err:      "",
+		NodeName: "testnode",
+	}
+
+	pr := &tailcfg.PingRequest{
+		URL: "testurl.com",
```
sailorfrag, line 139: Is this going to send a HTTP request to `testurl.com`?

I'd recommend something under `example.com` because that's guaranteed to always be reserved. But also, I think the `httpc` should be a special test http client that doesn't actually make real requests but instead lets you inspect the contents. Or perhaps set the URL to a localhost HTTP server set up by the test that captures the result to inspect.

simenghe: now using httptest server

control/controlclient/direct.go
```diff
@@ -1303,3 +1303,59 @@ func (c *Direct) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) error {
 
 	return nil
 }
+
+// tsmpPing sends a Ping to pr.IP, and sends an http request back to pr.URL
+// with ping response data.
+func tsmpPing(logf logger.Logf, c *http.Client, pr *tailcfg.PingRequest, pinger Pinger) error {
+	var err error
+	if pr.URL == "" {
+		return errors.New("invalid PingRequest with no URL")
```
sailorfrag, line 1312: In general, it's nice to put the various errors in exported package variables so that the caller can test for what the error is rather than just printing it out. In this case, it doesn't really matter, we can adjust that later if needed.

sailorfrag (approved):
I think this is OK for now after fixing the format strings.

sailorfrag (approved):

PR #2441, opened 2021-07-16
tstest/integration/vms: codegen for top level tests

Xe:
This moves the distribution definitions into a maintainable hujson file
instead of just existing as constants in `distros.go`. Comments are
maintained from the inline definitions.

This uses jennifer[1] for hygenic source tree creation. This allows us
to generate a unique top-level test for each VM run. This should
hopefully help make the output of `go test` easier to read.

This also separates each test out into its own top-level test so that we
can better track the time that each distro takes. I really wish there
was a way to have the `test_codegen.go` file _always_ run as a part of
the compile process instead of having to rely on people remembering to
run `go generate`, but I am limited by my tools.

This will let us remove the `-distro-regex` flag and use `go test -run`
to pick which distros are run.


tstest/integration/vms/gen/test_codegen.go
```diff
@@ -0,0 +1,57 @@
...
+package main
+
+import (
+	_ "embed"
+	"fmt"
+	"log"
+	"os"
+	"time"
+
+	. "github.com/dave/jennifer/jen"
```
Xe, line 16: The [upstream examples](https://github.com/dave/jennifer#jennifer) suggest the use of a dot import, so I use one here.

tstest/integration/vms/top_level_test.go
```diff
@@ -0,0 +1,121 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build linux
+
+package vms
+
+import "testing"
+
+// Code generated by tstest/integration/vms/gen/test_codegen.go DO NOT EDIT
```
sailorfrag, line 11: Should this be done in `go:generate` style? I'm not sure exactly what that entails, but we have a couple examples in the tree already (e.g. Cloner)

tstest/integration/vms/gen/test_codegen.go
sailorfrag: I think it's preferable not to do that anyway; it doesn't look like it'll make the code much more annoying to write.

tstest/integration/vms/top_level_test.go
Xe: It is!

sailorfrag (approved):

PR #2461, opened 2021-07-19
tstest/integration: taildrop integration test

JulianKnodt:
[WIP] Adds an integration test for taildrop, testing that taildrop should work in both
directions and that the files are identical to the original.


ipn/ipnlocal/local.go
```diff
@@ -2589,13 +2589,17 @@ func (b *LocalBackend) FileTargets() ([]*apitype.FileTarget, error) {
...
 	}
-	if !b.capFileSharing {
-		return nil, errors.New("file sharing not enabled by Tailscale admin")
-	}
-	for _, p := range nm.Peers {
-		if p.User != nm.User {
-			continue
+	/*
+		if !b.capFileSharing {
+			return nil, errors.New("file sharing not enabled by Tailscale admin")
```
JulianKnodt, line 2594: this is a temp change before changing test control, will not be merged

ipn/ipnlocal/local.go
```diff
@@ -2589,13 +2589,17 @@ func (b *LocalBackend) FileTargets() ([]*apitype.FileTarget, error) {
...
 		}
+	*/
+	for _, p := range nm.Peers {
+		/*
+			if p.User != nm.User {
+				continue
+			}
+		*/
```
JulianKnodt, line 2602: even with this commented out, peerAPI == "", see comment below about exact location, is peerAPI enabled in integration tests?

tstest/integration/testcontrol/testcontrol.go
```diff
@@ -293,20 +296,24 @@ func (s *Server) getUser(nodeKey tailcfg.NodeKey) (*tailcfg.User, *tailcfg.Login
 	if u, ok := s.users[nodeKey]; ok {
 		return u, s.logins[nodeKey]
 	}
-	id := tailcfg.UserID(len(s.users) + 1)
+	userID := tailcfg.UserID(len(s.users) + 1)
+	if s.AllNodesSameUser {
+		userID = 42
+	}
```
JulianKnodt, line 302: This is a hacky way to make all of the users in integration tests have the same id. For some reason it causes an overwrite in a map somewhere, resulting in unknown target, not in your tailnet

tstest/integration/integration_test.go
```diff
@@ -311,6 +311,96 @@ func TestAddPingRequest(t *testing.T) {
...
+	n2.AwaitRunning(t)
+
+	target := n2.AwaitIP(t)
+
+	tempDir := t.TempDir()
+	tmpFile, err := ioutil.TempFile(tempDir, "taildrop")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(tmpFile.Name())
```
bradfitz, line 356: no need; if it's under the testing.T.TempDir, the testing package will clean it all up

tstest/integration/integration_test.go
```diff
@@ -311,6 +311,96 @@ func TestAddPingRequest(t *testing.T) {
...
+	n2.AwaitListening(t)
+	n1.MustUp()
+	n2.MustUp()
+	n1.AwaitRunning(t)
+	n2.AwaitRunning(t)
+
+	target := n2.AwaitIP(t)
+
+	tempDir := t.TempDir()
+	tmpFile, err := ioutil.TempFile(tempDir, "taildrop")
```
bradfitz, line 352: seems overkill to use TempFile in an empty directory.

just use ioutil.WriteFile?

tstest/integration/integration_test.go
```diff
@@ -311,6 +311,96 @@ func TestAddPingRequest(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(string(targetsOutput), err)
+	}
+	if !bytes.Contains(targetsOutput, []byte(target.String())) {
+		t.Errorf("Missing target from cp -targets, want: %v, in: %v", target, targetsOutput)
+	}
+
+	// Need ALL_PROXY=SOCKS because peerapi was not configured to work with userspace-networking
+	// yet.
+	cpCmd := n1.Tailscale("ALL_PROXY=SOCKS", "file", "cp", tmpFile.Name(), fmt.Sprintf("%s:", target))
```
bradfitz, line 376: what is this ALL_PROXY=SOCKS doing here? it's being passed as an argument? to the cmd/tailscale command? I'm missing something. How does this work?

bradfitz (commented):
Nice test!

tstest/integration/integration_test.go
JulianKnodt: hmmmmm I have no idea how it works exactly, but by setting this it makes some `os.Setenv` cause it to use SOCKS5 lol
it's not a direct argument though, just an env var

JulianKnodt:
ah shit I just realized that this actually isn't copying the file, it's just reading from the same file both times, I changed it to be a bit more clear and it's not getting it now

tstest/integration/integration_test.go
JulianKnodt: oh wait this is just broken never mind

JulianKnodt:
@Xe I think the networking issue you mentioned is coming up here! Any help would be appreciated in how to fix it

PR #2462, opened 2021-07-19
util/hujson: initial commit a HuJSON AST parser and packer

dsnet:
This package provides a HuJSON parser into an AST represented by the following Go types:
* Value: A JSON literal, object, or array with surrounding whitespace/comments.
* Extra: Raw bytes of whitespace and comments.
* Literal: Raw bytes of a JSON literal (i.e., null, boolean, string, or number).
* Object: A JSON object.
* Array: A JSON array.

The Parse function parses raw HuJSON input as a Value.
The Value.Pack method serializes the AST as raw HuJSON output.

### DIFF
--- a/util/hujson/types.go
+++ b/util/hujson/types.go
@@ -1,0 +1,450 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package hujson contains a parser and packer for the HuJSON format.
+//
+// HuJSON is an extension of standard JSON (as defined in RFC 8259) in order to
+// make it more suitable for humans and configuration files. In particular,
+// it supports line comments (e.g., //...), block comments (e.g., /*...*/), and
+// trailing commas after the last member or element in a JSON object or array.
+//
+//
+// Functionality
+//
+// The Parse function parses HuJSON input as a Value,
+// which is a syntax tree exactly representing the input.
+// Comments and whitespace are represented using the Extra type.
+// Composite types in JSON are represented using the Object and Array types.
+// Primitive types in JSON are represented using the Literal type.
+// The Value.Pack method serializes the syntax tree as raw output,
+// which is byte-for-byte identical to the input if no transformations
+// were performed on the value.
+//
+// A HuJSON value can be transformed using the Minimize, Standardize,
+// NormalizeNames, and Reformat methods. Each of these methods mutate the value
+// in place. Call the Clone method in order to preserve the original value.
+// The Minimize and Standardize methods coerces HuJSON into valid standard JSON.
+// The Reformat method formats the value; it is similar to `go fmt`,
+// but instead for the HuJSON and standard JSON format.
+//
+//
+// Grammar
+//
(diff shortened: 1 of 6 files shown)


bradfitz:
Was putting this in https://github.com/tailscale/hujson already ruled out?

util/hujson/types.go
```diff
@@ -0,0 +1,398 @@
...
+			return math.Inf(-1)
+		}
+	}
+	return 0
+}
+
+// Object is an exact syntactic representation of a JSON object.
+type Object struct {
+	// Members are the members of a JSON object.
+	Members [][2]Value
```
bradfitz, line 262: can you make `[2]Value` a named type so docs can be attached to it? What are the values at index 0 and 1? etc.

util/hujson/find.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	return nil
+}
+
+func (b Literal) equalString(s string) bool {
+	// Fast-path: Assume there are no escape characters.
+	if len(b) >= 2 && b[0] == '"' && b[len(b)-1] == '"' && bytes.IndexByte(b, '\\') < 0 {
+		return string(b[len(`"`):len(b)-len(`"`)]) == s
+	}
+	// Slow-path: Unescape the string and then compare it.
```
bradfitz, line 67: // TODO: implement the slow path without allocations :)

util/hujson/types.go
```diff
@@ -0,0 +1,398 @@
...
+// Value is an exact syntactic representation of a JSON value.
+// The starting and ending byte offsets are populated when parsing,
+// but are otherwise ignored when packing.
+type Value struct {
+	// BeforeExtra are the comments and whitespace before Value.
+	BeforeExtra Extra
+	// StartOffset is the offset of the first byte in Value.
+	StartOffset int
+	// Value is the JSON value itself.
+	Value value // Literal | *Object | *Array
```
bradfitz, line 89: unexported type on an exported field is a little weird

dsnet:
I didn't think much about it.

What's the compatibility agreement policy for `github.com/tailscale/hujson`? Should we just blast away the old code? Or place it under a `v2` sub-directory (something I think @crawshaw  suggested)?

For my own interest, how do we generally decide whether something is a new repo or something goes within this repo?

util/hujson/types.go
dsnet: Done.

util/hujson/find.go
dsnet: I removed this file. Will submit in future PR if we need this functionality.

util/hujson/types.go
dsnet: Done.

dsnet:
Unless someone prefers otherwise. I'm thinking of just comitting this under `util`. 

We can't do `github.com/tailscale/hujson` unless we're willing to make a breaking change there. Even if we do, we'll probably have to go through some dependency dance for the code that currently depends on that module.

We could put it under `github.com/tailscale/hujson/v2`, but I'm not sure we want to commit to creating another module, especially one under the `github.com` domain name.

dsnet:
Moving to https://github.com/tailscale/hujson/pull/6.

PR #2464, opened 2021-07-20
util/deephash: make hash type opaque

dsnet:
The fact that Hash returns a [sha256.Size]byte leaks details about
the underlying hash implementation. This could very well be any other
hashing algorithm with a possible different block size.

Abstract this implementation detail away by declaring an opaque type
that is comparable. While we are changing the signature of UpdateHash,
rename it to just Update to reduce stutter (e.g., deephash.Update).

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -53,8 +53,17 @@ func (h *hasher) setBufioWriter(w *bufio.Writer) (old *bufio.Writer) {
 	return old
 }
 
-// Hash returns the raw SHA-256 (not hex) of v.
-func (h *hasher) Hash(v interface{}) (hash [sha256.Size]byte) {
+// Sum is an opaque checksum type that is comparable.
+type Sum struct {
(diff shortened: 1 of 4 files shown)


util/deephash/deephash.go
```diff
@@ -53,8 +53,17 @@ func (h *hasher) setBufioWriter(w *bufio.Writer) (old *bufio.Writer) {
...
+// Sum is an opaque checksum type that is comparable.
+type Sum struct {
+	sum [sha256.Size]byte
+}
+
+func (s Sum) String() string {
+	return hex.EncodeToString(s.sum[:])
+}
+
 // Hash returns the raw SHA-256 (not hex) of v.
```
bradfitz, line 65: these docs could be updated now. the "not hex" part is a little old at this point.

bradfitz (approved):

util/deephash/deephash.go

Issue #1591, opened 2021-03-26
systemd-networkd removes tailscale routing table entries


pfactum:
**Describe the bug**

If during the boot the networking/DNS is not available yet at the moment tailscaled starts, the tailscale interface never comes up and stucks in a loop like:

```
Mar 26 12:25:01 rain tailscaled[283]: logtail: dial "log.tailscale.io:443" failed: dial tcp: lookup log.tailscale.io: Temporary failure in name resolution (in 1ms)
Mar 26 12:25:01 rain tailscaled[283]: logtail: upload: log upload of 307 bytes compressed failed: Post "https://log.tailscale.io/c/tailnode.log.tailscale.io/…": dial tcp: lookup log.tailscale.io: Temporary failure in name resolution
Mar 26 12:25:01 rain tailscaled[283]: logtail: backoff: 39 msec
…
Mar 26 12:26:14 rain tailscaled[404]: trying bootstrapDNS("derp3.tailscale.com", "68.183.179.66") for "login.tailscale.com" ...
Mar 26 12:26:14 rain tailscaled[404]: bootstrapDNS("derp3.tailscale.com", "68.183.179.66") for "login.tailscale.com" error: Get "https://derp3.tailscale.com/bootstrap-dns?q=login.tailscale.com": dial tcp 68.183.179.66:443: connect: network is unreachable
…
Mar 26 12:26:14 rain tailscaled[404]: control: authRoutine: backoff: 10588 msec
Mar 26 12:26:18 rain tailscaled[404]: logtail: dial "log.tailscale.io:443" failed: dial tcp 34.210.105.16:443: connect: network is unreachable (in 3ms)
Mar 26 12:26:18 rain tailscaled[404]: logtail: upload: log upload of 422 bytes compressed failed: Post "https://log.tailscale.io/c/tailnode.log.tailscale.io/…": dial tcp 34.210.105.16:443: connect: network is unreachable
Mar 26 12:26:18 rain tailscaled[404]: logtail: backoff: 12570 msec
Mar 26 12:26:24 rain tailscaled[404]: control: authRoutine: state:authenticating; wantLoggedIn=true
Mar 26 12:26:24 rain tailscaled[404]: control: direct.TryLogin(token=false, flags=0)
Mar 26 12:26:24 rain tailscaled[404]: control: doLogin(regen=false, hasUrl=false)
```

even if the networking is up and running a little bit later, and required hosts are perfectly available through telnet.

Restarting tailscaled fixes the issue.

[...]

PR #2466, opened 2021-07-20
wgengine/router: take a link monitor

bradfitz:
Prep for #1591 which will need to make Linux's router react to changes
that the link monitor observes.

The router package already depended on the monitor package
transitively. Now it's explicit.

### DIFF
--- a/wgengine/router/router_default.go
+++ b/wgengine/router/router_default.go
@@ -7,12 +7,16 @@
 package router
 
 import (
+	"fmt"
+	"runtime"
+
 	"golang.zx2c4.com/wireguard/tun"
 	"tailscale.com/types/logger"
+	"tailscale.com/wgengine/monitor"
 )
 
-func newUserspaceRouter(logf logger.Logf, tunname string, tunDev tun.Device, netChanged func()) Router {
-	return NewFakeRouter(logf, tunname, tunDev, netChanged)
+func newUserspaceRouter(logf logger.Logf, tunname string, tunDev tun.Device, linkMon *monitor.Mon) Router {
+	panic(fmt.Sprintf("unsupported OS %q", runtime.GOOS))
 }
 
 func cleanup(logf logger.Logf, interfaceName string) {
--- a/wgengine/router/router_linux_test.go
+++ b/wgengine/router/router_linux_test.go
@@ -18,6 +18,8 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"golang.zx2c4.com/wireguard/tun"
 	"inet.af/netaddr"
+	"tailscale.com/types/logger"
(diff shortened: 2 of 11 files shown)


wgengine/router/router_default.go
```diff
@@ -7,12 +7,16 @@
...
+
 	"golang.zx2c4.com/wireguard/tun"
 	"tailscale.com/types/logger"
+	"tailscale.com/wgengine/monitor"
 )
 
-func newUserspaceRouter(logf logger.Logf, tunname string, tunDev tun.Device, netChanged func()) Router {
-	return NewFakeRouter(logf, tunname, tunDev, netChanged)
+func newUserspaceRouter(logf logger.Logf, tunname string, tunDev tun.Device, linkMon *monitor.Mon) Router {
+	panic(fmt.Sprintf("unsupported OS %q", runtime.GOOS))
```
crawshaw, line 19: is there a netstack-only os this is going to break?

wgengine/router/router_linux_test.go
```diff
@@ -314,8 +316,15 @@ ip route add throw 192.168.0.0/24 table 52` + basic,
...
+	mon, err := monitor.New(logger.Discard)
+	if err != nil {
+		t.Fatal(err)
+	}
+	mon.Start()
+	defer mon.Close()
+
 	fake := NewFakeOS(t)
-	router, err := newUserspaceRouterAdvanced(t.Logf, "tailscale0", fake.netfilter4, fake.netfilter6, fake, true, true)
+	router, err := newUserspaceRouterAdvanced(t.Logf, "tailscale0", mon, fake.netfilter4, fake.netfilter6, fake, true, true)
```
crawshaw, line 327: not going to cause test flakes when a linkchange comes through during the test?

crawshaw (approved):

wgengine/router/router_linux_test.go
bradfitz: Turns out this code didn't even compile before. It called a function which didn't exist in the code.

But we have no GOOS for which WireGuard compiles today but we don't.

So this is a useless change that can't currently be exercised.

wgengine/router/router_default.go
bradfitz: Turns out this code didn't even compile before. It called a function which didn't exist in the code.

But we have no GOOS for which WireGuard compiles today but we don't.

So this is a useless change that can't currently be exercised.

wgengine/router/router_linux_test.go
bradfitz: Well, it's currently unused, so that'd be impressive.

And with how I plan to use it, no.

PR #2469, opened 2021-07-20
tstest/integration: build binaries only once

josharian:
The existing code relied on the Go build cache to avoid
needless work when obtaining the tailscale binaries.

For non-obvious reasons, the binaries were getting re-linked
every time, which added 600ms or so on my machine to every test.

Instead, build the binaries exactly once, before doing anything else.
This reduces the time to run 'go test -count=5' from 34s to 10s
on my machine.

### DIFF
--- a/tstest/integration/integration.go
+++ b/tstest/integration/integration.go
@@ -43,38 +43,68 @@ import (
 	"tailscale.com/version"
 )
 
-// Binaries are the paths to a tailscaled and tailscale binary.
-// These can be shared by multiple nodes.
-type Binaries struct {
-	Dir    string // temp dir for tailscale & tailscaled
-	Daemon string // tailscaled
-	CLI    string // tailscale
-}
-
-// BuildTestBinaries builds tailscale and tailscaled, failing the test
-// if they fail to compile.
-func BuildTestBinaries(t testing.TB) *Binaries {
-	td := t.TempDir()
-	build(t, td, "tailscale.com/cmd/tailscaled", "tailscale.com/cmd/tailscale")
-	return &Binaries{
-		Dir:    td,
-		Daemon: filepath.Join(td, "tailscaled"+exe()),
-		CLI:    filepath.Join(td, "tailscale"+exe()),
+// CleanupBinaries cleans up any resources created by calls to BinaryDir, TailscaleBinary, or TailscaledBinary.
+// It should be called from TestMain after all tests have completed.
+func CleanupBinaries() {
+	buildOnce.Do(func() {})
+	if binDir != "" {
+		os.RemoveAll(binDir)
 	}
 }
 
-// buildMu limits our use of "go build" to one at a time, so we don't
-// fight Go's built-in caching trying to do the same build concurrently.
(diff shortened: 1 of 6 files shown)


josharian:
I clearly messed up the vms part of this PR, but I don't see why/how, and my attempts to get the vms test running on my local nixos box have failed (#2472 and #2473). I'll revisit this PR once those issues are fixed and I can run the vms tests locally...or if @Xe spots the bug in my code in the meantime. :)

josharian:
(depends on #3479)

tstest/integration/vms/vms_test.go
```diff
@@ -52,6 +54,17 @@ var (
 	}()
 )
 
+func TestMain(m *testing.M) {
+	flag.Parse()
+	err, cleanup := integration.BuildTestBinaries()
```
bradfitz, line 59: I'd rather `sync.Once` it so we don't need to build binaries until/if they're first needed.

josharian: Done

tstest/integration/integration.go
```diff
@@ -43,38 +43,39 @@ import (
...
+var (
+	binariesOnce sync.Once
+	binaries     Binaries
+)
+
+// TestBinaries returns tailscale and tailscaled binaries.
+// The returned binaries are valid for the duration of the test.
+func TestBinaries(tb testing.TB) Binaries {
+	binariesOnce.Do(func() {
+		bindir := tb.TempDir()
```
bradfitz, line 55: But TempDir is removed automatically when a test ends, no?

how does this work with two tests running in serial, both using this?

josharian: oh right. duh. that's why i did it that way. (that's what I get for letting the pr rot for so long.)

how did you want this to work with sync.Once, then? provide a global cleanup function that clients call from TestMain after running tests?

josharian: Did the global cleanup function thing. PTAL.

josharian:
Windows race test is in unrelated code. Appears to be a brand new test flake. (Sigh.)

josharian:
Ping

tstest/integration/integration.go
```diff
@@ -43,38 +43,67 @@ import (
...
-// Binaries are the paths to a tailscaled and tailscale binary.
-// These can be shared by multiple nodes.
-type Binaries struct {
-	Dir    string // temp dir for tailscale & tailscaled
-	Daemon string // tailscaled
-	CLI    string // tailscale
+// CleanupBinaries cleans up any resources created by calls to BinaryDir, TailscaleBinary, or TailscaledBinary.
+// It should be called from TestMain after all tests have completed.
+func CleanupBinaries() {
+	if binDir != "" {
```
bradfitz, line 49: do you want a `buildOnce.Do(func() {})` first?

bradfitz (approved):

tstest/integration/integration.go
josharian: Good idea. Done.

Issue #2443, opened 2021-07-16
util/deephash: ambiguous hashing when using AppendTo


dsnet:
In this snippet: https://github.com/tailscale/tailscale/blob/391207bbcf11171bd3fbf6a8667f94fea0b1e23f/util/deephash/deephash.go#L141-L147
we blindly use the `AppendTo` without any form of framing.

Hashing shares many properties with formatting/parsing. If the format used for hashing cannot be parsed in an unambiguous way, then it implies that two different values may hash to the same thing.

Here's an example of two different values with the same hash:
```go
type bytes []byte

func (p bytes) AppendTo(b []byte) []byte {
	return append(b, p...)
}

func main() {
	fmt.Printf("%x\n", deephash.Hash([]bytes{{}, {0, 0, 0, 0, 0, 0, 0, 1}})) // b716e801d6be55e4999e0f72dc570689b8b002775af6cdb2344103187aed5dbe
	fmt.Printf("%x\n", deephash.Hash([]bytes{{0, 0, 0, 0, 0, 0, 0, 1}, {}})) // b716e801d6be55e4999e0f72dc570689b8b002775af6cdb2344103187aed5dbe
}
```

A trivial way to fix this is to prepend the size to the `AppendTo` output.

\cc @bradfitz

josharian:
I'm not sure that prepending the size is a complete fix. You might still get collisions with other kinds of hashed output, exactly because AppendTo allows arbitrary output. We probably need a (type, length) tuple, where type is itself probably a (length, data) tuple.

(E.g. do we correctly distinguish between `uint16(1)` and `int16(1)` right now?)

The other question is whether we aim for correctness against adversarial inputs. The existing system is probably "good enough for government work" in that the likelihood of accidents is very low, albeit not as a low as the probability of a collision from the underlying sha256 hash function.

bradfitz:
I'd rather we do it correctly so we don't have to worry about whether future use cases are suitable for the "good enough" implementation.

dsnet:
The principle for collision-free hashes (assuming the `hash` itself is a sufficiently conflict-free hashing algorithm) is whether the format written to `hash.Write` can be unambiguously parsed. For a given Go type, every record written should either always be fixed size, or variable length with sufficient information to denote when the record terminates. If either condition is met, then unambigious parsing is possible since we know the Go type for every record written.

Other cases of ambiguous parsing:
[...]

PR #2470, opened 2021-07-20
util/deephash: improve cycle detection

dsnet:
The previous algorithm used a map of all visited pointers.
The strength of this approach is that it quickly prunes any nodes
that we have ever visited before. The detriment of the approach
is that pruning is heavily dependent on the order that pointers
were visited. This is especially relevant for hashing a map
where map entries are visited in a non-deterministic manner,
which would cause the map hash to be non-deterministic
(which defeats the point of a hash).

This new algorithm uses a stack of all visited pointers,
similar to how github.com/google/go-cmp performs cycle detection.
When we visit a pointer, we push it onto the stack, and when
we leave a pointer, we pop it from the stack.
Before visiting a pointer, we first check whether the pointer exists
anywhere in the stack. If yes, then we prune the node.
The detriment of this approach is that we may hash a node more often
than before since we do not prune as aggressively.

The set of visited pointers up until any node is only the
path of nodes up to that node and not any other pointers
that may have been visited elsewhere. This provides us
deterministic hashing regardless of visit order.
We can now delete hashMapFallback and associated complexity,
which only exists because the previous approach was non-deterministic
in the presence of cycles.

This fixes a failure of the old algorithm where obviously different
values are treated as equal because the pruning was too aggresive.
See https://github.com/tailscale/tailscale/issues/2443#issuecomment-883653534

The new algorithm is slightly slower since it prunes less aggresively:
```
Hash-8              66.1µs ± 1%    68.8µs ± 1%   +4.09%        (p=0.000 n=19+19)
HashMapAcyclic-8    63.0µs ± 1%    62.5µs ± 1%   -0.76%        (p=0.000 n=18+19)
TailcfgNode-8       9.79µs ± 2%    9.88µs ± 1%   +0.95%        (p=0.000 n=19+17)
HashArray-8          643ns ± 1%     653ns ± 1%   +1.64%        (p=0.000 n=19+19)
```
However, a slower but more correct algorithm seems
more favorable than a faster but incorrect algorithm.


util/deephash/deephash.go
```diff
@@ -30,14 +30,19 @@ type hasher struct {
 	h       hash.Hash
 	bw      *bufio.Writer
 	scratch [scratchSize]byte
-	visited map[uintptr]bool
+
+	// visitStack is a stack of pointers visited.
+	// Pointers are pushed onto the stack when visited, and popped when leaving.
+	// The integer value is the depth at which the pointer was visited.
+	// The length of this stack should be zero after ever hashing operation.
```
josharian, line 37: every

util/deephash/deephash.go
```diff
@@ -30,14 +30,19 @@ type hasher struct {
...
+	visitStack map[uintptr]int
```
josharian, line 38: Using a uintptr as the key here will break if there's a moving GC. Maybe add:

```go
import _ "go4.org/unsafe/assume-no-moving-gc"
```

Pointers change today on stack copies, but I think we're saved from that because anything that gets passed to a hasher will get heap-allocated today. But I'm a bit nervous that future escape analysis improvements could change that, in which case recursive print calls could cause the stack to grow, thereby breaking cycle detection.

I'd feel more comfortable if we used unsafe.Pointer here so we didn't have to worry about that, assuming we can find a safe way to get an unsafe.Pointer out of package reflect.

util/deephash/deephash.go
```diff
@@ -175,48 +183,33 @@ func (h *hasher) print(v reflect.Value) (acyclic bool) {
...
-			if !h.print(v.Index(i)) {
-				acyclic = false
-			}
+			h.print(v.Index(i))
 		}
-		return acyclic
 	case reflect.Interface:
-		return h.print(v.Elem())
+		h.print(v.Elem())
 	case reflect.Map:
```
josharian, line 200: Check for nil and write `0` in that case, to mirror the pointer code above?

dsnet: Future update to code :)

util/deephash/deephash.go

util/deephash/deephash.go
dsnet: I prefer this be an `unsafe.Pointer`, but I can understand the aversion towards using `unsafe`.

josharian: I'd like this to be an unsafe.Pointer too. Sorry if that didn't come through clearly above. @bradfitz you OK with that?

josharian:
depaware is https://github.com/tailscale/depaware/. Install, run `depaware -update ./cmd/tailscale{,d}`.

The code LGTM, but I'd like to get resolution on using unsafe.Pointer (and ideally start using it) before submitting.

util/deephash/deephash.go
bradfitz: yeah, I'm fine with that.

util/deephash/deephash.go
```diff
@@ -120,48 +127,54 @@ func (h *hasher) int(i int) {
...
+			return
+		}
+
+		// Check for cycle.
 		ptr := v.Pointer()
-		if visited[ptr] {
-			return false
+		if idx, ok := h.visitStack[ptr]; ok {
+			w.WriteByte(2) // indicates cycle
+			h.uint(uint64(idx))
```
bradfitz, line 163: perfect. this is what I was wanting, encoding a reference to a previous thing.

bradfitz (approved):

util/deephash/deephash.go
dsnet: Alright, I switched it to using `unsafe.Pointer`. I created a `pointer` wrapper type to reduce that number of references to `unsafe.Pointer` and document that we only depend on it for comparability and not something more sketchy.

util/deephash/deephash.go
```diff
@@ -306,33 +292,40 @@ func (h *hasher) hashMapAcyclic(v reflect.Value) (acyclic bool) {
...
+	delete(v, p)
+}
+
+// pointer is a thin wrapper over unsafe.Pointer.
+// We only rely on comparability of pointers; we cannot rely on uintptr since
+// that would break if Go ever switched to a moving GC.
+type pointer struct{ p unsafe.Pointer }
+
+func pointerOf(v reflect.Value) pointer {
+	return pointer{unsafe.Pointer(v.Pointer())}
```
josharian, line 330: Is it safe to do this? This is converting from a uintptr to an unsafe.Pointer, which is not on the list of allowed unsafe.Pointer operations. Does checkptr complain?

dsnet: > which is not on the list of allowed unsafe.Pointer operations.

On https://pkg.go.dev/unsafe#Pointer, I see:

> (5) Conversion of the result of reflect.Value.Pointer or reflect.Value.UnsafeAddr from uintptr to Pointer.

Doesn't that cover this case?

josharian: Yes! Thanks. :) LGTM.

PR #2474, opened 2021-07-20
tstest/integration: add upnp test

JulianKnodt:
Adds a upnp integration test which ensures that portmapping gets called.
It's a bit sketchy since I didn't want to depend on miniupnp, but it checks that the flow is correct.


tstest/integration/integration_test.go
```diff
@@ -795,7 +830,33 @@ func (tt *trafficTrap) Err() error {
...
+var rootDesc []byte
+
+//go:embed testdata/addAnyPortMapping.xml
+var addAnyPortMapping []byte
+
+//go:embed testdata/getExternalIP.xml
+var getExternalIP []byte
+
 func (tt *trafficTrap) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.URL.EscapedPath() == "/rootDesc.xml" {
```
bradfitz, line 843: worth some comments about what this is? Or better, pull this into a method with a nice name so it's obvious what it is.

```go
     if tt.serveUPnP(w, r) {
            return
     }

// serveUPnP acts like a UPnP gateway device.
// It reports whether the request r was handled. (That is, whether r looked
// like a UPnP query)
func (tt *trafficTrap) serveUPnP(...) (handled bool) {
    
    return false
}

File: tstest/integration/integration_test.go [hunk 2]
```diff
@@ -359,6 +361,38 @@ func TestNoControlConnectionWhenDown(t *testing.T) {
 	d2.MustCleanShutdown(t)
 }
 
+func TestUPnPPortMapping(t *testing.T) {
+	// t.Skip()
```
bradfitz, line 365: delete

tstest/integration/integration_test.go
```diff
@@ -359,6 +361,38 @@ func TestNoControlConnectionWhenDown(t *testing.T) {
...
+
+	t.Parallel()
+	bins := BuildTestBinaries(t)
+
+	env := newTestEnv(t, bins)
+	defer env.Close()
+
+	n1 := newTestNode(t, env)
+
+	os.Setenv("TS_DISABLE_UPNP", "false")
```
bradfitz, line 375: this doesn't work in a t.Parallel

JulianKnodt:
Closing, the IGD tests cover the same code

PR #2475, opened 2021-07-20
net/portmapper: return correct upnp port

JulianKnodt:
Previously, this was incorrectly returning the internal port, and using that with the external
exposed IP when it did not use WANIPConnection2. In the case when we must provide a port, we
return it instead.

Noticed this while implementing the integration test for upnp.

### DIFF
--- a/net/portmapper/upnp.go
+++ b/net/portmapper/upnp.go
@@ -7,6 +7,7 @@ package portmapper
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"net/url"
 	"time"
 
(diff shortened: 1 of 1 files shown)


net/portmapper/upnp.go
```diff
@@ -86,6 +86,7 @@ const tsPortMappingDesc = "tailscale-portmap"
 // addAnyPortMapping abstracts over different UPnP client connections, calling the available
 // AddAnyPortMapping call if available for WAN IP connection v2, otherwise defaulting to the old
 // behavior of calling AddPortMapping with port = 0 to specify a wildcard port.
+// It returns the new external port, or an error.
```
bradfitz, line 89: why does it just return something the caller already passed in?

why not just return an error?

bradfitz: Wait, the caller passes `prevPort` for this argument.

How does this even work? `prevPort` will usually be 0 on startup.

JulianKnodt: It can either return something the caller passed in, or can return an entirely new port from the AddAnyPortMapping. I've updated the docs for that

bradfitz (approved):
Ah! Totally missed that, thanks. I see now.

JulianKnodt:
@bradfitz I'm not entirely sure if 0 is handled correctly though, you are right in that case, since it will just return 0 and the IP. If that's the case, should we just roll for a random port and try portmapping it?

PR #2476, opened 2021-07-20
util/deephash: include type as part of hash for interfaces

dsnet:
A Go interface may hold any number of different concrete types.
Just because two underlying value hashes to the same thing
does not mean two values are identical if they have different
concrete types. As such, include the type in the hash.

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -189,7 +189,15 @@ func (h *hasher) print(v reflect.Value) (acyclic bool) {
 		}
 		return acyclic
 	case reflect.Interface:
-		return h.print(v.Elem())
+		if v.IsNil() {
+			w.WriteByte(0) // indicates nil
+			return true
+		}
+		v = v.Elem()
+
+		w.WriteByte(1) // indicates visiting interface value
+		h.hashType(v.Type())
+		return h.print(v)
 	case reflect.Map:
 		// TODO(bradfitz): ideally we'd avoid these map
 		// operations to detect cycles if we knew from the map
@@ -335,4 +343,44 @@ func (h *hasher) hashMapFallback(v reflect.Value) (acyclic bool) {
 	}
 	w.WriteString("}\n")
 	return acyclic
+}
+
+func (h *hasher) hashType(t reflect.Type) {
+	// Unwrap anonymous pointers to named types since they are so common.
+	name := t.Name()
+	isAnonPtr := name == "" && t.Kind() == reflect.Ptr
+	if isAnonPtr {
+		t = t.Elem()
+		name = t.Name()
+	}
+
+	// Format the qualified name of the type.
+	if name != "" {
+		size := h.scratch[:8]
+		qualifiedName := h.scratch[8:8] // reserve space for size
+		if isAnonPtr {
+			qualifiedName = append(qualifiedName, '*')
+		}
+		if pkgPath := t.PkgPath(); pkgPath != "" {
+			qualifiedName = strconv.AppendQuote(qualifiedName, pkgPath)
+		}
+		qualifiedName = append(qualifiedName, '.')
+		qualifiedName = append(qualifiedName, name...)
+		binary.LittleEndian.PutUint64(size, uint64(len(qualifiedName)))
+		h.bw.Write(size)
+		h.bw.Write(qualifiedName)
+		return
+	}
+
+	// For unnamed types, we cannot easily hash the type.
+	// Hash the address of the *reflect.rtype, which we assume to be stable
+	// within the lifetime of a program.
+	if isAnonPtr {
+		t = reflect.PtrTo(t)
+	}
+	rtypePtr := reflect.ValueOf(t).Pointer() // address of *reflect.rtype
+	record := h.scratch[:16]
+	binary.LittleEndian.PutUint64(record[:8], math.MaxUint64) // value to distinguish it from size field for qualified name
+	binary.LittleEndian.PutUint64(record[8:], uint64(rtypePtr))
+	h.bw.Write(record)
 }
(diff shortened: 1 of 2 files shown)


util/deephash/deephash.go
```diff
@@ -336,3 +344,43 @@ func (h *hasher) hashMapFallback(v reflect.Value) (acyclic bool) {
...
+	name := t.Name()
+	isAnonPtr := name == "" && t.Kind() == reflect.Ptr
+	if isAnonPtr {
+		t = t.Elem()
+		name = t.Name()
+	}
+
+	// Format the qualified name of the type.
+	if name != "" {
+		size := h.scratch[:8]
```
josharian, line 359: I think we could use 4 (uint32) here.

util/deephash/deephash.go
```diff
@@ -336,3 +344,43 @@ func (h *hasher) hashMapFallback(v reflect.Value) (acyclic bool) {
...
+		qualifiedName := h.scratch[8:8] // reserve space for size
+		if isAnonPtr {
+			qualifiedName = append(qualifiedName, '*')
+		}
+		if pkgPath := t.PkgPath(); pkgPath != "" {
+			qualifiedName = strconv.AppendQuote(qualifiedName, pkgPath)
```
josharian, line 365: Do we need to quote the string? ISTM that `append(qualifiedName, pkgPath...)` would suffice.

util/deephash/deephash.go
```diff
@@ -336,3 +344,43 @@ func (h *hasher) hashMapFallback(v reflect.Value) (acyclic bool) {
...
+		}
+		qualifiedName = append(qualifiedName, '.')
```
josharian, line 367: Only append `.` if there's a pkgPath?

util/deephash/deephash.go
```diff
@@ -336,3 +344,43 @@ func (h *hasher) hashMapFallback(v reflect.Value) (acyclic bool) {
...
+		qualifiedName = append(qualifiedName, name...)
+		binary.LittleEndian.PutUint64(size, uint64(len(qualifiedName)))
+		h.bw.Write(size)
```
josharian, line 370: It probably doesn't matter for performance, but you could do this in a single write with a bit of refactoring above.

util/deephash/deephash_test.go
```diff
@@ -21,6 +22,33 @@ import (
 	"tailscale.com/wgengine/wgcfg"
 )
 
+func TestHash(t *testing.T) {
```
josharian, line 25: Hurray, tests!

util/deephash/deephash.go
```diff
@@ -336,3 +344,43 @@ func (h *hasher) hashMapFallback(v reflect.Value) (acyclic bool) {
 	w.WriteString("}\n")
 	return acyclic
 }
+
+func (h *hasher) hashType(t reflect.Type) {
```
josharian, line 348: I wonder whether we oughtn't have a call to hashType at the beginning of print. Maybe that's coming up in a future change. :)

josharian (commented):
It's nice to have you here, Joe!

josharian (approved):

util/deephash/deephash.go
dsnet: Got rid of this approach.

util/deephash/deephash.go
dsnet: Got rid of this approach.

util/deephash/deephash.go
dsnet: Got rid of this approach.

util/deephash/deephash.go
dsnet: Got rid of this approach.

dsnet:
I wasn't comfortable with hashing the `*reflect.rtype` address. Since `reflect.Type` is comparable and the set of `reflect.Type` in a program is likely fixed, I just created a global mapping of `reflect.Type` to some arbitrary and unique integer.

josharian:
We’ve gone to some pains to avoid un-GC-able memory. Let’s discuss some other options here.

dsnet:
Another alternative is to use `go4.org/intern` and hash the address of the `*intern.Value`, but then we would be relying on Go not having a moving GC.

It kinda stinks that you can't get a unique number or string for a `reflect.Type`. I don't care what the value is, I just need it to be unique. We had the same problem in `go-cmp` and we don't solve it correctly there.

josharian:
Hashing the intern.Value won’t keep it alive, so that won’t provide a stable hash. I think we should use the reflect.rtype, myself. It is exactly what we want, and we are lucky enough to have enough Go expertise on the team to be able to adjust as necessary. Particularly if we can write some tests that will likely fail if the undocumented invariants we rely on change. We could also use build tags next to some docs to force someone to consciously certify that the code still is valid.

Yet another alternative is to file a proposal for new API to fix this properly. Then while it goes through the process and we wait for it to show up in a release, we add it in our own fork. We’ve done this for other APIs.

util/deephash/deephash.go
dsnet: I don't think we should call it at the beginning of _every_ `print`, but we might want to consider doing it for the very top-level concrete type.

At present, all `deephash` usages in the code base always pass the same concrete type at the top-level, so it's not a big deal.

dsnet:
And we're back with the `*reflect.rtype` approach.

util/deephash/deephash.go
josharian: > we might want to consider doing it for the very top-level concrete type.

...and every interface we encounter, because the concrete types there might vary from call to call?

josharian (approved):
Btw, we do a squash (as appropriate) and rebase to keep a clean, linear history...in short, we emulate gerrit over GitHub.

PR #2479, opened 2021-07-21
Chore: Add more helpful error message

chakflying:
Version mismatch occurs on ArchLinux if I updated Tailscale but didn't reboot.

### DIFF
--- a/ipn/message.go
+++ b/ipn/message.go
@@ -241,7 +241,7 @@ func (bc *BackendClient) GotNotifyMsg(b []byte) {
 		log.Fatalf("BackendClient.Notify: cannot decode message (length=%d, %#q): %v", len(b), b, err)
 	}
 	if n.Version != version.Long && !bc.AllowVersionSkew {
-		vs := fmt.Sprintf("GotNotify: Version mismatch! frontend=%#v backend=%#v",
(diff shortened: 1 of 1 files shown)


ipn/message.go
```diff
@@ -241,7 +241,7 @@ func (bc *BackendClient) GotNotifyMsg(b []byte) {
 		log.Fatalf("BackendClient.Notify: cannot decode message (length=%d, %#q): %v", len(b), b, err)
 	}
 	if n.Version != version.Long && !bc.AllowVersionSkew {
-		vs := fmt.Sprintf("GotNotify: Version mismatch! frontend=%#v backend=%#v",
+		vs := fmt.Sprintf("GotNotify: Version mismatch! Please try to reboot if you just updated Tailscale. frontend=%#v backend=%#v",
```
DentonGentry, line 244: Package install is expected to stop the running tailscaled and start the new version immediately.

We do not require a reboot to update the tailscale package, and don't want to print a message telling people to do so.

DentonGentry:
If you send email to support@tailscale.com with the IP address of the system where this happened, we might be able to help figure out why tailscaled did not update during package install.

Printing a message telling people to reboot is not something we want to do.

PR #2480, opened 2021-07-21
tstest/integration/vms: test DNS configuration

Xe:
This uses a neat little tool to dump the output of DNS queries to
standard out. This is the first end-to-end test of DNS that runs against
actual linux systems. The /etc/resolv.conf test may look superflous,
however this will help for correlating system state if one of the DNS
tests fails.

Depends on #2491 and #2492

### DIFF
--- a/tstest/integration/testcontrol/testcontrol.go
+++ b/tstest/integration/testcontrol/testcontrol.go
@@ -43,6 +43,7 @@ type Server struct {
 	DERPMap     *tailcfg.DERPMap // nil means to use prod DERP map
 	RequireAuth bool
 	Verbose     bool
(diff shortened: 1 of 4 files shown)


tstest/integration/testcontrol/testcontrol.go
```diff
@@ -655,6 +657,7 @@ func (s *Server) MapResponse(req *tailcfg.MapRequest) (res *tailcfg.MapResponse,
 		Debug: &tailcfg.Debug{
 			DisableUPnP: "true",
 		},
+		DNSConfig: s.DNSConfig,
```
Xe, line 660: Is this the right way to do it?

PR #2483, opened 2021-07-21
util/deephash: disambiguate hashing of AppendTo

### LINKED ISSUE #2443: util/deephash: ambiguous hashing when using AppendTo

dsnet:
Prepend size to AppendTo output.

Fixes #2443

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -147,8 +147,10 @@ func (h *hasher) print(v reflect.Value) (acyclic bool) {
 		// Use AppendTo methods, if available and cheap.
 		if v.CanAddr() && v.Type().Implements(appenderToType) {
 			a := v.Addr().Interface().(appenderTo)
-			scratch := a.AppendTo(h.scratch[:0])
-			w.Write(scratch)
+			size := h.scratch[:8]
+			record := a.AppendTo(size)
+			binary.LittleEndian.PutUint64(record, uint64(len(record)-len(size)))
+			w.Write(record)
 			return true
 		}
 	}
--- a/util/deephash/deephash_test.go
+++ b/util/deephash/deephash_test.go
@@ -22,6 +22,12 @@ import (
 	"tailscale.com/wgengine/wgcfg"
 )
 
+type appendBytes []byte
+
+func (p appendBytes) AppendTo(b []byte) []byte {
+	return append(b, p...)
+}
+
 func TestHash(t *testing.T) {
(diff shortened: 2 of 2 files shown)


bradfitz (approved):

util/deephash/deephash.go
```diff
@@ -147,8 +147,10 @@ func (h *hasher) print(v reflect.Value) (acyclic bool) {
 		// Use AppendTo methods, if available and cheap.
 		if v.CanAddr() && v.Type().Implements(appenderToType) {
 			a := v.Addr().Interface().(appenderTo)
-			scratch := a.AppendTo(h.scratch[:0])
-			w.Write(scratch)
+			size := h.scratch[:8]
+			record := a.AppendTo(size)
+			binary.LittleEndian.PutUint64(record, uint64(len(record)-len(size)))
```
josharian, line 152: You don't really need to subtract len(size) here, because all that matters is that all lengths are distinct.

At which point you could simplify a little bit by eliminating the `size` local variable, which I think is a net readability win because it reduces the temptation to think that we should be writing the size to `size`.

```go
record := a.AppendTo(h.scratch[:8]) // save space at the beginning of record
binary.LittleEndian.PutUint64(record, uint64(len(record)))
w.Write(record)
```

josharian (approved):

PR #2486, opened 2021-07-22
tstime/mono: add new package, use everywhere

josharian:
We call time.Now in 4-6 places per packet.
At 50+ns per call, it adds up enough to show up in pprof.

time.Now is twice as expensive as we need,
because it gets both wall time and monotonic time.
We need only monotonic time.

This PR introduces tstime/mono.Time and
converts all but one per-packet time.Now call
to use it. (The "but one" is in wireguard-go.)

tstime.Mono is about 50% faster than time.Now,
at the cost of using a go:linkname to grab
the runtime's internal monotonic clock routine.

This shrinks the "current time" pprof box,
but doesn't eliminate it.

The next step here is to decide which of our
uses of time would accept a granularity of 1-4ms,
and switch them to use a coarse monotonic clock.
That's faster yet than the regular monotonic clock,
but it also means making our own copies of
the package runtime assembly that gets the time.


crawshaw (commented):
General strategy looks good. It's somewhat unfortunate how big a diff is created by `s/time.Time/tstime.Mono/`, but I think it's for the best that we don't do something tricky like a rename import to keep using time.Time.

A quick survey of these uses of time.Now suggest they could all be replaced with a coarse variant.

bradfitz:
Bikeshed: package `tstime/mono` (or even something outside of `tailscale.com` entirely) and then `mono.Time` as the type? Same length as `time.Time`.

josharian:
> Bikeshed: package tstime/mono (or even something outside of tailscale.com entirely) and then mono.Time as the type? Same length as time.Time.

Done.

> A quick survey of these uses of time.Now suggest they could all be replaced with a coarse variant.

I recall seeing one (DERP timing?) that appeared to want millisecond resolution. Maybe that'd be OK with coarse time? Not sure. I'll look again.

Once this is in (after the 1.12 release!), I'll pull in my coarse time implementations and do a bit more retrofitting. I plan to use the same type (mono.Time) for coarse time, with the only difference being calling `mono.CoarseNow` and `mono.CoarseSince` instead of `mono.Now` and `mono.Since`, so the diff should be much smaller.


Now to figure out why tests pass locally but not on CI...

josharian:
Ready for review at last.

ipn/ipnstate/ipnstate.go
```diff
@@ -90,7 +91,7 @@ type PeerStatus struct {
 	RxBytes       int64
 	TxBytes       int64
 	Created       time.Time // time registered with tailcontrol
-	LastWrite     time.Time // time last packet sent
+	LastWrite     mono.Time // time last packet sent
```
bradfitz, line 94: How does this serialize in JSON in `tailscale status --json` ?

josharian: Badly. I've changed it to format as a time.Time, for backwards compatibility. Of course, if the wall clock changes, it won't be responsive to that change.

net/tstun/wrap.go
```diff
@@ -64,7 +65,7 @@ type Wrapper struct {
 
 	closeOnce sync.Once
 
-	lastActivityAtomic int64 // unix seconds of last send or receive
+	lastActivity mono.Time // time of last send or receive, accessed atomically
```
bradfitz, line 68: keep the old variable name? the Atomic suffix is ugly and scary which is a feature.

I guess you didn't like the stutter in `lastActivityAtomic.LoadAtomic` etc? I'm fine with it if you are.

tstime/mono/mono.go
```diff
@@ -0,0 +1,113 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package mono provides fast monotonic time.
```
bradfitz, line 5: Maybe flesh out this package comment a bit with some numbers and a big scary warning that it should almost never be used and std time is plenty fast?

bradfitz (approved):

tstime/rate/rate.go
```diff
@@ -0,0 +1,89 @@
...
+// bursts of at most b tokens.
+func NewLimiter(r Limit, b int) *Limiter {
+	if b < 1 {
+		panic("bad burst, must be at least 1")
+	}
+	return &Limiter{limit: r, burst: float64(b)}
+}
+
+// AllowN reports whether an event may happen now.
+func (lim *Limiter) Allow() bool {
```
crawshaw, line 59: If you plan to add a `mono.NowCoarse`, then I think it's worth making this Allow take a `now` parameter, so the caller can choose if they want coarse or fine-grained timing on their token bucket

crawshaw (approved):

net/tstun/wrap.go
josharian: Yeah, I was removing the stutter, but don't feel strongly. Re-added.

tstime/mono/mono.go
josharian: Done

tstime/rate/rate.go
josharian: Ack, although I'd rather do that as part of adding coarse mono time.

PR #2489, opened 2021-07-22
Redo controlclient backtrace scrub without using regex

DentonGentry:
This is an attempt to remove regex from iOS.
Please do feel free to do so differently if desired, I wanted to try it.

### DIFF
--- a/control/controlclient/debug.go
+++ b/control/controlclient/debug.go
@@ -11,10 +11,11 @@ import (
 	"fmt"
 	"log"
 	"net/http"
-	"regexp"
 	"runtime"
 	"strconv"
+	"strings"
 	"time"
+	"unicode"
 )
 
 func dumpGoroutinesToURL(c *http.Client, targetURL string) {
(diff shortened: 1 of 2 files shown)


wgengine/netstack/netstack.go
```diff
@@ -145,25 +163,8 @@ func (ns *Impl) Start() error {
...
-		if !ok {
-			ns.logf("netstack: could not parse local address %s for incoming TCP connection", ip)
-			return false
-		}
-		if !ns.isLocalIP(ip) {
-			ns.addSubnetAddress(pn, ip)
-		}
-		return tcpFwd.HandlePacket(tei, pb)
-	})
-	ns.ipstack.SetTransportProtocolHandler(udp.ProtocolNumber, udpFwd.HandlePacket)
```
JulianKnodt, line 166: are these changes related to the removing the regex?

control/controlclient/debug.go
```diff
@@ -42,28 +43,51 @@ func dumpGoroutinesToURL(c *http.Client, targetURL string) {
 	}
 }
 
-var reHexArgs = regexp.MustCompile(`\b0x[0-9a-f]+\b`)
+func isHexString(s string) bool {
+	if len(s) <= 2 || s[0:2] != "0x" {
+		return false
+	}
+	_, err := strconv.ParseUint(s[2:], 16, 64)
```
JulianKnodt, line 50: One thing that worries me about this is that it may overflow if the hex string is big enough. Does that matter? I'm not sure if opting for a for loop which checks every character would be better, but it'd be more explicit. I am also curious whether capital hex such as 0xFF is valid, which wasn't previously accepted?

wgengine/netstack/netstack.go
DentonGentry: Oops: I must not have gone back to main before making my branch, these are Brad's changes for a different bug which I had patched in to test.

I re-made the branch and pushed again.

control/controlclient/debug.go
```diff
@@ -42,28 +43,51 @@ func dumpGoroutinesToURL(c *http.Client, targetURL string) {
 	}
 }
 
-var reHexArgs = regexp.MustCompile(`\b0x[0-9a-f]+\b`)
+func isHexString(s string) bool {
+	if len(s) <= 2 || s[0:2] != "0x" {
```
bradfitz, line 47: drop the 0

control/controlclient/debug.go
```diff
@@ -42,28 +43,51 @@ func dumpGoroutinesToURL(c *http.Client, targetURL string) {
...
+	if unicode.IsLetter(r) || unicode.IsNumber(r) {
+		return false
+	}
+	return true
+}
+
+func scrubGoroutineDump(buf []byte) []byte {
+	saw := map[string]string{} // "0x123" => "v1%3" (unique value 1 and its value mod 8)
+	sbuf := string(buf)
+	for _, field := range strings.FieldsFunc(sbuf, separator) {
```
bradfitz, line 64: I think this allocates too much, making this whole array. Ideally we'd do this without an additional pass over the data.

control/controlclient/debug.go
```diff
@@ -42,28 +43,51 @@ func dumpGoroutinesToURL(c *http.Client, targetURL string) {
...
+func separator(r rune) bool {
+	if unicode.IsLetter(r) || unicode.IsNumber(r) {
+		return false
+	}
+	return true
+}
+
+func scrubGoroutineDump(buf []byte) []byte {
+	saw := map[string]string{} // "0x123" => "v1%3" (unique value 1 and its value mod 8)
+	sbuf := string(buf)
```
bradfitz, line 63: I'd like to avoid this allocation too. The input []byte is already quite big and this doubles it.

control/controlclient/debug.go
DentonGentry: Added a test for capitalized hex numbers, even if they were allowed before we want to suppress them.

As it is a backtrace from the go runtime, I don't think it can be bigger than 64 bits. We don't run on machines with wider integers than that.

If a very wide hex number appears, the worst that happens is we don't suppress it.

DentonGentry:
As we have other dependencies on regex, pausing work on this.

bradfitz:
Sent https://github.com/tailscale/tailscale/pull/2494 to not do extra allocs. (Actually, it's more efficient than the old regexp version alloc-wise)

PR #2493, opened 2021-07-22
disable vm tests on every commit to main

Xe:
This experiment apparently failed.


DentonGentry:
I wouldn't say it failed, and if we do really believe that the problem is mostly nixos-unstable then it would be fine to leave the rest in for a while and see. If there are still flakes after that, then yes we'd want to take it out of the regular CI until we can address it.

bradfitz (approved):

PR #2497, opened 2021-07-22
tstime: add Append3339Nano

josharian:
And use it in log tail.

This eliminates the use of time.Format.
It shaves off ~9k from the iOS binary size.


tstime/tstime.go
```diff
@@ -141,3 +142,71 @@ func Parse3339(s string) (time.Time, error) {
 func Parse3339B(b []byte) (time.Time, error) {
 	return parse3339m(mem.B(b))
 }
+
+func Append3339Nano(t time.Time, b []byte) []byte {
```
dsnet, line 146: Nit: perhaps AppendRFC3339Nano to make clean it's a specialized form of `time.RFC3339Nano`?

tstime/tstime.go
```diff
@@ -141,3 +142,71 @@ func Parse3339(s string) (time.Time, error) {
...
+		b = append(b, '+')
+	}
+	zoneOffset /= 60
+	b = appendInt(b, int64(zoneOffset/60), 2)
+	b = append(b, ':')
+	b = appendInt(b, int64(zoneOffset%60), 2)
+	return b
+}
+
+func appendInt(b []byte, n int64, width int) []byte {
```
dsnet, line 181: Aside: I've seen wrapper logic to `strconv.AppendUint` to ensure a fixed width many times that i wonder if the strconv package itself should support this.

dsnet (approved):

tstime/tstime.go
josharian: Yeah, that'd be nice.

josharian:
Turns out that `-ldflags=-dumpdep` only shows you a single dependency. After cutting out this one...there are more. Plenty more. time.AppendFormat isn't going anywhere.

PR #2499, opened 2021-07-22
net/portmapper: Add discovery to upnp

JulianKnodt:
Previously, it was just a brittle check for a specific port on my own network.
It seems that routers are free to pick whatever port they'd like, so if we'd really like to
support any router it's necessary to have discovery. This is the smallest possible change
to add it in.

This makes upnp work on Denton's box, so hopefully it should enable more uses.

### DIFF
--- a/net/portmapper/upnp.go
+++ b/net/portmapper/upnp.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"math/rand"
 	"net/url"
+	"sync"
 	"time"
 
 	"github.com/tailscale/goupnp/dcps/internetgateway2"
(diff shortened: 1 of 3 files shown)


net/portmapper/upnp.go
```diff
@@ -169,11 +192,63 @@ func getUPnPClient(ctx context.Context, gw netaddr.IP) (upnpClient, error) {
...
 	}()
 
-	select {
-	case client := <-clients:
-		return client, nil
-	case <-ctx.Done():
-		return nil, ctx.Err()
+	// In addition to pinging specific endpoints, if we are able to find any
+	discoClients := make(chan upnpClient, 3)
+	var total int32 = 6
```
bradfitz, line 197: what is 6? I see 3 below.

net/portmapper/upnp.go
```diff
@@ -169,11 +192,63 @@ func getUPnPClient(ctx context.Context, gw netaddr.IP) (upnpClient, error) {
...
+	var done int32
+	discoCtxt, discoCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	go func() {
+		var err error
+		ip1Clients, _, err := internetgateway2.NewWANIPConnection1Clients(discoCtxt)
+		if err == nil && len(ip1Clients) > 0 {
+			discoClients <- ip1Clients[0]
+		}
+		if atomic.AddInt32(&done, 1) == total {
```
bradfitz, line 206: can we use some different mechanism? this is a little magic and frail. why do we even need to close discoClients at all?

bradfitz:
Too late for 1.12. We'll have to make port discover better for 1.14. Can we instead use the old code we used in 1.10 and earlier?

JulianKnodt:
@bradfitz 1.10 and earlier I think also hardcoded a port, so I don't think it's any better than the current implementation, but it will hit a different port so you can leave it in if that helps

bradfitz:
No longer applicable I believe. Let me know if there are parts of this you'd want to rebase and get in.

JulianKnodt:
Should be good? I think it does the same thing as what's currently in but less efficiently

PR #2502, opened 2021-07-23
util/deephash: hash uint{8,16,32,64} explicitly

dsnet:
Instead of hashing the humanly formatted forms of a number,
hash the native machine bits of the integers themselves.

There is a small performance gain for this:

	Hash-8              75.7µs ± 1%    76.0µs ± 2%    ~            (p=0.315 n=10+9)
	HashMapAcyclic-8    63.1µs ± 3%    61.3µs ± 1%  -2.77%        (p=0.000 n=10+10)
	TailcfgNode-8       10.3µs ± 1%    10.2µs ± 1%  -1.48%        (p=0.000 n=10+10)
	HashArray-8         1.07µs ± 1%    1.05µs ± 1%  -1.79%        (p=0.000 n=10+10)

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -21,7 +21,6 @@ import (
 	"hash"
 	"math"
 	"reflect"
-	"strconv"
 	"sync"
 	"time"
 	"unsafe"
@@ -91,8 +90,8 @@ func Hash(v interface{}) (s Sum) {
 	once.Do(func() {
 		seed = uint64(time.Now().UnixNano())
 	})
-	h.uint(seed)
-	h.print(reflect.ValueOf(v))
+	h.hashUint64(seed)
+	h.hashValue(reflect.ValueOf(v))
 	return h.sum()
 }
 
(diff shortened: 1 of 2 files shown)


dsnet:
Approximate increase of 28 lines of source code and ~1KiB in compile size. This is offset by #2502

util/deephash/deephash.go
```diff
@@ -116,19 +115,25 @@ type appenderTo interface {
...
-	binary.BigEndian.PutUint64(h.scratch[:8], i)
-	h.bw.Write(h.scratch[:8])
+func (h *hasher) hashUint8(i uint8) {
+	h.bw.WriteByte(i)
 }
-
-func (h *hasher) int(i int) {
-	binary.BigEndian.PutUint64(h.scratch[:8], uint64(i))
+func (h *hasher) hashUint16(i uint16) {
+	binary.BigEndian.PutUint16(h.scratch[:2], i)
```
josharian, line 122: Nit: Let's use LittleEndian by default in cases in which it doesn't matter. Our most active platforms are all little-endian.

josharian (approved):

util/deephash/deephash.go
dsnet: Done.

Issue #2506, opened 2021-07-23
cmd/derper: skipping expvar "derp_average_queue_duration_ms" with undeclared Prometheus type


bradfitz:
fe54721e3 added a new `average_queue_duration_ms` metric, but varz says:

```
# skipping expvar "derp_average_queue_duration_ms" (Go type expvar.Func returning float64) with undeclared Prometheus type
```

/cc @JulianKnodt

PR #2507, opened 2021-07-23
tsweb: add case for f64 in logging

JulianKnodt:
A previously added metric which was float64 was being ignored in tsweb, because it previously
only accepted int64 and ints. It can be handled in the same way as ints.

Fixes: #2506

### DIFF
--- a/tsweb/tsweb.go
+++ b/tsweb/tsweb.go
@@ -409,7 +409,7 @@ func VarzHandler(w http.ResponseWriter, r *http.Request) {
 		case expvar.Func:
 			val := v()
 			switch val.(type) {
-			case int64, int:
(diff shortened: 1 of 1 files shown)


bradfitz:
Can you instead update tsweb/tsweb.go? (and thus keep the float type)

```go
switch v := kv.Value.(type) {
case expvar.Func:
        val := v()
        switch val.(type) {
        case int64, int:
                fmt.Fprintf(w, "# TYPE %s %s\n%s %v\n", name, typ, name, val)
        default:
                fmt.Fprintf(w, "# skipping expvar func %q returning unknown type %T\n", name, val)
        }


```

bradfitz (approved):

PR #2513, opened 2021-07-26
wgengine: re-set DNS config on Linux after a major link change

bradfitz:
Updates #2458 (maybe fixes it)

### DIFF
--- a/wgengine/userspace.go
+++ b/wgengine/userspace.go
@@ -110,6 +110,7 @@ type userspaceEngine struct {
 	lastRouterSig       deephash.Sum // of router.Config
 	lastEngineSigFull   deephash.Sum // of full wireguard config
 	lastEngineSigTrim   deephash.Sum // of trimmed wireguard config
+	lastDNSConfig       *dns.Config
 	recvActivityAt      map[tailcfg.DiscoKey]time.Time
 	trimmedDisco        map[tailcfg.DiscoKey]bool // set of disco keys of peers currently excluded from wireguard config
 	sentActivityAt      map[netaddr.IP]*int64     // value is atomic int64 of unixtime
@@ -752,6 +753,7 @@ func (e *userspaceEngine) Reconfig(cfg *wgcfg.Config, routerCfg *router.Config, 
 
 	e.wgLock.Lock()
 	defer e.wgLock.Unlock()
+	e.lastDNSConfig = dnsCfg
 
 	peerSet := make(map[key.Public]struct{}, len(cfg.Peers))
 	e.mu.Lock()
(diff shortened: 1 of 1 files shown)


wgengine/userspace.go
```diff
@@ -1076,6 +1078,23 @@ func (e *userspaceEngine) linkChange(changed bool, cur *interfaces.State) {
 	health.SetAnyInterfaceUp(up)
 	e.magicConn.SetNetworkUp(up)
 
+	// Hacky workaround for Linux DNS issue 2458: on
+	// suspend/resumse or whenever NetworkManager is started, it
```
DentonGentry, line 1082: resumse

wgengine/userspace.go
```diff
@@ -1076,6 +1078,23 @@ func (e *userspaceEngine) linkChange(changed bool, cur *interfaces.State) {
...
+	// nukes all systemd-resolved configs. So reapply our DNS
+	// config on major link change.
+	if runtime.GOOS == "linux" && changed {
```
DentonGentry, line 1085: Presumably we only need to do this for the systemd-resolved case. I guess it has the downside of rewriting /etc/resolv.conf more often in the Direct case, if users have written scripts to make local modifications after Tailscale comes up. However its also likely to fix other cases of broken DNS, just by rewriting more often.

DentonGentry (approved):

danderson (approved):

wgengine/userspace.go
crawshaw: Elsewhere we have an issue suggesting we fsnotify resolv.conf and rewrite it on change, so we always win. I think that’s generally the right attitude: if we think we own it, own it.

crawshaw (approved):

frioux:
I installed locally built versions (see this:)

```bash
$ tailscale version 
date.20210603
  go version: go1.16.5
$ tailscaled -version
date.20210603
  go version: go1.16.5
```

But when I try to start it I get this:

```bash
$ sudo tailscale up
2021/07/26 08:38:57 GotNotify: Version mismatch! frontend="date.20210603" backend="1.11.137-ta5fb8e073-gfad17047b"
backend error: GotNotify: Version mismatch! frontend="date.20210603" backend="1.11.137-ta5fb8e073-gfad17047b"
```

DentonGentry:
That means tailscaled is still the release version, while the tailscale CLI command is the one you built.

If you want to permanently switch to the version you built:
1. go build cmd/tailscaled
1. sudo systemctl stop tailscaled
1. sudo cp tailscaled /usr/sbin/tailscaled
1. sudo systemctl start tailscaled

frioux:
Ah I didn't realize tailscale down was different from stopping the service.  Thanks.  This fixed it!

PR #2519, opened 2021-07-26
net/dnscache: make Dialer try all resolved IPs

bradfitz:


dsnet (approved):
concurrency looks good.

dsnet:
still looks good after `failBoost` change.

net/dnscache/dnscache.go
```diff
@@ -314,39 +314,129 @@ func Dialer(fwd DialContextFunc, dnsCache *Resolver) DialContextFunc {
...
+	}
+}
+
+// fallbackDelay is how long to wait between trying subsequent
+// addresses when multiple options are available.
+// 300ms is the same as Go's Happy Eyeballs fallbackDelay value.
+const fallbackDelay = 300 * time.Millisecond
+
+// raceDial tries to dial port on each ip in ips, starting a new race
+// dial every 300ms apart, returning whichever completes first.
```
crawshaw, line 355: could s/300ms/fallbackDelay/ in the comments so there's less chance of them getting out of date in the future

net/dnscache/dnscache.go
```diff
@@ -314,39 +314,129 @@ func Dialer(fwd DialContextFunc, dnsCache *Resolver) DialContextFunc {
...
-		ip, ip6, _, err := dnsCache.LookupIP(ctx, host)
+		ip, ip6, allIPs, err := dnsCache.LookupIP(ctx, host)
 		if err != nil {
 			return nil, fmt.Errorf("failed to resolve %q: %w", host, err)
 		}
-		dst := net.JoinHostPort(ip.String(), port)
-		if debug {
-			log.Printf("dnscache: dialing %s, %s for %s", network, dst, address)
+		i4s := v4addrs(allIPs)
+		if len(i4s) < 2 {
```
crawshaw, line 329: So if we serve exactly one ipv4 address, we try ipv4 and fall back to ipv6. If we serve multiple ipv4 addresses, we race ipv4 and ipv6 and take whichever comes back first. Does the fact we aren't favoring ipv4 in the latter case cause problems?

That is: are we happy treating ipv6 as an equal citizen on a dual stack system? Many systems in the past have favored the ipv4 connection on the theory that ipv6 is not as well maintained by dual stack ISPs and produces unexpected failures. Are those days behind us, as in, does chrome treat ipv6 equally?

crawshaw (approved):

DentonGentry (approved):

net/dnscache/dnscache.go
bradfitz: Already merged, but I can push that separately.

net/dnscache/dnscache.go
bradfitz: It still favors IPv4 in the case where there are multiple IPv4 + IPv6: the v4s are first in the list of things to try, and they get a head start.

I'm happy to promote IPv6 a little more in the future, but only once this dialing code has more context on whether IPv6 is expected to work. I wanted to make the least invasive change for 1.12. I can put more work+testing into it for the next release.

The whole reason for this conditional was being conservative and not changing the code paths for the existing situation. But arguably having an unexercised code path that might only be found to be buggy in the future after a DNS change is also kinda scary, so perhaps I should just remove this "fast path" and just always use the raceDial path?


Although arguably that's

PR #2525, opened 2021-07-27
wgengine/userspace: add support to automatically enable/disable the tailscale protocol in BIRD.

maisem:
When the node determines that it is the primary subnet router it will enable/disable the `tailscale` protocol in BIRD.
The node determines this by looking for overlapping routes in `PrimaryRoutes` and `RouteableIPs`.

`PrimaryRoutes` is what control has determined should be routed by us.
`RoutableIPs` is what is configured by `--advertise-routes`.


bradfitz (changes requested):
Let's figure out the protocol between control first.

I sent https://github.com/tailscale/tailscale/pull/2527 as a possible approach.

maisem:
> Let's figure out the protocol between control first.
> 
> I sent #2527 as a possible approach.

Rebased on #2527

chirp/chirp.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return fmt.Errorf("failed to enable %s: %v", protocol, out)
+}
+
+func (b *BIRDClient) exec(cmd string) (string, error) {
+	if _, err := b.conn.Write([]byte(cmd)); err != nil {
+		return "", err
+	}
+	buf := make([]byte, 4096)
+	if _, err := b.conn.Read(buf); err != nil {
```
bradfitz, line 62: this isn't a safe way to read from a net.Conn. there are zero guarantees about how much will be returned. you need to loop.

I assume the output is framed somehow? with a length? with a newline character?

wgengine/userspace.go
```diff
@@ -93,6 +94,7 @@ type userspaceEngine struct {
 	linkMon           *monitor.Mon
 	linkMonOwned      bool   // whether we created linkMon (and thus need to close it)
 	linkMonUnregister func() // unsubscribes from changes; used regardless of linkMonOwned
+	birdClient        *chirp.BIRDClient
```
bradfitz, line 97: `// or nil` I assume?

wgengine/userspace.go
```diff
@@ -841,6 +872,22 @@ func (e *userspaceEngine) Reconfig(cfg *wgcfg.Config, routerCfg *router.Config,
...
+		e.logf("wgengine: Reconfig: configuring BIRD")
+		var err error
+		if isSubnetRouter {
+			err = e.birdClient.EnableProtocol("tailscale")
+		} else {
+			err = e.birdClient.DisableProtocol("tailscale")
+		}
+		if err != nil {
+			// Log but don't fail here.
+			e.logf("wgengine: Reconfig: configuring BIRD failed: %v", err)
```
bradfitz, line 885: I'm not 100% sure but I think we've been moving to getting the word "error" in all failure log prints to ease grepping logs for "error" to find bad stuff.

So like "error configuring BIRD: %v"

wgengine/userspace.go
```diff
@@ -741,6 +755,17 @@ func (e *userspaceEngine) updateActivityMapsLocked(trackDisco []tailcfg.DiscoKey
 	e.tundev.SetDestIPActivityFuncs(e.destIPActivityFuncs)
 }
 
+func hasOverlap(aips, rips []netaddr.IPPrefix) bool {
```
bradfitz, line 758: comment that this is different than

https://pkg.go.dev/inet.af/netaddr#IPPrefix.Overlaps
https://pkg.go.dev/inet.af/netaddr#IPSet.Overlaps

This is not about any partial overlap.

chirp/chirp.go
maisem: Would a `bufio.Scanner` work?

wgengine/userspace.go
maisem: Done

chirp/chirp.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil, err
+	}
+	return b, nil
+}
+
+// BIRDClient handles communication with the BIRD Internet Routing Daemon.
+type BIRDClient struct {
+	socket  string
+	conn    net.Conn
+	scanner *bufio.Scanner
```
bradfitz, line 37: I'd probably just name this "bs".

cmd/tailscaled/tailscaled.go
```diff
@@ -379,6 +387,12 @@ func tryEngine(logf logger.Logf, linkMon *monitor.Mon, name string) (e wgengine.
 		ListenPort:  args.port,
 		LinkMonitor: linkMon,
 	}
+	if args.birdSocketPath != "" && createBIRDClient != nil {
+		conf.BIRDClient, err = createBIRDClient(args.birdSocketPath)
```
bradfitz, line 391: do you want to do some logging at least before this line, as createBIRDClient can block (potentially forever) if bird doesn't reply with its greeting banner.

```go
    log.Printf("Connecting to bird at %s ...", args.birdSocketPath)
    conf.BIRDClient, err = createBIRDClient(args.birdSocketPath)
```

cmd/tailscaled/tailscaled_bird.go
```diff
@@ -0,0 +1,19 @@
...
+
+package main
+
+import (
+	"tailscale.com/chirp"
+	"tailscale.com/wgengine"
+)
+
+func init() {
+	createBIRDClient = func(ctlSocket string) (wgengine.BIRDClient, error) {
```
bradfitz, line 16: createBIRDClient = chirp.New

?

bradfitz (approved):

cmd/tailscaled/tailscaled_bird.go
maisem: mismatched return types.

PR #2527, opened 2021-07-27
tailcfg: add Node.PrimaryRoutes

bradfitz:


crawshaw (approved):

maisem (approved):
LGTM, however you might want to rev the protocol number.

bradfitz:
> LGTM, however you might want to rev the protocol number.

We can once needed/used. For now it's just an ignored field so server doesn't need to know.

PR #2530, opened 2021-07-27
cmd/tailscale/cli: document that empty string disable exit nodes, routes

bradfitz:
Updates #2529


rosszurowski (approved):
I do think Avery's point about supporting `--no-exit-node` or `--exit-node=false` is a fair one, but this is definitely a step up.

PR #2531, opened 2021-07-27
derp: add session flows

JulianKnodt:
A flow tracks exchanges from some source to destination, for the purpose of better understanding
who is using DERP. It currently tracks the number of packets sent, as well as the number of
bytes sent with the packets.


bradfitz:
Closing for now, but filed https://github.com/tailscale/tailscale/issues/3560 to track getting this updated & submitted.

PR #2532, opened 2021-07-27
net/tstun: buffer outbound channel

josharian:
The handoff between tstun.Wrap's Read and poll methods
is one of the per-packet hotspots. It shows up in pprof.

Making outbound buffered increases throughput.

It is hard to measure exactly how much, because the numbers
are highly variable, but I'd estimate it at about 1%,
using the best observed max throughput across three runs.

### DIFF
--- a/net/tstun/wrap.go
+++ b/net/tstun/wrap.go
@@ -153,9 +153,10 @@ func Wrap(logf logger.Logf, tdev tun.Device) *Wrapper {
 		// a goroutine should not block when setting it, even with no listeners.
 		bufferConsumed: make(chan struct{}, 1),
 		closed:         make(chan struct{}),
-		outbound:       make(chan tunReadResult),
-		eventsUpDown:   make(chan tun.Event),
-		eventsOther:    make(chan tun.Event),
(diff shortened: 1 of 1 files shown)


net/tstun/wrap.go
```diff
@@ -153,7 +153,7 @@ func Wrap(logf logger.Logf, tdev tun.Device) *Wrapper {
 		// a goroutine should not block when setting it, even with no listeners.
 		bufferConsumed: make(chan struct{}, 1),
 		closed:         make(chan struct{}),
-		outbound:       make(chan tunReadResult),
+		outbound:       make(chan tunReadResult, 1),
```
crawshaw, line 156: This needs a comment saying that it can be unbuffered, but the buffer improves performance.

Did you try with other numbers, or is 1 all the protocol allows?

crawshaw (approved):

net/tstun/wrap.go
josharian: In theory the protocol allows more than 1, because we occasionally inject packets, but in practice, 1 is all that will ever matter.

PR #2534, opened 2021-07-27
net/dns/resolver: EDNS OPT record off-by-one

crawshaw:
I don't know how to get access to a real packet. Basing this commit
entirely off:

```
       +------------+--------------+------------------------------+
       | Field Name | Field Type   | Description                  |
       +------------+--------------+------------------------------+
       | NAME       | domain name  | MUST be 0 (root domain)      |
       | TYPE       | u_int16_t    | OPT (41)                     |
       | CLASS      | u_int16_t    | requestor's UDP payload size |
       | TTL        | u_int32_t    | extended RCODE and flags     |
       | RDLEN      | u_int16_t    | length of all RDATA          |
       | RDATA      | octet stream | {attribute,value} pairs      |
       +------------+--------------+------------------------------+
```

From https://datatracker.ietf.org/doc/html/rfc6891#section-6.1.2

### DIFF
--- a/net/dns/resolver/forwarder.go
+++ b/net/dns/resolver/forwarder.go
@@ -125,6 +125,7 @@ func clampEDNSSize(packet []byte, maxSize uint16) {
 		return
 	}
 
+	// https://datatracker.ietf.org/doc/html/rfc6891#section-6.1.2
 	opt := packet[len(packet)-optFixedBytes:]
 
 	if opt[0] != 0 {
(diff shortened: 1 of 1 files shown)


DentonGentry (approved):
Based on #eng channel: we think this is correct.

sailorfrag:
Fix looks correct, but we need a unit test that fails on the old code; this shouldn't have been able to get past the tests.

PR #2538, opened 2021-07-28
cmd/tsshd: switch from github.com/kr/pty to github.com/creack/pty

qbit:
The kr/pty module moved to creack/pty per the kr/pty README[1].

creack/pty brings in support for a number of OS/arch combos that
are lacking in kr/pty.

Run `go mod tidy` while here.

[1] https://github.com/kr/pty/blob/master/README.md

### DIFF
--- a/cmd/tsshd/tsshd.go
+++ b/cmd/tsshd/tsshd.go
@@ -29,8 +29,8 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/creack/pty"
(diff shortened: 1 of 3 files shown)


bradfitz:
Thanks. We require the DCO sign-off line in commit messages, though.

If you agree with https://developercertificate.org/ then `git commit --amend -s` to update your commit message, then `git push -f` the branch.

qbit:
Updated!

PR #2539, opened 2021-07-28
ipn/ipnstate: add json struct tags to PingResult

simenghe:

Adding struct tags to `PingResult`. `PingResult`s will be sent to adminhttp etc via http.

### DIFF
--- a/ipn/ipnstate/ipnstate.go
+++ b/ipn/ipnstate/ipnstate.go
@@ -418,25 +418,25 @@ func osEmoji(os string) string {
 // PingResult contains response information for the "tailscale ping" subcommand,
 // saying how Tailscale can reach a Tailscale IP or subnet-routed IP.
 type PingResult struct {
-	IP       string // ping destination
-	NodeIP   string // Tailscale IP of node handling IP (different for subnet routers)
-	NodeName string // DNS name base or (possibly not unique) hostname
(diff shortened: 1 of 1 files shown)


ipn/ipnstate/ipnstate.go
```diff
@@ -418,25 +418,25 @@ func osEmoji(os string) string {
...
-	DERPRegionID int
+	DERPRegionID int `json:"derpRegionID"`
 
 	// DERPRegionCode is the three-letter region code
 	// corresponding to DERPRegionID.
 	// It is not currently set for TSMP pings.
-	DERPRegionCode string
+	DERPRegionCode string `json:"derpRegionCode"`
 
 	// PeerAPIPort is set by TSMP ping responses for peers that
```
sailorfrag, line 441: Should PeerAPIPort also have an explicit json name? Right now the tag is just specifying omitempty

sailorfrag (commented):
This seems fine although I am curious if there's a strong motivating reason to put in tags? This is more about filling in my incomplete understanding of Go JSON usage, rather than implied disapproval.

simenghe:
> I'm also curious about the motivation. This is an internal-detail and not a public facing API, right? So conventions don't really matter.



> I'm also curious about the motivation. This is an internal-detail and not a public facing API, right? So conventions don't really matter.

understood!

PR #2542, opened 2021-07-28
net/tsaddr: remove IsULA, replace with netaddr.IP.IsPrivate

mdlayher:
Signed-off-by: Matt Layher <mdlayher@gmail.com>

IsPrivate covers both the RFC1918 IPv4 ranges and the IPv6 ULA ranges, so the IsULA helper is no longer necessary.


mdlayher:
Added a second commit to fix a minor regression since IsPrivate will accept both address families.

DentonGentry (approved):

Issue #2414, opened 2021-07-13
cmd/derper: mesh derpers using private IPv4 addresses


bradfitz:
https://docs.digitalocean.com/products/billing/bandwidth/ says:

> Any data transfer sent by a Droplet using a public network interface will count against the transfer pool for that Droplet’s account. All IPv6 traffic uses the public interface. Data transfer between Droplets over the VPC network uses a private network interface.

But currently we mesh inter-region traffic using DNS names which map to the public v4/v6 addresses.

We should teach our transport dialer's to use the VPC private IPv4s instead of those names. Or just /etc/hosts it.

bradfitz:
and whoops, some of our derpers don't even have a VPC

bradfitz:
Yeah, we've done this enough at this point. Not everywhere, but we're good at it at this point and know how to do it as needed now, and have the right monitoring, etc.

PR #2549, opened 2021-07-29
cmd/derper: mesh over VPC network

bradfitz:
Updates #2414

### DIFF
--- a/cmd/derper/mesh.go
+++ b/cmd/derper/mesh.go
@@ -9,7 +9,9 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net"
 	"strings"
+	"time"
 
 	"tailscale.com/derp"
 	"tailscale.com/derp/derphttp"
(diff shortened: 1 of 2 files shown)


bradfitz:
```
Jul 29 17:45:49 derp4-fra-c derper[18434]: 2021/07/29 17:45:49 connected to derp4c-vpc.tailscale.com (10.114.0.4) instead of derp4c
Jul 29 17:45:49 derp4-fra-c derper[18434]: 2021/07/29 17:45:49 connected to derp4e-vpc.tailscale.com (10.114.0.3) instead of derp4e
Jul 29 17:45:49 derp4-fra-c derper[18434]: 2021/07/29 17:45:49 connected to derp4d-vpc.tailscale.com (10.114.0.5) instead of derp4d
```

cmd/derper/mesh.go
```diff
@@ -39,6 +41,34 @@ func startMeshWithHost(s *derp.Server, host string) error {
...
+		var d net.Dialer
+		var r net.Resolver
+		if port == "443" && strings.HasSuffix(host, ".tailscale.com") {
+			base := strings.TrimSuffix(host, ".tailscale.com")
+			subCtx, cancel := context.WithTimeout(ctx, 2*time.Second)
+			defer cancel()
+			vpcHost := base + "-vpc.tailscale.com"
+			ips, _ := r.LookupIP(subCtx, "ip", vpcHost)
+			if len(ips) > 0 {
+				vpcAddr := net.JoinHostPort(ips[0].String(), port)
```
crawshaw, line 60: Is there much value in using the result of LookupIP? Doesn't DialContext with the hostname do something clever here about trying mulitple IPs? If so, you could `net.JoinHostPort(vpcHost, port)` and let it do the lookup again.

crawshaw (approved):

cmd/derper/mesh.go
bradfitz: Not a huge difference. I mostly wanted different log messages for the difference cases.

PR #2550, opened 2021-07-29
net/dns/resolver: allow an extra alloc for go closure allocation

### LINKED ISSUE #2545: net/dns/resolver: allocs test failing on Go 1.17
```
$ go test -v -run=TestAllocs ./net/dns/resolver
=== RUN   TestAllocs
    tsdns_test.go:945: forward: allocs = 3; want 2
--- FAIL: TestAllocs (0.00s)
FAIL
FAIL    tailscale.com/net/dns/resolver  0.003s
FAIL
```

cc @josharian

josharian:
Go 1.17 switches to a register ABI on amd64 platforms.
Part of that switch is that go and defer calls use an argument-less
closure, which allocates. This means that we have an extra
alloc in some DNS work. That's unfortunate but not a showstopper,
and I don't see a clear path to fixing it.
The other performance benefits from the register ABI will all
but certainly outweigh this extra alloc.

Fixes #2545

### DIFF
--- a/net/dns/resolver/tsdns_test.go
+++ b/net/dns/resolver/tsdns_test.go
@@ -931,8 +931,11 @@ func TestAllocs(t *testing.T) {
 		query []byte
 		want  int
 	}{
-		// Name lowercasing and response slice created by dns.NewBuilder.
-		{"forward", dnspacket("test1.ipn.dev.", dns.TypeA, noEdns), 2},
+		// Name lowercasing, response slice created by dns.NewBuilder,
+		// and closure allocation from go call.
+		// (Closure allocation only happens when using new register ABI,
+		// which is amd64 with Go 1.17, and probably more platforms later.)
+		{"forward", dnspacket("test1.ipn.dev.", dns.TypeA, noEdns), 3},
 		// 3 extra allocs in rdnsNameToIPv4 and one in marshalPTRRecord (dns.NewName).
 		{"reverse", dnspacket("4.3.2.1.in-addr.arpa.", dns.TypePTR, noEdns), 5},
 	}


crawshaw (approved):
Oh interesting. Wasn't it considered a Big Deal when they switched to not allocating for every defer? Or can they still avoid the allocation for easy method defers, like `defer mu.Unlock()`?

josharian:
> Or can they still avoid the allocation for easy method defers, like `defer mu.Unlock()`?

The key question is whether the closure itself can be stack allocated.

Empirically, things are OK. A benchmark that does `mu.Lock(); defer mu.Unlock()` runs in 18ns/op, 0 allocs/op. A benchmark that does `mu.Lock(); go mu.Unlock()` runs in 680ns/op, 1 alloc/op.

Many defer closures can go on the stack, including `mu.Unlock`. But almost all `go` closures are going to escape. (I guess if you're only dealing with references to global variables you could put it on the newly created goroutine's stack, but that's going to require a delicate dance with the runtime.) Anyway, for now, the compiler makes them escape unilaterally. From `wrapGoDefer` (in the walk pass):

```go
	// This is set to true if the closure we're generating escapes
	// (needs heap allocation).
	cloEscapes := func() bool {
		if n.Op() == ir.OGO {
			// For "go", assume that all closures escape.
			return true
		}
		// For defer, just use whatever result escape analysis
		// has determined for the defer.
		return n.Esc() != ir.EscNever
	}()
```

Maybe this'll get improved later. The comment for `wrapGoDefer` does say:

```go
// wrapGoDefer wraps the target of a "go" or "defer" statement with a
// new "function with no arguments" closure. Specifically, it converts
//
//   defer f(x, y)
//
// to
//
//   x1, y1 := x, y
//   defer func() { f(x1, y1) }()
//
// This is primarily to enable a quicker bringup of defers under the
// new register ABI; by doing this conversion, we can simplify the
// code in the runtime that invokes defers on the panic path.
```

PR #2567, opened 2021-08-02
derp: reduce server memory by 30% by removing persistent bufio.Writer

bradfitz:


crawshaw (approved):
This seems fine, but why do we have a buffer here? Is it to save syscalls? Is it because we don't handle backpressure well from the TCP send buffer in the kernel?

What happens... if we don't have one?

bradfitz:
> This seems fine, but why do we have a buffer here? Is it to save syscalls? Is it because we don't handle backpressure well from the TCP send buffer in the kernel?

Because we construct DERP frames over a series of writes: first header, then body (or multiple body chunks).

Also, when copying packets, if multiple are waiting, we send them all together in one write until there's nothing left to send and then only do 1 final flush when idle.

> What happens... if we don't have one?

We end up writing 3 or 4 packets instead of 1, including that many extra TLS records.

PR #2573, opened 2021-08-03
logpolicy: only log panics when running under systemd

crawshaw:
Given that https://github.com/golang/go/issues/42888 is coming, this
catches most practical panics without interfering in our development
environments.

### DIFF
--- a/logpolicy/logpolicy.go
+++ b/logpolicy/logpolicy.go
@@ -187,6 +187,10 @@ func runningUnderSystemd() bool {
 	return false
 }
 
+func redirectStderrToLogPanics() bool {
+	return runningUnderSystemd() || os.Getenv("TS_PLEASE_PANIC") != ""
+}
+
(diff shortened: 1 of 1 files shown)


logpolicy/logpolicy.go
```diff
@@ -435,12 +439,16 @@ func New(collection string) *Policy {
 		c.HTTPC = &http.Client{Transport: newLogtailTransport(u.Host)}
 	}
 
-	filchBuf, filchErr := filch.New(filepath.Join(dir, cmdName), filch.Options{
-		ReplaceStderr: true,
-	})
+	opts := filch.Options{}
+	if redirectStderrToLogPanics() {
```
bradfitz, line 443: or just
```
.... filch.Options{
    ReplaceStderr: redirectStderrToLogPanics(),
}
```

(kinda like it was)

Either way.

bradfitz (approved):

PR #2572, opened 2021-08-03
net/portmapper: fix UPnP probing

bradfitz:
Prior to Tailscale 1.12 it detected UPnP on any port.
Starting with Tailscale 1.11.x, it stopped detecting UPnP on all ports.

Then use the Location header to find the rootDesc.xml, rather than hard-coding the URL + port.

And remove the three goroutines doing redundant & racing rootDesc.xml fetches.

And use netns so it works while using exit nodes.

Works against pfSense now.

Updates #2109

PR #2575, opened 2021-08-03
net/portmapper: reuse root device across calls

JulianKnodt:
Before, we were sending 3 http requests to the same device, which was both costly and likely
caused it to be slower than necessary.


bradfitz:
did you see I'd already done this as part of https://github.com/tailscale/tailscale/pull/2572/files#diff-aa077f884f6428db68a1b9ed0acb657a4e83c023ae0fb39c36fdbb34e37630feR174 ?

JulianKnodt:
ah I missed that, should I just close this then?

Issue #2563, opened 2021-08-01
Implement Port Control Protocol


DentonGentry:
Seen in logs for an unrelated issue:

```
portmapper: unexpected PMP response: 02 80
portmapper: unexpected PMP response: 02 81 00 02 a2 a9 00 00 00 4f 8f d2
```

This happens when parsePMPResponse returns !ok.

```
func parsePMPResponse(pkt []byte) (res pmpResponse, ok bool) {
        if len(pkt) < 12 {
                return
        }
        ver := pkt[0]
        if ver != 0 {
                return
        }
```

The RFC for NAT-PMP https://datatracker.ietf.org/doc/html/rfc6886#section-1.1 says, _"The version number in the packet header is 0 for NAT-PMP and 2 for PCP, so the packets are easily distinguished."_

So this is the Port Control Protocol being seen in the wild.

bradfitz:
Which version logged this?

We know PCP exists in the wild, but it was so rarely the only option that we haven't implemented it (despite being easy). But now that we do more than 1, adding a 3rd should be easy. (@JulianKnodt, interested?)

I'm surprised at this log message, though. Is a PMP query resulting in a PCP response?

DentonGentry:
This came from 1.12.1 on an iOS device. The user reported a different problem, which I'm looking into, and happened to notice this.

It does appear that we sent a NAT-PMP request and we got back a PCP response.

JulianKnodt:
I can try it out next week! altho that one I'd need to dig thru the spec which hopefully will be short

JulianKnodt:
This issue should be closed after 1.14 when this is included

DentonGentry:
Fixed in 1.14.

PR #2577, opened 2021-08-03
net/portmapper: add pcp portmapping

JulianKnodt:
This adds PCP portmapping, moving all existing PCP content into a separate file, and parsing the
response from PCP portmapping requests. It hooks into the existing way we handle PMP mapping.

Tested this out on Denton's router, and seems to work fine, but would appreciate others also checking it out.
Also seems to work on pfSense on Brad's DO instance.

Updates #2563

### DIFF
--- a/net/portmapper/pcp.go
+++ b/net/portmapper/pcp.go
@@ -5,10 +5,14 @@
 package portmapper
 
 import (
+	"context"
 	"crypto/rand"
 	"encoding/binary"
+	"fmt"
+	"time"
 
 	"inet.af/netaddr"
+	"tailscale.com/net/netns"
 )
 
 // References:
@@ -34,41 +38,92 @@ const (
 	pcpTCPMapping = 6  // portmap TCP
 )
 
-// pcpMapRequest generates a PCP packet with a MAP opcode.
-func pcpMapRequest(myIP netaddr.IP, mapToLocalPort int, delete bool) []byte {
-	const udpProtoNumber = 17
-	lifetimeSeconds := uint32(1)
-	if delete {
-		lifetimeSeconds = 0
+type pcpMapping struct {
+	gw       netaddr.IP
+	internal netaddr.IPPort
+	external netaddr.IPPort
+
+	renewAfter time.Time
+	goodUntil  time.Time
+}
+
+func (p *pcpMapping) GoodUntil() time.Time     { return p.goodUntil }
+func (p *pcpMapping) RenewAfter() time.Time    { return p.renewAfter }
+func (p *pcpMapping) External() netaddr.IPPort { return p.external }
+func (p *pcpMapping) Release(ctx context.Context) {
+	uc, err := netns.Listener().ListenPacket(ctx, "udp4", ":0")
+	if err != nil {
+		return
 	}
-	const opMap = 1
+	defer uc.Close()
+	pkt := buildPCPRequestMappingPacket(p.internal.IP(), p.internal.Port(), p.external.Port(), 0)
+	uc.WriteTo(pkt, netaddr.IPPortFrom(p.gw, pcpPort).UDPAddr())
+}
 
-	// 24 byte header + 36 byte map opcode
-	pkt := make([]byte, (32+32+128)/8+(96+8+24+16+16+128)/8)
+// buildPCPRequestMappingPacket generates a PCP packet with a MAP opcode.
+// To create a packet which deletes a mapping, lifetimeSec should be set to 0.
+// If prevPort is not known, it should be set to 0.
+func buildPCPRequestMappingPacket(myIP netaddr.IP, localPort, prevPort uint16, lifetimeSec uint32) (pkt []byte) {
+	// note: lifetimeSec = 0 implies delete the mapping, should that be special-cased here?
 
-	// The header (https://tools.ietf.org/html/rfc6887#section-7.1)
-	pkt[0] = 2 // version
-	pkt[1] = opMap
-	binary.BigEndian.PutUint32(pkt[4:8], lifetimeSeconds)
+	// 24 byte common PCP header + 36 bytes of MAP-specific fields
+	pkt = make([]byte, 24+36)
+	pkt[0] = pcpVersion
+	pkt[1] = pcpOpMap
+	binary.BigEndian.PutUint32(pkt[4:8], lifetimeSec)
 	myIP16 := myIP.As16()
-	copy(pkt[8:], myIP16[:])
+	copy(pkt[8:24], myIP16[:])
 
-	// The map opcode body (https://tools.ietf.org/html/rfc6887#section-11.1)
 	mapOp := pkt[24:]
-	rand.Read(mapOp[:12]) // 96 bit mappping nonce
-	mapOp[12] = udpProtoNumber
-	binary.BigEndian.PutUint16(mapOp[16:], uint16(mapToLocalPort))
+	rand.Read(mapOp[:12]) // 96 bit mapping nonce
+
+	// TODO should this be a UDP mapping? It looks like it supports "all protocols" with 0, but
+	// also doesn't support a local port then.
+	mapOp[12] = pcpUDPMapping
+	binary.BigEndian.PutUint16(mapOp[16:18], localPort)
+	binary.BigEndian.PutUint16(mapOp[18:20], prevPort)
+
 	v4unspec := netaddr.MustParseIP("0.0.0.0")
 	v4unspec16 := v4unspec.As16()
 	copy(mapOp[20:], v4unspec16[:])
 	return pkt
 }
 
+func parsePCPMapResponse(resp []byte) (*pcpMapping, error) {
+	if len(resp) < 60 {
+		return nil, fmt.Errorf("Does not appear to be PCP MAP response")
+	}
+	res, ok := parsePCPResponse(resp[:24])
+	if !ok {
+		return nil, fmt.Errorf("Invalid PCP common header")
+	}
+	if res.ResultCode != pcpCodeOK {
+		return nil, fmt.Errorf("PCP response not ok, code %d", res.ResultCode)
+	}
+	// TODO don't ignore the nonce and make sure it's the same?
+	externalPort := binary.BigEndian.Uint16(resp[42:44])
+	externalIPBytes := [16]byte{}
+	copy(externalIPBytes[:], resp[44:])
+	externalIP := netaddr.IPFrom16(externalIPBytes)
+
+	external := netaddr.IPPortFrom(externalIP, externalPort)
+
+	lifetime := time.Second * time.Duration(res.Lifetime)
+	now := time.Now()
+	mapping := &pcpMapping{
+		external:   external,
+		renewAfter: now.Add(lifetime / 2),
+		goodUntil:  now.Add(lifetime),
+	}
+
+	return mapping, nil
+}
+
 // pcpAnnounceRequest generates a PCP packet with an ANNOUNCE opcode.
 func pcpAnnounceRequest(myIP netaddr.IP) []byte {
 	// See https://tools.ietf.org/html/rfc6887#section-7.1
 	pkt := make([]byte, 24)
-	pkt[0] = pcpVersion // version
+	pkt[0] = pcpVersion
 	pkt[1] = pcpOpAnnounce
 	myIP16 := myIP.As16()
 	copy(pkt[8:], myIP16[:])
--- a/net/portmapper/pcp.go
+++ b/net/portmapper/pcp.go
@@ -59,14 +59,20 @@ func (p *pcpMapping) Release(ctx context.Context) {
 		return
 	}
 	defer uc.Close()
-	pkt := buildPCPRequestMappingPacket(p.internal.IP(), p.internal.Port(), p.external.Port(), 0)
+	pkt := buildPCPRequestMappingPacket(p.internal.IP(), p.internal.Port(), p.external.Port(), 0, p.external.IP())
 	uc.WriteTo(pkt, netaddr.IPPortFrom(p.gw, pcpPort).UDPAddr())
 }
 
 // buildPCPRequestMappingPacket generates a PCP packet with a MAP opcode.
 // To create a packet which deletes a mapping, lifetimeSec should be set to 0.
 // If prevPort is not known, it should be set to 0.
-func buildPCPRequestMappingPacket(myIP netaddr.IP, localPort, prevPort uint16, lifetimeSec uint32) (pkt []byte) {
+// If prevExternalIP is not known, it should be set to 0.0.0.0.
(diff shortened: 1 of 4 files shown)


net/portmapper/portmapper.go
```diff
@@ -652,75 +669,4 @@ func (c *Client) Probe(ctx context.Context) (res ProbeResult, err error) {
...
-}
-
-type pcpResponse struct {
-	OpCode     uint8
-	ResultCode uint8
-	Lifetime   uint32
-	Epoch      uint32
-}
-
-func parsePCPResponse(b []byte) (res pcpResponse, ok bool) {
```
bradfitz, line 715: can you do the code movement in its own commit first?

net/portmapper/portmapper.go
```diff
@@ -3,12 +3,11 @@
 // license that can be found in the LICENSE file.
 
 // Package portmapper is a UDP port mapping client. It currently allows for mapping over
-// NAT-PMP and UPnP, but will perhaps do PCP later.
+// NAT-PMP, UPnP, and PCP.
```
bradfitz, line 6: technically this belongs in the second commit :)

net/portmapper/portmapper.go
```diff
@@ -652,75 +650,4 @@ func (c *Client) Probe(ctx context.Context) (res ProbeResult, err error) {
...
-	pkt := make([]byte, 24)
-	pkt[0] = pcpVersion // version
-	pkt[1] = pcpOpAnnounce
-	myIP16 := myIP.As16()
-	copy(pkt[8:], myIP16[:])
-	return pkt
-}
-
-// pcpMapRequest generates a PCP packet with a MAP opcode.
-func pcpMapRequest(myIP netaddr.IP, mapToLocalPort int, delete bool) []byte {
```
bradfitz, line 679: why's this being deleted? won't we need to make map requests to use PCP?

JulianKnodt: it wasn't being used previously, and I initially didn't notice it existed when I wrote `buildPCPMapRequestPacket`. I took the parts of this function that were good and moved it into that function

net/portmapper/portmapper.go
```diff
@@ -413,6 +415,11 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
 		return netaddr.IPPort{}, err
 	}
 
+	pcpReqMapping := buildPCPRequestMappingPacket(myIP, localPort, prevPort, pcpMapLifetimeSec)
```
bradfitz, line 418: I don't think we should race a PMP & PCP mapping. That's going to be a little chaotic.

Only send one or the other, preferring PMP for now.

net/portmapper/pcp.go
```diff
@@ -34,41 +38,89 @@ const (
 	pcpTCPMapping = 6  // portmap TCP
 )
 
-// pcpMapRequest generates a PCP packet with a MAP opcode.
```
bradfitz: restore comment. new version lacks a comment.

you can then document that lifetimeSec 0 means delete.

net/portmapper/portmapper.go
```diff
@@ -432,24 +439,38 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
...
-			}
-			if pres.OpCode == pmpOpReply|pmpOpMapUDP {
-				m.external = m.external.WithPort(pres.ExternalPort)
-				d := time.Duration(pres.MappingValidSeconds) * time.Second
-				now := time.Now()
-				m.goodUntil = now.Add(d)
-				m.renewAfter = now.Add(d / 2) // renew in half the time
-				m.epoch = pres.SecondsSinceEpoch
+			version := res[0]
+			switch version {
```
bradfitz, line 443: add a default case with a log on unexpected version?

net/portmapper/portmapper.go
```diff
@@ -432,24 +439,38 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
...
+					d := time.Duration(pres.MappingValidSeconds) * time.Second
+					now := time.Now()
+					m.goodUntil = now.Add(d)
+					m.renewAfter = now.Add(d / 2) // renew in half the time
+					m.epoch = pres.SecondsSinceEpoch
+				}
+			case pcpVersion:
+				pcpMapping, err := parsePCPMapResponse(res[:n])
+				if err != nil {
+					continue
```
bradfitz, line 467: log it at least? this goes nowhere now?

net/portmapper/portmapper.go
```diff
@@ -409,10 +414,18 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
 		}
 	}
 
-	// And ask for a mapping.
-	pmpReqMapping := buildPMPRequestMappingPacket(localPort, prevPort, pmpMapLifetimeSec)
-	if _, err := uc.WriteTo(pmpReqMapping, pmpAddru); err != nil {
-		return netaddr.IPPort{}, err
+	if haveRecentPMP {
+		// And ask for a mapping.
+		pmpReqMapping := buildPMPRequestMappingPacket(localPort, prevPort, pmpMapLifetimeSec)
```
bradfitz, line 419: the scope of this variable is two lines, so this variable can be shorter. (and be the same as the ones on line 425/426 to match)

Maybe "pkt" to match your other caller of these above?

net/portmapper/portmapper.go
```diff
@@ -409,10 +414,18 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
...
+		if _, err := uc.WriteTo(pmpReqMapping, pmpAddru); err != nil {
+			return netaddr.IPPort{}, err
+		}
+	} else {
+		// Only do PCP mapping in the case when PMP did not appear to be available recently.
```
bradfitz, line 424: so we send PCP unconditionally all the time even if we've never heard about its availability from an earlier probe?

I guess that's what we did for PMP previously?

Seems a little less than ideal. Maybe we should just bail out earlier if we haven't heard from anything.

net/portmapper/portmapper.go
```diff
@@ -434,24 +447,41 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
...
+					d := time.Duration(pres.MappingValidSeconds) * time.Second
+					now := time.Now()
+					m.goodUntil = now.Add(d)
+					m.renewAfter = now.Add(d / 2) // renew in half the time
+					m.epoch = pres.SecondsSinceEpoch
+				}
+			case pcpVersion:
+				pcpMapping, err := parsePCPMapResponse(res[:n])
+				if err != nil {
+					c.logf("Failed to get PCP mapping: %v", err)
```
bradfitz, line 475: generally style is to start log/error messages with lowercase, as they compose more nicely when there are prefixes before them.  I think vet or staticcheck or something even complains about that.

it also makes grepping logs for "fail" and "error" easier, as you don't have to do a case insensitive search

net/portmapper/portmapper.go
```diff
@@ -434,24 +447,41 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
...
+					continue
+				}
+				pcpMapping.internal = m.internal
+				c.mu.Lock()
+				defer c.mu.Unlock()
+				c.mapping = pcpMapping
+				return pcpMapping.external, nil
+			default:
+				c.logf("Unknown PMP/PCP version number: %d", version)
```
bradfitz, line 484: unknown (lowercase).  dump the packet?

net/portmapper/pcp_test.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"testing"
+)
+
+var examplePCPMapResponse = []byte{2, 129, 0, 0, 0, 0, 28, 32, 0, 2, 155, 237, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 129, 112, 9, 24, 241, 208, 251, 45, 157, 76, 10, 188, 17, 0, 0, 0, 4, 210, 4, 210, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 135, 180, 175, 246}
+
+func TestParsePCPMapResponse(t *testing.T) {
+	mapping, err := parsePCPMapResponse(examplePCPMapResponse)
+	if err != nil {
+		t.Errorf("Failed to parse PCP Map Response: %v", err)
```
bradfitz, line 16: Fatalf if the test can't proceed usefully

net/portmapper/pcp_test.go
```diff
@@ -0,0 +1,21 @@
...
+	}
+	if err == nil && mapping == nil {
```
bradfitz, line 18: then remove the err == nil part.

net/portmapper/pcp_test.go
```diff
@@ -0,0 +1,21 @@
...
+		t.Errorf("Got nil mapping when expected non-nil")
```
bradfitz, line 19: Fatalf, then see if mapping is what you expect, other than its nilness

Or reflect.DeepEqual mapping against an expected value.

net/portmapper/portmapper.go
JulianKnodt: oops that's not what I intended, let me rewrite it

net/portmapper/portmapper.go
```diff
@@ -409,10 +414,18 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
...
-	if _, err := uc.WriteTo(pmpReqMapping, pmpAddru); err != nil {
-		return netaddr.IPPort{}, err
+	if !haveRecentPMP && haveRecentPCP {
+		// Only do PCP mapping in the case when PMP did not appear to be available recently.
+		pkt := buildPCPRequestMappingPacket(myIP, localPort, prevPort, pcpMapLifetimeSec)
+		if _, err := uc.WriteTo(pkt, pcpAddru); err != nil {
+			return netaddr.IPPort{}, err
+		}
+	} else {
+		// And ask for a mapping.
```
bradfitz, line 424: This comment style, staring with "And ask" no longer reads well in context of just the else.

net/portmapper/portmapper.go
```diff
@@ -409,10 +414,18 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
 		}
 	}
 
-	// And ask for a mapping.
-	pmpReqMapping := buildPMPRequestMappingPacket(localPort, prevPort, pmpMapLifetimeSec)
-	if _, err := uc.WriteTo(pmpReqMapping, pmpAddru); err != nil {
-		return netaddr.IPPort{}, err
+	if !haveRecentPMP && haveRecentPCP {
```
bradfitz, line 417: (line 412 up there where stupid github won't let me comment)  .... you're still writing to PMP to ask for the external address in the PCP case.

how do we get the external IP with PCP?

JulianKnodt: It's contained in the PCP packet, so we don't need to do anything special for it

net/portmapper/portmapper.go
```diff
@@ -401,18 +420,31 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
 
 	pmpAddr := netaddr.IPPortFrom(gw, pmpPort)
```
bradfitz, line 421: just make this variable be called pxpAddr and use it for both. then ditch pcpAddr{,u}

net/portmapper/portmapper.go
```diff
@@ -401,18 +420,31 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
...
 	pmpAddru := pmpAddr.UDPAddr()
+	pcpAddr := netaddr.IPPortFrom(gw, pcpPort)
+	pcpAddru := pcpAddr.UDPAddr()
 
-	// Ask for our external address if needed.
-	if m.external.IP().IsZero() {
-		if _, err := uc.WriteTo(pmpReqExternalAddrPacket, pmpAddru); err != nil {
+	preferPCP := DisablePMP || (!haveRecentPMP && haveRecentPCP)
```
bradfitz, line 426: shouldn't need to check DisablePMP here? If we didn't send the PMP solicitation out before, haveRecentPMP will be false, right?

But if we do need it, why don't we also need to check DisablePCP?

net/portmapper/portmapper.go
```diff
@@ -371,23 +378,35 @@ func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPor
...
+		return netaddr.IPPort{}, NoMappingError{ErrNoPortMappingServices}
+	}
+
 	// If we just did a Probe (e.g. via netchecker) but didn't
 	// find a PMP service, bail out early rather than probing
 	// again. Cuts down latency for most clients.
 	haveRecentPMP := c.sawPMPRecentlyLocked()
+	haveRecentPCP := c.sawPCPRecentlyLocked()
 
+	m := &pmpMapping{
```
bradfitz, line 395: is this used for PCP as well? if not, why is only one created up here, and why PMP?

net/portmapper/portmapper.go
```diff
@@ -337,12 +342,18 @@ func (c *Client) createMapping() {
...
+var wildcardIP = netaddr.MustParseIP("0.0.0.0")
+
 // createOrGetMapping either creates a new mapping or returns a cached
 // valid one.
 //
 // If no mapping is available, the error will be of type
 // NoMappingError; see IsNoMappingError.
 func (c *Client) createOrGetMapping(ctx context.Context) (external netaddr.IPPort, err error) {
+	if DisableUPnP && DisablePCP && DisablePMP {
+		return
```
bradfitz, line 355: don't use naked returns on functions over 2 or 3 lines

also, why isn't this `return netaddr.IPPort{}, NoMappingError{ErrGatewayRange}` like line 359? Docs do say:

```
// If no mapping is available, the error will be of type
// NoMappingError; see IsNoMappingError.
```

JulianKnodt: Since all of them are disabled it feels odd to me that it's a `NoMappingError`, but certainly it's more right than returning nil

net/portmapper/portmapper.go
JulianKnodt: this is from the old code, so I mainly left this in to preserve that, but PMP also requires two steps, getting the mapping and getting the external IP. I just shifted it around so that if it doesn't gets called it won't be allocated. I left a comment above it documenting the behaviour

net/portmapper/portmapper.go
JulianKnodt: I'm not sure what you mean by PMP solicitation, other than the the request for the external IP, but I don't think that will update haveRecentPMP

I updated it to make sure that it won't try PCP if PCP is disabled

PR #2582, opened 2021-08-04
ipn/ipnstate: move tailscale status "active" determination to tailscaled

### LINKED ISSUE #2579: cmd/tailscale: status shows all nodes active (at main)
Monotonic time regression: all lines in `tailscale status` show active due to:

```go
// peerActive reports whether ps has recent activity.
//
// TODO: have the server report this bool instead.
func peerActive(ps *ipnstate.PeerStatus) bool {
        return !ps.LastWrite.IsZero() && mono.Since(ps.LastWrite) < 2*time.Minute
}
```

And `ps.LastWrite.IsZero()` is now always false, even for nodes which have never written.

because the json parse of the zero time ends up doing:

```go
// UnmarshalJSON sets t according to data.
// This is best-effort only. Time does not survive a MarshalJSON/UnmarshalJSON round trip unchanged.
func (t *Time) UnmarshalJSON(data []byte) error {
        var tt time.Time
        err := tt.UnmarshalJSON(data)
        if err != nil {
                return err
        }
        *t = Now().Add(-time.Since(tt))
        return nil
}

// Now returns the current monotonic time.
func Now() Time {
        // On a newly started machine, the monotonic clock might be very near zero.
        // Thus mono.Time(0).Before(mono.Now.Add(-time.Minute)) might yield true.
        // The corresponding package time expression never does, if the wall clock is correct.
        // Preserve this correspondence by increasing the "base" monotonic clock by a fair amount.
        const baseOffset int64 = 1 << 55 // approximately 10,000 hours in nanoseconds
        return Time(now() + baseOffset)
}
```

Time to do that TODO:

```
// TODO: have the server report this bool instead.
```

That is, add a field to `ipnstate.PeerStatus`: `Active bool`.

(Reported by @danderson)

bradfitz:
Fixes #2579


crawshaw (approved):
Looks right, but let Josh have a pass?

bradfitz:
Going to merge for now. Happy to make edits later.

PR #2586, opened 2021-08-04
net/dns: replace resolver IPs with type for DoH

crawshaw:


bradfitz:
This commit message gives me zero background/motivation for this change.

What's happening here?

net/dns/resolver/tsdns.go
```diff
@@ -121,9 +140,33 @@ func WriteIPPorts(w *bufio.Writer, vv []netaddr.IPPort) {
 	w.WriteByte(']')
 }
 
+func WriteResolverName(w *bufio.Writer, r ResolverName) {
```
bradfitz, line 143: docs

net/dns/resolver/tsdns.go
```diff
@@ -121,9 +140,33 @@ func WriteIPPorts(w *bufio.Writer, vv []netaddr.IPPort) {
...
+		var b []byte
+		for _, ip := range r.BootstrapResolution {
+			ip.AppendTo(b[:0])
+			w.Write(b)
+		}
+		w.WriteByte(')')
+	}
+}
+
+func WriteResolverNames(w *bufio.Writer, resolvers []ResolverName) {
```
bradfitz, line 156: docs

net/dns/resolver/tsdns.go
```diff
@@ -62,6 +63,24 @@ type packet struct {
...
+	//  - "http(s?)://resolver.com/query-tmpl" for DNS over HTTP(S?)
+	Addr string
+
+	// BootstrapResolution is an optional suggested resolution for the
+	// DoT/DoH resolver, if the resolver URL does not reference an IP
+	// address directly.
+	// BootstrapResolution may be empty, in which case clients should
+	// look up the DoT/DoH server using their local "classic" DNS
+	// resolver.
+	BootstrapResolution []netaddr.IP
```
bradfitz, line 81: could these be URL parameters on Addr if Addr is a URL-ish thing? Maybe not for efficiency. Is this used yet?

net/dns/resolver/tsdns.go
```diff
@@ -62,6 +63,24 @@ type packet struct {
 	addr netaddr.IPPort // src for a request, dst for a response
 }
 
+// ResolverName is the configuration for a DNS resolver.
+type ResolverName struct {
+	// Addr is the address of the DNS resolver, one of:
+	//  - A plain IP address for a "classic" UDP+TCP DNS resolver
+	//  - An IP:port pair
```
bradfitz, line 70: what does this mean? same as first bullet but alternate port? same port for both UDP+TCP?

net/dns/resolver/tsdns.go
```diff
@@ -62,6 +63,24 @@ type packet struct {
 	addr netaddr.IPPort // src for a request, dst for a response
 }
 
+// ResolverName is the configuration for a DNS resolver.
```
bradfitz, line 66: sounds like the name of a resolver.

what's wrong with tailcfg.DNSResolver ? seems like this is a copy of that? (why a copy?)

net/dns/resolver/tsdns.go
crawshaw: I don't want to deviate from the tailcfg type by inventing a new type here. Otherwise, sounds possible.

crawshaw:
Commentary added, and a test to net/dns showing how it will be used by an exit node for forwarding.

net/dns/resolver/forwarder.go
```diff
@@ -206,19 +208,19 @@ func resolversWithDelays(ipps []netaddr.IPPort) []resolverAndDelay {
 	// per address family.
 	total := map[hostAndFam]int{}
 
-	rr := make([]resolverAndDelay, len(ipps))
-	for _, ipp := range ipps {
-		ip := ipp.IP()
-		if host, ok := knownDoH[ip]; ok {
+	rr := make([]resolverAndDelay, len(resolvers))
+	for _, r := range resolvers {
+		if host, ok := knownDoH[r.Addr]; ok {
```
bradfitz, line 213: it's cheaper to keep knownDoH keyed as netaddr.IP as it was but instead just ParseIP here... the parsing doesn't allocate, but the ip.String() in the map lookup later does.

net/dns/config.go
```diff
@@ -65,10 +66,21 @@ func (c Config) hasRoutes() bool {
 	return len(c.Routes) > 0
 }
 
-// hasDefaultResolversOnly reports whether the only resolvers in c are
-// DefaultResolvers.
-func (c Config) hasDefaultResolversOnly() bool {
-	return c.hasDefaultResolvers() && !c.hasRoutes()
+// hasDefaultIPResolversOnly reports whether the only resolvers in c are
+// DefaultResolvers, and that those resovlers are simple IP addresses.
```
bradfitz, line 70: typo: resolvers

net/dns/manager.go
```diff
@@ -100,9 +101,12 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
 		// case where cfg is entirely zero, in which case these
 		// configs clear all Tailscale DNS settings.
 		return rcfg, ocfg, nil
-	case cfg.hasDefaultResolversOnly():
+	case cfg.hasDefaultIPResolversOnly():
 		// Trivial CorpDNS configuration, just override the OS
 		// resolver.
+		// TODO: for OSs that support it, pass IP:port and DoH
```
bradfitz, line 107: OSes

net/dns/manager.go
```diff
@@ -159,22 +163,27 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
 		if err != nil {
 			return resolver.Config{}, OSConfig{}, err
 		}
-		rcfg.Routes["."] = toIPPorts(bcfg.Nameservers)
+		var defaultRoutes []dnstype.Resolver
+		for _, ip := range bcfg.Nameservers {
+			defaultRoutes = append(defaultRoutes, dnstype.Resolver{Addr: netaddr.IPPortFrom(ip, 53).String()})
```
bradfitz, line 168: maybe add a dnstype.ResolverFromIP constructor func?

net/dns/manager.go
```diff
@@ -159,22 +163,27 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
...
-	for _, ipp := range ipps {
-		ret = append(ret, ipp.IP())
+// toIPsOnly returns only the IP portion of dnstype.Resolver.
+// Only safe to use if the resolvers slice has been cleared of
+// DoH or custom-port entries with somethnig like hasDefaultIPResolversOnly.
+func toIPsOnly(resolvers []dnstype.Resolver) (ret []netaddr.IP) {
+	for _, r := range resolvers {
+		if ipp, err := netaddr.ParseIPPort(r.Addr); err == nil && ipp.Port() == 53 {
+			ret = append(ret, ipp.IP())
+		} else if ip, err := netaddr.ParseIP(r.Addr); err == nil {
```
bradfitz, line 184: redundantly parsing r.Addr. do it once earlier at top of loop body?

net/dns/manager.go
```diff
@@ -159,22 +163,27 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
...
-// always pointing at port 53.
-// https://github.com/tailscale/tailscale/issues/1666 tracks making
-// that not true, if we ever want to.
-func toIPsOnly(ipps []netaddr.IPPort) (ret []netaddr.IP) {
-	ret = make([]netaddr.IP, 0, len(ipps))
-	for _, ipp := range ipps {
-		ret = append(ret, ipp.IP())
+// toIPsOnly returns only the IP portion of dnstype.Resolver.
+// Only safe to use if the resolvers slice has been cleared of
+// DoH or custom-port entries with somethnig like hasDefaultIPResolversOnly.
```
bradfitz, line 179: typo: something

net/dns/manager.go
crawshaw: I don't follow. The first parse is for IP:port, the second is for a bare IP.

crawshaw:
PTAL

ipn/ipnlocal/local.go
```diff
@@ -1898,12 +1888,17 @@ func (b *LocalBackend) authReconfig() {
...
-	if err != nil {
-		return netaddr.IPPort{}, fmt.Errorf("[unexpected] non-IP resolver %q", cfg.Addr)
+func parseResolver(cfg dnstype.Resolver) dnstype.Resolver {
+	if ip, err := netaddr.ParseIP(cfg.Addr); err == nil {
+		// Add 53 here for bare IPs for consistency with previous data type.
+		return dnstype.Resolver{
+			Addr: netaddr.IPPortFrom(ip, 53).String(),
+		}
+	}
+	return dnstype.Resolver{
```
bradfitz, line 1898: return cfg?

ipn/ipnlocal/local.go
```diff
@@ -1898,12 +1888,17 @@ func (b *LocalBackend) authReconfig() {
 	b.initPeerAPIListener()
 }
 
-func parseResolver(cfg dnstype.Resolver) (netaddr.IPPort, error) {
-	ip, err := netaddr.ParseIP(cfg.Addr)
-	if err != nil {
-		return netaddr.IPPort{}, fmt.Errorf("[unexpected] non-IP resolver %q", cfg.Addr)
+func parseResolver(cfg dnstype.Resolver) dnstype.Resolver {
```
bradfitz, line 1891: hardly a parse anymore. maybe `normalizeResolver` ?

net/dns/resolver/forwarder.go
```diff
@@ -192,11 +194,11 @@ func (f *forwarder) Close() error {
 	return nil
 }
 
-// resolversWithDelays maps from a set of DNS server ip:ports (currently
-// the port is always 53) to a slice of a type that included a
-// startDelay. So if ipps contains e.g. four Google DNS IPs (two IPv4
-// + twoIPv6), this function partition adds delays to some.
-func resolversWithDelays(ipps []netaddr.IPPort) []resolverAndDelay {
+// resolversWithDelays maps from a set of DNS server names to a slice of
+// a type that included a startDelay. So if ipps contains e.g. four
```
bradfitz, line 198: there is no `ipps` in scope anymore

net/dns/resolver/forwarder.go
```diff
@@ -317,11 +338,24 @@ func (f *forwarder) getDoHClient(ip netaddr.IP) (urlBase string, c *http.Client,
 				if !strings.HasPrefix(netw, "tcp") {
 					return nil, fmt.Errorf("unexpected network %q", netw)
 				}
-				c, err := nsDialer.DialContext(ctx, "tcp", net.JoinHostPort(ip.String(), "443"))
+				toDial := host
+				if _, err := netaddr.ParseIP(toDial); err != nil && len(bootstrapRes) > 0 {
+					// Need to resolve host name, and this is
```
bradfitz, line 343: I don't want to pay this resolution cost for our current set of knownDoH servers because they're all unnecessary: they have fixed IP addresses.

Only for OpenDNS etc do we need to do DNS lookups for DoH bootstrapped.

Can we do this in a later version? This is adding a new feature (and a new unnecessary cost in some cases) to a PR that should be just about changing types.

net/dns/resolver/forwarder.go
```diff
@@ -407,7 +454,11 @@ func (f *forwarder) send(ctx context.Context, fq *forwardQuery, dst netaddr.IPPo
 	fq.closeOnCtxDone.Add(conn)
 	defer fq.closeOnCtxDone.Remove(conn)
 
-	if _, err := conn.WriteTo(fq.packet, dst.UDPAddr()); err != nil {
+	udpAddr := &net.UDPAddr{
```
bradfitz, line 457: I think the UDPAddr method is easier to read than the stutter of the next line.

Revert this part? `dst.UDPAddr()` can just be `ipp.UDPAddr()`, no?

net/dns/resolver/forwarder.go
```diff
@@ -381,19 +415,32 @@ func (f *forwarder) sendDoH(ctx context.Context, urlBase string, c *http.Client,
 //
 // send expects the reply to have the same txid as txidOut.
 //
-func (f *forwarder) send(ctx context.Context, fq *forwardQuery, dst netaddr.IPPort) ([]byte, error) {
-	ip := dst.IP()
+func (f *forwarder) send(ctx context.Context, fq *forwardQuery, rr *resolverAndDelay) ([]byte, error) {
```
bradfitz, line 418: why the pointer to resolverAndDelay? It's pretty small, no?

net/dns/resolver/forwarder.go
crawshaw: It doesn't execute for knownDoH because getKnownDoHClient passes in an empty `bootstrapRes` slice so this code never runs.

The idea is we only do the resolution here if we can do a better job than DialContext (which is only if we have a bootstrapRes value.)

I'll look into pulling it out.

bradfitz: Well, even when len(bootstrapRes) == 0, aren't you still doing an implicit DNS lookup because `toDial` is the hostname?

net/dns/resolver/forwarder.go
```diff
@@ -380,20 +384,32 @@ func (f *forwarder) sendDoH(ctx context.Context, urlBase string, c *http.Client,
 // send sends packet to dst. It is best effort.
 //
 // send expects the reply to have the same txid as txidOut.
-//
-func (f *forwarder) send(ctx context.Context, fq *forwardQuery, dst netaddr.IPPort) ([]byte, error) {
-	ip := dst.IP()
+func (f *forwarder) send(ctx context.Context, fq *forwardQuery, rr resolverAndDelay) ([]byte, error) {
+	if strings.HasPrefix(rr.name.Addr, "http://") {
+		return nil, fmt.Errorf("https:// resolvers not supported yet")
```
bradfitz, line 389: still says https for http

net/dns/config.go
```diff
@@ -65,10 +66,21 @@ func (c Config) hasRoutes() bool {
...
-func (c Config) hasDefaultResolversOnly() bool {
-	return c.hasDefaultResolvers() && !c.hasRoutes()
+// hasDefaultIPResolversOnly reports whether the only resolvers in c are
+// DefaultResolvers, and that those resolvers are simple IP addresses.
+func (c Config) hasDefaultIPResolversOnly() bool {
+	if !c.hasDefaultResolvers() || c.hasRoutes() {
+		return false
+	}
+	for _, r := range c.DefaultResolvers {
+		if ipp, err := netaddr.ParseIPPort(r.Addr); err == nil && ipp.Port() == 53 {
```
bradfitz, line 76: I forget whether ParseIPPort allocs on error. Probably doesn't matter, but maybe worth a TODO. Maybe we can check the scheme prefix (or only call this Parse if it doesn't contain "://" or something)

net/dns/manager.go
bradfitz: whoops

bradfitz (approved):

PR #2594, opened 2021-08-05
net/portmapper: add sketch of a TestIGD for portmapper testing

bradfitz:


JulianKnodt:
Seems like it'll be a useful tool to prevent regressions in the future, probably we'd like to be able to snapshot what we're currently sending so we can make sure it doesn't change, at least for PCP and PMP?

JulianKnodt:
Can this be closed now?

Issue #2596, opened 2021-08-05
Forward DNS traffic to exit node


crawshaw:
Today, any device using an exit node needs to be configured to use a DNS server that is reachable via the exit node. If you setup Tailscale in its default configuration, and your machine's DNS is still your local router (192.168.1.1:53), your connection to it will break when you turn on exit nodes.

Instead, we should forward all DNS traffic to the exit node for it to resolve. We can do this by having the exit node server DNS-over-HTTP via the peer API.

DentonGentry:
Dup of https://github.com/tailscale/tailscale/issues/1713

PR #2599, opened 2021-08-05
types/dnstype: introduce new package for Resolver

crawshaw:
So the type can be used in net/dns without introducing a tailcfg
dependency.

For #2596


bradfitz (approved):
Once CI is fixed.

PR #2601, opened 2021-08-05
cmd/tailscaled: don't require root for --cleanup

josharian:
Without this, the integration tests fail locally for me:

```
--- FAIL: TestCollectPanic (7.61s)
    integration.go:74: built [tailscale.com/cmd/tailscaled tailscale.com/cmd/tailscale] in 1.59s
    integration_test.go:102: initial run: tailscaled requires root; use sudo tailscaled (or use --tun=userspace-networking)
    integration_test.go:108: cleanup failed: exit status 1: "tailscaled requires root; use sudo tailscaled (or use --tun=userspace-networking)\n"
    stuntest.go:64: STUN server shutdown
FAIL
FAIL	tailscale.com/tstest/integration	9.678s
FAIL
```

### DIFF
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -142,7 +142,7 @@ func main() {
 		os.Exit(0)
 	}
 
-	if runtime.GOOS == "darwin" && os.Getuid() != 0 && !strings.Contains(args.tunname, "userspace-networking") {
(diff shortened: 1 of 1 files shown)


bradfitz (approved):
Probably not exactly the right fix long-term (I'd rather see cleanup handled nicely on darwin), but good enough for now.

josharian:
> Probably not exactly the right fix long-term (I'd rather see cleanup handled nicely on darwin), but good enough for now.

Ack.

PR #2607, opened 2021-08-06
net/dns/resolver: use forwarded dns txid directly

### LINKED ISSUE #2597: 100.100.100.100 DNS forwarding fails when server canonicalizes responses
A user reported an instance where a query of the form `ABCD.example.com` fails to resolve through 100.100.100.100 but works correctly if sent directly to the DNS server. After examining packet captures, @danderson and I discovered that the DNS server's response modified the question section: the response referred to `abcd.example.com`.

For historical reasons, we hash the DNS question and combine that with the original DNS transaction ID, so if the response's question is modified, DNS forwarding will drop the packet.

In the logs, this appears as `dns: error: txid doesn't match`. If you encounter this error, as a workaround try using a completely lowercase hostname. Failing that, if you can capture the response packet from the server (e.g. with tcpdump or wireshark), look the contents of the question section of the server's response.

The workaround is somewhat awkward, instead we should solve the problem by simply passing the original transaction ID through rather than performing hashing -- it is not needed with the current implementation.

Front logo [Front conversations](https://app.frontapp.com/open/top_3g029)

sailorfrag:
Previously, we hashed the question and combined it with the original
txid which was useful when concurrent queries were multiplexed on a
single local source port. We encountered some situations where the DNS
server canonicalizes the question in the response (uppercase converted
to lowercase in this case). This includes a new test to cover that
situation.

Fixes #2597

### DIFF
--- a/net/dns/resolver/forwarder.go
+++ b/net/dns/resolver/forwarder.go
@@ -10,7 +10,6 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"hash/crc32"
 	"io"
 	"io/ioutil"
 	"math/rand"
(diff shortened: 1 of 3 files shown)


net/dns/resolver/forwarder.go
```diff
@@ -65,44 +64,12 @@ func getTxID(packet []byte) txid {
...
-			// Corrupt packet; don't crash.
-			return txid(dnsid)
-		}
-	}
-
-	hash := crc32.ChecksumIEEE(packet[headerBytes:offset])
-	return (txid(hash) << 32) | txid(dnsid)
+	// Previously, we hashed the question and combined it with the original txid
+	// which was useful when concurrent queries were multiplexed on a single
+	// local source port. We encountered some situations where the DNS server
```
bradfitz, line 69: seems like this second sentence ends too early. I'm left asking, " ... and? so?"

bradfitz (approved):
LGTM other than the comment nit

danderson (approved):

DentonGentry:
This one fixes a customer-impacting problem, we should build an unstable once submitted and cherry-pick it for the next 1.12 release.

PR #2609, opened 2021-08-06
net/portmapper: add PCP tests

JulianKnodt:
This adds a PCP test to the IGD test server, by hardcoding in a few observed packets from
Denton's box.

### DIFF
--- a/net/portmapper/igd_test.go
+++ b/net/portmapper/igd_test.go
@@ -11,8 +11,10 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"sync"
+	"testing"
 
 	"inet.af/netaddr"
+	"tailscale.com/types/logger"
 )
 
 // TestIGD is an IGD (Intenet Gateway Device) for testing. It supports fake
@@ -21,13 +23,23 @@ type TestIGD struct {
 	upnpConn net.PacketConn // for UPnP discovery
 	pxpConn  net.PacketConn // for NAT-PMP and/or PCP
 	ts       *httptest.Server
+	logf     logger.Logf
+
+	// do* will log which packets are sent, but will not reply to unexpected packets.
 
 	doPMP  bool
 	doPCP  bool
-	doUPnP bool // TODO: more options for 3 flavors of UPnP services
+	doUPnP bool
 
 	mu       sync.Mutex // guards below
 	counters igdCounters
+}
+
+// TestIGDOptions are options
+type TestIGDOptions struct {
+	PMP  bool
+	PCP  bool
+	UPnP bool // TODO: more options for 3 flavors of UPnP services
 }
 
 type igdCounters struct {
@@ -38,15 +50,21 @@ type igdCounters struct {
 	numPMPDiscoRecv      int32
 	numPCPRecv           int32
 	numPCPDiscoRecv      int32
+	numPCPMapRecv        int32
+	numPCPOtherRecv      int32
 	numPMPPublicAddrRecv int32
 	numPMPBogusRecv      int32
+
+	numFailedWrites  int32
+	invalidPCPMapPkt int32
 }
 
-func NewTestIGD() (*TestIGD, error) {
+func NewTestIGD(logf logger.Logf, t TestIGDOptions) (*TestIGD, error) {
 	d := &TestIGD{
-		doPMP:  true,
-		doPCP:  true,
-		doUPnP: true,
+		logf:   logf,
+		doPMP:  t.PMP,
+		doPCP:  t.PCP,
+		doUPnP: t.UPnP,
 	}
 	var err error
 	if d.upnpConn, err = testListenUDP(); err != nil {
@@ -72,6 +90,10 @@ func (d *TestIGD) TestPxPPort() uint16 {
 
 func (d *TestIGD) TestUPnPPort() uint16 {
 	return uint16(d.upnpConn.LocalAddr().(*net.UDPAddr).Port)
+}
+
+func testIPAndGateway() (gw, ip netaddr.IP, ok bool) {
+	return netaddr.IPv4(127, 0, 0, 1), netaddr.IPv4(1, 2, 3, 4), true
 }
 
 func (d *TestIGD) Close() error {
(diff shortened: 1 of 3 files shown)


net/portmapper/igd_test.go
```diff
@@ -151,5 +171,69 @@ func (d *TestIGD) handlePMPQuery(pkt []byte, src netaddr.IPPort) {
...
+		d.pxpConn.WriteTo(examplePCPMapResponse, src.UDPAddr())
+	default:
+		// unknown op code, ignore it for now.
+		d.inc(&d.counters.numPCPOtherRecv)
+		return
+	}
+}
+
+func TestPCPIGD(t *testing.T) {
+	t.Skip()
```
JulianKnodt, line 204: Note, I'm skipping this test cause it passes locally but doesn't pass in CI and I'm not sure why

net/portmapper/igd_test.go
```diff
@@ -88,6 +99,9 @@ func (d *TestIGD) serveUPnPDiscovery() {
 	buf := make([]byte, 1500)
 	for {
 		n, src, err := d.upnpConn.ReadFrom(buf)
+		if !d.doUPnP {
```
bradfitz, line 102: you should move this later so you still increment the counter but don't just don't reply.

then we can test whether clients are behaving (e.g. not sending when we don't want them to send).

net/portmapper/igd_test.go
```diff
@@ -124,8 +138,14 @@ func (d *TestIGD) servePxP() {
 		default:
 			continue
 		case pmpVersion:
+			if !d.doPMP {
```
bradfitz, line 141: same

net/portmapper/pcp_test.go
```diff
@@ -12,6 +12,8 @@ import (
 
 var examplePCPMapResponse = []byte{2, 129, 0, 0, 0, 0, 28, 32, 0, 2, 155, 237, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 129, 112, 9, 24, 241, 208, 251, 45, 157, 76, 10, 188, 17, 0, 0, 0, 4, 210, 4, 210, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 135, 180, 175, 246}
 
+var examplePCPDiscoResponse = []byte{2, 128, 0, 0, 0, 0, 0, 0, 0, 5, 39, 196, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
```
bradfitz, line 15: I'd rather not hard-code this whole response packet. I'd rather see us build replies.

net/portmapper/igd_test.go
bradfitz: well, let's figure that out then.

what failures do you see on CI?

bradfitz:
Okay, you can rebase this now.

net/portmapper/igd_test.go
```diff
@@ -21,10 +24,13 @@ type TestIGD struct {
...
 	pxpConn  net.PacketConn // for NAT-PMP and/or PCP
 	ts       *httptest.Server
+	t        *testing.T
 
-	doPMP  bool
-	doPCP  bool
-	doUPnP bool // TODO: more options for 3 flavors of UPnP services
+	// do* will log which packets are sent, but will not reply to unexpected packets.
+
+	doPMP  syncs.AtomicBool
```
bradfitz, line 31: why do these need to be atomic? they're written before they're ever read, no?

net/portmapper/igd_test.go
```diff
@@ -38,16 +44,19 @@ type igdCounters struct {
...
 	numPCPRecv           int32
 	numPCPDiscoRecv      int32
+	numPCPMapRecv        int32
+	numPCPOtherRecv      int32
 	numPMPPublicAddrRecv int32
 	numPMPBogusRecv      int32
 }
 
-func NewTestIGD() (*TestIGD, error) {
+func NewTestIGD(t *testing.T) (*TestIGD, error) {
```
bradfitz, line 53: Is t only used for logging? Maybe it should be a `logf logger.Logf` instead then? Then people can pass in `logger.Discard` to shut it up.

net/portmapper/igd_test.go
```diff
@@ -164,5 +184,89 @@ func (d *TestIGD) handlePMPQuery(pkt []byte, src netaddr.IPPort) {
...
+	}
+	switch op {
+	case pcpOpAnnounce:
+		d.inc(&d.counters.numPCPDiscoRecv)
+		if !d.doPCP.Get() {
+			return
+		}
+		resp := buildPCPDiscoResponse(pkt)
+		if _, err := d.pxpConn.WriteTo(resp, src.UDPAddr()); err != nil {
+			d.t.Errorf("failed to write disco reply: %v", err)
```
bradfitz, line 207: oh, there are failures deep in here. Maybe just increment a counter instead and then the caller can fail on it.

I kinda wanted to keep this test stuff independent of the testing package in case we want to reuse it later in, say, integration tests components where there might not be a *testing.T if we're running this in a different process.

net/portmapper/igd_test.go
```diff
@@ -164,5 +184,89 @@ func (d *TestIGD) handlePMPQuery(pkt []byte, src netaddr.IPPort) {
...
+		}
+	case pcpOpMap:
+		if len(pkt) < 60 {
+			d.t.Errorf("Got too short packet for op map: %v", pkt)
```
bradfitz, line 211: mention PCP

bradfitz:
In the commit message:

* PCP instead of pcp
* mention test/testing instead of igd (which is a bizarre rare term)

So, like: `net/portmapping: add PCP tests`

net/portmapper/igd_test.go
JulianKnodt: I don't think they need to be, but the linux-race test reports it as a race, so I just changed it to atomic

net/portmapper/igd_test.go
JulianKnodt: I added some points where it errors in the tester if it fails to write, but that should be an error on both ends.

net/portmapper/igd_test.go
```diff
@@ -164,5 +188,90 @@ func (d *TestIGD) handlePMPQuery(pkt []byte, src netaddr.IPPort) {
...
+		t.Logf("port map changed")
+		t.Logf("have mapping: %v", c.HaveMapping())
+	})
+	c.testPxPPort = igd.TestPxPPort()
+	c.testUPnPPort = igd.TestUPnPPort()
+	c.SetGatewayLookupFunc(testIPAndGateway)
+	return c
+}
+
+func TestPCPIGD(t *testing.T) {
```
bradfitz, line 244: move this to portmapper_test so it's together with the other tests.

I want igd_test.go to be purely about the portmapper server implementation that's currently for tests but might be used for integration test environments later.

net/portmapper/igd_test.go
```diff
@@ -164,5 +188,90 @@ func (d *TestIGD) handlePMPQuery(pkt []byte, src netaddr.IPPort) {
...
+	igd, err := NewTestIGD(t.Logf)
```
bradfitz, line 245: maybe we need a more meta helper that returns both the igd and client and a cleanup func the caller can defer.

like:

```
     client, igd, cleanup := newPortmapTest(IGDOptions{PCP: true, PMP: true, UPnP: true})
     defer cleanup()
```

Then you can probably ditch the atomics since it won't be changing at runtime.

net/portmapper/igd_test.go
bradfitz: Also, the IGD suffix is a bit nerdy. Maybe just "TestICP" or "TestICPIntegration" which matches Probe.

But you could also rename TestProbeIntegration to TestProbe, first renaming TestProbe to something else

net/portmapper/igd_test.go
JulianKnodt: I don't think it needs to create both (altho I can't imagine a case where you have just the IGD test server)

net/portmapper/igd_test.go
JulianKnodt: I've changed it to `TestPCPIntegration`, hopefully that's good enough

bradfitz (approved):

PR #2614, opened 2021-08-09
hostinfo: add fly.io detection

DentonGentry:

### DIFF
--- a/hostinfo/hostinfo.go
+++ b/hostinfo/hostinfo.go
@@ -28,6 +28,7 @@ const (
 	Heroku          = EnvType("hr")
 	AzureAppService = EnvType("az")
 	AWSFargate      = EnvType("fg")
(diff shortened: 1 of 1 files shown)


hostinfo/hostinfo.go
```diff
@@ -28,6 +28,7 @@ const (
 	Heroku          = EnvType("hr")
 	AzureAppService = EnvType("az")
 	AWSFargate      = EnvType("fg")
+	FlyDotIo        = EnvType("fy")
```
bradfitz, line 31: Not sure how we got on this two letter code thing, but sure?

I'd also accept `fly` here :)

bradfitz (approved):

hostinfo/hostinfo.go
DentonGentry: Went with `fly`

PR #2615, opened 2021-08-09
net/portmapper: actually test something in TestProbeIntegration

bradfitz:
And work around GitHub Actions localhost UDP firewalls. Empirically,
localhost UDP (at least on certain ports) doesn't work on GitHub
actions. So use 0.0.0.0 there instead, as discovered by Julian.

Co-Author: @JulianKnodt

### DIFF
--- a/net/portmapper/portmapper.go
+++ b/net/portmapper/portmapper.go
@@ -14,6 +14,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"os"
 	"sync"
 	"time"
 
@@ -55,6 +56,8 @@ type Client struct {
 	logf         logger.Logf
 	ipAndGateway func() (gw, ip netaddr.IP, ok bool)
 	onChange     func() // or nil
+	testPxPPort  uint16 // if non-zero, pxpPort to use for tests
+	testUPnPPort uint16 // if non-zero, uPnPPort to use for tests
 
 	mu sync.Mutex // guards following, and all fields thereof
 
@@ -113,7 +116,8 @@ func (c *Client) HaveMapping() bool {
 //
 // All fields are immutable once created.
 type pmpMapping struct {
-	gw         netaddr.IP
+	c          *Client
+	gw         netaddr.IPPort
 	external   netaddr.IPPort
 	internal   netaddr.IPPort
 	renewAfter time.Time // the time at which we want to renew the mapping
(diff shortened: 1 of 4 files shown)


JulianKnodt:
The one thing is, it doesn't work for windows, and I have no idea how to fix that

bradfitz:
Turns out Windows runs a service on 1900 for its own UPnP stuff, which I guess makes sense, so instead this now doesn't use those well-known ports in tests.

Still doesn't work on my Windows machine yet, though. 😢

bradfitz:
(PTAL, though)

net/portmapper/portmapper.go
```diff
@@ -671,7 +690,7 @@ func (c *Client) Probe(ctx context.Context) (res ProbeResult, err error) {
 		}
 		port := addr.(*net.UDPAddr).Port
 		switch port {
-		case upnpPort:
```
JulianKnodt, line 674: Instead of upcasting these ports, can we cast the addr.Port to uint16?

net/portmapper/portmapper.go
```diff
@@ -55,6 +55,8 @@ type Client struct {
 	logf         logger.Logf
 	ipAndGateway func() (gw, ip netaddr.IP, ok bool)
 	onChange     func() // or nil
+	testPxPPort  uint16 // if non-zero, pxpPort to use
+	testUPnPPort uint16 // if non-zero, uPnPPort to use
```
JulianKnodt, line 59: can we specify in comments these are only set in tests?

net/portmapper/portmapper.go
```diff
@@ -113,7 +115,7 @@ func (c *Client) HaveMapping() bool {
 //
 // All fields are immutable once created.
 type pmpMapping struct {
-	gw         netaddr.IP
+	gw         netaddr.IPPort
```
JulianKnodt, line 118: I guess this also makes sense considering a mapping's gateway won't just be the IP but also the port to access it, the change feels a bit odd but seems more right

net/portmapper/portmapper.go
bradfitz: Done.

net/portmapper/portmapper.go
bradfitz: Done. Although I thought the field name made that clear. Now it's double clear :)

bradfitz:
The final hurdle on GitHub was netns which bound the socket to an interface that was unable to reach localhost.

PR #2616, opened 2021-08-09
[WIP] net/portmapper: fix igd tests

JulianKnodt:
[WIP] Fixing igd gateway to actually run in CI

### DIFF
--- a/net/portmapper/igd_test.go
+++ b/net/portmapper/igd_test.go
@@ -6,6 +6,7 @@ package portmapper
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"net"
 	"net/http"
@@ -13,6 +14,7 @@ import (
 	"sync"
 
 	"inet.af/netaddr"
+	"tailscale.com/net/netns"
 )
 
 // TestIGD is an IGD (Intenet Gateway Device) for testing. It supports fake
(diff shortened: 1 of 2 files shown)


net/portmapper/portmapper_test.go
```diff
@@ -74,7 +74,7 @@ func TestProbeIntegration(t *testing.T) {
 	})
 
 	c.SetGatewayLookupFunc(func() (gw, self netaddr.IP, ok bool) {
-		return netaddr.IPv4(127, 0, 0, 1), netaddr.IPv4(1, 2, 3, 4), true
+		return netaddr.IPv4(0, 0, 0, 0), netaddr.IPv4(1, 2, 3, 4), true
```
bradfitz, line 77: uh, we can dial 0.0.0.0 ?

net/portmapper/igd_test.go
```diff
@@ -49,10 +51,10 @@ func NewTestIGD() (*TestIGD, error) {
 		doUPnP: true,
 	}
 	var err error
-	if d.upnpConn, err = net.ListenPacket("udp", "127.0.0.1:1900"); err != nil {
+	if d.upnpConn, err = netns.Listener().ListenPacket(context.Background(), "udp4", ":1900"); err != nil {
```
bradfitz, line 54: this will be angry at macOS with its firewall pop-up dialogs. we really don't want to listen on non-localhost on macos.

net/portmapper/portmapper_test.go
JulianKnodt: yea not totally sure how this works... I was just permuting as many things as possible

net/portmapper/igd_test.go
JulianKnodt: I'm on mac, but haven't noticed anything from this?

JulianKnodt:
Fun fun fun https://github.com/golang/go/issues/6290

net/portmapper/igd_test.go
bradfitz: Depends on settings. You might have it off.

Screen Shot 2021-08-09 at 2 18 00 PM
Screen Shot 2021-08-09 at 2 18 07 PM

JulianKnodt: ah you're right the firewall is off

PR #2618, opened 2021-08-10
version: remove rsc.io/goversion dependency

josharian:
rsc.io/goversion is really expensive.
Running version.ReadExe on tailscaled on darwin
allocates 47k objects, almost 11mb.

All we want is the module info. For that, all we need to do
is scan through the binary looking for the magic start/end strings
and then grab the bytes in between them.

We can do that easily and quickly with nothing but a 64k buffer.


version/cmdname.go
```diff
@@ -48,3 +50,84 @@ func CmdName() string {
 	}
 	return ret
 }
+
+// findModuleInfo returns the Go module info from the executable file.
+func findModuleInfo(file string) (s string, err error) {
```
danderson, line 55: Add a test so we detect regressions if the magic strings change?

josharian: Done.

bradfitz (approved):

PR #2629, opened 2021-08-10
tsweb: also support LabelMaps from expvar.Map, without metrics

bradfitz:
We want to use tsweb to format Prometheus-style metrics from
our temporary golang.org/x/net/http2 fork, but we don't want http2
to depend on the tailscale.com module to use the concrete type
tailscale.com/metrics.LabelMap. Instead, let a expvar.Map be used
instead of it's annotated sufficiently in its name.

### DIFF
--- a/tsweb/tsweb.go
+++ b/tsweb/tsweb.go
@@ -364,18 +364,25 @@ func VarzHandler(w http.ResponseWriter, r *http.Request) {
 
 	var dump func(prefix string, kv expvar.KeyValue)
 	dump = func(prefix string, kv expvar.KeyValue) {
-		name := prefix + kv.Key
-
+		key := kv.Key
 		var typ string
+		var label string
 		switch {
 		case strings.HasPrefix(kv.Key, "gauge_"):
 			typ = "gauge"
-			name = prefix + strings.TrimPrefix(kv.Key, "gauge_")
+			key = strings.TrimPrefix(kv.Key, "gauge_")
 
 		case strings.HasPrefix(kv.Key, "counter_"):
 			typ = "counter"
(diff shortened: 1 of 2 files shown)


tsweb/tsweb.go
```diff
@@ -364,18 +364,25 @@ func VarzHandler(w http.ResponseWriter, r *http.Request) {
...
-			name = prefix + strings.TrimPrefix(kv.Key, "gauge_")
+			key = strings.TrimPrefix(kv.Key, "gauge_")
 
 		case strings.HasPrefix(kv.Key, "counter_"):
 			typ = "counter"
-			name = prefix + strings.TrimPrefix(kv.Key, "counter_")
+			key = strings.TrimPrefix(kv.Key, "counter_")
+		}
+		if strings.HasPrefix(key, "labelmap_") {
+			key = strings.TrimPrefix(key, "labelmap_")
```
dsnet, line 380: Should this be a new variable?  Otherwise, cases where it has a `labelmap_` as a prefix, but doesn't have a subsequent `_` will have mutated the key unnecessarily.

danderson (approved):

dsnet (approved):

tsweb/tsweb.go
bradfitz: I think that's fine? We kinda want to eat it at that point.

If anything I can print an error comment like some of the cases.

dsnet: Up to you. If you want to keep it as is, that's fine too.

bradfitz: Done.

PR #2630, opened 2021-08-10
net/portmapper: log upnp information

JulianKnodt:
This logs some basic statistics for UPnP, so that tailscale can better understand what routers
are being used and how to connect to them.

Would it be better to just log the body, and then parse it later to understand what types are where?

### DIFF
--- a/net/portmapper/portmapper.go
+++ b/net/portmapper/portmapper.go
@@ -714,7 +714,10 @@ func (c *Client) Probe(ctx context.Context) (res ProbeResult, err error) {
 				res.UPnP = true
 				c.mu.Lock()
 				c.uPnPSawTime = time.Now()
-				c.uPnPMeta = meta
+				if c.uPnPMeta != meta {
+					c.logf("UPnP meta changed: %+v", meta)
+					c.uPnPMeta = meta
+				}
 				c.mu.Unlock()
 			}
 		case c.pxpPort(): // same value for PMP and PCP
--- a/net/portmapper/upnp.go
+++ b/net/portmapper/upnp.go
@@ -311,6 +311,11 @@ func (c *Client) getUPnPPortMapping(
 
 type uPnPDiscoResponse struct {
 	Location string
+	// Server describes what version the UPnP is, such as MiniUPnPd/2.x.x
+	Server string
+	// USN is the serial number of the device, which also contains
+	// what kind of UPnP service is being offered, i.e. InternetGatewayDevice:2
+	USN string
 }
 
 // parseUPnPDiscoResponse parses a UPnP HTTP-over-UDP discovery response.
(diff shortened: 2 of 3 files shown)


net/portmapper/upnp.go
```diff
@@ -314,12 +314,18 @@ type uPnPDiscoResponse struct {
 }
 
 // parseUPnPDiscoResponse parses a UPnP HTTP-over-UDP discovery response.
-func parseUPnPDiscoResponse(body []byte) (uPnPDiscoResponse, error) {
+func parseUPnPDiscoResponse(body []byte, logf logger.Logf) (uPnPDiscoResponse, error) {
```
bradfitz, line 317: I don't really like pure parser functions having side effects. I'd instead parse more into uPnPDiscoResponse and then have the caller log that.

net/portmapper/upnp_test.go
```diff
@@ -50,7 +51,7 @@ func TestParseUPnPDiscoResponse(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := parseUPnPDiscoResponse([]byte(tt.headers))
+			got, err := parseUPnPDiscoResponse([]byte(tt.headers), logger.Discard)
```
bradfitz, line 54: as a bonus, these tests would then actually start testing things, rather than throwing it away

net/portmapper/upnp.go
```diff
@@ -321,5 +323,7 @@ func parseUPnPDiscoResponse(body []byte) (uPnPDiscoResponse, error) {
 		return r, err
 	}
 	r.Location = res.Header.Get("Location")
+	r.Server = res.Header.Get("Server")
+	r.USN = res.Header.Get("USN")
```
bradfitz, line 327: technically `Get("Usn")` saves an alloc here and is equivalent 😛

JulianKnodt: eh how come changing the caps removes an alloc?

net/portmapper/portmapper.go
```diff
@@ -714,7 +714,10 @@ func (c *Client) Probe(ctx context.Context) (res ProbeResult, err error) {
 				res.UPnP = true
 				c.mu.Lock()
 				c.uPnPSawTime = time.Now()
-				c.uPnPMeta = meta
+				if c.uPnPMeta != meta {
+					c.logf("[upnp-stats] meta changed: %+v", meta)
```
bradfitz, line 718: `UPnP meta changed: %+v` is fine.

I'm not a fan of the pretending-we-have-structured-logging with [foo] annotations, at least when they're one-off cases.

net/portmapper/upnp.go
```diff
@@ -311,6 +311,8 @@ func (c *Client) getUPnPPortMapping(
 
 type uPnPDiscoResponse struct {
 	Location string
+	Server   string
```
bradfitz, line 314: add example values in comments on what what these are. I have no clue what a USN is. And Server sounds like it's a URL to the server.

bradfitz (approved):

bradfitz:
LGTM; ship it

Issue #2635, opened 2021-08-11
cmd/microproxy: tracking bug for deleting


bradfitz:
The microproxy varz converter has caused too much pain. Time to delete it.

Plan:
* expand tsweb VarzHandler capabilities
* make control's have a new varz handler in Prometheus format (maybe on a separate port?)
* verify it matches microproxy's output
* adjust prometheus to not scrape the 4430 port
* turn down microproxy, run control's prometheus metrics handler on the same port that microproxy used to run on
* delete microproxy

/cc @danderson

PR #2636, opened 2021-08-11
tsweb: make VarzHandler capable of walking structs with reflect

bradfitz:
To be used by control, per linked bug's plan.

Updates #2635

### DIFF
--- a/tsweb/tsweb.go
+++ b/tsweb/tsweb.go
@@ -19,6 +19,7 @@ import (
 	_ "net/http/pprof"
 	"os"
 	"path/filepath"
+	"reflect"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


tsweb/tsweb.go
```diff
@@ -396,6 +397,29 @@ func VarzHandler(w http.ResponseWriter, r *http.Request) {
 				dump(name+"_", kv)
 			})
 			return
+		case PrometheusMetricsReflectRooter:
+			root := v.PrometheusMetricsReflectRoot()
+			rv := reflect.ValueOf(root)
+			if rv.Type().Kind() == reflect.Ptr {
+				rv = rv.Elem()
```
dsnet, line 404: and if rv is nil?

tsweb/tsweb.go
```diff
@@ -463,3 +497,33 @@ func writeMemstats(w io.Writer, ms *runtime.MemStats) {
...
+func foreachExportedStructField(rv reflect.Value, f func(fieldOrJSONName, promTyp string, rv reflect.Value)) {
+	t := rv.Type()
+	for i, n := 0, t.NumField(); i < n; i++ {
+		sf := t.Field(i)
+		name := sf.Name
+		if v := sf.Tag.Get("json"); v != "" {
+			if i := strings.Index(v, ","); i != -1 {
+				v = v[:i]
+			}
+			if v != "" {
```
dsnet, line 510: `v != "" && v != "-"`

tsweb/tsweb.go
```diff
@@ -463,3 +497,33 @@ func writeMemstats(w io.Writer, ms *runtime.MemStats) {
...
+				name = v
+			}
+		}
+		promTyp := sf.Tag.Get("metrictype")
+		if promTyp != "" || sf.Type.Kind() == reflect.Struct {
```
dsnet, line 515: Or pointer to struct?

tsweb/tsweb.go
```diff
@@ -396,6 +397,29 @@ func VarzHandler(w http.ResponseWriter, r *http.Request) {
...
+				return
+			}
+			foreachExportedStructField(rv, func(fieldOrJSONName, promTyp string, rv reflect.Value) {
+				mname := name + "_" + fieldOrJSONName
+				switch rv.Kind() {
+				case reflect.Int, reflect.Int64:
+					fmt.Fprintf(w, "# TYPE %s %s\n%s %v\n", mname, promTyp, mname, rv.Int())
+				case reflect.Float64:
+					fmt.Fprintf(w, "# TYPE %s %s\n%s %v\n", mname, promTyp, mname, rv.Float())
+				case reflect.Struct:
```
dsnet, line 417: Or pointer to struct?

tsweb/tsweb.go
dsnet: Stepping back a little. Do we really want to use the name from the `json` tag?

```go
type MyStruct {
    FooBar int    `json:"bar"`
    FizzBuzz int  `json:"-"`
}
```
For the above struct, `FooBar` will be named `bar`, but `FizzBuzz` will be named `FizzBuzz`. Seems a bit inconsistent.

tsweb/tsweb.go
bradfitz: PTAL; more cases & tests, including that.

tsweb/tsweb.go
bradfitz: yup, done in latest rev

tsweb/tsweb.go
```diff
@@ -396,6 +397,32 @@ func VarzHandler(w http.ResponseWriter, r *http.Request) {
 				dump(name+"_", kv)
 			})
 			return
+		case PrometheusMetricsReflectRooter:
+			root := v.PrometheusMetricsReflectRoot()
+			rv := reflect.ValueOf(root)
+			if rv.Type().Kind() == reflect.Ptr {
+				if rv.Pointer() == 0 {
```
dsnet, line 404: Optional: `rv.IsNil()` instead?

tsweb/tsweb.go
bradfitz: Done, with new test.

tsweb/tsweb.go
```diff
@@ -463,3 +500,42 @@ func writeMemstats(w io.Writer, ms *runtime.MemStats) {
...
+			if v != "" {
+				name = v
+			}
+		}
+		promTyp := sf.Tag.Get("metrictype")
+		if promTyp != "" || sf.Type.Kind() == reflect.Struct {
+			f(name, promTyp, rv.Field(i))
+		} else if sf.Type.Kind() == reflect.Ptr && sf.Type.Elem().Kind() == reflect.Struct {
+			fv := rv.Field(i)
+			if fv.Pointer() != 0 {
```
dsnet, line 526: Optional: fv.IsNil() instead?

tsweb/tsweb.go
```diff
@@ -396,6 +397,32 @@ func VarzHandler(w http.ResponseWriter, r *http.Request) {
...
+			}
+			foreachExportedStructField(rv, func(fieldOrJSONName, promTyp string, rv reflect.Value) {
+				mname := name + "_" + fieldOrJSONName
+				switch rv.Kind() {
+				case reflect.Int, reflect.Int64:
+					fmt.Fprintf(w, "# TYPE %s %s\n%s %v\n", mname, promTyp, mname, rv.Int())
+				case reflect.Float64:
+					fmt.Fprintf(w, "# TYPE %s %s\n%s %v\n", mname, promTyp, mname, rv.Float())
+				case reflect.Struct:
+					dump(name+"_", expvar.KeyValue{Key: fieldOrJSONName, Value: expVarPromStructRoot{rv.Addr().Interface()}})
```
dsnet, line 421: rv.Addr() only works if the parent struct is addressable. In the event that the parent struct is not a pointer, then this panics.

I believe that's possible since we permit both structs and pointer to structs at the top-level. I think the former case would trigger a panic here.

tsweb/tsweb.go
bradfitz: Yes, because we're replacing a big JSON-encoding struct of struct of struct now and they all have struct tag names today.

tsweb/tsweb.go
bradfitz: Done.

tsweb/tsweb.go
bradfitz: Done.

tsweb/tsweb.go
dsnet: SGTM.

tsweb/tsweb.go
bradfitz: Done.

dsnet (approved):

tsweb/tsweb.go
```diff
@@ -396,6 +397,37 @@ func VarzHandler(w http.ResponseWriter, r *http.Request) {
...
+				if rv.IsNil() {
+					return
+				}
+				rv = rv.Elem()
+			}
+			if rv.Type().Kind() != reflect.Struct {
+				fmt.Fprintf(w, "# skipping expvar %q; unknown root type\n", name)
+				return
+			}
+			foreachExportedStructField(rv, func(fieldOrJSONName, promTyp string, rv reflect.Value) {
```
josharian, line 413: Maybe s/promTyp/metricTyp/

tsweb/tsweb.go
```diff
@@ -396,6 +397,37 @@ func VarzHandler(w http.ResponseWriter, r *http.Request) {
...
+				mname := name + "_" + fieldOrJSONName
+				switch rv.Kind() {
+				case reflect.Int, reflect.Int64:
```
josharian, line 416: Add uint, uint64, and 32 bit integer variants. Someone will use them soon enough, and it's easy enough to get them right now.

tsweb/tsweb_test.go
```diff
@@ -421,9 +449,54 @@ func TestVarzHandler(t *testing.T) {
...
+	BarC int64 `json:"bar" metrictype:"counter"`
+	Omit int   `json:"-" metrictype:"counter"`
+}
+
+type SomeStats struct {
+	Nested       SomeNested  `json:"nestvalue"`
+	NestedPtr    *SomeNested `json:"nestptr"`
+	NestedNilPtr *SomeNested `json:"nestnilptr"`
+	CurX         int         `json:"curX" metrictype:"gauge"`
+	NoMetricType int         `json:"curX" metrictype:""`
```
josharian, line 469: s/curX/noMetric/

tsweb/tsweb.go
```diff
@@ -463,3 +505,42 @@ func writeMemstats(w io.Writer, ms *runtime.MemStats) {
 	c("frees", ms.Frees, "cumulative count of heap objects freed")
 	c("num_gc", uint64(ms.NumGC), "number of completed GC cycles")
 }
+
+func foreachExportedStructField(rv reflect.Value, f func(fieldOrJSONName, promTyp string, rv reflect.Value)) {
+	t := rv.Type()
+	for i, n := 0, t.NumField(); i < n; i++ {
+		sf := t.Field(i)
+		name := sf.Name
+		if v := sf.Tag.Get("json"); v != "" {
```
josharian, line 514: It is possible to construct pathological collisions in which a json tag is the same as a different field's name. I guess the answer is...if it hurts, don't do that?

tsweb/tsweb.go
```diff
@@ -463,3 +505,42 @@ func writeMemstats(w io.Writer, ms *runtime.MemStats) {
 	c("frees", ms.Frees, "cumulative count of heap objects freed")
 	c("num_gc", uint64(ms.NumGC), "number of completed GC cycles")
 }
+
+func foreachExportedStructField(rv reflect.Value, f func(fieldOrJSONName, promTyp string, rv reflect.Value)) {
+	t := rv.Type()
+	for i, n := 0, t.NumField(); i < n; i++ {
+		sf := t.Field(i)
+		name := sf.Name
```
josharian, line 513: Do we need any special handling of embedded fields here? I guess it's OK to have the embedded field be a part of the naming scheme.

josharian (approved):

tsweb/tsweb.go
bradfitz: done

tsweb/tsweb.go
bradfitz: yeah, that's my view

tsweb/tsweb.go
bradfitz: yup. fine by me.

PR #2638, opened 2021-08-11
wgengine: allow both fixed and random port

JulianKnodt:
Previously, we only attempted either a fixed port xor a random port. Instead, we would like to
try both (sequentially), which should provide a higher rate of success.

Context:
@DentonGentry suggested that currently we only do one of either random ports or fixed ports  Ideally, we would like to try both out, instead of having the user manually select one or the other, and then picking the best of both.

Testing this out on CI to get the integration tests.


JulianKnodt:
Closing, will probably be superseded by another impl later

PR #2643, opened 2021-08-12
ipn/ipnlocal: update requested tags in host info

### LINKED ISSUE #2641: "tailscale up --advertise-tags=tag:server" silently does nothing without --force-reauth
I just brought a server up (running 1.12.3), but forgot to tag it as a server:

```
tailscale up
```

So I went to add the tag:

```
tailscale up --advertise-tags=tag:server
```

It looked like it succeeded. No errors, but it wasn't showing up as a server on the admin console.

`tailscale debug -prefs` showed it there.

I tried:

```
tailscale up
tailscale down
```

Same. Not there yet.

So I tried:

```
tailscale up --advertise-tags=tag:server --force-reauth
```

And then it worked.

Seems like a bug. If a reauth is required to adjust tags, then we should at least make the first thing I tried generate an error telling you that it's not permitted.

/cc @apenwarr @crawshaw

MDrollette:
Fixes #2641

`Hostinfo.RequestTags` was only being set on Start which meant that modifying `--advertise-tags` on a running instance would do nothing unless a `--reset` or `--force-reauth` was also supplied.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -783,8 +783,6 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 
 	b.inServerMode = b.prefs.ForceDaemon
 	b.serverURL = b.prefs.ControlURLOrDefault()
-	hostinfo.RoutableIPs = append(hostinfo.RoutableIPs, b.prefs.AdvertiseRoutes...)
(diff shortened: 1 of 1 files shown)


MDrollette:
Is that just a CI hiccup?

bradfitz:
Thanks!

PR #2651, opened 2021-08-13
tailcfg & controlclient: add caching for Node

JulianKnodt:
This adds a sync.Pool for tailcfg.Node so that they are not allocated on every iteration. It
seems this is the only place that Nodes are being cloned, so we can hopefully save most of the
cost of cloning them all the time.

Updates tailscale/corp#2405

### DIFF
--- a/control/controlclient/map.go
+++ b/control/controlclient/map.go
@@ -123,7 +123,12 @@ func (ms *mapSession) netmapForResponse(resp *tailcfg.MapResponse) *netmap.Netwo
 	if resp.Node != nil {
 		ms.lastNode = resp.Node
 	}
-	if node := ms.lastNode.Clone(); node != nil {
+	if ms.lastNode != nil {
+		node := nm.SelfNode
+		if node == nil {
+			node = &tailcfg.Node{}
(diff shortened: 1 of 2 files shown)


bradfitz:
Sorry, this isn't necessarily safe and also doesn't look effective. It's not safe because we don't know where those Nodes might've escaped to, and it's not effective because the Puts in a loop mean that most won't be retained. Only a few will.

If we're going to reuse this memory between map updates, we don't need or want a sync.Pool at all: we have all the previous nodes already. We can just use the old ones with the same NodeIDs. But we have to make sure those update writes won't race with other parties still hanging on to those nodes. We have to make sure that's safe+documented first.

bradfitz (changes requested):
See prior comment.

JulianKnodt:
Hm ok, I'll look at reusing memory more safely then w/o the pool

JulianKnodt:
Hm I'm just going to drop this as I'm not so sure what to mark downstream as it shouldn't be updated due to races, and also it seems like it'd be a much larger change than I expected.

PR #2652, opened 2021-08-13
Add some tails and some scales

sonovawolf:


bradfitz:
You were missing a trailing newline in the file, which triggered a bug in the parser. I fixed that, but you'll need to rebase.

Also:

* add the Signed-Off-By line (git commit -s)
* commit message title should start with "words: verb ...."; see https://github.com/golang/go/wiki/CommitMessage or look at "git log" to see examples
* you'll probably want to use your corp email address as your git author, not your gmail

sonovawolf:
@bradfitz I should have incorporated your suggestions now!

Issue #2498, opened 2021-07-22
Support FreeBSD as an exit node


DentonGentry:
FreeBSD systems, including pfSense and opnSense, currently can not function as [exit nodes](https://tailscale.com/kb/1103/exit-nodes/). This issue tracks the effort to implement that support. 

<img src="https://frontapp.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_3h041)

bradfitz:
We could at least do hybrid netstack mode on FreeBSD trivially like we do on Windows, synology, etc.

Wouldn't be super efficient, but would work for lite usage.

DentonGentry:
Verified that enabling hybrid netstack mode allows my OPNsense (FreeBSD 12.1) node to function as an exit node for both IPv4 and IPv6.

bradfitz:
ship it :)

DentonGentry:
Report from the user who initially requested this, after compiling tailscaled from head for OPNsense:

_After a reboot, I'm able to use the opnSense router as an exit node!_

_I'll put some traffic through it and see if we run into any issues. Currently it seems I'm capped at around 130Mbps (tested through speedtest.net and fast.com. Home bandwidth is 600mbps but realistically get about 300, datacenter is 10gig) which I'm happy with for a full tunnel VPN connection, but I feel like it should theoretically be higher._

_I don't see any spikes in CPU or RAM usage, so we're probably good. I'll try to do some iperf tests later this week._

0xSteeW:
Hello, I'm trying to get Tailscale working as an exit node on my FreeBSD Vps, however, the log points out "some manual modifications" are needed for getting it to work. Is there any place where I can find how to do that?

0xSteeW:
Nevermind! It just works after creating a NAT from tailscale IPs to the external interface with pf.

lipidity:
@oezh using normal PF config. This kind of thing worked for me on OpenBSD: https://www.openbsd.org/faq/pf/nat.html

mats852:
> @oezh using normal PF config. This kind of thing worked for me on OpenBSD: https://www.openbsd.org/faq/pf/nat.html

Could you provide an example? That would be really appreciated! Thanks in advance.

FiloSottile:
For posterity, this is the `/etc/pf.conf` that works for me to enable exit nodes on OpenBSD 7.0.

```
set skip on lo0

[...]

PR #2655, opened 2021-08-15
Use hybrid netstack mode for FreeBSD

DentonGentry:
Allows FreeBSD to function as an exit node in the same way
that Windows and Tailscaled-on-MacOS do.

Updates https://github.com/tailscale/tailscale/issues/2498


bradfitz (approved):

bradfitz:
Maybe add a RELNOTE= line to the commit message? We're still not great at that, but it'd help.

PR #2656, opened 2021-08-15
Dockerfile: use alpine:latest.

DentonGentry:
golang:1.16-alpine is kept updates as Alpine versions
advance, we can't specify an older release without having
it break at some point due to incompatibility.

Reported by @kubeworm
https://twitter.com/kubeworm/status/1426751941519020033

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,6 +61,6 @@ RUN go install -tags=xversion -ldflags="\
       -X tailscale.com/version.GitCommit=$VERSION_GIT_HASH" \
       -v ./cmd/...
 
-FROM alpine:3.11
+FROM alpine:3.14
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -61,6 +61,6 @@ RUN go install -tags=xversion -ldflags="\
       -X tailscale.com/version.GitCommit=$VERSION_GIT_HASH" \
       -v ./cmd/...
 
-FROM alpine:3.11
+FROM alpine:latest
```
bradfitz, line 64: Using "latest" means this isn't reproducible. We can't bisect to a prior Tailscale commit and get a reproducible Docker build.

Can you just bump this up to a new version instead? (that's effectively how go.mod etc work)

bradfitz (approved):

PR #2665, opened 2021-08-17
logtail/filch: preallocate a scanner buffer

josharian:
Scanning log lines is a frequent source of allocations.
Pre-allocate a re-usable buffer.

This still doesn't help when there are giant log lines.
Those will still be problematic from an iOS memory perspective.
For more on that, see https://github.com/tailscale/corp/issues/2423.

(For those who cannot follow that link, it is a discussion
of particular problematic types of log lines for
particular categories of customers. The "categories of customers"
part is the reason that it is a private issue.)

There is also a latent bug here. If we ever encounter
a log line longer than bufio.MaxScanTokenSize,
then bufio.Scan will return an error,
and we'll truncate the file and discard the rest of the log.
That's not good, but bufio.MaxScanTokenSize is really big,
so it probably doesn't matter much in practice now.
Unfortunately, it does prevent us from easily capping the potential
memory usage here, on pain of losing log entries.

### DIFF
--- a/logtail/filch/filch.go
+++ b/logtail/filch/filch.go
@@ -30,6 +30,15 @@ type Filch struct {
 	alt       *os.File
 	altscan   *bufio.Scanner
 	recovered int64
+	// buf is an initial buffer for altscan.
+	// As of August 2021, 99.96% of all log lines
+	// are below 4096 bytes in length.
+	// Since this cutoff is arbitrary, instead of using 4096,
+	// we subtract off the size of the rest of the struct
+	// so that the whole struct takes 4096 bytes
+	// (less on 32 bit platforms).
+	// This reduces allocation waste.
+	buf [4096 - 48]byte
 }
 
 // TryReadline implements the logtail.Buffer interface.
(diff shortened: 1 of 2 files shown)


logtail/filch/filch.go
```diff
@@ -30,6 +30,15 @@ type Filch struct {
...
 	recovered int64
+	// buf is an initial buffer for altscan.
+	// As of August 2021, 99.96% of all log lines
+	// are below 4096 bytes in length.
+	// Since this cutoff is arbitrary, instead of using 4096,
+	// we subtract off the size of the rest of the struct
+	// so that the whole struct takes 4096 bytes
+	// (less on 32 bit platforms).
+	// This reduces allocation waste.
+	buf [4096 - 48]byte
```
bradfitz, line 41: add a test for unsafe.Sizeof(Filch{}) ?

josharian: Done.

bradfitz (approved):

PR #2666, opened 2021-08-17
all: prevent common large log lines

josharian:
- control/controlclient: do not periodically print full netmap
- ipn/ipnlocal: log number of packet filters rather than entire filter


josharian:
I anticipate that this may be unpopular. I won't submit until I get an OK from both @bradfitz and @danderson.

josharian:
Marking as draft, because @danderson points out that we need the u= part of the netmap line for grinder, at least in the short term.

josharian:
Ready for review @danderson @bradfitz @crawshaw

danderson (approved):

PR #2671, opened 2021-08-17
wgengine: log connection metrics

JulianKnodt:
Adds counter for connection types, which aren't currently bubbled up anywhere but can be easily
in the future.


wgengine/pendopen.go
```diff
@@ -111,6 +112,12 @@ func (e *userspaceEngine) trackOpenPostFilterOut(pp *packet.Parsed, t *tstun.Wra
 	}
 
 	flow := flowtrack.Tuple{Proto: pp.IPProto, Src: pp.Src, Dst: pp.Dst}
+	// Accepted a flow, log the flow kind
```
bradfitz, line 115: move this after the iOS check (see comment on line 122)

wgengine/userspace.go
```diff
@@ -133,6 +135,19 @@ type userspaceEngine struct {
 	// Lock ordering: magicsock.Conn.mu, wgLock, then mu.
 }
 
+type metrics struct {
+	// offline is the number of nodes magicsock attempted to connect to which are offline.
+	offline uint32
+	// relayed is the number of nodes magicsock connected thru over DERP.
```
bradfitz, line 141: "number of nodes" isn't right on these.

this is about the number of new flows

wgengine/pendopen.go
```diff
@@ -122,6 +123,13 @@ func (e *userspaceEngine) trackOpenPostFilterOut(pp *packet.Parsed, t *tstun.Wra
 		}
 	}
 
+	// Accepted a flow, log the flow kind
+	if tsaddr.IsTailscaleIP(flow.Dst.IP()) {
```
bradfitz, line 127: This is a different definition of "relayed". You're doing "using a subnet relay", but I thought what we wanted for stats was direct connects vs. DERP relays. The two are independent.

JulianKnodt: Is there a convenient way to check for DERPs here? I assume I'd need the DERP map so I guess I need to move this somewhere else

JulianKnodt: I've added a metric for derps in magicsock

PR #2676, opened 2021-08-18
tailscaled: try migrating old state on synology devices

maisem:

### DIFF
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -163,6 +163,34 @@ func main() {
 	}
 }
 
+func trySynologyMigration(p string) error {
+	if runtime.GOOS != "linux" || distro.Get() != distro.Synology {
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


DentonGentry (approved):

cmd/tailscaled/tailscaled.go
```diff
@@ -163,6 +163,34 @@ func main() {
 	}
 }
 
+func trySynologyMigration(p string) error {
+	if distro.Get() != distro.Synology {
```
bradfitz, line 167: if runtime.GOOS != "linux" || ....

Then this would func body will get compiled away on other platforms.

bradfitz (commented):
LGTM otherwise

cmd/tailscaled/tailscaled.go
maisem: oh! sweet i didn't know about this!

PR #2677, opened 2021-08-18
wgengine/router: pass in AmbientCaps when calling `ip rule`

maisem:

### DIFF
--- a/wgengine/router/runner.go
+++ b/wgengine/router/runner.go
@@ -13,6 +13,9 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"syscall"
+
+	"golang.org/x/sys/unix"
 )
 
 // commandRunner abstracts helpers to run OS commands. It exists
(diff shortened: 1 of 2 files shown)


wgengine/router/runner.go
```diff
@@ -55,7 +56,11 @@ func (o osCommandRunner) output(args ...string) ([]byte, error) {
 		return nil, errors.New("cmd: no argv[0]")
 	}
 
-	out, err := exec.Command(args[0], args[1:]...).CombinedOutput()
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		AmbientCaps: []uintptr{uintptr(12)},
```
bradfitz, line 61: 12 looks super weird.

use a named constant. one surely exists already.

also, big fat comment what's going on here and why. should we only do this on Synology? any harm in doing it on other Linux?

bradfitz: looks like `unix.CAP_NET_ADMIN` ?

maisem: Not sure if there is any harm to do it on other distros, it's safer to do Synology only

wgengine/router/runner.go
```diff
@@ -23,7 +24,9 @@ type commandRunner interface {
 	output(...string) ([]byte, error)
 }
 
-type osCommandRunner struct{}
+type osCommandRunner struct {
+	ambientCaps []uintptr
```
bradfitz, line 28: This is now a value type with pointers in it, which isn't my favorite (https://github.com/golang/go/wiki/CodeReviewComments#receiver-type)

Maybe just a `ambientCapNetAdmin bool` instead of a slice?

And add a comment regardless.

DentonGentry (approved):
LGTM (but give Brad one more look at it please)

bradfitz (approved):
Much more clear now! Thanks.

PR #2679, opened 2021-08-18
cli: web advertise exit node button

crawshaw:
A couple of gnarly assumptions in this code, as always with the async
message thing.

UI button is based on the DNS settings in the admin panel.

### DIFF
--- a/cmd/tailscale/cli/web.go
+++ b/cmd/tailscale/cli/web.go
@@ -14,6 +14,7 @@ import (
 	"flag"
 	"fmt"
 	"html/template"
+	"io"
 	"io/ioutil"
 	"log"
 	"net"
@@ -26,6 +27,7 @@ import (
 	"strings"
 
 	"github.com/peterbourgon/ff/v3/ffcli"
+	"inet.af/netaddr"
 	"tailscale.com/client/tailscale"
 	"tailscale.com/ipn"
 	"tailscale.com/tailcfg"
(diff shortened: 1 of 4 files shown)


bradfitz:
I'm happy to defer to @maisem on this review :)

cmd/tailscale/cli/web.html
```diff
@@ -86,14 +86,28 @@ <h3 class="text-3xl font-semibold mb-3">Log in</h3>
...
+		<a href="#" class="mb-4 link font-medium js-loginButton" target="_blank">Reauthenticate</a>
+	</div>
 	{{ end }}
 </main>
 <script>(function () {
-let loginButtons = document.querySelectorAll(".js-loginButton");
+const advertiseExitNode = {{.AdvertiseExitNode}};
 let fetchingUrl = false;
+var data = {
+	AdvertiseRoutes: "{{.AdvertiseRoutes}}",
```
maisem, line 105: Maybe remove this as it's not being used?

cmd/tailscale/cli/web.go
```diff
@@ -49,11 +51,13 @@ func init() {
...
-	Status       string
-	DeviceName   string
-	IP           string
+	Profile           tailcfg.UserProfile
+	SynologyUser      string
+	Status            string
+	DeviceName        string
+	IP                string
+	AdvertiseExitNode bool
+	AdvertiseRoutes   string
```
maisem, line 60: Remove this, it's not being used

maisem (approved):
I could not figure out how this worked until I actually ran the code :)

cmd/tailscale/cli/web.html
crawshaw: I tried this, but it involves unwinding a stack of other code and I'd rather not, as all of this is neatly in place to support advertising subnet routes. At this point all the "server" code is done and all that's needed on top of this PR is a UI.

crawshaw:
Updated spacing and button wording as per @sonovawolf's advice.

PR #2682, opened 2021-08-18
wgengine: predict min.Peers length across calls

josharian:
The number of peers we have will be pretty stable across time.
Allocate roughly the right slice size.
This reduces memory usage when there are many peers.

### DIFF
--- a/wgengine/userspace.go
+++ b/wgengine/userspace.go
@@ -108,6 +108,7 @@ type userspaceEngine struct {
 
 	wgLock              sync.Mutex // serializes all wgdev operations; see lock order comment below
 	lastCfgFull         wgcfg.Config
+	lastNMinPeers       int
 	lastRouterSig       deephash.Sum // of router.Config
 	lastEngineSigFull   deephash.Sum // of full wireguard config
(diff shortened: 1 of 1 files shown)


wgengine/userspace.go
```diff
@@ -606,7 +607,7 @@ func (e *userspaceEngine) maybeReconfigWireguardLocked(discoChanged map[key.Publ
 	// based on the full config. Prune off all the peers
 	// and only add the active ones back.
 	min := full
-	min.Peers = nil
+	min.Peers = make([]wgcfg.Peer, 0, e.lastNMinPeers)
```
crawshaw, line 610: this number might be `len(e.lastCfgFull.Peers)-len(e.trimmedDisco)`? maybe I'm overthinking. :)

crawshaw (approved):

bradfitz (approved):

wgengine/userspace.go
josharian: Maybe, but this logic is tricky. I'd rather keep it simple.

PR #2689, opened 2021-08-19
derp: throttle client sends if server advertises rate limits

bradfitz:
derp: throttle client sends if server advertises rate limits

### DIFF
--- a/derp/derp_client.go
+++ b/derp/derp_client.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"golang.org/x/crypto/nacl/box"
+	"golang.org/x/time/rate"
 	"tailscale.com/types/key"
 	"tailscale.com/types/logger"
 )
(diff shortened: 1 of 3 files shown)


danderson:
Sfine so far, though, we may be able to get the same result by just throttling the whole TCP connection server side, per what apenwarr was saying. I'm not sure what we'd do differently by explicitly knowing the rate limit on the client.

bradfitz:
on the client side we can prioritize disco over data

bradfitz:
PTAL

crawshaw (approved):
I believe we have a queue somewhere before this send in magicsock. If so does that make this rate limit a head drop?

bradfitz:
> I believe we have a queue somewhere before this send in magicsock. If so does that make this rate limit a head drop?

This limiter is a head drop, between draining the magicsock queue and writing to the network. The other one is a tail drop if the queue is full, like if the write blocks in the kernel due to TCP backpressure.

PR #2692, opened 2021-08-20
control/controlclient: add call to tsmpPing

simenghe:
Held back on spinning up the new goroutine to call the tsmpPing. This PR enables clients to respond with the PingResults via POST request. 

This should be merged after the 1.14 release


raggi:
Ah, just found this, patches landed recently obsolete it.

PR #2694, opened 2021-08-23
hostinfo: set DeviceModel from Linux devicetree model

bradfitz:


danderson (approved):
Maybe a comment to explain what is? Say something about the device tree being something embedded boards use, so we can use it to identify the make/model of things like rpi?

(I had to look up whether that would yield anything on my desktop, and worked out from there where it's likely to yield data)

bradfitz:
Done.

PR #2699, opened 2021-08-23
tsweb: add vars for unix process start time and version

bradfitz:
To be scraped in the Go expvar JSON format, as a string is involved.

For a future tool to record when processes restarted exactly, and at
what version.


crawshaw (approved):
Do we build all our binaries with enough magic to get version.Long populated by anything useful?

bradfitz:
> Do we build all our binaries with enough magic to get version.Long populated by anything useful?

Everything we build from corp/redo and that's running in prod, yes.

PR #2709, opened 2021-08-25
tailcfg,ipn/ipnlocal: support DNSConfig.Routes with empty values [mapver 23]

### LINKED ISSUE #2706: tailcfg: inaccurate DNSConfig.Routes docs?
@danderson, the docs on DNSConfig.Routes say:

```go
        // Routes maps DNS name suffixes to a set of DNS resolvers to                                                                                          
        // use. It is used to implement "split DNS" and other advanced DNS                                                                                     
        // routing overlays.                                                                                                                                   
        // Map keys must be fully-qualified DNS name suffixes, with a                                                                                          
        // trailing dot but no leading dot.                                                                                                                    
        Routes map[string][]dnstype.Resolver `json:",omitempty"`
```

Notably:

> must be fully-qualified DNS name suffixes, with a trailing dot but no leading dot

But the control plane sends:

```
    "DNSConfig": {
        "Routes": {
            "ts-dns.test": [
                {   
                    "Addr": "100.120.55.47"
                }
            ],
            "ts.tailscale.com": [
                {   
                    "Addr": "8.8.8.8"
                },
                {   
                    "Addr": "8.8.4.4"
                }
            ]
        },
```

Which is right?

Also, the tailcfg docs say nothing about a route with an empty value, whereas the net/dns.Config docs add:

```
        // A Routes entry with no resolvers means the route should be                                                                                          
        // authoritatively answered using the contents of Hosts.                                                                                               
        Routes map[dnsname.FQDN][]netaddr.IPPort
```

I assume that's the same semantics?

Oh, except it seems that the conversion from the tailcfg.DNSConfig to dns.Config makes that impossible, looking at `ipnlocal/local.go`:

```go
                for suffix, resolvers := range nm.DNS.Routes {
                        fqdn, err := dnsname.ToFQDN(suffix)
                        if err != nil {
                                b.logf("[unexpected] non-FQDN route suffix %q", suffix)
                        }
                        for _, resolver := range resolvers {
                                res, err := parseResolver(resolver)
                                if err != nil {
                                        b.logf(err.Error())
                                        continue
                                }
                                dcfg.Routes[fqdn] = append(dcfg.Routes[fqdn], res)
                        }
                }
```

bradfitz:
Fixes #2706
Updates #1235


bradfitz:
With this, I'm able to use our new *.corp.ts.net LetsEncrypt cert domain with ExtraRecords without breaking my home DNS bare label queries (e.g. `ping ss3`), as as now properly reply with "No such name" when querying e.g. `ss3.corp.ts.net`, as it now gets treated as a local domain for which we're authoratitive.

Screen Shot 2021-08-25 at 9 09 54 AM

(It requires the corresponding control change to send that empty value in the Routes, which I locally hacked up for testing)

PR #2711, opened 2021-08-26
wgengine/magicsock: delete legacy NAT traversal logic.

danderson:
Replace the legacy packet handling logic with an enhanced
discoEndpoint that can handle peers with no discovery key,
by communicating with them exclusively over DERP.

Co-authored-by: Josh Bleecher Snyder <josh@tailscale.com>


bradfitz:
ship it. There might be bugs but they won't be found in further review.

PR #2734, opened 2021-08-30
testcontrol: replace panic with error

crawshaw:
I have seen this once in the VM test (caused by an EOF, I believe on
shutdown) that didn't need to cause the test to fail.


bradfitz (approved):
How about just accepting io.EOF but if you read any bytes, then panic if they're invalid?

I like being super strict by default in tests.

crawshaw:
more invasive but sure

PR #2737, opened 2021-08-30
tstest/integration: prefix logs with logid

crawshaw:
The VM test has two tailscaled instances running and interleaves the
logs. Without a prefix it is impossible to figure out what is going on.

It might be even better to include the [ABCD] node prefix here as well.
Unfortunately lots of interesting logs happen before tailscaled has a
node key, so it wouldn't be a replacement for a short ID.

### DIFF
--- a/tstest/integration/integration.go
+++ b/tstest/integration/integration.go
@@ -33,6 +33,7 @@ import (
 	"go4.org/mem"
 	"tailscale.com/derp"
 	"tailscale.com/derp/derphttp"
+	"tailscale.com/logtail"
 	"tailscale.com/net/stun/stuntest"
 	"tailscale.com/smallzstd"
 	"tailscale.com/tailcfg"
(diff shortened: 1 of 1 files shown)


tstest/integration/integration.go
```diff
@@ -213,6 +214,22 @@ func (lc *LogCatcher) Reset() {
 }
 
 func (lc *LogCatcher) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// `POST /c/<collection-name>/<private-ID>`
+	if r.Method != "POST" {
+		log.Printf("bad logcatcher method: %v", r.Method)
+		http.Error(w, "only POST is supported", 400)
```
bradfitz, line 220: missing "return" after both these http.Error?

tstest/integration/integration.go
```diff
@@ -213,6 +214,22 @@ func (lc *LogCatcher) Reset() {
 }
 
 func (lc *LogCatcher) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// `POST /c/<collection-name>/<private-ID>`
```
bradfitz, line 217: remove backticks?

tstest/integration/integration.go
```diff
@@ -213,6 +214,22 @@ func (lc *LogCatcher) Reset() {
...
+	if r.Method != "POST" {
+		log.Printf("bad logcatcher method: %v", r.Method)
+		http.Error(w, "only POST is supported", 400)
+	}
+	pathParts := strings.Split(strings.TrimPrefix(r.URL.Path, "/c/"), "/")
+	if len(pathParts) != 2 {
+		log.Printf("bad logcatcher path: %q", r.URL.Path)
+		http.Error(w, "bad URL", 400)
```
bradfitz, line 225: return

bradfitz (approved):

PR #2738, opened 2021-08-30
vms: make ssh-keygen quiet by default

crawshaw:
Always succeeds. Its output clutters the log.

### DIFF
--- a/tstest/integration/vms/harness_test.go
+++ b/tstest/integration/vms/harness_test.go
@@ -96,7 +96,11 @@ func newHarness(t *testing.T) *Harness {
 	hs := &http.Server{Handler: mux}
 	go hs.Serve(ln)
 
-	run(t, dir, "ssh-keygen", "-t", "ed25519", "-f", "machinekey", "-N", ``)
+	cmd := exec.Command("ssh-keygen", "-t", "ed25519", "-f", "machinekey", "-N", "")
+	cmd.Dir = dir
+	if out, err := cmd.CombinedOutput(); err != nil {
+		t.Fatalf("ssh-keygen: %v, %s", err, out)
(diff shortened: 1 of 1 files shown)


tstest/integration/vms/harness_test.go
```diff
@@ -96,7 +96,12 @@ func newHarness(t *testing.T) *Harness {
 	hs := &http.Server{Handler: mux}
 	go hs.Serve(ln)
 
-	run(t, dir, "ssh-keygen", "-t", "ed25519", "-f", "machinekey", "-N", ``)
+	cmd := exec.Command("ssh-keygen", "-t", "ed25519", "-f", "machinekey", "-N", ``)
+	cmd.Dir = dir
+	if out, err := cmd.CombinedOutput(); err != nil {
+		t.Logf("ssh-keygen: %s", out)
```
bradfitz, line 102: just one Fatalf("ssh-keygen: %v, %s", err, out) ?  but either way.

tstest/integration/vms/harness_test.go
```diff
@@ -96,7 +96,12 @@ func newHarness(t *testing.T) *Harness {
 	hs := &http.Server{Handler: mux}
 	go hs.Serve(ln)
 
-	run(t, dir, "ssh-keygen", "-t", "ed25519", "-f", "machinekey", "-N", ``)
+	cmd := exec.Command("ssh-keygen", "-t", "ed25519", "-f", "machinekey", "-N", ``)
```
bradfitz, line 99: I prefer to only use backticks when in regexps or when they're really needed. I'd prefer `""` there, but minor.

bradfitz (approved):

PR #1919, opened 2021-05-13
ipn/store: native AWS SSM integration

mvisonneau:
This change introduces a new storage provider for the state file. It
allows users to leverage AWS SSM parameter store natively within
`tailscaled`. eg:

```shell
~$ tailscaled -state arn:aws:ssm:eu-west-1:123456789:parameter/foo
```

With this feature, users running tailscale agent over AWS won't have to tinker around or provision a 1GB volume in order to dynamically store the state file.

known limitations:
- it is not currently possible to specific a custom KMS key ID

@bradfitz, I reckon this change relates to #1851. If you are happy with the less invasive decomposition I have done in here I believe we can close #1851 :shrug:

PR #2742, opened 2021-08-30
ipn/store: add ability to store data as k8s secrets.

maisem:


crawshaw:
Might be good to get a K8S expert in here too (@danderson, @bradfitz?) but happy to review tonight.

cmd/tailscaled/tailscaled.go
```diff
@@ -115,6 +116,7 @@ func main() {
 	flag.StringVar(&args.socketpath, "socket", paths.DefaultTailscaledSocket(), "path of the service unix socket")
 	flag.StringVar(&args.birdSocketPath, "bird-socket", "", "path of the bird unix socket")
 	flag.BoolVar(&printVersion, "version", false, "print version information and exit")
+	flag.StringVar(&args.kubesecret, "kube-secret", "", "use kubernetes secrets as the state store")
```
bradfitz, line 119: this description sounds like a bool argument.  What is the string value? Also, capitalize Kubernetes, no?

bradfitz: But from https://github.com/tailscale/tailscale/pull/1919 I think the plan we were somewhat converging on was to use magic state path prefixes to mean different state backends.

In any case, we should have a plan for all three.

If we do separate flags for each, then we we need check that they're all mutually exclusive.

cmd/tailscaled/tailscaled.go
```diff
@@ -258,7 +261,7 @@ func run() error {
 		return nil
 	}
 
-	if args.statepath == "" {
+	if args.statepath == "" && args.kubesecret == "" {
```
bradfitz, line 264: and if they're both set?

ipn/ipnserver/server.go
```diff
@@ -61,6 +61,9 @@ type Options struct {
 	// StatePath is the path to the stored agent state.
 	StatePath string
 
+	// KubeSecret is the name of the kubernetes secret to use.
```
bradfitz, line 64: // KubeSecret, if not empty, is the ....

And Kubernetes capitalized.

And what's it used for? And what's the relation between this and StatePath?

kube/client.go
```diff
@@ -0,0 +1,169 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package kube provides a client to interact with Kubernetes.
```
bradfitz, line 5: Bit fat warning that this is Tailscale-internal and not meant for external consumption and is not a stable API.

kube/client.go
```diff
@@ -0,0 +1,169 @@
...
+	"time"
+)
+
+const (
+	saPath     = "/var/run/secrets/kubernetes.io/serviceaccount"
+	defaultURL = "https://kubernetes.default.svc"
+)
+
+func readFile(n string) ([]byte, error) {
+	return ioutil.ReadFile(filepath.Join(saPath, n))
```
bradfitz, line 28: os.ReadFile these days

kube/client.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+	if out != nil {
+		return json.NewDecoder(resp.Body).Decode(out)
+	}
+	return nil
+}
+
+// GetSecret fetches the secret from the Kubernetes API.
+func (c *Client) GetSecret(name string) (*Secret, error) {
```
bradfitz, line 152: take a context, here and everywhere else where network requests are involved

kube/client.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+func readFile(n string) ([]byte, error) {
+	return ioutil.ReadFile(filepath.Join(saPath, n))
+}
+
+// Client handles connections to Kubernetes.
+// It expects to be run inside a cluster.
+type Client struct {
+	mu          sync.RWMutex
```
bradfitz, line 34: just a regular sync.Mutex

This doesn't look like one of the rare times that RWMutex actually does better. You're not holding the RLock for long periods of time.

cmd/tailscaled/tailscaled.go
```diff
@@ -78,6 +78,7 @@ var args struct {
 	port           uint16
 	statepath      string
 	socketpath     string
+	kubesecret     string
```
crawshaw, line 81: we've got an odd mix of styles here, but given the flag is `kube-secret`, I think this may as well be `kubeSecret`

docs/k8s/role.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
crawshaw, line 1: could this file (or this directory in a readme maybe) have some more pointers/instructions? maybe just a pointer to a KB article?

kube/client.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return nil, err
+	}
+	caCert, err := readFile("ca.crt")
+	if err != nil {
+		return nil, err
+	}
+	cp := x509.NewCertPool()
+	if ok := cp.AppendCertsFromPEM(caCert); !ok {
+		return nil, fmt.Errorf("error in creating root cert pool")
```
crawshaw, line 54: `kube:` in the error so we know where it came from. (The other errors give it away because they include the file path.)

crawshaw (commented):
Agree with Brad we should have a somewhat-unified story for picking state stores

docs/k8s/role.yaml
```diff
@@ -0,0 +1,23 @@
...
+kind: Role
+metadata:
+  namespace: default
+  name: tailscale
+rules:
+- apiGroups: [""] # "" indicates the core API group
+  resources: ["secrets"]
+  verbs: ["create", "get", "update"]
+  resouceNames: ["tailscale"]
```
mayakacz, line 10:
```suggestion
  resourceNames: ["tailscale"]
```

cmd/tailscaled/tailscaled.go
maisem: Moved to `statepath` with `kube:` as a prefix

cmd/tailscaled/tailscaled.go
maisem: moved to a single flag

cmd/tailscaled/tailscaled.go
```diff
@@ -111,7 +111,7 @@ func main() {
 	flag.StringVar(&args.socksAddr, "socks5-server", "", `optional [ip]:port to run a SOCK5 server (e.g. "localhost:1080")`)
 	flag.StringVar(&args.tunname, "tun", defaultTunName(), `tunnel interface name; use "userspace-networking" (beta) to not use TUN`)
 	flag.Var(flagtype.PortValue(&args.port, 0), "port", "UDP port to listen on for WireGuard and peer-to-peer traffic; 0 means automatically select")
-	flag.StringVar(&args.statepath, "state", paths.DefaultTailscaledStateFile(), "path of state file")
+	flag.StringVar(&args.statepath, "state", paths.DefaultTailscaledStateFile(), "path of state file; prefix with 'kube:' to use Kubernetes secrets")
```
bradfitz, line 114: same comment as earlier: what is the suffix after kube: ?

ipn/ipnserver/server.go
```diff
@@ -613,9 +613,18 @@ func Run(ctx context.Context, logf logger.Logf, logid string, getEngine func() (
 
 	var store ipn.StateStore
 	if opts.StatePath != "" {
-		store, err = ipn.NewFileStore(opts.StatePath)
-		if err != nil {
-			return fmt.Errorf("ipn.NewFileStore(%q): %v", opts.StatePath, err)
+		const kubePrefix = "kube:"
+		switch {
+		case strings.HasPrefix(opts.StatePath, kubePrefix):
+			store, err = ipn.NewKubeStore(opts.StatePath[len(kubePrefix):])
```
bradfitz, line 619: I'd just use strings.TrimPrefix for clarity/consistency with the line above.

It'll be imperceptible slower but won't even allocate.

bradfitz: Also, assign it to a variable so it's clear what the suffix _is_

```go
    dingus := strings.TrimPrefix(opts.StatePath, kubePrefix)
    store, err := ipn.NewKubeStore(dingus)
```

ipn/ipnserver/server.go
```diff
@@ -613,9 +613,18 @@ func Run(ctx context.Context, logf logger.Logf, logid string, getEngine func() (
...
+			if err != nil {
+				return fmt.Errorf("ipn.NewKubeStore(%q): %v", opts.StatePath, err)
```
bradfitz, line 621: technically ipn.NewKubeStore was called with `dingus`, not `opts.StatePath`.

another reason for a variable.

ipn/store.go
```diff
@@ -55,6 +58,75 @@ type StateStore interface {
...
 }
 
+// KubeStore is a StateStore that uses a Kubernetes Secret for persistence.
+type KubeStore struct {
+	client     *kube.Client
+	secretName string
+}
+
+// NewKubeStore returns a new KubeStore that persists to the provided secret.
+func NewKubeStore(secret string) (*KubeStore, error) {
```
bradfitz, line 68: secretName ?

ipn/store.go
```diff
@@ -55,6 +58,75 @@ type StateStore interface {
 	WriteState(id StateKey, bs []byte) error
 }
 
+// KubeStore is a StateStore that uses a Kubernetes Secret for persistence.
+type KubeStore struct {
+	client     *kube.Client
+	secretName string
+}
+
+// NewKubeStore returns a new KubeStore that persists to the provided secret.
```
bradfitz, line 67: to the named secret?

kube/client.go
```diff
@@ -0,0 +1,170 @@
...
+func (c *Client) GetSecret(ctx context.Context, name string) (*Secret, error) {
+	s := &Secret{Data: make(map[string][]byte)}
+	if err := c.doRequest(ctx, "GET", c.secretURL(name), nil, s); err != nil {
+		return nil, err
+	}
+	return s, nil
+}
+
+// CreateSecret creates a secret in the Kubernetes API.
+func (c *Client) CreateSecret(ctx context.Context, in *Secret) error {
```
bradfitz, line 162: `s` instead of `in`?

Go doesn't typically annotate pointers as in or out parameters unless they violate convention.

bradfitz (approved):

docs/k8s/role.yaml
maisem: I added a readme file. Will update with a link when we have a KB article

PR #2744, opened 2021-08-30
wgengine/magicsock: misc trivial cleanups

danderson:


danderson:
I suggest reviewing this commit-by-commit. Each one is a trivial change, but some of the trivial changes are mechanical renamings, so the overall squashed diff looks noisy compared to individual commit diffs.

bradfitz (approved):

Issue #2741, opened 2021-08-30
wgengine/netstack: remove alloc-per-packet


bradfitz:
The netstack mode data path includes this full copy+alloc per packet:

```go
func (ns *Impl) injectOutbound() {
        for {
                packetInfo, ok := ns.linkEP.ReadContext(context.Background())
                if !ok {
                        ns.logf("[v2] ReadContext-for-write = ok=false")
                        continue
                }
                pkt := packetInfo.Pkt
                hdrNetwork := pkt.NetworkHeader()
                hdrTransport := pkt.TransportHeader()

                full := make([]byte, 0, pkt.Size())
                full = append(full, hdrNetwork.View()...)
                full = append(full, hdrTransport.View()...)
                full = append(full, pkt.Data().AsRange().AsView()...)
                if debugNetstack {
                        ns.logf("[v2] packet Write out: % x", full)
                }
                if err := ns.tundev.InjectOutbound(full); err != nil {
                        log.Printf("netstack inject outbound: %v", err)
                        return
                }

        }
}
```

That was lazy get-it-working code and known to be inefficient at the time, but we haven't gotten back to improving it yet.

/cc @stapelberg @crawshaw

PR #2747, opened 2021-08-31
wgengine/netstack: switch injectOutbound to full-duplex pipes

ok-john:
Thought I'd take a quick crack at #2741

I use a pipe (two synchronous, full-duplex *net.Conn ) that gets allocated once per call to injectUnbound(). 

The tests never failed while writing this; so I'm somewhat sus about it. 

I added a deadline just out of habit; 30 seconds is completely arbitrary, likely needs changing.
```go
// Deadline is optional - thoughts?
d, n := time.Second*30, time.Now()
r, w := net.Pipe()
r.SetDeadline(n.Add(d))
w.SetDeadline(n.Add(d))
```

I left the buffer allocation for the full packet.
```go
full := make([]byte, 0, pkt.Size())
```

But instead of performing appends to full we write to the raw conn `w` in a new go routine.
```go
go func() {
	w.Write(pkt.Data().AsRange().AsView())
	w.Write(pkt.NetworkHeader().View())
	w.Write(pkt.TransportHeader().View())
}()

```

The successive read from the `r` conn will block until all writes have been consumed.
```go
if wrote, err := r.Read(full); debugNetstack {
        ...
}
```

BTW how do you generally benchmark the netstack package? 

I'm not saying this is better (haven't benched it) - but the heavy use of `append` as you said is "get-it-working code" so I'd wager most implementations that use reader/writer or `copy` should be an improvement.

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -405,28 +405,36 @@ func (ns *Impl) DialContextUDP(ctx context.Context, addr string) (*gonet.UDPConn
 }
 
 func (ns *Impl) injectOutbound() {
-	for {
-		packetInfo, ok := ns.linkEP.ReadContext(context.Background())
-		if !ok {
+	// Deadline is optional - thoughts?
+	d, n := time.Second*30, time.Now()
+	r, w := net.Pipe()
+	r.SetDeadline(n.Add(d))
+	w.SetDeadline(n.Add(d))
+	defer r.Close()
+	defer w.Close()
+	for packetInfo, ok := ns.linkEP.ReadContext(context.Background()); ; packetInfo, ok = ns.linkEP.ReadContext(context.Background()) {
+		if !ok { // thoughts on getting rid of this branch?
 			ns.logf("[v2] ReadContext-for-write = ok=false")
 			continue
 		}
 		pkt := packetInfo.Pkt
-		hdrNetwork := pkt.NetworkHeader()
-		hdrTransport := pkt.TransportHeader()
-
 		full := make([]byte, 0, pkt.Size())
-		full = append(full, hdrNetwork.View()...)
-		full = append(full, hdrTransport.View()...)
-		full = append(full, pkt.Data().AsRange().AsView()...)
-		if debugNetstack {
-			ns.logf("[v2] packet Write out: % x", full)
+		go func() {
+			w.Write(pkt.Data().AsRange().AsView())
+			w.Write(pkt.NetworkHeader().View())
+			w.Write(pkt.TransportHeader().View())
+		}()
+		if wrote, err := r.Read(full); debugNetstack {
+			// I didn't add a nil check for this error
+			// as InjectUnbound throws one. Idea being that less branches in this function the better
+			// not sure though - thoughts?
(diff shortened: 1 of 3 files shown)


ok-john:
Interesting - so I added another [PR on my fork ](https://github.com/ok-john/tailscale/pull/1)- all tests pass except for this [magicsock windows test](https://github.com/ok-john/tailscale/pull/1/checks?check_run_id=3468696159#step:5:485).

wgengine/netstack/netstack.go
```diff
@@ -405,28 +405,36 @@ func (ns *Impl) DialContextUDP(ctx context.Context, addr string) (*gonet.UDPConn
...
+	for packetInfo, ok := ns.linkEP.ReadContext(context.Background()); ; packetInfo, ok = ns.linkEP.ReadContext(context.Background()) {
+		if !ok { // thoughts on getting rid of this branch?
 			ns.logf("[v2] ReadContext-for-write = ok=false")
 			continue
 		}
 		pkt := packetInfo.Pkt
-		hdrNetwork := pkt.NetworkHeader()
-		hdrTransport := pkt.TransportHeader()
-
 		full := make([]byte, 0, pkt.Size())
```
bradfitz, line 421: This alloc per packet is still here, so how does this help the optimization bug?

ok-john: > This alloc per packet is still here,

Unless a buffer is provided to this function to handle each packet than it will need to allocate one, no?

> so how does this help the optimization bug?

So both implementations allocate a buffer of size `full`.

```go
full := make([]byte, 0, pkt.Size())
```

The current implementation uses append - these calls will re-allocate `full` multiple times (to my understanding), in general I've thought it best to use `io.Copy`, `copy` or `subtle.ConstantTimeCopy` when populating buffers over append, but, I did not spend a decade developing the language so please do tell me if I'm wrong.

```go
full = append(full, hdrNetwork.View()...)
full = append(full, hdrTransport.View()...)
full = append(full, pkt.Data().AsRange().AsView()...)
```

My implementation we're trading this reallocation of the `full` buffer for multiple writes to the in-memory pipe.

```go
go func() {
	w.Write(pkt.Data().AsRange().AsView())
	w.Write(pkt.NetworkHeader().View())
	w.Write(pkt.TransportHeader().View())
}()
```

Then we populate the empty, already allocated buffer `full` with a single read from the Pipe, avoiding the resizing or cost associated with the use of append. 

```go
r.Read(full)
```

Like I mentioned though I didn't benchmark this code, I simply was scoping out your repository, saw the open issue and was like oh huh _maybe you could use a pipe_. I raised the PR as the tests all passed on my end and I figured you'd have some easy CI/CD for bench-marking changes to `netstack`.

ok-john: Is packet-size a constant here?
```go
pkt := packetInfo.Pkt
full := make([]byte, 0, pkt.Size())
```
Is then sum of lengths across these three values always == `pkt.Size()` as well?
```go
w.Write(pkt.Data().AsRange().AsView())
w.Write(pkt.NetworkHeader().View())
w.Write(pkt.TransportHeader().View())
```

Issue #2756, opened 2021-08-31
derp: add graceful shutdown frame type


bradfitz:
Add a DERP frame that the server can send to clients that says, "I'm shutting down for a restart. You should reconnect in N milliseconds, trying for M ms rotal, without picking a new DERP home."

Then we can gracefully restart derper for binary updates without big swings in where users are for 5-15 minutes and without the corresponding control/map update flood.

The "N milliseconds" part then also lets us potentially smooth out the reconnects, if that ends up mattering. In practice we can just make N and M be 0 ms and 5,000 ms or something.

bradfitz:
The type exists at this point (as of Aug 31) but the client just ignores it in magicsock and the server doesn't ever send it.

DentonGentry:
For Tailscalars reading along, search for a "DERP-to-client: Begone!" eng doc.

Issue #2746, opened 2021-08-31
derp: add "health" frame


bradfitz:
So derp server can report an unhealthy state (like dup keys) to the derp client.

bradfitz:
Client part done.

Server doesn't yet send it, though.

PR #2760, opened 2021-08-31
derp: add new health update and server restarting frame types

bradfitz:
Updates #2756
Updates #2746

### DIFF
--- a/derp/derp.go
+++ b/derp/derp.go
@@ -101,6 +101,20 @@ const (
 
 	framePing = frameType(0x12) // 8 byte ping payload, to be echoed back in framePong
 	framePong = frameType(0x13) // 8 byte payload, the contents of the ping being replied to
+
+	// frameHealth is sent from server to client to tell the client
+	// if their connection is unhealthy somehow. Currently the only unhealthy state
+	// is whether the connection is detected as a duplicate.
+	// The entire frame body is the text of the error message. An empty message
+	// clears the error state.
+	frameHealth = frameType(0x14)
+
+	// frameRestarting is sent from server to client for the
+	// server to declare that it's restarting. Payload is two big
+	// endian uint32 durations in milliseconds: when to reconnect,
+	// and how long to try total. See ServerRestartingMessage docs for
+	// more details on how the client should interpret them.
+	frameRestarting = frameType(0x15)
 )
 
 var bin = binary.BigEndian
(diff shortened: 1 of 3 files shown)


derp/derp.go
```diff
@@ -101,6 +101,19 @@ const (
...
+	// if they connection is unhealthy somehow. Currently the only unhealthy state
+	// is whether the connection is detected as a duplicate.
+	// The entire frame body is the text of the error message. An empty message
+	// clears the error state.
+	frameHealth = frameType(0x14)
+
+	// frameRestarting is sent from server to client for the
+	// server to declare that it's restarting. Payload is two big
+	// endian uint32 duration in milliseconds: when to reconnect,
+	// and how long to try total.
```
dsnet, line 115: What's the purpose of "how long to try total"? If that time period expires, what's the alternative? And if there is a valid alternative, is the client allowed to try that alternative before this period as ended?

derp/derp_client.go
```diff
@@ -369,6 +370,38 @@ type KeepAliveMessage struct{}
...
+// connection health state.
+type HealthMessage struct {
+	// Problem, if non-empty, is a description of why the connection
+	// is unhealthy.
+	//
+	// The empty string means the connection is healthy again.
+	//
+	// The default condition is healthy, so the server doesn't
+	// broadcast a HealthMessage until a problem exists.
+	Problem string
```
crawshaw, line 383: In programming we have a thousand ways to say something is wrong, but not one way to say everything is fine.

derp/derp_client.go
```diff
@@ -369,6 +370,38 @@ type KeepAliveMessage struct{}
...
+
+func (HealthMessage) msg() {}
+
+// ServerRestartingMessage is a one-way message from server to client,
+// advertising that the server is restarting.
+type ServerRestartingMessage struct {
+	// ReconnectIn is an advisory duration that the client should wait
+	// before attempting to reconnect. It might be zero.
+	// It exists for the server to smear out the reconnects.
+	ReconnectIn time.Duration
```
crawshaw, line 394: This is fine, but your commit message makes it sound like the client will definitely not pick a  new home. Seems reasonable that if a server says it's going away for entire seconds to pick a new home.

crawshaw (approved):

derp/derp.go
bradfitz: It's explained more on the ServerRestartingMessage field docs. I'll reference those.

derp/derp_client.go
crawshaw: oh I probably extrapolated from https://github.com/tailscale/tailscale/issues/2756,  which I think was in the email notification above this PR.

Issue #2708, opened 2021-08-25
tailscale up should block until tailscaled is ready


rosszurowski:
In our guides for running Tailscale on serverless environments, we often use a block like this:

```sh
/app/tailscaled --state=/var/lib/tailscale/tailscaled.state --socket=/var/run/tailscale/tailscaled.sock &
until /app/tailscale up --authkey=${TAILSCALE_AUTHKEY} --hostname=fly-app
do
    sleep 0.1
done
```

We do this because we need to wait for tailscaled to spin up and be ready before we pass control on to the next commands. 

We should change the default behavior of `tailscale up` to block until `tailscaled` is ready. It seems like it should have a negligible effect on the CLI experience in most cases, and would make scripting much easier.

Crawshaw warns that there are tricky edge cases and we'll need to more accurately define what "ready" looks like.

bradfitz:
@DentonGentry and I were discussing this earlier in https://github.com/tailscale/tailscale-www/pull/291#pullrequestreview-730066992

bradfitz:
> We should change the default behavior of tailscale up to block until tailscaled is ready. It seems like it should have a negligible effect on the CLI experience in most cases, and would make scripting much easier.

Or we do the thing that's been discussed elsewhere and not require such cases to even use two binaries. Let's remove `cmd/tailscale` from the picture entirely and just teach `cmd/tailscaled` to come up fully on its own.

rosszurowski:
Will this make it into the next point release or minor release (1.14.1 vs. 1.16)? Once we have decent uptake of a new client that supports it, I'll revise our guides to remove the `until` checks.

bradfitz:
> Will this make it into the next point release or minor release (1.14.1 vs. 1.16)? Once we have decent uptake of a new client that supports it, I'll revise our guides to remove the until checks.

Added to today's eng meeting agenda to decide.

PR #2762, opened 2021-08-31
safesocket: add connect retry loop to wait for tailscaled

bradfitz:
Updates #2708


safesocket/safesocket.go
```diff
@@ -29,9 +33,54 @@ func ConnCloseWrite(c net.Conn) error {
 	return c.(closeable).CloseWrite()
 }
 
+var processStartTime = time.Now()
+
+// tailscaledStillStarting reports whether tailscaled is probably
```
maisem, line 38: This adds tailscale specific code in an otherwise generic library.

bradfitz: Do you propose we create `tailscale.com/safesocket/tailscalesafesocket` ?

bradfitz: (That was a joke.)

Look at `safesocket_darwin.go` to convince yourself that safesocket is Tailscale-specific already. :)

dsnet:
What's the rationale for this approach, which incurs a new dependency over getting `tailscaled` to start listening on a socket ASAP, and then use that channel to further communicate whether it is ready or not?

bradfitz:
> What's the rationale for this approach, which incurs a new dependency, 

We already depend on this package.

> over getting tailscaled to start listening on a socket ASAP, 

It *already does* start listening on the socket ASAP.

> and then use that channel to further communicate whether it is ready or not?

Because that's racey. That's where we're at today. A "tailscaled& tailscale" could still schedule the tailscale part to run first.

safesocket/safesocket.go
maisem: Lol well ok then

safesocket/safesocket_ps.go
```diff
@@ -0,0 +1,37 @@
...
+		}
+		for _, proc := range procs {
+			name := proc.Executable()
+			const tailscaled = "tailscaled"
+			if len(name) < len(tailscaled) {
+				continue
+			}
+			// Do case insensitive comparison for Windows,
+			// notably, and ignore any ".exe" suffix.
+			if strings.EqualFold(name[:len(tailscaled)], tailscaled) {
```
dsnet, line 30: This only works for case insensitivity in ASCII, which I assume is sufficient? For non-ASCII, `name[:len(tailscaled)]` will truncate the string if there are multi-byte Unicode characters.

dsnet (commented):
> Because that's racey. That's where we're at today. A "tailscaled& tailscale" could still schedule the tailscale part to run first.

The new logic is also still racy. Hard-coding things like <2s and >5s suggests that we're making assumptions about what a good startup looks like. That said, I guess it's better than before.

dsnet (approved):

safesocket/safesocket_ps.go
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build linux || windows || darwin || freebsd
+// +build linux windows darwin freebsd
```
maisem, line 6: openbsd?

maisem (approved):

safesocket/safesocket_ps.go
bradfitz: The OpenBSD build failure is why I pulled this out to its own file in PS2. In the first version I had it in a portable file and discovered that the ps package doesn't support OpenBSD. So oh well.

safesocket/safesocket_ps.go
bradfitz: yup, sufficient. The second arg is ASCII anyway.

Issue #1889, opened 2021-05-09
tailscale up doesn't wait for ipn.Running


tavianator:
I have a GitHub Actions runner that recently started failing once it started using 1.8.0.  After looking at the [logs](https://github.com/tavianator/bfs/runs/2539062744?check_suite_focus=true), it seems like tailscale hasn't finished connecting by the time it tries to use it.  Is there a way to make `tailscale up` wait for the connection to finish?  Or is there another command that will wait?

DentonGentry:
tailscale up does block until the network connection itself is up, but it is quite possible that the timing of when the DNS servers are configured and /etc/resolv.conf updated has changed in 1.8. There have been a number of changes in the DNS implementation for the 1.8 release. That could lead to the behavior seen of a hostname not resolving immediately after "tailscale up" completes.

tavianator:
I tried with hardcoded IPs too, but it times out: https://github.com/tavianator/bfs/runs/2538977411?check_suite_focus=true

On the other hand, running `tailscale status; tailscale netcheck` first made it work: https://github.com/tavianator/bfs/runs/2539035192?check_suite_focus=true

DentonGentry:
Ok, that's odd.

tavianator:
Running `sudo tailscale up && ping 100...` locally works fine.  The gh actions runner is doing `tailscale up --authkey=...` though.

tavianator:
My latest attempt to work around this ran into "unexpected state: Starting" from `tailscale status`: https://github.com/tavianator/bfs/runs/2605558790?check_suite_focus=true

tavianator:
So I see here that `tailscale up` seems to wait for either `Starting` or `Running`: https://github.com/tailscale/tailscale/blob/85df1b0fa7d565f5b3deddd825fd08eafbd71632/cmd/tailscale/cli/up.go#L383-L393

Should it be changed to just wait for `Running`?  If that's the right fix I can send a PR

calderonth:
I can confirm I'm seeing the same issue when running from a Github action workflow, used to work with `1.6.0` now not working.

DentonGentry:
Fixed by https://github.com/tailscale/tailscale/pull/2763
Fix will appear in Tailscale 1.16.

PR #2763, opened 2021-08-31
cmd/tailscale/cli: make up block until state Running, not just Starting

### LINKED ISSUE #2708: tailscale up should block until tailscaled is ready

bradfitz:
At "Starting", the DERP connection isn't yet up. After the first netmap
and DERP connect, then it transitions into "Running".

Fixes #2708

### DIFF
--- a/cmd/tailscale/cli/up.go
+++ b/cmd/tailscale/cli/up.go
@@ -372,8 +372,8 @@ func runUp(ctx context.Context, args []string) error {
 	c, bc, pumpCtx, cancel := connect(ctx)
 	defer cancel()
 
-	startingOrRunning := make(chan bool, 1) // gets value once starting or running
-	gotEngineUpdate := make(chan bool, 1)   // gets value upon an engine update
+	running := make(chan bool, 1)         // gets value once in state ipn.Running
+	gotEngineUpdate := make(chan bool, 1) // gets value upon an engine update
 	pumpErr := make(chan error, 1)
 	go func() { pumpErr <- pump(pumpCtx, bc, c) }()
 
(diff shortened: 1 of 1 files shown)


bradfitz:
I think this is right?

I don't remember why it previously waited for Starting too. Need to do some git archaeology.

bradfitz:
The Starting or Running goes back to the first commit: a8d8b8719 ...

```go
				case ipn.Starting, ipn.Running:
					// Done full authentication process
					cancel()
				}
```

bradfitz:
The Starting check goes back to when it was `cmd/ipn` in its very first commit: https://github.com/tailscale/corp/commit/e723ccec471004b54cd578a9dfaca2872a4e425f

```go
	bc := ipn.NewBackendClient(clientToServer)
	opts := ipn.Options{
		Prefs:      prefs,
		ServerURL:  *server,
		LoginFlags: lf,
		Notify: func(n ipn.Notify) {
			log.Printf("Notify: %v\n", n)
			if s := n.State; s != nil {
				switch *s {
				case ipn.NeedsLogin:
					bc.StartLoginInteractive()
				case ipn.NeedsMachineAuth:
					fmt.Fprintf(os.Stderr, "\nTo authorize your machine, visit (as admin):\n\n\t%s/admin/machines\n\n", *server)
				case ipn.Starting, ipn.Running:
					// Done full authentication process
					cancel()
				}
			}
			if url := n.BrowseToURL; url != nil {
				fmt.Fprintf(os.Stderr, "\nTo authenticate, visit:\n\n\t%s\n\n", *url)
			}
			if p := n.Prefs; p != nil {
				prefs = *p
				saveConfig(*config, *p)
			}
		},
	}
	bc.Start(opts)
	pump(ctx, bc, c)
```

danderson (approved):
Seems fine. Is there a common scenario in which we would end up stuck in Starting for legit reasons, that this would break? Offhand I can't think, but that's the only risk introduced by this change.

rosszurowski (approved):
Seems reasonable to me! I'd echo @danderson's concerns, but also can't think of any cases where we'd be stuck in running — and if there are some, those are bugs we should probably fix.

tavianator:
Seems like this should fix https://github.com/tailscale/tailscale/issues/1889 as well

PR #2765, opened 2021-08-31
wgengine/magicsock: simplify lazy reconfig workflow

danderson:


danderson:
Suggested review: one commit at a time, each one is an independent small refactor.

wgengine/magicsock/magicsock.go
```diff
@@ -2814,59 +2867,20 @@ func (c *Conn) ParseEndpoint(endpointStr string) (conn.Endpoint, error) {
...
-		}
-	}))
-	de.updateFromNode(node)
-	c.peerMap.upsertDiscoEndpoint(de)
+	if ep, ok := c.peerMap.endpointForNodeKey(tailcfg.NodeKey(pk)); ok {
+		return ep, nil
+	}
 
-	return de, nil
+	// We should never be telling WireGuard about a new peer
```
crawshaw: Does the wireguard reconfig block until all calls to ParseEndpoint are done? If not, two rapid updates could add and remove an endpoint, with the slow following ParseEndpoint coming in after the second one and not finding the node key.

crawshaw (approved):
Seems almost too good to be true

wgengine/magicsock/magicsock.go
danderson: Yes, all ParseEndpoint happens synchronously during Wireguard reconfig. Additionally, we hold the wgengine lock during the combination of the magicsock and wireguard reconfig, so we should be guaranteed that updates always happen in lockstep: first we tell magicsock about a new peer, then we tell wireguard about the same peer.

wgengine/userspace.go
```diff
@@ -403,7 +403,6 @@ func NewUserspaceEngine(logf logger.Logf, conf Config) (_ Engine, reterr error)
 	e.logf("Starting link monitor...")
 	e.linkMon.Start()
 	e.logf("Starting magicsock...")
-	e.magicConn.Start()
 	close(e.magicConnStarted)
```
bradfitz, line 406: this whole dance is no longer necessary. You can nuke the `magicConnStarted` field.

PR #2768, opened 2021-09-01
cmd/derper: listen on host of flag server addr for port 80 and 3478

### LINKED ISSUE #2767: derp: listen on server host for port 80 and stun port 3478
When using custom derp on the server with multiple IP addresses, we would like to bind derp 80, 443 and stun 3478 to a certain IP.

```sh
$ derper --hostname your-hostname.com -a 10.240.240.10:443
2021/09/01 13:37:24 no config path specified; using /var/lib/derper/derper.key
2021/09/01 13:37:24 derper: serving on 10.240.240.10:443 with TLS
2021/09/01 13:37:24 listen tcp :80: bind: address already in use
```

Luckily derp command provides flag `-a` to customize which address to bind for port 443. But port `:80` and `:3478` are hard-coded at the moment.

https://github.com/tailscale/tailscale/blob/7a7aa8f2b0887761ef51d66634b1f3bf730c71be/cmd/derper/derper.go#L227-L231

https://github.com/tailscale/tailscale/blob/7a7aa8f2b0887761ef51d66634b1f3bf730c71be/cmd/derper/derper.go#L251-L254

It'd be great if derp command listens to all the ports on the address from flag `-a`.

chuangbo:
Fix #2767

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -117,6 +117,11 @@ func main() {
 		tsweb.DevMode = true
 	}
 
+	listenHost, _, err := net.SplitHostPort(*addr)
+	if err != nil {
+		log.Fatalf("invalid server address: %v", err)
+	}
+
 	var logPol *logpolicy.Policy
 	if *logCollection != "" {
 		logPol = logpolicy.New(*logCollection)
@@ -181,7 +186,7 @@ func main() {
 	debug.Handle("traffic", "Traffic check", http.HandlerFunc(s.ServeDebugTraffic))
 
 	if *runSTUN {
-		go serveSTUN()
+		go serveSTUN(listenHost)
 	}
 
 	httpsrv := &http.Server{
@@ -199,7 +204,6 @@ func main() {
 		WriteTimeout: 30 * time.Second,
 	}
 
-	var err error
 	if letsEncrypt {
 		if *certDir == "" {
 			log.Fatalf("missing required --certdir flag")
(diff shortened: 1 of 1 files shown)


cmd/derper/derper.go
```diff
@@ -226,7 +230,7 @@ func main() {
 		}
 		go func() {
 			port80srv := &http.Server{
-				Addr:         ":80", // the default, but to be explicit
+				Addr:         listenHost + ":80", // the default, but to be explicit
```
danderson, line 233: use net.JoinHostPort, not +, or you'll construct invalid IPv6 listen addrs.

cmd/derper/derper.go
```diff
@@ -248,8 +252,8 @@ func main() {
 	}
 }
 
-func serveSTUN() {
-	pc, err := net.ListenPacket("udp", ":3478")
+func serveSTUN(host string) {
+	pc, err := net.ListenPacket("udp", host+":3478")
```
danderson, line 256: Same here, use net.JoinHostPort.

danderson (commented):
This seems fine to me. @bradfitz thoughts?

cmd/derper/derper.go
chuangbo: Thank you for your advice.

cmd/derper/derper.go
```diff
@@ -226,7 +230,7 @@ func main() {
 		}
 		go func() {
 			port80srv := &http.Server{
-				Addr:         ":80", // the default, but to be explicit
+				Addr:         net.JoinHostPort(listenHost, "80"), // the default, but to be explicit
```
bradfitz, line 233: this comment is no longer accurate; I'd just delete it now.

bradfitz:
I suppose this is okay, but it does effectively lock you in to only one address family (IPv4 or IPv6).

I'll ask some more questions on the bug: #2767

bradfitz:
Just remove that one comment and then I think this is good.

cmd/derper/derper.go
chuangbo: Thank you for your advice.

Issue #1033, opened 2020-12-16
Magic DNS causes intermittent DNS refusals


Smittyvb:
**Describe the bug**
When Magic DNS is enabled and `tailscaled` is running, DNS requests are sometimes incorrectly refused.

**To Reproduce**
Run ``for i in `seq 1 100`; do host example.net; done | grep "REFUSED"`` with Magic DNS disabled, and notice the lack of output (good!). Enable Magic DNS, and you should see something like this (bad):
```
Host example.net not found: 5(REFUSED)
Host example.net not found: 5(REFUSED)
Host example.net not found: 5(REFUSED)
Host example.net not found: 5(REFUSED)
Host example.net not found: 5(REFUSED)
Host example.net not found: 5(REFUSED)
Host example.net not found: 5(REFUSED)
Host example.net not found: 5(REFUSED)
Host example.net not found: 5(REFUSED)
[...]
```
I have seen between 0 and 41 errors. When `tailscaled` isn't running, or Magic DNS is enabled, I never see these errors. Every time I run the command, I tend get less errors (it's quite jittery though). The error rate appears to be a per-domain thing, though. Change the test domain and your error rate will go up. Also, if I use a domain that doesn't exist (e.g. `a.example.com`) the error rate stays constant.

Here is a graph of errors for a hostname over time, that I had never resolved before:
![image](https://user-images.githubusercontent.com/10530973/102411316-bc933700-3fbf-11eb-9e90-13c90d27a5b4.png)

**Expected behavior**
DNS queries aren't refused by Magic DNS.

**Version information:**
 - Device: Desktop computer
 - OS: Pop!_OS
 - OS version: 20.10
 - Tailscale version: Reproduced on da1bad51cdc5b78b149a536b0a4300a99cfa88f5 (currently most recent commit)

Smittyvb:
Oh, I figured this out! My list of DNS resolvers for Magic DNS had two entries: `8.8.8.8`, and my educational institution's internal resolver, which is flaky (or maybe doesn't work at all?) when I'm not on their network. (all other port 53 traffic is blocked on their network, so I added that as a secondary resolver) So sometimes Tailscale would use that bad resolver (despite it being a secondary resolver), see failure from it, then give up.

PR #2772, opened 2021-09-01
tsdns: don't forward transient DNS errors

Smittyvb:
When a DNS server claims to be unable or unwilling to handle a request, instead of passing that refusal along to the client, just treat it as any other error trying to connect to the DNS server. This prevents DNS requests from failing based on if a server can respond with a transient error before another server is able to give an actual response. DNS requests only failing *sometimes* is really hard to find the cause of (such as in #1033).


bradfitz:
Ideally this would have some tests. I don't know DNS well enough to approve this as is. @sailorfrag, @danderson, do you?

sailorfrag:
As currently written, this will turn various errors returned by a server into timeouts. In the event that we're racing multiple servers (either within Tailscale or the OS is racing Tailscale DNS with other DNS servers), that's arguably preferable (with a stronger argument for soft errors).

Better would be to return one of the errors (the first one?) back to the client as soon as we've exhausted the resolvers we're racing against each other. That would be a much more involved change.

I think this is a good idea, but without the extra part I'd want to be conservative about which error codes we consume. Server failure seems good, I'm less sure about the others (e.g. refused seems like a policy decision and perhaps queries for that domain should never have been configured to go to that server).

Smittyvb:
I've added some tests for the rcode parsing logic and changed it to only act on `SERVFAIL` responses.

Also, GitHub seems to require approval for running GH Actions again on this PR (which is weird, because they were approved on this PR before and I'm not a first-time contributor)

sailorfrag:
OK I think this should wait until 1.16 is released before merging so we don't make extra DNS changes close to release. I think DNS error handling needs a lot more attention but this feels like it's a start.

Issue #2775, opened 2021-09-02
cmd/tailscale: make status show health errors


bradfitz:
"tailscale status" currently just shows peers (when Running).

"tailscale status" shows other things when in other states, like "Not logged in" or "Stopped", etc.

We discussed making "tailscale status" (in its default non-JSON mode) show some other metadata like health check problems at the top of the status output.

The JSON output mode remains the stable API for parsing with tools.

PR #2776, opened 2021-09-02
wgengine/magicsock, health, ipn/ipnstate: track DERP-advertised health

bradfitz:
And add health check errors to ipnstate.Status (tailscale status --json).

Updates #2746
Updates #2775

### DIFF
--- a/ipn/ipnstate/ipnstate.go
+++ b/ipn/ipnstate/ipnstate.go
@@ -38,6 +38,11 @@ type Status struct {
 	TailscaleIPs []netaddr.IP // Tailscale IP(s) assigned to this node
 	Self         *PeerStatus
 
+	// Health contains health check problems.
+	// Empty means everything is good. (or at least that no known
+	// problems are detected)
+	Health []string
+
 	// MagicDNSSuffix is the network's MagicDNS suffix for nodes
 	// in the network such as "userfoo.tailscale.net".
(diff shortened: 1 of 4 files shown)


ipn/ipnstate/ipnstate.go
```diff
@@ -38,6 +38,11 @@ type Status struct {
 	TailscaleIPs []netaddr.IP // Tailscale IP(s) assigned to this node
 	Self         *PeerStatus
 
+	// Health contains health check problems.
+	// Empty means everything is good. (or at least that no known
+	// problems are detected)
+	Health []string
```
crawshaw, line 44: Not parsable by frontends in this form.

If someday we want to report health to particular network nodes, and render it nicely in the UI next to the node, we'll want something more here. Should we pre-empt it and make a `type Health struct { Problem []string }` to add fields to?

crawshaw (approved):

ipn/ipnstate/ipnstate.go
bradfitz: for per-peer health, we can put that on the Peer struct itself, no?

bradfitz: that is, on `Peer map[key.Public]*PeerStatus`

crawshaw: sounds good

PR #2778, opened 2021-09-02
types/key: add MachinePrivate and MachinePublic.

danderson:
Suggested review order: look at types/key things first, that's the interesting API bits. Everything else is mechanical replacements of prior key types throughout the codebase to get things building and tests passing again.

Design principles I tried to follow:

 - All cryptographic key types live in types/key, so that we can curate and test all of them together for consistent interfaces and behaviors. This does add another dependency to tailcfg, but similar to netaddr, I believe cryptographic keys are a sufficiently core type for Tailscale that they deserve first-class use in our wire protocols.
 - There is one type per key _purpose_, not one type per mathematical type of key. IOW, it's Machine{Public,Private}, Disco{Public,Private}, Wireguard{Public,Private} and not Curve25519{Public,Private} or RSA{Public,Private}. This makes the type system help us not confuse keys, which is lovely in the bits of our code that juggle 2-3 different kinds of keys in close proximity. It also makes it harder to reuse a single key across multiple protocols, which is a smell in crypto design. This PR only introduces machine keys, so we can focus on the general API design and goals, future keys will happen with similar mechanical changes.
 - The types are opaque value types. You can't get at the key bytes directly, but you can compare keys for equality. (this actually makes these types nicer to use in code than the prior types, which required a bunch of bytes.Equal instead of ==)
 - Entropy is always available, and the code panics if crypto/rand can't get bytes. This means constructing keypairs never fails, calling code gets cleaner.
 - Zero keys panic if used for any "cryptographically meaningful" operation (e.g. deriving a public key, attempting to encrypt with it). Attempting to use an all-zero key is sufficiently catastrophic that I want to blow up hard. Some of our prior types did this, but not all and not consistently.
 - Each key type exposes exactly the crypto primitives that it expects to be used with, and no other. You can't execute X25519 with a key intended for NaCl, or vice versa.
 - Key serializations include a type prefix to identify the kind of key the text is intended to represent, and parsers verify that the serialized type matches. In some places we use "legacy" untyped serializations (e.g. the /key endpoint in control), so serialization functions also exist to/from untyped hex strings, but are only added as the current code requires them, and are marked deprecated in godoc to discourage further use.
 - Aside from mechanical function renames and removal of error handling, the new type is a drop-in replacement for the old types. That means they serialize the same (important since these keys are serialized both in control and clients), methods behave the same, and API purity is compromised using deprecated methods where necessary to ensure 1:1 compatibility.


types/key/util.go
```diff
@@ -0,0 +1,111 @@
...
+// clamping requirement.
+//
+// PLEASE NOTE that not all Curve25519 curve values require
+// clamping. When implementing a new key type that uses Curve25519,
+// you must evaluate whether that particular key's use requires
+// clamping. Here are some existing uses and whether you should clamp
+// private keys at creation.
+//
+//  - NaCl box: yes, clamp at creation.
+//  - WireGuard: yes, clamp at creation.
```
crawshaw, line 46: I'm not sure this is true. I believe wireguard will accept any key and clamp it before use. The idea is any random bytes can be used as a wireguard key.

types/persist/persist.go
```diff
@@ -44,15 +45,18 @@ func (p *Persist) Equals(p2 *Persist) bool {
 		return false
 	}
 
-	return p.LegacyFrontendPrivateMachineKey.Equal(p2.LegacyFrontendPrivateMachineKey) &&
+	return p.LegacyFrontendPrivateMachineKey == p2.LegacyFrontendPrivateMachineKey &&
```
crawshaw, line 48: not concerned about constant-time?

I don't think it matters here, but there's an argument to be made for making the key type incomparable so that non-constant-time operations don't slip into the wrong place. May not be worth the hassle of not being able to use maps on the public key though, maybe just the private ones?

crawshaw (commented):
Generally looks good. Any changes to the way keys, esp. short keys are printed in logs? I have found it hard to touch the key types in the past without accidentally changing log lines.

types/key/util.go
danderson: Hm, you're right. The thing that confused me was that we imported the clamping from wireguard-go during some refactor, because at one point we were talking to wireguard-go with our own Go structs, and so had to safely initialize our own keys.

However, now, wireguard-go only ever ingests private keys through uapi, which clamps private keys during the decoding from hex.

Updated the docstring, and will adjust our key creation when I get to that key type.

types/persist/persist.go
danderson: Hmm, interesting.

My first impulse was that this is not an issue, because private keys being what they are, we never transmit them over the wire. Thus, this kind of comparison would only take place on the machine that already has access to both private keys, so I don't think there's an exploitable oracle here.

However, the prior code used constant time comparison, so this is strictly a regression, and I don't want semantic changes at this stage of the transition. Plus, the cost of me being wrong about that oracle thing would be very high.

I suppose the next question is: should we apply the same treatment to public keys? We compare those copiously throughout the client and server, and my stance is that they're public, non-sensitive data that it's fine to compare like strings. Could someone use that to map out the public keys in use, through a timing oracle? Does that buy anyone useful information? I need to think/research about this more. I would really like to preserve the map keyability of public keys, because we use that fairly heavily (perhaps that's the actual bug...)

danderson: Switched MachinePrivate to be incomparable with a constant-time Equal.

Still thinking about the threats of a public key comparison oracle, won't submit until I've convinced myself either way and documented an answer.

danderson (commented):
I took care to not change the key serialization formats, either debug strings or on-disk storage. Both should be identical to the prior code.

crawshaw (approved):
I don't think we need to worry about public key comparisons.

PR #2783, opened 2021-09-02
cmd/derper: refactor STUN path for testing, add serverSTUN benchmark

bradfitz:
Real goal is to eliminate some allocs in the STUN path, but that requires
work in the standard library.

For now this just adds a benchmark.

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -49,6 +49,26 @@ var (
 	verifyClients = flag.Bool("verify-clients", false, "verify clients to this DERP server through a local tailscaled instance.")
 )
 
+var (
+	stats           = new(metrics.Set)
+	stunDisposition = &metrics.LabelMap{Label: "disposition"}
+	stunAddrFamily  = &metrics.LabelMap{Label: "family"}
+
+	stunReadError  = stunDisposition.Get("read_error")
+	stunNotSTUN    = stunDisposition.Get("not_stun")
+	stunWriteError = stunDisposition.Get("write_error")
+	stunSuccess    = stunDisposition.Get("success")
+
+	stunIPv4 = stunAddrFamily.Get("ipv4")
+	stunIPv6 = stunAddrFamily.Get("ipv6")
+)
+
+func init() {
+	stats.Set("counter_requests", stunDisposition)
+	stats.Set("counter_addrfamily", stunAddrFamily)
+	expvar.Publish("stun", stats)
+}
+
 type config struct {
 	PrivateKey wgkey.Private
 }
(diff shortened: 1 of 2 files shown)


cmd/derper/derper.go
```diff
@@ -254,39 +274,27 @@ func serveSTUN() {
...
+		err error
 	)
-	stats.Set("counter_requests", stunDisposition)
-	stats.Set("counter_addrfamily", stunAddrFamily)
-	expvar.Publish("stun", stats)
-
-	var buf [64 << 10]byte
 	for {
-		n, addr, err := pc.ReadFrom(buf[:])
+		n, ua, err = pc.ReadFromUDP(buf[:])
```
bradfitz, line 288: apparently this still escapes, probably from the pc.WriteTo on line 314. bah.

More reason to get netaddr in std, to make UDP more efficient in Go.

dsnet: Fascinating: https://blog.filippo.io/efficient-go-apis-with-the-inliner/
Found this while reading the code to see why `UDPConn.ReadFromUDP` didn't automatically allocate a `UDPAddr`.

dsnet: Lame. So `UDPConn.WriteToUDP` only escapes the input `addr` because it's returned in the error message. I tried doing the same function outlining trick, but can't seem to get the outer function to be inlineable since calling a function to do the error handling exceeds the inline budget.

dsnet (approved):
It's unfortunate that we can't get rid of the `ua` allocation.

What are your thoughts on copying `addr` in `UDPConn.WriteToUDP` in the error case so that the input `addr` does not escape? This will slow down the error path, but improve the common case.

bradfitz:
@dsnet, you want to take this over if you see a way to make it work? I mailed it mostly for discussion.

dsnet:
Yea, I can look into derper performance in general.

cmd/derper/derper.go
josharian: How about constructing a new, identical *net.UDPAddr to return in the error case? Something like:

```go
	if !ok {
                addrClone := *a
		return 0, &OpError{Op: "write", Net: c.fd.net, Source: c.fd.laddr, Addr: &addrClone, Err: syscall.EINVAL}
	}
```

I think then it'll only allocate on the actual error path.

If that doesn't work, there may be other things we can try. Ping me and we can pair on this. :)

josharian: Oh, I see that you suggested that same thing lower down. :) Anyway, I think that that is a good idea. I assume you're on it, but let me know if you'd like my help.

Issue #2784, opened 2021-09-02
metrics: currentFDs code is duplicated in 3 places, inefficient


bradfitz:
The `func currentFDs() int` was duplicated 3 times (once in tailscale, twice in corp) and it's inefficient, accounting for 50% of our allocations in one of our servers.

Unify, make efficient.

bradfitz:
Works. When I started monitoring FD count the other day, notice it jump from 10 kB/s to 25 kB/s...

<img width="722" alt="Screen Shot 2021-09-02 at 8 56 34 PM" src="https://user-images.githubusercontent.com/2621/131948403-12f11fea-a9f8-4783-b457-9fe8ffd882cd.png">

Then deploying this fix, back down:

<img width="709" alt="Screen Shot 2021-09-02 at 8 55 52 PM" src="https://user-images.githubusercontent.com/2621/131948466-61685e5e-f104-4f6f-839b-b351101f1b48.png">

PR #2785, opened 2021-09-02
metrics: optimize CurrentFDs to not allocate on Linux

bradfitz:
It was 50% of our allocs on one of our servers. (!!)

Updates #2784


bradfitz:
sorry

metrics/fds_linux.go
```diff
@@ -4,10 +4,109 @@
...
+	nbuf := 0                      // end valid data in buf
+	dirent := &syscall.Dirent{}
+	var err error
+	for {
+		if bufp >= nbuf {
+			bufp = 0
+			nbuf, err = readDirent(fd, buf)
+			if err != nil {
+				log.Printf("currentFDs: readDirent: %v", err)
+				return 0
```
dsnet, line 41: In the event of an error, would returning the current count be better than just reporting zero?

metrics/fds_linux.go
```diff
@@ -4,10 +4,109 @@
...
+		limit = nameBufLen
+	}
+	nameLen := bytes.IndexByte(nameBuf[:limit], 0)
+	if nameLen < 0 {
+		panic("failed to find terminating 0 byte in dirent")
+	}
+	return uint64(nameLen)
+}
+
+func parseDirEnt(dirent *syscall.Dirent, buf []byte) (consumed int, name []byte) {
```
crawshaw, line 71: Can this stuff go upstream in x/sys?

dsnet (approved):
Gross.

metrics/fds_linux.go
bradfitz: I think we're more likely to notice if the graph drops to 0

metrics/fds_linux.go
```diff
@@ -4,10 +4,109 @@
...
+	return
+}
+
+// According to https://golang.org/doc/go1.14#runtime
+// A consequence of the implementation of preemption is that on Unix systems, including Linux and macOS
+// systems, programs built with Go 1.14 will receive more signals than programs built with earlier releases.
+//
+// This causes syscall.Open and syscall.ReadDirent sometimes fail with EINTR errors.
+// We need to retry in this case.
+func open(path string, mode int, perm uint32) (fd int, err error) {
```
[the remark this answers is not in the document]
bradfitz: whoops, fixed.

metrics/fds_linux.go
bradfitz: It is, but it allocs a bunch there.

metrics/fds_linux.go
```diff
@@ -4,10 +4,109 @@
...
+		if len(name) == 0 || string(name) == "." || string(name) == ".." {
+			continue
+		}
+		count++
+	}
+}
+
+func direntNamlen(dirent *syscall.Dirent) uint64 {
+	const fixedHdr = uint16(unsafe.Offsetof(syscall.Dirent{}.Name))
+	nameBuf := (*[unsafe.Sizeof(dirent.Name)]byte)(unsafe.Pointer(&dirent.Name[0]))
```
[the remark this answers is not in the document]
bradfitz: This is code I had from elsewhere that I adapted. Simplified.

metrics/fds_linux.go
dsnet: Fair argument.

bradfitz:
PTAL

metrics/fds_linux.go
crawshaw: I mean, can you upstream a new variant that doesn't alloc?

If we can't justify it upstream, can we justify carrying it here? Are these allocs important?

bradfitz: Started the process: https://github.com/golang/go/issues/48161

trunkd numbers says it's important

metrics/fds_linux.go
```diff
@@ -4,10 +4,106 @@
...
+		consumed, name := parseDirEnt(dirent, buf[bufp:nbuf])
+		bufp += consumed
+		if len(name) == 0 || string(name) == "." || string(name) == ".." {
+			continue
+		}
+		count++
+	}
+}
+
+func direntNamlen(dirent *syscall.Dirent) uint64 {
```
crawshaw, line 49: Weird for a len of something held in memory to be a uint64 and not an int.

metrics/fds_linux.go
```diff
@@ -4,10 +4,106 @@
...
+		if dirent.Name[i] == 0 {
+			return uint64(i)
+		}
+	}
+	panic("failed to find terminating 0 byte in dirent")
+}
+
+func parseDirEnt(dirent *syscall.Dirent, buf []byte) (consumed int, name []byte) {
+	// golang.org/issue/37269
+	copy((*[unsafe.Sizeof(syscall.Dirent{})]byte)(unsafe.Pointer(dirent))[:], buf)
```
crawshaw, line 66: unsafe.Slice and below?

crawshaw (approved):

metrics/fds_linux.go
bradfitz: it's actually longer, but sure:

```
        copy((*[unsafe.Sizeof(syscall.Dirent{})]byte)(unsafe.Pointer(dirent))[:], buf)
=>
        copy(unsafe.Slice((*byte)(unsafe.Pointer(dirent)), unsafe.Sizeof(syscall.Dirent)), buf)
```

metrics/fds_linux.go
bradfitz: done

metrics/fds_linux.go
bradfitz: done

evanphx:
Not to arm chair review a persons work, though I do have a question: Looking at unix.ParseDirent, could you have preallocated a large slice (`make([]string, 0, 10000)`) and passed that in as the `names` argument to unix.ParseDirent to prevent it from doing any allocations? Since you only care about the count variable, you don't even care about goroutines clobbering the contents of your names slice.

dsnet:
> Not to arm chair review a persons work, though I do have a question: Looking at unix.ParseDirent, could you have preallocated a large slice (make([]string, 0, 10000)) and passed that in as the names argument to unix.ParseDirent to prevent it from doing any allocations? Since you only care about the count variable, you don't even care about goroutines clobbering the contents of your names slice.

Two thoughts:
* The `os` package lacks API to read into an existing `[]os.DirEntry`.
* I'm fairly certain the Go race detector won't be happen about concurrent mutations of the same slice, even if we don't care about the results. See https://github.com/golang/go/commit/eb43ce2d7711ad963de4860b70495a7aba3271c5 for a similar situation.

bradfitz:
@evanphx, feel free to try it and see!

evanphx:
@bradfitz: bitmoji

PR #2786, opened 2021-09-02
all: bump minimum Go to 1.17

bradfitz:
In prep for using 1.17 features.

Note the go.mod changes are due to:
https://golang.org/doc/go1.17#go-command

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module tailscale.com
 
-go 1.16
+go 1.17
 
 require (
 	github.com/alexbrainman/sspi v0.0.0-20210105120005-909beea2cc74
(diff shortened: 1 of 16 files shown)


go.mod
```diff
@@ -52,3 +52,148 @@ require (
 	inet.af/peercred v0.0.0-20210318190834-4259e17bb763
 	inet.af/wf v0.0.0-20210516214145-a5343001b756
 )
+
+require (
```
crawshaw, line 56: 😱

crawshaw (approved):

go.mod
bradfitz: it was already in go.sum. Now it's just also here. At least it's in its own block.

dsnet: I was under the impression that Go module pruning was to reduce the amount of junk into the `go.sum` file, so I'm a bit surprised that I don't see any changes to `go.sum`.

Per https://go.googlesource.com/proposal/+/master/design/36460-lazy-module-loading.md#effect-on-size:
> Under this approach, the set of modules recorded in the **go.mod file would in most cases increase** beyond the set recorded in Go 1.14. However, the set of modules recorded in the **go.sum file would decrease**: irrelevant modules would no longer be included.

dsnet (approved):

go.mod
crawshaw: What's scary is I kept being told all these dependencies in go.sum don't matter, we don't really use them. Now they're here, so I guess we do use them? protos? snappy? terminal color output?

crawshaw: To answer my own question:

```
go mod why -m github.com/golang/protobuf
# github.com/golang/protobuf
tailscale.com/tstest/integration/vms
tailscale.com/tstest/integration/vms.test
github.com/google/goexpect
google.golang.org/grpc/status
google.golang.org/genproto/googleapis/rpc/status
github.com/golang/protobuf/proto
```

This is the second thing the goexpect module has done to make me sad in two days, I think I'll go fork it.

dsnet: Wow. `goexpect` depends on the entire `grpc` module [just for status code like things](https://github.com/google/goexpect/blob/master/expect.go)? That's overkill...

For the `cmp` module, I felt strongly that it have a very slim set of dependencies, despite many people in Google (understandably) asking that it have first-class support for protobufs.

PR #2788, opened 2021-09-02
tstest/integration/vms: use fork of goexpect to avoid proto/grpc dep

crawshaw:
Use a fork of goexpect to avoid unwanted dependencies.


dsnet:
Does this mean that we're settling on `goexpect`? I thought we were also considering `quicktest` and other things?

dsnet (approved):

crawshaw:
I fell into the same trap of assuming it was an assertion library, until I looked at it. :)

crawshaw:
CI failure is unrelated bug.

PR #2789, opened 2021-09-02
net/dns: teach OpenBSD's manager to talk to resolvd(8).

qbit:
OpenBSD 6.9 and up has a daemon which handles nameserver configuration. This MR
teaches the OpenBSD dns manager to check if resolvd is being used. If it is, it
will use the route(8) command to tell resolvd to add the Tailscale dns entries
to resolv.conf

Signed-off-by: Aaron Bieber <aaron@bolddaemon.com>

### DIFF
--- a/net/dns/manager_openbsd.go
+++ b/net/dns/manager_openbsd.go
@@ -31,7 +31,7 @@ func NewOSConfigurator(logf logger.Logf, interfaceName string) (OSConfigurator, 
 // newOSConfigEnv are the funcs newOSConfigurator needs, pulled out for testing.
 type newOSConfigEnv struct {
 	fs          wholeFileFS
-	rcIsResolvd func(resolvConfContents []byte) string
+	rcIsResolvd func(resolvConfContents []byte) bool
 }
 
 func newOSConfigurator(logf logger.Logf, interfaceName string, env newOSConfigEnv) (ret OSConfigurator, err error) {
@@ -70,5 +70,5 @@ func rcIsResolvd(resolvConfContents []byte) bool {
 	if bytes.Contains(resolvConfContents, []byte("# resolvd:")) {
 		return true
 	}
-	return fase
+	return false
 }
--- a/net/dns/manager_openbsd.go
+++ b/net/dns/manager_openbsd.go
@@ -61,7 +61,7 @@ func newOSConfigurator(logf logger.Logf, interfaceName string, env newOSConfigEn
 	}
 
 	dbg("resolvd", "missing")
-	return newDirectManager(), nil
+	return newDirectManager(logf), nil
 }
 
 func rcIsResolvd(resolvConfContents []byte) bool {
--- a/net/dns/manager_openbsd.go
+++ b/net/dns/manager_openbsd.go
@@ -4,8 +4,72 @@
 
 package dns
 
-import "tailscale.com/types/logger"
+import (
+	"bytes"
+	"fmt"
+	"os"
 
-func NewOSConfigurator(logf logger.Logf, _ string) (OSConfigurator, error) {
-	return newDirectManager(logf), nil
+	"tailscale.com/types/logger"
+)
+
+type kv struct {
+	k, v string
+}
+
+func (kv kv) String() string {
+	return fmt.Sprintf("%s=%s", kv.k, kv.v)
+}
+
+func NewOSConfigurator(logf logger.Logf, interfaceName string) (OSConfigurator, error) {
+	return newOSConfigurator(logf, interfaceName,
+		newOSConfigEnv{
+			rcIsResolvd: rcIsResolvd,
+			fs:          directFS{},
+		})
+}
+
+// newOSConfigEnv
+type newOSConfigEnv struct {
+	fs          wholeFileFS
+	rcIsResolvd func(resolvConfContents []byte) string
+}
+
+func newOSConfigurator(logf logger.Logf, interfaceName string, env newOSConfigEnv) (ret OSConfigurator, err error) {
+	var debug []kv
+	dbg := func(k, v string) {
+		debug = append(debug, kv{k, v})
+	}
+	defer func() {
+		if ret != nil {
+			dbg("ret", fmt.Sprintf("%T", ret))
+		}
+		logf("dns: %v", debug)
+	}()
+
+	bs, err := env.fs.ReadFile(resolvConf)
+	if os.IsNotExist(err) {
+		dbg("rc", "missing")
+		return newDirectManager(logf), nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("reading /etc/resolv.conf: %w", err)
+	}
+
+	switch env.rcIsResolvd(bs) {
+	case "resolvd":
+		dbg("resolvd", "yes")
+		return newResolvdManager(logf, interfaceName)
+	default:
+		dbg("resolvd", "missing")
+		return newDirectManager(logf), nil
+	}
+}
+
+func rcIsResolvd(resolvConfContents []byte) string {
+	// If we have the string "# resolvd:" in resolv.conf resolvd(8) is
+	// managing things.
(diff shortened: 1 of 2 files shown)


bradfitz:
Thanks! Assigning to @danderson for review.

qbit:
After putting this together I realized that OpenBSD 6.9's `route` doesn't have the `nameserver` command.  So this will break 6.9 machines. I'll have to think about how to keep compat with 6.9.

net/dns/manager_openbsd.go
```diff
@@ -4,8 +4,72 @@
...
+
+func NewOSConfigurator(logf logger.Logf, interfaceName string) (OSConfigurator, error) {
+	return newOSConfigurator(logf, interfaceName,
+		newOSConfigEnv{
+			rcIsResolvd: rcIsResolvd,
+			fs:          directFS{},
+		})
+}
+
+// newOSConfigEnv
```
danderson, line 31: // newOSConfigEnv are the funcs newOSConfigurator needs, pulled out for testing.

net/dns/manager_openbsd.go
```diff
@@ -4,8 +4,72 @@
...
+	case "resolvd":
+		dbg("resolvd", "yes")
+		return newResolvdManager(logf, interfaceName)
+	default:
+		dbg("resolvd", "missing")
+		return newDirectManager(), nil
+	}
+}
+
+func rcIsResolvd(resolvConfContents []byte) string {
```
danderson, line 68: I'd make this func return a bool if it's making a binary decision. I believe you patterned this on linux's resolvOwner, which has to return an N-way decision between all the different managers. OpenBSD is mercifully simpler :)

net/dns/resolvd.go
```diff
@@ -0,0 +1,53 @@
...
+
+	for _, s := range config.Nameservers {
+		args = append(args, s.String())
+	}
+
+	cmd := exec.Command("/sbin/route", args...)
+	return cmd.Run()
+}
+func (m *resolvdManager) SupportsSplitDNS() bool {
+	return true
```
danderson, line 44: I don't think this is correct, based on the implementation of SetDNS. Split DNS support here means that the overall manager can tell the OS to only use a set of resolvers for specific domain suffixes. The implementation of SetDNS doesn't make use of MatchDomains, so it wouldn't implement the configuration that the DNS manager wanted.

Can resolvd do per-suffix DNS routes? If so, this can be true, but SetDNS needs to obey MatchDomains in what it configures.

net/dns/resolvd.go
```diff
@@ -0,0 +1,53 @@
...
+	args := []string{
+		"nameserver",
+		m.ifName,
+	}
+
+	for _, s := range config.Nameservers {
+		args = append(args, s.String())
+	}
+
+	cmd := exec.Command("/sbin/route", args...)
```
danderson, line 40: Is this constructed command safe to invoke repeatedly, with either the same nameservers or a different set? I assume so, but want to explicitly confirm since I don't have an OpenBSD box to test on.

net/dns/resolvd.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// resolvdManager is an OSConfigurator which uses route(1) to teach OpenBSD's
+// resolvd(8) about DNS servers.
+type resolvdManager struct {
+	logf   logger.Logf
+	ifName string
+}
+
+func (m *resolvdManager) SetDNS(config OSConfig) error {
```
danderson, line 30: SetDNS also needs to propagate config.SearchDomains into the OS, for single-label name expansion to work right.

net/dns/resolvd.go
```diff
@@ -0,0 +1,53 @@
...
+
+	cmd := exec.Command("/sbin/route", args...)
+	return cmd.Run()
+}
+func (m *resolvdManager) SupportsSplitDNS() bool {
+	return true
+}
+
+func (m *resolvdManager) GetBaseConfig() (OSConfig, error) {
+	return OSConfig{}, ErrGetBaseConfigNotSupported
```
danderson, line 48: If resolvd doesn't support per-suffix DNS routing, this needs to be able to read the base OS configuration, so that we can emulate split DNS within tailscale. Not necessary if SupportsSplitDNS() is true and SetDNS supports it natively though.

net/dns/resolvd.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (m *resolvdManager) Close() error {
```
danderson, line 51: Do you need to clear any settings here? Or does resolvd clean up when the interface vanishes?

danderson (commented):
Thanks for the PR! Main feedback is that this OSConfigurator doesn't implement our definition of split DNS, so either it needs to do that (if resolvd supports it), or this has to be able to read + entirely replace the prior OS configuration so that we can emulate split DNS within tailscaled.

The manpages for resolvd and unwind are a bit light, but afaict resolvd supports neither overriding all other DNS configuration, nor per-suffix DNS routing. As such, I'm worried that resolvd cannot implement either of the two behaviors we require to make DNS work correctly :/ I hope I'm wrong and the manpages are just missing the bits I'm looking for.

net/dns/resolvd.go
qbit: `resolvd` handles cleanup when the interface is removed.

net/dns/resolvd.go
qbit: Yep, it adds/removes accordingly.

```
qbit@tal[0]:~$ cat /etc/resolv.conf                                                                                                                                                                         
nameserver 100.100.100.100 # resolvd: tun0
nameserver 10.6.0.2 # resolvd: em0
search qbit.github.beta.tailscale.net
qbit@tal[0]:~$ doas route nameserver tun0 100.100.100.100 
qbit@tal[0]:~$ cat /etc/resolv.conf                       
nameserver 100.100.100.100 # resolvd: tun0
nameserver 10.6.0.2 # resolvd: em0
search qbit.github.beta.tailscale.net
qbit@tal[0]:~$ doas route nameserver tun0 100.100.100.100 1.1.1.1 
qbit@tal[0]:~$ cat /etc/resolv.conf                               
nameserver 1.1.1.1 # resolvd: tun0
nameserver 100.100.100.100 # resolvd: tun0
nameserver 10.6.0.2 # resolvd: em0
search qbit.github.beta.tailscale.net
qbit@tal[0]:~$ doas route nameserver tun0 100.100.100.100         
qbit@tal[0]:~$ cat /etc/resolv.conf                       
nameserver 100.100.100.100 # resolvd: tun0
nameserver 10.6.0.2 # resolvd: em0
search qbit.github.beta.tailscale.net
qbit@tal[0]:~$ 
```

net/dns/resolvd.go
qbit: I don't think resolvd can do per-suffix DNS routes. I am not completely sure I understand DNSR, but it seems that resolved's approach of ordering nameservers based off route priority _almost_ achieves the same thing (but maybe that's just my lack of understanding :P).

I guess making this `false` is the correct path forward.

net/dns/resolvd.go
qbit: Ok, updated to propagate SearchDomains.

net/dns/resolvd.go
qbit: GetBaseConfig now returns a partially populated OSConfig. I only had it grab the SearchDomains since resolvd handles the rest.

qbit:
OK, I think I resolved everything (possibly introduced some new things too :P)!

qbit:
I believe I have addressed all the issues for this PR. I have a squashed version ready to push but I wanted to leave the history for anyone wanting to review again.

Should I just push the squash'd version?

bradfitz:
We're all back from vacation now. Leaving for @danderson to review again. I pushed the button to run the CI.

qbit:
Awesome, hope it was a good vacation! Thanks for pushin the button!

PR #2792, opened 2021-09-03
types/key: add a special key with custom serialization for control private keys

danderson:


types/key/control_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+package key
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+func TestControlKey(t *testing.T) {
+	serialized := `{"PrivateKey":[36,132,249,6,73,141,249,49,9,96,49,60,240,217,253,57,3,69,248,64,178,62,121,73,121,88,115,218,130,145,68,254]}`
```
bradfitz, line 13: hah, what an encoding!

/me checks that you didn't actually submit our actual prod private key to source 😛

bradfitz (approved):

types/key/control_test.go
danderson: I didn't, don't panic :) Generated a test key with NewControl(), serialized it, manually verified the serialization matched the preexisting control plane serialization, then hardcoded the test key.

danderson:
For the record: also reproduced the original deserialization failure in a local test control, and verified that this fixes it.

Issue #2619, opened 2021-08-10
TLS handshake issue when running a custom DERP server


hronro:
![screenshot](https://user-images.githubusercontent.com/15059605/128657304-0ed14483-f36a-4848-9d11-e66f15e630e0.png)

Any ideas about this?

DentonGentry:
The level of support we provide in operating a DERP server is fairly limited.
As the client appears to be disconnecting, I would check that the certificate generated for the DERP server is correct.

If that isn't it, I would suggest running wireshark on either the DERP server or client to see what is happening.

bradfitz:
Are those log lines from your own device, or from random scanning from being on the internet?

That is, is it working, ignoring the log spam?

hronro:
@DentonGentry 
Thank you for your answer. I would like to know what's the location of the certificate? And does the certificate generated by derper automatically or I need to create the certificate manually by using a 3rd-party service like Let's encrypt?

bradfitz:
@hronro, see the `--help`:

```
$ derper --help
Usage of derper:
  -a string
        server address (default ":443")
  -bootstrap-dns-names string
        optional comma-separated list of hostnames to make available at /bootstrap-dns
  -c string
        config file path
  -certdir string
        directory to store LetsEncrypt certs, if addr's port is :443 (default "/home/bradfitz/.cache/tailscale/derper-certs")
  -dev
        run in localhost development mode
  -hostname string
        LetsEncrypt host name, if addr's port is :443 (default "derp.tailscale.com")
  -logcollection string
        If non-empty, logtail collection to log to
  -mesh-psk-file string
        if non-empty, path to file containing the mesh pre-shared key file. It should contain some hex string; whitespace is trimmed. (default "/home/bradfitz/keys/derp-mesh.key")
  -mesh-with string
        optional comma-separated list of hostnames to mesh with; the server's own hostname can be in the list
  -stun
        also run a STUN server
  -verify-clients
        verify clients to this DERP server through a local tailscaled instance.
```

It does LetsEncrypt for you automatically if you use the default `-a :443` listen address. You'll find the cert in `$HOME/.cache/tailscale/derper-certs`.

Is Tailscale connecting to it and working?

Do you have some other server or party in front of `derper`?

hronro:
@bradfitz 
Thanks for your answer.
After some investigation, I found sign/resign certs on LetsEncrypt is somehow blocked by China gov.
[...]

PR #2793, opened 2021-09-03
cmd/derper: support customized server certificate

SilverBut:
In #2619, @hronro asked if users could provide an certificate by themself, instead of using the automatically retrieved one via Let's Encrypt.

This pull request adds a new "-certmode" argument for control certificate source. Default argument "letsencrypt" would keep the old behaviour. 

In "manual" mode, certs are searched in given "-certdir" argument. The filename being searched is decided by HTTPS request's SNI field, and in format `${certdir}/${hostname}.{crt/key}`. If cert can not be loaded on start, error is thrown.

Solves #2619 and #2794

Signed-off-by: SilverBut <SilverBut@users.noreply.github.com>

### DIFF
--- a/cmd/derper/cert.go
+++ b/cmd/derper/cert.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
@@ -6,6 +6,7 @@ package main
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"net/http"
@@ -24,7 +25,7 @@ type certProvider interface {
 	HTTPHandler(fallback http.Handler) http.Handler
 }
 
-func certProviderByCertMode(mode, dir string) (certProvider, error) {
+func certProviderByCertMode(mode, dir, hostname string) (certProvider, error) {
 	if dir == "" {
 		return nil, errors.New("missing required --certdir flag")
 	}
@@ -32,25 +33,44 @@ func certProviderByCertMode(mode, dir string) (certProvider, error) {
 	case "letsencrypt":
 		certManager := &autocert.Manager{
 			Prompt:     autocert.AcceptTOS,
-			HostPolicy: autocert.HostWhitelist(*hostname),
+			HostPolicy: autocert.HostWhitelist(hostname),
 			Cache:      autocert.DirCache(dir),
 		}
-		if *hostname == "derp.tailscale.com" {
+		if hostname == "derp.tailscale.com" {
 			certManager.HostPolicy = prodAutocertHostPolicy
 			certManager.Email = "security@tailscale.com"
 		}
 		return certManager, nil
 	case "manual":
-		return &manualCertManager{
-			certdir: dir,
-		}, nil
+		return NewManualCertManager(dir, hostname)
 	default:
 		return nil, fmt.Errorf("unsupport cert mode: %q", mode)
 	}
 }
 
 type manualCertManager struct {
-	certdir string
+	cert     *tls.Certificate
+	hostname string
+}
+
+// NewManualCertManager returns a cert provider which read certificate by given hostname on create.
+func NewManualCertManager(certdir, hostname string) (certProvider, error) {
+	keyname := unsafeHostnameCharacters.ReplaceAllString(hostname, "")
+	crtPath := filepath.Join(certdir, keyname+".crt")
+	keyPath := filepath.Join(certdir, keyname+".key")
+	cert, err := tls.LoadX509KeyPair(crtPath, keyPath)
+	if err != nil {
+		return nil, fmt.Errorf("can not load x509 key pair for hostname %q: %w", keyname, err)
+	}
+	// ensure hostname matches with the certificate
+	x509Cert, err := x509.ParseCertificate(cert.Certificate[0])
+	if err != nil {
+		return nil, fmt.Errorf("can not load cert: %w", err)
+	}
+	if x509Cert.VerifyHostname(hostname) != nil {
+		return nil, errors.New("refuse to load cert: hostname mismatch with key")
+	}
+	return &manualCertManager{cert: &cert, hostname: hostname}, nil
 }
 
 func (m *manualCertManager) TLSConfig() *tls.Config {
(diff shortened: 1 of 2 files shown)


SilverBut:
Hi @DentonGentry. I'm not sure this is the right way to add support for customized cert you mentioned in #2619, since this is not well discussed yet. Please feel free to edit if necessary ;)

cmd/derper/cert.go
```diff
@@ -0,0 +1,96 @@
...
+
+func isAvailableCertMode(mode string) bool {
+	switch mode {
+	case "letsencrypt", "manual":
+		return true
+	}
+	return false
+}
+
+func certProviderByCertMode(mode string, dir string) certProvider {
```
bradfitz, line 35: this can be written `(mode, dir string)`

cmd/derper/cert.go
```diff
@@ -0,0 +1,96 @@
...
+	if dir == "" {
+		log.Fatalf("missing cert directory")
+	}
+	switch mode {
+	case "letsencrypt":
+		certManager := &autocert.Manager{
+			Prompt:     autocert.AcceptTOS,
+			HostPolicy: autocert.HostWhitelist(*hostname),
+			Cache:      autocert.DirCache(*certDir),
```
bradfitz, line 44: I think you meant `dir` instead of the flag `*certDir` value?

cmd/derper/cert.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	// RFC basically only allow [0-9a-zA-Z\.-] in SNI
+	// We will follow it for security purpose
+	// Also, unnecessary to check concated string
+	keyname = unsafeHostnameCharacters.ReplaceAllString(keyname, "")
+	crtPath := filepath.Join(m.certdir, keyname+".crt")
+	keyPath := filepath.Join(m.certdir, keyname+".key")
+	cert, err := tls.LoadX509KeyPair(crtPath, keyPath)
+	if err != nil {
+		return nil, fmt.Errorf("can not load x509 key pair for hostname %s: %w", keyname, err)
```
bradfitz, line 89: use %q instead of %s ... it shows empty, binary, surrounding whitespace, etc.

cmd/derper/cert.go
```diff
@@ -0,0 +1,96 @@
...
+	"path/filepath"
+	"regexp"
+
+	"golang.org/x/crypto/acme/autocert"
+)
+
+var unsafeHostnameCharacters = regexp.MustCompile(`[^a-zA-Z0-9-\.]`)
+
+type certProvider interface {
+	// TLSConfig provide TLS related stuff for HTTPS library
```
bradfitz, line 21: instead of "stuff": "TLSConfig returns the TLS configuration for the HTTP server."

(and end in a period)

cmd/derper/cert.go
```diff
@@ -0,0 +1,96 @@
...
+			Prompt:     autocert.AcceptTOS,
+			HostPolicy: autocert.HostWhitelist(*hostname),
+			Cache:      autocert.DirCache(*certDir),
+		}
+		if *hostname == "derp.tailscale.com" {
+			certManager.HostPolicy = prodAutocertHostPolicy
+			certManager.Email = "security@tailscale.com"
+		}
+		return certManager
+	case "manual":
```
bradfitz, line 51: kinda a bummer to duplicate this switch/check between the two funcs. Maybe this can return (certProvider, error) instead and then you can remove `func isAvailableCertMode`?

cmd/derper/derper.go
```diff
@@ -54,6 +54,13 @@ type config struct {
 }
 
 func loadConfig() config {
+	if !isAvailableCertMode(*certMode) {
+		log.Panicf("invalid cert mode")
```
bradfitz, line 58: log.Fatalf("invalid cert mode %q", *certMode)

(no panic)

bradfitz (commented):
Thanks!

cmd/derper/cert.go
SilverBut: Oh, my fault, didn't test under old mode...

SilverBut:
Instead of load cert for every request, now the cert is only loaded once when `derper` starts, since it only support one hostname.

cmd/derper/cert.go
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
```
bradfitz, line 1: wrong year

bradfitz: (only set it to the current year for new files, not on existing files)

cmd/derper/cert.go
```diff
@@ -0,0 +1,95 @@
...
+		return certManager, nil
+	case "manual":
+		return NewManualCertManager(dir, hostname)
+	default:
+		return nil, fmt.Errorf("unsupport cert mode: %q", mode)
+	}
+}
+
+type manualCertManager struct {
+	cert     *tls.Certificate
```
bradfitz, line 52: this does mean that you need a derper restart to rotate the cert out, but that's acceptable I suppose, considering how quickly it starts up.

cmd/derper/derper.go
```diff
@@ -39,6 +38,7 @@ var (
 	dev           = flag.Bool("dev", false, "run in localhost development mode")
 	addr          = flag.String("a", ":443", "server address")
 	configPath    = flag.String("c", "", "config file path")
+	certMode      = flag.String("certmode", "letsencrypt", "mode for getting a cert. possible options: manual, letsencrypt")
```
bradfitz, line 41: did you consider making this just be a `--letsencrypt=true` bool flag? Do you foresee other cert modes in the future?

cmd/derper/cert.go
SilverBut: fixed

cmd/derper/derper.go
SilverBut: Yes. Letsencrypt is just an ACME provider and maybe someone want to use alternate provider (like [DigiCert](https://docs.digicert.com/certificate-tools/Certificate-lifecycle-automation-index/automation-user-guide/install-automation-agent-webserver/)). Maybe one day, some one want to implement a "acmev2" mode, and allowing customize endpoint, but that's out of the scope of this PR.

cmd/derper/cert.go
SilverBut: Yes. In addition, reconnection DERP from the server is cheaper, and write a watcher for the cert file might be not necessary. I think for a busy DERP server, reduce per-connection cost is more important.
In addition, this would encourage users restart their DERP sever to catch up latest implementation.

PR #2806, opened 2021-09-07
ipn/ipnlocal: start adding some netmap to DNS config tests

bradfitz:
Follow-up to #2805


crawshaw (approved):
This is good stuff. If you want something else to think on: I'd love to have these same DNSConfig / want objects to plug into testcontrol, to run real nslookup requests on a VM in the integration tests. Anywhere obvious we could put them that's accessible to both?

PR #2814, opened 2021-09-08
all: update tests to use tstest.MemLogger

bradfitz:
And give MemLogger a mutex, as one caller had, which does match the logf
contract better.


crawshaw:
LGTM

PR #2825, opened 2021-09-09
api: document new API to authorize a device

### LINKED ISSUE #2813: FR: API to approve machine additions
When device approvals are enabled on a Tailnet, the admin must manually approve each device on the machine page.

Adding an API to approve a node would allow integration with posture checking and endpoint security services.

bradfitz:
Fixes #2813

### DIFF
--- a/api.md
+++ b/api.md
@@ -13,6 +13,8 @@ Currently based on {some authentication method}. Visit the [admin panel](https:/
   - Routes
     - [GET device routes](#device-routes-get)
     - [POST device routes](#device-routes-post)
+  - Authorize machine
+    - [POST device authorized](#device-authorized-post)
 * **[Tailnets](#tailnet)**
   - ACLs
     - [GET tailnet ACL](#tailnet-acl-get)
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -234,6 +236,33 @@ Response
...
+
+##### Example
+
+```
+curl 'https://api.tailscale.com/api/v2/device/11055/authorized' \
+-u "tskey-yourapikey123:" \
+--data-binary '{"authorized": true}'
+```
+
+The response is 2xx on success. The body is currently an empty JSON
```
bradfitz, line 263: This last part is a bit weird but I'm not sure what's normal here. Return a 204 (no content) status code? Return the Device object itself after mutation? Return just the authorization status? But that's always true on success.

For now it's just an empty JSON object.

api.md
```diff
@@ -234,6 +236,33 @@ Response
 }
 ```
 
+<a name=device-authorized-post></div>
```
DentonGentry, line 239: I see that all of the other API sections do this, but I don't understand why there is a `</div>` instead of `</a>`

DentonGentry (approved):

api.md
bradfitz: Hah, nor do I. I just copied this from another section without noticing. Will fix.

Issue #2827, opened 2021-09-09
Running tailscale on CircleCI fails with SO_BINDTODEVICE


chrisgoffinet:
I am attempting to use tailscale 1.14.0 with userspace-networking on CircleCI. It looks to be failing with `setting SO_BINDTODEVICE: operation not permitted`, I made sure to run as `root`. 

Here is the log:
```
$ sudo tailscaled -verbose 10 --tun=userspace-networking
logtail started
Program starting: v1.14.0-t5cff36945-g809e87bba, Go 1.17-tsdd6b269c77: []string{"tailscaled", "-verbose", "10", "--tun=userspace-networking"}
LogID: 7226a55e02478af0f1ea71db909049bc7042675461dba456521d594577bd3870
logpolicy: using system state directory "/var/lib/tailscale"
logpolicy.Read /var/lib/tailscale/tailscaled.log.conf: open /var/lib/tailscale/tailscaled.log.conf: no such file or directory
wgengine.NewUserspaceEngine(tun "userspace-networking") ...
using fake (no-op) tun device
using fake (no-op) OS network configurator
using fake (no-op) DNS configurator
dns: using dns.noopManager
link state: interfaces.State{defaultRoute=eth0 ifs={eth0:[192.168.96.3/20]} v4=true v6=false}
magicsock: unable to bind udp4 port 0: listen udp4 :0: setting SO_BINDTODEVICE: operation not permitted
wgengine.NewUserspaceEngine(tun "userspace-networking") error: wgengine: magicsock: initialBind IPv4 failed: failed to bind any ports (tried [0])
wgengine.New: wgengine: magicsock: initialBind IPv4 failed: failed to bind any ports (tried [0])
flushing log.
logger closing down
logtail: dial "log.tailscale.io:443" failed: dial tcp 34.210.105.16:443: setting SO_BINDTODEVICE: operation not permitted (in 1ms)
logtail: upload: log upload of 369 bytes compressed failed: Post "https://log.tailscale.io/c/tailnode.log.tailscale.io/38ca472036f521dff487f563bdbb5282e18fed63ef2620c93c820b10596b1d4b": dial tcp 34.210.105.16:443: setting SO_BINDTODEVICE: operation not permitted
logtail: backoff: 12 msec
logtail: dial "log.tailscale.io:443" failed: dial tcp 34.210.105.16:443: setting SO_BINDTODEVICE: operation not permitted (in 1ms)
[...]

Issue #2822, opened 2021-09-09
tailscaled failing to start on GKE Autopilot cluster in userspace-networking


BrianAdams:
Simply trying to run the container on [Google GKE Autopilot ](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-overview)

Turned on all the capabilities that are allowed:
```
        securityContext:
          capabilities:
            add: ["NET_BIND_SERVICE","SETPCAP","MKNOD","CHOWN","DAC_OVERRIDE","FOWNER","FSETID","KILL","SETGID","SETUID","SYS_CHROOT","SETFCAP"]      
```

GKE Autopilot does drop the NET_RAW capability after start up if that is at all relevant.

But when tailscaled runs it fails with the unable to bind udp4 port 0: ... SO_BINDTODEVICE: operation not permitted.

I am using the[ google cloud run instructions](https://tailscale.com/kb/1108/cloudrun/) to create the Docker container.

Tried with both 
`--tun=userspace-networking --socks5-server=localhost:1055`
and
`--tun=userspace-networking`

```
failed to connect to local tailscaled (which appears to be running). Got error: Get "http://local-tailscaled.sock/localapi/v0/status": dial unix /var/run/tailscale/tailscaled.sock: connect: no such file or directory
2021/09/09 05:10:25 logtail started
2021/09/09 05:10:25 Program starting: v1.14.0-t5cff36945-g809e87bba, Go 1.17-tsdd6b269c77: []string{"/app/tailscaled", "--tun=userspace-networking"}
2021/09/09 05:10:25 LogID: f045a2528d8f0acedd93da4d1d970a3f252997628654077d393bd2d4d1cc30c9
2021/09/09 05:10:25 logpolicy: using system state directory "/var/lib/tailscale"
logpolicy.Read /var/lib/tailscale/tailscaled.log.conf: open /var/lib/tailscale/tailscaled.log.conf: no such file or directory
2021/09/09 05:10:25 wgengine.NewUserspaceEngine(tun "userspace-networking") ...
2021/09/09 05:10:25 dns: using dns.noopManager
2021/09/09 05:10:25 link state: interfaces.State{defaultRoute=eth0 ifs={eth0:[10.52.0.155/26]} v4=true v6=false}
2021/09/09 05:10:25 magicsock: unable to bind udp4 port 0: listen udp4 :0: setting SO_BINDTODEVICE: operation not permitted
[...]

PR #2828, opened 2021-09-09
cmd/tailscaled: disable netns in userspace-networking mode

bradfitz:
Updates #2827
Updates #2822

### DIFF
--- a/net/netns/netns.go
+++ b/net/netns/netns.go
@@ -19,12 +19,23 @@ import (
 	"net"
 
 	"inet.af/netaddr"
+	"tailscale.com/syncs"
 )
+
+var disabled syncs.AtomicBool
+
+// Disable disables netns for the process.
+func Disable() {
+	disabled.Set(true)
+}
 
 // Listener returns a new net.Listener with its Control hook func
 // initialized as necessary to run in logical network namespace that
 // doesn't route back into Tailscale.
 func Listener() *net.ListenConfig {
+	if disabled.Get() {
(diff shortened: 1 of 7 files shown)


net/netns/netns.go
```diff
@@ -19,12 +19,23 @@ import (
 	"net"
 
 	"inet.af/netaddr"
+	"tailscale.com/syncs"
 )
 
+var disabled syncs.AtomicBool
+
+// Disable disables netns for the process.
+func Disable() {
```
DentonGentry, line 28: Recommend we log that netns is being disabled.

DentonGentry (approved):

net/netns/netns.go
bradfitz: We already log that userspace-networking is being used, though? They're 1:1 so it doesn't strictly add any information.

I don't really mind but I'm trying to be a little bit more cautious of adding lots of logs.

DentonGentry: +1

bradfitz: yeah, right before this is:

```
                logf("wgengine.NewUserspaceEngine(tun %q) ...", name)
```

Where the `name` is `"userspace-networking"`

bradfitz: Or did you want to backport this to 1.14.x?

DentonGentry: I wasn't going to recommend that, the people running into it so far would likely be fine trying an unstable release.

PR #2834, opened 2021-09-11
portlist: reduce allocs on Linux

bradfitz:
Notably, it no longer allocates proportional to the number of open
sockets on the machine. Any alloc reduction numbers are a little
contrived with such a reduction but e.g. on a machine with 50,000
connections open:

```
name          old time/op    new time/op    delta
ParsePorts-6    57.7ms ± 6%    32.8ms ± 3%   -43.04%  (p=0.000 n=9+10)

name          old alloc/op   new alloc/op   delta
ParsePorts-6    24.0MB ± 0%     0.0MB ± 0%  -100.00%  (p=0.000 n=10+9)

name          old allocs/op  new allocs/op  delta
ParsePorts-6      100k ± 0%        0k ± 0%   -99.99%  (p=0.000 n=10+10)
```

Updates tailscale/corp#2566

### DIFF
--- a/portlist/portlist_linux.go
+++ b/portlist/portlist_linux.go
@@ -17,6 +17,7 @@ import (
 	"syscall"
 	"time"
 
+	"go4.org/mem"
 	"golang.org/x/sys/unix"
 	"tailscale.com/syncs"
 )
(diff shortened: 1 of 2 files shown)


bradfitz:
This should fix the CPU (garbage) performance on the control server with its bazillion fds.

Benchmark numbers later. Computer window closing.

crawshaw (approved):

PR #2838, opened 2021-09-13
portlist: cache field index position between runs, cut two more allocs (Linux)

bradfitz:
```
name          old time/op    new time/op    delta
ParsePorts-6    6.41ms ± 7%    3.15ms ± 2%  -50.84%  (p=0.000 n=9+9)

name          old alloc/op   new alloc/op   delta
ParsePorts-6      408B ± 0%      216B ± 0%  -47.06%  (p=0.002 n=8+10)

name          old allocs/op  new allocs/op  delta
ParsePorts-6      7.00 ± 0%      4.00 ± 0%  -42.86%  (p=0.000 n=10+10)
```

Updates tailscale/corp#2566

### DIFF
--- a/portlist/portlist_linux.go
+++ b/portlist/portlist_linux.go
@@ -81,7 +81,7 @@ func parsePorts(r *bufio.Reader, fileBase string) ([]Port, error) {
 	var ret []Port
 
 	// skip header row
-	_, err := r.ReadString('\n')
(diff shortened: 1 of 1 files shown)


crawshaw (approved):
I *think* this is right, but it's starting to get hard to tell. :)

Issue #1526, opened 2021-03-17
Block looped Tailscale traffic in tstun


danderson:
We're hunting an elusive problem where sometimes mac machines seem to generate a routing loop, and so we transmit large amount of our own wireguard packets. To gather more information and mitigate the problem, we should implement a blocklist in tstun.

Magicsock has a list of all peer endpoints it's currently aware of. We should plumb that list into tstun as a blocklist, such that any attempt to send to those endpoints via Tailscale itself will drop the packets and log `[unexpected]` to help us diagnose.

bradfitz:
Maybe related:

https://forum.tailscale.com/t/switch-from-cellular-to-wifi-tailscale-still-uses-cellular-to-exchange-data/441/6

> I have been trialling the new 1.6 release.
>
> Two days ago while I was on wifi at the office all day I managed to use 6.5GB of cellular data.

maxburke:
From #2917: the user encountering it is on macOS 11.6, it's a quad-core i5 Intel Macbook Pro. No other VPN software at the moment - though he had Tunnelblick installed up until yesterday, it is now uninstalled as of the time the issue happened. His computer has not been rebooted since uninstalling Tunnelblick. No antivirus though he has Automox installed for patch monitoring.

We have MagicDNS enabled, our MagicDNS servers are on a private subnet in Azure that are accessible through the subnet router in question.

bradfitz:
> We have MagicDNS enabled, our MagicDNS servers are on a private subnet in Azure that are accessible through the subnet router in question.

That does seem to be a theme in reproducing this. Thanks.

dblohm7:
Message from a user who emailed support:

> MacBook pro was on home network ([192.168.1.0/24](http://192.168.1.0/24)), and slept/lid shut.
Moved from home network to external network.
Lid open/wake and connect to the external network.
Shortly after, and for hours after (until I noticed), massive network traffic upload from MacBook Pro to Jumpbox on home network. Possibly via DERP node in Sydney.
Traffic did not exit Jumpbox. Basically flat lined jumpbox/connection/etc. for several hours.
Traffic flood only stopped by turning Tailscale off on MacBook Pro and then back on again.

woutwoot:
I believe I'm also seeing this issue in the form of high CPU usage for IPNExtension on macOS. 
[...]

PR #2840, opened 2021-09-13
net/tstun: block looped disco traffic

bradfitz:
Updates #1526 (maybe fixes? time will tell)

### DIFF
--- a/net/tstun/wrap.go
+++ b/net/tstun/wrap.go
@@ -7,6 +7,7 @@
 package tstun
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
@@ -19,7 +20,9 @@ import (
 	"golang.zx2c4.com/wireguard/device"
 	"golang.zx2c4.com/wireguard/tun"
 	"inet.af/netaddr"
+	"tailscale.com/disco"
 	"tailscale.com/net/packet"
+	"tailscale.com/tailcfg"
 	"tailscale.com/tstime/mono"
 	"tailscale.com/types/ipproto"
 	"tailscale.com/types/logger"
@@ -76,6 +79,7 @@ type Wrapper struct {
 
 	destIPActivity atomic.Value // of map[netaddr.IP]func()
 	destMACAtomic  atomic.Value // of [6]byte
+	discoKey       atomic.Value // of tailcfg.DiscoKey
 
 	// buffer stores the oldest unconsumed packet from tdev.
 	// It is made a static buffer in order to avoid allocations.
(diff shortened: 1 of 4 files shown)


danderson (approved):

net/tstun/wrap.go
```diff
@@ -481,6 +494,16 @@ func (t *Wrapper) filterIn(buf []byte) filter.Response {
 				f(data)
 			}
 		}
+	case ipproto.UDP:
+		if pkt := p.Payload(); disco.LooksLikeDiscoWrapper(pkt) {
+			discoFrom := pkt[len(disco.Magic):][:len(tailcfg.DiscoKey{})]
+			if selfDiscoPub, ok := t.discoKey.Load().(tailcfg.DiscoKey); ok {
```
josharian, line 500: `ok && bytes.Equal(...) {`

net/tstun/wrap.go
```diff
@@ -481,6 +494,16 @@ func (t *Wrapper) filterIn(buf []byte) filter.Response {
 				f(data)
 			}
 		}
+	case ipproto.UDP:
+		if pkt := p.Payload(); disco.LooksLikeDiscoWrapper(pkt) {
+			discoFrom := pkt[len(disco.Magic):][:len(tailcfg.DiscoKey{})]
```
josharian, line 499: Can we add a package disco method to extract `discoFrom`, rather than having tstun know about the structure of disco packets?

net/tstun/wrap.go
```diff
@@ -481,6 +494,16 @@ func (t *Wrapper) filterIn(buf []byte) filter.Response {
 				f(data)
 			}
 		}
+	case ipproto.UDP:
+		if pkt := p.Payload(); disco.LooksLikeDiscoWrapper(pkt) {
```
josharian, line 498: Some inversion of control to reduce indentation would be nice here.

```go
pkt := p.Payload()
if !disco.LooksLikeDiscoWrapper(pkt) {
  break
}
// etc
```

net/tstun/wrap.go
```diff
@@ -481,6 +494,16 @@ func (t *Wrapper) filterIn(buf []byte) filter.Response {
 				f(data)
 			}
 		}
+	case ipproto.UDP:
```
josharian, line 497: Add a short comment here explaining why we're looking at UDP traffic.

net/tstun/wrap.go
```diff
@@ -481,6 +494,16 @@ func (t *Wrapper) filterIn(buf []byte) filter.Response {
...
+		if pkt := p.Payload(); disco.LooksLikeDiscoWrapper(pkt) {
+			discoFrom := pkt[len(disco.Magic):][:len(tailcfg.DiscoKey{})]
+			if selfDiscoPub, ok := t.discoKey.Load().(tailcfg.DiscoKey); ok {
+				if bytes.Equal(selfDiscoPub[:], discoFrom) {
+					t.logf("[unexpected] received self disco package over tstun; dropping")
```
josharian, line 502: To confirm, when issue 1526 reproduces, would we expect to see one of these log lines, or one per packet send (for other reasons)? I know we have a rate-limiter, but even going through the rate-limited once per packet is going to be not cheap. Not that I see any good alternatives.

Is there any other info we want as part of debugging this? (I'd suggest remote port+IP but I guess that has PII problems.)

josharian (approved):

net/tstun/wrap.go
bradfitz: I think that stopping one will stop the flood.

But we'll see. We'll have the users test it before the release and we can adjust.

net/tstun/wrap.go
bradfitz: pulled out

net/tstun/wrap.go
bradfitz: done

net/tstun/wrap.go
bradfitz: done

PR #2845, opened 2021-09-14
cmd/tailscale/web: support TLS from env vars.

DentonGentry:
pfSense stores its SSL certificate and key in the PHP config.
We wrote PHP code to pull the two out of the PHP config and
into environment variables before running "tailscale web".

The pfSense web UI is served over https, we need "tailscale web"
to also support https in order to put it in an <iframe>.

### DIFF
--- a/cmd/tailscale/cli/web.go
+++ b/cmd/tailscale/cli/web.go
@@ -7,6 +7,7 @@ package cli
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	_ "embed"
 	"encoding/json"
 	"encoding/xml"
@@ -19,6 +20,7 @@ import (
 	"net/http"
 	"net/http/cgi"
 	"net/url"
+	"os"
 	"os/exec"
 	"runtime"
 	"strings"
@@ -83,6 +85,29 @@ var webArgs struct {
 	cgi    bool
 }
 
+func tlsConfigFromEnvironment() *tls.Config {
+	crt := os.Getenv("TLS_CRT_PEM")
+	key := os.Getenv("TLS_KEY_PEM")
+	if crt == "" || key == "" {
+		return nil
+	}
+
+	// We support passing in the complete certificate and key from environment
+	// variables because pfSense stores its cert+key in the PHP config. We populate
+	// TLS_CRT_PEM and TLS_KEY_PEM from PHP code before starting tailscale web.
+	// These are the PEM-encoded Certificate and Private Key.
+
+	cert, err := tls.X509KeyPair([]byte(crt), []byte(key))
+	if err != nil {
+		log.Printf("tlsConfigFromEnvironment: %v", err)
+
+		// Fallback to unencrypted HTTP.
+		return nil
+	}
+
+	return &tls.Config{Certificates: []tls.Certificate{cert}}
+}
+
 func runWeb(ctx context.Context, args []string) error {
 	if len(args) > 0 {
 		log.Fatalf("too many non-flag arguments: %q", args)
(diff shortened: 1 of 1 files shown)


cmd/tailscale/cli/web.go
```diff
@@ -83,6 +85,35 @@ var webArgs struct {
 	cgi    bool
 }
 
+func tlsCertFromEnvironment(listen string) (*http.Server, error) {
+	crt := os.Getenv("TLS_CRT")
+	key := os.Getenv("TLS_KEY")
+	if crt != "" && key != "" {
+		fmt.Println(crt)
```
bradfitz, line 92: debug? delete?

cmd/tailscale/cli/web.go
```diff
@@ -83,6 +85,35 @@ var webArgs struct {
 	cgi    bool
 }
 
+func tlsCertFromEnvironment(listen string) (*http.Server, error) {
+	crt := os.Getenv("TLS_CRT")
```
bradfitz, line 89: these names would make me assume these are paths to files on disks, not the PEM contents themselves.

cmd/tailscale/cli/web.go
```diff
@@ -83,6 +85,35 @@ var webArgs struct {
 	cgi    bool
 }
 
+func tlsCertFromEnvironment(listen string) (*http.Server, error) {
```
bradfitz, line 88: it's named as if it returns a "tlsCert" but it returns a Server. And it returning (nil, nil) is worth documenting.

cmd/tailscale/cli/web.go
```diff
@@ -83,6 +85,35 @@ var webArgs struct {
...
+		cert, err := tls.X509KeyPair([]byte(crt), []byte(key))
+		if err != nil {
+			return nil, err
+		}
+
+		tlsConfig := &tls.Config{
+			Certificates: []tls.Certificate{cert},
+		}
+
+		// Build a server:
```
bradfitz, line 105: comment doesn't add anything. the server and tlsConfig variables don't much either. could just `return &http.Server{...}, nil` probably

cmd/tailscale/cli/web.go
```diff
@@ -83,6 +85,35 @@ var webArgs struct {
 	cgi    bool
 }
 
+func tlsCertFromEnvironment(listen string) (*http.Server, error) {
+	crt := os.Getenv("TLS_CRT")
+	key := os.Getenv("TLS_KEY")
+	if crt != "" && key != "" {
```
bradfitz, line 91: Go in general prefers early returns and outdenting

so inverting this,

```go
   if crt == "" || key == "" {
       return nil, nil
   }
   ....
```

cmd/tailscale/cli/web.go
```diff
@@ -96,8 +127,14 @@ func runWeb(ctx context.Context, args []string) error {
 		return nil
 	}
 
-	log.Printf("web server running on: %s", urlOfListenAddr(webArgs.listen))
-	return http.ListenAndServe(webArgs.listen, http.HandlerFunc(webHandler))
+	server, err := tlsCertFromEnvironment(webArgs.listen)
+	if server != nil && err == nil {
```
bradfitz, line 131: You don't handle the err != nil case.

cmd/tailscale/cli/web.go
DentonGentry: Removed.

cmd/tailscale/cli/web.go
DentonGentry: Added "_PEM" to environment variable names.

cmd/tailscale/cli/web.go
DentonGentry: Renamed tlsConfigFromEnvironment and made it return tls.Config.

cmd/tailscale/cli/web.go
DentonGentry: Made it return tls.Config.

cmd/tailscale/cli/web.go
DentonGentry: Done.

cmd/tailscale/cli/web.go
DentonGentry: No longer has the err here.

bradfitz (approved):

PR #2851, opened 2021-09-15
cmd/tailscaled: move Windows state to ProgramData

DentonGentry:
ipn/ipnserver: move Windows state to ProgramData

C:\WINDOWS\system32\config\systemprofile\AppData\Local\
is frequently cleared for almost any reason: Windows updates,
System Restore, even various System Cleaner utilities.

The server-state.conf file in AppData\Local could be deleted
at any time, which would break login until the node is removed
from the Admin Panel allowing it to create a new key.

Carefully copy any AppData state to ProgramData at startup.
If copying the state fails, continue to use AppData so at
least there will be connectivity. If there is no state,
use ProgramData.

We also migrate the log.conf file. Very old versions of
Tailscale named the EXE tailscale-ipn, so the log conf was
tailscale-ipn.log.conf and more recent versions preserved
this filename and cmdName in logs. In this migration we
always update the filename to
c:\ProgramData\Tailscale\tailscaled.log.conf

Updates https://github.com/tailscale/tailscale/issues/2856

### DIFF
--- a/ipn/ipnserver/server.go
+++ b/ipn/ipnserver/server.go
@@ -20,6 +20,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"os/user"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
@@ -37,6 +38,7 @@ import (
 	"tailscale.com/log/filelogger"
 	"tailscale.com/logtail/backoff"
 	"tailscale.com/net/netstat"
+	"tailscale.com/paths"
 	"tailscale.com/safesocket"
 	"tailscale.com/smallzstd"
 	"tailscale.com/types/logger"
@@ -581,6 +583,28 @@ func (s *server) writeToClients(n ipn.Notify) {
 	}
 }
 
+// tryWindowsAppDataMigration attempts to copy the Windows state file
+// from its old location to the new location. (Issue 2856)
+//
+// Tailscale 1.14 and before stored state under %LocalAppData%
+// (usually "C:\WINDOWS\system32\config\systemprofile\AppData\Local"
+// when tailscaled.exe is running as a non-user system service).
+// However it is frequently cleared for almost any reason: Windows
+// updates, System Restore, even various System Cleaner utilities.
+//
+// Returns a string of the path to use for the state file.
+// This will be a fallback %LocalAppData% path if migration fails,
+// a %ProgramData% path otherwise.
+func tryWindowsAppDataMigration(path string) string {
+	if path != paths.DefaultTailscaledStateFile() {
+		// If they're specifying a non-default path, just trust that they know
+		// what they are doing.
+		return path
+	}
+	oldFile := filepath.Join(os.Getenv("LocalAppData"), "Tailscale", "server-state.conf")
+	return paths.TryConfigFileMigration(oldFile, path)
+}
+
 // Run runs a Tailscale backend service.
 // The getEngine func is called repeatedly, once per connection, until it returns an engine successfully.
 func Run(ctx context.Context, logf logger.Logf, logid string, getEngine func() (wgengine.Engine, error), opts Options) error {
@@ -614,23 +638,27 @@ func Run(ctx context.Context, logf logger.Logf, logid string, getEngine func() (
 	var store ipn.StateStore
 	if opts.StatePath != "" {
 		const kubePrefix = "kube:"
+		path := opts.StatePath
 		switch {
-		case strings.HasPrefix(opts.StatePath, kubePrefix):
-			secretName := strings.TrimPrefix(opts.StatePath, kubePrefix)
+		case strings.HasPrefix(path, kubePrefix):
+			secretName := strings.TrimPrefix(path, kubePrefix)
 			store, err = ipn.NewKubeStore(secretName)
 			if err != nil {
 				return fmt.Errorf("ipn.NewKubeStore(%q): %v", secretName, err)
 			}
 		default:
-			store, err = ipn.NewFileStore(opts.StatePath)
+			if runtime.GOOS == "windows" {
+				path = tryWindowsAppDataMigration(path)
+			}
+			store, err = ipn.NewFileStore(path)
 			if err != nil {
-				return fmt.Errorf("ipn.NewFileStore(%q): %v", opts.StatePath, err)
+				return fmt.Errorf("ipn.NewFileStore(%q): %v", path, err)
 			}
 		}
 		if opts.AutostartStateKey == "" {
 			autoStartKey, err := store.ReadState(ipn.ServerModeStartKey)
 			if err != nil && err != ipn.ErrStateNotExist {
-				return fmt.Errorf("calling ReadState on %s: %w", opts.StatePath, err)
+				return fmt.Errorf("calling ReadState on %s: %w", path, err)
 			}
 			key := string(autoStartKey)
 			if strings.HasPrefix(key, "user-") {
--- a/paths/migrate.go
+++ b/paths/migrate.go
@@ -1,0 +1,54 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package paths
+
+import (
+	"log"
+	"os"
+	"path/filepath"
+)
+
+// TryConfigFileMigration carefully copies the contents of oldFile to
+// newFile, returning the path which should be used to read the config.
+// - if newFile already exists, don't modify it just return its path
+// - if neither oldFile nor newFile exist, return newFile for a fresh
+//   default config to be written to.
+// - if oldFile exists but copying to newFile fails, return oldFile so
+//   there will at least be some config to work with.
+func TryConfigFileMigration(oldFile, newFile string) string {
+	_, err := os.Stat(newFile)
+	if err == nil {
+		// Common case for a system which has already been migrated.
+		return newFile
+	}
+	if !os.IsNotExist(err) {
+		log.Printf("TryConfigFileMigration failed; new file: %v", err)
+		return newFile
+	}
+
+	contents, err := os.ReadFile(oldFile)
+	if err != nil {
+		// Common case for a new user.
+		return newFile
+	}
+
+	os.MkdirAll(filepath.Dir(newFile), 0700)
+	err = os.WriteFile(newFile, contents, 0600)
+	if err != nil {
+		removeErr := os.Remove(newFile)
+		if removeErr != nil {
+			log.Printf("TryConfigFileMigration failed; write newFile no cleanup: %v, remove err: %v",
+				err, removeErr)
+			return oldFile
+		}
+		log.Printf("TryConfigFileMigration failed; write newFile: %v", err)
+		return oldFile
+	}
+
+	log.Printf("TryConfigFileMigration: successfully migrated: from %v to %v",
+		oldFile, newFile)
+
+	return newFile
+}
--- a/paths/paths.go
+++ b/paths/paths.go
@@ -55,7 +55,7 @@ func DefaultTailscaledStateFile() string {
 		return f()
 	}
 	if runtime.GOOS == "windows" {
-		return filepath.Join(os.Getenv("LocalAppData"), "Tailscale", "server-state.conf")
+		return filepath.Join(os.Getenv("ProgramData"), "Tailscale", "server-state.conf")
 	}
 	return ""
 }
--- a/log/filelogger/log.go
+++ b/log/filelogger/log.go
@@ -36,7 +36,7 @@ func New(fileBasePrefix, logID string, logf logger.Logf) logger.Logf {
 	if logf == nil {
 		panic("nil logf")
 	}
-	dir := filepath.Join(os.Getenv("LocalAppData"), "Tailscale", "Logs")
+	dir := filepath.Join(os.Getenv("ProgramData"), "Tailscale", "Logs")
 
 	if err := os.MkdirAll(dir, 0700); err != nil {
 		log.Printf("failed to create local log directory; not writing logs to disk: %v", err)
--- a/logpolicy/logpolicy.go
+++ b/logpolicy/logpolicy.go
@@ -372,15 +372,34 @@ func New(collection string) *Policy {
 
 	cfgPath := filepath.Join(dir, fmt.Sprintf("%s.log.conf", cmdName))
 
-	// The Windows service previously ran as tailscale-ipn.exe, so
-	// let's keep using that log base name if it exists.
 	if runtime.GOOS == "windows" && cmdName == "tailscaled" {
-		const oldCmdName = "tailscale-ipn"
-		oldPath := filepath.Join(dir, oldCmdName+".log.conf")
-		if fi, err := os.Stat(oldPath); err == nil && fi.Mode().IsRegular() {
-			cfgPath = oldPath
-			cmdName = oldCmdName
+		// Tailscale 1.14 and before stored state under %LocalAppData%
+		// (usually "C:\WINDOWS\system32\config\systemprofile\AppData\Local"
+		// when tailscaled.exe is running as a non-user system service).
+		// However it is frequently cleared for almost any reason: Windows
+		// updates, System Restore, even various System Cleaner utilities.
+		//
+		// The Windows service previously ran as tailscale-ipn.exe, so
+		// machines which ran very old versions might still have their
+		// log conf named %LocalAppData%\tailscale-ipn.log.conf
+		//
+		// Machines which started using Tailscale more recently will have
+		// %LocalAppData%\tailscaled.log.conf
+		//
+		// Attempt to migrate the log conf to C:\ProgramData\Tailscale
+		oldDir := filepath.Join(os.Getenv("LocalAppData"), "Tailscale")
+
+		oldPath := filepath.Join(oldDir, "tailscaled.log.conf")
+		if fi, err := os.Stat(oldPath); err != nil || !fi.Mode().IsRegular() {
+			// *Only* if tailscaled.log.conf does not exist,
+			// check for tailscale-ipn.log.conf
+			oldPathOldCmd := filepath.Join(oldDir, "tailscale-ipn.log.conf")
+			if fi, err := os.Stat(oldPathOldCmd); err == nil && fi.Mode().IsRegular() {
+				oldPath = oldPathOldCmd
+			}
 		}
+
+		cfgPath = paths.TryConfigFileMigration(oldPath, cfgPath)
 	}
 
 	var oldc *Config


cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,34 @@ var subCommands = map[string]*func([]string) error{
 	"debug":                   &debugModeFunc,
 }
 
+func tryWindowsAppDataMigration() {
```
bradfitz, line 100: worth some comments.

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,34 @@ var subCommands = map[string]*func([]string) error{
...
+	if runtime.GOOS != "windows" {
+		return
+	}
+
+	localAppData := filepath.Join(os.Getenv("LocalAppData"), "Tailscale")
```
bradfitz, line 105: maybe call this variable `oldDir` for clarity? both that it's the old location and that it's a directory and not a file.

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,34 @@ var subCommands = map[string]*func([]string) error{
...
+	_, err := os.Stat(localAppData)
+	if os.IsNotExist(err) {
+		return
```
bradfitz, line 108:
```
// Common case.
return
```

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,34 @@ var subCommands = map[string]*func([]string) error{
...
+	}
+	if err != nil {
+		log.Printf("tryWindowsAppDataMigration failed; no AppDataLocal: %v", err)
```
bradfitz, line 111: "no" isn't quite right... if it were "no" you would've returned on 108. it's likely permissions here.

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,34 @@ var subCommands = map[string]*func([]string) error{
...
+		return
+	}
+
+	programData := filepath.Join(os.Getenv("ProgramData"), "Tailscale")
+	_, err = os.Stat(programData)
+	if err == nil || os.IsNotExist(err) {
```
bradfitz, line 117: if it already exists (err == nil), don't we want to bail and not try to move a directory atop (in?) a directory?

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,34 @@ var subCommands = map[string]*func([]string) error{
...
+		err = os.Rename(localAppData, programData)
+		if err != nil {
+			log.Printf("tryWindowsAppDataMigration rename failed: %v", err)
+			return
+		}
+	} else {
+		log.Printf("tryWindowsAppDataMigration ProgramData stat failed: %v", err)
```
bradfitz, line 124: do this earlier, return early, then outdent the rest?

paths/paths.go
```diff
@@ -55,7 +55,16 @@ func DefaultTailscaledStateFile() string {
 		return f()
 	}
 	if runtime.GOOS == "windows" {
-		return filepath.Join(os.Getenv("LocalAppData"), "Tailscale", "server-state.conf")
+		programData := filepath.Join(os.Getenv("ProgramData"), "Tailscale", "server-state.conf")
+		if _, err := os.Stat(programData); err == nil {
+			return programData
+		}
+		localAppData := filepath.Join(os.Getenv("LocalAppData"), "Tailscale", "server-state.conf")
```
bradfitz, line 62: // This is where Tailscale 1.14 and earlier stored the server-state. (Issue NNN)
// We still recognize it as a fallback.

cmd/tailscaled/tailscaled.go
DentonGentry: Added comments.

cmd/tailscaled/tailscaled.go
DentonGentry: Done.

cmd/tailscaled/tailscaled.go
DentonGentry: Done.

cmd/tailscaled/tailscaled.go
DentonGentry: Changed to "stat failed"

cmd/tailscaled/tailscaled.go
DentonGentry: Oh, yeah that is an important one. Logged a message if both ProgramData and AppData\Local exist, but since there is a legitimate case where that can happen, don't do anything else.

cmd/tailscaled/tailscaled.go
DentonGentry: Done.

paths/paths.go
DentonGentry: Done.

cmd/tailscaled/tailscaled.go
bradfitz: "stat failed" is redundant with its own stringification of the error from stat:

```
bradfitz@tsdev:~$ cat stat.go 
package main

import (
        "fmt"
        "os"
)

func main() {
        _, err := os.Stat("/root/.ssh")
        fmt.Printf("foo: %v\n", err)
}
bradfitz@tsdev:~$ go run stat.go 
foo: stat /root/.ssh: permission denied
```

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,53 @@ var subCommands = map[string]*func([]string) error{
 	"debug":                   &debugModeFunc,
 }
 
+// Tailscale 1.14 and before stored state in
```
bradfitz, line 100: s/in/under/

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,53 @@ var subCommands = map[string]*func([]string) error{
...
+// C:\WINDOWS\system32\config\systemprofile\AppData\Local
```
bradfitz, line 101: in %LocalAppData% (usually "C:\WINDOWS\system32\config\systemprofile\AppData\Local"
when tailscaled.exe is running as a non-user system service)

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,53 @@ var subCommands = map[string]*func([]string) error{
...
+	}
+
+	oldDir := filepath.Join(os.Getenv("LocalAppData"), "Tailscale")
+	_, err := os.Stat(oldDir)
+	if os.IsNotExist(err) {
+		// Common case.
+		return
+	}
+	if err != nil {
+		log.Printf("tryWindowsAppDataMigration failed; cannot stat AppDataLocal: %v", err)
```
bradfitz, line 120: "tryWindowsAppDataMigration failed: %v", err

... is probably sufficient.

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,53 @@ var subCommands = map[string]*func([]string) error{
...
+		return
+	} else if !os.IsNotExist(err) {
+		log.Printf("tryWindowsAppDataMigration ProgramData stat failed: %v", err)
+		return
+	}
+
+	// Atomically move AppData\Local\Tailscale to ProgramData\Tailscale
+	err = os.Rename(oldDir, programData)
+	if err != nil {
+		log.Printf("tryWindowsAppDataMigration rename failed: %v", err)
```
bradfitz, line 142: "rename" is also redundant

cmd/tailscaled/tailscaled.go
DentonGentry: Done.

cmd/tailscaled/tailscaled.go
DentonGentry: Done.

cmd/tailscaled/tailscaled.go
DentonGentry: Done.

cmd/tailscaled/tailscaled.go
DentonGentry: Done.

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,54 @@ var subCommands = map[string]*func([]string) error{
 	"debug":                   &debugModeFunc,
 }
 
+// Tailscale 1.14 and before stored state under %LocalAppData%
+// (usually "C:\WINDOWS\system32\config\systemprofile\AppData\Local"
+// when tailscaled.exe is running as a non-user system service).
+// However it is frequently cleared for almost any reason: Windows
+// updates, System Restore, even various System Cleaner utilities.
+//
+// We carefully move any state in AppData\Local to ProgramData
```
bradfitz, line 106: Let's reword this from "we" to the subject (the func name) and make it first like https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences 

Like:

```
// tryWindowsAppDataMigration attempts to move the Windows state storage directory
// from its old location to the new location. (Issue NNN)
//
// Tailscale 1.14 and before ...
//
// This function does not return an error; on rename failure, code elsewhere
// in Tailscale will continue to use the old location as a fallback.

```

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,54 @@ var subCommands = map[string]*func([]string) error{
...
+	_, err = os.Stat(programData)
+	if err == nil {
+		// Both AppData\Local\Tailscale and ProgramData\Tailscale exist. There are
+		// cases where this can happen, like if one downgrades from a version which
+		// stores state in ProgramData back to an earlier version that re-creates
+		// state in AppData\Local.
+		// This is unusual so we log it, but don't otherwise mess with the files.
+		log.Printf("tryWindowsAppDataMigration ProgramData and AppDataLocal both exist")
+		return
+	} else if !os.IsNotExist(err) {
```
bradfitz, line 137: drop else, move if to new line? (early return + outdent style)

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,54 @@ var subCommands = map[string]*func([]string) error{
...
+
+	programData := filepath.Join(os.Getenv("ProgramData"), "Tailscale")
+	_, err = os.Stat(programData)
+	if err == nil {
+		// Both AppData\Local\Tailscale and ProgramData\Tailscale exist. There are
+		// cases where this can happen, like if one downgrades from a version which
+		// stores state in ProgramData back to an earlier version that re-creates
+		// state in AppData\Local.
+		// This is unusual so we log it, but don't otherwise mess with the files.
+		log.Printf("tryWindowsAppDataMigration ProgramData and AppDataLocal both exist")
```
bradfitz, line 135: missing colon after tryWindowsAppDataMigration?

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,54 @@ var subCommands = map[string]*func([]string) error{
...
+		return
+	} else if !os.IsNotExist(err) {
+		log.Printf("tryWindowsAppDataMigration failed; ProgramData: %v", err)
```
bradfitz, line 138: "... failed: %v" is probably enough, as the error will include ProgramData in it. or replace ProgramData with "new dir: " maybe?

cmd/tailscaled/tailscaled.go
DentonGentry: Done.

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,56 @@ var subCommands = map[string]*func([]string) error{
...
+		return
+	}
+
+	// Atomically move AppData\Local\Tailscale to ProgramData\Tailscale
+	err = os.Rename(oldDir, programData)
+	if err != nil {
+		log.Printf("tryWindowsAppDataMigration: %v", err)
+		return
+	}
```
bradfitz, line 147: log.Printf("tryWindowsAppDataMigration: successfully migrated: from %v to %v", oldDir, programData)

cmd/tailscaled/tailscaled.go
DentonGentry: Done.

cmd/tailscaled/tailscaled.go
DentonGentry: Done.

cmd/tailscaled/tailscaled.go
DentonGentry: Replaced with new dir

cmd/tailscaled/tailscaled.go
DentonGentry: Done.

bradfitz (commented):
LGTM but I'd like to either test it myself (and remind myself how the windows filesystem behaves in the process) or have others also review this, given its sensitivity.

But if you need to submit it to get an unstable build out, go for it. We can always tweak later.

DentonGentry:
I'm content to wait to submit until we've fully tested it. I want to try it on Windows 7, I've only tried Windows 10 so far.

@dblohm7 this is the proposal of how to migrate our files from AppData\Local to ProgramData.

dblohm7:
> 
> 
> I'm content to wait to submit until we've fully tested it. I want to try it on Windows 7, I've only tried Windows 10 so far.
> 
> @dblohm7 this is the proposal of how to migrate our files from AppData\Local to ProgramData.

Looks reasonable.

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,60 @@ var subCommands = map[string]*func([]string) error{
 	"debug":                   &debugModeFunc,
 }
 
+// tryWindowsAppDataMigration attempts to move the Windows state storage
+// directory from its old location to the new location. (Issue 2282)
```
bradfitz, line 101: change this to 2856 now that I filed #2856 with an overview.

I linked 2282 from that bug.

Also add `Updates #2856` to the commit mesage.

DentonGentry: Done.

DentonGentry:
Windows 7 somehow ended up with a copy in ProgramData\Tailscale and another copy in AppData\Local\Tailscale

Maybe os.Rename on Windows 7 doesn't have an atomic move available in the filesystem, so it synthesizes one with a copy + delete?

Leaving a stale copy in AppData\Local\Tailscale is survivable, not necessarily derailing the idea.

DentonGentry:
Windows 7 and Windows 10 display the same behavior:
- the first time Tailscale is run, it *copies* (not moves, copies) the files from %LOCALAPPDATA%\Tailscale to %PROGRAMDATA%\Tailscale . The %LOCALAPPDATA%\Tailscale files are still there.
- If I unsinstll Tailscale and `rmdir /s %PROGRAMDATA%\Tailscale` to try again, then after reinstalling Tailscale the UI just says "Please restart the Tailscale Windows Service."

This is clearly not acceptable, I'm marking the PR as a draft to prevent submission and will look at why the service doesn't start.

bradfitz:
I guess I missed something... what changed to warrant this as ready again?

DentonGentry:
(got derailed by kids)

The system-state.conf file **does** get moved to C:\ProgramData.

Logs are still being written to C:\Windows\System32\config\systemprofile\AppData\Local\Tailscale\Logs which is preventing the AppData\Local directory from being fully removed. Looking at that now.

The second part of the last update: "If I unsinstall Tailscale and rmdir /s %PROGRAMDATA%\Tailscale to try again, then after reinstalling Tailscale the UI just says "Please restart the Tailscale Windows Service."
happens because system-state.conf did get moved. So I had recreated the original problem, system-state.conf is gone but the node key still exists. Attempting to reinstall 1.14 in this situation means it will say the backend needs to be restarted.

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,70 @@ var subCommands = map[string]*func([]string) error{
...
+	contents, err := os.ReadFile(oldFile)
+	if err != nil {
+		log.Printf("tryWindowsAppDataMigration failed; read oldFile: %v", err)
+		return
+	}
+
+	os.MkdirAll(filepath.Dir(newFile), 0755)
+	err = os.WriteFile(newFile, contents, 0600)
+	if err != nil {
+		remove_err := os.Remove(newFile)
```
bradfitz, line 149: removeErr instead of underscores: https://golang.org/doc/effective_go#mixed-caps

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,70 @@ var subCommands = map[string]*func([]string) error{
...
+		if remove_err == nil {
+			log.Printf("tryWindowsAppDataMigration failed; write newFile: %v", err)
+			return
```
bradfitz, line 152: move these two returns out into one at the end of the "if removeErr" block, so it's obvious that it doesn't fall through? and saves a line.

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,70 @@ var subCommands = map[string]*func([]string) error{
 	"debug":                   &debugModeFunc,
 }
 
+// tryWindowsAppDataMigration attempts to move the Windows state storage
```
bradfitz, line 100: s/move/copy/

cmd/tailscaled/tailscaled.go
```diff
@@ -96,6 +97,70 @@ var subCommands = map[string]*func([]string) error{
...
+// directory from its old location to the new location. (Issue 2856)
+//
+// Tailscale 1.14 and before stored state under %LocalAppData%
+// (usually "C:\WINDOWS\system32\config\systemprofile\AppData\Local"
+// when tailscaled.exe is running as a non-user system service).
+// However it is frequently cleared for almost any reason: Windows
+// updates, System Restore, even various System Cleaner utilities.
+//
+// This function does not return an error; on rename failure, code elsewhere
```
bradfitz, line 109: not "rename" anymore

maybe just "on failure"

DentonGentry:
Now only copies the state file, leaving it in the %LOCALAPPDATA% location to allow downgrade back to 1.14.

One concerning thing is that I never see the tryWindowsAppDataMigration for success, and I should. The C:\ProgramData\Tailscale directory appears with a copy of the state file in it.

cmd/tailscaled/tailscaled.go
DentonGentry: Done

cmd/tailscaled/tailscaled.go
DentonGentry: Done

cmd/tailscaled/tailscaled.go
DentonGentry: Done

cmd/tailscaled/tailscaled.go
DentonGentry: Done

crawshaw (approved):

cmd/tailscaled/tailscaled.go
```diff
@@ -141,6 +206,10 @@ func main() {
 		log.Fatalf("tailscaled does not take non-flag arguments: %q", flag.Args())
 	}
 
+	if args.statepath == paths.DefaultTailscaledStateFile() {
+		tryWindowsAppDataMigration()
```
bradfitz, line 210: I think this call is at the wrong stage of the Windows service lifecycle. See the "if beWindowsSubprocess" up on line 200. That's the path where we capture logs.

So you have the wrong process doing this migration.

This way is only taken when running tailscaled.exe by hand in a cmd.exe window or using the tswin debug tool.

DentonGentry: Should tryWindowsAppDataMigration() be called from beWindowsSubprocess() right before it calls startIPNServer() ?

bradfitz: Ideally we'd place it in one place that's shared by both code paths: the foreground tailscaled mode on Windows as well as the service mode.

ipnserver.Run perhaps?

I think I'd make DefaultTailscaledStateFile not do any fallback work; just have it return the new path only. 

The fallback work could probably be in ipnsever.Run before it calls ipn.NewFileStore

DentonGentry: Moved to ipnserver.Run, moved the fallback into ipnserver.
Verified that the log message appears now: `2021-09-17 16:11:58.8096887 -0700 PDT: 2021/09/17 16:11:58 tryWindowsAppDataMigration: successfully migrated: from C:\Windows\system32\config\systemprofile\AppData\Local\Tailscale\server-state.conf to C:\ProgramData\Tailscale\server-state.conf`

ipn/ipnserver/server.go
```diff
@@ -581,6 +583,76 @@ func (s *server) writeToClients(n ipn.Notify) {
...
+// tryWindowsAppDataMigration attempts to copy the Windows state storage
+// directory from its old location to the new location. (Issue 2856)
+//
+// Tailscale 1.14 and before stored state under %LocalAppData%
+// (usually "C:\WINDOWS\system32\config\systemprofile\AppData\Local"
+// when tailscaled.exe is running as a non-user system service).
+// However it is frequently cleared for almost any reason: Windows
+// updates, System Restore, even various System Cleaner utilities.
+//
+// On failure, return the old location as a fallback where appropriate.
```
bradfitz, line 595: clarify what it returns on success (or in general).

ipn/ipnserver/server.go
```diff
@@ -581,6 +583,76 @@ func (s *server) writeToClients(n ipn.Notify) {
...
+	if os.IsNotExist(err) {
+		// Common case for a freshly installed system.
+		return path
+	}
+	if err != nil {
+		log.Printf("tryWindowsAppDataMigration failed; AppDataLocal: %v", err)
+		return path
+	}
+
+	newFile := filepath.Join(os.Getenv("ProgramData"), "Tailscale", "server-state.conf")
```
bradfitz, line 614: do you even need this anymore filepath.Join and Getenv anymore? isn't this the same value as the `path` variable by definition now, after line 597's check?

Likewise, I'd move this up as the second return path, after "just trust they know what they're doing", like:


```go
	if path != paths.DefaultTailscaledStateFile() {
		// If they're specifying a non-default path, just trust that they know
		// what they are doing.
		return path
	}
        if _, err := os.Stat(path); err == nil {
              // Common case: the requested path already exists on disk, so it was either
              // a fresh install, or was already migrated.
              return path
       }
```

ipn/ipnserver/server.go
```diff
@@ -581,6 +583,76 @@ func (s *server) writeToClients(n ipn.Notify) {
...
+		// to %PROGRAMDATA% while leaving the original in %LOCALAPPDATA%.
+		return path
+	}
+	if !os.IsNotExist(err) {
+		log.Printf("tryWindowsAppDataMigration failed; new file: %v", err)
+		return path
+	}
+
+	// copy server-state.conf from %LOCALAPPDATA% to %PROGRAMDATA%
+	contents, err := os.ReadFile(oldFile)
```
bradfitz, line 627: no need to stat the oldFile first. You can just jump right to ReadFile it.

bradfitz (commented):
Commit message should now start with "ipn/ipnserver: move ..."

ipn/ipnserver/server.go
DentonGentry: Done

ipn/ipnserver/server.go
DentonGentry: Removed.
Could not make it a single line conditional because line 610 references err.

ipn/ipnserver/server.go
DentonGentry: Done

bradfitz:
BTW, I think we also need to effectively copy this migration code for tailscale.com/logpolicy and its `logsDir` func. It'll do the right thing going forward, but we don't want log IDs to move.

Maybe pull this code out into a shared func that both of the two places can share?

DentonGentry:
Moved file copy code to paths/migrate.go and made it handle the logs conf as well.

There is handling for very old Windows versions with a tailscale-ipn.log.conf file, which preserves the cmdName as "tailscale-ipn". This version does still handle finding a tailscale-ipn.log.conf, but unconditionally renames it to c:\ProgramData\tailscaled.log.conf and makes the cmdName be "tailscaled"

paths/migrate.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package paths returns platform and user-specific default paths to
```
bradfitz, line 5: Only one file in a package (conventionally the one named after the package itself) should have a package comment. Otherwise they get concatenated together in weird ways.

So delete these 3 lines and update the paths.go version if desired.

paths/migrate.go
```diff
@@ -0,0 +1,59 @@
...
+		return newFile
+	}
+	if !os.IsNotExist(err) {
+		log.Printf("TryConfigFileMigration failed; new file: %v", err)
+		return newFile
+	}
+
+	contents, err := os.ReadFile(oldFile)
+	if err != nil {
+		log.Printf("TryConfigFileMigration failed; read oldFile: %v", err)
```
bradfitz, line 36: in the IsNotExist case I think we want to be silent here. Otherwise there's scary log spam for all new users.

bradfitz (commented):
Looking nice! And close.

bradfitz:
Commit message title update:

```
ipn/ipnserver, paths, logpolicy: move Window config files out of %LocalAppData%
```

paths/migrate.go
DentonGentry: Done

paths/migrate.go
DentonGentry: Done

paths/migrate.go
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package paths
```
bradfitz, line 4: you need a blank line before this, otherwise the copyright comment is the package doc

you can verify with `go doc .` in the directory

bradfitz: otherwise LGTM with the caveat that I haven't actually patched it in and tried it. But seems right and I trust your testing.

DentonGentry: Added blank line.

PR #2853, opened 2021-09-15
cmd/cloner: reduce diff noise when changing command

josharian:
Spelling out the command to run for every type
means that changing the command makes for a large, repetitive diff.
Stop doing that.


cmd/cloner/cloner.go
```diff
@@ -149,7 +149,9 @@ const header = `// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserve
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// Code generated by tailscale.com/cmd/cloner -type %s; DO NOT EDIT.
+// Code generated by:
```
bradfitz, line 152: Does still this match the prescribed regexp for generated files? I thought the DO NOT EDIT had to be on the same line as "Code generated by" ?

bradfitz: https://golang.org/s/generatedcode ... 

> To convey to humans and machine tools that code is generated, generated source should have a line that matches the following regular expression (in Go syntax):
>
> `^// Code generated .* DO NOT EDIT\.$`

josharian: Grumble grumble nice things grumble grumble humans vs machines.

Fixed, thanks.

bradfitz (approved):

PR #2854, opened 2021-09-15
.github/workflows: skip net/dnsfallback go generate check on release branch

josharian:


.github/workflows/go_generate.yml
```diff
@@ -28,7 +28,14 @@ jobs:
         run: |
           mkdir gentools
           go build -o gentools/stringer golang.org/x/tools/cmd/stringer
-          PATH="$PATH:$(pwd)/gentools" go generate ./...
+          if [[ "${{github.ref}}" == release-branch/* ]]
```
bradfitz, line 31: whoa, == matches against globs? TIL.

.github/workflows/go_generate.yml
```diff
@@ -28,7 +28,14 @@ jobs:
         run: |
           mkdir gentools
           go build -o gentools/stringer golang.org/x/tools/cmd/stringer
```
bradfitz, line 30: I don't think we need this anymore. Awhile back I updated all the raw "stringer" calls to use "go run golang.org/x/tools/cmd/stringer" instead.

bradfitz (approved):

.github/workflows/go_generate.yml
josharian: Shell. 🤪

.github/workflows/go_generate.yml
josharian: Ack. Will fix in separate commit in this PR.

bradfitz (approved):

josharian:
Windows race failure is unrelated. Filed https://github.com/tailscale/tailscale/issues/2855 for later. (I'm busy working on a different flaky test now. 😢)

PR #2857, opened 2021-09-15
ipn/ipnlocal: add failing test, reduce UpdateEndpoints calls, start concurrency work

josharian:
- all: close fake userspace engines when tests complete
- ipn/ipnlocal: add failing test
- control/controlclient: replace TODO with explanation
- ipn/ipnlocal: only call UpdateEndpoints when the endpoints change
- ipn/ipnlocal: use a switch statement instead of an else-if chain
- ipn/ipnlocal: inline LocalBackend.getEngineStatus
- ipn/ipnlocal: add LocalBackend.broadcastStatusChanged

### DIFF
--- a/control/controlclient/client.go
+++ b/control/controlclient/client.go
@@ -69,7 +69,7 @@ type Client interface {
 	SetNetInfo(*tailcfg.NetInfo)
 	// UpdateEndpoints changes the Endpoint structure that will be sent
 	// in subsequent node registration requests.
-	// TODO: localPort seems to be obsolete, remove it.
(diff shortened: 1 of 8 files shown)


josharian:
This is a smorgasbord of commits: The beginning of work towards fixing https://github.com/tailscale/corp/issues/2579, some cleanup, and an unrelated bug/optimization fix. Please review commits independently.

crawshaw (approved):
Thanks for the commit stack, that made this easy to review.

josharian:
> So you should be able to use b.statusChanged.Broadcast directly.

Good point. I'll do that.

PR #2865, opened 2021-09-16
ipnlocal: fix sync.Cond-related test flakiness

josharian:
- Revert "ipn/ipnlocal: remove locks around sync.Cond.Broadcast call"
- ipn/ipnlocal: add LocalBackend.broadcastStatusChanged

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -601,18 +601,12 @@ func (b *LocalBackend) findExitNodeIDLocked(nm *netmap.NetworkMap) (prefsChanged
 func (b *LocalBackend) setWgengineStatus(s *wgengine.Status, err error) {
 	if err != nil {
 		b.logf("wgengine status error: %v", err)
-
-		b.statusLock.Lock()
-		b.statusChanged.Broadcast()
-		b.statusLock.Unlock()
+		b.broadcastStatusChanged()
 		return
 	}
 	if s == nil {
 		b.logf("[unexpected] non-error wgengine update with status=nil: %v", s)
-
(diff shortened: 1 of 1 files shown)


bradfitz:
https://pkg.go.dev/sync#Cond.Broadcast says:

> It is allowed but not required for the caller to hold c.L during the call.

So why does this fix things? Add some comments to the code so it's not reverted?

josharian:
I don't know. I'm happy to investigate more, but I'd like to get at least the revert submitted right away, to get the build stable again. You OK with that?

josharian:
> So why does this fix things? Add some comments to the code so it's not reverted?

The deadlock occurs in the Wait call here:

```go
// Requests the wgengine status, and does not return until the status
// was delivered (to the usual callback).
func (b *LocalBackend) requestEngineStatusAndWait() {
	b.logf("requestEngineStatusAndWait")

	b.statusLock.Lock()
	go b.e.RequestStatus()
	b.logf("requestEngineStatusAndWait: waiting...")
	b.statusChanged.Wait() // temporarily releases lock while waiting
	b.logf("requestEngineStatusAndWait: got status update.")
	b.statusLock.Unlock()
}
```

If the call to `go b.e.RequestStatus()` completes all of its work, including calling `b.statusChanged.Broadcast()` before we reach the `b.statusChanged.Wait()` line, then no one ever unblocks the waiter.

With the calls to Lock, though, the call to Broadcast cannot occur until the waiter is blocked in the wait.

It's not pretty, but it works. I'll add a comment.

josharian:
PTAL

bradfitz (approved):

PR #2868, opened 2021-09-17
util/codegen, cmd/cloner: organize and simplify cloner

josharian:
- util/codegen: add package
- util/codegen: add AssertStructUnchanged
- cmd/cloner: add Clone context to regen struct assignments
- cmd/cloner: simplify code
- cmd/cloner: delete some debug code
- util/codegen: add NamedTypes
- cmd/cloner: unify switch cases
- util/codegen: add ContainsPointers

### DIFF
--- a/util/codegen/codegen.go
+++ b/util/codegen/codegen.go
@@ -72,12 +72,12 @@ func NamedTypes(pkg *packages.Package) map[string]*types.Named {
 }
 
 // AssertStructUnchanged generates code that asserts at compile time that type t is unchanged.
+// thisPkg is the package containing t.
 // tname is the named type corresponding to t.
 // ctx is a single-word context for this assertion, such as "Clone".
-// thisPkg is the package containing t.
 // If non-nil, AssertStructUnchanged will add elements to imports
 // for each package path that the caller must import for the returned code to compile.
-func AssertStructUnchanged(t *types.Struct, tname, ctx string, thisPkg *types.Package, imports map[string]struct{}) []byte {
+func AssertStructUnchanged(t *types.Struct, thisPkg *types.Package, tname, ctx string, imports map[string]struct{}) []byte {
 	buf := new(bytes.Buffer)
 	w := func(format string, args ...interface{}) {
 		fmt.Fprintf(buf, format+"\n", args...)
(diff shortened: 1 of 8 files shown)


josharian:
I am in the process of hacking up a cmd/viewer, similar to cmd/cloner, but that generates readonly views rather than copies. It shares a lot of code and structure with cmd/cloner. This PR is preliminary work on cloner, cleaning up the code a bit and extracting some obviously re-usable parts.

Review commit-by-commit for best effect.

dsnet:
I'm starting review now. Just wanted to comment that a few weeks ago I was exploring how to make the cloner logic nicer with the use of generics. I'd love to work with you on seeing in we can get that integrated in the future.

josharian:
> Just wanted to comment that a few weeks ago I was exploring how to make the cloner logic nicer with the use of generics.

Cool, SGTM.

I'd also like to make the logic a bit less one-off: Instead of panicking on pointers-to-pointers, it should do recursive codegen and (as a consequence) handle a broader swath of types. Sounds scary, but in practice probably quite straightforward. But not urgent, either.

Happy to work with you on either/both of those efforts. I'll put up a WIP PR for cmd/viewer once this one is in so you can see what I'm up to there.

util/codegen/codegen.go
```diff
@@ -38,3 +40,42 @@ func WriteFormatted(code []byte, path string) error {
...
 	return nil
 }
+
+// AssertStructUnchanged generates code that asserts at compile time that type t is unchanged.
+// tname is the named type corresponding to t.
+// ctx is a single-word context for this assertion, such as "Clone".
+// thisPkg is the package containing t.
+// If non-nil, AssertStructUnchanged will add elements to imports
+// for each package path that the caller must import for the returned code to compile.
+func AssertStructUnchanged(t *types.Struct, tname, ctx string, thisPkg *types.Package, imports map[string]struct{}) []byte {
```
dsnet, line 80: Optional:  Perhaps `thisPkg` should come right between `t` to match the fact that we usually refer to qualified type as `PkgPath.TypeName`

dsnet (approved):
Sigh... GitHub makes it hard to review these. It's all mechanical changes.

dsnet:
My apologies for the delay, some drama with the kiddos ;)

util/codegen/codegen.go
josharian: Done. Did it as an additional commit to avoid pointless merge pain.

PR #2874, opened 2021-09-18
tailcfg, health: add way for control plane to add problems to health check

bradfitz:
So if the control plane knows that something's broken about the node, it can
include problem(s) in MapResponse and "tailscale status" will show it.
(and GUIs in the future, as it's in ipnstate.Status/JSON)

This also bumps the MapRequest.Version, though it's not strictly
required. Doesn't hurt.


crawshaw (approved):
Hmm I feel like we discussed this with the last health check change and there was some other mechanism for this. Oh well.

Issue #2132, opened 2021-06-15
Windows flushdns not always working


DentonGentry:
The Windows client issues an `ipconfig /flushdns` when the Tailscale link connects. For corporate customers this allows use of public Outlook addresses while on the public Internet, and private Tailscale IPs when connected.

User reports that the DNS cache flush is not always working, they frequently see the public Outlook IP address being used after the Tailscale connection comes up.

PR #2875, opened 2021-09-19
wgengine: flush DNS cache after major link change.

DentonGentry:
Fixes https://github.com/tailscale/tailscale/issues/2132

### DIFF
--- a/net/dns/flush_default.go
+++ b/net/dns/flush_default.go
@@ -1,0 +1,12 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build !windows
+// +build !windows
+
+package dns
+
+func flushCaches() error {
+	return nil
+}
--- a/net/dns/flush_windows.go
+++ b/net/dns/flush_windows.go
@@ -9,11 +9,20 @@ import (
 	"os/exec"
 )
 
-// Flush clears the local resolver cache.
-func Flush() error {
+func flushCaches() error {
 	out, err := exec.Command("ipconfig", "/flushdns").CombinedOutput()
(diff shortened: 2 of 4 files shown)


net/dns/flush_default.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build !windows
+// +build !windows
+
+package dns
+
+func Flush() error {
```
bradfitz, line 10: forget to add a file to this PR?

Also, a convention is to put the exported "func Flush" in a portable file where you can attach docs (just once) and then have 2+ unexported versions without docs.

DentonGentry: There was already a flush_windows.go, called from router_windows.go.

Added unexported flushCaches(). Windows retains its public Flush() method, because router_windows cannot easily get access to the dns.Manager to call FlushCaches().

net/dns/flush_windows.go
```diff
@@ -9,11 +9,20 @@ import (
...
 		return fmt.Errorf("%v (output: %s)", err, out)
 	}
 	return nil
 }
+
+// Flush clears the local resolver cache.
+//
+// Only Windows has a public dns.Flush, needed in router_windows.go. Other
+// platforms like Linux need a different flush implementation depending on
+// the Dns Manager. There is a FlushCaches method on the Manager which
```
bradfitz, line 24: DNS manager or DNSManager 

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

bradfitz (approved):
LGTM after minor nit

net/dns/flush_windows.go
DentonGentry: Done.

PR #2884, opened 2021-09-20
Danderson/node key

danderson:


types/key/node.go
```diff
@@ -0,0 +1,244 @@
...
+// created by SealTo, and returns the inner cleartext if ciphertext is
+// a valid box from p to k.
+func (k NodePrivate) OpenFrom(p NodePublic, ciphertext []byte) (cleartext []byte, ok bool) {
+	if k.IsZero() || p.IsZero() {
+		panic("can't open with zero keys")
+	}
+	if len(ciphertext) < 24 {
+		return nil, false
+	}
+	var nonce [24]byte
```
josharian, line 120: Could do `nonce := (*[24]byte)(ciphertext)`

types/key/node.go
```diff
@@ -0,0 +1,244 @@
...
+	copy(nonce[:], ciphertext)
+	return box.Open(nil, ciphertext[len(nonce):], &nonce, &p.k, &k.k)
+}
+
+func (k NodePrivate) UntypedHexString() string {
+	return hex.EncodeToString(k.k[:])
+}
+
+// NodePublic is the public portion of a a NodePrivate.
```
josharian, line 129: a a

types/key/node.go
```diff
@@ -0,0 +1,244 @@
...
+	return ret
+}
+
+// MarshalText implements encoding.TextMarshaler.
+func (k NodePrivate) MarshalText() ([]byte, error) {
+	return toHex(k.k[:], nodePrivateHexPrefix), nil
+}
+
+// MarshalText implements encoding.TextUnmarshaler.
+func (k *NodePrivate) UnmarshalText(b []byte) error {
```
josharian, line 91: Do we want to reject unmarshalling into a non-zero NodePrivate?

types/key/node.go
```diff
@@ -0,0 +1,244 @@
...
+// string format, please use MarshalText/UnmarshalText.
+func ParseNodePublicUntyped(raw mem.RO) (NodePublic, error) {
+	var ret NodePublic
+	if err := parseHex(ret.k[:], raw, mem.B(nil)); err != nil {
+		return NodePublic{}, err
+	}
+	return ret, nil
+}
+
+func NodePublicFromRaw32(raw mem.RO) NodePublic {
```
josharian, line 150: So Untyped is hex-encoded and Raw is not? Is there a package level comment somewhere about naming conventions?

Also, a doc string for this function would be nice; it would be non-obvious to me w/o the implementation.

bradfitz (approved):
Skimmed. All seems straightforward.

Fix depaware first?

types/key/node.go
```diff
@@ -0,0 +1,244 @@
...
+// brackets.
+func (k NodePublic) ShortString() string {
+	return debug32(k.k)
+}
+
+func (k NodePublic) AppendTo(buf []byte) []byte {
+	return append(buf, k.k[:]...)
+}
+
+func (k *NodePublic) ReadFrom(r io.Reader) (int64, error) {
```
josharian, line 175: ReadFromRaw, since it is using the bytes directly? Document that it reads exactly 32 bytes on success? Docs please, in any case.

types/key/node.go
```diff
@@ -0,0 +1,244 @@
...
+	var z NodePublic
+	if *k != z {
+		return 0, errors.New("refusing to read into non-zero NodePublic")
+	}
+	n, err := io.ReadFull(r, k.k[:])
+	return int64(n), err
+}
+
+func (k NodePublic) WriteTo(w io.Writer) (int64, error) {
```
josharian, line 184: WriteToRaw? Docs please.

types/key/node.go
```diff
@@ -0,0 +1,244 @@
...
+func (k NodePublic) MarshalText() ([]byte, error) {
+	return toHex(k.k[:], nodePublicHexPrefix), nil
+}
+
+// MarshalText implements encoding.TextUnmarshaler.
+func (k *NodePublic) UnmarshalText(b []byte) error {
+	return parseHex(k.k[:], mem.B(b), mem.S(nodePublicHexPrefix))
+}
+
+// wireguardGoString prints k in the same format used by wireguard-go.
```
josharian, line 225: Nit, but capital G in WireGuard.

types/key/node.go
```diff
@@ -0,0 +1,244 @@
...
+	return string(bs)
+}
+
+// MarshalText implements encoding.TextMarshaler.
+func (k NodePublic) MarshalText() ([]byte, error) {
+	return toHex(k.k[:], nodePublicHexPrefix), nil
+}
+
+// MarshalText implements encoding.TextUnmarshaler.
+func (k *NodePublic) UnmarshalText(b []byte) error {
```
josharian, line 221: Refuse to overwrite non-zero k?

types/key/node_test.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func TestNodeSerialization(t *testing.T) {
+	serialized := `{
+      "Priv": "privkey:40ab1b58e9076c7a4d9d07291f5edf9d1aa017eb949624ba683317f48a640369",
+      "Pub":"nodekey:50d20b455ecf12bc453f83c2cfdb2a24925d06cf2598dcaa54e91af82ce9f765"
+    }`
+
+	// Carefully check that the expected serialized data decodes and
+	// reencodes to the expected keys. These types are serialized to
```
josharian, line 49: re-encodes

derp/derp_client.go
```diff
@@ -220,15 +210,16 @@ func (c *Client) send(dstKey key.Public, pkt []byte) (ret error) {
 	c.wmu.Lock()
 	defer c.wmu.Unlock()
 	if c.rate != nil {
-		pktLen := frameHeaderLen + len(dstKey) + len(pkt)
+		pktLen := frameHeaderLen + 32 + len(pkt)
```
josharian, line 213: Can we add an exported const to package key? Then this can read `+ key.NodeLen +` or the like. Or perhaps better, a Len method, so it reads `+ dstKey.Len() +`.

derp/derp_client.go
```diff
@@ -220,15 +210,16 @@ func (c *Client) send(dstKey key.Public, pkt []byte) (ret error) {
...
 		if !c.rate.AllowN(time.Now(), pktLen) {
 			return nil // drop
 		}
 	}
-	if err := writeFrameHeader(c.bw, frameSendPacket, uint32(len(dstKey)+len(pkt))); err != nil {
+
+	if err := writeFrameHeader(c.bw, frameSendPacket, uint32(32+len(pkt))); err != nil {
```
josharian, line 219: Ditto

derp/derp_server.go
```diff
@@ -774,10 +773,11 @@ func (c *sclient) handleFrameClosePeer(ft frameType, fl uint32) error {
 	if !c.canMesh {
 		return fmt.Errorf("insufficient permissions")
 	}
-	var targetKey key.Public
-	if _, err := io.ReadFull(c.br, targetKey[:]); err != nil {
+	var bs [32]byte
+	if _, err := io.ReadFull(c.br, bs[:]); err != nil {
 		return err
 	}
+	targetKey := key.NodePublicFromRaw32(mem.B(bs[:]))
```
josharian, line 780: Why do a ReadFull into an untyped buffer and then do a FromRaw32? Can we instead do a single key.ReadFrom call? (I have already forgotten the exact name.) And then maybe we get delete FromRaw32?

derp/derp_server.go
```diff
@@ -1018,9 +1018,9 @@ func (s *Server) verifyClient(clientKey key.Public, info *clientInfo) error {
 }
 
 func (s *Server) sendServerKey(lw *lazyBufioWriter) error {
-	buf := make([]byte, 0, len(magic)+len(s.publicKey))
+	buf := make([]byte, 0, len(magic)+32)
```
josharian, line 1021: s.publicKey.Len()

derp/derp_server.go
```diff
@@ -1124,21 +1117,19 @@ func (s *Server) recvClientKey(br *bufio.Reader) (clientKey key.Public, info *cl
...
+	var bs [32]byte
+	if _, err := io.ReadFull(br, bs[:]); err != nil {
 		return zpub, nil, err
 	}
-	var nonce [24]byte
-	if _, err := io.ReadFull(br, nonce[:]); err != nil {
-		return zpub, nil, fmt.Errorf("nonce: %v", err)
-	}
-	msgLen := int(fl - minLen)
+	clientKey = key.NodePublicFromRaw32(mem.B(bs[:]))
```
josharian, line 1124: Again, collapsed ReadFull+FromRaw32 into a ReadFrom.

derp/derp_server.go
```diff
@@ -1481,7 +1472,7 @@ func (c *sclient) sendPacket(srcKey key.Public, contents []byte) (err error) {
 	withKey := !srcKey.IsZero()
 	pktLen := len(contents)
 	if withKey {
-		pktLen += len(srcKey)
+		pktLen += 32
```
josharian, line 1475: srcKey.Len()

derp/derp_server.go
```diff
@@ -1696,31 +1687,18 @@ func (s *Server) ConsistencyCheck() error {
 // It is ~4x slower than io.ReadFull(br, key),
 // but it prevents key from escaping and thus being allocated.
```
josharian, line 1688: Have you confirmed that key.ReadFrom prevents key from escaping? (Is this property protected by an allocs-per-run test, and if not, will you add one? Thanks.)

derp/derp_server.go
```diff
@@ -1696,31 +1687,18 @@ func (s *Server) ConsistencyCheck() error {
...
-	}
-	return nil
+func readPublicKey(br *bufio.Reader, key *key.NodePublic) error {
+	_, err := key.ReadFrom(br)
+	return err
 }
 
 // writePublicKey writes key to bw.
 // It is ~3x slower than bw.Write(key[:]),
 // but it prevents key from escaping and thus being allocated.
```
josharian, line 1697: Same questions.

tstest/integration/integration_test.go
```diff
@@ -543,7 +543,17 @@ func (n *testNode) AwaitResponding(t testing.TB) {
 	t.Helper()
 	n.AwaitListening(t)
 
-	st := n.MustStatus(t)
+	var st *ipnstate.Status
```
josharian, line 546: This seems unrelated. Revert? Or at least put in a separate commit?

types/netmap/netmap_test.go
```diff
@@ -6,17 +6,22 @@ package netmap
...
 	"inet.af/netaddr"
 	"tailscale.com/tailcfg"
+	"tailscale.com/types/key"
 )
 
-func testNodeKey(b byte) (ret tailcfg.NodeKey) {
-	for i := range ret {
-		ret[i] = b
+func testNodeKey(b byte) key.NodePublic {
+	k, err := key.ParseNodePublicUntyped(mem.S(strings.Repeat(fmt.Sprintf("%02x", b), 32)))
```
josharian, line 20: This is a long line, but what it accomplishes is very simple. Add a short doc comment or example to the function so the reader doesn't have to puzzle it out?

wgengine/magicsock/magicsock_test.go
```diff
@@ -1329,11 +1323,11 @@ func TestSetNetworkMapChangingNodeKey(t *testing.T) {
 	var buf tstest.MemLogger
 	conn.logf = buf.Logf
 
-	conn.SetPrivateKey(wgkey.Private{0: 1})
+	conn.SetPrivateKey(key.NewNode())
 
 	discoKey := tailcfg.DiscoKey{31: 1}
-	nodeKey1 := tailcfg.NodeKey{0: 'N', 1: 'K', 2: '1'}
-	nodeKey2 := tailcfg.NodeKey{0: 'N', 1: 'K', 2: '2'}
+	nodeKey1, _ := key.ParseNodePublicUntyped(mem.S("1000000000000000000000000000000000000000000000000000000000000000"))
```
josharian, line 1329: This is the second or third time I've seen something like this. I wonder whether we should have an API to get a "memorable"/"readable"'/"numbered" key. The API could include the word "Test" in it. Or even require you to pass in a testing.TB, which is hard to get ahold of when not in a test.

josharian (approved):
Thanks for doing this.

derp/derp_server.go
danderson: Also done, per above.

wgengine/magicsock/magicsock_test.go
danderson: Will do separate from the big refactor PR that has constant rebase conflicts: https://github.com/tailscale/tailscale/issues/2889

types/key/node.go
danderson: Ended up rewriting this to a different, much more specific func while processing other feedback.

types/key/node.go
danderson: Now implements something much more specific and alloc-free. And documented.

derp/derp_client.go
danderson: I keep going back and forth about this, because this length is only relevant for one particular serialization, and it's not even the most common serialization in our codebase. In my happy place, nobody would need to care how big keys are.

As a compromise to get rid of magic numbers, I added `RawLen` to NodePublic, and renamed ReadRaw32... and WriteRaw32... to just ReadRaw... and WriteRaw..., to try and convey that this is one family of functions that deals with one particular serialization of NodePublic. WDYT?

derp/derp_client.go
danderson: Resolving to keep conversation about API in prior comment. Resolution here is same as abover.

derp/derp_server.go
danderson: I was doing a mechanical change and didn't notice this pattern. Fixed, but there are still some FromRaw32 calls in derp_client that aren't as obviously deletable yet. Closer to a tight API though...

tstest/integration/integration_test.go
danderson: Oops, this was a change I thought I needed during earlier debugging, but turned out that no.

derp/derp_client.go
josharian: SGTM

PR #2885, opened 2021-09-20
version: fix CmdName on the tailscale-ipn.exe binary

bradfitz:
Don't return "wg64", "wg32", etc.

### DIFF
--- a/version/modinfo_test.go
+++ b/version/modinfo_test.go
@@ -5,6 +5,7 @@
 package version
 
 import (
+	"flag"
 	"os/exec"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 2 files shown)


DentonGentry (approved):

version/modinfo_test.go
```diff
@@ -36,3 +37,18 @@ func exe() string {
...
+	exe := *findModuleInfoName
+	if exe == "" {
+		t.Skip("skipping without --module-info-file filename")
+	}
+	cmd := cmdName(exe)
+	mod, err := findModuleInfo(exe)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("Got %q from: %s", cmd, mod)
```
crawshaw, line 53: should there be a t.Error in here somewhere?

crawshaw (approved):
Doesn't seem wrong.

PR #2891, opened 2021-09-21
Set perms on store directory under C:\ProgramData

dblohm7:


bradfitz:
An attempt was made! :)

Issue #2856, opened 2021-09-15
windows: move server state file out of %LocalAppData%


bradfitz:
On Windows, Tailscale uses `%LocalAppData%\Tailscale` as its storage directory for various things. The meaning of %LocalAppData% varies paths depending on whether it's running as a regular user or as the Local System account.

tailscaled.exe, the system service, runs as the Local System and thus its `%LocalAppData%` is `C:\Windows\System32\config\systemprofile\AppData\Local`. Under there it stores:

* incoming Taildrop files (`%LocalAppData%\Tailscale\files`)
* Logs (`%LocalAppData%\Tailscale\Logs`)
* server-state.conf (`%LocalAppData%\Tailscale\server-state.conf`), which is a JSON file containing:
  - the machine key
  - whether Tailscale should run in "unattended mode" and if so, the node key/etc for that

Note how that resides under `C:\Windows\System32`.

We've gotten a number of reports (#2282, ...) that Windows upgrades and other system checkpoint/clean/restore/etc tools nuke the Windows\System32 %LocalAppData% directory, which means Tailscale loses its machine key (and other lesser things).

The per-user node keys (stored in user-specific `%LocalAppData%`) are not wiped, but the control plane rejects them if they're not associated with the right machine key. When Tailscaled starts after having lost its machine key `server-state.conf` file, it assumes it's being run for the first time and makes a new one. But then the control plane is unhappy that the keys don't line up, and Tailscale is broken until the user logs out and logs back in (IIUC), creating a new node key.

Tracking bug for stopping using the system `%LocalAppData%` directory.

We may also want to move Taildrop and Logs?

/cc @danderson @DentonGentry @dblohm7

bradfitz:
Tailscale 1.14.4 is out with these fixes.

bradfitz:
And https://tailscale.com/blog/windows-updates/ is published, asking users to upgrade.

PR #2893, opened 2021-09-21
ipn, paths: Ensure that the state directory for Windows has the corre…

dblohm7:
…ct perms

ProgramData has a permissive ACL. For us to safely store machine-wide
state information, we must set a more restrictive ACL on our state directory.
We set the ACL so that only talescaled's user (ie, LocalSystem) and the
Administrators group may access our directory.

We must include Administrators to ensure that logs continue to be easily
accessible; omitting that group would force users to use special tools to
log in interactively as LocalSystem, which is not ideal.

(Note that the ACL we apply matches the ACL that was used for LocalSystem's
AppData\Local).

There are two cases where we need to reset perms: One is during migration
from the old location to the new. The second case is for clean installations
where we are creating the file store for the first time.

Updates #2856


ipn/store.go
```diff
@@ -182,7 +184,12 @@ func NewFileStore(path string) (*FileStore, error) {
 		if os.IsNotExist(err) {
 			// Write out an initial file, to verify that we can write
 			// to the path.
-			os.MkdirAll(filepath.Dir(path), 0755) // best effort
+			dirPath := filepath.Dir(path)
+			os.MkdirAll(dirPath, 0755) // best effort
+			if runtime.GOOS == "windows" {
```
bradfitz, line 189: we shouldn't need this check if you already have a !windows implementation in that paths package.

How about we collapse these 6 lines down to:

```go
    if err := paths.MkStateDir(filepath.Dir(path)); err != nil {
           return nil, fmt.Errorf("creating state directory: %w", err)
    }
```

And remove the `best effort` comment.

Then document that `MkStateDir` both creates it if needed (with the right permissions to start, so no 0755 to 0700 window on Unix too) and fixes its permissions if needed.

paths/paths_windows.go
```diff
@@ -0,0 +1,140 @@
...
+func getTokenPrimaryGroupInfo(token windows.Token) (*windows.Tokenprimarygroup, error) {
+	buf, err := getTokenInfo(token, windows.TokenPrimaryGroup)
+	if err != nil {
+		return nil, err
+	}
+
+	return (*windows.Tokenprimarygroup)(unsafe.Pointer(&buf[0])), nil
+}
+
+type UserSids struct {
```
bradfitz, line 45: I don't think we want this exported, do make it lowercase. Otherwise it's part of the public API.

paths/paths_windows.go
```diff
@@ -0,0 +1,140 @@
...
+
+	dacl, err := windows.ACLFromEntries(explicitAccess, nil)
+	if err != nil {
+		return err
+	}
+
+	// We now reset the file's owner, primary group, and DACL.
+	// We also must pass PROTECTED_DACL_SECURITY_INFORMATION so that our new ACL
+	// does not inherit any ACL entries from the parent directory.
+	const setFlags windows.SECURITY_INFORMATION = windows.OWNER_SECURITY_INFORMATION |
```
bradfitz, line 134: the type here is unnecessary. And given the small scope this name lives (one line), I'd just name it "flags", so:

```
   const flags = windows.OWNER_SECURITY_INFORMATION | ....
```

paths/paths_windows.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// We need the info for our current user as SIDs
+	sids, err := getCurrentUserSids()
+	if err != nil {
+		return err
+	}
+
+	// We also need the SID for the Administrators group so that admins may
+	// easily access logs.
+	administratorsGroupSid, err := windows.CreateWellKnownSid(windows.WinBuiltinAdministratorsSid)
```
bradfitz, line 93: maybe just `adminGroupSid` ?

bradfitz:
Also, rebase & squash the commits down into 1 commit.

paths/paths_windows.go
dblohm7: Oops! Good catch!

paths/paths_windows.go
dblohm7: Sure.

paths/paths_windows.go
dblohm7: Acknowledged.

ipn/store.go
dblohm7: Ok, will do.

paths/paths_windows.go
dblohm7: Actually this is giving me problems. If I do:
```
flags := windows.OWNER_SECURITY_INFORMATON | ...
```
Then the compiler complains about `flags`'s type when I pass it into `SetNamedSecurityInfo`.
Wrapping that with a conversion:
```
windows.SetNamedSecurityInfo(..., windows.SECURITY_INFORMATION(flags), ...)
```
Makes the compiler complain that `constant 2147483655 overflows int`

Unfortunately from what I can tell, this is because the constants in the `windows` package are not explicitly typed as `windows.SECURITY_INFORMATION` or `uint32`. `windows.PROTECTED_DACL_SECURITY_INFORMATION` is `0x80000000`.

Thoughts @bradfitz?

bradfitz: > If I do:
> flags := windows.OWNER_SECURITY_INFORMATON | ...

I said `const flags = ` ... :) 

See https://go.dev/blog/constants ... they're one of the coolest parts of Go.

paths/paths.go
```diff
@@ -59,3 +59,34 @@ func DefaultTailscaledStateFile() string {
...
+// permisions:
+// If it creates the directory, then the directory is created with the correct permissions.
+// If the directory already exists, we ensure that the permissions are set
+// correctly.
+func MkStateDir(dirPath string) error {
+	info, err := os.Stat(dirPath)
+	if err == nil && !info.IsDir() {
+		// dirPath already exists but is not a directory
+		return os.ErrInvalid
+	} else {
```
bradfitz, line 73: Go style is to do early returns and outdent the rest, without much nesting of elses.

paths/paths.go
```diff
@@ -59,3 +59,34 @@ func DefaultTailscaledStateFile() string {
...
+			return err
+		}
+
+		// The directory does not exist yet. Create it.
+		err = os.MkdirAll(dirPath, 0700)
+
+		// On non-Windows we're done and can just return.
+		// On Windows, when the MkdirAll succeeds, we still need to call
+		// ensureStateDirPerms to set the correct ACL.
+		if runtime.GOOS != "windows" || err != nil {
```
bradfitz, line 84: I'd drop the runtime.GOOS check here and just unconditionally call ensureStateDirPerms.

If you want to avoid the `os.Chmod` call in the common case (already 0700), then pass in `info` (more commonly spelled `fi` for `FileInfo`) to `ensureStateDirPerms`.

Then this is just:

```go
      if err := os.MkdirAll(dirPath, 0700); err != nil {
            return err
      }
      return ensureStateDirPerms(dirPath, fi)
```
 
(and a lot of the comments can go away or be simplified)

And then it handles the case where due to earlier bugs or people messing around the `dirPath` was 0755.

paths/paths_windows.go
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
```
bradfitz, line 1: 2021

paths/paths_windows.go
```diff
@@ -0,0 +1,140 @@
...
+	return (*windows.Tokenprimarygroup)(unsafe.Pointer(&buf[0])), nil
+}
+
+type userSids struct {
+	User         *windows.SID
+	PrimaryGroup *windows.SID
+}
+
+func getCurrentUserSids() (*userSids, error) {
+	token := windows.GetCurrentProcessToken()
```
bradfitz, line 51: I think you'll want OpenCurrentProcessToken because we still support Windows 7 and GetCurrentProcessToken is only supported from Windows 8+.

paths/paths_windows.go
dblohm7: Ah, there we go. Thanks.

paths/paths.go
dblohm7: Since `os.MkdirAll` shouls return an error if `dirPath` is not a directory, I eliminated the `os.Stat`. So we just go ahead and do that, followed by the perm fixup.

paths/paths.go
```diff
@@ -59,3 +59,13 @@ func DefaultTailscaledStateFile() string {
 	}
 	return ""
 }
+
+// MkStateDir ensures that the directory dirPath both exists and has the correct
```
bradfitz, line 63: elaborate on what the "correct permissions" are?

And maybe what "StateDir" is? (that it's meant for the system daemon's configuration directory containing the machine keys, etc)

bradfitz:
Don't worry about the "go generate" failure. That's due to some prod experiment going on.

Lowercase verb "ensure" in the commit message.

bradfitz:
Otherwise LGTM

PR #2905, opened 2021-09-22
net/socks5: verify that the requested port is a uint16.

danderson:


bradfitz:
But why? This can't happen?

danderson:
My mistake, I parsed what the code was doing wrong. wontfix.

Issue #2264, opened 2021-06-28
Support Subnet Routes In userspace-networking Socks5 Proxy


alex-richman-onesignal:
Currently when running in the userspace-networking mode the Socks5 server only supports proxying connections to Tailscale IPs (and external IPs, see https://github.com/tailscale/tailscale/issues/1617).  The server should be able to proxy to subnet routes advertised to the tailscale network by other nodes.

There's a comment in the codebase to this effect already:
https://github.com/tailscale/tailscale/blob/088b90b15f4a3b18b0938f04b9c08e979f9381bb/cmd/tailscaled/tailscaled.go#L254

<img src="https://frontapp.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_3b85t)

jakedgy:
This could be a workaround for Tailscale devices that accept routes within an advertised subnet (https://github.com/tailscale/tailscale/issues/2590).

You pick up a bit more control in that you get to decide if network traffic is even sent to the proxy.

PR #2907, opened 2021-09-23
Support subnet routes in userspace networking mode

jakedgy:
Adds support for accepted subnet routes when running in userspace networking mode with a SOCKS5 proxy.

Fixes https://github.com/tailscale/tailscale/issues/2264.

This does not check the RouteAll pref, so the SOCKS proxy will use the Netstack dialer for all advertised routes regardless of prefs.  I could not find a straight-forward way of determining the state of the RouteAll flag within the SOCKS proxy.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -1781,6 +1781,7 @@ func (b *LocalBackend) authReconfig() {
 	if err == wgengine.ErrNoChanges {
 		return
 	}
+	b.e.NotifyPrefs(prefs)
 	b.logf("[v1] authReconfig: ra=%v dns=%v 0x%02x: %v", prefs.RouteAll, prefs.CorpDNS, flags, err)
 
 	b.initPeerAPIListener()
--- a/net/socks5/tssocks/tssocks.go
+++ b/net/socks5/tssocks/tssocks.go
@@ -94,8 +94,6 @@ func (d *dialer) useNetstackForIP(ip netaddr.IP) bool {
 	if d.ns == nil {
 		return false
 	}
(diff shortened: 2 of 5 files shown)


jakedgy:
Is there an easy way to get read access to Prefs.RouteAll without leaking details about the Engine/Backend?

Maybe with a Handle to the LocalBackend?

ipn/ipnlocal/local.go
```diff
@@ -1781,6 +1781,7 @@ func (b *LocalBackend) authReconfig() {
 	if err == wgengine.ErrNoChanges {
 		return
 	}
+	b.e.NotifyPrefs(prefs)
```
jakedgy, line 1784: I added an engine callback for prefs updates so that tssocks could be more intelligent about whether or not subnet traffic should go over the proxy dialer.

This works well as I've tested it, but I'm not sure if this is the best place to invoke this callback--or if it's appropriate to expose these prefs to the engine at all.

jakedgy: My biggest problem is that the wgcfg.Config.Peers AllowedIPs is different than the netmap.Peers AllowedIPs depending on the value of Prefs.RouteAll.

bradfitz:
@jakedgy, I'm probably the right person to review this but I'm in the middle of a couple fires so probably won't be able to get to it for a few days.

jakedgy:
> @jakedgy, I'm probably the right person to review this but I'm in the middle of a couple fires so probably won't be able to get to it for a few days.

No worries at all.  I appreciate you taking the time to review it whenever you are available.

jakedgy:
Closing this because #2264 fixes it.

PR #2909, opened 2021-09-23
cmd/tailscaled: appease a security scanner

bradfitz:
There are two reasons this can't ever go to actual logs,
but rewrite it to make it happy.

Fixes tailscale/corp#2695


danderson (approved):
CodeQL isn't convinced, but I feel good marking as a false positive.

PR #2916, opened 2021-09-23
ipn/localapi, cmd/tailscale: add CPU & memory profile support, debug command

bradfitz:
This was already possible on Linux if you ran tailscaled with --debug
(which runs net/http/pprof), but it requires the user have the Go
toolchain around.

Also, it wasn't possible on macOS, as there's no way to run the IPNExtension
with a debug server (it doesn't run tailscaled).

And on Windows it's super tedious: beyond what users want to do or
what we want to explain.

Instead, put it in "tailscale debug" so it works and works the same on
all platforms. Then we can ask users to run it when we're debugging something
and they can email us the output files.


client/tailscale/tailscale.go
```diff
@@ -154,6 +154,18 @@ func Goroutines(ctx context.Context) ([]byte, error) {
...
 }
 
+// Profile returns a pprof profile of the Tailscale daemon.
+func Profile(ctx context.Context, pprofType string, sec int) ([]byte, error) {
+	var secArg string
+	if sec < 0 || sec > 300 {
+		return nil, errors.New("duration out of range")
+	}
+	if sec != 0 || pprofType == "profile" {
+		secArg = fmt.Sprint(sec)
```
dsnet, line 164: Shouldn't this be:
```go
secArg = fmt.Sprintf("&seconds=%v", sec)
```
and remove the "&seconds=" from the query string below?

ipn/localapi/profile.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func init() {
+	serveProfileFunc = serveProfile
+}
+
+func serveProfile(w http.ResponseWriter, r *http.Request) {
+	name := r.FormValue("name")
+	if name == "profile" {
+		// CPU is space
```
dsnet, line 25: I don't understand this comment.

dsnet (approved):

client/tailscale/tailscale.go
bradfitz: You'd think, right? But the non-profile handlers treat non-empty string seconds as being a delta:

https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/net/http/pprof/pprof.go;l=238

And then you hit:

https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/net/http/pprof/pprof.go;l=259;drc=refs%2Ftags%2Fgo1.17.1

ipn/localapi/profile.go
bradfitz: Neither do I! Maybe I meant to write "special". But removed it and put it in a switch instead which makes that obvious.

josharian (approved):

cmd/tailscale/cli/debug.go
```diff
@@ -36,6 +36,9 @@ var debugCmd = &ffcli.Command{
 		fs.BoolVar(&debugArgs.netMap, "netmap", true, "whether to include netmap in --ipn mode")
 		fs.BoolVar(&debugArgs.localCreds, "local-creds", false, "print how to connect to local tailscaled")
 		fs.StringVar(&debugArgs.file, "file", "", "get, delete:NAME, or NAME")
+		fs.StringVar(&debugArgs.cpuFile, "cpu-profile", "", "if non-empty, grab a CPU profile for --profile-sec seconds and write it to this file")
```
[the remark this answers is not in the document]
bradfitz: I considered that first, but that's only consistent for us. It's not locally consistent with the other flags in this package.

cmd/tailscale/cli/debug.go
```diff
@@ -36,6 +36,9 @@ var debugCmd = &ffcli.Command{
...
+		fs.StringVar(&debugArgs.memFile, "mem-profile", "", "if non-empty, grab a memory profile and write it to this file")
+		fs.IntVar(&debugArgs.cpuSec, "profile-seconds", 15, "number of seconds to run a CPU profile for, when --cpu-profile is non-empty")
```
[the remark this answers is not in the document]
bradfitz: I really didn't want to deal with 5ms rounding to 0 seconds, etc. I imagine nobody will actually set this flag?

PR #2924, opened 2021-09-24
ipn: test TestFileStore in a fresh subdirectory

### LINKED ISSUE #2923: Move test state directories into a subdir of /tmp
We cannot currently return a proper `error` from `paths/ensureStateDirPerms` because some tests are using `/tmp` as the state directory. We go to `chmod` `/tmp`, that obviously fails, etc.

I'd like to tweak a few paths that we set up in our harnesses so that the state info is dumped into a subdirectory of `/tmp`, thus enabling successful `chmod`s and allowing me to fix that function to return errors.

bradfitz:
Fixes #2923

### DIFF
--- a/ipn/store_test.go
+++ b/ipn/store_test.go
@@ -5,8 +5,7 @@
 package ipn
 
 import (
-	"io/ioutil"
-	"os"
+	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


dblohm7 (approved):
LGTM

PR #2926, opened 2021-09-24
ipn, paths: Unconditionally attempt to set state dir perms, but only …

### LINKED ISSUE #2925: Relax chmod of server state directory
As of 9ebb5d4205d50d54c1874b404b0a7d83cdaa77ca we now chmod the parent directory of the server-state.conf file to 0700.

But that means if a user runs:

```
$ tailscaled --tun=userspace-networking --state=$HOME/foo.state
```

... now we're doing chmod 0700 on their $HOME.

I think we should change that before release, to only chmod the parent if it's a directory we would've otherwise created. (easy heuristic: is the parent base named "Tailscale" or "tailscale" ?)

/cc @DentonGentry @dblohm7 @danderson

dblohm7:
…if the state dir is ours

We unconditionally set appropriate perms on the statefile dir.

We look at the basename of the statefile dir, and if it is "tailscale", then
we set perms as appropriate.

Fixes #2925
Updates #2856

### DIFF
--- a/ipn/store.go
+++ b/ipn/store.go
@@ -170,6 +170,11 @@ func (s *FileStore) String() string { return fmt.Sprintf("FileStore(%q)", s.path
 
 // NewFileStore returns a new file store that persists to path.
 func NewFileStore(path string) (*FileStore, error) {
+	// We unconditionally call this to ensure that our perms are correct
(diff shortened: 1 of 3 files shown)


bradfitz (approved):
lowercase verb after colon in commit message title

Issue #2934, opened 2021-09-26
Unable to use tailscaled on read-only filesystem


mdevaev:
The current release broke run on a read-only filesystem on embedded Linux. Previous releases (like 1.12.0) worked fine, but I can't say from which version it broke.

Previously, it was possible to use the following algorithm:
1. Remount the file system to RW.
2. Run the tailscaled daemon.
3. Perform tailscaled up
4. Reboot.
5. After a reboot, the entire root file system becomes read-only (this is a feature of our distribution), but tailscale worked fine.

The new version has completely broken this use case.
1. At startup, it tries to make a [`chmod /var/lib/tailscale`](https://github.com/tailscale/tailscale/blob/21e9f98fc12d767c2408195da19d9893f9de9224/paths/paths_unix.go#L65) and if this did not work, it refuses to continue working.
2. The log1 and log2 files are also written in /var/lib, but there is no way to place them somewhere else in the directory. Could you please move them to /var/log or provide an option that does this - this location is usually mounted in tmpfs on embedded Linux systems. Logs in /var/lib also violate the FHS standard.

At the moment, tailscale is actively used by [PiKVM](https://github.com/pikvm/pikvm/blob/master/pages/tailscale.md) as a recommended method of cloud access. Updating to a new version breaks access to user's devices.

srepac:
version 1.14.0-1 tailscale doesn't have the issue reported here.

Here's the md5sum and file info for the previous working tailscale package for reference:
```
# md5sum tailscale-1.14.0-1-armv7h.pkg.tar.xz
8bd53995127f74f34a27a632c8097dd0  tailscale-1.14.0-1-armv7h.pkg.tar.xz

# ls -l tailscale-1.14.0-1-armv7h.pkg.tar.xz
-rw-r--r-- 1 root root 4803728 Aug 24 13:06 tailscale-1.14.0-1-armv7h.pkg.tar.xz
```
I hope this helps to resolve the issue.

srepac:
Here's proof that tailscale down and up works with readonly FS of pikvm using the 1.14.0 version

```[root@usbkvm ~]# mount | grep mmc
/dev/mmcblk0p2 on / type ext4 (rw,relatime)
/dev/mmcblk0p3 on /var/lib/kvmd/msd type ext4 (ro,nosuid,nodev,noexec,relatime,nodioread_nolock,nodelalloc,errors=remount-ro,data=journal)
[...]

PR #2935, opened 2021-09-26
paths: skip unix chmod if state directory is already 0700

bradfitz:
Updates #2934

### DIFF
--- a/paths/paths_unix.go
+++ b/paths/paths_unix.go
@@ -8,6 +8,7 @@
 package paths
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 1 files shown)


paths/paths_unix.go
```diff
@@ -62,10 +63,21 @@ func xdgDataHome() string {
...
+	}
+	if !fi.IsDir() {
+		return fmt.Errorf("expected %v to be a directory; is %v", dir, fi.Mode())
+	}
+	const perm = 0700
+	if fi.Mode().Perm() == perm {
+		// Already correct.
+		return nil
+	}
+	return os.Chmod(dir, perm)
```
DentonGentry, line 82: This would still be a behavior change compared with previous versions: tailscaled will refuse to start, especially problematic if the node is remote.

I understand that merely logging it doesn't do much as nobody looks at the logs, but would like to at least consider the tradeoff of exit vs continue.

bradfitz: Until we have a comprehensive integration tests in a read-only builder, read-only disks are not a supported configuration. But this gives people a path forward: they can fix their permissions once manually.

paths/paths_unix.go
```diff
@@ -62,10 +63,21 @@ func xdgDataHome() string {
...
 		return nil
 	}
-
-	return os.Chmod(dirPath, 0700)
+	fi, err := os.Stat(dir)
+	if err != nil {
+		return err
+	}
+	if !fi.IsDir() {
+		return fmt.Errorf("expected %v to be a directory; is %v", dir, fi.Mode())
```
odeke-em, line 75: Did you want to perhaps quote the directory name with %q?

paths/paths_unix.go
bradfitz: On the bug they say that this fix works for them.

paths/paths_unix.go
bradfitz: Done.

paths/paths_unix.go
bradfitz: Okay to submit, @DentonGentry?

dblohm7 (approved):
ok

Issue #2442, opened 2021-07-16
High battery usage on mobile platforms


gillescoolen:
When running the Android 12 beta on a Pixel 5, I am getting a noticeable decrease in battery life when using Tailscale.

![Screenshot_20210716-213255.png](https://user-images.githubusercontent.com/31668393/125999895-49d8f902-e6f2-48dc-8956-7aa67f6ae011.png)

<img src="https://frontapp.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_3f6a9)

DentonGentry:
For Tailscale to accumulate 22 hours of runtime, it is either stuck in a loop or there is a pretty constant stream of traffic going over the VPN.

Could you send either the email address associated with the account or the Tailscale IP address of this phone to support@tailscale.com, so we can look into those?

For comparison, this is what Android 12 on a Pixel 3a is showing:
![Screenshot_20210716-123743](https://user-images.githubusercontent.com/1140756/126000448-7b69f88a-1e04-4561-a15b-e641fc7fc314.png)

gillescoolen:
Will do!

DentonGentry:
There is a subnet router on your account which the Android phone is communicating with often enough that it keeps trying to establish a direct connection but never manages to so. It is able to communicate using the DERP relay, but I think the main part of the power consumption is in waking the radio up so often trying to make a direct connection.

You can likely see this on the subnet router system in `journalctl -u tailscaled`, grep for "magicsock: disco:"
The output of `tailscale status` on the subnet router will likely show something like:

$ tailscale status
...
100.x.y.z  phone        user@   android active; relay "lhr", tx 1108 rx 188

As for what to do about it, the most straightforward thing would be to get them able to make a direct connection so they'll stop having to try so hard. This could include:
- whatever the firewall is from your ISP that the subnet router is sitting behind, if you can enable NAT-PMP on it then the subnet router will have one more tool it can use.
- with the 1.12 release in a couple weeks we'll add UPnP as well, so if that firewall does not have NAT-PMP but does have UPnP that could help.
- open UDP port 41641 in the ISP firewall, allowing UDP port 41641 packets in to at least the subnet router.

gillescoolen:
I see, I'll move the subnet router to another host which has the port open. Thanks for taking your time to figure it out!

buxel:
[...]

PR #2951, opened 2021-09-28
all: disable TCP keep-alives on iOS/Android

bradfitz:
For mobile battery life.

Updates #2442
Updates tailscale/corp#2750


bradfitz:
For testing.

net/netknob/netknob.go
```diff
@@ -0,0 +1,26 @@
...
+	switch runtime.GOOS {
+	case "ios", "android":
+		// Disable TCP keep-alives on mobile platforms.
+		// See https://github.com/golang/go/issues/48622.
+		//
+		// TODO(bradfitz): in 1.17.x, try disabling TCP
+		// keep-alives on for all platforms.
+		return -1
+	}
+	return 0
```
josharian, line 25: Looks to me from the rest of the diff that this should be `return 30 * time.Second`.

josharian (approved):

net/netknob/netknob.go
```diff
@@ -0,0 +1,26 @@
...
+package netknob
+
+import (
+	"runtime"
+	"time"
+)
+
+// PlatformTCPKeepAlive returns the default net.Dialer.KeepAlive
+// value for the current runtime.GOOS.
+func PlatformTCPKeepAlive() time.Duration {
```
crawshaw, line 15: I feel like there should be a NetKnob object to pass around, so that one day control can be involved in configuring it (and we can use it for testing). But maybe later?

crawshaw (approved):

net/netknob/netknob.go
bradfitz: Done.

net/netknob/netknob.go
bradfitz: Maybe. :)

PR #2952, opened 2021-09-28
cmd/tailscaled: set StateDirectoryMode=0700 in tailscaled.service

bradfitz:
Updates #2934


crawshaw (approved):
Is this mode just for creation? I.e. it leaves existing directories alone?

bradfitz:
> Is this mode just for creation? I.e. it leaves existing directories alone?

Sounds like it also changes them. But so do we in tailscaled, so doesn't much matter which layer gets there first.

PR #2953, opened 2021-09-28
cmd/tailscale/web: use html/template for Synology token redirects

maisem:

### DIFF
--- a/cmd/tailscale/cli/web.go
+++ b/cmd/tailscale/cli/web.go
@@ -43,11 +43,18 @@ var webCSS string
 //go:embed auth-redirect.html
 var authenticationRedirectHTML string
 
-var tmpl *template.Template
+//go:embed syno-token-redirect.html
+var synoTokenRedirectHTML string
+
+var (
+	webTmpl               *template.Template
+	synoTokenRedirectTmpl *template.Template
+)
 
 func init() {
-	tmpl = template.Must(template.New("web.html").Parse(webHTML))
-	template.Must(tmpl.New("web.css").Parse(webCSS))
+	webTmpl = template.Must(template.New("web.html").Parse(webHTML))
+	template.Must(webTmpl.New("web.css").Parse(webCSS))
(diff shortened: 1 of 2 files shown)


cmd/tailscale/cli/web.go
```diff
@@ -43,11 +43,18 @@ var webCSS string
...
-var tmpl *template.Template
+//go:embed syno-token-redirect.html
+var synoTokenRedirectHTML string
+
+var (
+	webTmpl               *template.Template
+	synoTokenRedirectTmpl *template.Template
+)
 
 func init() {
```
bradfitz, line 54: can we not do this parsing work until the "web" command starts?

it's a little silly for "tailscale status" to do all this work.

cmd/tailscale/cli/web.go
```diff
@@ -246,49 +253,41 @@ func synoAuthn() (string, error) {
...
+func authRedirect(w http.ResponseWriter, r *http.Request) (bool, error) {
 	if distro.Get() == distro.Synology {
 		return synoTokenRedirect(w, r)
 	}
-	return false
+	return false, nil
 }
 
-func synoTokenRedirect(w http.ResponseWriter, r *http.Request) bool {
+func synoTokenRedirect(w http.ResponseWriter, r *http.Request) (bool, error) {
```
bradfitz, line 263: If you're going to return an error, you don't want to accept a `w` in the first place. You want to return the HTML content to write instead, so you know that `w` is still untouched by the time you go to write the 500 in the caller.

cmd/tailscale/cli/web.go
```diff
@@ -246,49 +253,41 @@ func synoAuthn() (string, error) {
...
-req.send(null);
-</script>
-</body></html>
-`
-
 func webHandler(w http.ResponseWriter, r *http.Request) {
-	if authRedirect(w, r) {
+	type mi map[string]interface{}
+	if ok, err := authRedirect(w, r); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
```
bradfitz, line 287: You don't know that `w` is in a good state to write this.

PR #2954, opened 2021-09-28
cmd/tailscaled: add support for running an HTTP proxy

### LINKED ISSUE #2289: Userspace Networking should support HTTP proxy
There are many applications which do not work well with SOCKS5 including unfortunately [Kubernetes](https://github.com/kubernetes/kubernetes/pull/84205).

It would be great if there could be an option to expose a standard HTTP/HTTPS proxy in addition to the SOCKS5 one.

bradfitz:
This adds support for tailscaled to be an HTTP proxy server.
It shares the same backend dialing code as the SOCK5 server, but the
client protocol is HTTP (including CONNECT), rather than SOCKS.

Fixes #2289


crawshaw (approved):
No tests?

bradfitz:
> No tests?

Maybe later, especially if I move it elsewhere or share it from something else. It's so cookie-cutter, though.

PR #2957, opened 2021-09-28
cmd/tailscaled util/winutil: make tailscaled aware of SessionChange e…

### LINKED ISSUE #2956: Add handling for Windows session unlock events to tailscaled
As discussed today, we want the ability to optionally flush the DNS cache when a session is unlocked.

There are two options for obtaining session change events from Windows:
* By Window message.
* By service notification.

Since we can use existing service notifications to handle this, it makes the most sense to handle this in our tailscaled service. As discussed in today's eng meeting, this will be off by default.

dblohm7:
…vents, and when enabled, flush the dns cache upon session unlock

For the service, all we need to do is handle the `svc.SessionChange` command.
Upon receipt of a `windows.WTS_SESSION_UNLOCK` event, we fire off a goroutine to flush the DNS cache.
(Windows expects responses to service requests to be quick, so we don't want to do that synchronously.)

This is gated on an integral registry value named `FlushDnsOnSessionUnlock`,
whose value we obtain during service initialization. We add `winutil.GetRegInteger`
to support this.

(See [this link](https://docs.microsoft.com/en-us/windows/win32/api/winsvc/nc-winsvc-lphandler_function_ex)
for information re: handling `SERVICE_CONTROL_SESSIONCHANGE`.)

Fixes #2956

### DIFF
--- a/cmd/tailscaled/tailscaled_windows.go
+++ b/cmd/tailscaled/tailscaled_windows.go
@@ -34,6 +34,7 @@ import (
 	"tailscale.com/net/dns"
 	"tailscale.com/net/tstun"
 	"tailscale.com/types/logger"
+	"tailscale.com/util/winutil"
 	"tailscale.com/version"
 	"tailscale.com/wf"
 	"tailscale.com/wgengine"
@@ -42,6 +43,8 @@ import (
 )
 
 const serviceName = "Tailscale"
+
+var flushDNSOnSessionUnlock bool
 
 func isWindowsService() bool {
 	v, err := svc.IsWindowsService()
@@ -63,6 +66,7 @@ type ipnService struct {
 func (service *ipnService) Execute(args []string, r <-chan svc.ChangeRequest, changes chan<- svc.Status) (bool, uint32) {
 	changes <- svc.Status{State: svc.StartPending}
 
+	flushDNSOnSessionUnlock = winutil.GetRegInteger("FlushDNSOnSessionUnlock", 0) != 0
 	ctx, cancel := context.WithCancel(context.Background())
 	doneCh := make(chan struct{})
 	go func() {
@@ -81,6 +85,9 @@ func (service *ipnService) Execute(args []string, r <-chan svc.ChangeRequest, ch
 			case svc.Stop:
 				cancel()
 			case svc.Interrogate:
+				changes <- cmd.CurrentStatus
+			case svc.SessionChange:
+				handleSessionChange(cmd)
 				changes <- cmd.CurrentStatus
 			}
 		}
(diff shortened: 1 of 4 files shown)


bradfitz:
maybe send the winutil part separately? at least in its own commit, even if in this same PR?

bradfitz:
commit message 2 title:

```
cmd/tailscaled: conditionally flush Windows DNS cache on SessionChange
```

Notably: it's not all the time ("conditionally"), and it notes which OS this is for.

And no need to say "tailscaled" a second time when the subject is already the part before the colon.

cmd/tailscaled/tailscaled_windows.go
```diff
@@ -43,6 +44,8 @@ import (
 
 const serviceName = "Tailscale"
 
+var flushDnsOnSessionUnlock bool
```
bradfitz, line 47: flushDNSOn ... https://github.com/golang/go/wiki/CodeReviewComments#initialisms

bradfitz:
Fix the build failure with:

```
$ go generate tailscale.com/tstest/integration
$ git commit -a --amend
```

cmd/tailscaled/tailscaled_windows.go
```diff
@@ -63,6 +66,7 @@ type ipnService struct {
 func (service *ipnService) Execute(args []string, r <-chan svc.ChangeRequest, changes chan<- svc.Status) (bool, uint32) {
 	changes <- svc.Status{State: svc.StartPending}
 
+	flushDNSOnSessionUnlock = winutil.GetRegInteger("FlushDNSOnSessionUnlock", 0) != 0
```
bradfitz, line 69: I don't know what the Windows convention on initialisms is. Feel free to use FlushDnsOnSessionUnlock if that matches Windows style for registry keys. Unless we already picked one style in our other registry keys.

I was only commenting on the Go variable name before.

cmd/tailscaled/tailscaled_windows.go
```diff
@@ -264,6 +271,15 @@ func startIPNServer(ctx context.Context, logid string) error {
 	return err
 }
 
+func handleSessionChange(chgRequest svc.ChangeRequest) {
+	if chgRequest.Cmd != svc.SessionChange || chgRequest.EventType != windows.WTS_SESSION_UNLOCK ||
+		!flushDNSOnSessionUnlock {
+		return
+	}
+
```
bradfitz, line 279: should we log here? or does dns.Flush already log? Even if it does, it doesn't log _why_ it's being flushed, so maybe we should log here anyway.

cmd/tailscaled/tailscaled_windows.go
dblohm7: Yeah, there isn't really a hard convention, so I might as well keep it mostly in sync with the Go variable.

cmd/tailscaled/tailscaled_windows.go
dblohm7: I added a log entry.

cmd/tailscaled/tailscaled_windows.go
```diff
@@ -264,6 +271,16 @@ func startIPNServer(ctx context.Context, logid string) error {
...
+	log.Printf("Received WTS_SESSION_UNLOCK event, initiating DNS flush.")
+	go dns.Flush()
```
DentonGentry, line 281: Can we log if the dns.Flush returns an error (which I think can happen, if ipconfig fails to exec)?

I think this means running an anonymous function as the goroutine:
```
go func() {
        err := dns.Flush()
        ...log on error, etc...
        }()
```

dblohm7: Sure, done

bradfitz:
LGTM. That second commit message is a bit more Markdown-heavy than we usually write, but it's fine... not worth changing.

(Usually the commit messages are just plaintext with maybe sometimes backticks for clarity)

PR #2964, opened 2021-09-29
cmd/tailscale: let up --authkey be of form file:/path/to/secret

### LINKED ISSUE #2958: `tailscale up` allows the authkey to be read by the (unix) environment
As far as I can tell, I can't pass the authkey to `tailscale up` via stdin or via a filename. It would be nice to be able to do so, and keep the authkey a bit more secret. (I do acknowledge this is somewhat theoretical, and that one has already lost if this is a practical concern. All the same, I try to follow it when I can.)

(title somewhat carefully worded to not avoid alarmism)

ref: https://smallstep.com/blog/command-line-secrets/

bradfitz:
Fixes #2958


maisem (approved):

colemickens:
Very fast, much speed. Thank you!

PR #2972, opened 2021-10-01
net/dns/resolver: add unsecured Quad9 resolvers

FiloSottile:
DNSSEC is an availability issue, as recently demonstrated by the
Slack issue, with limited security advantage. DoH on the other hand
is a critical security upgrade. This change adds DoH support for the
non-DNSSEC endpoints of Quad9.

https://www.quad9.net/service/service-addresses-and-features#unsec

### DIFF
--- a/net/dns/resolver/forwarder.go
+++ b/net/dns/resolver/forwarder.go
@@ -712,4 +712,10 @@ func init() {
 	addDoH("149.112.112.112", "https://dns.quad9.net/dns-query")
 	addDoH("2620:fe::fe", "https://dns.quad9.net/dns-query")
 	addDoH("2620:fe::fe:9", "https://dns.quad9.net/dns-query")
+	
+	// Quad9 -DNSSEC
+	addDoH("9.9.9.10", "https://dns10.quad9.net/dns-query")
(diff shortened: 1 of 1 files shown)


danderson (approved):

danderson:
gobench error is spurious, unless extending an array by 4 elements made every single one of our benchmarks worse by a statistically significant margin.

bradfitz:
We don't do DNS for these DoH endpoints anyway so this is kinda a no-op change.

FiloSottile:
> We don't do DNS for these DoH endpoints anyway so this is kinda a no-op change.

I'm not sure I understand, I thought these were the DNS IPs that get magically upgraded to DoH? What does not doing DNS mean? If you meant DNSSEC, the resolver does it for you on 9.9.9.9, and doesn't on 9.9.9.10, meaning Slack would have been down on the former, but up on the latter.

bradfitz:
Ah, okay. I meant that we never do a DNS lookup on `"dns10.quad9.net"`. That's why only some DoH providers are in this list: we're only doing the "easy" ones right now that don't need bootstrap DNS because their IPs are all anycast.

I thought you meant something about the `"dns10.quad9.net"` records wasn't signed.

PR #2973, opened 2021-10-01
net/tlsdial: bake in LetsEncrypt's ISRG Root X1 root

### LINKED ISSUE #1609: Bake in at least LetsEncrypt root cert
Enough users have problems with including root CA certs on their machines, it'd probably be useful for us to bake in at least the LetsEncrypt root cert to our binaries so we don't require that dependency.

That is, this should work:

```
# mkdir /tmp/foo
# touch /tmp/foo/bar
# SSL_CERT_FILE=/tmp/foo/bar SSL_CERT_DIR=/tmp/foo tailscaled 
```

Currently that results in lots of stuff like:

```
Received error: TryLogin: fetch control key: Get "https://login.tailscale.com/key": x509: certificate signed by unknown authority
control: authRoutine: backoff: 203 msec
logtail: dialed "log.tailscale.io:443" in 313ms
control: authRoutine: state:authenticating; wantLoggedIn=true
control: direct.TryLogin(token=false, flags=0)
control: doLogin(regen=false, hasUrl=false)
Received error: TryLogin: fetch control key: Get "https://login.tailscale.com/key": x509: certificate signed by unknown authority
control: authRoutine: backoff: 499 msec
control: authRoutine: state:authenticating; wantLoggedIn=true
control: direct.TryLogin(token=false, flags=0)
control: doLogin(regen=false, hasUrl=false)
Received error: TryLogin: fetch control key: Get "https://login.tailscale.com/key": x509: certificate signed by unknown authority
control: authRoutine: backoff: 309 msec
```

(That won't help all on-prem users using different control servers with different/custom CAs, but this isn't meant to address that. This is meant to reduce our support load for the tailscale.com users.)

bradfitz:
We still try the host's x509 roots first, but if that fails (like if
the host is old), we fall back to using LetsEncrypt's root and
retrying with that.

tlsdial was used in the three main places: logs, control, DERP. But it
was missing in dnsfallback. So added it there too, so we can run fine
now on a machine with no DNS config and no root CAs configured.

Also, move SSLKEYLOGFILE support out of DERP. tlsdial is the logical place
for that support.

Fixes #1609

### DIFF
--- a/net/tlsdial/tlsdial.go
+++ b/net/tlsdial/tlsdial.go
@@ -15,8 +15,23 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
+	"log"
+	"os"
+	"strconv"
+	"sync"
+	"sync/atomic"
 	"time"
 )
+
+var counterFallbackOK int32 // atomic
+
+// If SSLKEYLOGFILE is set, it's a file to which we write our TLS private keys
+// in a way that WireShark can read.
+//
+// See https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format
+var sslKeyLogFile = os.Getenv("SSLKEYLOGFILE")
+
+var debug, _ = strconv.ParseBool(os.Getenv("TS_DEBUG_TLS_DIAL"))
 
 // Config returns a tls.Config for connecting to a server.
 // If base is non-nil, it's cloned as the base config before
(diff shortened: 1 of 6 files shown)


net/tlsdial/tlsdial.go
```diff
@@ -30,6 +42,52 @@ func Config(host string, base *tls.Config) *tls.Config {
...
+		conf.KeyLogWriter = f
+	}
+
+	if conf.InsecureSkipVerify {
+		panic("unexpected base.InsecureSkipVerify")
+	}
+	if conf.VerifyConnection != nil {
+		panic("unexpected base.VerifyConnection")
+	}
+	conf.InsecureSkipVerify = true
```
crawshaw, line 60: maybe add to the line `// security provided by VerifyConnection below`

crawshaw (approved):

net/tlsdial/tlsdial.go
bradfitz: Done. I add the same comment below but didn't copy it. 

At some point I should unify that code, but this PR didn't feel like the time for it.

bradfitz: PTAL

crawshaw (approved):
Code looks good, how do we test it?

bradfitz:
> Code looks good, how do we test it?

I added automated tests to this PR, as well as comments in the source about how I tested it by hand. PTAL.

crawshaw (approved):
Very nice!

PR #2978, opened 2021-10-01
cmd/tailscale: make cert subcommand give hints on access denied

bradfitz:
Lot of people have been hitting this.

Now it says:

    $ tailscale cert tsdev.corp.ts.net
    Access denied: cert access denied

    Use 'sudo tailscale cert' or 'tailscale up --operator=$USER' to not require root.

### DIFF
--- a/client/tailscale/tailscale.go
+++ b/client/tailscale/tailscale.go
@@ -77,6 +77,20 @@ type errorJSON struct {
 	Error string
 }
 
+// AccessDeniedError is an error due to permissions.
+type AccessDeniedError struct {
+	err error
+}
+
+func (e *AccessDeniedError) Error() string { return fmt.Sprintf("Access denied: %v", e.err) }
+func (e *AccessDeniedError) Unwrap() error { return e.err }
+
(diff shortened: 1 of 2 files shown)


client/tailscale/tailscale.go
```diff
@@ -87,6 +101,14 @@ func bestError(err error, body []byte) error {
 	return err
 }
 
+func errorMessageFromBody(body []byte) string {
+	var j errorJSON
+	if err := json.Unmarshal(body, &j); err == nil && j.Error != "" {
```
DentonGentry, line 106: if err != nil I think we return `return strings.TrimSpace(string(body))` which is maybe ok, but would print a string containing whatever as coming from Tailscale and telling them to do it.

bradfitz: Yes, it does. That string is "cert access denied" today. From:

```go
func (h *Handler) serveCert(w http.ResponseWriter, r *http.Request) {
        if !h.PermitWrite {
                http.Error(w, "cert access denied", http.StatusForbidden)
                return
        }
```

DentonGentry (approved):

PR #2998, opened 2021-10-05
ipn: remove unused Prefs.OSVersion and Prefs.DeviceModel

bradfitz:
iOS and Android no longer use these. They both now (as of today)
use the hostinfo.SetFoo setters instead.


bradfitz:
Android update was https://github.com/tailscale/tailscale-android/commit/a5bed46c9c123e662fc8f6f4109da8c33e8fdeea

PR #3005, opened 2021-10-06
cmd/tailscaled: set the correct flag for receiving session change events

dblohm7:
This feature wasn't working until I realized that we also need to opt into
the events. MSDN wasn't so generous as to make this easy to deduce.

Updates #2956


bradfitz (approved):

bradfitz:
Commit message title should say Windows somewhere.

PR #3015, opened 2021-10-07
net/dns: fall back to copy+delete if fs.Rename fails.

### LINKED ISSUE #3000: net/dns.directManager should try copying+truncating if moving files fails
In some container runtimes, /etc/resolv.conf is a writable bind-mount from outside the container. This means that /etc/resolv.conf and /etc are on different filesystems, and so `mv` and `rename(2)` don't work. This breaks net/dns.directManager, which tries to backup /etc/resolv.conf by renaming the file, and updates /etc/resolv.conf by writing to a tempfile and atomically renaming it to /etc/resolv.conf.

As a fallback, we could detect that the initial "rename to create backup" fails, and fall back to an alternate mode where we create the backup by copying resolv.conf, and update resolv.conf by truncating and rewriting it in place. That will open a small race where DNS lookups can fail in the window between the truncate and the write, but that should be a very short window, and it's better than non-functioning DNS.

danderson:
Fixes #3000.

(also a couple tiny cleanup commits that are distinct, last commit in the chain has the behavioral change)

### DIFF
--- a/net/dns/direct.go
+++ b/net/dns/direct.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 
 	"inet.af/netaddr"
+	"tailscale.com/types/logger"
 	"tailscale.com/util/dnsname"
 )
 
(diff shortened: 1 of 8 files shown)


danderson:
The new logic does trigger, as indicated by the logs. It fails because I try to delete /etc/resolv.conf, which is also not allowed - guess I need to truncate it to zero instead.

DentonGentry:
I cloned this branch, added Truncate support, and verified that it can now write /etc/resolv.conf to enable MagicDNS in a Codespaces container. https://github.com/tailscale/tailscale/commit/d7d20f44b6a5062be90c57eee43bd48a6d607f74
https://github.com/tailscale/tailscale/commits/dgentry/dns-copy

Issue #3026, opened 2021-10-08
[Bug]: detect certain resolvconf /etc/resolv.conf files as systemd-resolved


DentonGentry:
We're seen /etc/resolv.conf files like these a few times:

```
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "systemd-resolve --status" to see details about the actual nameservers.
```

The file was created by resolvconf, and tailscaled decides to handle it using the resolvconf path, but the only nameserver in the file is systemd-resolved. The systemd-resolved handling in tailscaled allows more options, like only adding 100.100.100.100 for example.com.beta.tailscale.net and letting the system handle everything else.

I think we can detect cases like this and use systemd-resolved. It will mean making [resolvOwner](https://github.com/tailscale/tailscale/blob/1f506d23514a601701a1d54ce85f4ab116bbbf73/net/dns/direct.go#L95) parse further into the file to see if a later line overrides the first match found.

Issue #2428, opened 2021-07-14
Tailscale 1.10.1 is once again clobbering resolv.conf


taisph:
The `resolv.conf` file is once again consistently getting overwritten by tailscale instead of using systemd-resolve. It is also forcing all DNS lookups to go to 100.100.100.100 although "override local DNS" is off in the online settings.

Reverting to 1.8.7 seems to fix the issue.

1.10.1
  tailscale commit: 6b6016130a794dd7332ff79abbb0a63a220d1dec
  other commit: 49e1dcc205c8e95f667a7d66b5f2484282041965
  go version: go1.16.4-tsa2a536c

Ubuntu 20.04.2 LTS

See also #589.

DentonGentry:
I think there are two ways to fix it:
- in https://login.tailscale.com/admin/dns disable Magic DNS and remove all DNS server configuration
- if you want to keep DNS configuration for other devices, you can tell the Ubuntu machine to not apply any DNS config using "tailscale up --allow-dns=false"

taisph:
> I think there are two ways to fix it:

Keeping tailscale at 1.8.7 seems to be a better fix than outright disabling DNS. Without DNS I would have to go back to OpenVPN.

bradfitz:
@taisph, what's your Tailscale IP so we can look at logs?

taisph:
> @taisph, what's your Tailscale IP so we can look at logs?

Sent to support@tailscale.com. Thanks for taking a look.

bradfitz:
```
2021-07-29 20:33:07.019423645 +0200 +0200: Program starting: v1.12.1-t07819c160-g935db9f62, Go 1.16.6-ts6fa85e8: []string{"/usr/sbin/tailscaled", "--state=/var/lib/tailscale/tailscaled.state", "--socket=/run/tailscale/tailscaled.sock", "--port", "41641"}
2021-07-29 20:33:07.019436675 +0200 +0200: LogID: dc59e81377132d75fc344d7037bfac20a816bb9a587bab371d0fde07b471cdc6
2021-07-29 20:33:07.019447965 +0200 +0200: logpolicy: using $STATE_DIRECTORY, "/var/lib/tailscale"
2021-07-29 20:33:07.020406005 +0200 +0200: wgengine.NewUserspaceEngine(tun "tailscale0") ...
2021-07-29 20:33:07.023063253 +0200 +0200: router: v6nat = true
[...]

PR #3027, opened 2021-10-08
net/dns: detect when resolvconf points to systemd-resolved.

DentonGentry:
There are /etc/resolv.conf files out there where resolvconf wrote
the file but pointed to systemd-resolved as the nameserver.
We're better off handling those as systemd-resolved.

```
#  Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "systemd-resolve --status" to see details about the actual nameservers.
```

Fixes https://github.com/tailscale/tailscale/issues/3026


DentonGentry:
Also note https://github.com/tailscale/tailscale/issues/2428 which seems relevant, DBus behavior varies depending on compatibility mode.

PR #3050, opened 2021-10-12
cmd/tailscale/cli: add file ls

mattn:
Current implementation of `tailscale file` command does not have ways to recognize whether the host is receiving files. This change add `tailscale file ls` sub-command. And -json flag will be useful to check the file is too large.

### DIFF
--- a/cmd/tailscale/cli/file.go
+++ b/cmd/tailscale/cli/file.go
@@ -7,6 +7,7 @@ package cli
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"flag"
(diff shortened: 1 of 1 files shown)


mattn:
I want to add feature that auto receiving files sent from others in [tailscale-systray](https://github.com/mattn/tailscale-systray). Currently this can't implement following features.

* Display account e-mail address
* Auto receiving files since `tailscale get` require root permission. Also can't overwrite files if same filename already exists.

So I want to add features to tailscale command.
* Way to get account e-mail address.
* Get filename received from others. (to remove them before writing)

PR #3054, opened 2021-10-12
Bug template: remove "[Bug]: " prefix for the common case.

apenwarr:
Also shorten "[FR]:" to "FR:" to save precious subject line space.

I don't mind a prefix to distinguish feature requests, but the majority
of cases are bugs. Let's preserve as many chars as possible for the
specific topic when looking at subject lines in gmail.

(Now, if only it wouldn't include [tailscale/tailscale] on every
message...)

Signed-off-by: Avery Pennarun <apenwarr@tailscale.com>

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,6 +1,6 @@
 name: Bug report
 description: File a bug report
-title: "[Bug]: "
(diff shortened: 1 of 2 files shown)


apenwarr:
Sounds like no objections to the current change. I can see a bit of value to the FR: prefix so I'll not change it for now. I guess we can see if it gets annoying.

PR #3058, opened 2021-10-12
Bug report template: remove empty 'title' field.

apenwarr:
Mysteriously, GitHub can't parse it if it's an empty string rather than just missing.


DentonGentry:
In related news, I think I found a way to improve our bug metrics.

PR #3061, opened 2021-10-12
wintun 0.14 bump

dblohm7:
Need this to satisfy wintun update in corp. Note that `Pool` [is no longer a thing](https://git.zx2c4.com/wireguard-go/commit/?id=82d2aa87aa623cb5143a41c3345da4fb875ad85d).


dblohm7:
> 
> 
> Can you merge these into one commit (so there's no git bissect window where things are broken) and then include a `go mod tidy` in that same commit?

Done.

PR #3085, opened 2021-10-14
net/dnsfallback: update derp11a record

crawshaw:


dsnet:
Oh gosh. This is unrelated. It's what I would argue is a bug in the "go generate" logic.
It sees the "go:generate" string in the file and treats this as generated code.
Such strings should be ignored if they're within a string literal and not a comment.

dsnet:
Fixed in #3086.

PR #3087, opened 2021-10-14
Bradfitz/exit node fixes

bradfitz:
(Two separate commits.)


bradfitz:
Both the Windows and Windows-race CI failures were different flakes. 😭

PR #3113, opened 2021-10-16
wgengine/magicsock: don't Rebind after STUN error if closed.

DentonGentry:
https://github.com/tailscale/tailscale/pull/3014 added a
rebind on STUN failure, which means there can now be a
tailscale.com/wgengine/magicsock.(*RebindingUDPConn).ReadFromNetaddr
in progress at the end of the test waiting for a STUN
response which will never arrive.

This causes a test flake due to the resource leak in those
cases where the Conn decided to rebind. For whatever reason,
it mostly flakes with Windows.

If the Conn is closed, don't Rebind after a send error.


DentonGentry:
That'll teach me not to re-run the Windows race test after it passed once.

Ok, clearly this doesn't fix it.

DentonGentry:
Windows-race has passed four times in a row with no flakes.

Issue #3088, opened 2021-10-14
wgengine/magicsock: wgdev.Reconfig: unexpected endpoint for peer, expected the peer's public key


bradfitz:
I can `tailscale ping` (disco ping) but not `tailscale ping --tsmp` a Linux 1.16.0 machine.

That Linux 1.16.0 machine's logs say:

```
Oct 14 13:28:16 tsbi tailscaled[7707]: wgdev.Reconfig: unexpected endpoint "84c0cfca85ac0aaf1086c76f7e7ff85cffb8ca98c5c87495da667ce3b09af306" for peer "[/Bifd]", expected the peer's public key
```

/cc @danderson @DentonGentry @crawshaw

bradfitz:
I hacked up testcontrol to send a non-stop stream of changes (randomly adding/deleting/changing node/disco keys) and added some random sleeps in the client and promoted the error to panics, but so far I've been unable to reproduce it.

But a clue...

The error message says:

```
Oct 14 13:28:16 tsbi tailscaled[7707]: wgdev.Reconfig: unexpected endpoint "84c0cfca85ac0aaf1086c76f7e7ff85cffb8ca98c5c87495da667ce3b09af306" for peer "[/Bifd]", expected the peer's public key
```

That comes from:

```
case key.EqualString("endpoint"):
        // TODO: our key types are all over the place, and this
        // particular one can't parse a mem.RO or a []byte without
        // allocating. We don't reconfigure wireguard often though, so
        // this is okay.
        s := value.StringCopy()
        k, err := wgkey.ParseHex(s)
        if err != nil {
                return fmt.Errorf("invalid endpoint %q for peer %q, expected a hex public key", s, peer.PublicKey.ShortString())
        }
        if k != peer.PublicKey {
                return fmt.Errorf("unexpected endpoint %q for peer %q, expected the peer's public key", s, peer.PublicKey.ShortString())
        }
```

So how would they get out of sync?

The clue is `84c0cfca85ac0aaf1086c76f7e7ff85cffb8ca98c5c87495da667ce3b09af306` in short form is `[hMDPy]` which is my Mac desktop on my tailscale.com email account (nodeid:49969443346629937 100.90.243.7)

But `[/Bifd]` is the same physical machine & same Tailscale IPNExtension process, but logged into my gmail account (nodeid:23936249232029309 100.82.88.80).

Notably, they share the same disco key when switching accounts.

[...]

PR #3121, opened 2021-10-17
wgengine/magicsock: track which NodeKey each DiscoKey was last for

bradfitz:
This adds new fields (currently unused) to discoInfo to track what the
last verified (unambiguous) NodeKey a DiscoKey last mapped to, and
when.

Updates #3088


bradfitz:
(Rebased past the unrelated data race fix)

PR #3141, opened 2021-10-19
wgengine/magicsock: turn down timeouts in tests

crawshaw:
Before:

	--- PASS: TestActiveDiscovery (11.78s)
	    --- PASS: TestActiveDiscovery/facing_easy_firewalls (5.89s)
	    --- PASS: TestActiveDiscovery/facing_nats (5.89s)
	    --- PASS: TestActiveDiscovery/simple_internet (0.89s)

After:

	--- PASS: TestActiveDiscovery (1.98s)
	    --- PASS: TestActiveDiscovery/facing_easy_firewalls (0.99s)
	    --- PASS: TestActiveDiscovery/facing_nats (0.99s)
	    --- PASS: TestActiveDiscovery/simple_internet (0.89s)

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -3151,16 +3151,6 @@ const (
 	// are sent.
 	heartbeatInterval = 2 * time.Second
 
-	// discoPingInterval is the minimum time between pings
-	// to an endpoint. (Except in the case of CallMeMaybe frames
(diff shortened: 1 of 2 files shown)


danderson:
I dropped the virtual clock PR after consensus that it would be more invasive than acceptable (in particular to make it work with general I/O timeouts and such). If we want one again, we can revive that effort (esp. now we have our own toolchain :P )

PR #3150, opened 2021-10-20
scripts/installer.sh: Correct support for Oracle Linux.

danderson:
Rebase + tweak of #3146 

Co-Authored-By: Jonathan Hult <jhult@mythics.com>


danderson:
In general I want to try and stick to the "preferred" command for each distro release, and we already have the machinery for yum vs. dnf because of other redhat derivatives, so it's not a big pain to support.

Thanks for testing!

Issue #3157, opened 2021-10-20
Run Tailscale in the browser


bradfitz:
Tracking bug for running Tailscale in the browser with WebAssembly, with the goal of doing end-to-end WireGuard w/ netstack from the browser to a remote peer, doing SSH/VNC/RDP (with e.g. xterm.js, https://github.com/novnc/noVNC, etc), where private keys etc all live in the browser session.

bradfitz:
@mihaip finished this up. It shipped: https://tailscale.com/blog/ssh-console/ Thanks, Mihai!

(Improvements will be tracked in separate issues.)

teaalltr:
What's the status of this? Has it been implemented?

d1vanloon:
@teaalltr see the previous comment. This shipped as part of the web-based Tailscale SSH Console.

PR #3165, opened 2021-10-21
cmd/derper, derp/derphttp: add websocket support

bradfitz:
Updates #3157


bradfitz:
Not sufficiently tested yet, but that'll come.

bradfitz:
It was ready but then I decided to add the client to the same PR to share some code before I submitted it in the wrong place. But then I didn't finish it so marked it as draft.

Issue #3197, opened 2021-10-27
Tailscale fails to detect/use Eero's UPnP


danderson:
### What is the issue?

My network uses an Eero gateway, which supports UPnP. `tailscale netcheck` and node hostinfo both report that no portmapping protocols are available on my LAN.

### Steps to reproduce

Get an eero network, turn on UPnP in network settings, use tailscale.

### OS

Linux, macOS, Windows, Android

### Bug report

BUG-7f282f26c77f1efeb8d494a99187c1d55a4346b814e69cd004df6dda231794b1-20211027204508Z-4920474da648f13d

PR #3202, opened 2021-10-27
net/portmapper: ignore IGD SSDP responses from !defaultgw

danderson:
Now that we multicast the SSDP query, we can get IGD offers from
devices other than the current device's default gateway. We don't want
to accidentally bind ourselves to those.

Updates #3197

### DIFF
--- a/net/portmapper/portmapper.go
+++ b/net/portmapper/portmapper.go
@@ -752,10 +752,14 @@ func (c *Client) Probe(ctx context.Context) (res ProbeResult, err error) {
 			}
 			return res, err
 		}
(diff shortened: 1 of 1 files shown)


joshbenner:
Note to future users: this can cause UPnP and NAT-PMP to not be detected behind OPNSense (and presumably pfSense) in a failover pair, since the routers respond on their interface IP rather than the virtual IP probably used as the gateway. I was able to work around this for tailscale by port forwarding 5153/udp from the virtual gateway IP to the primary router's LAN interface IP. Not perfect (will not work during failover), but gets things working during normal operations.

DentonGentry:
Could you file a bug about this, with a description? The intent of this PR was not to preclude use of portmapping in valid cases.

PR #3223, opened 2021-10-29
logtail/filch: limit buffer file size to 50MB

maisem:
If for some reason there is no network connectivity, the log files keep growing.  This attempts to limit the max size we consume to 50MB per file.

image

### DIFF
--- a/logtail/filch/filch.go
+++ b/logtail/filch/filch.go
@@ -17,8 +17,11 @@ import (
 
 var stderrFD = 2 // a variable for testing
 
+const defaultMaxFileSize = 50 << 20
+
 type Options struct {
 	ReplaceStderr bool // dup over fd 2 so everything written to stderr comes here
+	MaxFileSize   int
 }
 
 // A Filch uses two alternating files as a simplistic ring buffer.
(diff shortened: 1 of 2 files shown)


logtail/filch/filch.go
```diff
@@ -91,6 +93,17 @@ func (f *Filch) scan() ([]byte, error) {
 func (f *Filch) Write(b []byte) (int, error) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
+	fi, err := f.cur.Stat()
```
crawshaw, line 96: it would be nice to avoid the stat on every write. put a log counter in Filch, n and check every 100 writes?

crawshaw (approved):
I think this works. Any chance to write a test with maxFileSize dialed way down?

logtail/filch/filch.go
josharian: Or go stateless and do checks randomly. Or put a size counter in Filch, at which point you don't need a stat ever.

maisem:
> I think this works. Any chance to write a test with maxFileSize dialed way down?

Added a test to verify that we can only read maxFileSize x 2 bytes.

crawshaw (approved):
LGTM, thanks!

Issue #3220, opened 2021-10-29
"wgengine: watchdog timeout on Reconfig" and tailscaled exits


DentonGentry:
### What is the issue?

No particular pattern to it has been noticed, but user reports that tailscaled exits fairly often.
We see a crash report.

```
2021-10-29 12:09:35.96721086 +0000 UTC: wgengine watchdog stacks:
goroutine profile: total 89
22 @ 0x437c36 0x40630c 0x405d78 0x8520a5 0x4681a1
#	0x8520a4	golang.zx2c4.com/wireguard/device.(*Peer).RoutineSequentialReceiver+0x144	golang.zx2c4.com/wireguard@v0.0.0-20210905140043-2ef39d47540c/device/receive.go:406

22 @ 0x437c36 0x40630c 0x405d78 0x85477a 0x4681a1
#	0x854779	golang.zx2c4.com/wireguard/device.(*Peer).RoutineSequentialSender+0x119	golang.zx2c4.com/wireguard@v0.0.0-20210905140043-2ef39d47540c/device/send.go:418

3 @ 0x437c36 0x447a12 0x878dd4 0x4681a1
#	0x878dd3	tailscale.com/wgengine/magicsock.(*Conn).runDerpWriter+0x1d3	tailscale.com@v1.16.1/wgengine/magicsock/magicsock.go:1575

2 @ 0x437c36 0x4305de 0x462a29 0x4d19b2 0x4d2afa 0x4d2ae8 0x57e069 0x58c165 0x5c25fd 0x4a36f8 0x5c27e5 0x5bfe52 0x5c56ef 0x5c56f0 0x543263 0x543a0c 0x6e57be 0x6e75e7 0x704505 0x7044f4 0x877c9d 0x4681a1
#	0x462a28	internal/poll.runtime_pollWait+0x88				runtime/netpoll.go:229
#	0x4d19b1	internal/poll.(*pollDesc).wait+0x31				internal/poll/fd_poll_runtime.go:84
#	0x4d2af9	internal/poll.(*pollDesc).waitRead+0x259			internal/poll/fd_poll_runtime.go:89
#	0x4d2ae7	internal/poll.(*FD).Read+0x247					internal/poll/fd_unix.go:167
#	0x57e068	net.(*netFD).Read+0x28						net/fd_posix.go:56
#	0x58c164	net.(*conn).Read+0x44						net/net.go:183
[...]

PR #3226, opened 2021-10-29
wgengine/wgcfg: always close io.Pipe

josharian:
In DeviceConfig, we did not close r after calling FromUAPI.
If FromUAPI returned early due to an error, then it might
not have read all the data that IpcGetOperation wanted to write.
As a result, IpcGetOperation could hang, as in #3220.

We were also closing the wrong end of the pipe after IpcSetOperation
in ReconfigDevice.

To ensure that we get all available information to diagnose
such a situation, include all errors anytime something goes wrong.

This should fix the immediate crashing problem in #3220.
We'll then need to figure out why IpcGetOperation was failing.


josharian:
Tests pass again. :) PTAL

wgengine/magicsock/magicsock_test.go
```diff
@@ -296,10 +297,19 @@ func meshStacks(logf logger.Logf, mutateNetmap func(idx int, nm *netmap.NetworkM
...
 			if err := m.Reconfig(wg); err != nil {
-				if ctx.Err() != nil || errors.Is(err, errConnClosed) {
+				if ctx.Err() != nil {
 					// shutdown race, don't care.
 					return
 				}
+				var me multierror.MultipleErrors
+				if errors.As(err, &me) {
+					for _, e := range me {
+						if errors.Is(e, errConnClosed) {
```
crawshaw, line 307: woah, does an errors.Is not see into a multierror? that seems wrong. does unwrap make it impossible to implement an iterator?

crawshaw (approved):

crawshaw:
oh it totally can, `errors.Is` looks for a method named `Is`, so a multierror can implement that and do the iteration.

looks like multierror hasn't been touched in 6 years. want to fork it under github.com/tailscale/multierror and implement Is, instead of putting the loop in the test?

josharian:
Now with a shiny new package. PTAL. 

CI failure is a pre-existing flake. Filed https://github.com/tailscale/tailscale/issues/3247.

crawshaw (approved):

PR #3227, opened 2021-10-29
cmd/tailscaled, wgengine/netstack: always wire up netstack

bradfitz:
Even if not in use. We plan to use it for more stuff later.

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -54,13 +54,23 @@ type Impl struct {
 	// port other than accepting it and closing it.
 	ForwardTCPIn func(c net.Conn, port uint16)
 
-	ipstack     *stack.Stack
-	linkEP      *channel.Endpoint
-	tundev      *tstun.Wrapper
-	e           wgengine.Engine
-	mc          *magicsock.Conn
-	logf        logger.Logf
-	onlySubnets bool // whether we only want to handle subnet relaying
+	// ProcessLocalIPs is whether netstack should handle incoming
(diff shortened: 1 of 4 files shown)


wgengine/netstack/netstack.go
```diff
@@ -54,13 +54,20 @@ type Impl struct {
...
-	e           wgengine.Engine
-	mc          *magicsock.Conn
-	logf        logger.Logf
-	onlySubnets bool // whether we only want to handle subnet relaying
+	// ProcessAll is whether netstack should handle all incoming traffic.
+	ProcessAll bool
+
+	// ProcessSubnets is whether netstack should traffic to
+	// non-local IPs (e.g. it should be a subnet router). This
+	// option is ignored ProcessAll is set anyway.
```
DentonGentry, line 62: "This option is ignored ProcessAll is set anyway."

Something got deleted from that sentence.

crawshaw (approved):
The code looks fine but I'd like to know if it has any impact on throughput. Any benchmarks we can run?

bradfitz:
> The code looks fine but I'd like to know if it has any impact on throughput. Any benchmarks we can run?

It should be negligible. A method call that reads a false field and does nothing. Nothing actually goes through netstack when it's off.

wgengine/netstack/netstack.go
bradfitz: Whoops. I'll rewrite. And might rename these fields. They're subtle and confusing.

bradfitz: PTAL; renamed & rewritten.

PR #3235, opened 2021-11-01
scripts/installer: handle fedora

Xe:
We missed a switch case.

### DIFF
--- a/scripts/installer.sh
+++ b/scripts/installer.sh
@@ -250,6 +250,9 @@ main() {
 				OS_UNSUPPORTED=1
 			fi
 			;;
+		fedora)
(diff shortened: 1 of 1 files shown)


scripts/installer.sh
```diff
@@ -250,6 +250,9 @@ main() {
 				OS_UNSUPPORTED=1
 			fi
 			;;
+    fedora)
```
maisem, line 253: White space mismatch

maisem (approved):

danderson (approved):
Arguably we don't need to clutter that switch with all the cases where we have nothing to do, but LGTM for consistency with current practice.

Issue #3233, opened 2021-10-31
netstack: unaligned 64-bit atomic operation on MIPS


DentonGentry:
From https://github.com/tailscale/tailscale/issues/2129#issuecomment-955765999:

Different version, different arch 
```
panic: unaligned 64-bit atomic operation

goroutine 193 [running]:
runtime/internal/atomic.panicUnaligned(...)
        runtime/internal/atomic/unaligned.go:8
runtime/internal/atomic.lockAndCheck(...)
        runtime/internal/atomic/atomic_mipsx.go:38
runtime/internal/atomic.Store64(0x214a56c, 0x8000)
        runtime/internal/atomic/atomic_mipsx.go:104 +0x94
inet.af/netstack/atomicbitops.(*AlignedAtomicInt64).Store(...)
        inet.af/netstack@v0.0.0-20210622165351-29b14ebc044e/atomicbitops/aligned_64bit.go:39
inet.af/netstack/tcpip.(*SocketOptions).SetReceiveBufferSize(0x214a4f4, 0x8000, 0x0)
        inet.af/netstack@v0.0.0-20210622165351-29b14ebc044e/tcpip/socketops.go:663 +0x1c8
inet.af/netstack/tcpip/transport/udp.newEndpoint(0x2098000, 0x800, 0x20daba0)
        inet.af/netstack@v0.0.0-20210622165351-29b14ebc044e/tcpip/transport/udp/endpoint.go:187 +0x270
inet.af/netstack/tcpip/transport/udp.(*ForwarderRequest).CreateEndpoint(0x20dab80, 0x20daba0)
        inet.af/netstack@v0.0.0-20210622165351-29b14ebc044e/tcpip/transport/udp/forwarder.go:79 +0x11c
tailscale.com/wgengine/netstack.(*Impl).acceptUDP(0x20d3b00, 0x20dab80)
        tailscale.com@v1.16.2/wgengine/netstack/netstack.go:597 +0xb4
inet.af/netstack/tcpip/transport/udp.(*Forwarder).HandlePacket(...)
        inet.af/netstack@v0.0.0-20210622165351-29b14ebc044e/tcpip/transport/udp/forwarder.go:47
tailscale.com/wgengine/netstack.(*Impl).wrapProtoHandler.func1({0x35, {0x236fbcc, 0x4}, 0x9cfa, {0x236fbc8, 0x4}}, 0x206b950)
[...]

PR #3236, opened 2021-11-01
tstest/archtest: add GOARCH-specific tests, run via qemu-user

bradfitz:
Updates #3233

### DIFF
--- a/tstest/archtest/qemu_test.go
+++ b/tstest/archtest/qemu_test.go
@@ -1,0 +1,75 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build linux && amd64 && !race
+// +build linux,amd64,!race
+
+package archtest
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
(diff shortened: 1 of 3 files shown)


tstest/archtest/qemu_test.go
```diff
@@ -0,0 +1,75 @@
...
+			cmd.Env = append(os.Environ(), "GOARCH="+arch.Goarch)
+			out, err := cmd.CombinedOutput()
+			if err != nil {
+				if strings.Contains(string(out), "fatal error: sigaction failed") && !inCI {
+					t.Skip("skipping; qemu too old. use 5.x.")
+				}
+				t.Errorf("failed: %s", out)
+			}
+			sub := fmt.Sprintf("I am linux/%s", arch.Goarch)
+			if !bytes.Contains(out, []byte(sub)) {
```
DentonGentry, line 70: I don't really understand why this is here. I see that it is checking that the test actually ran by looking for a message it would have logged, but why is that especially important here?

DentonGentry (approved):

tstest/archtest/qemu_test.go
bradfitz: Just as proof that we made it into our code at all and didn't e.g. os.Exit(0) early or crash in a way that failed without an exit code.

PR #3251, opened 2021-11-03
Support build and release on pushed tags to ghcr

### LINKED ISSUE #3250: FR: Support build and release on pushed tags
### Tell us about your idea!

Build docker containers and push them to ghcr when releasing tag

Should support at least ghcr, linux/amd64 and linux/arm64

chrisgoffinet:
Build linux/amd64 and linux/arm64

Fixes #3250

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -1,0 +1,44 @@
+name: Build and Release
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
+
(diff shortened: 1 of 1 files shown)


DentonGentry:
Thank you, though I suspect we won't do it this way. We have an autobuilder which builds most of the platforms, we're likely to implement container image push alongside those.

chrisgoffinet:
No problem!

PR #3255, opened 2021-11-03
ipn/ipnlocal: add owner-only debug handler to get process env

bradfitz:
For debugging Synology. Like the existing goroutines handler, in that
it's owner-only.

### DIFF
--- a/ipn/ipnlocal/peerapi.go
+++ b/ipn/ipnlocal/peerapi.go
@@ -6,6 +6,7 @@ package ipnlocal
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"hash/crc32"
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/peerapi.go
```diff
@@ -710,3 +716,28 @@ func (h *peerAPIHandler) handleServeGoroutines(w http.ResponseWriter, r *http.Re
...
+
+func (h *peerAPIHandler) handleServeEnv(w http.ResponseWriter, r *http.Request) {
+	if !h.isSelf {
+		http.Error(w, "not owner", http.StatusForbidden)
+		return
+	}
+	var data struct {
+		Hostname     string
+		GOOS, GOARCH string
+		Distro       string
```
maisem, line 728: Use`hostinfo.New()`?

maisem (approved):

ipn/ipnlocal/peerapi.go
bradfitz: Ooooh right, good call.

PR #3231, opened 2021-10-31
cmd/derper custom port support TLS

oocococo:
allow TLS in custom port

PR #3263, opened 2021-11-04
cmd/derper: allow unsecure port configuration

### LINKED ISSUE #3232: FR: DERP support custom port
### Tell us about your idea!

derper allow TLS in custom port, when using manual certmode

443 usage is restricted in some region (e.g. China...

derper allow us to use custom port, however non-443 port won't allow TLS connection
if you try to config a custom port in ACL, relay can't work, and the log complain something like `tls: first record does not look like a TLS handshake`

Can we allow TLS on custom port?

### How should we solve this?

We can assume, if `certmode` is `manual`, it is attend to allow TLS connection on custom port.

So i made PR  [#3231](https://github.com/tailscale/tailscale/pull/3231)

_No response_

### Anything else?

I'm not sure is traffic through DERP is encrypted end-to-end using TLS or Wireguard, if It's is TLS, then, there's no need to allow derper without TLS, isn't it?

arnaud-dezandee:
Following on https://github.com/tailscale/tailscale/pull/3231

This PR aims to allow the configuration of the unsecure HTTP port, which serves both the debug/profiling and acme challenges if enabled.
I added 1 new CLI flags:
- `--http-port`

Fixes #3232

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -12,6 +12,7 @@ import (
 	"errors"
 	"expvar"
 	"flag"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
@@ -36,6 +37,7 @@ import (
 var (
 	dev           = flag.Bool("dev", false, "run in localhost development mode")
 	addr          = flag.String("a", ":443", "server address")
+	httpPort      = flag.Int("http-port", 80, "The port on which to serve HTTP. Set to -1 to disable")
 	configPath    = flag.String("c", "", "config file path")
 	certMode      = flag.String("certmode", "letsencrypt", "mode for getting a cert. possible options: manual, letsencrypt")
 	certDir       = flag.String("certdir", tsweb.DefaultCertDir("derper-certs"), "directory to store LetsEncrypt certs, if addr's port is :443")
@@ -250,24 +252,26 @@ func main() {
 			w.Header().Set("Content-Security-Policy", "default-src 'none'; frame-ancestors 'none'; form-action 'none'; base-uri 'self'; block-all-mixed-content; plugin-types 'none'")
 			mux.ServeHTTP(w, r)
 		})
-		go func() {
-			port80srv := &http.Server{
-				Addr:        net.JoinHostPort(listenHost, "80"),
-				Handler:     certManager.HTTPHandler(tsweb.Port80Handler{Main: mux}),
-				ReadTimeout: 30 * time.Second,
-				// Crank up WriteTimeout a bit more than usually
-				// necessary just so we can do long CPU profiles
-				// and not hit net/http/pprof's "profile
-				// duration exceeds server's WriteTimeout".
-				WriteTimeout: 5 * time.Minute,
-			}
-			err := port80srv.ListenAndServe()
-			if err != nil {
-				if err != http.ErrServerClosed {
-					log.Fatal(err)
+		if *httpPort > -1 {
+			go func() {
+				port80srv := &http.Server{
+					Addr:        net.JoinHostPort(listenHost, fmt.Sprintf("%d", *httpPort)),
+					Handler:     certManager.HTTPHandler(tsweb.Port80Handler{Main: mux}),
+					ReadTimeout: 30 * time.Second,
+					// Crank up WriteTimeout a bit more than usually
+					// necessary just so we can do long CPU profiles
+					// and not hit net/http/pprof's "profile
+					// duration exceeds server's WriteTimeout".
+					WriteTimeout: 5 * time.Minute,
 				}
-			}
-		}()
+				err := port80srv.ListenAndServe()
+				if err != nil {
+					if err != http.ErrServerClosed {
+						log.Fatal(err)
+					}
+				}
+			}()
+		}
 		err = httpsrv.ListenAndServeTLS("", "")
 	} else {
 		log.Printf("derper: serving on %s", *addr)


danderson (approved):
This also means the /debug/ handler won't be available when running self-hosted, but I assume that's fine.

bradfitz:
The premise is wrong: we don't use port 80 for acme challenges.

cmd/derper/derper.go
```diff
@@ -235,24 +235,26 @@ func main() {
...
-				// necessary just so we can do long CPU profiles
-				// and not hit net/http/pprof's "profile
-				// duration exceeds server's WriteTimeout".
-				WriteTimeout: 5 * time.Minute,
-			}
-			err := port80srv.ListenAndServe()
-			if err != nil {
-				if err != http.ErrServerClosed {
-					log.Fatal(err)
+		if (*certMode == "letsencrypt") {
```
bradfitz, line 238: don't need these parens.

But let's do this more explicitly with a new flag, rather than tying it to the TLS mode.

It has nothing to do with LetsEncrypt.

arnaud-dezandee: @bradfitz I've updated the PR with a more comprehensive changeset following your comment. Let me know

bradfitz:
Sorry, I don't like all the new flags and deprecations. This makes it too complicated.

Please add at most one new flag without deprecating existing ones or changing the current behavior for users not using the new flag.

You can change the `-a` flag to clarify that it's for TLS if it it's port :443 or `--certmode=manual` is given.

Then I guess add `--http-port` that defaults to 80. And say that -1 means means disabled. And that it's only used when the server's running in TLS mode ("see the -a flag").

arnaud-dezandee:
Hello @bradfitz,
I updated the PR with only 1 additionnal CLI flag as you suggested and pull rebased the commit.
Let me know

cmd/derper/derper.go
```diff
@@ -128,10 +132,18 @@ func main() {
 		tsweb.DevMode = true
 	}
 
-	listenHost, _, err := net.SplitHostPort(*addr)
+	if *insecurePort != 80 && *certMode == "letsencrypt" {
```
bradfitz, line 135: what does port 80 have to do with letsencrypt in this context?

cmd/derper/derper.go
```diff
@@ -128,10 +132,18 @@ func main() {
...
+		log.Fatalf("ERR: invalid http-port value with certmode letsencrypt. Use --certmode=manual to disable.")
```
bradfitz, line 136: drop ERR.

but maybe just delete all this

cmd/derper/derper.go
```diff
@@ -128,10 +132,18 @@ func main() {
...
+	}
+
+	bindAddress, securePortStr, err := net.SplitHostPort(*addr)
```
bradfitz, line 139: we don't generally use "bind" and use Go vocab "listen" instead like it was

cmd/derper/derper.go
```diff
@@ -35,7 +38,8 @@ import (
 
 var (
 	dev           = flag.Bool("dev", false, "run in localhost development mode")
-	addr          = flag.String("a", ":443", "server address")
+	addr          = flag.String("a", ":443", "The server address on which to serve HTTPS")
```
bradfitz, line 41: it used to support running non-HTTPS on the `-a` listen address. No longer?

cmd/derper/derper.go
```diff
@@ -200,82 +210,91 @@ func main() {
...
+
+			log.Printf("derper: listening encrypted on %s", secureServer.Addr)
+			err = secureServer.ListenAndServeTLS("", "")
+			if err != nil && err != http.ErrServerClosed {
+				log.Fatalf("derper: %v", err)
+			}
+		}()
 	}
+
+	runtime.Goexit()
```
bradfitz, line 297: don't use this.

cmd/derper/derper.go
```diff
@@ -200,82 +210,91 @@ func main() {
...
-			mux.ServeHTTP(w, r)
-		})
+	if *insecurePort > 0 {
 		go func() {
-			port80srv := &http.Server{
-				Addr:        net.JoinHostPort(listenHost, "80"),
-				Handler:     certManager.HTTPHandler(tsweb.Port80Handler{Main: mux}),
+			var handler http.Handler = tsweb.Port80Handler{Main: mux}
+			if *insecurePort == 80 && *certMode == "letsencrypt" {
+				handler = certManager.HTTPHandler(tsweb.Port80Handler{Main: mux})
```
bradfitz, line 225: Oh, this is probably the root of the confusion. The type is named "Port80" but that just means "not TLS".  We don't actually need to listen on port 80 for LetsEncrypt; we don't do http-01 challenges.

cmd/derper/derper.go
```diff
@@ -200,82 +210,91 @@ func main() {
...
 		}()
-		err = httpsrv.ListenAndServeTLS("", "")
-	} else {
-		log.Printf("derper: serving on %s", *addr)
-		err = httpsrv.ListenAndServe()
 	}
-	if err != nil && err != http.ErrServerClosed {
-		log.Fatalf("derper: %v", err)
+
+	if securePort > 0 {
```
bradfitz, line 247: this whole block gets very indented fast

can this PR do less refactoring? Or can we do refactoring in its own no-op commit that's purely about refactoring?

arnaud-dezandee:
Alright, here is the minimal changeset required to achieve my needs. Hopes this time it will be ok for you @bradfitz 
Regards

cmd/derper/derper.go
```diff
@@ -36,6 +36,7 @@ import (
 var (
 	dev           = flag.Bool("dev", false, "run in localhost development mode")
 	addr          = flag.String("a", ":443", "server address")
+	httpPort      = flag.String("http-port", "80", "The port on which to serve HTTP. Set to 0 to disable")
```
bradfitz, line 39: 0 for ports typically means auto-select. Since you used a String, how about saying the empty string means to not run it instead.

arnaud-dezandee: On this aspect, I followed Kubernetes guidance which seems to use `0` as a disable value for this kind of listening handler notably on the `kube-apiserver` with CLI flags like `--insecure=port=0` or on the `kubelet` with `--healthz-port int32     Default: 10248`. They all use `set to 0 to disable`.
I've never seen `0` as auto-select behavior, do you have an example in mind ?

arnaud-dezandee: also `--http-port=""` or `--http-port` cli flags looks a bit confusing to me to disable said port

bradfitz: With all operating systems, listening on port 0 asks the operating system to pick a random port.

See https://www.google.com/search?q=port+0+wildcard+listen+random+port with various links.

We use & depend on that for e.g. integration tests: https://github.com/tailscale/tailscale/blob/v1.18.1/tstest/integration/integration_test.go#L710

Then our test runner just observes what port number was given to us by the kernel.

Kubernetes is apparently special-casing 0 to mean "disabled". I hadn't seen that style before.

Since it's still a string, how about special-casing the word `"off"`? Does that read nicer? `--http-port=off`?

arnaud-dezandee: Well, I was not aware of `0` meaning asking the kernel for a random free port. Thanks for pointing that out
Then, if you rely on this behavior I guess the best value is to use `-1` as you suggested from the start which is a widespread "disable this" value.
I'll also set the flag back as an `Int`

PR #3265, opened 2021-11-04
ipn/ipnlocal: handle key extensions after key has already expired

### LINKED ISSUE #2515: Issues with extend key/reauthentication
We have just reached 6 months since setting up tailscale, so most of our users are having to reauthenticate. 

I had to reauthenticate for a VM, extending the key did not let me connect. When I connected in via my physical machine Tailscale showed the greyed out icon (as if disconnected) and I right clicked and logged in, which got me working again.

One of my colleagues also had a key expire. In that case they did not have a way to access it, and I had to RDP from my VM in the office, into their machine via local IP. In that case tailscale appeared to be connected, and I just did a cycle of disconnect, reconnect. This seemed to resolve the connection issue, but the key is still showing as about to expire.

It's not clear from the instructions how/if you can reauthenticate a windows machine while connected to it via RDP over tailscale? For the time being I have turned off key expiry for users that do not have an alternative way of connecting to the machine.

maisem:
Fixes #2515

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -122,6 +122,7 @@ type LocalBackend struct {
 	engineStatus     ipn.EngineStatus
 	endpoints        []tailcfg.Endpoint
 	blocked          bool
+	keyExpired       bool
 	authURL          string // cleared on Notify
 	authURLSticky    string // not cleared on Notify
 	interact         bool
@@ -465,7 +466,21 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 
 	b.mu.Lock()
 	wasBlocked := b.blocked
+	keyExpiryExtended := false
+	if st.NetMap != nil {
+		wasExpired := b.keyExpired
+		isExpired := !st.NetMap.Expiry.IsZero() && st.NetMap.Expiry.Before(time.Now())
+		if wasExpired && !isExpired {
+			keyExpiryExtended = true
+		}
+		b.keyExpired = isExpired
+	}
 	b.mu.Unlock()
+
+	if keyExpiryExtended && wasBlocked {
+		// Key extended, unblock the engine
+		b.blockEngineUpdates(false)
+	}
 
 	if st.LoginFinished != nil && wasBlocked {
 		// Auth completed, unblock the engine
@@ -1774,6 +1789,12 @@ func (b *LocalBackend) NetMap() *netmap.NetworkMap {
 	return b.netMap
 }
 
+func (b *LocalBackend) isEngineBlocked() bool {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.blocked
+}
+
 // blockEngineUpdate sets b.blocked to block, while holding b.mu. Its
 // indirect effect is to turn b.authReconfig() into a no-op if block
 // is true.
@@ -2396,6 +2417,7 @@ func (b *LocalBackend) nextState() ipn.State {
 		wantRunning = b.prefs.WantRunning
 		loggedOut   = b.prefs.LoggedOut
 		st          = b.engineStatus
+		keyExpired  = b.keyExpired
 	)
 	b.mu.Unlock()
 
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/local.go
```diff
@@ -2358,7 +2360,14 @@ func (b *LocalBackend) enterState(newState ipn.State) {
...
 			systemd.Status("Stopped; run 'tailscale up' to log in")
 		}
-	case ipn.Starting, ipn.NeedsMachineAuth:
+	case ipn.Starting:
+		if oldState == ipn.NeedsLogin && !keyExpired && blocked {
+			// This means that key was extended **after** expiry,
+			// unblock engine updates.
+			b.blockEngineUpdates(false)
+		}
+		fallthrough
```
bradfitz, line 2369: this isn't intuitive on its own (I understand how you got here from the old code, but people won't have that context later). e.g. How's Starting related to NeedsMachineAuth?

I'd rather you just copy/paste those two lines if we need them in this case.

maisem: makes sense, copied the two lines.

ipn/ipnlocal/local.go
```diff
@@ -2358,7 +2360,16 @@ func (b *LocalBackend) enterState(newState ipn.State) {
...
 		}
-	case ipn.Starting, ipn.NeedsMachineAuth:
+	case ipn.Starting:
+		if oldState == ipn.NeedsLogin && !keyExpired && blocked {
+			// This means that key was extended **after** expiry,
+			// unblock engine updates.
+			b.blockEngineUpdates(false)
+		}
+		b.authReconfig()
+		// Needed so that UpdateEndpoints can run
```
bradfitz, line 2370: it's not obvious what line (prior or following) this comment is referring to, or why they make UpdateEndpoints run?

same on line 2374

ipn/ipnlocal/local.go
```diff
@@ -2358,7 +2360,16 @@ func (b *LocalBackend) enterState(newState ipn.State) {
 		if authURL == "" {
 			systemd.Status("Stopped; run 'tailscale up' to log in")
 		}
-	case ipn.Starting, ipn.NeedsMachineAuth:
+	case ipn.Starting:
+		if oldState == ipn.NeedsLogin && !keyExpired && blocked {
```
bradfitz, line 2364: This all seems oddly specific.

Seems like there might be a bunch of ways to get stuck and we're addressing just one we saw.

Why is key expiry only sticking on NeedsLogin => Starting?

Why don't we unblock in LocalBackend.setClientStatus instead where it's already unblocked?

I mean, how do we know the `blocked` here being true means that it was blocked due to the key being expired?

maisem: > Why don't we unblock in LocalBackend.setClientStatus instead where it's already unblocked?

That makes a lot more sense, how's this?

ipn/ipnlocal/local.go
```diff
@@ -463,10 +464,18 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 		return
 	}
 
+	keyExpired := st.NetMap != nil && !st.NetMap.Expiry.IsZero() && time.Until(st.NetMap.Expiry) <= 0
```
bradfitz, line 467: time.Until reads a bit awkward to me at least here. Maybe `time.Now().After(st.NetMap.Expiry)` ?

bradfitz (commented):
No updates to `ipn/ipnlocal/state_test.go` ?

ipn/ipnlocal/local.go
maisem: Added an explicit check for NetMap and a comment to improve readability.

ipn/ipnlocal/local.go
maisem: I honestly have no idea what it does/means either, but I'm not changing that code now so leaving it as is.

ipn/ipnlocal/local.go
```diff
@@ -463,10 +464,18 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 		return
 	}
 
+	keyExpired := st.NetMap != nil && !st.NetMap.Expiry.IsZero() && time.Now().After(st.NetMap.Expiry)
 	b.mu.Lock()
 	wasBlocked := b.blocked
+	wasKeyExpired := b.keyExpired
+	b.keyExpired = keyExpired
```
bradfitz, line 471: I think you only want to mutate b.keyExpired if st.NetMap != nil.  Otherwise it should be left how it was.

bradfitz (approved):

ipn/ipnlocal/local.go
maisem: good point. makes the logic easier to follow too.

PR #3266, opened 2021-11-05
add in dockerignore

lpmi-13:
extra files in the build container probably don't matter so much due to the multi-stage build, but if there are additional files that can be ignored, particularly tests or artefacts that don't need to be copied over, it would speed up the build, and there's no reason not to do it.

Signed-off-by: Adam Leskis <leskis@gmail.com>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+*.md
+AUTHORS
+LICENSE
+PATENTS
+*.txt


DentonGentry:
We use a multistage build for this reason, so we have the short list of things to copy out rather than a long list of things to clean up.

I don't think we want to add this: an ignore list makes it look like it is safe to rely on, which takes us down the path of needing to keep it updated and being surprised when it isn't complete.

PR #3277, opened 2021-11-08
net/interfaces/windows: update Tailscale interface detection logic to account for new wintun naming.

### LINKED ISSUE #3260: Use of exit nodes fails with Windows 7-32 in 1.17.186
### What is the issue?

In a Windows 7 (32 bit) VM running 1.17.186, enabling an exit node results in DNS being unable to resolve addresses. The global DNS servers for my tailnet are set to Google (8.8.8.8, 8.8.4.4, and the IPv6 addresses for those).

Note that:
- Windows 10 shows no such issue, DNS is working with an exit node
- Android 12 and Android 6 show similar behavior, DNS stops working: https://github.com/tailscale/tailscale/issues/3259

### Steps to reproduce

_No response_

_No response_

### OS

Windows

### OS version

Windows 7

### Tailscale version

1.17.186

### Bug report

_No response_

maisem:
With wintun 0.14, we end up binding to the wrong default interface. This is likely related to the new installation mechanism which probably also changed the naming of the tunnel. We are looking for `Tailscale Tunnel` in the description, but now it's `Wintun Userspace Tunnel #24`. 

Fixes #3260

### DIFF
--- a/tsconst/interface.go
+++ b/tsconst/interface.go
@@ -7,5 +7,6 @@
 package tsconst
 
 // WintunInterfaceDesc is the description attached to Tailscale
-// interfaces on Windows. This is set by our modified WinTun driver.
+// interfaces on Windows. This is set by the WinTun driver.
 const WintunInterfaceDesc = "Tailscale Tunnel"
+const WintunInterfaceDesc0_14 = "Wintun Userspace Tunnel"
--- a/net/interfaces/interfaces_windows.go
+++ b/net/interfaces/interfaces_windows.go
@@ -116,8 +116,12 @@ func notTailscaleInterface(iface *winipcfg.IPAdapterAddresses) bool {
 	// TODO(bradfitz): do this without the Description method's
 	// utf16-to-string allocation. But at least we only do it for
 	// the virtual interfaces, for which there won't be many.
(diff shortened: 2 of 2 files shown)


tsconst/interface.go
```diff
@@ -9,3 +9,4 @@ package tsconst
 // WintunInterfaceDesc is the description attached to Tailscale
 // interfaces on Windows. This is set by our modified WinTun driver.
```
DentonGentry, line 10: While we're in here:
"This is set by our modified WinTun driver."

Recommend: "This is set by the WinTun driver."

We de-forked wintun some time ago.

DentonGentry (approved):

bradfitz:
I'm too late, but I was going to say: isn't there a GUID or something we could use rather than a string match?

maisem:
> I'm too late, but I was going to say: isn't there a GUID or something we could use rather than a string match?

That's probably something we should do, though the issue is that we have call sites in a bunch of places where we don't have access to the LUID. My attempts at getting the LUID from the GUID we set in tstun[1] failed with a `The parameter is incorrect error` on Windows 7 32bit so I gave up.

[1]  https://github.com/tailscale/tailscale/blob/main/net/tstun/tun_windows.go#L15

Issue #2642, opened 2021-08-12
userspace-networking incoming TCP doesn't always work right away


bradfitz:
I've noticed a few times now, most recently with 1.12.3, that a tailscaled running in userspace-networking mode doesn't always forward incoming TCP connections to localhost reliably for a bit after the process is started.

Note that I can ping it and the first curl fails, but then it works a bit later: 
```
crow5k:~ $ ping foo
PING foo.tailscale.com.beta.tailscale.net (100.127.0.49): 56 data bytes
64 bytes from 100.127.0.49: icmp_seq=0 ttl=64 time=161.106 ms
^C
--- foo.tailscale.com.beta.tailscale.net ping statistics ---
2 packets transmitted, 1 packets received, 50.0% packet loss
round-trip min/avg/max/stddev = 161.106/161.106/161.106/0.000 ms

crow5k:~ $ curl http://foo:8383/debug/varz
curl: (7) Failed to connect to foo port 8383: Connection refused

crow5k:~ $ curl http://foo:8383/debug/varz
[works, output omitted]
```

michaelyork:
+1, have seen this exact same behavior. Happy to share logs if helpful.

bradfitz:
Maybe related to lazy wg config, @danderson? Or some lazy/async netstack config?

Example today:

logs: fceaf5a9b386893703e09cd23941b959ebdde1fd05e324206bcc35be29ea585e
starting at: 2021-09-23 20:55:34.683177622

My desktop to server's ssh:
```
2021-09-23 20:55:43.479715179 +0000 UTC: Accept: TCP{100.123.101.40:61047 > 100.127.0.48:22} 64 tcp ok
```

... only one filter log line, but it failed quickly several times in a row, getting immediate connection resets.

Me taking a bug report:
```
2021-09-23 20:55:57.782114316 +0000 UTC: user bugreport: BUG-fceaf5a9b386893703e09cd23941b959ebdde1fd05e324206bcc35be29ea585e-20210923205557Z-868457bada3f20f2
```

Me trying again and it then working:

```
2021-09-23 20:56:02.777491432 +0000 UTC: Accept: TCP{100.123.101.40:61053 > 100.127.0.48:22} 64 tcp ok
2021-09-23 20:56:02.927086398 +0000 UTC: netstack: forwarding incoming connection to 127.0.0.1:22
```

bradfitz:
The 255.255.255.255/32 thing didn't fix it.

[...]

PR #3282, opened 2021-11-09
go.mod: bump inet.af/netstack

bradfitz:
Updates #2642 (I'd hoped, but doesn't seem to fix it)


bradfitz:
Ugh, this is all broken on non-Linux now.

Sigh. Tear.

bradfitz:
@crawshaw, PTAL. It doesn't fix the bug, but it advances our netstack dep by several months of commits (for better or worse?).

crawshaw (approved):
Worth getting in now at the beginning of the next cycle.

Issue #2500, opened 2021-07-22
Ubuntu disables Tailscale's apt sources; actively revert that back?


bradfitz:
I just ran:

```
root@frac:~# apt-get update
Hit:1 http://dl.google.com/linux/chrome/deb stable InRelease
Hit:2 http://us.archive.ubuntu.com/ubuntu hirsute InRelease                     
Hit:3 http://security.ubuntu.com/ubuntu hirsute-security InRelease              
Hit:4 http://us.archive.ubuntu.com/ubuntu hirsute-updates InRelease
Hit:5 http://us.archive.ubuntu.com/ubuntu hirsute-backports InRelease
Reading package lists... Done

root@frac:~# apt-get install tailscale
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
tailscale is already the newest version (1.9.44).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

Wut? 1.9.44 is the latest?

```
root@frac:~# cat /etc/apt/sources.list.d/tailscale.list
# Tailscale packages for ubuntu focal
# deb https://pkgs.tailscale.com/unstable/ubuntu groovy main # disabled on upgrade to groovy
```

Oh, damn.

Ubuntu just removes our entry on any distro upgrade.

Perhaps our tailscaled (old, but still running) should do a HEAD request to `pkgs.tailscale.com` periodically and if/once it sees the current $DISTRIB_CODENAME accessible, uncomment those lines from the `/etc/apt/sources.list.d/tailscale.list` file.

/cc @danderson @crawshaw @Xe

DentonGentry:
Possibly related, Ubuntu 21 and Debian 11 want third-parties to change how their repos are signed: https://github.com/tailscale/tailscale/issues/1876

(draft PR, still working on tests: https://github.com/tailscale/corp/pull/2257)

danderson:
Not related particularly, just a sharp edge of some kind of upgrade script.

I worry about having tailscaled "helpfully" edit package configuration files owned by the admin, that feels a bit borderline to me. I don't have better ideas though, short of "can we figure out how to make Debian stop doing that?" Does whatever upgrade script this is actively check if a new version's repo is available?

bradfitz:
@creachadair just hit this today:

```
$ sudo tailscale update
This will update Tailscale from 1.94.2 to 1.96.4. Continue? [Y/n]
open /etc/apt/sources.list.d/tailscale.list: no such file or directory

[...]

PR #3286, opened 2021-11-10
hostinfo, control/controlclient: tell control when Ubuntu has disabled Tailscale's sources

### LINKED ISSUE #3177: control/controlclient: tell control when Ubuntu has disabled Tailscale's sources
(Related but separate from #2500)

I just upgraded my Ubuntu desktop to Impish, which again disabled my `/etc/apt/sources.list.d/tailscale.list` file, commenting it out with:

`# disabled on upgrade to impish`

... at the end.

#2500 tracks fighting that and undoing it.
 
But this bug tracks a safer thing we can always do: have tailscaled detect that it's commented out and report that to control, so we can surface it as a warning in the admin panel.

/cc @danderson @DentonGentry @crawshaw

bradfitz:
Fixes #3177
Updates #2500

### DIFF
--- a/hostinfo/hostinfo.go
+++ b/hostinfo/hostinfo.go
@@ -7,11 +7,14 @@
 package hostinfo
 
 import (
+	"bufio"
 	"io"
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"sync/atomic"
+	"time"
 
 	"go4.org/mem"
 	"tailscale.com/tailcfg"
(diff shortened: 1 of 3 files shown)


hostinfo/hostinfo.go
```diff
@@ -224,3 +227,55 @@ func inKubernetes() bool {
...
 }
+
+type etcAptSrcResult struct {
+	mod      time.Time
+	disabled bool
+}
+
+var etcAptSrcCache atomic.Value // of etcAptSrcResult
+
+// DisabledEtcAptSource reports whether Ubuntu (or similar) has disables
```
crawshaw, line 238: disabled

crawshaw (approved):

PR #3290, opened 2021-11-11
wgengine/wgcfg: recover from mismatched PublicKey/Endpoints

josharian:
In rare circumstances (tailscale/corp#3016), the PublicKey
and Endpoints can diverge.

This by itself doesn't cause any harm, but our early exit
in response did, because it prevented us from recovering from it.

Remove the early exit.


bradfitz (approved):

josharian:
Going to delay merging this until after 1.18, unless I hear opinions otherwise.

PR #3306, opened 2021-11-15
net/dns: correctly handle NetworkManager-managed DNS that points to resolved

### LINKED ISSUE #3304: Linux DNS manager selects wrong implementation for NM+resolved
### What is the issue?

Given a /etc/resolv.conf that looks like:

```
$ ls -l /etc/resolv.conf
lrwxrwxrwx 1 root root 31 Aug 26 12:50 /etc/resolv.conf -> /run/NetworkManager/resolv.conf

$ cat /etc/resolv.conf
# Generated by NetworkManager
nameserver 127.0.0.53
options edns0 trust-ad
```

The linux DNS manager concludes that NetworkManager owns the configuration, and per our "don't use NM because bugs" policy, selects the only other alternative, direct editing of /etc/resolv.conf.

We need to instead detect that NM is pointing at systemd-resolved, and program resolved.

### Steps to reproduce

_No response_

_No response_

### OS

_No response_

### OS version

ubuntu 21.10

### Tailscale version

1.16.2

### Bug report

BUG-a1a262c98a7285d3d91bfecfdbf10c1795b232b5ab3bbe9f39109935ca264c93-20211115060549Z-a001ab64a5728186

Front logo [Front conversations](https://app.frontapp.com/open/top_3ujm9)

danderson:
Fixes #3304


net/dns/manager_linux.go
```diff
@@ -180,20 +180,51 @@ func dnsMode(logf logger.Logf, env newOSConfigEnv) (ret string, err error) {
...
-		//
-		// If you ever try to put NMManager back here, keep in mind
-		// that versions >=1.26.6 will ignore DNS configuration
-		// anyway, so you still need a fallback path that uses
-		// directManager.
 		dbg("rc", "nm")
-		return "direct", nil
+		// Sometimes, NetworkManager owns the configuration but points
+		// it at systemd-resolved.
+		if err := resolvedIsActuallyResolver(bs); err != nil {
```
bradfitz, line 186: add a same-line comment indicating what err != nil here means? Does this mean it is or isn't actually the resolver?

net/dns/manager_linux.go
```diff
@@ -180,20 +180,51 @@ func dnsMode(logf logger.Logf, env newOSConfigEnv) (ret string, err error) {
...
+			// directManager.
+			return "direct", nil
+		}
+		dbg("nm-resolved", "yes")
+
+		if err := env.dbusPing("org.freedesktop.resolve1", "/org/freedesktop/resolve1"); err != nil {
+			dbg("resolved", "no")
+			return "direct", nil
+		}
+		if err := env.dbusPing("org.freedesktop.NetworkManager", "/org/freedesktop/NetworkManager/DnsManager"); err != nil {
```
bradfitz, line 209: add a comment in here about when this happens? The /etc/resolv.conf file made us think NetworkManager was in use, but then it turns out it's not running?

bradfitz (approved):
I'll trust you on this. :)

Left some comments on new comments that help future features (or me)

net/dns/manager_linux.go
danderson: Added a docstring to resolvedIsActuallyResolver, explaining what it does.

net/dns/manager_linux.go
danderson: Done. In this codepath, we're trying to figure out if we have to program DNS via NetworkManager, even though resolved is nominally in charge. Step one of that is "is NM even running?".

PR #3311, opened 2021-11-15
cmd/tailscale/cli: migrate hidden debug subcommand to use subcomands

bradfitz:
It was a mess of flags. Use subcommands under "debug" instead.

And document loudly that it's not a stable interface.

### DIFF
--- a/cmd/tailscale/cli/debug.go
+++ b/cmd/tailscale/cli/debug.go
@@ -24,39 +24,66 @@ import (
 )
 
 var debugCmd = &ffcli.Command{
-	Name: "debug",
-	Exec: runDebug,
+	Name:     "debug",
+	Exec:     runDebug,
+	LongHelp: `"tailscale debug" contains misc debug facilities; it is not a stable interface.`,
 	FlagSet: (func() *flag.FlagSet {
 		fs := newFlagSet("debug")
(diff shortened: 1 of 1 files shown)


bradfitz:
(This is all code movement. The diff looks scary than it is.)

sailorfrag (approved):

cmd/tailscale/cli/debug.go
```diff
@@ -24,39 +24,66 @@ import (
...
+				fs.BoolVar(&prefsArgs.pretty, "pretty", false, "If true, pretty-print output")
+				return fs
+			})(),
+		},
+		&ffcli.Command{
+			Name:      "watch-ipn",
+			Exec:      runWatchIPN,
+			ShortHelp: "subscribe to IPN message bus",
+			FlagSet: (func() *flag.FlagSet {
+				fs := newFlagSet("watch-ipn")
```
DentonGentry, line 74: So this will be: `tailscale debug watch-ipn --netmap`, where the old one was `tailscale debug --ipn --netmap` ?
(seems fine)

DentonGentry (approved):

cmd/tailscale/cli/debug.go
bradfitz: Well, it's always defaulted to true, so the way you'd disable the verbose netmap is to do `tailscale debug watch-ipn --netmap=false`. The common case would be `tailscale debug watch-ipn` without flags.

Issue #3307, opened 2021-11-15
add client metrics


bradfitz:
Tracking bug for adding metrics ("varz") to the client software for debugging and letting users have more visibility.

(I thought we had such a bug already but I can't find it.)

bradfitz:
Sufficiently done now, and shipped in Tailscale 1.18.

More things need to be annotated, but the infrastructure and a bunch of metrics are there.

Other bugs will track specific things to annotate.

PR #3313, opened 2021-11-16
util/clientmetric, logtail: log metric changes

bradfitz:
Updates #3307


logpolicy/logpolicy.go
```diff
@@ -500,6 +501,9 @@ func New(collection string) *Policy {
 		},
 		HTTPC: &http.Client{Transport: newLogtailTransport(logtail.DefaultHost)},
 	}
+	if collection == "tailnode.log.tailscale.io" {
```
sailorfrag, line 504: optional: I wonder if this should be a constant

util/clientmetric/clientmetric.go
```diff
@@ -7,18 +7,24 @@
...
 )
 
 var (
-	mu          sync.Mutex
+	mu          sync.Mutex // guards vars in this block
 	metrics     = map[string]*Metric{}
-	sortedDirty bool
-	sorted      []*Metric
+	numWireID   int       // how many wireIDs have been allocated
+	lastDelta   time.Time // time of last call to EncodeLogTailMetricsDelta
```
sailorfrag, line 25: optional: would it make sense to attach the information for generating deltas to the particular log stream rather than being global? That would make it easier to send logs to two destinations concurrently if we want to do that in future.

util/clientmetric/clientmetric.go
```diff
@@ -133,3 +149,123 @@ func WritePrometheusExpositionFormat(w io.Writer) {
...
+}
+
+const hextable = "0123456789abcdef"
+
+// writeHexVarint writes v to the buffer as a hex-encoded varint.
+func (b *deltaEncBuf) writeHexVarint(v int64) {
+	n := binary.PutVarint(b.scratch[:], v)
+	const hextable = "0123456789abcdef"
+	for _, o := range b.scratch[:n] {
+		b.buf.WriteByte(hextable[o>>4])
```
sailorfrag, line 268: Just curious, why do this yourself rather than using `encoding/hex`?

sailorfrag (approved):

util/clientmetric/clientmetric.go
bradfitz: I was trying to avoid another scratch buffer but really I just forgot about bytes.Buffer.Grow. I can use Buffer.Grow and then get at the dest bytes for encoding/hex to write into using Buffer.Bytes(). Will fix.

util/clientmetric/clientmetric.go
bradfitz: It might make sense, but I don't want to overdesign and complicate the code [yet] for something we might never use.

logpolicy/logpolicy.go
bradfitz: Will do.

PR #3326, opened 2021-11-16
control/controlclient: treat status 4XX errors as user visible

dsnet:
For auth keys with tags, it is possible that use of the key
results in a 400 error because the tag associated with the key
no longer exists (it existed when the key was originally created).

Prior to this change, `tailscale up -authkey=...` would loop
indefinitely without any problems reported to the user.

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -421,8 +421,19 @@ func (c *Direct) doLogin(ctx context.Context, opt loginOpt) (mustRegen bool, new
 	if res.StatusCode != 200 {
 		msg, _ := ioutil.ReadAll(res.Body)
 		res.Body.Close()
-		return regen, opt.URL, fmt.Errorf("register request: http %d: %.200s",
-			res.StatusCode, strings.TrimSpace(string(msg)))
+		var err error
+		if 400 <= res.StatusCode && res.StatusCode < 500 {
+			const maxErrorLength = 1024
+			if len(msg) > maxErrorLength {
+				err = UserVisibleError(msg[:maxErrorLength]) + "..."
+			} else {
+				err = UserVisibleError(msg)
+			}
+		} else {
+			err = fmt.Errorf("register request: http %d: %.200s",
+				res.StatusCode, strings.TrimSpace(string(msg)))
+		}
+		return regen, opt.URL, err
 	}
 	resp := tailcfg.RegisterResponse{}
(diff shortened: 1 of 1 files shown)


dsnet:
This is a follow-up to #3257

control/controlclient/direct.go
```diff
@@ -421,8 +421,12 @@ func (c *Direct) doLogin(ctx context.Context, opt loginOpt) (mustRegen bool, new
 	if res.StatusCode != 200 {
 		msg, _ := ioutil.ReadAll(res.Body)
 		res.Body.Close()
-		return regen, opt.URL, fmt.Errorf("register request: http %d: %.200s",
+		err := fmt.Errorf("register request: http %d: %.200s",
```
bradfitz, line 424: This is an ugly message to be popped up in a Windows dialog box.

We control the server, so we should instead pass back the exact error message we want to show.

dsnet: Done. I kept some sensible max length limit.

After this change I'll scrub through the control server, making sure that error paths that return 4XX return something sensible.

control/controlclient/direct.go
```diff
@@ -421,8 +421,19 @@ func (c *Direct) doLogin(ctx context.Context, opt loginOpt) (mustRegen bool, new
 	if res.StatusCode != 200 {
 		msg, _ := ioutil.ReadAll(res.Body)
 		res.Body.Close()
-		return regen, opt.URL, fmt.Errorf("register request: http %d: %.200s",
-			res.StatusCode, strings.TrimSpace(string(msg)))
+		var err error
+		if 400 <= res.StatusCode && res.StatusCode < 500 {
```
bradfitz, line 425: How about we just don't send 400s from control's serveRegister then? And instead use RegisterResponse.Error and the line 452 path below?

(and we should fix that logging immediately before it to instead logf the error and not logf the "got response" until after the Error != "" path)

bradfitz: I just don't like assuming that these 400s have any sort of defined body ... a plain text string? HTML? a JSON response?

I don't like just blindly shoving one of those 3 types into a MessageBox.

It'd be nice to know it's actually a text string, which RegisterResponse.Error already has.

dsnet: It's kinda gross that we return status 200, but also have an error to report.

Given that this code path already exists, I'll modify the control server to use that.

PR #3328, opened 2021-11-17
net/tstun, wgengine: add packet-level and drop metrics

bradfitz:
Primarily tstun work, but some MagicDNS stuff spread into wgengine.

No wireguard reconfig metrics (yet).

Updates #3307

### DIFF
--- a/net/tstun/wrap.go
+++ b/net/tstun/wrap.go
@@ -27,6 +27,7 @@ import (
 	"tailscale.com/types/key"
 	"tailscale.com/types/logger"
 	"tailscale.com/types/pad32"
+	"tailscale.com/util/clientmetric"
 	"tailscale.com/wgengine/filter"
 )
 
@@ -421,11 +422,13 @@ func (t *Wrapper) filterOut(p *packet.Parsed) filter.Response {
 	if p.IPProto == ipproto.UDP && // disco is over UDP; avoid isSelfDisco call for TCP/etc
 		t.isSelfDisco(p) {
 		t.logf("[unexpected] received self disco out packet over tstun; dropping")
+		metricPacketOutDropSelfDisco.Add(1)
 		return filter.DropSilently
 	}
 
 	if t.PreFilterOut != nil {
 		if res := t.PreFilterOut(p, t); res.IsDrop() {
+			// Handled by userspaceEngine.handleLocalPackets (quad-100 DNS primarily).
 			return res
 		}
 	}
@@ -437,6 +440,7 @@ func (t *Wrapper) filterOut(p *packet.Parsed) filter.Response {
 	}
 
 	if filt.RunOut(p, t.filterFlags) != filter.Accept {
+		metricPacketOutDropFilter.Add(1)
 		return filter.Drop
 	}
 
@@ -471,6 +475,8 @@ func (t *Wrapper) Read(buf []byte, offset int) (int, error) {
 	if res.err != nil {
 		return 0, res.err
 	}
+
+	metricPacketOut.Add(1)
 	pkt := res.data
 	n := copy(buf[offset:], pkt)
 	// t.buffer has a fixed location in memory.
@@ -496,6 +502,7 @@ func (t *Wrapper) Read(buf []byte, offset int) (int, error) {
 	if !isInjectedPacket && !t.disableFilter {
 		response := t.filterOut(p)
 		if response != filter.Accept {
+			metricPacketOutDrop.Add(1)
 			// Wireguard considers read errors fatal; pretend nothing was read
 			return 0, nil
 		}
@@ -529,6 +536,7 @@ func (t *Wrapper) filterIn(buf []byte) filter.Response {
 	if p.IPProto == ipproto.UDP && // disco is over UDP; avoid isSelfDisco call for TCP/etc
 		t.isSelfDisco(p) {
 		t.logf("[unexpected] received self disco in packet over tstun; dropping")
+		metricPacketInDropSelfDisco.Add(1)
 		return filter.DropSilently
 	}
 
@@ -558,6 +566,7 @@ func (t *Wrapper) filterIn(buf []byte) filter.Response {
 	}
 
 	if outcome != filter.Accept {
+		metricPacketInDropFilter.Add(1)
 
 		// Tell them, via TSMP, we're dropping them due to the ACL.
 		// Their host networking stack can translate this into ICMP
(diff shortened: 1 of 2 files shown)


net/tstun/wrap.go
```diff
@@ -721,3 +732,15 @@ func (t *Wrapper) InjectOutbound(packet []byte) error {
 func (t *Wrapper) Unwrap() tun.Device {
 	return t.tdev
 }
+
+var (
+	metricPacketIn              = clientmetric.NewGauge("tstun_in")
```
danderson, line 737: "in" and "out" are not very descriptive. Just like "inbound" and "outbound", I will always have to look up what frame of reference the original author was using. Can you use absolute locators instead, e.g. {to,from}_{kernel,wg} or somesuch?

wgengine/userspace.go
```diff
@@ -1554,3 +1557,8 @@ func (ls fwdDNSLinkSelector) PickLink(ip netaddr.IP) (linkName string) {
 	}
 	return ""
 }
+
+var (
+	metricMagicDNSPacketIn = clientmetric.NewGauge("magicdns_packet_in") // for 100.100.100.100
+	metricForwardToOS      = clientmetric.NewGauge("packet_forward_to_os")
```
danderson, line 1563: reflect rather than forward? It's a packet we got from localhost, which we shouldn't have received, and must reflect back to the kernel for things to work correctly

danderson (approved):

net/tstun/wrap.go
bradfitz: I dunno, to me the frame of reference is `tailscaled`, the entity that's doing the logging and counting. `iptables` calls it `INPUT` and `OUTPUT` and that's not confusing? Is the confusion here that there's also wireguard involved?

Also, `in` and `out` match what we call `filterIn` and `filterOut`. And all the tstun hooks are named with `In$` and `Out$` names.

I can't think of something that's unambiguous for you but also not distractingly long.

I guess maybe `tstun_in_from_wg` as the prefix, so then `tstun_in_from_wg_drop_self_disco`?

And `tstun_out_to_wg` and `tstun_out_to_wg_drop_filter`?

You like those?

wgengine/userspace.go
bradfitz: done

net/tstun/wrap.go
danderson: The thing is, tailscaled has two inputs and outputs: one with the kernel, and one with the tailnet. And in different places in our code, we call either one in/out :/

So when I read "in", I have to figure out: did this mean "in from the tailnet", or "in from the kernel" ? Likewise when I see "out". This is inside tailscaled, so I can't just say "well it's into tailscaled", I also have to figure out into tailscaled _from where_. iptables doesn't have that problem as much, because all its inputs and outputs are the same kind, so it can reason about them as a bundle. But for us, kernel vs. tailnet is an important distinction, because kernel in == talinet out, usually.

I already get confused by tstun's in/out, and empirically I'm not alone given that we initially put the disco looping bugfix on the wrong leg of tstun :P 

I don't really know how to resolve this without making unwieldy names :/ Maybe we can shorten them further like: `tstun_wg_in` and `tstun_tun_in` ? That's unambiguously providing both a point of origin and a direction, so I don't have to go read the code to know where the counter is sitting?

bradfitz: > I already get confused by tstun's in/out, and empirically I'm not alone given that we initially put the disco looping bugfix on the wrong leg of tstun :P

Well, we had little clue what macOS was doing and were largely guessing.

> `tstun_wg_in` and `tstun_tun_in`

Those names confuse me more. "wg_in" can mean "incoming from wireguard from the tailnet" or "we're feeding it in to wireguard to send out". Likewise with "tun".

My compromise with `tstun_in_from_wg` and `tstun_out_to_wg` was that they still started with `in` and `out` like before.

You can pick the suffixes after `tstun_in_` and `tstun_out_`, but let's stick with "in" and "out" prefixes and just agree on what in/out mean in a tstun context where those terms are already widely used. The metrics are largely intended to debug which paths we're flowing through the code, so it makes sense to name them after the code.

And in the context of this code, "in" means another a packet is arriving from somewhere else (in from the tailnet) and "out" means we're sending a packet like iptables OUTPUT (out to the tailnet, or out to 100.100.100.100).

wgengine/userspace.go
```diff
@@ -1554,3 +1557,8 @@ func (ls fwdDNSLinkSelector) PickLink(ip netaddr.IP) (linkName string) {
 	}
 	return ""
 }
+
+var (
+	metricMagicDNSPacketIn = clientmetric.NewGauge("magicdns_packet_in") // for 100.100.100.100
+	metricReflectToOS      = clientmetric.NewGauge("packet_reflect_to_os")
```
josharian, line 1563: Suggest `wgengine_packet_reflect_to_os`. The `tstun` namespacing above seems like a thing worth doing more of.

net/tstun/wrap.go
```diff
@@ -721,3 +732,15 @@ func (t *Wrapper) InjectOutbound(packet []byte) error {
 func (t *Wrapper) Unwrap() tun.Device {
 	return t.tdev
 }
+
+var (
+	metricPacketIn              = clientmetric.NewGauge("tstun_in_from_wg")
+	metricPacketInDrop          = clientmetric.NewGauge("tstun_in_from_wg_drop")
```
josharian, line 738: Maybe s/in/recv/ and s/out/send/ throughout?

josharian (approved):

net/tstun/wrap.go
josharian: Oh, I see this got discussed extensively above. I'm still tempted by send_to and recv_from, but am ok not attempting to repaint the shed.

wgengine/userspace.go
bradfitz: I didn't do it here because it's not really a function of wgengine at a design level. It's a weird hook installed from afar and really it could be done as part of tstun. The exact place it's installed today isn't incredibly relevant.

In the prefixed cases earlier it's a core function of that package.

net/tstun/wrap.go
bradfitz: I'm fine with `tstun_send` and `tstun_recv` too (instead of out and in, respectively) but I imagine @danderson would find those ambiguous for the same reason as above.

bradfitz:
I'm going to merge this because both @danderson and @josharian approved it, and I think the new names address both address the code and @danderson's disambiguation requirements. But if they don't, let me know @danderson and I can rename them.

PR #3331, opened 2021-11-17
util/clientmetric: optimize memory layout for finding updates

bradfitz:
Updates #3307

### DIFF
--- a/util/clientmetric/clientmetric.go
+++ b/util/clientmetric/clientmetric.go
@@ -22,11 +22,27 @@ import (
 var (
 	mu          sync.Mutex // guards vars in this block
 	metrics     = map[string]*Metric{}
-	numWireID   int       // how many wireIDs have been allocated
-	lastDelta   time.Time // time of last call to EncodeLogTailMetricsDelta
-	sortedDirty bool      // whether sorted needs to be rebuilt
-	sorted      []*Metric // by name
+	numWireID   int         // how many wireIDs have been allocated
+	lastDelta   time.Time   // time of last call to EncodeLogTailMetricsDelta
+	sortedDirty bool        // whether sorted needs to be rebuilt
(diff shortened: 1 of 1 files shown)


util/clientmetric/clientmetric.go
```diff
@@ -22,12 +22,27 @@ import (
...
-	sortedDirty bool      // whether sorted needs to be rebuilt
-	sorted      []*Metric // by name
+	numWireID   int         // how many wireIDs have been allocated
+	lastDelta   time.Time   // time of last call to EncodeLogTailMetricsDelta
+	sortedDirty bool        // whether sorted needs to be rebuilt
+	sorted      []*Metric   // by name
+	lastLogVal  []scanEntry // by Metric.regIdx
+	unsorted    []*Metric   // by Metric.regIdx
+
+	// valBackBuf is a set of free contiguous int64s to assign to Metric.v
```
josharian, line 32: I suggest calling this `freelist` or `int64alloc` or the like. This makes it clearer that the purpose is a custom int64 allocator.

"to assign to" isn't technically quite right--we're taking the address, not just assigning.

util/clientmetric/clientmetric.go
```diff
@@ -22,12 +22,27 @@ import (
...
+	// Any memory address in len(valPackBuf) is free for use.
+	// They're contiguous to reduce memory pressure during diff scans.
```
josharian, line 34: s/memory pressure/cache churn/

josharian (approved):

util/clientmetric/clientmetric.go
bradfitz: done

util/clientmetric/clientmetric.go
bradfitz: done

PR #3338, opened 2021-11-18
wgengine/magicsock: fix data race on endpoint.discoKey

josharian:
endpoint.discoKey is protected by endpoint.mu.
endpoint.sendDiscoMessage was reading it without holding the lock.
This showed up in a CI failure and is readily reproducible locally.

The fix is in two parts.

First, for Conn.enqueueCallMeMaybe, eliminate the one-line helper method endpoint.sendDiscoMessage; call Conn.sendDiscoMessage directly.
This makes it more natural to read endpoint.discoKey in a context
in which endpoint.mu is already held.

Second, for endpoint.sendDiscoPing, explicitly pass the disco key
as an argument. Again, this makes it easier to read endpoint.discoKey
in a context in which endpoint.mu is already held.

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -2102,7 +2102,7 @@ func (c *Conn) enqueueCallMeMaybe(derpAddr netaddr.IPPort, de *endpoint) {
 	for _, ep := range c.lastEndpoints {
 		eps = append(eps, ep.Addr)
 	}
-	go de.sendDiscoMessage(derpAddr, &disco.CallMeMaybe{MyNumber: eps}, discoLog)
+	go de.c.sendDiscoMessage(derpAddr, de.publicKey, de.discoKey, &disco.CallMeMaybe{MyNumber: eps}, discoLog)
 }
 
 // discoInfoLocked returns the previous or new discoInfo for k.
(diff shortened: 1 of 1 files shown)


josharian:
@danderson @bradfitz do you think we should try to squeeze this into 1.18? It's not obvious to me how big a risk this poses in practice, and how that risk compares to the risk of pulling in this change.

bradfitz:
> @danderson @bradfitz do you think we should try to squeeze this into 1.18? It's not obvious to me how big a risk this poses in practice, and how that risk compares to the risk of pulling in this change.

Yes, we should. It's a regression from this cycle and data races aren't okay; I don't want to ever debug an issue and wonder if it's race-related.

wgengine/magicsock/magicsock.go
```diff
@@ -3544,13 +3544,16 @@ func (de *endpoint) removeSentPingLocked(txid stun.TxID, sp sentPing) {
 	delete(de.sentPing, txid)
 }
 
-// sendDiscoPing sends a ping with the provided txid to ep.
+// sendDiscoPing sends a ping with the provided txid to ep using discoKey.
```
bradfitz, line 3547: clarify whose discoKey this is. ours vs ep's.

bradfitz (approved):

wgengine/magicsock/magicsock.go
josharian: Done.

PR #3344, opened 2021-11-18
log/filelogger: make filelogger remove redundant date before adding a date

bradfitz:
At some point since filelogger was added on Windows, the log hierarchy
above it changed such that a log.Printf writes to filelogger and includes
the log package's own date. But then filelogger adds another.

Rather than debug everything above and risk removing the prefix when
run by tailscaled, instead just remove the log package's prefix
very late right before we go to add the filelogger's own.


bradfitz:
Added a second commit for the logtail issue.

Tested on Windows both ways: via tailscaled directly, and via a service, checking both logtail output and local disk logs.

maisem (approved):

PR #3348, opened 2021-11-18
net/netns: thread logf into control functions

josharian:
So that darwin can log there without panicking during tests.

### DIFF
--- a/net/dnsfallback/dnsfallback.go
+++ b/net/dnsfallback/dnsfallback.go
@@ -90,7 +90,7 @@ func Lookup(ctx context.Context, host string) ([]netaddr.IP, error) {
 // serverName and serverIP of are, say, "derpN.tailscale.com".
 // queryName is the name being sought (e.g. "controlplane.tailscale.com"), passed as hint.
 func bootstrapDNSMap(ctx context.Context, serverName string, serverIP netaddr.IP, queryName string) (dnsMap, error) {
-	dialer := netns.NewDialer()
+	dialer := netns.NewDialer(log.Printf)
 	tr := http.DefaultTransport.(*http.Transport).Clone()
 	tr.Proxy = tshttpproxy.ProxyFromEnvironment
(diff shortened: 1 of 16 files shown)


josharian:
I'm not entirely sure we should do this. But this is what it looks like. :)

net/dnsfallback/dnsfallback.go
```diff
@@ -90,7 +90,7 @@ func Lookup(ctx context.Context, host string) ([]netaddr.IP, error) {
 // serverName and serverIP of are, say, "derpN.tailscale.com".
 // queryName is the name being sought (e.g. "controlplane.tailscale.com"), passed as hint.
 func bootstrapDNSMap(ctx context.Context, serverName string, serverIP netaddr.IP, queryName string) (dnsMap, error) {
-	dialer := netns.NewDialer()
+	dialer := netns.NewDialer(log.Printf)
```
bradfitz, line 93: missing plumbing here.

I wonder if we should have some `logfFromContext(ctx)` func that defaults to log.Printf.

bradfitz (commented):
Okay, not as bad as I'd feared.

For after 1.18 I assume.

net/dnsfallback/dnsfallback.go
josharian: Most of the callers of bootstrapDNSMap are near startup code (including logging subsystem startup) and use log.Printf themselves. I could thread it a few levels deeper, but it'd still be log.Printf, so I stopped here for now.

> I wonder if we should have some logfFromContext(ctx) func that defaults to log.Printf.

And then start passing around contexts instead of logf everywhere? That'd be really nice. And it'd make changing our logging system a lot easier, if we ever actually do that. cc @danderson 

Work for another day, though.

josharian:
After 1.18 is now!

bradfitz (approved):

PR #3349, opened 2021-11-18
cmd/allsrc: WIP tool to print out all a program's source

bradfitz:


bradfitz:
Going to move this into depaware instead.

PR #3353, opened 2021-11-19
cmd/tailtlsproxy: introduce HTTPS proxy bridge

Xe:
This allows you to bridge existing HTTP/S services into your tailnet using the Let's Encrypt[ 1] functionality. This will allow you to run multiple services on the same computer and still have them point at different target HTTP services.

This also includes an example systemd template unit that lets you easily set up multiple instances of tailtlsproxy on the same machine.

[1]: https://tailscale.com/blog/tls-certs/

### DIFF
--- a/cmd/tailtlsproxy/main.go
+++ b/cmd/tailtlsproxy/main.go
@@ -1,0 +1,86 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"crypto/tls"
+	"flag"
+	"log"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"os"
+	"time"
+
+	"tailscale.com/client/tailscale"
+	"tailscale.com/tsnet"
+)
+
+func envOr(name, def string) string {
+	if val, ok := os.LookupEnv(name); ok {
+		return val
+	}
+
+	return def
+}
+
+var (
+	bind     = flag.String("bind", ":443", "TCP hostport to bind to for TLS http traffic")
+	hostname = flag.String("hostname", envOr("HOSTNAME", "toolname"), "hostname to register on tailnet (can be set with $HOSTNAME)")
+	auth     = flag.Bool("auth", false, "if set, authenticate with tailscale (enables verbose logging)")
+	v        = flag.Bool("v", false, "if set, enable verbose tailscale logs")
+	to       = flag.String("to", envOr("TO", "http://127.0.0.1:3030"), "HTTP/S url to reverse proxy to (can be set with $TO)")
+)
+
+func main() {
+	os.Setenv("TAILSCALE_USE_WIP_CODE", "true")
+	flag.Parse()
+
+	srv := tsnet.Server{
+		Hostname: *hostname,
+	}
+
+	if *auth || *v {
+		srv.Logf = log.Printf
+	} else {
+		srv.Logf = func(string, ...interface{}) {}
+	}
+
+	if *auth {
+		os.Setenv("TS_LOGIN", "1")
+	}
+
+	ln, err := srv.Listen("tcp", *bind)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	u, err := url.Parse(*to)
+	if err != nil {
+		log.Fatalf("%s wasn't a valid URL: %v", *to, err)
+	}
+
+	h := httputil.NewSingleHostReverseProxy(u)
+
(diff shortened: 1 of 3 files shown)


cmd/tailtlsproxy/main.go
```diff
@@ -0,0 +1,86 @@
...
+var (
+	bind     = flag.String("bind", ":443", "TCP hostport to bind to for TLS http traffic")
+	hostname = flag.String("hostname", envOr("HOSTNAME", "toolname"), "hostname to register on tailnet (can be set with $HOSTNAME)")
+	auth     = flag.Bool("auth", false, "if set, authenticate with tailscale (enables verbose logging)")
+	v        = flag.Bool("v", false, "if set, enable verbose tailscale logs")
+	to       = flag.String("to", envOr("TO", "http://127.0.0.1:3030"), "HTTP/S url to reverse proxy to (can be set with $TO)")
+)
+
+func main() {
+	os.Setenv("TAILSCALE_USE_WIP_CODE", "true")
```
bradfitz, line 38: I'd rather not check this line in.

cmd/tailtlsproxy/main.go
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package main
```
bradfitz, line 5: needs package doc.

But I'm not a huge fan of this proposed name.

I thought we were going to put this functionality in tailscaled.

cmd/tailtlsproxy/main.go
```diff
@@ -0,0 +1,86 @@
...
+	u, err := url.Parse(*to)
+	if err != nil {
+		log.Fatalf("%s wasn't a valid URL: %v", *to, err)
+	}
+
+	h := httputil.NewSingleHostReverseProxy(u)
+
+	ln = tls.NewListener(ln, &tls.Config{
+		GetCertificate: func(chi *tls.ClientHelloInfo) (*tls.Certificate, error) {
+			c, err := tailscale.GetCertificate(chi)
```
apenwarr, line 69: Looks like you're not actually using the hostname argument anywhere, and instead you're just randomly trying letsencrypt on every incoming hostname request. Which probably works, but is maybe a little questionable.

cmd/tailtlsproxy/main.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	u, err := url.Parse(*to)
+	if err != nil {
+		log.Fatalf("%s wasn't a valid URL: %v", *to, err)
+	}
+
+	h := httputil.NewSingleHostReverseProxy(u)
+
+	ln = tls.NewListener(ln, &tls.Config{
```
apenwarr, line 67: It doesn't seem like anything will ever refresh the letsencrypt cert?

cmd/tailtlsproxy/main.go
```diff
@@ -0,0 +1,86 @@
...
+		},
+	})
+
+	s := &http.Server{
+		IdleTimeout: 5 * time.Minute,
+		Addr:        *bind,
+		Handler:     h,
+	}
+
+	log.Printf("listening for https on https://%s.your.tailcert.domain and forwarding to %s", *hostname, *to)
```
bradfitz, line 83: you can get that exact URL from the lower layers. It's in the ipnstate.Status.

cmd/tailtlsproxy/main.go
```diff
@@ -0,0 +1,86 @@
...
+	"net/url"
+	"os"
+	"time"
+
+	"tailscale.com/client/tailscale"
+	"tailscale.com/tsnet"
+)
+
+func envOr(name, def string) string {
+	if val, ok := os.LookupEnv(name); ok {
```
apenwarr, line 22: it doesn't seem like much value is created by using environment variables. Maybe just stick to plain args?

cmd/tailtlsproxy/main.go
apenwarr, line 83: I don't think we can actually request a cert for just <hostname>, we have to spell out the entire <hostname>.<domain>, don't we? At least that's how it was from the CLI last time I checked. If so, then the user should just specify the fqdn in *hostname and you don't have to have the ugly .your.tailcert.domain in the help message.

cmd/tailtlsproxy/main.go
Xe: This makes it easier to configure using the `/etc/defaults/tailtlsproxy-$INSTANCE` file

cmd/tailtlsproxy/main.go
bradfitz: GetCertificate does as needed.

Issue #3351, opened 2021-11-19
Tailscaled fails to start after updating to Ver 1.18.0 (arm32) [when IP_MULTIPLE_TABLES is not set]


apocnz:
### What is the issue?

Tailscale fails to start with version 1.18.0 (arm32 linux)

### Steps to reproduce

Running tailscaled from cli/console produces the following

```
Program starting: v1.18.0-t22d969975-g7022e5a4c, Go 1.17.2-ts7037d3ea51: []string{"tailscaled", "-tun", "tailscale0"}
LogID: d08b1955d71bacd9a9c60294120db178bcb58730919f538c76aaff1277cd740a
logpolicy: using system state directory "/var/lib/tailscale"
wgengine.NewUserspaceEngine(tun "tailscale0") ...
router: disabling tunneled IPv6 due to system IPv6 config: kernel doesn't support IPv6 policy routing: operation not supported
dns: [rc=unknown ret=direct]
dns: using *dns.directManager
link state: interfaces.State{defaultRoute=eth0 ifs={eth0:[192.168.0.101/24 192.168.88.88/24]} v4=true v6=false}
magicsock: disco key = d:2851c9d8f76971fc
Creating wireguard device...
Bringing wireguard device up...
Bringing router up...
external route: up
wgengine.NewUserspaceEngine(tun "tailscale0") error: operation not supported
wgengine.New: operation not supported
flushing log.
logger closing down
logtail: dialed "log.tailscale.io:443" in 291ms
```

### Are there any recent changes that introduced the issue?

Updated from 1.16.2_arm to 1.18.0_arm

### OS

Linux

### OS version

Custom Linux distro, based on Kernel 3.18.20

### Tailscale version

1.18.0_arm

bradfitz:
Hm, mystery. I read through the code & the changes and I'm not seeing the issue.

But I sent https://github.com/tailscale/tailscale/pull/3355 to make it more clear in the error messages which initialization path is generating the error.

We'll roll an unstable build with that change and that should point us toward where the problem is.

bradfitz:
Does it work if you run `tailscaled` with the `TS_DEBUG_USE_IP_COMMAND=1` environment variable set?

apocnz:
Hi Brad, 

> Does it work if you run `tailscaled` with the `TS_DEBUG_USE_IP_COMMAND=1` environment variable set?

[...]

PR #3355, opened 2021-11-19
cmd/tailscaled: disambiguate some startup failure error messages

bradfitz:
Updates #3351


bradfitz:
Then I'd like to build an unstable with this, so that user on the bug can try it.

DentonGentry (approved):

maisem (approved):

PR #3361, opened 2021-11-20
wgengine/monitor: ignore adding/removing uninteresting IPs

josharian:
One of the most common "unexpected" log lines is:

"network state changed, but stringification didn't"

One way that this can occur is if an interesting interface
(non-Tailscale, has interesting IP address)
gains or loses an uninteresting IP address (link local or loopback).

The fact that the interface is interesting is enough for EqualFiltered
to inspect it. The fact that an IP address changed is enough for
EqualFiltered to declare that the interfaces are not equal.

But the State.String method reasonably declines to print any
uninteresting IP addresses. As a result, the network state appears
to have changed, but the stringification did not.

The String method is correct; nothing interesting happened.

This change fixes this by adding an IP address filter to EqualFiltered
in addition to the interface filter. This lets the network monitor
ignore the addition/removal of uninteresting IP addresses.

### DIFF
--- a/net/interfaces/interfaces.go
+++ b/net/interfaces/interfaces.go
@@ -345,9 +345,18 @@ func (s *State) String() string {
 	return sb.String()
 }
 
+// An InterfaceFilter indicates whether EqualFiltered should use i when deciding whether two States are equal.
+// ips are all the IPPrefixes associated with i.
+type InterfaceFilter func(i Interface, ips []netaddr.IPPrefix) bool
+
+// An IPFilter indicates whether EqualFiltered should use ip when deciding whether two States are equal.
+// ip is an ip address associated with some interface under consideration.
+type IPFilter func(ip netaddr.IP) bool
+
 // EqualFiltered reports whether s and s2 are equal,
-// considering only interfaces in s for which filter returns true.
-func (s *State) EqualFiltered(s2 *State, filter func(i Interface, ips []netaddr.IPPrefix) bool) bool {
+// considering only interfaces in s for which filter returns true,
+// and considering only IPs for those interfaces for which filterIP returns true.
+func (s *State) EqualFiltered(s2 *State, useInterface InterfaceFilter, useIP IPFilter) bool {
 	if s == nil && s2 == nil {
 		return true
 	}
@@ -364,7 +373,7 @@ func (s *State) EqualFiltered(s2 *State, filter func(i Interface, ips []netaddr.
 	}
 	for iname, i := range s.Interface {
 		ips := s.InterfaceIPs[iname]
-		if !filter(i, ips) {
+		if !useInterface(i, ips) {
 			continue
 		}
 		i2, ok := s2.Interface[iname]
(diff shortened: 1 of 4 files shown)


josharian:
I wrote this PR based entirely on what I saw in the logs and what I saw in the code. I don't know how to reproduce this locally, or write meaningful tests. Hints on that front are welcome. Or just a careful code review, I guess.

josharian:
Ping

net/interfaces/interfaces.go
```diff
@@ -402,13 +426,23 @@ func prefixesEqual(a, b []netaddr.IPPrefix) bool {
...
-// FilterInteresting reports whether i is an interesting non-Tailscale interface.
-func FilterInteresting(i Interface, ips []netaddr.IPPrefix) bool {
+// UseInterestingInterfaces reports whether i is an interesting non-Tailscale interface.
+func UseInterestingInterfaces(i Interface, ips []netaddr.IPPrefix) bool {
 	return !isTailscaleInterface(i.Name, ips) && anyInterestingIP(ips)
 }
 
-// FilterAll always returns true, to use EqualFiltered against all interfaces.
-func FilterAll(i Interface, ips []netaddr.IPPrefix) bool { return true }
+// UseInterestingIPs reports whether ip is an interesting IP address.
```
bradfitz, line 434: This name sounds really weird without knowing about type IPFilter.

Maybe it should be:

    // UseInterestingIPs returns an IPFilter that only matches ....
    func UseInterestingIPs() IPFilter {
         return ...
    }

But also: if this is now exported API, it's all the more important to document what an "interesting" IP means.

bradfitz (approved):
This extra complexity would be much more compelling with a nice new test.

But this is fine if you want.

Do you know what was actually happening? An IP was changing on an interface that was ignored?

josharian:
> Do you know what was actually happening? An IP was changing on an interface that was ignored?

An ignored IP was changing on an interface that was not ignored.

josharian:
> This extra complexity would be much more compelling with a nice new test.

Added a test that matches what I saw in the logs. I was hoping for something higher level, but it should at least illustrate the problem.

> This name sounds really weird without knowing about type IPFilter.

Fixed, I think. PTAL.

net/interfaces/interfaces.go
```diff
@@ -594,10 +629,7 @@ func anyInterestingIP(pfxs []netaddr.IPPrefix) bool {
 // should log in interfaces.State logging. We don't need to show
 // localhost or link-local addresses.
 func isInterestingIP(ip netaddr.IP) bool {
-	if ip.IsLoopback() || ip.IsLinkLocalUnicast() {
```
bradfitz, line 597: Arbitrary churn but sure :)

bradfitz (approved):
Thanks for the test!

net/interfaces/interfaces.go
josharian: I had to read it twice to make sure I understood it. (If this, then false?) This helped me, so figured it'd help the next person.

PR #3368, opened 2021-11-22
hostinfo: report when running in Docker Desktop.

DentonGentry:

### DIFF
--- a/hostinfo/hostinfo.go
+++ b/hostinfo/hostinfo.go
@@ -87,6 +87,7 @@ const (
 	AWSFargate      = EnvType("fg")
 	FlyDotIo        = EnvType("fly")
 	Kubernetes      = EnvType("k8s")
+	DockerDesktop   = EnvType("dde")
 )
 
 var envType atomic.Value // of EnvType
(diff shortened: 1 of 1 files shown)


hostinfo/hostinfo.go
```diff
@@ -87,6 +87,7 @@ const (
 	AWSFargate      = EnvType("fg")
 	FlyDotIo        = EnvType("fly")
 	Kubernetes      = EnvType("k8s")
+	DockerDesktop   = EnvType("dd")
```
bradfitz, line 90: How about "dde" if that's what the TS_HOST_ENV is, rather than keeping track of two and possibly mixing them up later?

More generally, maybe if TS_HOST_ENV is non-empty we should pass it through unchanged, so it's a general mechanism for the future?

bradfitz (approved):

hostinfo/hostinfo.go
DentonGentry: Changed to "dde"

I did think about passing through TS_HOST_ENV as is, but was concerned that sets the bar to pollute our metrics really low. As it is today someone would have to change code to send us junk HostInfo, not just set an environment variable.

bradfitz: Fair 'nuff.

PR #3369, opened 2021-11-22
build_docker.sh: use `github.com/tailscale/mkctr` instead of docker

maisem:
We recently added support for building cross platform docker containers using `docker buildx`. However, docker does cross platform builds by emulating the target arch using qemu. That is obviously very slow to build and consumes 100% of the cpu, and when run on `arm64` machines has a tendency to fail completely.

This in contrast, is a small go binary which uses `GOOS= GOARCH= go build` directly to compile the binaries and then uses [google/go-containerregistry](https://github.com/google/go-containerregistry) to create and publish the new containers based on the desired platform.

This does introduce an extra step to generate a base alpine image, however the base image is going to change less frequently than our code so I think the extra step is ok.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,6 +56,5 @@ RUN GOARCH=$TARGETARCH go install -tags=xversion -ldflags="\
       -X tailscale.com/version.GitCommit=$VERSION_GIT_HASH" \
       -v ./cmd/tailscale ./cmd/tailscaled
 
-FROM alpine:3.14
-RUN apk add --no-cache ca-certificates iptables iproute2 ip6tables
+FROM ghcr.io/tailscale/alpine-base:3.14
 COPY --from=build-env /go/bin/* /usr/local/bin/
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -1,0 +1,6 @@
+# Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+FROM alpine:3.14
+RUN apk add --no-cache ca-certificates iptables iproute2 ip6tables
--- a/build_docker.sh
+++ b/build_docker.sh
@@ -21,8 +21,15 @@ set -eu
 
 eval $(./build_dist.sh shellvars)
 
-docker build \
-  --build-arg VERSION_LONG=$VERSION_LONG \
-  --build-arg VERSION_SHORT=$VERSION_SHORT \
-  --build-arg VERSION_GIT_HASH=$VERSION_GIT_HASH \
-  -t tailscale:$VERSION_SHORT -t tailscale:latest .
+go run github.com/tailscale/mkctr@latest \
+  --base="ghcr.io/tailscale/alpine-base:3.14" \
+  --gopaths="\
+    tailscale.com/cmd/tailscale:/usr/local/bin/tailscale, \
+    tailscale.com/cmd/tailscaled:/usr/local/bin/tailscaled" \
+  --ldflags="\
+    -X tailscale.com/version.Long=${VERSION_LONG} \
+    -X tailscale.com/version.Short=${VERSION_SHORT} \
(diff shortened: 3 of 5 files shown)


go.mod
```diff
@@ -82,9 +83,14 @@ require (
 	github.com/blakesmith/ar v0.0.0-20190502131153-809d4375e1fb // indirect
 	github.com/bombsimon/wsl/v3 v3.1.0 // indirect
 	github.com/cavaliercoder/go-cpio v0.0.0-20180626203310-925f9528c45e // indirect
+	github.com/containerd/stargz-snapshotter/estargz v0.10.0 // indirect
 	github.com/daixiang0/gci v0.2.7 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/denis-tingajkin/go-header v0.3.1 // indirect
+	github.com/docker/cli v20.10.10+incompatible // indirect
+	github.com/docker/distribution v2.7.1+incompatible // indirect
+	github.com/docker/docker v20.10.10+incompatible // indirect
```
bradfitz, line 92: I don't think this cmd/mkctr tool really needs to live in tailscale/tailscale. How about a new repo at github.com/tailscale/mkctr or something?

build_docker.sh
```diff
@@ -21,8 +21,15 @@ set -eu
 
 eval $(./build_dist.sh shellvars)
 
-docker build \
-  --build-arg VERSION_LONG=$VERSION_LONG \
-  --build-arg VERSION_SHORT=$VERSION_SHORT \
-  --build-arg VERSION_GIT_HASH=$VERSION_GIT_HASH \
-  -t tailscale:$VERSION_SHORT -t tailscale:latest .
+go run ./cmd/mkctr \
```
bradfitz, line 24: go run github.com/tailscale/mkctr

During dev you can `go mod edit -replace github.com/tailscale/mkctr=$HOME/src/github.com/tailscale/mkctr` or whatever.

go.mod
maisem: yeah i was considering that too, but i thought getting initial thoughts here would be easier

maisem: https://github.com/tailscale/mkctr/pull/1

bradfitz (approved):
LGTM at least. @danderson any thoughts?

Dockerfile
```diff
@@ -56,6 +56,5 @@ RUN GOARCH=$TARGETARCH go install -tags=xversion -ldflags="\
       -X tailscale.com/version.GitCommit=$VERSION_GIT_HASH" \
       -v ./cmd/tailscale ./cmd/tailscaled
 
-FROM alpine:3.14
-RUN apk add --no-cache ca-certificates iptables iproute2 ip6tables
```
danderson, line 60: Why remove these deps? Tailscale still needs them?

Dockerfile.base
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
```
danderson, line 1: What's this file for? Does't seem used by anything else here.

Also .base includes regular alpine and .not-base includes alpine-base. ???

maisem: this is the extra step that i mentioned in the commit message, the idea is that we can construct and publish the base images independently.
so we build `ghcr.io/tailscale/alpine-base` once and then future tailscale container builds can rely on that without having to rebuild from `alpine`.

I'm open to alternate suggestions.

Dockerfile
bradfitz: technically we don't need ca-certificates or iproute2 anymore. (but doesn't hurt to keep 'em)

but definitely need iptables and ip6tables still.

Dockerfile.base
danderson: Ah, gotcha. We're probably going to want a story soon for automating these rebuilds, otherwise we're eventually going to ship CVEs that alpine's already patched. But LGTM otherwise.

danderson (approved):

PR #3370, opened 2021-11-22
cmd/tailscale: add --json for up

DentonGentry:
Print JSON to stdout containing everything needed for
authentication.

### DIFF
--- a/cmd/tailscale/cli/up.go
+++ b/cmd/tailscale/cli/up.go
@@ -6,6 +6,8 @@ package cli
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
@@ -69,6 +71,7 @@ func newUpFlagSet(goos string, upArgs *upArgsT) *flag.FlagSet {
 	upf := newFlagSet("up")
 
 	upf.BoolVar(&upArgs.qr, "qr", false, "show QR code for login URLs")
+	upf.BoolVar(&upArgs.json, "json", false, "output in JSON format (WARNING: format subject to change)")
 	upf.BoolVar(&upArgs.forceReauth, "force-reauth", false, "force reauthentication")
 	upf.BoolVar(&upArgs.reset, "reset", false, "reset unspecified settings to their default values")
 
@@ -124,6 +127,7 @@ type upArgsT struct {
 	authKeyOrFile          string // "secret" or "file:/path/to/secret"
 	hostname               string
 	opUser                 string
+	json                   bool
 }
 
 func (a upArgsT) getAuthKey() (string, error) {
@@ -140,6 +144,33 @@ func (a upArgsT) getAuthKey() (string, error) {
 }
 
 var upArgs upArgsT
+
+// Fields output when `tailscale up --json` is used. Two JSON blocks will be output.
+//
+// When "tailscale up" is run it first outputs a block with AuthURL and QR populated,
+// providing the link for where to authenticate this client. BackendState would be
+// valid but boring, as it will almost certainly be "NeedsLogin". Error would be
+// populated if something goes badly wrong.
+//
+// When the client is authenticated by having someone visit the AuthURL, a second
+// JSON block will be output. The AuthURL and QR fields will not be present, the
+// BackendState and Error fields will give the result of the authentication.
+// Ex:
+// {
+//    "AuthURL": "https://login.tailscale.com/a/0123456789abcdef",
+//    "QR": "data:image/png;base64,0123...cdef"
+//    "BackendState": "NeedsLogin"
+// }
+// {
+//    "BackendState": "Running"
+// }
+//
+type upOutputJSON struct {
+	AuthURL      string `json:",omitempty"` // Authentication URL of the form https://login.tailscale.com/a/0123456789
+	QR           string `json:",omitempty"` // a DataURL (base64) PNG of a QR code AuthURL
+	BackendState string `json:",omitempty"` // name of state like Running or NeedsMachineAuth
+	Error        string `json:",omitempty"` // description of an error
+}
 
 func warnf(format string, args ...interface{}) {
 	printf("Warning: "+format+"\n", args...)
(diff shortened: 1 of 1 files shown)


cmd/tailscale/cli/up.go
```diff
@@ -451,15 +455,37 @@ func runUp(ctx context.Context, args []string) error {
 		}
 		if url := n.BrowseToURL; url != nil && printAuthURL(*url) {
 			printed = true
-			fmt.Fprintf(Stderr, "\nTo authenticate, visit:\n\n\t%s\n\n", *url)
-			if upArgs.qr {
+			if upArgs.json {
+				type fields struct {
```
bradfitz, line 459: if we're going to support this as an interface, it's worth making this a top-level type with docs on what the fields mean.

cmd/tailscale/cli/up.go
```diff
@@ -451,15 +455,37 @@ func runUp(ctx context.Context, args []string) error {
...
+					AuthURL string
+					QR      string `json:",omitempty"`
```
bradfitz, line 461: is this always included, or only with the --qr flag?

cmd/tailscale/cli/up.go
```diff
@@ -451,15 +455,37 @@ func runUp(ctx context.Context, args []string) error {
 		}
 		if url := n.BrowseToURL; url != nil && printAuthURL(*url) {
 			printed = true
-			fmt.Fprintf(Stderr, "\nTo authenticate, visit:\n\n\t%s\n\n", *url)
-			if upArgs.qr {
+			if upArgs.json {
+				type fields struct {
+					AuthURL string
```
bradfitz, line 460: is this omitempty too?

What is the signal that the state has changed and we're now successfully up?

cmd/tailscale/cli/up.go
bradfitz: OH! It's a data URL, not the text output. Also worth mentioning.

cmd/tailscale/cli/up.go
DentonGentry: Moved to top level

cmd/tailscale/cli/up.go
DentonGentry: Documented the field.

cmd/tailscale/cli/up.go
DentonGentry: Added omitempty.

It otherwise behaves the same as tailscale up, blocking until the authentication is done. I'd expected to have the parent read from the child process until it receives the JSON, do whatever with the AuthURL and QR image, and then block on wait until the child exits.

cmd/tailscale/cli/up.go
```diff
@@ -138,6 +142,12 @@ func (a upArgsT) getAuthKey() (string, error) {
 
 var upArgs upArgsT
 
+// Fields output when `tailscale up --json` is used.
+type upJsonOutput struct {
```
bradfitz, line 146: upJSONOutput or upJSON or upOutputJSON but no mixed case initialisms per https://github.com/golang/go/wiki/CodeReviewComments#initialisms

cmd/tailscale/cli/up.go
```diff
@@ -138,6 +142,12 @@ func (a upArgsT) getAuthKey() (string, error) {
...
+	AuthURL string `json:",omitempty"` // Authentication URL of the form https://login.tailscale.com/a/0123456789
+	QR      string `json:",omitempty"` // a DataURL (base64) PNG of a QR code AuthURL
+}
```
bradfitz, line 149: I think we want a final message that says success rather than assuming that EOF means success.

Likewise: failures, like bogus auth codes or bogus tags.

DentonGentry: Added:

```
    {
      "AuthURL": "https://login.tailscale.com/a/0123456789",
      "QR": "data:image/png;base64,iV...QmCC"
    }
    
 ...time passes, until I click on the login link...
 
    {
      "BackendState": "Running"
    }
```

It can end with BackendState Running or NeedsAuthorization. If I use invalid tags it ends up as Running but the admin panel says the tags are invalid. I don't think that is being communicated back to the client as part of the authentication flow.

cmd/tailscale/cli/up.go
DentonGentry: Done.

DentonGentry:
This is working now in the intended client platform.

cmd/tailscale/cli/up.go
```diff
@@ -546,6 +580,20 @@ func runUp(ctx context.Context, args []string) error {
 	}
 }
 
+func printUpDoneJSON(state, errorString string) {
+	type upDoneJSON struct {
```
bradfitz, line 584: let's merge these two types together... they're all one thing the client will parse. I know you're not using Go, but imagine you were using Go... you'd want one type to unmarshal into.

Then use omitempty annotations on the fields that aren't always present including BackendState.

DentonGentry: Done.

cmd/tailscale/cli/up.go
```diff
@@ -609,6 +664,16 @@ func runUp(ctx context.Context, args []string) error {
 	}
 }
 
+func printUpDoneJSON(state, errorString string) {
```
bradfitz, line 667: change `state` from type `string` to `ipn.State`?

cmd/tailscale/cli/up.go
```diff
@@ -609,6 +664,16 @@ func runUp(ctx context.Context, args []string) error {
...
+	js := &upOutputJSON{BackendState: state, Error: errorString}
+	data, err := json.MarshalIndent(js, "", "  ")
```
bradfitz, line 669: "\t" for indent? up to you, but fewer bytes on the wire and more semantically an indent (Go bias, admittedly)

bradfitz (approved):

cmd/tailscale/cli/up.go
DentonGentry: Done.

cmd/tailscale/cli/up.go
DentonGentry: Done.

Issue #3373, opened 2021-11-22
Enable HSTS on DERP servers


danderson:
### Tell us about your idea!

DERP servers need HSTS.

### What are you trying to do?

The automated security scanners we use for compliance things keep panicking when they discover that our DERP servers, which do not serve anything of value over HTTPS to _browsers_, don't set HSTS headers. Since HSTS has no effect on non-browser clients, and those are the only ones talking to DERP servers for non-debug purposes, we might as well enable HSTS to quiet the automation, rather than have to continually mark them as "exceptions" requiring paperwork to justify.

bradfitz:
This would break plain HTTP access to http://derp10.ts.tailscale.com/debug/varz access over Tailscale from browsers, no?

danderson:
It shouldn't, because you don't access derp10.ts.tailscale.com over HTTPS (no certs have that name), so HSTS will never kick in. URLs need to be reachable over HTTPS for HSTS to kick in, so it'll only happen for the public URLs (e.g. https://derp10.tailscale.com).

But to be sure, I'm deploying this the same way I did HSTS on our other properties: start with a 10min HSTS timeout, so we can look around and see if anything breaks.

bradfitz:
Oh, so it's not wildcard `*.tailscale.com`? I'm confusing an earlier proposal I guess.

danderson:
Yeah, HSTS works based on whatever domain you use to access the site. What I'm doing here is making visits to derpX.tailscale.com say "force HTTPS for derpX.tailscale.com and *.derpX.tailscale.com".

The tricky one you're thinking of is HSTS on https://tailscale.com, because if I `includeSubdomains` there, then it would force HTTPS on everything in our domain, including the semi-private ts.tailscale.com tree. I'm not doing that in this change :)

danderson:
Deployed with a 10min HSTS lifetime to derp10.tailscale.com as a canary. If all goes well, will crank the expiry up to 2y (HSTS best practice) and roll out to all derps.

/cc @mayakacz

PR #3374, opened 2021-11-22
cmd/derper: enable HSTS when serving over HTTPS.

danderson:
Starting with a short lifetime, to verify nothing breaks.

Updates #3373

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -235,6 +235,14 @@ func main() {
 			cert.Certificate = append(cert.Certificate, s.MetaCert())
 			return cert, nil
 		}
+		httpsrv.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Security scanners get cranky when HTTPS sites don't set
+			// HSTS. Set it even though derper doesn't really serve
+			// anything of interest to browsers (and API clients like
+			// tailscale don't obey HSTS).
+			w.Header().Set("Strict-Transport-Security", "max-age=600; includeSubDomains")
+			mux.ServeHTTP(w, r)
+		})
 		go func() {
 			port80srv := &http.Server{
 				Addr:        net.JoinHostPort(listenHost, "80"),


bradfitz (approved):

cmd/derper/derper.go
```diff
@@ -235,6 +235,14 @@ func main() {
 			cert.Certificate = append(cert.Certificate, s.MetaCert())
 			return cert, nil
 		}
+		httpsrv.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Security scanners get cranky when HTTPS sites don't set
+			// HSTS. Set it even though derper doesn't really serve
+			// anything of interest to browsers (and API clients like
+			// tailscale don't obey HSTS).
+			w.Header().Set("Strict-Transport-Security", "max-age=600; includeSubDomains")
```
bradfitz, line 243: only do this is r.TLS != nil ?

Line 249 won't use this, though, right?

bradfitz: (Do the HSTS scanners expect to see it on HTTP responses too?)

If so, you'll need to teach this code for which domain names to include it.

danderson: This is already in an `if serveTLS` block.

And yes, it wraps the TLS handler only, so port 80 continues to serve without HSTS directives (and besides, HSTS says browsers must ignore HSTS directives sent over HTTP anyway, because it's a DoS vector).

bradfitz: LGTM

danderson: To clarify: no, HSTS headers MUST NOT be set on HTTP responses. And if HSTS headers are present in HTTP responses, browsers MUST ignore them, because otherwise a mitm could force-enable HSTS on your site by editing your HTTP traffic, which would lead to a DoS if your site doesn't do HTTPS.

So, I think this change is correct: when answering HTTPS requests, set HSTS headers; otherwise, don't.

PR #3375, opened 2021-11-22
net/portmapper: PMP probe failures are not unexpected

josharian:
There are lots of lines in the logs of the form:

portmapper: unexpected PMP probe response: {OpCode:128 ResultCode:3
SecondsSinceEpoch:NNN MappingValidSeconds:0 InternalPort:0
ExternalPort:0 PublicAddr:0.0.0.0}

ResultCode 3 here means a network failure, e.g. the NAT box itself has
not obtained a DHCP lease. This is not an indication that something
is wrong in the Tailscale client, so change the wording here
to reflect that.

### DIFF
--- a/net/portmapper/portmapper.go
+++ b/net/portmapper/portmapper.go
@@ -798,7 +798,12 @@ func (c *Client) Probe(ctx context.Context) (res ProbeResult, err error) {
 				c.logf("unexpected PCP probe response: %+v", pres)
 			}
 			if pres, ok := parsePMPResponse(buf[:n]); ok {
-				if pres.OpCode == pmpOpReply|pmpOpMapPublicAddr && pres.ResultCode == pmpCodeOK {
+				if pres.OpCode != pmpOpReply|pmpOpMapPublicAddr {
+					c.logf("unexpected PMP probe response opcode: %+v", pres)
+					continue
+				}
+				switch pres.ResultCode {
+				case pmpCodeOK:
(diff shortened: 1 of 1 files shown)


net/portmapper/portmapper.go
```diff
@@ -808,7 +808,7 @@ func (c *Client) Probe(ctx context.Context) (res ProbeResult, err error) {
 					c.mu.Unlock()
 					continue
 				}
-				c.logf("unexpected PMP probe response: %+v", pres)
+				c.logf("PMP probe failed: %+v", pres)
```
bradfitz, line 811: Line 801 lists the type we're expecting.

If there are more types we're expecting, I'd prefer to add new lines to recognize other things (even if we do nothing and log nothing).

But I'd like to keep this line for learning about even weirder things.

bradfitz: I mean, this is also fine as-is (it's just rewording it), but maybe we want to recognize code 3 or whatever and log a specific message for that?

bradfitz (approved):

bradfitz:
> net/portmapper: PMP probe failures are not unexpected

Not the usual verb tense.

net/portmapper/portmapper.go
josharian: Done. I kept all the detail, in case we want to do analytics about why NAT-PMP gets rejected.

bradfitz (approved):

PR #3376, opened 2021-11-22
hostinfo: Change Windows implementation to directly query version inf…

dblohm7:
…ormation using API and registry

We replace the cmd.exe invocation with RtlGetNtVersionNumbers for the first
three fields. On Windows 10+, we query for the fourth field which is available
via the registry.

The fourth field is not really documented anywhere; Firefox has been querying
it successfully since Windows 10 was released, so we can be pretty confident in
its longevity at this point.


hostinfo/hostinfo_windows.go
```diff
@@ -21,19 +22,37 @@ func osVersionWindows() string {
...
-	s = strings.TrimPrefix(s, "Microsoft Windows [")
-	s = strings.TrimSuffix(s, "]")
-
-	// "Version 10.x.y.z", with "Version" localized. Keep only stuff after the space.
-	if sp := strings.Index(s, " "); sp != -1 {
-		s = s[sp+1:]
+	major, minor, build := windows.RtlGetNtVersionNumbers()
+	s := fmt.Sprintf("%d.%d.%d", major, minor, build)
+	// Windows 11 still uses 10.0 as its (major.minor) internally
+	if major == 10 && minor == 0 {
```
bradfitz, line 28: do you need to check minor == 0 or is major == 10 enough?

bradfitz (approved):

bradfitz:
lowercase verb "change" after colon in commit message

hostinfo/hostinfo_windows.go
dblohm7: Here's the dilemma that I've had in mind: as an undocumented registry value, this value isn't really formalized anywhere, so it is unclear whether it will work beyond the 10.0 major.minor pair. I suppose if it fails, it's just going to fail silently, so it doesn't matter that much. What do you think, @bradfitz ?

bradfitz: Well, it's not like you're panicking the binary on failure. You're just conditionally appending it to the string, so I think it's fine to always do (well, when major == 10) and append it on err == nil.

dblohm7: SGTM.

Issue #3360, opened 2021-11-19
wgengine/router test fails on ppc64le 


panekj:
### What is the issue?

`wgengine/router` test fails on ppc64le due to `netlink` dependency

```
--- FAIL: TestDebugListRules (0.00s)
    --- FAIL: TestDebugListRules/v4 (0.00s)
        router_linux_test.go:796: address family not supported by protocol
FAIL
FAIL	tailscale.com/wgengine/router	0.037s
```

https://gitlab.alpinelinux.org/panekj/aports/-/jobs/542353

### Steps to reproduce

Run `go test ./...` on ppc64le architecture

### Are there any recent changes that introduced the issue?

https://github.com/tailscale/tailscale/commit/d36c0d3566ee097fe69882771835a2f1434d0fbc

### OS

Linux

### OS version

Alpine Linux 3.15_rc4 (Edge)

### Tailscale version

1.18.0

bradfitz:
Does the test pass in your CI on other platforms?

panekj:
Yes, you can see [here](https://gitlab.alpinelinux.org/panekj/aports/-/pipelines/100233) all other arches pass successfully

bradfitz:
Well that's weird.

Maybe the netlink package has a bug on that platform.

I guess for now we can make that arch fall back to the code path that runs the "ip" command instead. (We kept that code in out of paranoia)

bradfitz:
@panekj, do you have the Linux kernel config for your ppc64le builder machine? Maybe it just lacks policy routing support altogether.

Does it have IP_MULTIPLE_TABLES set?

bradfitz:
1.18.1 is out now. Try again?

panekj:
@bradfitz that seems to fix the issue, but it also broke s390x now
```
--- FAIL: TestIgnoreLocallyBoundPorts (0.00s)
    portlist_test.go:43: didn't expect to find test's localhost ephemeral port
FAIL
FAIL	tailscale.com/portlist	0.031s
```

bradfitz:
> @bradfitz that seems to fix the issue

Thanks for confirming! I'll close this bug.

> but it also broke s390x now

We didn't change anything about s390x or the portlist code in 1.18.1 or even between 1.16.0 and 1.18.0.

[...]

PR #3378, opened 2021-11-22
wgengine/router: demote TestDebugListRules fail to skip

bradfitz:
Updates #3360


josharian:
Skip only on ppc?

josharian (approved):

bradfitz:
> Skip only on ppc?

No. It can be on any kernel without it. It just happens to be that that machine probably lacks kernel support for what we're trying to list.

The point of Debug tests is to just t.Log stuff, not fail.

Issue #879, opened 2020-10-30
Windows 8.1:  tshttpproxy: winhttp: GetProxyForURL("...log.tailscale.io..."): The parameter is incorrect./0x57


apenwarr:
Steps:
- Install tailscale 1.1.1313 on fresh Windows 8.1

Expected:
- It works

Actual:
- Frontend starts, and seems to connect to backend, but never manages to start up completely (windows menus neither show "please restart the backend" nor the Login... button)
- I can't find any backend log files in c:\windows.
- Uploaded logs show the following, infinitely repeating, and with no backoff timer: [logid d1de47d0c1c]

```
10-30,07:47:23.457 tshttpproxy: winhttp: GetProxyForURL("https://log.tailscale.io/c/tailnode.log.tailscale.io/<...>"): The parameter is incorrect./0x57
10-30,07:47:23.545 tshttpproxy: winhttp: GetProxyForURL("https://log.tailscale.io/c/tailnode.log.tailscale.io/<...>"): The parameter is incorrect./0x57
```

I see at least three possible bugs here:

1) The logs are uploading even though there's a GetProxyForURL error
2) We are requesting GetProxyForURL far too often
3) There is no backoff even though there's a GetProxyForURL error
4) There shouldn't be a GetProxyForURL error in the first place.

I didn't observe this problem on Windows 7 or Windows 10. Windows 8.1 users can still fall back to v1.0.5 until this is fixed, so I won't set it as a v1.2.0 release blocker, even though it's a regression.

bradfitz:
We're plugging into Go's https://golang.org/pkg/net/http/#Transport.Proxy to specify a function to return the proxy per HTTP request. `tshttpproxy_windows.go` already caches failures, but only for certain error codes. `The parameter is incorrect./0x57` is not one of the ones we recognize.

The errors are harmless; the result of any error from winhttp is that we just don't use a proxy.

We're seeing so many error messages because we're making that many HTTP requests (for each chunk of logs). But we're making that many HTTP requests because we're logging about the failure to make those error messages, which causes more HTTP requests. I think that issue, which isn't exactly one of the 4 bugs you listed, is the bigger problem.

bradfitz:
Verified the change works:

<img width="1092" alt="Screen Shot 2020-11-03 at 9 42 58 AM" src="https://user-images.githubusercontent.com/2621/98021313-0aa9fd80-1db9-11eb-815c-0c70a8980cb9.png">

[...]

PR #3381, opened 2021-11-22
net/tshttpproxy: use correct size for Windows BOOL argument

josharian:
The Windows BOOL type is an int32. We were using a bool,
which is a one byte wide. This could be responsible for the
ERROR_INVALID_PARAMETER errors we were seeing for calls to
WinHttpGetProxyForUrl.

We manually checked all other existing Windows syscalls
for similar mistakes and did not find any.

Updates #879

Co-authored-by: Aaron Klotz <aaron@tailscale.com>


bradfitz:
OMG YES! Good find!

bradfitz:
This has haunted me for ages.

bradfitz (approved):

Issue #2716, opened 2021-08-26
teach wgengine/bench to disco


danderson:
wgengine/bench is a benchmark suite that compares bare UDP to Tailscale. It does so by hand-building magicsocks and plumbing them together, but does so without DERPs or discovery keys - IOW, by relying on the legacy NAT traversal codepath that I'm in the middle of deleting.

If this bench suite is still useful to current performance work, refactor this and a bunch of the magicsock test suite so that it sets up a proper disco-based test.

DentonGentry:
I believe this was fixed by https://github.com/tailscale/tailscale/pull/2711

Barring further comment, we'll expect to close this during the next regular bug scrub.

PR #3390, opened 2021-11-23
wgengine/bench: attempt to fix the wireguard benchmark

bradfitz:
Not there yet. Too much copy/paste code in here. It needs
to be refactored. (or factored at all)

Updates #2716


josharian (approved):
This all looks plausible. Get it building and passing and I'm good with it. Refactoring and cleanup would be nice, but supererogatory.

Issue #1713, opened 2021-04-16
Forward DNS traffic to exit node


darshinimashar:
Our user suggested the feature for 1.8+ , When using the MagicDNS/100.100.100.100 local DNS resolver within Tailscale, AND if an exit node is currently configured to be active, then the local tailscaled resolver should forward DNS requests to the exit node tailscaled resolver, and allow the exit node to resolve the name on behalf of the client. This would generally allow geo-DNS based records to return an IP that is “closer” to the exit node (vs. “closer” to the client). 

***** One could argue this is actually a security and/or privacy requirement, as it further masks the location of the actual client and furthers the appearance that the exit node represents the actual client.

<img src="https://frontapp.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_319zl)

danderson:
Yes, this is the plan, and I hope to get it working for 1.8. It looks like I never filed a bug for it, so this is a great tracking bug.

DentonGentry:
This feature is present in 1.19.x builds in https://pkgs.tailscale.com/unstable/ and will be in the 1.20 release build.

bradfitz:
This shipped in 1.20.0.

PR #3392, opened 2021-11-23
net/dns/resolver, ipn/ipnlocal: wire up peerapi DoH server to DNS forwarder

bradfitz:
Updates #1713

### DIFF
--- a/ipn/ipnlocal/peerapi.go
+++ b/ipn/ipnlocal/peerapi.go
@@ -29,6 +29,7 @@ import (
 	"unicode"
 	"unicode/utf8"
 
+	"golang.org/x/net/dns/dnsmessage"
 	"inet.af/netaddr"
 	"tailscale.com/client/tailscale/apitype"
 	"tailscale.com/hostinfo"
@@ -767,6 +768,8 @@ func (h *peerAPIHandler) replyToDNSQueries() bool {
 	return h.isSelf || h.ps.b.OfferingExitNode()
 }
 
+// handleDNSQuery implements a DoH server (RFC 8484) over the peerapi.
+// It's not over HTTPS as the spec dictates, but rather HTTP-over-WireGuard.
 func (h *peerAPIHandler) handleDNSQuery(w http.ResponseWriter, r *http.Request) {
 	if h.ps.resolver == nil {
 		http.Error(w, "DNS not wired up", http.StatusNotImplemented)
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/peerapi.go
```diff
@@ -776,13 +779,40 @@ func (h *peerAPIHandler) handleDNSQuery(w http.ResponseWriter, r *http.Request)
...
+	}
 	if publicError != "" {
 		http.Error(w, publicError, http.StatusBadRequest)
 		return
 	}
-	// TODO(bradfitz): owl.
-	fmt.Fprintf(w, "## TODO: got %d bytes of DNS query", len(q))
+	ctx, cancel := context.WithTimeout(r.Context(), 5*time.Second)
+	defer cancel()
+	res, err := h.ps.resolver.ForwardSync(ctx, q, h.remoteAddr)
```
danderson, line 797: I don't think this is right. We want to hand peerapi queries to the OS for resolution, not to our internal forwarder. Depending on the DNS configuration, it doesn't have the full DNS config.

danderson (changes requested):

ipn/ipnlocal/peerapi.go
bradfitz: I don't think asking the OS is quite the right level. We don't want them to loop back into us and get our own MagicDNS names. We want "internet only" DNS.

That doesn't quite matter now, but will if/when we allow sharing exit nodes. The person using the exit node will have their DNS set to 100.100.100.100 which will first add any MagicDNS names for their tailnet, but then they'll forward to peerapi DoH which should not add any more MagicDNS names.

But, yeah, I guess we need more here in the case of an exit node that that's using systemd-resolved and doesn't have global/fallback DNS configured.

bradfitz: Updated the PR.

Renamed `ForwardSync` to `HandleExitNodeDNSQuery` to better reflect what it's trying to do.

And I handled the systemd-resolved Split DNS case on Linux at least now, even if it not wonderfully. I left some TODOs to clean it up.

My goal here is to get enough working on the peerapi end that I can then work on the client side.

ipn/ipnlocal/peerapi.go
```diff
@@ -776,13 +779,40 @@ func (h *peerAPIHandler) handleDNSQuery(w http.ResponseWriter, r *http.Request)
...
+			q = dnsQueryForName(name, r.FormValue("t"))
+		}
+	}
 	if publicError != "" {
 		http.Error(w, publicError, http.StatusBadRequest)
 		return
 	}
-	// TODO(bradfitz): owl.
-	fmt.Fprintf(w, "## TODO: got %d bytes of DNS query", len(q))
+	ctx, cancel := context.WithTimeout(r.Context(), 5*time.Second)
```
crawshaw, line 795: why 5? this is longer than a typical dns query timeout, no?

ipn/ipnlocal/peerapi.go
```diff
@@ -817,3 +847,86 @@ func dohQuery(r *http.Request) (dnsQuery []byte, publicErr string) {
...
+		if err == dnsmessage.ErrSectionDone {
+			break
+		}
+		if err != nil {
+			return err
+		}
+		if h.Class != dnsmessage.ClassINET {
+			continue
+		}
+		switch h.Type {
```
crawshaw, line 907: no SRV records? any other obvious types missing?

net/dns/resolver/tsdns.go
```diff
@@ -298,6 +299,58 @@ func (r *Resolver) NextResponse() (packet []byte, to netaddr.IPPort, err error)
...
+			// TODO(bradfitz): this is a lazy assumption. Do better, and
+			// maybe move the HandleExitNodeDNSQuery method to the dns.Manager
+			// instead? But this works for now.
+			err = r.forwarder.forwardWithDestChan(ctx, packet{q, from}, ch, resolverAndDelay{
+				name: dnstype.Resolver{
+					Addr: "127.0.0.1:53",
+				},
+			})
+		default:
+			// TODO(bradfitz): if we're on an exit node
```
crawshaw, line 326: a potential alternative is doing the query and stripping out magicdns answers (because we know the search domains list), which might let us be more generic about how we ask questions of the OS?

crawshaw (approved):
it's an improvement over what we've got, though some thoughts in the comments,

ipn/ipnlocal/peerapi.go
bradfitz: This isn't prod code. This is only the debug pretty printing path for us to test it with curl.

ipn/ipnlocal/peerapi.go
bradfitz: arbitrary. You have a number you like?

net/dns/resolver/tsdns.go
bradfitz: not really, because it's not like OS APIs provide an API like "give me a DNS query packet and I'll give you a DNS answer packet".

instead, they have gethostbyname and such.  So we're going to need to do some OS-specific mapping of DNS query packets to OS APIs anyway.

ipn/ipnlocal/peerapi.go
crawshaw: ha no, i have no answers! but add an `// arbitrary` so that future archeologists don't spend too many hours trying to decipher your sage timeout choices

bradfitz: Done:

```
        // Some timeout that's short enough to be noticed by humans 
        // but long enought that it's longer than real DNS timeouts. 
        const arbitraryTimeout = 5 * time.Second

        ctx, cancel := context.WithTimeout(r.Context(), arbitraryTimeout)
```

PR #3393, opened 2021-11-23
net/tstun: remove "self disco out packet" logging

josharian:
When this happens, it is incredibly noisy in the logs.
It accounts for about a third of all remaining
"unexpected" log lines from a recent investigation.

It's not clear that we know how to fix this,
we have a functioning workaround,
and we now have a (cheap and efficient) metric for this,
so we can still detect and measure its occurrence.

Given all that, drop the log line.


bradfitz:
I'm afraid humans will miss it debugging support tickets if we don't include it sometimes.

Can we log it once every minute at least?

Note that there's a second place in that file with such an unexpected log.

josharian:
Done. PTAL

bradfitz (approved):

PR #3394, opened 2021-11-23
cmd/tailscale: add ip -1 flag

josharian:
- cmd/tailscale: improve ip subcommand docs
- cmd/tailscale: add ip -1 flag

### DIFF
--- a/cmd/tailscale/cli/ip.go
+++ b/cmd/tailscale/cli/ip.go
@@ -18,9 +18,9 @@ import (
 
 var ipCmd = &ffcli.Command{
 	Name:       "ip",
-	ShortUsage: "ip [-4] [-6] [peername]",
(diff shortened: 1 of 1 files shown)


josharian:
If this goes in, where do I put release notes?

bradfitz (approved):

bradfitz:
> And we do we have a general process for remembering to update docs (to use this new flag) after a new release?

We've started to sometimes do the whole RELNOTE= thing in the commit mesage.

DentonGentry:
If you put `RELNOTE=tailscale ip adds a -1 flag` in the commit message, we'll pick it up in the release notes (it is a manual process so a human does look at it before it goes in).

If that RELNOTE message also mentions other docs to update, we're likely to do so.

josharian:
Added RELNOTE. Thanks.

PR #3395, opened 2021-11-23
ipn/ipnlocal: fix logging

josharian:
We were missing an argument here.
Also, switch to %q, in case anything weird
is happening with these strings.

Updates tailscale/corp#461


bradfitz:
LGTM

PR #3396, opened 2021-11-23
net/dns: handle comments in resolv.conf

josharian:
Currently, comments in resolv.conf cause our parser to fail,
with error messages like:

ParseIP("192.168.0.100 # comment"): unexpected character (at " # comment")

Fix that.

Noticed while looking through logs.

### DIFF
--- a/net/dns/direct.go
+++ b/net/dns/direct.go
@@ -56,8 +56,11 @@ func readResolv(r io.Reader) (config OSConfig, err error) {
 		}
 
 		if strings.HasPrefix(line, "nameserver") {
-			nameserver := strings.TrimPrefix(line, "nameserver")
-			nameserver = strings.TrimSpace(nameserver)
+			s := strings.TrimPrefix(line, "nameserver")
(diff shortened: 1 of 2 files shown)


net/dns/direct.go
```diff
@@ -50,6 +50,10 @@ func readResolv(r io.Reader) (config OSConfig, err error) {
 	scanner := bufio.NewScanner(r)
 	for scanner.Scan() {
 		line := strings.TrimSpace(scanner.Text())
+		i := strings.IndexByte(line, '#')
+		if i >= 0 {
+			line = line[:i]
+		}
 
 		if strings.HasPrefix(line, "nameserver") {
```
bradfitz, line 58: so this matches "nameserver1.2.3.4" without a space, eh? :)

bradfitz (approved):

net/dns/direct.go
josharian: picky picky.

done. (nice to have tests!)

PR #3401, opened 2021-11-24
cmd/derper: set Content-Security-Policy on DERPs.

danderson:
It's a basic "deny everything" policy, since DERP's HTTP
server is very uninteresting from a browser POV. But it
stops every security scanner under the sun from reporting
"dangerously configured" HTTP servers.

Updates tailscale/corp#3119

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -236,11 +236,18 @@ func main() {
 			return cert, nil
 		}
 		httpsrv.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			// Security scanners get cranky when HTTPS sites don't set
-			// HSTS. Set it even though derper doesn't really serve
-			// anything of interest to browsers (and API clients like
-			// tailscale don't obey HSTS).
+			// Set HTTP headers to appease automated security scanners.
+			//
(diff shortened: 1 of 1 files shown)


bradfitz (approved):

cmd/derper/derper.go
```diff
@@ -236,11 +236,18 @@ func main() {
...
+			// Security automation gets cranky when HTTPS sites don't
+			// set HSTS, and when they don't specify a content
+			// security policy for XSS mitigation.
+			//
+			// DERP's HTTP interface is only ever used for debug
+			// access (for which trivial safe policies work just
+			// fine), and by DERP clients which don't obey any of
+			// these browser-centric headers anyway.
 			w.Header().Set("Strict-Transport-Security", "max-age=63072000; includeSubDomains")
+			w.Header().Set("Content-Security-Policy", "default-src 'none'; frame-ancestors 'none'; form-action 'none'; base-uri 'self'; block-all-mixed-content; plugin-types 'none'")
```
bradfitz, line 250: this won't break the websocket CORS access, will it?

Doesn't look like it.

danderson: Shouldn't, these policies are all about what a browser does while it's parsing/rendering HTML and running javascript.

PR #3403, opened 2021-11-24
net/dns: make exit node DNS ask OSConfigurator for backup resolvers

bradfitz:
Updates #1713

/cc @crawshaw

### DIFF
--- a/net/dns/direct.go
+++ b/net/dns/direct.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 
 	"inet.af/netaddr"
+	"tailscale.com/types/dnstype"
 	"tailscale.com/types/logger"
 	"tailscale.com/util/dnsname"
 )
@@ -181,6 +182,24 @@ func (m *directManager) readResolvFile(path string) (OSConfig, error) {
 		return OSConfig{}, err
 	}
 	return readResolv(bytes.NewReader(b))
+}
+
+func (m *directManager) GetExitNodeForwardResolver() (ret []dnstype.Resolver, retErr error) {
+	for _, filename := range []string{backupConf, resolvConf} {
+		if oc, err := m.readResolvFile(filename); err == nil {
+			for _, ip := range oc.Nameservers {
+				if ip != netaddr.IPv4(100, 100, 100, 100) {
+					ret = append(ret, dnstype.Resolver{Addr: netaddr.IPPortFrom(ip, 53).String()})
+				}
+			}
+			if len(ret) > 0 {
+				return ret, nil
+			}
+		} else if !os.IsNotExist(err) && retErr == nil {
+			retErr = err
+		}
+	}
+	return nil, retErr
 }
 
 // ownedByTailscale reports whether /etc/resolv.conf seems to be a
--- a/net/dns/manager.go
+++ b/net/dns/manager.go
@@ -62,6 +62,11 @@ func (m *Manager) Set(cfg Config) error {
 	if err != nil {
 		return err
 	}
+	exitNodeBackupResolvers, err := m.os.GetExitNodeForwardResolver()
+	if err != nil {
+		return err
+	}
+	rcfg.ExitNodeBackupResolvers = exitNodeBackupResolvers
 
 	m.logf("Resolvercfg: %v", logger.ArgWriter(func(w *bufio.Writer) {
 		rcfg.WriteToBufioWriter(w)
--- a/net/dns/osconfig.go
+++ b/net/dns/osconfig.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 
 	"inet.af/netaddr"
+	"tailscale.com/types/dnstype"
 	"tailscale.com/util/dnsname"
 )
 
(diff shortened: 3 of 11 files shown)


net/dns/resolved.go
```diff
@@ -332,6 +333,10 @@ func (m *resolvedManager) GetBaseConfig() (OSConfig, error) {
 	return OSConfig{}, ErrGetBaseConfigNotSupported
 }
 
+func (m *resolvedManager) GetExitNodeForwardResolver() ([]dnstype.Resolver, error) {
+	return []dnstype.Resolver{{Addr: "127.0.0.53:53"}}, nil
```
crawshaw, line 337: Here's hoping no-one ever sets `DNSStubListener=no` in resolved.conf.

crawshaw (approved):

net/dns/resolved.go
bradfitz: That case will be fixed in a follow-up CL that also makes exit node DNS proxy work on Windows: using gethostbyname etc (via Go net package DNS lookups).

danderson: FWIW: nobody runs resolved and disables the stub listener. It'd break too much stuff.

net/dns/manager.go
```diff
@@ -62,6 +62,11 @@ func (m *Manager) Set(cfg Config) error {
 	if err != nil {
 		return err
 	}
+	exitNodeBackupResolvers, err := m.os.GetExitNodeForwardResolver()
```
danderson, line 65: Put this in compileconfig, that's where cfg gets split into rcfg and ocfg.

net/dns/osconfig.go
```diff
@@ -18,20 +19,35 @@ type OSConfigurator interface {
...
 	// after SetDNS has been called to set a Tailscale configuration.
 	// Only works when SupportsSplitDNS=false.
-
+	//
 	// Implementations that don't support getting the base config must
 	// return ErrGetBaseConfigNotSupported.
 	GetBaseConfig() (OSConfig, error)
+
+	// GetExitNodeForwardResolver returns the resolver(s) that should
+	// be used as a fallback for the exit node's DNS-over-HTTP peerapi
```
danderson, line 40: Not a fallback, it's where peerapi requests should go always. Whether global resolvers are configured or not is irrelevant.

net/dns/direct.go
```diff
@@ -183,6 +184,24 @@ func (m *directManager) readResolvFile(path string) (OSConfig, error) {
 	return readResolv(bytes.NewReader(b))
 }
 
+func (m *directManager) GetExitNodeForwardResolver() (ret []dnstype.Resolver, retErr error) {
+	for _, filename := range []string{backupConf, resolvConf} {
+		if oc, err := m.readResolvFile(filename); err == nil {
+			for _, ip := range oc.Nameservers {
+				if ip != netaddr.IPv4(100, 100, 100, 100) {
```
danderson, line 191: No need to specialcase quad 100. If that's what the exit node uses for DNS, it's perfectly fine for peerapi to loop back into it.

net/dns/osconfig.go
```diff
@@ -90,3 +106,16 @@ func (a OSConfig) Equal(b OSConfig) bool {
...
 // doesn't support reading the underlying configuration out of the OS.
 var ErrGetBaseConfigNotSupported = errors.New("getting OS base config is not supported")
+
+func getExitNodeForwardResolverFromBaseConfig(o OSConfigurator) (ret []dnstype.Resolver, retErr error) {
+	oc, err := o.GetBaseConfig()
+	if err != nil {
+		return nil, err
+	}
+	for _, ip := range oc.Nameservers {
+		if ip != netaddr.IPv4(100, 100, 100, 100) {
```
danderson, line 116: Same, specialcase not needed. peerapi is not in the resolution path for quad-100, therefore sending from peerapi to quad-100 is not a problem, and is desirable.

net/dns/resolver/tsdns.go
```diff
@@ -83,6 +83,14 @@ type Config struct {
 	// LocalDomains is a list of DNS name suffixes that should not be
 	// routed to upstream resolvers.
 	LocalDomains []dnsname.FQDN
+	// ExitNodeBackupResolvers are where the local node when
+	// acting as an exit node and serving a DNS proxy should
+	// forward DNS requests to in the case where there are no
+	// routes found. For example, for Linux systemd-resolved
```
danderson, line 89: "no routes" ? I don't understand.

danderson (commented):
Fairly sure this is overcomplicating things. Just read /etc/resolv.conf and forward to whatever it says, whether or not it's quad-100, whether or not resolved is in charge. That should always produce the correct result on linux, and produces the least incorrect result on other platforms.

bradfitz:
> Fairly sure this is overcomplicating things. 

We don't want to loop back to 100.100.100.100 and get the exit node's MagicDNS names blended in if it's in a different tailnet (later)

net/dns/resolver/tsdns.go
bradfitz: If the Tailnet DNS config has neither global resolvers or Split DNS routes.

net/dns/direct.go
bradfitz: We don't want to loop back to 100.100.100.100 and get the exit node's MagicDNS names blended in if it's in a different tailnet.

We can discuss in eng meeting.

bradfitz:
New strategy: https://github.com/tailscale/tailscale/pull/3436

PR #3408, opened 2021-11-25
ipn/ipnlocal: restrict exit node DoH server based on ACL'ed packet filter

bradfitz:
Don't be a DoH DNS server to peers unless the Tailnet admin has permitted
that peer autogroup:internet access.

Updates #1713


bradfitz:
Now with a first commit that fixes filter.NewAllowAllForTest (with its own test), and then updated the now-second commit to add tests.

crawshaw (approved):
Certainly an improvement.

PR #3412, opened 2021-11-26
ipn/ipnlocal: populate self status from netmap in ipnlocal, not magicsock.

### LINKED ISSUE #1933: "tailscale status" doesn't always reflect a change in hostname
I am currently writing a puppet module to install and configure tailscale on linux.
I am having trouble accessing the current state of the client.
`/var/lib/tailscale/tailscaled.state` and `tailscale status --json` seem not to provide the same data considering hostname.

What is the expected way to evaluate the current state?
Is there some time to wait until the state is updated?
Should the `tailscaled` service be restarted after running `tailscale up`? 

In the following I will describe some scenarios. 
I have the feeling that there are some problems when using an empty string or the actual hostname as value for `--hostname`.
Overall after playing around a little bit I could not find a consistent wrong behaviour.

Btw I would really appreciate an official puppet module.

---
Example A: 

We have a host with hostname "one.two.tree.four".
When upgrading from tailscale version ~1.2.x to 1.8.3 the hostname reported to tailscale had been "one".
I require to have the full hostname reported to tailscale.

My puppet module accesses the current state by using the following command:
```
cat /var/lib/tailscale/tailscaled.state | jq -r "._daemon" | base64 --decode | jq
```

And executes the following command if `Hostname` (or actually `AdvertiseRoutes`) is not as expected.
```
tailscale up --authkey=$AUTH_KEY --hostname=one.two.tree.four --advertise-routes=
```

The hostname is correctly updated in `tailscale status --json` but not in `/var/lib/tailscale/tailscaled.state`.
Therefore on the next puppet run `tailscale up` is executed again.

---
Example B: 

When setting the hostname, the hostname is updated in the `tailscaled.state` but not in `tailscale status`.
But after a restart, it seems to work.

```
# Returns "one.two.three"
tailscale status --json | jq -r '.Self.HostName | values' 

tailscale up --hostname=VALUE

# Still returns "one.two.three"
tailscale status --json | jq -r '.Self.HostName | values'

# Returns "VALUE"
cat /var/lib/tailscale/tailscaled.state | jq -r "._daemon" | base64 --decode | jq -r '.Hostname  | values'

systemd restart tailscaled.state


# Returns "VALUE" as expected
tailscale status --json | jq -r '.Self.HostName | values'

# Returns "VALUE" as expected
cat /var/lib/tailscale/tailscaled.state | jq -r "._daemon" | base64 --decode | jq -r '.Hostname  | values'
```

---

Example C: 

Kind of the reverse of Example B.
When setting the hostname, the hostname is updated in `tailscale status` but not in `tailsclae.state`.
But I think after restarting the `tailscaled` service the hostname got resetted to the value in `tailscale.state`.

I could not reproduce this misbehaviour for this bug report but it happened a few times.

---

System:

```
# Ubuntu 18.04.5 LTS

tailscale version 
1.8.3
  tailscale commit: d0e86b08c99d23686e2c2416f1e3c93d5c00863a
  other commit: 03346d1fbc21a0ae7e02abc7a7b5e62569c4bc55
  go version: go1.16.4-ts1b677aa
```

danderson:
Fixes #1933

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -350,8 +350,18 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 		}
 	})
 	sb.MutateSelfStatus(func(ss *ipnstate.PeerStatus) {
-		if b.netMap != nil && b.netMap.SelfNode != nil {
-			ss.ID = b.netMap.SelfNode.StableID
+		if b.netMap != nil {
+			ss.HostName = b.netMap.Hostinfo.Hostname
+			ss.DNSName = b.netMap.Name
+			ss.UserID = b.netMap.User
+			if sn := b.netMap.SelfNode; sn != nil {
+				ss.ID = sn.StableID
+				if c := sn.Capabilities; len(c) > 0 {
+					ss.Capabilities = append([]string(nil), c...)
+				}
+			}
+		} else {
+			ss.HostName, _ = os.Hostname()
 		}
 		for _, pln := range b.peerAPIListeners {
 			ss.PeerAPIURL = append(ss.PeerAPIURL, pln.urlStr)
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -17,7 +17,6 @@ import (
 	"math"
 	"math/rand"
 	"net"
-	"os"
 	"reflect"
 	"runtime"
 	"sort"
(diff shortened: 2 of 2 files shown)


DentonGentry (approved):
I don't really understand the motivation for ipnlocal vs magicsock, but the code to implement that choice looks solid.

ipn/ipnlocal/local.go
```diff
@@ -350,8 +350,18 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 		}
 	})
 	sb.MutateSelfStatus(func(ss *ipnstate.PeerStatus) {
-		if b.netMap != nil && b.netMap.SelfNode != nil {
-			ss.ID = b.netMap.SelfNode.StableID
+		if b.netMap != nil {
+			ss.HostName = b.netMap.Hostinfo.Hostname
+			ss.DNSName = b.netMap.Name
+			ss.UserID = b.netMap.User
+			if b.netMap.SelfNode != nil {
```
bradfitz, line 357: The expression `b.netMap.SelfNode` appears 3 times, so maybe....

```
if sn := b.netMap.SelfNode; sn != nil {
    ...
```

bradfitz (approved):

Issue #3046, opened 2021-10-11
Older Android releases do not upload logs due to DST Root CA X3 expiration


DentonGentry:
### What is the issue?

The Android 1.16 release does not upload any logs. Prior releases built against SDK 29, but we had to update to SDK 30 by November 1 (https://github.com/tailscale/tailscale/issues/2293) and it looks like we're hitting a new restriction. From `adb logcat`:

```
10-10 18:22:24.406 32261 32289 I com.tailscale.ipn: goSetupLogs: success
10-10 18:22:24.425 32261 32261 W Thread-9: type=1400 audit(0.0:32521): avc: denied { read } for name="route" dev="proc" ino=4026532077 scontext=u:r:untrusted_app:s0:c205,c257,c512,c768 tcontext=u:object_r:proc_net:s0 tclass=file permissive=0 app=com.tailscale.ipn
10-10 18:22:24.448 32295 32295 W ip      : type=1400 audit(0.0:32522): avc: denied { bind } for scontext=u:r:untrusted_app:s0:c205,c257,c512,c768 tcontext=u:r:untrusted_app:s0:c205,c257,c512,c768 tclass=netlink_route_socket permissive=0 bug=b/155595000 app=com.tailscale.ipn
```

Other functions of the app like connectivity and Taildrop were tested, but I didn't think to check log uploads.

### OS

Android

### Tailscale version

1.16

DentonGentry:
False alarm: there are logs uploading now, using both 1.16.0 from the Play Store and from a tailscale-debug.apk I just built from main. The pixel-3a running 1.16.0 from the Play Store is f5181ad482bcce6a12ead74c88395cf8c74f59ef4128e98af95a931f16b4da39, and fresh logs are appearing.

I don't know what was going on 3 days ago. I checked both a Pixel-3a and Nexus-7 and didn't see any logs appearing. Maybe I got the log IDs wrong for both of them.

PR #3413, opened 2021-11-26
logpolicy: export NewLogtailTransport for Android

DentonGentry:
Android doesn't use logpolicy and currently has enough
unique stuff about its logging that makes it difficult to
do so. For example, its logsDir comes from Gio.

Export NewLogtailTransport to let Android use it.

Updates https://github.com/tailscale/tailscale/issues/3046

### DIFF
--- a/logpolicy/logpolicy.go
+++ b/logpolicy/logpolicy.go
@@ -501,7 +501,7 @@ func New(collection string) *Policy {
 			}
 			return w
 		},
-		HTTPC: &http.Client{Transport: newLogtailTransport(logtail.DefaultHost)},
+		HTTPC: &http.Client{Transport: NewLogtailTransport(logtail.DefaultHost)},
 	}
 	if collection == logtail.CollectionNode {
 		c.MetricsDelta = clientmetric.EncodeLogTailMetricsDelta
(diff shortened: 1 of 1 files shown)


logpolicy/logpolicy.go
```diff
@@ -571,9 +571,12 @@ func (p *Policy) Shutdown(ctx context.Context) error {
 	return nil
 }
 
-// newLogtailTransport returns the HTTP Transport we use for uploading
-// logs to the given host name.
-func newLogtailTransport(host string) *http.Transport {
+// NewLogtailTransport returns an HTTP Transport particularly suited to uploading
+// logs to the given host name. This includes:
+//   - If DNS lookup fails, consult the bootstrap DNS list of Tailscale hostnames.
+//   - If TLS connection fails, try again using a built-in LetsEncrypt's ISRG Root X1
```
bradfitz, line 577: missing word in here? or: "try again using LetsEncrypt's built-in ..."

bradfitz (approved):

logpolicy/logpolicy.go
DentonGentry: Done.

PR #3421, opened 2021-11-26
net/dns/resolver: add clientmetrics for DNS

bradfitz:
Fixes tailscale/corp#1811

### DIFF
--- a/net/dns/resolver/forwarder.go
+++ b/net/dns/resolver/forwarder.go
@@ -385,6 +385,7 @@ func (f *forwarder) sendDoH(ctx context.Context, urlBase string, c *http.Client,
 	}
 	defer f.releaseDoHSem()
 
+	metricDNSFwdDoH.Add(1)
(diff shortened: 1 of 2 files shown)


danderson (approved):
Hmm, that's a lot of metrics. Hopefully the compression on the reporting will mean that most of these ~never change and so it's all fine. LGTM though, I'm just musing about what we're going to discover about the limits of this API :)

DentonGentry (approved):

bradfitz:
> Hmm, that's a lot of metrics. Hopefully the compression on the reporting will mean that most of these ~never change and so it's all fine. LGTM though, I'm just musing about what we're going to discover about the limits of this API :)

Yeah, the error cases will never make it to the wire/disk. But they'll consume a bit of CPU in scanning for changes (but at least they're contiguously laid out in memory now for fast scanning). There are several things we can do if/when it comes time for it, though, like lazily publish varz on first use, and make metric increment slower by keeping a dirty list of metrics or last change epoch value.

PR #3422, opened 2021-11-27
Support running SOCKS5 and HTTP proxies on the same port.

### LINKED ISSUE #3248: cmd/tailscaled: support running SOCKS5 & HTTP outbound proxy on same port
We should support running the SOCKS5 and HTTP outbound proxy on the same port number.

We can pick the right protocol based on what the client send.

danderson:
Fixes #3248


net/proxymux/mux.go
```diff
@@ -0,0 +1,134 @@
...
+import (
+	"io"
+	"net"
+	"sync"
+)
+
+// Mux accepts connections on l and passes connections through to
+// either socksListener or httpListener, depending the first byte sent
+// by the client.
+func Mux(l net.Listener) (socksListener, httpListener net.Listener) {
```
bradfitz, line 21: s/l/ln/ which Go style has moved towards (avoiding the variable `l` which looks like `1` in many people's fonts)

net/proxymux/mux.go
```diff
@@ -0,0 +1,134 @@
...
+		c:      make(chan net.Conn),
+		closed: make(chan struct{}),
+	}
+	hl := &listener{
+		addr:   l.Addr(),
+		c:      make(chan net.Conn),
+		closed: make(chan struct{}),
+	}
+
+	go func() {
```
bradfitz, line 33: make this a method on something so it shows up nicely in goroutine dumps? otherwise it's hard to keep track of tons of closure goroutines.

net/proxymux/mux.go
bradfitz: rename "Mux" to "SplitSOCKSAndHTTP" or something to keep this package generic for the future for other muxing or configurability? then the name also describes the order of the result params so caller can't transpose so easily.

(this is almost the tcpproxy thing we worked on before if you squint)

net/proxymux/mux.go
```diff
@@ -0,0 +1,134 @@
...
+			go routeConn(conn, sl, hl)
+		}
+	}()
+
+	return sl, hl
+}
+
+func routeConn(c net.Conn, socksListener, httpListener *listener) {
+	var b [1]byte
+	if _, err := io.ReadFull(c, b[:]); err != nil {
```
bradfitz, line 50: this has no read timeout so this can block forever. (and if something can, we know it will)

Set a 15 second read deadline here and clear after (setting it to time.Time{})?

net/proxymux/mux.go
```diff
@@ -0,0 +1,134 @@
...
+		c.Close()
+		return
+	}
+	conn := &connWithOneByte{
+		Conn: c,
+		b:    b[0],
+	}
+
+	// First byte of a SOCKS5 session is a version byte set to 5.
+	var l *listener
```
bradfitz, line 60: ln

net/proxymux/mux.go
```diff
@@ -0,0 +1,134 @@
...
+func (l *listener) Addr() net.Addr {
+	return l.addr
+}
+
+// connWithOneByte is a net.Conn that returns b for the first read
+// request, then forwards everything else to Conn.
+type connWithOneByte struct {
+	net.Conn
+
+	mu    sync.Mutex
```
bradfitz, line 117: not needed; Read on a TCPConn isn't defined to be safe for concurrent reads

so I'd rather get a data race if anybody ever tried to do that.

net/proxymux/mux_test.go
```diff
@@ -0,0 +1,172 @@
...
+	s := mkScaffold(t)
+	defer s.Close()
+
+	s.httpListener.Close()
+	s.socksListener.Close()
+	s.checkURL(s.httpClient, true)
+	s.checkURL(s.socksClient, true)
+}
+
+type scaffold struct {
```
bradfitz, line 54: I can live with this name, but my life is better the less I see this term :)

I usually go with "env" or "world" or "fooTest" or something. But I guess people know this too, so sure.

cmd/tailscaled/tailscaled.go
```diff
@@ -515,18 +515,38 @@ func newNetstack(logf logger.Logf, e wgengine.Engine) (*netstack.Impl, error) {
...
-func mustStartTCPListener(name, addr string) net.Listener {
-	if addr == "" {
-		return nil
+func mustStartProxyListeners(socksAddr, httpAddr string) (socksListener, httpListener net.Listener) {
+	if socksAddr == httpAddr && socksAddr != "" && !strings.HasSuffix(socksAddr, ":0") {
+		ln, err := net.Listen("tcp", socksAddr)
+		if err != nil {
+			log.Fatalf("proxy listener: %v", err)
+		}
+		return proxymux.Mux(ln)
```
bradfitz, line 524: yeah, I think this will read nicer here with a name like suggested above

cmd/tailscaled/tailscaled.go
```diff
@@ -515,18 +515,38 @@ func newNetstack(logf logger.Logf, e wgengine.Engine) (*netstack.Impl, error) {
 	return netstack.Create(logf, tunDev, e, magicConn)
 }
 
-func mustStartTCPListener(name, addr string) net.Listener {
-	if addr == "" {
-		return nil
+func mustStartProxyListeners(socksAddr, httpAddr string) (socksListener, httpListener net.Listener) {
```
bradfitz, line 518: specify that they can be nil returns

bradfitz (approved):
Yay!

net/proxymux/mux.go
danderson: Renamed, adjusted varnames

PR #3440, opened 2021-11-30
tailcfg, ipn/ipnlocal: advertise a Service when exit node DNS proxy available

bradfitz:
Updates #1713

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -1753,13 +1753,23 @@ func (b *LocalBackend) getPeerAPIPortForTSMPPing(ip netaddr.IP) (port uint16, ok
 
 func (b *LocalBackend) peerAPIServicesLocked() (ret []tailcfg.Service) {
 	for _, pln := range b.peerAPIListeners {
-		proto := tailcfg.ServiceProto("peerapi4")
+		proto := tailcfg.PeerAPI4
 		if pln.ip.Is6() {
-			proto = "peerapi6"
+			proto = tailcfg.PeerAPI6
 		}
 		ret = append(ret, tailcfg.Service{
 			Proto: proto,
 			Port:  uint16(pln.port),
+		})
+	}
+	switch runtime.GOOS {
+	case "linux", "freebsd", "openbsd", "illumos", "darwin":
+		// These are the platforms currently supported by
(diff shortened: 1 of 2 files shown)


tailcfg/tailcfg.go
```diff
@@ -379,18 +379,48 @@ func (h *Hostinfo) CheckRequestTags() error {
...
+	//        port number that the peerapi is running on the
+	//        node's Tailscale IPv4 address.
+	//     * "peerapi6": peerapi is available on IPv6; Port is the
+	//        port number that the peerapi is running on the
+	//        node's Tailscale IPv6 address.
+	//     * "peerapi-dns": the local peerapi service supports
+	//        being a DNS proxy (when the node is an exit
+	//        node). For this service, the Port number is really
+	//        the version number of the service.
+	Proto ServiceProto // TCP or UDP
```
crawshaw, line 412: now you have a big fancy comment you don't need the trailing comment

tailcfg/tailcfg.go
```diff
@@ -379,18 +379,48 @@ func (h *Hostinfo) CheckRequestTags() error {
...
+
+	// Port is the port number.
```
crawshaw, line 414: "...the service is listening on." ? then drop the trailing comment?

ipn/ipnlocal/local.go
```diff
@@ -1753,15 +1753,25 @@ func (b *LocalBackend) getPeerAPIPortForTSMPPing(ip netaddr.IP) (port uint16, ok
...
-			proto = "peerapi6"
+			proto = tailcfg.PeerAPI6
 		}
 		ret = append(ret, tailcfg.Service{
 			Proto: proto,
 			Port:  uint16(pln.port),
 		})
 	}
+	switch runtime.GOOS {
+	case "linux", "freebsd", "openbsd", "illumos", "darwin":
```
crawshaw, line 1766: I wish this weren't so far away from the implementations of HandleExitNodeDNSQuery.

crawshaw (approved):

ipn/ipnlocal/local.go
bradfitz: Same, but ran out of ideas. Welcome any good ones. Ideally this is gone in a day or so?

PR #3450, opened 2021-11-30
net/dns: windows DNS experiments

bradfitz:


bradfitz:
Looks like the default timeout is 10 seconds (https://docs.microsoft.com/en-us/troubleshoot/windows-server/networking/dns-client-resolution-timeouts), which means if things go poorly, we could amass a ton of threads in a short amount of time if we're an exit node for a bunch of clients and we lose our internet for a bit or something.

So we probably want the async interface.

cc @dblohm7

dblohm7:
Workin' on it

bradfitz:
> Workin' on it

Learn anything?

PR #3454, opened 2021-11-30
types/logger: add Context and related helpers

josharian:
We often need both a log function and a context.
We can do this by adding the log function as a context value.
This commit adds helper glue to make that easy.
It is designed to allow incremental adoption.

Updates tailscale/corp#3138

### DIFF
--- a/types/logger/logger.go
+++ b/types/logger/logger.go
@@ -18,12 +18,35 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"context"
 )
 
 // Logf is the basic Tailscale logger type: a printf-like func.
 // Like log.Printf, the format need not end in a newline.
 // Logf functions must be safe for concurrent use.
 type Logf func(format string, args ...interface{})
+
+// A Context is a context.Context that should contain a custom log function, obtainable from FromContext.
+// If no log function is present, FromContext will return log.Printf.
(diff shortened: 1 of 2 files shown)


josharian:
cc @danderson @crawshaw who also have Opinions about logging

types/logger/logger.go
```diff
@@ -18,13 +18,36 @@ import (
...
 
+// A Context is a context.Context that should contain a custom log function, obtainable from FromContext.
+// If no log function is present, FromContext will return log.Printf.
+// To construct a Context, use Add
+type Context context.Context
+
+type logfKey struct{}
+
+// FromContext extracts a log function from ctx.
+func FromContext(ctx Context) Logf {
```
crawshaw, line 38: so we log a line by writing `logger.FromContext(ctx).Logf("foo")`? seems a teensy bit wordy. no obvious alternatives in the typical go vocab, but maybe something like `logger.Printf(ctx, "foo")`?

josharian: For sample usage see https://github.com/tailscale/corp/pull/3151

crawshaw: I have had enough different opinions over the years that I'll leave this to someone else.

bradfitz (approved):

PR #3455, opened 2021-11-30
ipn/ipnlocal, net/dns/resolver: use exit node's DoH proxy when available

bradfitz:
Updates #1713

### DIFF
--- a/ipn/ipnlocal/peerapi.go
+++ b/ipn/ipnlocal/peerapi.go
@@ -849,7 +849,7 @@ func (h *peerAPIHandler) handleDNSQuery(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 	w.Header().Set("Content-Type", "application/dns-message")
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/peerapi.go
```diff
@@ -849,7 +849,7 @@ func (h *peerAPIHandler) handleDNSQuery(w http.ResponseWriter, r *http.Request)
 		return
 	}
 	w.Header().Set("Content-Type", "application/dns-message")
-	w.Header().Set("Content-Length", strconv.Itoa(len(q)))
+	w.Header().Set("Content-Length", strconv.Itoa(len(res)))
```
crawshaw, line 852: 💀

crawshaw (approved):

ipn/ipnlocal/peerapi.go
bradfitz: Well, at least the client side caught it with an UnexpectedEOF! :)

PR #3457, opened 2021-11-30
net/tsdial: start of new package to unify all outbound dialing complexity

bradfitz:
For now this just deletes the net/socks5/tssocks implementation (and
the DNSMap stuff from wgengine/netstack) and moves it into net/tsdial.

Then initialize a Dialer early in tailscaled, currently only use for the
outbound and SOCKS5 proxies. It will be plumbed more later. Notably, it
needs to get down into the DNS forwarder for exit node DNS forwading
in netstack mode. But it will also absorb all the peerapi setsockopt
and netns Dial and tlsdial complexity too.

Updates #1713


net/tsdial/dnsmap.go
```diff
@@ -0,0 +1,109 @@
...
+	"strings"
+
+	"inet.af/netaddr"
+	"tailscale.com/types/netmap"
+	"tailscale.com/util/dnsname"
+)
+
+// DNSMap maps MagicDNS names (both base + FQDN) to their first IP.
+// It should not be mutated once created.
+type DNSMap map[string]netaddr.IP
```
danderson, line 21: use dnsname.FQDN as the key? Forces running dnsname.ToFQDN to populate, which eliminates issues around trailing dot vs. no trailing dot.

net/tsdial/dnsmap.go
```diff
@@ -0,0 +1,109 @@
...
+	ipp, pippErr := netaddr.ParseIPPort(addr)
+	if pippErr == nil {
+		return ipp, nil
+	}
+	host, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		// addr is malformed.
+		return netaddr.IPPort{}, err
+	}
+	if net.ParseIP(host) != nil {
```
danderson, line 80: netaddr.ParseIP?

net/tsdial/dnsmap.go
```diff
@@ -0,0 +1,109 @@
...
+		return netaddr.IPPort{}, pippErr
+	}
+	port16, err := strconv.ParseUint(port, 10, 16)
+	if err != nil {
+		return netaddr.IPPort{}, fmt.Errorf("invalid port in address %q", addr)
+	}
+
+	// Host is not an IP, so assume it's a DNS name.
+
+	// Try MagicDNS first, else otherwise a real DNS lookup.
```
danderson, line 92: else otherwise

net/tsdial/dnsmap.go
```diff
@@ -0,0 +1,109 @@
...
+	ip := m[host]
+	if !ip.IsZero() {
+		return netaddr.IPPortFrom(ip, uint16(port16)), nil
+	}
+
+	// No MagicDNS name so try real DNS.
+	var r net.Resolver
```
danderson, line 99: TODO: hook to net/dnscache?

net/tsdial/dnsmap.go
```diff
@@ -0,0 +1,109 @@
...
+	"strconv"
+	"strings"
+
+	"inet.af/netaddr"
+	"tailscale.com/types/netmap"
+	"tailscale.com/util/dnsname"
+)
+
+// DNSMap maps MagicDNS names (both base + FQDN) to their first IP.
+// It should not be mutated once created.
```
danderson, line 20: Make stronger? "must not" ?

danderson (approved):

net/tsdial/dnsmap.go
bradfitz: well, per its docs, it holds bare names too, without FQDNs, so you can `curl http://foo` over SOCKS/HTTP proxy.

So dsnname.FQDN would be confusing in a different way. I can list some example keys formats, though.

net/tsdial/dnsmap.go
bradfitz: done

net/tsdial/dnsmap.go
bradfitz: done

net/tsdial/dnsmap.go
bradfitz: done

net/tsdial/dnsmap.go
bradfitz: (also, this is just code movement so I'm reluctant to do too much rewriting during a git mv... but we can later)

PR #3462, opened 2021-12-01
Replace AWS SDK v1 dependency with v2

### LINKED ISSUE #3461: FR: don't depend on two different AWS SDKs

go.mod depends on two AWS SDKs: monolithic old one (v1), and more recent modular v2:

https://github.com/tailscale/tailscale/blob/56c72d9cde8b2851f0753be4c81cf4d5af23c8a8/go.mod#L10-L11

### How should we solve this?

v1 SDK dependency is only used inside a single test, and can be upgraded to v2 SDK.

Keeping dependency on a rather fat SDK.

### Anything else?

_No response_

artyom:
This change drops AWS SDKv1 dependency, leaving only SDK v2 in use.

Closes #3461

Signed-off-by: Artyom Pervukhin <github@artyom.dev>


bradfitz:
LGTM, thank you!

PR #3464, opened 2021-12-01
wgengine/monitor: fix docs on Mon.InterfaceState

bradfitz:
The behavior was changed in March (in 7f174e84e627dc04bc91352e9897f22cf4724191)
but that change forgot to update these docs.

### DIFF
--- a/wgengine/monitor/monitor.go
+++ b/wgengine/monitor/monitor.go
@@ -104,8 +104,10 @@ func New(logf logger.Logf) (*Mon, error) {
 	return m, nil
 }
 
-// InterfaceState returns the state of the machine's network interfaces,
-// without any Tailscale ones.
+// InterfaceState returns the latest snapshot of the machine's network
+// interfaces.
+//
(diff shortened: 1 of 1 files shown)


wgengine/monitor/monitor.go
```diff
@@ -104,8 +104,10 @@ func New(logf logger.Logf) (*Mon, error) {
 	return m, nil
 }
 
-// InterfaceState returns the state of the machine's network interfaces,
-// without any Tailscale ones.
+// InterfaceState returns the latest snapshot of the machine's network
```
josharian, line 107: Nit: put the whole first sentence on the first line.

josharian (approved):

wgengine/monitor/monitor.go
bradfitz: That goes over 80 characters and my editor wraps comments before that. The rest of this file (e.g. the several funcs just below this) are consistent at wrapping before 80 too.

So I'll keep as is. There's at least a blank line before the second paragraph?

PR #3465, opened 2021-12-01
net/portmapper: deflake TestPCPIntegration

josharian:
Logging in goroutines after the test completed
caused data races and panics. Prevent that.


josharian:
Linux failure is a QEMU apt-get failure.

crawshaw (approved):

PR #3466, opened 2021-12-01
net/tsdial: move macOS/iOS peerapi sockopt logic from LocalBackend

bradfitz:


net/tsdial/tsdial.go
```diff
@@ -67,6 +67,23 @@ func (d *Dialer) SetLinkMonitor(mon *monitor.Mon) {
 	d.linkMon = mon
 }
 
+func (d *Dialer) interfaceIndexLocked(ifName string) (index int, ok bool) {
+	if d.linkMon == nil {
+		return 0, false
+	}
+	st := d.linkMon.InterfaceState()
+	tunIf, ok := st.Interface[ifName]
```
danderson, line 75: tunIf is a bit surprising in a generic func. Maybe iface?

danderson (approved):

net/tsdial/tsdial.go
```diff
@@ -67,6 +67,23 @@ func (d *Dialer) SetLinkMonitor(mon *monitor.Mon) {
...
+	if !ok {
+		return 0, false
+	}
+	return tunIf.Index, true
+
```
bradfitz, line 80: whoops, just noticed this. deleted. will repush after you're done reviewing.

net/tsdial/tsdial.go
bradfitz: Done.

PR #3472, opened 2021-12-02
control/controlhttp: package to get a controlbase.Conn over HTTP(S)

danderson:
Before I go deeper down this rabbithole, could you confirm that there's no easier way to do "get a net.Conn via 101 Switching Protocols" with net/http than this? I'm having to reimplement and paraphrase a bunch of net/http's connection establishment logic to deal with the various proxying cases. Not very much code, but quite subtle :(

On the plus side, with a little more work, this one piece of code could be used for both DERP and control protocol dialing. But I feel like I'm missing some obvious way to hijack a conn client-side and as a result be able to lean much more on net/http...

### DIFF
--- a/control/controlhttp/client.go
+++ b/control/controlhttp/client.go
@@ -1,0 +1,242 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package controlhttp implements the Tailscale 2021 control protocol
+// base transport over HTTP.
+//
+// This tunnels the protocol in control/controlbase over HTTP with a
+// variety of compatibility fallbacks for handling picky or deep
+// inspecting proxies.
+//
+// In the happy path, a client makes a single cleartext HTTP request
+// to the server, the server responds with 101 Switching Protocols,
+// and the control base protocol takes place over plain TCP.
+//
+// In the compatibility path, the client does the above over HTTPS,
+// resulting in double encryption (once for the control transport, and
+// once for the outer TLS layer).
+package controlhttp
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"net"
+	"net/http"
+	"net/http/httptrace"
+	"net/url"
+
+	"tailscale.com/control/controlbase"
+	"tailscale.com/net/dnscache"
+	"tailscale.com/net/dnsfallback"
+	"tailscale.com/net/netns"
+	"tailscale.com/net/tlsdial"
+	"tailscale.com/net/tshttpproxy"
+	"tailscale.com/types/key"
+)
+
+// upgradeHeader is the value of the Upgrade HTTP header used to
+// indicate the Tailscale control protocol.
+const (
+	upgradeHeaderValue  = "tailscale-control-protocol"
+	handshakeHeaderName = "X-Tailscale-Handshake"
+)
+
+// Dial connects to the HTTP server at addr, requests to switch to the
+// Tailscale control protocol, and returns an established control
+// protocol connection.
+//
+// If Dial fails to connect using addr, it also tries to tunnel over
+// TLS to <addr's host>:443 as a compatibility fallback.
+func Dial(ctx context.Context, addr string, machineKey key.MachinePrivate, controlKey key.MachinePublic) (*controlbase.Conn, error) {
+	host, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		return nil, err
+	}
+	a := &dialParams{
+		ctx:        ctx,
+		host:       host,
+		httpPort:   port,
+		httpsPort:  "443",
+		machineKey: machineKey,
+		controlKey: controlKey,
+		proxyFunc:  tshttpproxy.ProxyFromEnvironment,
+	}
+	return a.dial()
+}
+
+type dialParams struct {
+	ctx        context.Context
+	host       string
+	httpPort   string
+	httpsPort  string
+	machineKey key.MachinePrivate
+	controlKey key.MachinePublic
+	proxyFunc  func(*http.Request) (*url.URL, error) // or nil
+
+	// For tests only
+	insecureTLS bool
+}
+
+func (a *dialParams) dial() (*controlbase.Conn, error) {
+	init, cont, err := controlbase.ClientDeferred(a.machineKey, a.controlKey)
+	if err != nil {
+		return nil, err
+	}
+
+	u := &url.URL{
+		Scheme: "http",
+		Host:   net.JoinHostPort(a.host, a.httpPort),
+		Path:   "/switch",
+	}
+	conn, httpErr := a.tryURL(u, init)
+	if httpErr == nil {
+		ret, err := cont(a.ctx, conn)
+		if err != nil {
+			conn.Close()
+			return nil, err
+		}
+		return ret, nil
+	}
+
+	// Connecting over plain HTTP failed, assume it's an HTTP proxy
+	// being difficult and see if we can get through over HTTPS.
+	u.Scheme = "https"
+	u.Host = net.JoinHostPort(a.host, a.httpsPort)
+	init, cont, err = controlbase.ClientDeferred(a.machineKey, a.controlKey)
+	if err != nil {
+		return nil, err
+	}
+	conn, tlsErr := a.tryURL(u, init)
+	if tlsErr == nil {
+		ret, err := cont(a.ctx, conn)
+		if err != nil {
+			conn.Close()
+			return nil, err
+		}
+		return ret, nil
+	}
+
+	return nil, fmt.Errorf("all connection attempts failed (HTTP: %v, HTTPS: %v)", httpErr, tlsErr)
+}
+
+func (a *dialParams) tryURL(u *url.URL, init []byte) (net.Conn, error) {
+	dns := &dnscache.Resolver{
+		Forward:          dnscache.Get().Forward,
+		LookupIPFallback: dnsfallback.Lookup,
+		UseLastGood:      true,
+	}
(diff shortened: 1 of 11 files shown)


bradfitz:
Actually, there is one slightly sneaky way you can hijack a connection as long as it's the first HTTP response on a connection:

https://pkg.go.dev/net/http/httptrace#GotConnInfo and its Conn field.

If you use that trace hook, you'll have the underlying Conn.

Then just do a single HTTP request and when you see the 101 Switching Protocols, start using the Conn I think.

Looks like Transport will return 101 without an error; there's code/tests for it:

```go
                is1xx := 100 <= resCode && resCode <= 199
                // treat 101 as a terminal status, see issue 26161                                                                                                                       
                is1xxNonTerminal := is1xx && resCode != StatusSwitchingProtocols
```

danderson:
Wow, that's gross, I love it. Do you think we can lean on Hyrum's Law that way? Can you think of anything in the pipeline for net/http that would break it?

Prototyping, thanks for the wisdom.

danderson:
Redone with httptrace, and with one test. I'm minded to add more tests to verify that things work with all the various proxy/fallback options, although it's a little irritating to wrangle TLS during testing. Thoughts?

danderson:
Now featuring tests that wire up all different kinds of proxies and proxying policies, and check we can dial through all of them and upgrade to Noise.

danderson:
I can reproduce the race failure, but I have no clue what's up with it. If you notice something while reviewing, please help :)

control/controlhttp/client.go
```diff
@@ -0,0 +1,180 @@
...
+	// Technically, the underlying net.Conn is wrapped in a
+	// bufio.Reader within Transport, but the way the control protocol
+	// works, the client is the first to transmit after the protocol
+	// switch. This means the buffer within the Transport will never
+	// contain more than just the HTTP response, and as a result it's
+	// safe to steal the raw net.Conn and use that.
+	var switchedConn net.Conn
+	trace := httptrace.ClientTrace{
+		GotConn: func(info httptrace.GotConnInfo) {
+			switchedConn = info.Conn
```
bradfitz, line 141: I can't convince myself that this isn't technically a data race, which is why I'd used a channel before

control/controlhttp/client.go
```diff
@@ -0,0 +1,180 @@
...
+// protocol connection.
+//
+// If Dial fails to connect using addr, it also tries to tunnel over
+// TLS to <addr's host>:443 as a compatibility fallback.
+func Dial(ctx context.Context, addr string, machineKey key.MachinePrivate, controlKey key.MachinePublic) (*controlbase.Conn, error) {
+	host, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		return nil, err
+	}
+	a := attempt{
```
bradfitz, line 55: if all the methods are on the pointer, use &attempt here?

control/controlhttp/client.go
```diff
@@ -0,0 +1,180 @@
...
+	// From here we've confirmed the response is a protocol switch, so
+	// the switchedConn will no longer be touched by Transport. It's
+	// safe for us to handle, and also our responsibility to close on
+	// error.
+
+	if switchedConn == nil {
+		return nil, fmt.Errorf("httptrace didn't provide a connection")
+	}
+
+	if next := resp.Header.Get("Upgrade"); next != "tailscale-control-protocol" {
```
bradfitz, line 174: you'd made a `const upgradeHeader` for this.

control/controlhttp/http_test.go
```diff
@@ -0,0 +1,389 @@
...
+func TestControlHTTP(t *testing.T) {
+	tests := []struct {
+		name  string
+		proxy proxy
+		//rejectHTTP bool
+	}{
+		{
+			name:  "no_proxy",
+			proxy: nil,
+		}, // direct connection
```
bradfitz, line 34: move this comment above the block like the cases below?

control/controlhttp/server.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, errors.New("no next protocol in HTTP request")
+	}
+	if next != upgradeHeader {
+		http.Error(w, "unknown next protocol", http.StatusBadRequest)
+		return nil, fmt.Errorf("client requested unhandled next protocol %q", next)
+	}
+
+	hijacker, ok := w.(http.Hijacker)
+	if !ok {
+		http.Error(w, "make request over HTTP1", http.StatusBadRequest)
```
bradfitz, line 37: I think the hot new retcon spelling is HTTP/1

control/controlhttp/client.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+type attempt struct {
+	ctx        context.Context
+	host       string
+	httpPort   string
+	httpsPort  string
+	machineKey key.MachinePrivate
+	controlKey key.MachinePublic
+	proxyFunc  func(*http.Request) (*url.URL, error)
```
bradfitz, line 74: // or nil

bradfitz (approved):

sailorfrag (approved):
I have no comments to add beyond what Brad already wrote

control/controlhttp/client.go
```diff
@@ -0,0 +1,180 @@
...
+	if switchedConn == nil {
+		return nil, fmt.Errorf("httptrace didn't provide a connection")
+	}
+
+	if next := resp.Header.Get("Upgrade"); next != "tailscale-control-protocol" {
+		switchedConn.Close()
+		return nil, fmt.Errorf("server switched to unexpected protocol %q", next)
+	}
+
+	return switchedConn, nil
```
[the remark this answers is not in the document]
danderson: Hmm, this is a bit tricky without breaking the currently very nice property of the core crypto package, which is that it doesn't leak any of the handshake's internal cryptographic state. It takes one net.Conn, does the entire handshake synchronously, and returns another net.Conn. So, by definition, I don't know the payload to send until I can hand it a net.Conn.

I don't know how to resolve this without breaking the nice property that the dangerous cryptography is all hidden and hard to misuse. I'd love to pick your brain for ideas.

control/controlhttp/http_test.go
```diff
@@ -0,0 +1,389 @@
...
+
+	"tailscale.com/net/socks5"
+	"tailscale.com/types/key"
+)
+
+func TestControlHTTP(t *testing.T) {
+	tests := []struct {
+		name  string
+		proxy proxy
+		//rejectHTTP bool
```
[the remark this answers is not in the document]
danderson: Deleted, it got moved into the proxy config layer.

control/controlhttp/http_test.go
danderson: Oops, dunno how that happened, must have been in the wrong spot and got gofmt'd more wrong.

control/controlhttp/server.go
danderson: Done.

control/controlhttp/client.go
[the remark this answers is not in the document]
danderson: Pushed one more commit that dirties the base transport API (and introduces new misuse risks, which I think I've guarded against). But in return, the noise and HTTP handshakes are now intertwined, and we save a round-trip. PTAL @bradfitz ? Are you okay with the API tradeoffs this needs, or should I stick with the extra RTT?

danderson:
Damn, the tests are flaky, stuff times out sometimes. Debugging, but PTAL nonetheless.

control/controlhttp/server.go
```diff
@@ -0,0 +1,90 @@
...
+	conn, brw, err := hijacker.Hijack()
+	if err != nil {
+		return nil, fmt.Errorf("hijacking client connection: %w", err)
+	}
+	if err := brw.Flush(); err != nil {
+		conn.Close()
+		return nil, fmt.Errorf("flushing hijacked HTTP buffer: %w", err)
+	}
+	if brw.Reader.Buffered() > 0 {
+		conn = &drainBufConn{conn, brw.Reader}
```
bradfitz, line 61: I see... you're getting it from the HTTP body. That's a little fragile. I'd prefer to do what HTTP/2 does (and websockets kinda) and send the buffered beginning part base64 in a new HTTP header.

bradfitz: (In any case, I'm very glad to be getting rid of the RTT. If we're going to have a single control region, we need to do everything we can to eliminate RTTs.)

danderson: Can you say more about what's fragile about using the body? I find this surprising, so would like to document it for future readers.

bradfitz: I was thinking of proxies that could read+forward only the headers and add "Expect: 100-continue". But I guess this would survive that.

Also if you've already sent a request body and the server replies with a non-2xx, the connection's basically dead at that point (out of sync) and can't be reused by the client. The server's probably going to close it immediately after sending the 2xx, since it doesn't want to deal with the rando request body for a request it rejected.

But if we're not ever going to send a non-2xx to these, maybe that's fine too.

I guess it's fine. It just weirds me out a bit.

danderson: Yeah, for auth cases we get all the creds we need out of band. And for this particular protocol, we're only using HTTP as a stepping stone where we fully expect to either successfully stop being HTTP, or torch the entire connection on failure. Still happy to switch to headers if that makes you less nervous, but it sounds like we're okay with using the body?...

danderson:
Still hitting test flakes, pretty repeatably on CI and very infrequently on my workstation (needs 1-10k attempts, and even then it's occasional). Not submitting until I've root-caused.

danderson:
I don't get it. I'm now running TestControlHTTP/no_proxy in a loop, over 100k runs on my machine, with and without race, and no failures. Yet CI is able to fail every single time, with an error I can't reproduce and that makes no sense.

Help @bradfitz? See anything obvious in the code that would make the client read an EOF instead of headers?

bradfitz (approved):

Issue #3475, opened 2021-12-02
tsdial: use ExitDNS when SOCKS/HTTP dialing


bradfitz:
When an exit node is in use, and that exit node supports "ExitDNS" (the new peerapi DoH server in 1.19.x+), then the tsdial.UserDial func should only do DNS resolution via ExitDNS, and not via the host as it does now (in dnsmap.go).

bradfitz:
Note: #1970 was closed the other day because we now send traffic through exit nodes, but that only made userspace-networking (SOCKS) mode reach parity with 1.18-and-below exit node behavior.

In 1.20 we're trying to do also do all DNS through exit nodes (the "ExitDNS" DoH proxy), so this bug is about making outbound proxy dialing also do that when in TUN-free userspace-networking mode. (or even when `--accept-dns=false`)

PR #3478, opened 2021-12-02
net/tsdial: give netstack a Dialer, start refactoring name resolution

bradfitz:
This starts to refactor tsdial.Dialer's name resolution to have
different stages: in-memory MagicDNS vs system resolution. A future
change will plug in ExitDNS resolution.

This also plumbs a Dialer into netstack and unexports the dnsMap
internals.

And it removes some of the async AddNetworkMapCallback usage and
replaces it with synchronous updates of the Dialer's netmap
from LocalBackend, since the LocalBackend has the Dialer too.

Updates #3475

### DIFF
--- a/net/tsdial/tsdial.go
+++ b/net/tsdial/tsdial.go
@@ -18,6 +18,7 @@ import (
 
 	"inet.af/netaddr"
 	"tailscale.com/net/netknob"
+	"tailscale.com/types/netmap"
 	"tailscale.com/wgengine/monitor"
 )
 
@@ -43,7 +44,7 @@ type Dialer struct {
 	peerDialer     *net.Dialer
 
 	mu      sync.Mutex
-	dns     DNSMap
+	dns     dnsMap
 	tunName string // tun device name
 	linkMon *monitor.Mon
 }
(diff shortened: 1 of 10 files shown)


net/tsdial/tsdial.go
```diff
@@ -102,26 +103,55 @@ func (d *Dialer) PeerDialControlFunc() func(network, address string, c syscall.R
 	return peerDialControlFunc(d)
 }
 
-// SetDNSMap sets the current map of DNS names learned from the netmap.
-func (d *Dialer) SetDNSMap(m DNSMap) {
-	// TODO(bradfitz): update this to be aware of DNSConfig
-	// ExtraNames and CertDomains.
+// SetNetMap sets the current network map and notably, the DNS names
+// in its DNS configuration.
+func (d *Dialer) SetNetMap(nm *netmap.NetworkMap) {
```
danderson, line 108: netmaps are not a good interchange format for configuration, because they do not blend in local preferences. I'm sad that we continue to treat it as ground truth :(

danderson (approved):

net/tsdial/tsdial.go
bradfitz: It's getting less and less with each commit. What it really needs is the DNSConfig extra names & info on whether this node is v4, v6, or both, all of which the NetMap has. The next commit will make the Dialer start to learn about prefs (so it knows whether to use ExitDNS)

PR #3485, opened 2021-12-02
cmd/tailscale: don't allow setting exit node to known invalid value

### LINKED ISSUE #3081: Linux exit node selection doesn't warn when selecting bogus IP
If you do `tailscale up --exit-node=a.b.c.d` where `a.b.c.d` isn't an approved exit node, there's no error.

Instead, things just kinda break silently.

Somewhat related: #3062 (let `--exit-node=` take a name instead of just an IP)

bradfitz:
Fixes #3081

### DIFF
--- a/cmd/tailscale/cli/up.go
+++ b/cmd/tailscale/cli/up.go
@@ -193,12 +193,38 @@ func calcAdvertiseRoutes(advertiseRoutes string, advertiseDefaultRoute bool) ([]
 	return routes, nil
 }
 
+// peerWithTailscaleIP returns the peer in st with the provided
+// Tailscale IP.
+func peerWithTailscaleIP(st *ipnstate.Status, ip netaddr.IP) (ps *ipnstate.PeerStatus, ok bool) {
+	for _, ps := range st.Peer {
+		for _, ip2 := range ps.TailscaleIPs {
+			if ip == ip2 {
+				return ps, true
+			}
+		}
+	}
+	return nil, false
+}
+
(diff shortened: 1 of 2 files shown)


cmd/tailscale/cli/up.go
```diff
@@ -193,12 +193,38 @@ func calcAdvertiseRoutes(advertiseRoutes string, advertiseDefaultRoute bool) ([]
...
+// address they want to use as an exit node.
 func exitNodeIPOfArg(arg string, st *ipnstate.Status) (ip netaddr.IP, err error) {
 	if arg == "" {
 		return ip, errors.New("invalid use of exitNodeIPOfArg with empty string")
 	}
 	ip, err = netaddr.ParseIP(arg)
 	if err == nil {
+		// If we're online already and have a netmap, double check that the IP
+		// address specified is valid.
+		if st.BackendState == "Running" {
```
maisem, line 219: Do we have to check for running in the other code path as well?

maisem (approved):

cmd/tailscale/cli/up.go
bradfitz: Nope. Normally it's allowed to set an arbitrary IP address when you're not running and there's no netmap. The backend will lazily remap a Tailscale IP to a Node StableID.

So this check is saying "If we're Running and we know the state of the world, then check here to fail them early."

The check below is trying to get any IP at all to give to the backend, which won't accept a name. It doesn't matter what state we're in. If can map it in Stopped state, that's good enough.

PR #3492, opened 2021-12-03
ipn/ipnlocal, net/tsdial: make SOCKS/HTTP dials use ExitDNS

### LINKED ISSUE #3475: tsdial: use ExitDNS when SOCKS/HTTP dialing

bradfitz:
And simplify, unexport some tsdial/netstack stuff in the the process.

Fixes #3475


maisem (approved):
Can you make the code scanning alert go away?

PR #3499, opened 2021-12-03
tstest/integration: allow tests to pass on machines also running the macOS GUI

josharian:
- all: minor code cleanup
- tstest/integration: fix running with -verbose-tailscale
- safesocket: let callers control whether to dial IPNExtension as a fallback

### DIFF
--- a/safesocket/safesocket.go
+++ b/safesocket/safesocket.go
@@ -57,10 +57,65 @@ func tailscaledStillStarting() bool {
 	return tailscaledProcExists()
 }
 
-// Connect connects to either path (on Unix) or the provided localhost port (on Windows).
-func Connect(path string, port uint16) (net.Conn, error) {
+// A ConnectionStrategy is a plan for how to connect to tailscaled or equivalent (e.g. IPNExtension on macOS).
+type ConnectionStrategy struct {
+	// For now, a ConnectionStrategy is just a unix socket path, a TCP port,
+	// and a flag indicating whether to try fallback connections options.
+	path     string
+	port     uint16
+	fallback bool
+	// Longer term, a ConnectionStrategy should be an ordered list of things to attempt,
+	// with just the information required to connection for each.
+	//
+	// We have at least these cases to consider (see issue 3530):
+	//
+	//   tailscale sandbox | tailscaled sandbox | OS      | connection
+	//   ------------------|--------------------|---------|-----------
+	//   no                | no                 | unix    | unix socket
+	//   no                | no                 | Windows | TCP/port
+	//   no                | no                 | wasm    | memconn
+	//   no                | Network Extension  | macOS   | TCP/port/token, port/token from lsof
+	//   no                | System Extension   | macOS   | TCP/port/token, port/token from lsof
+	//   yes               | Network Extension  | macOS   | TCP/port/token, port/token from readdir
+	//   yes               | System Extension   | macOS   | TCP/port/token, port/token from readdir
+	//
+	// Note e.g. that port is only relevant as an input to Connect on Windows,
+	// that path is not relevant to Windows, and that neither matters to wasm.
+}
+
+// DefaultConnectionStrategy returns a default connection strategy.
+// The default strategy is to attempt to connect in as many ways as possible.
+// It uses path as the unix socket path, when applicable,
+// and defaults to WindowsLocalPort for the TCP port when applicable.
+// It falls back to auto-discovery across sandbox boundaries on macOS.
+// TODO: maybe take no arguments, since path is irrelevant on Windows? Discussion in PR 3499.
+func DefaultConnectionStrategy(path string) *ConnectionStrategy {
+	return &ConnectionStrategy{path: path, port: WindowsLocalPort, fallback: true}
+}
+
+// UsePort modifies s to use port for the TCP port when applicable.
+// UsePort is only applicable on Windows, and only then
+// when not using the default for Windows.
+func (s *ConnectionStrategy) UsePort(port uint16) {
+	s.port = port
+}
+
+// UseFallback modifies s to set whether it should fall back
+// to connecting to the macOS GUI's tailscaled
+// if the Unix socket path wasn't reachable.
+func (s *ConnectionStrategy) UseFallback(b bool) {
+	s.fallback = b
+}
+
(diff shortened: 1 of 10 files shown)


bradfitz:
Bool params are always kinda gross. Can we do something else?

josharian:
> Bool params are always kinda gross. Can we do something else?

I knew you'd say that. I did such a nice job of making the call sites pretty and all. :P

I guess I can make an enum or a bit flag or something.

josharian:
> I knew you'd say that.

I can hear my Parent Voice in my head saying: "If you knew that, why did you do it?"

Stupid Parent Voice.

bradfitz:
What about my ol' favorite: `TS_DEBUG_CONNECT_TO_MACEXT=false` that the integration tests can set early for all subprocesses?

Then no API changes.

There's precedent:

```
        cmd.Env = append(os.Environ(),
                "TS_LOG_TARGET="+n.env.LogCatcherServer.URL,
                "HTTP_PROXY="+n.env.TrafficTrapServer.URL,
                "HTTPS_PROXY="+n.env.TrafficTrapServer.URL,
                "TS_DEBUG_TAILSCALED_IPN_GOOS="+ipnGOOS,
                "TS_LOGS_DIR="+t.TempDir(),
        )
```

josharian:
Could, I guess, but I think the new behavior is preferable.

bradfitz:
> Could, I guess, but I think the new behavior is preferable.

What new behavior? The new API? I didn't like safesocket taking an int and string as it is, as one is always useless. Another one is even worse. I really don't want more arguments. If anything I'd take fewer.

A new func would be fine, though, if you need `ConnectSocket` to only take one string (no int, no bool) and connect exactly to that string socket without fallback.

bradfitz:
... and that environment variable to modify the behavior of Connect for child processes if needed.

josharian:
> What new behavior?

The new behavior that if you pass a --socket argument on the command line it doesn't fall back to IPNExtension. I'll add a new func.

bradfitz:
> The new behavior that if you pass a --socket argument on the command line it doesn't fall back to IPNExtension. I'll add a new func.

Oh, I agree that's nice. But I'd even want that even if the `--socket` is given explicitly and has the default value, though.

josharian:
I tried making this into a neatly factored "try X then Y then Z" style API and it got hairy. So I fell back to wrapping up the inputs in an opaque type, as we discussed.

> I'd even want that even if the --socket is given explicitly and has the default value, though.

Sadly, I don't believe the package `flags` API lets us distinguish between a user providing the default value to a flag vs the user not using the flag at all.

bradfitz:
> Sadly, I don't believe the package flags API lets us distinguish between a user providing the default value to a flag vs the user not using the flag at all.

It does:
https://pkg.go.dev/flag#FlagSet.Visit
demo:
https://go.dev/play/p/YinuC_lyJgH

safesocket/safesocket.go
```diff
@@ -57,10 +57,60 @@ func tailscaledStillStarting() bool {
...
+func DefaultConnectionStrategy(path string) *ConnectionStrategy {
+	return &ConnectionStrategy{path: path, port: WindowsLocalPort, fallback: true}
+}
+
+// UsePort modifies s to use port for the TCP port when applicable.
+func (s *ConnectionStrategy) UsePort(port uint16) {
+	s.port = port
+}
+
+// UseFallback modifies s to fall back if and only if b.
```
bradfitz, line 100: "if and only if" is a little weird in the context of "fall back", which implies it won't always do it always

// UseFallback modifies s to set whether it should fall back to connecting to the macOS GUI's Tailscaled
// if the Unix socket path wasn't reachable.

Or something.

safesocket/safesocket.go
```diff
@@ -57,10 +57,60 @@ func tailscaledStillStarting() bool {
...
+	// Note e.g. that port is only relevant as an input to Connect on Windows,
+	// that path is not relevant to Windows, and that neither matters to wasm.
+}
+
+// DefaultConnectionStrategy returns a default connection strategy.
+// The default strategy is to attempt to connect in as many ways as possible.
+// It uses path as the unix socket path, when applicable,
+// and defaults to WindowsLocalPort for the TCP port when applicable.
+// It falls back to auto-discovery across sandbox boundaries on macOS.
+func DefaultConnectionStrategy(path string) *ConnectionStrategy {
```
bradfitz, line 91: little weird that on Windows it requires this path that's ignored.

how about not taking any arguments here and using the default, and then if they set the unix socket path, then they can use ExactPath instead?

But already an improvement, so fine as is too.

safesocket/safesocket.go
```diff
@@ -57,10 +57,60 @@ func tailscaledStillStarting() bool {
...
+	return &ConnectionStrategy{path: path, port: WindowsLocalPort, fallback: true}
+}
+
+// UsePort modifies s to use port for the TCP port when applicable.
```
bradfitz, line 95: mention only applicable on Windows?

safesocket/safesocket.go
```diff
@@ -57,10 +57,60 @@ func tailscaledStillStarting() bool {
 	return tailscaledProcExists()
 }
 
-// Connect connects to either path (on Unix) or the provided localhost port (on Windows).
-func Connect(path string, port uint16) (net.Conn, error) {
+// A ConnectionStrategy is a plan for how to connect to tailscaled.
```
bradfitz, line 60: to tailscaled, or equivalent (e.g. IPNExtension on macOS)

bradfitz (approved):

safesocket/safesocket.go
josharian: Done.

josharian:
> It does:

Ooooh! Done.

safesocket/safesocket.go
josharian: done

safesocket/safesocket.go
josharian: donedone

safesocket/safesocket.go
josharian: We have something to pass at every callsite, so I'm going to leave as-is for now. I'll add a TODO.

PR #3503, opened 2021-12-04
wgengine/filter: let unknown IPProto match if IP okay & match allows all ports

bradfitz:


crawshaw (approved):
While here, should we add some more obscure protos to this filter?

bradfitz:
Like what? This covers all protocols. You want port parsers for certain ones? Which?

tinoue:
Excellent solution. Thaks!

amosbird:
Hmm, with this PR I still cannot make ipip or gre tunnel work under two tailscale endpoints. Do I need special configuration?

DentonGentry:
An ACL rule allowing `"dst": "something:*"` is expected to allow IPinIP and GRE through. If that isn't working please open an issue, we may not notice comments on closed pull requests.

PR #3508, opened 2021-12-06
support running in a FreeBSD jail

tzneal:
Since devd apparently can't be made to work in a FreeBSD jail
fall back to polling.

Fixes tailscale#2858

### DIFF
--- a/wgengine/monitor/monitor_freebsd.go
+++ b/wgengine/monitor/monitor_freebsd.go
@@ -25,10 +25,11 @@ type devdConn struct {
 	conn net.Conn
 }
 
-func newOSMon(logf logger.Logf, _ *Mon) (osMon, error) {
+func newOSMon(logf logger.Logf, m *Mon) (osMon, error) {
 	conn, err := net.Dial("unixpacket", "/var/run/devd.seqpacket.pipe")
 	if err != nil {
-		return nil, fmt.Errorf("devd dial error: %v", err)
+		logf("devd dial error: %v, falling back to polling method", err)
+		return newPollingMon(logf, m)
 	}
 	return &devdConn{conn}, nil
 }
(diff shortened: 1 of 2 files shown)


tzneal:
I verified this in a jail on a TrueNAS install.  I had started looking into making a plugin for TrueNAS and this was step one.

tzneal:
And after some more experimentation,  it can't really be used the way I wanted to which was to make the NAS itself available or perform subnet routing.  You only get access to the jail, and subnet routing is Linux only for now.  

The only use case I can think of, which is still somewhat interesting though, is using tailscale as a sort of 'sidecar' in an existing jail that you are running services on to make those services available.

bradfitz:
FreeBSD does subnet routing using gvisor's netstack, in userspace. See `tailscaled.go`'s:

```go
func shouldWrapNetstack() bool {
        if e := os.Getenv("TS_DEBUG_WRAP_NETSTACK"); e != "" {
                v, err := strconv.ParseBool(e)
                if err != nil {
                        log.Fatalf("invalid TS_DEBUG_WRAP_NETSTACK value: %v", err)
                }
                return v
        }
                return true
        }
        switch runtime.GOOS {
        case "windows", "darwin", "freebsd":
                // Enable on Windows and tailscaled-on-macOS (this doesn't                                                                                                                    
                // affect the GUI clients), and on FreeBSD.                                                                                                                                   
                return true
        }
        return false
}
```

But it won't do ICMP; only TCP & UDP. Have you tried it?

tzneal:
@bradfitz  You're right, it does work, thanks!  I had read on https://tailscale.com/kb/1019/subnets/ that "Currently, we only support Linux devices as subnet routers. We plan to support this feature on Windows and macOS in the future."  and assumed it didn't work without testing it.

josharian:
Hi, Todd! Nice to see you here. :)

(We tinkered together on the Go compiler some years back.)

tzneal:
> Hi, Todd! Nice to see you here. :)
> 
> (We tinkered together on the Go compiler some years back.)

Hey Josh, you too! That's actually how I found out about tailscale.  Several of the really smart guys I knew from working on the go compiler were going to work on tailscale, so I knew it had to be interesting :laughing:

PR #3509, opened 2021-12-06
taildrop: add --progress flag to 'tailscale file cp'

andrew-d:
This PR adds a `--progress` flag to the `tailscale file cp` command, which prints the upload progress (at a rate of one status report per second).

I found myself wanting this today while using taildrop with a multi-GB file on a moderately fast connection, so I figured that I'd submit this as a proposal; I'm happy to make changes to the log format or the general approach depending on what you folks prefer.


cmd/tailscale/cli/file.go
```diff
@@ -156,6 +159,12 @@ func runCp(ctx context.Context, args []string) error {
 		if cpArgs.verbose {
 			log.Printf("sending %q to %v/%v/%v ...", name, target, ip, stableID)
 		}
+		if cpArgs.progress {
+			rc := progress.New(fileContents, contentLength, time.Second)
+			defer rc.Close()
```
josharian, line 164: This is happening in a loop, processing one file after another. `defer` in a loop smells a bit weird. If you use this to `cp` multiple files, does the progress bar do the right thing?

util/progress/reader.go
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
```
josharian, line 1: Rather than rolling our own progress bar, can we use an existing, well-tested one? https://github.com/schollz/progressbar looks popular.

For code size purposes, we will probably also want to build tag any progress-bar-related stuff, so that it doesn't get compiled in on mobile, where it will never be used.

josharian (commented):
I left a few code level comments. I'm agnostic as to whether we should have a progress bar at all. Before working on the code any more, we should decide that question.

Opinions on that, @bradfitz @danderson et al?

util/progress/reader.go
andrew-d: Yep, I'm happy to switch to a third-party dependency if you'd prefer that.  Do you want me to push a commit using that library?

cmd/tailscale/cli/file.go
andrew-d: Good call; yes, this does misbehave. Just pushed a commit fixing this and tested locally with two files.

josharian:
Sorry for the delay. We've been discussing internally, and I've been distracted.

Consensus seems to be a reluctant yes on progress bars, given:

* only when we're 100% sure stdout is a terminal, so we don't write ascii art into pipes
* no color/ANSI unless we're absolutely sure the escapes are going to work
* no updates faster than about once a second

Alternatives considered:

* report our progress to a daemon/wrapper that did it if desired
* print status on SIGINFO/SIGUSR1

I'll have to go look at the third party packages code to decide whether we're OK taking a dependency on them.

Ball's in my court, but realistically replies will be slow. Apologies in advance for that.

caquino:
I wonder if for most users, utilizing `pv` would be enough to replace this functionality.

The only downside I can see is that the resulting file is named `stdin.something` , but if I could put the filename after colon it would solve the problem.

the commands I used are:

single file copy
`pv source-file | tailscale file cp - remote-machine:`

multiple files
`tar -czf - /source-folder| (pv -p --timer --rate --bytes | tailscale file cp - remote-machine:)`

pv-cp

Gottarocket:
If this is implemented would love a change in android app too. Yes it shows progress but when we touch any part of the screen or if we go to home screen, the send-to popup disappears. A notification on top showing progress is appreciated  so we can know progress when there is no popup.

PR #3518, opened 2021-12-06
Makefile: add spk and pushspk targets for iterative Synology development

bradfitz:
(using the new https://github.com/tailscale/tailscale-synology/pull/94)


bradfitz:
Once you have root@your-synology's $HOME/.ssh/authorized_keys set up, then you can just run:

```
$ SYNOHOST=10.2.3.4 make pushspk
```

And push changes live to your Synology in seconds.

maisem (approved):

PR #3524, opened 2021-12-06
go.mod: mass update with go get -u.

danderson:
Gets ahead of dependabot slightly, but the updates are minor.


bradfitz (approved):

crawshaw (approved):
Does that go.sum file need a `go mod tidy`?

danderson:
It's already tidied - or at least, `go mod tidy` won't shrink it any further.

Issue #2179, opened 2021-06-20
Taildrop on Synology


simondarren:
Could not find the drop. Turned on home dir.

DentonGentry:
The available releases are at: https://github.com/tailscale/tailscale-synology/releases

1. In the Synology DSM web admin UI, open the Package Center.
1. Press the Manual install button and provide the SPK file downloaded from https://github.com/tailscale/tailscale-synology/releases
1. Follow the wizard until done.
1. At this point tailscaled should be up and running.
1. SSH into the machine, and run sudo tailscale up to get a URL, visit this URL with a browser to authenticate the Synology NAS.

bradfitz:
I interpreted the super-minimal bug report about being unable to find the files Taildropped to a Synology NAS.

@simondarren, could you elaborate?

simondarren:
ty @bradfitz yes.  

@DentonGentry I have it running on my Synology fine. I just could not locate the taildrop. I get no error on the drop, I see nothing in the synology logs.

bradfitz:
Only on Mac and Windows do we automatically put the files somewhere useful.

On Linux (which includes Synology) they remain in the tailscaled-owned /var directory until they're "picked up".

But I suppose we could/should do something better on Synology, perhaps moving them to the user's home directory, or a "Tailscale" subdirectory thereof.

jpm:
@bradfitz I don't know how much you guys are willing to customize this behavior on Synology, but it would be very nice if there was some configuration parameter so users can set the location for the drops. I would perhaps choose my existing "Downloads" shared folder, maybe on a "Tailscale" subdirectory or something.

We could then add the write permission to the `tailscale` user on that shared folder, and it's all set up. That would allow the software to inject the files into the "regular" flow of how files are stored / shared within DSM.

jpm:
@bradfitz That sounds great.

bradfitz:
Sent https://github.com/tailscale/tailscale/pull/3525 implementing this.

DentonGentry:
Fixed in the 1.18.2 Synology release by https://github.com/tailscale/tailscale/commit/6c44133d8f1cfe06c33ac924daf425d8dc7f5fab. You can sideload it from https://github.com/tailscale/tailscale-synology/releases now, or we expect it to appear in the Package Center in (hopefully) January.

PR #3525, opened 2021-12-06
ipn/{ipnserver,ipnlocal}: support incoming Taildrop on Synology

bradfitz:
If the user has a "Taildrop" shared folder on startup and
the "tailscale" system user has read/write access to it,
then the user can "tailscale file cp" to their NAS.

Updates #2179 (would be fixes, but not super ideal/easy yet)

### DIFF
--- a/ipn/ipnserver/server.go
+++ b/ipn/ipnserver/server.go
@@ -759,6 +759,18 @@ func New(logf logger.Logf, logid string, store ipn.StateStore, eng wgengine.Engi
 	b.SetDecompressor(func() (controlclient.Decompressor, error) {
 		return smallzstd.NewDecoder(nil)
 	})
+	if distro.Get() == distro.Synology {
+		// See if they have a "Taildrop" share.
+		// See https://github.com/tailscale/tailscale/issues/2179#issuecomment-982821319
+		path, err := findSynologyTaildropDir()
+		if err != nil {
+			logf("Synology Taildrop support: %v", err)
+		} else {
+			logf("Synology Taildrop: using %v", path)
+			b.SetDirectFileRoot(path)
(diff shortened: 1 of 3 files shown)


ipn/ipnserver/server.go
```diff
@@ -1114,3 +1126,17 @@ func (ln *listenerWithReadyConn) Accept() (net.Conn, error) {
...
 	return ln.Listener.Accept()
 }
+
+// findSynologyTaildropDir looks for the first volume containing a
+// "Taildrop" directory.  We'd run "synoshare --get Taildrop" command
+// but on DSM7 at least, we lack permissions to run that.
+func findSynologyTaildropDir() (dir string, err error) {
+	const name = "Taildrop"
+	for i := 1; i <= 16; i++ {
+		dir = fmt.Sprintf("/volume%v/%s", i, name)
```
DentonGentry, line 1136: The path to the tailscale application has turned out to vary depending on whether the Synology system was first installed with DSM6/7 or if it started out life with a much earlier version and was repeatedly upgraded. Synology seems to keep putting app files in the location where that really old version first put them.

Do we have any reason to believe that volume handling would be similar, maybe different on some devices?

I guess the best way to proceed is ship it and see if there are systems where it doesn't work.

DentonGentry (approved):

ipn/ipnserver/server.go
bradfitz: > Do we have any reason to believe that volume handling would be similar, maybe different on some devices?

I've never seen anything but /volumeN/$NAME on any of my Synology devices on either DSM 6 or 7.

But nothing surprises me anymore. Wait and see I guess?

It tried to do it the "right" way but I got access denied querying that info, so not sure what else to do.

PR #3527, opened 2021-12-06
ipn/ipnlocal: add HTTP/2 h2c server support to peerapi on non-mobile platforms

bradfitz:
To make ExitDNS cheaper.

Might not finish client-side support in December before 1.20, but at
least server support can start rolling out ahead of clients being
ready for it.

Tested with curl against peerapi.

Updates #1713


crawshaw (approved):
Huh, I've not heard of H2C before. I tried googling it and the first link is about it accidentally being invented as a vulnerability in H2 servers? https://www.darkreading.com/edge-articles/h2c-smuggling-a-new-devastating-kind-of-http-request

PR #3532, opened 2021-12-08
Support "Taildrop" on TrueNAS

tzneal:
I was experimenting with TrueNAS for another reason and happened to see Brad's change to add this support for dropping files directly to Synology store.  This implements the same feature for TrueNAS.  I don't feel strongly about it and may not be using TrueNAS much in the future, but figured I'd mail the change in case it was wanted.  This was tested on both FreeBSD TrueNAS Core and Debian/Linux TrueNAS Scale.

### DIFF
--- a/ipn/ipnserver/server.go
+++ b/ipn/ipnserver/server.go
@@ -759,17 +759,21 @@ func New(logf logger.Logf, logid string, store ipn.StateStore, eng wgengine.Engi
 	b.SetDecompressor(func() (controlclient.Decompressor, error) {
 		return smallzstd.NewDecoder(nil)
 	})
-	if distro.Get() == distro.Synology {
+
+	dg := distro.Get()
+	switch dg {
+	case distro.Synology, distro.TrueNAS:
 		// See if they have a "Taildrop" share.
 		// See https://github.com/tailscale/tailscale/issues/2179#issuecomment-982821319
-		path, err := findSynologyTaildropDir()
+		path, err := findTaildropDir(dg)
 		if err != nil {
-			logf("Synology Taildrop support: %v", err)
+			logf("%s Taildrop support: %v", dg, err)
 		} else {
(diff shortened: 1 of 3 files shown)


version/distro/distro.go
```diff
@@ -70,6 +73,7 @@ func freebsdDistro() Distro {
 	case have("/usr/local/sbin/opnsense-shell"):
 		return OPNsense
 	case have("/usr/local/bin/freenas-debug"):
+		// TrueNAS Core runs on FreeBSD
```
DentonGentry, line 76: If this is an area you are familiar with: is there something we could use to distinguish TrueNAS from FreeNAS?

Should we even try, or are they treated interchangeably in forums / software packages / etc ?

tzneal: I'm sorry, but I'm not 100% sure, I'm new to TrueNAS.  From what I can tell though, TrueNAS Core is the new name for FreeNAS as of a few months ago (https://www.truenas.com/freenas/).

ipn/ipnserver/server.go
```diff
@@ -1140,3 +1154,26 @@ func findSynologyTaildropDir() (dir string, err error) {
...
+		return dir, nil
+	}
+
+	// but if running on the host, it may be something like /mnt/Primary/Taildrop
+	fis, err := ioutil.ReadDir("/mnt")
+	if err != nil {
+		return "", fmt.Errorf("error reading /mnt: %w", err)
+	}
+	for _, fi := range fis {
+		dir = fmt.Sprintf("/mnt/%s/%s", fi.Name(), name)
```
bradfitz, line 1173: hopefully this won't trigger some NFS automounter that then hangs and this stat never completes

(past NFS scars)

But looks plausible! :)

bradfitz (approved):
Looks great, thanks!

PR #3534, opened 2021-12-08
ipn/ipnstate, cmd/tailscale: add Online bool to tailscale status & --json

### LINKED ISSUE #3533: `tailscale status --json` returns `0001-01-01T00:00:00Z` for node `LastSeen`
### What is the issue?

Running `tailscale status --json` often returns `0001-01-01T00:00:00Z` for `LastSeen`, both under `Self` and `Peer`. It would appear that if a machine has been disconnected for a little while then it will return a sensible timestamp (the time when it disconnected), but if the machine is online it will return `0001-01-01T00:00:00Z`. Note that the Tailscale dashboard (`login.tailscale.com/admin/machines`) always shows a sensible `lastSeen`: the API this dashboard uses (`https://login.tailscale.com/admin/api/machines`) seems to return the correct timestamps.

Also note that `LastHandshake` and `LastWrite` appear to exhibit a similar issue.

### Steps to reproduce

Connect a few machines to Tailscale and run `tailscale status --json` on one of the machines. You should see `"LastSeen": "0001-01-01T00:00:00Z"` in places where you'd expect a timestamp matching what can be seen on the dashboard (e.g. if a machine is connected, the timestamp should be current-ish). 

_No response_

### OS

Linux

### OS version

Debian 11.1

### Tailscale version

1.18.1

### Bug report

BUG-3342af652b3e43b9cd37ee32ba8f0a5aa33e81a96f6681823126c3b30346c200-20211208142817Z-f88bd32615d64f4f

bradfitz:
Fixes #3533

### DIFF
--- a/cmd/tailscale/cli/status.go
+++ b/cmd/tailscale/cli/status.go
@@ -29,7 +29,22 @@ var statusCmd = &ffcli.Command{
 	Name:       "status",
 	ShortUsage: "status [--active] [--web] [--json]",
 	ShortHelp:  "Show state of tailscaled and its connections",
-	Exec:       runStatus,
+	LongHelp: strings.TrimSpace(`
+
+JSON FORMAT
+
+Warning: this format has changed between releases and might change more
+in the future.
+
+For a description of the fields, see the "type Status" declaration at:
+
+https://github.com/tailscale/tailscale/blob/main/ipn/ipnstate/ipnstate.go
+
+(and be sure to select branch/tag that corresponds to the version
+ of Tailscale you're running)
+
+`),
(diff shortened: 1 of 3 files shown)


bradfitz:
Looks like this:

Screen Shot 2021-12-08 at 9 10 27 AM

crawshaw (approved):
I think we should document somewhere what we mean by "offline". Maybe in the `--help`?

bradfitz:
> I think we should document somewhere what we mean by "offline". Maybe in the --help?

Added pointers to docs in --help.

PR #3538, opened 2021-12-08
ipn/ipnstate: remove old deprecated TailAddr IPv4-only field

bradfitz:
It's been a bunch of releases now since the TailscaleIPs slice
replacement was added.


crawshaw:
But which release? Are all those clients gone? If not, upgrading one to the latest tailscale will break it

bradfitz:
> But which release? Are all those clients gone? If not, upgrading one to the latest tailscale will break it

I don't follow.

This is for `tailscale status --json`. Nothing about the control plane.

This deprecated field name is synthesized on the client in tailscaled for the CLI, but the better named TailscaleIPs slice with IPv6 support and a full name (rather than TailAddr) was added in like 1.8.

crawshaw (approved):
oh i see

Issue #3540, opened 2021-12-09
RHEL install error - Bad GPG signature


JayWStapleton:
### What is the issue?

https://tailscale.com/install.sh fails on a new, updated RHEL8 install with error:

> Error: Failed to download metadata for repo 'tailscale-stable': repomd.xml GPG signature verification error: Bad GPG signature

### Steps to reproduce

$ curl -fsSL https://tailscale.com/install.sh | sh

Installing Tailscale for rhel 8, using method dnf

+ sudo dnf config-manager --add-repo https://pkgs.tailscale.com/stable/rhel/8/tailscale.repo

Updating Subscription Management repositories.

Adding repo from: https://pkgs.tailscale.com/stable/rhel/8/tailscale.repo

+ sudo dnf install tailscale

Updating Subscription Management repositories.

Tailscale stable                                                                                                   860  B/s | 800  B     00:00    

Tailscale stable                                                                                                   6.9 kB/s | 3.1 kB     00:00    

Importing GPG key 0x957F5868:

 Userid     : "Tailscale Inc. (Package repository signing key) <info@tailscale.com>"

 Fingerprint: 2596 A99E AAB3 3821 893C 0A79 458C A832 957F 5868

 From       : https://pkgs.tailscale.com/stable/rhel/8/repo.gpg

Is this ok [y/N]: Tailscale stable                                           [            ===                                      ] ---  B/s |   0Tailscale stable                                                                                                   880  B/s | 800  B     00:00    

Error: Failed to download metadata for repo 'tailscale-stable': repomd.xml GPG signature verification error: Bad GPG signature

### OS

Linux

### OS version

Red Hat Enterprise Linux 8

### Tailscale version

1.18

crawshaw:
I'll take a look at this

crawshaw:
I believe the GPG error is an intermittent failure that when RHEL has updated itself goes away. Other than that, I think we need to auto-answer yes to installing tailscale.

DentonGentry:
Substantially fixed by https://github.com/tailscale/tailscale/pull/3541

9034725985:
If you see this error on 

`time sudo dnf offline-upgrade download -y && time sudo dnf offline-upgrade reboot`, try `dnf check-update` 

```bash dnf check-update
Tailscale stable                                                                                                                                                                                              1.1 kB/s | 800  B     00:00    
Tailscale stable                                                                                                                                                                                               19 kB/s | 3.1 kB     00:00    
[...]

PR #3541, opened 2021-12-09
scripts/installer.sh: press Y on RHEL

crawshaw:
For #3540


DentonGentry (approved):
This sounds REALLY familiar, like we did something like this already or maybe have another PR open. I can't find another PR though.

PR #3563, opened 2021-12-14
Proposed 1.18.3 branch

bradfitz:
cherry-picks w/ @DentonGentry


josharian:
Maybe https://github.com/tailscale/tailscale/commit/ca1b3fe235ca617c4128fcc00d924e85535b4b8e?

Reviewing the rest of the list now...

josharian:
What's the rationale behind including https://github.com/tailscale/tailscale/pull/3563/commits/6c44133d8f1cfe06c33ac924daf425d8dc7f5fab?

Seems like a new feature not a bug-fix, and Synology is on a weird release cycle anyway.

josharian:
Similar question for https://github.com/tailscale/tailscale/pull/3563/commits/972bcccc3603c8ab57aaaedfc8f73d29e447a681 -- seems more like a new feature than a bug fix, and the filter code is on the security-sensitive side. But maybe it fixes a bug that's just not linked from the commit message?

Anyway, that's all my kibitzing. LGTM once you both are happy with the set of cherry-picks, after considering the input above.

bradfitz:
> What's the rationale behind including 6c44133?
>
> Seems like a new feature not a bug-fix, and Synology is on a weird release cycle anyway.

Because we need to get Synology a 1.18.x build for their weird release cycle, and it's pretty low risk. (only affects Synology)

bradfitz:
> Similar question for 972bccc -- seems more like a new feature than a bug fix, and the filter code is on the security-sensitive side. But maybe it fixes a bug that's just not linked from the commit message?

It's indeed borderline but we decided to include it because it's a safe new code path (that only is invoked if it was going to reject a packet anyway) and it'd help somebody that currently needs it.

bradfitz:
> Maybe ca1b3fe?

Definitely! That's actually the one I meant to get first and got distracted.

Issue #3570, opened 2021-12-15
Windows: tailscaled service never logs to file


dblohm7:
### What is the issue?

The service `tailscaled`'s log entries are uploaded, but they are not written to the `tailscale-service` log files on disk (only the `tailscaled` subprocess's log entries are written to disk). This is bad for troubleshooting errors when the `tailscaled` service itself cannot dial out; those errors never make it into the log file.

I have a PR ready that initializes the `filelogger` during `logpolicy` init and plugs it in ahead of `logtail`.

### OS

Windows

### Tailscale version

1.18.0

PR #3571, opened 2021-12-15
ipn/ipnserver, log/filelogger, logpolicy: when tailscaled is running …

dblohm7:
…as a windows service, ensure the service's log messages are written to file

This patch moves the Windows-only initialization of the filelogger into logpolicy.
Previously we only did it when babysitting the tailscaled subprocess, but this meant
that log messages from the service itself never made it to disk. This meant that
if logtail could not dial out, its log messages would be lost.

I modified filelogger.New to work a bit differently and added a `maybeWrapForPlatform`
to logpolicy to ensure that the filelogger is plugged in front of logtail ASAP.

Fixes https://github.com/tailscale/tailscale/issues/3570


bradfitz:
Let's chat about all this when you're around.

bradfitz:
> log messages from the service itself never made it to disk.

Which log messages are those?

> This meant that if logtail could not dial out, its log messages would be lost.

It's intentional that logtail's errors (as used by the logwriter backoff `bo` field) never loop back into that log writer, otherwise it'd amplify. This is purely about those logtail dial error messages not going to disk?

dblohm7:
> Which log messages are those?

Any log message generated by the parent service.

> It's intentional that logtail's errors (as used by the logwriter backoff `bo` field) never loop back into that log writer, otherwise it'd amplify.

That makes sense. I do find it hard to get my head around all the complexity w/ respect to who is plugged into what.

> This is purely about those logtail dial error messages not going to disk?

That would be the main impetus, yes, though there are some other useful entries that wouldn't make it to disk (such as DNS cache flushing upon screen unlock, when enabled)

dblohm7:
Though I don't think that's a concern here, as the filelogger is essentially a tee off the existing setup.

log/filelogger/log.go
```diff
@@ -26,30 +27,30 @@ const (
 	maxFiles = 50
 )
 
-// New returns a logf wrapper that appends to local disk log
+// New returns a Writer that appends to local disk log
 // files on Windows, rotating old log files as needed to stay under
 // file count & byte limits.
-func New(fileBasePrefix, logID string, logf logger.Logf) logger.Logf {
+func New(fileBasePrefix, logID string, inner *log.Logger) io.Writer {
```
bradfitz, line 33: so, I like everything about this PR (like moving the file logging up from ipnserver) except this signature change. An io.Writer is logically a byte stream and not record oriented like logf is. Generally when an io.Writer is used record-wise (or a passed in one has that constraint), it's documented, like the `log` package saying:

> Each logging operation makes a single call to the Writer's Write method.

That'd be one option, to just document it here.

But I'd prefer to stay in the `logf` type world if possible.

I see why that's tricky, though... logpolicy is what generates the LogID.

Still thinking.

bradfitz: What do you think about #3582?

PR #3573, opened 2021-12-15
ipn/ipnlocal: deflake (mostly) TestStateMachine

bradfitz:
I'm sick of this flaking. Even if this isn't the right fix, it
stops the alert fatigue.

Updates #3020

### DIFF
--- a/ipn/ipnlocal/state_test.go
+++ b/ipn/ipnlocal/state_test.go
@@ -87,8 +87,9 @@ func (nt *notifyThrottler) drain(count int) []ipn.Notify {
 type mockControl struct {
 	tb         testing.TB
 	opts       controlclient.Options
-	logf       logger.Logf
+	logfActual logger.Logf
 	statusFunc func(controlclient.Status)
+	preventLog syncs.AtomicBool
 
 	mu          sync.Mutex
 	calls       []string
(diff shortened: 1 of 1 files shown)


bradfitz:
This almost always passes now:

```
go test -v -failfast -run=StateMachine -count=100 ./ipn/ipnlocal/
```

Before it could rarely pass with 20.

josharian (approved):
I'd prefer t.Cleanup over defer, but OK as-is.

crawshaw (approved):

bradfitz:
I prefer defer when you don't need cleanup-at-a-distance, which is what I think of t.Cleanup as for... for test helpers that need to register some cleanup. Otherwise I think defer is nicely explicit.

Is there any advantage of t.Cleanup I don't know of?

josharian:
> Is there any advantage of t.Cleanup I don't know of?

Mainly:

* I don't have to think about where it's being called from, which makes refactoring and reviewing easier. I've noticed that as I work on tests I frequently end up converting them, usually as part of refactoring out common setup/teardown.
* It makes the intent clearer: It is intended to be test-level, not function-level, even though in this case those coincide.

But as I said, I don't feel strongly.

bradfitz:
Sure, done. Little uglier with the closure, but it's all ugly, so no real change :)

PR #3585, opened 2021-12-16
scripts/installer.sh fix APT_SYSTEMCTL_START error

JayWStapleton:
fix error: 'sh: 411: APT_SYSTEMCTL_START: parameter not set' on certain debian distributions

### DIFF
--- a/scripts/installer.sh
+++ b/scripts/installer.sh
@@ -24,6 +24,7 @@ main() {
 	VERSION=""
 	PACKAGETYPE=""
 	APT_KEY_TYPE="" # Only for apt-based distros
+	APT_SYSTEMCTL_START=false # Only needs to be true for Kali
 
 	if [ -f /etc/os-release ]; then
 		# /etc/os-release populates a number of shell variables. We care about the following:


DentonGentry:
To fix the DCO error you use: `git commit --amend --signoff`
Please make sure the email address it populates is the tailscale.com one. You can use `git config` to set that email address permanently.

scripts/installer.sh
```diff
@@ -24,6 +24,7 @@ main() {
 	VERSION=""
 	PACKAGETYPE=""
 	APT_KEY_TYPE="" # Only for apt-based distros
+	APT_SYSTEMCTL_START=true # Only for apt-based distros
```
DentonGentry, line 27: This should be false by default. Only Kali (so far) needs the quirky behavior.

JayWStapleton: As of today, it won't install on debian or Ubuntu without that flag set

DentonGentry: I'm saying, change it from "true" to "false".
Debian and Ubuntu don't need us to call systemctl after install, only Kali does.
The bug right now is that the variable is completely undefined, and we set -x at the top of the script.

DentonGentry (approved):

PR #3587, opened 2021-12-16
ipn/ipnlocal: make TestShrinkDefaultRoute hermetic

### LINKED ISSUE #3580: TestShrinkDefaultRoute is not hermetic, fails on some people's machines
@DentonGentry got this failure at e67182d44606d05ede8400514fa28b0bc7211f91 but only on one of his machines:

```
--- FAIL: TestShrinkDefaultRoute (0.00s)
    local_test.go:204: shrink("::/0").Contains(2602:xxxx:7b4a:ff60::xxxx) = false, want true
FAIL
FAIL    tailscale.com/ipn/ipnlocal      0.610s
```

Looks like it calls func `shrinkDefaultRoute` which uses `interfaceRoutes` which calls `interfaces.ForeachInterfaceAddress`, so it depends on your local machine's interface addresses.

Instead, `shrinkDefaultRoute` should take an extra param of the same func type of `interfaceRoutes` so the test can supply a fixed version.

josharian:
Make shrinkDefaultRoute a pure function.
Instead of calling interfaceRoutes, accept that information as parameters.
Hard-code those parameters in TestShrinkDefaultRoute.

Fixes #3580

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -1020,7 +1020,12 @@ func (b *LocalBackend) updateFilter(netMap *netmap.NetworkMap, prefs *ipn.Prefs)
 				// wifi": you get internet access, but to additionally
 				// get LAN access the LAN(s) need to be offered
 				// explicitly as well.
-				s, err := shrinkDefaultRoute(r)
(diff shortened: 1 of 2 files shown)


bradfitz (approved):
Thanks!

PR #3589, opened 2021-12-16
wgengine/netstack: don't remove 255.255.255.255/32 from netstack

bradfitz:
Updates #2642 (maybe fixes?)


crawshaw (approved):
I don't understand why at all (and if you have any idea, please add it to the commit comment), but sure.

bradfitz:
Added a bit to commit but I don't have a great idea other than it's not our address to remove and it added it itself and it's a magic IP and parts of netstack might depend on it.

PR #3592, opened 2021-12-17
logpolicy, ipn/ipnserver: connect to logtail via tailscaled when needed

bradfitz:
This is for use by the Windows GUI client to log via when an
exit node is in use, so the logs don't go out via the exit node and
instead go directly, like tailscaled's. The dialer tried to do that
in the unprivileged GUI by binding to a specific interface, but the
"Internet Kill Switch" installed by tailscaled for exit nodes
precludes that from working and instead the GUI fails to dial out.
So, go through tailscaled (with a CONNECT request) instead.

Fixes tailscale/corp#3169


ipn/ipnserver/proxyconnect.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		logf("error CONNECT dialing %v: %v", hostPort, err)
+		io.WriteString(c, "HTTP/1.1 502 Fail\r\n\r\nConnect failure.\n")
+		return
+	}
+	defer back.Close()
+
+	io.WriteString(c, "HTTP/1.1 200 OK\r\n\r\n")
+
+	errc := make(chan error, 1)
```
josharian, line 64: 2? Then the reader doesn't have to think about whether having multiple senders below is safe b/c of GC or what.

ipn/ipnserver/proxyconnect.go
```diff
@@ -0,0 +1,74 @@
...
+// exit node is in use, so the logs don't go out via the exit node and
+// instead go directly, like tailscaled's. The dialer tried to do that
+// in the unprivileged GUI by binding to a specific interface, but the
+// "Internet Kill Switch" installed by tailscaled for exit nodes
+// precludes that from working and instead the GUI fails to dial out.
+// So, go through tailscaled (with a CONNECT request) instead.
+func (s *Server) handleProxyConnectConn(ctx context.Context, br *bufio.Reader, c net.Conn, logf logger.Logf) {
+	defer c.Close()
+
+	c.SetReadDeadline(time.Now().Add(5 * time.Second))
```
josharian, line 31: Why is this the right duration? Something like: `// 5s ought to be long enough to make the initial connection`?

ipn/ipnserver/server.go
```diff
@@ -238,12 +238,26 @@ func bufferHasHTTPRequest(br *bufio.Reader) bool {
 		mem.Contains(mem.B(peek), mem.S(" HTTP/"))
 }
 
+// bufferIsConnect reports whether br looks like it's likely an HTTP
+// CONNECT request.
+func bufferIsConnect(br *bufio.Reader) bool {
+	peek, _ := br.Peek(br.Buffered())
```
josharian, line 244: Why Buffered instead of 4 (or `len("CONN")`)? As is it works b/c this gets called from serveConn that does a peek(4) already, but changing that might then break this...

ipn/ipnserver/server.go
```diff
@@ -238,12 +238,26 @@ func bufferHasHTTPRequest(br *bufio.Reader) bool {
...
+	return mem.HasPrefix(mem.B(peek), mem.S("CONN"))
```
josharian, line 245: I wish the compiler was good enough to make `strings`/`bytes` as nice as `mem` to use here.

josharian (approved):

ipn/ipnserver/server.go
bradfitz: Indeed.

ipn/ipnserver/proxyconnect.go
bradfitz: done

ipn/ipnserver/server.go
bradfitz: I added an "Invariant: ..." doc instead to the func's docs.

I don't want the Peek to cause any actual I/O to happen. I want to only use what's available. So I instead documented the caller must've already peeked at least 4 bytes.

ipn/ipnserver/proxyconnect.go
bradfitz: Done.

Issue #3363, opened 2021-11-21
iOS Tailscale "eats battery


ProfDrYoMan:
### What is the issue?

Device runtime cut by half running tailscale always on.
Did not happen with wirequard.

PS: I did check for old bug reports and solutions on that, but there seem not to be any.

### Steps to reproduce

Enable tailscale and wait.

### OS

iOS

### OS version

15.1

### Tailscale version

1.16.1

bradfitz:
The prior bug was #2442 but the poster closed it when they were happy enough.

But plenty more remains. It largely depends on some configuration (DNS settings, exit nodes, other stuff running/configured on iOS) but there's a lot we can do to improve things across the board.

In any case, we're working on it, and Tailscale 1.18.0 has a lot of new metrics to track it so we can gather data from our own devices & the field to see which fixes would be the most effective first.

We can keep this bug open, but a request to others who find this bug: please refrain from posting "this happens to me too!" or posting screenshots of your iOS battery usage graphs. Those aren't helpful. Use the "Subscribe" button on the right or use emoji reactions which don't send email or otherwise clutter the discussion here.

ProfDrYoMan:
Thanks for keeping this in mind as a serious need to improve!

Some info:
- I got one exit node which I route all traffic thru (always). Here it would be good if the app would allow to configure excludes, e.g., no VPN if connected to my known internal SSIDs.
- I am using a subnet route on the same node as the exit node which is connecting to my internal net.
- I am using the DNS feature to route all DNS traffic to my internal DNS (to allow pi-hole use always).

bradfitz:
@ProfDrYoMan, yeah, so you're basically hitting the worst case scenario. We're tracking all those things, though, so it should get better.

ProfDrYoMan:
What has got most impact? Can you order?
E.g. Exit Node > Subnet > DNS

Asking because I might be able to put the DNS virtually onto exit node which would remove the need for a subnet route.

pilif:
On my iPhone (iPhone 13 Pro, iOS 15.3.1), I see high background energy usage even when not connected, but the app still running. In this state, I would not think any networking issues to be the underlying cause.

Force-quitting the app naturally fixes the battery usage - until I start the app again.

bradfitz:
Sub-bugs of this one:

- [ ] https://github.com/tailscale/tailscale/issues/540
- [ ] ...

DentonGentry:
[...]

PR #3594, opened 2021-12-19
net/dns/resolver: send NXDOMAIN to iOS DNS-SD/Bonjour queries

bradfitz:
Don't just ignore them. See if this makes them calm down.

Updates #3363

### DIFF
--- a/net/dns/resolver/forwarder.go
+++ b/net/dns/resolver/forwarder.go
@@ -598,7 +598,17 @@ func (f *forwarder) forwardWithDestChan(ctx context.Context, query packet, respo
 	// out, playing on Sonos still works.
 	if hasRDNSBonjourPrefix(domain) {
 		metricDNSFwdDropBonjour.Add(1)
-		return nil
+		res, err := nxDomainResponse(query)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


net/dns/resolver/forwarder.go
```diff
@@ -598,6 +598,17 @@ func (f *forwarder) forwardWithDestChan(ctx context.Context, query packet, respo
...
 		metricDNSFwdDropBonjour.Add(1)
+		res, err := nxDomainResponse(query)
+		if err != nil {
+			f.logf("error parsing bonjour query: %v", err)
+			return nil
+		}
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case responseChan <- res:
```
crawshaw, line 609: Should we block here, or just drop the response?

crawshaw (approved):

net/dns/resolver/forwarder.go
bradfitz: I think it's fine... this is exactly what the packet writing code at the bottom does, and the caller's timeout is 5 seconds. So worst case it blocks for 5 seconds if userspace engine is having a bad day reading these out.

PR #3606, opened 2021-12-21
net/dns/resolver: add debug HTML handler to see what DNS traffic was forwarded

bradfitz:
for debugging iOS DNS mystery


crawshaw (approved):
You thinking we leave it in the prod release?

bradfitz:
> You thinking we leave it in the prod release?

Yeah, it seems harmless. Off by default, and restricted.

PR #3608, opened 2021-12-22
tailcfg: add Hostinfo.BasicallyEqual

bradfitz:


tailcfg/tailcfg.go
```diff
@@ -571,12 +571,31 @@ func (h *Hostinfo) Equal(h2 *Hostinfo) bool {
...
+		return false
+	}
+	a := *h
+	b := *h2
+	if !a.NetInfo.BasicallyEqual(b.NetInfo) {
+		return false
+	}
+	a.NetInfo = nil
+	b.NetInfo = nil
+	return a.Equal(&b)
```
crawshaw, line 596: I think we need to filter out Services. (And that means our future plans of adding potentially interesting services to Services should be rethought.)

There's also a lot of machines out there that bounce their Hostname around. Not sure what to do about that one. Definitely not startup critical, but also not really "basically equal"

bradfitz:
Closing until we know whether we need this.

PR #3612, opened 2021-12-23
tempfork: remove pprof fork

cristaloleg:
Looks like it's not used at all + it was fixed in upstream https://github.com/golang/go/commit/33511fb959f8f0edd5e831a4b41523daf9d84e87#diff-bf7f9c7e4e4ce995f8412397216bb3fec637d5af147b89ee38e2826f2b253345


DentonGentry:
To set expectations: it is likely to be January before a review of this PR is done.

cristaloleg:
Ok for me)

bradfitz:
Actually, we're still using this code from our iOS app (in a different repo).

And even though the html/template dep has been removed in Go 1.16, net/http/pprof still has init-time registration of default routes, which we still don't want on iOS for linker code reduction reasons.  (by registering those handlers on the global DefaultServeMux, the linker can't deadcode those funcs even if we never use that DefaultServeMux)

Not everybody is on iOS 15 with the 50 MB of NetworkExtension memory limit yet (some are still using iOS 14 and earlier with a 15 MB limit), and we're also not using Go 1.18 with its relocation improvements yet.

So I don't think we're ready to switch to using the standard library's `net/http/pprof` yet.

/cc @josharian

cristaloleg:
Ah, that's how things looks like. Haven't found that it's used on iOS. In this case feel free to close the PR.

Issue #3619, opened 2021-12-24
magicsock, derp, netcheck: behave better when UDP writes fail


bradfitz:
Several intertwined bugs here,

In 306b03d71ef3e4a352c71c2a64b66c6678eef1687a7e755d6f70cc8916721ed8, @JayWStapleton was trying to test out running a custom DERP server and forcing it by blocking all his non-DNS UDP:

```
2021-12-23 13:00:26.828410023 -0500 -0500: magicsock: last netcheck reported send error. Rebinding.
2021-12-23 13:00:26.828630129 -0500 -0500: magicsock: closing connection to derp-900 (rebind), age 27s
2021-12-23 13:00:26.82867139 -0500 -0500: magicsock: 0 active derp conns
2021-12-23 13:00:26.83593777 -0500 -0500: magicsock: adding connection to derp-900 for home-keep-alive
2021-12-23 13:00:26.836034517 -0500 -0500: magicsock: 1 active derp conns: derp-900=cr0s,wr0s
2021-12-23 13:00:26.836300914 -0500 -0500: portmapper: createOrGetMapping: write udp4 0.0.0.0:54082->10.0.2.2:5351: sendto: operation not permitted
2021-12-23 13:00:26.851156956 -0500 -0500: derphttp.Client.Recv: connecting to derp-900 (jay)
2021-12-23 13:00:26.85487967 -0500 -0500: health("overall"): error: not connected to home DERP region 900
2021-12-23 13:00:26.97326754 -0500 -0500: magicsock: derp-900 connected; connGen=1
2021-12-23 13:00:26.973431216 -0500 -0500: health("overall"): ok
2021-12-23 13:00:28.829194344 -0500 -0500: netcheck: report: udp=false v4=false v6=false mapvarydest= hair= portmap= derp=0
2021-12-23 13:00:52.066540179 -0500 -0500: magicsock: starting endpoint update (periodic)
2021-12-23 13:00:52.066590635 -0500 -0500: magicsock: last netcheck reported send error. Rebinding.
2021-12-23 13:00:52.066670633 -0500 -0500: magicsock: closing connection to derp-900 (rebind), age 25s
2021-12-23 13:00:52.066683199 -0500 -0500: magicsock: 0 active derp conns
2021-12-23 13:00:52.068200102 -0500 -0500: magicsock: adding connection to derp-900 for home-keep-alive
[...]

PR #3633, opened 2021-12-29
wgengine/magicsock: don't unconditionally close DERP connections on rebind

bradfitz:
Only if the source address isn't on the currently active interface or
a ping of the DERP server fails.

Updates #3619

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -42,6 +42,7 @@ import (
 	"tailscale.com/net/netns"
 	"tailscale.com/net/portmapper"
 	"tailscale.com/net/stun"
+	"tailscale.com/net/tsaddr"
 	"tailscale.com/syncs"
 	"tailscale.com/tailcfg"
 	"tailscale.com/tstime"
@@ -233,6 +234,7 @@ type Conn struct {
 	idleFunc               func() time.Duration // nil means unknown
 	testOnlyPacketListener nettype.PacketListener
 	noteRecvActivity       func(key.NodePublic) // or nil, see Options.NoteRecvActivity
+	linkMon                *monitor.Mon         // or nil
 
 	// ================================================================
 	// No locking required to access these fields, either because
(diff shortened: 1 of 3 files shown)


bradfitz:
TODO before I submit: fix DefaultRouteInterface on Windows to have its key line up with the interface name in interfaces.State and audit the other non-Linux OSes and remove the `State.DefaultRouteInterface` warning.

wgengine/magicsock/magicsock.go
```diff
@@ -2393,6 +2396,42 @@ func (c *Conn) closeAllDerpLocked(why string) {
...
+		dc := ad.c
+		go func() {
+			ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+			defer cancel()
+			if err := dc.Ping(ctx); err != nil {
+				c.mu.Lock()
+				defer c.mu.Unlock()
+				c.closeDerpLocked(regionID, "rebind-ping-fail")
+				if !c.privateKey.IsZero() {
+					c.startDerpHomeConnectLocked()
```
crawshaw, line 2425: I don’t understand why this is done only on ping failure

crawshaw: That is, don’t we need this on the local IP check above?

crawshaw (approved):

wgengine/magicsock/magicsock.go
bradfitz: The caller, Conn.Rebind, does it.

But I can pull it out into a new method like "closeOrReconnectDerpLocked" that makes it more clear.

crawshaw: Ah I see now. Maybe move that startDerpHomeConnectLocked from Rebind into this function?

crawshaw (approved):

bradfitz:
> Ah I see now. Maybe move that startDerpHomeConnectLocked from Rebind into this function?

Done. PTAL.

crawshaw (approved):

crawshaw:
Much clearer, thanks.

PR #3637, opened 2021-12-29
net/interfaces: define DefaultRouteInterface and State.DefaultRouteInterface

bradfitz:
It was pretty ill-defined before and mostly for logging. But I wanted
to start depending on it, so define what it is and make Windows match
the other operating systems, without losing the log output we had
before. (and add tests for that)

### DIFF
--- a/net/interfaces/interfaces.go
+++ b/net/interfaces/interfaces.go
@@ -172,6 +172,7 @@ func sortIPs(s []netaddr.IP) {
 type Interface struct {
 	*net.Interface
 	AltAddrs []net.Addr // if non-nil, returned by Addrs
+	Desc     string     // extra description (used on Windows)
 }
 
 func (i Interface) IsLoopback() bool { return isLoopback(i.Interface) }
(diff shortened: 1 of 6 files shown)


net/interfaces/interfaces.go
```diff
@@ -278,13 +279,16 @@ type State struct {
 	// instead of Wifi. This field is not populated by GetState.
 	IsExpensive bool
 
-	// DefaultRouteInterface is the interface name for the machine's default route.
+	// DefaultRouteInterface is the interface name for the
```
crawshaw, line 282: Is this used to bind values anywhere? I’m curious if we’re going from not binding to an interface to binding, which will change our behavior under all sorts of untested edge cases.

bradfitz: Not on Windows yet, no. See `netns_windows.go` which does it by hand still. A future change can make it use this instead.

crawshaw: so if I read this correctly, this PR is a no-op for edge cases. in that case, lgtm

crawshaw (approved):

PR #3641, opened 2021-12-30
netcheck,portmapper,magicsock: ignore some UDP write errors on Linux

bradfitz:
Treat UDP send EPERM errors as a lost UDP packet, not something super
fatal. That's just the Linux firewall preventing it from going out.

And add a leaf package net/neterror for that (and future) policy that
all three packages can share, with tests.

Updates #3619


wgengine/magicsock/magicsock.go
```diff
@@ -1206,13 +1207,15 @@ func (c *Conn) sendUDP(ipp netaddr.IPPort, b []byte) (sent bool, err error) {
...
+var errEPERM error = syscall.EPERM
+
 // sendUDP sends UDP packet b to addr.
 // See sendAddr's docs on the return value meanings.
 func (c *Conn) sendUDPStd(addr *net.UDPAddr, b []byte) (sent bool, err error) {
 	switch {
 	case addr.IP.To4() != nil:
 		_, err = c.pconn4.WriteTo(b, addr)
-		if err != nil && c.noV4.Get() {
+		if err != nil && (c.noV4.Get() || errors.Is(err, errEPERM)) {
```
crawshaw, line 1218: Something somewhere deserves a comment about why we ignore permanent errors

crawshaw: There’s also some net package abstract layer for this right? The comment should probably mention why we don’t use it

crawshaw (approved):

wgengine/magicsock/magicsock.go
bradfitz: PTAL

crawshaw (approved):
Fancy!

PR #3653, opened 2022-01-03
tsnet: add Start method to allow connecting to the tailnet without requiring a call to Dial/Listen

maisem:

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -67,8 +67,9 @@ type Server struct {
 }
 
 // Dial connects to the address on the tailnet.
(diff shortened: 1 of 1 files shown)


tsnet/tsnet.go
```diff
@@ -80,7 +81,8 @@ func (s *Server) doInit() {
 	}
 }
 
-func (s *Server) init() error {
+// Start connects the server to the tailnet.
```
bradfitz, line 84: Mention that it's optional and done implicitly by Dial or Listen.

bradfitz (approved):

PR #3661, opened 2022-01-04
cmd/hello: migrate to hello.ts.net as the hostname

bradfitz:
But still support hello.ipn.dev for a bit.

Updates tailscale/corp#1327

### DIFF
--- a/cmd/hello/hello.go
+++ b/cmd/hello/hello.go
@@ -2,13 +2,15 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// The hello binary runs hello.ipn.dev.
+// The hello binary runs hello.ts.net.
 package main // import "tailscale.com/cmd/hello"
 
 import (
 	"context"
+	"crypto/tls"
 	_ "embed"
 	"encoding/json"
+	"errors"
 	"flag"
 	"html/template"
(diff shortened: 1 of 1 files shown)


danderson (approved):

cmd/hello/hello.go
```diff
@@ -69,11 +72,31 @@ func main() {
...
-			)
+			hs := &http.Server{
+				Addr: *httpsAddr,
+				TLSConfig: &tls.Config{
+					GetCertificate: func(hi *tls.ClientHelloInfo) (*tls.Certificate, error) {
+						switch hi.ServerName {
+						case "hello.ts.net":
+							return tailscale.GetCertificate(hi)
+						case "hello.ipn.dev":
+							c, err := tls.LoadX509KeyPair(
```
maisem, line 83: cache this instead of loading everytime?

maisem (approved):

cmd/hello/hello.go
bradfitz: I considered it but it's going away soon anyway and is almost never hit. Not sure it matters complicating the control flow.

bradfitz: Also as written this means we can replace the hello.ipn.dev cert on disk without restarting the server :)

PR #3663, opened 2022-01-04
ipnlocal, net/{dns,tsaddr,tstun}, wgengine: support MagicDNS on IPv6

### LINKED ISSUE #3660: MagicDNS does not work in IPv6 only mode
### What is the issue?

When IPv4 is disabled, there is no mechanism for a host to contact 100.100.100.100 for MagicDNS resolution. 

### Steps to reproduce

Set  `"disableipv4": true,` in ACL
tailscale device automatically sets nameservers to auto-discover ({fec0:0:0:ffff::1, fec0:0:0:ffff::2, fec0:0:0:ffff::3})
tailscale sets a route to 100.100.100.100 for MagicDNS, but that fails because there's no IPv4 address on that interface
Cannot access tailnet computers by machine name. 
tailscale will still resolve machine names - `tailscale ping mymachine` works but `ping mymachine` does not

_No response_

### OS

_No response_

### OS version

_No response_

### Tailscale version

_No response_

### Bug report

_No response_

bradfitz:
Fixes #3660

RELNOTE=MagicDNS now works over IPv6 when CGNAT IPv4 is disabled.

### DIFF
--- a/wgengine/userspace.go
+++ b/wgengine/userspace.go
@@ -56,7 +56,10 @@ import (
 
 const magicDNSPort = 53
 
-var magicDNSIP = netaddr.IPv4(100, 100, 100, 100)
+var (
+	magicDNSIP   = tsaddr.TailscaleServiceIP()
+	magicDNSIPv6 = tsaddr.TailscaleServiceIPv6()
+)
 
 // Lazy wireguard-go configuration parameters.
 const (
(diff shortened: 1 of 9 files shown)


bradfitz:
This uses @Xe's proposed `fd7a:115c:a1e0::53` as the Service IP.

cc @crawshaw @danderson

wgengine/userspace.go
```diff
@@ -508,22 +514,38 @@ func (e *userspaceEngine) pollResolver() {
...
+			h := packet.UDP6Header{
+				IP6Header: packet.IP6Header{
+					Src: magicDNSIPv6,
+					Dst: to.IP(),
+				},
+				SrcPort: magicDNSPort,
+				DstPort: to.Port(),
+			}
+			hlen := h.Len()
+			// TODO(dmytro): avoid this allocation without importing tstun quirks into dns.
```
crawshaw, line 544: 🤣

crawshaw (approved):

wgengine/userspace.go
bradfitz: 🤷‍♂️🤪 he might come back!

PR #3664, opened 2022-01-04
wgengine/magicsock: reconnect to DERP when its definition changes

bradfitz:


bradfitz:
(is there a tracking bug for this? I only saw that one email.)

DentonGentry (approved):

DentonGentry:
I don't think a bug was ever opened, I'd forgotten the symptom altogether until that email.

PR #3666, opened 2022-01-05
wgengine/magicsock: reduce disco ping heartbeat aggressiveness a bit

bradfitz:
Bigger changes coming later, but this should improve things a bit in
the meantime.

Updates #3363

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -3274,7 +3274,7 @@ const (
 	// try to keep an established endpoint peering alive.
 	// It's also the idle time at which we stop doing STUN queries to
 	// keep NAT mappings alive.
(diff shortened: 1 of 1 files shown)


crawshaw (commented):
When I see magic numbers change I wonder why. The commit message and comments doesn't explain why the change is being made, and it doesn't explain why it doesn't break any existing uses.

bradfitz:
Updated comment message with background.

crawshaw:
Leaving for @danderson. Looks like a dangling sentence in the commit message:

`* heartbeat from 2 seconds to 3 seconds: reducing rate from every`

bradfitz:
Updated commit message.

danderson (approved):

Issue #3669, opened 2022-01-06
android: Makefile doesn't work on macOS


bradfitz:
Normally I do Android work on a headless Linux machine but I wanted to have lower iteration time on UI work in the emulator so I decided to try doing development work on my Mac:

```
crow5k:tailscale-android $ make tailscale-debug.apk
go: downloading tailscale.com v1.1.1-0.20220103195654-dd45bba76b3a
/bin/sh: /Users/bradfitz/.cache/tailscale-android-go-118f76c18c3f755ba0e6211e170a6422871f2791/go/bin/go: No such file or directory
/bin/sh: /Users/bradfitz/.cache/tailscale-android-go-118f76c18c3f755ba0e6211e170a6422871f2791/go/bin/go: No such file or directory
mktemp: illegal option -- -
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix 
want: okay
got: bad
rm -rf /Users/bradfitz/.cache/tailscale-android-go-*
wget https://github.com/tailscale/go/releases/download/build-118f76c18c3f755ba0e6211e170a6422871f2791/linux.tar.gz -O ""
/bin/sh: wget: command not found
make: *** [toolchain] Error 127
```

bradfitz:
First attempt was:

```patch
diff --git a/Makefile b/Makefile
index 04de26d..da81e5e 100644
--- a/Makefile
+++ b/Makefile
@@ -42,11 +42,9 @@ ifneq ($(TOOLCHAINWANT),$(TOOLCHAINSUM))
        @echo want: $(TOOLCHAINWANT)
        @echo got: $(TOOLCHAINSUM)
        rm -rf ${HOME}/.cache/tailscale-android-go-*
-       $(eval tmpfile=$(shell mktemp --suffix=.tgz))
-       wget https://github.com/tailscale/go/releases/download/build-$(TOOLCHAINREV)/linux.tar.gz -O "$(tmpfile)"
        mkdir -p $(TOOLCHAINDIR)
-       tar xzf $(tmpfile) -C $(TOOLCHAINDIR)
-       rm $(tmpfile)
+       curl --silent -L https://github.com/tailscale/go/releases/download/build-$(TOOLCHAINREV)/linux.tar.gz | \
+               tar -C $(TOOLCHAINDIR) -zxv
 endif
 
 $(DEBUG_APK): toolchain
```

But then I realized it only ever downloads a linux-amd64 toolchain.

PR #3670, opened 2022-01-06
cmd/printdep: add flag to print out Go toolchain tarball URL

bradfitz:
Updates #3669

### DIFF
--- a/cmd/printdep/printdep.go
+++ b/cmd/printdep/printdep.go
@@ -9,18 +9,38 @@ package main
 import (
 	"flag"
 	"fmt"
+	"log"
+	"runtime"
 	"strings"
 
 	ts "tailscale.com"
 )
 
 var (
-	goToolchain = flag.Bool("go", false, "print the supported Go toolchain git hash (a github.com/tailscale/go commit)")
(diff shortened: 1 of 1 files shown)


cmd/printdep/printdep.go
```diff
@@ -9,18 +9,38 @@ package main
...
 	flag.Parse()
 	if *goToolchain {
 		fmt.Println(strings.TrimSpace(ts.GoToolchainRev))
 	}
+	if *goToolchainURL {
+		var suffix string
+		switch runtime.GOARCH {
+		case "amd64":
+			// None
+		case "arm64":
```
DentonGentry, line 34: We build static binaries for a number of other architectures like MIPS, will we need any handling here?

DentonGentry (approved):

cmd/printdep/printdep.go
bradfitz: We don't; only four: https://github.com/tailscale/go/releases

bradfitz: (That is, these are the four platforms we support for development)

PR #3677, opened 2022-01-07
wgengine/netstack: add env knob to turn on netstack debug logs

bradfitz:
Except for the super verbose packet-level dumps. Keep those disabled
by default with a const.

Updates #2642

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"log"
 	"net"
+	"os"
 	"os/exec"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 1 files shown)


wgengine/netstack/netstack.go
```diff
@@ -560,6 +563,7 @@ func (ns *Impl) forwardTCP(client *gonet.TCPConn, clientRemoteIP netaddr.IP, wq
 	go func() {
 		select {
 		case <-notifyCh:
+			ns.logf("[v2] netstack: forwardTCP notifyCh fired; canceling context for %s", dialAddrStr)
```
maisem, line 566: Should this be in if block with debugNetstack?

maisem (approved):

wgengine/netstack/netstack.go
bradfitz: Yeah, guess so. Done.

bradfitz:
> I keep meaning to wire in a pcap dump.

What does that give you over running tcpdump?

Issue #2815, opened 2021-09-08
Tailscale breaks WSL2 DNS


PieterScheffers:
On Windows login, Tailscale is started. When I open Ubuntu 18.04 on WSL2 no DNS is working.
```
$ dig google.com
; <<>> DiG 9.11.3-1ubuntu1.12-Ubuntu <<>> google.com
;; global options: +cmd
;; connection timed out; no servers could be reached
```

When opening `/etc/wsl.conf`, I see Tailscale has added the following snippet:
```
# added by tailscale
[network]
generateResolvConf = false
```

When I remove this snippet from `/etc/wsl.conf` and then terminate and start WSL again, the DNS works again.
```
C:\WINDOWS\system32> wsl --terminate Ubuntu-18.04
C:\WINDOWS\system32> wsl --distribution Ubuntu-18.04

$ dig google.com
; <<>> DiG 9.11.3-1ubuntu1.12-Ubuntu <<>> google.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 54109
;; flags: qr rd ad; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0
;; WARNING: recursion requested but not available

;; QUESTION SECTION:
;google.com.                    IN      A

;; ANSWER SECTION:
google.com.             0       IN      A       142.250.179.206

;; Query time: 21 msec
;; SERVER: 172.29.64.1#53(172.29.64.1)
;; WHEN: Wed Sep 08 09:16:49 CEST 2021
;; MSG SIZE  rcvd: 54
```

On every restart of Windows the snippet is added to `/etc/wsl.conf` again.

DentonGentry:
The write to /etc/wsl.conf was added in https://github.com/tailscale/tailscale/commit/1147c7fd4f1cdb2eb9727cb34a5af0bf09ca84f7 to fix https://github.com/tailscale/tailscale/issues/775.

When first booted with "generateResolvConf = false" present, what is the content of /etc/resolv.conf? I think we'd expect to have written a "nameserver 100.100.100.100" line there.

PieterScheffers:
The `/etc/resolv.conf` file doesn't exist when the DNS isn't working.

DentonGentry:
In https://login.tailscale.com/admin/dns do you have either Magic DNS enabled, or any nameservers added?

PieterScheffers:
[...]

PR #3687, opened 2022-01-10
net/dns: make WSL network configuration opt-in for now

bradfitz:
Tailscale seems to be breaking WSL configurations lately.  Until we
understand what changed, turn off Tailscale's involvement by default
and make it opt-in.

Updates #2815

### DIFF
--- a/net/dns/manager_windows.go
+++ b/net/dns/manager_windows.go
@@ -7,8 +7,10 @@ package dns
 import (
 	"errors"
 	"fmt"
+	"os"
 	"os/exec"
 	"sort"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


net/dns/manager_windows.go
```diff
@@ -34,6 +36,8 @@ const (
 	versionKey = `SOFTWARE\Microsoft\Windows NT\CurrentVersion`
 )
 
+var configureWSL, _ = strconv.ParseBool(os.Getenv("TS_DEBUG_CONFIGURE_WSL"))
```
DentonGentry, line 39: We'd advise people to set a System Environment variable? It looks like those would apply to the Windows service tailscaled.

crawshaw (approved):

crawshaw:
https://github.com/tailscale/tailscale/issues/3690

bradfitz:
Per discussion with @DentonGentry, we're just going to disable it for 1.20.

PR #3692, opened 2022-01-10
ipn/ipnserver: provide means of setting debug env vars on Windows

### LINKED ISSUE #3688: Provide easy means of setting environment variables on Windows
Tailscale has a number of environment variable based debug knobs.

On Linux, we can debug with users by having them set `/etc/default/tailscaled`.

We should provide something similar on Windows, which tailscaled can read before it runs itself as a child process in the ipnserver.BabysitProc.

That'd help with e.g. https://github.com/tailscale/tailscale/pull/3687 so people can turn that back on.

/cc @crawshaw @dblohm7

bradfitz:
Fixes #3688

### DIFF
--- a/ipn/ipnserver/server.go
+++ b/ipn/ipnserver/server.go
@@ -20,6 +20,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"os/user"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


ipn/ipnserver/server.go
```diff
@@ -1184,3 +1193,47 @@ func findTrueNASTaildropDir(name string) (dir string, err error) {
 	}
 	return "", fmt.Errorf("shared folder %q not found", name)
 }
+
+func loadExtraEnv() (env []string, err error) {
+	if runtime.GOOS != "windows" {
+		return nil, nil
+	}
+	name := filepath.Join(os.Getenv("ProgramData"), "Tailscale", "tailscaled-env.txt")
```
crawshaw, line 1201: Rather than an absolute path can we find this relative to the executable, for debug setups that run from a directory rather than using the installer?

crawshaw (commented):
The fact that we have this and registry keys on windows is a bit confusing. What's the logic for what goes where?

bradfitz:
> The fact that we have this and registry keys on windows is a bit confusing. What's the logic for what goes where?

Supported things can go in the registry.

Debug crap that's debug-portable between Linux/Windows/etc can continue to go in the environment. (We have dozens already... this just makes it usable on Windows when running as a service.... you can already set them by hand on Windows but only for tailscaled in the foreground in an admin cmd.exe)

ipn/ipnserver/server.go
bradfitz: Not worth it. If you're using a debug setup and running a custom binary from cmd.exe, you can set the environment variable by hand already in cmd.exe or powershell. You don't need this mechanism for that.

This is purely for running as a service.

crawshaw (approved):

Issue #3584, opened 2021-12-16
Tailscale Windows registry (policy) settings should be under HKLM\Software\Polices\


AdrianFletcher:
### What are you trying to do?

I'm trying to manage the Windows policies of Tailscale in an automated and predictable fashion.

### How should we solve this?

It's generally considered best practice to locate software policies under the HKLM\Software\Policies registry folder. Tailscale currently stores these under [HKLM\Software\Tailscale IPN](https://tailscale.com/kb/1102/windows-config-and-log-files/).

This allows the software policies to play nicely with Group Policy and Endpoint/InTune should your team wish to push out any ADMX files or a sysadmin create their own.

### What is the impact of not solving this?

I'm pushing out configuration changes via a custom installation script (reg add, and reg delete on removal). This does not (easily) allow differentiation of settings depending on the machine/user however.

DentonGentry:
Presumably we would check HKLM\Software\Policies first, then look in HKLM\Software\Tailscale IPN?
Or is there a way to alias a registry key in two locations, for migrations like this?

dblohm7:
Unfortunate the public API for the registry does not allow for any kind of alias/symlink type of thing.

Probably the best way to deal with this is to have policy retrieval code that always accesses the "right" location and migrates if necessary.

AdrianFletcher:
If memory serves, the Policy key cannot be written to except by an elevated process (e.g. the System account), so you might need to read from both, or perhaps log an error/deprecated warning on reading the keys from HKLM\Software\Tailscale IPN?

dblohm7:
We can do it from `tailscaled`

dblohm7:
Policies will be read from `HKLM\Software\Policies` starting with the next unstable build.

boonedoggle:
Are there any plans to provide ADMX for Intune configuration?

PR #3693, opened 2022-01-10
cmd/tailscaled, util/winutil: add accessor functions for Windows syst…

dblohm7:
…em policies

This patch adds new functions to be used when accessing system policies,
and revises callers to use the new functions. They first attempt the new
registry path for policies, and if that fails, attempt to fall back to the
legacy path.

The remaining changes will be done in corp.

Updates https://github.com/tailscale/tailscale/issues/3584

### DIFF
--- a/util/winutil/winutil.go
+++ b/util/winutil/winutil.go
@@ -9,7 +9,31 @@ package winutil
 // are stored. This constant is a non-empty string only when GOOS=windows.
 const RegBase = regBase
 
-// GetRegString looks up a registry path in our local machine path, or returns
+// GetPolicyString looks up a registry value in the local machine's path for
+// system policies, or returns the given default if it can't.
+// Use this function to read values that may be set by sysadmins via the MSI
+// installer or via GPO. For registry settings that you do *not* want to be
+// visible to sysadmin tools, use GetRegString instead.
+//
+// This function will only work on GOOS=windows. Trying to run it on any other
+// OS will always return the default value.
+func GetPolicyString(name, defval string) string {
+	return getPolicyString(name, defval)
+}
+
+// GetPolicyInteger looks up a registry value in the local machine's path for
+// system policies, or returns the given default if it can't.
+// Use this function to read values that may be set by sysadmins via the MSI
+// installer or via GPO. For registry settings that you do *not* want to be
+// visible to sysadmin tools, use GetRegInteger instead.
+//
+// This function will only work on GOOS=windows. Trying to run it on any other
+// OS will always return the default value.
+func GetPolicyInteger(name string, defval uint64) uint64 {
+	return getPolicyInteger(name, defval)
+}
+
+// GetRegString looks up a registry path in the local machine path, or returns
 // the given default if it can't.
 //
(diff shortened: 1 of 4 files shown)


util/winutil/winutil.go
```diff
@@ -30,27 +33,45 @@ func GetDesktopPID() (pid uint32, ok bool) {
...
 // OS will always return the default value.
-func GetRegString(name, defval string) string {
-	key, err := registry.OpenKey(registry.LOCAL_MACHINE, RegBase, registry.READ)
+func GetPolicyString(name, defval string) string {
+	s, err := getRegString(regPolicyBase, name)
 	if err != nil {
-		log.Printf("registry.OpenKey(%v): %v", RegBase, err)
-		return defval
+		// Fall back to the legacy path for policies
+		return GetRegString(name, defval)
```
bradfitz, line 45: pretty confusing having two funcs called here in GetPolicyString both called "get reg string" but with different case and different signatures.

It took me a few reads to figure out what was going on.

Should we rename some things and get "OrDefault" in the name of one?

util/winutil/winutil_notwindows.go
```diff
@@ -9,6 +9,20 @@ package winutil
 
 const RegBase = ``
 
+// GetPolicyString looks up a registry value in our local machine's path for
+// system policies, or returns the given default if it can't.
+//
+// This function will only work on GOOS=windows. Trying to run it on any other
```
bradfitz, line 15: rather than repeating docs over multiple files with multiple builds tags, I often prefer to put the exported func + docs in one portable file and then call an unexported one that differs per build tag file, which then doesn't need docs.

util/winutil/winutil.go
```diff
@@ -16,7 +16,10 @@ import (
 	"golang.org/x/sys/windows/registry"
 )
 
-const RegBase = `SOFTWARE\Tailscale IPN`
+const (
+	RegBase       = `SOFTWARE\Tailscale IPN`
```
sailorfrag, line 20: I feel like these should either both be exported or neither should be exported.

util/winutil/winutil.go
sailorfrag: I think `GetRegString` and `GetRegInteger` can just be removed entirely and we can migrate all callers to the Policy versions. Then the existing behavior can simply be included in the Policy functions.

util/winutil/winutil.go
bradfitz: but if they stay mixed, they shouldn't be in a const ( ... ) block together

bradfitz (changes requested):
Marking "request changes" to see if that makes GitHub bot stop notifying.

util/winutil/winutil_notwindows.go
dblohm7: Fixing

util/winutil/winutil.go
dblohm7: Yeah, I'll do something about the naming, however I do not want to get rid of the non-policy variants, as is is not always desirable to expose all registry settings to admin policy.

bradfitz (approved):

PR #3709, opened 2022-01-11
cmd/tailscale: let 'tailscale up --reset' do a pref edit

### LINKED ISSUE #3702: tailscale up --reset forces reauthentication
### What is the issue?

running tailscale up --reset forces reauth on Windows. 

### Steps to reproduce

`tailscale up --reset`
Any other arguments work fine

tested on 1.18.0, 1.18.2, 1.19.187 on Windows 10, 11

Troubleshooting: 
Tried clean reinstall of tailscale 1.18.2, removing the state files, and the problem persists

_No response_

### OS

Windows

### OS version

Windows 10, 11

### Tailscale version

1.18.0, 1.18.2, 1.19.187

### Bug report

_No response_

bradfitz:
The --reset shouldn't imply that a Start is necessary.

Fixes #3702

### DIFF
--- a/cmd/tailscale/cli/cli_test.go
+++ b/cmd/tailscale/cli/cli_test.go
@@ -787,6 +787,16 @@ func TestUpdatePrefs(t *testing.T) {
 			wantJustEditMP: &ipn.MaskedPrefs{WantRunningSet: true},
 		},
 		{
+			name:  "just_edit_reset",
+			flags: []string{"--reset"},
+			curPrefs: &ipn.Prefs{
+				ControlURL: ipn.DefaultControlURL,
+				Persist:    &persist.Persist{LoginName: "crawshaw.github"},
+			},
+			env:            upCheckEnv{backendState: "Running"},
+			wantJustEditMP: &ipn.MaskedPrefs{WantRunningSet: true},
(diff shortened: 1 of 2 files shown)


maisem:
I don't understand why this is only an issue on windows though?

DentonGentry (approved):

maisem:
Can you also clarify the commit message? What do you mean by "Start"?

bradfitz:
Start is one of the backend.Client methods, as opposed to EditPrefs. It's in the code just following this.

This was also affecting Linux, but the behavior was different because tailscaled defaults to backend-owned prefs on Linux but GUI-owned client prefs on Windows.

bradfitz:
> Can you also clarify the commit message? What do you mean by "Start"?

Clarified.

maisem:
> Start is one of the backend.Client methods, as opposed to EditPrefs. It's in the code just following this.
> 
> 
> 
> This was also affecting Linux, but the behavior was different because tailscaled defaults to backend-owned prefs on Linux but GUI-owned client prefs on Windows.

Ah that makes sense, thanks!

maisem (approved):

PR #3716, opened 2022-01-12
cmd/tailscale/cli/web: fix typo where the html template data was being replaced instead of being appended to.

maisem:
This caused the exit node toggle to stop working as ",10.0.6.1/32" is invalid.

### DIFF
--- a/cmd/tailscale/cli/web.go
+++ b/cmd/tailscale/cli/web.go
@@ -375,7 +375,7 @@ func webHandler(w http.ResponseWriter, r *http.Request) {
 			data.AdvertiseExitNode = true
 		} else {
 			if data.AdvertiseRoutes != "" {
-				data.AdvertiseRoutes = ","
+				data.AdvertiseRoutes += ","
 			}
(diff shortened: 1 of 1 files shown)


danderson (approved):

cmd/tailscale/cli/web.go
```diff
@@ -375,7 +375,7 @@ func webHandler(w http.ResponseWriter, r *http.Request) {
 			data.AdvertiseExitNode = true
 		} else {
 			if data.AdvertiseRoutes != "" {
-				data.AdvertiseRoutes = ","
+				data.AdvertiseRoutes += ","
```
bradfitz, line 378: actually! we have a better API for this now.

See 2cfc96aa901433ae407883ce8856244b3a196a49

bradfitz: oh, but this is some weird template thing

bradfitz (approved):

crawshaw (approved):

Issue #3710, opened 2022-01-11
Using Tailscale 1.19.187 on an aarch64 Synology in Userspace Mode Doesn't Work as an Exit Node


Xe:
### What is the issue?

I tried to use my synology as an exit node. It failed to let me search things on google. Changing to another exit node in my network let me confirm the outgoing IP address as well as search things on google.

### Steps to reproduce

1. Enable exit node on your synology
2. enable it as an exit node in the web panel
3. try to use it
4. internet traffic doesn't work

### OS

Synology

### OS version

Synology DSM 7

### Tailscale version

1.19.187

### Bug report

BUG-302bab5de9dd81635703607289caa4bebdcbce06ed44706f8f08886d0bbf24e2-20220111213157Z-2041ae85c824ff3f

bradfitz:
I think this is sufficiently fixed now.

Issue #2922, opened 2021-09-24
Synology DS416 DSM6 armv7 package (tailscale-armv7-1.14.3-008-dsm6) stops working after 1 minute 


Markmaster:
Hi.
I running Synology NAS DS416 DSM 6.2.4-25556, and I updated the previous version to the current release tailscale-armv7-1.14.3-008-dsm6.spk.
After one minute or above pings are lost and the package is stopped.
Any next manual run will lead to only a stopped package.

bradfitz:
What's your Tailscale IP so we can look up your logs? (Your Tailscale IP is safe to share publicly; it's useless to anybody but you)

bradfitz:
And do you see any crashes in your Synology logs, at `/var/packages/Tailscale/var/tailscaled.stdout.log` ?

DentonGentry:
Almost identical report earlier today: https://github.com/tailscale/tailscale/issues/2920

maisem:
@Markmaster Just published https://github.com/tailscale/tailscale-synology/releases/tag/v1.14.4. Can you please try that?

Markmaster:
> @Markmaster Just published https://github.com/tailscale/tailscale-synology/releases/tag/v1.14.4. Can you please try that?

Yes, I'll try this now.

The strange thing is I'm not signed for replies on this issues tracker.

Markmaster:
> And do you see any crashes in your Synology logs, at `/var/packages/Tailscale/var/tailscaled.stdout.log` ?

Markmaster:
> And do you see any crashes in your Synology logs, at `/var/packages/Tailscale/var/tailscaled.stdout.log` ?

Can I upload this file or send it to some email for troubleshooting?
The IP is 100.81.117.88

bradfitz:
@Markmaster, why did you close this bug? I assume that was a mistake?

A couple suspicious things in logs:
```
2021-10-04 12:56:49.795691123 +0300 +0300: Program starting: v1.14.6-t422a83b3d-g721194f59, Go 1.17-tsdd6b269c77: []string{"/volume1/@appstore/Tailscale/bin/tailscaled", "--state=/var/packages/Tailscale/etc/tailscaled.state", "-
-socket=/var/packages/Tailscale/etc/tailscaled.sock", "--port=41641"}
[...]

PR #3719, opened 2022-01-12
wgengine/netstack: make userspace ping work when tailscaled has CAP_NET_RAW

bradfitz:
Updates #3710

### DIFF
--- a/wgengine/netstack/netstack_linux.go
+++ b/wgengine/netstack/netstack_linux.go
@@ -1,0 +1,20 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package netstack
+
+import (
+	"os/exec"
+	"syscall"
+
+	"golang.org/x/sys/unix"
+)
+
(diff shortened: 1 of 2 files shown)


wgengine/netstack/netstack_linux.go
```diff
@@ -0,0 +1,20 @@
...
+
+package netstack
+
+import (
+	"os/exec"
+	"syscall"
+
+	"golang.org/x/sys/unix"
+)
+
```
maisem, line 13: there is some weird behavior on DSM6 which led to having to special case DSM7
https://github.com/tailscale/tailscale/blob/main/wgengine/router/router_linux.go#L180-L189

maisem: https://github.com/tailscale/tailscale/issues/2922#issuecomment-946026420

bradfitz: Done; PTAL

bradfitz: Confirmed that my tailscaled runs as uid 0 on DSM6 and non-0 on DSM7

maisem (approved):

Issue #2698, opened 2021-08-23
tailscale stops working after wake up from sleep on Windows 10


vanc:
I'm running the latest stable version 1.12.3. Started experiencing the issue a couple of days ago.

So basically, after the PC wakes up from sleep, traffic couldn't go through the tunnel. From other peers, the PC wasn't connected.

If I restart the app, the menu bar will have a message "Please restart the Tailscale Windows Service". But if check the tailscale Windows service via the "Service" app, it's actually running. If check the process list from task manager, the process tailscald is right there.

![image](https://user-images.githubusercontent.com/588742/130476929-5c6cf755-2b0b-40de-bb8c-818503aab440.png)

![image](https://user-images.githubusercontent.com/588742/130477079-0fb2eec6-84ff-48cc-bfe2-d555adfc9f69.png)

My Windows version is 21H1 (OS Build 19043.1165). On Aug 15, [KB5005033 ](https://support.microsoft.com/en-us/topic/august-10-2021-kb5005033-os-builds-19041-1165-19042-1165-and-19043-1165-b4c77d08-435a-4833-b9f7-e092372079a4) was installed. I suspect the issue was related this update.

I also tried install/reinstall the tailscale app when issue happened, with the same issue. The only solution is to reboot the PC. But that's apparently very inconvenient as I need to restore all kinds of VMs and my dev environments.

Please let me know if any other information is needed.

bradfitz:
What's your Tailscale IP?

vanc:
> What's your Tailscale IP?

For the problemic Windows instance, it was 100.88.64.72.

bradfitz:
@vanc, your logs were never uploaded with that bug report.

Can you email support at tailscale.com with your logs from C:\Windows\System32\config\systemprofile\AppData\Local\Tailscale\Logs?  (and reference this github bug in your email)

DentonGentry:
No recent reports and cannot reproduce.
return -ETIMEDOUT;

PR #3728, opened 2022-01-13
cmd/tailscale: use html/template for synoTokenRedirect

josharian:
The GitHub code scanner flagged this as a security vulnerability.
I don't believe it was, but I couldn't convince myself of it 100%.
Err on the safe side and use html/template to generate the HTML,
with all necessary escaping.


josharian:
I manually tested this one function locally, in isolation, and got plausible output out of it. I did not attempt to test it in an end-to-end manner.

bradfitz (approved):

maisem:
I tried fixing this earlier and completely forgot about it. Thanks! Can you add "Fixes #2698"

maisem (approved):

Issue #150, opened 2020-03-06
derp: DERP Return Path Optimization (DRPO)


bradfitz:
Currently DERP does asymmetric routing. If A and B have different home nodes (say, SF and NY), then A sends to B via NY and B sends to A via SF.

But there's an optimization we could do:

One of the packets has to arrive first. When A (connected to its home node SF), gets a packet from B via DERP SF, it can add a "DERP routing table entry" for that public key to go via that specific DERP TCP connection, keeping that routing entry alive until either: a) the TCP connection dies (for whatever reason), or b) the server sends a "previous sender disconnected" frame.

So then the server would just need to keep track of all the public keys that have sent a node a message, and upon a TCP connection, the DERP SF server can tell A that B is disconnected, and A can remove its routing table entry.

So now, when A wants to send a reply packet (or any packet) to B, it first looks to see if it has an explicit DERP route for B, otherwise it falls back to using B's DERP home node. That also means less latency on the first reply, since it doesn't need to do the TCP+TLS+DERP round trips to set up the connection. And it minimizes overall connections & memory use.

Again, this is just an optimization (that should trigger a lot; we could add varz) and the fallback behavior would be the same as today.

/cc @crawshaw @danderson

apenwarr:
Note that this might not be optimal at all: in general (and to way overgeneralize, sigh), packets leaving a cloud provider we get charged for, while packets entering a cloud provider we don't. If someone wants to run their own DERP server inside their own network, we'd like to use that whenever possible for incoming packets - it (in theory) won't cost anything to retransmit them to machines on the same LAN. Whereas a DERP relay hosted elsewhere, even colocated with the transmitter, might pay extra fees.

bradfitz:
@apenwarr, when we do self-hosted DERP, this policy is easy to tweak or make dynamic. It's already behind a flag to go either way. When we update the derpmap and make it dynamic, we can make it a bool on the servers.

But for our own 5 current DERP servers, the goal here is to reduce the number of open TCP connections nodes need to maintain, and reduce initial handshake latency over DERP while NAT punching proceeds independently.

If A is connecting to B via derp1, it'd be nice to have B be able to WG handshake back over derp1 ~immediately, rather than opening a connection to derp2 first (so many RTTs: TCP+TLS+DERP).

So for our public DERPs I plan to enable this at some point. I want to be able to see global DERP connections graphs before & after, so I'll probably make it a flag sent by control so we can toggle it server-side and watch numbers.

[...]

PR #3730, opened 2022-01-13
wgengine/magicsock: enable DERP Return Path Optimization (DRPO)

bradfitz:
Turning this on at the beginning of the 1.21.x dev cycle, for 1.22.

Updates #150

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -68,7 +68,7 @@ func useDerpRoute() bool {
 	if v, ok := ob.Get(); ok {
 		return v
 	}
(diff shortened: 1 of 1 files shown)


DentonGentry (approved):

bradfitz:
> I assume we'll try turning it on by control flag during the 1.22 dev period

No, this turns it on for everybody. The control flag is now an emergency off switch, not an opt-in switch.

crawshaw:
Yeah that's fine, I meant that we can try out DRPO in 1.20 before we release 1.22.

bradfitz:
Yup!

PR #3732, opened 2022-01-13
logtail: cap the buffer size in encodeText

josharian:
This started as an attempt to placate GitHub's code scanner,
but it's also probably generally a good idea.

### DIFF
--- a/logtail/logtail.go
+++ b/logtail/logtail.go
@@ -431,6 +431,21 @@ func (l *Logger) encodeText(buf []byte, skipClientTime bool) []byte {
 	// For now just factor in a dozen.
 	overhead += 12
 
+	// Put a sanity cap on buf's size.
+	max := 16 << 10
+	if l.lowMem {
+		max = 255
+	}
+	var nTruncated int
+	if len(buf) > max {
+		nTruncated = len(buf) - max
+		// TODO: this can break a UTF-8 character
+		// mid-encoding.  We don't tend to log
+		// non-ASCII stuff ourselves, but e.g. client
+		// names might be.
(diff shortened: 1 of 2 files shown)


logtail/logtail.go
```diff
@@ -431,6 +431,21 @@ func (l *Logger) encodeText(buf []byte, skipClientTime bool) []byte {
 	// For now just factor in a dozen.
 	overhead += 12
 
+	// Put a sanity cap on buf's size.
+	max := 16384
```
bradfitz, line 435: 16 << 10 looks less mysterious

After 4096 I start forgetting my powers of two in decimal

logtail/logtail.go
```diff
@@ -469,14 +484,10 @@ func (l *Logger) encodeText(buf []byte, skipClientTime bool) []byte {
...
-		if l.lowMem && i > 254 {
-			// TODO: this can break a UTF-8 character
-			// mid-encoding.  We don't tend to log
-			// non-ASCII stuff ourselves, but e.g. client
-			// names might be.
-			b = append(b, "…"...)
-			break
-		}
+	}
+	if nTruncated > 0 {
```
bradfitz, line 488: testable?

logtail/logtail.go
josharian: I feel the opposite, but sure, done.

logtail/logtail.go
josharian: tested

bradfitz (approved):

crawshaw (approved):

PR #3734, opened 2022-01-13
net/dnscache: don't cancel the TLS context before writing to the result channel.

danderson:
Cancelling the context makes the timeout goroutine race with the write that
reports a successful TLS handshake, so you can end up with a successful TLS
handshake that mysteriously reports that it timed out after ~0s in flight.

The context is always canceled and cleaned up as the function exits, which
happens mere microseconds later, so just let function exit clean up and
thereby avoid races.

### DIFF
--- a/net/dnscache/dnscache.go
+++ b/net/dnscache/dnscache.go
@@ -444,22 +444,9 @@ func TLSDialer(fwd DialContextFunc, dnsCache *Resolver, tlsConfigBase *tls.Confi
 		}
 		tlsConn := tls.Client(tcpConn, cfg)
 
-		errc := make(chan error, 2)
 		handshakeCtx, handshakeTimeoutCancel := context.WithTimeout(ctx, 5*time.Second)
 		defer handshakeTimeoutCancel()
-		done := make(chan bool)
-		defer close(done)
-		go func() {
-			select {
-			case <-done:
-			case <-handshakeCtx.Done():
-				errc <- errTLSHandshakeTimeout
-			}
-		}()
-		go func() {
-			errc <- tlsConn.Handshake()
-		}()
-		if err := <-errc; err != nil {
+		if err := tlsConn.HandshakeContext(handshakeCtx); err != nil {
 			tcpConn.Close()
 			// TODO: if err != errTLSHandshakeTimeout,
 			// assume it might be some captive portal or
--- a/net/dnscache/dnscache.go
+++ b/net/dnscache/dnscache.go
@@ -457,9 +457,7 @@ func TLSDialer(fwd DialContextFunc, dnsCache *Resolver, tlsConfigBase *tls.Confi
 			}
 		}()
 		go func() {
-			err := tlsConn.Handshake()
-			handshakeTimeoutCancel()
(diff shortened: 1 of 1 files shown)


net/dnscache/dnscache.go
```diff
@@ -457,9 +457,7 @@ func TLSDialer(fwd DialContextFunc, dnsCache *Resolver, tlsConfigBase *tls.Confi
 			}
 		}()
 		go func() {
-			err := tlsConn.Handshake()
-			handshakeTimeoutCancel()
-			errc <- err
+			errc <- tlsConn.Handshake()
```
bradfitz, line 460: really all this crap can be simplified now that this is in std as of Go 1.17: https://pkg.go.dev/crypto/tls#Conn.HandshakeContext

net/dnscache/dnscache.go
```diff
@@ -457,9 +457,7 @@ func TLSDialer(fwd DialContextFunc, dnsCache *Resolver, tlsConfigBase *tls.Confi
 			}
 		}()
 		go func() {
-			err := tlsConn.Handshake()
-			handshakeTimeoutCancel()
```
bradfitz, line 461: ah! I see. it made the L452 goroutine readable and send to errc first sometimes.

bradfitz (approved):
LGTM. Up to you to switch to HandshakeContext in this PR or a future one.

DentonGentry:
Would we want this in 1.20.1?

If the functionality works and just leads to annoying-to-us logging, that isn't as important to get into 1.20.1.

net/dnscache/dnscache.go
danderson: Yup, exactly.

net/dnscache/dnscache.go
danderson: Replaced with HandshakeContext.

danderson:
It's probably not necessary for 1.20.1. In production code, the only effect is that we might have to retry ~once to get a working control or DERP connection, because we mistakenly judge a working connection to be bad. It makes tests flaky, but doesn't hugely affect the operation of tailscaled, since we retry everything with backoff anyway.

bradfitz (approved):

bradfitz:
> Would we want this in 1.20.1?
>
> If the functionality works and just leads to annoying-to-us logging, that isn't as important to get into 1.20.1.

I'd be fine (and slightly supportive) of Dave's first rev of this PR getting backported, but not the bigger change to use HandshakeContext. That I'd rather let bake a bit.

danderson:
Well, I just wiped that first rev to replace it with HandshakeContext :). I don't think anyone would notice this bug outside of tests, so I'm meh on backporting, but sounds like on balance everyone's neutral or in favor. I still have the old commit in my reflog somewhere, I'll arrange this as 2 commits. Stand by.

bradfitz:
LGTM as 2 commits, thanks!

danderson:
I've cherry-picked commit a9da6b73 to `release-branch/1.20`, so it'll be in the next patch release.

PR #3735, opened 2022-01-13
tstime/rate: deflake TestLongRunningQPS

### LINKED ISSUE #3733: tstime/rate: TestLongRunningQPS test flake
At just above a076aaecc6cba71f6869a26cabcf6644071877cc on the CI ...

```
--- FAIL: TestLongRunningQPS (5.00s)
    rate_test.go:208: numOK = 596, want 597 (ideal 600.327094)
FAIL
exit status 1
FAIL	tailscale.com/tstime/rate	5.042s
```

josharian:
This test set the bar too high.
Just a couple of missed timers was enough to fail.
Change the test to more of a sanity check.
While we're here, run it for just 1s instead of 5s.

Prior to this change, on a 13" M1 MPB, with

stress -p 512 ./rate.test -test.run=QPS

I saw 90%+ failures.

After this change, I'm at 30k runs with no failures yet.

Fixes #3733

### DIFF
--- a/tstime/rate/rate_test.go
+++ b/tstime/rate/rate_test.go
@@ -185,7 +185,7 @@ func TestLongRunningQPS(t *testing.T) {
 	// This will still offer ~500 requests per second,
 	// but won't consume outrageous amount of CPU.
 	start := time.Now()
-	end := start.Add(5 * time.Second)
+	end := start.Add(1 * time.Second)
 	ticker := time.NewTicker(2 * time.Millisecond)
 	defer ticker.Stop()
 	for now := range ticker.C {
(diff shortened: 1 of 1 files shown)


tstime/rate/rate_test.go
```diff
@@ -185,7 +185,7 @@ func TestLongRunningQPS(t *testing.T) {
 	// This will still offer ~500 requests per second,
 	// but won't consume outrageous amount of CPU.
 	start := time.Now()
-	end := start.Add(5 * time.Second)
+	end := start.Add(1 * time.Second)
```
crawshaw, line 188: Doesn't this increase the likelihood of a flake on a highly loaded machine?

crawshaw (approved):

bradfitz (approved):

tstime/rate/rate_test.go
bradfitz: How about trying it for 1 second up to 5 tries?

josharian: It is not obvious to me the impact. It might actually make it more reliable, if the failures occur when a single long pause happens--then there's a smaller window to get unlucky with the scheduler. I'm inclined to keep it shorter for now, and get fancy with retries if/when there's an actual problem. It went from >99% failure to 0% failure in my tests here.

josharian:
Windows test failure is unrelated. But depressing.

PR #3740, opened 2022-01-14
net/dns/resolvd: store nameservers

qbit:
Currently only search domains are stored. This was an oversight
(under?) on my part.

As things are now, when MagicDNS is on and "Override local DNS" is
off, the dns forwarder has to timeout before names resolve. This
introduces a pretty annoying lang that makes everything feel
extremely slow. You will also see an error: "upstream nameservers
not set".

I tested with "Override local DNS" on and off. In both situations
things seem to function as expected (and quickly).

Signed-off-by: Aaron Bieber <aaron@bolddaemon.com>

### DIFF
--- a/net/dns/resolvd.go
+++ b/net/dns/resolvd.go
@@ -16,6 +16,7 @@ import (
 	"regexp"
 	"strings"
 
+	"inet.af/netaddr"
 	"tailscale.com/types/logger"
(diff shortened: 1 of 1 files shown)


bradfitz (approved):
LGTM. I'll unify our three+ /etc/resolv.conf parsers later.

They're all in different states of bugginess.

PR #3749, opened 2022-01-17
Update installer.sh

JayWStapleton:
Add -y flag for yum-config-manager and yum install lines to stop it from failing while waiting for interactive confirmation. 

Tested and working on Oracle Linux 7

### DIFF
--- a/scripts/installer.sh
+++ b/scripts/installer.sh
@@ -427,8 +427,8 @@ main() {
 		yum)
 			set -x
 			$SUDO yum install yum-utils
-			$SUDO yum-config-manager --add-repo "https://pkgs.tailscale.com/$TRACK/$OS/$VERSION/tailscale.repo"
-			$SUDO yum install tailscale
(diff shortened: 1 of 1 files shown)


scripts/installer.sh
```diff
@@ -427,8 +427,8 @@ main() {
 		yum)
 			set -x
 			$SUDO yum install yum-utils
-			$SUDO yum-config-manager --add-repo "https://pkgs.tailscale.com/$TRACK/$OS/$VERSION/tailscale.repo"
-			$SUDO yum install tailscale
+			$SUDO yum-config-manager -y --add-repo "https://pkgs.tailscale.com/$TRACK/$OS/$VERSION/tailscale.repo"
+			$SUDO yum install tailscale -y
```
bradfitz, line 431: the -y doesn't come between install and tailscale?

bradfitz (approved):

scripts/installer.sh
JayWStapleton: This is how I tested it.

JayWStapleton (commented):
Signed-off-by: Jay Stapleton <jay@tailscale.com>

PR #3753, opened 2022-01-17
wgengine/netstack: fix netstack ping timeout on darwin

bradfitz:
-W is milliseconds on darwin, not seconds, and empirically it's
milliseconds after a 1 second base.

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -426,6 +426,11 @@ func (ns *Impl) userPing(dstIP netaddr.IP, pingResPkt []byte) {
 	switch runtime.GOOS {
 	case "windows":
 		err = exec.Command("ping", "-n", "1", "-w", "3000", dstIP.String()).Run()
+	case "darwin":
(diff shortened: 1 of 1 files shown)


apenwarr:
Ah, the reason I never noticed this is if you don't use -c, then -W doesn't seem to matter, and the timeout is defined by -i instead. If you use -c, -W seems to extend the timeout after the *last* ping.

```
$ time ping -W3000 -i5 -c2 8.8.8.9
PING 8.8.8.9 (8.8.8.9): 56 data bytes
Request timeout for icmp_seq 0

--- 8.8.8.9 ping statistics ---
2 packets transmitted, 0 packets received, 100.0% packet loss

real	0m13.144s
```

(The first "Request timeout" message was after 5 seconds, not 3)

apenwarr (approved):

DentonGentry:
Noting for future reference: FreeBSD also uses a -W argument in milliseconds, and empirically is also 1 second plus the extra milliseconds passed in as an argument.

PR #3770, opened 2022-01-19
wgengine/netstack: clear TCP ECN bits before giving to gvisor

bradfitz:
Updates #2642


bradfitz:
WIP; needs tests and hasn't even been tested by hand

net/packet/packet.go
```diff
@@ -180,7 +183,7 @@ func (q *Parsed) decode4(b []byte) {
 			}
 			q.Src = q.Src.WithPort(binary.BigEndian.Uint16(sub[0:2]))
 			q.Dst = q.Dst.WithPort(binary.BigEndian.Uint16(sub[2:4]))
-			q.TCPFlags = TCPFlag(sub[13]) & 0x3F
+			q.TCPFlags = TCPFlag(sub[13]) & 0xFF
```
danderson, line 186: Isn't this the same as removing `& 0xFF` entirely? You're masking off none bits of an 8-bit value.

net/packet/packet.go
```diff
@@ -282,7 +285,7 @@ func (q *Parsed) decode6(b []byte) {
 		}
 		q.Src = q.Src.WithPort(binary.BigEndian.Uint16(sub[0:2]))
 		q.Dst = q.Dst.WithPort(binary.BigEndian.Uint16(sub[2:4]))
-		q.TCPFlags = TCPFlag(sub[13]) & 0x3F
+		q.TCPFlags = TCPFlag(sub[13]) & 0xFF
```
danderson, line 288: Same

wgengine/netstack/netstack.go
```diff
@@ -793,3 +796,56 @@ func stringifyTEI(tei stack.TransportEndpointID) string {
...
+func removeECNBits(p *packet.Parsed) {
+	if p.IPVersion == 0 {
+		return
+	}
+	if p.IPProto != ipproto.TCP {
+		// TODO(bradfitz): handle non-TCP too? for now only trying to
+		// fix the Issue 2642 problem.
+		return
+	}
+	if p.TCPFlags&(packet.TCPCWR|packet.TCPECNEcho) == 0 {
```
danderson, line 810: Similar to TCPSynAck, I'd make a combined constant for this called something like TCPECN, so it's more obvious that you're masking ECN bits to someone who hasn't recently re-read that "ECN Echo" and "Congestion Window Reduced" are the two flags TCP ECN uses.

wgengine/netstack/netstack.go
```diff
@@ -793,3 +796,56 @@ func stringifyTEI(tei stack.TransportEndpointID) string {
...
+		// Nothing to do.
+		return
+	}
+
+	// Clear flags.
```
danderson, line 815: Feels a bit dicey to be reaching into parsed packets this far from where they're parsed, but I'll hold my nose since this is for debugging.

For more future robustness, make it a method on packet.Parsed, so the mangling of raw bytes is in the same place as the parsing code?

wgengine/netstack/netstack_test.go
```diff
@@ -74,3 +77,52 @@ func getMemStats() (ms runtime.MemStats) {
 	runtime.ReadMemStats(&ms)
 	return
 }
+
+func TestRemoveECNBits(t *testing.T) {
+	// withECNHex is a TCP SYN packet with ECN bits set in the TCP
+	// header as captured by Wireshark on macOS against the
+	// Tailscale interface.  For whatever reason, the ECN bits are
```
danderson, line 84: "Since the Transmission Control Protocol (TCP) does not perform congestion control on control packets (pure ACKs, SYN, FIN segments), control packets are usually not marked as ECN-capable."

That's why :) You captured a SYN, which indicates its ECN-awareness in the TCP headers, but by design does not advertise ECN-awareness at the IP layer.

danderson (approved):

net/packet/packet.go
bradfitz: whoops, yeah. I was reviewing it in wireshark which groups the prior byte into a uint16 and just brain-o'ed it here.

will remove.

wgengine/netstack/netstack.go
bradfitz: This isn't just for debugging; this is meant to be the fix.

> For more future robustness, make it a method on packet.Parsed, so the mangling of raw bytes is in the same place as the parsing code?

So that's how I started, but then I got lost in horrible checksum calculations and debating pulling out & rewriting gvisor's checksumming (because I hate its string-based tcpip.Address type) and then I moved it to wgengine/netstack so I could just depend on the gvisor packages, but then ultimately I ended up not doing that and finding out how to incrementally update the checksum.

But now that I've done that, I can move it back to net/packet.

Will do.

wgengine/netstack/netstack.go
```diff
@@ -793,3 +796,56 @@ func stringifyTEI(tei stack.TransportEndpointID) string {
 func ipPortOfNetstackAddr(a tcpip.Address, port uint16) (ipp netaddr.IPPort, ok bool) {
 	return netaddr.FromStdAddr(net.IP(a), int(port), "") // TODO(bradfitz): can do without allocs
 }
+
+// removeECNBits modifies p to remove ECN bits, if any.
+func removeECNBits(p *packet.Parsed) {
+	if p.IPVersion == 0 {
```
DentonGentry: I don't understand why this is here. It is harmless, just can't think of cases where it would be needed.

wgengine/netstack/netstack.go
bradfitz: done.

wgengine/netstack/netstack_test.go
bradfitz: Thanks, updated comment.

net/packet/packet.go
```diff
@@ -424,6 +434,40 @@ func (q *Parsed) IsEchoResponse() bool {
 	}
 }
 
+// RemoveECNBits modifies p and its underlying memory buffer to remove
+// ECN bits, if any. It reports whether it did so.
+//
+// It currently only does the TCP flags.
+func (p *Parsed) RemoveECNBits() bool {
```
danderson, line 441: Nit: none of the callers seem to care about the return value. Remove it? But it's also fine as-is.

danderson (approved):

net/packet/packet.go
bradfitz: test uses it! :)

I figured it might be useful in the future to decide when you need to deep copy the result or something, knowing the underlying buffer has been mutated.

wgengine/netstack/netstack.go
danderson: It's a necessary precondition to accessing the packet bytes. We're only guaranteed to have a TCP+IP header in the buffer if we successfully parsed it as some kind of IP plus TCP.

bradfitz:
Okay, now that the bug is understood (https://github.com/google/gvisor/issues/7075) I'm going to submit this workaround that we can cherry-pick back to the 1.20.x release.

Then we can fix the bug upstream in gvisor, pull in new gvisor (along with months of other changes) and use that in 1.22 and revert the workaround.

(FYI @DentonGentry)

PR #3778, opened 2022-01-20
words: fallout from mudpuppy-gate.

danderson:
Salamanders also have no scales. I checked the interweb, and there
doesn't seem to be any subspecies that would let us claim that
*some* salamanders are scaley.

But they are tailey, for sure.


catzkorn (approved):
Lets hope there isn't an amphibian uprising any time soon 😬

PR #3783, opened 2022-01-21
Makefile: update `make spk` target to use the new go spk builder

maisem:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,6 @@
 IMAGE_REPO ?= tailscale/tailscale
+SYNO_ARCH ?= "amd64"
+SYNO_DSM ?= "7"
 
 usage:
 	echo "See Makefile"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -32,9 +34,12 @@ staticcheck:
 	go run honnef.co/go/tools/cmd/staticcheck -- $$(go list ./... | grep -v tempfork)
 
 spk:
-	go run github.com/tailscale/tailscale-synology@main --version=build -o tailscale.spk --source=.
+	PATH="${PWD}/tool:${PATH}" ./tool/go run github.com/tailscale/tailscale-synology@main -o tailscale.spk --source=. --goarch=${SYNO_ARCH} --dsm-version=${SYNO_DSM}
+
+spkall:
```
bradfitz, line 39: that `-o` won't work with all

bradfitz (approved):
LGTM otherwise

Makefile
maisem: oops, thanks

PR #3784, opened 2022-01-21
cmd/derper: record TLS versions used for requests.

danderson:
Surveying the fleet prior to turning off old/unused/insecure
TLS versions.

Updates tailscale/corp#3615

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -51,9 +51,11 @@ var (
 )
 
 var (
-	stats           = new(metrics.Set)
-	stunDisposition = &metrics.LabelMap{Label: "disposition"}
-	stunAddrFamily  = &metrics.LabelMap{Label: "family"}
+	stats             = new(metrics.Set)
+	stunDisposition   = &metrics.LabelMap{Label: "disposition"}
+	stunAddrFamily    = &metrics.LabelMap{Label: "family"}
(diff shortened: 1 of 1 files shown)


bradfitz (approved):
LGTM but also interesting would be active sessions broken down by TLS version in addition to just new connection counters if you want to also add that.

That'd tell us the difference between e.g. some buggy/old scrapper hitting us a bunch that's not actually a DERP client.

danderson:
Hmm, possibly. I copied this code over from our other server, since right now I suspect the answer is 0.0qps for both TLS 1.0 and 1.1, and I didn't want to make something custom if that's all that happens.

Would you care about tracking TLS version in the long term, rather than as a one-off now? If so, I can make it session-based.

bradfitz:
> Would you care about tracking TLS version in the long term, rather than as a one-off now? If so, I can make it session-based.

Yeah, long term would be interesting, to watch TLS 1.3 adoption over time. At some point we may want to lock/ratchet it. It should always be 1.3 if there's no MITM. I want to see what percent of active sessions are corp-MITM'ed.

danderson:
Done, PTAL. Now have a gauge-per-version for active requests, and a per-request counter for seeing any short-lived requests.

bradfitz (approved):

PR #3794, opened 2022-01-24
cmd/tailscale: Propagate tailscaled 403s as AccessDeniedErrors

davideger:
linux users get helpful advice to run tailscale with --operator=$USER
when they try to 'tailscale file {cp,get}' but are mysteriously forbidden.

Signed-off-by: David Eger <david.eger@gmail.com>

### DIFF
--- a/client/tailscale/tailscale.go
+++ b/client/tailscale/tailscale.go
@@ -104,6 +104,10 @@ func doLocalRequestNiceError(req *http.Request) (*http.Response, error) {
 		if server := res.Header.Get("Tailscale-Version"); server != "" && server != version.Long && onVersionMismatch != nil {
 			onVersionMismatch(version.Long, server)
 		}
+		if res.StatusCode == 403 {
+			all, _ := ioutil.ReadAll(res.Body)
+			return nil, &AccessDeniedError{errors.New(errorMessageFromBody(all))}
+		}
 		return res, nil
 	}
 	if ue, ok := err.(*url.Error); ok {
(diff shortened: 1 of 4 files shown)


cmd/tailscale/cli/file.go
```diff
@@ -324,7 +324,7 @@ func runFileGet(ctx context.Context, args []string) error {
 	for {
 		wfs, err = tailscale.WaitingFiles(ctx)
 		if err != nil {
-			return fmt.Errorf("getting WaitingFiles: %v", err)
```
bradfitz, line 327: rather than remove this, use %w (wrapped error) rather than %v?

I assume you removed this just so `tailscale.IsAccessDenied` would work?

client/tailscale/tailscale.go
```diff
@@ -179,10 +183,6 @@ func send(ctx context.Context, method, path string, wantStatus int, body io.Read
 		return nil, err
 	}
 	if res.StatusCode != wantStatus {
-		if res.StatusCode == 403 {
```
bradfitz, line 182: any harm in leaving this? Are there any code paths that use send directly instead of `doLocalRequestNiceError` and would no longer get AccessDeniedErrors?

bradfitz (commented):
Thanks!

client/tailscale/tailscale.go
davideger: `tailscale file get .` calls doLocalRequestNiceError() instead of going through send() because it wants access to `res.ContentLength` as far as I can tell, so that's what prompted the move of this bit down the stack.

cmd/tailscale/cli/file.go
davideger: that's right; i'm a bit new to golang, didn't know %w was a thing :+1:

davideger (commented):
thanks for the education about %w :-)

bradfitz (approved):

PR #3801, opened 2022-01-24
envknob: add new package for all the strconv.ParseBool(os.Getenv(..))

bradfitz:
A new package can also later record/report which knobs are checked and
set. It also makes the code cleaner & easier to grep for env knobs.


crawshaw (approved):
Would be nice to have package doc commentary on the difference between controlknobs and the new package.

maybe it should have debug in the package name?

Issue #3802, opened 2022-01-24
SSH over Tailscale


bradfitz:
Since all packets sent over Tailscale have an associated identity thanks to Tailscale, a lot of authentication work can be simplified. Notably, SSH could be made much less tedious by taking advantage of your TCP connection's identity. The [cmd/tsshd](https://github.com/tailscale/tailscale/tree/main/cmd/tsshd) was the start of these experiments.

This is a tracking bug for more such SSH work.

DentonGentry:
Tailscale SSH feature launched: https://tailscale.com/blog/tailscale-ssh/

I imagine we should have new issues tracking future feature additions.

PR #3803, opened 2022-01-24
wgengine/netstack: add an SSH server experiment

bradfitz:
Disabled by default.

To use, run tailscaled with:

    TS_SSH_ALLOW_LOGIN=you@bar.com

And enable with:

    $ TAILSCALE_USE_WIP_CODE=true tailscale up --ssh=true

Then ssh [any-user]@[your-tailscale-ip] for a root bash shell.
(both the "root" and "bash" part are temporary)

Updates #3802


cmd/tailscale/cli/up.go
```diff
@@ -81,6 +82,8 @@ func acceptRouteDefault(goos string) bool {
 
 var upFlagSet = newUpFlagSet(effectiveGOOS(), &upArgs)
 
+func inTest() bool { return flag.Lookup("test.v") != nil }
```
crawshaw, line 85: ha clever

crawshaw (approved):
Looks disabled enough by default. :)

PR #3809, opened 2022-01-25
ipn/ipnserver: add TS_PERMIT_CERT_UID envknob to give webservers cert access

bradfitz:
So you can run Caddy etc as a non-root user and let it have access to
get certs.

Updates caddyserver/caddy#4541

### DIFF
--- a/ipn/ipnserver/server.go
+++ b/ipn/ipnserver/server.go
@@ -32,6 +32,7 @@ import (
 	"inet.af/netaddr"
 	"inet.af/peercred"
 	"tailscale.com/control/controlclient"
+	"tailscale.com/envknob"
 	"tailscale.com/ipn"
 	"tailscale.com/ipn/ipnlocal"
 	"tailscale.com/ipn/localapi"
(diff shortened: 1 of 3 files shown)


ipn/ipnserver/server.go
```diff
@@ -445,6 +446,26 @@ func (s *Server) localAPIPermissions(ci connIdentity) (read, write bool) {
 	return false, false
 }
 
+// connCanFetchCerts reports whether ci is allowed to fetch HTTPS
+// certs from this server when it wouldn't otherwise be able to.
+//
+// That is, this reports whether ti should grant additional
```
maisem, line 452: ci?

maisem (approved):

ipn/ipnserver/server.go
bradfitz: done

bradfitz:
Chatted with @danderson and we decided this is fine for now. We might do something fancier later.

mholt:
Is there anything special I need to do to get this to work?

I have a `caddy` user with UID 998, I added this to /etc/default/tailscaled:

```
TS_PERMIT_CERT_UID=998
```

(Also tried it in quotes: `TS_PERMIT_CERT_UID="998"`)

I was also sure to build Tailscale and Tailscaled from source at 03caa95bf21c1c59272c860c586eb0ff37a5a712 and verified with `tailscale --version` that it is on that commit.

But from Caddy I'm still getting `Access denied: cert access denied` errors (unless I run it as root). Any ideas why?

mholt:
Ah, `cp` returned an error status ("text file busy") because the service was running when I replaced `tailscaled` which is subtly different from `tailscale` -- oops. Always check your exit status, kids.

Working now. Sorry for the noise!

bradfitz:
> Ah, cp returned an error status ("text file busy") 

Protip: use [install(1)](https://linux.die.net/man/1/install) :)

ksylvan:
How does this work on MacOS? I have tailscale installed via `brew install --cask tailscale` and I don't know where to set the environment variable for tailscaled. Please help! Thanks.

PR #3813, opened 2022-01-26
wgengine/magicsock: make debugUseDerpRoute an opt.Bool.

### LINKED ISSUE #3812: ios build broken by envknob change
### What is the issue?

41fd4eab5c13b12dbafc16a4496af63f20432d03 added envknobs, which broke the wgengine/magicsock build for iOS:

```
wgengine/magicsock/magicsock.go:64:31: debugUseDerpRoute.Get undefined (type bool has no field or method Get)
```

This is because ios defines its own debugknobs_ios.go as constants, to generate smaller binaries.

@bradfitz not sure how to fix this. Switching debugknobs_ios to envknobs will probably reinflate binaries, so we might as well get rid of this hack... Right? Especially since ios gave us more memory?

danderson:
Can still be constant, just needs the extra methods.

Fixes #3812

### DIFF
--- a/wgengine/magicsock/debugknobs_ios.go
+++ b/wgengine/magicsock/debugknobs_ios.go
@@ -4,16 +4,18 @@
 
 package magicsock
 
+import "tailscale.com/types/opt"
+
 // All knobs are disabled on iOS.
 // Further, they're const, so the toolchain can produce smaller binaries.
 const (
-	debugDisco              = false
-	debugOmitLocalAddresses = false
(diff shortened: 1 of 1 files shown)


crawshaw (approved):

sailorfrag (approved):

bradfitz (approved):

bradfitz:
Whoops. Guess we're missing some CI coverage.

bradfitz:
Merging so I can send a CI change atop this.

Issue #1518, opened 2021-03-16
Break out gvisor netstack to a separate package


danderson:
The current netstack, for google organizational reasons, ships as part of gvisor. This means that trying to update netstack also pulls in gRPC and a bunch of other gvisor-only dependencies.

This is silly, and causes problems because nothing in this ecosystem is ever stable or compatible with other things.

As a result, I had to commit 84aba349d9a8d4e43585856c8155385f2569d35a to revert d42f8b7f9a29a288058a03ebdf740e11b843bf26. We tried to bump netstack to pick up a bugfix for 32-bit platforms, and that ended up making the etcd client fail to compile because of conflicting gRPC versions. Tailscale 1.6.0 is going to ship with broken 32-bit userspace networking as a result.

We want to make inet.af/netstack, which will be just the netstack code yanked out of the gvisor repo with a script. That should lead to a much smaller (maybe even nil?) go.mod, which will be much nicer for projects that just want to do userspace networking (like us) and not G all the visor.

/cc @namansood, @bradfitz to note the gvisor revert.

bradfitz:
The commit cross-reference didn't work, but: https://github.com/inetaf/netstack

Mostly done.

Need to clean up the sync automation scripts a bit and add to the repo.

bradfitz:
Current shell script hack job is: https://gist.github.com/bradfitz/889d7c3db6d0907f3653a56a88c38415

bradfitz:
Not quite done. The shell script isn't great yet and it's not automated or even included in the repo.

DentonGentry:
Is https://github.com/inetaf/netstack considered enough for this purpose?

bradfitz:
I wanted to keep this open until the repo at least has my hacky update script. But it's really hacky so I kept hoping I'd clean it up a bit.

bradfitz:
Latest iteration of the hacky script: https://gist.github.com/bradfitz/4ccf99a80820b3811683fa4803c64979

But we should really rewrite that to be robust.

Also, we should probably get some automated CI on that, doing it for every gvisor commit and running our tests, to prevent other regressions like https://github.com/google/gvisor/issues/6839

DentonGentry:
gvisor added tests to build on multiple platforms: https://buildkite.com/gvisor/pipeline/builds/10194

bradfitz:
As of Go 1.17 with its lazy module loading, we no longer need netstack in its own module. We can use the upstream gvisor module without breaking our corp build now.

[...]

PR #3815, opened 2022-01-26
go.mod: bump netstack, switch to upstream netstack

bradfitz:
Now that Go 1.17 has module graph pruning
(https://go.dev/doc/go1.17#go-command), we should be able to use
upstream netstack without breaking our private repo's build
that then depends on the tailscale.com Go module.

This is that experiment.

Updates #1518 (the original bug to break out netstack to own module)
Updates #2642 (this updates netstack, but doesn't remove workaround)


go.mod
```diff
@@ -56,6 +56,7 @@ require (
 	golang.org/x/tools v0.1.8
 	golang.zx2c4.com/wireguard v0.0.0-20211116201604-de7c702ace45
 	golang.zx2c4.com/wireguard/windows v0.4.10
+	gvisor.dev/gvisor v0.0.0-20220126021142-d8aa030b2591
 	honnef.co/go/tools v0.2.2
 	inet.af/netaddr v0.0.0-20211027220019-c74959edd3b6
 	inet.af/netstack v0.0.0-20211120045802-8aa80cf23d3c
```
maisem, line 61: I'd have expected this to disappear.

maisem (approved):

go.mod
bradfitz: Why? That's the module name that netstack is in.

bradfitz:
I'm not going to submit this yet until we discuss at an eng meeting. But the experiment worked; this doesn't break our corp build anymore (it did with Go 1.16)

go.mod
maisem: Didn't we fully move to upstream? I thought the inet.af/netstack was the fork that we're removing

maisem: run `go mod tidy`. these lines should disappear.

bradfitz: oh, you're commenting on line 62. The highlight above made me think you were commenting on line 59. 🤦‍♂️

bradfitz: The reason `go mod tidy` didn't clean it for me was because I had local test files not in git that were keeping `inet.af/netstack` alive. Fixed.

bradfitz:
Per eng meeting, we're going ahead with this.

PR #3818, opened 2022-01-26
scripts/installer: add Centos Stream 9 support

Xe:
Blocked on package server support being deployed

### DIFF
--- a/scripts/installer.sh
+++ b/scripts/installer.sh
@@ -142,7 +142,7 @@ main() {
 				VERSION="tumbleweed"
 				PACKAGETYPE="zypper"
 				;;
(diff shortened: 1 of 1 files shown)


scripts/installer.sh
```diff
@@ -142,7 +142,7 @@ main() {
 				VERSION="tumbleweed"
 				PACKAGETYPE="zypper"
 				;;
- 			arch)
+			arch)
```
Xe, line 145: This is a spacing fix that I only noticed when using Vim

catzkorn (approved):

bradfitz (approved):

PR #3820, opened 2022-01-26
cmd/derpprobe: check derper TLS certs too

bradfitz:
Screen Shot 2022-01-26 at 11 57 03 AM

### DIFF
--- a/cmd/derpprobe/derpprobe.go
+++ b/cmd/derpprobe/derpprobe.go
@@ -9,7 +9,9 @@ import (
 	"bytes"
 	"context"
 	crand "crypto/rand"
+	"crypto/x509"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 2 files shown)


cmd/derpprobe/derpprobe.go
```diff
@@ -93,6 +103,27 @@ func getOverallStatus() (o overallStatus) {
 			}
 		}
 	}
+
+	var subjs []string
+	for k := range certs {
+		subjs = append(subjs, k)
+	}
+	sort.Strings(subjs)
+	reissueTime := time.Unix(1643226768, 0)     // assume certs before this need reissuance by LetsEncrypt due to ALPN bug
```
maisem, line 112: Maybe make this a const with a small description of the let's encrypt issue

bradfitz: Done

PR #3821, opened 2022-01-26
cmd/derper: provide support for forcing autocert renewals

bradfitz:


cmd/derper/cert.go
```diff
@@ -39,6 +42,24 @@ func certProviderByCertMode(mode, dir, hostname string) (certProvider, error) {
 		if hostname == "derp.tailscale.com" {
 			certManager.HostPolicy = prodAutocertHostPolicy
 			certManager.Email = "security@tailscale.com"
+			if v, err := os.ReadFile("/home/derp/forcerenew.secret"); err == nil {
+				proto := strings.TrimSpace(string(v))
+				am2 := *certManager
```
danderson, line 47: Is it racy to use two autocert managers on the same cache dir? Do they assume things about not being mutated surprisingly?

bradfitz: Should be fine, but I ended up going a different route (in acme/autocert) rather than testing this :)

PR #3831, opened 2022-01-27
cmd/tailscale: fix up --reset, again

### LINKED ISSUE #3830: FR: make --reset remove all flags when called with no other arguments

currently, `tailscale up --reset ` does not have an effect when called on it's own. It is intended to be used with other arguments to reset all non-specified flags to default. 

The intuitive behaviour would be to have it reset all flags to default when called with no other arguments. 

### How should we solve this?

make `tailscale up --reset` set all flags to default values

confusing UX

### Anything else?

_No response_

bradfitz:
Also fix a somewhat related printing bug in the process where
some paths would print "Success." inconsistently even
when there otherwise was no output (in the EditPrefs path)

Fixes #3830
Updates #3702 (which broke it once while trying to fix it)

### DIFF
--- a/cmd/tailscale/cli/up.go
+++ b/cmd/tailscale/cli/up.go
@@ -387,7 +387,8 @@ func prefsFromUpArgs(upArgs upArgsT, warnf logger.Logf, st *ipnstate.Status, goo
 	return prefs, nil
 }
 
-// updatePrefs updates prefs based on curPrefs
+// updatePrefs returns how to edit preferences based on the
+// flag-provided 'prefs' and the currently active 'curPrefs'.
 //
 // It returns a non-nil justEditMP if we're already running and none of
 // the flags require a restart, so we can just do an EditPrefs call and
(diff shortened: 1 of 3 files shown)


cmd/tailscale/cli/up.go
```diff
@@ -424,11 +425,16 @@ func updatePrefs(prefs, curPrefs *ipn.Prefs, env upCheckEnv) (simpleUp bool, jus
 		env.upArgs.authKeyOrFile == "" &&
 		!controlURLChanged &&
 		!tagsChanged
+
 	if justEdit {
 		justEditMP = new(ipn.MaskedPrefs)
 		justEditMP.WantRunningSet = true
 		justEditMP.Prefs = *prefs
-		env.flagSet.Visit(func(f *flag.Flag) {
+		visitFlags := env.flagSet.Visit
```
maisem, line 433: oh!

maisem (approved):

PR #3841, opened 2022-01-29
derp/derphttp, wgengine/magicsock: prefer IPv6 to DERPs when IPv6 works

### LINKED ISSUE #3838: derp/derphttp: dialNode IPv4 vs IPv6 race should prefer v6
Currently the DERP dialer races v4 and v6:

```go
        if shouldDialProto(n.IPv4, netaddr.IP.Is4) {
                startDial(n.IPv4, "tcp4")
        }
        if shouldDialProto(n.IPv6, netaddr.IP.Is6) {
                startDial(n.IPv6, "tcp6")
        }
```

If it looks like the machine has working global IPv6 (e.g. a previous UDP STUN to that same DERP's global v6 worked), it should prefer IPv6 on the dialing.

This is better for various reasons: less NAT work for clients, phones/carriers, and potentially cheaper server costs for us.

bradfitz:
Fixes #3838

### DIFF
--- a/derp/derphttp/derphttp_client.go
+++ b/derp/derphttp/derphttp_client.go
@@ -26,6 +26,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"go4.org/mem"
@@ -63,6 +64,12 @@ type Client struct {
 
 	ctx       context.Context // closed via cancelCtx in Client.Close
 	cancelCtx context.CancelFunc
+
+	// addrFamSelAtomic is the last AddressFamilySelector set
+	// by SetAddressFamilySelector. It's an atomic because it needs
+	// to be accessed by multiple racing routines started while
+	// Client.conn holds mu.
+	addrFamSelAtomic atomic.Value // of AddressFamilySelector
 
 	mu           sync.Mutex
 	preferred    bool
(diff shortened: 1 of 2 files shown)


DentonGentry (approved):

derp/derphttp/derphttp_client.go
```diff
@@ -583,6 +616,18 @@ func (c *Client) dialNode(ctx context.Context, n *tailcfg.DERPNode) (net.Conn, e
 	startDial := func(dstPrimary, proto string) {
 		nwait++
 		go func() {
+			if proto == "tcp4" && c.preferIPv6() {
```
danderson, line 619: Why can't we use net.Dialer instead of all this machinery? Doesn't it already have logic to do happy eyeball racing?

I'm sure there's a good reason, I just can't see it with the tiny amount of context that github offers.

bradfitz: Our DERP dialer has all those tailcfg/derpmap.go knobs,

https://pkg.go.dev/tailscale.com/tailcfg#DERPNode

... that let us/users control much more than what net.Dialer allows. It lets us do DNS or no DNS (hard-code v4 and/or v6 addresses), and control which cert is checked (which might different from the DNS lookup for domain fronting).

There's no way to use net.Dialer and retain all those knobs. But fortunately there's not much to net.Dialer's implementation that's not what this is doing, racing some goroutines.

Issue #3842, opened 2022-01-30
DNS resolution not working after turning exit node 


cepera-ang:
### What is the issue?

After updating from Tailscale 1.18 to Tailscale 1.20.2 I no longer can use exit node functionality. I have ubuntu cloud machine as exit node (named vpn) and a windows machine. After enabling exit node on windows I get all DNS requests going to 100.100.100.100 and dying in timeout. The same requests from older version work flawlessly. 

For ex, windows, 1.20.2, exit node off:
```
λ nslookup github.com
Server:  one.one.one.one
Address:  1.1.1.1

Non-authoritative answer:
Name:    github.com
Address:  140.82.121.3
```

Windows, 1.20.2, exit node on:

```
λ nslookup github.com
Server:  UnKnown
Address:  100.100.100.100

DNS request timed out.
    timeout was 2 seconds.
DNS request timed out.
    timeout was 2 seconds.
DNS request timed out.
    timeout was 2 seconds.
```

Linux, 1.18.2, exit node on or off, whatever (same result) :
```
user@user-pi:~$ nslookup github.com 100.100.100.100
Server:         100.100.100.100
Address:        100.100.100.100#53

Non-authoritative answer:
Name:   github.com
Address: 140.82.121.4
```

Linux, 1.20.2, exit node on: 
```
user@user-pi:~$ tailscale version
1.20.2
  tailscale commit: 312750ddd288cf4073cfaef56a45102b9c1e8421
  other commit: 2c164d9c7443e2f3014fa54ea45e946b35152680
  go version: go1.17.6-tse44d304e54
user@user-pi:~$ nslookup github.com 100.100.100.100
;; connection timed out; no servers could be reached

```

Well, anyway, it seems like 100.100.100.100 not working anywhere in 1.20.2 for me. 

I see some changes related to DNS and exit nodes in release notes. Is there some configuration I have to do, in order to get this working again? 

### Are there any recent changes that introduced the issue?

Updated Tailscale everywhere to the latest version. 

### OS

Linux, Windows

### OS version

Ubuntu 20.04.3 LTS (GNU/Linux 5.11.0-1027-oracle aarch64), Microsoft Windows [Version 10.0.19044.1466]

### Tailscale version

1.20.2

### Bug report

BUG-c2f835af9713719097081eaf7976601903d023065d119901ad8e2e1799922664-20220130093427Z-bd88e452804a0817

Murgeye:
[...]

PR #3851, opened 2022-01-31
ipn/ipnlocal, wgengine/netstack: use netstack for peerapi server

bradfitz:
We're finding a bunch of host operating systems/firewalls interact poorly
with peerapi. We either get ICMP errors from the host or users need to run
commands to allow the peerapi port:

https://github.com/tailscale/tailscale/issues/3842#issuecomment-1025133727

... even though the peerapi should be an internal implementation detail.

Rather than fight the host OS & firewalls, this change handles the
server side of peerapi entirely in netstack (except on iOS), so it
never makes its way to the host OS where it might be messed with. Two
main downsides are:

1) netstack isn't as fast, but we don't really need speed for peerapi.
   And actually, with fewer trips to/from the kernel, we might
   actually make up for some of the netstack performance loss by
   staying in userspace.

2) tcpdump / Wireshark etc packet captures will no longer see the peerapi
   traffic. Oh well. Crawshaw's been wanting to add packet capture server
   support to tailscaled, so we'll probably do that sooner now.

A future change might also then use peerapi for the client-side
(except on iOS).

Updates #3842 (probably fixes, as well as many exit node issues I bet)

Front logo [Front conversations](https://app.frontapp.com/open/top_49gpd)

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -215,7 +215,7 @@ func NewLocalBackend(logf logger.Logf, logid string, store ipn.StateStore, diale
 	wiredPeerAPIPort := false
 	if ig, ok := e.(wgengine.InternalsGetter); ok {
 		if tunWrap, _, ok := ig.GetInternals(); ok {
-			tunWrap.PeerAPIPort = b.getPeerAPIPortForTSMPPing
+			tunWrap.PeerAPIPort = b.GetPeerAPIPort
 			wiredPeerAPIPort = true
 		}
 	}
@@ -1788,7 +1788,9 @@ func (b *LocalBackend) setPrefsLockedOnEntry(caller string, newp *ipn.Prefs) {
 	b.send(ipn.Notify{Prefs: newp})
 }
 
-func (b *LocalBackend) getPeerAPIPortForTSMPPing(ip netaddr.IP) (port uint16, ok bool) {
+// GetPeerAPIPort returns the port number for the peerapi server
+// running on the provided IP.
+func (b *LocalBackend) GetPeerAPIPort(ip netaddr.IP) (port uint16, ok bool) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 	for _, pln := range b.peerAPIListeners {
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/local.go
```diff
@@ -1799,6 +1801,27 @@ func (b *LocalBackend) getPeerAPIPortForTSMPPing(ip netaddr.IP) (port uint16, ok
 	return 0, false
 }
 
+// ServePeerAPIConnection served an already-accepted connection c.
```
DentonGentry, line 1804: ServePeerAPIConnection serves? (present tense)

wgengine/netstack/netstack.go
```diff
@@ -82,9 +82,12 @@ type Impl struct {
...
 	logf      logger.Logf
 	dialer    *tsdial.Dialer
-	ctx       context.Context    // alive until Close
-	ctxCancel context.CancelFunc // called on Close
-	lb        *ipnlocal.LocalBackend
+	ctx       context.Context        // alive until Close
+	ctxCancel context.CancelFunc     // called on Close
+	lb        *ipnlocal.LocalBackend // or nil
+
+	peerapi4Atomic uint32 // of uint16
```
DentonGentry, line 89: I don't understand what "of uint16" is intended to convey. This is a TCP port number, maybe?

crawshaw:
General concept sounds good. I don't have time to review this morning.

ipn/ipnlocal/local.go
```diff
@@ -1788,7 +1788,9 @@ func (b *LocalBackend) setPrefsLockedOnEntry(caller string, newp *ipn.Prefs) {
 	b.send(ipn.Notify{Prefs: newp})
 }
 
-func (b *LocalBackend) getPeerAPIPortForTSMPPing(ip netaddr.IP) (port uint16, ok bool) {
+// GetPeerAPIPort returns the port number for the peerapi server
+// running on the provided IP.
+func (b *LocalBackend) GetPeerAPIPort(ip netaddr.IP) (port uint16, ok bool) {
```
DentonGentry, line 1793: So we bind to port 0 and let the kernel pick a port, but then pick off incoming packets to the port in netstack?

The main thing I'm trying to think through is if there could be a case where we close the kernel socket, potentially allowing some other process to bind to that same ephemeral port, but we neglect to cleanup the netstack listener and end up intercepting traffic which doesn't belong to us.

maisem:
for posterity, why no iOS? we shouldn't be hitting memory limits anymore, so is it for battery usage?

ipn/ipnlocal/local.go
bradfitz: that's why on every incoming SYN to netstack, we double-check that the PeerAPIPort is accurate.

bradfitz:
> for posterity, why no iOS? we shouldn't be hitting memory limits anymore, so is it for battery usage?

Only iOS 15 gives us 50MB. Something like half of users are still on iOS 14.

We can enable it conditionally at runtime, but that doesn't help the binary size bloat, which still hurts us with our current version of Go due to the relocations. Josh's relocation work in the Go linker might make it possible for us to do that once we switch to Go 1.18.

JayWStapleton (commented):
Tested and confirmed that it solved a recurring exit node DNS error we've been seeing.

ipn/ipnlocal/local.go

wgengine/netstack/netstack.go
bradfitz: reworded

ipn/ipnlocal/local.go
bradfitz: > So we bind to port 0 and let the kernel pick a port, but then pick off incoming packets to the port in netstack?

Yes.

PR #3853, opened 2022-01-31
logtail: add a few new methods to PublicID

josharian:
These are for use in our internal systems.

### DIFF
--- a/logtail/id.go
+++ b/logtail/id.go
@@ -7,6 +7,7 @@ package logtail
 import (
 	"crypto/rand"
 	"crypto/sha256"
+	"encoding/binary"
 	"encoding/hex"
 	"errors"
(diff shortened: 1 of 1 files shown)


logtail/id.go
```diff
@@ -156,3 +157,29 @@ func fromHexChar(c byte) (byte, bool) {
...
+		}
+	}
+	return false // equal
+}
+
+func (id PublicID) IsZero() bool {
+	return id == PublicID{}
+}
+
+func (id PublicID) Prefix16() uint16 {
```
dsnet, line 175: We don't use Prefix16 and Prefix32 anymore, so might as well nuke them.

dsnet (approved):

Issue #3797, opened 2022-01-24
FR: list subnets advertised in API device list


lewisstevens1:
### What are you trying to do?

I would like to be able to list all the devices and maybe return a value such as a hostname / if it's advertising subnets.
This way I will be able to pass the device to the second command.

Also, I would like to be able to disable key expiry, via the API with the output of the first command.

### How should we solve this?

Add this via the API

### What is the impact of not solving this?

The server will keep expiring (everything has to be automated)

DentonGentry:
Setting a device to never expire using the API is being tracked in https://github.com/tailscale/tailscale/issues/2418 (and is currently being worked on).

> I would like to be able to list all the devices and maybe return a value such as a hostname / if it's advertising subnets.
This way I will be able to pass the device to the second command.

You're asking for a different API call to return the hostname and whether it is advertising subnets? Something more than https://github.com/tailscale/tailscale/blob/main/api.md#device-get ?

lewisstevens1:
Hello @DentonGentry, thankyou #2418 will cover that side of this. 

In regards to whether it is advertising subnets, yes, we need to be able to find which is the latest Tailscale instance that has been created that is currently active.

We don't seem to have a way to tell that:

- Filter if the computer is currently online (we have to take current time and iterate if any were active in the last 5 mins etc which can get quite confusing, so when we do launches of new instances we are not 100% sure if someone is using Tailscale at all).

- Whether it is the tailscale instance other than our computers (which the only way I can think of checking if if its advertising subnets) I have tried taking `advertisedRoutes` and trying to convert the list into a boolean, but I guess if there is a way to maybe filter by tags that may be a better solution.


![Screenshot 2022-01-26 at 11 25 57](https://user-images.githubusercontent.com/10538348/151155443-2d9a7ef6-d02a-48ea-9d8b-52c3691a5c7b.jpg)

DentonGentry:
@lewisstevens1 you might also look at the output of `tailscale status --json` on the device.

DentonGentry:
Regarding subnets advertised: does https://github.com/tailscale/tailscale/blob/main/api.md#device-routes-get do what is needed?

DentonGentry:
Closing as https://github.com/tailscale/tailscale/blob/main/api.md#device-routes-get appears to provide the needed functionality, please reopen if not.

PR #3861, opened 2022-02-02
tsweb: add a Server, which serves HTTP/HTTPS with various good defaults.

danderson:
First pass at an API and general structure for a unified "build me an HTTP/S server with sensible defaults", while still being transparent enough to allow for tweaking. Thoughts?

(marking draft because lacking tests, and I need to check against our existing HTTP servers to see what tweaks each would require, and if that still results in a net positive)

Updates #3797


tsweb/server.go
```diff
@@ -0,0 +1,345 @@
...
+	"tailscale.com/metrics"
+	"tailscale.com/types/logger"
+)
+
+const defaultTimeout = 10 * time.Second
+
+// ServerConfig is the initial configuration of a Server, for
+// consumption by NewServer.
+type ServerConfig struct {
+	// Name is a human-readable, alphanumeric name for the HTTP
```
bradfitz, line 27: give an example string literal in the comment (such as "foo") or (such as "foo.example.com") or (such as "Brad's Nice server")

Notably to clarify whether it's a hostname (with dots) or some bare name or some descriptive name.

Alphanumeric sounds like just "foo"?

tsweb/server.go
```diff
@@ -0,0 +1,345 @@
...
+	// server.
+	// The name is for internal use only, for things like naming cache
+	// directories, identifying the source of autocert emails,
+	// human-readable debug pages, ...
+	Name string
+	// Addr specifies the TCP address for the server to listen on, in
```
bradfitz, line 33: is there a default value?

tsweb/server.go
```diff
@@ -0,0 +1,345 @@
...
+	// is automatically configured and a second listener is set up on
+	// port 80 to handle HTTP to HTTPS redirection. All other ports
+	// result in HTTP-only serving.
+	Addr string
+	// Handler is the HTTP.Handler to invoke to handle
+	// requests. Cannot be nil.
+	Handler http.Handler
+	// Hostnames is the list of hostnames that the server accepts for
+	// serving TLS requests.
+	Hostnames []string
```
bradfitz, line 44: maybe something like https://pkg.go.dev/golang.org/x/crypto/acme/autocert#HostPolicy instead?

then callers can use funcs if needed but otherwise use helpers like https://pkg.go.dev/golang.org/x/crypto/acme/autocert#HostWhitelist ?

(just using autocert as example)

tsweb/server.go
```diff
@@ -0,0 +1,345 @@
...
+
+	// HSTSNoSubdomains is whether to tell browsers to only apply
+	// strict HTTPS serving to the current domain being requested,
+	// rather than the request domain and all its subdomains (the
+	// default).
+	HSTSNoSubdomains bool
+
+	// RequestLog is a log sink to write access logs for requests
```
bradfitz, line 52: this loses all the structure, though?  (byte counts, http status type, etc)

tsweb/server.go
```diff
@@ -0,0 +1,345 @@
...
+	// handled by the server. If nil, no access logs are written.
+	RequestLog logger.Logf
+
+	// TODO: some kind of dev mode? Alter how request logging and
+	// error handling is done to be more human-friendly?
+}
+
+// Server is an HTTP+HTTPS server, with a bunch of safe defaults
```
bradfitz, line 60: Mention the constructor, as that's what really does the "preconfigured" part, right?

tsweb/server.go
danderson: Added words. Basically has to be valid as part of an email address, and as a directory name.

tsweb/server.go
danderson: Documented.

tsweb/server.go
danderson: Changed to HostPolicy. The line I'm trying to walk is avoiding too much boilerplate in common cases, and I pictured the common case as "here's the 1 hostname I have". Other servers could go monkey with autocert configuration by hand before starting the server. But `[]string{` vs `autocert.HostWhitelist` isn't that much extra boilerplate.

tsweb/server.go
danderson: Switched to a structured callback. Again my goal here was that everything currently writes to a Logf, so by having a structured callback here I'm forcing the same boilerplate onto everyone in order to logf it. Maybe that won't be terrible, I dunno.

tsweb/server.go
danderson: Added more words on how Server should be used (use constructor, customize if desired, then ListenAndServe)

tsweb/server.go
```diff
@@ -0,0 +1,385 @@
...
+			s.CertManager.HostPolicy = func(context.Context, string) error {
+				return errors.New("no TLS serving allowed")
+			}
+		}
+		s.HTTPS = &http.Server{
+			Addr:         cfg.Addr,
+			Handler:      s,
+			ReadTimeout:  defaultTimeout,
+			WriteTimeout: defaultTimeout,
+			IdleTimeout:  defaultTimeout,
```
bradfitz, line 184: 10 seconds idle timeout is quite a bit lower than what we currently have in places, isn't it?

tsweb/server.go
```diff
@@ -0,0 +1,385 @@
...
+		if s.CertManager.HostPolicy == nil {
+			s.CertManager.HostPolicy = func(context.Context, string) error {
+				return errors.New("no TLS serving allowed")
+			}
+		}
+		s.HTTPS = &http.Server{
+			Addr:         cfg.Addr,
+			Handler:      s,
+			ReadTimeout:  defaultTimeout,
+			WriteTimeout: defaultTimeout,
```
bradfitz, line 183: this WriteTimeout is for the entirely response payload. That means 10 seconds to download a 45 MB tar.gz download of Tailscale, for instance. It's often not what people expect.

tsweb/server.go
```diff
@@ -0,0 +1,385 @@
...
+			label = "1.2"
+		case tls.VersionTLS13:
+			label = "1.3"
+		}
+		defer s.tlsRequests.Add(label, 1)
+		s.tlsActive.Add(label, 1)
+		defer s.tlsActive.Add(label, -1)
+	}
+
+	for k, v := range s.AlwaysHeaders {
```
bradfitz, line 286: how about only set these if they're not already present in the map? Then callers can opt-out per Handler if they need to (e.g. be embeddable in a frame)

That's what Go does itself for things like the "Date" header... it's automatic to add Date unless you set the "Date" map key to `nil`.

bradfitz:
(Closing; this became `safeweb` in this repo)

Issue #869, opened 2020-10-29
windows: failed to look up user from userid: No mapping between account names and security IDs was done.


bradfitz:
User reports, with 1.1.556:

![image (5)](https://user-images.githubusercontent.com/2621/97598813-93ddc080-19c4-11eb-8c18-4516ea2de7d4.png)

Internet says:
* https://docs.microsoft.com/en-us/troubleshoot/windows-server/group-policy/scecli-1202-events
* https://www.rebeladmin.com/2016/01/how-to-fix-error-no-mapping-between-account-names-and-security-ids-in-active-directory/

/cc @alexbrainman

bradfitz:
Sent https://github.com/tailscale/tailscale/pull/871

bradfitz:
Good enough for now.

DentonGentry:
Does still happen in the wild: `2021-11-05 18:02:06.1458677 -0500 -0500: 2021/11/05 18:02:06 [warning] issue 869: os/user.LookupId failed; ignoring`

bradfitz:
Reopening for @dblohm7 to investigate properly, as my earlier fix was a band-aid at best.

dblohm7:
For starters, I filed https://github.com/golang/go/issues/49509

PR #3862, opened 2022-02-02
ipn/ipnserver, util/winutil: update workaround for os/user.LookupId f…

dblohm7:
…ailures on Windows to reject SIDs from deleted/invalid security principals

Our current workaround made the user check too lax, thus allowing deleted
users. This patch adds a helper function to winutil that checks that the
uid's SID represents a valid Windows security principal.

Now if `lookupUserFromID` determines that the SID is invalid, we simply
propagate the error.

Updates https://github.com/tailscale/tailscale/issues/869

### DIFF
--- a/util/winutil/winutil_notwindows.go
+++ b/util/winutil/winutil_notwindows.go
@@ -2,23 +2,12 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-//go:build !windows
-// +build !windows
-
 package winutil
 
-const RegBase = ``
+const regBase = ``
 
-// GetRegString looks up a registry path in our local machine path, or returns
-// the given default if it can't.
-//
-// This function will only work on GOOS=windows. Trying to run it on any other
-// OS will always return the default value.
-func GetRegString(name, defval string) string { return defval }
+func getRegString(name, defval string) string { return defval }
 
-// GetRegInteger looks up a registry path in our local machine path, or returns
-// the given default if it can't.
-//
-// This function will only work on GOOS=windows. Trying to run it on any other
-// OS will always return the default value.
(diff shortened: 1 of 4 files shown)


util/winutil/winutil_notwindows.go
```diff
@@ -22,3 +22,13 @@ func GetRegString(name, defval string) string { return defval }
 // This function will only work on GOOS=windows. Trying to run it on any other
 // OS will always return the default value.
 func GetRegInteger(name string, defval uint64) uint64 { return defval }
+
+// IsSIDValidPrincipal determines whether the SID contained in uid represents a
```
bradfitz, line 26: only document this once in a portable file, otherwise you'll either have to mirror it in two files and keep it in sync or it won't show up in godoc for people on the 'wrong' platform(s).

common pattern is put the docs on the exported func in a portable file and then delegate to an unexported func/method in platform-specific files.

dblohm7: Got it. I've reorganized the affected functions according to that convention.

util/winutil/winutil_windows.go
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build windows
```
bradfitz, line 5: this is fine but these lines are unnecessary when you the filename is already a _GOOS.go or _GOARCH.go or _GOOS_GOARCH.go suffix as it is in this case.

see https://pkg.go.dev/cmd/go#hdr-Build_constraints

bradfitz (approved):

PR #3863, opened 2022-02-02
chirp: handle multiline responses from BIRD

maisem:
Also add tests to verify the parsing logic.

Fixes tailscale/corp#3639

### DIFF
--- a/chirp/chirp.go
+++ b/chirp/chirp.go
@@ -10,6 +10,7 @@ import (
 	"bufio"
 	"fmt"
 	"net"
+	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


sailorfrag (approved):

chirp/chirp.go
```diff
@@ -10,6 +10,7 @@ import (
 	"bufio"
 	"fmt"
 	"net"
+	"regexp"
```
bradfitz, line 13: FYI we try to avoid the regexp package for iOS/binary size reasons, but since this isn't on iOS, it's fine

maisem: does darwin imply iOS?

``` 
// tailscaled_bird.go
//go:build linux || darwin || freebsd || openbsd
// +build linux darwin freebsd openbsd
```

DentonGentry (approved):

chirp/chirp.go
maisem: oh right, no tailscaled on iOS

PR #3864, opened 2022-02-02
wgengine/netstack: start supporting different SSH users

bradfitz:
Updates #3802

### DIFF
--- a/wgengine/netstack/ssh.go
+++ b/wgengine/netstack/ssh.go
@@ -107,7 +107,13 @@ func (ns *Impl) handleSSH(s ssh.Session) {
 		return
 	}
 
-	cmd := exec.Command("/bin/bash")
+	var cmd *exec.Cmd
+	sshUser := s.User()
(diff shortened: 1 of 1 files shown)


wgengine/netstack/ssh.go
```diff
@@ -107,7 +107,13 @@ func (ns *Impl) handleSSH(s ssh.Session) {
 		return
 	}
 
-	cmd := exec.Command("/bin/bash")
+	var cmd *exec.Cmd
+	sshUser := s.User()
+	if os.Getuid() != 0 || sshUser == "root" {
+		cmd = exec.Command("/bin/bash")
+	} else {
+		cmd = exec.Command("/bin/su", "-", sshUser)
```
maisem, line 115: fine for now, but maybe use `/usr/bin/env su`? e.g. this doesn't work on NixOS

maisem (approved):

wgengine/netstack/ssh.go
bradfitz: done

PR #3865, opened 2022-02-02
ipn: add `CurrentTailnet` info to `tailscale status --json`

rosszurowski:
We need this for some Docker-related features. There's currently no way to detect the MagicDNS status from the `status --json` output, so this commit adds it.

Fixes tailscale/corp#3670

Signed-off-by: Ross Zurowski <ross@rosszurowski.com>

### DIFF
--- a/ipn/ipnstate/ipnstate.go
+++ b/ipn/ipnstate/ipnstate.go
@@ -33,10 +33,6 @@ type Status struct {
 	//  "Starting", "Running".
 	BackendState string
 
-	// TailnetName is the name of the network that's currently in
-	// use.
-	TailnetName string
-
 	AuthURL      string       // current URL provided by control to authorize client
 	TailscaleIPs []netaddr.IP // Tailscale IP(s) assigned to this node
 	Self         *PeerStatus
@@ -46,12 +42,14 @@ type Status struct {
 	// problems are detected)
 	Health []string
 
-	// MagicDNSSuffix is the network's MagicDNS suffix for nodes
-	// in the network such as "userfoo.tailscale.net".
-	// There are no surrounding dots.
-	// MagicDNSSuffix should be populated regardless of whether a domain
-	// has MagicDNS enabled.
+	// This field is the legacy name of CurrentTailnet.MagicDNSSuffix.
+	//
+	// Deprecated: use CurrentTailnet.MagicDNSSuffix instead.
 	MagicDNSSuffix string
+
+	// CurrentTailnet is information about the tailnet that the node
+	// is currently connected to. When not connected, this field is nil.
+	CurrentTailnet *TailnetStatus
 
 	// CertDomains are the set of DNS names for which the control
 	// plane server will assist with provisioning TLS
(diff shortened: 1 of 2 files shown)


ipn/ipnstate/ipnstate.go
```diff
@@ -53,6 +53,9 @@ type Status struct {
 	// has MagicDNS enabled.
 	MagicDNSSuffix string
 
+	// MagicDNSEnabled is whether or not the network has MagicDNS enabled.
```
bradfitz, line 56: You should probably clarify that even if this is true, MagicDNS is not necessarily enabled on this node if `--accept-dns=false`. That is, this is purely about whether it's enabled on the tailnet.

bradfitz (approved):

ipn/ipnstate/ipnstate.go
rosszurowski: Good point!

bradfitz: in fact maybe the field should be more clearly named too then.

Or maybe we should make a Tailnet-specific JSON object to put all such fields.

bradfitz: (because who really reads docs? people just run `tailscale status --json` and guess what things mean usually)

rosszurowski: Any objection to this kind of structure within the `Status` struct?

```go
Tailnet struct {
  Name string
  MagicDNSEnabled bool
}
```

bradfitz: Don't do an anonymous struct in this case. Make a named type.

rosszurowski: Yeah, I'll make a named struct. Just wanted your thoughts on the structure.

ipn/ipnstate/ipnstate.go
```diff
@@ -33,10 +33,6 @@ type Status struct {
 	//  "Starting", "Running".
 	BackendState string
 
-	// TailnetName is the name of the network that's currently in
```
rosszurowski, line 36: This field was never released in a public build, so it seems ideal to move it before folks rely on it.

ipn/ipnstate/ipnstate.go
```diff
@@ -64,6 +63,26 @@ type Status struct {
 	User map[tailcfg.UserID]tailcfg.UserProfile
 }
 
+// TailnetStatus is a subset of status information related to the
+// current tailnet.
```
bradfitz, line 67: I think this field itself can describe any tailnet, like when we have multiple at once.

I'd drop this "current" and move "current" to the currently-undocumented Tailnet field above.

Maybe call that field CurrentTailnet?

ipn/ipnstate/ipnstate.go
```diff
@@ -51,8 +47,11 @@ type Status struct {
 	// There are no surrounding dots.
 	// MagicDNSSuffix should be populated regardless of whether a domain
 	// has MagicDNS enabled.
+	// This field is duplicated into Tailnet.MagicDNSSuffix.
 	MagicDNSSuffix string
 
+	Tailnet TailnetStatus
```
bradfitz, line 53: Make it a pointer so it's JSON null when you're logged out? But then document what nil here means.
 
Otherwise you need to document the meaning of Tailnet.Name == "".

bradfitz:
@rosszurowski, @DentonGentry, we need to finish this effort before 1.22 is cut. We don't want to ship with the `Status.TailnetName` otherwise we can never really remove it.

rosszurowski:
I can get to this tomorrow. Sorry, been closing out other things first.

ipn/ipnstate/ipnstate.go
```diff
@@ -64,6 +65,26 @@ type Status struct {
 	User map[tailcfg.UserID]tailcfg.UserProfile
 }
 
+// TailnetStatus is a subset of status information related to the
```
bradfitz, line 68: // TailnetStatus is information about a Tailscale network ("tailnet").

rosszurowski:
Ok, updated this PR with your suggested changes, which made sense to me. PTAL!

ipn/ipnstate/ipnstate.go
```diff
@@ -64,6 +65,26 @@ type Status struct {
...
+// a tailnet this node is connected to.
+type TailnetStatus struct {
+	// TailnetName is the name of the network that's currently in
```
bradfitz, line 71: subject doesn't match field. Say "Name is the ..."

rosszurowski:
Pushed up doc fixes, will merge after tests pass, just to be safe.

ipn/ipnstate/ipnstate.go
juanfont: > (because who really reads docs? people just run `tailscale status --json` and guess what things mean usually)
 
Hey there ^^ ! I've just landed here searching for a bug in our integration tests on tailscale < 1.22

PR #3870, opened 2022-02-03
chirp: remove regex dependency

maisem:

### DIFF
--- a/chirp/chirp.go
+++ b/chirp/chirp.go
@@ -10,7 +10,6 @@ import (
 	"bufio"
 	"fmt"
 	"net"
-	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


chirp/chirp.go
```diff
@@ -89,7 +88,20 @@ func (b *BIRDClient) exec(cmd string, args ...interface{}) (string, error) {
 	return b.readResponse()
 }
 
-var respCodeRegex = regexp.MustCompile(`^\d{4}[ -]`)
+// hasResponseCode reports whether the provided byte slice is
+// prefixed with the BIRD response code.
```
bradfitz, line 92: "with a" ?

chirp/chirp.go
```diff
@@ -89,7 +88,20 @@ func (b *BIRDClient) exec(cmd string, args ...interface{}) (string, error) {
...
+// Equivalent regex: `^\d{4}[ -]`.
+func hasResponseCode(s []byte) bool {
+	if len(s) < 5 {
+		return false
+	}
+	for i := 0; i < 4; i++ {
+		if s[i] < '0' || '9' < s[i] {
```
bradfitz, line 99: I think the normal Go style is the other way around, with the boundaries on the outside:

```go
   for _, b := range s[:4] {
         if '0' <= b && b <= '9' {
              continue
         }
         return false
    }
```

bradfitz (approved):

chirp/chirp.go
maisem: Done

chirp/chirp.go
maisem: Done

PR #3888, opened 2022-02-07
tailcfg, hostinfo: put envtype in Hostinfo

crawshaw:
Maybe this is unnecessary? Is this information somewhere else?


bradfitz:
This is already included in an "env=" attribute in `hostinfo_linux.go` which currently covers all interesting cases (all env types are currently Linux or a Linux VM on something else)

crawshaw:
@maisem points out that EnvType is in OSVersion on linux. should we do that for all OSs? (because it doesn't seem to be in there for windows, but maybe that's fine for now.)

crawshaw:
I'll drop this for now.

Issue #3761, opened 2022-01-19
Synology Outbound Connections not Persistent After Reboot (DSM 7 / 


rncole:
### What is the issue?

After following the instructions at [https://tailscale.com/kb/1152/synology-outbound/](https://tailscale.com/kb/1152/synology-outbound/), outbound connections are not persistent after restarting the synology device.  Re-running the setcap (below) command and restarting the Tailscale service are required to re-enable.

setcap command:
`sudo setcap cap_net_admin+eip /var/packages/Tailscale/target/bin/tailscaled`

### Steps to reproduce

Restarted Synology

### OS

Synology

### OS version

Synology DSM 7 (DSM 7.0.1-42218)

### Tailscale version

1.20.1-2017

bradfitz:
Huh, weird. I thought that was persistent and was previously working. Does DSM proactively remove that bit from disk at boot or does it never make it to to disk? Is that in-memory only?

DSM7 has systemd. We should probably make our instructions be about installing a start-up unit that fixes things. Perhaps just making the unit be `/var/packages/Tailscale/target/bin/tailscaled synoinit` or something so we can add new startup items over time.

cc @maisem, @danderson, @Xe

rncole:
I did some more testing and it doesn't appear to be the cap.  **Apologies for mis-identifying the initial problem.**  The actual problem is that the `tun` device permissions are getting reset on reboot on DSM7.

Background: I use Tailscale for a Synology-Synology sync where one is behind a T-Mobile CGN rendering most other VPN solutions non-starters.

Here's the getcap output while working:
> $ sudo getcap /var/packages/Tailscale/target/bin/tailscaled
> /var/packages/Tailscale/target/bin/tailscaled = cap_net_admin+eip

After a reboot, here's the same:
> $ sudo getcap /var/packages/Tailscale/target/bin/tailscaled
> /var/packages/Tailscale/target/bin/tailscaled = cap_net_admin+eip

So that part is persistent, however, outbound connections were now no longer functioning.  Re-setting the cap with setcap and restarting Tailscale alone still did not allow outbound.

I re-ran the rest of the commands (placed into a shell script):
> tailscale-outbound-enable.sh 
> mknod: /dev/net/tun: File exists

At this point (now operating correctly) the following is the status as expected:
>/dev$ ls -lah | grep net
>drwxr-xr-x  2 root root        60 Jan 19 11:10 net

> /dev$ ls -lah net
[...]

PR #3890, opened 2022-02-07
cmd/tailscale: add "tailscale configure-host" to prep a Synology machine at boot

bradfitz:
Updates #3761


bradfitz:
I'm flexible on the `on-boot` name. Ideas welcome.

cmd/tailscaled/onboot.go
```diff
@@ -0,0 +1,44 @@
...
+		return errors.New("unsupported arguments")
+	}
+	if runtime.GOOS != "linux" || distro.Get() != distro.Synology {
+		return errors.New("only implemented on Synology")
+	}
+	if uid := os.Getuid(); uid != 0 {
+		return fmt.Errorf("must be run as root, not %q (%v)", os.Getenv("USER"), uid)
+	}
+	if _, err := os.Stat("/dev/net/tun"); os.IsNotExist(err) {
+		if err := os.MkdirAll("/dev/net", 0755); err != nil {
```
maisem, line 30: users have reported that the permissions can be wrong, so do the `chmod` outside of the `IsNotExist` check

maisem (approved):

maisem:
> I'm flexible on the `on-boot` name. Ideas welcome.

`tailscale configure-system`?

cmd/tailscaled/onboot.go
bradfitz: done

bradfitz:
PTAL

cmd/tailscale/cli/configure-host.go
```diff
@@ -0,0 +1,78 @@
...
+	if runtime.GOOS != "linux" || distro.Get() != distro.Synology {
+		return errors.New("only implemented on Synology")
+	}
+	if uid := os.Getuid(); uid != 0 {
+		return fmt.Errorf("must be run as root, not %q (%v)", os.Getenv("USER"), uid)
+	}
+	osVer := hostinfo.GetOSVersion()
+	if !strings.HasPrefix(osVer, "Synology 7") {
+		return fmt.Errorf("unsupported Synology version %q", osVer)
+	}
```
maisem, line 54: instead of erroring out, let's just make sure the tun device exists and has the correct permissions, we can skip the `setcap` bit on DSM6.

And instead of checking for `Synology 7`, check for `Synology 6` so that we are more future proof.

bradfitz: well there's also Synology 5. And who knows what Synology 8 will be like. I'd kinda like to shut down and do nothing by default to be safe until we know what we're dealing with.

bradfitz: PTAL

maisem (approved):

bradfitz:
Tested that it works on my DSM7 after deleting tun device & rebooting the DSM. Tailscale came up in TUN mode.

PR #3902, opened 2022-02-10
util/winutil: migrate corp's winutil into OSS.

dblohm7:
It makes the most sense to have all our utility functions reside in one place.
There was nothing in corp that could not reasonably live in OSS.

I also updated `StartProcessAsChild` to no longer depend on `futureexec`,
thus reducing the amount of code that needed migration. I tested this change
with `tswin` and it is working correctly.

I have a follow-up PR to remove the corresponding code from corp.

### DIFF
--- a/util/winutil/winutil_windows.go
+++ b/util/winutil/winutil_windows.go
@@ -5,7 +5,11 @@
 package winutil
 
 import (
+	"errors"
+	"fmt"
 	"log"
+	"os/exec"
+	"runtime"
 	"syscall"
 
 	"golang.org/x/sys/windows"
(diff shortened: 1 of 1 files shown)


util/winutil/winutil_windows.go
```diff
@@ -18,15 +22,21 @@ const (
...
 	if hwnd == 0 {
-		return 0, false
+		err = errors.New("no Shell process is present")
+		return
 	}
+
 	windows.GetWindowThreadProcessId(hwnd, &pid)
-	return pid, pid != 0
+	if pid == 0 {
+		err = fmt.Errorf("invalid PID for HWND %v", hwnd)
```
bradfitz, line 36: Get into the habit of avoid a "naked return" unless your func is like 2 lines.

return 0, fmt.Errorf(...)
...

return pid, nil

util/winutil/winutil_windows.go
```diff
@@ -18,15 +22,21 @@ const (
...
 // GetDesktopPID searches the PID of the process that's running the
-// currently active desktop and whether it was found.
+// currently active desktop or error.
 // Usually the PID will be for explorer.exe.
-func GetDesktopPID() (pid uint32, ok bool) {
+func GetDesktopPID() (pid uint32, err error) {
 	hwnd := windows.GetShellWindow()
 	if hwnd == 0 {
-		return 0, false
+		err = errors.New("no Shell process is present")
```
bradfitz, line 30: add global `var ErrNoShell = ...` and then:

return 0, ErrNoShell

... so this doesn't allocate in the two common cases?

Then you can document it returns ErrNoShell when it's missing.

But OTOH what code cares about the distinction of ok bool vs error anyway?

util/winutil/winutil_windows.go
```diff
@@ -130,3 +140,114 @@ func isSIDValidPrincipal(uid string) bool {
...
+	}
+	tp.PrivilegeCount = 1
+	tp.Privileges[0].Attributes = windows.SE_PRIVILEGE_ENABLED
+	return windows.AdjustTokenPrivileges(t, false, &tp, 0, nil, nil)
+}
+
+// StartProcessAsChild starts exePath process as a child of parentPID.
+// StartProcessAsChild copies parentPID's environment variables into
+// the new process, along with any optional environment variables in extraEnv.
+func StartProcessAsChild(parentPID uint32, exePath string, extraEnv []string) error {
```
bradfitz, line 173: for archaeology reasons, in the commit message can you give Alex Brainman (github `@alexbrainman`) attribution for this original code, and reference the corp git hash this is coming from?

bradfitz (approved):

PR #3904, opened 2022-02-10
tailcfg: add start of SSH policy to be sent from control plane to nodes

bradfitz:
Updates #3802

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -1157,6 +1157,10 @@ type MapResponse struct {
 	// sees.
 	Health []string `json:",omitempty"`
 
+	// SSHPolicy, if non-nil, updates the SSH policy for how incoming
+	// SSH connections should be handled.
+	SSHPolicy *SSHPolicy `json:",omitempty"`
+
 	// Debug is normally nil, except for when the control server
 	// is setting debug settings on a node.
 	Debug *Debug `json:",omitempty"`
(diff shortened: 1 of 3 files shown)


tailcfg/tailcfg.go
```diff
@@ -1369,3 +1373,59 @@ type SetDNSRequest struct {
...
+	// match. If it does, the map entry's value is the actual
+	// user that's logged in.
+	SSHUser map[string]string `json:"sshUser"`
+
+	// Action is the outcome to task.
+	// A nil or invalid action means to deny.
+	Action *SSHAction `json:"action"`
+}
+
+// SSHPrincipal is either a particular node or a user on any node.
```
maisem, line 1405: What about tags? Do you intend on solving that by using `StableNodeID`?

tailcfg/tailcfg.go
```diff
@@ -1369,3 +1373,59 @@ type SetDNSRequest struct {
...
+// An SSH rule is a match predicate and associated action for an incoming SSH connection.
+type SSHRule struct {
+	// Principal matches an incoming connection. If the connection
+	// matches anything in this list and also matches SSHUser,
+	// then Action is applied.
+	Principal []*SSHPrincipal `json:"principal"`
+
+	// SSHUser are the SSH users that this rule matches.
+	// It is a map from either ssh-user|"*" => local-user.
+	// The map must contain a key for either ssh-user or "*" to
```
maisem, line 1395: what if both match? I assume the `ssh-user` match takes precedence? Add comment?

tailcfg/tailcfg.go
```diff
@@ -1369,3 +1373,59 @@ type SetDNSRequest struct {
...
+	User UserID       `json:"user,omitempty"`
+}
+
+// SSHAction is how to handle an incoming connection.
+// At most one field should be non-zero.
+type SSHAction struct {
+	// Message, if non-empty, is shown to the user before the
+	// action occurs.
+	Message string
+
```
maisem, line 1418: This deviates from the no deny policy in the ACLs, and as you mention elsewhere makes the ordering of rules matter. Can we simplify it and only have Accepts?

tailcfg/tailcfg.go
```diff
@@ -1157,6 +1157,10 @@ type MapResponse struct {
 	// sees.
 	Health []string `json:",omitempty"`
 
+	// SSHPolicy, if non-nil, updates the SSH policy for how incoming
```
maisem, line 1160: update `CurrentMapRequestVersion`?

tailcfg/tailcfg.go
```diff
@@ -1369,3 +1373,59 @@ type SetDNSRequest struct {
 	// Value is the value to add.
 	Value string
 }
+
+// SSHPolicy is the policy for how to handle incoming SSH connections
+// over Tailscale.
+type SSHPolicy struct {
+	// Rules are the rules to process for an incoming SSH
+	// connection. The first matching rule takes its action and
```
mayakacz, line 1381: Does this mean that if I have multiple rules, they're evaluated in order? I guess that makes sense. If we're allowing you to specify multiple SSH users though, and want to default to the non-root one, but you list them in a different order, does that mean we might use the root one instead?

tailcfg/tailcfg.go
```diff
@@ -1369,3 +1373,59 @@ type SetDNSRequest struct {
...
+
+// An SSH rule is a match predicate and associated action for an incoming SSH connection.
+type SSHRule struct {
+	// Principal matches an incoming connection. If the connection
+	// matches anything in this list and also matches SSHUser,
+	// then Action is applied.
+	Principal []*SSHPrincipal `json:"principal"`
+
+	// SSHUser are the SSH users that this rule matches.
+	// It is a map from either ssh-user|"*" => local-user.
```
mayakacz, line 1394: I think this is saying we'll pass through the user that is authenticated in Tailscale? If so, yup.

tailcfg/tailcfg.go
bradfitz: Yeah, but keep in mind that this SSHPolicy is not the user facing bits at all. This is the internal representation that tailscaled can handle, analogous to how ACL JSON compiles down to the internal PacketFilter/FilterRule:

https://pkg.go.dev/tailscale.com/tailcfg#MapResponse.PacketFilter
https://pkg.go.dev/tailscale.com/tailcfg#FilterRule

So this is just setting up tailscaled to be able to do configurable sets of things from control where we can implement higher level policies and UIs that map down to this.

tailcfg/tailcfg.go
bradfitz: Added a comment about why it's needed.

tailcfg/tailcfg.go
bradfitz: done

tailcfg/tailcfg.go
bradfitz: Yes, we can send the whole set down.

tailcfg/tailcfg.go
bradfitz: Not yet. Once it's wired up to anything useful I can.

bradfitz:
PTAL; ready to be submitted.

Still not wired up to anything, though. That'll be the next commit.

maisem (approved):

tailcfg/tailcfg.go
```diff
@@ -1157,6 +1157,10 @@ type MapResponse struct {
 	// sees.
 	Health []string `json:",omitempty"`
 
+	// SSHPolicy, if non-nil, updates the SSH policy for how incoming
+	// SSH connections should be handled.
+	SSHPolicy *SSHPolicy
```
crawshaw, line 1162: omitempty?

tailcfg/tailcfg.go
```diff
@@ -1369,3 +1373,82 @@ type SetDNSRequest struct {
...
 	Value string
 }
+
+// SSHPolicy is the policy for how to handle incoming SSH connections
+// over Tailscale.
+type SSHPolicy struct {
+	// Rules are the rules to process for an incoming SSH
+	// connection. The first matching rule takes its action and
+	// stops processing further rules.
+	Rules []*SSHRule
```
crawshaw, line 1383: given you're naming fields inside SSHRule, name this "rules"?

tailcfg/tailcfg.go
```diff
@@ -1369,3 +1373,82 @@ type SetDNSRequest struct {
...
+	// Action is the outcome to task.
+	// A nil or invalid action means to deny.
+	Action *SSHAction `json:"action"`
+}
+
+// SSHPrincipal is either a particular node or a user on any node.
+// At most one field should be non-zero specified.
+type SSHPrincipal struct {
+	Node StableNodeID `json:"node,omitempty"`
+	User UserID       `json:"user,omitempty"`
```
crawshaw, line 1421: We have a StableUserID on the server, I think we should add it to tailcfg and use it here. I'd rather not introduce a use for the old UserID here.

tailcfg/tailcfg.go
```diff
@@ -1369,3 +1373,82 @@ type SetDNSRequest struct {
...
+	// at least during beta testing period, when people can
+	// temporarily write their own SSHPolicy JSON file.
+}
+
+// SSHAction is how to handle an incoming connection.
+// At most one field should be non-zero.
+type SSHAction struct {
+	// Message, if non-empty, is shown to the user before the
+	// action occurs.
+	Message string
```
crawshaw, line 1433: Message and Reject don't have custom json names

crawshaw (approved):

tailcfg/tailcfg.go
bradfitz: done

tailcfg/tailcfg.go
bradfitz: done

tailcfg/tailcfg.go
bradfitz: adjusted

tailcfg/tailcfg.go
bradfitz: done

PR #3917, opened 2022-02-12
control/controlclient: select newer certificate

sailorfrag:
If multiple certificates match when selecting a certificate, use the one
issued the most recently (as determined by the NotBefore timestamp).
This also adds some tests for the function that performs that
comparison.

Updates tailscale/coral#6

### DIFF
--- a/control/controlclient/sign_supported.go
+++ b/control/controlclient/sign_supported.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/tailscale/certstore"
 	"tailscale.com/tailcfg"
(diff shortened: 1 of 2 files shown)


control/controlclient/sign_supported.go
```diff
@@ -73,23 +74,46 @@ func isSubjectInChain(subject string, chain []*x509.Certificate) bool {
 	return false
 }
 
-func selectIdentityFromSlice(subject string, ids []certstore.Identity) (certstore.Identity, []*x509.Certificate) {
+func selectIdentityFromSlice(subject string, ids []certstore.Identity, now time.Time) (certstore.Identity, []*x509.Certificate) {
+	bestCandidate := struct {
```
bradfitz, line 78: Less syntaxy:

```go
var bestCandidate struct {
    ...
}
```

control/controlclient/sign_supported_test.go
```diff
@@ -0,0 +1,238 @@
+// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
```
bradfitz, line 1: wrong year

bradfitz (approved):

control/controlclient/sign_supported.go
```diff
@@ -73,23 +74,46 @@ func isSubjectInChain(subject string, chain []*x509.Certificate) bool {
...
+		if len(bestCandidate.chain) > 0 && bestCandidate.chain[0].NotBefore.After(chain[0].NotBefore) {
+			continue
+		}
+
+		bestCandidate.id = id
+		bestCandidate.chain = chain
 	}
 
-	return nil, nil
+	return bestCandidate.id, bestCandidate.chain
```
DentonGentry, line 116: If no suitable certificate is found, will bestCandidate.id be nil from the initialization on line 79?

DentonGentry (approved):

control/controlclient/sign_supported.go
sailorfrag: Yes, they'll be `nil, nil`

sailorfrag: Also, there is a test case specifically covering this and it verifies that `nil` is returned.

Issue #1548, opened 2021-03-19
cmd/tailscaled: tracking bug for ongoing log leveling work


bradfitz:
We introduced log levels and fixed the bulk of the log spam issues in #282 but work remains.

This is an ongoing tracking bug for future log leveling work.

And for tailscaled flags to control log output/uploading.

luismanson:
Hello, is this active? I have been poking with syslog on a Raspberry Pi and noticed Tailscale logging being too verbose.
Is some form of logging control comming?

DentonGentry:
Substantial work on reducing log verbosity has already been done, though there is more that can be done and would be tracked here. There is not currently a large effort to further reduce overall logging, though there are a couple places where the code gets into a state where it logs stuff too often. Those are being resolved by fixing the state it can get into.

josharian:
@luismanson are there particular log entries that are of concern, or is this a general "there's too much"?

luismanson:
Thanks for your quick reply @DentonGentry 

@josharian  thanks for your interest,  in my case I've seen my 7Mb 'daemon.log" is like >30% filled with tailscale related messaged, most of them are some variation of:
`"tailscaled[423]: Accept: TCP{123.123.193.26:45542 > 123.123.159.56:22} 104 tcp non-syn`

Some quick numbers:
$ grep -i tailscale daemon.log|grep -v Accept|wc -l
    4340
$ grep -i tailscale daemon.log|grep Accept|wc -l
    16118

dijit:
Tailscale logging is indeed rather verbose; in fact it's the kind of output I would expect if I ran it with the `--verbose` flag, since every connection, drop, health status and so on is a log line.

sample from my laptop (note: I didn't have to remove any lines from any _other_ service on my laptop)

```May 21 16:40:39 trace.rundeln.drk.sc tailscaled[438]: health("overall"): error: not in map poll
May 21 16:40:40 trace.rundeln.drk.sc tailscaled[438]: [RATELIMIT] format("control: mapRoutine: %s") (2 dropped)
May 21 16:40:40 trace.rundeln.drk.sc tailscaled[438]: control: mapRoutine: state:authenticated
May 21 16:40:40 trace.rundeln.drk.sc tailscaled[438]: trying bootstrapDNS("derp7.tailscale.com", "167.179.89.145") for "login.tailscale.com" ...
[...]

PR #3918, opened 2022-02-12
all: adjust some log verbosity

bradfitz:
Updates #1548


control/controlclient/direct.go
```diff
@@ -214,7 +214,7 @@ func (c *Direct) SetHostinfo(hi *tailcfg.Hostinfo) bool {
 	}
 	c.hostinfo = hi.Clone()
 	j, _ := json.Marshal(c.hostinfo)
-	c.logf("HostInfo: %s", j)
+	c.logf("[v1] HostInfo: %s", j)
```
DentonGentry, line 217: Isn't this one something we use?

control/controlclient/direct.go
```diff
@@ -467,9 +467,9 @@ func (c *Direct) doLogin(ctx context.Context, opt loginOpt) (mustRegen bool, new
 	//	- user is disabled
 
 	if resp.AuthURL != "" {
-		c.logf("AuthURL is %v", resp.AuthURL)
+		c.logf("[v1] AuthURL is %v", resp.AuthURL)
```
DentonGentry, line 470: This is the log message which prints https://login.tailscale.com/a/0123456789 ? We do have people who have used this, to see the login URL in cloud instance logs for example so they can authenticate the system even without having SSH access to see the output of tailscale up.

DentonGentry (approved):

control/controlclient/direct.go
bradfitz: (resolved on chat)

control/controlclient/direct.go
bradfitz: Reverted this line.

PR #3943, opened 2022-02-15
tsweb: add DebugHandler.FlagSet

josharian:
For runtime-modifiable values.


josharian:
Demo of my beautiful UI:

https://user-images.githubusercontent.com/67496/153975741-cfd0a9c1-d441-46e4-88af-b5a6b122c304.mov

tsweb/debug.go
```diff
@@ -139,3 +148,161 @@ func gcHandler(w http.ResponseWriter, r *http.Request) {
...
+	return all
+}
+
+func (d *DebugHandler) handleFlags(w http.ResponseWriter, r *http.Request) {
+	d.flagmu.Lock()
+	defer d.flagmu.Unlock()
+
+	var userError string
+	var modified string
+	if r.Method == http.MethodPost {
```
bradfitz, line 198: I prefer "POST" but won't fight the constant if you prefer.

tsweb/debug.go
```diff
@@ -139,3 +148,161 @@ func gcHandler(w http.ResponseWriter, r *http.Request) {
...
+// Sample usage:
+//      flagset, flagc := debug.FlagSet()
+//      flagset.Int("max", 0, "maximum number of bars")
+//      flagset.String("s", "qux", "default name for new foos")
+//      go func() {
+//          for change := range flagc {
+//              // TODO: handle change, which will contain values for keys "max" and/or "s"
+//          }
+//      }()
+func (d *DebugHandler) FlagSet() (fs *flag.FlagSet, c chan map[string]interface{}) {
```
bradfitz, line 167: `<-chan`

tsweb/debug.go
```diff
@@ -139,3 +148,161 @@ func gcHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	dot := &struct {
+		Error    string
+		Modified string
+		Flags    []copiedFlag
+	}{
+		Error:    userError,
+		Modified: modified,
+		Flags:    copyFlags(d.flagset),
```
dsnet, line 271: What's the fidelity of the round-trip stringification and flag parsing? That is, we don't have confidence whether `flag.Value.String` will be parsed by `flag.Value.Set` to be identical to `flag.Value.` I think we should have a sanity check that `value.Set(value.String()).String() == value.String()` before proceeding here.

I'm concerned that trying to change flag X accidentally changes flag Y because of the lack of round-trip fidelity.

dsnet: A test like this also guarantees that the roll-back logic on L236 would succeed.

dsnet:
Another thought: should flags be persistent by default? What happens if make runtime changes and forget that we made them, but those changes happen to be critical to the operation of the service. Restarting the service will wipe out the flags and potentially cause the service to not come back up.

tsweb/debug.go
josharian: The thought of a non-round-tripping flag.Value makes me shudder in horror. But sure, will add a check.

josharian:
> should flags be persistent

My gut reaction: If you want persistent flags, you should change them in the systemd service file. This is for tinkering, for emergency recovery, etc. Plus with persistence you need to figure out a default storage location and serialization, figure out error handling paths, etc. Plus then you have a non-code config, which means migrating to a new machine will cause mysterious outages.

dsnet:
I'm okay with proceeding with non-persistent flags, but will voice that I feel uncomfortable that they may become load bearing in unfortunate ways.

josharian:
OK. I'll add another bit of warning text on the page.

FWIW, I think it unlikely they'll become load bearing, exactly because they are not persistent.

josharian:
It looks like I am unlikely to finish this. If anyone feels inspired to revive it, feel free to. :)

PR #3944, opened 2022-02-15
net/dns/resolvconffile: unify three /etc/resolv.conf parsers into new package

bradfitz:
(Diff looks long because a table of tests was effectively duplicated.)


DentonGentry (approved):
I looked over the direct.go handling in dealing with rename(), because that is tricky in some environments like gitpod where /etc/resolv.conf is a bindmount. This should be fine, it is only moving the code which writes to the in-memory buffer.

bradfitz:
> I looked over the direct.go handling in dealing with rename(), because that is tricky in some environments like gitpod where /etc/resolv.conf is a bindmount. This should be fine, it is only moving the code which writes to the in-memory buffer.

Yeah, I stayed away from anything tricky like that.

For now I just want the three unified so we can fix/tighten the parsing in one place. Currently all three behaved differently (yay)

PR #3958, opened 2022-02-17
cmd/tailscale: add --auth-key flag as a synonym for --authkey

josharian:
That way humans don't have to remember which is correct.

The help text renders like this:

  --auth-key string
    	synonym for authkey
  --authkey string
    	node authorization key; if it begins with "file:", then it's a path to a file containing the authkey


bradfitz (commented):
How about we only have one flag, whichever we think is best, probably `--auth-key` and rewrite os.Args before flag.Parse to convert the old-style one?

josharian:
> How about we only have one flag, whichever we think is best, probably --auth-key and rewrite os.Args before flag.Parse to convert the old-style one?

Done.

cmd/tailscale/cli/cli_test.go
```diff
@@ -1023,3 +1026,27 @@ func TestExitNodeIPOfArg(t *testing.T) {
...
+		{in: []string{"a", "-authkey=3", "b"}, want: []string{"a", "--auth-key=3", "b"}},
+		{in: []string{"a", "-auth-key=4", "b"}, want: []string{"a", "-auth-key=4", "b"}},
+		{in: []string{"a", "--authkey", "5", "b"}, want: []string{"a", "--auth-key", "5", "b"}},
+		{in: []string{"a", "-authkey", "6", "b"}, want: []string{"a", "--auth-key", "6", "b"}},
+		{in: []string{"a", "authkey", "7", "b"}, want: []string{"a", "authkey", "7", "b"}},
+	}
+
+	for _, tt := range tests {
+		got := CleanUpArgs(tt.in)
+		c.Assert(got, qt.DeepEquals, tt.want)
```
bradfitz, line 1050: when this fails, it never includes the `tt.in` context. I guess you're relying on all the want values having distinct numbers in them, eh?

cmd/tailscale/cli/cli.go
```diff
@@ -104,6 +104,23 @@ func newFlagSet(name string) *flag.FlagSet {
...
 }
 
+// CleanUpArgs rewrites command line arguments for simplicity and backwards compatibility.
+// In particular, it rewrites --authkey to --auth-key.
+func CleanUpArgs(args []string) []string {
+	out := make([]string, 0, len(args))
+	for _, arg := range args {
+		// Rewrite --authkey to --auth-key, and --authkey=x to --auth-key=x,
+		// and the same for the -authkey variant.
+		if strings.HasPrefix(arg, "--authkey") || strings.HasPrefix(arg, "-authkey") {
```
bradfitz, line 114: technically this should match `/^-{1,2}authkey\b/` and this is missing the `\b` part; it matches `--authkeyfoo` still.

Probably not a problem in practice. At least yet. :)

bradfitz (approved):

bradfitz:
Add a RELNOTE= line to commit message to help us write release notes later?

mayakacz:
Updates https://github.com/tailscale/corp/issues/3486

cmd/tailscale/cli/cli.go
josharian: I think the only relevant `\b` here is `=`. (This is argv, not shell, so spaces don't get any magic treatment.) Fixed.

PR #3963, opened 2022-02-17
ipnlocal, tailssh: start moving host key stuff into the right spot

bradfitz:
Make tailssh ask LocalBackend for the SSH hostkeys, as we'll need to
distribute them to peers.

For now only the hacky use-same-as-actual-host mode is implemented.

Updates #3802

### DIFF
--- a/ipn/ipnlocal/ssh.go
+++ b/ipn/ipnlocal/ssh.go
@@ -1,0 +1,48 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build linux
+// +build linux
+
+package ipnlocal
+
+import (
+	"errors"
+	"io/ioutil"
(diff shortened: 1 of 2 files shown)


maisem (approved):

ipn/ipnlocal/ssh.go
```diff
@@ -0,0 +1,48 @@
...
+		if os.IsNotExist(err) {
+			continue
+		}
+		if err != nil {
+			return nil, err
+		}
+		signer, err := ssh.ParsePrivateKey(hostKey)
+		if err != nil {
+			return nil, err
+		}
```
maisem, line 41: cache them?

bradfitz: eh, how often does a machine get an incoming SSH connection? not thousands of times per second.

maisem: I was more speaking in terms of the callers of the LocalBackend, not sure how frequently they'll be fetching this or over PeerAPI?

PR #3966, opened 2022-02-17
tailcfg, ipn/ipnlocal: add Hostinfo.SSH_HostKeys, send when SSH enabled

bradfitz:
Updates #3802


maisem (approved):
Change LGTM, pending whatever name we end up on.

Issue #2057, opened 2021-06-06
Taildrop: "tailscale get" does not preserve timestamps


apenwarr:
Steps:
- Taildrop a bunch of files to my linux box over several days, letting them accumulate in the incoming folder
- `tailscale get .`

Expected:
- The files can be sorted in order of arrival, in Chrome's file browser

Actual:
- They all have exactly the same timestamp, so I can't tell which file was the most recent.

It might not actually be desirable to preserve the actual time the file was originally sent, which could have been months ago, causing some confusion in the output folder. On the other hand, it would be really nice to at least have some fake timestamps that are in chronological order of the times the files were received.

wizzard0:
I believe the common practice is to preserve the 'last-modified' date from the original file while bumping 'created' to the time of arrival/sending

DentonGentry:
Still an issue with 1.36:
```
root@dgentry:~# stat ./testfile
  File: ./testfile
  Size: 6935828   	Blocks: 13552      IO Block: 4096   regular file
Device: fc01h/64513d	Inode: 96          Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2023-02-12 15:54:19.219412796 +0000
Modify: 2023-02-12 15:54:19.243413055 +0000
Change: 2023-02-12 15:54:19.243413055 +0000
 Birth: 2023-02-12 15:54:19.219412796 +0000
```

yojimble:
Still an issue, using 1.80.3 all transferred photos have lost their creation dates which makes sorting difficult

Issue #3548, opened 2021-12-10
TailDrop should default file overwrite as optional param


gamefreak613:
Feature request: I'd like to automatically overwrite files sent via TailDrop if one exists already at the destination. Right now (at least in windows) it renames the newer file, appending a number in the filename, keeping both files. Perhaps as part of the send, it could be an option to either overwrite or keep existing files?

DentonGentry:
Since this bug was filed, `tailscale file` added a `--conflict=(skip|overwrite|rename)` argument. It defaults to skip, but can be set to `--conflict=overwrite`

https://github.com/tailscale/tailscale/commit/5be42c0af1c9c8899e8383b19d7284a058005bd0

error-try-again:
I believe the overwrite flag is bugged.

When executing:

```
ExecStart=tailscale file get --verbose --loop --conflict=overwrite "/home/void/taildrops"
```

I hit the following error when the subsequent case is called

```
void on null in ~/.config/systemd
$ systemctl --user status tailreceive
● tailreceive.service - File Receiver Service for Taildrop
     Loaded: loaded (/home/void/.config/systemd/user/tailreceive.service; enabled; vendor preset: enabled)
     Active: active (running) since Fri 2024-02-16 23:09:11 AEST; 12min ago
   Main PID: 2873436 (tailscale)
      Tasks: 5 (limit: 38263)
     Memory: 3.8M
        CPU: 42ms
     CGroup: /user.slice/user-1000.slice/user@1000.service/app.slice/tailreceive.service
             └─2873436 tailscale file get --verbose --loop /home/void/Downloads

Feb 16 23:20:20 null tailscale[2873436]: refusing to overwrite file: open /home/void/Downloads/setup-script.sh: file exists
Feb 16 23:20:20 null tailscale[2873436]: moved 0/1 files
```

```
case skipOnExist:
if _, statErr := os.Stat(targetFile); statErr == nil {
	// we can stat a file at that path: so it already exists.
	return nil, fmt.Errorf("refusing to overwrite file: %w", err)
}
```
in the `openFileOrSubstitute` function

`setup-script.sh` is an arbitrary file I'm sharing from a macbook.

Issue #2324, opened 2021-07-04
taildrop: receive files on linux needs sudo


mcristina422:
The [docs](https://tailscale.com/kb/1106/taildrop/) currently make it clear that `sudo tailscale file get .` is the proper way to receive files on linux. Although the error presented when missing `sudo` doesn't really make that apparent to the user.

```
$ tailscale file get .
getting WaitingFiles: HTTP 403 Forbidden: file access denied
 (expected 200)
```

related to https://github.com/tailscale/tailscale/issues/2312, where this UX confusion will get improved 
version: 1.10

DentonGentry:
I think #2312 is sufficient to track the issue, closing as a dup.

PR #3967, opened 2022-02-18
cmd/tailscale: add file get options for dealing with existing files

davideger:
This addresses #3548 for the tailscale users who use 'tailscale file get' to receive Taildrops.

### DIFF
--- a/cmd/tailscale/cli/file.go
+++ b/cmd/tailscale/cli/file.go
@@ -15,6 +15,7 @@ import (
 	"mime"
 	"net/http"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 	"time"
@@ -286,22 +287,116 @@ func runCpTargets(ctx context.Context, args []string) error {
 	return nil
 }
 
+// onConflict is a flag.Value for the --conflict flag's three string options.
+type onConflict string
+
+const (
+	skipOnExist         onConflict = "skip"
+	overwriteExisting   onConflict = "overwrite" //  Overwrite any existing file at the target location
+	createNumberedFiles onConflict = "rename"    //  Create an alternately named file in the style of Chrome Downloads
+)
+
+func (v *onConflict) String() string { return string(*v) }
+
+func (v *onConflict) Set(s string) error {
+	if s == "" {
+		*v = skipOnExist
+		return nil
+	}
+	*v = onConflict(strings.ToLower(s))
+	if *v != skipOnExist && *v != overwriteExisting && *v != createNumberedFiles {
+		return fmt.Errorf("%q is not one of (skip|overwrite|rename)", s)
+	}
+	return nil
+}
+
 var fileGetCmd = &ffcli.Command{
 	Name:       "get",
-	ShortUsage: "file get [--wait] [--verbose] <target-directory>",
+	ShortUsage: "file get [--wait] [--verbose] [--conflict=(skip|overwrite|rename)] <target-directory>",
 	ShortHelp:  "Move files out of the Tailscale file inbox",
 	Exec:       runFileGet,
 	FlagSet: (func() *flag.FlagSet {
 		fs := newFlagSet("get")
 		fs.BoolVar(&getArgs.wait, "wait", false, "wait for a file to arrive if inbox is empty")
 		fs.BoolVar(&getArgs.verbose, "verbose", false, "verbose output")
+		fs.Var(&getArgs.conflict, "conflict", `behavior when a conflicting (same-named) file already exists in the target directory.
+	skip:       skip conflicting files: leave them in the taildrop inbox and print an error. get any non-conflicting files
+	overwrite:  overwrite existing file
+	rename:     write to a new number-suffixed filename`)
 		return fs
 	})(),
 }
 
-var getArgs struct {
-	wait    bool
-	verbose bool
+var getArgs = struct {
+	wait     bool
+	verbose  bool
+	conflict onConflict
+}{conflict: skipOnExist}
+
+func numberedFileName(dir, name string, i int) string {
+	ext := path.Ext(name)
+	return filepath.Join(dir, fmt.Sprintf("%s (%d)%s",
+		strings.TrimSuffix(name, ext),
+		i, ext))
+}
+
+func openFileOrSubstitute(dir, base string, action onConflict) (*os.File, error) {
+	targetFile := filepath.Join(dir, base)
+	f, err := os.OpenFile(targetFile, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0644)
+	if err == nil {
+		return f, nil
+	}
+	// Something went wrong trying to open targetFile as a new file for writing.
+	switch action {
+	default:
+		// This should not happen.
+		return nil, fmt.Errorf("file issue. how to resolve this conflict? no one knows.")
+	case skipOnExist:
+		if _, statErr := os.Stat(targetFile); statErr == nil {
+			// we can stat a file at that path: so it already exists.
+			return nil, fmt.Errorf("refusing to overwrite file: %w", err)
+		}
+		return nil, fmt.Errorf("failed to write; %w", err)
+	case overwriteExisting:
+		// remove the target file and create it anew so we don't fall for an 
+		// attacker who symlinks a known target name to a file he wants changed.
+		if err = os.Remove(targetFile); err != nil {
+			return nil, fmt.Errorf("unable to remove target file: %w", err)
+		}
+		if f, err = os.OpenFile(targetFile, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0644); err != nil {
+			return nil, fmt.Errorf("unable to overwrite: %w", err)
+		}
+		return f, nil
+	case createNumberedFiles:
+		// It's possible the target directory or filesystem isn't writable by us,
+		// not just that the target file(s) already exists.  For now, give up after
+		// a limited number of attempts.  In future, maybe distinguish this case
+		// and follow in the style of https://tinyurl.com/chromium100
+		maxAttempts := 100
+		for i := 1; i < maxAttempts; i++ {
+			if f, err = os.OpenFile(numberedFileName(dir, base, i), os.O_RDWR|os.O_CREATE|os.O_EXCL, 0644); err == nil {
+				return f, nil
+			}
+		}
+		return nil, fmt.Errorf("unable to find a name for writing %v, final attempt: %w", targetFile, err)
+	}
+}
+
+func receiveFile(ctx context.Context, wf apitype.WaitingFile, dir string) (targetFile string, size int64, err error) {
+	rc, size, err := tailscale.GetWaitingFile(ctx, wf.Name)
+	if err != nil {
+		return "", 0, fmt.Errorf("opening inbox file %q: %w", wf.Name, err)
+	}
+	f, err := openFileOrSubstitute(dir, wf.Name, getArgs.conflict)
+	if err != nil {
+		return "", 0, err
+	}
+	_, err = io.Copy(f, rc)
+	rc.Close()
+	if err != nil {
+		return "", 0, fmt.Errorf("failed to write %v: %v", f.Name(), err)
+	}
+	return f.Name(), size, f.Close()
 }
 
 func runFileGet(ctx context.Context, args []string) error {
@@ -330,47 +425,40 @@ func runFileGet(ctx context.Context, args []string) error {
 			break
 		}
 		if getArgs.verbose {
-			log.Printf("waiting for file...")
+			printf("waiting for file...")
 		}
 		if err := waitForFile(ctx); err != nil {
 			return err
 		}
 	}
 
+	var errs []error
 	deleted := 0
 	for _, wf := range wfs {
-		rc, size, err := tailscale.GetWaitingFile(ctx, wf.Name)
+		writtenFile, size, err := receiveFile(ctx, wf, dir)
 		if err != nil {
-			return fmt.Errorf("opening inbox file %q: %v", wf.Name, err)
-		}
-		targetFile := filepath.Join(dir, wf.Name)
-		of, err := os.OpenFile(targetFile, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0644)
-		if err != nil {
-			if _, err := os.Stat(targetFile); err == nil {
-				return fmt.Errorf("refusing to overwrite %v", targetFile)
-			}
-			return err
-		}
-		_, err = io.Copy(of, rc)
-		rc.Close()
-		if err != nil {
-			return fmt.Errorf("failed to write %v: %v", targetFile, err)
-		}
-		if err := of.Close(); err != nil {
-			return err
+			errs = append(errs, err)
+			continue
 		}
 		if getArgs.verbose {
-			log.Printf("wrote %v (%d bytes)", wf.Name, size)
+			printf("wrote %v as %v (%d bytes)\n", wf.Name, writtenFile, size)
 		}
-		if err := tailscale.DeleteWaitingFile(ctx, wf.Name); err != nil {
-			return fmt.Errorf("deleting %q from inbox: %v", wf.Name, err)
+		if err = tailscale.DeleteWaitingFile(ctx, wf.Name); err != nil {
+			errs = append(errs, fmt.Errorf("deleting %q from inbox: %v", wf.Name, err))
+			continue
 		}
 		deleted++
 	}
 	if getArgs.verbose {
-		log.Printf("moved %d files", deleted)
+		printf("moved %d/%d files\n", deleted, len(wfs))
 	}
-	return nil
+	if len(errs) == 0 {
+		return nil
+	}
+	for _, err := range errs[:len(errs)-1] {
+		outln(err)
+	}
+	return errs[len(errs)-1]
 }
 
 func wipeInbox(ctx context.Context) error {


cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,119 @@ func runCpTargets(ctx context.Context, args []string) error {
 	return nil
 }
 
+type existingFileBehavior int
+
+const (
+	FailOnExist         existingFileBehavior = iota
```
bradfitz, line 293: Exported consts of an unexported type is a bit weird in Go.

But also this is just a bit overkill. The strings are fine. Let's ditch the type here and the map on line 298 and the String method, etc.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,119 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+		return fmt.Errorf("'%s' not valid option for existing file behavior", s)
+	}
+	*v = val
+	return nil
+}
+
 var fileGetCmd = &ffcli.Command{
 	Name:       "get",
-	ShortUsage: "file get [--wait] [--verbose] <target-directory>",
+	ShortUsage: "file get [--wait] [--verbose] [--on_existing=(abort|overwrite|autonumber)] <target-directory>",
```
bradfitz, line 328: I think the flag convention we use is hyphens, not underscores.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,119 @@ func runCpTargets(ctx context.Context, args []string) error {
...
 	ShortHelp:  "Move files out of the Tailscale file inbox",
 	Exec:       runFileGet,
 	FlagSet: (func() *flag.FlagSet {
 		fs := newFlagSet("get")
 		fs.BoolVar(&getArgs.wait, "wait", false, "wait for a file to arrive if inbox is empty")
 		fs.BoolVar(&getArgs.verbose, "verbose", false, "verbose output")
+		fs.Var(&getArgs.onExisting, "on_existing", "behavior when target file already exists")
```
bradfitz, line 335: how about "conflict" instead of "on_existing"

 --conflict={abort,overwrite,rename}

?

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,119 @@ func runCpTargets(ctx context.Context, args []string) error {
...
 		return fs
 	})(),
 }
 
-var getArgs struct {
-	wait    bool
-	verbose bool
+type FileGetArgs struct {
```
bradfitz, line 340: why renamed? why exported?

bradfitz (commented):
Please squash into one commit.

Look at git history for examples of how commit messages are formatted.

Thanks!

bradfitz (changes requested):
(See prior comments)

davideger:
Thanks as always for the gentle feedback.  Hopefully this version looks better.

I noticed you started converting IO in the tailscale cli to using a custom wrapper printf (for js/wasm hackery?), so I replaced the log lines I was writing here with that printf.

davideger:
to answer your questions, "why did you... ?"

The answer is simply: "embarrassing golang and git noob errors" so thanks for training me up.

More than you probably wanted to know,  my brain took: ` v := exp ` to be the simple type-elided version of what must be ` v : Type = exp`,  But when I tried:
```
var getArgs : struct { ... } = { onConflict: FailOnExist }
```
it didn't work (golang doesn't actually use : as type separator in this way as OCaml and Python do), and I random walked into the first version I sent which compiled 😂

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,106 @@ func runCpTargets(ctx context.Context, args []string) error {
 	return nil
 }
 
+type onConflict string
+
+const (
+	skipOnExist         onConflict = "skip"
+	overwriteExisting              = "overwrite" //  Overwrite any existing file at the target location
```
bradfitz, line 294: This doesn't do what you think it does: https://go.dev/play/p/587SMC5Rn-K

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,106 @@ func runCpTargets(ctx context.Context, args []string) error {
...
 }
 
-var getArgs struct {
-	wait    bool
-	verbose bool
+var getArgs = struct {
+	wait     bool
+	verbose  bool
+	conflict onConflict
+}{conflict: skipOnExist}
```
bradfitz, line 333: why do you need to initialize this? It will be done by flag.Parse before the code runs.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,106 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+
+func numberedFileName(dir string, name string, i int) string {
```
bradfitz, line 335: Drop one of the `string` types...

`(dir, name string, i int)`

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,106 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+	extension := path.Ext(name)
+	numberedFileName := fmt.Sprintf("%s (%d)%s", name[0:len(name)-len(extension)], i, extension)
```
bradfitz, line 337: drop the 0

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,106 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+}
+
+func receiveFile(ctx context.Context, wf apitype.WaitingFile, dir string) (targetFile string, size int64, err error) {
+	rc, size, err := tailscale.GetWaitingFile(ctx, wf.Name)
+	targetFile = filepath.Join(dir, wf.Name)
+	if err != nil {
+		err = fmt.Errorf("opening inbox file %q: %w", wf.Name, err)
+		return
+	}
+	of, err := os.OpenFile(targetFile, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0644)
```
bradfitz, line 348: I think the rest of this would be simpler if you pulled this part out into its own func, passing in targetFile and getArgs.conflict and returning what OpenFile already returns

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,106 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+	if err != nil {
+		var err2 error = nil
```
bradfitz, line 350: all variables in Go are zero initialized.

and err2 isn't a good variable name.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,106 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+		switch getArgs.conflict {
+		case skipOnExist:
```
bradfitz, line 352: personally I'd add a default case here to catch invalid input and then drop the whole custom type + flag.Var thing.

But I guess you want the flag.Parse to fail early even before there are conflicts, eh? That's fine, but let's still add a default here, even if it's just panics "internal error" or returns an error.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,106 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+			of, err = os.OpenFile(targetFile, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+			if err != nil {
+				err = fmt.Errorf("unable to overwrite %v. %w", targetFile, err)
+				return
+			}
+		case createNumberedFiles:
+			// Why have max attempts?
+			// It's possible the target directory or filesystem isn't writable by us,
+			// not just that the target file name is not writable by us.
+			maxAttempts := 1000
```
bradfitz, line 367: or 100 :)

cmd/tailscale/cli/file.go
davideger: rightie oh. thanks.

cmd/tailscale/cli/file.go
davideger: done

cmd/tailscale/cli/file.go
davideger: done

cmd/tailscale/cli/file.go
davideger: done

cmd/tailscale/cli/file.go
davideger: done

cmd/tailscale/cli/file.go
davideger: done

davideger:
Re-syncing to keep up with head.  Please let me know if there's something more you'd like to see.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,109 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+var getArgs = struct {
+	wait     bool
+	verbose  bool
+	conflict onConflict
+}{conflict: skipOnExist}
+
+func numberedFileName(dir, name string, i int) string {
+	extension := path.Ext(name)
+	numberedFileName := fmt.Sprintf("%s (%d)%s", name[:len(name)-len(extension)], i, extension)
+	return filepath.Join(dir, numberedFileName)
```
bradfitz, line 337: this variable name isn't really pulling its weight. inline into this expression, over multiple lines?

```go
   return filepath.Join(dir, fmt.Sprintf("%s (%d)%s",
            strings.TrimSuffix(name, ext)
            i, ext))
```

?

cmd/tailscale/cli/file.go
bradfitz: This was marked as resolved, but how? Did you see this comment?

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,109 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+}
+
+func (v *onConflict) Set(s string) error {
+	if s == "" {
+		*v = skipOnExist
+		return nil
+	}
+	*v = onConflict(strings.ToLower(s))
+	if *v != skipOnExist && *v != overwriteExisting && *v != createNumberedFiles {
+		return fmt.Errorf("'%s' not one of (skip|overwrite|rename)", s)
```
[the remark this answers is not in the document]
davideger: thanks, done.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,109 @@ func runCpTargets(ctx context.Context, args []string) error {
...
-	wait    bool
-	verbose bool
+var getArgs = struct {
+	wait     bool
+	verbose  bool
+	conflict onConflict
+}{conflict: skipOnExist}
+
+func numberedFileName(dir, name string, i int) string {
+	extension := path.Ext(name)
```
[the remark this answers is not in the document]
davideger: done.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,109 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+			for i := 1; i < maxAttempts; i++ {
+				targetFile = numberedFileName(dir, base, i)
+				if of, err = os.OpenFile(targetFile, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0644); err == nil {
+					return
+				}
+			}
+			err = fmt.Errorf("unable to find a name for writing %v.  original error: %w", targetFile, openError)
+		}
+	}
+	return
```
bradfitz, line 372: sorry, no naked returns in funcs this long. Only when it's a few lines long at most.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,109 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+	conflict onConflict
+}{conflict: skipOnExist}
+
+func numberedFileName(dir, name string, i int) string {
+	extension := path.Ext(name)
+	numberedFileName := fmt.Sprintf("%s (%d)%s", name[:len(name)-len(extension)], i, extension)
+	return filepath.Join(dir, numberedFileName)
+}
+
+func openFileOrSubstitute(dir, base string, action onConflict) (of *os.File, err error) {
```
bradfitz, line 341: I wouldn't name the result params here unless you needed to, but I don't see a reason to.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,110 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+		default:
+			return nil, fmt.Errorf("file issue. how to resolve this conflict? no one knows.")
+		case skipOnExist:
+			if _, err = os.Stat(targetFile); err == nil {
+				err = fmt.Errorf("refusing to overwrite %v. %w", targetFile, openError)
+			}
+			err = fmt.Errorf("failed to write; %w", openError)
+		case overwriteExisting:
+			if of, err = os.OpenFile(targetFile, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644); err != nil {
+				err = fmt.Errorf("unable to overwrite %v. %w", targetFile, err)
```
bradfitz, line 355: return fmt.Errorf(...  to overwrite file: %w" ...). (the %w from err will include the filename, so no need to include targetFile again)

cmd/tailscale/cli/file.go
davideger: changed to be explicit at each return site here and below.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,109 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+
+func numberedFileName(dir, name string, i int) string {
+	extension := path.Ext(name)
+	numberedFileName := fmt.Sprintf("%s (%d)%s", name[:len(name)-len(extension)], i, extension)
+	return filepath.Join(dir, numberedFileName)
+}
+
+func openFileOrSubstitute(dir, base string, action onConflict) (of *os.File, err error) {
+	targetFile := filepath.Join(dir, base)
+	of, openError := os.OpenFile(targetFile, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0644)
```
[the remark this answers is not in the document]
davideger: sounds good.

cmd/tailscale/cli/file.go
davideger: done

cmd/tailscale/cli/file.go
davideger: done, good to know the convention. probably clearer this way.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,109 @@ func runCpTargets(ctx context.Context, args []string) error {
...
 	Name:       "get",
-	ShortUsage: "file get [--wait] [--verbose] <target-directory>",
+	ShortUsage: "file get [--wait] [--verbose] [--conflict=(skip|overwrite|rename)] <target-directory>",
 	ShortHelp:  "Move files out of the Tailscale file inbox",
 	Exec:       runFileGet,
 	FlagSet: (func() *flag.FlagSet {
 		fs := newFlagSet("get")
 		fs.BoolVar(&getArgs.wait, "wait", false, "wait for a file to arrive if inbox is empty")
 		fs.BoolVar(&getArgs.verbose, "verbose", false, "verbose output")
+		fs.Var(&getArgs.conflict, "conflict", "behavior when target file already exists")
```
bradfitz, line 323: I ran this and the output is:

```
USAGE
  file get [--wait] [--verbose] [--conflict=(skip|overwrite|rename)] <target-directory>

FLAGS
  -conflict skip  behavior when target file already exists
  -verbose=false  verbose output
  -wait=false     wait for a file to arrive if inbox is empty
```

I think the conflict flag's docs should be fleshed out here. Maybe:

`"how to handle conflicting filenames: \"overwrite\" to overwrite the local existing file, \"rename\" to make a new number-suffixed local name that's unique, or \"skip\" to ....`

Actually, what does "skip" even do? It looks like it fails and doesn't skip?

Ah, it gathers accumulated errors and prints them all, but doesn't delete the waiting file from the "inbox" area.

cmd/tailscale/cli/file.go
```diff
@@ -330,47 +418,40 @@ func runFileGet(ctx context.Context, args []string) error {
...
 	}
 	if getArgs.verbose {
-		log.Printf("moved %d files", deleted)
+		printf("moved %d/%d files", deleted, len(wfs))
 	}
-	return nil
+	if len(errs) == 0 {
+		return nil
+	}
+	for _, err := range errs[1:] {
```
bradfitz, line 451: How about the other way? return the last one and print errs[:len(errs-1)] here instead? So they're printed in order.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,109 @@ func runCpTargets(ctx context.Context, args []string) error {
 	return nil
 }
 
+type onConflict string
```
bradfitz, line 290: // onConflict is a flag.Value for the --conflict flag's three string options.

cmd/tailscale/cli/file.go
```diff
@@ -286,22 +287,109 @@ func runCpTargets(ctx context.Context, args []string) error {
...
+
+const (
+	skipOnExist         onConflict = "skip"
+	overwriteExisting   onConflict = "overwrite" //  Overwrite any existing file at the target location
+	createNumberedFiles onConflict = "rename"    //  Create an alternately named file in the style of Chrome Downloads
+)
+
+func (v *onConflict) String() string {
```
bradfitz, line 298: this can be all on one line if you want

cmd/tailscale/cli/file.go
bradfitz: Maybe "error" instead of "skip"? 🤔

bradfitz (commented):
Pretty close!

cmd/tailscale/cli/file.go
davideger: sure thing

cmd/tailscale/cli/file.go
davideger: done

davideger:
Thanks for bearing with me.  This PR has gotten much better with your help.

davideger:
Hi Brad,

I wrote a little HOWTO for Linux users to finish setting up Taildrop [here](https://github.com/davideger/blog/blob/main/taildrop_on_linux.md) that shows how to resolve #2057, #2312 and #2324.  While writing the HOWTO, I updated the `overwrite` behavior added in this PR so we do a remove+write O_EXCL to avoid accidentally following symlinks out of the target directory.

Let me know if there's more you'd like to see in this PR.  Cheers!

bradfitz (approved):
Thanks!

PR #3968, opened 2022-02-18
cmd/tailscaled: add `-state=mem:` to support creation of an ephemeral node

maisem:
RELNOTE=`tailscaled --state=mem:` registers as an ephemeral node and
does not store state to disk.

### DIFF
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"inet.af/netaddr"
+	"tailscale.com/control/controlclient"
 	"tailscale.com/envknob"
 	"tailscale.com/ipn"
 	"tailscale.com/ipn/ipnserver"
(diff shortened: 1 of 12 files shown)


ipn/ipnserver/server.go
```diff
@@ -668,9 +673,12 @@ func StateStore(path string, logf logger.Logf) (ipn.StateStore, error) {
 	if path == "" {
 		return &ipn.MemoryStore{}, nil
 	}
+	const ephemeralPrefix = "ephemeral:"
```
bradfitz, line 676: how about just making this be "mem:" ?

cmd/tailscaled/tailscaled.go
```diff
@@ -238,8 +239,19 @@ func ipnServerOpts() (o ipnserver.Options) {
 			o.VarRoot = dir
 		}
 	}
+	if strings.HasPrefix(statePathOrDefault(), "ephemeral:") {
+		// Register as an ephemeral node.
+		o.LoginFlags |= controlclient.LoginEphemeral
```
bradfitz, line 244: no need for the `|` .... we made this `o` and there's no assignment to `LoginFlags` before this.

bradfitz (commented):
but no objections otherwise

ipn/ipnserver/server.go
bradfitz: I like `mem:`. The flag's about where things are stored.

Ephemeral is just an optimization that falls out naturally from that. (the fact that we then know to aggressively GC those)

`mem` is also much shorter.

And it leaves it open for us to adding an explicit `--type={,regular,ephemeral}` flag somewhere later too, orthogonal to whether storage was in memory or on disk.

cmd/tailscaled/tailscaled.go
maisem: Done

bradfitz (approved):
Add a RELNOTE= line to the commit message before submitting.

PR #3969, opened 2022-02-18
types/logger, logtail: add mechanism to do structured JSON logs

bradfitz:
e.g. the change to ipnlocal in this commit ultimately logs out:

```
{"logtail":{"client_time":"2022-02-17T20:40:30.511381153-08:00","server_time":"2022-02-18T04:40:31.057771504Z"},"type":"Hostinfo","val":{"GoArch":"amd64","Hostname":"tsdev","IPNVersion":"1.21.0-date.20220107","OS":"linux","OSVersion":"Debian 11.2 (bullseye); kernel=5.10.0-10-amd64"}}
```

### DIFF
--- a/logtail/logtail.go
+++ b/logtail/logtail.go
@@ -530,6 +530,9 @@ func (l *Logger) encode(buf []byte, level int) []byte {
 			"client_time": now.Format(time.RFC3339Nano),
 		}
 	}
+	if level > 0 {
+		obj["v"] = level
+	}
 
 	b, err := json.Marshal(obj)
 	if err != nil {
@@ -577,6 +580,7 @@ var (
 	openBracketV = []byte("[v")
 	v1           = []byte("[v1] ")
 	v2           = []byte("[v2] ")
+	vJSON        = []byte("[v\x00JSON]") // precedes log level '0'-'9' byte, then JSON value
 )
 
 // level 0 is normal (or unknown) level; 1+ are increasingly verbose
@@ -589,6 +593,15 @@ func parseAndRemoveLogLevel(buf []byte) (level int, cleanBuf []byte) {
 	}
 	if bytes.Contains(buf, v2) {
 		return 2, bytes.ReplaceAll(buf, v2, nil)
+	}
+	if i := bytes.Index(buf, vJSON); i != -1 {
+		rest := buf[i+len(vJSON):]
+		if len(rest) >= 2 {
+			v := rest[0]
+			if v >= '0' && v <= '9' {
+				return int(v - '0'), rest[1:]
+			}
+		}
 	}
 	return 0, buf
 }
--- a/types/logger/logger.go
+++ b/types/logger/logger.go
@@ -9,7 +9,9 @@ package logger
 
 import (
 	"bufio"
+	"bytes"
 	"container/list"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -34,6 +36,49 @@ type Logf func(format string, args ...interface{})
 type Context context.Context
 
 type logfKey struct{}
+
+// jenc is a json.Encode + bytes.Buffer pair wired up to be reused in a pool.
+type jenc struct {
+	buf bytes.Buffer
+	enc *json.Encoder
+}
+
+var jencPool = &sync.Pool{New: func() interface{} {
+	je := new(jenc)
+	je.enc = json.NewEncoder(&je.buf)
+	return je
+}}
+
+// JSON marshals v as JSON and writes it to logf formatted with the annotation to make logtail
+// treat it as a structured log.
+//
+// The recType is the record type and becomes the key of the wrapper
+// JSON object that is logged. That is, if recType is "foo" and v is
+// 123, the value logged is {"foo":123}.
+//
+// Do not use recType "logtail" or "v", with any case. Those are
+// reserved for the logging system.
+//
+// The level can be from 0 to 9. Levels from 1 to 9 are included in
+// the logged JSON object, like {"foo":123,"v":2}.
+func (logf Logf) JSON(level int, recType string, v interface{}) {
+	je := jencPool.Get().(*jenc)
+	defer jencPool.Put(je)
+	je.buf.Reset()
+	je.buf.WriteByte('{')
+	je.enc.Encode(recType)
+	je.buf.Truncate(je.buf.Len() - 1) // remove newline from prior Encode
+	je.buf.WriteByte(':')
+	if err := je.enc.Encode(v); err != nil {
+		logf("[unexpected]: failed to encode structured JSON log record of type %q / %T: %v", recType, v, err)
+		return
+	}
+	je.buf.Truncate(je.buf.Len() - 1) // remove newline from prior Encode
+	je.buf.WriteByte('}')
+	// Magic prefix recognized by logtail:
+	logf("[v\x00JSON]%d%s", level%10, je.buf.Bytes())
+
+}
 
 // FromContext extracts a log function from ctx.
 func FromContext(ctx Context) Logf {
--- a/types/logger/logger_test.go
+++ b/types/logger/logger_test.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	qt "github.com/frankban/quicktest"
+	"tailscale.com/tailcfg"
 )
 
 func TestFuncWriter(t *testing.T) {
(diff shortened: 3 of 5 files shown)


bradfitz:
cc @danderson @dsnet @crawshaw

logtail/logtail.go
```diff
@@ -590,5 +591,8 @@ func parseAndRemoveLogLevel(buf []byte) (level int, cleanBuf []byte) {
 	if bytes.Contains(buf, v2) {
 		return 2, bytes.ReplaceAll(buf, v2, nil)
 	}
+	if i := bytes.Index(buf, vJSON); i != -1 {
+		return 1, buf[i+len(vJSON):]
```
dsnet, line 595: Why is JSON level 1?

types/logger/logger.go
```diff
@@ -35,6 +36,23 @@ type Context context.Context
 
 type logfKey struct{}
 
+// J marshals v as JSON and writes it to logf formatted with the annotation to make logtail
+// treat it as a structured log.
+func J(logf Logf, recType string, v interface{}) {
+	// TODO: pool of json.Encoder+bytes.Buffer pre-wired pairs
+	enc, err := json.Marshal(struct {
```
dsnet, line 43: There's a benefit to encoding the type as a key in the JSON object, rather than as a member field. This ensures that the type has to appear before the payload at a syntactic level. One of my frustrations implementing `google.protobuf.Any` in JSON is that we could not always know the type before we parsed the body, which made a few optimizations difficult.

Thus, something like:
```go
enc, err := json.Marshal(val)
if err != nil {
    ...
}
logf("[v\x00JSON:{%q:%s}", recType, enc)
```

(I'm making the assumption that `recType` contains no special characters such that "%q" is equivalent to a JSON serialized string)

dsnet:
I'm a proponent of structured logging. Is there a reason we didn't a embrace this earlier?

logtail/logtail.go
josharian: So that it goes to logcatcher but not syslog/etc. This helps reduce the volume of logs stored locally, which is an ongoing source of (not unreasonable) complaints.

logtail/logtail.go
```diff
@@ -577,6 +577,7 @@ var (
 	openBracketV = []byte("[v")
 	v1           = []byte("[v1] ")
 	v2           = []byte("[v2] ")
+	vJSON        = []byte("[v\x00JSON:")
```
josharian, line 580: Why no trailing `]`? E.g. `[v\x00JSON] `? I'm sure there's a reason, I just can't intuit what it is. Document it?

types/logger/logger.go
```diff
@@ -35,6 +36,23 @@ type Context context.Context
 
 type logfKey struct{}
 
+// J marshals v as JSON and writes it to logf formatted with the annotation to make logtail
+// treat it as a structured log.
+func J(logf Logf, recType string, v interface{}) {
+	// TODO: pool of json.Encoder+bytes.Buffer pre-wired pairs
```
josharian, line 42: We have this for unmarshal in util/jsonutil, so it'd be a natural addition there. Although if I recall correctly, everyone hates that. :P And anyway, @dsnet is going to fix all our JSON performance issues. :)

types/logger/logger.go
```diff
@@ -35,6 +36,23 @@ type Context context.Context
 
 type logfKey struct{}
 
+// J marshals v as JSON and writes it to logf formatted with the annotation to make logtail
+// treat it as a structured log.
+func J(logf Logf, recType string, v interface{}) {
```
josharian, line 41: Suggest s/J/JSON/. It's not that long, and it's a lot clearer.

josharian: What happens when `v` is not a struct/map? I'd like to see some tests for at least a slice and a basic type.

Also, corner case, but what if `v` contains a field called `logtail`? (It's not impossible to imagine this happening legitimately, if you were logging from within the logger.)

types/logger/logger.go
dsnet: 😏

logtail/logtail.go
dsnet: JSON is at-least self-terminating, so we don't absolutely need some out-of-band terminator. A trailing ']' is aesthetically nicer, but slightly more work later to strip it later.

types/logger/logger.go
bradfitz: Wait, Logf is a type. How about a JSON method on logf?

`logf.JSON(1, "Hostinfo", hi)`

... with the verbosity level thrown in there too?

josharian: SGTM, but in an ideal world, the verbosity level would be some named constant rather than a bare int, to make it more obvious at call sites what it means.

bradfitz:
PTAL

types/logger/logger.go
```diff
@@ -35,6 +37,46 @@ type Context context.Context
...
+// JSON marshals v as JSON and writes it to logf formatted with the annotation to make logtail
+// treat it as a structured log.
+//
+// The recType is the record type and becomes the key of the wrapper
+// JSON object that is logged. That is, if recType is "foo" and v is
+// 123, the value logged is {"foo":123}.
+//
+// The level can be from 0 to 9. Levels from 1 to 9 are included in
+// the logged JSON object, like {"foo":123,"v":2}.
+func (logf Logf) JSON(level int, recType string, v interface{}) {
```
josharian, line 61: `recType == "v"`: Don't do that?

types/logger/logger.go
dsnet: BTW (in case anyone is thinking about it), I like the fact that we're using our own specified type name rather than the fully qualified Go type name. Unfortunately, Go type names are not stable over time ever since type aliases were introduced (see https://github.com/golang/go/issues/36345).

types/logger/logger.go
bradfitz: oh, right. I'll also warn against "logtail"

logtail/logtail.go
```diff
@@ -590,5 +594,14 @@ func parseAndRemoveLogLevel(buf []byte) (level int, cleanBuf []byte) {
 	if bytes.Contains(buf, v2) {
 		return 2, bytes.ReplaceAll(buf, v2, nil)
 	}
+	if i := bytes.Index(buf, vJSON); i != -1 {
+		rest := buf[i+len(vJSON):]
+		if len(rest) >= 2 {
+			v := rest[0]
+			if v >= '0' && v <= '9' {
```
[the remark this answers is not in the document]
bradfitz: if we have %02d log levels we've lost.

I'm going to lock ourselves into 10.

types/logger/logger.go
```diff
@@ -35,6 +37,46 @@ type Context context.Context
...
+	je.buf.Truncate(je.buf.Len() - 1) // remove newline from prior Encode
+	je.buf.WriteByte(':')
+	if err := je.enc.Encode(v); err != nil {
+		logf("[unexpected]: failed to encode structured JSON log record of type %q / %T: %v", recType, v, err)
+		return
+	}
+	je.buf.Truncate(je.buf.Len() - 1) // remove newline from prior Encode
+	je.buf.WriteByte('}')
+	// Magic prefix recognized by logtail:
+	logf("[v\x00JSON]%d%s", level%10, je.buf.Bytes())
```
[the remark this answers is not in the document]
bradfitz: 🤷‍♂️

types/logger/logger_test.go
```diff
@@ -211,3 +212,13 @@ func TestContext(t *testing.T) {
 	logf("a")
 	c.Assert(called, qt.IsTrue)
 }
+
+func TestJSON(t *testing.T) {
+	var buf bytes.Buffer
+	var logf Logf = func(f string, a ...interface{}) { fmt.Fprintf(&buf, f, a...) }
+	logf.JSON(1, "foo", &tailcfg.Hostinfo{})
+	want := "[v\x00JSON]1{\"foo\":{\"OS\":\"\",\"Hostname\":\"\"}}"
```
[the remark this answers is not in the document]
bradfitz: can't do \x00 in backticks. But I guess I could concat.

logtail/logtail.go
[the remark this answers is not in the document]
josharian: The number of children you have and the number of log levels you have should have the same order of magnitude.

types/logger/logger_test.go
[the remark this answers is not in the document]
josharian: Yeah, something like `"[v\x00JSON]"+...`

types/logger/logger.go
bradfitz: It's not necessarily just a type name, though.

Think of cases like: ("FooBeforeLogin", foo) ... ("FooAfterLogin", foo).

Let's try this as-is and we can always refactor later since it's an internal detail not going over the wire.

PR #3976, opened 2022-02-18
tailcfg: make MapResponse.ControlTime a pointer

josharian:
Otherwise omitempty doesn't work.

This is wire-compatible with a non-pointer type, so switching
is safe, now and in the future.


josharian:
Also, filed https://github.com/golang/go/issues/51261 for a vet check so that we don't make this mistake in the future.

PR #3983, opened 2022-02-19
ipn/ipnserver: let TS_PERMIT_CERT_UID contain a username too, not just uid

bradfitz:
Don't make users map their system's "caddy" (or whatever) system user
to its userid. We can do that. Support either a uid or a username.

RELNOTE=TS_PERMIT_CERT_UID can contain a uid or username

### DIFF
--- a/ipn/ipnserver/server.go
+++ b/ipn/ipnserver/server.go
@@ -458,6 +458,29 @@ func (s *Server) localAPIPermissions(ci connIdentity) (read, write bool) {
 	return false, false
 }
 
+// userIDFromString maps from either a numeric user id in string form
+// ("998") or username ("caddy") to its string userid ("998").
+// It returns the empty string on error.
+func userIDFromString(v string) string {
+	if v == "" || isAllDigit(v) {
+		return v
+	}
+	u, err := user.Lookup(v)
+	if err != nil {
+		return ""
+	}
+	return u.Uid
+}
+
+func isAllDigit(s string) bool {
+	for i := 0; i < len(s); i++ {
+		if b := s[i]; b < '0' || b > '9' {
+			return false
+		}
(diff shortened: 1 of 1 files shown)


bradfitz:
cc @mholt

bradfitz:
cc @mayakacz @WalterHub

ipn/ipnserver/server.go
```diff
@@ -458,6 +458,43 @@ func (s *Server) localAPIPermissions(ci connIdentity) (read, write bool) {
...
+func userIDFromString(v string) string {
+	if v == "" || isAllDigit(v) {
+		return v
+	}
+	uidMapMu.Lock()
+	defer uidMapMu.Unlock()
+	if id, ok := uidMap[v]; ok {
+		return id
+	}
+	u, err := user.Lookup(v)
```
DentonGentry, line 478: I'm trying to think through whether there are cases for a long-running tailscaled where an older cached value would be wrong (since we have no way of knowing to clear it here). I guess if someone deletes a user from /etc/passwd, or if LDAP is in use, a UID could change. It doesn't seem harmful though.

DentonGentry (approved):

ipn/ipnserver/server.go
bradfitz: Actually I'm not sure why I even added a cache. It's pointless. Removed.

DentonGentry (approved):

mholt:
Ohh, very nice -- thank you! That's definitely more convenient.

ksylvan:
How does this work on MacOS? I have tailscale installed via `brew install --cask tailscale` and I don't know where to set the environment variable for tailscaled. Please help! Thanks.

PR #3987, opened 2022-02-20
hostinfo: detect NSIS vs MSI package type on Windows

bradfitz:


hostinfo/hostinfo.go
```diff
@@ -51,15 +51,47 @@ func GetOSVersion() string {
...
+			// Atypical. Not worth trying to detect. Likely open
+			// source tailscaled or a developer running by hand.
+			return ""
+		}
+		nsisUninstaller := filepath.Join(dir, "Uninstall-Tailscale.exe")
+		_, err = os.Stat(nsisUninstaller)
+		if err == nil {
+			return "nsis"
+		}
+		if os.IsNotExist(err) {
```
DentonGentry, line 83: I've installed from MSI but I have an Uninstall-Tailscale.exe in %PATH%.

The system was running NSIS before, then I installed MSI. The MSI installer is supposed to have uninstalled NSIS before proceeding, though I didn't actually watch it do so.

```
C:\Users\dgentry>dir "C:\Program Files (x86)\Tailscale IPN\Uninstall-Tailscale.exe"
 Volume in drive C has no label.
 Volume Serial Number is AA73-648A

 Directory of C:\Program Files (x86)\Tailscale IPN

02/19/2022  11:54 PM            50,762 Uninstall-Tailscale.exe
               1 File(s)         50,762 bytes
               0 Dir(s)  168,263,729,152 bytes free

C:\Users\dgentry>
```

bradfitz: But that's in the wrong Program Files path. Due to a bug in our NSIS installer, that's our 64-bit binary in the 32-bit directory.

Once you install the MSI you should have the correct files in `C:\Program Files\`

This code uses os.Executable to see what binary you're actually running and looks at sibling files in that directory.

DentonGentry: I completely uninstalled Tailscale and installed again using the MSI, and Uninstall-Tailscale.exe is gone now. The previous MSI installation apparently failed? Dunno.

DentonGentry (approved):

dblohm7:
MSI also writes a sentinel value to the registry. That is probably the most authoritative way to check this. I'll write a follow-up patch.

PR #3998, opened 2022-02-21
ipn/{ipnlocal,ipnstate}: add Tags and PrimaryRoutes to PeerStatus

maisem:
This adds `Tags` and `PrimaryRoutes` to the output of `tailscale status --json`

cc @mayakacz 

### DIFF
--- a/ipn/ipnstate/ipnstate.go
+++ b/ipn/ipnstate/ipnstate.go
@@ -101,14 +101,22 @@ type PeerStatusLite struct {
 }
 
 type PeerStatus struct {
-	ID        tailcfg.StableNodeID
-	PublicKey key.NodePublic
-	HostName  string // HostInfo's Hostname (not a DNS name or necessarily unique)
-	DNSName   string
-	OS        string // HostInfo.OS
-	UserID    tailcfg.UserID
-
+	ID           tailcfg.StableNodeID
+	PublicKey    key.NodePublic
+	HostName     string // HostInfo's Hostname (not a DNS name or necessarily unique)
+	DNSName      string
+	OS           string // HostInfo.OS
+	UserID       tailcfg.UserID
 	TailscaleIPs []netaddr.IP // Tailscale IP(s) assigned to this node
+
+	// Tags are the list of ACL tags applied to this node.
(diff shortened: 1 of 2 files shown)


ipn/ipnstate/ipnstate.go
```diff
@@ -108,8 +108,12 @@ type PeerStatus struct {
 	OS        string // HostInfo.OS
 	UserID    tailcfg.UserID
 
+	Tags []string // Tags assigned to this node
```
bradfitz, line 111: omitempty?

And maybe promote to a lines-before formal comment and describe this more: is this only the ones that were accepted or the ones that were requested?

ipn/ipnstate/ipnstate.go
```diff
@@ -108,8 +108,12 @@ type PeerStatus struct {
...
+
 	TailscaleIPs []netaddr.IP // Tailscale IP(s) assigned to this node
 
+	PrimaryRoutes []netaddr.IPPrefix // Routes being routed via this node
```
bradfitz, line 115: same. omitempty & elaborate. Does this include TailscaleIPs /32 and /128? Or only a subset of offered routes?

ipn/ipnstate/ipnstate.go
maisem: Done, also added a pointer to tailcfg#Node.Tags.

ipn/ipnstate/ipnstate.go
maisem: Done. No they don't; described in comments.

bradfitz (approved):

bradfitz:
All RELNOTE= to commit.

Also, if this for 1.22? Or should it wait until after 1.22 is branched? cc @DentonGentry

maisem:
> All RELNOTE= to commit.
> 
> 
> 
> Also, if this for 1.22? Or should it wait until after 1.22 is branched? cc @DentonGentry 

Oops! Will coordinate with @DentonGentry

bradfitz:
Go update the release notes wiki now, before you forget?

PR #3999, opened 2022-02-21
go.mod: update github.com/mdlayher/netlink to 1.6.0

danderson:
This unbreaks some downstream users of tailscale who end up
with build errors from importing a v0 indirect dependency.

Specifically for zekjur and gokrazy.


mdlayher (approved):
LGTM thanks!

PR #4001, opened 2022-02-22
ipn/ipnlocal: use views for Peer.PrimaryRoutes and Peer.Tags 

maisem:
RELNOTE=`tailscale status --json` now shows Tags and PrimaryRoutes


DentonGentry (approved):

bradfitz (commented):
Maybe they should be views! They're both in the tree now? And can do JSON?

maisem:
> Maybe they should be views! They're both in the tree now? And can do JSON?

Done. PTAL

ipn/ipnstate/ipnstate.go
```diff
@@ -111,12 +112,12 @@ type PeerStatus struct {
 
 	// Tags are the list of ACL tags applied to this node.
 	// See tailscale.com/tailcfg#Node.Tags for more information.
-	Tags []string `json:",omitempty"`
+	Tags views.StringSlice `json:",omitempty"`
```
bradfitz, line 115: does omitempty work on a view? (add a little test?)

maisem: Apparently not, the only way around this would be to make this a pointer which is gross.

ipn/ipnstate/ipnstate.go
```diff
@@ -111,12 +112,12 @@ type PeerStatus struct {
 
 	// Tags are the list of ACL tags applied to this node.
 	// See tailscale.com/tailcfg#Node.Tags for more information.
-	Tags []string `json:",omitempty"`
+	Tags views.StringSlice
```
bradfitz, line 115: To get omitempty working, you'll probably need to make these pointer-to-view

maisem: Done

bradfitz (approved):

bradfitz:
Add RELNOTE first

maisem:
> Add RELNOTE first

Heh, thanks for the reminder! We already have a RELNOTE for this in the wiki, but it was missed in the previous commit message. I'll add it here just so that it's on some commit message

PR #4002, opened 2022-02-22
net/tstun: set link speed to SPEED_UNKNOWN

### LINKED ISSUE #3933: tailscale0 interface exposes wrong speed to kernel
### What is the issue?

I'm using Prometheus to monitor network interfaces saturation by node_network_speed_bytes. There has been a lot of alerts regarding `tailscale0` interface only. I looked into the traffics on the interface, and there's little traffic that would never trigger such alert. Looking into `/sys/class/net/tailscale0/speed`, apparently Tailscale interface speed is set to a wrong value.

### Steps to reproduce

1. Value of `/sys/class/net/tailscale0/speed`: `10`
2. Output of `ethtool tailscale0`:
```
$ ethtool tailscale0
Settings for tailscale0:
        Supported ports: [  ]
        Supported link modes:   Not reported
        Supported pause frame use: No
        Supports auto-negotiation: No
        Supported FEC modes: Not reported
        Advertised link modes:  Not reported
        Advertised pause frame use: No
        Advertised auto-negotiation: No
        Advertised FEC modes: Not reported
        Speed: 10Mb/s
        Duplex: Full
        Auto-negotiation: off
        Port: Twisted Pair
        PHYAD: 0
        Transceiver: internal
        MDI-X: Unknown
        Current message level: 0x00007fa1 (32673)
                               drv ifup tx_err tx_queued intr tx_done rx_status pktdata hw wol
        Link detected: yes
```

_No response_

### OS

Linux

### OS version

CentOS 8 / 4.18.0-348.2.1.el8_5.x86_64

### Tailscale version

1.20.2-t312750ddd-g2c164d9c7

### Bug report

_No response_

dshynkev:
This should fix #3933 by setting the link speed on the TUN interface to `SPEED_UNKNOWN`, as is appropriate for a non-physical interface. Prometheus should be aware of this special case now: prometheus/node_exporter#1989.

A less satisfying option would be to set a high value via `setLinkSpeed(iface, 10_000)` to get 10Gbps or similar. This may be better for software that isn't aware of SPEED_UNKNOWN.

This arguably belongs in `wireguard-go` with the rest of the TUN setup (in fact, `setLinkSpeed` is modeled after said setup from `tun/tun_linux.go`), but putting it here allows for it to seamlessly apply to the TAP case. It seems wrong to indiscriminately do this in `CreateUnmonitoredTUNFromFD` in `wireguard-go`, but perhaps I'm mistaken; I can send in a patch there if you think that's the right place.


net/tstun/linkattrs_linux.go
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
```
bradfitz, line 1: wrong year, unless you've been sitting on this and waiting to mail it.

net/tstun/linkattrs_linux.go
```diff
@@ -0,0 +1,99 @@
...
+// license that can be found in the LICENSE file.
+
+package tstun
+
+import (
+	"fmt"
+	"unsafe"
+
+	"golang.zx2c4.com/wireguard/tun"
+
```
bradfitz, line 12: delete line

net/tstun/linkattrs_linux.go
```diff
@@ -0,0 +1,99 @@
...
+	cmd     uint32
+	_       [8]byte
+	speedLo uint16
+	_       [14]byte
+	speedHi uint16
+	_       [34]byte
+}
+
+// ifreq is struct ifreq from linux/if.h.
+// We use this instead of unix.Ifreq because the latter
```
bradfitz, line 27: We can just modify x/sys/unix to add the bits you need for both this and probably the `ethtool_cmd` above. (cc @mdlayher @tklauser who are very active maintainers of x/sys/unix)

I'd rather we do that over maintaining this copy.

net/tstun/linkattrs_linux.go
```diff
@@ -0,0 +1,99 @@
...
+// does not expose a setter for ifr_data (ifreqData.withData).
+type ifreq struct {
+	name [unix.IFNAMSIZ]byte
+	data unsafe.Pointer
+	_    [40]byte // data is actually a member of a large union
+}
+
+// linkSpeedUnknown is SPEED_UNKNOWN from linux/ethtool.h,
```
bradfitz, line 35: maybe a constant in x/sys/unix for this too?

bradfitz (commented):
Thanks!

DentonGentry:
One note on timing: this change in particular can have downstream impacts on monitoring or other infrastructure which are impossible for us to predict. I'd request that this go in soon after we've released a new major version, so that it can get maximum exposure in the unstable channel trying to find anything it impacts.

So even if the PR were ready today, I'd ask that it wait until 1.22.0 is branched before it gets submitted to main.

net/tstun/linkattrs_linux.go
bradfitz: Actually, there's a TODO from @mdlayher already:

```go
// TODO(mdlayher): export as IfreqData? For now we can provide helpers such as
// IoctlGetEthtoolDrvinfo which use these APIs under the hood.

// An ifreqData is an Ifreq which carries pointer data. To produce an ifreqData,
// use the Ifreq.withData method.
type ifreqData struct {
        name [IFNAMSIZ]byte
        // A type separate from ifreq is required in order to comply with the
        // unsafe.Pointer rules since the "pointer-ness" of data would not be
        // preserved if it were cast into the byte array of a raw ifreq.
        data unsafe.Pointer
        // Pad to the same size as ifreq.
        _ [len(ifreq{}.Ifru) - SizeofPtr]byte
}

// withData produces an ifreqData with the pointer p set for ioctls which require
// arbitrary pointer data.
func (ifr Ifreq) withData(p unsafe.Pointer) ifreqData {
        return ifreqData{
                name: ifr.raw.Ifrn,
                data: p,
        }
}
```

net/tstun/linkattrs_linux.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"fmt"
+	"unsafe"
+
+	"golang.zx2c4.com/wireguard/tun"
+
+	"golang.org/x/sys/unix"
+)
+
+// ethtoolCmd is struct ethtool_cmd from linux/ethtool.h
```
bradfitz, line 16:
```
 * struct ethtool_cmd - DEPRECATED, link control and status                                                                                                                                                               
 * This structure is DEPRECATED, please use struct ethtool_link_settings.                                                                                                                                                 
```

Did you try using `ethtool_link_settings`?

dshynkev: Yes to this. That didn't work at least on my machine, in the sense that the GLINKSETTINGS cmd returns 0 but fills the struct with speed = 0, and submitting a SLINKSETTINGS fails with EINVAL. The recommended way is to try GLINKSETTINGS first and then fall back, which is probably the right thing to do after all, but the silent failure (?) is unfortunate. I'll investigate more while the x/sys/unix change is pending.

dshynkev: Ah, I remembered a suspected cause and it checks out: per docs, speed = 0 can be observed when the link is down, and presumably one cannot modify it in that state either. This doesn't go away when the TUN device observes EventUp; it probably has to be done via the link monitor or by implementing `waitForInterfaceUp` on linux. No such caveat is given for ethtool_cmd, and, in fact, it appears to work just fine where it is.

net/tstun/linkattrs_linux.go
mdlayher: In addition to the above, it may be doable to use the ethtool netlink interface instead for some of these interactions. Though that one is fairly new and frankly I am not sure what kernels you all support for Tailscale.

https://github.com/mdlayher/ethtool

PRs certainly welcome.

mdlayher: Been there, gotta love that. +1 to a new ifreq ioctl wrapper like the one for EthtoolDrvinfo though assuming genetlink is not amenable due to kernel version support.

bradfitz: This doesn't affect iOS so 1.5K is fine.

mdlayher: Cool. As we know Linux won't break userspace, so if adding a new ioctl to x/sys feels less intrusive then feel free rather than pulling my package. I just figured I'd offer another option.

DentonGentry:
We released 1.22.0 late yesterday. Subsequent 1.24.x builds will be done from a release branch, where we can control what goes in.

Ideally this would go in relatively soon in order to get as much exposure in unstable builds as we can: https://pkgs.tailscale.com/unstable/

We'd expect to release 1.24.0 in about a month. If we get relatively close to that date, we'd again ask that this not go in until after 1.24.0 is branched.

dshynkev:
Didn't mean to close this: pushed wrong ref.

This is not in the final form yet: SPEED_UNKNOWN will go into x/sys/unix, and I'll try to merge the netlink setter into mdlayher/ethtool. However, it should probably work like this instead of the ioctl: reading prior state is unnecessary.

go.mod
```diff
@@ -28,6 +28,7 @@ require (
 	github.com/jsimonetti/rtnetlink v0.0.0-20211203074127-fd9a11f42291
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
 	github.com/klauspost/compress v1.13.6
+	github.com/mdlayher/genetlink v1.0.0
```
mdlayher, line 31: I recommend v1.2.0 at this point.

mdlayher (approved):
LGTM, nice.

net/tstun/linkattrs_linux.go
```diff
@@ -0,0 +1,68 @@
...
+
+	b, err := ae.Encode()
+	if err != nil {
+		return err
+	}
+
+	_, err = conn.Execute(
+		genetlink.Message{
+			Header: genetlink.Header{
+				Command: unix.ETHTOOL_MSG_LINKMODES_SET,
```
mdlayher, line 52: AFAIK all of the set operations require CAP_NET_ADMIN but you probably need that in other areas in Tailscale anyway?

dshynkev: Except in userspace-networking mode AFAIU, but this code won't run in that mode (and failure is OK in this path anyway).

mdlayher (approved):
LGTM!

dshynkev:
Thanks for the review! I'll try to get the setter upstreamed, as a last thing, to make this change smaller.

net/tstun/linkattrs_linux.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"github.com/mdlayher/genetlink"
+	"github.com/mdlayher/netlink"
+	"golang.org/x/sys/unix"
+	"golang.zx2c4.com/wireguard/tun"
+)
+
+// linkSpeedUnknown is SPEED_UNKNOWN from linux/ethtool.h,
+// which is appropriate for a virtual interface.
+const linkSpeedUnknown = ^uint32(0)
```
bradfitz, line 16: use the new constant that's merged?

bradfitz:
Run `make updatedeps` to fix the depaware failure.

bradfitz (approved):

net/tstun/linkattrs_notlinux.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build !linux
+// +build !linux
+
+package tstun
+
+func setLinkAttrs(iface tun.Device) error {
```
bradfitz, line 10: This "tun." reference without an import fails all the non-Linux CI builders.

dshynkev: Oops, my bad :/

bradfitz:
Nice seeing ya around again! 😀

PR #4008, opened 2022-02-22
net/netutil: fix regression where peerapi would get closed after 1st req

bradfitz:
I introduced a bug in 8fe503057da26 when unifying oneConnListener
implementations.

The NewOneConnListenerFrom API was easy to misuse (its Close method
closes the underlying Listener), and we did (via http.Serve, which
closes the listener after use, which meant we were close the peerapi's
listener, even though we only wanted its Addr)

Instead, combine those two constructors into one and pass in the Addr
explicitly, without delegating through to any Listener.

### DIFF
--- a/net/netutil/netutil.go
+++ b/net/netutil/netutil.go
@@ -8,34 +8,49 @@ package netutil
 import (
 	"io"
 	"net"
+	"sync"
 )
 
-// NewOneConnListener returns a net.Listener that returns c on its first
-// Accept and EOF thereafter. The Listener's Addr is a dummy address.
-func NewOneConnListener(c net.Conn) net.Listener {
-	return NewOneConnListenerFrom(c, dummyListener{})
-}
-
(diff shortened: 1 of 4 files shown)


net/netutil/netutil.go
```diff
@@ -8,36 +8,51 @@ package netutil
...
-
-// NewOneConnListenerFrom returns a net.Listener wrapping ln where
-// its Accept returns c on the first call and io.EOF thereafter.
-func NewOneConnListenerFrom(c net.Conn, ln net.Listener) net.Listener {
-	return &oneConnListener{c, ln}
+// NewOneConnListener returns a net.Listener that returns c on its
+// first Accept and EOF thereafter.
+//
+// The returned Listener's Addr method returns addr if non-nil. If nil,
+// the a dummy address is returned.
```
maisem, line 18: `the a`

maisem (approved):

PR #4020, opened 2022-02-23
ssh/tailssh: flesh out env, support non-pty commands

bradfitz:
Updates #3802

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -150,7 +150,7 @@ func (srv *server) handleSSH(s ssh.Session) {
 	}
 	action, localUser, ok := evalSSHPolicy(pol, sctx)
 	if ok && action.Message != "" {
-		io.WriteString(s, action.Message)
+		io.WriteString(s.Stderr(), strings.Replace(action.Message, "\n", "\r\n", -1))
 	}
 	if !ok || action.Reject {
 		logf("ssh: access denied for %q from %v", uprof.LoginName, srcIP)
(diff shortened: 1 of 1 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -150,7 +150,7 @@ func (srv *server) handleSSH(s ssh.Session) {
 	}
 	action, localUser, ok := evalSSHPolicy(pol, sctx)
 	if ok && action.Message != "" {
-		io.WriteString(s, action.Message)
+		io.WriteString(s.Stderr(), strings.Replace(action.Message, "\n", "\n\r", -1))
```
maisem, line 153: fine for now, but we might end up adding a bunch of carriage returns when the input is `\r\n` already. consider pulling it out into a different func and adding a comment.

ssh/tailssh/tailssh.go
```diff
@@ -160,62 +160,105 @@ func (srv *server) handleSSH(s ssh.Session) {
...
-		if u.Username != localUser {
-			fmt.Fprintf(s, "can't switch user\n")
+		cmd = exec.Command(loginShell(lu.Uid))
+	} else {
+		if rawCmd := s.RawCommand(); rawCmd != "" {
+			cmd = exec.Command("/usr/bin/env", "su",
+				"-c", rawCmd,
+				localUser)
+			cmd.Dir = lu.HomeDir
+			cmd.Env = append(cmd.Env,
```
maisem, line 186: pull this out into a func like you did for loginShell

maisem (approved):

ssh/tailssh/tailssh.go
bradfitz: done

ssh/tailssh/tailssh.go
bradfitz: tweaked to \r\n for now but until we have half of a plan here, leaving as-is instead of pretending we do

PR #4022, opened 2022-02-23
ssh/tailssh: break a method into half in prep for testing

bradfitz:
Updates #3802

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -9,6 +9,7 @@
 package tailssh
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -102,7 +103,6 @@ func (srv *server) sshPolicy() (_ *tailcfg.SSHPolicy, ok bool) {
 func (srv *server) handleSSH(s ssh.Session) {
 	lb := srv.lb
 	logf := srv.logf
-
 	sshUser := s.User()
 	addr := s.RemoteAddr()
 	logf("Handling SSH from %v for user %v", addr, sshUser)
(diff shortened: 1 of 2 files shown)


maisem:
commit message has `ssh/tailsh`

ssh/tailssh/tailssh.go
```diff
@@ -167,7 +165,29 @@ func (srv *server) handleSSH(s ssh.Session) {
...
+	}
+	srv.handleAcceptedSSH(s, sessionClose, sctx, lu)
+}
+
+// handleAcceptedSSH handles s once it's been accepted and determined
+// that it should run as user lu.
+//
+// sessionClose, if non-nil, is readable when the session should
+// be forcefully closed with an error that says why.
+func (srv *server) handleAcceptedSSH(s ssh.Session, sessionClose <-chan error, sctx *sshContext, lu *user.User) {
```
maisem, line 184: use a context with timeout/cancel? you lose the error but that's more canonical?

maisem: or you could embed https://pkg.go.dev/context#Context in `sshContext`, make it also expose a `CancelWithErr` function and then have it override the `ctx.Err()` to instead return the err from cancel.

`context.WithTimeoutError`/`context.WithCancelError` might be something I want in general tbh, would be useful to figure out things like who canceled the context and why.

ssh/tailssh/tailssh.go
```diff
@@ -167,7 +166,31 @@ func (srv *server) handleSSH(s ssh.Session) {
 		s.Exit(1)
 	}
 
-	logf("ssh: connection from %v %v to %v@ => %q. command = %q, env = %q", srcIP, uprof.LoginName, sshUser, localUser, s.Command(), s.Environ())
+	var ctx context.Context = context.Background()
+	if action.SesssionDuration != 0 {
+		sctx := newSSHContext()
+		ctx = sctx
+		t := time.AfterFunc(action.SesssionDuration, func() {
+			sctx.CloseWithError(userVisibleError{fmt.Errorf("Session timeout of %v elapsed.", action.SesssionDuration)})
```
maisem, line 174: should this wrap `context.DeadlineExceeded` with a `fmt.Sprintf("...: %w")`?

maisem (approved):

PR #4025, opened 2022-02-24
ipn/ipnlocal: generate tailscaled-owned SSH keys as needed

bradfitz:
Updates #3802

### DIFF
--- a/ipn/ipnlocal/ssh.go
+++ b/ipn/ipnlocal/ssh.go
@@ -8,10 +8,20 @@
 package ipnlocal
 
 import (
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
+	"sync"
 
 	"golang.org/x/crypto/ssh"
 	"tailscale.com/envknob"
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/ssh.go
```diff
@@ -8,28 +8,99 @@
...
 
 var useHostKeys = envknob.Bool("TS_USE_SYSTEM_SSH_HOST_KEYS")
 
-func (b *LocalBackend) GetSSH_HostKeys() ([]ssh.Signer, error) {
-	// TODO(bradfitz): generate host keys, at least as needed if
-	// an existing SSH server didn't put them on disk. But also
-	// because people may want tailscale-specific ones. For now be
-	// lazy and reuse the host ones.
-	return b.getSystemSSH_HostKeys()
+var keyTypes = []string{"rsa", "ecdsa", "ed25519"}
```
maisem, line 32: add a comment that these are only used by tailscale to generate ssh host keys and doesn't impact system generated ssh key behavior.

ipn/ipnlocal/ssh.go
```diff
@@ -8,28 +8,99 @@
...
+	var priv interface{}
+	switch typ {
+	default:
+		return nil, fmt.Errorf("unsupported key type %q", typ)
+	case "ed25519":
+		_, priv, err = ed25519.GenerateKey(rand.Reader)
+	case "ecdsa":
+		priv, err = ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	case "rsa":
+		priv, err = rsa.GenerateKey(rand.Reader, 2048)
```
maisem, line 87: make it a constant somewhere i guess. All of this is only to satisfy the client so doesn't really matter as all of this goes over Tailscale anyways.

ipn/ipnlocal/ssh.go
maisem: oh wait it does. comments pls

maisem (approved):

PR #4029, opened 2022-02-24
ssh/tailssh: create login sessions for new connections

maisem:


ssh/tailssh/tailssh.go
```diff
@@ -165,6 +167,13 @@ func (srv *server) handleSSH(s ssh.Session) {
 		logf("ssh: user Lookup %q: %v", localUser, err)
 		s.Exit(1)
 	}
+	ci.localUser = localUser
+	uid, err := strconv.ParseUint(lu.Uid, 10, 32)
+	if err != nil {
+		logf("ssh: user Lookup %q: %v", localUser, err)
+		s.Exit(1)
```
bradfitz, line 174: don't forget "return" after this

ssh/tailssh/tailssh.go
```diff
@@ -297,6 +319,71 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
...
+	conn, err := dbus.SystemBus()
+	if err != nil {
+		return err
+	}
+	name, objectPath := "org.freedesktop.login1", "/org/freedesktop/login1"
+	obj := conn.Object(name, dbus.ObjectPath(objectPath))
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	call := obj.CallWithContext(ctx, "org.freedesktop.login1.Manager.CreateSession", dbus.FlagNoAutoStart, sessionID)
```
bradfitz, line 332: CreateSession in releaseSession?

ssh/tailssh/tailssh.go
```diff
@@ -297,6 +319,71 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
...
+	return call.Err
+}
+
+func startSession(ci *sshConnInfo) (sessionID string, existing bool, err error) {
+	conn, err := dbus.SystemBus()
+	if err != nil {
+		// DBus probably not running.
+		return "", false, err
```
bradfitz, line 340: return err nil?

otherwise caller would fail if dbus isn't running

ssh/tailssh/tailssh.go
maisem: Ah, the line above is also borked. Will fix

ssh/tailssh/child.go
```diff
@@ -0,0 +1,214 @@
...
+	"log/syslog"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+	"syscall"
+	"unsafe"
+
+	"github.com/gliderlabs/ssh"
+	"github.com/kr/pty"
```
bradfitz, line 24: use github.com/creack/pty; that's what we already use (it's a fork of kr's)

ssh/tailssh/login1.go
```diff
@@ -0,0 +1,125 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
```
bradfitz, line 1: login1.go? what's with that filename?

ssh/tailssh/login1.go
```diff
@@ -0,0 +1,125 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build linux || (darwin && !ios)
```
bradfitz, line 5: this should be linux-only. None of this dbus stuff will actually work on darwin, even if it compiles.

ssh/tailssh/tailssh_test.go
```diff
@@ -245,8 +243,8 @@ func TestSSH(t *testing.T) {
 	}
 
 	t.Run("env", func(t *testing.T) {
-		cmd := execSSH("env")
-		cmd.Env = append(os.Environ(), "LANG=foo")
+		cmd := execSSH("LANG=foo env")
```
bradfitz, line 246: how's this work?

ssh/tailssh/child.go
```diff
@@ -0,0 +1,214 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
```
bradfitz, line 1: 2022

ssh/tailssh/tailssh.go
```diff
@@ -317,27 +286,23 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
...
-func setWinsize(f *os.File, w, h int) {
-	syscall.Syscall(syscall.SYS_IOCTL, f.Fd(), uintptr(syscall.TIOCSWINSZ),
-		uintptr(unsafe.Pointer(&struct{ h, w, x, y uint16 }{uint16(h), uint16(w), 0, 0})))
-}
-
 type sshConnInfo struct {
-	// now is the time to consider the present moment for the
-	// purposes of rule evaluation.
-	now time.Time
+	StartTime time.Time
```
bradfitz, line 290: missing docs on this one.

why are these all exported now?

bradfitz: maybe do one separate PR where those are renamed, to get diff size down?

maisem: I exported these when I was doing json marshaling to the child process, but that lead to really weird tty behavior. I can revert this bit, if you prefer

ssh/tailssh/tailssh.go
```diff
@@ -117,70 +109,77 @@ func (srv *server) sshPolicy() (_ *tailcfg.SSHPolicy, ok bool) {
 	return nil, false
 }
 
-func (srv *server) handleSSH(s ssh.Session) {
-	lb := srv.lb
+func (srv *server) evaluatePolicy(sshUser string, addr net.Addr) (*tailcfg.SSHAction, *sshConnInfo, error) {
```
bradfitz, line 112: addr of what?

this all needs comments.

ssh/tailssh/tailssh.go
```diff
@@ -208,97 +220,54 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
...
-		if err != nil {
-			logf("ssh: start error: %f", err)
-			s.Exit(1)
-			return
-		}
-		fromCmd, toCmd = stdout, stdin
-		go func() { io.Copy(s.Stderr(), stderr) }()
+	cmd.Env = append(cmd.Env, s.Environ()...)
+	logf("Running: %q, %v", cmd.Args, cmd.Env)
+	logf("Running: %q, %v", cmd.Args, s.Environ())
```
bradfitz, line 243: dup-ish log statement

go.mod
```diff
@@ -167,6 +167,7 @@ require (
 	github.com/kisielk/gotool v1.0.0 // indirect
 	github.com/kr/fs v0.1.0 // indirect
 	github.com/kr/pretty v0.3.0 // indirect
+	github.com/kr/pty v1.1.8 // indirect
```
bradfitz, line 170: creack/pty

ssh/tailssh/tailssh_test.go
maisem: i don't understand the question. We should not be passing in any local env variables over SSH, however we also want to allow users to be able to specify env variables before command they run.

ssh/tailssh/child.go
```diff
@@ -0,0 +1,246 @@
...
+	if err := cmd.Start(); err != nil {
+		return nil, nil, nil, err
+	}
+	// When using a pty we don't get a separate reader for stderr.
+	return pty, pty, nil, nil
+}
+
+func setWinsize(f *os.File, w, h int) {
+	syscall.Syscall(syscall.SYS_IOCTL, f.Fd(), uintptr(syscall.TIOCSWINSZ),
+		uintptr(unsafe.Pointer(&struct{ h, w, x, y uint16 }{uint16(h), uint16(w), 0, 0})))
```
bradfitz, line 219: https://pkg.go.dev/golang.org/x/sys/unix#Winsize

bradfitz: https://pkg.go.dev/golang.org/x/sys/unix#IoctlSetWinsize

ssh/tailssh/tailssh.go
```diff
@@ -330,14 +313,21 @@ type sshConnInfo struct {
...
+	// dst is the Tailscale IP that the connection came for.
+	dst netaddr.IPPort
 
 	// node is srcIP's node.
 	node *tailcfg.Node
 
 	// uprof is node's UserProfile.
 	uprof *tailcfg.UserProfile
+
+	// localUser is the username of the local user as configured
```
bradfitz, line 328: let's remove this

bradfitz: (this is an input struct to the policy eval)

ssh/tailssh/tailssh.go
```diff
@@ -330,14 +313,21 @@ type sshConnInfo struct {
 	// sshUser is the requested local SSH username ("root", "alice", etc).
 	sshUser string
 
-	// srcIP is the Tailscale IP that the connection came from.
-	srcIP netaddr.IP
+	// src is the Tailscale IP that the connection came from.
```
bradfitz, line 316: not just "IP" on these two

ssh/tailssh/tailssh.go
```diff
@@ -208,97 +237,56 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
...
-		}
-		fromCmd, toCmd = stdout, stdin
-		go func() { io.Copy(s.Stderr(), stderr) }()
+	cmd.Env = append(cmd.Env, s.Environ()...)
+	cmd.Env = append(cmd.Env,
+		fmt.Sprintf("SSH_CLIENT=%s %d %d", ci.src.IP(), ci.src.Port(), ci.dst.Port()),
+		fmt.Sprintf("SSH_CONNECTION=%s %d %s %d", ci.src.IP(), ci.src.Port(), ci.dst.IP(), ci.dst.Port()),
+	)
+	logf("Running: %q, %v", cmd.Args, cmd.Env)
+	stdin, stdout, stderr, err := startChild(s, cmd)
```
bradfitz, line 264: startHelper? or some name that's not child

ssh/tailssh/tailssh.go
```diff
@@ -208,97 +237,56 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
...
-		toCmd.Close()
+		stdin.Close()
 	}()
 	go func() {
-		_, err := io.Copy(s, fromCmd) // stdout
+		_, err := io.Copy(s, stdout)
 		logf("ssh: stdout copy: %v", err)
 	}()
-
+	if stderr != nil {
```
bradfitz, line 284: comment here like `// stderr is nil for ptys`?

ssh/tailssh/tailssh.go
bradfitz: can you also add a comment here about where the child (incubator?) takes over and starts running its code?

ssh/tailssh/child.go
```diff
@@ -0,0 +1,246 @@
...
+
+const debugChild = false
+
+// sshChild is the entrypoint to the `tailscaled be-child ssh` subcommand.
+// This adds a necessary layer of indirection between tailscaled and the
+// process that we are about to spawn. One of the primary requirements
+// is to inform the system of new login session from the user. This is
+// typically necessary for mounting home directories and decrypting file
+// systems.
+// It expects the following env variables:
```
bradfitz, line 72: use args?

ssh/tailssh/child.go
```diff
@@ -0,0 +1,246 @@
...
+	}()
+	cmd.Env = append(cmd.Env, "TS_HAS_TTY=1")
+	if ptyReq.Term != "" {
+		cmd.Env = append(cmd.Env, fmt.Sprintf("TERM=%s", ptyReq.Term))
+	}
+	cmd.Stdin = tty
+	cmd.Stdout = tty
+	cmd.Stderr = tty
+
+	go func() {
```
bradfitz, line 205: goroutine dumps look better when goroutines have names. "go resizeSSHChild(...)" or something?

ssh/tailssh/child.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	if ptyPath, err := ptyName(ptyFile); err == nil {
+		cmd.Env = append(cmd.Env, "TS_TTY_NAME="+ptyPath)
+	}
+
+	return ptyFile, tty, nil
+}
+
+func startChild(s ssh.Session, cmd *exec.Cmd) (stdin io.WriteCloser, stdout, stderr io.Reader, err error) {
```
bradfitz, line 181: clarify in a comment that cmd is the "tailscaled be-child ssh ..."

ssh/tailssh/child.go
```diff
@@ -0,0 +1,246 @@
...
+		os.Exit(1)
+	}
+
+	euid := uint64(os.Geteuid())
+	if euid == 0 {
+		// Inform the system that we are about to login someone.
+		// We can only do this if we are running as root.
+		remoteUser := os.Getenv("TS_REMOTE_TAGS")
+		if remoteUser == "" {
+			remoteUser = os.Getenv("TS_REMOTE_USER")
```
bradfitz, line 105: using env vars for this scares me for security reasons.

people can ssh -E and send malicious stuff here maybe?

ssh/tailssh/incubator.go
```diff
@@ -0,0 +1,281 @@
...
+	if len(args) > 0 {
+		incubatorArgs = append(incubatorArgs, fmt.Sprintf("--cmd-args=%q", strings.Join(args, " ")))
+	}
+	return exec.CommandContext(ctx, tailscaled, incubatorArgs...)
+}
+
+const debugChild = false
+
+// beIncubator is the entrypoint to the `tailscaled be-child ssh` subcommand.
+// It is responsible for informinf the system of a new login session for the user.
```
bradfitz, line 74: typo

ssh/tailssh/incubator.go
```diff
@@ -0,0 +1,281 @@
...
+// This is typically necessary for mounting home directories and decrypting file
```
bradfitz, line 75: sometimes necessary

ssh/tailssh/incubator_linux.go
```diff
@@ -0,0 +1,170 @@
...
+func ptyNameLinux(f *os.File) (string, error) {
+	var n uint32
+	_, _, e := syscall.Syscall(syscall.SYS_IOCTL, f.Fd(), syscall.TIOCGPTN, uintptr(unsafe.Pointer(&n)))
+	if e != 0 {
+		return "", e
+	}
+	return fmt.Sprintf("pts/%d", n), nil
+}
+
+// callLogin1 invokes the provided method over D-Bus.
```
bradfitz, line 36: .... provided method of the "login1" service over ...

(to explain why this method is named as such)

ssh/tailssh/tailssh.go
```diff
@@ -198,6 +212,19 @@ func (srv *server) handleSSH(s ssh.Session) {
...
+	<-ctx.Done()
+	err := ctx.Err()
+	if serr, ok := err.(SSHTerminationError); ok {
+		msg := serr.SSHTerminationMessage()
+		if msg != "" {
+			io.WriteString(s.Stderr(), "\r\n\r\n"+msg+"\r\n\r\n")
+		}
+	}
+	srv.logf("terminating SSH session from %v: %v", ci.src.IP(), err)
+	cmd.Process.Kill()
```
bradfitz, line 225: It'd be nice to avoid this on Linux if it's already exited nicely on its own. Because on Linux (until `pidfd`: https://lwn.net/Articles/801319/) the Kill here can kill an unrelated process in the future, as I don't think Go yet uses pidfds, so if this pid recycles, it could be a different process.

Instead of passing in a cmd, can you pass in a `killProc func()` and then have the caller pass in a func that's disarmed (perhaps via a sync.Once shared between this goroutine and the post-Wait cleanup).

Then document this method too.

ssh/tailssh/tailssh.go
```diff
@@ -208,97 +235,42 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
...
-		_, err := io.Copy(s, fromCmd) // stdout
+		_, err := io.Copy(s, stdout)
 		logf("ssh: stdout copy: %v", err)
 	}()
-
+	// stderr is nil for ptys.
+	if stderr != nil {
+		go func() {
+			_, err := io.Copy(s.Stderr(), stderr)
+			logf("ssh: stderr copy: %v", err)
```
bradfitz, line 271: TODO on all three of these to avoid these logf calls on normal termination?

ssh/tailssh/incubator.go
```diff
@@ -0,0 +1,281 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
```
bradfitz, line 4: Maybe a comment up here explaining what an "incubator" is?  (the middle of the three processes, running as root?)

ssh/tailssh/incubator.go
```diff
@@ -0,0 +1,281 @@
...
+		hasTTY     = flags.Bool("has-tty", false, "is the output attached to a tty")
+		cmdName    = flags.String("cmd", "", "the cmd to launch")
+		cmdArgs    = flags.String("cmd-args", "", "the args for cmd")
+	)
+	if err := flags.Parse(args); err != nil {
+		return err
+	}
+	logf := logger.Discard
+	if debugChild {
+		// The only place we can log is syslog.
```
bradfitz, line 98: // We don't own stdout or stderr, so the ...

ssh/tailssh/incubator.go
```diff
@@ -0,0 +1,281 @@
...
+		if sl, err := syslog.New(syslog.LOG_INFO|syslog.LOG_DAEMON, "tailscaled-ssh"); err == nil {
+			logf = log.New(sl, "", 0).Printf
+		}
+	}
+
+	euid := uint64(os.Geteuid())
+	if euid == 0 {
+		// Inform the system that we are about to log someone in.
+		// We can only do this if we are running as root.
```
bradfitz, line 107: On Linux, perhaps. But maybe some system/environment in the future won't have this requirement.

Maybe this euid check could also be in startLoginSession.

But we can do that later if/when needed.

ssh/tailssh/incubator.go
```diff
@@ -0,0 +1,281 @@
...
+		closer, err := startLoginSession(logf, uint32(*uid), *localUser, *remoteUser, *remoteIP, *ttyName)
+		if err == nil && closer != nil {
+			defer closer()
+		}
+	}
+	if euid != *uid {
+		// Switch users if required before starting the desired process.
+		if err := syscall.Setuid(int(*uid)); err != nil {
```
bradfitz, line 115: I was going to warn you that this doesn't work, but then I remember that this finally landed in Go 1.16 after like 10 years of discussion.

Previously it was only per-thread. Now we have https://pkg.go.dev/syscall#AllThreadsSyscall which Setuid uses.

But be warned:

> AllThreadsSyscall is unaware of any threads that are launched explicitly by cgo linked code, so the function always returns ENOTSUP in binaries that use cgo.

That's a pretty big disclaimer.

Maybe we should fail earlier with a nice error message if the "cgo" build tag is present?

bradfitz (approved):

ssh/tailssh/incubator.go
maisem: As discussed on slack, this isn't an issue. syscall.Setuid already has a cgo variant. 
https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/syscall/syscall_linux.go;l=1205-1207

ssh/tailssh/incubator.go
maisem: Moved to maybeStartLoginSession

ssh/tailssh/tailssh.go
maisem: I did another way, PTAL

ssh/tailssh/incubator.go
```diff
@@ -0,0 +1,288 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// This file contains the code for the incubator process.
+// Taiscaled launches the incubator as the same user as it was launched as.
+// The incbuator then registers a new session with the OS, sets it's own UID to
```
bradfitz, line 7: its

ssh/tailssh/tailssh.go
```diff
@@ -117,70 +112,90 @@ func (srv *server) sshPolicy() (_ *tailcfg.SSHPolicy, ok bool) {
...
-		logf("tsshd: rejecting non-Tailscale addr %v", ta.IP)
-		s.Exit(1)
-		return
+		return netaddr.IPPort{}, fmt.Errorf("non-Tailscale addr %v", ta.IP)
 	}
+	return netaddr.IPPortFrom(tanetaddr, uint16(ta.Port)), nil
+}
+
+// evaluatePolicy returns the SSHAction after evaluating the sshUser and
+// remoteAddr against the SSHPolicy. remoteAddr must be a Tailscale IP.
```
bradfitz, line 131: localAddr too?

ssh/tailssh/tailssh.go
```diff
@@ -117,70 +112,90 @@ func (srv *server) sshPolicy() (_ *tailcfg.SSHPolicy, ok bool) {
...
+func (srv *server) evaluatePolicy(sshUser string, localAddr, remoteAddr net.Addr) (*tailcfg.SSHAction, *sshConnInfo, string, error) {
```
bradfitz, line 132: um what's that third string doing in there? surprise! a string!

ssh/tailssh/tailssh.go
```diff
@@ -208,98 +240,63 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
...
+		defer cancel()
+		go srv.handleSessionTermination(ctx, s, ci, cmd, exitOnce)
+	}
+	go func() {
+		_, err := io.Copy(stdin, s)
 		if err != nil {
-			logf("running shell: %v", err)
-			s.Exit(1)
-			return
+			// TODO: don't log in the success case.
```
bradfitz, line 270: isn't this done now?

ssh/tailssh/tailssh.go
```diff
@@ -208,98 +240,63 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
...
-		stdin, stdout, stderr, err := startWithStdPipes(cmd)
+		stdin.Close()
+	}()
+	go func() {
+		_, err := io.Copy(s, stdout)
 		if err != nil {
-			logf("ssh: start error: %f", err)
-			s.Exit(1)
-			return
+			// TODO: don't log in the success case.
```
bradfitz, line 278: and here

bradfitz (approved):
LGTM after comments addressed.

ssh/tailssh/tailssh.go
```diff
@@ -208,98 +240,63 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
...
+	if err != nil {
+		logf("start failed: %v", err.Error())
+		s.Exit(1)
+		return
+	}
+	// We use this sync.Once to ensure that we only terminate the process once,
+	// either it exits itself or is terminated
+	var exitOnce *sync.Once
+	if ctx.Done() != nil {
+		exitOnce = &sync.Once{}
```
bradfitz, line 262: I'd just make Line 260 a var and use &exitOnce in places, but sure, this works too.

ssh/tailssh/tailssh.go
maisem: There's an EIO coming from somewhere and I want to chase it down

ssh/tailssh/tailssh.go
```diff
@@ -208,98 +240,63 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
...
-		logf("ssh: stdin copy: %v", err)
-		toCmd.Close()
-	}()
-	go func() {
-		_, err := io.Copy(s, fromCmd) // stdout
-		logf("ssh: stdout copy: %v", err)
-	}()
-
 	err = cmd.Wait()
+	if exitOnce != nil {
```
bradfitz, line 293: ... then you wouldn't need this.

bradfitz: (I mean the if statement. Still need to disarm it with the func(){}

PR #4030, opened 2022-02-24
ssh/tailssh: add start of real ssh tests

bradfitz:
Updates #3802

### DIFF
--- a/ssh/tailssh/tailssh_test.go
+++ b/ssh/tailssh/tailssh_test.go
@@ -8,11 +8,24 @@
 package tailssh
 
 import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"os/exec"
+	"os/user"
 	"testing"
 	"time"
 
+	"github.com/gliderlabs/ssh"
 	"inet.af/netaddr"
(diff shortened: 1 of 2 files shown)


ssh/tailssh/tailssh_test.go
```diff
@@ -155,3 +168,75 @@ func TestMatchRule(t *testing.T) {
...
+		uprof:   &tailcfg.UserProfile{},
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	ss.Handler = func(s ssh.Session) {
+		srv.handleAcceptedSSH(ctx, s, ci, u)
+	}
+
+	ln, err := net.Listen("tcp4", "127.0.0.1:0")
```
maisem, line 214: you can use https://pkg.go.dev/tailscale.com/net/nettest#Listener then you don't have to do the whole port dance.

maisem: oh, oops. I thought you were calling from within the process. You're exec-ing out to `ssh`. This is fine.

maisem (approved):

PR #4031, opened 2022-02-24
cmd/tailscaled/childproc: add be-child registration mechanism

bradfitz:
For ssh and maybe windows service babysitter later.

Updates #3802


cmd/tailscaled/childproc/childproc.go
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package childproc allows other packages to register be-child process
```
danderson, line 5: Could you add a reference to the top-level command you mean? It took me a later function docstring to realize that "be-child" meant specifically "a subcommand of tailscaled"

danderson (approved):

cmd/tailscaled/childproc/childproc.go
bradfitz: read the next line? it says "with tailscaled" ?

bradfitz: reworded

Issue #3488, opened 2021-12-03
Implement 2021 Control Protocol


danderson:
We have a design for a new control protocol that uses Noise and other nice things. This is the tracking bug to wire it up and transition the client over to it incrementally.

danderson:
https://github.com/tailscale/tailscale/pull/2261 was the initial chunk of code for this, with more to follow.

ericlagergren:
How does this relate to https://tailscale.com/blog/tailscale-key-management/

Is there a white paper or design document?

danderson:
It's a revised version of what's described in that post. Uses a different base layer transport, and lays the groundwork for some nice security features (but that aren't going to materialize immediately, this bug is about getting the basics online). The design's not public yet, but will be once this is something the clients actually use.

DentonGentry:
The ts2021 control protocol design shipped with client version 1.24.0 and is in use by a majority of the device fleet.

I'm going to mark the issue as fixed. We've mostly been using it as a reference for PRs relating to the feature, and we can continue to do so for further work even after it is closed.

ericlagergren:
> The design's not public yet, but will be once this is something the clients actually use.

Since it has shipped does this mean a white paper or design document is available? 😃

frankli0324:
I also landed here looking for some material on the new control protocol.

PR #4035, opened 2022-02-26
controlhttp: add some docs, change Dial's path from /switch to /ts2021

bradfitz:
When I deployed server-side changes, I put the upgrade handler at /ts2021
instead of /switch. We could move the server to /switch, but ts2021 seems
more specific and better, but I don't feel strongly.

Updates #3488

### DIFF
--- a/control/controlhttp/client.go
+++ b/control/controlhttp/client.go
@@ -40,11 +40,19 @@ import (
 	"tailscale.com/types/key"
 )
 
-// upgradeHeader is the value of the Upgrade HTTP header used to
-// indicate the Tailscale control protocol.
 const (
-	upgradeHeaderValue  = "tailscale-control-protocol"
+	// upgradeHeader is the value of the Upgrade HTTP header used to
+	// indicate the Tailscale control protocol.
+	upgradeHeaderValue = "tailscale-control-protocol"
+
+	// handshakeHeaderName is the HTTP request header that can
+	// optionally contain base64-encoded initial handshake
+	// payload, to save an RTT.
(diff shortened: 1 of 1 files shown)


control/controlhttp/client.go
```diff
@@ -40,11 +40,19 @@ import (
 	"tailscale.com/types/key"
 )
 
-// upgradeHeader is the value of the Upgrade HTTP header used to
-// indicate the Tailscale control protocol.
 const (
-	upgradeHeaderValue  = "tailscale-control-protocol"
+	// upgradeHeader is the value of the Upgrade HTTP header used to
+	// indicate the Tailscale control protocol.
+	upgradeHeaderValue = "tailscale-control-protocol"
```
danderson, line 46: Any thoughts on varying this header value, rather than the switch URL?

I have no strong feelings, just occurred to me that there's slight redundancy between the URL and the header now, if the URL also specifies the exact protocol being switched to.

control/controlhttp/client.go
```diff
@@ -53,6 +61,9 @@ const (
 //
 // If Dial fails to connect using addr, it also tries to tunnel over
 // TLS to <addr's host>:443 as a compatibility fallback.
+//
+// The provided ctx is only used for the initial connection, until
+// Dial returns. It does not affect the connection once established.
```
danderson, line 66: Could you throw a TODO in here to do something with conn read/write timeouts? I think I neglected to plumb the ctx deadline all the way into the socket, so there are some ways this package (and possibly also controlbase?) can be slowlorised at the moment.

PR #4037, opened 2022-02-27
ipn/store: add common package for instantiating ipn.StateStores

maisem:
ipn/store: add common package for instantiating ipn.StateStores
Also move KubeStore and MemStore into their own package.

RELNOTE: tsnet now supports providing a custom ipn.StateStore.

cc @rosszurowski 


bradfitz (commented):
D:\OtherDisk\Tailscale has a colon.

ipn/ipnlocal/local.go
```diff
@@ -2166,6 +2166,10 @@ func (b *LocalBackend) SetVarRoot(dir string) {
 // It returns an empty string if there's no configured or discovered
 // location.
 func (b *LocalBackend) TailscaleVarRoot() string {
+	if strings.ContainsRune(b.varRoot, ':') {
+		// Not writable.
```
crawshaw, line 2170: Worth adding an example of what is being detected here.

crawshaw (approved):
Add a test?

ipn/store.go
```diff
@@ -14,14 +14,66 @@ import (
...
 	"sync"
 	"time"
 
 	"tailscale.com/atomicfile"
 	"tailscale.com/kube"
+	_ "tailscale.com/kube"
 	"tailscale.com/paths"
 )
 
+// NewStoreFunc returns a StateStore for the provided path.
```
bradfitz, line 27: clarify that the path is the part after the colon?

ipn/store.go
```diff
@@ -14,14 +14,66 @@ import (
...
+	"file:": func(path string) (StateStore, error) {
+		path = strings.TrimPrefix(path, "file:")
+		return NewFileStore(path)
+	},
+}
+
+// ResolveStore returns a StateStore based on the provided path
+// and registered stores. It returns nil when there is no matching
+// path.
+func ResolveStore(path string) (StateStore, error) {
```
bradfitz, line 47: why "Resolve"? Why not NewStore?

ipn/store.go
```diff
@@ -14,14 +14,66 @@ import (
...
+		return new(MemoryStore), nil
+	},
+	"file:": func(path string) (StateStore, error) {
+		path = strings.TrimPrefix(path, "file:")
+		return NewFileStore(path)
+	},
+}
+
+// ResolveStore returns a StateStore based on the provided path
+// and registered stores. It returns nil when there is no matching
```
bradfitz, line 45: what part is nil? (nil, nil)? Why is that a good result? That seems bogus in like all cases.

ipn/store.go
maisem: That's dependent on the implementation. Added comments in `RegisterStore` and `NewStore`

ipn/store.go
maisem: Wasn't ready for re-review, renamed to `NewStore`.

ipn/store.go
maisem: Returning an error now.

tsnet/tsnet.go
```diff
@@ -108,21 +112,28 @@ func (s *Server) start() error {
...
 	}
 
-	s.dir = s.Dir
-	if s.dir == "" {
+	s.rootPath = s.Dir
+	if strings.HasPrefix(s.rootPath, "mem:") && !s.Ephemeral {
+		return fmt.Errorf("in-memory store is only supported for Ephemeral nodes")
+	}
+	switch {
+	case strings.ContainsRune(s.rootPath, ':'):
```
bradfitz, line 120: Just Contains, not ContainsRune.

But this needs a comment and I didn't see an answer to my Windows concern.

ipn/store.go
```diff
@@ -14,14 +14,73 @@ import (
...
+			return sf(path)
+		}
+	}
+	return nil, fmt.Errorf("%v: %w", path, ErrNoMatchingStore)
+}
+
+// RegisterStore registers a prefix to be used for
+// NewStore. It panics if the prefix is empty, or has
+// already been defined.
+// The provided fn is called with the path passed to NewStore,
```
bradfitz, line 64: semicolon, not colon

ipn/store.go
```diff
@@ -14,14 +14,73 @@ import (
...
+// the prefix is not stripped.
+// By default the following stores are registered:
+//
+//   * if the string begins with "mem:", the suffix
+//     is ignored and an in-memory store is used.
+//   * if the string begins with "kube:", the suffix
```
bradfitz, line 70: This isn't "by default". In the other file in this PR you had to blank import it to get the side effect of its registration?

ipn/ipnserver/server.go
```diff
@@ -696,34 +695,15 @@ func StateStore(path string, logf logger.Logf) (ipn.StateStore, error) {
 	if path == "" {
```
bradfitz, line 695: Line 685: isn't "file:" also a special case now? It's optional? Only used for Windows?

bradfitz (changes requested):
(earlier comments)

ipn/store.go
maisem: These are the default ones (are in the same package), I had to blank import the aws one.

tsnet/tsnet.go
maisem: I was adding tests so missed this bit, removed this. This was the last place left where I introduced it, and now it's also removed.

ipn/ipnlocal/local.go
maisem: Removed this, added tests for the generic prefix based lookup mechanism.

ipn/ipnserver/server.go
maisem: Removed those from here and pointed to ipn.Store.

ipn/ipnserver/server.go
```diff
@@ -681,49 +680,22 @@ func tryWindowsAppDataMigration(logf logger.Logf, path string) string {
...
-//   * empty string means to use an in-memory store
-//   * if the string begins with "mem:", the suffix
-//     is ignored and an in-memory store is used.
-//   * if the string begins with "kube:", the suffix
-//     is a Kubernetes secret name
-//   * if the string begins with "arn:", the value is
-//     an AWS ARN for an SSM.
+// The path should correspond to a registered store in
+// ipn.NewStore.
+// An empty path means to use an in-memory store
```
bradfitz, line 685: missing period.

ipn/ipnserver/server.go
```diff
@@ -681,49 +680,22 @@ func tryWindowsAppDataMigration(logf logger.Logf, path string) string {
...
-//
-//   * empty string means to use an in-memory store
-//   * if the string begins with "mem:", the suffix
-//     is ignored and an in-memory store is used.
-//   * if the string begins with "kube:", the suffix
-//     is a Kubernetes secret name
-//   * if the string begins with "arn:", the value is
-//     an AWS ARN for an SSM.
+// The path should correspond to a registered store in
+// ipn.NewStore.
```
bradfitz, line 684: Restore some of the deleted docs so people have some examples of how to use this. You can refer to something else for more details.

ipn/store.go
bradfitz: These docs are all too indirect.

They need to be fleshed out to be more useful without chasing a few levels of indirection to find the content.

ipn/store.go
```diff
@@ -14,14 +14,79 @@ import (
...
 	"tailscale.com/kube"
+	_ "tailscale.com/kube"
 	"tailscale.com/paths"
 )
 
+// NewStoreFunc returns a StateStore for the provided path.
+type NewStoreFunc func(path string) (StateStore, error)
+
+func init() {
+	registerDefaultStores()
```
bradfitz, line 31: do this via a sync.Once and not in func init, to not slow down start-up for e.g. the tailscaled be-child case.

cmd/tailscaled/tailscaled.go
```diff
@@ -55,6 +55,9 @@ import (
 	"tailscale.com/wgengine/monitor"
 	"tailscale.com/wgengine/netstack"
 	"tailscale.com/wgengine/router"
+
+	// Initialize the aws store.
+	_ "tailscale.com/ipn/store/aws"
```
bradfitz, line 60: why is kube and aws not on equal footing? they should probably both require importing.

but if you use tsnet and forget one, the built-in package should know about their prefixes enough to tell you a useful error like, "Yo, you forgot to import _ "..../kubee""

bradfitz (changes requested):
Can you flesh out the docs? It seems too vague & indirect.

Some concrete examples would help for at least a few of the cases, even if it's redundant.

ipn/ipnserver/server.go
maisem: I removed this entire function to remove the confusion.

ipn/store.go
maisem: Done.

cmd/tailscaled/tailscaled.go
maisem: I moved KubeStore to it's own package, and added add a generic `tailscale.com/ipn/store` package to instantiate all the stores so that it's all in one place.

ipn/store/stores.go
```diff
@@ -0,0 +1,245 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package store provides various implementation
```
bradfitz, line 5: cut off sentence

bradfitz (commented):
package state?
type Store interface?

state.Store ?
state.MemoryStore ?

maisem:
> package state?
> type Store interface?
> state.Store ?

The interface needs to be in a different package from the lookup stuff to not cause a circular dependency. Otherwise we would need to go back to the `_` imports.

cmd/tailscaled/depaware.txt
```diff
@@ -184,8 +184,11 @@ tailscale.com/cmd/tailscaled dependencies: (generated by github.com/tailscale/de
         tailscale.com/ipn/ipnstate                                   from tailscale.com/client/tailscale+
         tailscale.com/ipn/localapi                                   from tailscale.com/ipn/ipnserver
         tailscale.com/ipn/policy                                     from tailscale.com/ipn/ipnlocal
-        tailscale.com/ipn/store/aws                                  from tailscale.com/ipn/ipnserver
-        tailscale.com/kube                                           from tailscale.com/ipn
+        tailscale.com/ipn/store                                      from tailscale.com/cmd/tailscaled
+        tailscale.com/ipn/store/aws                                  from tailscale.com/ipn/store
```
bradfitz, line 188: Seems like a waste of names "aws" and "kube" for these little packages.

Maybe awsstore and kubestore ?

cmd/tailscaled/tailscaled_windows.go
```diff
@@ -335,8 +337,12 @@ func startIPNServer(ctx context.Context, logid string) error {
 			return nil, fmt.Errorf("%w\n\nlogid: %v", res.Err, logid)
 		}
 	}
-
-	store, err := ipnserver.StateStore(statePathOrDefault(), logf)
+	statePath := store.TryWindowsAppDataMigration(logf, statePathOrDefault())
+	st, err := store.New(statePath)
```
bradfitz, line 341: "sto" instead of "st" which we use for "state" in some places?

ipn/ipnlocal/local_test.go
```diff
@@ -457,7 +458,7 @@ func TestLazyMachineKeyGeneration(t *testing.T) {
 	panicOnMachineKeyGeneration = true
 
 	var logf logger.Logf = logger.Discard
-	store := new(ipn.MemoryStore)
+	store := new(mem.Store)
```
bradfitz, line 461: in all these other places we use "store" as the variable name, so it's probably fine to use in tailscaled_windows.go even if you shadow the package name for a few lines

ipn/store/kube/store_kube.go
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package kube contains an Kubernetes Secret based StateStore implementation.
```
bradfitz, line 5: ... contains a StateStore implementation using Kubernetes Secrets.

ipn/store/kube/store_kube.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"context"
+	"time"
+
+	"tailscale.com/ipn"
+	"tailscale.com/kube"
+)
+
+// Store is a StateStore that uses a Kubernetes Secret for persistence.
```
bradfitz, line 16: is an ipn.StateStore

ipn/store/stores.go
```diff
@@ -0,0 +1,200 @@
...
+//   * if the string begins with "mem:", the suffix
+//     is ignored and an in-memory store is used.
+//   * if the string begins with "file:", the suffix
+//     is a file name.
+//   * if the string begins with "arn:", the suffix
+//     an AWS ARN for an SSM.
+//   * if the string begins with "kube:", the suffix
+//     is a Kubernetes secret name
+func New(path string) (ipn.StateStore, error) {
+	regOnce.Do(func() {
```
bradfitz, line 68: regOnce.Do(registerDefaultStores)

ipn/store/stores.go
```diff
@@ -0,0 +1,200 @@
...
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"tailscale.com/atomicfile"
+	"tailscale.com/ipn"
+	"tailscale.com/ipn/store/aws"
```
bradfitz, line 22: This shouldn't be linked on iOS. Nor Kube.

ipn/store/stores.go
```diff
@@ -0,0 +1,200 @@
...
+// registered store.
+var ErrNoMatchingStore = errors.New("no matching store")
+
+// New returns a StateStore based on the provided path
+// and registered stores.
+// By default the following stores are registered:
+//
+//   * if the string begins with "mem:", the suffix
+//     is ignored and an in-memory store is used.
+//   * if the string begins with "file:", the suffix
```
bradfitz, line 61: Clarify the relation to NewFileStore here? With the New constructor you need the "file:" prefix but the other one you don't?

maisem: I made NewFileStore strip out `file:` if provided.

ipn/store/stores.go
maisem: Done

ipn/store/stores.go
maisem: Done

ipn/ipnlocal/local_test.go
maisem: I can't do that because we need to try `store.NewFileStore` if `store.New` fails with `ErrNoMatchingStore`

ipn/store/awsstore/store_aws.go
```diff
@@ -5,8 +5,8 @@
 //go:build linux
 // +build linux
 
-// Package aws contains an AWS SSM StateStore implementation.
-package aws
+// Package awsstore contains an AWS SSM StateStore implementation.
```
bradfitz, line 8: implementation of ipn.StateStore.

ipn/store/mem/store_mem.go
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package mem provides an in-memory StateStore implementation.
```
bradfitz, line 5: implementation of ipn.StateStore.

ipn/store/mem/store_mem.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"tailscale.com/ipn"
+)
+
+// New returns a new Store.
+func New(string) (ipn.StateStore, error) {
+	return new(Store), nil
+}
+
+// Store is a store that keeps state in memory only.
```
bradfitz, line 20: in an ipn.StateStore that keeps

ipn/store/stores.go
```diff
@@ -0,0 +1,196 @@
...
+// and registered stores.
+// It will return ErrNoMatchingStore if the path is not
+// prefixed with a known prefix.
+// By default the following stores are registered:
+//
+//   * if the string begins with "mem:", the suffix
+//     is ignored and an in-memory store is used.
+//   * if the string begins with "file:", the suffix
+//     is a file name.
+//   * if the string begins with "arn:", the suffix
```
bradfitz, line 59: (Linux-only) in parens after this and next one?

ipn/store/stores_linux.go
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package store provides various implementation of ipn.StateStore.
```
bradfitz, line 5: delete broken-looking package doc. (it's already in the portable non-linux file and here you have a bogus newline after)

ipn/store/stores_test.go
```diff
@@ -1,24 +1,84 @@
...
 	"testing"
 
+	"tailscale.com/ipn"
+	"tailscale.com/ipn/store/mem"
 	"tailscale.com/tstest"
 )
 
-func testStoreSemantics(t *testing.T, store StateStore) {
+func TestNewStore(t *testing.T) {
+	regOnce.Do(func() {
```
bradfitz, line 18: regOnce.Do(registerDefaultStores)

ipn/store/stores.go
```diff
@@ -0,0 +1,189 @@
...
+//   * (Linux-only) if the string begins with "kube:",
+//     the suffix is a Kubernetes secret name
+//   * In all other cases, the path is treated as a filepath.
+func New(logf logger.Logf, path string) (ipn.StateStore, error) {
+	regOnce.Do(registerDefaultStores)
+	for prefix, sf := range knownStores {
+		if strings.HasPrefix(path, prefix) {
+			// We can't strip the prefix here as some NewStoreFunc (like arn:)
+			// expect the prefix.
+			return sf(path)
```
bradfitz, line 61: should we pass the logger.Logf down to the Provider?

ipn/store/stores.go
```diff
@@ -0,0 +1,189 @@
...
+
+	"tailscale.com/atomicfile"
+	"tailscale.com/ipn"
+	"tailscale.com/ipn/store/mem"
+	"tailscale.com/paths"
+	"tailscale.com/types/logger"
+)
+
+// Provider returns a StateStore for the provided path.
+type Provider func(path string) (ipn.StateStore, error)
```
bradfitz, line 28: instead of "path", how about "arg"? Then document here that arg includes the colon part.

// The arg is of the form "prefix:rest", where prefix was previously registered with Register.

bradfitz (approved):

ipn/store/stores.go
maisem: Good idea, done.

ipn/store/stores.go
maisem: Done.

Issue #4038, opened 2022-02-27
UDM Pro requires fwmark rules on table 201


jtcressy:
### What is the issue?

I happen to be running tailscale on a device that puts the default route on another table (table 201) instead of putting it on the main table. Specifically, i'm running on a unifi UDM pro.
If I manually re-add the default route to the main table, tailscaled is able to connect to the control servers just fine.

I'm on version 1.20.4

Below is some relevant log output:
```
2022/02/27 19:45:28 trying bootstrapDNS("derp8d.tailscale.com", "178.62.44.132") for "controlplane.tailscale.com" ...
2022/02/27 19:45:28 bootstrapDNS("derp8d.tailscale.com", "178.62.44.132") for "controlplane.tailscale.com" error: Get "https://derp8d.tailscale.com/bootstrap-dns?q=controlplane.tailscale.com": dial tcp 178.62.44.132:443: connect: network is unreachable
2022/02/27 19:45:28 trying bootstrapDNS("derp5.tailscale.com", "2001:19f0:5801:10b7:5400:2ff:feaa:284c") for "controlplane.tailscale.com" ...
2022/02/27 19:45:28 bootstrapDNS("derp5.tailscale.com", "2001:19f0:5801:10b7:5400:2ff:feaa:284c") for "controlplane.tailscale.com" error: Get "https://derp5.tailscale.com/bootstrap-dns?q=controlplane.tailscale.com": dial tcp [2001:19f0:5801:10b7:5400:2ff:feaa:284c]:443: connect: network is unreachable
2022/02/27 19:45:28 trying bootstrapDNS("derp6.tailscale.com", "68.183.90.120") for "controlplane.tailscale.com" ...
2022/02/27 19:45:28 bootstrapDNS("derp6.tailscale.com", "68.183.90.120") for "controlplane.tailscale.com" error: Get "https://derp6.tailscale.com/bootstrap-dns?q=controlplane.tailscale.com": dial tcp 68.183.90.120:443: connect: network is unreachable
2022/02/27 19:45:28 trying bootstrapDNS("derp8b.tailscale.com", "2a03:b0c0:1:d0::ec1:e001") for "controlplane.tailscale.com" ...
[...]

PR #4042, opened 2022-02-28
net/interfaces: get Linux default route from netlink as fallback

bradfitz:
If it's in a non-standard table, as it is on Unifi UDM Pro, apparently.

Updates #4038 (probably fixes, but don't have hardware to verify)


net/interfaces/interfaces_linux.go
```diff
@@ -145,7 +147,54 @@ func defaultRoute() (d DefaultRouteDetails, err error) {
...
+	if err != nil {
+		return d, err
+	}
+	defer c.Close()
+	rms, err := c.Route.List()
+	if err != nil {
+		return d, err
+	}
+	for _, rm := range rms {
+		if rm.Attributes.Gateway == nil || rm.Attributes.Src != nil || rm.Attributes.Dst != nil {
```
maisem, line 181: Add a comment why we're ignoring this.

net/interfaces/interfaces_linux.go
```diff
@@ -145,7 +147,54 @@ func defaultRoute() (d DefaultRouteDetails, err error) {
...
+	// as a fallback for weird environments where netlink might be
+	// banned but /proc/net/route is emulated (e.g. stuff like
+	// Cloud Run?).
+	return defaultRouteFromNetlink()
+}
+
+func defaultRouteFromNetlink() (d DefaultRouteDetails, err error) {
+	c, err := rtnetlink.Dial(&netlink.Config{Strict: true})
+	if err != nil {
+		return d, err
```
maisem, line 173: Can we wrap these errors so that we know they're coming from the new netlink code?

net/interfaces/interfaces_linux.go
```diff
@@ -145,7 +147,54 @@ func defaultRoute() (d DefaultRouteDetails, err error) {
...
+	//
+	// Then we should just always try to use the cached route
+	// table from netlink every time, and only use /proc/net/route
+	// as a fallback for weird environments where netlink might be
+	// banned but /proc/net/route is emulated (e.g. stuff like
+	// Cloud Run?).
+	return defaultRouteFromNetlink()
+}
+
+func defaultRouteFromNetlink() (d DefaultRouteDetails, err error) {
```
maisem, line 170: Document d can be nil?

bradfitz: It can't because it's a struct.

maisem: can be zero.

maisem: also fix struct comment pls
https://github.com/tailscale/tailscale/blob/v1.22.0/net/interfaces/interfaces.go#L690

maisem (approved):

bradfitz:
Sent #4045 for the Windows flakes.

PR #4046, opened 2022-03-01
cmd/derpprobe: don't alert for smaller failures.

DentonGentry:
There is a Cosmic Background level of DERP Unreachability,
with individual nodes or regions becoming unreachable briefly
and returning a short time later.

Returning a 500 error pages a human. Being awoken at 3am for
a transient error is annoying.

For relatively small levels of badness don't page a human,
just post to Slack. If the outage impacts a significant fraction
of the DERP fleet, then page a human.

### DIFF
--- a/cmd/derpprobe/derpprobe.go
+++ b/cmd/derpprobe/derpprobe.go
@@ -19,7 +19,9 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"os"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -54,7 +56,15 @@ var (
 
 func main() {
 	flag.Parse()
+
+	// proactively load the DERP map. Nothing terrible happens if this fails, so we ignore
+	// the error. The Slack bot will print a notification that the DERP map was empty.
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	_, _ = getDERPMap(ctx)
+
 	go probeLoop()
+	go slackLoop()
 	log.Fatal(http.ListenAndServe(*listen, http.HandlerFunc(serve)))
 }
 
@@ -138,10 +148,14 @@ func getOverallStatus() (o overallStatus) {
 func serve(w http.ResponseWriter, r *http.Request) {
 	st := getOverallStatus()
 	summary := "All good"
-	if len(st.bad) > 0 {
+	if (float64(len(st.bad)) / float64(len(st.bad)+len(st.good))) > 0.25 {
+		// This will generate an alert and page a human.
+		// It also ends up in Slack, but as part of the alert handling pipeline not
+		// because we generated a Slack notification from here.
 		w.WriteHeader(500)
 		summary = fmt.Sprintf("%d problems", len(st.bad))
 	}
+
 	io.WriteString(w, "<html><head><style>.bad { font-weight: bold; color: #700; }</style></head>\n")
 	fmt.Fprintf(w, "<body><h1>derp probe</h1>\n%s:<ul>", summary)
 	for _, s := range st.bad {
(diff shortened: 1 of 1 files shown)


cmd/derpprobe/derpprobe.go
```diff
@@ -50,6 +52,7 @@ var (
 	lastDERPMap   *tailcfg.DERPMap
 	lastDERPMapAt time.Time
 	certs         = map[string]*x509.Certificate{}
+	reportedBad   bool
```
bradfitz, line 55: this makes it sound like a one-time false=>true variable. Maybe "inBadState"? Or at least a comment like `// whether we're currently in report-worthy failure state`

cmd/derpprobe/derpprobe.go
```diff
@@ -135,13 +138,71 @@ func getOverallStatus() (o overallStatus) {
...
+	if err != nil {
+		return err
+	}
+
+	webhookUrl := os.Getenv("SLACK_WEBHOOK")
+	if webhookUrl == "" {
+		return errors.New("No SLACK_WEBHOOK configured")
+	}
+
+	req, err := http.NewRequest(http.MethodPost, webhookUrl, bytes.NewBuffer(slackBody))
```
bradfitz, line 156: just "POST" is preferred over the const usually

bradfitz: and bytes.NewReader is a little cheaper & minimal for what you need here

cmd/derpprobe/derpprobe.go
```diff
@@ -135,13 +138,71 @@ func getOverallStatus() (o overallStatus) {
...
+		} else {
+			mu.Lock()
+			reportedBad = true
+			mu.Unlock()
+		}
+	}
+
+	mu.Lock()
+	if len(st.bad) == 0 && reportedBad {
+		err := notifySlack("All DERPs recovered.")
```
DentonGentry, line 199: Could rearrange this to not hold the lock across the POST to Slack.

cmd/derpprobe/derpprobe.go
```diff
@@ -135,13 +138,71 @@ func getOverallStatus() (o overallStatus) {
...
+	}
+	req.Header.Add("Content-Type", "application/json")
+
+	// if Slack is down, don't break the DERP monitoring. Timeout quickly.
+	client := &http.Client{Timeout: 3 * time.Second}
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+
```
bradfitz, line 168: defer resp.Body.Close()

cmd/derpprobe/derpprobe.go
```diff
@@ -135,13 +138,71 @@ func getOverallStatus() (o overallStatus) {
...
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(resp.Body)
```
bradfitz, line 170: this isn't guaranteed to read everything.

```go
    if resp.StatusCode != 200 { return errrors.New(resp.Status) }
    body, _ := io.ReadAll(resp.Body)
    if string(body) != "ok" { .... 
```

cmd/derpprobe/derpprobe.go
bradfitz, line 199: Doing this 3 second operation with mu held does mean that all the calls to setState may now block for 3 seconds, slowing down the probe.

But I guess that's probably tolerable, as those have 60 second context timeouts.

cmd/derpprobe/derpprobe.go
```diff
@@ -135,13 +138,71 @@ func getOverallStatus() (o overallStatus) {
...
+}
+
 func serve(w http.ResponseWriter, r *http.Request) {
 	st := getOverallStatus()
 	summary := "All good"
-	if len(st.bad) > 0 {
+	if (float64(len(st.bad)) / float64(len(st.bad)+len(st.good))) > 0.25 {
+		// This will generate an alert and page a human.
+		// It also ends up in Slack, as part of the alert pipeline.
 		w.WriteHeader(500)
```
bradfitz, line 183: I thought the plan was to always say 200 OK to Cloudradar and only have it monitor whether derpprobe is running, and then let derpprobe decide what makes it to Slack

cmd/derpprobe/derpprobe.go
```diff
@@ -135,13 +138,71 @@ func getOverallStatus() (o overallStatus) {
...
 		summary = fmt.Sprintf("%d problems", len(st.bad))
+	} else if len(st.bad) > 0 {
+		summary = fmt.Sprintf("%d problems", len(st.bad))
+		err := notifySlack(strings.Join(st.bad, "\n"))
```
bradfitz, line 187: so this'll notify slack only when the HTML is fetched?

how about having one dedicated notify-slack goroutine running in a loop and notifying slack whenever a condition persists for over N seconds?

cmd/derpprobe/derpprobe.go
DentonGentry: Done (also rearranged to not be a global now)

cmd/derpprobe/derpprobe.go
DentonGentry: Right now the only way we have to summon a human is cloudradar initiating a page, so at least for now if enough of the DERP fleet is out to be concerning I think returning 500 is the best way to handle it.

cmd/derpprobe/derpprobe.go
DentonGentry: Switched to a goroutine.

cmd/derpprobe/derpprobe.go
DentonGentry: Done.

DentonGentry:
One problem I'm having in testing is very frequent timeouts:
- most often when trying to contact Sydney
- next most often for Bangalore
- sometimes even DFW times out

I'm running my prober instance in Digital Ocean's Toronto DC. This PR won't be submitted until the timeouts are understood.

bradfitz (approved):

danderson (approved):

Issue #1866, opened 2021-05-06
Raspberry Pi relay node images


renne:
Hi, I suggest SD-Card images using the Raspberry Pi models as pre-configured relay nodes.
The user would just have to configure subnets/routes in a web-interface and authenticate the device with the tailscale network.

DentonGentry:
Are you familiar with https://github.com/gokrazy/gokrazy ?

We've discussed using it to build an image for an RPi to be a subnet relay appliance

renne:
No, I don't know gokrazy, yet.
It would also be great if the SD image automagically advertizes the subnets it receives via DHCP/RA.
Just send someone an SD-card with the image, he/she plugs it into a Raspberry Pi, adds routes in his/her router and authorizes the relay node with Tailscale.

renne:
The relay node should also use IPv6 router advertisement to advertise all IPv6 subnets of the tailscale network including subnets of other relay nodes.

That way only IPv4 subnet routes have to be configured manually on the router in the local network.

DentonGentry:
Re: gokrazy, though still appealing, it currently does not work well with tailscale because we shell out for utilities like iptables which gokrazy does not have.

bradfitz:
As you might've noticed from the flurry of recent commits referencing this bug, I've started working on this a bit, motivated almost entirely by the fact that [Gokrazy](https://gokrazy.org/) now supports encrypted wifi:

https://twitter.com/zekjur/status/1498567764444958723

That means we can now have a form on Tailscale's side (e.g. `https://tailscale.com/pi` or something) to download a dynamically generated ~60M raspberry pi image that starts Tailscale on boot, even baking in Tailscale auth keys (so it comes up automatically on boot headless, without requiring a monitor or keyboard) and wifi passwords, and being a subnet router even. And we can use the web configurator thing (tailscale web) that's minimally working currently for Synology. (we could flesh it out to do more options)

/cc @stapelberg

bradfitz:
> It would also be great if the SD image automagically advertizes the subnets it receives via DHCP/RA.
> Just send someone an SD-card with the image, he/she plugs it into a Raspberry Pi, adds routes in his/her router and
> authorizes the relay node with Tailscale.

That's a fun idea. That might be too aggressive to do by default, but your current LAN CIDR(s) could definitely be prominently recommended as options.

digaus:
Any progress on this? 🙂

DentonGentry:
[...]

Issue #391, opened 2020-05-15
Program routes and policy rules using netlink, not iptables binary


danderson:
The linux router currently programs the linux network stack with `ip` and `iptables`. `iptables` is unavoidable because the corresponding netlink APIs are not safe for direct use. However, everything we do with `ip` we could just do directly over a netlink socket.

This would in particular help embedded systems that ship with the Busybox version of `ip`, which is stripped down and doesn't understand advanced commands like policy routing rules, which Tailscale requires.

stapelberg:
> `iptables` is unavoidable because the corresponding netlink APIs are not safe for direct use.

Can you expand on this? Which APIs do you need, and in which way are they unsafe?

I was about to open an issue suggesting to use https://github.com/google/nftables/ where available to avoid the iptables dependency. This would help make tailscale work on https://gokrazy.org/, where only Go is available (no C userland, i.e. no iptables).

danderson:
For one, we can't depend on nftables, because we need to support OSes that haven't switched to nftables under the hood (e.g. pre-Buster debian), and we have to be cross-compatible with other software that doesn't support nftables (e.g. Docker, Kubernetes). The main way we can do that safely is to continue using the `iptables` binary, which will do "the right thing" based on OS/administrator policy.

And the legacy iptables is unsafe to use over netlink, because the API does replacement of the entire firewall at once, and doesn't have a way of preventing concurrent writes from squashing each other. The userspace tools work around this (poorly) with file-based locks, which already goes wrong in situations like containers.

If we can (a) positively detect that nftables, and only nftables is in use on the system; (b) we can manipulate the "legacy iptables converted to nftables" rules in nftables without conflicting with `iptables` itself... Then we could possibly make that work. But it's a bunch of work that, frankly, only makes sense for something like gokrazy, and not really for any other OS. So, that's going to be pretty low priority from my POV.

stapelberg:
Thanks for the additional details, I agree with your view point.

I expect that using nftables directly might become more attractive in a few years as more and more installations run on the legacy-converted-to-nftables model.

Would an nftables fallback code path be something that you would accept if people contributed it, or is there a lot of churn and hence maintenance burden in that part of the tailscale code base?

danderson:
I would love to not depend on `iptables` where possible. The part I'm unsure about is whether we can reliably detect that yes, it's definitely safe to use nftables directly without breaking the system.

[...]

PR #4053, opened 2022-03-02
cmd/tailscaled: default to userspace-networking mode on gokrazy, set paths

bradfitz:
One of the current few steps to run Tailscale on gokrazy is to
specify the --tun=userspace-networking flag:

https://gokrazy.org/userguide/install/tailscale/

Instead, make it the default for now. Later we can change the
default to kernel mode if available and fall back to userspace
mode like Synology, once #391 is done.

Likewise, set default paths for Gokrazy, as its filesystem hierarchy
is not the Linux standard one. Instead, use the conventional paths as
documented at https://gokrazy.org/userguide/install/tailscale/.

Updates #1866

RELNOTE=default to userspace-networking mode on gokrazy

/cc @stapelberg

### DIFF
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -70,11 +70,27 @@ func defaultTunName() string {
 		// as a magic value that uses/creates any free number.
 		return "utun"
 	case "linux":
-		if distro.Get() == distro.Synology {
+		switch distro.Get() {
+		case distro.Synology:
 			// Try TUN, but fall back to userspace networking if needed.
 			// See https://github.com/tailscale/tailscale-synology/issues/35
 			return "tailscale0,userspace-networking"
+		case distro.Gokrazy:
+			// Gokrazy doesn't yet work in tun mode because the whole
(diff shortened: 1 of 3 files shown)


cmd/tailscaled/tailscaled.go
```diff
@@ -70,11 +70,27 @@ func defaultTunName() string {
...
 			// See https://github.com/tailscale/tailscale-synology/issues/35
 			return "tailscale0,userspace-networking"
+		case distro.Gokrazy:
+			// Gokrazy doesn't yet work in tun mode because the whole
+			// Gokrazy thing is no C code, and Tailscale currently
+			// depends on the iptables binary for Linux's
+			// wgengine/router.
+			// But on Gokrazy there's no legacy iptables, so we could use netlink
+			// to program nft-iptables directly. It just isn't done yet.
+			// TODO(bradfitz): that's issue 391.
```
DentonGentry, line 85: Recommend the full URL for "391" as various editors will turn that into a clickable link.

bradfitz: done

PR #4062, opened 2022-03-03
    control/controlclient: make SetDNS attempt to go over Noise first

maisem:
Updates #3488

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -1254,6 +1254,45 @@ func (c *Direct) getNoiseClient() (*noiseClient, error) {
 	return np.(*noiseClient), nil
 }
 
+// setDNSNoise sends the SetDNSRequest request to the control plane server over Noise,
+// requesting a DNS record be created or updated.
+func (c *Direct) setDNSNoise(ctx context.Context, req *tailcfg.SetDNSRequest) error {
+	newReq := *req
+	newReq.Version = tailcfg.CurrentCapabilityVersion
+	np, err := c.getNoiseClient()
+	if err != nil {
+		return err
+	}
+	bodyData, err := json.Marshal(newReq)
+	if err != nil {
+		return err
+	}
+	res, err := np.Post(fmt.Sprintf("https://%v/%v", np.serverHost, "machine/set-dns"), "application/json", bytes.NewReader(bodyData))
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+	if res.StatusCode != 200 {
+		msg, _ := ioutil.ReadAll(res.Body)
+		return fmt.Errorf("set-dns response: %v, %.200s", res.Status, strings.TrimSpace(string(msg)))
+	}
+	var setDNSRes tailcfg.SetDNSResponse
+	if err := json.NewDecoder(res.Body).Decode(&setDNSRes); err != nil {
+		c.logf("error decoding SetDNSResponse: %v", err)
+		return fmt.Errorf("set-dns-response: %w", err)
+	}
+
+	return nil
+}
+
+// noiseConfigured reports whether the client can communicate with Control
+// over Noise.
+func (c *Direct) noiseConfigured() bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return !c.serverNoiseKey.IsZero()
+}
+
 // SetDNS sends the SetDNSRequest request to the control plane server,
 // requesting a DNS record be created or updated.
 func (c *Direct) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) (err error) {
@@ -1263,6 +1302,9 @@ func (c *Direct) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) (err er
 			metricSetDNSError.Add(1)
 		}
 	}()
+	if c.noiseConfigured() {
+		return c.setDNSNoise(ctx, req)
+	}
 	c.mu.Lock()
 	serverKey := c.serverKey
 	c.mu.Unlock()
@@ -1301,7 +1343,7 @@ func (c *Direct) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) (err er
 	var setDNSRes tailcfg.SetDNSResponse
 	if err := decode(res, &setDNSRes, serverKey, machinePrivKey); err != nil {
 		c.logf("error decoding SetDNSResponse with server key %s and machine key %s: %v", serverKey, machinePrivKey.Public(), err)
-		return fmt.Errorf("set-dns-response: %v", err)
+		return fmt.Errorf("set-dns-response: %w", err)
 	}
 
 	return nil
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -1298,7 +1298,7 @@ func (c *Direct) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) (err er
 		msg, _ := ioutil.ReadAll(res.Body)
 		return fmt.Errorf("set-dns response: %v, %.200s", res.Status, strings.TrimSpace(string(msg)))
 	}
-	var setDNSRes struct{} // no fields yet
+	var setDNSRes tailcfg.SetDNSResponse
 	if err := decode(res, &setDNSRes, serverKey, machinePrivKey); err != nil {
 		c.logf("error decoding SetDNSResponse with server key %s and machine key %s: %v", serverKey, machinePrivKey.Public(), err)
 		return fmt.Errorf("set-dns-response: %v", err)
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -1534,6 +1534,9 @@ type SetDNSRequest struct {
 	Value string
 }
 
+// SetDNSResponse is the response to a SetDNSRequest.
+type SetDNSResponse struct{}
+
 // SSHPolicy is the policy for how to handle incoming SSH connections
 // over Tailscale.
 type SSHPolicy struct {


control/controlclient/direct.go
```diff
@@ -200,6 +206,17 @@ func NewDirect(opts Options) (*Direct, error) {
 	return c, nil
 }
 
+// Close closes any open idle connections.
+// TODO(maisem): make this do more.
+func (c *Direct) Close() error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.noiseClient != nil {
+		c.noiseClient.CloseIdleConnections()
```
bradfitz, line 215: if we're closing the connection I think we want to be more aggressive than this and actually close all the underlying TCP connection(s) immediately

maisem: I don't follow, wdym? Are you saying that we don't keep any underlying TCP connection open at all?

maisem: Done

control/controlclient/noise.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	nc.mu.Lock()
+	nc.serverKey = serverKey
+	nc.conns = append(nc.conns, conn)
+	nc.mu.Unlock()
+	return conn, nil
+}
+
+func loadServerNoiseKey(ctx context.Context, httpc *http.Client, serverURL string) (k key.MachinePublic, _ error) {
+	url := fmt.Sprintf("%s/server-key?mapRequestVersion=%d", serverURL, tailcfg.CurrentMapRequestVersion)
```
bradfitz, line 112: unusual param case? how about just `mapver` or `v`?

control/controlclient/noise.go
```diff
@@ -0,0 +1,118 @@
...
+	u, err := url.Parse(serverURL)
+	if err != nil {
+		return nil, err
+	}
+	conn, err := controlhttp.Dial(ctx, u.Host, priKey, serverKey)
+	if err != nil {
+		return nil, err
+	}
+	nc.mu.Lock()
+	nc.serverKey = serverKey
```
bradfitz, line 105: as long as this doesn't reset it back to an earlier version, once we do key rotation.  (I'm worried about a concurrent outstanding Dial + key rotation at once, and the Dial fighting the key rotation)

control/controlclient/noise.go
```diff
@@ -0,0 +1,118 @@
...
+	nc.conns = append(nc.conns, conn)
```
bradfitz, line 106: maybe a map instead of a list? so you can remove them from the map when the conn closes?

control/controlclient/noise.go
```diff
@@ -0,0 +1,118 @@
...
+	"tailscale.com/tailcfg"
+	"tailscale.com/types/key"
+	"tailscale.com/util/multierr"
+)
+
+// noiseClient provides a http.Client to connect to tailcontrol over
+// the ts2021 protocol.
+type noiseClient struct {
+	http.Client // HTTP client used to talk to tailcontrol
+	priKey      key.MachinePrivate
```
bradfitz, line 29: priKey sounds like priorityKey. `privKey`.

control/controlclient/noise.go
```diff
@@ -0,0 +1,118 @@
...
+	serverKey   key.MachinePublic
```
bradfitz, line 30: `serverPub` or `serverPubKey` ?

control/controlclient/noise.go
```diff
@@ -0,0 +1,118 @@
...
+	serverURL   string
```
bradfitz, line 31: example. http? https? just hostname? trailing slash?

control/controlclient/direct.go
```diff
@@ -1190,6 +1209,58 @@ func sleepAsRequested(ctx context.Context, logf logger.Logf, timeoutReset chan<-
...
+	if err != nil {
+		return nil, err
+	}
+
+	np, err = newNoiseClient(ctx, k, su)
+	if err != nil {
+		return nil, err
+	}
+	c.mu.Lock()
+	c.noiseClient = np
```
bradfitz, line 1231: at this point the noiseClient might've already been assigned by a concurrent call to getNoiseClient who won the race.

do you want to check that here and close the one you just made on Line 1226 and return the winner one?

control/controlclient/noise.go
maisem: Is this better?

control/controlclient/direct.go
```diff
@@ -1199,6 +1270,12 @@ func (c *Direct) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) (err er
 			metricSetDNSError.Add(1)
 		}
 	}()
+	// Try noise first.
+	if err := c.setDNSNoise(ctx, req); err == nil {
+		return nil
+	} else {
+		c.logf("SetDNS failed over noise, falling back to NaCL box: %v", err)
```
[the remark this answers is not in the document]
maisem: How about leave it in for now and remove once we have a little more confidence and definitely before 1.24?

control/controlclient/noise.go
```diff
@@ -0,0 +1,162 @@
...
+			// We lost the race, use the key that won.
+			serverKey = nc.serverPubKey
+		}
+		nc.mu.Unlock()
+	}
+
+	conn, err := controlhttp.Dial(ctx, serverHost, priKey, serverKey)
+	if err != nil {
+		return nil, err
+	}
```
maisem, line 143: this is kind of annoying. We need to give `controlhttp.Dial` an addr which goes over the http endpoint.
https://github.com/tailscale/tailscale/blob/0e62a7d1a247eb312f340b08e92942f73149f11f/control/controlhttp/client.go#L58-L82

control/controlclient/direct.go
```diff
@@ -950,29 +968,25 @@ func encode(v interface{}, serverKey key.MachinePublic, mkey key.MachinePrivate)
...
-	}
 	if res.StatusCode != 200 {
-		return key.MachinePublic{}, fmt.Errorf("fetch control key: %d: %s", res.StatusCode, string(b))
+		return nil, fmt.Errorf("fetch control key: %d", res.StatusCode)
 	}
-	k, err := key.ParseMachinePublicUntyped(mem.B(b))
-	if err != nil {
-		return key.MachinePublic{}, fmt.Errorf("fetch control key: %v", err)
+	var out tailcfg.OverTLSPublicKeyResponse
+	if err := json.NewDecoder(res.Body).Decode(&out); err != nil {
```
bradfitz, line 986: You need to be prepared for an old control server here that doesn't return JSON even with the `?v=N` param. So you need to read the res.Body into memory and first try JSON, else treat it as the legacy pub key (without the `mkey:` prefix).

maisem: Good catch, done.

control/controlclient/direct.go
```diff
@@ -950,29 +956,38 @@ func encode(v interface{}, serverKey key.MachinePublic, mkey key.MachinePrivate)
...
+		return nil, fmt.Errorf("create control key request: %v", err)
 	}
-	req = req.WithContext(ctx)
 	res, err := httpc.Do(req)
 	if err != nil {
-		return key.MachinePublic{}, fmt.Errorf("fetch control key: %v", err)
+		return nil, fmt.Errorf("fetch control key: %v", err)
 	}
 	defer res.Body.Close()
 	b, err := ioutil.ReadAll(io.LimitReader(res.Body, 1<<16))
```
bradfitz, line 984: while you're here: 64 << 10 instead of 1 << 6.

Go style is to only do shifts of 10, 20, 30, etc for KB, MB, GB.

control/controlclient/direct.go
```diff
@@ -950,29 +956,38 @@ func encode(v interface{}, serverKey key.MachinePublic, mkey key.MachinePrivate)
...
+
+	// Some old control servers might not be updated to send the new format.
+	// Fallback to the old way just in case.
 	k, err := key.ParseMachinePublicUntyped(mem.B(b))
 	if err != nil {
-		return key.MachinePublic{}, fmt.Errorf("fetch control key: %v", err)
+		return nil, multierr.New(jsonErr, err)
 	}
-	return k, nil
+	out.LegacyPublicKey = k
```
bradfitz, line 1003: sure it's in a zeroed state here? I think technically json.Unmarshal can leave it half filled and still return an error midway through?

control/controlclient/direct.go
maisem: Done

control/controlclient/direct.go
maisem: I hadn't considered that, thanks!

tailcfg/tailcfg.go
```diff
@@ -1534,6 +1534,9 @@ type SetDNSRequest struct {
 	Value string
 }
 
+// SetDNSResponse is the repsonse to a SetDNSRequest.
```
bradfitz, line 1537: typo. But you could also send this separately if you wanted.

control/controlclient/noise.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (c *noiseConn) Close() error {
+	if err := c.Conn.Close(); err != nil {
+		return err
+	}
+	c.pool.connClosed(c.id)
+	return nil
+}
+
+// noiseClient provides a http.Client to connect to tailcontrol over
```
bradfitz, line 38: break off the http2 stuff from the SetDNS stuff into separate PRs?

control/controlclient/direct.go
```diff
@@ -1213,6 +1288,13 @@ func (c *Direct) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) (err er
 			metricSetDNSError.Add(1)
 		}
 	}()
+	// Try noise first.
+	// TODO(maisem/bradfitz): remove the fallback before 1.24.
```
bradfitz, line 1292: I'd rather do it now. We can turn off noise if we find problems from the /key?v=28 URL.

That is: if there's noise, use noise only.

control/controlclient/direct.go
```diff
@@ -1204,6 +1218,67 @@ func sleepAsRequested(ctx context.Context, logf logger.Logf, timeoutReset chan<-
...
+		// We lost the race, clean up.
+		np.Close()
+		np = c.noiseClient
+	} else {
+		c.noiseClient = np
+	}
+	return np, nil
+}
+
+func (c *Direct) setDNSNoise(ctx context.Context, req *tailcfg.SetDNSRequest) error {
```
bradfitz, line 1255: req.Version = tailcfg.CurrentCapabilityVersion?

just document that this mutates req?

control/controlclient/direct.go
```diff
@@ -203,6 +204,19 @@ func NewDirect(opts Options) (*Direct, error) {
 	return c, nil
 }
 
+// Close closes the underlying noise connection.
```
bradfitz, line 207: Noise

and connection(s) ?

control/controlclient/noise.go
```diff
@@ -0,0 +1,142 @@
...
+	"time"
+
+	"golang.org/x/net/http2"
+	"tailscale.com/control/controlbase"
+	"tailscale.com/control/controlhttp"
+	"tailscale.com/types/key"
+	"tailscale.com/util/multierr"
+)
+
+type noiseConn struct {
```
bradfitz, line 24: docs

tailcfg/tailcfg.go
maisem: Added a separate commit, don't think we need a separate PR.

control/controlclient/noise.go
```diff
@@ -0,0 +1,142 @@
...
+		Transport: &http2.Transport{
+			DialTLS:   np.Dial,
+			AllowHTTP: true,
+		},
+	}
+
+	return np, nil
+}
+
+// connClosed is called from noiseConn.Close to clean up the connection
```
bradfitz, line 84: // connClosed removes the connection with the provided ID from the pool
// of active connections.

?

control/controlclient/noise.go
maisem: Sure

control/controlclient/direct.go
maisem: That would still require a client restart to drop the cached key, no?

bradfitz: `tailcfg.Debug.Exit` 🤷‍♂️

control/controlclient/noise.go
maisem: https://github.com/tailscale/tailscale/pull/4092

control/controlclient/direct.go
maisem: created a copy of req

control/controlclient/noise.go
maisem: Rebased

control/controlclient/direct.go
```diff
@@ -1254,6 +1254,45 @@ func (c *Direct) getNoiseClient() (*noiseClient, error) {
...
+	}
+	defer res.Body.Close()
+	if res.StatusCode != 200 {
+		msg, _ := ioutil.ReadAll(res.Body)
+		return fmt.Errorf("set-dns response: %v, %.200s", res.Status, strings.TrimSpace(string(msg)))
+	}
+	var setDNSRes tailcfg.SetDNSResponse
+	if err := json.NewDecoder(res.Body).Decode(&setDNSRes); err != nil {
+		c.logf("error decoding SetDNSResponse: %v", err)
+		return fmt.Errorf("set-dns-response: %v", err)
```
bradfitz, line 1282: %w ?

bradfitz (approved):

control/controlclient/direct.go
maisem: fixed both places

PR #4063, opened 2022-03-03
cmd/tailscale/cli, ipn: move exit node IP parsing and validation from…

dblohm7:
… cli into prefs.

We need to be able to provide the ability for the GUI clients to resolve and set
the exit node IP from an untrusted string, thus enabling the ability to specify
that information via enterprise policy.

This patch moves the relevant code out of the handler for `tailscale up`,
into a method on `Prefs` that may then be called by GUI clients.

### DIFF
--- a/cmd/tailscale/cli/cli_test.go
+++ b/cmd/tailscale/cli/cli_test.go
@@ -19,7 +19,6 @@ import (
 	"tailscale.com/ipn"
 	"tailscale.com/ipn/ipnstate"
 	"tailscale.com/tstest"
-	"tailscale.com/types/key"
 	"tailscale.com/types/persist"
 	"tailscale.com/types/preftype"
 	"tailscale.com/version/distro"
@@ -632,7 +631,7 @@ func TestPrefsFromUpArgs(t *testing.T) {
 			st: &ipnstate.Status{
 				TailscaleIPs: []netaddr.IP{netaddr.MustParseIP("100.105.106.107")},
 			},
-			wantErr: `cannot use 100.105.106.107 as the exit node as it is a local IP address to this machine, did you mean --advertise-exit-node?`,
+			wantErr: `cannot use 100.105.106.107 as an exit node as it is a local IP address to this machine; did you mean --advertise-exit-node?`,
 		},
 		{
 			name: "warn_linux_netfilter_nodivert",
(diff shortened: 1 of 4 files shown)


ipn/prefs.go
```diff
@@ -477,6 +482,92 @@ func (p *Prefs) SetAdvertiseExitNode(runExit bool) {
...
+func checkIPNotSelf(st *ipnstate.Status, ip netaddr.IP) bool {
+	for _, tip := range st.TailscaleIPs {
+		if ip == tip {
+			return false
+		}
+	}
+	return true
+}
+
+// ClearExitNodeIP sets the ExitNodeIP to the zero value.
```
bradfitz, line 507: clarify that this doesn't mutate the ExitNodeID, though?

ipn/prefs.go
```diff
@@ -477,6 +482,92 @@ func (p *Prefs) SetAdvertiseExitNode(runExit bool) {
...
+			if ip == ip2 {
+				return ps, true
+			}
+		}
+	}
+	return nil, false
+}
+
+func checkIPNotSelf(st *ipnstate.Status, ip netaddr.IP) bool {
+	for _, tip := range st.TailscaleIPs {
```
bradfitz, line 499: s/tip/selfIP/ ?

ipn/prefs.go
```diff
@@ -477,6 +482,92 @@ func (p *Prefs) SetAdvertiseExitNode(runExit bool) {
...
+		if !checkIPNotSelf(st, ip) {
+			return ip, ErrExitNodeLocalIP
+		}
+		return ip, nil
+	default:
+		return ip, fmt.Errorf("ambiguous exit node name %q", s)
+	}
+}
+
+// SetExitNodeIP validates and sets the ExitNodeIP from a user-provided string
```
bradfitz, line 562: final period.

And clarify what types of user-provided strings are acceptable.

ipn/prefs.go
```diff
@@ -477,6 +482,92 @@ func (p *Prefs) SetAdvertiseExitNode(runExit bool) {
...
+func (p *Prefs) SetExitNodeIP(s string, st *ipnstate.Status) error {
+	ip, err := validateExitNodeIP(s, st)
+	if err == nil {
+		p.ExitNodeIP = ip
```
bradfitz, line 566: and document that this doesn't mutate ExitNodeID.

(should it return an error if ID is non-empty?)

bradfitz (commented):
Add an `Updates tailscale/corp#nnnn` line to commit message?

cmd/tailscale/cli/cli_test.go
```diff
@@ -632,7 +631,7 @@ func TestPrefsFromUpArgs(t *testing.T) {
 			st: &ipnstate.Status{
 				TailscaleIPs: []netaddr.IP{netaddr.MustParseIP("100.105.106.107")},
 			},
-			wantErr: `cannot use 100.105.106.107 as the exit node as it is a local IP address to this machine, did you mean --advertise-exit-node?`,
+			wantErr: `100.105.106.107: cannot use this address as an exit node as it is a local IP address to this machine; did you mean --advertise-exit-node?`,
```
bradfitz, line 634: I kinda liked the old word order more ("the" to "an" is fine, and semicolon is better, though)

ipn/prefs.go
dblohm7: Actually we probably want to mutate both (the purpose is for clearing out any persisted exit node info). Updated.

ipn/prefs.go
dblohm7: Sure.

ipn/prefs.go
dblohm7: Good idea.

cmd/tailscale/cli/cli_test.go
dblohm7: This is a consequence of the fact that the patch splits some of that error message by moving some of that code into `prefs.go` while leaving some of it in `up.go`. (Some parts of that message seemed more appropriate for just the command line use case.)

I'll go over this and reconsider.

dblohm7: OK I redid this in a way that preserves the ordering in this message.

ipn/prefs.go
```diff
@@ -467,6 +474,106 @@ func (p *Prefs) SetAdvertiseExitNode(runExit bool) {
...
+
+type ExitNodeLocalIPError struct {
+	hostOrIP string
+}
+
+func (e ExitNodeLocalIPError) Error() string {
+	return fmt.Sprintf("cannot use %s as an exit node as it is a local IP address to this machine", e.hostOrIP)
+}
+
+func validateExitNodeIP(s string, st *ipnstate.Status) (ip netaddr.IP, err error) {
```
bradfitz, line 513: keep the old func name? The test you moved is still named `TestExitNodeIPOfArg` after all.
 
Not sure "validate" adds much when the signature already returns an error. Validation is implicit.

ipn/prefs.go
```diff
@@ -467,6 +474,106 @@ func (p *Prefs) SetAdvertiseExitNode(runExit bool) {
...
+	return true
+}
+
+// ClearExitNodeInfo sets the ExitNodeID and ExitNodeIP to their zero values.
+func (p *Prefs) ClearExitNodeInfo() {
+	p.ExitNodeID = ""
+	p.ExitNodeIP = netaddr.IP{}
+}
+
+type ExitNodeLocalIPError struct {
```
bradfitz, line 505: exported API needs docs.

ipn/prefs.go
```diff
@@ -467,6 +474,106 @@ func (p *Prefs) SetAdvertiseExitNode(runExit bool) {
...
+	hostOrIP string
+}
+
+func (e ExitNodeLocalIPError) Error() string {
```
bradfitz, line 509: this is a value receiver but you used it as a pointer everywhere else. Be consistent either way.

ipn/prefs.go
```diff
@@ -467,6 +474,106 @@ func (p *Prefs) SetAdvertiseExitNode(runExit bool) {
...
+			return ip, &ExitNodeLocalIPError{s}
+		}
+		return ip, nil
+	default:
+		return ip, fmt.Errorf("ambiguous exit node name %q", s)
+	}
+}
+
+// SetExitNodeIP validates and sets the ExitNodeIP from a user-provided string
+// specifying either an IP address or a MagicDNS host name. This method does
```
bradfitz, line 564: clarify that it's only the MagicDNS base name ("foo", not "foo.bar.beta.tailscale.net")

ipn/prefs.go
```diff
@@ -467,6 +474,106 @@ func (p *Prefs) SetAdvertiseExitNode(runExit bool) {
...
+		for _, ip2 := range ps.TailscaleIPs {
+			if ip == ip2 {
+				return ps, true
+			}
+		}
+	}
+	return nil, false
+}
+
+func checkIPNotSelf(st *ipnstate.Status, ip netaddr.IP) bool {
```
bradfitz, line 490: this name isn't clear. boolean funcs with negated names rarely are.

Docs would help, but so would renaming it.

Or just get rid of this and instead make the caller use `!tsaddr.IPsContains(st.TailscaleIPs, ip)` ... but you'd need to add that variant to the `tsaddr` package. We have a dozen variants but not that one.

ipn/prefs.go
```diff
@@ -467,6 +474,106 @@ func (p *Prefs) SetAdvertiseExitNode(runExit bool) {
...
+	for _, selfIP := range st.TailscaleIPs {
+		if ip == selfIP {
+			return false
+		}
+	}
+	return true
+}
+
+// ClearExitNodeInfo sets the ExitNodeID and ExitNodeIP to their zero values.
```
bradfitz, line 499: Drop "Info" ?

bradfitz (approved):

PR #4076, opened 2022-03-06
tailcfg: add OverTLSPublicKeyResponse for the new response from /key

bradfitz:
Updates #3488

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -1596,3 +1596,27 @@ type SSHAction struct {
 	// value.
 	HoldAndDelegate string `json:"holdAndDelegate,omitempty"`
 }
+
+// OverTLSPublicKeyResponse is the JSON response to /key?v=<n>
+// over HTTPS (regular TLS) to the Tailscale control plane server,
+// where the 'v' argument is the client's current capability version
+// (previously known as the "MapRequest version").
+//
+// The "OverTLS" prefix is to loudly declare that this exchange
+// doesn't happen over Noise and can be intercepted/MITM'ed by
+// enterprise/corp proxies where the orgnanization can put TLS roots
+// on devices.
+type OverTLSPublicKeyResponse struct {
+	// LegacyPublic specifies the control plane server's original
+	// NaCl crypto_box machine key.
+	// It will be zero for sufficiently new clients, based on their
+	// advertised "v" parameter (the CurrentMapRequestVersion).
+	// In that case, only the newer Noise-based transport may be used
+	// using the PublicKey field.
+	LegacyPublicKey key.MachinePublic `json:"legacyPublicKey"`
+
+	// PublicKey specifies the server's public key for the
+	// Noise-based control plane protocol. (see packages
+	// control/controlbase and control/controlhttp)
+	PublicKey key.MachinePublic `json:"publicKey"`
+}


bradfitz:
Thoughts? Objections?

maisem:
> Thoughts? Objections?

I like this better than what we are currently doing. We should add a test for the old endpoint though to make sure we don't accidentally regress.

tailcfg/tailcfg.go
```diff
@@ -1596,3 +1604,18 @@ type SSHAction struct {
...
+// over HTTPS (regular TLS) to the Tailscale control plane server,
+// where <n> is the client's CurrentMapRequestVersion.
+type ServerMetaResponse struct {
+	// PublicNacl specifies the server's original NaCl-based machine key.
+	// It will be zero for sufficiently new clients, based on their
+	// advertised "v" parameter (the CurrentMapRequestVersion).
+	// In that case, only the Noise-based transport may be used.
+	PublicKeyNacl key.MachinePublic `json:"publicKeyNacl"`
+
+	// PublicKeyNoise specifies the server's public key for the
```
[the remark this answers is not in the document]
bradfitz: This is oss. You mean controlbase, or ...?

tailcfg/tailcfg.go
```diff
@@ -26,6 +26,14 @@ import (
 
 // CurrentMapRequestVersion is the current MapRequest.Version value.
 //
+// Despite the historical name ("map request"), more generally it
```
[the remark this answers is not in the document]
bradfitz: Sure, renaming works. `CapabilityVersion`?

tailcfg/tailcfg.go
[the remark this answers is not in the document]
maisem: control/controlhttp?

tailcfg/tailcfg.go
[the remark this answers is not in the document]
maisem: how about `CapabilityRevision`? both wfm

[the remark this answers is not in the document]
bradfitz: Easier mentally to switch stuff like mapVer to capVer. Only one letter off and sounds the same?

tailcfg/tailcfg.go
```diff
@@ -1596,3 +1604,18 @@ type SSHAction struct {
...
+
+// ServerMetaResponse is the JSON response to a request to /meta?v=<n>
+// over HTTPS (regular TLS) to the Tailscale control plane server,
+// where <n> is the client's CurrentMapRequestVersion.
+type ServerMetaResponse struct {
+	// PublicNacl specifies the server's original NaCl-based machine key.
+	// It will be zero for sufficiently new clients, based on their
+	// advertised "v" parameter (the CurrentMapRequestVersion).
+	// In that case, only the Noise-based transport may be used.
+	PublicKeyNacl key.MachinePublic `json:"publicKeyNacl"`
```
maisem, line 1616: Want to add the fields for key rotation?

tailcfg/tailcfg.go
```diff
@@ -1596,3 +1604,18 @@ type SSHAction struct {
 	// value.
 	HoldAndDelegate string `json:"holdAndDelegate,omitempty"`
 }
+
+// ServerMetaResponse is the JSON response to a request to /meta?v=<n>
+// over HTTPS (regular TLS) to the Tailscale control plane server,
+// where <n> is the client's CurrentMapRequestVersion.
+type ServerMetaResponse struct {
+	// PublicNacl specifies the server's original NaCl-based machine key.
```
crawshaw, line 1612: NaCl is a library with several cryptographic primitives. NaCl I believe calls this crypto_box, which is sometimes called naclbox (but maybe not outside of Go?).

tailcfg/tailcfg.go
```diff
@@ -1596,3 +1604,18 @@ type SSHAction struct {
...
+	// It will be zero for sufficiently new clients, based on their
+	// advertised "v" parameter (the CurrentMapRequestVersion).
+	// In that case, only the Noise-based transport may be used.
+	PublicKeyNacl key.MachinePublic `json:"publicKeyNacl"`
+
+	// PublicKeyNoise specifies the server's public key for the
+	// Noise-based control plane protocol.
+	PublicKeyNoise key.MachinePublic `json:"publicKeyNoise"`
```
crawshaw, line 1620: So we're still relying on TLS roots for trusting who the server is. I guess that makes sense seeing as we haven't invented some new mechanism. (On my phone so it's hard to cross reference right now.)

But before we go down this road, I'd like to think through the ramifications of routing ts2021 through DERP. My understanding is we're going to make a noise session then do http2 over that, is that right? If so, then all server metadata, except the key used to establish the connection should be done over the noise channel. That means /meta is too generic a term, because no extra metadata should be outside of noise.

Also, should we introduce some separation here for commands that happen over noise and those that don't?

tailcfg/tailcfg.go
bradfitz: How about just LegacyPublicKey and PublicKey and we can remove the Legacy one later?

tailcfg/tailcfg.go
bradfitz: This being over TLS is already discussed in the comment above. This is mostly about converting the existing single-line text HTTP response to be JSON instead, so we can add key rotation date fields later. And it lets us combine the two public keys (for now) into one HTTP round trip. 

There are no plans for any other meta.

We can not name this "meta" if you want and only use the existing `/key` path but return JSON instead of a single line when the `?v=N` parameter is present and >= 27.

Then I guess we'd call this `KeyResponse` instead.

crawshaw: Yeah I think naming it something else reduces the risk someone accidentally adds any other metadata to this struct in the future. For newer clients, other discussions should happen over noise. Maybe the comment should point that out too.

This struct is very different than the rest of the structs in tailcfg. Everything else is sent/received over a private channel with the server, this one is out in the open.

bradfitz: PTAL?

tailcfg/tailcfg.go
bradfitz: Done.

tailcfg/tailcfg.go
```diff
@@ -1596,3 +1596,30 @@ type SSHAction struct {
...
+	// using the PublicKey field.
+	LegacyPublicKey key.MachinePublic `json:"legacyPublicKey"`
+
+	// PublicKey specifies the server's public key for the
+	// Noise-based control plane protocol.
+	PublicKey key.MachinePublic `json:"publicKey"`
+
+	// Expiry, if non-nil, is the time at which the public key(s)
+	// expire.
+	Expiry *time.Time `json:"expiry,omitempty"`
```
crawshaw, line 1624: Expiry is an example of something that shouldn't be distributed in plain text. (E.g. when a TLS cert includes an expiry, it's covered by the signature of the certificate.)

crawshaw (commented):
other than expiry, lgtm

crawshaw (approved):

tailcfg/tailcfg.go
bradfitz: Removed for now but I still think we want something here like this, perhaps named differently.

Currently the client caches the public key forever and never re-checks. It's TOFU for life of the process, which means we can never rotate the nacl public key. I want a field like this to say that the client should at least re-query the public key periodically, but we can deal with that later.

crawshaw: I agree we want an expiry. But the expiry needs proof associated that the server key agrees with it. The easy way for us to do that is to return it in a /meta request inside the noise channel.

stormshield-gt:
Sorry if the question is naive, but does that mean that a corporate proxy could impersonate the tailscale coordination server if it gives its public key instead in `OverTLSPublicKeyResponse`? For instance an attacker could install `headscale` on a compromised corporate proxy to do this?

PR #4078, opened 2022-03-06
tailcfg: rename map request version to "capability version"

bradfitz:
And add a CapabilityVersion type, primarily for documentation.

This makes MapRequest.Version, RegisterRequest.Version, and
SetDNSRequest.Version all use the same version, which will avoid
confusing in the future if Register or SetDNS ever changed their
semantics on Version change. (Currently they're both always 1)

This will requre a control server change to allow a
SetDNSRequest.Version value other than 1 to be deployed first.


crawshaw (approved):

maisem (approved):

bradfitz:
PTAL. I redefined Register and SetDNS version to always be 1 when over Nacl crypto_box. That'll be less invasive for other control plane servers.

crawshaw (approved):

PR #4087, opened 2022-03-07
wgengine/router: delete legacy netfilter rule cleanup [Linux]

bradfitz:
This was just cleanup for an ancient version of Tailscale. Any such machines
have upgraded since then.


bradfitz:
I think this is right? I can't see how it's useful nowadays.

Mostly my motivation is that this shuts up logspam from people who don't have the "comment" module.

danderson (approved):
Yeah, s'fine. I think that'll mean upgrading from pre-1.0 straight to 1.24+ might break in weird ways, but that should be okay.

PR #4092, opened 2022-03-07
control/controlclient: add Noise client

maisem:
From https://github.com/tailscale/tailscale/pull/4062#discussion_r821081272

Updates #3488


control/controlclient/direct.go
```diff
@@ -1204,6 +1218,40 @@ func sleepAsRequested(ctx context.Context, logf logger.Logf, timeoutReset chan<-
 	}
 }
 
+// getNoiseClient returns the noise client, creating one if one doesn't exist.
+func (c *Direct) getNoiseClient(ctx context.Context) (*noiseClient, error) {
+	c.mu.Lock()
+	su := c.serverURL
+	np := c.noiseClient
+	sk := c.serverNoiseKey
```
bradfitz, line 1226: `serverPub := c.serverNoisePubKey` ?

control/controlclient/noise.go
```diff
@@ -0,0 +1,146 @@
...
+	return nil
+}
+
+// noiseClient provides a http.Client to connect to tailcontrol over
+// the ts2021 protocol.
+type noiseClient struct {
+	*http.Client // HTTP client used to talk to tailcontrol
+	privKey      key.MachinePrivate
+	serverPubKey key.MachinePublic
+	connCount    int
```
bradfitz, line 48: next ID or last ID to be allocated? key of connPool?

control/controlclient/noise.go
```diff
@@ -0,0 +1,146 @@
...
+
+	serverHost string // the host:port part of serverURL
+
+	mu       sync.Mutex
+	connPool map[int]*noiseConn // All created connections. TODO(maisem): clean up dead connections.
```
bradfitz, line 53: You don't delete from this if the http2 connection dies on its own (e.g. trunkd restart, GOAWAY message, etc)

control/controlclient/noise.go
```diff
@@ -0,0 +1,146 @@
...
+	serverHost := nc.serverHost
+	connID := nc.connCount
+	if nc.connPool == nil {
+		nc.connPool = make(map[int]*noiseConn)
+	}
+	nc.connCount++
+	nc.mu.Unlock()
+
+	// This context is only used for the dial.
+	// Pick an arbitrary timeout.
```
bradfitz, line 129: // Timeout is a little arbitrary, but plenty long enough for even the
// highest latency links.

control/controlclient/noise.go
```diff
@@ -0,0 +1,146 @@
...
+
+	conn, err := controlhttp.Dial(ctx, serverHost, priKey, serverKey)
+	if err != nil {
+		return nil, err
+	}
+
+	nc.mu.Lock()
+	defer nc.mu.Unlock()
+	if !nc.serverPubKey.IsZero() {
+		nc.serverPubKey = serverKey
```
bradfitz, line 141: what's happening here? This came from nc.serverPubKey above? This adjusting for some race? But why's it updating the client?

control/controlclient/noise.go
```diff
@@ -0,0 +1,146 @@
...
+			errors = append(errors, err)
+		}
+	}
+	return multierr.New(errors...)
+}
+
+// Dial opens a new connection to tailcontrol, fetching the server noise key
+// if not cached. The network and address params are ignored as it only dials
+// the server it was created for.
+func (nc *noiseClient) Dial(_, _ string, cfg *tls.Config) (net.Conn, error) {
```
bradfitz, line 116: cfg is also unused. _ it too? 

The signature then looks (already looks) bizarre enough to be worth a comment about why it is how it is. (say what it's implementing)

control/controlclient/noise.go
```diff
@@ -0,0 +1,146 @@
...
+		serverPubKey: serverPubKey,
+		privKey:      priKey,
+		serverHost:   host,
+	}
+
+	// Create a new http.Client which dials out using nc.Dial.
+	np.Client = &http.Client{
+		Transport: &http2.Transport{
+			DialTLS:   np.Dial,
+			AllowHTTP: true,
```
bradfitz, line 81: comment

control/controlclient/direct.go
```diff
@@ -203,6 +204,19 @@ func NewDirect(opts Options) (*Direct, error) {
 	return c, nil
 }
 
+// Close closes the underlying Noise connection(s).
```
danderson, line 207: Why (s)? There should never be more than 1?

control/controlclient/noise.go
danderson, line 53: Confused, why isn't this 1 conn with http2 multiplexing?

control/controlclient/direct.go
```diff
@@ -1204,6 +1218,40 @@ func sleepAsRequested(ctx context.Context, logf logger.Logf, timeoutReset chan<-
...
+	if err != nil {
+		return nil, err
+	}
+
+	np, err = newNoiseClient(ctx, k, sk, su)
+	if err != nil {
+		return nil, err
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
danderson, line 1244: use a SingleFlight to avoid having these races?

control/controlclient/noise.go
danderson, line 116: Make this function unexported? I was very confused about it until I realized that it's an internal implementation detail of the http2 Transport configuration.

Also still confused by why we need all this machinery to manage multiple connections, when presumably we're dialing a single destination and (I thought) http2 would provide the multiplexing on top of the Conn.

I'm worried about this specifically because it's going to break some things we could do with noise connections, like binding hardware attestations to the underlying Noise state. We can't do that if we might have 15 different conns kicking around, my assumption was that we'd be running http2 muxing over a single conn, so it would be tractable to bind cryptographic state to the underlying connection.

danderson: IOW: with noise connections, each Conn is stateful and can carry more contextual information than just what was initially handshaken. Stuff like "client has proven to me, _on this specific Conn_, that they're running on the expected hardware". If we now let API calls float between different Conns, we lose that binding, and we can no longer do any strong credential cloning prevention, as one example.

bradfitz: That's fine. If/when we do that, we can do that in the Dial at the beginning before the http2 part starts.

control/controlclient/noise.go
bradfitz: It is, but http2 can GOAWAY to drain the connection (e.g. trunkd graceful shutdowns)

control/controlclient/direct.go
maisem: I don't think `singleflight` helps us much here. We need to be able to create multiple connections (in case of trunkd restarts, local client errors etc), and we need the mutex anyways for cleaning up/closing.

control/controlclient/noise.go
maisem: Unexported

control/controlclient/noise.go
maisem: Remnant from when this was fetching the server key

control/controlclient/noise.go
maisem: hence the TODO :)
Any suggestions on how to detect/cleanup?

control/controlclient/noise.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	np := &noiseClient{
+		serverPubKey: serverPubKey,
+		privKey:      priKey,
+		serverHost:   host,
+	}
+
+	// Create a new http.Client which dials out using nc.Dial.
+	np.Client = &http.Client{
+		Transport: &http2.Transport{
```
bradfitz, line 79: set ReadIdleTimeout too? To like 60 seconds?

control/controlclient/noise.go
```diff
@@ -0,0 +1,141 @@
...
+			DialTLS: np.dial,
+			// Turn on AllowHTTP as we don't rely on TLS.
```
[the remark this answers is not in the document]
bradfitz: Like why can't we just send requests to `https://controlplane.tailscale.com/machine/foo` with the `https://` scheme?

control/controlclient/noise.go
```diff
@@ -0,0 +1,141 @@
...
+	for _, c := range conns {
+		if err := c.Close(); err != nil {
+			errors = append(errors, err)
+		}
+	}
+	return multierr.New(errors...)
+}
+
+// dial opens a new connection to tailcontrol, fetching the server noise key
+// if not cached. It implments tls.Dial but ignores all params as it only
```
bradfitz, line 115: typo implements

It implements the signature needed by http2.Transport.DialTLS.

control/controlclient/noise.go
bradfitz: `// active connections not yet closed; see noiseConn.Close`

bradfitz (approved):
LGTM after wording nits

control/controlclient/noise.go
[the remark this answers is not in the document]
maisem: I didn't realize that we could ¯\_(ツ)_/¯. Tested and removed

control/controlclient/direct.go
maisem: oops, you're right this can be simplified by singleflight. I thought this was in `noise.go`.

control/controlclient/direct.go
```diff
@@ -72,6 +73,9 @@ type Direct struct {
 	mu             sync.Mutex        // mutex guards the following fields
 	serverKey      key.MachinePublic // original ("legacy") nacl crypto_box-based public key
 	serverNoiseKey key.MachinePublic
+	noiseClient    *noiseClient
+
+	sfGroup singleflight.Group
```
bradfitz, line 78: say what it does/guards, how it's used, something. Otherwise it's just:

`integer int`

control/controlclient/direct.go
```diff
@@ -1204,6 +1221,32 @@ func sleepAsRequested(ctx context.Context, logf logger.Logf, timeoutReset chan<-
...
+	np, err, _ := c.sfGroup.Do("noise", func() (interface{}, error) {
+		k, err := c.getMachinePrivKey()
+		if err != nil {
+			return nil, err
+		}
+
+		return newNoiseClient(k, serverNoiseKey, c.serverURL)
+	})
+	c.mu.Lock()
+	c.noiseClient = np.(*noiseClient)
```
bradfitz, line 1245: this will panic if nip is nil on error.

You want `c.noiseClient, _ = ....`

bradfitz: But also this part can have multiple goroutines racing on it.

You want to do the assignment inside the `Do` func.

maisem: I'd rather just handle the error explicitly. Done

control/controlclient/direct.go
maisem: Done

control/controlclient/direct.go
bradfitz: you still want to do the assignment inside the Do

maisem: Moved to `Do`

control/controlclient/direct.go
```diff
@@ -1204,6 +1221,37 @@ func sleepAsRequested(ctx context.Context, logf logger.Logf, timeoutReset chan<-
...
+		k, err := c.getMachinePrivKey()
+		if err != nil {
+			return nil, err
+		}
+
+		nc, err = newNoiseClient(k, serverNoiseKey, c.serverURL)
+		if err != nil {
+			return nil, err
+		}
+		c.noiseClient = nc
```
bradfitz, line 1246: not enough. You still need to Lock/Unlock c.mu here. this is running in a separate goroutine and others can read it via a call to getNoiseClient.

maisem: This is surprisingly easy to get wrong :/

PR #4100, opened 2022-03-08
control/controlclient: fix Noise HTTP/2 regression from earlier commit

bradfitz:
Fix regression from 21069124db caught by tests in another repo.

The HTTP/2 Transport that was being returned had a ConnPool that never
dialed.

Updates #3488


maisem (approved):
I assume you've tested this works now

bradfitz:
Yup. I have an alias which does `go mod edit -replace tailscale.com=$HOME/src/tailscale.com` for testing the corp repo against HEAD oss.

PR #4102, opened 2022-03-08
control/controlclient: make RegisterRequest go over Noise

maisem:
Updates https://github.com/tailscale/tailscale/issues/3488

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -310,11 +310,18 @@ type loginOpt struct {
 	Logout bool
 }
 
+// httpClient provides a common interface for the noiseClient and
+// the NaCl box http.Client.
+type httpClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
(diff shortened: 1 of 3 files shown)


bradfitz (approved):

maisem:
Updated to pass in `request.Version` for Noise and added a commit to make `Auto.Shutdown` call `Direct.Close`. PTAL @bradfitz

control/controlclient/direct.go
```diff
@@ -310,11 +310,16 @@ type loginOpt struct {
 	Logout bool
 }
 
+type httpClient interface {
```
bradfitz, line 313: I guess if you didn't use a RoundTripper that means you meant for this to be http.Client-like and follow redirects?

bradfitz (approved):

control/controlclient/direct.go
maisem: added a comment, it's not meant to stop redirects (which should not be happening anyways, maybe that belongs in a different layer though). The goal with this is to just have a single interface that can be used between noise and nacl

PR #4108, opened 2022-03-08
1.22 picks

bradfitz:


DentonGentry:
Does the toolchain update deliver something compelling?
Feel free to Slack if not appropriate for a public issue.

I ask because patch releases get far less testing, if there is a regression somewhere we are unlikely to catch it.

bradfitz:
Full diff is here:
https://github.com/tailscale/go/compare/fd6b85b80f5fbbfaddef34e7ab6ac83384c50e4d..dce70b6d327c7a30b81701f4cc134b56c4e6c229

And bugs: https://github.com/golang/go/issues?q=milestone%3AGo1.17.8+label%3ACherryPickApproved

Notes here:
https://go.dev/doc/devel/release#go1.17.minor

> go1.17.8 (released 2022-03-03) includes a security fix to the regexp/syntax package, as well as bug fixes to the compiler, runtime, the go command, and the crypto/x509, and net packages. See the Go 1.17.8 milestone on our issue tracker for details.

Thoughts on each:
* We don't use regexp
* We do have RISC-V builds
* https://github.com/golang/go/issues/51153 might fix some DNS issues somewhere for us
* don't know about https://github.com/golang/go/issues/48171 
* https://github.com/golang/go/issues/50732 doesn't impact users
* the cmd/go change doesn't affect us.

The diff is so tiny, I'd do it just to be current and have the latest number shown so it doesn't look like we're behind on security updates.

DentonGentry (approved):

bradfitz:
Oh, and it seems that our tree doesn't have the x509 change anyway. Upstream Go has it, but not us:

https://github.com/golang/go/compare/go1.17.7..go1.17.8#diff-e3cea16ed73b68bdaa41bd72dffb3cbbabe2528712d72080de7e4def46c0dbf4

/cc @josharian

maisem (approved):

bradfitz:
> Maybe add https://github.com/tailscale/tailscale/commit/d9a7205be5470f6ea03944c76e2ee2b5a1906041. Others LGTM

The risk/reward seems off on that one to me.

PR #4111, opened 2022-03-08
control/controlhttp: send expected control public key in upgrade request

bradfitz:
So we can do key rotation later and have small windows of overlapping
valid server keys.

Updates #3488

### DIFF
--- a/control/controlhttp/server.go
+++ b/control/controlhttp/server.go
@@ -44,6 +44,18 @@ func AcceptHTTP(ctx context.Context, w http.ResponseWriter, r *http.Request, pri
 		return nil, fmt.Errorf("decoding base64 handshake header: %v", err)
 	}
 
+	if wantPub := r.Header.Get(serverPubHeaderName); wantPub != "" {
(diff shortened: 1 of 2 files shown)


danderson (changes requested):
<don't mind me, I read this change wrong>

control/controlhttp/server.go
```diff
@@ -44,6 +44,18 @@ func AcceptHTTP(ctx context.Context, w http.ResponseWriter, r *http.Request, pri
 		return nil, fmt.Errorf("decoding base64 handshake header: %v", err)
 	}
 
+	if wantPub := r.Header.Get(serverPubHeaderName); wantPub != "" {
+		// If the client declared the public key they expect to speak to,
+		// check it.
+		// TODO: replace the 'private' parameter with a func/interface
+		// that looks up the private key as a functio of the public key
```
DentonGentry, line 51: "functio" missing an n

bradfitz:
We did 02ad987e24bccab4b20e6a5ee67dd0a42eb6844f instead.

PR #4117, opened 2022-03-09
Allow using tailscale cli with gokrazy

joneskoo:
Enable use of command line arguments with tailscale cli on gokrazy. Before this change using arguments like "up" would cause tailscale cli to be repeatedly restarted by gokrazy process supervisor.

We never want to have gokrazy restart tailscale cli, even if user would manually start the process.

Expected usage is that user creates files:

`flags/tailscale.com/cmd/tailscale/flags.txt`:

    up

`flags/tailscale.com/cmd/tailscaled/flags.txt`:

    --statedir=/perm/tailscaled/
    --tun=userspace-networking

Then tailscale prints URL for user to log in with browser.

Alternatively it should be possible to use up with auth key to allow unattended gokrazy installs.

cc @bradfitz @stapelberg

### DIFF
--- a/cmd/tailscale/cli/cli.go
+++ b/cmd/tailscale/cli/cli.go
@@ -125,16 +125,21 @@ func CleanUpArgs(args []string) []string {
 }
 
 // Run runs the CLI. The args do not include the binary name.
-func Run(args []string) error {
+func Run(args []string) (err error) {
 	if len(args) == 1 && (args[0] == "-V" || args[0] == "--version") {
 		args = []string{"version"}
 	}
-	if runtime.GOOS == "linux" && distro.Get() == distro.Gokrazy && len(args) == 0 &&
+	if runtime.GOOS == "linux" && distro.Get() == distro.Gokrazy &&
 		os.Getenv("GOKRAZY_FIRST_START") == "1" {
-		// Exit with 125 otherwise the CLI binary is restarted
-		// forever in a loop by the Gokrazy process supervisor.
-		// See https://gokrazy.org/userguide/process-interface/
-		os.Exit(125)
+		defer func() {
+			// Exit with 125 otherwise the CLI binary is restarted
+			// forever in a loop by the Gokrazy process supervisor.
+			// See https://gokrazy.org/userguide/process-interface/
+			if err != nil {
+				log.Println(err)
+			}
+			os.Exit(125)
+		}()
 	}
 
 	var warnOnce sync.Once
@@ -201,7 +206,7 @@ change in the future.
 		}
 	})
 
-	err := rootCmd.Run(context.Background())
+	err = rootCmd.Run(context.Background())
 	if tailscale.IsAccessDeniedError(err) && os.Getuid() != 0 && runtime.GOOS != "windows" {
 		return fmt.Errorf("%v\n\nUse 'sudo tailscale %s' or 'tailscale up --operator=$USER' to not require root.", err, strings.Join(args, " "))
 	}


bradfitz:
Please rewrite commit message to match normal style (see git log). Same as Go std style.

`cmd/tailscale: allow using ...`

cmd/tailscale/cli/cli.go
```diff
@@ -125,16 +125,20 @@ func CleanUpArgs(args []string) []string {
...
-		os.Exit(125)
+	if runtime.GOOS == "linux" && distro.Get() == distro.Gokrazy {
+		defer func() {
+			// Exit with 125 otherwise the CLI binary is restarted
+			// forever in a loop by the Gokrazy process supervisor.
+			// See https://gokrazy.org/userguide/process-interface/
+			if err != nil {
+				log.Println(err)
+			}
+			os.Exit(125)
```
bradfitz, line 140: Can we only do this when our parent is the gokrazy process supervisor and not the breakglass shell?

If I'm in the breakglass shell, I don't want any special behavior here.

The `GOKRAZY_FIRST_START` condition provided that property before, but now it's gone.

/cc @stapelberg

stapelberg: You could just keep the GOKRAZY_FIRST_START check around, I think?

bradfitz: Yeah, works for me.

joneskoo: Fixed and commit message updated. All good?

stapelberg (approved):

bradfitz (approved):

bradfitz:
Merged & cherry-picked to 1.22.x so this'll be in the upcoming 1.22.1 release.

bradfitz:
And btw, these flags should no longer be necessary:

```
--statedir=/perm/tailscaled/
--tun=userspace-networking
```

The gokrazy docs can be simplified once 1.22.1 is out.

stapelberg:
Great, thanks @bradfitz!

oliverpool:
I am running tailscale, via gokrazy on a 1st gen raspberry pi and I have an issue, likely because the pi is so slow:
```
2022/04/02 15:13:52 failed to connect to local tailscaled (which appears to be running as tailscaled, pid 361). Got error: Failed to connect to local Tailscale daemon for /localapi/v0/status; not running? Error: dial unix /perm/tailscaled/tailscaled.sock: connect: no such file or directory
2022/04/02 15:13:52 gokrazy: process should not be supervised, stopping
2022/04/02 15:13:52 gokrazy: exit status 125
```

When I manually restart the process, it runs fine (because tailscaled had by then enough time to start).

I think that changing the behavior in `tailscale/cmd/tailscale/tailscale.go` would solve this issue: only exit with status 125 if there was no error (independently of `GOKRAZY_FIRST_START`)
https://github.com/tailscale/tailscale/blob/9f604f2bd3b48fc4464727b7ac6f5ee56d02413c/cmd/tailscale/tailscale.go#L24-L27

```go
 if err := cli.Run(args); err != nil { 
 	fmt.Fprintln(os.Stderr, err) 
 	os.Exit(1) 
 } else if runtime.GOOS == "linux" && distro.Get() == distro.Gokrazy {
	// Exit with 125 otherwise the CLI binary is restarted
	// forever in a loop by the Gokrazy process supervisor.
	// See https://gokrazy.org/userguide/process-interface/
	os.Exit(125) 
}
```

I think this would be cleaner:
- `os.Exit` is called only in the `main` package
- special case for gokrazy is still in only one file (`tailscale.go` instead of `cli.go`)
- only a special handling when there is no error (cc @bradfitz)
> If I'm in the breakglass shell, I don't want any special behavior here.

Shall I craft a PR with this? (and try on my old pi, if it works?)

DentonGentry:
It would be fine to send a PR to do this, with this explanation in the commit message.

Note that we don't always pay attention to merged PRs, if you want additional correspondence on this please send the PR to comment on or open a new GitHub issue.

oliverpool:
Thinking more about this, I think the custom logic for gokrazy can be completely drop, once the gokrazy-supervisor is adjusted, not to restart services which successfully started: https://github.com/gokrazy/gokrazy/pull/127

> Note that we don't always pay attention to merged PRs

Thank you for the hint. I thought that I would comment here, to let readers have the full context (and notify the people who already discussed this). I would probably have opened a new issue, if you didn't reply here.

I will open a PR once the gokrazy supervisor supports this.

PR #4125, opened 2022-03-09
Update docs/k8s dockerfile to allow running the run.sh

phumberdroz:
The dockerfile as is does not work since you can not run the run.sh. 

This makes it a bit easier for other people to use this.

### DIFF
--- a/docs/k8s/Dockerfile
+++ b/docs/k8s/Dockerfile
@@ -4,4 +4,5 @@
 
 FROM ghcr.io/tailscale/tailscale:latest
 COPY run.sh /run.sh
+RUN chmod +x /run.sh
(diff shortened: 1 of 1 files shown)


maisem:
The file is already marked executable, which platform/OS are you using to build the Dockerfile?

maisem:
The file is already marked executable, which platform/OS are you using to build the Dockerfile?

phumberdroz:
MacOS and I downloaded it via the zip option (I did not clone it) I think that might have striped the file permissions

phumberdroz:
Feel free to close if you feel this is not needed. :)

maisem:
This is obsolete now as the `run.sh` file is in the container

PR #4127, opened 2022-03-09
control/controlbase: make Conn.Write return consumed bytes

### LINKED ISSUE #4126: control/controlbase: Conn.Write can return more succesful bytes written than input provided
We just noticed a `*bufio.Writer` crash writing to a `*controlbase.Conn` where the `*controlbase.Conn` returned (n, nil) where n was greater than the length of its provided input buffer to Write.

Specifically, it returned (4143, nil) for an input []byte of len 4105, which tripped up this part of bufio.Writer.Write:

```go
                if b.Buffered() == 0 {
                        // Large write, empty buffer. 
                        // Write directly from p to avoid copy. 
                        n, b.err = b.wr.Write(p)
....
                p = p[n:] // <----- panic is here 
```

/cc @danderson @maisem

maisem:
Currently `Write` returns the number of ciphertext bytes written.
According to the docs for io.Writer, Write should return the amount
of bytes consumed from the input.
```
// Write writes len(p) bytes from p to the underlying data stream.
// It returns the number of bytes written from p (0 <= n <= len(p))
// and any error encountered that caused the write to stop early.
// Write must return a non-nil error if it returns n < len(p).
// Write must not modify the slice data, even temporarily.
Write(p []byte) (n int, err error)
```

Fixes #4126

### DIFF
--- a/control/controlbase/conn.go
+++ b/control/controlbase/conn.go
@@ -267,18 +267,16 @@ func (c *Conn) Write(bs []byte) (n int, err error) {
 
 		ciphertext, err := c.encryptLocked(toSend)
 		if err != nil {
-			return 0, err
+			return sent, err
 		}
-
-		n, err := c.conn.Write(ciphertext)
-		sent += n
-		if err != nil {
+		if _, err := c.conn.Write(ciphertext); err != nil {
 			// Return the raw error on the Write that actually
 			// failed. For future writes, return that error wrapped in
 			// a desync error.
 			c.tx.err = errPartialWrite{err}
 			return sent, err
 		}
+		sent += len(toSend)
 	}
 	return sent, nil
 }


bradfitz (approved):

danderson (dismissed):
io.Writer is specified to always return n == len(input) on success, so the loop is unnecessary.

Additionally, the returned amount would correspond to the amount of ciphertext written, which is a different length to the plaintext, so I don't think this complies with the contract of io.Writer either way - it's returning an n that's unrelated to the input.

Why do you need this? This Conn cannot recover from a partial write because the cipher states have fallen out of sync, so there's nothing actionable the caller can do with the value of n, even if it was accurate (which it isn't, the ciphertext is longer than the plaintext).

control/controlbase/conn.go
```diff
@@ -267,18 +267,20 @@ func (c *Conn) Write(bs []byte) (n int, err error) {
...
-		n, err := c.conn.Write(ciphertext)
-		sent += n
-		if err != nil {
-			// Return the raw error on the Write that actually
-			// failed. For future writes, return that error wrapped in
-			// a desync error.
-			c.tx.err = errPartialWrite{err}
 			return sent, err
 		}
+		for len(ciphertext) > 0 {
```
bradfitz, line 272: yeah, this loop isn't needed.

maisem:
> io.Writer is specified to always return n == len(input) on success, so the loop is unnecessary.

Which loop?

> Additionally, the returned amount would correspond to the amount of ciphertext written, which is a different length to the plaintext, so I don't think this complies with the contract of io.Writer either way - it's returning an n that's unrelated to the input.

Previously it was returning the amount of ciphertext written, now it returns the amount of plaintext written.

> Why do you need this? This Conn cannot recover from a partial write because the cipher states have fallen out of sync, so there's nothing actionable the caller can do with the value of n, even if it was accurate (which it isn't, the ciphertext is longer than the plaintext).

See https://github.com/tailscale/tailscale/issues/4126. We are currently seeing panics

```
panic: runtime error: slice bounds out of range [4143:4105]

goroutine 275 [running]:
bufio.(*Writer).Write(0xc0002fc740, {0xc000164000, 0xc000144650, 0x44fb94})
	bufio/bufio.go:654 +0x1bd
golang.org/x/net/http2.(*bufferedWriter).Write(0xc0001dc0c0, {0xc000164000, 0x1009, 0x1300})
	golang.org/x/net@v0.0.0-20220127200216-cd36cc0744dd/http2/http2.go:280 +0xd6
golang.org/x/net/http2.(*Framer).endWrite(0xc00032c2a0)
	golang.org/x/net@v0.0.0-20220127200216-cd36cc0744dd/http2/frame.go:371 +0xc5
golang.org/x/net/http2.(*Framer).WriteDataPadded(0xc00032c2a0, 0x1768818, 0x0, {0xc0007d8000, 0x1000, 0xc0001dc0c0}, {0x0, 0x0, 0xc000144798})
	golang.org/x/net@v0.0.0-20220127200216-cd36cc0744dd/http2/frame.go:694 +0x38a
golang.org/x/net/http2.(*Framer).WriteData(...)
	golang.org/x/net@v0.0.0-20220127200216-cd36cc0744dd/http2/frame.go:652
golang.org/x/net/http2.(*writeData).writeFrame(0xc000347740, {0x19ab838, 0xc0002fe180})
	golang.org/x/net@v0.0.0-20220127200216-cd36cc0744dd/http2/write.go:109 +0x53
golang.org/x/net/http2.(*serverConn).writeFrameAsync(0xc0002fe180, {{0x19981b0, 0xc000347740}, 0xc0001a6510, 0xc000020780})
	golang.org/x/net@v0.0.0-20220127200216-cd36cc0744dd/http2/server.go:781 +0x67
created by golang.org/x/net/http2.(*serverConn).startFrameWrite
	golang.org/x/net@v0.0.0-20220127200216-cd36cc0744dd/http2/server.go:1184 +0x359
FAIL	tailscale.io/control/controlclient	0.067s
FAIL
```

bradfitz (approved):

PR #4135, opened 2022-03-10
ssh/tailssh, ipnlocal, controlclient: fetch next SSHAction from network

bradfitz:
Updates #3802


ssh/tailssh/tailssh.go
```diff
@@ -200,18 +202,32 @@ func (srv *server) handleSSH(s ssh.Session) {
...
-	if action.Reject {
-		logf("ssh: access denied for %q from %v", ci.uprof.LoginName, ci.src.IP())
-		s.Exit(1)
-		return
-	}
-	if !action.Accept || action.HoldAndDelegate != "" {
-		fmt.Fprintf(s, "TODO: other SSHAction outcomes")
-		s.Exit(1)
-		return
+ProcessAction:
```
maisem, line 205: Can this keep waiting forever? looks like it can, add comment and/or way to make it not.

bradfitz: fetchSSHAction returns an error on context timeout

maisem: yes, but on success it can return an Action which points to the same HoldAndDelegate URL. So it'll keep fetching the same thing over and over again

bradfitz: Done.

maisem (approved):

PR #4142, opened 2022-03-10
ssh/tailssh: start login shell, fix arg passing, width/height mismatch

bradfitz:
Updates #3802

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -69,11 +69,14 @@ func newIncubatorCommand(ctx context.Context, ci *sshConnInfo, lu *user.User, ta
 		"--remote-user=" + remoteUser,
 		"--remote-ip=" + ci.src.IP().String(),
 		"--cmd=" + name,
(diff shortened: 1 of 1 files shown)


maisem (approved):

ssh/tailssh/incubator.go
```diff
@@ -157,8 +156,9 @@ func beIncubator(args []string) error {
 func (srv *server) launchProcess(ctx context.Context, s ssh.Session, ci *sshConnInfo, lu *user.User) (cmd *exec.Cmd, stdin io.WriteCloser, stdout, stderr io.Reader, err error) {
 	shell := loginShell(lu.Uid)
 	var args []string
+	args = append(args, "-l")
```
bradfitz, line 159: oh, is this right in the RawCommand case? I guess we don't want a login shell when we run a command externally, @maisem?

maisem: yeah we don't. Only do this when RawCommand is empty.

Issue #4146, opened 2022-03-11
ssh/tailssh: set pty termios settings


bradfitz:
Our SSH server isn't yet setting the pty termio settings. Using `stty -a`, you can see the difference. This Go-based JSON-supporting stty makes it easier to see:

https://github.com/u-root/u-root/tree/main/cmds/core/stty

Diff from normal to ours:

```diff
bradfitz@tsdev:~$ diff -u stty-normal.json  stty-ts.json
--- stty-normal.json	2022-03-11 07:45:12.833294697 -0800
+++ stty-ts.json	2022-03-11 07:45:07.765250539 -0800
@@ -1,12 +1,12 @@
 {
 	"Ispeed": 0,
 	"Ospeed": 0,
-	"Row": 24,
-	"Col": 80,
+	"Row": 62,
+	"Col": 143,
 	"CC": {
 		"eof": 4,
-		"eol": 255,
-		"eol2": 255,
+		"eol": 0,
+		"eol2": 0,
 		"erase": 127,
 		"intr": 3,
 		"kill": 21,
@@ -27,7 +27,7 @@
 		"echo": true,
 		"echoctl": true,
 		"echoe": true,
-		"echok": false,
+		"echok": true,
 		"echoke": true,
 		"echonl": false,
 		"echoprt": false,
@@ -39,14 +39,14 @@
 		"ignbrk": false,
 		"igncr": false,
 		"ignpar": false,
-		"imaxbel": true,
+		"imaxbel": false,
 		"inlcr": false,
 		"inpck": false,
 		"isig": true,
 		"istrip": false,
 		"iuclc": false,
-		"iutf8": true,
-		"ixany": true,
+		"iutf8": false,
+		"ixany": false,
 		"ixoff": false,
 		"ixon": true,
 		"noflsh": false,
@@ -61,7 +61,7 @@
 		"parenb": false,
 		"parmrk": false,
 		"parodd": false,
-		"pendin": true,
+		"pendin": false,
 		"tostop": false,
 		"xcase": false
 	}
```

(Ignore the Row/Col differences, of course)

/cc @maisem

bradfitz:
Even if I `stty dump` the JSON of a normal OpenSSH session's pty and `stty load` it in Tailscale session's pty, then start a new GNU Screen session and switch back and forth between two windows where one is Emacs in go-mode showing indented code in color, it gets all corrupted (loses indentation, etc) when switching back to Emacs.

So something beyond just the termios, it seems.

[...]

PR #4151, opened 2022-03-11
ssh/tailssh: make pty termios options match OpenSSH

bradfitz:
Still not sure the exact rules of how/when/who's supposed to set
these, but this works for now on making them match. Baby steps.
Will research more and adjust later.

Updates #4146 (but not enough to fix it, something's still wrong)
Updates #3802


ssh/tailssh/incubator.go
```diff
@@ -209,12 +210,37 @@ func startWithPTY(cmd *exec.Cmd, ptyReq ssh.Pty) (ptyFile *os.File, err error) {
...
+		tios.CC["eol"] = 255
+		tios.CC["eol2"] = 255
+		tios.Opts["echok"] = false
+		tios.Opts["imaxbel"] = true
+		tios.Opts["iutf8"] = true
+		tios.Opts["ixany"] = true
+		tios.Opts["pendin"] = true
+		if _, err := tios.STTY(int(fd)); err != nil {
+			ctlErr = fmt.Errorf("STTY: %w", err)
+			return
```
maisem, line 237: Add a comment describing this pls

maisem (approved):

ssh/tailssh/incubator.go
bradfitz: done

maisem (approved):

PR #4153, opened 2022-03-12
ssh/tailssh: set DBUS_SESSION_BUS_ADDRESS, SSH_TTY and server version

maisem:
- ssh/tailssh: start sending the server version
- ssh/tailssh: set DBUS_SESSION_BUS_ADDRESS and SSH_TTY variables

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"os/user"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"syscall"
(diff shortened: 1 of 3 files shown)


ssh/tailssh/incubator.go
```diff
@@ -262,6 +262,7 @@ func startWithPTY(cmd *exec.Cmd, ptyReq ssh.Pty) (ptyFile *os.File, err error) {
 	updateStringInSlice(cmd.Args, "--has-tty=false", "--has-tty=true")
 	if ptyName, err := ptyName(ptyFile); err == nil {
 		updateStringInSlice(cmd.Args, "--tty-name=", "--tty-name="+ptyName)
+		cmd.Env = append(cmd.Env, fmt.Sprintf("SSH_TTY=/dev%s", ptyName))
```
bradfitz, line 265: Is this missing a slash between dev and %s?

ssh/tailssh/tailssh.go
```diff
@@ -67,6 +67,7 @@ func (srv *server) newSSHServer() (*ssh.Server, error) {
 		ChannelHandlers: map[string]ssh.ChannelHandler{
 			"direct-tcpip": ssh.DirectTCPIPHandler,
 		},
+		Version:                     "SSH-2.0-tailssh_0.1",
```
bradfitz, line 70: Let's drop the 0.1 part and just say "SSH-2.0-Tailscale" at most?

bradfitz (approved):

ssh/tailssh/incubator.go
maisem: No, but it's cleaner to use `filepath.Join` here.

Issue #4129, opened 2022-03-09
ssh/tailssh: add policy config for port forwarding


bradfitz:
Make the tailcfg.SSHPolicy support enabling vs disabling port forwarding

filviu:
Silly question - this looks merged for over a year yet I can't find any documentation (maybe I'm overlooking something obvious).

I was searching for this as it's a blocker for us, mainly due to management requirements to switch to tailscale ssh on some machines.

Thanks!

PR #4157, opened 2022-03-13
tailcfg: add field to allow LocalPortForwarding in SSHAction

maisem:
Updates #3802, #4129

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -69,7 +69,7 @@ func (srv *server) newSSHServer() (*ssh.Server, error) {
 			"direct-tcpip": ssh.DirectTCPIPHandler,
 		},
 		Version:                     "SSH-2.0-Tailscale",
-		LocalPortForwardingCallback: srv.portForward,
+		LocalPortForwardingCallback: srv.mayForwardLocalPortTo,
 	}
 	for k, v := range ssh.DefaultRequestHandlers {
 		ss.RequestHandlers[k] = v
@@ -103,11 +103,15 @@ type server struct {
 
 var debugPolicyFile = envknob.String("TS_DEBUG_SSH_POLICY_FILE")
 
-// portForward reports whether the ctx should be allowed to port forward
+// mayForwardLocalPortTo reports whether the ctx should be allowed to port forward
 // to the specified host and port.
 // TODO(bradfitz/maisem): should we have more checks on host/port?
-func (srv *server) portForward(ctx ssh.Context, destinationHost string, destinationPort uint32) bool {
-	return srv.isActiveSession(ctx)
+func (srv *server) mayForwardLocalPortTo(ctx ssh.Context, destinationHost string, destinationPort uint32) bool {
+	ss, ok := srv.getSessionForContext(ctx)
+	if !ok {
+		return false
+	}
+	return ss.action.AllowLocalPortForwarding
 }
 
 // sshPolicy returns the SSHPolicy for current node.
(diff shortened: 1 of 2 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -96,16 +96,17 @@ type server struct {
...
 
-// portForward reports whether the ctx should be allowed to port forward
+// localPortForward reports whether the ctx should be allowed to port forward
 // to the specified host and port.
 // TODO(bradfitz/maisem): should we have more checks on host/port?
-func (srv *server) portForward(ctx ssh.Context, destinationHost string, destinationPort uint32) bool {
-	return srv.isActiveSession(ctx)
+func (srv *server) localPortForward(ctx ssh.Context, destinationHost string, destinationPort uint32) bool {
+	a := srv.sessionAction(ctx)
+	return a.LocalPortForwarding
```
bradfitz, line 109: I was going to say your API change on line 313 below was prone to misuse and this here looks what I was afraid of: nil ptr derefs after blind use trusting it'll return the right thing.

tailcfg/tailcfg.go
```diff
@@ -1624,6 +1624,10 @@ type SSHAction struct {
 	// response, it should be re-fetched as long as the SSH
 	// session is open.
 	HoldAndDelegate string `json:"holdAndDelegate,omitempty"`
+
+	// LocalPortForwarding, if true, allows accepted connections
+	// to optionally use local port forwarding.
```
bradfitz, line 1629: to use local port forwarding if requested.

?

tailcfg/tailcfg.go
```diff
@@ -1624,6 +1624,10 @@ type SSHAction struct {
...
+	LocalPortForwarding bool `json:"localPortForwarding,omitempty"`
```
bradfitz, line 1630: How about "AllowLocalPortForwarding" ?

ssh/tailssh/tailssh.go
maisem: Sigh, yes. Fixed

tailcfg/tailcfg.go
maisem: sure

tailcfg/tailcfg.go
maisem: sure

ssh/tailssh/tailssh.go
```diff
@@ -96,16 +96,17 @@ type server struct {
...
 	// mu protects activeSessions.
 	mu             sync.Mutex
-	activeSessions map[string]bool
+	activeSessions map[string]*tailcfg.SSHAction
 }
 
 var debugPolicyFile = envknob.String("TS_DEBUG_SSH_POLICY_FILE")
 
-// portForward reports whether the ctx should be allowed to port forward
+// localPortForward reports whether the ctx should be allowed to port forward
```
bradfitz, line 104: doesn't sound like a boolean func.

how about `mayForwardLocalPortTo`

ssh/tailssh/tailssh.go
```diff
@@ -96,16 +96,17 @@ type server struct {
...
 // to the specified host and port.
 // TODO(bradfitz/maisem): should we have more checks on host/port?
-func (srv *server) portForward(ctx ssh.Context, destinationHost string, destinationPort uint32) bool {
-	return srv.isActiveSession(ctx)
+func (srv *server) localPortForward(ctx ssh.Context, destinationHost string, destinationPort uint32) bool {
```
bradfitz, line 107: I don't think this method should be on `*server`.

I think we need a new type for an established session, that `launchProcess` or something earlier should return.

ssh/tailssh/tailssh.go
```diff
@@ -256,7 +257,7 @@ ProcessAction:
 		})
 		defer t.Stop()
 	}
-	srv.handleAcceptedSSH(ctx, s, ci, lu)
+	srv.handleAcceptedSSH(ctx, s, action, ci, lu)
```
bradfitz, line 260: yeah, this suggests we should have a new type before this, and then `handleAcceptedSSH` can be renamed to `run` or something on that new type.

tailcfg/tailcfg.go
```diff
@@ -1624,6 +1624,10 @@ type SSHAction struct {
 	// response, it should be re-fetched as long as the SSH
 	// session is open.
 	HoldAndDelegate string `json:"holdAndDelegate,omitempty"`
+
+	// AllowLocalPortForwarding, if true, allows accepted connections
```
bradfitz, line 1628: this part in this file LGTM though

bradfitz (changes requested):

bradfitz (approved):

PR #4158, opened 2022-03-13
ssh/tailssh: start of implementing optional session recording

bradfitz:
To asciinema cast format.

Updates #3802

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -15,12 +15,14 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
 	"os/exec"
 	"os/user"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
@@ -432,11 +434,16 @@ func (ss *sshSession) handleSSHAgentForwarding(s ssh.Session, lu *user.User) err
 	return nil
 }
 
+// recordSSH is a temporary dev knob to test the SSH recording
+// functionality and support off-node streaming.
+//
+// TODO(bradfitz,maisem): move this to SSHPolicy.
+var recordSSH = envknob.Bool("TS_DEBUG_LOG_SSH")
+
 // run is the entrypoint for a newly accepted SSH session.
 //
-// When ctx is done, the session is forcefully terminated. If its Err
-// is an SSHTerminationError, its SSHTerminationMessage is sent to the
-// user.
+// It handles ss once it's been accepted and determined
+// that it should run.
 func (ss *sshSession) run() {
 	srv := ss.srv
 	srv.startSession(ss)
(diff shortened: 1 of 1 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -332,6 +334,8 @@ func (srv *server) endSession(s ssh.Session) {
 	delete(srv.activeSessions, s.Context().(ssh.Context).SessionID())
 }
 
+var recordSSH = envknob.Bool("TS_LOG_SSH")
```
maisem, line 337: Make this part of the action?

ssh/tailssh/tailssh.go
```diff
@@ -357,6 +361,24 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
 	// See https://github.com/tailscale/tailscale/issues/4146
 	s.DisablePTYEmulation()
 
+	var rec *recording // or nil if disabled
+	var err error
+	ptyReq, _, isPtyReq := s.Pty()
+	if recordSSH && isPtyReq { // for now only record pty sessions
+		env := envMap(s.Environ())
```
maisem, line 368: Should this be cmd.Env?

ssh/tailssh/tailssh.go
```diff
@@ -357,6 +361,24 @@ func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *ssh
 	// See https://github.com/tailscale/tailscale/issues/4146
 	s.DisablePTYEmulation()
 
+	var rec *recording // or nil if disabled
+	var err error
+	ptyReq, _, isPtyReq := s.Pty()
+	if recordSSH && isPtyReq { // for now only record pty sessions
```
maisem, line 367: Can you move all the recording setup to a different function. Maybe have it return a ReadWriteCloser?

ssh/tailssh/tailssh.go
```diff
@@ -509,3 +531,138 @@ func matchesPrincipal(ps []*tailcfg.SSHPrincipal, ci *sshConnInfo) bool {
...
+		return nil, err
+	}
+	f, err := ioutil.TempFile(dir, fmt.Sprintf("ssh-session-%v-*.cast", now.UnixNano()))
+	if err != nil {
+		return nil, err
+	}
+	rec.out = f
+
+	// {"version": 2, "width": 221, "height": 84, "timestamp": 1647146075, "env": {"SHELL": "/bin/bash", "TERM": "screen"}}
+	type CastHeader struct {
```
maisem, line 560: Why in-line?

ssh/tailssh/tailssh.go
bradfitz: I should name this TS_TEMP_DEBUG_LOG_SSH or something; the real recording mechanism will stream off-device. Otherwise it's not secure having the machine being sshed into also storing the recordings.

This whole PR is just a baby step towards the mechanism itself but isn't the final product.

ssh/tailssh/tailssh.go
bradfitz: Probably, yeah, but it requires rearranging a bunch of stuff. We want to get this env before we start the process.

Another reason to make a new type.

ssh/tailssh/tailssh.go
bradfitz: Why scoped wider? Does anything read this?

ssh/tailssh/tailssh.go
bradfitz: Read?

And it needs to make two writer streams below. So, WriteCloser isn't enough.

There's definitely refactoring I want to do, but not necessarily to this part.

bradfitz:
I won't submit this until we've done some refactoring to the ssh package to make a new `acceptedSession` type or something.

bradfitz:
PTAL; rebased after all the `sshSession` stuff.

I think this is ready now as a checkpoint. Plenty of work remains later to tie it into SSHPolicy and stream it off-node, etc, but I'd rather submit in small chunks.

maisem (approved):

PR #4161, opened 2022-03-13
wip ssh refactor

maisem:
Signed-off-by: Maisem Ali <maisem@tailscale.com>

PR #4162, opened 2022-03-13
ssh/tailssh: add a new sshSession type to clean up existing+future code

bradfitz:
Updates #3802


bradfitz:
(Merge of @maisem's #4161 and my similar WIP branch.)

ssh/tailssh/incubator.go
```diff
@@ -158,31 +162,32 @@ func beIncubator(args []string) error {
...
+	shell := loginShell(ss.localUser.Uid)
 	var args []string
-	if rawCmd := s.RawCommand(); rawCmd != "" {
+	if rawCmd := ss.RawCommand(); rawCmd != "" {
 		args = append(args, "-c", rawCmd)
 	} else {
 		args = append(args, "-l") // login shell
 	}
-	ptyReq, winCh, isPty := s.Pty()
+	ptyReq, winCh, isPty := ss.Pty()
```
maisem, line 173: move this down, near usage

ssh/tailssh/incubator.go
```diff
@@ -158,31 +162,32 @@ func beIncubator(args []string) error {
...
 	)
-	srv.logf("ssh: starting: %+v", cmd.Args)
+	ss.logf("ssh: starting: %+v", cmd.Args)
 
 	if !isPty {
 		stdin, stdout, stderr, err = startWithStdPipes(cmd)
 		return
 	}
-	pty, err := srv.startWithPTY(cmd, ptyReq)
+	pty, err := ss.srv.startWithPTY(cmd, ptyReq)
```
maisem, line 190: can you move this to be function on session?

ssh/tailssh/incubator.go
```diff
@@ -54,11 +54,15 @@ var maybeStartLoginSession = func(logf logger.Logf, uid uint32, localUser, remot
 
 // newIncubatorCommand returns a new exec.Cmd configured with
 // `tailscaled be-child ssh` as the entrypoint.
-// If tailscaled is empty, the desired cmd is executed directly.
-func newIncubatorCommand(ctx context.Context, ci *sshConnInfo, lu *user.User, tailscaled, name string, args []string) *exec.Cmd {
-	if tailscaled == "" {
+//
+// If ss.srv.tailscaledPath is empty, this method is equivalent to
```
maisem, line 58: I didn't want to pass the server down to session, well especially not have it access private fields. You can either make sshSession also have `tailscaledPath`, or create a processLauncher

ssh/tailssh/tailssh.go
```diff
@@ -244,19 +245,37 @@ ProcessAction:
...
 	}
-	srv.handleAcceptedSSH(ctx, s, ci, lu)
+	ss.run()
+}
+
+// sshSession is an accepted Tailscale SSH session.
+type sshSession struct {
+	ssh.Session
+	idH      string // the RFC4253 sec8 hash H; don't share outside process
+	sharedID int64  // ID that's shared with control; time.Now().UnixNano()
```
maisem, line 265: why not just use a random ID here instead of UnixNano?

ssh/tailssh/tailssh.go
```diff
@@ -244,19 +245,37 @@ ProcessAction:
...
+	logf     logger.Logf
+
+	ctx       *sshContext // implements context.Context
+	srv       *server
+	connInfo  *sshConnInfo
+	action    *tailcfg.SSHAction
+	localUser *user.User
+
+	cmd *exec.Cmd
```
maisem, line 274: add a comment on when this is filled in etc

ssh/tailssh/tailssh.go
```diff
@@ -290,20 +309,20 @@ func (srv *server) fetchSSHAction(ctx context.Context, url string) (*tailcfg.SSH
...
 		if serr, ok := err.(SSHTerminationError); ok {
 			msg := serr.SSHTerminationMessage()
 			if msg != "" {
-				io.WriteString(s.Stderr(), "\r\n\r\n"+msg+"\r\n\r\n")
+				io.WriteString(ss.Stderr(), "\r\n\r\n"+msg+"\r\n\r\n")
 			}
 		}
-		srv.logf("terminating SSH session from %v: %v", ci.src.IP(), err)
-		cmd.Process.Kill()
+		ss.srv.logf("terminating SSH session from %v: %v", ss.connInfo.src.IP(), err)
```
maisem, line 324: ss.logf

ssh/tailssh/tailssh.go
```diff
@@ -312,75 +331,107 @@ func (srv *server) handleSessionTermination(ctx context.Context, s ssh.Session,
...
+
+// run is the entrypoint for a newly accepted SSH session.
 //
 // When ctx is done, the session is forcefully terminated. If its Err
 // is an SSHTerminationError, its SSHTerminationMessage is sent to the
 // user.
-func (srv *server) handleAcceptedSSH(ctx context.Context, s ssh.Session, ci *sshConnInfo, lu *user.User) {
-	srv.startSession(s)
-	defer srv.endSession(s)
+func (ss *sshSession) run() {
```
maisem, line 382: keep handleAcceptedSSH so that it can munge with server level data and have it call ss.run.

ssh/tailssh/tailssh.go
```diff
@@ -312,75 +331,107 @@ func (srv *server) handleSessionTermination(ctx context.Context, s ssh.Session,
...
 			logf("ssh: can't switch to user %q from process euid %v", localUser, euid)
-			fmt.Fprintf(s, "can't switch user\n")
-			s.Exit(1)
+			fmt.Fprintf(ss, "can't switch user\n")
+			ss.Exit(1)
 			return
 		}
 	}
 
+	go ss.waitContext()
```
maisem, line 412: no point doing this if we are unable to launch the process, move it down after call to `launchProcess`

ssh/tailssh/incubator.go
maisem: processLaucher will actually help us test this code better in general.

ssh/tailssh/tailssh.go
```diff
@@ -290,20 +309,20 @@ func (srv *server) fetchSSHAction(ctx context.Context, url string) (*tailcfg.SSH
...
 			msg := serr.SSHTerminationMessage()
 			if msg != "" {
-				io.WriteString(s.Stderr(), "\r\n\r\n"+msg+"\r\n\r\n")
+				io.WriteString(ss.Stderr(), "\r\n\r\n"+msg+"\r\n\r\n")
 			}
 		}
-		srv.logf("terminating SSH session from %v: %v", ci.src.IP(), err)
-		cmd.Process.Kill()
+		ss.srv.logf("terminating SSH session from %v: %v", ss.connInfo.src.IP(), err)
+		ss.cmd.Process.Kill()
```
maisem, line 325: this might be nil the way it's written right now.

ssh/tailssh/incubator.go
bradfitz: That's misleading; it doesn't use anything from sshSession.

ssh/tailssh/incubator.go
bradfitz: Moved down a bit, even though that makes this diff bigger. I tried to keep the diff minimal during changes like this, so you can see the new type/usage with fewer distractions.

ssh/tailssh/incubator.go
bradfitz: I definitely _do_ want an up pointer back up to the server from sshSession. That's a very common pattern.  (e..g. peerAPIHandler up to peerAPIServer up to LocalBackend)

I don't want to copy a rando `tailscaledPath` into sshSession when it'll never vary per session or really have anything to do with a session.

ssh/tailssh/tailssh.go
bradfitz: I was thinking of sorting and lifetime uniqueness.

But I can do date+randhex.

ssh/tailssh/incubator.go
bradfitz: oh, okay, I see how it works now

ssh/tailssh/tailssh.go
bradfitz: It's awkward with the special context type. I added a constructor instead which helps the tests and breaks up the caller. How's that look?

bradfitz:
PTAL

bradfitz:
Whoops, @maisem approved in a video chat.

maisem:
yep, LGTM

PR #4172, opened 2022-03-15
github/windows: improve caching in -race builder

raggi:
Applying the same logic as for the non-race builder, using a dedicated cache key to avoid build cache trees from stomping each other.


raggi:
Before: 8m31s
After: 4m3s

bradfitz (approved):

PR #4190, opened 2022-03-17
syncs: use TryLock and TryRLock instead of unsafe

josharian:
The docs say:

Note that while correct uses of TryLock do exist, they are rare,
and use of TryLock is often a sign of a deeper problem in a particular use of mutexes.

Rare code! Or bad code! Who can tell!


bradfitz (approved):
Yay!

PR #4196, opened 2022-03-17
tailcfg: add Targets to SSHPolicy

maisem:
Updates #3802

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -1562,7 +1562,7 @@ type SSHRule struct {
 	RuleExpires *time.Time `json:"ruleExpires,omitempty"`
 
 	// Principals matches an incoming connection. If the connection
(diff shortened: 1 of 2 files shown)


tailcfg/tailcfg.go
```diff
@@ -1579,6 +1579,9 @@ type SSHRule struct {
 	// Action is the outcome to task.
 	// A nil or invalid action means to deny.
 	Action *SSHAction `json:"action"`
+
+	// Targets matches an incoming connection for the target destination.
```
bradfitz, line 1583: 1) this should go before the Action field.

but:

2) why would we sent a SSHPolicy down to a node talking about Targets under than the node we sent the policy too? The only Target that is applicable is the current node itself.

maisem:
spoke offline, not doing it this way.

PR #4198, opened 2022-03-17
tsnet: add some usability polish, remove WIP env var restriction

bradfitz:


tsnet/tsnet.go
```diff
@@ -94,17 +95,25 @@ func (s *Server) Start() error {
...
 
+// Close stops the server.
+//
+// It must not be called before or concurrently with Start.
+func (s *Server) Close() error {
+	s.shutdownCancel()
+	s.lb.Shutdown()
+	s.linkMon.Close()
+	return nil
+}
```
maisem, line 106: we should also close the `nettest.Listener`
It also overrides the global `tailscale.TailscaledDialer`, so maybe set that to the old value? I really want `tailscale.Client`, so callers can decide what to do

bradfitz: Done. 

About the dialer: oh well. Can clean later.

maisem (approved):

Issue #4201, opened 2022-03-17
net/tshttpproxy: winhttp: Open: The parameter is incorrect.


dblohm7:
### What is the issue?

We're passing `WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY` as the `dwAccessType` argument to [`WinHttpOpen`](https://docs.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpopen). Unfortunately that value is not yet supported pre-Windows 8.1, where we should be using the deprecated `WINHTTP_ACCESS_TYPE_DEFAULT_PROXY` flag instead.

### OS

Windows

bradfitz:
Another cherry-pick candidate?

/cc @DentonGentry

PR #4202, opened 2022-03-17
net/tshttpproxy: ensure we pass the correct flags to WinHttpOpen on W…

dblohm7:
…in7 and Win8.0

The best flag to use on Win7 and Win8.0 is deprecated in Win8.1, so we resolve
the flag depending on OS version info.

Fixes https://github.com/tailscale/tailscale/issues/4201


net/tshttpproxy/tshttpproxy_windows.go
```diff
@@ -146,20 +148,33 @@ func proxyFromWinHTTP(ctx context.Context, urlStr string) (proxy *url.URL, err e
...
 	winHTTP_AUTO_DETECT_TYPE_DHCP       = 0x00000001
 	winHTTP_AUTO_DETECT_TYPE_DNS_A      = 0x00000002
 )
 
+// Windows 8.1 is actually Windows 6.3 under the hood. Yay, marketing!
+const win8dot1Ver = "6.3"
+
+// We need to use a different flag for proxy resolution depending on whether
+// or not we're running < Windows 8.1
+var proxyFlag = func() uint32 {
```
bradfitz, line 164: we try to minimize init-time work. It makes binaries start slower and makes the linker unable to discard otherwise unused code.

I'd do this lazily as needed instead. You can use an atomic.Value to cache it between calls to winHTTPOpen.

dblohm7: ack

net/tshttpproxy/tshttpproxy_windows.go
```diff
@@ -146,20 +149,42 @@ func proxyFromWinHTTP(ctx context.Context, urlStr string) (proxy *url.URL, err e
...
 	winHTTP_AUTOPROXY_ALLOW_AUTOCONFIG  = 0x00000100
 	winHTTP_AUTOPROXY_AUTO_DETECT       = 1
 	winHTTP_AUTO_DETECT_TYPE_DHCP       = 0x00000001
 	winHTTP_AUTO_DETECT_TYPE_DNS_A      = 0x00000002
 )
 
+// Windows 8.1 is actually Windows 6.3 under the hood. Yay, marketing!
+const win8dot1Ver = "6.3"
+
+// We need to use a different accessType flag for proxy resolution depending on
```
bradfitz, line 163: // accessType is the XXXX parameter we pass to XXXX>
// We need to use a ....

bradfitz (approved):

bradfitz:
`make updatedeps` to fix depaware

PR #4205, opened 2022-03-17
derp: set Basic Constraints on metacert

bradfitz:
See https://github.com/golang/go/issues/51759#issuecomment-1071147836

Once we deploy this, tailscaled should work again for macOS users with
Go 1.18.

Updates golang/go#51759

### DIFF
--- a/derp/derp_server.go
+++ b/derp/derp_server.go
@@ -453,6 +453,9 @@ func (s *Server) initMetacert() {
 		// Windows requires NotAfter and NotBefore set:
 		NotAfter:  time.Now().Add(30 * 24 * time.Hour),
 		NotBefore: time.Now().Add(-30 * 24 * time.Hour),
+		// Per https://github.com/golang/go/issues/51759#issuecomment-1071147836,
+		// macOS requires BasicConstraints when subject == issuer:
+		BasicConstraintsValid: true,
 	}
 	cert, err := x509.CreateCertificate(crand.Reader, tmpl, tmpl, pub, priv)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


derp/derp_server.go
```diff
@@ -445,6 +450,22 @@ func (s *Server) initMetacert() {
...
 	}
+
+	type basicConstraints struct {
+		IsCA       bool `asn1:"optional"`
+		MaxPathLen int  `asn1:"optional"`
+	}
+	extEnc, err := asn1.Marshal(basicConstraints{IsCA: true, MaxPathLen: -1})
+	if err != nil {
+		log.Fatal(err)
+	}
```
maisem, line 461: Add a comment linking to https://pkg.go.dev/crypto/x509#Certificate.MaxPathLen.

maisem (approved):

derp/derp_server.go
```diff
@@ -453,7 +474,11 @@ func (s *Server) initMetacert() {
 		// Windows requires NotAfter and NotBefore set:
 		NotAfter:  time.Now().Add(30 * 24 * time.Hour),
 		NotBefore: time.Now().Add(-30 * 24 * time.Hour),
+		// macOS requires BasicConstraints when subject == issuer:
```
josharian, line 477: Colon here plus parens on next line is weird.

josharian (approved):

derp/derp_server.go
bradfitz: Done.

derp/derp_server.go
bradfitz: Actually simplified this a ton after reading those docs!

maisem (approved):

Issue #4194, opened 2022-03-17
Re-enable `go generate` and `staticcheck` CI


maisem:
`go generate` fails because of https://github.com/golang/go/issues/51629
~~`staticcheck` fails because of https://github.com/dominikh/go-tools/issues/1166~~

danderson:
One more upstream bug that breaks cmd/cloner: https://github.com/golang/go/issues/51629

PR #4206, opened 2022-03-17
.github/workflows: work around golang/go#51629

danderson:
Updates #4194

### DIFF
--- a/.github/workflows/go-generate-without-stringer.sh
+++ b/.github/workflows/go-generate-without-stringer.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env sh
+#
+# This is a temporary hack to work around
+# https://github.com/golang/go/issues/51629 , wherein the stringer
+# generator doesn't work with generics.
+#
+# This script is the equivalent of `go generate ./...`, except that it
+# only runs generate on packages that don't try to use stringer.
+
+set -e
+
+find . -name '*.go' | xargs grep -l go:generate | xargs -n1 dirname | sort -u | while read dir; do
+	if ! egrep "cmd/(stringer|cloner)" $dir/*.go; then
+		set -x
+		go generate -tags=hermetic $dir
+		set +x
+	fi
+done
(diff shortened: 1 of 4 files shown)


maisem:
I disabled the check yesterday, will have to enable it again from the UI

danderson:
Confused: we have a workflow for go_generate in .github/workflows, but it doesn't seem to run in CI?...

maisem:
I just renabled it, want to push again?

.github/workflows/go-generate-without-stringer.sh
```diff
@@ -0,0 +1,16 @@
...
+# This is a temporary hack to work around
+# https://github.com/golang/go/issues/51629 , wherein the stringer
+# generator doesn't work with generics.
+#
+# This script is the equivalent of `go generate ./...`, except that it
+# only runs generate on packages that don't try to use stringer.
+
+find . -name '*.go' | xargs grep -l go:generate | xargs -n1 dirname | sort -u | while read dir; do
+	if ! grep cmd/stringer $dir/*.go; then
+		set -x
```
maisem, line 12: Is this intended to be equivalent to `go generate || true`?

josharian (approved):

.github/workflows/go-generate-without-stringer.sh
danderson: Sort of. My intent is to skip go generate tasks that we know are going to fail, while still reporting failures for others. What I appear to have discovered, is that this workflow never complained if `go generate` fails (vs. succeeds but produces a diff), oops. Working on it.

net/dnsfallback/update-dns-fallbacks.go
```diff
@@ -19,6 +19,7 @@ import (
 )
 
 func main() {
+	log.Printf("RUN RUN RUN")
```
maisem, line 22: Debug?

maisem (approved):

net/dnsfallback/update-dns-fallbacks.go
danderson: Oops! Fixed.

PR #4208, opened 2022-03-17
cmd/proxy-to-grafana: use grafana's authproxy to log in tailnet users

nickoneill:
Moving this into cmd for now... can create a new repo if we decide to go in that direction. Now using the new whois rather than exec'ing status from the cli, and added tsnet so it can be easily served on 80/443.

@maisem I would like to include tls on this as well... the other tls work for the builder seems like a lot of boilerplate to configure 80->443 redirect, request a cert, etc. Does it makes sense to wrap that into some convenience that I we can reuse in the tsnet package?

### DIFF
--- a/cmd/proxy-to-grafana/proxy-to-grafana.go
+++ b/cmd/proxy-to-grafana/proxy-to-grafana.go
@@ -1,0 +1,154 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// proxy-to-grafana is a reverse proxy which identifies users based on their
+// originating Tailscale identity and maps them to corresponding Grafana
+// users, creating them if needed.
+//
+// It uses Grafana's AuthProxy feature:
+// https://grafana.com/docs/grafana/latest/auth/auth-proxy/
+//
+// Set the TS_AUTHKEY environment variable to have this server automatically
+// join your tailnet, or look for the logged auth link on first start.
+//
+// Use this Grafana configuration to enable the auth proxy:
+//
+// ```
+// [auth.proxy]
+// enabled = true
+// header_name = X-WEBAUTH-USER
+// header_property = username
+// auto_sign_up = true
+// whitelist = 127.0.0.1
+// headers = Name:X-WEBAUTH-NAME
+// enable_login_token = true
+// ```
+package main
+
+import (
+	"context"
+	"crypto/tls"
+	"flag"
+	"fmt"
+	"log"
+	"net"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"strings"
+	"time"
+
+	"tailscale.com/client/tailscale"
+	"tailscale.com/tailcfg"
+	"tailscale.com/tsnet"
+)
+
+var (
+	hostname     = flag.String("hostname", "", "Tailscale hostname to serve on, used as the base name for MagicDNS or subdomain in your domain alias for HTTPS.")
+	backendAddr  = flag.String("backend-addr", "", "Address of the Grafana server served over HTTP, in host:port format. Typically localhost:nnnn.")
+	tailscaleDir = flag.String("state-dir", "./", "Alternate directory to use for Tailscale state storage. If empty, a default is used.")
+	useHTTPS     = flag.Bool("use-https", false, "Serve over HTTPS via your *.ts.net subdomain if enabled in Tailscale admin.")
+)
+
+func main() {
+	flag.Parse()
+	if *hostname == "" || strings.Contains(*hostname, ".") {
+		log.Fatal("missing or invalid --hostname")
+	}
+	if *backendAddr == "" {
+		log.Fatal("missing --backend-addr")
+	}
+	ts := &tsnet.Server{
+		Dir:      *tailscaleDir,
+		Hostname: *hostname,
+	}
+
+	url, err := url.Parse(fmt.Sprintf("http://%s", *backendAddr))
+	if err != nil {
+		log.Fatalf("couldn't parse backend address: %v", err)
+	}
+
+	proxy := httputil.NewSingleHostReverseProxy(url)
+	originalDirector := proxy.Director
+	proxy.Director = func(req *http.Request) {
+		originalDirector(req)
+		modifyRequest(req)
+	}
+
+	var ln net.Listener
+	if *useHTTPS {
+		ln, err = ts.Listen("tcp", ":443")
+		ln = tls.NewListener(ln, &tls.Config{
+			GetCertificate: tailscale.GetCertificate,
+		})
+
+		go func() {
+			// wait for tailscale to start before trying to fetch cert names
+			for i := 0; i < 60; i++ {
+				st, err := tailscale.Status(context.Background())
+				if err != nil {
+					log.Fatal(err)
+				}
+				log.Printf("tailscale status: %v", st.BackendState)
+				if st.BackendState == "Running" {
+					break
+				}
+				time.Sleep(time.Second)
+			}
+
+			l80, err := ts.Listen("tcp", ":80")
+			if err != nil {
+				log.Fatal(err)
+			}
+			name, ok := tailscale.ExpandSNIName(context.Background(), *hostname)
+			if !ok {
+				log.Fatalf("can't get hostname for https redirect")
+			}
+			if err := http.Serve(l80, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				http.Redirect(w, r, fmt.Sprintf("https://%s", name), http.StatusMovedPermanently)
+			})); err != nil {
+				log.Fatal(err)
+			}
+		}()
+	} else {
+		ln, err = ts.Listen("tcp", ":80")
+	}
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Printf("proxy-to-grafana running at %v, proxying to %v", ln.Addr(), *backendAddr)
+	log.Fatal(http.Serve(ln, proxy))
+}
+
+func modifyRequest(req *http.Request) {
+	// with enable_login_token set to true, we get a cookie that handles
+	// auth for paths that are not /login
+	if req.URL.Path != "/login" {
+		return
+	}
+
+	user, err := getTailscaleUser(req.Context(), req.RemoteAddr)
+	if err != nil {
+		log.Printf("error getting Tailscale user: %v", err)
+		return
+	}
+
+	req.Header.Set("X-Webauth-User", user.LoginName)
+	req.Header.Set("X-Webauth-Name", user.DisplayName)
+}
+
+func getTailscaleUser(ctx context.Context, ipPort string) (*tailcfg.UserProfile, error) {
+	whois, err := tailscale.WhoIs(ctx, ipPort)
+	if err != nil {
+		return nil, fmt.Errorf("failed to identify remote host: %w", err)
+	}
+	if len(whois.Node.Tags) != 0 {
+		return nil, fmt.Errorf("tagged nodes are not users")
+	}
+	if whois.UserProfile == nil || whois.UserProfile.LoginName == "" {
+		return nil, fmt.Errorf("failed to identify remote user")
+	}
+
+	return whois.UserProfile, nil
+}


cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,121 @@
+// proxy-to-grafana uses a reverse proxy which identifies users based on their originating tailscale node and creates / logs in users to grafana. Uses the auth proxy feature for grafana: https://grafana.com/docs/grafana/latest/auth/auth-proxy/
```
bradfitz, line 1: Add copyright header in oss

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,121 @@
...
+//
+// Provide the env var TS_AUTHKEY to have this server automatically join your Tailnet.
+//
+// Use this grafana configuration to enable the auth proxy:
+
```
bradfitz, line 6: Need // here otherwise the packagedoc is only the comment adjacent to the package line

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,121 @@
...
+	"strings"
+
+	"inet.af/netaddr"
+	"tailscale.com/client/tailscale"
+	"tailscale.com/net/tsaddr"
+	"tailscale.com/tailcfg"
+	"tailscale.com/tsnet"
+)
+
+var (
```
bradfitz, line 36: Clarify whether this is full or base name

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,121 @@
...
+	hostname     = flag.String("hostname", "", "Tailscale hostname to serve on")
+	upstreamAddr = flag.String("upstream-addr", "", "Address of the upstream Grafana server, in host:port format")
```
bradfitz, line 38: Backend is more common than upstream I think

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,121 @@
...
+	tailscaleDir = flag.String("state-dir", "", "Directory in which to store the Tailscale auth state")
```
bradfitz, line 39: Optional?

maisem: You'd want it for certs

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,121 @@
...
+		log.Printf("error getting Tailscale user: %v", err)
+		return
+	}
+	// try to make these emails not collide with real users if any
+	email := strings.Replace(user.LoginName, "@", "-auto@", 1)
+	req.Header.Set("X-WEBAUTH-USER", email)
+	req.Header.Set("X-WEBAUTH-NAME", user.DisplayName)
+}
+
+func getTailscaleUser(ctx context.Context, ip string) (*tailcfg.UserProfile, error) {
```
maisem, line 111: You probably should also check the node isn't tagged.

maisem (commented):
I can send a PR to add a 80->443 redirect handler in tsweb

cmd/proxy-to-grafana/proxy-to-grafana.go
bradfitz: But it should be automatic. I'd like this to be described like "alternate directory to use for Tailscale state storage. If empty, a default is used."

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,121 @@
...
+		return
+	}
+	user, err := getTailscaleUser(req.Context(), req.RemoteAddr)
+	if err != nil {
+		log.Printf("error getting Tailscale user: %v", err)
+		return
+	}
+	// try to make these emails not collide with real users if any
+	email := strings.Replace(user.LoginName, "@", "-auto@", 1)
+	req.Header.Set("X-WEBAUTH-USER", email)
```
bradfitz, line 107: X-Webauth-User
X-Webauth-Name

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,121 @@
...
+	req.Header.Set("X-WEBAUTH-NAME", user.DisplayName)
+}
+
+func getTailscaleUser(ctx context.Context, ip string) (*tailcfg.UserProfile, error) {
+	whois, err := tailscale.WhoIs(ctx, ip)
+	if err != nil {
+		return nil, fmt.Errorf("failed to identify remote host: %w", err)
+	}
+	if whois.UserProfile == nil || whois.UserProfile.LoginName == "" {
+		return nil, fmt.Errorf("failed to identify remote user: %w", err)
```
bradfitz, line 117: err is nil here

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,121 @@
...
+	hostname     = flag.String("hostname", "", "Tailscale hostname to serve on")
+	upstreamAddr = flag.String("upstream-addr", "", "Address of the upstream Grafana server, in host:port format")
+	tailscaleDir = flag.String("state-dir", "", "Directory in which to store the Tailscale auth state")
+)
+
+func main() {
+	flag.Parse()
+	if *hostname == "" {
+		log.Fatal("missing --hostname")
+	}
```
bradfitz, line 46: Also fail if it contains a period?

cmd/proxy-to-grafana/proxy-to-grafana.go
bradfitz: capital Tailscale and Grafana

and wrap this to 76 or 80-ish.

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// proxy-to-grafana uses a reverse proxy which identifies users based on their
```
bradfitz, line 5: is a reverse proxy

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,129 @@
...
+// originating Tailscale node and creates / logs in users to grafana. Uses the
```
bradfitz, line 6: Grafana capitalized

bradfitz: It uses

It uses Grafana's AuthProxy

("Grafana AuthProxy" is how they seem to mostly spell it in their docs)

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	proxy := httputil.NewSingleHostReverseProxy(url)
+	originalDirector := proxy.Director
+	proxy.Director = func(req *http.Request) {
+		originalDirector(req)
+		modifyRequest(req)
+	}
+
+	ltsn, err := ts.Listen("tcp", ":80")
```
bradfitz, line 75: common variable name for listeners is `ln`

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Printf("serving access to %s on port 80", *backendAddr)
```
bradfitz, line 79: This is the first/only log in the program, so it should identify itself.

"proxy-to-grafana running at %v, proxying to %v ..."

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,129 @@
...
+	log.Fatal(http.Serve(ltsn, ProxyRequestHandler(proxy)))
+}
+
+func ProxyRequestHandler(proxy *httputil.ReverseProxy) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		proxy.ServeHTTP(w, r)
+	})
+}
```
bradfitz, line 87: So this actually doesn't do anything. You can just write `proxy` on line 80 instead of `ProxyRequestHandler(proxy)`.

nickoneill: Done

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: 👍

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: 👍

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: 👍

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: 👍

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,121 @@
...
+		originalDirector(req)
+		modifyRequest(req)
+	}
+
+	ltsn, err := ts.Listen("tcp", ":80")
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Printf("serving access to %s on port 80", *upstreamAddr)
+	log.Fatal(http.Serve(ltsn, http.HandlerFunc(ProxyRequestHandler(proxy))))
```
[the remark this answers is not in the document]
nickoneill: 👍

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: 👍

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: Change made - but why is this more efficient?

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: 👍

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: 👍

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: 👍

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	var ln net.Listener
+	if *useHTTPS {
+		ln, err = ts.Listen("tcp", ":443")
+		ln = tls.NewListener(ln, &tls.Config{
+			GetCertificate: tailscale.GetCertificate,
+		})
+	} else {
+		ln, err = ts.Listen("tcp", ":80")
```
bradfitz, line 85: even in https mode you should also have 80 redirect to 443, though?

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,134 @@
...
+
+	ipp, err := netaddr.ParseIPPort(req.RemoteAddr)
+	if err != nil {
+		log.Printf("bad remote address")
+		return
+	}
+	if !tsaddr.IsTailscaleIP(ipp.IP()) {
+		log.Printf("not a tailscale IP")
+		return
+	}
```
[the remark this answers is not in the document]
nickoneill: I was thinking this would prevent access from non-tailscale interfaces, but since it uses tsnet now it only has the one interface, will remove.

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,134 @@
...
+	user, err := getTailscaleUser(req.Context(), req.RemoteAddr)
+	if err != nil {
+		log.Printf("error getting Tailscale user: %v", err)
+		return
+	}
+	// try to make these emails not collide with real users if any
+	email := strings.Replace(user.LoginName, "@", "-auto@", 1)
```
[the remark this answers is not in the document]
nickoneill: I kept it from the original POC, though existing users with the same email as ts identities don't break the implementation. I don't think it's important enough to flag though, if it doesn't seem valuable then let's just get rid of it.

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: Ah yeah

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// proxy-to-grafana is a reverse proxy which identifies users based on their
+// originating Tailscale node and creates / logs in users to Grafana. It uses
+// Grafana's AuthProxy feature:
+// https://grafana.com/docs/grafana/latest/auth/auth-proxy/
+//
+// Provide the env var TS_AUTHKEY to have this server automatically join your
```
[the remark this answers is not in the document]
nickoneill: 👍

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// proxy-to-grafana is a reverse proxy which identifies users based on their
+// originating Tailscale node and creates / logs in users to Grafana. It uses
```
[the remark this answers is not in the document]
nickoneill: 👍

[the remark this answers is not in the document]
nickoneill: 👍

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,135 @@
...
+	"net/url"
+	"strings"
+
+	"tailscale.com/client/tailscale"
+	"tailscale.com/tailcfg"
+	"tailscale.com/tsnet"
+)
+
+var (
+	hostname     = flag.String("hostname", "", "Tailscale hostname to serve on, used as the full name for MagicDNS or subdomain in your domain alias for HTTPS")
```
maisem, line 47: Full name? Like do I have to do foo.example.com?

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,135 @@
...
+	backendAddr  = flag.String("backend-addr", "", "Address of the Grafana server, in host:port format")
+	tailscaleDir = flag.String("state-dir", "./", "Alternate directory to use for Tailscale state storage. If empty, a default is used.")
+	useHTTPS     = flag.Bool("use-https", false, "Serve over SSL via your *.ts.net subdomain if enabled in Tailscale admin.")
```
maisem, line 50: TLS

maisem: Or HTTPS

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,135 @@
...
+		ln, err = ts.Listen("tcp", ":443")
+		ln = tls.NewListener(ln, &tls.Config{
+			GetCertificate: tailscale.GetCertificate,
+		})
+		go func() {
+			l80, err := ts.Listen("tcp", ":80")
+			if err != nil {
+				log.Fatal(err)
+			}
+			if err := http.Serve(l80, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
maisem, line 89: You can use tsweb.Port80Handler

cmd/proxy-to-grafana/proxy-to-grafana.go
maisem: I think this is supposed to be base name

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,135 @@
...
+				http.Redirect(w, r, fmt.Sprintf("https://%s", r.Host), http.StatusMovedPermanently)
```
maisem, line 90: You probably want to use tailscale.ExpandSNIName to get to the full domain name here

nickoneill: Thanks! I was wondering where I might get this

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: <we paired and this was not immediately useful, may be though once we enable the debug endpoint in the future>

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: 👍

maisem (approved):

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,154 @@
...
+	"time"
+
+	"tailscale.com/client/tailscale"
+	"tailscale.com/tailcfg"
+	"tailscale.com/tsnet"
+)
+
+var (
+	hostname     = flag.String("hostname", "", "Tailscale hostname to serve on, used as the base name for MagicDNS or subdomain in your domain alias for HTTPS")
+	backendAddr  = flag.String("backend-addr", "", "Address of the Grafana server, in host:port format")
```
bradfitz, line 49: Clarify that this is always HTTP and not HTTPS? And that it should usually be localhost:nnnn?

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,154 @@
...
+		log.Fatal("missing --backend-addr")
+	}
+	ts := &tsnet.Server{
+		Dir:      *tailscaleDir,
+		Hostname: *hostname,
+	}
+
+	url, err := url.Parse(fmt.Sprintf("http://%s", *backendAddr))
+	if err != nil {
+		panic(err)
```
bradfitz, line 69: log.Fatalf with a nicer error message. No need for the user to get a stack trace. This ain't Python :)

bradfitz: (That is: panics are okay for internal programmer errors, but not for errors from user input)

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil {
+		log.Printf("error getting Tailscale user: %v", err)
+		return
+	}
+
+	req.Header.Set("X-Webauth-User", user.LoginName)
+	req.Header.Set("X-Webauth-Name", user.DisplayName)
+}
+
+func getTailscaleUser(ctx context.Context, ip string) (*tailcfg.UserProfile, error) {
```
bradfitz, line 141: `ipPort string` probably. WhoIs takes both, but `req.RemoteAddr` is ip:port.

bradfitz (approved):

bradfitz:
(Good to submit without further review after minor edits)

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: Done

cmd/proxy-to-grafana/proxy-to-grafana.go
nickoneill: Done

nickoneill:
🎉 merging when checks finish

PR #4209, opened 2022-03-18
prober: library to build healthchecking probers.

danderson:

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -1,0 +1,62 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package prober
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+const maxHTTPBody = 4 << 20 // MiB
+
+// HTTP returns a Probe that healthchecks an HTTP URL.
+//
+// The Probe sends a GET request for url, expects an HTTP 200
+// response, and verifies that want is present in the response
+// body. If the URL is HTTPS, the probe further checks that the TLS
+// certificate is good for at least the next 7 days.
+func HTTP(url, wantText string) Probe {
+	return func(ctx context.Context) error {
+		return probeHTTP(ctx, url, []byte(wantText))
+	}
+}
+
+func probeHTTP(ctx context.Context, url string, want []byte) error {
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
+	if err != nil {
+		return fmt.Errorf("constructing request: %w", err)
+	}
+
+	// Get a completely new transport each time, so we don't reuse a
+	// past connection.
+	tr := http.DefaultTransport.(*http.Transport).Clone()
+	defer tr.CloseIdleConnections()
+	c := &http.Client{
+		Transport: tr,
+	}
+
+	resp, err := c.Do(req)
+	if err != nil {
+		return fmt.Errorf("fetching %q: %w", url, err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("fetching %q: status code %d, want 200", url, resp.StatusCode)
+	}
+
+	bs, err := io.ReadAll(&io.LimitedReader{resp.Body, maxHTTPBody})
+	if err != nil {
+		return fmt.Errorf("reading body of %q: %w", url, err)
+	}
+
+	if !bytes.Contains(bs, want) {
+		return fmt.Errorf("body of %q does not contain %q", url, want)
+	}
+
+	return nil
+}
--- a/prober/prober.go
+++ b/prober/prober.go
@@ -1,0 +1,235 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package prober implements a simple blackbox prober. Each probe runs
+// in its own goroutine, and run results are recorded as Prometheus
+// metrics.
+package prober
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"sync"
+	"time"
+
+	"tailscale.com/metrics"
+)
+
+// Probe is a function that probes something and reports whether the
+// probe succeeded. The provided context must be used to ensure timely
+// cancellation and timeout behavior.
+type Probe func(context.Context) error
+
+// a Prober manages a set of probes and keeps track of their results.
+type Prober struct {
+	// Time-related functions that get faked out during tests.
+	now       func() time.Time
+	newTicker func(time.Duration) ticker
+
+	// lastStart is the time, in seconds since epoch, of the last time
+	// each probe started a probe cycle.
+	lastStart metrics.LabelMap
+	// lastEnd is the time, in seconds since epoch, of the last time
+	// each probe finished a probe cycle.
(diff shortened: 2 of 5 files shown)


prober/http.go
```diff
@@ -0,0 +1,62 @@
...
+
+const maxHTTPBody = 4 << 20 // MiB
+
+// HTTP returns a Probe that healthchecks an HTTP URL.
+//
+// The Probe sends a GET request for url, expects an HTTP 200
+// response, and verifies that want is present in the response
+// body. If the URL is HTTPS, the probe further checks that the TLS
+// certificate is good for at least the next 7 days.
+func HTTP(url, wantText string) Probe {
```
maisem, line 23: how about taking in a `func() *http.Request`? and add a helper for GET? would allow adding POSTs.


I don't know what we are probing, maybe not required.

prober/http.go
```diff
@@ -0,0 +1,62 @@
...
+		return fmt.Errorf("constructing request: %w", err)
+	}
+
+	// Get a completely new transport each time, so we don't reuse a
+	// past connection.
+	tr := http.DefaultTransport.(*http.Transport).Clone()
+	defer tr.CloseIdleConnections()
+	c := &http.Client{
+		Transport: tr,
+	}
```
maisem, line 41: For my edification, is there a nicer way to do this? I guess this works but it seems like a big hammer, which maybe a good thing here? Should we also be testing long lived connections?

prober/http.go
```diff
@@ -0,0 +1,62 @@
...
+
+	resp, err := c.Do(req)
+	if err != nil {
+		return fmt.Errorf("fetching %q: %w", url, err)
```
maisem, line 45: maybe known error types so that we can have varz?

prober/prober.go
```diff
@@ -0,0 +1,166 @@
...
+	return func() {
+		p.mu.Lock()
+		stopped := p.probes[name]
+		p.mu.Unlock()
+		cancel()
+		<-stopped
+	}
+}
+
+func (p *Prober) probeLoop(ctx context.Context, name string, interval time.Duration, fun Probe) {
```
maisem, line 91: add a comment that the first probe is after interval.

prober/prober.go
```diff
@@ -0,0 +1,166 @@
...
+	start := p.start(name)
+	defer func() {
+		if r := recover(); r != nil {
+			log.Printf("probe %s panicked: %v", name, r)
+			p.end(name, start, errors.New("panic"))
+		}
+	}()
+	timeout := time.Duration(float64(interval) * 0.8)
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
```
maisem, line 118: add a function comment about the 80% behavior

prober/prober.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+}
+
+func (p *Prober) runProbe(ctx context.Context, name string, interval time.Duration, fun Probe) {
+	start := p.start(name)
+	defer func() {
+		if r := recover(); r != nil {
+			log.Printf("probe %s panicked: %v", name, r)
+			p.end(name, start, errors.New("panic"))
+		}
```
maisem, line 114: why allow probes to panic? document pls

prober/tls.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	if err := conn.Handshake(); err != nil {
+		return fmt.Errorf("TLS handshake error with %q: %w", hostname, err)
+	}
+	if err := conn.VerifyHostname(hostname); err != nil {
+		return fmt.Errorf("Host %q TLS verification failed: %w", hostname, err)
+	}
+
+	latestAllowedExpiration := time.Now().Add(7 * 24 * time.Hour) // 7 days from now
+	if expires := conn.ConnectionState().PeerCertificates[0].NotAfter; expires.After(latestAllowedExpiration) {
```
maisem, line 40: i think you want the opposite.

prober/prober.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// New returns a new Prober.
+func New() *Prober {
+	return &Prober{
+		lastStart:     metrics.LabelMap{Label: "probe"},
+		lastEnd:       metrics.LabelMap{Label: "probe"},
+		lastResult:    metrics.LabelMap{Label: "probe"},
+		lastLatency:   metrics.LabelMap{Label: "probe"},
+		alertDuration: metrics.LabelMap{Label: "probe"},
```
maisem, line 46: what's the alert duration

prober/prober.go
```diff
@@ -0,0 +1,166 @@
...
+type Prober struct {
+	// Maps are keyed by probe name
+	lastStart     metrics.LabelMap
+	lastEnd       metrics.LabelMap
+	lastResult    metrics.LabelMap
+	lastLatency   metrics.LabelMap
+	alertDuration metrics.LabelMap
+
+	mu     sync.Mutex // protects all following fields
+	probes map[string]chan struct{}
```
maisem, line 36: activeProbesChans?

prober/http.go
danderson: Hmm, possibly, but usage of these functions is already quite nested:

```
p.Run("my-probe", 10*time.Second, prober.HTTP("https://example.com", "Sign in"))
```

I'm worried about readability with a third level of nesting.

For right now, I'd like feature parity with our existing probers, which only do HTTP GET, TCP connect, and TLS validation. I'd say let's revisit when we need fancier HTTP stuff?

prober/http.go
danderson: AFAICT, there's no cleaner way to be certain that a client will create a completely fresh TCP connection on every probe. I'd like to be wrong though :)

Testing long-lived connections I think is going to depend a lot on the behavior of the endpoint under test, so I'm thinking for those we start with custom\ Prober.Run funcs, and see if a pattern emerges.

prober/http.go
danderson: Hmm. This is a bit terrible to export in a form Prometheus will understand, so to send richer alerts we'd have to speak the alertmanager protocol directly, and make the prober fully stateful. I'd prefer to avoid that for now, but if we don't have that, this prober needs an HTTP UI where you can ask it about recent history. I'll think about how to do this.

prober/prober.go
danderson: Done.

prober/prober.go
danderson: Done.

prober/http.go
maisem: I just meant like export counters for the number of times we've seen a type of known error. Nothing too fancy

danderson: Hmm, maybe, but being a periodic prober, it's kinda weird for it to keep track of errors seen over its lifetime, unconnected from the probe cycles? Maybe? I'm not sure.

I think I'd like to refactor the probing API a little bit if we want this, to let probes set their own vars. Okay if I do in a followup?

maisem: yeah, can add later as needed as well. This is fine for now!

prober/prober_test.go
```diff
@@ -0,0 +1,293 @@
...
+	"sync"
+	"testing"
+	"time"
+
+	"tailscale.com/syncs"
+	"tailscale.com/tstest"
+)
+
+const (
+	probeInterval        = 10 * time.Second // So expvars that are integer numbers of seconds change
```
maisem, line 22: 8 seconds? 10 won't divide nicely by 4 :)

maisem (approved):

prober/prober_test.go
danderson: It's all nanoseconds under the hood, so it should be fine, but sure :)

Issue #4051, opened 2022-03-02
ssh/tailssh: TestSSH sometimes flaking on CI


bradfitz:
On our GitHub CI builders, TestSSH is sometimes failing with:

```
2022-03-02T03:32:17.8513614Z --- FAIL: TestSSH (0.19s)
2022-03-02T03:32:17.8514422Z     --- FAIL: TestSSH/env (0.03s)
2022-03-02T03:32:17.8517168Z         tailssh_test.go:256: USER = ""; want "runner"
2022-03-02T03:32:17.8517663Z         tailssh_test.go:259: HOME = ""; want "/home/runner"
2022-03-02T03:32:17.8518300Z         tailssh_test.go:262: PWD = ""; want "/home/runner"
2022-03-02T03:32:17.8518706Z         tailssh_test.go:265: no SHELL
2022-03-02T03:32:17.8519089Z         tailssh_test.go:268: LANG = ""; want "foo"
2022-03-02T03:32:17.8519464Z         tailssh_test.go:270: got: map[]
2022-03-02T03:32:17.8519689Z FAIL
2022-03-02T03:32:17.8525053Z exit status 1
2022-03-02T03:32:17.8525709Z FAIL	tailscale.com/ssh/tailssh	0.192s
```

/cc @maisem

maisem:
The problem lies in the way we are spinning up new processes. We inherit a lot of environmental properties from the parent in this case tailscaled. The login session PR fixes this by redoing the way we set the child processes environment in general.

bradfitz:
Phew!

Out of curiosity (I'm happy enough if it's fixed one way or another), happen to know which environment we're inheriting that's screwing things up only sometimes?

josharian:
This doesn't reproduce for me locally in a loop, even running for a long time. But if I run multiple of these tests concurrently it starts failing immediately.

bradfitz:
Just hit both this and #5209 in the same run, at 036334e9135037d99679f0a9811c7cd93cbf152d

```
ok  	tailscale.com/smallzstd	0.057s [no tests to run]
==================
WARNING: DATA RACE
Read at 0x00c0000e0060 by goroutine 231:
  github.com/tailscale/golang-x-crypto/ssh.(*channel).WriteExtended()
[...]

PR #4216, opened 2022-03-18
ssh/tailssh_test: skip TestSSH/stdin in CI

maisem:
Updates #4051


maisem:
It's very flaky

crawshaw (approved):

Issue #4217, opened 2022-03-18
Application-defined capabilities


bradfitz:
Currently, the Tailscale ACL system lets you grant access to IP protocols (TCP, UDP, etc) and ports (80, 443, 22, etc).

Those can be viewed as as a capability grant: the admin grants the capability for a node X to do "tcp/443" on node Y.

But that's only a network-level capability. We can say that you have TCP/3306 access, but there's no support for saying, "You have read-only access to the database" vs "You have read/write access to the database."

In Tailscale 1.6 or so started adding support for the "WhoIs" handler and the localapi for application to ask, "Who is the remote node connecting to me?"

That lets you write Tailscale authentication proxies that speak Tailscale in the front (Tailscale crypto, Tailscale authentication) and then map it to another system on the backend, filling in the authentication as desired. But if that backend system requires some lazy user provisioning or access control (authz, not authn), there's no way from the Tailscale ACL config to grant certain users/groups access to higher-level application capabilities ("mysql.com/rw", "mysql.com/ro"), etc.

@maisem and I propose extending the Tailscale ACL syntax to permit arbitrary capabilities where we currently accept ports.

That is "80", "udp/443", and "tcp/8000-8999" are capabilities today. But they're capabilities for the network.

We propose that you be able to write:

```
{
    "src": ["group:all"],
    "dst": ["mon:tcp/80,443"],
},
{
    "src": ["group:eng"],
    "dst": ["mon:cap/grafana.com/rw"],
}
```

A capability is an arbitrary string (`"grafana.com/rw"`) that is granted between a pair of nodes, identified by IPs.  (or in this case, `"mon"` being a Host alias to a Tailscale IP)

Then, the `WhoIs` lookup response is extended to include the capabilities that the remote user has:

```patch
diff --git a/client/tailscale/apitype/apitype.go b/client/tailscale/apitype/apitype.go
index cae573a4..c5d10dff 100644
--- a/client/tailscale/apitype/apitype.go
+++ b/client/tailscale/apitype/apitype.go
@@ -11,6 +11,9 @@
 type WhoIsResponse struct {
        Node        *tailcfg.Node
        UserProfile *tailcfg.UserProfile
+
+       // Caps are extra capabilities that the remote Node has to this node.
+       Caps []string `json:",omitempty"`
 }
```

Implementation linked below, which might answer further questions.

/cc @maisem @nickoneill @crawshaw @mayakacz @Xe

danderson:
[...]

PR #4218, opened 2022-03-18
all: add arbitrary capability support

bradfitz:
Updates #4217

### DIFF
--- a/wgengine/filter/filter.go
+++ b/wgengine/filter/filter.go
@@ -27,10 +27,12 @@ type Filter struct {
 	// destination within local, regardless of the policy filter
 	// below.
 	local *netaddr.IPSet
+
 	// logIPs is the set of IPs that are allowed to appear in flow
 	// logs. If a packet is to or from an IP not in logIPs, it will
 	// never be logged.
 	logIPs *netaddr.IPSet
+
 	// matches4 and matches6 are lists of match->action rules
 	// applied to all packets arriving over tailscale
 	// tunnels. Matches are checked in order, and processing stops
@@ -38,6 +40,11 @@ type Filter struct {
 	// match is to drop the packet.
 	matches4 matches
 	matches6 matches
+
+	// cap4 and cap6 are the subsets of the matches that are about
+	// capability grants, partitioned by source IP address family.
+	cap4, cap6 matches
+
 	// state is the connection tracking state attached to this
 	// filter. It is used to allow incoming traffic that is a response
 	// to an outbound connection that this node made, even if those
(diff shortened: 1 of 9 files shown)


wgengine/filter/filter.go
```diff
@@ -205,6 +214,27 @@ func matchesFamily(ms matches, keep func(netaddr.IP) bool) matches {
 	return ret
 }
 
+// capMatchesFamily returns the subset of ms for which keep(srcNet.IP)
+// and the match is a capability grant.
+func capMatchesFamily(ms matches, keep func(netaddr.IP) bool) matches {
```
maisem, line 219: `capMatchFamily` implies that it is an inherent understanding of IP families. Maybe rename this and `matchesFamily` to `matchesFunc`. Or leave it as is ¯\\\_(ツ)\_/¯

wgengine/filter/filter.go
```diff
@@ -291,6 +321,30 @@ func (f *Filter) CheckTCP(srcIP, dstIP netaddr.IP, dstPort uint16) Response {
 	return f.RunIn(pkt, 0)
 }
 
+// AppendCaps appends to base the capabilities that srcIP has talking
+// to dstIP.
+func (f *Filter) AppendCaps(base []string, srcIP, dstIP netaddr.IP) []string {
```
maisem, line 326: optional: this is doing too much, maybe split out `Caps`?

wgengine/filter/filter.go
maisem: why not match on `cap.Net` too, to be consistent with `matchesFamily`?

wgengine/filter/match.go
```diff
@@ -47,12 +47,18 @@ func (npr NetPortRange) String() string {
 	return fmt.Sprintf("%v:%v", npr.Net, npr.Ports)
 }
 
+type CapMatch struct {
+	Net netaddr.IPPrefix
+	Cap string
+}
```
maisem, line 53: docs

wgengine/filter/match.go
```diff
@@ -47,12 +47,18 @@ func (npr NetPortRange) String() string {
 	return fmt.Sprintf("%v:%v", npr.Net, npr.Ports)
 }
 
+type CapMatch struct {
+	Net netaddr.IPPrefix
```
maisem, line 51: rename Net to Dst?

wgengine/filter/match.go
```diff
@@ -47,12 +47,18 @@ func (npr NetPortRange) String() string {
 	return fmt.Sprintf("%v:%v", npr.Net, npr.Ports)
 }
 
+type CapMatch struct {
```
maisem, line 50: This name reuse really threw me off in the code review and I had to open it in VSCode to actually figure out what was going on.

wgengine/filter/match.go
bradfitz: done

wgengine/filter/match.go
bradfitz: done

maisem (approved):

PR #4222, opened 2022-03-19
go.mod: bump netstack

raggi:
Primarily this is for google/gvisor@f375784d83852b1e3ff20cc9de0648b3c0cf8525 and the
related commits that provide buffer pooling for the endpoint code paths
we use.


bradfitz (approved):
You can `go mod tidy -compat=1.17` too.

PR #4230, opened 2022-03-19
cmd/tailscale: Add file get --loop flag.

davideger:
To "automatically receive taildrop files to my Downloads directory,"
a user currently has to run 'tailscale file get' in a loop.
Add a "looping" / daemon option to the file get command so no shell
scripts will be required for this behavior.   Addresses #2312.

Signed-off-by: David Eger <david.eger@gmail.com>

### DIFF
--- a/cmd/tailscale/cli/file.go
+++ b/cmd/tailscale/cli/file.go
@@ -318,6 +318,7 @@ var fileGetCmd = &ffcli.Command{
 	FlagSet: (func() *flag.FlagSet {
 		fs := newFlagSet("get")
 		fs.BoolVar(&getArgs.wait, "wait", false, "wait for a file to arrive if inbox is empty")
+		fs.BoolVar(&getArgs.loop, "loop", false, "run get in a loop, receiving files as they come in")
 		fs.BoolVar(&getArgs.verbose, "verbose", false, "verbose output")
 		fs.Var(&getArgs.conflict, "conflict", `behavior when a conflicting (same-named) file already exists in the target directory.
 	skip:       skip conflicting files: leave them in the taildrop inbox and print an error. get any non-conflicting files
@@ -329,6 +330,7 @@ var fileGetCmd = &ffcli.Command{
 
 var getArgs = struct {
 	wait     bool
+	loop     bool
 	verbose  bool
 	conflict onConflict
 }{conflict: skipOnExist}
@@ -358,7 +360,7 @@ func openFileOrSubstitute(dir, base string, action onConflict) (*os.File, error)
 		}
 		return nil, fmt.Errorf("failed to write; %w", err)
 	case overwriteExisting:
-		// remove the target file and create it anew so we don't fall for an 
+		// remove the target file and create it anew so we don't fall for an
 		// attacker who symlinks a known target name to a file he wants changed.
 		if err = os.Remove(targetFile); err != nil {
 			return nil, fmt.Errorf("unable to remove target file: %w", err)
@@ -387,16 +389,69 @@ func receiveFile(ctx context.Context, wf apitype.WaitingFile, dir string) (targe
 	if err != nil {
 		return "", 0, fmt.Errorf("opening inbox file %q: %w", wf.Name, err)
 	}
+	defer rc.Close()
 	f, err := openFileOrSubstitute(dir, wf.Name, getArgs.conflict)
 	if err != nil {
 		return "", 0, err
 	}
 	_, err = io.Copy(f, rc)
-	rc.Close()
 	if err != nil {
+		f.Close()
 		return "", 0, fmt.Errorf("failed to write %v: %v", f.Name(), err)
 	}
 	return f.Name(), size, f.Close()
+}
+
+func runFileGetOneBatch(ctx context.Context, dir string) []error {
+	var wfs []apitype.WaitingFile
+	var err error
+	var errs []error
+	for len(errs) == 0 {
+		wfs, err = tailscale.WaitingFiles(ctx)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("getting WaitingFiles: %w", err))
+			break
+		}
+		if len(wfs) != 0 || !(getArgs.wait || getArgs.loop) {
+			break
+		}
+		if getArgs.verbose {
+			printf("waiting for file...")
+		}
+		if err := waitForFile(ctx); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	deleted := 0
+	for i, wf := range wfs {
+		if len(errs) > 100 {
+			// Likely, everything is broken.
+			// Don't try to receive any more files in this batch.
+			errs = append(errs, fmt.Errorf("too many errors in runFileGetOneBatch(). %d files unexamined", len(wfs) - i))
+			break
+		}
+		writtenFile, size, err := receiveFile(ctx, wf, dir)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if getArgs.verbose {
+			printf("wrote %v as %v (%d bytes)\n", wf.Name, writtenFile, size)
+		}
+		if err = tailscale.DeleteWaitingFile(ctx, wf.Name); err != nil {
+			errs = append(errs, fmt.Errorf("deleting %q from inbox: %v", wf.Name, err))
+			continue
+		}
+		deleted++
+	}
+	if deleted == 0 && len(wfs) > 0 {
+		// persistently stuck files are basically an error
+		errs = append(errs, fmt.Errorf("moved %d/%d files", deleted, len(wfs)))
+	} else if getArgs.verbose {
+		printf("moved %d/%d files\n", deleted, len(wfs))
+	}
+	return errs
 }
 
 func runFileGet(ctx context.Context, args []string) error {
@@ -413,45 +468,28 @@ func runFileGet(ctx context.Context, args []string) error {
 	if fi, err := os.Stat(dir); err != nil || !fi.IsDir() {
 		return fmt.Errorf("%q is not a directory", dir)
 	}
-
-	var wfs []apitype.WaitingFile
-	var err error
-	for {
-		wfs, err = tailscale.WaitingFiles(ctx)
-		if err != nil {
-			return fmt.Errorf("getting WaitingFiles: %w", err)
-		}
-		if len(wfs) != 0 || !getArgs.wait {
-			break
-		}
-		if getArgs.verbose {
-			printf("waiting for file...")
-		}
-		if err := waitForFile(ctx); err != nil {
-			return err
+	if getArgs.loop {
+		for {
+			errs := runFileGetOneBatch(ctx, dir)
+			for _, err := range errs {
+				outln(err)
+			}
+			if len(errs) > 0 {
+				// It's possible whatever caused the error(s) (e.g. conflicting target file,
+				// full disk, unwritable target directory) will re-occur if we try again so
+				// let's back off and not busy loop on error.
+				//
+				// If we've been invoked as:
+				//    tailscale file get --conflict=skip ~/Downloads
+				// then any file coming in named the same as one in ~/Downloads will always
+				// appear as an "error" until the user clears it, but other incoming files
+				// should be receivable when they arrive, so let's not wait too long to
+				// check again.
+				time.Sleep(5 * time.Second)
+			}
 		}
 	}
-
-	var errs []error
-	deleted := 0
-	for _, wf := range wfs {
-		writtenFile, size, err := receiveFile(ctx, wf, dir)
-		if err != nil {
-			errs = append(errs, err)
-			continue
-		}
-		if getArgs.verbose {
-			printf("wrote %v as %v (%d bytes)\n", wf.Name, writtenFile, size)
-		}
-		if err = tailscale.DeleteWaitingFile(ctx, wf.Name); err != nil {
-			errs = append(errs, fmt.Errorf("deleting %q from inbox: %v", wf.Name, err))
-			continue
-		}
-		deleted++
-	}
-	if getArgs.verbose {
-		printf("moved %d/%d files\n", deleted, len(wfs))
-	}
+	errs := runFileGetOneBatch(ctx, dir)
 	if len(errs) == 0 {
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


bradfitz:
how about --loop instead so we don't imply this daemonizes?

davideger:
sounds good, changed to --loop.

cmd/tailscale/cli/file.go
```diff
@@ -489,9 +509,10 @@ func waitForFile(ctx context.Context) error {
 	c, bc, pumpCtx, cancel := connect(ctx)
 	defer cancel()
 	fileWaiting := make(chan bool, 1)
+	notifyError := make(chan error)
```
bradfitz, line 512: make this 1-buffered like above.

cmd/tailscale/cli/file.go
```diff
@@ -413,52 +415,70 @@ func runFileGet(ctx context.Context, args []string) error {
...
 		return fmt.Errorf("%q is not a directory", dir)
 	}
-
-	var wfs []apitype.WaitingFile
-	var err error
-	for {
-		wfs, err = tailscale.WaitingFiles(ctx)
-		if err != nil {
-			return fmt.Errorf("getting WaitingFiles: %w", err)
+	for true {  // Infinite loop
```
bradfitz, line 418: no need for "true" or the comment. You can assume readers know the basics of the language, even if you're just learning it :)

cmd/tailscale/cli/file.go
```diff
@@ -413,52 +415,70 @@ func runFileGet(ctx context.Context, args []string) error {
...
-		printf("moved %d/%d files\n", deleted, len(wfs))
-	}
-	if len(errs) == 0 {
-		return nil
-	}
-	for _, err := range errs[:len(errs)-1] {
-		outln(err)
 	}
-	return errs[len(errs)-1]
+	// unreachable
```
bradfitz, line 480: then you don't need this comment or the return line.

See https://go.dev/ref/spec#Terminating_statements

cmd/tailscale/cli/file.go
bradfitz, line 418: but I think we'll want to refactor this both to unindent & simplify it, pulling the body of the loop into its own func

cmd/tailscale/cli/file.go
```diff
@@ -413,52 +415,70 @@ func runFileGet(ctx context.Context, args []string) error {
...
-			errs = append(errs, fmt.Errorf("deleting %q from inbox: %v", wf.Name, err))
-			continue
+		if len(errs) > 0 {
+			// It's possible whatever caused the error(s) (e.g. conflicting target file,
+			// full disk, unwritable target directory) will re-occur if we try again so
+			// let's back off and not busy loop on error.
+			//
+			// However, let's not wait too long so other files may get received if
+			// there's just one problematic file.
+			time.Sleep(5 * time.Second)
```
bradfitz, line 477: we don't always want this sleep. only in the loop case.

but I think you actually want to use `tailscale.com/logtail/backoff.BackOff`

cmd/tailscale/cli/file.go
davideger: ooooh, go's spec has a list of recognized idioms.  unreachable is... sorta turing complete, so i figured it was a "compiler not smart enough to figure it out" thing.  thanks for the pointer.

cmd/tailscale/cli/file.go
davideger: the sleep only got executed in the loopy case on error, but the structure was making that really non-obvious.  (the "non-loop" case had returns in all of its branches).

So the failure scenario that had me shove a 5 second sleep in there is someone running:
  `tailscale file get --conflict=skip --loop`

Say tailscaled has received an `image.jpg` along with a bunch of PDFs the user is trying to save for later, and we already have an `image.jpg` in our Downloads directory.  `tailscale file get` will record an error along the lines of "hey, so I couldn't write to Downloads/image.jpg, because you told me to "skip" instead of "overwrite", and you've already got one there, so that file is wedged in your inbox."  it's a persistent error that will happen every time but is sort of ok.  we're happy to receive *other* files, just not that one.  and we don't want to start backing off and waiting ten minutes to receive any file because there's one wedged file.  on the other hand, if we ask tailscaled to tell us when it's got files, hey, it's got files! there's this `image.jpg` it keeps telling us about.

davideger:
It took me a little while to figure out how to hold the rate limiter correctly, but I think I have it now.  log messages are ratelimited, though we keep checking for new files every 5 seconds in the presence of persistent errors.  Thoughts?

cmd/tailscale/cli/file.go
```diff
@@ -384,6 +387,7 @@ func openFileOrSubstitute(dir, base string, action onConflict) (*os.File, error)
 
 func receiveFile(ctx context.Context, wf apitype.WaitingFile, dir string) (targetFile string, size int64, err error) {
 	rc, size, err := tailscale.GetWaitingFile(ctx, wf.Name)
+	defer rc.Close()
```
bradfitz, line 390: this panics if rc is nil on err

cmd/tailscale/cli/file.go
```diff
@@ -449,9 +443,62 @@ func runFileGet(ctx context.Context, args []string) error {
...
 		printf("moved %d/%d files\n", deleted, len(wfs))
 	}
+	return errs
+}
+
+func runFileGet(ctx context.Context, args []string) error {
+	if len(args) != 1 {
+		return errors.New("usage: file get <target-directory>")
+	}
+	log.SetFlags(0)
```
bradfitz, line 459: no need to do this every loop? move out earlier?

cmd/tailscale/cli/file.go
```diff
@@ -449,9 +443,62 @@ func runFileGet(ctx context.Context, args []string) error {
...
+	dir := args[0]
+	if dir == "/dev/null" {
+		return wipeInbox(ctx)
+	}
+
+	if fi, err := os.Stat(dir); err != nil || !fi.IsDir() {
+		return fmt.Errorf("%q is not a directory", dir)
+	}
+	nlLogf := logger.Logf(func(format string, args ...any) {
+				printf(format + "\n", args...)
```
bradfitz, line 470: gofmt?

cmd/tailscale/cli/file.go
davideger: thanks for the catch, moved down past the err check here.

cmd/tailscale/cli/file.go
davideger: done

cmd/tailscale/cli/file.go
davideger: not sure i follow.

bradfitz: ah, whoops. I misread!

cmd/tailscale/cli/file.go
```diff
@@ -449,9 +443,62 @@ func runFileGet(ctx context.Context, args []string) error {
...
+	nlLogf := logger.Logf(func(format string, args ...any) {
+		printf(format+"\n", args...)
+	})
+	if getArgs.loop {
+		// the rate limiter library effectively uses half of our quota for [RATELIMIT]
+		// messages.  burst has to be at least 2 so we can save up quota for printing
+		// a real message and a summary RATELIMIT message.
+		quietLogf := logger.RateLimitedFn(nlLogf, 15*time.Minute, 4, 125)
+		quietOutln := func(err error) {
+			quietLogf(strings.ReplaceAll(err.Error(), "%", "%%"))
```
bradfitz, line 478: `quietLogf("%v", err)` instead of shoving it into the format string?

davideger: That's actually intentional.  We don't want to rate limit all messages together where the format string is "%v" ;-)  Then we'd get a "random sample error message" every half an hour.

It's odd enough that I changed it when you asked and then remembered why I'd done it, and then changed it back, so I've added a  comment.

types/logger/logger.go
```diff
@@ -164,7 +164,9 @@ func RateLimitedFnWithClock(logf Logf, f time.Duration, burst int, maxCache int,
 		msgLim   = make(map[string]*limitData) // keyed by logf format
 		msgCache = list.New()                  // a rudimentary LRU that limits the size of the map
 	)
-
+	if burst < 2 {
+		panic("RateLimiter requires burst >= 2 so it can always emit a rate limit warning in addition to one log.")
```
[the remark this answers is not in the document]
davideger: sure

cmd/tailscale/cli/file.go
bradfitz: I'm already not a fan of the rate limiting we have in tailscaled. Why's it spreading to the CLI?

Especially with all this mechanism to fight the rate limiter that you added in this PR?

Can we just.... not rate limit here?

You already have the 5 second sleep on failures. (I thought that was going to be an incremental backoff with logtail/backoff.BackOff?). But in any case, what's the point of the quietLogf?

davideger: Here's the scenario is this: You're running `tailscale file get --skip --loop ~/Downloads` on your linux file server in some systemd user unit to babysit it and capture logs.  Let's call this the "taildrop receiver."  This setup is a bit better than sudo'ing it because as you, it can't be tricked into overwriting system files: it just doesn't have permission.

Ok, so you're browsing on your phone reading papers and slinging the random pdf at your file server.  Then you send some `paper.dvi` at your file server, but there's already a `paper.dvi` in the taildrop directory.  So that file is wedged, but it's one of many you're uploading.  However, every time the taildrop receiver polls tailscaled for new files (every 5 seconds, since there's a persistent error), it logs this same persistent error.  Soon your logs are filling up with this same persistent error.  Every 5 seconds.  Nothing's changed since last time.  It's possible you notice you didn't receive the new `paper.dvi` and look at the logs and see the issue and fix it.  If so great.  You move the other paper to a new name, the taildrop receiver delivers the new file, and the log stops printing out this error message.

If not, in a day you're storing ~1.3MB of logs saying you couldn't receive this `paper.dvi`  With rate limiting to 30 minutes that's down to about ~60KB.  Maybe this is all chump change?  It's duplicated for each file though.  Seems a waste.

Alternates:
1 We could print the error to the log every time, no rate limiting beyond our 5 second persistent error backoff.  It's 1.3MB/day/error.  Acceptable maybe?  Much better than filling the log at line rate.
2  We could save a map of every error we've seen here and print them exactly once.
- Con: there's no way to notice the error "is not happening any more" -- once suppressed, always supressed.
[   Since we're explicitly giving up after 100 failed files in any batch, we've effectively capped the number of log messages we might see if a peer (future feature) starts sending us 2x of a bunch of differently named files, so at least there's not a memory leak attack. ]

Do you have a preference for one of the above two instead of this approach?

The reason I didn't go with your suggestion of using backoff.BackOff for the entire receiver (just the logs) was to keep the flow of other, non-wedged files coming in at a reasonable rate.

bradfitz: I'd keep a map keyed off WaitingFile to track its most recent error and when it started failing, logging only on change, and maybe a summary of outstanding errors every N minutes (10, 15, 30? you pick)

I'm waiting for this to grow a Prometheus exporter too so you could plug in monitoring 😂

davideger: I like it: errors get logged on their rising and falling edges.

 i'm ambivalent to the periodic flushing, but it's in this version.  could easily be ripped out.  also not sure how clear `groundhogsDay` is to the younger generation, so if you or they have better naming ideas i'm all ears.

cmd/tailscale/cli/file.go
```diff
@@ -449,9 +442,99 @@ func runFileGet(ctx context.Context, args []string) error {
...
+		return
+	}
+	t := time.Now()
+	for k, v := range gdl.log {
+		if t.Sub(v.lastReported) > gdl.stutterTime {
+			outln("Persistent error for", t.Sub(v.firstSeen), ":", k)
+			v.lastReported = t
+		}
+	}
+}
```
bradfitz, line 499: Let's delete all this new log limiting code for now and just get this merged in the basic form that logs without limits.

davideger: done.

cmd/tailscale/cli/file.go
```diff
@@ -387,54 +389,45 @@ func receiveFile(ctx context.Context, wf apitype.WaitingFile, dir string) (targe
...
-	if fi, err := os.Stat(dir); err != nil || !fi.IsDir() {
-		return fmt.Errorf("%q is not a directory", dir)
-	}
-
+func runFileGetOneBatch(ctx context.Context, dir string) []error {
 	var wfs []apitype.WaitingFile
 	var err error
-	for {
+	var errs []error
+	for len(errs) < 2 {
```
bradfitz, line 409: why 2? add a little comment saying what's special about this number?

cmd/tailscale/cli/file.go
```diff
@@ -387,54 +389,45 @@ func receiveFile(ctx context.Context, wf apitype.WaitingFile, dir string) (targe
...
-			return err
+			errs = append(errs, err)
 		}
 	}
 
-	var errs []error
 	deleted := 0
 	for _, wf := range wfs {
+		if len(errs) > 100 {
+			// Likely, everything is broken. Let's just stop.
```
bradfitz, line 428: just stop this batch, right? the outer loop will still run forever?

bradfitz (approved):
Sorry, life stuff came up last week.

Looks good, thanks. One minor question...

cmd/tailscale/cli/file.go
davideger: converted to 0, which I think is correct having read the code i was calling.

i had thought `GetWaitingFile` might return an error if there were no files waiting. (1 error), and I wanted the loop to go around at least once through `WaitForFiles` and a `GetWaitingFile` re-attempt if there were initially no files waiting.  looking at the code, that's not how `GetWaitingFile` works.

cmd/tailscale/cli/file.go
davideger: that's right, the outer loop will still run forever.  this short-circuiting is mostly to help limit logging (or error memory map size as we briefly had in this cl) if we get flooded with incoming files and have persistent errors.

bradfitz (approved):

davideger:
I see an approve (thank you) but no merge yet?

PR #4232, opened 2022-03-20
prober: run all probes once on initial registration.

danderson:
Turns out, it's annoying to have to wait the entire interval
before getting any monitorable data, especially for very long
interval probes like hourly/daily checks.


maisem (approved):
Heh

PR #4234, opened 2022-03-20
cmd/tailscale, ipn/ipn{local,server}: add start of CLI admin API + over Noise

bradfitz:


bradfitz:
Early look. Needs polish, docs, etc.

Also needs a server push, to fix the colons in the HTTP basic auth "username" (NodeKey) over Noise.

cmd/tailscale/cli/admin.go
```diff
@@ -0,0 +1,133 @@
...
+		},
+	},
+}
+
+var adminArgs struct {
+	tokenFile string
+	tailnet   string
+	apiBase   string
+}
+
```
maisem, line 48: let's make this an actual client library.

ipn/ipnlocal/local.go
```diff
@@ -3253,3 +3253,35 @@ func (b *LocalBackend) DoNoiseRequest(req *http.Request) (*http.Response, error)
...
+		http.Error(w, "no node public key", http.StatusBadGateway)
+		return
+	}
+
+	outR := r.Clone(r.Context())
+	outR.RequestURI = ""
+	outR.URL.Scheme = "https"
+	outR.URL.Host = "unused"
+
+	outR.SetBasicAuth(nodePub.UntypedHexString(), "")
```
maisem, line 3274: i think you want nodePub.MarshalText

bradfitz: nope. read my chats to you :)

cmd/tailscale/cli/admin.go
bradfitz: For sure but not yet.

crawshaw: when it's a real client library please make it replace the partially-done one we have in corp

bradfitz: The problem is nobody wants a client library in oss with our massive go.mod and bleeding edge Go dependency.

So a client library really needs to live elsewhere and work on older Gos, so I don't intend to do it in this commit (as it's the wrong repo)

The separate repo would probably then need to be like https://github.com/tailscale/tscert which we made for Caddy, which includes a thinned down version of the client/tailscale package for speaking to the LocalAPI.  (as we'll want the real control API to use Local API for the Noise proxying)

maisem: Shouldn't [lazy module loading](https://go.dev/ref/mod#lazy-loading) make our go.mod size irrelevant? it wouldn't fix the older Go version problem though.

cmd/tailscale/cli/admin.go
```diff
@@ -0,0 +1,231 @@
...
+			return nil, err
+		}
+		if st.BackendState != "Running" {
+			return nil, fmt.Errorf("Tailscale must be running; currently in state %q", st.BackendState)
+		}
+		if st.CurrentTailnet == nil {
+			return nil, fmt.Errorf("no CurrentTailnet in status")
+		}
+		tailnet = st.CurrentTailnet.Name
+		// TODO(bradfitz): put apiBase in *ipnstate.TailnetStatus? update apiBase here?
```
maisem, line 83: I can send a PR for this. `ControlURL`?

cmd/tailscale/cli/admin.go
```diff
@@ -0,0 +1,231 @@
...
+		st, err := tailscale.StatusWithoutPeers(context.Background())
+		if err != nil {
+			return nil, err
+		}
+		if st.BackendState != "Running" {
+			return nil, fmt.Errorf("Tailscale must be running; currently in state %q", st.BackendState)
+		}
+		if st.CurrentTailnet == nil {
+			return nil, fmt.Errorf("no CurrentTailnet in status")
+		}
```
maisem, line 81: this will probably be easier to read if you split the setup between client types

maisem: i mean this entire function not just the highlighted part

PR #4239, opened 2022-03-20
tailcfg: add DERPRegion anycast fields

bradfitz:


tailcfg/derpmap.go
```diff
@@ -68,6 +68,20 @@ type DERPRegion struct {
 	// away to a new region without Avoid set.
 	Avoid bool `json:",omitempty"`
 
+	// AnycastIPv4 and AnycastIPv6, if non-empty, specify that
+	// this region is discoverable via Anycast STUN queries. The
+	// STUN responses will indicate which region replied.
```
crawshaw, line 73: So in this model, there is still one region per BigDERP node deployed? But each region entry in the map repeats the anycast address?

Is the idea here that clients will continue to send STUNs to every region? I've been hunting for a way for our US+EU clients to avoid sending a single anycast STUN + a STUN to India. Maybe we prefer anycast and try a second round of STUNs if the RTT exceeds some threshold? (I have been trying to make that server-side logic inside the anycast stun server, but it's tricky.)

Issue #4113, opened 2022-03-09
control/controlbase: reduce memory usage of `Conn`


bradfitz:
The `controlbase.Conn` is currently:

```go
const (
        // maxMessageSize is the maximum size of a protocol frame on the
        // wire, including header and payload.
        maxMessageSize = 4096
)

type Conn struct {
        conn          net.Conn
        version       uint16
        peer          key.MachinePublic
        handshakeHash [blake2s.Size]byte
        rx            rxState
        tx            txState
}

// rxState is all the Conn state that Read uses.
type rxState struct {
        sync.Mutex
        cipher    cipher.AEAD
        nonce     nonce
        buf       [maxMessageSize]byte
        n         int    // number of valid bytes in buf
        next      int    // offset of next undecrypted packet
        plaintext []byte // slice into buf of decrypted bytes
}

// txState is all the Conn state that Write uses.
type txState struct {
        sync.Mutex
        cipher cipher.AEAD
        nonce  nonce
        buf    [maxMessageSize]byte
        err    error // records the first partial write error for all future calls
}
```

This is a tracking bug to investigate reducing the size of Conn, which is currently over 8KB.

This doesn't matter for the client, but it'll matter server-side with tons of connections.

I suspect most most of the tx pools are empty at any moment, for instance. That halves the size.

And for reading, we probably don't need the recv buf either until the 3 byte header read unblocks.

So we can pool both.

For #3488

/cc @danderson @maisem @josharian @crawshaw

crawshaw:
Ideally this will be in trunkd, so we do get to distribute the memory use.

bradfitz:
It already is in trunkd. We still care about memory in trunkd, though. I spent a fair bit of time working on it with the TLS+http2 connections.

PR #4240, opened 2022-03-20
control/controlbase: use less memory when idle (remove txState.buf)

bradfitz:
Uses 4KB less per Conn.

Updates #4113


bradfitz:
Will do the rxState buffer in its own PR, to keep them both smaller.

danderson (approved):

PR #4241, opened 2022-03-21
control/controlbase: use less memory when idle (remove rxState.buf)

### LINKED ISSUE #4113: control/controlbase: reduce memory usage of `Conn`

bradfitz:
Uses 4KB less per Conn.

Fixes #4113

### DIFF
--- a/control/controlbase/conn.go
+++ b/control/controlbase/conn.go
@@ -52,10 +52,11 @@ type rxState struct {
 	sync.Mutex
 	cipher    cipher.AEAD
 	nonce     nonce
-	buf       [maxMessageSize]byte
-	n         int    // number of valid bytes in buf
-	next      int    // offset of next undecrypted packet
-	plaintext []byte // slice into buf of decrypted bytes
+	buf       *maxMsgBuffer   // or nil when reads exhausted
+	n         int             // number of valid bytes in buf
+	next      int             // offset of next undecrypted packet
+	plaintext []byte          // slice into buf of decrypted bytes
+	hdrBuf    [headerLen]byte // small buffer used when buf is nil
 }
 
 // txState is all the Conn state that Write uses.
@@ -88,6 +89,10 @@ func (c *Conn) Peer() key.MachinePublic {
 // readNLocked reads into c.rx.buf until buf contains at least total
 // bytes. Returns a slice of the total bytes in rxBuf, or an
 // error if fewer than total bytes are available.
+//
+// It may be called with a nil c.rx.buf only if total == headerLen.
+//
+// On success, c.rx.buf will be non-nil.
 func (c *Conn) readNLocked(total int) ([]byte, error) {
 	if total > maxMessageSize {
 		return nil, errReadTooBig{total}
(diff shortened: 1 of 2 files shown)


bradfitz:
I had to mail this quickly so haven't given it my usual once-over. Will do later before submitting, but just in case you want to review it early....

control/controlbase/conn.go
```diff
@@ -96,8 +97,26 @@ func (c *Conn) readNLocked(total int) ([]byte, error) {
 		if total <= c.rx.n {
 			return c.rx.buf[:total], nil
 		}
-
-		n, err := c.conn.Read(c.rx.buf[c.rx.n:])
+		var n int
+		var err error
+		if c.rx.buf == nil && c.rx.n == 0 && total == headerLen {
```
danderson, line 102: This branch is suspect: if one of the latter two conditions is false, the else would end up reading into a nil buffer. So either the latter conditions aren't necessary and are always true here, or there are more alternatives needed?

danderson (commented):
Brain's mush today, so only did a very quick pass. I need to re-read more carefully tomorrow, the buffer dance is much trickier in the read path.

control/controlbase/conn.go
bradfitz: Tightened it up. PTAL.

control/controlbase/conn.go
```diff
@@ -96,8 +101,29 @@ func (c *Conn) readNLocked(total int) ([]byte, error) {
...
+			}
+			// Optimization to reduce memory usage.
+			// Most connections are blocked forever waiting for
+			// a read, so we don't want c.rx.buf to be allocated until
+			// we know there's data to read. Instead, when we're
+			// waiting for data to arrive here, read into the
+			// 3 byte hdrBuf:
+			n, err = c.conn.Read(c.rx.hdrBuf[:])
+			if n > 0 {
+				if n > headerLen {
```
danderson, line 118: This check's probably over-defensive, if net.Conn.Read ever returns n > buf, we have big problems :)

control/controlbase/conn.go
```diff
@@ -190,6 +216,14 @@ func (c *Conn) decryptOneLocked() error {
 		c.rx.next = 0
 	}
 
+	// Return out buffer to the pool if it's empty, lest we be
```
danderson, line 219: typo: out > our? I think? Unless this was copy pasted from the tx path and you meant "the out buffer", but this isn't an out buffer.

control/controlbase/conn_test.go
```diff
@@ -224,6 +226,55 @@ func TestConnStd(t *testing.T) {
...
+	if testing.Short() {
+		num = 100
+	}
+	runtime.GC()
+	var ms0 runtime.MemStats
+	runtime.ReadMemStats(&ms0)
+
+	for i := 0; i < num; i++ {
+		client, server := pair(t)
+		defer server.Close()
```
danderson, line 245: no client.Close?

danderson (approved):

control/controlbase/conn.go
bradfitz: Removed. It'll panic two lines later if so anyway.

control/controlbase/conn_test.go
bradfitz: done

PR #4242, opened 2022-03-21
ipn/ipn{local,server}: add TailnetStatus.ControlURL

maisem:
For https://github.com/tailscale/tailscale/pull/4234#discussion_r830668184


bradfitz (commented):
So it turns out I don't think we need this. I'm not actually sure what it helps, now that I've seen we don't need it.

Also, it doesn't have a distinction between the API server or not (we use `https://api.tailscale.com/api/....`)

PR #4250, opened 2022-03-21
prober: support adding key/value labels to probes.

danderson:
This lets especially dynamically-registered probes export their results with a bunch more dimensions to slice by in dashboards, so you can e.g. show "only probes for DERP in Singapore", or "only production backup checks".

The labels are still static after probe creation, because prometheus doesn't deal well with per-datapoint cardinality explosion. This API forces you to think a little bit about adding labels that are about the probe, not about a particular probing event.

Two commits. First one is a pure refactor that uses PrometheusVar for exporting probe results, without altering the exported data. The second commit alters the probe export to include custom labels.

### DIFF
--- a/prober/prober.go
+++ b/prober/prober.go
@@ -16,6 +16,7 @@ import (
 	"io"
 	"log"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 )
@@ -56,7 +57,7 @@ func (p *Prober) Expvar() expvar.Var {
 // Run executes fun every interval, and exports probe results under probeName.
 //
 // Registering a probe under an already-registered name panics.
-func (p *Prober) Run(name string, interval time.Duration, fun ProbeFunc) *Probe {
+func (p *Prober) Run(name string, interval time.Duration, labels map[string]string, fun ProbeFunc) *Probe {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	if _, ok := p.probes[name]; ok {
@@ -75,6 +76,7 @@ func (p *Prober) Run(name string, interval time.Duration, fun ProbeFunc) *Probe 
 		doProbe:  fun,
 		interval: interval,
 		tick:     ticker,
+		labels:   labels,
 	}
 	p.probes[name] = probe
 	go probe.loop()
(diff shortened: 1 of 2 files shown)


prober/prober.go
```diff
@@ -149,6 +107,11 @@ type Probe struct {
 	doProbe  ProbeFunc
 	interval time.Duration
 	tick     ticker
+
+	mu     sync.Mutex
+	start  time.Time
+	end    time.Time
+	result bool
```
bradfitz, line 117: what's result? has a result? has a good result?

bradfitz (approved):
Other than test failure.

prober/prober.go
```diff
@@ -192,35 +158,120 @@ func (p *Probe) run() {
...
+
+	bs, err := json.Marshal(out)
 	if err != nil {
-		v = 0
+		return fmt.Sprintf(`{"error": %q}`, err)
+	}
+	return string(bs)
+}
+
+func (v varExporter) WritePrometheus(w io.Writer, prefix string) {
```
maisem, line 234: way too much going on here to understand. Please add docs?

prober/prober.go
```diff
@@ -192,35 +158,120 @@ func (p *Probe) run() {
...
+	defer p.mu.Unlock()
+	p.end = end
+	p.result = err == nil
+}
+
+type varExporter struct {
+	p *Prober
+}
+
+func (v varExporter) String() string {
```
maisem, line 195: same

prober/prober.go
maisem: specifically i care about what fields it exports and what the output format looks like so that I can debug it.

maisem: and this should have a test to verify output matches whats expected.

danderson:
Wow, I thought I'd added tests for this. Sigh, mondays. Adding tests and stuff before un-drafting again.

danderson:
Okay, added more comments and tests for the two expvar formats. @maisem PTAL

danderson:
I ended up squashing the changes into the wrong commit of this PR, so I'll do a squash and merge to keep the timeline in main clean. Sorry for the confusion.

maisem (approved):

PR #4261, opened 2022-03-23
tsnet: fail with nice message with go1.18 on darwin

bradfitz:
Like we did in ead16b24ec64ef1c82956011f119c500ae360d07 for tailscaled.

Updates #4258
Due to golang/go#51759

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -15,6 +15,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


bradfitz:
cc @bmizerany

tsnet/tsnet.go
```diff
@@ -92,6 +93,9 @@ func (s *Server) Dial(ctx context.Context, network, address string) (net.Conn, e
 // Start connects the server to the tailnet.
 // Optional: any calls to Dial/Listen will also call Start.
 func (s *Server) Start() error {
+	if runtime.GOOS == "darwin" && runtime.Version() == "go1.18" {
+		log.Fatalf("Tailscale is broken on macOS with go1.18 due to upstream bug https://github.com/golang/go/issues/51759; use 1.18.1+ or Tailscale's Go fork")
```
bradfitz, line 97: yes, this is a lot.Fatalf in a func returning an error. Meh.

It's better than a panic, which also brought down the whole program! :)

maisem: Didn't we do something with compile time checks for the go version?

bradfitz: You're thinking of 1db46919ab02676faa65535c18da9c45988e3a30 which made Go 1.17 have nice error messages when building HEAD. This is about using Go 1.18(.0) to build anything.  (we'd need to cherry-pick this back)

maisem: Oh I was wondering if we could detect go 1.18.0 vs go 1.18.X via build tags

This is fine, it's better to have the error at least tell you what's wrong.

maisem (approved):

PR #4268, opened 2022-03-23
envknob, ipn/ipnlocal: add SSH admin kill switch

bradfitz:
Updates #3802

### DIFF
--- a/envknob/envknob.go
+++ b/envknob/envknob.go
@@ -142,3 +142,10 @@ func LookupInt(envVar string) (v int, ok bool) {
 // UseWIPCode is whether TAILSCALE_USE_WIP_CODE is set to permit use
 // of Work-In-Progress code.
 func UseWIPCode() bool { return Bool("TAILSCALE_USE_WIP_CODE") }
+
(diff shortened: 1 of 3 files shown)


envknob/envknob.go
```diff
@@ -142,3 +142,6 @@ func LookupInt(envVar string) (v int, ok bool) {
 // UseWIPCode is whether TAILSCALE_USE_WIP_CODE is set to permit use
 // of Work-In-Progress code.
 func UseWIPCode() bool { return Bool("TAILSCALE_USE_WIP_CODE") }
+
+// CanSSHD is whether the Tailscale SSH server is allowed to run.
```
maisem, line 146: Can you clarify what happens if ssh is already enabled?

bradfitz: It doesn't start.

bradfitz: It's not allowed to run. I think the comment's pretty clear already.

bradfitz: That is, it says nothing about state transitions here at all.

bradfitz: PTAL

maisem (approved):

PR #4272, opened 2022-03-24
client/tailscale, cmd/tailscale, localapi: add 'tailscale nc'

bradfitz:
This adds a "tailscale nc" command that acts a bit like "nc", but
dials out via tailscaled via localapi.

This is a step towards a "tailscale ssh", as we'll use "tailscale nc"
as a ProxyCommand for in some cases (notably in userspace mode).

But this is also just useful for debugging & scripting.

Updates #3802

RELNOTE=tailscale nc


ipn/localapi/localapi.go
```diff
@@ -542,6 +546,84 @@ func (h *Handler) serveSetExpirySooner(w http.ResponseWriter, r *http.Request) {
...
+		_, err := io.Copy(outConn, reqConn)
+		errc <- err
+	}()
+	<-errc
+}
+
+// drainBufConn is a net.Conn with an initial bunch of bytes in a
+// bufio.Reader. Read drains the bufio.Reader until empty, then passes
+// through subsequent reads to the Conn directly.
+type drainBufConn struct {
```
maisem, line 611: We have this in a few places, might be time to put it in `netutil`?

client/tailscale/tailscale.go
```diff
@@ -419,6 +420,77 @@ func SetDNS(ctx context.Context, name, value string) error {
...
+	if !ok {
+		res.Body.Close()
+		return nil, errors.New("http Transport did not provide a writable body")
+	}
+	return &wrappedConn{switchedConn, rwc}, nil
+}
+
+type wrappedConn struct {
+	net.Conn
+	rwc io.ReadWriteCloser
```
maisem, line 479: Go question, can we do:
```
type wrappedConn struct {
  net.Conn
  io.ReadWriteCloser
}

wc := wrappedConn{conn, resp.Body}
```

client/tailscale/tailscale.go
```diff
@@ -419,6 +420,77 @@ func SetDNS(ctx context.Context, name, value string) error {
...
+	rwc, ok := res.Body.(io.ReadWriteCloser)
+	if !ok {
+		res.Body.Close()
+		return nil, errors.New("http Transport did not provide a writable body")
+	}
+	return &wrappedConn{switchedConn, rwc}, nil
+}
+
+type wrappedConn struct {
+	net.Conn
```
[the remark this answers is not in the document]
bradfitz: there's a defer already

client/tailscale/tailscale.go
bradfitz: Try: https://go.dev/play/p/Q22iddFt1WL

maisem: So it doesn't work if there is any ambiguity? 

https://go.dev/play/p/9bhuCnlGezi

client/tailscale/tailscale.go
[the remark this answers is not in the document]
maisem: Where?

ipn/localapi/localapi.go
bradfitz: https://github.com/tailscale/tailscale/pull/4273

bradfitz:
PTAL

maisem (approved):

PR #4275, opened 2022-03-24
client/tailscale, cmd/tailscale, localapi: add 'tailscale nc' (actually)

bradfitz:
Adds missing file from fc12cbfcd381b31de905ac431d76fb8c054006a2.

GitHub was having issues earlier and it was all green because the
checks never actually ran, but the DCO non-Actions check at least did,
so "green" and I merged, not realizing it hadn't really run anything.

Updates #3802

### DIFF
--- a/cmd/tailscale/cli/nc.go
+++ b/cmd/tailscale/cli/nc.go
@@ -1,0 +1,63 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package cli
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+
+	"github.com/peterbourgon/ff/v3/ffcli"
+	"tailscale.com/client/tailscale"
+)
+
+var ncCmd = &ffcli.Command{
+	Name:       "nc",
+	ShortUsage: "nc <hostname-or-IP> <port>",
+	ShortHelp:  "Connect to a port on a host, connected to stdin/stdout",
+	Exec:       runNC,
+}
+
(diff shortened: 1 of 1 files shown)


cmd/tailscale/cli/nc.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+var ncCmd = &ffcli.Command{
+	Name:       "nc",
+	ShortUsage: "nc <hostname-or-IP> <port>",
+	ShortHelp:  "Connect to a port on a host, connected to stdin/stdout",
+	Exec:       runNc,
+}
+
+func runNc(ctx context.Context, args []string) error {
```
maisem, line 26: should this be runNC?

cmd/tailscale/cli/nc.go
```diff
@@ -0,0 +1,63 @@
...
+		return errors.New("usage: nc <hostname-or-IP> <port>")
+	}
+
+	hostOrIP, portStr := args[0], args[1]
+	port, err := strconv.ParseUint(portStr, 10, 16)
+	if err != nil {
+		return fmt.Errorf("invalid port number %q", portStr)
+	}
+
+	// TODO(bradfitz): also add UDP too, via flag?
```
maisem, line 47: add flag and say it's not supported yet?
`nc` uses `-u`
```
     -u      Use UDP instead of the default option of TCP.
````

cmd/tailscale/cli/nc.go
bradfitz: done

cmd/tailscale/cli/nc.go
bradfitz: naah. I don't want to advertise something that's not supported.

maisem (approved):

PR #4277, opened 2022-03-24
cmd/tailscale/cli: add start of 'ssh' subcommand

bradfitz:
Updates #3802


cmd/tailscale/cli/ssh.go
```diff
@@ -0,0 +1,74 @@
...
+	arg, argRest := args[0], args[1:]
+	username, host, ok := strings.Cut(arg, "@")
+	if !ok {
+		host = arg
+		lu, err := user.Current()
+		if err != nil {
+			return nil
+		}
+		username = lu.Username
+	}
```
maisem, line 42: why not just pass `arg[0]` through directly? `ssh` should do the right thing

cmd/tailscale/cli/ssh.go
```diff
@@ -0,0 +1,74 @@
...
+	ssh, err := exec.LookPath("ssh")
+	if err != nil {
+		// TODO(bradfitz): use Go's crypto/ssh client instead
+		// of failing. But for now:
+		return fmt.Errorf("no system 'ssh' command found: %w", err)
+	}
+	argv := append([]string{
+		"ssh",
+		"-o", "ProxyCommand tailscale nc %h %p",
```
maisem, line 51: you might want to use the full path for tailscale? and possibly pass any sockets.

cmd/tailscale/cli/ssh.go
bradfitz: I mean, so would the string "lskdjflksdjf".  I dunno, because it just doesn't matter?

Initially I forgot it and it didn't work and then remembered that syscall.Exec requires it so I quickly wrote "ssh".

cmd/tailscale/cli/ssh.go
bradfitz: Done.

cmd/tailscale/cli/ssh.go
bradfitz: Left a comment.

bradfitz:
PTAL

cmd/tailscale/cli/ssh.go
```diff
@@ -0,0 +1,98 @@
...
+		"-o", fmt.Sprintf("ProxyCommand %s --socket=%s nc %%h %%p",
+			shellescape.Quote(tailscaleBin),
+			shellescape.Quote(rootArgs.socket),
+		),
+
+		// Explicitly rebuild the user@host argument rather than
+		// passing it through.  In general, the use of OpenSSH's ssh
+		// binary is a crutch for now.  We don't want to be
+		// Hyrum-locked into passing through all OpenSSH flags to the
+
```
maisem, line 67: remove line?

maisem (approved):

PR #4279, opened 2022-03-24
cmd/tailscale: make status --peers=false work earlier + in JSON mode

bradfitz:
And return an error if you use non-flag arguments.

### DIFF
--- a/cmd/tailscale/cli/status.go
+++ b/cmd/tailscale/cli/status.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"net"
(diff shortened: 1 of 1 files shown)


cmd/tailscale/cli/status.go
```diff
@@ -69,7 +70,14 @@ var statusArgs struct {
 }
 
 func runStatus(ctx context.Context, args []string) error {
-	st, err := tailscale.Status(ctx)
+	if len(args) > 0 {
+		return errors.New("unexpected non-flag arguments to 'tailscale status'")
+	}
+	getStatus := tailscale.Status
+	if !statusArgs.peers {
```
maisem, line 77: Where was this being used previously?

maisem (approved):

cmd/tailscale/cli/status.go
bradfitz: there was some manual filtering later in this file.

but I have a change coming after this one that lets you see filter the status (including CLI+JSON) to specific peers and I'll unify the two then

PR #4283, opened 2022-03-25
net/dnscache: do not call LookupIPFallback if the context was canceled.

maisem:
Two changes
```
net/dnsfallback: do not attempt lookups of IPs.

When the context is canceled, dc.dialOne returns an error from line 345.
This causes the defer on line 312 to try to resolve the host again, which
triggers a dns lookup of "127.0.0.1" from derp.
```

```
net/dnscache: do not call LookupIPFallback if the context was canceled.

Currently if the passed in host is an IP, Lookup still attempts to
resolve it with a dns server. This makes it just return the IP directly.
```

Updates tailscale/corp#4475

### DIFF
--- a/net/dnscache/dnscache.go
+++ b/net/dnscache/dnscache.go
@@ -312,7 +312,7 @@ func (d *dialer) DialContext(ctx context.Context, network, address string) (retC
 	defer func() {
 		// On failure, consider that our DNS might be wrong and ask the DNS fallback mechanism for
 		// some other IPs to try.
-		if ret == nil || d.dnsCache.LookupIPFallback == nil || dc.dnsWasTrustworthy() {
+		if ret == nil || ctx.Err() != nil || d.dnsCache.LookupIPFallback == nil || dc.dnsWasTrustworthy() {
 			return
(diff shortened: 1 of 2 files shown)


danderson (changes requested):
Won't this break fallback resolution in the case where your DNS is misconfigured and the request times out? For this to work, we'd have to nest contexts somehow and give the fallback lookup some time to be attempted still.

maisem:
> Won't this break fallback resolution in the case where your DNS is misconfigured and the request times out? 

It doesn't change anything semantically, in the sense that if the context was canceled the call to lookup was going to fail anyway as the context is reused. This primarily prevents a few function calls and some logs. See attached corp bug on how it's flaking, it's basically trying to lookup "127.0.0.1" from derp.


> For this to work, we'd have to nest contexts somehow and give the fallback lookup some time to be attempted still.

I agree that's something we should do but that seems independent of this change because of what I said above.

danderson (approved):
Okay, I agree this is still broken like before, but with less noise. Progress!

PR #4284, opened 2022-03-25
Allow subnet routing in tsnet

juliens:
This PR alllows subnet routing in tsnet.
To be able to handle subnet routing, I have to add a mechanism to know if tcp conn are handled by the "forwardTCP" or not.


clarkmcc:
Hi, I'm wondering if this feature will ever be supported? I'd like to embed tsnet in my application so that I can expose specific IP addresses that are accessible to my application, to my tailscale network. This requires that tsnet be able to act as a subnet router.

PR #4288, opened 2022-03-25
ipn/ipnstate: put SSH Host Keys in ipnstate.PeerStatus

bradfitz:
Updates #3802

### DIFF
--- a/ipn/ipnstate/ipnstate.go
+++ b/ipn/ipnstate/ipnstate.go
@@ -145,6 +145,9 @@ type PeerStatus struct {
 	PeerAPIURL   []string
 	Capabilities []string `json:",omitempty"`
 
+	// SSH_HostKeys are the node's SSH host keys, if known.
+	SSH_HostKeys []string `json:"sshHostKeys,omitempty"`
+
 	// ShareeNode indicates this node exists in the netmap because
 	// it's owned by a shared-to user and that node might connect
 	// to us. These nodes should be hidden by "tailscale status"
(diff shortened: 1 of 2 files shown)


ipn/ipnstate/ipnstate.go
```diff
@@ -145,6 +145,9 @@ type PeerStatus struct {
 	PeerAPIURL   []string
 	Capabilities []string `json:",omitempty"`
 
+	// SSH_HostKeys are the node's SSH host keys, if known.
+	SSH_HostKeys []string `json:"sshHostKeys,omitempty"`
```
maisem, line 149: views.Slice[string]?

bradfitz: I'd rather do them all at the same time. For now I value consistency.

maisem: we kinda already are inconsistent here, but sure.

bradfitz: Also, you'd want the pointer. Above we do have one `Tags *views.Slice[string]`

maisem: e.g. Tags and PrimaryRoutes

maisem (approved):

ipn/ipnstate/ipnstate.go
bradfitz: I really wish we didn't need that pointer. Ugh.

/cc @dsnet (for omitzero in json)

maisem: yeah you'd need the pointer for `omitempty` to work

PR #4291, opened 2022-03-26
tempfork: temporarily fork gliderlabs/ssh and x/crypto/ssh

bradfitz:
While we rearrange/upstream things.

Updates #3802


maisem (commented):
I don't see the crypto ssh fork

bradfitz:
Where'd you look? Because it's at the same as its Go import path would suggest: https://github.com/tailscale/golang-x-crypto

bradfitz:
Oh, you only looked at the files changed.

See the commit message text: https://github.com/tailscale/tailscale/pull/4291/commits/5567039434ad0014e99b40585ffef76b449c84dd

maisem (approved):
Ah, sorry I was looking at the first message in the PR.

PR #4297, opened 2022-03-27
prober: used keyed initializer for LimitedReader.

danderson:
Reported by go vet.

### DIFF
--- a/prober/http.go
+++ b/prober/http.go
@@ -49,7 +49,7 @@ func probeHTTP(ctx context.Context, url string, want []byte) error {
 		return fmt.Errorf("fetching %q: status code %d, want 200", url, resp.StatusCode)
 	}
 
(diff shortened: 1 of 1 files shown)


prober/http.go
```diff
@@ -49,7 +49,10 @@ func probeHTTP(ctx context.Context, url string, want []byte) error {
 		return fmt.Errorf("fetching %q: status code %d, want 200", url, resp.StatusCode)
 	}
 
-	bs, err := io.ReadAll(&io.LimitedReader{resp.Body, maxHTTPBody})
+	bs, err := io.ReadAll(&io.LimitedReader{
```
bradfitz, line 52: I would just use the constructor: https://pkg.go.dev/io#LimitReader

and keep it on one line like you had it

bradfitz (approved):

prober/http.go
danderson: Oops, duh, fixed.

PR #4298, opened 2022-03-27
util/groupmember: remove redundant code

### LINKED ISSUE #3001: util/groupmember: delete package once we use Go 1.18
The util/groupmember package will become redundant once we have:

https://github.com/golang/go/commit/30faf968b1f348e944db3bde24c13462125007b1

... in our tree.

Either via Go 1.18 or cherry-picking it into tailscale/go.

/cc @maisem @josharian

maisem:
Now that we have https://github.com/golang/go/commit/30faf968b1f348e944db3bde24c13462125007b1
this is no longer needed.

Fixes #3001


util/groupmember/groupmember.go
```diff
@@ -7,14 +7,44 @@
...
+	}
+	for _, ugid := range ugids {
+		if gid == ugid {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+var groupIDCache sync.Map // of string
```
bradfitz, line 37: // from string (group name) to string (Gid)

util/groupmember/groupmember.go
```diff
@@ -7,14 +7,44 @@
...
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+var groupIDCache sync.Map // from string (group name) to string (Gid)
+
+func getGroupID(groupName string) (string, error) {
+	s, ok := groupIDCache.Load(groupName)
```
bradfitz, line 40: caches for security stuff like this scare me a bit.

also, a more common pattern I think is:

      if s, ok := groupIDCache.Load(groupName).(string); ok {
             return s, nil
       }

That's safe even if somebody puts the wrong value type in that key.

bradfitz: But I guess it was already caching, so sure?

bradfitz (approved):

util/groupmember/groupmember.go
maisem: I can just drop the cache

maisem: Dropped the cache

maisem: PTAL

bradfitz: Either way. Or a TODO to investigate whether it matters.

In any case, the cache should probably be up a level on the func itself (is this user in this group?) and only cache positive caching for 10 seconds or so.

bradfitz (approved):

PR #4301, opened 2022-03-28
net/netutil: add CheckIPForwardingLinux

### LINKED ISSUE #4300: Minor: sysctl issue with interfaces with "." in the name
### What is the issue?

I've got some interfaces that have a period in their name, e.g. "vlan.130".  tailscale seems to use `sysctl` to check for forwarding status, which doesn't work.  Tailscale reports:

```
Warning: Couldn't check system's IP forwarding configuration, subnet routing/exit nodes may not work: couldn't check net.ipv4.conf.vlan.130.forwarding (exit status 255)

# sysctl net.ipv4.conf.vlan.130.forwarding
sysctl: cannot stat /proc/sys/net/ipv4/conf/vlan/130/forwarding: No such file or directory
# ls /proc/sys/net/ipv4/conf/vlan.130/forwarding
forwarding
```

Note that the directory node name should be "vlan.130" but sysctl interprets everything with a period as a delimiter.

### Steps to reproduce

_No response_

_No response_

### OS

Linux

### OS version

_No response_

### Tailscale version

_No response_

### Bug report

_No response_

maisem:
Combine the code between `LocalBackend.CheckIPForwarding` and
`controlclient.ipForwardingBroken`.

Fixes #4300

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -18,7 +18,6 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"os/exec"
 	"reflect"
 	"runtime"
 	"strings"
@@ -39,6 +38,7 @@ import (
 	"tailscale.com/net/dnsfallback"
 	"tailscale.com/net/interfaces"
 	"tailscale.com/net/netns"
+	"tailscale.com/net/netutil"
 	"tailscale.com/net/tlsdial"
 	"tailscale.com/net/tshttpproxy"
 	"tailscale.com/tailcfg"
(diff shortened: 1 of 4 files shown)


net/netutil/netutil.go
```diff
@@ -116,3 +126,143 @@ func (w wrappedConn) Write(bs []byte) (int, error) {
 func (w wrappedConn) Close() error {
 	return w.rwc.Close()
 }
+
+func reqProtocols(routes []netaddr.IPPrefix, state *interfaces.State) (v4, v6 bool) {
```
bradfitz, line 130: What does this do?

maisem: Restructured and added more comments

net/netutil/netutil.go
```diff
@@ -116,3 +126,195 @@ func (w wrappedConn) Write(bs []byte) (int, error) {
...
+		for _, pfx := range addrs {
+			localIPs[pfx.IP()] = true
+		}
+	}
+
+	for _, r := range routes {
+		// It's possible to advertise a route to one of the local
+		// machine's local IPs. IP forwarding isn't required for this
+		// to work, so we shouldn't warn for such exports.
+		if r.IsSingleIP() && localIPs[r.IP()] {
```
bradfitz, line 150: what about IsSingleIP && IsTailscaleIP? Does `routes` include `AllowedIPs` which includes the Tailscale self IPs? Document?

net/netutil/netutil.go
```diff
@@ -7,9 +7,19 @@ package netutil
```
bradfitz, line 7: let's move all this forwarding stuff into its own file in netutil.

net/netutil/netutil.go
```diff
@@ -116,3 +126,195 @@ func (w wrappedConn) Write(bs []byte) (int, error) {
...
+	}
+	return fmt.Sprintf(k, iface)
+}
+
+type sysctlFormat int
+
+const (
+	dotFormat   sysctlFormat = iota
+	slashFormat sysctlFormat = iota
+)
```
bradfitz, line 284: why a different format from lines 289-290 below?  I'd use the lower style.

net/netutil/netutil.go
```diff
@@ -116,3 +126,195 @@ func (w wrappedConn) Write(bs []byte) (int, error) {
...
+}
+
+// CheckIPForwardingLinux reports whether IP forwarding is enabled correctly
+// for subnet routing and exit node functionality on any interface.
+// The state param can be nil, in which case interfaces.GetState is used.
+// It returns an error if it is unable to determine if IP forwarding is enabled.
+// It returns a warning describing configuration issues if IP forwarding is
+// non-functional or partly functional.
+func CheckIPForwardingLinux(routes []netaddr.IPPrefix, state *interfaces.State) (warn, err error) {
+	const kbLink = "\nSee https://tailscale.com/kb/1104/enable-ip-forwarding/"
```
bradfitz, line 169:
```
if runtime.GOOS != "linux" {
   return nil, nil
}
```

?

net/netutil/netutil.go
```diff
@@ -116,3 +126,195 @@ func (w wrappedConn) Write(bs []byte) (int, error) {
...
+
+// ipForwardingEnabled reports whether the IP Forwarding is enabled for the
+// given interface.
+// The iface param determines which interface to check against, "" means to check
+// global config.
+// It tries to lookup the value directly from `/proc/sys`, and fallsback to
+// using `sysctl` on failure.
+func ipForwardingEnabled(p protocol, iface string) (bool, error) {
+	k := ipForwardSysctlKey(slashFormat, p, iface)
+	bs, err := os.ReadFile(filepath.Join("/proc/sys", k))
```
bradfitz, line 301: Is this trying to be Linux-only or portable, or what?

net/netutil/netutil.go
maisem: Documented that this should only be `--advertised-routes`.

net/netutil/netutil.go
maisem: bad copy paste. done

net/netutil/netutil.go
maisem: Done, and also added a thing to respond with a warning for BSD.

bradfitz (approved):
Assuming you've tested this well :)

maisem:
> Assuming you've tested this well :)

I tested what I could :)

PR #4315, opened 2022-03-29
tailscaled: reduce gc churn

raggi:
Usage of userspace-networking is increasing, and the aggressive GC
tuning causes a significant reduction in performance in that mode.

---

I was debugging a different issue and noticed that the GC frequency under moderate throughput (100mbit) in userspace mode was very high - multiple times per second.

The default tuning seems more suitable when running in userspace mode.

Observed metrics from the change tested against a 2015 macbook pro over wifi (not great test):

100mbit sustained UDP:
GOGC=10: dropping ~15% of packets
GOGC=100: dropping 0%

TCP throughput test:
GOGC=10:   100mbit
GOGC=100:  350mbit

Throughput wise for this scenario, the new tuning causes userspace mode and normal mode to perform the same.

### DIFF
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -25,7 +25,6 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime"
-	"runtime/debug"
 	"strings"
 	"syscall"
 	"time"
@@ -130,14 +129,6 @@ var subCommands = map[string]*func([]string) error{
 }
 
 func main() {
-	// We aren't very performance sensitive, and the parts that are
-	// performance sensitive (wireguard) try hard not to do any memory
-	// allocations. So let's be aggressive about garbage collection,
-	// unless the user specifically overrides it in the usual way.
-	if _, ok := os.LookupEnv("GOGC"); !ok {
-		debug.SetGCPercent(10)
-	}
-
 	printVersion := false
 	flag.IntVar(&args.verbose, "verbose", 0, "log verbosity level; 0 is default, 1 or higher are increasingly verbose")
 	flag.BoolVar(&args.cleanup, "cleanup", false, "clean up system state and exit")


cmd/tailscaled/tailscaled.go
```diff
@@ -181,6 +173,18 @@ func main() {
 		os.Exit(0)
 	}
 
+	if !strings.Contains(args.tunname, "userspace-networking") {
```
danderson, line 176: This assumes that environments in which we userspace-networking aren't memory constrained. I'm not sure that' a good bet for things like AWS Lambda environments?

All I remeber about this tuning is that it mattered a lot to mobile devices, I don't know how we ended up making this global. If we do want to be in the business of doing this, feels like we should figure out more broadly why/when we want this aggressive setting?

raggi: A typical outcome for this change would be 15mb -> 25mb heap. This doesn't change our churned set size, only the size of the "free list". I feel confident that for most users a 4x throughput performance and reduction in dropped traffic is worth the overhead.

This may be a problem for specific mobile environments, but I'm not sure how general of a problem it is. AIUI the latest macOS/iOS limits are 50mb now, and Tailscale's use recent Go changes reduced the cost of relocations.

Do we have platform specific budgets written down somewhere?

danderson: We don't have strict memory budgets figured out, no. The only hard limits we have come from iOS:
 - iOS <15: 15MiB, which has to include a bunch of static program overhead (https://tailscale.com/blog/go-linker/ has the gory details, I've forgotten the nuances).
 - iOS >=15: 50MiB

AFAIK, every other platform is unconstrained on a technical level, though we're usually a system-level daemon so we should be economical where possible, in both CPU and RAM. But those aren't hard limits, just "don't be so greedy that people notice tailscale hogging resources."

Unless I've forgotten about other constraints we have, my bias would be to only set the aggressive GC tune on iOS, and leave the runtime default (or a different, less aggressive custom tune if we have one we can defend rationally) everywhere else.

cc @bradfitz, thoughts?

raggi: I have updated the change with the following proposal:

Only apply this tuning to iOS now. I'd like to find a way to detect iOS 15 and remove it there too.

cmd/tailscaled/tailscaled_ios.go
```diff
@@ -0,0 +1,27 @@
...
+	// performance sensitive (wireguard) try hard not to do any memory
+	// allocations. So let's be aggressive about garbage collection,
+	// unless the user specifically overrides it in the usual way.
+
+	// TODO(raggi): detect ios 15+ and do not apply this there. This tuning
+	// creates a strong throughput reduction in userspace-networking mode as it
+	// clears the sync.Pools too frequently.
+	if _, ok := os.LookupEnv("GOGC"); !ok {
+		debug.SetGCPercent(10)
+	}
```
bradfitz, line 26: There is no tailscaled on iOS and even if there were, there'd be no way for users to set environment variables anyway.

bradfitz (approved):

PR #4323, opened 2022-03-29
Allow any port for HTTPS when using Noise over TLS

juanfont:
When connecting to `/ts2021` it is assumed that the control server is running in tcp/443 if using the fallback HTTPS method. This sometimes might not be the case :)

This PR allows to pass any addr to `controlhttp.Dial` - or if the addr does not include a port assume tcp/80 for http and tcp/443 for https.

Signed-off-by: Juan Font Alonso <juanfontalonso@gmail.com>

### DIFF
--- a/control/controlclient/noise.go
+++ b/control/controlclient/noise.go
@@ -7,7 +7,6 @@ package controlclient
 import (
 	"context"
 	"crypto/tls"
-	"fmt"
 	"math"
 	"net"
 	"net/http"
@@ -50,7 +49,9 @@ type noiseClient struct {
 	dialer       *tsdial.Dialer
 	privKey      key.MachinePrivate
 	serverPubKey key.MachinePublic
-	serverHost   string // the host:port part of serverURL
+	host         string // the host part of serverURL
+	httpPort     string // the default port to call
+	httpsPort    string // the fallback Noise-over-https port
 
 	// mu only protects the following variables.
 	mu       sync.Mutex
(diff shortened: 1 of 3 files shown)


control/controlclient/noise.go
```diff
@@ -65,8 +65,12 @@ func newNoiseClient(priKey key.MachinePrivate, serverPubKey key.MachinePublic, s
 		// If there is an explicit port specified use it.
 		host = u.Host
 	} else {
-		// Otherwise, controlhttp.Dial expects an http endpoint.
-		host = fmt.Sprintf("%v:80", u.Hostname())
+		// Otherwise, pass the usual ports to controlhttp.Dial
+		if u.Scheme == "https" {
```
maisem, line 69: The serverURL almost always has `https` as the scheme (except in tests), as written this will make TLS the default mode.

juanfont: I might've misunderstood what this code does, but I don't see how this would make TLS the default option for Noise.

The rationale for this is that we do use https for our integration tests, but usually not in 443...

juanfont: @maisem would it be possible for you to have another look at this?

maisem: Sorry this dropped off my radar.

The way it's written as it is, is to make it possible for tailscale to run in as many environments as possible. The way this works is by trying to reach out to the noise endpoint first over port 80 and then 500ms later over 443.
It prefers port 80 to avoid double encryption as noise is already encrypted and with TLS it is going to do noise-over-TLS, which isn't very useful except for environments where HTTP is completely blocked.

Having said that, you mentioned your requirement is to make this work in your integration test env which uses TLS. For that you could either not use TLS for the noise part, or if you really want to test it noise-over-TLS then I'd recommend adding an additional field to pass in as the TLS port instead of saying that the HTTP and HTTPS ports are the same.

control/controlclient/noise.go
```diff
@@ -65,22 +66,28 @@ func newNoiseClient(priKey key.MachinePrivate, serverPubKey key.MachinePublic, s
 	if err != nil {
 		return nil, err
 	}
-	var host string
+	var p80 string
```
maisem, line 69: plaintextPort
tlsPort

maisem: or httpPort and httpsPort to be consistent

juanfont:
@maisem a kind reminder :)

juanfont:
@maisem Hi Maisem, is there anything I can help with to move this along? Thanks!

maisem:
Apologize for the delay, I kept losing this. I just ran the tests in our other repo and it seems to work fine.

maisem (approved):

juanfont:
@maisem  \o/ thank you so much!

Would it be acceptable to raise a PR to bump the ClientCapabilityVersion, so we can detect when this patch was merged?

The idea would be to only offer Noise (thought /key) to those clients above that version.

maisem:
`//	28: 2022-03-09: client can communicate over Noise.`
Would that not work?

juanfont:
For those Headscale users exposing it in tcp/443 yes, but any other https port would require this patch.

maisem:
sure! something like?
```
//	39: 2022-08-15: client can talk Noise over arbitrary HTTPS port
```

juanfont:
That be perfect :)

Issue #3616, opened 2021-12-24
FR: Support IPv4 subnet routes even when `disableIPv4` is true


PaulMcMillan:
### What are you trying to do?

This is a bit of a tricky one.

I need to use `disableIPv4` to prevent Tailscale's 100.64/10 addresses from conflicting with existing CGNAT in other networks I access. I'm happy to use Tailscale's IPv6 addresses for talking to devices running Tailscale.

However, I also desire the ability to use tailscale to route IPv4 subnets.

Experimentally, v4 subnet routing seems to stop working when I set `disableIPv4` to true. This isn't unreasonable, but I want to have my cake and eat it too!

### How should we solve this?

Can we route IPv4 subnets over Tailscale's IPv6 addressing scheme?

### What is the impact of not solving this?

I can't use both `disableIPv4` and IPv4 subnet routing at the same time.

Subnet routing is a killer feature (it's why I pay for a personal account), so I'll probably go back to `disableIPv4: false` and manually stopping Tailscale when I access a network with conflicting CGNAT addresses.

Related: #1381

DentonGentry:
I really don't see a way to do this in a reasonable way. If the node has no Tailscale IPv4 address, I don't see what we would put in the IP source address field of the packet we send to the subnet router.

Barring further comment, we'll expect to close this during the next regular bug scrub.

PaulMcMillan:
Yeah, I agree it's kinda a weird ask, and may not be technically feasible.

What about using one of the 6 to 4 mappings? I could probably live with using an IPv6 mapped address rather than the native IPv4 ones if the tailscale router node knew how to forward that traffic.

normanr:
Is the plan to do IPv6-to-IPv4 at the "egress" site? Something like [tayga](http://www.litech.org/tayga/) (but built into `tailscaled`)?

DentonGentry:
https://github.com/tailscale/tailscale/issues/826#issuecomment-1120251404 describes one such mechanism available now.

normanr:
ahh, I wasn't obvious to me at first glace that #4330 also included the code to do the v6-to-v4 mapping (I thought it was only reserving the address space). It looks like netstack is doing _exactly_ what tayga could do, just integrated within tailscaled.

DentonGentry:
I think this works; without an IPv4 address I don't know what else we could do. Barring further comment. we'll expect to close this during the next regular bug scrub.

PaulMcMillan:
[...]

PR #4330, opened 2022-03-30
net/tsaddr, wgengine/netstack: add IPv6 range that forwards to site-relative IPv4

bradfitz:
This defines a new magic IPv6 prefix, fd7a:115c:a1e0:b1a::/64, a subset of our
existing /48, where the final 32 bits are an IPv4 address, and the middle 32 bits are
a user-chosen "site ID".

e.g., I can say my home LAN's "site ID" is "bbbb:0000" and then
advertise its IPv4 via IPv6, like:

    tailscale up --advertise-routes=fd7a:115c:a1e0:b1a:bbbb::/80

Then people in my tailnet can:

     $ curl '[fd7a:115c:a1e0:b1a:bbbb::10.2.0.230]'
     <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" ....

Or if only want to export 10.2.0.0/16:

    tailscale up --advertise-routes=fd7a:115c:a1e0:b1a:bbbb::10.2.0.0/112

(112 being 128-16)

Updates #3616, etc

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -431,6 +431,8 @@ func (ns *Impl) peerAPIPortAtomic(ip netaddr.IP) *uint32 {
 	}
 }
 
+var viaRange = tsaddr.TailscaleViaRange()
+
 // shouldProcessInbound reports whether an inbound packet should be
 // handled by netstack.
 func (ns *Impl) shouldProcessInbound(p *packet.Parsed, t *tstun.Wrapper) bool {
@@ -452,6 +454,9 @@ func (ns *Impl) shouldProcessInbound(p *packet.Parsed, t *tstun.Wrapper) bool {
 	}
 	if ns.isInboundTSSH(p) && ns.processSSH() {
 		return true
+	}
+	if p.IPVersion == 6 && viaRange.Contains(p.Dst.IP()) {
+		return ns.lb != nil && ns.lb.ShouldHandleViaIP(p.Dst.IP())
 	}
 	if !ns.ProcessLocalIPs && !ns.ProcessSubnets {
 		// Fast path for common case (e.g. Linux server in TUN mode) where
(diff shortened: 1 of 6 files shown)


bradfitz:
We'd probably want to make this web UI understand the range and format it nicer...

Screen Shot 2022-03-30 at 8 54 23 AM

crawshaw:
This looks to me like step one in automatically sharing out local nodes, excellent!

wgengine/netstack/netstack.go
```diff
@@ -775,6 +787,13 @@ func (ns *Impl) forwardUDP(client *gonet.UDPConn, wq *waiter.Queue, clientAddr,
 		backendRemoteAddr = &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: int(port)}
 		backendListenAddr = &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: int(srcPort)}
 	} else {
+		if dstIP := dstAddr.IP(); viaRange.Contains(dstIP) {
+			dial16 := dstIP.As16()
+			dstAddr = netaddr.IPPortFrom(
+				netaddr.IPv4(dial16[12], dial16[13], dial16[14], dial16[15]),
```
maisem, line 793: `netaddr.IPFrom4(dial16[12:])`?

maisem (approved):

wgengine/netstack/netstack.go
bradfitz: You mean `netaddr.IPFrom4(*(*[4]byte)(dial16[12:]))`? 🤮 😝

bradfitz: But https://github.com/golang/go/issues/46505 would make it slightly less gross.

bradfitz: Done. PTAL. Moved to a common func in tsaddr.

Issue #4326, opened 2022-03-30
debug: make DERP send frame (framePeerGone) as limited reply as "nobody's here"


bradfitz:
Imagine node A is trying to reach node B.

Node A thinks that Node B was last homed at DERP region, say, 5.

But let's say that either Node B disappeared or the control server's broken or Node A's client is broken or... whatever. Assume that Node A can connect to DERP-5 and send frame type `frameSendPacket` to send to B, but then...

_nothing_

Currently we just do nothing (besides increment some varz in `(*Server).recordDrop`), and Node A is none the wiser that Node B isn't in that region.

It'd be nice to send a message (rate-limited) telling Node A that Node B's not there so Node A can locally log "oh, well that's messed up.", which would be useful during debugging connectivity problems.

But we already have `framePeerGone` used for DRPO (#150) that's very close to what we want. We could probably just reuse it, perhaps with a reason byte (reasons being: because they disconnected (the current use, when not listed), or because you tried to send to them, thinking they were here).

/cc @DentonGentry @crawshaw @danderson @maisem

danderson:
Only question I have is what happens during startup race conditions.
 - Client B starts up, locates home derp, reports derp5 as home to control
 - Client A gets told B is at derp5
 - A connects to derp5, sends frame, receives peerGone
 - B connects to derp5
 - ???

If the client doesn't alter its behavior upon receiving peerGone, I guess it's fine and becomes just a debugging aid. Is there anything actionable and inexpensive we'd want to do upon receiving a peerGone?

bradfitz:
Your race is #212 which I'm still unhappy about :)

But yeah, I was thinking this would be purely for debugging and it wouldn't alter the behavior. Except it'd still clear the DRPO route mapping if set, but that's why the frame exists.

PR #4331, opened 2022-03-30
derp: add (*Server).IsClientConnectedForTest func. 

maisem:
This allows tests to verfiy that a DERP connection was actually
established.

Related to #4326
Updates tailscale/corp#2579

### DIFF
--- a/derp/derp_server.go
+++ b/derp/derp_server.go
@@ -391,6 +391,18 @@ func (s *Server) isClosed() bool {
 	return s.closed
 }
 
+// IsClientConnectedForTest reports whether the client with specified key is connected.
+// This is used in tests to verify that nodes are connected.
+func (s *Server) IsClientConnectedForTest(k key.NodePublic) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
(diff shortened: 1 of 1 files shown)


bradfitz:
`/derp_server` isn't a package :)

crawshaw (approved):

derp/derp_server.go
```diff
@@ -391,6 +391,18 @@ func (s *Server) isClosed() bool {
 	return s.closed
 }
 
+// IsClientConnected reports whether the client with specified key is connected.
+// This is used in tests to verify that nodes are connected to derp.
```
bradfitz, line 395: to DERP? or just "are connected." 

Are you going to poll this? 😬  I guess if it deflakes.

Can you rename to make it include `ForTest` in the name? See `git grep ForTest` in the root for examples.

bradfitz (approved):

derp/derp_server.go
maisem: Yeah I'm polling this, would be nice to have a callback or something. But that can happen in the future.
Renamed to `IsClientConnectedForTest`

PR #4339, opened 2022-04-01
tailcfg, ssh/tailssh: optionally support SSH public keys in wire policy

bradfitz:
Updates #3802

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -9,8 +9,10 @@
 package tailssh
 
 import (
+	"bytes"
 	"context"
 	"crypto/rand"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -40,6 +42,8 @@ import (
 	"tailscale.com/types/logger"
 )
 
+var sshVerboseLogging = envknob.Bool("TS_DEBUG_SSH_VLOG")
+
 // TODO(bradfitz): this is all very temporary as code is temporarily
 // being moved around; it will be restructured and documented in
 // following commits.
@@ -50,6 +54,9 @@ func Handle(logf logger.Logf, lb *ipnlocal.LocalBackend, c net.Conn) error {
 	if err != nil {
 		return err
 	}
+	// TODO(bradfitz): make just one server for the whole process. rearrange
+	// netstack's hooks to be a constructor given a lb instead. Then the *server
+	// will have a HandleTailscaleConn method.
 	srv := &server{
 		lb:             lb,
 		logf:           logf,
@@ -60,6 +67,10 @@ func Handle(logf logger.Logf, lb *ipnlocal.LocalBackend, c net.Conn) error {
 		return err
 	}
 	ss.HandleConn(c)
+
+	// Return nil to signal to netstack's interception that it doesn't need to
+	// log. If ss.HandleConn had problems, it can log itself (ideally on an
+	// sshSession.logf).
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -208,12 +243,36 @@ func (srv *server) evaluatePolicy(sshUser string, localAddr, remoteAddr net.Addr
...
+	}
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	if res.StatusCode != http.StatusOK {
+		return nil, errors.New(res.Status)
+	}
+	all, err := io.ReadAll(io.LimitReader(res.Body, 1<<10))
```
maisem, line 266: 1KB seems too small, maybe `4<<10`?

ssh/tailssh/tailssh.go
```diff
@@ -654,15 +722,15 @@ func matchRule(r *tailcfg.SSHRule, ci *sshConnInfo) (a *tailcfg.SSHAction, local
 	if r.RuleExpires != nil && ci.now.After(*r.RuleExpires) {
 		return nil, "", errRuleExpired
 	}
-	if !matchesPrincipal(r.Principals, ci) {
-		return nil, "", errPrincipalMatch
-	}
```
maisem, line 659: why is this moving?

ssh/tailssh/tailssh.go
```diff
@@ -609,6 +668,10 @@ type sshConnInfo struct {
 	// now is the time to consider the present moment for the
 	// purposes of rule evaluation.
 	now time.Time
+	// fetchPublicKeysURL, if non-nil, is a func to fetch the public
+	// keys of a URL. The strings are in the the typical public
+	// key "type base64-string [comment]" format seen at e.g. https://github.com/USER.keys
+	fetchPublicKeysURL func(url string) ([]string, error)
```
maisem, line 674: this is never nil?

ssh/tailssh/tailssh.go
```diff
@@ -124,6 +132,31 @@ func (srv *server) mayForwardLocalPortTo(ctx ssh.Context, destinationHost string
...
+	a, ci, _, err := srv.evaluatePolicy(sshUser, localAddr, remoteAddr, nil)
+	if err == nil && (a.Accept || a.HoldAndDelegate != "") {
+		// Policy doesn't require a cert.
+		return false
+	}
+
+	// Is there any rule that looks like it'd require a cert for
+	// this sshUser?
+	for _, r := range pol.Rules {
+		for _, p := range r.Principals {
```
maisem, line 151: should you also be checking against the sshUser here?

ssh/tailssh/tailssh.go
bradfitz: Done

ssh/tailssh/tailssh.go
bradfitz: I want to avoid fetching SSH certs in the case where the SSH user can't match anyway.

This moves it as late as possible.

bradfitz:
PTAL

ssh/tailssh/tailssh.go
```diff
@@ -124,6 +148,47 @@ func (srv *server) mayForwardLocalPortTo(ctx ssh.Context, destinationHost string
...
+	if err == nil && (a.Accept || a.HoldAndDelegate != "") {
+		// Policy doesn't require a public key.
+		return false
+	}
+
+	// Is there any rule that looks like it'd require a public key for this
+	// sshUser?
+	now := time.Now()
+	for _, r := range pol.Rules {
+		if r.RuleExpires != nil && now.After(*r.RuleExpires) {
```
maisem, line 169: add `IsExpired()` to the Rule?

ssh/tailssh/tailssh.go
```diff
@@ -124,6 +148,47 @@ func (srv *server) mayForwardLocalPortTo(ctx ssh.Context, destinationHost string
 	return ss.action.AllowLocalPortForwarding
 }
 
+// askForPubKey reports whether the SSH server, during the auth negotiation
+// phase, should requires that the client send an SSH public key. (or, more
+// specifically, that "none" auth isn't acceptable)
+func (srv *server) askForPubKey(sshUser string, localAddr, remoteAddr net.Addr) bool {
```
maisem, line 154: maybe `requiresPubKey`?

bradfitz: done

ssh/tailssh/tailssh.go
bradfitz: we generally don't have code in the tailcfg package.

also, having that would mean it would need to use time.Now() and we don't here.

but I'll see if there's anywhere to pull it out

bradfitz: done

maisem (approved):

Issue #4174, opened 2022-03-15
Support winget release channel officially


cwilliams5:
### What is the issue?

Minor issue...but I use winget for easy management, and we've missed the last 2 tailscale versions. Don't forget us xD

![image](https://user-images.githubusercontent.com/10100743/158312524-54bf40a9-17c7-490b-8b9e-a49109b72ba1.png)

### Steps to reproduce

winget upgrade tailscale
winget list --name tailscale
note package is 3 releases behind

### OS

Windows

### OS version

Win 11

### Tailscale version

irrelevant

DentonGentry:
Tailscale does not distribute a package via winget.

Is this the one? Says 1.18.2. https://winget.run/pkg/tailscale/tailscale
Or this one? Looks like 1.20.4. https://github.com/microsoft/winget-pkgs/tree/afd1f813d3def5f9aca4deb4d32defb92456b1c6/manifests/t/tailscale/tailscale

cwilliams5:
> Tailscale does not distribute a package via winget.

Well, lesson learned for me there then. I had assumed they were doing verification of some sort and packages were official. Go ahead and close this ticket if you want and I'll switch to an official release channel. I suppose perhaps you find value in maintaining this officially so no one else gets tricked and there is an easy package managed version, but hey I get attention must go lots of places. 

> Or this one? Looks like 1.20.4. https://github.com/microsoft/winget-pkgs/tree/afd1f813d3def5f9aca4deb4d32defb92456b1c6/manifests/t/tailscale/tailscale

That's the one I had picked, yes.

bradfitz:
It's at least downloading our official binaries. That's just a metadata repository, like https://chocolatey.org/ is.

We've been meaning to automate our Chocolatey metadata updates (that's #303) and I suppose we should do the same here. They'd probably use very similar mechanisms (send/submit PRs).

/cc @maisem @dblohm7

ItzLevvie:
> @cwilliams5: we've missed the last 2 tailscale versions. Don't forget us xD

One of the main reasons why that package is behind several versions is because the community has made more than 1000+ PRs in a single day which has hit various rate limits and broke Azure Pipelines for the winget-pkgs repository for a few days.

> @cwilliams5: Well, lesson learned for me there then.

It's all good as long as the application is being downloaded from official sources which is what's happening here.

What I mean by "official sources" is the Download URL that comes straight from the publisher despite the WinGet package being community-maintained.

[...]

Issue #2257, opened 2021-06-26
Make Windows package available in WinGet


MichelLopez:
The Tailscape Windows installer should be available from the Windows package manager WinGet.

More details at 
[https://docs.microsoft.com/en-us/windows/package-manager/winget/](https://docs.microsoft.com/en-us/windows/package-manager/winget/)
[https://github.com/microsoft/winget-cli](https://github.com/microsoft/winget-cli)

DentonGentry:
https://docs.microsoft.com/en-us/windows/package-manager/package/ lists MSIX, MSI, or EXE as a requirement. I guess the current NSIS installer is an EXE? Probably. We're tracking a move to MSI packaging in https://github.com/tailscale/tailscale/issues/595

https://github.com/microsoft/winget-pkgs is the package repository. We list on chocolatey now, which some articles say might be supported as a package repository in the future but not now.

bradfitz:
https://github.com/microsoft/winget-pkgs/tree/master/manifests/t/tailscale/tailscale

bradfitz:
I'm going to reopen this as a tracking bug for deciding whether we want to take over Winget maintenance, with some automation.

johnny-salz:
Hi, I tried to scan issues about that topic and this one seems like the most appropriate but it is still unclear if Tailscale officially present on winget? I had no luck with automatically updating it there, logs say that I need to download appropriate MSI for my machine. Apparently it tries to download x86 version on x64 machine?

```
> winget upgrade Tailscale.Tailscale
Found Tailscale [Tailscale.Tailscale] Version 1.96.3
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Successfully verified installer hash
Starting package install...
Installer failed with exit code: 1603
Installer log is available at: AppData\Local\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir\Tailscale.Tailscale.1.96.3-26-03-25-12-11-47.log
```

```
=== Logging started: 25.03.2026  12:01:39 ===
Action start 12:01:39: INSTALL.
Action start 12:01:39: FindRelatedProducts.
Action ended 12:01:39: FindRelatedProducts. Return value 1.
Action start 12:01:39: SaveAdminConsoleProperty.
Action ended 12:01:39: SaveAdminConsoleProperty. Return value 1.
Action start 12:01:39: SaveAdvertiseExitNodeProperty.
[...]

PR #4350, opened 2022-04-04
Automating winget submission

gnpaone:
Related to https://github.com/tailscale/tailscale/issues/4174 and https://github.com/tailscale/tailscale/issues/2257. Please create a Personal Access Token for winget to this repo and add in this workflow

### DIFF
--- a/.github/workflows/windows-winget.yml
+++ b/.github/workflows/windows-winget.yml
@@ -1,0 +1,38 @@
+name: Submit to winget
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+
+defaults:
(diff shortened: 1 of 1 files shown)


gnpaone:
Oh ok. But using Github Actions is an option too as winget-pkgs repo also exists in Github so it would be all integrated. Anyways, let this be an option too I guess.

DentonGentry:
At this point: thank you for the thought, but this is not how we want build artifacts to be generated.

gnpaone:
> At this point: thank you for the thought, but this is not how we want build artifacts to be generated.

Ok

PR #4360, opened 2022-04-05
tailcfg: add omitempty to all fields of Hostinfo

dsnet:
This reduces the noise when marshaling only a subset of this type.

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -452,16 +452,14 @@ type Service struct {
 // Because it contains pointers (slices), this type should not be used
 // as a value type.
 type Hostinfo struct {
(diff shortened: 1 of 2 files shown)


dsnet:
@crawshaw I don't know what "when all the iOS apps are updated with the latest swift version" quite means, so I'm unable to verify whether this change breaks anything.

crawshaw:
This JSON gets passed into the Swift frontend in iOS/macOS.

If a field not marked optional (`?`) is missing in the Swift JSON decoder, it errors out and crashes the frontend. So that's the thing to check before submitting this.

dsnet:
I think we're good since the struct definitions have `?` on them for the past 2 years now.

crawshaw (approved):

PR #4361, opened 2022-04-05
net/dns/resolver: return symbolic PTR records for TS service IPs

### LINKED ISSUE #1233: MagicDNS should have PTR records for tailscale service machines
MagicDNS should respond with PTR records for 100.100.100.100 and 100.101.102.103. It should also have A records for whatever hostnames are returned.

This would help any admins when they're looking at things like netstat to be able to find what those addresses are related to.

I tried looking up various forms of `hello.ipn.dev` (`hello-ipn-dev`, `hello.ipn.dev.beta.tailscale.net`, `hello-ipn-dev.tailscale.net.beta.tailscale.net`, etc) at 100.100.100.100, but nothing worked. Unexpectedly I discovered that `hello.ipn.dev` resolves perfectly fine using regular dns (which was a surprise to me).

twitchyliquid64:
Fixes #1233 

Open questions:

 - What should the specific domain be? I yeeted `nodedns.tailscale-internals` but very open to bike-shed. In particular, should we make it an actual real domain, and make a corresponding forward A/AAAA record resolvable via public DNS?
 - ~Should we provide a PTR record for `100.101.102.103` (hello) as well? Afaict its optional?~ Already works thanks to the existing PTR implementation for nodes on the tailnet

### DIFF
--- a/net/dns/resolver/tsdns.go
+++ b/net/dns/resolver/tsdns.go
@@ -34,6 +34,8 @@ import (
 	"tailscale.com/wgengine/monitor"
 )
 
+const dnsSymbolicFQDN = "magicdns.localhost-tailscale-daemon."
+
 // maxResponseBytes is the maximum size of a response from a Resolver. The
 // actual buffer size will be one larger than this so that we can detect
 // truncation in a platform-agnostic way.
@@ -553,6 +555,18 @@ func (r *Resolver) resolveLocal(domain dnsname.FQDN, typ dns.Type) (netaddr.IP, 
 		return netaddr.IP{}, dns.RCodeNameError
 	}
 
+	// We return a symbolic domain if someone does a reverse lookup on the
+	// DNS endpoint. To round out this special case, we also do the inverse
+	// (returning the endpoint IP if someone looks up the symbolic domain).
+	if domain == dnsSymbolicFQDN {
+		switch typ {
+		case dns.TypeA:
+			return tsaddr.TailscaleServiceIP(), dns.RCodeSuccess
+		case dns.TypeAAAA:
+			return tsaddr.TailscaleServiceIPv6(), dns.RCodeSuccess
+		}
+	}
+
 	r.mu.Lock()
 	hosts := r.hostToIP
 	localDomains := r.localDomains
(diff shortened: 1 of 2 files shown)


net/dns/resolver/tsdns.go
```diff
@@ -644,6 +644,14 @@ func (r *Resolver) resolveLocalReverse(name dnsname.FQDN) (dnsname.FQDN, dns.RCo
 		return "", dns.RCodeRefused
 	}
 
+	// If someone curiously does a reverse lookup on the DNS IP, we
+	// return a domain that helps indicate that Tailscale is using
+	// this IP for a special purpose and it is not a node on their
+	// tailnet.
+	if ip == tsaddr.TailscaleServiceIP() || ip == tsaddr.TailscaleServiceIPv6() {
+		return "nodedns.tailscale-internals.", dns.RCodeSuccess
```
bradfitz, line 652: Maybe `magicdns.localhost-tailscale-daemon.`? Key bits:

* MagicDNS is what we call the feature
* it "tailscale" in there
* the "localhost" tells them to not be concerned about unencrypted DNS traffic because it didn't leave the machine

bradfitz:
Put the `Fixes #nnn` in the commit message. (PR too is fine, but the commit message is the history of record)

Also, the title should be "net/dns/resolver: ..."

In general we follow the Go style for commit messages: https://github.com/golang/go/wiki/CommitMessage

net/dns/resolver/tsdns.go
twitchyliquid64: SGTM!

twitchyliquid64:
Done, thanks!

Out of curiosity, what do you do if your change touches multiple packages? Or is it intended that you split a change into one commit per package?

net/dns/resolver/tsdns.go
twitchyliquid64: Yeppers, done!

bradfitz (approved):

danderson (approved):
The name's a bit of a mouthful, but that's probably actually a good thing for a name that's purely informational and never intended to be typed into anything by humans. LGTM, thanks!

PR #4364, opened 2022-04-06
net/dns, net/dns/resolver, wgengine: refactor DNS request path

twitchyliquid64:
Previously, method calls into the DNS manager/resolver types handled DNS requests rather than DNS packets. This is fine for UDP as one packet corresponds to one request or response, however will not suit an implementation that supports DNS over TCP.

To support PRs implementing this in the future, wgengine delegates all handling/construction of packets to the magic DNS endpoint, to the DNS types themselves. Handling IP packets at this level enables future support for both UDP and TCP.

### DIFF
--- a/net/dns/resolver/tsdns_test.go
+++ b/net/dns/resolver/tsdns_test.go
@@ -39,7 +39,7 @@ var (
 	testipv4Arpa = dnsname.FQDN("4.3.2.1.in-addr.arpa.")
 	testipv6Arpa = dnsname.FQDN("f.0.e.0.d.0.c.0.b.0.a.0.9.0.8.0.7.0.6.0.5.0.4.0.3.0.2.0.1.0.0.0.ip6.arpa.")
 
-	magicdnsDstUdp = netaddr.MustParseIPPort("100.100.100.100:53")
+	magicDNSv4Port = netaddr.MustParseIPPort("100.100.100.100:53")
 )
 
 var dnsCfg = Config{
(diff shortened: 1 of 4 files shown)


net/dns/resolver/tsdns_test.go
```diff
@@ -37,6 +38,8 @@ var (
 
 	testipv4Arpa = dnsname.FQDN("4.3.2.1.in-addr.arpa.")
 	testipv6Arpa = dnsname.FQDN("f.0.e.0.d.0.c.0.b.0.a.0.9.0.8.0.7.0.6.0.5.0.4.0.3.0.2.0.1.0.0.0.ip6.arpa.")
+
+	magicdnsDstUdp = netaddr.MustParseIPPort("100.100.100.100:53")
```
bradfitz, line 42: magicDNSv4Port ?

https://github.com/golang/go/wiki/CodeReviewComments/#initialisms

wgengine/userspace.go
```diff
@@ -512,38 +512,9 @@ func (e *userspaceEngine) pollResolver() {
...
-			h := packet.UDP6Header{
-				IP6Header: packet.IP6Header{
-					Src: magicDNSIPv6,
-					Dst: to.IP(),
-				},
-				SrcPort: magicDNSPort,
-				DstPort: to.Port(),
-			}
-			hlen := h.Len()
-			// TODO(dmytro): avoid this allocation without importing tstun quirks into dns.
```
bradfitz, line 542: I guess we're just accepting this allocation at this point? :)

bradfitz: (I mean, it's DNS traffic and not super high volume so that's probably fine)

twitchyliquid64: Given its a straightforward change, I've updated the code to read the offset and allocate enough space originally. I had a quick look through `tstun`, it always gets imported anyway so I don't see any downside to reading the const straight from the package.

danderson (approved):
Something to consider for future changes to the MagicDNS endpoint: back in the day, we had to build things this way because we couldn't bundle netstack into all clients (because of ios memory limits). Now that we can (and are, I think?) bundling netstack everywhere, should quad-100 just be served purely through netstack, so we can handle all this request processing as boring TCP and UDP sockets?

Not required for this change, obviously, just planting seeds for the future.

bradfitz:
How much refactoring do we want to do right before 1.24.0?

PR #4366, opened 2022-04-07
api: update acl/validate data format

ramyfication:
Expected input is JSON formatted []policy.ACLTest


bradfitz:
How does it validate things then if you don't send the rest of the ACL rules?

ramyfication:
It runs validation tests against the tailnet's active ACL

bradfitz (approved):

PR #4369, opened 2022-04-07
hostinfo, tailcfg: add desktop environment to hostinfo

twitchyliquid64:
From the machines tab its hard to differenciate desktop Linux installs from server Linux installs. Transmitting this information should make this determination a lot easier.

Due to the reality that tailscaled is likely a system process, the standard checks based on XDG_SESSION_TYPE or DISPLAY environment variables are not possible (those variables won't be set). Instead, we look for listening unix sockets that are typical of desktop installs.

### DIFF
--- a/hostinfo/hostinfo.go
+++ b/hostinfo/hostinfo.go
@@ -17,10 +17,13 @@ import (
 
 	"go4.org/mem"
 	"tailscale.com/tailcfg"
+	"tailscale.com/types/opt"
 	"tailscale.com/util/dnsname"
 	"tailscale.com/util/lineread"
 	"tailscale.com/version"
 )
+
+var started = time.Now()
 
 // New returns a partially populated Hostinfo for the current host.
 func New() *tailcfg.Hostinfo {
@@ -31,6 +34,7 @@ func New() *tailcfg.Hostinfo {
 		Hostname:    hostname,
 		OS:          version.OS(),
 		OSVersion:   GetOSVersion(),
+		Desktop:     desktop(),
 		Package:     packageTypeCached(),
 		GoArch:      runtime.GOARCH,
 		DeviceModel: deviceModel(),
@@ -97,6 +101,7 @@ func GetEnvType() EnvType {
 var (
 	deviceModelAtomic atomic.Value // of string
 	osVersionAtomic   atomic.Value // of string
+	desktopAtomic     atomic.Value // of opt.Bool
 	packagingType     atomic.Value // of string
 )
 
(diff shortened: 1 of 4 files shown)


twitchyliquid64:
I'll also note that I had to manually update the generated file due to https://github.com/golang/go/issues/51629

hostinfo/hostinfo_linux.go
```diff
@@ -26,6 +26,48 @@ func init() {
...
+	}
+	if seenWayland {
+		return "wayland"
+	}
+	if seenX {
+		return "X11"
+	}
+	if seenDbus {
+		return "user-dbus"
+	}
```
bradfitz, line 69: I thought we were just going to do a bool, to answer the question of whether it's a Linux server or a Linux desktop.

How important is Sway vs Wayland vs X11? That's a pretty esoteric detail. Why not GNOME vs KDE vs XFCE at that point, for instance?

If we do a bool, it should be a tristate (opt.Bool), though, so we can distinguish old clients that didn't check.

twitchyliquid64: My original thinking was that X11 vs Wayland would be useful information, if we ever wanted to make a GUI or tray client on Linux. Regardless, I've changed this to be an `opt.Bool`.

bradfitz (approved):

twitchyliquid64:
Changed logic to only cache the desktop/not-desktop determination after a minute - based on slack convo.

hostinfo/hostinfo.go
```diff
@@ -117,6 +122,31 @@ func deviceModel() string {
...
+	lineread.File("/proc/net/unix", func(line []byte) error {
+		seenDesktop = seenDesktop || strings.Contains(string(line), " @/tmp/dbus-")
+		seenDesktop = seenDesktop || strings.Contains(string(line), ".X11-unix")
+		seenDesktop = seenDesktop || strings.Contains(string(line), "/wayland-1")
+		return nil
+	})
+	ret.Set(seenDesktop)
+
+	// Only cache after a minute - compositors might not have started yet.
+	if time.Now().After(started.Add(time.Minute)) {
```
bradfitz, line 144: if time.Since(started) > time.Minute {

bradfitz (approved):
LGTM otherwise

hostinfo/hostinfo.go
```diff
@@ -117,6 +122,31 @@ func deviceModel() string {
...
+	if v := desktopAtomic.Load(); v != nil {
+		v, _ := v.(opt.Bool)
+		return v
+	}
+
+	seenDesktop := false
+	lineread.File("/proc/net/unix", func(line []byte) error {
+		seenDesktop = seenDesktop || strings.Contains(string(line), " @/tmp/dbus-")
+		seenDesktop = seenDesktop || strings.Contains(string(line), ".X11-unix")
+		seenDesktop = seenDesktop || strings.Contains(string(line), "/wayland-1")
```
bradfitz, line 138: mem.Contains(mem.B(line), mem.S("/wayland-1"))

bradfitz:
LGTM

PR #4370, opened 2022-04-07
control/controlbase: make the protocol version number selectable.

danderson:
This is so that we can plumb our client capability version through
the protocol as the Noise version. The capability version increments
more frequently than strictly required (the Noise version only needs
to change when cryptographically-significant changes are made to
the protocol, whereas the capability version also indicates changes
in non-cryptographically-significant parts of the protocol), but this
gives us a safe pre-auth way to determine if the client supports
future protocol features, while still relying on Noise's strong
assurance that the client and server have agreed on the same version.

Currently, the server executes the same protocol regardless of the
version number, and just presents the version to the caller so they
can do capability-based things in the upper RPC protocol. In future,
we may add a ratchet to disallow obsolete protocols, or vary the
Noise handshake behavior based on requested version.

### DIFF
--- a/control/controlclient/noise.go
+++ b/control/controlclient/noise.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"math"
 	"net"
 	"net/http"
 	"net/url"
@@ -17,6 +18,7 @@ import (
 	"golang.org/x/net/http2"
 	"tailscale.com/control/controlbase"
 	"tailscale.com/control/controlhttp"
+	"tailscale.com/tailcfg"
 	"tailscale.com/types/key"
 	"tailscale.com/util/multierr"
 )
@@ -146,7 +148,12 @@ func (nc *noiseClient) dial(_, _ string, _ *tls.Config) (net.Conn, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 
-	conn, err := controlhttp.Dial(ctx, nc.serverHost, nc.privKey, nc.serverPubKey)
+	if tailcfg.CurrentCapabilityVersion > math.MaxUint16 {
+		// Panic, because a test should have started failing several
+		// thousand version numbers before getting to this point.
+		panic("capability version is too high to fit in the wire protocol")
+	}
+	conn, err := controlhttp.Dial(ctx, nc.serverHost, nc.privKey, nc.serverPubKey, uint16(tailcfg.CurrentCapabilityVersion))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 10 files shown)


tailcfg/tailcfg.go
```diff
@@ -34,7 +34,7 @@ import (
 // Previously (prior to 2022-03-06), it was known as the "MapRequest
 // version" or "mapVer" or "map cap" and that name and usage persists
 // in places.
-type CapabilityVersion int
+type CapabilityVersion uint16
```
bradfitz, line 37: I don't think this will be worth the pain.

Just explode elsewhere if this ever goes over uint16 in size.  (or write some test that fails if it goes over 64k)

bradfitz (approved):
LGTM otherwise

tailcfg/tailcfg.go
danderson: Switched tailcfg type back to int. I kept uint16 in controlbase and controlhttp, you okay with that?

Added a test that will fail when tailcfg version exceeds 60k, as an early warning to revise our wire protocol. Also added a hard panic() in controlclient if we ever try to start a session with a >uint16 version. Hard panic because the test should give us very early warning that the panic is coming, so something's very wrong if that codepath trips.

control/controlclient/noise.go
```diff
@@ -146,7 +147,12 @@ func (nc *noiseClient) dial(_, _ string, _ *tls.Config) (net.Conn, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 
-	conn, err := controlhttp.Dial(ctx, nc.serverHost, nc.privKey, nc.serverPubKey)
+	if tailcfg.CurrentCapabilityVersion > 65535 {
```
bradfitz, line 150: math.MaxUint16 ?

control/controlclient/noise_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+// maxAllowedNoiseVersion is the highest we expect the Tailscale
+// capability version to ever get. It's a value close to 2^16, but
+// with enough leeway that we get a very early warning that it's time
+// to rework the wire protocol to allow larger versions, while still
+// giving us headroom to bump this test and fix the build.
+//
+// Code elsewhere in the client will panic() if the tailcfg capability
+// version exceeds 16 bits, so take a failure of this test seriously.
+const maxAllowedNoiseVersion = 60000
```
bradfitz, line 17: math.MaxUint16 - 5000 ?  :)

bradfitz (approved):

control/controlclient/noise.go
danderson: Bah, I never remember those magic constants. Fixed.

PR #4372, opened 2022-04-07
cmd/mkpkg: add name argument

Xe:
I would like to package proxy-to-grafana and similar tools in .deb/.rpm packages. As a first step to that, I want to be able to pass a package name and description to the `mkpkg` tool.

Also included is a patch to `shell.nix` to make it track recent changes in how `goimports` is referenced.

### DIFF
--- a/cmd/mkpkg/main.go
+++ b/cmd/mkpkg/main.go
@@ -45,6 +45,8 @@ func parseEmptyDirs(s string) []string {
 
 func main() {
 	out := getopt.StringLong("out", 'o', "", "output file to write")
+	name := getopt.StringLong("name", 'n', "tailscale", "package name")
+	description := getopt.StringLong("description", 'd', "The easiest, most secure, cross platform way to use WireGuard + oauth2 + 2FA/SSO", "package description")
 	goarch := getopt.StringLong("arch", 'a', "amd64", "GOARCH this package is for")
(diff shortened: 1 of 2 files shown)


cmd/mkpkg/main.go
```diff
@@ -45,6 +45,8 @@ func parseEmptyDirs(s string) []string {
 
 func main() {
 	out := getopt.StringLong("out", 'o', "", "output file to write")
+	name := getopt.StringLong("name", 'n', "tailscale", "package name")
```
bradfitz, line 48: Please don't use getopt. I thought we'd removed it everywhere. This might be the only last user in any of our trees?

bradfitz (approved):

cmd/mkpkg/main.go
Xe: Which package should I replace it with? I can make a followup PR to do that.

danderson: This program's needs are very simple, so just the stdlib `flag` package.

For commands that need complex subcommands and stuff, atm we use ffcli, but I don't think you should use it for this program.

I agree that should be a followup change though.

bradfitz: the standard library one we use everywhere else

danderson (approved):

PR #4374, opened 2022-04-07
tstest/integration/vms: fix docs, qemu-img invocation

twitchyliquid64:


crawshaw (approved):

tstest/integration/vms/README.md
```diff
@@ -26,19 +26,16 @@ This test also requires the following:
...
   `kvm` group should suffice)
 
-This optionally requires an AWS profile to be configured at the [default
-path](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html).
-The S3 bucket is set so that the requester pays. Please keep this in mind when
-running these tests on your machine. If you are uncomfortable with the cost from
-downloading from S3, you should pass the `-no-s3` flag to disable downloads from
-S3. However keep in mind that some distributions do not use stable URLs for each
-individual image artifact, so there may be spurious test failures as a result.
+The `-no-s3` flag is needed to disable downloads from S3, which require
```
bradfitz, line 29: even if Go supports both styles, we tend write two hyphens in our docs to not weird people out who don't know non-GNU styles.

bradfitz (approved):

tstest/integration/vms/README.md
twitchyliquid64: Ah whelp looks like i forgot to `-f` so this didnt make it in. I'll do a follow-up.

PR #4375, opened 2022-04-07
shell.nix: use tailscale-go for compilation

### LINKED ISSUE #4231: ./build_dist.sh & friends (tools/go) do not work out of the box on NixOS
### What is the issue?

The installed `tailscale-go` does not have its dynamic-linker paths adjusted for a destination NixOS system.

A mildly-terrifying (but would work until you update glibc) fix is:

```diff
diff --git a/tool/go b/tool/go
index 8f7485f9..96c68d7a 100755
--- a/tool/go
+++ b/tool/go
@@ -74,6 +74,15 @@ get_cached() {
     mkdir -p "$TOOLCHAIN"
     (cd "$TOOLCHAIN" && tar --strip-components=1 -xf "$archive")
     echo "$REV" >$mark
+
+    # NixOS installations do not have glibc in the default location on the filesystem,
+    # however tailscale-go expects it. As such we patch the dynamic linker paths to
+    # Make this work.
+    if [[ $(uname -v) == *"NixOS"* ]]; then
+      PATCHELF=$(nix eval --raw nixpkgs.patchelf)/bin/patchelf
+      $PATHELF --set-rpath $(nix eval --raw nixpkgs.glibc)/lib $TOOLCHAIN_GO
+      $PATHELF --set-interpreter $(nix eval --raw nixpkgs.glibc)/lib/ld-linux-x86-64.so.2 $TOOLCHAIN_GO
+    fi
 }

 if [ "${REV}" = "SKIP" ] ||
```

### Steps to reproduce

1. Be on NixOS or a derivative
2. `git clone https://github.com/tailscale/tailscale && cd tailscale`
3. `nix-shell`
4. `./build_dist.sh tailscale.com/cmd/tailscale`

You'll get the error:

```
bash: /home/xxx/.cache/tailscale-go/bin/go: No such file or directory
```

_No response_

### OS

Linux, Other

### OS version

Linux xxx 5.10.106 #1-NixOS SMP Wed Mar 16 13:16:03 UTC 2022 x86_64 GNU/Linux

### Tailscale version

main

### Bug report

_No response_

twitchyliquid64:
This change builds a derivation for tailscale-go and makes it available in the
users development environment. This is consistent with the shell.nix in corp/.

Once go1.18 is in a stable Nixpkgs release we can avoid relying on derivations
from nixpkgs head. For now, this works well, and the fetched derivations are
cached in the Nix store according to the usual rules.

Fixes #4231

Signed-off-by: Tom DNetto <tom@tailscale.com>

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,3 +1,4 @@
+# TODO(tom): Use system nixpkgs exclusively once 1.18 is in a stable release.
 # This is a shell.nix file used to describe the environment that tailscale needs
 # for development. This includes a lot of the basic tools that you need in order
 # to get started. We hope this file will be useful for users of Nix on macOS or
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -9,16 +10,42 @@
...
-  #  - gotools for goimports, a robust formatting tool for Go source code
-  #  - gopls, the language server for Go to increase editor integration
-  #  - git, the version control program (used in some scripts)
-  buildInputs = with pkgs; [
-    go gotools gopls git
-  ];
-}
+{
+	pkgs ? import <nixpkgs> {},
+	nixosUnstable ? import (fetchTarball https://github.com/NixOS/nixpkgs/archive/refs/heads/master.tar.gz) { },
```
danderson, line 16: use `nixpkgs-unstable.tar.gz` rather than `master.tar.gz`. The master branch only passes basic CI checks, whereas the release branches only get bumped after Hydra pre-builds all the packages and runs a more extensive regression suite.

danderson (approved):

PR #4378, opened 2022-04-07
wgengine/netstack: disable refsvfs2 leak tracking

### LINKED ISSUE #4309: Leak checking is partially enabled in netstack
Netstack does not explicitly initialize its leak checker, and the 0 value performs some leak checking activities:

https://github.com/google/gvisor/blob/34623f4d753169d8af72a75c8e72bbd27e64c01b/pkg/refsvfs2/refs_map.go#L52

https://github.com/google/gvisor/blob/34623f4d7531/pkg/refs/refcounter.go#L222

The leak checker may be useful, so I think I'd like to drive this through a flag, config or envknob, however there's also a need with the leak checker to actively drive it to make the most use of it, see: https://github.com/google/gvisor/blob/34623f4d753169d8af72a75c8e72bbd27e64c01b/pkg/refsvfs2/refs_map.go#L126

I think we may also want to enable panic leak checks in tests, if we can do so without adding flakes.

raggi:
In addition an envknob (TS_DEBUG_NETSTACK_LEAK_MODE) now provides access
to set leak tracking to more useful values.

Fixes #4309


cmd/tailscaled/depaware.txt
```diff
@@ -130,7 +130,7 @@ tailscale.com/cmd/tailscaled dependencies: (generated by github.com/tailscale/de
         gvisor.dev/gvisor/pkg/linewriter                             from gvisor.dev/gvisor/pkg/log
         gvisor.dev/gvisor/pkg/log                                    from gvisor.dev/gvisor/pkg/context+
         gvisor.dev/gvisor/pkg/rand                                   from gvisor.dev/gvisor/pkg/tcpip/network/hash+
-        gvisor.dev/gvisor/pkg/refs                                   from gvisor.dev/gvisor/pkg/refsvfs2
+        gvisor.dev/gvisor/pkg/refs                                   from gvisor.dev/gvisor/pkg/refsvfs2+
```
raggi, line 133: I'm not entirely sure what this change in syntax on the right hand side means, but I am aware of pre-existing dependencies between these packages. Perhaps this is triggered by actually exercising them in our init paths?

bradfitz (approved):

Issue #4144, opened 2022-03-11
gvisor: panic: Incrementing non-positive count 0xc000533180 on stack.PacketBuffer


DentonGentry:
### What is the issue?

From email report:

we compiled 1.22.1 and are experiencing problems on two of our macOS machines. tailscaled crashes directly after startup.

The ports 6883, 6885 are used by a local program which is running in the background before tailscaled is started.

We do see this error on our 2 Intel macOS systems. Arm macOS does not show the error.

We had a similar problem with 1.22.0. Downgrading on one of the intel machines to 1.20.4 solved the problem.

```
2022/03/10 15:08:59 logtail started
2022/03/10 15:08:59 Program starting: v1.22.1-tc8fb4f8c7, Go 1.17.8-tsdce70b6: []string{"/usr/local/bin/tailscaled", "-state=/var/db/tailscale/tailscaled.state", "-tun", "utun9"}
2022/03/10 15:08:59 LogID: 459c4a8ac464c58028859fb0d0bb79ee88adf8db6a27e083c7a7d63242103f9e
2022/03/10 15:08:59 logpolicy: using system state directory "/Library/Tailscale"
2022/03/10 15:08:59 wgengine.NewUserspaceEngine(tun "utun9") ...
2022/03/10 15:08:59 dns: using dns.noopManager
2022/03/10 15:08:59 link state: interfaces.State{defaultRoute=en13 ifs={en13:[10.0.5.6/24]} v4=true v6=false}
2022/03/10 15:08:59 magicsock: disco key = d:f376519d5ab5ccd3
2022/03/10 15:08:59 Creating wireguard device...
2022/03/10 15:08:59 Bringing wireguard device up...
2022/03/10 15:08:59 Bringing router up...
2022/03/10 15:08:59 external route: up
2022/03/10 15:08:59 Clearing router settings...
2022/03/10 15:08:59 Starting link monitor...
2022/03/10 15:08:59 Engine created.
2022/03/10 15:09:01 Start
2022/03/10 15:09:01 using backend prefs for "_daemon": Prefs{ra=true dns=true want=true routes=[0.0.0.0/0 ::/0 10.0.5.0/24] snat=true Persist{lm=, o=, n=[jR0DB] u="REDACTED"}}
2022/03/10 15:09:01 Backend: logs: be:REDACTED fe:
2022/03/10 15:09:01 control: client.Login(false, 0)
2022/03/10 15:09:01 control: doLogin(regen=false, hasUrl=false)
[...]

PR #4379, opened 2022-04-08
go.mod: bump netstack for clone reset fix

raggi:
In tracking down issue #4144 and reading through the netstack code in
detail, I discovered that the packet buf Clone path did not reset the
packetbuf it was getting from the sync.Pool. The fix was sent upstream
https://github.com/google/gvisor/pull/7385, and this bump pulls that in.
At this time there is no known path that this fixes, however at the time
of upstream submission this reset at least one field that could lead to
incorrect packet routing if exercised, a situation that could therefore
lead to an information leak.


crawshaw (approved):

bradfitz:
Please reference a bug or so more.

bradfitz:
Also "go.mod: bump netstack ..."

raggi:
description added, sorry for not doing so first time :)

bradfitz:
Put a # before 4144 so they're crosslinked

raggi:
fwiw the two issues are only linked by the fact that I found the nestack bug while scouring the code for #4144 - this does not relate to it directly.

bradfitz:
LGTM

PR #4385, opened 2022-04-08
wgengine/monitor: ignore duplicate RTM_NEWADDRs

raggi:
Two commits, the first contains the introduction of an envknob for live inspection of this path. Using this envknob I discovered several follow-up issues described below and included in the change in the second commit.

--

Ignoring the events at this layer is the simpler path for right now, a
broader change should follow to suppress irrelevant change events in a
higher layer so as to avoid related problems with other monitoring paths
on other platforms.  This approach may also carry a small risk that it
applies an at-most-once invariant low in the chain that could be assumed
otherwise higher in the code.

I adjusted the newAddrMessage type to include interface index rather
than a label, as labels are not always supplied, and in particular on my
test hosts they were consistently missing for ipv6 address messages.

I adjusted the newAddrMessage.Addr field to be populated from
Attributes.Address rather than Attributes.Local, as again for ipv6
.Local was always empty, and with ipv4 the .Address and .Local contained
the same contents in each of my test environments.

### DIFF
--- a/wgengine/monitor/monitor_linux.go
+++ b/wgengine/monitor/monitor_linux.go
@@ -15,9 +15,12 @@ import (
 	"github.com/mdlayher/netlink"
 	"golang.org/x/sys/unix"
 	"inet.af/netaddr"
+	"tailscale.com/envknob"
 	"tailscale.com/net/tsaddr"
 	"tailscale.com/types/logger"
 )
+
+var debugNetlinkMessages = envknob.Bool("TS_DEBUG_NETLINK")
 
 // unspecifiedMessage is a minimal message implementation that should not
 // be ignored. In general, OS-specific implementations should use better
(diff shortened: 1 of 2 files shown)


wgengine/monitor/monitor_linux.go
```diff
@@ -35,6 +35,12 @@ type nlConn struct {
 	logf     logger.Logf
 	conn     *netlink.Conn
 	buffered []netlink.Message
+
+	// addrCache is used to suppress duplicate RTM_NEWADDR messages, it is
+	// populated by RTM_NEWADDR messages and de-populated by RTM_DELADDR.
+	// Address change notifications will only passed up to the callees if the
+	// address is not a member of this set. See #4282.
+	addrCache map[string]map[netaddr.IP]struct{}
```
bradfitz, line 43: whenever you have non-obvious map keys like strings, document what the key is in a comment

also, I'm kinda anti-struct{} for sets. I just use bool true. It makes set membership tests so much cleaner to write.

raggi: 👍

raggi:
staticcheck seems to be bugged.

raggi:
> staticcheck is having a good friday drink.

whoops, no, it's me who needs a good friday drink.

bradfitz (approved):

PR #4396, opened 2022-04-11
tstest/integration/vms: test on stable nixos (21.11)

twitchyliquid64:
I would like to do some more customized integration tests in the future, (specifically, bringing up a mitm proxy and testing tailscaled through that) so hoping to bring back the nixos wiring to support that.


Xe (approved):
LGTM! Happy to see this come back!

Feel free to delete most of the distros in the list. Most of those links 404.

Issue #4329, opened 2022-03-30
wgengine/router: check for required kernel modules


bradfitz:
Fail nicer on Linux if iptables kernel modules are missing.

e.g. `xt_mark` in:
https://github.com/tailscale/tailscale/issues/4145#issuecomment-1082662468

we can probably just exec modprobe for each and not even worry about parsing `/proc/modules` and checking which are built-in, etc.

We might need to require one of two passing sets (for legacy vs nftables)

/cc @danderson @raggi

raggi:
I should know this, but I don't remember the semantics of `modprobe` in various container environments.

PR #4397, opened 2022-04-12
wgengine/router,util/kmod: load & log xt_mark

raggi:
Attempt to load the xt_mark kernel module when it is not present. If the
load fails, log error information.

It may be tempting to promote this failure to an error once it has been
in use for some time, so as to avoid reaching an error with the iptables
invocation, however, there are conditions under which the two stages may
disagree - this change adds more useful breadcrumbs.

Example new output from tailscaled running under my WSL2:

```
router: ensure module xt_mark: "/usr/sbin/modprobe xt_mark" failed: exit status 1; modprobe: FATAL: Module xt_mark not found in directory /lib/modules/5.10.43.3-microsoft-standard-WSL2
```

Background:

There are two places to lookup modules, one is `/proc/modules` "old",
the other is `/sys/module/` "new".

There was query_modules(2) in linux <2.6, alas, it is gone.

In a docker container in the default configuration, you would get
/proc/modules and /sys/module/ both populated. lsmod may work file,
modprobe will fail with EPERM at `finit_module()` for an unpriviliged
container.

In a priviliged container the load may *succeed*, if some conditions are
met. This condition should be avoided, but the code landing in this
change does not attempt to avoid this scenario as it is both difficult
to detect, and has a very uncertain impact.

In an nspawn container `/proc/modules` is populated, but `/sys/module`
does not exist. Modern `lsmod` versions will fail to gather most module
information, without sysfs being populated with module information.

In WSL2 modules are likely missing, as the in-use kernel typically is
not provided by the distribution filesystem, and WSL does not mount in a
module filesystem of its own. Notably the WSL2 kernel supports iptables
marks without listing the xt_mark module in /sys/module, and
/proc/modules is empty.

On a recent kernel, we can ask the capabilities system about SYS_MODULE,
that will help to disambiguate between the non-privileged container case
and just being root. On older kernels these calls may fail.

Update #4329


util/kmod/kmod.go
```diff
@@ -0,0 +1,157 @@
...
+	prefix := make([]byte, len(name)+1)
+	copy(prefix, name)
+	prefix[len(prefix)-1] = ' '
+	scanner := bufio.NewScanner(procMod)
+	for scanner.Scan() {
+		if bytes.HasPrefix(scanner.Bytes(), prefix) {
+			return true, nil
+		}
+	}
+	return false, nil
```
bradfitz, line 54: maybe use lineread.File instead? (from our util/lineread package)

util/kmod/kmod.go
```diff
@@ -0,0 +1,157 @@
...
+	if _, err := os.Stat(filepath.Join("/sys/module", name)); err == nil {
+		return true, nil
+	}
+	procMod, err := os.Open("/proc/modules")
+	if err != nil {
+		return false, fmt.Errorf("module %s not found in /sys/module or /proc/modules: %w", name, err)
+	}
+	defer procMod.Close()
+
+	prefix := make([]byte, len(name)+1)
```
bradfitz, line 45: `prefix := mem.S(name + " ")` is one alloc (for the string concat, not the `mem.S`)

and then below:

`if mem.HasPrefix(mem.B(line), prefix) { .... }`

bradfitz: (that's using `go4.org/mem` which we use pretty often in the tree)

util/kmod/kmod.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// could not determine a well known result from capabilities, make an
+	// assumption based on uid.
+	if os.Getuid() == 0 {
+		return true, nil
+	}
+	return false, fmt.Errorf("not running as root, and unable to check kernel module capabilities")
+}
+
+// findExecutable checks paths for a path that exists and is executable by the current user.
```
bradfitz, line 84: sounds too close to os.Executable.  maybe call it `firstExecutable`?

util/kmod/kmod.go
```diff
@@ -0,0 +1,157 @@
...
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		err = fmt.Errorf("%q failed: %w; %s", fmt.Sprintf("%s %s", modprobePath, name), err, bytes.TrimSpace(out))
+	}
+	return err
+}
+
+// tryInstallModule attempts to find a modprobe binary to run either in
+// well-known paths, or in $PATH, and runs it. If it can not find a modprobe to
+// run, it instead falls back to a syscall interface to attemtp to install a
```
bradfitz, line 106: typo: attempt

util/kmod/kmod.go
```diff
@@ -0,0 +1,157 @@
...
+	// could not determine a well known result from capabilities, make an
+	// assumption based on uid.
+	if os.Getuid() == 0 {
+		return true, nil
+	}
+	return false, fmt.Errorf("not running as root, and unable to check kernel module capabilities")
+}
+
+// findExecutable checks paths for a path that exists and is executable by the current user.
+func findExecutable(paths []string) string {
```
bradfitz, line 85: make this `paths ...string` to clean up call site?

util/kmod/kmod.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		err = fmt.Errorf("unable to find modprobe(1), and load of module %s failed with: %w", name, err)
+	}
+	return err
+}
+
+// EnsureModule attempts to ensure that the given module is installed, returning
+// true only if it has been found or successfully installed, otherwise false is
+// returned along with a list of informational errors about probe attempts.
+func EnsureModule(name string) (bool, []error) {
```
bradfitz, line 128: return just an error and use `tailscale.com/util/multierr` instead?

wgengine/router/router_linux.go
```diff
@@ -175,6 +176,14 @@ func newUserspaceRouterAdvanced(logf logger.Logf, tunname string, linkMon *monit
 		}
 	}
 
+	if !envknob.Bool("TS_DISABLE_MODPROBE") {
```
bradfitz, line 179: generally we put these envknob calls as package globals for the side effect of them registering their use, which gets logged/reported.

bradfitz (approved):

util/kmod/kmod.go
raggi: Neat, nice package, thanks.

PR #4400, opened 2022-04-12
cmd/nginx-auth: create new Tailscale NGINX auth service

Xe:
This conforms to the NGINX subrequest result authentication protocol[1] using the NGINX module `ngx_http_auth_request_module`. This is based on the example that @peterkeen provided on Twitter[2], but with several changes to make things more tightly locked down:

* This listens over a UNIX socket instead of a TCP socket to prevent leakage to the network
* This uses systemd socket activation so that systemd owns the socket and can then lock down the service to the bare minimum required to do its job without having to worry about dropping permissions
* This provides additional information in HTTP response headers that can be useful for integrating with various services
* This has a script to automagically create debian and redhat packages for easier distribution

This will be written about on the Tailscale blog. There is more information in README.md.

[1]: https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-subrequest-authentication/
[2]: https://github.com/peterkeen/tailscale/blob/main/cmd/nginx-auth-proxy/nginx-auth-proxy.go

### DIFF
--- a/cmd/nginx-auth/README.md
+++ b/cmd/nginx-auth/README.md
@@ -1,0 +1,135 @@
+# nginx-auth
+
+This is a tool that allows users to use Tailscale Whois authentication with
+NGINX as a reverse proxy. This allows users that already have a bunch of
+services hosted on an internal NGINX server to point those domains to the
+Tailscale IP of the NGINX server and then seamlessly use Tailscale for
+authentication.
+
+Many thanks to [@zrail](https://twitter.com/zrail/status/1511788463586222087) on
+Twitter for introducing the basic idea and offering some sample code. This
+program is based on that sample code with security enhancements. Namely:
+
+* This listens over a UNIX socket instead of a TCP socket, to prevent
+  leakage to the network
+* This uses systemd socket activation so that systemd owns the socket
+  and can then lock down the service to the bare minimum required to do
+  its job without having to worry about dropping permissions
+* This provides additional information in HTTP response headers that can
+  be useful for integrating with various services
+
+## Configuration
+
+In order to protect a service with this tool, do the following in the respective
+`server` block:
+
+Create an authentication location with the `internal` flag set:
+
+```nginx
+location /auth {
+  internal;
+
+  proxy_pass http://unix:/run/tailscale.nginx-auth.sock;
+  proxy_pass_request_body off;
+
+  proxy_set_header Host $http_host;
+  proxy_set_header Remote-Addr $remote_addr;
+  proxy_set_header Remote-Port $remote_port;
+  proxy_set_header Original-URI $request_uri;
+}
+```
+
+Then add the following to the `location /` block:
+
+```
+auth_request /auth;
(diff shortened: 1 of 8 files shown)


cmd/nginx-auth/nginx-auth.go
```diff
@@ -0,0 +1,116 @@
...
+			log.Printf("can't extract tailnet name from hostname %q", info.Node.Name)
+			return
+		}
+		tailnet, _, ok = strings.Cut(tailnet, ".beta.tailscale.net")
+		if !ok {
+			w.WriteHeader(http.StatusUnauthorized)
+			log.Printf("can't extract tailnet name from hostname %q", info.Node.Name)
+			return
+		}
+
```
maisem, line 77: we really should add this to the `WhoIsResponse`

cmd/nginx-auth/nginx-auth.go
```diff
@@ -0,0 +1,116 @@
...
+		go func(ln net.Listener) {
+			log.Printf("listening on %s", ln.Addr())
+			log.Fatal(http.Serve(ln, mux))
+		}(ln)
+	}
+
+	for {
+		select {}
+	}
```
maisem, line 115: maybe add a sync.WaitGroup and wait on that?

something like

```
var wg sync.WaitGroup
wg.Add(len(listeners))
for _, ln := range listeners {
    go func(ln net.Listener) {
        defer wg.Done()
        log.Printf("listening on %s", ln.Addr())
        http.Serve(ln, mux)
    }(ln)
}
wg.Wait()
```

cmd/nginx-auth/nginx-auth.go
Xe: We should, but I imagine that we will need to keep this hack around for a while regardless.

cmd/nginx-auth/nginx-auth.go
Xe: I kinda want this to explode horribly if any of the listeners exits though. Do you have any suggestions for this?

maisem: You can keep your original code then and add a comment that we want it to explode and state the reason why

cmd/nginx-auth/README.md
```diff
@@ -0,0 +1,134 @@
...
+Most of the time you can set `X-Webauth-User` to the contents of the
+`Tailscale-User` header, but some services may not accept a username with an `@`
+symbol in it. If this is the case, set `X-Webauth-User` to the `Tailscale-Login`
+header.
+
+The `Tailscale-Tailnet` header can help you identify which tailnet the session
+is coming from. If you are using node sharing, this can help you make sure that
+you aren't giving administrative access to people outside your tailnet. You will
+need to be sure to check this in your application code. If you use OpenResty,
+you may be able to do more complicated access controls.
```
maisem, line 122: If you use OpenResty, you may be able to add more complex access controls than those provided by NGINX?

maisem (approved):

PR #4403, opened 2022-04-13
bump x/tools for go/packages fix

bradfitz:


.github/workflows/go_generate.yml
```diff
@@ -25,13 +25,14 @@ jobs:
...
-        # go generate --tags=hermetic ./...
         run: |
-          set -e
-          ./.github/workflows/go-generate-without-stringer.sh
+          if [[ "${{github.ref}}" == release-branch/* ]]
+          then
+            pkgs=$(go list ./... | grep -v dnsfallback)
+          else
+            pkgs=$(go list ./... | grep -v dnsfallback)
+          fi
```
maisem, line 34: are these two identical?

maisem (approved):

.github/workflows/go_generate.yml
bradfitz: yes, it's how it was before. It's so we can make them different easily in the future I believe.

PR #4405, opened 2022-04-13
net/dns, util/publicdns: extract public DNS mapping into own package

phirework:
This extracts DOH mapping of known public DNS providers in
`net/dns/resolverforwarder.go` into its own package, to be consumed by other repos


util/publicdns/publicdns.go
```diff
@@ -0,0 +1,75 @@
...
+func DohV6(base string) (ip netaddr.IP, ok bool) {
+	for _, ip := range DohIPsOfBase[base] {
+		if ip.Is6() {
+			return ip, true
+		}
+	}
+	return ip, false
+}
+
+func init() {
```
bradfitz, line 30: so in general, best practice is to avoid init funcs because they slow startup of programs that import this package and don't use it. It turns out our two importing packages will always use it, but we might want to get the API right to start, returning accessor functions that use a `sync.Once` to do this func once if/when needed. Then the linker can see when they're not used and discard it all.

so one package-level:

```go
var populateOnce sync.Once
```

And then name this `func init()` as `func populate()` and then at accessors, at the top:

```go
   populateOnce.Do(populate)
   return ...
```

util/publicdns/publicdns.go
```diff
@@ -0,0 +1,75 @@
...
+
+var DohIPsOfBase = map[string][]netaddr.IP{}
+
+func addDoH(ipStr, base string) {
+	ip := netaddr.MustParseIP(ipStr)
+	KnownDoH[ip] = base
+	DohIPsOfBase[base] = append(DohIPsOfBase[base], ip)
+}
+
+func DohV6(base string) (ip netaddr.IP, ok bool) {
```
bradfitz, line 21: exported symbols need proper docs (higher bar for exported stuff than package internals) in normal Go complete sentence style (https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences)

Also, this would be `DoH` per https://github.com/golang/go/wiki/CodeReviewComments#initialisms probably.

util/publicdns/publicdns.go
```diff
@@ -0,0 +1,75 @@
...
+	addDoH("2606:4700:4700::1003", "https://family.cloudflare-dns.com/dns-query")
+
+	// Google
+	addDoH("8.8.8.8", "https://dns.google/dns-query")
+	addDoH("8.8.4.4", "https://dns.google/dns-query")
+	addDoH("2001:4860:4860::8888", "https://dns.google/dns-query")
+	addDoH("2001:4860:4860::8844", "https://dns.google/dns-query")
+
+	// OpenDNS
+	// TODO(bradfitz): OpenDNS is unique amongst this current set in that
```
bradfitz, line 56: if we ever include these, we'll need to change the exported API of this package to let our forwarder code know when this is the case.  But fine for now.

util/publicdns/publicdns.go
```diff
@@ -0,0 +1,75 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package publicdns contains mapping and helpers for working with
+// public DNS providers
+package publicdns
+
+import "inet.af/netaddr"
+
+var KnownDoH = map[netaddr.IP]string{} // 8.8.8.8 => "https://..."
```
bradfitz, line 11:
```go
// KnownDoH returns a map of well-known public DNS IPs to their DoH URL.
// The returned map should not be modified.
func KnownDoH() map[netaddr.IP]string {
       populateOnce.Do(populate)
       return knownDoH
}
```

util/publicdns/publicdns.go
```diff
@@ -0,0 +1,75 @@
...
+
+var DohIPsOfBase = map[string][]netaddr.IP{}
```
bradfitz, line 13: similar

util/publicdns/publicdns.go
phirework: Ah good to know, thanks!

net/dns/publicdns/publicdns.go
```diff
@@ -0,0 +1,97 @@
...
+var populateOnce sync.Once
+
+// KnownDoH returns a map of well-known public DNS IPs to their DoH URL.
+// The returned map should not be modified.
+func KnownDoH() map[netaddr.IP]string {
+	populateOnce.Do(populate)
+	return knownDoH
+}
+
+// DoHIPsOfBase returns the public DNS provider associated with a given IP addr.
```
bradfitz, line 26: text isn't quite right. kinda backwards.

... returns a map of DNS server IP addresses keyed by their DoH URL. It is the inverse of KnownDoH.

bradfitz (approved):
LGTM otherwise

PR #4406, opened 2022-04-13
ipn/ipnlocal: make peerIPs return a sorted slice

maisem:
Currently peerIPs doesn't do any sorting of the routes it returns. This
is typically fine, however imagine the case of an HA subnet router
failover. When a route R moves from peer A to peer B, the output of
peerIPs changes. This in turn causes all the deephash check inside
wgengine to fail as the hashed value of [R1, R2] is different than
the hashed value of [R2, R1]. When the hash check failes, it causes
wgengine to reconfigure all routes in the OS. This is especially
problematic for macOS and iOS where we use the NetworkExtension.

This commit makes it that the peerIPs are always sorted when returned,
thus making the hash be consistent as long as the list of routes remains
static.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -2397,7 +2397,18 @@ func peerRoutes(peers []wgcfg.Peer, cgnatThreshold int) (routes []netaddr.IPPref
 	} else {
 		routes = append(routes, cgNATIPs...)
 	}
+
+	sort.Slice(routes, func(i, j int) bool {
+		return ipPrefixLess(routes[i], routes[j])
+	})
 	return routes
+}
+
+func ipPrefixLess(ri, rj netaddr.IPPrefix) bool {
+	if ri.IP() == rj.IP() {
+		return ri.Bits() < rj.Bits()
+	}
+	return ri.IP().Less(rj.IP())
 }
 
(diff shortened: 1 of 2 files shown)


maisem:
cc @nickoneill

ipn/ipnlocal/local.go
```diff
@@ -2397,9 +2397,20 @@ func peerRoutes(peers []wgcfg.Peer, cgnatThreshold int) (routes []netaddr.IPPref
 	} else {
 		routes = append(routes, cgNATIPs...)
 	}
+
+	sort.Slice(routes, func(i, j int) bool {
+		return ipPrefixLess(routes[i], routes[j])
+	})
```
bradfitz, line 2403: where's the non-determinism in the func above? I see no maps.

Is the input slice random from some caller?

bradfitz (approved):

ipn/ipnlocal/local.go
maisem: > Say node A and B both advertise the same route
> Control picks node A as primary
> A stops advertising route, B is primary now
> This causes ordering in peerIPs to change because the AllowedIPs slice in peers changed
> Note A and B are still connected

Copied from slack for posterity

PR #4408, opened 2022-04-14
Move magicDNS to netstack instead of handcrafted UDP

twitchyliquid64:
- [x] Moves magicDNS-specific handling out of `resolver.Resolver` & into `dns.Manager`. This greatly simplifies the Resolver to solely issuing queries and returning responses, without channels.
 - [x] Pass `*dns.Manager` in initialization of `netstack.Impl` so it can service DNS
 - [x] Hook _From Host_ (traffic received from the TUN) traffic in `netstack.Impl` so netstack can handle MagicDNS
 - [x] Implement shim in `netstack.Impl.acceptUDP` to handle MagicDNS requests
 - [x] ~Implement shim in `netstack.Impl.acceptTCP` to handle MagicDNS requests~ For a future PR, this is big enough already.

Note to reviewers: This change was not conducive to being split into multiple atomic PRs. Instead, I've split it into multiple commits that can be reviewed in progression.

### DIFF
--- a/net/dns/manager.go
+++ b/net/dns/manager.go
@@ -6,19 +6,50 @@ package dns
 
 import (
 	"bufio"
+	"context"
+	"errors"
 	"runtime"
+	"sync/atomic"
 	"time"
 
 	"inet.af/netaddr"
 	"tailscale.com/health"
 	"tailscale.com/net/dns/resolver"
+	"tailscale.com/net/packet"
+	"tailscale.com/net/tsaddr"
 	"tailscale.com/net/tsdial"
+	"tailscale.com/net/tstun"
 	"tailscale.com/types/dnstype"
 	"tailscale.com/types/ipproto"
 	"tailscale.com/types/logger"
+	"tailscale.com/util/clientmetric"
 	"tailscale.com/util/dnsname"
 	"tailscale.com/wgengine/monitor"
 )
+
+var (
+	magicDNSIP   = tsaddr.TailscaleServiceIP()
+	magicDNSIPv6 = tsaddr.TailscaleServiceIPv6()
+)
+
+var (
+	errFullQueue = errors.New("request queue full")
+)
+
+// maxActiveQueries returns the maximal number of DNS requests that be
+// can running.
+// If EnqueueRequest is called when this many requests are already pending,
+// the request will be dropped to avoid blocking the caller.
+func maxActiveQueries() int32 {
+	if runtime.GOOS == "ios" {
+		// For memory paranoia reasons on iOS, match the
+		// historical Tailscale 1.x..1.8 behavior for now
+		// (just before the 1.10 release).
+		return 64
+	}
+	// But for other platforms, allow more burstiness:
+	return 256
+}
 
 // We use file-ignore below instead of ignore because on some platforms,
 // the lint exception is necessary and on others it is not,
@@ -31,9 +62,23 @@ import (
 // Such operations should be wrapped in a timeout context.
 const reconfigTimeout = time.Second
 
+type response struct {
+	pkt  []byte
+	from netaddr.IPPort // where the packet needs to be sent
+}
+
 // Manager manages system DNS settings.
 type Manager struct {
 	logf logger.Logf
+
+	// When netstack is not used, Manager implements magic DNS.
+	// In this case, responses tracks completed DNS requests
+	// which need a response, and NextPacket() synthesizes a
+	// fake IP+UDP header to finish assembling the response.
+	//
+	// TODO(tom): Rip out once all platforms use netstack.
+	responses           chan response
+	activeQueriesAtomic int32
 
 	resolver *resolver.Resolver
 	os       OSConfigurator
@@ -46,9 +91,10 @@ func NewManager(logf logger.Logf, oscfg OSConfigurator, linkMon *monitor.Mon, di
 	}
 	logf = logger.WithPrefix(logf, "dns: ")
 	m := &Manager{
-		logf:     logf,
-		resolver: resolver.New(logf, linkMon, linkSel, dialer),
-		os:       oscfg,
+		logf:      logf,
+		resolver:  resolver.New(logf, linkMon, linkSel, dialer),
+		os:        oscfg,
+		responses: make(chan response),
 	}
 	m.logf("using %T", m.os)
 	return m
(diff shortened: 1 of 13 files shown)


twitchyliquid64:
Rebased onto main and should be good for review!

wgengine/netstack/netstack.go
```diff
@@ -436,7 +508,7 @@ func (ns *Impl) peerAPIPortAtomic(ip netaddr.IP) *uint32 {
 
 var viaRange = tsaddr.TailscaleViaRange()
 
-// shouldProcessInbound reports whether an inbound packet should be
+// shouldProcessInbound reports whether an inbound (wg -> us) packet should be
```
bradfitz, line 511: what's "us"?

also: WG for WireGuard

wgengine/netstack/netstack.go
```diff
@@ -390,7 +425,7 @@ func (ns *Impl) DialContextUDP(ctx context.Context, ipp netaddr.IPPort) (*gonet.
 	return gonet.DialUDP(ns.ipstack, nil, remoteAddress, ipType)
 }
 
-func (ns *Impl) injectOutbound() {
+func (ns *Impl) inject() {
```
bradfitz, line 428: time to document this probably?

wgengine/netstack/netstack.go
```diff
@@ -358,6 +364,35 @@ func (ns *Impl) updateIPs(nm *netmap.NetworkMap) {
 	}
 }
 
+// handleLocalPackets is hooked into the tun datapath for packets leaving
+// the host and arriving at tailscaled. This method returns filter.DropSilently
+// to intercept a packet for handling, for instance traffic to quad-100.
+func (ns *Impl) handleLocalPackets(p *packet.Parsed, t *tstun.Wrapper) filter.Response {
+	// If its not traffic to the service IP (ie: magicDNS) we don't
```
bradfitz, line 371: it's

bradfitz: i.e.

bradfitz (commented):
You're not doing any TCP interception to 100.100.100.100 yet in this PR, it seems?

twitchyliquid64:
> You're not doing any TCP interception to 100.100.100.100 yet in this PR, it seems?

This PR is big enough so I'm going to put that in a separate one. Ive got a working draft at https://github.com/tailscale/tailscale/commit/bd9539ded53acb8cefe664e6e750e2e7a5755db4, I'll clean it up and send a PR once this one is in.

I'll address your comments this afternoon.

net/tstun/wrap.go
```diff
@@ -135,9 +135,14 @@ type Wrapper struct {
 	PreFilterIn FilterFunc
 	// PostFilterIn is the inbound filter function that runs after the main filter.
 	PostFilterIn FilterFunc
-	// PreFilterOut is the outbound filter function that runs before the main filter
-	// and therefore sees the packets that may be later dropped by it.
-	PreFilterOut FilterFunc
+	// PreFilterOutNetstack is the outbound filter function that runs before the main filter
```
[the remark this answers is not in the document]
twitchyliquid64: I've revised to `PreFilterFromTunToNetstack` and `PreFilterFromTunToEngine`.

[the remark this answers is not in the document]
twitchyliquid64: Though on second read im thinking maybe it should be `PreFilterFromTunForNetstack` or `PreFilterFromTunNetstackHook` or something.

net/dns/manager.go
```diff
@@ -31,10 +62,24 @@ import (
 // Such operations should be wrapped in a timeout context.
 const reconfigTimeout = time.Second
 
+type response struct {
+	pkt  []byte
+	from netaddr.IPPort // where the packet needs to be sent
```
danderson, line 67: Why is it `from`, if this is where the packet needs to be sent? The comment and code don't seem to match.

net/dns/manager.go
```diff
@@ -31,10 +62,24 @@ import (
...
 // Manager manages system DNS settings.
 type Manager struct {
 	logf logger.Logf
 
+	// When netstack is not used, Manager implements magic DNS.
+	// In this case, responses tracks completed DNS requests
+	// which need a response, and NextPacket() synthesizes a
+	// fake IP+UDP header to finish assembling the response.
+	//
+	// TODO(tom): Rip out once all platforms use netstack.
```
danderson, line 79: Confused: I thought the plan is for all platforms to use netstack immediately?

net/tstun/wrap.go
```diff
@@ -692,6 +692,27 @@ func (t *Wrapper) SetFilter(filt *filter.Filter) {
 	t.filter.Store(filt)
 }
 
+// InjectInboundDirect makes the Wrapper device behave as if a packet
```
danderson, line 695: InjectOutboundDirect is the wrong name, should be InjectInboundPacketBuffer.

I still don't like these "Inbound" and "Outbound" relative descriptors that don't tell you what they're relative to, but I see you're doing this for consistency with other similar helpers, LGTM.

net/tstun/wrap.go
```diff
@@ -692,6 +692,27 @@ func (t *Wrapper) SetFilter(filt *filter.Filter) {
...
+// This path is typically used to deliver synthesized packets to the
+// host networking stack.
+func (t *Wrapper) InjectInboundPacketBuffer(pkt *stack.PacketBuffer) error {
+	buf := make([]byte, PacketStartOffset + pkt.Size())
+
+	n := copy(buf[PacketStartOffset:], pkt.NetworkHeader().View())
+	n += copy(buf[PacketStartOffset+n:], pkt.TransportHeader().View())
+	n += copy(buf[PacketStartOffset+n:], pkt.Data().AsRange().AsView())
+	if n != pkt.Size() {
+		panic("unexpected: revisit assumptions")
```
danderson, line 709: What assumptions should we revisit? :) "unexpected packet size after copy" would be better for when a user reports us this panic in the wild.

tsnet/tsnet.go
```diff
@@ -182,12 +182,12 @@ func (s *Server) start() error {
 		return err
 	}
 
-	tunDev, magicConn, ok := eng.(wgengine.InternalsGetter).GetInternals()
+	tunDev, magicConn, d, ok := eng.(wgengine.InternalsGetter).GetInternals()
```
danderson, line 185: dns, rather than d? The single-letter var looks out of place in context.

wgengine/netstack/netstack_test.go
```diff
@@ -39,12 +39,12 @@ func TestInjectInboundLeak(t *testing.T) {
 	if !ok {
 		t.Fatal("not an InternalsGetter")
 	}
-	tunWrap, magicSock, ok := ig.GetInternals()
+	tunWrap, magicSock, d, ok := ig.GetInternals()
```
danderson, line 42: same, d is a big nondescript unless I go find GetInternals' signature.

danderson (approved):
Sorry this took so long. Change stack LGTM.

This set of commits does not include wiring up the filter func in userspace.go to make the new DNS resolution path live. Is that intended?

net/dns/manager.go
twitchyliquid64: To/From relativism strikes again XD Its the address where the request came from, I'll update the comment.

net/dns/manager.go
twitchyliquid64: Yes, but I didnt want to make this commit series really hard to unravel if we needed to change directions quickly. Once its simmered, I'll pull out these fallback paths.

twitchyliquid64:
The new DNS path is wired up here:

https://github.com/tailscale/tailscale/pull/4408/files#diff-66f4f26f7c61ca193846818b1068d474ea39b3b9ad5efd0e51438f35b88de59eR239

I'll note that the old path remains wired: but because of how the two filters work (the netstack one takes precedence and takes the packets from the queue), the old path wont see traffic on binaries that have netstack installed.

wgengine/netstack/netstack_test.go
bradfitz: @twitchyliquid64, this wasn't done.

twitchyliquid64:
Heh, might have missed a -f. Will check / fix when I’m back in front of
my computer

net/dns/resolver/forwarder.go
```diff
@@ -601,17 +597,6 @@ type forwardQuery struct {
 	// ...
 }
 
-// forward forwards the query to all upstream nameservers and waits for
-// the first response.
-//
-// It either sends to f.responses and returns nil, or returns a
-// non-nil error (without sending to the channel).
-func (f *forwarder) forward(query packet) error {
-	ctx, cancel := context.WithTimeout(f.ctx, responseTimeout)
```
mihaip, line 610: @twitchyliquid64 `responseTimeout` is now unused/unenforced, was this an intentional change?

(noticed while discussing #5308 with @fylie, it seemed like that 5 second timeout might explain observed behavior, but we're not actually enforcing it anywhere).

twitchyliquid64: The timeout was refactored into `Resolver`: See tsdns.go:281. The other timeout is on the TCP DNS path (45 seconds for an idle connection), in `manager.go` ~line 415.

mihaip: Thanks, that makes sense. Sent out #5629 to clean it up.

PR #4410, opened 2022-04-14
net/dns/publicdns: add missing call to sync.Once.Do

phirework:


bradfitz:
`net/dns/publicdns: add missing call to sync.Once.Do`

bradfitz (approved):

PR #4417, opened 2022-04-15
control/controlclient: wire up PingRequest for disco & TSMP

raggi:
We could land this now, I've tested a round-trip.

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -850,7 +850,7 @@ func (c *Direct) sendMapRequest(ctx context.Context, maxPolls int, cb func(*netm
 
 		if pr := resp.PingRequest; pr != nil && c.isUniquePingRequest(pr) {
 			metricMapResponsePings.Add(1)
-			go answerPing(c.logf, c.httpc, pr)
+			go answerPing(c.logf, c.httpc, pr, c.pinger)
 		}
 		if u := resp.PopBrowserURL; u != "" && u != sess.lastPopBrowserURL {
 			sess.lastPopBrowserURL = u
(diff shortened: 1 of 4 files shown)


control/controlclient/direct.go
```diff
@@ -1414,20 +1425,20 @@ func postPingResult(now time.Time, logf logger.Logf, c *http.Client, pr *tailcfg
 		return err
 	}
 	// Send the results of the Ping, back to control URL.
```
bradfitz, line 1427: PingResult doesn't distinguish between TSMP and disco it seems? More reason to make a new JSON POST body type in tailcfg.

tailcfg/tailcfg.go
```diff
@@ -1218,6 +1219,41 @@ type PingRequest struct {
...
+	Endpoint string
+
+	// DERPRegionID is non-zero DERP region ID if DERP was used.
+	// It is not currently set for TSMP pings.
+	DERPRegionID int
+
+	// DERPRegionCode is the three-letter region code
+	// corresponding to DERPRegionID.
+	// It is not currently set for TSMP pings.
+	DERPRegionCode string
```
bradfitz, line 1245: omitempty on probably all of these?

tailcfg/tailcfg.go
```diff
@@ -1218,6 +1219,41 @@ type PingRequest struct {
...
+// Typically populated from an ipnstate.PingResult used in `tailscale ping`.
+type PingResponse struct {
+	Type string // ping type, such as TSMP or disco.
+
+	IP       string // ping destination
+	NodeIP   string // Tailscale IP of node handling IP (different for subnet routers)
+	NodeName string // DNS name base or (possibly not unique) hostname
+
+	Err            string
+	LatencySeconds float64
```
bradfitz, line 1232: docs? is this one set on errors too?

ipn/ipnstate/ipnstate.go
```diff
@@ -513,6 +515,22 @@ type PingResult struct {
 	// TODO(bradfitz): details like whether port mapping was used on either side? (Once supported)
 }
 
+func (pr *PingResult) ToPingResponse(pingType string) tailcfg.PingResponse {
```
bradfitz, line 518: it's probably a big enough struct to be a pointer at this point?

bradfitz: oh, you use it as a pointer elsewhere already. so yeah, here too.

control/controlclient/direct.go
```diff
@@ -1181,29 +1181,47 @@ func (c *Direct) isUniquePingRequest(pr *tailcfg.PingRequest) bool {
...
+	}
+	for _, t := range strings.Split(pr.Types, ",") {
+		switch t {
+		case "TSMP", "disco":
+			go doPingerPing(logf, c, pr, pinger, t)
+		// TODO(tailscale/corp#754)
+		// case "host":
+		// case "peerapi":
+		default:
+			logf("unsupported ping request type: %s", t)
```
bradfitz, line 1201: %q for bogus strings

tailcfg/tailcfg.go
```diff
@@ -1218,6 +1219,41 @@ type PingRequest struct {
...
+
+	// Endpoint is the ip:port if direct UDP was used.
+	// It is not currently set for TSMP pings.
+	Endpoint string
+
+	// DERPRegionID is non-zero DERP region ID if DERP was used.
+	// It is not currently set for TSMP pings.
+	DERPRegionID int
```
bradfitz, line 1240: so if we want to see if a disco ping resulted in a direct connection we'll need to send several PingRequests from control spaced out by a couple seconds to see if it eventually finds a path?

raggi: yes, I suspect we'll need to make a number of adjustments in the protocol once we aim to cover all of the behaviors.

I've been debating myself between sending more full round-trips or adding a field equivalent to `--until-direct`, and whether I might at some point want to send more than one ping response per pong token. I had a code path in an early prototype in control, though the complexity was much higher so I cut back to focus on getting some results sooner.

In the interest of working through the end to end incrementally, I think I want to defer this decision for some follow-up, though that might be soon.

bradfitz: SGTM. This is fine for now. Not a bad thing to send multiple pings anyway. Maybe that's how it'll even end up.

tailcfg/tailcfg.go
raggi: It might be, however if it is not set, Err should be set. Added in a doc string.

bradfitz (approved):

PR #4419, opened 2022-04-15
go.mod: bump u-root

raggi:
We only use the termios subpackage, so we're unaffected by
CVE-2020-7665, but the bump will let dependabot return to slumber.

Fixes https://github.com/tailscale/tailscale/security/dependabot/2


maisem (approved):
Thanks!

PR #4428, opened 2022-04-17
ipn/ipnlocal: only call updateFilter with mutex held

### LINKED ISSUE #4427: ipn/ipnlocal: (*LocalBackend).updateFilter data race?
Looks like `(*LocalBackend).updateFilter` is called inconsistently, sometimes with the lock held and sometimes without?

And looks like it needs to, as it reads/writes filterHash at: `deephash.Update(&b.filterHash`

/cc @maisem @danderson

bradfitz:
And rename to updateFilterLocked to prevent future mistakes.

Fixes #4427

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -119,13 +119,12 @@ type LocalBackend struct {
 	sshAtomicBool         syncs.AtomicBool
 	sshServer             SSHServer // or nil
 
-	filterHash deephash.Sum
-
 	filterAtomic            atomic.Value // of *filter.Filter
 	containsViaIPFuncAtomic atomic.Value // of func(netaddr.IP) bool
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/local.go
```diff
@@ -1825,6 +1826,12 @@ func (b *LocalBackend) setPrefsLockedOnEntry(caller string, newp *ipn.Prefs) {
 	userID := b.userID
 	cc := b.cc
 
+	// [GRINDER STATS LINE] - please don't remove (used for log parsing)
```
maisem, line 1829: why is this moving up?

maisem: oh i see

maisem (approved):
!!!

PR #1585, opened 2021-03-25
wgengine/monitor: on wall time jump, synthesize network change event

bradfitz:
... to force rebinds of TCP connections

Fixes #1555
Updates tailscale/felicity#4

PR #4429, opened 2022-04-17
wgengine/monitor: do not set timeJumped on iOS/Android

maisem:
In `(*Mon).Start` we don't run a timer to update `(*Mon).lastWall` on iOS and
Android as their sleep patterns are bespoke. However, in the debounce
goroutine we would notice that the the wall clock hadn't been updated
since the last event would assume that a time jump had occurred. This would
result in non-events being considered as major-change events.

This commit makes it so that `(*Mon).timeJumped` is never set to `true`
on iOS and Android.

### DIFF
--- a/wgengine/monitor/monitor.go
+++ b/wgengine/monitor/monitor.go
@@ -185,13 +185,7 @@ func (m *Mon) Start() {
 	}
 	m.started = true
 
-	switch runtime.GOOS {
-	case "ios", "android":
-		// For battery reasons, and because these platforms
-		// don't really sleep in the same way, don't poll
-		// for the wall time to detect for wake-for-sleep
-		// walltime jumps.
-	default:
+	if shouldMonitorTimeJump {
 		m.wallTimer = time.AfterFunc(pollWallTimeInterval, m.pollWallTime)
 	}
 
@@ -304,16 +298,9 @@ func (m *Mon) debounce() {
 		} else {
 			m.mu.Lock()
 
-			// See if we have a queued or new time jump signal.
-			m.checkWallTimeAdvanceLocked()
-			timeJumped := m.timeJumped
-			if timeJumped {
-				m.logf("time jumped (probably wake from sleep); synthesizing major change event")
-			}
-
 			oldState := m.ifState
-			ifChanged := !curState.EqualFiltered(oldState, interfaces.UseInterestingInterfaces, interfaces.UseInterestingIPs)
-			if ifChanged {
+			changed := !curState.EqualFiltered(oldState, interfaces.UseInterestingInterfaces, interfaces.UseInterestingIPs)
+			if changed {
 				m.gwValid = false
 				m.ifState = curState
 
@@ -322,9 +309,14 @@ func (m *Mon) debounce() {
 						jsonSummary(oldState), jsonSummary(curState))
 				}
 			}
-			changed := ifChanged || timeJumped
-			if changed {
-				m.timeJumped = false
+			// See if we have a queued or new time jump signal.
+			if shouldMonitorTimeJump && m.checkWallTimeAdvanceLocked() {
+				m.resetTimeJumpedLocked()
+				if !changed {
+					// Only log if it wasn't an interesting change.
+					m.logf("time jumped (probably wake from sleep); synthesizing major change event")
+					changed = true
+				}
 			}
 			for _, cb := range m.cbs {
 				go cb(changed, m.ifState)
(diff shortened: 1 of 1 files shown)


maisem:
Looking at logs from an iOS device with high battery usage over 24 hours
```
cat logs | grep -E "synthesizing major" | wc -l                                                                                                                                                                ~
      89
```

Similarly from another iOS device which isn't
```
cat logs.m | grep -E "synthesizing major" | wc -l                                                                                                                                                              ~
       4
```


Looking further, a single `synthesized major change event` results in many `endpoint updates`
```
2022-04-17 18:01:33.148633 +0000 UTC: magicsock: starting endpoint update (link-change-major)
2022-04-17 20:05:22.991312 +0000 UTC: monitor: time jumped (probably wake from sleep); synthesizing major change event
2022-04-17 20:05:22.992937 +0000 UTC: magicsock: starting endpoint update (link-change-major)
2022-04-17 20:05:23.248184 +0000 UTC: magicsock: ReSTUN: endpoint update active, need another later ("link-change-major")
2022-04-17 20:05:33.614057 +0000 UTC: magicsock: starting endpoint update (link-change-major)
2022-04-17 20:05:33.865806 +0000 UTC: magicsock: starting endpoint update (link-change-major)
2022-04-17 20:07:39.7659 +0000 UTC: monitor: time jumped (probably wake from sleep); synthesizing major change event
2022-04-17 20:07:39.767367 +0000 UTC: magicsock: starting endpoint update (link-change-major)
2022-04-17 20:07:40.174202 +0000 UTC: magicsock: ReSTUN: endpoint update active, need another later ("link-change-major")
2022-04-17 20:10:05.002711 +0000 UTC: monitor: time jumped (probably wake from sleep); synthesizing major change event
2022-04-17 20:10:05.003664 +0000 UTC: magicsock: starting endpoint update (link-change-major)
2022-04-17 20:10:06.085954 +0000 UTC: magicsock: starting endpoint update (link-change-major)
2022-04-17 20:10:06.343136 +0000 UTC: magicsock: starting endpoint update (link-change-major)
2022-04-17 20:10:07.07617 +0000 UTC: magicsock: starting endpoint update (link-change-major)
2022-04-17 20:10:08.683335 +0000 UTC: magicsock: ReSTUN: endpoint update active, need another later ("link-change-major")
2022-04-17 20:29:02.5321 +0000 UTC: monitor: time jumped (probably wake from sleep); synthesizing major change event
2022-04-17 20:29:02.53627 +0000 UTC: magicsock: starting endpoint update (link-change-major)
2022-04-17 20:39:40.381161 +0000 UTC: monitor: time jumped (probably wake from sleep); synthesizing major change event
2022-04-17 20:39:40.384685 +0000 UTC: magicsock: starting endpoint update (link-change-major)
2022-04-17 20:39:40.636415 +0000 UTC: magicsock: ReSTUN: endpoint update active, need another later ("link-change-major")
2022-04-17 20:39:40.646682 +0000 UTC: magicsock: starting endpoint update (link-change-major)
```

bradfitz:
I'll have to go dig through history to remind myself the reason for this but I thought this whole mechanism was added in part for mobile sleeping phones.

maisem:
> I'll have to go dig through history to remind myself the reason for this but I thought this whole mechanism was added in part for mobile sleeping phones.

Here is the place where we already have this check, and as a result we never start a timer to update `lastWall`. This causes all minor changes to be treated as major.

```golang
// Start starts the monitor.
// A monitor can only be started & closed once.
func (m *Mon) Start() {
	m.mu.Lock()
	defer m.mu.Unlock()
	if m.started || m.closed {
		return
	}
	m.started = true

	switch runtime.GOOS {
	case "ios", "android":
		// For battery reasons, and because these platforms
		// don't really sleep in the same way, don't poll
		// for the wall time to detect for wake-for-sleep
		// walltime jumps.
	default:
		m.wallTimer = time.AfterFunc(pollWallTimeInterval, m.pollWallTime)
	}

	if m.om == nil {
		return
	}
	m.goroutines.Add(2)
	go m.pump()
	go m.debounce()
}
```

maisem:
Looks like this was introduced mostly as-is in https://github.com/tailscale/tailscale/pull/1585.
So it's been an issue since it was introduced.

wgengine/monitor/monitor.go
```diff
@@ -372,8 +372,16 @@ func (m *Mon) pollWallTime() {
 // came out of sleep.
 func (m *Mon) checkWallTimeAdvanceLocked() {
 	now := wallTime()
-	if now.Sub(m.lastWall) > pollWallTimeInterval*3/2 {
-		m.timeJumped = true
+	switch runtime.GOOS {
+	case "android", "ios":
```
bradfitz, line 376: can you make a new package-level unexported constant for this so we don't duplicate this set of geese in two places and risk them getting out of sync again?

maisem: Good idea. Done.

wgengine/monitor/monitor.go
```diff
@@ -367,13 +364,28 @@ func (m *Mon) pollWallTime() {
 	m.wallTimer.Reset(pollWallTimeInterval)
 }
 
+func shouldMonitorTimeJump() bool {
```
bradfitz, line 367:
```
// shouldMonitorTimeJump is whether we keep a regular periodic timer running in the
// background watching for jumps in wall time.
//
// We don't do this on mobile platforms for blah blah blah.
const shouldMonitorTimeJump = runtime.GOOS != "android" && runtime.GOOS != "ios"
```

wgengine/monitor/monitor.go
```diff
@@ -367,13 +364,28 @@ func (m *Mon) pollWallTime() {
...
+}
+
 // checkWallTimeAdvanceLocked updates m.timeJumped, if wall time jumped
 // more than 150% of pollWallTimeInterval, indicating we probably just
 // came out of sleep.
 func (m *Mon) checkWallTimeAdvanceLocked() {
 	now := wallTime()
-	if now.Sub(m.lastWall) > pollWallTimeInterval*3/2 {
-		m.timeJumped = true
+	if shouldMonitorTimeJump() {
```
bradfitz, line 385: how about:

```go
if !shouldMonitorTimeJump {
    panic("unreachable") // if callers are correct
}
```

then fix the other caller.

bradfitz (approved):

wgengine/monitor/monitor.go
maisem: Done

wgengine/monitor/monitor.go
maisem: Done

bradfitz (approved):

PR #4433, opened 2022-04-18
cmd/tailscale: [ssh] enable StrictHostKeyChecking mode

maisem:
Updates #3802

### DIFF
--- a/cmd/tailscale/cli/ssh.go
+++ b/cmd/tailscale/cli/ssh.go
@@ -79,9 +79,13 @@ func runSSH(ctx context.Context, args []string) error {
 	argv := append([]string{
 		ssh,
 
+		// Only trust SSH hosts that we know about.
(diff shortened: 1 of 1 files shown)


bradfitz (approved):
Fine as is, but in a later change we should also fail nicely when the target isn't a known Tailscale SSH server.

pkpowell:
I'm getting `Host key verification failed.` with this change, despite having `StrictHostKeyChecking no` in my .ssh/config file.
```
No ED25519 host key is known for <node> and you have requested strict checking.
Host key verification failed.
```

I'm also getting
```
debug1: load_hostkeys: fopen '/Users/ard/Library/Application: No such file or directory
debug1: load_hostkeys: fopen Support/tailscale/ssh_known_hosts': No such file or directory
```
which looks like the space isn't being quoted between `Application` and `Support`.
Maybe these are related? This is on macos 11 and 12

PR #4436, opened 2022-04-18
net/dns/resolver: make DoH dialer use existing dnscache happy eyeball dialer

bradfitz:
Simplify the ability to reason about the DoH dialing code by reusing the
dnscache's dialer we already have.

Also, reduce the scope of the "ip" variable we don't want to close over.

This necessarily adds a new field to dnscache.Resolver:
SingleHostStaticResult, for when the caller already knows the IPs to be
returned.

### DIFF
--- a/net/dns/resolver/forwarder.go
+++ b/net/dns/resolver/forwarder.go
@@ -15,6 +15,7 @@ import (
 	"math/rand"
 	"net"
 	"net/http"
+	"net/url"
 	"runtime"
 	"sort"
 	"strconv"
@@ -26,6 +27,7 @@ import (
 	"inet.af/netaddr"
 	"tailscale.com/hostinfo"
 	"tailscale.com/net/dns/publicdns"
+	"tailscale.com/net/dnscache"
 	"tailscale.com/net/neterror"
 	"tailscale.com/net/netns"
 	"tailscale.com/net/tsdial"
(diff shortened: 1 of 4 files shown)


net/dns/resolver/forwarder.go
```diff
@@ -335,40 +336,43 @@ func (f *forwarder) packetListener(ip netaddr.IP) (packetListener, error) {
...
+		return "", nil, false
+	}
+	c, ok = f.getKnownDoHClientForProvider(urlBase)
+	if !ok {
+		return "", nil, false
 	}
+	return urlBase, c, true
+}
 
+func (f *forwarder) getKnownDoHClientForProvider(urlBase string) (c *http.Client, ok bool) {
```
bradfitz, line 348: this split is so the `ip` variable can't be accidentally used in the code below and that it's clear that the `*http.Client` returned is purely for the `urlBase`.

net/dns/resolver/forwarder.go
```diff
@@ -335,40 +336,43 @@ func (f *forwarder) packetListener(ip netaddr.IP) (packetListener, error) {
...
 	}
 	nsDialer := netns.NewDialer(f.logf)
+	dialer := dnscache.Dialer(nsDialer.DialContext, &dnscache.Resolver{AllHostStaticResult: allIPs})
 	c = &http.Client{
 		Transport: &http.Transport{
 			IdleConnTimeout: dohTransportTimeout,
 			DialContext: func(ctx context.Context, netw, addr string) (net.Conn, error) {
 				if !strings.HasPrefix(netw, "tcp") {
 					return nil, fmt.Errorf("unexpected network %q", netw)
 				}
```
bradfitz, line 366: this check remains pretty useless.  I could either remove it, or add more checks: that the port number is "443", for instance.

net/dnscache/dnscache.go
```diff
@@ -108,6 +112,19 @@ var debug = envknob.Bool("TS_DEBUG_DNS_CACHE")
 // If err is nil, ip will be non-nil. The v6 address may be nil even
 // with a nil error.
 func (r *Resolver) LookupIP(ctx context.Context, host string) (ip, v6 net.IP, allIPs []net.IPAddr, err error) {
```
bradfitz, line 114: this signature continues to be stupid & old (predating netaddr.IP).

it's on my list to clean up.

net/dns/resolver/forwarder.go
maisem: add a comment so that it doesn't regress

net/dnscache/dnscache.go
```diff
@@ -108,6 +112,19 @@ var debug = envknob.Bool("TS_DEBUG_DNS_CACHE")
...
+	if r.AllHostStaticResult != nil {
```
maisem, line 115: check for host string as well here and return an error if it doesn't match, or panic

maisem: I guess it says `AllHost`, fine as is

maisem (approved):

net/dnscache/dnscache.go
bradfitz: that's not something we can do here. There is no specific host that's preconfigured on resolver.

net/dns/resolver/forwarder.go
bradfitz: done.

net/dnscache/dnscache.go
maisem: I meant something like `SingleHostStaticResult []netaddr.IP` and `SingleHostName []string`

maisem (approved):

net/dnscache/dnscache.go
bradfitz: Done.

maisem (approved):

PR #4441, opened 2022-04-19
ssh/tailssh: start moving auth checks earlier, adding banner

bradfitz:
Updates #3802

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -87,7 +87,7 @@ func init() {
 
 // HandleSSHConn handles a Tailscale SSH connection from c.
 func (srv *server) HandleSSHConn(c net.Conn) error {
-	ss, err := srv.newSSHServer()
+	ss, err := srv.newSSHServer(c)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -87,7 +87,7 @@ func init() {
 
 // HandleSSHConn handles a Tailscale SSH connection from c.
 func (srv *server) HandleSSHConn(c net.Conn) error {
-	ss, err := srv.newSSHServer()
+	ss, err := srv.newSSHServer(c)
```
bradfitz, line 90: we pass this now earlier, as we'll use its identity to conditionally configure the ssh.Server

ssh/tailssh/tailssh.go
```diff
@@ -109,8 +109,28 @@ func (srv *server) OnPolicyChange() {
...
+		ServerConfigCallback: func(ctx ssh.Context) *gossh.ServerConfig {
+			conf := &gossh.ServerConfig{
+				NoClientAuth: true,
+				NoClientAuthCallback: func(m gossh.ConnMetadata) (*gossh.Permissions, error) {
+					if srv.requiresPubKey(m.User(), toIPPort(m.LocalAddr()), toIPPort(m.RemoteAddr())) {
+						return nil, errors.New("public key required") // any non-nil error will do
+					}
+					return nil, nil
+				},
+				BannerCallback: func(m gossh.ConnMetadata) string {
```
bradfitz, line 123: This is where we should do the policy evaluation! We have the ssh user at this point.

PR #4442, opened 2022-04-19
ssh/tailssh: terminate ssh auth early if no policy can match

maisem:
Updates #3802

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -87,7 +87,7 @@ func init() {
 
 // HandleSSHConn handles a Tailscale SSH connection from c.
 func (srv *server) HandleSSHConn(c net.Conn) error {
-	ss, err := srv.newSSHServer()
+	ss, err := srv.newConn()
 	if err != nil {
 		return err
 	}
@@ -109,34 +109,95 @@ func (srv *server) OnPolicyChange() {
 	}
 }
 
-func (srv *server) newSSHServer() (*ssh.Server, error) {
-	ss := &ssh.Server{
+// conn represents a single SSH connection and its associated
+// ssh.Server.
+type conn struct {
+	*ssh.Server
+	srv *server
+	ci  *sshConnInfo // set by NoClientAuthCallback
+
+	insecureSkipTailscaleAuth bool // used by tests.
+}
+
+func (c *conn) logf(format string, args ...any) {
+	if c.ci == nil {
+		c.srv.logf(format, args...)
+		return
+	}
+	format = fmt.Sprintf("%v: %v", c.ci.String(), format)
+	c.srv.logf(format, args...)
+}
+
+// PublicKeyHandler implements ssh.PublicKeyHandler is called by the the
+// ssh.Server when the client presents a public key.
+func (c *conn) PublicKeyHandler(ctx ssh.Context, key ssh.PublicKey) bool {
+	if c.ci == nil {
+		return false
+	}
+	if _, ok := c.srv.canProceed(c.ci, key); !ok {
+		c.logf("rejecting SSH public key %s", bytes.TrimSpace(gossh.MarshalAuthorizedKey(key)))
+		return false
+	}
+	c.logf("accepting SSH public key %s", bytes.TrimSpace(gossh.MarshalAuthorizedKey(key)))
+	return true
+}
+
+// errPubKeyRequired is returned by NoClientAuthCallback to make the client
+// resort to public-key auth; not user visible.
+var errPubKeyRequired = errors.New("ssh publickey required")
+
+// NoClientAuthCallback implements gossh.NoClientAuthCallback and is called by
+// the the ssh.Server when the client first connects with the "none"
+// authentication method.
+func (c *conn) NoClientAuthCallback(cm gossh.ConnMetadata) (*gossh.Permissions, error) {
+	if c.insecureSkipTailscaleAuth {
+		return nil, nil
+	}
+	ci, err := c.srv.connInfo(cm.User(), toIPPort(cm.LocalAddr()), toIPPort(cm.RemoteAddr()))
+	if err != nil {
+		c.logf("failed to get conninfo: %v", err)
+		return nil, gossh.ErrDenied
+	}
+	c.ci = ci
+	if _, ok := c.srv.canProceed(ci, nil /* without public key */); ok {
+		return nil, nil
+	}
+	if c.srv.havePubKeyPolicy(ci) {
+		return nil, errPubKeyRequired
+	}
+	return nil, gossh.ErrDenied
+}
+
+// ServerConfig implements ssh.ServerConfigCallback.
+func (c *conn) ServerConfig(ctx ssh.Context) *gossh.ServerConfig {
+	return &gossh.ServerConfig{
+		// OpenSSH presents this on failure as `Permission denied (tailscale).`
+		ImplictAuthMethod:    "tailscale",
+		NoClientAuth:         true, // required for the NoClientAuthCallback to run
+		NoClientAuthCallback: c.NoClientAuthCallback,
+	}
+}
+
+func (srv *server) newConn() (*conn, error) {
+	c := &conn{srv: srv}
+	c.Server = &ssh.Server{
+		Version:           "SSH-2.0-Tailscale",
 		Handler:           srv.handleSSH,
 		RequestHandlers:   map[string]ssh.RequestHandler{},
 		SubsystemHandlers: map[string]ssh.SubsystemHandler{},
+
 		// Note: the direct-tcpip channel handler and LocalPortForwardingCallback
 		// only adds support for forwarding ports from the local machine.
 		// TODO(maisem/bradfitz): add remote port forwarding support.
 		ChannelHandlers: map[string]ssh.ChannelHandler{
 			"direct-tcpip": ssh.DirectTCPIPHandler,
 		},
-		Version:                     "SSH-2.0-Tailscale",
 		LocalPortForwardingCallback: srv.mayForwardLocalPortTo,
-		NoClientAuthCallback: func(m gossh.ConnMetadata) (*gossh.Permissions, error) {
-			if srv.requiresPubKey(m.User(), toIPPort(m.LocalAddr()), toIPPort(m.RemoteAddr())) {
-				return nil, errors.New("public key required") // any non-nil error will do
-			}
-			return nil, nil
-		},
-		PublicKeyHandler: func(ctx ssh.Context, key ssh.PublicKey) bool {
-			if srv.acceptPubKey(ctx.User(), toIPPort(ctx.LocalAddr()), toIPPort(ctx.RemoteAddr()), key) {
-				srv.logf("accepting SSH public key %s", bytes.TrimSpace(gossh.MarshalAuthorizedKey(key)))
-				return true
-			}
-			srv.logf("rejecting SSH public key %s", bytes.TrimSpace(gossh.MarshalAuthorizedKey(key)))
-			return false
-		},
+
+		PublicKeyHandler:     c.PublicKeyHandler,
+		ServerConfigCallback: c.ServerConfig,
 	}
+	ss := c.Server
 	for k, v := range ssh.DefaultRequestHandlers {
 		ss.RequestHandlers[k] = v
 	}
@@ -153,7 +214,7 @@ func (srv *server) newSSHServer() (*ssh.Server, error) {
 	for _, signer := range keys {
 		ss.AddHostKey(signer)
 	}
-	return ss, nil
+	return c, nil
 }
 
 // mayForwardLocalPortTo reports whether the ctx should be allowed to port forward
@@ -167,37 +228,24 @@ func (srv *server) mayForwardLocalPortTo(ctx ssh.Context, destinationHost string
 	return ss.action.AllowLocalPortForwarding
 }
 
-// requiresPubKey reports whether the SSH server, during the auth negotiation
-// phase, should requires that the client send an SSH public key. (or, more
-// specifically, that "none" auth isn't acceptable)
-func (srv *server) requiresPubKey(sshUser string, localAddr, remoteAddr netaddr.IPPort) bool {
+// havePubKeyPolicy reports whether any policy rule may provide access by means
+// of a ssh.PublicKey.
+func (srv *server) havePubKeyPolicy(ci *sshConnInfo) bool {
+	// Is there any rule that looks like it'd require a public key for this
+	// sshUser?
 	pol, ok := srv.sshPolicy()
 	if !ok {
 		return false
 	}
-	a, ci, _, err := srv.evaluatePolicy(sshUser, localAddr, remoteAddr, nil)
-	if err == nil && (a.Accept || a.HoldAndDelegate != "") {
-		// Policy doesn't require a public key.
-		return false
-	}
-	if ci == nil {
-		// If we didn't get far enough along through evaluatePolicy to know the Tailscale
-		// identify of the remote side then it's going to fail quickly later anyway.
-		// Return false to accept "none" auth and reject the conn.
-		return false
-	}
-
-	// Is there any rule that looks like it'd require a public key for this
-	// sshUser?
 	for _, r := range pol.Rules {
 		if ci.ruleExpired(r) {
 			continue
 		}
-		if mapLocalUser(r.SSHUsers, sshUser) == "" {
+		if mapLocalUser(r.SSHUsers, ci.sshUser) == "" {
 			continue
 		}
 		for _, p := range r.Principals {
-			if principalMatchesTailscaleIdentity(p, ci) && len(p.PubKeys) > 0 {
+			if len(p.PubKeys) > 0 && principalMatchesTailscaleIdentity(p, ci) {
 				return true
 			}
 		}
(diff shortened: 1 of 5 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -253,42 +268,43 @@ func toIPPort(a net.Addr) (ipp netaddr.IPPort) {
...
-// evaluating the sshUser and remoteAddr against the SSHPolicy. The remoteAddr
-// and localAddr params must be Tailscale IPs.
-//
-// The return sshConnInfo will be non-nil, even on some errors, if the
-// evaluation made it far enough to resolve the remoteAddr to a Tailscale IP.
-func (srv *server) evaluatePolicy(sshUser string, localAddr, remoteAddr netaddr.IPPort, pubKey ssh.PublicKey) (_ *tailcfg.SSHAction, _ *sshConnInfo, localUser string, _ error) {
-	pol, ok := srv.sshPolicy()
-	if !ok {
-		return nil, nil, "", fmt.Errorf("tailssh: rejecting connection; no SSH policy")
+func (srv *server) connInfo(sshUser string, localAddr, remoteAddr netaddr.IPPort) (*sshConnInfo, error) {
```
bradfitz, line 271: // connInfo returns a partially populated sshConnInfo from the provided arguments, validating
// only that they represent a known Tailscale identity.

ssh/tailssh/tailssh.go
```diff
@@ -205,12 +210,21 @@ func (srv *server) requiresPubKey(sshUser string, localAddr, remoteAddr netaddr.
...
-		return false
+		return nil, false
 	}
-	return a.Accept || a.HoldAndDelegate != ""
+	if !a.Accept && a.HoldAndDelegate == "" {
+		return nil, false
+	}
+	lu, err = user.Lookup(localUser)
+	if err != nil {
+		return nil, false
```
bradfitz, line 225: log on this lookup error?

ssh/tailssh/tailssh.go
```diff
@@ -226,6 +240,7 @@ func (srv *server) sshPolicy() (_ *tailcfg.SSHPolicy, ok bool) {
 		return pol, true
 	}
 	if debugPolicyFile != "" {
+		srv.logf("reading debug SSH policy file: %v", debugPolicyFile)
```
bradfitz, line 243: "error reading ..."

(I like the word "error" in errors so it's easier to consistently grep for them)

ssh/tailssh/tailssh.go
```diff
@@ -253,42 +268,43 @@ func toIPPort(a net.Addr) (ipp netaddr.IPPort) {
...
+
+	ci.node = node
+	ci.uprof = &uprof
+	return ci, nil
+}
+
+// evaluatePolicy returns the SSHAction, sshConnInfo and localUser after
+// evaluating the sshUser and remoteAddr against the SSHPolicy. The remoteAddr
+// and localAddr params must be Tailscale IPs.
+func (srv *server) evaluatePolicy(ci *sshConnInfo, pubKey gossh.PublicKey) (_ *tailcfg.SSHAction, localUser string, _ error) {
```
bradfitz, line 298: // The pubKey may be nil for "none" auth.

I assume?

ssh/tailssh/tailssh.go
```diff
@@ -391,35 +407,27 @@ func (srv *server) handleSSH(s ssh.Session) {
...
-		}
-	}
-	ss := srv.newSSHSession(s, ci, lu)
-	ss.logf("handling new SSH connection from %v (%v) to ssh-user %q", ci.uprof.LoginName, ci.src.IP(), sshUser)
-	action, err = ss.resolveTerminalAction(action)
+	action, lu, err := srv.resolveTerminalAction(s, ci)
 	if err != nil {
-		ss.logf("resolveTerminalAction: %v", err)
-		io.WriteString(s.Stderr(), "Access denied: failed to resolve SSHAction.\n")
+		srv.logf("resolveTerminalAction: %v", err)
```
bradfitz, line 418: I'm trying to only do logfs on sshSession, but it's too early here, so try to include enough context for debugging. Maybe include ci? Maybe *sshConnInfo can be a Stringer?

ssh/tailssh/tailssh.go
```diff
@@ -391,35 +407,27 @@ func (srv *server) handleSSH(s ssh.Session) {
...
+		io.WriteString(s.Stderr(), "Access denied: failed to resolve SSHAction.\r\n")
```
bradfitz, line 419: SSHAction is a nerdy thing for us, but this is visible to end users. "failed during authorization check" ?

ssh/tailssh/tailssh.go
```diff
@@ -431,7 +439,18 @@ func (srv *server) handleSSH(s ssh.Session) {
 // Any action with a Message in the chain will be printed to ss.
```
bradfitz, line 439: docs on line 436 (new side 436) are stale now. "returns action" but no "action" is passed in now.

ssh/tailssh/tailssh.go
```diff
@@ -827,7 +844,7 @@ type sshConnInfo struct {
 	// pubKey is the public key presented by the client, or nil
 	// if they haven't yet sent one (as in the early "none" phase
 	// of authentication negotiation).
-	pubKey ssh.PublicKey
+	_pubKey ssh.PublicKey
```
bradfitz, line 847: what's going on here with the underscore!?

ssh/tailssh/tailssh.go
```diff
@@ -122,19 +122,37 @@ func (srv *server) newSSHServer() (*ssh.Server, error) {
...
 			}
-			return nil, nil
+			return nil, gossh.ErrDenied
 		},
 		PublicKeyHandler: func(ctx ssh.Context, key ssh.PublicKey) bool {
-			if srv.acceptPubKey(ctx.User(), toIPPort(ctx.LocalAddr()), toIPPort(ctx.RemoteAddr()), key) {
-				srv.logf("accepting SSH public key %s", bytes.TrimSpace(gossh.MarshalAuthorizedKey(key)))
-				return true
+			conn := ctx
+			ci, err := srv.connInfo(conn.User(), toIPPort(conn.LocalAddr()), toIPPort(conn.RemoteAddr()))
```
bradfitz, line 141: this is all wasteful, doing up to 17 ssh.ConnInfo calls with the same info.

can you make this ci once early, once conn.User is first known?

ssh/tailssh/tailssh.go
```diff
@@ -122,19 +122,37 @@ func (srv *server) newSSHServer() (*ssh.Server, error) {
...
+			if err != nil {
+				return false
+			}
+			if _, yes := srv.canProceed(ci, key); !yes {
+				srv.logf("rejecting SSH public key %s", bytes.TrimSpace(gossh.MarshalAuthorizedKey(key)))
+				return false
+			}
+			srv.logf("accepting SSH public key %s", bytes.TrimSpace(gossh.MarshalAuthorizedKey(key)))
```
bradfitz, line 149: for who?

include ci stringification in both these.

ssh/tailssh/tailssh.go
```diff
@@ -122,19 +122,37 @@ func (srv *server) newSSHServer() (*ssh.Server, error) {
...
+			return true
+		},
+		ServerConfigCallback: func(ctx ssh.Context) *gossh.ServerConfig {
+			return &gossh.ServerConfig{
+				ImplictAuthMethod: "tailscale",
```
bradfitz, line 154: add a comment above this saying what the end result is (that the OpenSSH client shows it parens at failure)

ssh/tailssh/tailssh.go
maisem: Oh this was supposed to be deleted, but I initially only renamed it to find usage breaks

ssh/tailssh/tailssh.go
maisem: not an error, just log that it's reading the debug file

ssh/tailssh/tailssh.go
maisem: it's not partially populated, it's always fully populated now

bradfitz: it doesn't have pubKey.

bradfitz: oh, you actually removed it now? that was the underscore thing

maisem: yeah connInfo doesn't have pubkey now.

ssh/tailssh/tailssh.go
```diff
@@ -122,19 +125,37 @@ func (srv *server) newSSHServer() (*ssh.Server, error) {
 		},
 		Version:                     "SSH-2.0-Tailscale",
 		LocalPortForwardingCallback: srv.mayForwardLocalPortTo,
-		NoClientAuthCallback: func(m gossh.ConnMetadata) (*gossh.Permissions, error) {
-			if srv.requiresPubKey(m.User(), toIPPort(m.LocalAddr()), toIPPort(m.RemoteAddr())) {
-				return nil, errors.New("public key required") // any non-nil error will do
+		NoClientAuthCallback: func(conn gossh.ConnMetadata) (*gossh.Permissions, error) {
+			var err error
+			ci, err = srv.connInfo(conn.User(), toIPPort(conn.LocalAddr()), toIPPort(conn.RemoteAddr()))
```
bradfitz, line 130: This is just one `:=` refactoring away to silently break the code below where `ci == nil` on line 144 will then be true.

Hmm. Wish we had better tests here.

ssh/tailssh/tailssh.go
```diff
@@ -122,19 +125,37 @@ func (srv *server) newSSHServer() (*ssh.Server, error) {
 		},
 		Version:                     "SSH-2.0-Tailscale",
 		LocalPortForwardingCallback: srv.mayForwardLocalPortTo,
-		NoClientAuthCallback: func(m gossh.ConnMetadata) (*gossh.Permissions, error) {
-			if srv.requiresPubKey(m.User(), toIPPort(m.LocalAddr()), toIPPort(m.RemoteAddr())) {
-				return nil, errors.New("public key required") // any non-nil error will do
+		NoClientAuthCallback: func(conn gossh.ConnMetadata) (*gossh.Permissions, error) {
```
bradfitz, line 128: upstream gliderlabs/ssh doesn't have this hook. I had removed it in my https://github.com/tailscale/tailscale/pull/4441 and moved it to ServerConfigCallback instead. Since you're already using ServerConfigCallback, do you want to also do that in this PR, then also removing that modification to our gliderlabs/ssh fork in this PR too like #4441?

ssh/tailssh/tailssh.go
```diff
@@ -122,19 +125,37 @@ func (srv *server) newSSHServer() (*ssh.Server, error) {
...
+			var err error
+			ci, err = srv.connInfo(conn.User(), toIPPort(conn.LocalAddr()), toIPPort(conn.RemoteAddr()))
+			if err != nil {
+				srv.logf("failed to get conninfo: %v", err)
+				return nil, gossh.ErrDenied
+			}
+			if _, withoutPubKey := srv.canProceed(ci, nil); withoutPubKey {
```
bradfitz, line 135: not a huge fan of all these new ways of spelling "ok" you're adding like "yes" and "withoutPubKey". This one in particular is very verbose.

You could write "ok" instead and then put "// Can proceed without a public key" or `canProceed(ci, nil /* without pubkey */)`

ssh/tailssh/tailssh.go
```diff
@@ -167,33 +188,20 @@ func (srv *server) mayForwardLocalPortTo(ctx ssh.Context, destinationHost string
 	return ss.action.AllowLocalPortForwarding
 }
 
-// requiresPubKey reports whether the SSH server, during the auth negotiation
-// phase, should requires that the client send an SSH public key. (or, more
-// specifically, that "none" auth isn't acceptable)
-func (srv *server) requiresPubKey(sshUser string, localAddr, remoteAddr netaddr.IPPort) bool {
+// havePubKeyPolicy reports whether any policy may provide access by means of a
```
bradfitz, line 191: any policy rule

ssh/tailssh/tailssh.go
```diff
@@ -167,33 +188,20 @@ func (srv *server) mayForwardLocalPortTo(ctx ssh.Context, destinationHost string
...
-	// sshUser?
 	for _, r := range pol.Rules {
 		if ci.ruleExpired(r) {
 			continue
 		}
-		if mapLocalUser(r.SSHUsers, sshUser) == "" {
+		if mapLocalUser(r.SSHUsers, ci.sshUser) == "" {
 			continue
 		}
 		for _, p := range r.Principals {
```
bradfitz, line 207: move the len(p.PubKeys) > 0 part of the short circuit first as it's cheaper

ssh/tailssh/tailssh.go
```diff
@@ -253,42 +272,46 @@ func toIPPort(a net.Addr) (ipp netaddr.IPPort) {
...
+func (srv *server) evaluatePolicy(ci *sshConnInfo, pubKey gossh.PublicKey) (_ *tailcfg.SSHAction, localUser string, _ error) {
+	pol, ok := srv.sshPolicy()
+	if !ok {
+		return nil, "", fmt.Errorf("tailssh: rejecting connection; no SSH policy")
 	}
-	a, localUser, ok := evalSSHPolicy(pol, ci)
+	a, localUser, ok := srv.evalSSHPolicy(pol, ci, pubKey)
 	if !ok {
-		return nil, ci, "", fmt.Errorf("ssh: access denied for %q from %v", uprof.LoginName, ci.src.IP())
+		return nil, "", fmt.Errorf("ssh: access denied for %q from %v", ci.uprof.LoginName, ci.src.IP())
```
bradfitz, line 312: it's not so much that access was denied as there was no matching access grant.  This is a little misleading because the following line isn't the opposite of "access denied": line 314 could return an SSHAction that's Rejected.

bradfitz (approved):

ssh/tailssh/tailssh.go
maisem: Created a new struct for this. PTAL

ssh/tailssh/tailssh.go
```diff
@@ -109,34 +109,75 @@ func (srv *server) OnPolicyChange() {
...
 }
 
-func (srv *server) newSSHServer() (*ssh.Server, error) {
-	ss := &ssh.Server{
+// singleConnSSHServer represents a single SSH connection and it's associated
+// ssh.Server.
+type singleConnSSHServer struct {
+	*ssh.Server
+	srv  *server
+	logf logger.Logf
```
bradfitz, line 117: maybe be a method instead?

ssh/tailssh/tailssh.go
```diff
@@ -109,34 +109,87 @@ func (srv *server) OnPolicyChange() {
...
+func (c *conn) logf(format string, args ...any) {
+	if c.ci == nil {
+		c.srv.logf(format, args...)
+		return
+	}
+	format = fmt.Sprintf("%v: %v", c.ci.String(), format)
+	c.srv.logf(format, args...)
+}
+
+func (c *conn) PublicKeyHandler(ctx ssh.Context, key ssh.PublicKey) bool {
```
bradfitz, line 131: // PublicKeyHandler is the gossh.Server.PublicKeyHandler implementation.

or whatever

ssh/tailssh/tailssh.go
```diff
@@ -109,34 +109,87 @@ func (srv *server) OnPolicyChange() {
...
+	}
+	if _, ok := c.srv.canProceed(c.ci, key); !ok {
+		c.logf("rejecting SSH public key %s", bytes.TrimSpace(gossh.MarshalAuthorizedKey(key)))
+		return false
+	}
+	c.logf("accepting SSH public key %s", bytes.TrimSpace(gossh.MarshalAuthorizedKey(key)))
+	return true
+}
+
+func (c *conn) NoClientAuthCallback(cm gossh.ConnMetadata) (*gossh.Permissions, error) {
```
bradfitz, line 143: same

ssh/tailssh/tailssh.go
```diff
@@ -109,34 +109,87 @@ func (srv *server) OnPolicyChange() {
...
+	if err != nil {
+		c.logf("failed to get conninfo: %v", err)
+		return nil, gossh.ErrDenied
+	}
+	c.ci = ci
+	if _, ok := c.srv.canProceed(ci, nil /* without public key */); ok {
+		return nil, nil
+	}
+	if c.srv.havePubKeyPolicy(ci) {
+		return nil, fmt.Errorf("needs more info")
```
bradfitz, line 157: return nil, errPubKeyRequired


var errPubKeyRequired = errors.New("some error") // to make NoClientAuthCallback proceed to public-key auth; not user visible

ssh/tailssh/tailssh.go
```diff
@@ -109,34 +109,87 @@ func (srv *server) OnPolicyChange() {
...
+	}
+	return nil, gossh.ErrDenied
+}
+
+func (c *conn) ServerConfig(ctx ssh.Context) *gossh.ServerConfig {
```
bradfitz, line 162: say what hook this implements

ssh/tailssh/tailssh.go
```diff
@@ -109,34 +109,87 @@ func (srv *server) OnPolicyChange() {
...
+		ImplictAuthMethod:    "tailscale",
+		NoClientAuth:         true,
+		NoClientAuthCallback: c.NoClientAuthCallback,
+	}
+}
+
+func (srv *server) newConn() (*conn, error) {
+	c := &conn{
+		srv: srv,
+	}
```
bradfitz, line 174: one line?

bradfitz (approved):

PR #4443, opened 2022-04-19
net/ping, control/controlclient: add ICMP pings

raggi:
```
control: ping to 8.8.8.8 completed in 0.0129. pinger.Ping took 17.361535ms seconds
control: postPingResult: sending ping results to http://localhost:31544/pong/a42f028ba0636fea63d348a80bb6b185 ...
control: postPingResult complete to http://localhost:31544/pong/a42f028ba0636fea63d348a80bb6b185 (after 1ms)```

### DIFF
--- a/cmd/tailscale/cli/ping.go
+++ b/cmd/tailscale/cli/ping.go
@@ -18,6 +18,7 @@ import (
 	"github.com/peterbourgon/ff/v3/ffcli"
 	"tailscale.com/ipn"
 	"tailscale.com/ipn/ipnstate"
+	"tailscale.com/tailcfg"
 )
 
 var pingCmd = &ffcli.Command{
@@ -26,7 +27,7 @@ var pingCmd = &ffcli.Command{
 	ShortHelp:  "Ping a host at the Tailscale layer, see how it routed",
 	LongHelp: strings.TrimSpace(`
 
-The 'tailscale ping' command pings a peer node at the Tailscale layer
+The 'tailscale ping' command pings a peer node from the Tailscale layer
 and reports which route it took for each response. The first ping or
 so will likely go over DERP (Tailscale's TCP relay protocol) while NAT
 traversal finds a direct path through.
@@ -48,7 +49,8 @@ relay node.
 		fs := newFlagSet("ping")
 		fs.BoolVar(&pingArgs.verbose, "verbose", false, "verbose output")
 		fs.BoolVar(&pingArgs.untilDirect, "until-direct", true, "stop once a direct path is established")
-		fs.BoolVar(&pingArgs.tsmp, "tsmp", false, "do a TSMP-level ping (through IP + wireguard, but not involving host OS stack)")
+		fs.BoolVar(&pingArgs.tsmp, "tsmp", false, "do a TSMP-level ping (through wireguard, but not either host OS stack)")
+		fs.BoolVar(&pingArgs.icmp, "icmp", false, "do a ICMP-level ping (through wireguard, but not the local host OS stack)")
 		fs.IntVar(&pingArgs.num, "c", 10, "max number of pings to send")
 		fs.DurationVar(&pingArgs.timeout, "timeout", 5*time.Second, "timeout before giving up on a ping")
 		return fs
@@ -60,7 +62,18 @@ var pingArgs struct {
 	untilDirect bool
 	verbose     bool
 	tsmp        bool
+	icmp        bool
 	timeout     time.Duration
+}
+
+func pingType() tailcfg.PingType {
+	if pingArgs.tsmp {
+		return tailcfg.PingTSMP
+	}
+	if pingArgs.icmp {
+		return tailcfg.PingICMP
+	}
+	return tailcfg.PingDisco
 }
 
 func runPing(ctx context.Context, args []string) error {
@@ -111,7 +124,7 @@ func runPing(ctx context.Context, args []string) error {
 	anyPong := false
 	for {
 		n++
-		bc.Ping(ip, pingArgs.tsmp)
+		bc.Ping(ip, pingType())
 		timer := time.NewTimer(pingArgs.timeout)
 		select {
 		case <-timer.C:
@@ -132,10 +145,10 @@ func runPing(ctx context.Context, args []string) error {
 			if pr.DERPRegionID != 0 {
 				via = fmt.Sprintf("DERP(%s)", pr.DERPRegionCode)
 			}
-			if pingArgs.tsmp {
+			if via == "" {
 				// TODO(bradfitz): populate the rest of ipnstate.PingResult for TSMP queries?
-				// For now just say it came via TSMP.
-				via = "TSMP"
+				// For now just say which protocol it used.
+				via = string(pingType())
 			}
 			anyPong = true
 			extra := ""
@@ -143,7 +156,7 @@ func runPing(ctx context.Context, args []string) error {
 				extra = fmt.Sprintf(", %d", pr.PeerAPIPort)
 			}
 			printf("pong from %s (%s%s) via %v in %v\n", pr.NodeName, pr.NodeIP, extra, via, latency)
-			if pingArgs.tsmp {
+			if pingArgs.tsmp || pingArgs.icmp {
 				return nil
 			}
 			if pr.Endpoint != "" && pingArgs.untilDirect {
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -126,9 +126,9 @@ type Options struct {
 
 // Pinger is a subset of the wgengine.Engine interface, containing just the Ping method.
 type Pinger interface {
-	// Ping is a request to start a discovery or TSMP ping with the peer handling
-	// the given IP and then call cb with its ping latency & method.
-	Ping(ip netaddr.IP, useTSMP bool, cb func(*ipnstate.PingResult))
+	// Ping is a request to start a ping with the peer handling the given IP and
+	// then call cb with its ping latency & method.
+	Ping(ip netaddr.IP, pingType tailcfg.PingType, cb func(*ipnstate.PingResult))
 }
 
 type Decompressor interface {
@@ -1197,11 +1197,10 @@ func answerPing(logf logger.Logf, c *http.Client, pr *tailcfg.PingRequest, pinge
 		return
 	}
 	for _, t := range strings.Split(pr.Types, ",") {
-		switch t {
-		case "TSMP", "disco":
-			go doPingerPing(logf, c, pr, pinger, t)
+		switch pt := tailcfg.PingType(t); pt {
+		case tailcfg.PingTSMP, tailcfg.PingDisco, tailcfg.PingICMP:
+			go doPingerPing(logf, c, pr, pinger, pt)
 		// TODO(tailscale/corp#754)
-		// case "host":
 		// case "peerapi":
 		default:
 			logf("unsupported ping request type: %q", t)
@@ -1402,13 +1401,13 @@ func (c *Direct) DoNoiseRequest(req *http.Request) (*http.Response, error) {
 
 // doPingerPing sends a Ping to pr.IP using pinger, and sends an http request back to
 // pr.URL with ping response data.
-func doPingerPing(logf logger.Logf, c *http.Client, pr *tailcfg.PingRequest, pinger Pinger, pingType string) {
+func doPingerPing(logf logger.Logf, c *http.Client, pr *tailcfg.PingRequest, pinger Pinger, pingType tailcfg.PingType) {
 	if pr.URL == "" || pr.IP.IsZero() || pinger == nil {
 		logf("invalid ping request: missing url, ip or pinger")
 		return
 	}
 	start := time.Now()
-	pinger.Ping(pr.IP, pingType == "TSMP", func(res *ipnstate.PingResult) {
+	pinger.Ping(pr.IP, pingType, func(res *ipnstate.PingResult) {
 		// Currently does not check for error since we just return if it fails.
 		postPingResult(start, logf, c, pr, res.ToPingResponse(pingType))
 	})
--- a/ipn/message.go
+++ b/ipn/message.go
@@ -52,8 +52,8 @@ type SetPrefsArgs struct {
 }
 
 type PingArgs struct {
-	IP      string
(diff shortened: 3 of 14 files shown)

### CODE REVIEW

File: net/ping/ping.go [hunk 1]
```diff
@@ -0,0 +1,133 @@
...
+// Package ping provides utilities for ICMP pings. It currently delegates to
+// ping programs, in order to avoid capability and API restrictions.
+package ping
+
+import (
+	"errors"
+	"io"
+	"os"
+	"os/exec"
+	"regexp"
```
bradfitz, line 14: we've historically not used used the regexp package anywhere reachable from iOS because it brings in huge unicode tables.

control/controlclient/icmp_pinger.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"inet.af/netaddr"
+	"tailscale.com/ipn/ipnstate"
+	"tailscale.com/net/ping"
+)
+
+// ICMPPinger wraps the ping package in a Pinger interface.
+type ICMPPinger struct{}
+
+func (i *ICMPPinger) Ping(ip netaddr.IP, useTSMP bool, cb func(*ipnstate.PingResult)) {
```
bradfitz, line 20: func (ICMPPinger)

control/controlclient/direct.go
```diff
@@ -1198,8 +1198,9 @@ func answerPing(logf logger.Logf, c *http.Client, pr *tailcfg.PingRequest, pinge
 		switch t {
 		case "TSMP", "disco":
 			go doPingerPing(logf, c, pr, pinger, t)
+		case "host":
+			go doPingerPing(logf, c, pr, &ICMPPinger{}, t)
```
bradfitz, line 1202: ICMPPinger{} unless you foresee this growing state or configuration over time?

bradfitz (commented):
So you gave up on the idea of using netstack, eh?

raggi:
> So you gave up on the idea of using netstack, eh?

The issue expands into a compatibility matrix problem: in an overwhelming number of environments we can find and run a ping command, but the opposite is true for writing ICMP packets, where being a user, or being on windows makes that difficult. On Windows you need to use `Iphlpapi.dll` with the usual runtime dynamic linking strategy. It might be worth revisiting in time, but in the interest of landing a working solution now, this will cover a broad set of use cases with easy to understand failure conditions.

net/ping/ping.go
raggi: Ah, that's a very fair point, and this is relatively easy to parse by hand, I'll do so.

control/controlclient/direct.go
raggi: I don't, I want to actually replace the pinger abstraction here, but I didn't want to inline that refactor with this change.

raggi:
> > > So you gave up on the idea of using netstack, eh?
> > 
> > 
> > The issue expands into a compatibility matrix problem: in an overwhelming number of environments we can find and run a ping command, but the opposite is true for writing ICMP packets, where being a user, or being on windows makes that difficult. On Windows you need to use Iphlpapi.dll with the usual runtime dynamic linking strategy. It might be worth revisiting in time, but in the interest of landing a working solution now, this will cover a broad set of use cases with easy to understand failure conditions.
> 
> We might be talking about different things.
> 
> All our pings are meant to only go to Tailscale peers, so we can just inject the trivial ICMP packet (by hand even without netstack!) into Wireguard and from Wireguard without any TUN or os/exec or host kernel help.
> 
> I don't think we need PingRequest (at least for now) to support pinging e.g. 8.8.8.8.

I agree that 8.8.8.8 is not necessarily that useful to ping in many scenarios, being able to ping the hosts perspective of a target does have useful diagnostic capabilities. It can help determine if magicdns is working, if routing is working, if subnet routing is working, and so on - from the send side. If we only care about reachability of the remote then the tsmp and/or disco pings are probably usable?

I could constrain this back down to only tailscale nodes, and inject a packet, but I think that becomes very closely equivalent to the disco ping, where the send side operates entirely inside tailscaled, and the receive side drops out in [s.userPing](https://github.com/tailscale/tailscale/blob/v1.22.2/wgengine/netstack/netstack.go#L558) as modified here, or the host tun rx. What are the conditions in which an injected echo request would provide usefully different results from a disco ping request?

bradfitz:
> It can help determine if magicdns is working, if routing is working, if subnet routing is working, and so on - from the send side. If we only care about reachability of the remote then the tsmp and/or disco pings are probably usable?

So I viewed ICMP as mostly a test of whether the destination node's host stack was working and its firewall wasn't blocking things.

The main problem we hit when TSMP works but other things don't is that the host has firewalls enabled (Linux, Windows, Mac... like everything)

I'm less excited about trying to test MagicDNS & routing from the client's side because we're running in such different environments: in a sandbox on macOS/iOS, and as root/admin on Linux/Windows.... it's not what the end user's stuff is running as. For that, we'd want to ask the GUI (if present on macOS and Windows) to do the ping.

We can already look at the routing table ourselves and just send that for debugging.

cmd/tailscale/cli/ping.go
```diff
@@ -50,6 +50,7 @@ relay node.
 		fs.BoolVar(&pingArgs.verbose, "verbose", false, "verbose output")
 		fs.BoolVar(&pingArgs.untilDirect, "until-direct", true, "stop once a direct path is established")
 		fs.BoolVar(&pingArgs.tsmp, "tsmp", false, "do a TSMP-level ping (through IP + wireguard, but not involving host OS stack)")
+		fs.BoolVar(&pingArgs.icmp, "icmp", false, "do a ICMP-level ping (through IP + wireguard. Remote host OS stack only)")
```
raggi, line 53: I don't like having two args here, but I also wasn't too keen on breaking the UI interface - what do you think?

wgengine/userspace.go
```diff
@@ -1313,6 +1327,55 @@ func (e *userspaceEngine) mySelfIPMatchingFamily(dst netaddr.IP) (src netaddr.IP
 	return netaddr.IP{}, errors.New("no self address in netmap matching address family")
 }
 
+func (e *userspaceEngine) sendICMPEcho(ip netaddr.IP, peer *tailcfg.Node, res *ipnstate.PingResult, cb func(*ipnstate.PingResult)) {
```
raggi, line 1330: I can probably factor TSMP and ICMP together, but the logic can get a little dense. Happy to go in that direction if it seems worthwhile.

raggi (commented):
I switched to ICMP internal injection and reception, PTAL.

ipn/message.go
```diff
@@ -56,8 +56,8 @@ type FakeExpireAfterArgs struct {
 }
 
 type PingArgs struct {
-	IP      string
-	UseTSMP bool
+	IP   string
+	Type string
```
raggi, line 60: I think this isn't used by other ipn API clients, so the change is ok, but do we have any additional process or considerations for such changes? Are there things I should do to help me check I didn't break anything?

wgengine/wgengine.go
```diff
@@ -154,9 +154,9 @@ type Engine interface {
 	// status builder.
 	UpdateStatus(*ipnstate.StatusBuilder)
 
-	// Ping is a request to start a discovery ping with the peer handling
-	// the given IP and then call cb with its ping latency & method.
-	Ping(ip netaddr.IP, useTSMP bool, cb func(*ipnstate.PingResult))
+	// Ping is a request to start a ping with the peer handling the given IP and
```
bradfitz, line 157: clarify what the set of valid "pingType" strings are.

wgengine/userspace.go
```diff
@@ -1373,6 +1436,19 @@ func (e *userspaceEngine) setTSMPPongCallback(data [8]byte, cb func(packet.TSMPP
 	}
 }
 
+func (e *userspaceEngine) setICMPEchoCallback(idSeq uint32, cb func()) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	if e.icmpEchoCallback == nil {
+		e.icmpEchoCallback = map[uint32]func(){}
+	}
```
bradfitz, line 1444: you can drop this init now if you use the new helpers we added recently:

```go
        if cb == nil {
		delete(e.icmpEchoCallback, idSeq)
	} else {
		mak.Set(&e.icmpEchoCallback, idSeq, cb)
	}
}
```

Your call. (it's safe to delete from a nil map too)

wgengine/userspace.go
bradfitz: is this an echo request or echo reply?

net/tstun/wrap.go
```diff
@@ -144,6 +144,11 @@ type Wrapper struct {
 	// OnTSMPPongReceived, if non-nil, is called whenever a TSMP pong arrives.
 	OnTSMPPongReceived func(packet.TSMPPongReply)
 
+	// OnICMPEchoReceived, if non-nil, is called whenever a ICMP echo response
```
bradfitz, line 147: Put Response in the name too? Otherwise I think we'll be confused every time we read it.

net/tstun/wrap.go
```diff
@@ -561,6 +566,16 @@ func (t *Wrapper) filterIn(buf []byte) filter.Response {
 		}
 	}
 
+	if p.IsEchoResponse() {
+		if f := t.OnICMPEchoReceived; f != nil {
```
bradfitz, line 570: ; f != nil && f(p) {

wgengine/userspace.go
```diff
@@ -136,6 +136,7 @@ type userspaceEngine struct {
 	networkMapCallbacks map[*someHandle]NetworkMapCallback
 	tsIPByIPPort        map[netaddr.IPPort]netaddr.IP          // allows registration of IP:ports as belonging to a certain Tailscale IP for whois lookups
 	pongCallback        map[[8]byte]func(packet.TSMPPongReply) // for TSMP pong responses
+	icmpEchoCallback    map[uint32]func()                      // for ICMP echo responses
```
bradfitz, line 139: document map key

I guess pongCallback has the same problem

wgengine/userspace.go
raggi: Neat! Generics!

wgengine/userspace.go
raggi: I'll do both!

cmd/tailscale/cli/ping.go
bradfitz: it's fine for now. Later we'll make a mode that by default with no flags does everything and prints a nice report.

control/controlclient/direct.go
```diff
@@ -126,9 +126,9 @@ type Options struct {
 
 // Pinger is a subset of the wgengine.Engine interface, containing just the Ping method.
 type Pinger interface {
-	// Ping is a request to start a discovery or TSMP ping with the peer handling
-	// the given IP and then call cb with its ping latency & method.
-	Ping(ip netaddr.IP, useTSMP bool, cb func(*ipnstate.PingResult))
+	// Ping is a request to start a ping with the peer handling the given IP and
+	// then call cb with its ping latency & method.
+	Ping(ip netaddr.IP, pingType string, cb func(*ipnstate.PingResult))
```
bradfitz, line 131: clarify the set of valid pingTypes for this string. (especially important as this is the interface's docs)

maybe we want a `type PingType string` somewhere onto which we can attach docs?

ipn/message.go
bradfitz: we don't generally try too hard to be compatible at this layer. (we generally don't support mix & matching versions of `cmd/tailscale` and `tailscaled`)

net/packet/icmp.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	crand "crypto/rand"
+
+	"encoding/binary"
+)
+
+// ICMPEchoPayload generates a new random ID/Sequence pair, and returns a uint32
+// derived from them, along with the id, sequence and given payload in a buffer.
+// It returns an error if the random source could not be read.
+func ICMPEchoPayload(payload []byte) (uint32, []byte) {
```
bradfitz, line 16: name results?

wgengine/userspace.go
```diff
@@ -1316,6 +1337,55 @@ func (e *userspaceEngine) mySelfIPMatchingFamily(dst netaddr.IP) (src netaddr.IP
 	return netaddr.IP{}, errors.New("no self address in netmap matching address family")
 }
 
+func (e *userspaceEngine) sendICMPEchoRequest(ip netaddr.IP, peer *tailcfg.Node, res *ipnstate.PingResult, cb func(*ipnstate.PingResult)) {
```
bradfitz, line 1340: change `ip` to `dstIP`? At first read I thought maybe `ip` was the source and `peer` was the destination, but that `peer` really isn't doing much is it?  Fine to keep. I think if the argument is `dstIP` or it had a bit of docs then it wouldn't be confusing.

wgengine/wgengine.go
```diff
@@ -154,9 +154,10 @@ type Engine interface {
...
+	// then call cb with its ping latency & method. pingType can be TSMP, ICMP,
```
bradfitz, line 158: "TSMP", "ICMP", or "disco" and ...

cmd/tailscale/cli/ping.go
bradfitz, line 53: the "Remote host OS stack only" makes sense to me but only because I've thought about this a lot.

Most users will be confused. I see it kinda matches the line above, but will they sort correctly? it might be confusing in alphabetical order.

Maybe: "send an ICMP-level ping via wireguard (bypasses local host network stack)"

bradfitz (approved):

control/controlclient/direct.go
raggi: I started doing that, but it became a cyclic import problem. I'll go back and do it and introduce a package if needs be.

cmd/tailscale/cli/ping.go
raggi: I went back and forward with some different language. I think where I landed is ok, trying to keep it terse and relatively broadly understandable.

wgengine/userspace.go
raggi: Yeah, I think this might be able to be refactored down, it's tempting to at some point pull the whole set of pings off of endpoint out to a separate object, as well because this is all very large.

PR #4448, opened 2022-04-19
cmd/proxy-to-grafana: prevent premature termination

bmizerany:
This commit changes proxy-to-grafana to report errors while polling for
tailscaled status instead of terminating at the first sign of an error.
This allows tailscale some time to come up before the proxy decides to
give up.


bradfitz:
Nice. Add the DCO thingy? `git commit --amend -s`

bmizerany:
Fixed DCO, I think.

PR #4450, opened 2022-04-19
net/dns: schedule DoH upgrade explicitly, fix Resolver.Addr confusion

bradfitz:
Two changes in one:

* make DoH upgrades an explicitly scheduled send earlier, when we come
  up with the resolvers-and-delay send plan. Previously we were
  getting e.g.  four Google DNS IPs and then spreading them out in
  time (for back when we only did URL) but then later we added DoH
  upgrading at the UDP packet layer, which resulted in sometimes
  multiple DoH queries to the same provider running (each doing happy
  eyeballs dialing to 4x IPs themselves) for each of the 4 source IPs.
  Instead, take those 4 Google/Cloudflare IPs and schedule 5 things:
  first the DoH query (which can use all 4 IPs), and then each of the
  4 IPs as UDP later.

* clean up the dnstype.Resolver.Addr confusion; half the code was
  using it as an IP string (as documented) as half was using it as
  an IP:port (from some prior type we used), primarily for tests.
  Instead, document it was being primarily an IP string but also
  accepting an IP:port for tests, then add an accessor method on it
  to get the IPPort and use that consistently everywhere.


net/dns/resolver/forwarder.go
```diff
@@ -49,6 +50,11 @@ const (
 	// arbitrary.
 	dohTransportTimeout = 30 * time.Second
 
+	// dohTransportTimeout is how much of a head start to give a DoH query
+	// that was upgraded from a well-known public DNS provider's IP before
+	// normal UDP mode is attempted as a fallback.
+	dohHeadStart = 1500 * time.Millisecond
```
danderson, line 56: 1.5s delays when DoH is sad is about as bad as DNS being down, from a UX POV. Any chance we can break this down into a first-connection-setup delay (where 1.5s is okay), vs. a request over an already connected channel (where we'd probably want to try fallbacks sooner) ?

danderson (approved):

net/dns/resolver/forwarder.go
bradfitz: I'm open to other values here.

Deciding based on whether it's connected or not is pretty tricky, though. It's less work (but still annoying) to know whether there's definitely not a connection, but when there is a TCP connection open, it's not clear whether it'll work right away or time out and work later, etc.

I'd rather we just lower this for now. But to what?

bradfitz: 500ms?

twitchyliquid64: I'm not sure if we are able to access `net.Conn` underneath all the various abstractions, but could we use `SetReadDeadline()` + `SetWriteDeadline()` to cause the request to error faster in the case we are re-using an existing connection?

twitchyliquid64 (approved):
LGTM

net/dns/resolver/forwarder.go
bradfitz: it's an http.Client -> http.Transport -> http2.Transport -> h2ConnPool -> maybe more than 1 net.Conn!

we could use the various ClientTrace hooks (https://pkg.go.dev/net/http/httptrace#ClientTrace) to see some of the lifecycle, etc.

bradfitz: changed to 500ms for now. still an eternity compared to how quickly I see Google/Cloudflare complete a cold DoH query anyway.

net/dns/resolver/forwarder.go
```diff
@@ -49,6 +50,11 @@ const (
 	// arbitrary.
 	dohTransportTimeout = 30 * time.Second
 
+	// dohTransportTimeout is how much of a head start to give a DoH query
+	// that was upgraded from a well-known public DNS provider's IP before
+	// normal UDP mode is attempted as a fallback.
+	dohHeadStart = 500 * time.Millisecond
```
maisem, line 56: Comment doesn't match

net/dns/resolver/forwarder.go
```diff
@@ -228,65 +234,55 @@ func (f *forwarder) Close() error {
...
-					} else if total4 == 1 {
-						startDelay += wellKnownHostBackupDelay
-					}
-					if done[key6] > 0 {
-						startDelay += wellKnownHostBackupDelay
-					}
-				}
-				done[hostAndFam{host, ip.BitLen()}]++
+		if host, ok := publicdns.KnownDoH()[ip]; ok {
+			// We already did the DoH query early. These
```
maisem, line 274: Incomplete sentence

net/dns/resolver/forwarder.go
```diff
@@ -228,65 +234,55 @@ func (f *forwarder) Close() error {
...
+			startDelay = dohHeadStart
+			key := hostAndFam{host, ip.BitLen()}
+			if done[key] > 0 {
+				startDelay += wellKnownHostBackupDelay
```
maisem, line 278: Should this be a multiple of the number of times?

Issue #4377, opened 2022-04-07
HTTPS DERP latency checks don't work correctly with proxy that does ssl mitm


normanr:
### What is the issue?

I'm running a client in a restricted network, but `tailscale netcheck` shows latencies for all DERP nodes (even though the restrictive proxy denied the connection request for most (but not all) of them).

If I use a non-mitm port on the proxy (which looks at the CONNECT hostname), then only the DERP hosts that are allowed return latency information. If I use a mitm port on the proxy (which accepts the CONNECT request, starts TLS, and looks at the http headers, and then returns a 503), then latencies are shown for _all_ hosts. Also looking at the proxy logs, _all_ of the requests fail because the `Host` header is set to `derp-unused-hostname.tld`.

I think two fixes would help:
- Use `node.Hostname` to construct the url in `measureHTTPSLatency` https://github.com/tailscale/tailscale/blob/858ab8017253a870a35a72676a9c83cf5fdf2bac/net/netcheck/netcheck.go#L1039 similar to how it's used in `runHTTPOnlyChecks` https://github.com/tailscale/tailscale/blob/858ab8017253a870a35a72676a9c83cf5fdf2bac/net/netcheck/netcheck.go#L978
- Return an error if `resp.StatusCode > 299`

### Steps to reproduce

Force tailscale to connect through an https_proxy that does mitm (eg: squid with ssl_bump).

example config snippet for squid:
```
http_port 3129 ssl-bump \
  generate-host-certificates=on \
  tls-cert=/etc/squid/certs/squid-ca-cert-key.pem \
  tls-cafile=/etc/squid/certs/cacert.pem \
  options=NO_SSLv3,NO_TLSv1,NO_TLSv1_1
tls_outgoing_options min-version=1.2
ssl_bump stare all      # mimic the Client Hello, drop unsupported extensions
ssl_bump bump all       # terminate and establish new TLS connection
```

Then setup acls so that only some derp servers are allowed.

### OS

Linux

### OS version

Debian testing (proxy) and Raspberry Pi (client)

### Tailscale version

1.22.2

normanr:
also when starting tailscaled it seems to pick a derp server at random, and keeps trying that single server, and never trying any other ones.

twitchyliquid64:
Can you share a little more about your setup? in particular, how are you informing tailscale about the proxy and how are you preventing traffic from hitting the WAN except through the proxy?

[...]

PR #4452, opened 2022-04-19
derp/derphttp,net/netcheck: improve netcheck behavior under MITM proxies

twitchyliquid64:
Updates #4377

In cases where tailscale is operating behind a MITM proxy, we need to consider that a lot more of the internals of our HTTP requests are visible and may be used as part of authorization checks. As such, we need to 'behave' as closely as possible to ideal.

 - Some proxies do authorization or consistency checks based the on Host header or HTTP URI, instead of just the IP/hostname/SNI. As such, we need to construct a `*http.Request` with a valid URI everytime HTTP is going to be used on the wire, even if its over TLS.
   Aside from the singular instance in net/netcheck, I couldn't find anywhere else a http.Request was constructed incorrectly.

 - Some proxies may deny requests, typically by returning a 403 status code. We should not consider these requests as a valid latency check, so netcheck semantics have been updated to consider >299 status codes as a failed probe.

### DIFF
--- a/derp/derphttp/derphttp_client.go
+++ b/derp/derphttp/derphttp_client.go
@@ -538,12 +538,17 @@ func (c *Client) tlsClient(nc net.Conn, node *tailcfg.DERPNode) *tls.Conn {
 	return tls.Client(nc, tlsConf)
 }
 
-func (c *Client) DialRegionTLS(ctx context.Context, reg *tailcfg.DERPRegion) (tlsConn *tls.Conn, connClose io.Closer, err error) {
+// DialRegionTLS returns a TLS connection to a DERP node in the given region.
+//
+// DERP nodes for a region are tried in sequence according to their order
+// in the DERP map. TLS is initiated on the first node where a socket is
+// established.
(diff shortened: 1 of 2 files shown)


derp/derphttp/derphttp_client.go
```diff
@@ -538,10 +538,10 @@ func (c *Client) tlsClient(nc net.Conn, node *tailcfg.DERPNode) *tls.Conn {
 	return tls.Client(nc, tlsConf)
 }
 
-func (c *Client) DialRegionTLS(ctx context.Context, reg *tailcfg.DERPRegion) (tlsConn *tls.Conn, connClose io.Closer, err error) {
+func (c *Client) DialRegionTLS(ctx context.Context, reg *tailcfg.DERPRegion) (tlsConn *tls.Conn, connClose io.Closer, node *tailcfg.DERPNode, err error) {
```
bradfitz, line 541: can you add some docs while you're here?

It needed docs before and it really needs docs now.

bradfitz (approved):

derp/derphttp/derphttp_client.go
bradfitz: (and full sentence too, per https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences)

PR #4456, opened 2022-04-19
tshttpproxy: support synology proxy configuration

### LINKED ISSUE #2605: synology: investigate what the HTTP proxy option does
Synology supports using an HTTP proxy but not SOCKS. What is this used for?

Screen Shot 2021-08-06 at 8 13 15 AM

In userspace mode we already support being a SOCKS5 proxy in tailscaled and we have an open tracking bug IIRC to also be an HTTP proxy in tailscaled. 

Maybe this would let people use things like Hyper Backup on Synology without TUN.

/cc @maisem @crawshaw

### LINKED ISSUE #4395: Synology client does not use proxy configuration
### What is the issue?

A customer reports that the Synology package does not use the proxy server set in the Synology configuration [HS issue 506].

### Steps to reproduce

1. Set up a Synology device on a network that blocks outgoing port 80 and 443 but has a proxy server
2. Configure Synology device to use the proxy server
3. Install Tailscale package

_No response_

### OS

Synology

### OS version

_No response_

### Tailscale version

_No response_

### Bug report

_No response_

raggi:
Fixes #4395
Fixes #2605

### DIFF
--- a/net/tshttpproxy/tshttpproxy_synology.go
+++ b/net/tshttpproxy/tshttpproxy_synology.go
@@ -1,0 +1,132 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build linux
+// +build linux
+
+package tshttpproxy
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	"tailscale.com/util/lineread"
+)
+
+// These vars are overridden for tests.
+var (
+	synologyProxyConfigPath = "/etc/proxy.conf"
+
+	openSynologyProxyConf = func() (io.ReadCloser, error) {
+		return os.Open(synologyProxyConfigPath)
+	}
+)
+
+var cache struct {
+	sync.Mutex
+	proxy   *url.URL
+	updated time.Time
+}
+
+func synologyProxyFromConfigCached(req *http.Request) (*url.URL, error) {
+	if req.URL == nil {
+		return nil, nil
+	}
+
+	cache.Lock()
+	defer cache.Unlock()
+
+	modtime := mtime(synologyProxyConfigPath)
+
+	if cache.updated == modtime {
+		return cache.proxy, nil
+	}
+
+	val, err := synologyProxyFromConfig(req)
+	cache.proxy = val
+
+	cache.updated = modtime
+
+	return val, err
+}
+
+func synologyProxyFromConfig(req *http.Request) (*url.URL, error) {
+	r, err := openSynologyProxyConf()
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	defer r.Close()
+
+	return parseSynologyConfig(r)
+}
+
+func parseSynologyConfig(r io.Reader) (*url.URL, error) {
+	cfg := map[string]string{}
+
+	if err := lineread.Reader(r, func(line []byte) error {
+		// accept and skip over empty lines
+		line = bytes.TrimSpace(line)
+		if len(line) == 0 {
+			return nil
+		}
+
+		key, value, ok := strings.Cut(string(line), "=")
+		if !ok {
+			return fmt.Errorf("missing \"=\" in proxy.conf line: %q", line)
+		}
+		cfg[string(key)] = string(value)
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	if cfg["proxy_enabled"] != "yes" {
+		return nil, nil
+	}
+
+	proxyURL := new(url.URL)
+	if cfg["auth_enabled"] == "yes" {
+		proxyURL.User = url.UserPassword(cfg["proxy_user"], cfg["proxy_pwd"])
+	}
+
+	proxyURL.Scheme = "https"
+	host, port := cfg["https_host"], cfg["https_port"]
+	if host == "" {
+		proxyURL.Scheme = "http"
+		host, port = cfg["http_host"], cfg["http_port"]
+	}
+
+	if host == "" {
+		return nil, nil
+	}
+
+	if port != "" {
+		proxyURL.Host = net.JoinHostPort(host, port)
+	} else {
+		proxyURL.Host = host
+	}
+
+	return proxyURL, nil
+}
+
+// mtime stat's path and returns it's modification time. If path does not exist,
+// it returns the unix epoch.
+func mtime(path string) time.Time {
+	fi, err := os.Stat(path)
+	if err != nil {
+		return time.Unix(0, 0)
+	}
+	return fi.ModTime()
+}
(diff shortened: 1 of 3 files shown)


net/tshttpproxy/tshttpproxy_synology.go
```diff
@@ -0,0 +1,123 @@
...
+	"os"
+	"sync"
+	"time"
+
+	"tailscale.com/util/lineread"
+	"tailscale.com/version/distro"
+)
+
+func init() {
+	if distro.Get() == distro.Synology {
```
bradfitz, line 26: rather than do this disto.Get() work at init-time, I'd prefer you just do `sysProxyEnv = linuxSysProxyEnv` and then do this `distro.Get()` lazily if/when the proxy call is ever made

net/tshttpproxy/tshttpproxy_synology.go
```diff
@@ -0,0 +1,123 @@
...
+	sync.Mutex
+	val        *url.URL
+	lastUpdate time.Time
+}
+
+func synologyProxyFromConfigCached(req *http.Request) (*url.URL, error) {
+	cachedProxy.Lock()
+	defer cachedProxy.Unlock()
+
+	if time.Since(cachedProxy.lastUpdate) < time.Minute {
```
bradfitz, line 41: time.Since isn't super well defined when working on a zero time.Time: https://go.dev/play/p/P7S_BY2iHJe

It should be considered random.

Instead: `if time.Now().Before(cachedProxy.lastUpdate.Add(time.Minute))` or something, perhaps pulling part of that out into a `tooOld` or `expiry` variable.

net/tshttpproxy/tshttpproxy_synology.go
```diff
@@ -0,0 +1,123 @@
...
+		return cachedProxy.val, nil
+	}
+
+	defer func() {
+		cachedProxy.lastUpdate = time.Now()
+	}()
```
bradfitz, line 47: is this leftover debugging where you had log prints in here earlier?

just assign this on line 51 without a defer?

net/tshttpproxy/tshttpproxy_synology.go
```diff
@@ -0,0 +1,123 @@
...
+		if len(line) == 0 {
+			return nil
+		}
+
+		parts := bytes.SplitN(line, []byte("="), 2)
+		if len(parts) != 2 {
+			return fmt.Errorf("missing \"=\" in proxy.conf line: %q", line)
+		}
+		key, value := parts[0], parts[1]
+		cfg[string(key)] = string(value)
```
bradfitz, line 90:
```go
k, v, ok := strings.Cut(string(line), "=")
if !ok {
    return ...
}
cfg[k] = v
```

net/tshttpproxy/tshttpproxy_synology.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	if cfg["proxy_enabled"] != "yes" {
+		return nil, nil
+	}
+
+	var proxyURL url.URL
```
bradfitz, line 100: proxyURL := new(url.URL)

to keep it as a pointer throughout, since that's what you're returning.

avoids accidental value copies in the future

net/tshttpproxy/tshttpproxy_synology.go
raggi: The zero value of time.Time has a documented well defined wall time, and the output seems to agree with that.

I assume the source of potential instability comes from the monotonic value, which is signed. Is there some condition in which the monotonic value on the system reads as a sub-0 time?

net/tshttpproxy/tshttpproxy_synology.go
raggi: Embarassingly, it's mental damage from too much time in raii. Anyway, defer removed.

net/tshttpproxy/tshttpproxy_synology.go
raggi: Neat, I knew this would be useful when I saw it in the notes, but I promptly forgot to use it :)

net/tshttpproxy/tshttpproxy_synology.go
raggi: I used IsZero, as I started reading through the behaviors in time post monotonic patches, and it's a bit of work I'd rather just avoid needing to think about for the Add case as well, which appears to also have monotonic portability / range concerns.

net/tshttpproxy/tshttpproxy_synology.go
```diff
@@ -0,0 +1,137 @@
...
+	sync.Mutex
+	proxy   *url.URL
+	updated time.Time
+}
+
+func synologyProxyFromConfigCached(req *http.Request) (*url.URL, error) {
+	cache.Lock()
+	defer cache.Unlock()
+
+	if !cache.updated.IsZero() && time.Since(cache.updated) < time.Minute {
```
bradfitz, line 58: I can see users getting confused that they changed their synology settings and then it doesn't work for a minute.

Isn't it cheap enough to do a Lstat of the file per call? We're about to make a network request, after all. A local lstat to the dentry cache should be lost in the noise.

net/tshttpproxy/tshttpproxy_synology.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+func init() {
+	sysProxyFromEnv = linuxSysProxyFromEnv
+}
+
+var (
+	distroOnce         sync.Once
+	distroProxyFromEnv func(*http.Request) (*url.URL, error)
+)
```
bradfitz, line 33: this is all overkill. distro.Get() is already using a sync.Once.

The it's fine to do a string comparison per call to `linuxSysProxyFromEnv`.

```go
func linuxSysProxyFromEnv(req *http.Request) (*url.URL, error) {
         return synologyProxyFromConfigCached(req)
    }
    return nil, nil
}
```

net/tshttpproxy/tshttpproxy_synology.go
```diff
@@ -0,0 +1,137 @@
...
+			return nil, nil
+		}
+		return nil, err
+	}
+	defer r.Close()
+
+	return parseSynologyConfig(r)
+}
+
+var openSynologyProxyConf = func() (io.ReadCloser, error) {
```
bradfitz, line 86: This looks very weird without a comment saying it's a `// variable for testing` or something

bradfitz (approved):

bradfitz:
It's close enough that I don't need to take another look, unless you want me to.

net/tshttpproxy/tshttpproxy_synology.go
raggi: I'd need to do a full stat, and now done.. I tend to be untrusting of vendors that they've applied appropriate flags to avoid flash wear - just paranoia not wanting to wear peoples devices out prematurely.

net/tshttpproxy/tshttpproxy_synology.go
raggi: Fair point, the once was also guarding the update, I didn't look at Get to see that it was once'd, makes sense.

bradfitz (approved):

PR #4460, opened 2022-04-20
cmd/nginx-auth: maintainer scripts and tailnet checking

Xe:
This adds the `Expected-Tailnet` header to the auth proxy. This allows you to set an expected tailnet so that you can be sure that node sharing users can't access certain internal services.

This also adds Debian and RPM maintainer scripts to the install process to ensure that systemd state is properly updated on a package update.

### DIFF
--- a/cmd/nginx-auth/nginx-auth.go
+++ b/cmd/nginx-auth/nginx-auth.go
@@ -17,6 +17,7 @@ import (
 	"net"
 	"net/http"
 	"net/netip"
+	"net/url"
 	"os"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


cmd/nginx-auth/nginx-auth.go
```diff
@@ -75,6 +75,12 @@ func main() {
 			return
 		}
 
+		if expectedTailnet := r.Header.Get("Expected-Tailnet"); expectedTailnet != "" && expectedTailnet != tailnet {
+			w.WriteHeader(http.StatusForbidden)
+			log.Printf("user is part of tailnet %s, wanted: %s", tailnet, expectedTailnet)
```
maisem, line 80: query escape both I guess

Issue #4459, opened 2022-04-20
ipv6 not forwarding through exit node properly. 


JayWStapleton:
### What is the issue?

when I use an exit node which has ipv6, I can't ping6 from my client node.

### Steps to reproduce

Exit node: fresh install debian bullseye on Vultr with ipv6 address provisioned. I can ping6 from this machine. v 1.22.2

Client: Consistent with Android (1.23.210), MacOS (1.23.207), Linux (1.22.2). 

on the exit node:
```
14:43:38.230712 tailscale0 In  IP6 fd7a:115c:a1e0:ab12:4843:cd96:6276:9a50 > 2607:f8b0:4006:808::200e: ICMP6, echo request, id 1604, seq 13, length 16
14:43:38.230781 enp1s0 Out IP6 fd7a:115c:a1e0:ab12:4843:cd96:6276:9a50 > 2607:f8b0:4006:808::200e: ICMP6, echo request, id 1604, seq 13, length 16
```
Packet is being sent "from" fd7a:115c:a1e0 address, rather than global ipv6 address of the node.
The machine's global IPv6 is 2001:19f0:5:2e7e:5400:3ff:fec0:8200/64

UFW set to ACCEPT all on ts-forward 
net.ipv6.conf.all.forwarding = 1 set 

### OS

Linux

### OS version

debian bullseye

### Tailscale version

1.22.2

### Bug report

BUG-ae1a8fdb1af6c70543fc81178c22e714fccf5cbe4ba43c86b008b8f2e7919cd2-20220420145958Z-21ea9eccc1e5bcae

bradfitz:
Copying from Slack:

> oh! I know.
> Jay has no IPv6 at home so he has no source IPv6 address to send from. So his machine uses his TS ULA address as the source and we forward it along.
> I don't think we've ever done v6 masquerading
> That is: Tailscale exit nodes can't give a non-v6 machine v6 access.
> which is both sad and exciting that we get to fix/enable this

danderson:
From slack:

We have a couple ways we could solve this:
 - Fully principled: implement NAT64 (NAT from IPv6 to IPv4) in userspace. The kernel doesn't implement NAT64, so we'd have to do it ourselves. Reference implementations to learn from are be TAYGA, Jool and WrapSix. This is fairly tricky to get right, and IMO not great value for money right now. 
[...]

PR #4461, opened 2022-04-20
wgengine/router: make supportsV6NAT check catch more cases

bradfitz:
Updates #4459


danderson (approved):
Fine I guess, but I'll repeat my question from the other place: do you remember why we try to figure this out through heuristics, rather than just running the iptables command and deciding if stuff works based on that? Running the command should autoload the appropriate modules if missing, should work on legacy iptables or nftables, etc. etc.

Was there a reason we didn't document why we're playing whackamole with unstable kernel interfaces instead?

bradfitz:
@danderson, filed #4462 for ongoing cleanup.

Issue #4449, opened 2022-04-19
Taildrop to Android TV fails, peerAPI accept fails


DentonGentry:
### What is the issue?

Telemetry as the Android TV device is starting up says the peerAPI failed. It is complaining about accept, not listen, which I don't understand:

```
2022-04-19 15:36:31.254779454 +0000 UTC: 9.6M/136.3M peerapi.Accept: accept tcp6 [fd7a:115c:a1e0:ab12:4843:cd96:625e:1003]:39235: accept: invalid argument
2022-04-19 15:36:31.264521163 +0000 UTC: 9.6M/136.3M peerapi.Accept: accept tcp4 100.94.16.3:39235: accept: invalid argument
```

When I try to send a file via taildrop to the Android TV device the ACL allows it, but taildrop on the sending device shows failure immediately:
```
2022-04-19 17:34:47.698071922 +0000 UTC: 8.6M/152.5M Accept: TCP{100.100.254.110:61794 > 100.94.16.3:39235} 64 tcp ok
2022-04-19 17:34:47.708856213 +0000 UTC: 8.6M/152.5M Accept: TCP{100.94.16.3:39235 > 100.100.254.110:61794} 40 ok out
```

### OS

Android

### OS version

Android 10

### Tailscale version

1.23.210-tc13be0c50-g3f8df48d23d

### Bug report

Cannot easily so do on Android TV. LogID is b29366e3a3852e6d03091af1fb9cd98a28402e0c2c298ac93e4a00739aa2926e, timestamps shown in telemetry snippets above.

bradfitz:
Please clarify which direction this is. The PR title is confusion. 

I think this is a dup of https://github.com/tailscale/tailscale/issues/4293#issuecomment-1090715082 if this is about sending TO Android.

bradfitz:
https://github.com/tailscale/tailscale/pull/4465 is submitted and should fix this for 1.24.0. Keeping open until confirmed.

DentonGentry:
I built a 1.23.230 Android build at 13f75b9667c0522323b7b4c8a1762debbf0e7866, which includes https://github.com/tailscale/tailscale/commit/1b89662eff16397d77be4634f114f12f0e2967f9, and tried to taildrop a file from a macOS system, which failed and logged:
`default	20:26:17.948165-0700	IPNExtension	http: proxy error: dial tcp 100.75.181.70:1: connect: connection refused`

When the Android app starts I see it listening on port 1:
```
[...]

Issue #4293, opened 2022-03-27
Taildrop: 502 Bad Gateway attempting to send a file from a Linux server to Android


GOATS2K:
### What is the issue?

`tailscale file cp --verbose loner.mp4 pixel-6-pro:`

```
2022/03/27 01:49:06 sending "loner.mp4" to pixel-6-pro/100.110.71.64/ngH2ag7CNTRL ...
502 Bad Gateway:
```

I expected the file to be sent.

### Steps to reproduce

I ran the `tailscaile file cp` command on my server, running Ubuntu 20.04.

Attempted sending the file to a Pixel 6 Pro, running Android 12.

### OS

Linux, Android

### OS version

Ubuntu 20.04

### Tailscale version

1.22.2

### Bug report

BUG-a553851f8705625b010d90125777946fb62e76eb1e8a768368bed8b1b8e1dc1d-20220327020308Z-ab2e23156be1b736

bradfitz:
[Notes to self]

on the sending side:
```
2022-03-27 01:49:07.021133382 +0000 UTC: http: proxy error: dial tcp 100.110.71.64:42858: connect: connection refused
```

on the receiving side (c809a371970690b9dd6740b0b6caca05bf07896398c8747e102d1e513dbd406f) a bit earlier, perhaps from other attempt:
```
2022-03-27 01:18:16.815868906 +0000 UTC: 12.8M/236.5M peerapi.Accept: accept tcp6 [fd7a:115c:a1e0:ab12:4843:cd96:626e:4740]:42858: accept: invalid argument
2022-03-27 01:18:16.824863331 +0000 UTC: 12.8M/236.5M peerapi.Accept: accept tcp4 100.110.71.64:42858: accept: invalid argument
```

In bd90781b34d2558f27196194ba871f1cda60ffe6 we started using netstack for peerapi, though, for everywhere but iOS, so that should include Android.

bradfitz:
@GOATS2K, if this is reproducible, could you tell me what this says from Linux...

First, `tailscale ping -tsmp 100.110.71.64` and you'll see something like:

```
pong from pixel-6-pro (100.110.71.64, 42858) via TSMP in 100ms
```

That number after the comma is the port number (which might change over time) of the "peerapi" service that Taildrop uses.

Then try to curl that port number...

```
$ curl -v http://100.110.71.64:42858
```

What does that say?

GOATS2K:
Oddly enough, file transfers work now!

```
[...]

Issue #3986, opened 2022-02-20
Android Exit Node not passing traffic


cyrus104:
### What is the issue?

I'm testing Tailscale in my homelab and have several subnets. The VMs are able to talk via a firewall rule / route on an OPNsense router. One device is an Pixel 6 Pro that I want to use as the exit node (using the debug secret function). The other device is an Ubuntu VM on a different VLAN.

After making my first mistake, I enabled the exit node in the web interface which got me to here.

I have tested a tailscale ping of 8.8.8.8 and a local network accessible to the Android only `10.*.40.4` and `10.*.10.4` (a different subnet but only accessible from the vlan the phone is on. It's blocked on the vlan the vm is on. I'm using this as a test. The output for both are the same:
pong from pixel via DERP(sea) in 535ms (normal for me right) 
Interesting when I regular ping the time is 3 times 

Android Setup:
VLAN: 10.*.40.157/24

Ubuntu VM Setup:
192.168.*.19/24

tailscale ping 100.100.100.100
pong from pixel via DERP(sea) in 535ms (normal for me right)

dig@100.100.100.100 google.com
1 server found
global options; +cmd
connection timed out; no servers could be reached

### OS

Linux, Android

### OS version

Ubuntu 21.04, Android 12

### Tailscale version

1.20.4, 1.20.2

### Bug report

BUG-4ab0854226c742e3d4ca193566696dc1ac8d667175807e29f0706ed8a7657da1-20220220023919Z-49240ffdcfa6270f

bradfitz:
> tailscale ping 100.100.100.100

That command doesn't make sense, as 100.100.100.100 doesn't exist anywhere on the network. It's a virtual IP that runs on every Tailscale node. Did you really run that, or did you make that up while trying to redact your Tailscale IP?

Your Tailscale IPs are safe to share publicly. Nobody can do anything with them. Yours are 100.123.6.37 (Ubuntu) and 100.114.213.119 (Android). What does `tailscale ping 100.114.213.119` from your Ubuntu box say? I suspect it'll still use DERP because your router doesn't seem to support hairpinning.

Your Android phone seems to be changing DERP homes all the time. Is there some other VPN/proxy involved here?

cyrus104:
[...]

PR #4465, opened 2022-04-20
ipn/ipnlocal: make peerapi listener on Android avoid the kernel

bradfitz:
We intercept the peerapi port in netstack anyway, so there's no reason
the linux kernel on Android needs to know about it. It's only getting
in the way and causing problems for reasons we don't fully understand.
But we don't even need to understand it because it's not relevant
anymore.

Instead, provide a dummy net.Listener that just sits and blocks to
pacify the rest of the code that assumes it can be stuck in a
Listener.Accept call and call Listener.Close and Listener.Addr.

We'll likely do this for all platforms in the future, if/when we also
link in netstack on iOS.

Updates #4449
Updates #4293
Updates #3986

### DIFF
--- a/ipn/ipnlocal/peerapi.go
+++ b/ipn/ipnlocal/peerapi.go
@@ -386,6 +386,14 @@ func (s *peerAPIServer) OpenFile(baseName string) (rc io.ReadCloser, size int64,
 }
 
 func (s *peerAPIServer) listen(ip netaddr.IP, ifState *interfaces.State) (ln net.Listener, err error) {
+	// Android for whatever reason often has problems creating the peerapi listener.
+	// But since we started intercepting it with netstack, it's not even important that
+	// we have a real kernel-level listener. So just create a dummy listener on Android
+	// and let netstack intercept it.
+	if runtime.GOOS == "android" {
+		return newFakePeerAPIListener(ip), nil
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/peerapi.go
```diff
@@ -1047,3 +1055,32 @@ func writePrettyDNSReply(w io.Writer, res []byte) (err error) {
...
+func newFakePeerAPIListener(ip netaddr.IP) net.Listener {
+	return &fakePeerAPIListener{
+		addr:   netaddr.IPPortFrom(ip, 1).TCPAddr(),
+		closed: make(chan struct{}),
+	}
+}
+
+// fakePeerAPIListener is a net.Listener that has an Addr method returning a TCPAddr
+// for a given IP on port 1 (arbitrary) and can be Closed, but otherwise Accept
+// just blocks forever until closed.
```
maisem, line 1068: why block forever? add comment on what that achieves

maisem (approved):

ipn/ipnlocal/peerapi.go
bradfitz: added big comment

maisem (approved):

PR #4466, opened 2022-04-20
wgengine/monitor: ignore OS-specific uninteresting interfaces

maisem:
Currently we ignore these interfaces in the darwin osMon but then would consider it
interesting when checking if anything had changed.

### DIFF
--- a/wgengine/monitor/monitor.go
+++ b/wgengine/monitor/monitor.go
@@ -316,8 +316,9 @@ func (m *Mon) debounce() {
 				m.ifState = curState
 
 				if s1, s2 := oldState.String(), curState.String(); s1 == s2 {
-					m.logf("[unexpected] network state changed, but stringification didn't: %v\nold: %s\nnew: %s\n", s1,
-						jsonSummary(oldState), jsonSummary(curState))
+					m.logf("[unexpected] network state changed, but stringification didn't: %v", s1)
+					m.logf("[unexpected] old: %s", jsonSummary(oldState))
+					m.logf("[unexpected] new: %s", jsonSummary(curState))
 				}
 			}
 			// See if we have a queued or new time jump signal.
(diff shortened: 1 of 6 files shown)


wgengine/monitor/monitor_linux.go
```diff
@@ -64,6 +64,8 @@ func newOSMon(logf logger.Logf, m *Mon) (osMon, error) {
 	return &nlConn{logf: logf, conn: conn, addrCache: make(map[uint32]map[netaddr.IP]bool)}, nil
 }
 
+func (c *nlConn) IsInterestingInterface(_ string) bool { return true }
```
bradfitz, line 67: `iface string` is fine. looks less mysterious

wgengine/monitor/polling.go
```diff
@@ -38,6 +38,10 @@ type pollingMon struct {
 	stop      chan struct{}
 }
 
+func (pm *pollingMon) IsInterestingInterface(_ string) bool {
```
bradfitz, line 41: same

wgengine/monitor/monitor.go
```diff
@@ -282,6 +286,10 @@ func (m *Mon) notifyRuleDeleted(rdm ipRuleDeletedMessage) {
 	}
 }
 
+func (m *Mon) useInterestingInterfaces(i interfaces.Interface, ips []netaddr.IPPrefix) bool {
```
bradfitz, line 289: docs on this method now.

Or just call it `isInterfaceInteresting` and clarify that `ips` is its current IPs.

bradfitz (approved):

wgengine/monitor/monitor.go
```diff
@@ -42,6 +42,10 @@ type osMon interface {
 	// until the osMon is closed. After a Close, the returned
 	// error is ignored.
 	Receive() (message, error)
+
+	// IsInterestingInterface reports whether the provided interface is
+	// considered interesting.
```
bradfitz, line 47: pretty tautological docs here.

what's it used for?

bradfitz (approved):

PR #4469, opened 2022-04-20
ssh/tailssh: support expansions in public key fetch URL too

bradfitz:
Updates #3802

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -541,6 +541,22 @@ func (srv *server) expandDelegateURL(ci *sshConnInfo, lu *user.User, actionURL s
 	).Replace(actionURL)
 }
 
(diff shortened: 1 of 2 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -541,6 +541,22 @@ func (srv *server) expandDelegateURL(ci *sshConnInfo, lu *user.User, actionURL s
...
+		return pubKeyURL
+	}
+	var localPart string
+	var loginName string
+	if ci.uprof != nil {
+		loginName = ci.uprof.LoginName
+		localPart, _, _ = strings.Cut(loginName, "@")
+	}
+	return strings.NewReplacer(
+		"$LOGINNAME_EMAIL", loginName,
```
maisem, line 555: I don't like the names but it's all internal so sure

maisem (approved):

PR #4481, opened 2022-04-21
util/kmod: call unix.Capget directly, without kernel.org/pub/linux/libs/security/libcap/cap

bradfitz:
The kernel.org/pub/linux/libs/security/libcap/cap or its dependencies somehow
broke the Android build, bringing in cgo stuff or something.

It was quicker to remove the dependency than debug.


raggi (approved):

bradfitz:
Nevermind. Reverting prior change instead for now.

PR #4485, opened 2022-04-21
.github/workflows: add cross-android

raggi:


bradfitz:
You forgot the `#` sign in the commit message so it doesn't cross-ref between the issue.

But also this isn't a fix. If it's not running `gio` to do the build and making an *.apk, it's not enough yet.

This is fine too, but keep the commit message saying `Updates #nnnn` for now.

raggi:
I made the PR to see the action run, sadly draft PRs don't run actions. I didn't intend for it to be ready yet - I wanted to see what the outcome would be given the environment and toolchain. I'm not sure this produces useful signal - it wouldn't fail on the kmod case as-is - the linker seems to ignore the missing cgo.

PR #4486, opened 2022-04-21
shell.nix: Update go toolchain

twitchyliquid64:
Maybe in the future we should have `go.toolchain.rev` files like we have in control, and then (given control already pulls in OSS) controls `shell.nix` can read the values from OSS. Just an idea.


bradfitz:
> Maybe in the future we should have go.toolchain.rev files like we have in control, and then (given control already pulls in OSS) controls shell.nix can read the values from OSS. Just an idea.

Go for it.

bradfitz (approved):

PR #4488, opened 2022-04-21
net/dns/resolver: support magic resolution of via-<siteid>.<ip4> domains

twitchyliquid64:
For now, I've opted not to do the inverse (reverse-lookup on these IPs into these via domain names) because I think that would look weird in stuff like traceroute - but I can add it in if you like.

### DIFF
--- a/net/dns/resolver/tsdns.go
+++ b/net/dns/resolver/tsdns.go
@@ -17,6 +17,7 @@ import (
 	"os"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -633,6 +634,10 @@ func (r *Resolver) resolveLocal(domain dnsname.FQDN, typ dns.Type) (netaddr.IP, 
 			return tsaddr.TailscaleServiceIPv6(), dns.RCodeSuccess
 		}
 	}
+	// Special-case: 'via-<siteid>.<ipv4>' queries.
+	if ip, ok := r.parseViaDomain(domain, typ); ok {
+		return ip, dns.RCodeSuccess
+	}
 
 	r.mu.Lock()
 	hosts := r.hostToIP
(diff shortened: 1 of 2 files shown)


net/dns/resolver/tsdns_test.go
```diff
@@ -348,6 +348,9 @@ func TestResolveLocal(t *testing.T) {
 		{"ns-nxdomain", "test3.ipn.dev.", dns.TypeNS, netaddr.IP{}, dns.RCodeNameError},
 		{"onion-domain", "footest.onion.", dns.TypeA, netaddr.IP{}, dns.RCodeNameError},
 		{"magicdns", dnsSymbolicFQDN, dns.TypeA, netaddr.MustParseIP("100.100.100.100"), dns.RCodeSuccess},
+		{"via_hex", dnsname.FQDN("via-0xff.1.2.3.4."), dns.TypeAAAA, netaddr.MustParseIP("fd7a:115c:a1e0:b1a:0:ff:102:304"), dns.RCodeSuccess},
```
bradfitz, line 351: oh, I assumed these would be rooted at the MagicDNS suffix, without the "." at the end here.

so the test input would be "via-0xff.1.2.3.4.corp.beta.tailscale.net." or something.

net/dns/resolver/tsdns.go
```diff
@@ -708,6 +702,57 @@ func (r *Resolver) resolveLocal(domain dnsname.FQDN, typ dns.Type) (netaddr.IP,
...
+			return tsaddr.TailscaleServiceIPv6(), true
+		}
+	}
+
+	// As a convenient mapping into Tailscales 'Via Range', we support
+	// resolving any queries of the form 'via-<X>.<IPv4-address>' into
+	// their corresponding IPv6 address, where X is a base 10 or base 16
+	// siteID.
+	fqdn := string(domain)
+	if firstDot := strings.Index(fqdn, "."); len(fqdn) > 10 && firstDot > 0 && strings.HasPrefix(fqdn, "via-") {
```
bradfitz, line 727: that 10 is mysterious.  if the goal is to do something cheap early, it's kinda too late as you've already called strings.Index.

I'd pull this into a func where you can return early a bunch instead of making this a bunch of nested ifs indenting far.

```go 
...
      if typ == dns.TypeAAAA {
          if ip, ok := parseViaName(domain); ok {
                 return ip, true
          }
      }
      // TODO: other magic in the future?
      return  netaddr.IP{}, false
}
```

Then like:

```go
func parseViaName(domain dnsname.FQDN) (ip netaddr.IP, ok bool) {
     s := string(domain)
     if len(s) < len("via-X.0.0.0.0") {
            return
      }
      if !strings.HasPrefix(s, "via-") {
            return
      }
      s = s[len("via-"):]
      ....
```

(Note that `len("via-")`compiles to 4)
```

File: net/dns/resolver/tsdns_test.go (hunk 1 above)
Comment (bradfitz) [reply net/dns/resolver/tsdns_test.go:351]: Let's make it work both ways, rooted or not?

File: net/dns/resolver/tsdns.go (hunk 2 above)
Comment (twitchyliquid64) [reply net/dns/resolver/tsdns.go]: Ahhhh thats so much nicer!

File: net/dns/resolver/tsdns_test.go (hunk 1 above)
Comment (twitchyliquid64) [reply net/dns/resolver/tsdns_test.go:351]: I went on a bit of a rabbithole here, as I couldnt get DNS queries to come in using the search domains even though they are correctly configured. I discovered that:

 - A domain with any dot is considered fully qualified, so search domains arent consulted
 - Its technically not valid to have domains that are just numbers, so this scheme is technically invalid.

As such, it only works rootless, and everyone except us would just reject the query anyway.

File: net/dns/resolver/tsdns.go [hunk 3]
```diff
@@ -708,6 +714,53 @@ func (r *Resolver) resolveLocal(domain dnsname.FQDN, typ dns.Type) (netaddr.IP,
...
+		return netaddr.IP{}, false // badly formed, dont respond
+	}
+	suffix := ip4.As4()
+
+	var prefix uint64
+	if len(siteID) > 2 && siteID[:2] == "0x" {
+		prefix, err = strconv.ParseUint(siteID[2:], 16, 32)
+	} else {
+		prefix, err = strconv.ParseUint(siteID, 10, 32)
+	}
```
bradfitz, line 753: you don't need this conditional. Just use ParseUint(..., 0, 32)

the 0 means automatic.

net/dns/resolver/tsdns.go
```diff
@@ -708,6 +714,53 @@ func (r *Resolver) resolveLocal(domain dnsname.FQDN, typ dns.Type) (netaddr.IP,
...
+	if err != nil {
+		return netaddr.IP{}, false // badly formed, dont respond
+	}
+
+	ip := tsaddr.TailscaleViaRange().IP().As16()
+	binary.BigEndian.PutUint32(ip[8:12], uint32(prefix))
+	copy(ip[12:], suffix[:])
```
bradfitz, line 760: there's a func in tsaddr to do all this

return tsaddr.MapVia(uint32(prefix), netaddr.IPPrefixFrom(ip4, ip4.BitLen()))

bradfitz (approved):

PR #4495, opened 2022-04-21
cmd/tailscale/cli: do not use syscall.Exec from macOS sandbox

maisem:


bradfitz (approved):

bradfitz:
sigh

PR #4496, opened 2022-04-22
cmd/tailscale: use double quotes in the ssh subcommands

maisem:
Single-quote escaping is insufficient apparently.

Updates #3802


bradfitz (approved):
Suuure

PR #4497, opened 2022-04-22
ssh/tailssh: avoid user ssh configuration in tests

raggi:
The tests were hanging for me, probably waiting for `ssh` to exit, but it may not under some conditions, such as if it's matched my broad ControlMaster config. Avoiding the user config will avoid any such test unpredictability.

### DIFF
--- a/ssh/tailssh/tailssh_test.go
+++ b/ssh/tailssh/tailssh_test.go
@@ -265,6 +265,8 @@ func TestSSH(t *testing.T) {
 
 	execSSH := func(args ...string) *exec.Cmd {
 		cmd := exec.Command("ssh",
+			"-F",
+			"none",
 			"-v",
 			"-p", fmt.Sprint(port),
 			"-o", "StrictHostKeyChecking=no",


ssh/tailssh/tailssh_test.go
```diff
@@ -265,6 +265,8 @@ func TestSSH(t *testing.T) {
 
 	execSSH := func(args ...string) *exec.Cmd {
 		cmd := exec.Command("ssh",
+			"-F",
+			"none",
```
maisem, line 269: /dev/null?

maisem (approved):

ssh/tailssh/tailssh_test.go
raggi: No, because that means I have to check if `NUL:` works on with ssh on Windows, and I was being lazy.

`none` is an ssh magic value.

raggi: Well, turns out `NUL:` works too, but like I said, `none` is a well-known value to ssh meaning no config:

```
C:\Users\raggi>ssh -F foo ...
Can't open user config file foo: No such file or directory

C:\Users\raggi>ssh -F none rag.pub
Warning: the ECDSA host key for '...' differs from the key for the IP address '...'
```

PR #4500, opened 2022-04-22
ipn/store: add GCP secrets manager store

terorie:
- adds ipn/store/gcpstore package using GCP secret manager backend
- adds `gcp:` as supported prefix for state path
- deduplicates definitions of `testStoreSemantics` using new storetest package

### Motivation

Tailscale supports persistent state backends for Kubernetes (`kube:`) and AWS (`arn:`) but not for GCP.

When deploying Tailscale as a proxy using single-node Compute Engine instance groups similarly to the [k8s example](https://github.com/tailscale/tailscale/tree/main/docs/k8s), the only choice is the in-memory state backend.
This causes machines to lose their IP addresses on VM replacement and get a different hostname like `<machine>-1`.

Using the gcpstore backend, Tailscale can store state in [GCP secret manager](https://cloud.google.com/secret-manager) which allows secrets up to 64KiB.

VMs on GCP can then start the Tailscale daemon like so:
```
tailscaled --state="gcp:projects/my-project/secrets/tailscale-state-my-vm"
```


DentonGentry:
When you're ready for review, please mark the PR as not being a Draft.

terorie:
> When you're ready for review, please mark the PR as not being a Draft.

Yes, will do, after testing in a real environment.

> I took a quick look but then saw it doesn't compile yet.
> 
> Also, `make updatedeps` shows that it adds quite a few dependencies to the project

@bradfitz This amount of new deps looks hard to justify for such a small feature. 😔 It will only get worse when support for more clouds is needed.

The Prometheus and Terraform projects were facing a similar problem with service discovery and state backends respectively – lots of vendors to support but it was unrealistic to maintain all of them in the main tree.
Both of them opted for generic file-based and HTTP-based backends. When folks need a new vendor they can just add a sidecar that monitors a file or accepts HTTP requests without modifying the core software.

So, do you think a "httpstore" (using GET/POST) or "fsstore" (using JSON file) for IPN is a better alternative? If so, which one do you prefer?

bradfitz:
httpstore HTTP sidecar sounds good to us.

Issue #4276, opened 2022-03-24
Intermittent DNS resolution failure on macOS using open source installation


djmarcin:
### What is the issue?

We get intermittent failures to resolve a DNS name via Magic DNS. This occurs both when using the short name and FQDN.

### Steps to reproduce

Our setup is:
- Install via https://github.com/tailscale/tailscale/wiki/Tailscaled-on-macOS. We use networksetup commands to set the dns servers and search domains. We also have an additional search domain for our internal services.
```
TAILSCALE_VERSION=v1.22.2
go install -v tailscale.com/cmd/tailscale{,d}@"$TAILSCALE_VERSION"
sudo tailscaled install-system-daemon
tailscale up --authkey "$AUTH_KEY" --accept-routes
sudo networksetup -setdnsservers Ethernet 100.100.100.100 8.8.8.8 8.8.4.4
sudo networksetup -setsearchdomains Ethernet metawork.com.beta.tailscale.net some.internal.domain
```

CI jobs run on these machines and they frequently (several times a day) fail with an error indicating they cannot resolve another tailscale host:
```
ERROR: java.io.IOException: io.grpc.StatusRuntimeException: UNAVAILABLE: Unable to resolve host buildkite-macos-x86-64-230eff.metawork.com.beta.tailscale.net.
```

This happens whether we use a hostname that should be found on the search domain like `buildkite-macos-x86-64-230eff` or a FQDN like `buildkite-macos-x86-64-230eff.metawork.com.beta.tailscale.net.`. This only seems to happen on our macOS hosts though, we have never seen this error on linux hosts. This happens on both arm64 and x86_64 machines.

The tailscale bugreport is from a machine that failed DNS resolution, but about 20 minutes prior to the `tailscale bugreport` command. The approximate times of failure was 10:48PT and 10:52PT

### Are there any recent changes that introduced the issue?

We recently updated from 1.14.4 where we were not having this issue to 1.22.2. I am not certain this was the only change to our systems in this time, though.

### OS

macOS

### OS version

Darwin buildkite-macos-arm64-8f2450.local 20.3.0 Darwin Kernel Version 20.3.0: Thu Jan 21 00:06:51 PST 2021; root:xnu-7195.81.3~1/RELEASE_ARM64_T8101 arm64

### Tailscale version

1.22.2

### Bug report

[...]

PR #4508, opened 2022-04-23
net/dns: add start of a tailscaled-on-macOS DNS OSConfigurator

bradfitz:
This populates DNS suffixes ("ts.net", etc) in /etc/resolver/* files
to point to 100.100.100.100 so MagicDNS works.

It also sets search domains.

Updates #4276

### DIFF
--- a/net/dns/manager_darwin.go
+++ b/net/dns/manager_darwin.go
@@ -1,0 +1,127 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package dns
+
+import (
+	"bytes"
+	"errors"
+	"os"
+
+	"go4.org/mem"
+	"tailscale.com/types/logger"
+	"tailscale.com/util/mak"
+)
+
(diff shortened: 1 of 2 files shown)


crawshaw (commented):
I don't see how a NetworkExtension can edit /etc, and I don't understand what stops this code from being called when running under the NetworkExtension.

bradfitz:
> I don't see how a NetworkExtension can edit /etc, 

Read the commit message title. This is tailscaled-on-macOS, not Network Extension. This runs as root.

> and I don't understand what stops this code from being called when running under the NetworkExtension.

Because the GUI versions call NewUserspaceEngine with DNS set to a router.CallbackRouter. They don't use this code path.

Note how this code for Darwin previously use a `NewNoopManager()`. Obviously this code wasn't being used in the GUI Mac version before with a no-op manager.

bradfitz:
Updated to also set search suffixes.

crawshaw (approved):
It would be nice if we could figure out how to make it easy to spot when opening a file which deployment variant it’s for. I see now there is a reference in the comment on the type, but it is still going to take me a couple minutes next time I come back to this code.

PR #4513, opened 2022-04-25
ipn/store: add HTTP store

terorie:
- adds ipn/store/httpstore package, which supports storing IPN state using a generic HTTP API
- adds `http:` and `https:` as supported prefixes for state path
- deduplicates definitions of testStoreSemantics using new storetest package

Continuation of https://github.com/tailscale/tailscale/pull/4500


terorie:
Please let me know how we should go about adding docs for the HTTP API. I can provide a knowledge base article in Markdown format or an example repo.

ipn/store/httpstore/store_http.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+// httpStore is a HTTP-backed store with in-memory write-through caching.
+type httpStore struct {
+	httpClient *http.Client
+	requestURL string
+	memory     mem.Store
+}
+
+const contentType = "application/vnd.tailscale.ipn-state+json"
```
terorie, line 47: Need confirmation from Tailscale maintainers whether it's ok to use this MIME type

terorie:
We're switching away from Tailscale, so we don't need this patch anymore. Feel free to create a new PR based on my changes if needed.

PR #4519, opened 2022-04-25
wgengine/monitor: do not ignore changes to pdp_ip*

maisem:

### DIFF
--- a/wgengine/monitor/monitor_darwin.go
+++ b/wgengine/monitor/monitor_darwin.go
@@ -115,7 +115,8 @@ func addrType(addrs []route.Addr, rtaxType int) route.Addr {
 func (m *darwinRouteMon) IsInterestingInterface(iface string) bool {
 	baseName := strings.TrimRight(iface, "0123456789")
 	switch baseName {
-	case "llw", "awdl", "pdp_ip", "ipsec":
+	// TODO(maisem): figure out what this list should actually be.
+	case "llw", "awdl", "ipsec":
 		return false
 	}
 	return true


bradfitz (changes requested):
Insufficiently justified commit message and no bug reference.

Doesn't answer "Why?"

crawshaw (approved):
Add a comment mentioning we want pdp_ip?

Googled around and stack overflow says:

| ap1, en0, en1 are names of the interfaces on iOS as well as on Mac. If you type in Terminal on Mac ifconfig you would get the same, en0, en1, etc.
| pdp_ip interfaces are those that are used for 3G and cellular data, while ap1 is used to represent currently active data connection, Wi-Fi, cellular data or bluetooth.

So I'm back to thinking, if our connection defaults over the ap1 interface, it's iOS's job to stop us running a TCP connection over LTE when we're on wifi.

maisem:
@crawshaw looking at logs the defaultRoute we pick is always `en0` or `pdp_ip0`

maisem:
> Insufficiently justified commit message and no bug reference.
> 
> Doesn't answer "Why?"

Fixed

bradfitz:
> Our current theory on battery consumption is that magicsock is resorting

One current theory (among other things) ... 

> to using the IPv6 over LTE even on WiFi. One thing that could explain
> this is that we do not get link change updates for the LTE modem as we
> ignore then in this list. This commit makes us not ignore changes to

typo: them

> `pdp_ip` as a test.

PR #4520, opened 2022-04-25
tstest/integration/vms: use hujson.Standardize instead of hujson.Unmarshal

dsnet:
The hujson package transition to just being a pure AST
parser and formatter for HuJSON and not an unmarshaler.

Thus, parse HuJSON as such, convert it to JSON,
and then use the standard JSON unmarshaler.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/go-ole/go-ole v1.2.6
 	github.com/godbus/dbus/v5 v5.0.6
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da
-	github.com/google/go-cmp v0.5.7
+	github.com/google/go-cmp v0.5.8
 	github.com/google/uuid v1.3.0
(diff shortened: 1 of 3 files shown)


bradfitz:
Okay, but why? Why not keep an Unmarshal method in hujson that does exactly those 4 lines so everybody doesn't need to do that?

dsnet:
This is the only usage of `hujson.Unmarshal` in our entire code-base. If we do this, we can delete the entire fork of `encoding/json`.

bradfitz:
But why can't `hujson.Unmarshal` call std `encoding/json.Unmarshal`?

dsnet:
We could, but that helper function is going to be using current `encoding/json` package. Hypothetically speaking, there could be a new `json` package in the future (😏). I'm a little uncomfortable adding a helper that implicitly makes assumptions about how unmarshaling works.

bradfitz (approved):
Still not a fan of this API usability regression, but not enough to object more.

Issue #4522, opened 2022-04-25
Postgres connection leaks


zenazn:
### What is the issue?

One of the primary uses my organization relies on Tailscale for is to connect to our production Postgres database from our employee laptops. We commonly use `psql` and a MacOS GUI named [TablePlus](https://tableplus.com/). Our database runs inside [Render](https://render.com/), where we run a subnet router (currently version 1.22.2), I believe in userspace mode (but I might be making that up—the deployment is based on [this](https://github.com/render-examples/tailscale))

Twice now, we've encountered issues with Tailscale-mediated connections to the database piling up, resulting in us saturating the maximum number of connections the database allows. The number of "extra" connections is far in excess of the number of connections we expect, and has the classic "sawtooth" pattern of a resource leak:

<img width="529" alt="Screen Shot 2022-04-25 at 4 12 16 PM" src="https://user-images.githubusercontent.com/12960/165189240-e7e14674-82c4-4e66-9fba-01f8f5386129.png">

Our Postgres server is configured to use TCP keepalives to detect idle clients that have disappeared. I forget the exact settings, but they're pretty reasonable (~10m of inactivity, 5 probes every minute or similar). The leaked connections are all in state `idle` (and not `idle in transaction` or similar), so I'd expect the TCP keepalives to be an effective solution on an "ordinary" network.

Frustratingly for both of us, I'm not certain this is a Tailscale bug! But given the centrality of networking to my understanding of the problem, I figured I'd open a ticket anyways and see if you folks have any ideas (or have heard similar rumblings from other customers). As always, happy to provide more detail if needed. Given the slow speed of the leak, I obviously don't have much by means of reproducing test cases or smoking guns, but happy to do some legwork to collect data if that's helpful.

### OS

Linux, macOS

### Tailscale version

1.22.2

### Bug report

BUG-1b1b6d797befc704c22048932ffcb72ce2677f8c2c08667f3bf69a4a4ee1ae83-20220425231410Z-d0c5b4f2ec891f08

DentonGentry:
It is a lot easier to imagine how tailscale might be breaking Postgres connections than it is to imagine keeping connections alive longer than expected.

When functioning as a subnet router I believe userspace-networking is forwarding packet-by-packet. It shouldn't be terminating the TCP connection on one side and making a new connection on the other, which is how one might imagine it could absorb TCP keepalives without passing them on to the other side.

[...]

PR #4534, opened 2022-04-26
wgengine/netstack: close forwarded TCP connections when incoming TCP dies

bradfitz:
Updates #4522

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -734,11 +734,15 @@ func (ns *Impl) forwardTCP(client *gonet.TCPConn, clientRemoteIP netaddr.IP, wq 
 		_, err := io.Copy(client, server)
 		connClosed <- err
 	}()
-	err = <-connClosed
-	if err != nil {
-		ns.logf("proxy connection closed with error: %v", err)
+	select {
+	case err := <-connClosed:
+		if err != nil {
+			ns.logf("proxy connection closed with error: %v", err)
+		}
+		ns.logf("[v2] netstack: forwarder connection to %s closed", dialAddrStr)
+	case <-ctx.Done():
+		ns.logf("[v2] netstack: context done, closing TCP forward conn to %s", dialAddrStr)
 	}
-	ns.logf("[v2] netstack: forwarder connection to %s closed", dialAddrStr)
 }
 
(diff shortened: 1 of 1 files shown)


bradfitz:
@raggi, is this right?

When does the `notifyCh` fire from here:

```
waitEntry, notifyCh := waiter.NewChannelEntry(waiter.EventHUp) // TODO(bradfitz): right EventMask?
```

DentonGentry (approved):
LGTM to me, it will let it not be blocked waiting for a <-connClosed which will never come.

bradfitz:
I'm just not sure when `notifyCh` fires so I'm afraid it might be too soon and break active connections.

raggi:
> waiter.NewChannelEntry(waiter.EventHUp)

This is essentially the same as POLLHUP so a relatively catch-all for disconnection, but it does not imply EOF (more on that below). It is the right signal.

Good catch, this is a subtle case where not enough data is flowing to cause a teardown from the data plane.

I'm wary of dropping tail buffers due to the eager closes. Right now I think because this is dropping out in the fall-back path in io.Copy it's not trivial to do much better, but it would be better to follow a shutdown path on hup, but it's a little more tricky to plumb as either we need to wire up server read shutdown on client hup notify in order to guarantee a clean overall shutdown path, assuming dropped packets and naïve copy.

bradfitz:
@raggi, what about adding `time.Sleep(2 * time.Second) // wait for pending copies to maybe finish 🤪 `?  

Is that effectively what you mean? We can `shutdown` (er, https://pkg.go.dev/net#TCPConn.CloseWrite) one of the parts and wait for the io.Copy to close on its own?

raggi:
I believe that there is a possible condition to fix here, and having reviewed it in more detail I think there might also be some code paths in netstack that aren't always signalling wake-up when we might need them to. Despite this, I was unable to reproduce _this condition_ using any obvious methods, so I'd like to revisit this with a finer toothcomb before we land changes. For posterity, a patch implementing the wordy thing I was suggesting above is https://github.com/tailscale/tailscale/commit/19defaf6dd292cf5fe021f092d2fe642a1d98877, though similarly I'm not sure this is strictly necessary and want to go through the netstack behavior carefully or produce a functional test to demonstrate the need.

bradfitz:
I forget the context here. Closing.

PR #4536, opened 2022-04-26
ipn/ipnlocal: add Wake-on-LAN function to peerapi

bradfitz:
No CLI support yet. Just the curl'able version if you know the peerapi
port. (like via a TSMP ping)

Updates #306


ipn/ipnlocal/peerapi.go
```diff
@@ -848,6 +857,61 @@ func (h *peerAPIHandler) handleServeDNSFwd(w http.ResponseWriter, r *http.Reques
...
+		return
+	}
+	var password []byte // TODO(bradfitz): support?
+	st, err := interfaces.GetState()
+	if err != nil {
+		http.Error(w, "failed to get interfaces state", http.StatusInternalServerError)
+		return
+	}
+	var res struct {
+		Sent   int
```
maisem, line 886: maybe return which interfaces it was sent on?

bradfitz: Done. PTAL

maisem (approved):

PR #4539, opened 2022-04-26
ipn/ipnlocal/peerapi: add endpoint to list local interfaces

maisem:

### DIFF
--- a/ipn/ipnlocal/peerapi.go
+++ b/ipn/ipnlocal/peerapi.go
@@ -567,6 +567,9 @@ func (h *peerAPIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	case "/v0/wol":
 		h.handleWakeOnLAN(w, r)
 		return
+	case "/v0/interfaces":
+		h.handleServeInterfaces(w, r)
+		return
 	}
 	who := h.peerUser.DisplayName
 	fmt.Fprintf(w, `<html>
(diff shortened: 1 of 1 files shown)


bradfitz (commented):
routing table too while you're here?

maisem:
> routing table too while you're here?

yep, just realized that I want that and the default route

maisem:
> routing table too while you're here?

I'll do that in a different PR as it requires changing a bunch of code to get an OS independent view of the routing table

ipn/ipnlocal/peerapi.go
```diff
@@ -581,6 +584,40 @@ This is my Tailscale device. Your device is %v.
...
+	if err != nil {
+		http.Error(w, err.Error(), 500)
+	}
+	w.Header().Set("Content-Type", "text/html; charset=utf-8")
+	fmt.Fprintln(w, "<h1>Interfaces</h1>")
+	fmt.Fprintf(w, "<h3>Default route is %q(%d)</h3>\n", dr.InterfaceName, dr.InterfaceIndex)
+
+	fmt.Fprintln(w, "<table>")
+	fmt.Fprint(w, "<tr>")
+	for _, v := range []any{"Index", "Name", "MTU", "Flags", "Addrs"} {
```
bradfitz, line 607: are we at the point where we're using `any` because its fewer characters than `string`? :)

bradfitz (approved):

ipn/ipnlocal/peerapi.go
maisem: I had non-string values in the slice below, and this was added later. :)

PR #4548, opened 2022-04-27
control/controlhttp: start port 443 fallback sooner if 80's stuck

### LINKED ISSUE #4544: control/controlhttp: port 443 fallback should start sooner if port 80 is hanging
Looking at the controlhttp code, it seems like if the port 80 dial hangs (firewall, etc) as opposed to failing immediately due to a RST or something, then it can just hang until the context's canceled and never make it to the port 443 fallback.

Some logs (like in #4541) suggest this is happening:

```
2022-04-26 23:28:51.363440274 +0000 UTC: control: TryLogin: register request: Post "https://controlplane.tailscale.com/machine/register": all connection attempts failed (HTTP: context deadline exceeded, HTTPS: context deadline exceeded)
```

We should do something more Happy Eyeballs-ish / racing where we start the 443 dial as a fallback after a couple seconds or something.

/cc @danderson @maisem @crawshaw

bradfitz:
Fixes #4544

### DIFF
--- a/control/controlhttp/client.go
+++ b/control/controlhttp/client.go
@@ -30,6 +30,7 @@ import (
 	"net/http"
 	"net/http/httptrace"
 	"net/url"
+	"time"
 
 	"tailscale.com/control/controlbase"
 	"tailscale.com/net/dnscache"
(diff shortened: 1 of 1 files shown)


control/controlhttp/client.go
```diff
@@ -98,48 +99,105 @@ type dialParams struct {
...
+
+	var err80, err443 error
+	for {
+		select {
+		case <-a.ctx.Done():
+			return nil, fmt.Errorf("connection attempts aborted by context: %w", a.ctx.Err())
+		case res := <-ch:
+			if res.err == nil {
+				if res.u == u443 && err80 == nil || res.u == u80 && !try443Timer.Stop() {
+					// If either port 443 worked while port 80 dial is still
```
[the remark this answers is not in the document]
bradfitz: PTAL

maisem (approved):

maisem (approved):

control/controlhttp/client.go
```diff
@@ -98,48 +99,98 @@ type dialParams struct {
...
+				if try443Timer.Stop() {
+					go try(u443)
+				} // else we lost the race and it started already which is what we want
+			case u443:
+				err443 = res.err
+			default:
+				panic("invalid")
+			}
+			if err80 != nil && err443 != nil {
+				return nil, fmt.Errorf("all connection attempts failed (HTTP: %v, HTTPS: %v)", err80, err443)
```
maisem, line 184: %w?

bradfitz: thought about it but can't justify making just one (which?) %w without some rationale on how I pick the "better" error to wrap

maisem: the one that's not a context error maybe? or the first one?

dsnet (approved):

control/controlhttp/client.go
bradfitz: I'll keep thinking about this but I'm not sure what to do yet, and it's not a regression from what we had before so I'll do it later in a separate change.

PR #4550, opened 2022-04-27
net/tsdial: add SystemDial as a wrapper on netns.Dial

maisem:
The connections returned from SystemDial are automatically closed when
there is a major link change.

Also plumb through the dialer to the noise client so that connections
are auto-reset when moving from cellular to WiFi etc.

Updates #3363

### DIFF
--- a/control/controlhttp/client.go
+++ b/control/controlhttp/client.go
@@ -25,7 +25,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
 	"net"
 	"net/http"
 	"net/http/httptrace"
@@ -34,7 +33,6 @@ import (
 	"tailscale.com/control/controlbase"
 	"tailscale.com/net/dnscache"
 	"tailscale.com/net/dnsfallback"
-	"tailscale.com/net/netns"
 	"tailscale.com/net/netutil"
 	"tailscale.com/net/tlsdial"
 	"tailscale.com/net/tshttpproxy"
@@ -65,7 +63,7 @@ const (
 //
 // The provided ctx is only used for the initial connection, until
 // Dial returns. It does not affect the connection once established.
-func Dial(ctx context.Context, addr string, machineKey key.MachinePrivate, controlKey key.MachinePublic, protocolVersion uint16) (*controlbase.Conn, error) {
+func Dial(ctx context.Context, addr string, machineKey key.MachinePrivate, controlKey key.MachinePublic, protocolVersion uint16, dialer dnscache.DialContextFunc) (*controlbase.Conn, error) {
 	host, port, err := net.SplitHostPort(addr)
 	if err != nil {
 		return nil, err
@@ -79,6 +77,7 @@ func Dial(ctx context.Context, addr string, machineKey key.MachinePrivate, contr
 		controlKey: controlKey,
 		version:    protocolVersion,
 		proxyFunc:  tshttpproxy.ProxyFromEnvironment,
+		dialer:     dialer,
 	}
 	return a.dial()
 }
@@ -92,6 +91,7 @@ type dialParams struct {
 	controlKey key.MachinePublic
 	version    uint16
 	proxyFunc  func(*http.Request) (*url.URL, error) // or nil
+	dialer     dnscache.DialContextFunc
 
 	// For tests only
 	insecureTLS bool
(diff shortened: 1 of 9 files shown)


net/tsdial/tsdial.go
```diff
@@ -91,10 +113,51 @@ func (d *Dialer) SetExitDNSDoH(doh string) {
 	}
 }
 
+func (d *Dialer) Close() {
+	d.mu.Lock()
```
bradfitz, line 117: looks like a bug to not Unlock this too. big comment if that's your intention.

but even though that's clever, it usually leads to pain. e.g. it prevents double Closes which should usually be allowed.

so I'd add an Unlock and add a `closed bool`

control/controlhttp/client.go
```diff
@@ -65,7 +64,7 @@ const (
 //
 // The provided ctx is only used for the initial connection, until
 // Dial returns. It does not affect the connection once established.
-func Dial(ctx context.Context, addr string, machineKey key.MachinePrivate, controlKey key.MachinePublic, protocolVersion uint16) (*controlbase.Conn, error) {
+func Dial(ctx context.Context, addr string, machineKey key.MachinePrivate, controlKey key.MachinePublic, protocolVersion uint16, dialer *tsdial.Dialer) (*controlbase.Conn, error) {
```
bradfitz, line 67: does this actually need the whole `*tsdial.Dialer` type or just the dial func?

net/tsdial/tsdial.go
```diff
@@ -91,10 +113,51 @@ func (d *Dialer) SetExitDNSDoH(doh string) {
...
+
+func (d *Dialer) closeSysConn(id int) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	c, ok := d.activeSysConns[id]
+	if !ok {
+		return nil
+	}
+	delete(d.activeSysConns, id)
+	go c.Close()
```
bradfitz, line 159: add comment that we're throwing away this Close error value and we're fine with that.

net/tsdial/tsdial.go
```diff
@@ -91,10 +113,51 @@ func (d *Dialer) SetExitDNSDoH(doh string) {
...
+	}
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	for id, c := range d.activeSysConns {
+		go c.Close()
+		delete(d.activeSysConns, id)
+	}
+}
+
+func (d *Dialer) closeSysConn(id int) error {
```
bradfitz, line 151: should this even return an error? it's kinda misleading. at least document it always returns nil?

net/tsdial/tsdial.go
```diff
@@ -91,10 +113,51 @@ func (d *Dialer) SetExitDNSDoH(doh string) {
 	}
 }
 
+func (d *Dialer) Close() {
```
bradfitz, line 116: Close without a return isn't an `io.Closer` which is sometimes annoying

bradfitz (commented):
LGTM overall

control/controlhttp/client.go
maisem: Dial func is sufficient, done

net/tsdial/tsdial.go
maisem: Removed the error return

net/tsdial/tsdial.go
maisem: This is where I actually wanted to add the error return :)

net/tsdial/tsdial.go
maisem: Done

net/tsdial/tsdial.go
```diff
@@ -197,6 +264,42 @@ func ipNetOfNetwork(n string) string {
 	return "ip"
 }
 
+// SystemDial connects to the provided network address without going over
+// Tailscale. It prefers going over the default interface and closes existing
+// connections if the default interface changes. It is used to connect to
+// Control and DERPs.
```
bradfitz, line 270: and (in the future, as of 2022-04-27) DERPs.

net/tsdial/tsdial.go
```diff
@@ -46,12 +51,31 @@ type Dialer struct {
...
+	tunName           string // tun device name
+	linkMon           *monitor.Mon
+	linkMonUnregister func()
+	exitDNSDoHBase    string                 // non-empty if DoH-proxying exit node in use; base URL+path (without '?')
+	dnsCache          *dnscache.MessageCache // nil until first first non-empty SetExitDNSDoH
+	nextSysConnID     int
+	activeSysConns    map[int]net.Conn // active connections not yet closed
+}
+
+// sysConn represents a connection that was created using d.SystemDial.
```
bradfitz, line 69: s/represents/wraps/
+
// It exists to track which net.Conns are still open.

bradfitz (approved):

PR #4556, opened 2022-04-28
Fix shellcheck errors in k8s script

tolkonepiu:
Fixed all the warnings generated by spellcheck ([SC3010](https://github.com/koalaman/shellcheck/wiki/SC3010), [SC3014](https://github.com/koalaman/shellcheck/wiki/SC3014), [SC2236](https://github.com/koalaman/shellcheck/wiki/SC2236), [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086)).

Signed-off-by: Pavel Popov <popoffpavel@yandex.ru>

### DIFF
--- a/docs/k8s/run.sh
+++ b/docs/k8s/run.sh
@@ -4,7 +4,7 @@
 
 #! /bin/sh
 
-export PATH=$PATH:/tailscale/bin
+export PATH="$PATH:/tailscale/bin"
 
 AUTH_KEY="${AUTH_KEY:-}"
 ROUTES="${ROUTES:-}"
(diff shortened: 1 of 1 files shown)


DentonGentry (commented):
The intent is to de-bashify the script?
Did that cause a particular failure?

tolkonepiu:
@DentonGentry 

> The intent is to de-bashify the script?

It focuses on refactoring this script to be POSIX compliant.

> Did that cause a particular failure?

I haven't tried running the original script. I think that there are no errors in it, apart from problems in [Dockerfile](https://github.com/tailscale/tailscale/pull/4125).

DentonGentry:
You didn't run the script before sending a PR to change it.

raggi:
As a first change, I would suggest sticking to:

- Add quotes around PATH modification
- Move from `[[` to `[`
- Prefer `-n` over `! -z`
- Use `=` rather than `==` in tests.

For now I would like to avoid the introduction of `eval` that seems like it's attempting to solve for word splitting challenges, but it potentially introduces another layer of splitting - shellcheck is perhaps giving bad advice here? It might be better to simply explicitly field split `*_ARGS`. If we really want to solve for rational word splitting in argument expansion then avoiding aggregating into a single ARGS flag is likely a better strategy but a bigger change.

Given that this script is targeted at containers though, it seems like explicitly using bash or zsh or some explicit shell and making use of arrays could simplify things quite a bit.

Issue #4557, opened 2022-04-28
control/controlhttp: port 80 noise upgrade problem


bradfitz:
Hey @danderson, @maisem, more controlv2 connection woes on port 80.

This is 1.24.1 on Windows:

```
(logID "d2796595a7110830caab2cb5cfd469abb69ad546b619c678736b515bf5d52b2f")
2022-04-28T09:37:30.359+08:00: Program starting: v1.24.1-t1a9302b1e-g1331ed583, Go 1.18.1-ts710a0d8610: []string{"C:\\Program Files\\Tailscale\\tailscaled.exe"}
2022-04-28T09:37:30.360+08:00: LogID: d2796595a7110830caab2cb5cfd469abb69ad546b619c678736b515bf5d52b2f
2022-04-28T09:37:30.360+08:00: logpolicy: using dir C:\ProgramData\Tailscale
2022-04-28T09:37:30.367+08:00: Running service...
2022-04-28T09:37:30.369+08:00: registry.OpenKey(SOFTWARE\Policies\Tailscale): The system cannot find the file specified.
2022-04-28T09:37:30.378+08:00: exec: "C:\\Program Files\\Tailscale\\tailscaled.exe" [/subproc d2796595a7110830caab2cb5cfd469abb69ad546b619c678736b515bf5d52b2f]
2022-04-28T09:37:30.430+08:00: Program starting: v1.24.1-t1a9302b1e-g1331ed583: []string{"C:\\Program Files\\Tailscale\\tailscaled.exe", "/subproc", "d2796595a7110830caab2cb5cfd469abb69ad546b619c678736b515bf5d52b2f"}
2022-04-28T09:37:30.430+08:00: subproc mode: logid=d2796595a7110830caab2cb5cfd469abb69ad546b619c678736b515bf5d52b2f
....
2022-04-28T09:37:49.021+08:00: control: [v1] TryLogin: register request: Post "https://controlplane.tailscale.com/machine/register": reading response header: read tcp 192.168.87.99:50582->52.29.8.43:80: i/o timeout
2022-04-28T09:37:49.022+08:00: control: [v1] sendStatus: authRoutine-report: state:authenticating
2022-04-28T09:37:49.022+08:00: Received error: register request: Post "https://controlplane.tailscale.com/machine/register": reading response header: read tcp 192.168.87.99:50582->52.29.8.43:80: i/o timeout
[...]

PR #4559, opened 2022-04-28
control/controlhttp: don't assume port 80 upgrade response will work

bradfitz:
Just because we get an HTTP upgrade response over port 80, don't
assume we'll be able to do bi-di Noise over it. There might be a MITM
corp proxy or anti-virus/firewall interfering. Do a bit more work to
validate the connection before proceeding to give up on the TLS port
443 dial.

Updates #4557 (probably fixes)

### DIFF
--- a/control/controlhttp/http_test.go
+++ b/control/controlhttp/http_test.go
@@ -17,6 +17,7 @@ import (
 	"strconv"
 	"sync"
 	"testing"
+	"time"
 
 	"tailscale.com/control/controlbase"
 	"tailscale.com/net/socks5"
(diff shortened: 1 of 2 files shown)


maisem (approved):
Either add a test or a TODO to add a test to make sure we don't regress here when we have a unfriendly MITM proxy

bradfitz:
> Either add a test or a TODO to add a test to make sure we don't regress here when we have a unfriendly MITM proxy

I added a test being an unfriendly proxy and verified the test failed before and passes now.

control/controlhttp/http_test.go
```diff
@@ -232,7 +256,14 @@ func (s *socksProxy) Start(t *testing.T) (url string) {
 	}
 	s.ln = ln
 	s.clientConnAddrs = map[string]bool{}
-	s.proxy.Logf = t.Logf
+	s.proxy.Logf = func(format string, a ...any) {
+		s.Lock()
+		defer s.Unlock()
+		if s.closed {
+			return
+		}
```
maisem, line 264: wow lol

maisem (approved):

control/controlhttp/http_test.go
bradfitz: to fix a windows-race test flake in this code

PR #4562, opened 2022-04-28
net/tshttpproxy: use http as the scheme for proxies

maisem:
Currently we try to use `https://` when we see `https_host`, however
that doesn't work and results in errors like `Received error: fetch
control key: Get "https://controlplane.tailscale.com/key?v=32":
proxyconnect tcp: tls: first record does not look like a TLS handshake`

This indiciates that we are trying to do a HTTPS request to a HTTP
server. Googling suggests that the standard is to use `http` regardless
of `https` or `http` proxy

Updates #4395, #2605

### DIFF
--- a/net/tshttpproxy/tshttpproxy_synology.go
+++ b/net/tshttpproxy/tshttpproxy_synology.go
@@ -96,15 +96,15 @@ func parseSynologyConfig(r io.Reader) (*url.URL, error) {
 		return nil, nil
 	}
 
-	proxyURL := new(url.URL)
+	proxyURL := &url.URL{
+		Scheme: "http", // regardless of proxy type
+	}
 	if cfg["auth_enabled"] == "yes" {
 		proxyURL.User = url.UserPassword(cfg["proxy_user"], cfg["proxy_pwd"])
 	}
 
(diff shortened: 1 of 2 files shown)


maisem:
More info in thread https://tailscale.slack.com/archives/CTU3D5HAS/p1651164332735139

raggi:
The test needs an update too, otherwise lgtm

net/tshttpproxy/tshttpproxy_synology.go
```diff
@@ -101,10 +101,10 @@ func parseSynologyConfig(r io.Reader) (*url.URL, error) {
 		proxyURL.User = url.UserPassword(cfg["proxy_user"], cfg["proxy_pwd"])
 	}
 
-	proxyURL.Scheme = "https"
+	// Scheme is always `http` regardless of the type of proxy.
+	proxyURL.Scheme = "http"
```
bradfitz, line 105: put it up on line 99 in a composite literal?

```go
proxyURL := &url.URL{
    Scheme: "http", // regardless of proxy type
}
```

bradfitz (approved):

PR #4568, opened 2022-04-28
safesocket: log warning when running sandboxed Mac binary as root

### LINKED ISSUE #3063: safesocket: warn against sudo on macOS
If a user tries to use the macOS GUI binary as the CLI, it's sandboxed, and we have a whole complicated mechanism to make it work.

But if you use e.g. `sudo /Applications/Tailscale.app/Contents/MacOS/Tailscale cert xxxxx.ts.net`

Then you get:

```
Get "http://local-tailscaled.sock/localapi/v0/cert/xxx.ts.net?type=pair": safesocket: failed to connect to /var/run/tailscaled.socket: dial unix /var/run/tailscaled.socket: connect: no such file or directory; failed to connect to Tailscale IPNExtension: no sameuserproof token found in TS_MACOS_CLI_SHARED_DIR "/private/var/root/Library/Group Containers/W5364U7YZB.group.io.tailscale.ipn.macos”
```

... because sudo makes the "sameuserproof token" not available.

We should modify safesocket to detect if sudo was used on macOS and if we were otherwise going to return an error like that, instead tell them to try again without sudo.

/cc @DentonGentry

mihaip:
It won't work, provide a clue in the error output.

Fixes #3063

### DIFF
--- a/safesocket/safesocket_darwin.go
+++ b/safesocket/safesocket_darwin.go
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 )
 
 func init() {
@@ -49,6 +50,8 @@ func localTCPPortAndTokenMacsys() (port int, token string, err error) {
 	return port, auth, nil
 }
 
+var warnAboutRootOnce sync.Once
+
 func localTCPPortAndTokenDarwin() (port int, token string, err error) {
 	// There are two ways this binary can be run: as the Mac App Store sandboxed binary,
 	// or a normal binary that somebody built or download and are being run from outside
@@ -82,6 +85,14 @@ func localTCPPortAndTokenDarwin() (port int, token string, err error) {
 					}
 				}
 			}
+		}
+		if os.Geteuid() == 0 {
+			// Log a warning as the clue to the user, in case the error
+			// message is swallowed. Only do this once since we may retry
+			// multiple times to connect, and don't want to spam.
+			warnAboutRootOnce.Do(func() {
+				fmt.Fprintf(os.Stderr, "Warning: The CLI is running as root from within a sandboxed binary. It cannot reach the local tailscaled, please try again as a regular user.\n")
+			})
 		}
 		return 0, "", fmt.Errorf("failed to find sandboxed sameuserproof-* file in TS_MACOS_CLI_SHARED_DIR %q", dir)
 	}


safesocket/safesocket_darwin.go
```diff
@@ -49,6 +50,8 @@ func localTCPPortAndTokenMacsys() (port int, token string, err error) {
 	return port, auth, nil
 }
 
+var warnAboutRootOnce sync.Once
```
maisem, line 53: Add a comment why here.

safesocket/safesocket_darwin.go
```diff
@@ -83,6 +86,11 @@ func localTCPPortAndTokenDarwin() (port int, token string, err error) {
 				}
 			}
 		}
+		if os.Getuid() == 0 {
+			warnAboutRootOnce.Do(func() {
```
mihaip, line 90: Done as a separate warning log because the error returned by this function is swallowed (we continue on the fallback path at https://github.com/tailscale/tailscale/blob/a1dbd56c18ccd480c72e4d7876065a52bde19acc/client/tailscale/tailscale.go#L67-L71).

safesocket/safesocket_darwin.go
```diff
@@ -83,6 +86,11 @@ func localTCPPortAndTokenDarwin() (port int, token string, err error) {
 				}
 			}
 		}
+		if os.Getuid() == 0 {
```
maisem, line 89: I think you should use `os.Geteuid` here

safesocket/safesocket_darwin.go
maisem, line 90: Add a comment here too about why this exists, the rest of the file is very verbose on comments might as well continue that here.

maisem (commented):
Change LGTM

safesocket/safesocket_darwin.go
bradfitz: When you run `sudo`, is the `TS_MACOS_CLI_SHARED_DIR` environment set in the new environment? I always forget how that works. But I assume if you tested this, then it works as-is.

bradfitz (approved):
lowercase verb after colon in the commit message, but LGTM otherwise, thanks!

safesocket/safesocket_darwin.go
mihaip: Environment variables are not inherited when using `sudo` (unless you use `-E`), but in this case we first start running as root, then we set the environment variable, then we run the CLI (in the same process, where we read it back out), so it's not a concern.

safesocket/safesocket_darwin.go
mihaip: TIL

safesocket/safesocket_darwin.go
mihaip: Added some comments where it's actually used, but this one seemed self-explanatory.

maisem (approved):

PR #4574, opened 2022-04-29
Dockerfile: narrow host's mount directory scope

pataquets:
Recommend a narrowly defined directory access scope to confine tailscale container, for security reasons.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@
 #
 # To run the tailscaled agent:
 #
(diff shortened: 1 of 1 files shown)


pataquets:
Docker's default behaviour for `-v` switch is to create directories on host (with `root:root` ownership) if they do not exist.
Since I was using `docker-compose`, I fired a shell and ran the command to make sure it works :+1:.

PR #4575, opened 2022-04-29
client/tailscale: move/copy all package funcs to new LocalClient type

bradfitz:
Remove all global variables, and clean up tsnet and cmd/tailscale's usage.

This is in prep for using this package for the web API too (it has the
best package name).

### DIFF
--- a/client/tailscale/tailscale.go
+++ b/client/tailscale/tailscale.go
@@ -38,24 +38,59 @@ import (
 	"tailscale.com/tailcfg"
 )
 
-var (
-	// TailscaledSocket is the tailscaled Unix socket. It's used by the TailscaledDialer.
-	TailscaledSocket = paths.DefaultTailscaledSocket()
+// defaultLocalClient is the default LocalClient when using the legacy
+// package-level functions.
+var defaultLocalClient LocalClient
 
-	// TailscaledSocketSetExplicitly reports whether the user explicitly set TailscaledSocket.
-	TailscaledSocketSetExplicitly bool
+// LocalClient is a client to Tailscale's "local API", communicating with the
+// Tailscale daemon on the local machine. Its API is not necessarily stable and
+// subject to changes between releases. Some API calls have stricter
+// compatibility guarantees, once they've been widely adopted. See method docs
+// for details.
+//
+// Its zero value is valid to use.
+//
+// Any exported fields should be set before using methods on the type
+// and not changed thereafter.
+type LocalClient struct {
+	// Dial optionally specifies an alternate func that connects to the local
+	// machine's tailscaled or equivalent. If nil, a default is used.
+	Dial func(ctx context.Context, network, addr string) (net.Conn, error)
 
-	// TailscaledDialer is the DialContext func that connects to the local machine's
-	// tailscaled or equivalent.
-	TailscaledDialer = defaultDialer
-)
+	// Socket specifies an alternate path to the local Tailscale socket.
+	// If empty, a platform-specific default is used.
+	Socket string
 
-func defaultDialer(ctx context.Context, network, addr string) (net.Conn, error) {
+	// UseSocketOnly, if true, tries to only connect to tailscaled via the
+	// Unix socket and not via fallback mechanisms as done on macOS when
+	// connecting to the GUI client variants.
+	UseSocketOnly bool
+
+	// tsClient does HTTP requests to the local Tailscale daemon.
+	// It's lazily initialized on first use.
+	tsClient     *http.Client
+	tsClientOnce sync.Once
+}
+
+func (lc *LocalClient) socket() string {
+	if lc.Socket != "" {
+		return lc.Socket
+	}
+	return paths.DefaultTailscaledSocket()
+}
+
+func (lc *LocalClient) dialer() func(ctx context.Context, network, addr string) (net.Conn, error) {
+	if lc.Dial != nil {
+		return lc.Dial
+	}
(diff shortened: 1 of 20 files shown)


tsnet/tsnet.go
```diff
@@ -263,7 +275,7 @@ func (s *Server) start() error {
 	s.localAPIListener = lal
 
 	// Override the Tailscale client to use the in-process listener.
```
mihaip, line 277: Should this comment be removed since we're no longer overriding global state?

client/tailscale/tailscale.go
```diff
@@ -38,24 +38,57 @@ import (
...
+	// UseSocketOnly, if true, tries to only connect to tailscaled via the
+	// Unix socket and not via fallback mechanisms as done on macOS when
+	// connecting to the GUI client variants.
+	UseSocketOnly bool
 
-	// TailscaledSocketSetExplicitly reports whether the user explicitly set TailscaledSocket.
-	TailscaledSocketSetExplicitly bool
+	// tsClient does HTTP requests to the local Tailscale daemon.
+	// We lazily initialize the client in case the caller wants to
+	// override TailscaledDialer.
```
mihaip, line 68: `TailscaledDialer` is no longer a thing, right?

client/tailscale/tailscale.go
```diff
@@ -571,7 +639,7 @@ func GetCertificate(hi *tls.ClientHelloInfo) (*tls.Certificate, error) {
 }
 
 // ExpandSNIName expands bare label name into the the most likely actual TLS cert name.
-func ExpandSNIName(ctx context.Context, name string) (fqdn string, ok bool) {
+func (lc *LocalClient) ExpandSNIName(ctx context.Context, name string) (fqdn string, ok bool) {
```
maisem, line 642: leave this on the package level as well

client/tailscale/tailscale.go
```diff
@@ -38,24 +38,57 @@ import (
 	"tailscale.com/tailcfg"
 )
 
-var (
-	// TailscaledSocket is the tailscaled Unix socket. It's used by the TailscaledDialer.
-	TailscaledSocket = paths.DefaultTailscaledSocket()
+// Local is the default LocalClient when using the legacy package-level functions.
+var Local LocalClient
```
maisem, line 42: should this be a pointer?

client/tailscale/tailscale.go
bradfitz: done

client/tailscale/tailscale.go
maisem: I'm worried about the sync.Once being copied

maisem: https://pkg.go.dev/sync#Once says it should not be copied

client/tailscale/tailscale.go
bradfitz: done

tsnet/tsnet.go
maisem: There should be a RELNOTE for this as it's a breaking change

client/tailscale/tailscale.go
```diff
@@ -38,24 +38,59 @@ import (
...
```
mihaip, line 42: Maybe if this was not public it would be less of an issue? Seems like we want people to move away from it being a singleton, no need to expose it.

bradfitz: removed.

client/tailscale/tailscale.go
bradfitz: PTAL

tsnet/tsnet.go
bradfitz: Done.

mihaip (approved):
I assume it's intentional that we added deprecated stubs for only ~4 global functions and others (e.g. `CheckIPForwarding`) are gone.

cmd/proxy-to-grafana/proxy-to-grafana.go
```diff
@@ -62,6 +62,12 @@ func main() {
 		Hostname: *hostname,
 	}
 
+	// TODO(bradfitz,maisem): move this to a method on tsnet.Server probably.
+	if err := ts.Start(); err != nil {
```
maisem, line 66: you can drop the call to `ts.Start` and just call `ts.LocalClient`
```
localClient, err := ts.LocalClient(); err != nil {
   ...
}
```

bradfitz: I could but I like the explicit Start.

bradfitz:
> I assume it's intentional that we added deprecated stubs for only ~4 global functions and others (e.g. CheckIPForwarding) are gone.

Yeah, those are the ones that anybody in the community besides us use.

tsnet/tsnet.go
```diff
@@ -90,6 +91,17 @@ func (s *Server) Dial(ctx context.Context, network, address string) (net.Conn, e
 	return s.dialer.UserDial(ctx, network, address)
 }
 
+// LocalClient returns a LocalClient that speaks to s.
+//
+// It will start the server if it has not been started yet. If the server's
+// already been started, it doesn't return an error.
```
maisem, line 97: `started successfully,`

bradfitz: then arguably it didn't start :)

bradfitz: done

maisem (approved):

PR #4576, opened 2022-04-29
all: gofmt all

bradfitz:
Well, goimports actually (which adds the normal import grouping order we do)


mihaip (approved):
Time for a CI check?

bradfitz:
> Time for a CI check?

Probably. It's a little annoying though because gofmt rules are allowed to change between releases and have a few times in the past and then you get annoying errors while people e.g. upgrade their editors which might not be using the exact same version of Go or goimports that our build is using.

PR #4580, opened 2022-04-29
ipn: remove FakeExpireAfter Backend function

mihaip:
No callers remain (last one was removed with
tailscale/corp@1c095ae08fcc21362db65035a117672a71b4a081), and it's
pretty esoteric.


bradfitz (approved):
🎉

PR #4581, opened 2022-04-29
ssh/tailssh,ipn/ipnlocal: terminate any active sessions on `up --ssh=false`

maisem:
Currently the ssh session isn't terminated cleanly, instead the packets
are just are no longer routed to the in-proc SSH server. This makes it
so that clients get a disconnection when the `RunSSH` pref changes to
`false`.

Updates #3802


bradfitz (approved):

bradfitz:
It'd be nice if we had a small window to send them ssh packets but not accept them back or something so we could RST them nicely and maybe say "closed!" But this is good enough for now.

maisem:
I get a "access revoked" message

PR #4589, opened 2022-04-30
cmd/viewer: add codegen utility to generate Views

maisem:


bradfitz:
Can we put views in their own `foo_view.go` file? And then move the existing hand-written views to such view files first, so this diff is then readable to see how the machine-generated code varies? (or not)

maisem:
Yeah sgtm, I'll break this up further.

tailcfg/tailcfg_view.go
```diff
@@ -0,0 +1,92 @@
+package tailcfg
```
bradfitz, line 8: copyright header

cmd/cloner/cloner.go
```diff
@@ -21,14 +21,15 @@ import (
...
+	"time"
 
 	"golang.org/x/tools/go/packages"
 	"tailscale.com/util/codegen"
 )
 
 var (
 	flagTypes     = flag.String("type", "", "comma-separated list of types; required")
-	flagOutput    = flag.String("output", "", "output file; required")
+	_             = flag.String("output", "", "deprecated: not used")
```
bradfitz, line 32: just delete?

cmd/cloner/cloner.go
```diff
@@ -58,19 +59,22 @@ func main() {
 		log.Fatalf("wrong number of packages: %d", len(pkgs))
 	}
 	pkg := pkgs[0]
-	buf := new(bytes.Buffer)
-	imports := make(map[string]struct{})
+	cloneBuf := new(bytes.Buffer)
+	viewBuf := new(bytes.Buffer)
+	cloneImports := make(map[string]struct{})
+	viewImports := make(map[string]struct{})
```
bradfitz, line 65: high level question: why merge this functionality into the cmd/cloner tool? It seems unrelated, or unrelated enough to be its own specific tool.

If they need to share some code, make a package?

ipn/ipn_clone.go
```diff
@@ -1,9 +1,9 @@
-// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
 // Code generated by tailscale.com/cmd/cloner; DO NOT EDIT.
-//go:generate go run tailscale.com/cmd/cloner -type=Prefs -output=prefs_clone.go
+//go:generate go run tailscale.com/cmd/cloner -type=Prefs
```
bradfitz, line 6: did I already lose this battle once? I don't like go:generate in generated files. I like being able to nuke them and "go generate" them back into existence, and have the go:generate line at the source of truth (the input data to the generation)

tailcfg/tailcfg_view.go
```diff
@@ -1,13 +1,120 @@
...
+	"tailscale.com/types/dnstype"
+	"tailscale.com/types/key"
 	"tailscale.com/types/opt"
 	"tailscale.com/types/views"
 )
 
+// View returns a readonly view of User.
+func (p *User) View() UserView { return UserView{p} }
+
+type UserView struct {
```
bradfitz, line 25: generate docs for the types at least? and mention that methods are only safe to call if Valid?

wgengine/wgcfg/wgcfg_view.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"inet.af/netaddr"
+	"tailscale.com/types/key"
+	"tailscale.com/types/views"
+)
+
+// View returns a readonly view of Config.
+func (p *Config) View() ConfigView { return ConfigView{p} }
+
+type ConfigView struct {
```
bradfitz, line 22: do we need this? I'd prefer to opt-in to which views it generates, rather than it just generating code for everything we need we want a Clone for.

bradfitz (commented):
Also: tests?

wgengine/wgcfg/wgcfg_view.go
maisem: I disagree, I think having the ability to defer a `Clone` by providing a `View` is really useful. `View` also relies on `Clone` to provide the `AsStruct` functionality.

As for this exact usage, turns out we can indeed replace our current usage of `Clone` with `View`.
https://github.com/tailscale/tailscale/blob/7f45734663aecae79c8dd86214bbf872ecc3baf8/wgengine/userspace.go#L873

maisem: https://github.com/tailscale/tailscale/blob/7f45734663aecae79c8dd86214bbf872ecc3baf8/wgengine/userspace.go#L1457-L1467

bradfitz: I'm fine with views implying clones, but I'd like use of cloner to not require our religion about views. (Or if this is purely an internal tool, let's say so?)

In any case, I prefer N small tools to 1 megatool that does everything. And I prefer explicitness (generate this view!) over magic.

maisem: I assumed it was internal because it was in this repo and not a separate one. I'll modify it to add a cmd/viewer

bradfitz: Sure. But then we need to update the docs. The package doc only talks about Clone.

And I don't want to see any types where Cloner would've worked but Viewer fails and that prevents Cloner from working.

bradfitz: cmd/viewer SGTM. I missed your last sentence.

util/codegen/codegen.go
```diff
@@ -11,13 +11,126 @@ import (
...
+		}
+		val = strings.Trim(val, "\"")
+		for _, v := range strings.Split(val, ",") {
+			if v == "noclone" {
+				return true
+			}
+		}
+	}
+	return false
+}
```
bradfitz, line 64: https://pkg.go.dev/reflect#StructTag

cmd/viewer/tests/tests.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"fmt"
+
+	"inet.af/netaddr"
+)
+
+//go:generate go run tailscale.com/cmd/viewer --type=Value,Struct,Map,Slices
+
+type Value struct {
```
bradfitz, line 16: StructWithoutPtrs?

cmd/viewer/tests/tests.go
```diff
@@ -0,0 +1,46 @@
...
+	Int int
+	Pfx netaddr.IPPrefix
+}
+
+type Map struct {
+	M map[string]int
+}
+
+type Struct struct {
```
bradfitz, line 25: StructWithPtrs?

cmd/viewer/tests/tests_view.go
```diff
@@ -0,0 +1,249 @@
...
+import (
+	"encoding/json"
+	"errors"
+
+	"go4.org/mem"
+	"inet.af/netaddr"
+	"tailscale.com/types/views"
+)
+
+//go:generate go run tailscale.com/cmd/cloner  -clonefunc=false -type=Value,Struct,Map,Slices
```
bradfitz, line 18: blank line after this so it doesn't become the godoc

cmd/viewer/tests/tests_view.go
```diff
@@ -0,0 +1,249 @@
...
+// View returns a readonly view of Value.
+func (p *Value) View() ValueView {
+	return ValueView{ж: p}
+}
+
+// ValueView provides a read-only view on Value.
```
bradfitz, line 24: "view of" ? "view over" ?

cmd/viewer/tests/tests_view.go
```diff
@@ -0,0 +1,249 @@
...
+// It's methods should only be called if `Valid()` returns true.
```
bradfitz, line 25: Its :)

cmd/viewer/viewer.go
```diff
@@ -0,0 +1,288 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Viewer is a tool to automate the creation of read-only wrapper structs.
```
bradfitz, line 5: of "view" wrapper types that provide read-only accessor methods to the underlying type

or something

tailcfg/tailcfg_view.go
```diff
@@ -2,99 +2,323 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+// Code generated by tailscale.com/util/codegen (viewer); DO NOT EDIT.
```
bradfitz, line 5: say "by /cmd/viewer" or wahtever

tailcfg/tailcfg_view.go
```diff
@@ -2,99 +2,323 @@
...
+// View returns a readonly view of User.
+func (p *User) View() UserView {
+	return UserView{ж: p}
+}
 
-// HostinfoView is a read-only accessor for Hostinfo.
-// See Hostinfo.
-type HostinfoView struct {
+// UserView provides a read-only view on User.
+// It's methods should only be called if `Valid()` returns true.
```
bradfitz, line 29: add a blank "//" line separating these two sentences

tailcfg/tailcfg_view.go
```diff
@@ -2,99 +2,323 @@
...
+	"tailscale.com/types/dnstype"
+	"tailscale.com/types/key"
 	"tailscale.com/types/opt"
+	"tailscale.com/types/structs"
 	"tailscale.com/types/views"
 )
 
-// View returns a read-only accessor for hi.
-func (hi *Hostinfo) View() HostinfoView { return HostinfoView{hi} }
+//go:generate go run tailscale.com/cmd/cloner  -clonefunc=true -type=User,Node,Hostinfo,NetInfo,Login,DNSConfig,RegisterResponse,DERPRegion,DERPMap,DERPNode
```
bradfitz, line 22: reminder to make the "go generate" step on re-parse package and if type checking or whatever fails then say loudly to stderr, "hey, you gotta run this a second time"

types/views/views.go
```diff
@@ -14,6 +14,78 @@ import (
...
+	if *x != nil {
+		return errors.New("already initialized")
+	}
+	if len(b) == 0 {
+		return nil
+	}
+	return json.Unmarshal(b, x)
+}
+
+// View represents the corresponding View of a Viewable. The concrete types are
```
bradfitz, line 27: clarify this is always a struct and that slices use ViewSlice?

StructView? ViewStruct?

bradfitz: actually it's a pointer to struct!

types/views/views.go
```diff
@@ -14,6 +14,78 @@ import (
...
+	// Valid reports whether the underlying Viewable is nil.
+	Valid() bool
+	// AsStruct returns a deep-copy of the underlying value.
+	// It returns nil, if Valid() is false.
+	AsStruct() T
+}
+
+// Viewable is any type that has had a corresponding View generated by the use
+// of tailscale.com/cmd/viewer.
+type Viewable[T any, V View[T]] interface {
```
bradfitz, line 39: type ViewCloner ?

like io.ReadCloser, io.ReadWriter, etc

types/views/views.go
```diff
@@ -14,6 +14,78 @@ import (
...
+// typically generated by tailscale.com/cmd/viewer.
+type View[T any] interface {
+	// Valid reports whether the underlying Viewable is nil.
+	Valid() bool
+	// AsStruct returns a deep-copy of the underlying value.
+	// It returns nil, if Valid() is false.
+	AsStruct() T
+}
+
+// Viewable is any type that has had a corresponding View generated by the use
```
bradfitz, line 37: and Clone

types/views/views.go
```diff
@@ -14,6 +14,78 @@ import (
...
+
+// SliceOfViews returns a ViewSlice for x.
+func SliceOfViews[T Viewable[T, V], V View[T]](x []T) ViewSlice[T, V] {
+	return ViewSlice[T, V]{x}
+}
+
+// ViewSlice is a read-only wrapper around a struct which should only be exposed
+// as a View.
+type ViewSlice[T Viewable[T, V], V View[T]] struct {
+	// It is named distinctively to make you think of how dangerous it is to escape
```
bradfitz, line 56: "It" is not Go comment style :)

bradfitz: // ж is the underlying mutable value, named with a hard-to-type
// character that looks pointy like a pointer.
// It's named distinctively ....

types/views/views.go
```diff
@@ -14,6 +14,78 @@ import (
...
+}
+
+// SliceOfViews returns a ViewSlice for x.
+func SliceOfViews[T Viewable[T, V], V View[T]](x []T) ViewSlice[T, V] {
+	return ViewSlice[T, V]{x}
+}
+
+// ViewSlice is a read-only wrapper around a struct which should only be exposed
+// as a View.
+type ViewSlice[T Viewable[T, V], V View[T]] struct {
```
bradfitz, line 55: SliceView

bradfitz (approved):

dsnet:
I just saw this! Really cool! I was playing around with something on and off that looked very similar to this. I had no idea that you were working on this.

I'm intrigued that you ended up with both `views.Slice` and `views.SliceView` to handle both cases of slices of mutable and immutable types. I came to the same conclusion in my prototype.

I noticed that maps are a TODO. That gets more complicated since you end up with 4 possibilities:
1. map with immutable key and immutable value (e.g., map[string]string)
2. map with immutable key and mutable value (e.g., map[string]*tar.Header)
3. map with mutable key and immutable value (e.g.,  map[*tar.Header]string)
4. map with mutable key and mutable value (e.g., map[*tar.Header]*tar.Header)

Of course,  cases 3 and 4 are much more rare.

PR #4598, opened 2022-05-02
net/dns, wgengine: implement DNS over TCP

### LINKED ISSUE #1123: Support DNS over TCP
**Describe the bug**
Magic DNS doesn't support TCP, preventing queries with long records from working.

**To Reproduce**
```bash
# Works, outputs long record: (UDP request, TCP response)
dig @8.8.8.8 -t TXT longrecord.smitop.com

# No record and "Message parser reports malformed message packet": (UDP request, TCP response)
dig @100.100.100.100 -t TXT longrecord.smitop.com

# Doesn't even go through the Magic DNS system, "open-conn-track: timeout opening (100.66.99.116:48353 => 100.100.100.100:53); no associated peer node" (TCP request, TCP response)
dig @100.100.100.100 +tcp smitop.com
```

**Expected behavior**
Magic DNS dosen't support DNS over TCP, currently it only supports UDP. As such, long records that don't fit in a UDP packet fail (even for IPs on the public Internet). It at least shouldn't return malformed packets. This also causes issues with DNSSEC, where answer sizes are larger due to being signed (not a security issue though since I believe failure here would just cause the answer to be rejected))

**Version information:**
 - OS: Pop!_OS 20.10
 - Tailscale version: Tip of `main`

twitchyliquid64:
Fixes #1123

### DIFF
--- a/net/dns/manager.go
+++ b/net/dns/manager.go
@@ -7,7 +7,9 @@ package dns
 import (
 	"bufio"
 	"context"
+	"encoding/binary"
 	"errors"
+	"io"
 	"net"
 	"runtime"
 	"sync/atomic"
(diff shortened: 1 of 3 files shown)


net/dns/manager_tcp_test.go
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package dns
```
bradfitz, line 4: blank line separating the comment and package line, otherwise it becomes a package doc comment

net/dns/manager.go
```diff
@@ -346,6 +348,125 @@ func (m *Manager) Query(ctx context.Context, bs []byte, from netaddr.IPPort) ([]
...
+				return
+			case s.queries <- buff:
+			}
+		}
+	}
+}
+
+// HandleTCPConn implements magicDNS over TCP, taking a connection and
+// servicing DNS requests sent down it.
+func (m *Manager) HandleTCPConn(conn net.Conn, from netaddr.IPPort) {
```
bradfitz, line 457: what is "from"? is it not conn.RemoteAddr?

net/dns/manager.go
```diff
@@ -346,6 +348,125 @@ func (m *Manager) Query(ctx context.Context, bs []byte, from netaddr.IPPort) ([]
...
+
+// dnsTCPSession services DNS requests sent over TCP.
+type dnsTCPSession struct {
+	m *Manager
+
+	conn net.Conn
+	from netaddr.IPPort
+
+	responses chan []byte // DNS replies pending writing
+	queries   chan []byte // DNS requests from reading
```
bradfitz, line 371: what's the point of this channel from the read side to the write side?

can't the reader start the handleQuery goroutines?

net/dns/manager.go
```diff
@@ -346,6 +348,125 @@ func (m *Manager) Query(ctx context.Context, bs []byte, from netaddr.IPPort) ([]
...
+	s := dnsTCPSession{
+		m:         m,
+		conn:      conn,
+		from:      from,
+		responses: make(chan []byte),
+		queries:   make(chan []byte),
+	}
+	s.ctx, s.closeCtx = context.WithCancel(context.Background())
+	go s.readRoutine()
+	s.handle()
```
bradfitz, line 467: name these so they match names? One is handleReads and the other is handleWrites?

net/dns/manager.go
twitchyliquid64: Changed to `srcAddr` - its converted to the correct type in netstack using `netaddr.IPPortFrom`.

net/dns/manager.go
twitchyliquid64: Good point - fixed

net/dns/manager.go
```diff
@@ -346,6 +348,123 @@ func (m *Manager) Query(ctx context.Context, bs []byte, from netaddr.IPPort) ([]
...
+				}
+				s.m.logf("tcp read (len): %v", err)
+				return
+			}
+			if int(reqLen) > maxReqSizeTCP {
+				s.m.logf("tcp request too large (%d > %d)", reqLen, maxReqSizeTCP)
+				return
+			}
+
+			buff := make([]byte, int(reqLen))
```
bradfitz, line 437: `buf` is more common

bradfitz (approved):

Issue #4324, opened 2022-03-29
FR: support DNS names that refer to "any node with tag T"


josharian:
### What are you trying to do?

I'm spinning up tagged ephemeral Tailscale nodes on fly.io to serve our internal dashboard.

Every time I deploy, new nodes get created and the old nodes get killed.

As a result, I have no user-friendly way to reach the dashboard.

I can't use IPs because those change every time an ephemeral node gets created. I can't use hostname-based DNS because the lifetimes of the nodes overlap, so the entries end up with numeric suffixes.

### How should we solve this?

I'd like to be able to have a DNS name that maps to "any currently-online node with tag T". Maybe something like this in the ACL:

```
 "hosts": {
    "dashboard": ["tag:dashboard:*"],
  },
```

There are lots of potential semantics for how to resolve the DNS name (most recently seen, load balance, lowest ping latency / first to respond to peer request, random). I don't really care which the semantics are.

I'd also be happy with other ways to fix this, as long as they're low effort + low maintenance.

### What is the impact of not solving this?

I will either (1) hack together a script to update the ACL file after every deployment; (2) play subnet router games, as suggested by @maisem, or mostly realistically (3) copy/paste IP addresses until my brain explodes and then go do 1 or 2.

### Anything else?

Related: https://github.com/tailscale/tailscale/issues/465

Somewhat related: https://github.com/tailscale/tailscale/issues/1543

cc @maisem @crawshaw @mayakacz

maisem:
>play subnet router games, as suggested by @maisem

Also referred here

> ... would most likely require https://github.com/tailscale/tailscale/issues/465, or some workaround using subnet routers and externally assigned VIPs.
https://github.com/tailscale/tailscale/issues/502#issuecomment-1053766639


The proposal was to reserve any IP from the RFC1918 space (or really any space), and treat it as a VIP. This VIP is something that the ephemeral nodes would advertise using `--advertise-routes` and can be auto approved using [AutoApprovers]( https://tailscale.com/kb/1018/acls/#auto-approvers-for-routes-and-exit-nodes). You will still have to somehow make sure to listen on that VIP from your code though, and make your code dial out to the VIP instead of the Tailscale IP.

But as you mention, it's not user friendly :-)

mayakacz:
> "any currently-online node with tag T". 

What does this represent? A specific service?

josharian:
>> "any currently-online node with tag T".

> What does this represent? A specific service?

[...]

PR #4599, opened 2022-05-02
tsnet: add Server.AdvertiseTags for plumbing to ipn.Prefs

bmizerany:
This commit makes it possible for embedded Tailscale nodes to advertise their own tags.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -66,6 +66,10 @@ type Server struct {
 	// as an Ephemeral node (https://tailscale.com/kb/1111/ephemeral-nodes/).
 	Ephemeral bool
 
+	// AdvertisedTags are tags to advertise for ACL enforcement.
+	// See ipn.Prefs.AdvertisedTags for more information.
+	AdvertiseTags []string
+
 	initOnce         sync.Once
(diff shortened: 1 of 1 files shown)


DentonGentry:
At the time `--advertise-tags` was implemented in the tailscale CLI, there was no support for specifying tags in authkeys: https://tailscale.com/blog/acl-tags-ga/

Nowadays we'd recommend using authkeys to assign the tags. Before adding the support to tsnet for the mechanism that we don't really want to promulgate: would tags+authkeys work for your use?

bmizerany:
@DentonGentry I had a hunch tsnet omitted AdvertiseTags for reasons, and it wasn't a "TODO". Thank you for cementing this in.

Tags+authkeys works for me but for my (probably weird?) situation where I'm bringing up multiple tsnet.Server's in a single-process unit test to test interactions between the nodes based on certain tag configurations, and it's been nice to modify tags with fluidity without generating authkeys for each permutation and combination I was to test, which I was doing by hand, but can easily use the API.

Just to give another ticket a bump: The goal of the test is to play with tags as a way to perform service discovery because I'm having the same itch @josharian is in #4324.

PR #4600, opened 2022-05-03
control/controlclient: store netinfo and hostinfo separately

maisem:
Currently, when SetNetInfo is called it sets the value on
hostinfo.NetInfo. However, when SetHostInfo is called it overrides all
of the hostinfo object which may race with SetNetInfo.
Instead of trying to solve that race, store NetInfo separately and combine when needed.

Updates #tailscale/corp#4824 (maybe fixes)

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -80,12 +80,12 @@ type Direct struct {
 	sfGroup     singleflight.Group // protects noiseClient creation.
 	noiseClient *noiseClient
 
-	persist      persist.Persist
-	authKey      string
-	tryingNewKey key.NodePrivate
-	expiry       *time.Time
-	// hostinfo is mutated in-place while mu is held.
+	persist       persist.Persist
+	authKey       string
+	tryingNewKey  key.NodePrivate
+	expiry        *time.Time
 	hostinfo      *tailcfg.Hostinfo // always non-nil
+	netinfo       *tailcfg.NetInfo
 	endpoints     []tailcfg.Endpoint
 	everEndpoints bool   // whether we've ever had non-empty endpoints
 	localPort     uint16 // or zero to mean auto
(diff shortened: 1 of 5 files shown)


ipn/ipnlocal/local.go
```diff
@@ -935,8 +935,7 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 	httpTestClient := b.httpTestClient
 
 	if b.hostinfo != nil {
-		hostinfo.Services = b.hostinfo.Services // keep any previous session and netinfo
-		hostinfo.NetInfo = b.hostinfo.NetInfo
+		hostinfo.Services = b.hostinfo.Services // keep any previous services
 	}
 	b.hostinfo = hostinfo
```
raggi, line 940: With the unlock/lock dance that happens higher up, it's very hard for me to assert the safety of this code.

Should the local hostinfo variable be instead intiialized only after the b.cc != nil check and subsequent lock recycle?

control/controlclient/status.go
```diff
@@ -75,9 +74,8 @@ type Status struct {
 	// package, but we have some automated tests elsewhere that need to
 	// use them. Please don't use these fields.
 	// TODO(apenwarr): Unexport or remove these.
-	State    State
-	Persist  *persist.Persist  // locally persisted configuration
-	Hostinfo *tailcfg.Hostinfo // current Hostinfo data
```
raggi, line 80: Is `Hostinfo` just not used here? It's not immediately obvious this is related

ipn/ipnlocal/local.go
maisem: At this point, the lock is still held. It is unlocked further down on line 979

control/controlclient/status.go
maisem: It isn't being used anywhere. See commit message

control/controlclient/direct.go
```diff
@@ -208,7 +208,12 @@ func NewDirect(opts Options) (*Direct, error) {
 	if opts.Hostinfo == nil {
 		c.SetHostinfo(hostinfo.New())
 	} else {
+		ni := opts.Hostinfo.NetInfo
+		opts.Hostinfo.NetInfo = nil
```
bradfitz, line 212: is this ours to mutate? should it be cloned first?

maisem: `// non-nil passes ownership, nil means to use default using os.Hostname, etc`

bradfitz (approved):

PR #4601, opened 2022-05-03
tailcfg: remove some documented DebugFlags that no longer exist

bradfitz:
Update tailscale/corp#5007


danderson (approved):
No longer exist meaning the clients no longer set them?

bradfitz:
> No longer exist meaning the clients no longer set them?

Meaning the server doesn't respect them.

The clients never set them; they were only set via the dev envknob that forced them to be passed through. They were just for us to debug things manually.

danderson:
Ah, I thought clients used this in places, for some reason. Thanks for tidying up!

PR #4613, opened 2022-05-05
util/clientmetric: add Reader type for parsing encoded client metric updates

mihaip:
Extracted from the logprint package in the corp repository.


bradfitz (commented):
Any reason for it to be in this repo?

I think it'd be easier to keep it where it was, but moved out to a package there. Then it'll be in the same repo as its callers.

mihaip:
Figured it made sense to have the reader next to the writer, but happy to move it. Make a new `misc/clientmetric` package?

mihaip:
Abandoned.

PR #4621, opened 2022-05-05
client/tailscale: fix ExpandSNIName on non-default LocalClient

bradfitz:
It was using a mix.

Found by @maisem.


nickoneill (approved):
subtle!

PR #4623, opened 2022-05-05
wgengine/netstack, ipn/ipnlocal: serve http://100.100.100.100/

bradfitz:
For future stuff.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -3313,3 +3313,38 @@ func (b *LocalBackend) HandleSSHConn(c net.Conn) error {
 	}
 	return b.sshServer.HandleSSHConn(c)
 }
+
+// HandleQuad100Port80Conn serves http://100.100.100.100/ on port 80 (and
+// the equivalent tsaddr.TailscaleServiceIPv6 address).
+func (b *LocalBackend) HandleQuad100Port80Conn(c net.Conn) {
+	var s http.Server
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/local.go
```diff
@@ -3313,3 +3313,31 @@ func (b *LocalBackend) HandleSSHConn(c net.Conn) error {
...
+// HandleQuad100Port80Conn serves http://100.100.100.100/ on port 80 (and
+// the equivalent tsaddr.TailscaleServiceIPv6 address).
+func (b *LocalBackend) HandleQuad100Port80Conn(c net.Conn) {
+	var s http.Server
+	s.Handler = http.HandlerFunc(b.handleQuad100Port80Conn)
+	s.Serve(netutil.NewOneConnListener(c, nil))
+}
+
+func (b *LocalBackend) handleQuad100Port80Conn(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("X-Frame-Options", "DENY")
```
twitchyliquid64, line 3326: Could we also set a CSP?

```go
w.Header().Set("Content-Security-Policy", "default-src 'self';")
```

twitchyliquid64 (approved):

ipn/ipnlocal/local.go
bradfitz: Done

maisem (commented):
Nice!

PR #4625, opened 2022-05-05
util/clientmetric: add ResetLastDeltaForTest function

mihaip:
Necessary to force flushing of client metrics more aggressively in
dev/test mode.

### DIFF
--- a/util/clientmetric/clientmetric.go
+++ b/util/clientmetric/clientmetric.go
@@ -255,6 +255,10 @@ func EncodeLogTailMetricsDelta() string {
 	return enc.buf.String()
 }
 
+func ResetLastDeltaForTest() {
+	lastDelta = time.Time{}
+}
+
 var deltaPool = &sync.Pool{
 	New: func() any {
 		return new(deltaEncBuf)


util/clientmetric/clientmetric.go
```diff
@@ -255,6 +255,10 @@ func EncodeLogTailMetricsDelta() string {
 	return enc.buf.String()
 }
 
+func ResetLastDeltaForTest() {
```
bradfitz, line 258: Normally I'd say "all exported stuff needs proper docs" but given as how this is a ForTest func, meh.

bradfitz (approved):

bradfitz:
BTW, one thing you can do to not pollute the package documentation (as viewed on https://pkg.go.dev/tailscale.com/util/clientmetric or local godoc, etc) is something like this, with a package variable of unexported type with public methods:

```go
var TestHooks testHooks

type testHooks struct{}

func (testHooks) ResetLastDelta() { .... }
```

Then the only thing that's rendered in docs is a single line without a TOC entry:

```
var TestHooks testHooks
```

bradfitz:
Runnable example of that pattern: https://go.dev/play/p/dmr8N6tmBTr

mihaip:
`TestHooks` var approach works for me, switched to that.

mihaip:
> `TestHooks` var approach works for me, switched to that.

Except I forgot to actually stage the changes 🤦‍♂️ . Will send a followup.

PR #4633, opened 2022-05-06
ssh/tailssh: exec into `login` when launching a shell

### LINKED ISSUE #4627: ssh/tailssh: motd support
We don't do pam_motd. (https://linux.die.net/man/8/pam_motd)

OpenSSH Ubuntu default is:

```
UsePAM yes
```

Ubuntu has `/etc/pam.d/login` with:

```
# (Replaces the `MOTD_FILE' option in login.defs)
# This includes a dynamically generated part from /run/motd.dynamic
# and a static (admin-editable) part from /etc/motd.
session    optional   pam_motd.so motd=/run/motd.dynamic
session    optional   pam_motd.so noupdate
```

Its motd is updated by http://manpages.ubuntu.com/manpages/trusty/man5/update-motd.5.html

Maybe we should just best effort parse `/etc/pam.d/login` etc and try to act alike.

Worst case no MOTD (like today) and not the end of the world.

/cc @maisem @danderson @raggi @Xe

maisem:
This has the added benefit of displaying the MOTD and reducing our
dependency on the DBus interface.

Fixes #4627
Updates #3802

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -51,7 +51,7 @@ var ptyName = func(f *os.File) (string, error) {
 // On success, it may return a non-nil close func which must be closed to
 // release the session.
 // See maybeStartLoginSessionLinux.
-var maybeStartLoginSession = func(logf logger.Logf, uid uint32, localUser, remoteUser, remoteHost, tty string) (close func() error, err error) {
+var maybeStartLoginSession = func(logf logger.Logf, ia incubatorArgs) (close func() error, err error) {
 	return nil, nil
 }
 
@@ -133,6 +133,38 @@ func (stdRWC) Close() error {
 	return nil
 }
 
+type incubatorArgs struct {
+	uid          uint64
+	gid          int
+	groups       string
+	localUser    string
+	remoteUser   string
+	remoteIP     string
+	ttyName      string
+	hasTTY       bool
+	cmdName      string
+	isSFTP       bool
+	loginCmdPath string
+	cmdArgs      []string
+}
+
+func parseIncubatorArgs(args []string) (a incubatorArgs) {
+	flags := flag.NewFlagSet("", flag.ExitOnError)
+	flags.Uint64Var(&a.uid, "uid", 0, "the uid of local-user")
+	flags.IntVar(&a.gid, "gid", 0, "the gid of local-user")
+	flags.StringVar(&a.groups, "groups", "", "comma-separated list of gids of local-user")
+	flags.StringVar(&a.localUser, "local-user", "", "the user to run as")
+	flags.StringVar(&a.remoteUser, "remote-user", "", "the remote user/tags")
+	flags.StringVar(&a.remoteIP, "remote-ip", "", "the remote Tailscale IP")
+	flags.StringVar(&a.ttyName, "tty-name", "", "the tty name (pts/3)")
+	flags.BoolVar(&a.hasTTY, "has-tty", false, "is the output attached to a tty")
+	flags.StringVar(&a.cmdName, "cmd", "", "the cmd to launch (ignored in sftp mode)")
+	flags.BoolVar(&a.isSFTP, "sftp", false, "run sftp server (cmd is ignored)")
+	flags.Parse(args)
+	a.cmdArgs = flags.Args()
+	return a
+}
+
 // beIncubator is the entrypoint to the `tailscaled be-child ssh` subcommand.
 // It is responsible for informing the system of a new login session for the user.
 // This is sometimes necessary for mounting home directories and decrypting file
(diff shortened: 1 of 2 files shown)


ssh/tailssh/incubator.go
```diff
@@ -170,6 +170,15 @@ func beIncubator(args []string) error {
 	}
 
 	euid := uint64(os.Geteuid())
+	if !*sftpMode && len(cmdArgs) == 1 && cmdArgs[0] == "-l" && euid == 0 {
+		// If we are trying to launch a login shell, just exec into login instead.
+		lp, err := exec.LookPath("login")
+		if err != nil {
+			return err
+		}
+		return unix.Exec(lp, []string{lp, "-f", *localUser}, subprocEnviron())
```
danderson, line 179: `login` clears the environment by default, so subprocEnviron won't be applied.

You can pass `-p` to preserve the environment, but then see below re: additional environment sanitization.

ssh/tailssh/incubator.go
```diff
@@ -233,6 +242,19 @@ func beIncubator(args []string) error {
 	return cmd.Run()
 }
 
+// subprocEnviron filters and returns subset of os.Environ.
+func subprocEnviron() []string {
+	all := os.Environ()
+	ret := all[:0]
+	for _, e := range all {
+		if strings.HasPrefix(e, "PATH=") {
+			continue
```
danderson, line 251: Why do we remove PATH? This just makes me wonder what else is in this environment that needs to be nuked. Is tailscaled setting a restricted env for the incubator?

ssh/tailssh/incubator.go
```diff
@@ -252,6 +274,7 @@ func (ss *sshSession) launchProcess() error {
 
 	ci := ss.conn.info
 	cmd.Env = append(cmd.Env,
```
danderson, line 276: Style thing: rather than append to Env, could you make this an assignment? `cmd.Env = []string{...}`. That way it's 100% obvious what the child's environment will be, without having to trace down the chain of newIncubatorCommand and exec.CommandContext to figure out if they inherit the current env.

(this is important because sanitizing the environment has some security implications, so being able to quickly reason about what env is getting set matters a lot)

ssh/tailssh/incubator.go
danderson: Ah, I see below that it's kind of so that the incubator can find stuff like login. Hmm. At least add a comment here explaining why PATH is special? But also I'm wondering if we can do all those path lookups in the parent and keep the incubator less surprising.

ssh/tailssh/incubator.go
```diff
@@ -233,6 +242,19 @@ func beIncubator(args []string) error {
 	return cmd.Run()
 }
 
+// subprocEnviron filters and returns subset of os.Environ.
+func subprocEnviron() []string {
+	all := os.Environ()
```
danderson, line 247: Could we make this function use an allowlist? I'm scared of things getting injected into the environment by stuff like the dynamic loader, when we really only want to pass through SSH_CONNECTION and SSH_CLIENT and nothing else.

ssh/tailssh/incubator.go
```diff
@@ -443,6 +465,7 @@ func (ss *sshSession) startWithPTY() (ptyFile *os.File, err error) {
 	cmd.Stdout = tty
 	cmd.Stderr = tty
 
+	ss.logf("starting pty command: %+v", cmd.Args)
```
danderson, line 468: Excessively spammy? Or do we want this for an audit log?

danderson (approved):
General principle LGTM, but I'd like to tighten up environment sanitization a bit, for easier future audit of "okay what environment will actually get set"

ssh/tailssh/incubator.go
maisem: this is moving and not a new log line, I agree it's spammy but I don't want to drop it yet. Also it's not terrible because it's not like you get multiple ssh reqs/s

ssh/tailssh/incubator.go
maisem: I initially passed in `nil` but that made my `TERM` variable not work. So maybe `login` does something more fancy?

ssh/tailssh/incubator.go
maisem: I removed the PATH and instead just passed the path for `login` as a flag

ssh/tailssh/incubator.go
```diff
@@ -160,11 +164,16 @@ func parseIncubatorArgs(args []string) (a incubatorArgs) {
...
 	flags.StringVar(&a.cmdName, "cmd", "", "the cmd to launch (ignored in sftp mode)")
 	flags.BoolVar(&a.isSFTP, "sftp", false, "run sftp server (cmd is ignored)")
+	flags.StringVar(&a.loginCmdPath, "login-cmd", "", "the path to `login` cmd")
 	flags.Parse(args)
 	a.cmdArgs = flags.Args()
 	return a
 }
 
+func (a incubatorArgs) isShell() bool {
+	return len(a.cmdArgs) == 1 && a.cmdArgs[0] == "-l"
```
danderson, line 174: Probably quite false-positivey. `ls -l` matches this.

Pass an incubatorArg for "am I spawning a login shell" ?

ssh/tailssh/incubator.go
```diff
@@ -186,6 +195,11 @@ func beIncubator(args []string) error {
 	}
 
 	euid := uint64(os.Geteuid())
+	runningAsRoot := euid == 0
+	if runningAsRoot && ia.isShell() && !ia.isSFTP && ia.loginCmdPath != "" {
+		// If we are trying to launch a login shell, just exec into login instead.
+		return unix.Exec(ia.loginCmdPath, []string{ia.loginCmdPath, "-f", ia.localUser, "-h", ia.remoteIP}, os.Environ())
```
danderson, line 201: Still doesn't preserve the os.Environ() you pass it, `login` nukes the environment by default. If you want to preserve it, need to pass `-p`

danderson: Speaking of that, where did the env sanitization go? We still want SSH_CONNECTION and SSH_CLIENT envvars to be passed through, but this is just passing through all of os.Environ(), which still contains PATH and stuff, I think?

danderson: Basically I'm worried about if we're strictly controlling what ends up in this environment. Is it exactly SSH_* (generated by us) plus the allowlisted envs that the client can set before forking the shell?

danderson (approved):

ssh/tailssh/incubator.go
maisem: Done

ssh/tailssh/incubator.go
maisem: >Speaking of that, where did the env sanitization go? We still want SSH_CONNECTION and SSH_CLIENT envvars to be passed through, but this is just passing through all of os.Environ(), which still contains PATH and stuff, I think?

This happens in `tailscaled`, the env passed down to the `incubator` is always sanitized.

danderson: Still not passing `-p` to the login process, so it's still going to nuke the entire parent environment.

maisem: Done

danderson (approved):

Issue #4635, opened 2022-05-07
Add support for maps in views


maisem:
Figure out how to expose maps in views. Currently they are unsupported.

maisem:
I tried linking this from my phone but it just created an issue without linking it ¯\_(ツ)_/¯ 
https://github.com/tailscale/tailscale/pull/4589#issuecomment-1120151741

cc @dsnet

DentonGentry:
I think this is addressed by https://github.com/tailscale/tailscale/commit/d04afc697cbb81fb765fdee8847b8e33e34537be, I don't know what else there is to do.

Barring further comment, we'll expect to close this during the next regular bug scrub.

PR #4636, opened 2022-05-07
cmd/viewer,types/views: add support for views of maps

maisem:
Updates #4635

### DIFF
--- a/types/views/views.go
+++ b/types/views/views.go
@@ -9,6 +9,8 @@ package views
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
+	"reflect"
 
 	"inet.af/netaddr"
 	"tailscale.com/net/tsaddr"
@@ -97,8 +99,14 @@ type Slice[T any] struct {
 	ж []T
 }
 
-// SliceOf returns a Slice for the provided slice.
-func SliceOf[T any](x []T) Slice[T] { return Slice[T]{x} }
+// SliceOf returns a Slice for the provided slice for immutable values.
+// It panics if the value type contains pointers.
+func SliceOf[T any](x []T) Slice[T] {
+	if ev := reflect.TypeOf(x).Elem(); containsMutable(ev) {
+		panic(fmt.Sprintf("slice value type %q has pointers", ev.Name()))
+	}
+	return Slice[T]{x}
+}
 
 // MarshalJSON implements json.Marshaler.
 func (v Slice[T]) MarshalJSON() ([]byte, error) {
@@ -186,8 +194,52 @@ func (v *IPPrefixSlice) UnmarshalJSON(b []byte) error {
 	return v.ж.UnmarshalJSON(b)
 }
 
-// MapOf returns a read-only view over m.
+// containsMutable reports whether the provided type has anything mutable.
+func containsMutable(t reflect.Type) bool {
+	switch x := fmt.Sprintf("%v.%v", t.PkgPath(), t.Name()); x {
+	case "time.Time",
+		"inet.af/netaddr.IP":
+		return false
+	}
+	k := t.Kind()
+	switch k {
+	case reflect.Bool,
+		reflect.Int,
+		reflect.Int8,
+		reflect.Int16,
+		reflect.Int32,
+		reflect.Int64,
+		reflect.Uint,
+		reflect.Uint8,
+		reflect.Uint16,
+		reflect.Uint32,
+		reflect.Uint64,
+		reflect.Float32,
+		reflect.Float64,
+		reflect.Complex64,
+		reflect.Complex128,
+		reflect.String:
+		return false
+	case reflect.Array: // Not a slice.
+		return containsMutable(t.Elem()) && t.Len() > 0
+	case reflect.Struct:
+		for i := 0; i < t.NumField(); i++ {
+			f := t.Field(i)
+			if containsMutable(f.Type) {
+				return true
+			}
+		}
+		return false
+	}
+	return true
+}
+
+// MapOf returns a read-only view over m for immutable values.
+// It panics if the value type contains pointers.
 func MapOf[K comparable, V comparable](m map[K]V) Map[K, V] {
+	if ev := reflect.TypeOf(m).Elem(); containsMutable(ev) {
+		panic(fmt.Sprintf("map value type %q has pointers", ev.Name()))
+	}
 	return Map[K, V]{m}
 }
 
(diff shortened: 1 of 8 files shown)


types/views/views.go
```diff
@@ -181,3 +181,80 @@ func (v IPPrefixSlice) MarshalJSON() ([]byte, error) {
...
 }
+
+// MapOf returns a read-only view over m.
+func MapOf[K comparable, V any](m map[K]V) Map[K, V] {
+	return Map[K, V]{m}
+}
+
+// Map is a read-only accessor over a map.
+type Map[K comparable, V any] struct {
+	// It is named distinctively to make you think of how dangerous it is to escape
```
bradfitz, line 192: This seems to be the old comment. Didn't we flesh it out elsewhere?

types/views/views.go
```diff
@@ -181,3 +181,80 @@ func (v IPPrefixSlice) MarshalJSON() ([]byte, error) {
 func (v *IPPrefixSlice) UnmarshalJSON(b []byte) error {
 	return v.ж.UnmarshalJSON(b)
 }
+
+// MapOf returns a read-only view over m.
+func MapOf[K comparable, V any](m map[K]V) Map[K, V] {
+	return Map[K, V]{m}
+}
+
+// Map is a read-only accessor over a map.
```
bradfitz, line 190: say which variant this is for?

types/views/views.go
```diff
@@ -181,3 +181,80 @@ func (v IPPrefixSlice) MarshalJSON() ([]byte, error) {
 func (v *IPPrefixSlice) UnmarshalJSON(b []byte) error {
 	return v.ж.UnmarshalJSON(b)
 }
+
+// MapOf returns a read-only view over m.
+func MapOf[K comparable, V any](m map[K]V) Map[K, V] {
+	return Map[K, V]{m}
```
bradfitz, line 187: should this panic if V's type has pointers in it?

types/views/views.go
```diff
@@ -181,3 +181,80 @@ func (v IPPrefixSlice) MarshalJSON() ([]byte, error) {
...
+}
+
+// ForEach calls f for every k,v par in the underlying map.
+func (m Map[K, V]) ForEach(f func(k K, v V)) {
+	for k, v := range m.ж {
+		f(k, v)
+	}
+}
+
+// ForEach calls f for every k,v par in the underlying map.
```
bradfitz, line 224: par? pair? entry?

types/views/views.go
```diff
@@ -181,3 +181,80 @@ func (v IPPrefixSlice) MarshalJSON() ([]byte, error) {
...
+	}
+}
+
+// MapFn is like Map but with a func to convert values from T to V.
+// It is used to provide map of slices and views.
+type MapFn[K comparable, T any, V any] struct {
+	// It is named distinctively to make you think of how dangerous it is to escape
+	// to callers. You must not let callers be able to mutate it.
+	ж  map[K]T
+	fn func(T) V
```
bradfitz, line 238: maybe `wrapv` instead of `fn`?

types/views/views.go
maisem: Oh yeah, I'll update this file too. We updated the generator, these are still manually generated

types/views/views.go
maisem: PTAL

types/views/views.go
```diff
@@ -181,3 +185,90 @@ func (v IPPrefixSlice) MarshalJSON() ([]byte, error) {
...
+	return v, ok
+}
+
+// ForEach calls f for every k,v pair in the underlying map.
+func (m Map[K, V]) ForEach(f func(k K, v V)) {
+	for k, v := range m.ж {
+		f(k, v)
+	}
+}
+
```
dsnet, line 287: Might be worth exposing a `Has` method too since they're useful for using within a Go expression.

types/views/views.go
```diff
@@ -181,3 +185,90 @@ func (v IPPrefixSlice) MarshalJSON() ([]byte, error) {
...
+func MapFnOf[K comparable, T any, V any](m map[K]T, f func(T) V) MapFn[K, T, V] {
+	return MapFn[K, T, V]{
+		ж:     m,
+		wrapv: f,
+	}
+}
+
+// MapFn is like Map but with a func to convert values from T to V.
+// It is used to provide map of slices and views.
+type MapFn[K comparable, T any, V any] struct {
```
dsnet, line 298: You already have a `ViewCloner` constraint, I think we should make that less specific on structs and more general purpose.

When I was exploring this problem, I came up with: https://go.dev/play/p/Kn5HUXEWMwe

dsnet: In my snippet, my `Viewer` constraint analogous to your `ViewCloner` constraint.

types/views/views.go
maisem: We can't generate it because `ж` is unexported

types/views/views.go
```diff
@@ -53,6 +53,8 @@ func SliceOfViews[T ViewCloner[T, V], V StructView[T]](x []T) SliceView[T, V] {
 // SliceView is a read-only wrapper around a struct which should only be exposed
 // as a View.
 type SliceView[T ViewCloner[T, V], V StructView[T]] struct {
+	// ж is the underlying mutable value, named with a hard-to-type
```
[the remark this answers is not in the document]
maisem: I'd rather keep it here to be consistent with all other uses of Views. Even here it serves the purpose to being explicit that the underlying value should be handled with care.

types/views/views.go
```diff
@@ -181,3 +193,152 @@ func (v IPPrefixSlice) MarshalJSON() ([]byte, error) {
 func (v *IPPrefixSlice) UnmarshalJSON(b []byte) error {
 	return v.ж.UnmarshalJSON(b)
 }
+
+// containsPointers reports whether the provided type is a pointer or wraps any
+// pointers.
+func containsPointers(t reflect.Type) bool {
```
dsnet, line 199: (I thought I made this comment, but I can't find any record of it. Strange.)

I think this is too narrow. What we really want is to know whether this contains any mutable values.

In Go, this can handle through the following kinds:
* pointers
* slices (since the user can overwrite elements or take the address of them)
* maps (since the user can overwrite or delete an entry value)

We may also want to reject the following kinds:
* channels
* funcs

since those are technically mutable (but there is also no obvious immutable view of them).

types/views/views.go
```diff
@@ -181,3 +193,147 @@ func (v IPPrefixSlice) MarshalJSON() ([]byte, error) {
...
+		reflect.Uint32,
+		reflect.Uint64,
+		reflect.Float32,
+		reflect.Float64,
+		reflect.Complex64,
+		reflect.Complex128,
+		reflect.String:
+		return false
+	case reflect.Array: // Not a slice.
+		return t.Len() > 0
```
dsnet, line 225: This should be:
```go
return containsMutable(t.Elem()) && t.Len() > 0
```

dsnet (approved):

types/views/views.go
maisem: yep! done.

PR #4637, opened 2022-05-07
net/dns: add MagicDNS DNS-over-TLS support

bradfitz:
Updates #915


bradfitz:
I've only tested that this compiles and I can connect to it with `openssl s_client`.

Other than that, I have no clue whether this actually works. (weekend hacking during kids' cartoons and time is gone)

net/dns/manager.go
```diff
@@ -454,17 +463,61 @@ func (s *dnsTCPSession) handleReads() {
...
+		srcAddr:     srcAddr,
+		responses:   make(chan []byte),
+		readClosing: make(chan struct{}),
 	}
 	s.ctx, s.closeCtx = context.WithCancel(context.Background())
 	go s.handleReads()
 	s.handleWrites()
 }
 
+// HandleTCPConn implements magicDNS over DNS-over-TLS, taking a connection and
```
DentonGentry, line 477: :s/HandleTCPConn/HandleDNSoverTLSConn/

net/dns/manager.go
```diff
@@ -454,17 +463,61 @@ func (s *dnsTCPSession) handleReads() {
...
+// servicing DNS requests sent down it.
+func (m *Manager) HandleDNSoverTLSConn(conn net.Conn, srcAddr netaddr.IPPort) {
```
danderson, line 479: I don't understand how this will work to Android's satisfaction. The TLS cert will be invalid, and we have no way of generating a valid one. Does android just not care?...

net/dns/manager.go
```diff
@@ -454,17 +463,61 @@ func (s *dnsTCPSession) handleReads() {
...
+		conn.Close()
+		return
+	}
+	template := x509.Certificate{
+		SerialNumber: big.NewInt(1),
+		Subject: pkix.Name{
+			Organization: []string{"Tailscale MagicDNS"},
+		},
+		NotBefore: time.Now(),
+		NotAfter:  time.Now().Add(time.Hour * 24 * 180),
```
maisem, line 492: Oh no, this is going to be terrible for tagged nodes (key expiry disabled)

net/dns/manager.go
```diff
@@ -454,17 +463,61 @@ func (s *dnsTCPSession) handleReads() {
...
 	}
 	s.ctx, s.closeCtx = context.WithCancel(context.Background())
 	go s.handleReads()
 	s.handleWrites()
 }
 
+// HandleTCPConn implements magicDNS over DNS-over-TLS, taking a connection and
+// servicing DNS requests sent down it.
+func (m *Manager) HandleDNSoverTLSConn(conn net.Conn, srcAddr netaddr.IPPort) {
+	privKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
DentonGentry, line 480: This will work in Android's PrivateDNS=Automatic mode, where it opportunistically tries to use DoT and does not validate the certificate it receives.

With Android's PrivateDNS=On setting I don't think this will be sufficient, it wants to see a real certificate signed by a root of trust. We should be able to do that eventually if the tailnet has certificate support set up and we're willing to proactively obtain a cert for android-name.tails-scales.ts.net.

net/dns/manager.go
maisem: Is there an upstream bug we can track that says that android doesn't care?

bradfitz: Yeah see the bug. There are two modes.

net/dns/manager.go
bradfitz: I don't know what that has to do with anything here?

maisem: Isn't this cert going to expire after 6 months? I assume you picked 6 months because it's the same as the key expiry period?

bradfitz: It's making a new cert on each TCP connection!

This was a proof of concept to see if it made Android happy.

6 months has nothing to do with key expiry or anything. It's copy/paste from some Go cert generation code elsewhere.

The latest version caches the cert, which is bad in a new way: if your Tailscale process runs for 6 months your cert can expire. But that means you're not restarting Tailscale or your phone or applying security updates for 6 months, which seems unlikely.

maisem (approved):

bradfitz:
I don't plan to either merge this or further refine it until somebody tests it on Android. @DentonGentry is probably the most capable Android+DNS tester here. :)

DentonGentry:
This does seem to work on Android.

I've set PrivateDNS to Automatic, and added several log messages. With PrivateDNS=Automatic Android silently falls back to UDP as needed, the logs let me see that it is using TLS.

```
func (m *Manager) HandleDNSoverTLSConn(conn net.Conn, srcAddr netaddr.IPPort) {
        ...
        m.logf("HandleDNSoverTLSConn: new connection %v", tlsConn.RemoteAddr())
        m.HandleTCPConn(tlsConn, srcAddr)
}
func (s *dnsTCPSession) handleReads() {
...
                                s.m.logf("handleReads DNS from %v", s.conn.RemoteAddr())
                                go s.handleQuery(buf)

```

I see the log messages in `adb logcat -s "com.tailscale.ipn:*"`
```
05-08 08:13:32.390 18446 18470 I com.tailscale.ipn: 27.5M/123.1M dns: HandleDNSoverTLSConn: new connection 10.1.10.131:44872
05-08 08:13:32.401 18446 18470 I com.tailscale.ipn: 27.5M/123.1M dns: handleReads DNS from 10.1.10.131:44872
05-08 08:13:43.815 18446 18477 I com.tailscale.ipn: 28.2M/123.1M dns: HandleDNSoverTLSConn: new connection 100.124.110.51:38206
05-08 08:13:43.822 18446 18497 I com.tailscale.ipn: 28.2M/123.1M dns: handleReads DNS from 100.124.110.51:38206
05-08 08:13:43.823 18446 18497 I com.tailscale.ipn: 28.2M/123.1M dns: handleReads DNS from 100.124.110.51:38206
...
05-08 08:14:12.607 18446 18465 I com.tailscale.ipn: 18.0M/123.1M dns: handleReads DNS from 100.124.110.51:38206
05-08 08:14:43.809 18446 18470 I com.tailscale.ipn: 18.4M/123.1M dns: HandleDNSoverTLSConn: new connection 100.124.110.51:38278
05-08 08:14:43.819 18446 18470 I com.tailscale.ipn: 18.5M/123.1M dns: handleReads DNS from 100.124.110.51:38278
05-08 08:14:43.820 18446 18470 I com.tailscale.ipn: 18.5M/123.1M dns: handleReads DNS from 100.124.110.51:38278
...
05-08 08:15:02.763 18446 18470 I com.tailscale.ipn: 19.0M/123.1M dns: handleReads DNS from 100.124.110.51:38278
05-08 08:15:29.657 18446 18467 I com.tailscale.ipn: 19.7M/123.5M dns: HandleDNSoverTLSConn: new connection 100.124.110.51:38294
05-08 08:15:29.672 18446 18465 I com.tailscale.ipn: 19.8M/123.5M dns: handleReads DNS from 100.124.110.51:38294
05-08 08:15:29.673 18446 18465 I com.tailscale.ipn: 19.8M/123.5M dns: handleReads DNS from 100.124.110.51:38294
05-08 08:16:02.813 18446 18470 I com.tailscale.ipn: 17.3M/123.6M dns: HandleDNSoverTLSConn: new connection 100.124.110.51:38298
05-08 08:16:02.827 18446 18476 I com.tailscale.ipn: 17.4M/123.6M dns: handleReads DNS from 100.124.110.51:38298
...etc, etc...
```

Once Android sets up a TLS connection it uses it for a while. If I leave the phone sitting idle, it will set up a new TLS connection the next time I use it.

----

This PR gets us a step closer to having PrivateDNS=On work. I think the remaining steps (to be done later) would include:
1. Issue a real certificate. We'd have to think through the right tradeoff regarding the public ledger vs the user having to configure things.
1. The user has to type a hostname into the Private DNS config. That hostname needs to match the certificate, and DNS needs to resolve to 100.100.100.100 (or something).

PrivateDNS-On

bradfitz:
> This does seem to work on Android.

So previously only "Off" worked and now both "Off" and "Automatic" work?

That is, do we gain anything with this change as-is before issuing actual certs?

One thing we could do for actual certs is issue them out of a different namespace, like `<NODE-ID-HEX>._node.ts.net`. But I don't understand how Android does bootstrap DNS resolution of the DoT server. Does it look up that "Private DNS provider hostname" in your screenshot using UDP DNS?

DentonGentry:
PrivateDNS=Automatic worked even without this change, in that it will fall back to UDP for everything. By itself this change wouldn't accomplish much, its main benefit would be as a step toward PrivateDNS=On.

At least with Android 12 when you switch to PrivateDNS = On, it helpfully pre-populates `dns.google`. If you want anything else you have to type it out.

dns google

In PrivateDNS=On mode, bootstrapping uses regular UDP/TCP DNS to look up "dns.google" or "1dot1dot1dot1.cloudflare-dns.com" or "pixel3.tails-scales.ts.net" or whatever, then makes a TLS connection and asserts that the offered certificate passes validation and matches the name.

[It might have special handling with hard-coded IP addresses for some of those, I guess]

DentonGentry:
This change does help with a case in Linux: systemd-resolved can be set to always use DNS-over-TLS, and right now enabling MagicDNS will break if the user has enabled this setting.

bradfitz:
> This change does help with a case in Linux: systemd-resolved can be set to always use DNS-over-TLS, and right now enabling MagicDNS will break if the user has enabled this setting.

Okay, @maisem, you win.

(He was just making the point that people might use this on servers and I was like, "Why!?!?")

bradfitz:
So clearly we need to get a cert for `magic.ts.net` and add a MagicDNS DNS entry from that to 100.100.100.100 and then just give out all phones that cert (including its private key!) 🤪

DentonGentry:
It could also be `magic.tails-scales.ts.net`.
Or both, one if you've enabled certificates on your tailnet and the other available if you haven't.

bradfitz:
> Or both, one if you've enabled certificates on your tailnet and the other available if you haven't.

I'm trying to think of ways that avoid opt-in/info-leakage stuff. And nicer to have a short easy-to-document name rather than documenting the steps to look up the name.

*especially* if this is something users will have to manually configure every time they turn Tailscale on/off? Or can we programmatically update this Private DNS setting from the app?

danderson:
Does android let us specify both the IP and expected hostname? If so, we could park something like dot.ts.net and issue certs for hash-like names under that.

We definitely need to keep the private keys per-device though, distributing a private key would get our cert promptly burned.

DentonGentry:
> android let us specify both the IP and expected hostname

Android requires the hostname, which it expects to be the certificate name and also to resolve to the IP address of the DoT server (using regular UDP-based DNS).

It is entered as a text string. I find no API for us to populate it as a selection, we'd presumably want to copy it to the clipboard to be pasted into place.

PrivateDNS-On

net/dns/manager.go
```diff
@@ -454,17 +469,87 @@ func (s *dnsTCPSession) handleReads() {
...
+// mode doesn't verify certs.
+//
+// See https://github.com/tailscale/tailscale/issues/915.
+func genSelfSignedDoTCert() (tls.Certificate, error) {
+	privKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		return tls.Certificate{}, err
+	}
+	template := x509.Certificate{
+		SerialNumber: big.NewInt(1),
```
twitchyliquid64, line 514: The serial number needs to be unique / randomly generated. Some copypasta:

```go
	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
```

net/dns/manager.go
```diff
@@ -454,17 +469,87 @@ func (s *dnsTCPSession) handleReads() {
...
+		Subject: pkix.Name{
+			Organization: []string{"Tailscale MagicDNS"},
```
twitchyliquid64, line 516: It might be a good idea to set the common name to something device specific too if such a field is easily obtainable.

twitchyliquid64 (commented):
LGTM for the self-signed cert stuff, but I'm not sure about issuing TLS certs for `magic.ts.net` with a non-private key + public CA: its a significant enough deviation from the web PKI model to warrant caution, and could even be incompatible with CA obligations (need to double check that one).

PR #4652, opened 2022-05-10
types/views: remove alloc in hot path

maisem:

### DIFF
--- a/types/views/views.go
+++ b/types/views/views.go
@@ -9,8 +9,6 @@ package views
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
-	"reflect"
 
 	"inet.af/netaddr"
 	"tailscale.com/net/tsaddr"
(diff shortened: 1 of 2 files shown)


types/views/views.go
```diff
@@ -196,9 +196,8 @@ func (v *IPPrefixSlice) UnmarshalJSON(b []byte) error {
 
 // containsMutable reports whether the provided type has anything mutable.
 func containsMutable(t reflect.Type) bool {
```
bradfitz, line 198: we need to memoize the crap out of this

bradfitz (approved):

types/views/views.go
```diff
@@ -100,11 +98,8 @@ type Slice[T any] struct {
 }
 
 // SliceOf returns a Slice for the provided slice for immutable values.
-// It panics if the value type contains pointers.
+// It is the callers responsibility to make sure V is immutable.
```
bradfitz, line 101: caller's

bradfitz (approved):

PR #4658, opened 2022-05-11
cmd/tailscaled: let SOCKS5 listen on a unix socket

### LINKED ISSUE #4657: FR: Support SOCKS proxy binding to a unix socket

I am using TS in a shared userspace environment where it would be better to avoid opening ports if possible. This could be accomplished by using the same `file:` syntax [already established](https://github.com/tailscale/tailscale/blob/87b44aa311aa535f243c11ff53e32f1b4217f9c2/cmd/tailscale/cli/up.go#L153) with the `tailscale --auth-key` arg:

```
tailscaled -tun="userspace-networking" -socks5-server="file:/example/proxy.sock"
```

This would then let me mount the path into a userspace container for other software, without other users in the shared environment having access to the proxy.

### How should we solve this?

Should be a small change to the `net.Listen` [during setup](https://github.com/tailscale/tailscale/blob/87b44aa311aa535f243c11ff53e32f1b4217f9c2/cmd/tailscaled/tailscaled.go#L601).

I got this working with a small patch, and will file a PR shortly.

Some possible workarounds for my use case:

1. Modify the containers to run tailscaled inside them instead (not ideal; the containers many and customer-provided)

2. Run tailscaled as a sidecar container and bridge the networking over (more palatable, would have to check how that plays with the [singularity container runtime](https://sylabs.io/guides/3.5/user-guide/introduction.html) we sometimes use.

### Anything else?

_No response_

kofalt:
Allows userspace networking without binding a port.

Uses the same `file:` syntax [already established](https://github.com/tailscale/tailscale/blob/87b44aa311aa535f243c11ff53e32f1b4217f9c2/cmd/tailscale/cli/up.go#L153) with the `tailscale --auth-key` arg:

```
tailscaled -tun="userspace-networking" -socks5-server="file:/example/proxy.sock"
```

Ideally, this patch would also clean up the file when tailscaled exits. If that's needed for this to merge, I was hoping a maintainer could add that or provide a suggestion.

Closes #4657

### DIFF
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -598,9 +598,14 @@ func mustStartProxyListeners(socksAddr, httpAddr string) (socksListener, httpLis
 
 	var err error
 	if socksAddr != "" {
-		socksListener, err = net.Listen("tcp", socksAddr)
+		socksType := "tcp"
+		if strings.HasPrefix(socksAddr, "file:") {
+			socksType = "unix"
+			socksAddr = strings.TrimPrefix(socksAddr, "file:")
+		}
+		socksListener, err = net.Listen(socksType, socksAddr)
 		if err != nil {
-			log.Fatalf("SOCKS5 listener: %v", err)
+			log.Fatalf("SOCKS5 %s listener: %v", socksType, err)
 		}
 		if strings.HasSuffix(socksAddr, ":0") {
 			// Log kernel-selected port number so integration tests


DentonGentry:
It looks like the PR description doesn't match the content of the commit message. The commit message just has a signed-off-by line. Could you add the description to the commit as well?

It would be great to have a test somehow, or we're likely to break this support at some point.

kofalt:
Hi Denton, nice to meet you over here as well. I wasn't sure what style was ideal so I just aped parent commit 87b44aa3. I've updated the commit to include some extra detail - happy to make further changes.

As for testing, this is how I was checking things:

```sh
# The curl team is currently discussing a PR to support unix sockets:
# https://curl.se/mail/archive-2022-04/0017.html
# https://github.com/curl/curl/pull/8668

# For now, just turn the socket back into a port:
socat TCP-LISTEN:1080,fork UNIX-CONNECT:"/example/proxy.sock"

# Expect a healthy HTTP response
ALL_PROXY="socks5://localhost:1080/" curl --connect-timeout 1 "some-ts-peer's-server"
```

To write a test for this behavior, you'd want to speak to the open socket on an established tailnet using Go instead.
I'm not familiar with your test suites to accomplish that.

bradfitz:
It's great that curl added support for SOCKS-over-Unix sockets!

But now that curl can do that, let's just use the existing Tailscaled unix socket (`/var/run/tailscale/tailscaled.sock`). That already speaks different protocols so we can make it also supports SOCKS5.

kofalt:
Sounds fine to me. If that's how you'd like it to work I will probably close this and let someone else take a stab at #4657. This was just a quick hack; working with the main socket & protocol detection is beyond my familiarity level.

DentonGentry:
Closing in favor of https://github.com/tailscale/tailscale/issues/4657

PR #4659, opened 2022-05-11
util/clientmetric: add gauge_ name prefix when uploading names

mihaip:
The prefix is a signal to tsweb to treat this as a gauge metric when
generating the Prometheus version.

### DIFF
--- a/util/clientmetric/clientmetric.go
+++ b/util/clientmetric/clientmetric.go
@@ -241,7 +241,7 @@ func EncodeLogTailMetricsDelta() string {
 			m.wireID = numWireID
 		}
 		if m.lastNamed.IsZero() || now.Sub(m.lastNamed) > metricLogNameFrequency {
-			enc.writeName(m.Name())
+			enc.writeName(m.Name(), m.Type())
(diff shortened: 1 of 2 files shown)


util/clientmetric/clientmetric.go
```diff
@@ -271,9 +271,16 @@ type deltaEncBuf struct {
 // writeName writes a "name" (N) record to the buffer, which notes
 // that the immediately following record's wireID has the provided
 // name.
-func (b *deltaEncBuf) writeName(name string) {
+func (b *deltaEncBuf) writeName(name string, typ Type) {
+	var name_prefix string
```
bradfitz, line 275: Go doesn't really use underscores in variable names. They're super rare. https://go.dev/doc/effective_go#mixed-caps

bradfitz (approved):

util/clientmetric/clientmetric.go
mihaip: Brain fart. It'll take a little while longer for ~10 years of Python snake_case naming to wear off.

PR #4672, opened 2022-05-13
wgengine/filter: preallocate some hot slices in MatchesFromFilterRules

twitchyliquid64:
Profiling identified this as a fairly hot path for growing a slice.

Given this is only used in control & when a new packet filter is recieved, this shouldnt be hot in the client.

### DIFF
--- a/wgengine/filter/tailcfg.go
+++ b/wgengine/filter/tailcfg.go
@@ -28,7 +28,14 @@ func MatchesFromFilterRules(pf []tailcfg.FilterRule) ([]Match, error) {
 	var erracc error
 
 	for _, r := range pf {
-		m := Match{}
+		// Profiling determined that this function was spending a lot
+		// of time in runtime.growslice. As such, we attempt to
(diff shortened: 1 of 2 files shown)


danderson (approved):

wgengine/filter/tailcfg.go
```diff
@@ -28,7 +28,14 @@ func MatchesFromFilterRules(pf []tailcfg.FilterRule) ([]Match, error) {
 	var erracc error
 
 	for _, r := range pf {
-		m := Match{}
+		// Profiling determined that this function was spending a lot
+		// of time in runtime.growslice. As such, we attempt to
+		// pre-allocate some slices. Multipliers were chosen arbitrarily.
```
bradfitz, line 33: arbitrarily means it could actually be worse, though?

twitchyliquid64: growslice is a different codepath to preallocation i think? so we are amortizing multiple allocation & copies to one or at least less.

PR #4673, opened 2022-05-13
tsweb: memoize the string forms of HTTP response codes.

danderson:
Saves 1-2 allocs per HTTP request after warmup, thanks to
avoiding strconv and fmt.

### DIFF
--- a/tsweb/tsweb.go
+++ b/tsweb/tsweb.go
@@ -23,6 +23,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"go4.org/mem"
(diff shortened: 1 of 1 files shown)


bradfitz (commented):
factor it out into a func?

danderson:
Done, PTAL?

tsweb/tsweb.go
```diff
@@ -302,15 +303,38 @@ func (h retHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if h.opts.StatusCodeCounters != nil {
-		key := fmt.Sprintf("%dxx", msg.Code/100)
-		h.opts.StatusCodeCounters.Add(key, 1)
+		h.opts.StatusCodeCounters.Add(responseCodeString(msg.Code/100), 1)
```
bradfitz, line 306: /100*100, no?

bradfitz: oh, nevermind, I see :)

bradfitz (approved):

PR #4675, opened 2022-05-14
ipn/ipnlocal: add missing place where we set the SSH atomic

bradfitz:
This fixes the "tailscale up --authkey=... --ssh" path which wasn't setting the bit.

Updates #3802


bradfitz:
Untested so far. I just read the code starting cli/up.go to LocalBackend.Start and looks like this is the path.

maisem (approved):

bradfitz:
Tested. Clarified commit message.

PR #4683, opened 2022-05-15
types/views: add SliceIterator to make iterating over Slices nicer

maisem:

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -3128,8 +3128,8 @@ func peerAPIBase(nm *netmap.NetworkMap, peer *tailcfg.Node) string {
 	}
 	var p4, p6 uint16
 	svcs := peer.Hostinfo.Services()
-	for i, n := 0, svcs.Len(); i < n; i++ {
-		s := svcs.At(i)
+	for i := svcs.Iterator(); i.Valid(); i.Next() {
+		s := i.Val()
 		switch s.Proto {
 		case tailcfg.PeerAPI4:
 			p4 = s.Port
@@ -3256,9 +3256,9 @@ func exitNodeCanProxyDNS(nm *netmap.NetworkMap, exitNodeID tailcfg.StableNodeID)
 		if p.StableID != exitNodeID {
 			continue
 		}
-		services := p.Hostinfo.Services()
-		for i, n := 0, services.Len(); i < n; i++ {
-			s := services.At(i)
+		svcs := p.Hostinfo.Services()
+		for i := svcs.Iterator(); i.Valid(); i.Next() {
+			s := i.Val()
 			if s.Proto == tailcfg.PeerAPIDNS && s.Port >= 1 {
 				return peerAPIBase(nm, p) + "/dns-query", true
 			}
--- a/types/views/views.go
+++ b/types/views/views.go
@@ -88,6 +88,36 @@ func (v SliceView[T, V]) AsSlice() []V {
 	return v.AppendTo(nil)
 }
 
+func (v SliceView[T, V]) Iterator() SliceViewIterator[T, V] {
+	return SliceViewIterator[T, V]{}
+}
+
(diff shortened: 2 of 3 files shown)


bradfitz (commented):
I want to see benchmarks including allocs for this, views with At, and regular slices without generics.

maisem:
> I want to see benchmarks including allocs for this, views with At, and regular slices without generics.
```
goos: darwin
goarch: arm64
pkg: tailscale.com/types/views
BenchmarkSliceIterator-10           	 3689444	       319.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkSliceManualIteration-10    	 3745927	       319.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkSlice-10                   	 3741626	       319.6 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	tailscale.com/types/views	4.813s
```

ipn/ipnlocal/local.go
```diff
@@ -3128,8 +3128,8 @@ func peerAPIBase(nm *netmap.NetworkMap, peer *tailcfg.Node) string {
 	}
 	var p4, p6 uint16
 	svcs := peer.Hostinfo.Services()
-	for i, n := 0, svcs.Len(); i < n; i++ {
-		s := svcs.At(i)
+	for i := svcs.Iterator(); i.Valid(); i.Next() {
```
bradfitz, line 3131: How about just `Iter()`. And I guess `Valid` works, but something even shorter like `Ok` is tempting.

And it'd be nice of the whole thing were shorter somehow. Like renaming Valid to Next and making Next advance, so it operates like https://pkg.go.dev/database/sql#Rows.Next and dropping the final PostStmt.

types/views/views.go
```diff
@@ -88,6 +88,36 @@ func (v SliceView[T, V]) AsSlice() []V {
...
+	return si
+}
+
+// Valid reports whether the iterator points to a valid value.
+func (si SliceViewIterator[T, V]) Valid() bool {
+	return si.i < len(si.s.ж) // it's cheaper to use the underlying value here.
+}
+
+// Val returns the value at the iterator position.
+func (si SliceViewIterator[T, V]) Val() V { return si.s.At(si.i) }
```
dsnet, line 119: Optional: personally I like `Value` over `Val`. What's 2 characters?

types/views/views.go
```diff
@@ -88,6 +88,36 @@ func (v SliceView[T, V]) AsSlice() []V {
...
+//   for i := s.Iterator(); i.Valid(); i = i.Next() {
+//     v := i.Val()
+//   }
+type SliceViewIterator[T ViewCloner[T, V], V StructView[T]] struct {
+	s SliceView[T, V]
+	i int
+}
+
+// Next increments the iterator.
+func (si SliceViewIterator[T, V]) Next() SliceViewIterator[T, V] {
```
dsnet, line 108: I'm confused. This modifies a local copy and returns it.

But above in the `local.go` example, `i.Next()` is called without assigning the output to anything.

Should this be declared on a pointer receiver instead?
```go
func (si *SliceViewIterator[T, V]) Next()
```

Given that this is a concrete type, I suspect that the compiler will prove that usages do not escape and we don't have any allocations.

types/views/views.go
```diff
@@ -88,6 +88,36 @@ func (v SliceView[T, V]) AsSlice() []V {
...
+}
+
+// SliceViewIterator provides an iterator interface over a SliceView.
+//
+// A typical usage example is
+//   var s Slice[T]
+//   for i := s.Iterator(); i.Valid(); i = i.Next() {
+//     v := i.Val()
+//   }
+type SliceViewIterator[T ViewCloner[T, V], V StructView[T]] struct {
```
dsnet, line 102: It is a bit weird that the iterator patterns for slice here and in `Map.Range` are drastically different. I don't know how to unify the two in a way that performs well since you can't easily capture the iteration state of a map in a variable.

PR #4690, opened 2022-05-16
net/dns: timeout dns queries after 10 seconds

twitchyliquid64:
Fixes https://github.com/tailscale/corp/issues/5198

The upstream forwarder will block indefinitely on `udpconn.ReadFrom` if no
reply is recieved, due to the lack of deadline on the connection object.

There still isn't a deadline on the connection object, but the automatic closing
of the context on deadline expiry will close the connection via `closeOnCtxDone`,
unblocking the read and resulting in a normal teardown.


bradfitz:
I look forward to reviewing this and auditing all the contexts, timeouts, and channels in this code, but I won't have time until tonight.

net/dns/manager.go
```diff
@@ -339,6 +341,9 @@ func (m *Manager) Query(ctx context.Context, bs []byte, from netaddr.IPPort) ([]
 		// continue
 	}
 
+	ctx, cancel := context.WithTimeout(ctx, dnsQueryTimeout)
```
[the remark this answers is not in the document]
maisem: `10s` is a long time for a DNS query, no? Should this be lower?

maisem (approved):

maisem (commented):
oh I didn't see brad's comment, wait for his review pls :)

twitchyliquid64:
WILCO.

net/dns/manager.go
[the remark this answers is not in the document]
twitchyliquid64: My intention was that this timeout just puts an upper bound on resource usage - I expect the downstream (iOS in your case) DNS client will retry far more aggressively than this.

net/dns/manager.go
```diff
@@ -331,6 +331,8 @@ func (m *Manager) NextPacket() ([]byte, error) {
 	return buf, nil
 }
 
+const dnsQueryTimeout = 10 * time.Second
+
 func (m *Manager) Query(ctx context.Context, bs []byte, from netaddr.IPPort) ([]byte, error) {
```
bradfitz, line 336: // Query executes a DNS query (encoded in bs) from the provided IPPort.
// It's called for requests arriving over TCP and UDP.
// The bs query is in _____ form (is it always UDP form, TCP form, varies depending on transport type?).

net/dns/manager.go
```diff
@@ -331,6 +331,8 @@ func (m *Manager) NextPacket() ([]byte, error) {
 	return buf, nil
 }
```
bradfitz, line 332: commit message:

`net/dns: time out DNS ...`

(timeout = noun, time out = verb, and DNS all caps in English)

net/dns/manager.go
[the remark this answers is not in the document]
bradfitz: Add some commentary about this (upper bound, etc) on the constant above.

bradfitz, line 344: This can be moved to after the maxActiveQueries right before the `m.resolver.Query` call.

But why isn't it in `(*Resolver).Query`, actually?

Or even inside `(*Resolver).Query` it only needs to be done in the `errNotOurName` forwarding path, otherwise we'll answer right away (or have an answer right away), meaning you can avoid the context creation work.

In fact, we're already creating a new context in `(*forwarder).forwardWithDestChan`:

```go
	// Guarantee that the ctx we use below is done when this function returns.
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()
```

That `WithCancel` could be a `WithTimeout` and not increase the total number of context wrappers created.

But lawlz... `(*Resolver).Query` also has a `ctx, cancel := context.WithCancel(ctx)` with a deferred cancel.

Everybody loves that pattern so every layer does it I guess.

In retrospect I think the middle layer might be best: as you want to nuke `forwarderWithDestChan` anyway.

So move it to the immediate caller of forwarding, the `ctx, cancel := context.WithCancel(ctx)` in `(*Resolver).Query`, changing that to a `WithTimeout`?

net/dns/manager.go
```diff
@@ -460,7 +465,7 @@ func (m *Manager) HandleTCPConn(conn net.Conn, srcAddr netaddr.IPPort) {
 		responses:    make(chan []byte),
 		readClosing:  make(chan struct{}),
 	}
-	s.ctx, s.closeCtx = context.WithCancel(context.Background())
+	s.ctx, s.closeCtx = context.WithCancel(m.ctx)
```
bradfitz, line 468: LGTM but this is unrelated, right?

bradfitz (approved):

net/dns/manager.go
twitchyliquid64: Yepp

net/dns/manager.go
twitchyliquid64: Done

net/dns/manager.go
twitchyliquid64: Done

PR #4700, opened 2022-05-17
logtail: add instance metadata to the entry logtail

mihaip:
Allows instances that are running with the same machine ID (due to
cloning) to be distinguished.

For tailscale/corp#5244

### DIFF
--- a/logtail/logtail.go
+++ b/logtail/logtail.go
@@ -8,6 +8,8 @@ package logtail
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
+	"encoding/binary"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -15,6 +17,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -62,6 +65,17 @@ type Config struct {
 	// DrainLogs, if non-nil, disables automatic uploading of new logs,
 	// so that logs are only uploaded when a token is sent to DrainLogs.
 	DrainLogs <-chan struct{}
+
+	// IncludeProcID, if true, results in an ephemeral process identifier being
+	// included in logs. The ID is random and not guaranteed to be globally
+	// unique, but it can be used to distinguish between different instances
+	// running with same PrivateID.
+	IncludeProcID bool
+
+	// IncludeProcSequence, if true, results in an ephemeral sequence number
+	// being included in the logs. The sequence number is incremented for each
+	// log message sent, but is not peristed across process restarts.
+	IncludeProcSequence bool
 }
 
 func NewLogger(cfg Config, logf tslogger.Logf) *Logger {
@@ -84,6 +98,17 @@ func NewLogger(cfg Config, logf tslogger.Logf) *Logger {
 		}
 		cfg.Buffer = NewMemoryBuffer(pendingSize)
 	}
+	var procID uint32
+	if cfg.IncludeProcID {
+		keyBytes := make([]byte, 4)
+		rand.Read(keyBytes)
+		procID = binary.LittleEndian.Uint32(keyBytes)
+		if procID == 0 {
+			// 0 is the empty/off value, assign a different (non-zero) value to
+			// make sure we still include an ID (actual value does not matter).
+			procID = 7
+		}
+	}
 	l := &Logger{
 		privateID:      cfg.PrivateID,
 		stderr:         cfg.Stderr,
@@ -99,6 +124,9 @@ func NewLogger(cfg Config, logf tslogger.Logf) *Logger {
 		timeNow:        cfg.TimeNow,
 		bo:             backoff.NewBackoff("logtail", logf, 30*time.Second),
 		metricsDelta:   cfg.MetricsDelta,
+
+		procID:              procID,
+		includeProcSequence: cfg.IncludeProcSequence,
 
 		shutdownStart: make(chan struct{}),
 		shutdownDone:  make(chan struct{}),
@@ -136,6 +164,11 @@ type Logger struct {
 	explainedRaw   bool
 	metricsDelta   func() string // or nil
 	privateID      PrivateID
+
+	procID              uint32
+	includeProcSequence bool
+	writeLock           sync.Mutex // guards increments of procSequence
+	procSequence        uint64
 
 	shutdownStart chan struct{} // closed when shutdown begins
 	shutdownDone  chan struct{} // closed when shutdown complete
@@ -253,8 +286,6 @@ func (l *Logger) drainPending(scratch []byte) (res []byte) {
 		if b[0] != '{' || !json.Valid(b) {
 			// This is probably a log added to stderr by filch
 			// outside of the logtail logger. Encode it.
-			// Do not add a client time, as it could have been
-			// been written a long time ago.
 			if !l.explainedRaw {
 				fmt.Fprintf(l.stderr, "RAW-STDERR: ***\n")
 				fmt.Fprintf(l.stderr, "RAW-STDERR: *** Lines prefixed with RAW-STDERR below bypassed logtail and probably come from a previous run of the program\n")
@@ -263,7 +294,10 @@ func (l *Logger) drainPending(scratch []byte) (res []byte) {
 				l.explainedRaw = true
 			}
 			fmt.Fprintf(l.stderr, "RAW-STDERR: %s", b)
-			b = l.encodeText(b, true, 0)
+			// Do not add a client time, as it could have been
+			// been written a long time ago. Don't include instance key or ID
+			// either, since this came from a different instance.
+			b = l.encodeText(b, true, 0, 0, 0)
 		}
 
 		if entries > 0 {
@@ -437,14 +471,24 @@ func (l *Logger) send(jsonBlob []byte) (int, error) {
 
 // TODO: instead of allocating, this should probably just append
 // directly into the output log buffer.
-func (l *Logger) encodeText(buf []byte, skipClientTime bool, level int) []byte {
+func (l *Logger) encodeText(buf []byte, skipClientTime bool, procID uint32, procSequence uint64, level int) []byte {
 	now := l.timeNow()
 
 	// Factor in JSON encoding overhead to try to only do one alloc
 	// in the make below (so appends don't resize the buffer).
-	overhead := 13
+	overhead := len(`{"text": ""}\n`)
+	includeLogtail := !skipClientTime || procID != 0 || procSequence != 0
+	if includeLogtail {
+		overhead += len(`"logtail": {},`)
+	}
 	if !skipClientTime {
-		overhead += 67
+		overhead += len(`"client_time": "2006-01-02T15:04:05.999999999Z07:00",`)
+	}
+	if procID != 0 {
+		overhead += len(`"proc_id": 4294967296,`)
+	}
+	if procSequence != 0 {
+		overhead += len(`"proc_seq": 9007199254740992,`)
 	}
 	// TODO: do a pass over buf and count how many backslashes will be needed?
 	// For now just factor in a dozen.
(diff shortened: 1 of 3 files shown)


mihaip:
Let me know if this is on the right track. 

Have not added the node ID yet, partly because I wasn't sure how to get it (pass in a function at https://github.com/tailscale/tailscale/blob/58b2f181695ad6a6321b102dd55cc170dcf317ff/cmd/tailscaled/tailscaled.go#L286 that uses the `LocalBackend` from the ipnserver?), partly because it seemed like a layering violation (and not something that we strictly need).

logtail/logtail.go
```diff
@@ -137,6 +160,10 @@ type Logger struct {
 	metricsDelta   func() string // or nil
 	privateID      PrivateID
 
+	instanceKey             *string
```
bradfitz, line 163: pointer-to-string to mean optional isn't super common and generally requires docs.  Better to just use the string zero value for unset (but still with little same-line doc like `// or empty if off`)

logtail/logtail.go
```diff
@@ -437,14 +465,24 @@ func (l *Logger) send(jsonBlob []byte) (int, error) {
 
 // TODO: instead of allocating, this should probably just append
 // directly into the output log buffer.
-func (l *Logger) encodeText(buf []byte, skipClientTime bool, level int) []byte {
+func (l *Logger) encodeText(buf []byte, skipClientTime bool, instanceKey *string, instanceSequence int64, level int) []byte {
 	now := l.timeNow()
 
 	// Factor in JSON encoding overhead to try to only do one alloc
 	// in the make below (so appends don't resize the buffer).
 	overhead := 13
```
dsnet, line 473: Instead of magical length constants, use a `len`:
```go
overhead := len(`"logtail": `)
```
and so forth.

In fact, I'm not sure where 13 came from, so my suggestion might not be using the right string literal.

logtail/logtail.go
```diff
@@ -84,6 +97,13 @@ func NewLogger(cfg Config, logf tslogger.Logf) *Logger {
 		}
 		cfg.Buffer = NewMemoryBuffer(pendingSize)
 	}
+	var instanceKey *string
+	if cfg.IncludeInstanceKey {
+		keyBytes := make([]byte, 4)
```
bradfitz, line 102: I'd do a bit bigger. this isn't much random.

logtail/logtail.go
```diff
@@ -437,14 +465,24 @@ func (l *Logger) send(jsonBlob []byte) (int, error) {
...
 	// Factor in JSON encoding overhead to try to only do one alloc
 	// in the make below (so appends don't resize the buffer).
 	overhead := 13
+	includeLogtail := !skipClientTime || instanceKey != nil || instanceSequence != 0
+	if includeLogtail {
+		overhead += 20
+	}
 	if !skipClientTime {
-		overhead += 67
+		overhead += 52
```
dsnet, line 479: For example:
```go
overhead += len(`"client_time": "0001-01-01T00:00:00.000000000Z", `)
```

logtail/logtail.go
```diff
@@ -468,10 +506,34 @@ func (l *Logger) encodeText(buf []byte, skipClientTime bool, level int) []byte {
...
+			b = append(b, `"client_time": "`...)
+			b = now.AppendFormat(b, time.RFC3339Nano)
+			b = append(b, '"')
+			needsComma = true
+		}
+		if instanceKey != nil {
+			if needsComma {
+				b = append(b, ',')
+			}
+			b = append(b, `"instance_key": "`...)
```
bradfitz, line 522: proc_key, proc_seq ?

"instance" already has a meaning and I'd prefer to not overload that.

logtail/logtail.go
dsnet: This should either be large enough that random collisions never happens (in which cases this would need to be size of a MD5 or SHA256 hash), or short enough where we know it can occasionally happen with sufficiently low probability and anything processing doesn't choke on duplicates.

IIUC, our use-case is the latter, where some collisions is okay. I'm a vote for a 32-bit value.

logtail/logtail.go
```diff
@@ -137,6 +160,10 @@ type Logger struct {
 	metricsDelta   func() string // or nil
 	privateID      PrivateID
 
+	instanceKey             *string
+	includeInstanceSequence bool
+	instanceSequence        int64
```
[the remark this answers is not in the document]
dsnet: Thinking about it more, I'm okay with this being an `uint64`. It will take a very long time (i.e., 2^53 messages) for it to be indistinguishable as a `float64`. Also, when the process first starts, the number will be small. As it grows bigger, the leading prefix will have the same bytes, and thus compress well.

logtail/logtail.go
bradfitz: Sure. I guess you'd need 50 million nodes cloned off a log id to have a 1% risk of a collision.

32's fine.

I'm just used to disco keys (which made me think of this whole thing) which are 32 bytes, for different reasons.

logtail/logtail.go
mihaip: It's a unit32 now, switched to 0 being the off value

logtail/logtail.go
mihaip: Stayed with 32 bit, and stored it as a uint32 to avoid the string overhead.

logtail/logtail.go
mihaip: Ended up with `proc_id` and `proc_seq`

logtail/logtail.go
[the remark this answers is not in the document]
mihaip: Switched to `uint64`

logtail/logtail.go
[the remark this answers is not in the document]
mihaip: Yep, went with `procID uint32`

logtail/logtail.go
```diff
@@ -468,10 +506,34 @@ func (l *Logger) encodeText(buf []byte, skipClientTime bool, level int) []byte {
 	b := make([]byte, 0, len(buf)+overhead)
 	b = append(b, '{')
 
-	if !skipClientTime {
-		b = append(b, `"logtail": {"client_time": "`...)
-		b = now.AppendFormat(b, time.RFC3339Nano)
-		b = append(b, "\"}, "...)
+	if includeLogtail {
+		b = append(b, `"logtail": {`...)
+		needsComma := false
```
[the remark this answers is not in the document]
mihaip: Did this, except I did a subslice more directly -- `TrimRight` docs say that it looks at UTF-8 encoded code points, which seems like it would have more overhead.

logtail/logtail.go
mihaip: I like this, switched to it (didn't realize the compiler would be smart enough to generate a constant, but it does -- https://go.godbolt.org/z/hPGoTbhrW).

I believe the 13 is due to `{"text": ""}\n`

logtail/logtail.go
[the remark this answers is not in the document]
dsnet: No need to worry about that. `TrimRight` has been optimized to handle a single-byte cutset pretty efficiently 😄: https://go-review.googlesource.com/c/go/+/323318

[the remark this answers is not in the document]
dsnet: Also, I think using `TrimRight` is more correct in the rare case where none of the `logtail` fields are outputed, in which case you will accidentally truncate something that is not a `,`.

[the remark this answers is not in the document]
mihaip: Nice, TIL

logtail/logtail.go
```diff
@@ -62,6 +65,17 @@ type Config struct {
 	// DrainLogs, if non-nil, disables automatic uploading of new logs,
 	// so that logs are only uploaded when a token is sent to DrainLogs.
 	DrainLogs <-chan struct{}
+
+	// IncludeProcId, if true, results in an ephemeral process identifier being
+	// included in logs. The ID is random and not guaranteed to be globally
+	// unique, but it can be used to distinguish between different instances
+	// running with same PrivateID.
+	IncludeProcId bool
```
dsnet, line 73: Nit: In Go, usually ID is written as "ID" (e.g., `dbx.TailnetID`) instead of "Id".

logtail/logtail.go
```diff
@@ -137,6 +160,13 @@ type Logger struct {
 	metricsDelta   func() string // or nil
 	privateID      PrivateID
 
+	procID              uint32
+	includeProcSequence bool
+	// procSequence is increment atomically, pad32.Four is needed to ensure that
```
dsnet, line 165: s/increment/incremented/

logtail/logtail.go
```diff
@@ -84,6 +98,12 @@ func NewLogger(cfg Config, logf tslogger.Logf) *Logger {
 		}
 		cfg.Buffer = NewMemoryBuffer(pendingSize)
 	}
+	var procID uint32
+	if cfg.IncludeProcId {
+		keyBytes := make([]byte, 4)
+		rand.Read(keyBytes)
+		procID = binary.LittleEndian.Uint32(keyBytes)
```
dsnet, line 105: In the 1 in 4 billion case where this is 0, we should add 1 since we specially treat 0 elsewhere as not being specified.

logtail/logtail.go
```diff
@@ -437,14 +468,24 @@ func (l *Logger) send(jsonBlob []byte) (int, error) {
...
+func (l *Logger) encodeText(buf []byte, skipClientTime bool, procID uint32, procSequence uint64, level int) []byte {
 	now := l.timeNow()
 
 	// Factor in JSON encoding overhead to try to only do one alloc
 	// in the make below (so appends don't resize the buffer).
-	overhead := 13
+	overhead := len(`{"text": ""}\n`)
+	includeLogtail := !skipClientTime || procID != 0 || procSequence != 0
+	if includeLogtail {
+		overhead += len(`"logtail": {,,},`)
```
dsnet, line 479: I think we can drop the commas, since we already account for those elsewhere:
```go
overhead += len(`"logtail": {},`)
```

logtail/logtail.go
```diff
@@ -468,10 +509,26 @@ func (l *Logger) encodeText(buf []byte, skipClientTime bool, level int) []byte {
...
+			b = append(b, `"proc_id": `...)
+			b = strconv.AppendUint(b, uint64(procID), 10)
+			b = append(b, ',')
+		}
+		if procSequence != 0 {
+			b = append(b, `"proc_seq": `...)
+			b = strconv.AppendUint(b, procSequence, 10)
+			b = append(b, ',')
+		}
+		// Drop trailing comma
```
dsnet, line 529: Nit: Comment seems self-explanatory from the code itself.

dsnet (approved):

dsnet:
I think there's a race in this code.

In order for the `proc_seq` to be useful, we need it to be in monotonically increasing order.

Suppose, we have two concurrent calls to `Logger.Write` racing with each other. Consider the following sequence of events:
* Call#1: this calls `l.encode` which then executes `atomic.AddUint64(&l.procSequence, 1)` which returns 1
* Call#1 sleeps and call#2 runs.
* Call#2: this calls `l.encode` which then executes `atomic.AddUint64(&l.procSequence, 1)` which returns 2
* Call#2: this calls `l.send`, which records a message marked as sequence 2
* Call#2 sleeps and call#1 runs.
* Call#1: this calls `l.send`, which records a message marked as sequence 1

Now they are recorded out-of-order. Rather than `procSequence` being an atomic integer, I think we need a lock over this critical section:
```go
b := l.encode(buf, level)
_, err := l.send(b)
```

mihaip:
Good point, added the locking. I'm assuming we're not concerned with `Write` blocking for longer than before?

dsnet:
> Good point, added the locking. I'm assuming we're not concerned with Write blocking for longer than before?

Hard to say. If this becomes a problem, I see some low-hanging fruit to optimize the critical section to be faster.

dsnet (approved):

logtail/logtail.go
```diff
@@ -84,6 +98,15 @@ func NewLogger(cfg Config, logf tslogger.Logf) *Logger {
 		}
 		cfg.Buffer = NewMemoryBuffer(pendingSize)
 	}
+	var procID uint32
+	if cfg.IncludeProcID {
+		keyBytes := make([]byte, 4)
+		rand.Read(keyBytes)
+		procID = binary.LittleEndian.Uint32(keyBytes)
+		if procID == 0 {
+			procID = 7
```
bradfitz, line 107: what's 7? this deserves a comment.

bradfitz (approved):

logtail/logtail.go
```diff
@@ -136,6 +162,11 @@ type Logger struct {
 	explainedRaw   bool
 	metricsDelta   func() string // or nil
 	privateID      PrivateID
+	writeLock      sync.Mutex
```
bradfitz, line 165: say what this guards.

oh, you wrote it below.

the convention is immutable stuff at the top and then blocks of:

```go

   mu sync.Mutex // guards following
   foo T
  
```

Issue #4686, opened 2022-05-16
Five second timeout for DNS lookup in 1.25.77


DentonGentry:
### What is the issue?

"ping example.com" freezes for 5 seconds before starting to ping in 1.25.77, which was not seen with 1.25.40. Notably, I believe 1.25.77 is the first unstable build to move MagicDNS to use netstack https://github.com/tailscale/tailscale/pull/4408

strace -tt ping apple.com shows:
```
03:07:31.543860 setsockopt(5, SOL_IP, IP_RECVERR, [1], 4) = 0
03:07:31.543953 connect(5, {sa_family=AF_INET, sin_port=htons(53), sin_addr=inet_addr("100.100.100.100")}, 16) = 0
03:07:31.544076 poll([{fd=5, events=POLLOUT}], 1, 0) = 1 ([{fd=5, revents=POLLOUT}])
03:07:31.544182 sendmmsg(5, [{msg_hdr={msg_name=NULL, msg_namelen=0, msg_iov=[{iov_base="\354\325\1\0\0\1\0\0\0\0\0\0\5apple\3com\0\0\1\0\1", iov_len=27}], msg_iovlen=1, msg_controllen=0, msg_flags=0}, msg_len=27}, {msg_hdr={msg_name=NULL, msg_namelen=0, msg_iov=[{iov_base="Y\332\1\0\0\1\0\0\0\0\0\0\5apple\3com\0\0\34\0\1", iov_len=27}], msg_iovlen=1, msg_controllen=0, msg_flags=0}, msg_len=27}], 2, MSG_NOSIGNAL) = 2
03:07:31.544363 poll([{fd=5, events=POLLIN}], 1, 5000) = 1 ([{fd=5, revents=POLLIN}])
03:07:31.577430 ioctl(5, FIONREAD, [43]) = 0
03:07:31.577562 recvfrom(5, "\354\325\201\200\0\1\0\1\0\0\0\0\5apple\3com\0\0\1\0\1\300\f\0\1\0"..., 2048, 0, {sa_family=AF_INET, sin_port=htons(53), sin_addr=inet_addr("100.100.100.100")}, [28->16]) = 43
03:07:31.577748 poll([{fd=5, events=POLLIN}], 1, 4966

...this is where the ping paused for a few seconds. I hit return a few times...

) = 0 (Timeout)
03:07:36.545435 poll([{fd=5, events=POLLOUT}], 1, 0) = 1 ([{fd=5, revents=POLLOUT}])
[...]

PR #4708, opened 2022-05-19
wgengine/netstack: handle multiple magicDNS queries per UDP socket

twitchyliquid64:
Fixes: #4686 
See: https://github.com/tailscale/tailscale/issues/4686#issuecomment-1132219403


bradfitz (commented):
It seems like using netstack for this was a huge regression, no?

Before we just plucked these packets out as they went by and didn't have to worry about deadlines.

150ms is both too long and too short.

twitchyliquid64:
We could in theory switch back to the old path for UDP without too much hassle: just have `handleLocalPackets` return `filter.Accept` for magicDNS UDP and it will use the old path.

I'm not happy about the extra overhead thanks to the deadline (looking through the gonet adapter, it boils down to a timer and a channel), but I don't think its the end of the world given how cold this path is.

> 150ms is both too long and too short.

Come again? :joy:

bradfitz:
> > 150ms is both too long and too short.
>
> Come again? 😂

It's too short for the first DNS query over UDP and too long for the last one that never arrives.

bradfitz:
What does gonet/netstack actually get us, though?

twitchyliquid64:
> It's too short for the first DNS query over UDP and too long for the last one that never arrives.

Its fine for the first one, as the 'connection' (i.e. the `PacketConn`) doesnt actually exist till the first UDP packet comes in and gets queued to be read. Agree on it being too long though.

> What does gonet/netstack actually get us, though?

Mostly, a consistent abstraction layer (yay sockets) in an otherwise chaotic world. The bit where packets get flung around like 6 different paths within tailscaled and get mangled all over the place took me ages to get my head around.

With UDP being entirely stateless, theres no technical need for us to use netstack. I'm fine with going back to using `NextPacket()` / `EnqueuePacket()` for DNS UDP and skipping netstack, but the tradeoff there is now we have more places/paths for people to understand, rather than converging on a shared abstraction (netstack) for all our simulated traffic.

DentonGentry:
Even for UDP I'd think there is an advantage to netstack. I don't see any IP fragmentation handling in `NextPacket()` / `EnqueuePacket()`, which very large responses might conceivably use. People with the power to do so will tailor their domain's DNS responses to fit within one packet, but then we might fragment it anyway for the 1280 byte MTU.

bradfitz:
Oh, I hadn't even considered IP fragmentation at all.

How's that work with netstack? This PR's just doing a `c.Read` and not a `PacketConn.ReadFrom` or whatever. How do you know boundaries between multiple DNS packets and how they're different than boundaries between fragments?

twitchyliquid64:
Defrag should happen up the stack, so we should only see each assembled packet.

I've always observed that each call to `read()` on a UDP socket returns no more than one packet (ie: not stream oriented like TCP). Now that I actually look at the manpage, it seems that semantic is only guaranteed with `recv()` / `recvfrom()`, so I'll switch to `ReadFrom`.

bradfitz (approved):

PR #4709, opened 2022-05-19
ipn/ipnserver: fix build on js/wasm

bradfitz:
Broken by 3dedcd1640 but we don't have CI coverage yet.


mihaip (approved):
Mention that this is for #3157?

PR #4711, opened 2022-05-20
cmd/tailscale/cli: fix ssh CLI command breaking the Wasm build

mihaip:
Adds a stub for syscall.Exec when GOOS=js. We also had a separate branch
for Windows, might as well use the same mechanism there too.

For #3157


bradfitz (approved):
execSSH is the Go case

cmd/tailscale/cli/ssh_exec_windows.go
```diff
@@ -0,0 +1,25 @@
...
+
+package cli
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+)
+
+func execSsh(ssh string, argv []string) error {
```
maisem, line 13: `execSSH`

maisem (approved):

PR #4720, opened 2022-05-20
control/controlclient: avoid Noise protocol for js/wasm for now

bradfitz:
Updates #3157

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -1097,6 +1097,12 @@ func loadServerPubKeys(ctx context.Context, httpc *http.Client, serverURL string
 	var out tailcfg.OverTLSPublicKeyResponse
 	jsonErr := json.Unmarshal(b, &out)
 	if jsonErr == nil {
+		if runtime.GOOS == "js" {
+			// As of 2022-05-20 it's not possible for js/wasm to make a bidi
+			// Noise connection to the control plane. Instead, for now, pretend
+			// like the server can't do Noise to force use of the old protocol.
+			out.PublicKey = key.MachinePublic{}
+		}
 		return &out, nil
 	}
 


control/controlclient/direct.go
```diff
@@ -1097,6 +1097,12 @@ func loadServerPubKeys(ctx context.Context, httpc *http.Client, serverURL string
 	var out tailcfg.OverTLSPublicKeyResponse
 	jsonErr := json.Unmarshal(b, &out)
 	if jsonErr == nil {
+		if runtime.GOOS == "js" {
+			// As of 2022-05-20 it's not possible for js/wasm to make a bidi
+			// Noise connection to the control plane. Instead, for now, pretend
+			// like the server can't do Noise to force use of the old protocol.
+			out.PublicKey = key.MachinePublic{}
```
mihaip, line 1104: https://github.com/tailscale/tailscale/blob/880a6713e7d0df2d20c3b663781a61a1e2b191eb/tailcfg/tailcfg.go#L1655-L1661 says that the legacy key will "will be zero for sufficiently new clients". Based on reading the control code (https://github.com/tailscale/corp/blob/bebeb7810864059946c57bcb87689db99465485c/control/control.go#L6026-L6040) it seems like we populate it unconditionally, but this may break at some point.

mihaip (approved):

control/controlclient/direct.go
bradfitz: Yeah. I'm assuming we'll fix the js/wasm client to be able to do noise by then.

The tricky bit is that browsers have historically sucked at doing bidi HTTP traffic other than websockets and our trunkd servers don't yet do noise-over-websockets. We figured we'd need to eventually, though.

PR #4724, opened 2022-05-20
types/views: add package level Range

bmizerany:
While using `cmd/viewer` and the `views` package (which are super nice!), I found that when errors can happen while ranging over maps, preventing err shadowing gets tricky, and the the desire to "just return an error" creeps up, so I wrote a little helper and thought it might be nice if it could have a home here since it's pretty generic.

An example of easy mistake before Range:

```
mv := views.MapOf(map[int]bool{...})

var err error
mv.Range(func(k, v string) bool {
       _, err = db.Exec(...)
       if err != nil {
           // maybe sets error to non-nil
           return false
       }
       _, err := db.Exec(...)
       if err != nil {
          // maybe sets shadowed err to non-nil, but outer err is still nil.
          return false
       }
       return true
})
if err != nil {
   // ... if the first db.Exec returns a nil error; but the second does not, this err is still nil
}
```

After:

```
err := views.Range(mv, func(k, v string) error {
      _, err := db.Exec(...)
     if err != nil {
         return err
     }
     _, err = db.Exec(...)
     return err
})
```

One could also argue it aids in clarity.

I hope this is helpful and makes it in!

Also: I'm sure my docs here could be better. Feedback there would very much be appreciated.

### DIFF
--- a/types/views/views.go
+++ b/types/views/views.go
@@ -298,3 +298,21 @@ func (m MapFn[K, T, V]) Range(f MapRangeFn[K, V]) {
 		}
 	}
 }
+
+type Ranger[K comparable, V any] interface {
+	Range(f MapRangeFn[K, V])
+}
+
+// Range is like Map.Range but it takes a func that breaks the range if it
+// returns an error.  It returns the error returned by f, if any.
+//
+// It is used to where errors may occur while ranging over maps and shadowing
+// errors gets tricky without it.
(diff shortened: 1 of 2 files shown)


DentonGentry:
Thanks!

To set expectations: the person who did most of the work on Views is unavailable next week. Please have patience, it will likely be more than a week before someone has review comments.

bmizerany:
No rush!

Issue #4738, opened 2022-05-26
FR: Implement LoginInteractive via localapi


jtcressy:
### What are you trying to do?

To do just the opposite of /localapi/v0/logout, I would like to instead kick-off interactive login; basically call the function StartLoginInteractive() from LocalBackend in ipn/ipnlocal/local.go#L1686 except via the localapi

This way I can start the interactive login process (basically populate an AuthURL in backend status) from an external tool without wrapping a shell around the tailscale binary. The tailscale binary is capable of it by sending requests directly through some kind of socket which I don't yet understand. It skips the localapi http interface somehow and somewhat reminds me of gRPC instead.

### How should we solve this?

Just like the logout handler here https://github.com/tailscale/tailscale/blob/fc5839864bfbd5bfed512cbb6f05e32894e93938/ipn/localapi/localapi.go#L346
we should create a new handler called `serveInteractiveLogin` or similar that instead of calling `LocalBackend`'s `LogoutSync()` function:
https://github.com/tailscale/tailscale/blob/fc5839864bfbd5bfed512cbb6f05e32894e93938/ipn/localapi/localapi.go#L355
^ called there, defined here:
https://github.com/tailscale/tailscale/blob/fc5839864bfbd5bfed512cbb6f05e32894e93938/ipn/ipnlocal/local.go#L2830

We could simply call `StartLoginInteractive()` as defined here:
https://github.com/tailscale/tailscale/blob/fc5839864bfbd5bfed512cbb6f05e32894e93938/ipn/ipnlocal/local.go#L1686

### What is the impact of not solving this?

The only way to kick off interactive login is to run `tailscale up` with no arguments (or with `--reset`) from a shell and it would be much cleaner to use the local client (`import tailscale.com/client/tailscale`) from another golang program to execute commands to the local API

DentonGentry:
The currently supported way to do this is `tailscale up --json`
I understand that the bug states this isn't a desired solution, nonetheless it is available now.

jtcressy:
> The currently supported way to do this is `tailscale up --json` I understand that the bug states this isn't a desired solution, nonetheless it is available now.

I did try using this, but one annoyance is this command still hangs (blocks) until authentication completes. I worked around it by placing it in a goroutine in my application so that it runs parallel, but ideally it should eventually be an api call with context.Background().

More work is in progress, going to take a swing at updating client/tailscale next.

maisem:
Fixed in #6450

PR #4739, opened 2022-05-26
ipn/localapi: implement LoginInteractive via localapi

jtcressy:
Updates: #4738

This is the server-side change for implementing an Interactive Login via the localapi.

I noticed that the function `StartLoginInteractive()` on `LocalBackend` has no error return type, so i'll be looking to the core maintainers/reviewers if this is acceptable for localapi. I think it would be beneficial to catch errors and return 500 just like the logout api but i'm leaving that open for review.

### DIFF
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -109,6 +109,8 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		h.serveStatus(w, r)
 	case "/localapi/v0/logout":
 		h.serveLogout(w, r)
+	case "/localapi/v0/login-interactive":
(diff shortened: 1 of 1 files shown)


bradfitz (approved):
Looks good. It'd be a bit nicer if you actually used this new localapi in the client (cmd/tailscale/cli) to help the migration away from the old IPC mechanism but we can do that later.

jtcressy:
> Looks good. It'd be a bit nicer if you actually used this new localapi in the client (cmd/tailscale/cli) to help the migration away from the old IPC mechanism but we can do that later.
> 
> 

I'm going to see if I can get it implemented in client/tailscale next, but I'll also look into cmd/tailscale/cli unless y'all beat me to it.

PR #4740, opened 2022-05-26
wgengine/router: delete hardcoded link-local address on Windows

### LINKED ISSUE #4647: When using MagicDNS on Windows 10, you can't ping yourself by bare hostname
### What is the issue?

When using MagicDNS, you can't ping your own machine by hostname.

```
[xena@xatci ~] $ ping xatci

Pinging xatci [fe80::99d0:ec2d:b2e7:536b%5] with 32 bytes of data:
Request timed out.

Ping statistics for fe80::99d0:ec2d:b2e7:536b%5:
    Packets: Sent = 1, Received = 0, Lost = 1 (100% loss),
Control-C
```

### Steps to reproduce

1. Create fresh windows VM from either the Windows 10 MSIX packaging environment or [this vagrant box](https://app.vagrantup.com/gusztavvargadr/boxes/windows-10)
2. run `ping (hostname)`
3. see it work
4. install Tailscale and log in
5. run `ping (hostname)`
6. see it not work

_No response_

### OS

Windows

### OS version

10, 11

### Tailscale version

1.25.0

### Bug report

BUG-495ff6490a5112f9cb8a60d1e2598749bd609d5110641115b67dffe609b339be-20220509150038Z-0e69a407d89e9728

twitchyliquid64:
Fixes #4647

It seems that Windows creates a link-local address for the TUN driver, seemingly based on the (fixed) adapter GUID. This results in a fixed MAC address, which for some reason doesn't handle loopback correctly. Given the derived link-local
address is preferred for lookups (thanks LLMNR), traffic which addresses the current node by hostname uses this broken address and never works.

To address this, we remove the broken link-local address from the wintun adapter.

@raggi IIRC you have a Windows machine? Are you able to test this live?

### DIFF
--- a/wgengine/router/ifconfig_windows.go
+++ b/wgengine/router/ifconfig_windows.go
@@ -27,6 +27,8 @@ import (
 	"tailscale.com/wgengine/winnet"
 )
 
+var wintunLinkLocal = netaddr.MustParseIP("fe80::99d0:ec2d:b2e7:536b")
+
 // monitorDefaultRoutes subscribes to route change events and updates
 // the Tailscale tunnel interface's MTU to match that of the
 // underlying default route.
@@ -576,7 +578,18 @@ func excludeIPv6LinkLocal(in []*net.IPNet) (out []*net.IPNet) {
 	out = in[:0]
 	for _, n := range in {
 		if len(n.IP) == 16 && n.IP.IsLinkLocalUnicast() {
-			continue
+			// Windows creates a fixed link-local address for wintun,
+			// which doesn't seem to route correctly. Unfortunately, LLMNR returns this
+			// address for lookups by the hostname, and Windows prefers using it.
+			// This means that local traffic addressed to the machine's hostname breaks.
+			//
+			// While we otherwise preserve link-local addresses, we delete
+			// this one to force lookups to use a working address.
+			//
+			// See: https://github.com/tailscale/tailscale/issues/4647
+			if ip, ok := netaddr.FromStdIP(n.IP); !ok || wintunLinkLocal != ip {
+				continue // filter this IPNet
+			}
 		}
 		out = append(out, n)
 	}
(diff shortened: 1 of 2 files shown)


twitchyliquid64:
I'm also not sure if `n.Contains` is the right way to do the check (over something like simple equality), thoughts?

raggi:
> @raggi IIRC you have a Windows machine? Are you able to test this live?

Yep, will do so

wgengine/router/ifconfig_windows.go
```diff
@@ -576,7 +578,18 @@ func excludeIPv6LinkLocal(in []*net.IPNet) (out []*net.IPNet) {
...
+			// Windows creates a fixed link-local address for wintun,
+			// which doesn't seem to route correctly. Unfortunately, LLMNR returns this
+			// address for lookups by the hostname, and Windows prefers using it.
+			// This means that local traffic addressed to the machine's hostname breaks.
+			//
+			// While we otherwise preserve link-local addresses, we delete
+			// this one to force lookups to use a working address.
+			//
+			// See: https://github.com/tailscale/tailscale/issues/4647
+			if isWintun := n.Contains(wintunLinkLocal); !isWintun {
```
raggi, line 590: The netaddr package provides an easy path to equality:

```
winTunLinkLocal = netaddr.MustParseIP(...)

netaddr.FromStdIP(n.IP) == winTunLinkLocal
```

raggi (approved):

raggi:
> >@raggi IIRC you have a Windows machine? Are you able to test this live?
> Yep, will do so

It took a bit, as I had to diagnose and fix a regression from a while ago, but it's good that was caught.

This worked great on my Windows 11 machine.

wgengine/router/ifconfig_windows.go
twitchyliquid64: Its weird to me that the route is registered as a `/64`. Would that mean that for `fe80::99d0:ec2d:b2e7`, it would get any link-local address? Which I guess implies that my original `.Contains()` would delete every link-local address on the interface.

Maybe this is just a quirk of the API: as long as the IP is `fe80::99d0:ec2d:b2e7` from Windows, the subnet mask shouldn't matter. Do you recall if the API gave us that or just `fe80::/64`?

Issue #4607, opened 2022-05-04
Windows Split DNS is overridden by GPO configured by AD domains


sailorfrag:
### What is the issue?

When a Tailnet has split DNS configured, the Tailscale client configures Windows 8 and newer using NRPT to selectively send DNS queries for particular DNS suffixes to the Tailscale client. The Tailscale client then forwards the requests along to the appropriate DNS server.

Unfortunately, a couple of customers have reported that this does not work at all for them. Evidence indicates that the NRPT rules are simply being ignored. In both cases, they have configured a GPO to set various DNS settings (such as search domains).

### Steps to reproduce

1. Install Tailscale on a Windows 10 Enterprise computer that is part of an AD domain
2. Configure a GPO that sets DNS search domains
3. Configure split DNS on the Tailnet
4. Using `ping`, Wireshark and/or the other tool of your choice (not `nslookup` because it is not aware of fancy DNS lookup rules) check some hostnames that only resolve correctly if split DNS rules are being followed.

### Are there any recent changes that introduced the issue?

No, this appears to have been a problem since the last major Windows DNS implementation change.

### OS

Windows

### OS version

Confirmed on Windows 10, probably affects 8 and newer

### Tailscale version

1.24.0

sailorfrag:
I'm still actively investigating this issue and will provide updates when available.

sailorfrag:
In terms of workarounds for now, I think the best I can suggest is to configure "override local DNS" and configure the global resolvers to one of the public DNS services that use anycast (in particular, Google at 8.8.8.8, Cloudflare at 1.1.1.1, Quad9 at 9.9.9.9 will all automatically be recognized by Tailscale and automatically upgrade to using DoH); those should provide reasonably good localized DNS results regardless of location. The full list of public DNS services that our client recognizes is at https://github.com/tailscale/tailscale/blob/main/net/dns/publicdns/publicdns.go -- services not in that list should still work, but won't use DoH.

sailorfrag:
Easier reproduction step: in the key `HKLM\Software\Policies\Microsoft\Windows NT\DNSClient` (create it if needed), create a new string value SearchList and put something in there (it's a comma separated list of domain names). Split DNS immediately stops working.

[...]

PR #4742, opened 2022-05-26
net/dns, paths, util/winutil: change net/dns/windowsManager NRPT mana…

dblohm7:
…gement to support more than 50 domains.

AFAICT this isn't documented on MSDN, but based on the issue referenced below,
NRPT rules are not working when a rule specifies > 50 domains.

This patch modifies our NRPT rule generator to split the list of domains
into chunks as necessary, and write a separate rule for each chunk.

For compatibility reasons, we continue to use the hard-coded rule ID, but
as additional rules are required, we generate new GUIDs. Those GUIDs are
stored under the Tailscale registry path so that we know which rules are ours.

I made some changes to winutils to add additional helper functions in support
of both the code and its test: I added additional registry accessors, and also
moved some token accessors from paths to util/winutil.

Fixes https://github.com/tailscale/coral/issues/63


net/dns/manager_windows.go
```diff
@@ -104,31 +134,95 @@ func delValue(key registry.Key, name string) error {
...
+	dr := len(domains) % nrptMaxDomainsPerRule
+
+	domainRulesLen := dq
+	if dr > 0 {
+		domainRulesLen++
+	}
+
+	nrptRuleIDs := loadRuleSubkeyNames()
+	for len(nrptRuleIDs) < domainRulesLen {
+		guid, err := windows.GenerateGUID()
```
sailorfrag, line 165: So when I was looking at GPO examples online, it looks like the registry key doesn't actually have to be a GUID, that's just a convention.

I've seen other examples that amounted to the equivalent of Tailscale1, Tailscale2, Tailscale3, etc.

net/dns/manager_windows.go
```diff
@@ -90,10 +109,21 @@ func (m windowsManager) ifPath(basePath string) string {
...
 		return err
 	}
+	return winutil.DeleteRegValue(nrptRuleIDValueName)
+}
+
+func delRuleKeys(nrptRuleIDs []string) error {
+	for _, rid := range nrptRuleIDs {
+		keyName := nrptBase + rid
+		if err := registry.DeleteKey(registry.LOCAL_MACHINE, keyName); err != nil && err != registry.ErrNotExist {
+			return err
```
sailorfrag, line 124: Would it be better to log and continue?

net/dns/manager_windows.go
```diff
@@ -59,7 +78,7 @@ func NewOSConfigurator(logf logger.Logf, interfaceName string) (OSConfigurator,
 	// boot up. The bootstrap resolver logic will save us, but it
 	// slows down start-up a bunch.
 	if ret.nrptWorks {
-		ret.delKey(nrptBase)
+		delAllRuleKeys()
```
sailorfrag, line 81: Log the error?

net/dns/manager_windows.go
```diff
@@ -104,31 +134,95 @@ func delValue(key registry.Key, name string) error {
...
+		guid, err := windows.GenerateGUID()
+		if err != nil {
+			return err
+		}
+		nrptRuleIDs = append(nrptRuleIDs, guid.String())
+	}
+
+	// Remove any surplus rules that are no longer needed.
+	ruleIDsToRemove := nrptRuleIDs[domainRulesLen:]
+	delRuleKeys(ruleIDsToRemove)
```
sailorfrag, line 174: log errors

net/dns/manager_windows.go
```diff
@@ -104,31 +134,95 @@ func delValue(key registry.Key, name string) error {
...
+
+	// We need to save the list of rule IDs to our Tailscale registry key so that
+	// we know which rules are ours during subsequent modifications to NRPT rules.
+
+	// Omit nrptSingleRuleID at index 0 when writing our list of rule IDs to the
```
sailorfrag, line 179: Seems like it would be simpler just to include `nrptSingleRuleID`?

sailorfrag (approved):
Also https://github.com/tailscale/tailscale/issues/4607#issuecomment-1130586168 indicates that maybe there's some kind of thing we should be doing to kick DNSClient into reloading its config.

net/dns/manager_windows.go
dblohm7: Per [this document](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-gpnrpt/2d34f260-1e9e-4a52-ac91-2056dfd29702) it is a GUID.

net/dns/manager_windows.go
dblohm7: Yeah, I'll change this.

net/dns/manager_windows.go
dblohm7: I updated the implementation to log.

net/dns/manager_windows.go
dblohm7: Yeah, I'm revising this.

Issue #4750, opened 2022-05-28
win7-32 Tailscale 1.25.x: exit node & taildrop receive no worky


DentonGentry:
### What is the issue?

On a 32 bit VM running Windows 7-32 and Tailscale 1.25.98, the following features don't work:
- when running as an exit node, clients cannot resolve DNS lookups
- receiving files sent via Taildrop (tried sending from macsys, Android, and Ubuntu)

I note a common factor in both is other nodes being able to contact the PeerAPI. We do see packets from other nodes to the PeerAPI without errors in the logs:

2022-05-27 20:29:19.2646484 -0700 PDT: Accept: TCP{100.102.2.96:42646 > 100.126.109.3:1} 60 tcp ok
2022-05-27 20:29:20.2294921 -0700 PDT: Accept: TCP{100.102.2.96:42646 > 100.126.109.3:1} 60 tcp ok
2022-05-27 20:29:22.2548828 -0700 PDT: Accept: TCP{100.102.2.96:42646 > 100.126.109.3:1} 60 tcp ok
2022-05-27 20:29:34.6972656 -0700 PDT: Accept: TCP{100.102.2.96:42646 > 100.126.109.3:1} 60 tcp ok

A bugreport from the Win7 machine shortly after sending it a taildrop file failed:
BUG-fbf02ec95dfadad29da9aa320cdf54d48ee63dbad1b911f3932731dc70377e9f-20220528033404Z-2fbf9d44b76ac92b

### OS

Windows

### OS version

7

### Tailscale version

1.25.98

### Bug report

BUG-fbf02ec95dfadad29da9aa320cdf54d48ee63dbad1b911f3932731dc70377e9f-20220528033404Z-2fbf9d44b76ac92b

DentonGentry:
The issue still happens with 1.25.103. The new `tailscale ping --peerapi` also fails:
```
# tailscale ping --peerapi ie8win7
ping "100.126.109.3" timed out
ping "100.126.109.3" timed out
^C
# tailscale ping --tsmp ie8win7
pong from ie8win7 (100.126.109.3, 1) via TSMP in 3ms
```

DentonGentry:
Bug report from the Win7 system shortly before sending `tailscale ping -peerapi win7` from an ubuntu system:
BUG-fbf02ec95dfadad29da9aa320cdf54d48ee63dbad1b911f3932731dc70377e9f-20220603134027Z-7f930e933915eb56

The ping from the Ubuntu system failed:
[...]

PR #4751, opened 2022-05-28
cmd/tailscale, ipn/ipnlocal: add "peerapi" ping type

bradfitz:
For debugging when stuff like #4750 isn't working.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -1700,6 +1700,27 @@ func (b *LocalBackend) StartLoginInteractive() {
 }
 
 func (b *LocalBackend) Ping(ctx context.Context, ip netaddr.IP, pingType tailcfg.PingType) (*ipnstate.PingResult, error) {
+	if pingType == tailcfg.PingPeerAPI {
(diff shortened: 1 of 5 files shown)


DentonGentry (approved):

ipn/ipnlocal/local.go
```diff
@@ -1715,6 +1736,51 @@ func (b *LocalBackend) Ping(ctx context.Context, ip netaddr.IP, pingType tailcfg
 	}
 }
 
+func peerForIP(nm *netmap.NetworkMap, ip netaddr.IP) (_ *tailcfg.Node, ok bool) {
```
maisem, line 1739: We should at some point just have a `map[netaddr.IP]*tailcfg.Node`

maisem: Maybe on the netmap struct

maisem (approved):

ipn/ipnlocal/local.go

bradfitz: Rather, I moved the method there for now. We can do the map later.

PR #4753, opened 2022-05-28
ssh/tailssh: handle session terminations better

maisem:


bradfitz (approved):

maisem:
@bradfitz said LGTM on slack

PR #4754, opened 2022-05-29
tsd: add package with System type to unify subsystem init, discovery

bradfitz:
Updates #8036

----

This gets rid of all the weird hacks like ResolvingEngine and
InternalsGetter for different subsystems to get at other parts. And
simplifies some places where we were passing an increasing number of
things around, rather than 1 "all the stuff" thing.

This still isn't wonderful, but I think it cleans things up a bit and
gets us closer to a nice place. There's still too much boilerplate;
there's probably room for various helpers to set things up, especially
in tests.

This removes four methods from wgengine.Engine that should've never
really belonged there.


cmd/tailscaled/tailscaled.go
```diff
@@ -620,44 +624,55 @@ func tryEngine(logf logger.Logf, netMon *netmon.Monitor, dialer *tsdial.Dialer,
...
 		}
 	}
-	e, err = wgengine.NewUserspaceEngine(logf, conf)
+	e, err := wgengine.NewUserspaceEngine(logf, conf)
 	if err != nil {
-		return nil, onlyNetstack, err
+		return onlyNetstack, err
 	}
-	return e, onlyNetstack, nil
+	e = wgengine.NewWatchdog(e)
```
maisem, line 671: this is new?

maisem: oh i see, looks right

PR #4756, opened 2022-05-29
ipn/ipnlocal: shutdown sshServer on tailscale down

maisem:
Also lazify SSHServer initialization to allow restarting the server on a
subsequent `tailscale up`

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -125,8 +125,7 @@ type LocalBackend struct {
 	newDecompressor       func() (controlclient.Decompressor, error)
 	varRoot               string // or empty if SetVarRoot never called
 	sshAtomicBool         syncs.AtomicBool
-	sshServer             SSHServer // or nil
-	shutdownCalled        bool      // if Shutdown has been called
+	shutdownCalled        bool // if Shutdown has been called
 
 	filterAtomic            atomic.Value // of *filter.Filter
 	containsViaIPFuncAtomic atomic.Value // of func(netaddr.IP) bool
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/local.go
```diff
@@ -1969,16 +1964,23 @@ func (b *LocalBackend) setPrefsLockedOnEntry(caller string, newp *ipn.Prefs) {
 		cc.Login(nil, controlclient.LoginDefault)
 	}
 
+	if oldp.RunSSH && oldp.WantRunning {
```
bradfitz, line 1967: if you add a method like `ShouldSSHBeUp()` on the prefs, then this is just:

````go
    if oldp.ShouldSSHBeUp() && !newp.ShouldSSHBeUp() {
     
````

.... which reads a bit nicer

File: ipn/ipnlocal/local.go [hunk 2]
```diff
@@ -1969,16 +1964,23 @@ func (b *LocalBackend) setPrefsLockedOnEntry(caller string, newp *ipn.Prefs) {
...
+		if !newp.WantRunning || !newp.RunSSH {
+			b.mu.Lock()
```
bradfitz, line 1969: why re-acquiring the lock? can this be done before the unlock on 1930?

ipn/ipnlocal/local.go
```diff
@@ -3367,11 +3369,28 @@ func (b *LocalBackend) DoNoiseRequest(req *http.Request) (*http.Response, error)
...
-	if b.sshServer == nil {
-		return errors.New("no SSH server")
+func (b *LocalBackend) sshServerOrInit() (_ SSHServer, err error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.sshServer != nil {
+		return b.sshServer, nil
+	}
+	if newSSHServer == nil {
+		return nil, errors.New("no SSH server")
```
bradfitz, line 3379: "no SSH server support" ?

bradfitz (approved):

ipn/ipnlocal/local.go
maisem: Done

ipn/ipnlocal/local.go
maisem: Good catch, done.

Issue #4619, opened 2022-05-05
FR: Add to "tailscale status --json" if a client is using an exit node, and if so.


baandab:
### What are you trying to do?

I am writing a script to ensure that my clients are using an exit mode.   When I parse "tailscale status --json", I cannot tell from the .Self portion of the JSON if the client is using an exit mode.  

```
    "Online": true,
    "KeepAlive": false,
    "ExitNode": false,
    "ExitNodeOption": false,
    "Active": false,
```

### How should we solve this?

Add to the "tailscale status --json" two new parameters to the .Self:

ExitNodeUsage:  true or false
ExitNodeHostaname:  <name of host>

### What is the impact of not solving this?

Currently just forcing the use of the exit node, but would prefer to be more surgical and only force it when it is not on:

`Tailscale up --accept-routes --exit-node=100.XXX.XX.XX --exit-node-allow-lan-access=true --reset`

nyghtowl:
PR #4761 Addresses this issue by including ExitNodeStatus in the json output when an exit node is enabled in the net and set for usage.

Issue #4729, opened 2022-05-24
Missing JSON Output with `up --json` Command in Tailnet on Mac


nyghtowl:
### What is the issue?

When in a tailnet on Mac, the `/Applications/Tailscale.app/Contents/MacOS/Tailscale up --json` has no output. It does work on Linux

### Steps to reproduce

**Step to Reproduce**
Run `/Applications/Tailscale.app/Contents/MacOS/Tailscale up --json` on a Mac 

**Result**
No output

**Expected output**

{
        "AuthURL": "https://....",
        "QR": "....",
        "BackendState": "...."
}

### OS

macOS

### OS version

macOS Monterey

### Tailscale version

1.24.2

### Bug report

BUG-d0bd1c24b81c1248b525c46c8a165f4017b867738a7158a71c894057228f355b-20220524001257Z-66fac9a57fad9b78

mihaip:
As best as I can tell, this is behaving across platforms. If Tailscale is already up/connected, `up --json` outputs nothing. If it's not, it will output JSON states:

```
~$ /Users/mihai/Library/Developer/Xcode/DerivedData/IPN-gegnlsvbjygrcefuhrpkzxtpjnqu/Build/Products/Debug/Tailscale.app/Contents/MacOS/Tailscale up --json
{
	"AuthURL": "https://login.tailscale.com/a/79cbdaddf439",
[...]

PR #4761, opened 2022-05-31
ipn/ipnstate: add ExitNodeStatus to share the exit node if it is in use.  #4619

nyghtowl:


ipn/ipnlocal/local.go
```diff
@@ -453,6 +453,24 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 	}
 }
 
+// exitNodeIP returns the exit node IP from and netMap.
```
bradfitz, line 456: from and?

ipn/ipnlocal/local.go
```diff
@@ -453,6 +453,24 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
...
+func (b *LocalBackend) exitNodeIP() (ip netaddr.IP) {
```
bradfitz, line 457: no need to name the result

see https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

bradfitz (commented):
Squash down into one commit?  With commit message format: https://golang.org/wiki/CommitMessage

bradfitz:
> Addresses Issue https://github.com/tailscale/tailscale/issues/4729

Conventionally we write either `Fixes #nnnn` or `Updates #nnnnn`

ipn/ipnlocal/local.go
nyghtowl: I've seen named results in other parts of the codebase so confirming our approach moving forward is to avoid this?

ipn/ipnlocal/local.go
```diff
@@ -453,6 +453,24 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
...
+		for _, bP := range b.netMap.Peers {
+			if bP.StableID == b.prefs.ExitNodeID {
+				if len(bP.Addresses) > 0 {
+					return bP.Addresses[0].IP()
+				}
+				break
+			}
+		}
+	}
+	return
```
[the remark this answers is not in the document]
nyghtowl: Going with documented and zero value return.

nyghtowl:
It provides the IP for the ExitNode. I talked with Dave A. about this approach of addressing the ask here to provide the IP if it has been set in prefs and to provide "" if it isn't being used.

bradfitz:
The commit message isn't quite right yet. You have:

```
ipn/ipnstate: add ExitNodeIP to share exit node in use  #4729
Signed-off-by: nyghtowl <warrick@tailscale.com
```

It should be:

```
ipn/ipnstate: add ExitNodeIP to share exit node in use

Updates #4729

Signed-off-by: nyghtowl <warrick@tailscale.com
```

But is #4729 even the right bug? I don't see how they're related.

ipn/ipnlocal/local.go
```diff
@@ -453,6 +453,24 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 	}
 }
 
+// exitNodeIP returns the exit node IP from prefs.
+func (b *LocalBackend) exitNodeIP() netaddr.IP {
+	if !b.prefs.ExitNodeIP.IsZero() {
+		return b.prefs.ExitNodeIP
+	} else {
```
bradfitz, line 460: remove else and outdent rest

ipn/ipnlocal/local.go
```diff
@@ -453,6 +453,24 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
...
+		// convert exit node ID to IP
+		for _, bP := range b.netMap.Peers {
```
bradfitz, line 462: what does `bP` stand for? How about just `p` for "peer"?

ipn/ipnlocal/local.go
```diff
@@ -453,6 +453,24 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 	}
 }
 
+// exitNodeIP returns the exit node IP from prefs.
+func (b *LocalBackend) exitNodeIP() netaddr.IP {
+	if !b.prefs.ExitNodeIP.IsZero() {
```
bradfitz, line 458: b.prefs is guarded by b.mu.  So either you need to acquire b.mu (and make sure that's safe for callers) or rename this `exitNodeIPLocked` (the Locked suffix conventionally means the mutex must already be held by the caller)

Also, check that b.prefs is non-nil.

ipn/ipnlocal/local.go
```diff
@@ -453,6 +453,24 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 	}
 }
 
+// exitNodeIP returns the exit node IP from prefs.
```
bradfitz, line 456: , or the zero value if an exit node is not in use or the current exit node is no longer in the netmap.

ipn/ipnlocal/local.go
nyghtowl: backend Peer but I'll change to p

ipn/ipnstate/ipnstate.go
```diff
@@ -132,8 +132,9 @@ type PeerStatus struct {
 	LastHandshake  time.Time // with local wireguard
 	Online         bool      // whether node is connected to the control plane
 	KeepAlive      bool
-	ExitNode       bool // true if this is the currently selected exit node.
-	ExitNodeOption bool // true if this node can be an exit node (offered && approved)
+	ExitNode       bool       // true if this is the currently selected exit node.
+	ExitNodeOption bool       // true if this node can be an exit node (offered && approved)
+	ExitNodeIP     netaddr.IP // exit node IP; assume zero value means exit node is unset
```
bradfitz, line 137: I just realized that this is a little weird that this is an IP address, because which? IPv4? IPv6? other ones in the future when we do "TailNICs" letting a node have multiple identities.

At least document this as being "one of the IPs of the exit node"?  Probably time for a non-same-line comment.

```
// ExitNodeIP is set to one of the IP addresses of the exit node when an exit
// node is in use.
//
// It's set to the zero value when either an exit node is not in use, or the selected exit node is no longer present in the netmap.
```

ipn/ipnlocal/local.go
```diff
@@ -501,6 +523,7 @@ func (b *LocalBackend) populatePeerStatusLocked(sb *ipnstate.StatusBuilder) {
 			ShareeNode:     p.Hostinfo.ShareeNode(),
 			ExitNode:       p.StableID != "" && p.StableID == b.prefs.ExitNodeID,
 			ExitNodeOption: exitNodeOption,
+			ExitNodeIP:     b.exitNodeIPLocked(),
```
bradfitz, line 526: Oh, sorry, I just realized this on at the wrong level of the status. You're putting this on PeerStatus which is in a loop for each peer (line 485).

We want just one of these in the status, not per-peer.

But we already have the ExitNode bool on PeerStatus and the TailscaleIPs field on the Peer, so all this info is already in the status output isn't it?

This change is taking the original issue's FR a bit too literally probably.

They don't want to do a loop, I guess, in which case putting it top-level as a sibling of the `Self` attribute is probably best.

nyghtowl: I've moved the status up to Self which makes sense from a local status view of whether an exit node is used.

On peer and on Self, ExitNode is whether it is an exit node, not if its using one and what its using which is the goal of ExitNodeIP. It makes sense that we'd also show if the peers are set to use an exit node and what node that is. Looking at that, I'd need to add that field to Node. Does that make sense to add ExitNodeIP field to Node?

ipn/ipnlocal/local.go
```diff
@@ -432,6 +432,7 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 			ss.HostName = b.netMap.Hostinfo.Hostname
 			ss.DNSName = b.netMap.Name
 			ss.UserID = b.netMap.User
+			ss.ExitNodeIP = b.exitNodeIPLocked()
```
bradfitz, line 435: I don't think this should be on Self (which is of type PeerStatus) at all, because then we have an unused field for all the Peers (which are also of type PeerStatus).

ipn/ipnstate/ipnstate.go
```diff
@@ -81,6 +84,18 @@ type TailnetStatus struct {
 	MagicDNSEnabled bool
 }
 
+// ExitNodeStatus is the exit node used by the local backend.
+type ExitNodeStatus struct {
+	// Congigured is the exit node's ID.
```
bradfitz, line 89: typo

ipn/ipnstate/ipnstate.go
```diff
@@ -81,6 +84,18 @@ type TailnetStatus struct {
...
+	Configured tailcfg.StableNodeID
+
+	// Online is whether the node is alive.
+	Online *bool
```
bradfitz, line 93: no need for the pointer

ipn/ipnstate/ipnstate.go
```diff
@@ -81,6 +84,18 @@ type TailnetStatus struct {
...
+
+	//FirstIP is the exit node's first IP address.
```
bradfitz, line 95: space after //

and clarify that this is the Tailscale IP

Or at this point, now that we have a struct, want to just do `TailscaleIPs []netaddr.IP` like Status and PeerStatus? Then it matches those two.

ipn/ipnlocal/local.go
```diff
@@ -425,6 +425,12 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 			s.CurrentTailnet.MagicDNSEnabled = b.netMap.DNS.Proxied
 			s.CurrentTailnet.Name = b.netMap.Domain
 		}
+		if b.prefs != nil && !b.prefs.ExitNodeID.IsZero() {
+			s.ExitNodeStatus = &ipnstate.ExitNodeStatus{}
+			s.ExitNodeStatus.Configured = b.prefs.ExitNodeID
+			s.ExitNodeStatus.Online = b.exitNodeOnline()
+			s.ExitNodeStatus.FirstIP = b.exitNodeIP()
```
bradfitz, line 432:
```go
s.ExitNodeStatus = &ipnstate.ExitNodeStatus{
      Configured: b.prefs.ExitNodeID,
....
}
```

ipn/ipnstate/ipnstate.go
bradfitz: But how about just naming this field "ID"?

Then docs are:

// ID is the currently selected exit node.

ipn/ipnstate/ipnstate.go
```diff
@@ -38,6 +38,9 @@ type Status struct {
 	TailscaleIPs []netaddr.IP // Tailscale IP(s) assigned to this node
 	Self         *PeerStatus
 
+	// ExitNodeStatus is the exit node information this node is using if in use.
```
bradfitz, line 41: Might be worth calling out the zero value a bit more, like:

```
// ExitNodeStatus describes the current exit node.
// If nil, an exit node is not in use.
```

ipn/ipnstate/ipnstate.go
```diff
@@ -81,6 +84,18 @@ type TailnetStatus struct {
 	MagicDNSEnabled bool
 }
 
+// ExitNodeStatus is the exit node used by the local backend.
+type ExitNodeStatus struct {
+	// Congigured is the exit node's ID.
+	Configured tailcfg.StableNodeID
+
+	// Online is whether the node is alive.
```
bradfitz, line 92: "the node" is slightly ambiguous. Probably worth clarifying "exit node"

bradfitz:
What happened here?

ipn/ipnstate/ipnstate.go
```diff
@@ -38,6 +38,10 @@ type Status struct {
 	TailscaleIPs []netaddr.IP // Tailscale IP(s) assigned to this node
 	Self         *PeerStatus
 
+	// ExitNodeStatus describes the current exit node.
+	// If nil, an exit nodes is not in use.
```
bradfitz, line 42: node singular

ipn/ipnstate/ipnstate.go
```diff
@@ -81,6 +85,18 @@ type TailnetStatus struct {
 	MagicDNSEnabled bool
 }
 
+// ExitNodeStatus is the exit node used by the local backend.
```
bradfitz, line 88: "local backend" is an internal term and these is somewhat user-facing docs.

Maybe just repeat some of the docs above?

// ExitNodeStatus describes the state of an exit node.

ipn/ipnstate/ipnstate.go
```diff
@@ -81,6 +85,18 @@ type TailnetStatus struct {
...
+type ExitNodeStatus struct {
+	// ID is the exit node's ID.
+	ID tailcfg.StableNodeID
+
+	// Online is whether the exit node is alive.
+	Online bool
+
+	// TailscaleIPS are the exit node's IP addresses assigned to the node.
```
bradfitz, line 96: IPs (on both lines)

ipn/ipnlocal/local.go
```diff
@@ -425,6 +425,13 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 			s.CurrentTailnet.MagicDNSEnabled = b.netMap.DNS.Proxied
 			s.CurrentTailnet.Name = b.netMap.Domain
 		}
+		if b.prefs != nil && !b.prefs.ExitNodeID.IsZero() {
```
bradfitz, line 428: move this into the `if b.netMap != nil` part too one line up?

ipn/ipnlocal/local.go
```diff
@@ -425,6 +425,13 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
...
+			s.ExitNodeStatus = &ipnstate.ExitNodeStatus{
+				ID:           b.prefs.ExitNodeID,
+				Online:       b.exitNodeOnline(),
+				TailscaleIPS: b.exitNodeIPS(),
```
bradfitz, line 432: these are both linear scans over all peers and both assume methods below look like they're safe to use anywhere but aren't documented with the `Locked` suffix and don't check that `b.netMap` is nil.

How about instead you add a method to to the `b.netMap` type like `PeerWithStableID` so you only do the linear search once (and later we can optimize it into a map; I think such a TODO already exists somewhere) and then this code is like:

```go
     if b.prefs != nil {
              if exitPeer, ok := b.netMap.PeerWithStableID(b.prefs.ExitNodeID); ok {
                  s.ExitNodeStatus = &ipnstate.ExitNodeStatus{
                       ....
```

ipn/ipnlocal/local.go
```diff
@@ -424,6 +424,16 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 			s.CurrentTailnet.MagicDNSSuffix = b.netMap.MagicDNSSuffix()
 			s.CurrentTailnet.MagicDNSEnabled = b.netMap.DNS.Proxied
 			s.CurrentTailnet.Name = b.netMap.Domain
+			if b.prefs != nil && !b.prefs.ExitNodeID.IsZero() {
+				if exitPeer, ok := b.netMap.PeerWithStableID(b.prefs.ExitNodeID); ok {
+					s.ExitNodeStatus = &ipnstate.ExitNodeStatus{
+						ID:           b.prefs.ExitNodeID,
+						Online:       *exitPeer.Online,
```
bradfitz, line 431: this can crash if `exitPeer.Online` is nil.

bradfitz (approved):

bradfitz:
The commit message might need updating:

> ipn/ipnstate: add ExitNodeStatus to share the exit node if it  ....

Symbols have changed since then, right?

bradfitz:
Oh, commit message is fine. Just the PR title here confused me.

bradfitz:
But the git title (first line of commit message) shouldn't be so wide.

Some guides online say 80, or 76, or 72, or 50 characters. 

I'd probably say 76 or 80 is fine.

PR #4767, opened 2022-06-01
 hostinfo: use ByteSliceToString from golang.org/x/sys/unix

tklauser:
Use `unix.ByteSliceToString` in `osVersionFreebsd` and `osVersionLinux` to
convert the `Utsname.Release` `[]byte` field to `string`.

### DIFF
--- a/hostinfo/hostinfo_freebsd.go
+++ b/hostinfo/hostinfo_freebsd.go
@@ -27,12 +27,7 @@ func osVersionFreebsd() string {
 
 	var attrBuf strings.Builder
 	attrBuf.WriteString("; version=")
-	for _, b := range un.Release {
-		if b == 0 {
-			break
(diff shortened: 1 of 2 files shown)


hostinfo/hostinfo_freebsd.go
```diff
@@ -25,15 +25,7 @@ func osVersionFreebsd() string {
 	un := unix.Utsname{}
 	unix.Uname(&un)
 
-	var attrBuf strings.Builder
-	attrBuf.WriteString("; version=")
```
bradfitz, line 29: This string is lost now?

tklauser: :facepalm: fixed

bradfitz:
Nice seeing you around here! :)

PR #4775, opened 2022-06-02
net/dns/resolver: add support for <IPv4>.via-<site-id>

maisem:
Currently we only support `via-<site-id>.<IPv4>`, however that does not
work with Google Chrome which parses `http://via-1.10.0.0.1` as a search
string and not as a URL. This commit introduces `<IPv4>.via-<site-id>`
(`http://10.0.0.1.via-1`) which is parsed correctly by Chrome.

Updates #3616

### DIFF
--- a/net/dns/resolver/tsdns.go
+++ b/net/dns/resolver/tsdns.go
@@ -261,7 +261,6 @@ func (r *Resolver) Close() {
 // bound on per-query resource usage.
 const dnsQueryTimeout = 10 * time.Second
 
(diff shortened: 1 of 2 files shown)


twitchyliquid64 (approved):

bradfitz (approved):
But add some comments in the code about why there are two and Chrome and which is the now-preferred way and that one is deprecated and we should remove it in N releases from YYYY-MM-DD?

PR #4777, opened 2022-06-02
net/dns: update Windows split DNS settings to work alongside other NR…

dblohm7:
…PT entries set by group policy.

When there are group policy entries for the NRPT that do not belong to Tailscale,
we recognize that we need to add ourselves to group policy and use that registry
key instead of the local one. We also refresh the group policy settings as
necessary to ensure that our changes take effect immediately.

Fixes https://github.com/tailscale/tailscale/issues/4607

### DIFF
--- a/net/dns/nrpt_windows.go
+++ b/net/dns/nrpt_windows.go
@@ -1,0 +1,331 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package dns
+
+import (
+	"fmt"
+	"strings"
+
+	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/registry"
+	"tailscale.com/types/logger"
+	"tailscale.com/util/dnsname"
+	"tailscale.com/util/winutil"
+)
+
+const (
+	dnsBaseGP     = `SOFTWARE\Policies\Microsoft\Windows NT\DNSClient`
+	nrptBaseLocal = `SYSTEM\CurrentControlSet\Services\Dnscache\Parameters\DnsPolicyConfig`
+	nrptBaseGP    = `SOFTWARE\Policies\Microsoft\Windows NT\DNSClient\DnsPolicyConfig`
+
+	nrptOverrideDNS = 0x8 // bitmask value for "use the provided override DNS resolvers"
+
+	// Apparently NRPT rules cannot handle > 50 domains.
+	nrptMaxDomainsPerRule = 50
+
+	// This is the legacy rule ID that previous versions used when we supported
+	// only a single rule. Now that we support multiple rules are required, we
+	// generate their GUIDs and store them under the Tailscale registry key.
+	nrptSingleRuleID = `{5abe529b-675b-4486-8459-25a634dacc23}`
+
+	// This is the name of the registry value we use to save Rule IDs under
+	// the Tailscale registry key.
+	nrptRuleIDValueName = `NRPTRuleIDs`
+)
+
+var (
+	libUserenv          = windows.NewLazySystemDLL("userenv.dll")
+	procRefreshPolicyEx = libUserenv.NewProc("RefreshPolicyEx")
+)
+
+const _RP_FORCE = 1 // Flag for RefreshPolicyEx
+
+// nrptRuleDatabase ensapsulates access to the Windows Name Resolution Policy
+// Table (NRPT).
+type nrptRuleDatabase struct {
+	logf      logger.Logf
+	ruleIDs   []string
+	writeAsGP bool
+	isGPDirty bool
+}
+
+func newNRPTRuleDatabase(logf logger.Logf) *nrptRuleDatabase {
+	ret := &nrptRuleDatabase{logf: logf}
+	ret.loadRuleSubkeyNames()
+	ret.initWriteAsGP()
+	logf("nrptRuleDatabase using group policy: %v\n", ret.writeAsGP)
+	// Best-effort: if our NRPT rule exists, try to delete it. Unlike
+	// per-interface configuration, NRPT rules survive the unclean
+	// termination of the Tailscale process, and depending on the
+	// rule, it may prevent us from reaching login.tailscale.com to
+	// boot up. The bootstrap resolver logic will save us, but it
+	// slows down start-up a bunch.
+	ret.DelAllRuleKeys()
+	return ret
+}
+
+func (db *nrptRuleDatabase) loadRuleSubkeyNames() {
+	result := winutil.GetRegStrings(nrptRuleIDValueName, nil)
+	if result == nil {
+		// Use the legacy rule ID if none are specified in our registry key
+		result = []string{nrptSingleRuleID}
+	}
+	db.ruleIDs = result
+}
+
+// initWriteAsGP determines which registry path should be used for writing
+// NRPT rules. If there are rules in the GP path that don't belong to us, then
+// we should use the GP path.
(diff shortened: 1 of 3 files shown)


DentonGentry:
At this point even if we get an approval I would request that this not be submitted until after 1.26 branches, as it doesn't seem suitable as a last-moment submission. It would be best if it go in the early part of the 1.27.x test cycle to maximize the exposure amongst Windows sites with interesting group policy usage.

net/dns/nrpt_windows.go
```diff
@@ -0,0 +1,297 @@
...
+		return
+	}
+	defer dnsKey.Close()
+
+	ki, err := dnsKey.Stat()
+	if err != nil {
+		return
+	}
+
+	// If the dnsKey contains any values, then we need to use the GP key.
```
sailorfrag, line 99: It also looks like empty subkeys can also trigger DNS Client: https://github.com/tailscale/tailscale/issues/4607#issuecomment-1146908637

So filtering out our subkeys is irrelevant unless we're willing to delete an empty `DnsPolicyConfig`.

net/dns/nrpt_windows.go
```diff
@@ -0,0 +1,297 @@
...
+	db.ruleIDs = nil
+	return nil
+}
+
+// delRuleKeys removes the NRPT rules specified by nrptRuleIDs from the
+// Windows registry. It attempts to remove the rules from both possible registry
+// keys: the local key and the group policy key.
+func (db *nrptRuleDatabase) delRuleKeys(nrptRuleIDs []string) error {
+	for _, rid := range nrptRuleIDs {
+		keyNameLocal := nrptBaseLocal + `\` + rid
```
sailorfrag, line 158: Is there no equivalent to `filepath.Join()` for registry keys? :-(

I always end up worrying about random quoting type issues, but I suppose the left is a constant and the right is a GUID, so it should be fine.

net/dns/nrpt_windows.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil
+	}
+
+	if err := winutil.SetRegStrings(nrptRuleIDValueName, ruleIDsToWrite); err != nil {
+		return err
+	}
+	db.ruleIDs = ruleIDsToWrite
+
+	doms := make([]string, 0, nrptMaxDomainsPerRule)
+	for i := 0; i < domainRulesLen; i++ {
```
sailorfrag, line 220: optional: I think this loop would be simpler if written as a range over `domains` and it moves to the next rule when the current rule is full. I may be wrong; it'll definitely have both pros and cons.

net/dns/nrpt_windows.go
```diff
@@ -0,0 +1,297 @@
...
+	return nil
+}
+
+// Refresh notifies the Windows group policy engine when policies have changed.
+func (db *nrptRuleDatabase) Refresh() {
+	if !db.isGPDirty {
+		return
+	}
+	ok, _, err := procRefreshPolicyEx.Call(
+		uintptr(1), // Refresh computer policy, not user policy
```
sailorfrag, line 253: is there a relevant constant name for that?

net/dns/nrpt_windows.go
```diff
@@ -0,0 +1,297 @@
...
+		uintptr(_RP_FORCE),
```
sailorfrag, line 254: Rather than forcing, is there a way that we could actually create a real policy that Windows loads and applies? It seems like that would ensure all the correct mechanisms get tickled along the way, and perhaps would make cleanup easier too.

net/dns/nrpt_windows.go
```diff
@@ -0,0 +1,297 @@
...
+	isGPDirty bool
+}
+
+func newNRPTRuleDatabase(logf logger.Logf) *nrptRuleDatabase {
+	ret := &nrptRuleDatabase{logf: logf}
+	ret.loadRuleSubkeyNames()
+	ret.initWriteAsGP()
+	// Best-effort: if our NRPT rule exists, try to delete it. Unlike
+	// per-interface configuration, NRPT rules survive the unclean
+	// termination of the Tailscale process, and depending on the
```
sailorfrag, line 60: Does the uninstaller clear these? Maybe we should think about how to do that.

Are they cleared on clean shutdown?

sailorfrag (approved):

sailorfrag:
Sorry for the delay, I started doing the review a half dozen times but then some interruption happened each time.

net/dns/nrpt_windows.go
```diff
@@ -0,0 +1,297 @@
...
+		// Use the legacy rule ID if none are specified in our registry key
+		result = []string{nrptSingleRuleID}
+	}
+	db.ruleIDs = result
+}
+
+// initWriteAsGP determines which registry path should be used for writing
+// NRPT rules. If there are rules in the GP path that don't belong to us, then
+// we should use the GP path.
+func (db *nrptRuleDatabase) initWriteAsGP() {
```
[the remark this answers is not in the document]
dblohm7: Maybe use the `RegisterGPNotification` stuff like I used in the test to monitor for group policy flushes?

I think I'd prefer doing that in a follow-up PR, but it seems like a good idea.

net/dns/nrpt_windows.go
dblohm7: AFAICT from testing, we need to delete an empty `DnsPolicyConfig` or else the policy engine still thinks there are rules, even if there aren't any. I've modified the code to incorporate that.

net/dns/nrpt_windows.go
dblohm7: No, there's nothing like that, unfortunately.

net/dns/nrpt_windows.go
dblohm7: It's actually a Windows `BOOL` (which in Go is a `int32`). I suppose I could just make one.

net/dns/nrpt_windows.go
dblohm7: You know, I think we might have an issue there with other things too (eg changes made to resolv.conf in WSL not being reverted during uninstall). I'll file a follow-up issue.

dblohm7: https://github.com/tailscale/corp/issues/5621

dblohm7: IIRC we do clear them when we shutdown cleanly, though. I'll double-check.

net/dns/nrpt_windows.go
dblohm7: Yeah, it seems a bit cleaner. Nice catch!

net/dns/nrpt_windows.go
```diff
@@ -0,0 +1,297 @@
...
+	// the Tailscale registry key.
+	nrptRuleIDValueName = `NRPTRuleIDs`
+)
+
+var (
+	libUserenv          = windows.NewLazySystemDLL("userenv.dll")
+	procRefreshPolicyEx = libUserenv.NewProc("RefreshPolicyEx")
+)
+
+const _RP_FORCE = 1
```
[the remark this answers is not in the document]
dblohm7: AFAIK `nrptOverrideDNS` is not in any SDK header files under any name, so we need to declare it ourselves.

If you look at a package like `x/sys/windows`, the constants' public names match their C names, so I was matching that but with a leading underscore to keep it within the package.

net/dns/nrpt_windows.go
dblohm7: Unfortunately the changes aren't being seen without the force flag, so for now at least I think we need to keep it.

sailorfrag (approved):

dblohm7:
FYI my unit test hangs in GitHub CI but not locally, hence the repeated pushes. I'm trying to narrow this down.

PR #4782, opened 2022-06-03
tsnet: use proper log ID

### LINKED ISSUE #3866: tsnet: use proper log ID
### What is the issue?

I noticed a log ID conflict for multiple of my nodes. It's coming from inside the house:

https://github.com/tailscale/tailscale/blob/d7962e3bcf80c4d0837ac565f1cd17f87be3d07e/tsnet/tsnet.go#L174

If nothing else, we should generate a new log ID each time to use; better than a hard-coded string.

### Steps to reproduce

_No response_

_No response_

### OS

_No response_

### OS version

_No response_

### Tailscale version

_No response_

### Bug report

_No response_

willnorris:
refactor logpolicy config loading to make it easier to reuse from outside the package.  Within tsnet, setup a basic logtail config.

Fixes #3866

### DIFF
--- a/logpolicy/logpolicy.go
+++ b/logpolicy/logpolicy.go
@@ -95,17 +95,16 @@ type Policy struct {
 	PublicID logtail.PublicID
 }
 
-// Initialize the Config with the collection.
-// If the PrivateID is zero, it generates a new one.
-func (c *Config) Initialize(collection string) {
-	c.Collection = collection
-	if c.PrivateID.IsZero() {
-		var err error
-		c.PrivateID, err = logtail.NewPrivateID()
-		if err != nil {
-			panic("logtail.NewPrivateID should never fail")
-		}
-		c.PublicID = c.PrivateID.Public()
+// NewConfig creates a Config with collection and a newly generated PrivateID.
+func NewConfig(collection string) *Config {
+	id, err := logtail.NewPrivateID()
+	if err != nil {
+		panic("logtail.NewPrivateID should never fail")
+	}
+	return &Config{
+		Collection: collection,
+		PrivateID:  id,
+		PublicID:   id.Public(),
 	}
 }
 
@@ -116,7 +115,7 @@ func (c *Config) Validate(collection string) error {
 	case c.Collection != collection:
 		return fmt.Errorf("config collection %q does not match %q", c.Collection, collection)
 	case c.PrivateID.IsZero():
-		return errors.New("config has zero PrivateID ")
+		return errors.New("config has zero PrivateID")
 	case c.PrivateID.Public() != c.PublicID:
 		return errors.New("config PrivateID does not match PublicID")
 	}
@@ -510,12 +509,11 @@ func New(collection string) *Policy {
 	newc, err := ConfigFromFile(cfgPath)
 	if err != nil {
 		earlyLogf("logpolicy.ConfigFromFile %v: %v", cfgPath, err)
-		newc = &Config{}
-		newc.Collection = collection
+		newc = NewConfig(collection)
 	}
 	if err := newc.Validate(collection); err != nil {
 		earlyLogf("logpolicy.Config.Validate for %q: %v", cfgPath, err)
-		newc.Initialize(collection)
+		newc := NewConfig(collection)
 		if err := newc.Save(cfgPath); err != nil {
 			earlyLogf("logpolicy.Config.Save for %v: %v", cfgPath, err)
 		}
--- a/logpolicy/logpolicy.go
+++ b/logpolicy/logpolicy.go
@@ -95,6 +95,34 @@ type Policy struct {
 	PublicID logtail.PublicID
 }
 
+// Initialize the Config with the collection.
+// If the PrivateID is zero, it generates a new one.
+func (c *Config) Initialize(collection string) {
+	c.Collection = collection
+	if c.PrivateID.IsZero() {
+		var err error
+		c.PrivateID, err = logtail.NewPrivateID()
+		if err != nil {
+			panic("logtail.NewPrivateID should never fail")
+		}
+		c.PublicID = c.PrivateID.Public()
+	}
+}
+
+// Validate verifies that the Config matches the collection,
+// and that the PrivateID and PublicID pair are sensible.
+func (c *Config) Validate(collection string) error {
+	switch {
+	case c.Collection != collection:
+		return fmt.Errorf("config collection %q does not match %q", c.Collection, collection)
+	case c.PrivateID.IsZero():
+		return errors.New("config has zero PrivateID ")
+	case c.PrivateID.Public() != c.PublicID:
+		return errors.New("config PrivateID does not match PublicID")
+	}
+	return nil
+}
+
 // ToBytes returns the JSON representation of c.
 func (c *Config) ToBytes() []byte {
 	data, err := json.MarshalIndent(c, "", "\t")
@@ -105,7 +133,7 @@ func (c *Config) ToBytes() []byte {
 }
 
 // Save writes the JSON representation of c to stateFile.
-func (c *Config) save(stateFile string) error {
+func (c *Config) Save(stateFile string) error {
 	c.PublicID = c.PrivateID.Public()
 	if err := os.MkdirAll(filepath.Dir(stateFile), 0750); err != nil {
 		return err
@@ -117,7 +145,16 @@ func (c *Config) save(stateFile string) error {
 	return nil
 }
 
-// ConfigFromBytes parses a a Config from its JSON encoding.
+// ConfigFromFile reads a Config from a JSON file.
+func ConfigFromFile(statefile string) (*Config, error) {
+	b, err := os.ReadFile(statefile)
+	if err != nil {
+		return nil, err
+	}
+	return ConfigFromBytes(b)
+}
+
+// ConfigFromBytes parses a Config from its JSON encoding.
 func ConfigFromBytes(jsonEnc []byte) (*Config, error) {
 	c := &Config{}
 	if err := json.Unmarshal(jsonEnc, c); err != nil {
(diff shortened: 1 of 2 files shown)


tsnet/tsnet.go
```diff
@@ -180,11 +182,19 @@ func (s *Server) start() error {
 
 	// TODO(bradfitz): start logtail? don't use filch, perhaps?
 	// only upload plumbed Logf?
+	pol := logpolicy.New(logtail.CollectionNode)
+	defer func() {
```
dsnet, line 186: I'm not sure this runs at the right time. The `start` method appears to execute and then return quickly. It doesn't seem to block while the server is still running.

tsnet/tsnet.go
```diff
@@ -180,11 +182,19 @@ func (s *Server) start() error {
 
 	// TODO(bradfitz): start logtail? don't use filch, perhaps?
 	// only upload plumbed Logf?
+	pol := logpolicy.New(logtail.CollectionNode)
```
dsnet, line 185: I'm concerned about where this is going to locally store the logs data.

It appears to me that by default it figures it out from a bunch of environment variables:
https://github.com/tailscale/tailscale/blob/dce2409b15837f30885405b8b1d27e7b7fd6bf7a/logpolicy/logpolicy.go#L148-L218

However, I would have expect this to use `Server.Dir`.

dsnet: Part of the problem is that the `logpolicy` package was designed with the assumption that it's used within a main binary where only one instantiation ever occurs.

We may need to extend the API to support running many `logpolicy` instances within a single process.

dsnet:
BTW, if you want to pair program on this, I'd be happy to jump on a GVC, just let me know.

tsnet/tsnet.go
```diff
@@ -180,11 +182,19 @@ func (s *Server) start() error {
 
 	// TODO(bradfitz): start logtail? don't use filch, perhaps?
 	// only upload plumbed Logf?
```
bradfitz, line 184: What @dsnet said (assumptions about globals and only being run once early in the life of the program) and this comment here about "only upload plumbed logf" is why this bug is annoying and I skipped it all earlier :)

willnorris: @dsnet and I paired for an hour this morning, so I'll be pushing up more changes later.  I think things are in much better shape now.

logpolicy/logpolicy.go
```diff
@@ -573,6 +540,49 @@ func New(collection string) *Policy {
 	}
 }
 
+// LoadOrCreateConfig will load the logpolicy.Config from the specified path.
+// If it does not exist, a new config will be created.  If provided collection
+// does not match the loaded config, it will be overwritten and a new logID
+// generated.  Changes are saved back to cfgPath.
+func LoadOrCreateConfig(cfgPath string, collection string, logf logger.Logf) Config {
```
dsnet, line 547: This is an awkward API for an exported function.

What are you're thoughts on the following:
```go
// LoadConfig reads a JSON file as a Config.
func LoadConfig(file string) (Config, error)

// Store atomically writes the Config as a JSON file.
func (Config) Store(file string) error

// Validate verifies that the Config matches the collection,
// and that the PrivateID and PublicID pair and sensible.
func (Config) Validate(collection string) bool

// Initialize initializes the Config with the collection.
// If the PrivateID is zero, it generates a new one.
func (*Config) Initialize(collection string)
```

Also, I think the ways that we would use this in `logpolicy` and `tsnet` would slightly differ.
`logpolicy` probably does a best effort since we want `tailscaled` to be as autonomous as possible.
It'd be a sad day if upgrading tailscaled prevented it from starting up.

Possible usage in `logpolicy.go`:
```go
c, err := LoadConfig(cfgPath)
if err != nil {
	earlyLogf("logpolicy.LoadConfig %v: %v", cfgPath, err)
}
if !c.Validate(collection) {
	c.Initialize(collection)
	if err := c.Store(cfgPath); err != nil {
		earlyLogf("logpolicy.Config.Store %v: %v", cfgPath, err)
	}
}
```

On the other hand, I feel like `tsnet` adheres to stronger checks since they are usually managed services,
so we would want to eagerly return an error.

Possible usage in `tsnet.go`:
```go
c, err := logpolicy.LoadConfig(cfgPath)
switch {
case err == nil:
	if !c.Validate(logtail.CollectionNode) {
		return fmt.Errorf("logpolicy.Validate failed for %v", cfgPath)
	}
case os.IsNotExist(err):
	c.Initialize(logtail.CollectionNode)
	if err := c.Store(cfgPath); err != nil {
		return fmt.Errorf("logpolicy.Config.Store for %v: %w", cfgPath, err)
	}
default:
	return fmt.Errorf("logpolicy.LoadConfig for %v: %w", cfgPath, err)
}
```

https://go.dev/play/p/is9IXDK367g

willnorris: totally agreed... I didn't *love* this, but really like breaking it apart like you have!  I'll probably pick this back up next week.

bradfitz:
Ping me on chat when this is ready for review again in case I miss it.

tsnet/tsnet.go
```diff
@@ -178,8 +188,34 @@ func (s *Server) start() error {
...
-	// only upload plumbed Logf?
+	cfgPath := filepath.Join(s.rootPath, "tailscaled.log.conf")
+	collection := logtail.CollectionNode
+
+	lpc, err := logpolicy.ConfigFromFile(cfgPath)
+	if err != nil {
+		logf("logpolicy.ConfigFromFile %v: %v", cfgPath, err)
+	}
+	if !lpc.Validate(collection) {
+		logf("logpolicy.Config: config collection %q does not match %q", lpc.Collection, collection)
```
dsnet, line 199:
```go
return fmt.Errorf("logpolicy.Config.Validate %v: %w", cfgPath, err)
```

dsnet (approved):

tsnet/tsnet.go
```diff
@@ -178,8 +188,34 @@ func (s *Server) start() error {
...
 	}
 
-	// TODO(bradfitz): start logtail? don't use filch, perhaps?
-	// only upload plumbed Logf?
+	cfgPath := filepath.Join(s.rootPath, "tailscaled.log.conf")
+	collection := logtail.CollectionNode
+
+	lpc, err := logpolicy.ConfigFromFile(cfgPath)
+	if err != nil {
+		logf("logpolicy.ConfigFromFile %v: %v", cfgPath, err)
```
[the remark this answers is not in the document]
willnorris: why a different behavior from tailscaled?  I saw you had that, but couldn't rationalize why we would treat tsnet differently.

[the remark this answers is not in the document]
dsnet: Several thoughts:
1. Logging, but not erroring, is the prior behavior of `logpolicy`, so we shouldn't change that.
2. `tailscaled` prioritizes starting up at all costs.
    * For example, if you run `tailscaled` on a Raspberrry Pi in the woods, ensuring connectivity is of paramount importance. You wouldn't want to drive out to the woods to diagnose the problem.
3. `tsnet` is most often used in managed systems where the guarantee that things start up exactly as specified is more likely higher stronger prioty.
    * As a user of `tsnet`, I have a strong expectation that the logs for a given service will always go to a specific log stream ID. In the rare event of an IO error causing `LoadFromConfig` to fail, I'd rather have `tsnet` complain and for me to look into the issue than for it to automatically regenerate a new logstream ID and overwrite the original log config file.
    * I should have access to diagnose the problem since I can SSH into the machine and poke around.

logpolicy/logpolicy.go
```diff
@@ -469,38 +499,15 @@ func New(collection string) *Policy {
...
-			log.Fatalf("logpolicy: NewPrivateID() should never fail")
-		}
+		earlyLogf("logpolicy.ConfigFromFile %v: %v", cfgPath, err)
 	}
-	newc.PublicID = newc.PrivateID.Public()
-	if newc != *oldc {
-		if err := newc.save(cfgPath); err != nil {
-			earlyLogf("logpolicy.Config.Save: %v", err)
+	if !newc.Validate(collection) {
+		earlyLogf("logpolicy.Config: config collection %q does not match %q", newc.Collection, collection)
```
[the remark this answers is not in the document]
dsnet: SGTM.

tsnet/tsnet.go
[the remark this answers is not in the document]
willnorris: 👍🏻 I can buy that.

willnorris:
okay, PTA(nother)L

dsnet (approved):

Xe (approved):

tsnet/tsnet.go
```diff
@@ -178,8 +188,35 @@ func (s *Server) start() error {
 		return fmt.Errorf("%v is not a directory", s.rootPath)
 	}
 
-	// TODO(bradfitz): start logtail? don't use filch, perhaps?
-	// only upload plumbed Logf?
+	cfgPath := filepath.Join(s.rootPath, "tailscaled.log.conf")
```
bradfitz, line 191: tsnet.log.conf?

tsnet/tsnet.go
```diff
@@ -178,8 +188,35 @@ func (s *Server) start() error {
...
+
+	lpc, err := logpolicy.ConfigFromFile(cfgPath)
+	switch {
+	case os.IsNotExist(err):
+		lpc.Initialize(logtail.CollectionNode)
```
bradfitz, line 196: In normal Go conventions, `lpc` would be nil here, because unless documented explicitly, a func returning (T, error) should be assumed to have a `nil` or otherwise undefined T when error is non-nil.

So I think you should precede this with:

   lpc = new(...)
 
But at that point I'd rather see an explicit constructor, like:

    lpc = logtail.NewFoo(logtail.CollectionNode)

tsnet/tsnet.go
```diff
@@ -178,8 +188,35 @@ func (s *Server) start() error {
...
+		}
+	case err != nil:
+		return fmt.Errorf("logpolicy.LoadConfig for %v: %w", cfgPath, err)
+	}
+	if err := lpc.Validate(logtail.CollectionNode); err != nil {
+		return fmt.Errorf("logpolicy.Config.Validate for %v: %w", cfgPath, err)
+	}
+	logid := lpc.PublicID.String()
+
+	f, err := filch.New(filepath.Join(s.rootPath, "tailscaled"), filch.Options{ReplaceStderr: false})
```
bradfitz, line 208: not "tailscaled"

tsnet/tsnet.go
```diff
@@ -178,8 +188,35 @@ func (s *Server) start() error {
...
+	if err != nil {
+		return fmt.Errorf("error creating filch: %w", err)
+	}
+	c := logtail.Config{
+		Collection: lpc.Collection,
+		PrivateID:  lpc.PrivateID,
+		Stderr:     ioutil.Discard, // log everything to Buffer
+		Buffer:     f,
+		// TODO: consider adding compression. Look at logpolicy.go
```
bradfitz, line 217: let's do that now?

tsnet/tsnet.go
willnorris: done

tsnet/tsnet.go
willnorris: done

tsnet/tsnet.go
willnorris: done

logpolicy/logpolicy.go
```diff
@@ -95,6 +95,34 @@ type Policy struct {
...
+}
+
+// Validate verifies that the Config matches the collection,
+// and that the PrivateID and PublicID pair are sensible.
+func (c *Config) Validate(collection string) error {
+	switch {
+	case c.Collection != collection:
+		return fmt.Errorf("config collection %q does not match %q", c.Collection, collection)
+	case c.PrivateID.IsZero():
+		return errors.New("config has zero PrivateID ")
```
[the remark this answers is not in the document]
willnorris: done

logpolicy/logpolicy.go
```diff
@@ -95,6 +95,34 @@ type Policy struct {
 	PublicID logtail.PublicID
 }
 
+// Initialize the Config with the collection.
```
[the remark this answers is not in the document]
willnorris: done

bradfitz (approved):

tsnet/tsnet.go
dsnet: (Same [point as above](https://github.com/tailscale/tailscale/pull/4782#discussion_r891537228))

`tsnet` already uses the `tailscaled` file naming.

tsnet/tsnet.go
willnorris: oh, right 😕.  I don't have strong feelings on which it is, but yeah it should probably be consistent.

dsnet: The prior precedence occurs here: https://github.com/tailscale/tailscale/blob/09363064b544cd00e56fd4142e457119292e2588/tsnet/tsnet.go#L266

PR #4783, opened 2022-06-03
ipn/ipnserver, cmd/tailscaled: fix peerapi on Windows

bradfitz:
We weren't wiring up netstack.Impl to the LocalBackend in some cases
on Windows. This fixes Windows 7 when run as a service.

Updates #4750 (fixes after pull in to corp repo)


maisem (approved):

maisem (approved):

ipn/ipnserver/server.go
```diff
@@ -747,6 +748,9 @@ func Run(ctx context.Context, logf logger.Logf, ln net.Listener, store ipn.State
 	if err != nil {
 		return err
 	}
+	if ns != nil {
+		ns.SetLocalBackend(server.LocalBackend())
+	}
```
bradfitz, line 753: this is the missing wiring that `tailscaled.go` did but I forgot to do in `tailscaled_windows.go`.

ipn/ipnserver/server.go
```diff
@@ -659,7 +660,7 @@ func (s *Server) writeToClients(n ipn.Notify) {
 // The getEngine func is called repeatedly, once per connection, until it returns an engine successfully.
 //
 // Deprecated: use New and Server.Run instead.
-func Run(ctx context.Context, logf logger.Logf, ln net.Listener, store ipn.StateStore, linkMon *monitor.Mon, dialer *tsdial.Dialer, logid string, getEngine func() (wgengine.Engine, error), opts Options) error {
+func Run(ctx context.Context, logf logger.Logf, ln net.Listener, store ipn.StateStore, linkMon *monitor.Mon, dialer *tsdial.Dialer, logid string, getEngine func() (wgengine.Engine, *netstack.Impl, error), opts Options) error {
```
DentonGentry, line 663: Chonky argv

(which, if I understood it correctly, is something the Parts may help with so we don't have to pass all of these every time?)

DentonGentry (approved):

ipn/ipnserver/server.go
maisem: can ns ever be nil here?

ipn/ipnserver/server.go
bradfitz: yes but also: this whole `ipnserver.Run` entry point is deprecated and only used by Windows.

Windows only uses it because a bunch of things on Windows sporadically fail on start-up due to various async net races inside the Windows kernel so we have to keep retrying until things are happy.

We really need to move the retries around (or figure out what to actually wait for) so we can get rid of `ipnserver.Run` entirely.

ipn/ipnserver/server.go
bradfitz: In the test only I think? I just didn't want to think about it and deal with crashes later.

PR #4785, opened 2022-06-03
ipn/ipnlocal: prevent attempting to run SSH on Synology for now

bradfitz:
On DSM7 as a non-root user it'll run into problems.

And we haven't tested on DSM6, even though it might work, but I doubt
it.

Updates #3802
Updates tailscale/corp#5468

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -1833,7 +1833,10 @@ func (b *LocalBackend) checkPrefsLocked(p *ipn.Prefs) error {
 	if p.RunSSH {
 		switch runtime.GOOS {
 		case "linux":
-			// okay
+			if distro.Get() == distro.Synology && !envknob.UseWIPCode() {
+				return errors.New("The Tailscale SSH server does not run on Synology.")
+			}
+			// otherwise okay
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/local.go
```diff
@@ -1833,7 +1833,10 @@ func (b *LocalBackend) checkPrefsLocked(p *ipn.Prefs) error {
 	if p.RunSSH {
 		switch runtime.GOOS {
 		case "linux":
-			// okay
+			if distro.Get() == distro.Synology && !envknob.UseWIPCode() {
```
DentonGentry, line 1836: I don't know what `!envknob.UseWIPCode()` does but assume it is appropriate here.

DentonGentry (approved):

maisem (approved):

ipn/ipnlocal/local.go
bradfitz: It lets adventurous people (perhaps building it themselves and running as root or in Docker, etc) turn on our general I-know-what-I'm-doing env knob to let it run anyway. Same as darwin support a few lines down.

Issue #3808, opened 2022-01-25
cmd/tailscale: tailscale up --operator= doesn't reset operator


bradfitz:
The `up --operator=` flag was meant to be a little implicit to be less annoying, but it's too magic right now.

Even explicitly mentioning it and setting it to an empty value doesn't reset it. Likewise `--reset`:

```
$ go run ./cmd/tailscale up --operator=
$ go run ./cmd/tailscale up --reset
$ go run ./cmd/tailscale debug prefs
{
....
        "OperatorUser": "bradfitz",
        "Config": {
```

bradfitz:
Use your actual username to see.

It does seem that `--reset` got fixed, but the explicitly-empty bug remains:

```
bradfitz@tsdev:~/src/tailscale.com$ go run ./cmd/tailscale debug prefs | jq .OperatorUser
"bradfitz"
bradfitz@tsdev:~/src/tailscale.com$ go run ./cmd/tailscale up --operator=
bradfitz@tsdev:~/src/tailscale.com$ go run ./cmd/tailscale debug prefs | jq .OperatorUser
"bradfitz"
bradfitz@tsdev:~/src/tailscale.com$ go run ./cmd/tailscale up --reset
bradfitz@tsdev:~/src/tailscale.com$ go run ./cmd/tailscale debug prefs | jq .OperatorUser
null
```

(Tested at 69b535c01fe.)

Soypete:
As it stands if ```--reset``` is not called and the operator is blank then it defaults to the user
https://github.com/tailscale/tailscale/blame/69b535c01feb1066a92670d400afe870aca02b20/cmd/tailscale/cli/up.go#L853

Soypete:
https://github.com/tailscale/tailscale/pull/4795
adding tests to capture fix

Soypete:
let comment in the PR. I find that we need to change cli flag functionality to address this issue

bradfitz:
@Soypete, I've added a failing test you can use in your PR to test your fix:

https://github.com/tailscale/tailscale/commit/64bd0fc320fcd7a708bc290991def919895e53db

No other test changes or flag default changes should be required.

You should only need to modify `applyImplicitPrefs`, probably passing in a `flag.FlagSet` and having it check whether `--operator` was explicit set in the flagset.

PR #4795, opened 2022-06-06
cli/ : update up command to reset on blank string operator

Soypete:


cmd/tailscale/cli/cli_test.go
```diff
@@ -493,7 +494,7 @@ func TestCheckForAccidentalSettingReverts(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
-			applyImplicitPrefs(newPrefs, tt.curPrefs, tt.curUser)
+			// applyImplicitPrefs(newPrefs, tt.curPrefs, tt.curUser)
```
bradfitz, line 497: temporary debug?

cmd/tailscale/cli/up.go
```diff
@@ -105,7 +105,7 @@ func newUpFlagSet(goos string, upArgs *upArgsT) *flag.FlagSet {
 	upf.StringVar(&upArgs.advertiseRoutes, "advertise-routes", "", "routes to advertise to other nodes (comma-separated, e.g. \"10.0.0.0/8,192.168.0.0/24\") or empty string to not advertise routes")
 	upf.BoolVar(&upArgs.advertiseDefaultRoute, "advertise-exit-node", false, "offer to be an exit node for internet traffic for the tailnet")
 	if safesocket.GOOSUsesPeerCreds(goos) {
-		upf.StringVar(&upArgs.opUser, "operator", "", "Unix username to allow to operate on tailscaled without sudo")
+		upf.StringVar(&upArgs.opUser, "operator", os.Getenv("USER"), "Unix username to allow to operate on tailscaled without sudo")
```
maisem, line 108: this looks like a change in behavior?

Soypete: this + additions on line 809-812 replicate the removed function `applyImplicitPrefs()`

cmd/tailscale/cli/cli_test.go
Soypete: the function was removed and the functionality replaced... I need to remove the commend and function

cmd/tailscale/cli/cli_test.go
```diff
@@ -912,14 +945,16 @@ func TestUpdatePrefs(t *testing.T) {
...
+					if tt.curPrefs.OperatorUser == justEditMP.Prefs.OperatorUser {
+						t.Fatalf("current Operator User: %s, changes operator user: %s\n", tt.curPrefs.OperatorUser, justEditMP.OperatorUser)
+					}
+				}
 				justEditMP.Prefs = ipn.Prefs{} // uninteresting
 			}
 			if !reflect.DeepEqual(justEditMP, tt.wantJustEditMP) {
-				t.Logf("justEditMP != wantJustEditMP; following diff omits the Prefs field, which was %+v", oldEditPrefs)
-				t.Fatalf("justEditMP: %v\n\n: ", cmp.Diff(justEditMP, tt.wantJustEditMP, cmpIP))
+				t.Fatalf("justEditMP != wantJustEditMP; following diff omits the Prefs field, which was %+v\n", justEditMP)
```
Soypete, line 957: checks for the existing functionality

cmd/tailscale/cli/cli_test.go
```diff
@@ -912,14 +945,16 @@ func TestUpdatePrefs(t *testing.T) {
...
 				t.Fatalf("simpleUp=%v, want %v", simpleUp, tt.wantSimpleUp)
 			}
-			var oldEditPrefs ipn.Prefs
 			if justEditMP != nil {
-				oldEditPrefs = justEditMP.Prefs
+				if tt.wantOperatorChange {
+					if tt.curPrefs.OperatorUser == justEditMP.Prefs.OperatorUser {
+						t.Fatalf("current Operator User: %s, changes operator user: %s\n", tt.curPrefs.OperatorUser, justEditMP.OperatorUser)
+					}
+				}
```
Soypete, line 953: check for edge case in issue #3808

Soypete (commented):
this was checked via test and manually

cmd/tailscale/cli/cli_test.go
maisem: can you remove it in this PR? different commit is fine.

cmd/tailscale/cli/up.go
bradfitz: We can't make this change. Like Maisem says, this is a behavior change. We can't opt people in to this feature.

net/tstun/constants.go
```diff
@@ -0,0 +1,16 @@
...
+// DefaultMTU is the Tailscale default MTU for now.
+//
+// wireguard-go defaults to 1420 bytes, which only works if the
+// "outer" MTU is 1500 bytes. This breaks on DSL connections
+// (typically 1492 MTU) and on GCE (1460 MTU?!).
+//
+// 1280 is the smallest MTU allowed for IPv6, which is a sensible
+// "probably works everywhere" setting until we develop proper PMTU
+// discovery.
+const DefaultMTU = 1280
```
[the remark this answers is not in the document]
Soypete: working on this now! thanks!

cmd/tailscale/cli/up.go
Soypete: when I traced the behavior of the flags through the setup we initialize the default env value to be user [here](https://github.com/tailscale/tailscale/blob/4005134263db39b5a9e0f13f6db0e260f871a5c7/cmd/tailscale/cli/up.go#L463) and then we had an implicit check in the `applyImplicitPrefs()` that replaced that value with any non-empty string value set in a flag. 

This, to me, is the exact same functionality as explicitly setting the default value to user and removing the check for non-empty string. 

If this is an incorrect understanding of functionality, please help me understand? I might need an additional conversation on a meet or slack hangout to really understand the difference. 

additionally, what defines behavior we can and can't change?

Soypete: [the docs](https://tailscale.com/kb/1080/cli/) say that the default is `root`. So if you could help me understand what I am not understanding by suggesting this change

bradfitz: This value appears in the `up --help` text. I don't want to advertise any default for that flag, as by default there's no operator user.

That is, I don't like how it now does this:
```
bradfitz@tsdev:~/src/tailscale.com$ go run ./cmd/tailscale up --help 2>&1 | grep -A 1 operator
  --operator string
        Unix username to allow to operate on tailscaled without sudo (default bradfitz)

bradfitz@tsdev:~/src/tailscale.com$ go run --exec=sudo ./cmd/tailscale up --help 2>&1 | grep -A 1 operator
  --operator string
        Unix username to allow to operate on tailscaled without sudo (default root)
```

Both are misleading: the first makes it sound like the default is that `bradfitz` can operate `tailscaled` without sudo. (not true, but users might think that if they were the ones who installed it or something, guessing that we might've noticed that).  And the "root can operate tailscaled without sudo" part is also misleading, as root can always do anything without sudo.

Also, the fact that you had to add `wantOperatorChange: true` to some of the unrelated test cases is a bit of a red flag that there are behavior changes.

Soypete: > I don't like how it now does this
This to me should like we need a documentation change to match existing functionality. I am happy to work on that if it is needed. 
>  I don't want to advertise any default for that flag 
After the investigation and work put into the is #3808 I find that the only way to achieve the desired reset behavior without explicitly calling the `--reset` flag is to change functionality. If that is bigger than the scope of this issue I wonder if it is worth addressing.

Soypete:
this is too big and bloated. closing and moving to other [pr](https://github.com/tailscale/tailscale/pull/4899)

PR #4799, opened 2022-06-06
build_docker.sh: add run.sh as an entrypoint to the docker image

### LINKED ISSUE #4071: FR: Provide entrypoint for the Docker container

I'm trying to run a Tailscale subnet router on ECS Fargate.

Currently I have this entrypoint, cribbed from https://github.com/tailscale/tailscale/blob/main/docs/k8s/run.sh. It's a bit awkward running `tailscaled` and `tailscale` in the same container.

```
entrypoint = ["/bin/sh", "-c", "tailscaled --tun=userspace-networking & PID=$!; tailscale up --authkey $TAILSCALE_AUTHKEY --advertise-routes=10.30.0.0/16,10.31.0.0/16,10.32.0.0/16 --accept-dns=false && wait $PID"]
```

### How should we solve this?

I would imagine providing something similar to the [run.sh](https://github.com/tailscale/tailscale/blob/main/docs/k8s/run.sh) would work pretty well and be generally useful. It would let users provide flags for `tailscaled` and `tailscale` and reliably handle failures from both.

If I want to get a more robust script my options are (I think):

1. Create and publish a new Docker image which adds a script to the built-in container. This is fine but adds more maintenance cost to update it.
2. Source the full script in the entrypoint (if it would fit)

### Anything else?

Here's part of my Terraform config for this:

<details>

```terraform
resource "aws_ecs_task_definition" "tailscale" {
  # ...
  requires_compatibilities = ["FARGATE"]
  network_mode             = "awsvpc"

  container_definitions = jsonencode([
    {
      name       = "tailscale"
      image      = "tailscale/tailscale:v1.22.0"
      entrypoint = ["/bin/sh", "-c", "tailscaled --tun=userspace-networking & PID=$!; tailscale up --authkey $TAILSCALE_AUTHKEY --advertise-routes=10.30.0.0/16,10.31.0.0/16,10.32.0.0/16 --accept-dns=false && wait $PID"]

      essential = true
      secrets = [{
        "name" : "TAILSCALE_AUTHKEY",
        "valueFrom" : local.ts_subnet_router_authkey
      }]
    }
  ])
}
```

</details>

maisem:
Fixes #4071

### DIFF
--- a/build_docker.sh
+++ b/build_docker.sh
@@ -40,7 +40,9 @@ go run github.com/tailscale/mkctr \
     -X tailscale.com/version.Long=${VERSION_LONG} \
     -X tailscale.com/version.Short=${VERSION_SHORT} \
     -X tailscale.com/version.GitCommit=${VERSION_GIT_HASH}" \
+  --files="docs/k8s/run.sh:/tailscale/run.sh" \
   --base="${BASE}" \
   --tags="${TAGS}" \
   --repos="${REPOS}" \
-  --push="${PUSH}"
+  --push="${PUSH}" \
+  /bin/sh /tailscale/run.sh
--- a/docs/k8s/Dockerfile
+++ b/docs/k8s/Dockerfile
@@ -1,7 +1,0 @@
-# Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
-# Use of this source code is governed by a BSD-style
-# license that can be found in the LICENSE file.
-
-FROM ghcr.io/tailscale/tailscale:latest
-COPY run.sh /run.sh
-CMD "/run.sh"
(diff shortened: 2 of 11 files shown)


bradfitz (approved):

maisem:
I tested locally, working on setting up a k8s cluster.

maisem:
> LGTM. Normally we don't bump copyright years, btw.
> 
> Have you tested this after recent rename edits?

Yeah, confirmed that it works.

shadycuz:
@maisem I was looking at which docker containers have this new feature https://hub.docker.com/r/tailscale/tailscale/tags and thats when I noticed that dockerhub lists tailscale `1.27.X` and this repo only goes up to `1.26.X` is that a bug? Did someone accidentally push the wrong tag to dockerhub?

maisem:
@shadycuz odd numbers are built from HEAD and signify unstable builds

shadycuz:
@maisem  Thanks =), I see your commit is referenced in the `1.26.0` and `1.26.1` tags, so I'm guessing I will find your changes in those docker containers <3. 

Thanks for this. It was really important and will make tailscale much easier to implement on K8's. 🎉

PR #4801, opened 2022-06-06
cmd/controlclient: wire up PingRequest peerapi pings too

bradfitz:
Updates tailscale/corp#754

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -124,11 +124,10 @@ type Options struct {
 	Pinger Pinger
 }
 
-// Pinger is a subset of the wgengine.Engine interface, containing just the Ping method.
+// Pinger is the LocalBackend.Ping method.
 type Pinger interface {
-	// Ping is a request to start a ping with the peer handling the given IP and
-	// then call cb with its ping latency & method.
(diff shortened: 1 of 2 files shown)


control/controlclient/direct.go
```diff
@@ -1417,10 +1414,19 @@ func doPingerPing(logf logger.Logf, c *http.Client, pr *tailcfg.PingRequest, pin
...
-	})
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	res, err := pinger.Ping(ctx, pr.IP, pingType)
+	if err != nil {
+		if ctx.Err() != nil {
+			return
+		}
```
maisem, line 1425: that means that you won't log context timeout errors, why not?

maisem: Also maybe log a ping request so that we can confirm that it made it to the client?

maisem (approved):

raggi (approved):
I kind of want to make the wgengine side quack the same interface, but that can be cleaned up later.

bradfitz:
> I kind of want to make the wgengine side quack the same interface, but that can be cleaned up later.

I actually want to kill it. I've been working on removing all its dumb methods: https://github.com/tailscale/tailscale/pull/4754

PR #4812, opened 2022-06-07
tailcfg: add TailscaleSSHEnabled helper check

rosszurowski:
This commit adds a helper to check if SSH is enabled. We're currently checking the SSH_HostKeys field in a few places, but later plan to add an explicit bool. This helper makes the check and any future changes easier.

Signed-off-by: Ross Zurowski <ross@rosszurowski.com>

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -475,6 +475,14 @@ type Hostinfo struct {
 	//       require changes to Hostinfo.Equal.
 }
 
+// TailscaleSSHEnabled reports whether or not this node is acting as a
+// Tailscale SSH server.
+func (hi *Hostinfo) TailscaleSSHEnabled() bool {
+	// Currently, we use `SSH_HostKeys` as a proxy for this. However, we may later
+	// include non-Tailscale host keys, and will add a separate flag to rely on.
+	return hi != nil && len(hi.SSH_HostKeys) > 0
+}
+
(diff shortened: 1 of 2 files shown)


tailcfg/tailcfg.go
```diff
@@ -475,6 +475,14 @@ type Hostinfo struct {
 	//       require changes to Hostinfo.Equal.
 }
 
+// SSHEnabled returns whether or not this node is acting as a
```
bradfitz, line 478: "reports whether"

tailcfg/tailcfg.go
```diff
@@ -475,6 +475,14 @@ type Hostinfo struct {
...
+// Tailscale SSH server.
+func (hi *Hostinfo) SSHEnabled() bool {
```
bradfitz, line 480: Shall we name it less ambiguously? TailscaleSSHEnabled?

tailcfg/tailcfg.go
```diff
@@ -475,6 +475,14 @@ type Hostinfo struct {
...
+	// Currently, we use `SSH_HostKeys` as a proxy, but we may later rely on a
+	// separate flag.
```
bradfitz, line 482: mention that in the future SSH_HostKeys might include non-Tailscale SSH host keys?

bradfitz (approved):

PR #4818, opened 2022-06-08
Update a Taildrop error message

praneetloke:
Specifically, when trying to use the Taildrop feature while not connected to the Tailnet.

For reference, the video below captures the error message. It's really minor but I didn't pay attention to my tailscale status and was checking everything else but my actual connection to the tailnet.

https://user-images.githubusercontent.com/1466314/172544322-fc7b4d2f-be3e-4613-8ad7-f3a3c332d5ec.mp4

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -3108,7 +3108,7 @@ func (b *LocalBackend) FileTargets() ([]*apitype.FileTarget, error) {
 	defer b.mu.Unlock()
 	nm := b.netMap
 	if b.state != ipn.Running || nm == nil {
-		return nil, errors.New("not connected")
+		return nil, errors.New("not connected to the Tailnet")
 	}
 	if !b.capFileSharing {
 		return nil, errors.New("file sharing not enabled by Tailscale admin")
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/local.go
```diff
@@ -3108,7 +3108,7 @@ func (b *LocalBackend) FileTargets() ([]*apitype.FileTarget, error) {
 	defer b.mu.Unlock()
 	nm := b.netMap
 	if b.state != ipn.Running || nm == nil {
-		return nil, errors.New("not connected")
+		return nil, errors.New("not connected to the Tailnet")
```
DentonGentry, line 3111: Please lower-case "tailnet."

"Tailscale" is a name and should be capitalized. We consider _tailnet_ to be a regular noun.

ipn/ipnlocal/local_test.go
```diff
@@ -511,13 +511,13 @@ func TestLazyMachineKeyGeneration(t *testing.T) {
 func TestFileTargets(t *testing.T) {
 	b := new(LocalBackend)
 	_, err := b.FileTargets()
-	if got, want := fmt.Sprint(err), "not connected"; got != want {
+	if got, want := fmt.Sprint(err), "not connected to the Tailnet"; got != want {
```
praneetloke, line 514:
```suggestion
	if got, want := fmt.Sprint(err), "not connected to the tailnet"; got != want {
```

ipn/ipnlocal/local.go
praneetloke: Done.

PR #4819, opened 2022-06-08
cmd/tsconnect: initial scaffolding for Tailscale Connect browser client

mihaip:
Runs a Tailscale client in the browser (via a WebAssembly build of the
`wasm` package) and allows SSH access to machines. The `wasm` package exports
a `newIPN()` function, which returns a simple JS object with methods like
`start()`, `login()`, `logout()` and `ssh()`. The golang.org/x/crypto/ssh
package is used for the SSH client.

Terminal emulation and QR code rendering is done via NPM packages (`xterm`
and `qrcode` respectively), thus we also need a JS toolchain that can
install and bundle them. Yarn is used for installation, and esbuild
handles loading them and bundling for production serving.

Updates #3157


mihaip:
Still very preliminary, but I figured it was time to start cleaning things up to get this merged.

Open questions (mostly for Ross):
- There's now a yarn (and node) dependency. Should that be encapsulated in a script that pulls down a specific version, similar to tool/go?
- If we want this to match the look and feel of tailscale.com, we'd need to import CSS and other styles (though we already do some of this for the web CLI command)
- The UI may end up being more complex, and thus doing via vanilla JS may get annoying (especially to handle state changes for the backend and SSH sessions). It seems heavyweight to bring in React though, do you have preferences here?

rosszurowski:
> There's now a yarn (and node) dependency. Should that be encapsulated in a script that pulls down a specific version, similar to tool/go?

Yeah, it's important we (a) ensure everyone's on the exact same versions of node/yarn (or npm, we could start that transition here), and (b) that it's easy to get hacking on any part of the codebase.

As a _temporary_ approach, you could lock `engines` in `package.json` to exact versions and leave the toolchain for another PR. But we will need to the toolchain set up before long.

> If we want this to match the look and feel of tailscale.com, we'd need to import CSS and other styles (though we already do some of this for the web CLI command)

We currently have two kinds of CSS: (a) "simple" pages like our auth pages, `tailscale web`, and other one-offs. This is hand-written and embedded as a static file. And (b) "complex" pages which use Tailwind. Ideally, we'd use Tailwind everywhere — it'd give us a consistent and maintainable dev experience. The places where we haven't, it's simply because it's felt like overkill to introduce a build step.

Since we're adding a build step here anyways, using Tailwind would be my preference.

> The UI may end up being more complex, and thus doing via vanilla JS may get annoying (especially to handle state changes for the backend and SSH sessions). It seems heavyweight to bring in React though, do you have preferences here?

We don't have a formal policy, but on the website we use [Preact](https://preactjs.com/) for cases too complicated for vanilla JS (eg. KB search, downloads page). It's close enough to React + hooks that it gives us the same dev experience, just in a much smaller package size.

tailcfg/tailcfg.go
```diff
@@ -483,6 +483,8 @@ func (hi *Hostinfo) TailscaleSSHEnabled() bool {
 	return hi != nil && len(hi.SSH_HostKeys) > 0
 }
 
+func (v HostinfoView) TailscaleSSHEnabled() bool { return v.ж.TailscaleSSHEnabled() }
```
rosszurowski, line 486: Ah, thank you! I figured there was more places to add this, but wasn't sure what they were.

ssh/browser/src/index.js
```diff
@@ -0,0 +1,28 @@
...
+  fetch(`./dist/${wasmUrl}`),
+  go.importObject
+).then((result) => {
+  go.run(result.instance)
+  const ipn = newIPN({
+    // Persist IPN state in sessionStorage in development, so that we don't need
+    // to re-authorize every time we reload the page.
+    stateStorage: DEBUG ? sessionStateStorage : undefined,
+  })
+  console.log("will run")
```
rosszurowski, line 21: Same with these, maybe drop 'em behind a `DEBUG` check.

rosszurowski (commented):
I can't run it locally, but happy to leave more in-depth comments on the JS/CSS once I can!

ssh/browser/browser.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	if len(result.Warnings) > 0 {
+		log.Printf("ESBuild Warnings:\n")
+		for _, w := range result.Warnings {
+			log.Printf("%v", w)
+		}
+	}
+}
+
+//go:embed dist/* index.html
```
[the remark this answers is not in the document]
mihaip: Added a placeholder file to make sure the directory exists (same approach as the admin panel in corp), hopefully that should fix the bootstrapping problem.

tailcfg/tailcfg.go
mihaip: Will move this to a separate commit, it's likely we'll want this sooner

(I checked with Maisem that the views code gen tool does not handle custom methods)

ssh/browser/browser.go
```diff
@@ -0,0 +1,186 @@
...
+	"runtime"
+	"strconv"
+
+	esbuild "github.com/evanw/esbuild/pkg/api"
+)
+
+var (
+	dev   = flag.Bool("dev", false, "Run in dev mode (watching JS and CSS)")
+	build = flag.Bool("build", false, "Run in build mode (generating static assets)")
+	serve = flag.Bool("serve", false, "Run in serve mode (serving static assets)")
```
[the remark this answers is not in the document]
mihaip: `-dev` is meant to be a dev server, will clarify the descriptions.

ssh/browser/src/notifier.js
```diff
@@ -0,0 +1,73 @@
...
+      showLogoutButton(ipn)
+      break
+  }
+  const stateNode = document.getElementById("state")
+  stateNode.textContent = stateLabel ?? ""
+}
+
+export function notifyNetMap(ipn, netMapStr) {
+  const netMap = JSON.parse(netMapStr)
+  console.log("Received net map: " + JSON.stringify(netMap, null, 2))
```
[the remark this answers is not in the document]
mihaip: It's pretty noisy, moved to being behind a DEBUG check.

ssh/browser/src/index.js
mihaip: Leftover debugging code, removed.

ssh/browser/wasm/wasm_js.go
```diff
@@ -0,0 +1,406 @@
...
+	jsIPN := &jsIPN{
+		dialer: dialer,
+		srv:    srv,
+		lb:     lb,
+	}
+
+	return map[string]any{
+		"run": js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+			if len(args) != 1 {
+				log.Fatal("Usage: run(callbacks)")
```
bradfitz, line 112: document this interface a bit more? exactly 1 callback? what's the callback? When's it called? in what state(s)?

ssh/browser/wasm/wasm_js.go
```diff
@@ -0,0 +1,406 @@
...
+			if len(args) != 0 {
+				log.Printf("Usage: logout()")
+				return nil
+			}
+			jsIPN.logout()
+			return nil
+		}),
+		"ssh": js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+			if len(args) != 6 {
+				log.Printf("Usage: ssh(hostname, writeFn, readFn, row, cols, onDone)")
```
bradfitz, line 136: rows or row? cols is plural.

ssh/browser/wasm/wasm_js.go
```diff
@@ -0,0 +1,406 @@
...
+	setReadFn.Invoke(js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+		input := args[0].String()
+		_, err := stdin.Write([]byte(input))
+		if err != nil {
+			writeError("Write Input", err)
+		}
+		return nil
+	}))
+
+	err = session.RequestPty("xterm", rows, cols, ssh.TerminalModes{})
```
bradfitz, line 301: do you handle terminal resizes?

ssh/browser/wasm/wasm_js.go
```diff
@@ -0,0 +1,406 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package that generates JS bindings for Tailscale via WebAssembly.
```
bradfitz, line 5: // The wasm package is a WebAssembly module that provides Tailscale APIs to JavaScript.

?

ssh/browser/browser.go
```diff
@@ -0,0 +1,213 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Builds and serves the static site that is generated for the browser/Wasm
```
bradfitz, line 5: // The browser command ...

https://github.com/golang/go/wiki/CodeReviewComments#package-comments

ssh/browser/browser.go
```diff
@@ -0,0 +1,213 @@
...
+var (
+	dev   = flag.Bool("dev", false, "Run in dev build and serve mode")
+	build = flag.Bool("build", false, "Run in production build mode (generating static assets)")
+	serve = flag.Bool("serve", false, "Run in production serve mode (serving static assets)")
+	addr  = flag.String("addr", ":9090", "address to listen on")
+)
+
+func main() {
+	flag.Parse()
+
```
bradfitz, line 39:
```
if sum(*dev, *build, *serve) != 1 {
    log.Fatalf("exactly one of --dev, --build, --serve expected")
}
```

?

ssh/browser/browser.go
```diff
@@ -0,0 +1,213 @@
...
+	cmd := exec.Command(filepath.Join(runtime.GOROOT(), "bin", "go"), args...)
+	cmd.Env = append(os.Environ(), "GOOS=js", "GOARCH=wasm")
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+// installJsDeps installs the JavaScript dependencies specified by package.json
+func installJsDeps() error {
```
bradfitz, line 191: installJSDeps per https://github.com/golang/go/wiki/CodeReviewComments#initialisms but given as how this isn't exported API I don't care so much.

bradfitz (approved):
Seems fine on quick skim. Don't have time to review fully yet. I'll leave the JS review to others. :)

ssh/browser/browser.go
mihaip: Ended up going with a `cmd` arg instead of three mutually exclusive flags, that seemed to more clearly communicate intent.

mihaip:
Did a cleanup pass on how the dev/build/serve code is structured so that it's not all in one file, probably worth another look before merging.

mihaip:
(there are many things left to do, including the yarn boostrap, but this is getting large enough at is)

cmd/tsconnect/common.go
```diff
@@ -0,0 +1,103 @@
...
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+// installJSDeps installs the JavaScript dependencies specified by package.json
+func installJSDeps() error {
+	log.Printf("Installing JS deps...\n")
+	return exec.Command("yarn").Run()
```
bradfitz, line 102: Which "yarn"? I don't have any yarn in my path.

Also, this won't print stdout/stderr anywhere so it'll just kinda fail mysteriously.

cmd/tsconnect/common.go
```diff
@@ -0,0 +1,103 @@
...
+	"os/exec"
+	"path"
+	"path/filepath"
+	"runtime"
+	"strconv"
+
+	esbuild "github.com/evanw/esbuild/pkg/api"
+)
+
+type EntryPointMap = map[string]string
```
bradfitz, line 20: With a `=` this is a type alias. Type aliases are almost never what you want.

I'd just drop it. And then document this a bit. What are the keys/values?

cmd/tsconnect/common.go
```diff
@@ -0,0 +1,103 @@
...
+// JS/CSS bundling, which is  handled by esbuild).
+func buildDeps(dev bool) error {
+	if err := copyWasmExec(); err != nil {
+		return fmt.Errorf("Cannot copy wasm_exec.js: %v", err)
+	}
+	if err := buildWasm(dev); err != nil {
+		return fmt.Errorf("Cannot build main.wasm: %v", err)
+	}
+	if err := installJSDeps(); err != nil {
+		return fmt.Errorf("Cannot install JS deps: %v", err)
```
bradfitz, line 62: %w on all these instead of %v?

bradfitz (approved):

cmd/tsconnect/common.go
mihaip: Ended up dropping this preprocessing altogether, and thus the need for this type. I now persist the entire esbuild metadata JSON, and build this map at serving time (since we only need to do it once, during server startup).

cmd/tsconnect/common.go
mihaip: Added output when there's an error. Bootstrapping yarn (and node) is left to a followup commit.

PR #4821, opened 2022-06-08
cmd/tailscale: add 'debug ts2021' Noise connectivity subcommand

bradfitz:
Updates #3488


cmd/tailscale/cli/debug.go
```diff
@@ -425,3 +441,64 @@ func runVia(ctx context.Context, args []string) error {
...
+		}
+		return c, err
+	}
+
+	conn, err := controlhttp.Dial(ctx, net.JoinHostPort(ts2021Args.host, "80"), machinePrivate, keys.PublicKey, uint16(ts2021Args.version), dialFunc)
+	log.Printf("controlhttp.Dial = %p, %v", conn, err)
+
+	gotPeer := conn.Peer()
+	if gotPeer != keys.PublicKey {
+		log.Printf("peer = %v, want %v", gotPeer, keys.PublicKey)
```
maisem, line 497: this will panic if err is not nil

bradfitz: whoops, missed the return above

maisem (approved):

PR #4824, opened 2022-06-08
control/controlhttp: use secure WebSockets protocol by default

mihaip:
Forcing the insecure protocol (and perserving the port number) is only
desired for localhost testing, in prod we need to use wss:// to avoid
mixed-content errors.

### DIFF
--- a/control/controlhttp/client_js.go
+++ b/control/controlhttp/client_js.go
@@ -16,7 +16,7 @@ import (
 	"tailscale.com/types/key"
 )
 
-// Variant of Dial that tunnels the request over WebScokets, since we cannot do
+// Variant of Dial that tunnels the request over WebSockets, since we cannot do
 // bi-directional communication over an HTTP connection when in JS.
 func Dial(ctx context.Context, addr string, machineKey key.MachinePrivate, controlKey key.MachinePublic, protocolVersion uint16, dialer dnscache.DialContextFunc) (*controlbase.Conn, error) {
 	init, cont, err := controlbase.ClientDeferred(machineKey, controlKey, protocolVersion)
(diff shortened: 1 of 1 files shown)


control/controlhttp/client_js.go
```diff
@@ -28,9 +28,15 @@ func Dial(ctx context.Context, addr string, machineKey key.MachinePrivate, contr
 	if err != nil {
 		return nil, err
 	}
+	wsScheme := "wss"
+	wsHost := host
+	if host == "localhost" {
+		wsScheme = "ws"
+		wsHost = net.JoinHostPort(host, addr)
```
bradfitz, line 35: Line 27 makes this super confusing... addr changes from a host:port to just a port. Rename the variable on line 27?

mihaip: Yeah, not sure what I was thinking there. Renamed that to `port`

control/controlhttp/client_js.go
```diff
@@ -16,21 +16,27 @@ import (
...
-	host, addr, err := net.SplitHostPort(addr)
+	host, port, err := net.SplitHostPort(addr)
 	if err != nil {
 		return nil, err
 	}
+	wsScheme := "wss"
+	wsHost := host
+	if host == "localhost" {
+		wsScheme = "ws"
+		wsHost = net.JoinHostPort(host, port)
```
bradfitz, line 35: so now this is just `wsHost = addr` ?

control/controlhttp/client_js.go
```diff
@@ -16,21 +16,27 @@ import (
...
 		return nil, err
 	}
 
-	host, addr, err := net.SplitHostPort(addr)
+	host, port, err := net.SplitHostPort(addr)
 	if err != nil {
 		return nil, err
 	}
+	wsScheme := "wss"
+	wsHost := host
```
bradfitz, line 32: and this is just to make 80 into 443, right?

related: https://github.com/tailscale/tailscale/pull/4323

cc @maisem

bradfitz: nevermind, this is client_js.go

bradfitz (approved):

control/controlhttp/client_js.go
mihaip: True, we don't actually need to parse out `port`. Simplified further.

control/controlhttp/client_js.go
mihaip: I think client_js.go could be folded into client.go (as a third fallback to HTTP/HTTPS) if we wanted to do that bug and/or tailscale/corp#5542.

PR #4825, opened 2022-06-09
wgengine/magicsock: fix panic in http debug server

coadler:
Fixes an panic in `(*magicsock.Conn).ServeHTTPDebug` when the `recentPongs` ring buffer for an endpoint wraps around.


bradfitz (approved):

bradfitz:
> cc @bradfitz, this has sat for a while

whoops, thanks for the ping!

PR #4831, opened 2022-06-09
tsnet: do not log using log.Printf by default

dsnet:
Printing to log.Printf by default made sense when there was
no automatic uploading to log.tailscale.io.
However, now that we support that by default,
logging to log.Printf if Server.Logf is nil seems superfluous.

The tsnet package is also intended to be instantiated as a library.
Thus, a single process can multiple tsnet instances running simultaneously.
The default behavior of each logging to log.Printf makes the
os.Stderr  of the main process unreadable.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -11,7 +11,6 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


willnorris (approved):
This seems reasonable to me, but would want to make sure @bradfitz isn't aware of some other reason to keep this.

bradfitz:
Doesn't this mean it's hard for people to get their auth URL?

Or do we still print that to stderr?

bradfitz:
The doesn't answer my question? What's the user experience setting up a new node with this new default?

bradfitz:
As long as we still log.Printf the AuthURL when we're in needs-to-login state, I'm fine with this change.

bradfitz:
Ping @dsnet

dsnet:
I'm going to close this for now. I still prefer for logging to `log.Printf` *not* be the default, but I understand the need to expose the Auth URL somehow. I don't have any answers that's better than the status quo. We can revisit this if we come up with a better solution.

PR #4840, opened 2022-06-12
Portmapping: several commits

DentonGentry:
1. Log pcpCodeNotAuthorized more clearly. The user could get portmap support by enabling it in the router.
1. Add a test for the Sagemcom FAST3890V3. 
1. Send SSDP discovery for InternetGatewayDevice specifically, in addition to ssdp:all. Sagemcom devices only respond to IGD.

### DIFF
--- a/net/portmapper/portmapper.go
+++ b/net/portmapper/portmapper.go
@@ -749,9 +749,16 @@ func (c *Client) Probe(ctx context.Context) (res ProbeResult, err error) {
 		// See https://github.com/tailscale/tailscale/issues/3197 for
 		// an example of a device that strictly implements UPnP, and
 		// only responds to multicast queries.
+		//
+		// Then we send a discovery packet looking for
+		// urn:schemas-upnp-org:device:InternetGatewayDevice:1 specifically, not
+		// just ssdp:all, because there appear to be devices which only send
+		// their first descriptor (like urn:schemas-wifialliance-org:device:WFADevice:1)
+		// in response to ssdp:all. https://github.com/tailscale/tailscale/issues/3557
 		metricUPnPSent.Add(1)
 		uc.WriteTo(uPnPPacket, upnpAddr)
 		uc.WriteTo(uPnPPacket, upnpMulticastAddr)
+		uc.WriteTo(uPnPIGDPacket, upnpMulticastAddr)
 	}
 
 	buf := make([]byte, 1500)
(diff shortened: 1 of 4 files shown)


net/portmapper/portmapper.go
```diff
@@ -876,6 +883,11 @@ var uPnPPacket = []byte("M-SEARCH * HTTP/1.1\r\n" +
 	"ST: ssdp:all\r\n" +
 	"MAN: \"ssdp:discover\"\r\n" +
 	"MX: 2\r\n\r\n")
+var uPnPIGDPacket = []byte("M-SEARCH * HTTP/1.1\r\n" +
```
bradfitz, line 886: comment on this one, with reference to bug? (copy some of the comment from above)

net/portmapper/portmapper.go
```diff
@@ -876,6 +883,11 @@ var uPnPPacket = []byte("M-SEARCH * HTTP/1.1\r\n" +
...
+	"HOST: 239.255.255.250:1900\r\n" +
+	"ST: urn:schemas-upnp-org:device:InternetGatewayDevice:1\r\n" +
```
bradfitz, line 888: It's not enough to send one UDP packet with both of these ST values or something?

net/portmapper/portmapper.go
```diff
@@ -749,9 +749,16 @@ func (c *Client) Probe(ctx context.Context) (res ProbeResult, err error) {
...
+		//
+		// Then we send a discovery packet looking for
+		// urn:schemas-upnp-org:device:InternetGatewayDevice:1 specifically, not
+		// just ssdp:all, because there appear to be devices which only send
+		// their first descriptor (like urn:schemas-wifialliance-org:device:WFADevice:1)
+		// in response to ssdp:all. https://github.com/tailscale/tailscale/issues/3557
 		metricUPnPSent.Add(1)
 		uc.WriteTo(uPnPPacket, upnpAddr)
 		uc.WriteTo(uPnPPacket, upnpMulticastAddr)
+		uc.WriteTo(uPnPIGDPacket, upnpMulticastAddr)
```
bradfitz, line 761: Bit of a bummer that we're up to 5 packets now (PMP + PCP + 3 UPnP?) for portmapping discovery. We'll probably want a better scheduler for these at some point that's more aware of what's worked/not in the past. I forget how smart our current one is, but I suspect not great.

Oh well.

net/portmapper/upnp_test.go
```diff
@@ -35,6 +35,11 @@ const (
...
 )
 
+// Sagemcom FAST3890V3
```
bradfitz, line 38: issue reference too?

bradfitz (approved):

bradfitz:
Can you make every commit have an `Updates #nnn` or `Fixes #nnnn` where relevant?

net/portmapper/portmapper.go
DentonGentry: Added comment with reference to bug.

net/portmapper/portmapper.go
DentonGentry: We don't know. We could try it, but we also don't know whether multiple ST fields would break other routers.

The SAGEMCOM model known to not respond well to `ssdp:all` hasn't had any for sale on eBay recently.

net/portmapper/upnp_test.go
DentonGentry: Added bug link

bradfitz (approved):

net/portmapper/portmapper.go
DentonGentry: I think we can send SSDP way less often than we currently do, whether portmap protocols respond shouldn't change often. Maybe probe for all protocols a couple times shortly after startup.

PR #4844, opened 2022-06-14
Docker Entrypoint: `TAILSCALED_ARGS` -> `TS_TAILSCALED_ARGS` + allow parameter expansion of `TS_TAILSCALED_ARGS`

epk:
## Why

It would be nice to be able to pass on args to `tailscaled` without having to build my own image with a custom entrypoint. 
Another option here would be to have a new env var `TS_DAEMON_EXTRA_ARGS` or similar.

Open to other ideas


rodrigc:
Similar idea to what I had here: https://github.com/tailscale/tailscale/pull/4987

maisem:
Sorry I missed this one, #4987 just merged which adds similar functionality.

PR #4848, opened 2022-06-14
wgengine/router: set SkipAsSource on IPv6 LL addrs

raggi:
Link-local addresses on the Tailscale interface are not routable.
Ideally they would be removed, however, a standing concern exists that
the operating system will attempt to re-add them which would lead to
thrashing.

Setting SkipAsSource attempts to avoid production of packets using the
address as a source in any default behaviors. It should also result in
the address not being added to the DNS cache, so `ping (hostname)`
should no longer ping these addresses.

Issue #4647


wgengine/router/ifconfig_windows.go
```diff
@@ -574,26 +574,21 @@ func deltaNets(a, b []*net.IPNet) (add, del []*net.IPNet) {
...
+}
+
+func sliceSplit[T any](in []T, f func(T) bool) (a, b []T) {
+	a = in[:0]
+	b = in[:0]
+	for _, v := range in {
+		if f(v) {
+			a = append(a, v)
+		} else {
+			b = append(b, v)
```
bradfitz, line 588: this doesn't seem right. they're aliasing the same memory.


Also doesn't seem like the right package for it.

raggi:
> Put "Windows" somewhere in the git commit title somewhere (even at the end in parens/brackets is fine)
> 
> And is this `Fixes #4647` or `Updates #4647`? Pick one or the other.
> 
> Did you verify it does the name lookup suppression?

No, it's not tested yet.

wgengine/router/ifconfig_windows.go
bradfitz: e.g. https://go.dev/play/p/B3HiKEthA45

bradfitz (changes requested):

wgengine/router/ifconfig_windows.go
raggi: yeah, not surprised, copilot wrote this code

raggi:
> Put "Windows" somewhere in the git commit title somewhere (even at the end in parens/brackets is fine)
> 
> And is this `Fixes #4647` or `Updates #4647`? Pick one or the other.
> 
> Did you verify it does the name lookup suppression?

Verified now, PTAL.

wgengine/router/ifconfig_windows.go
```diff
@@ -622,28 +605,43 @@ func unicastIPNets(ifc *winipcfg.IPAdapterAddresses) []*net.IPNet {
...
+// to apply SkipAsSource to prevent them from being used, a the addresses will
+// not be routed on the Tailscale interface.
 func syncAddresses(ifc *winipcfg.IPAdapterAddresses, want []*net.IPNet) error {
-	var erracc error
+	var erracc []error
 
 	got := unicastIPNets(ifc)
 	add, del := deltaNets(got, want)
-	del = excludeIPv6LinkLocal(del)
+	ll, del := iter.Split(del, isIPv6LinkLocal)
```
maisem, line 617: Can you do `AppendMatching` and `FilterNotMatching` instead? That way you don't have to allocate two slices

maisem: You could also do the check in the del loop and not split the slices anyways.

util/iter/split.go
```diff
@@ -0,0 +1,18 @@
+// Package iter contains common utilities over iterable types.
+package iter
+
+// Split splits a slice into two slices with the left slice containing all
+// values for which f returns true, and the right containing all values for
+// which f returns false.
+func Split[T any](in []T, f func(T) bool) ([]T, []T) {
```
bradfitz, line 11: (trues, falses []T) ?

util/iter/split.go
```diff
@@ -0,0 +1,18 @@
+// Package iter contains common utilities over iterable types.
+package iter
+
+// Split splits a slice into two slices with the left slice containing all
+// values for which f returns true, and the right containing all values for
+// which f returns false.
```
bradfitz, line 10: mention aliasing, that the results don't alias in?

util/iter/split.go
```diff
@@ -0,0 +1,22 @@
...
+
+// Package iter contains common utilities over iterable types.
+package iter
+
+// Split splits a slice into two slices with the left slice containing all
+// values for which f returns true, and the right containing all values for
+// which f returns false.
+func Split[T any](in []T, f func(T) bool) ([]T, []T) {
+	a := make([]T, 0, len(in))
+	b := make([]T, 0, len(in))
```
bradfitz, line 13: doesn't seem worth allocating these aggressively. In common cases I can imagine all are false or true.

Or: allocate exactly one slice and fill it in from the front vs back and then reverse the back half at the end.

Do you guarantee ordering? document?

wgengine/router/ifconfig_windows.go
```diff
@@ -622,28 +604,51 @@ func unicastIPNets(ifc *winipcfg.IPAdapterAddresses) []*net.IPNet {
...
-			erracc = fmt.Errorf("deleting IP %q: %w", *a, err)
+			erracc = append(erracc, fmt.Errorf("deleting IP %q: %w", *a, err))
 		}
 	}
 
 	for _, a := range add {
 		err := ifc.LUID.AddIPAddress(*a)
 		if err != nil {
-			erracc = fmt.Errorf("adding IP %q: %w", *a, err)
+			erracc = append(erracc, fmt.Errorf("adding IP %q: %w", *a, err))
```
bradfitz, line 635: I'm a little concerned that these were bounded before and now we could have a ton of them in the multierror.

How about returning at most one per type (del, add, get-ll, set-ll)?

wgengine/router/ifconfig_windows.go
```diff
@@ -622,28 +604,51 @@ func unicastIPNets(ifc *winipcfg.IPAdapterAddresses) []*net.IPNet {
...
+		}
+	}
+
+	for _, a := range ll {
+		mib, err := ifc.LUID.IPAddress(a.IP)
+		if err != nil {
+			erracc = append(erracc, fmt.Errorf("setting skip-as-source on IP %q: unable to retrieve MIB: %w", *a, err))
+			continue
+		}
+		mib.SkipAsSource = true
```
bradfitz, line 645: and if they already have SkipAsSource set? can we avoid this Set call?

wgengine/router/ifconfig_windows.go
```diff
@@ -622,28 +604,51 @@ func unicastIPNets(ifc *winipcfg.IPAdapterAddresses) []*net.IPNet {
 // doing the minimum number of AddAddresses & DeleteAddress calls.
 // This avoids the full FlushAddresses.
 //
-// Any IPv6 link-local addresses are not deleted.
+// Any IPv6 link-local addresses are not deleted out of caution as some
+// configurations may repeatedly re-add them. Link-local addresses are adjusted
+// to apply SkipAsSource to prevent them from being used, a the addresses will
```
bradfitz, line 609: "a the"

wgengine/router/ifconfig_windows.go
raggi: I don't think we expect a lot of addresses on these interfaces, or a lot of errors.

If errors do come up, it seemed more useful for diagnosis to have more information, the code before was simply replacing the error, dropping all but the last.

The error accumulation is unrelated to the intent of the change, I was just trying to cleanup on the way past, but I'm not really clear on what is useful here - I don't know if one per type is more or less useful for diagnosis than just having the last error.

I think I'll continue to go with diff minimization so we can get the user issue addressed promptly. I've reverted the accumulation.

wgengine/router/ifconfig_windows.go
raggi: Done

wgengine/router/ifconfig_windows.go
bradfitz: Oh, I was thinking of O(n peers) but right, this is our self addresses. Nevermind.

bradfitz (approved):

PR #4858, opened 2022-06-15
util/deephash: generate type-specific hasher funcs

bradfitz:
```
name                old time/op    new time/op    delta
Hash-8                71.1µs ± 2%    71.5µs ± 1%     ~     (p=0.114 n=9+8)
HashPacketFilter-8    8.39µs ± 1%    4.83µs ± 2%  -42.38%  (p=0.000 n=8+9)
HashMapAcyclic-8      56.2µs ± 1%    56.9µs ± 2%   +1.17%  (p=0.035 n=10+9)
TailcfgNode-8         6.49µs ± 2%    3.54µs ± 1%  -45.37%  (p=0.000 n=9+9)
HashArray-8            729ns ± 2%     566ns ± 3%  -22.30%  (p=0.000 n=10+10)

name                old alloc/op   new alloc/op   delta
Hash-8                 24.0B ± 0%     24.0B ± 0%     ~     (all equal)
HashPacketFilter-8     24.0B ± 0%     24.0B ± 0%     ~     (all equal)
HashMapAcyclic-8       0.00B          0.00B          ~     (all equal)
TailcfgNode-8          0.00B          0.00B          ~     (all equal)
HashArray-8            0.00B          0.00B          ~     (all equal)

name                old allocs/op  new allocs/op  delta
Hash-8                  1.00 ± 0%      1.00 ± 0%     ~     (all equal)
HashPacketFilter-8      1.00 ± 0%      1.00 ± 0%     ~     (all equal)
HashMapAcyclic-8        0.00           0.00          ~     (all equal)
TailcfgNode-8           0.00           0.00          ~     (all equal)
HashArray-8             0.00           0.00          ~     (all equal)
```


util/deephash/deephash.go
```diff
@@ -161,26 +162,218 @@ func (h *hasher) hashUint64(i uint64) {
...
-	if !v.IsValid() {
-		return
+// returns ok if it was handled; else slow path runs
+type typeHasherFunc func(h *hasher, v reflect.Value) (ok bool)
+
+// A non-nil typeHasherFuture is populated into the typeHasher map
+// when its type is first requested, before its func is created.
+// Its func field fn is only populated once the type has been created.
+// This is used for recursive types.
+type typeHasherFuture struct {
```
dsnet, line 185: I think it would be simpler to solve the cyclic-types problem with lazy loading. This avoids this "partial" state problem you have going on here and subtle bugs that can occur as a result (see my comment in `hashValue`).

The lazy-loading approach uses a `sync.Once` that triggers upon the first call to `typeHasherFunc`, which performs the lookup lazily.

bradfitz:
Pushed more work. Now up to:

```
name                old time/op    new time/op    delta
HashPacketFilter-8    3.45µs ± 5%    2.36µs ± 8%  -31.53%  (p=0.000 n=9+9)

name                old alloc/op   new alloc/op   delta
HashPacketFilter-8     24.0B ± 0%     24.0B ± 0%     ~     (all equal)

name                old allocs/op  new allocs/op  delta
HashPacketFilter-8      1.00 ± 0%      1.00 ± 0%     ~     (all equal)
```

util/deephash/deephash.go
```diff
@@ -161,26 +162,218 @@ func (h *hasher) hashUint64(i uint64) {
 	binary.LittleEndian.PutUint64(h.scratch[:8], i)
 	h.bw.Write(h.scratch[:8])
 }
+func (h *hasher) hashVarint(x uint64) {
```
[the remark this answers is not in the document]
bradfitz: > I'm surprised this is faster than appending 8-byte uint64 every time.

I haven't measured this part in isolation. Maybe it's not. I was just lazy when hashing a signed integer.

> Are a vast majority of our integers close to zero or something?

Port numbers sure are.

util/deephash/deephash.go
```diff
@@ -161,26 +162,218 @@ func (h *hasher) hashUint64(i uint64) {
...
+			vLen := v.Len()
+			h.hashUint64(uint64(vLen))
+			for i := 0; i < vLen; i++ {
+				if !efu.fn(h, v.Index(i)) {
+					return false
+				}
+			}
+			return true
+		}
+	case reflect.Struct:
```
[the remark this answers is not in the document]
bradfitz: In progress. More in latest rev just pushed.

util/deephash/deephash.go
```diff
@@ -161,26 +162,218 @@ func (h *hasher) hashUint64(i uint64) {
...
+	}
+	return false // TODO: more
+}
+
+func (h *hasher) hashValue(v reflect.Value) {
+	if !v.IsValid() {
+		return
+	}
+	fn := getTypeHasher(v.Type())
+	if fn(h, v) {
```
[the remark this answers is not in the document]
bradfitz: oh, with concurrent callers. Yeah, I think I see.

util/deephash/deephash.go
dsnet: Right now you have:
```go
for i, n := 0, t.NumField(); i < n; i++ {
	...
	fields = append(fields, indexAndFuture{i, getTypeHasherFutureLocked(sf.Type)})
}
return func(h *hasher, v reflect.Value) bool {
	for _, f := range fields {
		...
	}
	return true
}
```
the partial state handling is to make sure the recursive call to `getTypeHasherFutureLocked` doesn't result in infinite recursive.

I'm proposing it become like:
```go
var once sync.Once
init := func() {
	for i, n := 0, t.NumField(); i < n; i++ {
		...
		fields = append(fields, indexAndFuture{i, getTypeHasherFutureLocked(sf.Type)})
	}
}
return func(h *hasher, v reflect.Value) bool {
	once.Do(init)
	for _, f := range fields {
		...
	}
	return true
}
```
Thus, we don't know have to worry about infinite recursion with `getTypeHasherFutureLocked`.

dsnet:
I wanted to confirm, is this ready for review again?

bradfitz:
> I wanted to confirm, is this ready for review again?

Not sure. Probably not. I rebased it over the weekend so it'd compile and pass tests again, but I haven't looked at the result of that rebase closely yet. It might still be a mess.

It's on my list to send to you properly this week.

util/deephash/deephash.go
```diff
@@ -190,11 +203,379 @@ type typeInfo struct {
...
+func (h *hasher) hashTimev(v reflect.Value) bool {
+	var t time.Time
+	if v.CanAddr() {
+		t = *(v.Addr().Interface().(*time.Time))
+	} else {
+		t = v.Interface().(time.Time) // allocs
+	}
+	b := t.AppendFormat(h.scratch[:0], time.RFC3339Nano)
+	h.bw.Write(b)
+	h.hashUint8(uint8(len(b))) // more than sufficient width; if not, good enough.
```
dsnet, line 513: Elsewhere in the package, we've been using a length-prefix, rather than a length-suffix. Perhaps do a length-prefix to be consistent?

```go
b := t.AppendFormat(h.scratch[:1], time.RFC3339Nano)
b[0] = len(b) - 1 // more than sufficient width; if not, good enough.
h.bw.Write(b)
```

util/deephash/deephash.go
```diff
@@ -190,11 +203,379 @@ type typeInfo struct {
...
+				return false
+			}
+		}
+		return true
+	}
+}
+
+func noopHasherFunc(h *hasher, v reflect.Value) bool { return true }
+
+func hashArray(t reflect.Type, eti *typeInfo) typeHasherFunc {
```
dsnet, line 552: Perhaps consistently name functions that return `typeHasherFunc` as `genHashXXX` or something?

dsnet (approved):

util/deephash/deephash.go
```diff
@@ -190,11 +203,379 @@ type typeInfo struct {
...
+				return eti.hasher()(h, v.Elem())
+			}
+		}
+	}
+
+	return func(h *hasher, v reflect.Value) bool {
+		if debug {
+			log.Printf("unhandled type %v", v.Type())
+		}
+		return false
```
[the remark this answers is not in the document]
bradfitz: Yup

util/deephash/deephash.go
```diff
@@ -190,11 +203,379 @@ type typeInfo struct {
...
+	return true
+}
+
+// hashString hashes v, of kind time.Time.
+func (h *hasher) hashTimev(v reflect.Value) bool {
+	var t time.Time
+	if v.CanAddr() {
+		t = *(v.Addr().Interface().(*time.Time))
+	} else {
+		t = v.Interface().(time.Time) // allocs
```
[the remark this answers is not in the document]
bradfitz: Yeah, I was thinking of the wrong direction. It's fine to get it back out. Removing the comment.

util/deephash/deephash.go
bradfitz: done

util/deephash/deephash.go
```diff
@@ -142,6 +143,34 @@ func Hash(v any) (s Sum) {
 	return h.sum()
 }
 
+// HasherForType is like Hash, but it returns a Hash func that's specialized for
+// the provided reflect type, avoiding a map lookup per value.
+func HasherForType(t reflect.Type) (hash func(any) (s Sum)) {
```
dsnet, line 148: Optional: Generics?
```go
func HasherForType[T any]() func(T) Sum
```

dsnet (approved):

util/deephash/deephash.go
bradfitz: Done.

PR #4863, opened 2022-06-15
net/dns/resolver: respond with SERVFAIL if all upstreams fail

### LINKED ISSUE #4722: Reverse DNS queries to 100.100.100.100 time out
### What is the issue?

I set up tailscale on a FreeBSD machine.  Suddenly, ssh into that machine over the public internet started taking forever to connect.  I tracked it down to sshd performing a reverse DNS lookup which timed out.  For now I've worked around it with `UseDNS no` in `sshd_config`, but I'm reporting the issue anyway.

It's not a big deal for me on Linux because split DNS works.  But on FreeBSD I just have `nameserver 100.100.100.100` in `/etc/resolve.conf` so regular system reverse DNS queries (e.g. `getnameinfo()`) also time out.

### Steps to reproduce

I can reproduce directly like this:

```console
$ time drill -x $MY_IP @100.100.100.100 # FreeBSD
Error: error sending query: Could not send or receive, because of network error
drill -x $MY_IP @100.100.100.100  0.00s user 0.01s system 0% cpu 15.092 total
$ time dig -x $MY_IP @100.100.100.100 # Linux

; <<>> DiG 9.18.2 <<>> -x $MY_IP @100.100.100.100
;; global options: +cmd
;; connection timed out; no servers could be reached

dig -x $MY_IP @100.100.100.100  0.00s user 0.01s system 0% cpu 15.030 total
```

_No response_

### OS

Linux, Other

### OS version

Arch Linux, FreeBSD 13.1-RELEASE

### Tailscale version

1.24.2

### Bug report

BUG-c4e4b6372b7c646441002f47d0b13e818dd3966cb7a216cda19e8b5308f7b517-20220520184002Z-6346e1494f0f52cf

twitchyliquid64:
Fixes #4722

This represents a fairly big change from being silent, so will need a lot of testing just in case there are side effects.

### DIFF
--- a/net/dns/resolver/tsdns.go
+++ b/net/dns/resolver/tsdns.go
@@ -278,7 +278,12 @@ func (r *Resolver) Query(ctx context.Context, bs []byte, from netaddr.IPPort) ([
 		defer cancel()
 		err = r.forwarder.forwardWithDestChan(ctx, packet{bs, from}, responses)
 		if err != nil {
-			return nil, err
+			select {
+			case resp := <-responses:
+				return resp.bs, err
+			default:
+				return nil, err
+			}
 		}
 		return (<-responses).bs, nil
 	}
--- a/net/dns/resolver/tsdns_test.go
+++ b/net/dns/resolver/tsdns_test.go
@@ -1417,3 +1417,41 @@ func TestUnARPA(t *testing.T) {
 		}
 	}
 }
+
+func TestServfail(t *testing.T) {
+	server := serveDNS(t, "127.0.0.1:0", "test.site.", miekdns.HandlerFunc(func(w miekdns.ResponseWriter, req *miekdns.Msg) {
+		m := new(miekdns.Msg)
+		m.Rcode = miekdns.RcodeServerFailure
+		w.WriteMsg(m)
+	}))
+	defer server.Shutdown()
+
+	r := newResolver(t)
+	defer r.Close()
+
+	cfg := dnsCfg
+	cfg.Routes = map[dnsname.FQDN][]*dnstype.Resolver{
+		".": {{Addr: server.PacketConn.LocalAddr().String()}},
+	}
+	r.SetConfig(cfg)
+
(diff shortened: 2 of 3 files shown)


twitchyliquid64:
Friendly ping :)

net/dns/resolver/tsdns_test.go
```diff
@@ -1417,3 +1417,41 @@ func TestUnARPA(t *testing.T) {
 		}
 	}
 }
+
+func TestServfail(t *testing.T) {
```
bradfitz, line 1421: Can you add a little comment above this test saying what the overall goal of the test is, what it's testing? Maybe a bug link if applicable?

net/dns/resolver/tsdns.go
```diff
@@ -278,7 +278,12 @@ func (r *Resolver) Query(ctx context.Context, bs []byte, from netaddr.IPPort) ([
 		defer cancel()
 		err = r.forwarder.forwardWithDestChan(ctx, packet{bs, from}, responses)
 		if err != nil {
-			return nil, err
+			select {
```
bradfitz, line 281: // Best effort: use any error sent by forwardWithDestChan.
// This is present in some errors paths, such as when all upstream
// DNS servers replied with an error.

bradfitz (approved):

twitchyliquid64:
I'm going back on forth on whether we should respond SERVFAIL for any upstream error as long as they all errored, or only respond SERVFAIL if upstreams responded SERVFAIL (`if firstErr == errServerFailure {`). Wdyt @bradfitz?

DentonGentry:
We may be building 1.28 next week. If this is to be in it, we should get it into an unstable as it is a relatively large change in behavior.

twitchyliquid64:
WILCO.

> I'm going back on forth on whether we should respond SERVFAIL for any upstream error as long as they all errored, or only respond SERVFAIL if upstreams responded SERVFAIL (if firstErr == errServerFailure {).

Settling on only responding with SERVFAIL if an upstream did for now.

PR #4864, opened 2022-06-16
controlclient: preserve NetInfo on Hostinfo update

crawshaw:
We have two separate functions, SetHostinfo and SetNetInfo, called by
different systems at different times.

But we store the NetInfo object inside the Hostinfo object.
Usually this is fine if we:

	1. SetHostinfo
	2. SetNetInfo

But it is possible to:

	1. SetNetInfo  -- controlclient now knows about our home DERP
	2. SetHostinfo -- NetInfo is now nil

The result is controlclient loses track of our PreferredDERP.
In our end-to-end test, this appears as the client being connected to a
test DERP server, but the server never receiving the information.

This is visible in the logs of the end-to-end test by searching for the
string "[v1] HostInfo" and seeing that even though a DERP home is well
established before the log line, there is no NetInfo to speak of.

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -242,7 +242,11 @@ func (c *Direct) SetHostinfo(hi *tailcfg.Hostinfo) bool {
 	if hi.Equal(c.hostinfo) {
 		return false
 	}
-	c.hostinfo = hi.Clone()
+	hi = hi.Clone()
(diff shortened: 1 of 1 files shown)


crawshaw:
Needs a test before I submit. I haven't come up with a good one yet, all this stuff is very racy.

crawshaw:
doesn't work

PR #4869, opened 2022-06-16
util/deephash: fix map hashing to actually hash elements

### LINKED ISSUE #4868: util/deephash: map hashing is broken
The map hash code tries to has each map item's key & element, but accidentally hashes its element as `v` (the map itself) instead of `e` (the element):

https://github.com/tailscale/tailscale/blob/4005134263db39b5a9e0f13f6db0e260f871a5c7/util/deephash/deephash.go#L345

Found when I removed the cycle detection overhead for non-recursive types and then one of the tests stack overflowed and crashed in this code.... whoops.

This means we've never been hashing maps correctly.

Worse, we only hash one random element and then the cycle detection stops us (thinking it found a recursive map), and the hash output isn't even stable between runs of anything containing maps.

Like.... *tailcfg.DNSConfig used by wgengine:

```
	routerChanged := deephash.Update(&e.lastRouterSig, routerCfg, dnsCfg)
```

Which has:

```
// Config is a DNS configuration.
type Config struct {
...
	Routes map[dnsname.FQDN][]*dnstype.Resolver
...
	Hosts map[dnsname.FQDN][]netaddr.IP
}
```

😬

bradfitz:
Fixes #4868

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -342,7 +342,7 @@ func (h *hasher) hashMap(v reflect.Value) {
 		e.SetIterValue(iter)
 		mh.h.reset()
 		mh.h.hashValue(k)
-		mh.h.hashValue(v)
+		mh.h.hashValue(e)
 		sum.xor(mh.h.sum())
 	}
 	h.bw.Write(append(h.scratch[:0], sum.sum[:]...)) // append into scratch to avoid heap allocation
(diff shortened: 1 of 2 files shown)


dsnet:
Given how simple of a test exposes the problem, I'm shocked we had no map tests beforehand

andrew-d (approved):

raggi (approved):

PR #4870, opened 2022-06-16
util/deephash: don't track cycles on non-recursive types

bradfitz:
```
Hash-8              67.3µs ±20%    76.5µs ±16%     ~     (p=0.143 n=10+10)
HashMapAcyclic-8    63.0µs ± 2%    56.3µs ± 1%  -10.65%  (p=0.000 n=10+8)
TailcfgNode-8       9.18µs ± 2%    6.52µs ± 3%  -28.96%  (p=0.000 n=9+10)
HashArray-8          732ns ± 3%     709ns ± 1%   -3.21%  (p=0.000 n=10+10)

Hash-8               24.0B ± 0%     24.0B ± 0%     ~     (all equal)
HashMapAcyclic-8     0.00B          0.00B          ~     (all equal)
TailcfgNode-8        0.00B          0.00B          ~     (all equal)
HashArray-8          0.00B          0.00B          ~     (all equal)

Hash-8                1.00 ± 0%      1.00 ± 0%     ~     (all equal)
HashMapAcyclic-8      0.00           0.00          ~     (all equal)
TailcfgNode-8         0.00           0.00          ~     (all equal)
HashArray-8           0.00           0.00          ~     (all equal)
```

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -125,7 +125,7 @@ func Hash(v any) (s Sum) {
 		seed = uint64(time.Now().UnixNano())
 	})
 	h.hashUint64(seed)
-	h.hashValue(reflect.ValueOf(v))
+	h.hashValue(reflect.ValueOf(v), false)
 	return h.sum()
 }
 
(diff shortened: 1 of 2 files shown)


util/deephash/deephash.go
```diff
@@ -164,26 +164,144 @@ func (h *hasher) hashUint64(i uint64) {
...
+			reflect.Uint16,
+			reflect.Uint32,
+			reflect.Uint64,
+			reflect.Uintptr,
+			reflect.Float32,
+			reflect.Float64,
+			reflect.Complex64,
+			reflect.Complex128,
+			reflect.String,
+			reflect.Func:
```
dsnet, line 241: Do we care about Chan, UnsafePointer?

util/deephash/deephash.go
```diff
@@ -164,26 +164,144 @@ func (h *hasher) hashUint64(i uint64) {
...
+
+		switch t.Kind() {
+		default:
+			panic("unhandled kind " + t.Kind().String())
+		case reflect.Interface:
+			// Assume the worst for now. TODO(bradfitz): in some cases
+			// we should be able to prove that it's not recursive. Not worth
+			// it for now.
+			return true
+		case reflect.Array, reflect.Chan, reflect.Pointer, reflect.Slice:
```
dsnet, line 265: Do we really need to visit a Chan?

util/deephash/deephash.go
```diff
@@ -164,26 +164,144 @@ func (h *hasher) hashUint64(i uint64) {
...
+			}
+			if visitType(t.Elem()) {
+				return true
+			}
+		case reflect.Struct:
+			if t.String() == "intern.Value" {
+				// Otherwise its interface{} makes this return true.
+				return false
+			}
+			if t.Size() == 0 {
```
dsnet, line 279: Redundant: Already checked above on L244.

util/deephash/deephash.go
```diff
@@ -164,26 +164,144 @@ func (h *hasher) hashUint64(i uint64) {
...
+		return false
+	}
+	return visitType(t)
+}
+
+func (h *hasher) hashValue(v reflect.Value, forceCycleChecking bool) {
+	if !v.IsValid() {
+		return
 	}
+	ti := getTypeInfo(v.Type())
```
dsnet, line 298: There's a race here with concurrent hashes of the same type where `ti.eleTypeInfo` and `ti.keyTypeInfo` are `nil` when first stored into the map, and populated asynchronously.

Perhaps we should do:
```go
type typeInfo struct {
	done uint32
	wait chan struct{} // closed when done is atomically set to 1

	...
}

func getTypeInfoLocked(t reflect.Type) *typeInfo {
	...
	ti := &typeInfo{wait: make(chan struct{}), ...}
	typeInfoMap.Store(t, ti)
	defer func() {
		close(ti.wait)
		atomic.StoreUint32(&ti.done, 1)
	}
	...
}


func (h *hasher) hashValue(v reflect.Value, forceCycleChecking bool) {
	...
	ti := getTypeInfo(v.Type())
	if atomic.LoadUint32(&ti.done) == 0 {
		<-ti.wait
	}
	...
}
```

dsnet: Semantically, we could rely on `<-ti.wait` alone, but apparently an atomic load is 7x faster for the common case.

dsnet (changes requested):

util/deephash/deephash.go
bradfitz: Done.

util/deephash/deephash.go
bradfitz: Yes:

```
	type RecursiveChan chan *RecursiveChan
```

I added two new tests:

```
		{val: make(RecursiveChan), want: true},
		{val: make(chan int), want: false},
```

bradfitz: Admittedly we won't be able to hash them anyway, but I'd rather this func be as correct as possible.

dsnet: I guess there is technically a cycle in the type-graph, but not a cycle in the value-graph (for hashing concerns) since we will never try to hash a channel.

SGTM.

util/deephash/deephash.go
bradfitz: PTAL

util/deephash/deephash.go
```diff
@@ -164,26 +165,165 @@ func (h *hasher) hashUint64(i uint64) {
...
+		ti.keyTypeInfo = getTypeInfoLocked(t.Key(), incomplete)
+		fallthrough
+	case reflect.Ptr, reflect.Slice, reflect.Array:
+		ti.elemTypeInfo = getTypeInfoLocked(t.Elem(), incomplete)
+	}
+
+	close(ti.completeCh)
+	atomic.StoreInt32(&ti.atomicComplete, 1)
+
+	ti.keyTypeInfo.waitComplete()
```
dsnet, line 225: Perhaps move:
```go
ti.keyTypeInfo.waitComplete()
ti.elemTypeInfo.waitComplete()
```
to the `typeInfo.waitComplete` method?

util/deephash/deephash.go
```diff
@@ -164,26 +165,165 @@ func (h *hasher) hashUint64(i uint64) {
...
+var typeInfoMapPopulate sync.Mutex // just for adding to typeInfoMap
+
+func getTypeInfo(t reflect.Type) *typeInfo {
+	if f, ok := typeInfoMap.Load(t); ok {
+		return f.(*typeInfo)
 	}
+	typeInfoMapPopulate.Lock()
+	defer typeInfoMapPopulate.Unlock()
+	ti := getTypeInfoLocked(t, map[reflect.Type]*typeInfo{})
+	ti.waitComplete()
```
dsnet, line 195: Given the use of an `incomplete` map, I'm not sure we need the synchronization guaranteed to us by `atomicComplete` and `completeCh` anymore since this all  runs under a global `typeInfoMapPopulate` mutex. The `typeInfo` is guaranteed to be complete by the time `getTypeInfoLocked` returns since there are no goroutines spawned in that call-tree.

dsnet: Also, there's a minor waste of memory where the `typeInfo` of sub-types can be repeated computed.

Let type A reference types B and C.

Suppose I concurrently call `getTypeInfo` on A and C:
* `getTypeInfo(A)` will only cause A to be stored into `typeInfoMap`, even though it did the work to compute B and C.
* `getTypeInfo(C)` will now redo the work to compute C since it couldn't find C in `typeInfoMap`.

I think this can be solved by doing:
```go
func getTypeInfo(t reflect.Type) *typeInfo {
	...
	typeInfoMapPopulate.Lock()
	defer typeInfoMapPopulate.Unlock()
	newTypes := map[reflect.Type]*typeInfo{}
	ti := getTypeInfoLocked(t, newTypes)
	for t, ti := range newTypes {
		typeInfoMap.Store(t, ti)
	}
	return ti
}
```

dsnet (approved):
It looks correct, but seems more complicated than necessary and uses more memory than necessary.

util/deephash/deephash.go
dsnet: Disregard. I think I've come to the conclusion that `waitComplete` is unnecessary given that `incomplete` map approach you're taking.

util/deephash/deephash.go
bradfitz: > Also, there's a minor waste of memory where the typeInfo of sub-types can be repeatedly computed.

Ah, whoops!

> I think this can be solved by doing:

Yeah, LGTM, thanks! Done.

dsnet (approved):
Yay!

PR #4884, opened 2022-06-19
control/controlclient: start simplifying netmap fetch APIs

bradfitz:
Step 1 of many, cleaning up the direct/auto client & restarting map
requests that leads to all the unnecessary map requests.

Updates tailscale/corp#5761

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -620,18 +620,27 @@ func inTest() bool { return flag.Lookup("test.v") != nil }
 
 // PollNetMap makes a /map request to download the network map, calling cb with
 // each new netmap.
-//
-// maxPolls is how many network maps to download; common values are 1
-// or -1 (to keep a long-poll query open to the server).
-func (c *Direct) PollNetMap(ctx context.Context, maxPolls int, cb func(*netmap.NetworkMap)) error {
-	return c.sendMapRequest(ctx, maxPolls, cb)
+func (c *Direct) PollNetMap(ctx context.Context, cb func(*netmap.NetworkMap)) error {
+	return c.sendMapRequest(ctx, -1, false, cb)
+}
+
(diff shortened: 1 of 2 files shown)


control/controlclient/direct.go
```diff
@@ -703,6 +712,16 @@ func (c *Direct) sendMapRequest(ctx context.Context, maxPolls int, cb func(*netm
...
+
+		// On initial startup before we know our endpoints, set the ReadOnly flag
+		// to tell the control server not to distribute out our (empty) endpoints to peers.
+		// Presumably we'll learn our endpoints in a half second and do another post
+		// with useful results. The first POST just gets us the DERP map which we
+		// need to do the STUN queries to discover our endpoints.
+		// TODO(bradfitz): we skip this optimization in tests, though,
+		// because the e2e tests are currently hyperspecific about the
+		// ordering of things. The e2e tests need love.
+		ReadOnly: readOnly || (len(epStrs) == 0 && !everEndpoints && !inTest()),
```
crawshaw, line 724: Something I observed in testing is endpoints being sent to the control server before a DERP connection had been established. In fact, the entire NetInfo block of Hostinfo was nil.

I don't know how common this is in reality, but maybe we should also set ReadOnly if `hi.NetInfo == nil`?

crawshaw (approved):
Are you testing changes against the controlclient tests in the other repository? Better to catch the bugs before submitting here rather than on bump oss.

control/controlclient/direct.go
bradfitz: This change has no change in behavior. This is just refactoring.

The plan is for PollNetMap to always be read-only and stream forever without ever being interrupted and any changes to netmap/etc will happen in lite endpoint update requests that don't abort the long poll.

PR #4889, opened 2022-06-20
adding --noconfirm to pacman installation

SofianeHamlaoui:
As you can see, the automatic installation won't work because you need to confirm the installation manually on arch-based distributions.

So you need to add `--noconfirm` to pacman installation command to bypass the confirmations steps. ( same as "-y" in Debian )

And now it'll just work perfectly !


DentonGentry (approved):
Thank you!

PR #4893, opened 2022-06-20
cmd/gitops-pusher: add new GitOps assistant

Xe:
This is for an upcoming blogpost on how to manage Tailscale ACLs using a
GitOps flow. This tool is intended to be used in CI and will allow users
to have a git repository be the ultimate source of truth for their ACL
file. This enables ACL changes to be proposed, approved and discussed
before they are applied.


cmd/gitops-pusher/main.go
```diff
@@ -0,0 +1,209 @@
+package main
```
bradfitz, line 6: rename from `main.go` to `gitops-pusher.go`, add copyright, godoc package doc,

cmd/gitops-pusher/main.go
```diff
@@ -0,0 +1,209 @@
...
+
+	req.SetBasicAuth(apiKey, "")
+	req.Header.Set("Content-Type", "application/hujson")
+	req.Header.Set("If-Match", oldEtag)
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+
```
bradfitz, line 128: defer resp.Body.Close() here and elsewhere

cmd/gitops-pusher/main.go
```diff
@@ -0,0 +1,214 @@
...
+	req.SetBasicAuth(apiKey, "")
+	req.Header.Set("Accept", "application/hujson")
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", err
+	}
+
+	if resp.StatusCode != http.StatusOK {
```
bradfitz, line 209:
```if got, want := resp.StatusCode, http.StatusOK; got != want {
    .... , got, want)
```

... to not repeat the got & want values.

cmd/gitops-pusher/README.md
```diff
@@ -0,0 +1,48 @@
...
+      - name: Install gitops-pusher
+        run: go install tailscale.com/cmd/gitops-pusher@latest
+              
+      - name: Deploy ACL
+        if: github.event_name == 'push'
+        env:
+          TS_API_KEY: ${{ secrets.TS_API_KEY }}
+          TS_TAILNET: ${{ secrets.TS_TAILNET }}
+        run: |
+          ~/go/bin/gitops-pusher --policy-file ./policy.hujson apply
```
bradfitz, line 35: how about just "go run ..."

But that's pretty heavy in a GitHub Action to pull in the `tailscale.com` module which is huge.

This should probably live somewhere smaller.

Xe: :+1: I'll go figure out somewhere to put it, but this should work until then.

cmd/gitops-pusher/gitops-pusher.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	defer fin.Close()
+
+	h := sha256.New()
+	_, err = io.Copy(h, fin)
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("\"%x\"", h.Sum(nil)), nil
```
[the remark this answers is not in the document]
Xe: With the Go hash type, you collect data into the Hash value and then call `Sum` when you want to get the final checksum. This is slightly weird as written because the ETag header value is also quoted. I did it like this to avoid having to shuck or parse the string.

cmd/gitops-pusher/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: ACL tests
+        if: github.event_name == 'pull_request'
```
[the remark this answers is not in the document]
Xe: Yes

PR #4899, opened 2022-06-21
cmd/tailscale: make up respect explicitly empty --operator= value

Soypete:
Fixes: #3808 

change the implicit fields checks on the `--operator` flag to allow it to be explicitly set to an empty string

Signed-off-by: soypete <miriah@tailscale.com>

### DIFF
--- a/cmd/tailscale/cli/cli_test.go
+++ b/cmd/tailscale/cli/cli_test.go
@@ -493,14 +493,16 @@ func TestCheckForAccidentalSettingReverts(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
-			applyImplicitPrefs(newPrefs, tt.curPrefs, tt.curUser)
-			var got string
-			if err := checkForAccidentalSettingReverts(newPrefs, tt.curPrefs, upCheckEnv{
+			upEnv := upCheckEnv{
 				goos:          goos,
 				flagSet:       flagSet,
 				curExitNodeIP: tt.curExitNodeIP,
 				distro:        tt.distro,
-			}); err != nil {
+				user:          tt.curUser,
+			}
+			applyImplicitPrefs(newPrefs, tt.curPrefs, upEnv)
+			var got string
+			if err := checkForAccidentalSettingReverts(newPrefs, tt.curPrefs, upEnv); err != nil {
 				got = err.Error()
 			}
 			if strings.TrimSpace(got) != tt.want {
@@ -784,6 +786,10 @@ func TestUpdatePrefs(t *testing.T) {
 		curPrefs *ipn.Prefs
 		env      upCheckEnv // empty goos means "linux"
 
+		// checkUpdatePrefsMutations, if non-nil, is run with the new prefs after
+		// updatePrefs might've mutated them (from applyImplicitPrefs).
+		checkUpdatePrefsMutations func(t *testing.T, newPrefs *ipn.Prefs)
+
 		wantSimpleUp   bool
 		wantJustEditMP *ipn.MaskedPrefs
 		wantErrSubtr   string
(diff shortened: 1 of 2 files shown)


bradfitz:
Test failures look real.

cmd/tailscale/cli/cli_test.go
```diff
@@ -489,18 +490,20 @@ func TestCheckForAccidentalSettingReverts(t *testing.T) {
 			flagSet := newUpFlagSet(goos, &upArgs)
 			flags := CleanUpArgs(tt.flags)
 			flagSet.Parse(flags)
+			env := upCheckEnv{
```
bradfitz, line 493: move this down a bit to just before `applyImplicitPrefs`. As written this makes it look like `prefsFromUpArgs` needs this.

cmd/tailscale/cli/up.go
```diff
@@ -881,13 +881,22 @@ func checkForAccidentalSettingReverts(newPrefs, curPrefs *ipn.Prefs, env upCheck
...
+// applyImplicitPrefs mutates prefs to add implicit preferences for the user operator.
+// If the operator flag is passed no action is taken, otherwise this only needs to be set if it doesn't
 // match the current user.
-//
-// curUser is os.Getenv("USER"). It's pulled out for testability.
-func applyImplicitPrefs(prefs, oldPrefs *ipn.Prefs, curUser string) {
-	if prefs.OperatorUser == "" && oldPrefs.OperatorUser == curUser {
+func applyImplicitPrefs(prefs, oldPrefs *ipn.Prefs, env upCheckEnv) {
+	explicitlySet := false
+	upFlagSet.Visit(func(f *flag.Flag) {
```
bradfitz, line 889: rather than the global, use `env.flagSet`?

cmd/tailscale/cli/up.go
```diff
@@ -881,13 +881,22 @@ func checkForAccidentalSettingReverts(newPrefs, curPrefs *ipn.Prefs, env upCheck
...
+		if f.Name == "operator" {
+			explicitlySet = true
+		}
+	})
+
+	if explicitlySet && prefs.OperatorUser == "" {
```
bradfitz, line 895: merge this check into line 899, adding a `&& !explicitOperator`?

then we can add future implicit prefs without this short-circuiting the rest of the func.

(rename `explicitlySet` to `explicitOperator` too, to be clear which thing's explicit)

cmd/tailscale/cli/up.go
Soypete: I am learning all about flag sets today

cmd/tailscale/cli/up.go
Soypete: happy to make the change. Is this for better readability or performance?

cmd/tailscale/cli/cli_test.go
```diff
@@ -847,6 +850,26 @@ func TestUpdatePrefs(t *testing.T) {
 				WantRunningSet:            true,
 			},
 		},
+		{
```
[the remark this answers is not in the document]
Soypete: oh thanks... I must have missed one. My intention was to pull all the love test code and them remove the erroneous `env` changes

bradfitz:
Looking good. Can you squash this down to one commit now in our usual commit message format? http://go/commit

Notably, start with a verb, and say `Fixes #nnn` on its own line.

Like:  `cmd/tailscale: make up respect explicitly empty --operator= value`

bradfitz:
It's down to 2 commits now. Can you make it one, including the Fixes line?

cmd/tailscale/cli/up.go
bradfitz: This reverted.

cmd/tailscale/cli/up.go
bradfitz: This reverted.

bradfitz (approved):

Issue #4883, opened 2022-06-19
util/deephash: add a full fuzz test


josharian:
### What is the issue?

Following up on #4871, here's one written for go-fuzz:

```go
//go:build gofuzz

package deephash

import (
        "encoding/json"
        "fmt"
        "reflect"
)

func Fuzz(data []byte) int {
        if len(data) == 0 {
                return -1
        }
        cut := int(data[0])
        data = data[1:]
        if cut > len(data) {
                return -1
        }
        buf1, buf2 := data[:cut], data[cut:]
        var x, y any
        err := json.Unmarshal(buf1, &x)
        if err != nil {
                return -1
        }
        err = json.Unmarshal(buf2, &y)
        if err != nil {
                return -1
        }
        if (Hash(x) == Hash(y)) != reflect.DeepEqual(x, y) {
                fmt.Printf("Hash(x) = %v\n", Hash(x))
                fmt.Printf("Hash(y) = %v\n", Hash(y))
                fmt.Printf("DeepEqual? %v\n", reflect.DeepEqual(x, y))
                fmt.Printf("x (%T): %+v\n", x, x)
                fmt.Printf("y (%T): %+v\n", y, y)
                panic("mismatch")
        }
        return 0
}
```

This ~immediately caught two issues:

```go
func TestFloatEdgeCases(t *testing.T) {
       f1 := 0.0
       f2 := -f1
       if h1, h2 := Hash(f1), Hash(f2); (h1 == h2) != (f1 == f2) {
               t.Errorf("bogus: f1=%#v, f2=%#v", f1, f2)
       }
}
```

And another (which I didn't turn into a test):

```
Hash(x) = 9975ddef9205194a611087d342abd4d17886dcdf1e8a9ecd390a6604c8ab976c
Hash(y) = 9975ddef9205194a611087d342abd4d17886dcdf1e8a9ecd390a6604c8ab976c
DeepEqual? false
x ([]interface {}): []
y (float64): 0
panic: mismatch
```

There might be others lurking after those two are fixed.

cc @dsnet @bradfitz

bradfitz:
`TestFloatEdgeCases` is a problem with the test. They hash differently, but zero and negative zero are true for `==`: https://go.dev/play/p/o4X_uIJkrwL

josharian:
[...]

PR #4901, opened 2022-06-22
util/deephash: fix collisions between different types

bradfitz:
cc @josharian 

Updates #4883


maisem (approved):

josharian (approved):
After this commit, I'm at 20 minutes of fuzzing with no failures, which is a very promising sign.

josharian:
After approximately an hour of fuzzing, the only bug I've found is a bug in go-fuzz (whee): https://github.com/dvyukov/go-fuzz/issues/341

PR #4915, opened 2022-06-23
fix(docs): pass TS_ROUTES correctly as make kwargs

beejiujitsu:
# Problem

`make subnet-router` does not pick up `TS_ROUTES` set as shell env var.

# Solution

Update docs to reflect intended `make subnet-router` effects.

# Result

Able to deploy subnet-router with shell-specified subnet routes.


beejiujitsu:
Is the tailscale github particpation dead?

PR #4932, opened 2022-06-25
/usr/bin/login on Mac requires different argument ordering

### LINKED ISSUE #4931: Tailscale SSH to MacOS 12.4 (Monterey) doesn't work
### What is the issue?

I enabled Tailscale SSH on my Mac, however when I access it from my Linux workstation, I get an error:

```
$ tailscale ssh xxx@yyy
Last login: Sat Jun 25 18:30:18 on ttys003
login: --h: No such file or directory
Connection to yyy.zzz.beta.tailscale.net. closed.
```

### Steps to reproduce

Followed these instructions to install Tailscale on my Mac:
https://github.com/tailscale/tailscale/wiki/Tailscaled-on-macOS

Started Tailscaled with:

```
sudo TAILSCALE_USE_WIP_CODE=1 tailscaled
```

Setup ACLs, then connected to it from a Linux host.

Yes, the new Tailscale SSH feature. We couldn't have this bug without it!

### OS

macOS

### OS version

MacOS 12.4 (Monterey)

### Tailscale version

Current `main` - 1.27.0+dev?

### Bug report

_No response_

aeijdenberg:
Fixes #4931 - well seems to - with this change I am able to `ssh` in successfully.

I don't know if other versions of MacOS would be different, and I'm not sure the best way to make this conditional in the code, but thought I'd contribute this as a step in the direction of solving.

### DIFF
--- a/ssh/tailssh/incubator_darwin.go
+++ b/ssh/tailssh/incubator_darwin.go
@@ -4,6 +4,18 @@
 
 package tailssh
 
+import "syscall"
+
 func (ia *incubatorArgs) loginArgs() []string {
 	return []string{ia.loginCmdPath, "-fp", "-h", ia.remoteIP, ia.localUser}
+}
+
+func setGroups(groupIDs []int) error {
+	// darwin returns "invalid argument" if more than 16 groups are passed to syscall.Setgroups
+	// some info can be found here:
+	// https://opensource.apple.com/source/samba/samba-187.8/patches/support-darwin-initgroups-syscall.auto.html
+	// this fix isn't great, as anyone reading this has probably just wasted hours figuring out why
+	// some permissions thing isn't working, due to some arbitrary group ordering, but it at least allows
+	// this to work for more things than it previously did.
+	return syscall.Setgroups(groupIDs[:16])
 }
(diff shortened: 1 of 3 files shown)


ssh/tailssh/incubator_darwin.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build darwin
+// +build darwin
```
bradfitz, line 6: FYI these lines aren't necessary when the filename already ends in a common thing like `_darwin.go`.

bradfitz (commented):
Thanks! Looks good other than minor things.

Can you make your commit message match our existing style?

(`git log` to see, but it's basically https://golang.org/wiki/CommitMessage)

Like:

```
ssh/tailssh: fix /usr/bin/login args on macOS

Fixes #nnn
```

bradfitz:
In the future try to open separate PRs for unrelated changes rather than adding new commits to one that's already been reviewed.

This is good enough for now, but the second one doesn't really "Fixes" the issue as noted in the commit message. So I'll merge this and then re-open that bug.

Thanks for the changes!

PR #4936, opened 2022-06-25
util/deephash: add packet filter benchmark

bradfitz:
(breaking up parts of another change)

This adds a PacketFilter hashing benchmark with an input that both
contains every possible field, but also is somewhat representative in
the shape of what real packet filters contain.

### DIFF
--- a/util/deephash/deephash_test.go
+++ b/util/deephash/deephash_test.go
@@ -324,6 +324,57 @@ func BenchmarkHash(b *testing.B) {
 	}
 }
 
+func ptrTo[T any](v T) *T { return &v }
+
+// filterRules is a packet filter that has both everything populated (in its
+// first element) and also a few entries that are the typical shape for regular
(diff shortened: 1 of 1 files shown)


util/deephash/deephash_test.go
```diff
@@ -324,6 +324,55 @@ func BenchmarkHash(b *testing.B) {
 	}
 }
 
+func ptrTo[T any](v T) *T { return &v }
+
+// filterRules is a 1-element FilterRule with everything populated.
```
maisem, line 329: 1-element?

maisem (approved):
Results?

util/deephash/deephash_test.go
bradfitz: Updated comment to more match the commit message.

bradfitz:
> Results?

Not interesting at this point. I just want this in for the upcoming optimization to compare against.

PR #4937, opened 2022-06-25
util/deephash: add canMemHash func + typeInfo property

bradfitz:
Currently unused. (breaking up a bigger change)


crawshaw (approved):
Looks right.

Issue #4939, opened 2022-06-26
ssh/tailssh: do setgroups on darwin properly


bradfitz:
See #4938 from @aeijdenberg.

There's a darwin-specific setgroups dance. See comments in https://opensource.apple.com/source/samba/samba-187.8/patches/support-darwin-initgroups-syscall.auto.html

https://github.com/tailscale/tailscale/commit/9294a14a376e44d68087af40c97918431073276a is a workaround but not the ideal fix.

aeijdenberg:
Thanks @bradfitz , I'm still having a poke around - even with those other bits, I still don't have my use-case quite working.

e.g. I want the Mac to say things - this works:

```bash
$ ssh xxx@yyy
Last login: Sun Jun 26 15:25:32 on tty??
xxx@Adams-iMac ~ % say hi
# works, computer says "hi" out loud
```

This doesn't work:

```bash
$ ssh xxx@yyy say hi
Open speech channel failed: -915
exit status 1
```

Interestingly if I run `tailscaled` with `sudo tailscaled` (vs letting it start via Launchd), then frustratingly it does work again. There's clearly some cleverness re Mac login processes we need to overcome.

I'm going to have a play to see if running the command as: `/usr/bin/login -fpq xxx say hi`  will work, `man login` on Mac shows a 2nd form:

```
     login [-pq] [-h hostname] [user]
     login -f [-lpq] [-h hostname] [user [prog [args...]]]
```

(this is a fun weekend project - I can't pretend the use-case is important: https://github.com/continusec/timedcomputershutdown#timedcomputershutdown)

DentonGentry:
Was this fixed by https://github.com/tailscale/tailscale/commit/0582829e00271a63c3abcb9e12700d9b223d386d ?

Barring further comment, we'll expect to close this during the next regular bug scrub.

DentonGentry:
Fixed by https://github.com/tailscale/tailscale/commit/0582829e00271a63c3abcb9e12700d9b223d386d

PR #4941, opened 2022-06-26
ssh/tailssh: try launching commands with /usr/bin/login on macOS

aeijdenberg:
Perhaps fixes some aspects of #4939 by avoiding that portion of code.
 
I really don't know if this is a good solution or not - I'm not that familiar with the internals of what happens when someone tries to login to a system. I just noticed that `/usr/bin/login` on Mac had the ability to accept program name and args, and noted this happened to fix the issue I was seeing (details in #4939).

Signed-off-by: Adam Eijdenberg <adam@continusec.com>

### DIFF
--- a/ssh/tailssh/incubator_darwin.go
+++ b/ssh/tailssh/incubator_darwin.go
@@ -4,7 +4,17 @@
 
 package tailssh
 
-import "syscall"
+import (
+	"os"
+	"syscall"
+
+	"golang.org/x/sys/unix"
+	"tailscale.com/types/logger"
+)
+
+func init() {
+	maybeRunCommandWithLogin = maybeRunCommandWithLoginDarwin
+}
 
 func (ia *incubatorArgs) loginArgs() []string {
 	return []string{ia.loginCmdPath, "-fp", "-h", ia.remoteIP, ia.localUser}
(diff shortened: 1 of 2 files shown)


ssh/tailssh/incubator_darwin.go
```diff
@@ -19,3 +29,15 @@ func setGroups(groupIDs []int) error {
...
 	return syscall.Setgroups(groupIDs[:16])
 }
+
+// maybeRunCommandWithLoginDarwin is the darwin implementation of maybeRunCommandWithLogin
+func maybeRunCommandWithLoginDarwin(logf logger.Logf, ia incubatorArgs) (bool, error) {
+	if ia.isSFTP || ia.isShell || ia.cmdName == "" || ia.loginCmdPath == "" {
+		return false, nil
+	}
+
+	// if we run it this way on Mac, this seems to solve various user ID issues
```
bradfitz, line 39: This comment makes sense to you today but will make sense to ~nobody in a month or two.

Either say more or include a URL to more info.

ssh/tailssh/incubator_darwin.go
```diff
@@ -19,3 +29,15 @@ func setGroups(groupIDs []int) error {
...
+	return true, unix.Exec(ia.loginCmdPath, append(
```
bradfitz, line 40: unix.Exec return value, eh? So what's the point of the bool result parameter? If it returns true, it's always an error?

bradfitz (commented):
(Leaving for @maisem to review)

maisem:
@aeijdenberg I tested this and it works fine 🎉  , can you address @bradfitz's comments?

ssh/tailssh/incubator_darwin.go
aeijdenberg: If it doesn't handle it (ie if not on Mac), then it returns false. I put that comment where the variable if is defined in `incubator.go`. You're technically right, since `unix.Exec()` is special, and I guess won't will return `nil`, I guess we could special case the method signature to mean: "return nil if unable to handle this"?

Felt simpler to me to return 2 values. Can change if you feel strongly.

I used `unix.Exec()` rather than `exec.Command()` as that matched the other invocation of `/usr/bin/login` from `incubator.go`.  It also means that the process should get killed if the connection drops.

ssh/tailssh/incubator_darwin.go
aeijdenberg: Ack, done.

bradfitz:
@maisem, can you take this over?

PR #4956, opened 2022-06-28
tsnet: add Server.AuthKey field

bradfitz:
... so callers can provide the AuthKey via mechanisms other than
environment variables which means multiple Servers can't be started
concurrently in the same process without coordination.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -74,6 +74,13 @@ type Server struct {
 	// as an Ephemeral node (https://tailscale.com/kb/1111/ephemeral-nodes/).
 	Ephemeral bool
 
+	// AuthKey, if non-empty, is the auth key to create the node
+	// and will be preferred over the TS_AUTHKEY environment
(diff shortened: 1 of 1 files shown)


danderson (approved):

tsnet/tsnet.go
```diff
@@ -306,7 +320,7 @@ func (s *Server) start() error {
 		logf("LocalBackend state is %v; running StartLoginInteractive...", st)
 		s.lb.StartLoginInteractive()
 	} else if authKey != "" {
-		logf("TS_AUTHKEY is set; but state is %v. Ignoring authkey. Re-run with TSNET_FORCE_LOGIN=1 to force use of authkey.", st)
+		logf("Authkey is set; but state is %v. Ignoring authkey. Re-run with TSNET_FORCE_LOGIN=1 to force use of authkey.", st)
```
bradfitz, line 323: Fixed this misleading error message too.

Issue #4924, opened 2022-06-24
"Authentication Failure" when running SSHing into an Arch/Manjaro box


mike-hearn:
### What is the issue?

When trying to use Tailscale's SSH functionality into an Arch machine (in my case from MacOS -> Arch), it returns the error `Authentication Failure`

This issue was first flagged on the forum, and there's some additional information there - https://forum.tailscale.com/t/tailscale-ssh-authentication-failure-on-manjaro-arch/2564

### Steps to reproduce

1. Install Tailscale + SSH on an Arch box
2. Add that box to the Tailscale network, as well as whatever client machine you are using
3. SSH into the Arch box
4. Observe the failure message
5. Hold back your tears

### Are there any recent changes that introduced the issue?

It's the new beta SSH feature, so bugs are expected! I still love you, Tailscale.

### OS

Linux, macOS

### OS version

5.17.0-rc7-asahi-next-20220310-5-2-ARCH

### Tailscale version

1.26.1

### Bug report

BUG-b8f072d4dc1c38925c280914f94579d937dd0af19aacde6633fc20c7d5e0eed4-20220624193515Z-5ef9537512f1ef2f

mike-hearn:
Here is some additional logging provided by sundbp on the Tailscale forum:

```Jun 22 23:16:04 thehost tailscaled[2623]: 100.78.255.16:53068->theuser@100.81.41.35:22: starting session: 20220622T221603-4880af8916
Jun 22 23:16:04 thehost tailscaled[2623]: ssh-session(20220622T221603-4880af8916): handling new SSH connection from theuser@gmail.com (100.78.255.16) to ssh-user "theuser"
Jun 22 23:16:04 thehost tailscaled[2623]: ssh-session(20220622T221603-4880af8916): access granted to theuser@gmail.com as ssh-user "theuser"
Jun 22 23:16:04 thehost tailscaled[2623]: ssh-session(20220622T221603-4880af8916): starting pty command: [/usr/bin/tailscaled be-child ssh --uid=1000 --gid=1000 --groups=1000,998,995,993,991,988,987,986,3,90,9>
Jun 22 23:16:04 thehost login[3240032]: pam_warn(remote:account): function=[pam_sm_acct_mgmt] flags=0 service=[remote] terminal=[/dev/pts/30] user=[theuser] ruser=[<unknown>] rhost=[100.78.255.16]
Jun 22 23:16:04 thehost login[3240032]: Authentication failure
```

[...]

PR #4963, opened 2022-06-28
ssh/tailssh: fix Tailscale SSH to Linux Arch machines

bradfitz:
See https://github.com/tailscale/tailscale/issues/4924#issuecomment-1168201823

Arch uses a different login binary that makes the -h flag set the PAM
service to "remote". So if they don't have that configured, don't pass -h.

Thanks to @eddiezane for debugging!

Updates #4924


maisem (approved):

eddiezane (approved):
I think this is reasonable.

Thanks for the quick fix!

Also :eyes: on that distro package.

Issue #3102, opened 2021-10-15
macOS: Networking errors occur within Chrome when sufficient concurrent network activity exists


shorthandguy:
### What is the issue?

Whenever using TailScale (App version: 1.14.6) on macOS (Big Sur 11.6), when I generate sufficient concurrent network activity within Chrome (issue may not be browser-specific, but the tool that makes it easiest to reproduce this issue is one that works best in Chrome), then on almost every occasion, Chrome will generate a variety of networking-related errors. 

The activity that is primarily happening when the issue occurs is concurrent uploads of media to remote servers (via Chrome). 

I have spent too many hours attempting to perform Root Cause Analysis into this issue, however, nothing that has been tried has been fruitful. This has included:
- Working with the administrator to enable and disable override DNS,
- Working with the administrator to enable and disable MagicDNS
- Using the same DNS servers as what is configured within TailScale
- Using alternative DNS servers 
- Disabling IPv6 on the local network (the ISP generates IPv6 addresses, but these addresses appear to be cycled), but disabling IPv6 on the network and only using IPv4 did not solve the issue
- Upgrading to the latest MacOS client
- Upgrading Chrome to the latest version (v94.0.4606.81)
- Flushing sockets and DNS cache within Chrome and the DNS cache within macOS
- Disabling all extensions within Chrome
- Using Incognito mode in Chrome
- Using a standalone instance of Chrome
- Looked through Net Logs of Chrome and Console logs of macOS
- And various other things..

Our TailScale network is running the latest server software, uses publicly accessible common DNS servers, and uses exit nodes.

Whilst it would be easy to possibly point to either the web application, the tool or the browser itself, the _only_ variable that is consistent is that the error _only_ occurs when connected to TailScale (I can't select to not use 'Use TailScale DNS Server', otherwise my Internet connectivity ceases to work). 

I have tested this issue with TailScale disabled, plus also tested it whilst connected to both NordVPN and VPN Unlimited, and the issue never happens in those cases. 

The primary error that I encounter in Chrome is 'ERR_NETWORK_CHANGED', but the main other error has been 'ERR_NAME_NOT_RESOLVED', 

About the only time these errors _haven't_ occurred has been when i've slowed down the media upload process sufficiently that it didn't end up with as much data in-flight. 

[...]

PR #4965, opened 2022-06-28
wgengine/router: avoid unncessary routing configuration changes

mihaip:
The iOS and macOS networking extension API only exposes a single setter
for the entire routing and DNS configuration, and does not appear to
do any kind of diffing or deltas when applying changes. This results
in spurious "network changed" errors in Chrome, even when the
`OneCGNATRoute` flag from df9ce972c79023e0b0535bffee6afb3d88e61dc3 is
used (because we're setting the same configuration repeatedly).

Since we already keep track of the current routing and DNS configuration
in CallbackRouter, use that to detect if they're actually changing, and
only invoke the platform setter if it's actually necessary.

Updates #3102

### DIFF
--- a/wgengine/router/router.go
+++ b/wgengine/router/router.go
@@ -7,6 +7,8 @@
 package router
 
 import (
+	"reflect"
+
 	"golang.zx2c4.com/wireguard/tun"
 	"inet.af/netaddr"
 	"tailscale.com/types/logger"
@@ -70,6 +72,16 @@ type Config struct {
 	SubnetRoutes     []netaddr.IPPrefix     // subnets being advertised to other Tailscale nodes
 	SNATSubnetRoutes bool                   // SNAT traffic to local subnets
 	NetfilterMode    preftype.NetfilterMode // how much to manage netfilter rules
+}
+
+func (a *Config) Equal(b *Config) bool {
+	if a == nil && b == nil {
+		return true
+	}
+	if (a == nil) != (b == nil) {
+		return false
+	}
+	return reflect.DeepEqual(a, b)
 }
 
 // shutdownConfig is a routing configuration that removes all router
--- a/wgengine/router/callback.go
+++ b/wgengine/router/callback.go
@@ -13,7 +13,7 @@ import (
 // CallbackRouter is an implementation of both Router and dns.OSConfigurator.
 // When either network or DNS settings are changed, SetBoth is called with both configs.
 // Mainly used as a shim for OSes that want to set both network and
-// DNS configuration simultaneously (iOS, android).
+// DNS configuration simultaneously (Mac, iOS, Android).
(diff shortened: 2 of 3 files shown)


wgengine/router/router.go
```diff
@@ -72,6 +72,49 @@ type Config struct {
...
+	}
+	if a.SNATSubnetRoutes != b.SNATSubnetRoutes {
+		return false
+	}
+	if a.NetfilterMode != b.NetfilterMode {
+		return false
+	}
+
+	for i := range a.LocalAddrs {
+		if a.LocalAddrs[i] != b.LocalAddrs[i] {
```
mihaip, line 96: Wasn't sure if the "can compare with ==" comment in netaddr is only true for `IP`, or if we can also use it for `IPPrefix`

wgengine/router/router.go
```diff
@@ -72,6 +72,49 @@ type Config struct {
 	NetfilterMode    preftype.NetfilterMode // how much to manage netfilter rules
 }
 
+func (a *Config) Equal(b *Config) bool {
```
bradfitz, line 75: it's probably adding a test like tailcfg's TestHostinfoEqual here, so a test fails if somebody adds a new field to Config and forgets to update this method.

At least just this part of the test:

```go
func TestHostinfoEqual(t *testing.T) {
	hiHandles := []string{
		"IPNVersion", "FrontendLogID", "BackendLogID",
		"OS", "OSVersion", "Desktop", "Package", "DeviceModel", "Hostname",
		"ShieldsUp", "ShareeNode",
		"GoArch",
		"RoutableIPs", "RequestTags",
		"Services", "NetInfo", "SSH_HostKeys",
	}
	if have := fieldsOf(reflect.TypeOf(Hostinfo{})); !reflect.DeepEqual(have, hiHandles) {
		t.Errorf("Hostinfo.Equal check might be out of sync\nfields: %q\nhandled: %q\n",
			have, hiHandles)
	}
```

bradfitz (approved):

wgengine/router/router.go
bradfitz: also IPPrefix and IPPort and IPRange. I guess docs could be more clear.

wgengine/router/router.go
mihaip: Would it also make sense to use `reflect.DeepEqual` in the implementation of `Equal` for `Config` (that's what `Hostinfo` does)?

bradfitz: it's not as fast and can allocate and nil vs empty slices are different. So if those don't matter then that's fine.

mihaip: This is not a hotspot as far as I can tell, so I switched to it. Added tests too.

bradfitz (approved):

PR #4970, opened 2022-06-29
Dockerfile: bump alpine to 3.16

maisem:


Dockerfile
```diff
@@ -56,5 +56,5 @@ RUN GOARCH=$TARGETARCH go install -ldflags="\
       -X tailscale.com/version.GitCommit=$VERSION_GIT_HASH" \
       -v ./cmd/tailscale ./cmd/tailscaled
 
-FROM ghcr.io/tailscale/alpine-base:3.14
```
bradfitz, line 59: no latest please. We want these to be reproducible.

bradfitz (approved):

PR #4972, opened 2022-06-29
Docker: add ALPINE.txt to manage alpine versions

maisem:


danderson (commented):
Why? This adds a test that 2 numbers in different files are the same. Does one of the numbers get updated automatically somehow, or do we end up with 2 numbers to maintain manually instead of one?

maisem:
We were already maintaining the same number in multiple files (see https://github.com/tailscale/corp/pull/5936). Happy to do it any other way if you have suggestions?

The goal here is to 
1. make it so that the number doesn't diverge
2. not define the number in corp, only in oss

There is no great way (that I know of) to automatically ingest a file to figure what the base image should be during the docker build phase.

danderson (approved):

PR #4976, opened 2022-06-29
ipn/ipnlocal: prefer to use one CGNAT route on the Mac

mihaip:
Together with 06aa1416325424ab2881a9e2442697f9707f363b this minimizes
the number of NEPacketTunnelNetworkSettings updates that we have to do,
and thus avoids Chrome interrupting outstanding requests due to
(perceived) network changes.

Updates #3102


ipn/ipnlocal/local.go
```diff
@@ -2232,6 +2232,49 @@ func (b *LocalBackend) authReconfig() {
...
+		if v, ok := nm.Debug.OneCGNATRoute.Get(); ok {
+			return v
+		}
+	}
+	// Also prefer to do this on the Mac, so that we don't need to constantly
+	// update the network extension configuration (which is distruptive to
+	// Chrome, see https://github.com/tailscale/tailscale/issues/3102). Only
+	// use fine-grained routes if another interfaces is also using the CGNAT
+	// IP range.
+	if versionOS == "macOS" {
```
bradfitz, line 2252: Maybe?

switch versionOS {
case "macOS":

}

Do you think we'll be adding iOS here soon too? Or others?

net/interfaces/interfaces.go
```diff
@@ -474,7 +474,7 @@ func hasTailscaleIP(pfxs []netaddr.IPPrefix) bool {
 	return false
 }
 
-func isTailscaleInterface(name string, ips []netaddr.IPPrefix) bool {
+func IsTailscaleInterface(name string, ips []netaddr.IPPrefix) bool {
```
bradfitz, line 477: exported so needs docs now

ipn/ipnlocal/local.go
```diff
@@ -2232,6 +2232,49 @@ func (b *LocalBackend) authReconfig() {
 	b.initPeerAPIListener()
 }
 
+// shouldUseOneCGNATRoute returns whether we should prefer to make one big
```
bradfitz, line 2235: "reports whether" (sorry)

I can still hear Rob Pike saying "The function returns regardless."

bradfitz (approved):

net/interfaces/interfaces.go
mihaip: No longer need to export this now that there's a `HasCGNATInterface` helper method.

ipn/ipnlocal/local.go
```diff
@@ -2232,6 +2232,39 @@ func (b *LocalBackend) authReconfig() {
...
+// shouldUseOneCGNATRoute reports whether we should prefer to make one big
+// CGNAT /10 route rather than a /32 per peer.
+//
+// The versionOS is a Tailscale-style version ("iOS", "macOS") and not
+// a runtime.GOOS.
+func shouldUseOneCGNATRoute(nm *netmap.NetworkMap, logf logger.Logf, versionOS string) bool {
+	// Explicit enabling or disabling always take precedence.
+	if nm.Debug != nil {
+		if v, ok := nm.Debug.OneCGNATRoute.Get(); ok {
+			return v
```
bradfitz, line 2244: do you want to log here too? maybe have the caller log instead, rather than duplicating 2258-2263?

bradfitz: oh, you only want to log on macOS, I see. maybe a new func then.

bradfitz (approved):

Issue #4926, opened 2022-06-24
ThreadSanitizer failures in windows-race


dblohm7:
### What is the issue?

We've got CI failing in some tests in windows-race. In each of these cases, the test program cannot even start.

I did find [this Go issue](https://github.com/golang/go/issues/46099) discussing the problem, but it looked like nobody really understood the reason behind the `VirtualAlloc` failure.

The base address that TSan wants to use for allocating shadow memory is invalid because it is higher than the maximum virtual address permitted by the kernel for user-mode VM allocation requests.

If I'm reading the existing analysis so far, the concern is around the computation of the base address [here](https://github.com/golang/go/blame/fa6aa872225f8d33a90d936e7a81b64d2cea68e1/src/runtime/race.go#L366=).

### OS

Windows

raggi:
If I disable ASLR on the binary, it works, e.g.

```
C:\Users\raggi\src\tailscale>editbin /HIGHENTROPYVA:NO tailscaled.test.exe
Microsoft (R) COFF/PE Editor Version 14.26.28805.0
Copyright (C) Microsoft Corporation.  All rights reserved.


C:\Users\raggi\src\tailscale>editbin /DYNAMICBASE:NO tailscaled.test.exe
Microsoft (R) COFF/PE Editor Version 14.26.28805.0
Copyright (C) Microsoft Corporation.  All rights reserved.


C:\Users\raggi\src\tailscale>tailscaled.test.exe
PASS
```

DentonGentry:
https://github.com/tailscale/tailscale/pull/4977 solved the immediate failures by downgrading MinGW. I'm not sure what else we can do here.

mihaip:
#5359 removed windows-race altogether, so I think this is obsolete.

PR #4977, opened 2022-06-29
.github/workflows: downgrade MinGW to work around windows-race failures

mihaip:
Updates #4926


raggi:
LGTM, we could also pass `-aslr=false` in the build arguments for the same effect.

raggi (approved):

dblohm7 (approved):
ty!

PR #4979, opened 2022-06-29
ipn/ipnlocal, net/dns*, util/cloudenv: specialize DNS config on Google Cloud

### LINKED ISSUE #4911: dns: support Google Cloud (GCP) DNS forwarding by default
Hacker News [comment](https://news.ycombinator.com/item?id=31844996):

> Consistently I’m unable to use Tailscale on a GCP instance and also use GCP services cleanly, because it messes with the DNS route to the metadata server. Otherwise, it’s a great product.

We should detect when we're running on GCP (https://pkg.go.dev/cloud.google.com/go/compute/metadata) and automatically add its service resolver hierarchy as a fallback resolver split DNS route.

So even if the distro being run on GCP doesn't have systemd-resolved or something reasonable, we can still do the expected thing.

/cc @danderson @twitchyliquid64

bradfitz:
This does three things:

* If you're on GCP, it adds a *.internal DNS split route to the
  metadata server, so we never break GCP DNS names. This lets people
  have some Tailscale nodes on GCP and some not (e.g. laptops at home)
  without having to add a Tailnet-wide *.internal DNS route.
  If you already have such a route, though, it won't overwrite it.

* If the 100.100.100.100 DNS forwarder has nowhere to forward to,
  it forwards it to the GCP metadata IP, which forwards to 8.8.8.8.
  This means there are never errNoUpstreams ("upstream nameservers not set")
  errors on GCP due to e.g. mangled /etc/resolv.conf (GCP default VMs
  don't have systemd-resolved, so it's likely a DNS supremacy fight)

* makes the DNS fallback mechanism use the GCP metadata IP as a
  fallback before our hosted HTTP-based fallbacks

I created a default GCP VM from their web wizard. It has no
systemd-resolved.

I then made its /etc/resolv.conf be empty and deleted its GCP
hostnames in /etc/hosts.

I then logged in to a tailnet with no global DNS settings.

With this, tailscaled writes /etc/resolv.conf (direct mode, as no
systemd-resolved) and sets it to 100.100.100.100, which then has
regular DNS via the metadata IP and *.internal DNS via the metadata IP
as well. If the tailnet configures explicit DNS servers, those are used
instead, except for *.internal.

This also adds a new util/cloudenv package based on version/distro
where the cloud type is only detected once. We'll likely expand it in
the future for other clouds, doing variants of this change for other
popular cloud environments.

Fixes #4911

RELNOTES=Google Cloud DNS improvements


ipn/ipnlocal/local.go
```diff
@@ -2237,7 +2238,7 @@ func (b *LocalBackend) authReconfig() {
 //
 // The versionOS is a Tailscale-style version ("iOS", "macOS") and not
 // a runtime.GOOS.
-func dnsConfigForNetmap(nm *netmap.NetworkMap, prefs *ipn.Prefs, logf logger.Logf, versionOS string) *dns.Config {
+func dnsConfigForNetmap(nm *netmap.NetworkMap, prefs *ipn.Prefs, logf logger.Logf, versionOS string, cloud cloudenv.Cloud) *dns.Config {
```
danderson, line 2241: Docstring needs an update, it currently lists all the things that influence the resulting config. Or we could make it not say specifics, since the list of things that influence the config is getting fairly long.

ipn/ipnlocal/local.go
```diff
@@ -2325,10 +2326,14 @@ func dnsConfigForNetmap(nm *netmap.NetworkMap, prefs *ipn.Prefs, logf logger.Log
 			dcfg.Routes[dom] = nil // resolve internally with dcfg.Hosts
 		}
 	}
+	if cloud == cloudenv.GCP {
```
danderson, line 2329: Maybe a comment explaining the intent? Without knowing the background, I'd expect the preexisting VM DNS config to cover this, so why do we need to care? I think we want this half so that "override local DNS" keeps working, and half to work around broken base configs, right?

ipn/ipnlocal/local.go
```diff
@@ -2381,6 +2386,7 @@ func dnsConfigForNetmap(nm *netmap.NetworkMap, prefs *ipn.Prefs, logf logger.Log
 		// settings or we break all DNS resolution.
 		//
 		// https://github.com/tailscale/tailscale/issues/1713
+
```
danderson, line 2389: bonus space?

danderson (approved):

twitchyliquid64 (approved):

bradfitz:
PTAL. Added comments and made the DNS fallback mechanism also use GCP.

net/dnscache/dnscache.go
```diff
@@ -107,6 +108,29 @@ func (r *Resolver) fwd() *net.Resolver {
...
 }
 
+// cloudHostResolver returns a Resolver for the current cloud hosting environment.
+// It currently only supports Google Cloud.
+func (r *Resolver) cloudHostResolver() (v *net.Resolver, ok bool) {
+	switch runtime.GOOS {
+	case "android", "ios", "darwin":
+		return nil, false
+	case "windows":
+		// TODO(bradfitz): remove this restriction once we're using Go 1.19
```
maisem, line 118: at some point it would be nice to have a searchable TODO for Go bumps.

bradfitz:
Now with more comments for @maisem.

maisem (approved):

danderson (approved):

PR #4982, opened 2022-06-30
ipn/ipnlocal, net/dns*, util/cloudenv: add AWS DNS support

### LINKED ISSUE #4980: dns: add AWS DNS support
Like #4911/#4979 did for GCP, we could also add AWS support:

https://docs.aws.amazon.com/vpc/latest/userguide/vpc-dns.html

> The Amazon DNS server does not reside within a specific subnet or Availability Zone in a VPC. It's located at the address 169.254.169.253 (and the reserved IP address at the base of the VPC IPv4 network range, plus two) and fd00:ec2::253. For example, the Amazon DNS Server on a 10.0.0.0/16 network is located at 10.0.0.2. For VPCs with multiple IPv4 CIDR blocks, the DNS server IP address is located in the primary CIDR block.

https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resolver.html

> When you create a VPC using Amazon VPC, Route 53 Resolver automatically uses a Resolver on the VPC to answer DNS queries for local Amazon VPC domain names for EC2 instances (ec2-192-0-2-44.compute-1.amazonaws.com) and records in private hosted zones (acme.example.com). For all other domain names, Resolver performs recursive lookups against public name servers.
>
> When you create a VPC, the Route 53 Resolver that is created by default, maps to a DNS server that runs on a reserved IP address for the VPC network range, plus 2. For example, the DNS Server on a 10.0.0.0/16 network is located at 10.0.0.2.

/cc @maisem

### LINKED ISSUE #4981: net/dns: rework how GCP DNS fallback for *.internal works
As I went to do the AWS variant, I realized the fix in #4979 for #4911 for GCP is probably not ideal.

The GCP VM I was testing on (the default from their instance creation wizard) doesn't have systemd-resolved, so I was using Tailscale's DNS direct mode for testing.

But my AWS vanilla VM does have systemd-resolved, and I realize we don't want to register an explicit `*.internal` split DNS route with systemd-resolved if one already exists.

The change in #4979 on a machine with systemd-resolved will add a new *.internal route to 100.100.100.100 (which will then proxy to the right place) rather than just hitting the right place directly (well, from systemd-resolved).

I think instead we just want to modify the forwarder code to pretend it had a *.internal route there, but never actually tell the DNS Manager & OSConfigurator about it.

Thoughts, @danderson?

bradfitz:
And remove the GCP special-casing from ipn/ipnlocal; do it only in the
forwarder for *.internal.

Fixes #4980
Fixes #4981


net/dns/resolver/forwarder.go
```diff
@@ -303,12 +313,34 @@ func resolversWithDelays(resolvers []*dnstype.Resolver) []resolverAndDelay {
 // The memory referenced by routesBySuffix should not be modified.
 func (f *forwarder) setRoutes(routesBySuffix map[dnsname.FQDN][]*dnstype.Resolver) {
 	routes := make([]route, 0, len(routesBySuffix))
+
+	var cloudResolver []*dnstype.Resolver
+	var cloudHostFallback []resolverAndDelay
+	if ip := cloudenv.Get().ResolverIP(); ip != "" {
+		cloudResolver = []*dnstype.Resolver{{Addr: ip}}
+		cloudHostFallback = resolversWithDelays(cloudResolver)
```
danderson, line 321: Do sync.Once and keep it around? The cloud env isn't going to change over time, so it can be done once per process.

net/dns/resolver/forwarder.go
```diff
@@ -303,12 +313,34 @@ func resolversWithDelays(resolvers []*dnstype.Resolver) []resolverAndDelay {
...
+		}
 		routes = append(routes, route{
 			Suffix:    suffix,
 			Resolvers: resolversWithDelays(rs),
 		})
 	}
+
+	switch cloudenv.Get() {
+	case cloudenv.GCP, cloudenv.AWS:
+		_, haveInternal := routesBySuffix["internal."]
```
danderson, line 335: Both use .internal for stuff? TIL

net/dnscache/dnscache.go
```diff
@@ -119,14 +119,21 @@ func (r *Resolver) cloudHostResolver() (v *net.Resolver, ok bool) {
 		// which supports net.Resolver.PreferGo on Windows.
 		return nil, false
 	}
-	if cloudenv.Get() != cloudenv.GCP {
+	var ip string
+	switch cloudenv.Get() {
+	case cloudenv.GCP:
+		ip = cloudenv.GoogleMetadataAndDNSIP
```
danderson, line 125: Use cloudenv.Get().ResolverIP() ? Saves you a bunch of switching and stuff.

danderson (approved):
Sure. To the concerns I voiced in #4982, I think a lot of the extra "inject stuff" logic could be lifted into at least manager.go, so that the existing DNS config logs reflect our actual behavior fully, but this is fine and I can play with a refactor separately once this is in.

net/dns/resolver/forwarder.go
bradfitz: Yeah, I also didn't know that.

net/dnscache/dnscache.go
bradfitz: Right. That was a late addition. Will do.

net/dns/resolver/forwarder.go
bradfitz: done

bradfitz:
For now I at least made it print out the cloud-specific DNS tweaks in this line:

```
dns: Resolvercfg: {Routes:{.:[]} Hosts:7 LocalDomains:[ts.net. bradxxx.tailscale.net.]+65arpa, cloud="gcp"}
```

Issue #4984, opened 2022-06-30
util/cloudenv: improve AWS detection


bradfitz:
#4980 adds DNS-specific tweaks when running on AWS (e.g. bootstrap DNS, not shadowing *.internal) but only works if we can detect that we're running on AWS.

Our current implementation of that detection is super minimal:

```go
func getCloud() Cloud {
	// TODO(bradfitz): also detect AWS on Windows, etc. Just try to hit the metadata server
	// and see if it's there? But it might be turned off. Do some small-timeout DNS request
	// to 169.254.169.253 and see if it replies? But which DNS request?
	if runtime.GOOS == "linux" {
		biosVendorB, _ := os.ReadFile("/sys/class/dmi/id/bios_vendor")
		biosVendor := strings.TrimSpace(string(biosVendorB))
		if biosVendor == "Amazon EC2" || strings.HasSuffix(biosVendor, ".amazon") {
			return AWS
		}
	}
....
```

We should do something that works for BSDs & Windows on AWS too.

See that TODO above.

Issue #4983, opened 2022-06-30
net/dns: add Azure DNS support


bradfitz:
Like GCP (#4911) and AWS (#4980), also do Azure-specific DNS stuff.

Notes:
https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-name-resolution-for-vms-and-role-instances

> Access to the recursive resolvers in Azure is provided via the virtual IP 168.63.129.16.

And looks like the suffix is `.internal.cloudapp.net` ?

DentonGentry:
I think https://github.com/tailscale/tailscale/pull/4986 implemented this.

Barring further comment, we'll expect to close this during the next regular bug scrub.

DentonGentry:
Fixed by https://github.com/tailscale/tailscale/pull/4986

PR #4986, opened 2022-06-30
util/cloudenv: add Azure support & DNS IPs

bradfitz:
And rewrite cloud detection to try to do only zero or one metadata
discovery request for all clouds, only doing a first (or second) as
confidence increases. Work remains for Windows, but a start.

And add Cloud to tailcfg.Hostinfo, which helped with testing using
"tailcfg debug hostinfo".

Updates #4983 (Linux only)
Updates #4984


util/cloudenv/cloudenv.go
```diff
@@ -66,20 +81,88 @@ func Get() Cloud {
...
+	tr := &http.Transport{
+		DisableKeepAlives: true,
+		Dial: (&net.Dialer{
+			Timeout: maxWait,
+		}).Dial,
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), maxWait)
+	defer cancel()
+
+	req, _ := http.NewRequestWithContext(ctx, "HEAD", "http://"+CommonNonRoutableMetadataIP, nil)
```
DentonGentry, line 134: I think this turns into a IMDSv1 request, a GET (HEAD, in this case) to the local Metadata server. IMDSv1 is disabled in many AWS deployments now and, more problematically, may have CloudWatch configured to watch for IMDSv1 requests.

bradfitz: AWS IMDSv1 and AWS IMDSv2 and GCP metadata and Azure metadata are all four just HTTP to this same IP address.

To use IMDSv2 you need to make an HTTP request like this without any token to get a token. (`curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60"`)

The CloudWatch `MetadataNoToken` metric is mentioned at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/viewing_metrics_with_cloudwatch.html but doesn't say what exactly it counts. I confidently assume it doesn't count requests to get a token. I less confidently assume it shouldn't count no-op requests like a `HEAD /` that aren't trying to do anything, but who knows.

I could try to get a token instead, but that might just fire a CloudWatch-like counter on Azure or GCP instead.

I suppose I could experiment with CloudWatch and see if it triggers with this.

Looking at their SDK,

https://docs.aws.amazon.com/sdk-for-go/api/aws/ec2metadata/#EC2Metadata.Available
```
// Available returns if the application has access to the EC2 Metadata service.
// Can be used to determine if application is running within an EC2 Instance and
// the metadata service is available.
func (c *EC2Metadata) Available() bool {
```

.... it fetches a token (which would've been enough to answer the question of whether EC2Metadata was available) and then proceeds to fetch the instance-id using that token.

So we could fetch a token I guess?

But look a few lines up: on Linux, we don't even make it this far if we already found from the /sys info that we're on EC2. So this path would only trigger for Windows in the future (currently Windows returns false for all clouds)

bradfitz: Well, ....

```
$ for x in $(seq 0 500); do curl  -I http://169.254.169.254/; done
....
```
Screen Shot 2022-06-30 at 8 50 06 AM

That answers that question.

bradfitz: But no spike from:

```
ubuntu@bradprod:~$ for x in $(seq 0 500); do curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 5" ; done
```

Screen Shot 2022-06-30 at 9 23 55 AM

bradfitz: @DentonGentry, PTAL.

util/cloudenv/cloudenv.go
```diff
@@ -66,20 +82,102 @@ func Get() Cloud {
...
+		Dial: (&net.Dialer{
+			Timeout: maxWait,
+		}).Dial,
 	}
-	if gcpmetadata.OnGCE() {
+	ctx, cancel := context.WithTimeout(context.Background(), maxWait)
+	defer cancel()
+
+	// We want to hit CommonNonRoutableMetadataIP to see if we're on AWS, GCP,
+	// or Azure. All three (and many othersA) use the same metadata IP.
```
bradfitz, line 136: whoops, typo. Fixed. Will push/rebase before merge.

util/cloudenv/cloudenv.go
```diff
@@ -66,20 +82,103 @@ func Get() Cloud {
...
+	defer cancel()
+
+	// We want to hit CommonNonRoutableMetadataIP to see if we're on AWS, GCP,
+	// or Azure. All three (and many others) use the same metadata IP.
+	//
+	// But to avoid triggering the AWS CloudWatch "MetadataNoToken" metric (for which
+	// there might be an alert registered?), make our initial request be a token
+	// request. This only works on AWS, but the failing HTTP response on other clouds gives
+	// us enough clues about which cloud we're on.
+	req, err := http.NewRequestWithContext(ctx, "PUT", "http://"+CommonNonRoutableMetadataIP+"/latest/api/token", strings.NewReader(""))
```
DentonGentry, line 142: To my knowledge, a PUT will be ok.

DentonGentry (approved):

Issue #388, opened 2020-05-14
Add DERP region support


bradfitz:
We currently have one 1 VM per DERP region.

For both scalability and high availability, that's a bit scary. It's not super scary because we're only using super cheap VMs and we can scale vertically a bunch, and worst case clients fall over to another region with more latency for a few minutes (or hours, depending on how bad the outage is).

But we can do better.

Plan is to run N VMs per region and do inter-VM routing within a region, but minimize that by having domains prefer a certain VM in that region. But if it's down, they'll also quickly fail over to another.

Then the control server will give each consistently hash domains onto the available DERP nodes per region.

bradfitz:
Forgot to reference it, but also: https://github.com/tailscale/tailscale/commit/3fa58303d0dd206609362d68736f1039801ffd8d

bradfitz:
This is sufficiently done.

We're at least running two nodes in all the major regions for HA purposes now and intra-region packet forwarding works. We're not load balancing yet but that's just a config tweak we can do when needed.

kylecarbs:
@bradfitz what small tweak would be necessary? We're trying to DERP in a Kubernetes cluster where there are numerous dynamic replicas with internal IPs, load-balanced under a single external IP.

DentonGentry:
@kylecarbs I think getting DERP working within an environment like that is well beyond what can just be answered in a quick update on this bug. The current custom derp feature is considered done: it works, as a single node, in a straightforward hosting environment like EC2.

If you really want to discuss it you're welcome to contact support@tailscale.com.

If I may ask: why do you want to run your own DERP server?

bradfitz:
Short answer, see these `cmd/derper` flags:

```
  -mesh-psk-file string
    	if non-empty, path to file containing the mesh pre-shared key file. It should contain some hex string; whitespace is trimmed.
  -mesh-with string
    	optional comma-separated list of hostnames to mesh with; the server's own hostname can be in the list
```

Happy to discuss more elsewhere.

PR #4988, opened 2022-06-30
tailcfg: add HTTPForTests to enable non-TLS servers in testing

kylecarbs:
DERP currently requires HTTPS for tests with derphttp.NewRegionClient, as
there is no way to force HTTP. This adds a new HTTPForTests property to
DERPNode to force HTTP.

It seems potentially excessive to add an option to tailcfg.DERPMap for this case,
as NewClient consumes a URL that would determine whether HTTPS is used. I'm
very open to feedback on this approach, and happy to entirely change it!

### DIFF
--- a/derp/derphttp/derphttp_client.go
+++ b/derp/derphttp/derphttp_client.go
@@ -178,7 +178,10 @@ func (c *Client) targetString(reg *tailcfg.DERPRegion) string {
 	return fmt.Sprintf("region %d (%v)", reg.RegionID, reg.RegionCode)
 }
 
-func (c *Client) useHTTPS() bool {
+func (c *Client) useHTTPS(node *tailcfg.DERPNode) bool {
+	if node.HTTPForTests {
+		return false
+	}
 	if c.url != nil && c.url.Scheme == "http" {
 		return false
 	}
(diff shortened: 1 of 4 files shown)


bradfitz:
I'm missing some context. What's the problem?

kylecarbs:
magicsock [creates](https://github.com/coder/tailscale/blob/aa37aece9c007de64042ba2f882b4f0c703b2c1b/wgengine/magicsock/magicsock.go#L1362-L1380) a DERP region client, but there is no way to specify a target URL (and reasonably, because the regions are dynamic from the DERPMap). Because of this, there is no way to specify whether DERP is being served over HTTP.

We have tests that serve DERP over HTTP. We could refactor to use HTTPS instead, but I figured this was worth exploring first.

DentonGentry:
As with the earlier conversation on https://github.com/tailscale/tailscale/issues/388, could you email support@tailscale.com to discuss it? Custom DERP is available for people to use, but derper is also a big piece of infrastructure we run. If there will be PRs of substantive things which y'all want to upstream, we'd like to talk about it as it impacts our production.

Brad and I both monitor support@tailscale.com, and would see it.

bradfitz:
Or we can do a quick call to get on the same page: https://calendly.com/bradfitz

kylecarbs:
I sent an email. I'm very happy to chat with y'all about what we're up to, let's follow up in the email thread!

kylecarbs:
Seems fine for us to do that instead, I'll go ahead and close this PR!

I believe there's still a use-case for DERP over HTTP when users access Coder over localhost, but I'll follow up if I encounter any issues.

PR #4989, opened 2022-06-30
net/dns: fix resolved dbus issue

### LINKED ISSUE #4645: Sudden High CPU Usage With Tailscale VPN and systemd-resolved when dbus restarts
### What is the issue?

I was running the tailscale VPN daemon (no devices were logged in). Out of nowhere, I heard the fans spin up, and saw that both the tailscale daemon and systemd-resolved were using 100% CPU. I restarted the systemd-resolved service and both processes dropped to near-0 cpu usage.

### Steps to reproduce

_No response_

_No response_

### OS

Linux

### OS version

Ubuntu 22.04

### Tailscale version

1.24.2

### Bug report

BUG-ec1aaed5a0ab790bcc95321a028585ebe4ac88a7ebe5dbe6e169105b186c328e-20220508232821Z-057aacabef55d99d

nyghtowl:
When dbus restarts it can cause the tailscaled to crash because the nil signal was not handled in resolved. It requires manual restart and on restart it doesn't recognize systemd is running so defaults to direct. Fixing so the nil signal is handled and tailscaled stayes connected to systemd when dbus restarted.

Fixes #4645

### DIFF
--- a/net/dns/resolved.go
+++ b/net/dns/resolved.go
@@ -14,6 +14,8 @@ import (
 	"net"
 	"strings"
 	"sync"
+	"tailscale.com/logtail/backoff"
+	"time"
 
 	"github.com/godbus/dbus/v5"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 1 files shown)


bradfitz:
This avoids the tailscaled crash on `systemctl restart dbus` (as might happen during a dbus update),

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x95ae78]

goroutine 31 [running]:
tailscale.com/net/dns.(*resolvedManager).resync(0xc0000f2a80, {0xe963d0, 0xc000114680}, 0xc000434360)
        /home/bradfitz/src/tailscale.com/net/dns/resolved.go:141 +0x118
created by tailscale.com/net/dns.newResolvedManager
        /home/bradfitz/src/tailscale.com/net/dns/resolved.go:111 +0x24f
```

from when this channel is closed and `signal` is `nil`,

```go
		case signal := <-signals:
			// In theory the signal was filtered by DBus, but if
			// AddMatchSignal in the constructor failed, we may be
			// getting other spam.
			if signal.Path != dbusPath || signal.Name != dbusInterface+"."+dbusOwnerSignal {
				continue
			}
```

But even if tailscaled stays up with this change, it's no longer going to ever reconnect to dbus, right?

One thing we keep talking about is doing a "meta Linux DNS manager" that constantly reevaluates the state of the world and reacts on changes. (e.g. something else rewriting /etc/resolv.conf or dbus coming/going or DNS manager processes starting/stopping). Rather than our current state of the world where we make the decision once on start-up. Something like a meta manager would be responsible for trying to reconnect to dbus on close.

> It requires manual restart and on restart it doesn't recognize systemd is running so defaults to direct. 

I don't follow that bit. Normally tailscaled is run under systemd and will restart on crash. Are you saying that on crash & restart it doesn't come back up in systemd-resolved mode?

/cc @danderson

nyghtowl:
Partially works but getting error

`wgengine: watchdog timeout on Reconfig`

Can Close be emptied out and is the synclock still needed?

bradfitz:
Closing this one; superseded by your new PR: https://github.com/tailscale/tailscale/pull/5026

PR #4991, opened 2022-06-30
util/cloudenv: avoid goroutine leak from gcpmetadata.OnGCE()

mihaip:
gcpmetadata.OnGCE() races multiple goroutines to determine if we're on
GCP, which trips up E2E tests in corp. For now avoid calling it
altogether under tests.

### DIFF
--- a/util/cloudenv/cloudenv.go
+++ b/util/cloudenv/cloudenv.go
@@ -6,6 +6,7 @@
 package cloudenv
 
 import (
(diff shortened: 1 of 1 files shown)


mihaip:
`gcpmetadata.OnGCE() ` call is gone with 6b71568eb73104ae13f4660d86cd53ac7583f5a5, not necessary.

PR #5000, opened 2022-07-02
go.mod: bump wireguard-go to set CLOEXEC on tun/netlink fds

### LINKED ISSUE #4992: mosh-server retains a fd to /dev/net/tun even after tailscaled terminates
### What is the issue?

I started a mosh session inside tailscale ssh, then upgraded tailscaled using apt-get. The new instance of tailscaled was unable to start, because the tailscale0 interface was in use.

`fuser /dev/net/tun` revealed the pid of mosh-server, a daemon which (intentionally) sticks around even after ssh is disconnected, so that the session can resume next time the network comes up. Killing mosh-server allowed tailscaled to restart.

It's likely that the ssh session inherited the /dev/net/tun fd because it lacked the FD_CLOEXEC flag. If that's the case, it's probably true of several other (less disruptive) file descriptors as well.

### Steps to reproduce

- Run tailscaled server (in this case I tested both 1.25.98 and 1.27.57)
- tailscale up --ssh
- mosh into the server
- From the mosh session or an unrelated session, stop and restart tailscaled

Expected:
- tailscaled restarts
- mosh session resumes

Actual:
- tailscaled complains in the log that tailscale0 is already in use, and terminates
- `fuser /dev/net/tun` shows mosh-server using the tun fd.
- Killing mosh-server allows tailscaled to start.

Obviously does not occur without tailscale-ssh.

### OS

Linux

### OS version

Debian 11.1

### Tailscale version

1.27.57

### Bug report

BUG-71e4de5749941943e9623ef9b80459ab2dab186100c021f84cf45b2ff3fc8c72-20220630235847Z-3a2c9156f63670cd

bradfitz:
To get:
https://github.com/WireGuard/wireguard-go/commit/c31a7b1ab47807f01613a571cc480f79d5fb4181

Diff:
https://github.com/WireGuard/wireguard-go/compare/95b48cdb3961..c31a7b1ab47807f01613a571cc480f79d5fb4181

Fixes #4992

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,7 +57,7 @@ require (
 	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211
 	golang.org/x/time v0.0.0-20211116232009-f0f3c7e86c11
 	golang.org/x/tools v0.1.11
-	golang.zx2c4.com/wireguard v0.0.0-20220317000134-95b48cdb3961
+	golang.zx2c4.com/wireguard v0.0.0-20220703234212-c31a7b1ab478
 	golang.zx2c4.com/wireguard/windows v0.4.10
 	gvisor.dev/gvisor v0.0.0-20220407223209-21871174d445
 	honnef.co/go/tools v0.4.0-0.dev.0.20220404092545-59d7a2877f83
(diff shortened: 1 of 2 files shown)


maisem (approved):

DentonGentry (approved):

bradfitz:
There's also a netlink route fd leaking, so I sent https://git.zx2c4.com/wireguard-go/log/?h=bf/cloexec to Jason.

bradfitz:
I've updated that change to Jason to do all platforms and do all the open/socket calls.

Once that's merged, this change should just become a `go.mod` bump of wireguard-go. (I verified there's nothing else that'd be in the bump that'd be scary for a point release)

bradfitz:
PTAL. Now this is just a go.mod bump.

DentonGentry (approved):

PR #5009, opened 2022-07-05
tka: implement AUM and Key types

twitchyliquid64:
This is the first in a series of PRs implementing the internals for the Tailnet Key Authority. This PR implements the `AUM` and `Key` types, which are used by pretty much everything else. Future PRs:

 - The `State` type & related machinery
 - The `Chonk` (storage) type & `Mem` implementation
 - The `Authority` type and sync implementation

### DIFF
--- a/tka/aum.go
+++ b/tka/aum.go
@@ -1,0 +1,256 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package tka
+
+import (
+	"bytes"
+	"crypto/ed25519"
+	"errors"
+	"fmt"
+
+	"github.com/fxamacker/cbor/v2"
+	"golang.org/x/crypto/blake2s"
+)
+
+// AUMHash represents the BLAKE2s digest of an Authority Update Message (AUM).
+type AUMHash [blake2s.Size]byte
+
+// AUMSigHash represents the BLAKE2s digest of an Authority Update
+// Message (AUM), sans any signatures.
+type AUMSigHash [blake2s.Size]byte
+
+// AUMKind describes valid AUM types.
+type AUMKind uint8
+
+// Valid AUM types. Do NOT reorder.
+const (
+	AUMInvalid AUMKind = iota
+	// An AddKey AUM describes a new key trusted by the TKA.
+	//
+	// Only the Key optional field may be set.
+	AUMAddKey
+	// A RemoveKey AUM describes hte removal of a key trusted by TKA.
+	//
+	// Only the KeyID optional field may be set.
+	AUMRemoveKey
+	// A DisableNL AUM describes the disablement of TKA.
+	//
+	// Only the DisablementSecret optional field may be set.
+	AUMDisableNL
+	// A NoOp AUM carries no information and is used in tests.
+	AUMNoOp
+	// A UpdateKey AUM updates the metadata or votes of an existing key.
+	//
+	// Only KeyID, along with either/or Meta or Votes optional fields
+	// may be set.
+	AUMUpdateKey
+	// A Checkpoint AUM specifies the full state of the TKA.
+	//
+	// Only the State optional field may be set.
+	AUMCheckpoint
+)
+
+func (k AUMKind) String() string {
+	switch k {
+	case AUMInvalid:
+		return "invalid"
+	case AUMAddKey:
+		return "add-key"
+	case AUMRemoveKey:
+		return "remove-key"
+	case AUMDisableNL:
+		return "disable-nl"
+	case AUMNoOp:
+		return "no-op"
+	case AUMCheckpoint:
+		return "checkpoint"
+	case AUMUpdateKey:
+		return "update-key"
+	default:
+		return fmt.Sprintf("AUM?<%d>", int(k))
+	}
+}
+
+// AUM describes an Authority Update Message.
+//
+// The rules for adding new types of AUMs (MessageKind):
+// - CBOR key IDs must never be changed.
+// - New AUM types must not change semantics that are manipulated by other
+//   AUM types.
+// - The serialization of existing data cannot change (in other words, if
+//   an existing serialization test in aum_test.go fails, you need to try a
+//   different approach).
+//
+// The rules for adding new fields are as follows:
+// - Must all be optional.
+// - An unset value must not result in serialization overhead. This is
+//   necessary so the serialization of older AUMs stays the same.
+// - New processing semantics of the new fields must be compatible with the
+//   behavior of old clients (which will ignore the field).
+// - No floats!
+type AUM struct {
+	MessageKind AUMKind `cbor:"1,keyasint"`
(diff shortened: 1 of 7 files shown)


tka/aum.go
```diff
@@ -0,0 +1,233 @@
...
+// - An unset value must not result in serialization overhead. This is
+//   necessary so the serialization of older AUMs stays the same.
+// - New processing semantics of the new fields must be compatible with the
+//   behavior of old clients (which will ignore the field).
+// - No floats!
+type AUM struct {
+	MessageKind AUMKind `cbor:"1,keyasint"`
+	PrevAUMHash []byte  `cbor:"2,keyasint"`
+
+	// Used for AddKey
```
crawshaw, line 97: Typically in the format `// Key is __, used for AddKey`. And below.

crawshaw (commented):
These types seem fine, but somewhere we need to have a discussion about CBOR.

The question everyone (and me!) will ask on seeing it is: why not JSON? That explanation needs to live in comments in here somewhere, possibly the package docs. (In fact, I've been trying to make the argument myself and haven't really been succeeding, I suspect we could use JSON with our own canonicalization rules, and so if I'm wrong we need to explain that.)

tka/aum.go
twitchyliquid64: Done

tka/aum.go
```diff
@@ -0,0 +1,256 @@
...
+		return h, true
+	}
+	return h, false
+}
+
+func (a *AUM) sign25519(priv ed25519.PrivateKey) {
+	key := Key{Kind: Key25519, Public: priv.Public().(ed25519.PublicKey)}
+	sigHash := a.SigHash()
+
+	a.Signatures = append(a.Signatures, Signature{
```
crawshaw, line 250: What should this do if the signature is already in signatures? It's fine to duplicate?

tka/key.go
```diff
@@ -0,0 +1,118 @@
...
+	// Weighting is used to deterministically resolve branches in the AUM
+	// chain (i.e. forks, where two AUMs exist with the same parent).
+	Votes uint `cbor:"2,keyasint"`
+
+	// Public encodes the public key of the key. For 25519 keys,
+	// this is simply the point on the curve representing the public
+	// key.
+	Public []byte `cbor:"3,keyasint"`
+
+	Meta map[string]string `cbor:"12,keyasint,omitempty"`
```
crawshaw, line 49: This could really use a comment. What are the keys / values?

tka/key.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	return nil
+}
+
+// KeyID identifies the key used in a signature.
+//
+// For 25519 keys: The 32-byte public key.
+type KeyID []byte
+
+// Signature describes a signature over some data.
```
crawshaw, line 96: describes a <kind of goes here> signature

crawshaw (approved):

tka/aum.go
twitchyliquid64: It's fine (safe) to duplicate - multiple signatures with the same KeyID will not result in a multiple of the signature weight. But its definitely a waste of space.

I don't thin we need to check for duplicates here, its a private method as well.

tka/key.go
twitchyliquid64: Done. This is an arbitrary K:V container for human-friendly metadata about the key thats irrelevant to the operation of TKA. Name, Created, whatever.

tka/aum.go
```diff
@@ -0,0 +1,256 @@
...
+	// understood + widely-implemented, and already proven for use in signing
+	// assertions through its use by FIDO2 devices.
+	out := bytes.NewBuffer(make([]byte, 0, 128))
+	encoder, err := cbor.CTAP2EncOptions().EncMode()
+	if err != nil {
+		// Deterministic validation of encoding options, should
+		// never fail.
+		panic(err)
+	}
+	if err := encoder.NewEncoder(out).Encode(a); err != nil {
```
dsnet, line 212: This should be using [`cbor.CanonicalEncOptions`](https://pkg.go.dev/github.com/fxamacker/cbor/v2#CanonicalEncOptions), right?

tka/aum.go
```diff
@@ -0,0 +1,256 @@
...
+	// Why CBOR and not something like JSON?
+	//
+	// The main function of an AUM is to carry signed data. Signatures are
+	// over digests, so the serialized representation must be deterministic.
+	// Further, experience with other attempts (JWS/JWT,SAML,X509 etc) has
+	// taught us that even subtle behaviors such as how you handle invalid
+	// or unrecognized fields + any invariants in subsequent re-serialization
+	// can easily lead to security-relevant logic bugs. Its certainly possible
+	// to invent a workable scheme by massaging a JSON parsing library, though
+	// profoundly unwise.
```
dsnet, line 198: I suspect that it's not so much determinism that you need, but rather stability or canonicalization. Determinism is usually defined as the same program producing the same output, but there's no guarantee that different versions of the program or different implementations output the same thing.

I suspect that the property you want is that the serialization is stable where different versions of the program produce the same output. Canonical output goes a step even further than all implementations must produce the same output.

In order of strictness, it goes from non-deterministic to deterministic to stable to canonical.

---

There's been advances in canonical JSON in the recent years. [RFC 8785](https://datatracker.ietf.org/doc/html/rfc8785) defines a canonical representation of JSON, with a Go implementation in [`json.RawValue.Canonicalize`](https://pkg.go.dev/github.com/go-json-experiment/json#RawValue.Canonicalize).

tka/aum.go
```diff
@@ -0,0 +1,256 @@
...
+		if a.KeyID != nil || a.State != nil || a.Key != nil || a.Votes != nil || a.Meta != nil {
+			return errors.New("DisableNL AUMs may only a disablement secret")
+		}
+	}
+
+	return nil
+}
+
+// Serialize returns the given AUM in a serialized format.
+func (a *AUM) Serialize() []byte {
```
dsnet, line 188: Is there a reason why this isn't a `MarshalBinary` method?
https://pkg.go.dev/encoding#BinaryMarshaler

tka/aum.go
twitchyliquid64: `CanonicalEncOptions` is less strict than the CTAP2 variety, which is also battle-tested for use in signatures by FIDO2 keys. Also the 'Canonical CBOR' rules allow tags, which would provide more valid encodings for a given data than we want.

tka/aum.go
twitchyliquid64: Didn't know the nuance of all those terms! In that case, we want 'canonical'.

Between the poor efficiency of encoding raw bytes and its use for signed data being an antipattern (see JWS/JWTs), I think we would have more issues from canonical JSON than we would from something like CBOR.

tka/aum.go
twitchyliquid64: Nope, I'll rename in a future revision!

(Though I admit I like serialize because it makes me thing of extruding cereal lol)

tka/aum.go
dsnet: Ah, got it. I couldn't tell from the name of "CTAP2" that this is a form of canonicalization.

PR #5014, opened 2022-07-06
tka: implement State and logic for applying AUMs

twitchyliquid64:
This is the second PR in a series implementing the Tailnet Key Authority. This PR implements the `State` type, which is the instantaneous value of the TKA at a moment in time. A `State` is mutated by applying an AUM, resulting in a new, independent state (like applying a log).

PRs to follow:
 - The `Chonk` (storage) type & `Mem` implementation
 - The `Authority` type and sync implementation

### DIFF
--- a/tka/state.go
+++ b/tka/state.go
@@ -1,0 +1,204 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package tka
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+
+	"golang.org/x/crypto/argon2"
+)
+
+// ErrNoSuchKey is returned if the key referenced by a KeyID does not exist.
+var ErrNoSuchKey = errors.New("key not found")
+
+// State describes Tailnet Key Authority state at an instant in time.
+//
+// State is mutated by applying Authority Update Messages (AUMs), resulting
+// in a new State.
+type State struct {
+	// LastAUMHash is the blake2s digest of the last-applied AUM.
+	// Because AUMs are strictly ordered and form a hash chain, we
+	// check the previous AUM hash in an update we are applying
+	// is the same as the LastAUMHash.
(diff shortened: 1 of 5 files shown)


tka/state.go
```diff
@@ -0,0 +1,189 @@
...
+
+	return Key{}, ErrNoSuchKey
+}
+
+// Clone makes an independent copy of State.
+//
+// NOTE: There is a difference between a nil slice and an empty
+//       slice for encoding purposes, so an implementation of Clone()
+//       must take care to preserve this.
+func (s State) Clone() State {
```
crawshaw, line 45: Now that you're in the oss repo you can do this with ./cmd/cloner. Add a line like:

```
//go:generate go run tailscale.com/cmd/cloner -type=State
```

tka/state.go
```diff
@@ -0,0 +1,189 @@
...
+
+// ErrNoSuchKey is returned if the key referenced by a KeyID does not exist.
+var ErrNoSuchKey = errors.New("key not found")
+
+// State describes Tailnet Key Authority state at an instant in time.
+//
+// State is mutated by applying Authority Update Messages (AUMs), resulting
+// in a new State.
+type State struct {
+	LastAUMHash        []byte   `cbor:"1,keyasint"`
```
crawshaw, line 23: I think these fields could use some commentary.

Are hashes and secrets always fixed lengths? I'm wondering if we should be using named array types for them `type AUMHash [32]byte`. That would mean using zeros instead of nil, but might be particularly nice with a `type Secret [64]byte` type.

crawshaw (approved):

tka/state.go
twitchyliquid64: I spent a while trying to get this to work, but ultimately I think the assumptions in `cloner` are too different and its not super worthwhile adding in the logic for just this case. Specifically:

1. `cloner` assumes a nil slice and an empty slice are equivalent
2. `cloner` works on pointer receiver (and assumes pointer receiver) methods, whereas we are trying to intentionally pass by value for values which can't be empty

PR #5017, opened 2022-07-07
tka: implement Chonk type & an in-memory implementation

twitchyliquid64:
This PR is the third in a series implementing the Tailnet Key Authority. Chonks are responsible for efficient storage of AUMs and other TKA state.

For testing/prototyping I've implemented an in-memory version, but once we start to use this from tailscaled we'll implement a file-based version.

Future PRs:
 - Implementing skeleton for `Authority` along with AUM-walking & chain resolution internals: `fastForward()`, `computeStateAt()`, `computeActiveChain()` etc
 - `SyncOffer` computation and the remaining API surface

### DIFF
--- a/tka/tailchonk.go
+++ b/tka/tailchonk.go
@@ -1,0 +1,158 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package tka
+
+import (
+	"os"
+	"sync"
+)
+
+// Chonk implementations provide durable storage for AUMs and other
+// TKA state.
+//
+// All methods must be thread-safe.
+//
+// The name 'tailchonk' was coined by @catzkorn.
+type Chonk interface {
+	// AUM returns the AUM with the specified digest.
+	//
+	// If the AUM does not exist, then os.ErrNotExist is returned.
+	AUM(hash AUMHash) (AUM, error)
+
+	// ChildAUMs returns all AUMs with a specified previous
+	// AUM hash.
+	ChildAUMs(prevAUMHash AUMHash) ([]AUM, error)
+
+	// CommitVerifiedAUMs durably stores the provided AUMs.
+	// Callers MUST ONLY provide AUMs which are verified (specifically,
+	// a call to aumVerify() must return a nil error).
+	// as the implementation assumes that only verified AUMs are stored.
+	CommitVerifiedAUMs(updates []AUM) error
+
+	// Heads returns AUMs for which there are no children. In other
+	// words, the latest AUM in all possible chains (the 'leaves').
+	Heads() ([]AUM, error)
+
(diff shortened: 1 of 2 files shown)


DentonGentry:
Please tell me the Chonk type is for something really large, an Absolute Unit of a datastructure.

twitchyliquid64:
Its the storage abstraction for storing the current window of signed update messages, so I definitely imagine something heckin large (corgi-butt vibes for sure)

tka/tailchonk.go
```diff
@@ -0,0 +1,158 @@
...
+// TKA state.
+//
+// All methods must be thread-safe.
+//
+// The name 'tailchonk' was coined by @catzkorn.
+type Chonk interface {
+	// GetAUM returns the AUM with the specified digest.
+	//
+	// If the AUM does not exist, then os.ErrNotExist is returned.
+	GetAUM(hash AUMHash) (AUM, error)
```
crawshaw, line 22: "Getters" in go often drop the Get, if they can get away with it. Usually on methods you can. I think you can here, without creating confusion?

tka/tailchonk.go
```diff
@@ -0,0 +1,158 @@
...
+// Chonk implementations provide durable storage for AUMs and other
+// TKA state.
+//
+// All methods must be thread-safe.
+//
+// The name 'tailchonk' was coined by @catzkorn.
+type Chonk interface {
+	// GetAUM returns the AUM with the specified digest.
+	//
+	// If the AUM does not exist, then os.ErrNotExist is returned.
```
crawshaw, line 21: os.ErrNotExist is an alias these days for fs.ErrNotExist, which requires fewer dependencies.

I'm not sure it's exactly the error the error to use, but I guess it's ok for now.

tka/tailchonk.go
```diff
@@ -0,0 +1,158 @@
...
+	GetAUM(hash AUMHash) (AUM, error)
+
+	// GetChildAUMs returns all AUMs with a specified previous
+	// AUM hash.
+	GetChildAUMs(prevAUMHash AUMHash) ([]AUM, error)
```
crawshaw, line 26: If the AUM does not exist does this return a particular error, or an empty slice?

crawshaw (approved):

tka/tailchonk.go
twitchyliquid64: Neither! Theres no requirement to know the parent AUM, as the 'child' AUMs all reference the parent hash.

tka/tailchonk.go
twitchyliquid64: I'm hoping that LTO is smart enough to remove anything that doesn't get used, but in any case I'm guessing the os package is pulled in anyway?

If it ends up mattering, I can make a specific `ErrNotExist` error for this package.

tka/tailchonk.go
twitchyliquid64: Done.

PR #5025, opened 2022-07-08
ipn/localapi: add API for uploading client metrics

mihaip:
Clients may have platform-specific metrics they would like uploaded
(e.g. extracted from MetricKit on iOS). Add a new local API endpoint
that allows metrics to be updated by a simple name/value JSON-encoded
struct.

### DIFF
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -21,6 +21,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"inet.af/netaddr"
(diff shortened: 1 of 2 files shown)


ipn/localapi/localapi.go
```diff
@@ -730,6 +743,50 @@ func (h *Handler) serveDial(w http.ResponseWriter, r *http.Request) {
...
+		if metric, ok := metrics[m.Name]; ok {
+			metric.Add(int64(m.Value))
+		} else {
+			var metric *clientmetric.Metric
+			switch m.Type {
+			case "counter":
+				metric = clientmetric.NewCounter(m.Name)
+			case "gauge":
+				metric = clientmetric.NewGauge(m.Name)
+			default:
```
bradfitz, line 777: These will panic if a localapi client provides an m.Name that's already in use, though.

I guess inside an HTTP handler that'll get caught and it'll be an HTTP error at least?

Either add a comment saying that's acceptable, or maybe add new api to the clientmetric package to test whether a metric name is already published?

bradfitz (approved):

ipn/localapi/localapi.go
mihaip: Yeah, the panics were being recovered (though I don't actually see where that's happening), but it doesn't seem ideal to rely on that.

Added a `HasPublished` API and made the API call return a 400 in that case.

bradfitz (approved):

PR #5026, opened 2022-07-08
net/dns: fix resolved dbus issue

nyghtowl:


bradfitz:
Squash commits & fix signed-off-by line & un-draft this & ping me when this is ready for review?

Looks like it doesn't work yet probably because the configs chan is never initialized anywhere?

net/dns/resolved.go
```diff
@@ -74,73 +76,164 @@ type resolvedLinkDomain struct {
...
+func (m *resolvedManager) run(ctx context.Context) {
+	var (
+		conn      *dbus.Conn
+		signals   chan *dbus.Signal
+		r1Manager dbus.BusObject
+	)
+	bo := backoff.NewBackoff("resolvedSetDNS", m.logf, 30*time.Second)
+	needsReconnect := make(chan bool, 1)
+	defer close(needsReconnect)
+	defer close(m.configCR)
```
bradfitz, line 148: same. But this one's arguably worse because this goroutine is the reader of this channel and close is a write operation. So you can be closing a channel concurrently with a SetDNS call which would then panic

best to just remove this one

net/dns/resolved.go
```diff
@@ -74,73 +76,164 @@ type resolvedLinkDomain struct {
...
+
+		r1Manager = conn.Object(dbusResolvedObject, dbus.ObjectPath(dbusResolvedPath))
+
+		// Only receive the DBus signals we need to resync our config on
+		// resolved restart. Failure to set filters isn't a fatal error,
+		// we'll just receive all broadcast signals and have to ignore
+		// them on our end.
+		if err = conn.AddMatchSignal(dbus.WithMatchObjectPath(dbusPath), dbus.WithMatchInterface(dbusInterface), dbus.WithMatchMember(dbusOwnerSignal), dbus.WithMatchArg(0, dbusResolvedObject)); err != nil {
+			m.logf("[v1] Setting DBus signal filter failed: %v", err)
+		}
```
bradfitz, line 179: this error is fine to ignore?

net/dns/resolved.go
```diff
@@ -160,26 +253,25 @@ func (m *resolvedManager) resync(ctx context.Context, signals chan *dbus.Signal)
...
-			m.mu.Unlock()
 			if err != nil {
 				m.logf("failed to configure systemd-resolved: %v", err)
 			}
 		}
 	}
 }
 
-func (m *resolvedManager) syncLocked(ctx context.Context) error {
+// sync is only called from the run goroutine to sync.
```
bradfitz, line 268: once you have a comment at all, the comment should be in the normal form to start, and then you can end with this info.

So start with saying what sync _does_.

But while I'm looking at the body, I'm thinking `setConfigOverDBus` or something would be a better name.

net/dns/resolved.go
```diff
@@ -74,73 +76,164 @@ type resolvedLinkDomain struct {
...
+	}
 }
 
-func (m *resolvedManager) resync(ctx context.Context, signals chan *dbus.Signal) {
-	defer close(m.syncerDone)
+func (m *resolvedManager) run(ctx context.Context) {
+	var (
+		conn      *dbus.Conn
+		signals   chan *dbus.Signal
+		r1Manager dbus.BusObject
```
bradfitz, line 143: what's the meaning of "r1Manager" used throughout this change? comment here with a reference?

net/dns/resolved.go
```diff
@@ -74,73 +76,164 @@ type resolvedLinkDomain struct {
...
+	)
+	bo := backoff.NewBackoff("resolvedSetDNS", m.logf, 30*time.Second)
```
bradfitz, line 145: "resolved-dbus" ?

net/dns/resolved.go
```diff
@@ -13,7 +13,9 @@ import (
 	"fmt"
 	"net"
 	"strings"
-	"sync"
+	"time"
+
+	"tailscale.com/logtail/backoff"
```
bradfitz, line 19: delete empty line here. just two sections is fine: std vs all else

net/dns/resolved.go
```diff
@@ -74,73 +76,164 @@ type resolvedLinkDomain struct {
...
+
 	for {
 		select {
 		case <-ctx.Done():
+			if r1Manager == nil {
+				return
+			}
+			// RevertLink on connection if signals are interrupted in order to create a new connection.
+			if call := r1Manager.CallWithContext(ctx, dbusResolvedInterface+".RevertLink", 0, m.ifidx); call.Err != nil {
+				// TODO(raggi): return fmt.Errorf("RevertLink: %w", call.Err)
```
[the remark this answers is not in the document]
nyghtowl: This was part of the Close function originally. When moving things around, I thought it was needed to create a new link which @raggi and I discussed that its not. So comment needs cleanup but I thought we still needed it.

net/dns/resolved.go
nyghtowl: Original code ignored it and its noted as not being a fatal error because the code has more broadcast signals to review.

net/dns/resolved.go
```diff
@@ -74,73 +75,161 @@ type resolvedLinkDomain struct {
...
+				configCR.res <- fmt.Errorf("resolved DBus does not have a connection")
+				continue
+			}
+			err := m.setConfigOverDBus(ctx, rManager, configCR.config)
+			configCR.res <- err
+		case <-needsReconnect:
+			if err := reconnect(); err != nil {
+				m.logf("[v1] SystemBus reconnect error %T", err)
+			}
+			needsReconnect <- false
```
bradfitz, line 214: this isn't right. it'll spin forever.

net/dns/resolved.go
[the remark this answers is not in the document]
nyghtowl: After discussion between @bradfitz and @danderson - retaining for future if needed and doesn't hurt.

bradfitz:
In the commit message, this line isn't of the right format that GitHub + tools recognize:

```
Co-authored-by:James Tucker (raggi)
```

Needs angle brackets & an email-ish identifier associated with James on GitHub. (see git history for a valid one).

Also space after colon, and put it in the block at the bottom above the Signed-off-by

bradfitz:
I tried this out and restarted dbus and tried to change my DNS settings and it logged:

```
failed to configure resolved: <nil>
```

bradfitz:
The various log prints should have a `dns: ` prefix.

Right now I see e.g.:

```
systemd-resolved restarted, syncing DNS config
```

net/dns/resolved.go
```diff
@@ -74,73 +75,163 @@ type resolvedLinkDomain struct {
...
+			conn.Close()
+		}
+	}()
+
+	// Reconnect the systemBus if disconnected.
+	reconnect := func() error {
+		var err error
+		signals = make(chan *dbus.Signal, 16)
+		conn, err = dbus.SystemBus()
+		m.logf("[v1] Connecting DBUS or reconnecting to recover disconnect.")
```
bradfitz, line 156:
```
if err != nil {
    m.logf("dns: dbus connection error: %v", err)
} else {
     m.logf("dns: [v1] dbus connected")
}
```

net/dns/resolved.go
```diff
@@ -74,73 +75,163 @@ type resolvedLinkDomain struct {
...
 			// getting other spam.
 			if signal.Path != dbusPath || signal.Name != dbusInterface+"."+dbusOwnerSignal {
 				continue
 			}
+			if lastConfig.IsZero() {
+				continue
+			}
 			// signal.Body is a []any of 3 strings: bus name, previous owner, new owner.
 			if len(signal.Body) != 3 {
 				m.logf("[unexpectected] DBus NameOwnerChanged len(Body) = %d, want 3")
```
bradfitz, line 237: typo in [unexpected]

net/dns/resolved.go
```diff
@@ -301,20 +389,9 @@ func (m *resolvedManager) GetBaseConfig() (OSConfig, error) {
...
-	}
-
-	select {
-	case <-m.syncerDone:
-	case <-ctx.Done():
-		m.logf("timeout in systemd-resolved syncer shutdown")
-	}
+	m.cancel()
 
+	// No need to do anything on close which is handled in run.
```
bradfitz, line 394: Maybe instead of this comment, just:

     m.cancel() // stops the 'run' method goroutine

bradfitz (commented):
Minor nits aside, LGTM

bradfitz:
Let's tweak the commit message a bit:

* more descriptive subject than just "fix issue"... maybe: "net/dns: handle D-Bus restarts in resolved manager" ?
* word wrap the body to ~76/80 chars, with space after period

bradfitz (approved):

PR #5034, opened 2022-07-10
net/netcheck: cancel derp probe if node address is nil

kylecarbs:
This was causing a delay of 3s (stunProbeTimeout) if STUN was not being
used on a DERP node (STUNPort set to -1).


net/netcheck/netcheck.go
```diff
@@ -1230,6 +1230,7 @@ func (rs *reportState) runProbe(ctx context.Context, dm *tailcfg.DERPMap, probe
 
 	addr := c.nodeAddr(ctx, node, probe.proto)
 	if addr == nil {
+		cancelSet()
```
bradfitz, line 1233: I don't think this is right. Looking at the caller,

```go
        for _, probeSet := range plan {
                setCtx, cancelSet := context.WithCancel(ctx)
                go func(probeSet []probe) {
                        for _, probe := range probeSet {
                                go rs.runProbe(setCtx, dm, probe, cancelSet)
                        }
                        <-setCtx.Done()
```

... each probeSet can involve multiple nodes. Just because one failed, that doesn't mean the others in the set might not succeed.

Why is nodeAddr returning nil for you?

Perhaps what we need to make the caller do is note when all `runProbe` calls in probeSet have returned and then select on either, like:

```go
   select {
   case <-setCtx.Done():
   case <-allRunProbeCallsDone:
   }
```

But I'd like to understand the problem more first.

kylecarbs: Ahh, you're right. This would cause problems if a single node were to fail.

`nodeAddr` returns nil when `STUNPort` is set to -1, which is supported based on these docs:

https://github.com/tailscale/tailscale/blob/d6b8a18b09d2c24111777d51f79fd63680023438/tailcfg/derpmap.go#L124-L127

If a DERPNode is specified that doesn't use STUN, a timeout presently takes place. It'd be preferred to skip STUN entirely than await the timeout:

https://github.com/tailscale/tailscale/blob/6bed78125933a4fe1cad38c98b4a9dbd0b4f9df4/net/netcheck/netcheck.go#L876-L877

I'll adjust this to use a channel. Thanks for the thorough review, that could have been bad 🤦!

bradfitz: Converted this PR to draft meanwhile.

kylecarbs:
@bradfitz I fixed this up to use a channel. It's not the simplest code to follow, but I didn't want to pass the WaitGroup to a child func.

Happy to make any improvements, but I've confirmed this removes the STUN delay when no DERP servers support it.

kylecarbs:
@bradfitz friendly ping! Happy to make any changes necessary, I'm trying to get rid of our fork :)

kylecarbs:
No biggie! Our fork is really minor, so it's trivial to maintain.

PR #5037, opened 2022-07-11
tka: implement consensus & state computation internals

twitchyliquid64:
This PR implements the nitty-gritty consensus internals. `fastForward` and `computeStateAt` are the 'core' algorithms that drive the whole thing.

A lot of the delta are tests or the new test helper `chaintest_test.go`.

### DIFF
--- a/tka/tka.go
+++ b/tka/tka.go
@@ -4,3 +4,335 @@
 
 // Package tka (WIP) implements the Tailnet Key Authority.
 package tka
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"os"
+	"sort"
+)
+
+// A chain describes a linear sequence of updates from Oldest to Head,
+// resulting in some State at Head.
+type chain struct {
+	Oldest AUM
+	Head   AUM
+
+	state State
+
+	// Set to true if the AUM chain intersects with the active
+	// chain from a previous run.
+	chainsThroughActive bool
+}
+
+// computeChainCandidates returns all possible chains based on AUMs stored
+// in the given tailchonk. A chain is defined as a unique (oldest, newest)
+// AUM tuple. chain.state is not yet populated in returned chains.
+//
+// If lastKnownOldest is provided, any chain that includes the given AUM
+// has the chainsThroughActive field set to true. This bit is leveraged
+// in computeActiveAncestor() to filter out irrelevant chains when determining
+// the active ancestor from a list of distinct chains.
+func computeChainCandidates(storage Chonk, lastKnownOldest *AUMHash, maxIter int) ([]chain, error) {
+	heads, err := storage.Heads()
+	if err != nil {
+		return nil, fmt.Errorf("reading heads: %v", err)
+	}
+	candidates := make([]chain, len(heads))
+	for i := range heads {
+		// Oldest is iteratively computed below.
+		candidates[i] = chain{Oldest: heads[i], Head: heads[i]}
+	}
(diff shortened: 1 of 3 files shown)


tka/tka.go
```diff
@@ -4,3 +4,331 @@
...
+func pickNextAUM(state State, candidates []AUM) AUM {
+	switch len(candidates) {
+	case 0:
+		panic("pickNextAUM called with empty candidate set")
+	case 1:
+		return candidates[0]
+	}
+
+	// Oooof, we have some forks in the chain. We need to pick which
+	// one to use by applying the Fork Resolution Algorithm :sparkles:
```
crawshaw, line 99: it's utf8, you can paste ✨ into the comment :)

tka/tka.go
```diff
@@ -4,3 +4,331 @@
...
+	//
+	// The rules are this:
+	// 1. The child with the highest signature weight is chosen.
```
crawshaw, line 102: You could inline these comments below

tka/tka.go
```diff
@@ -4,3 +4,331 @@
...
+	// Not strictly necessary, but simplifies checks in tests.
+	sort.Slice(candidates, func(i, j int) bool {
+		ih, jh := candidates[i].Oldest.Hash(), candidates[j].Oldest.Hash()
+		return bytes.Compare(ih[:], jh[:]) < 0
+	})
+
+	// candidates.Oldest needs to be computed by working backwards from
+	// head as far as we can.
+	iterAgain := true // if theres still work to be done.
+	for i := 0; i < maxIter && iterAgain; i++ {
```
crawshaw, line 56: what happens if you hit maxIter? no need to error out? (it's passed into computeActiveChain and could use some docs)

tka/tka.go
```diff
@@ -4,3 +4,331 @@
...
+// ancestor of the valid chain.
+//
+// If all the chains end up having the same ancestor, then thats the
+// only possible ancestor, ezpz. However if there are multiple distinct
+// ancestors, that means there are distinct chains, and we need some
+// hint to choose what to use. For that, we rely on the chainsThroughActive
+// bit, which signals to us that that ancestor was part of the
+// chain in a previous run.
+func computeActiveAncestor(storage Chonk, chains []chain) (AUMHash, error) {
+	// Dedupe possible ancestors, tracking if they were part of
```
crawshaw, line 262: why would there be a duplicate? it feels like there's something to `computeChainCandidates` I don't understand yet

crawshaw (approved):
I'm not convinced I really understand `computeChainCandidates`, might need to talk it through sometime, but after that everything seems straightforward.

tka/tka.go
twitchyliquid64: VIBES ✨

tka/tka.go
twitchyliquid64: Welp missed that, added!

twitchyliquid64:
If everything goes well and storage is exactly how we expect it, then there will never be more than one distinct chain (ie: all chains have a common ancestor). `computeChainCandidates()` and `computeActiveAncestor()` is all about being resilient to if that isnt the case for whatever reason.

I have half a mind to pull that out: just compute the active chain from the last known ancestor (as per the TODO).

Does that make more sense?

crawshaw:
Don’t you still need the candidates concept for the case where you get one aum then the server rolls back and you get a different one sent to you?

crawshaw:
I see. Probably worth keeping for that case then.

PR #5039, opened 2022-07-12
tsweb: add Float expvar support in varz

DentonGentry:
We make assertions about stringification of 0.5. IEEE floating point and
all reasonable proprietary floating point can exactly represent 0.5.
We don't make assertions about other floating point values, too brittle
in tests.

### DIFF
--- a/tsweb/tsweb.go
+++ b/tsweb/tsweb.go
@@ -466,6 +466,12 @@ func writePromExpVar(w io.Writer, prefix string, kv expvar.KeyValue) {
 		}
 		fmt.Fprintf(w, "# TYPE %s %s\n%s %v\n", name, typ, name, v.Value())
 		return
+	case *expvar.Float:
+		if typ == "" {
+			typ = "gauge"
+		}
+		fmt.Fprintf(w, "# TYPE %s %s\n%s %v\n", name, typ, name, v.Value())
+		return
 	case *metrics.Set:
 		v.Do(func(kv expvar.KeyValue) {
 			writePromExpVar(w, name+"_", kv)
--- a/tsweb/tsweb_test.go
+++ b/tsweb/tsweb_test.go
@@ -334,6 +334,9 @@ func TestVarzHandler(t *testing.T) {
 		t.Logf("Got: %s", rec.Body.Bytes())
 	})
 
+	half := new(expvar.Float)
+	half.Set(0.5)
+
 	tests := []struct {
 		name string
 		k    string // key name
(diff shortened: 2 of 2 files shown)


tsweb/tsweb_test.go
```diff
@@ -514,6 +542,52 @@ foo_AUint16 65535
 	}
 }
 
+func TestVarzHandlerExponentialFloats(t *testing.T) {
+	t.Run("float_log", func(t *testing.T) {
+		rec := httptest.NewRecorder()
+		VarzHandler(rec, httptest.NewRequest("GET", "/", nil))
+		t.Logf("Got: %s", rec.Body.Bytes())
+	})
```
bradfitz, line 550: this test doesn't test anything? leftover debug?

tsweb/tsweb.go
```diff
@@ -466,6 +466,12 @@ func writePromExpVar(w io.Writer, prefix string, kv expvar.KeyValue) {
 		}
 		fmt.Fprintf(w, "# TYPE %s %s\n%s %v\n", name, typ, name, v.Value())
 		return
+	case *expvar.Float:
+		if typ == "" {
+			typ = "gauge"
+		}
+		fmt.Fprintf(w, "# TYPE %s %s\n%s %g\n", name, typ, name, v.Value())
```
bradfitz, line 473: why the %g in particular? and all the testing for `e` below?

https://github.com/Showmax/prometheus-docs/blob/master/content/docs/instrumenting/exposition_formats.md says anything that strconv.ParseFloat can parse.

which means anything works, including the good ol' %v default: https://go.dev/play/p/CwQ7xQK-oay

bradfitz (approved):

tsweb/tsweb_test.go
DentonGentry: Removed the whole function.

tsweb/tsweb.go
DentonGentry: That is fair. I had been comparing to derp-bwmon's current `prometheus/client_golang` metrics to keep the output roughly the same, but it doesn't need to exactly match so long as Prometheus can scrape it.

tsweb/tsweb_test.go
```diff
@@ -358,6 +361,31 @@ func TestVarzHandler(t *testing.T) {
...
+			"float_with_type_counter",
+			"counter_foo",
+			half,
+			"# TYPE foo counter\nfoo 0.5\n",
+		},
+		{
+			"float_with_type_gauge",
+			"gauge_foo",
+			half,
+			"# TYPE foo gauge\nfoo 0.5\n",
```
maisem, line 387: Add a test for large floats?

maisem (approved):

tsweb/tsweb_test.go
maisem: Oh I just saw the commit message

tsweb/tsweb.go
bradfitz: We use `prometheus/client_golang` anywhere!? :)

Issue #3767, opened 2022-01-19
Shared node DNS issue


JayWStapleton:
### What is the issue?

When user A has nodes shared with user B, and user B has a node shared with user A, MagicDNS does not resolve for the shared nodes in either direction. 

Also, shared nodes did not show up in `tailscale status`

### Steps to reproduce

```
jay@nautilus-debian:~$ tailscale status
100.89.187.98   nautilus-deb-test    jstaplet@    linux   -
100.117.133.12  ip-10-0-0-114.domain1.beta.tailscale.net jay@         linux   -
```

Then shared nautilus-deb-test back to domain1, then

```
jay@nautilus-debian:~$ tailscale status
100.89.187.98   nautilus-deb-test    jstaplet@    linux   -
jay@nautilus-debian:~$ tailscale ping 100.117.133.12
pong from device-of-shared-to-user (100.117.133.12) via DERP(ord) in 137ms
pong from device-of-shared-to-user (100.117.133.12) via DERP(ord) in 51ms
```

Remove the share *to* <domain1> from nautilus-deb-test, and the node shared *from* domain1 are once again visible in `tailscale status`

This does not happen for every bidirectional sharing. 

### Tailscale version

1.20, 1.19, 1.18 tested

### Bug report

BUG-2492d4547400cf7c4d2bf08350f6bcc990cf998d12390462fefd1f1c3c269eb5-20220119185350Z-15ac31a90257d9d0

JayWStapleton:
BUG-2492d4547400cf7c4d2bf08350f6bcc990cf998d12390462fefd1f1c3c269eb5-20220119194536Z-771d220a6437de19 domain1

BUG-993dda06e35b03cc437fc71494d430affa92cded991e38e3903637d62894ada4-20220119194524Z-9b809bad0e3c5f14
domain2

bradfitz:
Dup of #3756?

JayWStapleton:
Might be the same root cause. I tested on linux systems. The initial notification came from a user with Windows on both nodes. This only seems to appear where there are nodes shared both ways between two tailnets.

Delikt:
i have the same issue, after i share bi-directional the second device between two User Accounts, the output of tailscale status is just 1 Device (the own one) but before while i just shared the first device tailscale status shows me all two devices

Kernel: x86_64 Linux 5.13.0-44-generic
OS: Linux Mint 20.3 Una
Tailscale version: 1.24.2 (on both nodes the same specification)

[...]

PR #5050, opened 2022-07-13
ipn/ipnlocal: put DNS suffixes of shared nodes as split DNS routes

bradfitz:
Updates #3756
Updates #3767


mihaip:
Superseded by https://github.com/tailscale/corp/pull/6074, which fixes this on the control/server side.

PR #5051, opened 2022-07-13
net/dns: add Windows group policy notifications to the NRPT rule manager

dblohm7:
As discussed in previous PRs, we can register for notifications when group
policies are updated and act accordingly.

This patch changes nrptRuleDatabase to receive notifications that group policy
has changed and automatically move our NRPT rules between the local and
group policy subkeys as needed.

### DIFF
--- a/net/dns/manager_windows_test.go
+++ b/net/dns/manager_windows_test.go
@@ -5,6 +5,7 @@
 package dns
 
 import (
+	"context"
 	"fmt"
 	"math/rand"
 	"strings"
@@ -19,11 +20,6 @@ import (
 )
 
 const testGPRuleID = "{7B1B6151-84E6-41A3-8967-62F7F7B45687}"
-
-var (
-	procRegisterGPNotification   = libUserenv.NewProc("RegisterGPNotification")
-	procUnregisterGPNotification = libUserenv.NewProc("UnregisterGPNotification")
-)
 
 func TestManagerWindowsLocal(t *testing.T) {
 	if !isWindows10OrBetter() || !winutil.IsCurrentProcessElevated() {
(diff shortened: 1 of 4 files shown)


net/dns/manager_windows_test.go
```diff
@@ -332,6 +435,37 @@ func getSavedDomainsForRule(base, ruleID string) ([]string, error) {
 	return result, err
 }
 
+func genRandomSubdomains(t *testing.T, n int) []dnsname.FQDN {
```
sailorfrag, line 438: I'm generally a fan of avoiding randomly generated test cases unless it uses a fixed seed or displays the seed because sometimes it results in flakes if certain inputs fail but you can't reproduce what the test did.

net/dns/manager_windows_test.go
```diff
@@ -384,3 +518,97 @@ func (trk *gpNotificationTracker) Close() error {
...
+		windows.REG_NOTIFY_CHANGE_NAME|windows.REG_NOTIFY_THREAD_AGNOSTIC, rw.evtGP, true)
+}
+
+func (rw *regKeyWatcher) wait() error {
+	handles := []windows.Handle{
+		rw.evtLocal,
+		rw.evtGP,
+	}
+
+	waitCode, err := windows.WaitForMultipleObjects(
```
sailorfrag, line 597: would it be good to add some sort of timeout (ideally derived from a context but I'm not picky)

sailorfrag (approved):

net/dns/manager_windows_test.go
dblohm7: Using a context would add a lot more complexity here (due to wiring in its done channel), which I'd take care of if this wasn't test code. In this case I'll just set the timeout directly.

net/dns/manager_windows_test.go
dblohm7: I don't expect too many problems here given that I'm limiting the length of the generated text and restricting it to the lowercase English alphabet, but I see what you're getting at. I'll log the seed.

PR #5056, opened 2022-07-14
net/netcheck: try ICMP if UDP is blocked

andrew-d:
In addition to trying to guess the latency via HTTP(S), we can send an ICMP echo request to each DERP region if UDP is blocked. Currently this is only using IPv4 ICMP, but we should be able to add IPv6 support to this without too much trouble.

Feedback appreciated; this is my first time looking at the `net/netcheck` code here.


net/ping/ping.go
```diff
@@ -0,0 +1,230 @@
...
+//
+// A new instance should be created for each concurrent set of ping requests;
+// this type should not be reused.
+type Pinger struct {
+	c       net.PacketConn
+	Logf    logger.Logf
+	Verbose bool
+	timeNow func() time.Time
+	id      uint16 // uint16 per RFC 792
+	seq     uint32 // uint16 per RFC 792, but we don't have an atomic function; use uint32
```
bradfitz, line 40: I might just not use atomics here and put these under the `mu sync.Mutex`. Is it really hot enough to matter?

bradfitz (commented):
fun!

crawshaw:
Tangential thought: we discussed STUN to anycast addresses to guess at a local DERP (alongside STUNing a handful of core servers like 1 US, 1 EU, etc) and including the DERP region in the return packet. If UDP is blocked we could similarly send an ICMP echo to the anycast address, and report the DERP region in the echo reply payload.

net/netcheck/netcheck.go
```diff
@@ -893,6 +894,48 @@ func (c *Client) GetReport(ctx context.Context, dm *tailcfg.DERPMap) (_ *Report,
 	}
 	rs.stopTimers()
 
+	// Try ICMP latency to determine DERP latency if we don't get any UDP
+	// replies; we'll try HTTP as well below, but ICMP is more reliable.
```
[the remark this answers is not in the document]
andrew-d: I took a stab at this in the second commit I just pushed; is that what you mean?

net/ping/ping.go
```diff
@@ -0,0 +1,233 @@
...
+	seq := p.seq
+	p.mu.Unlock()
+
+	m := icmp.Message{
+		Type: ipv4.ICMPTypeEcho,
+		Code: 0,
+		Body: &icmp.Echo{
+			ID:   int(p.id),
+			Seq:  int(seq),
+			Data: []byte(""),
```
sailorfrag, line 200: Optional, fine as a follow on: there's only 16 bits of ID space, it might be good to also include some payload (the public disco key maybe?) and compare that matches.

net/ping/ping.go
```diff
@@ -0,0 +1,233 @@
...
+	if ch != nil {
+		ch <- response{t: now}
+	}
+}
+
+func (p *Pinger) Send(ctx context.Context, dest net.Addr) (time.Duration, error) {
+	// Use sequential sequence numbers on the assumption that we will not
+	// wrap around when using a single Pinger instance
+	p.mu.Lock()
+	p.seq++
```
sailorfrag, line 190: optional: a random sequence might help with the ID space issue

net/ping/ping.go
```diff
@@ -0,0 +1,233 @@
...
+	conn, err := netns.Listener(logf).ListenPacket(ctx, "ip4:icmp", "0.0.0.0")
+	if err != nil {
+		return nil, err
+	}
+
+	return &Pinger{
+		c:       conn,
+		Logf:    logf,
+		timeNow: time.Now,
+		id:      uint16(id[0]) | uint16(id[1])<<8,
```
sailorfrag, line 85: optional: This looks correct but `id: binary.LittleEndian.Uint16(id[:])` is probably more typo-proof.

net/ping/ping.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	p.wg.Add(1)
+	go p.run(ctx)
+	return p, nil
+}
+
+func newUnstarted(ctx context.Context, logf logger.Logf) (*Pinger, error) {
+	var id [2]byte
+	_, err := rand.Read(id[:])
```
sailorfrag, line 71: optional: it would be a bit better for testing purposes to make it possible to pass in a random number source rather than using the system random number generator. As is, it's easy enough for a test to just override `id` after, so it's not a big problem.

sailorfrag (approved):

net/ping/ping.go
andrew-d: Good call on this; I updated this to use the (unique) `Name` field of the DERP server as the packet's data, and took this opportunity to flesh out tests a bunch.

net/netcheck/netcheck.go
```diff
@@ -905,6 +907,14 @@ func (c *Client) GetReport(ctx context.Context, dm *tailcfg.DERPMap) (_ *Report,
...
 	rs.stopTimers()
 
+	// Try ICMP latency to determine DERP latency if we don't get any UDP
+	// replies; we'll try HTTP as well below.
+	if !rs.anyUDP() && ctx.Err() == nil {
+		if err := c.measureAllICMPLatency(ctx, rs, dm); err != nil {
+			c.logf("measureAllICMPLatency: %v", err)
+		}
+	}
+
```
bradfitz, line 917: per chat: do this in parallel with the HTTPS check

net/ping/ping.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+}
+
+// Send sends an ICMP Echo Request packet to the destination, waits for a
+// response, and returns the duration between when the request was sent and
+// when the reply was received.
+//
+// If provided, "data" is sent with the packet and is compared upon receiving a
+// reply.
+func (p *Pinger) Send(ctx context.Context, dest net.Addr, data []byte) (time.Duration, error) {
```
bradfitz, line 206: use `netip.AddrPort`

net/netcheck/netcheck.go
```diff
@@ -1086,12 +1096,86 @@ func (c *Client) measureHTTPSLatency(ctx context.Context, reg *tailcfg.DERPRegio
...
+	for _, reg := range need {
+		go func(reg *tailcfg.DERPRegion) {
+			defer wg.Done()
+			if d, err := c.measureICMPLatency(ctx, reg, p); err != nil {
+				c.logf("[v1] netcheck: measuring ICMP latency of %v (%d): %v", reg.RegionCode, reg.RegionID, err)
+			} else {
+				c.logf("[v1] netcheck: ICMP latency of %v (%d): %v", reg.RegionCode, reg.RegionID, d)
+				rs.mu.Lock()
+				rs.report.RegionLatency[reg.RegionID] = d
```
bradfitz, line 1127: is this map non-nil already?

If unsure, you can use `mak.Set(&rs.report.RegionLatency, reg.RegionID, d)`

net/netcheck/netcheck.go
```diff
@@ -1086,12 +1096,86 @@ func (c *Client) measureHTTPSLatency(ctx context.Context, reg *tailcfg.DERPRegio
...
+	//
+	// TODO(andrew-d): this is a bit ugly
+	nodeAddr := c.nodeAddr(ctx, node, probeIPv4)
+	if !nodeAddr.IsValid() {
+		return 0, fmt.Errorf("no address for node %v", node.Name)
+	}
+	udpAddr := net.UDPAddrFromAddrPort(nodeAddr)
+
+	addr := &net.IPAddr{
+		IP:   udpAddr.IP,
```
bradfitz, line 1161: net.IP(nodeAddr.Addr().AsSlice())

net/netcheck/netcheck.go
```diff
@@ -1086,12 +1096,86 @@ func (c *Client) measureHTTPSLatency(ctx context.Context, reg *tailcfg.DERPRegio
...
+		Zone: udpAddr.Zone,
```
bradfitz, line 1162: nodeAddr.Addr().Zone()

bradfitz (approved):

net/ping/ping.go
sailorfrag: I'm thinking of colliding with unrelated pings originating from the same host. They tend to start at 0 so if we don't, we're more likely to miss those.

Actually, we'd greatly reduce the risk of collisions with pings from other processes by checking that the reply's source IP matches where we sent the ping to.

sailorfrag: ... the hostname of the derp server in the payload probably suffices to distinguish the responses anyway

Issue #5028, opened 2022-07-09
FR: Support ability to send directory using CLI.


stefins:
### What are you trying to do?

Currently It is not possible to send directory using the `tailscale` cli

### How should we solve this?

We can zip the directory and treat it as a file.

PR #5063, opened 2022-07-15
cmd/tailscale: add ability to send directory via CLI

stefins:
Hi,

This PR solves #5028.

Before this `tailscale` was unable to send directories through the CLI. It raised an `"directories not supported"` error.

After this a directory can be send by
`tailscale file cp directory_name target_ip:` 

Thanks :)


stefins:
any plan of reviewing this PR :)

cmd/tailscale/cli/file.go
```diff
@@ -140,7 +141,17 @@ func runCp(ctx context.Context, args []string) error {
 				return err
 			}
 			if fi.IsDir() {
-				return errors.New("directories not supported")
+				name = fi.Name() + ".zip"
```
DentonGentry, line 144: We wouldn't want to leave the zip file in their filesystem, and ideally it would work by streaming the data from the ZipWriter without writing out a file at all.

stefins: Sure I'm looking into it.

stefins:
@DentonGentry Is this what you had in mind?

Issue #5029, opened 2022-07-09
tstun.New("tailscale0"): device or resource busy


dragon2611:
### What is the issue?

After an apt upgrade (which upgraded from 1.26.1)

```
tailscaled
logtail started
Program starting: v1.26.2-t9e47908f1-g27ee7cf2f, Go 1.18.3-ts04d67b90d8: []string{"tailscaled"}
LogID: a784de426e271c29e54c8c68876a80e59afc8880a3e466611e34fc7d61cbcb7f
logpolicy: using system state directory "/var/lib/tailscale"
wgengine.NewUserspaceEngine(tun "tailscale0") ...
Linux kernel version: 5.10.0-15-amd64
'modprobe tun' successful
/dev/net/tun: Dcrw-rw-rw-
wgengine.NewUserspaceEngine(tun "tailscale0") error: tstun.New("tailscale0"): device or resource busy
flushing log.
logger closing down
createEngine: tstun.New("tailscale0"): device or resource busy
```

### Steps to reproduce

Upgrade from 1.26.1 on debian

### OS version

debian 11

### Tailscale version

1.26.2

Note: I manually started tailscaled in this instance to debug why the service was crashing.

dragon2611:
Deleting the tailscale0 interface `ip link delete tailscale0` fixes but you have to be able to get into the machine outside of tailscale first to do that.

DentonGentry:
Can you provide either a "tailscale bugreport" r the IP address of the node where this happened so we can look into it.

dragon2611:
`BUG-3351824445b319768ac4de53e892c32e8ab5faa2338e548fb098a901c52ac24c-20220709171609Z-e95dfb48df9c85e5`

DentonGentry:
Unfortunately there isn't anything suspicious in the telemetry.

It seems like the tun device in the kernel just got stuck.

```
2022-07-09 16:27:53.903281437 +0000 UTC: Program starting: v1.26.2-t9e47908f1-g27ee7cf2f, Go 1.18.3-ts04d67b90d8: []string{"/usr/sbin/tailscaled", "--cleanup"}
2022-07-09 16:27:54.535590941 +0000 UTC: Program starting: v1.26.2-t9e47908f1-g27ee7cf2f, Go 1.18.3-ts04d67b90d8: []string{"/usr/sbin/tailscaled", "--cleanup"}
[...]

PR #5069, opened 2022-07-16
net/tstun: diagnose /dev/net/tun fd leak, give better failure message

bradfitz:
Updates #5029

### DIFF
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -515,7 +515,7 @@ func tryEngine(logf logger.Logf, linkMon *monitor.Mon, dialer *tsdial.Dialer, na
 	} else {
 		dev, devName, err := tstun.New(logf, name)
 		if err != nil {
-			tstun.Diagnose(logf, name)
+			tstun.Diagnose(logf, name, err)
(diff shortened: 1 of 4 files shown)


DentonGentry (approved):

bradfitz:
Demo, repro'ed from running v1.26.0 to get into the bad state:

```
logpolicy: using system state directory "/var/lib/tailscale"
SOCKS5 listening on 127.0.0.1:40869
wgengine.NewUserspaceEngine(tun "tailscale0") ...
TUN device tailscale0 is busy; another process probably still has it open (from old version of Tailscale that had a bug)
To fix, kill the process that has it open. Find with:

$ sudo lsof -n /dev/net/tun

... and then kill those PID(s)
wgengine.NewUserspaceEngine(tun "tailscale0") error: tstun.New("tailscale0"): device or resource busy
flushing log.
logger closing down
createEngine: tstun.New("tailscale0"): device or resource busy


root@tsdev:~# sudo lsof -n /dev/net/tun
COMMAND       PID     USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
mosh-serv 3506677 bradfitz   10u   CHR 10,200    0t111  278 /dev/net/tun
bash      3506678 bradfitz   10u   CHR 10,200    0t111  278 /dev/net/tun
```

PR #5071, opened 2022-07-17
cmd/tailscale/cli/web: add QNAP NAS_SID authentication

DentonGentry:
QTS 5.0 doesn't always pass a qtoken, in some circumstances
it sends a NAS_SID cookie for us to verify instead.


bradfitz (approved):
Get "web" and "QNAP" into title and drop final period?

I guess you were able to test this? Use our two QNAP devices or?

DentonGentry:
Updated title.

I bought a QNAP TS-231K to work with.

PR #5073, opened 2022-07-17
QNAP: detect QTS version, support Taildrop folder

DentonGentry:


hostinfo/hostinfo_linux.go
```diff
@@ -48,6 +49,15 @@ func linuxDeviceModel() string {
 	return ""
 }
 
+func getQnapQtsVersion(version_info []byte) string {
```
maisem, line 52: Doesn't follow go naming guidelines.

ipn/ipnserver/server.go
```diff
@@ -1163,6 +1165,26 @@ func findTrueNASTaildropDir(name string) (dir string, err error) {
...
+		if fi.IsDir() {
+			return dir, nil
+		}
+
+		// share/Taildrop is usually a symlink to CACHEDEV1_DATA/Taildrop/ or some such.
+		fullpath, err := filepath.EvalSymlinks(dir)
+		if err != nil {
+			return "", fmt.Errorf("symlink to shared folder %q not found", name)
+		}
+		if fi, err = os.Stat(fullpath); err == nil && fi.IsDir() {
```
maisem, line 1181: `:=`

maisem (approved):

hostinfo/hostinfo_linux.go
```diff
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"regexp"
```
bradfitz, line 15: I guess this is okay on Linux but we've tried to avoid this package because it brings in huge Unicode tables and makes binaries bigger.

But iOS now brings it in any for UPnP so maybe it's time to accept it.

hostinfo/hostinfo_linux.go
```diff
@@ -48,6 +49,15 @@ func linuxDeviceModel() string {
 	return ""
 }
 
+func getQnapQtsVersion(version_info []byte) string {
+	re := regexp.MustCompile(`QTSFW_([\d.]+)`)
+	match := re.FindSubmatch(version_info)
```
bradfitz, line 54: FindStringSubmatch

hostinfo/hostinfo_linux.go
```diff
@@ -48,6 +49,15 @@ func linuxDeviceModel() string {
...
+	if len(match) == 0 {
```
bradfitz, line 55: matching nil is more convention for the regexp package callers.

Its docs do say:

> A return value of nil indicates no match.

hostinfo/hostinfo_linux.go
DentonGentry: looking into this

ipn/ipnserver/server.go
DentonGentry: Done

hostinfo/hostinfo_linux.go
```diff
@@ -48,6 +49,15 @@ func linuxDeviceModel() string {
 	return ""
 }
 
+func getQnapQtsVersion(version_info []byte) string {
+	re := regexp.MustCompile(`QTSFW_([\d.]+)`)
```
bradfitz, line 53: We shouldn't compile on each call. Compiling the regexp is quite heavy.

OTOH, we shouldn't compile it in init either, ideally, especially as most systems aren't QNAP.

I see that GetOSVersion isn't cached (except on some platforms that call SetOSVersion) so that means this will be run on every call to hostinfo.New, which I think is often enough. (new map calls, etc)

Maybe we should add some caching to osVersion Linux at least. It's not super cheap currently in any case.

Not your cleanup, though. You can just add a TODO here, like `// TODO(bradfitz,dgentry): cache this or caller(s)`

ipn/ipnserver/server.go
```diff
@@ -1163,6 +1165,26 @@ func findTrueNASTaildropDir(name string) (dir string, err error) {
 	return "", fmt.Errorf("shared folder %q not found", name)
 }
 
+// findQnapTaildropDir checks if a Shared Folder named "Taildrop" exists.
+func findQnapTaildropDir(name string) (string, error) {
+	dir := fmt.Sprintf("/share/%s", name)
+	if fi, err := os.Stat(dir); err == nil {
```
bradfitz, line 1171: invert and early return & outdent rest

hostinfo/hostinfo_linux.go
maisem: https://zchee.github.io/golang-wiki/CodeReviewComments/

hostinfo/hostinfo_linux.go
DentonGentry: Made version_info be a string and switched to FindStringSubmatch

hostinfo/hostinfo_linux.go
DentonGentry: Made the regex compile once the first time it is used, added a TODO about caching at a higher layer.

ipn/ipnserver/server.go
DentonGentry: Done

hostinfo/hostinfo_linux.go
DentonGentry: Done. I thought this referred to the function name, just realized it referred to `version_info`

hostinfo/hostinfo_linux.go
DentonGentry: Replaced use of regex, we're looking for one fixed pattern.

hostinfo/hostinfo_linux.go
DentonGentry: Replaced use of regex

hostinfo/hostinfo_linux.go
```diff
@@ -48,7 +50,22 @@ func linuxDeviceModel() string {
 	return ""
 }
 
+var qnapRegexOnce sync.Once
+var qnapRegex *regexp.Regexp
+
+func getQnapQtsVersion(version_info string) string {
```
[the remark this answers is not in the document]
DentonGentry: Done

hostinfo/hostinfo_linux.go
```diff
@@ -48,7 +50,22 @@ func linuxDeviceModel() string {
 	return ""
 }
 
+var qnapRegexOnce sync.Once
```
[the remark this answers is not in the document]
DentonGentry: Replaced use of regex.

hostinfo/hostinfo_linux.go
maisem: Ah yeah, I should have been clearer.

hostinfo/hostinfo_linux.go
```diff
@@ -48,7 +48,17 @@ func linuxDeviceModel() string {
 	return ""
 }
 
+func getQnapQtsVersion(versionInfo string) string {
+	for _, field := range strings.Fields(versionInfo) {
+		if strings.HasPrefix(field, "QTSFW_") {
+			return "QTS " + field[6:]
```
bradfitz, line 54: len("QTSFW_") also compiles to 6

or:

     if suffix := strings.TrimPrefix(field, "QTSFW_"); suffix != field {
             return "QTS "+suffix
     }

hostinfo/hostinfo_linux_test.go
```diff
@@ -0,0 +1,38 @@
...
+	version_info := `commit 2910d3a594b068024ed01a64a0fe4168cb001a12
+Date:   2022-05-30 16:08:45 +0800
+================================================
+* QTSFW_5.0.0
+remotes/origin/QTSFW_5.0.0`
+
+	got := getQnapQtsVersion(version_info)
+	want := "QTS 5.0.0"
+	if got != want {
+		t.Errorf("got %v; want %v", got, want)
```
bradfitz, line 24: use %q for failing strings. %q shows empty, trailing whitespace, binary escaped, etc.

(here and elsewhere)

bradfitz (approved):

hostinfo/hostinfo_linux_test.go
DentonGentry: Done.

hostinfo/hostinfo_linux.go
DentonGentry: Went with TrimPrefix, very nice.

PR #5083, opened 2022-07-18
derp: plumb '/derp' request context through

catzkorn:
This change is required to implement tracing for derp.


derp/derp_test.go
```diff
@@ -475,7 +475,7 @@ func newTestServer(t *testing.T) *testServer {
 			// TODO: register c in ts so Close also closes it?
 			go func(i int) {
 				brwServer := bufio.NewReadWriter(bufio.NewReader(c), bufio.NewWriter(c))
-				go s.Accept(c, brwServer, fmt.Sprintf("test-client-%d", i))
+				go s.Accept(context.Background(), c, brwServer, fmt.Sprintf("test-client-%d", i))
```
maisem, line 478: this should be something scoped to the test. So maybe do something like at the start of the test and then use `ctx` here.
```
defer cancel()
```

bradfitz (approved):

Issue #3160, opened 2021-10-21
health: add health check item if DNS server doesn't seem reachable


bradfitz:
If somebody sets their global DNS server, to, say 192.168.1.1 and that IP address isn't reachable from a certain node, we should detect that and add it as a health check item (so it shows up in "tailscale status", etc) that their DNS config is not viable.

/cc @danderson

DentonGentry:
Draft pull request in https://github.com/tailscale/tailscale/pull/5086 would fix this

PR #5086, opened 2022-07-18
net/dns: add healthcheck for upstream DNS servers

andrew-d:
Per the suggestion in #3160, query `ipv4only.arpa` to see if the upstream DNS server(s) are functional and include that in the health status. Marked as WIP since there's some TODOs that I would appreciate feedback on. More generally, are we comfortable merging this as-is or should we put this in some form of "monitor" mode where we report metrics but don't update the overall health status until we're confident this works?

### DIFF
--- a/health/health.go
+++ b/health/health.go
@@ -73,6 +73,9 @@ const (
 
 	// SysTKA is the name of the tailnet key authority subsystem.
 	SysTKA = Subsystem("tailnet-lock")
+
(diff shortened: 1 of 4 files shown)


mnordhoff:
What happens if IANA has a DNS outage but the rest of the DNS works?

andrew-d:
Per the suggestion from @bradfitz, we now only healthcheck if we notice that a query timed out (defined as a context deadline expiring or receiving an i/o deadline exceeded error).

PR #5090, opened 2022-07-19
ipn/ipnlocal: logout ephemeral nodes on shutdown

maisem:

### DIFF
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -404,7 +404,6 @@ func run() error {
 	// want to keep running.
 	signal.Ignore(syscall.SIGPIPE)
 	go func() {
-		defer dialer.Close()
 		select {
 		case s := <-interrupt:
 			logf("tailscaled got signal %v; shutting down", s)
@@ -437,6 +436,7 @@ func run() error {
 	if err != nil {
 		return fmt.Errorf("safesocket.Listen: %v", err)
 	}
+	defer dialer.Close()
 
 	err = srv.Run(ctx, ln)
 	// Cancelation is not an error: it is the only way to stop ipnserver.
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -342,7 +342,19 @@ func (b *LocalBackend) onHealthChange(sys health.Subsystem, err error) {
 // can no longer be used after Shutdown returns.
 func (b *LocalBackend) Shutdown() {
 	b.mu.Lock()
+	if b.shutdownCalled {
+		b.mu.Unlock()
+		return
+	}
 	b.shutdownCalled = true
+
+	if b.loginFlags&controlclient.LoginEphemeral != 0 {
+		b.mu.Unlock()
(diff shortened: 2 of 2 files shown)


ipn/ipnlocal/local.go
```diff
@@ -342,7 +342,17 @@ func (b *LocalBackend) onHealthChange(sys health.Subsystem, err error) {
...
 	b.mu.Lock()
+	if b.shutdownCalled {
+		b.mu.Unlock()
+		return
+	}
 	b.shutdownCalled = true
+
+	if b.loginFlags&controlclient.LoginEphemeral != 0 {
+		b.mu.Unlock()
+		b.logout(b.ctx, true) // best effort
```
bradfitz, line 353: can we bound the time here? Or start it async in its own goroutine?

Not a fan of blocking Shutdown with network work.

ipn/ipnlocal/local.go
```diff
@@ -342,7 +342,17 @@ func (b *LocalBackend) onHealthChange(sys health.Subsystem, err error) {
...
+		b.mu.Lock()
```
bradfitz, line 354: any invariants you need to re-check now that you've dropped the lock and reacquired it?

ipn/ipnlocal/local.go
maisem: Time bound sgtm
Running a different goroutine doesn't help shutdown

ipn/ipnlocal/local.go
maisem: Nothing can restart the backend, so no.

bradfitz (approved):

ipn/ipnlocal/local.go
```diff
@@ -342,7 +342,19 @@ func (b *LocalBackend) onHealthChange(sys health.Subsystem, err error) {
...
+		b.mu.Unlock()
+		return
+	}
 	b.shutdownCalled = true
+
+	if b.loginFlags&controlclient.LoginEphemeral != 0 {
+		b.mu.Unlock()
+		ctx, cancel := context.WithTimeout(b.ctx, 5*time.Second)
+		defer cancel()
+		b.logout(ctx, true) // best effort
```
mihaip, line 355: May as well use `LogoutSync` so the reader doesn't have to figure out what the boolean arg does?

cmd/tailscaled/tailscaled.go
```diff
@@ -437,6 +436,7 @@ func run() error {
 	if err != nil {
 		return fmt.Errorf("safesocket.Listen: %v", err)
 	}
+	defer dialer.Close()
```
mihaip, line 439: Out of curiosity, why this change?

mihaip (approved):

ipn/ipnlocal/local.go
maisem: Good idea, done.

cmd/tailscaled/tailscaled.go
maisem: It was being closed too early, which resulted in the call to LogoutSync not being able to make the network request to control.

PR #5093, opened 2022-07-19
ssh/tailssh: add password-forcing workaround for buggy SSH clients

bradfitz:
If the username includes a suffix of +password, then we accept
password auth and just let them in like it were no auth.

This exists purely for SSH clients that get confused by seeing success
to their initial auth type "none".

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -48,6 +48,13 @@ var (
 	sshVerboseLogging = envknob.RegisterBool("TS_DEBUG_SSH_VLOG")
 )
 
+const (
+	// forcePasswordSuffix is the suffix at the end of a username that forces
+	// Tailscale SSH into password authentication mode to work around buggy SSH
+	// clients that get confused by successful replies to auth type "none".
+	forcePasswordSuffix = "+password"
+)
+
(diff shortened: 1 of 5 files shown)


maisem:
>If the username includes a suffix of +pass or +password,
It only works with `+password`

bradfitz:
I'd updated the commit message but not the PR.

ssh/tailssh/tailssh.go
```diff
@@ -197,6 +204,7 @@ type conn struct {
 	idH    string
 	connID string // ID that's shared with control
 
+	anyPasswordIsOkay       bool  // set by NoClientAuthCallback; whether any password is okay for this connection
```
maisem, line 207: is noPubKeyPolicyAuthError no longer used?

bradfitz: I didn't touch that?

bradfitz: deleted.

maisem: sent https://github.com/tailscale/tailscale/pull/5898

maisem (approved):

ssh/tailssh/tailssh.go
maisem: lol that was a race. I'm going to close that PR

PR #5097, opened 2022-07-19
types/key: add naclbox shared key wrapper type + Seal method

bradfitz:
So the control plane can stop doing precomputations on each naclbox
message.


types/key/machine.go
```diff
@@ -105,6 +105,33 @@ func (k MachinePrivate) SealTo(p MachinePublic, cleartext []byte) (ciphertext []
 	return box.Seal(nonce[:], cleartext, &nonce, &p.k, &k.k)
 }
 
+// Seal wraps cleartext into a NaCl box (see
+// golang.org/x/crypto/nacl) using the shared key k as generated
+// by MachinePrivate.SharedKey.
+//
+// The returned ciphertext is a 24-byte nonce concatenated with the
+// box value.
+func (k MachinePrecomputedSharedKey) Seal(cleartext []byte) (ciphertext []byte) {
```
danderson, line 114: Nit: put the function after the type declaration + its "constructor" func?

danderson (approved):

PR #5101, opened 2022-07-20
cmd/tsconnect: allow building static resources in a different directory

mihaip:
When using tsconnect as a module in another repo, we cannot write to
the ./dist directory (modules directories are read-only by default -
there is a -modcacherw flag for `go get` but we can't count on it).

We add a -distdir flag that is honored by both the build and serve
commands for where to place output in.

Somewhat tedious because esbuild outputs paths relative to the working
directory, so we need to do some extra munging to make them relative
to the output directory.

### DIFF
--- a/cmd/tsconnect/common.go
+++ b/cmd/tsconnect/common.go
@@ -38,7 +38,7 @@ func commonSetup(dev bool) (*esbuild.BuildOptions, error) {
 	return &esbuild.BuildOptions{
 		EntryPoints: []string{"src/index.js", "src/index.css"},
 		Loader:      map[string]esbuild.Loader{".wasm": esbuild.LoaderFile},
-		Outdir:      "./dist",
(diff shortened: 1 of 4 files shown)


cmd/tsconnect/common.go
```diff
@@ -103,3 +103,12 @@ func installJSDeps() error {
 	}
 	return err
 }
+
+// EsbuildMetadata is the subset of metadata struct (described by
+// https://esbuild.github.io/api/#metafile) that we care about for mapping
+// from entry points to hashed file names.
+type EsbuildMetadata = struct {
```
bradfitz, line 110: no = sign

bradfitz (approved):

cmd/tsconnect/common.go
mihaip: Too much TypeScript muscle memory.

PR #5109, opened 2022-07-21
types/views: add SliceContains, View.ContainsFunc, View.IndexFunc

bradfitz:
We were starting to write these elsewhere as little unexported copies
in misc places.

### DIFF
--- a/types/views/views.go
+++ b/types/views/views.go
@@ -132,6 +132,43 @@ func (v Slice[T]) AsSlice() []T {
 	return v.AppendTo(v.ж[:0:0])
 }
 
+// IndexFunc returns the first index of an element in v satisfying f(e),
+// or -1 if none do.
+//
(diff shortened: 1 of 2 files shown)


types/views/views.go
```diff
@@ -132,6 +132,39 @@ func (v Slice[T]) AsSlice() []T {
...
+			return true
+		}
+	}
+	return false
+}
+
+// SliceContains reports whether v contains element e.
+//
+// As it runs in O(n) time, use with care.
+func SliceContains[T comparable](v Slice[T], e T) bool {
```
maisem, line 159: Put this on the Slice itself?

bradfitz: can't. slices aren't necessarily comparable and can't add constraints on a method: https://go.dev/play/p/pNj8OAApOGt

maisem (approved):

PR #5113, opened 2022-07-21
cmd/gitops-pusher: port to use ffcli

Xe:
This is getting more complicated, so I want to port this to use ffcli like other commands instead of yoloing a command line framework. 

### DIFF
--- a/cmd/gitops-pusher/gitops-pusher.go
+++ b/cmd/gitops-pusher/gitops-pusher.go
@@ -20,20 +20,92 @@ import (
 	"strings"
 	"time"
 
+	"github.com/peterbourgon/ff/v3/ffcli"
 	"github.com/tailscale/hujson"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/gitops-pusher/gitops-pusher.go
```diff
@@ -20,79 +20,142 @@ import (
...
+			return err
 		}
 
-	case "test":
+		return nil
+	}
+}
+
+func test(tailnet, apiKey string) func(context.Context, []string) error {
+	return func(ctx context.Context, args []string) error {
```
Soypete, line 63: am I understing this correctly? 
```ffcli.Command.Exec()``` required a function of the form ```func (context.Context, []string)``` but args are not actually be used by any of these functions?

Soypete (approved):

PR #5115, opened 2022-07-21
cmd/ssh-auth-none-demo: add SSH demo server w/ auth type none

bradfitz:
Change-Id: Ief8b1d904a4304ddb72e76d52fa8c50871c704f6


bradfitz:
Whoops, I forgot about this and sent/merged #5886 instead.

PR #5116, opened 2022-07-21
tka,types/key: implement direct node-key signatures

twitchyliquid64:
Am I holding `key.NodePublic` right?

Can I use `Raw32()` & `NodePublicFromRaw32()` and avoid the prefix overhead?

### DIFF
--- a/tka/sig.go
+++ b/tka/sig.go
@@ -1,0 +1,99 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package tka
+
+import (
+	"bytes"
+	"crypto/ed25519"
+	"errors"
+	"fmt"
+
+	"github.com/fxamacker/cbor/v2"
+	"github.com/hdevalence/ed25519consensus"
+	"golang.org/x/crypto/blake2s"
+)
+
+// SigKind describes valid NodeKeySignature types.
+type SigKind uint8
+
+const (
+	SigInvalid SigKind = iota
+	// SigDirect describes a signature over a specific node key, using
+	// the keyID specified.
+	SigDirect
+)
+
+func (s SigKind) String() string {
+	switch s {
+	case SigInvalid:
+		return "invalid"
+	case SigDirect:
+		return "direct"
+	default:
+		return fmt.Sprintf("Sig?<%d>", int(s))
+	}
+}
+
+// NodeKeySignature encapsulates a signature that authorizes a specific
+// node key, based on verification from keys in the tailnet key authority.
+type NodeKeySignature struct {
+	// SigKind identifies the variety of signature.
+	SigKind SigKind `cbor:"1,keyasint"`
+	// Pubkey identifies the public key which is being certified.
+	Pubkey []byte `cbor:"2,keyasint"`
+
+	// KeyID identifies which key in the tailnet key authority should
+	// be used to verify this signature. Only set for SigDirect and
+	// SigCredential signature kinds.
+	KeyID []byte `cbor:"3,keyasint,omitempty"`
+
+	// Signature is the packed (R, S) ed25519 signature over the rest
+	// of the structure.
+	Signature []byte `cbor:"4,keyasint,omitempty"`
+}
+
+// sigHash returns the cryptographic digest which a signature
+// is over.
+//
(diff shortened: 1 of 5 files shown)


danderson:
What is the concern with prefix overhead? The purpose of having typed strings is to help prevent errors where a string key gets deserialized into the wrong type - defeating the help Go's type system is trying to provide to prevent cross-purpose key reuse. Especially if this is in textual protocols, I'd prefer to preserve that property unless the performance is truly unacceptable, somehow?

tka/sig.go
```diff
@@ -0,0 +1,118 @@
...
+		panic(err)
+	}
+	if err := encoder.NewEncoder(out).Encode(s); err != nil {
+		// Writing to a bytes.Buffer should never fail.
+		panic(err)
+	}
+	return out.Bytes()
+}
+
+// verifySignature returns a nil error if the signature is valid & certifies
```
danderson, line 90: Returning a nil error on success is a common pattern, don't need to spell it out. I'd word as: `verifySignature verifies that s is a valid signature for nodePublic...`

ellipsis because I actually can't tell what verificationKey does here, and how it differs from whatever's already in the signature. Please explain that :)

tka/tka.go
```diff
@@ -543,3 +546,18 @@ func (a *Authority) Inform(updates []AUM) error {
 	a.state = c.state
 	return nil
 }
+
+// NodePermitted returns a nil error if the node key is authorized by the tailnet key
```
danderson, line 550: Same, no need to spell out that a nil error is the success condition, for simple functions that only return an error.

`NodePermitted checks that nodeKeySignature authorizes nodeKey to be a member of the tailnet.`

twitchyliquid64:
It bloats the size of an NKS: 32bytes raw encoding vs 64bytes hex + 8 bytes prefix. Given this is going to need to be sent for every node in the netmap, I can see the overhead adding up fast.

> preserve that property unless the performance is truly unacceptable

Your call: I think Raw32 is acceptable in this case, but happy to fall in line if you think the size thing is no biggie

types/key/node.go
```diff
@@ -265,6 +265,14 @@ func (k NodePublic) Less(other NodePublic) bool {
 	return bytes.Compare(k.k[:], other.k[:]) < 0
 }
 
+// Equal reports whether k is bytewise equal to an other NodePublic.
+//
+// Not that this function can return false for two different encodings
```
danderson, line 270: typo, not->note. But also, I don't understand what this sentence is trying to say. You're comparing one NodePublic to another, there should not be any way to end up with encoding differences, or this package is buggy and needs fixing.

types/key/node.go
```diff
@@ -265,6 +265,14 @@ func (k NodePublic) Less(other NodePublic) bool {
...
+// of the same public key.
+func (k NodePublic) Equal(other NodePublic) bool {
```
danderson, line 272: No need to define Equal methods for `FooPublic` types, they are deliberately designed to be comparable as value types, so you can use `==` and get the right result. `Equal` method exist on private keys to enforce constant-time comparison (and the `FooPrivate` struct definitions contain a hack that make the structs incomparable using `==`).

danderson:
I'm okay with using Raw32 here, because I'd consider the CBOR blobs to be a binary encoding that is an implementation detail of a cryptographic protocol. That's a place where I would _hope_ we're already being extremely careful of making changes (similar to the guts of our noise protocol implementation), so I'm okay with efficiency overruling misuse-resistance in a small way.

twitchyliquid64:
Thinking this some more, I think im okay with a small prefix as long as the payload is not hex-encoded (ie: only 2 bytes overhead). Ive gone ahead & implemented `binaryMarshaler` to provide a more compact encoding while also having a distinguishing prefix.

tka/sig.go
```diff
@@ -0,0 +1,118 @@
...
+type SigKind uint8
+
+const (
+	SigInvalid SigKind = iota
+	// SigDirect describes a signature over a specific node key, using
+	// the keyID specified.
+	SigDirect
+)
+
+func (s SigKind) String() string {
```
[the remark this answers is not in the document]
twitchyliquid64: I get it does then funky string compaction thing, but I'm not too crash hot on how it encodes unknown values. Do we have to use it?

twitchyliquid64:
Re-assigning as Dave is OOO.

crawshaw (approved):

PR #5119, opened 2022-07-22
cmd/tsconnect: pin yarn and node

mihaip:
Adds a tool/yarn helper script that uses specific versions of yarn and
node, downloading them if necessary.

Modeled after tool/go (and the yarn and node Redo scripts from the
corp repo).


tool/yarn
```diff
@@ -0,0 +1,71 @@
...
+    fi
+    if [ "$ARCH" = "aarch64" ]; then
+        ARCH="arm64"
+    fi
+    NODE_URL="https://nodejs.org/dist/v${REV}/node-v${REV}-${OS}-${ARCH}.tar.gz"
+    install_tool "node" $REV $NODE_DIR $NODE_URL
+}
+
+install_yarn() {
+    read -r REV < "$(dirname "$0")/yarn.rev"
```
maisem, line 28: Add the equivalent of SKIP from `./tool/go`? Something that would use whatever is already available in the environment. Otherwise it'll break on nixos

mihaip: Done. Also added a way to specify a custom yarn path when building, so that we can use `tool/yarn` when doing deploys from corp (that wrapper has nixos-specific logic, which I don't want to replicate).

tool/yarn
```diff
@@ -0,0 +1,79 @@
...
+    read -r REV < "$(dirname "$0")/node.rev"
+    NODE_URL="https://nodejs.org/dist/v${REV}/node-v${REV}-${OS}-${ARCH}.tar.gz"
+    install_tool "node" $REV $NODE_DIR $NODE_URL
+}
+
+install_yarn() {
+    YARN_URL="https://github.com/yarnpkg/yarn/releases/download/v$REV/yarn-v$REV.tar.gz"
+    install_tool "yarn" $YARN_REV $YARN_DIR $YARN_URL
+}
+
```
maisem, line 33: The variables aren't correct, it didn't work on my machine until I made these changes.

```suggestion
install_node() {
    read -r NODE_REV < "$(dirname "$0")/node.rev"
    NODE_URL="https://nodejs.org/dist/v${NODE_REV}/node-v${NODE_REV}-${OS}-${ARCH}.tar.gz"
    install_tool "node" $NODE_REV $NODE_DIR $NODE_URL
}

install_yarn() {
    YARN_URL="https://github.com/yarnpkg/yarn/releases/download/v${YARN_REV}/yarn-v${YARN_REV}.tar.gz"
    install_tool "yarn" $YARN_REV $YARN_DIR $YARN_URL
}
```

mihaip: Oops, didn't do an end-to-end test. Only the `install_yarn` side is actually needed, fixed that.

maisem: I only did the NODE_REV stuff to be consistent.

mihaip: Fair enough, and likely safer (I don't know, and don't want to learn bash/shell variable scoping/shadowing rules).

maisem (approved):

tool/yarn
maisem: Neither do I :)

PR #5124, opened 2022-07-22
cmd/gitops-pusher: add etag cache file for the three version problem

Xe:
This allows gitops-pusher to detect external ACL changes. I'm not sure what to call this problem, so I've been calling it the "three version problem" in my notes. The basic problem is that at any given time we only have two versions of the ACL file at any given point: the version in CONTROL and the one in the git repo. In order to check if there has been tampering of the ACL files in the admin panel, we need to have a _third_ version to compare against.

In this case I am not storing the old ACL entirely (though that could be a reasonable thing to add in the future), but only its sha256sum. This allows us to detect if the shasum in control matches the shasum we expect, and if that expectation fails, then we can react accordingly.

This will require additional configuration in CI, but I'm sure that can be done.

### DIFF
--- a/cmd/gitops-pusher/gitops-pusher.go
+++ b/cmd/gitops-pusher/gitops-pusher.go
@@ -27,11 +27,23 @@ import (
 var (
 	rootFlagSet  = flag.NewFlagSet("gitops-pusher", flag.ExitOnError)
 	policyFname  = rootFlagSet.String("policy-file", "./policy.hujson", "filename for policy file")
+	cacheFname   = rootFlagSet.String("cache-file", "./version-cache.json", "filename for the previous known version hash")
 	timeout      = rootFlagSet.Duration("timeout", 5*time.Minute, "timeout for the entire CI run")
 	githubSyntax = rootFlagSet.Bool("github-syntax", true, "use GitHub Action error syntax (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message)")
+
+	modifiedExternallyFailure = make(chan struct{}, 1)
 )
 
-func apply(tailnet, apiKey string) func(context.Context, []string) error {
+func modifiedExternallyError() {
+	if *githubSyntax {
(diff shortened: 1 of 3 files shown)


cmd/gitops-pusher/gitops-pusher.go
```diff
@@ -27,11 +27,23 @@ import (
...
 	timeout      = rootFlagSet.Duration("timeout", 5*time.Minute, "timeout for the entire CI run")
 	githubSyntax = rootFlagSet.Bool("github-syntax", true, "use GitHub Action error syntax (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message)")
+
+	modifiedExternallyFailure = make(chan struct{}, 1)
 )
 
-func apply(tailnet, apiKey string) func(context.Context, []string) error {
+func modifiedExternallyError() {
+	if *githubSyntax {
+		fmt.Printf("::error file=%s,line=1,col=1,title=Policy File Modified Externally::The policy file was modified externally in the admin console.\n", *policyFname)
```
Soypete, line 39: does this error follow a specific github error syntax?

cmd/gitops-pusher/gitops-pusher.go
```diff
@@ -157,6 +206,10 @@ func main() {
 		fmt.Println(err)
 		os.Exit(1)
 	}
+
+	if len(modifiedExternallyFailure) != 0 {
```
Soypete, line 210: modified externally? like the file is updated by a human or github?

Soypete (approved):

cmd/gitops-pusher/gitops-pusher.go
Xe: yes: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message

cmd/gitops-pusher/gitops-pusher.go
Xe: Like the file is updated in the admin panel yeah

PR #5125, opened 2022-07-22
ipn/ipnlocal: ignore empty SSH host key files

bradfitz:


ipn/ipnlocal/ssh.go
```diff
@@ -119,7 +120,7 @@ func (b *LocalBackend) getSystemSSH_HostKeys() (ret []ssh.Signer, err error) {
 	for _, typ := range keyTypes {
 		filename := "/etc/ssh/ssh_host_" + typ + "_key"
 		hostKey, err := ioutil.ReadFile(filename)
```
maisem, line 122: should we also be following symlinks?

maisem (approved):

ipn/ipnlocal/ssh.go
bradfitz: we already do. are you saying should we not?

maisem: Nvm, I thought that was part of the bug. Carry on.

PR #5153, opened 2022-07-22
net/dns: remove systemd-resolved ping

nyghtowl:
Ping not needed and duplicate to confirm systemd-resolved working.


nyghtowl:
@danderson Does this line up to the ping removal we discussed on Monday?

net/dns/manager_linux.go
```diff
@@ -78,18 +78,6 @@ func dnsMode(logf logger.Logf, env newOSConfigEnv) (ret string, err error) {
...
-	// for systemd-resolved. If it's active on the machine, this
-	// will make it start up and write the /etc/resolv.conf file
-	// before it replies to the ping. (see how systemd's
-	// src/resolve/resolved.c calls manager_write_resolv_conf
-	// before the sd_event_loop starts)
-	resolvedUp := env.dbusPing("org.freedesktop.resolve1", "/org/freedesktop/resolve1") == nil
-	if resolvedUp {
-		dbg("resolved-ping", "yes")
-	}
-
```
bradfitz, line 92: I think this one's fine to keep. It's the `if !resolvedUp ` cases below we'd remove.

danderson: Yeah, keep this one. Its goal is to give resolved a chance to start up and fix /etc/resolv.conf before we read it, to increase the chances that we'll read a good config.

bradfitz:
Leaving for @danderson.

danderson (approved):

PR #5168, opened 2022-07-25
tka: support processing non-primary forks, scenario-driven tests

twitchyliquid64:
- Fix a bug where processing valid-but-not-primary-chain AUMs from a peer would make processing explode, by supporting computing the state through non-primary chains.
 - Start implementing some scenario-driven tests

### DIFF
--- a/tka/chaintest_test.go
+++ b/tka/chaintest_test.go
@@ -30,13 +30,21 @@ type testchainNode struct {
 	HashSeed   int
 	Template   string
 	SignedWith string
+
+	// When set, uses this hash as the parent hash when
+	// Parent is not set.
+	//
+	// Set when a testChain is based on a different one
+	// (in scenario_test.go).
+	ParentHash *AUMHash
 }
 
 // testChain represents a constructed web of AUMs for testing purposes.
 type testChain struct {
-	Nodes     map[string]*testchainNode
-	AUMs      map[string]AUM
-	AUMHashes map[string]AUMHash
+	FirstIdent string
+	Nodes      map[string]*testchainNode
+	AUMs       map[string]AUM
+	AUMHashes  map[string]AUMHash
 
 	// Configured by options to NewTestchain()
 	Template    map[string]AUM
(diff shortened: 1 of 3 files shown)


tka/chaintest_test.go
```diff
@@ -30,13 +30,21 @@ type testchainNode struct {
 	HashSeed   int
 	Template   string
 	SignedWith string
+
+	// When set, uses this hash as the parent hash when
+	// Parent is not set.
+	//
+	// Set when a testChain is based on a different one
+	// (in scenario_test.go).
+	ParentHash AUMHash
```
crawshaw, line 39: "When set" does this mean an AUMHash of zero bytes is unset?

I'm fine with that, but elsewhere you have used pointers-to-thing and nil to represent unset. Probably worth being consistent even in tests.

tka/tka.go
```diff
@@ -160,12 +164,18 @@ func advanceChain(state State, candidates []AUM) (next *AUM, out State, err erro
...
+// the given advancer to get+apply the next update. This process is repeated
+// until the given termination function returns true or there is no more
+// progress possible.
 //
 // The last-processed AUM, and the state computed after applying the last AUM,
 // are returned.
-func fastForward(storage Chonk, maxIter int, startState State, done func(curAUM AUM, curState State) bool) (AUM, State, error) {
+func fastForwardWithAdvancer(
+	storage Chonk, maxIter int, startState State,
+	advancer func(state State, candidates []AUM) (next *AUM, out State, err error),
```
crawshaw, line 176: I see how you got here, but this is really the outer limit of what is practical with passing closures around before it becomes unreadable. :)

You're also pretty close to being able to combine advancer and done, but we can think about that another time.

crawshaw (approved):

tka/chaintest_test.go
twitchyliquid64: Done

tka/tka.go
twitchyliquid64: Yeah definitely XD If we end up needing to further complicate fast forward, i think its time for another abstraction.

PR #5170, opened 2022-07-25
cmd/tsconnect: add README with instructions

mihaip:
Outlines basic development, build and serving workflows.


catzkorn (approved):
Will add running with a local control at a later point!

PR #5171, opened 2022-07-25
logtail: perform zstd compression with single-segment mode

dsnet:
The "single-segment" mode encodes the expected uncompressed size
in the zstd frame, which the server can use as an optimization.

Also, add a new Config.CompressTransport flag.
Compression is an implementation detail of log uploading.
It could be zstd today or a different format in the future.
Most users of logtail should not care what compression is used.


logtail/logtail.go
```diff
@@ -44,17 +46,18 @@ type Encoder interface {
...
+	PrivateID         PrivateID        // machine-specific private identifier
+	BaseURL           string           // if empty defaults to "https://log.tailscale.io"
+	HTTPC             *http.Client     // if empty defaults to http.DefaultClient
+	SkipClientTime    bool             // if true, client_time is not written to logs
+	LowMemory         bool             // if true, logtail minimizes memory use
+	TimeNow           func() time.Time // if set, subsitutes uses of time.Now
+	Stderr            io.Writer        // if set, logs are sent here instead of os.Stderr
+	StderrLevel       int              // max verbosity level to write to stderr; 0 means the non-verbose messages only
+	Buffer            Buffer           // temp storage, if nil a MemoryBuffer
+	NewZstdEncoder    func() Encoder   // if set, used to compress logs for transmission
```
willnorris, line 59: I find this API a little confusing, at least without additional documentation that NewZstdEncoder and CompressTransport generally shouldn't be used together.

Is the new field really necessary?  Any reason not to just set SingleSegment in the two places where we are setting NewZstdEncoder?

dsnet: I'd rather move in the direction of hiding any details of zstd from the public API of `logtail`. It's an implementation detail and its exposure makes it harder to change the underlying protocol in the future.

It turns out that the only usages of `NewZstdEncoder` are in the `tsnet` and `logpolicy` package. I'm thinking of just deleting it so there's no ambiguity how the two options compose together.

mihaip: +1 to removing `NewZstdEncoder`

willnorris (approved):

mihaip (approved):

bradfitz:
@dsnet, you going to get back to this or should we close it?

PR #5172, opened 2022-07-26
cmd/tsconnect: add Tailwind CSS support and switch to using it

### LINKED ISSUE #5129: cmd/tsconnect: Fix warnings in package.json
Warning about missing license.

Should also rename it from @tailscale/ssh

### LINKED ISSUE #5136: cmd/tsconnect: Set up Tailwind
See also https://github.com/tailscale/corp/issues/5110

mihaip:
Integrates Tailwind CSS as an esbuild plugin that invokes the CLI
to process the input. It takes ~400ms, so it seems like the easiest
option (vs running a separate process for dev mode).

Existing minimal look and feel is replicated with Tailwind classes,
mostly to prove that the entire system works, including unused
class removal.

Also fixes yarn warnings about package.json not having a license
(which were showing up when invoking any scripts).

Fixes #5136
Fixes #5129


bradfitz (approved):

cmd/tsconnect/common.go
```diff
@@ -111,3 +118,30 @@ type EsbuildMetadata struct {
 		EntryPoint string `json:"entryPoint,omitempty"`
 	} `json:"outputs,omitempty"`
 }
+
+func setupEsbuildTailwind(build esbuild.PluginBuild, dev bool) {
+	build.OnLoad(esbuild.OnLoadOptions{
+		Filter: "./src/index.css$",
+	}, func(args esbuild.OnLoadArgs) (esbuild.OnLoadResult, error) {
+		start := time.Now()
+		yarnArgs := []string{"--silent", "tailwind", "-i", args.Path}
```
rosszurowski, line 127: I assume esbuild handles all the file watching for us here?

It may later end up being faster to rely on tailwind's own `--watch` flag, but this seems like a fine place to start.

rosszurowski (approved):
This is great! We can tweak the design later on, but introducing Tailwind will make that much easier.

cmd/tsconnect/common.go
mihaip: esbuild invokes this every time the CSS is requested, so it adds a bit of latency. Managing a persistent watch process will add some complexity, we look into it if the slowdown becomes noticeable during development.

PR #5178, opened 2022-07-26
.github/workflows: add gofmt check

bradfitz:


.github/workflows/staticcheck.yml
```diff
@@ -21,6 +21,9 @@ jobs:
     - name: Check out code
       uses: actions/checkout@v3
 
+    - name: Run gofmt
+      run: "gofmt -d $(find . -maxdepth 1 -type d -and -not -name '.git' -and -not -name '.' -or -name '*.go')"
```
maisem, line 25: Do we use `gofmt` or `goimports`?

maisem (approved):

.github/workflows/staticcheck.yml
bradfitz: I'd be happy with only gofmt for now.

You're asking whether we want to use `goimports --format-only` instead here? Could.

maisem: I mean I use `goimports` locally, if we want to not confuse new people joining the team I would like tools to be consistent across CI and local dev. So yes `goimports --format-only`.

DentonGentry (approved):
I still don't understand why `go fmt` produced changes in derp-bwmon. I ran it locally again this morning, and it didn't rearrange the order of the imports.

Running `go fmt` again now, after your change rearranging the imports, still doesn't suggest changing them back to alphabetical. For me it just ignores the order of the imports.

```
$ go version
go version go1.18 linux/amd64
```

--------

Hang on, what the heck is gofmt? And `goimports --format-only`?
I guess that explains it.

bradfitz:
> Hang on, what the heck is gofmt? And goimports --format-only?

`gofmt` works on files/directories. It only formats things.

`go fmt` is the wrapper from the `cmd/go` tool. It works at the package level and understands things like `./...` etc.

`goimports` adds import management/sorting on top of gofmt.

bradfitz:
> I still don't understand why go fmt produced changes in derp-bwmon. 

`goimports` did, which I necessarily ran to fix imports.

.github/workflows/staticcheck.yml
bradfitz: done

PR #5193, opened 2022-07-27
cmd/tailscaled: add opt-in support for linking CLI into daemon

### LINKED ISSUE #2233: Merge tailscale and tailscaled binaries
For --tun=userspace-networking in particular, and especially with ephemeral authkeys, it would be nice if it were easier to launch and configure a tailscaled in a single step. Also, our tailscale and tailscaled binaries contain a ton of overlapping code which would be greatly reduced if we just had a single binary; this matters a lot on small systems like openwrt.

I suggest we try merging the two binaries as follows:

1) Add a new `tailscale server` subcommand to the tailscale program. This takes the same options as current tailscaled and runs the same code as current tailscaled. For backwards compat, we can then supply a "tailscaled" wrapper script that just runs `exec tailscale server "$@"`

2) Add a new "tailscale run" (or something like that) subcommand that mostly takes the same options as `tailscale up`, but launches its own server and then brings it up. Perhaps this should only work with authkeys, so that it's guaranteed to be noninteractive. It would be nice if this could also (optionally) fork into the background after making a connection, so that it's easy to insert into container startup scripts, without having to write silly retry loops.

bradfitz:
Doesn't help much, though.

    $ go install --tags=ts_include_cli ./cmd/tailscaled/
    $ ls -lh ~/go/bin/tailscaled
    -rwxr-xr-x 2 bradfitz bradfitz 34M Jul 27 11:00 /home/bradfitz/go/bin/tailscaled
    $ go install --tags= ./cmd/tailscaled/
    $ ls -lh ~/go/bin/tailscaled
    -rwxr-xr-x 1 bradfitz bradfitz 23M Jul 27 11:00 /home/bradfitz/go/bin/tailscaled
    $ ls -lh ~/go/bin/tailscale
    -rwxr-xr-x 1 bradfitz bradfitz 13M Jul 25 21:30 /home/bradfitz/go/bin/tailscale

Fixes #2233

### DIFF
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -128,6 +128,8 @@ var subCommands = map[string]*func([]string) error{
 	"be-child":                &beChildFunc,
 }
 
+var beCLI func() // non-nil if CLI is linked in
+
 func main() {
 	printVersion := false
 	flag.IntVar(&args.verbose, "verbose", 0, "log verbosity level; 0 is default, 1 or higher are increasingly verbose")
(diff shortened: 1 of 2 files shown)


andrew-d (approved):
👍

PR #5196, opened 2022-07-27
tka: implement API surface for generating updates

twitchyliquid64:
Based on the builder pattern.

### DIFF
--- a/tka/builder.go
+++ b/tka/builder.go
@@ -1,0 +1,111 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package tka
+
+import (
+	"fmt"
+)
+
+// UpdateBuilder implements a builder for changes to the tailnet
+// key authority.
+//
+// Finalize must be called to compute the update messages, which
+// must then be applied to all Authority objects using Inform().
+type UpdateBuilder struct {
+	a      *Authority
+	signer func(*AUM) error
+
+	state  State
+	parent AUMHash
+
+	out []AUM
+}
+
+func (b *UpdateBuilder) mkUpdate(update AUM) error {
+	prevHash := make([]byte, len(b.parent))
+	copy(prevHash, b.parent[:])
+	update.PrevAUMHash = prevHash
+
+	if b.signer != nil {
+		if err := b.signer(&update); err != nil {
+			return fmt.Errorf("signing failed: %v", err)
+		}
+	}
+	if err := update.StaticValidate(); err != nil {
+		return fmt.Errorf("generated update was invalid: %v", err)
+	}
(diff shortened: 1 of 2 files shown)


tka/builder.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	// Generate UpdateKey messages for existing keys.
+	for k, info := range b.votesToChange {
+		if _, err := state.GetKey(info.keyID); err != nil {
+			return nil, fmt.Errorf("failed reading key %x: %v", info.keyID, err)
+		}
+		update := AUM{MessageKind: AUMUpdateKey, Votes: &info.votes, KeyID: info.keyID}
+		// If theres also a meta update for this key, roll it into one update.
+		if info, hasMetaUpdate := b.metaToChange[k]; hasMetaUpdate {
```
sailorfrag, line 118: I think this is the only reason for the builder fields to be maps, I wonder if it would be fine to just skip this optimization and use slices instead. Or even preserve it with slices -- yes that makes this a O(_n_<sup>2</sup>) operation but I expect _n_ to be small.

Or if you are doing these optimizations, I feel like there are others that are missing:
* Cancel out adds and deletes
* Cancel out vote/meta changes if there's a delete
* This covers [meta, vote]; also check for [vote, meta].

Also the vote/meta split could be resolved by making `SetKeyMeta` take an additional parameter for vote changes, which could be 0 or something like `const noVoteChange = 0`.

If you add the same operation twice (delete a particular key, change the vote for a particular key, etc), it'll silently overwrite the old instance, is that intentional?

tka/builder_test.go
```diff
@@ -0,0 +1,152 @@
...
+	// + checking if the key has been removed.
+	if err := a.Inform(updates); err != nil {
+		t.Fatalf("could not apply generated updates: %v", err)
+	}
+	if _, err := a.state.GetKey(key2.ID()); err != ErrNoSuchKey {
+		t.Errorf("GetKey(key2).err = %v, want %v", err, ErrNoSuchKey)
+	}
+}
+
+func TestAuthorityBuilderSetKeyVote(t *testing.T) {
```
sailorfrag, line 82: If you're keeping the "update vote with meta" optimzation, there should be a test case that covers that.

tka/builder_test.go
```diff
@@ -0,0 +1,152 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package tka
+
+import (
+	"testing"
```
sailorfrag, line 8: I think there should be some big test cases for multiple operations: both multiple of the same operation and a mix of different operations

twitchyliquid64:
Updated based on our convo: We now avoid being smart and use the `StaticValidate()` + `applyVerifiedAUM()` methods to check for correctness. PTAL.

sailorfrag (approved):

Issue #4811, opened 2022-06-07
http proxy not working since 1.22.x


steom:
just to inform that 
Improved HTTPS proxy handling for windows client from 1.24.0, 124.2 and still today 1.26.0
broke support for http proxy.

version 1.22.2 works great instead!

control: control server key from https://controlplane.tailscale.com: ts2021=[fSeS+], legacy=[nlFWp]
control: RegisterReq: onode= node=[25gj+] fup=false
tshttpproxy: winhttp: GetProxyForURL("http://controlplane.tailscale.com:80/ts2021"): ERROR_INVALID_PARAMETER [unexpected]

logtail: dial "log.tailscale.io:443" failed: dial tcp 34.229.201.48:443: 
Received error: PollNetMap: Post "https://controlplane.tailscale.com/machine/map": connection attempts aborted by context: context deadline exceeded
trying bootstrapDNS("derp9b.tailscale.com", "144.202.67.195") for "log.tailscal.io" ...
bootstrapDNS("derp9b.tailscale.com", "144.202.67.195") for "log.tailscale.io" eror: Get "https://derp9b.tailscale.com/bootstrap-dns?q=log.tailscale.io": dialcp 144.202.67.195:443: 
trying bootstrapDNS("derp8b.tailscale.com", "2a03:b0c0:1:d0::ec1:e001") for "lo.tailscale.io" ...
bootstrapDNS("derp8b.tailscale.com", "2a03:b0c0:1:d0::ec1:e001") for "log.tailsale.io" error: Get "https://derp8b.tailscale.com/bootstrap-dns?q=log.tailscale.
o": dial tcp [2a03:b0c0:1:d0::ec1:e001]:443: 
trying bootstrapDNS("derp8c.tailscale.com", "206.189.16.32") for "log.tailscaleio" ...

bradfitz:
cc @sailorfrag @dblohm7

steom:
bug still present in 1.28.0
version 1.22.2 works great instead!

dblohm7:
@steom What do your logs look like in 1.28.0? Are there still proxy errors in them? Can you post a snippet or send a bug report ID?

steom:
tshttpproxy: CONNECT response from http://proxy.lan.org:80 for target "controlplane.tailscale.com:443" (auth "Basic xxxxxxxxxxxxxxxxxx...
(34 total bytes)"): 200 Connection established
control: control server key from https://controlplane.tailscale.com: ts2021=[fSe
[...]

PR #5197, opened 2022-07-27
net/tshttpproxy: fix incorrect type in Windows implementation, switch…

dblohm7:
… to mkwinsyscall, fix memory leak

The definition of winHTTPProxyInfo was using the wrong type (uint16 vs uint32)
for its first field. I fixed that type.

Furthermore, any UTF16 strings returned in that structure must be explicitly
freed. I added code to do this.

Finally, since this is the second time I've seen type safety errors in this code,
I switched the native API calls over to use wrappers generated by mkwinsyscall.
I know that would not have helped prevent the previous two problems, but every
bit helps IMHO.

Updates https://github.com/tailscale/tailscale/issues/4811


net/tshttpproxy/tshttpproxy_windows.go
```diff
@@ -221,30 +201,46 @@ type autoProxyOptions struct {
...
 // https://docs.microsoft.com/en-us/windows/win32/api/winhttp/ns-winhttp-winhttp_proxy_info
 type winHTTPProxyInfo struct {
-	AccessType  uint16
+	AccessType  uint32
 	Proxy       *uint16
 	ProxyBypass *uint16
 }
 
-var proxyForURLOpts = &autoProxyOptions{
+type winHGlobal windows.Handle
```
bradfitz, line 209: why do we need this type?

net/tshttpproxy/mksyscall.go
```diff
@@ -0,0 +1,8 @@
+package tshttpproxy
+
+//go:generate go run golang.org/x/sys/windows/mkwinsyscall -output zsyscall_windows.go mksyscall.go
+
+//sys globalFree(hglobal winHGlobal) (err error) [failretval==0] = kernel32.GlobalFree
```
bradfitz, line 5: ugh I guess https://go-review.googlesource.com/c/sys/+/413435 stalled out

bradfitz (approved):
LGTM after license header fix

net/tshttpproxy/tshttpproxy_windows.go
dblohm7: In terms of "will it compile," we don't, but `HGLOBAL` is considered to be distinct from `HANDLE` as far as the SDK is concerned. I wanted to reflect that.

PR #5199, opened 2022-07-27
Revert "net/dns: remove systemd-resolved ping"

nyghtowl:
Reverts tailscale/tailscale#5153


bradfitz:
Why? The commit message needs to say why (briefly) and include a reference to a bug.

bradfitz:
I assume you saw the windows-race build failure and assumed it was related? Looks like our CI is depending on the sourceforge servers to get mingw32 or something and they had server problems? You can file a separate bug about that, perhaps.

nyghtowl:
Appears there is an issue on this build failure https://github.com/tailscale/tailscale/issues/4926

PR #5200, opened 2022-07-27
build_dist.sh: add --box and --extra-small flag to produce smaller and boxed binaries

raggi:
- `--box` when ./cmd/tailscaled is built with this flag, it builds a
  "toybox" style binary that includes tailscale and tailscaled.
- `--extra-small` strip the output binary and omit some dependencies
  (currently AWS integration).

### DIFF
--- a/build_dist.sh
+++ b/build_dist.sh
@@ -45,4 +45,25 @@ EOF
 	exit 0
 fi
 
-exec ./tool/go build -ldflags "-X tailscale.com/version.Long=${LONG} -X tailscale.com/version.Short=${SHORT} -X tailscale.com/version.GitCommit=${GIT_HASH}" "$@"
+tags=""
+ldflags="-X tailscale.com/version.Long=${LONG} -X tailscale.com/version.Short=${SHORT} -X tailscale.com/version.GitCommit=${GIT_HASH}"
+
+# build_dist.sh arguments must precede go build arguments.
+while [ "$#" -gt 1 ]; do
+	case "$1" in
+	--extra-small)
+		shift
+		ldflags="$ldflags -w -s"
+		tags="${tags:+$tags,}ts_omit_aws"
+		;;
+	--box)
+		shift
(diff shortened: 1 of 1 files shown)


build_dist.sh
```diff
@@ -45,4 +45,22 @@ EOF
...
+buildflags=""
+ldflags="-X tailscale.com/version.Long=${LONG} -X tailscale.com/version.Short=${SHORT} -X tailscale.com/version.GitCommit=${GIT_HASH}"
+
+# build_dist.sh arguments must precede go build arguments.
+while [ "$#" -gt 1 ]; do
+	case "$1" in
+	--extra-small)
+		shift
+		ldflags="$ldflags -w -s"
+		buildflags="$buildflags -tags=ts_include_cli,ts_omit_aws"
```
andrew-d, line 57: Should the `--extra-small` flag imply `ts_include_cli`? Or are there cases where we want to produce a small non-combined binary/binaries?

andrew-d (approved):
LGTM with one question

build_dist.sh
raggi: Good question, I'm not sure.

Here for a 64-bit linux build, this stripped tailscaled is 18MB with tailscale inside, and 15MB without it, so 20% for the rest of the functionality.

What do you think?

I guess I could also add a special flag to make it look something like this:

```bash
$ ./build_dist.sh --extra-small tailbox
```

And have that produce `tailscaled` and `tailscale` as hardlinks?

raggi: The other thing we could do is to actually make ./cmd/tailbox exist, then you could build _that_ target explicitly to get a toybox'd style binary.

raggi:
aside: wanted https://github.com/golang/go/issues/51692

build_dist.sh
raggi: I split these into two arguments for now, see new description.

andrew-d (approved):
Love the array conditions - very nice 😃

PR #5202, opened 2022-07-27
cmd/tsconnect: add basic panic handling

### LINKED ISSUE #5132: cmd/tsconnect: Handle Go panics
e.g. I ran into one when generating the JS version of the netmap

```
wasm_exec.js:22 [signal 0xb code=0x0 addr=0x0 pc=0x0]
wasm_exec.js:22 
wasm_exec.js:22 goroutine 32 [running]:
wasm_exec.js:22 main.(*jsIPN).run.func2.2(0xff68c0)
wasm_exec.js:22 	tailscale.com/cmd/tsconnect/wasm/wasm_js.go:192 +0xe
```

Should both handle that nil and have a signal to the JS that the go side is dead.

mihaip:
The go wasm process exiting is a sign of an unhandled panic. Also
add a explicit recover() call in the notify callback, that's where most
logic bugs are likely to happen (and they may not be fatal).

Also fixes the one panic that was encountered (nill pointer dereference
when generating the JS view of the netmap).

Fixes #5132

### DIFF
--- a/cmd/tsconnect/wasm/wasm_js.go
+++ b/cmd/tsconnect/wasm/wasm_js.go
@@ -114,6 +114,7 @@ func newIPN(jsConfig js.Value) map[string]any {
 					notifyState(state: int): void,
 					notifyNetMap(netMap: object): void,
 					notifyBrowseToURL(url: string): void,
+					notifyPanicRecover(err: string): void,
 				})`)
 				return nil
 			}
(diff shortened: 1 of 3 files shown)


cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -166,6 +167,16 @@ func (i *jsIPN) run(jsCallbacks js.Value) {
...
 
 	i.lb.SetNotifyCallback(func(n ipn.Notify) {
+		// Panics in the notify callback are likely due to be due to bugs in
+		// this bridging module (as opposed to actual bugs in Tailscale) and
+		// thus may be recoverable. Let the UI know, and allow the user to
+		// choose if they want to reload the page.
+		defer func() {
+			if r := recover(); r != nil {
+				fmt.Println("Panic recovered:", r)
+				jsCallbacks.Call("notifyPanicRecover", fmt.Sprintf("%v", r))
```
bradfitz, line 177: or just fmt.Sprint(r)

bradfitz (approved):

PR #5207, opened 2022-07-28
logtail,logpolicy: tweak minor cosmetic things

bradfitz:
Just reading the code again in prep for some alloc reductions.

### DIFF
--- a/logtail/logtail.go
+++ b/logtail/logtail.go
@@ -167,8 +167,9 @@ type Logger struct {
 
 	procID              uint32
 	includeProcSequence bool
-	writeLock           sync.Mutex // guards increments of procSequence
-	procSequence        uint64
+
+	writeLock    sync.Mutex // guards increments of procSequence
+	procSequence uint64
 
 	shutdownStart chan struct{} // closed when shutdown begins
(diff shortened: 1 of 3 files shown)


logtail/logtail.go
```diff
@@ -659,10 +660,12 @@ func (l *Logger) Write(buf []byte) (int, error) {
...
 	}
+
 	l.writeLock.Lock()
-	b := l.encode(buf, level)
-	_, err := l.send(b)
-	l.writeLock.Unlock()
+	defer l.writeLock.Unlock()
+
+	b := l.encodeLocked(buf, level)
+	_, err := l.sendLocked(b)
```
maisem, line 668: should this really be done while holding the lock? (not suggesting you change it in this pr, just asking)

maisem (approved):

logtail/logtail.go
bradfitz: It doesn't really do much, but I didn't put much thought into it yet.

The filch implementation of Buffer and the memory implementation have different ownership lifetimes. Need to figure out a design that works for both of them, or unify and remove the interface. Or something.

PR #5208, opened 2022-07-28
tailcfg: add wire fields/docs for resuming streaming map sessions

bradfitz:
No implementation yet.

Updates #7175

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -979,6 +979,25 @@ type MapRequest struct {
 	Stream      bool // if true, multiple MapResponse objects are returned
 	Hostinfo    *Hostinfo
 
+	// MapSessionHandle, if non-empty, is a request to reattach to a previous
+	// map session after a previous map session was interrupted for whatever
+	// reason. Its value is an opaque string as returned by
+	// MapResponse.MapSessionHandle.
+	//
+	// When set, the client must also send MapSessionSeq to specify the last
+	// processed message in that prior session.
+	//
+	// The server may choose to ignore the request for any reason and start a
+	// new map session. This is only applicable when Stream is true.
+	MapSessionHandle string `json:",omitempty"`
+
+	// MapSessionSeq is the sequence number in the map session identified by
+	// MapSesssionHandle that was most recently processed by the client.
+	// It is only applicable when MapSessionHandle is specified.
+	// If the server chooses to honor the MapSessionHandle request, only sequence
+	// numbers greater than this value will be returned.
+	MapSessionSeq int64 `json:",omitempty"`
+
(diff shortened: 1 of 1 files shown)


bradfitz:
Thoughts on something like this?

tailcfg/tailcfg.go
```diff
@@ -979,6 +979,25 @@ type MapRequest struct {
...
+	// The server may choose to ignore the request for any reason and start a
+	// new map session. This is only applicable when Stream is true.
+	MapSessionHandle string `json:",omitempty"`
+
+	// MapSessionSeq is the sequence number in the map session identified by
+	// MapSesssionHandle that was most recently processed by the client.
+	// It is only applicable when MapSessionHandle is specified.
+	// If the server chooses to honor the MapSessionHandle request, only sequence
+	// numbers greater than this value will be returned.
+	MapSessionSeq int `json:",omitempty"`
```
andrew-d, line 999: Do we want to use an explicit `int32` / `int64` here, for 32-bit clients? Presumably overflowing a 32-bit integer is pretty unlikely.

tailcfg/tailcfg.go
```diff
@@ -1313,6 +1332,20 @@ type PingResponse struct {
 }
 
 type MapResponse struct {
+	// MapSessionHandle optionally specifies a unique opaque handle for this
+	// stateful MapResponse session. Servers may not send it, and it's only sent
```
andrew-d, line 1336: Maybe "servers may choose not to send it", since I read this as "it is not permitted for a server to send this" at first?

tailcfg/tailcfg.go
```diff
@@ -1313,6 +1332,20 @@ type PingResponse struct {
...
+	// on the first MapResponse in a stream. The client can determine whether
+	// it's reattaching to a prior stream by seeing whether this value matches
+	// the reuqested MapRequest.MapSessionHandle.
```
andrew-d, line 1339:
```suggestion
	// the requested MapRequest.MapSessionHandle.
```

andrew-d (approved):
Sounds good with some minor spelling/wording suggestions.

tailcfg/tailcfg.go
bradfitz: I figured server-side we'd just end the session after 2 billion.

tailcfg/tailcfg.go
bradfitz: Done

tailcfg/tailcfg.go
andrew-d: 👍

bradfitz: But int64 works. There won't be JS clients.

PR #5228, opened 2022-07-29
shell.nix: update Go toolchain to 1.19

twitchyliquid64:
Same as in corp lol

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -12,8 +12,8 @@
 
 {
 	pkgs ? import <nixpkgs> {},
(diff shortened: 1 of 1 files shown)


bradfitz:
No, in corp this is automatically updated. What's the story here?

bradfitz (approved):
but sure

twitchyliquid64:
Because oss cant rely on corp it needs to be updated manually.

This change ensure ppl doing `nix-shell` in this repo will get the correct toolchain.

PR #5230, opened 2022-07-29
tka,types/key: implement NLPrivate glue for tailnet key authority keys

twitchyliquid64:
It looks like the new toolchain has different rules for `go fmt`, so it seems a bunch of the comments got reformatted.


types/key/nl.go
```diff
@@ -0,0 +1,74 @@
...
+	// nlPrivateHexPrefix is the prefix used to identify a
+	// hex-encoded network-lock key.
+	nlPrivateHexPrefix = "nlpriv:"
+)
+
+// NLPrivate is a node-managed network-lock key, used for signing
+// node-key signatures and authority update messages.
+type NLPrivate struct {
+	_ structs.Incomparable // because == isn't constant-time
+	k [ed25519.PrivateKeySize]byte
```
sailorfrag, line 25: any particular reason to use a byte array rather than ed25519.PrivateKey?

sailorfrag (approved):

types/key/nl.go
twitchyliquid64: Not really, I'm mostly following the pattern of using an array rather than a slice, I presume this established as the pattern to minimize allocations.

PR #5231, opened 2022-07-29
cmd/tsconnect: allow SSH username to be specified

### LINKED ISSUE #5139: cmd/tsconnect: Allow SSH user to be specified
Right now we rely on `defaultUser`, which is not documented

mihaip:
Redoes the UI to be a form, with a username field and a host drop-down
(we also show non-Tailscale SSH hosts now, in a separate section)

Fixes #5139

### DIFF
--- a/cmd/tsconnect/index.html
+++ b/cmd/tsconnect/index.html
@@ -12,7 +12,27 @@
         <div class="text-gray-600" id="state">Loading…</div>
       </header>
     </div>
-    <div id="peers" class="container mx-auto px-4"></div>
+    <form
+      id="ssh-form"
+      class="container mx-auto px-4 hidden flex justify-center"
+    >
+      <input type="text" class="input username" placeholder="Username" />
+      <div class="select-with-arrow mx-2">
+        <select class="select"></select>
+      </div>
+      <input
+        type="submit"
+        class="button bg-green-500 border-green-500 text-white hover:bg-green-600 hover:border-green-600"
+        value="SSH"
+      />
+    </form>
+    <div id="no-ssh" class="container mx-auto px-4 hidden text-center">
+      None of your machines have
+      <a href="https://tailscale.com/kb/1193/tailscale-ssh/" class="link"
+        >Tailscale SSH</a
+      >
+      enabled. Give it a try!
+    </div>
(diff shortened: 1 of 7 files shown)


mihaip:
Deployed at https://tsconnect.fly.dev/ if you want play with it.

bradfitz:
But now the hosts drop-down is all nodes, not just ones that can Tailscale SSH

mihaip:
> But now the hosts drop-down is all nodes, not just ones that can Tailscale SSH

Technically it should work if you're willing to do username/password credentials (have not tested this).

bradfitz:
> Technically it should work if you're willing to do username/password credentials (have not tested this).

I don't think we should permit password auth. I don't want to encourage people to type their passwords into rando websites.

mihaip:
> I don't think we should permit password auth. I don't want to encourage people to type their passwords into rando websites.

Fair enough, switched back to showing just nodes with Tailscale SSH.

cmd/tsconnect/index.html
```diff
@@ -12,7 +12,27 @@ <h1 class="text-3xl font-bold grow">Tailscale Connect</h1>
...
+        value="SSH"
+      />
+    </form>
+    <div id="no-ssh" class="container mx-auto px-4 hidden text-center">
+      None of your machines have
+      <a href="https://tailscale.com/kb/1193/tailscale-ssh/" class="link"
+        >Tailscale SSH</a
+      >
+      enabled. Give it a try!
+    </div>
```
catzkorn, line 35: Might be known behaviour but this is being triggered when a machine isn't auth'd and retains the QR/login link below. Even though it says Needs Auth top right, it might be a bit confusing. I don't think this was happening prior but might be wrong.

Screen Shot 2022-07-29 at 3 03 16 PM

catzkorn (approved):

cmd/tsconnect/index.html
mihaip: Don't think that's a regression(we had a similar empty check for the list of SSH peers), but I agree that it's a bug. Added a check on the Go side to not send the netmap notification if we're not in the `Running` state.

PR #5249, opened 2022-08-01
tsweb: add HandlerOptions.Quiet499s

dsnet:
Client termination of a request is a fairly common situation
where we may want to suppress logs for.
The Quiet499s suppresses such logs similar to Quiet200s.


bradfitz (approved):

dsnet:
Dropping. There are other codes that I want to ignore and the API doesn't scale adding each code.

Somewhat of a hack, but I can specify a custom `Logf` function that checks whether it was called with `logf("%s", tswenb.AccessLogRecord)` and manually check the `AccessLogRecord` to see if it is something I should ignore.

PR #5250, opened 2022-08-01
tsweb: mark AccessLogRecord fields as omitempty

dsnet:
If the field is the zero value, then avoid serializing the field.
This reduces verbosity in server logs.


bradfitz (approved):
Kinda weird to do this for the super common ones that are never empty, but sure.

dsnet:
Didn't really want to think about what exactly was common versus not 😏

PR #5251, opened 2022-08-01
tsweb: add HTTPError.Header

dsnet:
The Hdr field allows the server to specify specific headers to set.
Example use case: server returns 429 with the "Retry-After" header set.


bradfitz:
Hdr is pretty short for a field name.

dsnet:
Matches `Msg` being pretty short 😐. Do we want to be consistent or not? I personally prefer `Header` as well.

bradfitz (approved):
Let's go with `Header` then.

dsnet:
Done.

bradfitz:
typo in commit message: `tsbweb`

PR #5254, opened 2022-08-02
cmd/tsconnect: make terminal resizable

### LINKED ISSUE #5150: cmd/tsconnect: Make terminals resizable
Easiest thing may be to make them take up the entire window (possibly allowing them to to be popped up in a new window via `window.open` and then hooking up the window resize to the [onResize xterm.js event](https://xtermjs.org/docs/api/terminal/classes/terminal/#onresize) and/or the `WindowChange` method on the session.

mihaip:
Makes the terminal container DOM node as large as the window (except for
the header) via flexbox. The xterm.js terminal is then sized to fit via
xterm-addon-fit. Once we have a computed rows/columns size, and we can
tell the SSH session of the computed size.

Required introducing an IPNSSHSession type to allow the JS to control
the SSH session once opened. That alse allows us to programatically
close it, which we do when the user closes the window with the session
still active.

I initially wanted to open the terminal in a new window instead (so that
it could be resizable independently of the main window), but xterm.js
does not appear to work well in that mode (possibly because it adds an
IntersectionObserver to pause rendering when the window is not visible,
and it ends up doing that when the parent window is hidden -- see
xtermjs/xterm.js@87dca56dee8018a17b5cb22ec844b7013629da63)

Fixes #5150

### DIFF
--- a/cmd/tsconnect/wasm/wasm_js.go
+++ b/cmd/tsconnect/wasm/wasm_js.go
@@ -142,11 +142,10 @@ func newIPN(jsConfig js.Value) map[string]any {
 				log.Printf("Usage: ssh(hostname, userName, termConfig)")
 				return nil
 			}
-			go jsIPN.ssh(
+			return jsIPN.ssh(
 				args[0].String(),
 				args[1].String(),
 				args[2])
-			return nil
 		}),
 	}
 }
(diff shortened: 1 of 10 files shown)


cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -256,13 +255,42 @@ func (i *jsIPN) logout() {
...
 }
 
-func (i *jsIPN) ssh(host, username string, termConfig js.Value) {
-	writeFn := termConfig.Get("writeFn")
-	setReadFn := termConfig.Get("setReadFn")
-	rows := termConfig.Get("rows").Int()
-	cols := termConfig.Get("cols").Int()
-	onDone := termConfig.Get("onDone")
+func (i *jsIPN) ssh(host, username string, termConfig js.Value) map[string]any {
+	jsSSHSession := jsSSHSession{
```
bradfitz, line 259: add a `&` here. Your Run method's on the pointer receiver here, so don't make it possible to accidentally do value copies in the future. Stay in pointers from the beginning.

catzkorn (approved):

PR #5260, opened 2022-08-02
wgengine/magicsock: avoid RebindingUDPConn mutex in common read/write case

bradfitz:


wgengine/magicsock/magicsock.go
```diff
@@ -2974,11 +2974,22 @@ func (c *Conn) ParseEndpoint(nodeKeyStr string) (conn.Endpoint, error) {
 // RebindingUDPConn is a UDP socket that can be re-bound.
 // Unix has no notion of re-binding a socket, so we swap it out for a new one.
 type RebindingUDPConn struct {
-	mu    sync.Mutex
+	// pconnAtomic is the same as pconn, but doesn't require acquiring mu. It's
+	// used for reads/writes and only upon failure do the reads/writes then
+	// check pconn (after acquiring mu) to see if there's been a rebind
+	// meanwhile.
```
crawshaw, line 2980: Be good to add a note here about why we need to keep pconn at all.

maisem (approved):
LGTM after adding comment as @crawshaw asked

PR #5270, opened 2022-08-02
tailcfg: add CapabilityFileSharingSendTo to identify FileTargets

maisem:
This adds the inverse to CapabilityFileSharingSend so that senders can
identify who they can Taildrop to.


ipn/ipnlocal/local.go
```diff
@@ -3201,6 +3205,15 @@ func (b *LocalBackend) hasCapFileSharing() bool {
 	return b.capFileSharing
 }
 
+func sliceContains[T comparable](s []T, v T) bool {
```
bradfitz, line 3208: https://pkg.go.dev/golang.org/x/exp/slices#Contains

tailcfg/tailcfg.go
```diff
@@ -1503,6 +1504,9 @@ const (
 	// CapabilityFileSharingSend grants the ability to receive files from a
 	// node that's owned by a different user.
 	CapabilityFileSharingSend = "https://tailscale.com/cap/file-send"
+	// CapabilityFileSharingSendTo grants the ability to send files to a
+	// node that's owned by a different user.
```
bradfitz, line 1508: this isn't worded clearly.

It almost sounds backwards.

ipn/ipnlocal/local.go
maisem: Oh right, done

tailcfg/tailcfg.go
maisem: reworded, PTAL

maisem: also moved it to the other capability section to be clear that its not in the filter but in`Node.Capabilities`

tailcfg/tailcfg.go
```diff
@@ -1497,6 +1498,9 @@ const (
 	CapabilityAdmin       = "https://tailscale.com/cap/is-admin"
 	CapabilitySSH         = "https://tailscale.com/cap/ssh"         // feature enabled/available
 	CapabilitySSHRuleIn   = "https://tailscale.com/cap/ssh-rule-in" // some SSH rule reach this node
+	// CapabilityFileSharingSendTo grants the current node the ability to send
+	// files to a different node.
+	CapabilityFileSharingSendTo = "https://tailscale.com/cap/file-send-to"
```
bradfitz, line 1503: CapabilityFileSharingTarget

bradfitz (approved):

Issue #5133, opened 2022-07-22
cmd/tsconnect: Refactor `precompress` code into shared package


mihaip:
https://github.com/tailscale/corp/blob/8af43cb8db9c1b450ebc554fa0736fadb08178a7/misc/precompress/precompress.go and https://github.com/tailscale/tailscale/blob/898695e312ddbe670ac9e35c5b1ea0f4df6ab797/cmd/tsconnect/build.go#L123-L190 are basically duplicates, as is the serving code (https://github.com/tailscale/corp/blob/8af43cb8db9c1b450ebc554fa0736fadb08178a7/control/adminhttp/panel.go#L64-L79 and https://github.com/tailscale/tailscale/blob/898695e312ddbe670ac9e35c5b1ea0f4df6ab797/cmd/tsconnect/serve.go#L111-L131).

Should probably move it to `tsweb` so that it can be used in both places.

PR #5277, opened 2022-08-03
cmd/tsconnect,util/precompress: move precompression to its own package

mihaip:
We have very similar code in corp, moving it to tsweb allows it to be
reused.

Updates #5133


tsweb/precompress.go
```diff
@@ -0,0 +1,129 @@
...
+	return eg.Wait()
+}
+
+type PrecompressOptions struct {
+	// FastCompression controls whether compression should be optimized for
+	// speed rather than size.
+	FastCompression bool
+	// ProgressFn is invoked when a file in the directory is about to be
+	// compressed.
+	ProgressFn func(path string)
```
bradfitz, line 59: is this optional?

tsweb/precompress.go
```diff
@@ -0,0 +1,129 @@
...
+	"os"
+	"path"
+	"path/filepath"
+
+	"github.com/andybalholm/brotli"
+	"golang.org/x/sync/errgroup"
+)
+
+// PrecompressDir compresses static assets in dirPath using Gzip and Brotli, so
+// that they can be later served with OpenPrecompressedFile. Meant to be run
```
bradfitz, line 22: It's meant ....

tsweb/precompress.go
```diff
@@ -0,0 +1,129 @@
...
+// during build steps, to avoid the cost of repeatedly compressing unchanging
+// resources.
+func PrecompressDir(dirPath string, options PrecompressOptions) error {
```
bradfitz, line 25: kinda weird to have this in the tsweb package.

It also means all our tsweb code now brings in brotli?

mihaip: I can move it to `util/precompress`?

tsweb/precompress.go
mihaip: Yes. Added a comment, is there a canonical way to indicate that it is?

mihaip: Looks like a "if non-nil" clause is idiomatic.

tsweb/precompress.go
mihaip: Moved to `util/precompress` as discussed.

bradfitz (approved):

PR #5278, opened 2022-08-03
tsnet: log out ephemeral nodes on Close()

maisem:

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -130,6 +130,25 @@ func (s *Server) Start() error {
 //
 // It must not be called before or concurrently with Start.
 func (s *Server) Close() error {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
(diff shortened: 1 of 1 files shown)


willnorris (approved):

tsnet/tsnet.go
```diff
@@ -144,8 +149,6 @@ func (s *Server) Close() error {
 	s.listeners = nil
 
 	// Perform a best-effort final flush.
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-	defer cancel()
 	s.logtail.Shutdown(ctx)
```
bradfitz, line 152: The time might've been exhausted at this point. Do the two five second things concurrently?

bradfitz:
"log out" is the verb.

tsnet/tsnet.go
maisem: Done

bradfitz (approved):

PR #5292, opened 2022-08-04
all: use atomic.Pointer

maisem:
Also add some missing docs.

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -1151,8 +1151,8 @@ var clockNow = time.Now
 
 // opt.Bool configs from control.
 var (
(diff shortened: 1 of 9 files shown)


control/controlclient/direct.go
```diff
@@ -1151,8 +1151,8 @@ var clockNow = time.Now
 
 // opt.Bool configs from control.
```
bradfitz, line 1152: it does say "opt.Bool" here.

The comment is _factored_ out! 😂

maisem: too far, doesn't follow standards all over the code base :P

bradfitz (approved):
LGTM once it compiles :)

PR #5294, opened 2022-08-04
tsconnect: add flag to specify control server

catzkorn:
To improve the local development experience, this change allows a
control url to be passed in with the `--control-url=` flag.

If no flag is passed, the default remains the Tailscale controlled
control server set by `ipn.DefaultControlURL`.

### DIFF
--- a/cmd/tsconnect/tsconnect.go
+++ b/cmd/tsconnect/tsconnect.go
@@ -22,6 +22,7 @@ var (
 	distDir         = flag.String("distdir", "./dist", "path of directory to place build output in")
 	yarnPath        = flag.String("yarnpath", "../../tool/yarn", "path yarn executable used to install JavaScript dependencies")
 	fastCompression = flag.Bool("fast-compression", false, "Use faster compression when building, to speed up build time. Meant to iterative/debugging use only.")
+	devControl      = flag.String("dev-control", "", "URL of a development control server to be used with dev. If provided without specifying dev, an error will be returned.")
 )
 
 func main() {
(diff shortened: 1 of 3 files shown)


cmd/tsconnect/tsconnect.go
```diff
@@ -22,6 +22,7 @@ var (
 	distDir         = flag.String("distdir", "./dist", "path of directory to place build output in")
 	yarnPath        = flag.String("yarnpath", "../../tool/yarn", "path yarn executable used to install JavaScript dependencies")
 	fastCompression = flag.Bool("fast-compression", false, "Use faster compression when building, to speed up build time. Meant to iterative/debugging use only.")
+	controlURL      = flag.String("control-url", "", "URL of the Tailscale control server. If empty, defaults to DefaultControlURL.")
```
maisem, line 25: Document that this only works in dev mode

DentonGentry: Someday this might be used by a customer on the US Compliance server, or one of the onprem sites.

maisem: Sure, but as implemented it only works in dev mode

mihaip: Better yet fail with an error if it's used in other modes.

mihaip (approved):

cmd/tsconnect/tsconnect.go
maisem: But also document, or maybe rename to `--dev-control`?

mihaip (approved):

PR #5297, opened 2022-08-04
all: use syncs.AtomicValue

maisem:


syncs/syncs.go
```diff
@@ -21,6 +21,49 @@ func initClosedChan() <-chan struct{} {
 	return ch
 }
 
+// AtomicValue is the generic version of atomic.Value.
+type AtomicValue[T any] struct {
```
dsnet, line 25: Why is this implemented in terms of the older `atomic.Value` instead of the newer `atomic.Pointer`? The latter is not as efficient since it's dealing with an interface value under the hood.

maisem: @bradfitz mentioned that we shouldn't add indirection. I don't have any strong feelings on this

dsnet: Isn't an interface already an indirection? It's composed of a pointer to the type and a pointer to be value.

I don't feel strong about this. Frist step is to get all these usages on a common API and then we can figure out how to improve it under the hood.

dsnet (approved):

PR #5298, opened 2022-08-04
tka,types/key: remove dependency for tailcfg & types/ packages on tka

twitchyliquid64:
Following the pattern elsewhere, we create a new tka-specific types package for the types
that need to couple between the serialized structure types, and tka.


types/tkatype/tkatype.go
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package tkatype defines types for working with the tka package.
```
crawshaw, line 5: You may want to add a comment here not to add dependencies, because you want to avoid extra deps in the tka package.

crawshaw (approved):

crawshaw:
Thanks!

PR #5299, opened 2022-08-04
tsnet: fix closing of filch buffer

dsnet:
It should be safe to initialize multiple Server instances
without any resource leaks what-so-ever.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -91,6 +91,7 @@ type Server struct {
 	shutdownCtx      context.Context
 	shutdownCancel   context.CancelFunc
 	localClient      *tailscale.LocalClient
+	logbuffer        *filch.Filch
 	logtail          *logtail.Logger
 
 	mu        sync.Mutex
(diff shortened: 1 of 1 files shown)


willnorris (approved):

tsnet/tsnet.go
```diff
@@ -238,15 +240,15 @@ func (s *Server) start() error {
 	}
 	logid := lpc.PublicID.String()
 
-	f, err := filch.New(filepath.Join(s.rootPath, "tailscaled"), filch.Options{ReplaceStderr: false})
+	s.logbuffer, err = filch.New(filepath.Join(s.rootPath, "tailscaled"), filch.Options{ReplaceStderr: false})
```
bradfitz, line 243: missing cleanup in error paths below, like monitor.New failing

bradfitz: Maybe time to reuse:

```go
	var closePool closeOnErrorPool
	defer closePool.closeAllIfError(&reterr)
```

From wgengine?

Soypete (approved):

tsnet/tsnet.go
dsnet: There are a number of resource leaks on the error path here. I'll fix that in a subsequent PR since this PR is clearly an improvement over the status quo.

PR #5300, opened 2022-08-04
cmd/tsconnect: switch UI to Preact

### LINKED ISSUE #5137: cmd/tsconnect: Set up Preact

### LINKED ISSUE #5273: cmd/tsconnect: list of SSH peers is slow to get populated
We only send the netmap to JS if we're in the "running" state (https://github.com/tailscale/tailscale/commit/c06758c83b23de3a45d0b441221fedc0a73a3f4e#diff-17cfb2d953cf64e47108124d3599ce9db769412392ea9a7af931d3aa906b8183R181) to avoid showing an empty list while waiting for machine authorization. However, depending on how the netmap vs. state change callbacks happen, we may drop a populated netmap on the floor.

We should hold on to all netmap updates, and only control UI visibility based on the state (switching the UI to Preact #5137 would make this easier).

mihaip:
Reduces the amount of boilerplate to render the UI and makes it easier to
respond to state changes (e.g. machine getting authorized, netmap changing,
etc.)

Preact adds ~13K to our bundle size (5K after Brotli) thus is a neglibible
size contribution. We mitigate the delay in rendering the UI by having a static
placeholder in the HTML.

Required bumping the esbuild version to pick up evanw/esbuild#2349, which
makes it easier to support Preact's JSX code generation.

Fixes #5137
Fixes #5273


catzkorn:
Tagging @phirework for some eyes on this!

cmd/tsconnect/src/header.tsx
```diff
@@ -0,0 +1,40 @@
...
+        Logout
+      </button>
+    )
+  }
+  return (
+    <div class="bg-gray-100 border-b border-gray-200 pt-4 pb-2">
+      <header class="container mx-auto px-4 flex flex-row items-center">
+        <h1 class="text-3xl font-bold grow">Tailscale Connect</h1>
+        <div class="text-gray-600">{stateText}</div>
+        {logoutButton}
```
phirework, line 26: Since `logoutButton` is only assigned once and there's no early return, imo it'd be a bit more idiomatic for this to be a `{state === IPNState.Running && (...$COMPONENT)}` syntax. Not a blocker / just a preference!

cmd/tsconnect/src/app.tsx
```diff
@@ -0,0 +1,124 @@
...
+  ipn?: IPN
+  ipnState: IPNState
+  netMap?: IPNNetMap
+  browseToURL?: string
+  goPanicError?: string
+}
+
+class App extends Component<{}, AppState> {
+  state: AppState = { ipnState: IPNState.NoState }
+  #goPanicTimeout?: number
```
phirework, line 22: Why the hash?

cmd/tsconnect/src/index.ts
```diff
@@ -4,55 +4,26 @@
...
-  panicDetailNode.textContent = "Click to reload"
-  panicNode.appendChild(panicDetailNode)
-  panicNode.addEventListener("click", () => location.reload(), {
-    once: true,
-  })
-  document.body.appendChild(panicNode)
-  setTimeout(() => {
-    panicNode!.remove()
-  }, 10000)
+  app.runWithIPN(ipn)
```
phirework, line 26: This is so pleasant 😁

cmd/tsconnect/src/index.ts
```diff
@@ -4,55 +4,26 @@
...
+import { renderApp } from "./app"
 
-const go = new Go()
-WebAssembly.instantiateStreaming(
-  fetch(`./dist/${wasmUrl}`),
-  go.importObject
-).then((result) => {
+async function main() {
+  const app = await renderApp()
+  const go = new Go()
```
phirework, line 12: This is cool, how are you telling JS about Go without `globalThis`?

cmd/tsconnect/src/ssh.tsx
```diff
@@ -0,0 +1,156 @@
...
+  onDone,
+}: {
+  def: SSHSessionDef
+  ipn: IPN
+  onDone: () => void
+}) {
+  return (
+    <div
+      class="flex-grow bg-black p-2 overflow-hidden"
+      ref={(node) => {
```
phirework, line 45: It doesn't look like the function is returning anything, what is the ref handler holding onto?

cmd/tsconnect/src/app.tsx
mihaip: It's an ES2022 private field: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#ecmascript-private-fields

cmd/tsconnect/src/header.tsx
mihaip: I prefer this style because it result in less indentation and nesting of punctuation.

cmd/tsconnect/src/index.ts
mihaip: It's added by the Go/WASM type definitions: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/golang-wasm-exec/index.d.ts

We specify the types as dev dependencies in package.json (https://github.com/tailscale/tailscale/blob/b81d219b24dc49ca91967837fddef274c71be51d/cmd/tsconnect/package.json#L6), and TypeScript automatically picks them up.

cmd/tsconnect/src/ssh.tsx
mihaip: I'm using `ref` as a callback to know when the component has been rendered and there's a DOM node I can attach the xterm.js instance to. If there's a more idiomatic way with to do with hooks I'm happy to switch (we never did a full hooks port at my old job, so I don't have much production experience with them).

cmd/tsconnect/src/app.tsx
phirework: TIL! Thanks.

cmd/tsconnect/src/ssh.tsx
phirework: Ohhh that's clever. The more idiomatic way would be to combine a `useRef` on this div with a `useEffect` that has that ref in the dependency array, like what we're doing with the [ACL editor](https://github.com/tailscale/corp/blob/main/control/adminhttp/panel/src/components/acl-editor/acl-editor.tsx#L56) but I think the logic is the same in the end. Up to you if you want to switch!

phirework (approved):

cmd/tsconnect/src/ssh.tsx
mihaip: Keeping this for now mostly so I can merge the commit as is (don't have a dev environment on my vacation machine, but I don't want to have this PR block Charlotte's work by having that be done on top of the old DOM code). Will look into switching to hooks next week.

PR #5301, opened 2022-08-04
tsnet: cleanup resources upon start failure

dsnet:
In a partially initialized state, we should cleanup
all prior resources when an error occurs.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -10,6 +10,7 @@ package tsnet
 import (
 	"context"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"net"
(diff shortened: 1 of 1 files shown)


tsnet/tsnet.go
```diff
@@ -357,9 +366,22 @@ func (s *Server) start() error {
 			logf("localapi serve error: %v", err)
 		}
 	}()
+	closePool.add(s.localAPIListener)
```
bradfitz, line 369: this one's kinda pointless?

bradfitz (approved):

tsnet/tsnet.go
```diff
@@ -182,7 +183,10 @@ func (s *Server) getAuthKey() string {
 	return os.Getenv("TS_AUTHKEY")
 }
 
-func (s *Server) start() error {
+func (s *Server) start() (reterr error) {
+	var closePool closeOnErrorPool
+	defer closePool.closeAllIfError(&reterr)
```
Soypete, line 188: so just to make sure I understand, if an error is returned, this. loop does not run

Soypete (approved):

tsnet/tsnet.go
dsnet: The set of closed resources matches what occurs in `Server.Shutdown`. It's probably worth keeping so that there's correlation between the two.

tsnet/tsnet.go
dsnet: Correct.

PR #5314, opened 2022-08-06
util/must: rename Do->Get, add Do

maisem:

### DIFF
--- a/util/must/must.go
+++ b/util/must/must.go
@@ -5,11 +5,20 @@
 // Package must assists in calling functions that must succeed.
 //
 // Example usage:
-//	var target = must.Do(url.Parse(...))
+//
+//	var target = must.Get(url.Parse(...))
+//	must.Do(close())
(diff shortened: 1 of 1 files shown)


dsnet (approved):
SGTM.

One word of caution is that it would be tempting to write:
```go
f := must.Get(os.Open(...))
defer must.Do(f.Close())
```
but this does not do what the user expects since `f.Close()` is called immediately and the returned error is then passed to the deferred call to `must.Do`.

bradfitz (approved):

bradfitz:
> One word of caution is that it would be tempting to write:
>
> f := must.Get(os.Open(...))
> defer must.Do(f.Close())
> but this does not do what the user expects since f.Close() is called immediately and the returned error is then passed to the deferred call to must.Do.

If that becomes common, we could always add `must.Close(io.Closer)` too.

PR #5315, opened 2022-08-06
tka: stable text representation of AUMHash

crawshaw:
This makes debugging easier, you can pass an AUMHash to a printf and get
a string that is easy to debug.

Also rearrange how directories/files work in the FS store: use the first
two characters of the string representation as the prefix directory, and
use the entire AUMHash string as the file name. This is again to aid
debugging: you can `ls` a directory and line up what prints out easily
with what you get from a printf in debug code.

### DIFF
--- a/tka/tailchonk.go
+++ b/tka/tailchonk.go
@@ -6,8 +6,6 @@ package tka
 
 import (
 	"bytes"
-	"encoding/base32"
-	"encoding/hex"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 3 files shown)


tka/tailchonk.go
```diff
@@ -297,7 +299,7 @@ func (c *FS) scanHashes(eachHashInfo func(*fsHashInfo)) error {
 		if !prefix.IsDir() {
 			continue
 		}
-		pb, err := hex.DecodeString(prefix.Name())
+		pb, err := base32StdNoPad.DecodeString(prefix.Name())
```
twitchyliquid64, line 302: Now that the hash is fully in the (file) filename, do we need to decode the directory name at all anymore? And I guess the other assumption (which i have no idea if it holds) is that 2 characters is always a valid base32 symbol?

tka/tailchonk.go
```diff
@@ -190,7 +188,8 @@ func ChonkDir(dir string) (*FS, error) {
 // fsHashInfo describes how information about an AUMHash is represented
 // on disk.
 //
-// The CBOR-serialization of this struct is stored to base/hex(hash[0])/base32(hash[1:])
+// The CBOR-serialization of this struct is stored to base/XX/base32(hash)
+// where AA are the first two characters of base32(hash).
```
twitchyliquid64, line 192: where XX?

twitchyliquid64 (approved):

tka/tailchonk.go
crawshaw: Removed, you're right we don't need it. Indeed, my goal here was to stop thinking of the prefix directory as base32, and just as "a prefix component of the aum hash string."

Issue #5144, opened 2022-07-22
cmd/tsconnect: Allow auth key to be passed in


mihaip:
Allow `AuthKey` for the `ipn.Options` at https://github.com/tailscale/tailscale/blob/898695e312ddbe670ac9e35c5b1ea0f4df6ab797/cmd/tsconnect/wasm/wasm_js.go#L209-L221 to be passed in.

May want to allow reading it from a query parameter in dev mode, but in prod mode we should probably have it be passed in via JS/`postMessage`, so that it doesn't get logged or otherwise leak too easily.

mihaip:
In dev mode, reading it from the query string on the JS side and then adding it to the `IPNConfig` that gets passed to `newIPN` would be a good start.

catzkorn:
Ty! Going to work on this tomorrow

mihaip:
This is done.

PR #5320, opened 2022-08-08
tsconnect: pass in authkey in dev mode

catzkorn:
This change allows for an auth key to be specified as a url query param
for use in development mode. If an auth key is specified and valid, it
will authorize the client for use immediately.

Updates #5144

### DIFF
--- a/cmd/tsconnect/src/index.ts
+++ b/cmd/tsconnect/src/index.ts
@@ -18,10 +18,18 @@ async function main() {
   go.run(wasmInstance.instance).then(() =>
     app.handleGoPanic("Unexpected shutdown")
   )
+
+  const params = new URLSearchParams(window.location.search)
+  const authKey = params.get("authkey") ?? undefined
+
   const ipn = newIPN({
     // Persist IPN state in sessionStorage in development, so that we don't need
     // to re-authorize every time we reload the page.
     stateStorage: DEBUG ? sessionStateStorage : undefined,
+    // authKey allows for an auth key to be
+    // specified as a url param which automatically
+    // authorizes the client for use.
+    authKey: DEBUG ? authKey : undefined,
   })
   app.runWithIPN(ipn)
 }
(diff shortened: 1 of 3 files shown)


cmd/tsconnect/src/index.ts
```diff
@@ -14,10 +14,22 @@ WebAssembly.instantiateStreaming(
 ).then((result) => {
   // The Go process should never exit, if it does then it's an unhandled panic.
   go.run(result.instance).then(() => handleGoPanic())
+
+  const params = new URLSearchParams(window.location.search)
+  var authKey = ""
```
mihaip, line 19: `let` is generally preferred to `var` because it had better scoping rules. https://eslint.org/docs/latest/rules/no-var describes some of the tradeoffs and could be used to automatically check for this, I'm pretty sure we already have it for the admin portal. Will investigate when I'm back.

(I do wish Go, Swift and JS would agree on what `var` and `let` mean, switching between quickly makes my head hurt)

mihaip: Though in this case you can say as a one-liner and have it be `const`:

```
const authKey = params.get("authkey") ?? undefined
```

`??` is the [nullish-coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator).

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -69,6 +69,12 @@ func newIPN(jsConfig js.Value) map[string]any {
 		store = &jsStateStore{jsStateStorage}
 	}
 
+	jsAuthKey := jsConfig.Get("authKey")
+	var authKey string
+	if !jsAuthKey.IsUndefined() && jsAuthKey.Type() == js.TypeString {
```
mihaip, line 74: I think `jsAuthKey.Type() == js.TypeString` is all you need. Undefined is its own type in JS (and mapped to `js.TypeUndefined` in the Go JS bindings).

mihaip (approved):

PR #5322, opened 2022-08-09
net/dns/resolver: return SERVFAIL when no upstream resolvers set

maisem:
Otherwise we just keep looping over the same thing again and again.

```
dns udp query: upstream nameservers not set
dns udp query: upstream nameservers not set
dns udp query: upstream nameservers not set
```

### DIFF
--- a/net/dns/resolver/forwarder.go
+++ b/net/dns/resolver/forwarder.go
@@ -76,8 +76,6 @@ const (
 	wellKnownHostBackupDelay = 200 * time.Millisecond
 )
 
-var errNoUpstreams = errors.New("upstream nameservers not set")
-
 // txid identifies a DNS transaction.
 //
 // As the standard DNS Request ID is only 16 bits, we extend it:
@@ -205,7 +203,7 @@ type forwarder struct {
 	//
 	// That is, if we're running on GCP or AWS where there's always a well-known
 	// IP of a recursive resolver, return that rather than having callers return
-	// errNoUpstreams. This fixes both normal 100.100.100.100 resolution when
+	// SERVFAIL. This fixes both normal 100.100.100.100 resolution when
 	// /etc/resolv.conf is missing/corrupt, and the peerapi ExitDNS stub
 	// resolver lookup.
 	cloudHostFallback []resolverAndDelay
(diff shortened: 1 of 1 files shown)


net/dns/resolver/forwarder.go
```diff
@@ -698,7 +696,17 @@ func (f *forwarder) forwardWithDestChan(ctx context.Context, query packet, respo
 		resolvers = f.resolvers(domain)
 		if len(resolvers) == 0 {
 			metricDNSFwdErrorNoUpstream.Add(1)
-			return errNoUpstreams
+			res, err := nxDomainResponse(query)
```
DentonGentry, line 699: It would be good to log something for Support to be able to easily see that there is a problem preventing DNS from working. It doesn't need to log a line for every query.

bradfitz (approved):
Code seems fine at least. Will defer to others on whether this is correct.

net/dns/resolver/forwarder.go
maisem: There's already a metric that we can expose for that.

danderson (commented):
To be correct per DNS, I think this needs to return SERVFAIL, not NXDOMAIN. NXDOMAIN is an authoritative "the domain you asked for doesn't exist", SERVFAIL is "I'm unable to give you an answer for your query".

twitchyliquid64 (approved):

maisem:
Trying to figure out if it is indeed SERVFAIL or NODATA

net/dns/resolver/forwarder.go
DentonGentry: We have the fleetwide metrics coalesced in dashboards. I don't know how to see metrics from an individual client. Do we have a reasonable way for support to look at this?

DentonGentry (changes requested):
would like to address the question about support seeing that something is going wrong before this gets submitted.

net/dns/resolver/forwarder.go
```diff
@@ -205,7 +203,7 @@ type forwarder struct {
 	//
 	// That is, if we're running on GCP or AWS where there's always a well-known
 	// IP of a recursive resolver, return that rather than having callers return
-	// errNoUpstreams. This fixes both normal 100.100.100.100 resolution when
+	// NXDOMAIN. This fixes both normal 100.100.100.100 resolution when
```
andrew-d, line 206:
```suggestion
	// SERVFAIL. This fixes both normal 100.100.100.100 resolution when
```

?

maisem:
>would like to address the question about support seeing that something is going wrong before this gets submitted.

I've added a log line, but it is going to slightly less spammy than what it was as it won't cause internal retries. PTAL

DentonGentry (approved):

net/dns/resolver/forwarder.go
```diff
@@ -698,7 +696,18 @@ func (f *forwarder) forwardWithDestChan(ctx context.Context, query packet, respo
 		resolvers = f.resolvers(domain)
 		if len(resolvers) == 0 {
 			metricDNSFwdErrorNoUpstream.Add(1)
-			return errNoUpstreams
+			f.logf("no upstream resolvers set, returning SERVFAIL")
+			res, err := servfailResponse(query)
+			if err != nil {
+				f.logf("building servfail response: %v", err)
+				return nil
```
bradfitz, line 703: why's this nil? add a comment? it looks like a mistake otherwise.

maisem: Sent https://github.com/tailscale/tailscale/pull/5326

PR #5323, opened 2022-08-09
tailcfg, control/controlclient: make Debug settings sticky in a map session [capver 37]

### LINKED ISSUE #4843: tailcfg: MapResponse.Debug is kinda a mess
Unlike all other fields, MapResponse.Debug of nil doesn't mean no change.

But we can't easily fix that, because half the fields on `type Debug` are settings and have are imperative commands to the client.

At minimum, we should add struct tags to each field to distinguish which are which and add a `CloneSettings` method that returns a clone of `*Debug` (mapping nil to nil) and only copying the settings, not the commands.

/cc @maisem

bradfitz:
Fixes #4843

### DIFF
--- a/control/controlclient/map.go
+++ b/control/controlclient/map.go
@@ -15,6 +15,7 @@ import (
 	"tailscale.com/types/key"
 	"tailscale.com/types/logger"
 	"tailscale.com/types/netmap"
+	"tailscale.com/types/opt"
 	"tailscale.com/wgengine/filter"
 )
 
@@ -46,6 +47,7 @@ type mapSession struct {
 	lastDomain             string
 	lastHealth             []string
 	lastPopBrowserURL      string
+	stickyDebug            tailcfg.Debug // accumulated opt.Bool values
 
 	// netMapBuilding is non-nil during a netmapForResponse call,
 	// containing the value to be returned, once fully populated.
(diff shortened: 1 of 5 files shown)


control/controlclient/map.go
```diff
@@ -114,6 +116,29 @@ func (ms *mapSession) netmapForResponse(resp *tailcfg.MapResponse) *netmap.Netwo
...
+	debug := resp.Debug
+	if debug != nil {
+		if debug.RandomizeClientPort {
+			debug.SetRandomizeClientPort.Set(true)
+		}
+		if debug.ForceBackgroundSTUN {
+			debug.SetForceBackgroundSTUN.Set(true)
+		}
+		copyDebugOptBools(&ms.stickyDebug, debug)
+	}
```
maisem, line 128: else if?

tailcfg/tailcfg.go
```diff
@@ -1335,6 +1336,15 @@ type Debug struct {
 	// periodicReSTUN), regardless of inactivity.
 	ForceBackgroundSTUN bool `json:",omitempty"`
 
+	// SetForceBackgroundSTUN controls whether magicsock should
+	// always do its background STUN queries (see magicsock's
+	// periodicReSTUN), regardless of inactivity.
+	//
+	// As for capver 37, this field is what's set on the fire
```
maisem, line 1343: 🔥

control/controlclient/map.go
bradfitz: done

control/controlclient/map.go
```diff
@@ -114,6 +116,28 @@ func (ms *mapSession) netmapForResponse(resp *tailcfg.MapResponse) *netmap.Netwo
...
+	if debug != nil {
+		copyDebugOptBools(debug, &ms.stickyDebug)
+		if !debug.ForceBackgroundSTUN {
+			debug.ForceBackgroundSTUN, _ = ms.stickyDebug.SetForceBackgroundSTUN.Get()
+		}
+		if !debug.RandomizeClientPort {
+			debug.RandomizeClientPort, _ = ms.stickyDebug.SetRandomizeClientPort.Get()
+		}
+	}
```
maisem, line 139: this will only turn it on, but not off?

tailcfg/tailcfg.go
maisem: s/fire/wire

tailcfg/tailcfg.go
```diff
@@ -1365,6 +1375,15 @@ type Debug struct {
 	// fixed port.
 	RandomizeClientPort bool `json:",omitempty"`
 
+	// SetRandomizeClientPort is whether magicsock should UDP bind to
+	// :0 to get a random local port, ignoring any configured
+	// fixed port.
+	//
+	// As for capver 37, this field is what's set on the fire
```
maisem, line 1382: same

control/controlclient/map.go
bradfitz: This is denormalizing the opt.Bool field into the bool. The `debug` field here is freshly allocated (either from the JSON parse or the `new` above, so it would never be true unless control set it true, in which case that means the server hasn't been updated to use the new field and we should respect the server's bool from the pre-opt.Bool field.

maisem: ah right

maisem (approved):

PR #5327, opened 2022-08-09
.github/workflows: use a matrix for staticcheck GOOS/GOARCHes

bradfitz:


.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,108 @@
...
+      env:
+        SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      if: failure() && github.event_name == 'push'
+
+  staticcheck:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        goos: [linux, windows, darwin]
+        goarch: [amd64]
```
DentonGentry, line 69: Would darwin ARM be likely to catch things which darwin amd64 does not?

DentonGentry (approved):

.github/workflows/static-analysis.yml
bradfitz: Nothing in oss currently that would select different files.

PR #5330, opened 2022-08-09
Multiple instances in windows

amandal-cn:


.github/workflows/release-windows.yml
```diff
@@ -0,0 +1,58 @@
+name: Build and Release [windows]
```
DentonGentry, line 1: We're not going to use this for releases, and I'd ask to remove it from the PR please.

DentonGentry (changes requested):

paths/paths_windows.go
```diff
@@ -99,5 +99,5 @@ func ensureStateDirPerms(dirPath string) error {
 // LegacyStateFilePath returns the legacy path to the state file when it was stored under the
 // current user's %LocalAppData%.
 func LegacyStateFilePath() string {
-	return filepath.Join(os.Getenv("LocalAppData"), "Tailscale", "server-state.conf")
+	return filepath.Join(os.Getenv("LocalAppData"), "Tailscale", "server-state-trilogy.conf")
```
DentonGentry, line 102: This is a backwards compatibility feature for very old versions of the Windows client. You shouldn't update this filename, those old versions did not have "trilogy" in the filename.

PR #5336, opened 2022-08-10
cmd/derper: fix data race & server panic in manual cert mode

### LINKED ISSUE #4082: Derper TLS handshake error in China: remote error: tls: internal error
### What is the issue?

It is very strange, i build an custom derper server but it always randomly broken after some times of running.
It usually be 1~6 days depends on usage. High pressure broken it sooner.
even if the option '-verify-clients' is enable, derper might broken within one hour.

Go version i used is 1.17.1 and derper server we build follow the official manual.

Then it was broken, syslog will has dozens of error logs like:
`Mar  7 11:59:52 localhost derper[268829]: 2022/03/07 11:59:52 http: TLS handshake error from 111.12.xx.xx:51574: remote error: tls: internal error`

and after about 5-10 minus then will start to generated tons of error like:
`Mar  7 12:05:00 localhost derper[268829]: 2022/03/07 21:52:56 http: TLS handshake error from 42.90.xx.xx:34461: write tcp 172.xx.xx.xx:12341->42.90.xx.xx:34461: write: connection reset by peer`

When the derper server broken, i can still see that derper marked as available in `tailscale netcheck` but actually it cannot be used, so i lost connection with many devices.

### Steps to reproduce

Use letsencrypt to generate a cert.
Build a derper server using that cert and with option: `-stun -a :12345 -certmode manual -verify-clients --certdir /xxx/derper-certs -http-port -1 -hostname xx.xxx.com` 
Wait hours or days until broken.

_No response_

### OS

Linux

### OS version

centos8

### Tailscale version

derper@v1.1.1-0.20220225000201 & tailscale@1.22.0

### Bug report

derper has no bugreport option

bradfitz:
(Thanks for debugging, Roland!)

Fixes #4082

### DIFF
--- a/cmd/derper/cert.go
+++ b/cmd/derper/cert.go
@@ -20,6 +20,11 @@ var unsafeHostnameCharacters = regexp.MustCompile(`[^a-zA-Z0-9-\.]`)
 
 type certProvider interface {
 	// TLSConfig creates a new TLS config suitable for net/http.Server servers.
+	//
+	// The returned Config must have a GetCertificate function set and that
+	// function must return a unique *tls.Certificate for each call. The
+	// returned *tls.Certificate will be mutated by the caller to append to the
+	// (*tls.Certificate).Certificate field.
 	TLSConfig() *tls.Config
 	// HTTPHandler handle ACME related request, if any.
 	HTTPHandler(fallback http.Handler) http.Handler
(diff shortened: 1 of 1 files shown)


maisem (approved):

IT-dragon:
i use port 81 with v1.30.0.But still recreate this error.

DentonGentry:
The problem is in the derper binary, which you compile yourself.
Did you update the source code and build derper again?

If you did build the derper from current source code and still see the problem, please open an issue https://github.com/tailscale/tailscale/issues

PR #5337, opened 2022-08-10
util/sha256: new package

dsnet:
The hash.Hash provided by sha256.New is much more efficient
if we always provide it with data a multiple of the block size.
This avoids double-copying of data into the internal block
of sha256.digest.x. Effectively, we are managing a block ourselves
to ensure we only ever call hash.Hash.Write with full blocks.

Performance:

	name    old time/op    new time/op    delta
	Hash    33.5µs ± 1%    20.6µs ± 1%  -38.40%  (p=0.000 n=10+9)

The logic has gone through CPU-hours of fuzzing.

### DIFF
--- a/util/sha256x/sha256.go
+++ b/util/sha256x/sha256.go
@@ -1,0 +1,149 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package sha256x is like crypto/sha256 with extra methods.
+// It exports a concrete Hash type
+// rather than only returning an interface implementation.
+package sha256x
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+	"hash"
+)
+
+var _ hash.Hash = (*Hash)(nil)
+
+// Hash is a hash.Hash for SHA-256,
+// but has efficient methods for hashing fixed-width integers.
+type Hash struct {
+	// The optimization is to maintain our own block and
+	// only call h.Write with entire blocks.
+	// This avoids double-copying of buffers within sha256.digest itself.
+	// However, it does mean that sha256.digest.x goes unused,
+	// which is a waste of 64B.
+
+	h  hash.Hash              // always *sha256.digest
+	x  [sha256.BlockSize]byte // equivalent to sha256.digest.x
+	nx int                    // equivalent to sha256.digest.nx
+}
+
+func New() *Hash {
+	return &Hash{h: sha256.New()}
+}
+
+func (h *Hash) Write(b []byte) (int, error) {
+	h.HashBytes(b)
+	return len(b), nil
+}
+
+func (h *Hash) Sum(b []byte) []byte {
+	if h.nx > 0 {
+		// This causes block mis-alignment. Future operations will be correct,
+		// but are less efficient until Reset is called.
+		h.h.Write(h.x[:h.nx])
+		h.nx = 0
+	}
+	return h.h.Sum(b)
+}
+
+func (h *Hash) Reset() {
+	h.h.Reset()
+	h.nx = 0
+}
+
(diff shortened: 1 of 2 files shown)


util/sha256/sha256.go
```diff
@@ -0,0 +1,144 @@
+// Package sha256 is like crypto/sha256,
```
maisem, line 1: license

bradfitz, line 1: can we give this a different package name that doesn't collide with crypto/sha256?

util/sha256/sha256_test.go
```diff
@@ -0,0 +1,146 @@
+package sha256
```
maisem, line 1: same

util/sha256/sha256.go
```diff
@@ -0,0 +1,144 @@
...
+	// NOTE: This method is carefully written to be inlineable.
+	if h.nx <= len(h.x)-1 {
+		h.x[h.nx] = n
+		h.nx += 1
+	} else {
+		h.hashUint8Slow(n)
+	}
+}
+
+//go:noinline
```
bradfitz, line 69: why?

bradfitz: oh, because otherwise it blocks HashUint8 from being inlined?

util/sha256/sha256.go
dsnet: I called it `sha256fixed`. Happy to hear alternative names.

util/sha256/sha256.go
dsnet: Yep...

dsnet:
Wow... you two are insanely fast.

util/sha256/sha256.go
dsnet: I'll add a comment to the effect.

util/sha256fixed/sha256.go
```diff
@@ -0,0 +1,148 @@
...
+	// This avoids double-copying of buffers within sha256.digest itself.
+	// However, it does mean that sha256.digest.x goes unused,
+	// which is a waste of 64B.
+
+	h  hash.Hash              // always *sha256.digest
+	x  [sha256.BlockSize]byte // equivalent to sha256.digest.x
+	nx int                    // equivalent to sha256.digest.nx
+}
+
+func New() *Hash {
```
bradfitz, line 31: how much of the win is that we're using a concrete type and not going through interface indirect calls?

util/sha256fixed/sha256.go
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package sha256fixed is like crypto/sha256,
+// but efficiently hashes fixed-width integers.
+package sha256fixed
```
bradfitz, line 7: sha256x? "Package sha256 is like crypto/sha256 with extra methods. It exports a concrete Hasher
type rather than only returning an interface implementation."

bradfitz: (Because I suspect this'll grow more stuff than just a few integer hashing methods)

util/sha256fixed/sha256.go
dsnet: The 40% win is purely from avoiding the double copy.
The benchmarks currently go through an interface to keep them fairer.
When we consider the fact that we will actually be calling concrete methods,
the performance gains are even better at 140%. 😄

util/sha256fixed/sha256.go
dsnet: SGTM.

util/sha256fixed/sha256.go
bradfitz: hell yeah

bradfitz (approved):

PR #5339, opened 2022-08-11
util/deephash: use sha256x

dsnet:
Switch deephash ot use sha256x.Hash.

We also modify sha256x.Hash to export the underlying hash.Hash
for testing purposes so that we can intercept all hash.Hash calls.

Performance:

	Hash-24                19.8µs ± 1%    19.2µs ± 1%  -3.01%  (p=0.000 n=10+10)
	HashPacketFilter-24    2.61µs ± 0%    2.53µs ± 1%  -3.01%  (p=0.000 n=8+10)
	HashMapAcyclic-24      31.3µs ± 1%    29.8µs ± 0%  -4.80%  (p=0.000 n=10+9)
	TailcfgNode-24         1.83µs ± 1%    1.82µs ± 2%    ~     (p=0.305 n=10+10)
	HashArray-24            344ns ± 2%     323ns ± 1%  -6.02%  (p=0.000 n=9+10)

The performance gains is not as dramatic as sha256x over sha256 due to:
1. Most of the hashing already occuring through the direct memory hashing logic, and
2. what does not go through direct memory hashing is slowed down by reflect.


util/deephash/deephash.go
```diff
@@ -1010,5 +972,14 @@ func (h *hasher) hashType(t reflect.Type) {
...
+	h.HashUint64(uint64(rtypeAddr))
+}
+
+func unsafeStringToBytes(s string) []byte {
+	type stringHeader struct {
+		p unsafe.Pointer
+		n int
+	}
+	p := (*stringHeader)(unsafe.Pointer(&s))
+	return unsafe.Slice((*byte)(p.p), p.n)
```
bradfitz, line 984: These are always super scary. This is an accidental refactor or two away from being a major problem with random crashes & corruption.

Can we instead use go4.org/mem and have sha256x have a HashMem(mem.RO) method?

Or at least make sha256x have a HashString method and do this unsafety very locally such that the []byte doesn't escape out to a caller?

dsnet: I went with adding `sha256x.Hash.HashString` instead. Now the unsafe cast is contained in just that method, which should be easier to reason about.

util/sha256x/sha256.go
```diff
@@ -24,13 +25,16 @@ type Hash struct {
 	// However, it does mean that sha256.digest.x goes unused,
 	// which is a waste of 64B.
 
-	h  hash.Hash              // always *sha256.digest
+	// H is the underlying hash.Hash.
+	// The hash.Hash.BlockSize must be equal to sha256.BlockSize.
+	// It is exported only for testing purposes.
+	H  hash.Hash              // usually a *sha256.digest
```
bradfitz, line 31: How about keep this unexported by add a package-level exported variable of unexported type, like `util/clientmetric` does next door to this directory:

```
util/clientmetric/clientmetric.go:var TestHooks testHooks
util/clientmetric/clientmetric.go:type testHooks struct{}
util/clientmetric/clientmetric.go:func (testHooks) ResetLastDelta() {
```

Then add a method like `(testHooks).SetHashUnderlyingHash(*Hash, hash.Hash)` and have the caller test code type assert that on the `testHooks` type.

Keeps the API clean looking.

bradfitz (approved):

util/sha256x/sha256.go
dsnet: I think I'm going to export this for general use in the future.

I'm realizing that `sha256x` is probably name since this isn't actually about `sha256`, but rather a concrete implementation of `hash.Hash` with a particular block size.

According to https://en.wikipedia.org/wiki/Comparison_of_cryptographic_hash_functions#Parameters, common hash functions have the following block sizes (in bytes):
* 32B: GOST, PANAMA
* 64B: BLAKE2s, BLAKE3, MD4, MD5, RIPEMD, RIPEMD-128, RIPEMD-256, RIPEMD-160, RIPEMD-320, SHA-0, SHA-1, SHA-224, SHA-256, Tiger(2), Whirpool
* 128B: BLAKE2b, HAVAL, SHA-384, SHA-512

However, SHA3 is really odd in that it doesn't have a block size that's a power of 2. Yuck.

dsnet: Right now, the `Hash` type has a hardcoded block of 64B, which makes it suitable for most of the crypto hashes. If we increase it to 128B, it would be suitable for all crypto hashes except SHA3.

bradfitz: But you still submitted this saying "It is exported only for testing purposes." when I suggested an alternative way to test.

PR #5344, opened 2022-08-11
client,cmd/tailscale,ipn,tka,types: implement tka initialization flow

twitchyliquid64:
This PR implements the client-side of initializing network-lock with the Coordination server.

Meaningful pieces:
 - `tailcfg` changes: Describes the two RPCs with control to initialize network-lock. These are sent over the noise protocol.
 - `cmd/tailscale` changes: Implements `tailscale lock init` and `tailscale lock status`.

Most of the rest is just glue.

I don't actually initialize the local tka (ie: `ipnlocal.LocalBackend.tka`) because I figure control will drive that, like for other nodes that arent the ones doing the initialization.

/cc @maisem as we were discussing this


ipn/ipnstate/ipnstate.go
```diff
@@ -67,6 +67,16 @@ type Status struct {
 	User map[tailcfg.UserID]tailcfg.UserProfile
 }
 
+type NetworkLockStatus struct {
+	Enabled bool
```
bradfitz, line 71: docs on both these

ipn/ipnstate/ipnstate.go
```diff
@@ -67,6 +67,16 @@ type Status struct {
...
+
+	// Head describes the AUM hash of the leaf AUM.
+	Head *[32]byte
```
bradfitz, line 74: what does nil here mean?

ipn/ipnstate/ipnstate.go
```diff
@@ -67,6 +67,16 @@ type Status struct {
...
+
+	// Pubkey describes the nodes' network-lock public key.
+	Pubkey key.NLPublic
```
bradfitz, line 77: elsewhere we spell this out as PublicKey

client/tailscale/localclient.go
```diff
@@ -680,6 +681,42 @@ func (lc *LocalClient) Ping(ctx context.Context, ip netip.Addr, pingtype tailcfg
...
+		return nil, fmt.Errorf("error: %w", err)
+	}
+	pr := new(ipnstate.NetworkLockStatus)
+	if err := json.Unmarshal(body, pr); err != nil {
+		return nil, err
+	}
+	return pr, nil
+}
+
+// NetworkLockInit Initializes the tailnet key authority.
```
bradfitz, line 697: lowercase initializes

ipn/ipnstate/ipnstate.go
twitchyliquid64: `Head` is nil if network lock is not enabled. Added a comment.

ipn/ipnstate/ipnstate.go
twitchyliquid64: Updated.

cmd/tailscale/cli/network-lock.go
```diff
@@ -0,0 +1,101 @@
...
+		return errors.New("network-lock is already enabled")
+	}
+
+	// Parse the set of initially-trusted keys.
+	// Keys are specified using their key.NLPublic.MarshalText representation,
+	// with an optional '#<votes>' suffix.
+	var keys []tka.Key
+	for i, a := range args {
+		var key key.NLPublic
+		spl := strings.Split(a, "#")
```
sailorfrag, line 49: `#` is probably not going to make it through the shell unquoted anyway, maybe the suffix should use a different character? It would be unfortunate for someone to do `tailscale lock init key1#1 key2#1 key3#1` and only get `key1` because the rest were treated as a comment

sailorfrag, line 49: `SplitN` n=2?

ipn/localapi/localapi.go
```diff
@@ -791,6 +796,54 @@ func (h *Handler) serveUploadClientMetrics(w http.ResponseWriter, r *http.Reques
 	json.NewEncoder(w).Encode(struct{}{})
 }
 
+func (h *Handler) serveTkaStatus(w http.ResponseWriter, r *http.Request) {
```
sailorfrag, line 799: should this one be checking the method too? Perhaps localapi should be calling this as a `GET`?

cmd/tailscale/cli/network-lock.go
twitchyliquid64: Good catch, I'll use `?` instead. There aren't good options given that `:` is taken, but I can see why the pound/bang is particularly heinous.

cmd/tailscale/depaware.txt
```diff
@@ -1,9 +1,13 @@
 tailscale.com/cmd/tailscale dependencies: (generated by github.com/tailscale/depaware)
 
+        filippo.io/edwards25519                                      from github.com/hdevalence/ed25519consensus
+        filippo.io/edwards25519/field                                from filippo.io/edwards25519
    W 💣 github.com/alexbrainman/sspi                                 from github.com/alexbrainman/sspi/negotiate+
    W    github.com/alexbrainman/sspi/internal/common                 from github.com/alexbrainman/sspi/negotiate
    W 💣 github.com/alexbrainman/sspi/negotiate                       from tailscale.com/net/tshttpproxy
+        github.com/fxamacker/cbor/v2                                 from tailscale.com/tka
```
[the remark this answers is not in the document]
twitchyliquid64: I've gone through them already as part of building TKA, they seem appropriate for me:

1. The cbor package is the most battle-tested and well-used of the cbor encoding packages.
2. The remaining packages are for `eed25519consensus` and its deps, which are Filippo recommended and also pretty used in the crypto space, and seem fine to me.

cmd/tailscale/cli/network-lock.go
twitchyliquid64: Ooooo, TIL! DOone.

ipn/localapi/localapi.go
twitchyliquid64: Done.

Issue #4880, opened 2022-06-18
FR: nginx-auth with shared nodes


mshunshin:
### What is the issue?

nginx-auth works as you would expect with computers on your own tailnet.

However, when sharing a computer with another person (we are both gmail.com users) the code fails.

This is because the info.Node struct in the code does not have the same format when the user is external - consequently the go code that extracts the tailnet name from the username fails and the code bails out

info.Node.Name for when I am accessing it on my own network
matt-laptop2.matthew-shunshin.gmail.com.beta.tailscale.net.

info.Node.Name when someone I have shared the computer with accesses it
""

The code should probably chop up the UserProfile.LoginName 
strings.Split(info.UserProfile.LoginName, "@")[1]
to get it.

### Steps to reproduce

look at the output of tailscale.nginx-auth service when you access a computer on your tailnet, and one where someone you have shared it with does.

### Are there any recent changes that introduced the issue?

no

### Tailscale version

1.26.0

### Bug report

BUG-3b620f32633d2f7553615175634f51778c810efb16504ab33d486205e514790c-20220618152948Z-b02eb7ac2fdfda89

DentonGentry:
Possibly fixed by https://github.com/tailscale/tailscale/pull/5346

mshunshin:
I don't think it does as the merged code doesn't appear to source the user information to return in the shared node case.

willnorris:
Oh interesting.. I hadn't seen this issue when I did #5346.  I was focusing on node information (which is intentionally not provided for shared nodes), but it sounds like you're running into issues with user info?  I'll poke around some more.

willnorris:
actually, reading a little more closely, maybe we are talking about the same thing.  Information about the "sharee" node is intentionally removed, and it can't simply be parsed from the login name.  What node information were you trying to get at?

mshunshin:
oops - actually it does solve it - I had patched my copy to mangle the Tailscale-Tailnet header as that was the quickest way to get it working.

but really - I can now look at the Tailscale-Login line in all cases - shared or not.

Thanks!

willnorris:
awesome, glad to hear it!

PR #5346, opened 2022-08-11
cmd/nginx-auth: allow use of shared nodes

willnorris:
When sharing nodes, the name of the sharee node is not exposed (instead it is hardcoded to "device-of-shared-to-user"), which means that we can't determine the tailnet of that node.  Don't immediately fail when that happens, since it only matters if "Expected-Tailnet" is used.

### DIFF
--- a/cmd/nginx-auth/nginx-auth.go
+++ b/cmd/nginx-auth/nginx-auth.go
@@ -63,17 +63,24 @@ func main() {
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


Xe (approved):

mshunshin:
Hi - I don't think this entirely solvers the problem in #4880 as the merged commit doens't appear to source data on the user in the shared node case.

Issue #5161, opened 2022-07-25
net/dnscache: use Go 1.19's Resolver.PreferGo


bradfitz:
Tracking bug to use the new PreferGo stuff in Go 1.19 for our DNS caching on all platforms (now that it works on Windows in Go 1.19)

https://tip.golang.org/doc/go1.19#net

see golang/go#33097

DentonGentry:
Tailscale at `HEAD` uses Go 1.19 now.

DentonGentry:
https://github.com/tailscale/tailscale/pull/4979/ added some support for PreferGo in dnscache.go

bradfitz:
This bug is really about redoing all our currently-kinda-rando DNS caching inside tailscaled to use a DNS-protocol-aware (record-level) cache using `net.Dialer.Resolver` to our own impl with `net.Resolver.Dial` and `net.Resolver.PreferGo` set true.

PR #5349, opened 2022-08-11
net/dnscache: try IPv6 addresses first

andrew-d:
I saw our graph of IPv6 vs IPv4 connections, and I suspect that this is the reason that we're seeing so many v4 connections. It looks like we're trying the first IP and then waiting, which means we're trying an IPv4 address first and not actually racing a v6 connection.

### DIFF
--- a/net/dnscache/dnscache.go
+++ b/net/dnscache/dnscache.go
@@ -523,6 +523,21 @@ func (dc *dialCall) raceDial(ctx context.Context, ips []netip.Addr) (net.Conn, e
 		return nil, errors.New("no IPs")
 	}
 
+	// Partition candidate list and then merge such that an IPv6 address is
+	// in the first spot if present, and then addresses are interleaved.
+	// This ensures that we're trying an IPv6 address first, then
+	// alternating between v4 and v6 in case one of the two networks is
+	// broken.
+	var iv4, iv6 []netip.Addr
+	for _, ip := range ips {
+		if ip.Is6() {
+			iv6 = append(iv6, ip)
+		} else {
+			iv4 = append(iv4, ip)
+		}
+	}
+	ips = interleaveSlices(iv6, iv4)
+
 	go func() {
 		for i, ip := range ips {
 			if i != 0 {
@@ -578,6 +593,21 @@ func (dc *dialCall) raceDial(ctx context.Context, ips []netip.Addr) (net.Conn, e
 			return nil, ctx.Err()
 		}
 	}
+}
+
+// interleaveSlices combines two slices of the form [a, b, c] and [x, y, z]
+// into a slice with elements interleaved; i.e. [a, x, b, y, c, z].
+func interleaveSlices[T any](a, b []T) []T {
+	var (
+		i   int
+		ret = make([]T, 0, len(a)+len(b))
+	)
+	for i = 0; i < len(a) && i < len(b); i++ {
+		ret = append(ret, a[i], b[i])
+	}
+	ret = append(ret, a[i:]...)
+	ret = append(ret, b[i:]...)
+	return ret
 }
 
 func v4addrs(aa []net.IPAddr) (ret []netip.Addr) {
--- a/net/dnscache/dnscache_test.go
+++ b/net/dnscache/dnscache_test.go
@@ -140,3 +140,27 @@ func TestResolverAllHostStaticResult(t *testing.T) {
 		t.Errorf("bad dial error got %q; want %q", got, want)
 	}
 }
+
+func TestInterleaveSlices(t *testing.T) {
+	testCases := []struct {
+		name string
+		a, b []int
+		want []int
+	}{
+		{name: "equal", a: []int{1, 3, 5}, b: []int{2, 4, 6}, want: []int{1, 2, 3, 4, 5, 6}},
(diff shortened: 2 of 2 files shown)


net/dnscache/dnscache.go
```diff
@@ -523,6 +523,21 @@ func (dc *dialCall) raceDial(ctx context.Context, ips []netip.Addr) (net.Conn, e
 		return nil, errors.New("no IPs")
 	}
 
+	// Partition candidate list and then merge such that an IPv6 address is
+	// in the first spot if present, and then addresses are interleaved.
+	// This ensures that we're trying an IPv6 address first, then
+	// alternating between v4 and v6 in case one of the two networks is
+	// broken.
+	var iv4, iv6 []netip.Addr
```
maisem, line 531: Won't a sort be better?

maisem: Oh I see, you're alternating.

andrew-d: I don't actually want these IPs to be sorted, since otherwise we'll direct the majority of the traffic to the lowest-sorting IP. I'd actually rather randomly shuffle these to spread the load around, though I figured that could be a follow-up.

andrew-d: At least for now, keeping the original presumably-at-least-sorta-random order and then merging them such that there's an IPv6 addr first means that we're not making things any worse.

maisem: I meant StableSort with a func that uses the protocol as the Less func.

net/dnscache/dnscache.go
```diff
@@ -523,6 +523,21 @@ func (dc *dialCall) raceDial(ctx context.Context, ips []netip.Addr) (net.Conn, e
...
+	for _, ip := range ips {
+		if ip.Is6() {
+			iv6 = append(iv6, ip)
+		} else if ip.Is4() {
```
maisem, line 535: do you need the else if? if so comment

net/dnscache/dnscache.go
```diff
@@ -580,6 +595,19 @@ func (dc *dialCall) raceDial(ctx context.Context, ips []netip.Addr) (net.Conn, e
 	}
 }
 
+func mergeSlices[T any](a, b []T) []T {
```
maisem, line 598: should have a comment on what the expected output is. like merge based on what

maisem: maybe `interleaveSlices`?

andrew-d: Much better name :+1:

net/dnscache/dnscache_test.go
```diff
@@ -140,3 +140,26 @@ func TestResolverAllHostStaticResult(t *testing.T) {
...
+		{a: []int{1, 3, 5}, b: []int{2, 4, 6}, want: []int{1, 2, 3, 4, 5, 6}},
+		{a: []int{1, 3, 5}, b: []int{2, 4}, want: []int{1, 2, 3, 4, 5}},
+		{a: []int{1, 3}, b: []int{2, 4, 6}, want: []int{1, 2, 3, 4, 6}},
+		{a: []int{1}, b: []int{2, 4, 6}, want: []int{1, 2, 4, 6}},
+		{a: nil, b: []int{2, 4, 6}, want: []int{2, 4, 6}},
+		{a: nil, b: nil, want: nil},
+	}
+
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("TestCase%d", i), func(t *testing.T) {
```
maisem, line 158: test cases should be named, numbering is annoying because it requires counting when debugging.

maisem (approved):

net/dnscache/dnscache.go
```diff
@@ -580,6 +595,21 @@ func (dc *dialCall) raceDial(ctx context.Context, ips []netip.Addr) (net.Conn, e
 	}
 }
 
+// interleaveSlices combines two slices of the form [a, b, c] and [x, y, z]
+// into a slice with elements interleaved; i.e. [a, x, b, y, c, z].
+func interleaveSlices[T any](a, b []T) []T {
+	var (
+		i   int
+		ret []T
```
bradfitz, line 603: you could size this correctly to start

bradfitz (commented):
I'm out right now. I'd like to re-read all this code again before, especially as I want to redo all this for https://github.com/tailscale/tailscale/issues/5161

andrew-d:
@bradfitz - sure, happy to hold off on merging until you're back

bradfitz:
> @bradfitz - sure, happy to hold off on merging until you're back

If you and @maisem are happy already feel free to merge. I can re-read all this code later. It's not like we're cutting a new release imminently.

net/dnscache/dnscache.go
bradfitz: Yeah, StableSort SGTM. That also means we don't allocate two more slices. (Assuming we own ips here... don't remember)

maisem: the issue with StableSort is that it'll make it so that all IPv6 addresses are tried first. We really do want to alternate between IPv6 and IPv4 I think.

We should also really simplify this logic, it's very hard to understand. Not that I know how to simplify just yet.

PR #5351, opened 2022-08-12
util/sha256x: rename Hash as Block512

dsnet:
Rename Hash as Block512 to indicate that this is a general-purpose
hash.Hash for any algorithm that operates on 512-bit block sizes.

A subsequent commit will move the sha256x package to hashx.
This is done separately to avoid confusing git.

### DIFF
--- a/util/sha256x/sha256.go
+++ b/util/sha256x/sha256.go
@@ -2,51 +2,76 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// Package sha256x is like crypto/sha256 with extra methods.
-// It exports a concrete Hash type
-// rather than only returning an interface implementation.
-package sha256x
+// Package hashx provides a concrete implementation of [hash.Hash]
+// that operates on a particular block size.
+package hashx
 
 import (
-	"crypto/sha256"
 	"encoding/binary"
+	"fmt"
 	"hash"
 	"unsafe"
 )
 
-var _ hash.Hash = (*Hash)(nil)
+var _ hash.Hash = (*Block512)(nil)
 
-// Hash is a hash.Hash for SHA-256,
-// but has efficient methods for hashing fixed-width integers.
-type Hash struct {
-	// The optimization is to maintain our own block and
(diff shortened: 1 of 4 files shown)


util/sha256x/sha256.go
```diff
@@ -2,79 +2,92 @@
...
+	// never allows the provided buffer to escape.
+	hash.Hash
+
+	x  [512 / 8]byte
+	nx int
 }
 
-func New() *Hash {
-	return &Hash{H: sha256.New()}
+func New(h hash.Hash) (*Block512, error) {
```
bradfitz, line 48: New512?

Presumably this package will grow other sizes based on the name of the package and Block512?

dsnet: Good catch.

util/sha256x/sha256.go
```diff
@@ -2,79 +2,92 @@
...
+	b := &Block512{Hash: h}
+	if len(b.x)%h.BlockSize() != 0 {
+		return b, fmt.Errorf("hashx.Block512: inefficient use of hash.Hash with %d-bit block size", 8*h.BlockSize())
```
bradfitz, line 51: So that's an error? But not an error?

Docs on this function?

Above it says:

```
// The hash still operates correctly even with misaligned block sizes,
// but operates less efficiently.
```

So why does this return an error?

util/sha256x/sha256.go
dsnet: I was trying to have `Block` be dynamically sized, but I couldn't get `HashUintXXX` to be inlineable no matter how much rearranging I tried since the extra operate to fetch the slice length goes above the budget.

util/sha256x/sha256.go
dsnet: I don't feel strongly whether this should return an error. The only reason to use `hashx` is for performance, so I figured it would be a good sanity check that someone was not passing misaligned algorithms.

dsnet: Do the latest set of changes look alright? Thanks!

bradfitz (approved):

PR #5352, opened 2022-08-12
util/deephash: use binary encoding of time.Time

dsnet:
Formatting a time.Time as RFC3339 is slow (see https://go.dev/issue/54093).

Now that we have efficient hashing of fixed-width integers,
just hash the time.Time as a binary value.

Performance:

	Hash-24                19.0µs ± 1%    18.6µs ± 1%   -2.03%  (p=0.000 n=10+9)
	TailcfgNode-24         1.79µs ± 1%    1.40µs ± 1%  -21.74%  (p=0.000 n=10+9)

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -11,6 +11,9 @@
 // The definition of equality is identical to reflect.DeepEqual except:
 //   - Floating-point values are compared based on the raw bits,
 //     which means that NaNs (with the same bit pattern) are treated as equal.
+//   - time.Time are compared based on whether they are the same instant in time
+//     and also in the same zone offset. Monotonic measurements and zone names
+//     are ignored as part of the hash.
 //   - Types which implement interface { AppendTo([]byte) []byte } use
 //     the AppendTo method to produce a textual representation of the value.
 //     Thus, two values are equal if AppendTo produces the same bytes.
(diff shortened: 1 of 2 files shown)


util/deephash/deephash.go
```diff
@@ -525,9 +525,13 @@ func (h *hasher) hashComplex128v(v addressableValue) bool {
...
-	b := t.AppendFormat(h.scratch[:1], time.RFC3339Nano)
-	b[0] = byte(len(b) - 1) // more than sufficient width; if not, good enough.
-	h.HashBytes(b)
+	h.HashUint64(uint64(t.Unix()))
+	h.HashUint32(uint32(t.Nanosecond()))
+	// TODO(joetsai): Do we care about the time zone? Including it means that
+	// t1.Equal(t2) is not identical to Hash(t1) == Hash(t2),
+	// which is an odd semantic.
+	_, offset := t.Zone()
+	h.HashUint32(uint32(offset))
```
dsnet, line 534: Personally, I'm surprised that we include the time zone as part of the hash. I vote that we remove it. However, I don't know whether any usages of `deephash` depend on the zone being part of the comparison or not.

bradfitz: Remember that deephash was originally defined as the SHA256 of the JSON we sent to clients, because the protocol is JSON, and if the JSON changed, we wanted the hash to change.

With that in mind, consider:

https://go.dev/play/p/n17DwNjMp_q

Those times are equal, but have different JSON.

Also keep in mind that the Swift date/time parser is very picky and we've had 2 or 3 big bugs in the past because we sent the wrong time.Time down to Go which passed through to Swift and broke the Mac GUI.

So I'd like to continue to be paranoid about hashing everything in a time.Time.

dsnet: Done. I documented the desired semantic for `time.Time`.

dsnet: And added a fuzz test to verify that `Hash(t1) == Hash(t2)` is semantically equivalent to `t1.Format(RFC3339Nano) == t2.Format(RFC3339Nano)`.

I'll run it for an hour before submitting.

dsnet: 1 hour later. 650M timestamps fuzzed, no issues.

bradfitz: Lgtm

bradfitz:
LGTM

PR #5357, opened 2022-08-12
control/controlclient,tailcfg: capver 40 create KeySignature field on tailcfg.Node

twitchyliquid64:
We carve out a space to put the node-key signature (eventually will be used on tailnets where network lock is enabled).


tailcfg/tailcfg.go
```diff
@@ -1797,6 +1801,10 @@ type PeerChange struct {
 	// Key, if non-nil, means that the NodeID's wireguard public key changed.
 	Key *key.NodePublic `json:",omitempty"`
 
+	// KeySignature, if non-nil, means that the signature of the wireguard
+	// public key has changed.
+	KeySignature tkatype.MarshaledSignature `json:",omitempty"`
```
bradfitz, line 1806: Don't add to PeerChange now. This would need to be in its own change with a capver bump.

twitchyliquid64: Will do!

Why is it unsafe without masking it behind a capver? I would have thought it had no effect when unset.

bradfitz: It's just kinda confusing in the history, why there's a no-op unused field.

control/controlclient/map.go
```diff
@@ -310,6 +310,9 @@ func undeltaPeers(mapRes *tailcfg.MapResponse, prev []*tailcfg.Node) {
 				if v := ec.Capabilities; v != nil {
 					n.Capabilities = *v
 				}
+				if v := ec.KeySignature; v != nil {
+					n.KeySignature = v
+				}
```
bradfitz, line 315: oh, you did make it work.

okay, just bump capver then and put [capver NN] in the commit message?

bradfitz: and you can probably probably the deephash mention in the commit message. That's not the main point of the change.

twitchyliquid64: Done.

bradfitz (approved):

PR #5358, opened 2022-08-12
net/stun: convert to use net/netip.AddrPort

### LINKED ISSUE #5281: net/stun: convert to use net/netip.AddrPort
Make this use netip.AddrPort:

```
func ParseResponse(b []byte) (tID TxID, addr []byte, port uint16, err error) {
```

And this too:

```
func Response(txID TxID, ip net.IP, port uint16) []byte {
```

Part of #5162

skriptble:
Convert ParseResponse and Response to use netip.AddrPort instead of
net.IP and separate port.

Fixes #5281

### DIFF
--- a/net/stun/stun.go
+++ b/net/stun/stun.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"hash/crc32"
 	"net"
+	"net/netip"
 )
 
 const (
@@ -151,20 +152,18 @@ func foreachAttr(b []byte, fn func(attrType uint16, a []byte) error) error {
 }
 
 // Response generates a binding response.
-func Response(txID TxID, ip net.IP, port uint16) []byte {
-	if ip4 := ip.To4(); ip4 != nil {
-		ip = ip4
-	}
+func Response(txID TxID, addrPort netip.AddrPort) []byte {
+	addr := addrPort.Addr()
+
 	var fam byte
-	switch len(ip) {
-	case net.IPv4len:
+	if addr.Is4() {
 		fam = 1
-	case net.IPv6len:
+	} else if addr.Is6() {
 		fam = 2
-	default:
+	} else {
 		return nil
 	}
-	attrsLen := 8 + len(ip)
+	attrsLen := 8 + addr.BitLen()/8
 	b := make([]byte, 0, headerLen+attrsLen)
 
 	// Header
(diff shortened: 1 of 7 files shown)


net/stun/stun.go
```diff
@@ -151,20 +152,21 @@ func foreachAttr(b []byte, fn func(attrType uint16, a []byte) error) error {
...
+		bytLen = 4
+	} else if addr.Is6() {
 		fam = 2
-	default:
+		bytLen = 16
+	} else {
 		return nil
 	}
-	attrsLen := 8 + len(ip)
+	attrsLen := 8 + bytLen
```
bradfitz, line 169: instead of `bytLen` you could also use `addr.BitLen()/8` (https://go.dev/play/p/_OMhPgqfq8T)

net/stun/stun.go
```diff
@@ -175,12 +177,12 @@ func Response(txID TxID, ip net.IP, port uint16) []byte {
...
 	b = appendU16(b, attrXorMappedAddress)
-	b = appendU16(b, uint16(4+len(ip)))
+	b = appendU16(b, uint16(4+bytLen))
 	b = append(b,
 		0, // unused byte
 		fam)
-	b = appendU16(b, port^0x2112) // first half of magicCookie
-	for i, o := range []byte(ip) {
+	b = appendU16(b, addrPort.Port()^0x2112) // first half of magicCookie
+	for i, o := range addr.AsSlice() {
```
bradfitz, line 185: this allocates, which is not ideal

you could do:

```
ipa := addr.As16()
for i, o := range ipa[16-bytLen:] {
```

net/stun/stun.go
```diff
@@ -225,40 +226,48 @@ func ParseResponse(b []byte) (tID TxID, addr []byte, port uint16, err error) {
 				return err
 			}
 			if len(a) == 16 {
-				addr6, port6 = a, p
+				var ip6 [16]byte
+				copy(ip6[:], a)
```
bradfitz, line 230: no need for these explicit copies as of Go 1.17: https://go.dev/doc/go1.17#language

If you want you can write:

```
     netip.AddrFrom16(*(*[16]byte)([]byte(a)))
```

But it'll be about identical performance.

bradfitz (approved):

net/stun/stun.go
skriptble: I was thinking about that but didn't know if it'd be too messy. I like this better.

net/stun/stun.go
skriptble: Updated!

net/stun/stun.go
skriptble: Ah, I was looking for this syntax. Updated!

net/stun/stun.go
```diff
@@ -193,24 +193,23 @@ func Response(txID TxID, ip net.IP, port uint16) []byte {
 // ParseResponse parses a successful binding response STUN packet.
 // The IP address is extracted from the XOR-MAPPED-ADDRESS attribute.
 // The returned addr slice is owned by the caller and does not alias b.
```
bradfitz, line 195: Docs need updating here.

But there's no addr _slice_ returned anymore. And it's a value type, so no need to document ownership or aliasing.

bradfitz (approved):
Still LGTM, other than doc nit. But no need to wait for me to review that again before you submit.

PR #5359, opened 2022-08-12
.github/workflows: delete flaky windows-race.yml

bradfitz:
It flakes more often than it runs. It provides no value and builds
failure blindness, making people get used to submitting on red.

Bye.


skriptble (approved):
Hooray! Now my CI checks will be green 😄

PR #5360, opened 2022-08-12
tka: Use strict decoding settings, implement Unserialize()

twitchyliquid64:
Unfortunately I can't use the regular names `MarshalBinary()` and `UnmarshalBinary()` due to cbor semantics :(

### DIFF
--- a/tka/aum.go
+++ b/tka/aum.go
@@ -212,6 +212,10 @@ func (a *AUM) StaticValidate() error {
 }
 
 // Serialize returns the given AUM in a serialized format.
+//
+// We would implement encoding.BinaryMarshaler, except that would
+// unfortunately get called by the cbor marshaller resulting in infinite
+// recursion.
 func (a *AUM) Serialize() []byte {
 	// Why CBOR and not something like JSON?
 	//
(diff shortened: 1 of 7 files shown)


tka/aum.go
```diff
@@ -212,6 +212,10 @@ func (a *AUM) StaticValidate() error {
 }
 
 // Serialize returns the given AUM in a serialized format.
+//
+// We would implement encoding.BinaryMarshaler, except that would
+// unfortunately get called by the cbor marshaller resulting in infinite
```
sailorfrag, line 217: Does the cbor marshaller have a `CBORMarshal()` or something that you could implement to get around that; I believe that's the way to avoid the same problem with JSON and `TextMarshaler`

Hmm. Or I guess you could do something like:
```golang
type aumData struct {
  // Actual fields go here
}

type AUM aumData

func (a *AUM) MarshalBinary() (data byte[], err error) {
  return doCborEncooding((*aumData)(a))
}
```

Maybe this is fine as-is; I'm just trying to come up with ideas for _that_ particular problem raised in the comments.

sailorfrag (approved):

tka/aum.go
twitchyliquid64: Unfortunately no :( I gave the `aumData` approach a go, but it ends up feeling very fragile & I was worried we would introduce complexity/bugs just to solve this little thing.

Issue #1659, opened 2021-04-06
Resolution of single-label names via quad-100 is laggy for Windows


danderson:
Found while implementing split DNS on Windows.

Setup: enable MagicDNS on your domain, configure a resolver other than your LAN's resolver (e.g. 8.8.8.8), add search domains other than MagicDNS (e.g. tailscale.com), enable split DNS mode. Ensure your LAN resolver also sets some search domain, other than the ones set by Tailscale (e.g. universe.tf).

Dataflow: requests for `<domain>.beta.tailscale.net` and `tailscale.com` flow to the 100.100.100.100 resolver on localhost. `<domain>.beta.tailscale.net` gets handled by the MagicDNS hosts table, `tailscale.com` gets forwarded to 8.8.8.8. Everything else falls through to the LAN resolvers.

Open a command prompt, and start pinging single-label things whose names resolve through these different paths:
 - `ping go` -> `go.universe.tf`, resolution is ~instant via LAN resolvers and pings begin.
 - `ping derp1` -> `derp1.tailscale.com`, resolution via 100.100.100.100 forwarded to 8.8.8.8, but resolution stalls for 1-2s before pings begin.
 - `ping acrux` (a MagicDNS name) -> `acrux.natulte.net.beta.tailscale.net`, but resolution stalls for 1-2s before pings begin.

Additional spam logging in tailscaled shows MagicDNS receiving A and AAAA queries promptly, and returning responses promptly (either forwarded from 8.8.8.8, or from its internal hosts table). So, seemingly the delay is not in tailscaled itself, but some external process.

Separately, Wireshark shows Windows sending LLMNR queries to the LAN, simultaneously with the MagicDNS queries. It does _not_ send LLMNR requests for names that resolve through the regular LAN resolver, only for our queries.

Hypothesis: waiting on LLMNR timeout is what's slowing us down.

Question: why do only our queries get LLMNR'd? What's different in our config vs. the default DNS config?

Rant: Microsoft does not provide documentation on the algorithm used for resolving single-label DNS names, so we're left guessing.

Datapoint: I turned off LLMNR on the Tailscale network adapter (since it can't multicast anyway). This didn't change anything.

danderson:
@bradfitz remembers that Perkeep once hit something about maybe SOA records relating to custom DNS resolvers, and will add details here once dug up.

danderson:
Also, disabling LLMNR globally didn't help, so it seems to be something else.

bradfitz:
[...]

PR #5361, opened 2022-08-12
net/dns: [win] respond with SERVFAIL queries when no resolvers

maisem:
Currently we forward unmatched queries to the default resolver on
Windows. This results in duplicate queries being issued to the same
resolver which is just wasted.

Updates #1659


raggi:
Do we want to have this behavior on all platforms?

maisem:
> Do we want to have this behavior on all platforms?

This is already the behavior on other platforms, Windows was just special.

raggi (approved):

PR #5343, opened 2022-08-11
go.mod: bump wireguard/windows, which moves to using net/netip

bradfitz:
Updates #5162

PR #5364, opened 2022-08-12
net/dns: call LUID.SetDNS instead of writing to registry directly

maisem:
This calls the approrpiate Win32 API to do the same.

Updates #1659

### DIFF
--- a/net/dns/manager_windows.go
+++ b/net/dns/manager_windows.go
@@ -7,10 +7,10 @@ package dns
 import (
 	"errors"
 	"fmt"
+	"net"
 	"net/netip"
 	"os/exec"
 	"sort"
-	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


bradfitz:
Typo of "appropriate" in commit message/PR text.

net/dns/manager_windows.go
```diff
@@ -134,53 +143,35 @@ func (m windowsManager) setPrimaryDNS(resolvers []netip.Addr, domains []dnsname.
...
-			return err
-		}
-	} else if err := key4.SetStringValue("SearchList", strings.Join(domStrs, ",")); err != nil {
+	// Disable LLMNR on the Tailscale interface. We don't do
+	// multicast, and we certainly don't do LLMNR, so it's pointless
+	// to make Windows try it.
+	//
+	// TODO(maisem): The same effect can be achieved by setting
+	// winipcfg.DnsInterfaceSettings.EnableLLMNR to 0. Currently upstream does
+	// not support setting the DnsInterfaceSettingsFlagEnableLLMNR flag, use it
```
danderson, line 160: Is there an upstream bug/PR somewhere? Link?

net/dns/manager_windows.go
```diff
@@ -118,14 +118,23 @@ func (m windowsManager) setSplitDNS(resolvers []netip.Addr, domains []dnsname.FQ
...
+	tsLUID, err := winipcfg.LUIDFromGUID(&tsGUID)
+	if err != nil {
+		return err
+	}
+	ips4 := make([]net.IP, 0, len(resolvers))
+	ips6 := make([]net.IP, 0, len(resolvers))
 	for _, ip := range resolvers {
 		if ip.Is4() {
-			ipsv4 = append(ipsv4, ip.String())
+			b := ip.As4()
```
danderson, line 133: If we're upstreaming changes to winipcfg, maybe also propose one to use net/netip types?

danderson (approved):

dblohm7:
@maisem Could you please clarify what the impetus is for doing this? Is the concern that using the API might allow the changes to take effect more quickly than directly editing the registry?

net/dns/manager_windows.go
maisem: It already does - https://github.com/tailscale/tailscale/pull/5343

net/dns/manager_windows.go
maisem: Not yet, will file.

maisem:
this was a shot in the dark when trying to figure out windows dns stuff, don't think its necessary any more.

PR #5362, opened 2022-08-12
util/winutil: consolidate interface specific registry keys

maisem:
Code movement to allow reuse in a follow up PR.

Updates #1659

Signed-off-by: Maisem Ali <maisem@tailscale.com>

PR #5365, opened 2022-08-12
net/dns: disable NetBIOS on Tailscale interfaces

maisem:
Like LLMNR, NetBIOS also adds resolution delays and we don't support it
anyway so just disable it on the interface.

Updates https://github.com/tailscale/tailscale/issues/1659

### DIFF
--- a/net/dns/manager_windows.go
+++ b/net/dns/manager_windows.go
@@ -167,9 +167,9 @@ func (m windowsManager) setPrimaryDNS(resolvers []netip.Addr, domains []dnsname.
 		return err
 	}
 
-	// Disable LLMNR on the Tailscale interface. We don't do
-	// multicast, and we certainly don't do LLMNR, so it's pointless
-	// to make Windows try it.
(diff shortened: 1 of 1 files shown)


maisem:
The first commit is basically #5362.

net/dns/manager_windows.go
```diff
@@ -207,10 +207,14 @@ func (m windowsManager) SetDNS(cfg OSConfig) error {
 	// configuration only, routing one set of things to the "split"
 	// resolver and the rest to the primary.
 
-	// Unconditionally disable dynamic DNS updates on our interfaces.
+	// Unconditionally disable dynamic DNS updates and NetBIOS on our
+	// interfaces.
```
bradfitz, line 211: Say why a bit? Maybe even reference a bug for more background reading?

bradfitz (approved):
Second commit LGTM (with a bit more comments added), assuming the first one goes away after #5362 is in.

PR #5366, opened 2022-08-13
net/dns: [win] set Nameserver on Tailscale interface when SNR is disabled

maisem:
Windows has a feature called Smart Name Resolution. When it is on, for
short name DNS queries (i.e. queries without a suffix) Windows issues
A and AAAA requests on all interfaces which have a DNS server associated
with them.

These unicast requests are issued in parallel for each interface, and within
each interface they are issued in serial for each Search Domain on the
interface. Querying an interface stops once a Search Domain returns a
positive response.

  1. If there is a positive response from the Nameserver on the primary interface
     (one with the lowest metric); it is used and all other queries stop.

  2. If none of the Search Domains on the primary interface return a
     positive response; it issues MDNS and LLMNR queries in parallel on
     all interfaces and waits for either a positive response from these
     a timeout of about 1s.

  3. If a positive multicast response is returned from the primary interface,
     that is returned.

  4. For each interface in ascending metric order:
     a. If there is any positive DNS response it is returned.
     b. If there is any positive multicast response it is returned.

  5. Else; it returns "no such host".

When SNR is off, unicast queries on all interfaces are preferred over
multicast queries. As such, multicast queries are not even issued until
all interfaces are first exhausted with unicast queries.

  1. Like previously, unicast requests are issued in parallel for each interface, and within
     each interface they are issued in serial for each Search Domain on the
     interface. Querying an interface stops once a Search Domain returns
     a positive response.

  2. In case of positive responses, the one from the lowest interface
     metric wins.

  3. If none of the unicast queries across all interfaces return a
     response; it issues MDNS and LLMNR queries in parallel on
     all interfaces.

  4. In case of positive responses, the one from the lowest interface
metric wins.

  5. Else it times out after 1s; and returns "no such host".

Today we do not set 100.100.100.100 as the Nameserver on the Tailscale interface.
This results in Tailscale getting the lowest priority in Search Domains, which
in turn results in a 2.3s of short name queries that resolve over MagicDNS
as Windows waits for MDNS/LLMNR queries to timeout from other
interfaces.

Setting the nameserver results in Tailscale getting all queries from the OS and
becoming the primary resolver as it has the lowest interface metric (5).
This resolves the 2.3s delay and short name queries now resolve
immediately.

This also means that 100.100.100.100 receives queries for domains it did
not sign up for (e.g. google.com), we now (1cff719) get around this by
returning SERVFAIL for those queries making Windows use the configured
resolver on other interfaces.

This unfortunately introduces a 1s delay for short name queries that
would previously resolve instanteously from the local network. These
queries are now slow as Windows waits for a timeout for the MDNS query
it sent over the primary interface (Tailscale). (Note we already disable
LLMNR on the Tailscale interface).

The way around the 1s delay is to globally turn off SNR which is not
something we can do for all users, however we can detect that we are in
that mode and configure the OS correctly so that when SNR is turned off
all short name queries resolve instantly. When SNR is turned on, we
still have the 2.3s delay for MagicDNS short names but at least we can
write a KB article to explain this behavior.

Updates #1659

### DIFF
--- a/net/dns/manager_windows.go
+++ b/net/dns/manager_windows.go
@@ -250,20 +250,31 @@ func (m windowsManager) setPrimaryDNS(resolvers []netip.Addr, domains []dnsname.
 	return nil
 }
 
+// isSmartNameResolutionDisabled reports whether Multi-homed Smart Name
+// Resolution is enabled in the Windows Registry. We use this to decide
+// how to configure the DNS servers on the Tailscale interface.
+func (m windowsManager) isSmartNameResolutionDisabled() bool {
+	key, err := winutil.OpenKeyWait(registry.LOCAL_MACHINE, `SOFTWARE\Policies\Microsoft\Windows NT\DNSClient`, registry.SET_VALUE)
+	if err != nil {
+		return false
+	}
+	defer key.Close()
+	val, _, err := key.GetIntegerValue("DisableSmartNameResolution")
+	return err == nil && val == 1
+}
+
 func (m windowsManager) SetDNS(cfg OSConfig) error {
 	// We can configure Windows DNS in one of two ways:
 	//
-	//  - In primary DNS mode, we set the NameServer and SearchList
-	//    registry keys on our interface. Because our interface metric
-	//    is very low, this turns us into the one and only "primary"
-	//    resolver for the OS, i.e. all queries flow to the
-	//    resolver(s) we specify.
-	//  - In split DNS mode, we set the Domain registry key on our
-	//    interface (which adds that domain to the global search list,
-	//    but does not contribute other DNS configuration from the
-	//    interface), and configure an NRPT (Name Resolution Policy
-	//    Table) rule to route queries for our suffixes to the
-	//    provided resolver.
+	// - Primary DNS mode, we set the NameServer and SearchList registry keys
+	// on our interface. Because our interface metric is very low, this turns us
+	// into the "primary" resolver for the OS.
+	//
+	// - Split DNS mode, we only set the Domain registry key on our interface
+	// (which adds that domain to the global search list, but does not
+	// contribute other DNS configuration from the interface), and configure an
+	// NRPT (Name Resolution Policy Table) rule to route queries for our
+	// suffixes to the provided resolver.
 	//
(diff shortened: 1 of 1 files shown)


net/dns/manager_windows.go
```diff
@@ -219,22 +219,46 @@ func (m windowsManager) SetDNS(cfg OSConfig) error {
...
-		return errors.New("cannot set per-domain resolvers on Windows 7")
 	} else {
 		if err := m.setSplitDNS(cfg.Nameservers, cfg.MatchDomains); err != nil {
 			return err
 		}
-		// Still set search domains on the interface, since NRPT only
-		// handles query routing and not search domain expansion.
+	}
+
+	isSmartNameResolutionDisabled := func() bool {
```
danderson, line 235: Standalone helper method? Doesn't seem to have to be an embedded func.

net/dns/manager_windows.go
```diff
@@ -219,22 +219,46 @@ func (m windowsManager) SetDNS(cfg OSConfig) error {
...
+		key, err := m.openKey(`SOFTWARE\Policies\Microsoft\Windows NT\DNSClient`)
+		if err != nil {
+			return false
+		}
+		defer key.Close()
+		val, _, err := key.GetIntegerValue("DisableSmartNameResolution")
+		return err == nil && val == 1
+	}()
+
+	// We still need to set both the search domains and nameservers on the
```
danderson, line 245: Top-level comment is now wrong, we don't always set nameserver. Needs a bit of rewording so the first paragraph only talks about search domains, and the next paragraph talks about when/why we set Nameserver.

danderson (approved):

dblohm7 (approved):
OK once @danderson's issues are resolved.

sailorfrag:
In case it's useful for future reference, Microsoft has [some docs](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2008-R2-and-2008/dd197552(v=ws.10)?redirectedfrom=MSDN#part-2-querying-a-dns-server) that cover at least part of this

maisem:
Closing this as it never got merged

PR #5367, opened 2022-08-13
util/deephash: completely refactor

dsnet:
DO NOT SUBMIT

This is an experiment to see where we would want this to go.
If we're already using unsafe to hash memory,
then embrace the use of unsafe and stay out of reflect.

Rather than a typeHasherFunc that operates on

	func(*hasher, addressableValue)

we operate on

	func(*hasher, unsafe.Pointer)

where the unsafe.Pointer is identical to addressableValue.UnsafePointer.

We drop support for interface{ AppendTo([]byte) []byte }.
While that was more efficient in the prior pure-reflect implementation,
it is no longer true today. Instead, rely on direct memory hashing
of crypto-key types that have this method, and/or add direct
support for types that cannot be memory hashed
(for example time.Time and netip.Addr).

Performance:

	Hash-24                18.4µs ± 1%    13.5µs ± 1%  -26.82%  (p=0.000 n=9+9)
	HashPacketFilter-24    2.52µs ± 3%    2.24µs ± 1%  -11.14%  (p=0.000 n=10+10)
	TailcfgNode-24         1.42µs ± 1%    1.35µs ± 1%   -4.99%  (p=0.000 n=10+10)
	HashArray-24            324ns ± 1%     312ns ± 1%   -3.76%  (p=0.000 n=9+10)

SHA-256 hashing now takes up ~55% of the benchmark time.
We will get an additional boost in performance when Go's crypto/sha256
package uses native instructions. See https://go.dev/cl/408795.

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -14,9 +14,6 @@
 //   - time.Time are compared based on whether they are the same instant in time
 //     and also in the same zone offset. Monotonic measurements and zone names
 //     are ignored as part of the hash.
-//   - Types which implement interface { AppendTo([]byte) []byte } use
-//     the AppendTo method to produce a textual representation of the value.
-//     Thus, two values are equal if AppendTo produces the same bytes.
 //
 // WARNING: This package, like most of the tailscale.com Go module,
 // should be considered Tailscale-internal; we make no API promises.
@@ -27,8 +24,7 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
-	"log"
-	"math"
+	"net/netip"
 	"reflect"
 	"sync"
 	"time"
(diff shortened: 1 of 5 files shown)


dsnet:
I'd like your thoughts on this general approach before I spend more time on it. I'll try to break it apart into smaller changes if possible, but I got carried away in refactoring.

Major changes:
* `typeHasherFunc` takes an `unsafe.Pointer` instead of a `addressableValue`.
* We unify the slow and fast paths, resulting in about ~300 LOC deleted. It also means that our unit tests are more thorough since we don't have gaps all over all coverage since the logic decided to go through the slow path or fast path unexpectedly.
* We drop support for `interface{ AppendTo([]byte) []byte }` since it was always slower than binary hashing.
* We add support for hashing `netip.Addr`, which was the only type that really used `AppendTo`.
* Cycle detection for slices is now correctly implemented.
* The `isRecursiveType` and `canMemHash` functions are mostly unchanged.

Performance wise, we are now limited by `crypto/sha256` (takes up 55% of CPU time), which I think is a great result.

dsnet:
The latest commit wraps `unsafe.Pointer` in a `pointer` abstraction. When built with `-race`, it performs type and bounds checking for every operation performed upon it. I would argue that this is now safer than the logic before. See `pointer_race.go`.

bradfitz:
I didn't get a chance to review this yet. Busy weekend & people over working in my garage today. Will look on flight.

dsnet:
Take your time. I think I found a path to gradually get to this state over multiple PRs.

dsnet:
HEAD is pretty close to what this PR does. I think HEAD is actually cleaner now.

HEAD is just missing a few edgecases with regard to slices.

PR #5371, opened 2022-08-14
net/dns: [win] add MagicDNS entries to etc/hosts

maisem:
This works around the 2.3s delay in short name lookups when SNR is
enabled.
C:\Windows\System32\drivers\etc\hosts file. We only add known hosts that
match the search domains, and we populate the list in order of
Search Domains so that our matching algorithm mimics what Windows would
otherwise do itself if SNR was off.

Updates #1659


net/dns/manager_windows.go
```diff
@@ -111,6 +115,55 @@ func (m windowsManager) setSplitDNS(resolvers []netip.Addr, domains []dnsname.FQ
...
+		footer = "# TailscaleHostsSectionEnd"
+	)
+	var out bytes.Buffer
+	var inSection bool
+	for sc.Scan() {
+		line := strings.TrimSpace(sc.Text())
+		if line == header {
+			if !inSection {
+				inSection = true
+			}
```
andrew-d, line 138: Is there a reason this is conditional?

net/dns/manager_windows.go
```diff
@@ -111,6 +115,55 @@ func (m windowsManager) setSplitDNS(resolvers []netip.Addr, domains []dnsname.FQ
...
+			continue
+		}
+		if line == footer {
+			if inSection {
+				inSection = false
+			}
```
andrew-d, line 144: Same q as above?

net/dns/manager_windows.go
```diff
@@ -111,6 +115,55 @@ func (m windowsManager) setSplitDNS(resolvers []netip.Addr, domains []dnsname.FQ
...
+	if len(hosts) > 0 {
+		fmt.Fprintln(&out, header)
+		fmt.Fprintln(&out)
+		for _, he := range hosts {
+			fmt.Fprintf(&out, "%s\t%s\n", he.Addr, strings.Join(he.Hosts, "\t"))
+		}
+		fmt.Fprintln(&out)
+		fmt.Fprintln(&out, footer)
+	}
+	return atomicfile.WriteFile(`C:\Windows\System32\drivers\etc\hosts`, out.Bytes(), 0 /* ignored */)
```
andrew-d, line 164: Is this going to cause problems if we don't use the Windows `\r\n` line ending convention? At the very least, it'll make it annoying if someone wants to edit this file with Notepad or something, I would think.

net/dns/manager_windows.go
maisem: ¯\_(ツ)_/¯ didn't think too much about it. I can make it unconditional if you prefer.

net/dns/manager_windows.go
maisem: Good catch, done.

andrew-d (approved):
LGTM after test imports fix; would appreciate 👀  from @danderson on general "does this make sense" grounds.

net/dns/manager.go
```diff
@@ -137,6 +139,47 @@ func (m *Manager) Set(cfg Config) error {
...
+func compileHostEntries(cfg Config) (hosts []*HostEntry) {
+	didLabel := make(map[string]bool, len(cfg.Hosts))
+	for _, sd := range cfg.SearchDomains {
+		for h, ips := range cfg.Hosts {
+			if !sd.Contains(h) || h.NumLabels() != (sd.NumLabels()+1) {
+				continue
+			}
+			ipHosts := []string{string(h.WithTrailingDot()), ""}[:1]
+			if label := dnsname.FirstLabel(string(h)); !didLabel[label] {
+				mak.Set(&didLabel, label, true)
```
danderson, line 155: didLabel was initialized explicitly, no need to use magical mak.Set here, `didLabel[label] = true` works

net/dns/manager.go
```diff
@@ -137,6 +139,47 @@ func (m *Manager) Set(cfg Config) error {
...
+// The entries are compiled in the order of the search domains, then the hosts.
+// The returned list is sorted by the first hostname in each entry.
+func compileHostEntries(cfg Config) (hosts []*HostEntry) {
+	didLabel := make(map[string]bool, len(cfg.Hosts))
+	for _, sd := range cfg.SearchDomains {
+		for h, ips := range cfg.Hosts {
+			if !sd.Contains(h) || h.NumLabels() != (sd.NumLabels()+1) {
+				continue
+			}
+			ipHosts := []string{string(h.WithTrailingDot()), ""}[:1]
```
danderson, line 153: Mysterious. I can reverse engineer that you're saving an allocation further down, but it's a really unusual construct. Comment, at least?

danderson: Also arguably unnecessary. We don't reconfigure DNS that often, a little garbage should be fine? I don't mind keeping it, it's just a bit out of place since this isn't super hot code.

net/dns/manager_windows.go
```diff
@@ -111,6 +115,65 @@ func (m windowsManager) setSplitDNS(resolvers []netip.Addr, domains []dnsname.FQ
...
+		return nil, err
+	}
+	if len(hosts) > 0 {
+		fmt.Fprintln(&out, header)
+		for _, c := range comments {
+			fmt.Fprintln(&out, c)
+		}
+		fmt.Fprintln(&out)
+		for _, he := range hosts {
+			fmt.Fprintf(&out, "%s\t%s\n", he.Addr, strings.Join(he.Hosts, "\t"))
```
danderson, line 156: Use spaces instead of tabs? Mild style preference, if windows is happy with tabs I don't really care. It's just uncommon to see tabs in the wild :)

net/dns/manager_windows.go
```diff
@@ -111,6 +115,65 @@ func (m windowsManager) setSplitDNS(resolvers []netip.Addr, domains []dnsname.FQ
...
+		}
+		fmt.Fprintln(&out)
+		fmt.Fprintln(&out, footer)
+	}
+	return bytes.ReplaceAll(out.Bytes(), []byte("\n"), []byte("\r\n")), nil
+}
+
+// setHosts sets the hosts file to contain the given host entries.
+func (m windowsManager) setHosts(hosts []*HostEntry) error {
+	b, err := os.ReadFile(`C:\Windows\System32\drivers\etc\hosts`)
```
danderson, line 166: Is C: guaranteed to be the location of the Windows install? If not, is there a portable way to locate System32?

net/dns/manager_windows.go
```diff
@@ -111,6 +115,65 @@ func (m windowsManager) setSplitDNS(resolvers []netip.Addr, domains []dnsname.FQ
...
+	if err != nil {
+		return err
+	}
+	outB, err := setTailscaleHosts(b, hosts)
+	if err != nil {
+		return err
+	}
+	return atomicfile.WriteFile(`C:\Windows\System32\drivers\etc\hosts`, outB, 0 /* ignored */)
```
danderson, line 174: "file mode, ignored on windows"

net/dns/manager_windows.go
```diff
@@ -233,11 +299,25 @@ func (m windowsManager) SetDNS(cfg OSConfig) error {
...
+		// query routing and not search domain expansion.
 		if err := m.setPrimaryDNS(nil, cfg.SearchDomains); err != nil {
 			return err
 		}
+
+		// As we are not the primary resolver in this setup, we need to
+		// explicitly set some single name hosts to ensure that we can resolve
+		// them quickly and get around the 2.3s delay that otherwise occurs due
+		// to multicast timeouts.
+		if err := m.setHosts(cfg.Hosts); err != nil {
```
danderson, line 318: What else writes etc/hosts on windows? Are we going to have to monitor the file contents and reapply DNS configuration if it gets changed? Worried about us getting into a fight with e.g. DHCP on the machine.

danderson (approved):
Looks right to me. Main concern I have is concurrent editing/overwriting, and if we're going to have to fight to maintain the contents of etc/hosts. But assuming that file never changes, this is fine.

net/dns/manager_windows.go
andrew-d: Ah, this is a good point; we probably want to use [`GetSystemDirectory`](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsystemdirectorya) or [`SHGetKnownFolderPath`](https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetknownfolderpath) with `FOLDERID_System` here?

net/dns/manager_windows.go
maisem: @raggi says there isn't anything and that's what I've seen empirical too.

PR #5374, opened 2022-08-15
.github/workflows: cancel previous CI runs on PR update

maisem:
Currently scheduled runs are not canceled when we update PRs, those
results are unused and the resources are just wasted. Instead of doing
that, this PR makes it so that when a PR is updated (force-pushed or new
commit added) the previous runs are canceled.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,5 +1,10 @@
 name: CIFuzz
 on: [pull_request]
+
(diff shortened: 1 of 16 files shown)


bradfitz (approved):

maisem:
CI failures fixed in #5375

PR #5380, opened 2022-08-15
cmd/tailscaled: enable hybrid netstack mode on openbsd too

bradfitz:
Apparently OpenBSD can forward packets with manual configuration,

https://github.com/tailscale/tailscale/issues/2498#issuecomment-1114216999

But this makes it work by default. People doing things by hand can
set TS_DEBUG_WRAP_NETSTACK=0 in the environment.


DentonGentry (approved):

DentonGentry:
The /etc/pf.conf lines in https://github.com/tailscale/tailscale/issues/2498#issuecomment-1114216999 might work on FreeBSD as well.

PR #5390, opened 2022-08-16
net/dnscache: use net/netip

### LINKED ISSUE #5282: net/dnscache: use net/netip
Lot of `net.IP` in there:

```
bradfitz@tsdev:~/src/tailscale.com/net/dnscache$ git grep net.IP
dnscache.go:    ip, ip6 net.IP
dnscache.go:    allIPs  []net.IPAddr
dnscache.go:    ip      net.IP       // either v4 or v6
dnscache.go:    ip6     net.IP       // nil if no v4 or no v6
dnscache.go:    allIPs  []net.IPAddr // 1+ v4 and/or v6
dnscache.go:func (r *Resolver) LookupIP(ctx context.Context, host string) (ip, v6 net.IP, allIPs []net.IPAddr, err error) {
dnscache.go:                    ipa := &net.IPAddr{
dnscache.go:                    return ip4, nil, []net.IPAddr{{IP: ip4}}, nil
dnscache.go:            return ip, nil, []net.IPAddr{{IP: ip}}, nil
dnscache.go:func (r *Resolver) lookupIPCache(host string) (ip, ip6 net.IP, allIPs []net.IPAddr, ok bool) {
dnscache.go:func (r *Resolver) lookupIPCacheExpired(host string) (ip, ip6 net.IP, allIPs []net.IPAddr, ok bool) {
dnscache.go:func (r *Resolver) lookupIP(host string) (ip, ip6 net.IP, allIPs []net.IPAddr, err error) {
dnscache.go:                            ips = append(ips, net.IPAddr{
dnscache.go:func (r *Resolver) addIPCache(host string, ip, ip6 net.IP, allIPs []net.IPAddr, d time.Duration) {
dnscache.go:func v4addrs(aa []net.IPAddr) (ret []netip.Addr) {
dnscache.go:func v6addrs(aa []net.IPAddr) (ret []netip.Addr) {
```

Part of #5162

skriptble:
Removes usage of net.IP and net.IPAddr where possible from net/dnscache.

Fixes #5282

Signed-off-by: Kris Brandow <kris.brandow@gmail.com>

### DIFF
--- a/net/dnscache/dnscache.go
+++ b/net/dnscache/dnscache.go
@@ -25,6 +25,8 @@ import (
 	"tailscale.com/util/singleflight"
 )
 
+var zaddr netip.Addr
+
 var single = &Resolver{
 	Forward: &net.Resolver{PreferGo: preferGoResolver()},
 }
@@ -90,14 +92,14 @@ type Resolver struct {
 
 // ipRes is the type used by the Resolver.sf singleflight group.
 type ipRes struct {
-	ip, ip6 net.IP
-	allIPs  []net.IPAddr
+	ip, ip6 netip.Addr
+	allIPs  []netip.Addr
 }
 
 type ipCacheEntry struct {
-	ip      net.IP       // either v4 or v6
-	ip6     net.IP       // nil if no v4 or no v6
-	allIPs  []net.IPAddr // 1+ v4 and/or v6
+	ip      netip.Addr   // either v4 or v6
+	ip6     netip.Addr   // nil if no v4 or no v6
+	allIPs  []netip.Addr // 1+ v4 and/or v6
 	expires time.Time
 }
 
@@ -147,34 +149,28 @@ var debug = envknob.Bool("TS_DEBUG_DNS_CACHE")
 //
 // If err is nil, ip will be non-nil. The v6 address may be nil even
 // with a nil error.
-func (r *Resolver) LookupIP(ctx context.Context, host string) (ip, v6 net.IP, allIPs []net.IPAddr, err error) {
+func (r *Resolver) LookupIP(ctx context.Context, host string) (ip, v6 netip.Addr, allIPs []netip.Addr, err error) {
 	if r.SingleHostStaticResult != nil {
 		if r.SingleHost != host {
-			return nil, nil, nil, fmt.Errorf("dnscache: unexpected hostname %q doesn't match expected %q", host, r.SingleHost)
+			return zaddr, zaddr, nil, fmt.Errorf("dnscache: unexpected hostname %q doesn't match expected %q", host, r.SingleHost)
 		}
 		for _, naIP := range r.SingleHostStaticResult {
-			ipa := &net.IPAddr{
-				IP:   naIP.AsSlice(),
-				Zone: naIP.Zone(),
+			if !ip.IsValid() && naIP.Is4() {
+				ip = naIP
 			}
-			if ip == nil && naIP.Is4() {
-				ip = ipa.IP
+			if !v6.IsValid() && naIP.Is6() {
+				v6 = naIP
 			}
-			if v6 == nil && naIP.Is6() {
-				v6 = ipa.IP
-			}
-			allIPs = append(allIPs, *ipa)
+			allIPs = append(allIPs, naIP)
 		}
 		return
 	}
-	if ip := net.ParseIP(host); ip != nil {
-		if ip4 := ip.To4(); ip4 != nil {
-			return ip4, nil, []net.IPAddr{{IP: ip4}}, nil
-		}
+	if ip, err := netip.ParseAddr(host); err == nil {
+		ip = ip.Unmap()
 		if debug {
 			log.Printf("dnscache: %q is an IP", host)
 		}
-		return ip, nil, []net.IPAddr{{IP: ip}}, nil
+		return ip, zaddr, []netip.Addr{ip}, nil
 	}
 
 	if ip, ip6, allIPs, ok := r.lookupIPCache(host); ok {
@@ -205,7 +201,7 @@ func (r *Resolver) LookupIP(ctx context.Context, host string) (ip, v6 net.IP, al
 			if debug {
 				log.Printf("dnscache: error resolving %q: %v", host, res.Err)
 			}
-			return nil, nil, nil, res.Err
+			return zaddr, zaddr, nil, res.Err
 		}
 		r := res.Val
 		return r.ip, r.ip6, r.allIPs, nil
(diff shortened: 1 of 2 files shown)


net/dnscache/dnscache.go
```diff
@@ -205,34 +201,34 @@ func (r *Resolver) LookupIP(ctx context.Context, host string) (ip, v6 net.IP, al
...
+			return netip.Addr{}, netip.Addr{}, nil, res.Err
 		}
 		r := res.Val
 		return r.ip, r.ip6, r.allIPs, nil
 	case <-ctx.Done():
 		if debug {
 			log.Printf("dnscache: context done while resolving %q: %v", host, ctx.Err())
 		}
-		return nil, nil, nil, ctx.Err()
+		return netip.Addr{}, netip.Addr{}, nil, ctx.Err()
```
maisem, line 212: maybe create 
`var zero netip.Addr`
and return that instead? it'll make it easier to read

net/dnscache/dnscache.go
```diff
@@ -394,29 +400,21 @@ func (d *dialer) DialContext(ctx context.Context, network, address string) (retC
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve %q: %w", host, err)
 	}
-	i4s := v4addrs(allIPs)
+	i4s := allIPs
```
maisem, line 403: doesn't this need to filter out IPv4s?

skriptble: From what I can tell this function wasn't filtering v4 addresses it was converting `net.IPAddr`s to `netip.Addr`s and wasn't actually removing v6 addresses. But looking at the code further down that appends the `v6addrs` (which does filter to only v6 addresses), maybe this code is supposed to do that. I'll update it to do that.

cc: @bradfitz

net/dnscache/dnscache.go
bradfitz: Or `zip` :)

skriptble: `zaddr`? :D

bradfitz: It will always be `netaddr.IP` to me :)

net/dnscache/dnscache.go
bradfitz: The current implementation of `func v4addrs` sure looks busted to me.

Definitely seems like the intent was to only keep the v4s. I wonder when it regressed. Or if it was always broken.

skriptble: Ah yeah, looks like it happened when the code was [converted to use `netip.Addr`](https://github.com/tailscale/tailscale/commit/8725b1405626bdec4d8693c9900992a5f4f00680#diff-da14a8a0bec0c36724730cc90cbbbc8eb8ef36192725dee561e02e62dbc6d9a1L586). I'll change it to return only v4 addresses and add back the v6 filtering as well.

bradfitz: Whoooops

bradfitz: Can you do the fix in a separate commit from the type change? So it's not lost in the noise.

Same PR is fine. Or separate. Your call.

skriptble: Sure, I'll throw up another PR for it.

net/dnscache/dnscache.go
```diff
@@ -25,6 +25,8 @@ import (
 	"tailscale.com/util/singleflight"
 )
 
+var zaddr netip.Addr
```
maisem, line 28: oh i meant as a var in the func, global var seems overkill?

maisem: oh i see it is used in multiple funcs

net/dnscache/dnscache.go
```diff
@@ -252,7 +250,26 @@ func (r *Resolver) lookupTimeoutForHost(host string) time.Duration {
 	return 10 * time.Second
 }
 
-func (r *Resolver) lookupIP(host string) (ip, ip6 net.IP, allIPs []net.IPAddr, err error) {
+func ip2addrs(n []net.IPAddr) []netip.Addr {
+	ips := make([]netip.Addr, 0, len(n))
+	for _, ip := range n {
+		if len(ip.IP) == 4 {
```
maisem, line 256: docs say
> Note that in this documentation, referring to an IP address as an IPv4 address or an IPv6 address is a semantic property of the address, not just the length of the byte slice: a 16-byte slice can still be an IPv4 address.

skriptble: From reading the code this looks like it's referring to 4in6 addresses. I'm fine with converting 4in6 addresses to IPv4 addresses here. Wasn't sure if we wanted to preserve the 4in6 format for some reason.

net/dnscache/dnscache.go
skriptble: Yeah, there's a bunch of instances of it, so it seemed better to just make one global declaration.

net/dnscache/dnscache.go
```diff
@@ -252,7 +250,26 @@ func (r *Resolver) lookupTimeoutForHost(host string) time.Duration {
...
+			ips = append(ips, netip.AddrFrom4(*(*[4]byte)(ip.IP)))
+		} else {
+			ips = append(ips, netip.AddrFrom16(*(*[16]byte)(ip.IP)).WithZone(ip.Zone))
+		}
+	}
+	return ips
+}
+
+func ip2addr(ip net.IP) netip.Addr {
+	if len(ip) == 4 {
```
maisem, line 266: netip.AddrFromSlice?

maisem: or https://pkg.go.dev/go4.org/netipx?utm_source=godoc#FromStdIP

skriptble: There's a `bool` that's returned from that, which we could ignore, but then we'd have silent failures if something went wrong. With this we just fail loudly because something is very wrong if we're getting a `net.IP` that isn't either 4 or 16 bytes long.

maisem: we could panic on `!ok` :)

skriptble: That's more lines of code to do the same thing :P

maisem: But it makes readability better because I can go read docs in the library.

skriptble: I don't follow, how is reading the docs for `AddrFromSlice` any better than reading the docs for `AddrFrom4` and `AddrFrom16`? If we want it to have better readability I can add a doc comment to the function explaining what it does, but I don't think that changing this to use `AddrFromSlice` adds enough. As is this is pretty straightforward code.

net/dnscache/dnscache.go
```diff
@@ -610,21 +610,20 @@ func interleaveSlices[T any](a, b []T) []T {
...
 	}
 	return ret
 }
 
-func v6addrs(aa []net.IPAddr) (ret []netip.Addr) {
+func v6addrs(aa []netip.Addr) (ret []netip.Addr) {
 	for _, a := range aa {
-		if ip, ok := netip.AddrFromSlice(a.IP); ok && ip.Is6() {
-			ret = append(ret, ip)
+		if a.Is6() {
```
bradfitz, line 625: `&& !a.Is4In6()` probably? otherwise the same IP can match both v4addrs and v6addrs

net/dnscache/dnscache.go
```diff
@@ -147,34 +149,30 @@ var debug = envknob.Bool("TS_DEBUG_DNS_CACHE")
...
-				v6 = ipa.IP
+			if !v6.IsValid() && naIP.Is6() {
+				v6 = naIP
 			}
-			allIPs = append(allIPs, *ipa)
+			allIPs = append(allIPs, naIP)
 		}
 		return
 	}
 	if ip := net.ParseIP(host); ip != nil {
```
bradfitz, line 168: use netip.ParseAddr instead?

net/dnscache/dnscache.go
skriptble: Good catch, I'll update this.

net/dnscache/dnscache.go
bradfitz: then you don't even need ip2addr?

skriptble: Oh, I don't think I even noticed that when I first wrote this code. Updated.

net/dnscache/dnscache.go
skriptble: Updated to convert the address to IPv4 here.

net/dnscache/dnscache.go
```diff
@@ -262,55 +270,46 @@ func (r *Resolver) lookupIP(host string) (ip, ip6 net.IP, allIPs []net.IPAddr, e
 
 	ctx, cancel := context.WithTimeout(context.Background(), r.lookupTimeoutForHost(host))
 	defer cancel()
-	ips, err := r.fwd().LookupIPAddr(ctx, host)
-	if err != nil || len(ips) == 0 {
+	ipsn, err := r.fwd().LookupIPAddr(ctx, host)
```
maisem, line 273: use LookupNetIP instead?

skriptble: Updated.

maisem (approved):

PR #5391, opened 2022-08-16
skriptble/dnscache v4filter fix

skriptble:


skriptble:
I'll merge this after #5390 is merged.

bradfitz:
Can you make the fix commit reference the commit which regressed it?

(when in doubt, error on the side of cross-linking everything in commit messages; PRs don't matter as much for archaeology, as they don't make it into the git history) 

And ideally we'd fix the regression first, and then do the type change after.

skriptble:
Sure, what's the standard way that we reference other commits?

And I'll change this PR to fix the bug before the other changes that I made and then collapse the update I made here into #5390.

bradfitz:
> Sure, what's the standard way that we reference other commits?

Just first 7-10 hex digits of the git hash is fine. No need for a full URLs. GitHub will link things.

bradfitz (approved):

PR #5404, opened 2022-08-19
util/deephash: expand fast-path capabilities

dsnet:
Add support for maps and interfaces to the fast path.
Add cycle-detection to the pointer handling logic.
This logic is generally copied from the slow path.

A future commit will delete the slow path once
the fast path never falls back to the slow path.

Performance:

	Hash-24                18.5µs ± 1%    14.9µs ± 2%  -19.52%  (p=0.000 n=10+10)
	HashPacketFilter-24    2.54µs ± 1%    2.60µs ± 1%   +2.19%  (p=0.000 n=10+10)
	HashMapAcyclic-24      31.6µs ± 1%    30.5µs ± 1%   -3.42%  (p=0.000 n=9+8)
	TailcfgNode-24         1.44µs ± 2%    1.43µs ± 1%     ~     (p=0.171 n=10+10)
	HashArray-24            324ns ± 1%     324ns ± 2%     ~     (p=0.425 n=9+9)

The additional cycle detection logic doesn't incur much slow down
since it only activates if a type is recursive, which does not apply
for any of the types that we care about.

There is a notable performance boost since we switch from the fath path
to the slow path less often. Most notably, a struct with a field that
could not be handled by the fast path would previously cause
the entire struct to go through the slow path.

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -14,7 +14,7 @@
 //   - time.Time are compared based on whether they are the same instant in time
 //     and also in the same zone offset. Monotonic measurements and zone names
 //     are ignored as part of the hash.
-//   - netip.Addr are compared based on a shallow comparison of the struct. 
+//   - netip.Addr are compared based on a shallow comparison of the struct.
 //
 // WARNING: This package, like most of the tailscale.com Go module,
 // should be considered Tailscale-internal; we make no API promises.
@@ -25,7 +25,6 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
-	"log"
 	"math"
 	"net/netip"
 	"reflect"
@@ -246,7 +245,7 @@ func (ti *typeInfo) hasher() typeHasherFunc {
 }
 
 func (ti *typeInfo) buildHashFuncOnce() {
-	ti.hashFuncLazy = genTypeHasher(ti.rtype)
+	ti.hashFuncLazy = genTypeHasher(ti)
 }
 
 func (h *hasher) hashBoolv(v addressableValue) bool {
@@ -380,13 +379,8 @@ func genHashPtrToMemoryRange(eleType reflect.Type) typeHasherFunc {
 	}
 }
 
-const debug = false
-
-func genTypeHasher(t reflect.Type) typeHasherFunc {
-	if debug {
-		log.Printf("generating func for %v", t)
-	}
-
+func genTypeHasher(ti *typeInfo) typeHasherFunc {
+	t := ti.rtype
 	switch t.Kind() {
 	case reflect.Bool:
 		return (*hasher).hashBoolv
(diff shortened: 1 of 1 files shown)


util/deephash/deephash.go
```diff
@@ -436,30 +430,65 @@ func genTypeHasher(t reflect.Type) typeHasherFunc {
 		default:
 			return genHashStructFields(t)
 		}
+	case reflect.Map:
+		return func(h *hasher, v addressableValue) bool {
+			// TODO: Check for nil.
```
bradfitz, line 435: who should do this? :)

seems worth doing now, though. What'll happen otherwise?

util/deephash/deephash.go
```diff
@@ -436,30 +430,65 @@ func genTypeHasher(t reflect.Type) typeHasherFunc {
...
+			}
+			va := newAddressableValue(v.Elem().Type())
+			va.Set(v.Elem())
 
-	return func(h *hasher, v addressableValue) bool {
-		if debug {
-			log.Printf("unhandled type %v", v.Type())
+			h.HashUint8(1) // indicates visiting interface value
+			h.hashType(va.Type())
+			h.hashValue(va, true)
```
bradfitz, line 487: we don't need recursive checks here?

dsnet: In pure Go, cycles can only occur through pointers, maps, and slices. We already do cycle detection for pointers and maps, and there's a TODO to add it for slices.

Technically, there is an implicit pointer inside a Go interface, but it is impossible to create a cycle without using unsafe. I don't think we need to handle this case.

util/deephash/deephash.go

util/deephash/deephash.go
bradfitz: I was skeptical that you couldn't make a cycle through interfaces so went to make a little test but it crashed before I could test anything:

```go
func TestInterfaceCycle(t *testing.T) {
	type T struct {
		v any
	}
	a := new(T)
	b := new(T)
	a.v = b
	b.v = a
	Hash(a)
}
```

```
--- FAIL: TestInterfaceCycle (0.00s)
panic: reflect: reflect.Value.Set using value obtained using unexported field [recovered]
	panic: reflect: reflect.Value.Set using value obtained using unexported field

goroutine 89 [running]:
testing.tRunner.func1.2({0x1012c0960, 0x1400019c490})
	/Users/bradfitz/.cache/tailscale-go/src/testing/testing.go:1396 +0x1c8
testing.tRunner.func1()
	/Users/bradfitz/.cache/tailscale-go/src/testing/testing.go:1399 +0x380
panic({0x1012c0960, 0x1400019c490})
	/Users/bradfitz/.cache/tailscale-go/src/runtime/panic.go:884 +0x204
reflect.flag.mustBeExportedSlow(0x14000065a68?)
	/Users/bradfitz/.cache/tailscale-go/src/reflect/value.go:238 +0x70
reflect.flag.mustBeExported(...)
	/Users/bradfitz/.cache/tailscale-go/src/reflect/value.go:229
reflect.Value.Set({0x1012b7780?, 0x140005c60e0?, 0x0?}, {0x1012b7780?, 0x1400019c410?, 0x101202524?})
	/Users/bradfitz/.cache/tailscale-go/src/reflect/value.go:2155 +0x84
tailscale.com/util/deephash.genTypeHasher.func3(0x14000188000, {{0x1012cb9c0?, 0x1400019c3f0?, 0x10000010131d920?}})
	/Users/bradfitz/src/tailscale.com/util/deephash/deephash.go:483 +0x158
tailscale.com/util/deephash.structHasher.hash({{0x140001b41b0?, 0x14000065bd8?, 0x14000065c38?}}, 0x14000188000, {{0x1012e5420?, 0x1400019c3f0?, 0x101212820?}})
	/Users/bradfitz/src/tailscale.com/util/deephash/deephash.go:360 +0x1a0
tailscale.com/util/deephash.(*hasher).hashValueWithType(0x14000188000, {{0x1012e5420?, 0x1400019c3f0?, 0x16?}}, 0x140001a3080, 0x14?)
	/Users/bradfitz/src/tailscale.com/util/deephash/deephash.go:678 +0x84
tailscale.com/util/deephash.(*hasher).hashValue(0x14000188000?, {{0x1012e5420?, 0x1400019c3f0?, 0x1010d9650?}}, 0xf0?)
	/Users/bradfitz/src/tailscale.com/util/deephash/deephash.go:672 +0x80
tailscale.com/util/deephash.Hash({0x1012b7780?, 0x1400019c3f0})
	/Users/bradfitz/src/tailscale.com/util/deephash/deephash.go:154 +0x2ec
tailscale.com/util/deephash.TestInterfaceCycle(0x140005029c0?)
	/Users/bradfitz/src/tailscale.com/util/deephash/deephash_test.go:920 +0xa4
testing.tRunner(0x140005829c0, 0x10133b968)
	/Users/bradfitz/.cache/tailscale-go/src/testing/testing.go:1446 +0x10c
created by testing.(*T).Run
	/Users/bradfitz/.cache/tailscale-go/src/testing/testing.go:1493 +0x300
FAIL	tailscale.com/util/deephash	0.084s
FAIL
```

dsnet: If you export the `v` field, then the test passes.

The reason we don't need to worry about cycles is because there's still an explicit pointer in the value. Traversing the 
1. `*T`
2. `T`
3. `any`
4. `*T` (saw this pointer before!)

In regards to the panic, it appears that we already have a pre-existing bug where we call `reflect.Value.Set` without first verifying that the provided value doesn't have the "read-only" bit set. I'll fix that in a separate PR since this isn't a regression and we'll need to fix it a few places.

dsnet: Hahaha... the places where we rely on `reflect.Value.SetIterXXX` are okay due to a bug in reflect. Great...

I filed https://github.com/golang/go/issues/54628

dsnet: It seems that #5367, which is where I'm trying to move `deephash` to, avoids this problem entirely since everything goes through an `unsafe.Pointer`, and thus dropping any "read-only" bits in the `reflect.Value` that could bite us.

dsnet: In case you curious, you can construct a cycle with pure interfaces through:
```go
type T struct { V any }
v := new(T)
v.V = T{}
type interfaceHeader struct{ Type, Data unsafe.Pointer }
(*interfaceHeader)(unsafe.Pointer(&a.V)).Data = unsafe.Pointer(a)
Hash(a)
```
I don't think it's worth checking for this. Any assignment of a non-pointer into a Go interface is going to force that value to be heap allocated, guaranteeing that a cycle can't happen without `unsafe`.

bradfitz (approved):

Issue #5396, opened 2022-08-18
FR: Improve connectivity by sharing intermediate router information to peers


awsong:
### What are you trying to do?

![image](https://user-images.githubusercontent.com/787544/185341584-0aa96525-21a5-47bf-9fcb-08e6806fd543.png)

As shown in above picture, node A and B are not able to get direct connection because they can only see their local addresses and public IP address via STUN server. They can't see each other's 192 addresses. When node C joins, it can get direct connections with both A and B. C will connect to B through 192.168.2.1. Now if C shares this information with B, B will add 192.168.2.1 to its address list and exchange the info with A. A then is able to reach B through 192.168.2.1 and get direct connection. 

I tried to setup a custom DERP on 192 network and it makes A and B connect directly. So I believe sharing this would help the overall connectivity. 

### How should we solve this?

Share address information between connecting nodes.

### What is the impact of not solving this?

Nodes reside in intranet with intermediate routes may stuck with relay.

DentonGentry:
Nodes do report every endpoint IP address they can find, the issue is that they have no visibility into IP addresses at that middle layer without the DERP server there.

Nodes report:
- all of their local interface IPs
- any public port they manage to open using a portmapping protocol like UPnP, NAT-PMP, or PCP
- any public address they can figure out by sending a STUN packet to the DERP servers, asking what public IP address it sees.

Without the DERP "C" you added, the nodes can figure out:
![no DERP C](https://user-images.githubusercontent.com/1140756/185440370-19fe1af9-581e-46fe-bcde-941afb6ce22e.png)

Adding a DERP at "C" also provided a new STUN server, allowing the nodes to figure out their IP address by asking the DERP:
![DERP C](https://user-images.githubusercontent.com/1140756/185440573-cb73c061-1a12-433b-9b7a-2d101a8d2052.png)

awsong:
Sorry, I meant that when node C is a normal tailnet node communicating with B, it should let B know the 192 address.

awsong:
I got PoC patch here which I tried and worked. What it does is sending the destination IP-Port pair in disco ping message so the receiving party can store it as its local address if it's not through DERP. 

```diff 
diff --git a/disco/disco.go b/disco/disco.go
index d8eb504d..ab9a6415 100644
--- a/disco/disco.go
+++ b/disco/disco.go
@@ -112,6 +112,8 @@ type Ping struct {
 	// TxID is a random client-generated per-ping transaction ID.
[...]

PR #5406, opened 2022-08-19
Improve NAT traversal by sharing endpoint info between peers.

awsong:
Signed-off-by: Andy Song <wsong.cn@gmail.com>

Related to issue #5396 

Sending dest IP-Port in disco Ping message, the recieving party
will then be able to figure out its "public facing ip/port" relative
to sending party. This effectively makes the sending party a STUN.
Pre-condition: the two parties already get direct connection.

### DIFF
--- a/types/logger/rusage_syscall.go
+++ b/types/logger/rusage_syscall.go
@@ -9,12 +9,13 @@ package logger
 
 import (
 	"runtime"
-	"syscall"
+
+	"golang.org/x/sys/unix"
 )
 
 func rusageMaxRSS() float64 {
-	var ru syscall.Rusage
-	err := syscall.Getrusage(syscall.RUSAGE_SELF, &ru)
+	var ru unix.Rusage
+	err := unix.Getrusage(unix.RUSAGE_SELF, &ru)
 	if err != nil {
 		return 0
 	}


disco/disco.go
```diff
@@ -118,6 +118,8 @@ type Ping struct {
 	// netmap data to reduce the discokey:nodekey relation from 1:N to
 	// 1:1.
 	NodeKey key.NodePublic
+
+	Dst netip.AddrPort
```
bradfitz, line 122: This needs to be documented.

DentonGentry:
We're about to start testing the 1.30 release. Even if the code review proceeds and is approved, I'll ask that we not merge it until after the release-branch/1.30 is created. This would go into the first unstable build in the 1.31 series.

disco/disco.go
awsong: > This needs to be documented.
        // Dst is the destination IP/Port of the ping message, it's receiving
        // node's public facing address relative to sending node. The receiving
        // node will add this to its local endpoint list upon receiving the
        // ping message.

Will this work? If the comment is OK I'll add to my PR.

awsong:
Sorry I accidentially discarded this commit when I tried to sync my repo with upstream. Should I open a new PR or can I do anything on top of this one?

PR #5413, opened 2022-08-22
doctor: add package for running in-depth healthchecks; use in bugreport

andrew-d:
This is kind of a monstrous change, and once I get off this plane and back home I'll post some additional sample output. For now:

<details>
  <summary>Sample output from my macOS laptop</summary>

```
route: {Kind: IPv4, Dst: default, GatewayAddr: 172.19.131.2, OutputIf: en0, Flags: [gateway global static up]}
route: {Kind: IPv4, Dst: default, GatewayIf: utun3, OutputIf: utun3, Flags: [global ifscope static up]}
route: {Kind: IPv4, Dst: 100.64.0.0/10, GatewayIf: utun3, OutputIf: utun3, Flags: [static up]}
route: {Kind: IPv4, Dst: 100.100.100.100/32, GatewayIf: utun3, OutputIf: utun3, Flags: [static up]}
route: {Kind: IPv4, Dst: 127.0.0.0/8, GatewayAddr: 127.0.0.1, OutputIf: lo0, Flags: [static up]}
route: {Kind: IPv4, Dst: 127.0.0.1, GatewayAddr: 127.0.0.1, OutputIf: lo0, Flags: [host up]}
route: {Kind: IPv4, Dst: 169.254.0.0/16, GatewayIf: en0, OutputIf: en0, Flags: [static up]}
route: {Kind: IPv4, Dst: 224.0.0.0/4, GatewayIf: en0, OutputIf: en0, Flags: [multicast static up]}
route: {Kind: IPv4, Dst: 224.0.0.0/4, GatewayIf: utun3, OutputIf: utun3, Flags: [ifscope multicast static up]}
route: {Kind: IPv4, Dst: 255.255.255.255/32, GatewayIf: en0, OutputIf: en0, Flags: [static up]}
route: {Kind: IPv4, Dst: 255.255.255.255/32, GatewayIf: utun3, OutputIf: utun3, Flags: [ifscope static up]}
route: {Kind: IPv6, Dst: default, GatewayAddr: fd7a:115c:a1e0::, OutputIf: utun3, Flags: [gateway global ifscope up]}
route: {Kind: IPv6, Dst: ::1, GatewayAddr: ::1, OutputIf: lo0, Flags: [host up]}
route: {Kind: IPv6, Dst: fd7a:115c:a1e0::/48, GatewayAddr: fe80::f22f:4bff:fe01:151b%utun3, OutputIf: utun3, Flags: [up]}
route: {Kind: IPv6, Dst: fd7a:115c:a1e0:ab12:4843:cd96:624b:e86c, GatewayIf: utun3, OutputIf: lo0, Flags: [host up]}
```

(edited to remove a bunch of similar routes that don't really help demonstrate anything)
</details>


DentonGentry:
As 1.30 testing has started, and this is large, and `tailscale bugreport` is essential, I'd ask that we do the code review but not submit this until release-branch/1.30 has branched (which might be next week).

JayWStapleton:
I feel this should be explicit opt-in
Like: `tailscale bugreport --extended` which returns a list of everything we collect. With a `proceed?` prompt

This information will be in our logs for (ever/a long time) and people should be aware of what they're sending us

andrew-d:
> I feel this should be explicit opt-in Like: `tailscale bugreport --extended` which returns a list of everything we collect. With a `proceed?` prompt
> 
> This information will be in our logs for (ever/a long time) and people should be aware of what they're sending us

Oh, good call and thanks for the reminder; was meaning to do that and forgot. Added a `--doctor` flag to the `tailscale bugreport` CLI and local API that defaults to `false` if not specified.

doctor/routetable/routetable_bsd.go
```diff
@@ -0,0 +1,256 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build darwin || freebsd
```
DentonGentry, line 5: probably openbsd as well (we have a small but mighty contingent of OpenBSD users)

DentonGentry (approved):
Having made it conditional on `--doctor`, I think this would be fine to go in to 1.30. Even if there is a cohort of weird systems out there which make the doctor code die, we'd still be able to get regular `tailscale bugreport` strings out of them.

bradfitz:
(sorry, on phone. More later. Internally I'm fine with doctor but I'd rather not expose that name in the CLI or UI. Bug doctor? Weird vet.)

andrew-d:
> I'm not a huge fan of the name doctor.

Thoughts on Jay's `--extended` flag? Could call it "extended diagnostics", and rename the `tailscale debug` subcommand to `tailscale debug extended-diagnostics`?

doctor/routetable/routetable_linux.go
```diff
@@ -0,0 +1,167 @@
...
+	routes, err := netlink.RouteListFiltered(netlink.FAMILY_ALL, filter, netlink.RT_FILTER_TABLE)
+	if err != nil {
+		return nil, err
+	}
+
+	var ret []routeEntry
+	for _, route := range routes {
+		if route.Family != netlink.FAMILY_V4 && route.Family != netlink.FAMILY_V6 {
+			continue
+		}
```
bradfitz, line 136: Make getRouteTable take a `max int` parameter?

I run Tailscale on an internet router that has a full internet routing table with a million routes.

types/logger/logger.go
```diff
@@ -141,6 +141,8 @@ var rateFree = []string{
 	"SetPrefs: %v",
 	"peer keys: %s",
 	"v%v peers: %v",
+	// debug messages printed by 'tailscale bugreport'
+	"doctor: ",
```
bradfitz, line 145: diag: ?

cmd/tailscale/cli/bugreport.go
```diff
@@ -16,6 +17,15 @@ var bugReportCmd = &ffcli.Command{
 	Exec:       runBugReport,
 	ShortHelp:  "Print a shareable identifier to help diagnose issues",
 	ShortUsage: "bugreport [note]",
+	FlagSet: (func() *flag.FlagSet {
+		fs := newFlagSet("doctor")
+		fs.BoolVar(&bugReportArgs.doctor, "doctor", false, "run additional in-depth checks")
```
bradfitz, line 22: diagnose? diag? Or make it a comma-separated list of check names with an "all" alias? `--check=all` ?

doctor/routetable/routetable_bsd.go
```diff
@@ -0,0 +1,256 @@
...
+		populateGateway(&re, ifsByIdx, rm.Addrs[unix.RTAX_GATEWAY])
+	}
+
+	if outif, ok := ifsByIdx[rm.Index]; ok {
+		re.OutputIf = outif.Name
+	}
+	return re, true
+}
+
+func getRouteTable() ([]routeEntry, error) {
```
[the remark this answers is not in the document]
andrew-d: Yeah, I was struggling to find a good cross-platform abstraction for "give me a route table" since the concept of a route table isn't particularly cross-platform.

andrew-d:
> Really this should just be on by default, maybe even without a flag.

I think @JayWStapleton's point about this gathering a bunch of possibly-sensitive data is a good one; I'm happy to turn it on in the future by default, but I don't want to go right to "on by default for everyone" without at least a bit of transition, thinking more about redacting the data, and digging into our log deletion a bit more.

As for `tailscale debug`, I was thinking of something like 9125a8061644dfdb933dea698642b568a94b95a6 - you can run `tailscale debug diagnostics` to print the debug messages that would be logged, but without actually logging. Intentionally not a stable CLI API, but at least it's visible?

bradfitz:
`tailscale debug diagnostics` is a little redundant. Maybe `tailscale debug dump` or `print`? Also shorter.

andrew-d:
> `tailscale debug diagnostics` is a little redundant. Maybe `tailscale debug dump` or `print`? Also shorter.

Sure, makes sense - was going to add that in another PR, so will rename to `debug dump` there 👍

doctor/routetable/routetable_bsd.go
andrew-d: I don't have an OpenBSD system to test on, but I will set up a VM and test this package as a follow-up PR 👍

doctor/routetable/routetable_bsd.go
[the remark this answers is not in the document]
bradfitz: It's mostly cross platform, though. There are some per-OS special fields but we could shove them into a 'Sys any` field/method like https://pkg.go.dev/io/fs#FileInfo and its Sys.

andrew-d:
@bradfitz - I took a pass at making this cross-platform; we now have the `routeEntry` type, which is the same on every platform, and a `Sys` field that contains per-platform information about each route. Thoughts?

doctor/routetable/routetable_linux.go
```diff
@@ -0,0 +1,230 @@
...
+	// Table is the routing table index of this route.
+	Table int
+	// Src is the source of the route (if any)
+	Src netip.Addr
+	// Proto describes the source of the route--i.e. what caused this route
+	// to be added to the route table.
+	Proto netlink.RouteProtocol
+	// Priority is the route's priority.
+	Priority int
+	// Scope is the route's scope
```
bradfitz, line 37: missing period here and Src above

client/tailscale/localclient.go
```diff
@@ -269,8 +269,9 @@ func (lc *LocalClient) Profile(ctx context.Context, pprofType string, sec int) (
 }
 
 // BugReport logs and returns a log marker that can be shared by the user with support.
-func (lc *LocalClient) BugReport(ctx context.Context, note string) (string, error) {
-	body, err := lc.send(ctx, "POST", "/localapi/v0/bugreport?note="+url.QueryEscape(note), 200, nil)
+func (lc *LocalClient) BugReport(ctx context.Context, note string, diagnose bool) (string, error) {
```
bradfitz, line 272: per chat: keep this method as-is and add a new BugReportOpt or something that takes a struct of things

and document diagnose bool

doctor/doctor.go
```diff
@@ -0,0 +1,76 @@
...
+	"tailscale.com/types/logger"
+)
+
+// Check is the interface defining a singular check.
+//
+// A check should log information that it gathers using the provided log
+// function, and should attempt to make as much progress as possible in error
+// conditions.
+type Check interface {
+	Name() string
```
bradfitz, line 22: document some example strings formats this might return? Is this "My Good Check" or "good-check" ?

net/routetable/routetable.go
```diff
@@ -0,0 +1,148 @@
...
+		ip = ip.WithZone(r.Zone)
+	}
+	return ip.String() + "/" + strconv.Itoa(r.Prefix.Bits())
+}
+
+// RouteType describes the type of a route.
+type RouteType int
+
+const (
+	// An unspecified route type
```
bradfitz, line 113: before-line comments should be complete sentences starting with the symbol as the subject. 

maybe just make these same-line comments instead? then they can be sentence fragments and lack capitalization/punctuation.

either wya.

net/routetable/routetable_default.go
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
```
bradfitz, line 1: routetable_other.go ? or _stub?

net/routetable/routetable_linux.go
```diff
@@ -0,0 +1,231 @@
...
+		return "link"
+	case unix.RT_SCOPE_SITE:
+		return "site"
+	default:
+		return strconv.Itoa(r.Scope)
+	}
+}
+
+// Get returns route entries from the system route table, limited to at most
+// 'max' results.
```
bradfitz, line 142: drop single quotes around max. Go generally avoids that.

doctor/routetable/routetable.go
```diff
@@ -0,0 +1,35 @@
...
+const MaxRoutes = 1000
+
+// Check implements the doctor.Check interface.
+type Check struct{}
+
+func (c Check) Name() string {
+	return "routetable"
+}
+
+func (c Check) Run(_ context.Context, log logger.Logf) error {
```
bradfitz, line 26: name this param `logf` so at call site (for both humans + automated tools) it's obviously fmt-ish call

doctor/routetable/routetable.go
```diff
@@ -0,0 +1,35 @@
...
+	"tailscale.com/types/logger"
+)
+
+// MaxRoutes is the maximum number of routes that will be displayed.
+const MaxRoutes = 1000
+
+// Check implements the doctor.Check interface.
+type Check struct{}
+
+func (c Check) Name() string {
```
bradfitz, line 22: can drop receiver name here and line 26

doctor/doctor_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+type testCheck1 struct{}
+
+func (t testCheck1) Name() string { return "testcheck1" }
+func (t testCheck1) Run(_ context.Context, log logger.Logf) error {
+	log("check 1")
+	return nil
+}
+
+func assertContains[T comparable](t *testing.T, arr []T, val T) {
```
bradfitz, line 50: use https://pkg.go.dev/github.com/frankban/quicktest ?

bradfitz (approved):

PR #5414, opened 2022-08-22
Fastpath for processing AUMs in the common case, tx NodeID during tka setup

twitchyliquid64:
- Transmit the nodeID so the server knows who we are. This is checked against the machine key used to authenticate the session, so attempts to spoof this value will result in an error.
 - Implements a fast path for processing AUMs which just chain off of HEAD. I expect this to be the common case.

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -1829,25 +1829,31 @@ type PeerChange struct {
 // TKAInitBeginRequest submits a genesis AUM to seed the creation of the
 // tailnet's key authority.
 type TKAInitBeginRequest struct {
-	NodeID NodeID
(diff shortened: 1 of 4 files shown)


tailcfg/tailcfg.go
```diff
@@ -1845,6 +1845,8 @@ type TKAInitBeginResponse struct {
 // TKAInitFinishRequest finalizes initialization of the tailnet key authority
 // by submitting node-key signatures for all existing nodes.
 type TKAInitFinishRequest struct {
+	NodeID NodeID
```
crawshaw, line 1848: document which node this is the ID for

crawshaw (approved):

PR #5420, opened 2022-08-23
tka: support rotating node-keys in node-key signatures

twitchyliquid64:


tka/sig.go
```diff
@@ -42,17 +48,47 @@ func (s SigKind) String() string {
...
 	// SigCredential signature kinds.
 	KeyID []byte `cbor:"3,keyasint,omitempty"`
 
-	// Signature is the packed (R, S) ed25519 signature over the rest
-	// of the structure.
+	// Signature is the packed (R, S) ed25519 signature over all other
+	// fields of the structure.
 	Signature []byte `cbor:"4,keyasint,omitempty"`
+
+	// Nested describes a NodeKeySignature which authorizes the node-key
```
sailorfrag, line 63: I think this might be worth some elaboration.

I believe from the rest of the code that `Nested` contains the original `NodeKeySignature` that approved this node (an approved verification key) _or_ the previous rotation's `NodeKeySignature` (which will in turn nest more rotations, eventually terminating in the original approval). Is that correct?

I wonder if it would be better to make it an array since the nesting might get pretty deep and it'll be more efficient to allocate the entire array upfront.

It might also be good to have a plan for compaction even if we don't implement that right away.

tka/sig.go
```diff
@@ -97,18 +133,56 @@ func (s *NodeKeySignature) Unserialize(data []byte) error {
...
+		if !ed25519.Verify(ed25519.PublicKey(verifyPub[:]), sigHash[:], s.Signature) {
+			return errors.New("invalid signature")
+		}
+
+		// Recurse to verify the signature on the nested structure.
+		var nestedPub key.NodePublic
+		if err := nestedPub.UnmarshalBinary(s.Nested.Pubkey); err != nil {
+			return fmt.Errorf("nested pubkey: %v", err)
+		}
+		if err := s.Nested.verifySignature(nestedPub, verificationKey); err != nil {
```
sailorfrag, line 168: How well does Go handle deep recursive function calls? For tailnets that require daily key rotation, I could see this easily being hundreds or even thousands of levels deep.

tka/sig.go
```diff
@@ -42,17 +48,47 @@ func (s SigKind) String() string {
...
+	if len(s.RotationPubkey) > 0 {
+		return ed25519.PublicKey(s.RotationPubkey), true
+	}
+
+	switch s.SigKind {
+	case SigRotation:
+		if s.Nested == nil {
+			return nil, false
+		}
+		return s.Nested.rotationPublic()
```
sailorfrag, line 87: will this be optimized into tail recursion? Otherwise it might be better to make this a loop.

tka/sig_test.go
```diff
@@ -30,11 +32,54 @@ func TestSigDirect(t *testing.T) {
 		t.Errorf("sigHash changed after signing: %x != %x", sig.SigHash(), sigHash)
 	}
 
-	if err := sig.verifySignature(key); err != nil {
+	if err := sig.verifySignature(node.Public(), k); err != nil {
 		t.Fatalf("verifySignature() failed: %v", err)
 	}
 }
 
+func TestSigNested(t *testing.T) {
```
sailorfrag, line 40: Suggestion for additional tests:
1. Deeply nested signature (100+)
2. First signature fails
3. Last signature fails
4. An intermediate signature fails
5. The last signature is valid but not actually on the approver list

tka/tka.go
```diff
@@ -23,7 +24,7 @@ var cborDecOpts = cbor.DecOptions{
 	TagsMd:      cbor.TagsForbidden,
 
 	// Arbitrarily-chosen maximums.
-	MaxNestedLevels:  8,
+	MaxNestedLevels:  16, // Most likely to be hit for SigRotation sigs.
```
sailorfrag, line 27: Is this going to be enough?

tka/sig.go
twitchyliquid64: For compaction: My hope was that control would (eventually) implement telling a node with a network-lock key to direct-sign a signature, providing the non-direct signature as evidence that the node-key is authorized.

I'm hoping to avoid nesting more than 2 or three deep. The rotation key shouldnt change (or shouldnt change frequently), so when the rotation happens, the node can drop the outer signature, and make a new one using its rotation key. IE: it stays 2 deep even when it rotates.

> I believe from the rest of the code that Nested contains the original NodeKeySignature that approved this node (an approved verification key) or the previous rotation's NodeKeySignature (which will in turn nest more rotations, eventually terminating in the original approval). Is that correct?

Yes, with the caveat that I don't want the nesting to grow unbounded as per above.

tka/sig.go
twitchyliquid64: I think it has no issue with tens of thousands but it does cap it eventually. In our case, we'll hit our CBOR nesting limit well before that.

tka/sig.go
twitchyliquid64: I don't think so, but I'm also hoping to keep nesting a few levels deep so hopeful optimizations like an array or a loop wont be necessary.

tka/tka.go
twitchyliquid64: Thinking on it some more, I think we could get away with a hard limit of 3 (direct -> rotation, or credential -> rotation -> rotation, I havent implemented credential yet).

tka/sig.go
sailorfrag: Okay, I guess I was expecting that the rotation key would also rotate at each rotation. Maybe add a comment about that too!

sailorfrag (approved):

tka/sig_test.go
twitchyliquid64: Done.

tka/sig.go
twitchyliquid64: Done.

Issue #5415, opened 2022-08-23
cmd/tsconnect: generate NPM package


mihaip:
We wish to embed the functionality into other contexts/a large app, thus making it reusable is necessary. We should be able to run esbuild with a different entyrpoint to generate the sources for an npm package. esbuild does not support generating type declarations (see evanw/esbuild#95), thus we may need to manually invoke `tsc` for that (see https://github.com/VulcanJS/npm-the-right-way for an example of how to set that up).

This should include the wrapper around loading the .wasm and creating an SSH session (and rendering it in xterm.js), but should not include login (including QR code generation) or any Preact UI.

- [x] Refactor code to allow the non-React parts to be used independently
- [x] `esbuild`-based build script for package
- [x] Move xterm.js-based SSH session to the package
- [x] GitHub Action for publishing new package versions (auto-generate version number using `version/version.go`)

PR #5421, opened 2022-08-24
cmd/tsconnect: extract NPM package for reusing in other projects

mihaip:
`src/` is broken up into several subdirectories:
- `lib/` and `types`/ for shared code and type definitions (more code
  will be moved here)
- `app/` for the existing Preact-app
- `pkg/` for the new NPM package

A new `build-pkg` esbuild-based command is added to generate the files
for the NPM package. To generate type definitions (something that esbuild
does not do), we set up `dts-bundle-generator`.

Includes additional cleanups to the Wasm type definitions (we switch to
string literals for enums, since exported const enums are hard to use
via packages).

Also allows the control URL to be set a runtime (in addition to the
current build option), so that we don't have to rebuild the package
for dev vs. prod use.

Updates #5415


cmd/tsconnect/pkg/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "author": "Tailscale Inc.",
+  "description": "Tailscale Connect SDK",
+  "license": "BSD-3-Clause",
+  "name": "@tailscale/connect",
+  "type": "module",
```
rosszurowski, line 6: This means we're only exporting ES Module versions here. Along with the fact that we're using `fetch` internally, that means this package won't be compatible with Node or with older browsers. I assume that's ok?

cmd/tsconnect/package.json
```diff
@@ -13,7 +14,8 @@
     "xterm-addon-fit": "^0.5.0"
   },
   "scripts": {
-    "lint": "tsc --noEmit"
+    "lint": "tsc --noEmit",
+    "pkg-types": "dts-bundle-generator --inline-declare-global=true --no-banner -o pkg/pkg.d.ts src/pkg/pkg.ts"
```
rosszurowski, line 18: Eventually we'll likely want to split some of the IPN types into its own package. For example, [in the Docker Extension](https://github.com/tailscale/docker-extension/blob/main/ui/src/tailscale.ts#L8) I needed to manually re-implement all the IPN types too.

Nothing to do now, just flagging FYSA.

cmd/tsconnect/README.md
```diff
@@ -28,3 +28,13 @@ To serve them, run:
 ```
 
 By default the build output is placed in the `dist/` directory and embedded in the binary, but this can be controlled by the `-distdir` flag. The `-addr` flag controls the interface and port that the serve listens on.
```
rosszurowski, line 30: Does this change deprecate the use of the `dist` folder? Or is that still helpful for local testing?

rosszurowski (approved):
A few small comments, but generally looks good to me! I assume any CI publishing steps are going to be added in later PRs.

cmd/tsconnect/pkg/package.json
mihaip: Yeah, it's meant for use in the admin panel only.

cmd/tsconnect/package.json
mihaip: Huh, hadn't realized that the Docker extension has a TypeScript component, that's a good observation.

cmd/tsconnect/README.md
mihaip: `dist/` is still used when building the app (what's served at https://tsconnect.fly.dev/). That still seems useful as a testbed (that doesn't depend on control), so I'm keeping it around for now.

mihaip:
>  I assume any CI publishing steps are going to be added in later PRs.

Yep, this was getting to be quite large as it was. #5415 has all the subtasks.

PR #5426, opened 2022-08-24
tailcfg: implement wire format for map request/response tka comms

twitchyliquid64:
Beginning to spitball what some of the regular TKA comms should look like. Thoughts?


twitchyliquid64:
Also /cc @crawshaw

sailorfrag (approved):

crawshaw (approved):
Looks like a good start! The format is adjustable until we ship it.

PR #5428, opened 2022-08-25
cmd/tsconnect: add NPM publish workflow

mihaip:
Updates #5415


mihaip:
The action would normally not run on PRs, but I temporarily enabled it for my branch and it appears to work: https://github.com/tailscale/tailscale/runs/8023317435?check_suite_focus=true

.github/workflows/tsconnect-pkg-publish.yml
```diff
@@ -0,0 +1,30 @@
...
+        # Build with build_dist.sh to ensure that version information is embedded.
+        # GOROOT is specified so that the Go/Wasm that is trigged by build-pk
+        # also picks up our custom Go toolchain.
+        run: |
+          ./build_dist.sh tailscale.com/cmd/tsconnect
+          GOROOT="${HOME}/.cache/tailscale-go" ./tsconnect build-pkg
+
+      - name: Publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.TSCONNECT_NPM_PUBLISH_AUTH_TOKEN }}
```
rosszurowski, line 29: How did this token get generated? It's just private to your account right now? I wonder if we should make an org, but maybe that can come later, since I think we'd need to pay.

cmd/tsconnect/build-pkg.go
```diff
@@ -41,4 +47,33 @@ func runBuildPkg() {
...
 
+	if err := updateVersion(); err != nil {
+		log.Fatalf("Cannot update version: %v", err)
+	}
+
+	log.Printf("Built package version %s", version.Long)
+}
+
+func updateVersion() error {
+	packageJSONBytes, err := os.ReadFile("pkg-package.json")
```
rosszurowski, line 58: Should we call this `package.json.tmpl` or something to imply it's not the final file? `pkg-package` feels like "package package" which doesn't tell me much.

cmd/tsconnect/build-pkg.go
```diff
@@ -41,4 +47,33 @@ func runBuildPkg() {
...
+
+	var packageJSON map[string]any
+	packageJSONBytes, err = hujson.Standardize(packageJSONBytes)
+	if err != nil {
+		return fmt.Errorf("Could not standardize template package.json: %w", err)
+	}
+	if err := json.Unmarshal(packageJSONBytes, &packageJSON); err != nil {
+		return fmt.Errorf("Could not unmarshal package.json: %w", err)
+	}
+	packageJSON["version"] = version.Long
```
rosszurowski, line 71: Looks like the version info that gets embedded looks like this `1.29.0-tf23741275`. I'm curious how package managers like npm handle hashes at the end like that… is there a reason we don't want to truncate down to a semver string?

rosszurowski (approved):

.github/workflows/tsconnect-pkg-publish.yml
mihaip: It's tied to my account. Will look into moving it to an org token once we get access to it (see [Slack thread](https://tailscale.slack.com/archives/CQQ1AUCCE/p1661370944918929))

cmd/tsconnect/build-pkg.go
mihaip: Yeah, `package.json.tmpl` is better.

cmd/tsconnect/build-pkg.go
mihaip: They're OK with it -- it gets treated as a [prerelease version identifier in semver](https://semver.org/#spec-item-9). This is generally good, because we don't bump the patch version in OSS, and without the Git hash it we would always end up with `1.29.0` and thus no way to distinguish.

PR #5429, opened 2022-08-25
util/cstruct: add package for decoding padded C structures

andrew-d:
I was working on my "dump iptables rules using only syscalls" branch and
had a bunch of C structure decoding to do. Rather than manually
calculating the padding or using unsafe trickery to actually cast
variable-length structures to Go types, I'd rather use a helper package
that deals with padding for me.

Padding rules were taken from the following article:
  http://www.catb.org/esr/structure-packing/

### DIFF
--- a/util/cstruct/cstruct.go
+++ b/util/cstruct/cstruct.go
@@ -1,0 +1,179 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package cstruct provides a helper for decoding binary data that is in the
+// form of a padded C structure.
+package cstruct
+
+import (
+	"errors"
+	"io"
+
+	"tailscale.com/util/endian"
+)
+
+// Size of a pointer-typed value, in bits
+const pointerSize = 32 << (^uintptr(0) >> 63)
+
+// We assume that non-64-bit platforms are 32-bit; we don't expect Go to run on
+// a 16- or 8-bit architecture any time soon.
+const is64Bit = pointerSize == 64
+
+// Decoder reads and decodes padded fields from a slice of bytes. All fields
+// are decoded with native endianness.
+//
+// Methods of a Decoder do not return errors, but rather store any error within
+// the Decoder. The first error can be obtained via the Err method; after the
+// first error, methods will return the zero value for their type.
+type Decoder struct {
+	b    []byte
+	off  int
+	err  error
+	dbuf [8]byte // for decoding
+}
+
+// NewDecoder creates a Decoder from a byte slice.
+func NewDecoder(b []byte) *Decoder {
+	return &Decoder{b: b}
+}
+
+var errUnsupportedSize = errors.New("unsupported size")
+
+func padBytes(offset, size int) int {
+	if offset == 0 || size == 1 {
+		return 0
+	}
+	remainder := offset % size
+	return size - remainder
+}
+
+func (d *Decoder) getField(b []byte) error {
+	size := len(b)
+
+	// We only support fields that are multiples of 2 (or 1-sized)
+	if size != 1 && size&1 == 1 {
+		return errUnsupportedSize
+	}
+
+	// Fields are aligned to their size
+	padBytes := padBytes(d.off, size)
+	if d.off+size+padBytes > len(d.b) {
+		return io.EOF
+	}
+	d.off += padBytes
+
+	copy(b, d.b[d.off:d.off+size])
+	d.off += size
+	return nil
+}
+
+// Err returns the first error that was encountered by this Decoder.
+func (d *Decoder) Err() error {
+	return d.err
+}
+
+// Offset returns the current read offset for data in the buffer.
+func (d *Decoder) Offset() int {
+	return d.off
+}
+
+// Byte returns a single byte from the buffer.
+func (d *Decoder) Byte() byte {
+	if d.err != nil {
+		return 0
+	}
+
+	if err := d.getField(d.dbuf[0:1]); err != nil {
+		d.err = err
+		return 0
+	}
+	return d.dbuf[0]
+}
+
+// Byte returns a number of bytes from the buffer based on the size of the
+// input slice. No padding is applied.
+//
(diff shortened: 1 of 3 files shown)


util/cstruct/cstruct.go
```diff
@@ -0,0 +1,208 @@
...
+func (d *Decoder) Err() error {
+	return d.err
+}
+
+// Offset returns the current read offset for data in the buffer.
+func (d *Decoder) Offset() int {
+	return d.off
+}
+
+// Byte reutrns a single byte from the buffer.
```
bradfitz, line 73: typo returns

util/cstruct/cstruct.go
```diff
@@ -0,0 +1,208 @@
...
+func (d *Decoder) Byte() byte {
+	if d.err != nil {
+		return 0
+	}
+
+	var b [1]byte
```
bradfitz, line 79: move this/some buf to Decoder? maybe it provably doesn't escape + allocate, but putting it on Decoder means not thinking about it.

util/cstruct/cstruct.go
```diff
@@ -0,0 +1,208 @@
...
+	d.off += size
+}
+
+// Uint16 returns a uint16 decoded from the buffer.
+func (d *Decoder) Uint16() uint16 {
+	if d.err != nil {
+		return 0
+	}
+
+	var buf [2]byte
```
bradfitz, line 113: yeah, could be a small buffer shared by all these

util/cstruct/cstruct.go
```diff
@@ -0,0 +1,208 @@
...
+
+	var b [1]byte
+	if err := d.getField(b[:]); err != nil {
+		d.err = err
+		return 0
+	}
+	return b[0]
+}
+
+// Byte reutrns a number of bytes from the buffer based on the size of the
```
bradfitz, line 87: typo

util/cstruct/cstruct.go
```diff
@@ -0,0 +1,208 @@
...
+
+	var buf [2]byte
+	if err := d.getField(buf[:]); err != nil {
+		d.err = err
+		return 0
+	}
+	return endian.Native.Uint16(buf[:])
+}
+
+// Uint32 returns a uint32 decoded from the buffer.
```
bradfitz, line 121: none of the godoc mentions endianness is native? add something somewhere?

util/cstruct/cstruct.go
```diff
@@ -0,0 +1,208 @@
...
+	var buf [8]byte
+	if err := d.getField(buf[:]); err != nil {
+		d.err = err
+		return 0
+	}
+	return int64(endian.Native.Uint64(buf[:]))
+}
+
+// TODO(andrew): remove unsafe with build tags?
+var pointerSize = int(unsafe.Sizeof(uintptr(0)))
```
bradfitz, line 208: you need neither unsafe nor build tags... e.g. https://github.com/tailscale/tailscale/blob/main/types/pad32/pad32.go#L8-L9

bradfitz (approved):

util/cstruct/cstruct.go
andrew-d: Added a comment on the type.

util/cstruct/cstruct.go
andrew-d: This is a neat trick; did something similar to make a `const pointerSize`

bradfitz: You're in good company:

```
bradfitz@tsdev:~/hack/go$ git grep -F ">> 63" | grep uint
misc/cgo/errors/testdata/malloc.go:		size = C.size_t(0x7fffffff << (32 * (^uintptr(0) >> 63)))
src/cmd/compile/internal/ssa/poset.go:const uintSize = 32 << (^uint(0) >> 63) // 32 or 64
src/cmd/vendor/golang.org/x/tools/go/analysis/passes/shift/shift.go:		// like ^uint(0) >> 63 for 32/64 bit detection and compatibility.
src/crypto/subtle/constant_time.go:	return int((uint64(uint32(x^y)) - 1) >> 63)
src/fmt/scan.go:	intBits     = 32 << (^uint(0) >> 63)
src/fmt/scan.go:	uintptrBits = 32 << (^uintptr(0) >> 63)
src/image/png/paeth.go:const intSize = 32 << (^uint(0) >> 63)
src/internal/goarch/goarch.go:const PtrSize = 4 << (^uintptr(0) >> 63)
src/math/bits/bits.go:const uintSize = 32 << (^uint(0) >> 63) // 32 or 64
src/math/const.go:	intSize = 32 << (^uint(0) >> 63) // 32 or 64
src/math/fma.go:	sign = uint32(b >> 63)
src/runtime/internal/math/math_test.go:	UintptrSize = 32 << (^uintptr(0) >> 63)
src/runtime/malloc.go:	_64bit = 1 << (^uintptr(0) >> 63) / 2
src/runtime/string_test.go:const intSize = 32 << (^uint(0) >> 63)
src/strconv/atoi.go:const intSize = 32 << (^uint(0) >> 63)
src/strconv/decimal.go:const uintSize = 32 << (^uint(0) >> 63)
src/vendor/golang.org/x/sys/cpu/hwcap_linux.go:	uintSize = int(32 << (^uint(0) >> 63))
```

util/cstruct/cstruct.go
```diff
@@ -0,0 +1,178 @@
...
+		return 0
+	}
+
+	switch pointerSize {
+	case 32:
+		return uintptr(d.Uint32())
+	case 64:
+		return uintptr(d.Uint64())
+	default:
+		panic("unsupported")
```
bradfitz, line 161: I'd actually put this in an `func init()` check. It's cheap enough to do at init-time.

Or really, you just need a bool const like `const is64Bit = (32 << (^uintptr(0) >> 63)) == 64`. Then the other branch gets `deadcode`d away by the toolchain.

bradfitz: Or both constants + init.

bradfitz (approved):
prior comments optional

util/cstruct/cstruct.go
bradfitz: Actually not sure what I was thinking about yesterday. A switch on a constant value will deadcode away the other branches anyway.

PR #5433, opened 2022-08-25
licenses: add GitHub Action to update license notices

willnorris:
This will update a licenses/tailscale.md file with all of our go dependencies and their respective licenses. Notices for other clients will be triggered by similar actions in other repos.

Updates tailscale/corp#5780

Example output that this will result in: https://github.com/willnorris/tailscale/blob/main/licenses/tailscale.md

### DIFF
--- a/.github/licenses.tmpl
+++ b/.github/licenses.tmpl
@@ -1,6 +1,6 @@
 # Tailscale CLI and daemon dependencies
 
-The follow open source dependencies are used to build the [tailscale][] and
+The following open source dependencies are used to build the [tailscale][] and
 [tailscaled][] commands. These are primarily used on Linux and BSD variants as
 well as an [option for macOS][].
 
--- a/.github/licenses.tmpl
+++ b/.github/licenses.tmpl
@@ -1,0 +1,17 @@
+# Tailscale CLI and daemon dependencies
+
+The follow open source dependencies are used to build the [tailscale][] and
+[tailscaled][] commands. These are primarily used on Linux and BSD variants as
+well as an [option for macOS][].
+
+[tailscale]: https://pkg.go.dev/tailscale.com/cmd/tailscale
+[tailscaled]: https://pkg.go.dev/tailscale.com/cmd/tailscaled
+[option for macOS]: https://tailscale.com/kb/1065/macos-variants/
+
+## Go Packages
+
+Some packages may only be included on certain architectures or operating systems.
+
+{{ range . }}
+ - [{{.Name}}](https://pkg.go.dev/{{.Name}}) ([{{.LicenseName}}]({{.LicenseURL}}))
+{{- end }}
--- a/.github/workflows/go-licenses.yml
+++ b/.github/workflows/go-licenses.yml
@@ -1,0 +1,62 @@
+name: go-licenses
+
+on:
+  # run action when a change lands in the main branch which updates go.mod or
+  # our license template file. Also allow manual triggering.
+  push:
+    branches:
+      - main
+    paths:
+      - go.mod
+      - .github/licenses.tmpl
+      - .github/workflows/go-licenses.yml
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-$${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  tailscale:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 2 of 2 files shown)


.github/licenses.tmpl
```diff
@@ -0,0 +1,17 @@
+# Tailscale CLI and daemon dependencies
+
+The follow open source dependencies are used to build the [tailscale][] and
```
andrew-d, line 3:
```suggestion
The following open source dependencies are used to build the [tailscale][] and
```

.github/workflows/go-licenses.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Run go-licenses
+        env:
+          # include all build tags to include platform-specific dependencies
+          GOFLAGS: "-tags=cgo,darwin,freebsd,ios,js,linux,openbsd,wasm,windows"
+        run: |
+          mkdir -p licenses
+          go-licenses report tailscale.com/cmd/tailscale tailscale.com/cmd/tailscaled > licenses/tailscale.md --template .github/licenses.tmpl
+
+      - name: Get access token
+        uses: tibdex/github-app-token@f717b5ecd4534d3c4df4ce9b5c1c2214f0f7cd06 # v1.6.0
```
danderson, line 45: You reviewed the contents of this action's repo and the one below at this revision? Admittedly passing our access tokens to third-party code makes me nervous, given that this one has the power to send apparently-genuine PRs.

danderson (approved):

.github/workflows/go-licenses.yml
```diff
@@ -0,0 +1,62 @@
...
+          go-version-file: go.mod
+
+      - name: Install go-licenses
+        run: |
+          go install github.com/google/go-licenses@v1.2.2-0.20220825154955-5eedde1c6584
+
+      - name: Run go-licenses
+        env:
+          # include all build tags to include platform-specific dependencies
+          GOFLAGS: "-tags=cgo,darwin,freebsd,ios,js,linux,openbsd,wasm,windows"
```
DentonGentry, line 39: Do you have thoughts on how to accommodate Android?

Windows pulls in some additional dependencies for its UI, like lxn, will we handle those from here later or some other way?

.github/workflows/go-licenses.yml
```diff
@@ -0,0 +1,62 @@
...
+        with:
+          token: ${{ steps.generate-token.outputs.token }}
+          author: License Updater <noreply@tailscale.com>
+          branch: licenses/cli
+          branch-suffix: short-commit-hash
+          commit-message: "licenses: update tailscale{,d} licenses"
+          title: "licenses: update tailscale{,d} licenses"
+          body: Triggered by ${{ github.repository }}@${{ github.sha }}
+          delete-branch: true
+          team-reviewers: opensource-license-reviewers
```
DentonGentry, line 62: I had no idea one could assign PR reviews to a GitHub team. This could be really useful in other contexts.

DentonGentry (approved):

.github/workflows/go-licenses.yml
willnorris: yep, that is the biggest concern with this action (and why I wanted your review), and something I talked through with @DentonGentry yesterday.  

Using a GitHub App gives us [pretty granular permissions](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens)... the app is only installed on this repo, and branch protection ensures it can't touch the main branch.  I've done a basic review of the code of both actions, and they seem to do only what they say on the tin.  I'm also pinning at specific versions of everything so the code can't change underneath us.

.github/workflows/go-licenses.yml
willnorris: oops, forgot to include android here!  I'll add that.

lxn should show up in the generated list for windows which will run as a similar action inside corp.  For non-go dependencies, we'll simply include them directly in the respective template file and maintain those lists manually (which shouldn't be too bad since they don't change often)

.github/workflows/go-licenses.yml
willnorris: hmmm... I just checked again and it appears this commit actually [only includes the compiled version](https://github.com/tibdex/github-app-token/tree/f717b5ecd4534d3c4df4ce9b5c1c2214f0f7cd06) of the code.  I was actually reviewing at main.  :-\

PR #5435, opened 2022-08-25
tka: implement thawing/freezing, document thread-unsafety

twitchyliquid64:
Control will shortly have a SQL-based tailchonk, so we'll need this to avoid a long-lived transaction.

I didn't see a way to do locking safely within the `Authority`, so I've added docs to delegate it to the caller.

/cc @crawshaw

### DIFF
--- a/tka/tka.go
+++ b/tka/tka.go
@@ -35,6 +35,11 @@ var cborDecOpts = cbor.DecOptions{
 // Authority objects can either be created from an existing, non-empty
 // tailchonk (via tka.Open()), or created from scratch using tka.Bootstrap()
 // or tka.Create().
+//
+// The Authority is NOT thread-safe. Specifically:
+//   - Calls to Inform() must be done in an exclusive section
+//   - No method calls may be pending during a call to Thaw()
+//   - All other methods are safe for concurrent use with other methods.
 type Authority struct {
 	head           AUM
 	oldestAncestor AUM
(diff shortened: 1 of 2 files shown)


crawshaw (commented):
Do we really need to `Thaw` an Authority? Can't we create a new one?

Also wouldn't `Thaw` have to adjust the internal state of the authority to the underlying Chonk storage?

twitchyliquid64:
Everytime you `Open()` an authority you need to do a bunch of computation to determine the correct state, which is stupid expensive if you do it for every map request. If you can be sure all mutations go through the same `Authority` (ie: we do the memory thing we do in control) then theres no point in doing an expensive `computeActiveChain()`.

crawshaw:
OK, so how do we deal with:

1. create a DB chonk
2. do stuff to it
3. tx.Rollback (or tx.Commit fails)

Now that authority state is out of step with the underlying chonk storage.

twitchyliquid64:
Good point. I guess we can either:

1. Invalidate the frozen authority and `Open()` for the next transaction after a rollback
2. Redo this PR so that you can have multiple `FrozenAuthority` objects - one for pre transaction and one for if it goes through. And then we can use the one corresponding to the success of the transaction.

twitchyliquid64:
I'm leaning towards 1 because its easy and we have a read-heavy workload.

sailorfrag:
I wonder if the need to do this indicates that we need to adjust the `Authority` API and/or the `Chonk` interface to incorporate the notion of transactions. I think it's also important to be clear about the DB semantics regarding external writers; it sounds like there's an underlying assumption that nobody else will mutate `Chonk` data between transactions which is consistent with what other parts of control are doing, but probably needs to be made clear.

crawshaw:
I agree that there seems to be a layering issue here. I think the current idea of a Chonk-per-tailner-per-tx may be the wrong move.

twitchyliquid64:
I'm not entirely sure what the right layering is. I guess half of the problem is that `Inform()` expects when `chonk.CommitVerifiedAUMs()` returns, they are committed (something that has no notion of a rollback).

Im not sure how, but maybe we want to plumb understanding of rollbacks into `Inform()` ?

Everything else in `tka.Authority` is read-only, which is nice.

twitchyliquid64:
Really all that needs to change is that `Inform()` 'holds on' to the `[]tka.AUM` until the actual commit, in which case it does the commit and checks for success.

twitchyliquid64:
@sailorfrag had an idea I really like: What if we stopped storing the `chonk` as a member on Authority (we only need it for startup and Inform() anyway), and passed it in method parameter to Inform()? That way:

  - Chonks can be parameterized for a specific query rather than be a long-running storage backend
  - Authority objects are now entirely a state snapshot, so we can make them cloneable
  - We could even make `Inform()` return a new Authority based on the application of the new AUMs, and then the wrapping logic can store either the original or the 'informed' authority based on rollback or commit.
  
I like this. @crawshaw thoughts?

twitchyliquid64:
solo buenos vibras

PR #5436, opened 2022-08-25
ipn/ipnlocal: fix Taildrop not returning any sharing targets

mihaip:
The CapabilityFileSharingTarget capability added by eb32847d8525e61c7435
is meant to control the ability to share with nodes not owned by the
current user, not to restrict all sharing (the coordination server is
not currently populating the capability at all)

Fixes tailscale/corp#6669


maisem (approved):
Thanks for fixing it!

PR #5437, opened 2022-08-25
tka: move disablement logic out-of-band from AUMs

twitchyliquid64:
It doesn't make a ton of sense for disablement to be communicated as an AUM, because
any failure in the AUM or chain mechanism will mean disablement wont function.

Instead, tracking of the disablement secrets remains inside the state machine, but
actual disablement and communication of the disablement secret is done by the caller.

### DIFF
--- a/tka/aum.go
+++ b/tka/aum.go
@@ -59,10 +59,6 @@ const (
 	//
 	// Only the KeyID optional field may be set.
 	AUMRemoveKey
-	// A DisableNL AUM describes the disablement of TKA.
-	//
-	// Only the DisablementSecret optional field may be set.
-	AUMDisableNL
 	// A NoOp AUM carries no information and is used in tests.
 	AUMNoOp
 	// A UpdateKey AUM updates the metadata or votes of an existing key.
@@ -84,8 +80,6 @@ func (k AUMKind) String() string {
 		return "add-key"
 	case AUMRemoveKey:
 		return "remove-key"
-	case AUMDisableNL:
-		return "disable-nl"
 	case AUMNoOp:
 		return "no-op"
 	case AUMCheckpoint:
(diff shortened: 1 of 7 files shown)


tka/aum.go
```diff
@@ -189,23 +185,16 @@ func (a *AUM) StaticValidate() error {
...
 			return errors.New("Checkpoint AUMs may only specify State")
 		}
-	case AUMDisableNL:
-		if len(a.DisablementSecret) == 0 {
-			return errors.New("DisableNL AUMs must specify a disablement secret")
-		}
-		if a.KeyID != nil || a.State != nil || a.Key != nil || a.Votes != nil || a.Meta != nil {
-			return errors.New("DisableNL AUMs may only specify a disablement secret")
-		}
 	}
```
sailorfrag, line 198: reject unknown/unused message kinds?

tka/aum.go
```diff
@@ -59,10 +59,9 @@ const (
 	//
 	// Only the KeyID optional field may be set.
 	AUMRemoveKey
-	// A DisableNL AUM describes the disablement of TKA.
-	//
-	// Only the DisablementSecret optional field may be set.
-	AUMDisableNL
+	// Formerly DisableNL, This AUM type is unused but kept to
+	// maintain ordering.
+	AUMUnused1
```
sailorfrag, line 64: I feel like maybe either:
- The test inputs that use this ID should be kept around to make sure that existing data continues to process without making things explode, or
- Removing the ID altogether makes sense, and we assume that old data using the old IDs will never be seen.

Right now it seems like if `AUMUnused1` is found, that's not an error, but nothing generates it anymore and it's ignored if seen.

sailorfrag (approved):

tka/aum.go
twitchyliquid64: Done.

tka/aum.go
twitchyliquid64: Ive gone ahead and removed it altogether, given theres nothing relying on this in the field yet. That did necessitate updating a bunch of tests, but no biggie.

sailorfrag (approved):

PR #5442, opened 2022-08-25
cmd/tailscale: add licenses subcommand

willnorris:
`tailscale licenses` will direct users to https://tailscale.com/licenses/tailscale for dependencies and licenses.

I'm not currently adding license information to tailscaled, since I believe we pretty much always distribute them together, and the link above covers the dependencies for both binaries.

Updates tailscale/corp#5780

### DIFF
--- a/cmd/tailscale/cli/licenses.go
+++ b/cmd/tailscale/cli/licenses.go
@@ -6,6 +6,7 @@ package cli
 
 import (
 	"context"
+	"runtime"
 
 	"github.com/peterbourgon/ff/v3/ffcli"
 )
@@ -19,11 +20,23 @@ var licensesCmd = &ffcli.Command{
 }
 
 func runLicenses(ctx context.Context, args []string) error {
+	var licenseURL string
+	switch runtime.GOOS {
+	case "android":
+		licenseURL = "https://tailscale.com/licenses/android"
+	case "darwin", "ios":
+		licenseURL = "https://tailscale.com/licenses/apple"
+	case "windows":
+		licenseURL = "https://tailscale.com/licenses/windows"
+	default:
+		licenseURL = "https://tailscale.com/licenses/tailscale"
+	}
+
 	outln(`
 Tailscale wouldn't be possible without the contributions of thousands of open
 source developers. To see the open source packages included in Tailscale and
 their respective license information, visit:
 
-    https://tailscale.com/licenses/tailscale`)
+    ` + licenseURL)
 	return nil
 }
--- a/cmd/tailscale/cli/licenses.go
+++ b/cmd/tailscale/cli/licenses.go
@@ -1,0 +1,29 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package cli
+
+import (
+	"context"
+
+	"github.com/peterbourgon/ff/v3/ffcli"
+)
+
+var licensesCmd = &ffcli.Command{
+	Name:       "licenses",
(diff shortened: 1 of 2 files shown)


cmd/tailscale/cli/licenses.go
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
```
DentonGentry, line 1: 2022

cmd/tailscale/cli/licenses.go
```diff
@@ -0,0 +1,29 @@
...
+	Exec:       runLicenses,
+}
+
+func runLicenses(ctx context.Context, args []string) error {
+	outln(`
+Tailscale wouldn't be possible without the contributions of thousands of open
+source developers. To see the open source packages included in Tailscale and
+their respective license information, visit:
+
+    https://tailscale.com/licenses/tailscale`)
```
DentonGentry, line 27: macOS systems and Windows systems can also run the `tailscale licenses` command (on macOS via `/Applications/Tailscale.app/Contents/MacOS/Tailscale licenses`). Should we check GOOS?

Tailscaled-on-MacOS would be a corner case, its GOOS=darwin but its actual dependencies would not be the same as licenses/apple

DentonGentry (approved):

cmd/tailscale/cli/licenses.go
willnorris: psh.  I have it on good authority it's [March 907th, 2020](https://covidstandardtime.com/).

cmd/tailscale/cli/licenses.go
willnorris: https://tailscale.com/licenses/tailscale includes the dependencies for the CLI for all architectures, so it will still be the right link.  Which is why https://tailscale.com/licenses/apple additionally points to the CLI page at the top, since we do distribute the CLI with it.  This is the 80% solution that ensures we're listing all of our dependencies, even if the grouping is a tad weird.

willnorris: So it would be correct for the CLI on all platforms to point to https://tailscale.com/licenses/tailscale, and then the system tray app on macOS will point to https://tailscale.com/licenses/apple

willnorris:
We will also want to add a link in the web UI that `tailscale web` serves, since that is the primary UI for synology, home assistant, and maybe others.  But there's not an obvious place to put it, so I'll maybe get some thoughts from @rosszurowski  on where to add it.

cmd/tailscale/cli/licenses.go
DentonGentry: Legit. It does feel like we never left March 2020.

cmd/tailscale/cli/licenses.go
DentonGentry: I'll admit to possibly overthinking this.

`/Applications/Tailscale.app/Contents/MacOS/Tailscale.app` links to `wireguard-apple` and uses it in operation.

`wireguard-apple` will not appear in `/Applications/Tailscale.app/Contents/MacOS/Tailscale.app licenses` because it is not in `tailscale.md`, because it doesn't appear in the OSS repo.

macOS only has to one giant app bundle, not separate tailscaled and tailscale binaries.

willnorris: ah, got it.  I misunderstood that part.  And not overthinking it... I'd like to get this right.  I'll add in a GOOS check

willnorris:
@DentonGentry ptal.  I doubt that android and iOS expose a CLI, but there's no harm in having it here.

DentonGentry (approved):

PR #5444, opened 2022-08-26
chirp: add a 10s timeout when communicating with BIRD

andrew-d:
Prior to this change, if BIRD stops responding wgengine.watchdogEngine
will crash tailscaled.

This happens because in wgengine.userspaceEngine, we end up blocking
forever trying to write a request to or read a response from BIRD with
wgLock held, and then future watchdog'd calls will block on acquiring
that mutex until the watchdog kills the process. With the timeout, we at
least get the chance to print an error message and decide whether we
want to crash or not.

Updates tailscale/coral#72

### DIFF
--- a/chirp/chirp.go
+++ b/chirp/chirp.go
@@ -11,15 +11,31 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"time"
+)
+
+const (
+	// Maximum amount of time we should wait when reading a response from BIRD.
+	responseTimeout = 10 * time.Second
 )
 
 // New creates a BIRDClient.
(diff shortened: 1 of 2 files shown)


maisem:
Context?

maisem (approved):

maisem:
Do you need a similar timeout for writes?

andrew-d:
> Do you need a similar timeout for writes?

Maybe? Writes to a closed socket will cause an `io.EOF` or "closed" error... though I guess BIRD could always just stop reading data until the socket buffer fills up and things hang. Let me add that, actually.

bradfitz (approved):

PR #5445, opened 2022-08-26
tka: make storage a parameter rather than an Authority struct member

twitchyliquid64:
Updates #5435

Based on the discussion in #5435, we can better support transactional data models
by making the underlying storage layer a parameter (which can be specialized for
the request) rather than a long-lived member of Authority.

Now that Authority is just an instantaneous snapshot of state, we can do things
like provide idempotent methods and make it cloneable, too.

/cc @crawshaw

### DIFF
--- a/tka/sync.go
+++ b/tka/sync.go
@@ -43,7 +43,19 @@ const (
 	ancestorsSkipShift = 2
 )
 
-func (a *Authority) syncOffer() (SyncOffer, error) {
+// SyncOffer returns an abbreviated description of the current AUM
+// chain, which can be used to synchronize with another (untrusted)
+// Authority instance.
+//
+// The returned SyncOffer structure should be transmitted to the remote
+// Authority, which should call MissingAUMs() using it to determine
+// AUMs which need to be transmitted. This list of AUMs from the remote
+// can then be applied locally with Inform().
+//
+// This SyncOffer + AUM exchange should be performed by both ends,
+// because its possible that either end has AUMs that the other needs
+// to find out about.
+func (a *Authority) SyncOffer(storage Chonk) (SyncOffer, error) {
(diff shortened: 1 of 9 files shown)


twitchyliquid64:
Using `InformIdempotent()` we can use exactly the same approach to transactions as `Cache[]`: Use `tx.addCommitCallback` to writeback the `Authority` returned to the cache.

tka/sync.go
```diff
@@ -96,8 +96,8 @@ func (a *Authority) syncOffer() (SyncOffer, error) {
 // This SyncOffer + AUM exchange should be performed by both ends,
 // because its possible that either end has AUMs that the other needs
 // to find out about.
-func (a *Authority) SyncOffer() (SyncOffer, error) {
-	return a.syncOffer()
+func (a *Authority) SyncOffer(storage Chonk) (SyncOffer, error) {
+	return a.syncOffer(storage)
 }
```
sailorfrag, line 101: Export `a.syncOffer` and then this should be unnecessary?

tka/tka.go
```diff
@@ -39,8 +39,15 @@ type Authority struct {
 	head           AUM
 	oldestAncestor AUM
 	state          State
+}
 
-	storage Chonk
+// Clone duplicates the Authority structure.
+func (a *Authority) Clone() Authority {
```
sailorfrag, line 45: Would it be better to return a `*Authority`? I thought that was the usual convention but I'm unclear on if there is a meaningful distinction in this case.

sailorfrag (approved):

tka/tka.go
twitchyliquid64: Done.

tka/sync.go
twitchyliquid64: Done.

PR #5447, opened 2022-08-26
wgengine: print in-flight operations on watchdog trigger

andrew-d:
In addition to printing goroutine stacks, explicitly track all in-flight
operations and print them when the watchdog triggers (along with the
time they were started at). This should make debugging watchdog failures
easier, since we can look at the longest-running operation(s) first.

### DIFF
--- a/wgengine/watchdog.go
+++ b/wgengine/watchdog.go
@@ -8,10 +8,12 @@
 package wgengine
 
 import (
+	"fmt"
 	"log"
 	"net/netip"
 	"runtime/pprof"
 	"strings"
+	"sync"
 	"time"
 
 	"tailscale.com/envknob"
@@ -38,11 +40,17 @@ func NewWatchdog(e Engine) Engine {
 		return e
 	}
 	return &watchdogEngine{
-		wrap:    e,
-		logf:    log.Printf,
-		fatalf:  log.Fatalf,
-		maxWait: 45 * time.Second,
+		wrap:     e,
+		logf:     log.Printf,
+		fatalf:   log.Fatalf,
+		maxWait:  45 * time.Second,
+		inFlight: make(map[inFlightKey]time.Time),
 	}
+}
+
+type inFlightKey struct {
+	op  string
+	ctr uint64
 }
 
 type watchdogEngine struct {
@@ -50,9 +58,31 @@ type watchdogEngine struct {
 	logf    func(format string, args ...any)
 	fatalf  func(format string, args ...any)
 	maxWait time.Duration
+
+	// Track the start time(s) of in-flight operations
+	inFlightMu  sync.Mutex
+	inFlight    map[inFlightKey]time.Time
+	inFlightCtr uint64
 }
 
 func (e *watchdogEngine) watchdogErr(name string, fn func() error) error {
+	// Track all in-flight operations so we can print more useful error
+	// messages on watchdog failure
+	e.inFlightMu.Lock()
+	key := inFlightKey{
+		op:  name,
+		ctr: e.inFlightCtr,
+	}
+	e.inFlightCtr++
+	e.inFlight[key] = time.Now()
+	e.inFlightMu.Unlock()
+
+	defer func() {
+		e.inFlightMu.Lock()
+		defer e.inFlightMu.Unlock()
+		delete(e.inFlight, key)
+	}()
+
 	errCh := make(chan error)
 	go func() {
 		errCh <- fn()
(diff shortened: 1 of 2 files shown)


wgengine/watchdog.go
```diff
@@ -38,21 +40,49 @@ func NewWatchdog(e Engine) Engine {
...
 type watchdogEngine struct {
 	wrap    Engine
 	logf    func(format string, args ...any)
 	fatalf  func(format string, args ...any)
 	maxWait time.Duration
+
+	// Track the start time(s) of in-flight operations
+	inFlight    map[inFlightKey]time.Time
+	inFlightCtr uint64
+	inFlightMu  sync.Mutex
```
bradfitz, line 65: the mutex conventionally goes before the chunk of fields it guards

wgengine/watchdog.go
```diff
@@ -38,21 +40,49 @@ func NewWatchdog(e Engine) Engine {
...
+	}
+	e.inFlightCtr++
+	e.inFlight[key] = time.Now()
+	e.inFlightMu.Unlock()
+
+	defer func() {
+		e.inFlightMu.Lock()
+		defer e.inFlightMu.Unlock()
+		delete(e.inFlight, key)
+	}()
```
bradfitz, line 84: sure is a bunch more work now for every engine method.

oh well I guess.

(I've never been a huge fan of this watchdog thing but it's been useful a couple times I suppose)

wgengine/watchdog.go
```diff
@@ -66,6 +96,21 @@ func (e *watchdogEngine) watchdogErr(name string, fn func() error) error {
 		buf := new(strings.Builder)
 		pprof.Lookup("goroutine").WriteTo(buf, 1)
 		e.logf("wgengine watchdog stacks:\n%s", buf.String())
+
+		// Opportunistically try to print the currently in-flight
+		// operations, but don't block to avoid a deadlock.
+		if e.inFlightMu.TryLock() {
```
bradfitz, line 102: why would you expect this mutex to be contended?

its critical section is tiny, no?

wgengine/watchdog.go
```diff
@@ -66,6 +96,21 @@ func (e *watchdogEngine) watchdogErr(name string, fn func() error) error {
...
+			defer e.inFlightMu.Unlock()
+
+			// Print everything as a single string to avoid log
+			// rate limits
+			var b []byte
+			for k, t := range e.inFlight {
+				b = fmt.Appendf(b, "in-flight[%d]: name=%s start=%s\n", k.ctr, k.op, t.Format(time.RFC3339Nano))
+			}
+			e.logf("wgengine watchdog in-flight:\n%s", string(b))
```
bradfitz, line 111: release inFlightMu before the call to e.logf?

bradfitz (approved):

wgengine/watchdog.go
andrew-d: I have no idea why I thought that. Moved to a regular `Lock` instead of `TryLock`.

wgengine/watchdog.go
```diff
@@ -66,6 +96,18 @@ func (e *watchdogEngine) watchdogErr(name string, fn func() error) error {
 		buf := new(strings.Builder)
 		pprof.Lookup("goroutine").WriteTo(buf, 1)
 		e.logf("wgengine watchdog stacks:\n%s", buf.String())
+
+		// Collect the list of in-flight operations for debugging.
+		var b []byte
+		e.inFlightMu.Lock()
+		for k, t := range e.inFlight {
+			b = fmt.Appendf(b, "in-flight[%d]: name=%s start=%s\n", k.ctr, k.op, t.Format(time.RFC3339Nano))
```
bradfitz, line 104: want to also print how long ago the operation was started? (which we usually round to a millisecond)

wgengine/watchdog.go
```diff
@@ -66,6 +96,18 @@ func (e *watchdogEngine) watchdogErr(name string, fn func() error) error {
...
+		}
+		e.inFlightMu.Unlock()
+
+		// Print everything as a single string to avoid log
+		// rate limits.
+		e.logf("wgengine watchdog in-flight:\n%s", string(b))
```
bradfitz, line 110: no need for the string(...) ... that just allocates unnecessarily

bradfitz (approved):

wgengine/watchdog.go
andrew-d: TIL - thanks

PR #5449, opened 2022-08-26
wgengine/wgcfg: only write peer headers if necessary

andrew-d:
On sufficiently large tailnets, even writing the peer header (~95 bytes)
can result in a large amount of data that needs to be serialized and
deserialized. Only write headers for peers that need to have their
configuration changed.

### DIFF
--- a/wgengine/wgcfg/writer.go
+++ b/wgengine/wgcfg/writer.go
@@ -16,8 +16,11 @@ import (
 
 // ToUAPI writes cfg in UAPI format to w.
 // Prev is the previous device Config.
-// Prev is required so that we can remove now-defunct peers
-// without having to remove and re-add all peers.
+//
+// Prev is required so that we can remove now-defunct peers without having to
+// remove and re-add all peers, and so that we can avoid writing information
+// about peers that have not changed since the previous time we wrote our
+// Config.
 func (cfg *Config) ToUAPI(logf logger.Logf, w io.Writer, prev *Config) error {
 	var stickyErr error
 	set := func(key, value string) {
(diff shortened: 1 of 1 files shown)


wgengine/wgcfg/writer.go
```diff
@@ -49,13 +49,29 @@ func (cfg *Config) ToUAPI(logf logger.Logf, w io.Writer, prev *Config) error {
 	// Add/configure all new peers.
 	for _, p := range cfg.Peers {
 		oldPeer, wasPresent := old[p.PublicKey]
+
```
raggi, line 52: The comment on ToUAPI should describe the new use of prev

wgengine/wgcfg/writer.go
```diff
@@ -49,13 +49,29 @@ func (cfg *Config) ToUAPI(logf logger.Logf, w io.Writer, prev *Config) error {
...
+		// We only want to write the peer header/version if we're about
```
raggi, line 53: It might be a good defense to always have these be true, unconditionally, if `!wasPresent`, as they're reflecting on a zero-value, and not setting the zero value may imply a default that isn't strictly zero.

raggi (changes requested):

wgengine/wgcfg/writer.go
andrew-d: This is technically a behavioural change, but I'm pretty sure it doesn't matter in practice since the zero values would always compare as non-equal.

andrew-d:
@raggi - PTAL; any additional thoughts?

raggi (approved):

PR #5451, opened 2022-08-26
wgengine: use Go API rather than UAPI for status

raggi:


danderson:
We switched _from_ this to uapi at some point in the past. Anyone remember why we did? I'm worried there's a chesterton's fence lurking here.

bradfitz:
> We switched from this to uapi at some point in the past. Anyone remember why we did? I'm worried there's a chesterton's fence lurking here.

We switched away from this when we switched away from our own wireguard-go fork to upstream wireguard-go.

The idea at the time was that UAPI was slow but we could probably make it faster, which we did, over a number of rounds. But I think we made it about as fast as we can.

raggi:
> @raggi, want to rebase this on the `wgengine/wgint` package for now instead?

done!

wgengine/userspace.go
```diff
@@ -990,126 +989,42 @@ func (e *userspaceEngine) getStatus() (*Status, error) {
...
+		if peer == nil {
+			continue
 		}
+
+		var p ipnstate.PeerStatusLite
+		p.NodeKey = key
+		p.RxBytes = int64(wgint.PeerRxBytes(peer))
+		p.TxBytes = int64(wgint.PeerTxBytes(peer))
+		p.LastHandshake = time.Unix(0, wgint.PeerLastHandshakeNano(peer))
+		peers = append(peers, p)
```
bradfitz, line 1027: I'd probably just

```go
peers = append(peers, ipnstate.PeerStatusLite{
     NodeKey: ...
     RxBytes: ...
})
```

but either way

bradfitz (approved):

wgengine/userspace.go
raggi: I'll apply that in the second PR when the code moves into the per-peer getter.

raggi: Oh, actually, I'm using the return var there, so 🤷🏼 :)

Issue #5340, opened 2022-08-11
Does the Tailscale NixOS package use a buggy version of Go?


bradfitz:
In https://github.com/tailscale/tailscale/issues/5156#issuecomment-1211556460, I saw that @mjibson was hitting a Go runtime bug that we fixed in our official 1.28.0 binaries.

But @mjibson's 1.28.0 systems are NixOS.

Those NixOS packages are maintained by Tailscale folk who use NixOS: @danderson, @Xe, @maisem, @twitchyliquid64.

Does NixOS not have Go 1.18.5?

bradfitz:
Sure doesn't look like it:

https://github.com/NixOS/nixpkgs/commits/master/pkgs/development/compilers/go/1.18.nix .... only up to 1.18.4.

Go 1.18.5 came out 2022-08-01, nine days ago:

https://go.dev/doc/devel/release#go1.18.minor

maisem:
Apparently not
https://github.com/NixOS/nixpkgs/blob/36cc29d837e7232e3176e4651e8e117a6f231793/pkgs/development/compilers/go/1.18.nix#L59

We should make the NixOS pacakges also use our fork of go.

bradfitz:
> We should make the NixOS pacakges also use our fork of go.

I'd kinda rather not. We don't want to have a fork. It just lets us move quickly when we need to.

But sure, if it's not objectionable to the NixOS way of doing things?

danderson:
Getting the toolchain fork into upstream nixos is... dicey. We have to use the "official" go toolchain nix machinery, but then we'd be maintaining such a build outside of the purview of the toolchain maintainers. I've put off trying just to not deal with the aggro of "you're using private APIs!"

Any ETA on a point release to fix the runtime bug? Those get pushed into nixos pretty promptly, and we can agitate for a backport.

bradfitz:
Fixed by https://github.com/NixOS/nixpkgs/commit/eba5f13c08cd9fda50b4cf0e6499b7e39e729bc4

PR #5456, opened 2022-08-27
tailcfg: add Hostinfo.GoVersion

bradfitz:
So next time something like #5340 happens we can identify all affected
nodes and have the control plane send them health warnings.


maisem (approved):

DentonGentry (approved):
Sure, seems safe enough.

PR #5457, opened 2022-08-27
util/deephash: rely on direct memory hashing for primitive kinds

dsnet:
Rather than separate functions to hash each kind,
just rely on the fact that these are direct memory hashable,
thus simplifying the code.


bradfitz:
Looks like 32-bit test cases need updating.

dsnet:
Fixed.

bradfitz (approved):

PR #5459, opened 2022-08-27
util/deephash: use unsafe.Pointer instead of reflect.Value

dsnet:
Use of reflect.Value.SetXXX panics if the provided argument was
obtained from an unexported struct field.
Instead, pass an unsafe.Pointer around and convert to a
reflect.Value when necessary (i.e., for maps and interfaces).
Converting from unsafe.Pointer to reflect.Value guarantees that
none of the read-only bits will be populated.

When running in race mode, we attach type information to the pointer
so that we can type check every pointer operation.
This also type-checks that direct memory hashing is within
the valid range of a struct value.

We add test cases that previously caused deephash to panic,
but now pass.

Performance:

	Hash              14.1µs ± 1%    14.1µs ± 1%    ~     (p=0.590 n=10+9)
	HashPacketFilter  2.53µs ± 2%    2.44µs ± 1%  -3.79%  (p=0.000 n=9+10)
	TailcfgNode       1.45µs ± 1%    1.43µs ± 0%  -1.36%  (p=0.000 n=9+9)
	HashArray         318ns ± 2%     318ns ± 2%    ~      (p=0.541 n=10+10)
	HashMapAcyclic    32.9µs ± 1%    31.6µs ± 1%  -4.16%  (p=0.000 n=10+9)

There is a slight performance gain due to the use of unsafe.Pointer
over reflect.Value methods. Also, passing an unsafe.Pointer (1 word)
on the stack is cheaper than passing a reflect.Value (3 words).

Performance gains are diminishing since SHA-256 hashing now dominates the runtime.

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -24,11 +24,9 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"encoding/hex"
-	"net/netip"
 	"reflect"
 	"sync"
 	"time"
-	"unsafe"
 
 	"tailscale.com/util/hashx"
 )
@@ -59,19 +57,6 @@ import (
 //	  hash of every map entry. With a sufficiently strong hash, this value is
 //	  theoretically "parsable" by looking up the hash in a magical map that
 //	  returns the set of entries for that given hash.
-
-// addressableValue is a reflect.Value that is guaranteed to be addressable
-// such that calling the Addr and Set methods do not panic.
-//
-// There is no compile magic that enforces this property,
-// but rather the need to construct this type makes it easier to examine each
-// construction site to ensure that this property is upheld.
-type addressableValue struct{ reflect.Value }
-
-// newAddressableValue constructs a new addressable value of type t.
-func newAddressableValue(t reflect.Type) addressableValue {
-	return addressableValue{reflect.New(t).Elem()} // dereferenced pointer is always addressable
-}
 
 const scratchSize = 128
 
@@ -134,12 +119,16 @@ func Hash(v any) (s Sum) {
 
 	rv := reflect.ValueOf(v)
 	if rv.IsValid() {
-		var va addressableValue
+		var t reflect.Type
+		var p pointer
 		if rv.Kind() == reflect.Pointer && !rv.IsNil() {
-			va = addressableValue{rv.Elem()} // dereferenced pointer is always addressable
+			t = rv.Type().Elem()
+			p = pointerOf(rv)
 		} else {
-			va = newAddressableValue(rv.Type())
+			t = rv.Type()
+			va := reflect.New(t).Elem()
 			va.Set(rv)
+			p = pointerOf(va.Addr())
 		}
 
 		// Always treat the Hash input as an interface (it is), including hashing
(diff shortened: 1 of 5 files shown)


util/deephash/deephash.go
```diff
@@ -452,46 +437,22 @@ func (h *hasher) hashAddrv(v addressableValue) {
 }
 
 func makeMemHasher(n uintptr) typeHasherFunc {
-	return func(h *hasher, v addressableValue) {
-		h.HashBytes(unsafe.Slice((*byte)(v.Addr().UnsafePointer()), n))
-	}
-}
-
-// hashSliceMem hashes v, of kind Slice, with a memhash-able element type.
-func (h *hasher) hashSliceMem(v addressableValue) {
```
dsnet, line 461: `hashSliceMem` is deleted because we inline where it is used in `genTypeHasher` so that we can access the element size from the type.

util/deephash/deephash.go
```diff
@@ -452,46 +437,22 @@ func (h *hasher) hashAddrv(v addressableValue) {
...
+			pe := p.arrayIndex(i, nb)
+			eti.hasher()(h, pe)
 		}
 	}
 }
 
-func noopHasherFunc(h *hasher, v addressableValue) {}
-
 func genHashArray(t reflect.Type, eti *typeInfo) typeHasherFunc {
-	if t.Size() == 0 {
```
dsnet, line 488: This check is unnecessary since we check for mem-hashable types at the start of `genTypeHasher`.
A zero length array is always mem-hashable.

util/deephash/deephash.go
```diff
@@ -452,46 +437,22 @@ func (h *hasher) hashAddrv(v addressableValue) {
...
-		return noopHasherFunc
-	}
-	et := t.Elem()
-	if typeIsMemHashable(et) {
```
dsnet, line 492: This check is unnecessary since we check for mem-hashable types at the start of `genTypeHasher`.

util/deephash/deephash.go
```diff
@@ -594,44 +557,13 @@ func (h *hasher) hashMap(v addressableValue, ti *typeInfo, checkCycles bool) {
...
 		sum.xor(mh.h.sum())
 	}
 	h.HashBytes(append(h.scratch[:0], sum.sum[:]...)) // append into scratch to avoid heap allocation
 }
 
-// visitStack is a stack of pointers visited.
-// Pointers are pushed onto the stack when visited, and popped when leaving.
-// The integer value is the depth at which the pointer was visited.
-// The length of this stack should be zero after every hashing operation.
-type visitStack map[pointer]int
```
dsnet, line 608: Moved to `pointer.go`.

util/deephash/deephash.go
```diff
@@ -348,92 +334,91 @@ func genTypeHasher(ti *typeInfo) typeHasherFunc {
...
+				h.visitStack.push(pm)
+				defer h.visitStack.pop(pm)
 			}
 			h.HashUint8(1) // indicates visiting a map
-			h.hashMap(v, ti, ti.isRecursive)
+			h.hashMap(v, ti)
 		}
 	case reflect.Pointer:
 		et := t.Elem()
-		if typeIsMemHashable(et) {
```
dsnet, line 371: Unnecessary since we check for mem-hashable types at the start of `genTypeHasher`.

util/deephash/pointer.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"net/netip"
+	"reflect"
+	"time"
+	"unsafe"
+)
+
+// unsafePointer is an untyped pointer.
+// It is the caller's responsibility to call operations on the correct type.
+type unsafePointer struct{ p unsafe.Pointer }
```
bradfitz, line 16: can you include a list of what types of untyped pointers are used in this package? there aren't more than a half dozen I think?

for memhashable types, to the value itself.
for slices, ... to the slice header or to the first value?
for arrays, to the first value,
etc

bradfitz (approved):

PR #5462, opened 2022-08-27
tailcfg, control/controlclient, ipn/ipnlocal: add c2n (control-to-node) system

bradfitz:
This lets the control plane can make HTTP requests to nodes.

Then we can use this for future things rather than slapping more stuff
into MapResponse, etc.


bradfitz:
(will add tests before marking it ready)

ipn/ipnlocal/c2n.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"io"
+	"net/http"
+)
+
+func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
+	switch r.URL.Path {
+	case "/hi":
+		io.WriteString(w, "hi\n")
```
maisem, line 15: Let's do an echo instead, it helps test payloads too

bradfitz: sure

ipn/ipnlocal/c2n.go
```diff
@@ -0,0 +1,21 @@
...
+	"io"
+	"net/http"
+)
+
+func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
+	switch r.URL.Path {
+	case "/echo":
+		// Test handler.
+		body, _ := io.ReadAll(r.Body)
+		w.Write(body)
```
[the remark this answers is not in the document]
bradfitz: thanks 🙄

bradfitz:
Ready for review. cc @andrew-d

control/controlclient/direct.go
```diff
@@ -1217,9 +1223,13 @@ func (c *Direct) answerPing(pr *tailcfg.PingRequest) {
 		c.logf("invalid PingRequest with no URL")
 		return
 	}
-	if pr.Types == "" {
+	switch pr.Types {
+	case "":
 		answerHeadPing(c.logf, httpc, pr)
 		return
+	case "c2n":
```
maisem, line 1230: return an error if `c2n` and `c.noiseConfigured == false`?

bradfitz: I need this for tests. There's a TODO from you in our integration tests to make noise work with testcontrol

maisem: I guess I should go fix that 😄

maisem: can you add a env override for now?

maisem (approved):

PR #5464, opened 2022-08-27
util/deephash: move map logic to separate function

dsnet:
This helps pprof better identify which Go kinds take the most time
since the kind is always in the function name.

There is a minor adjustment where we hash the length of the map
to be more on the cautious side.

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -58,13 +58,10 @@ import (
 //	  theoretically "parsable" by looking up the hash in a magical map that
 //	  returns the set of entries for that given hash.
 
-const scratchSize = 128
-
 // hasher is reusable state for hashing a value.
 // Get one via hasherPool.
 type hasher struct {
 	hashx.Block512
-	scratch    [scratchSize]byte
 	visitStack visitStack
 }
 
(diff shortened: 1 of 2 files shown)


util/deephash/deephash.go
```diff
@@ -476,6 +455,59 @@ func (seh sliceElementHasher) hash(h *hasher, p pointer) {
...
+			if idx, ok := h.visitStack.seen(pm); ok {
+				h.HashUint8(2) // indicates cycle
+				h.HashUint64(uint64(idx))
+				return
+			}
+			h.visitStack.push(pm)
+			defer h.visitStack.pop(pm)
+		}
+		h.HashUint8(1)                // indicates visiting map entries
+		h.HashUint64(uint64(v.Len())) // MUSTDO
```
bradfitz, line 486: what is MUSTDO?

dsnet: Ooops. It's a reminder to myself to think about whether this is necessary. I couldn't prove to myself that it is. But I couldn't prove to myself that it isn't either. I think it's safer to include.

bradfitz (approved):

PR #5467, opened 2022-08-27
util/deephash: require pointer in API

dsnet:
The entry logic of Hash has extra complexity to make sure
we always have an addressable value on hand.
If not, we heap allocate the input.
For this reason we document that there are performance benefits
to always providing a pointer.
Rather than documenting this, just enforce it through generics.

Also, delete the unused HasherForType function.
It's an interesting use of generics, but not well tested.
We can resurrect it from code history if there's a need for it.

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -6,7 +6,7 @@
 // without looping. The hash is only valid within the lifetime of a program.
 // Users should not store the hash on disk or send it over the network.
 // The hash is sufficiently strong and unique such that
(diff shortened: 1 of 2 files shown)


maisem (approved):

bradfitz (approved):
I wanted to start using HasherForType soon. Pity to delete it before it's even had a chance.

But I guess I can resurrect it later too, unless you can keep it easily now.

dsnet:
> But I guess I can resurrect it later too, unless you can keep it easily now.

I'll resurrect it.

PR #5471, opened 2022-08-28
util/deephash: handle slice edge-cases

dsnet:
It is unclear whether the lack of checking nil-ness of slices
was an oversight or a deliberate feature.
Lacking a comment, the assumption is that this was an oversight.

Also, expand the logic to perform cycle detection for recursive slices.
We do this on a per-element basis since a slice is semantically
equivalent to a list of pointers.


util/deephash/deephash.go
```diff
@@ -326,17 +331,34 @@ func makeSliceHasher(t reflect.Type) typeHasherFunc {
...
 	var hashElem typeHasherFunc
+	var isRecursive bool
 	init := func() {
 		hashElem = getTypeInfo(t.Elem()).hasher()
+		isRecursive = typeIsRecursive(t)
 	}
 
 	return func(h *hasher, p pointer) {
 		pa := p.sliceArray()
+		if pa.isNil() {
```
dsnet, line 342: If the lack of nil-checking is deliberate, I'll put a comment here to the effect.
The `deepprint` predecessor never did nil-checks for slices,
but it feels weird and inconsistent that we do nil checks for maps, but not slices.

At the end of the day, the semantics here is more governed by how other packages use this,
and I don't have the best insight into that.

dsnet:
This is the last semantic change to `deephash` I have stacked up.
I might revamp the tests in the future, but I feel that `deephash` is in a good state after this.

bradfitz (approved):

util/deephash/deephash.go
```diff
@@ -303,17 +308,38 @@ func makeSliceHasher(t reflect.Type) typeHasherFunc {
...
+					if idx, ok := h.visitStack.seen(pe.p); ok {
+						h.HashUint8(2) // indicates cycle
+						h.HashUint64(uint64(idx))
+						return
+					}
+					h.HashUint8(1) // indicates visiting slice element
+					h.visitStack.push(pe.p)
+					defer h.visitStack.pop(pe.p)
+					hashElem(h, pe)
+				}()
```
bradfitz, line 340: these immediately-called-funcs are kinda gross style.

just for the defer?

dsnet: Yea. I realized looking at the code earlier that it was buggy without it. Did you have an alternative suggestion?

bradfitz: Using method? Not using a defer?

dsnet: I realized that the recursive case is identical to normal hashing of elements, but with additional logic before.
Implemented it as a wrapper around the original `hashElem`.

bradfitz:
LGTM

Issue #697, opened 2020-08-20
illumos support


nshalman:
**Is your feature request related to a problem? Please describe.**

The `main` branch of tailscale/tailscale doesn't compile on/for illumos.

**Describe the solution you'd like**

I have hosts where `GOOS=illumos` and I want to run Tailscale on them.
Ideally I'd like to be able to just build from `main`.

**Describe alternatives you've considered**

I have configured Wireguard manually, and I've used Algo, but Tailscale is pretty sweet.

**Additional context**

I have managed to hack together working binaries based on [a fork of wireguard-go](https://github.com/nshalman/wireguard-go-illumos-wip/tree/tailscale-illumos) built on top of the work of someone else who does not have the bandwidth to get the changes upstream, but has no objection to them being upstreamed. My branch is [here](https://github.com/tailscale/tailscale/compare/main...nshalman:illumos).

Attempting to merge `main` fully this evening fails with errors I haven't had time to debug yet:
```$ GOOS=illumos go build tailscale.com/cmd/tailscaled
# tailscale.com/wgengine/tsdns
wgengine/tsdns/forwarder.go:211:23: undefined: os.ErrDeadlineExceeded
wgengine/tsdns/forwarder.go:228:23: undefined: os.ErrDeadlineExceeded
# tailscale.com/derp/derphttp
derp/derphttp/derphttp_client.go:518:9: undefined: tls.Dialer
``` 

Some questions on my mind:
- Is it reasonable to open up a PR against `tailscale/wireguard-go` with a cleaned up commit to bring in illumos support?
- Is there documentation on how to run the tests (I'm still relatively new to Go and am not finding it obvious...)
- Can I help provide documentation/tooling for how to get a usable test environment for illumos CI builds?

jclulow:
>     * The main thing we'd need for an illumos CI is instructions on how to run an Illumos box in a cheap cloud somewhere (e.g. DigitalOcean VPS), and getting the Github Actions runner working on it so we can plug it into our existing CI.

If DO is an option, this will be pretty easy to get going.  I have a pipeline for emitting images of OpenIndiana and OmniOS that will run in a droplet -- you can use the "import custom image" function and provide it a URL and it all pretty much just works.

>     * As an aside, we would also need some operational stuff in general: how do we package stuff for Illumos? How do we run stuff under SMF? Can this be automated in a similar vein to Debian packages?

[...]

PR #5473, opened 2022-08-28
cmd/tailscale: add emoji for illumos in status subcommand

nshalman:
As I close in on opening the PR for #697 it would be nice for non-illumos machines to give their illumos peers a nicer emoji instead of the default alien.
I think the Sun is fitting for SunOS operating systems.

It looks prettier in my browser when running the status web view than it does in the PR.


bradfitz (approved):

DentonGentry:
I had no idea this function was in the codebase.

bradfitz:
It's kinda a debug thing. It's more prominent on windows but even then pretty hidden. I'd love to see it fleshed out.

PR #5474, opened 2022-08-28
wgengine/router: allow disco packets through iptables on Linux

### LINKED ISSUE #3824: Try to open the host firewall for magicsock on linux
### What is the issue?

On linux, we don't try to open port 41641 for magicsock, partly because it's unnecessary for easy<>easy NAT traversal, and partly because the magicsock port can roam on rebind to another number, so we'd have to play some dicey firewall whackamole.

However, not opening the firewall breaks one NAT traversal case that should be a slam dunk:
 - server behind a Hard NAT
 - client behind a NAT with UPnP support
 - client machine has a stateful host firewall (e.g. ufw)

This should be a trivial<>hard combination, because UPnP makes the client side NAT disappear. But, the host firewall is still there, and will eat disco probes coming from the hard side, because they come from a surprising source port that the client hasn't communicated with. It _would_ communicate with it if the disco probe made it to tailscaled, and that would immediately get a working p2p link, but we can't get past the stateful firewall, so we never discover that trivially feasible path.

This makes what should be a trivial<>hard NAT case degrade back into easy<>hard, which you'll recall has no easy solutions for us right now (the bday paradox is infeasible in practice because of shallow session tables on routers). IOW, something that really ought to work amazingly well, doesn't :(

This _does_ work on Windows, because on Windows we talk to the WFP (firewall framework), and tell it to allow all UDP to our cryptographically strong process identity, so all inbound traffic just works regardless of how we rebind over time.

On linux, we can't do that, because netfilter has no way to ask what process is about to receive the packet, if it's allowed to proceed. The only thing we could implement would be to open up the specific port numbers we're using, and dynamically reconfigure the firewall every time we rebind magicsock and roam to a new port.

This is possible, but risky: if tailscaled crashes or stops without the opportunity for cleanup, that open port will stick around for ever. We have `tailscaled --cleanup` running for this automatically on stop for systemd-based distros, maybe that's enough?...

A more horrible alternative, for situations where we have network admin powers on the system (which we need anyway for firewall opening), would be to listen on a raw socket and attach an ebpf program to it that filters only traffic destined to tailscaled, and whose header contains the disco magic number. That would give us an extra receive funnel of all inbound disco traffic, before it ever hits netfilter. That way, we could grab those disco packets, use them to learn of a new usable endpoint, and punch our way out without having to imprecisely open the host firewall.

Neither option is particularly pleasant to contemplate, but getting trivial<>hard NAT cases to work would be a pretty big deal to double digit %age of our fleet, based on our portmapping stats.

bradfitz:
Fixes #3824

### DIFF
--- a/wgengine/router/router_linux.go
+++ b/wgengine/router/router_linux.go
@@ -1112,6 +1112,30 @@ func (r *linuxRouter) addNetfilterBase() error {
 	return nil
 }
 
+// allowDiscoRule is the iptables rule to allow all UDP discovery packets in.
+// See https://github.com/tailscale/tailscale/issues/3824.
+var allowDiscoRule = []string{
+	"-p", "udp", // UDP
+	"!", "-f", // not fragmented parts of UDP packets
+	"-m", "u32", "--u32",
+	"0>>22&0x3C@8=0x5453f09f && 0>>22&0x3C@10&0xffff=0x92ac", // check first 6 bytes of UDP payload is disco packet magic
+	"-m", "comment", "--comment", "Allow Tailscale NAT traversal",
+	"-j", "ACCEPT",
+}
+
+// allowDiscoRule6 is allowDiscoRule without the fragment check that doesn't
+// apply to IPv6, and with the right offsets for the IPv6 UDP payload.
+//
+// Note: this doesn't work if IPv6 extension headers are present. Oh well.
+// It'll help most users.
+var allowDiscoRule6 = []string{
+	"-p", "udp", // UDP
+	"-m", "u32", "--u32",
+	"0x30=0x5453f09f && 0x32&0xffff=0x92ac", // check first 6 bytes of UDP payload is disco packet magic
+	"-m", "comment", "--comment", "Allow Tailscale NAT traversal",
+	"-j", "ACCEPT",
+}
+
 // addNetfilterBase4 adds some basic IPv4 processing rules to be
 // supplemented by later calls to other helpers.
 func (r *linuxRouter) addNetfilterBase4() error {
@@ -1128,6 +1152,10 @@ func (r *linuxRouter) addNetfilterBase4() error {
 	args = []string{"!", "-i", r.tunname, "-s", tsaddr.CGNATRange().String(), "-j", "DROP"}
 	if err := r.ipt4.Append("filter", "ts-input", args...); err != nil {
 		return fmt.Errorf("adding %v in v4/filter/ts-input: %w", args, err)
+	}
+	if err := r.ipt4.Append("filter", "ts-input", allowDiscoRule...); err != nil {
+		// Not worth returning an error about. (Maybe they lack u32 kernel match support.)
+		r.logf("ignoring error adding allow-disco in v4/filter/ts-input: %v", err)
 	}
 
 	// Forward all traffic from the Tailscale interface, and drop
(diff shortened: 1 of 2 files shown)


DentonGentry (approved):
Do we want to do this literally the day before we build 1.30?
There really isn't time to test it.

bradfitz:
What tests should we do? I see it counting up upon getting disco packets. And I made it best effort in case their kernel doesn't have u32 support.

DentonGentry:
I'd want to try it on a system which can't handle unaligned accesses (we'd expect these accesses to be aligned, and maybe iptables handles unaligned accesses anyway, but I'd like to try it). I have an RPi somewhere I can dust off to build a tailscaled for.

bradfitz:
I can test on arm. But I just realized I didn't test IPv6 and I think the rule offsets are wrong on v6.

DentonGentry (approved):
Go ahead and submit when ready, if we find anything we'll handle it then.

danderson:
IIRC comments didn't work on the busybox implementation of iptables, or some other old iptables binary. We hit a bunch of compat issues because of that. I would remove the comment if you want best compat. Which is a shame because the rule sure looks sketchy without an explanation :)

But I'm also not clear on why this is needed. If we're sending bidirectionally, the linux firewall is just like any other stateful firewall, and it'll all work out after a few packets.

danderson:
nvm on the "why do we need this", I went and read the original bug. I guess that danderson guy knows what he's talking about.

danderson (changes requested):
Thinking more, I think this is unsafe as written. This allows any program on the system to receive network traffic they don't expect. A user might reasonably expect that a server behind their host firewall won't receive external traffic, but we're saying that any process can, as long as the packet has the right preamble. I think this is too wide a hole to open, especially if people are running random poorly written services on the assumption that the firewall is protecting them from RCE.

Looking at iptables now to see what we might add to improve this...

danderson:
Okay, `iptables-extensions` is as usual poor-to-useless for this problem.

IMO, if we want to do this with iptables, we'd have to make the match be specific to our magicsock port numbers, and update the rule whenever magicsock rebinds. It's an irritating amount of plumbing, and it creates brief races where the allowed packets don't go to the right destination... But I can't think of an obvious way to exploit that, and the race won't harm NAT traversal since a future packet will still get through and get everything working again.

Or we do the AF_PACKET+BPF trick, which we can do from pure Go without needing an ebpf compiler or any such machinery.

bradfitz:
> IMO, if we want to do this with iptables, we'd have to make the match be specific to our magicsock port numbers, and update the rule whenever magicsock rebinds. It's an irritating amount of plumbing, and it creates brief races where the allowed packets don't go to the right destination... But I can't think of an obvious way to exploit that, and the race won't harm NAT traversal since a future packet will still get through and get everything working again.

Not a terrible amount of plumbing. And once that's done we don't even need the scary-looking u32 match. We can just say "all UDP" and that's much more readable for admins.

> Or we do the AF_PACKET+BPF trick, which we can do from pure Go without needing an ebpf compiler or any such machinery.

What's that trick? Amusingly, Googling that yields [The wrong way to filter sockets with BPF](https://natanyellin.com/posts/ebpf-filtering-done-right/). 😂

andrew-d:
Y'all are really good at nerd-sniping 😂  Here's a commit with an example of the "BPF trick"–using a BPF packet filter to look for disco packets and then reading them in Go. I confirmed that it dumps disco packets on a machine running Tailscale, though I haven't tested it in cases where a firewall blocks them.

https://github.com/tailscale/tailscale/commit/4e3158464fde135c7e1945fb833e26b180f0ff03

bradfitz:
> Y'all are really good at nerd-sniping 😂 Here's a commit with an example of the "BPF trick"–using a BPF packet filter to look for disco packets and then reading them in Go. I confirmed that it dumps disco packets on a machine running Tailscale, though I haven't tested it in cases where a firewall blocks them.
>
> https://github.com/tailscale/tailscale/commit/4e3158464fde135c7e1945fb833e26b180f0ff03

Ah, fun! So we can make this be a 4th packet source in magicsock (v4, v6, derp, this)?

bradfitz:
(left some comments on that commit, in lieu of a PR)

bradfitz:
Closing this in favor of @andrew-d's WIP changes.

PR #5475, opened 2022-08-29
Improve NAT traversal by sharing endpoint info between peers.

awsong:
Signed-off-by: Andy Song <wsong.cn@gmail.com>

Related to issue #5396.

Sending dest IP-Port in disco Ping message, the recieving party
will then be able to figure out its "public facing ip/port" relative
to sending party. This effectively makes the sending party a STUN.
Pre-condition: the two parties already get direct connection.


wgengine/magicsock/magicsock.go
```diff
@@ -231,6 +231,7 @@ type Conn struct {
 	// This block mirrors the contents and field order of the Options
 	// struct. Initialized once at construction, then constant.
 
+	interEp                netip.AddrPort
```
bradfitz, line 234: what's the locking rules for reading/writing this? It shouldn't be in this section, which is static stuff.

Is it guarded by mu? Then it needs to go below mu (after line 320).

But I don't think that's currently the case. It seems the current reading/writing is a data race.

Also, this field needs docs on what it is. It isn't obvious from the name.

bradfitz (changes requested):
Can you flesh out the commit message and explain what this helps?

awsong:
> Can you flesh out the commit message and explain what this helps?

This helps the situation mentioned in #5396 , should we repeat it in commit comments or just reference it?

wgengine/magicsock/magicsock.go
awsong: > what's the locking rules for reading/writing this? It shouldn't be in this section, which is static stuff.
> 
> Is it guarded by mu? Then it needs to go below mu (after line 320).
> 
> But I don't think that's currently the case. It seems the current reading/writing is a data race.
> 
> Also, this field needs docs on what it is. It isn't obvious from the name.

Not sure I'm able to answer your question, not quit familiar with the general locking shceme here. 
The intention of this field is to temporarly store this interim endpoint information received from Ping message and use it when it's time to update local endpoint list.

wgengine/magicsock/magicsock.go
```diff
@@ -338,6 +338,14 @@ type Conn struct {
 
 	closed bool // Close was called
 
+	// interEps (intermediate Endpoints) holds the endpoints that are
+	// discovered in Ping messages received from peers. When a peer is
+	// located in private network, it could potentially see more/different
+	// endpoint addresses of ourselves from STUN. And this extra information
+	// could help other peers in the same private network. See issu #5396
+	// for details.
+	interEps map[netip.AddrPort]bool
```
tilosp, line 347: Simply adding all discovered endpoints to a map and never removing them seems problematic. If a host has changing ips and is running for long this would lead to an indefinitely growing list.

This problem could be solved by only keeping the newest endpoint for each remote peer. So basically `map[somepeerid]netip.AddrPort`

tilosp (changes requested):

wgengine/magicsock/magicsock.go
awsong: Thanks for commenting. Changed according to your suggestion.

tilosp (approved):

bradfitz:
I don't think we want to accept this change. It only benefits nodes temporarily in a relatively uncommon network configuration.

If it always helped people in that configuration that'd be more interesting, but since this only helps when a 3rd node helps out the first two, that's kinda fringe. I'd rather we do something else.

awsong:
> I don't think we want to accept this change. It only benefits nodes temporarily in a relatively uncommon network configuration.
> 
> If it always helped people in that configuration that'd be more interesting, but since this only helps when a 3rd node helps out the first two, that's kinda fringe. I'd rather we do something else.

Understand your point. But I really can't think of a way how to help without a 3rd node. 

Maybe traceroute could give some extra information, but that info doesn't seem enough.

awsong:
> I don't think we want to accept this change. It only benefits nodes temporarily in a relatively uncommon network configuration.

The configuration might not be that uncommon. For ISPs who deploy CGNAT, all the nodes behind home router will benefit if one node is running Tailscale w/o router or on its router.

bradfitz:
But I'm going to close this PR per comments above. We can discuss other possible solutions on the issue. I imagine what we'll want to do is let you run a `derper` binary or similar on your internal NAT network in STUN-only mode that doesn't affect the hard-vs-easy NAT detection.

PR #5483, opened 2022-08-29
wgengine/wgint: add shady temporary package to get at wireguard internals

bradfitz:
For #5451


crawshaw (approved):
This go package is definitely wearing a trenchcoat.

PR #5484, opened 2022-08-29
wgengine/magicsock: use AF_PACKET socket + BPF to read disco messages

andrew-d:
This is entirely optional (i.e. failing in this code is non-fatal) and
only enabled on Linux for now. Additionally, this new behaviour can be
disabled by setting the TS_DEBUG_DISABLE_AF_PACKET environment variable.

Updates https://github.com/tailscale/tailscale/issues/3824
Replaces https://github.com/tailscale/tailscale/pull/5474


wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,316 @@
...
+		// Next header indicates we have a UDP packet, so the header
+		// offset is immediately following the IPv6 header.
+		bpf.LoadConstant{Dst: bpf.RegX, Val: ethHeaderSize + ipv6.HeaderLen},
+		bpf.Jump{Skip: 6}, // move past fragment checks
+
+		// Check if the "Next Header" field is the "Fragment" header
+		bpf.JumpIf{Cond: bpf.JumpEqual, Val: 0x2C, SkipTrue: 1, SkipFalse: 0},
+		bpf.RetConstant{Val: 0x0}, // skip packet
+
+		// If it's the fragment header, check whether the protocol is UDP
```
danderson, line 120: You could almost certainly also decline to handle fragmented IPv6 packets. IPv6 strongly states that fragmentation is understanding-optional in IPv6, meaning that most of the internet doesn't generate any.

But if you do handle it, you need to add an extra check to only process packets with fragment offset == 0, otherwise you'll have false positives on later fragments.

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,316 @@
...
+		return
+	}
+	c.listenDiscoProto(&c.bpfConn4, "IPv4", magicsockFilterV4)
+	c.listenDiscoProto(&c.bpfConn6, "IPv6", magicsockFilterV6)
+}
+
+func (c *Conn) listenDiscoProto(fdptr *int, name string, filter []bpf.Instruction) {
+	// We create our socket manually here, rather than using
+	// {ipv4,ipv6}.NewPacketConn, because they don't support SOCK_RAW.
+	fd, err := unix.Socket(unix.AF_PACKET, unix.SOCK_RAW, int(htons(unix.ETH_P_ALL)))
```
danderson, line 162: c, err := net.ListenPacket("ip4:17", "0.0.0.0")

Will make a raw UDP-only listener. As a bonus, the packets you receive won't have an ethernet envelope, meaning they'll work on non-ethernet interfaces and save you a little BPF code. And you can then use x/net/ipv4 to set the BPF filter, and end up with packet listener that integrate gracefully with the runtime netpoller.

Same for IPv6.

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,316 @@
...
+	if c.bpfConn6 != 0 {
+		unix.Close(c.bpfConn6)
+	}
+}
+
+// handleDiscoPacket is called by listenPacketsWithFilter when a packet
+// matching a BPF filter is received. It can be passed either an IPv4 or IPv6
+// packet and will handle either case.
+func (c *Conn) handleDiscoPacket(packet []byte, from unix.Sockaddr) error {
+	pkt := gopacket.NewPacket(packet, layers.LayerTypeEthernet, gopacket.NoCopy)
```
danderson, line 190: Another benefit of using net.ListenPacket above, is that you can use net/packet to pull the packet apart here, and avoid the gopacket dependency just for dissecting ethernet.

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,316 @@
...
+// BPF filter to it, and then calls 'cb' with all packets that are received
+// from the socket (and thus ones that match the given filter).
+func (c *Conn) listenPacketsWithFilter(fd int, filt []bpf.Instruction, cb func([]byte, unix.Sockaddr) error) error {
+	// This filter ignores all packets
+	zeroFilter := bpfFilter{bpf.RetConstant{Val: 0x0}}
+	if err := zeroFilter.ApplyTo(fd); err != nil {
+		return fmt.Errorf("applying zero BPF filter: %w", err)
+	}
+
+	// Drain any existing packets in the queue, since things can
```
danderson, line 247: Unnecessary, the generic receive code will do the right thing even when handed non-disco packets, won't it?

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,316 @@
...
+	)
+	for {
+		n, _ /* from */, _ /* err */ := unix.Recvfrom(fd, drain[:], unix.MSG_DONTWAIT)
+		if n == -1 {
+			// We assume the error here means there is nothing left
+			// to read from the socket, which is exactly what we
+			// want. If there are any further errors, we'll bail out below anyway.
+			break
+		}
+		nDrain++
```
danderson, line 262: Unused.

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,316 @@
...
+			return err
+		}
+	}
+}
+
+// bpfFilter represents a BPF filter program that can be applied to a socket.
+type bpfFilter []bpf.Instruction
+
+// ApplyTo applies the current filter to a socket with the provided file descriptor.
+func (filter bpfFilter) ApplyTo(fd int) error {
```
danderson, line 289: Can delete all of this once using x/net/ipv{4,6} :)

wgengine/magicsock/magicsock_nonlinux.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
```
danderson, line 1: We usually call these files `_default.go`, to indicate that it's the filler implementation when nothing better exists.

Also means we don't have to rename it when we add this for e.g. the BSDs, which are likely to just work with linux's code since they also support classic BPF.

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,316 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build linux
+// +build linux
```
bradfitz, line 6: delete

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,316 @@
...
+	udp := udpL.(*layers.UDP)
+	srcPort := udp.SrcPort
+
+	srcAddr, ok := netip.AddrFromSlice(srcIP)
+	if !ok {
+		return nil
+	}
+
+	src := netip.AddrPortFrom(srcAddr, uint16(srcPort))
+	success := c.handleDiscoMessage(udp.Payload, src, key.NodePublic{})
```
bradfitz, line 221: are you just handling disco messages twice via both this new path and the old path in some cases?

I figured we'd do some initial runtime probe on start-up to see if this BPF stuff was working and if so then disable the handleDiscoMessage-via-UDP path in magicsock

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,317 @@
...
+	debugDisableAF_PACKET = envknob.Bool("TS_DEBUG_DISABLE_AF_PACKET")
+)
+
+// Convert our disco magic number into a uint32 and uint16 to test
+// against. We panic on an incorrect length here rather than try to be
+// generic with our BPF instructions below.
+//
+// Note that BPF uses network byte order (big-endian) when loading data
+// from a packet, so that is what we use to generate our magic numbers.
+var magic1, magic2 = (func() (uint32, uint16) {
```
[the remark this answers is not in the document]
bradfitz: I agree about dropping the func but I'd rather re-use disco.Magic directly or at least have the constants in the disco package and tested there.

I'd personally go with:

```
var (
    magic1 = binary.BigEndian.Uint32([]byte(disco.Magic[0:4]))
    magic2 = binary.BigEndian.Uint16([]byte(disco.Magic[4:6]))
)
```

Or:

```
const (
   magic1 = disco.Magic[0] << 24 | disco.Magic[1] << 16 ...
   magic2 ...
)
```

danderson:
Closing this PR so I can reopen it and make andrew the reviewer.

PR #5485, opened 2022-08-29
logtail: do not log when backing off

dsnet:
We may be backing off due to rate-limiting.
Generating more logs is counter-productive.

### DIFF
--- a/logtail/logtail.go
+++ b/logtail/logtail.go
@@ -17,6 +17,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


bradfitz (commented):
It should at least log to stderr, no?

bradfitz:
(or whatever layer's not to be uploaded)

dsnet:
> It should at least log to stderr, no?

Done. It's not all that useful of a log message, but it doesn't hurt either (as long as it doesn't get uploaded).

bradfitz (approved):

PR #5486, opened 2022-08-29
tka: implement credential signatures (key material delegation)

twitchyliquid64:
This will be needed to support preauth-keys with network lock in the future, so getting the core mechanics out of the way now.

### DIFF
--- a/tka/sig.go
+++ b/tka/sig.go
@@ -33,6 +33,19 @@ const (
 	// SigRotation signature and sign it again with their rotation key. That
 	// way, SigRotation nesting should only be 2 deep in the common case.
 	SigRotation
+	// SigCredential describes a signature over a specifi public key, signed
+	// by a key in the tailnet key authority referenced by the specified keyID.
+	// In effect, SigCredential delegates the ability to make a signature to
+	// a different public/private key pair.
+	//
+	// It is intended that a different public/private key pair be generated
+	// for each different SigCredential that is created. Implementors must
+	// take care that the private side is only known to the entity that needs
+	// to generate the wrapping SigRotation signature, and it is immediately
+	// discarded after use.
+	//
+	// SigCredential is expected to be nested in a SigRotation signature.
+	SigCredential
 )
 
 func (s SigKind) String() string {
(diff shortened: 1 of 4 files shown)


tka/sig.go
```diff
@@ -74,6 +84,9 @@ type NodeKeySignature struct {
 	//
 	// Intermediate SigRotation signatures may omit this value to use the
 	// parent one.
+	//
+	// SigCredential signatures use this field to specify the public key
```
sailorfrag, line 88: Should we change the name of the field to not refer to rotation then?

`IntermediatePubkey` or something?

tka/sig.go
```diff
@@ -139,14 +152,17 @@ func (s *NodeKeySignature) Unserialize(data []byte) error {
 }
 
 // verifySignature checks that the NodeKeySignature is authentic, certified
-// by the given verificationKey, and authorizes the given nodeKey.
+// by the given verificationKey, and (for all signatures except SigCredential)
```
sailorfrag, line 155: Perhaps enumerate the ones that it's true for, rather than the exceptions, so it won't go stale the next time something is added 😄

tka/sig_test.go
```diff
@@ -175,6 +175,82 @@ func TestSigNested_DeepNesting(t *testing.T) {
 	}
 }
 
+func TestSigCredential(t *testing.T) {
```
sailorfrag, line 178: Is there a test that the credential can't be used more than once?

sailorfrag (approved):

tka/sig_test.go
twitchyliquid64: Theres nothing wrong with using a credential multiple times for authorizing different node keys. Unless you mean a check that SigCredential cant nest another signature?

tka/sig.go
twitchyliquid64: Ive gone with `WrappingPubkey`, wdyt?

tka/sig_test.go
sailorfrag: > SigCredential describes a signature over a single-use public key

I feel like that implies it's only usable once.

Anyway that makes preauth keys really powerful and I guess generation and distribution of those will need to be done in a way that control never sees them.

twitchyliquid64: 100%, we definitely need to make sure control never sees them. I was thinking that, after someone generates a preauth key in the admin panel, they can do a `tailscale lock key-wrap` to generate and encode the `SigCredential` and its private key around the preauth key.

I'll update the language to remove the confusion that the public key can only be used once, rather than that the generated public key is intended to only be used for a single SigCredential.

PR #5487, opened 2022-08-29
tailcfg: redo/implement TKA wire format for enablement, sync, disablement

twitchyliquid64:
Let the bikeshedding commence XD

/cc @bradfitz

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -903,11 +903,6 @@ type MapRequest struct {
 	Stream      bool // if true, multiple MapResponse objects are returned
 	Hostinfo    *Hostinfo
 
-	// TKA describes request parameters relating to a local instance of
-	// the tailnet key authority. This field is omitted if a local instance
-	// is not running.
-	TKA *TKAMapRequest `json:",omitempty"`
-
 	// Endpoints are the client's magicsock UDP ip:port endpoints (IPv4 or IPv6).
 	Endpoints []string
 	// EndpointTypes are the types of the corresponding endpoints in Endpoints.
@@ -1347,9 +1342,15 @@ type MapResponse struct {
 	// ControlTime, if non-zero, is the current timestamp according to the control server.
 	ControlTime *time.Time `json:",omitempty"`
 
-	// TKA, if non-nil, describes updates for the local instance of the
-	// tailnet key authority.
-	TKA *TKAMapResponse `json:",omitempty"`
+	// TKAInfo describes the control plane's view of tailnet
+	// key authority (TKA) state.
+	//
+	// An initial nil TKAInfo indicates that the control plane
+	// believes TKA should not be enabled. An initial non-nil TKAInfo
+	// indicates the control plane believes TKA should be enabled.
+	// A nil TKAInfo in a mapresponse stream (i.e. a 'delta' mapresponse)
+	// indicates no change from the value sent earlier.
+	TKAInfo *TKAInfo `json:",omitempty"`
 
 	// Debug is normally nil, except for when the control server
 	// is setting debug settings on a node.
@@ -1851,85 +1852,6 @@ type PeerChange struct {
 	// It's a pointer to a slice for "omitempty", to allow differentiating
 	// a change to empty from no change.
 	Capabilities *[]string `json:",omitempty"`
-}
-
-// TKAInitBeginRequest submits a genesis AUM to seed the creation of the
-// tailnet's key authority.
-type TKAInitBeginRequest struct {
-	NodeID NodeID // NodeID of the initiating client
-
-	GenesisAUM tkatype.MarshaledAUM
-}
-
-// TKASignInfo describes information about an existing node that needs
-// to be signed into a node-key signature.
-type TKASignInfo struct {
-	NodeID     NodeID // NodeID of the node-key being signed
-	NodePublic key.NodePublic
-
-	// RotationPubkey specifies the public key which may sign
-	// a NodeKeySignature (NKS), which rotates the node key.
-	//
-	// This is necessary so the node can rotate its node-key without
-	// talking to a node which holds a trusted network-lock key.
-	// It does this by nesting the original NKS in a 'rotation' NKS,
-	// which it then signs with the key corresponding to RotationPubkey.
-	//
-	// This field expects a raw ed25519 public key.
-	RotationPubkey []byte
-}
-
-// TKAInitBeginResponse describes node information which must be signed to
-// complete initialization of the tailnets' key authority.
-type TKAInitBeginResponse struct {
-	NeedSignatures []TKASignInfo
-}
-
-// TKAInitFinishRequest finalizes initialization of the tailnet key authority
-// by submitting node-key signatures for all existing nodes.
-type TKAInitFinishRequest struct {
-	NodeID NodeID // NodeID of the initiating client
-
-	Signatures map[NodeID]tkatype.MarshaledSignature
-}
-
-// TKAInitFinishResponse describes the successful enablement of the tailnet's
-// key authority.
-type TKAInitFinishResponse struct{}
-
-// TKAMapRequest describes request parameters relating to the tailnet key
-// authority instance on this node. This information is transmitted as
-// part of the MapRequest.
-type TKAMapRequest struct {
-	// Head is the AUMHash of the latest authority update message committed
-	// by this node.
-	Head string // tka.AUMHash.String
-}
-
-// TKAMapResponse encodes information for the tailnet key authority
-// instance on the node. This information is transmitted as
-// part of the MapResponse.
-//
-// If there are no updates to be transmitted (in other words, if both
-// control and the node have the same head hash), len(Updates) == 0 and
-// WantSync is false.
-//
-// If control has updates that build off the head hash reported by the
-// node, they are simply transmitted in Updates (avoiding the more
-// expensive synchronization process).
-//
-// In all other cases, WantSync is set to true, and the node is expected
-// to reach out to control separately to synchronize.
-type TKAMapResponse struct {
-	// Updates is any AUMs that control believes the node should apply.
-	Updates []tkatype.MarshaledAUM `json:",omitempty"`
-
-	// WantSync is set by control to request the node complete AUM
-	// synchronization.
-	//
-	// TODO(tom): Implement AUM synchronization, probably as noise endpoints
-	// /machine/tka/sync/offer & /machine/tka/sync/send.
-	WantSync bool `json:",omitempty"`
 }
 
 // DerpMagicIP is a fake WireGuard endpoint IP address that means to
--- a/tailcfg/tka.go
+++ b/tailcfg/tka.go
@@ -1,0 +1,161 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package tailcfg
+
+import (
+	"tailscale.com/types/key"
+	"tailscale.com/types/tkatype"
+)
+
+// TKAInitBeginRequest submits a genesis AUM to seed the creation of the
+// tailnet's key authority.
+type TKAInitBeginRequest struct {
+	// NodeID is the node of the initiating client.
+	// It must match the machine key being used to communicate over noise.
+	NodeID NodeID
+
+	// GenesisAUM is the initial (genesis) AUM that the node generated
+	// to bootstrap tailnet key authority state.
+	GenesisAUM tkatype.MarshaledAUM
+}
+
+// TKASignInfo describes information about an existing node that needs
+// to be signed into a node-key signature.
+type TKASignInfo struct {
+	// NodeID is the ID of the node which needs a signature. It must
+	// correspond to NodePublic.
+	NodeID NodeID
+	// NodePublic is the node (Wireguard) public key which is being
+	// signed.
+	NodePublic key.NodePublic
+
+	// RotationPubkey specifies the public key which may sign
+	// a NodeKeySignature (NKS), which rotates the node key.
+	//
+	// This is necessary so the node can rotate its node-key without
+	// talking to a node which holds a trusted network-lock key.
+	// It does this by nesting the original NKS in a 'rotation' NKS,
+	// which it then signs with the key corresponding to RotationPubkey.
+	//
+	// This field expects a raw ed25519 public key.
+	RotationPubkey []byte
+}
+
+// TKAInitBeginResponse is the JSON response from a /tka/init/begin RPC.
+// This structure describes node information which must be signed to
+// complete initialization of the tailnets' key authority.
+type TKAInitBeginResponse struct {
+	// NeedSignatures specify information about the nodes in your tailnet
+	// which need initial signatures to function once the tailnet key
+	// authority is in use. The generated signatures should then be
+	// submitted in a /tka/init/finish RPC.
+	NeedSignatures []TKASignInfo
+}
+
+// TKAInitFinishRequest is the JSON request of a /tka/init/finish RPC.
+// This RPC finalizes initialization of the tailnet key authority
+// by submitting node-key signatures for all existing nodes.
+type TKAInitFinishRequest struct {
+	// NodeID is the node ID of the initiating client.
+	NodeID NodeID
+
+	// Signatures are serialized tka.NodeKeySignatures for all nodes
+	// in the tailnet.
+	Signatures map[NodeID]tkatype.MarshaledSignature
+}
+
+// TKAInitFinishResponse is the JSON response from a /tka/init/finish RPC.
+// This schema describes the successful enablement of the tailnet's
+// key authority.
+type TKAInitFinishResponse struct {
+	// Nothing. (yet?)
+}
+
+// TKAInfo encodes the control plane's view of tailnet key authority (TKA)
+// state. This information is transmitted as part of the MapResponse.
+type TKAInfo struct {
+	// Head describes the hash of the latest AUM applied to the authority.
+	// Head is encoded as tka.AUMHash.MarshalText.
+	//
+	// If the Head state differs to that known locally, the node should perform
+	// synchronization via a separate RPC.
+	//
+	// TODO(tom): Implement AUM synchronization as noise endpoints
+	// /machine/tka/sync/offer & /machine/tka/sync/send.
+	Head string `json:",omitempty"`
+
+	// Disabled indicates the control plane believes TKA should be disabled,
+	// and the node should reach out to fetch a disablement
+	// secret. If the disablement secret verifies, then the node should then
+	// disable TKA locally.
+	// This field exists to disambiguate a nil TKAInfo in a delta mapresponse
+	// from a nil TKAInfo indicating TKA should be disabled.
+	//
+	// TODO(tom): Implement /machine/tka/boostrap as a noise endpoint, to
+	// communicate the genesis AUM & any disablement secrets.
+	Disabled bool `json:",omitempty"`
+}
+
+// TKABootstrapRequest is sent by a node to get information necessary for
+// enabling or disabling the tailnet key authority.
+type TKABootstrapRequest struct {
+	// Head represents the node's head AUMHash (tka.Authority.Head), if
+	// network lock is enabled.
+	Head string
+}
+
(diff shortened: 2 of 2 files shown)


tailcfg/tailcfg.go
```diff
@@ -1330,9 +1329,9 @@ type MapResponse struct {
 	// ControlTime, if non-zero, is the current timestamp according to the control server.
 	ControlTime *time.Time `json:",omitempty"`
 
-	// TKA, if non-nil, describes updates for the local instance of the
-	// tailnet key authority.
-	TKA *TKAMapResponse `json:",omitempty"`
+	// TKAInstructions, if non-nil, describes updates for the local instance
+	// of the tailnet key authority.
+	TKA *TKAInstructions `json:",omitempty"`
```
bradfitz, line 1334: We'd discussed using the c2n http mechanism from control to node for this, no? Or is this something else?

twitchyliquid64: Is everything in map request/response going to be moving to c2n? We should use whichever mechanism is going to be appropriate for tailnet-wide 'settings' changes and the like.

bradfitz: depends on whether it's imperative or declarative.

declarative stuff will stay on MapResponse.

But "do this" will be c2n.

twitchyliquid64: WRT our offsite discussion: I was mostly thinking of using c2n for doing things like asking a node to direct-sign a rotation-signature, which fits a bit better into the c2n model.

tailcfg/tailcfg.go
```diff
@@ -1330,9 +1329,9 @@ type MapResponse struct {
 	// ControlTime, if non-zero, is the current timestamp according to the control server.
 	ControlTime *time.Time `json:",omitempty"`
 
-	// TKA, if non-nil, describes updates for the local instance of the
-	// tailnet key authority.
-	TKA *TKAMapResponse `json:",omitempty"`
+	// TKAInstructions, if non-nil, describes updates for the local instance
```
bradfitz, line 1332: The field's named TKA, not TKAInstructions

tailcfg/tailcfg.go
twitchyliquid64: Okay gotcha, then the 'plz resign' stuff will be c2n, and TKAInstructions should stay in <notification-channel> / map response

tailcfg/tailcfg.go
twitchyliquid64: Fixed.

twitchyliquid64:
Refactored to be as we discussed @bradfitz. Also went ahead and spitballed what the separate RPCs will look like:

 - "bootstrap" to communicate enablement and disablement values
 - sync "offer" and sync "send" to do the actual AUM exchange process

tailcfg/tailcfg.go
```diff
@@ -1347,9 +1342,9 @@ type MapResponse struct {
 	// ControlTime, if non-zero, is the current timestamp according to the control server.
 	ControlTime *time.Time `json:",omitempty"`
 
-	// TKA, if non-nil, describes updates for the local instance of the
-	// tailnet key authority.
-	TKA *TKAMapResponse `json:",omitempty"`
+	// TKAInfo, if non-nil, describes controls' view of tailnet key
```
bradfitz, line 1345: describes the control plane's view ....

tailcfg/tailcfg.go
```diff
@@ -1347,9 +1342,9 @@ type MapResponse struct {
...
+	// authority state.
```
bradfitz, line 1346: Document what nil means for both the initial MapResponse and on streams ones, where zero for all these fields means "no change from before".

For the initial MapResponse, is a nil TKAInfo equivalent to a pointer to the zero value?

tailcfg/tailcfg.go
```diff
@@ -1897,39 +1892,88 @@ type TKAInitFinishRequest struct {
...
-//
-// In all other cases, WantSync is set to true, and the node is expected
-// to reach out to control separately to synchronize.
-type TKAMapResponse struct {
-	// Updates is any AUMs that control believes the node should apply.
-	Updates []tkatype.MarshaledAUM `json:",omitempty"`
-
-	// WantSync is set by control to request the node complete AUM
-	// synchronization.
+// TKAInfo encodes controls' view of tailnet key authority (TKA) state.
```
bradfitz, line 1895: It would be "control's" (it's not a possessive of a plural / ending in -s, etc)

But normally we don't say just "control" in oss docs and prefer "the control plane" or "the coordination server". It's not entirely consistent.

tailcfg/tailcfg.go
```diff
@@ -1897,39 +1892,88 @@ type TKAInitFinishRequest struct {
...
+// This information is transmitted as part of the MapResponse.
+type TKAInfo struct {
+	// Head describes the hash of the latest AUM applied to the authority.
+	// Head is encoded as tka.AUMHash.MarshalText()
```
bradfitz, line 1899: drop parens & add trailing period.  Go docs generally omit the parens when describing funcs.

tailcfg/tailcfg.go
```diff
@@ -1897,39 +1892,88 @@ type TKAInitFinishRequest struct {
...
+	// Enabled describes whether the tailnet key authority should be enabled.
+	//  - False: TKA should be disabled. If it is enabled locally, the node
+	//    should reach out to fetch a disablement secret. If the disablement
+	//    secret verifies, then the node should then disable TKA locally.
+	//  - True: TKA should be enabled. If it is not, the node should reach
+	//    out via separate RPC to fetch the genesis AUM and initialize.
+	//
+	// TODO(tom): Implement /machine/tka/boostrap as a noise endpoint, to
+	// communicate the genesis AUM & any disablement secrets.
+	Enabled bool `json:",omitempty"`
```
bradfitz, line 1917: probably don't need omitempty here. the outer one's already omit empty so when we're looking at this output with TS_DEBUG_MAP=1 it'll be useful to see an explicit "Enabled: false" fly by on stderr.

tailcfg/tailcfg.go
```diff
@@ -1897,39 +1892,88 @@ type TKAInitFinishRequest struct {
...
+	Head string
+}
+
+// TKABootstrapResponse encodes values necessary to enable or disable
+// the tailnet key authority (TKA).
+type TKABootstrapResponse struct {
+	// GenesisAUM encodes the initial AUM necessary to initialize TKA.
+	GenesisAUM tkatype.MarshaledAUM `json:",omitempty"`
+
+	// DisablementSecret encodes a secret necessary to disable TKA.
```
bradfitz, line 1934: encodes in what form?

tailcfg/tailcfg.go
```diff
@@ -1897,39 +1892,88 @@ type TKAInitFinishRequest struct {
...
+	DisablementSecret []byte `json:",omitempty"`
+}
+
+// TKASyncOfferRequest encodes a request to synchronize tailnet key authority
+// state (TKA). Values of type tka.AUMHash are encoded as strings in their
+// MarshalText() form.
```
bradfitz, line 1940: drop parens

tailcfg/tailcfg.go
```diff
@@ -1897,39 +1892,88 @@ type TKAInitFinishRequest struct {
...
+type TKASyncOfferRequest struct {
+	// Head represents the nodes' head AUMHash (tka.Authority.Head()). This
+	// corresponds to tka.SyncOffer.Head.
+	Head string
+	// Ancestors represents selection of ancestor AUMHash values ascending
```
bradfitz, line 1945: the selection?

but I don't quite follow this so maybe I'm confused

tailcfg/tailcfg.go
```diff
@@ -1897,39 +1892,88 @@ type TKAInitFinishRequest struct {
...
+	// from the current head. This corresponds to tka.SyncOffer.Ancestors.
+	Ancestors []string
+}
+
+// TKASyncOfferResponse encodes a response in synchronizing a nodes'
```
bradfitz, line 1950: a node's

tailcfg/tailcfg.go
```diff
@@ -1897,39 +1892,88 @@ type TKAInitFinishRequest struct {
...
+// tailnet key authority state. Values of type tka.AUMHash are encoded as
+// strings in their MarshalText() form.
```
bradfitz, line 1952: parens

tailcfg/tailcfg.go
twitchyliquid64: Im bad at english :joy: :joy: I didnt know the rule for possession with trailing apostrophe required it to end in -s. I'll fix lol

tailcfg/tailcfg.go
twitchyliquid64: The type `tkatype.MarshaledAUM` is just a `[]byte`, so its the serialized AUM.

tailcfg/tailcfg.go
twitchyliquid64: A selection (2nd ancestor, 4th ancestor, 8th ancestor, 16th ancestor etc) of ancestor hashes are sent to help compute what we call in the design a 'tail intersection'. This is all a mirror of the fields of `tka.SyncOffer`.

tailcfg/tailcfg.go
```diff
@@ -1347,9 +1342,11 @@ type MapResponse struct {
 	// ControlTime, if non-zero, is the current timestamp according to the control server.
 	ControlTime *time.Time `json:",omitempty"`
 
-	// TKA, if non-nil, describes updates for the local instance of the
-	// tailnet key authority.
-	TKA *TKAMapResponse `json:",omitempty"`
+	// TKAInfo, if non-nil, describes the control plane's view of tailnet
+	// key authority state.
+	//
+	// A nil TKAInfo indicates no changes from before.
```
bradfitz, line 1348: You forgot half my request. What's an initial nil mean? Does it mean Enabled==false, Head==""?

tailcfg/tailcfg.go
```diff
@@ -1897,39 +1894,95 @@ type TKAInitFinishRequest struct {
...
-// to reach out to control separately to synchronize.
-type TKAMapResponse struct {
-	// Updates is any AUMs that control believes the node should apply.
-	Updates []tkatype.MarshaledAUM `json:",omitempty"`
-
-	// WantSync is set by control to request the node complete AUM
-	// synchronization.
+// A nil pointer to this struct indicates no changes since the last time
+// TKAInfo was communicated.
+type TKAInfo struct {
```
bradfitz, line 1902: want to put all these TKA types in a new file `tka.go` like we did for `tailcfg/derpmap.go`?

twitchyliquid64: Done.

tailcfg/tailcfg.go
twitchyliquid64: A nil `TKAInfo` means don't touch the local TKA state, rather than declaring that tka should or shouldnt be enabled or anything like that (that would mean that not sending TKAInfo because it hasnt changed has the side effect of reaching out to get the disablement secret, regardless of if its actually disabled).

Ive updated the comment.

tailcfg/tailcfg.go
```diff
@@ -1347,9 +1342,12 @@ type MapResponse struct {
 	// ControlTime, if non-zero, is the current timestamp according to the control server.
 	ControlTime *time.Time `json:",omitempty"`
 
-	// TKA, if non-nil, describes updates for the local instance of the
-	// tailnet key authority.
-	TKA *TKAMapResponse `json:",omitempty"`
+	// TKAInfo, if non-nil, describes the control plane's view of tailnet
+	// key authority (TKA) state.
+	//
+	// A nil TKAInfo indicates that the control plane has no information
```
bradfitz, line 1348: now you lost the "nil" means no change bit.

Each MapResponse field has two contexts: the initial response, and deltas. For deltas, they're nowadays all the same: nil means "no change from the previous streamed one". We still document that even though it's the same for all of them. But the initial value also should be documented.

"no information to influence" .... what if the node had NL already enabled and then reconnects and gets nil here? Is that a transition to disabled true?

twitchyliquid64: "nil means no change" and "nil means dont touch the local tka" are both true: Because the first time round with the non-nil TKAInfo would have resulted in a local state change. If that makes sense?

bradfitz: No, because these fields are all declarative. MapResponse only says the state of the world. The nil-ness is only a compression detail for bandwidth. It has no semantic difference.

Ignoring the compression, if we didn't do that, what does nil mean? It can't mean "don't touch" because it can't mean something imperative. 

Does a nil mean NL isn't enabled?

twitchyliquid64: Oh I see what youre trying to do. We could remove `TKAInfo.Enabled` and rely just on nil vs non-nil, and that would be fully declarative, but that has some side-effects we need to accept: A client with a local TKA state talking to a server which doesnt undestand these fields is going to result in an RPC to get the disablement secret (which will 404), because not supporting this field implies NL is disabled.

This is messier than the other fields because NL is fundamentally about not blindly trusting the declarations from control: It always needs to involve consideration of local state.

twitchyliquid64:
PTAL at the latest revision, which moves forward with the declarative approach. I think in this case the combination of overloading nil to mean different things for the initial response vs a delta response, coupled with being control declarative when in reality its the local state dictating how tka state is updated ends up being messier.

tailcfg/tailcfg.go
```diff
@@ -1347,9 +1342,15 @@ type MapResponse struct {
 	// ControlTime, if non-zero, is the current timestamp according to the control server.
 	ControlTime *time.Time `json:",omitempty"`
 
-	// TKA, if non-nil, describes updates for the local instance of the
-	// tailnet key authority.
-	TKA *TKAMapResponse `json:",omitempty"`
+	// TKAInfo, describes the control plane's view of tailnet
```
bradfitz, line 1345: drop this comma

tailcfg/tka.go
```diff
@@ -0,0 +1,142 @@
...
+	"tailscale.com/types/key"
+	"tailscale.com/types/tkatype"
+)
+
+// TKAInitBeginRequest submits a genesis AUM to seed the creation of the
+// tailnet's key authority.
+type TKAInitBeginRequest struct {
+	NodeID NodeID // NodeID of the initiating client
+
+	GenesisAUM tkatype.MarshaledAUM
```
bradfitz, line 17: docs?

tailcfg/tka.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// TKASignInfo describes information about an existing node that needs
+// to be signed into a node-key signature.
+type TKASignInfo struct {
+	NodeID     NodeID // NodeID of the node-key being signed
```
bradfitz, line 23: let's go with formal before-line comments on all these. they're important enough.

// NodeID is the node whose node key is requested to be signed. It must correspond to NodePublic.

or something

tailcfg/tka.go
```diff
@@ -0,0 +1,142 @@
...
+	NodePublic key.NodePublic
```
bradfitz, line 24: same. also some comments.

tailcfg/tka.go
```diff
@@ -0,0 +1,142 @@
...
+	// which it then signs with the key corresponding to RotationPubkey.
+	//
+	// This field expects a raw ed25519 public key.
+	RotationPubkey []byte
+}
+
+// TKAInitBeginResponse describes node information which must be signed to
+// complete initialization of the tailnets' key authority.
+type TKAInitBeginResponse struct {
+	NeedSignatures []TKASignInfo
```
bradfitz, line 41: // NeedSignatures are the ....

tailcfg/tka.go
```diff
@@ -0,0 +1,142 @@
...
+	// This is necessary so the node can rotate its node-key without
+	// talking to a node which holds a trusted network-lock key.
+	// It does this by nesting the original NKS in a 'rotation' NKS,
+	// which it then signs with the key corresponding to RotationPubkey.
+	//
+	// This field expects a raw ed25519 public key.
+	RotationPubkey []byte
+}
+
+// TKAInitBeginResponse describes node information which must be signed to
```
bradfitz, line 38: // TKAInitBeginResponse is the JSON response from the foo handler.

tailcfg/tka.go
```diff
@@ -0,0 +1,142 @@
...
+// complete initialization of the tailnets' key authority.
+type TKAInitBeginResponse struct {
+	NeedSignatures []TKASignInfo
+}
+
+// TKAInitFinishRequest finalizes initialization of the tailnet key authority
+// by submitting node-key signatures for all existing nodes.
+type TKAInitFinishRequest struct {
+	NodeID NodeID // NodeID of the initiating client
```
bradfitz, line 47: I don't follow this. This request is over noise and we know the client.

tailcfg/tka.go
```diff
@@ -0,0 +1,142 @@
...
+
+	Signatures map[NodeID]tkatype.MarshaledSignature
+}
+
+// TKAInitFinishResponse describes the successful enablement of the tailnet's
```
bradfitz, line 52: ... is the response from the foo handler.

tailcfg/tka.go
```diff
@@ -0,0 +1,142 @@
...
+// key authority.
+type TKAInitFinishResponse struct{}
```
bradfitz, line 54:
```
struct {
   // Nothing. (yet?)
}
```

tailcfg/tka.go
```diff
@@ -0,0 +1,142 @@
...
+	// Disabled indicates the control plane believes TKA should be disabled,
+	// and the node should reach out to fetch a disablement
+	// secret. If the disablement secret verifies, then the node should then
+	// disable TKA locally.
+	// This field exists to disambiguate a nil TKAInfo in a delta mapresponse
+	// from a nil TKAInfo indicating TKA should be disabled.
+	//
+	// TODO(tom): Implement /machine/tka/boostrap as a noise endpoint, to
+	// communicate the genesis AUM & any disablement secrets.
+	Disabled bool
```
bradfitz, line 78: now that it's reversed meaning and this will almost always be false, we can add omitempty back

bradfitz (approved):
But LGTM otherwise.

tailcfg/tka.go
twitchyliquid64: My understanding is that the noise session is authenticated with the machine key, which may correspond to multiple nodes (ie: Windows). So the initiator needs to signal which node they are.

I saw this pattern in other RPCs that are done to control: If there is another way please link me because I would prefer an alternative lol

bradfitz: Oh, no, you're right. The Noise RPCs are by machine key.

tailcfg/tka.go
```diff
@@ -0,0 +1,161 @@
...
+
+import (
+	"tailscale.com/types/key"
+	"tailscale.com/types/tkatype"
+)
+
+// TKAInitBeginRequest submits a genesis AUM to seed the creation of the
+// tailnet's key authority.
+type TKAInitBeginRequest struct {
+	// NodeID of the initiating client.
```
bradfitz, line 15: complete sentences:

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

This sentence lacks a verb. Maybe:

```
// NodeID is the node of the initiating client.
// It must match the machine key being used to communicate over noise.
```

tailcfg/tka.go
```diff
@@ -0,0 +1,161 @@
...
+	NodeID NodeID
+
+	// The initial (genesis) AUM that the node generated to bootstrap
```
bradfitz, line 18: per https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences start with the name of the field

// GenesisAUM is ...

tailcfg/tka.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// TKASyncOfferResponse encodes a response in synchronizing a node's
+// tailnet key authority state. Values of type tka.AUMHash are encoded as
+// strings in their MarshalText form.
+type TKASyncOfferResponse struct {
+	// Head represents the control plane's head AUMHash (tka.Authority.Head).
+	// This corresponds to tka.SyncOffer.Head.
+	Head string
+	// Ancestors represents a selection of ancestor AUMHash values ascending
```
bradfitz, line 137: blank line before?

PR #5490, opened 2022-08-29
wgengine/magicsock: use AF_PACKET socket + BPF to read disco messages

danderson:
This is entirely optional (i.e. failing in this code is non-fatal) and
only enabled on Linux for now. Additionally, this new behaviour can be
disabled by setting the TS_DEBUG_DISABLE_AF_PACKET environment variable.

Updates #3824
Replaces #5474

Co-authored-by: Andrew Dunham <andrew@du.nham.ca>


wgengine/magicsock/magicsock.go
```diff
@@ -1665,12 +1683,12 @@ func (c *Conn) receiveIPv4(b []byte) (n int, ep conn.Endpoint, err error) {
...
 // ok is whether this read should be reported up to wireguard-go (our
 // caller).
-func (c *Conn) receiveIP(b []byte, ipp netip.AddrPort, cache *ippEndpointCache) (ep *endpoint, ok bool) {
+func (c *Conn) receiveIP(b []byte, ipp netip.AddrPort, cache *ippEndpointCache, checkDisco bool) (ep *endpoint, ok bool) {
 	if stun.Is(b) {
 		c.stunReceiveFunc.Load()(b, ipp)
 		return nil, false
 	}
-	if c.handleDiscoMessage(b, ipp, key.NodePublic{}) {
+	if checkDisco && c.handleDiscoMessage(b, ipp, key.NodePublic{}) {
```
bradfitz, line 1691: You're assuming that just because we listened that the BPF is working? I'd want to send a test packet to ourselves and test that we get it first before turning off our UDP disco processing here.

Also, don't we still want to look to see if it's disco here and discard it rather than letting it fall through to wireguard?

wgengine/magicsock/magicsock.go
```diff
@@ -4192,4 +4216,10 @@ var (
 	// metricDERPHomeChange is how many times our DERP home region DI has
 	// changed from non-zero to a different non-zero.
 	metricDERPHomeChange = clientmetric.NewCounter("derp_home_change")
+
+	// Disco packets received through AF_PACKET read path
+	metricRecvDiscoPacketIPv4        = clientmetric.NewCounter("magicsock_disco_recv_packet_ipv4")
```
bradfitz, line 4221: these counter metric names don't really say that they're AF_PACKET/bpf'ed

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,233 @@
...
+	debugDisableAF_PACKET = envknob.Bool("TS_DEBUG_DISABLE_AF_PACKET")
+)
+
+// Convert our disco magic number into a uint32 and uint16 to test
+// against. We panic on an incorrect length here rather than try to be
+// generic with our BPF instructions below.
+//
+// Note that BPF uses network byte order (big-endian) when loading data
+// from a packet, so that is what we use to generate our magic numbers.
+var magic1, magic2 = (func() (uint32, uint16) {
```
bradfitz, line 43: two consts or two vars without the func literal at least? no need to check the disco length

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,233 @@
...
+var (
+	magicsockFilterV4 = []bpf.Instruction{
+		// Protocol is guaranteed to be UDP because we told
+		// net.ListenPacket we wanted a UDP tap.
+
+		// Disco packets are so small they should never get
+		// fragmented, and we don't want to handle reassembly.
+		bpf.LoadAbsolute{Off: 6, Size: 2},
+		bpf.JumpIf{Cond: bpf.JumpBitsSet, Val: 0x2000, SkipTrue: 7, SkipFalse: 0},
+		bpf.JumpIf{Cond: bpf.JumpBitsSet, Val: 0x1fff, SkipTrue: 6, SkipFalse: 0},
```
andrew-d, line 62: Maybe worth a comment on what these two tests are checking for?

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,233 @@
...
+
+	go c.receiveDisco(pc)
+	return pc, nil
+}
+
+func (c *Conn) receiveDisco(pc net.PacketConn) {
+	var buf [1500]byte
+	for {
+		n, src, err := pc.ReadFrom(buf[:])
+		if err == net.ErrClosed {
```
andrew-d, line 163: `errors.Is(err, net.ErrClosed)` probably, per the `ErrClosed` godoc

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,233 @@
...
+	"tailscale.com/disco"
+	"tailscale.com/envknob"
+	"tailscale.com/types/ipproto"
+	"tailscale.com/types/key"
+	"tailscale.com/util/clientmetric"
+)
+
+const (
+	udpHeaderSize          = 8
+	ipv6HeaderSize         = 40
```
andrew-d, line 26: I just discovered this now; `ipv6.HeaderLen` is a constant

wgengine/magicsock/magicsock_linux.go
```diff
@@ -0,0 +1,224 @@
...
+var (
+	magicsockFilterV4 = []bpf.Instruction{
+		// Protocol is guaranteed to be UDP because we told
+		// net.ListenPacket we wanted a UDP tap.
+
+		// Disco packets are so small they should never get
+		// fragmented, and we don't want to handle reassembly.
+		bpf.LoadAbsolute{Off: 6, Size: 2},
+		bpf.JumpIf{Cond: bpf.JumpBitsSet, Val: 0x2000, SkipTrue: 7, SkipFalse: 0},
+		bpf.JumpIf{Cond: bpf.JumpBitsSet, Val: 0x1fff, SkipTrue: 6, SkipFalse: 0},
```
andrew-d, line 53: Maybe worth a comment on what these two tests are checking for?

andrew-d (commented):
Initial read-through review

wgengine/magicsock/magicsock.go
danderson: The latter: yes, noticed that and fixed.

The former: no, I don't think it's necessary to self-test. The combination of ListenPacket and SetBPF succeeding is sufficient, IMO.

wgengine/magicsock/magicsock_linux.go
danderson: Replaced with a const magic and a test to enforce that it matches the disco magic in the disco packet.

wgengine/magicsock/magicsock_linux.go
danderson: Done.

wgengine/magicsock/magicsock.go
danderson: Updated to explicitly say "bpf". Also removed the "invalid" counters, since we already track bad disco packets in more detail in the main handler.

wgengine/magicsock/magicsock.go
```diff
@@ -4192,4 +4232,8 @@ var (
 	// metricDERPHomeChange is how many times our DERP home region DI has
 	// changed from non-zero to a different non-zero.
 	metricDERPHomeChange = clientmetric.NewCounter("derp_home_change")
+
+	// Disco packets received through AF_PACKET read path
```
andrew-d, line 4236:
```suggestion
	// Disco packets received through raw socket read path
```

andrew-d (approved):
🎉

wgengine/magicsock/magicsock.go
danderson: I repent: I implemented the self-test because it was easy enough to do, and it revealed that the IPv6 BPF program is broken and doesn't appear to receive disco traffic. So, um, debugging that now.

bradfitz: hah, lolsobyay?

danderson: All glory to @andrew-d for figuring it out: when you create a raw UDP socket, the bytes the BPF program receives as input vary based on address family: for a UDPv4 raw socket, it's the entire IP+UDP+payload. For UDPv6, it's just UDP+payload. So obviously, our programs that assumed IP+UDP+payload didn't work.

Cursed, but true. And now fixed. On the plus side, I have empirical proof that the self-test and fallback work correctly :zany_face:

andrew-d (approved):
The different BPF behaviour is deeply cursed, but this does appear to work 😵‍💫

danderson:
Waiting on 1.30 to merge this, because I'm leery of changing magicsock's NAT traversal behavior on the eve of release. Also, this code as-is generates a fair bit more noise in TS_DEBUG_DISCO when multiple tailscaled instances are running per machine, since they now all see each others' disco packets. I would like to implement a read-time filter for that (not in BPF, just in userspace), but when I tried doing that it transpired that magicsock's RebindingUDPConn always gave me port 0 as the listening port, instead of the "effective" bound port... So that's something else to debug tomorrow.

bradfitz (approved):

bradfitz:
> when multiple tailscaled instances are running per machine

I wouldn't worry about that for the first version. Even with a dozen tsnet instances on a single machine, disco traffic isn't super busy.

If you wanted to be super fancy you could probably use a bpf map to filter on the set of `senderDiscoPub [32]byte` (which follows the disco magic) you care about, keeping that map updated as the netmap changes.

danderson:
Classic BPF doesn't have access to maps, so to do that we'd have to upgrade to full fat eBPF, which involves pulling in a new compiler toolchain and C headers and eugh :(

My worry is not so much traffic load, but confusion in debugging. While building this I confused myself multiple times into thinking the receiver was broken, when magicsock was in fact complaining about the disco traffic of a neighbor tailscaled.

I won't spend much time trying to get it working for the first round, but it feels like I've stumbled on a different bug somewhere that's not yielding the right port numbers...

PR #5494, opened 2022-08-30
cmd/tsconnect: output errors to the JS console too

mihaip:
We were just outputting them to the terminal, but that's hard to debug
because we immediately tear down the terminal when getting an error.

### DIFF
--- a/cmd/tsconnect/src/lib/ssh.ts
+++ b/cmd/tsconnect/src/lib/ssh.ts
@@ -27,26 +27,39 @@ export function runSSHSession(
 
   term.focus()
 
-  const sshSession = ipn.ssh(def.hostname, def.username, {
-    writeFn: (input) => term.write(input),
-    setReadFn: (hook) => (onDataHook = hook),
+  let resizeObserver: ResizeObserver | undefined
+  let handleBeforeUnload: ((e: BeforeUnloadEvent) => void) | undefined
+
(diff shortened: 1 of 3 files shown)


cmd/tsconnect/src/lib/ssh.ts
```diff
@@ -27,26 +27,39 @@ export function runSSHSession(
 
   term.focus()
 
-  const sshSession = ipn.ssh(def.hostname, def.username, {
-    writeFn: (input) => term.write(input),
-    setReadFn: (hook) => (onDataHook = hook),
+  let resizeObserver: ResizeObserver | undefined
```
mihaip, line 30: It was necessary to add declarations for these earlier than their use below because `onDone` can be invoked synchronously, before they're initialized.

cmd/tsconnect/src/lib/ssh.ts
```diff
@@ -27,26 +27,39 @@ export function runSSHSession(
...
+  let handleBeforeUnload: ((e: BeforeUnloadEvent) => void) | undefined
+
+  const sshSession = ipn.ssh(def.hostname + "2", def.username, {
+    writeFn(input) {
+      term.write(input)
+    },
```
catzkorn, line 36: Why has the structure of this changed in comparison to the existing implementation?

catzkorn (approved):

Issue #1543, opened 2021-03-18
Renaming MagicDNS URLs to. machine name only.


darshinimashar:
Also, for DNS, the Customer would really like to make DNS aliases such as **example.ourdomain.beta.tailscale.net** always point to a machine name **staging-1** or something like that. 

This way, they can easily access the machines.

guidoiaquinti:
I would love to see as well support for an internal `.tailscale` domain name (example: `nas.tailscale`, `pi.tailscale`) and maybe even subzones (example: `nas.home.tailscale` VS `nas.office.tailscale`) without the need of a dedicated/external DNS server.

MingSpace:
This feature will easily customize the DNS of the home server , just use tailscale 😀

> I would love to see as well support for an internal `.tailscale` domain name (example: `nas.tailscale`, `pi.tailscale`) and maybe even subzones (example: `nas.home.tailscale` VS `nas.office.tailscale`) without the need of a dedicated/external DNS server.

bradfitz:
0debb99f0852e8fa53ab6c725966f8dd99b3fbc9 added client support for a new ExtraRecords field. It's not yet supported server-side.

awoimbee:
Sorry to be this person, but: Any news on this @bradfitz ? It's awkward that it has been partially implemented for 7 months.
This would simplify the process described here https://github.com/tailscale/tailscale/issues/1748#issuecomment-824121315 quite a bit by removing the need to host a DNS server.

bradfitz:
> It's awkward that it has been partially implemented for 7 months.

Well, it was "partially implemented" because we needed the part we implemented (and now use) for the LetsEncrypt support.

Exposing more of it for this has a ton of product & security considerations and interacts often surprisingly with various DNS stacks on different platforms, which will increase our support burden if we ship something that seems to work sometimes but not on all platforms in all configs.

Example of problems:

* if you add a `foo.net` record, we need to make sure your OS queries the 100.100.100.100 resolver for that record. But some OSes only let us route `*.foo.net` to us, which means what do we do with `bar.foo.net`? We'd have to forward that on upstream, but some OSes don't make it easy or possible for us to figure out what the system's underlying/native DNS servers are. So then we'd require you specify an upstream DNS server in your Tailnet config, which we're trying to move away from so we can enable MagicDNS on by default.
[...]

PR #5497, opened 2022-08-30
net/dns: set OS DNS to 100.100.100.100 for route-less ExtraRecords [cap 41]

bradfitz:
If ExtraRecords (Hosts) are specified without a corresponding split
DNS route and global DNS is specified, then program the host OS DNS to
use 100.100.100.100 so it can blend in those ExtraRecords.

Updates #1543

### DIFF
--- a/net/dns/config.go
+++ b/net/dns/config.go
@@ -91,6 +91,30 @@ func (c Config) hasDefaultIPResolversOnly() bool {
 	return true
 }
 
+// hasHostsWithoutSplitDNSRoutes reports whether c contains any Host entries
+// that aren't covered by a SplitDNS route suffix.
+func (c Config) hasHostsWithoutSplitDNSRoutes() bool {
+	// TODO(bradfitz): this could be more efficient, but we imagine
+	// the number of SplitDNS routes and/or hosts will be small.
+	for host := range c.Hosts {
+		if !c.hasSplitDNSRouteForHost(host) {
+			return true
+		}
+	}
+	return false
+}
+
+// hasSplitDNSRouteForHost reports whether c contains a SplitDNS route
+// that contains hosts.
+func (c Config) hasSplitDNSRouteForHost(host dnsname.FQDN) bool {
+	for route := range c.Routes {
+		if route.Contains(host) {
+			return true
+		}
+	}
+	return false
+}
+
 func (c Config) hasDefaultResolvers() bool {
 	return len(c.DefaultResolvers) > 0
 }
--- a/net/dns/manager.go
+++ b/net/dns/manager.go
@@ -207,9 +207,14 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
 		// case where cfg is entirely zero, in which case these
 		// configs clear all Tailscale DNS settings.
 		return rcfg, ocfg, nil
-	case cfg.hasDefaultIPResolversOnly():
-		// Trivial CorpDNS configuration, just override the OS
-		// resolver.
+	case cfg.hasDefaultIPResolversOnly() && !cfg.hasHostsWithoutSplitDNSRoutes():
+		// Trivial CorpDNS configuration, just override the OS resolver.
+		//
+		// If there are hosts (ExtraRecords) that are not covered by an existing
+		// SplitDNS route, then we don't go into this path so that we fall into
+		// the next case and send the extra record hosts queries through
+		// 100.100.100.100 instead where we can answer them.
+		//
 		// TODO: for OSes that support it, pass IP:port and DoH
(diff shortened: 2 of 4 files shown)


net/dns/config.go
```diff
@@ -91,6 +91,24 @@ func (c Config) hasDefaultIPResolversOnly() bool {
 	return true
 }
 
+func (c Config) hasHostsWithoutSplitDNSRoutes() bool {
```
crawshaw, line 94: So in the case where no extra dns records are set, this returns false? (Which means we follow current expected paths.)

danderson:
This won't work on all platforms, as documented at https://docs.google.com/document/d/11CvDEvT9QOE6p7Qo9k_wiXudqe-IHhcM0UgYAriYqyw/edit#heading=h.p5dwwtdrgwq0 . What's the plan to make this actually work everywhere?

danderson (changes requested):
NACK, because this will break DNS on at least macOS and iOS in some configurations.

crawshaw:
I don't follow. This sets the name server to 100.100.100.100 in that case, so we can respond to these names.

maisem:
Doesn't work on macOS, I tested this last week. Happy to context dump.

net/dns/config.go
bradfitz: yes. No change unless this returns true.

bradfitz:
> Doesn't work on macOS, I tested this last week. Happy to context dump.

I talked to @maisem now. He tested something else and didn't understand the background on this change.

bradfitz:
> NACK, because this will break DNS on at least macOS and iOS in some configurations.

I don't think so. This is only when override local DNS is set so we have fallback resolvers for quad100 to use.

bradfitz:
@maisem, you say this works on macOS? also with MagicDNS enabled?

net/dns/config.go
```diff
@@ -91,6 +91,24 @@ func (c Config) hasDefaultIPResolversOnly() bool {
...
+	for host := range c.Hosts {
```
danderson, line 95: So this is O(N^2) in hosts x routes. Probably okay?

net/dns/manager.go
```diff
@@ -207,7 +207,7 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
 		// case where cfg is entirely zero, in which case these
 		// configs clear all Tailscale DNS settings.
 		return rcfg, ocfg, nil
-	case cfg.hasDefaultIPResolversOnly():
+	case cfg.hasDefaultIPResolversOnly() && !cfg.hasHostsWithoutSplitDNSRoutes():
 		// Trivial CorpDNS configuration, just override the OS
```
danderson, line 211: Comment needs updating, this codepath is no longer just trivial CorpDNS config.

danderson (approved):

maisem:
> @maisem, you say this works on macOS? also with MagicDNS enabled?

Yeah with MagicDNS enabled and `Override Local DNS` set

net/dns/config.go
```diff
@@ -91,6 +91,24 @@ func (c Config) hasDefaultIPResolversOnly() bool {
...
+		if !c.hasSplitDNSRouteForHost(host) {
+			return true
+		}
+	}
+	return false
+}
+
+func (c Config) hasSplitDNSRouteForHost(host dnsname.FQDN) bool {
```
maisem, line 103: docs pls

net/dns/config.go
maisem, line 94: docs

net/dns/manager.go
```diff
@@ -207,7 +207,7 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
 		// case where cfg is entirely zero, in which case these
 		// configs clear all Tailscale DNS settings.
 		return rcfg, ocfg, nil
-	case cfg.hasDefaultIPResolversOnly():
+	case cfg.hasDefaultIPResolversOnly() && !cfg.hasHostsWithoutSplitDNSRoutes():
```
maisem, line 210: comment here describing why?

maisem (approved):

net/dns/manager.go
bradfitz: I dunno, the next case's comment basically already says it:

```
	// Default resolvers plus other stuff always ends up proxying
	// through quad-100.
```

"plus other stuff"

net/dns/config.go
bradfitz: commented

DentonGentry (approved):
Fine to submit in time for 1.30 if that is what we decide to do.

PR #5499, opened 2022-08-30
ipn/localapi: send Tailscale version in ACME User-Agent

andrew-d:
Requested by a friend at Let's Encrypt.

cc @mcpherrinm

### DIFF
--- a/ipn/localapi/cert.go
+++ b/ipn/localapi/cert.go
@@ -38,6 +38,7 @@ import (
 	"tailscale.com/envknob"
 	"tailscale.com/ipn/ipnstate"
 	"tailscale.com/types/logger"
+	"tailscale.com/version"
 	"tailscale.com/version/distro"
 )
 
@@ -185,7 +186,10 @@ func (h *Handler) getCertPEM(ctx context.Context, logf logger.Logf, traceACME fu
 	if err != nil {
 		return nil, fmt.Errorf("acmeKey: %w", err)
 	}
-	ac := &acme.Client{Key: key}
+	ac := &acme.Client{
+		Key:       key,
+		UserAgent: "tailscaled/" + version.Long,
+	}
 
 	a, err := ac.GetReg(ctx, "" /* pre-RFC param */)
 	switch {
--- a/version/version.go
+++ b/version/version.go
@@ -23,7 +23,8 @@ var Long = ""
 
 // Short is a short version number for this build, of the form
 // "x.y.z" for builds stamped in the usual way (see
-// build_dist.sh in the root) or, for binaries built by hand with the                                                                            // go tool, it's like Long's dev form, but ending at the date part,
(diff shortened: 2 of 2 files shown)


ipn/localapi/cert.go
```diff
@@ -185,7 +186,10 @@ func (h *Handler) getCertPEM(ctx context.Context, logf logger.Logf, traceACME fu
 	if err != nil {
 		return nil, fmt.Errorf("acmeKey: %w", err)
 	}
-	ac := &acme.Client{Key: key}
+	ac := &acme.Client{
+		Key:       key,
+		UserAgent: "tailscaled " + version.String(),
```
bradfitz, line 191: This String isn't what you think it is

version/version.go
```diff
@@ -23,7 +23,8 @@ var Long = ""
 
 // Short is a short version number for this build, of the form
 // "x.y.z" for builds stamped in the usual way (see
-// build_dist.sh in the root) or, for binaries built by hand with the                                                                            // go tool, it's like Long's dev form, but ending at the date part,
+// build_dist.sh in the root) or, for binaries built by hand with the
```
danderson, line 26: Spurious change? I can't tell what's different.

danderson (approved):

DentonGentry (approved):
Fine to submit in time for 1.30.

version/version.go
mcpherrinm: The "split" github view hides it with line wrapping, but there's a missing newline here being inserted.  More obvious if you look at the current file in main: https://github.com/tailscale/tailscale/blob/27f36f77c364562bc1ca660c241291e23606e54d/version/version.go#L26

ipn/localapi/cert.go
```diff
@@ -185,7 +186,10 @@ func (h *Handler) getCertPEM(ctx context.Context, logf logger.Logf, traceACME fu
 	if err != nil {
 		return nil, fmt.Errorf("acmeKey: %w", err)
 	}
-	ac := &acme.Client{Key: key}
+	ac := &acme.Client{
+		Key:       key,
+		UserAgent: "tailscaled " + version.Long,
```
bradfitz, line 191: don't you want a slash instead of a space? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

bradfitz (approved):

ipn/localapi/cert.go
mcpherrinm: A slash would be consistent with the standard use of the User Agent header and other ACME clients, though from Let's Encrypt's perspective we don't care too much about the separator.

andrew-d: Good call; switched to a slash.

bradfitz (approved):

Issue #5425, opened 2022-08-24
userspace-networking as exit-node without IPv6 connectivity breaks Chrome 104+


KayLEvans:
### What is the issue?

Unable to browse websites through exit node on google chrome. Chrome returns a ERR_CONNECTION_RESET error. Firefox and Safari works fine. I ran a wireshark capture on the tailscale vpn tunnel and it appears that chrome is trying to access the ipv6 address of the website however the exit node's ISP does not have any ipv6 capability, which I think is the cause of this error. If chrome has previously cached the ipv4 address of that website then it will work fine, but if you try to use incognito mode or guest mode in chrome then the connection reset error appears.

### Steps to reproduce

1. Advertise exit node on opnsense router.
2. Enable exit node in admin panel.
3. Activate exit node on tailscale client.
4. Attempt to browse websites using exit node. Chrome returns ERR_CONNECTION_RESET error. When exit node is disabled, browsing returns to normal as expected.
5. Use a different web browser and browsing works fine.

### OS

macOS

### OS version

MacOS Monterey 12.5

### Tailscale version

1.28.0

### Bug report

BUG-ada8048276570b184149432e6f7106aa37c0a258779dd63085611b52d9f593f8-20220813134359Z-2b4b0289c9e2c92b

mclarence:
> Thanks for letting us know! Would you happen to know the ISP?

The ISP is TPG Australia.

I would like to add that I have replicated it on a windows machine as well. I have also used a different device (synology NAS) as an exit node and the same issue still persists.

DentonGentry:
Interesting. We know of cases where the opposite happens, that the IPv6 ULA that Tailscale uses discourages Chrome from using IPv6: https://github.com/tailscale/tailscale/issues/2299

ookees:
This is similar to what is happening to me as of Chrome 104. It happens on all chromium based browsers as of 104. Anything before 104 works. Safari works fine, Firefox also works but only if I turn off DNS over HTTPS. Turning off secure DNS in chrome doesn't fix it. This happens on both Mac and Android 13.

My exit node is a Mac running Monterey.

willnorris:
For folks seeing this issue, can you confirm whether you have IPv6 connectivity on your network (specifically from your exit node)?  I'm guessing that you don't.  You can test at https://test-ipv6.com/

ookees:
Tailscale's response to my support email:

[...]

PR #5503, opened 2022-08-30
wgengine/netstack: only accept connection after dialing

andrew-d:
If we accept a forwarded TCP connection before dialing, we can
erroneously signal to a client that we support IPv6 (or IPv4) without
that actually being possible. Instead, we only complete the client's TCP
handshake after we've dialed the outbound connection; if that fails, we
respond with a RST.

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -743,47 +743,64 @@ func (ns *Impl) acceptTCP(r *tcp.ForwarderRequest) {
 			ns.removeSubnetAddress(dialIP)
 		}
 	}()
+
 	var wq waiter.Queue
-	ep, err := r.CreateEndpoint(&wq)
-	if err != nil {
-		ns.logf("CreateEndpoint error for %s: %v", stringifyTEI(reqDetails), err)
-		r.Complete(true) // sends a RST
-		return
+
+	// We can't actually create the endpoint or complete the inbound
+	// request until we're sure that the connection can be handled by this
+	// endpoint. This function sets up the TCP connection and should be
+	// called immediately before a connection is handled.
+	createConn := func() *gonet.TCPConn {
+		ep, err := r.CreateEndpoint(&wq)
+		if err != nil {
+			ns.logf("CreateEndpoint error for %s: %v", stringifyTEI(reqDetails), err)
+			r.Complete(true) // sends a RST
+			return nil
+		}
+		r.Complete(false)
+
+		// SetKeepAlive so that idle connections to peers that have forgotten about
+		// the connection or gone completely offline eventually time out.
+		// Applications might be setting this on a forwarded connection, but from
+		// userspace we can not see those, so the best we can do is to always
+		// perform them with conservative timing.
+		// TODO(tailscale/tailscale#4522): Netstack defaults match the Linux
+		// defaults, and results in a little over two hours before the socket would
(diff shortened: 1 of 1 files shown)


wgengine/netstack/netstack.go
```diff
@@ -743,86 +743,119 @@ func (ns *Impl) acceptTCP(r *tcp.ForwarderRequest) {
...
-	ns.forwardTCP(c, clientRemoteIP, &wq, dialAddr)
+
+	if !ns.forwardTCP(createConn, clientRemoteIP, &wq, dialAddr) {
+		r.Complete(true) // sends a RST
+	}
 }
 
-func (ns *Impl) forwardTCP(client *gonet.TCPConn, clientRemoteIP netip.Addr, wq *waiter.Queue, dialAddr netip.AddrPort) {
-	defer client.Close()
+func (ns *Impl) forwardTCP(getClient func() *gonet.TCPConn, clientRemoteIP netip.Addr, wq *waiter.Queue, dialAddr netip.AddrPort) (handled bool) {
```
bradfitz, line 850: maybe worth making a named type for that func. quite the signature.

wgengine/netstack/netstack.go
```diff
@@ -840,13 +873,29 @@ func (ns *Impl) forwardTCP(client *gonet.TCPConn, clientRemoteIP netip.Addr, wq
...
+	// respond to the client with a RST. Either way, the caller no longer
+	// needs to clean up the client connection.
+	handled = true
+
+	// We dialed the connection; we can complete the client's TCP handshake.
+	client := getClient()
+	if client == nil {
+		return
+	}
+	defer client.Close()
```
bradfitz, line 897: where was this before?

bradfitz (approved):
LGTM assuming it works!

crawshaw (approved):
Conceptually looks good! Not an easy thing to write a test for. :)

wgengine/netstack/netstack.go
andrew-d: At the top of `forwardTCP`

wgengine/netstack/netstack.go
andrew-d: Something like `type GetClientFunc func() *gonet.TCPConn`? Or do you mean passing all the arguments (`getClient`, `clientRemoteIP`, etc.) as a struct?

bradfitz: oh, nevermind, I thought all that clientRemoteIP etc were part of the signature. nevermind me.

carry on

willnorris (approved):
tested and confirmed that this resolved the issue on my Synology running DSM 7.

bradfitz:
But does actual IPv6 still work? 😜

mihaip:
Should this reference #5425?

bradfitz:
> Should this reference #5425?

I often write footer lines like `Updates #nnn (maybe fixes!)` so it's not recognized by GitHub regexps and auto-closed, lest it doesn't fix it.

DentonGentry:
This would be fine to go in now, `release-branch/1.30` was created this morning.

andrew-d:
I launched two servers in Digital Ocean, with IPv6 enabled on both, and confirmed that IPv6 access through the exit node works fine (using `curl -6 https://ifconfig.co`). Given that, and with @willnorris reporting that it fixes things on his Synology, I'm going to merge this now.

PR #5491, opened 2022-08-29
util/uniq: use generics instead of reflect

andrew-d:
This takes 75% less time per operation per some benchmarks on my mac.

Signed-off-by: Andrew Dunham <andrew@du.nham.ca>

PR #5504, opened 2022-08-30
util/uniq: add ModifySliceFunc

andrew-d:
Follow-up to #5491. This was used in control... oops!


phirework (approved):

bradfitz (approved):
It is used? I didn't find a case earlier.

phirework:
cfgdb and adminhttp have a handful of calls that compare based on a property of the element, e.g. https://github.com/tailscale/corp/blob/main/control/cfgdb/cfgdb.go#L5352

PR #5505, opened 2022-08-31
cmd/tsconnect: add dev-pkg command for two-sided development

mihaip:
Allows imports of the NPM package added by 1a093ef4822b973ec86d481924690349eddba5cb
to be replaced with import("http://localhost:9090/pkg/pkg.js"), so that
changes can be made in parallel to both the module and code that uses
it (without any need for NPM publishing or even building of the package).

Updates #5415


catzkorn (approved):
Tried this out locally, very handy!

PR #5506, opened 2022-08-31
words: this title should have been a pun, but I chickened out

catzkorn:

### DIFF
--- a/words/scales.txt
+++ b/words/scales.txt
@@ -113,6 +113,11 @@ bee
 bearded
 beardie
 pogona
+chicken
+hen
+rooster
+quail
+grouse
 
 # Musical scales
 acoustic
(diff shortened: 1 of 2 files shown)


apenwarr:
Should we do hen and rooster while we’re here?

words/scales.txt
```diff
@@ -113,6 +113,11 @@ bee
 bearded
 beardie
 pogona
+chicken
```
danderson, line 116: These creatures have scales? Wha?

catzkorn: On their legs! I fact checked

catzkorn: > Scaled Skin: This covers the legs and feet and is thought to be vestigial from the dinosaur age. It also provides protection for the legs.

sailorfrag (approved):
How fowl

crawshaw (approved):
looking forward to bearded-chicken

PR #5507, opened 2022-08-31
cmd/nginx-auth: ability to check application capabilities

madjam002:
A header `Expected-Cap` can now be passed which can be used to assert that `WhoIsResponse.Caps` contains the provided capability.

This allows for ACL rules to define capabilities that can be checked here in cmd/nginx-auth, meaning that ACLs can be used to control access to various web services at Layer 7.

Note that as the Tailscale control plane doesn't yet support `action: "grant"` ACL rules yet (as I guess it's still WIP as per https://github.com/tailscale/tailscale/issues/4217), this was tested using Headscale and a patch to allow Application Capabilities.


madjam002:
I am using this with Kubernetes and the NGINX Ingress Controller. With this PR it's now possible to attach annotations to an Ingress resource like so:

```
nginx.ingress.kubernetes.io/auth-url: "http://unix:/var/run/tailscale/tailscale.nginx-auth.sock"
nginx.ingress.kubernetes.io/auth-snippet: |
  proxy_set_header Remote-Addr $remote_addr;
  proxy_set_header Remote-Port $remote_port;
  proxy_set_header Expected-Cap "nginx.ingress.kubernetes.io/allow/grafana.mycorp.internal";
```

And then define a corresponding Tailscale ACL rule to allow access:
```
{
  "action": "grant",
  "src": ["group:allowed_grafana"],
  "dst": ["tag:ingress_controller"],
  "cap": ["nginx.ingress.kubernetes.io/allow/grafana.mycorp.internal"]
}
```

cmd/nginx-auth/nginx-auth.go
```diff
@@ -89,6 +90,20 @@ func main() {
...
+		if expectedCap := r.Header.Get("Expected-Cap"); expectedCap != "" {
+			if info.Caps == nil {
+				w.WriteHeader(http.StatusForbidden)
+				log.Printf("user does not have any caps, wanted: %s", url.QueryEscape(expectedCap))
+				return
+			}
+
+			if !slices.Contains(info.Caps, expectedCap) {
+				w.WriteHeader(http.StatusForbidden)
+				log.Printf("user is missing expected cap, has: %s, wanted: %s", strings.Join(info.Caps[:], ","), url.QueryEscape(expectedCap))
```
willnorris, line 102: any reason this isn't just `strings.Join(info.Caps, ",")` ?

willnorris (approved):
As you note, it looks like the exact feature and format of capabilities is still being worked out.  But since it's already exposed in the WhoisResponse (and because this PR doesn't make any particular assumption about the format), I don't necessarily see any issue with passing it through to applications.

willnorris:
Actually, thinking about this a bit more, I do worry a bit about the confusion of adding this into nginx-auth before it's actually supported in Tailscale.

madjam002:
> Actually, thinking about this a bit more, I do worry a bit about the confusion of adding this into nginx-auth before it's actually supported in Tailscale.

Yeah that's fair enough, I wanted to just open this PR to suggest the idea and provide a working implementation!

willnorris:
Removing myself as a reviewer here and leaving a note for future implementors or reviewers... The syntax of capabilities has changed since the WIP this PR was built against.  We now have a map of application IDs to a basically opaque slice of "stuff".  That "stuff" is application-dependent, so it will require some careful thought about how to expose that to applications sitting behind nginx-auth, or how to build authorization rules inside nginx-auth itself.

willnorris (changes requested):
looks like because I previously approved this, that sticks even though I've removed myself as a reviewer.  Marking "request changes" just to indicate that this will need to be rewritten to use the new capability map.

PR #5511, opened 2022-08-31
ci(lint): add shell linter - Differential ShellCheck

jamacku:
Differential ShellCheck is a GitHub action that performs differential ShellCheck scans on shell scripts changed via PR and reports results directly in PR. The beauty of differential scans is that you will get reports only about newly added defects. 

Since this repository has few shell scripts, I think you might find some value in having a shell linter. Differential ShellCheck action is able to produce reports in SARIF format. GitHub understands this format and is able to display it nicely as a PR comment, and on the `Files Changed` tab, please see below.

image

image

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -1,0 +1,26 @@
+name: Differential ShellCheck
+on:
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    permissions:
+      security-events: write
+      pull-requests: write
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/differential-shellcheck.yml
```diff
@@ -0,0 +1,26 @@
...
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    permissions:
+      security-events: write
+      pull-requests: write
```
bradfitz, line 15: why?

.github/workflows/differential-shellcheck.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
bradfitz, line 26: why? seems shady.

.github/workflows/differential-shellcheck.yml
jamacku: [Permissions array](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) explicitly specify which permissions will have `GITHUB_TOKEN`.

`security-events: write`: allows to upload SARIF report to GitHub for further processing.
`pull-requests: write`: allows to GitHub bot to comment on PR (please see images above).

.github/workflows/differential-shellcheck.yml
jamacku: Why shady? [Token](https://github.com/jamacku/differential-shellcheck#token) is used for uploading SARIF to GitHub. The thing with `GITHUB_TOKEN` is that some actions use it as a default value, so you don't even know that action uses it. That seems a bit shady to me. But I understand your concerns.

Feel free to inspect [sources of action](https://github.com/redhat-plumbers-in-action/differential-shellcheck/tree/main/src), and I'm happy to answer your questions. Also, you can have a look at some repositories that are already using this action (e.g. [systemd](https://github.com/systemd/systemd/actions/workflows/differential-shellcheck.yml)).

DentonGentry:
I think that running this in CI with these permissions is not the right tradeoff for us. We'd be happy to accept PRs updating the shell scripts if you want to run the repository through the tool yourself, but I'm going to decline to run this as part of our CI.

jamacku:
> I think running this in CI with these permissions is not the right tradeoff for us. We'd be happy to accept PRs updating the shell scripts if you want to run the repository through the tool yourself, but I'm going to decline to run this as part of our CI.

Just for the record, you already have a workflow that is running with a similar permissions - [codeql](https://github.com/tailscale/tailscale/blob/main/.github/workflows/codeql-analysis.yml)

PR #5515, opened 2022-08-31
tsnet: add Server method to listener

willnorris:
Allow callers to verify that a net.Listener is a tsnet.listener by type asserting against this Server method, as well as providing access to the underlying Server.

This is initially being added to support the caddy integration in caddyserver/caddy#5002.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -541,6 +541,9 @@ func (ln *listener) Close() error {
 	return nil
 }
 
+// Server returns the tsnet Server associated with the listener.
+func (ln *listener) Server() *Server { return ln.s }
+
 type addr struct{ ln *listener }
 
 func (a addr) Network() string { return a.ln.key.network }
--- a/tsnet/tsnet_test.go
+++ b/tsnet/tsnet_test.go
@@ -1,0 +1,18 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package tsnet
+
+import "testing"
+
+// TestListener_Server ensures that the listener type always keeps the Server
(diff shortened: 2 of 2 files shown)


tsnet/tsnet.go
```diff
@@ -540,6 +540,7 @@ func (ln *listener) Close() error {
 	}
 	return nil
 }
+func (ln *listener) Server() *Server { return ln.s }
```
bradfitz, line 543: blank line before.

and document? well, it's exported on unexported type, so how would people use this? sniff it out of an undocumented interface? then add a test so we don't remove this in the future?

willnorris: yes, basically.  See `tsnetListener` in https://github.com/tailscale/caddy-tailscale-auth/pull/1/files.  Will add docs and test.

willnorris:
okay, PTAL

tsnet/tsnet_test.go
```diff
@@ -0,0 +1,8 @@
+package tsnet
```
bradfitz, line 1: copyright header.

tsnet/tsnet_test.go
```diff
@@ -0,0 +1,8 @@
...
+
+// ensure that the listener type always keeps the Server method, which is used
+// by some external applications to identify a tsnet.Listener from other
+// net.Listeners, as well as access the underlying Server.
+var _ interface {
+	Server() *Server
+} = new(listener)
```
bradfitz, line 8: just make a regular TestFoo test?

bradfitz: But if you'd rather check this at compile time, then put this in tsnet.go itself.

And write `= (*listener)(nil)`

willnorris: done

tsnet/tsnet_test.go
willnorris: done

bradfitz (approved):

PR #5516, opened 2022-09-01
cmd/nginx-auth/nginx-auth: update auth to allow for new domains

nyghtowl:
With MagicDNS GA, we are giving every tailnet a tailnet-<hex>.ts.net name.
We will only parse out if legacy domains include beta.tailscale.net; otherwise,
set tailnet to the full domain format going forward.

### DIFF
--- a/cmd/nginx-auth/nginx-auth.go
+++ b/cmd/nginx-auth/nginx-auth.go
@@ -75,12 +75,7 @@ func main() {
 				log.Printf("can't extract tailnet name from hostname %q", info.Node.Name)
 				return
 			}
-			tailnet, _, ok = strings.Cut(tailnet, ".beta.tailscale.net")
-			if !ok {
-				w.WriteHeader(http.StatusUnauthorized)
-				log.Printf("can't extract tailnet name from hostname %q", info.Node.Name)
-				return
-			}
+			tailnet = strings.TrimSuffix(tailnet, ".beta.tailscale.net")
 		}
 
 		if expectedTailnet := r.Header.Get("Expected-Tailnet"); expectedTailnet != "" && expectedTailnet != tailnet {


cmd/nginx-auth/nginx-auth.go
```diff
@@ -75,12 +75,7 @@ func main() {
 				log.Printf("can't extract tailnet name from hostname %q", info.Node.Name)
 				return
 			}
-			tailnet, _, ok = strings.Cut(tailnet, ".beta.tailscale.net")
-			if !ok {
-				w.WriteHeader(http.StatusUnauthorized)
-				log.Printf("can't extract tailnet name from hostname %q", info.Node.Name)
-				return
-			}
+			tailnet, _, ok = strings.TrimSuffix(tailnet, ".beta.tailscale.net")
```
maisem, line 78: tailnet = strings.TrimSuffx(tailnet, ".beta.tailscale.net")

maisem:
@willnorris can you review this?

willnorris:
I think this change should be fine, but it doesn't actually solve the problem.  

We probably don't need to return a 401 when failing to extract the tailnet name, either when removing the `beta.tailscale.net` suffix, or the node name prefix right above that.  I'm pretty sure that successfully extracting the tailnet name really only matters when using `Expected-Tailnet`, and that check would likely fail anyway if the tailnet name could be determined so no need to fail early.  @xe can confirm since this was their initial code.

However, this will still break things since these new `tailnet-HEX.ts.net` names no longer have the tailnet name in them.  We still should just add the Tailnet to [`tailcfg.Node`](https://github.com/tailscale/tailscale/blob/cfbbcf6d071dd13addbc1b38ef993a08d557ba22/tailcfg/tailcfg.go#L166), and then we can remove all of this logic.  Or perhaps we really care about the tailnet of the user and it should be on [`tailcfg.UserProfile`](https://github.com/tailscale/tailscale/blob/cfbbcf6d071dd13addbc1b38ef993a08d557ba22/tailcfg/tailcfg.go#L145)?

PR #5524, opened 2022-09-01
cmd/tailscale/cli: add backwards compatibility 'up' processing for legacy client

bradfitz:
Updates tailscale/corp#6781

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -164,6 +164,7 @@ type LocalBackend struct {
 	authURL          string // cleared on Notify
 	authURLSticky    string // not cleared on Notify
 	interact         bool
+	egg              bool
 	prevIfState      *interfaces.State
 	peerAPIServer    *peerAPIServer // or nil
 	peerAPIListeners []*peerAPIListener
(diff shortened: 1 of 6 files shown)


willnorris (approved):

ipn/ipnlocal/local.go
```diff
@@ -2015,6 +2015,11 @@ func (b *LocalBackend) isDefaultServerLocked() bool {
 
 func (b *LocalBackend) EditPrefs(mp *ipn.MaskedPrefs) (*ipn.Prefs, error) {
 	b.mu.Lock()
+	if len(mp.AdvertiseTags) > 0 && mp.AdvertiseTags[len(mp.AdvertiseTags)-1] == "egg" {
+		mp.AdvertiseTags = mp.AdvertiseTags[:len(mp.AdvertiseTags)-1]
```
willnorris, line 2019: what happens if this doesn't get removed from AdvertiseTags?  Will that cause problems anywhere else?  I'm thinking in particular what if for some reason it's not the last item in the slice, which this assumes?

bradfitz: earlier code (and code in control) filters out ones that don't start with "tag:". But I changed it to be less mysterious. PTAL.

willnorris (approved):

PR #5526, opened 2022-09-01
wgengine: fix race on endpoints in getStatus

raggi:


andrew-d (approved):

wgengine/userspace.go
```diff
@@ -1005,6 +1005,7 @@ func (e *userspaceEngine) getStatus() (*Status, error) {
 	closing := e.closing
 	peerKeys := make([]key.NodePublic, len(e.peerSequence))
 	copy(peerKeys, e.peerSequence)
+	localAddrs := append([]tailcfg.Endpoint(nil), e.endpoints...)
```
bradfitz, line 1008: https://pkg.go.dev/golang.org/x/exp/slices#Clone is a bit prettier these days

localAddrs := slices.Clone(e.endpoints)

bradfitz (approved):

bradfitz:
This commit message could've been improved in a few ways:

* say it's a data race
* reference when it was introduced
* referenced a bug number

PR #5529, opened 2022-09-01
cmd/derper: add support for unpublished bootstrap DNS entries

andrew-d:

### DIFF
--- a/cmd/derper/bootstrap_dns.go
+++ b/cmd/derper/bootstrap_dns.go
@@ -17,16 +17,31 @@ import (
 	"tailscale.com/syncs"
 )
 
-var dnsCache syncs.AtomicValue[[]byte]
+const refreshTimeout = time.Minute
 
-var bootstrapDNSRequests = expvar.NewInt("counter_bootstrap_dns_requests")
+type dnsEntryMap map[string][]net.IP
+
+var (
+	dnsCache            syncs.AtomicValue[dnsEntryMap]
+	dnsCacheBytes       syncs.AtomicValue[[]byte] // of JSON
+	unpublishedDNSCache syncs.AtomicValue[dnsEntryMap]
+)
+
+var (
+	bootstrapDNSRequests = expvar.NewInt("counter_bootstrap_dns_requests")
+	publishedDNSHits     = expvar.NewInt("counter_bootstrap_dns_published_hits")
+	publishedDNSMisses   = expvar.NewInt("counter_bootstrap_dns_published_misses")
+	unpublishedDNSHits   = expvar.NewInt("counter_bootstrap_dns_unpublished_hits")
+	unpublishedDNSMisses = expvar.NewInt("counter_bootstrap_dns_unpublished_misses")
+)
 
 func refreshBootstrapDNSLoop() {
-	if *bootstrapDNS == "" {
+	if *bootstrapDNS == "" && *unpublishedDNS == "" {
 		return
 	}
 	for {
 		refreshBootstrapDNS()
+		refreshUnpublishedDNS()
 		time.Sleep(10 * time.Minute)
 	}
 }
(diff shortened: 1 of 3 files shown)


danderson (approved):

cmd/derper/bootstrap_dns.go
```diff
@@ -35,10 +46,32 @@ func refreshBootstrapDNS() {
...
+func refreshUnpublishedDNS() {
+	if *unpublishedDNS == "" {
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), refreshTimeout)
 	defer cancel()
-	names := strings.Split(*bootstrapDNS, ",")
+
+	dnsEntries := resolveList(ctx, strings.Split(*unpublishedDNS, ","))
```
bradfitz, line 68: keep in mind that strings.Split on an empty string will yield a 1-length slice of an empty string: https://go.dev/play/p/m5X9rfocwem

cmd/derper/bootstrap_dns.go
```diff
@@ -48,21 +81,32 @@ func refreshBootstrapDNS() {
...
+			m := map[string][]net.IP{q: ips}
+			j, err := json.MarshalIndent(m, "", "\t")
+			if err == nil {
+				w.Write(j)
+				return
+			}
+		}
+	}
+
+	// Fall back to the global DERP map
```
bradfitz, line 109: not really a "global DERP map".

More like the public set of cached DNS names?

cmd/derper/bootstrap_dns.go
```diff
@@ -17,16 +17,27 @@ import (
...
+type dnsEntryMap map[string][]net.IP
+
+var (
+	dnsCache            syncs.AtomicValue[[]byte]
+	unpublishedDNSCache syncs.AtomicValue[dnsEntryMap]
+)
+
+var (
+	bootstrapDNSRequests = expvar.NewInt("counter_bootstrap_dns_requests")
+	unpublishedDNSHits   = expvar.NewInt("counter_bootstrap_dns_unpublished_hits")
```
bradfitz, line 31: want to do all 4 counters instead of just this 1? {unpublished,published} x {hits,misses} ?  (in addition to the overall _requests one)

cmd/derper/bootstrap_dns.go
andrew-d: Should be handled by the `if` statement at the top that exits if the string is empty

bradfitz: whoops, missed that

cmd/derper/bootstrap_dns.go
```diff
@@ -48,21 +87,42 @@ func refreshBootstrapDNS() {
...
+		} else {
+			unpublishedDNSMisses.Add(1)
+		}
+
+		// Also track whether this is a cache hit in our published DNS cache.
+		if _, ok := dnsCache.Load()[q]; ok {
+			publishedDNSHits.Add(1)
+		} else {
+			publishedDNSMisses.Add(1)
+		}
```
bradfitz, line 122: I was imagining a different definition for these counters:

* if the "q" name is one of the 4 at https://derp10.tailscale.com/bootstrap-dns and we have a non-empty value, that's a `publishedDNSHits`.
* else, if the "q" name is one of the 4 and its value is currently `[]`, that's a `publishedDNSMisses`
* else, ... unpublished hits vs miss

But a given `q` can't be both published and unpublished.

andrew-d: Makes sense; changed that

cmd/derper/bootstrap_dns.go
```diff
@@ -17,16 +17,31 @@ import (
...
 
-var dnsCache syncs.AtomicValue[[]byte]
+const refreshTimeout = time.Minute
 
-var bootstrapDNSRequests = expvar.NewInt("counter_bootstrap_dns_requests")
+type dnsEntryMap map[string][]net.IP
+
+var (
+	dnsCache            syncs.AtomicValue[dnsEntryMap]
+	dnsCacheBytes       syncs.AtomicValue[[]byte]
```
bradfitz, line 26: same line comment like `// of JSON` ?

cmd/derper/bootstrap_dns.go
```diff
@@ -48,21 +87,47 @@ func refreshBootstrapDNS() {
...
-	// and are randomized per request,
-	// so keeping a connection open is pointlessly expensive.
+	// Bootstrap DNS requests occur cross-regions, and are randomized per
+	// request, so keeping a connection open is pointlessly expensive.
 	w.Header().Set("Connection", "close")
+
+	// Try answering a query from our hidden map first
+	if q := r.URL.Query().Get("q"); q != "" {
+		if ips, ok := unpublishedDNSCache.Load()[q]; ok {
+			unpublishedDNSHits.Add(1)
```
bradfitz, line 104: is it still a hit of len(ips) == 0? 🤷

bradfitz (approved):

cmd/derper/bootstrap_dns.go
andrew-d: Changed this to only respond (and only increment the metric) if there's a non-zero number of IPs in the list, since returning an empty list isn't helpful to anybody.

PR #5535, opened 2022-09-02
wgengine/magicsock: add filter to ignore disco to old/other ports

raggi:
Incoming disco packets are now dropped unless they match one of the
current bound ports, or have a zero port*.

The BPF filter passes all packets with a disco header to the raw packet
sockets regardless of destination port (in order to avoid needing to
reconfigure BPF on rebind).

If a BPF enabled node has just rebound, due to restart or rebind, it may
receive and reply to disco ping packets destined for ports other than
those which are presently bound. If the pong is accepted, the pinging
node will now assume that it can send WireGuard traffic to the pinged
port - such traffic will not reach the node as it is not destined for a
bound port.

*The zero port is ignored, if received. This is a speculative defense
and would indicate a problem in the receive path, or the BPF filter.
This condition is allowed to pass as it may enable traffic to flow,
however it will also enable problems with the same symptoms this patch
otherwise fixes.

Fixes https://github.com/tailscale/tailscale/issues/5536

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -2989,11 +2989,13 @@ type RebindingUDPConn struct {
 
 	mu    sync.Mutex // held while changing pconn (and pconnAtomic)
 	pconn nettype.PacketConn
+	port  uint16
 }
 
 func (c *RebindingUDPConn) setConnLocked(p nettype.PacketConn) {
 	c.pconn = p
 	c.pconnAtomic.Store(p)
+	c.port = uint16(c.localAddrLocked().Port)
 }
 
 // currentConn returns c's current pconn, acquiring c.mu in the process.
@@ -3055,6 +3057,12 @@ func (c *RebindingUDPConn) ReadFromNetaddr(b []byte) (n int, ipp netip.AddrPort,
 	}
 }
 
+func (c *RebindingUDPConn) Port() uint16 {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.port
+}
+
 func (c *RebindingUDPConn) LocalAddr() *net.UDPAddr {
 	c.mu.Lock()
 	defer c.mu.Unlock()
@@ -3079,6 +3087,7 @@ func (c *RebindingUDPConn) closeLocked() error {
 	if c.pconn == nil {
 		return errNilPConn
 	}
+	c.port = 0
 	return c.pconn.Close()
 }
 
--- a/wgengine/magicsock/magicsock_linux.go
+++ b/wgengine/magicsock/magicsock_linux.go
@@ -195,11 +195,11 @@ func (c *Conn) listenRawDisco(family string) (io.Closer, error) {
 	}
 	pc.SetReadDeadline(time.Time{})
 
-	go c.receiveDisco(pc)
+	go c.receiveDisco(pc, family == "ip6")
 	return pc, nil
 }
 
-func (c *Conn) receiveDisco(pc net.PacketConn) {
+func (c *Conn) receiveDisco(pc net.PacketConn, isIPV6 bool) {
 	var buf [1500]byte
 	for {
 		n, src, err := pc.ReadFrom(buf[:])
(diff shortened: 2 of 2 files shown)


wgengine/magicsock/magicsock_linux.go
```diff
@@ -213,6 +213,19 @@ func (c *Conn) receiveDisco(pc net.PacketConn) {
 			// Too small to be a valid UDP datagram, drop.
 			continue
 		}
+
+		dstPort := binary.BigEndian.Uint16(buf[2:4])
+
+		// TODO(raggi): unsure if these are memoized by pcon, if not we should do so.
+		v4Port := c.pconn4.LocalAddr().Port
```
danderson, line 220: You get correct port numbers back from this? Our original change had almost exactly this, but the pconns returned Port=0 all the time, leading to dropping all disco traffic.

danderson (approved):
Assuming the pconn's LocalAddr works correctly, LGTM. Thanks!

wgengine/magicsock/magicsock_linux.go
raggi: I'm sure on a system without ipv6 for example, the v6 one returns 0, as it has a zero value explicitly returned from the fake conn that's in there.

A 0 compare should not result in all traffic being dropped however, as it shouldn't match the incoming port number. I can explicitly skip 0 values as a defensive measure though.

Were you getting 0 ports from the UDP header? That would be more concerning as it could lead to a broader issue.

danderson: No, the port numbers in the packet were fine, but all disco packets were getting dropped because the comparisons were all 0 == 39482 or whatever.

Are you running tailscaled with a --port argument? I was running in userspace mode without --port, if that's relevant to what I saw.

It's also possible my code was just wrong at the time, I didn't spend much time trying to debug it because I thought it was unnecessary.

wgengine/magicsock/magicsock_linux.go
```diff
@@ -213,6 +213,19 @@ func (c *Conn) receiveDisco(pc net.PacketConn) {
...
+		v6Port := c.pconn6.LocalAddr().Port
+
+		if dstPort != uint16(v4Port) && dstPort != uint16(v6Port) {
```
bradfitz, line 223: I assume this is temporary and you'll actually only check the v4port for IPv4 srcs and v6Port for IPv6 srcs?

bradfitz: (that is, this will need to move after the srcIP initialization below)

raggi: Yes, this is addressed. I addressed it slightly differently, by passing in the family instead.

wgengine/magicsock/magicsock_linux.go
raggi: Ah, userspace mode, interesting. I shall have to test that.

wgengine/magicsock/magicsock.go
```diff
@@ -3023,9 +3023,13 @@ type RebindingUDPConn struct {
 
 	mu    sync.Mutex // held while changing pconn (and pconnAtomic)
 	pconn nettype.PacketConn
+	// laddr is laziliy memoized for frequent access from disco, cleared when
+	// the pconn is updated, and populated by the next call to LocalAddr.
+	laddr *net.UDPAddr
```
bradfitz, line 3028: store a netip.AddrPort instead? and then don't worry about the defensive copy stuff

wgengine/magicsock/magicsock_linux.go
raggi: Userspace seems fine. I also tested with both a fixed port number (which effectively mitigates the original issue too), and with `--port 0`.

I'm not able to reproduce the condition you describe.

wgengine/magicsock/magicsock.go
raggi: Oh, PacketConn.LocalAddr() actually has a very loose specification. Sadly netip.AddrPort does quite meet the specification so I can't just return that.

I can just add a .Port() extension here instead.

raggi: done, PTAL.

wgengine/magicsock/magicsock_linux.go
```diff
@@ -213,6 +213,37 @@ func (c *Conn) receiveDisco(pc net.PacketConn) {
...
 		}
+
+		dstPort := binary.BigEndian.Uint16(buf[2:4])
+		if dstPort == 0 {
+			c.logf("[unexpected] disco raw: received packet for port 0")
+		}
+
+		var acceptPort uint16
+		if isIPV6 {
+			acceptPort = c.pconn6.Port()
```
bradfitz, line 224: Looks like `pconn6` can be nil. There are a bunch of checks for it being nil in magicsock. See `initialBind`.

Do `if isIPV6 && p.conn6 != nil` here?

The pconn4 is always non-nil.

wgengine/magicsock/magicsock_linux.go
```diff
@@ -213,6 +213,37 @@ func (c *Conn) receiveDisco(pc net.PacketConn) {
...
+			continue
+		}
+
+		// Speculative defense: if dstPort == 0, all disco traffic would be
+		// dropped in conditions such as a host with no ipv6.  This is logged
+		// above, and ignored here so that disco might still flow. If this
+		// condition occurs, it could lead to bugs with behavior such as #5536
+		// with symptoms such as peers only able to communicate disco and no
+		// wireguard data for the duration of a disco session due to a port
+		// mismatch.
```
bradfitz, line 241: I don't exactly follow this.

raggi: See the conversation with @danderson above. I tried to reword it a few ways. It's a subtle condition that I believe should never happen, but Dave reports having seen.

I removed the doc string and the dstPort condition as its effectively covered by new conditions above.

wgengine/magicsock/magicsock_linux.go
raggi: Good call, I made a new PR https://github.com/tailscale/tailscale/pull/5537 and rebased atop it to remove this condition.

PR #5537, opened 2022-09-02
wgengine/magicksock: remove nullability of RebindingUDPConns

raggi:
Both RebindingUDPConns now always exist. the initial bind (which now
just calls rebind) now ensures that bind is called for both, such that
they both at least contain a blockForeverConn. Calling code no longer
needs to assert their state.

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -262,8 +262,8 @@ type Conn struct {
 	// pconn4 and pconn6 are the underlying UDP sockets used to
 	// send/receive packets for wireguard and other magicsock
 	// protocols.
-	pconn4 *RebindingUDPConn
-	pconn6 *RebindingUDPConn
+	pconn4 RebindingUDPConn
+	pconn6 RebindingUDPConn
 
 	// closeDisco4 and closeDisco6 are io.Closers to shut down the raw
 	// disco packet receivers. If nil, no raw disco receiver is
@@ -570,7 +570,7 @@ func NewConn(opts Options) (*Conn, error) {
 	}
 	c.linkMon = opts.LinkMonitor
 
-	if err := c.initialBind(); err != nil {
+	if err := c.rebind(keepCurrentPort); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


bradfitz (commented):
This doesn't seem sufficient. Will comment more when at a computer.

wgengine/magicsock/magicsock.go
```diff
@@ -2822,17 +2808,7 @@ func (c *Conn) ReSTUN(why string) {
 }
 
 func (c *Conn) initialBind() error {
```
bradfitz, line 2810: this initialBind is probably not worth keeping at this point? just move the rebind to the sole caller?

wgengine/magicsock/magicsock.go
```diff
@@ -578,14 +578,12 @@ func NewConn(opts Options) (*Conn, error) {
...
-		GetSTUNConn4:        func() netcheck.STUNConn { return c.pconn4 },
+		GetSTUNConn4:        func() netcheck.STUNConn { return &c.pconn4 },
 		SkipExternalNetwork: inTest(),
 		PortMapper:          c.portMapper,
 	}
 
-	if c.pconn6 != nil {
-		c.netChecker.GetSTUNConn6 = func() netcheck.STUNConn { return c.pconn6 }
-	}
+	c.netChecker.GetSTUNConn6 = func() netcheck.STUNConn { return &c.pconn6 }
```
bradfitz, line 586: I thought this might've been sad, making this hook always populated non-nil even when there was no kernel IPv6 support, but it turns out that's actually better behavior! otherwise netcheck tries to bind to ipv6 again just to waste time failing.

bradfitz, line 586: move this up to after line 581

wgengine/magicsock/magicsock.go
```diff
@@ -2930,13 +2901,13 @@ func (c *Conn) rebind(curPortFate currentPortFate) error {
 	if runtime.GOOS == "js" {
 		return nil
 	}
+	if err := c.bindSocket(&c.pconn6, "udp6", curPortFate); err != nil {
```
bradfitz, line 2904: okay, I understand why this is safe now: all the failure paths from bindSocket do `ruc.setConnLocked(newBlockForeverConn())` so other calls to the Rebinding conn won't crash later.

bradfitz (approved):

PR #5545, opened 2022-09-05
net/dnsfallback: cache most recent DERP map on disk

andrew-d:
This is especially helpful as we launch newer DERPs over time, and older
clients have progressively out-of-date static DERP maps baked in. After
this, as long as the client has successfully connected once, it'll cache
the most recent DERP map it knows about.

Resolves an in-code comment from @bradfitz

### DIFF
--- a/net/dnsfallback/dnsfallback.go
+++ b/net/dnsfallback/dnsfallback.go
@@ -20,8 +20,12 @@ import (
 	"net/http"
 	"net/netip"
 	"net/url"
+	"os"
+	"reflect"
+	"sync/atomic"
 	"time"
 
+	"tailscale.com/atomicfile"
 	"tailscale.com/net/netns"
 	"tailscale.com/net/tlsdial"
 	"tailscale.com/net/tshttpproxy"
@@ -39,6 +43,7 @@ func Lookup(ctx context.Context, host string) ([]netip.Addr, error) {
 	}
 
 	dm := getDERPMap()
+
 	var cands4, cands6 []nameIP
 	for _, dr := range dm.Regions {
 		for _, n := range dr.Nodes {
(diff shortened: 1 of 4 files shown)


net/dnsfallback/dnsfallback.go
```diff
@@ -128,12 +132,42 @@ type dnsMap map[string][]netip.Addr
...
+
+	// Merge in any DERP servers from the cached map that aren't in the
+	// static map; this ensures that we're getting new region(s) while not
+	// overriding the built-in fallbacks if things go horribly wrong and we
+	// get a bad DERP map.
+	//
+	// TODO(andrew): should we expect OmitDefaultRegions here? We're not
+	// forwarding traffic, just resolving DNS, so maybe we can ignore that
+	// value anyway?
+	if cached := cachedDERPMap.Load(); cached != nil {
```
bradfitz, line 145: invert test, return dm early, and outdent rest?  (won't be able to use a ShortVarDecl in the `if` but oh well)

net/dnsfallback/dnsfallback.go
```diff
@@ -143,3 +177,58 @@ func getDERPMap() *tailcfg.DERPMap {
 
 //go:embed dns-fallback-servers.json
 var staticDERPMapJSON []byte
+
+// cachedDERPMap is the path to a cached DERP map that we loaded from our on-disk cache
+var cachedDERPMap syncs.AtomicValue[*tailcfg.DERPMap]
```
bradfitz, line 182: could just use the standard library for this as it's just a pointer... https://pkg.go.dev/sync/atomic#Pointer

net/dnsfallback/dnsfallback.go
```diff
@@ -143,3 +177,58 @@ func getDERPMap() *tailcfg.DERPMap {
...
+
+// cachePath is the path to the DERP map cache file
+var cachePath string
+
+// UpdateCache stores the DERP map cache back to disk.
+func UpdateCache(c *tailcfg.DERPMap) {
```
bradfitz, line 188: I guess we take ownership of c? Or the caller must promise to not mutate it at least?

net/dnsfallback/dnsfallback.go
```diff
@@ -143,3 +177,58 @@ func getDERPMap() *tailcfg.DERPMap {
...
+	d, err := json.Marshal(c)
```
bradfitz, line 189: could we check if it's unchanged first before doing the rest of this work?

andrew-d: Done using `reflect.DeepEqual`; could do something else if you'd prefer. This doesn't happen particularly often, AFAIK?

PR #5550, opened 2022-09-06
net/dnsfallback: allow setting log function

andrew-d:
This broke a test in corp that enforces we don't use the log package.


net/dnsfallback/dnsfallback.go
```diff
@@ -213,11 +215,10 @@ func UpdateCache(c *tailcfg.DERPMap) {
 	if cachePath != "" {
 		err = atomicfile.WriteFile(cachePath, d, 0600)
 		if err != nil {
-			log.Printf("[v1] dnsfallback: UpdateCache error writing: %v", err)
+			logf("[v1] dnsfallback: UpdateCache error writing: %v", err)
 			return
 		}
 	}
-	log.Printf("[v2] dnsfallback: UpdateCache succeeded")
```
andrew-d, line 220: Removed this line because it's not actually helpful, IMO

net/dnsfallback/dnsfallback_test.go
```diff
@@ -25,6 +25,7 @@ func TestGetDERPMap(t *testing.T) {
 }
 
 func TestCache(t *testing.T) {
+	SetLogger(t.Logf)
```
bradfitz, line 28: Restore on exit from test?

bradfitz (approved):
Not great log plumbing but sure. :)

I dislike a lot of our current situation anyway so this is fine for now

PR #5558, opened 2022-09-06
control/controlclient,ipn/ipnlocal: wire tka enable/disable

twitchyliquid64:
Implement enablement and disablement of TKA based on communications with the control plane.

Theres also some amount of plumbing for testing the noise endpoints (we don't have any tests hitting those!?) and making the disablementKDF function public.


ipn/ipnlocal/network-lock.go
```diff
@@ -31,6 +33,104 @@ type tkaState struct {
...
+		return errors.New("SelfNode missing")
+	}
+
+	isEnabled := b.tka != nil
+	wantEnabled := nm.TKAEnabled
+	if isEnabled != wantEnabled {
+		// Regardless of whether we are moving to disabled or enabled, we
+		// need information from the tka bootstrap endpoint.
+		b.mu.Unlock()
+		bs, err := b.tkaFetchBootstrap(nm.SelfNode.ID, nm.TKAHead)
```
sailorfrag, line 68: Is the bootstrap endpoint already written? I think it's probably more useful to pass in the backend's last known head rather than the netmap current head, since the latter is something that control should be able to figure out on its own (modulo some version skew, but a slightly newer head that incorporates the netmap head seems like it should be pretty easy to recover from)

ipn/ipnlocal/network-lock.go
```diff
@@ -31,6 +33,104 @@ type tkaState struct {
...
+			return err
+		}
+
+		if wantEnabled && !isEnabled {
+			if err := b.tkaBootstrapFromGenesisLocked(bs.GenesisAUM); err != nil {
+				return fmt.Errorf("bootstrap: %v", err)
+			}
+			isEnabled = true
+		} else if !wantEnabled && isEnabled {
+			if b.tka.authority.ValidDisablement(bs.DisablementSecret) {
```
sailorfrag, line 80: else error of some sort?

ipn/ipnlocal/network-lock.go
```diff
@@ -31,6 +33,104 @@ type tkaState struct {
...
+	if !b.CanSupportNetworkLock() {
+		return errors.New("network lock not supported in this configuration")
+	}
+
+	var genesis tka.AUM
+	if err := genesis.Unserialize(g); err != nil {
+		return fmt.Errorf("reading genesis: %v", err)
+	}
+
+	chonkDir := filepath.Join(b.TailscaleVarRoot(), "chonk")
```
sailorfrag, line 113: I think this is repeated enough times it should be its own receiver

ipn/ipnlocal/network-lock.go
```diff
@@ -31,6 +33,104 @@ type tkaState struct {
...
+	if err := os.Mkdir(chonkDir, 0755); err != nil && !os.IsExist(err) {
```
sailorfrag, line 114: Is there anything sensitive in there? Would 0700 be better?

ipn/ipnlocal/network-lock_test.go
```diff
@@ -0,0 +1,232 @@
...
+
+	ts, client := fakeNoiseServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer r.Body.Close()
+		switch r.URL.Path {
+		case "/machine/tka/bootstrap":
+			body := new(tailcfg.TKABootstrapRequest)
+			if err := json.NewDecoder(r.Body).Decode(body); err != nil {
+				t.Fatal(err)
+			}
+			if body.NodeID != 420 {
```
sailorfrag, line 85: check for expected head?

ipn/ipnlocal/network-lock.go
twitchyliquid64: Sounds good, done!

ipn/ipnlocal/network-lock.go
```diff
@@ -31,6 +33,104 @@ type tkaState struct {
...
+	}
+
+	isEnabled := b.tka != nil
+	wantEnabled := nm.TKAEnabled
+	if isEnabled != wantEnabled {
+		// Regardless of whether we are moving to disabled or enabled, we
+		// need information from the tka bootstrap endpoint.
+		b.mu.Unlock()
+		bs, err := b.tkaFetchBootstrap(nm.SelfNode.ID, nm.TKAHead)
+		b.mu.Lock()
```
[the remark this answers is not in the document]
twitchyliquid64: No - bootstrap returns values for enablement or disablement but doesnt indicate one way or the other - thats in the netmap.

ipn/ipnlocal/network-lock.go
twitchyliquid64: Everything here is public (no secrets) - the NL key is in the prefs file, along with the other keys (so i presume it is permissioned appropriately).

ipn/ipnlocal/network-lock.go
```diff
@@ -31,6 +33,115 @@ type tkaState struct {
...
+
+	isEnabled := b.tka != nil
+	wantEnabled := nm.TKAEnabled
+	if isEnabled != wantEnabled {
+		// Regardless of whether we are moving to disabled or enabled, we
+		// need information from the tka bootstrap endpoint.
+		b.mu.Unlock()
+		var ourHead tka.AUMHash
+		if b.tka != nil {
+			ourHead = b.tka.authority.Head()
```
sailorfrag, line 70: Should this be happening with the lock still held? Seems like at minimum there's a race with `b.tka` possibly changing or going nil

ipn/ipnlocal/network-lock.go
sailorfrag: I actually was referring to if the disablement isn't valid 😄

sailorfrag (approved):
Sorry for the delay, github is terrible at showing what's changed since last time if there's a rebase mixed in.

A couple things that should be trivial to resolve, feel free to merge after making changes or writing why you disagree.

ipn/ipnlocal/network-lock_test.go
```diff
@@ -0,0 +1,242 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
```
bradfitz, line 1: 2022

ipn/ipnlocal/network-lock.go
```diff
@@ -31,6 +33,115 @@ type tkaState struct {
...
+//   - Enablement: nm.TKAEnabled but b.tka == nil
+//     ∴ reach out to /machine/tka/boostrap to get the genesis AUM, then
+//     initialize TKA.
+//   - Disablement: !nm.TKAEnabled but b.tka != nil
+//     ∴ reach out to /machine/tka/boostrap to read the disablement secret,
+//     then verify and clear tka local state.
+//   - Sync needed: b.tka.Head != nm.TKAHead
+//     ∴ complete multi-step synchronization flow.
+//   - Everything up to date: All other cases.
+//     ∴ chill out & crack open a cold one.
```
bradfitz, line 49: let's be serious here

also, what's with the unicode?

ipn/ipnlocal/network-lock.go
```diff
@@ -31,6 +33,115 @@ type tkaState struct {
 	storage   *tka.FS
 }
 
+// TKASyncIfNeeded examines tka info reported from the control plane,
```
bradfitz, line 36: TKA (acronym so capitalized)

control/controlclient/map.go
```diff
@@ -152,9 +156,17 @@ func (ms *mapSession) netmapForResponse(resp *tailcfg.MapResponse) *netmap.Netwo
 		DERPMap:         ms.lastDERPMap,
 		Debug:           debug,
 		ControlHealth:   ms.lastHealth,
+		TKAEnabled:      ms.lastTKAInfo != nil && !ms.lastTKAInfo.Disabled,
 	}
 	ms.netMapBuilding = nm
 
+	if ms.lastTKAInfo != nil && ms.lastTKAInfo.Head != "" {
+		if err := nm.TKAHead.UnmarshalText([]byte(ms.lastTKAInfo.Head)); err != nil {
+			ms.logf("unmarshalling TKAHead: %v", err)
```
bradfitz, line 165: "error unmarshalling TKAHead: %v"

(get "error" in the log print for easier grepping)

ipn/ipnlocal/network-lock.go
bradfitz, line 36: lowercase tkaSyncIfNeeded to match subject

ipn/ipnlocal/network-lock.go
```diff
@@ -31,6 +33,115 @@ type tkaState struct {
...
+	}
+
+	return nil
+}
+
+// chonkPath returns the absolute path to the directory in which TKA
+// state (the 'tailchonk') is stored.
+func (b *LocalBackend) chonkPath() string {
+	return filepath.Join(b.TailscaleVarRoot(), "chonk")
+}
```
bradfitz, line 108: Does this weird "chonk" word need to extend to paths on disk?

Can we name it something more like "network-lock"?

ipn/ipnlocal/network-lock.go
```diff
@@ -237,3 +348,50 @@ func (b *LocalBackend) tkaInitFinish(nm *netmap.NetworkMap, nks map[tailcfg.Node
...
+	if err := json.NewEncoder(&req).Encode(bootstrapReq); err != nil {
+		return nil, fmt.Errorf("encoding request: %v", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+	if err := ctx.Err(); err != nil {
+		return nil, fmt.Errorf("ctx: %w", err)
+	}
+	req2, err := http.NewRequestWithContext(ctx, "GET", "https://example.com/machine/tka/bootstrap", &req)
```
bradfitz, line 376: is example.com what we do elsewhere? How about just "control"? or "control-noise" ?

sailorfrag: I've seen `https://example.com` in some places and `https://unused` in others, the latter maybe being slightly more prevalent.

bradfitz: unused is better than example.com which looks like it's docs

twitchyliquid64: `httptest` will only accept `example.com` unfortunately.

ipn/ipnlocal/network-lock.go
twitchyliquid64: Its the therefore symbol!

ipn/ipnlocal/network-lock.go
sailorfrag: I'd suggest using the control URL then, and tests can override that?

ipn/ipnlocal/network-lock.go
twitchyliquid64: I don't want to return an error as that'll stop processing, and that would end up being another holdback attack vector from control. I'll log an error instead.

ipn/ipnlocal/network-lock.go
twitchyliquid64: Good catch, fixed!

ipn/ipnlocal/network-lock.go
twitchyliquid64: Updating to 'tka' to match the subsystem name.

ipn/ipnlocal/network-lock.go
twitchyliquid64: That unfortunately won't work either, as the TLS config, certificate, and key are quite literally hardcoded in Go: https://cs.opensource.google/go/go/+/refs/tags/go1.19.1:src/net/http/internal/testcert/testcert.go;l=14;drc=d8220ac32fc73e69aa6d58890b6b5277d277825f

We could do the work to generate a fake config, cert, and key, but I'd rather not because I like that `example.com` indicates the request.Do() flow is tested somewhere whereas `unused` suggests theres no such test.

ipn/ipnlocal/network-lock.go
```diff
@@ -237,3 +351,50 @@ func (b *LocalBackend) tkaInitFinish(nm *netmap.NetworkMap, nks map[tailcfg.Node
...
 	}
 }
+
+// tkaFetchBootstrap sends a /machine/tka/bootstrap RPC to the control plane
+// over noise. This is used to get values necessary to enable or disable TKA.
+func (b *LocalBackend) tkaFetchBootstrap(nodeID tailcfg.NodeID, head tka.AUMHash) (*tailcfg.TKABootstrapResponse, error) {
+	bootstrapReq := tailcfg.TKABootstrapRequest{
+		NodeID: nodeID,
+	}
+	if head != (tka.AUMHash{}) {
```
maisem, line 361: add a IsZero func?

ipn/ipnlocal/network-lock.go
sailorfrag: Oh sorry, I was jumping a bunch without explaining my thought further. If the request is made to the control URL, that makes this code clearer. To make testing with `httptest` work correctly, set the control URL to `example.com` when initializing a test. Would that satisfy everything?

twitchyliquid64: I'm not sure I follow. Both the noise and test http clients use custom transports, which wire up their own connections and hence dont look at the host part of the URL at all. The only thing that looks at the host part is the TLS verification logic, which (because the server cert is hardcoded to `example.com`) will fail for any other domain.

twitchyliquid64: Actually I just figured out a way to get to the client TLS config from the http.Client, meaning I can disable verification and we can just used `unused`.

PR #5563, opened 2022-09-07
cmd/derper: filter out useless HTTP error logs

andrew-d:
These errors aren't actionable and just fill up logs with useless data. See the following Go issue for more details:
  https://golang.org/issue/26918

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"time"
 
+	"go4.org/mem"
 	"golang.org/x/time/rate"
 	"tailscale.com/atomicfile"
 	"tailscale.com/derp"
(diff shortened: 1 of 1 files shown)


danderson (approved):

cmd/derper/derper.go
```diff
@@ -460,3 +464,21 @@ func (l *rateLimitedListener) Accept() (net.Conn, error) {
...
+// logFilter is used to filter out useless error logs that are logged to
+// the net/http.Server.ErrorLog logger.
+type logFilter struct{}
+
+func (*logFilter) Write(p []byte) (int, error) {
+	if bytes.HasSuffix(p, []byte(": EOF\n")) ||
+		bytes.HasSuffix(p, []byte(": i/o timeout\n")) ||
+		bytes.HasSuffix(p, []byte(": read: connection reset by peer\n")) ||
+		bytes.HasSuffix(p, []byte(": remote error: tls: bad certificate\n")) ||
+		bytes.HasSuffix(p, []byte(": tls: first record does not look like a TLS handshake\n")) {
```
bradfitz, line 477: I forget if this ever got faster. It used to suck with those byte slices.

I still kinda prefer mem.HasSuffix(mem.B(..), mem.S(...))

andrew-d: Tested this out in godbolt (https://godbolt.org/z/1EE7dzbKz) and verified that with byte slices, it constructs a new one on each call to this function, whereas with `mem.S` they're static. Switched to that.

bradfitz (approved):

PR #5579, opened 2022-09-08
cmd/tsconnect: enable web links addon in the terminal

### LINKED ISSUE #5247: cmd/tsconnect: special-case BrowseToURL notifications from SSH check mode
Right now they bring up a big QR code similar to the initial log in one (and do not get cleared). We should treat them differently if the IPN state is already in `Running`

mihaip:
More user friendly, and as a side-effect we handle SSH check mode better, since the URL that's output is now clickable.

Fixes #5247

### DIFF
--- a/cmd/tsconnect/package.json
+++ b/cmd/tsconnect/package.json
@@ -11,7 +11,8 @@
     "tailwindcss": "^3.1.6",
     "typescript": "^4.7.4",
     "xterm": "5.0.0-beta.58",
-    "xterm-addon-fit": "^0.5.0"
+    "xterm-addon-fit": "^0.5.0",
+    "xterm-addon-web-links": "0.7.0-beta.6"
   },
   "scripts": {
     "lint": "tsc --noEmit",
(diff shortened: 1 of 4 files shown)


cmd/tsconnect/package.json
```diff
@@ -11,7 +11,8 @@
     "tailwindcss": "^3.1.6",
     "typescript": "^4.7.4",
     "xterm": "mihaip/xterm.js#95100e3c870b59348bb4fa5504bab9f9317bac67",
-    "xterm-addon-fit": "^0.5.0"
+    "xterm-addon-fit": "^0.5.0",
+    "xterm-addon-web-links": "0.7.0-beta.6"
```
mihaip, line 15: Had to use the beta version of the addon because we're using the HEAD version of xterm.js, which has some API changes that the addon needs to pick up (xtermjs/xterm.js@7a17a26e18c9640821508cb987ddf43bb505c6dc).

phirework (approved):

PR #5586, opened 2022-09-09
cmd/tailscale: add licenses link to web UI

willnorris:
The `tailscale web` UI is the primary interface for Synology and Home Assistant users (and perhaps others), so is the logical place to put our open source license notices.  I don't love adding things to what is currently a very minimal UI, but I'm not sure of a better option.

Updates tailscale/corp#5780

Screen Shot 2022-09-09 at 12 19 55 PM

### DIFF
--- a/cmd/tailscale/cli/web.html
+++ b/cmd/tailscale/cli/web.html
@@ -11,7 +11,7 @@
 </head>
 
 <body class="py-14">
-<main class="container max-w-lg mx-auto py-6 px-8 bg-white rounded-md shadow-2xl" style="width: 95%">
+<main class="container max-w-lg mx-auto mb-8 py-6 px-8 bg-white rounded-md shadow-2xl" style="width: 95%">
 	<header class="flex justify-between items-center min-width-0 py-2 mb-8">
 		<svg width="26" height="26" viewBox="0 0 23 23" title="Tailscale" fill="none" xmlns="http://www.w3.org/2000/svg"
 			class="flex-shrink-0 mr-4">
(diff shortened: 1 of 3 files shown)


cmd/tailscale/cli/web.html
```diff
@@ -99,6 +99,9 @@ <h3 class="text-3xl font-semibold mb-3">Log in</h3>
 		<a href="#" class="mb-4 link font-medium js-loginButton" target="_blank">Reauthenticate</a>
 	</div>
 	{{ end }}
+	<div class="text-right">
```
rosszurowski, line 102: Putting it inside the white card on the right hand side makes it feel…

1. Related to the other Tailscale content inside the card
2. Prominent because of all the negative space around it

image

Instead, I think this would work better underneath the card as a little footer.

Screenshot 2022-09-12 at 12 55 02@2x

You can do this by adding this HTML block under the `</main>` line below:

```html
<footer class="container max-w-lg mx-auto text-center">
	<a class="text-xs text-gray-500 hover:text-gray-600" href="{{ .LicensesURL }}">Open Source Licenses</a>
</footer>
```

And then just add a `mb-8` to the `class` attribute on the `main` up on line 14

rosszurowski (commented):
Change is good, I'd just adjust the position/styles so its relationship to the other content is clearer.

cmd/tailscale/cli/web.html
willnorris: done

Screen Shot 2022-09-12 at 10 15 19 AM

rosszurowski (approved):
Lovely! Thanks for adding this!

Issue #2331, opened 2021-07-05
Activating tailscale kills ipv4 access to Rapberry Pi 4 (Raspberry Pi OS)


Bleistein:
We are running tailscaled in version 1.10  and 1.10.1 on a raspberrypi 4. Networking is done via wlan. If the tailscale interface is up we loose connectivity via ipv4 on the plain interface (no ping, no ssh). ipv6 access is working without problems. Issuing tailscale up/down commands toggles the problem...

DentonGentry:
One guess would be address conflict: Tailscale uses the CGNAT range of addresses, 100.64.0.0/10, which is also used by some ISPs. If your ISP uses the same range of addresses there can be a problem routing.

Otherwise, we're probably not going to be able to figure it out based on what is in the bug report. Taiscale runs on a bunch of Raspberry Pi systems without seeing symptoms like this.

If you can send either the email address you use to log in or the Tailscale IP address of the Raspberry Pi system to support@tailscale.com, we can look into it further.

Bleistein:
We have used 100.64.0.0/10 in the past for our wireguard setup. But the static routing tables in the router associated with this where inactive. We deleted them now just in case - but no change in behaviour. I sent e-mail and IP address to support. Thanks.

Bleistein:
We had a bit more time to analyse which interfaces are used for routing... 

We are trying to ping the raspberrypi machine having the tailscale interface up from a machine in the same subnet with addresses 10.0.6.31/24 and fd99:e25d:c2c:0:4100:a6d5:97c4:b1b3. The network traffic is handled by a wireless lan.

On the machine issuing the ping command packets (macOS Big Sur) to the raspberrypi are routed via the expected interface (en0 in our case) independent of using ipv4 or ipv6.

On the raspberrypi we are now checking which interfaces are used for talking back to the macos machine.

```
ip route get fd99:e25d:c2c::1ccd:a298:20a8:1511
fd99:e25d:c2c:0:1ccd:a298:20a8:1511 from :: dev wlan0 proto ra src fd99:e25d:c2c:0:4100:a6d5:97c4:b1b3 metric 303 mtu 1480 pref medium
```

Using the *ipv6* address the system we are using to ping the raspberrypi machine is reached via the wlan0 interface what is expected behaviour for our network.

```
ip route get 10.0.6.31
10.0.6.31 dev tailscale0 table 52 src 100.98.230.83 uid 1000
    cache
```

[...]

PR #5592, opened 2022-09-11
net/netns: add TS_FORCE_LINUX_BIND_TO_DEVICE for Linux

bradfitz:
For debugging a macOS-specific magicsock issue. macOS runs in bind-to-interface mode always. This lets me force Linux into the same mode as macOS, even if the Linux kernel supports SO_MARK, as it usually does.

Updates #2331 etc


bradfitz:
cc @supersat :)

maisem (approved):

Issue #1634, opened 2021-04-01
Support Hotel Wifi


frioux:
**Describe the bug**
I went to a hotel that uses open wifi, DNS interception, and an http login, and instead of `http://example.com` getting intercepted, it just never loaded.  Disabling Tailscale allowed me to log in.

**To Reproduce**
Not sure I can help with this.  If there is some tracing I can do at this hotel I can for the next ~hour, but it seemed pretty typical.

**Expected behavior**
Best case scenario tailscale could detect this situation and allow it.  Alternately maybe it could somehow let me know what's happening?  I'm not sure.

**Version information:**
 - Device: Laptop
 - OS: Linux
 - OS version: Ubuntu 20.04
 - Tailscale version: 1.6.0

**Additional context**
My phone runs tailscale and *didn't* have this problem.  It's an Android.

For fun: I'm at the Brisas Del Mar, Inn at the Beach in Santa Barbara.

bradfitz:
What's the network do without Tailscale? Enter your room number and last name and then the wifi stops being intercepted?

I'd guess we don't get a network change event in that case. Maybe we need to fall back to some polling to wait for the network to stop being a lie.

/cc @danderson @DentonGentry

frioux:
well example.com goes to some non-example.com url and asks for a password (which is on the back of the hotel card.)  I didn't look at the actual mechanism but my *guess* is that the DNS that comes from DHCP gave me a DNS server that does.  FWIW with tailscale I never got the initial MITM'd page, so it seems like TS didn't even attempt that DNS server.

danderson:
Yeah, captive portals typically redirect your HTTP traffic with lying DNS responses, to present a captive portal. Once you've signed in, it notes that your IP/MAC address is allowed and stops doing that.

To properly fix, we'd need to have tailscale do captive portal detection in the background, and do... something if that says we've captured. Not sure what though.

In the meantime, split DNS support in 1.8 will at least help with this, in the sense that you can have MagicDNS without hijacking all DNS on the system, so the captive portal would still work.

frioux:
I agree re split DNS and suspect it will almost completely resolve the issue.  If you care, here's a blog post documenting how some things detect this situation (pretty non-standard) https://www.intraway.com/blog/how-browser-identify-captive-portals

sailorfrag:
[...]

PR #5593, opened 2022-09-11
net/netcheck: add check for captive portal

andrew-d:
This doesn't change any behaviour for now, other than maybe running a full netcheck more often. The intent is to start gathering data on captive portals, and additionally, seeing this in the 'tailscale netcheck' command should provide a bit of additional information to users.

Updates #1634

### DIFF
--- a/net/netcheck/netcheck.go
+++ b/net/netcheck/netcheck.go
@@ -14,6 +14,7 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+	"math/rand"
 	"net"
 	"net/http"
 	"net/netip"
@@ -113,6 +114,10 @@ type Report struct {
 	GlobalV4 string // ip:port of global IPv4
 	GlobalV6 string // [ip]:port of global IPv6
 
+	// CaptivePortal is set when we think there's a captive portal that is
+	// intercepting HTTP traffic.
+	CaptivePortal opt.Bool
+
 	// TODO: update Clone when adding new fields
 }
 
@@ -176,6 +181,10 @@ type Client struct {
 	// If nil, portmap discovery is not done.
 	PortMapper *portmapper.Client // lazily initialized on first use
 
+	// For tests
+	testEnoughRegions      int
+	testCaptivePortalDelay time.Duration
+
 	mu       sync.Mutex            // guards following
 	nextFull bool                  // do a full region scan, even if last != nil
 	prev     map[time.Time]*Report // some previous reports
@@ -193,12 +202,23 @@ type STUNConn interface {
 }
 
 func (c *Client) enoughRegions() int {
+	if c.testEnoughRegions > 0 {
+		return c.testEnoughRegions
+	}
 	if c.Verbose {
 		// Abuse verbose a bit here so netcheck can show all region latencies
 		// in verbose mode.
 		return 100
 	}
 	return 3
+}
+
+func (c *Client) captivePortalDelay() time.Duration {
+	if c.testCaptivePortalDelay > 0 {
+		return c.testCaptivePortalDelay
+	}
+	// Chosen semi-arbitrarily
+	return 200 * time.Millisecond
 }
 
 func (c *Client) logf(format string, a ...any) {
@@ -784,13 +804,35 @@ func (c *Client) GetReport(ctx context.Context, dm *tailcfg.DERPMap) (_ *Report,
 	}
 	c.curState = rs
 	last := c.last
+
+	// Even if we're doing a non-incremental update, we may want to try our
+	// preferred DERP region for captive portal detection. Save that, if we
+	// have it.
+	var preferredDERP int
+	if last != nil {
+		preferredDERP = last.PreferredDERP
+	}
+
 	now := c.timeNow()
+
+	doFull := false
 	if c.nextFull || now.Sub(c.lastFull) > 5*time.Minute {
+		doFull = true
+	}
+	// If the last report had a captive portal and reported no UDP access,
+	// it's possible that we didn't get a useful netcheck due to the
+	// captive portal blocking us. If so, make this report a full
+	// (non-incremental) one.
+	if !doFull && last != nil {
+		doFull = !last.UDP && last.CaptivePortal.EqualBool(true)
+	}
+	if doFull {
 		last = nil // causes makeProbePlan below to do a full (initial) plan
 		c.nextFull = false
 		c.lastFull = now
 		metricNumGetReportFull.Add(1)
 	}
+
 	rs.incremental = last != nil
 	c.mu.Unlock()
 
@@ -875,6 +917,48 @@ func (c *Client) GetReport(ctx context.Context, dm *tailcfg.DERPMap) (_ *Report,
 
 	plan := makeProbePlan(dm, ifState, last)
 
+	// If we're doing a full probe, also check for a captive portal. We
+	// delay by a bit to wait for UDP STUN to finish, to avoid the probe if
+	// it's unnecessary.
+	captivePortalDone := syncs.ClosedChan()
+	captivePortalStop := func() {}
+	if !rs.incremental {
+		// NOTE(andrew): we can't simply add this goroutine to the
+		// `NewWaitGroupChan` below, since we don't wait for that
+		// waitgroup to finish when exiting this function and thus get
+		// a data race.
+		ch := make(chan struct{})
+		captivePortalDone = ch
+
+		tmr := time.AfterFunc(c.captivePortalDelay(), func() {
+			defer close(ch)
+			found, err := c.checkCaptivePortal(ctx, dm, preferredDERP)
+			if err != nil {
+				c.logf("[v1] checkCaptivePortal: %v", err)
+				return
+			}
+			rs.report.CaptivePortal.Set(found)
+		})
+
+		captivePortalStop = func() {
+			// Don't cancel our captive portal check if we're
+			// explicitly doing a verbose netcheck.
+			if c.Verbose {
+				return
+			}
+
+			if tmr.Stop() {
+				// Stopped successfully; need to close the
+				// signal channel ourselves.
+				close(ch)
+				return
+			}
+
+			// Did not stop; do nothing and it'll finish by itself
+			// and close the signal channel.
+		}
+	}
+
 	wg := syncs.NewWaitGroupChan()
 	wg.Add(len(plan))
 	for _, probeSet := range plan {
@@ -895,9 +979,17 @@ func (c *Client) GetReport(ctx context.Context, dm *tailcfg.DERPMap) (_ *Report,
 	case <-stunTimer.C:
 	case <-ctx.Done():
 	case <-wg.DoneChan():
+		// All of our probes finished, so if we have >0 responses, we
+		// stop our captive portal check.
+		if rs.anyUDP() {
+			captivePortalStop()
+		}
 	case <-rs.stopProbeCh:
 		// Saw enough regions.
 		c.vlogf("saw enough regions; not waiting for rest")
+		// We can stop the captive portal check since we know that we
+		// got a bunch of STUN responses.
+		captivePortalStop()
 	}
 
 	rs.waitHairCheck(ctx)
@@ -966,6 +1058,9 @@ func (c *Client) GetReport(ctx context.Context, dm *tailcfg.DERPMap) (_ *Report,
 		wg.Wait()
 	}
 
+	// Wait for captive portal check before finishing the report.
+	<-captivePortalDone
+
 	return c.finishAndStoreReport(rs, dm), nil
 }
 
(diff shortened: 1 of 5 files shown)


net/netcheck/netcheck.go
```diff
@@ -785,12 +792,25 @@ func (c *Client) GetReport(ctx context.Context, dm *tailcfg.DERPMap) (_ *Report,
 	c.curState = rs
 	last := c.last
 	now := c.timeNow()
+	shouldFull := false
```
maisem, line 795: doFull?

net/netcheck/netcheck.go
```diff
@@ -980,6 +1015,33 @@ func (c *Client) finishAndStoreReport(rs *reportState, dm *tailcfg.DERPMap) *Rep
...
+}
+
+func (c *Client) checkCaptivePortal(ctx context.Context) (bool, error) {
+	ctx, cancel := context.WithTimeout(ctx, captivePortalTimeout)
+	defer cancel()
+
+	req, _ := http.NewRequestWithContext(ctx, "GET", generate204URL, nil)
+	r, err := noRedirectClient.Do(req)
+	if err != nil {
+		return false, nil
```
maisem, line 1039: Return err? Or drop the error return

net/netcheck/netcheck.go
```diff
@@ -980,6 +1015,33 @@ func (c *Client) finishAndStoreReport(rs *reportState, dm *tailcfg.DERPMap) *Rep
...
+		return http.ErrUseLastResponse
+	},
+
+	// Remaining fields are the same as the default client.
+	Transport: http.DefaultClient.Transport,
+	Jar:       http.DefaultClient.Jar,
+	Timeout:   http.DefaultClient.Timeout,
+}
+
+func (c *Client) checkCaptivePortal(ctx context.Context) (bool, error) {
```
maisem, line 1032: This never returns an err

maisem: Document what the bool represents

net/netcheck/netcheck.go
andrew-d: Oops; was supposed to return that. I want to distinguish between "error making a request" and "positive response that isn't the status code we expect".

net/netcheck/netcheck.go
```diff
@@ -980,6 +1015,38 @@ func (c *Client) finishAndStoreReport(rs *reportState, dm *tailcfg.DERPMap) *Rep
...
+// checkCaptivePortal returns whether or not we think the system is behind a
+// captive portal, detected by making a request to a URL that we know should
+// return a "204 No Content" response and checking if that's what we get.
+//
+// The boolean return is whether we think we have a captive portal.
+func (c *Client) checkCaptivePortal(ctx context.Context) (bool, error) {
+	ctx, cancel := context.WithTimeout(ctx, captivePortalTimeout)
+	defer cancel()
+
+	req, _ := http.NewRequestWithContext(ctx, "GET", generate204URL, nil)
```
maisem, line 1041: Let's not throw away the error here, if someone changes this code in the future this might make the Do call panic.

maisem (approved):

maisem:
Test failures look real.

andrew-d:
> Test failures look real.

Yeah, looks like we never actually wait for the `NewWaitGroupChan` [here](https://github.com/tailscale/tailscale/pull/5593/files#diff-f9c9570d6425603afd944c122b0805ca6d8612f079142adca496bee42051ee41R898) to finish. I'll take a look at this tomorrow; I'm not confident that we actually want to wait, and I might just move the captive portal check out of that waitgroup.

andrew-d:
@maisem - Okay, had to tweak the integration tests and make the URL configurable, but... tests are good. Mind taking another look?

net/netcheck/netcheck.go
```diff
@@ -785,12 +792,25 @@ func (c *Client) GetReport(ctx context.Context, dm *tailcfg.DERPMap) (_ *Report,
...
 	if c.nextFull || now.Sub(c.lastFull) > 5*time.Minute {
+		doFull = true
+	}
+	// If the last report had a captive portal and reported no UDP access,
+	// it's possible that we didn't get a useful netcheck due to the
+	// captive portal blocking us. If so, make this report a full
+	// (non-incremental) one.
+	if !doFull && last != nil {
+		v, ok := last.CaptivePortal.Get()
+		doFull = v && ok && !last.UDP
```
bradfitz, line 805: `doFull =  !last.UDP && last.CaptivePort.EqualBool(true)`

(cheaper stuff first and then use EqualBool instead of Get)

net/netcheck/netcheck.go
```diff
@@ -980,6 +1023,45 @@ func (c *Client) finishAndStoreReport(rs *reportState, dm *tailcfg.DERPMap) *Rep
 	return report
 }
 
+var generate204URL = envknob.StringOrDefault(
+	"TS_NETCHECK_GENERATE_204_URL",
+	"http://controlplane.tailscale.com/generate_204",
```
bradfitz, line 1028: looks like this generates a 404

But I'd prefer to not hard-code this URL.

Can it not use DERP servers from the DERP map instead?

net/netcheck/netcheck.go
```diff
@@ -980,6 +1023,45 @@ func (c *Client) finishAndStoreReport(rs *reportState, dm *tailcfg.DERPMap) *Rep
...
+}
+
+// checkCaptivePortal reports whether or not we think the system is behind a
+// captive portal, detected by making a request to a URL that we know should
+// return a "204 No Content" response and checking if that's what we get.
+//
+// The boolean return is whether we think we have a captive portal.
+func (c *Client) checkCaptivePortal(ctx context.Context) (bool, error) {
+	defer noRedirectClient.CloseIdleConnections()
+	ctx, cancel := context.WithTimeout(ctx, captivePortalTimeout)
```
bradfitz, line 1050: this would usually fail on most flights nowadays with how many RTTs there are involved

(at least it's not TLS too!)

net/netcheck/netcheck.go
```diff
@@ -875,6 +895,26 @@ func (c *Client) GetReport(ctx context.Context, dm *tailcfg.DERPMap) (_ *Report,
 
 	plan := makeProbePlan(dm, ifState, last)
 
+	// If we're doing a full probe, also check for a captive portal.
```
bradfitz, line 898: I had assumed we'd only do this when UDP failed. You think STUN could work but port 80 would be intercepted?

net/netcheck/netcheck.go
andrew-d: It's possible that I'm hitting a different server from you, but...
```
$ curl -v http://controlplane.tailscale.com/generate_204
*   Trying 18.157.173.201:80...
* Connected to controlplane.tailscale.com (18.157.173.201) port 80 (#0)
> GET /generate_204 HTTP/1.1
> Host: controlplane.tailscale.com
> User-Agent: curl/7.83.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 204 No Content
< Date: Mon, 12 Sep 2022 16:17:07 GMT
< 
* Connection #0 to host controlplane.tailscale.com left intact
```

This is currently served by trunkd, but I could add the `/generate_204` endpoint to the DERPs, yeah - easy enough to do, if you'd prefer.

net/netcheck/netcheck.go
andrew-d: Not on a "good" captive portal, no, but I have no idea how less-good ones behave, and HTTP requests are pretty cheap.

net/netcheck/netcheck.go
andrew-d: Hm, good point; lemme just rely on the `overallProbeTimeout` value instead of a sub-timeout here.

net/netcheck/netcheck.go
andrew-d: https://github.com/tailscale/tailscale/pull/5601

bradfitz: Using DERPs means headscale needn't be hitting our servers

andrew-d: Okay, re-pushed to use the DERP servers. I can't merge this PR until we've deployed #5601 to all the DERPs, though, since the current behaviour is "redirect to HTTPS" and that reports as "has a captive portal", unfortunately.

bradfitz: And if people are running old DERPers? What will happen if we false positive detect a captive portal? Is the plan to turn off their Tailscale routes?

andrew-d: False-positives don't do anything right now other than cause "full" netchecks to happen more frequently, and even then, only if there's no UDP access. I'm pretty okay with that for people running old custom DERPs, TBH.

andrew-d: @bradfitz - We have now deployed all the DERPs with this route (with thanks to @DentonGentry for doing most of the work). This PR should be ready for review now.

net/netcheck/netcheck.go
```diff
@@ -113,6 +114,10 @@ type Report struct {
 	GlobalV4 string // ip:port of global IPv4
 	GlobalV6 string // [ip]:port of global IPv6
 
+	// CaptivePortal is set when we think there's a captive portal that is
```
bradfitz, line 117: so despite it being a tristate, we're only using two of its states: unset and true? Why not a bool then?

net/netcheck/netcheck.go
```diff
@@ -980,6 +1030,53 @@ func (c *Client) finishAndStoreReport(rs *reportState, dm *tailcfg.DERPMap) *Rep
...
+	req, err := http.NewRequestWithContext(ctx, "GET", "http://"+node.HostName+"/generate_204", nil)
+	if err != nil {
+		return false, err
+	}
+	r, err := noRedirectClient.Do(req)
+	if err != nil {
+		return false, err
+	}
+	c.logf("[v2] checkCaptivePortal url=%q status_code=%d", req.URL.String(), r.StatusCode)
+
```
bradfitz, line 1076: defer r.Body.Close()

net/netcheck/netcheck.go
andrew-d: We only check if it's `true` to change behaviour, but all three states are valid:

* `unset` for incremental netchecks
* `true` (during non-incremental netchecks) "made request, got non-204"
* `false` (during non-incremental netchecks) for "made request, got 204"

bradfitz: The wording of this comment is "is set when we think there's a captive portal"

I guess what does "is set" mean? I interpret it to mean a non-empty string opt.Bool: so either "true" or "false".

But "false" doesn't mean "we think there's a captive portal"

andrew-d: Ah, gotcha; update the comment in-code to be more clear.

net/netcheck/netcheck.go
```diff
@@ -980,6 +1031,54 @@ func (c *Client) finishAndStoreReport(rs *reportState, dm *tailcfg.DERPMap) *Rep
...
+// captive portal, detected by making a request to a URL that we know should
+// return a "204 No Content" response and checking if that's what we get.
+//
+// The boolean return is whether we think we have a captive portal.
+func (c *Client) checkCaptivePortal(ctx context.Context, dm *tailcfg.DERPMap, preferredDERP int) (bool, error) {
+	defer noRedirectClient.CloseIdleConnections()
+
+	// If we have a preferred DERP region with more than one node, try
+	// that; otherwise, pick a random one not marked as "Avoid".
+	if preferredDERP == 0 || (preferredDERP != 0 && len(dm.Regions[preferredDERP].Nodes) == 0) {
```
bradfitz, line 1056: `dm.Regions[preferredDERP]` might be nil in some weird cases, like derpmap changes

(especially if a user is working on their custom DERP map in their ACL file)

net/netcheck/netcheck.go
```diff
@@ -980,6 +1031,54 @@ func (c *Client) finishAndStoreReport(rs *reportState, dm *tailcfg.DERPMap) *Rep
...
+		for id, reg := range dm.Regions {
+			if reg.Avoid || len(reg.Nodes) == 0 {
+				continue
+			}
+			rids = append(rids, id)
+		}
+		preferredDERP = rids[rand.Intn(len(rids))]
+	}
+
+	node := dm.Regions[preferredDERP].Nodes[0]
```
bradfitz, line 1067: same

net/netcheck/netcheck_test.go
```diff
@@ -115,6 +115,9 @@ func TestWorksWhenUDPBlocked(t *testing.T) {
 	// OS IPv6 test is irrelevant here, accept whatever the current
 	// machine has.
 	want.OSHasIPv6 = r.OSHasIPv6
+	// Captive portal test is irrelevent; accept what the current report
```
bradfitz, line 118: typo: irrelevant

bradfitz: hah, above too

net/netcheck/netcheck.go
bradfitz: They're cheap but it's a lot of client nodes in the field. Can we at least delay the captive portal check like 50ms and abort it if we get any UDP STUN reply in those 50ms?

net/netcheck/netcheck.go
andrew-d: Good catch; added a check for whether the entry is `nil`.

net/netcheck/netcheck.go
andrew-d: This is handled above now by setting `preferredDERP` to a valid index

net/netcheck/netcheck.go
andrew-d: Added a 200ms delay, chosen semi-randomly as a value that resulted in no captive portal checks firing on my home network.

bradfitz (approved):

Issue #2452, opened 2021-07-18
Add DNS-Over-HTTPS Support for NextDNS service


sreeram-boyapati:
nextdns.io is a customizable dns service with private blocklists and crowdsourced adblocker and tracker blocklists. It is highly configurable and supports all major platforms.

Using NextDNS as DNS nameservers to use those custom blacklists and magicDNS on android phones as well as desktop is a major requirement (at least for me) to use tailscale in my workflow.

#2435 added support for Google DNS, Cloudflare. NextDNS would be another great addition. Firefox added it as the second DOH provider along with CloudFlare in 2019 [1].

#915 tracks private DNS support with tailscale in Android 12 and other versions.

1. https://www.zdnet.com/article/mozilla-to-add-second-dns-over-https-doh-provider-in-firefox/

sa1:
Seems like the recent PR maps IP addresses to hardcoded DoH urls.
https://github.com/tailscale/tailscale/blob/9ef3d6f412c2298c9e11cea5466498353c679bdf/net/dns/resolver/forwarder.go#L516

This wouldn't work for NextDNS because there's a custom url for everyone. It would help if the admin settings accept a DoH url instead of an IP.

DentonGentry:
That could get complex: depending on the capabilities of the OS, we often push the DNS configuration to the platform and not handle DNS ourselves out of Magic DNS. For example:
- recent Windows versions have a very flexible name resolution system and can handle most anything
- Android has very limited capabilities, we often handle DNS ourselves in the Tailscale app
- MacOS/iOS and Linux are somewhere in between

Accepting a URL in the DNS config means needing to look into OS capabilities for DoH.

bradfitz:
If the user specified an explicit DoH URL in their Tailnet config and the OS couldn't handle it natively (as most can't), we'd just have to wire up the host DNS config such that 100.100.100.100 was the DNS server and we'd then handle forwarding it. At least that's been what I've been assuming we'll do.

DentonGentry:
I believe that implementing https://github.com/tailscale/tailscale/issues/74 to support DNS-over-HTTPS and/or DNS-over-TLS handles this.

Duplicate of https://github.com/tailscale/tailscale/issues/74

bradfitz:
I'm going to reopen this one because I've been looking into it for other reasons and it turns out that NextDNS is unique enough in some ways to warrant & require some custom work.

We can this bug to track rather than opening a new one.

[...]

PR #5595, opened 2022-09-11
net/dns/publicdns: also support NextDNS DoH query parameters

bradfitz:
The plan has changed. Doing query parameters rather than path + heades. NextDNS added support for query parameters.

Updates #2452


bradfitz:
cc @rs

andrew-d (approved):

Issue #5602, opened 2022-09-12
wgengine/magicsock: seeing disco pings to IPv6-mapped-IPv4 addresses


bradfitz:
Noticed on Linux at HEAD (16939f0d56e4543a5f9e3),

```
ping(100.108.228.17): sending disco ping to [XvMEx] ub2004desk ...
magicsock: disco: failed to send *disco.Ping to [::ffff:209.180.x.y]:46078: write udp6 [::]:56726->[::ffff:209.180.x.y]:46078: sendto: network is unreachable
magicsock: disco: node [XvMEx] d:b508a2efe1de7f28 now using 10.2.200.239:46078
```

Note the ipv6-mapped ipv4 addresses there. I suspect something (maybe the new BPF disco stuff?) isn't unmapping somewhere.

/cc @raggi @danderson

bradfitz:
Related to STUN somewhere:

> magicsock: endpoints changed: 209.180.x.y:50451 (stun), [::ffff:209.180.x.y]:50451 (stun), 10.0.0.52:50451 (local), 10.2.0.52:50451 (local), 172.17.0.1:50451 (local), [2602:xx:...:4148]:50451 (local)

bradfitz:
It's from:

```
	if nr.GlobalV6 != "" {
		addAddr(ipp(nr.GlobalV6), tailcfg.EndpointSTUN)
	}
```

That `GlobalV6` string is literally a 4-in-6 ipport string:

```
panic: BOOM got 4 in 6 from "[::ffff:209.180.x.y]:48895"

goroutine 302 [running]:
tailscale.com/wgengine/magicsock.(*Conn).determineEndpoints.func1({0xc0008ba140, 0x1e})
        /home/bradfitz/src/tailscale.com/wgengine/magicsock/magicsock.go:1075 +0xc5
tailscale.com/wgengine/magicsock.(*Conn).determineEndpoints(0xc000158380, {0xf32bb8, 0xc0004aec40})
        /home/bradfitz/src/tailscale.com/wgengine/magicsock/magicsock.go:1116 +0x379
tailscale.com/wgengine/magicsock.(*Conn).updateEndpoints(0xc000158380, {0xdc8080, 0xe})
        /home/bradfitz/src/tailscale.com/wgengine/magicsock/magicsock.go:672 +0x1f0
created by tailscale.com/wgengine/magicsock.(*Conn).updateEndpoints.func1
[...]

PR #5603, opened 2022-09-12
net/stun: unmap IPv4 addresses in 16 byte STUN replies

bradfitz:
Updates #5602

### DIFF
--- a/net/stun/stun.go
+++ b/net/stun/stun.go
@@ -208,7 +208,7 @@ func ParseResponse(b []byte) (tID TxID, addr netip.AddrPort, err error) {
 		b = b[:attrsLen] // trim trailing packet bytes
 	}
 
-	var addr6, fallbackAddr, fallbackAddr6 netip.AddrPort
+	var fallbackAddr netip.AddrPort
 
 	// Read through the attributes.
 	// The the addr+port reported by XOR-MAPPED-ADDRESS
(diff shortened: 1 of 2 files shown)


net/stun/stun.go
```diff
@@ -223,7 +223,7 @@ func ParseResponse(b []byte) (tID TxID, addr netip.AddrPort, err error) {
 				return err
 			}
 			if len(a) == 16 {
-				addr6 = netip.AddrPortFrom(netip.AddrFrom16(*(*[16]byte)([]byte(a))), p)
+				addr6 = netip.AddrPortFrom(netip.AddrFrom16(*(*[16]byte)([]byte(a))).Unmap(), p)
```
skriptble, line 226: This works, but I think this is a little confusing that we might wind up calling something `addr6` when it is in fact an IPv4 address. Not sure that matters though.

danderson (approved):

net/stun/stun.go
bradfitz: PTAL. Removed the two `6` variables. Not sure why they were there.

net/stun/stun.go
```diff
@@ -223,7 +223,7 @@ func ParseResponse(b []byte) (tID TxID, addr netip.AddrPort, err error) {
 				return err
 			}
 			if len(a) == 16 {
-				addr6 = netip.AddrPortFrom(netip.AddrFrom16(*(*[16]byte)([]byte(a))), p)
+				addr = netip.AddrPortFrom(netip.AddrFrom16(*(*[16]byte)([]byte(a))).Unmap(), p)
```
skriptble, line 226: If we're getting rid of `addr6` then we can just do this:
```go
a, _ := netip.AddrFromSlice(slc)
addr = netip.AddrPortFrom(a.Unmap(), p)
```

skriptble (approved):

net/stun/stun.go
bradfitz: Sure, done. 

I'm trying to avoid too much cleanup with fixes at the same time, though.

There's more to be done here: the allocations from xorMappedAddress and mappedAddress are unnecessary. I can send that as a separate change, though.

skriptble: Sounds good!

PR #5605, opened 2022-09-12
cmd/tsconnect: allow hostname to be specified

mihaip:
The auto-generated hostname is nice as a default, but there are cases where the client has a more specific name that it can generate.

### DIFF
--- a/cmd/tsconnect/wasm/wasm_js.go
+++ b/cmd/tsconnect/wasm/wasm_js.go
@@ -61,24 +61,28 @@ func main() {
 func newIPN(jsConfig js.Value) map[string]any {
 	netns.SetEnabled(false)
 
-	jsStateStorage := jsConfig.Get("stateStorage")
 	var store ipn.StateStore
-	if jsStateStorage.IsUndefined() {
-		store = new(mem.Store)
-	} else {
+	if jsStateStorage := jsConfig.Get("stateStorage"); !jsStateStorage.IsUndefined() {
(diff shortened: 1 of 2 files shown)


cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -81,6 +81,14 @@ func newIPN(jsConfig js.Value) map[string]any {
 		authKey = jsAuthKey.String()
 	}
 
+	jsHostname := jsConfig.Get("hostname")
+	var hostname string
+	if jsHostname.Type() == js.TypeString {
```
phirework, line 86: Is `jsHostname` (or any of the other params) re-used? If not, wouldn't it be better for GC to declare it in the if scope like `if jsHostname := jsConfig.Get("hostname"); jsHostname == js.TypeString {`?

mihaip: Good point. I was being consistent with the other option fields (`jsAuthKey`, etc.) but they can all be rewritten in this style.

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -384,7 +390,6 @@ func (s *jsSSHSession) Run() {
 		return
 	}
 	defer sshConn.Close()
-	write("SSH Connected\r\n")
```
mihaip, line 387: Unrelated, but this is some debug logging that I happened to notice.

phirework (approved):

PR #5606, opened 2022-09-12
wgengine: set fwmark masks in netfilter & ip rules

twitchyliquid64:
Linux only.

This change masks the bitspace used when setting and querying the fwmark on packets. This allows tailscaled to play nicer with other networking software on the host, assuming the other networking software is also using fwmarks & a different mask.

IPTables / mark module has always supported masks (kversion 2.6 or something ridiculous), so this is safe on the netfilter front.

However on the `ip` utility front, busybox only gained support for parsing + setting masks in 1.33.0, so we make sure we arent such a version before we add the "/<mask>" syntax to an ip rule command. The 'OG' `ip` command from iproute2 has had support since 2016. In any case thats not hugely important, as in practice we only use the `ip` command if `TS_DEBUG_USE_IP_COMMAND=` is set.

### DIFF
--- a/wgengine/router/router_linux.go
+++ b/wgengine/router/router_linux.go
@@ -50,13 +50,21 @@ const (
 // Empirically, most of the documentation on packet marks on the
 // internet gives the impression that the marks are 16 bits
 // wide. Based on this, we theorize that the upper two bytes are
-// relatively unused in the wild, and so we consume bits starting at
-// the 17th.
+// relatively unused in the wild, and so we consume bits 16:23 (the
+// third byte).
 //
 // The constants are in the iptables/iproute2 string format for
 // matching and setting the bits, so they can be directly embedded in
 // commands.
 const (
+	// The mask for reading/writing the 'firewall mask' bits on a packet.
+	// See the comment on the const block on why we only use the third byte.
+	//
+	// We claim bits 16:23 entirely. For now we only use the lower four
+	// bits, leaving the higher 4 bits for future use.
+	tailscaleFwmarkMask    = "0xff0000"
+	tailscaleFwmarkMaskNum = 0xff0000
+
 	// Packet is from Tailscale and to a subnet route destination, so
 	// is allowed to be routed through this machine.
 	tailscaleSubnetRouteMark = "0x40000"
@@ -104,6 +112,7 @@ type linuxRouter struct {
 	ipRuleAvailable bool // whether kernel was built with IP_MULTIPLE_TABLES
 	v6Available     bool
 	v6NATAvailable  bool
+	fwmaskWorks     bool // whether we can use 'ip rule...fwmark <mark>/<mask>'
 
 	// ipPolicyPrefBase is the base priority at which ip rules are installed.
 	ipPolicyPrefBase int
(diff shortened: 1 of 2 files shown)


wgengine/router/router_linux.go
```diff
@@ -39,6 +40,11 @@ const (
 	netfilterOn       = preftype.NetfilterOn
 )
 
+// busyboxVersionRE extracts the major, minor, and patch versions
+// from a busybox banner string.
+// See: https://regex101.com/r/XGoKtD/1 for analysis.
+var busyboxVersionRE = regexp.MustCompile("(?m)^BusyBox\\Wv([0-9]+)\\.([0-9]+)\\.([0-9]+)\\W")
```
bradfitz, line 46: use raw string literals for regexps.

But we've generally tried to avoid the regexp package for binary size reasons because it brings in huge unicode tables for matching e.g. `\p{Greek}`. If you must, though. I guess it's Linux-only.

wgengine/router/router_linux_test.go
```diff
@@ -811,3 +811,79 @@ func TestCheckIPRuleSupportsV6(t *testing.T) {
...
+   or: busybox --show SCRIPT
+   or: busybox --install [-s] [DIR]
+   or: function [arguments]...
+
+	BusyBox is a multi-call binary that combines many common Unix
+	utilities into a single executable.  Most people will create a
+	link to busybox for each function they wish to use and BusyBox
+	will act like whatever it was invoked as.
+
+Currently defined functions:
```
bradfitz, line 832: little unnecessary to include all these commands

wgengine/router/router_linux.go
twitchyliquid64: I used it as its already used for DNS INI parsing, but i missed that code was conditional for windows. I can re-write it into string scanning/indexing/splitting methods if thats preferable? should be pretty straightforward.

wgengine/router/router_linux.go
```diff
@@ -206,6 +234,75 @@ func newUserspaceRouterAdvanced(logf logger.Logf, tunname string, linkMon *monit
...
+	case major > 1:
+		return true, nil
+	case major == 1 && minor >= 33:
+		return true, nil
+	default:
+		return false, nil
+	}
+}
+
+func busyboxParseVersion(output string) (major, minor, patch int, err error) {
```
danderson, line 288: Without regex:

```
fs := strings.Fields(output)
if len(fs) < 2 { err... }
if fs[0] != "BusyBox" { err... }
fs = strings.Split(strings.TrimPrefix(fs[1], "v"), ".")
if len(fs) != 3 { err... }
// atoi etc. here
```

danderson (approved):

lanrat:
> /cc @lanrat when you get a chance can you try this patch on your openWRT system, and see if you get `ip rule` output like `from all fwmark 0x80000/0xff0000 lookup main` ?

Output of `ip rule` with this patch.

```sh
root@GL-AR750S:~# ip rule
0:	from all lookup local
1001:	from all iif eth0.2 lookup 1
1002:	from all iif wlan-sta lookup 2
1310:	from all fwmark 0x80000/0xff0000 lookup main
1330:	from all fwmark 0x80000/0xff0000 lookup default
1350:	from all fwmark 0x80000/0xff0000 unreachable
1370:	from all lookup 52
2001:	from all fwmark 0x100/0x3f00 lookup 1
2002:	from all fwmark 0x200/0x3f00 lookup 2
2061:	from all fwmark 0x3d00/0x3f00 blackhole
2062:	from all fwmark 0x3e00/0x3f00 unreachable
32766:	from all lookup main
32767:	from all lookup default
```

lanrat:
Exit node behavior still does not seem to work.

twitchyliquid64:
Lets do another debug sesh to work out whats up with the exit node functionality separately - I'm guessing it something specific to openWRT.

PR #5611, opened 2022-09-12
wgengine/magicsock: move firstDerp check after nil derpMap check

kylecarbs:
This fixes a race condition which caused `c.muCond.Broadcast()` to never fire in the `firstDerp` if block. It resulted in `Close()` hanging forever.


wgengine/magicsock/magicsock.go
```diff
@@ -1356,15 +1356,15 @@ func (c *Conn) derpWriteChanOfAddr(addr netip.AddrPort, peer key.NodePublic) cha
 	}
 	c.logf("magicsock: adding connection to derp-%v for %v", regionID, why)
```
bradfitz, line 1357: Little weird that this log message will be printed just to ... not do that and return nil instead.

Why not the three lines up even early?

But how does this situation even arise? A test would be nice if it's not too much work.

kylecarbs: I moved the few lines up.... seems reasonable to me.

Crazy enough I'm not sure how to replicate this. I ran into it when _seemingly_ using the library properly, but obviously not. Somehow the derp mapping or region provided becomes nil, but I don't know how to get it in that state. I can write a test that just mock-sets the values, but that seems to miss the point somewhat.

PR #5613, opened 2022-09-12
tka: implement machinery for node-key denylist

twitchyliquid64:


tka/aum.go
```diff
@@ -161,14 +183,14 @@ func (a *AUM) StaticValidate() error {
 		if a.Key == nil {
 			return errors.New("AddKey AUMs must contain a key")
 		}
-		if a.KeyID != nil || a.State != nil || a.Votes != nil || a.Meta != nil {
+		if a.KeyID != nil || a.State != nil || a.Votes != nil || a.Meta != nil || a.NodeKey != nil {
```
sailorfrag, line 186: This is getting cumbersome, I wonder if we want to change the `AUM` struct to use some other approach to varying the fields for each `AUMKind`

I almost wonder if something like this would be good:
```golang
type AUM struct {
  MessageKind AUMKind
  PrevAUMHash []byte
  MessageParams []byte // or is there a cbor.RawMessage?
}
```

And then `MessageParams` would be one of several possible struct types depending on what `MessageKind` says.

tka/aum.go
```diff
@@ -70,6 +71,16 @@ const (
...
+	// A AddDenylistNodeKey AUM describes the addition of a node key to
+	// the denylist.
+	//
+	// Only the NodeKey optional field may be set.
+	AUMAddDenylistNodeKey
+	// A RemoveDenylistNodeKey AUM describes the removal of a node key from
+	// the denylist.
+	//
+	// Only the NodeKey optional field may be set.
+	AUMRemoveDenylistNodeKey
```
sailorfrag, line 83: Would it be better to make the deny list append-only? I guess that makes the limit problematic.

tka/state.go
```diff
@@ -193,6 +207,32 @@ func (s State) applyVerifiedAUM(update AUM) (State, error) {
...
+			if bytes.Equal(update.NodeKey, s.BannedNodekeys[i]) {
+				idx = i
+				break
+			}
+		}
+		if idx < 0 {
+			return State{}, errors.New("no such entry")
+		}
+		out := s.cloneForUpdate(&update)
+		out.BannedNodekeys = append(out.BannedNodekeys[:idx], out.BannedNodekeys[idx+1:]...)
```
sailorfrag, line 233: Would it be better to do the O(1) (but not order-preserving) "swap with the last element" strategy rather than a O(n) shift? I suppose it's already O(n) with the scan, so maybe that doesn't matter.

tka/state.go
```diff
@@ -205,6 +245,7 @@ func (s State) applyVerifiedAUM(update AUM) (State, error) {
 const (
 	maxDisablementSecrets = 32
 	maxKeys               = 512
+	maxDenylistEntries    = 64
```
sailorfrag, line 248: I feel like this makes the node deletion flow kind of awkward, in that sometimes you'll hit the deny list limit and have to rotate your signing key, and in the event that there's more than one signing key, that might mean being forced to revoke one of the other signing keys. I'm not sure how to do that without making the node creation flow more burdensome.

sailorfrag (approved):

tka/aum.go
twitchyliquid64: I originally opted not to nest message subtypes because that form usually lends itself to accidental type confusion issues. Additionally, I'm hesitant to change the layout of the AUM struct now as it would represent a massive refactor.

If the issue is mostly how cumbersome the checks are, what if we used reflection to invert it? So instead of listing the fields that arent allowed, we list the fields that are?

```go
if !a.OnlySet("NodeKey") {
  return errors.New("only NodeKey may be set on AddDenylistNodeKey AUMs")
}
```

wdyt?

tka/state.go
twitchyliquid64: Ooooo, what is the swap with last element strat? 

```go
out.BannedNodeKeys[idx] = out.BannedNodeKeys[len(out.BannedNodeKeys)-1]
out.BannedNodeKeys = out.BannedNodeKeys[:len(out.BannedNodeKeys)-2]
```

or such? does that work for all values of `idx`?

tka/aum.go
twitchyliquid64: The larger the state gets, the larger each checkpoint AUM will get, which will not trend in ideal directions. I think in general we should try and keep append-only data out of TKA state, but I can see how a revocation list is kind of hard to clean up.

tka/state.go
twitchyliquid64: Yeah, I haven't really thought through how this works from a UX perspective. Im thinking that even if we do decide to merge this, we should probably keep this feature fairly hidden and only mention it to users who are stuck in a bind & use this primitive.

I pulled the `64` entries limit out of thin air: It corresponds to about 2.5kb of data. Perhaps we can bump it up much higher?

tka/aum.go
sailorfrag: Honestly the more I think this through, the more I'm feeling like a revocation list maybe doesn't really belong in the TKA state; it doesn't fit the data model very well.

If we're OK with retaining the extra complexity for a feature we might not use in future, I guess leaving this with the 64k max is maybe not horrible? Alternatively, rather than allowing selective removal maybe only support add and clear, or even simply "replace"? I'm having a hard time coming up with a reason to remove a single key, whereas if the signing keys are all rotated out, then "clear" seems reasonable enough.

tka/state.go
sailorfrag: Yeah, that's the trick. I think the second line is just `len-1` though.

If `len(out.BannedNodeKeys) == 1` then skip the first assignment (the only element is the one being deleted).

tka/state.go
sailorfrag: I think any fixed limit such that the number of free entries is less than the number of devices in the tailnet is going to have limited utility. I guess 64 is big enough to make it useful for an emergency push for a handful of devices being compromised; anything bigger than that and you really need to rotate the signing keys + re-sign everything.

twitchyliquid64:
Yeah, I'm also starting to think storing the actual revocation list isnt a good fit for TKA.

One thing that could be: Storing the URL to a revocation list? The glue would be more complex, but that way we would give something of a CRL without ballooning the size of the state machine.

sailorfrag:
> One thing that could be: Storing the URL to a revocation list? The glue would be more complex, but that way we would give something of a CRL without ballooning the size of the state machine.

That sounds like a bunch of annoying client complexity (and should control check the CRL regularly?) but yeah I think that's probably the best approach I see. It might be worth running the idea by a security person first. Is it worth putting the code in for that upfront or could it be added later without confusing older clients?

twitchyliquid64:
Yeah, I definitely think we can do that later. Old clients can ignore the AUMs for updating the CRL and ignore the field in the state they read from a checkpoint. The only difference is old clients might accept a banned node key, which isnt ideal but wouldnt affect consensus.

twitchyliquid64:
Echoing some messages in slack: we're in agreement we shouldnt store a full CRL in tka state, and the complexity of fetching a CRL URL is non-trivial (handling the case where the endpoint you are reaching for the CRL is revoked, fail open/secure, concurrency when fetching, how often to update etc).

For now, we'll leave this to the future. It should be possible to implement this as a new field + AUMs that old clients will safely ignore.

Issue #5114, opened 2022-07-21
FR: Make it easier for tailscaled to set --port at startup


dblohm7:
### What are you trying to do?

This came out of a support issue. The user has an uncooperative firewall and ended up running `tailscaled --port` manually. They would like `tailscaled` to use the `--port` option every time it starts.

This is not very intuitive to set up on Windows (though entirely possible if you know the right commands).

### How should we solve this?

Their feature request is to be able to save the `--port` setting persistently (either via command line, GUI, or both) so that tailscaled always uses that preferred setting when it starts.

### What is the impact of not solving this?

In this case, the user disabled `tailscaled`'s entry in the service control manager and is instead using the Windows Task Scheduler to start `tailscaled` automatically with their preferred port configuration.

DentonGentry:
I think it doesn't store the port in the config now because the config is opened too late. We could try to open the config earlier, or maybe close and reopen the port in question after we've read the config.

Or I guess add a registry key for the Windows Service port number, though that seems oddly platform specific.

bradfitz:
I thought we already had support for reading a file of env vars or flags from disk for Windows and/or macOS to approximate the Linux `/etc/default/tailscaled` story. But I sure can't find it. Maybe we'd just talked about it.

bradfitz:
> I thought we already had support for reading a file of env vars or flags from disk for Windows and/or macOS to approximate the Linux /etc/default/tailscaled story. But I sure can't find it. Maybe we'd just talked about it.

Ah, yes, I'd done a lot of this already in b03170b901b1bb89f815e589750a76ea4a2254bf for #3688

But I guess it doesn't work with the envknob package globals, so I'll tie that and 74674b110daeb2cdda6f76ad3d69ef4cfba6e5b9 together.

/cc @andrew-d

bradfitz:
Seems like this happened in 1.32.0.

Put `PORT=1234` in `C:\ProgramData\Tailscale\tailscaled-env.txt` and restart the Tailscale service (`net stop Tailscale`, `net start Tailscale`)

misaligar:
https://tailscale.com/kb/1011/log-mesh-traffic?tab=macos this page links here for disabling logs using Tailscale on macOS. I don't understand what I should be doing here. Can someone please help?

vickunwu:
[...]

PR #5618, opened 2022-09-13
cmd/tailscaled, tailcfg, hostinfo: add flag to disable logging + support

### LINKED ISSUE #5617: FR: Opt out of client log upload

I'm trying to disable the upload of client logs to log.tailscale.io

### How should we solve this?

Provide an option to disable remote logging

Currently there is no way to effectively opt-out of log uploads. On https://tailscale.com/security/ you say:

> You can block [log.tailscale.io](https://tailscale.com/kb/1011/log-mesh-traffic/) on your clients to prevent logging client logs. If you do so, Tailscale may not be able to provide technical support.

However, since https://github.com/tailscale/tailscale/pull/3333 was merged, this is no longer possible using DNS. Blocking IPs is even more brittle.

### Anything else?

_No response_

bradfitz:
As noted in #5617, our documented method of blocking log.tailscale.io DNS no longer works due to bootstrap DNS.

Instead, provide an explicit flag (--no-logs-no-support) and/or env variable (TS_NO_LOGS_NO_SUPPORT=true) to explicitly disable logcatcher uploads. It also sets a bit on Hostinfo to say that the node is in that mode so we can end any support tickets from such nodes more quickly.

This does not yet provide an easy mechanism for users on some platforms (such as Windows, macOS, Synology) to set flags/env. On Linux you'd used /etc/default/tailscaled typically. Making it easier to set flags for other platforms is tracked in #5114.

Fixes #5617
Fixes tailscale/corp#1475

### DIFF
--- a/logpolicy/logpolicy.go
+++ b/logpolicy/logpolicy.go
@@ -539,7 +539,10 @@ func New(collection string) *Policy {
 		conf.IncludeProcSequence = true
 	}
 
-	if val := getLogTarget(); val != "" {
+	if envknob.NoLogsNoSupport() {
+		log.Println("You have disabled logging. Tailscale will not be able to provide support.")
+		conf.HTTPC = &http.Client{Transport: noopPretendSuccessTransport{}}
+	} else if val := getLogTarget(); val != "" {
 		log.Println("You have enabled a non-default log target. Doing without being told to by Tailscale staff or your network administrator will make getting support difficult.")
 		conf.BaseURL = val
 		u, _ := url.Parse(val)
@@ -734,4 +737,15 @@ func goVersion() string {
 		return v + "-race"
 	}
 	return v
+}
+
+type noopPretendSuccessTransport struct{}
+
+func (noopPretendSuccessTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	io.ReadAll(req.Body)
+	req.Body.Close()
+	return &http.Response{
+		StatusCode: 200,
+		Status:     "200 OK",
+	}, nil
 }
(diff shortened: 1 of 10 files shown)


DentonGentry (approved):
Also fixes https://github.com/tailscale/corp/issues/1475

logpolicy/logpolicy.go
```diff
@@ -735,3 +738,14 @@ func goVersion() string {
...
+
+type noopPretendSuccessTransport struct{}
+
+func (noopPretendSuccessTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	io.ReadAll(req.Body)
+	return &http.Response{
+		StatusCode: 200,
+		Status:     "200 OK",
+	}, nil
+
```
andrew-d, line 750: Extra newline

logpolicy/logpolicy.go
```diff
@@ -735,3 +738,14 @@ func goVersion() string {
 	}
 	return v
 }
+
+type noopPretendSuccessTransport struct{}
+
+func (noopPretendSuccessTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	io.ReadAll(req.Body)
```
andrew-d, line 745: IIRC the `RoundTripper` contract is to close the body? From the docs:

> 	// RoundTrip should not modify the request, except for
>	// consuming and closing the Request's Body. RoundTrip may

mayakacz:
To clarify: does this block only `log.tailscale.io` or other logging as well?

logpolicy/logpolicy.go
bradfitz: Done. Although in this case it doesn't matter (for now!) because all our Close methods are no-ops. But good for the future.

andrew-d (approved):

PR #5620, opened 2022-09-13
docker: add ability to use a custom control socket

iSchluff:
Allows for specifying a custom control socket in the docker entrypoint, which is useful for sharing it with sidecar containers


bradfitz:
Leaving for @maisem.

docs/k8s/run.sh
```diff
@@ -81,11 +82,11 @@ if [[ ! -z "${TS_EXTRA_ARGS}" ]]; then
 fi
 
 echo "Running tailscale up"
-tailscale --socket=/tmp/tailscaled.sock up ${UP_ARGS}
+tailscale --socket=${TS_SOCKET} up ${UP_ARGS}
```
maisem, line 85: Add quotes here and below

maisem: Oh I guess you don't need them below. Only here is fine

iSchluff: resolved

maisem (approved):

maisem:
Thanks!

PR #5634, opened 2022-09-14
portlist: normalise space delimited process names

### LINKED ISSUE #3781: Normalize ssh service descriptions to suppress "N of 10-100 startups"
I saw a Linux node updating its Services from:

`"Services":[{"Proto":"tcp","Port":22,"Description":"sshd_config [listener] 1 of 10-100 startups"}`
to
`"Services":[{"Proto":"tcp","Port":22,"Description":"sshd_config [listener] 0 of 10-100 startups"}`

(note the 1 to 0 change)

We should normalize that so we don't send such noisy updates to control.

kradalby:
Some process names are not only separated by NUL and contains spaces. This PR will check for these special cases and extract the base process name.

This fixes #3781 and other similar cases:

`sshd_config [listener] 1 of 10-100 startups` -> `sshd_config`

`nginx worker` -> `nginx`

Not sure if this approach might have bad performance impact, and that it might be better to tackle this in the Linux specific code where we split by `\x00`.

### DIFF
--- a/portlist/clean.go
+++ b/portlist/clean.go
@@ -26,6 +26,9 @@ func argvSubject(argv ...string) string {
 		ret = filepath.Base(argv[1])
 	}
 
+	// Handle space separated argv
+	ret, _, _ = strings.Cut(ret, " ")
+
 	// Remove common noise.
 	ret = strings.TrimSpace(ret)
 	ret = strings.TrimSuffix(ret, ".exe")
--- a/portlist/clean_test.go
+++ b/portlist/clean_test.go
@@ -31,6 +31,22 @@ func TestArgvSubject(t *testing.T) {
 			in:   []string{"/bin/mono", "/sbin/exampleProgram.bin"},
 			want: "exampleProgram.bin",
 		},
+		{
+			in:   []string{"/usr/bin/sshd_config [listener] 1 of 10-100 startups"},
+			want: "sshd_config",
+		},
+		{
+			in:   []string{"/usr/bin/sshd [listener] 0 of 10-100 startups"},
+			want: "sshd",
+		},
+		{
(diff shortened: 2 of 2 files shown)


portlist/clean.go
```diff
@@ -24,6 +24,12 @@ func argvSubject(argv ...string) string {
 	case ret == "mono" && len(argv) >= 2:
 		// .Net programs execute as `mono actualProgram.exe`.
 		ret = filepath.Base(argv[1])
+	case strings.Contains(ret, " "):
+		argv2 := strings.Split(ret, " ")
```
bradfitz, line 28: Split allocates. Use two strings.Cut instead?

bradfitz:
Remove `/clean` from the commit message. Just the package name, not the file.

portlist/clean.go
kradalby: Oh, I could put it straight into `ret`, I suppose it will always be found.

portlist/clean.go
```diff
@@ -24,6 +24,8 @@ func argvSubject(argv ...string) string {
 	case ret == "mono" && len(argv) >= 2:
 		// .Net programs execute as `mono actualProgram.exe`.
 		ret = filepath.Base(argv[1])
+	case strings.Contains(ret, " "):
```
bradfitz, line 27: technically this is unnecessary. strings.Cut is defined to return its input unchanged if it doesn't contain the substring. So this could be after the switch with just `ret, _, _ = strings.Cut(ret, " ")`

bradfitz (approved):

portlist/clean.go
kradalby: neat

PR #5638, opened 2022-09-14
wgengine/magicsock: add client flag and envknob to disable heartbeat

phirework:
My notes from uhhhh six weeks ago aren't as clear as I remembered them being so this could be totally off base.

Change-Id: Idc9a72748e74145b068d67e6dd4a4ffe3932efd0

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -1456,6 +1456,10 @@ type Debug struct {
 	// re-enabled for the lifetime of the process.
 	DisableLogTail bool `json:",omitempty"`
 
+	// EnableSilentDisco disables the use of heartBeatTimer in magicsock and attempts to
+	// handle disco silently. See issue #540 for details.
+	EnableSilentDisco bool `json:",omitempty"`
+
 	// Exit optionally specifies that the client should os.Exit
 	// with this code.
 	Exit *int `json:",omitempty"`
--- a/wgengine/magicsock/debugknobs.go
+++ b/wgengine/magicsock/debugknobs.go
@@ -35,6 +35,9 @@ var (
 	debugReSTUNStopOnIdle = envknob.RegisterBool("TS_DEBUG_RESTUN_STOP_ON_IDLE")
 	// debugAlwaysDERP disables the use of UDP, forcing all peer communication over DERP.
 	debugAlwaysDERP = envknob.RegisterBool("TS_DEBUG_ALWAYS_USE_DERP")
+	// debugEnableSilentDisco disables the use of heartbeatTimer on the endpoint struct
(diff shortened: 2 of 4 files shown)


bradfitz:
This provides a way to turn off the heartbeat, but it doesn't replace it with any other disco pings at any stage. We also need to occasionally convince ourselves we're using the right path when there's existing traffic.

We could do this in a bunch of little commits as long as they all reference a tracking bug and the commit messages clarify where it's at.

tailcfg/tailcfg.go
```diff
@@ -79,7 +79,8 @@ type CapabilityVersion int
 //   - 40: 2022-08-22: added Node.KeySignature, PeersChangedPatch.KeySignature
 //   - 41: 2022-08-30: uses 100.100.100.100 for route-less ExtraRecords if global nameservers is set
 //   - 42: 2022-09-06: NextDNS DoH support; see https://github.com/tailscale/tailscale/pull/5556
-const CurrentCapabilityVersion CapabilityVersion = 42
+//   - 43: 2022-09-14: adds Debug.EnableSilentDisco
+const CurrentCapabilityVersion CapabilityVersion = 43
```
bradfitz, line 83: I don't think we want to bump this quite yet as it doesn't work yet.

Control can send this bit blindly to nodes regardless of whether they actually support it and old nodes will ignore it, so I don't think we want to advertise this until it works. Meanwhile you can make control send it to just your node during testing. (or really, use the env var)

wgengine/magicsock/debugknobs.go
```diff
@@ -35,6 +35,9 @@ var (
 	debugReSTUNStopOnIdle = envknob.RegisterBool("TS_DEBUG_RESTUN_STOP_ON_IDLE")
 	// debugAlwaysDERP disables the use of UDP, forcing all peer communication over DERP.
 	debugAlwaysDERP = envknob.RegisterBool("TS_DEBUG_ALWAYS_USE_DERP")
+	// debugEnableSilentDisco disables the use of heartbeatTimer on the endpoint struct
+	// and attempts to
```
bradfitz, line 39: unfinished sentence

tailcfg/tailcfg.go
```diff
@@ -1456,6 +1457,10 @@ type Debug struct {
 	// re-enabled for the lifetime of the process.
 	DisableLogTail bool `json:",omitempty"`
 
+	// EnableSilentDisco disables the use of heartBeatTimer in magicsock and attempts to
+	// handle disco silently.
```
bradfitz, line 1461: add "See issue 540." here and elsewhere?

bradfitz (approved):

PR #5639, opened 2022-09-15
net/speedtest: retune to meet iperf on localhost in a VM

raggi:
- removed some in-flow time calls
- increase buffer size to 2MB to overcome syscall cost
- move relative time computation from record to report time

### DIFF
--- a/net/speedtest/speedtest.go
+++ b/net/speedtest/speedtest.go
@@ -11,11 +11,11 @@ import (
 )
 
 const (
-	blockSize       = 32000                 // size of the block of data to send
+	blockSize       = 2 * 1024 * 1024       // size of the block of data to send
 	MinDuration     = 5 * time.Second       // minimum duration for a test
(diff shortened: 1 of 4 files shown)


net/speedtest/speedtest.go
```diff
@@ -37,26 +37,26 @@ type configResponse struct {
...
 }
 
 func (r Result) MBitsPerSecond() float64 {
-	return r.MegaBits() / (r.IntervalEnd - r.IntervalStart).Seconds()
+	return r.MegaBits() / r.IntervalEnd.Sub(r.IntervalStart).Seconds()
 }
 
 func (r Result) MegaBytes() float64 {
-	return float64(r.Bytes) / 1000000.0
+	return float64(r.Bytes) / 1024 / 1024
```
jwhited, line 51: nit: if this is network bandwidth context we could keep it as mega instead of mebi

jwhited (approved):

PR #5644, opened 2022-09-15
Remove /var/lib mount recommendation

GrigoriyMikhalkin:
Can we remove `/var/lib` mount recommendation for starting `taislcaled` in container? It looks unnecessary and excessive security wise. Is there good reason for doing that when running a container? At least in all my scenarios i was able to run `tailscaled` without mounting it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@
 #
 # To run the tailscaled agent:
 #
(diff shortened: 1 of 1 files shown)


DentonGentry:
Removing the /var/lib mount just means there is no persistent data storage, so every time you start the container it will need to authenticate again. The Dockerfile is currently structured so you'll have one, consistent node which restarts as needed but keeps the same IP address and hostname.

PR #5645, opened 2022-09-15
Implement rate limiting on a per-client basis for DERP

twitchyliquid64:
This PR implements an experimental control loop for a per-client throughput ratelimit, based on egress utilization of the network interface and an open control loop.

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -57,6 +57,11 @@ var (
 
 	acceptConnLimit = flag.Float64("accept-connection-limit", math.Inf(+1), "rate limit for accepting new connection")
 	acceptConnBurst = flag.Int("accept-connection-burst", math.MaxInt, "burst limit for accepting new connection")
+
+	egressInterface = flag.String("egress-interface", "", "the interface to monitor for automatic ratelimit tuning")
+	egressDataLimit = flag.Int("egress-data-limit", 100*1024*1024/8, "the bandwidth in bytes/s the server will try to stay under, only applies if egress-interface is set")
+	clientDataMin   = flag.Int("client-data-min-limit", 1024*1024/8, "minimum bandwidth in bytes/s for a single client, only applies if egress-interface is set")
+	clientDataBurst = flag.Int("client-data-burst", 3*1024*1024, "burst limit in bytes for forwarded data from a single client, only applies if egress-interface is set")
 )
 
 var (
@@ -153,6 +158,12 @@ func main() {
 
 	s := derp.NewServer(cfg.PrivateKey, log.Printf)
 	s.SetVerifyClient(*verifyClients)
+
+	if *egressInterface != "" && *egressDataLimit > 0 {
+		if err := s.StartEgressRateLimiter(*egressInterface, *egressDataLimit, *clientDataMin, *clientDataBurst); err != nil {
+			log.Fatalf("failed to start egress rate limiter: %v", err)
+		}
+	}
 
 	if *meshPSKFile != "" {
 		b, err := ioutil.ReadFile(*meshPSKFile)
--- a/derp/derp_server.go
+++ b/derp/derp_server.go
@@ -107,6 +107,9 @@ type Server struct {
 	metaCert    []byte // the encoded x509 cert to send after LetsEncrypt cert+intermediate
 	dupPolicy   dupPolicy
 
+	clientDataLimit *uint64 // limit for how many bytes/s of content a client can send; atomic
+	clientDataBurst int     // burst limit for how many bytes/s of content a client can send
+
 	// Counters:
 	packetsSent, bytesSent       expvar.Int
 	packetsRecv, bytesRecv       expvar.Int
@@ -314,7 +317,10 @@ func NewServer(privateKey key.NodePrivate, logf logger.Logf) *Server {
 		sentTo:               map[key.NodePublic]map[key.NodePublic]int64{},
 		avgQueueDuration:     new(uint64),
 		keyOfAddr:            map[netip.AddrPort]key.NodePublic{},
+		clientDataLimit:      new(uint64),
+		clientDataBurst:      10 * 1024 * 1024, // 10Mb default burst
 	}
+	atomic.StoreUint64(s.clientDataLimit, 12*1024*1024) // 12Mb/s default ratelimit
 	s.initMetacert()
 	s.packetsRecvDisco = s.packetsRecvByKind.Get("disco")
 	s.packetsRecvOther = s.packetsRecvByKind.Get("other")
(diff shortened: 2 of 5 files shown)


cmd/derper/derper.go
```diff
@@ -57,6 +57,9 @@ var (
 
 	acceptConnLimit = flag.Float64("accept-connection-limit", math.Inf(+1), "rate limit for accepting new connection")
 	acceptConnBurst = flag.Int("accept-connection-burst", math.MaxInt, "burst limit for accepting new connection")
+
+	clientDataLimit = flag.Int("client-data-limit", 4*1024*1024, "rate limit for forwarded data from a single client")
+	clientDataBurst = flag.Int("client-data-burst", 10*1024*1024, "burst limit for forwarded data from a single client")
```
bradfitz, line 62: clarify what the unit is

derp/derp_server.go
```diff
@@ -294,7 +297,10 @@ type Conn interface {
 
 // NewServer returns a new DERP server. It doesn't listen on its own.
 // Connections are given to it via Server.Accept.
-func NewServer(privateKey key.NodePrivate, logf logger.Logf) *Server {
+//
+// clientDataLimit and clientDataBurst configure per-client token-bucket
+// rate/burst limits, in bytes of Wireguard traffic per second.
+func NewServer(privateKey key.NodePrivate, logf logger.Logf, clientdataLimit, clientDataBurst int) *Server {
```
bradfitz, line 303: rather than extra positional parameters (especially easily-transposable ones of the same type!), I'd prefer a method on Server to set this.

cmd/derper/derper.go
```diff
@@ -57,6 +57,11 @@ var (
 
 	acceptConnLimit = flag.Float64("accept-connection-limit", math.Inf(+1), "rate limit for accepting new connection")
 	acceptConnBurst = flag.Int("accept-connection-burst", math.MaxInt, "burst limit for accepting new connection")
+
+	egressInterface = flag.String("egress-interface", "", "the interface to monitor for automatic ratelimit tuning")
```
danderson, line 61: Why do we need to monitor specific interfaces? I'd have expected us to measure all non-mesh traffic directly.

cmd/derper/derper.go
```diff
@@ -57,6 +57,11 @@ var (
...
+	egressDataLimit = flag.Int("egress-data-limit", 100*1024*1024/8, "the bandwidth in bytes/s the server will try to stay under, only applies if egress-interface is set")
+	clientDataMin   = flag.Int("client-data-min-limit", 1024*1024/8, "minimum bandwidth in bytes/s for a single client, only applies if egress-interface is set")
```
danderson, line 63: How do clientDataMin and egressDataLimit interact? If I DoS a DERP server with clients, can I make the server's owner spend more money than egressDataLimit?

cmd/derper/derper.go
```diff
@@ -57,6 +57,11 @@ var (
 
 	acceptConnLimit = flag.Float64("accept-connection-limit", math.Inf(+1), "rate limit for accepting new connection")
 	acceptConnBurst = flag.Int("accept-connection-burst", math.MaxInt, "burst limit for accepting new connection")
+
+	egressInterface = flag.String("egress-interface", "", "the interface to monitor for automatic ratelimit tuning")
+	egressDataLimit = flag.Int("egress-data-limit", 100*1024*1024/8, "the bandwidth in bytes/s the server will try to stay under, only applies if egress-interface is set")
```
danderson, line 62: Default to 200, since we just changed prod to that.

derp/derp_server.go
```diff
@@ -107,6 +107,9 @@ type Server struct {
 	metaCert    []byte // the encoded x509 cert to send after LetsEncrypt cert+intermediate
 	dupPolicy   dupPolicy
 
+	clientDataLimit *uint64 // limit for how many bytes/s of content a client can send; atomic
```
danderson, line 110: Use atomic.Int64 ? Can't be misused non-atomically, otherwise identical.

derp/derp_server.go
```diff
@@ -314,7 +317,10 @@ func NewServer(privateKey key.NodePrivate, logf logger.Logf) *Server {
 		sentTo:               map[key.NodePublic]map[key.NodePublic]int64{},
 		avgQueueDuration:     new(uint64),
 		keyOfAddr:            map[netip.AddrPort]key.NodePublic{},
+		clientDataLimit:      new(uint64),
+		clientDataBurst:      10 * 1024 * 1024, // 10Mb default burst
```
danderson, line 321: There's a flag for this, I don't think it should be hardcoded?

derp/derp_server.go
```diff
@@ -314,7 +317,10 @@ func NewServer(privateKey key.NodePrivate, logf logger.Logf) *Server {
...
 	}
+	atomic.StoreUint64(s.clientDataLimit, 12*1024*1024) // 12Mb/s default ratelimit
```
danderson, line 323: Same?

derp/limiter.go
```diff
@@ -0,0 +1,171 @@
...
+
+import (
+	"io/ioutil"
+	"strconv"
+	"strings"
+	"time"
+)
+
+func readTxBytes(interfaceName string) (uint64, error) {
+	v, err := ioutil.ReadFile("/sys/class/net/" + interfaceName + "/statistics/tx_bytes")
```
danderson, line 15: Doing it this way means that if I'm scp-ing something off a derp server, all clients on the server will get throttled. I don't think that's what we want? I'd strongly prefer that we do the accounting of sent bytes within derper, rather than rely on undifferentiated OS stats.

Plus this means derper can't run on !linux any more, which previously was possible.

derp/limiter.go
```diff
@@ -0,0 +1,171 @@
...
+		return 0, err
+	}
+	tx, err := strconv.Atoi(strings.TrimSpace(string(v)))
+	if err != nil {
+		return 0, err
+	}
+	return uint64(tx), nil
+}
+
+type egressLimiter struct {
```
danderson, line 26: docstring pls, not clear to me what this is meant to do from just the struct definition.

derp/limiter.go
```diff
@@ -0,0 +1,171 @@
...
+		limitBytesSec: limitBytesSec,
+		minBytesSec:   minBytesSec,
+		lastTxBytes:   initial,
+		controlLoop:   newLimiterLoop(limitBytesSec, time.Now()),
+	}, err
+}
+
+// Limit returns the current rate limit value based on interface utilization.
+func (e *egressLimiter) Limit() (uint64, error) {
+	rx, err := readTxBytes(e.interfaceName)
```
danderson, line 52: rx in the var name, tx in the function?

derp/limiter.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	if uint64(limit) > e.limitBytesSec {
+		limit = float64(e.limitBytesSec)
+	}
+	return uint64(limit), nil
+}
+
+// PID loop values for the dynamic ratelimit.
+// The wikipedia page on PID is recommended reading if you are not familiar
+// with PID loops or open-loop control theory.
```
danderson, line 72: closed-loop, not open-loop, if we have a PID controller

derp/limiter.go
```diff
@@ -0,0 +1,171 @@
...
+	if dt > 0 {
+		d = -(float64(egressBytesPerSec-l.lastEgress) / dt) * limiterD
+	}
+	// Proportional term.
+	p := limiterP * err
+
+	l.lastEgress = egressBytesPerSec
+	l.lastUpdate = now
+	output := p + l.integral + d
+	// fmt.Printf("in=%d, out=%0.3f:   p=%0.2f d=%0.2f i=%0.2f\n", egressBytesPerSec, output, p, d, l.integral)
```
danderson, line 169: Remove debug stuff

derp/limiter_test.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"testing"
+	"time"
+)
+
+func mb(mb uint64) uint64 {
+	return mb * 1024 * 1024
+}
+
+func TestLimiterLoopGradual(t *testing.T) {
```
danderson, line 16: Make this a table-based test, with successive amounts of traffic and the expected limiter output, and verifies that at each step the loop settles on the expected output.

For bonus points, also have tests that shuffle the table and verify that everything still works, since the loop should settle on the same new output regardless of past values.

danderson (commented):
Still a bit worried about introducing the requirement that we tune a bunch of magic numbers, as we talked about last week. With more tests to verify the loop's behavior under various traffic conditions (step changes, oscillating inputs, etc.), I'd feel more confident.

PR #5646, opened 2022-09-15
cmd/tailscale, client, ipn, tailcfg: add network lock modify command

sailorfrag:


cmd/tailscale/cli/network-lock.go
```diff
@@ -99,3 +104,77 @@ func runNetworkLockStatus(ctx context.Context, args []string) error {
...
+func runNetworkLockModify(ctx context.Context, addArgs, removeArgs []string) error {
+	st, err := localClient.NetworkLockStatus(ctx)
+	if err != nil {
+		return fixTailscaledConnectError(err)
+	}
+	if st.Enabled {
+		return errors.New("network-lock is already enabled")
+	}
+
+	// Parse the set of initially-trusted keys.
```
twitchyliquid64, line 135: Maybe we should move the parseArgs closure to its own function now its duplicated?

ipn/ipnlocal/network-lock.go
```diff
@@ -137,6 +138,122 @@ func (b *LocalBackend) NetworkLockInit(keys []tka.Key) error {
...
+	if err != nil {
+		return err
+	}
+
+	lastHead := aums[len(aums)-1].Hash()
+	if !slices.Equal(head[:], lastHead[:]) {
+		return errors.New("central tka head differs from submitted AUM, try again")
+	}
+
+	if err := b.tka.authority.Inform(b.tka.storage, aums); err != nil {
```
twitchyliquid64, line 201: Do you think we should `Inform()` locally or let the update come through from control in the usual / mapResponsy fashion? I was originally thinking the latter.

ipn/ipnlocal/network-lock.go
```diff
@@ -137,6 +138,122 @@ func (b *LocalBackend) NetworkLockInit(keys []tka.Key) error {
...
+}
+
+func (b *LocalBackend) sendAUMs(aums []tka.AUM) (head tka.AUMHash, err error) {
+	mAUMs := make([]tkatype.MarshaledAUM, len(aums))
+	for i := range aums {
+		mAUMs[i] = aums[i].Serialize()
+	}
+
+	var req bytes.Buffer
+	if err := json.NewEncoder(&req).Encode(tailcfg.TKASyncSendRequest{
```
twitchyliquid64, line 215: We should add a new bool, `Interactive`, so control knows it can do the additional validation for head updates (as oppoosed to the other case where control is learning about AUMs it lost during data loss, and should only reject if its invalid/unsigned).

ipn/localapi/localapi.go
```diff
@@ -852,6 +854,40 @@ func (h *Handler) serveTkaInit(w http.ResponseWriter, r *http.Request) {
...
+		RemoveKeys []tka.Key
+	}
+	var req modifyRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, "invalid JSON body", 400)
+		return
+	}
+
+	if err := h.b.NetworkLockModify(req.AddKeys, req.RemoveKeys); err != nil {
+		http.Error(w, "initialization failed: "+err.Error(), http.StatusInternalServerError)
```
twitchyliquid64, line 878: update error?

sailorfrag:
I believe I've cleaned this up from the comments, please amend as desired or merge; I'm off on vacation for a while.

twitchyliquid64 (approved):

PR #5648, opened 2022-09-15
control: add "ControlDialPlan" field to MapResponse; use when connecting

andrew-d:
(split into N commits to make it easier to see what's a refactor and what's new; will squash before merging)

This field allows the control server to provide explicit information
about how to connect to it; useful if the client's link status can
change after the initial connection, or if the DNS settings pushed by
the control server break future connections.

Change-Id: I720afe6289ec27d40a41b3dcb310ec45bd7e5f3e


andrew-d:
When using the test dial plan in the commit, logs look like this when run with verbosity >= 2:
```
controlhttp: trying to dial "controlplane.tailscale.com" @ 13.0.0.1
controlhttp: waiting 0.50 seconds before dialing "controlplane.tailscale.com" @ 2a05:d014:386:201:7200:6340:b22f:7df6
controlhttp: waiting 0.50 seconds before dialing "controlplane.tailscale.com" @ 18.157.173.201
controlhttp: trying to dial "controlplane.tailscale.com" @ 18.157.173.201
controlhttp: trying to dial "controlplane.tailscale.com" @ 2a05:d014:386:201:7200:6340:b22f:7df6
controlhttp: succeeded dialing "controlplane.tailscale.com" @ 18.157.173.201 from dial plan
control: RegisterReq: got response; nodeKeyExpired=false, machineAuthorized=false; authURL=true
[...]
```

tailcfg/tailcfg.go
```diff
@@ -1375,6 +1375,39 @@ type MapResponse struct {
...
 	Debug *Debug `json:",omitempty"`
+
+	// ControlDialPlan tells the client how to connect to the control
+	// server. An initial nil is equivalent to new(ControlDialPlan).
+	// A subsequent streamed nil means no change.
+	ControlDialPlan *ControlDialPlan `json:",omitempty"`
+}
+
+// ControlDialPlan is instructions from the control server to the client on how
+// to connect to the control server; useful for maintaining connection if the
```
bradfitz, line 1386: a complete sentence follows a semicolon. "useful for" lacks a subject.

tailcfg/tailcfg.go
```diff
@@ -1375,6 +1375,39 @@ type MapResponse struct {
...
+// connecting to the control server.
+type ControlIPCandidate struct {
+	// IP is the address to attempt connecting to.
+	IP netip.Addr
+
+	// DialStartSec is the number of seconds after the beginning of the
+	// connection process to wait before trying this candidate.
+	DialStartDelaySec float64 `json:",omitempty"`
+
+	// DialTimeoutSec is the timeout for a connection to this candidate.
```
bradfitz, line 1404: timeout from the beginning or from DialStartDelaySec?

control/controlhttp/constants.go
```diff
@@ -18,3 +27,40 @@ const (
 	// to do the protocol switch is located.
 	serverUpgradePath = "/ts2021"
 )
+
+// DialOpts contains options when calling Dial.
+type DialOpts struct {
+	// The following fields are required
```
bradfitz, line 33: typically required fields are positional parameters and Opts are for optional ones.

I wonder if this shouldn't just be like https://pkg.go.dev/net#Dialer and be a struct of things, rather than a `Dial(context.Context)` method.

Then there are no opts and it's less weird to have required "options".

andrew-d: Pushed another commit at the end that does this; thoughts? It does remove a bunch of mostly-useless copying around, which is nice.

control/controlhttp/http_test.go
```diff
@@ -176,17 +176,17 @@ func testControlHTTP(t *testing.T, param httpTestParam) {
...
 	}
 
-	a := dialParams{
-		host:              "localhost",
-		httpPort:          strconv.Itoa(httpLn.Addr().(*net.TCPAddr).Port),
-		httpsPort:         strconv.Itoa(httpsLn.Addr().(*net.TCPAddr).Port),
-		machineKey:        client,
-		controlKey:        server.Public(),
-		version:           testProtocolVersion,
+	a := Dialer{
```
bradfitz, line 179: &Dialer

(to stay in same type as receiver methods and prevent accidental value copies)

control/controlhttp/http_test.go
```diff
@@ -608,20 +608,20 @@ func TestDialPlan(t *testing.T) {
...
-				httpPort:          httpPort,
-				httpsPort:         httpsPort,
-				machineKey:        client,
-				controlKey:        server.Public(),
-				version:           testProtocolVersion,
-				insecureTLS:       true,
-				dialer:            dialer.Dial,
-				testFallbackDelay: 50 * time.Millisecond,
-				dialPlan:          &plan,
+			a := Dialer{
```
bradfitz, line 611: same

control/controlhttp/constants.go
```diff
@@ -28,33 +31,43 @@ const (
 	serverUpgradePath = "/ts2021"
 )
 
-// DialOpts contains options when calling Dial.
-type DialOpts struct {
-	// The following fields are required
-
+// Dialer contains configuration on how to dial the Tailscale control server.
+type Dialer struct {
```
bradfitz, line 35: can the controlhttp move to a Dialer type be in its own change first without any dialplan stuff?

andrew-d: Sure, lemme break this out into a new PR

andrew-d: #5661

control/controlhttp/http_test.go
andrew-d: Addressed in #5661

andrew-d:
@bradfitz - Rebased on `main` to pick up the changes from #5661

control/controlclient/direct.go
```diff
@@ -76,6 +77,8 @@ type Direct struct {
 	popBrowser             func(url string) // or nil
 	c2nHandler             http.Handler     // or nil
 
+	dialPlan *atomic.Pointer[tailcfg.ControlDialPlan] // can be nil
```
bradfitz, line 80: this is a pointer to a pointer. (is that intentional? feels very odd) if so, which pointer can be nil? both?

andrew-d: Yes, this is a pointer-to-pointer, and no, I'm not amazingly happy with this; suggestions appreciated. This is because `atomic.Pointer` isn't copyable, and we need to thread the value through several layers of abstraction.

Either can be nil; if the outermost pointer is nil, it's because there's no dial plan storage provided (e.g. the zero value for `Options`). If the innermost (atomic) pointer is nil, it's because we haven't yet received a dial plan from the control server, for example on the first connection.

bradfitz: how about `dialPlan func() *tailcfg.ControlDialPlan`?

andrew-d: And then thread that through the entire stack with the `atomic.Pointer.Load` function as the top-level? Yeah, that should work; lemme try it.

andrew-d: That cleaned things up a bit, but we still need to store a `ControlDialPlan` between instances of `Direct`, so this still needs to be a pointer-to-a-pointer. I could make a wrapper type for `atomic.Pointer` that just checks if the receiver is nil and returns nil, though? `syncs.NilableAtomicPointer[tailcfg.ControlDialPlan]` that does nothing if the receiver is nil on `Store` and returns `nil` on `Load`?

andrew-d: Switched to an interface after chatting today; no more pointers-to-pointers!

andrew-d:
@bradfitz - this is ready for review; removed all of my "TODO"s that I had for testing.

control/controlclient/direct.go
```diff
@@ -915,6 +932,14 @@ func (c *Direct) sendMapRequest(ctx context.Context, maxPolls int, readOnly bool
 		} else {
 			vlogf("netmap: got new map")
 		}
+		if resp.ControlDialPlan != nil {
+			if c.dialPlan != nil {
+				c.logf("netmap: got new dial plan from control")
+				c.dialPlan.Store(resp.ControlDialPlan)
+			} else {
+				c.logf("netmap: new dial plan; nowhere to store it")
```
bradfitz, line 940: if this is expected, it maybe should be [v1]. If it's unexpected, mark it with `[unexpected]` which is our convention used elsewhere for things that shouldn't happen but aren't fatal-worthy.

control/controlclient/noise.go
```diff
@@ -53,6 +54,10 @@ type noiseClient struct {
 	httpPort     string // the default port to call
 	httpsPort    string // the fallback Noise-over-https port
 
+	// Optional DialPlan received from the control server previously; can
```
bradfitz, line 57: If it's a before-line comment, use a complete sentence.

// dialPlan optionally ...

control/controlclient/noise.go
```diff
@@ -61,7 +66,7 @@ type noiseClient struct {
 
 // newNoiseClient returns a new noiseClient for the provided server and machine key.
 // serverURL is of the form https://<host>:<port> (no trailing slash).
```
bradfitz, line 68: //
// dialPlan may be nil.

(if true)

control/controlclient/noise.go
```diff
@@ -155,16 +161,52 @@ func (nc *noiseClient) dial(_, _ string, _ *tls.Config) (net.Conn, error) {
...
+	}
+
+	// If we have a dial plan, then set our timeout as slightly longer than
+	// the maximum amount of time contained therein; we assume that
+	// explicit instructions on timeouts are more useful than a single
+	// hard-coded timeout.
+	//
+	// The default value of 5 is chosen so that, when there's no dial plan,
+	// we retain the previous behaviour of 10 seconds end-to-end timeout.
+	var timeoutSec float64 = 5
```
bradfitz, line 182: or `timeoutSec := 5.0`

or `timeout := 5 * time.Second` and do the conversions below.

we generally try to use `time.Duration` whenever possible except on the wire where it kinda sucks

control/controlclient/noise.go
```diff
@@ -155,16 +161,52 @@ func (nc *noiseClient) dial(_, _ string, _ *tls.Config) (net.Conn, error) {
...
+	// Be extremely defensive and ensure that the timeout is in the range
+	// [5, 60] seconds (e.g. if we accidentally get a negative number).
+	if timeoutSec > 60 {
+		timeoutSec = 60
+	} else if timeoutSec < 5 {
+		timeoutSec = 5
+	}
+
+	timeout := time.Duration(timeoutSec * float64(time.Second))
+	log.Printf("controlclient: Noise dial timeout: %v", timeout)
```
bradfitz, line 206: debug? remove? or [v1] maybe? or only log it on failure?

control/controlhttp/client.go
```diff
@@ -83,6 +89,159 @@ func (a *Dialer) httpsFallbackDelay() time.Duration {
 }
 
 func (a *Dialer) dial(ctx context.Context) (*controlbase.Conn, error) {
+	// If we don't have a dial plan, just fall back to dialing the single
+	// host we know about.
+	useDialPlan := envknob.BoolDefaultTrue("TS_USE_CONTROL_DIAL_PLAN")
```
bradfitz, line 94: if you use an envknob like this then it won't be recorded earlier as in use and logged in the early logs which can help with debugging.

I might keep this here (for the DefaultTrue bit) but add a global like:

```go
var _ = envknob.RegisterBool("TS_USE_CONTROL_DIAL_PLAN") // to record at init time whether it's in use
```

control/controlhttp/client.go
```diff
@@ -28,18 +28,24 @@ import (
...
 	"time"
 
 	"tailscale.com/control/controlbase"
+	"tailscale.com/envknob"
 	"tailscale.com/net/dnscache"
 	"tailscale.com/net/dnsfallback"
 	"tailscale.com/net/netutil"
 	"tailscale.com/net/tlsdial"
 	"tailscale.com/net/tshttpproxy"
+	"tailscale.com/tailcfg"
```
bradfitz, line 48: I was going to say I thought Dave wanted this package to be kinda Tailscale-free and just a Noise thing, but I guess this was already a dep and we lost that cleanliness property some time ago:

```
bradfitz@book1pro controlhttp % ~/go/bin/deppath tailscale.com/control/controlhttp /tailcfg
["tailscale.com/control/controlhttp" "tailscale.com/net/dnsfallback" "tailscale.com/net/netns" "tailscale.com/net/interfaces" "tailscale.com/hostinfo" "tailscale.com/tailcfg"]
["tailscale.com/control/controlhttp" "tailscale.com/net/dnsfallback" "tailscale.com/tailcfg"]
["tailscale.com/control/controlhttp" "tailscale.com/net/netutil" "tailscale.com/net/interfaces" "tailscale.com/hostinfo" "tailscale.com/tailcfg"]
```

control/controlhttp/client.go
```diff
@@ -83,6 +89,159 @@ func (a *Dialer) httpsFallbackDelay() time.Duration {
...
+		for _, result := range results {
+			if result.conn != nil {
+				result.conn.Close()
+			}
+		}
+
+		// We don't drain asynchronously after this point, so notify our
+		// channel when we return.
+		if a.drainFinished != nil {
+			defer close(a.drainFinished)
```
bradfitz, line 216: why defer here?

control/controlhttp/client.go
```diff
@@ -83,6 +89,159 @@ func (a *Dialer) httpsFallbackDelay() time.Duration {
...
+		results[i].conn = nil // so we don't close it in the defer
+		return conn, nil
+	}
+
+	// If we get here, then we didn't get anywhere with our dial plan; fall back to just using DNS.
+	a.logf("controlhttp: failed dialing using DialPlan; falling back to DNS")
+	return a.dialHost(ctx, netip.Addr{})
+}
+
+func (a *Dialer) dialHost(ctx context.Context, addr netip.Addr) (*controlbase.Conn, error) {
```
bradfitz, line 244: document what addr means here? zero value seems to be special.

bradfitz (commented):
I haven't finished reading Dialer.dial (the meaty part)

control/controlclient/direct.go
andrew-d: Added `[unexpected]`

control/controlclient/noise.go
andrew-d: Used `:=` and kept it as a float, since otherwise we just move the casting below (since `time.Duration` isn't an integer type and we need to support non-integer seconds).

control/controlclient/noise.go
andrew-d: Debug; removed

control/controlhttp/client.go
andrew-d: Leftover from a refactor; good catch

control/controlclient/direct.go
```diff
@@ -133,6 +135,20 @@ type Options struct {
...
+
+	// DialPlan contains and stores a previous dial plan that we received
+	// from the control server; if nil, we fall back to using DNS.
+	//
+	// If we receive a new DialPlan from the server, this value will be
+	// updated.
+	DialPlan ControlDialPlanner
+}
+
+// ControlDialPlanner is implemented by an atomic.Pointer to
```
bradfitz, line 147:
```
// ControlDialPlanner is the interface optionally supplied when creating a control client
// to control exactly how TCP connections to the control plane are dialed.
//
// It is usually implemented by an atomic.Pointer.
type .. {
    // Load returns the current plan for how to connect to control.
    //
    // It can (or can not?) be nil. If so that means ...

    // Store updates the dial plan from new directions from the control server.
    // The dial plan can span multiple connections to the control plan.
    // That is, an dial plan received when connected over wi-fi is still valid for
    // a subsequent connection over LTE after a network switch.
```

andrew-d: Used this with minor tweaks :+1:

control/controlhttp/client.go
```diff
@@ -82,7 +88,165 @@ func (a *Dialer) httpsFallbackDelay() time.Duration {
...
+		// we're done.
+		results = append(results, res)
+	}
+
+	// After we finish this function, close any remaining open connections;
+	// we remove a successful connection from the slice below so we don't
+	// close it.
+	defer func() {
+		for _, result := range results {
+			if result.conn != nil {
```
bradfitz, line 210: This seemed wrong to me at first until I saw below where you nilled it out. Add little comment?

control/controlhttp/client.go
```diff
@@ -82,7 +88,165 @@ func (a *Dialer) httpsFallbackDelay() time.Duration {
...
+			continue
+		}
+
+		a.logf("[v1] controlhttp: succeeded dialing %q @ %v from dial plan", a.Hostname, result.addr)
+		conn = result.conn
+		results[i].conn = nil // so we don't close it in the defer
+		return conn, nil
+	}
+
+	// If we get here, then we didn't get anywhere with our dial plan; fall back to just using DNS.
```
bradfitz, line 241: do we want to at least log some/one of the errors for debugging before we fall back to DNS?

bradfitz (approved):

PR #5651, opened 2022-09-15
cmd/derper, control/controlhttp: disable WebSocket compression

mihaip:
The data that we send over WebSockets is encrypted and thus not compressible. Additionally, Safari has a broken implementation of compression (see nhooyr/websocket#218) that makes enabling it actively harmful.

Fixes tailscale/corp#6943


bradfitz (approved):
Wow

Issue #5653, opened 2022-09-15
sidecar.yaml uses incorrect value for the kubernetes secret variable


tleerai:
### What is the issue?

The [README.md](https://github.com/tailscale/tailscale/blob/main/docs/k8s/README.md?plain=1#L19) file for the k8s documentation advises users to create a secret with string data which does not match the value being used in the sidecar.yaml:

```
   stringData:
     TS_AUTH_KEY: tskey-...
```

However the [sidecar container](https://github.com/tailscale/tailscale/blob/main/docs/k8s/sidecar.yaml#L26) is actually instructed to use a different name:

```
    - name: TS_AUTH_KEY
      valueFrom:
        secretKeyRef:
          name: tailscale-auth
          key: AUTH_KEY
          optional: true
```

PR #5654, opened 2022-09-15
Updated value for the kubernetes secret used by the sidecar container

tleerai:
This changes sidecar.yaml so it uses the same value for the `key` stanza as the README.md and run.sh

Fixes: https://github.com/tailscale/tailscale/issues/5653

Signed-off-by: Tyler Lee <tyler.lee@radius.ai>

### DIFF
--- a/docs/k8s/README.md
+++ b/docs/k8s/README.md
@@ -16,7 +16,7 @@ There are quite a few ways of running Tailscale inside a Kubernetes Cluster, som
    metadata:
      name: tailscale-auth
    stringData:
-     TS_AUTH_KEY: tskey-...
+     AUTH_KEY: tskey-...
    ```
 
 1. Tailscale (v1.16+) supports storing state inside a Kubernetes Secret.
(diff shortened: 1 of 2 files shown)


docs/k8s/README.md
```diff
@@ -16,7 +16,7 @@ There are quite a few ways of running Tailscale inside a Kubernetes Cluster, som
    metadata:
      name: tailscale-auth
    stringData:
-     TS_AUTH_KEY: tskey-...
+     AUTH_KEY: tskey-...
```
DentonGentry, line 19: run.sh uses TS_AUTH_KEY: https://github.com/tailscale/tailscale/blob/027111fb5a3785c4cc3a68d8ad347eece380e9e7/docs/k8s/run.sh#L9

sidecar.yaml needs to be updated to use TS_AUTH_KEY instead.

DentonGentry (changes requested):

docs/k8s/README.md
tleerai: Fixed: e7476e2dd66151590695bae9018d5e3c51e089e5

DentonGentry (approved):

PR #5657, opened 2022-09-16
ipn/{ipnlocal,localapi}: use strs.CutPrefix, add more domain validation

bradfitz:
The GitHub CodeQL scanner flagged the localapi's cert domain usage as a problem because user input in the URL made it to disk stat checks.

The domain is validated against the ipnstate.Status later, and only authenticated root/configured users can hit this, but add some paranoia anyway.


andrew-d (approved):
Thanks!

PR #5661, opened 2022-09-16
control/controlhttp: move Dial options into options struct

andrew-d:
This turns 'dialParams' into something more like net.Dialer, where configuration fields are public on the struct.

Split out of #5648

Change-Id: I0c56fd151dc5489c3c94fb40d18fd639e06473bc

### DIFF
--- a/cmd/tailscale/cli/debug.go
+++ b/cmd/tailscale/cli/debug.go
@@ -489,7 +489,15 @@ func runTS2021(ctx context.Context, args []string) error {
 		return c, err
 	}
 
-	conn, err := controlhttp.Dial(ctx, ts2021Args.host, "80", "443", machinePrivate, keys.PublicKey, uint16(ts2021Args.version), dialFunc)
+	conn, err := (&controlhttp.Dialer{
+		Hostname:        ts2021Args.host,
+		HTTPPort:        "80",
+		HTTPSPort:       "443",
+		MachineKey:      machinePrivate,
+		ControlKey:      keys.PublicKey,
+		ProtocolVersion: uint16(ts2021Args.version),
+		Dialer:          dialFunc,
+	}).Dial(ctx)
 	log.Printf("controlhttp.Dial = %p, %v", conn, err)
 	if err != nil {
 		return err
--- a/control/controlhttp/client.go
+++ b/control/controlhttp/client.go
@@ -40,57 +40,49 @@ import (
 	"tailscale.com/net/netutil"
 	"tailscale.com/net/tlsdial"
 	"tailscale.com/net/tshttpproxy"
-	"tailscale.com/types/key"
 )
 
-// Dial connects to the HTTP server at host:httpPort, requests to switch to the
-// Tailscale control protocol, and returns an established control
+var stdDialer net.Dialer
+
+// Dial connects to the HTTP server at this Dialer's Host:HTTPPort, requests to
+// switch to the Tailscale control protocol, and returns an established control
 // protocol connection.
 //
-// If Dial fails to connect using addr, it also tries to tunnel over
-// TLS to host:httpsPort as a compatibility fallback.
+// If Dial fails to connect using HTTP, it also tries to tunnel over TLS to the
+// Dialer's Host:HTTPSPort as a compatibility fallback.
 //
 // The provided ctx is only used for the initial connection, until
 // Dial returns. It does not affect the connection once established.
-func Dial(ctx context.Context, host string, httpPort string, httpsPort string, machineKey key.MachinePrivate, controlKey key.MachinePublic, protocolVersion uint16, dialer dnscache.DialContextFunc) (*controlbase.Conn, error) {
-	a := &dialParams{
-		host:       host,
-		httpPort:   httpPort,
-		httpsPort:  httpsPort,
-		machineKey: machineKey,
-		controlKey: controlKey,
-		version:    protocolVersion,
-		proxyFunc:  tshttpproxy.ProxyFromEnvironment,
-		dialer:     dialer,
+func (a *Dialer) Dial(ctx context.Context) (*controlbase.Conn, error) {
+	if a.Hostname == "" {
+		return nil, errors.New("required Dialer.Hostname empty")
 	}
 	return a.dial(ctx)
 }
 
-type dialParams struct {
-	host       string
-	httpPort   string
-	httpsPort  string
-	machineKey key.MachinePrivate
-	controlKey key.MachinePublic
-	version    uint16
-	proxyFunc  func(*http.Request) (*url.URL, error) // or nil
-	dialer     dnscache.DialContextFunc
-
-	// For tests only
-	insecureTLS       bool
-	testFallbackDelay time.Duration
+func (a *Dialer) logf(format string, args ...any) {
+	if a.Logf != nil {
+		a.Logf(format, args...)
+	}
 }
 
-// httpsFallbackDelay is how long we'll wait for a.httpPort to work before
-// starting to try a.httpsPort.
-func (a *dialParams) httpsFallbackDelay() time.Duration {
+func (a *Dialer) getProxyFunc() func(*http.Request) (*url.URL, error) {
+	if a.proxyFunc != nil {
+		return a.proxyFunc
+	}
+	return tshttpproxy.ProxyFromEnvironment
+}
+
+// httpsFallbackDelay is how long we'll wait for a.HTTPPort to work before
+// starting to try a.HTTPSPort.
+func (a *Dialer) httpsFallbackDelay() time.Duration {
 	if v := a.testFallbackDelay; v != 0 {
 		return v
 	}
 	return 500 * time.Millisecond
 }
 
-func (a *dialParams) dial(ctx context.Context) (*controlbase.Conn, error) {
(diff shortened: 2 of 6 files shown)


control/controlhttp/client.go
```diff
@@ -52,45 +51,29 @@ import (
 //
```
bradfitz, line 51: docs are stale now

control/controlhttp/client.go
```diff
@@ -52,45 +51,29 @@ import (
...
-		httpPort:   httpPort,
-		httpsPort:  httpsPort,
-		machineKey: machineKey,
-		controlKey: controlKey,
-		version:    protocolVersion,
-		proxyFunc:  tshttpproxy.ProxyFromEnvironment,
-		dialer:     dialer,
+func (a *Dialer) Dial(ctx context.Context) (*controlbase.Conn, error) {
+	if a.proxyFunc == nil {
+		a.proxyFunc = tshttpproxy.ProxyFromEnvironment
```
bradfitz, line 56: don't mutate a.

instead, have an accessor method on Dialer that returns a non-nil value based on a.

otherwise it's racey

control/controlhttp/constants.go
```diff
@@ -18,3 +28,51 @@ const (
 	// to do the protocol switch is located.
 	serverUpgradePath = "/ts2021"
 )
+
+// Dialer contains configuration on how to dial the Tailscale control server.
+type Dialer struct {
+	// Host is the host to connect to.
```
bradfitz, line 34: is the host name? (I assume it must be a DNS name and not a raw IP?)

and no port number, since that's below.

control/controlhttp/constants.go
```diff
@@ -18,3 +28,51 @@ const (
...
+	//
+	// This field is required.
+	Host string
+
+	// HTTPPort is the port number to use when making a HTTP connection.
+	//
+	// This field is required.
+	HTTPPort string
```
bradfitz, line 42: why required? Why not default to 80?

control/controlhttp/constants.go
```diff
@@ -18,3 +28,51 @@ const (
...
+
+	// HTTPSPort is the port number to use when making a HTTPS connection.
+	//
+	// This field is required.
+	HTTPSPort string
```
bradfitz, line 47: same? 443 seems reasonable.

control/controlhttp/constants.go
```diff
@@ -18,3 +28,51 @@ const (
...
+
+	// ProtocolVersion is the expected protocol version to negotiate.
+	//
+	// This field is required.
+	ProtocolVersion uint16
+
+	// Dialer is the dialer used to make outbound connections.
+	//
+	// This field is required.
+	Dialer dnscache.DialContextFunc
```
bradfitz, line 67: why? net.Dial seems like a fine default?

andrew-d: For this (and the other cases) I was replicating the prior behaviour; happy to change to a sensible default.

bradfitz: yeah, let's. I think it's just those 3 fields that have pretty obvious defaults.

control/controlhttp/constants.go
andrew-d: Renamed to `Hostname` to be more clear; added comment stating that this should not contain a port.

control/controlhttp/constants.go
andrew-d: Done; added defaults and updated comments.

control/controlhttp/client.go
```diff
@@ -100,14 +87,26 @@ func (a *dialParams) dial(ctx context.Context) (*controlbase.Conn, error) {
...
+	if a.HTTPSPort != "" {
+		port443 = a.HTTPSPort
+	} else {
+		port443 = "443"
+	}
+
 	u80 := &url.URL{
 		Scheme: "http",
-		Host:   net.JoinHostPort(a.host, a.httpPort),
+		Host:   net.JoinHostPort(a.Hostname, port80),
```
bradfitz, line 104: or some helper and like `net.JoinHostPort(a.Hostname, strDef(a.HTTPPort, "80"))` here and line 109

control/controlhttp/client.go
```diff
@@ -189,26 +188,34 @@ func (a *dialParams) dialURL(ctx context.Context, u *url.URL) (*controlbase.Conn
...
-	tr.Proxy = a.proxyFunc
+	tr.Proxy = a.getProxyFunc()
 	tshttpproxy.SetTransportGetProxyConnectHeader(tr)
-	tr.DialContext = dnscache.Dialer(a.dialer, dns)
+	tr.DialContext = dnscache.Dialer(dialer, dns)
 	// Disable HTTP2, since h2 can't do protocol switching.
 	tr.TLSClientConfig.NextProtos = []string{}
 	tr.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{}
-	tr.TLSClientConfig = tlsdial.Config(a.host, tr.TLSClientConfig)
+	tr.TLSClientConfig = tlsdial.Config(a.Hostname, tr.TLSClientConfig)
```
bradfitz, line 213: do we fail early anywhere with a nice error if a.Hostname is empty?

control/controlhttp/client.go
```diff
@@ -189,26 +188,34 @@ func (a *dialParams) dialURL(ctx context.Context, u *url.URL) (*controlbase.Conn
...
 		LookupIPFallback: dnsfallback.Lookup,
 		UseLastGood:      true,
 	}
+
+	var dialer dnscache.DialContextFunc
+	if a.Dialer != nil {
+		dialer = a.Dialer
+	} else {
+		dialer = new(net.Dialer).DialContext
+	}
```
bradfitz, line 203: you could write the strDefault as:

```go
func valOr[T comparable](v1, v2 T) T { ... }
```

And then use it here too.

But drop the `new(net.Dialer)` and make a package global `var stdDialer net.Dialer` and then just use `stdDialer.DialContext` here in the `valOr` call so it doesn't allocate?

control/controlhttp/constants.go
```diff
@@ -18,3 +28,51 @@ const (
...
 )
+
+// Dialer contains configuration on how to dial the Tailscale control server.
+type Dialer struct {
+	// Hostname is the hostname to connect to, with no port number.
+	//
+	// This field is required.
+	Hostname string
+
+	// HTTPPort is the port number to use when making a HTTP connection.
```
bradfitz, line 39: move all optional stuff after required fields

control/controlhttp/client.go
andrew-d: `dnscache.DialContextFunc does not implement comparable` unfortunately :unamused:

control/controlhttp/client.go
andrew-d: We do now!

control/controlhttp/client.go
bradfitz: well, could use T any and reflect.ValueOf(v1).IsZero(). I wish there was a magic constraint that's like `isZeroCheckable`

cmd/tailscale/cli/debug.go
```diff
@@ -489,7 +489,15 @@ func runTS2021(ctx context.Context, args []string) error {
 		return c, err
 	}
 
-	conn, err := controlhttp.Dial(ctx, ts2021Args.host, "80", "443", machinePrivate, keys.PublicKey, uint16(ts2021Args.version), dialFunc)
+	conn, err := (&controlhttp.Dialer{
+		Hostname:        ts2021Args.host,
+		HTTPPort:        "80",
+		HTTPSPort:       "443",
```
bradfitz, line 495: can remove these now, eh? either way

control/controlhttp/client.go
```diff
@@ -40,57 +40,49 @@ import (
...
-		httpPort:   httpPort,
-		httpsPort:  httpsPort,
-		machineKey: machineKey,
-		controlKey: controlKey,
-		version:    protocolVersion,
-		proxyFunc:  tshttpproxy.ProxyFromEnvironment,
-		dialer:     dialer,
+func (a *Dialer) Dial(ctx context.Context) (*controlbase.Conn, error) {
+	if a.Hostname == "" {
+		return nil, errors.New("no Hostname provided")
```
bradfitz, line 58: maybe "required Dialer.Hostname empty"

control/controlhttp/client_js.go
```diff
@@ -7,27 +7,31 @@ package controlhttp
...
-	"tailscale.com/types/key"
 )
 
 // Variant of Dial that tunnels the request over WebSockets, since we cannot do
 // bi-directional communication over an HTTP connection when in JS.
-func Dial(ctx context.Context, host string, httpPort string, httpsPort string, machineKey key.MachinePrivate, controlKey key.MachinePublic, protocolVersion uint16, dialer dnscache.DialContextFunc) (*controlbase.Conn, error) {
-	init, cont, err := controlbase.ClientDeferred(machineKey, controlKey, protocolVersion)
+func (d *Dialer) Dial(ctx context.Context) (*controlbase.Conn, error) {
+	if d.Hostname == "" {
+		return nil, errors.New("no Hostname provided")
```
bradfitz, line 22: maybe "required Dialer.Hostname empty"

bradfitz (approved):

PR #5667, opened 2022-09-16
cmd/tailscaled: use explicit equal sign in --port=$PORT in tailscaled.service

bradfitz:
Personal preference (so it's obvious it's not a bool flag), but it also matches the --state= before it.

Bonus: stop allowing PORT to sneak in extra flags to be passed as their own arguments, as $FOO and ${FOO} expand differently. (${FOO} is required to concat to strings)


DentonGentry (approved):

danderson (approved):
Style-wise, sure. Any concern about breaking existing users? I'm tempted to say no since we have an explicit means of providing extra flags on the very next line of the env file, but https://xkcd.com/1172/

PR #5680, opened 2022-09-17
cmd/tailscale/cli: [up] move lose-ssh check after other validations

maisem:
The check was happening too early and in the case of error would wait 5 s and then error out. This makes it so that it does validations before the SSH check.

### DIFF
--- a/cmd/tailscale/cli/cli_test.go
+++ b/cmd/tailscale/cli/cli_test.go
@@ -789,6 +789,10 @@ func TestUpdatePrefs(t *testing.T) {
 		curPrefs *ipn.Prefs
 		env      upCheckEnv // empty goos means "linux"
 
+		// sshOverTailscale specifies if the cmd being run over SSH over Tailscale.
+		// It is used to test the --accept-risks flag.
+		sshOverTailscale bool
+
 		// checkUpdatePrefsMutations, if non-nil, is run with the new prefs after
 		// updatePrefs might've mutated them (from applyImplicitPrefs).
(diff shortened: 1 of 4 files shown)


bradfitz (approved):

cmd/tailscale/cli/cli_test.go
```diff
@@ -789,6 +789,8 @@ func TestUpdatePrefs(t *testing.T) {
 		curPrefs *ipn.Prefs
 		env      upCheckEnv // empty goos means "linux"
 
+		sshOverTailscale bool
```
bradfitz, line 792: document what this means?

cmd/tailscale/cli/risks.go
```diff
@@ -28,12 +27,12 @@ func registerRiskType(riskType string) string {
...
 // in presentRiskToUser.
-func registerAcceptRiskFlag(f *flag.FlagSet) {
-	f.StringVar(&acceptedRisks, "accept-risk", "", "accept risk and skip confirmation for risk types: "+strings.Join(riskTypes, ","))
+func registerAcceptRiskFlag(f *flag.FlagSet, acceptedRisks *string) {
+	f.StringVar(acceptedRisks, "accept-risk", "", "accept risk and skip confirmation for risk types: "+strings.Join(riskTypes, ","))
 }
 
-// riskAccepted reports whether riskType is in acceptedRisks.
-func riskAccepted(riskType string) bool {
+// isRiskAccepted reports whether riskType is in acceptedRisks.
```
bradfitz, line 34: is in the comma-separated list of risks in acceptedRisks

bradfitz (approved):

Issue #4482, opened 2022-04-21
builder: run Android builder in tailscale repo


bradfitz:
We should have a CI builder that checks whether a change to tailscale/tailscale would break Android.

(like we do in corp for `if-oss-were-updated`)

/cc @raggi @DentonGentry

DentonGentry:
When this CI fails, what will we do? We won't be able to update the Android code to match a breaking API change until the OSS commit is available, but submitting the OSS commit will break the build.

We do fairly regularly make API changes in OSS which will require a matching change in the Android repo whenever we next update OSS.

raggi:
Are the dependencies cyclic?

DentonGentry:
Android depends on OSS, OSS does not depend on Android.
However I think having the CI in OSS makes it impossible to fix without breaking one of the two repos briefly.

The `if-corp-were-updated` CI, being in corp, doesn't have that issue.
If this CI were in `tailscale-android` it wouldn't have this issue.

raggi:
> However I think having the CI in OSS makes it impossible to fix without breaking one of the two repos briefly.

Isn't this already true? If an API change is made, there's a break in integration, until the dependent is updated.

What about CI makes this more true?

DentonGentry:
Say this CI is in `tailscale-android`, and a breaking OSS change is checked in:
- the next time a PR is sent to `tailscale-android`, this CI will fail. The person who wanted to submit their PR needs to go update OSS before they'll be able to submit their new PR. The person can take however long they need to update OSS, leaving their PR open.
- Android main is not broken. CI doesn't run continuously on main.

If this CI is in OSS, and an OSS change is being made which breaks Android, either:
- speculatively change the Android repo before being able to update OSS, breaking the Android `main`, or
- submit the PR with broken `it-android-were-updated` CI, breaking OSS `main`, then go immediately to the Android repo and update OSS. OSS `main` is broken until someone submits another, unrelated PR which will notice that Android has been updated.

With the CI in Android, you don't ever have to break `main`.
With the CI in OSS, one of the two repos is going to have a broken `main` for a while, which isn't a great look.

raggi:
This bug was written in part triggered by the fact that I'd been able to land code that won't build for Android into OSS, and we didn't notice until we tried to build Android. One primary goal is to add a check here that would prevent this situation. In the loosest approximation, this involves doing a build here with GOOS=android, and that must pass in order to submit.

[...]

PR #5687, opened 2022-09-18
.github/workflows: add cross-android

bradfitz:
This would've caught the regression from 7c49db02a before it was submitted so 42f1d92ae020c wouldn't have been necessary to fix it.

Updates #4482

### DIFF
--- a/.github/workflows/cross-android.yml
+++ b/.github/workflows/cross-android.yml
@@ -1,0 +1,54 @@
+name: Android-Cross
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - '*'
+
+concurrency:
+  group: ${{ github.workflow }}-$${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    if: "!contains(github.event.head_commit.message, '[ci skip]')"
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/cross-android.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: Android smoke build
+      # Super minimal Android build that doesn't even use CGO and doesn't build everything that's needed
+      # and is only arm64. But it's a smoke build: it's not meant to catch everything. But it'll catch
+      # some Android breakages early.
+      # TODO(bradfitz): better; see https://github.com/tailscale/tailscale/issues/4482
+      env:
+        GOOS: android
+        GOARCH: arm64
+      run: go install ./net/netns ./ipn/ipnlocal ./wgengine/magicsock/ ./wgengine/ ./wgengine/router/ ./util/dnsname/ ./ipn/ ./net/interfaces ./wgengine/router/ ./tailcfg/ ./types/logger/ ./net/dns
```
DentonGentry, line 39: If they can build via this mechanism, recommend adding `./hostinfo` and `./version` as well.

DentonGentry (approved):

.github/workflows/cross-android.yml
bradfitz: They're included already via deps from other stuff, but will add them explicitly as this list is already a hodgepodge of high level and low level stuff.

PR #5688, opened 2022-09-18
version/distro: detect Ubuntu separately from Debian

bradfitz:
(for an upcoming change where I wanted to easily distinguish these two)

### DIFF
--- a/version/distro/distro.go
+++ b/version/distro/distro.go
@@ -10,13 +10,16 @@ import (
 	"runtime"
 	"strconv"
 
+	"go4.org/mem"
 	"tailscale.com/syncs"
+	"tailscale.com/util/lineread"
 )
 
 type Distro string
 
 const (
 	Debian    = Distro("debian")
+	Ubuntu    = Distro("ubuntu")
 	Arch      = Distro("arch")
 	Synology  = Distro("synology")
 	OpenWrt   = Distro("openwrt")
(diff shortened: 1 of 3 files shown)


version/distro/distro.go
```diff
@@ -62,9 +65,20 @@ func linuxDistro() Distro {
...
 		return Synology
 	case have("/usr/local/bin/freenas-debug"):
-		// TrueNAS Scale runs on debian
+		// TrueNAS Scale runs on debian so test for it before Debian.
 		return TrueNAS
 	case have("/etc/debian_version"):
+		// Ubuntu also has an /etc/debian_version file, so see if it's actually Ubuntu.
+		isUbuntu := false
+		lineread.File("/etc/os-release", func(line []byte) error {
+			if mem.HasPrefix(mem.B(line), mem.S("ID=ubuntu")) {
```
DentonGentry, line 74: There are a largish number of almost-Ubuntu distributions where the ID=pop or zorin or elementary but ID_LIKE will be "ubuntu"

Depending on what you're planning, a LikeUbuntu() method might accommodate more of these distros.

DentonGentry (approved):

version/distro/distro.go
bradfitz: Yeah, at some point I might need a whole family tree set of methods. I'm kinda resisting the urge to do it before I need it, as fun as it sounds.

DentonGentry: Oooh there could be a `begat` package where Debian begat Ubuntu, and Ubuntu begat PopOS, and...

mihaip: That reminds me of the user agent detection logic we had at my old job. Chrome and Safari shared some bugs/workarounds because they dated to behavior from before the WebKit/Blink fork in 2013, and Opera and Edge switched rendering engines at some point, etc. so we had a "lineage" mechanism to rationalize some of that.

bradfitz:
Forgot why I wanted this.

Closing for now.

PR #5689, opened 2022-09-18
ipn/ipnlocal: add c2n method to get SSH username candidates

bradfitz:
For control to fetch a list of Tailscale SSH username candidates to filter against the Tailnet's SSH policy to present some valid candidates to a user.

Updates #3802
Updates tailscale/corp#7007

### DIFF
--- a/ipn/ipnlocal/ssh.go
+++ b/ipn/ipnlocal/ssh.go
@@ -19,11 +19,17 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"sync"
 
 	"github.com/tailscale/golang-x-crypto/ssh"
+	"go4.org/mem"
+	"golang.org/x/exp/slices"
+	"tailscale.com/tailcfg"
+	"tailscale.com/util/lineread"
 	"tailscale.com/util/mak"
 )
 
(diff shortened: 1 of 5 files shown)


ipn/ipnlocal/ssh.go
```diff
@@ -32,6 +38,70 @@ import (
...
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	if b.sshServer == nil {
+		return res, nil
+	}
+
+	add := func(u string) {
+		switch u {
+		case "nobody", "daemon", "sync":
```
andrew-d, line 53: Other things we may want to denylist that IIRC usually have a shell set: `mysql`, `postgres`.

bradfitz: They have shells set? Curious why!

andrew-d: No idea! But on my home (NixOS) server:
```
postgres:x:71:71:PostgreSQL server user:/var/lib/postgresql/13:/run/current-system/sw/bin/bash
```

maisem: Instead of hard coding it, maybe we can send a filter as a query param and only return unmatching usernames. Helps us iterate without bumping clients

bradfitz: I mean, we can also ignore things server-side.

maisem: We don't send the full user list, so it is plausible that none of the names are useful.

ipn/ipnlocal/ssh_test.go
```diff
@@ -40,3 +43,17 @@ func TestSSHKeyGen(t *testing.T) {
 		t.Errorf("got different keys on second call")
 	}
 }
+
+type fakeSSHServer struct {
+	SSHServer
+}
+
+func TestGetSSHUsernames(t *testing.T) {
+	b := new(LocalBackend)
```
maisem, line 52: skip in CI

bradfitz: checks at least that it doesn't crash

ipn/ipnlocal/ssh.go
bradfitz: I don't want to filter mysql or postgres by default: it's possible that you're running a database server and your only non-root user account is mysql or postgres.

If there's a better heuristic to exclude it, I'm all ears.

maisem (approved):

Issue #5693, opened 2022-09-19
cmd/tsconnect: make SSH connection timeout configurable


mihaip:
`mikan-ubuntu` on the corp tailnet is pretty slow to respond, and we can't reliably connect to it

PR #5695, opened 2022-09-20
cmd/tsconnect: allow SSH connection timeout to be overridden

mihaip:
5 seconds may not be enough if we're still loading the derp map and connecting to a slow machine.

Updates #5693

### DIFF
--- a/cmd/tsconnect/wasm/wasm_js.go
+++ b/cmd/tsconnect/wasm/wasm_js.go
@@ -360,6 +360,10 @@ func (s *jsSSHSession) Run() {
 	setReadFn := s.termConfig.Get("setReadFn")
 	rows := s.termConfig.Get("rows").Int()
 	cols := s.termConfig.Get("cols").Int()
+	timeoutSeconds := 5.0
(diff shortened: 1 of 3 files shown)


cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -360,14 +360,18 @@ func (s *jsSSHSession) Run() {
...
+	}
 	onDone := s.termConfig.Get("onDone")
 	defer onDone.Invoke()
 
 	writeError := func(label string, err error) {
 		writeErrorFn.Invoke(fmt.Sprintf("%s Error: %v\r\n", label, err))
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeoutSeconds*float64(time.Second)))
```
maisem, line 374: Canonical is `time.Duration(timeoutSeconds)*time.Second`

maisem (approved):

cmd/tsconnect/wasm/wasm_js.go
mihaip: But time.Duration is an `int64` so that would not allow the JS to use fractional second timeouts.

PR #5696, opened 2022-09-20
tailconfig: add tailnet field to node

nyghtowl:
Enable access to tailnet name from Node to track and avoid computing in unrelated functions like ngnix-auth

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -234,6 +234,7 @@ type Node struct {
 	ComputedName            string `json:",omitempty"` // MagicDNS base name (for normal non-shared-in nodes), FQDN (without trailing dot, for shared-in nodes), or Hostname (if no MagicDNS)
 	computedHostIfDifferent string // hostname, if different than ComputedName, otherwise empty
 	ComputedNameWithHost    string `json:",omitempty"` // either "ComputedName" or "ComputedName (computedHostIfDifferent)", if computedHostIfDifferent is set
+	TailnetName             string `json:",omitempty"` // SSO Tailnet name
 }
 
 // DisplayName returns the user-facing name for a node which should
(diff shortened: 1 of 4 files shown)


nyghtowl:
@willnorris is this what you were thinking about regarding getting the name into the Node as per https://github.com/tailscale/tailscale/pull/5516#issuecomment-1234274366

willnorris:
Yep, exactly!  Provided we're in agreement on what a "TailnetName" is given the naming discussions 😜.

willnorris:
I *think* this would be the `dbx.TailnetSID` from corp?  That would be a slight variation from what we are parsing out today, particularly for single-user Gmail tailnets.  Perviously, I think it would have been `username.gmail.com` and with `dbx.TailnetSID` it would be `username@gmail.com`.  But I honestly think that's okay.  It will be a breaking change, but for a very small number of users of an experimental part of the product.

tailcfg/tailcfg.go
```diff
@@ -234,6 +234,7 @@ type Node struct {
 	ComputedName            string `json:",omitempty"` // MagicDNS base name (for normal non-shared-in nodes), FQDN (without trailing dot, for shared-in nodes), or Hostname (if no MagicDNS)
 	computedHostIfDifferent string // hostname, if different than ComputedName, otherwise empty
 	ComputedNameWithHost    string `json:",omitempty"` // either "ComputedName" or "ComputedName (computedHostIfDifferent)", if computedHostIfDifferent is set
+	TailnetName             string `json:",omitempty"` // SSO Tailnet name
```
nyghtowl, line 237: Move out of computed and not SSO - change to TCD

nyghtowl:
Thanks for the feedback. It sounds like we may not be linking this to SID and that this needs to be allocated to the alias vs the SSO. Working to get confirmation on direction for identity.

willnorris:
sure, makes sense to be consistent with other places we identify the tailnet and use the same TCD identifier.

Issue #4432, opened 2022-04-18
NixOS machines with firewall management enabled can't use exit nodes


Xe:
### What is the issue?

When trying to use an exit node with a NixOS machine, all traffic to the outside world stops working.

The only workaround is to disable `networking.firewall.enable`, which is on by default.

### Steps to reproduce

1. Set up a NixOS vm with all default configuration except authentication [link](https://gist.github.com/Xe/96ec8586a5341aaf7201e8b993a091a2)
2. Enable Tailscale (`services.tailscale.enable = true;`)
3. Log in to Tailscale
4. Try to use an exit node
5. The internet stops working, including log uploads

### OS

Linux

### OS version

NixOS Unstable

### Tailscale version

1.22.2

### Bug report

BUG-c0601693c68cc9f6f1c4407f8af3804f06fbc6fc192a4a9c4476ed9154bd91fd-202204141517362-552ebdf5e2c22a6d

averagechris:
Thank you for looking into this @Xe I really appreciate it!

danderson:
Okay, I just hit this myself completely at random, and so as a result I went and debugged it. TL;DR: NixOS's reverse path filtering logic is too strict, and doesn't make allowances for policy routing.

More detail: Tailscale has to be careful to avoid routing loops with itself: when we add a default route into `tailscale0` for exit node use, the traffic generated by `tailscaled` (wireguard tunnels, control plane connection...) must _not_ use that route, otherwise we'll create a routing loop where the wireguard tunnels are trying to use themselves.

To make this work, we use Linux policy routing: we place all tailscale0 related routes into route table 52, and then set up a set of policy routing rules that say: if a packet has the packet mark 0x80000, then look up routes _without_ consulting table 52, so you get a routing decision as if tailscale didn't exist. Otherwise, use table 52 in route lookups.

So, by default, packets use Tailscale routes when deciding where to go, but tailscaled can set mark 0x80000 on its "underlay" connections, and those will not loop back into tailscale0, but rather will get routed as if tailscale didn't exist.

Separately, the NixOS firewall apparently recently added reverse path filtering logic very early in the firewall ruleset:

```
$ iptables -t raw -L PREROUTING
Chain PREROUTING (policy ACCEPT)
target     prot opt source               destination
nixos-fw-rpfilter  all  --  anywhere             anywhere

$ iptables -t raw -L nixos-fw-rpfilter
Chain nixos-fw-rpfilter (1 references)
target     prot opt source               destination
[...]

Issue #3310, opened 2021-11-15
Alpine linux: rp_filter=strict prevents exit node from working; save and restore fwmark?


apenwarr:
### What is the issue?

When using an exit node from an Alpine Linux client, where apparently rp_filter=1 (strict) by default, network connectivity through Tailscale seems to be blocked. A customer reported that setting rp_filter=2 (loose) fixed the problem.

Modern systemd-based distributions set rp_filter=2 by default, which created a raft of security problems when it first happens, but could be hiding this on most Linux distributions.

We don't know any reason rp_filter=1 shouldn't work. This warrants some investigation so we don't have to advise setting rp_filter=2 unnecessarily (systemd default security holes notwithstanding).

### OS

Linux

### OS version

Alpine

### Tailscale version

1.16.2

<img src="https://frontapp.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_3ul75)

apenwarr:
Customer mentioned that these two bugs sound similar: https://github.com/tailscale/tailscale/issues/2654
and https://github.com/tailscale/tailscale/issues/2564

ncfavier:
I don't know how tailscale works, but I wonder if this could be fixed using a similar trick as in `wg-quick`: https://github.com/WireGuard/wireguard-tools/commit/ebcf1ef8b1addbb888a6753fdee8f21231b656e6#diff-a62711f4efe175165bec9f288107ec4d172e8b50a6de3d592a399010a7cc2355R218

Basically saving the fwmark in the connection for outgoing packets and restoring it for incoming response packets so that rpfilter looks up the correct routing table.

DentonGentry:
So if we saved and restored the fwmark, then we wouldn't need to worry about whether the rp_filter is set to strict?

ncfavier:
That's the idea.

DentonGentry:
Also related: https://github.com/tailscale/tailscale/issues/591

joshspicer:
I'm here from the nixOS issue (https://github.com/tailscale/tailscale/issues/4432) and am hitting this today the default nixos firewall enabled. Is there any suggested workaround except for completely disabling the firewall?

DentonGentry:
`sysctl -w net.ipv4.conf.eth0.rp_filter=2`
will make the reverse path filter less strict, but allow the firewall to be left on.

[...]

PR #5703, opened 2022-09-20
net/netutil: add function to check rp_filter value

andrew-d:
Updates #4432

Change-Id: Ifc332a5747fc1feffdbb87437308cf8ecb21b0b0

### DIFF
--- a/net/netutil/ip_forward.go
+++ b/net/netutil/ip_forward.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/netip"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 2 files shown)


andrew-d:
This is what it looks like on my local machine:
```
$ sudo curl -fsS --unix-socket /tmp/tailscaled.usermode.sock http://localhost/localapi/v0/check-ip-forwarding | jq -r .Warning
IPv6 forwarding is disabled.
Subnet routes and exit nodes may not work correctly.
See https://tailscale.com/kb/1104/enable-ip-forwarding/
Interface docker0 has strict reverse-path filtering enabled
Subnet routes and exit nodes may not work correctly.
```

andrew-d:
I'm no expert on this, but I think this may not be sufficient; we may want to actually run this on _all_ calls to `tailscale up`, not just when we're advertising an exit node (which is when we currently check IP forwarding). Thoughts? I can pull this out into a new local endpoint without too much difficulty, if so?

danderson (approved):
Got an example of a network setup where RPF breaks exit nodes? It's one of the few configurations I can think of where RPF strictness shouldn't matter.

andrew-d:
@danderson - I think this is me being bad at reading comprehension; strict `rp_filter` breaks clients, not exit nodes. Pushed a change to have this run on all calls to `tailscale up` instead; thoughts?

ramyfication (commented):
Thanks for cc'ing me on this PR. I'm not familiar enough with this to give a proper review, but I've left some minor comments.

ipn/localapi/localapi.go
```diff
@@ -350,6 +352,23 @@ func (h *Handler) serveCheckIPForwarding(w http.ResponseWriter, r *http.Request)
...
+	if !h.PermitRead {
+		http.Error(w, "reverse path filtering check access denied", http.StatusForbidden)
+		return
+	}
+	var warning string
+	if err := h.b.CheckRPFilter(); err != nil {
+		warning = err.Error()
+	}
+	w.Header().Set("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(struct {
```
bradfitz, line 365: I'm not a fan of this unnamed type on both the server and client.

It's a protocol type; it's pretty important.

net/netutil/ip_forward.go
```diff
@@ -146,6 +147,87 @@ func CheckIPForwarding(routes []netip.Prefix, state *interfaces.State) (warn, er
...
+// The state param can be nil, in which case interfaces.GetState is used.
+//
+// The routes should only be advertised routes, and should not contain the
+// nodes Tailscale IPs.
+//
+// This function returns an error if it is unable to determine if reverse path
+// filtering is enabled, or a warning describing configuration issues if
+// reverse path fitering is non-functional or partly functional.
+func CheckReversePathFiltering(routes []netip.Prefix, state *interfaces.State) (warn, err error) {
+	if runtime.GOOS != "linux" {
```
bradfitz, line 163: will this ever be on anything other than Linux? Should the CLI only conditionally call this too to avoid some latency on other OSes?

ncfavier:
This would also need to check for an iptables or nftables rule invoking the `rpfilter` netfilter module; AFAICT the `rp_filter` sysctl is now deprecated.

There is hope of fixing the underlying problem though, see discussions in https://github.com/tailscale/tailscale/issues/3310#issuecomment-1271412885 and https://github.com/tailscale/tailscale/issues/4432.

andrew-d:
Force-pushed over this to just implement the "check for `rp_filter`" functionality, and will defer using it to another PR.

knyar (approved):

PR #5713, opened 2022-09-21
ipn/ipnlocal,tailcfg: Identify client using NodeKey in tka RPCs

twitchyliquid64:
Updates https://github.com/tailscale/corp/pull/7024

Makes TKA RPCs use `NodeKey` to identify the client, as we discussed. TKA isnt used anywhere yet so its safe to change things without worrying about backwards-compat.

### DIFF
--- a/tailcfg/tka.go
+++ b/tailcfg/tka.go
@@ -12,9 +12,11 @@ import (
 // TKAInitBeginRequest submits a genesis AUM to seed the creation of the
 // tailnet's key authority.
 type TKAInitBeginRequest struct {
-	// NodeID is the node of the initiating client.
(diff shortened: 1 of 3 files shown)


tailcfg/tka.go
```diff
@@ -12,9 +12,8 @@ import (
 // TKAInitBeginRequest submits a genesis AUM to seed the creation of the
 // tailnet's key authority.
 type TKAInitBeginRequest struct {
```
maisem, line 14: no CapVer?

twitchyliquid64: Done!

crawshaw (approved):

PR #5714, opened 2022-09-21
ipn/ipnlocal: return usernames when Tailscale SSH is enabled

maisem:
It was checking if the sshServer was initialized as a proxy, but that could either not have been initialized yet or Tailscale SSH could have been disabled after intialized.

Also bump tailcfg.CurrentCapabilityVersion


ipn/ipnlocal/local.go
```diff
@@ -3225,6 +3225,17 @@ func (b *LocalBackend) setNetMapLocked(nm *netmap.NetworkMap) {
 	}
 }
 
+// operatorUserName returns the current pref's OperatorUser's name), or the
```
bradfitz, line 3228: rando paren

ipn/ipnlocal/ssh.go
```diff
@@ -40,13 +40,10 @@ var keyTypes = []string{"rsa", "ecdsa", "ed25519"}
 
 func (b *LocalBackend) getSSHUsernames(req *tailcfg.C2NSSHUsernamesRequest) (*tailcfg.C2NSSHUsernamesResponse, error) {
```
bradfitz, line 41: document that this acquires b.mu now that it's not obvious?

bradfitz (approved):

PR #5718, opened 2022-09-21
cmd/tailscale/cli: [up] ignore allow-lan-access revert when no --exit-node

### LINKED ISSUE #3752: cmd/tailscale: disable up's revert protector on `--exit-node-allow-lan-access` when disabling exit node?
This is always annoying, disabling an exit node:

```
root@tsbi:~# tailscale up --exit-node=
Error: changing settings via 'tailscale up' requires mentioning all
non-default flags. To proceed, either re-run your command with --reset or
use the command below to explicitly mention the current value of
all non-default settings:

        tailscale up --exit-node= --exit-node-allow-lan-access

```

It's very verbose to have to specify `--exit-node-allow-lan-access=false` whenever I'm disabling an exit node.  (And you can't write `--exit-node-allow-lan-access` because then it complains that you can't use that option along with an empty `--exit-node=` value)

I think we might want to make an exception to our "no silent flag changes" rule and permit an implicit `--exit-node-allow-lan-access=true` => its flag default value of `false` here, but only when `--exit-node=` (blank, off).

/cc @maisem @danderson

maisem:
When we aren't using an exit node, the `ExitNodeAllowLANAccess` value is ignored. Make the CLI have the same behavior and not complain of accidental reverts of --exit-node-allow-lan-access if there is no exit-node configured.

Fixes #3752

### DIFF
--- a/cmd/tailscale/cli/cli_test.go
+++ b/cmd/tailscale/cli/cli_test.go
@@ -425,6 +425,57 @@ func TestCheckForAccidentalSettingReverts(t *testing.T) {
 			want: accidentalUpPrefix + " --hostname=foo --exit-node-allow-lan-access --exit-node=100.2.3.4",
 		},
 		{
+			name:          "exit_node_revert",
+			flags:         []string{"--hostname=foo"},
+			curExitNodeIP: netip.MustParseAddr("100.2.3.4"),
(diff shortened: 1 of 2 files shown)


bradfitz:
Do we still need this if we have `tailscale set` now?

PR #5720, opened 2022-09-22
docs/k8s: add TS prefix

hlts2:
# WHAT

I added `TS` prefix to auth key based on [READ.md](https://github.com/tailscale/tailscale/blob/main/docs/k8s/README.md)

# WHY

The documentation describes the auth key as `TS_AUTH_KEY`, but the manifest uses `AUTH_KEY`, which did not work properly.

### DIFF
--- a/docs/k8s/proxy.yaml
+++ b/docs/k8s/proxy.yaml
@@ -37,7 +37,7 @@ spec:
       valueFrom:
         secretKeyRef:
           name: tailscale-auth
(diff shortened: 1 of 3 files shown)


hlts2:
@bradfitz

Could you please review this PR? :bow:

DentonGentry (approved):

DentonGentry:
Thank you!

PR #5725, opened 2022-09-22
types/view: add ContainsNonExitSubnetRoutes func

soniaappasamy:
Signed-off-by: Sonia Appasamy <sonia@tailscale.com>

### DIFF
--- a/types/views/views.go
+++ b/types/views/views.go
@@ -216,6 +216,17 @@ func (v IPPrefixSlice) ContainsExitRoutes() bool {
 	return tsaddr.ContainsExitRoutes(v.ж.ж)
 }
 
+// ContainsNonExitSubnetRoutes reports whether v contains Subnet
+// Routes other than ExitNode Routes.
+func (v IPPrefixSlice) ContainsNonExitSubnetRoutes() bool {
+	for i := 0; i < v.Len(); i++ {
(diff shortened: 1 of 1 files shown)


types/views/views.go
```diff
@@ -216,6 +216,18 @@ func (v IPPrefixSlice) ContainsExitRoutes() bool {
 	return tsaddr.ContainsExitRoutes(v.ж.ж)
 }
 
+// ContainsNonExitSubnetRoutes reports whether v contains Subnet
+// Routes other than ExitNode Routes.
+func (v IPPrefixSlice) ContainsNonExitSubnetRoutes() bool {
+	if v.Len() == 0 {
```
bradfitz, line 222: How about:

 ```go
    for i := 0; i < v.Len(); i++ {
          if v.At(i).Bits() != 0 { return true } 
    }
    return false
```

Bit more explicit?

soniaappasamy: Logic is slightly different though, right? What if they have one exit route but not the other?

bradfitz: If they have only `0.0.0.0/0` your version would report true. Should it? I don't think so. Just because it has non-zero length and not exit nodes, that doesn't imply it's a subnet router.

bradfitz (approved):

PR #5726, opened 2022-09-22
tka: Checkpoint every 50 updates

twitchyliquid64:
Should effectively limit how far tka needs to iterate backwards in time to build a chain.

### DIFF
--- a/tka/builder.go
+++ b/tka/builder.go
@@ -6,6 +6,7 @@ package tka
 
 import (
 	"fmt"
+	"os"
 
 	"tailscale.com/types/tkatype"
 )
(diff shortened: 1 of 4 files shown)


twitchyliquid64:
The more i think about it, the more I'm wondering if 15 updates is too frequent. An AUM is probably on average ~150 bytes (32b parent + 64b sig + 60bish payload), so 15 updates is basically around 2.2k. I think its safe to let clients accumulate an order of magnitude more than that, especially given that iterating backwards through AUMs is hella efficient.

I guess the initial value is pretty arbitrary, but what if we boop it to 30?

tka/builder.go
```diff
@@ -92,8 +93,62 @@ func (b *UpdateBuilder) SetKeyMeta(keyID tkatype.KeyID, meta map[string]string)
...
+		if !hasParent {
+			// We've hit the genesis update, so the chain is shorter than the interval to checkpoint at.
+			needCheckpoint = false
+			break
+		}
+		cursor = parent
+	}
+
+	if needCheckpoint {
+		if err := b.generateCheckpoint(); err != nil {
```
sailorfrag, line 146: Should it also delete all of the earlier intermediate AUMs if it's replacing them with a checkpoint?

sailorfrag (approved):

sailorfrag:
> The more i think about it, the more I'm wondering if 15 updates is too frequent. An AUM is probably on average ~150 bytes (32b parent + 64b sig + 60bish payload), so 15 updates is basically around 2.2k. I think its safe to let clients accumulate an order of magnitude more than that, especially given that iterating backwards through AUMs is hella efficient.
> 
> I guess the initial value is pretty arbitrary, but what if we boop it to 30?

Maybe 100?

sailorfrag:
Although I'm also kind of wondering if there's a strong benefit to the small incremental AUMs instead of just making every one a checkpoint.

tka/builder.go
twitchyliquid64: I'd rather keep all the updates that lead to a new state as a kind of audit record - control will keep it forever but (eventually) clients will compact to a checkpoint. This will also let us determine if a bug in the consensus layer is caused by checkpoints vs applying one of the other AUMs.

sailorfrag: Okay, should we add some sort of verification that requires a checkpoint is equivalent to the state from evaluating everything since the last checkpoint (which, by induction, means it validates back to the beginning of time)?

twitchyliquid64: I think that ends up happening anyway? iirc `tka.Open` walks the chain to build the state for the Authority, and then we `state.applyVerifiedAUM` to compute the state we embed in the checkpoint.

sailorfrag: I mean more like `tka.Open` requires that each checkpoint matches the existing state at that point.

twitchyliquid64: Done, though we should remove that check for GA because we expect checkpoints may contain additional fields from new semantics that old clients dont understand. But a good correctness check in the meantime.

twitchyliquid64:
> Maybe 100?

Lets call splitzies at 50? :P

PR #5728, opened 2022-09-22
net/packet: fix filtering of short IPv4 fragments

### LINKED ISSUE #5727: Medium sized IPv4 fragments are dropped
### What is the issue?

IPv4 fragments up to 640 bytes are dropped.

IPv4 fragments less than 80 bytes should be dropped, but larger than 80 bytes should be fine.

### Steps to reproduce

Use an exit node, browse to http://icmpcheck.popcount.org/, observe second test fails.

_No response_

### OS

_No response_

### OS version

_No response_

### Tailscale version

_No response_

### Bug report

_No response_

raggi:
The fragment offset is an octet-offset rather than a byte offset. Extremely short fragments less than 80 bytes should be dropped, however, fragments of less than 640 bytes are ok.

Fixes #5727


net/packet/packet.go
```diff
@@ -18,7 +18,7 @@ import (
 const unknown = ipproto.Unknown
 
 // RFC1858: prevent overlapping fragment attacks.
-const minFrag = 60 + 20 // max IPv4 header + basic TCP header
+const minFragOct = (60 + 20) / 8 // max IPv4 header + basic TCP header in octets
```
DentonGentry, line 21: I agree that there is a problem, but this doesn't look right. It says the IPv4+TCP header is 10 octets, which is not true.

raggi: True, the terminology bug starts in the RFC. I'll change it to "fragment blocks".

bradfitz:
Let's add a test.

DentonGentry (approved):

raggi:
Pausing for a moment, reviewing the flags math. I flipped my test case to see it fail without the fix, and it didn't fail - spuriously. I think we're looking at the flags wrong.

raggi:
Ok, I have corrected the mask, but it's bit twiddling - please scrutinize.

net/packet/packet_test.go
```diff
@@ -265,6 +265,59 @@ var udp4ReplyDecode = Parsed{
 	Dst:     mustIPPort("5.6.7.8:123"),
 }
 
+// First TCP fragment of a packet with leading 24 bytes of 'a's
+var tcp4MediumFragmentBuffer = []byte{
+	// ip header up to checksum
```
bradfitz, line 270: nit: IP in comments. and TCP. and few cases below.

net/packet/packet.go
```diff
@@ -152,11 +152,12 @@ func (q *Parsed) decode4(b []byte) {
 	// it as Unknown. We can also treat any subsequent fragment that starts
 	// at such a low offset as Unknown.
 	fragFlags := binary.BigEndian.Uint16(b[6:8])
-	moreFrags := (fragFlags & 0x20) != 0
+	moreFrags := (fragFlags & 0x2000) != 0
```
bradfitz, line 155: nice find!

Looks right (now) to me!

bradfitz (approved):

PR #5730, opened 2022-09-22
tailcfg: add DataPlaneAuditLogID to Node

dblohm7:
We're adding a node-specific log ID to facilitate data-plane audit logging.

Updates https://github.com/tailscale/corp/issues/6991


tailcfg/tailcfg_view.go
```diff
@@ -173,6 +173,7 @@ func (v NodeView) MachineAuthorized() bool           { return v.ж.MachineAuthor
 func (v NodeView) Capabilities() views.Slice[string] { return views.SliceOf(v.ж.Capabilities) }
 func (v NodeView) ComputedName() string              { return v.ж.ComputedName }
 func (v NodeView) ComputedNameWithHost() string      { return v.ж.ComputedNameWithHost }
+func (v NodeView) DataPlaneAuditLogID() string       { return v.ж.DataPlaneAuditLogID }
```
dsnet, line 176: We talked about having two IDs:
* a node-specific audit log ID, and
* a tailnet-specific audit log ID.

Would it make sense to implement both in the same PR so we only have to bump the CapabilityVersion once?

tailcfg/tailcfg.go
```diff
@@ -80,7 +80,8 @@ type CapabilityVersion int
 //   - 41: 2022-08-30: uses 100.100.100.100 for route-less ExtraRecords if global nameservers is set
 //   - 42: 2022-09-06: NextDNS DoH support; see https://github.com/tailscale/tailscale/pull/5556
 //   - 43: 2022-09-21: clients can return usernames for SSH
-const CurrentCapabilityVersion CapabilityVersion = 43
+//   - 44: 2022-09-21: CapabilityDataPlaneAuditLogs, added Node.DataPlaneAuditLogID
+const CurrentCapabilityVersion CapabilityVersion = 44
```
bradfitz, line 84: I don't think you need to bump this. The client can't do anything new as of this commit that the server would be interested in.

dblohm7: OK.

tailcfg/tailcfg_view.go
dblohm7: Sure. The only reason why I had done it this way so far is because I needed to take care of the former in OSS before I could build things properly in corp, whereas the other one does not have this hard dependency. But it's easy enough for me to make that change at this point.

dsnet (approved):

tailcfg/tailcfg.go
```diff
@@ -235,6 +235,10 @@ type Node struct {
 	ComputedName            string `json:",omitempty"` // MagicDNS base name (for normal non-shared-in nodes), FQDN (without trailing dot, for shared-in nodes), or Hostname (if no MagicDNS)
 	computedHostIfDifferent string // hostname, if different than ComputedName, otherwise empty
 	ComputedNameWithHost    string `json:",omitempty"` // either "ComputedName" or "ComputedName (computedHostIfDifferent)", if computedHostIfDifferent is set
+
+	// DataPlaneAuditLogID is the per-node logtail ID used for data plane audit logging.
+	// It is non-empty if and only if CapabilityDataPlaneAuditLogs is set.
```
bradfitz, line 240: this "if and only if" bit isn't true as of this commit.

It's never set regardless :)

dblohm7: Well, it is in the sense that no tailnets have that feature flag turned on, so even though they're not tied in yet in corp, this still holds. But if you feel strongly enough about it... :-)

bradfitz: Yeah, but then your comment is making a promise that's a function of time rather than the code.

It means if in 3 months (when control has moved on) we sync oss back to this point in time and read the code, its promise in the client no longer holds.

PR #5731, opened 2022-09-22
net/dns: always attempt to read the OS config on macOS/iOS

maisem:
Also reconfigure DNS on iOS/macOS on link changes.


net/dns/manager.go
```diff
@@ -265,16 +265,23 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
...
+	isApple := runtime.GOOS == "darwin" || runtime.GOOS == "ios"
+	if isApple || !m.os.SupportsSplitDNS() {
 		// If the OS can't do native split-dns, read out the underlying
 		// resolver config and blend it into our config.
-		bcfg, err := m.os.GetBaseConfig()
-		if err != nil {
+		g, err := m.os.GetBaseConfig()
+		if err == nil {
+			bcfg = &g
+		} else if !isApple || err != ErrGetBaseConfigNotSupported {
```
bradfitz, line 276: is this right? It's at best confusing to read.

Maybe an extra if statement for just `else if isApple` with nothing in it would be easier to read?

maisem: Yeah I was just trying to rewrite that as well. How's this?

net/dns/manager.go
```diff
@@ -265,16 +265,26 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
 	rcfg.Routes = routes
 	ocfg.Nameservers = []netip.Addr{cfg.serviceIP()}
 
-	if m.os.SupportsSplitDNS() {
-		ocfg.MatchDomains = cfg.matchDomains()
-	} else {
+	var bcfg *OSConfig
+	isApple := runtime.GOOS == "darwin" || runtime.GOOS == "ios"
```
bradfitz, line 269: const isApple = ?

net/dns/manager.go
```diff
@@ -265,16 +265,26 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
...
+	if isApple || !m.os.SupportsSplitDNS() {
 		// If the OS can't do native split-dns, read out the underlying
 		// resolver config and blend it into our config.
-		bcfg, err := m.os.GetBaseConfig()
-		if err != nil {
+		g, err := m.os.GetBaseConfig()
```
bradfitz, line 273: what does `g` mean?

net/dns/manager.go
```diff
@@ -265,16 +265,26 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
...
+			// This is currently (2022-09-22) expected in certain iOS and macOS
+			// builds.
+		} else {
 			health.SetDNSOSHealth(err)
 			return resolver.Config{}, OSConfig{}, err
 		}
+	}
+
+	if bcfg == nil {
+		ocfg.MatchDomains = cfg.matchDomains()
```
bradfitz, line 286: this seemed easier to read before when this line was about "If SupportsSplitDNS".  

Leave a comment here?

net/dns/manager.go
```diff
@@ -265,16 +265,26 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
 	rcfg.Routes = routes
 	ocfg.Nameservers = []netip.Addr{cfg.serviceIP()}
 
-	if m.os.SupportsSplitDNS() {
-		ocfg.MatchDomains = cfg.matchDomains()
-	} else {
+	var bcfg *OSConfig
+	isApple := runtime.GOOS == "darwin" || runtime.GOOS == "ios"
+	if isApple || !m.os.SupportsSplitDNS() {
 		// If the OS can't do native split-dns, read out the underlying
```
bradfitz, line 271: Apple can't do native split DNS?

net/dns/manager.go
```diff
@@ -265,16 +265,34 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
 	rcfg.Routes = routes
 	ocfg.Nameservers = []netip.Addr{cfg.serviceIP()}
 
-	if m.os.SupportsSplitDNS() {
-		ocfg.MatchDomains = cfg.matchDomains()
-	} else {
+	var bcfg *OSConfig
```
bradfitz, line 268: // base config; non-nil if/when known

?

bradfitz (approved):

wgengine/userspace.go
```diff
@@ -1131,15 +1131,18 @@ func (e *userspaceEngine) linkChange(changed bool, cur *interfaces.State) {
...
-				e.logf("wgengine: error setting DNS config after major link change: %v", err)
-			} else {
-				e.logf("wgengine: set DNS config again after major link change")
+	if changed {
+		switch runtime.GOOS {
+		case "linux", "android", "ios", "darwin":
+			e.wgLock.Lock()
+			dnsCfg := e.lastDNSConfig
+			e.wgLock.Unlock()
+			if dnsCfg != nil {
```
bradfitz, line 1140: seems unrelated to the rest of this change?

net/dns/manager.go
```diff
@@ -265,22 +265,40 @@ func (m *Manager) compileConfig(cfg Config) (rcfg resolver.Config, ocfg OSConfig
...
+
+	if isApple || !m.os.SupportsSplitDNS() {
 		// If the OS can't do native split-dns, read out the underlying
 		// resolver config and blend it into our config.
-		bcfg, err := m.os.GetBaseConfig()
-		if err != nil {
+		cfg, err := m.os.GetBaseConfig()
+		if err == nil {
+			baseCfg = &cfg
+		} else if isApple && err == ErrGetBaseConfigNotSupported {
```
knyar, line 282: Should we change `GetBaseConfig` on macOS to actually return this error (https://github.com/tailscale/tailscale/pull/6128)? Otherwise it seems that the client compiled from source cannot configure DNS macOS:
```
health("dns-os"): error: [unexpected] unreachable
authReconfig: ra=0x100c626c0 dns=0x100c62720 0x01: [unexpected] unreachable
health("dns"): error: [unexpected] unreachable
```

maisem: Oh I thought it already did. Yes we should.

PR #5732, opened 2022-09-22
logtail: always record timestamps in UTC

dsnet:
Upstream optimizations to the Go time package will make unmarshaling of time.Time 3-6x faster. See:
* https://go.dev/cl/425116
* https://go.dev/cl/425197
* https://go.dev/cl/429862

The last optimization avoids a []byte -> string allocation if the timestamp string less than than 32B.
Unfortunately, the presence of a timezone breaks that optimization. Drop recording of timezone as this is non-essential information.

Most of the performance gains is upon unmarshal,
but there is also a slight performance benefit to
not marshaling the timezone as well.

### DIFF
--- a/logtail/logtail.go
+++ b/logtail/logtail.go
@@ -519,7 +519,7 @@ func (l *Logger) encodeText(buf []byte, skipClientTime bool, procID uint32, proc
 		b = append(b, `"logtail": {`...)
 		if !skipClientTime {
 			b = append(b, `"client_time": "`...)
-			b = now.AppendFormat(b, time.RFC3339Nano)
+			b = now.UTC().AppendFormat(b, time.RFC3339Nano)
 			b = append(b, `",`...)
 		}
 		if procID != 0 {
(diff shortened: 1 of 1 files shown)


dsnet:
\cc @JayWStapleton as someone from support to give a signal on whether timezones matter or not.

bradfitz:
This does make debugging a bit more difficult sometimes when they describe things in terms of their local time. That happens less than it used to now that we have `tailscale bugreport`, though.

I suppose instead we could add their local timezone offset to Hostinfo or something.

bradfitz (approved):

JayWStapleton:
> \cc @JayWStapleton as someone from support to give a signal on whether timezones matter or not.

I prefer it - it would be nice, but not essential to also record what tz the machine is in - not necessarily even in the logs. I'm often comparing logs across a couple of nodes, and translating the time zone each reference point is a source of error.

JayWStapleton (approved):

dblohm7 (approved):

Issue #5621, opened 2022-09-13
Detect which iptables to use in containers


bradfitz:
[we have copies of this bug scattered about as comments on other bugs, so creating one tracking bug]

On Linux, when Tailscale is running in a container, we should ask the kernel directly whether iptables-legacy or iptables-nft is the right interface and use the right binary, regardless of what the "iptables" binary is a symlink to on disk.

And if the right binary is not available, fail loudly.

That is, the host environment outside of the container could be using iptables-legacy vs iptables-nft and even if the host's Debian alternatives or equivalent is correct, the container's alternatives might be wrong.

This manifests in lots of weird iptables errors.

@andrew-d has done all the hard work for this bug already. Filing this bug to cross-reference other bugs against.

zyclonite:
what about creating an additional container tag like `-nft` or have an env to toggle?

i am running on fedora coreos and nftables by default so i tried out the following
https://github.com/zyclonite/tailscale-docker/blob/55b0edd41d406ac19aa994efb04f8afe496c8b3f/Dockerfile
and this seems to work all fine

DentonGentry:
It should be possible to detect this automatically, and much of the implementation to do so already exists. Requiring the user to use the correct container tag shouldn't be necessary.

zyclonite:
on coreos for example just detecting that could be tricky as both variants exist... checking for existing rules on one of them would also be quite a hack...
i would still prefer a way to have some toggle to force the container

rodrigc:
I don't know if replicating the logic done in Kubernetes to detect iptables vs. nftables is an option:
https://unix.stackexchange.com/a/589006

DentonGentry:
https://github.com/tailscale/tailscale/issues/391 is also relevant, if `tailscaled` can stop using `iptables`/`nftables` binaries altogether.

rodrigc:
@DentonGentry I am at [Tailscale Up](https://tailscale.dev/up) today,
and talked with a manager, Charlotte.  Charlotte mentioned that there is someone at tailscale who is scoping out a solution to address https://github.com/tailscale/tailscale/issues/391 .

That solution for that will probably solve https://github.com/tailscale/tailscale/issues/5621

rodrigc:
I verified [here](https://github.com/tailscale/tailscale/issues/391#issuecomment-1703900492) that the logic to detect nftables works inside a container

zyclonite:
[...]

PR #5734, opened 2022-09-23
util/linuxfw: add package to gather Linux firewall information

andrew-d:
This package is an initial implementation of something that can read
netfilter and iptables rules from the Linux kernel without needing to
shell out to an external utility; it speaks directly to the kernel using
syscalls and parses the data returned.

Currently this is read-only since it only knows how to parse a subset of
the available data.


maisem (commented):
I don't know if we should do this via Hostinfo, maybe c2n?

cmd/tailscale/depaware.txt
```diff
@@ -7,6 +7,12 @@ tailscale.com/cmd/tailscale dependencies: (generated by github.com/tailscale/dep
    W 💣 github.com/alexbrainman/sspi/negotiate                       from tailscale.com/net/tshttpproxy
         github.com/fxamacker/cbor/v2                                 from tailscale.com/tka
         github.com/golang/groupcache/lru                             from tailscale.com/net/dnscache
+   L    github.com/google/nftables                                   from tailscale.com/util/linuxfw
```
bradfitz, line 10: the CLI binary bringing in all this is a little surprising. And gvisor stuff below.

hostinfo/hostinfo_linux.go
```diff
@@ -168,3 +175,72 @@ func packageTypeLinux() string {
 	}
 	return ""
 }
+
+func linuxFW() *tailcfg.LinuxFW {
+	ret := &tailcfg.LinuxFW{
+		BinInfo: make(map[string]tailcfg.LinuxFWBinInfo),
+	}
+
+	n, err := linuxfw.DetectIptables()
```
bradfitz, line 184: why don't we put all this detection in its own package outside of hostinfo and have it register itself with the hostinfo package in its init?

then only tailscaled can empty-import it for its side effects.

bradfitz:
> I don't know if we should do this via Hostinfo, maybe c2n?

We're going to want to know this early at runtime for our own purposes in the Linux router code to make decisions on how to program iptables from containers.

@andrew-d, these commits/PR should reference #5621

tailcfg/tailcfg.go
```diff
@@ -721,6 +723,45 @@ func appendStructPtrDiff(base []string, pfx string, p1, p2 reflect.Value) (ret [
...
+	// Present is whether the binary is present on the system or not.
+	Present bool
+	// Version is the version string of the binary, if running it with
+	// `--version` succeeded.
+	Version string `json:",omitempty"`
+	// Error is the error encountered while trying to execute the binary to
+	// determine its version.
+	Error string `json:",omitempty"`
+	// Flavor is whether we were able to detect a "flavor" of the binary;
+	// this is only set for the "iptables" and "ip6tables" binaries.
```
bradfitz, line 761: and when it's set, what are the set/form of values it can be?

maisem:
> We're going to want to know this early at runtime for our own purposes in the Linux router code to make decisions on how to program iptables from containers.

Sure, but is hostinfo really the place to put it? This should just be on the link monitor and be refreshed on link changes

andrew-d:
@bradfitz @maisem - I rebased and updated this to only add the `util/linuxfw` package, not plumbing it into the rest of `tailscaled`, so that a future merge is smaller and more self-contained. PTAL?

util/linuxfw/linuxfw.go
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+// Package linuxfw returns the kind of firewall being used by the kernel
```
bradfitz, line 4: missing trailing period

util/linuxfw/helpers.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"bytes"
+	"encoding/hex"
+	"fmt"
+	"strings"
+	"unicode"
+)
+
+func btos(b []byte) string {
```
bradfitz, line 14: https://pkg.go.dev/golang.org/x/sys/unix#ByteSliceToString instead?

util/linuxfw/iptables.go
```diff
@@ -0,0 +1,821 @@
...
+	"errors"
+	"fmt"
+	"net"
+	"net/netip"
+	"strings"
+	"unsafe"
+
+	"github.com/josharian/native"
+	"golang.org/x/sys/unix"
+	linux_abi "gvisor.dev/gvisor/pkg/abi/linux"
```
bradfitz, line 19: underscore is a bit jarring/atypical

util/linuxfw/iptables.go
```diff
@@ -0,0 +1,821 @@
...
+			logf("iptables:   entry=%+v", ce.entry)
+		}
+	}
+	return nil
+}
+
+// DetectIptables returns the number of iptables rules that are present in the
+// system, ignoring the default "ACCEPT" rule present in the standard iptables
+// chains.
+func DetectIptables() (int, error) {
```
bradfitz, line 103: document: when does it return an error?

on 0? on error? on not supported by kernel?

util/linuxfw/iptables.go
```diff
@@ -0,0 +1,821 @@
...
+
+	// Must have space for the rest of the members
+	if err := p.assertLen(28); err != nil {
+		return entry{}, err
+	}
+
+	ret.NFCache = native.Endian.Uint32(p.getBytes(4))
+	targetOffset := int(native.Endian.Uint16(p.getBytes(2)))
+	nextOffset := int(native.Endian.Uint16(p.getBytes(2)))
+	/* Comeback = */ p.getBytes(4)
```
bradfitz, line 381: comeback?

util/linuxfw/iptables.go
```diff
@@ -0,0 +1,821 @@
...
+				ret = append(ret, 'X')
+			} else {
+				ret = append(ret, '.')
+			}
+		}
+		return string(ret)
+	}
+
+	if ip.InputInterface != "" {
+		fmt.Fprintf(&sb, `, InputInterface:%s/%s`, ip.InputInterface, translateMask(ip.InputInterfaceMask))
```
bradfitz, line 474: not a fan of backticks for string literals except 1) in regexps always, even when not needed, 2) when really needed because of newlines or lots of double quotes

util/linuxfw/linuxfw_unsupported.go
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+//go:build !linux || (linux && (386 || loong64))
```
bradfitz, line 4: the `linux &&` part is redundant but fine I guess

bradfitz (approved):

util/linuxfw/iptables.go
andrew-d: Unused field; that's the name that the kernel calls it. Expanded the comment a bit.

util/linuxfw/iptables.go
andrew-d: Switched everything to regular double-quotes.

Issue #5531, opened 2022-09-02
Tailscale 1.30.0 "Please Restart the Tailscale Windows Service" Win 11


artis3n:
### What is the issue?

Somewhere around v1.28, my Windows 11 machine stopped connection to Tailscale with the "Please restart the Tailsacle windows service" error. Restarting the windows service didn't resolve anything. I uninstalled and re-installed Tailscale, which didn't seem to have an impact. When I restart my machine, I do get the "Tailscale backend error:TUN: Error creating interface: The device is not ready for use" error I see on a couple of [older](https://github.com/tailscale/tailscale/issues/4225) [tickets](https://github.com/tailscale/tailscale/issues/4538), but none of the stated fixes seem to resolve anything for me. Upon uninstalling, I have no Tailscale network adapter or driver leftover in the system. I deleted the C:\Users\<me>\AppData\Local\Tailscale folder. The log files inside the directory were empty anyway. Restarting the machine + reinstalling Tailscale, immediately have the same errors.

Is there a way for me to generate logs? I am unable to successfully run `tailscale bugreport`.

![image](https://user-images.githubusercontent.com/6969296/188056736-5ee5af5f-291c-4948-a3f0-0df821a113c3.png)

![image](https://user-images.githubusercontent.com/6969296/188056750-d359da4f-7f23-4f17-a62c-a2d80bbddbf5.png)

### Steps to reproduce

:shrug:
At the moment, fresh install of Tailscale 1.30 on Win 11 is the only reproduction I can determine, although I'm sure there's some corrupt artifact of something lying around.

### OS

Windows

### OS version

11

### Tailscale version

1.30.0

DentonGentry:
> I deleted the C:\Users\<me>\AppData\Local\Tailscale folder.

I'd recommend also deleting `C:\ProgramData\Tailscale` and removing the device from https://login.tailscale.com/admin/machines

bradfitz:
Can you copy/paste that error dialog message here so I don't have to type all that hex from the screenshot?

(You can Ctrl-C it in Windows when that MsgBox is focused, even if you can't select the text)

danp:
If it helps, iOS image text extraction yields:

8ddd2cf82c5ef07a805e23d029ffae43b257bca0552ea3a7102eaa
08205ed9e0

artis3n:
[...]

PR #5739, opened 2022-09-23
cmd/tailscaled, util/winutil: log Windows service diagnostics when th…

dblohm7:
…e wintun device fails to install

I added new functions to winutil to obtain the state of a service and all its depedencies, serialize them to JSON, and write them to a Logf.

When tstun.New returns a wrapped ERROR_DEVICE_NOT_AVAILABLE, we know that wintun installation failed. We then log the service graph rooted at "NetSetupSvc". We are interested in that specific service because network devices will not install if that service is not running.

Updates https://github.com/tailscale/tailscale/issues/5531

Here is sample output:
```
State of service "NetSetupSvc" and its dependencies:
[
  {
    "serviceName": "NetSetupSvc",
    "serviceType": "WIN32",
    "state": "RUNNING",
    "startupType": "DEMAND_START",
    "triggers": [
      {
        "triggerType": 7,
        "action": 1,
        "triggerSubtype": "{2D7A2816-0C5E-45FC-9CE7-570E5ECDE9C9}",
        "dataItems": [
          {
            "dataType": 1,
            "data": "7508bca3230a8f12"
          }
        ]
      },
      {
        "triggerType": 7,
        "action": 1,
        "triggerSubtype": "{2D7A2816-0C5E-45FC-9CE7-570E5ECDE9C9}",
        "dataItems": [
          {
            "dataType": 1,
            "data": "7510bca3230a8f12"
          }
        ]
      },
      {
        "triggerType": 6,
        "action": 1,
        "triggerSubtype": "{BC90D167-9470-4139-A9BA-BE0BBBF5B74D}",
        "dataItems": [
          {
            "dataType": 2,
            "data": "610031003100310066003100630035002d0035003900320033002d0034003700630030002d0039006100360038002d006400300062006100660062003500370037003900300031000000"
          }
        ]
      }
    ]
  },
  {
    "serviceName": "RpcSs",
    "serviceType": "WIN32",
    "state": "RUNNING",
    "startupType": "AUTO_START"
  },
  {
    "serviceName": "RpcEptMapper",
    "serviceType": "WIN32",
    "state": "RUNNING",
    "startupType": "AUTO_START"
  },
  {
    "serviceName": "DcomLaunch",
    "serviceType": "WIN32",
    "state": "RUNNING",
    "startupType": "AUTO_START"
  }
]
```


util/winutil/svcdiag_windows.go
```diff
@@ -0,0 +1,297 @@
...
+		if err != nil {
+			logf("Failed retrieving Status for service %q: %v", svc.Name, err)
+		}
+
+		logEntries = append(logEntries, makeLogEntry(svc, status, config))
+	}
+
+	err := walkServices(rootSvcName, walkFn)
+	if err != nil {
+		logf("LogSvcState: %v", err)
```
andrew-d, line 36:
```suggestion
		logf("LogSvcState: error: %v", err)
```

Nitpick, so feel free to ignore: you can put `error: ` here and drop the `failed ...` from the `fmt.Errorf` calls below.

util/winutil/svcdiag_windows.go
```diff
@@ -0,0 +1,297 @@
...
+
+	deps := []*mgr.Service{rootSvc}
+	seen := make(map[string]struct{})
+
+	for err == nil && len(deps) > 0 {
+		err = func() error {
+			curSvc := deps[0]
+			defer curSvc.Close()
+
+			deps = deps[1:]
```
andrew-d, line 82: Popping from the front of a slice and pushing to the end allocates in proportion to the total # of entries, vs. a more stack-like approach where we both push/pop from the end of the slice. Unsure if order matters here, and it looks like there's not that many entries for what we're using it for, so up to you.

util/winutil/svcdiag_windows.go
```diff
@@ -0,0 +1,297 @@
...
+	if ti.triggerSubtype != nil {
+		m.TriggerSubtype = ti.triggerSubtype.String()
+	}
+	return json.Marshal(m)
+}
+
+type _SERVICE_TRIGGER_INFO struct {
+	cTriggers uint32
+	pTriggers *_SERVICE_TRIGGER
+	_         *byte
```
andrew-d, line 182: Could you add a comment explaining why we have this? e.g.
```suggestion
_/* pReserved */ *byte
```

util/winutil/svcdiag_windows.go
```diff
@@ -0,0 +1,297 @@
...
+		entry.StartupType = "SYSTEM_START"
+	case windows.SERVICE_AUTO_START:
+		if cfg.DelayedAutoStart {
+			entry.StartupType = "DELAYED_AUTO_START"
+		} else {
+			entry.StartupType = "AUTO_START"
+		}
+	case windows.SERVICE_DEMAND_START:
+		entry.StartupType = "DEMAND_START"
+		if triggerInfo, err := getSvcTriggerInfo(svc); err == nil {
```
andrew-d, line 261: Do we want an error field here?

andrew-d (commented):
Generally LGTM, with a few small requests. One other thing I noted: there's a few `unsafe.Slice`s that don't check bounds; are we assuming that Windows will never return a bad offset/count?

util/winutil/svcdiag_windows.go
```diff
@@ -0,0 +1,297 @@
...
+				if _, ok := seen[depName]; ok {
+					continue
+				}
+
+				depSvc, err := openServiceForRead(scm, depName)
+				if err != nil {
+					return fmt.Errorf("failed opening service %q: %w", depName, err)
+				}
+
+				deps = append(deps, depSvc)
```
sailorfrag, line 103: I think you want to set `seen[depSvc.Name]` here otherwise you can end up with multiple copies of the same not-yet-visited dependency?

util/winutil/svcdiag_windows.go
```diff
@@ -0,0 +1,297 @@
...
+			}
+
+			return nil
+		}()
+	}
+
+	// On error, deps might be non-empty. Its services are still open and must be closed.
+	for _, dep := range deps {
+		dep.Close()
+	}
```
sailorfrag, line 113: This could maybe be a defer?

util/winutil/svcdiag_windows.go
dblohm7: Yeah, these dependency graphs are not super deep, but I suppose it doesn't hurt to avoid a bit of churn.

util/winutil/svcdiag_windows.go
dblohm7: It's not super critical for our purposes, but it doesn't hurt. I'll add one.

util/winutil/svcdiag_windows.go
dblohm7: Sure, might as well be on the safe side...

util/winutil/svcdiag_windows.go
dblohm7: `depSvc.Name` == `depName` which we've already checked.

dblohm7:
> Generally LGTM, with a few small requests. One other thing I noted: there's a few `unsafe.Slice`s that don't check bounds; are we assuming that Windows will never return a bad offset/count?

As you notice in the `getSvcTriggerInfo` function, the API in question follows the Windows idiom where the caller invokes the API twice: the first time to obtain the buffer size, and then the second time with the actual buffer. Part of that API contract is that any pointers that are output into the results point within the same memory block. So yes, we're assuming that Windows is doing the right thing here.

util/winutil/svcdiag_windows.go
```diff
@@ -0,0 +1,303 @@
...
+	json, err := json.MarshalIndent(logEntries, "", "  ")
+	if err != nil {
+		logf("Error marshaling service log entries: %v", err)
+		return
+	}
+
+	var builder strings.Builder
+	builder.WriteString(`State of service "`)
+	builder.WriteString(rootSvcName)
+	builder.WriteString(`" and its dependencies:`)
```
andrew-d, line 49: Nitpick: this doesn't escape; one could instead do:
```suggestion
	builder.WriteString(`State of service `)
	fmt.Fprintf(&builder, "%q", rootSvcName)
	builder.WriteString(` and its dependencies:`)
```

andrew-d (approved):

PR #5740, opened 2022-09-24
tsnet: use one zstd encoder for all Servers

DentonGentry:
When a Server is destroyed we run into https://github.com/klauspost/compress/issues/370 with ever-increasing memory footprint from the zstd encoder structures.

From that issue: _"Since you are just using EncodeAll you can use a single instance for all files. This is safe and allows for concurrent compression."_

Fixes https://github.com/tailscale/corp/issues/6442

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -42,6 +42,9 @@ import (
 	"tailscale.com/wgengine/netstack"
 )
 
+var encoderOnce sync.Once
+var zstdEncoder logtail.Encoder
+
 // Server is an embedded Tailscale server.
 //
 // Its exported fields may be changed until the first call to Listen.
@@ -123,6 +126,14 @@ func (s *Server) LocalClient() (*tailscale.LocalClient, error) {
 // Optional: any calls to Dial/Listen will also call Start.
 func (s *Server) Start() error {
 	hostinfo.SetPackage("tsnet")
+	encoderOnce.Do(func() {
+		// on advice from https://github.com/klauspost/compress/issues/370
+		w, err := smallzstd.NewEncoder(nil)
+		if err != nil {
+			panic(err)
+		}
+		zstdEncoder = w
+	})
 	s.initOnce.Do(s.doInit)
 	return s.initErr
 }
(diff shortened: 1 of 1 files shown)


bradfitz (commented):
I'd prefer to avoid heavy init time work and sync.Once it instead which can be deadcoded.

tsnet/tsnet.go
```diff
@@ -254,11 +265,7 @@ func (s *Server) start() (reterr error) {
 		Stderr:     io.Discard, // log everything to Buffer
 		Buffer:     s.logbuffer,
 		NewZstdEncoder: func() logtail.Encoder {
-			w, err := smallzstd.NewEncoder(nil)
```
bradfitz, line 257: The `logtail.Encoder` interface has a `Close` method.

So this change is risky... if we later correctly change logtail to fix the `Close` leak, then this code breaks when the first tsnet server shuts down, causing the other still-running Servers to have broken logtail clients.

So two options:

1) wrap the returned zstdEncoder in a type with a dummy no-op Close method,
2) just fix the missing Close in logtail, making this PR no longer required

I'm curious why (2) isn't easier.

bradfitz: I don't actually see any obviously missing Close/Shutdown call in tsnet or logtail.

Is tsnet.Server.Close being called?

DentonGentry: > Is tsnet.Server.Close being called?

I think so:
https://github.com/tailscale/corp/blob/373f00bebd69d022c44145c95c65700c74518fb2/cmd/nodecop/nodecop.go#L118

The resource being leaked is allocated at https://github.com/klauspost/compress/blob/03c136c2259e3adcfae5c4a050ce6de1f3330597/zstd/encoder.go#L87, and I wonder if storing it in a channel on the next line is confusing the GC.

bradfitz: The GC is only relevant in that it can prevent a finalizer from running. But if you're relying on a finalizer for correctness, you've already kinda lost.

So I'd worry more about where we're not explicitly closing things.

maisem: The `tsnet.Server` was not calling `Close` on the `netstack.Impl` it created in `Start`. This was fixed in #5809.

Issue #4335, opened 2022-03-31
DNS traffic routed through exit nodes instead of direct connection


sylr:
### What is the issue?

AFAIU all my DNS traffic should be direct between machines (using Exit Nodes and configured with Tailscale DNS settings) and the configured resolvers (100.X.X.X and 100.Y.Y.Y) in MagicDNS.

However, I am seeing an unexpected volume of DNS queries transiting through my Exit Nodes.

### Steps to reproduce

On the Exit Node:

```
$ tcpdump -i tailscale0 port 53 | grep pwet.com
```

Then on my machine configured with the Exit Node:

- https://thisisatestfromfirefox.pwet.com (From Firefox)
- https://thisisatestfrombrave.pwet.com (From Brave)
- `host thisisatestfromclihost.pwet.com` (from terminal)
- `doggo thisisatestfromclihost.pwet.com` (from terminal with [doggo](https://github.com/mr-karan/doggo/))

tcpdump output

```
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on tailscale0, link-type RAW (Raw IP), capture size 262144 bytes
17:08:45.524647 IP tailscale-euc1c-infra-03.company.com.beta.tailscale.net.41264 > 100.100.100.100.domain: 43223+ AAAA? thisisatestfromfirefox.pwet.com. (49)
17:08:45.525156 IP tailscale-euc1c-infra-03.company.com.beta.tailscale.net.45975 > 100.100.100.100.domain: 47571+ A? thisisatestfromfirefox.pwet.com. (49)
17:08:45.525408 IP tailscale-euc1c-infra-03.company.com.beta.tailscale.net.54917 > dns-euc1b-infra-02.company.com.beta.tailscale.net.domain: 43223+ AAAA? thisisatestfromfirefox.pwet.com. (49)
17:08:45.525475 IP tailscale-euc1c-infra-03.company.com.beta.tailscale.net.36297 > dns-euc1b-infra-02.company.com.beta.tailscale.net.domain: 47571+ A? thisisatestfromfirefox.pwet.com. (49)
17:08:45.525554 IP tailscale-euc1c-infra-03.company.com.beta.tailscale.net.38827 > dns-euc1a-infra-01.company.com.beta.tailscale.net.domain: 43223+ AAAA? thisisatestfromfirefox.pwet.com. (49)
[...]

PR #5742, opened 2022-09-24
Add an option for tailscaled to disable the peerapi-dns service.

matrixuser0:
See also https://github.com/tailscale/tailscale/issues/4335

This is useful for exit node scenarios where we prefer DNS queries to go through the regular exit node routing rather than handled by the DoH service.

Why would we ever prefer that? One reason is if we are already hosting our own DNS infrastructure, and don't want to lose the association between clients and the DNS queries they are performing for security/firewall/logging reasons.

### DIFF
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -129,6 +129,7 @@ var args struct {
 	socksAddr      string // listen address for SOCKS5 server
 	httpProxyAddr  string // listen address for HTTP proxy server
 	disableLogs    bool
+	disablePeerAPIDNS    bool
 }
 
 var (
@@ -164,6 +165,7 @@ func main() {
 	flag.StringVar(&args.birdSocketPath, "bird-socket", "", "path of the bird unix socket")
 	flag.BoolVar(&printVersion, "version", false, "print version information and exit")
 	flag.BoolVar(&args.disableLogs, "no-logs-no-support", false, "disable log uploads; this also disables any technical support")
+	flag.BoolVar(&args.disablePeerAPIDNS, "no-peerapi-dns", false, "disable peer API DNS");
 
 	if len(os.Args) > 0 && filepath.Base(os.Args[0]) == "tailscale" && beCLI != nil {
 		beCLI()
@@ -458,7 +460,9 @@ func run() error {
 	if err != nil {
 		return fmt.Errorf("ipnserver.New: %w", err)
 	}
-	ns.SetLocalBackend(srv.LocalBackend())
+	backend := srv.LocalBackend()
+	backend.SetPeerAPIEnableDNS(!args.disablePeerAPIDNS)
+	ns.SetLocalBackend(backend)
 	if err := ns.Start(); err != nil {
 		log.Fatalf("failed to start netstack: %v", err)
 	}
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -167,6 +167,7 @@ type LocalBackend struct {
 	prevIfState      *interfaces.State
 	peerAPIServer    *peerAPIServer // or nil
 	peerAPIListeners []*peerAPIListener
+	peerAPIEnableDNS bool
 	loginFlags       controlclient.LoginFlags
 	incomingFiles    map[*incomingFile]bool
 	lastStatusTime   time.Time // status.AsOf value of the last processed status update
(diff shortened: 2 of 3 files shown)


DentonGentry:
> Why would we ever prefer that? One reason is if we are already hosting our own DNS infrastructure, and don't want to lose the association between clients and the DNS queries they are performing for security/firewall/logging reasons.

Are those DNS servers also running Tailscale and reached via a Tailscale 100.64.0.0/10 address, or a public IP address?

DentonGentry:
This proposes a larger change than might be obvious. We allow a DNS configuration in https://login.tailscale.com/admin/dns with "Override local DNS" turned off, meaning there are no configured global DNS servers.

On some platforms once we enable an exit node, we can no longer access the platform's configured DNS servers from inside the sandbox in which tailscaled runs. We have to forward the queries over to the exit node for resolution.

matrixuser0:
Wow, you are very responsive, great experience. Let me explain more the scenario where we find this useful (and same is true for the commenters in https://github.com/tailscale/tailscale/pull/5742)

You have a corporate DNS server (or multiple), and it's either public or reachable via Tailscale routes (i.e. subnet routers/exit node). It's not running Tailscale itself, so doesn't have a 100.64.0.0/10 address. DNS servers are globally configured in admin panel with "Override local DNS" enabled (the intention here is that it is always used by all clients).

Now for an endpoint, you want all Internet and internal traffic to go through your own infrastructure, e.g. for monitoring/security reasons. It is seemingly impossible currently to support this scenario, since you would want the clients to go through an exit node, but if you do so, then DNS is masqueraded by the exit node's peerapi DoH service, which the clients will automatically use. It becomes impossible to distinguish DNS traffic originating from tailscaled itself (for example to communicate with Tailscale) and DNS traffic from different clients using the exit node.

> On some platforms once we enable an exit node, we can no longer access the platform's configured DNS servers from inside the sandbox in which tailscaled runs. We have to forward the queries over to the exit node for resolution.

This is well understood, and I'm sure it's a good default for most use cases to proxy DNS queries through the exit node's DoH service.

DentonGentry:
> It's not running Tailscale itself, so doesn't have a 100.64.0.0/10 address.

Is that because it can't be on the tailnet, or that it currently is not?

Traffic to other nodes on the tailnet doesn't go through the exit node, it makes a direct connection. So even today, a DNS server would see the Tailscale IP address of each individual client sending a DNS request.

matrixuser0:
The actual deployment is more complicated, I simplified it. In our specific case, the DNS service cannot have a single tailnet IP address, since it consists of a global network of different hosts. We currently use DNAT to map it to the same IP address in multiple sites. Then we configure that single IP address in Tailscale as a global DNS override.

So maybe "it can't be on the tailnet" would not be the right conclusion, but rather this is mixed hybrid infrastructure between Tailscale and more classic networking/custom routing.

DentonGentry:
We'll need to talk about this some more internally. To set expectations: that will probably be next week, so there may not be further responses on the PR for a while.

As it is, I don't think we could support functionality like this on macOS due to the sandbox we run in. Having a feature flag which can't ever work on one of our large platforms has some negative consequences.

bradfitz:
I appreciate that you're trying to scratch your own itch by writing code & took the time to learn the codebase, but this isn't the way we'd go about this. In the future (in a couple releases) we'll probably start assuming that all peers have ExitDNS available, without them even advertising it. You're trying to express a policy by turning off a feature rather than explicitly supporting expressing that policy and making the DNS configurator etc respect it.

Also, it's not clear that we want to put the policy via flags to tailscaled which are usually pretty inaccessible to users. (e.g. not available at all on the macOS GUI builds, or iOS or Android)

The policy should either live via e.g. `tailscale up --advertise-exit-node --something-new` or in the admin console as part of the controlplane-to-node protocol.

But really we need to think about this product-wise before jumping into some minimal workaround.

You're free to continue to use this yourself in your local tailscaled but be warned that it might stop working in a few releases when we start assuming the ExitDNS service is always available.

matrixuser0:
@bradfitz That's only fair and I had no illusions of actually upstreaming this. In terms of product feedback, what can be drawn from our experiences is that the current DNS configuration system is challenging to align with requirements such as full control over all internet and DNS traffic in exit node scenarios.

If you can share more about roadmap with ExitDNS and how DNS queries will be routed through the network in the future I'd be very interested.

bradfitz:
@matrixuser0, want to chat sometime? Pick some time at https://calendly.com/bradfitz?

I'll close this for now, but I do want to make sure we have a way to do the things you want to do.

bradfitz:
Per chat, as a temporary hack, you can add:

```patch
diff --git a/net/dns/resolver/tsdns.go b/net/dns/resolver/tsdns.go
index 90e447020..c0b5b113f 100644
--- a/net/dns/resolver/tsdns.go
+++ b/net/dns/resolver/tsdns.go
@@ -12,6 +12,7 @@
        "errors"
        "fmt"
        "io"
+       "log"
        "net"
        "net/netip"
        "os"
@@ -376,6 +377,8 @@ func (r *Resolver) HandlePeerDNSQuery(ctx context.Context, q []byte, from netip.
                return marshalResponse(resp)
        }
 
+       log.Printf("DNS log: %v asked for %v", from, name)
+
        switch runtime.GOOS {
        default:
                return nil, errors.New("unsupported exit node OS")
```

Issue #5706, opened 2022-09-21
all: some resource leaks detected due to missing .Close() calls


odeke-em:
### What is the issue?

I ran Orijtech Inc's resource leak static analyzer [Static Major](https://github.com/marketplace/actions/staticmajor-analyzer) on Tailscale and it produced this report

```shell
$ staticmajor -resleak=true -resleak.fuzzy ./...
/github.com/tailscale/tailscale/chirp/chirp.go:41:3: leaking resource created on line 0
/github.com/tailscale/tailscale/client/tailscale/example/servetls/servetls.go:29:1: leaking resource created on line 0
/github.com/tailscale/tailscale/cmd/derper/derper.go:316:4: leaking resource created on line 0
/github.com/tailscale/tailscale/cmd/hello/hello.go:99:3: leaking resource created on line 0
/github.com/tailscale/tailscale/wgengine/monitor/monitor.go:96:3: leaking resource created on line 0
/github.com/tailscale/tailscale/control/controlclient/auto.go:97:3: leaking resource created on line 92
/github.com/tailscale/tailscale/control/controlclient/noise.go:103:3: leaking resource created on line 0
/github.com/tailscale/tailscale/portlist/poller.go:50:3: leaking resource created on line 0
/github.com/tailscale/tailscale/net/ping/ping.go:67:4: leaking resource created on line 57
/github.com/tailscale/tailscale/net/netcheck/netcheck.go:1193:3: leaking resource created on line 1190
/github.com/tailscale/tailscale/net/netcheck/netcheck.go:1202:3: leaking resource created on line 1190
/github.com/tailscale/tailscale/net/netcheck/netcheck.go:1225:3: leaking resource created on line 1190
/github.com/tailscale/tailscale/net/netcheck/netcheck.go:1230:3: leaking resource created on line 1190
/github.com/tailscale/tailscale/net/netcheck/netcheck.go:1238:3: leaking resource created on line 1190
/github.com/tailscale/tailscale/net/netcheck/netcheck.go:1243:3: leaking resource created on line 1190
/github.com/tailscale/tailscale/net/netcheck/netcheck.go:1249:2: leaking resource created on line 1190
/github.com/tailscale/tailscale/ipn/ipnlocal/local.go:3626:1: leaking resource created on line 0
/github.com/tailscale/tailscale/wgengine/netstack/netstack.go:943:4: leaking resource created on line 926
[...]

PR #5758, opened 2022-09-26
all: fix more resource leaks found by staticmajor

odeke-em:
Updates #5706

Signed-off-by: Emmanuel T Odeke <emmanuel@orijtech.com>

### DIFF
--- a/tsnet/example/tshello/tshello.go
+++ b/tsnet/example/tshello/tshello.go
@@ -25,10 +25,13 @@ var (
 func main() {
 	flag.Parse()
 	s := new(tsnet.Server)
+	defer s.Close()
 	ln, err := s.Listen("tcp", *addr)
 	if err != nil {
 		log.Fatal(err)
 	}
+	defer ln.Close()
+
 	if *addr == ":443" {
 		ln = tls.NewListener(ln, &tls.Config{
 			GetCertificate: tailscale.GetCertificate,
--- a/cmd/tailscaled/debug.go
+++ b/cmd/tailscaled/debug.go
@@ -88,6 +88,8 @@ func runMonitor(ctx context.Context, loop bool) error {
 	if err != nil {
 		return err
 	}
+	defer mon.Close()
+
 	mon.RegisterChangeCallback(func(changed bool, st *interfaces.State) {
 		if !changed {
 			log.Printf("Link monitor fired; no change")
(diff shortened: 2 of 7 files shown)


odeke-em:
Another kind ping @bradfitz :-) I hadn't sent those fixes in the prior PR, apologies.

tsnet/example/tshello/tshello.go
```diff
@@ -29,6 +29,11 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
+	defer func() {
```
bradfitz, line 32: prefer two defer calls over a defer of a closure

bradfitz: and then you can put the s.Close after s := and the ln.Close here

odeke-em: Done, thank you @bradfitz!

tsnet/example/tshello/tshello.go
```diff
@@ -25,10 +25,13 @@ var (
 func main() {
 	flag.Parse()
 	s := new(tsnet.Server)
+	s.Close()
```
bradfitz, line 28: forgot the defer

bradfitz: ping

odeke-em: Apologies for the late reply @bradfitz! I was busy but now back home. I've made the update, please take a look!

DentonGentry:
While it is possible the tests have always been incorrect and relied on not closing the FDs corrected in this PR, the test failures do appear to be new. We cannot merge this until the tests pass.

```
2022/10/10 22:44:08 s2.Read: read tcp [::1]:32787->[::1]:42188: use of closed network connection
exit status 1
FAIL	tailscale.com/wgengine/bench	0.228s
```

odeke-em:
Yup, thanks @DentonGentry indeed, I've fixed this in an order matching the receiver and transmitter's goroutines per 
Screen Shot 2022-10-10 at 4 01 50 PM

bradfitz:
Those bench changes are kinda convoluted. I'd prefer keeping defers where they were and waiting for goroutines to stop first before starting to return.

tsnet/example/tshello/tshello.go
```diff
@@ -25,10 +25,13 @@ var (
 func main() {
 	flag.Parse()
 	s := new(tsnet.Server)
+	defer s.Close()
```
bradfitz, line 28: this is unreachable, right? all paths below log.Fatal.

same for line 33.

bradfitz:
Oh, nevermind, now I see why you did what you did. I thought that func was self-contained instead of launching random goroutines and then returning.

odeke-em:
> Oh, nevermind, now I see why you did what you did. I thought that func was self-contained instead of launching random goroutines and then returning.

Yeah, that code sets up a transmitter and receiver and expects them to run as it sets up the actual benchmarking hence why the closes had to happen that way inside the respective goroutines!

PR #5759, opened 2022-09-27
wgengine/magicsock: set UDP socket buffer sizes to 7MB

raggi:
- At high data rates more buffer space is required in order to avoid
  packet loss during any cause of delay.
- On slower machines more buffer space is required in order to avoid
  packet loss while decryption & tun writing is underway.
- On higher latency network paths more buffer space is required in order
  to overcome BDP.
- On Linux set with SO_*BUFFORCE to bypass net.core.{r,w}mem_max.
- 7MB is the current default maximum on macOS 12.6
- Windows test is omitted, as Windows does not support getsockopt for
  these options.

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -69,6 +69,11 @@ const (
 	// _linux variant.
 	discoMagic1 = 0x5453f09f
 	discoMagic2 = 0x92ac
+
+	// UDP socket read/write buffer size (7MB). The value of 7MB is chosen as it
+	// is the max supported by a default configuration of macOS. Some platforms
+	// will silently clamp the value.
+	socketBufferSize = 7 << 20
 )
 
 // useDerpRoute reports whether magicsock should enable the DERP
@@ -2893,6 +2898,7 @@ func (c *Conn) bindSocket(ruc *RebindingUDPConn, network string, curPortFate cur
 			c.logf("magicsock: unable to bind %v port %d: %v", network, port, err)
 			continue
 		}
+		trySetSocketBuffer(pconn, c.logf)
 		// Success.
 		ruc.setConnLocked(pconn)
 		if network == "udp4" {
@@ -3946,6 +3952,20 @@ func (de *endpoint) handlePongConnLocked(m *disco.Pong, di *discoInfo, src netip
 		}
 	}
 	return
+}
+
+// portableTrySetSocketBuffer sets SO_SNDBUF and SO_RECVBUF on pconn to socketBufferSize,
+// logging an error if it occurs.
+func portableTrySetSocketBuffer(pconn nettype.PacketConn, logf logger.Logf) {
+	if c, ok := pconn.(*net.UDPConn); ok {
+		// Attempt to increase the buffer size, and allow failures.
+		if err := c.SetReadBuffer(socketBufferSize); err != nil {
+			logf("magicsock: failed to set UDP read buffer size to %d: %v", socketBufferSize, err)
+		}
+		if err := c.SetWriteBuffer(socketBufferSize); err != nil {
+			logf("magicsock: failed to set UDP write buffer size to %d: %v", socketBufferSize, err)
+		}
+	}
 }
 
 // addrLatency is an IPPort with an associated latency.
--- a/wgengine/magicsock/magicsock_linux.go
+++ b/wgengine/magicsock/magicsock_linux.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"net"
 	"net/netip"
+	"syscall"
 	"time"
 	"unsafe"
 
@@ -20,6 +21,8 @@ import (
 	"tailscale.com/envknob"
 	"tailscale.com/net/netns"
 	"tailscale.com/types/key"
+	"tailscale.com/types/logger"
+	"tailscale.com/types/nettype"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


wgengine/magicsock/magicsock_linux.go
```diff
@@ -288,3 +291,24 @@ func setBPF(conn net.PacketConn, filter []bpf.RawInstruction) error {
...
+	if c, ok := pconn.(*net.UDPConn); ok {
+		rc, err := c.SyscallConn()
+		if err != nil {
+			logf("magicsock: failed to get raw syscall conn: %v, falling back to portable interface", err)
+			portableTrySetSocketBuffer(pconn, logf)
+			return
+		}
+		rc.Control(func(fd uintptr) {
+			if err := syscall.SetsockoptInt(int(fd), syscall.SOL_SOCKET, syscall.SO_RCVBUFFORCE, socketBufferSize); err != nil {
+				logf("magicsock: failed to force-set UDP read buffer size to %d: %v", socketBufferSize, err)
```
danderson, line 307: Fall back to portable impl if these sets fail?

I guess it depends how old the FORCE sockopts are. Are they liable to not work on some of the older kernels we support (e.g. olde synology) ?

wgengine/magicsock/magicsock.go
```diff
@@ -69,6 +69,14 @@ const (
...
 	discoMagic1 = 0x5453f09f
 	discoMagic2 = 0x92ac
+
+	// UDP socket read/write buffer size (12MB). This will be set using
+	// SO_RECVBUF/SO_SNDBUF, and is often clamped by the OS to a much smaller
+	// value. This buffer includes the cost of packet headers and OS internal
+	// management structures (i.e. skb's on Linux). Linux has a special
+	// implementation to use SO_RCVBUFFORCE/SO_SNDBUFFORCE to bypass the clamp -
+	// see magicksock_linux.go.
+	socketBufferSize = 12 << 20
```
danderson, line 79: Why 12M? It's not a power of 2, which makes me think it was selected somehow. It'd be nice to have some breadcrumbs for future us to remember how this was arrived at (or explicitly state that it's an arbitrary high value, if that's what it is)

wgengine/magicsock/magicsock.go
```diff
@@ -69,6 +69,14 @@ const (
 	// _linux variant.
 	discoMagic1 = 0x5453f09f
 	discoMagic2 = 0x92ac
+
+	// UDP socket read/write buffer size (12MB). This will be set using
+	// SO_RECVBUF/SO_SNDBUF, and is often clamped by the OS to a much smaller
+	// value. This buffer includes the cost of packet headers and OS internal
+	// management structures (i.e. skb's on Linux). Linux has a special
+	// implementation to use SO_RCVBUFFORCE/SO_SNDBUFFORCE to bypass the clamp -
```
danderson, line 77: No need to speak to implementation details here IMO, the comment will bitrot sooner if we ever change implementations.

danderson (approved):

wgengine/magicsock/magicsock_linux.go
raggi: they require CAP_NET_ADMIN

wgengine/magicsock/magicsock.go
raggi: It is somewhat arbitrary, but I just set it down to 7MB to fit within macOS limits.

wgengine/magicsock/magicsock.go
raggi: Ok, though it's unfortunate to not leave breadcumbs for readers. With the methods being spread about, it's almost more messy to spread it around the methods instead.

wgengine/magicsock/magicsock_unix_test.go
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build unix
```
bradfitz, line 5: these lines shouldn't be required since the filename is _unix_test.go?

raggi: I'm probably getting something wrong:

```
~ts/tailscale set-socket-buffer-size % GOOS=windows go test -c ./wgengine/magicsock 
# tailscale.com/wgengine/magicsock [tailscale.com/wgengine/magicsock.test]
wgengine/magicsock/magicsock_unix_test.go:30:37: cannot use int(fd) (value of type int) as type syscall.Handle in argument to syscall.GetsockoptInt
wgengine/magicsock/magicsock_unix_test.go:34:37: cannot use int(fd) (value of type int) as type syscall.Handle in argument to syscall.GetsockoptInt```

raggi: https://pkg.go.dev/cmd/go#hdr-Build_constraints does not seem to document unix for file suffixes, only for tags. This seems to be in common with what appears in GOROOT.

wgengine/magicsock/magicsock.go
danderson: Not a strong opinion on my part if you'd rather keep it. My usual flow is to look for uses of the constant, and from there things usually come together easily. But I don't mind that much.

jwhited (approved):

PR #5765, opened 2022-09-27
ipn/ipnlocal, tka: Implement TKA synchronization with the control plane

twitchyliquid64:
This implement the client-side of TKA sync. I've opted to implement mocks for testing on this end, I'll do the usual controlclient tests on the PR for control.

### DIFF
--- a/ipn/ipnlocal/network-lock.go
+++ b/ipn/ipnlocal/network-lock.go
@@ -55,6 +55,7 @@ func (b *LocalBackend) tkaSyncIfNeededLocked(nm *netmap.NetworkMap) error {
 		// If the feature flag is not enabled, pretend we don't exist.
 		return nil
 	}
+	ourNodeKey := b.prefs.Persist.PrivateNodeKey.Public()
 
 	isEnabled := b.tka != nil
 	wantEnabled := nm.TKAEnabled
@@ -66,17 +67,16 @@ func (b *LocalBackend) tkaSyncIfNeededLocked(nm *netmap.NetworkMap) error {
 
 		// Regardless of whether we are moving to disabled or enabled, we
 		// need information from the tka bootstrap endpoint.
-		ourNodeKey := b.prefs.Persist.PrivateNodeKey.Public()
 		b.mu.Unlock()
 		bs, err := b.tkaFetchBootstrap(ourNodeKey, ourHead)
 		b.mu.Lock()
 		if err != nil {
-			return fmt.Errorf("fetching bootstrap: %v", err)
+			return fmt.Errorf("fetching bootstrap: %w", err)
 		}
 
 		if wantEnabled && !isEnabled {
 			if err := b.tkaBootstrapFromGenesisLocked(bs.GenesisAUM); err != nil {
-				return fmt.Errorf("bootstrap: %v", err)
+				return fmt.Errorf("bootstrap: %w", err)
 			}
 			isEnabled = true
 		} else if !wantEnabled && isEnabled {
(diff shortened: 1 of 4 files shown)


ipn/ipnlocal/network-lock.go
```diff
@@ -96,12 +96,92 @@ func (b *LocalBackend) tkaSyncIfNeededLocked(nm *netmap.NetworkMap) error {
...
+			return fmt.Errorf("inform failed: %v", err)
+		}
+	}
+
+	// NOTE(tom): We could short-circuit here if our HEAD equals the
+	// control-plane's head, but we don't just so control always has a
+	// copy of all forks that clients had.
+
+	b.mu.Unlock()
+	_, err = b.tkaDoSyncSend(ourNodeKey, toSendAUMs)
```
sailorfrag, line 177: Is there anything useful to do with the return value? It seems weird to throw it away

ipn/ipnlocal/network-lock.go
```diff
@@ -96,12 +96,92 @@ func (b *LocalBackend) tkaSyncIfNeededLocked(nm *netmap.NetworkMap) error {
...
+	out.Ancestors = make([]tka.AUMHash, len(ancestors))
+	for i, a := range ancestors {
+		if err := out.Ancestors[i].UnmarshalText([]byte(a)); err != nil {
+			return tka.SyncOffer{}, fmt.Errorf("ancestor[%d].UnmarshalText: %v", i, err)
+		}
+	}
+	return out, nil
+}
+
+// tkaSyncLocked synchronizes TKA state with control. b.mu must be held
```
sailorfrag, line 121: I feel like it might be nice to do this as a non-Locked receiver, only grabbing the mutex for the small moments when needed. There's nothing wrong with this approach, it just feels like it would be better to hold the lock for the minimum time required. It doesn't really change anything if you do unmarshalling while the lock is released except that without RAII, the lock management is more annoying that way.

ipn/ipnlocal/network-lock.go
```diff
@@ -96,12 +96,92 @@ func (b *LocalBackend) tkaSyncIfNeededLocked(nm *netmap.NetworkMap) error {
 	}
 
 	if isEnabled && b.tka.authority.Head() != nm.TKAHead {
-		// TODO(tom): Implement sync
+		if err := b.tkaSyncLocked(ourNodeKey); err != nil {
+			return fmt.Errorf("tka sync: %v", err)
```
sailorfrag, line 100:
```suggestion
			return fmt.Errorf("tka sync: %w", err)
```

This happens throughout the PR

ipn/ipnlocal/network-lock.go
```diff
@@ -405,3 +485,112 @@ func (b *LocalBackend) tkaFetchBootstrap(ourNodeKey key.NodePublic, head tka.AUM
...
+	}
+
+	var req bytes.Buffer
+	if err := json.NewEncoder(&req).Encode(syncReq); err != nil {
+		return nil, fmt.Errorf("encoding request: %v", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+	if err := ctx.Err(); err != nil {
```
sailorfrag, line 527: I don't think this is necessary to check up front?

sailorfrag (approved):

ipn/ipnlocal/network-lock.go
twitchyliquid64: Not really. Ive added logging if the returned HEAD is different to ours though so we have a logged indication of a consensus failure.

ipn/ipnlocal/network-lock.go
twitchyliquid64: I'm not sure about inverting it because `b.tka` needs the lock and its used in about 4 places, and we spend a lot of time computing stuff against `b.tka.authority` and `b.tka.storage`, so really the only times it safe to release the lock are the network RPCs.

Issue #5669, opened 2022-09-16
net/dns: OSConfig with Hosts formats as gibberish


bradfitz:
```
type OSConfig struct {
	// Hosts is a map of DNS FQDNs to their IPs, which should be added to the
	// OS's hosts file. Currently, (2022-08-12) it is only populated for Windows
	// in SplitDNS mode and with Smart Name Resolution turned on.
	Hosts []*HostEntry
```

write test for `m.logf("OScfg: %+v", ocfg)` that it isn't a bunch of pointers

PR #5766, opened 2022-09-27
net/dns: format OSConfig correctly with no pointers

andrew-d:
Fixes tailscale/tailscale#5669

### DIFF
--- a/net/dns/osconfig_test.go
+++ b/net/dns/osconfig_test.go
@@ -1,0 +1,44 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package dns
+
+import (
+	"fmt"
+	"net/netip"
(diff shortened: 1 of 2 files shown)


net/dns/osconfig_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"tailscale.com/util/dnsname"
+)
+
+func TestOSConfigPrintable(t *testing.T) {
+	ocfg := OSConfig{
+		Hosts: []*HostEntry{
+			{
+				Addr:  netip.AddrFrom4([4]byte{100, 1, 2, 3}),
+				Hosts: []string{"server", "client"},
```
maisem, line 20: Add another hostentry to make sure the formatting is legible.
I would also argue that hosts maybe should go at the end and have nameservers and search domains be first to prevent against logline truncation

andrew-d: Good call; moved to the end

maisem (approved):

PR #5768, opened 2022-09-27
util/deephash: add AppendTo method

andrew-d:
This method can be used to obtain the hex-formatted deephash.Sum instance without allocations.

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -106,7 +106,18 @@ func (s1 *Sum) xor(s2 Sum) {
 }
 
 func (s Sum) String() string {
+	// Note: if we change this, keep in sync with AppendTo
 	return hex.EncodeToString(s.sum[:])
+}
+
+// AppendTo appends the string encoding of this sum (as returned by the String
+// method) to the provided byte slice and returns the extended buffer.
+func (s Sum) AppendTo(b []byte) []byte {
+	// TODO: switch to upstream implementation if accepted:
+	// https://github.com/golang/go/issues/53693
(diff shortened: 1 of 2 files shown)


util/deephash/deephash.go
```diff
@@ -106,9 +106,18 @@ func (s1 *Sum) xor(s2 Sum) {
 }
 
 func (s Sum) String() string {
+	// Note: if we change this, keep in sync with AppendSum
 	return hex.EncodeToString(s.sum[:])
 }
 
+// AppendSum appends the string encoding of this sum (as returned by the String
+// method) to the provided byte slice and returns the extended buffer.
+func (s Sum) AppendSum(b []byte) []byte {
```
dsnet, line 115: The `netip` package uses the name `AppendTo` for a method of the same signature. Should we do so as well here? The "sum" part is a little redundant given that the type is `Sum`.

util/deephash/deephash.go
```diff
@@ -106,9 +106,18 @@ func (s1 *Sum) xor(s2 Sum) {
...
+	var lb [len(s.sum) * 2]byte
+	hex.Encode(lb[:], s.sum[:])
+	return append(b, lb[:]...)
```
dsnet, line 118: Yet another example where `hex.AppendEncode` would be a superior API. See https://github.com/golang/go/issues/53693

Perhaps put a TODO here if that proposal is ever accepted?

util/deephash/deephash.go
andrew-d: SGTM; changed

dsnet (approved):

PR #5770, opened 2022-09-28
Enable filtering accepted routes via --accept-routes

raggi:


cmd/tailscale/cli/up.go
```diff
@@ -93,7 +96,7 @@ func newUpFlagSet(goos string, upArgs *upArgsT) *flag.FlagSet {
 	upf.BoolVar(&upArgs.reset, "reset", false, "reset unspecified settings to their default values")
 
 	upf.StringVar(&upArgs.server, "login-server", ipn.DefaultControlURL, "base URL of control server")
-	upf.BoolVar(&upArgs.acceptRoutes, "accept-routes", acceptRouteDefault(goos), "accept routes advertised by other Tailscale nodes")
+	upf.StringVar(&upArgs.acceptRoutes, "accept-routes", acceptRouteDefault(goos), "accept routes advertised by other Tailscale nodes")
```
DentonGentry, line 99: Do we really want to do this? We've previously declined to do so in https://github.com/tailscale/tailscale/issues/286, for a couple reasons:
- the main use-case people proposed was in handling overlapping IPv4 subnets, which we handled via 4via6
- to not encourage the complexity of network topologies that dueling advertise-routes and accept-routes would imply.

raggi: I can't speak to the prior decision - we should definitely discuss this.

I can speak to some present conditions:

- Support often provide a "workaround" for this issue by suggesting that customers advertise a /23 rather than a /24, thereby allowing prefix length to resolve route priority.
- The aforementioned workaround does not work with Linux policy routing, and has no replacement.
- As this is part of the common support vocabulary, it is quite a common customer issue.

The use case I was responding to in this case is one where the user has two subnets they want to advertise. Most of the users nodes are Linux, and they are homed in one of the two advertised subnets. Not every endpoint that they want to reach is able to be a Tailscale node at this time. They do want Tailscale nodes that live in each of the subnets to be able to reach the other subnet - as such they want to accept one of the subnet routes, but not the other.

jorhett: > * the main use-case people proposed was in handling overlapping IPv4 subnets, which we handled with 4via6 instead

Explain "handled" in the context where a Linux node routes traffic received on its LAN interface onto the Tailscale network, the reply gets re-addressed by the node which currently "owns" the router and the original requesting node rejects the reply because the IP address changed?

This is NOT addressed in any form.

PR #5777, opened 2022-09-28
logpolicy: fix, test LogHost to work as documented

bradfitz:


logpolicy/logpolicy.go
```diff
@@ -74,7 +74,9 @@ func getLogTarget() string {
 // logtail server, or the default.
 func LogHost() string {
 	if v := getLogTarget(); v != "" {
-		return v
+		if u, err := url.Parse(v); err == nil {
```
andrew-d, line 77: Should we do this is in the `sync.Once` above?

andrew-d (approved):

logpolicy/logpolicy.go
bradfitz: I considered it but there are like barely any callers and none performance sensitive so I erred on making a smaller diff

bradfitz: But if you feel strongly, you/I can in a follow-up?

andrew-d: Nah, just wanted to check 👍

logpolicy/logpolicy_test.go
```diff
@@ -0,0 +1,37 @@
...
+	}{
+		{"", "log.tailscale.io"},
+		{"http://foo.com", "foo.com"},
+		{"https://foo.com", "foo.com"},
+		{"https://foo.com/", "foo.com"},
+		{"https://foo.com:123/", "foo.com"},
+	}
+	for _, tt := range tests {
+		reset()
+		os.Setenv("TS_LOG_TARGET", tt.env)
```
maisem, line 32: t.Setenv

PR #5779, opened 2022-09-28
tstime: add ParseDuration helper function

mihaip:
More expressive than time.ParseDuration, also accepting d (days) and w (weeks) literals.


mihaip:
(moving from corp because I want to use this in the logz viewer UI too, and this seems like the best package for it)

dsnet (approved):

PR #5785, opened 2022-09-29
tailcfg: add DERPNode.DERPBasePath support

### LINKED ISSUE #4197: FR: Make the request path to the DERP server configurable

Let's say I want to share my custom DERP server (long-term & simple hostname) with some friends, but didn't want to add their nodes to my tailnet, so I can only start the `derper` without the `--verify-clients` flag. The problem is that the absence of the `--verify-clients` flag means that anyone (maybe not my friend) who knows the hostname can use my DERP server, and I can't do anything about it.

But somehow, ideally, only my friends (A & B) are using my DERP server. I like my friend A very much, so I want A to have the best possible network connectivity when using my DERP server. And I hate my friend B because B exchanges tons of packets through my DERP server every single day, which not only costs me a lot, but also makes my friend A's experience bad. So I want to put some limits on my friend B. The first solution that came to my mind was that I could easily throttle bandwidth for different DERP request paths by putting an Nginx instance in front of my DERP server. But unfortunately, the `tailscale` doesn't care about the DERP request path at all.

So, I'd like to ask is it possible to make the `tailscale` DERP request-path aware?

### How should we solve this?

Well, at least something like [`tailcfg.DERPNode.DERPPath`](https://pkg.go.dev/tailscale.com/tailcfg#DERPNode) needs to be added.

No big deal, just might cause some trouble.

### Anything else?

In fact, we can do a lot of things with the request path, such as including a token in the request path for authentication. And since the request is TLSed, it's safe to do so.

aofei:
Fixes #4197

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -39,6 +39,7 @@ import (
 var (
 	dev        = flag.Bool("dev", false, "run in localhost development mode")
 	addr       = flag.String("a", ":443", "server HTTPS listen address, in form \":port\", \"ip:port\", or for IPv6 \"[ip]:port\". If the IP is omitted, it defaults to all interfaces.")
+	basePath   = flag.String("basepath", "", "serve under a sub-path")
 	httpPort   = flag.Int("http-port", 80, "The port on which to serve HTTP. Set to -1 to disable. The listener is bound to the same IP (if any) as specified in the -a flag.")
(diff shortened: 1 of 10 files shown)


bradfitz:
Sorry, but this ended up more invasive than we'd like to maintain. For the problem listed in #4197 you could instead have multiple hostnames and use SNI routing to different backends, though. Then you just need different DNS entries.

PR #5786, opened 2022-09-29
wgengine: log subnet router decision at v1 if we have a BIRD client

andrew-d:
Updates tailscale/coral#82

Change-Id: I398d75f7e178ff7c531ca09899c82cf974fc30c9


wgengine/userspace.go
```diff
@@ -853,6 +853,7 @@ func (e *userspaceEngine) Reconfig(cfg *wgcfg.Config, routerCfg *router.Config,
 	isSubnetRouter := false
 	if e.birdClient != nil && nm != nil && nm.SelfNode != nil {
 		isSubnetRouter = hasOverlap(nm.SelfNode.PrimaryRoutes, nm.Hostinfo.RoutableIPs)
+		e.logf("[v1] Reconfig: isSubnetRouter=%v because of hasOverlap(%v, %v)", isSubnetRouter, nm.SelfNode.PrimaryRoutes, nm.Hostinfo.RoutableIPs)
```
bradfitz, line 856: "=false because of foo(X, Y)" makes it sound like foo(X, Y) is true

bradfitz (approved):
but LGTM otherwise

PR #5788, opened 2022-09-29
logpolicy: directly expose the logtail server URL

dsnet:
Callers of LogHost often jump through hoops to undo the loss of information dropped by LogHost (e.g., the HTTP scheme).

### DIFF
--- a/logpolicy/logpolicy.go
+++ b/logpolicy/logpolicy.go
@@ -70,8 +70,19 @@ func getLogTarget() string {
 	return getLogTargetOnce.v
 }
 
+// LogURL is the base URL for the configured logtail server, or the default.
+// It is guaranteed to not terminate with any forward slashes.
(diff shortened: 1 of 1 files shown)


logpolicy/logpolicy.go
```diff
@@ -70,8 +70,18 @@ func getLogTarget() string {
 	return getLogTargetOnce.v
 }
 
+// LogURL is the base URL for the configured logtail server, or the default.
```
bradfitz, line 73: Document & guarantee that it doesn't end in a slash so callers can blindly append "/foo" and not worry about double slashes.

dsnet: Done.

mihaip (approved):

bradfitz (approved):

PR #5818, opened 2022-10-03
net/tunstats: new package to track per-connection counters

dsnet:
High-level API:

	type Statistics struct { ... }
	type Counts struct { TxPackets, TxBytes, RxPackets, RxBytes uint64 }
	func (*Statistics) UpdateTx([]byte)
	func (*Statistics) UpdateRx([]byte)
	func (*Statistics) Extract() map[flowtrack.Tuple]Counts

The API accepts a []byte instead of a packet.Parsed so that a future implementation can directly hash the address and port bytes, which are contiguous in most IP packets.
This will be useful for a custom concurrent-safe hashmap implementation.

### DIFF
--- a/net/tunstats/stats.go
+++ b/net/tunstats/stats.go
@@ -1,0 +1,78 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package tunstats maintains statistics about connections
+// flowing through a TUN device (which operate at the IP layer).
+package tunstats
+
+import (
+	"sync"
+
+	"tailscale.com/net/flowtrack"
+	"tailscale.com/net/packet"
+)
+
+// Statistics maintains counters for every connection.
+// All methods are safe for concurrent use.
+// The zero value is ready for use.
+type Statistics struct {
+	mu sync.Mutex
+	m  map[flowtrack.Tuple]Counts
+}
+
+// Counts are statistics about a particular connection.
+type Counts struct {
+	TxPackets uint64 `json:"txPkts,omitempty"`
+	TxBytes   uint64 `json:"txBytes,omitempty"`
+	RxPackets uint64 `json:"rxPkts,omitempty"`
+	RxBytes   uint64 `json:"rxBytes,omitempty"`
+}
+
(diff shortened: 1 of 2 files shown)


net/tunstats/stats.go
```diff
@@ -0,0 +1,78 @@
...
+// UpdateRx updates the counters for a received IP packet.
+// The source and destination of the packet are inverted with respect to
+// the source and destination in flowtrack.Tuple.
+func (s *Statistics) UpdateRx(b []byte) {
+	s.update(b, true)
+}
+
+func (s *Statistics) update(b []byte, receive bool) {
+	var p packet.Parsed
+	p.Decode(b)
```
bradfitz, line 48: Let's not reparse the packet. We should only do that once per packet at most.

dsnet: The commit description explains:
> The API accepts a []byte instead of a packet.Parsed so that a future implementation can directly hash the address and port bytes, which are contiguous in most IP packets.
This will be useful for a custom concurrent-safe hashmap implementation.

I have a prototype concurrent map implementation that is lock-free and 4-8x faster. Part of the optimization is performing `maphash.Bytes(b[12:24])` in the common case and using the `[12]byte` as a key.

I can push that prototype out as a PR if you want to take a look.

bradfitz: Are we going to do something similar for wgengine/filter then?

dsnet: Quite possibly.

`wgengine/filter` relies on `flowtrack.Cache`, which is fundamentally keyed on `flowtrack.Tuple`. The `netip.AddrPort` is not particularly well-performing as a Go map key since it contains such a large data structure for IPv6 information. Also, the zone pointer is always nil since zones (IIUC) come from the ethernet interface and never from within the packet itself.

We could declare `type TupleV4 [13]byte` and `type TupleV6 [35]byte`. Now that we have generics, we can have `Cache` be parameterized on either `TupleV4` or `TupleV6`. The `wgenginer/filter` would then maintain two caches, one dedicated to v4 and the other to v6.

dsnet: Prototype hash table if you're interested: https://github.com/tailscale/tailscale/commit/eb840ad423b8918b8a700d236577898654326bdb

dblohm7 (approved):

PR #5821, opened 2022-10-03
ipn/ipnlocal,tka: Fix bugs found by integration testing

twitchyliquid64:
* `tka.State.staticValidateCheckpoint` could call methods on a contained key prior to calling StaticValidate on that key
 * Remove broken backoff / RPC retry logic from tka methods in ipn/ipnlocal, to be fixed at a later time
 * Deduplicate `LocalBackend.sendAUMsLocked` as it does the same thing as `tkaDoSyncSend`
 * Fix `ipn.LocalBackend.NetworkLockModify()` which would attempt to take `b.mu` twice and deadlock
 * Add methods on ipnlocal.LocalBackend to be used in integration tests

### DIFF
--- a/tka/state.go
+++ b/tka/state.go
@@ -249,6 +249,10 @@ func (s *State) staticValidateCheckpoint() error {
 		if err := k.StaticValidate(); err != nil {
 			return fmt.Errorf("key[%d]: %v", i, err)
 		}
+	}
+	// NOTE: The max number of keys is constrained (512), so
+	// O(n^2) is fine.
+	for i, k := range s.Keys {
 		for j, k2 := range s.Keys {
 			if i == j {
 				continue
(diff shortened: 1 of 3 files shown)


twitchyliquid64:
https://github.com/tailscale/corp/pull/7279 is dependent on this (and the context behind it)

tka/state.go
```diff
@@ -249,6 +249,8 @@ func (s *State) staticValidateCheckpoint() error {
 		if err := k.StaticValidate(); err != nil {
 			return fmt.Errorf("key[%d]: %v", i, err)
 		}
+	}
+	for i, k := range s.Keys {
```
crawshaw, line 253: maybe add a comment saying this quadratic loop is fine because s.Keys is required to be small (<1000)

crawshaw (approved):

tka/state.go
twitchyliquid64: Done.

PR #5823, opened 2022-10-04
wgengine/magicsock: make debug-level stuff not logged by default

bradfitz:
And start of mechanism for CLI or support to enable debug logging of a component for a fixed amount of time.

Updates #1548

### DIFF
--- a/ipn/backend.go
+++ b/ipn/backend.go
@@ -168,6 +168,11 @@ type PartialFile struct {
 //     LocalBackend.userID, a string like "user-$USER_ID" (used in
 //     server mode).
 //   - on Linux/etc, it's always "_daemon" (ipn.GlobalDaemonStateKey)
+//
+// Additionally, the StateKey can be debug setting name:
+//
+//   - "_debug_magicsock_until" with value being a unix timestamp stringified
+//   - "_debug_<component>_until" with value being a unix timestamp stringified
 type StateKey string
 
 type Options struct {
--- a/ipn/ipnlocal/c2n.go
+++ b/ipn/ipnlocal/c2n.go
@@ -8,6 +8,8 @@ import (
 	"encoding/json"
 	"io"
 	"net/http"
+	"strconv"
+	"time"
 
 	"tailscale.com/tailcfg"
 	"tailscale.com/util/clientmetric"
@@ -32,6 +34,21 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
 	case "/debug/metrics":
 		w.Header().Set("Content-Type", "text/plain")
 		clientmetric.WritePrometheusExpositionFormat(w)
+	case "/debug/component-logging":
+		component := r.FormValue("component")
+		secs, _ := strconv.Atoi(r.FormValue("secs"))
+		if secs == 0 {
+			secs -= 1
+		}
+		until := time.Now().Add(time.Duration(secs) * time.Second)
+		err := b.SetComponentDebugLogging(component, until)
+		var res struct {
+			Error string `json:",omitempty"`
+		}
+		if err != nil {
+			res.Error = err.Error()
+		}
+		writeJSON(res)
 	case "/ssh/usernames":
 		var req tailcfg.C2NSSHUsernamesRequest
 		if r.Method == "POST" {
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"go4.org/netipx"
+	"golang.org/x/exp/slices"
 	"tailscale.com/client/tailscale/apitype"
 	"tailscale.com/control/controlclient"
 	"tailscale.com/doctor"
@@ -55,6 +56,7 @@ import (
 	"tailscale.com/types/views"
 	"tailscale.com/util/deephash"
 	"tailscale.com/util/dnsname"
+	"tailscale.com/util/mak"
 	"tailscale.com/util/multierr"
 	"tailscale.com/util/osshare"
 	"tailscale.com/util/systemd"
@@ -186,6 +188,7 @@ type LocalBackend struct {
 	// *.partial file to its final name on completion.
 	directFileRoot          string
 	directFileDoFinalRename bool // false on macOS, true on several NAS platforms
+	componentLogUntil       map[string]componentLogState
 
 	// statusLock must be held before calling statusChanged.Wait() or
 	// statusChanged.Broadcast().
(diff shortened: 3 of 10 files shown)


ipn/ipnlocal/local.go
```diff
@@ -267,9 +269,56 @@ func NewLocalBackend(logf logger.Logf, logid string, store ipn.StateStore, diale
...
+		mc, err := b.magicConn()
+		if err != nil {
+			return err
+		}
+		newSt := componentLogState{until: until}
+		on := now.Before(until)
+		mc.SetDebugLoggingEnabled(on)
+
+		if oldSt, ok := b.componentLogUntil[component]; ok && oldSt.timer != nil {
+			oldSt.timer.Stop()
```
andrew-d, line 310: Are we okay with the (presumably-unlikely) case where we fail to stop the timer because it's in-flight?

wgengine/magicsock/magicsock.go
```diff
@@ -3637,7 +3658,7 @@ func (de *endpoint) pingTimeout(txid stun.TxID) {
 		return
 	}
 	if debugDisco() || !de.bestAddr.IsValid() || mono.Now().After(de.trustBestAddrUntil) {
-		de.c.logf("[v1] magicsock: disco: timeout waiting for pong %x from %v (%v, %v)", txid[:6], sp.to, de.publicKey.ShortString(), de.discoShort)
+		de.c.dlogf("[v1] magicsock: disco: timeout waiting for pong %x from %v (%v, %v)", txid[:6], sp.to, de.publicKey.ShortString(), de.discoShort)
```
andrew-d, line 3661: Is this something that occurs often enough to log only in debug or something that support would find helpful? Presumably not because it's `[v1]` right now...

wgengine/magicsock/magicsock.go
```diff
@@ -430,6 +433,24 @@ type Conn struct {
 	peerLastDerp map[key.NodePublic]int
 }
 
+// SetDebugLoggingEnabled controls whether spammy debug logging is enabled.
+//
+// Note that this is currently independent from the log levels, even though
+// they're pretty correlated: debugging logs should be [v1] (or higher), but
```
andrew-d, line 439: Why would we want these logs to be `[v1]` or above? Since they're not logged by default, turning on logging is (presumably?) a signal that we actually care about these and we want to see them... which we won't in any case if `tailscaled` isn't logging verbosely.

wgengine/magicsock/magicsock.go
bradfitz: These happen all the time. It's expected that like 90% of disco pings will fail.

ipn/ipnlocal/local.go
andrew-d: I'm thinking of the following sequence:

**T -10:** `SetComponentDebugLogging(component, T=0)`
**T 0:** timer.AfterFunc starts
**T 0.1:** `SetComponentDebugLogging(component, T=5)`, which sets debugging on, and calls `timer.Stop()` which returns false
**T 0.2:** timer.AfterFunc runs, sets debugging off

Now it looks like we enabled debug logging, but it's actually off.

andrew-d: (this is also super unlikely, so probably worth ignoring)

bradfitz: Fixed, I think. PTAL in any case.

andrew-d: Yep, I think that fixes it :+1: Thanks!

ipn/ipnlocal/local.go
```diff
@@ -267,9 +269,66 @@ func NewLocalBackend(logf logger.Logf, logid string, store ipn.StateStore, diale
...
+		return fmt.Errorf("unknown component %q", component)
+	}
+
+	now := time.Now()
+	on := now.Before(until)
+	newSt := componentLogState{until: until}
+	setEnabled(on)
+
+	if oldSt, ok := b.componentLogUntil[component]; ok && oldSt.timer != nil {
+		oldSt.timer.Stop()
```
maisem, line 315: do you need to worry about the timer not stopping in time?

ipn/ipnlocal/local.go
```diff
@@ -267,9 +270,93 @@ func NewLocalBackend(logf logger.Logf, logid string, store ipn.StateStore, diale
...
 
+	for _, component := range debuggableComponents {
+		key := componentStateKey(component)
+		if v, err := store.ReadState(key); err == nil {
+			if ut, err := strconv.ParseInt(string(v), 10, 64); err == nil {
+				until := time.Unix(ut, 0)
+				if until.After(time.Now()) { // conditional to avoid log spam at start when off
+					b.SetComponentDebugLogging(component, until)
+				}
+			}
```
maisem, line 281: so much nesting :lolsob:

bradfitz: yeah but it's 5 lines so still readable

ipn/backend.go
```diff
@@ -168,6 +168,11 @@ type PartialFile struct {
 //     LocalBackend.userID, a string like "user-$USER_ID" (used in
 //     server mode).
 //   - on Linux/etc, it's always "_daemon" (ipn.GlobalDaemonStateKey)
+//
+// Additionally, the StateKey can be debug setting name:
+//
+//   - "_debug_magicsock_until" with value being a unix timestamp stringified
+//   - "_debug_<component>_until" with value being a unix timestamp stringified
```
andrew-d, line 175: Do you want both of these lines?

ipn/ipnlocal/c2n.go
```diff
@@ -32,6 +34,21 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
 	case "/debug/metrics":
 		w.Header().Set("Content-Type", "text/plain")
 		clientmetric.WritePrometheusExpositionFormat(w)
+	case "/debug/component-logging":
+		component := r.FormValue("component")
+		secs, _ := strconv.Atoi(r.FormValue("secs"))
+		if secs == 0 {
+			secs -= 1
+		}
```
andrew-d, line 42: I assume this is intentional since `secs = -1` results in a time in the past that means "disable debug logging"?

ipn/ipnlocal/local.go
andrew-d: Could add a method on `store`? e.g. `store.ReadInt(key)` and/or `store.ReadTime(key)`?

andrew-d (approved):

ipn/ipnlocal/local.go
bradfitz: store's an interface so no, not really

ipn/ipnlocal/c2n.go
bradfitz: yeah

maisem (approved):

ipn/backend.go
andrew-d: SGTM

ipn/ipnlocal/local.go
andrew-d: Something like this?

```diff
diff --git ipn/ipnlocal/local.go ipn/ipnlocal/local.go
index de2c68d6..da904aea 100644
--- ipn/ipnlocal/local.go
+++ ipn/ipnlocal/local.go
@@ -272,12 +272,10 @@ func NewLocalBackend(logf logger.Logf, logid string, store ipn.StateStore, diale
 
        for _, component := range debuggableComponents {
                key := componentStateKey(component)
-               if v, err := store.ReadState(key); err == nil {
-                       if ut, err := strconv.ParseInt(string(v), 10, 64); err == nil {
-                               until := time.Unix(ut, 0)
-                               if until.After(time.Now()) { // conditional to avoid log spam at start when off
-                                       b.SetComponentDebugLogging(component, until)
-                               }
+               if ut, err := ipn.ReadStoreInt(store, key); err == nil {
+                       until := time.Unix(ut, 0)
+                       if until.After(time.Now()) { // conditional to avoid log spam at start when off
+                               b.SetComponentDebugLogging(component, until)
                        }
                }
        }
diff --git ipn/store.go ipn/store.go
index 53aa5dae..8749398d 100644
--- ipn/store.go
+++ ipn/store.go
@@ -6,6 +6,7 @@
 
 import (
        "errors"
+       "strconv"
 )
 
 // ErrStateNotExist is returned by StateStore.ReadState when the
@@ -48,3 +49,23 @@ type StateStore interface {
        // WriteState saves bs as the state associated with ID.
        WriteState(id StateKey, bs []byte) error
 }
+
+// ReadStoreInt reads an integer from a StateStore.
+func ReadStoreInt(store StateStore, id StateKey) (int64, error) {
+       v, err := store.ReadState(id)
+       if err != nil {
+               return 0, err
+       }
+
+       ii, err := strconv.ParseInt(string(v), 10, 64)
+       if err != nil {
+               return 0, err
+       }
+
+       return ii, nil
+}
+
+// PutStoreInt puts an integer into a StateStore.
+func PutStoreInt(store StateStore, id StateKey, val int) error {
+       return store.WriteState(id, []byte(strconv.Itoa(val)))
+}
```

:shrug:

bradfitz: Done; PTAL

maisem (approved):

andrew-d (approved):

PR #5827, opened 2022-10-04
ipn/ipnlocal: Filter peers with bad signatures when tka is enabled

twitchyliquid64:


bradfitz:
Nit: lowercase verb after colon in commit message

ipn/ipnlocal/network-lock.go
```diff
@@ -39,6 +39,36 @@ type tkaState struct {
...
+			toDelete = append(toDelete, i)
+		} else {
+			if err := b.tka.authority.NodeKeyAuthorized(p.Key, p.KeySignature); err != nil {
+				b.logf("Network lock is dropping peer %v(%v) due to failed signature check: %v", p.ID, p.StableID, err)
+				toDelete = append(toDelete, i)
+			}
+		}
+	}
+
+	// nm.Peers is ordered, so deletion must be order-preserving.
```
crawshaw, line 65: algorithmically problematic.

make toDelete a map with key idx, then create a new slice by iterating over the old peers slice, skipping each element that's in the toDelete map? (you can even reuse the memory I believe)

crawshaw (commented):
✨

ipn/ipnlocal/network-lock.go
creachadair: FWIW, it is not too hard to permute a slice in-place to remove elements. It's linear time, but does not require any reallocation.

Example: https://github.com/creachadair/filter/blob/v0.1.0/filter.go#L19

It may not be worthwhile to do that, but I thought I'd point it out, since pushing into and out of a map is kind of heavyweight.

twitchyliquid64: Unfortunately I cant do the swap-to-back-and-truncate-slice trick as the ordering must remain intact.

creachadair: That's not what this code does. It preserves the order of the retained items.

twitchyliquid64: Oooo gotcha

crawshaw (approved):

PR #5830, opened 2022-10-04
cmd/tailscale, ipn: enable debug logs when --report flag is passed to bugreport

andrew-d:
Change-Id: Id22e9f4a2dcf35cecb9cd19dd844389e38c922ec

### DIFF
--- a/client/tailscale/localclient.go
+++ b/client/tailscale/localclient.go
@@ -276,6 +276,12 @@ type BugReportOpts struct {
 	// Diagnose specifies whether to print additional diagnostic information to
 	// the logs when generating this bugreport.
 	Diagnose bool
+
+	// Record specifies, if non-nil, whether to perform a bugreport
+	// "recording"–generating an initial log marker, then waiting for
+	// this channel to be closed before finishing the request, which
+	// generates another log marker.
+	Record <-chan struct{}
 }
 
 // BugReportWithOpts logs and returns a log marker that can be shared by the
@@ -284,16 +290,40 @@ type BugReportOpts struct {
 // The opts type specifies options to pass to the Tailscale daemon when
 // generating this bug report.
 func (lc *LocalClient) BugReportWithOpts(ctx context.Context, opts BugReportOpts) (string, error) {
-	var qparams url.Values
+	qparams := make(url.Values)
 	if opts.Note != "" {
 		qparams.Set("note", opts.Note)
 	}
 	if opts.Diagnose {
 		qparams.Set("diagnose", "true")
 	}
+	if opts.Record != nil {
+		qparams.Set("record", "true")
+	}
 
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	var requestBody io.Reader
+	if opts.Record != nil {
+		pr, pw := io.Pipe()
+		requestBody = pr
+
+		// This goroutine waits for the 'Record' channel to be closed,
+		// and then closes the write end of our pipe to unblock the
+		// reader.
+		go func() {
+			defer pw.Close()
+			select {
+			case <-opts.Record:
+			case <-ctx.Done():
+			}
+		}()
+	}
+
+	// lc.send might block if opts.Record != nil; see above.
 	uri := fmt.Sprintf("/localapi/v0/bugreport?%s", qparams.Encode())
-	body, err := lc.send(ctx, "POST", uri, 200, nil)
+	body, err := lc.send(ctx, "POST", uri, 200, requestBody)
 	if err != nil {
 		return "", err
 	}
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -232,12 +232,16 @@ func (h *Handler) serveBugReport(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	logMarker := fmt.Sprintf("BUG-%v-%v-%v", h.backendLogID, time.Now().UTC().Format("20060102150405Z"), randHex(8))
-	if envknob.NoLogsNoSupport() {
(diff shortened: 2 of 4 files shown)


client/tailscale/localclient.go
```diff
@@ -284,16 +290,37 @@ type BugReportOpts struct {
...
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	var requestBody io.Reader
+	if opts.Record != nil {
+		pr, pw := io.Pipe()
+		requestBody = pr
+
+		go func() {
```
maisem, line 312: what are you doing here?

client/tailscale/localclient.go
```diff
@@ -276,6 +276,12 @@ type BugReportOpts struct {
 	// Diagnose specifies whether to print additional diagnostic information to
 	// the logs when generating this bugreport.
 	Diagnose bool
+
+	// Record specifies whether to perform a bugreport "recording" -
+	// generating an initial log marker, then waiting for this channel to
+	// be closed before finishing the request, which generates another log
+	// marker.
+	Record chan struct{}
```
bradfitz, line 284: `<-chan struct{}` ?

client/tailscale/localclient.go
```diff
@@ -276,6 +276,12 @@ type BugReportOpts struct {
 	// Diagnose specifies whether to print additional diagnostic information to
 	// the logs when generating this bugreport.
 	Diagnose bool
+
+	// Record specifies whether to perform a bugreport "recording" -
+	// generating an initial log marker, then waiting for this channel to
+	// be closed before finishing the request, which generates another log
+	// marker.
```
bradfitz, line 283: "if non-nil" somewhere in here?

client/tailscale/localclient.go
```diff
@@ -284,16 +290,37 @@ type BugReportOpts struct {
...
+	if opts.Record != nil {
+		pr, pw := io.Pipe()
+		requestBody = pr
+
+		go func() {
+			select {
+			case <-opts.Record:
+				pw.Write([]byte{' '})
+				pw.Close()
+			case <-ctx.Done():
```
bradfitz, line 317: don't you still want to close it? defer pw.Close() above the select?

bradfitz (approved):

client/tailscale/localclient.go
andrew-d: Left a comment; PTAL?

client/tailscale/localclient.go
andrew-d: I think so; added.

client/tailscale/localclient.go
maisem: wouldn't a close achieve the same effect?

bradfitz (approved):

client/tailscale/localclient.go
andrew-d: Yep, looks like; just tested that out and it works fine, so pushed that change here. PTAL?

ipn/localapi/localapi.go
```diff
@@ -247,11 +251,69 @@ func (h *Handler) serveBugReport(w http.ResponseWriter, r *http.Request) {
...
+			h.b.SetComponentDebugLogging(component, time.Time{})
+		}
+	}()
+
+	// NOTE(andrew): if we have anything else we want to do while recording
+	// a bugreport, we can add it here.
+
+	// Read from the client, or fail when it's canceled.
+	done := make(chan struct{})
+	close(done)
```
maisem, line 290: I don't follow what's going on here either. You create and then immediately close the channel. Then you spawn a goroutine which blocks on closed channel, but really means the goroutine is a no-op.

Are you missing a defer?

andrew-d: ... how did this ever work 😆 😅  I mean, it did, since I tested it, but yes this is supposed to be a `defer`.

ipn/localapi/localapi.go
```diff
@@ -247,11 +251,69 @@ func (h *Handler) serveBugReport(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}()
+	var buf [1]byte
+	_, err := r.Body.Read(buf[:])
+	r.Body.Close()
+
+	switch {
+	case err == nil:
+		// good
+	case errors.Is(err, io.EOF) || errors.Is(err, http.ErrBodyReadAfterClose):
```
maisem, line 306: You shouldn't need the second error now.

maisem: The ErrBodyReadAfterClose

ipn/localapi/localapi.go
```diff
@@ -247,11 +251,69 @@ func (h *Handler) serveBugReport(w http.ResponseWriter, r *http.Request) {
...
+	go func() {
+		select {
+		case <-r.Context().Done():
+			r.Body.Close()
+		case <-done:
+		}
+	}()
+	var buf [1]byte
+	_, err := r.Body.Read(buf[:])
+	r.Body.Close()
```
maisem, line 301: Re-reading this, 
Maybe `defer r.Body.Close()` works just as well as this channel

maisem: Also just realized this is the request body which is automatically closed

https://pkg.go.dev/net/http#Request.Body

andrew-d: Just tested this and it does appear to work now that we're just waiting for the client to close the connection, not send a byte; PTAL?

maisem (approved):

ipn/localapi/localapi.go
maisem: So much simpler! Thanks

andrew-d: And same to you! I'm much happier with this now TBH.

PR #5839, opened 2022-10-05
tsweb: allow HTTPError to unwrap errors

willnorris:
This is needed for the changes in tailscale/corp#7296


bradfitz (approved):
If it's important for something else, lock it in with a test?

PR #5849, opened 2022-10-05
net/flowtrack: add json tags to Tuple

dsnet:
By convention, JSON serialization uses camelCase.
Specify such names on the Tuple type.


dsnet:
The alternative is to just declare a separate type in `tunstats`.

bradfitz (approved):

PR #5852, opened 2022-10-05
net/tstun: rename statististics method

dsnet:
Rename StatisticsEnable as SetStatisticsEnabled to be consistent with other similarly named methods.

Rename StatisticsExtract as ExtractStatistics to follow the convention where methods start with a verb.
It was originally named with Statistics as a prefix so that statistics related methods would sort well in godoc, but that property no longer holds.

### DIFF
--- a/net/tstun/wrap.go
+++ b/net/tstun/wrap.go
@@ -843,15 +843,15 @@ func (t *Wrapper) Unwrap() tun.Device {
 	return t.tdev
 }
 
-// StatisticsEnable enables per-connections packet counters.
-// StatisticsExtract must be called periodically to avoid unbounded memory use.
-func (t *Wrapper) StatisticsEnable(enable bool) {
+// SetStatisticsEnabled enables per-connections packet counters.
+// ExtractStatistics must be called periodically to avoid unbounded memory use.
+func (t *Wrapper) SetStatisticsEnabled(enable bool) {
 	t.stats.enabled.Store(enable)
 }
 
-// StatisticsExtract extracts and resets the counters for all active connections.
+// ExtractStatistics extracts and resets the counters for all active connections.
 // It must be called periodically otherwise the memory used is unbounded.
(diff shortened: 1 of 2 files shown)


net/tstun/wrap.go
```diff
@@ -843,15 +843,15 @@ func (t *Wrapper) Unwrap() tun.Device {
 	return t.tdev
 }
 
-// StatisticsEnable enables per-connections packet counters.
-// StatisticsExtract must be called periodically to avoid unbounded memory use.
-func (t *Wrapper) StatisticsEnable(enable bool) {
+// SetStatisticsEnabled enables per-connections packet counters.
+// [ExtractStatistics] must be called periodically to avoid unbounded memory use.
```
bradfitz, line 847: remove the brackets

bradfitz (approved):

net/tstun/wrap.go
dsnet: We don't want to use godoc's new ability to hotlink identifiers?

bradfitz: there's an official syntax now? oh. I thought this was just the latest of a dozen personal styles people try to use. I've seen backticks, pipes, single quotes, double quotes, ... I've always just said remove all of them since nothing was the convention.

Where are the docs on this new style?

dsnet: Proposed in https://github.com/golang/proposal/blob/master/design/51082-godocfmt.md#new-syntax-for-documentation-links, and accepted in https://github.com/golang/go/issues/51082.

dsnet: Removed it for now. I found a bug in `pkgsite` anyways, where it wasn't handling scopes properly.

bradfitz: Oh, right, I've tuned those bugs out. I thought they'd never get approved 😂

PR #5855, opened 2022-10-06
wgengine/wgcfg: plumb down audit log IDs

dsnet:
The node and domain audit log IDs are provided in the map response, but are ultimately going to be used in wgengine since that's the layer that manages the tstun.Wrapper.

Do the plumbing work to get this field passed down the stack.

### DIFF
--- a/types/netmap/netmap.go
+++ b/types/netmap/netmap.go
@@ -76,6 +76,11 @@ type NetworkMap struct {
 	// Domain is the current Tailnet name.
 	Domain string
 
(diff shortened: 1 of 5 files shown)


dblohm7 (approved):

types/netmap/netmap.go
```diff
@@ -76,6 +76,8 @@ type NetworkMap struct {
 	// Domain is the current Tailnet name.
 	Domain string
 
+	DomainAuditLogID string
```
bradfitz, line 79: docs

PR #5857, opened 2022-10-06
ipn/ipnlocal: make tkaSyncIfNeededLocked an exclusive section

twitchyliquid64:
Running corp/ipn#TestNetworkLockE2E has a 1/300 chance of failing, and deskchecking suggests thats whats happening are two netmaps are racing each other to be processed through tkaSyncIfNeededLocked. This happens in the first place because we release b.mu during network RPCs.

To fix this, we make the tka sync logic an exclusive section, so two netmaps will need to wait for tka sync to complete serially (which is what we would want anyway, as the second run through probably wont need to sync).

### DIFF
--- a/ipn/ipnlocal/network-lock.go
+++ b/ipn/ipnlocal/network-lock.go
@@ -39,6 +39,8 @@ type tkaState struct {
 
 // tkaFilterNetmapLocked checks the signatures on each node key, dropping
 // nodes from the netmap who's signature does not verify.
+//
+// b.mu must be held.
 func (b *LocalBackend) tkaFilterNetmapLocked(nm *netmap.NetworkMap) {
 	if !envknob.UseWIPCode() {
 		return // Feature-flag till network-lock is in Alpha.
@@ -70,7 +72,7 @@ func (b *LocalBackend) tkaFilterNetmapLocked(nm *netmap.NetworkMap) {
 	nm.Peers = peers
 }
 
-// tkaSyncIfNeededLocked examines TKA info reported from the control plane,
+// tkaSyncIfNeeded examines TKA info reported from the control plane,
 // performing the steps necessary to synchronize local tka state.
 //
 // There are 4 scenarios handled here:
@@ -85,13 +87,19 @@ func (b *LocalBackend) tkaFilterNetmapLocked(nm *netmap.NetworkMap) {
 //   - Everything up to date: All other cases.
 //     ∴ no action necessary.
 //
-// b.mu must be held. b.mu will be stepped out of (and back in) during network
-// RPCs.
-func (b *LocalBackend) tkaSyncIfNeededLocked(nm *netmap.NetworkMap) error {
+// tkaSyncIfNeeded immediately takes b.takeSyncLock which is held throughout,
+// and may take b.mu as required.
+func (b *LocalBackend) tkaSyncIfNeeded(nm *netmap.NetworkMap) error {
 	if !envknob.UseWIPCode() {
 		// If the feature flag is not enabled, pretend we don't exist.
 		return nil
 	}
+
+	b.tkaSyncLock.Lock() // take tkaSyncLock to make this function an exclusive section.
+	defer b.tkaSyncLock.Unlock()
+	b.mu.Lock() // take mu to protect access to synchronized fields.
+	defer b.mu.Unlock()
+
 	ourNodeKey := b.prefs.Persist.PrivateNodeKey.Public()
 
 	isEnabled := b.tka != nil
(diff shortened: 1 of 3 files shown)


maisem (commented):
Can we get in a state where the wrong netmap wins the race?

twitchyliquid64:
Now I'm wondering if doing the second mutex can cause a deadlock ... I think yes?

> Can we get in a state where the wrong netmap wins the race?

Yeah, if the first netmap completes its RPCs after the second and hence rewrites state to be one update back.

bradfitz:
> Now I'm wondering if doing the second mutex can cause a deadlock ... I think yes?

If any caller can acquire them in different orders, yes. That's why whenever multiple locks are involve you need to loudly document the rules.

twitchyliquid64:
How terrible would it be if I held `b.mu` through the entire of TKA sync, including the network RPCs?

1. TKA sync should be extremely rare compared to netmap updates
2. Most syncs are a single RPC, but disablement and enablement will be 2 & recovering from control data loss will be 2

thoughts? do you have a good sense of the consequences of `b.mu` being held for a few seconds?

ipn/ipnlocal/network-lock.go
```diff
@@ -88,6 +88,9 @@ func (b *LocalBackend) tkaFilterNetmapLocked(nm *netmap.NetworkMap) {
 // b.mu must be held. b.mu will be stepped out of (and back in) during network
 // RPCs.
 func (b *LocalBackend) tkaSyncIfNeededLocked(nm *netmap.NetworkMap) error {
+	b.tkaSyncLock.Lock()
+	defer b.tkaSyncLock.Unlock()
```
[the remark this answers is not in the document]
twitchyliquid64: Done.

twitchyliquid64:
Ive settled on inverting the logic: tkaSyncLock must be taken first, then mu. This both feels right and is a natural consequence of needing to drop and acquire mu during tka RPCs. PTAL!

twitchyliquid64:
This does appear to fix the data race:

```shell
[nix-shell:~/corp]$ go test -failfast -run TestNetworkLockE2E --count 50 ./ipn/
ok  	tailscale.io/ipn	74.482s

[nix-shell:~/corp]$ go test -failfast -run TestNetworkLockE2E --count 300 ./ipn/
ok  	tailscale.io/ipn	446.043s

[nix-shell:~/corp]$ go test -failfast -run TestNetworkLockE2E --count 300 ./ipn/
ok  	tailscale.io/ipn	447.095s

[nix-shell:~/corp]$ go test -failfast -run TestNetworkLockE2E --count 300 ./ipn/
ok  	tailscale.io/ipn	447.464s
```

ipn/ipnlocal/network-lock.go
```diff
@@ -85,13 +87,19 @@ func (b *LocalBackend) tkaFilterNetmapLocked(nm *netmap.NetworkMap) {
...
 	if !envknob.UseWIPCode() {
 		// If the feature flag is not enabled, pretend we don't exist.
 		return nil
 	}
+
+	b.tkaSyncLock.Lock() // take tkaSyncLock to make this function an exclusive section.
+	defer b.tkaSyncLock.Unlock()
+	b.mu.Lock() // take mu to protect access to synchronized fields.
+	defer b.mu.Unlock()
+
```
maisem, line 102: this will deadlock on line 115

maisem: nvm, i read that as lock and then unlock

ipn/ipnlocal/network-lock.go
```diff
@@ -85,13 +87,19 @@ func (b *LocalBackend) tkaFilterNetmapLocked(nm *netmap.NetworkMap) {
 //   - Everything up to date: All other cases.
 //     ∴ no action necessary.
 //
-// b.mu must be held. b.mu will be stepped out of (and back in) during network
-// RPCs.
-func (b *LocalBackend) tkaSyncIfNeededLocked(nm *netmap.NetworkMap) error {
+// tkaSyncIfNeeded will take b.tkaSyncLock, and periodically (while holding
+// b.tkaSyncLock) b.mu. Neither lock may be held when calling this function.
```
maisem, line 91: tkaSyncIfNeeded immediately takes b.takeSyncLock which is held through out, and may take b.mu as required.

maisem (approved):

ipn/ipnlocal/network-lock.go
twitchyliquid64: Done.

PR #5858, opened 2022-10-06
ssh/tailssh: make local port forwarding work w/o shell session

### LINKED ISSUE #5091: Tailscale SSH doesn't allow port forwarding unless a command is executed
### What is the issue?

It seems like Tailscale SSH requires me execute a command or open a shell on the server before allowing port forwarding.

### Steps to reproduce

I try to set up port forwarding with the following command:

```bash
ssh root@100.x.x.x -N -L5432:examplehost:5432
```

The connection is established successfully.

Then I try to open a connection (from a different Terminal):

```bash
nc localhost:5432
```

This fails, and an error message is printed:
`channel 2: open failed: administratively prohibited: port forwarding is disabled`

If I connect without the `-N` flag, it works.

_No response_

### OS

Linux, macOS

### OS version

Server is Ubuntu 20.04 LTS, Client is macOS 12.4

### Tailscale version

1.28.0

### Bug report

_No response_

maisem:
Fixes #5091

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -331,6 +331,17 @@ func (srv *server) newConn() (*conn, error) {
 // to the specified host and port.
 // TODO(bradfitz/maisem): should we have more checks on host/port?
 func (c *conn) mayForwardLocalPortTo(ctx ssh.Context, destinationHost string, destinationPort uint32) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.finalAction == nil {
+		// We haven't yet authenticated the user, this probably means that they
+		// are just requesting port forwarding and not a shell.
+		// We also do not have a reader or writer in this case so we can't
+		// read "Ctrl-C" or write messages to the user to prompt for check mode.
+		if _, err := c.authorizeSessionLocked(ctx, nil, nil); err != nil {
+			return false
+		}
+	}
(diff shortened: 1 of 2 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -554,31 +565,43 @@ func (srv *server) fetchPublicKeysURL(url string) ([]string, error) {
 	return lines, err
 }
 
-func (c *conn) authorizeSession(s ssh.Session) (_ *contextReader, ok bool) {
+// authorizeSession authorizes the SSH session, returning an error if
+// authorization fails.
+// If a reader is provided, it will be monitored for a "Ctrl+C" sequence.
+// If a writer is provided, it will be used to write any messages from
+// the authorization process.
+func (c *conn) authorizeSession(sctx ssh.Context, reader io.Reader, writer io.Writer) (*contextReader, error) {
```
bradfitz, line 573: document that the returned contextReader can be nil?

tempfork/gliderlabs/ssh/context.go
```diff
@@ -139,6 +141,14 @@ func (ctx *sshContext) ServerVersion() string {
 	return ctx.Value(ContextKeyServerVersion).(string)
 }
 
+func (ctx *sshContext) PublicKey() PublicKey {
```
bradfitz, line 144: per chat, get rid of this change and do it in ssh/tailssh instead?

PR #5864, opened 2022-10-06
wgengine/netlog: new package for traffic flow logging

dsnet:
The Logger type managers a logtail.Logger for extracting statistics from a tstun.Wrapper.
So long as Shutdown is called, it ensures that logtail and statistic gathering resources are properly cleared up.

### DIFF
--- a/wgengine/netlog/logger.go
+++ b/wgengine/netlog/logger.go
@@ -1,0 +1,250 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package netlog provides a logger that monitors a TUN device and
+// periodically records any traffic into a log stream.
+package netlog
+
+import (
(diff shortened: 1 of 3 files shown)


wgengine/netlog/logger.go
```diff
@@ -0,0 +1,235 @@
...
+	nl.cancel()
+	nl.mu.Unlock()
+	nl.group.Wait() // do not hold lock while waiting
+	nl.mu.Lock()
+	err := nl.logger.Shutdown(ctx)
+
+	nl.nodeID = logtail.PrivateID{}
+	nl.domainID = logtail.PrivateID{}
+	nl.logger = nil
+	nl.cancel = nil
```
dblohm7, line 233: Is it worth `nil`ing out `nl.addrs` and `nl.prefixes` too?

dblohm7 (approved):

wgengine/netlog/logger.go
dsnet: Done.

PR #5865, opened 2022-10-07
ssh/tailssh: do the full auth flow during ssh auth

### LINKED ISSUE #5091: Tailscale SSH doesn't allow port forwarding unless a command is executed

maisem:
Fixes #5091


ssh/tailssh/tailssh.go
```diff
@@ -184,46 +178,106 @@ func (c *conn) logf(format string, args ...any) {
...
+		c.logf("failed to get conninfo: %v", err)
+		return gossh.ErrDenied.Error()
+	}
+	if err := c.doPolicyAuth(ctx, nil /* no pub key */); err != nil {
+		return ""
+	}
+	action := c.currentAction
+	for {
+		if action.Reject || action.Accept || action.Message != "" {
+			return action.Message
```
bradfitz, line 238: oh, right, this is part of why we did it in stderr before instead of banner: the whole NextAction stream was made assuming that there could be N actions in a row that all emitted a banner.

The BannerCallback only lets us do one. I guess that works for now since we're not currently ever sending more than 1 from the server.

maisem: So I think the constraint that we can only show one message to the user is fine.

Specifically this change makes Tailscale SSH be more compliant with the protocol as we do actually do auth in the auth phase now

This change also means that things like LocalPortForwarding no longer need to have a shell started. The fix in https://github.com/tailscale/tailscale/pull/5858 is very specific to LocalPortForwarding, this is fixes the underlying problem instead.

ssh/tailssh/tailssh.go
```diff
@@ -48,8 +48,19 @@ var (
 	sshVerboseLogging = envknob.RegisterBool("TS_DEBUG_SSH_VLOG")
 )
 
+// LocalBackend is the subset of ipnlocal.LocalBackend that we use.
+// It is used for testing.
+type LocalBackend interface {
```
bradfitz, line 53: does it need to be exported? could call it ipnLocalBackend if you want to keep LocalBackend part capitalized.

ssh/tailssh/tailssh_test.go
```diff
@@ -211,6 +221,245 @@ func TestMatchRule(t *testing.T) {
 
 func timePtr(t time.Time) *time.Time { return &t }
 
+type localState struct {
```
bradfitz, line 224: // localState implements ipnLocalBackend for testing.

ssh/tailssh/tailssh_test.go
```diff
@@ -211,6 +221,245 @@ func TestMatchRule(t *testing.T) {
...
+	sshEnabled    bool
+	matchingRule  *tailcfg.SSHRule
+	serverActions map[string]*tailcfg.SSHAction
```
bradfitz, line 227: what's this keyed by?

bradfitz (approved):

ssh/tailssh/tailssh_test.go
maisem: just the action id, added docs.

Issue #5778, opened 2022-09-28
tsweb: varz aren't sorted


bradfitz:
Our tsweb prometheus metrics text output sorts, but it sorts before stripping the type prefix.

So if you have "counter_foo_a" and "gauge_foo_b", then "foo_a" and "foo_b" aren't next to each other as expected.

mhlakhani:
Whoops: I was playing around with go and did not realize this would send a notification :)

Any objections if I submit a PR for this in a day or two after having looked at it with fresh eyes to make sure I understood the issue correctly?

DentonGentry:
By all means, feel free to send a PR.

Also just so you know: github doesn't notify anyone via email or push notification in their app when a PR in a fork references an issue like this. The link appeared in this issue, but doesn't generate any active notification.

bradfitz:
@mhlakhani, #5778 addresses structs, but not top-level expvars.

Also, #5778 looks like it adds garbage per ServeHTTP. The sorted keys should be cacheable.

The thing we still need to do is this bit:

```
var expvarDo = expvar.Do // pulled out for tests
```

The standard library maintains a sorted list of published expvars without regard to the type prefixes. We need to maintain our own sorted list.

bradfitz:
> By garbage, I assume you're referring to this: https://github.com/tailscale/tailscale/blob/main/tsweb/tsweb.go#L669-L674 

No, that doesn't allocate. It's the slices & maps that were to sort.

I changed that to only do it once per reflect.Type in #5889

PR #5869, opened 2022-10-07
tsweb: sort varz by name after stripping prefix (#5778)

mhlakhani:
This makes it easier to view prometheus metrics.

Added a test case which demonstrates the new behavior - the test initially failed as the output was ordered in the same order as the fields were declared in the struct (i.e. foo_a, bar_a, foo_b, bar_b). For that reason, I also had to change an existing test case to sort the fields in the new expected order.

Signed-off-by: Hasnain Lakhani <m.hasnain.lakhani@gmail.com>

### DIFF
--- a/tsweb/tsweb.go
+++ b/tsweb/tsweb.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


mhlakhani:
I **think** this matches the intended goal based on the issue description, please let me know if I misunderstood and I'm happy to change.

DentonGentry (approved):
Thanks!

PR #5874, opened 2022-10-08
ipn/ipnlocal: fix taildrop target list UI bug

bradfitz:
The macOS and iOS apps that used the /localapi/v0/file-targets handler were getting too many candidate targets. They wouldn't actually accept the file. This is effectively just a UI glitch in the wrong hosts being listed as valid targets from the source side.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -3419,10 +3419,7 @@ func (b *LocalBackend) FileTargets() ([]*apitype.FileTarget, error) {
 		return nil, errors.New("file sharing not enabled by Tailscale admin")
 	}
 	for _, p := range nm.Peers {
-		if len(p.Addresses) == 0 {
-			continue
-		}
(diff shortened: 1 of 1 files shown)


maisem (approved):

ipn/ipnlocal/local.go
```diff
@@ -3422,7 +3422,7 @@ func (b *LocalBackend) FileTargets() ([]*apitype.FileTarget, error) {
 		if len(p.Addresses) == 0 {
 			continue
 		}
-		if p.User != nm.User && b.peerHasCapLocked(p.Addresses[0].Addr(), tailcfg.CapabilityFileSharing) {
+		if p.User != nm.User && !b.peerHasCapLocked(p.Addresses[0].Addr(), tailcfg.CapabilityFileSharing) {
```
maisem, line 3425: CapabilityFileSharing needs to be `CapabilityFileSharingTarget`

maisem (approved):

DentonGentry (approved):

PR #5876, opened 2022-10-08
Add flag --exit-node-use-dns

### LINKED ISSUE #5875: Exit Node (without accept-dns) breaks DNS Config from web ui
### What is the issue?

I discovered that Tailscale does not resolve my DNS requests as expected (through my pi-hole which is set as the only Global Nameserver) if I use an exit node where "accept-dns" is disabled (for other reasons).

I did not expected that setting "--accept-dns=false" on an exit node does modify the behavior of clients using this exit node but with "--accept-dns=true". I understand & accept the concept of using an exit node should let clients feel like they are in the same network as the given exit node. However, I think this should primarily apply to users which do not configure a global DNS server or did not enable "Override local DNS" in the web UI, as I did both (see at steps to reproduce).

For now to have both an exit node & DNS as I expect it, I either need to
- set "--accept-dns=true" on the server, which I cannot do, because the server's applications should not rely on using my pi-hole at home.
- use another exit node without, which heavily slows down my network connection.
- disable the exit node, which is a major privacy issue for me in cellular networks or public networks.
- ignore my DNS setup, which is a privacy & security issue as pi-hole is set up to block domains containing ads, malware and so on.
As so, I first though tailscale is perfect as it allowed me to have secure DNS requests but a fast network speed because I don't route my network connections through my Pi at home but through an hosted server. However, this limitation makes tailscale nearly unusable to me.

### Steps to reproduce

My setup looks as follows:
- I have a pi-hole available (at 10.0.0.10) at home through a subnet router (called `pi`).
  - This should not become my exit node as it has a very slow network speed, but my home network is fast enough to route DNS traffic through it.
  - I want to use my pi-hole as a DNS blocker.
- I have an exit node server (called `server`), which is hosted by a provider.
  - This server should be my exit node as it has a really fast network speed.
  - This server is used for multiple purposes. It should not depend on the pi-hole DNS as this would break many services. That's why "accept-dns" is disabled for it.
- Tailscale DNS Config (web ui): Global Nameserver is set to only 10.0.0.10 (my pi-hole). I enabled "Override local DNS".

Given that `ads.com` is blocked by the pi-hole, if I configure tailscale on my client with said exit node enabled, dns requests are not handled as I would not assume it:
```
$ tailscale up --reset --accept-dns=true --exit-node=server --exit-node-allow-lan-access=true
$ dig ads.com +short
199.59.243.60
```

However, if I disable the exit node (or choose another one without "--accept-dns=false"), DNS requests get handled as I expect (pi-hole answers with `0.0.0.0` to indicate a blocked answer):
```
$ tailscale up --reset --accept-dns=true
$ dig ads.com +short
0.0.0.0
```

Don't know about any.

### OS

Linux

### OS version

Debian GNU/Linux testing/unstable

### Tailscale version

1.30.2

### Bug report

BUG-61ed798e7e05dd0ab5418968cc9693edc8b54953071b2f018aac499dabadd086-20221008113419Z-2be646c65961d209

Zocker1999NET:
It lets the user configure if DNS requests should be handled by the given exit node or should be handled as without an exit node (still routing the traffic through it, but not leveraging the tailscale daemon of the exit node itself; at least thats the plan).

This fixes #5875.

Known bugs:
- [ ] when upgrading from a running tailscale instance, the default seems to be false, not true, which would keep the same behavior on upgrading.

What is still to do:
- [ ] test without exit node (PR should change nothing)
- [x] test with exit node & with "Override local DNS" & with global nameserver set
- [ ] test with exit node & without "Override local DNS" & with global nameserver set
- [ ] test with exit node & without "Override local DNS" & without any global nameserver set
- [ ] test proper on other platforms than Linux (same tests than showed above)
  - [ ] Windows (do not have myself)
  - [ ] Mac OS X (do not have myself)
  - [ ] Android (UI missing; have myself)
  - [ ] iOS (UI missing; do not have myself)
- [ ] add automated tests for the new flag

This change only affects one device by itself, so only one needs to be updated (not the exit node). You should be able to revert to your old version after testing.

I test it the changes by building them, running the new tailscaled instead of the old one. I flip the new flag multiple times, verify that the log uses if enabled the DoH resolver of the exit node, or if disabled the configured global nameserver / the locally announced dns resolver. I verify that still tailscale is the main name resolver by checking for MagicDNS still working & verify that the correct DNS resolver is choosen by the same way as explained in #5875.

Signed-off-by: Felix Stupp <felix.stupp@banananet.work>

### DIFF
--- a/cmd/tailscale/cli/up.go
+++ b/cmd/tailscale/cli/up.go
@@ -98,6 +98,7 @@ func newUpFlagSet(goos string, upArgs *upArgsT) *flag.FlagSet {
 	upf.BoolVar(&upArgs.singleRoutes, "host-routes", true, "install host routes to other Tailscale nodes")
 	upf.StringVar(&upArgs.exitNodeIP, "exit-node", "", "Tailscale exit node (IP or base name) for internet traffic, or empty string to not use an exit node")
 	upf.BoolVar(&upArgs.exitNodeAllowLANAccess, "exit-node-allow-lan-access", false, "Allow direct access to the local network when routing traffic via an exit node")
+	upf.BoolVar(&upArgs.exitNodeUseDNS, "exit-node-use-dns", true, "let DNS requests be handled by the exit node when routing traffic via an exit node")
 	upf.BoolVar(&upArgs.shieldsUp, "shields-up", false, "don't allow incoming connections")
 	upf.BoolVar(&upArgs.runSSH, "ssh", false, "run an SSH server, permitting access per tailnet admin's declared policy")
 	upf.StringVar(&upArgs.advertiseTags, "advertise-tags", "", "comma-separated ACL tags to request; each must start with \"tag:\" (e.g. \"tag:eng,tag:montreal,tag:ssh\")")
@@ -136,6 +137,7 @@ type upArgsT struct {
 	singleRoutes           bool
 	exitNodeIP             string
 	exitNodeAllowLANAccess bool
+	exitNodeUseDNS         bool
 	shieldsUp              bool
 	runSSH                 bool
 	forceReauth            bool
@@ -286,7 +288,10 @@ func prefsFromUpArgs(upArgs upArgsT, warnf logger.Logf, st *ipnstate.Status, goo
 	}
 
 	if upArgs.exitNodeIP == "" && upArgs.exitNodeAllowLANAccess {
-		return nil, fmt.Errorf("--exit-node-allow-lan-access can only be used with --exit-node")
+		return nil, fmt.Errorf("allow-lan-access can only be used with --exit-node")
+	}
+	if upArgs.exitNodeIP == "" && !upArgs.exitNodeUseDNS { // "!" because default is true
+		return nil, fmt.Errorf("use-dns can only be used with --exit-node")
 	}
 
 	var tags []string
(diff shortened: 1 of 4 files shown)


DentonGentry:
I wish you would let the discussion on https://github.com/tailscale/tailscale/issues/5875 complete before sending a PR. This proposes a larger change than might be obvious. We allow a DNS configuration in https://login.tailscale.com/admin/dns with "Override local DNS" turned off, meaning there are no configured global DNS servers. The tailscale client is supposed to read out the resolvers from the underlying platform.

On some platforms once we enable an exit node, we can no longer access the platform's configured DNS servers from inside the sandbox in which tailscaled runs. We cannot implement the behavior of having "Override local DNS" turned off. Forwarding all queries to the exit node both:
- meets the goal of teleporting to the exit node's location
- makes it possible to function in these platforms.

Zocker1999NET:
> I wish you would let the discussion on https://github.com/tailscale/tailscale/issues/5875 complete before sending a PR.
Sorry, this was a synchronization issue. I began implementing & finishing this first draft before I saw your answer. I answered in the issue right after uploading this PR.

Zocker1999NET:
Because of [this implementation](https://github.com/tailscale/tailscale/issues/5875#issuecomment-1272348954), I decide to withdraw this PR, as this flag sadly will not solve the issue for macOS & iOS and may introduce issues to more common cases where "teleporting DNS requests" more properly is expected and maybe also required.

PR #5878, opened 2022-10-09
feat(darwin): add launchd log file

mritd:
add launchd log file

Signed-off-by: kovacs <mritd@linux.com>

### DIFF
--- a/cmd/tailscaled/install_darwin.go
+++ b/cmd/tailscaled/install_darwin.go
@@ -43,6 +43,12 @@ const darwinLaunchdPlist = `
   <key>RunAtLoad</key>
   <true/>
 
+  <key>StandardOutPath</key>
+  <string>/var/log/tailscaled.log</string>
+
+  <key>StandardErrorPath</key>
+  <string>/var/log/tailscaled.log</string>
+
 </dict>
 </plist>
 `


cmd/tailscaled/install_darwin.go
```diff
@@ -66,7 +72,7 @@ func uninstallSystemDaemonDarwin(args []string) (ret error) {
 			fmt.Printf("launchctl stop com.tailscale.tailscaled: %v, %s\n", err, out)
 			ret = err
 		}
-		out, err = exec.Command("launchctl", "unload", sysPlist).CombinedOutput()
+		out, err = exec.Command("launchctl", "unload", "-w", sysPlist).CombinedOutput()
```
DentonGentry, line 75: Do we know when launctl added the `-w` flag? macOS releases back to 10.13 are supported.

cmd/tailscaled/install_darwin.go
```diff
@@ -43,6 +43,12 @@ const darwinLaunchdPlist = `
   <key>RunAtLoad</key>
   <true/>
 
+  <key>StandardOutPath</key>
+  <string>/var/log/tailscaled.log</string>
+
+  <key>StandardErrorPath</key>
+  <string>/var/log/tailscaled.log</string>
+
```
bradfitz, line 51: Do these get rotated?

Is it okay/defined for stdout+stderr to have the same value?

cmd/tailscaled/install_darwin.go
mritd: I haven't found a changelog for launctl, so I'm not sure if the `-w` option is available in 10.13, unfortunately I don't currently have a 10.13 system.

So do I need to remove this option? Or do some other work?

cmd/tailscaled/install_darwin.go
mritd: Currently the log size is unlimited, a feasible solution is to increase the newsyslog configuration, such as `/etc/newsyslog.d/wifi.conf`.

If the log is not configured, it seems that the log will be discarded and cannot be viewed.

cmd/tailscaled/install_darwin.go
andrew-d: This appears to exist at least as of Sept. 2016, per the source code for `launchctl`:
https://github.com/apple-opensource/launchd/blob/master/support/launchctl.c#L2649-L2651

That source release is `launchd 842.92.1`, which was in macOS 10.9.5, before 10.13.

PR #5883, opened 2022-10-09
ssh/tailssh: add support for sending multiple banners

maisem:

### DIFF
--- a/ssh/tailssh/tailssh_test.go
+++ b/ssh/tailssh/tailssh_test.go
@@ -335,10 +335,10 @@ func TestSSHAuthFlow(t *testing.T) {
 	})
 
 	tests := []struct {
-		name       string
-		state      *localState
-		wantBanner string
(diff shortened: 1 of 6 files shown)


ssh/tailssh/tailssh_test.go
```diff
@@ -392,21 +393,24 @@ func TestSSHAuthFlow(t *testing.T) {
 					"check2": acceptRule.Action,
 				},
 			},
-			wantBanner: "url-here",
+			// TODO(maisem): make this test pass.
```
bradfitz, line 396: Does it pass? Delete Todo before merging.

bradfitz (approved):

ssh/tailssh/tailssh_test.go
maisem: Hah, I added this todo when the test wasn't working. It works now.

PR #5884, opened 2022-10-09
ssh/tailssh: close sshContext on context cancellation

maisem:
This was preventing tailscaled from shutting down properly if there were active sessions in certain states (e.g. waiting in check mode).

### DIFF
--- a/ssh/tailssh/context.go
+++ b/ssh/tailssh/context.go
@@ -5,6 +5,7 @@
 package tailssh
 
 import (
+	"context"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


ssh/tailssh/context.go
```diff
@@ -13,14 +14,16 @@ import (
 // that adds a CloseWithError method. Otherwise it's just a normalish
 // Context.
 type sshContext struct {
-	mu     sync.Mutex
-	closed bool
-	done   chan struct{}
-	err    error
+	underlying context.Context
+	cancel     context.CancelFunc
```
bradfitz, line 18: Docs. What's this close? Underlying?

bradfitz (approved):

ssh/tailssh/context.go
maisem: done.

PR #5886, opened 2022-10-10
cmd/ssh-auth-none-demo: add demo SSH server that acts like Tailscale SSH

bradfitz:
For SSH client authors to fix their clients without setting up Tailscale stuff.

### DIFF
--- a/cmd/ssh-auth-none-demo/ssh-auth-none-demo.go
+++ b/cmd/ssh-auth-none-demo/ssh-auth-none-demo.go
@@ -1,0 +1,171 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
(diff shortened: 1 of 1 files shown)


cmd/ssh-auth-none-demo/ssh-auth-none-demo.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	if len(keys) == 0 {
+		log.Fatal("no host keys")
+	}
+
+	srv := &ssh.Server{
+		Addr:    *addr,
+		Version: "Tailscale",
```
maisem, line 62: Can you add a none auth callback and have it write a banner too?

maisem (approved):

cmd/ssh-auth-none-demo/ssh-auth-none-demo.go
maisem: Discussed in chat, will do as a follow up

PR #5889, opened 2022-10-10
tsweb: remove allocs introduced by earlier change

bradfitz:
This removes the ~9 allocs added by #5869, while still keeping struct fields sorted (the previous commit's tests still pass). And add a test to lock it in that this shouldn't allocate.

Updates #5778


DentonGentry (approved):

mhlakhani:
thanks @bradfitz for fixing up my stuff - I learnt a bunch of new Go-isms this way :)

PR #5893, opened 2022-10-11
tsweb: sort top-level expvars after removing type prefixes

### LINKED ISSUE #5778: tsweb: varz aren't sorted

bradfitz:
Fixes #5778

### DIFF
--- a/tsweb/tsweb.go
+++ b/tsweb/tsweb.go
@@ -591,6 +591,18 @@ func writePromExpVar(w io.Writer, prefix string, kv expvar.KeyValue) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


bradfitz:
cc @mhlakhani

DentonGentry (approved):

mhlakhani:
thanks for tagging me here - this answers the questions I had, and taught me even more stuff :)

PR #5911, opened 2022-10-12
wgengine/magicsock: remove endpoint.wgEndpoint

dsnet:
This field seems seldom used and the documentation is wrong. It is simpler to just derive its original value dynamically when endpoint.DstToString is called.

This method is potentially used by wireguard-go,
but not in any code path is performance sensitive. All calls to it use it in conjunction with fmt.Printf, which is going to be slow anyways since it uses Go reflection.

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -2403,7 +2403,6 @@ func (c *Conn) SetNetworkMap(nm *netmap.NetworkMap) {
 			ep.discoKey = n.DiscoKey
 			ep.discoShort = n.DiscoKey.ShortString()
 		}
-		ep.wgEndpoint = n.Key.UntypedHexString()
 		ep.initFakeUDPAddr()
 		if debugDisco() { // rather than making a new knob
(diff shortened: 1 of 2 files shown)


bradfitz (commented):
I'd rather fix the docs than worry about the extra allocs (when?)

dsnet:
I'm not sure the allocation really matters. This method is called mostly for verbose debugging purposes: https://grep.app/search?q=DstToString&filter[repo][0]=WireGuard/wireguard-go

danderson (approved):
Approving on the basis that the code change is correct. I have no opinion about the performance question bradfitz raised, if it's really only used for debugging that seems fine, I guess?

bradfitz:
Look at the code you linked... it still calls that method unconditionally, even if the `log.Verbosef` is a no-op:

```go
if peer := entry.peer; peer.isRunning.Load() {
    device.log.Verbosef("Receiving cookie response from %s", elem.endpoint.DstToString())
    if !peer.cookieGenerator.ConsumeReply(&reply) {
```

It's also sent when generating the UAPI responses, and we still poll the UAPI stuff regularly?

Seems like this'll make regular background garbage.

dsnet:
Bleh. Alright. I'll send a revert that just changes the comment.

If the whole purpose of this was primarily for debugging, the `Endpoint` interface should have just had a `String` method that would lazily get called by `log.Verbosef`.

dsnet:
#5971

PR #5915, opened 2022-10-12
wgengine/router: fix MTU configuration on Windows

### LINKED ISSUE #5914: MTU is not set for IPv6 on Windows
### What is the issue?

I observed some connections failing periodically from my browser. Wireshark reveals that the MSS being sent in the first packet is 1360, and the webserver is dutifully replying with a large packet that never arrives through my exit node.

After some digging, I discovered that we are not setting the MTU on the Tailscale interface for IPv6.

```
PS C:\WINDOWS\system32>  netsh interface ipv4 show subinterface

       MTU  MediaSenseState      Bytes In     Bytes Out  Interface
----------  ---------------  ------------  ------------  -------------
      1280                1    1011077604      31499090  Tailscale


PS C:\WINDOWS\system32>  netsh interface ipv6 show subinterface

       MTU  MediaSenseState      Bytes In     Bytes Out  Interface
----------  ---------------  ------------  ------------  -------------
      1420                1      22824400       1434485  Tailscale

```

There is some code to do this in `ifconfig_windows.go` that seems to have lost function during evolution somewhere.

### Steps to reproduce

Connect to a service over Tailscale using IPv6 that respects and uses the Windows MSS advertisement, observe traffic loss.

Unknown, looks like the regression was a long time back.

### OS

Windows

### OS version

Windows 11

raggi:
Always set the MTU to the Tailscale default MTU. In practice we are
missing applying an MTU for IPv6 on Windows prior to this patch.

This is the simplest patch to fix the problem, the code in here needs
some more refactoring.

Fixes #5914

### DIFF
--- a/wgengine/router/ifconfig_windows.go
+++ b/wgengine/router/ifconfig_windows.go
@@ -24,6 +24,7 @@ import (
 	"tailscale.com/health"
 	"tailscale.com/net/interfaces"
 	"tailscale.com/net/tsaddr"
+	"tailscale.com/net/tstun"
 	"tailscale.com/util/multierr"
 	"tailscale.com/wgengine/winnet"
 )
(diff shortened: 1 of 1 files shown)


danderson (approved):

jinyi:
Does it break the behave of TS_DEBUG_MTU on Windows?

raggi:
> Does it break the behave of TS_DEBUG_MTU on Windows?

This change didn't introduce a regression on that axis, but it also did not fix that issue. Prior to this change MTU was never set for IPv6, with this change it is, but it does not respect TS_DEBUG_MTU, as several code paths do not.

Are you using TS_DEBUG_MTU regularly?

PR #5927, opened 2022-10-13
logtail: change batched upload mechanism to not use CPU when idle

bradfitz:
The mobile implementation had a 2 minute ticker going all the time to do a channel send. Instead, schedule it as needed based on activity.

Then we can be actually idle for long periods of time.

Updates #3363

### DIFF
--- a/logtail/logtail.go
+++ b/logtail/logtail.go
@@ -13,6 +13,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"os"
 	"strconv"
@@ -21,6 +22,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"tailscale.com/envknob"
 	"tailscale.com/logtail/backoff"
 	"tailscale.com/net/interfaces"
 	tslogger "tailscale.com/types/logger"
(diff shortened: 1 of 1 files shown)


logtail/logtail.go
```diff
@@ -247,18 +249,10 @@ func (l *Logger) Close() {
 // If the caller provides a DrainLogs channel, then unblock-drain-on-Write
```
nickoneill, line 249: `drainWake` now

nickoneill (approved):

maisem (approved):

dsnet:
Wow... so a 2 minute ticker makes a big difference in terms of battery usage?

bradfitz:
> Wow... so a 2 minute ticker makes a big difference in terms of battery usage?

It at least shows up I guess? cc @nickoneill @mihaip

dsnet:
The data-plane audit logging feature uses a 5-second ticker 😨.

I'll refactor it in the future to be activity based as well.

nickoneill:
I wouldn't call it a significant impact, but it did seem to make a visual difference in the usage chart when I set the timer to 6h rather than 2m 🙃

logtail/logtail.go
```diff
@@ -241,24 +253,16 @@ func (l *Logger) Close() {
...
 // drainBlock is called by drainPending when there are no logs to drain.
 //
-// In typical operation, every call to the Write method unblocks and triggers
-// a buffer.TryReadline, so logs are written with very low latency.
+// In typical operation, every call to the Write method unblocks and triggers a
+// buffer.TryReadline, so logs are written with very low latency.
 //
-// If the caller provides a DrainLogs channel, then unblock-drain-on-Write
-// is disabled, and it is up to the caller to trigger unblock the drain.
+// If the caller specified FlushInterface, drainWake is only sent to
```
mihaip, line 259: Should this have been `FlushDelay`?

PR #5930, opened 2022-10-13
cmd/netlogfmt: new package to pretty print network traffic logs

dsnet:
This package parses a JSON stream of netlog.Message from os.Stdin and pretty prints the contents as a stream of tables.

It supports reverse lookup of tailscale IP addresses if given an API key and the tailnet that these traffic logs belong to.


dsnet:
Happy to show a demo of this in action if you'd like. It's pretty cool.

dblohm7 (approved):

PR #5940, opened 2022-10-14
net/interfaces: improve default route detection 

### LINKED ISSUE #5879: High CPU and network bandwidth usage at special gateways
### Operating environment

- run using the command line
- tailscale(d) version: v1.30.2
- golang build version: go1.19.2
- System version: macOS 12.6 (21G115) / M1 Max(arm64)

### What happened

**The tailscaled process consumes a lot of CPU and sends a lot of traffic to the derper server.**

image

image

### Allow me to start with my findings:

When used with some other network proxy tools (such as [Surge](https://nssurge.com/)), if other tools set a proxy gateway (such as Surge enhanced mode will set the default gateway to 198.18.0.1), tailsaled seems to be unable to read the default gateway, which will cause tailscaled Stuck in some infinite loop.

### How to reproduce

- Download [Surge](https://nssurge.com/) App
- Turn on enhanced mode in the top menu
- Use the `netstat -rn` command to confirm that the default gateway is set to `198.18.0.1`
- Restart tailscaled several times with `tailscale down && tailscale up` command
- The tailscaled process will continue to consume a lot of CPU and keep sending a lot of traffic to the derper

### Log

To describe the problem more clearly, I modified the launchd configuration and generated the log(I submitted a PR #5878 incidentally to enable log files by default):

[tailscaled.log](https://github.com/tailscale/tailscale/files/9740696/tailscaled.log)

I tried looking at the code and it seems that there is a problem in the `interfaces_darwin.go` code, guessing that the impactful change might start with `04c2c5bd80e86b2113d4e04e840d79ec66981fda`.

image

### OS

macOS

### OS version

12.6 (21G115) / M1 Max(arm64)

### Tailscale version

v1.30.2

### Bug report

BUG-9f5e02a01c6c13045de0c809a1920718b3a3e585e740cb2555da520c13491e40-20221009030245Z-b4a892d63e1da8c7

knyar:
Instead of treating any interface with a non-ifscope route as a potential default gateway, now verify that a given route is actually a default route (0.0.0.0/0 or ::/0).

I would suggest reviewing each commit separately.

Fixes #5879


net/interfaces/interfaces_bsd.go
```diff
@@ -2,11 +2,10 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// This might work on other BSDs, but only tested on FreeBSD.
-// Originally a fork of interfaces_darwin.go with slightly different flags.
+// Common code for FreeBSD and Darwin.
```
bradfitz, line 5: Leave a comment that it might work on other BSDs too but untested? Notably OpenBSD, which we run on.

net/interfaces/interfaces_bsd.go
```diff
@@ -103,25 +91,52 @@ func likelyHomeRouterIPBSDFetchRIB() (ret netip.Addr, ok bool) {
...
+	const RTF_IFSCOPE = 0x1000000
+	if rm.Flags&RTF_IFSCOPE != 0 {
+		return false
+	}
+	if len(rm.Addrs) <= unix.RTAX_NETMASK {
+		return false
+	}
+
+	dst := rm.Addrs[unix.RTAX_DST]
+	netmask := rm.Addrs[unix.RTAX_NETMASK]
```
bradfitz, line 125: I assume RTAX_DST is less than RTAX_NETMASK? :)

net/interfaces/interfaces_darwin_test.go
```diff
@@ -67,19 +81,23 @@ func likelyHomeRouterIPDarwinExec() (ret netip.Addr, ok bool) {
 		if len(f) < 3 || !f[0].EqualString("default") {
```
bradfitz, line 81: < 4 now?

bradfitz (approved):

net/interfaces/interfaces_bsd.go
knyar: Yep, added a comment to clarify this.

PR #5601, opened 2022-09-12
cmd/derper: add /generate_204 endpoint

andrew-d:
For captive portal detection.

Signed-off-by: Andrew Dunham <andrew@tailscale.com>

PR #5942, opened 2022-10-14
cmd/derper, net/netcheck: add challenge/response to generate_204 endpoint

andrew-d:
The Lufthansa in-flight wifi generates a synthetic 204 response to the DERP server's `/generate_204` endpoint. This PR adds a basic challenge/response to the endpoint; something sufficiently complicated that it's unlikely to be implemented by a captive portal. We can then check for the expected response to verify whether we're being MITM'd.

Follow-up to #5601

Change-Id: I94a68c9a16a7be7290200eea6a549b64f02ff48f

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -325,9 +325,29 @@ func main() {
 	}
 }
 
+const (
+	noContentChallengeHeader = "X-Tailscale-Challenge"
+	noContentResponseHeader  = "X-Tailscale-Response"
+)
+
 // For captive portal detection
 func serveNoContent(w http.ResponseWriter, r *http.Request) {
+	if challenge := r.Header.Get(noContentChallengeHeader); challenge != "" {
+		badChar := strings.IndexFunc(challenge, func(r rune) bool {
+			return !isChallengeChar(r)
+		}) != -1
+		if len(challenge) <= 64 && !badChar {
+			w.Header().Set(noContentResponseHeader, "response "+challenge)
+		}
+	}
 	w.WriteHeader(http.StatusNoContent)
+}
+
+func isChallengeChar(c rune) bool {
+	// Semi-randomly chosen as a limited set of valid characters
+	return ('a' <= c && c <= 'z') || ('A' <= c && c <= 'Z') ||
+		('0' <= c && c <= '9') ||
+		c == '.' || c == '-' || c == '_'
 }
 
 // probeHandler is the endpoint that js/wasm clients hit to measure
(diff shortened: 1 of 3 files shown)


bradfitz (commented):
Rather than a new endpoint, how about keep using 204 but make it fancier? You can add a request header & response header to see if it's real.

Like:

GET /generate_204 HTTP/1.1
X-Tailscale-Challenge: foobar

HTTP/1.1 204 No Content
X-Tailscale-Response: raboof

Doesn't need to be a complicated challenge/response. Just something Tailscale-specific captive portals probably won't go out of their way to fake.

andrew-d:
@bradfitz - Yeah, I like that; just pushed something with a very basic challenge. I wanted something non-constant, so `"response " + challenge` seemed simple enough.

Thoughts? Either way I'll hold off on merging until I can deploy all the DERP servers, so any unstable builds don't falsely detect captive portals when the server doesn't do this.

cmd/derper/derper.go
```diff
@@ -327,6 +327,9 @@ func main() {
 
 // For captive portal detection
 func serveNoContent(w http.ResponseWriter, r *http.Request) {
+	if challenge := r.Header.Get("X-Tailscale-Challenge"); challenge != "" {
```
bradfitz, line 330: validate that it is only ascii non-whitespace? or only hex maybe? and not too long?

bradfitz (approved):
But LGTM overall.

cmd/derper/derper.go
andrew-d: Sure, easy enough - no reason to allow tons of data or strange characters here. Added a test as well, because I don't add constraints without testing them 😃

bradfitz (approved):

net/netcheck/netcheck.go
```diff
@@ -1116,13 +1116,20 @@ func (c *Client) checkCaptivePortal(ctx context.Context, dm *tailcfg.DERPMap, pr
 	if err != nil {
 		return false, err
 	}
+
+	chal := "tailscale " + node.HostName
```
mnordhoff, line 1120: Assuming this is the node's hostname or something, and assuming it's not exposed in other ways, isn't this a minor privacy issue?

andrew-d: Good question! In this case, `node` is the node of the DERP server, not the local machine that Tailscale is running on. So `node.HostName` would be something like `derp10.tailscale.com` here.

PR #5943, opened 2022-10-14
net/dns: print systemd-resolved ResolvConfMode

andrew-d:
The ResolvConfMode property is documented to return how systemd-resolved is currently managing /etc/resolv.conf. Include that information in the debug line, when available, to assist in debugging DNS issues.

Change-Id: I1ae3a257df1d318d0193a8c7f135c458ec45093e

### DIFF
--- a/net/dns/manager_linux.go
+++ b/net/dns/manager_linux.go
@@ -31,6 +31,7 @@ func NewOSConfigurator(logf logger.Logf, interfaceName string) (ret OSConfigurat
 	env := newOSConfigEnv{
 		fs:                directFS{},
 		dbusPing:          dbusPing,
+		dbusReadString:    dbusReadString,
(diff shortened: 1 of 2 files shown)


andrew-d:
(note: I still need to do some real-world testing; I'm currently on a plane and don't want to reconfigure DNS on my machine at home 😅)

bradfitz (approved):

danderson (approved):

andrew-d:
(the most recent changes are to call `org.freedesktop.DBus.Properties.Get` manually instead of using `GetProperty` so that we can apply a timeout)

andrew-d:
Confirmed this works locally:
```
dns: [resolved-ping=yes rc=resolved nm=no resolv-conf-mode=stub ret=systemd-resolved]
dns: using *dns.resolvedManager
```

PR #5946, opened 2022-10-14
cmd/tsconnect: pre-compress main.wasm when building the NPM package

mihaip:
This way we can do that once (out of band, in the GitHub action), instead of increasing the time of each deploy that uses the package.

.wasm is removed from the list of automatically pre-compressed extensions, an OSS bump and small change on the corp side is needed to make use of this change.

### DIFF
--- a/cmd/tsconnect/build-pkg.go
+++ b/cmd/tsconnect/build-pkg.go
@@ -12,6 +12,7 @@ import (
 	"path"
 
 	"github.com/tailscale/hujson"
+	"tailscale.com/util/precompress"
 	"tailscale.com/version"
 )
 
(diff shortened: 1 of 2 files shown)


bradfitz:
As long as we're not checking in the wasm files (compressed or not) to our repos I'm cool with this :)

Thanks!

bradfitz (approved):

mihaip:
> As long as we're not checking in the wasm files (compressed or not) to our repos I'm cool with this :)

Yeah, that doesn't change (we weren't before). The NPM package will go from ~22MB to 35MB, but we only download that on-demand.

PR #5949, opened 2022-10-14
derp: add TCP RTT metric on Linux

andrew-d:
Periodically poll the TCP RTT metric from all open TCP connections and update a (bucketed) histogram metric.

Change-Id: I6214902196b05bf7829c9d0ea501ce0e13d984cf

### DIFF
--- a/derp/derp_server_default.go
+++ b/derp/derp_server_default.go
@@ -1,0 +1,14 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build !linux
+// +build !linux
+
+package derp
+
+import "context"
+
+func (c *sclient) statsLoop(ctx context.Context) error {
+	return nil
+}
--- a/derp/derp_server_linux.go
+++ b/derp/derp_server_linux.go
@@ -1,0 +1,95 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package derp
+
+import (
+	"context"
+	"crypto/tls"
+	"net"
+	"time"
+
+	"golang.org/x/sys/unix"
+)
+
+func (c *sclient) statsLoop(ctx context.Context) error {
+	// If we can't get a TCP socket, then we can't send stats.
+	tcpConn := c.tcpConn()
+	if tcpConn == nil {
+		c.s.tcpRtt.Add("non-tcp", 1)
+		return nil
+	}
+	rawConn, err := tcpConn.SyscallConn()
+	if err != nil {
+		c.logf("error getting SyscallConn: %v", err)
+		c.s.tcpRtt.Add("error", 1)
+		return nil
+	}
+
+	const statsInterval = 10 * time.Second
+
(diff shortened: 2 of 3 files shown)


andrew-d:
(unsure if we want to do this, feedback appreciated!)

derp/derp_server_default.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
```
DentonGentry, line 1: 2022

derp/derp_server_linux.go
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2020 Tailscale Inc & AUTHORS All rights reserved.
```
DentonGentry, line 1: 2022

derp/derp_server_linux.go
```diff
@@ -0,0 +1,77 @@
...
+
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
+}
+
+func durationToLabel(dur time.Duration) string {
+	switch {
+	case dur < 10*time.Millisecond:
```
DentonGentry, line 60: The Labelmap says "le", which I assume is "less-than-or-equal", but the code is less-than.

DentonGentry (approved):
SGTM

derp/derp_server_linux.go
andrew-d: Good catch; thanks. Fixed!

derp/derp_server_linux.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"golang.org/x/sys/unix"
+)
+
+func (c *sclient) statsLoop(ctx context.Context) error {
+	// If we can't get a TCP socket, then we can't send stats.
+	tcpConn, ok := c.nc.(*net.TCPConn)
+	if !ok {
+		// TODO(andrew): metric?
+		return nil
```
maisem, line 20: Log the error and/or add metric to the label map?

derp/derp_server_linux.go
```diff
@@ -0,0 +1,77 @@
...
+			err = rawConn.Control(func(fd uintptr) {
+				tcpInfo, sysErr = unix.GetsockoptTCPInfo(int(fd), unix.IPPROTO_TCP, unix.TCP_INFO)
+			})
+			if err != nil || sysErr != nil {
+				continue statsLoop
+			}
+
+			// TODO(andrew): more metrics?
+			rtt := time.Duration(tcpInfo.Rtt) * time.Microsecond
+			c.s.tcpRtt.Get(durationToLabel(rtt)).Add(1)
```
maisem, line 50: You can use Add directly instead of Get.Add

maisem (approved):

maisem:
Do you have reasons why we shouldn't do this?

derp/derp_server_linux.go
andrew-d: 👍

andrew-d:
Just tested this locally with a DERP server and `derpprobe`:
```
$ curl -kfsSL --resolve derp1b.tailscale.com:3340:127.0.0.1 https://derp1b.tailscale.com:3340/debug/varz | grep derp_tcp_rtt
# TYPE derp_tcp_rtt counter
derp_tcp_rtt{le="10ms"} 15
```

> Do you have reasons why we shouldn't do this?

Not particularly, other than "just slightly more complexity", which IMO is fine.

PR #5951, opened 2022-10-15
logtail: default to 2s log flush delay on all platforms

bradfitz:
Per chat. This is close enough to realtime but massively reduces number of HTTP requests. (which you can verify with TS_DEBUG_LOGTAIL_WAKES and watching tailscaled run at start)

By contrast, this is set to 2 minutes on mobile.

### DIFF
--- a/logtail/logtail.go
+++ b/logtail/logtail.go
@@ -33,6 +33,8 @@ import (
 // Config.BaseURL isn't provided.
 const DefaultHost = "log.tailscale.io"
 
+const defaultFlushDelay = 2 * time.Second
+
 const (
 	// CollectionNode is the name of a logtail Config.Collection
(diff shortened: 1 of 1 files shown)


DentonGentry (approved):

crawshaw (approved):
If we have any tests that look at log output this might slow them down. Disable the delay in tests?

bradfitz:
> If we have any tests that look at log output this might slow them down. Disable the delay in tests?

We can't just check whether we're in a test because any integration test using this will be running as a child process and our usual trick if looking up the `"test.v"` flag won't work.

We'll need to use explicit env. Elsewhere we use `IN_TS_TEST` so I can use that. But also: we don't set this default if `TS_DEBUG_LOGTAIL_FLUSHDELAY` is set at all so the integration tests.

PR #5952, opened 2022-10-15
words: space, the final frontier

catzkorn:
Captains log. Stardate 100386.37.

Work is proceeding on the Words list as Tailscalars are forced to scavenge for more taily and scaley things.


words/scales.txt
```diff
@@ -118,8 +118,9 @@ hen
 rooster
 quail
 grouse
-
-# Musical scales
```
bradfitz, line 122: Why deleting the comment/grouping?

catzkorn: We don't use grouping anywhere else or order things in general. So between ordering them and grouping for consistency, and just removing the one grouping label, I chose to do the latter.

Plus, the mystic of why they are there is more fun 😛

sailorfrag (approved):

PR #5966, opened 2022-10-17
tsnet/examples/tshello: update example for LocalClient method

Xe:
Before this would silently fail if this program was running on a machine that was not already running Tailscale. This patch changes the WhoIs call to use the tsnet.Server LocalClient instead of the global tailscale LocalClient.

### DIFF
--- a/tsnet/example/tshello/tshello.go
+++ b/tsnet/example/tshello/tshello.go
@@ -32,13 +32,18 @@ func main() {
 	}
 	defer ln.Close()
 
+	lc, err := s.LocalClient()
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	if *addr == ":443" {
 		ln = tls.NewListener(ln, &tls.Config{
 			GetCertificate: tailscale.GetCertificate,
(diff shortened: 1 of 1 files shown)


DentonGentry (approved):

tsnet/example/tshello/tshello.go
```diff
@@ -32,13 +32,18 @@ func main() {
...
 	defer ln.Close()
 
+	lc, err := s.LocalClient()
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	if *addr == ":443" {
 		ln = tls.NewListener(ln, &tls.Config{
 			GetCertificate: tailscale.GetCertificate,
```
grahamc, line 42: I wonder if this should be going to the local client as well?

grahamc:
Thanks, Xe! This closely mirrors my own adaptation: https://github.com/DeterminateSystems/bonk/pull/1/commits/37bb4f6e3310e1aa5edc20e85f26e78f44b5e83f

PR #5969, opened 2022-10-17
tsnet,client/tailscale: add APIClient which runs API over Noise.

maisem:
Updates tailscale/corp#4383


tsnet/tsnet.go
```diff
@@ -481,6 +482,25 @@ func getTSNetDir(logf logger.Logf, confDir, prog string) (string, error) {
 	return newPath, nil
 }
 
+// APIClient returns a tailscale.Client that can be used to make authenticated
+// requests to the Tailscale control server.
+// It requires the user to set tailscale.I_Acknowledge_This_API_Is_Unstable.
+func (s *Server) APIClient(ctx context.Context) (*tailscale.Client, error) {
```
bradfitz, line 488: what do you plan to use this ctx for?

client/tailscale/tailscale.go
```diff
@@ -90,6 +93,26 @@ func (c *Client) setAuth(r *http.Request) {
 	}
 }
 
+type nodeKey key.NodePublic
+
+func (k nodeKey) modifyRequest(req *http.Request) {
```
bradfitz, line 98: type nodeKeyAuth? type noiseAuth?

at least put some comments here about why you're setting this basic auth header?

tsnet/tsnet.go
maisem: oops

bradfitz (approved):

PR #5974, opened 2022-10-17
cmd/tsconnect: add progress and connection callbacks

mihaip:
Allows UI to display slightly more fine-grained progress when the SSH connection is being established.

Updates tailscale/corp#7186


cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -385,6 +391,7 @@ func (s *jsSSHSession) Run() {
 		User:            s.username,
 	}
 
+	reportProgress("Starting SSH client…")
```
bradfitz, line 394: if you want to do more, on line 390 just above there's the HostKeyCallback you can do to show that you've gotten data from the SSH connection and it's doing the phase of the SSH protocol before auth, exchanging host keys.

bradfitz (approved):

cmd/tsconnect/wasm/wasm_js.go
mihaip: Good idea, done.

PR #5988, opened 2022-10-19
prober: add a DERP probe manager based on derpprobe

knyar:
This ensures that each DERP server is probed individually (TLS and STUN) and also manages per-region mesh probing. Actual probing code has been copied from cmd/derpprobe.

### DIFF
--- a/prober/derp.go
+++ b/prober/derp.go
@@ -1,0 +1,392 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package prober
+
+import (
+	"bytes"
+	"context"
+	crand "crypto/rand"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log"
+	"net"
+	"net/http"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"tailscale.com/derp"
+	"tailscale.com/derp/derphttp"
+	"tailscale.com/net/stun"
+	"tailscale.com/tailcfg"
+	"tailscale.com/types/key"
+	"tailscale.com/types/logger"
+)
+
+// derpProber dynamically manages several probes for each DERP server
+// based on the current DERPMap.
+type derpProber struct {
+	p            *Prober
+	derpMapURL   string
+	udpInterval  time.Duration
+	meshInterval time.Duration
+	tlsInterval  time.Duration
+
+	// Probe functions that can be overriden for testing.
+	tlsProbeFn  func(string) ProbeFunc
+	udpProbeFn  func(string, int) ProbeFunc
+	meshProbeFn func(string, string) ProbeFunc
+
+	sync.Mutex
+	lastDERPMap   *tailcfg.DERPMap
+	lastDERPMapAt time.Time
+	nodes         map[string]*tailcfg.DERPNode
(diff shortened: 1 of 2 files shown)


prober/derp.go
```diff
@@ -0,0 +1,379 @@
...
+	if port == 0 {
+		port = 3478
+	}
+	for {
+		ip := net.ParseIP(ipStr)
+		_, err := uc.WriteToUDP(req, &net.UDPAddr{IP: ip, Port: port})
+		if err != nil {
+			return 0, err
+		}
+		buf := make([]byte, 1500)
```
DentonGentry, line 219: I imagine this is 1500 to be the largest a packet can be, but 1500 isn't the largest a packet can be.

If a DERP is running in AWS it might use jumbo frames to get 9001 bytes.

Since these are UDP frames, I think they could be up to 64K and just get fragmented. I don't see anything which would prevent reassembled fragmented frames from being received.

All of which is to say that I imagine making the buffer as large as we expect a STUN frame to be is probably sufficient, and if we want to handle a maximum size frame then I'm not sure 1500 is enough.

prober/derp.go
```diff
@@ -0,0 +1,379 @@
...
+	if err != nil {
+		return 0, err
+	}
+	defer toc.Close()
+
+	// Wait a bit for from's node to hear about to existing on the
+	// other node in the region, in the case where the two nodes
+	// are different.
+	if from.Name != to.Name {
+		time.Sleep(100 * time.Millisecond) // pretty arbitrary
```
DentonGentry, line 265: I don't quite understand what this 100 msec is, but some of our distant DERPs will be more than 100 msec away from wherever this prober runs. If this is RTT I think this will skip some of the DERPs.

prober/derp_test.go
```diff
@@ -0,0 +1,102 @@
...
+	dm := &tailcfg.DERPMap{
+		Regions: map[int]*tailcfg.DERPRegion{
+			0: {
+				RegionID:   0,
+				RegionCode: "zero",
+				Nodes: []*tailcfg.DERPNode{
+					{
+						Name:     "n1",
+						RegionID: 0,
+						HostName: "derpn1.tailscale.fake",
```
DentonGentry, line 23: The TLD ".test" is reserved for tests and guaranteed to never appear as a real TLD on the Internet.

https://en.wikipedia.org/wiki/.test

DentonGentry (approved):

prober/derp.go
knyar: I believe this is to allow the information about a new node to propagate from the `to` DERP server to the `from` DERP server. Both DERP servers are in the same region, so 100ms seems sufficient?

prober/derp.go
knyar: I looked at the [original commit](https://github.com/tailscale/tailscale/commit/13ef8e3c06de619d25eea3611027d2b1143cc9cf) and it's unclear why 1500 was chosen, but according to [STUN RFC](https://www.rfc-editor.org/rfc/rfc5389#section-7.1) the UDP messages should not exceed path MTU, so I guess 1500 might have been a reasonable upper bound in a non-jumbo frame situation.

In practice we are sending a Binding request that is around 40 bytes in size, with the expected binding response being similarly small. Using a 64K buffer for the response seems a bit wasteful, but I have increased this to 9000 to accommodate jumbo frames and added a comment explaining rationale.

DentonGentry (approved):

Issue #4570, opened 2022-04-29
FR: Configurable auth key expiry


lincolnmantracer:
### What are you trying to do?

We use single-use auth keys to provision instances like the docs suggest, and we even do it by having [vault create the keys for us](https://github.com/davidsbond/vault-plugin-tailscale). It's pretty nifty. However, there's no way to specify the expiry when creating auth keys with the api, so the keys always have the default 90 day expiry. We don't really need that, since the machines start up immediately. An hour or even ten minutes would be sufficient for our case.

### How should we solve this?

It would be great to have a `ttl` or `expiry` argument to [`POST /keys`](https://github.com/tailscale/tailscale/blob/main/api.md#post-apiv2tailnettailnetkeys---create-a-new-key-for-a-tailnet).

lincolnmantracer:
While I'm asking for expansions to the tailnet key API, it would be really nice if we could specify a key description, like "key for the GCP subnet relay"

DentonGentry:
> it would be really nice if we could specify a key description
This is tracked in https://github.com/tailscale/tailscale/issues/4314

PR #5989, opened 2022-10-19
api.md: add expirySeconds as parameter to post Tailnet keys

maisem:
Updates #4570


willnorris (approved):

mayakacz:
Can you also add to API docs

Issue #540, opened 2020-07-09
Reduce heartbeat ping rate


bradfitz:
Currently, active peers heartbeat with each other every 2 seconds.

(@danderson and I had discussed this and considered several options; the overall constraint is we want to only ever be without a network connection for 5 seconds max, before failing back to DERP, so we need a heartbeat at least half that, ideally.)

Unfortunately, every 2 seconds is in both directions, and each ping also involves a pong.

So every 2 seconds:

```
srv: 8.9M/0.0M magicsock: disco: d:de3ee729b55e7f2f<-d:c9b9fc10e4b34938 ([MINwg], 167.71.156.251:41641)  got ping tx=0ada6c80d510
srv: 8.9M/0.0M magicsock: disco: d:de3ee729b55e7f2f->d:c9b9fc10e4b34938 ([MINwg], 167.71.156.251:41641) sent pong tx=0ada6c80d510
srv: 8.9M/0.0M magicsock: disco: d:de3ee729b55e7f2f->d:c9b9fc10e4b34938 ([MINwg], 167.71.156.251:41641) sent ping tx=764870a09f60
srv: 8.9M/0.0M magicsock: disco: d:de3ee729b55e7f2f<-d:c9b9fc10e4b34938 ([MINwg], 167.71.156.251:41641)  got pong tx=764870a09f60 latency=23ms pong.src=209.180.207.193:49249
```

Which works out to 2 packets per second.

I think we can get that down to 1/4 that.

We can have a bit in the Ping frame that says "I'm a heartbeat ping, no need to reply right away." (a ping frame used for health, not latency). Then the response Pong can both say 1) how long it was delay (so the pinger can still compute pretty accurate ping time, less the delay time), and 2), can carry its own Ping challenge for its direction along with the Pong response.

Then we'd be back down to just 1 packet every 2 seconds, with the pings from each direction 1 second out of phase.

Alternatively, we do something like elect the lexicographically lesser disco endpoint as the heartbeat pinger and only ping in one direction. The ponger then just trusts the path if it got pings from it. That optimization would only come into play when the path is symmetric, which I imagine it is almost always (I haven't analyzed logs yet, but it seems to be).

DentonGentry:
I think this is done in 1.4.

bradfitz:
This is a different optimization.

DentonGentry:
Related:
corp PR 2874 suppressed sending certain map poll updates to mobile clients
Tailscale 1.16 disabled TCP keepalives

[...]

PR #5994, opened 2022-10-19
wgengine/magicsock: set up pathfinder

phirework:
Sets up new file for separate silent disco goroutine, tentatively named pathfinder for now.

Updates #540

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -3352,7 +3352,11 @@ type endpoint struct {
 
 	pendingCLIPings []pendingCLIPing // any outstanding "tailscale ping" commands running
 
-	heartbeatDisabled bool // heartBeatTimer disabled for silent disco. See issue #540.
+	// The following fields are related to the new "silent disco"
+	// implementation that's a WIP as of 2022-10-20.
+	// See #540 for background.
+	heartbeatDisabled bool
+	pathFinderRunning bool
 }
 
 type pendingCLIPing struct {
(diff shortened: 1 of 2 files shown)


wgengine/magicsock/magicsock.go
```diff
@@ -3630,6 +3633,20 @@ func (de *endpoint) cliPing(res *ipnstate.PingResult, cb func(*ipnstate.PingResu
 }
 
 func (de *endpoint) send(b []byte) error {
+	// if heartbeat disabled, try using alternate path
+	if de.heartbeatDisabled {
```
bradfitz, line 3637: This is a data race, reading this field without a mutex.

wgengine/magicsock/magicsock.go
```diff
@@ -3630,6 +3633,20 @@ func (de *endpoint) cliPing(res *ipnstate.PingResult, cb func(*ipnstate.PingResu
...
+		if !de.pathFinderRunning {
+			de.startPathFinder()
+		}
+		if sendFunc, ok := de.atomicSenderFunc.LoadOk(); ok {
```
bradfitz, line 3641: I forgot that our syncs.AtomicValue does actually transition from nil to non-nil to nil again because it always stores typed nils in the interface. So this LoadOk isn't needed. We can use a nil func pointer as the sentinel value for "need to take the slow path"

wgengine/magicsock/magicsock.go
```diff
@@ -3630,6 +3633,20 @@ func (de *endpoint) cliPing(res *ipnstate.PingResult, cb func(*ipnstate.PingResu
...
+			return sendFunc(b)
+		}
+	}
+
+	return de.noisyDiscoSend(b)
```
bradfitz, line 3646: I'd rather not even be this invasive yet by adding a stack frame.

wgengine/magicsock/magicsock.go
```diff
@@ -3630,6 +3633,20 @@ func (de *endpoint) cliPing(res *ipnstate.PingResult, cb func(*ipnstate.PingResu
 }
 
 func (de *endpoint) send(b []byte) error {
+	// if heartbeat disabled, try using alternate path
+	if de.heartbeatDisabled {
+		if !de.pathFinderRunning {
```
bradfitz, line 3638: also no mutex

bradfitz (commented):
I sent https://github.com/tailscale/tailscale/pull/5997 for what I was thinking for just the send part, without the path finder stuff yet.

phirework:
@bradfitz rebased your changes from #5997 and left just the pathfinder skeleton and the `heartbeatDisabled` check.

bradfitz (approved):

PR #5998, opened 2022-10-20
wgengine/router: [linux] add before deleting interface addrs

maisem:
Deleting may temporarily result in no addrs on the interface, which results in all other rules (like routes) to get dropped by the OS.

I verified this fixes the problem.

### DIFF
--- a/wgengine/router/router_linux.go
+++ b/wgengine/router/router_linux.go
@@ -1532,25 +1532,10 @@ func cidrDiff(kind string, old map[netip.Prefix]bool, new []netip.Prefix, add, d
 		ret[cidr] = true
 	}
 
-	var delFail []error
-	for cidr := range old {
-		if newMap[cidr] {
-			continue
-		}
-		if err := del(cidr); err != nil {
-			logf("%s del failed: %v", kind, err)
-			delFail = append(delFail, err)
-		} else {
-			delete(ret, cidr)
-		}
-	}
-	if len(delFail) == 1 {
-		return ret, delFail[0]
-	}
-	if len(delFail) > 0 {
-		return ret, fmt.Errorf("%d delete %s failures; first was: %w", len(delFail), kind, delFail[0])
-	}
-
+	// We want to add before we delete, so that if there is no overlap, we don't
+	// end up in a state where we have no addresses on an interface as that
+	// results in other kernel entities (like routes) pointing to that interface
+	// to also be deleted.
 	var addFail []error
 	for cidr := range newMap {
 		if old[cidr] {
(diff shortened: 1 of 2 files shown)


wgengine/router/router_linux.go
```diff
@@ -1532,25 +1532,10 @@ func cidrDiff(kind string, old map[netip.Prefix]bool, new []netip.Prefix, add, d
...
-	if len(delFail) == 1 {
-		return ret, delFail[0]
-	}
-	if len(delFail) > 0 {
-		return ret, fmt.Errorf("%d delete %s failures; first was: %w", len(delFail), kind, delFail[0])
-	}
-
+	// We want to add before we delete, so that if there is no
+	// overlap, we don't end up in a state where we have no addresses
+	// on an interface as that results in other rules pointing to
```
bradfitz, line 1537: Rules or routes? Or "kernel entities like routes"

maisem: i like `kernel entities (like routes)`

wgengine/router/router_linux.go
```diff
@@ -1532,25 +1532,10 @@ func cidrDiff(kind string, old map[netip.Prefix]bool, new []netip.Prefix, add, d
...
-		return ret, delFail[0]
-	}
-	if len(delFail) > 0 {
-		return ret, fmt.Errorf("%d delete %s failures; first was: %w", len(delFail), kind, delFail[0])
-	}
-
+	// We want to add before we delete, so that if there is no overlap, we don't
+	// end up in a state where we have no addresses on an interface as that
+	// results in other kernel entities (like routes) pointing to that interface
+	// to also be deleted.
```
andrew-d, line 1538:
```suggestion
	// also being deleted.
```

bradfitz (commented):
This looks very testable! cidrDiff takes those funcs already.

wgengine/router/router_linux.go
```diff
@@ -1532,25 +1532,10 @@ func cidrDiff(kind string, old map[netip.Prefix]bool, new []netip.Prefix, add, d
...
-		}
-	}
-	if len(delFail) == 1 {
-		return ret, delFail[0]
-	}
-	if len(delFail) > 0 {
-		return ret, fmt.Errorf("%d delete %s failures; first was: %w", len(delFail), kind, delFail[0])
-	}
-
+	// We want to add before we delete, so that if there is no overlap, we don't
```
danderson, line 1535: For the record: I was worried that this might result in a race where the wrong source addr gets used briefly during a transition, but wireguard will just eat those packets because cryptokey routing won't match.

danderson (approved):

wgengine/router/router_linux.go
maisem: Yep

maisem:
> This looks very testable! cidrDiff takes those funcs already.

Added test, PTAL

wgengine/router/router_linux_test.go
```diff
@@ -839,3 +840,82 @@ Usage: busybox [function [arguments]...]
...
+
+		// Check that the final state is correct.
+		if len(fm) != len(tc.final) {
+			t.Fatalf("final state = %v, want %v", fm, tc.final)
+		}
+		for _, p := range tc.final {
+			if !fm[p] {
+				t.Errorf("final state = %v, want %v", fm, tc.final)
+			}
+		}
```
andrew-d, line 919: [`maps.Equal`](https://pkg.go.dev/golang.org/x/exp/maps#Equal)?

andrew-d (approved):
Thanks for the test

wgengine/router/router_linux_test.go
maisem: I have a map and a slice :(

andrew-d: 🤦 ignore that then 😅

PR #6003, opened 2022-10-20
prober: optionally spread probes over time

knyar:
By default all probes with the same probe interval that have been added together will run on a synchronized schedule, which results in spiky resource usage and potential throttling by third-party systems (for example, OCSP servers used by the TLS probes).

To address this, prober can now run in "spread" mode that will introduce a random delay before the first run of each probe.

### DIFF
--- a/prober/prober.go
+++ b/prober/prober.go
@@ -13,8 +13,10 @@ import (
 	"errors"
 	"expvar"
 	"fmt"
+	"hash/fnv"
 	"io"
 	"log"
+	"math/rand"
 	"sort"
 	"strings"
 	"sync"
@@ -28,6 +30,10 @@ type ProbeFunc func(context.Context) error
 
 // a Prober manages a set of probes and keeps track of their results.
 type Prober struct {
+	// Whether to spread probe execution over time by introducing a
+	// random delay before the first probe run.
+	spread bool
+
 	// Time-related functions that get faked out during tests.
 	now       func() time.Time
 	newTicker func(time.Duration) ticker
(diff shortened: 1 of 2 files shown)


prober/prober.go
```diff
@@ -127,12 +140,26 @@ func (p *Probe) Close() error {
...
 
 // probeLoop invokes runProbe on fun every interval. The first probe
-// is run after interval.
+// is run after a random delay (if spreading is enabled) or immediately.
 func (p *Probe) loop() {
 	defer close(p.stopped)
 
-	// Do a first probe right away, so that the prober immediately exports results for everything.
-	p.run()
+	if p.prober.spread && p.initialDelay > 0 {
```
DentonGentry, line 147: We certainly can make it random. However it occurs to me that the caller knows how many probers it is planning to add, and the interval it expects to have them probe. The caller could possibly pass in a value to spread the probers evenly across the interval.

DentonGentry (approved):

prober/prober.go
knyar: That's a good point. Currently the API to add a probe ([Run](https://github.com/tailscale/tailscale/blob/adec726fee58a8d5c8e54fed789abced23737cdb/prober/prober.go#L60)) is pretty hard to add optional parameters to. I am planning to change that to enable custom metrics reported by probes, and as part of that change I can also add ability for an individual probe to set its own `initialDelay`.

DentonGentry: SGTM

PR #6007, opened 2022-10-21
Bradfitz/1.32 picks

bradfitz:


bradfitz:
hey @willnorris, if you're feeling GitHub Action-y, the lack of CI on changes like this is a bit not ideal

DentonGentry (approved):
I think https://github.com/tailscale/tailscale/pull/6007/commits/5a461d9961cc70994255ed8a0c1f7a664247dbbe probably isn't necessary (it isn't built from the release-branch), but may be valuable if it makes future cherry-picks easier.

bradfitz:
> I think https://github.com/tailscale/tailscale/commit/5a461d9961cc70994255ed8a0c1f7a664247dbbe probably isn't necessary (it isn't built from the release-branch), but may be valuable if it makes future cherry-picks easier.

Yeah, I mostly included for reducing future conflicts.

Same with https://github.com/tailscale/tailscale/pull/6007/commits/d6b7ede54d530bed5ca65a11f3b3bbe9050e2e72 and non-user-facing typo fixes.

andrew-d (approved):
35f42ee650e0cba1e9ec643e7b6078a2d716b6e3 is probably unnecessary due to being a test-only fix, but also harmless

mihaip:
> hey @willnorris, if you're feeling GitHub Action-y, the lack of CI on changes like this is a bit not ideal

I think we need to change the `*` on the pull_request filter to `**` so that it matches target branches with slashes in them ([docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)). Or alternatively add `release-branch/*` to be more selective.

Issue #713, opened 2020-08-25
FR: Fast user switching


apenwarr:
(context added on 2022/09/14)

### What are you trying to do?
Log into multiple tailscale accounts at once (eg. example@gmail.com and example@example.com) and easy switch between accounts, i.e. improve the "Log in to a different account..." flow on a device, (possibly) without requiring re-authentication. Access only machines on the network you are currently logged into.

**FR to access all the machines on both networks simultaneously: https://github.com/tailscale/tailscale/issues/183**

### How should we solve this?

The basics of this are not that complicated to do purely client side:
- have ipn.Backend start multiple controlclient instances, one per login
- merge the netmaps from the two instances and feed the merged result to wgengine

Complexity arises from specific details:

1) Should we have multiple local IP addresses, one per account? Is that even possible on all platforms? Or should we get help from the control server to assign the same IP address to all accounts? But if we do that, what about accounts with custom IP subnet designations?

2) If we have multiple local IP addresses, there needs to be a good way to choose which local IP address to use for a given outgoing connection. (Based on routes?)

3) If we use a single local IP address, this could violate users' privacy expectations. Their distinctive identities end up tied together in the control server. It's a lot more elegant if we can support multi-login entirely client side.

4) If two accounts export the same subnet routes (--advertise-subnets), you can get conflicts between them.

5) If two accounts have different DNS overrides, they can conflict. (This isn't so bad if we use the DNS settings only for particular subdomains rather than globally.)

6) Corporate admins may not like the idea of a single machine being on both the corporate and non-corporate networks at once (although it's probably not as bad as being on both the Internet and the corporate network at once.)

happyalu:
Although not _at the same time_, I find it useful to be able to switch accounts to access different groups of machines from my laptop. Prior to 1.0, with the [tailscale-linux-switchuser](https://github.com/happyalu/tailscale-linux-switchuser/) tool I was saving copies of relay.conf for different email addresses and setting `relay.conf` to the required account. It seems that now the information is saved in `tailscaled.state` with a base64 encoded field. 

I tried to save this new file for different email addresses similarly, and copy the right one for the address I want to login to tailscale with, but that didn't seem to work even after restarting tailscaled. Is there an easy way to switch accounts in the current setup? I thought I'll ask before trying to dive down into finding the details of the new file :)

apenwarr:
Note: related issue: https://github.com/tailscale/tailscale/issues/183

apenwarr:
[...]

PR #6020, opened 2022-10-22
cmd/tailscale/cli: add login and switch subcommands

maisem:
```
USAGE
  login [flags]

"tailscale login" logs in this machine to your Tailscale network.

FLAGS
  --accept-dns, --accept-dns=false
    	accept DNS configuration from the admin panel (default true)
  --accept-routes, --accept-routes=false
    	accept routes advertised by other Tailscale nodes (default false)
  --advertise-exit-node, --advertise-exit-node=false
    	offer to be an exit node for internet traffic for the tailnet (default false)
  --advertise-routes string
    	routes to advertise to other nodes (comma-separated, e.g. "10.0.0.0/8,192.168.0.0/24") or empty string to not advertise routes
  --advertise-tags string
    	comma-separated ACL tags to request; each must start with "tag:" (e.g. "tag:eng,tag:montreal,tag:ssh")
  --auth-key string
    	node authorization key; if it begins with "file:", then it's a path to a file containing the authkey
  --exit-node string
    	Tailscale exit node (IP or base name) for internet traffic, or empty string to not use an exit node
  --exit-node-allow-lan-access, --exit-node-allow-lan-access=false
    	Allow direct access to the local network when routing traffic via an exit node (default false)
  --hostname string
    	hostname to use instead of the one provided by the OS
  --login-server string
    	base URL of control server (default https://controlplane.tailscale.com)
  --operator string
    	Unix username to allow to operate on tailscaled without sudo
  --qr, --qr=false
    	show QR code for login URLs (default false)
  --shields-up, --shields-up=false
    	don't allow incoming connections (default false)
  --ssh, --ssh=false
    	run an SSH server, permitting access per tailnet admin's declared policy (default false)
  --timeout duration
    	maximum amount of time to wait for tailscaled to enter a Running state; default (0s) blocks forever (default 0s)
```

```
USAGE
  tailscale switch <name>
  tailscale switch --list

"tailscale switch" switches between logged in profiles
```
Updates #713

cc @mihaip @willnorris @mayakacz


mayakacz:
* To switch, would you do `tailscale profile --switch home` or just `tailscale profile home`? I guess I'm not sure how `switch` would work.
* You also proposed `rename` in the doc. Did you forgo that here in favour of `delete`?
* How does `create` work with current state? Do I run `tailscale profile --create` and it creates a _blank_ profile? do I need to pass it an email? does it create a profile based on my current logged in user?

mayakacz:
I believe this is not in line with what's now in https://docs.google.com/document/d/1HfUIN-teHjniUzntwwtNQ_GZvu72YbN1aWAXxgPvlzk/edit#

bradfitz:
> `--list, --list=false`

Hah, "`--list=false`" wrote nobody ever.

bradfitz:
So we're going to have three commands (up, set, login) with all the prefs flags, eh?

cmd/tailscale/cli/switch.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/peterbourgon/ff/v3/ffcli"
+	"tailscale.com/ipn"
+)
+
+var switchCmd = &ffcli.Command{
+	Name:       "switch",
+	ShortHelp:  "Switches to a different Tailscale profile",
+	LongHelp:   `"tailscale switch" switches between logged in profiles`,
+	ShortUsage: "switch [<name>] [--list]",
```
bradfitz, line 22: It's more like `"switch --list\nswitch <name>\n"` (you can do multiple lines, right?)

cmd/tailscale/cli/switch.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	if profID == cp.ID {
+		printf("Already on profile %q\n", args[0])
+		os.Exit(0)
+	}
+	if err := localClient.SwitchProfile(ctx, profID); err != nil {
+		printf("Failed to switch to profile: %v\n", err)
+		os.Exit(1)
+	}
+	printf("Switched to profile %q\n", args[0])
```
bradfitz, line 82: Switched or Switching? Looks like it might still fail based on following for loop.

maisem:
> So we're going to have three commands (up, set, login) with all the prefs flags, eh?

I was hoping we can start hiding the pref flags from `up` in the next release and start steering users towards `tailscale login` to login and `tailscale set` to toggle prefs. We would leave `tailscale up` as just the analog to `tailscale down`

Another option would be to make `tailscale login` only have `--qr`, `--control-url`, `--authkey` and `--advertise-tags`.

bradfitz:
> Another option would be to make tailscale login only have --qr, --control-url, --authkey and --advertise-tags.

Maybe we add `--up=BOOL` (up, start, run?) defaulting to false, so people can say `--up=false` to login, and then change settings with `tailscale set`, and then `tailscale up`, if they want to avoid a time window where they would've been up an running with the wrong settings (like, say, shields being down)

mihaip (approved):

client/tailscale/localclient.go
```diff
@@ -903,3 +903,39 @@ func (r jsonReader) Read(p []byte) (n int, err error) {
...
+	}
+	body, err = lc.send(ctx, "GET", "/localapi/v0/profiles/", 200, nil)
+	if err != nil {
+		return
+	}
+	all, err = decodeJSON[[]ipn.LoginProfile](body)
+	return current, all, err
+}
+
+// NewProfile creates and switches to a new profile.
```
bradfitz, line 925: say more that it's unnamed, not persisted, etc? when does it get saved? etc

bradfitz: like say why this doesn't return a ProfileID

client/tailscale/localclient.go
```diff
@@ -903,3 +903,39 @@ func (r jsonReader) Read(p []byte) (n int, err error) {
...
+	return err
+}
+
+// SwitchProfile switches to the given profile.
+func (lc *LocalClient) SwitchProfile(ctx context.Context, profile ipn.ProfileID) error {
+	_, err := lc.send(ctx, "POST", "/localapi/v0/profiles/"+url.PathEscape(string(profile)), 204, nil)
+	return err
+}
+
+// DeleteProfile removes the profile with the given name.
```
bradfitz, line 937: given ID, no?

cmd/tailscale/cli/up.go
```diff
@@ -117,7 +120,14 @@ func newUpFlagSet(goos string, upArgs *upArgsT) *flag.FlagSet {
 		upf.BoolVar(&upArgs.forceDaemon, "unattended", false, "run in \"Unattended Mode\" where Tailscale keeps running even after the current GUI user logs out (Windows-only)")
 	}
 	upf.DurationVar(&upArgs.timeout, "timeout", 0, "maximum amount of time to wait for tailscaled to enter a Running state; default (0s) blocks forever")
-	registerAcceptRiskFlag(upf, &upArgs.acceptedRisks)
+
+	if cmd == "up" {
```
bradfitz, line 124: not doing the `--up=false` thing for "login", eh? (even with all the prefs flags too)

maisem: I was going to get back to this as it requires some a bunch more testing on when the login happens and when we run stop.

mayakacz:
> --auth-key string

Please confirm that `--authkey` also works.

Any reason to exclude (or maybe just not noted): `--netfilter-mode`, `--snat-subnet-routes` and `--unattended`?

How do you set name? Is it only with `tailscale set --name` (which I didn't think we were doing yet)? I don't see `tailscale login --name` listed.

maisem:
> > --auth-key string
> 
> Please confirm that `--authkey` also works.

It does


> 
> Any reason to exclude (or maybe just not noted): `--netfilter-mode`, `--snat-subnet-routes` and `--unattended`?

Those are there, just not visible cause I was running this on macOS where we do not have those options.

> 
> How do you set name? Is it only with `tailscale set --name` (which I didn't think we were doing yet)? I don't see `tailscale login --name` listed.

This is going to be a follow up change.

client/tailscale/localclient.go
maisem: yep.

mayakacz (approved):

client/tailscale/localclient.go
```diff
@@ -903,3 +903,43 @@ func (r jsonReader) Read(p []byte) (n int, err error) {
...
+		return
+	}
+	all, err = decodeJSON[[]ipn.LoginProfile](body)
+	return current, all, err
+}
+
+// NewProfile creates and switches to a new unnamed profile. The new profile is
+// not assigned an ID until it is persisted after a successful login.
+// In order to login to the new profile, the user must call LoginInteractive.
+func (lc *LocalClient) NewProfile(ctx context.Context) error {
```
bradfitz, line 928: NewT makes it sounds like it returns T.

How about `SwitchToEmptyProfile` or something?

client/tailscale/localclient.go
```diff
@@ -903,3 +903,43 @@ func (r jsonReader) Read(p []byte) (n int, err error) {
...
+
+// SwitchProfile switches to the given profile.
+func (lc *LocalClient) SwitchProfile(ctx context.Context, profile ipn.ProfileID) error {
+	_, err := lc.send(ctx, "POST", "/localapi/v0/profiles/"+url.PathEscape(string(profile)), 204, nil)
+	return err
+}
+
+// DeleteProfile removes the profile with the given ID.
+// If the profile is the current profile, an empty profile
+// will be selected as if NewProfile was called.
```
bradfitz, line 941: as if SwitchToEmptyProfile (or whatever)

cmd/tailscale/cli/cli.go
```diff
@@ -183,8 +183,12 @@ change in the future.
 		}
 	}
 	if envknob.UseWIPCode() {
-		rootCmd.Subcommands = append(rootCmd.Subcommands, idTokenCmd)
-		rootCmd.Subcommands = append(rootCmd.Subcommands, serveCmd)
+		rootCmd.Subcommands = append(rootCmd.Subcommands,
+			idTokenCmd,
+			serveCmd,
+			loginCmd,
```
bradfitz, line 189: Let's make these like "debug": present by default if you know they're there.

And then if the plan is for them to be alpha, then document them in their ShortHelp + LongHelp loudly that they're alpha. That's what @shayne is doing in his ingress CLI PR.

cmd/tailscale/cli/login.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/peterbourgon/ff/v3/ffcli"
+)
+
+var loginArgs upArgsT
+
+var loginCmd = &ffcli.Command{
+	Name:       "login",
+	ShortUsage: "login [flags]",
+	ShortHelp:  "Log in to a Tailscale account",
+	LongHelp:   `"tailscale login" logs in this machine to your Tailscale network.`,
```
bradfitz, line 20: "logs this machine in to " ?

cmd/tailscale/cli/switch.go
```diff
@@ -0,0 +1,122 @@
...
+	"time"
+
+	"github.com/peterbourgon/ff/v3/ffcli"
+	"tailscale.com/ipn"
+)
+
+var switchCmd = &ffcli.Command{
+	Name:      "switch",
+	ShortHelp: "Switches to a different Tailscale profile",
+	// ShortUsage: "switch <name>\r\n  switch --list",
```
bradfitz, line 21: ?

cmd/tailscale/cli/switch.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	var profID ipn.ProfileID
+	for _, p := range all {
+		if p.Name == args[0] {
+			profID = p.ID
+			break
+		}
+	}
+	if profID == "" {
+		printf("No profile named %q\n", args[0])
```
bradfitz, line 77: to stdout and not stderr for errors? (here and elsewhere)

cmd/tailscale/cli/switch.go
```diff
@@ -0,0 +1,122 @@
...
+		st, err := localClient.StatusWithoutPeers(ctx)
+		if err != nil {
+			outln("Error getting status:", err)
+			os.Exit(1)
+		}
+		switch st.BackendState {
+		case "NoState", "Starting":
+			// TODO(maisem): maybe add a way to subscribe to state changes to
+			// LocalClient.
+			time.Sleep(100 * time.Millisecond)
```
bradfitz, line 105: ouch. but yeah, I want to add the non-HTTP IPN protocol's streaming to a Watch method on LocalClient and then remove the non-HTTP protocol entirely.

bradfitz (approved):

cmd/tailscale/cli/switch.go
maisem: debug

cmd/tailscale/cli/cli.go
maisem: Done

client/tailscale/localclient.go
maisem: Done

client/tailscale/localclient.go
maisem: Done.

Issue #5958, opened 2022-10-16
High CPU consumption on Raspberry Pi 1B even while idle


lutfor-diu:
[cpu.out.txt](https://github.com/tailscale/tailscale/files/9801639/cpu.out.txt)
[cpu.out.txt](https://github.com/tailscale/tailscale/files/9801644/cpu.out.txt)
### What is the issue?

Resources hungry! Usage lots of cpu in ideal mode on raspberry pi. Please update it and make it more light

### Steps to reproduce

When using it raspberry pi 1b it consume 20% of cpu when in tailscale ideal mode. 

### Are there any recent changes that introduced the issue?

Yes, after recent update

### OS

Linux

### OS version

Bullsey

### Tailscale version

Tailscale latest version

bradfitz:
Without a bug report or profile or even version number there's little we can do here.

"Tailscale latest version" doesn't age well in bug reports.

DentonGentry:
- Is the Raspberry Pi being used as an exit node?
- Can you run "tailscale bugreport" on the Raspberry Pi during a time when its CPU is 20% and paste it here?

unnatural:
> Without a bug report or profile or even version number there's little we can do here.
> 
> "Tailscale latest version" doesn't age well in bug reports.

1.32.0

unnatural:
![IMG_20221016_211105](https://user-images.githubusercontent.com/32225836/196043241-865b7013-cf77-4379-a50d-84b5d0b4e36c.jpg)

DentonGentry:
Please run "tailscale bugreport" on the Raspberry Pi and paste it here.
A screenshot of the top output doesn't really help.

unnatural:
> Please run "tailscale bugreport" on the Raspberry Pi and paste it here. A screenshot of the top output doesn't really help.

Ok brother, previous tailscaled just consume 4-8% when in use.

unnatural:
root@DietPi:~# tailscale bugreport
BUG-2403f7f1b20fb1122c0ad05d6540ac4d0a45782d8ba3aa9c107d7bf0c0420d45-20221016152223Z-ebb88c91ca253a25

DentonGentry:
On the Raspberry Pi can you run:
tailscale debug -cpu-profile=/tmp/cpu.out --profile-seconds=60

and attach the /tmp/cpu.out file here?

bradfitz:
I see no smoking gun.

[...]

PR #6021, opened 2022-10-22
portlist: further reduce allocations on Linux

bradfitz:
Make Linux parsePorts also an append-style API and attach it to caller's provided append base memory.

And add a little string intern pool in front of the []byte to string for inode names.

    name       old time/op    new time/op    delta
    GetList-8    11.1ms ± 4%     9.8ms ± 6%  -11.68%  (p=0.000 n=9+10)

    name       old alloc/op   new alloc/op   delta
    GetList-8    92.8kB ± 2%    79.7kB ± 0%  -14.11%  (p=0.000 n=10+9)

    name       old allocs/op  new allocs/op  delta
    GetList-8     2.94k ± 1%     2.76k ± 0%   -6.16%  (p=0.000 n=10+10)

More coming. (the bulk of the allocations are in addProcesses and filesystem operations, most of which we should usually be able to skip)

Updates #5958

### DIFF
--- a/portlist/portlist_linux.go
+++ b/portlist/portlist_linux.go
@@ -6,6 +6,7 @@ package portlist
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io"
 	"os"
@@ -13,12 +14,14 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"syscall"
 	"time"
 
 	"go4.org/mem"
 	"golang.org/x/sys/unix"
+	"tailscale.com/util/mak"
 )
 
 // Reading the sockfiles on Linux is very fast, so we can do it often.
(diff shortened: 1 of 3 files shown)


portlist/portlist_linux.go
```diff
@@ -59,25 +88,20 @@ func appendListeningPorts(base []Port) ([]Port, error) {
...
 		}
-		if br == nil {
-			br = bufio.NewReader(f)
-		} else {
-			br.Reset(f)
-		}
-		ports, err := parsePorts(br, filepath.Base(fname))
+		br.Reset(f)
+		ret, err = appendParsePorts(ret, br, filepath.Base(fname))
 		f.Close()
```
DentonGentry, line 93: I think this results in br being put back in the syncpool still holding a reference to f, so f won't be garbage collected until the next time br gets used and Reset(eofReader).

Which is maybe fine, since it has closed the actual file and so won't hold it open.

portlist/portlist_linux.go
```diff
@@ -171,7 +198,7 @@ func parsePorts(r *bufio.Reader, fileBase string) ([]Port, error) {
 		ret = append(ret, Port{
 			Proto: proto,
 			Port:  uint16(portv),
-			inode: string(inoBuf),
+			inode: stringCache.get(inoBuf),
```
DentonGentry, line 201: If I'm reading this correctly, "socket:[inode_number]" strings will be put into the pool for unix local sockets, and there isn't a bound on the size of the pool. The potential number of inodes is large, if sockets are being created regularly.

Will there only ever be one socket, the one between tailscale and tailscaled?

DentonGentry (approved):

portlist/portlist_linux.go
bradfitz: See the `br.Reset` above:

```go
	br := bufioReaderPool.Get().(*bufio.Reader)
	defer bufioReaderPool.Put(br)
	defer br.Reset(eofReader)
```

That means before it's put back in the pool, we drop the reference to `f`.

portlist/portlist_linux.go
bradfitz: There will be one per listening process.

They're at least in a sync.Pool so if/when there's memory pressure the whole map will be dropped.

But I can make it better and drop the map if/when it gets double the expected size.

bradfitz: PTAL

DentonGentry: LGTM

DentonGentry (approved):

PR #6022, opened 2022-10-22
ipn/ipnlocal: add support for multiple profiles

maisem:

Updates #713

### DIFF
--- a/cmd/tailscale/cli/up.go
+++ b/cmd/tailscale/cli/up.go
@@ -634,27 +634,10 @@ func runUp(ctx context.Context, args []string) (retErr error) {
 		if err != nil {
 			return err
 		}
-		opts := ipn.Options{
-			StateKey:    ipn.GlobalDaemonStateKey,
+		bc.Start(ipn.Options{
 			AuthKey:     authKey,
 			UpdatePrefs: prefs,
-		}
-		// On Windows, we still run in mostly the "legacy" way that
-		// predated the server's StateStore. That is, we send an empty
-		// StateKey and send the prefs directly. Although the Windows
-		// supports server mode, though, the transition to StateStore
-		// is only half complete. Only server mode uses it, and the
-		// Windows service (~tailscaled) is the one that computes the
-		// StateKey based on the connection identity. So for now, just
-		// do as the Windows GUI's always done:
-		if effectiveGOOS() == "windows" {
-			// The Windows service will set this as needed based
-			// on our connection's identity.
-			opts.StateKey = ""
-			opts.Prefs = prefs
-		}
-
-		bc.Start(opts)
+		})
 		if upArgs.forceReauth {
 			startLoginInteractive()
 		}
--- a/control/controlclient/auto.go
+++ b/control/controlclient/auto.go
@@ -590,7 +590,7 @@ func (c *Auto) sendStatus(who string, err error, url string, nm *netmap.NetworkM
 	}
 	if nm != nil && loggedIn && synced {
 		pp := c.direct.GetPersist()
-		p = &pp
+		p = pp.AsStruct()
 	} else {
 		// don't send netmap status, as it's misleading when we're
 		// not logged in.
@@ -708,7 +708,7 @@ func (c *Auto) Shutdown() {
 // used exclusively in tests.
 func (c *Auto) TestOnlyNodePublicKey() key.NodePublic {
 	priv := c.direct.GetPersist()
-	return priv.PrivateNodeKey.Public()
+	return priv.PrivateNodeKey().Public()
 }
 
 func (c *Auto) TestOnlySetAuthKey(authkey string) {
--- a/control/controlclient/auto.go
+++ b/control/controlclient/auto.go
@@ -580,7 +580,7 @@ func (c *Auto) sendStatus(who string, err error, url string, nm *netmap.NetworkM
 
 	c.logf("[v1] sendStatus: %s: %v", who, state)
 
-	var p *persist.Persist
+	var p *persist.PersistView
 	var loginFin, logoutFin *empty.Message
 	if state == StateAuthenticated {
 		loginFin = new(empty.Message)
@@ -590,7 +590,7 @@ func (c *Auto) sendStatus(who string, err error, url string, nm *netmap.NetworkM
 	}
 	if nm != nil && loggedIn && synced {
 		pp := c.direct.GetPersist()
-		p = pp.AsStruct()
+		p = &pp
 	} else {
 		// don't send netmap status, as it's misleading when we're
 		// not logged in.
--- a/ipn/backend.go
+++ b/ipn/backend.go
@@ -20,13 +20,13 @@ import (
 type State int
 
 const (
-	NoState = State(iota)
-	InUseOtherUser
-	NeedsLogin
-	NeedsMachineAuth
-	Stopped
-	Starting
-	Running
+	NoState          State = 0
+	InUseOtherUser   State = 1
+	NeedsLogin       State = 2
+	NeedsMachineAuth State = 3
+	Stopped          State = 4
+	Starting         State = 5
+	Running          State = 6
 )
 
 // GoogleIDToken Type is the tailcfg.Oauth2Token.TokenType for the Google
@@ -153,21 +153,8 @@ type PartialFile struct {
 }
 
 // StateKey is an opaque identifier for a set of LocalBackend state
-// (preferences, private keys, etc.).
-//
-// The reason we need this is that the Tailscale agent may be running
-// on a multi-user machine, in a context where a single daemon is
-// shared by several consecutive users. Ideally we would just use the
-// username of the connected frontend as the StateKey.
-//
-// Various platforms currently set StateKey in different ways:
-//
-//   - the macOS/iOS GUI apps set it to "ipn-go-bridge"
-//   - the Android app sets it to "ipn-android"
-//   - on Windows, it's the empty string (in client mode) or, via
-//     LocalBackend.userID, a string like "user-$USER_ID" (used in
-//     server mode).
-//   - on Linux/etc, it's always "_daemon" (ipn.GlobalDaemonStateKey)
+// (preferences, private keys, etc.). It is also used as a key for
+// the various LoginProfiles that the instance may be signed into.
 //
 // Additionally, the StateKey can be debug setting name:
 //
@@ -178,21 +165,10 @@ type StateKey string
 type Options struct {
 	// FrontendLogID is the public logtail id used by the frontend.
 	FrontendLogID string
-	// StateKey and Prefs together define the state the backend should
-	// use:
-	//  - StateKey=="" && Prefs!=nil: use Prefs for internal state,
-	//    don't persist changes in the backend, except for the machine key
-	//    for migration purposes.
-	//  - StateKey!="" && Prefs==nil: load the given backend-side
-	//    state and use/update that.
-	//  - StateKey!="" && Prefs!=nil: like the previous case, but do
-	//    an initial overwrite of backend state with Prefs.
-	//
-	// NOTE(apenwarr): The above means that this Prefs field does not do
-	// what you probably think it does. It will overwrite your encryption
-	// keys. Do not use unless you know what you're doing.
-	StateKey StateKey
-	Prefs    *Prefs
+	// LegacyMigrationPrefs are used to migrate preferences from the
+	// frontend to the backend.
+	// If non-nil, they are imported as a new profile.
+	LegacyMigrationPrefs *Prefs `json:"Prefs"`
 	// UpdatePrefs, if provided, overrides Options.Prefs *and* the Prefs
 	// already stored in the backend state, *except* for the Persist
 	// Persist member. If you just want to provide prefs, this is
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -818,10 +818,10 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 		prefs.ControlURL = prefs.ControlURLOrDefault()
 		prefsChanged = true
 	}
-	if st.Persist != nil {
-		if !prefs.Persist.Equals(st.Persist) {
+	if st.Persist != nil && st.Persist.Valid() {
+		if !prefs.Persist.View().Equals(*st.Persist) {
 			prefsChanged = true
-			prefs.Persist = st.Persist.Clone()
+			prefs.Persist = st.Persist.AsStruct()
 		}
 	}
 	if st.URL != "" {
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -124,6 +124,7 @@ type LocalBackend struct {
 	keyLogf               logger.Logf        // for printing list of peers on change
 	statsLogf             logger.Logf        // for printing peers stats on change
 	e                     wgengine.Engine
+	pm                    *profileManager
 	store                 ipn.StateStore
 	dialer                *tsdial.Dialer // non-nil
 	backendLogID          string
@@ -138,6 +139,10 @@ type LocalBackend struct {
 	sshAtomicBool         atomic.Bool
 	shutdownCalled        bool // if Shutdown has been called
 
+	// lastProfileID tracks the last profile we've seen from the ProfileManager.
+	// It's used to detect when the user has changed their profile.
+	lastProfileID ipn.ProfileID
+
 	filterAtomic                 atomic.Pointer[filter.Filter]
 	containsViaIPFuncAtomic      syncs.AtomicValue[func(netip.Addr) bool]
 	shouldInterceptTCPPortAtomic syncs.AtomicValue[func(uint16) bool]
@@ -151,9 +156,6 @@ type LocalBackend struct {
 	notify         func(ipn.Notify)
 	cc             controlclient.Client
 	ccAuto         *controlclient.Auto // if cc is of type *controlclient.Auto
-	stateKey       ipn.StateKey        // computed in part from user-provided value
-	userID         string              // current controlling user ID (for Windows, primarily)
-	prefs          ipn.PrefsView       // may not be Valid.
 	inServerMode   bool
 	machinePrivKey key.MachinePrivate
 	nlPrivKey      key.NLPrivate
@@ -226,9 +228,14 @@ type clientGen func(controlclient.Options) (controlclient.Client, error)
 // but is not actually running.
 //
 // If dialer is nil, a new one is made.
-func NewLocalBackend(logf logger.Logf, logid string, store ipn.StateStore, dialer *tsdial.Dialer, e wgengine.Engine, loginFlags controlclient.LoginFlags) (*LocalBackend, error) {
+func NewLocalBackend(logf logger.Logf, logid string, store ipn.StateStore, stateKey ipn.StateKey, dialer *tsdial.Dialer, e wgengine.Engine, loginFlags controlclient.LoginFlags) (*LocalBackend, error) {
 	if e == nil {
 		panic("ipn.NewLocalBackend: engine must not be nil")
+	}
+
+	pm, err := newProfileManager(store, logf, stateKey)
+	if err != nil {
+		return nil, err
 	}
 
 	hi := hostinfo.New()
@@ -253,7 +260,8 @@ func NewLocalBackend(logf logger.Logf, logid string, store ipn.StateStore, diale
 		keyLogf:        logger.LogOnChange(logf, 5*time.Minute, time.Now),
 		statsLogf:      logger.LogOnChange(logf, 5*time.Minute, time.Now),
 		e:              e,
-		store:          store,
+		pm:             pm,
+		store:          pm.Store(),
 		dialer:         dialer,
 		backendLogID:   logid,
 		state:          ipn.NoState,
@@ -292,7 +300,7 @@ func NewLocalBackend(logf logger.Logf, logid string, store ipn.StateStore, diale
 
 	for _, component := range debuggableComponents {
 		key := componentStateKey(component)
-		if ut, err := ipn.ReadStoreInt(store, key); err == nil {
+		if ut, err := ipn.ReadStoreInt(pm.Store(), key); err == nil {
 			if until := time.Unix(ut, 0); until.After(time.Now()) {
 				// conditional to avoid log spam at start when off
 				b.SetComponentDebugLogging(component, until)
@@ -450,7 +458,7 @@ func (b *LocalBackend) linkChange(major bool, ifst *interfaces.State) {
 
 	// If the local network configuration has changed, our filter may
 	// need updating to tweak default routes.
-	b.updateFilterLocked(b.netMap, b.prefs)
+	b.updateFilterLocked(b.netMap, b.pm.CurrentPrefs())
 
 	if peerAPIListenAsync && b.netMap != nil && b.state == ipn.Running {
 		want := len(b.netMap.Addresses)
@@ -520,7 +528,7 @@ func stripKeysFromPrefs(p ipn.PrefsView) ipn.PrefsView {
 func (b *LocalBackend) Prefs() ipn.PrefsView {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	return stripKeysFromPrefs(b.prefs)
+	return stripKeysFromPrefs(b.pm.CurrentPrefs())
 }
 
 // Status returns the latest status of the backend and its
@@ -577,14 +585,14 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 			s.CurrentTailnet.MagicDNSSuffix = b.netMap.MagicDNSSuffix()
 			s.CurrentTailnet.MagicDNSEnabled = b.netMap.DNS.Proxied
 			s.CurrentTailnet.Name = b.netMap.Domain
-			if b.prefs.Valid() && !b.prefs.ExitNodeID().IsZero() {
-				if exitPeer, ok := b.netMap.PeerWithStableID(b.prefs.ExitNodeID()); ok {
+			if prefs := b.pm.CurrentPrefs(); prefs.Valid() && !prefs.ExitNodeID().IsZero() {
+				if exitPeer, ok := b.netMap.PeerWithStableID(prefs.ExitNodeID()); ok {
 					var online = false
 					if exitPeer.Online != nil {
 						online = *exitPeer.Online
 					}
 					s.ExitNodeStatus = &ipnstate.ExitNodeStatus{
-						ID:           b.prefs.ExitNodeID(),
+						ID:           prefs.ExitNodeID(),
 						Online:       online,
 						TailscaleIPs: exitPeer.Addresses,
 					}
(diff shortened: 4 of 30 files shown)


types/views/views.go
```diff
@@ -201,6 +201,12 @@ func (v IPPrefixSlice) AsSlice() []netip.Prefix {
 	return v.ж.AsSlice()
 }
 
+// Filter returns a new slice, not aliasing in, containing elements of in that
```
bradfitz, line 204: there is no `in` variable here.

ipn/backend.go
```diff
@@ -67,7 +67,7 @@ type Notify struct {
 
 	LoginFinished *empty.Message     // non-nil when/if the login process succeeded
 	State         *State             // if non-nil, the new or current IPN state
-	Prefs         *Prefs             // if non-nil, the new or current preferences
+	Prefs         PrefsView          // if non-nil, the new or current preferences
```
bradfitz: // if Valid, the new ...

ipn/ipnlocal/local.go
```diff
@@ -147,7 +147,7 @@ type LocalBackend struct {
 	ccAuto         *controlclient.Auto // if cc is of type *controlclient.Auto
 	stateKey       ipn.StateKey        // computed in part from user-provided value
 	userID         string              // current controlling user ID (for Windows, primarily)
-	prefs          *ipn.Prefs
+	prefs          ipn.PrefsView
```
bradfitz: // may not be Valid

ipn/ipnlocal/local.go
```diff
@@ -876,33 +876,33 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 
 // findExitNodeIDLocked updates b.prefs to reference an exit node by ID,
 // rather than by IP. It returns whether prefs was mutated.
-func (b *LocalBackend) findExitNodeIDLocked(nm *netmap.NetworkMap) (prefsChanged bool) {
+func (b *LocalBackend) findExitNodeIDLocked(prefs *ipn.Prefs, nm *netmap.NetworkMap) (prefsChanged bool) {
```
bradfitz, line 877: these docs are now confusing. Does it update b.prefs or prefs?

tstest/integration/integration_test.go
```diff
@@ -845,7 +845,7 @@ func (n *testNode) MustUp(extraArgs ...string) {
 func (n *testNode) MustDown() {
 	t := n.env.t
 	t.Logf("Running down ...")
-	if err := n.Tailscale("down").Run(); err != nil {
+	if err := n.Tailscale("down", "--accept-risk=lose-ssh").Run(); err != nil {
```
bradfitz, line 843: unrelated?

aside: should we have --accept-risk=all for stuff like this?

maisem: probably yeah, will send separately

cmd/tailscale/cli/up.go
```diff
@@ -647,7 +646,6 @@ func runUp(ctx context.Context, args []string) (retErr error) {
 		if effectiveGOOS() == "windows" {
 			// The Windows service will set this as needed based
```
mihaip, line 647: Remove comment (since it was presumably about the clearing of the `StateKey` opt)?

ipn/ipnlocal/local.go
```diff
@@ -835,19 +840,15 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
...
 	if prefsChanged {
-		if stateKey != "" {
-			if err := b.store.WriteState(stateKey, prefs.ToBytes()); err != nil {
-				b.logf("Failed to save new controlclient state: %v", err)
-			}
-		}
 		b.send(ipn.Notify{Prefs: prefs.View()})
 	}
+
+	// Now complete the lock-free parts of what we started while locked.
```
mihaip, line 851: Duplicates comment from line 846? (lock state of `b.mu` has not changed)

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,313 @@
...
+	pm.prefs = emptyPrefs
+	pm.currentState = ""
+	if err := pm.useCurrentProfileAtStartup(); err != nil {
+		return err
+	}
+	return pm.store.WriteState(ipn.ServerModeStartKey, nil)
+}
+
+// SetPrefs sets the current profile's prefs to the provided value. It also
+// saves the prefs to the StateStore. If the currnet profile is named, it stores
```
mihaip, line 68: Typo with "currnet"

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,313 @@
...
+// a copy of the provided prefs, which maybe accessed via CurrentPrefs.
```
mihaip, line 69: Typo with "maybe"

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,313 @@
...
+func (pm *ProfileManager) SetPrefs(prefsIn *ipn.Prefs) error {
```
mihaip, line 70: Any reason why this doesn't take a `PrefsView` to more clearly indicate that it doesn't mutate or take ownership of the prefs?

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,313 @@
...
+		}
+	}
+
+	if stateKey == "" && len(knownProfiles) == 0 && goos != "windows" {
+		knownProfiles = []string{"default"}
+		stateKey = "profile-default"
+
+		// No known profiles, see if there is a "_daemon" profile.
+		// If so, migrate it to the new format.
+		b, err := store.ReadState(ipn.GlobalDaemonStateKey)
```
mihaip, line 285: FWIW the Mac and iOS apps used a different state key for some (presumably historical/migration) reasons, so wed' need to read that too: https://github.com/tailscale/corp/blob/d96b92a627aeba0bec96591fc3e87ea7f8e66bb4/xcode/ipn-go-bridge/ipn.go#L245-L258

ipn/store.go
```diff
@@ -26,6 +26,13 @@ const (
 	// particular), but right now Unix daemons run with a single
 	// node-global state. To keep open the option of having per-user state
 	// later, the global state key doesn't look like a username.
+	//
+	// As of 2022-10-21, it has been superceded by profiles and is no longer
```
mihaip, line 30: Typo with "superceded"

ipn/ipnlocal/profiles.go
maisem: Done

ipn/ipnlocal/profiles.go
maisem: Good point, done.

cmd/tailscale/cli/up.go
```diff
@@ -648,9 +647,6 @@ func runUp(ctx context.Context, args []string) (retErr error) {
 		// StateKey based on the connection identity. So for now, just
```
mihaip, line 647: `StateKey`-related comment that probably needs updating.

ipn/store.go
```diff
@@ -26,6 +26,13 @@ const (
...
 	// node-global state. To keep open the option of having per-user state
 	// later, the global state key doesn't look like a username.
+	//
+	// As of 2022-10-21, it has been superseded by profiles and is no longer
+	// written to disk. It is only read at startup when there are no profiles,
+	// to migrate the state to the "default" profile.
+	// The existing state is left on disk in case the user downgrades to an
+	// older version of Tailscale that doesn't support profiles. We can
+	// remove this in a future release.
 	GlobalDaemonStateKey = StateKey("_daemon")
```
mihaip, line 36: Maybe rename this to `LegacyGlobalDaemonStateKey` or something like that to make it clearer that no new uses should be added?

control/controlclient/auto.go
```diff
@@ -590,7 +590,7 @@ func (c *Auto) sendStatus(who string, err error, url string, nm *netmap.NetworkM
 	}
 	if nm != nil && loggedIn && synced {
 		pp := c.direct.GetPersist()
-		p = &pp
+		p = pp.AsStruct()
```
mihaip, line 593: Should this be passing down the `PersistView` instead? I assume we don't expect `statusFunc` to mutate things.

ipn/backend.go
```diff
@@ -178,21 +178,11 @@ type StateKey string
 type Options struct {
```
mihaip, line 178: There's some `StateKey` comments right above this that need updating (as far as removing `_daemon` references, etc.)

ipn/ipnlocal/local.go
```diff
@@ -2012,6 +1911,7 @@ func (b *LocalBackend) setTCPPortsIntercepted(ports []uint16) {
 // setAtomicValuesFromPrefs populates sshAtomicBool and containsViaIPFuncAtomic
 // from the prefs p, which may be nil.
 func (b *LocalBackend) setAtomicValuesFromPrefs(p ipn.PrefsView) {
```
mihaip, line 1913: Comment and possibly function name may need to be updated since it does more now.

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,435 @@
...
+	"time"
+
+	"tailscale.com/ipn"
+	"tailscale.com/types/logger"
+	"tailscale.com/util/strs"
+)
+
+type loginProfile struct {
+	ID   string
+	Name string
```
mihaip, line 22: Can we persist a few more fields from the login? Ideally `DisplayName`, `LoginName`, and `ProfilePicURL`, so that we have have enough information to render the Mac menu item for all profiles?

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,435 @@
...
+		pm.prefs = emptyPrefs
+		pm.isNewProfile = true
+	} else {
+		return err
+	}
+	pm.currentUserID = uid
+	return nil
+}
+
+func (pm *ProfileManager) findProfileByKey(key ipn.StateKey) *loginProfile {
```
mihaip, line 81: Should this have a `Locked` suffix?

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,435 @@
...
+	}
+	if err := pm.store.WriteState(key, prefs.ToBytes()); err != nil {
+		pm.logf("WriteState(%q): %v", key, err)
+		return err
+	}
+	return nil
+}
+
+// Profiles returns the list of known profiles.
+func (pm *ProfileManager) Profiles() []string {
```
mihaip, line 190: Returning a struct with the `DisplayName`, `ProfilePicURL`, etc. fields would also be good for the Mac UI.

ipn/ipnlocal/local.go
```diff
@@ -4085,3 +3986,26 @@ func (b *LocalBackend) SetDevStateStore(key, value string) error {
 func (b *LocalBackend) ShouldInterceptTCPPort(port uint16) bool {
 	return b.shouldInterceptTCPPortAtomic.Load()(port)
 }
+
+func (b *LocalBackend) SwitchProfile(profile string) error {
+	if err := b.pm.SwitchProfile(profile); err != nil {
+		return nil
```
mihaip, line 3992: Should this be `return err`? I had a bug where I was passing in the profile name instead of the ID, and the "not found" error was being dropped.

Also, may want to add an explicit `ProfileID` type instead of `string` so that it's more obvious how profiles are identified.

maisem: ah yes it should be return err.

I was thinking about making the ProfileID a type too. Will do.

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,478 @@
...
+func (pm *ProfileManager) Profiles() []LoginProfile {
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
+	var profiles []LoginProfile
+	for _, p := range pm.knownProfiles {
+		if p.LocalUserID == pm.currentUserID {
+			profiles = append(profiles, *p)
+		}
+	}
+	return profiles
```
mihaip, line 229: Should probably sort these (by display or login name), they're in random/map iteration order right now (I know I can do the sort on the Mac UI side, but this way all clients will be consistent).

ipn/ipnlocal/state_test.go
```diff
@@ -918,6 +920,7 @@ func TestStateMachine(t *testing.T) {
 	}
 }
 
+/*
```
bradfitz, line 923: hah

ipn/localapi/localapi.go
```diff
@@ -71,16 +71,19 @@ var handler = map[string]localAPIHandler{
...
-	"set-dns":                 (*Handler).serveSetDNS,
-	"set-expiry-sooner":       (*Handler).serveSetExpirySooner,
-	"status":                  (*Handler).serveStatus,
-	"tka/init":                (*Handler).serveTKAInit,
-	"tka/modify":              (*Handler).serveTKAModify,
-	"tka/sign":                (*Handler).serveTKASign,
-	"tka/status":              (*Handler).serveTKAStatus,
-	"upload-client-metrics":   (*Handler).serveUploadClientMetrics,
-	"whois":                   (*Handler).serveWhoIs,
+	// TODO: rename to debug-profile or pprof-profile to separatate it from the
```
bradfitz, line 74: TODO(who)?

and typo: separate

bradfitz: But I don't think we can/should be renaming it at this point. Just add a comment like `// pprof (debug) profiles, not user profiles`

ipn/backend.go
```diff
@@ -178,21 +178,11 @@ type StateKey string
...
-	// NOTE(apenwarr): The above means that this Prefs field does not do
-	// what you probably think it does. It will overwrite your encryption
-	// keys. Do not use unless you know what you're doing.
-	StateKey StateKey
-	Prefs    *Prefs
+	// Prefs is the initial preferences to use. If nil, the current
+	// profile's preferences are loaded from the store.
+	// If non-nil, the Prefs are used as-is, and the state store is
+	// updated to match.
+	Prefs *Prefs
```
bradfitz, line 185:
```
LegacyMigrationPrefs *Prefs `json:"Prefs"` ?
```

ipn/localapi/localapi.go
maisem: I was going to add "/pprof" as an alias to "/profile".

maisem: and add the comment

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,467 @@
...
+	pm.prefs = emptyPrefs
+	pm.currentUserID = ""
+	pm.currentProfile = nil
+}
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
+func randID() string {
```
mihaip, line 121: Maybe call this `randProfileID` and have it return a `ProfileID` type directly?

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,467 @@
...
+
+	// isNewProfile is a sentinel value that indicates that the
+	// current profile is new and has not been saved to disk yet.
+	// It is reset to false after a call to SetPrefs with a filled
+	// in LoginName.
+	isNewProfile bool
+}
+
+// CurrentUser returns the current user ID. It is only non-empty on
+// Windows where we have a multi-user system.
```
mihaip, line 43: Are we going to want something like this for the macsys build too, which has a shared storage for all users? CC @nickoneill

ipn/ipnlocal/profiles.go
mihaip: (can also inline it in `newUnusedKey`, since it's only used once)

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,467 @@
...
+	defer pm.mu.Unlock()
+	if pm.currentProfile == nil {
+		return ipn.LoginProfile{}
+	}
+	return *pm.currentProfile
+}
+
+var ErrProfileNotFound = errors.New("profile not found")
+
+// DeleteProfile removes the profile with the given id. It is a no-op if the
```
mihaip, line 290: It does return an error though, so a caller may end up stopping instead of continuing if the profile does not exist.

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,467 @@
...
+
+func (pm *ProfileManager) writeKnownProfiles() error {
+	b, err := json.Marshal(pm.knownProfiles)
+	if err != nil {
+		return err
+	}
+	return pm.store.WriteState(ipn.KnownProfilesStateKey, b)
+}
+
+// NewProfile creates a new profile with the given name. It switches to the new
```
mihaip, line 317: No name is given.

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,467 @@
...
+// profile does not exist.
+func (pm *ProfileManager) DeleteProfile(profile ipn.ProfileID) error {
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
+	kp, ok := pm.knownProfiles[profile]
+	if !ok {
+		return ErrProfileNotFound
+	}
+	if kp.ID == pm.currentProfile.ID {
+		pm.NewProfile()
```
mihaip, line 300: Mention this side effect (and/or return an error -- perhaps the caller does not want a new profile and instead should switch to a different one first)?

ipn/prefs.go
```diff
@@ -681,3 +681,20 @@ func SavePrefs(filename string, p *Prefs) {
 		log.Printf("SavePrefs: %v\n", err)
 	}
 }
+
+// ProfileID is a system-wide unique identifier for a login profile.
```
bradfitz, line 685: what's it look like? examples? restrictions on allowed chars?

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,467 @@
...
+	prefs.WantRunning = false
+	return prefs.View()
+}()
+
+// Store returns the StateStore used by the ProfileManager.
+func (pm *ProfileManager) Store() ipn.StateStore {
+	return pm.store
+}
+
+func (pm *ProfileManager) ReadProfile(p ipn.ProfileID) (ipn.PrefsView, error) {
```
mihaip, line 339: Needs docs (but is also unused)?

ipn/store.go
```diff
@@ -40,8 +47,23 @@ const (
...
+	// KnownProfilesStateKey is the key under which we store the list of
+	// known profiles.
+	KnownProfilesStateKey = StateKey("_profiles")
+
+	// CurrentProfileStateKey is the key under which we store the current
+	// profile.
+	CurrentProfileStateKey = StateKey("_current-profile")
 )
 
+func CurrentProfileKey(userID string) StateKey {
```
bradfitz, line 60: docs

ipn/store.go
```diff
@@ -40,8 +47,23 @@ const (
 	// NLKeyStateKey is the key under which we store the node's
 	// network-lock node key, in its key.NLPrivate.MarshalText representation.
 	NLKeyStateKey = StateKey("_nl-node-key")
+
+	// KnownProfilesStateKey is the key under which we store the list of
```
bradfitz, line 51: store encoded how?

cmd/tailscale/cli/up.go
```diff
@@ -635,23 +635,11 @@ func runUp(ctx context.Context, args []string) (retErr error) {
...
-		// is only half complete. Only server mode uses it, and the
-		// Windows service (~tailscaled) is the one that computes the
-		// StateKey based on the connection identity. So for now, just
-		// do as the Windows GUI's always done:
 		if effectiveGOOS() == "windows" {
-			// The Windows service will set this as needed based
-			// on our connection's identity.
-			opts.StateKey = ""
-			opts.Prefs = prefs
+			opts.LegacyMigrationPrefs = prefs
```
bradfitz, line 642: all the comments about why this is windows-specific disappeared

ipn/ipnlocal/profiles.go
nickoneill: Seems reasonable. I haven't experimented with multi-user macOS installs in a while and maybe never with macsys builds, but prefixing the state files on macsys with a userid would be a nice improvement for anyone who uses more than one login.

maisem: that is going to need to happen at a different layer than the ProfileManager as we need to do UserID detection and have that plumbed in. It's not the same as using `os.GetUid()` 

See https://github.com/tailscale/tailscale/blob/446057d6134055ab60d96f4ceee67c9d89e812f3/ipn/ipnserver/server.go#L138-L185

ipn/ipnlocal/profiles.go
maisem: Thats what I had initially but then realized that there was no way to delete the last profile.

cmd/tailscale/cli/up.go
maisem: Actually I don't think we need this anymore either.

ipn/prefs.go
maisem: its auto-generated, added docs

ipn/ipnlocal/profiles.go
maisem: Done

ipn/ipnlocal/profiles.go
maisem: Updated docs

ipn/ipnlocal/profiles.go
maisem: Oh I added this when I thought we would need to pull Persist info out of the prefs and then decided to store them in the LoginProfile instead

ipn/ipnlocal/profiles.go
maisem: Updated docs.

ipn/ipnlocal/profiles.go
maisem: Updated docs.

ipn/ipnlocal/local.go
```diff
@@ -4093,3 +4006,34 @@ func (b *LocalBackend) SetDevStateStore(key, value string) error {
 func (b *LocalBackend) ShouldInterceptTCPPort(port uint16) bool {
 	return b.shouldInterceptTCPPortAtomic.Load()(port)
 }
+
+func (b *LocalBackend) SwitchProfile(profile ipn.ProfileID) error {
+	if b.CurrentProfile().ID == profile {
+		return nil
+	}
+	b.enterState(ipn.NoState) // Reset state.
```
mihaip, line 4014: May want to do an early check that the profile is known, otherwise you'll mutate the state but return an error.

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,458 @@
...
+	}
+
+	return pm, nil
+}
+
+func (pm *ProfileManager) migrateFromLegacyPrefs() error {
+	pm.NewProfile()
+	k := ipn.LegacyGlobalDaemonStateKey
+	switch runtime.GOOS {
+	case "ios", "darwin":
```
mihaip, line 436: `tailscaled` on macOS does not use `ipn-go-bridge`, it should still read from `LegacyGlobalDaemonStateKey`

mihaip (approved):

ipn/ipnlocal/profiles.go
maisem: Good catch! I thought about it last night but promptly forgot :)

ipn/ipnlocal/local.go
maisem: The lock ordering makes it slightly tricky. Done.

ipn/ipnlocal/local.go
```diff
@@ -227,7 +229,7 @@ type clientGen func(controlclient.Options) (controlclient.Client, error)
 // but is not actually running.
 //
 // If dialer is nil, a new one is made.
-func NewLocalBackend(logf logger.Logf, logid string, store ipn.StateStore, dialer *tsdial.Dialer, e wgengine.Engine, loginFlags controlclient.LoginFlags) (*LocalBackend, error) {
+func NewLocalBackend(logf logger.Logf, logid string, pm *ProfileManager, dialer *tsdial.Dialer, e wgengine.Engine, loginFlags controlclient.LoginFlags) (*LocalBackend, error) {
 	if e == nil {
```
bradfitz, line 233: can pm be nil? we panic here if e is nil. No comment on the field above.

ipn/ipnlocal/local.go
```diff
@@ -4093,3 +4012,33 @@ func (b *LocalBackend) SetDevStateStore(key, value string) error {
...
+	return b.pm.DeleteProfile(p)
+}
+
+func (b *LocalBackend) CurrentProfile() ipn.LoginProfile {
+	return b.pm.CurrentProfile()
+}
+
+func (b *LocalBackend) NewProfile() {
+	b.pm.NewProfile()
+}
```
bradfitz, line 4040: this is mysterious without a return value or docs

ipn/ipnlocal/local.go
```diff
@@ -4093,3 +4012,33 @@ func (b *LocalBackend) SetDevStateStore(key, value string) error {
...
+	b.setNetMapLocked(nil)                 // Reset netmap.
+	b.enterStateLockedOnEntry(ipn.NoState) // Reset state.
+	return b.Start(ipn.Options{})
+}
+
+func (b *LocalBackend) DeleteProfile(p ipn.ProfileID) error {
+	return b.pm.DeleteProfile(p)
+}
+
+func (b *LocalBackend) CurrentProfile() ipn.LoginProfile {
```
bradfitz, line 4034: can this return a zero value?

ipn/ipnlocal/local.go
```diff
@@ -4093,3 +4012,33 @@ func (b *LocalBackend) SetDevStateStore(key, value string) error {
...
+	if err := b.pm.SwitchProfile(profile); err != nil {
+		b.mu.Unlock()
+		return err
+	}
+	b.setNetMapLocked(nil)                 // Reset netmap.
+	b.enterStateLockedOnEntry(ipn.NoState) // Reset state.
+	return b.Start(ipn.Options{})
+}
+
+func (b *LocalBackend) DeleteProfile(p ipn.ProfileID) error {
```
bradfitz, line 4030: what's the error if the profile didn't exist?

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,455 @@
...
+			profiles = append(profiles, *p)
+		}
+	}
+	slices.SortFunc(profiles, func(a, b ipn.LoginProfile) bool {
+		return a.Name < b.Name
+	})
+	return profiles
+}
+
+// SwitchProfile switches to the profile with the given name.
```
bradfitz, line 214: document ErrProfileNotFound?

ipn/ipnlocal/profiles.go
```diff
@@ -0,0 +1,455 @@
...
+
+// CurrentProfile returns the name and ID of the current profile, or "" if the profile
+// is not named.
+func (pm *ProfileManager) CurrentProfile() ipn.LoginProfile {
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
+	return *pm.currentProfile
+}
+
+var ErrProfileNotFound = errors.New("profile not found")
```
bradfitz, line 277: // ErrProfileNotFound is returned by methods that accept a ProfileID.

ipn/ipnlocal/local.go
maisem: Made LocalBackend own and create profileManager (also unexported now)

ipn/ipnlocal/local.go
maisem: yeah, documented

ipn/ipnlocal/local.go
maisem: now with docs

PR #6031, opened 2022-10-23
ipn/*: add and use ipn.PrefsView

maisem:
Broken off from #6022.


types/views/views.go
```diff
@@ -201,6 +201,11 @@ func (v IPPrefixSlice) AsSlice() []netip.Prefix {
 	return v.ж.AsSlice()
 }
 
+// Filter returns a new slice, containing elements of in that match f.
```
bradfitz, line 204: no `in`

ipn/ipnlocal/local.go
```diff
@@ -874,35 +874,35 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
...
+	if !prefs.ExitNodeIP.IsValid() {
 		return false
 	}
 
 	// IP takes precedence over ID, so if both are set, clear ID.
-	if b.prefs.ExitNodeID != "" {
-		b.prefs.ExitNodeID = ""
+	if prefs.ExitNodeID != "" {
+		prefs.ExitNodeID = ""
 		prefsChanged = true
```
bradfitz, line 894: `prefsChanged = true` here but then "return false" below?!

ipn/ipnlocal/local.go
```diff
@@ -874,35 +874,35 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 	b.authReconfig()
 }
 
-// findExitNodeIDLocked updates b.prefs to reference an exit node by ID,
-// rather than by IP. It returns whether prefs was mutated.
-func (b *LocalBackend) findExitNodeIDLocked(nm *netmap.NetworkMap) (prefsChanged bool) {
+// findExitNodeIDLocked updates prefs to reference an exit node by ID, rather
+// than by IP. It returns whether prefs was mutated.
+func (b *LocalBackend) findExitNodeIDLocked(prefs *ipn.Prefs, nm *netmap.NetworkMap) (prefsChanged bool) {
```
mihaip, line 879: Nit: looks like this doesn't need to be a method on `LocalBackend` anymore, since you pass in everything it needs as arguments.

ipn/ipnlocal/local.go
maisem: good catch, its pre-existing though. Fixed

ipn/ipnlocal/local.go
maisem: Done

ipn/prefs.go
```diff
@@ -431,6 +441,10 @@ func NewPrefs() *Prefs {
 	}
 }
 
+func (p PrefsView) ControlURLOrDefault() string {
```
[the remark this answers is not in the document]
maisem: Done

bradfitz (approved):

mihaip (approved):

PR #6032, opened 2022-10-24
portlist: refactor, introduce OS-specific types

bradfitz:
Add an osImpl interface that can be stateful and thus more efficient between calls. It will later be implemented by all OSes but for now this change only adds a Linux implementation.

Remove Port.inode. It was only used by Linux and moves into its osImpl.

Don't reopen /proc/net/* files on each run. Turns out you can just keep then open and seek to the beginning and reread and the contents are fresh.

    name                    old time/op    new time/op    delta
    GetListIncremental-8    7.29ms ± 2%    6.53ms ± 1%  -10.50%  (p=0.000 n=9+9)

    name                   old alloc/op   new alloc/op   delta
    GetListIncremental-8    1.30kB ±13%    0.70kB ± 5%  -46.38%  (p=0.000 n=9+10)

    name                  old allocs/op  new allocs/op  delta
    GetListIncremental-8      33.2 ±11%      18.0 ± 0%  -45.82%  (p=0.000 n=9+10)

Updates #5958

### DIFF
--- a/portlist/portlist_linux.go
+++ b/portlist/portlist_linux.go
@@ -7,15 +7,15 @@ package portlist
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
-	"sync"
-	"sync/atomic"
 	"syscall"
 	"time"
 	"unsafe"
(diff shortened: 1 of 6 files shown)


portlist/portlist_linux.go
```diff
@@ -25,96 +25,133 @@ import (
...
+	v6Any       = "00000000000000000000000000000000:0000"
+	v4Localhost = "0100007F:"
+	v4Any       = "00000000:0000"
+)
+
+var eofReader = bytes.NewReader(nil)
 
-	for _, fname := range sockfiles {
+func (li *linuxImpl) AppendListeningPorts(base []Port) ([]Port, error) {
+	if runtime.GOOS == "android" {
```
DentonGentry, line 94: In the older code, Android 9 and earlier use this Linux implementation. I don't actually know what Android 10 and later have been doing here, do we know what the older Android releases will do when they run that Android code?

If not, we can figure it out the next time we update OSS in tailscale-android.

DentonGentry (approved):

portlist/portlist_linux.go
bradfitz: I'm basically giving up on all this code on Android. It's not worth maintaining for limited benefit (servers are rare on Android) and decreasing support (Android 10+ is only becoming more common over time.... in fact there are barely any Android 9 and earlier these days percentage-wise.

PR #6033, opened 2022-10-24
portlist: add a test that verifies changes are picked up over time

bradfitz:
To avoid annoying firewall dialogs on macOS and Windows, only run it on Linux by default without the flag.


bradfitz:
(the existing tests were a bit lacking for confidence working on the macOS & Windows impls)

DentonGentry (approved):

Issue #3876, opened 2022-02-04
Defender on Windows 11 appears to block tailscaled's repeated attempts to run `netstat.exe -na` as a "Risky action"


bgianfo:
### What is the issue?

I happened to be looking at the "Protection history" section in the `Windows Security` application on my windows 11 machine, and I saw the logs are littered with the following messages.

![image](https://user-images.githubusercontent.com/1212/152605179-7a6a6d02-847f-4e89-96aa-ed9a8a484fb9.png)

Curious what this was about, I used `ProcessExplorer` to watch the system and saw that `tailscaled.exe` appears to be the thing that is repeatedly spawning this process.
https://github.com/tailscale/tailscale/blob/297b3d6fa438c6953c5a032c7d6ae2f2c6bd4eb1/portlist/portlist_windows.go#L22

I haven't noticed this impacting any functionality, but I figured the team should know, and it would be better to get this fixed to avoid freaking folks out, like I did at first 😄. 

### Steps to reproduce

1) Install and Configure TailScale on Windows 11
2) Let the system run for a while.
3) `Start Menu` -> `Windows Security` -> `Virus & threat protection` -> `Protection history`
4) See the significant number of reports of `Risky Action blocked`.

### Are there any recent changes that introduced the issue?

Not that I'm aware of. 

### OS

Windows

### OS version

10.0.22000 Build 22000

### Tailscale version

1.20.2

### Bug report

BUG-3bfcd39348512a2918696599897f852e2e5ee06ce5b17b2a2271f47e09ca752f-20220205052437Z-223e882375f55e21

DentonGentry:
It should be possible to re-do this functionality using native Windows APIs.

bradfitz:
Native APIs are nicer than shelling out to things but that'd be a little stupid if doing the exact same thing with a different API is suddenly no longer "risky". Maybe they can adjust their risk engine instead? Maybe there's some escalation path to get blessed as not shady. (Wait, don't we sign all our binaries anyway? sigh)

PR #6034, opened 2022-10-24
portlist: use win32 calls instead of running netstat process [windows]

bradfitz:
Turns out using win32 instead of shelling out to child processes is a bit faster:

    name                  old time/op    new time/op    delta
    GetListIncremental-4     278ms ± 2%       0ms ± 7%  -99.93%  (p=0.000 n=8+10)

    name                  old alloc/op   new alloc/op   delta
    GetListIncremental-4     238kB ± 0%       9kB ± 0%  -96.12%  (p=0.000 n=10+8)

    name                  old allocs/op  new allocs/op  delta
    GetListIncremental-4     1.19k ± 0%     0.02k ± 0%  -98.49%  (p=0.000 n=10+10)


DentonGentry (approved):
Hurray that it also fixes https://github.com/tailscale/tailscale/issues/3876

portlist/portlist_windows.go
```diff
@@ -6,34 +6,124 @@ package portlist
...
+	if name == "." {
+		return ""
+	}
+	name = strings.TrimSuffix(name, ".exe")
+	name = strings.TrimSuffix(name, ".EXE")
+	return name
+}
+
+func appendListeningPorts([]Port) ([]Port, error) {
+	panic("unused on linux; needed to compile for now")
```
andrew-d, line 122: `linux`?

portlist/portlist_windows.go
```diff
@@ -6,34 +6,124 @@ package portlist
...
+}
+
+func addProcesses([]Port) ([]Port, error) {
+	panic("unused on linux; needed to compile for now")
```
andrew-d, line 126: `linux`?

bradfitz: whoops, copy/paste-o. fixed.

andrew-d (approved):

raggi (approved):

PR #6042, opened 2022-10-24
cmd,ipn/ipnlocal,tailcfg: implement TKA disablement

twitchyliquid64:
* Plumb disablement values through some of the internals of TKA enablement.
 * Transmit the node's TKA hash at the end of sync so the control plane understands each node's head.
 * Implement /machine/tka/disable RPC to actuate disablement on the control plane.
 * Send the TKA head hash in the map request. Changes to this value are communicated via sync-send; the map request is not restarted.

There is a partner PR for the control server I'll send shortly.

### DIFF
--- a/control/controlclient/auto.go
+++ b/control/controlclient/auto.go
@@ -561,6 +561,11 @@ func (c *Auto) SetNetInfo(ni *tailcfg.NetInfo) {
 	c.sendNewMapRequest()
 }
 
+// SetTKAHead updates the TKA head hash that map-request infrastructure sends.
+func (c *Auto) SetTKAHead(headHash string) {
+	c.direct.SetTKAHead(headHash)
+}
+
 func (c *Auto) sendStatus(who string, err error, url string, nm *netmap.NetworkMap) {
 	c.mu.Lock()
 	if c.closed {
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -94,6 +94,7 @@ type Direct struct {
 	hostinfo      *tailcfg.Hostinfo // always non-nil
 	netinfo       *tailcfg.NetInfo
 	endpoints     []tailcfg.Endpoint
+	tkaHead       string
 	everEndpoints bool   // whether we've ever had non-empty endpoints
 	lastPingURL   string // last PingRequest.URL received, for dup suppression
 }
@@ -314,6 +315,21 @@ func (c *Direct) SetNetInfo(ni *tailcfg.NetInfo) bool {
 	}
 	c.netinfo = ni.Clone()
 	c.logf("NetInfo: %v", ni)
+	return true
+}
+
+// SetNetInfo stores a new TKA head value for next update.
+// It reports whether the TKA head changed.
+func (c *Direct) SetTKAHead(tkaHead string) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if tkaHead == c.tkaHead {
+		return false
+	}
+
+	c.tkaHead = tkaHead
+	c.logf("tkaHead: %v", tkaHead)
 	return true
 }
 
@@ -829,6 +845,7 @@ func (c *Direct) sendMapRequest(ctx context.Context, maxPolls int, readOnly bool
 		Hostinfo:      hi,
 		DebugFlags:    c.debugFlags,
 		OmitPeers:     cb == nil,
+		TKAHead:       c.tkaHead,
 
 		// On initial startup before we know our endpoints, set the ReadOnly flag
 		// to tell the control server not to distribute out our (empty) endpoints to peers.
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -830,6 +830,16 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 			b.logf("[v1] TKA sync error: %v", err)
 		}
 		b.mu.Lock()
+		if b.tka != nil {
+			head, err := b.tka.authority.Head().MarshalText()
+			if err != nil {
+				b.logf("[v1] error marshalling tka head: %v", err)
+			} else {
+				b.cc.SetTKAHead(string(head))
+			}
+		} else {
+			b.cc.SetTKAHead("")
+		}
 
 		if !envknob.TKASkipSignatureCheck() {
 			b.tkaFilterNetmapLocked(st.NetMap)
@@ -1225,11 +1235,21 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 	b.cc = cc
 	b.ccAuto, _ = cc.(*controlclient.Auto)
 	endpoints := b.endpoints
+	var tkaHead string
+	if b.tka != nil {
+		head, err := b.tka.authority.Head().MarshalText()
+		if err != nil {
+			b.mu.Unlock()
+			return fmt.Errorf("marshalling tka head: %w", err)
+		}
+		tkaHead = string(head)
+	}
 	b.mu.Unlock()
 
 	if endpoints != nil {
 		cc.UpdateEndpoints(endpoints)
 	}
+	cc.SetTKAHead(tkaHead)
 
 	b.e.SetNetInfoCallback(b.setNetInfo)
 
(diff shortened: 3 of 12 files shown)


tailcfg/tailcfg.go
```diff
@@ -947,6 +947,11 @@ type MapRequest struct {
 	// EndpointTypes are the types of the corresponding endpoints in Endpoints.
 	EndpointTypes []EndpointType `json:",omitempty"`
 
+	// TKAHead describes the hash of the latest AUM applied to the local
+	// tailnet key authority, if one is operating.
+	// Head is encoded as tka.AUMHash.MarshalText.
```
bradfitz, line 952: // It is ...

ipn/ipnlocal/local.go
```diff
@@ -1221,11 +1231,22 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 	b.cc = cc
 	b.ccAuto, _ = cc.(*controlclient.Auto)
 	endpoints := b.endpoints
+	var tkaHead string
+	if b.tka != nil {
+		head, err := b.tka.authority.Head().MarshalText()
+		if err != nil {
+			return fmt.Errorf("marshalling tka head: %w", err)
```
bradfitz, line 1238: you're returning here with a lock held

ipn/ipnlocal/local.go
```diff
@@ -803,6 +803,16 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 			b.logf("[v1] TKA sync error: %v", err)
 		}
 		b.mu.Lock()
+		if b.tka != nil {
+			head, err := b.tka.authority.Head().MarshalText()
+			if err != nil {
+				b.logf("[v1] error marshalling tka head: %v", err)
+			} else {
+				b.cc.SetTKAHead(string(head))
```
bradfitz, line 811: should you be calling these with a lock held?

do those acquire locks themselves?

control/controlclient/direct.go
```diff
@@ -317,6 +318,21 @@ func (c *Direct) SetNetInfo(ni *tailcfg.NetInfo) bool {
 	return true
 }
 
+// SetNetInfo stores a new tka head value for next update.
+// True is returned if the tka head changed.
```
bradfitz, line 322: // It reports whether the TKA head changed.

(use TKA in docs as it's an acronym)

control/controlclient/auto.go
```diff
@@ -561,6 +561,17 @@ func (c *Auto) SetNetInfo(ni *tailcfg.NetInfo) {
...
 
+// SetTKAHead updates the TKA head hash that map-request infrastructure sends.
+// If the value has changed, a new map-request is queued.
+func (c *Auto) SetTKAHead(headHash string) {
+	if !c.direct.SetTKAHead(headHash) {
+		return
+	}
+
+	// Send new head hash to server
+	c.sendNewMapRequest()
```
bradfitz, line 572: we're trying to get away from this, restart map requests on changes. You send a new discrete HTTP request instead about what you want to say.

twitchyliquid64: We need this information in computing the map response for disablement.

Are there other places we tx information to control in a discrete RPC in the manner you're describing?

ipn/ipnlocal/local.go
twitchyliquid64: Control client has its own independent locking, which is acquired and dropped as part of the `SetTKAHead` call.

I can see this being a problem if ControlClient holds its own lock while it does a `setClientStatus` call, otherwise it should be fine?

ipn/ipnlocal/local.go
twitchyliquid64: Done.

ipn/ipnlocal/local.go
twitchyliquid64: Afaict theres no lock held during the callback:

```go
c.mu.Lock()
c.expiry = &nm.Expiry
c.mu.Unlock()

cb(nm)
```

control/controlclient/auto.go
bradfitz: Existing ones are:

* the "lite endpoint updates" which are map requests in name only but are effectively separate RPCs (and don't break the map poll)
* set-dns
* update-health

twitchyliquid64: Closing this out: The sync-send RPC (which has to be sent anyway) carries this information and on the control site gets wired into the node wakeup mechanics, hence avoiding the need to restart the map poll.

ipn/ipnlocal/network-lock.go
```diff
@@ -125,15 +127,10 @@ func (b *LocalBackend) tkaSyncIfNeeded(nm *netmap.NetworkMap) error {
...
 		} else if !wantEnabled && isEnabled {
-			if b.tka.authority.ValidDisablement(bs.DisablementSecret) {
-				b.tka = nil
-				isEnabled = false
-
-				if err := os.RemoveAll(b.chonkPath()); err != nil {
-					return fmt.Errorf("os.RemoveAll: %v", err)
-				}
+			if err := b.tkaApplyDisablementLocked(bs.DisablementSecret); err != nil {
+				b.logf("Disablement failed, leaving TKA enabled. Error: %v", err)
```
sailorfrag, line 131: Should this be a returned error rather than just logged?

ipn/ipnlocal/network-lock.go
```diff
@@ -474,6 +489,39 @@ func (b *LocalBackend) NetworkLockModify(addKeys, removeKeys []tka.Key) (err err
...
+	if err := b.CanSupportNetworkLock(); err != nil {
+		return err
+	}
+
+	var (
+		ourNodeKey key.NodePublic
+		head       tka.AUMHash
+	)
+
+	b.mu.Lock()
```
sailorfrag, line 503: optional: Move this into a separate function so you can defer the unlock? Alternatively, perhaps set an `err` to the error and return if non-nil after the unlock.

ipn/ipnlocal/network-lock.go
```diff
@@ -715,3 +769,45 @@ func (b *LocalBackend) tkaDoSyncSend(ourNodeKey key.NodePublic, aums []tka.AUM,
...
+	if err != nil {
+		return nil, fmt.Errorf("resp: %w", err)
+	}
+	if res.StatusCode != 200 {
+		body, _ := io.ReadAll(res.Body)
+		res.Body.Close()
+		return nil, fmt.Errorf("request returned (%d): %s", res.StatusCode, string(body))
+	}
+	a := new(tailcfg.TKADisableResponse)
+	err = json.NewDecoder(res.Body).Decode(a)
```
sailorfrag, line 806: On the server side, there's a defensive `LimitReader`, should we be doing that on the client side too?

sailorfrag (approved):

ipn/ipnlocal/network-lock.go
twitchyliquid64: I've opted to log here because returning early would mean that sync is not completed. I know holdbacks are technically out of scope, but the less the better. I've added a comment.

ipn/ipnlocal/network-lock.go
twitchyliquid64: Done the latter.

ipn/ipnlocal/network-lock.go
```diff
@@ -238,6 +234,21 @@ func (b *LocalBackend) tkaSyncLocked(ourNodeKey key.NodePublic) error {
 	return nil
 }
 
+// tkaApplyDisablementLocked checks a disablement secret and locally disables
+// TKA (if correct). An error is returned if disablement failed.
+//
+// b.mu must be held & TKA must be initialized.
+func (b *LocalBackend) tkaApplyDisablementLocked(secret []byte) error {
+	if b.tka.authority.ValidDisablement(secret) {
```
bradfitz, line 242: can `b.tka` be nil here?

ipn/ipnlocal/network-lock.go
twitchyliquid64: Done, here and other fetch methods.

ipn/ipnlocal/network-lock.go
twitchyliquid64: No, that codepath is guarded by `isEnabled` which itself is `b.tka != nil`

Issue #5142, opened 2022-07-22
cmd/tsconnect: Look into wasm_js.go binary size


mihaip:
Can do another pass of examining the wasm output with https://evmar.github.io/weave/. 

Can also try to run https://github.com/bradfitz/shotizam on it.

mihaip:
https://leaningtech.com/webvm-virtual-machine-with-networking-via-tailscale/#f268 mentions that running `wasm-opt` shaved off about 1MB. https://github.com/WebAssembly/binaryen#binaryen-optimizations has more details.

PR #6045, opened 2022-10-24
cmd/tsconnect: run wasm-opt on the generated wasm file

mihaip:
Saves about 1.4MB from the generated wasm file. The Brotli size is basically unchanged (it's actually slightly larger, by 40K), suggesting that most of the size delta is due to not inlining and other changes that were easily compressible.

However, it still seems worthwhile to have a smaller final binary, to reduce parse time and increase likelihood that we fit in the browser's disk cache. Actual performance appears to be unchanged.

Updates #5142


bradfitz (approved):
LGTM with the big caveat that you're asserting that you're confident this won't break the Go-generated wasm binary by wasmopt assuming that its output is from llvm or other C/Rust compilers.

I remember the UPX shrinker tool choking on Go binaries for years. Or rather, it would claim to work but then produce binaries that crashed at runtime.

mihaip:
I tried it locally and it seems to work. I found a few references to blog posts using it with Go binaries (and conversely, no issues in either https://github.com/golang/go or https://github.com/WebAssembly/binaryen/ about problems). We can revert if we start to get mystery crashes.

Issue #5972, opened 2022-10-17
control/controlhttp: associate node key(s) to noise conns


bradfitz:
Tracking bug for associating node keys to noise connections which will be used for future features.

@danderson will update this bug with a doc/design later. Just grabbing a bug number for now.

/cc @maisem @skriptble

danderson:
Background + design: https://docs.google.com/document/d/1KpJ2VSbX6VTWg3kVQcgUCANuRxybqM5CA8dWiydTR8o/edit#

PR #6052, opened 2022-10-25
client/tailscale, tsnet, ipn/ipnlocal: prove nodekey ownership over noise

bradfitz:
Updates #5972

### DIFF
--- a/client/tailscale/tailscale.go
+++ b/client/tailscale/tailscale.go
@@ -18,9 +18,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"net/url"
-
-	"tailscale.com/types/key"
 )
 
 // I_Acknowledge_This_API_Is_Unstable must be set true to use this package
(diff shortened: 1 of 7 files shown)


bradfitz:
Just started on this, remembering the code. What's done compiles, but it's not done yet.

client/tailscale/tailscale.go
```diff
@@ -93,25 +90,11 @@ func (c *Client) setAuth(r *http.Request) {
...
-	req.SetBasicAuth(nk, "")
-}
-
-// NewNoiseClient is a convenience method for instantiating a new Client
-// that uses the Noise protocol for authentication.
-//
-// tailnet is the globally unique identifier for a Tailscale network, such
-// as "example.com" or "user@gmail.com".
-func NewNoiseClient(tailnet string, noiseRoundTripper http.RoundTripper, nk key.NodePublic) *Client {
+// NewTSNetNoiseClient returns a  new Client that uses the Noise protocol for
```
maisem, line 93: two spaces in `returns a  new`

client/tailscale/tailscale.go
```diff
@@ -93,25 +90,11 @@ func (c *Client) setAuth(r *http.Request) {
...
+// authentication for use within a tsnet binary.
+func NewTSNetNoiseClient(noiseRoundTripper http.RoundTripper) *Client {
```
maisem, line 95: why is this TSNet specific?

control/controlclient/auto.go
```diff
@@ -730,3 +730,13 @@ func (c *Auto) SetDNS(ctx context.Context, req *tailcfg.SetDNSRequest) error {
 func (c *Auto) DoNoiseRequest(req *http.Request) (*http.Response, error) {
 	return c.direct.DoNoiseRequest(req)
 }
+
+// GetSingleUseNoiseRoundTripper returns a RoundTripper that can be only be used
+// once (and must be used once) to make a single HTTP request over the noise
+// channel to the coordination server.
+//
+// In addition to the RoundTripper, it returns the HTTP/2 channel's early noise
+// payload, if any.
```
maisem, line 739: add this to Direct too?

maisem (approved):

client/tailscale/tailscale.go
bradfitz: Because it requires you to be running in-process in tailscaled to be able to get a suitable RoundTripper from LocalBackend.

This doesn't yet work from another process (like cmd/tailscale CLI) because we never proxy noise stuff through tailscaled where it can grab a stateful noise conn to sign.

I ripped out all that "modify the *http.Request as an auth method" stuff which doesn't support stateful connections (you can't sign a request to a specific connection if you don't have the connection yet)

maisem: oh right, I forgot 🤦🏼‍♂️

maisem: maybe call it something else though? so that we don't have to rename later.

PR #6062, opened 2022-10-25
ipn/ipnlocal: set prefs before calling tkaSyncIfNeeded

maisem:
Caught this in a test in a different repo.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -803,21 +803,6 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 			prefs.Persist = st.Persist.Clone()
 		}
 	}
-	if st.NetMap != nil {
-		b.mu.Unlock() // respect locking rules for tkaSyncIfNeeded
-		if err := b.tkaSyncIfNeeded(st.NetMap); err != nil {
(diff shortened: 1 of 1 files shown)


twitchyliquid64 (approved):

ipn/ipnlocal/local.go
```diff
@@ -832,11 +817,24 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
...
 	if prefsChanged {
 		b.prefs = prefs.View()
 	}
 	if st.NetMap != nil {
+		b.mu.Unlock() // respect locking rules for tkaSyncIfNeeded
+		if err := b.tkaSyncIfNeeded(st.NetMap); err != nil {
+			b.logf("[v1] TKA sync error: %v", err)
+		}
+		b.mu.Lock()
+
```
bradfitz, line 833: cool, which invariants might have been invalidated now and needed to be checked again now that we've reacquired the lock? 😬

bradfitz: oh, this is just code movement

maisem: Yeah, but yes we should revisit this and see if we can make the lock ordering better.

PR #6068, opened 2022-10-26
cmd/tailscale/cli: add beginnings of `tailscale set`

maisem:

### DIFF
--- a/cmd/tailscale/cli/set.go
+++ b/cmd/tailscale/cli/set.go
@@ -1,0 +1,131 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package cli
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+
+	"github.com/peterbourgon/ff/v3/ffcli"
+	"tailscale.com/ipn"
+	"tailscale.com/safesocket"
+)
+
+var setCmd = &ffcli.Command{
+	Name:       "set",
+	ShortUsage: "set [flags]",
+	ShortHelp:  "Change specified preferences",
+	LongHelp: `"tailscale set" allows changing specific preferences.
+
+Unlike "tailscale up", this command does not require the complete set of desired settings.
+
+Only settings explicitly mentioned will be set. There are no default values.`,
+	FlagSet:   setFlagSet,
+	Exec:      runSet,
+	UsageFunc: usageFuncNoDefaultValues,
+}
+
+type setArgsT struct {
+	acceptRoutes           bool
+	acceptDNS              bool
+	exitNodeIP             string
+	exitNodeAllowLANAccess bool
+	shieldsUp              bool
+	runSSH                 bool
+	hostname               string
+	advertiseRoutes        string
+	advertiseDefaultRoute  bool
+	opUser                 string
+	acceptedRisks          string
+}
+
+func newSetFlagSet(goos string, setArgs *setArgsT) *flag.FlagSet {
+	setf := newFlagSet("set")
+
+	setf.BoolVar(&setArgs.acceptRoutes, "accept-routes", false, "accept routes advertised by other Tailscale nodes")
+	setf.BoolVar(&setArgs.acceptDNS, "accept-dns", false, "accept DNS configuration from the admin panel")
+	setf.StringVar(&setArgs.exitNodeIP, "exit-node", "", "Tailscale exit node (IP or base name) for internet traffic, or empty string to not use an exit node")
+	setf.BoolVar(&setArgs.exitNodeAllowLANAccess, "exit-node-allow-lan-access", false, "Allow direct access to the local network when routing traffic via an exit node")
+	setf.BoolVar(&setArgs.shieldsUp, "shields-up", false, "don't allow incoming connections")
+	setf.BoolVar(&setArgs.runSSH, "ssh", false, "run an SSH server, permitting access per tailnet admin's declared policy")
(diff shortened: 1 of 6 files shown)


cmd/tailscale/cli/set.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/peterbourgon/ff/v3/ffcli"
+	"tailscale.com/ipn"
+	"tailscale.com/safesocket"
+)
+
+var setCmd = &ffcli.Command{
+	Name:       "set",
+	ShortUsage: "set [flags]",
+	ShortHelp:  "Toggle preferences",
```
bradfitz, line 21: I think "toggle" implies changing a boolean on/off?

"Change specified preferences"

And then in LongHelp you can say more about how it differs from "up".

ipn/prefs.go
```diff
@@ -244,6 +244,17 @@ func (p *Prefs) ApplyEdits(m *MaskedPrefs) {
 	}
 }
 
+// IsEmpty reports true when there are no masks set or if m is nil.
```
bradfitz, line 247: reports whether

ipn/prefs.go
```diff
@@ -244,6 +244,17 @@ func (p *Prefs) ApplyEdits(m *MaskedPrefs) {
...
+func (m *MaskedPrefs) IsEmpty() bool {
+	if m == nil {
+		return true
+	}
+	m2 := *m
+	m2.Prefs = Prefs{}
+	var empty MaskedPrefs
+	return reflect.DeepEqual(m2, empty)
```
bradfitz, line 255: I think it'd be cheaper to do this with a reflect loop like Prefs.ApplyEdits instead of using DeepEqual (which allocates here)

bradfitz (approved):
Not sure why this is a draft. Looks done to me?

cmd/tailscale/cli/set.go
maisem: Updated, ptal.

cmd/tailscale/cli/set.go
```diff
@@ -0,0 +1,116 @@
...
+	setf.BoolVar(&setArgs.exitNodeAllowLANAccess, "exit-node-allow-lan-access", false, "Allow direct access to the local network when routing traffic via an exit node")
+	setf.BoolVar(&setArgs.shieldsUp, "shields-up", false, "don't allow incoming connections")
+	setf.BoolVar(&setArgs.runSSH, "ssh", false, "run an SSH server, permitting access per tailnet admin's declared policy")
+	setf.StringVar(&setArgs.hostname, "hostname", "", "hostname to use instead of the one provided by the OS")
+	setf.StringVar(&setArgs.advertiseRoutes, "advertise-routes", "", "routes to advertise to other nodes (comma-separated, e.g. \"10.0.0.0/8,192.168.0.0/24\") or empty string to not advertise routes")
+	setf.BoolVar(&setArgs.advertiseDefaultRoute, "advertise-exit-node", false, "offer to be an exit node for internet traffic for the tailnet")
+	if safesocket.GOOSUsesPeerCreds(goos) {
+		setf.StringVar(&setArgs.opUser, "operator", "", "Unix username to allow to operate on tailscaled without sudo")
+	}
+	registerAcceptRiskFlag(setf, &setArgs.acceptedRisks)
```
bradfitz, line 58: What is this used by? It gets ignored by `updateMaskedPrefsFromUpOrSetFlag` because of `if preflessFlag(flagName)`.

So what then uses it? Nothing?

cmd/tailscale/cli/set.go
```diff
@@ -0,0 +1,116 @@
...
+	"tailscale.com/safesocket"
+)
+
+var setCmd = &ffcli.Command{
+	Name:       "set",
+	ShortUsage: "set [flags]",
+	ShortHelp:  "Change specified preferences",
+	LongHelp: `"tailscale set" allows changing specific preferences.
+
+Unlike "tailscale up", this command does not require the complete set of desired settings.`,
```
bradfitz, line 24: Add a line:

Only settings explicitly mentioned will be set. There are no default values.

cmd/tailscale/cli/set.go
```diff
@@ -0,0 +1,116 @@
...
+	advertiseRoutes        string
+	advertiseDefaultRoute  bool
+	opUser                 string
+	acceptedRisks          string
+}
+
+func newSetFlagSet(goos string, setArgs *setArgsT) *flag.FlagSet {
+	setf := newFlagSet("set")
+
+	setf.BoolVar(&setArgs.acceptRoutes, "accept-routes", false, "accept routes advertised by other Tailscale nodes")
```
bradfitz, line 46: can we make our flag printer omit default values for all of these somehow? 

see the Go `flag` packages docs. In particular this part:

> The usage message will appear on a separate line for anything but a bool flag with a one-byte name. For bool flags, the type is omitted and if the flag name is one byte the usage message appears on the same line. The parenthetical default is omitted if the default is the zero value for the type. The listed type, here int, can be changed by placing a back-quoted name in the flag's usage string; the first such item in the message is taken to be a parameter name to show in the message and the back quotes are stripped from the message when displayed. For instance, given

bradfitz (approved):

cmd/tailscale/cli/set.go
maisem: Using it now for `lose-ssh`

cmd/tailscale/cli/set.go
maisem: Done.

bradfitz (commented):
Isn't turning ssh off when you're using it also a warning? I forget.

maisem:
>Isn't turning ssh off when you're using it also a warning? I forget.

Yes it is. It is the same behavior between `up` and `set`

bradfitz (approved):

PR #6069, opened 2022-10-26
wgengine/magicsock: add pathfinder file for alternate disco path

phirework:
Adds Pathfinder file for alternate discovery path with heartbeat turned off. Currently replicates largely existing behaviour and will not have any functional impact with the envknob turned off. Also changes heartbeat envknob from be settable after program startup.

Updates #540

### DIFF
--- a/wgengine/magicsock/pathfinder.go
+++ b/wgengine/magicsock/pathfinder.go
@@ -4,10 +4,97 @@
 
 package magicsock
 
-// startPathFinder initializes the atomicSendFunc, and
+import (
+	"fmt"
+	"net/netip"
+	"time"
+
+	"tailscale.com/tstime/mono"
+	"tailscale.com/types/key"
+)
+
+// startPathFinder initializes the sendFunc, and
 // will eventually kick off a goroutine that monitors whether
 // that sendFunc is still the best option for the endpoint
 // to use and adjusts accordingly.
 func (de *endpoint) startPathFinder() {
-	de.pathFinderRunning = true
+	go func() {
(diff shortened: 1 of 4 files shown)


wgengine/magicsock/pathfinder.go
```diff
@@ -4,10 +4,101 @@
...
+
+// updateSendPathIfNecessary optionally upates sendFunc
+// based on analysis of current conditions
+func (de *endpoint) updateSendPathIfNecessary(now mono.Time) {
+	de.mu.Lock()
+	defer de.mu.Unlock()
+
+	// default happy state is: use UDP, don't use Derp
+	useUDP := true
+	useDerp := false
```
phirework, line 44: This feels not-idiomatic, but it also seemed like the cleanest way to avoid a bunch of nested `.IsValid()` checks. Open to suggestions!

wgengine/magicsock/pathfinder.go
```diff
@@ -4,10 +4,101 @@
...
+		return
+	}
+
+	if useUDP && useDerp {
+		de.sendFunc.Store(de.sendDerpAndUDP(udpAddr, derpAddr, de.publicKey))
+	} else if useUDP {
+		de.sendSinglePath(udpAddr, de.publicKey)
+	} else if useDerp {
+		de.sendSinglePath(derpAddr, de.publicKey)
+	}
```
bradfitz, line 76: these return a func and then you're doing nothing with the result. Did you mean to Store it in the sendFunc?

phirework: I did, fixed.

PR #6078, opened 2022-10-26
docs/webhooks: add sample endpoint code

soniaappasamy:
Signed-off-by: Sonia Appasamy <sonia@tailscale.com>

### DIFF
--- a/docs/webhooks/example.go
+++ b/docs/webhooks/example.go
@@ -1,0 +1,149 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package webhooks provides example consumer code for Tailscale
+// webhooks.
+package webhooks
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type event struct {
+	Timestamp string            `json:"timestamp"`
+	Version   int               `json:"version"`
+	Type      string            `json:"type"`
+	Tailnet   string            `json:"tailnet"`
+	Message   string            `json:"message"`
+	Data      map[string]string `json:"data"`
+}
+
+const (
+	currentVersion = "v1"
+	secret         = "tskey-webhook-xxxxx" // sensitive, here just as an example
+)
+
+var (
+	errNotSigned     = errors.New("webhook has no signature")
+	errInvalidHeader = errors.New("webhook has an invalid signature")
+)
+
+func main() {
+	http.HandleFunc("/webhook", webhooksHandler)
+	if err := http.ListenAndServe(":80", nil); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func webhooksHandler(w http.ResponseWriter, req *http.Request) {
+	defer req.Body.Close()
+	events, err := verifyWebhookSignature(req, secret)
+	if err != nil {
+		log.Printf("error validating signature: %v\n", err)
+	} else {
+		log.Printf("events received %v\n", events)
+		// Do something with your events. :)
+	}
+
+	// The handler should always report 2XX except in the case of
+	// transient failures (e.g. database backend is down).
+	// Otherwise your future events will be blocked by retries.
+}
+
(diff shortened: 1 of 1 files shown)


docs/webhooks/example.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+var (
+	errNotSigned     = errors.New("webhook has no signature")
+	errInvalidHeader = errors.New("webhook has an invalid signature")
+)
+
+func main() {
+	http.HandleFunc("/webhook", webhooksHandler)
+	err := http.ListenAndServe(":80", nil)
```
dsnet, line 45: Might as well combine this statement on the if-statement.
```go
if err := ...; err != nil {
    ...
}
```

docs/webhooks/example.go
```diff
@@ -0,0 +1,138 @@
...
+
+		switch parts[0] {
+		case "t":
+			tsint, err := strconv.ParseInt(parts[1], 10, 64)
+			if err != nil {
+				return time.Time{}, nil, errInvalidHeader
+			}
+			timestamp = time.Unix(tsint, 0)
+		case currentVersion:
+			signatures[parts[0]] = parts[1]
```
andrew-d, line 127: This doesn't handle the case where there are multiple signatures of the same type... though I guess it also doesn't break and just takes the last one specified in the header. I don't know how much we want to optimize for, but perhaps a `map[string][]string` and then verifying in a loop above (on line 93)?

docs/webhooks/example.go
```diff
@@ -0,0 +1,138 @@
...
+	// If verified, return the events.
+	if err := json.Unmarshal(b, &events); err != nil {
+		return nil, err
+	}
+	return events, nil
+}
+
+// parseSignatureHeader splits header into its timestamp and included signatures.
+// The signatures are reported as a map of version (e.g. "v1") to signature content.
+func parseSignatureHeader(header string) (timestamp time.Time, signatures map[string]string, err error) {
```
dsnet, line 106: This allocates a lot, but looks correct. I guess that's okay for demo code.

docs/webhooks/example.go
soniaappasamy: Would we have multiple of the same type? Oh for rotation, they would both still be v1? And v2 only comes if there's a new secret format?

soniaappasamy: Nice, ok updated!

soniaappasamy:
I _think_ the github-code-scanning errors are ok to ignore?

docs/webhooks/example.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+}
+
+func webhooksHandler(w http.ResponseWriter, req *http.Request) {
+	defer req.Body.Close()
+	events, err := verifyWebhookSignature(req, secret)
+	if err != nil {
+		log.Printf("error validating signature: %v\n", err)
+	} else {
+		log.Printf("events received %v\n", events)
```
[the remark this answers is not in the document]
dsnet: This obviously leaks user input, but that kinda the point of this example. Safe to ignore

docs/webhooks/example.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func webhooksHandler(w http.ResponseWriter, req *http.Request) {
+	defer req.Body.Close()
+	events, err := verifyWebhookSignature(req, secret)
+	if err != nil {
+		log.Printf("error validating signature: %v\n", err)
```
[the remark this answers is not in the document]
dsnet: This is a false positive. The static analysis correctly detects that some parts of the input leak to the log statement, but fails to account for the fact that it is escaped. For example, we use the "%q" verb on L104.

dsnet (approved):

docs/webhooks/example.go
```diff
@@ -0,0 +1,149 @@
...
+	mac := hmac.New(sha256.New, []byte(secret))
+	mac.Write([]byte(fmt.Sprint(timestamp.Unix())))
+	mac.Write([]byte("."))
+	mac.Write(b)
+	want := hex.EncodeToString(mac.Sum(nil))
+
+	// Verify that the signatures match.
+	var match bool
+	for _, signature := range signatures[currentVersion] {
+		if signature == want {
```
andrew-d, line 98: Almost certainly not a real issue, but as a best-practice, we should be using [`subtle.ConstantTimeCompare`](https://pkg.go.dev/crypto/subtle#ConstantTimeCompare) here.

Issue #6037, opened 2022-10-24
Tailscale cannot connect on Windows Startup, probably DNS issue


VictorTomaili:
### What is the issue?

A few months ago I was not had such as issue but last few months when I start my computer, the tailscale is not connecting to the network. Most of the case, it needs to restart the tailscale service to fix that. And these days, It's not enough to restart service. I need to replace my DNS as 1.1.1.1 to fix that.  After a few seconds, tailscale is connecting to the network and I revert this DNS to automatic. I use tailscale DNS settings enabled. 

### Steps to reproduce

When I start my computer each day, with same or different internet connection. 
With my desktop pc and my laptop. Both of them same operating system and latest OS/tailscale version. 

### OS

Windows

### OS version

11 22H2 OSBuilld 22623.746

### Tailscale version

1.32.1

### Bug report

BUG-e5b33479a1a688f555a1731368bbc2414dc14001558a7c5f1ef75c8b37ce7e6b-20221024085038Z-3c0a94c24d1add09

andrew-d:
I believe that this is caused by #4134, since the logs show:
```
setPrivateNetwork(try=1): GetNetwork: Element not found.
setPrivateNetwork(try=2): GetNetwork: Element not found.
[...]
user bugreport health: network-category: GetNetwork: Element not found.
```

DentonGentry:
Duplicate of https://github.com/tailscale/tailscale/issues/4134

Issue #4134, opened 2022-03-10
Windows: setPrivateNetwork error 


JayWStapleton:
### What is the issue?

Seeing inconsistent connectivity, accompanied by log lines:
```setPrivateNetwork(try=13): SetCategory: Exception occurred. (<nil>)   ```

Also associated with 
```dns: error: context deadline exceeded ``` when trying to access a DNS server over a subnet router on the tailnet. 

note: `shields=up `

### OS

Windows

### OS version

10 osver="10.0.19044.1566"

### Tailscale version

1.22.0

### Bug report

BUG-4dc17f24c356fc1ac1feb9707f7f32306160a8b001a30853db7e46719c88b095-20220310165728Z-6fdcb5de091abe84

dblohm7:
That error message is a `DISP_E_EXCEPTION` error returned by [`IDispatch::Invoke`](https://docs.microsoft.com/en-us/windows/win32/api/oaidl/nf-oaidl-idispatch-invoke). I'm reviewing how the go-ole package is dealing with this and will follow up in a bit.

dblohm7:
Unfortunately there is a [bug in `go-ole`](https://github.com/go-ole/go-ole/blob/ee6c8cce8e87872a00da9b639db11e6ba8cc5f56/idispatch_windows.go#L190) where exception info is obtained but is being cleared before it is transferred into a go `error`. In other words, there actually was more useful information than `<nil>`, but it was prematurely cleared.

As a starting point, we could at least deliver a test build with proper handling of that information in the hope that we would then be able to obtain better logs.

dblohm7:
Correction, I'm reading this completely wrong because the `EXCEPINFO` implementation in `go-ole` is extremely confusing. eg it exposes a `Clear` method whose purpose is to free OLE `BSTR`s, but doesn't clear the entire structure.

I'll need to look more closely.

dblohm7:
The way that `go-ole` handles errors from `IDispatch` method invocations results in suppression of important error information. In the short term I can fork that repo and clean that up so that we can at least give them a build that hopefully provides a more descriptive error.

DentonGentry:
If anyone produces this error with 1.23.x from https://pkgs.tailscale.com/unstable/ or with 1.24 or later, please update this bug. There will be more information available for troubleshooting.

bradfitz:
[...]

PR #6080, opened 2022-10-26
wgengine/winnet: invoke some COM methods directly instead of through …

dblohm7:
…IDispatch.

Intermittently in the wild we are seeing failures when calling `INetworkConnection::GetNetwork`. It is unclear what the root cause is, but what is clear is that the error is happening inside the object's `IDispatch` invoker (as opposed to the method implementation itself).

This patch replaces our wrapper for `INetworkConnection::GetNetwork` with an alternate implementation that directly invokes the method, instead of using `IDispatch`. I also replaced the implementations of `INetwork::SetCategory` and `INetwork::GetCategory` while I was there.

This patch is speculative and tightly-scoped so that we could possibly add it to a dot-release if necessary.

Updates https://github.com/tailscale/tailscale/issues/6037

### DIFF
--- a/wgengine/winnet/winnet.go
+++ b/wgengine/winnet/winnet.go
@@ -9,6 +9,7 @@ package winnet
 
 import (
 	"fmt"
+	"syscall"
 	"unsafe"
 
 	"github.com/go-ole/go-ole"
@@ -43,6 +44,23 @@ type INetworkConnectionVtbl struct {
 
 type INetwork struct {
 	ole.IDispatch
+}
+
+type INetworkVtbl struct {
+	ole.IDispatchVtbl
+	GetName                    uintptr
+	SetName                    uintptr
+	GetDescription             uintptr
+	SetDescription             uintptr
+	GetNetworkId               uintptr
+	GetDomainType              uintptr
+	GetNetworkConnections      uintptr
+	GetTimeCreatedAndConnected uintptr
+	Get_IsConnectedToInternet  uintptr
+	Get_IsConnected            uintptr
+	GetConnectivity            uintptr
+	GetCategory                uintptr
+	SetCategory                uintptr
 }
 
 func NewNetworkListManager(c *ole.Connection) (*NetworkListManager, error) {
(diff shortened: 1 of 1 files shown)


bradfitz:
Updates #4134 too?

wgengine/winnet/winnet.go
```diff
@@ -45,6 +46,23 @@ type INetwork struct {
...
+	GetNetworkId               uintptr
+	GetDomainType              uintptr
+	GetNetworkConnections      uintptr
+	GetTimeCreatedAndConnected uintptr
+	Get_IsConnectedToInternet  uintptr
+	Get_IsConnected            uintptr
+	GetConnectivity            uintptr
+	GetCategory                uintptr
+	SetCategory                uintptr
+}
```
andrew-d, line 64: This appears to match the vtable that Microsoft generates for their Rust API bindings (my go-to for finding the structure for things without needing to dig through C++):
https://github.com/microsoft/windows-rs/blob/962d0a9f19d049a37f704b8ea3c7ab7efd1d95cb/crates/libs/windows/src/Windows/Win32/Networking/NetworkListManager/mod.rs#L248-L269

andrew-d (commented):
This is the first time that I'm looking at our COM code; super neat stuff 👍

dblohm7:
> Updates #4134 too?

Yeah. Fixed.

dblohm7:
> This is the first time that I'm looking at our COM code; super neat stuff 👍

TBH I despise the library we're using for this and am working on something much better.

raggi:
I made a basic debugging test, and I'm still seeing the same behavior with the patch:

```
func TestBasic(t *testing.T) {

	runtime.LockOSThread()
	defer runtime.UnlockOSThread()

	var c ole.Connection
	if err := c.Initialize(); err != nil {
		t.Fatal(err)
	}
	defer c.Uninitialize()

	m, err := NewNetworkListManager(&c)
	if err != nil {
		t.Fatal(err)
	}
	defer m.Release()

	cl, err := m.GetNetworkConnections()
	if err != nil {
		t.Fatal(err)
	}
	defer cl.Release()

	t.Logf("got %d connections", len(cl))
	for _, c := range cl {
		conn, err := c.GetNetwork()
		if err != nil {
			t.Errorf("GetNetwork: %v", err)
			continue
		}
		name, err := conn.GetName()
		if err != nil {
			t.Errorf("GetName: %v", err)
			continue
		} else {
			t.Logf("name: %s", name)
		}
		id, err := c.GetAdapterId()
		if err != nil {
			t.Fatal(err)
		}
		t.Logf("id: %s", id)
	}

}
```

```
=== RUN   TestBasic
    wininet_test.go:36: got 2 connections
    wininet_test.go:45: GetName: Element not found.
    wininet_test.go:45: GetName: Element not found.
--- FAIL: TestBasic (0.02s)
FAIL
FAIL    tailscale.com/wgengine/winnet   0.278s
FAIL
```

dblohm7:
As discussed with James, the same code does not repro on my machine. We're not calling `GetName` in this code path, so I'd like to land the vtable stuff for now so that we can give support an unstable build.

raggi (approved):
The code here is fine, and works in principle - following the pattern of other usage of the lib (e.g. the bluetooth library that uses it). It still doesn't seem to work on my Windows 11, but I think it's a step in the right direction anyway.

PR #6081, opened 2022-10-26
portlist, util/winutil: change Windows portlist implementation to dis…

dblohm7:
…ambiguate svchost processes.

We query the Windows Service Control Manager for information about service processes so that we can map services hosted by `svchost.exe` to the real name of the service being hosted. Unfortunately we need to query the process names of all running services to determine which ones are `svchost` processes, but we also temporarily cache them, which should help avoid some redundant system calls. I think this extra work is worth it, as so many system services are hosted in `svchost` that we would otherwise lose a lot of information compared to the previous implementation that shelled out to `netstat`.

This gives us similar fidelity to what we were getting when we were shelling out to `netstat`, with the added bonus that this implementation appears to be able to produce slightly better quality data.

### DIFF
--- a/portlist/portlist_windows.go
+++ b/portlist/portlist_windows.go
@@ -5,12 +5,8 @@
 package portlist
 
 import (
-	"path/filepath"
-	"strings"
-	"syscall"
 	"time"
 
-	"golang.org/x/sys/windows"
 	"tailscale.com/net/netstat"
 )
 
@@ -26,7 +22,7 @@ func init() {
 type famPort struct {
 	proto string
 	port  uint16
-	pid   uintptr
+	pid   uint32
 }
 
 type windowsImpl struct {
@@ -69,19 +65,25 @@ func (im *windowsImpl) AppendListeningPorts(base []Port) ([]Port, error) {
 		fp := famPort{
 			proto: "tcp", // TODO(bradfitz): UDP too; add to netstat
 			port:  e.Local.Port(),
-			pid:   uintptr(e.Pid),
+			pid:   uint32(e.Pid),
 		}
 		pm, ok := im.known[fp]
 		if ok {
 			pm.keep = true
 			continue
 		}
+		var process string
+		if e.OSMetadata != nil {
+			if module, err := e.OSMetadata.GetModule(); err == nil {
+				process = module
+			}
+		}
 		pm = &portMeta{
 			keep: true,
 			port: Port{
 				Proto:   "tcp",
 				Port:    e.Local.Port(),
-				Process: procNameOfPid(e.Pid),
+				Process: process,
 			},
 		}
 		im.known[fp] = pm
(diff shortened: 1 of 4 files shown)


portlist/portlist_windows.go
```diff
@@ -52,6 +53,8 @@ func (im *windowsImpl) AppendListeningPorts(base []Port) ([]Port, error) {
 		return nil, err
 	}
 
+	pi := newPidInfo()
```
bradfitz, line 56: let's do this lazily when it's first needed. Most calls (every second or so) there will be no changes in the set of open TCP connections so this work would be otherwise wasted and never used. (the line 78 continue would hit each time)

portlist/portlist_windows.go
```diff
@@ -92,29 +95,112 @@ func (im *windowsImpl) AppendListeningPorts(base []Port) ([]Port, error) {
...
-	const da = windows.PROCESS_QUERY_LIMITED_INFORMATION
-	h, err := syscall.OpenProcess(da, false, uint32(pid))
+// pidInfo is a mapping of process IDs to process names that incorporates
+// information from Windows' Service Control Manager. We use this information
+// to disambiguate svchost processes (when possible).
+type pidInfo map[uint32]string
+
+// newPidInfo instantiates a pidInfo and pre-populates it with PID mappings for
+// all Windows services that are currently running.
+func newPidInfo() (result pidInfo) {
```
bradfitz, line 109: can we only do all this Service Manager work if the process name is "svchost"? Seems overkill to do each time "python -m SimpleHTTPServer 8080" appears for a second and then disappears and then comes back. If we know right away that it's "python", then all the Service Manager querying is for nothing?

portlist/portlist_windows.go
bradfitz: I realized later that it might not've been obvious how often this function (AppendListeningPorts) runs. On Linux it's every 1 second. On Windows it's been 5 seconds because of its relative speed. Ideally it'd be nice to be 1 second also. 

Actually, ideally we'd not poll every N seconds at all on any platform and instead just subscribe (on Linux: via eBPF) to changes in the TCP/UDP listening tables.

But meanwhile assume that this runs every 5 seconds.

portlist/portlist_windows.go
dblohm7: Yeah, we can do that.

dblohm7:
I did some more digging into how programs like netstat itself handle this, and it turns out that if we request module information when calling `GetExtendedTcpTable`, we can resolve the information we need with additional API calls.

I refactored `netstat_windows` to work with this format. Right now I'm stuffing the entire `_MIB_TCP*ROW_OWNER_MODULE` into an `OSMetadata` field, with the belief that instances of `Entry` will be short-lived and we'll probably be either using `OSMetadata` or throwing away the `Entry` within a short period of time. (As opposed to only saving the stuff that `Entry` doesn't and then converting back and forth)

bradfitz:
Oh, sorry, I missed that this was ever updated. Can you rebase to fix the merge conflict? I'll try to review & understand it meanwhile.

bradfitz (approved):

PR #6087, opened 2022-10-27
cmd/containerboot: PID1 for running tailscaled in a container.

danderson:
This implements the same functionality as the former run.sh, but in Go and with a little better awareness of tailscaled's lifecycle.

Also adds TS_AUTH_ONCE, which fixes the unfortunate behavior run.sh had where it would unconditionally try to reauth every time if you gave it an authkey, rather than try to use it only if auth is actually needed. This makes it a bit nicer to deploy these containers in automation, since you don't have to run the container once, then go and edit its definition to remove authkeys.


cmd/containerboot/main.go
```diff
@@ -0,0 +1,401 @@
...
+		if err := installIPTablesRule(ctx, cfg.ProxyTo, st.TailscaleIPs); err != nil {
+			log.Fatalf("installing proxy rules: %v", err)
+		}
+	}
+	if cfg.KubeSecret != "" {
+		if err := storeDeviceID(ctx, cfg.KubeSecret, string(st.Self.ID)); err != nil {
+			log.Fatalf("storing device ID in kube secret: %v", err)
+		}
+	}
+}
```
andrew-d, line 142: Once `main` exits, I think the process exits, right? vs. the original `run.sh` where it waits forever before running?

cmd/containerboot/main.go
```diff
@@ -0,0 +1,401 @@
...
+			log.Printf("Failed to enable IP forwarding: %v", err)
+			log.Printf("To run tailscale as a proxy or router container, IP forwarding must be enabled.")
+			if cfg.InKubernetes {
+				log.Fatalf("You can either set the sysctls as a privileged initContainer, or run the tailscale container with privileged=true.")
+			} else {
+				log.Fatalf("You can fix this by running the container with privileged=true, or the equivalent in your container runtime that permits access to sysctls.")
+			}
+		}
+	}
+
```
andrew-d, line 104: For lack of a better place to put it: if this is PID1 do you need to reap child processes? Something like this?
https://go.dev/play/p/CoDZcQ8uZys

danderson: Switched the wait for just tailscaled to be a general PID reaper that also exits when the tailscaled PID exits.

andrew-d: Oh, that's much cleaner :+1:

cmd/containerboot/kube.go
```diff
@@ -0,0 +1,175 @@
...
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// findKeyInKubeSecret inspects the kube secret secretName for a data
+// field called "authkey", and returns its value if present.
+func findKeyInKubeSecret(ctx context.Context, secretName string) (string, error) {
+	req, err := http.NewRequest("GET", fmt.Sprintf("/api/v1/namespaces/%s/secrets/%s", kubeNamespace, secretName), nil)
```
andrew-d, line 30: LGTM from https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#read-secret-v1-core

cmd/containerboot/kube.go
```diff
@@ -0,0 +1,175 @@
...
+	m := map[string]map[string]string{
+		"stringData": map[string]string{
+			"device_id": deviceID,
+		},
+	}
+	var b bytes.Buffer
+	if err := json.NewEncoder(&b).Encode(m); err != nil {
+		return err
+	}
+	req, err := http.NewRequest("PATCH", fmt.Sprintf("/api/v1/namespaces/%s/secrets/%s?fieldManager=kubectl-patch", kubeNamespace, secretName), &b)
```
andrew-d, line 71: Do we want to use the `kubectl-patch` fieldManager here? I don't know a ton about this, but from [the docs](https://kubernetes.io/docs/reference/using-api/server-side-apply/) it appears to be for some sort of conflict detection; should we have our own?

Dockerfile
```diff
@@ -66,10 +66,12 @@ RUN GOARCH=$TARGETARCH go install -ldflags="\
...
+      -v ./cmd/tailscale ./cmd/tailscaled ./cmd/containerboot
 
 FROM alpine:3.16
 RUN apk add --no-cache ca-certificates iptables iproute2 ip6tables
 
 COPY --from=build-env /go/bin/* /usr/local/bin/
-COPY --from=build-env /go/src/tailscale/docs/k8s/run.sh /usr/local/bin/
+# For compat with the previous run.sh, although ideally you should be
+# using build_docker.sh which sets an entrypoint for the image.
+RUN ln -s /usr/local/bin/containerboot /usr/local/bin/run.sh
```
maisem, line 77: /tailscale/run.sh

cmd/containerboot/kube.go
```diff
@@ -0,0 +1,175 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
```
maisem, line 24: curious, does this blow up the binary size?

cmd/containerboot/kube.go
```diff
@@ -0,0 +1,175 @@
...
+func deleteAuthKey(ctx context.Context, secretName string) error {
+	m := []struct {
+		Op   string `json:"op"`
+		Path string `json:"path"`
+	}{
+		{
+			Op:   "remove",
+			Path: "/data/authkey",
+		},
+	}
```
andrew-d, line 93: Where does this come from? [The docs](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/) seem to indicate that this is a "JSON Patch, RFC 6902", so if so, worth adding a comment?

cmd/containerboot/main.go
```diff
@@ -0,0 +1,426 @@
...
+		AuthKey:         defaultEnv("TS_AUTH_KEY", ""),
+		Routes:          defaultEnv("TS_ROUTES", ""),
+		ProxyTo:         defaultEnv("TS_DEST_IP", ""),
+		DaemonExtraArgs: defaultEnv("TS_TAILSCALED_EXTRA_ARGS", ""),
+		ExtraArgs:       defaultEnv("TS_EXTRA_ARGS", ""),
+		InKubernetes:    os.Getenv("KUBERNETES_SERVICE_HOST") != "",
+		UserspaceMode:   defaultBool("TS_USERSPACE", true),
+		StateDir:        defaultEnv("TS_STATE_DIR", ""),
+		AcceptDNS:       defaultBool("TS_ACCEPT_DNS", false),
+		KubeSecret:      defaultEnv("TS_KUBE_SECRET", "tailscale"),
```
maisem, line 79: you can't default this here, as is you will fail on line 90

maisem: when running outside k8s

cmd/containerboot/kube.go
```diff
@@ -0,0 +1,175 @@
...
+	cp := x509.NewCertPool()
+	cp.AppendCertsFromPEM(bs)
+	kubeHTTP = &http.Transport{
+		TLSClientConfig: &tls.Config{
+			RootCAs: cp,
+		},
+		IdleConnTimeout: time.Second,
+	}
+}
```
andrew-d, line 154: I'm not a huge fan of complicated `init` functions; worth having a `sync.Once`/accessor? At the very least, having this in `init` means that this binary will print/exit with `log.Fatalf` before hitting the `log.SetPrefix` in `main` if anything here errors.

cmd/containerboot/kube.go
```diff
@@ -0,0 +1,175 @@
...
+)
+
+func init() {
+	// If running in Kubernetes, set things up so that doKubeRequest
+	// can talk successfully to the kube apiserver.
+	if os.Getenv("KUBERNETES_SERVICE_HOST") == "" {
+		return
+	}
+
+	kubeHost = os.Getenv("KUBERNETES_SERVICE_HOST") + ":" + os.Getenv("KUBERNETES_SERVICE_PORT_HTTPS")
```
andrew-d, line 128: If we have `KUBERNETES_SERVICE_HOST`, should we always expect `KUBERNETES_SERVICE_PORT`? Or do we need to handle when that second one is missing?

cmd/containerboot/main.go
```diff
@@ -0,0 +1,426 @@
...
+func startAndAuthTailscaled(ctx context.Context, cfg *settings) (*ipnstate.Status, error) {
+	args := tailscaledArgs(cfg)
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, unix.SIGTERM, unix.SIGINT)
+	// tailscaled runs without context, since it needs to persist
+	// beyond the startup timeout in ctx.
+	cmd := exec.Command("tailscaled", args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	log.Printf("Starting tailscaled")
```
andrew-d, line 172: Idle musing: we should look at the Linux process contract and see if we need to set anything in `cmd.SysProcAttr`; I think we want `Setpgid: true` here

cmd/containerboot/main.go
```diff
@@ -0,0 +1,426 @@
...
+		log.Printf("Received SIGTERM from container runtime, shutting down tailscaled")
+		cmd.Process.Signal(unix.SIGTERM)
+	}()
+	go func() {
+		// Reap all processes, since we are PID1 and need to collect
+		// zombies.
+		for {
+			var status unix.WaitStatus
+			pid, err := unix.Wait4(-1, &status, 0, nil)
+			if err != nil {
```
andrew-d, line 188: I think you probably need to handle `unix.EINTR` (and just continue) here?

cmd/containerboot/main.go
```diff
@@ -0,0 +1,401 @@
...
+	if cfg.InKubernetes && cfg.KubeSecret != "" && cfg.AuthKey == "" {
+		key, err := findKeyInKubeSecret(ctx, cfg.KubeSecret)
+		if err != nil {
+			log.Fatalf("Getting authkey from kube secret: %v", err)
+		}
+		if key != "" {
+			log.Print("Using authkey found in kube secret")
+			cfg.AuthKey = key
+		}
+		log.Print("No authkey found in kube secret and TS_AUTHKEY not provided, login will be interactive.")
```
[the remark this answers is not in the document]
danderson: Done

cmd/containerboot/main.go
danderson: Gaaah I want my tristate values :(

danderson: Fixed, l90 now checks the envvar directly, so that initialization block still captures all the defaults for everything.

cmd/containerboot/main.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+
+	if cfg.SOCKSProxyAddr != "" {
+		args = append(args, "--socks5-server="+cfg.SOCKSProxyAddr)
+	}
+	if cfg.HTTPProxyAddr != "" {
+		args = append(args, "--outbound-http-proxy-listen="+cfg.HTTPProxyAddr)
+	}
+	if cfg.DaemonExtraArgs != "" {
+		args = append(args, strings.Fields(cfg.DaemonExtraArgs)...)
```
andrew-d, line 288: `strings.Fields` is ... technically not the same as before, since it doesn't do shell escaping in the same way (e.g. `TS_TAILSCALED_EXTRA_ARGS="foo \"bar baz\"" ./run.sh` behaves differently now). I'm happy to ignore if you are, though.

cmd/containerboot/kube.go
maisem: We could use tailscale-container as the manager here

cmd/containerboot/kube.go
```diff
@@ -0,0 +1,175 @@
...
+)
+
+// findKeyInKubeSecret inspects the kube secret secretName for a data
+// field called "authkey", and returns its value if present.
+func findKeyInKubeSecret(ctx context.Context, secretName string) (string, error) {
+	req, err := http.NewRequest("GET", fmt.Sprintf("/api/v1/namespaces/%s/secrets/%s", kubeNamespace, secretName), nil)
+	if err != nil {
+		return "", err
+	}
+	resp, err := doKubeRequest(ctx, req)
```
maisem, line 34: Maybe some opportunity to share the code with the tailscale.com/kube package

maisem: Can happen later

cmd/containerboot/main.go
maisem: Oh this is annoying :(

cmd/containerboot/kube.go
danderson: Ack, will refactor once this is merged.

cmd/containerboot/main.go
danderson: I'm not following how this matters, for the constrained environment of this container?

cmd/containerboot/main.go
danderson: Done

cmd/containerboot/main.go
danderson: Looking at the args for both tailscaled and tailscale up, I don't see any reasonable commandline you could construct that embeds quoted elements with spaces. On that basis I don't feel strongly about adding a bunch of complexity here. WDYT?

andrew-d: I was thinking someone wanted to pass a `--hostname` or `--operator` with a value that has spaces; something like this: `TS_TAILSCALED_EXTRA_ARGS='--hostname "with spaces"'`, which would previously (correctly) set the hostname to `"with spaces"` and now... presumably just passes garbage?

I'm happy to say "don't do that" here, though.

cmd/containerboot/main.go
andrew-d: Huge caveat here: I am not a Unix/POSIX expert.

My understanding of process groups is that when a signal is directed to a process group, the signal is delivered to each process that is a member of the group. By default, when a shell `fork`s a subprocess, I believe that it makes the child process the leader of its own (new) process group. By launching `tailscaled` here ourselves, that isn't true any more, and if something delivers a signal to this process, both `containerboot` and `tailscaled` will be part of the same process group and both will receive the signal.

Since I don't know exactly how container runtimes work w.r.t. signals, I think the least-confusing approach here is to do exactly what you're doing to forward a signal from PID1 to the child `tailscaled`, and additionally make `tailscaled` the leader of its own process group so it doesn't get any signals except what `containerboot` explicitly sends it.

(again: not an expert here by any means)

cmd/containerboot/kube.go
maisem: https://kubernetes.io/docs/reference/kubectl/#in-cluster-authentication-and-namespace-overrides

cmd/containerboot/main.go
```diff
@@ -0,0 +1,423 @@
...
+		SOCKSProxyAddr:  defaultEnv("TS_SOCKS5_SERVER", ""),
+		HTTPProxyAddr:   defaultEnv("TS_OUTBOUND_HTTP_PROXY_LISTEN", ""),
+		Socket:          defaultEnv("TS_SOCKET", "/tmp/tailscaled.sock"),
+		AuthOnce:        defaultBool("TS_AUTH_ONCE", false),
+	}
+
+	if cfg.ProxyTo != "" && cfg.UserspaceMode {
+		log.Fatal("TS_DEST_IP is not supported with TS_USERSPACE")
+	}
+	if os.Getenv("TS_KUBE_SECRET") != "" && !cfg.InKubernetes {
```
maisem, line 89: nit: add a comment saying that we want to check the env var directly here in case someone accidentally refactors this in the future.

maisem (approved):

cmd/containerboot/main.go
andrew-d: SGTM

cmd/containerboot/main.go
danderson: sgtm, done. I think in the container setting it might not matter, but since all the plumbing is there in pid 1 to relay sigterm and do an orderly shutdown, we should make it actually work.

cmd/containerboot/main.go
danderson: I actually removed this test entirely, because technically it's breaking compat with run.sh: run.sh was completely happy to let you set TS_KUBE_SECRET, and just ignored it if not running in k8s. I could imagine that people have things set up to unconditionally pass TS_KUBE_SECRET, but then use that config both in docker for local dev and in k8s in prod. So, even though it might allow some buggy configs, I think we need to keep allowing it :(

maisem: oh sure, bug for bug compat sgtm

andrew-d (approved):
✅

PR #6097, opened 2022-10-27
control/controlhttp: add AcceptHTTP hook to add coalesced Server->Client write

bradfitz:
New plan for #5972. Instead of sending the public key in the clear (from earlier unreleased 246274b8e91) where the client might have to worry about it being dropped or tampered with and retrying, we'll instead send it post-Noise handshake but before the HTTP/2 connection begins.

This replaces the earlier extraHeaders hook with a different sort of hook that allows us to combine two writes on the wire in one packet.

Updates #5972

### DIFF
--- a/control/controlhttp/server.go
+++ b/control/controlhttp/server.go
@@ -9,7 +9,10 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"io"
+	"net"
 	"net/http"
+	"time"
 
 	"nhooyr.io/websocket"
 	"tailscale.com/control/controlbase"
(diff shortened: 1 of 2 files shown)


control/controlhttp/server.go
```diff
@@ -75,12 +75,30 @@ func AcceptHTTP(ctx context.Context, w http.ResponseWriter, r *http.Request, pri
...
 		return nil, fmt.Errorf("noise handshake failed: %w", err)
 	}
 
+	if earlyWrite != nil {
+		if deadline, ok := ctx.Deadline(); ok {
+			if err := conn.SetDeadline(deadline); err != nil {
+				return nil, fmt.Errorf("setting conn deadline: %w", err)
+			}
+			defer conn.SetDeadline(time.Time{})
+			if err := earlyWrite(nc.ProtocolVersion(), cwc); err != nil {
```
raggi, line 92: Does calling earlyWrite need to depend on having a deadline? If so it's worth documenting

danderson (approved):

raggi (approved):

control/controlhttp/server.go
maisem: should this be outside the deadline if check?

control/controlhttp/server.go
```diff
@@ -75,12 +75,30 @@ func AcceptHTTP(ctx context.Context, w http.ResponseWriter, r *http.Request, pri
 	}
 	conn = netutil.NewDrainBufConn(conn, brw.Reader)
 
-	nc, err := controlbase.Server(ctx, conn, private, init)
+	cwc := newWriteCorkingConn(conn)
+
+	nc, err := controlbase.Server(ctx, cwc, private, init)
 	if err != nil {
 		conn.Close()
```
maisem, line 82: you are not not calling `conn.Close` in all error paths

control/controlhttp/server.go
maisem: You are still not calling earlyWrite when there is no `ctx.Deadline`

maisem: If that is intended, that explicitly needs docs.

bradfitz: whoops, good catch

control/controlhttp/server.go
bradfitz: nice, done, thanks

control/controlhttp/server.go
```diff
@@ -75,12 +75,31 @@ func AcceptHTTP(ctx context.Context, w http.ResponseWriter, r *http.Request, pri
...
 		return nil, fmt.Errorf("noise handshake failed: %w", err)
 	}
 
+	if earlyWrite != nil {
+		if deadline, ok := ctx.Deadline(); ok {
+			if err := conn.SetDeadline(deadline); err != nil {
+				return nil, fmt.Errorf("setting conn deadline: %w", err)
```
maisem, line 89: missed this

control/controlhttp/server.go
```diff
@@ -75,12 +75,31 @@ func AcceptHTTP(ctx context.Context, w http.ResponseWriter, r *http.Request, pri
...
+			}
+			defer conn.SetDeadline(time.Time{})
+		}
+		if err := earlyWrite(nc.ProtocolVersion(), cwc); err != nil {
+			return nil, err
```
maisem, line 94: and this

maisem (approved):

control/controlhttp/server.go
maisem: missed `conn.Close()`

bradfitz: defer it is

maisem (approved):

PR #6112, opened 2022-10-28
ipn/ipnlocal: add tags and a few other details to self status

### LINKED ISSUE #6076: FR: Machine's own tags should be available in Self field when running `tailscale up --json`

I'm trying to install some automation script on my servers that discovers what its capabilities are on my tailnet based on what ACL tags are assigned to the node it's running on. It also should not have to directly contact tailscale control api for this information, as that would require distributing API keys which is less than ideal.

### How should we solve this?

Populate Self.Tags when querying local-api for Status

I have to manually signal my script with capabilities information on a machine after it connects to the tailnet, which gets tedious when onboarding new devices to my network.

### Anything else?

_No response_

knyar:
This makes tags, creation time, exit node option and primary routes for the current node exposed via `tailscale status --json`

Fixes #6076

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -584,11 +584,12 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 	sb.MutateSelfStatus(func(ss *ipnstate.PeerStatus) {
 		ss.Online = health.GetInPollNetMap()
 		if b.netMap != nil {
+			ss.InNetworkMap = true
 			ss.HostName = b.netMap.Hostinfo.Hostname
 			ss.DNSName = b.netMap.Name
 			ss.UserID = b.netMap.User
 			if sn := b.netMap.SelfNode; sn != nil {
-				ss.ID = sn.StableID
+				peerStatusFromNode(ss, sn)
 				if c := sn.Capabilities; len(c) > 0 {
 					ss.Capabilities = append([]string(nil), c...)
 				}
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/local.go
```diff
@@ -592,6 +593,14 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 				if c := sn.Capabilities; len(c) > 0 {
 					ss.Capabilities = append([]string(nil), c...)
 				}
+				if sn.Tags != nil {
+					v := views.SliceOf(sn.Tags)
+					ss.Tags = &v
+				}
+				ss.Created = sn.Created
+				ss.ExitNodeOption = tsaddr.PrefixesContainsFunc(sn.AllowedIPs, func(r netip.Prefix) bool {
```
maisem, line 601: tsaddr.ContainsExitRoutes

ipn/ipnlocal/local.go
```diff
@@ -592,6 +593,14 @@ func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func
 				if c := sn.Capabilities; len(c) > 0 {
 					ss.Capabilities = append([]string(nil), c...)
 				}
+				if sn.Tags != nil {
```
maisem, line 596: We should also add PrimaryRoutes

maisem: or maybe we should consolidate this with `populatePeerStatusLocked` to not have a lot of duplicated code.

ipn/ipnlocal/local.go
knyar: Oh nice, I have not noticed that.

ipn/ipnlocal/local.go
knyar: Done

maisem (approved):

PR #6116, opened 2022-10-28
ipn/ipnlocal: fix a log line having function pointers instead of values

mihaip:
Followup to using `ipn.PrefsView` (#6031).


bradfitz (approved):

maisem (approved):
oops! Thanks!
This keeps biting us as we use more views, we should add a staticcheck or something to catch these.

PR #6133, opened 2022-10-30
control/controlclient, tailcfg: add support for EarlyNoise payload before http/2

bradfitz:
Not yet used, but skipped over, parsed, and tested.

Updates #5972

### DIFF
--- a/control/controlclient/noise.go
+++ b/control/controlclient/noise.go
@@ -7,7 +7,10 @@ package controlclient
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
 	"encoding/json"
+	"errors"
+	"io"
 	"math"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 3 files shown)


control/controlclient/noise.go
```diff
@@ -34,6 +36,77 @@ type noiseConn struct {
...
+}
+
+func (c *noiseConn) RoundTrip(r *http.Request) (*http.Response, error) {
+	return c.h2cc.RoundTrip(r)
+}
+
+const hdrLen = 9 // http2 frame header size; also size of our early payload size header
+
+func (c *noiseConn) Read(p []byte) (n int, err error) {
+	if need := hdrLen - c.headerBytesRead; need > 0 {
```
danderson, line 56: This is a lot of extra logic in every Read call. Since this is all for a payload that's only seen once at the start of the connection, could you instead do all this at noiseConn construction time, and avoid all this overhead on each future call? As a bonus you can ReadFull and avoid a lot of the fiddly buffer management that you're forced to have inside Read.

bradfitz: PTAL

control/controlclient/noise.go
```diff
@@ -34,6 +37,57 @@ type noiseConn struct {
...
+
+func (r returnErrReader) Read([]byte) (int, error) { return 0, r.err }
+
+func (c *noiseConn) Read(p []byte) (n int, err error) {
+	c.readHeaderOnce.Do(c.readHeader)
+	return c.reader.Read(p)
+}
+
+func (c *noiseConn) readHeader() {
+	var hdr [9]byte
```
bradfitz, line 65: whoops, meant hdrLen. fixing.

danderson (approved):

control/controlclient/noise.go
```diff
@@ -34,6 +37,57 @@ type noiseConn struct {
...
+	reader            io.Reader     // (effectively Conn.Reader after header)
+	earlyPayloadReady chan struct{} // closed after earlyPayload is set (including set to nil)
+	earlyPayload      *tailcfg.EarlyNoise
+}
+
+func (c *noiseConn) RoundTrip(r *http.Request) (*http.Response, error) {
+	return c.h2cc.RoundTrip(r)
+}
+
+const hdrLen = 9 // http2 frame header size; also size of our early payload size header
```
maisem, line 51: define this closer to earlyPayloadMagic, or move that up here.

maisem (approved):

control/controlclient/noise.go
bradfitz: done w/ comments

Issue #5021, opened 2022-07-08
Connection reset by server while client is asleep


TomiBelan:
### What is the issue?

I use ssh to connect from my laptop to a server over Tailscale. When I suspend the laptop, the connection is reset after some time. When I resume the laptop and type any character, the client notices that the connection is broken and exits.

### Steps to reproduce

The client is a Windows laptop. I use various ssh clients (official ssh.exe, cygwin ssh.exe, libssh). I use `-o tcpkeepalive=no` ([manual](https://man.openbsd.org/ssh_config#TCPKeepAlive), [SE](https://unix.stackexchange.com/questions/34004/how-does-tcp-keepalive-work-in-ssh), [TLDP](https://tldp.org/HOWTO/TCP-Keepalive-HOWTO/usingkeepalive.html), [source](https://github.com/openssh/openssh-portable/blob/669a56bcfe73f8b985f2bba476ba834d55253acf/sshconnect.c#L525))

The server is a Linux machine running sshd from OpenSSH_7.6p1. The sshd_config is default. I am not using "Tailscale SSH". The server runs tailscaled unprivileged with `--tun=userspace-networking`, so sshd thinks it's a local TCP connection from 127.0.0.1.

1. Run the ssh client with the server's Tailscale IP and log in.
2. Put the laptop to sleep. Resume it after at least X hours (not sure about exact threshold).
3. Type any character in the ssh client. It prints: `client_loop: send disconnect: Connection reset by peer` and exits.

If I connect directly (not to Tailscale IP) immediately after resuming the laptop, and run `ps x`, it reveals that the `sshd: user@pts/N` process and the shell are already dead. I think this means that at some point while the laptop was asleep, either sshd or tailscaled tried to ping the client, timed out, and closed the connection. When the client sends a next packet, it receives RST.

I expect this will be hard to reproduce and you'll need more help from me to debug this. Is there anything else I should try or any information I can give you?

### Are there any recent changes that introduced the issue?

It used to work - the ssh connection used to survive overnight, as long as the shell was idle. The most likely recent change is that I upgraded the server from Tailscale 1.16.0 to 1.26.0.

When I downgraded the server side back to 1.16.0, it works again. That could be a coincidence, but it strongly suggests this is a regression between 1.16.0 and 1.26.0.

I'll try bisecting the version if I have time. But it will take a while. It doesn't happen with short sleeps, so I only get one try per day.

### OS

Linux, Windows

### OS version

[...]

PR #6147, opened 2022-10-31
wgengine/netstack: use 72h as the KeepAlive Idle time for Tailscale SSH

maisem:
Setting TCP KeepAlives for Tailscale SSH connections results in them unnecessarily disconnecting. However, we can't turn them off completely as that would mean we start leaking sessions waiting for a peer to come back which may have gone away forever (e.g. if the node was deleted from the tailnet during a session).

Updates #5021

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -812,7 +812,7 @@ func (ns *Impl) acceptTCP(r *tcp.ForwarderRequest) {
 	// request until we're sure that the connection can be handled by this
 	// endpoint. This function sets up the TCP connection and should be
 	// called immediately before a connection is handled.
-	createConn := func() *gonet.TCPConn {
+	createConn := func(opts ...tcpip.SettableSocketOption) *gonet.TCPConn {
 		ep, err := r.CreateEndpoint(&wq)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


andrew-d (approved):

wgengine/netstack/netstack.go
```diff
@@ -859,7 +861,14 @@ func (ns *Impl) acceptTCP(r *tcp.ForwarderRequest) {
 
 	if ns.lb != nil {
 		if reqDetails.LocalPort == 22 && ns.processSSH() && ns.isLocalIP(dialIP) {
-			c := createConn()
+			// Use a higher keepalive idle time for SSH connections, as they are
+			// typically long lived and idle connections are more likely to be
+			// intentional. Ideally we would turn this off entirely, but we can't
+			// tell the difference between a long lived connection that is idle
+			// vs a connection that is dead because the peer has gone away.
```
DentonGentry, line 868: I imagine our long-term fix is to get actual packet forwarding into netstack, rather than terminate the TCP session on one side and made a new one other the other side.

DentonGentry (approved):

wgengine/netstack/netstack.go
maisem: Yeah, but I think we will have something like this for Tailscale SSH even then as we have to handle the incoming connection in process.

PR #6154, opened 2022-10-31
cmd/tailscale,ipn: implement lock sign command

twitchyliquid64:
This implements `tailscale lock sign <node-key>`.

### DIFF
--- a/client/tailscale/localclient.go
+++ b/client/tailscale/localclient.go
@@ -36,6 +36,7 @@ import (
 	"tailscale.com/safesocket"
 	"tailscale.com/tailcfg"
 	"tailscale.com/tka"
+	"tailscale.com/types/key"
 )
 
 // defaultLocalClient is the default LocalClient when using the legacy
@@ -825,6 +826,25 @@ func (lc *LocalClient) NetworkLockModify(ctx context.Context, addKeys, removeKey
 		return nil, err
 	}
 	return pr, nil
+}
+
+// NetworkLockSign signs the specified node-key and transmits that signature to the control plane.
+// rotationPublic, if specified, must be an ed25519 public key.
+func (lc *LocalClient) NetworkLockSign(ctx context.Context, nodeKey key.NodePublic, rotationPublic []byte) error {
+	var b bytes.Buffer
+	type signRequest struct {
+		NodeKey        key.NodePublic
+		RotationPublic []byte
+	}
+
+	if err := json.NewEncoder(&b).Encode(signRequest{NodeKey: nodeKey, RotationPublic: rotationPublic}); err != nil {
+		return err
+	}
+
+	if _, err := lc.send(ctx, "POST", "/localapi/v0/tka/sign", 200, &b); err != nil {
+		return fmt.Errorf("error: %w", err)
+	}
+	return nil
 }
 
 // tailscaledConnectHint gives a little thing about why tailscaled (or
--- a/cmd/tailscale/cli/network-lock.go
+++ b/cmd/tailscale/cli/network-lock.go
@@ -26,6 +26,7 @@ var netlockCmd = &ffcli.Command{
 		nlStatusCmd,
 		nlAddCmd,
 		nlRemoveCmd,
+		nlSignCmd,
 	},
 	Exec: runNetworkLockStatus,
 }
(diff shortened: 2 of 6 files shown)


ipn/ipnlocal/network-lock.go
```diff
@@ -430,6 +430,45 @@ func (b *LocalBackend) NetworkLockKeyTrustedForTest(keyID tkatype.KeyID) bool {
 	return b.tka.authority.KeyTrusted(keyID)
 }
 
+// NetworkLockSign signs the given node-key and submits it to the control plane.
+// rotationPublic, if specified, must be an ed25519 public key.
+func (b *LocalBackend) NetworkLockSign(nodeKey key.NodePublic, rotationPublic []byte) error {
+	b.mu.Lock()
```
sailorfrag, line 436: Maybe move everything that needs the lock into a helper function so you can defer the unlock?

ipn/ipnlocal/network-lock.go
```diff
@@ -430,6 +430,45 @@ func (b *LocalBackend) NetworkLockKeyTrustedForTest(keyID tkatype.KeyID) bool {
...
+	if b.tka == nil {
+		b.mu.Unlock()
+		return errNetworkLockNotActive
+	}
+	if !b.tka.authority.KeyTrusted(b.nlPrivKey.KeyID()) {
+		b.mu.Unlock()
+		return errors.New("this node is not trusted by network lock")
```
sailorfrag, line 443: I wonder if this should be a warning rather than an error. It might be trusted in future, making the signature retroactively acceptable. Or maybe that's actually a terrible idea.

sailorfrag (approved):

tailcfg/tka.go
```diff
@@ -214,3 +214,22 @@ type TKADisableRequest struct {
...
 }
+
+// TKASubmitSignatureRequest transmits a node-key signature to the control plane.
+//
+// This is the request schema for a /tka/sign noise RPC.
+type TKASubmitSignatureRequest struct {
+	// Version is the client's capabilities.
+	Version CapabilityVersion
+
+	// NodeKey is the client's current node key.
```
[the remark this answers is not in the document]
twitchyliquid64: Done.

cmd/tailscale/cli/network-lock.go
```diff
@@ -163,3 +164,27 @@ func runNetworkLockModify(ctx context.Context, addArgs, removeArgs []string) err
...
+	ShortHelp:  "Signs a node-key and transmits that signature to the control plane",
+	Exec:       runNetworkLockSign,
+}
+
+// TODO(tom): Implement specifying the rotation key for the signature.
+func runNetworkLockSign(ctx context.Context, args []string) error {
+	switch len(args) {
+	case 0:
+		return errors.New("expected node-key as second argument")
+	case 1:
```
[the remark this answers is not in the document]
twitchyliquid64: Done.

client/tailscale/localclient.go
```diff
@@ -827,6 +828,25 @@ func (lc *LocalClient) NetworkLockModify(ctx context.Context, addKeys, removeKey
 	return pr, nil
 }
 
+// NetworkLockSign signs the specified node-key and transmits that signature to the control plane.
+// rotationPublic, if specified, must be an ed25519 public key.
+func (lc *LocalClient) NetworkLockSign(ctx context.Context, nodeKey key.NodePublic, rotationPublic []byte) error {
+	var b bytes.Buffer
+	type signRequest struct {
```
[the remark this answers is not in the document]
twitchyliquid64: I think it would be good to move all the unix-socket wire types to their own package, but if we do this we should do it for all the RPCs.

ipn/ipnlocal/network-lock.go
twitchyliquid64: I think I want the default to be 'only let you sign if the signature will work', but maybe we want a flag or something in future to let you submit a signature that doesnt work now but will shortly.

ipn/ipnlocal/network-lock.go
twitchyliquid64: Re-did everything into a closure and its much cleaner, thanks.

PR #6155, opened 2022-10-31
Add tailssh support on FreeBSD

patmaddox:
The changes to Tailscale are straightforward. I could use some help upstreaming the necessary changes to `u-root` (https://github.com/u-root/u-root/pull/2544).

Another possibility might be to try [term](https://pkg.go.dev/golang.org/x/term) or [go-tty](https://github.com/mattn/go-tty) for TTY support, since the dependency on `termios` appears minimal.


go.mod
```diff
@@ -281,3 +281,5 @@ require (
 	mvdan.cc/unparam v0.0.0-20211002134041-24922b6997ca // indirect
 	software.sslmate.com/src/go-pkcs12 v0.0.0-20210415151418-c5206de65a78 // indirect
 )
+
+replace github.com/u-root/u-root v0.9.0 => github.com/patmaddox/u-root v0.0.0-20221101021341-da6a6f63938d
```
patmaddox, line 285: Of course this is just for purposes of this PR. `u-root` will need to be updated.

ipn/ipnlocal/local.go
```diff
@@ -2120,6 +2120,7 @@ func (b *LocalBackend) checkSSHPrefsLocked(p *ipn.Prefs) error {
 		if !envknob.UseWIPCode() {
 			return errors.New("The Tailscale SSH server is disabled on macOS tailscaled by default. To try, set env TAILSCALE_USE_WIP_CODE=1")
 		}
+	case "freebsd": {}
```
danderson, line 2123: No need for `{}`, Go switch statements can be empty and don't fall through by default.

danderson:
Change to tailscale looks minimal and fine to me. So, modulo the u-root changes I have no problem shipping this.

I think we'd rather make u-root than tear it out and have to requal all platforms on a different library, but let's see how u-root upstreaming goes. We're not committers up there (I don't think?), so I can review but it won't carry any more weight than another internet rando.

patmaddox:
What behavior does all of the opcode stuff support anyway? I deleted it entirely, and it still worked when connecting from iTerm2. I suppose it's for other terminal applications that behave differently?

patmaddox:
What behavior does the opcode stuff support anyway? I deleted it all, and it still worked when SSHing from iTerm2. I suppose other clients behave differently and need it to respond to all those opcodes?

patmaddox:
The `u-root` changes have been merged upstream. This should be good to go.

bradfitz:
Looks like this breaks darwin?

bradfitz:
Okay, bump the dep and let's get this in!

bradfitz:
I bumped it in https://github.com/tailscale/tailscale/pull/6273. I want to make sure this gets in for our upcoming release.

bradfitz:
Merged in #6273, thanks!

PR #6165, opened 2022-11-01
wgengine/netlog: preserve Tailscale addresses for exit traffic

dsnet:
Exit node traffic is aggregated to protect the privacy of those using an exit node. However, it is reasonable to at least log which nodes are making most use of an exit node.

For a node using an exit node,
the source will be the taiscale IP address of itself, while the destination will be zeroed out.

For a node that serves as an exit node,
the source will be zeroed out,
while the destination will be tailscale IP address of the node that initiated the exit traffic.

### DIFF
--- a/wgengine/netlog/logger.go
+++ b/wgengine/netlog/logger.go
@@ -187,11 +187,13 @@ func recordStatistics(logger *logtail.Logger, nodeID tailcfg.StableNodeID, start
 		for p := range prefixes {
 			if p.Contains(a) && p.Bits() > 0 {
 				withinRoute = true
+				break
 			}
 		}
 		return withinRoute && tsaddr.IsTailscaleIP(a), withinRoute && !tsaddr.IsTailscaleIP(a)
 	}
 
+	exitTraffic := make(map[netlogtype.Connection]netlogtype.Counts)
 	for conn, cnts := range tunStats {
 		srcIsTailscaleIP, srcWithinSubnet := classifyAddr(conn.Src.Addr())
 		dstIsTailscaleIP, dstWithinSubnet := classifyAddr(conn.Dst.Addr())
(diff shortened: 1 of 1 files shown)


dsnet:
\cc @mayakacz

bradfitz:
nit: preserve, not preserved

dsnet:
> nit: preserve, not preserved

Fixed!

mayakacz:
Is this just Tailscale IP, or also port & protocol?

dsnet:
Just the Tailscale address (no port, no protocol). I'm a bit conservative about revealing less information unless there is a benefit of doing so.

ramyfication:
> For a node that serves as an exit node, the source will be zeroed out, while the destination will be tailscale IP address of the node that initiated the exit traffic.

In https://github.com/tailscale/tailscale-www/pull/1896, we wrote that:
```
For traffic from a node to a public device via an exit node, the source will be the Tailscale IP address, but the source port,
source protocol and destination will be empty. For traffic responses from a public device to a node via an exit node, the
destination will be the Tailscale IP address, but the destination port, destination protocol, and source will be empty.
```
Should we be capturing stats for both directions here?

dsnet:
To me those two paragraphs are more or less saying the same thing. There's some oddity about "source protocol" which should really just be "protocol", but generally they agree with each other.

The information captured on a node using an exit node and the information captured on a node acting as an exit node are more or less equivalent. The source and destination are inverted due to their inverted roles.

ramyfication:
> The information captured on a node using an exit node and the information captured on a node acting as an exit node are more or less equivalent. The source and destination are inverted due to their inverted roles.

For the node acting as an exit node, the source of traffic flowing through it is the node with the Tailscale IP, and it sends traffic to some external IP. To me, it doesn't look like the source & destination should be inverted.

dsnet:
For an exit node, it is acting on behalf the public internet. So the source address will end up being some public IP address, while the destination will be the Tailscale IP address that initiated the connection.

In general, the source is the node itself or something it is acting on behalf of (e.g., an external IP address in this case).

ramyfication:
Ack. My mental model was that we were capturing data about traffic leaving the exit node, so I had it reversed.

ramyfication (approved):

dsnet:
No worries. @dblohm7 and I were both very confused initially. You have to just play around with it and then you start to understand how things work by experience rather than by memorization.

PR #6174, opened 2022-11-02
tailcfg: add Node.UnsignedPeerAPIOnly to let server mark node as peerapi-only

bradfitz:


bradfitz:
(gotta run for now, tests later)

ipn/ipnlocal/local.go
```diff
@@ -1303,6 +1303,11 @@ func (b *LocalBackend) updateFilterLocked(netMap *netmap.NetworkMap, prefs ipn.P
 			localNetsB.AddPrefix(p)
 		}
 		packetFilter = netMap.PacketFilter
+
+		if packetFilterPermitsUnlockedNodes(netMap.Peers, packetFilter) {
+			b.logf("WARNING: server sent invalid packet filter permitting traffic to unlocked nodes; rejecting all packets now")
```
danderson, line 1308: Should probably be a health report too, since hitting this condition breaks the data plane.

bradfitz:
PTAL. Should be ready now.

ipn/ipnlocal/network-lock.go
```diff
@@ -49,27 +50,33 @@ func (b *LocalBackend) tkaFilterNetmapLocked(nm *netmap.NetworkMap) {
 		return // TKA not enabled.
 	}
 
-	toDelete := make(map[int]struct{}, len(nm.Peers))
+	var toDelete map[int]bool // peer index => true
 	for i, p := range nm.Peers {
+		if p.UnsignedPeerAPIOnly {
+			// No subject to TKA.
```
twitchyliquid64, line 56: s/No/Not

ipn/ipnlocal/local.go
```diff
@@ -1303,6 +1303,14 @@ func (b *LocalBackend) updateFilterLocked(netMap *netmap.NetworkMap, prefs ipn.P
 			localNetsB.AddPrefix(p)
 		}
 		packetFilter = netMap.PacketFilter
+
+		if packetFilterPermitsUnlockedNodes(netMap.Peers, packetFilter) {
+			err := errors.New("server sent invalid packet filter permitting traffic to unlocked nodes; rejecting all packets for safety")
```
twitchyliquid64, line 1308: Should we reference the node name or ID in the error so we have a thread to pull when debugging?

ipn/ipnlocal/local.go
```diff
@@ -1375,6 +1383,45 @@ func (b *LocalBackend) updateFilterLocked(netMap *netmap.NetworkMap, prefs ipn.P
...
+	if err != nil {
+		// Shouldn't happen, but if it does, fail closed.
+		return true
+	}
+	for _, m := range packetFilter {
+		for _, r := range m.Srcs {
+			if !s.OverlapsPrefix(r) {
+				continue
+			}
+			if len(m.Dsts) != 0 {
```
twitchyliquid64, line 1417: If Im reading this right: A packet filter with no `Dsts` term implies that it can hit netstack (ala PeerAPI and friends) but not get forwarded to the system?

ipn/ipnlocal/local_test.go
```diff
@@ -638,3 +639,86 @@ func TestInternalAndExternalInterfaces(t *testing.T) {
...
+		},
+		{
+			name: "unsigned-good",
+			peers: []*tailcfg.Node{
+				{ID: 1, UnsignedPeerAPIOnly: true},
+			},
+			want: false,
+		},
+		{
+			name: "unsigned-bad",
```
twitchyliquid64, line 669: For completeness, maybe add a 'bad' case where the `filter.Match.Srcs` is a superset of the peer's `AllowedIP` ?

tailcfg/tailcfg.go
```diff
@@ -231,6 +231,14 @@ type Node struct {
 	//    "https://tailscale.com/cap/file-sharing"
 	Capabilities []string `json:",omitempty"`
 
+	// UnsignedPeerAPIOnly means that this node is not signed nor subject to TKA
+	// restrictions. However, in exchange for that privilege, it does not get
+	// network access. It can only access this node's peerapi, which may not let
+	// it do anything. It is the tailscaled client's job to double-check the
+	// MapResponse's PacketFilter to verify that its AllowedIPs will not be
+	// accepted by the packet filter.
+	UnsignedPeerAPIOnly bool `json:",omitempty"`
```
twitchyliquid64, line 240: Do we need to bump capver so control knows its safe to do this, or gonna do the bump all at once for the feature this is dependent on?

twitchyliquid64 (approved):

ipn/ipnlocal/local.go
bradfitz: I'd rather keep it simple in the client. If we ever see this, we can run this same code as needed server-side but fleshed out to include debug info.

ipn/ipnlocal/local.go
bradfitz: Peerapi is intercepted before the packet filter.

tailcfg/tailcfg.go
bradfitz: Integers are cheap so I could do it now I guess, but I was going to wait.

But done.

ipn/ipnlocal/local_test.go
bradfitz: done

bradfitz:
PTAL

ipn/ipnlocal/local.go
twitchyliquid64: Is there anything else that happens before the packet filter that we need to make sure gets dropped for `UnsignedPeerAPIOnly` nodes? Anything something something exit node something DNS something SSH? (not sure if that stuff is also before the packet filter or not).

twitchyliquid64:
Looking through netstack.go it does seem that SSH et al is also before the packet filter, ala in the same intercept path as peer API. I'm a little uncomfortable about a network-lock bypass being that close to administrative access packet paths.

Could we either:
1. Add a specialcase DROP in the `acceptTCP` path for SSH and/or HTTP, or
2. (If its not too late to change tack) specialize `UnsignedPeerAPIOnly` to indicate which peerAPI services are permissible?

bradfitz:
@twitchyliquid64, start reading a bit earlier.

Read `net/tstun/wrap.go`, in `(*Wrapper).filterIn`. That's where packets arrive, in all cases (via `(*Wrapper).Write`).

Reading in order:

* TSMP pings would still be allowed from unsigned peers. That's fine. That's how you test if a node is present, and also the bootstrappy/old way of discovering their peerapi port(s)

* ICMP echo responses would still be handled, if it's a response to one we sent out.

* PreFilterIn is only used for outbound connection failure tracking, to see if our outbound TCP connections ever got SYNs back. That's harmless, like the ICMP responses. They both have to line up with something we sent out.

* then the packet filter is loaded

* if no packet filter, packet is dropped

* packet filter is evaluated for that packet, saved to `outcome`. (remember: lack of rules means reject, and this PR verifies there are no allowed rules)

* if `outcome` wasn't an `Accept` and we determine it's TCP and was for our peerapi port, we change the outcome to `Accept`. But only for the peerapi port.

* if `outcome` still isn't `Accept` after the peerapi check, it's dropped.

* finally we come to `PostFilterIn`, which injects it into netstack, the code you were reading that handles port 22, etc.

But it can't make it to SSH if you didn't have packet filter rules allowing it. (as we document in our KB, https://tailscale.com/kb/1193/tailscale-ssh/, Tailscale SSH requires both network level ACLs, "Ensure an ACL exists that allows the source to connect to the destination on port 22", and SSH policy rules)

When I said earlier that "Peerapi is intercepted before the packet filter.", that wasn't entirely accurate by some interpretations. The packet filter still runs. It's just that its decision is discarded specifically in the case of TCP connections to the peerapi port. But that's an implementation/optimization detail.

In any case, netstack SSH happens much later.

ipn/ipnlocal/local.go
bradfitz: See comment below.

twitchyliquid64 (approved):
My goodness. Thanks for digging into all that to put me at ease :sweat_smile: that makes sense.

LGTM. I'm going to add to my tka backlog to have a e2e test for the cases where packets shouldn't make it in, for future me's peace of mind XD

bradfitz:
Yeah, a test would be great.

On my walk tonight I was realizing how even the tstest/integration tests would be pretty easy to extend to test this.

twitchyliquid64:
I've been going ham on the ipn tests as of late - I'm quite partial to them + can test a lot of things that way. 

Ive not seen `packet.TailscaleRejectedHeader` before, looking for that reply might be a good candidate for testing ACL denial in such a test (that way the test wont pass if the packet path is broken). Things to think about another day :grinning:

PR #6176, opened 2022-11-03
wgengine/netstack: remove old unused handleSSH hook

bradfitz:
It's leftover from an earlier Tailscale SSH wiring and I forgot to delete it apparently.


catzkorn (approved):
Stampy stamp

PR #6183, opened 2022-11-03
util/winutil: remove log spam for missing registry keys

sailorfrag:
It's normal for HKLM\SOFTWARE\Policies\Tailscale to not exist but that currently produces a lot of log spam.

### DIFF
--- a/util/winutil/winutil_windows.go
+++ b/util/winutil/winutil_windows.go
@@ -82,7 +82,9 @@ func getRegInteger(name string, defval uint64) uint64 {
 func getRegStringInternal(subKey, name string) (string, error) {
 	key, err := registry.OpenKey(registry.LOCAL_MACHINE, subKey, registry.READ)
 	if err != nil {
-		log.Printf("registry.OpenKey(%v): %v", subKey, err)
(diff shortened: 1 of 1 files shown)


util/winutil/winutil_windows.go
```diff
@@ -148,6 +152,9 @@ func DeleteRegValue(name string) error {
 
 func deleteRegValueInternal(subKey, name string) error {
 	key, err := registry.OpenKey(registry.LOCAL_MACHINE, subKey, registry.SET_VALUE)
+	if err == registry.ErrNotExist {
```
DentonGentry, line 155: This changes the semantics, it would have returned an error. Presumably the caller already suppressed errors, does the caller need to pay attention to whatever other errors might be returned by line 160?

DentonGentry (approved):

util/winutil/winutil_windows.go
sailorfrag: The two places it's used actually do check for errors and will abort their cleanups if there is an error. I think this actually improves their semantics since in a cleanup it's fine if the thing being deleted doesn't exist.

PR #6188, opened 2022-11-03
tailcfg: bump CurrentCapabilityVersion for EarlyNoise

maisem:


bradfitz (approved):
Make the commit message link to the issue so we can find the related commits that led up to this?

PR #6198, opened 2022-11-04
tsnet/example/tshello: use strings.Cut

Xe:
strings.Cut allows us to be more precise here. This example was written before strings.Cut existed.

### DIFF
--- a/tsnet/example/tshello/tshello.go
+++ b/tsnet/example/tshello/tshello.go
@@ -57,8 +57,6 @@ func main() {
 }
 
 func firstLabel(s string) string {
-	if i := strings.Index(s, "."); i != -1 {
-		return s[:i]
-	}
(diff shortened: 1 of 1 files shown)


tsnet/example/tshello/tshello.go
```diff
@@ -57,8 +57,8 @@ func main() {
 }
 
 func firstLabel(s string) string {
-	if i := strings.Index(s, "."); i != -1 {
-		return s[:i]
+	if hostname, _, ok := strings.Cut(s, "."); ok {
+		return hostname
 	}
 	return s
```
bradfitz, line 63: Shorter:

```go
func firstLabel(s string) string {
    s, _, _ = strings.Cut(s, ".")
    return s
}
```

bradfitz (approved):

PR #6199, opened 2022-11-04
portlist: add macOS osImpl, finish migration to new style

bradfitz:
Previously:

* 036f70b7b4 for linux
* 35bee36549 for windows

This does macOS.

And removes all the compat code for the old style. (e.g. iOS, js are no longer mentioned; all platforms without implementations just default to not doing anything)

One possible regression is that platforms without explicit implementations previously tried to do the "netstat -na" style to get open ports (but not process names). Maybe that worked on FreeBSD and OpenBSD previously, but nobody ever really tested it. And it was kinda useless without associated process names. So better off removing those for now until they get a good implementation.


portlist/netstat.go
```diff
@@ -2,21 +2,23 @@
...
-	"strings"
+	"bufio"
+	"bytes"
+	"io"
+
+	"go4.org/mem"
 )
 
-func parsePort(s string) int {
+func parsePort(s mem.RO) int {
```
danderson, line 17: Not added in this PR, but: any chance of a docstring that explains why not net.SplitHostPort? I believe the answer is because we also have to parse `host.port` format (is that still true now that this is macOS only?).

portlist/netstat.go
```diff
@@ -41,50 +43,57 @@ func parsePort(s string) int {
...
-	lines := strings.Split(string(output), "\n")
-
-	var lastline string
-	var lastport Port
-	for _, line := range lines {
-		trimline := strings.TrimSpace(line)
-		cols := strings.Fields(trimline)
+// Nowadays, though, we only use it for macOS as of 2022-11-04.
+func appendParsePortsNetstat(base []Port, br *bufio.Reader) ([]Port, error) {
+	ret := base
```
danderson, line 61: Add some example lines in a comment, to make walking the parser logic easier?

danderson: Oh, or a comment pointing to tests for example inputs would do.

portlist/portlist_linux.go
```diff
@@ -27,6 +27,8 @@ import (
 
 func init() {
 	newOSImpl = newLinuxImpl
+	// Reading the sockfiles on Linux is very fast, so we can do it often.
```
danderson, line 30: Fast might still cost a bunch of battery on laptops? Is the increased resolution worth it?

danderson (approved):

portlist/netstat.go
bradfitz: Documented.

portlist/netstat.go
bradfitz: Done.

PR #6206, opened 2022-11-05
util/dirwalk, metrics, portlist: add new package for fast directory walking

bradfitz:
This is similar to the golang.org/x/tools/internal/fastwalk I'd previously written but not recursive and using mem.RO.

The metrics package already had some Linux-specific directory reading code in it. Move that out to a new general package that can be reused by portlist too, which helps its scanning of all /proc files:

    name                old time/op    new time/op    delta
    FindProcessNames-8    2.79ms ± 6%    2.45ms ± 7%  -12.11%  (p=0.000 n=10+10)

    name                old alloc/op   new alloc/op   delta
    FindProcessNames-8    62.9kB ± 0%    33.5kB ± 0%  -46.76%  (p=0.000 n=9+10)

    name                old allocs/op  new allocs/op  delta
    FindProcessNames-8     2.25k ± 0%     0.38k ± 0%  -82.98%  (p=0.000 n=9+10)


bradfitz:
(Weekend, so not adding any reviewers, but feel free if you're bored)

maisem (approved):

Issue #6207, opened 2022-11-06
wgengine/magicsock: NewConn test flake


bradfitz:
Bunch of these lately from github actions PR test runs. This is at bb2cba0cd1b0 ....

```
--- FAIL: TestNewConn (2.58s)
    magicsock_linux.go:307: magicsock: failed to force-set UDP read buffer size to 7340032: operation not permitted
    magicsock_linux.go:311: magicsock: failed to force-set UDP write buffer size to 7340032: operation not permitted
    magicsock_linux.go:307: magicsock: failed to force-set UDP read buffer size to 7340032: operation not permitted
    magicsock_linux.go:311: magicsock: failed to force-set UDP write buffer size to 7340032: operation not permitted
    magicsock.go:635: [v1] couldn't create raw v4 disco listener, using regular listener instead: raw disco listening disabled, SO_MARK unavailable
    magicsock.go:641: [v1] couldn't create raw v6 disco listener, using regular listener instead: raw disco listening disabled, SO_MARK unavailable
    magicsock.go:2273: magicsock: SetPrivateKey called (init)
    logger.go:99: netcheck: [v1] checkCaptivePortal: Get "http://d1.invalid/generate_204": context deadline exceeded
    logger.go:99: netcheck: [v1] report: udp=false v4=false icmpv4=false v6=false v6os=true mapvarydest= hair= portmap=? derp=0
    magicsock.go:1056: magicsock: home is now derp-1 ()
    magicsock.go:2580: magicsock: endpoints changed: 172.31.73.194:38106 (local)
    magicsock.go:1403: magicsock: adding connection to derp-1 for home-keep-alive
    magicsock.go:2568: magicsock: 1 active derp conns: derp-1=cr0s,wr0s
    magicsock.go:2557: magicsock: closing connection to derp-1 (conn-close), age 0s
    magicsock.go:2568: magicsock: 0 active derp conns
    logger.go:99: netcheck: [v1] report: udp=false v4=false icmpv4=false v6=false v6os=true mapvarydest= hair= portmap=? derp=0
    stuntest.go:64: STUN server shutdown
    resource.go:36: goroutine diff:
          []uint8(
          	"""
        - 	goroutine profile: total 2
        - 	1 @ 0x407fc 0x76414 0x1af364 0x1af180 0x1ac78c 0x5424e4 0x5420d0 0x5730fc 0x12249c 0x7c814
        + 	goroutine profile: total 3
[...]

PR #6209, opened 2022-11-06
net/netcheck: deflake (maybe) magicsock's TestNewConn

bradfitz:
Updates #6207

### DIFF
--- a/net/netcheck/netcheck.go
+++ b/net/netcheck/netcheck.go
@@ -19,6 +19,7 @@ import (
 	"net/netip"
 	"runtime"
 	"sort"
+	"strings"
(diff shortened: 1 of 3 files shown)


maisem (approved):
Sure, but how did it ever work?

bradfitz:
The test is failing because of an extra goroutine leftover at the end of the test. (in particular, a cgo DNS lookup)

The captive portal probe times out. But if its context cancellation doesn't plumb all the way (like into libc via cgo) then it has lingering goroutines after the test is done and checks for goroutine leaks.

GitHub Actions is just nice & slow and helps find things like that.

PR #6217, opened 2022-11-07
cmd/containerboot: don't write device ID into non-existent secret.

### LINKED ISSUE #6211: containerboot doesn't handle `--state=` and `--statedir=` simultaneously
Some sites use `run.sh` with a `--statedir=` argument to handle where to store TLS certificates and a `--state=` argument to put the tailscaled.state in the k8s secrets manager.

--------

_Originally posted by @OGKevin in https://github.com/tailscale/tailscale/issues/4837#issuecomment-1304764069_

I run the following sidecar: 

```
  ts-sidecar:
    Container ID:   containerd://e87205e22c690d9011c4d52166dcd26632d02ce92cc77cfac69701e76dea58cc
    Image:          ghcr.io/tailscale/tailscale:v1.33.132@sha256:766629890b2ce668b78ab880d813ea3059f85d634ec6ba079014bcb2d9d1316a
    Image ID:       ghcr.io/tailscale/tailscale@sha256:766629890b2ce668b78ab880d813ea3059f85d634ec6ba079014bcb2d9d1316a
    Port:           <none>
    Host Port:      <none>
    State:          Running
      Started:      Sun, 06 Nov 2022 10:53:52 +0100
    Ready:          True
    Restart Count:  0
    Environment:
      TS_TAILSCALED_EXTRA_ARGS:  --state=mem
      TS_USERSPACE:              false
      TS_AUTH_KEY:               <set to the key 'TS_AUTH_KEY' in secret 'tailscale-auth'>  Optional: false
```

Which ran fine, until upgrading to ` v1.33.141`. 

```
2022/11/06 09:39:23 Program starting: v1.33.141-tbb2cba0cd, Go 1.19.2-ts3fd24dee31: []string{"tailscaled", "--socket=/tmp/tailscaled.sock", "--state=kube:tailscale", "--statedir=/tmp", "--state=mem"}
```

Would still try to store the state in k8s secret, while `v1.33.132` seemed to ignore the first appearance of `--state` arg. 
Unfortunately this is hardcoded https://github.com/tailscale/tailscale/blob/da8def8e13d44df0c4183300c826523ba933cb37/cmd/containerboot/main.go#L265 😔 

I did not check the changelog between these 2 versions though.
But worst case, I would need to create a fork and remove the hardcoded state arg.

danderson:
Fixes #6211

### DIFF
--- a/cmd/containerboot/kube.go
+++ b/cmd/containerboot/kube.go
@@ -69,6 +69,23 @@ func findKeyInKubeSecret(ctx context.Context, secretName string) (string, error)
 // the kube secret secretName.
 func storeDeviceID(ctx context.Context, secretName, deviceID string) error {
 	kubeOnce.Do(initKube)
+
+	// First check if the secret exists at all. Even if running on
+	// kubernetes, we do not necessarily store state in a k8s secret.
+	req, err := http.NewRequest("GET", fmt.Sprintf("/api/v1/namespaces/%s/secrets/%s", kubeNamespace, secretName), nil)
+	if err != nil {
+		return err
+	}
+	resp, err := doKubeRequest(ctx, req)
+	if err != nil {
+		if resp != nil && resp.StatusCode >= 400 && resp.StatusCode <= 499 {
+			// Assume the secret doesn't exist, or we don't have
+			// permission to access it.
+			return nil
+		}
+		return err
+	}
+
 	m := map[string]map[string]string{
 		"stringData": map[string]string{
 			"device_id": deviceID,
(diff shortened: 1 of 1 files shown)


cmd/containerboot/kube.go
```diff
@@ -69,6 +69,23 @@ func findKeyInKubeSecret(ctx context.Context, secretName string) (string, error)
...
+
+	// First check if the secret exists at all. Even if running on
+	// kubernetes, we do not necessarily store state in a k8s secret.
+	req, err := http.NewRequest("GET", fmt.Sprintf("/api/v1/namespaces/%s/secrets/%s", kubeNamespace, secretName), nil)
+	if err != nil {
+		return err
+	}
+	resp, err := doKubeRequest(ctx, req)
+	if err != nil {
+		if resp != nil && resp.StatusCode/100 == 4 {
```
andrew-d, line 81: This `resp.StatusCode/100 == 4` is ... unusual, and I did a bit of a double-take; maybe something like `resp.StatusCode >= 400 && resp.StatusCode <= 499`?

andrew-d (approved):

cmd/containerboot/kube.go
danderson: Yeah, I felt clever at the time, but in retrospect it's not worth the keystroke savings.

PR #6220, opened 2022-11-07
cmd/tailscale, ipn/ipnlocal: add debug command to write to StateStore for dev

bradfitz:
Not for end users (unless directed by support). Mostly for ease of development for some upcoming webserver work.

### DIFF
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -61,6 +61,7 @@ var handler = map[string]localAPIHandler{
 	"component-debug-logging": (*Handler).serveComponentDebugLogging,
 	"debug":                   (*Handler).serveDebug,
 	"derpmap":                 (*Handler).serveDERPMap,
+	"dev-set-state-store":     (*Handler).serveDevSetStateStore,
 	"dial":                    (*Handler).serveDial,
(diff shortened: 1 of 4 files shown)


ipn/localapi/localapi.go
```diff
@@ -401,6 +402,23 @@ func (h *Handler) serveDebug(w http.ResponseWriter, r *http.Request) {
...
+	if !h.PermitWrite {
+		http.Error(w, "debug access denied", http.StatusForbidden)
+		return
+	}
+	if r.Method != "POST" {
+		http.Error(w, "POST required", http.StatusMethodNotAllowed)
+		return
+	}
+	if err := h.b.SetDevStateStore(r.FormValue("key"), r.FormValue("value")); err != nil {
+		http.Error(w, err.Error(), 500)
```
danderson, line 415: http.StatusInternalServerError?

danderson (approved):

ipn/localapi/localapi.go
bradfitz: Well, the rest of this file is all ", 500" like this and ", 400".

So I'll keep it like this for now for consistency and we can convert the whole file in its own pass.

PR #6224, opened 2022-11-07
ipn/ipnlocal, wgengine/netstack: start handling ports for future serving

bradfitz:


shayne:
Looks like we need to address the fast path optimization in `shouldProcessInbound`. In testing the PR my request is dropped on L562

https://github.com/tailscale/tailscale/blob/7b52d8ae3987abef75610898772e9786b232a3dc/wgengine/netstack/netstack.go#L559-L563

shayne:
> Looks like we need to address the fast path optimization in `shouldProcessInbound`. In testing the PR my request is dropped on L562
> 
> https://github.com/tailscale/tailscale/blob/7b52d8ae3987abef75610898772e9786b232a3dc/wgengine/netstack/netstack.go#L559-L563

Addressed by special casing 443 for now

ipn/ipnlocal/local.go
```diff
@@ -1956,13 +1961,44 @@ func (b *LocalBackend) loadStateLocked(key ipn.StateKey, prefs *ipn.Prefs) (err
...
+		f = func(p uint16) bool { return ports[0] == p || ports[1] == p || ports[2] == p }
+	default:
+		f = func(p uint16) bool {
+			for _, x := range ports {
+				if p == x {
+					return true
+				}
+			}
+			return false
+		}
```
andrew-d, line 1989: Why a loop instead of a `map`?

ipn/ipnlocal/local.go
```diff
@@ -1956,13 +1961,44 @@ func (b *LocalBackend) loadStateLocked(key ipn.StateKey, prefs *ipn.Prefs) (err
...
+}
+
 // setAtomicValuesFromPrefs populates sshAtomicBool and containsViaIPFuncAtomic
 // from the prefs p, which may be nil.
 func (b *LocalBackend) setAtomicValuesFromPrefs(p ipn.PrefsView) {
 	b.sshAtomicBool.Store(p.Valid() && p.RunSSH() && envknob.CanSSHD())
 
 	if !p.Valid() {
 		b.containsViaIPFuncAtomic.Store(tsaddr.NewContainsIPFunc(nil))
+		b.shouldInterceptTCPPortAtomic.Store(func(uint16) bool { return false })
```
andrew-d, line 2001: Why not `	b.shouldInterceptTCPPortAtomic.Store(nil)`?

ipn/ipnlocal/local.go
```diff
@@ -3434,6 +3471,21 @@ func (b *LocalBackend) setNetMapLocked(nm *netmap.NetworkMap) {
 			delete(b.nodeByAddr, k)
 		}
 	}
+
+	handlePorts := make([]uint16, 0, 4)
+	if b.ShouldRunSSH() {
+		handlePorts = append(handlePorts, 22)
+	}
+	for _, domain := range nm.DNS.CertDomains {
+		if v, err := b.store.ReadState(ipn.StateKey(fmt.Sprintf("_handle-tcp/%s", domain))); err == nil {
```
maisem, line 3480: I don't know what you are doing here with the state store. This needs comments.

ipn/ipnlocal/local.go
bradfitz: so caller can blindly call the func. we do this this atomic-func-documented-non-nil thing in a few places

bradfitz: but this should be setTCPPortsIntercepted(nil)

ipn/ipnlocal/local.go
bradfitz: For now only the CLI debug command. The CLI will come later.

The format is described only in https://docs.google.com/document/d/1c09VN98W77mf-qpZKD8tukqn7QNvT68kRaoqLsd8byE/edit# for now.

maisem: rest of the code is fine, but I think this should go in the same PR as the thing that writes this state.

bradfitz: done

ipn/ipnlocal/local.go
bradfitz: added a cutover for maps too

ipn/ipnlocal/local.go
maisem: Thanks, I can reason about this now!

ipn/store.go
```diff
@@ -64,3 +64,14 @@ func ReadStoreInt(store StateStore, id StateKey) (int64, error) {
...
 }
+
+// HandleTCPStateKeyForDomain returns a StateKey that stores the
+// whitespace-separated list of TCP ports that tailscaled should handle for the
+// given domain name (such as "foo.donkey-fish.ts.net")
+//
+// The returned value from the state store might be "443 8080", for example.
+// How the port is handled by tailscaled is subject to other StateKey
+// lookups. TODO(bradfitz): document those, once they're fleshed out.
+func HandleTCPStateKeyForDomain(domain string) StateKey {
```
maisem, line 75: Can you put this in prefs instead? I would rather we don't add more global keys with FUS.

bradfitz: It's qualified by the TCD name. I think it'd be pretty gross in prefs.

ipn/ipnlocal/local.go
andrew-d: That's what I meant and I apparently typo'd :+1:

ipn/ipnlocal/local.go
```diff
@@ -3976,5 +4042,59 @@ func (b *LocalBackend) SetDevStateStore(key, value string) error {
...
+	hs := &http.Server{
+		TLSConfig: &tls.Config{
+			GetCertificate: b.getTLSServeCert,
+		},
+		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			io.WriteString(w, "<h1>hello world</h1>this is tailscaled")
+		}),
+	}
+	hs.ServeTLS(netutil.NewOneConnListener(c, nil), "", "")
+}
```
shayne, line 4083: Just a heads up ran into an issue in WSL where there connection would hang in TLS handshake. Works fine for me under normal Linux.

bradfitz: Probably the good ol' MTU issues. Initial TLS ServerHellos are big and frequently trigger the existing MTU problems.

ipn/store.go
bradfitz: Updated to JSON with single key in StateStore, per call

PR #6226, opened 2022-11-08
api.md: make it clearer where to get the tailnet name in API calls

mihaip:
We added the tailnet organization name to to the settings page with tailscale/corp#6977, but the docs were not updated to reflect this. We later also changed "tailnet name" to refer to the MagicDNS hostname (tailscale/corp#7537), which further confuses things (see https://stackoverflow.com/questions/74132318).

Make it slightly clearer what is the expected value for tailnet names in API calls and how to get it.

### DIFF
--- a/api.md
+++ b/api.md
@@ -335,11 +335,12 @@ The response is 2xx on success. The response body is currently an empty JSON
 object.
 
 ## Tailnet
-A tailnet is the name of your Tailscale network.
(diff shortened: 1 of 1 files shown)


mihaip:
@willnorris you touched this last. Perhaps we should make the `-` magic value more prominent/lead with that?

willnorris (approved):

willnorris:
yeah, we could certainly do that.  For the foreseeable future where a user can only act on their own tailnet, pushing users to just always use `-` simplifies things a lot.

PR #6230, opened 2022-11-08
ipn/{ipnlocal,localapi}: move most of cert.go to ipnlocal

bradfitz:
Leave only the HTTP/auth bits in localapi.


bradfitz:
Just code movement, other than:

* rename `type keyPair` to `type TLSCertKeyPair`
* add a TODO about HTTP status codes and errors and tsweb

maisem (approved):

bradfitz:
Latest push: fix accidental exporting of CertDir that I'd done in an earlier attempt and forgot to revert.

And add the `!js` build tag, so this doesn't bloat wasm.

PR #6235, opened 2022-11-08
wgengine/netstack: fix shouldProcessInbound peerapi non-SYN handling

bradfitz:
It was eating TCP packets to peerapi ports to subnet routers.  Some of the TCP flow's packets went onward, some got eaten.  So some TCP flows to subnet routers, if they used an unfortunate TCP port number, got broken.

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -535,23 +535,30 @@ var viaRange = tsaddr.TailscaleViaRange()
 // WireGuard peer) should be handled by netstack.
 func (ns *Impl) shouldProcessInbound(p *packet.Parsed, t *tstun.Wrapper) bool {
 	// Handle incoming peerapi connections in netstack.
-	if ns.lb != nil && p.IPProto == ipproto.TCP {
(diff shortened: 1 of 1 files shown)


wgengine/netstack/netstack.go
```diff
@@ -535,23 +535,30 @@ var viaRange = tsaddr.TailscaleViaRange()
...
 			peerAPIPort = uint16(atomic.LoadUint32(ns.peerAPIPortAtomic(dstIP)))
 		}
-		if p.IPProto == ipproto.TCP && p.Dst.Port() == peerAPIPort {
+		dport := p.Dst.Port()
+		if dport == peerAPIPort {
+			return true
+		}
+
+		// Also handle SSH connections, if enabled.
+		if dport == 22 && ns.lb.ShouldRunSSH() {
```
shayne, line 559: can we use `isInboundTSSH(p *packet.Parsed)` here?

bradfitz: I deleted it because all its checks were redundant with what this block had already checked.

PR #6239, opened 2022-11-08
net/netcheck: don't use a space in the captive portal challenge

andrew-d:
The derpers don't allow whitespace in the challenge.

Change-Id: I93a8b073b846b87854fba127b5c1d80db205f658

### DIFF
--- a/net/netcheck/netcheck.go
+++ b/net/netcheck/netcheck.go
@@ -1126,7 +1126,10 @@ func (c *Client) checkCaptivePortal(ctx context.Context, dm *tailcfg.DERPMap, pr
 		return false, err
 	}
 
-	chal := "tailscale " + node.HostName
+	// Note: the set of valid characters in a challenge and the total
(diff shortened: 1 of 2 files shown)


net/netcheck/netcheck.go
```diff
@@ -1126,7 +1126,9 @@ func (c *Client) checkCaptivePortal(ctx context.Context, dm *tailcfg.DERPMap, pr
 		return false, err
 	}
 
-	chal := "tailscale " + node.HostName
+	// Note: the set of valid characters in a challenge and the total
+	// length is limited; see the derper code for more details
```
mihaip, line 1130: Maybe explicitly mention `isChallengeChar` as where this is enforced?

mihaip (approved):

net/netcheck/netcheck.go
andrew-d: Good call; done!

bradfitz (approved):

PR #6242, opened 2022-11-09
wgengine/netstack: optimize shouldProcessInbound, avoiding 4via6 lookups

bradfitz:
All IPv6 packets for the self address were doing netip.Prefix.Contains lookups.

If if we know they're for a self address (which we already previously computed and have sitting in a bool), then they can't be for a 4via6 range.

### DIFF
--- a/wgengine/netstack/netstack_test.go
+++ b/wgengine/netstack/netstack_test.go
@@ -259,6 +259,13 @@ func TestShouldHandlePing(t *testing.T) {
 	}
 }
 
+// looksLikeATailscaleSelfAddress reports whether addr looks like
+// a Tailscale self address, for tests.
+func looksLikeATailscaleSelfAddress(addr netip.Addr) bool {
+	return addr.Is4() && tsaddr.IsTailscaleIP(addr) ||
+		addr.Is6() && tsaddr.Tailscale4To6Range().Contains(addr)
+}
+
 func TestShouldProcessInbound(t *testing.T) {
 	testCases := []struct {
 		name  string
@@ -289,6 +296,7 @@ func TestShouldProcessInbound(t *testing.T) {
 					StateKey:    ipn.GlobalDaemonStateKey,
 					UpdatePrefs: prefs,
 				})
+				i.atomicIsLocalIPFunc.Store(looksLikeATailscaleSelfAddress)
 
 				// This should be handled even if we're
 				// otherwise not processing local IPs or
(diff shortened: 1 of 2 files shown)


bradfitz:
(Overkill number of reviewers, but people who've been looking at this code recently and Denton who likes 4via6 ... :))

wgengine/netstack/netstack_test.go
```diff
@@ -259,6 +259,13 @@ func TestShouldHandlePing(t *testing.T) {
 	}
 }
 
+// looksLikeATailscaleSelfAddress reports whether addr looks like
+// a Tailscale self address, for tests.
+func looksLikeATailscaleSelfAddress(addr netip.Addr) bool {
+	return addr.Is4() && tsaddr.IsTailscaleIP(addr) ||
```
DentonGentry, line 265: As written, an IPv6 Tailscale non-4via6 address will return false. I guess I'd expect this to be: `tsaddr.IsTailscaleIP(addr) || addr.Is6() && tsaddr.Tailscale4To6Range().Contains(addr)`

DentonGentry (approved):

wgengine/netstack/netstack_test.go
bradfitz: I don't think it will?

Looks right to me: https://go.dev/play/p/FhaV1vqZxqx

The confusion is likely `tsaddr.IsTailscaleIP`: it's not "Is a Tailscale self IP address", but:

```go
// IsTailscaleIP reports whether ip is an IP address in a range that
// Tailscale assigns from.
func IsTailscaleIP(ip netip.Addr) bool {
	if ip.Is4() {
		return CGNATRange().Contains(ip) && !ChromeOSVMRange().Contains(ip)
	}
	return TailscaleULARange().Contains(ip)
}

// TailscaleULARange returns the IPv6 Unique Local Address range that
// is the superset range that Tailscale assigns out of.
func TailscaleULARange() netip.Prefix {
	tsUlaRange.Do(func() { mustPrefix(&tsUlaRange.v, "fd7a:115c:a1e0::/48") })
	return tsUlaRange.v
}
```

.... the whole ULA /48 range, not just the bit we used to translate CGNAT v4s to v6.

DentonGentry: Oh: 4To6 and 4Via6 are two different things.

I interpreted `tsaddr.Tailscale4To6Range().Contains(addr)` as checking for a 4via6 address and didn't see how a regular Tailscale IPv6 address was handled.

DentonGentry (approved):

PR #6250, opened 2022-11-09
Make it possible to define controlURL for tsnet server

majst01:
If using tsnet for userspace networking based tailscale services, it is currently not possible (at least it is not obvious for me) to override the controlURL aka login-server.

With this PR i am able to do so. I dont know if this is the it is supposed to do, happy to either adapt this PR according your suggestions, or even better learn how to override controlURL in the right way.

All in all, using tsnet is pure magic, thanks for this, it opens so much new solutions no one could think of before.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -82,6 +82,13 @@ type Server struct {
 	// used.
 	AuthKey string
 
+	// ControlURL is explicitly not set to signal that
+	// it's not yet configured, which relaxes the CLI "up"
+	// safety net features. It will get set to DefaultControlURL
+	// on first up. Or, if not, DefaultControlURL will be used
+	// later anyway. Can be overridden with TS_CONTROL_URL environment variable.
+	ControlURL string
+
 	initOnce         sync.Once
 	initErr          error
 	lb               *ipnlocal.LocalBackend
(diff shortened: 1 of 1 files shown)


tsnet/tsnet.go
```diff
@@ -82,6 +82,13 @@ type Server struct {
 	// used.
 	AuthKey string
 
+	// ControlURL is explicitly not set to signal that
+	// it's not yet configured, which relaxes the CLI "up"
+	// safety net features. It will get set to DefaultControlURL
+	// on first up. Or, if not, DefaultControlURL will be used
+	// later anyway. Can be overridden with TS_CONTROL_URL environment variable.
```
bradfitz, line 89: This isn't quite the normal doc style we/Go use. Start by saying what it even is.

```
// ControlURL optionally specifies the coordination server URL.
//
// If empty, the TS_CONTROL_URL environment variable is used. If that's empty,
// the Tailscale default is used.
```

tsnet/tsnet.go
```diff
@@ -338,6 +345,12 @@ func (s *Server) start() (reterr error) {
 	prefs := ipn.NewPrefs()
 	prefs.Hostname = s.hostname
 	prefs.WantRunning = true
+	if os.Getenv("TS_CONTROL_URL") != "" {
+		prefs.ControlURL = os.Getenv("TS_CONTROL_URL")
+	}
+	if s.ControlURL != "" {
+		prefs.ControlURL = s.ControlURL
+	}
```
bradfitz, line 353: } else {
     prefs.ControlURL = os.Getenv("TS_CONTROL_URL")
}

bradfitz:
Also, DCO. And in commit message, match the existing style (see git log for examples, or https://github.com/golang/go/wiki/CommitMessage)

Like: `tsnet: add Server.ControlURL option`

bradfitz:
I'm cleaning up PRs so I sent https://github.com/tailscale/tailscale/pull/6318 with the minimal change for this, without the environment variable bit.

majst01:
Hi @bradfitz thanks for the effort !

bradfitz:
no prob!

Issue #3186, opened 2021-10-25
Windows `--unattended` doesn't start on boot until at least one login


12nick12:
### What is the issue?

Have to run "C:\Program Files (x86)\Tailscale IPN\tailscale.exe" up --login-server=https://scale.domain.com --unattended for it to start.

I was under that understanding that --unattended would make it start automatically. 

### Steps to reproduce

Have to run "C:\Program Files (x86)\Tailscale IPN\tailscale.exe" up --login-server=https://scale.domain.com --unattended for it to start

### OS version

Windows 10

### Tailscale version

1.16.1

jazzdan:
I have an EC2 user data script that looks like this (Pulumi code for reference):

```typescript
  const windowsUserData = `<powershell>
Start-Transcript -Path "C:\\UserData.log" -Append
Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://community.chocolatey.org/install.ps1'))
choco install tailscale -y
& "C:\\Program Files (x86)\\Tailscale IPN\\tailscale.exe" up --authkey=${tailscale_auth_key} --accept-dns --accept-routes --unattended
</powershell>`;

  const windows_build = new aws.ec2.Instance("windows build", {
    ami: "ami-08c213ebdf7b857c2",
    instanceType: "t2.2xlarge",
    associatePublicIpAddress: true,
    tags: {
      Name: "pulumi windows build",
    },
    rootBlockDevice: {
      volumeSize: 128,
    },
    keyName: "foo",
    vpcSecurityGroupIds: [sg.id],
    subnetId: publicSubnets[0],
    userData: windowsUserData,
    getPasswordData: true,
  });
```

When I start this EC2 instance the machine briefly appears in my Tailscale list of machines on login.tailscale.com. However it quickly disconnects. When I RDP in to the Windows box I see that Tailscale is running but is not authorized (it asks me to sign in). The logs for the user data script don't show any failures, but also don't show any output from Tailscale which is weird. If I wasn't seeing it connect briefly in the Tailscale console I would assume that my 24 hours of powershell experience has led me down a bad path, but it's clearly working at least temporarily:

```
PS C:\> cat C:\ProgramData\Amazon\EC2-Windows\Launch\Log\UserdataExecution.log
[...]

PR #6255, opened 2022-11-09
ipn/ipnlocal: ensure Persist information is saved to server mode star…

dblohm7:
…t state

Numerous issues have been filed concerning an inability to install and run Tailscale headlessly in unattended mode, particularly after rebooting. The server mode `Prefs` stored in `server-state.conf` were not being updated with `Persist` state once the node had been succesfully logged in.

Users have been working around this by finagling with the GUI to make it force a state rewrite. This patch makes that unnecessary by ensuring the required server mode state is updated when prefs are updated by the control client.

Fixes https://github.com/tailscale/tailscale/issues/3186

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -861,6 +861,7 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 		b.setNetMapLocked(st.NetMap)
 		b.updateFilterLocked(st.NetMap, b.prefs)
 	}
+	userID := b.userID
 	b.mu.Unlock()
 
 	// Now complete the lock-free parts of what we started while locked.
@@ -870,6 +871,8 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 				b.logf("Failed to save new controlclient state: %v", err)
 			}
 		}
+		b.writeServerModeStartState(userID, prefs.View())
+
 		p := prefs.View()
 		b.send(ipn.Notify{Prefs: &p})
 	}


ipn/ipnlocal/local.go
```diff
@@ -870,6 +871,8 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 				b.logf("Failed to save new controlclient state: %v", err)
 			}
 		}
+		b.writeServerModeStartState(userID, prefs.View())
```
bradfitz, line 874: so this is writing to disk in `setClientStatus`, which is basically the pubsub listener of state changes, but requires a bunch of stuff be up and working. So this works, but it's a little late.

All mutations to the unattended bit etc has to come through LocalBackend anyway, so we should be writing to the state store in e.g. SetPrefs, EditPrefs, Start.

(And sorry, you've found what's probably the oldest & worst code in the tree... we've been slowing moving towards EditPrefs and a new generation of APIs that are more explicit and documented but LocalBackend still deals with a bunch of styles.)

Looks like we should probably pull out all the assignments to `b.prefs` into a `setPrefsLocked` and place all the code that needs to react to prefs changes in one place.

e.g. there's a bunch of this sprinkled all over too:

```
	b.setAtomicValuesFromPrefs(b.prefs)
	b.setTCPPortsInterceptedFromNetmapAndPrefsLocked()
```

... whenever `b.prefs` is assigned.

bradfitz (commented):
yay!

ipn/ipnlocal/local.go
bradfitz: Oh, but it's trying to not do I/O with b.mu held, huh?

I guess setClientStatus (being async) is safer in that regard.

Eh, let's do this for now. Cleaning up LocalBackend is its own project.

bradfitz (approved):

crcastle:
@dblohm7 @bradfitz Not sure of the best place to share the below script, but I bumped into this PR when researching my issue.

I wanted to semi-automate Tailscale setup on Azure Windows VMs. I'm using them for testing so need to create and destroy them quickly. I also didn't want to expose the RDP port publicly, so wanted to use Tailscale.

Azure offers a way to easily [run a PowerShell script](https://learn.microsoft.com/en-us/azure/virtual-machines/windows/run-command#azure-portal) on the VM from https://portal.azure.com right after creating the VM. The script can also be run from the Azure CLI if you're using that instead of the web portal.

CleanShot 2024-03-23 at 15 44 25@2x

 Using that to get Tailscale installed and running was more difficult than I had hoped (more because of my lack of Windows knowledge than problems with Tailscale). The crux of the issue was that the script runs as the `NT AUTHORITY\SYSTEM` user instead of the normal user account one would login with. After that user installed and logged into Tailscale, the client would logout within 30 seconds. I think this is related to `--unattended` not persisting but not 100% sure. Moreover, in Windows there doesn't seem to be an easy to use `su` command, and the `NT AUTHORITY\SYSTEM` user seems to be restricted from impersonating other users without the "remote session" method I'm using below.

Anyway, here's the script in case any future Tailscale users need to do something similar. It downloads, installs, and logs into Tailscale given a Tailscale key and windows username and password. This works on Windows 10 and 11. Feel free to add it to any docs.

```powershell
$authkey = "tailscale-auth-key"
$username = 'domain\user'
$password = 'user-password'
$url = "https://pkgs.tailscale.com/stable/tailscale-setup-latest.exe"
$output = "C:\tailscale-setup-latest.exe"

# Disable first login privacy customization stuff to reduce number of clicks at first login
[microsoft.win32.registry]::SetValue("HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Microsoft\Windows\OOBE", "DisablePrivacyExperience", 1)
Get-ItemProperty -Path "HKLM:\SOFTWARE\Policies\Microsoft\Windows\OOBE" -Name "DisablePrivacyExperience"

# Disable browser first run customization stuff to reduce number of clicks at first browser use
[microsoft.win32.registry]::SetValue("HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Internet Explorer\Main", "DisableFirstRunCustomize", 2)
Get-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Internet Explorer\Main" -Name "DisableFirstRunCustomize"
[microsoft.win32.registry]::SetValue("HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Microsoft\Edge", "HideFirstRunExperience", 1)
Get-ItemProperty -Path "HKLM:\SOFTWARE\Policies\Microsoft\Edge" -Name "HideFirstRunExperience"

# Download Tailscale installer (and report duration)
$start_time = Get-Date
Invoke-WebRequest -Uri $url -OutFile $output -UseBasicParsing
Write-Output "Time taken to download: $((Get-Date).Subtract($start_time).Seconds) second(s)"

$securePassword = ConvertTo-SecureString $password -AsPlainText -Force
$credential = New-Object System.Management.Automation.PSCredential $username, $securePassword

# Enable Windows Remote Management so the Invoke-Command below can use a "remote session"
# TODO: determine if this is safe to leave enabled
winrm quickconfig -quiet

# Run Tailscale install and login as a regular windows admin user instead
# of the NT AUTHORITY\SYSTEM user so that the --unattended setting persists
$remoteSession=New-PSSession -credential $credential
Invoke-Command -session $remoteSession -scriptblock {
	C:\Windows\System32\whoami.exe

	# Install Tailscale (and report duration)
	$start_time = Get-Date
	Start-Process -Wait $Using:output -ArgumentList "/install","/quiet"
	Write-Output "Time taken to install: $((Get-Date).Subtract($start_time).Seconds) second(s)"

	# Start Tailscale
	Start-Process -Wait "C:\Program Files\Tailscale\tailscale.exe" -ArgumentList "up","--unattended","--authkey=$Using:authkey"

	sleep 1
	& "C:\Program Files\Tailscale\tailscale.exe" status
}
```

nickkhyl:
Thank you @crcastle! We're also currently working on a bunch of related improvements. Among other things, these would allow you and other users to configure Tailscale more easily from within the LocalSystem's security context, without the need to know and store local user credentials in a PowerShell script or enable WinRM.

hafometh88:
> Thank you @crcastle! We're also currently working on a bunch of related improvements. Among other things, these would allow you and other users to configure Tailscale more easily from within the LocalSystem's security context, without the need to know and store local user credentials in a PowerShell script or enable WinRM.

Are there any updates on this? Or is there an issue/PR I can track? :) Thanks

PR #6256, opened 2022-11-09
cmd/containerboot: add tests.

danderson:
Right now just one test, because it's already a lot of scaffolding code to look at. In a followup I will table-ize the test and exercise other configurations.

### DIFF
--- a/cmd/containerboot/main_test.go
+++ b/cmd/containerboot/main_test.go
@@ -1,0 +1,396 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build linux
+
+package main
+
+import (
+	"bytes"
+	_ "embed"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"net/netip"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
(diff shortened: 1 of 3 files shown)


cmd/containerboot/main_test.go
```diff
@@ -0,0 +1,438 @@
...
+func (k *kubeServer) Reset() {
+	k.Lock()
+	defer k.Unlock()
+	k.secret = map[string]string{}
+}
+
+func (k *kubeServer) Start() error {
+	root := filepath.Join(k.FSRoot, "var/run/secrets/kubernetes.io/serviceaccount")
+
+	cert, err := mkSelfSigned(filepath.Join(root, "ca.crt"))
```
maisem, line 283: `httptest.Server.StartTLS` _might_ be better.

cmd/containerboot/main_test.go
```diff
@@ -0,0 +1,438 @@
...
+			if strings.HasSuffix(line, " "+want) {
+				return
+			}
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	t.Fatalf("timed out waiting for wanted output line %q. Output:\n%s", want, b.String())
+}
+
+func waitArgs(t *testing.T, timeout time.Duration, root, path, wantArgs string) {
```
maisem, line 146: all args except root are self-explanatory.

maisem: add a comment?

maisem (approved):

cmd/containerboot/main_test.go
danderson: Replaced with httptest.Server.

cmd/containerboot/main_test.go
danderson: Documented both wait funcs, and renamed `root` to `removeStr`.

PR #6259, opened 2022-11-10
ipn/ipnlocal: add LocalBackend setPrefsLocked

bradfitz:
There were a bunch of places that assigned b.prefs, and a lot of copy/paste reactive work that wanted to run in response to b.prefs changing. Move it all to a new method.


bradfitz:
Nevermind, @maisem says this conflicts too much with his work. So ignoring for now.

PR #6260, opened 2022-11-10
syncs: add Map

dsnet:
Map is a concurrent safe map that is a trivial wrapper over a Go map and a sync.RWMutex.

It is optimized for use-cases where the entries change often, which is the opposite use-case of what sync.Map is optimized for.

The API is patterned off of sync.Map, but made generic.


dsnet:
\cc @soniaappasamy

syncs/syncs.go
```diff
@@ -152,3 +153,67 @@ func (s Semaphore) TryAcquire() bool {
...
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	value, ok = m.m[key]
+	return value, ok
+}
+
+func (m *Map[K, V]) Store(key K, value V) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if m.m == nil {
```
maisem, line 175: Not that it matters much here but you could save a few lines by using mak.Set but that has a trade off of someone having to look into the package for implementation.

syncs/syncs.go
```diff
@@ -152,3 +153,67 @@ func (s Semaphore) TryAcquire() bool {
 func (s Semaphore) Release() {
 	<-s.c
 }
+
+// Map is a Go map protected by a sync.RWMutex.
```
maisem, line 157: I think this deserves a comment saying that it should not be copied.

syncs/syncs.go
```diff
@@ -152,3 +153,67 @@ func (s Semaphore) TryAcquire() bool {
...
+func (m *Map[K, V]) Store(key K, value V) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if m.m == nil {
+		m.m = make(map[K]V)
+	}
+	m.m[key] = value
+}
+
+func (m *Map[K, V]) LoadOrStore(key K, value V) (actual V, loaded bool) {
```
maisem, line 181: It might be beneficial to try Load separately first as it uses RLock.

maisem (approved):

syncs/syncs.go
dsnet: Done. Since this includes a `sync.RWMutex`,  `go vet` will [complain if this type is shallow copied](https://golang.org/issues/8005).

syncs/syncs.go
dsnet: Done.

syncs/syncs.go
dsnet: Done.

maisem (approved):

soniaappasamy:
Nice, thanks!

PR #6261, opened 2022-11-10
Bradfitz/reverse proxy

bradfitz:
(multiple commits; maybe easiest to review in order)

### DIFF
--- a/ipn/ipnlocal/serve.go
+++ b/ipn/ipnlocal/serve.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/serve.go
```diff
@@ -37,6 +49,28 @@ func (b *LocalBackend) HandleInterceptedTCPConn(dport uint16, srcAddr netip.Addr
...
+		hs := &http.Server{
+			TLSConfig: &tls.Config{
+				GetCertificate: b.getTLSServeCertForPort(dport),
+			},
+			Handler: http.HandlerFunc(b.serveWebHandler),
+			BaseContext: func(_ net.Listener) context.Context {
+				return context.WithValue(context.Background(), serveHTTPContextKey{}, &serveHTTPContext{
+					SrcAddr:  srcAddr,
+					DestPort: dport,
+				})
```
shayne, line 67: Nice!

ipn/ipnlocal/serve.go
```diff
@@ -141,26 +159,84 @@ func (b *LocalBackend) serveWebHandler(w http.ResponseWriter, r *http.Request) {
...
+// * port number ("8080")
+// * host:port ("localhost:8080")
+// * full URL ("http://localhost:8080", in which case it's returned unchanged)
+func expandProxyArg(s string) string {
+	if s == "" {
+		return ""
+	}
+	if strings.HasPrefix(s, "http://") || strings.HasPrefix(s, "https://") {
+		return s
 	}
```
shayne, line 192: Just a thought... if we want to support a local https proxy with a (maybe self signed/invalid) cert. Weird case, but we'd do termination but the backend is https. The self signed case would require a `InsecureSkipVerify`)

bradfitz: Go at least uses a protocol scheme of `https+insecure://` for that. We could do the same?

PR #6262, opened 2022-11-10
cmd/containerboot: make a tests table, add more tests.

danderson:
Also fix a bug found while adding the tests, oops.

### DIFF
--- a/cmd/containerboot/main_test.go
+++ b/cmd/containerboot/main_test.go
@@ -23,20 +23,20 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"golang.org/x/sys/unix"
 	"tailscale.com/ipn/ipnstate"
+	"tailscale.com/tailcfg"
 )
 
 func TestContainerBoot(t *testing.T) {
-	d, err := os.MkdirTemp("", "containerboot")
-	if err != nil {
-		t.Fatal(err)
-	}
+	d := t.TempDir()
 
 	lapi := localAPI{FSRoot: d}
 	if err := lapi.Start(); err != nil {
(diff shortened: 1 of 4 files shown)


cmd/containerboot/main_test.go
```diff
@@ -23,13 +23,16 @@ import (
...
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"golang.org/x/sys/unix"
 	"tailscale.com/ipn/ipnstate"
+	"tailscale.com/tailcfg"
 )
 
 func TestContainerBoot(t *testing.T) {
```
maisem, line 38: oh maybe use t.TempDir so that we don't leak these?

cmd/containerboot/main_test.go
```diff
@@ -68,41 +90,281 @@ func TestContainerBoot(t *testing.T) {
...
+	runningSockPath := filepath.Join(d, "tmp/tailscaled.sock")
+
+	tests := []struct {
+		Name           string
+		Env            map[string]string
+		KubeSecret     map[string]string
+		WantArgs1      []string        // Wait for containerboot to run these commands...
+		Status1        ipnstate.Status // ... then report this status in LocalAPI.
+		WantArgs2      []string        // If non-nil, wait for containerboot to run these additional commands...
+		Status2        ipnstate.Status // ... then report this status in LocalAPI.
```
maisem, line 103: maybe add a TODO that if there is ever a third step it would be better to have a slice?

cmd/containerboot/main_test.go
```diff
@@ -68,41 +90,281 @@ func TestContainerBoot(t *testing.T) {
...
+			}
+			waitLogLine(t, 2*time.Second, cbOut, "Startup complete, waiting for shutdown signal")
 
-	waitLogLine(t, 2*time.Second, cbOut, "Startup complete, waiting for shutdown signal")
+			if test.WantKubeSecret != nil {
+				got := kube.Secret()
+				if diff := cmp.Diff(got, test.WantKubeSecret); diff != "" {
+					t.Fatalf("unexpected kube secret data (-got+want):\n%s", diff)
+				}
+			}
```
maisem, line 365: verify that kube.Secret doesn't exist if its not expected.

maisem (approved):

stavros-k:
Hey sorry to "bash in". While at it, can you please check this out?

Here, `proxyTo` is checked if it's an ipv4 address, else it assumes it's ipv6. Which might not be the case, as it could also be `""` (empty, aka not set). Can it check that is not empty, like it does with `routes`
Thanks

Unless this is checked somewhere before... which I missed. Then ignore me...

https://github.com/tailscale/tailscale/blob/b975bd1429cdb5f5a5325c32e6e513a2d6d498b4/cmd/containerboot/main.go#L356-L360

danderson:
Thanks for the report! You're right, I fixed the logic for routes but not for proxyIP. Fixed, and added regression tests that verify that the right forwarding settings are being configured/not configured.

PR #6268, opened 2022-11-10
ipn/localapi: rename /profile to /pprof

mihaip:
Avoids name collision with profiles for user switching.


mihaip:
I figured that since this is a semi-hidden debug command it wasn't likely to be important/frequently used enough for backward compatibility to be an issue.

PR #6269, opened 2022-11-10
ipn/localapi: add LocalAPI endpoints for profile switching

mihaip:
Exposes a REST-y API for interacting with the profile switching introduced in #6022.

### DIFF
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/exp/slices"
 	"tailscale.com/client/tailscale/apitype"
 	"tailscale.com/envknob"
 	"tailscale.com/health"
@@ -53,6 +54,7 @@ var handler = map[string]localAPIHandler{
 	"cert/":     (*Handler).serveCert,
 	"file-put/": (*Handler).serveFilePut,
 	"files/":    (*Handler).serveFiles,
+	"profiles/": (*Handler).serveProfiles,
 
 	// The other /localapi/v0/NAME handlers are exact matches and contain only NAME
 	// without a trailing slash:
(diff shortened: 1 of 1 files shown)


ipn/localapi/localapi.go
```diff
@@ -1037,6 +1038,69 @@ func (h *Handler) serveTKAModify(w http.ResponseWriter, r *http.Request) {
 	w.Write(j)
 }
 
+// serveProfiles serves profile switching-related endpoints. Supported methods
+// and paths are:
+// - GET /profiles/: list all profiles (JSON-encoded array of ipn.LoginProfiles)
+// - PUT /profiles/: add new profile (no response)
```
bradfitz, line 1044: this is a weird one. No request body and no response about what it auto-generated for you? Does it get a new random ID, or?

ipn/localapi/localapi.go
```diff
@@ -1037,6 +1038,69 @@ func (h *Handler) serveTKAModify(w http.ResponseWriter, r *http.Request) {
...
+// - GET /profiles/current: current profile (JSON-ecoded ipn.LoginProfile)
```
bradfitz, line 1045: What about `GET /profiles/<id>`? Not listed here.

ipn/localapi/localapi.go
```diff
@@ -1037,6 +1038,69 @@ func (h *Handler) serveTKAModify(w http.ResponseWriter, r *http.Request) {
...
+// - POST /profiles/<id>: switch to profile (no response)
+// - DELETE /profiles/<id>: delete profile (no response)
+func (h *Handler) serveProfiles(w http.ResponseWriter, r *http.Request) {
+	if !h.PermitWrite {
+		http.Error(w, "profiles access denied", http.StatusForbidden)
+		return
+	}
+	suffix := strings.TrimPrefix(r.URL.EscapedPath(), "/localapi/v0/profiles/")
```
bradfitz, line 1053: bit safer is:

```
suffix, ok := strs.CutPrefix(....)
if !ok {
   panic("misconfigured")
}
```

bradfitz: (CutPrefix is coming to Go 1.20; we backported it for now)

ipn/localapi/localapi.go
maisem: It is a bit weird because of the semantics where the profile is not persisted till the user logs in.
If the user switches out of the profile without logging in, they won't be able to switch back to the "empty" profile they previously created.

ipn/localapi/localapi.go
mihaip: There's no `LocalBackend` method to get a specific profile. I guess I could do it by iterating over the return value of `ListProfiles()`, but so far we have not needed it.

ipn/localapi/localapi.go
mihaip: Added a comment to say that `StartLoginInteractive()` or similar is needed to actually populate and persist the profile.

maisem (approved):
LGTM after addressing Brads comments.

ipn/localapi/localapi.go
```diff
@@ -1073,6 +1074,73 @@ func (h *Handler) serveTKAModify(w http.ResponseWriter, r *http.Request) {
...
+	if !ok {
+		panic("misconfigured")
+	}
+	if suffix == "" {
+		switch r.Method {
+		case http.MethodGet:
+			w.Header().Set("Content-Type", "application/json")
+			json.NewEncoder(w).Encode(h.b.ListProfiles())
+		case http.MethodPut:
+			h.b.NewProfile()
```
maisem, line 1100: this now returns an error.
and I think the code should be http.StatusCreated maybe?

mihaip: TIL about `StatusCreated`

ipn/localapi/localapi.go
mihaip: Added an error check and changed the response type to `StatusCreated`.

ipn/localapi/localapi.go
mihaip: Added support for `GET /profiles/<id>` too.

PR #6271, opened 2022-11-10
ipn/ipnlocal: support serving files/directories too

bradfitz:
Updates tailscale/corp#7515

### DIFF
--- a/ipn/ipnlocal/serve.go
+++ b/ipn/ipnlocal/serve.go
@@ -15,9 +15,12 @@ import (
 	"net/http/httputil"
 	"net/netip"
 	"net/url"
+	"os"
+	"path"
 	pathpkg "path"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"tailscale.com/ipn"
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/serve.go
```diff
@@ -219,6 +229,73 @@ func (b *LocalBackend) serveWebHandler(w http.ResponseWriter, r *http.Request) {
 	http.Error(w, "empty handler", 500)
 }
 
+func (b *LocalBackend) serveFileOrDirectory(w http.ResponseWriter, r *http.Request, fileOrDir, mountPoint string) {
+	fi, err := os.Stat(fileOrDir)
+	if err != nil {
+		if os.IsExist(err) {
```
andrew-d, line 235: `errors.Is(err, fs.ErrNotExist)`?

ipn/ipnlocal/serve.go
```diff
@@ -219,6 +229,73 @@ func (b *LocalBackend) serveWebHandler(w http.ResponseWriter, r *http.Request) {
 	http.Error(w, "empty handler", 500)
 }
 
+func (b *LocalBackend) serveFileOrDirectory(w http.ResponseWriter, r *http.Request, fileOrDir, mountPoint string) {
+	fi, err := os.Stat(fileOrDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			http.NotFound(w, r)
```
shayne, line 236: missing return?

shayne, line 236: missing return?

shayne (changes requested):

ipn/ipnlocal/serve.go
bradfitz: Good catch. Happens to be accidentally harmless in this case, but wrong nonetheless.

bradfitz: PTAL

shayne (approved):

PR #6272, opened 2022-11-11
net/connstats: invert network logging data flow

dsnet:
Previously, tstun.Wrapper and magicsock.Conn managed their own statistics data structure and relied on an external call to Extract to extract (and reset) the statistics.
This makes it difficult to ensure a maximum size on the statistics as the caller has no introspection into whether the number of unique connections is getting too large.

Invert the control flow such that a *connstats.Statistics is registered with tstun.Wrapper and magicsock.Conn. Methods on non-nil *connstats.Statistics are called for every packet. This allows the implementation of connstats.Statistics (in the future) to better control when it needs to flush to ensure bounds on maximum sizes.

The value registered into tstun.Wrapper and magicsock.Conn could be an interface, but that has two performance detriments:

1. Method calls on interface values are more expensive since they must go through a virtual method dispatch.

2. The implementation would need a sync.Mutex to protect the statistics value instead of using an atomic.Pointer.

Given that methods on constats.Statistics are called for every packet, we want reduce the CPU cost on this hot path.


net/connstats/stats.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// UpdateRxVirtual updates the counters for a received IP packet.
+// The source and destination of the packet are inverted with respect to
+// the source and destination in netlogtype.Connection.
+func (s *Statistics) UpdateRxVirtual(b []byte) {
+	s.updateVirtual(b, true)
+}
+
+func (s *Statistics) updateVirtual(b []byte, receive bool) {
```
dsnet, line 40: GitHub is being silly, this is identical to the removed `tunstats.Statististics.update` method.

net/connstats/stats.go
```diff
@@ -0,0 +1,109 @@
...
+	"sync"
+
+	"tailscale.com/net/packet"
+	"tailscale.com/types/netlogtype"
+)
+
+// Statistics maintains counters for every connection.
+// All methods are safe for concurrent use.
+// The zero value is ready for use.
+type Statistics struct {
```
dsnet, line 20: GitHub is being silly, this is mostly identical to the removed `tunstats.Statististics` type.

net/connstats/stats.go
```diff
@@ -0,0 +1,109 @@
...
+
+// UpdateRxPhysical updates the counters for a received wireguard packet.
+// The src is always a Tailscale IP address, representing some remote peer.
+// The dst is a remote IP address and port that corresponds
+// with some physical peer backing the Tailscale IP address.
+func (s *Statistics) UpdateRxPhysical(src netip.Addr, dst netip.AddrPort, n int) {
+	s.updatePhysical(src, dst, n, true)
+}
+
+func (s *Statistics) updatePhysical(src netip.Addr, dst netip.AddrPort, n int, receive bool) {
```
dsnet, line 80: This is moved from `magicsock.stats.m`.

dblohm7 (approved):

PR #6282, opened 2022-11-11
ipn/ipnlocal: listen to serve ports on netmap addrs

shayne:


ipn/ipnlocal/local.go
```diff
@@ -199,6 +199,7 @@ type LocalBackend struct {
 	// ServeConfig fields. (also guarded by mu)
 	lastServeConfJSON mem.RO              // last JSON that was parsed into serveConfig
 	serveConfig       ipn.ServeConfigView // or !Valid if none
+	serveListeners    map[uint16]map[string]net.Listener
```
bradfitz, line 202: always document what map keys/values are, especially for nest maps, like:

     serveListeners    map[uint16]map[string]net.Listener // foo => barID => listener

ipn/ipnlocal/local.go
```diff
@@ -3522,18 +3523,106 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked() {
...
 	}
 	b.setTCPPortsIntercepted(handlePorts)
 }
 
+// updateServeTCPPortNetMapAddrListeners starts a net.Listen for configured
+// Serve ports on all the node's addresses.
+// Existing Listeners are closed if port no longer in incoming ports list.
+//
+// b.mu must be held.
+func (b *LocalBackend) updateServeTCPPortNetMapAddrListeners(ports []uint16) {
```
bradfitz, line 3546: the method name should end in "Locked" if it's a "b.mu must be held" method

ipn/ipnlocal/local.go
```diff
@@ -3522,18 +3523,106 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked() {
...
+			}
+
+			lm := b.serveListeners[port]
+			if lm == nil {
+				lm = make(map[string]net.Listener)
+				b.serveListeners[port] = lm
+			}
+			lm[address] = l
+
+			go func() {
```
bradfitz, line 3600: we prefer long-lived goroutines to have nice names in stacks, so move this to a method rather than an anonymous func (closure? is it closing over anything? hard to tell at quick glance with the inner and outer func being big enough)

ipn/ipnlocal/local.go
```diff
@@ -3522,18 +3523,106 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked() {
...
+			}
+			address := net.JoinHostPort(ipAddr.String(), strconv.Itoa(int(p)))
+
+			if lm, ok := b.serveListeners[port]; ok {
+				if _, ok := lm[address]; ok {
+					continue // already listening
+				}
+			}
+
+			l, err := net.Listen(network, address)
```
bradfitz, line 3587: I generally prefer "ln" as the name for listeners over "l" which looks like 1 or I in many fonts

ipn/ipnlocal/local.go
```diff
@@ -3522,18 +3523,106 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked() {
...
+				defer l.Close()
+				b.logf("listening on %s", address)
+
+				for {
+					conn, err := l.Accept()
+					if err != nil {
+						b.logf("error accepting connection: %v", err)
+						return
+					}
+					srcAddr, err := netip.ParseAddrPort(conn.RemoteAddr().String())
```
bradfitz, line 3610: No need to round-trip though a string. You can use https://pkg.go.dev/net#TCPAddr.AddrPort instead.

conn.RemoteAddr().(*net.TCPAddr).AddrPort()

that'll panic if RemoteAddr isn't a TCPAddr, but it will be

ipn/ipnlocal/local.go
```diff
@@ -3522,18 +3523,106 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked() {
...
+					if err != nil {
+						b.logf("error parsing source address: %v", err)
+					}
+					getConn := func() (net.Conn, bool) { return conn, true }
+					sendRST := func() {
+						b.logf("RST of %v", srcAddr)
+					}
```
bradfitz, line 3617: conn.Close() I think.

ipn/ipnlocal/local.go
```diff
@@ -3522,18 +3523,106 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked() {
...
+					b.HandleInterceptedTCPConn(port, srcAddr, getConn, sendRST)
```
bradfitz, line 3618: This method sometimes blocks (for TCP copies at least, but not for HTTPS server)

So that means you can only Accept one thing at a time. You'll want to start this in its own goroutine.

ipn/ipnlocal/local.go
shayne: Thanks for the context!

ipn/ipnlocal/local.go
shayne: Great tip

ipn/ipnlocal/local.go
```diff
@@ -3522,18 +3523,99 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked() {
...
+	if nm.SelfNode == nil {
+		b.logf("netMap SelfNode is nil")
+		return
+	}
+
+	if b.serveListeners == nil {
+		b.serveListeners = make(map[netip.AddrPort]net.Listener)
+	}
+
+	for ap, ln := range b.serveListeners {
```
bradfitz, line 3564: this loop needs to move to the very top, before the `nm == nil` early return, no?

ipn/ipnlocal/local.go
```diff
@@ -3522,18 +3523,99 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked() {
...
+		if !slices.Contains(ports, ap.Port()) {
+			b.logf("closing listener %v", ap)
```
bradfitz, line 3566: let's get "serve" into some of these messages, like "closing serve listener %v"

ipn/ipnlocal/local.go
```diff
@@ -3522,18 +3523,99 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked() {
...
+			ln.Close()
+			delete(b.serveListeners, ap)
+		}
+	}
+
+	for _, a := range nm.Addresses {
+		for _, p := range ports {
+			addrPort := netip.AddrPortFrom(a.Addr(), p)
+			network := "tcp"
```
bradfitz, line 3575: probably can just hard-code "tcp" below. "tcp" means "TCP automatic or both as needed" whereas "tcp4" or "tcp6" means exactly those. I'd just `net.Listen("tcp", ...)` below. The actual type will match the IP address automatically.

ipn/ipnlocal/local.go
```diff
@@ -3522,18 +3523,99 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked() {
...
+
+// handleServeListenersAccept accepts connections for the serveListeners.
+// Calls HandleInterceptedTCPConn in a new goroutine for each accepted connection.
+func (b *LocalBackend) handleServeListenersAccept(ln net.Listener, port uint16) {
+	defer ln.Close()
+
+	for {
+		conn, err := ln.Accept()
+		if err != nil {
+			b.logf("error accepting connection: %v", err)
```
bradfitz, line 3606: if this was an intentional shutdown we shouldn't log a scary "listener closed EOF in use blah blah" message

Maybe each listener should be a type with a context and knowing whether it wants to be running or not. Then that listener can have a Close method that sets an atomic.Bool (or closes a context) and then closes the listener.

Then we can actually move the net.Listen into the serveListener.run goroutine and on listen failure (windows is weird) it can retry with a backoff (see the "backoff" type, which can backoff with a context, the same one you'd close on serveListener.close)

Then, if you get a real error from ln.Accept, you revert back to doing a net.Listen again (outer loop) until your context is closed.

ipn/ipnlocal/serve.go
```diff
@@ -37,6 +42,166 @@ type serveHTTPContext struct {
...
+		return
+	}
+	if nm.SelfNode == nil {
+		b.logf("netMap SelfNode is nil")
+		return
+	}
+
+	if b.serveListeners == nil {
+		b.serveListeners = make(map[netip.AddrPort]*serveListener)
+	}
```
bradfitz, line 140: don't need this anymore since it's a single-level map and you're using mak.Set on line 3651 below already

ipn/ipnlocal/serve.go
```diff
@@ -37,6 +42,166 @@ type serveHTTPContext struct {
...
+
+	for _, a := range nm.Addresses {
+		for _, p := range ports {
+			addrPort := netip.AddrPortFrom(a.Addr(), p)
+			if _, ok := b.serveListeners[addrPort]; ok {
+				continue // already listening
+			}
+
+			sl := newServeListener(context.Background(), addrPort, b.logf)
```
bradfitz, line 149: b.newServeListener(...) so it retains b, then line 152 below can just be "go sl.run()"

ipn/ipnlocal/serve.go
```diff
@@ -37,6 +42,166 @@ type serveHTTPContext struct {
...
+		cancel: cancel,
+		logf:   logf,
+
+		bo: backoff.NewBackoff("serve-listener", logf, 30*time.Second),
+	}
+}
+
+func (s *serveListener) Close() error {
+	s.cancel()
+	return s.ln.Close()
```
bradfitz, line 69: I think this s.ln access will be a data race.

make the ln field be a `syncs.AtomicValue[net.Listener]` instead?

ipn/ipnlocal/serve.go
```diff
@@ -37,6 +42,166 @@ type serveHTTPContext struct {
...
+			s.logf("serve failed to listen on %v, backing off: %v", s.ap, err)
+			s.bo.BackOff(s.ctx, err)
+			continue
+		}
+		s.ln = ln
+		break
+	}
+
+	s.logf("serve listening on %v", s.ap)
+	go s.handleServeListenersAccept(h)
```
bradfitz, line 89: why a new goroutine for the accepts? Once you have a listener, have a accept loop as long as the listener works, and then if your listener breaks (and context is still good) you can loop back around to doing another listen

ipn/ipnlocal/serve.go
```diff
@@ -37,6 +43,165 @@ type serveHTTPContext struct {
...
+
+func (s *serveListener) Run() {
+	for {
+		ln, err := net.Listen("tcp", s.ap.String())
+		if err != nil {
+			s.logf("serve failed to listen on %v, backing off: %v", s.ap, err)
+			s.bo.BackOff(s.ctx, err)
+			continue
+		}
+		s.ln.Store(ln)
```
bradfitz, line 85: s.closeListener.Store(ln.Close)

ipn/ipnlocal/serve.go
```diff
@@ -37,6 +43,165 @@ type serveHTTPContext struct {
...
+		}
+	}
+}
+
+// handleServeListenersAccept accepts connections for the Listener.
+// Calls incoming handler in a new goroutine for each accepted connection.
+func (s *serveListener) handleServeListenersAccept() error {
+	defer s.Close()
+
+	ln, ok := s.ln.LoadOk()
```
bradfitz, line 104: rather than load it, let's just pass it as an arg to the method

ipn/ipnlocal/serve.go
```diff
@@ -37,6 +43,165 @@ type serveHTTPContext struct {
 	DestPort uint16
 }
 
+type serveListener struct {
```
bradfitz, line 46:
```
// serveListener is the state of host-level net.Listen for a specific (Tailscale IP, serve port)
// combination. If there are two TailscaleIPs (v4 and v6) and three ports being served,
// then there will be six of these active and looping in their Run method.
//
// This is not used in userspace-networking mode.
//
// Most serve traffic is intercepted by netstack. This exists purely for connections
// from the machine itself, as that goes via the kernel, so we need to be in the
// kernel's listening/routing tables.
```

bradfitz (approved):

PR #6287, opened 2022-11-11
cmd/tailscale: Add tailnet name to cli and web output of status command

### LINKED ISSUE #6286: FR: add tailnet name to `tailscale status`

It could be nice to have a fast way of getting the tailnet name from cli.

### How should we solve this?

By letting the `tailscale status` command output tailnet name as well.

Learn the tailnet name by heart.

### Anything else?

_No response_

petteja:
Fixes #6286

### DIFF
--- a/cmd/tailscale/cli/status.go
+++ b/cmd/tailscale/cli/status.go
@@ -146,6 +146,9 @@ func runStatus(ctx context.Context, args []string) error {
 
 	var buf bytes.Buffer
 	f := func(format string, a ...any) { fmt.Fprintf(&buf, format, a...) }
+
+	f("%-15s %-20s\n", "tailnet name", st.CurrentTailnet.MagicDNSSuffix)
+
 	printPS := func(ps *ipnstate.PeerStatus) {
 		f("%-15s %-20s %-12s %-7s ",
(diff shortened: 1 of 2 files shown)


cmd/tailscale/cli/status.go
```diff
@@ -146,6 +146,9 @@ func runStatus(ctx context.Context, args []string) error {
 
 	var buf bytes.Buffer
 	f := func(format string, a ...any) { fmt.Fprintf(&buf, format, a...) }
+
+	f("%-15s %-20s\n", "tailnet name", st.CurrentTailnet.MagicDNSSuffix)
```
bradfitz, line 150: You should guard against CurrentTailnet being nil.

I wonder if the output format should be more like a header at the top, like Health check above is, rather than trying to blend in with the peers.

petteja: Thanks! Maybe the peers should have a heading as well, but I don't know if the heading should be "Peers", or "Machines" as in admin console, or  "Nodes" as in `tailscale status -web`.

DentonGentry:
I imagine this should be a header, ys.

petteja:
Yes, added a header similar to the health check header.

PR #6290, opened 2022-11-11
net/dns/resolver: add yet another 4via6 DNS form that's hopefully more robust

bradfitz:
```
$ dig +short @100.100.100.100 aaaa 10-2-5-3-via-7.foo-bar.ts.net fd7a:115c:a1e0:b1a:0:7:a02:503

$ dig +short @100.100.100.100 aaaa 10-2-5-3-via-7
fd7a:115c:a1e0:b1a:0:7:a02:503

$ ping 10-2-5-3-via-7
PING 10-2-5-3-via-7(fd7a:115c:a1e0:b1a:0:7:a02:503 (fd7a:115c:a1e0:b1a:0:7:a02:503)) 56 data bytes ...
```

### DIFF
--- a/net/dns/resolver/tsdns.go
+++ b/net/dns/resolver/tsdns.go
@@ -550,7 +550,7 @@ func (r *Resolver) resolveLocal(domain dnsname.FQDN, typ dns.Type) (netip.Addr, 
 			return tsaddr.TailscaleServiceIPv6(), dns.RCodeSuccess
 		}
 	}
-	// Special-case: 'via-<siteid>.<ipv4>' queries.
+	// Special-case: 4via6 DNS names.
 	if ip, ok := r.parseViaDomain(domain, typ); ok {
 		return ip, dns.RCodeSuccess
 	}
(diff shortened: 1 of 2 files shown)


net/dns/resolver/tsdns.go
```diff
@@ -650,14 +652,28 @@ func (r *Resolver) parseViaDomain(domain dnsname.FQDN, typ dns.Type) (netip.Addr
 
 	var siteID string
 	var ip4Str string
-	if strings.HasPrefix(fqdn, "via-") {
+	switch {
+	case strings.Contains(fqdn, "-via-"):
+		// Format number 3: "192-168-1-2-via-7" or "192-168-1-2-via-7.foo.ts.net."
```
DentonGentry, line 657: If we do eventually implement the named site-ids proposal I think this would be `192-168-1-2-via-toronto.foo.ts.net`, which seems fine.

net/dns/resolver/tsdns.go
```diff
@@ -650,14 +652,28 @@ func (r *Resolver) parseViaDomain(domain dnsname.FQDN, typ dns.Type) (netip.Addr
...
+		// Third time's a charm. The earlier two formats follow after this block.
+		firstLabel, domain, _ := strings.Cut(fqdn, ".") // "192-168-1-2-via-7"
+		if !(domain == "" || dnsname.HasSuffix(domain, "ts.net") || dnsname.HasSuffix(domain, "tailscale.net")) {
+			return netip.Addr{}, false
```
DentonGentry, line 661: One unfortunate side-effect (which also happened with the previous name format) is that an inadvertent hostname matching like `netflix-via-europe.example.com` can't be resolved, we don't fallback to the default case.

DentonGentry (approved):

net/dns/resolver/tsdns.go
bradfitz: I don't think that's true. If this `parseViaDomain` returns false, we fall through and keep matching the old way.

```go
	// Special-case: 'via-<siteid>.<ipv4>' queries.
	if ip, ok := r.parseViaDomain(domain, typ); ok {
		return ip, dns.RCodeSuccess
	}
....
....
```

DentonGentry: Aha! Ok.

bradfitz: Now that I see that comment, though, I'll update it, as that style is two styles old now.

Issue #6266, opened 2022-11-10
FR: Make the need for the --accept-routes flag explict in the Linux download page, under certain conditions


umm0n:
### What are you trying to do?

The [Linux download page](https://tailscale.com/download/linux) for Tailscale should have a small disclaimer about the need for the `--accept-routes` flag when you need to accept subnet routes from other nodes in the network, as explained in the [troubleshooting guide](https://tailscale.com/kb/1023/troubleshooting/#linux).

### How should we solve this?

Add the disclaimer to the Linux download page.

### What is the impact of not solving this?

When trying to make Tailscale as self-service as possible (but aimed at a technical audience) it would be useful to have this. Not having this disclaimer outright can lead to some confusion for users, when their connections fail while using `tailscale up` and nothing more. It would be simpler to include it in the download page, rather than have users hunt for it in the troubleshooting guide. I imagine it's a pretty common requirement (to accept subnet routes from other nodes).

DentonGentry:
We've added a warning to `tailscale up` and a health check to `tailscale status`, which I think is more likely to be at the moment one is likely to act on it. 

```
root@ubuntu:~# tailscale up
Some peers are advertising routes but --accept-routes is false
```

Closing as sufficiently addressed.

umm0n:
Excellent, thanks!

PR #6291, opened 2022-11-11
cmd/tailscale: make "up", "status" warn if routes and --accept-routes off

bradfitz:
Example output:

    # Health check:
    #     - Some peers are advertising routes but --accept-routes is false

Also, move "tailscale status" health checks to the bottom, where they won't be lost in large netmaps.

Updates #2053
Updates #6266


DentonGentry (approved):

bradfitz:
Making the health checks run ~immediately (milliseconds after!) after the up transition reminded me why we had that ugly "SSH"-substring-checking code in there for the SSH warnings. There's a lot of health spam on transition for a second.

So I also went and cleaned that up, adding a package of well-known error messages/substring constants for now that we want to warn up on up, and only made it show those, moving the SSH special case there, with some new tests.

Later we can clean further, making it the health warnings typed with metadata about how bad it is. Added a TODO for that.

PR #6295, opened 2022-11-12
ipn/ipnlocal: call restart backend on user changes

maisem:
This makes it so that the backend also restarts when users change, otherwise an extra call to Start was required.

Updates #713

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -2097,8 +2097,15 @@ func (b *LocalBackend) shouldUploadServices() bool {
 
 func (b *LocalBackend) SetCurrentUserID(uid string) {
 	b.mu.Lock()
-	b.pm.SetCurrentUser(uid)
-	b.mu.Unlock()
+	if b.pm.CurrentUser() == uid {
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/local.go
```diff
@@ -2097,8 +2097,15 @@ func (b *LocalBackend) shouldUploadServices() bool {
...
-	b.mu.Unlock()
+	if b.pm.CurrentUser() == uid {
+		b.mu.Unlock()
+		return
+	}
+	if err := b.pm.SetCurrentUser(uid); err != nil {
+		b.mu.Unlock()
+		return
+	}
+	b.resetForProfileChangeLockedOnEntry()
```
bradfitz, line 2108: oh no, more "LockedOnEntry"? I wanted to eliminate the one we had :(

I'll approve, but that sucks

bradfitz (approved):

ipn/ipnlocal/local.go
maisem: I agree, I plan on refactoring this package next.

PR #6296, opened 2022-11-12
cmd/tailscale/cli: [set] handle selectively modifying routes/exit node

maisem:
Noticed this while debugging something else, we would reset all routes if either `--advertise-exit-node` or `--advertise-routes` were set. This handles correctly updating them.

Also added tests.


DentonGentry:
Fixing this means that starting the exit node with `tailscale up --advertise-exit-node --advertise-routes=172.16.0.0/16` will result in other being able to use the exit node and also reach 172.16.0.0 through a local interface on the exit node?

maisem:
> Fixing this means that starting the exit node with `tailscale up --advertise-exit-node --advertise-routes=172.16.0.0/16` will result in other being able to use the exit node and also reach 172.16.0.0 through a local interface on the exit node?

This is unrelated to `tailscale up`, its for the new `tailscale set` subcommand added in 19b558657308a0cd6d8e2eac272737552fb04725.

cmd/tailscale/cli/set.go
```diff
@@ -128,3 +136,27 @@ func runSet(ctx context.Context, args []string) (retErr error) {
 	_, err = localClient.EditPrefs(ctx, maskedPrefs)
 	return err
 }
+
+func calcAdvertiseRoutesForSet(advertiseExitNodeSet, advertiseRoutesSet bool, curPrefs *ipn.Prefs, setArgs setArgsT) (routes []netip.Prefix, err error) {
```
bradfitz, line 140: this signature could use some docs, even if it's unexported

cmd/tailscale/cli/set_test.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"net/netip"
+	"reflect"
+	"testing"
+
+	"tailscale.com/ipn"
+	"tailscale.com/net/tsaddr"
+)
+
+func ptrTo[T any](v T) *T { return &v }
```
bradfitz, line 16: hah, we have this everywhere :)

bradfitz: (we'll clean 'em all later)

cmd/tailscale/cli/set_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+func TestCalcAdvertiseRoutesForSet(t *testing.T) {
+	pfx := netip.MustParsePrefix
+	tests := []struct {
+		name     string
+		exit     *bool
+		routes   *string
+		existing []netip.Prefix
```
bradfitz, line 24: "had" or "was" is a bit shorter? don't care, though

cmd/tailscale/cli/set_test.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+func ptrTo[T any](v T) *T { return &v }
+
+func TestCalcAdvertiseRoutesForSet(t *testing.T) {
+	pfx := netip.MustParsePrefix
+	tests := []struct {
+		name     string
+		exit     *bool
+		routes   *string
```
bradfitz, line 23: setExit, setRoutes ?

bradfitz (approved):
Nice tests!

cmd/tailscale/cli/set.go
maisem: Done

cmd/tailscale/cli/set_test.go
maisem: Done

cmd/tailscale/cli/set_test.go
maisem: Done

cmd/tailscale/cli/set_test.go
maisem: its a very useful func!

PR #6297, opened 2022-11-12
ipn/ipnlocal: delete profile on logout

maisem:
Updates #713

Signed-off-by: Maisem Ali <maisem@tailscale.com>

PR #6299, opened 2022-11-12
ipn/ipnlocal: make TKA tests not have side effects

maisem:
It left the envknob turned on which meant that running all the tests in the package had different behavior than running just any one test.


andrew-d (approved):

twitchyliquid64 (approved):
LGTM though I’ll note that I’ll be removing the requirement for the envknob in ~1-2 weeks!

maisem:
This was caught in a test failure on https://github.com/tailscale/tailscale/pull/6297. I fixed the failure in that PR, so it did catch an issue which is great!

However, individual tests shouldn't leak config knobs and result in failure far away from them. This PR is just meant to make sure tests clean up after themselves.

PR #6304, opened 2022-11-13
cmd/tailscale/cli: flesh out serve CLI and tests

shayne:


cmd/tailscale/cli/serve.go
```diff
@@ -101,7 +134,66 @@ func (e *serveEnv) stdout() io.Writer {
...
+
+	if sc != nil && sc.Web != nil && sc.Web[key] != nil &&
+		sc.Web[key].Handlers[path] != nil {
+		if (h.Proxy != "" && sc.Web[key].Handlers[path].Proxy == h.Proxy) ||
+			(h.Text != "" && sc.Web[key].Handlers[path].Text == h.Text) ||
+			(h.Path != "" && sc.Web[key].Handlers[path].Path == h.Path) {
+			// Nothing to do.
+			return nil
+		}
+	}
```
bradfitz, line 186: might be easier to blindly make the change to another copy of sc and then DeepEqual them to see if you need to call setServeConfig

cmd/tailscale/cli/serve_test.go
```diff
@@ -54,6 +55,18 @@ func TestServeConfigMutations(t *testing.T) {
...
 	})
 
+	// https
+	add(step{reset: true})
+	add(step{
+		command: cmd("/ proxy 3000"),
+		reset:   false,
+		want: &ipn.ServeConfig{Web: map[ipn.HostPort]*ipn.WebServerConfig{
+			"foo:443": {Handlers: map[string]*ipn.HTTPHandler{
+				"/": {Proxy: "localhost:3000"},
```
bradfitz, line 65: 127.0.0.1:3000 instead? I don't think "localhost" is hard-coded yet in Go and I don't want to worry (or even think) about hostname/DNS lookups in that path.

bradfitz (approved):

cmd/tailscale/cli/serve.go
```diff
@@ -101,7 +134,66 @@ func (e *serveEnv) stdout() io.Writer {
...
+	case "text":
+		h.Text = args[2]
+	default:
+		// TODO(shayne): fix, tell why
+		return flag.ErrHelp
+	}
+
+	sc, err := e.getServeConfig(ctx)
+	if err != nil {
+		// TODO(shayne): logging?
```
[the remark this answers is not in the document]
shayne: fair enough, wasn't sure if we wanted to ship any of these errors up

cmd/tailscale/cli/serve.go
```diff
@@ -63,9 +68,37 @@ type serveEnv struct {
...
+		return e.testSelfDNSName, nil
+	}
+	st, err := getLocalClientStatus(ctx)
+	if err != nil {
+		return "", fmt.Errorf("getting client status: %w", err)
+	}
+	if st.Self == nil {
+		return "", fmt.Errorf("no self node")
+	}
+	return st.Self.DNSName, nil
```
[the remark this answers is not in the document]
shayne: Yeah might as well here do it up here. Other than absoluteness is there any reason we have the FQDN dot at the root?

cmd/tailscale/cli/serve_test.go
shayne: image

cmd/tailscale/cli/serve.go
```diff
@@ -115,27 +191,435 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
...
+	h := new(ipn.HTTPHandler)
+
+	switch args[1] {
+	case "path":
+		if version.IsSandboxedMacOS() {
+			dir := "io.tailscale.ipn.macos"
+			if version.IsMacSysExt() {
+				dir = "io.tailscale.ipn.macsys"
+			}
+			printf("Warning: the macOS CLI runs in a sandbox; this binary's filesystem writes go to $HOME/Library/Containers/%s/Data\n", dir)
```
bradfitz, line 219: not "writes"

bradfitz: Let's just disallow "path" entirely on sandboxed macOS

cmd/tailscale/cli/serve.go
```diff
@@ -7,46 +7,95 @@ package cli
...
 func newServeCommand(e *serveEnv) *ffcli.Command {
 	return &ffcli.Command{
-		Name:       "serve",
-		ShortHelp:  "TODO",
-		ShortUsage: "serve {show-config|https|tcp|ingress} <args>",
-		LongHelp:   "", // TODO
-		Exec:       e.runServe,
-		FlagSet:    e.newFlags("serve", func(fs *flag.FlagSet) {}),
+		Name:      "serve",
+		ShortHelp: "!! ALPHA !! Serve from your Tailscale node",
```
bradfitz, line 39: [ALPHA] ...

cmd/tailscale/cli/serve.go
```diff
@@ -7,46 +7,95 @@ package cli
...
+		ShortUsage: strings.TrimSpace(`
+  serve [flags] <mount-point> {proxy|path|text} <arg>
+  serve [flags] <sub-command> [sub-flags] <args>`),
+		LongHelp: strings.TrimSpace(`
+!! ALPHA !!
```
bradfitz, line 44: this is more acceptable, but you also have more space here so you can elaborate.

`*** ALPHA; all of this is subject to change ***`

tailcfg/tailcfg.go
```diff
@@ -1682,6 +1682,11 @@ const (
 	CapabilityIngress = "https://tailscale.com/cap/ingress"
 )
 
+const (
+	// NodeAttrFunnel grants the ability for a node to host ingress traffic.
+	NodeAttrFunnel = "funnel"
```
maisem, line 1687: Looks unused, remove? Maybe move comment to `CapabilityIngress` instead?

maisem: oh, github fail. Can we reuse `CapabilityIngress` for the wire?

cmd/tailscale/cli/serve.go
```diff
@@ -7,46 +7,105 @@ package cli
...
+
+You can also choose to enable the Tailscale Funnel with:
+'tailscale serve funnel on'.
+
+This allows your services to be accessed over the internet,
+securely using TLS and of course, Tailscale.
+
+EXAMPLES
+  - To proxy requests to a local web server on port 3000:
+    $ tailscale serve / proxy 3000
```
maisem, line 58: unclear if this can be local address or can only be localhost.

tailcfg/tailcfg.go
shayne: in cli/serve.go
```go
if !slices.Contains(st.Self.Capabilities, tailcfg.NodeAttrFunnel) {
	return errors.New("Funnel not available. See https://tailscale.com/s/no-funnel")
}
```

maisem: GitHub decided to hide most of the PR cause the diff was too big 🤦🏼‍♂️

cmd/tailscale/cli/serve.go
shayne: Changing example to:
```
  - To proxy requests to a web server at 127.0.0.1:3000:
    $ tailscale serve / proxy 3000
```

Also error output is:
```console
$  sudo -E go run ./cmd/tailscale serve / proxy foobar:3000
only localhost or 127.0.0.1 proxies are currently supported
exit status 1
```

shayne: PTAL

cmd/tailscale/cli/cli.go
```diff
@@ -190,18 +190,19 @@ change in the future.
 	if envknob.UseWIPCode() {
 		rootCmd.Subcommands = append(rootCmd.Subcommands,
 			idTokenCmd,
-			serveCmd,
 		)
 	}
 
-	// Don't advertise the debug command, but it exists.
+	// Don't advertises these commands, but they're still explicitly available.
```
bradfitz, line 196: typo: advertise

cmd/tailscale/cli/serve.go
```diff
@@ -7,46 +7,105 @@ package cli
...
+*** ALPHA; all of this is subject to change ***
+
+The 'tailscale serve' set of commands allows you to serve
+content and local servers from your Tailscale node to
+your tailnet. 
+
+You can also choose to enable the Tailscale Funnel with:
+'tailscale serve funnel on'.
+
+This allows your services to be accessed over the internet,
```
bradfitz, line 53: It's not clear that this, down in its own paragraph, refers to Funnel and not Serve in general.

Let's move it up to line 51 and say "Funnel" instead of "This".

cmd/tailscale/cli/serve.go
```diff
@@ -58,13 +117,44 @@ func newServeCommand(e *serveEnv) *ffcli.Command {
...
 
 	// optional stuff for tests:
-	testFlagOut        io.Writer
-	testGetServeConfig func(context.Context) (*ipn.ServeConfig, error)
-	testSetServeConfig func(context.Context, *ipn.ServeConfig) error
-	testStdout         io.Writer
+	testFlagOut              io.Writer
+	testGetServeConfig       func(context.Context) (*ipn.ServeConfig, error)
+	testSetServeConfig       func(context.Context, *ipn.ServeConfig) error
+	testGetLocalClientStatus func(context.Context) (*ipnstate.Status, error)
```
maisem, line 129: I think it might be better to have a fake localClient instead of these. 
Maybe adding a `type localServeClient interface{}` with the required functions to facilitate.

Can happen in a follow up.

cmd/tailscale/cli/serve.go
```diff
@@ -7,46 +7,105 @@ package cli
...
+
+The 'tailscale serve' set of commands allows you to serve
+content and local servers from your Tailscale node to
+your tailnet. 
+
+You can also choose to enable the Tailscale Funnel with:
+'tailscale serve funnel on'.
+
+This allows your services to be accessed over the internet,
+securely using TLS and of course, Tailscale.
```
bradfitz, line 54: this isn't marketing copy. :) We can stick to the facts.

Funnel allows you to publish a 'tailscale serve' server publicly, open to
the entire internet. See https://tailscale.com/funnel.

(that shortlink is coming)

cmd/tailscale/cli/serve.go
```diff
@@ -7,46 +7,105 @@ package cli
...
+				UsageFunc: usageFunc,
 			},
 			{
-				Name:      "ingress",
-				Exec:      e.runServeIngress,
-				ShortHelp: "enable or disable ingress",
-				FlagSet:   e.newFlags("serve-ingress", func(fs *flag.FlagSet) {}),
+				Name:      "funnel",
+				Exec:      e.runServeFunnel,
+				ShortHelp: "turn on or off the Tailscale Funnel",
```
bradfitz, line 104: turn Tailscale Funnel on or off

cmd/tailscale/cli/serve.go
```diff
@@ -115,31 +229,457 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
...
+
+	for k, v := range sc.Web[hp].Handlers {
+		if v == h {
+			continue
+		}
+		// If the new mount point ends in / and another mount point
+		// shares the same prefix, remove the other handler.
+		// (e.g. /foo/ overwrites /foo)
+		// The opposite example is also handled.
+		if (strings.HasSuffix(mount, "/") && k == mount[:len(mount)-1]) ||
```
maisem, line 317: maybe:
```
m1 := strings.TrimSuffix(mount, "/")
m2 := strings.TrimSuffix(k, "/")
if m1 == m2 {
  delete(sc.Web[hp].Handlers, k)
}
```

cmd/tailscale/cli/serve.go
```diff
@@ -7,46 +7,105 @@ package cli
...
 			{
-				Name:      "ingress",
-				Exec:      e.runServeIngress,
-				ShortHelp: "enable or disable ingress",
-				FlagSet:   e.newFlags("serve-ingress", func(fs *flag.FlagSet) {}),
+				Name:      "funnel",
+				Exec:      e.runServeFunnel,
+				ShortHelp: "turn on or off the Tailscale Funnel",
+				FlagSet: e.newFlags("serve-funnel", func(fs *flag.FlagSet) {
+					fs.UintVar(&e.servePort, "serve-port", 443, "port to serve on (443, 8443 or 10000)")
```
bradfitz, line 106: just "port" for flag? and this is the port to turn on or off. 

Note (in LongHelp?) that it doesn't turn off serving on the tailnet; only the internet.

cmd/tailscale/cli/serve.go
```diff
@@ -101,7 +191,31 @@ func (e *serveEnv) stdout() io.Writer {
 	return os.Stdout
 }
 
+func (e *serveEnv) validServePortWithString() (uint16, string, error) {
```
bradfitz, line 194: this signature isn't obvious. Name the result parameters at least. Maybe also document what it's doing.

bradfitz: wait, all it does is also return strconv.Itoa also? the caller can do that.

then this can just be (port uint16, err error)

cmd/tailscale/cli/serve.go
```diff
@@ -115,31 +229,457 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
...
+		}
+	}
+	return false
 }
 
-func (e *serveEnv) runServeShowConfig(ctx context.Context, args []string) error {
+// isTCPForwardingOnPort checks serve config to see if
+// we're specifically forwarding TCP on the given port.
+func isTCPForwardingOnPort(sc *ipn.ServeConfig, port uint16) bool {
+	if sc == nil || sc.TCP == nil || sc.TCP[port] == nil {
```
bradfitz, line 455: it's safe to do reads on a nil map, so the sc.TCP == nil part is not needed

cmd/tailscale/cli/serve.go
```diff
@@ -115,31 +229,457 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
...
+	for p, h := range sc.TCP {
+		if h.TCPForward == "" {
+			continue
+		}
+		hp := ipn.HostPort(net.JoinHostPort(dnsName, strconv.Itoa(int(p))))
+		tlsStatus := "TLS over TCP"
+		if h.TerminateTLS != "" {
+			tlsStatus = "TLS terminated"
+		}
+		printf("┌── tcp://%s (%s, funnel on: %t)\n", hp, tlsStatus, isFunnelOn(sc, hp))
```
bradfitz, line 539: instead of "funnel on: false" and true, let's just say "(tailnet only)" and "(Funnel on)" or something?

cmd/tailscale/cli/serve.go
```diff
@@ -115,31 +229,457 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
...
+		for _, a := range st.TailscaleIPs {
+			ipp := net.JoinHostPort(a.String(), strconv.Itoa(int(p)))
+			printf("├── tcp://%s\n", ipp)
+		}
+		printf("└───► tcp://%s\n", h.TCPForward)
```
bradfitz, line 544: sure all this unicode stuff is going to work okay including in windows cmd.exe?

I think this is the first time we've done non-ASCII in the CLI

cmd/tailscale/cli/serve.go
```diff
@@ -115,31 +229,457 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
...
+	}
+	return nil
+}
+
+func httpHandlerExists(sc *ipn.ServeConfig, hp ipn.HostPort, mount string) bool {
+	_, ok := getHTTPHandler(sc, hp, mount)
+	return ok
+}
+
+func getHTTPHandler(sc *ipn.ServeConfig, hp ipn.HostPort, mount string) (handler *ipn.HTTPHandler, ok bool) {
```
maisem, line 380: you could have this just return nil when it doesn't exist

```
func httpHandlerExists(sc *ipn.ServeConfig, hp ipn.HostPort, mount string) bool {
	return getHTTPHandler(sc, hp, mount) != nil
}

// getHTTPHandler returns the HTTPHandler for the hp and mount. It returns
// nil if not present.
func getHTTPHandler(sc *ipn.ServeConfig, hp ipn.HostPort, mount string) *ipn.HTTPHandler {
	if sc.Web[hp] != nil {
		return sc.Web[hp].Handlers[mount]
	}
	return nil
}
```

looking up from a `nil` map is fine.

cmd/tailscale/cli/serve.go
```diff
@@ -115,31 +229,457 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
...
+	}
+	url := u.Scheme + "://" + host
+	if u.Port() != "" {
+		url += ":" + u.Port()
+	}
+	return url, nil
+}
+
+// isTCPForwardingAny checks if any TCP port is being forwarded.
+func isTCPForwardingAny(sc *ipn.ServeConfig) bool {
```
maisem, line 440: a lot of these should probably be helpers on the `ServeConfig` struct.
Fine to do later.

maisem (approved):

bradfitz (approved):

cmd/tailscale/cli/serve.go
shayne: #6372 created

cmd/tailscale/cli/serve.go
shayne: Done

cmd/tailscale/cli/serve.go
shayne: Done.

cmd/tailscale/cli/serve.go
shayne: Done.

cmd/tailscale/cli/serve.go
shayne: reasoning: was a convenience because we use ports as uint16 and as strs all over. Sometimes we need one or the other and often both. Happy to change.

```go
// validateServePort returns --serve-port flag value,
// or an error if the port is not a valid port to serve on.
func (e *serveEnv) validateServePort() (port uint16, err error) {
```

cmd/tailscale/cli/serve.go
shayne: Done.

cmd/tailscale/cli/serve.go
shayne: I like that!

cmd/tailscale/cli/serve.go
shayne: moving to ascii for now will follow up

cmd/tailscale/cli/serve.go
shayne: I agree. They're of general use but limited to the CLI right now.

cmd/tailscale/cli/serve.go
shayne: oof yeah should've been rolled up... done but also also checking `sc != nil`

cmd/tailscale/cli/serve.go
shayne: I actually think it was a mistake I put `serve-port` flag here and in TCP as well.

`--serve-port` scopes the entire command regardless of sub-command. It also maps existing servers 1:1 with the funnel port.

```console
# serve defaults to --serve-port=443
# serve listens locally on tailscale-ip;443 
# funnel on will funnel into tailscale-ip:443 be default
$ tailscale serve / proxy 3000
$ tailscale serve funnel on

# specifying --serve-port=10000 scopes the command
# to listen locally on tailscale-ip:10000
# funnel with explicit flag will funnel into tailscale-ip:10000
$ tailscale serve --serve-port=10000 tcp 5432
$ tailscale serve --serve-port=10000 funnel on
```

shayne: I've removed `--serve-port` from sub-commands, leaving only the top level. LMK

cmd/tailscale/cli/serve.go
shayne: Done.

Issue #4254, opened 2022-03-22
DNS breaks on NixOS system after resume


meithecatte:
### What is the issue?

If tailscale is enabled before I suspend my laptop, resolving domains hangs until I do `sudo tailscale down`. If I then do `sudo tailscale up`, everything works again.

`/etc/resolv.conf` looks as follows:
```
# Generated by resolvconf
search niedzejkob.github.beta.tailscale.net
nameserver 100.100.100.100
options edns0
```

### Steps to reproduce

1. Configure NixOS with settings similar to:
```
  networking.wireless.enable = true;
  # The global useDHCP flag is deprecated, therefore explicitly set to false here.
  # Per-interface useDHCP will be mandatory in the future, so this generated config
  # replicates the default behaviour.
  networking.useDHCP = false;
  networking.interfaces.wlp3s0.useDHCP = true;
  services.tailscale.enable = true;
```
(full config available [here](https://github.com/NieDzejkob/dotfiles.pub/blob/master/nixos/configuration.nix), probably mostly irrelevant)
2. Set up Tailscale with `sudo tailscale up`
3. Verify that `host github.com` resolves properly
4. Suspend the machine, and then wake it up. I've found the issue to reproduce regardless of how long the laptop is left suspended.
5. Attempt running `host github.com` and see that it hangs.
6. Observe that, nevertheless, domains under `niedzejkob.github.beta.tailscale.net` resolve properly.
7. Run `sudo tailscale down` and `sudo tailscale up`.
8. `host github.com` works again.

### Are there any recent changes that introduced the issue?

I installed Tailscale ;)

### OS

Linux

### OS version

NixOS unstable

### Tailscale version

1.22.2

### Bug report

BUG-691539f77b145a26bac1296772c1d69e36cb2e33f8ea3596aceb3ddeb2ac1e6c-20220322151817Z-124dbfbbe0e4b57b

bradfitz:
I think you'll want to use `services.resolved.enable = true`. In general that's the most complete DNS implementation for Linux and well supported with Tailscale.

/cc @danderson @Xe for NixOS stuff

NieDzejkob:
Thanks for the hint. I've enabled resolved, and now the DNS breaks the moment I run `sudo tailscale up`.

Michcioperz:
Didn't the original workaround (which I happen to still have in my config) also involve `services.resolved.dnssec = "false";`?

DentonGentry:
@NieDzejkob can you:
[...]

PR #6306, opened 2022-11-13
net/dns: make "direct" mode on Linux warn on resolv.conf fights

bradfitz:
Run an inotify goroutine and watch if another program takes over /etc/inotify.conf. Log if so.

For now this only logs. In the future I want to wire it up into the health system to warn (visible in "tailscale status", etc) about the situation, with a short URL to more info about how you should really be using systemd-resolved if you want programs to not fight over your DNS files on Linux.

Updates #4254 etc etc


bradfitz:
(I got sick of the same bug reports)

cmd/tailscaled/depaware.txt
```diff
@@ -71,6 +71,7 @@ tailscale.com/cmd/tailscaled dependencies: (generated by github.com/tailscale/de
         github.com/golang/groupcache/lru                             from tailscale.com/net/dnscache
         github.com/google/btree                                      from gvisor.dev/gvisor/pkg/tcpip/header+
         github.com/hdevalence/ed25519consensus                       from tailscale.com/tka
+   L 💣 github.com/illarion/gonotify                                 from tailscale.com/net/dns
```
danderson, line 74: I assume you looked at the use of unsafe and weren't horrified?

net/dns/direct_linux.go
```diff
@@ -0,0 +1,61 @@
...
+			}
+		}
+		if !match {
+			continue
+		}
+		m.checkForFileTrample()
+	}
+}
+
+func (m *directManager) closeInotifyOnClosed(ctx context.Context, in *gonotify.Inotify) {
```
danderson, line 58: OnDone rather than closed? Contexts don't close, even if this particular context is a close-ish context.

net/dns/direct_linux.go
```diff
@@ -0,0 +1,61 @@
...
+			return
+		}
+		if err != nil {
+			m.logf("dns: inotify read: %v", err)
+			return
+		}
+		var match bool
+		for _, ev := range events {
+			if ev.Name == resolvConf {
+				match = true
```
danderson, line 48: Can break out of the loop here

danderson (approved):

cmd/tailscaled/depaware.txt
bradfitz: Yeah, it's exactly what I expected. Nice minimal wrapper package. One little unsafe.Pointer use to copy things out of the buffer, like we have in util/dirwalk too.

net/dns/direct_linux.go
bradfitz: done 🥁

PR #6308, opened 2022-11-13
health, DNS stuff

bradfitz:
2 commits

### DIFF
--- a/health/health.go
+++ b/health/health.go
@@ -25,9 +25,10 @@ var (
 	// mu guards everything in this var block.
 	mu sync.Mutex
 
-	sysErr   = map[Subsystem]error{}                     // error key => err (or nil for no error)
-	watchers = map[*watchHandle]func(Subsystem, error){} // opt func to run if error state changes
-	timer    *time.Timer
+	sysErr    = map[Subsystem]error{}                     // error key => err (or nil for no error)
+	watchers  = map[*watchHandle]func(Subsystem, error){} // opt func to run if error state changes
+	warnables = map[*Warnable]struct{}{}                  // set of warnables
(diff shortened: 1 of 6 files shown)


health/health.go
```diff
@@ -67,16 +68,82 @@ const (
...
+	return warnOptFunc(func(w *Warnable) {
+		w.debugFlag = name
+	})
+}
+
+type warnOptFunc func(*Warnable)
+
+func (f warnOptFunc) mod(w *Warnable) { f(w) }
+
+// Warnable is a health check item that may or
```
maisem, line 104: Incomplete sentence

DentonGentry, line 104: Comment appears to be truncated.

maisem (approved):

net/dns/direct.go
```diff
@@ -336,7 +337,7 @@ func (m *directManager) checkForFileTrample() {
 		show = show[:1024]
 	}
 	m.logf("trample: resolv.conf changed from what we expected. did some other program interfere? current contents: %q", show)
-	// TODO(bradfitz): register with health package that something is wrong
+	warnTrample.Set(errors.New("Linux DNS config not ideal. /etc/resolv.conf overwritten. See https://tailscale.com/s/dns-fight"))
```
DentonGentry, line 340: We'll use https://tailscale.com/s/* to redirect to KB articles?

DentonGentry (approved):

net/dns/direct.go
bradfitz: Yeah, we use /s/ already in a handful of places in the client, as we can't update old clients in the field but we can update our redirect.txt file :)

PR #6309, opened 2022-11-13
ipn/ipnlocal: don't warn about serve listener failing on IPv6-less machines

### LINKED ISSUE #6303: ipn: serve IPv6 Listener fails on WSL 2
`serveListener` will try and listen on all addresses in the network map. WSL 2 is famously missing IPv6 support (https://github.com/microsoft/WSL/discussions/5855).

As of now it will continue to log errors (L96 below) at a frequency determined by the backoff.

One solution would be to check the `net.Interface` for Tailscale and see what addresses on the system match those in the network map.

My WSL 2 shows only RFC 4291:
```console
❯ ip -6 a s tailscale0
8: tailscale0: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1280 qdisc fq_codel state UNKNOWN group default qlen 500
    inet6 fe80::6b62:77c8:e5de:3239/64 scope link stable-privacy
```

https://github.com/tailscale/tailscale/blob/b5ac9172fd7749fe3dc9be79946ffdec4b6b1a8e/ipn/ipnlocal/serve.go#L89-L99

Filing before I forget.

/cc @bradfitz @soniaappasamy

bradfitz:
Fixes #6303

### DIFF
--- a/ipn/ipnlocal/serve.go
+++ b/ipn/ipnlocal/serve.go
@@ -93,7 +93,9 @@ func (s *serveListener) Run() {
 	for {
 		ln, err := net.Listen("tcp", s.ap.String())
 		if err != nil {
-			s.logf("serve failed to listen on %v, backing off: %v", s.ap, err)
+			if s.shouldWarnAboutListenError(err) {
+				s.logf("serve failed to listen on %v, backing off: %v", s.ap, err)
+			}
 			s.bo.BackOff(s.ctx, err)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


shayne (approved):
Thanks for this!

bradfitz:
Fair warning, though: I only barely tested it. I tested that HasIP works at least. But I didn't test it on WSL2. Want to give it a shot?

ipn/ipnlocal/serve.go
```diff
@@ -111,6 +113,17 @@ func (s *serveListener) Run() {
 	}
 }
 
+func (s *serveListener) shouldWarnAboutListenError(err error) bool {
+	if !s.b.e.GetLinkMonitor().InterfaceState().HasIP(s.ap.Addr()) {
+		// Machine doesn't have IPv6 enabled (or the IP is still being
```
maisem, line 118: Although the comment says IPv6, this code doesn't do any special handling for that.

maisem (approved):

ipn/ipnlocal/serve.go
bradfitz: Yeah, fair. It also would end up handling the userspace-networking path for IPv4, but we stopped that earlier (no need to even get this far if we know we're in userspace mode)

PR #6315, opened 2022-11-14
all: store NL keys per profile

maisem:
This moves the NetworkLock key from a dedicated StateKey to be part of the persist.Persist struct.
This struct is stored as part for ipn.Prefs and is also the place where we store the NodeKey.

It also moves the ChonkDir from "/tka" to "/tka-profile/<profile-id>". The rename was intentional
to be able to delete the "/tka" dir if it exists.

This means that we will have a unique key per profile, and a unique directory per profile.

Note: `tailscale logout` will delete the entire profile, including any keys. It currently does not
delete the ChonkDir.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -159,7 +159,6 @@ type LocalBackend struct {
 	ccAuto         *controlclient.Auto // if cc is of type *controlclient.Auto
 	inServerMode   bool
 	machinePrivKey key.MachinePrivate
-	nlPrivKey      key.NLPrivate
 	tka            *tkaState
 	state          ipn.State
 	capFileSharing bool // whether netMap contains the file sharing capability
@@ -832,6 +831,9 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 		if !prefs.Persist.View().Equals(*st.Persist) {
 			prefsChanged = true
 			prefs.Persist = st.Persist.AsStruct()
+			if err := b.initTKALocked(); err != nil {
+				b.logf("initTKALocked: %v", err)
+			}
 		}
 	}
 	if st.URL != "" {
@@ -1174,9 +1176,6 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 			return fmt.Errorf("initMachineKeyLocked: %w", err)
 		}
 	}
-	if err := b.initNLKeyLocked(); err != nil {
-		return fmt.Errorf("initNLKeyLocked: %w", err)
-	}
 
 	loggedOut := prefs.LoggedOut()
 
@@ -1232,7 +1231,6 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 	// but it won't take effect until the next Start().
 	cc, err := b.getNewControlClientFunc()(controlclient.Options{
 		GetMachinePrivateKey: b.createGetMachinePrivateKeyFunc(),
-		GetNLPrivateKey:      b.createGetNLPrivateKeyFunc(),
 		Logf:                 logger.WithPrefix(b.logf, "control: "),
 		Persist:              *persistv,
 		ServerURL:            b.serverURL,
@@ -1759,21 +1757,6 @@ func (b *LocalBackend) createGetMachinePrivateKeyFunc() func() (key.MachinePriva
 	}
 }
 
-func (b *LocalBackend) createGetNLPrivateKeyFunc() func() (key.NLPrivate, error) {
-	var cache syncs.AtomicValue[key.NLPrivate]
-	return func() (key.NLPrivate, error) {
-		b.mu.Lock()
-		defer b.mu.Unlock()
-		if v, ok := cache.LoadOk(); ok {
-			return v, nil
-		}
-
-		priv := b.nlPrivKey
-		cache.Store(priv)
-		return priv, nil
-	}
-}
-
 // initMachineKeyLocked is called to initialize b.machinePrivKey.
 //
 // b.prefs must already be initialized.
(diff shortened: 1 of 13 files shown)


types/persist/persist.go
```diff
@@ -36,6 +36,7 @@ type Persist struct {
 	Provider          string
 	LoginName         string
 	UserProfile       tailcfg.UserProfile
+	NetworkLockKey    key.NLPrivate
```
bradfitz, line 39: Clarify on the comment on line 18/19 where on disk this is stored? Is this still a mix of Windows user dir and tailscaled backend storage?

Can we remove LegacyFrontendPrivateMachineKey yet?

ipn/ipnlocal/local.go
```diff
@@ -4048,11 +3980,52 @@ func (b *LocalBackend) SwitchProfile(profile ipn.ProfileID) error {
...
+		return nil
+	}
+	if b.tka != nil && b.tka.profile == cp.ID {
+		// Already initialized.
+		return nil
+	}
+	root := b.TailscaleVarRoot()
+	if root == "" {
+		b.logf("network-lock unavailable; no state directory")
+		return nil
```
bradfitz, line 3996: do you have to `b.tka = nil` in this failure path too?

maisem: No, it can never have been initialized if the `TailscaleVarRoot` is empty.

maisem: but sure.

twitchyliquid64:
I'll review this today - theres a small list of coupling points I want to go through carefully.

maisem:
> I'll review this today - theres a small list of coupling points I want to go through carefully.

I found an issue with the ChonkDir, I didn't realize the path was being calculated in multiple places. Working on that right now.

types/persist/persist.go
maisem: Done.

>Can we remove LegacyFrontendPrivateMachineKey yet?

¯\\_(ツ)_/¯

ipn/ipnlocal/local.go
```diff
@@ -4048,11 +3980,57 @@ func (b *LocalBackend) SwitchProfile(profile ipn.ProfileID) error {
...
+	if cp.ID == "" {
+		b.tka = nil
+		return nil
+	}
+	if b.tka != nil {
+		if b.tka.profile == cp.ID {
+			// Already initialized.
+			return nil
+		}
+		b.tka = nil // reset
```
twitchyliquid64, line 3994: Its not clear to me what this does: Is this meant to clear a TKA state from a previous profile once you switch?

ipn/ipnlocal/network-lock.go
```diff
@@ -330,16 +332,17 @@ func (b *LocalBackend) NetworkLockStatus() *ipnstate.NetworkLockStatus {
...
+	var nlPriv key.NLPrivate
 	if p := b.pm.CurrentPrefs(); p.Valid() {
 		nkp := p.Persist().PublicNodeKey()
 		nodeKey = &nkp
+		nlPriv = p.Persist().NetworkLockKey
 	}
 
-	if b.tka == nil {
+	if b.tka == nil || nlPriv.IsZero() {
 		return &ipnstate.NetworkLockStatus{
```
twitchyliquid64, line 343: Its important that we display the profiles network-lock key when its not initialized, someone might need it when constructing a `tailscale lock init` invocation.

Can we update the logic to always return it in `NetworkLockStatus` if its set, rather than combining the check with `b.tka == nil` ?

ipn/ipnlocal/network-lock.go
```diff
@@ -522,8 +534,15 @@ func (b *LocalBackend) NetworkLockModify(addKeys, removeKeys []tka.Key) (err err
 	if b.tka == nil {
 		return errNetworkLockNotActive
 	}
+	var nlPriv key.NLPrivate
+	if p := b.pm.CurrentPrefs().Persist(); p != nil {
+		nlPriv = p.NetworkLockKey
+	}
+	if nlPriv.IsZero() {
+		return errMissingNetmap
```
twitchyliquid64, line 542: Either return `errNetworkLockNotActive` or customize the error, because the invariant of `b.tka != nil` with `nlPriv.IsZero()` should not be possible (or at least I really hope so).

ipn/ipnlocal/network-lock.go
maisem: Oh yeah you're right. When I wrote this I didn't fully realize what b.tka == nil meant. Will fix
Good catch!

twitchyliquid64 (approved):
LGTM from a network-lock side!

My only request is you run the corp `./ipn` tests as theres a lot of coverage there, but otherwise looks dope, :shipit:

ipn/ipnlocal/network-lock.go
maisem: Good point, I reordered the checks so that we verify `nlPriv.IsZero` before we check for `b.tka == nil`

ipn/ipnlocal/local.go
maisem: yep, added a comment.

maisem:
> My only request is you run the corp `./ipn` tests as theres a lot of coverage there, but otherwise looks dope, :shipit:

Tests pass now! 🎉

ipn/ipnlocal/network-lock.go
```diff
@@ -280,7 +281,10 @@ func (b *LocalBackend) tkaBootstrapFromGenesisLocked(g tkatype.MarshaledAUM) err
 		return fmt.Errorf("reading genesis: %v", err)
 	}
 
-	chonkDir := b.chonkPath()
+	chonkDir := b.chonkPathLocked()
```
twitchyliquid64, line 284: Nit: is there are reason we aren't doing `MkDirAll` ?

twitchyliquid64 (approved):

ipn/ipnlocal/network-lock.go
maisem: No real reason, I was just unsure if I wanted the `TailscaleVarRoot` be created with `0755`. Elsewhere it is created with 0700.

Issue #3137, opened 2021-10-18
Windows: Clean up COM initialization in tailscaled


dblohm7:
### What is the issue?

Last week I uncovered evidence that we are initializing single-threaded COM apartments in tailscaled, which we should never be doing. The reasoning is rather long-winded, so while I am happy to discuss with interested parties and document elsewhere, it is far too verbose to be included here.

Not only should we get rid of the STAs, but we should also be initializing COM on a process-wide basis early in the lifetime of the service. Doing so would give every thread implicit membership in COM's multi-threaded apartment, which among other important reasons, also means that we do not need to lock the OS thread every time we touch a COM API.

### OS

Windows

PR #6323, opened 2022-11-15
cmd/tailscaled, wgengine/router: use wingoes/com for COM initializati…

dblohm7:
…on instead of go-ole

This patch removes the crappy, half-backed COM initialization used by `go-ole` and replaces that with the `StartRuntime` function from `wingoes`, a library I have started which, among other things, initializes COM properly.

In particular, we should always be initializing COM to use the multithreaded apartment. Every single OS thread in the process becomes implicitly initialized as part of the MTA, so we do not need to concern ourselves as to whether or not any particular OS thread has initialized COM. Furthermore, we no longer need to lock the OS thread when calling methods on COM interfaces.

Single-threaded apartments are designed solely for working with Win32 threads that have a message pump; any other use of the STA is invalid.

Fixes https://github.com/tailscale/tailscale/issues/3137


dblohm7:
Some additional comments re: wingoes:

* Many moons ago I indicated to Denton that I was working on a Windows-specific library and asked him where he thought it should go. He was fine with it being under my personal GitHub.
* It implements its own `Error` type which encodes all the things in `HRESULT`s. This enables us to differentiate between successful "error codes" like `S_FALSE` and genuine errors. Of course, it implements `error` so that, in higher-level code, we can write functions and methods that are more idiomatic.
* `StartRuntime` is the accumulation of both readings (_Essential COM_) and lessons learned initializing COM in Firefox (I used to own all COM-related stuff in Gecko). When starting COM, you really need to do a bunch more than just `CoInitializeEx` at the beginning. You also need to initialize security and disable the catch-all exception handler in a very specific order. Finally, we shut off the `BSTR` cache because that hasn't been useful since the 90's, and we want to get that thing turned off before any dependencies inadvertently load `oleaut32.dll` and start allocating anything.
* A lot of the code in the `com` package is foundational stuff to prepare for a larger-scope project that generates bindings from `winmd` metadata. But you can get the gist of what is going on there: it mostly abstracts away an interface's reference count such that copying an object is perfectly safe in Go, without needing to fiddle with reference counting -- the GC itself releases the reference via a finalizer.

raggi (approved):

raggi:
Looks good. One consideration: keeping the init in the main package means it'll run almost last. It's a bit of a pain because we parameterize the initialization here which means it'll always need to be downstream of main init. Given the considerations mentioned in the description it could be ideal if we were able to arrange at least the MTA init and BSTR cache "up front" using something like an unnamed include and then do the security initialization in a convenient spot. This might look something like:

```go
import _ "github.com/dblohm7/wingoes/com/initmta"
```

The guts of which might be:

```go
func init() {
	must(startMTAImplicitly())
	must(globalOpts.Set(COMGLB_EXCEPTION_HANDLING, COMGLB_EXCEPTION_DONOT_HANDLE_ANY))
	setOaNoCache()
}
```

dblohm7:
> Looks good. One consideration: keeping the init in the main package means it'll run almost last. 

Hmm, good point. I'll definitely put some thought into this. Thanks!

PR #6324, opened 2022-11-15
ipn/ipnlocal: fix controlclient reentrancy when deleting profile during logout

mihaip:
Deletion of profiles on logout (#6297) added a `LocalBackend.Start()` call within `setClientStatus()`, but that's a callback from the controlclient. `Start()` ends calling back into the controlclient (to shut it down), and we end up stuck in a deadlock waiting for the `authDone` channel to be closed.

Fixed by making the Start call asynchronous. To reproduce this in a test case, we need to do an asynchronous logout, so add a CLI option for that.

### DIFF
--- a/client/tailscale/localclient.go
+++ b/client/tailscale/localclient.go
@@ -548,6 +548,11 @@ func (lc *LocalClient) Logout(ctx context.Context) error {
 	return err
 }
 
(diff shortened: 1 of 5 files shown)


mihaip:
Superseded by #6326

PR #6325, opened 2022-11-15
cmd/tailscale: access QNAP via localhost

DentonGentry:
QNAP 5.x works much better if we let Apache proxy
tailscale web, which means the URLs can no longer
be relative since apache sends us an internal
URL. Access QNAP authentication via
http://localhost:8080/ as documented in
https://download.qnap.com/dev/API_QNAP_QTS_Authentication.pdf

### DIFF
--- a/cmd/tailscale/cli/web.go
+++ b/cmd/tailscale/cli/web.go
@@ -224,8 +224,8 @@ func qnapAuthnQtoken(r *http.Request, user, token string) (string, *qnapAuthResp
 		"user":   []string{user},
 	}
 	u := url.URL{
-		Scheme:   r.URL.Scheme,
-		Host:     r.URL.Host,
+		Scheme:   "http",
+		Host:     "127.0.0.1:8080",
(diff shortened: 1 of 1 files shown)


cmd/tailscale/cli/web.go
```diff
@@ -224,8 +224,8 @@ func qnapAuthnQtoken(r *http.Request, user, token string) (string, *qnapAuthResp
 		"user":   []string{user},
 	}
 	u := url.URL{
-		Scheme:   r.URL.Scheme,
-		Host:     r.URL.Host,
+		Scheme:   "http",
```
maisem, line 227: Should this be conditional on the qnap version?

DentonGentry: In https://github.com/tailscale/tailscale-qpkg/tree/dgentry I gave up on QTS 4 support after QNAP pushed 5.0 to the unit even though I kept refusing it. I don't have a way to test it, there are a number of things which don't work on QTS 4.

maisem (approved):

PR #6326, opened 2022-11-15
ipn/ipnlocal: do controlclient.Shutdown in a different goroutine

maisem:
We do not need to wait for it to complete. And we might have to call Shutdown from callback from the controlclient which might already be holding a lock that Shutdown requires.

Updates #713

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -1141,9 +1141,7 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 		// into sync with the minimal changes. But that's not how it
 		// is right now, which is a sign that the code is still too
 		// complicated.
-		b.mu.Unlock()
-		b.cc.Shutdown()
-		b.mu.Lock()
+		b.resetControlClientLockedAsync()
 	}
 	httpTestClient := b.httpTestClient
 
@@ -1956,7 +1954,7 @@ func (b *LocalBackend) InServerMode() bool {
 }
 
 // Login implements Backend.
-// As of 2022-02-17, this is only exists for tests.
+// As of 2022-11-15, this is only exists for Android.
 func (b *LocalBackend) Login(token *tailcfg.Oauth2Token) {
 	b.mu.Lock()
 	b.assertClientLocked()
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/local.go
```diff
@@ -4053,6 +4058,7 @@ func (b *LocalBackend) resetForProfileChangeLockedOnEntry() error {
 	b.setNetMapLocked(nil) // Reset netmap.
 	// Reset the NetworkMap in the engine
 	b.e.SetNetworkMap(new(netmap.NetworkMap))
+	b.resetControlClientLocked()
```
mihaip, line 4061: Is this necessary? We end up calling `resetControlClientLocked` in `Start` either way.

ipn/ipnlocal/local.go
```diff
@@ -3238,6 +3236,17 @@ func (b *LocalBackend) requestEngineStatusAndWait() {
 	b.statusLock.Unlock()
 }
 
+func (b *LocalBackend) resetControlClientLocked() {
+	if b.cc == nil {
+		return
+	}
+	// Call Shutdown in a goroutine, as it may require
+	// locks that we're holding. We do not need to wait
```
mihaip, line 3244: Is it locks that we're holding? Based on the stack that I observed, the deadlock was contained with the control client itself.

mihaip (approved):
This is a lot simpler than #6324, so it seems preferable.

ipn/ipnlocal/local.go
maisem: I thought it was a lock but I could be wrong, reworded regardless.

ipn/ipnlocal/local.go
maisem: Good point. removed.

maisem:
@mihaip @bradfitz I had to modify TestStateMachine as it would never recreate the mockControl client and expected "Shutdown" and "New" events in order. However as we are now calling Shutdown in a goroutine, there was no way to guarantee the ordering, so instead I made the test also recreate the client when it was Shutdown. To ensure that shutdown was still being invoked I keep track of the previous client and make sure it is shutdown.

I also removed all the unnecessary "pause"/"unpause" and now only do them when it toggles state. This allowed me to remove all the places where we had `"unpause", "unpause", "unpause"`

bradfitz (approved):

Issue #5209, opened 2022-07-28
tailssh: race in x/crypto/ssh


andrew-d:
[This](https://github.com/tailscale/tailscale/runs/7564086514?check_suite_focus=true) test run has a data race in tailssh, which boils down to a race between this read and write in `x/crypto/ssh`:
https://github.com/golang/crypto/blob/0b941c09a5e1/ssh/channel.go#L233
https://github.com/golang/crypto/blob/0b941c09a5e1/ssh/channel.go#L542

Triggered by these two goroutines:
https://github.com/tailscale/tailscale/blob/a12aad6/ssh/tailssh/tailssh.go#L1057
https://github.com/tailscale/tailscale/blob/a12aad6/ssh/tailssh/tailssh.go#L1063

cc @bradfitz @maisem

bradfitz:
For posterity and saving clicks:

```
==================
WARNING: DATA RACE
Read at 0x00c0002541e0 by goroutine 74:
  github.com/tailscale/golang-x-crypto/ssh.(*channel).WriteExtended()
      /home/runner/go/pkg/mod/github.com/tailscale/golang-x-crypto@v0.0.0-20220428210705-0b941c09a5e1/ssh/channel.go:233 +0x6c
  github.com/tailscale/golang-x-crypto/ssh.(*extChannel).Write()
      /home/runner/go/pkg/mod/github.com/tailscale/golang-x-crypto@v0.0.0-202204[282](https://github.com/tailscale/tailscale/runs/7564086514?check_suite_focus=true#step:5:283)10705-0b941c09a5e1/ssh/channel.go:484 +0x78
  io.copyBuffer()
      /opt/hostedtoolcache/go/1.18.4/x64/src/io/io.go:428 +0x2dc
  io.Copy()
      /opt/hostedtoolcache/go/1.18.4/x64/src/io/io.go:385 +0xad
  tailscale.com/ssh/tailssh.(*sshSession).run.func4()
      /home/runner/work/tailscale/tailscale/ssh/tailssh/tailssh.go:1063 +0x37

Previous write at 0x00c0002541e0 by goroutine 73:
  github.com/tailscale/golang-x-crypto/ssh.(*channel).CloseWrite()
      /home/runner/go/pkg/mod/github.com/tailscale/golang-x-crypto@v0.0.0-20220428210705-0b941c09a5e1/ssh/channel.go:542 +0x4a
[...]

PR #6330, opened 2022-11-15
ssh/tailssh: only call CloseWrite when both stdout and stderr are done

maisem:
Updates #5209

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -28,6 +28,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	gossh "github.com/tailscale/golang-x-crypto/ssh"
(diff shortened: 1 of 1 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -1069,13 +1070,16 @@ func (ss *sshSession) run() {
 			ss.ctx.CloseWithError(err)
 		}
 	}()
+	var openOutputStreams atomic.Int32
+	openOutputStreams.Store(2)
```
mihaip, line 1074: We only conditionally call `io.Copy` for stderr (it may be nil), so it seems like the starting value could be either 1 or 2.

ssh/tailssh/tailssh.go
```diff
@@ -1069,13 +1070,20 @@ func (ss *sshSession) run() {
 			ss.ctx.CloseWithError(err)
 		}
 	}()
+	var openOutputStreams atomic.Int32
+	if ss.stderr != nil {
```
mihaip, line 1074: Getting nitpicky, but instead of repeating the conditional here, why not do an `.Add(1)` in each of the goroutines where you start the copy?

mihaip (approved):

ssh/tailssh/tailssh.go
maisem: setting this in the goroutines individually is racy as one goroutine might run before the other and potentially even run to completion.
e.g.
gr1 - Add(1)
gr1 - error copying
gr1 - Add(-1)
gr1 - Val == 0 => CloseWrite

gr2 - Add(1)
gr2 - error copying
gr2 - Add(-1)
gr2 - Val == 0 => CloseWrite

PR #6334, opened 2022-11-15
health, ipn/ipnlocal: when -no-logs-no-support is enabled, deny acces…

dblohm7:
…s to tailnets that have network logging enabled

We want users to have the freedom to start tailscaled with `-no-logs-no-support`, but that is obviously in direct conflict with tailnets that have network logging enabled.

When we detect that condition, we record the issue in health, notify the client, and then do a disconnect without logout.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -908,6 +908,21 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 	}
 
 	if st.NetMap != nil {
+		if envknob.NoLogsNoSupport() && hasCapability(st.NetMap, tailcfg.CapabilityDataPlaneAuditLogs) {
+			msg := "tailnet requires logging to be enabled. Remove -no-logs-no-support from tailscaled command line."
+			health.SetLocalLogConfigHealth(errors.New(msg))
+			// Connecting to this tailnet without logging is forbidden; boot us outta here.
+			b.mu.Lock()
+			prefs.WantRunning = false
+			p := prefs.View()
+			if err := b.pm.SetPrefs(p); err != nil {
+				b.logf("Failed to save new controlclient state: %v", err)
+			}
+			b.mu.Unlock()
+			np := stripKeysFromPrefs(p)
+			b.send(ipn.Notify{ErrMessage: &msg, Prefs: &np})
+			return
+		}
 		if netMap != nil {
 			diff := st.NetMap.ConciseDiffFrom(netMap)
 			if strings.TrimSpace(diff) == "" {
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/local.go
```diff
@@ -892,6 +892,13 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 	}
 
 	if st.NetMap != nil {
+		if envknob.NoLogsNoSupport() && hasCapability(st.NetMap, tailcfg.CapabilityDataPlaneAuditLogs) {
+			msg := "tailnet requires logging to be enabled. Remove -no-logs-no-support from tailscaled command line."
+			health.SetLocalLogConfigHealth(errors.New(msg))
+			b.send(ipn.Notify{ErrMessage: &msg})
+			b.ResetForClientDisconnect()
```
maisem, line 899: This has weird implications and I can't reason about them without reading the code in depth.
Can we force "WantRunning=false" instead?

dblohm7: I looked into this and TBH just setting `WantRunning = false` results in a worse experience than doing what we do with the current version of the patch.

Specifically, this iteration appears to work intuitively in the sense that, after removing `-no-logs-no-support` and restarting `tailscaled`, everything works again. `WantRunning = false` doesn't recover as nicely from this case -- in particular it looks like it necessitates that the user log in again.

At any rate, I'm not certain that we should be concerning ourselves with getting into a recoverable state after `-no-logs-no-support` is removed. I would suggest that any use of  `-no-logs-no-support` also implies that any unusual state that the client inadvertently ends up in (even after that flag's removal) is also unsupported. (Side note: maybe we should add some kind of permanent marker to a node's config to mark if that flag was ever turned on, even if it has since been turned off? That might be useful from a support perspective.) IMHO I think the important part is that we boot ourselves from the tailnet when necessary and provide a reasonable amount of explanation for that.

maisem: Oooh that's a good point.
Can you see what happens when you switch between tailnets that have audit logging enabled (using `tailscale login` and `tailscale switch`)

Regarding `ResetForClientDisconnect`, it is only used on windows when the frontend disconnects from backend. It is driven by `ipnserver.Server` which owns state. Reseting the client here might cause weird state inconsistency issues. I think you'll need a new marker for this.

dblohm7: Okay, so I was able to get your `WantRunning=false` suggestion working after all. Switching works too.

ipn/ipnlocal/local.go
```diff
@@ -908,6 +908,20 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
...
+			health.SetLocalLogConfigHealth(errors.New(msg))
+			// Connecting to this tailnet without logging is forbidden; boot us outta here.
+			b.mu.Lock()
+			prefs.WantRunning = false
+			p := prefs.View()
+			if err := b.pm.SetPrefs(p); err != nil {
+				b.logf("Failed to save new controlclient state: %v", err)
+			}
+			b.mu.Unlock()
+			b.send(ipn.Notify{ErrMessage: &msg, Prefs: &p})
```
maisem, line 922: use the new `b.notifyPrefs` method, it strips out information that should not be sent. Or use `stripKeysFromPrefs`.

maisem (approved):
LGTM after addressing the `stripKeysFromPrefs` comment

PR #6336, opened 2022-11-15
types/logid: move logtail ID types here

dsnet:
Many packages reference the logtail ID types,
but unfortunately pull in the transitive dependencies of logtail. Fix this problem by putting the log ID types in its own package with minimal dependencies.


logtail/id.go
```diff
@@ -4,180 +4,23 @@
...
-import (
-	"crypto/rand"
-	"crypto/sha256"
-	"encoding/binary"
-	"encoding/hex"
-	"errors"
-	"fmt"
-)
+// Deprecated: Use "tailscale.com/types/logid".PrivateID instead.
+type PrivateID = logid.PrivateID
```
dblohm7, line 8: looks like you're missing an import...

dblohm7 (approved):
OK with comment addressed + clean CI

logtail/id.go
dsnet: Foxed.

PR #6339, opened 2022-11-16
ipn, tailcfg: add Prefs, Hostinfo bool for whether ingress is on at all

bradfitz:
Optimization for control.

Updates tailscale/corp#7515

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -2193,6 +2193,20 @@ func (b *LocalBackend) SetPrefs(newp *ipn.Prefs) {
 	b.setPrefsLockedOnEntry("SetPrefs", newp)
 }
 
+// wantIngressLocked reports whether this node has ingress configured. This bool
+// is sent to the coordination server (in Hostinfo.WireIngress) as an
+// optimization hint to know primarily which nodes are NOT using ingress, to
+// avoid doing work for regular nodes.
+//
+// Even if the user's ServeConfig.AllowIngress map was manually edited in raw
+// mode and contains map entries with false values, sending true (from Len > 0)
+// is still fine. This is only an optimization hint for the control plane and
+// doesn't affect security or correctness. And we also don't expect people to
+// modify their ServeConfig in raw mode.
+func (b *LocalBackend) wantIngressLocked() bool {
+	return b.serveConfig.Valid() && b.serveConfig.AllowIngress().Len() > 0
+}
+
 // setPrefsLockedOnEntry requires b.mu be held to call it, but it
 // unlocks b.mu when done. newp ownership passes to this function.
 // It returns a readonly copy of the new prefs.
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/local.go
```diff
@@ -2971,6 +2975,14 @@ func (b *LocalBackend) applyPrefsToHostinfo(hi *tailcfg.Hostinfo, prefs ipn.Pref
...
 	}
 	hi.SSH_HostKeys = sshHostKeys
+
+	// The Hostinfo.WantIngress field tells control whether this node wants to
+	// be wired up for ingress connections. If harmless if it's accidentally
+	// true; the actual policy is controlled in tailscaled by ServeConfig. But
+	// if this is accidentally false, then control may not configure DNS
+	// properly. This exists as an optimization to control to program fewer DNS
+	// records that have ingress enabled but are not actually being used.
+	hi.WireIngress = b.wantIngressLocked() || prefs.WireIngress()
```
maisem, line 2985: Why is this not &&. Having two ways of controlling the same bit sounds scary

bradfitz: It's a hedge against the logic being wrong. The pref should be right and synced with wantIngressLocked (serveConfig), though. There's no downside to accidentally sending true. It's just an optimization for control.

There is, however, a downside to accidentally sending the wrong value: it doesn't get wired up by control.
 
So I'm erring on sending true.

Currently we send nothing and effectively treat all nodes as sending true.

bradfitz: PTAL. Redid it all.

bradfitz:
Ugh, but I just realized this can send a lite map update on the Hostinfo change, which means the control-side isStreaming checks won't update for a map poll already in flight...

```
func (c *Auto) SetHostinfo(hi *tailcfg.Hostinfo) {
	if hi == nil {
		panic("nil Hostinfo")
	}
	if !c.direct.SetHostinfo(hi) {
		// No changes. Don't log.
		return
	}

	// Send new Hostinfo to server
	c.sendNewMapRequest()
}
```

I guess it'd need to check whether WireIngress changed in Auto.SetHostinfo and then go into this path:

```
	c.mu.Lock()

	// If we're not already streaming a netmap, or if we're already stuck
	// in a lite update, then tear down everything and start a new stream
	// (which starts by sending a new map request)
	if !c.inPollNetMap || c.inLiteMapUpdate || !c.loggedIn {
		c.mu.Unlock()
		c.cancelMapSafely()
		return
	}
```

ipn/ipnlocal/local.go
```diff
@@ -2193,6 +2193,20 @@ func (b *LocalBackend) SetPrefs(newp *ipn.Prefs) {
...
+// optimization hint to know primarily which nodes are NOT using ingress, to
+// avoid doing work for regular nodes.
+//
+// Even if the user's ServeConfig.AllowIngress map was manually edited in raw
+// mode and contains map entries with false values, sending true (from Len > 0)
+// is still fine. This is only an optimization hint for the control plane and
+// doesn't affect security or correctness. And we also don't expect people to
+// modify their ServeConfig in raw mode.
+func (b *LocalBackend) wantIngressLocked() bool {
+	return b.serveConfig.Valid() && b.serveConfig.AllowIngress().Len() > 0
```
maisem, line 2207: Oh I don't think serveConfig is reset for profile changes

maisem:
Updated in https://github.com/tailscale/tailscale/pull/6358

bradfitz:
Superseded by https://github.com/tailscale/tailscale/pull/6358

PR #6346, opened 2022-11-16
client/tailscale/apitype: add LocalAPIHost const, use it

bradfitz:
Removes duplication.

### DIFF
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -34,6 +34,7 @@ import (
 	"tailscale.com/ipn/ipnlocal"
 	"tailscale.com/ipn/ipnstate"
 	"tailscale.com/net/netutil"
+	"tailscale.com/safesocket"
 	"tailscale.com/tailcfg"
 	"tailscale.com/tka"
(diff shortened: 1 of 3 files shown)


ipn/localapi/localapi.go
```diff
@@ -156,6 +161,24 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+	// Otherwise, any Host header we see should at most be an ip:port.
+	ap, err := netip.ParseAddrPort(h)
+	if err != nil {
+		return false
+	}
+	if runtime.GOOS == "windows" && ap.Port() != safesocket.WindowsLocalPort {
+		return false
+	}
+	return err == nil && ap.Addr().IsLoopback()
```
maisem, line 179: err is guaranteed `nil` here

maisem (approved):

ipn/localapi/localapi.go
bradfitz: done

DentonGentry (approved):

PR #6350, opened 2022-11-16
ipn/ipnlocal: move LocalBackend.validPopBrowserURL empty check from caller

bradfitz:
I was too late with review feedback to 513780f4f8.

Updates tailscale/corp#7948


maisem (approved):
Heh, I was going to send this after seeing your comments. Thanks!

PR #6352, opened 2022-11-16
ipn/ipnlocal: add some validation to PeerAPI

maisem:
Updates tailscale/corp#7948

### DIFF
--- a/ipn/ipnlocal/peerapi.go
+++ b/ipn/ipnlocal/peerapi.go
@@ -32,6 +32,7 @@ import (
 	"unicode/utf8"
 
 	"github.com/kortschak/wol"
+	"golang.org/x/exp/slices"
 	"golang.org/x/net/dns/dnsmessage"
 	"tailscale.com/client/tailscale/apitype"
 	"tailscale.com/envknob"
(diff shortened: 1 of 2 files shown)


bradfitz (approved):
add test case for Host "peer"

danderson (approved):

ipn/ipnlocal/peerapi.go
```diff
@@ -542,7 +543,30 @@ func (h *peerAPIHandler) logf(format string, a ...any) {
...
 
+// validPeerAPIRequest reports whether r is a valid PeerAPI request.
+func (h *peerAPIHandler) validPeerAPIRequest(r *http.Request) bool {
+	if r.Referer() != "" {
+		return false
+	}
+	if r.Header.Get("Origin") != "" {
+		return false
+	}
+	if r.Host == "peer" {
```
danderson, line 554: I would put this logic in a helper e.g. `validHost`. Then if we add more tests later on, this early return true doesn't end up skipping them.

danderson (approved):

ipn/ipnlocal/peerapi.go
maisem: good idea, done.

bradfitz: in the signature I sent you on chat, it was one method but it returned an error so tests could validate the expected error path fired

maisem: Made it return an error instead. PTAL

danderson (approved):

ipn/ipnlocal/peerapi.go
bradfitz: but this part doesn't return an error. it can fail for two reasons.

just move this back into the other method and return an error in all bad paths?

maisem: moving it back reduces readability when combined with "Then if we add more tests later on, this early return true doesn't end up skipping them." I made `validateHost` also return an error so that it's clear what the error is.

ipn/ipnlocal/peerapi.go
```diff
@@ -542,7 +543,37 @@ func (h *peerAPIHandler) logf(format string, a ...any) {
...
+	}
+	ap, err := netip.ParseAddrPort(r.Host)
+	if err != nil {
+		return err
+	}
+	hostIPPfx := netip.PrefixFrom(ap.Addr(), ap.Addr().BitLen())
+	if slices.Contains(h.ps.selfNode.Addresses, hostIPPfx) {
+		return nil
+	}
+	return fmt.Errorf("invalid Host header %q", r.Host)
```
bradfitz, line 558: we know it's an ip:port now. No need for %q. The reason is:  hostIPPfx not found in self addresses.

I'd invert the if above and make the failure paths be indented, then end in "return nil"

ipn/ipnlocal/peerapi.go
```diff
@@ -542,7 +543,37 @@ func (h *peerAPIHandler) logf(format string, a ...any) {
...
+	}
+	if r.Header.Get("Origin") != "" {
+		return errors.New("unexpected Origin")
+	}
+	return h.validateHost(r)
+}
+
 func (h *peerAPIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if err := h.validatePeerAPIRequest(r); err != nil {
+		h.logf("invalid request from %v: %v", h.remoteAddr, err)
```
bradfitz, line 573: I forget: does this h.logf already have a "peerapi: " prefix wrapped?

maisem: yeah

bradfitz (approved):

PR #6354, opened 2022-11-16
tailcfg: add CapabilityDebug

maisem:
Updates tailscale/corp#7948

### DIFF
--- a/ipn/ipnlocal/peerapi.go
+++ b/ipn/ipnlocal/peerapi.go
@@ -778,6 +778,17 @@ func (h *peerAPIHandler) canPutFile() bool {
 // canDebug reports whether h can debug this node (goroutines, metrics,
 // magicsock internal state, etc).
 func (h *peerAPIHandler) canDebug() bool {
+	// Reread the selfNode as it may have changed since the peerAPIServer
+	// was created.
+	// TODO(maisem): handle this in other places too.
+	nm := h.ps.b.NetMap()
+	if nm == nil || nm.SelfNode == nil {
+		return false
+	}
+	if !slices.Contains(nm.SelfNode.Capabilities, tailcfg.CapabilityDebug) {
+		// This node does not expose debug info.
+		return false
+	}
 	return h.isSelf || h.peerHasCap(tailcfg.CapabilityDebugPeer)
 }
 
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -1654,12 +1654,16 @@ type Oauth2Token struct {
 const (
 	// These are the capabilities that the self node has as listed in
 	// MapResponse.Node.Capabilities.
+	//
+	// We've since started referring to these as "Node Attributes" ("nodeAttrs"
+	// in the ACL policy file).
 
 	CapabilityFileSharing        = "https://tailscale.com/cap/file-sharing"
 	CapabilityAdmin              = "https://tailscale.com/cap/is-admin"
 	CapabilitySSH                = "https://tailscale.com/cap/ssh"                   // feature enabled/available
 	CapabilitySSHRuleIn          = "https://tailscale.com/cap/ssh-rule-in"           // some SSH rule reach this node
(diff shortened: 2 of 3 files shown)


tailcfg/tailcfg.go
```diff
@@ -1654,12 +1654,16 @@ type Oauth2Token struct {
...
+	//
+	// We've since started referring to these as "Node Attributes" ("nodeAttrs"
+	// in the ACL policy file).
 
 	CapabilityFileSharing        = "https://tailscale.com/cap/file-sharing"
 	CapabilityAdmin              = "https://tailscale.com/cap/is-admin"
 	CapabilitySSH                = "https://tailscale.com/cap/ssh"                   // feature enabled/available
 	CapabilitySSHRuleIn          = "https://tailscale.com/cap/ssh-rule-in"           // some SSH rule reach this node
 	CapabilityDataPlaneAuditLogs = "https://tailscale.com/cap/data-plane-audit-logs" // feature enabled
+	CapabilityDebug              = "https://tailscale.com/cap/debug"                 // exposes debug endpoints over the PeerAPI
```
bradfitz, line 1666: NodeAttrDebug ?

ipn/ipnlocal/peerapi.go
```diff
@@ -778,6 +778,9 @@ func (h *peerAPIHandler) canPutFile() bool {
 // canDebug reports whether h can debug this node (goroutines, metrics,
 // magicsock internal state, etc).
 func (h *peerAPIHandler) canDebug() bool {
+	if !slices.Contains(h.ps.selfNode.Capabilities, tailcfg.CapabilityDebug) {
```
bradfitz, line 781: when does a new peerAPIServer get created? because ps.selfNode isn't updated in-place or on ps, so if I update ACL noteAttrs server-side, when does this take effect?

bradfitz: Yeah, I think you want to do a call into LocalBackend instead to ask about the current netmap's self node capabilities, rather than trusting this h.ps.selfNode to be recent.

tailcfg/tailcfg.go
bradfitz: likewise, /nodeattr/debug in the "URL"?

bradfitz: but meh, it's fine, nevermind. let's keep it consistent.

ipn/ipnlocal/peerapi.go
maisem: Done, but we should remove ps.selfNode in that case. Will send follow up PR.

ipn/ipnlocal/peerapi.go
```diff
@@ -778,6 +778,13 @@ func (h *peerAPIHandler) canPutFile() bool {
 // canDebug reports whether h can debug this node (goroutines, metrics,
 // magicsock internal state, etc).
 func (h *peerAPIHandler) canDebug() bool {
+	nm := h.ps.b.NetMap()
+	if nm == nil {
```
bradfitz, line 782: nm == nil || nm.SelfNode == nil

ipn/ipnlocal/peerapi.go
```diff
@@ -778,6 +778,13 @@ func (h *peerAPIHandler) canPutFile() bool {
 // canDebug reports whether h can debug this node (goroutines, metrics,
 // magicsock internal state, etc).
 func (h *peerAPIHandler) canDebug() bool {
+	nm := h.ps.b.NetMap()
```
bradfitz, line 781: add a comment about why we're not using selfNode

bradfitz (approved):
LGTM after those comments

ipn/ipnlocal/peerapi.go
maisem: Done.

ipn/ipnlocal/peerapi.go
maisem: Done

PR #6355, opened 2022-11-16
ipn/ipnlocal: move selfNode from peerAPIServer to peerAPIHandler

maisem:
The peerAPIHandler is instantiated per PeerAPI call so it is guaranteed to have the latest selfNode.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -2784,7 +2784,6 @@ func (b *LocalBackend) initPeerAPIListener() {
 	ps := &peerAPIServer{
 		b:                       b,
 		rootDir:                 fileRoot,
-		selfNode:                selfNode,
 		directFileMode:          b.directFileRoot != "",
 		directFileDoFinalRename: b.directFileDoFinalRename,
 	}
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/local.go
```diff
@@ -2784,7 +2784,6 @@ func (b *LocalBackend) initPeerAPIListener() {
 	ps := &peerAPIServer{
 		b:                       b,
 		rootDir:                 fileRoot,
-		selfNode:                selfNode,
```
bradfitz, line 2787: remove the selfNode variable above then? It do anything? just keep the Addresses == 0 check

ipn/ipnlocal/peerapi.go
```diff
@@ -535,6 +541,7 @@ type peerAPIHandler struct {
 	ps         *peerAPIServer
 	remoteAddr netip.AddrPort
 	isSelf     bool                // whether peerNode is owned by same user as this node
+	selfNode   *tailcfg.Node       // this node
```
bradfitz, line 544: this node; always non-nil

bradfitz (approved):

ipn/ipnlocal/local.go
maisem: its used to calculate the fileRoot.

bradfitz: fair nuff

ipn/ipnlocal/peerapi.go
maisem: Done

PR #6357, opened 2022-11-16
ipn/ipnlocal: also accept service IP IPv6 literal in brackets for quad100

bradfitz:
The fix in 4fc8538e2 was sufficient for IPv6. Browsers (can?) send the IPv6 literal, even without a port number, in brackets.

Updates tailscale/corp#7948

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -3925,6 +3925,17 @@ func (b *LocalBackend) HandleQuad100Port80Conn(c net.Conn) {
 	s.Serve(netutil.NewOneConnListener(c, nil))
 }
 
+func validQuad100Host(h string) bool {
+	switch h {
+	case "",
+		tsaddr.TailscaleServiceIPString,
+		tsaddr.TailscaleServiceIPv6String,
+		"[" + tsaddr.TailscaleServiceIPv6String + "]":
+		return true
+	}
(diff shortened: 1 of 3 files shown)


DentonGentry (approved):

ipn/ipnlocal/local.go
```diff
@@ -3925,16 +3925,25 @@ func (b *LocalBackend) HandleQuad100Port80Conn(c net.Conn) {
 	s.Serve(netutil.NewOneConnListener(c, nil))
 }
 
+func validQuad100Host(h string) bool {
+	switch h {
+	case "",
+		tsaddr.TailscaleServiceIPString,
+		tsaddr.TailscaleServiceIPv6String,
+		"[" + tsaddr.TailscaleServiceIPv6String + "]":
```
danderson, line 3933: huh, didn't think that was a valid representation. I need to go read the RFCs again someday.

danderson (approved):

maisem (approved):

ipn/ipnlocal/local.go
bradfitz: well, it's the URL representation for IPv6 literals with colons to not look like a port number.

I didn't think browsers sent it when the implicit port number was used, but I guess they do. Without a port number.

PR #6359, opened 2022-11-16
cmd/{tailscale,tailscaled}: embed manifest into Windows binaries

### LINKED ISSUE #6429: Manifests for tailscale.exe and tailscaled.exe
We should be shipping manifests for these binaries so that we don't get weird behaviour with respect to API support.

andrew-d:
This uses a go:generate statement to create a bunch of .syso files that contain a Windows resource file. We check these in since they're less than 1KiB each, and are only included on Windows.

Fixes #6429 

Change-Id: I0512c3c0b2ab9d8d8509cf2037b88b81affcb81f


bradfitz (commented):
Do we need _386.syso too?

andrew-d:
> Do we need _386.syso too?

@bradfitz - Added, along with `windows/arm` as well (just in case).

cmd/mkmanifest/main.go
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// This is a simple helper utility to create a '.syso' file that contains a
```
bradfitz, line 5: // The mkmanifest command is a ...

or

// The mkmanifest tool is a ...

etc

bradfitz (approved):

cmd/mkmanifest/main.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// Add resources
+	rs.Set(winres.RT_MANIFEST, winres.ID(1), 0, manifest)
+
+	// Compile to a COFF object file
+	f, err := os.Create(out)
+	if err != nil {
+		log.Fatalf("error creating output file %q: %v", out, err)
+	}
+	defer f.Close()
```
bradfitz, line 46: never defer close on a file opened for write ... that's where errors can pop up

dblohm7 (approved):
This is fine, but FYI in the WWG we decided that we were always going to associate our PRs with issues. I have filed one [for this](https://github.com/tailscale/tailscale/issues/6429), but please update your commit message to reference it before merging.

andrew-d:
> This is fine, but FYI in the WWG we decided that we were always going to associate our PRs with issues. I have filed one [for this](https://github.com/tailscale/tailscale/issues/6429), but please update your commit message to reference it before merging.

Done; thanks for creating that.

PR #6361, opened 2022-11-16
ipn/ipnlocal,tka: generate a nonce for each TKA state

twitchyliquid64:
We generate a persistent 16-byte nonce thats fixed for the life of a tailnet key authority. The primary purpose of this is to randomize the genesis AUM hash, but such a value could be useful for future features.

### DIFF
--- a/ipn/ipnlocal/network-lock.go
+++ b/ipn/ipnlocal/network-lock.go
@@ -6,7 +6,9 @@ package ipnlocal
 
 import (
 	"bytes"
+	"crypto/rand"
 	"context"
+	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -413,6 +415,11 @@ func (b *LocalBackend) NetworkLockInit(keys []tka.Key, disablementValues [][]byt
 		return errors.New("no node-key: is tailscale logged in?")
 	}
 
+	var entropy [16]byte
+	if _, err := rand.Read(entropy[:]); err != nil {
+		return err
+	}
+
 	// Generates a genesis AUM representing trust in the provided keys.
 	// We use an in-memory tailchonk because we don't want to commit to
 	// the filesystem until we've finished the initialization sequence,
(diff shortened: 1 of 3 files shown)


tka/state.go
```diff
@@ -36,6 +36,13 @@ type State struct {
 
 	// Keys are the public keys currently trusted by the TKA.
 	Keys []Key `cbor:"3,keyasint"`
+
+	// StateID's are nonce's, generated on enablement and fixed for
+	// the lifetime of the Tailnet Key Authority. We generate 16-bytes
+	// worth of keyspace here just in case we come up with a cool future
+	// use for this.
+	StateID1 uint64 `cbor:"4,keyasint,omitempty"`
```
twitchyliquid64, line 44: btw these are omitempty so I don't have to update all the tests that depend on specific hash values.

sailorfrag: Any particular reason do this as `uint64` instead of `[16]byte` or `[]byte`?

ipn/ipnlocal/network-lock.go
```diff
@@ -424,6 +431,9 @@ func (b *LocalBackend) NetworkLockInit(keys []tka.Key, disablementValues [][]byt
 		//    - DisablementSecret: value needed to disable.
 		//    - DisablementValue: the KDF of the disablement secret, a public value.
 		DisablementSecrets: disablementValues,
+
+		StateID1: binary.LittleEndian.Uint64(entropy[:8]),
```
sailorfrag, line 435: Would it be simpler to do two 8 byte reads instead of one 16 byte read that gets sliced?

sailorfrag (approved):

ipn/ipnlocal/network-lock.go
creachadair: Maybe—but reading from the system entropy pool is potentially expensive, plus you'd then have to check for errors twice.

tka/state.go
twitchyliquid64: No strong reason: I just thought it would be easier to use as a `uint64` and it also encodes a bit nicer as CBOR.

PR #6363, opened 2022-11-17
tstime: fix ParseDuration for '6' digit

dsnet:
The cutset provided to strings.TrimRight was missing the digit '6', making it such that we couldn't parse something like "365d".


bradfitz (approved):
Hahah awesome

mihaip (approved):
LOL, "The Case Of The Missing 6"

PR #6368, opened 2022-11-17
ipn/ipnlocal: handle untagging nodes better

maisem:
We would end up with duplicate profiles for the node as the UserID would have chnaged. In order to correctly deduplicate profiles, we need to look at both the UserID and the NodeID. A single machine can only ever have 1 profile per NodeID and 1 profile per UserID.

Note: UserID of a Node can change when the node is tagged/untagged, and the NodeID of a device can change when the node is deleted so we need to check for both.

Updates #713

### DIFF
--- a/ipn/ipnlocal/profiles.go
+++ b/ipn/ipnlocal/profiles.go
@@ -70,13 +70,30 @@ func (pm *profileManager) SetCurrentUser(uid string) error {
 	return nil
 }
 
-func (pm *profileManager) findProfileByUserID(userID tailcfg.UserID) *ipn.LoginProfile {
+func (pm *profileManager) findProfilesByNodeID(nodeID tailcfg.StableNodeID) []*ipn.LoginProfile {
+	if nodeID.IsZero() {
+		return nil
+	}
+	var out []*ipn.LoginProfile
 	for _, p := range pm.knownProfiles {
-		if p.UserProfile.ID == userID {
-			return p
+		if p.NodeID == nodeID {
+			out = append(out, p)
 		}
 	}
-	return nil
+	return out
+}
+
+func (pm *profileManager) findProfilesByUserID(userID tailcfg.UserID) []*ipn.LoginProfile {
+	if userID.IsZero() {
+		return nil
+	}
+	var out []*ipn.LoginProfile
+	for _, p := range pm.knownProfiles {
+		if p.UserProfile.ID == userID {
+			out = append(out, p)
+		}
+	}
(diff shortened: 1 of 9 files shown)


ipn/ipnlocal/profiles.go
```diff
@@ -124,37 +141,45 @@ func init() {
...
+	if newPersist == nil || newPersist.LoginName == "" {
 		return pm.setPrefsLocked(prefs)
 	}
-	up := ps.UserProfile
+	up := newPersist.UserProfile
 	if up.LoginName == "" {
-		up.LoginName = ps.LoginName
+		// Backwards compatibility with old prefs files.
+		up.LoginName = newPersist.LoginName
+	} else if newPersist.LoginName != "" && newPersist.LoginName != up.LoginName {
```
mihaip, line 152: `newPersist.LoginName != ""` should always be true given the early return at line 145.

ipn/ipnlocal/profiles.go
```diff
@@ -124,37 +141,45 @@ func init() {
...
+		existing = append(existing, pm.findProfilesByNodeID(newPersist.NodeID)...)
+		if len(existing) == 0 {
 			cp.ID, cp.Key = newUnusedID(pm.knownProfiles)
-			cp.Name = ps.LoginName
+			cp.Name = up.LoginName
+			cp.NodeID = newPersist.NodeID
+		} else {
+			// Only one profile per user/nodeID should exist.
+			for _, p := range existing[1:] {
+				pm.DeleteProfile(p.ID)
```
mihaip, line 172: Intentionally ignoring any deletion errors?

ipn/ipnlocal/profiles.go
```diff
@@ -124,37 +141,45 @@ func init() {
...
+			}
+			cp = existing[0]
```
mihaip, line 174: Should this still copy the NodeID and UserID (since only one may be populated)?

mihaip (approved):

ipn/ipnlocal/profiles.go
maisem: Done

ipn/ipnlocal/profiles.go
maisem: yeah, added a comment.

ipn/ipnlocal/profiles.go
maisem: Done.

PR #6370, opened 2022-11-17
cmd/tailscale, util/quarantine: set quarantine flags on files from Taildrop

andrew-d:
This sets the "com.apple.quarantine" flag on macOS, and the "Zone.Identifier" alternate data stream on Windows.

Change-Id: If14f805467b0e2963067937d7f34e08ba1d1fa85


andrew-d:
I don't have a Windows machine conveniently here to test this, but on my Mac:

Screenshot 2022-11-17 at 12 27 43 PM Screenshot 2022-11-17 at 12 27 55 PM

util/quarantine/quarantine_darwin.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		return err
+	}
+
+	now := time.Now()
+
+	// We uppercase the UUID to match what other applications on macOS do
+	id := strings.ToUpper(uuid.New().String())
+
+	attrData := fmt.Sprintf("%s;%x;%s;%s",
```
mihaip, line 28: Pointer to documentation of this format? https://eclecticlight.co/2020/10/29/quarantine-and-the-quarantine-flag/ is the best that I've found, but there may be others.

mihaip (commented):
Note that this doesn't appear to affect the behavior of Taildrop when using the macOS GUI client. That does the Taildrop file moving around at https://github.com/tailscale/corp/blob/c4e919ae008994eb9e38423d3d79b2151758d7af/xcode/IPNExtension/IPNPacketTunnelProvider.swift#L143

I'm guessing Windows is similar, it does its own file copying in https://github.com/tailscale/corp/blob/c4e919ae008994eb9e38423d3d79b2151758d7af/win/ui/files.go#L39

util/quarantine/quarantine_darwin.go
andrew-d: Good call; added

andrew-d:
> I'm guessing Windows is similar, it does its own file copying in https://github.com/tailscale/corp/blob/c4e919ae008994eb9e38423d3d79b2151758d7af/win/ui/files.go#L39

Yep, once this is merged I was going to follow up with a PR to use the same `quarantine` package there.

andrew-d:
> Note that this doesn't appear to affect the behavior of Taildrop when using the macOS GUI client. That does the Taildrop file moving around at https://github.com/tailscale/corp/blob/c4e919ae008994eb9e38423d3d79b2151758d7af/xcode/IPNExtension/IPNPacketTunnelProvider.swift#L143

I don't know if we want to do this, but there's apparently a flag in `Info.plist` that enables this for all files by default:
https://developer.apple.com/documentation/bundleresources/information_property_list/lsfilequarantineenabled

This is used by Chrome:
https://chromium.googlesource.com/chromium/chromium/+/refs/heads/main/content/browser/file_metadata_mac.mm#95

mihaip:
> I don't know if we want to do this, but there's apparently a flag in Info.plist that enables this for all files by default

As far as I can tell, even without that Info.plist key we get quarantine attributes added, presumably because we're a sandboxed app (see https://twitter.com/Schwieb/status/1161085763272024064 and other discussion).

util/quarantine/quarantine_darwin.go
```diff
@@ -0,0 +1,53 @@
...
+	// We uppercase the UUID to match what other applications on macOS do
+	id := strings.ToUpper(uuid.New().String())
+
+	// This format is under-documented, but the following links contain a
+	// reasonably comprehensive overview:
+	//    https://eclecticlight.co/2020/10/29/quarantine-and-the-quarantine-flag/
+	//    https://nixhacker.com/security-protection-in-macos-1/
+	//    https://ilostmynotes.blogspot.com/2012/06/gatekeeper-xprotect-and-quarantine.html
+	attrData := fmt.Sprintf("%s;%x;%s;%s",
+		"0002",      // quarantine value; kLSQuarantineTypeOtherDownload
```
mihaip, line 34: Per https://nixhacker.com/security-protection-in-macos-1/ `kLSQuarantineTypeOtherDownload` is 1 not 2. I also see a leading 8 most of the time, which I think implies that this is a bitfield that has other values too. 

https://github.com/keybase/client/blob/457ebc8fc401c44dccf2ce684267b802fe61ea74/go/kbfs/libfuse/quarantine_darwin.go#L26-L28 has `_kLSQuarantineTypeOtherDownload = "0081;"`, which would agree with that.

Though if I use the AppKit APIs (`NSURL`'s `setResourceValue(..., forKey: kCFURLQuarantinePropertiesKey as URLResourceKey)`) with `kLSQuarantineTypeKey` set to `kLSQuarantineTypeOtherDownload` I end up with `0083`, so 🤷

mihaip (approved):
LGTM for Mac changes.

dblohm7 (approved):
I just looked at the Windows bits, but that part LGTM.

util/quarantine/quarantine_darwin.go
andrew-d: Whoops, off-by-one 🤦‍♂️  Switched to `1` and verified that this still works:
```
$ xattr -p com.apple.quarantine ./quarantine-test
0001;63766eee;Tailscale;b3e2ed8b-5e8d-4774-ba7e-9e6ff89adb06
```

Screenshot 2022-11-17 at 2 56 35 PM

PR #6371, opened 2022-11-17
tsweb: avoid dashes in Prometheus metric names

knyar:
Ideally we should strip other invalid characters too ([reference](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels)), but that would call for a regexp replacement which increases the number of allocations and makes `TestVarzHandlerSorting` fail.

### DIFF
--- a/tsweb/tsweb.go
+++ b/tsweb/tsweb.go
@@ -455,18 +455,30 @@ func WritePrometheusExpvar(w io.Writer, kv expvar.KeyValue) {
 	writePromExpVar(w, "", kv)
 }
 
-func writePromExpVar(w io.Writer, prefix string, kv expvar.KeyValue) {
-	key := kv.Key
+type prometheusMetricDetails struct {
+	Name  string
+	Type  string
+	Label string
+}
+
+var prometheusMetricCache sync.Map // string => *prometheusMetricDetails
+
+func prometheusMetric(prefix string, key string) (string, string, string) {
+	cachekey := prefix + key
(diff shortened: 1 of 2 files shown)


danderson (approved):

knyar:
After talking to @bradfitz, also added a `sync.Map` cache for this.

Soypete (approved):

tsweb/tsweb.go
```diff
@@ -455,26 +455,49 @@ func WritePrometheusExpvar(w io.Writer, kv expvar.KeyValue) {
...
+	case strings.HasPrefix(key, gaugePrefix):
 		typ = "gauge"
-		key = strings.TrimPrefix(kv.Key, gaugePrefix)
+		key = strings.TrimPrefix(key, gaugePrefix)
 
-	case strings.HasPrefix(kv.Key, counterPrefix):
+	case strings.HasPrefix(key, counterPrefix):
 		typ = "counter"
-		key = strings.TrimPrefix(kv.Key, counterPrefix)
+		key = strings.TrimPrefix(key, counterPrefix)
```
Soypete, line 481: I am not quite sure how our prefixing works. It seems to me that we hang if there is not a "gauge" or "counter" prefix, is that correct? does this mean we require all prom metrics to have this prefix?

knyar: I think the prefixes are optional and there is some logic that attempts to determine metric type based on expvar type, but it's definitely safer to have a prefix.

Longer term we should probably just switch to an open source Prometheus client instead of exposing metrics ourselves via expvar.

PR #6382, opened 2022-11-18
cmd/tailscale/cli: add ability to set short names for profiles

maisem:
This adds a `--nickname` flag to `tailscale login|set`.
    
Updates #713
    
### DIFF
--- a/cmd/tailscale/cli/cli_test.go
+++ b/cmd/tailscale/cli/cli_test.go
@@ -480,6 +480,19 @@ func TestCheckForAccidentalSettingReverts(t *testing.T) {
 			distro: "", // not Synology
 			want:   accidentalUpPrefix + " --hostname=foo --accept-routes",
 		},
+		{
+			name:  "profile_name_ignored_in_up",
+			flags: []string{"--hostname=foo"},
+			curPrefs: &ipn.Prefs{
+				ControlURL:       "https://login.tailscale.com",
+				CorpDNS:          true,
+				AllowSingleHosts: true,
+				NetfilterMode:    preftype.NetfilterOn,
+				ProfileName:      "foo",
+			},
+			goos: "linux",
+			want: "",
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/cmd/tailscale/cli/set.go
+++ b/cmd/tailscale/cli/set.go
@@ -43,11 +43,13 @@ type setArgsT struct {
 	advertiseDefaultRoute  bool
 	opUser                 string
 	acceptedRisks          string
+	profileName            string
 }
 
 func newSetFlagSet(goos string, setArgs *setArgsT) *flag.FlagSet {
 	setf := newFlagSet("set")
 
+	setf.StringVar(&setArgs.profileName, "nickname", "", "nickname for the login profile")
 	setf.BoolVar(&setArgs.acceptRoutes, "accept-routes", false, "accept routes advertised by other Tailscale nodes")
 	setf.BoolVar(&setArgs.acceptDNS, "accept-dns", false, "accept DNS configuration from the admin panel")
 	setf.StringVar(&setArgs.exitNodeIP, "exit-node", "", "Tailscale exit node (IP or base name) for internet traffic, or empty string to not use an exit node")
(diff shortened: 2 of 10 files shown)


maisem:
cc @mayakacz

cmd/tailscale/cli/set.go
```diff
@@ -43,11 +43,13 @@ type setArgsT struct {
 	advertiseDefaultRoute  bool
 	opUser                 string
 	acceptedRisks          string
+	profileName            string
 }
 
 func newSetFlagSet(goos string, setArgs *setArgsT) *flag.FlagSet {
 	setf := newFlagSet("set")
 
+	setf.StringVar(&setArgs.profileName, "name", "", "short name for the login profile")
```
mihaip, line 52: Feels like you'd want this to be `profile-name` or something more specific. Otherwise it gets confusing given that `hostname` is also settable (at least with `up`).

ipn/prefs.go
```diff
@@ -190,6 +190,11 @@ type Prefs struct {
 	// operate tailscaled without being root or using sudo.
 	OperatorUser string `json:",omitempty"`
 
+	// ProfileName is the desired name of the profile. If empty, the the Users
```
mihaip, line 193: Is "Users" intentionally capitalized?

mihaip:
Test failures look legit.

cmd/tailscale/cli/set.go
maisem: Done.
However, I also removed `--profile-name` from `tailscale up`. I think we are at the point where we should not add more flags to `tailscale up` and start steering folks to `tailscale login` and `tailscale set`.

ipn/prefs.go
maisem: Nope, fixed.

maisem:
> Test failures look legit.

Sadly these tests only run on linux so I occasionally forget to run the, when developing on my laptop.

cmd/tailscale/cli/cli_test.go
```diff
@@ -480,6 +480,19 @@ func TestCheckForAccidentalSettingReverts(t *testing.T) {
 			distro: "", // not Synology
 			want:   accidentalUpPrefix + " --hostname=foo --accept-routes",
 		},
+		{
+			name:  "profile_name_ignored_in_up",
+			flags: []string{"--hostname=foo"},
```
mihaip, line 485: Should this be `--profile-name`?

ipn/prefs.go
```diff
@@ -190,6 +190,11 @@ type Prefs struct {
 	// operate tailscaled without being root or using sudo.
 	OperatorUser string `json:",omitempty"`
 
+	// ProfileName is the desired name of the profile. If empty, then the users
```
mihaip, line 193: "user's"?

ipn/ipnlocal/profiles.go
```diff
@@ -174,9 +174,13 @@ func (pm *profileManager) SetPrefs(prefsIn ipn.PrefsView) error {
 		}
 		cp.LocalUserID = pm.currentUserID
 	}
+	if prefs.ProfileName() != "" {
```
mihaip, line 177: Do we want to enforce uniqueness for names?

mihaip (commented):
Should probably update the PR and commit message with the new flag name too.

cmd/tailscale/cli/set.go
mayakacz: Given it's `hostname` (one word) vs `profile-name` (two words), suggest either `profilename` or something like `nickname`.

maisem: well `hostname` is a single word https://en.wikipedia.org/wiki/Hostname 🤪 

nickname sgtm

cmd/tailscale/cli/cli_test.go
maisem: This checks that even though `--profile-name` arg is not defined AND we already have a `ProfileName` set it doesn't result in the accidental revert message.

ipn/ipnlocal/profiles.go
maisem: Good point.

maisem: Done. PTAL

mayakacz (approved):

mihaip (approved):

Issue #3569, opened 2021-12-15
client should reject both being and using an exit node at the same time


bradfitz:
I don't think we've ever considered whether using an exit node & running (advertising) an exit node at the same time should work.

Maybe it works today? Kinda? On some platforms?

But I don't think it works consistently on all platforms today if it even works on any.

And what should it do? Should the exit node that's running forward out of the exit node that you're using?

What if there's a loop?

etc.

Decide.

mayakacz:
From Slack conversation: shouldn't be able to be both exit node and use exit node. Whichever one you selected last is what it should be doing, and automatically disconnect/disable the other option.
We should probably warn you of this/ make this clear, esp. if you are being used as an exit node.

mayakacz:
> Maybe it works today? Kinda? On some platforms?

If it does, we should be clear where that's true. I don't think we've tested all cases.

In at least some cases, it doesn't work. Hit this today with macOS and it caused unexpected behaviour. Was advertising an exit node, then selected an exit node, and traffic did not work. Did not realize I was still advertising an exit node from something I did earlier.

Proposal: similar to how iOS hotspot/wifi interaction used to(?) work:
* If I am advertising an exit node, and try to use an exit node, then give me a warning that this will stop advertising an exit node, before doing so.
* If I am connected to an exit node, and try to advertise an exit node, same.
* If I am neither using an exit node, or advertising an exit node, then using these features works as they currently do

I think it's preferable to give a warning (before applying change), rather than say, make options not selectable, as I don't necessarily know why those are not available. Selecting the option will still tell me why and let me decide.

rweirtbg:
This does work on Linux (Ubuntu 22.04/20.04) and we desperately need this capability for our IOT application for both our IOT appliance to route its internet traffic through tailscale to bypass the customer LAN and improve reliability and the ability for the appliance to access our customer's LAN to reach other devices.

bradfitz:
@rweirtbg, I'm not exactly following your requirements from that description. Why can't you have the source clients use the ultimate exit node? What do you get from hopping via an intermediate exit node?

From #5710 we're thinking of removing this because there's a bunch of undefined behavior that varies per platform and we don't want to guarantee any behavior that "works" today when we didn't try to make anything work. We don't even know what it's supposed to do vs what it actually does.

Like: in your configuration, which LAN(s) should be accessible? How should DNS route?

rweirtbg:
[...]

PR #6383, opened 2022-11-18
ipn/ipnlocal: disallow exit nodes from using exit nodes

saljmz:
Nodes which have both -advertise-exit-node and -exit-node in prefs should continue have them until the next invocation of `tailscale up`.

Updates #3569.


DentonGentry:
We are approaching the cutoff for the 1.34.0 release. If this is going to go in — and it solves a real issue we have, would be nice if it does go in — it should be sent for review soon.

bradfitz (approved):
Looks plausible if it works!

bradfitz:
cc @maisem too

maisem (approved):

PR #6386, opened 2022-11-18
ipn/ipnserver: set a CSP for ServeHTMLStatus, refactor host validation

twitchyliquid64:


ipn/ipnserver/server.go
```diff
@@ -1029,13 +1029,16 @@ func (s *Server) localhostHandler(ci connIdentity) http.Handler {
 // Windows and via $DEBUG_LISTENER/debug/ipn when tailscaled's --debug flag
 // is used to run a debug server.
 func (s *Server) ServeHTMLStatus(w http.ResponseWriter, r *http.Request) {
-	// As this is only meant for debug, verify there's no DNS name being used to
-	// access this.
-	if strings.IndexFunc(r.Host, unicode.IsLetter) != -1 {
+	// As this is only meant for debug, verify we are only serving to a
+	// localhost or equivalent (i.e. 127.0.0.1) origin.
+	if r.Host != "localhost" && strings.IndexFunc(r.Host, unicode.IsLetter) != -1 {
```
bradfitz, line 1034: it won't be "localhost". It will be "localhost:nnnnnn".  You need a SplitHostPort to check the hostname. But at that point it's worth pulling this out into a func.

ipn/ipnserver/server.go
```diff
@@ -1029,13 +1029,16 @@ func (s *Server) localhostHandler(ci connIdentity) http.Handler {
...
 		http.Error(w, "invalid host", http.StatusForbidden)
 		return
 	}
 
+	w.Header().Set("Content-Security-Policy", `default-src 'none'; frame-ancestors 'none'; script-src 'none'; script-src-elem 'none'; script-src-attr 'none'`)
```
bradfitz, line 1039: time to put all this stuff in tsweb or something?

starting to get repetitive

ipn/ipnserver/server.go
twitchyliquid64: Maybe we can bring `validHost()` up a layer to ipnserver? Otherwise we could potentially get away with `strings.HasPrefix(r.Host, "localhost:")`

ipn/ipnserver/server.go
twitchyliquid64: Maybe as an option? We only want this set on stuff that must never be i-framed.

bradfitz: oh I meant as helper funcs/options, yeah, not by default

ipn/ipnserver/server.go
bradfitz: strings.HasPrefix on localhost: works for now

twitchyliquid64:
NB: I refactored the host check a layer up. `tsnet` will not hit this check, but localAPI listener is in-process so thats fine (it isn't in scope for being externally driven).

ipn/ipnserver/server.go
twitchyliquid64: I ended up bringing `validHost()` a layer up.

ipn/ipnserver/server.go
twitchyliquid64: I'll leave this for another PR as discussed on slack

bradfitz (commented):
I don't think this is right. The two things don't have the same protection rules because the ServeHTMLStatus can listen anywhere (on any port) by setting the debug listener to tailscaled, which also works on windows.

twitchyliquid64:
Ok - I've reverted back to the earlier version where the host check is as it was + we use `strings.HasPrefix` to check the host.

ipn/ipnserver/server.go
twitchyliquid64: Update: Doing `string.HasPrefix` now as we discussed.

bradfitz (approved):

Issue #5309, opened 2022-08-05
use Go 1.20's strings.CutPrefix, CutSuffix


bradfitz:
(In 6 months, when Go 1.20 is out)

bradfitz:
Upstream is https://github.com/golang/go/issues/42537

PR #6387, opened 2022-11-18
all: use strs.CutPrefix and strs.CutSuffix more

mihaip:
Updates places where we use HasPrefix + TrimPrefix to use the combined function.

Updates #5309


derp/derphttp/derphttp_client.go
```diff
@@ -1028,9 +1028,9 @@ var ErrClientClosed = errors.New("derphttp.Client closed")
 
 func parseMetaCert(certs []*x509.Certificate) (serverPub key.NodePublic, serverProtoVersion int) {
 	for _, cert := range certs {
-		if cn := cert.Subject.CommonName; strings.HasPrefix(cn, "derpkey") {
+		if cn, ok := strs.CutPrefix(cert.Subject.CommonName, "derpkey"); ok {
```
bradfitz, line 1031: not sure "cn" as a variable name is so accurate anymore, but also fine I suppose. Maybe "pubHex"?

bradfitz (approved):

derp/derphttp/derphttp_client.go
mihaip: Went with `pubHex`

PR #6390, opened 2022-11-18
ipn/ipnlocal: strip NetworkLockKey from Prefs

maisem:

### DIFF
--- a/ipn/ipnlocal/state_test.go
+++ b/ipn/ipnlocal/state_test.go
@@ -973,19 +973,20 @@ func TestEditPrefsHasNoKeys(t *testing.T) {
 		t.Errorf("Hostname = %q; want foo", p.Hostname())
 	}
 
-	// Test that we can't see the PrivateNodeKey.
 	if !p.Persist().PrivateNodeKey.IsZero() {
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/state_test.go
```diff
@@ -973,20 +973,21 @@ func TestEditPrefsHasNoKeys(t *testing.T) {
...
 	}
 
-	// Test that we can't see the PrivateNodeKey.
 	if !p.Persist().LegacyFrontendPrivateMachineKey.IsZero() {
 		t.Errorf("LegacyFrontendPrivateMachineKey = %v; want zero", p.Persist().LegacyFrontendPrivateMachineKey)
 	}
+
+	if !p.Persist().NetworkLockKey.IsZero() {
+		t.Errorf("NetworkLockKey= %v; want zero", p.Persist().NetworkLockKey)
+	}
```
andrew-d, line 990: I wonder if we could use some sort of `reflect`-based approach to find all fields with the substring `Key`?

andrew-d (approved):

ipn/ipnlocal/state_test.go
maisem: sgtm for a follow up

PR #6392, opened 2022-11-18
control/controlclient, wgengine/magicsock: ignore pre-disco (pre-0.100) peers

bradfitz:
There aren't any in the wild, other than one we ran on purpose to keep us honest, but we can bump that one forward to 0.100.


DentonGentry:
🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

Oh this will be SOOO nice.

phirework (commented):
We just need to remove `TestNoDiscoKey` from the test suite.

bradfitz:
> We just need to remove TestNoDiscoKey from the test suite.

Done.

bradfitz:
PTAL. Ended up going a different route, as an integration test revealed the old way was broken a bit.

phirework (approved):

PR #6393, opened 2022-11-18
words: add some cloven-hoofed ruminants

lrewega:
The Bovini tribe is sorely underrepresented.

### DIFF
--- a/words/tails.txt
+++ b/words/tails.txt
@@ -333,3 +333,10 @@ snares
 rockhopper
 royal
 macaroni
+bison
+buffalo
+gaur
+saola
+anoa
+banteng
+beefalo


DentonGentry:
The DCO error is because we require commits to have a Signed-Off-By line, indicating that the commit is suitable for open source release.

To add one:
```
git commit --amend --signoff
git push -f
```

lrewega:
> The DCO error is because we require commits to have a Signed-Off-By line

The error is really because I was being cheeky and using the GitHub interface rather than doing this the right way 🙃 

Signed off now.

words/tails.txt
```diff
@@ -272,3 +272,12 @@ qilin
...
 mercat
+bison
+buffalo
+gaur
+saola
+anoa
+banteng
+beefalo
+yak
+cow
```
catzkorn, line 283: These last two are duplicates, we have a test that checks for duplicates in the lists. Can you run the test and remove any duplicates?

lrewega: Wow, apologies -- GitHub search failed me here.

catzkorn:
(psst, they are still there!)

catzkorn (approved):
🐄

PR #6398, opened 2022-11-19
words: add a list of things you should yahoo!

raggi:


catzkorn:
_Other search engines are available_

raggi:
> Other search engines are available

quite right too, corrected!

catzkorn (approved):
For this, I would use duck duck go 😏🦆

PR #6401, opened 2022-11-19
cmd/tailscale/cli, ipn: move serve CLI funcs on to ServeConfig

shayne:


shayne:
Doing this before #6400

maisem (approved):
Nice!

PR #6403, opened 2022-11-19
words: where do penguins go when they lose their tail?

catzkorn:
To the retail store.


words/tails.txt
```diff
@@ -317,3 +317,19 @@ soay
 xalda
 ewe
 lamb
+great
+emperor
+king
+adelie
+chinstrap
+gentoo
```
andrew-d, line 325: Somehow I've made it this many years before discovering that "gentoo" isn't just a Linux distro but also a kind of penguin 🤦‍♂️

words/scales.txt
```diff
@@ -179,3 +179,18 @@ beaver
 hops
 salak
 mercat
+great
```
andrew-d, line 182: I don't think penguins have scales?

catzkorn: king-penguin-feet-antarctica-philippe-tulula-and-julie-alexander

catzkorn: > Do Penguins Have Scales?

Every part of a penguin’s body, from its head to its feet, is coated in scales and feathers.

andrew-d: 🤯

andrew-d (approved):

words/scales.txt
andrew-d: I'm also very happy that this means that `great-great` is now a valid tailnet name 🤣

catzkorn: great-gentoo, linux fans are gonna go NUTS

words/tails.txt
shayne: EverythingIsConnectedNeilDegrasseTysonGIF

PR #6408, opened 2022-11-19
ipn/ipnlocal: send Content-Security-Policy, etc to peerapi browser requests

bradfitz:
Updates tailscale/corp#7948


andrew-d (approved):
This seems like a reasonable set of heuristics to me! Especially since we have it behind the `validatePeerAPIRequest` call

PR #6412, opened 2022-11-20
words: good vibes only

catzkorn:
After consultation with @twitchyliquid64 , it has been agreed that a vibe, or vibes, can be felt in different quantifiable measures. That makes a vibe, or vibes, a scale thus it must be immortalized.


twitchyliquid64 (approved):
Vibes ✨

PR #6414, opened 2022-11-20
tailcfg: add TailscaleFunnelEnabled to HostInfo

shayne:
Adding this convenience for tailscale/corp#8015 and to avoid leaking WireIngress.


maisem:
Sure, but I don't understand what you mean by avoid leaking

maisem (approved):

shayne:
Leaking in the sense of keeping corp from referencing WireIngress if, in the future, semantics around this check changes.

Issue #6405, opened 2022-11-19
ipn/ipnlocal: allow serve of empty Text() handler


shayne:
In writing up the CLI docs KB I thought of the use case `tailscale serve / text ""` which would be useful for creating a clean blank page for the root of your serve/funnel.

If you don't specify the mount point `/` you are served an `http.NotFound(...)`, if you do the mount as above...

Currently we check against `!= ""`
https://github.com/tailscale/tailscale/blob/063eeefdca01ac110a8ae0e76e90c70e6eb6d1fe/ipn/ipnlocal/serve.go#L398-L402

Serving a 500 with the text "empty handler"
https://github.com/tailscale/tailscale/blob/063eeefdca01ac110a8ae0e76e90c70e6eb6d1fe/ipn/ipnlocal/serve.go#L428

This seem reasonable?

cc/ @bradfitz

bradfitz:
I'm not sure this is worth fixing. Or maybe we map `""` to `" "` (a single space) in the CLI, or U+200B ZERO WIDTH SPACE.

But I don't think we need to do anything in the backend or change types.

shayne:
I like the idea of mapping "" to some sentinel like U+200B. We wouldn't have to change the backend but it'd work as expected for the user.

PR #6415, opened 2022-11-20
cmd/tailscale/cli: allow empty text ("") serve

shayne:
Current behavior is broken. `tailscale serve text / ""` returns no error
and shows up in `tailscale serve status` but requests return a 500
"empty handler".

The commit changes the handler field `Text` to type of `*string`.

Closes https://github.com/tailscale/tailscale/issues/6405

### DIFF
--- a/cmd/tailscale/cli/serve.go
+++ b/cmd/tailscale/cli/serve.go
@@ -285,7 +285,7 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
 		}
 		h.Proxy = t
 	case "text":
-		h.Text = args[2]
+		h.Text = ptrTo(args[2])
 	default:
 		fmt.Fprintf(os.Stderr, "error: unknown serve type %q\n\n", args[1])
 		return flag.ErrHelp
@@ -547,8 +547,8 @@ func printWebStatusTree(sc *ipn.ServeConfig, hp ipn.HostPort) {
 			return "path", h.Path
 		case h.Proxy != "":
 			return "proxy", h.Proxy
-		case h.Text != "":
-			return "text", "\"" + elipticallyTruncate(h.Text, 20) + "\""
+		case h.Text != nil:
+			return "text", "\"" + elipticallyTruncate(*h.Text, 20) + "\""
 		}
 		return "", ""
 	}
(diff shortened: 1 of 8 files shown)


cmd/tailscale/cli/serve.go
```diff
@@ -548,7 +552,13 @@ func printWebStatusTree(sc *ipn.ServeConfig, hp ipn.HostPort) {
 		case h.Proxy != "":
 			return "proxy", h.Proxy
 		case h.Text != "":
```
maisem, line 554: We could make the Text field be `*string` instead

shayne: I'm open to that. I was trying to avoid changing the entire backend to support a single empty case. This approach seems harmless if imprecise.

maisem: I'm thinking about cases where users have automation that sets this up using the localapi.

The changes shouldn't be too much as json.Unmarshal should handle this correctly I think if we don't add omitempty.

shayne: OK I'll rework it.

shayne: PTAL

cmd/tailscale/cli/serve_test.go
```diff
@@ -49,6 +49,10 @@ func TestCleanMountPoint(t *testing.T) {
 	}
 }
 
+// strp returns a pointer to the string s.
+// Used for tests that need a pointer to a string.
+func strp(s string) *string { return &s }
```
maisem, line 54: We have ptrTo defined in a few places. I think I added it in `set_test.go` so you should be able to use it instead of this

cmd/tailscale/cli/serve.go
```diff
@@ -285,7 +285,7 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
 		}
 		h.Proxy = t
 	case "text":
-		h.Text = args[2]
```
maisem, line 288: Could also move ptrTo the package and use it here. ¯\\\_(ツ)\_/¯

maisem (approved):

cmd/tailscale/cli/serve_test.go
shayne: Moved it to `cli/cli.go`

cmd/tailscale/cli/serve.go
shayne: Done.

bradfitz (changes requested):
Let's not. This is too complicated for what it's worth. Let's just ban setting empty string in CLI. Caller can pass a space.

shayne:
OK.

Closing in favor of #6416

PR #6422, opened 2022-11-21
cmd/tailscale/cli: use mock impl of LocalClient for serve cmd

### LINKED ISSUE #6372: cmd/tailscale/cli: fake local client for serve tests
Follow up task from #6304 

Better to have a fake localClient instead. Maybe add type localServeClient interface{} with the required functions to facilitate.

shayne:
Create an interface and mock implementation of `tailscale.LocalClient` for serve command tests.

Updates #6304
Closes #6372

### DIFF
--- a/cmd/tailscale/cli/serve.go
+++ b/cmd/tailscale/cli/serve.go
@@ -30,9 +30,9 @@ import (
 	"tailscale.com/version"
 )
 
-var serveCmd = newServeCommand(&serveEnv{})
+var serveCmd = newServeCommand(&serveEnv{lc: &localClient})
 
-// newServeCommand returns a new "serve" subcommand using e as its environmment.
+// newServeCommand returns a new "serve" subcommand using e as its environment.
 func newServeCommand(e *serveEnv) *ffcli.Command {
 	return &ffcli.Command{
 		Name:      "serve",
@@ -127,8 +127,21 @@ func (e *serveEnv) newFlags(name string, setup func(fs *flag.FlagSet)) *flag.Fla
 	return fs
 }
 
+// localServeClient is an interface conforming to the subset of
+// tailscale.LocalClient. It includes only the methods used by the
+// serve command.
+//
+// The purpose of this interface is to allow tests to provide a mock.
+type localServeClient interface {
+	Status(context.Context) (*ipnstate.Status, error)
+	GetServeConfig(context.Context) (*ipn.ServeConfig, error)
+	SetServeConfig(context.Context, *ipn.ServeConfig) error
+}
+
 // serveEnv is the environment the serve command runs within. All I/O should be
 // done via serveEnv methods so that it can be faked out for tests.
+// Calls to localClient should be done via the lc field, which is an interface
+// that can be faked out for tests.
 //
 // It also contains the flags, as registered with newServeCommand.
 type serveEnv struct {
@@ -138,12 +151,11 @@ type serveEnv struct {
 	remove       bool // remove a serve config
 	json         bool // output JSON (status only for now)
 
+	lc localServeClient // localClient interface, specific to serve
+
 	// optional stuff for tests:
-	testFlagOut              io.Writer
-	testGetServeConfig       func(context.Context) (*ipn.ServeConfig, error)
-	testSetServeConfig       func(context.Context, *ipn.ServeConfig) error
-	testGetLocalClientStatus func(context.Context) (*ipnstate.Status, error)
-	testStdout               io.Writer
+	testFlagOut io.Writer
+	testStdout  io.Writer
 }
 
 // getSelfDNSName returns the DNS name of the current node.
@@ -157,15 +169,12 @@ func (e *serveEnv) getSelfDNSName(ctx context.Context) (string, error) {
 	return strings.TrimSuffix(st.Self.DNSName, "."), nil
 }
 
-// getLocalClientStatus calls LocalClient.Status, checks if
-// Status is ready.
+// getLocalClientStatus returns the Status of the local client.
 // Returns error if unable to reach tailscaled or if self node is nil.
+//
 // Exits if status is not running or starting.
 func (e *serveEnv) getLocalClientStatus(ctx context.Context) (*ipnstate.Status, error) {
-	if e.testGetLocalClientStatus != nil {
-		return e.testGetLocalClientStatus(ctx)
-	}
-	st, err := localClient.Status(ctx)
+	st, err := e.lc.Status(ctx)
 	if err != nil {
 		return nil, fixTailscaledConnectError(err)
 	}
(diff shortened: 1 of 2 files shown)


bradfitz (commented):
I'll have some feedback later tonight.

cmd/tailscale/cli/serve_test.go
```diff
@@ -631,23 +617,58 @@ func TestServeConfigMutations(t *testing.T) {
...
 
+// fakeLocalServeClient is a fake implementation of LocalClient
+// used by the serve command tests. It's not a full implementation,
+// just enough to test the serve command.
+//
+// The fake client is stateful, and is used to test manipulating
+// ServeConfig state.
+type fakeLocalServeClient struct {
+	current  *ipn.ServeConfig
+	newState *ipn.ServeConfig
```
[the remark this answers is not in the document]
maisem: yeah, why track the values differently?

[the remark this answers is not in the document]
shayne: Fine to simplify and change to a single field. But to answer your questions:

Two fields because 1) the way it was initially written; 2) the tests check their `want` against config passed to `SetServeConfig(...)` during a given step.

Example:
https://github.com/tailscale/tailscale/blob/4073613ffe18a7000825ae5501e083ac53fe9c93/cmd/tailscale/cli/serve_test.go#L69-L76

If we switch to one field then the step on L73 want would be an unchanged `ServeConfig` from the previous step, not nil:
```
{
    "AllowFunnel": {
        "foo.test.ts.net:443": true
    }
}
```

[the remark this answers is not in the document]
bradfitz: Yeah, but the old code was not trying to implement any interface. It was defining exactly what it needed for tests: a `reset` bool to clear the state if needed for that series of stateful steps, `newState` to capture what was saved (and whether a save even occured!), and `current`, to keep track of the accumulated state between steps when there isn't a `reset`.

I think you want to keep the `var newState *ipn.ServeConfig` line above unchanged so you can tell whether a save happened.

cmd/tailscale/cli/serve_test.go
```diff
@@ -631,23 +617,58 @@ func TestServeConfigMutations(t *testing.T) {
...
 		}
 	}
 }
 
+// fakeLocalServeClient is a fake implementation of LocalClient
+// used by the serve command tests. It's not a full implementation,
+// just enough to test the serve command.
+//
+// The fake client is stateful, and is used to test manipulating
+// ServeConfig state.
```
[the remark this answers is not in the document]
shayne: not sure about `fakeStatus.Clone()`, `ipnstate.Status` isn't currently part of the generated `ipn_clone.go`. The real implementation is a `json.Unmarshal()` of the localapi response body.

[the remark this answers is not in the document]
shayne: PTAL

cmd/tailscale/cli/serve_test.go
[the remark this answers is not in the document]
shayne: PTAL

cmd/tailscale/cli/serve_test.go
```diff
@@ -70,17 +71,17 @@ func TestServeConfigMutations(t *testing.T) {
 		command: cmd("funnel on"),
 		want:    &ipn.ServeConfig{AllowFunnel: map[ipn.HostPort]bool{"foo.test.ts.net:443": true}},
 	})
-	add(step{
+	add(step{ // want no change
 		command: cmd("funnel on"),
-		want:    nil, // nothing to save
+		want:    &ipn.ServeConfig{AllowFunnel: map[ipn.HostPort]bool{"foo.test.ts.net:443": true}},
```
bradfitz, line 76: This is the behavior change I was warning about. Line 58 above says "non-nil means we want a save of this value". You even added "// want no change" on line 74 but you had to fill this in to make this test pass, when it should be nil to mean no save occurred.

You could make your `fakeLocalServeClient` Set method also increment a counter and then the test can check whether the counter was incremented. If not, then nil out the "got" value before comparing to want.

cmd/tailscale/cli/serve_test.go
```diff
@@ -631,23 +631,59 @@ func TestServeConfigMutations(t *testing.T) {
...
 }
 
+// fakeLocalServeClient is a fake implementation of LocalClient
+// used by the serve command tests. It's not a full implementation,
+// just enough to test the serve command.
+//
+// The fake client is stateful, and is used to test manipulating
+// ServeConfig state.
+type fakeLocalServeClient struct {
+	mu     sync.Mutex
```
bradfitz, line 655: why the mutex? If want to keep it, you should fix all the callers above that access the config field without holding it, for consistency.

But I think it's safe to just drop the mutex and document that this implementation can't be used concurrently.

cmd/tailscale/cli/serve_test.go
shayne: That works. I should've asked before changing it in the last push. I misunderstood the feedback.

cmd/tailscale/cli/serve_test.go
shayne: Done.

cmd/tailscale/cli/serve_test.go
shayne: Implemented count as suggested below.

cmd/tailscale/cli/serve.go
```diff
@@ -127,8 +127,21 @@ func (e *serveEnv) newFlags(name string, setup func(fs *flag.FlagSet)) *flag.Fla
...
 
+// localServeClient is an interface conforming to the subset of
+// tailscale.LocalClient. It includes only the methods used by the
+// serve command.
+//
+// The purpose of this interface is to allow tests to provide a mock.
+type localServeClient interface {
+	Status(ctx context.Context) (*ipnstate.Status, error)
+	GetServeConfig(ctx context.Context) (*ipn.ServeConfig, error)
+	SetServeConfig(ctx context.Context, config *ipn.ServeConfig) error
```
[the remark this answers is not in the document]
shayne: Done.

cmd/tailscale/cli/serve.go
```diff
@@ -30,9 +30,9 @@ import (
 	"tailscale.com/version"
 )
 
-var serveCmd = newServeCommand(&serveEnv{})
+var serveCmd = newServeCommand(&serveEnv{lc: &localClient})
```
bradfitz, line 33: why the `&`?

The variable is of type:

```
var localClient *tailscale.LocalClient
```

And those three methods are already defined on the LocalClient pointer.

cmd/tailscale/cli/serve_test.go
```diff
@@ -631,23 +617,62 @@ func TestServeConfigMutations(t *testing.T) {
...
+// just enough to test the serve command.
+//
+// The fake client is stateful, and is used to test manipulating
+// ServeConfig state. This implementation cannot be used concurrently.
+type fakeLocalServeClient struct {
+	config   *ipn.ServeConfig
+	setCount uint // counts calls to SetServeConfig
+}
+
+// fakeStatus is a fake implementation of ipn.State.
```
bradfitz, line 649: a fake ipnstate.Status value for tests

cmd/tailscale/cli/serve_test.go
```diff
@@ -631,23 +617,62 @@ func TestServeConfigMutations(t *testing.T) {
...
 
+// fakeLocalServeClient is a fake implementation of LocalClient
+// used by the serve command tests. It's not a full implementation,
+// just enough to test the serve command.
+//
+// The fake client is stateful, and is used to test manipulating
+// ServeConfig state. This implementation cannot be used concurrently.
+type fakeLocalServeClient struct {
+	config   *ipn.ServeConfig
+	setCount uint // counts calls to SetServeConfig
```
bradfitz, line 646: I'd just use int

bradfitz (approved):

cmd/tailscale/cli/serve.go
shayne: I see the variable type as the struct:
https://github.com/tailscale/tailscale/blob/6edf357b96b28ee1be659a70232c0135b2ffedfd/cmd/tailscale/cli/cli.go#L127

bradfitz: Ah I was looking in a different package.

PR #6425, opened 2022-11-21
all: use named pipes on windows

maisem:


ipn/ipnserver/server_windows.go
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build windows
```
bradfitz, line 5: Not needed. Filename does this.

ipn/ipnserver/server_windows.go
```diff
@@ -0,0 +1,64 @@
...
+func getNamedPipeClientProcessId(h windows.Handle) (pid uint32, err error) {
+	r1, _, err := procGetNamedPipeClientProcessId.Call(uintptr(h), uintptr(unsafe.Pointer(&pid)))
+	if r1 > 0 {
+		return pid, nil
+	}
+	return 0, err
+}
+
+// getConnIdentity extracts the identity information from the connection.
+// It expects the passed-in connection to be a localhost TCP connection over
```
bradfitz, line 34: Huh?

safesocket/pipe_windows.go
```diff
@@ -26,20 +26,46 @@ func setFlags(network, address string, c syscall.RawConn) error {
 	})
 }
 
-// TODO(apenwarr): use named pipes instead of sockets?
-//
-//	I tried to use winio.ListenPipe() here, but that code is a disaster,
-//	built on top of an API that's a disaster. So for now we'll hack it by
-//	just always using a TCP session on a fixed port on localhost. As a
-//	result, on Windows we ignore the vendor and name strings.
-//	NOTE(bradfitz): Jason did a new pipe package: https://go-review.googlesource.com/c/sys/+/299009
```
bradfitz, line 35: Hah

ipn/ipnserver/server_windows.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	pid, err := getNamedPipeClientProcessId(h.Handle())
+	if err != nil {
+		return ci, fmt.Errorf("getNamedPipeClientProcessId: %v", err)
+	}
+	ci.Pid = int(pid)
+	uid, err := pidowner.OwnerOfPID(ci.Pid)
+	if err != nil {
+		var hint string
+		if runtime.GOOS == "windows" {
```
andrew-d, line 52: Is this necessary now that this is in a `_windows.go` file?

safesocket/pipe_windows.go
```diff
@@ -26,20 +26,46 @@ func setFlags(network, address string, c syscall.RawConn) error {
...
+
+	if err = sd.SetDACL(acl, true, false); err != nil {
+		return nil, 0, err
+	}
+	cfg := namedpipe.ListenConfig{
+		SecurityDescriptor: sd,
+		InputBufferSize:    256 * 1024,
+		OutputBufferSize:   256 * 1024,
+	}
+	lc, err := cfg.Listen(path)
```
andrew-d, line 66: It looks like this sets `FILE_PIPE_REJECT_REMOTE_CLIENTS`, which is good–named pipes can be used from remote clients by default, which is a huge footgun 😱

dblohm7 (changes requested):
One reason why that `namedpipe` package was stalled getting into the std lib was because I had some serious concerns with its implementation. I'd rather we did not use it.

maisem:
> One reason why that `namedpipe` package was stalled getting into the std lib was because I had some serious concerns with its implementation.

Can you clarify what the concerns are?

ipn/ipnserver/server_windows.go
maisem: fixed

ipn/ipnserver/server_windows.go
maisem: done

ipn/ipnserver/server_windows.go
maisem: Done

ipn/ipnserver/server_notwindows.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"net"
+
+	"inet.af/peercred"
+)
+
+// getConnIdentity returns the localhost TCP connection's identity information
+// (pid, userid, user). If it's not Windows (for now), it returns a nil error
+// and a ConnIdentity with NotWindows set true. It's only an error if we expected
+// to be able to map it and couldn't.
```
raggi, line 18: I'd like to see these documented conditions simplified, they contain a large amount of boolean logic right now in the description.

Ideally we can simplify this down to explicitly: the result is usable, or the result is not usable. Everything else should be implementation detail.

ipn/ipnserver/server_notwindows.go
```diff
@@ -0,0 +1,25 @@
...
+func (s *Server) getConnIdentity(c net.Conn) (ci connIdentity, err error) {
+	ci = connIdentity{Conn: c, NotWindows: true}
+	ci.NotWindows = true
+	_, ci.IsUnixSock = c.(*net.UnixConn)
+	ci.Creds, _ = peercred.Get(c)
```
raggi, line 23: This could error due to selinux policies or similar, why not return an error in this cases? Field debugging without logging the cause may be extra difficult.

ipn/ipnserver/server_windows.go
```diff
@@ -0,0 +1,56 @@
...
+func getNamedPipeClientProcessId(h windows.Handle) (pid uint32, err error) {
+	r1, _, err := procGetNamedPipeClientProcessId.Call(uintptr(h), uintptr(unsafe.Pointer(&pid)))
+	if r1 > 0 {
+		return pid, nil
+	}
+	return 0, err
+}
+
+// getConnIdentity extracts the identity information from the connection.
+// It expects the passed-in connection to be the c to be created over a Named Pipe.
```
raggi, line 31: `If c is not backed by a named pipe, an error is returned.`

safesocket/pipe_windows.go
```diff
@@ -26,20 +26,46 @@ func setFlags(network, address string, c syscall.RawConn) error {
...
+	if err := windows.RtlDefaultNpAcl(&acl); err != nil {
+		return nil, 0, err
+	}
+	defer windows.LocalFree(windows.Handle(unsafe.Pointer(acl)))
+	sd, err := windows.NewSecurityDescriptor()
+	if err != nil {
+		return nil, 0, err
+	}
+
+	allUsers, err := windows.CreateWellKnownSid(windows.WinBuiltinUsersSid)
```
raggi, line 41: We might consider using `Network Configuration Operators` which is a standard group that may be a better fit.

ipn/ipnserver/server_notwindows.go
maisem: This is a copy paste from before. This is purely code movement. I'd be happy to simplify it

ipn/ipnserver/server_notwindows.go
maisem: Again, this is all code movement.

maisem: I do not want to modify this behavior without also testing it on all the *nix platforms.

ipn/ipnserver/server_windows.go
maisem: Done

ipn/ipnserver/server_notwindows.go
maisem: reworded. PTAL

dblohm7:
> @dblohm7 I've added another commit which uses the `go-winio` pkg. Does that seem better?

I'll need to look. I think the other package was based on this one, so I need to see whether it has the same issues.

bradfitz:
I think everybody wants this in 1.34.0 so let's figure this out in the next day or two.

Otherwise I think we'll just go with whatever Wireguard-go uses, on the assumption that it must not be too bad if it has a bunch of users.

paths/paths.go
```diff
@@ -23,7 +23,7 @@ var AppSharedDir syncs.AtomicValue[string]
 // or the empty string if there's no reasonable default.
 func DefaultTailscaledSocket() string {
 	if runtime.GOOS == "windows" {
-		return ""
+		return `\\.\pipe\tailscaled`
```
maisem, line 26: Wireguard puts this in `\\.\pipe\ProtectedPrefix\Administrators\WireGuard`. I'm not sure if we should put ours under that path too. I can't find any documentation about this anywhere

e.g. `\\.\pipe\ProtectedPrefix\Administrators\tailscaled`

@dblohm7 @raggi thoughts?

DentonGentry:
> Otherwise I think we'll just go with whatever Wireguard-go uses, on the assumption that it
> must not be too bad if it has a bunch of users.

We have a working group for decisions like this, and we have enough time before 1.34 to work it out.
If we need to discuss it further we can do so.

dblohm7:
> > @dblohm7 I've added another commit which uses the `go-winio` pkg. Does that seem better?
> 
> I'll need to look. I think the other package was based on this one, so I need to see whether it has the same issues.

For our purposes I think this will be okay.

dblohm7:
> Otherwise I think we'll just go with whatever Wireguard-go uses, on the assumption that it must not be too bad if it has a bunch of users.

Having written a high-throughput Windows named pipe server, and having seen deadlocks happen from misused I/O completion ports, I am not comfortable with assuming anything. Having said that, I think we're okay in this case.

paths/paths.go
dblohm7: Looking...

dblohm7: https://www.tiraniddo.dev/2017/11/named-pipe-secure-prefixes.html

I'm spinning up a Windows 7 VM to see what this looks like.

dblohm7: @maisem I have definitely never heard of that before and it appears to be completely undocumented, but I also found that this setup works on Windows 7 exactly as described in that blog post. We might as well do that here, too.

safesocket/pipe_windows.go
```diff
@@ -26,20 +25,48 @@ func setFlags(network, address string, c syscall.RawConn) error {
...
-//
-//	I tried to use winio.ListenPipe() here, but that code is a disaster,
-//	built on top of an API that's a disaster. So for now we'll hack it by
-//	just always using a TCP session on a fixed port on localhost. As a
-//	result, on Windows we ignore the vendor and name strings.
-//	NOTE(bradfitz): Jason did a new pipe package: https://go-review.googlesource.com/c/sys/+/299009
 func listen(path string, port uint16) (_ net.Listener, gotPort uint16, _ error) {
-	lc := net.ListenConfig{
-		Control: setFlags,
+	// Construct the default named pipe security descriptor.
```
dblohm7, line 29: I don't think you need this. Just use the ACL that we're creating.

maisem: Will test, it didn't work last time I tried this. But I think I had the wrong DACL at the time.

safesocket/pipe_windows.go
dblohm7: That group does exist (it's an alias to other built-in stuff), but I'm fine with this code as-is for the time being.

Perhaps my only suggestion is that, since this ACL involves a well-known group, we don't need to go through the whole `ACLFromEntries` rigamarole every time -- we could just put the serialized SDDL string into the code as a constant and pass that straight into `PipeConfig`.

dblohm7 (changes requested):
This looks good! My only nit is that I'd like you to remove the `RtlDefaultNpAcl` bits and create the `DACL` with just our own ACL.

paths/paths.go
maisem: Done

safesocket/pipe_windows.go
maisem: Done, PTAL

dblohm7 (approved):

PR #6431, opened 2022-11-21
ipn/ipnlocal: call checkPrefs in Start too

maisem:
We were not calling checkPrefs on `opts.*Prefs` in (*LocalBackend).Start().

Updates #713


mihaip (approved):
There's some risk that people who had temporarily added `TAILSCALE_USE_WIP_CODE` to enable SSH on unsupported platforms (and then removed the env variable) will now have a `tailscaled` that won't start up, though I suppose it's on them for using WIP code.

PR #6434, opened 2022-11-22
ipn/localapi: add LocalAPI support for removing all profiles

mihaip:
Will be used to implement a "Log out all accounts" feature.

Updates #713

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -4204,6 +4204,16 @@ func (b *LocalBackend) DeleteProfile(p ipn.ProfileID) error {
 	return b.resetForProfileChangeLockedOnEntry()
 }
 
+// DeleteProfiles removes all known profiles.
+func (b *LocalBackend) DeleteAllProfiles() error {
+	b.mu.Lock()
+	if err := b.pm.DeleteAllProfiles(); err != nil {
+		b.mu.Unlock()
+		return err
+	}
+	return b.resetForProfileChangeLockedOnEntry()
+}
+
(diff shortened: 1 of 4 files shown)


ipn/ipnlocal/local.go
```diff
@@ -4204,6 +4204,16 @@ func (b *LocalBackend) DeleteProfile(p ipn.ProfileID) error {
 	return b.resetForProfileChangeLockedOnEntry()
 }
 
+// DeleteProfiles removes all known profiles.
```
bradfitz, line 4207: DeleteAllProfiles

ipn/ipnlocal/profiles.go
```diff
@@ -349,6 +350,21 @@ func (pm *profileManager) DeleteProfile(id ipn.ProfileID) error {
...
+	metricDeleteAllProfile.Add(1)
+
+	for _, kp := range pm.knownProfiles {
+		if err := pm.store.WriteState(kp.Key, nil); err != nil {
+			return err
+		}
+	}
+	pm.NewProfile()
+	maps.Clear(pm.knownProfiles)
+	return pm.writeKnownProfiles()
```
bradfitz, line 365: on failure, this leaves things is a kinda inconsistent state?

Should you try to writeKnownProfiles on your way out of the error path?

mihaip:
Maisem and I DM'ed about this a bit, and we'll probably want something different for "Log out all accounts" (it needs to talk to control for each account to revoke the key). Moving this back to draft.

mihaip:
Closing this, we haven't need the feature, and the useful bits were pulled into #7094

Issue #6417, opened 2022-11-20
ipn: get rid of IPN wire protocol, finish migration to HTTP everywhere


bradfitz:
The ipnserver has two protocols: the "IPN message bus" (streaming 4 byte length + JSON) ipn.Notify / ipn.Commands, and HTTP.

We almost entirely use HTTP these days, but some IPN message bus users remain.
 
Tracking bug to completely migrate to HTTP to simplify code.

/cc @maisem @mihaip

mihaip:
Closing this out, all clients have migrated to LocalAPI/watch-ipn-bus, and the old wire protocol is gone. Followup work is being tracked separately (e.g. tailscale/corp#8983).

PR #6435, opened 2022-11-22
ipn{,/ipnlocal}, client/tailscale: move Taildrop recv notifications to LocalAPI HTTP method

bradfitz:
Updates #6417

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -181,7 +181,8 @@ type LocalBackend struct {
 	peerAPIListeners []*peerAPIListener
 	loginFlags       controlclient.LoginFlags
 	incomingFiles    map[*incomingFile]bool
-	lastStatusTime   time.Time // status.AsOf value of the last processed status update
+	fileWaiters      map[*mapSetHandle]context.CancelFunc // handle => func to call on file received
+	lastStatusTime   time.Time                            // status.AsOf value of the last processed status update
 	// directFileRoot, if non-empty, means to write received files
 	// directly to this directory, without staging them in an
 	// intermediate buffered directory for "pick-up" later. If
@@ -1699,6 +1700,9 @@ func (b *LocalBackend) sendFileNotify() {
 	var n ipn.Notify
 
 	b.mu.Lock()
+	for _, wakeWaiter := range b.fileWaiters {
+		wakeWaiter()
+	}
 	notifyFunc := b.notify
 	apiSrv := b.peerAPIServer
 	if notifyFunc == nil || apiSrv == nil {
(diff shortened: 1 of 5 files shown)


ipn/ipnlocal/local.go
```diff
@@ -3558,6 +3562,47 @@ func (b *LocalBackend) WaitingFiles() ([]apitype.WaitingFile, error) {
 	return apiSrv.WaitingFiles()
 }
 
+// mapSetHandle is a minimal (but non-zero) value whose addresse serves as a map
```
andrew-d, line 3565:
```suggestion
// mapSetHandle is a minimal (but non-zero) value whose address serves as a map
```

ipn/ipnlocal/local.go
```diff
@@ -3558,6 +3562,47 @@ func (b *LocalBackend) WaitingFiles() ([]apitype.WaitingFile, error) {
...
+	b.setFileWaiter(handle, gotFileCancel)
+	defer b.setFileWaiter(handle, nil)
+
+	// After we've registered our waiter, check if anything's available. This
+	// guarantees there's no window for a file to arrive between checking and
+	// registering our waiter and us blocking forever.
+	go func() {
+		if ff, _ := b.WaitingFiles(); len(ff) > 0 {
+			gotFileCancel()
+		}
```
andrew-d, line 3595: You could return the `ff` from `AwaitFile` here and simplify the `for` loop in `serveFiles`, I think?

bradfitz: I didn't really want to reason about the races. Getting the list is cheap. I just want some cheap sleep+wake-up call.

Would you prefer it returned nothing instead? Then caller can check its own ctx.Err() that it passed in.

bradfitz: Oh I think I know what you mean now. Let me try.

bradfitz: PTAL

ipn/ipnlocal/local.go
```diff
@@ -3551,13 +3555,63 @@ func (b *LocalBackend) TestOnlyPublicKeys() (machineKey key.MachinePublic, nodeK
...
+		return ff, err
+	}
+
+	for {
+		gotFile, gotFileCancel := context.WithCancel(context.Background())
+		defer gotFileCancel()
+
+		handle := new(mapSetHandle)
+		b.setFileWaiter(handle, gotFileCancel)
+		defer b.setFileWaiter(handle, nil)
```
andrew-d, line 3595: Hm, technically these `defer`s won't fire until after the `for` loop is finished... probably fine in practice, though.

andrew-d (approved):

ipn/ipnlocal/local.go
bradfitz: Yeah. I couldn't convince myself it mattered.

PR #6444, opened 2022-11-22
derp: prevent concurrent access to multiForwarder map

### LINKED ISSUE #6445: derp: concurrent map iteration and map write
### What is the issue?

```
Nov 22 10:41:00 derp derper[1465140]: fatal error: concurrent map iteration and map write
Nov 22 10:41:00 derp derper[1465140]: goroutine 49915768 [running]:
Nov 22 10:41:00 derp derper[1465140]: tailscale.com/derp.multiForwarder.ForwardPacket(0x94?, {{0xf6, 0x33, 0x7c, 0xac, 0x1b, 0x5b, 0x45, 0x17, 0x4b, ...}}, ...)
Nov 22 10:41:00 derp derper[1465140]:         tailscale.com@v1.1.1-0.20221019152536-e966f024b0da/derp/derp_server.go:1651 +0x91
Nov 22 10:41:00 derp derper[1465140]: tailscale.com/derp.(*sclient).handleFrameSendPacket(0xc008719040, 0x80?, 0x8?)
Nov 22 10:41:00 derp derper[1465140]:         tailscale.com@v1.1.1-0.20221019152536-e966f024b0da/derp/derp_server.go:932 +0x416
```

### Steps to reproduce

_No response_

_No response_

### OS

_No response_

### OS version

_No response_

### Tailscale version

_No response_

### Bug report

_No response_

knyar:
Instead of iterating over the map to determine the preferred forwarder on every packet (which could happen concurrently with map mutations), store it separately in an atomic variable.

Fixes #6445

### DIFF
--- a/derp/derp_server.go
+++ b/derp/derp_server.go
@@ -40,6 +40,7 @@ import (
 	"tailscale.com/disco"
 	"tailscale.com/envknob"
 	"tailscale.com/metrics"
+	"tailscale.com/syncs"
 	"tailscale.com/types/key"
 	"tailscale.com/types/logger"
 	"tailscale.com/version"
@@ -1560,22 +1561,20 @@ func (s *Server) AddPacketForwarder(dst key.NodePublic, fwd PacketForwarder) {
 			// Duplicate registration of same forwarder. Ignore.
 			return
 		}
-		if m, ok := prev.(multiForwarder); ok {
-			if _, ok := m[fwd]; ok {
+		if m, ok := prev.(*multiForwarder); ok {
+			if _, ok := m.all[fwd]; ok {
 				// Duplicate registration of same forwarder in set; ignore.
 				return
 			}
-			m[fwd] = m.maxVal() + 1
+			m.add(fwd)
 			return
 		}
 		if prev != nil {
 			// Otherwise, the existing value is not a set,
 			// not a dup, and not local-only (nil) so make
-			// it a set.
-			fwd = multiForwarder{
-				prev: 1, // existed 1st, higher priority
-				fwd:  2, // the passed in fwd is in 2nd place
-			}
+			// it a set. `prev` existed first, so will have higher
+			// priority.
+			fwd = newMultiForwarder(prev, fwd)
 			s.multiForwarderCreated.Add(1)
 		}
 	}
@@ -1591,19 +1590,14 @@ func (s *Server) RemovePacketForwarder(dst key.NodePublic, fwd PacketForwarder) 
 	if !ok {
 		return
 	}
-	if m, ok := v.(multiForwarder); ok {
-		if len(m) < 2 {
+	if m, ok := v.(*multiForwarder); ok {
+		if len(m.all) < 2 {
 			panic("unexpected")
 		}
-		delete(m, fwd)
-		// If fwd was in m and we no longer need to be a
-		// multiForwarder, replace the entry with the
-		// remaining PacketForwarder.
-		if len(m) == 1 {
-			var remain PacketForwarder
-			for k := range m {
-				remain = k
-			}
+		if remain, isLast := m.deleteLocked(fwd); isLast {
+			// If fwd was in m and we no longer need to be a
+			// multiForwarder, replace the entry with the
+			// remaining PacketForwarder.
 			s.clientsMesh[dst] = remain
 			s.multiForwarderDeleted.Add(1)
 		}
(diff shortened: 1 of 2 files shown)


bradfitz:
What's the background? No reference to a bug and not much in the commit message.

Did you observe a race in practice? What's the race dumps?

Did a user report this? Was there a crash?

Thanks for fixing & the new test, in any case. I just want richer commit messages.

knyar:
> What's the background? No reference to a bug and not much in the commit message.

Sorry, I was waiting for CI to pass before moving this from draft to ready :)

derp/derp_test.go
```diff
@@ -785,6 +790,72 @@ func TestForwarderRegistration(t *testing.T) {
...
+	f.c <- packet
+	return nil
+}
+
+func TestMultiForwarder(t *testing.T) {
+	received := 0
+	var wg sync.WaitGroup
+	ch := make(chan []byte)
+	ctx, cancel := context.WithCancel(context.Background())
+	wg.Add(1)
```
bradfitz, line 808: Let's do `wg.Add(2)` here rather than worrying about whether there are concurrent Done calls with Add (which WaitGroup panics on).

Looks safe, though, because cancel isn't called until late, but I don't want readers to even think about that.

derp/derp_test.go
```diff
@@ -785,6 +790,72 @@ func TestForwarderRegistration(t *testing.T) {
...
+			s.RemovePacketForwarder(u, &channelFwd{1, ch})
+			s.AddPacketForwarder(u, channelFwd{1, ch})
+			s.RemovePacketForwarder(u, channelFwd{3, ch})
+			if ctx.Err() != nil {
+				return
+			}
+		}
+	}()
+
+	numMsgs := 1 << 12
```
bradfitz, line 844: Go generally only uses shifts of << 10, 20, 30 (for KB, MB, GB, etc)

derp/derp_test.go
```diff
@@ -785,6 +790,72 @@ func TestForwarderRegistration(t *testing.T) {
 	})
 }
 
+type channelFwd struct {
+	id int
```
bradfitz, line 794: what's this ID being used for? just debug prints, or?

derp/derp_server.go
```diff
@@ -1560,21 +1560,23 @@ func (s *Server) AddPacketForwarder(dst key.NodePublic, fwd PacketForwarder) {
 			// Duplicate registration of same forwarder. Ignore.
 			return
 		}
-		if m, ok := prev.(multiForwarder); ok {
-			if _, ok := m[fwd]; ok {
+		if m, ok := prev.(*multiForwarder); ok {
+			if m.exists(fwd) {
```
bradfitz, line 1564: so I think the prior assumption was all these multiForwarders were guarded by s.mu.

Where is that not the case? I didn't see the full race output.

knyar: Annoyingly, I could not capture stack traces of all goroutines before they expired from systemd journal, but I suspect what happened here was a map write concurrent with this [ForwardPacket](https://github.com/tailscale/tailscale/blob/0f7da5c7dc0ae30f02320ede18d20473ad7ca981/derp/derp_server.go#L933) that needs to iterate over the map to find the lowest id.

derp/derp_test.go
knyar: It's to make sure that different instances that reference the same channel hash to different values. I've added a comment to clarify.

derp/derp_test.go
knyar: Done.

derp/derp_test.go
knyar: Replaced with a number.

derp/derp_server.go
```diff
@@ -1635,26 +1632,57 @@ func (s *Server) RemovePacketForwarder(dst key.NodePublic, fwd PacketForwarder)
...
+		for k := range f.m {
+			return k, true
+		}
+	}
+	return nil, false
 }
 
-func (m multiForwarder) ForwardPacket(src, dst key.NodePublic, payload []byte) error {
+func (f *multiForwarder) ForwardPacket(src, dst key.NodePublic, payload []byte) error {
+	f.mu.RLock()
```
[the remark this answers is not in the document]
knyar: Yeah I have realized that it's also a bit silly to iterate over a map on every packet just to determine what forwarder to use, so I've replace the lock with a `syncs.AtomicValue` for the preferred forwarder. Thanks for the suggestion!

derp/derp_server.go
```diff
@@ -1635,27 +1629,63 @@ func (s *Server) RemovePacketForwarder(dst key.NodePublic, fwd PacketForwarder)
...
 // packets consistently to the same node and don't pick randomly.
-type multiForwarder map[PacketForwarder]uint8
+type multiForwarder struct {
+	fwd syncs.AtomicValue[PacketForwarder] // preferred forwarder.
+	all map[PacketForwarder]uint8          // all forwarders, protected by s.mu.
+}
 
-func (m multiForwarder) maxVal() (max uint8) {
-	for _, v := range m {
+// newMultiForwarder creates a new multiForwarder. First PacketForwarder
```
bradfitz, line 1637: // The first ...

derp/derp_server.go
```diff
@@ -1635,27 +1629,63 @@ func (s *Server) RemovePacketForwarder(dst key.NodePublic, fwd PacketForwarder)
...
-func (m multiForwarder) ForwardPacket(src, dst key.NodePublic, payload []byte) error {
-	var fwd PacketForwarder
-	var lowest uint8
-	for k, v := range m {
-		if fwd == nil || v < lowest {
-			fwd = k
-			lowest = v
+// delete removes a packet forwarder from the map. If only one forwarder remains after
+// the removal, it will be returned alongside a `true` boolean value.
+func (f *multiForwarder) delete(fwd PacketForwarder) (PacketForwarder, bool) {
```
bradfitz, line 1662: `(_ PacketForward, exactlyOneRemains bool)` for documentation.

I'm tempted to call that variable `uno` (for https://en.wikipedia.org/wiki/Uno_(card_game)) but that's probably too esoteric. In that game, you have to yell "UNO!" when you only have one card left :)

derp/derp_server.go
```diff
@@ -1635,27 +1629,63 @@ func (s *Server) RemovePacketForwarder(dst key.NodePublic, fwd PacketForwarder)
...
+		// based on the lowest index.
+		var lowestfwd PacketForwarder
+		var lowest uint8
+		for k, v := range f.all {
+			if lowestfwd == nil || v < lowest {
+				lowestfwd = k
+				lowest = v
+			}
 		}
+		f.fwd.Store(lowestfwd)
```
bradfitz, line 1676: be careful not to store nil here. That'll panic.

derp/derp_server.go
knyar: Fixed.

derp/derp_server.go
knyar: I've called it `isLast` to match the variable on line 1597 – it's a bit shorter and I think conveys the same meaning?

derp/derp_server.go
knyar: Added a nil check.

derp/derp_server.go
```diff
@@ -1635,27 +1629,65 @@ func (s *Server) RemovePacketForwarder(dst key.NodePublic, fwd PacketForwarder)
...
-func (m multiForwarder) ForwardPacket(src, dst key.NodePublic, payload []byte) error {
-	var fwd PacketForwarder
-	var lowest uint8
-	for k, v := range m {
-		if fwd == nil || v < lowest {
-			fwd = k
-			lowest = v
+// delete removes a packet forwarder from the map. If only one forwarder remains after
+// the removal, it will be returned alongside a `true` boolean value.
+func (f *multiForwarder) delete(fwd PacketForwarder) (_ PacketForwarder, isLast bool) {
```
bradfitz, line 1662: should we name this deleteLocked?

at least document what mutex must be held in the docs?

bradfitz (approved):

PR #6448, opened 2022-11-22
ipn/ipnlocal: call initTKALocked on backend start

twitchyliquid64:
Somewhere in the last week's changes to ipnlocal we broke TKA initialization. This manifests later, which upon getting a netmap TKA tries to initialize from scratch but finds it already had stored state:

```
tkaSyncIfNeeded: enabled=true, head=Q5UJDFVMPCZFQH6TLRE7KZSSBXOZGMKFLBV2LCGE75A7SSXFVL6A
TKA sync error: bootstrap: tka bootstrap: tailchonk is not empty
```

Is this the right place to call `initTKALocked` in the new, profiley world?


maisem (approved):
This looks correct.

PR #6450, opened 2022-11-22
ipn, cmd/tailscale/cli: add LocalAPI IPN bus watch, Start, convert CLI

bradfitz:
Updates #6417
Updates tailscale/corp#8051

### DIFF
--- a/client/tailscale/localclient.go
+++ b/client/tailscale/localclient.go
@@ -555,6 +555,20 @@ func (lc *LocalClient) EditPrefs(ctx context.Context, mp *ipn.MaskedPrefs) (*ipn
 	return decodeJSON[*ipn.Prefs](body)
 }
 
+// StartLoginInteractive starts an interactive login.
+func (lc *LocalClient) StartLoginInteractive(ctx context.Context) error {
+	_, err := lc.send(ctx, "POST", "/localapi/v0/login-interactive", http.StatusNoContent, nil)
+	return err
+}
+
+// Start applies the configuration specified in opts, and starts the
+// state machine.
+func (lc *LocalClient) Start(ctx context.Context, opts ipn.Options) error {
+	_, err := lc.send(ctx, "POST", "/localapi/v0/start", http.StatusNoContent, jsonBody(opts))
+	return err
+}
+
+// Logout logs out the current node.
 func (lc *LocalClient) Logout(ctx context.Context) error {
 	_, err := lc.send(ctx, "POST", "/localapi/v0/logout", http.StatusNoContent, nil)
 	return err
@@ -965,4 +979,79 @@ func (lc *LocalClient) SwitchProfile(ctx context.Context, profile ipn.ProfileID)
 func (lc *LocalClient) DeleteProfile(ctx context.Context, profile ipn.ProfileID) error {
 	_, err := lc.send(ctx, "DELETE", "/localapi/v0/profiles"+url.PathEscape(string(profile)), http.StatusNoContent, nil)
 	return err
+}
+
+// WatchIPNMask are filtering options for LocalClient.WatchIPNBus.
+//
+// The zero value is a valid WatchOpt that means to watch everything.
+//
+// TODO(bradfitz): flesh out.
+type WatchIPNMask uint64
+
+// WatchIPNBus subscribes to the IPN notification bus. It returns a watcher
+// once the bus is connected successfully.
+//
+// The context is used for the life of the watch, not just the call to
+// WatchIPNBus.
+//
+// The returned IPNBusWatcher's Close method must be called when done to release
+// resources.
+func (lc *LocalClient) WatchIPNBus(ctx context.Context, mask WatchIPNMask) (*IPNBusWatcher, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET",
+		"http://"+apitype.LocalAPIHost+"/localapi/v0/watch-ipn-bus?mask="+fmt.Sprint(mask),
+		nil)
+	if err != nil {
+		return nil, err
+	}
+	res, err := lc.doLocalRequestNiceError(req)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode != 200 {
+		res.Body.Close()
+		return nil, errors.New(res.Status)
+	}
+	dec := json.NewDecoder(res.Body)
+	return &IPNBusWatcher{
+		ctx:     ctx,
+		httpRes: res,
+		dec:     dec,
+	}, nil
+}
+
+// IPNBusWatcher is an active subscription (watch) of the local tailscaled IPN bus.
+// It's returned by LocalClient.WatchIPNBus.
+//
+// It must be closed when done.
+type IPNBusWatcher struct {
+	ctx     context.Context // from original WatchIPNBus call
+	httpRes *http.Response
+	dec     *json.Decoder
+
+	mu     sync.Mutex
+	closed bool
+}
+
+// Close stops the watcher and releases its resources.
+func (w *IPNBusWatcher) Close() error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	if w.closed {
+		return nil
+	}
+	w.closed = true
+	return w.httpRes.Body.Close()
+}
+
+// Next returns the next ipn.Notify from the stream.
+// If the context from LocalClient.WatchIPNBus is done, that error is returned.
+func (w *IPNBusWatcher) Next() (ipn.Notify, error) {
+	var n ipn.Notify
+	if err := w.dec.Decode(&n); err != nil {
+		if cerr := w.ctx.Err(); cerr != nil {
+			err = cerr
+		}
+		return ipn.Notify{}, err
+	}
+	return n, nil
 }
--- a/cmd/tailscale/cli/up.go
+++ b/cmd/tailscale/cli/up.go
@@ -15,11 +15,13 @@ import (
 	"log"
 	"net/netip"
 	"os"
+	"os/signal"
 	"reflect"
 	"runtime"
 	"sort"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	shellquote "github.com/kballard/go-shellquote"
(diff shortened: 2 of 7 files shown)


bradfitz:
Oh, I'm also told @danderson will like this change for some Kubernetes reason I'm not up to speed on. But hope it helps!

client/tailscale/localclient.go
```diff
@@ -966,3 +981,74 @@ func (lc *LocalClient) DeleteProfile(ctx context.Context, profile ipn.ProfileID)
...
+		nil)
+	if err != nil {
+		return nil, err
+	}
+	res, err := lc.doLocalRequestNiceError(req)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode != 200 {
+		defer res.Body.Close()
```
andrew-d, line 1012: Any reason to `defer` here?

client/tailscale/localclient.go
```diff
@@ -966,3 +981,74 @@ func (lc *LocalClient) DeleteProfile(ctx context.Context, profile ipn.ProfileID)
...
 }
+
+// WatchIPNMask are filtering options for LocalClient.WatchIPNBus.
+//
+// The zero value is a valid WatchOpt that means to watch everything.
+//
+// TODO(bradfitz): flesh out.
+type WatchIPNMask uint64
+
+// WatchIPNBus subscribes to the IPN notification bus and calls fn on any
```
maisem, line 992: no fn

client/tailscale/localclient.go
```diff
@@ -966,3 +981,74 @@ func (lc *LocalClient) DeleteProfile(ctx context.Context, profile ipn.ProfileID)
...
+	dec     *json.Decoder
+	closed  atomic.Bool
+}
+
+// Close stops the watcher and releases its resources.
+func (w *IPNBusWatcher) Close() error {
+	if w.closed.Load() {
+		return nil
+	}
+	w.closed.Store(true)
```
andrew-d, line 1039: This is racy.
```suggestion
	if !w.closed.CompareAndSwap(false, true) {
			return nil
	}
```

client/tailscale/localclient.go
bradfitz: Fixed. That was from an earlier signature.

cmd/tailscale/cli/up.go
```diff
@@ -660,10 +662,13 @@ func runUp(ctx context.Context, cmd string, args []string, upArgs upArgsT) (retE
 		if err != nil {
 			return err
 		}
-		bc.Start(ipn.Options{
+		err = localClient.Start(ctx, ipn.Options{
```
maisem, line 665: `if err := ...`

cmd/tailscale/cli/web.go
```diff
@@ -344,23 +356,32 @@ func webHandler(w http.ResponseWriter, r *http.Request) {
...
-		prefs, err := localClient.GetPrefs(r.Context())
-		if err != nil && !postData.Reauthenticate {
+
+		routes, err := calcAdvertiseRoutes(postData.AdvertiseRoutes, postData.AdvertiseExitNode)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			json.NewEncoder(w).Encode(mi{"error": err.Error()})
+			return
+		}
+		prefs.AdvertiseRoutes = routes
```
maisem, line 366: old?

cmd/tailscale/cli/web.go
```diff
@@ -344,23 +356,32 @@ func webHandler(w http.ResponseWriter, r *http.Request) {
...
-				return
-			}
-			prefs.AdvertiseRoutes = routes
 		}
 
 		w.Header().Set("Content-Type", "application/json")
-		url, err := tailscaleUp(r.Context(), prefs, postData.Reauthenticate)
+		log.Printf("tailscaleUp(reauth=%v) ...", postData.Reauthenticate)
+		url, err := tailscaleUp(r.Context(), st, postData.Reauthenticate)
+		log.Printf("tailscaleUp = (URL %v, %v)", url != "", err)
```
maisem, line 384: debug?

client/tailscale/localclient.go
```diff
@@ -966,3 +981,74 @@ func (lc *LocalClient) DeleteProfile(ctx context.Context, profile ipn.ProfileID)
...
+	return w.httpRes.Body.Close()
+}
+
+// Next returns the next ipn.Notify from the stream.
+// If the context is LocalClient.WatchIPNBus is done, that error is returned.
```
danderson, line 1044: Parse error. I think you meant the context _from_ LocalClient.WatchIPNBus? But I'm not sure.

cmd/tailscale/cli/web.go
maisem: if you want to update it to store the new value, use the output of `EditPrefs`?

ipn/ipnlocal/local.go
```diff
@@ -1690,24 +1691,70 @@ func (b *LocalBackend) readPoller() {
...
+//
+// WatchNotifications blocks until ctx is done.
+//
+// The provided fn will only be called with non-nil pointers. The caller must
+// not modify roNotify. If fn returns false, the watch also stops.
+//
+// Failure to consume a many notifications will result in dropped notifications.
+// There is currently (2022-11-22) no mechanism provided to detect when a
+// message has been dropped.
+func (b *LocalBackend) WatchNotifications(ctx context.Context, fn func(roNotify *ipn.Notify) (keepGoing bool)) {
```
maisem, line 1705: make a view?

ipn/ipnlocal/local.go
```diff
@@ -1690,24 +1691,70 @@ func (b *LocalBackend) readPoller() {
...
-func (b *LocalBackend) send(n ipn.Notify) {
+// WatchNotifications subscribes to the ipn.Notify message bus notification
+// messages.
+//
+// WatchNotifications blocks until ctx is done.
+//
+// The provided fn will only be called with non-nil pointers. The caller must
+// not modify roNotify. If fn returns false, the watch also stops.
+//
+// Failure to consume a many notifications will result in dropped notifications.
```
maisem, line 1702: type in `to consume a many`

ipn/ipnlocal/local.go
```diff
@@ -1690,24 +1691,70 @@ func (b *LocalBackend) readPoller() {
...
 	}
 
-	n.Version = version.Long
-	notifyFunc(n)
+	for _, ch := range b.notifyWatchers {
+		select {
+		case ch <- &n:
+		default:
+			// Drop the notification if the channel is full.
+		}
```
andrew-d, line 1750: Up to you, but another option would be prefer newer messages; something like:
```go
select {
case ch <- &n:
default:
	select {
	case <- ch: // drain unread message
	default:
	}
	ch <- &n
}
```

client/tailscale/localclient.go
bradfitz: Nope. It was movement and I didn't notice.

client/tailscale/localclient.go
bradfitz: I almost did that, but then one call can return nil before the other call is done closing.

OTOH a double close of a response body is safe.

danderson (commented):
Skim pass looks plausible. Given our recent excitement with LocalAPI though, I'd like to take more time to review carefully, unless someone else beats me to it.

Yes, this is useful for the container image, because it won't have to shell out to the CLI to auth any more.

client/tailscale/localclient.go
bradfitz: reverted to plain ol' mutex

andrew-d: Could also use a `sync.Once` if that's something we care about

ipn/ipnlocal/local.go
bradfitz: Not yet. When I update macOS I'm going to move its ipn.Notify merging code into here which is even better.

cmd/tailscale/cli/web.go
bradfitz: yes, debug I want to keep. Seems fine?

cmd/tailscale/cli/up.go
bradfitz: done

bradfitz:
> Given our recent excitement with LocalAPI though, I'd like to take more time to review carefully, unless someone else beats me to it.

It's behind the same checks that were recently hardened. Also, see https://github.com/tailscale/corp/issues/8051

ipn/ipnlocal/local.go
bradfitz: I don't think it works.

Also, this is all temporary. And the read-only rules only apply to within the tailscaled process. Once they come out the other side in user code then it's owned by the caller.

bradfitz:
Updated. PTAL.

cmd/tailscale/cli/up.go
```diff
@@ -535,109 +537,109 @@ func runUp(ctx context.Context, cmd string, args []string, upArgs upArgsT) (retE
...
+	defer cancelWatch()
+	watcher, err := localClient.WatchIPNBus(watchCtx, 0)
+	if err != nil {
+		return err
+	}
+	defer watcher.Close()
+
+	go func() {
+		interrupt := make(chan os.Signal, 1)
+		signal.Notify(interrupt, syscall.SIGINT, syscall.SIGTERM)
```
andrew-d, line 550: This apparently does compile on Windows, since there's made-up values in the `syscall` package, but does it actually work correctly? I guess we do have another usage in `cmd/tailscale/cli/cli.go`, so probably?

cmd/tailscale/cli/web.go
```diff
@@ -418,45 +427,46 @@ func webHandler(w http.ResponseWriter, r *http.Request) {
...
-		prefs = ipn.NewPrefs()
-		prefs.ControlURL = ipn.DefaultControlURL
-		prefs.WantRunning = true
-		prefs.CorpDNS = true
-		prefs.AllowSingleHosts = true
-		prefs.ForceDaemon = (runtime.GOOS == "windows")
-	}
-
-	if distro.Get() == distro.Synology {
-		prefs.NetfilterMode = preftype.NetfilterOff
```
andrew-d, line 433: Did we lose this behaviour in the refactor?

cmd/tailscale/cli/web.go
```diff
@@ -344,23 +356,32 @@
...
-				w.WriteHeader(http.StatusInternalServerError)
-				json.NewEncoder(w).Encode(mi{"error": err.Error()})
-				return
-			}
-			prefs.AdvertiseRoutes = routes
 		}
 
 		w.Header().Set("Content-Type", "application/json")
-		url, err := tailscaleUp(r.Context(), prefs, postData.Reauthenticate)
+		log.Printf("tailscaleUp(reauth=%v) ...", postData.Reauthenticate)
```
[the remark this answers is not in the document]
bradfitz: It's a frickin' bool from the network, github, relax. It's being logged on purpose.

ipn/localapi/localapi.go
```diff
@@ -586,6 +616,29 @@ func (h *Handler) serveLoginInteractive(w http.ResponseWriter, r *http.Request)
 	return
 }
 
+func (h *Handler) serveStart(w http.ResponseWriter, r *http.Request) {
+	if !h.PermitWrite {
+		http.Error(w, "access denied", http.StatusForbidden)
+		return
+	}
+	if r.Method != "POST" {
+		http.Error(w, "want POST", 400)
```
andrew-d, line 625:
```suggestion
		http.Error(w, "want POST", http.StatusBadRequest)
```

andrew-d (commented):
LGTM from me, but I'd prefer another +1 from someone else too since I'm pretty zonked out today 😅

cmd/tailscale/cli/web.go
bradfitz: Not needed. EditPrefs only edits the specified prefs (MaskedPrefs). We only needed this before because it was using the bad SetPrefs API which requires everything.

ipn/localapi/localapi.go
bradfitz: there are 25 occurrences of this in that file. Let's keep it for now and decide how to clean them all up in the same way in its own commit. Consistency is better for now.

cmd/tailscale/cli/up.go
bradfitz: This is just code movement from cli.go.

ipn/localapi/localapi.go
andrew-d: SGTM

client/tailscale/localclient.go
```diff
@@ -555,6 +555,20 @@ func (lc *LocalClient) EditPrefs(ctx context.Context, mp *ipn.MaskedPrefs) (*ipn
 	return decodeJSON[*ipn.Prefs](body)
 }
 
+// StartLoginInteractive starts an interactive login.
+func (lc *LocalClient) StartLoginInteractive(ctx context.Context) error {
+	_, err := lc.send(ctx, "POST", "/localapi/v0/login-interactive", http.StatusNoContent, nil)
```
maisem, line 560: this doesnt seem to be implemented

maisem (approved):

PR #6466, opened 2022-11-23
tsweb: add EnforceHostname helper for DNS rebinding mitigation

willnorris:
I'm still working on a few things, but still wanted to get this out for review of the basic approach.  A few notes:

 - I'm not sure if tsweb is the best package for this.  It's already a grab bag of random web related things, but I don't love that this pulls in the tailscale.com/client/tailscale package.  Maybe that's okay?  The other option would be putting it in tsnet, but it's not actually exclusive to tsnet use cases.
 - `hostEnforcer` isn't quite the right name for the type, since it's not actually doing the enforcement.  Still looking for better names.
 - I'm caching the local node's dnsname for 5 seconds so that we pickup name changes.  I'm not sure how expensive a StatusWithoutPeers call is, so we could definitely cache for longer if needed.


tsweb/tsweb.go
```diff
@@ -779,3 +781,67 @@ var (
...
+
+// EnforceHostname wraps a handler and enforces that all inbound requests have a Host header that matches the local Tailscale node.
+// Acceptable Host headers are bare IP addresses, the node's full DNS name (node.tailnet.ts.net) and the node's bare hostname.
+// This is used to help protect against DNS rebinding attacks.
+// If client is nil, a default client will be used.
+func EnforceHostname(h http.Handler, client *tailscale.LocalClient) http.Handler {
+	if client == nil {
+		client = &tailscale.LocalClient{}
+	}
+	enforcerAtomic := atomic.Pointer[hostEnforcer]{}
```
bradfitz, line 817: `var enforcerAtomic atomic.Pointer[hostEnforcer]`

... has less punctuation

tsweb/tsweb.go
```diff
@@ -779,3 +781,67 @@ var (
...
+			h.ServeHTTP(w, r)
+			return
+		}
+
+		enforcer := enforcerAtomic.Load()
+		// refresh the list of allowed hosts every 5 seconds
+		if now := time.Now(); enforcer == nil || now.After(enforcer.expiry) {
+			newEnforcer := &hostEnforcer{expiry: now.Add(5 * time.Second)}
+			if enforcerAtomic.CompareAndSwap(enforcer, newEnforcer) {
+				enforcer = enforcerAtomic.Load()
```
bradfitz, line 831: I think you mean if NOT CompareAndSwap. If it returned true, then the enforcer you want is newEnforcer. Only if you lost the race do you need to load it.

tsweb/tsweb.go
```diff
@@ -779,3 +781,67 @@ var (
...
+			}
+		}
+		allowedHosts, err := enforcer.allowedHosts(r.Context(), client)
+		if err != nil {
+			http.Error(w, "unable to get tailscale status", http.StatusInternalServerError)
+			return
+		}
+		for _, host := range allowedHosts {
+			if r.Host == host {
```
bradfitz, line 840: do you need to normalize case?

PR #6467, opened 2022-11-23
ipn/ipnauth: start splitting ipnserver into new ipnauth package

bradfitz:
We're trying to gut 90% of the ipnserver package. A lot will get deleted, some will move to LocalBackend, and a lot is being moved into this new ipn/ipnauth package which will be leaf-y and testable.

This is a baby step towards moving some stuff to ipnauth.

Update #6417
Updates tailscale/corp#8051


ipn/ipnauth/ipnauth.go
```diff
@@ -0,0 +1,204 @@
...
+	"tailscale.com/util/pidowner"
+	"tailscale.com/util/winutil"
+	"tailscale.com/version/distro"
+)
+
+// ConnIdentity represents the owner of a localhost TCP or unix socket connection
+// connecting to the LocalAPI.
+type ConnIdentity struct {
+	conn       net.Conn
+	notWindows bool // runtime.GOOS != "windows"
```
andrew-d, line 33: I know that this is all code movement, but ... any reason that we're storing this? Now that everything is an accessor, we could replace `NotWindows()` with `return runtime.GOOS != "windows"` and make it easier for the compiler to optimize.

andrew-d (approved):

ipn/ipnauth/ipnauth.go
bradfitz: I wish I remembered. I could make up some history but I'm sure the original is funnier.

Probably some half-done & forgotten migration? Maybe this time'll get it done! 🙃

PR #6469, opened 2022-11-23
ipn/ipnserver: move more connection acceptance logic to LocalBackend

bradfitz:
Follow-up to #6467 and #6506.

LocalBackend knows the server-mode state, so move more auth checking
there, removing some bookkeeping from ipnserver.Server.

Updates #6417
Updates tailscale/corp#8051


bradfitz:
Test failure is related. (Maybe not real, but related and caused by the change and not a flake)

Looking.

maisem (approved):

ipn/ipnserver/server.go
```diff
@@ -100,12 +100,13 @@ type Server struct {
...
 	bs   *ipn.BackendServer
 
-	mu             sync.Mutex
-	serverModeUser *user.User                         // or nil if not in server mode
-	lastUserID     string                             // tracks last userid; on change, Reset state for paranoia
-	allClients     map[net.Conn]*ipnauth.ConnIdentity // HTTP or IPN
-	clients        map[net.Conn]bool                  // subset of allClients; only IPN protocol
-	disconnectSub  map[chan<- struct{}]struct{}       // keys are subscribers of disconnects
+	// mu guards the fields that follow.
+	// lock order: mu, then LocalBackend.mu
```
andrew-d, line 104: 🙇

ipn/ipnserver/server.go
```diff
@@ -614,23 +577,11 @@ func New(logf logger.Logf, logid string, store ipn.StateStore, eng wgengine.Engi
 
 	}
 
-	var serverModeUser *user.User
-	if uid := b.CurrentUser(); uid != "" {
-		u, err := ipnauth.LookupUserFromID(logf, uid)
```
andrew-d, line 619: I think this changes the behaviour since we no longer error on deleted users; specifically:
```go
		if !winutil.IsSIDValidPrincipal(uid) {
			return nil, err
		}
```

bradfitz: But it was just a log statement, not erring out entirely.

And now an equivalent log is in ipnlocal in the profile manager.

andrew-d: Rubber-🦆 a bit out loud here:

Previously:
* If `b.CurrentUser()` returns a non-empty string...
* We call `LookupUserFromID` with that UID...
* Which will fail and return an error if `IsSIDValidPrincipal` fails (i.e. if the user is deleted)...
* Which causes us to not set `serverModeUser` here

Now:
* We get the user from `serverModeUid := b.pm.CurrentUser()` in `CheckIPNConnectionAllowed`...
* And then check the UID strings directly
* In `profileManager`, the `pm.currentUserID` field is set by `SetCurrentUser`
* I don't see a call to either `IsSIDValidPrincipal` or `LookupUserFromID` (which would call that) anywhere in the call tree

So, assuming I'm reading this all right... I think that if a user creates a profile, then is deleted, when we load that profile later we'll end up in a situation where we set the `serverModeUid` to that deleted user and the connection check will fail, and never get out of that situation because we don't check for whether the user exists any more.

(please double-check my work here!)

bradfitz: Sounds plausible, but at the same time I'm not sure we've ever considered or had any intentional design for handling deleted users. The admin could always stop Tailscale, nuke the state, and start over. Should the behavior be that starting up in server mode with a deleted user causes the first user to connect to win? The first _admin_ user to connect? Server mode to just be disabled? I dunno. I haven't thought about it. You, @maisem? For now I'm inclined to consider deleted users a corner case and let admins clean it up if they do that.

andrew-d: This check was added in d7962e3bc and refers to https://github.com/tailscale/tailscale/issues/869 FWIW

maisem: That does sound like a regression. We should probably move the check to profileManager when it first starts up. Can happen in a follow on though, I don't think it's critical. As Brad mentioned it's an untested edge case

andrew-d: I'm happy to do this in a follow-up, or not at all if we decide that 👍

bradfitz:
@maisem, I rebased this PR after your latest change. Thoughts on continuing to move logic like this? It's almost not worth it at this point, but I'd like all the logic in one place and not split between two packages.

maisem (changes requested):
LGTM, I'm going to send another PR to remove `b.inServerMode`

maisem (approved):

PR #6471, opened 2022-11-23
ipn/ipnlocal: restrict peerAPI endpoints for UnsignedPeerAPIOnly nodes

twitchyliquid64:
Probably not a great idea if such nodes can hit things like taildrop. An allowlist we can add to seems appropriate.

Am I holding this right / Is `h.peerNode` going to be populated in the way I'm assuming? Does this fit with your plans for the ipnauth package?

### DIFF
--- a/ipn/ipnlocal/peerapi.go
+++ b/ipn/ipnlocal/peerapi.go
@@ -620,6 +620,17 @@ func (h *peerAPIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("X-Frame-Options", "DENY")
 		w.Header().Set("X-Content-Type-Options", "nosniff")
 	}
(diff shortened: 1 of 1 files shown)


bradfitz (changes requested):
No, peer caps are what peerapi allows. This bool isn't named IngressOnly.

twitchyliquid64:
The outcome I want is that we minimize the attack surface for unsigned nodes as much as possible. Control can set the file-sharing capability and set `UnsignedPeerAPIOnly`, which would mean that taildrop is driveable by a compromised control even if the tailnet is locked. Similarly with `/env`.

bradfitz:
Blocking Taildrop & debug SGTM. Something like https://github.com/tailscale/tailscale/pull/6475 ?

twitchyliquid64:
Lol exactly XD

PR #6473, opened 2022-11-23
cmd/stunc: add command for making debug stun requests

raggi:


cmd/stunc/stunc.go
```diff
@@ -0,0 +1,53 @@
+package main
```
danderson, line 6: Missing license header, missing package doc

danderson (approved):

PR #6475, opened 2022-11-23
ipn/ipnlocal: lock down unsigned peers more

bradfitz:
Apparently there's no tracking bug?

Updates tailscale/corp#7515 for ingress/funnel at least.


twitchyliquid64 (approved):
banger pr

bradfitz:
> banger pr

😂 never gets old

PR #6476, opened 2022-11-23
envknob: add GOOS func

bradfitz:
Centralize the fake GOOS stuff, start to use it more. To be used more in the future.

### DIFF
--- a/envknob/envknob.go
+++ b/envknob/envknob.go
@@ -277,10 +277,8 @@ func SSHPolicyFile() string { return String("TS_DEBUG_SSH_POLICY_FILE") }
 // SSHIgnoreTailnetPolicy is whether to ignore the Tailnet SSH policy for development.
 func SSHIgnoreTailnetPolicy() bool { return Bool("TS_DEBUG_SSH_IGNORE_TAILNET_POLICY") }
 
-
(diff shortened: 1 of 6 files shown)


envknob/envknob.go
```diff
@@ -429,3 +427,22 @@ func applyKeyValueEnv(r io.Reader) error {
...
+// This isn't universally used.
+func GOOS() string {
+	// TODO(bradfitz): add a build tag like "ts_not_in_tests" that
+	// makes this compile to just "return runtime.GOOS" in the
+	// hope that the compiler can then inline+deadcode this away
+	// and reduce binary size outside of tests. Verify that first.
+	if v := String("TS_DEBUG_FAKE_GOOS"); v != "" {
+		return v
+	}
+	return runtime.GOOS
```
andrew-d, line 447: I was testing this earlier and it does appear to deadcode out; check this out:
https://godbolt.org/z/jx1h6T678

andrew-d (approved):

maisem (approved):

envknob/envknob.go
bradfitz: Cool. Well, I can't check on mobile. :) I forget whether the rules are still different interpackage too. Probably not.

PR #6479, opened 2022-11-24
net/dnsfallback: update the burned-in bootstrap DNS map.

danderson:

### DIFF
--- a/net/dnsfallback/dns-fallback-servers.json
+++ b/net/dnsfallback/dns-fallback-servers.json
@@ -6,25 +6,25 @@
 			"RegionName": "r1",
 			"Nodes": [
 				{
-					"Name": "1c",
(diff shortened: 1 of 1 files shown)


bradfitz:
I forget the details but @DentonGentry was at one point holding off on doing this for something. Probably best to discuss on Slack.

DentonGentry:
Personally, I would prefer not to update this list of IPs. We never gave up the original DigitalOcean/vultr VMs, my recommendation would be:
- expect to downsize those but operate them for the foreseeable future as BootstrapDNS servers
- update the compiled-in bootstrapDNS list when we start using the /24 of IPv4 addresses which we own, perhaps just a couple of them via Anycast.

danderson:
okay.

PR #6489, opened 2022-11-24
ipn/ipnserver: only permit the pre-HTTP LocalAPI protocol on Windows

bradfitz:
Updates #6417


bradfitz:
(Just trying to incrementally deadcode/tighten things down for most platforms for now, in case the Windows change and the the bigger #6469 and #6454 don't make it in in time....)

andrew-d (approved):

PR #6491, opened 2022-11-24
ipn/ipnauth, util/winutil: add temporary LookupPseudoUser workaround …

dblohm7:
…to address os/user.LookupId errors on Windows

I added util/winutil/LookupPseudoUser, which essentially consists of the bits that I am in the process of adding to Go's standard library.

We check the provided SID for "S-1-5-x" where 17 <= x <= 20 (which are the known pseudo-users) and then manually populate a os/user.User struct with the correct information.

Fixes https://github.com/tailscale/tailscale/issues/869


ipn/ipnauth/ipnauth.go
```diff
@@ -105,6 +105,16 @@ func GetConnIdentity(logf logger.Logf, c net.Conn) (ci *ConnIdentity, err error)
 func LookupUserFromID(logf logger.Logf, uid string) (*user.User, error) {
 	u, err := user.LookupId(uid)
 	if err != nil && runtime.GOOS == "windows" && errors.Is(err, syscall.Errno(0x534)) {
+		// See if uid resolves as a pseudo-user. Temporary workaround until
+		// https://github.com/golang/go/issues/49509 resolves and ships.
+		if u, err := winutil.LookupPseudoUser(uid); err == nil {
+			return u, err
```
bradfitz, line 111: `return u, nil` would require less thinking for future readers

ipn/ipnauth/ipnauth.go
```diff
@@ -105,6 +105,16 @@ func GetConnIdentity(logf logger.Logf, c net.Conn) (ci *ConnIdentity, err error)
...
+		}
+
+		// TODO(aaron): With LookupPseudoUser in place, I don't expect us to reach
```
bradfitz, line 114: add a clientmetric or so? they're super cheap basically free when not hit.

util/winutil/winutil_windows.go
```diff
@@ -492,3 +493,62 @@ func OpenKeyWait(k registry.Key, path RegistryPath, access uint32) (registry.Key
...
+		Username: fmt.Sprintf(`%s\%s`, domain, username),
+		Name:     username,
+		HomeDir:  homeDir,
+	}
+	return result, nil
+}
+
+// findHomeDirInRegistry finds the user home path based on the uid.
+// This is borrowed from Go's std lib.
+func findHomeDirInRegistry(uid string) (dir string, e error) {
```
bradfitz, line 543: s/e/err/

`e` is only used in syscall packages when it's actually an errNo or other non-error interface type

util/winutil/winutil_windows_test.go
```diff
@@ -0,0 +1,27 @@
+package winutil
+
+import (
+	"testing"
+)
+
+const (
+	localSystemSID = "S-1-5-18"
+	networkSID = "S-1-5-2"
```
bradfitz, line 9: is this gofmt'ed?

util/winutil/winutil_windows_test.go
```diff
@@ -0,0 +1,27 @@
+package winutil
```
bradfitz, line 1: copyright header

bradfitz (approved):

PR #6502, opened 2022-11-25
cmd/tailscale/cli: Improve messaging when Funnel is unavailable.

shayne:
There are three specific requirements for Funnel to work:
1) They must accept an invite.
2) They must enable HTTPS.
3) The "funnel" node attribute must be appropriately set up in the ACLs.

### DIFF
--- a/cmd/tailscale/cli/serve.go
+++ b/cmd/tailscale/cli/serve.go
@@ -685,8 +685,8 @@ func (e *serveEnv) runServeFunnel(ctx context.Context, args []string) error {
 	if err != nil {
 		return fmt.Errorf("getting client status: %w", err)
 	}
-	if !slices.Contains(st.Self.Capabilities, tailcfg.NodeAttrFunnel) {
-		return errors.New("Funnel not available. See https://tailscale.com/s/no-funnel")
+	if err := checkHasAccess(st.Self.Capabilities); err != nil {
+		return err
 	}
 	dnsName := strings.TrimSuffix(st.Self.DNSName, ".")
 	hp := ipn.HostPort(dnsName + ":" + srvPortStr)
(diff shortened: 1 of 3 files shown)


cmd/tailscale/cli/serve.go
```diff
@@ -709,3 +709,19 @@ func (e *serveEnv) runServeFunnel(ctx context.Context, args []string) error {
...
 }
+
+// checkHasAccess checks three things: 1) an invite was used to join the
+// Funnel alpha; 2) HTTPS is enabled; 3) the node has the "funnel" attribute.
+// If any of these are false, an error is returned describing the problem.
+func checkHasAccess(capabilities []string) error {
+	if !slices.Contains(capabilities, "https://tailscale.com/cap/warn-funnel-no-invite") {
+		return errors.New("Funnel not available; an invite is required to join the alpha. See https://tailscale.com/kb/1223/tailscale-funnel/.")
+	}
+	if !slices.Contains(capabilities, "https://tailscale.com/cap/warn-funnel-no-https") {
```
bradfitz, line 720: not not. Just "if slices.Contains..." on both these /warn ones.

Want to move them into tailcfg constants while you're here so then control can depend on them later?

bradfitz (approved):

cmd/tailscale/cli/serve.go
shayne: Too much 🦃 🍗  Thanks! Will make the constants.

shayne: PTAL

cmd/tailscale/cli/serve.go
```diff
@@ -709,3 +709,19 @@ func (e *serveEnv) runServeFunnel(ctx context.Context, args []string) error {
 	}
 	return nil
 }
+
+// checkHasAccess checks three things: 1) an invite was used to join the
+// Funnel alpha; 2) HTTPS is enabled; 3) the node has the "funnel" attribute.
+// If any of these are false, an error is returned describing the problem.
+func checkHasAccess(capabilities []string) error {
```
bradfitz, line 716: rename capabilities to nodeAttrs ?

and describe it in the godoc for this func a bit

bradfitz (approved):

cmd/tailscale/cli/serve.go
bradfitz: but LGTM still otherwise

PR #6509, opened 2022-11-25
ipn/{ipnlocal,ipnserver}: add some comments

bradfitz:


ipn/ipnlocal/local.go
```diff
@@ -2012,6 +2012,17 @@ func (b *LocalBackend) State() ipn.State {
...
+//
+// On non-Windows platforms, this usually returns false (because people don't
+// set unattended mode on other platforms) and also isn't checked on other
+// platforms.
+//
+// TODO(bradfitz): rename to InWindowsUnattendedMode or something? Or make this
+// return true on Linux etc and always be called? It's kinda messy now.
 func (b *LocalBackend) InServerMode() bool {
 	b.mu.Lock()
 	defer b.mu.Unlock()
```
maisem, line 2028: I wonder if we can also drop `inServerMode` and rely on `b.pm.CurrentPrefs().ForceDaemon()`

maisem (approved):

ipn/ipnlocal/local.go
maisem: Looks simple enough.

bradfitz: hopefully!

PR #6506, opened 2022-11-25
ipn/ipnserver: remove Server.serverModeUser

maisem:

PR #6510, opened 2022-11-25
ipn/ipnlocal,ipnserver: rename {,Set}CurrentUser to {,Set}CurrentUserID

maisem:
Address comments from https://github.com/tailscale/tailscale/pull/6506#discussion_r1032454064

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -2039,7 +2039,7 @@ func (b *LocalBackend) InServerMode() bool {
 func (b *LocalBackend) CheckIPNConnectionAllowed(ci *ipnauth.ConnIdentity) error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	serverModeUid := b.pm.CurrentUser()
+	serverModeUid := b.pm.CurrentUserID()
 	if serverModeUid == "" {
 		// Either this platform isn't a "multi-user" platform or we're not yet
 		// running as one.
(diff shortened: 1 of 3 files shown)


bradfitz (approved):

ipn/ipnserver/server.go
```diff
@@ -304,8 +304,14 @@ func (s *Server) checkConnIdentityLocked(ci *ipnauth.ConnIdentity) error {
 			return inUseOtherUserError{fmt.Errorf("Tailscale already in use by %s, pid %d", active.User().Username, active.Pid())}
 		}
 	}
-	if cu := s.b.CurrentUser(); cu != "" && cu != ci.UserID() {
-		return inUseOtherUserError{fmt.Errorf("Tailscale already in use by %s", s.b.CurrentUser())}
+	if cu := s.b.CurrentUserID(); cu != "" && cu != ci.UserID() {
+		// TODO(maisem): cache this?
+		u, err := user.LookupId(cu)
```
andrew-d, line 309: Should we use the `LookupUserByID` in ipnauth here?

bradfitz: yeah probably

maisem: Done

andrew-d (approved):

ipn/ipnlocal/local.go
```diff
@@ -2053,7 +2053,12 @@ func (b *LocalBackend) CheckIPNConnectionAllowed(ci *ipnauth.ConnIdentity) error
 		return errors.New("empty user uid in connection identity")
 	}
 	if uid != serverModeUid {
-		return fmt.Errorf("Tailscale running in server mode (uid=%v); connection from %q not allowed", serverModeUid, uid)
+		u, err := ipnauth.LookupUserFromID(b.logf, serverModeUid)
+		if err != nil {
+			return fmt.Errorf("Tailscale running in server mode (uid=%q); connection from %q not allowed", serverModeUid, uid)
+		} else {
+			return fmt.Errorf("Tailscale running in server mode (user=%q); connection from %q not allowed", u.Username, uid)
```
bradfitz, line 2060: these are still kinda gross with uids

Maybe make a little closure here that return a best effort stringification of a uid using ipnauth.LookupUserFromID and then use it in one fmt.Errorf return path for both the serverModeUid and uid?

bradfitz (approved):
LGTM otherwise

ipn/ipnlocal/local.go
maisem: added a method to LocalBackend instead of a closure.

PR #6511, opened 2022-11-25
ipn/ipnlocal: drop LocalBackend.inServerMode

maisem:

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -157,7 +157,6 @@ type LocalBackend struct {
 	notify         func(ipn.Notify)
 	cc             controlclient.Client
 	ccAuto         *controlclient.Auto // if cc is of type *controlclient.Auto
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/local.go
```diff
@@ -2027,7 +2025,7 @@ func (b *LocalBackend) State() ipn.State {
 func (b *LocalBackend) InServerMode() bool {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	return b.inServerMode
+	return b.pm.CurrentPrefs().ForceDaemon()
```
bradfitz, line 2028: this is guaranteed to be a Valid view?

bradfitz (approved):

ipn/ipnlocal/local.go
maisem: yes, updated docs.

PR #6518, opened 2022-11-26
ipn/ipnserver: remove protoSwitchConn shenanigans; just use http.Server early

bradfitz:
Now that everything's just HTTP, there's no longer a need to have a header-sniffing net.Conn wraper that dispatches which route to take. Refactor to just use an http.Server earlier instead.

Updates #6417

### DIFF
--- a/ipn/ipnserver/proxyconnect.go
+++ b/ipn/ipnserver/proxyconnect.go
@@ -7,15 +7,11 @@
 package ipnserver
 
 import (
-	"bufio"
-	"context"
 	"io"
 	"net"
 	"net/http"
-	"time"
 
 	"tailscale.com/logpolicy"
-	"tailscale.com/types/logger"
 )
 
 // handleProxyConnectConn handles a CONNECT request to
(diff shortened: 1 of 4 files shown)


ipn/ipnserver/server.go
```diff
@@ -82,66 +81,68 @@ type Server struct {
...
-	c.SetReadDeadline(time.Now().Add(30 * time.Second))
-	br.Peek(len("GET / HTTP/1.1\r\n")) // reasonable sniff size to get HTTP method
-	c.SetReadDeadline(time.Time{})
-
-	// Handle logtail CONNECT requests early. (See docs on handleProxyConnectConn)
-	if bufferIsConnect(br) {
-		s.handleProxyConnectConn(ctx, br, c, logf)
+func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "CONNECT" {
+		s.handleProxyConnectConn(w, r)
```
maisem, line 102: add a comment about this only being used by the windows GUI. Probably also build tag it as such?

maisem (approved):

ipn/ipnserver/proxyconnect.go
```diff
@@ -27,40 +23,41 @@ import (
...
 	}
 	defer back.Close()
 
+	hj, ok := w.(http.Hijacker)
+	if !ok {
+		panic("[unexpected] not a Hijacker")
+	}
+	c, br, err := hj.Hijack()
+	if err != nil {
+		s.logf("CONNECT hijack: %v", err)
```
andrew-d, line 56: The Go standard library writes an error in this case: https://pkg.go.dev/net/http#example-Hijacker
```
http.Error(w, err.Error(), http.StatusInternalServerError)
```

ipn/ipnserver/server.go
```diff
@@ -262,40 +263,41 @@ func (s *Server) addConn(c net.Conn) (ci *ipnauth.ConnIdentity, err error) {
...
 
-	if remain == 0 && s.resetOnZero {
-		if s.b.InServerMode() {
-			s.logf("client disconnected; staying alive in server mode")
-		} else {
-			s.logf("client disconnected; stopping server")
-			s.b.ResetForClientDisconnect()
+	onDone = func() {
+		s.mu.Lock()
+		delete(s.activeReqs, req) // TODO: this is racy, but we don't care
```
andrew-d, line 286: Why don't we care?

ipn/ipnserver/server.go
```diff
@@ -429,26 +436,34 @@ func (s *Server) Run(ctx context.Context, ln net.Listener) error {
...
+			if err != nil {
+				return context.WithValue(ctx, connIdentityContextKey{}, err)
 			}
-			s.logf("ipnserver: Accept: %v", err)
-			bo.BackOff(ctx, err)
-			continue
+			return context.WithValue(ctx, connIdentityContextKey{}, ci)
+		},
+		// Localhost connections are cheap; so only do
+		// keep-alives for a short periodof time, as these
```
andrew-d, line 452:
```suggestion
		// keep-alives for a short period of time, as these
```

andrew-d (approved):

ipn/ipnserver/server.go
bradfitz: Hahaha I think that was copilot. It shouldn't be racy at all. Will delete the comment.

ipn/ipnserver/proxyconnect.go
bradfitz: done

ipn/ipnserver/server.go
bradfitz: It has a lot of docs on the definition of the func, but I'll add a little here too.

ipn/ipnserver/server.go
andrew-d: Hahaha, amazing 🤣

andrew-d (approved):

PR #6521, opened 2022-11-26
cmd/tailscale/cli: [serve] rework commands based on feedback

### LINKED ISSUE #6519: cmd/tailscale/cli: relax ordering requirement of flags for serve sub-command
... to improve the ergonomics of the `tailscale serve` sub-command,

[ffcli ](https://pkg.go.dev/github.com/peterbourgon/ff/v3/ffcli) enforces the ordering of flags as it relates to their sub-commands. Ideally we'd be able to do the following:
```console
$ tailscale serve / text "Hello, world!"
$ # *presses up on keyboard*
$ tailscale serve / text "Hello, world!" --remove
$ # or
$ tailscale serve / --remove
$ tailscale serve --remove / # currently
$ # or
$ tailscale serve tcp 32400
$ # *presses up on keyboard*
$ tailscale serve tcp 32400 --remove
$ tailscale serve --remove tcp 32400 # currently
``

### PR DESCRIPTION
**Pull request by shayne**
Updates to the `serve` and `funnel` subcommands syntax below:
```console
$ tailscale serve https:<port> <mount-point> <source> [off]
$ tailscale serve tcp:<port> tcp://localhost:<local-port> [off]
$ tailscale serve tls-terminated-tcp:<port> tcp://localhost:<local-port> [off]
$ tailscale serve status [--json]

$ tailscale funnel <serve-port> {on|off}
$ tailscale funnel status [--json]

EXAMPLES
  - To proxy requests to a web server at 127.0.0.1:3000:
    $ tailscale serve https:443 / http://127.0.0.1:3000

  - To serve a single file or a directory of files:
    $ tailscale serve https:443 / /home/alice/blog/index.html
    $ tailscale serve https:443 /images/ /home/alice/blog/images

  - To serve simple static text:
    $ tailscale serve https:443 / text:"Hello, world!"

  - To forward raw TCP packets to a local TCP server on port 22:
    $ tailscale serve tcp:2222 tcp://localhost:22

  - To forward raw, TLS-terminated TCP packets to a local TCP server on port 80:
    $ tailscale serve tls-terminated-tcp:443 tcp://localhost:80
```

Fixes: #6674

### DIFF
--- a/cmd/tailscale/cli/funnel.go
+++ b/cmd/tailscale/cli/funnel.go
@@ -1,0 +1,138 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package cli
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"net"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/peterbourgon/ff/v3/ffcli"
+	"tailscale.com/ipn"
+	"tailscale.com/util/mak"
+)
+
+var funnelCmd = newFunnelCommand(&serveEnv{lc: &localClient})
+
+// newFunnelCommand returns a new "funnel" subcommand using e as its environment.
+// The funnel subcommand is used to turn on/off the Funnel service.
+// Funnel is off by default.
+// Funnel allows you to publish a 'tailscale serve' server publicly, open to the
+// entire internet.
+// newFunnelCommand shares the same serveEnv as the "serve" subcommand. See
+// newServeCommand and serve.go for more details.
+func newFunnelCommand(e *serveEnv) *ffcli.Command {
+	return &ffcli.Command{
+		Name:      "funnel",
+		ShortHelp: "[ALPHA] turn Tailscale Funnel on or off",
+		ShortUsage: strings.TrimSpace(`
+funnel <serve-port> {on|off}
+  funnel status [--json]
+`),
+		LongHelp: strings.Join([]string{
+			"Funnel allows you to publish a 'tailscale serve'",
+			"server publicly, open to the entire internet.",
+			"",
+			"Turning off Funnel only turns off serving to the internet.",
+			"It does not affect serving to your tailnet.",
+		}, "\n"),
+		Exec:      e.runFunnel,
+		UsageFunc: usageFunc,
+		Subcommands: []*ffcli.Command{
+			{
+				Name:      "status",
+				Exec:      e.runServeStatus,
+				ShortHelp: "show current serve/funnel status",
+				FlagSet: e.newFlags("funnel-status", func(fs *flag.FlagSet) {
+					fs.BoolVar(&e.json, "json", false, "output JSON")
+				}),
+				UsageFunc: usageFunc,
+			},
+		},
+	}
+}
+
+// runFunnel is the entry point for the "tailscale funnel" subcommand and
+// manages turning on/off funnel. Funnel is off by default.
+//
+// Note: funnel is only supported on single DNS name for now. (2022-11-15)
+func (e *serveEnv) runFunnel(ctx context.Context, args []string) error {
+	if len(args) != 2 {
+		return flag.ErrHelp
+	}
+
+	var on bool
+	switch args[1] {
+	case "on", "off":
+		on = args[1] == "on"
+	default:
+		return flag.ErrHelp
+	}
+	sc, err := e.lc.GetServeConfig(ctx)
+	if err != nil {
+		return err
+	}
+	if sc == nil {
+		sc = new(ipn.ServeConfig)
+	}
+	st, err := e.getLocalClientStatus(ctx)
+	if err != nil {
+		return fmt.Errorf("getting client status: %w", err)
+	}
+
+	port64, err := strconv.ParseUint(args[0], 10, 16)
+	if err != nil {
+		return err
+	}
+	port := uint16(port64)
+
+	if err := ipn.CheckFunnelAccess(port, st.Self.Capabilities); err != nil {
+		return err
+	}
+	dnsName := strings.TrimSuffix(st.Self.DNSName, ".")
+	hp := ipn.HostPort(dnsName + ":" + strconv.Itoa(int(port)))
+	if on == sc.AllowFunnel[hp] {
+		printFunnelWarning(sc)
+		// Nothing to do.
+		return nil
+	}
+	if on {
+		mak.Set(&sc.AllowFunnel, hp, true)
+	} else {
+		delete(sc.AllowFunnel, hp)
+		// clear map mostly for testing
+		if len(sc.AllowFunnel) == 0 {
+			sc.AllowFunnel = nil
+		}
+	}
+	if err := e.lc.SetServeConfig(ctx, sc); err != nil {
+		return err
+	}
+	printFunnelWarning(sc)
+	return nil
+}
+
+// printFunnelWarning prints a warning if the Funnel is on but there is no serve
+// config for its host:port.
+func printFunnelWarning(sc *ipn.ServeConfig) {
+	var warn bool
+	for hp, a := range sc.AllowFunnel {
+		if !a {
+			continue
+		}
+		_, portStr, _ := net.SplitHostPort(string(hp))
+		p, _ := strconv.ParseUint(portStr, 10, 16)
+		if _, ok := sc.TCP[uint16(p)]; !ok {
+			warn = true
+			fmt.Fprintf(os.Stderr, "Warning: funnel=on for %s, but no serve config\n", hp)
+		}
+	}
+	if warn {
+		fmt.Fprintf(os.Stderr, "         run: `tailscale serve --help` to see how to configure handlers\n")
+	}
+}
--- a/cmd/tailscale/cli/serve.go
+++ b/cmd/tailscale/cli/serve.go
@@ -35,8 +35,11 @@ func newServeCommand(e *serveEnv) *ffcli.Command {
 		Name:      "serve",
 		ShortHelp: "[ALPHA] Serve from your Tailscale node",
 		ShortUsage: strings.TrimSpace(`
-  serve [flags] <mount-point> {proxy|path|text} <arg>
-  serve [flags] <sub-command> [sub-flags] <args>`),
+serve https:<port> <mount-point> <source> [off]
+  serve tcp:<port> tcp://localhost:<local-port> [off]
+  serve tls-terminated-tcp:<port> tcp://localhost:<local-port> [off]
+  serve status [--json]
+`),
 		LongHelp: strings.TrimSpace(`
 *** ALPHA; all of this is subject to change ***
 
@@ -45,66 +48,40 @@ content and local servers from your Tailscale node to
 your tailnet. 
 
 You can also choose to enable the Tailscale Funnel with:
-'tailscale serve funnel on'. Funnel allows you to publish
+'tailscale funnel on'. Funnel allows you to publish
 a 'tailscale serve' server publicly, open to the entire
 internet. See https://tailscale.com/funnel.
 
 EXAMPLES
   - To proxy requests to a web server at 127.0.0.1:3000:
-    $ tailscale serve / proxy 3000
+    $ tailscale serve https:443 / http://127.0.0.1:3000
+
+	Or, using the default port:
+	$ tailscale serve https / http://127.0.0.1:3000
 
   - To serve a single file or a directory of files:
-    $ tailscale serve / path /home/alice/blog/index.html
-    $ tailscale serve /images/ path /home/alice/blog/images
+    $ tailscale serve https / /home/alice/blog/index.html
+    $ tailscale serve https /images/ /home/alice/blog/images
 
   - To serve simple static text:
-    $ tailscale serve / text "Hello, world!"
+    $ tailscale serve https:8080 / text:"Hello, world!"
+
+  - To forward raw TCP packets to a local TCP server on port 5432:
+    $ tailscale serve tcp:2222 tcp://localhost:22
+
+  - To forward raw, TLS-terminated TCP packets to a local TCP server on port 80:
+    $ tailscale serve tls-terminated-tcp:443 tcp://localhost:80
 `),
-		Exec: e.runServe,
-		FlagSet: e.newFlags("serve", func(fs *flag.FlagSet) {
-			fs.BoolVar(&e.remove, "remove", false, "remove an existing serve config")
-			fs.UintVar(&e.servePort, "serve-port", 443, "port to serve on (443, 8443 or 10000)")
-		}),
+		Exec:      e.runServe,
 		UsageFunc: usageFunc,
 		Subcommands: []*ffcli.Command{
 			{
 				Name:      "status",
 				Exec:      e.runServeStatus,
-				ShortHelp: "show current serve status",
+				ShortHelp: "show current serve/funnel status",
 				FlagSet: e.newFlags("serve-status", func(fs *flag.FlagSet) {
 					fs.BoolVar(&e.json, "json", false, "output JSON")
 				}),
-				UsageFunc: usageFunc,
-			},
-			{
-				Name:      "tcp",
-				Exec:      e.runServeTCP,
-				ShortHelp: "add or remove a TCP port forward",
-				LongHelp: strings.Join([]string{
-					"EXAMPLES",
-					"  - Forward TLS over TCP to a local TCP server on port 5432:",
-					"    $ tailscale serve tcp 5432",
-					"",
-					"  - Forward raw, TLS-terminated TCP packets to a local TCP server on port 5432:",
-					"    $ tailscale serve tcp --terminate-tls 5432",
-				}, "\n"),
-				FlagSet: e.newFlags("serve-tcp", func(fs *flag.FlagSet) {
-					fs.BoolVar(&e.terminateTLS, "terminate-tls", false, "terminate TLS before forwarding TCP connection")
-				}),
-				UsageFunc: usageFunc,
-			},
-			{
-				Name:       "funnel",
-				Exec:       e.runServeFunnel,
-				ShortUsage: "funnel [flags] {on|off}",
-				ShortHelp:  "turn Tailscale Funnel on or off",
-				LongHelp: strings.Join([]string{
-					"Funnel allows you to publish a 'tailscale serve'",
-					"server publicly, open to the entire internet.",
-					"",
-					"Turning off Funnel only turns off serving to the internet.",
-					"It does not affect serving to your tailnet.",
-				}, "\n"),
 				UsageFunc: usageFunc,
 			},
 		},
@@ -143,10 +120,7 @@ type localServeClient interface {
 // It also contains the flags, as registered with newServeCommand.
 type serveEnv struct {
 	// flags
-	servePort    uint // Port to serve on. Defaults to 443.
-	terminateTLS bool
-	remove       bool // remove a serve config
-	json         bool // output JSON (status only for now)
+	json bool // output JSON (status only for now)
 
 	lc localServeClient // localClient interface, specific to serve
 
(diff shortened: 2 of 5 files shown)


bradfitz (commented):
Sorry but I think this is too complication, both in code and for users: now only some subcommands have different flag order rules. I want them all to be consistent.

How about we use type "-" or "delete" something to delete instead? Then it's not a flag.

Alternatively we fix it for all subcommands in ffcli or whatever itself

shayne:
I want to move forward in improving our CLI. Happy to fix all commands, too. 😺 

I admit the code required to make this work isn't ideal. However, ffcli doesn't provide a hook to make this work without wrapping functionality as I did with `preExec`. With that, I don't think the code is overly complicated.

I think it's worth splitting things out regarding user confusion: a) how ffcli works now and b) inconsistency with other commands.

Regarding how ffcli works now, I think it's awkward that flags must be listed before arguments or sub-commands. Take, for example, this TCP serve command:

```console
$ tailscale serve tcp --terminate-tls 5432
$ # versus 
$ tailscale serve tcp 5442 --terminate-tls
```

The first example is correct, and the second fails with an undefined flag error. IMO both should work (why not?), and my preference is the second as I'm issuing the command `serve tcp 5432` and then tacking on `--terminate-tls`. 

Using the above example in considering removal:
```console
$ tailscale serve --remove tcp 5432
$ # versus 
$ tailscale serve tcp 5432 --remove
$ # versus 
$ tailscale serve delete tcp 5432
```

Again, the first example is correct. My thinking (and this PR) allows the first or second to work. It doesn't require documenting the flexibility; we could advertise: `tailscale serve tcp <port> [flags]`. I'm not a fan of another sub-command, if only because it requires rewriting the command vs. tacking on a `--remove` or `--rm` to the command you used to create the serve.

I do agree we should be consistent with the other commands / sub-commands. If you agree, I'd like to rewrite this in a non-serve specific but implement it only for the serve command. The approach will probably involve adding a layer between our app and ffcli, which would benefit other customizations that require spelunking (like UsageFunc). After review, I can post PRs for other commands.

In looking through other commands, there isn't a behavior change. Moving forward with relaxing flag ordering would allow documented and prior usage as before. The only material difference would be a flag could be successfully added to the end of the command or sub-command.

Only `tailscale file` has more than one sub-command with flags. Additionally, network lock has a subcommand, but only one has a flag (`log`).

bradfitz:
The problem with flexible flag ordering in general is that it introduces parsing ambiguities. Consider a "foo" and "bar" command that both have a "--flag" flag. 

```
$ tailscale foo bar --flag
$ tailscale foo --flag bar
```

That latter case is obviously a flag for `foo`, but with flexible ordering, what about the first case? Is it a flag for bar or a flag for foo that was listed after the positional arguments?

If we had a flag parsing library will well-defined rules that didn't have ambiguous cases I would object less, but having no defined rules and doing ad-hoc parsing for some commands isn't very appealing.

Doing deletion without flags seems better. We have "serve funnel {on,off}" after all. We don't use a flag there.

Also, "tailscale serve tcp 8080" is kinda weird. (Serve 8080, okay, but how?) The implicit "oh, to localhost:8080" is a reasonable default I suppose, but a bit too magic for my tastes. I'd prefer we could write:

```
     tailscale serve tcp 8080             # implicit target localhost:8080
     tailscale serve tcp 8080 localhost   # implicit target 8080
     tailscale serve tcp 8080 localhost:8080   # explicitly specifying the default target
     tailscale serve tcp 8080 10.0.0.2:8012    # explicitly mapped elsewhere
     tailscale serve tcp 8080 off        # delete it, like "serve funnel off"
```

shayne:
I recognize that flexible flag ordering introduces ambiguities. However, I'm not convinced it's worse than forcing positional flags when the serve CLI is more fluent and less directive. Take `serve tcp 8080` and `serve / text "hello"` as examples. They're more expressions than structured commands/sub-command in feel. Inserting our flags really breaks the readability and cohesiveness: `serve --serve-port=10000 tcp --terminate-tls 8080`.
<br/>

> Doing deletion without flags seems better. We have "serve funnel {on,off}" after all. We don't use a flag there.

Totally, you're right and I think this is a good way forward. I was confused a bit by your examples, like which `8080` was the local-port vs which `8080` was the serve-port.

I'll list a couple of examples below and let me know of the examples, which you prefer.

A)
```console
$ tailscale serve --help

tailscale serve <mount-point> {proxy|path|text} <arg> [<off> | --serve-port=(443|8443|10000)]
tailscale serve tcp <local-port> [<off> | --terminat-tls | --serve-port=(443|8443|10000)]

$ tailscale serve tcp 8080 # forwards to localhost:8080, serve port 443
$ tailscale serve tcp 8080 off

$ tailscale serve tcp 5432 --terminate-tls --serve-port=10000
$ tailscale serve tcp 5432 off --serve-port=10000 # not loving the flag w/ off

$ tailscale serve / proxy 3000 # proxies to localhost:3000, serve port 443
$ tailscale serve / off

$ tailscale serve / proxy localhost:8123 # proxies to localhost:8123, serve port 443
$ tailscale serve / off

$ tailscale serve / proxy localhost:8123 --serve-port=10000
$ tailscale serve / --serve-port=10000 off # flag doesn't feel good here either
```

B)
```console

$ tailscale serve --help

tailscale serve <mount-point> {proxy|path|text} <arg> [<serve-port> | <off>]
tailscale serve tcp <local-port> [<serve-port> | <off> | --terminate-tls]

$ tailscale serve tcp 8080 # forwards to localhost:8080, serve port 443
$ tailscale serve tcp 8080 off

$ tailscale serve tcp 5432 10000 --terminate-tls # forwards to localhost:5432, serving on tcp://100.x.y.z:10000
$ tailscale serve tcp 5432 10000 off

$ tailscale serve tcp localhost:4430 8443 --terminate-tls # forwards to localhost:4430 serving on tcp://100.x.y.z:8443
$ tailscale serve tcp localhost:4430 8443 off # or
$ tailscale serve tcp 4430 8443 off

$ tailscale serve / proxy 3000 # proxies to localhost:3000, serving at https://node.fun.ts.net
$ tailscale serve / off

$ tailscale serve / proxy localhost:8123 10000 # proxies to localhost:8123, serving at https://node.fun.ts.net:10000
$ tailscale serve / 10000 off

$ tailscale serve / path /var/www/ 8443 # serving at https://node.fun.ts.net:8443
$ tailscale serve / 8443 off
```

C)
```console

$ tailscale serve --help

tailscale serve <serve-port> <mount-point> {proxy|path|text} <arg> [<off>]
tailscale serve tcp <serve-port> <local-port> [<off> | --terminate-tls]

$ tailscale serve tcp 8080 # error: no implicit mapping of serve-port to local-port

$ tailscale serve tcp 443 8080 # forwards to localhost:8080, serving on tcp://100.x.y.z:443
$ tailscale serve tcp 443 8080 off

$ tailscale serve tcp 10000 5432 --terminate-tls # forwards to localhost:5432, serving on tcp://100.x.y.z:10000
$ tailscale serve tcp 10000 5432 off

$ tailscale serve tcp 8443 localhost:4430 --terminate-tls # forwards to localhost:4430 serving on tcp://100.x.y.z:8443
$ tailscale serve tcp 8443 localhost:4430 off # or
$ tailscale serve tcp 8443 4430 off

$ tailscale serve / proxy 3000 # error: no implicit serve-port

$ tailscale serve 443 / proxy 3000 # proxies to localhost:3000, serving at https://node.fun.ts.net
$ tailscale serve 443 / off

$ tailscale serve 10000 / proxy localhost:8123 #  serving at https://node.fun.ts.net:10000
$ tailscale serve 10000 / off

$ tailscale serve 8443 / path /var/www/ # serving at https://node.fun.ts.net:8443
$ tailscale serve 8443 / off
```

creachadair:
In the past I have had some success using a different "option" syntax in place of standard flags, for example "+key value", when I've needed to express this kind of structured argument list. It avoids tripping up pre-existing expectations about flags and signals to the reader that they're doing something different.

Perhaps that's not the right choice for this case, I don't know — but it's worth considering an option besides reworking flags if this case is indeed very different from other subcommand regimes.

bradfitz:
(c) is closest.

But I think all the integer literals by themselves are doing more harm than good. Explicit `localhost:8080` and maybe even explicit `ts` as in `ts:443` would be more clear what's happening.

bradfitz:
I'd also like to avoid flags here:

```
tailscale serve tcp 8443 localhost:4430 --terminate-tls
```

I wonder if like @creachadair says we could invent some syntax for that, like:

```
tailscale serve tcp ts:8443 localhost:8080,addtls
tailscale serve tcp ts:8443 localhost:8080+tls
tailscale serve tcp ts:8443+tls localhost:8080
tailscale serve tcp ts:8443 tls+localhost:8080
```

peterbourgon:
I'm happy to consider changes to ffcli that would help, here.

shayne:
@peterbourgon I appreciate your input and willingness to consider changes to ffcli. I take your points regarding arguments vs. flags.

A few things pop out in writing out sample commands: the command length is quite a bit longer than I'd like and requires more syntax than I was hoping for. There's also the issue of requiring the user to specify their hostname for the `DST` parameter. It's not ideal as it's randomly generated by default (example: `my-node.tailf310e.ts.net`).

The difference in style, flags vs. parameters/arguments, is more a matter of preference to me. A successful CLI for Funnel/serve is approachable, intuitive, and robust. I want to optimize for usability and balance brevity.

Your suggestions, @creachadair's, and @bradfitz's all contributed to another take. So here are my thoughts and a ~complete example. I'd appreciate any feedback, but I plan on getting a head start tomorrow.

 - Scheme explicitly includes the protocol, and for TCP, `tcp+tls://` indicates TLS packets over TCP, while `tcp://` implies terminated TLS (to be explicitly called out in CLI extended help)
 - Use `off` as a parameter. For HTTPS, turn off all web handlers or a specific mount point.
 - Demote text to requiring a `text:` pseudo protocol. It's mostly for debugging, so don't design around it. Allows clean `/path/like` parameters for file serving and `http,https,https+insecure` for protocols for the target.
 - Make `funnel` its own command. `tailscale funnel` vs. `tailscale serve funnel`. `tailscale status` is the same impl. as `tailscale serve status`.

D)

```shell

$ tailscale serve --help

tailscale serve https:<serve-port> <mount-point> <target> [off]
tailscale serve tcp+tls:<serve-port> tcp[+tls]://localhost:<local-port> [off]
tailscale serve status

$ tailscale serve tcp+tls:443 tcp+tls://127.0.0.1:8080 # forwards to localhost:8080, serving over tcp://100.x.y.z:443 (TLS)
$ tailscale serve tcp+tls:443 off

$ tailscale serve tcp+tls:10000 tcp://localhost:5432 # forwards to localhost:5432, serving on tcp://100.x.y.z:10000 (TLS terminated)
$ tailscale serve tcp+tls:10000 off

$ tailscale serve tcp+tls:8443 tcp://localhost:4430 # forwards to localhost:4430 serving on tcp://100.x.y.z:8443 (TLS terminated)
$ tailscale serve tcp+tls:8443 tcp://localhost:4430 off

$ tailscale serve https:443 / http://localhost:3000 # proxies to localhost:3000, serving at https://node.fun.ts.net
$ tailscale serve https:443 / off

$ tailscale serve https:10000 / http://localhost:8123 # proxies to localhost:8123, serving at https://node.fun.ts.net:10000
$ tailscale serve https:10000 / off

$ tailscale serve https:8443 / /var/www/ # serving at https://node.fun.ts.net:8443
$ tailscale serve https:8443 / off

$ tailscale serve https:443 / text:"Hello, world!" # static text, mostly used for debug
$ tailscale serve https:443 / off

$ tailscale serve https:443 off # turns off all https:443 serves

# move funnel outside of serve

$ tailscale funnel --help

tailscale funnel [--serve-port=(443|8443|10000)] {on|off}
tailscale funnel status

$ tailscale funnel on # default 443
$ tailscale funnel --serve-port=8443 on
$ tailscale funnel --serve-port=8443 off
```

shayne:
Implemented, and I welcome all feedback!

cmd/tailscale/cli/funnel.go
```diff
@@ -0,0 +1,166 @@
...
+		return err
+	}
+	port := uint16(port64)
+	switch port {
+	case 443, 8443, 10000:
+		// ok
+	default:
+		fmt.Fprintf(os.Stderr, "serve-port %d is invalid; must be 443, 8443 or 10000\n\n", port)
+		return flag.ErrHelp
+	}
```
bradfitz, line 83: please don't hard-code this in the client. We want to be able to change it server-side.

shayne: That makes sense. Would you happen to have a recommendation for how we deliver the ports to the client? We don't verify the `serveConfig` beyond `tailscaled`. Is there an existing mechanism? Map response?

shayne:
tailscale/corp#9761 adds nodeAttr to remove ports being hard-coded

cmd/tailscale/cli/funnel.go
shayne: Done. cc/ @maisem

cmd/tailscale/cli/serve.go
```diff
@@ -333,12 +331,31 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
...
-	srvPort, err := e.validateServePort()
-	if err != nil {
-		return err
+func isProxyTarget(source string) bool {
+	if strings.HasPrefix(source, "http") && strings.Index(source, "://") > 0 {
+		return true
+	}
+	// support "localhost:3000", for example
+	_, portStr, _ := strings.Cut(source, ":")
+	if allNumeric(portStr) && strings.HasSuffix(source, ":"+portStr) {
```
maisem, line 340: `strings.HasSuffix(source, ":"+portStr)` is going to be always true based on `allNumeric(portStr)`

maybe
```
_, portStr, ok := strings.Cut(source, ":")
if ok && allNumeric(portStr) {
```

cmd/tailscale/cli/serve.go
```diff
@@ -333,12 +331,31 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
...
+		return true
+	}
+	return false
+}
+
+func allNumeric(s string) bool {
```
maisem, line 346: add a comment

```
// allNumeric reports whether s only comprises of digits
// and has at least one digit.
```

cmd/tailscale/cli/serve.go
```diff
@@ -426,16 +441,111 @@ func expandProxyTarget(target string) (string, error) {
...
+	}
+
+	switch host {
+	case "localhost", "127.0.0.1":
+		// ok
+	default:
+		fmt.Fprintf(os.Stderr, "error: invalid TCP source %q\n", dest)
+		fmt.Fprint(os.Stderr, "must be one of: localhost or 127.0.0.1\n\n", dest)
+		return flag.ErrHelp
+	}
```
maisem, line 481: why?

cmd/tailscale/cli/funnel.go
```diff
@@ -0,0 +1,160 @@
...
+// The funnel subcommand is used to turn on/off the Funnel service.
+// Funnel is off by default.
+// Funnel allows you to publish a 'tailscale serve' server publicly, open to the
+// entire internet.
+// newFunnelCommand shares the same serveEnv as the "serve" subcommand. See
+// newServeCommand and serve.go for more details.
+func newFunnelCommand(e *serveEnv) *ffcli.Command {
+	return &ffcli.Command{
+		Name:      "funnel",
+		ShortHelp: "turn Tailscale Funnel on or off",
```
maisem, line 35: [ALPHA]

cmd/tailscale/cli/serve.go
shayne: To avoid network topology issues where reachability, return path, etc... This has been a limitation since the initial release.

cmd/tailscale/cli/serve.go
```diff
@@ -333,12 +331,31 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
...
 }
 
-func (e *serveEnv) handleWebServeRemove(ctx context.Context, mount string) error {
-	srvPort, err := e.validateServePort()
-	if err != nil {
-		return err
+func isProxyTarget(source string) bool {
+	if strings.HasPrefix(source, "http") && strings.Index(source, "://") > 0 {
+		return true
+	}
```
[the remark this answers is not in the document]
shayne: We support `case "http", "https", "https+insecure":`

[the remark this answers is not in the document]
maisem: ah, then lets check explitiy for those

```
for _, p := range []string{"http://", "https://", "https+insecure://"} {
  if strings.HasPrefix(source, p) {
    return true
  }
}
```

cmd/tailscale/cli/serve.go
maisem: I was hoping we would remove this in Beta :P 
We were discussing using `serve` in the k8s-operator/containerboot earlier today. This is one of the reasons we can not today.

Can happen later.

cmd/tailscale/cli/funnel.go
```diff
@@ -0,0 +1,160 @@
...
+		return errors.New("Funnel not available; an invite is required to join the alpha. See https://tailscale.com/kb/1223/tailscale-funnel/.")
+	}
+	if slices.Contains(nodeAttrs, tailcfg.CapabilityWarnFunnelNoHTTPS) {
+		return errors.New("Funnel not available; HTTPS must be enabled. See https://tailscale.com/kb/1153/enabling-https/.")
+	}
+	if !slices.Contains(nodeAttrs, tailcfg.NodeAttrFunnel) {
+		return errors.New("Funnel not available; \"funnel\" node attribute not set. See https://tailscale.com/kb/1223/tailscale-funnel/.")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
shayne: ✅

cmd/tailscale/cli/serve.go
```diff
@@ -35,78 +35,52 @@ func newServeCommand(e *serveEnv) *ffcli.Command {
...
-  serve [flags] <mount-point> {proxy|path|text} <arg>
-  serve [flags] <sub-command> [sub-flags] <args>`),
+serve https:<port> <mount-point> <source> [off]
+  serve tcp:<port> tcp://localhost:<local-port> [off]
+  serve tls-terminated-tcp:<port> tcp://localhost:<local-port> [off]
+  serve status [--json]
+`),
 		LongHelp: strings.TrimSpace(`
-*** ALPHA; all of this is subject to change ***
+*** BETA; all of this is subject to change ***
```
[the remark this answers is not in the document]
shayne: ✅

cmd/tailscale/cli/serve.go
```diff
@@ -223,39 +185,89 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
...
+	if !found {
 		return flag.ErrHelp
 	}
 
-	srvPort, err := e.validateServePort()
-	if err != nil {
-		return err
+	turnOff := "off" == args[len(args)-1]
+
+	if len(args) < 2 || srcType == "https" && len(args) < 3 && !turnOff {
```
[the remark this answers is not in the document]
shayne: ✅

cmd/tailscale/cli/serve.go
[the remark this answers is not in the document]
shayne: I hardcoded the three, if hat's ok. PTAL

cmd/tailscale/cli/serve.go
shayne: ✅

cmd/tailscale/cli/serve.go
shayne: ✅

cmd/tailscale/cli/funnel.go
shayne: ✅

cmd/tailscale/cli/serve.go
```diff
@@ -223,39 +188,94 @@ func (e *serveEnv) runServe(ctx context.Context, args []string) error {
...
 
-	srvPort, err := e.validateServePort()
-	if err != nil {
-		return err
+	srcType, srcPortStr, found := strings.Cut(args[0], ":")
+	if !found {
+		if srcType == "https" && srcPortStr == "" {
+			// Default https port to 443.
+			srcPortStr = "443"
+		} else {
```
shayne, line 204: I'd like to get this convenience in. I think 1) it's the common case, and 2) 443 isn't commonly written as a port since it's a default.

cmd/tailscale/cli/serve.go
```diff
@@ -45,68 +48,42 @@ content and local servers from your Tailscale node to
...
 You can also choose to enable the Tailscale Funnel with:
-'tailscale serve funnel on'. Funnel allows you to publish
+'tailscale funnel on'. Funnel allows you to publish
 a 'tailscale serve' server publicly, open to the entire
 internet. See https://tailscale.com/funnel.
 
 EXAMPLES
   - To proxy requests to a web server at 127.0.0.1:3000:
-    $ tailscale serve / proxy 3000
+    $ tailscale serve https:443 / http://127.0.0.1:3000
```
maisem, line 57: add an example that does only `https` and only `http` without the ports

cmd/tailscale/cli/serve.go
maisem: lets do `http` too without the port

cmd/tailscale/cli/serve.go
shayne: Discussed on Slack and decided we'll do any HTTP in a follow up

cmd/tailscale/cli/serve.go
maisem: doing later.

maisem (approved):

Issue #6526, opened 2022-11-26
cmd/tailscale: add custom DERP checker


bradfitz:
People often set up custom DERP servers without following all the instructions. This results in a large support load, especially proportional to the benefit custom DERP provides people in the first place. (Too often people just want to do it because they can, not because they strictly need it).

To reduce support load, we should add a debug tool to run a bunch of diagnostic checks against their custom DERP setup to tell them what they did wrong.

Related: #6478, etc

/cc @DentonGentry

PR #6527, opened 2022-11-26
cmd/tailscale: add start of "debug derp" subcommand

bradfitz:
Updates #6526


ipn/localapi/debugderp.go
```diff
@@ -0,0 +1,81 @@
...
+	//   try with LocalBackend's node key instead.
+	// * if they have more then one node, try to relay a packet between them
+	//   and see if it works (like cmd/derpprobe). But if server authz is on,
+	//   we won't be able to, so just warn. Say to turn that off, try again,
+	//   then turn it back on. TODO(bradfitz): maybe add a debug frame to DERP
+	//   protocol to say how many peers it's meshed with. Should match count
+	//   in DERPRegion. Or maybe even list all their server pub keys that it's peered
+	//   with.
+	// * try STUN queries
+	// * warn about IPv6 only
```
DentonGentry, line 79: Recommend:
```
* If their certificate is bad, either expired or just wrongly
  issued in the first place, tell them specifically that the
  cert is bad not just that the connection failed.
* If /generate_204 on port 80 cannot be reached, warn
  that they won't get captive portal connection and
  should allow port 80.
```

and maybe:
```
* If they have exactly one DERP region because they
  removed all of Tailscale's DERPs, warn that they have
  a SPOF that will hamper even direct connections from
  working.
```
except I think this is a common practice for Headscale users. If so, we should state somewhere that it is ok to keep Tailscale's DERPs in the map for Headscale users.

DentonGentry (approved):

ipn/localapi/debugderp.go
bradfitz: Added

PR #6528, opened 2022-11-27
cmd/tailscale: de-punycode hostnames in status display

bradfitz:
Still show original, but show de-punycode version in parens, similar to how we show DNS-less hostnames.


bradfitz:
We could? Want to file a bug with your notes?

PR #6531, opened 2022-11-27
ipn/ipnauth: make new WindowsUserID type to consolidate docs

bradfitz:
The "userID is empty everywhere but Windows" docs on lots of places but not everywhere while using just a string type was getting confusing. This makes a new type to wrap up those rules, however weird/historical they might be.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -2076,12 +2076,12 @@ func (b *LocalBackend) CheckIPNConnectionAllowed(ci *ipnauth.ConnIdentity) error
 	if !b.pm.CurrentPrefs().ForceDaemon() {
 		return nil
 	}
-	uid := ci.UserID()
+	uid := ci.WindowsUserID()
 	if uid == "" {
 		return errors.New("empty user uid in connection identity")
 	}
 	if uid != serverModeUid {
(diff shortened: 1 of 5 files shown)


ipn/ipnlocal/local.go
```diff
@@ -2257,13 +2257,13 @@ func (b *LocalBackend) shouldUploadServices() bool {
...
-func (b *LocalBackend) SetCurrentUserID(uid string) {
+func (b *LocalBackend) SetCurrentUserID(uid ipnauth.WindowsUserID) {
 	b.mu.Lock()
-	if b.pm.CurrentUserID() == uid {
+	if b.pm.CurrentUserID() == string(uid) {
 		b.mu.Unlock()
 		return
 	}
-	if err := b.pm.SetCurrentUserID(uid); err != nil {
+	if err := b.pm.SetCurrentUserID(string(uid)); err != nil {
```
maisem, line 2266: Make this also take WindowsUserID?

maisem (approved):

ipn/ipnlocal/local.go
bradfitz: Done. It ended up having to move the type to ipn instead of ipnauth but that seems fine. Otherwise `ipn` starts growing weird deps.

Issue #6522, opened 2022-11-26
ipn/ipnserver: remove Windows special cases, serve status immediately for GUI


bradfitz:
All platforms except Windows use `ipnserver.New` followed by `ipnserver.(*Server).Run`.

But `ipnserver.New` requires a `wgengine.Engine` and creating one on Windows can fail and required (actually: still requires) retries, so on Windows we use `ipnserver.Run` which takes a `getEngine func() (wgengine.Engine, *netstack.Impl, error)` argument instead, and then we retry the `getEngine` call as needed on each connection. That means the Windows GUI's reconnect loop (with backoff) loop eventually maybe gets a good one, and then we stick with using that `wgengine.Engine` (via the also-Windows-only `getEngineUntilItWorksWrapper` code).

It's all pretty hacky, and I don't like one platform being special-cased.

Even with the hack, the ipnserver still doesn't start up right away and the GUI says something like "Tailscale service not running", which isn't true. It's running but hanging to create a tun device.

Instead, we should start up the IPN HTTP server immediately (on all platforms) and create the engine async (on all platforms), and only block `/localapi/` HTTP request until the engine is created. Then we can add an always-serving ipnserver HTTP handler at like `/status` that returns a status enum + message for a Status of {Starting,Ready,PermanentlyFailed} and the Windows GUI can use that. Or maybe we continue to serve the WatchIPNBus handler even without an engine and serve those status messages over ipn.Notify.

In any case, this is a tracking bug to both:

1) unify the Windows and non-Windows ipnserver start-up paths,
2) make the Windows GUI not get blocked by wgengine.Engine creation and have it show what's actually happening. It shouldn't say the Tailscale Service is dead if it's not.

We should do all this even if we make the Windows wintun start-up consistently work, because even when it works, it can be slow (5-10 seconds) when the wintun driver hasn't even been installed and we want the post-installer first time experience of starting the Tailscale Windows GUI to not have "Service not running" errors for 5-10 seconds. But even if that were made fast, we still want to get rid of the `ipnserver.Run` entrypoint and have Windows use the same API has the other platforms.

/cc @andrew-d @maisem @dblohm7

PR #6535, opened 2022-11-27
ipn/ipnserver: change Server to let LocalBackend be supplied async

bradfitz:
This is step 1 of de-special-casing of Windows and letting the LocalAPI HTTP server start serving immediately, even while the rest of the world (notably the Engine and its TUN device) are being created, which can take a few to dozens of seconds on Windows.

With this change, the ipnserver.New function changes to not take an Engine and to return immediately, not returning an error, and let its Run run immediately. If its ServeHTTP is called when it doesn't yet have a LocalBackend, it returns an error. A TODO in there shows where a future handler will serve status before an engine is available.

Future changes will:

* delete a bunch of tailscaled_windows.go code and use this new API
* add the ipnserver.Server ServerHTTP handler to await the engine being available
* use that handler in the Windows GUI client

Updates #6522


ipn/ipnserver/server.go
```diff
@@ -11,69 +11,28 @@ import (
...
-
-	// LoginFlags specifies the LoginFlags to pass to the client.
-	LoginFlags controlclient.LoginFlags
-}
-
 // Server is an IPN backend and its set of 0 or more active localhost
 // TCP or unix socket connections talking to that backend.
 type Server struct {
-	b            *ipnlocal.LocalBackend
+	lb           syncs.AtomicValue[*ipnlocal.LocalBackend]
```
maisem, line 35: should this be a `atomic.Pointer`?

ipn/ipnserver/server.go
```diff
@@ -304,43 +282,53 @@ func (s *Server) addActiveHTTPRequest(req *http.Request, ci *ipnauth.ConnIdentit
...
+	if _, ok := s.lb.LoadOk(); ok {
+		panic("already set")
+	}
+	// Yes, this LoadOk+Store is technically racy. But
+	// syncs.AtomicValue.CompareAndSwap doesn't work with with a nil source, as
+	// the zero value of the underlying atomic.Value is holding a nil interface
+	// and the typed CompareAndSwap wrapper uses a typed nil as the swap source.
+	// But this is fine being best effort. It's only meant to catch obvious
+	// start-up API mistakes. It's not called at runtime.
+	s.lb.Store(lb)
```
maisem, line 313: then you can simplify this bit too.

maisem (approved):

ipn/ipnserver/server.go
bradfitz: sure, done.

ipn/ipnserver/server.go
bradfitz: done

PR #6537, opened 2022-11-28
cmd/tailscaled: unify the two Windows paths + separate IPN server path

bradfitz:
tailscaled on Windows had two entirely separate start-up paths for running
as a service vs in the foreground. It's been causing problems for ages.
This unifies the two paths, making them be the same as the path used
for every other platform.

Also, it uses the new async LocalBackend support in ipnserver.Server
so the Server can start serving HTTP immediately, even if tun takes
awhile to come up.

Updates #6535


bradfitz:
(Tests failing because signals aren't wired up yet)

bradfitz:
Ah, no, signals were just in the wrong place. Moved to ipnserver instead of getengine.

Tests pass now.

PTAL.

cmd/tailscaled/tailscaled.go
```diff
@@ -368,23 +374,87 @@ func run() error {
...
+	}()
+
+	err = srv.Run(ctx, ln)
+
+	if err != nil && lbErr.Load() != nil {
+		return fmt.Errorf("getLocalBackend error: %v", lbErr.Load())
+	}
+
+	// Cancelation is not an error: it is the only way to stop ipnserver.
+	if err != nil && err != context.Canceled {
```
andrew-d, line 436:
```suggestion
	if err != nil && !errors.Is(err, context.Canceled) {
```

To handle error wrapping?

cmd/tailscaled/tailscaled_windows.go
```diff
@@ -283,140 +300,6 @@ func beFirewallKillswitch() bool {
...
-			if res.Engine != nil {
-				return res.Engine, res.Netstack, nil
-			}
-			if time.Since(t0) < time.Minute || windowsUptime() < 10*time.Minute {
-				// Ignore errors during early boot. Windows 10 auto logs in the GUI
-				// way sooner than the networking stack components start up.
-				// So the network will fail for a bit (and require a few tries) while
-				// the GUI is still fine.
-				continue
-			}
```
andrew-d, line 397: It looks like this has been replaced by the `ERROR_DEVICE_NOT_AVAILABLE` check, above; are we sure that is the only error code? It looks like this code ignores all errors on early boot, not just that one.

andrew-d (commented):
First thoughts; need to run to a meeting so will re-read again later.

cmd/tailscaled/tailscaled.go
bradfitz: This was just movement, but SGTM. Done.

cmd/tailscaled/tailscaled_windows.go
bradfitz: Whoops, I'd missed that part in movement. Restored.

ipn/ipnserver/server.go
```diff
@@ -59,7 +63,64 @@ func (s *Server) mustBackend() *ipnlocal.LocalBackend {
...
+		Error string `json:"error,omitempty"`
+	}
+
+	lb := s.lb.Load()
+	if lb == nil {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		if wait, _ := strconv.ParseBool(r.FormValue("wait")); wait {
+			w.(http.Flusher).Flush()
+			lb, _ = s.awaitBackend(ctx)
+		}
```
andrew-d, line 113: Do we want to write `http.StatusServiceUnavailable` even when waiting?

cmd/tailscaled/tailscaled_windows.go
```diff
@@ -283,140 +301,6 @@ func beFirewallKillswitch() bool {
...
-		}
-	}()
-
-	// getEngine is called by ipnserver to get the engine. It's
-	// not called concurrently and is not called again once it
-	// successfully returns an engine.
-	getEngine := func() (wgengine.Engine, *netstack.Impl, error) {
-		if msg := envknob.String("TS_DEBUG_WIN_FAIL"); msg != "" {
-			return nil, nil, fmt.Errorf("pretending to be a service failure: %v", msg)
-		}
```
andrew-d, line 385: I don't know if it matters, but I don't see this in the refactored code.

cmd/tailscaled/tailscaled_windows.go
```diff
@@ -283,140 +301,6 @@ func beFirewallKillswitch() bool {
...
-			r.Close()
-			dev.Close()
-			return nil, nil, fmt.Errorf("engine: %w", err)
-		}
-		ns, err := newNetstack(logf, dialer, eng)
-		if err != nil {
-			return nil, nil, fmt.Errorf("newNetstack: %w", err)
-		}
-		ns.ProcessLocalIPs = false
-		ns.ProcessSubnets = shouldWrapNetstack()
```
andrew-d, line 338: Worth verifying that we didn't break/change the behaviour of netstack on Windows before merging, if you haven't already.

andrew-d (approved):
Generally LGTM with some small comments; feel free to address async if you'd rather.

cmd/tailscaled/tailscaled_windows.go
bradfitz: We haven't used this in ages. I can always add it back if/when needed. And name it something generic without "WIN" in the name.

Also maybe a sleep duration before the engine becomes available, to test the async bits, once there are more tests here.

ipn/ipnserver/server.go
bradfitz: Yeah, I figured that'd be the signal that "we're actually waiting".

Admittedly this isn't a very defined protocol.

cmd/tailscaled/tailscaled_windows.go
bradfitz: just verified. looks consistent.

PR #6545, opened 2022-11-28
api.md: change admin panel to admin console

JuliaTailscale:


bradfitz:
small nit: the verb tense in the git commit title should be "change" instead of "changed"

We generally try to stay consistent with the https://github.com/golang/go/wiki/CommitMessage style.

JuliaTailscale:
> small nit: the verb tense in the git commit title should be "change" instead of "changed"
> 
> We generally try to stay consistent with the https://github.com/golang/go/wiki/CommitMessage style.

Done; thank you!

DentonGentry:
The Continuous Integration check which failed is that we require a Signed-Off-By line in the commit message, that the submitter states that the code can be released under the open source license used by this repository.

The branch name in this commit is `JuliaTailscale:patch-1`, which is what the GitHub editor usually names its branches. Was this Pull Request generated from the editor in the GitHub website? If so, I don't know of a way to add a Signed-Off-By line in the editor but I think I have a way to fix it on your behalf.

----

If instead you generated this change from a git command line client, you can fix it like so:
```
git commit --amend --signoff
git push -f
```

bradfitz:
> Done; thank you!

You changed the PR title, not the git commit's title.  (In general we care more about the git commit message because it goes into history whereas PRs are more ephemeral at the time of review)

PR #6546, opened 2022-11-29
ipn,types/persist: store disallowed TKA's in prefs, lock local-disable

twitchyliquid64:
Implements `tailscale lock local-disable`, which shuts down tailnet lock on the local node. This can be done on all nodes if the disablement secrets are lost by the admin but they want to disable network lock.

/cc @maisem for changes to `persist.Persist`, and that I'm doing the right sequence to update that struct in `(*LocalBackend).NetworkLockForceLocalDisable`.

### DIFF
--- a/ipn/ipnlocal/network-lock.go
+++ b/ipn/ipnlocal/network-lock.go
@@ -25,6 +25,7 @@ import (
 	"tailscale.com/tka"
 	"tailscale.com/types/key"
 	"tailscale.com/types/netmap"
+	"tailscale.com/types/persist"
 	"tailscale.com/types/tkatype"
 	"tailscale.com/util/mak"
 )
@@ -134,7 +135,7 @@ func (b *LocalBackend) tkaSyncIfNeeded(nm *netmap.NetworkMap, prefs ipn.PrefsVie
 		}
 
 		if wantEnabled && !isEnabled {
-			if err := b.tkaBootstrapFromGenesisLocked(bs.GenesisAUM); err != nil {
+			if err := b.tkaBootstrapFromGenesisLocked(bs.GenesisAUM, prefs.Persist()); err != nil {
 				return fmt.Errorf("bootstrap: %w", err)
 			}
 			isEnabled = true
@@ -278,7 +279,7 @@ func (b *LocalBackend) chonkPathLocked() string {
 // tailnet key authority, based on the given genesis AUM.
 //
 // b.mu must be held.
-func (b *LocalBackend) tkaBootstrapFromGenesisLocked(g tkatype.MarshaledAUM) error {
+func (b *LocalBackend) tkaBootstrapFromGenesisLocked(g tkatype.MarshaledAUM, persist *persist.Persist) error {
 	if err := b.CanSupportNetworkLock(); err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


ipn/ipn_clone.go
```diff
@@ -24,10 +24,7 @@ func (src *Prefs) Clone() *Prefs {
 	*dst = *src
 	dst.AdvertiseTags = append(src.AdvertiseTags[:0:0], src.AdvertiseTags...)
 	dst.AdvertiseRoutes = append(src.AdvertiseRoutes[:0:0], src.AdvertiseRoutes...)
-	if dst.Persist != nil {
```
twitchyliquid64, line 27: Hey @maisem this is what happens if I run `go generate` in `ipn/`, I presume this is not intentional. Want me to put it back and ignore the go-generate github check?

maisem: This looks correct. Previously `Persist` was a value type and didn't contain any pointers or slices so it was safe to pass around by just doing a shallow copy. With your change it now has mutable data so it requires more protection around it which means that it needs to return a PersistView from the call to `PrefsView.Persist()`.

ipn/ipnlocal/network-lock.go
```diff
@@ -495,6 +509,31 @@ func (b *LocalBackend) NetworkLockKeyTrustedForTest(keyID tkatype.KeyID) bool {
...
+	id1, id2 := b.tka.authority.StateIDs()
+	stateID := fmt.Sprintf("%d:%d", id1, id2)
+
+	newPrefs := b.pm.CurrentPrefs().AsStruct().Clone() // .Persist should always be initialized here.
+	newPrefs.Persist.DisallowedTKAStateIDs = append(newPrefs.Persist.DisallowedTKAStateIDs, stateID)
+	if err := b.pm.SetPrefs(newPrefs.View()); err != nil {
+		return fmt.Errorf("saving prefs: %w", err)
+	}
+
+	if err := os.RemoveAll(b.chonkPathLocked()); err != nil {
```
sailorfrag, line 530: I'm not sure if we'd want to delete the whole chonk in case there are other tailnets being used that have network lock. Maybe just everything with a matching state ID. A scan shouldn't be too expensive?

types/persist/persist.go
```diff
@@ -70,7 +77,8 @@ func (p *Persist) Equals(p2 *Persist) bool {
 		p.LoginName == p2.LoginName &&
 		p.UserProfile == p2.UserProfile &&
 		p.NetworkLockKey.Equal(p2.NetworkLockKey) &&
-		p.NodeID == p2.NodeID
+		p.NodeID == p2.NodeID &&
+		reflect.DeepEqual(p.DisallowedTKAStateIDs, p2.DisallowedTKAStateIDs)
```
sailorfrag, line 81: Might want to make it accept permutations of the same IDs?

sailorfrag (approved):

ipn/ipnlocal/network-lock.go
twitchyliquid64: With the FUS changes chonks are now per-profile, so deleting the chonk should not affect any other tailnets with network lock:

`filepath.Join(b.TailscaleVarRoot(), "tka-profiles", string(b.pm.CurrentProfile().ID))`

types/persist/persist.go
twitchyliquid64: The only two places that `Equals` is used outside of tests is detecting when prefs changes for committing to disk, and for writing to the IPN bus. Given this, I think its best I leave it with the simple implementation: Equality checks might get expensive otherwise, and theres no consequence to returning false if the state IDs are in a different order.

PR #6557, opened 2022-11-29
ipn: re-run go generate, fix build

bradfitz:
Broken by #6546


bradfitz:
Ugh, then things don't even compile.

bradfitz:
This got invasive so I gave up.

PR #6560, opened 2022-11-29
ipn,types/persist: add DisallowedTKAStateIDs, refactor as view type

twitchyliquid64:
Supercedes https://github.com/tailscale/tailscale/pull/6557, precursor to trying https://github.com/tailscale/tailscale/pull/6546 again

### DIFF
--- a/types/persist/persist.go
+++ b/types/persist/persist.go
@@ -7,6 +7,7 @@ package persist
 
 import (
 	"fmt"
+	"reflect"
 
 	"tailscale.com/tailcfg"
 	"tailscale.com/types/key"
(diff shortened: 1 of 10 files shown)


types/persist/persist_test.go
```diff
@@ -133,6 +133,11 @@ func TestPersistEqual(t *testing.T) {
 			&Persist{NodeID: "abc"},
 			false,
 		},
+		{
+			&Persist{DisallowedTKAStateIDs: nil},
+			&Persist{DisallowedTKAStateIDs: []string{"0:0"}},
+			false,
+		},
```
maisem, line 140: add the positive case too

maisem (approved):

types/persist/persist.go
```diff
@@ -70,7 +77,8 @@ func (p *Persist) Equals(p2 *Persist) bool {
 		p.LoginName == p2.LoginName &&
 		p.UserProfile == p2.UserProfile &&
 		p.NetworkLockKey.Equal(p2.NetworkLockKey) &&
-		p.NodeID == p2.NodeID
+		p.NodeID == p2.NodeID &&
+		reflect.DeepEqual(p.DisallowedTKAStateIDs, p2.DisallowedTKAStateIDs)
```
bradfitz, line 81: do you care about nil vs empty slice? this treats them differently, but I doubt that matters. you might want to use that len-zero-map-to-nil func we have somewhere

bradfitz:
```
bradfitz@tsdev:~/src/tailscale.com$ git grep ^func | grep nil | grep any
wgengine/router/ifconfig_windows_test.go:func nilIfEmpty[E any](s []E) []E {
```

twitchyliquid64: Done.

PR #6562, opened 2022-11-29
ipn/ipnlocal: account for ControlURL when merging profiles

maisem:
We merge/dedupe profiles based on UserID and NodeID, however we were not accounting for ControlURLs.

Updates #713

### DIFF
--- a/ipn/ipnlocal/profiles.go
+++ b/ipn/ipnlocal/profiles.go
@@ -98,21 +98,25 @@ func (pm *profileManager) matchingProfiles(f func(*ipn.LoginProfile) bool) (out 
 	return out
 }
 
-func (pm *profileManager) findProfilesByNodeID(nodeID tailcfg.StableNodeID) []*ipn.LoginProfile {
+// findProfilesByNodeID returns all profiles that have the provided nodeID and
+// belong to the same control server.
+func (pm *profileManager) findProfilesByNodeID(controlURL string, nodeID tailcfg.StableNodeID) []*ipn.LoginProfile {
 	if nodeID.IsZero() {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/profiles.go
```diff
@@ -98,21 +98,21 @@ func (pm *profileManager) matchingProfiles(f func(*ipn.LoginProfile) bool) (out
 	return out
 }
 
-func (pm *profileManager) findProfilesByNodeID(nodeID tailcfg.StableNodeID) []*ipn.LoginProfile {
+func (pm *profileManager) findProfilesByNodeID(controlURL string, nodeID tailcfg.StableNodeID) []*ipn.LoginProfile {
```
bradfitz, line 101: method name doesn't quite match. either rename method and/or document what controlURL does?

bradfitz (approved):

ipn/ipnlocal/profiles.go
maisem: Done

ipn/prefs.go
```diff
@@ -741,4 +741,8 @@ type LoginProfile struct {
 	// It is only relevant on Windows where we have a multi-user system.
 	// It is assigned once at profile creation time and never changes.
 	LocalUserID WindowsUserID
+
+	// ControlURL is the URL of the control server that this profile is logged
+	// into.
+	ControlURL string
```
mihaip, line 747: Without a backfill existing profiles will not be considered to match the default control URL, though it's probably not worth doing anything about it for the few dozen people with unstable builds.

mihaip (approved):

ipn/prefs.go
maisem: that's what i was thinking too

Issue #5156, opened 2022-07-23
MacOS tailscale starts using 100% CPU and lots of network looping Wireguard frames


dmd:
### What is the issue?

<img width="313" alt="image" src="https://user-images.githubusercontent.com/41439/180605504-f3476e17-d11b-438f-a9de-a8ab030e306c.png">

About 3x a week on MacOS tailscale starts using 100% cpu and 20 Mbps of bandwidth for no apparent reason. Quitting and restarting it fixes the problem.

### OS

macOS

### OS version

occurred under 12.4, still occurring under 12.5

### Tailscale version

1.26.2

### Bug report

BUG-bdc30f2af8ebb0a7aa591939e9f240a3a57392e2de308db311aab847ee2a88a0-20220723124735Z-31c1419398423a01

DentonGentry:
Could you run Wireshark on the utun interface during one of the times when the network bandwidth is so high, and see what it is?

There will be a number of interfaces listed, from utun0 to (probably) utun5. You can find the right one using either `ifconfig` in Terminal.app, or leave Wireshark's initial page open for a while and watch the traffic graphs to see which one is really active.

dmd:
Yes, will do next time I see this behavior.

mjibson:
Have been seeing occasional 100% CPU on 1.28.0 on linux. `tailscale down` doesn't help, have to kill the process.

bradfitz:
@mjibson, any network traffic when it happens? Is it a DNS loop?

Can you grab a CPU profile next time it happens? `tailscale debug --cpu-profile=prof.cpu --profile-seconds=15`

mjibson:
Profile attached. wireshark on tailscale0 didn't show any packets during this.
[prof.zip](https://github.com/tailscale/tailscale/files/9306147/prof.zip)

bradfitz:
@mjibson, oh, that's #4760 which was a Go bug and now fixed upstream. The fix is in Go 1.18.5 or Go 1.19.

Our official binaries for Tailscale 1.28.0 were built at https://github.com/tailscale/tailscale/blob/v1.28.0/go.toolchain.rev (https://github.com/tailscale/go/commit/149f7d88f11384083d42ccbcdb31693510385ec6) which included the fix: https://github.com/tailscale/tailscale/commit/7fb6781bdaa4cccb61497ccf061fcd5a525a329e

So I guess you built Tailscale yourself. Update your Go.

[...]

PR #6566, opened 2022-11-30
net/netns, net/interfaces: explicitly bind sockets to the default interface on all Darwin variants

### LINKED ISSUE #6565: iOS keeps using cellular data after default route changes to WiFi
@dsnet reported that when his iPhone gets on WiFi it continues to communicate with his NAS over the cellular connection. We investigated, and got as far as noticing that Magicsock tries to do DISCO pings over the new endpoints that it was told about, but bi-directional communication is not possible:

These are the [logs from Joe’s phone](http://logz.ts.tailscale.com:9431/logs?start=2022-11-10T22%3A19%3A57Z&end=now&instances=d25cc550108d0dad040aa78f4dd581fad1a83d244ec43c6f0e4552e66c7b9edd), you can see that it both sends and gets a ping from his NAS over his LAN (192.168.0.101):

```
magicsock: disco: d:baf1be2bc22288d6->d:63f0e05d66bb491d ([K94YI], 192.168.0.101:41641) sent ping tx=8609cd4c7a33
magicsock: disco: d:baf1be2bc22288d6<-d:63f0e05d66bb491d ([K94YI], 192.168.0.101:41641) got ping tx=076ceb0acb7a
```

The [NAS’s logs](http://logz.ts.tailscale.com:9431/logs?start=2022-11-10T22%3A19%3A57Z&end=now&instances=afad3037420ff1d02cf2b22eeae5fee2c476b9912fc0d9fbd41fabfe9f1d9f2e) show the ping that it sent, but it doesn’t get the ping or pong from his phone:

```
magicsock: disco: d:63f0e05d66bb491d->d:baf1be2bc22288d6 ([wNcRJ], 192.168.0.183:41641) sent ping tx=076ceb0acb7a
```

We’re assuming that the iOS network stack is trying to send the UDP packets to 192.168.0.101 over the cell connection. It would be nice if when we discover new interfaces/endpoints we bind the sockets to those interfaces instead of letting iOS choose.

_Originally posted by @mihaip in https://github.com/tailscale/tailscale/issues/6152#issuecomment-1315910954_

mihaip:
We were previously only doing this for tailscaled-on-Darwin, but it also appears to help on iOS. Otherwise, when we rebind magicsock UDP connections after a cellular -> WiFi interface change they still keep using cellular one.

To do this correctly when using exit nodes, we need to exclude the Tailscale interface when getting the default route, otherwise packets cannot leave the tunnel. There are native macOS/iOS APIs that we can use to do this, so we allow those clients to override the implementation of `DefaultRouteInterfaceIndex`.

Updates #6565, may also help with #5156


mihaip:
Definitely open to alternate ideas here. Possible ways to make this less conservative:
- Make it be controlled by a debug flag from control, so that we can experiment with it
- Only bind to non-"expensive" interfaces (we can either wire this up from the Swift side, or get the interface type and see if it's `syscall.IFT_CELLULAR`)

mihaip:
Moving back to draft. Using an exit node is broken, will need to investigate more.

net/interfaces/interfaces_bsd.go
```diff
@@ -37,85 +39,118 @@ func defaultRoute() (d DefaultRouteDetails, err error) {
...
+// interface.
+func SetTailscaleInterfaceIndex(index uint32) {
+	tailscaleInterfaceIndex.Store(index)
 }
 
 var v4default = [4]byte{0, 0, 0, 0}
 var v6default = [16]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
 
-func isDefaultGateway(rm *route.RouteMessage) bool {
+func isDefaultGateway(rm *route.RouteMessage, allowInterfaceScopeRoute bool) bool {
```
bradfitz, line 144: little docs on this?

net/interfaces/interfaces_bsd.go
```diff
@@ -37,85 +39,118 @@ func defaultRoute() (d DefaultRouteDetails, err error) {
 }
 
 func DefaultRouteInterfaceIndex() (int, error) {
```
bradfitz, line 41: this is specifically for IPv4, right? docs while you're here?

I think elsewhere we have a TODO to add an address family arg to this if/when any caller needs to distinguish those. And IIRC we recently did need that but for Windows only so far.

net/interfaces/interfaces_bsd.go
```diff
@@ -37,85 +39,118 @@ func defaultRoute() (d DefaultRouteDetails, err error) {
...
+	}
+	return rms[0].Index, nil
+}
+
+func init() {
+	likelyHomeRouterIP = likelyHomeRouterIPBSDFetchRIB
+}
+
+func likelyHomeRouterIPBSDFetchRIB() (ret netip.Addr, ok bool) {
+	rms, err := defaultRouteMessages(math.MaxInt)
```
bradfitz, line 57: MaxInt is a little overkill. Also for the math package. Also because Go's int size (which varies by platform) might not match the kernel's anyway. I doubt you'd get more than 32 bits of routes from the kernel. Maybe just write 100 or something reasonable here?

net/interfaces/interfaces_bsd.go
```diff
@@ -37,85 +39,118 @@ func defaultRoute() (d DefaultRouteDetails, err error) {
...
+		return nil, fmt.Errorf("route.FetchRIB: %w", err)
 	}
 	msgs, err := parseRoutingTable(rib)
 	if err != nil {
-		return 0, fmt.Errorf("route.ParseRIB: %w", err)
+		return nil, fmt.Errorf("route.ParseRIB: %w", err)
 	}
+	tailscaleInterfaceIndex := int(tailscaleInterfaceIndex.Load())
+	allowInterfaceScopeRoute := false
+	rms := make([]*route.RouteMessage, 0, 1)
```
bradfitz, line 109: capacity 1? not like 2 for macs on wifi + ethernet?

bradfitz (approved):

net/interfaces/interfaces_bsd.go
mihaip: It does handle IPv6 though: https://github.com/tailscale/tailscale/blob/e7f9b92e1a394f4d5600feac223f79ed9c97d4ef/net/interfaces/interfaces_bsd.go#L175-L181

Added some docs.

bradfitz: No, I mean: it's possible to configure machines such that their default route for IPv4 is a different interface from IPv6. This API doesn't support passing which the caller cares about. That's fine. No need to fix today. Just document whatever the behavior is

dsnet (approved):
I'm no expert in the Apple networking code, but it generally seems sensible.

mihaip:
> I'm no expert in the Apple networking code, but it generally seems sensible.

It's mostly an FYI, though if you wanted, you could run a custom iOS build with this and https://github.com/tailscale/corp/pull/8201 to check if it fixes problem for you.

mihaip:
As discussed on Slack, the default-route-when-an-exit-node-is-used mode did not handle the macOS network service order preference. I've redone this to allow an alternate default interface function to be provided (implemented in tailscale/corp#8201), since the iOS/macOS native code has that information. 

It ends up being quite a but a bit simpler, but PTAL since there's significant changes from the last version.

PR #6569, opened 2022-11-30
ipn: be more consistent with omitempty on debug/rare Prefs

bradfitz:


bradfitz:
(noticed while testing #6568 on Windows)

maisem (approved):

PR #6574, opened 2022-11-30
ipn/ipnlocal,tailcfg: introduce capability to gate TKA init paths

twitchyliquid64:
**CAPVER 51**

Previously, `TAILSCALE_USE_WIP_CODE` was needed to hit a bunch of the TKA paths. With this change:

 - Enablement codepaths (`NetworkLockInit`) and initialization codepaths (`tkaBootstrapFromGenesisLocked` via `tkaSyncIfNeeded`) require either the WIP envknob or `CapabilityTailnetLockAlpha`.
 - Normal operation codepaths (`tkaSyncIfNeeded`, `tkaFilterNetmapLocked`) require TKA to be initialized, or either-or the envknob / capability.
 - Auxillary commands (ie: changing tka keys) require TKA to be initialized.

The end result is that it shouldn't be possible to initialize TKA (or subsequently use any of its features) without being sent the capability or setting the envknob on tailscaled yourself.

I've also pulled out a bunch of unnecessary checks for `CanSupportNetworkLock()`.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -162,6 +162,7 @@ type LocalBackend struct {
 	tka            *tkaState
 	state          ipn.State
 	capFileSharing bool // whether netMap contains the file sharing capability
+	capTailnetLock bool // whether netMap contains the tailnet lock capability
 	// hostinfo is mutated in-place while mu is held.
 	hostinfo *tailcfg.Hostinfo
 	// netMap is not mutated in-place once set.
@@ -869,6 +870,8 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 	}
 	// Prefs will be written out; this is not safe unless locked or cloned.
 	if st.NetMap != nil {
+		b.capTailnetLock = hasCapability(st.NetMap, tailcfg.CapabilityTailnetLockAlpha)
+
 		b.mu.Unlock() // respect locking rules for tkaSyncIfNeeded
 		if err := b.tkaSyncIfNeeded(st.NetMap, prefs.View()); err != nil {
 			b.logf("[v1] TKA sync error: %v", err)
--- a/ipn/ipnlocal/network-lock.go
+++ b/ipn/ipnlocal/network-lock.go
@@ -46,13 +46,21 @@ type tkaState struct {
 	filtered  []ipnstate.TKAFilteredPeer
 }
 
+// permitTKAInitLocked returns true if tailnet lock initialization may
+// occur.
+// b.mu must be held.
+func (b *LocalBackend) permitTKAInitLocked() bool {
+	return envknob.UseWIPCode() || b.capTailnetLock
+}
+
 // tkaFilterNetmapLocked checks the signatures on each node key, dropping
 // nodes from the netmap whose signature does not verify.
 //
 // b.mu must be held.
 func (b *LocalBackend) tkaFilterNetmapLocked(nm *netmap.NetworkMap) {
-	if !envknob.UseWIPCode() {
-		return // Feature-flag till network-lock is in Alpha.
+	// TODO(tom): Remove this guard for 1.35 and later.
+	if b.tka == nil && !b.permitTKAInitLocked() {
+		return
 	}
 	if b.tka == nil {
 		return // TKA not enabled.
@@ -61,7 +69,7 @@ func (b *LocalBackend) tkaFilterNetmapLocked(nm *netmap.NetworkMap) {
 	var toDelete map[int]bool // peer index => true
 	for i, p := range nm.Peers {
 		if p.UnsignedPeerAPIOnly {
-			// Not subject to TKA.
+			// Not subject to tailnet lock.
 			continue
 		}
 		if len(p.KeySignature) == 0 {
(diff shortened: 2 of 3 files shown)


ipn/ipnlocal/local.go
```diff
@@ -222,6 +222,8 @@ type LocalBackend struct {
 	// tkaSyncLock MUST be taken before mu (or inversely, mu must not be held
 	// at the moment that tkaSyncLock is taken).
 	tkaSyncLock sync.Mutex
+
+	capTailnetLockAlpha bool // whether netmap contains the tailnet lock alpha capability
```
bradfitz, line 226: what are the locking rules for this? It's in a weird section (under a different mutex)

bradfitz: looks like `b.mu`, not `tkaSyncLock`.

ipn/ipnlocal/network-lock.go
```diff
@@ -102,18 +106,21 @@ func (b *LocalBackend) tkaFilterNetmapLocked(nm *netmap.NetworkMap) {
...
-
-	b.logf("tkaSyncIfNeeded: enabled=%v, head=%v", nm.TKAEnabled, nm.TKAHead)
-
 	b.tkaSyncLock.Lock() // take tkaSyncLock to make this function an exclusive section.
 	defer b.tkaSyncLock.Unlock()
 	b.mu.Lock() // take mu to protect access to synchronized fields.
 	defer b.mu.Unlock()
 
+	// TODO(tom): Remove this guard for 1.35 and later.
+	if b.tka == nil && (!envknob.UseWIPCode() && !b.capTailnetLockAlpha) {
```
bradfitz, line 115: subjective but I think `.. && !(useWIP || capAlpha)` would read more naturally?

ipn/ipnlocal/network-lock.go
```diff
@@ -426,6 +429,15 @@ func (b *LocalBackend) NetworkLockInit(keys []tka.Key, disablementValues [][]byt
 	var ourNodeKey key.NodePublic
 	var nlPriv key.NLPrivate
 	b.mu.Lock()
+
+	// TODO(tom): Remove this guard for 1.35 and later.
+	if !envknob.UseWIPCode() && !b.capTailnetLockAlpha {
```
bradfitz, line 434: time for a method instead? you have it twice.

permitTailnetLockLocked

(hah.... LockLocked)

The second Locked is for `b.mu` must be held.

tailcfg/tailcfg.go
```diff
@@ -1730,6 +1731,12 @@ const (
 
 	// CapabilityWarnFunnelNoHTTPS indicates HTTPS has not been enabled for the tailnet.
 	CapabilityWarnFunnelNoHTTPS = "https://tailscale.com/cap/warn-funnel-no-https"
+
+	// CapabilityTailnetLockAlpha indicates the node is in the tailnet lock alpha,
```
bradfitz, line 1735: wrong section. These are the warnings.  Put this up in the "Node Attributes" section (under line 1701/1702)

ipn/ipnlocal/local.go
twitchyliquid64: Moved it up to the one for `b.mu`, which is the same for `capFileSharing`.

ipn/ipnlocal/network-lock.go
twitchyliquid64: I agree! updating everywhere.

ipn/ipnlocal/network-lock.go
twitchyliquid64: 2 is the magic number? XD

Will do.

tailcfg/tailcfg.go
twitchyliquid64: Doneskie

ipn/ipnlocal/network-lock.go
bradfitz: It's fine as-is. No need for a method really. The WIP part will go away anyway, right?

twitchyliquid64:
That feels nicer!

ipn/ipnlocal/local.go
```diff
@@ -150,18 +150,19 @@ type LocalBackend struct {
...
+	ccGen               clientGen    // function for producing controlclient; lazily populated
+	sshServer           SSHServer    // or nil, initialized lazily.
+	notify              func(ipn.Notify)
+	cc                  controlclient.Client
+	ccAuto              *controlclient.Auto // if cc is of type *controlclient.Auto
+	machinePrivKey      key.MachinePrivate
+	tka                 *tkaState
+	state               ipn.State
+	capFileSharing      bool // whether netMap contains the file sharing capability
+	capTailnetLockAlpha bool // whether netMap contains the tailnet lock alpha capability
```
sailorfrag, line 165: I kind of feel like `catpTailnetLock` would be less likely to go out of date in future if we keep that beyond the alpha timeframe. It's also shorter :-D

sailorfrag (approved):

ipn/ipnlocal/local.go
twitchyliquid64: Done.

ipn/ipnlocal/network-lock.go
```diff
@@ -46,13 +46,21 @@ type tkaState struct {
 	filtered  []ipnstate.TKAFilteredPeer
 }
 
+// permitTKAInitLocked returns true if tailnet lock initialization may
```
bradfitz, line 49: Go style nit: "reports whether" instead of "returns true if"

No need to fix, though.

PR #6579, opened 2022-11-30
tsweb: add transparent compression for StdHandler

raggi:
Implements inline compression for both gzip and brotli via the brotli library. The library requires that Content-Type is set. The implementation here explicitly avoids wrapping the ResponseWriter in cases where Accept-Encoding is not set so as to maximally attempt to get out of the way of hijack and upgrade concerns. It also avoids any attempt at compression if Content-Encoding is already set so that handlers that already perform compression are unaffected.


bradfitz:
We've been down this road before and had to roll it back after it took down control. See git history for bug references?

raggi:
> We've been down this road before and had to roll it back after it took down control. See git history for bug references?

I've looked through history in tsweb and control - there's a very mixed history. There are places where we still have specialized handlers doing this (why I was careful to avoid affecting them), and there are some specialized handlers that were reverted without a description in the commit message as to why. There are easy bugs to make with these implementations, such as handling WriteHeader in the wrong order or not flushing/closing the stream - which my tests here cover (and you've fixed in some of the implementations that remain).

I can't find a commit message that has a good match your description. I found one corp revert (9d74ea9eaa669b45bd521df8f86d7904310bd3cb) for a case, but the revert had no description as to why it was reverted. The code in question has a bug where it was not closing the gzip stream, so very likely it was in some cases not writing a complete stream.

I do see some additional/separate bugs where we're not doing ETags correctly in existing handlers, which I plan to come back and fix independently (they essentially all need a weak prefix).

bradfitz:
We were doing this for all serveMap calls before, which was a ton of extra memory/goroutines.

raggi:
@maisem kindly helped me find the other case in df29e238f3dac3c42508fcef0d38dce1d842e4ef which was also missing a call to `Close`, so the gzipwriters, associated buffers and pools were never getting freed/recycled.

tsweb/compress.go
```diff
@@ -0,0 +1,104 @@
...
+	"bufio"
+	"errors"
+	"io"
+	"net"
+	"net/http"
+
+	"github.com/andybalholm/brotli"
+)
+
+type compressingHandler struct {
```
bradfitz, line 17: why not export this so people can opt into it fine-grained where they want it rather than only tsweb and its weirdo handler signature?

tsweb/compress.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// WriteHeader implements http.ResponseWriter.
+func (w *compressingResponseWriter) WriteHeader(code int) {
+	// If a handler has already set a Content-Encoding, such as for precompressed
+	// assets, skip the compressing writer. This must be recorded before
+	// WriteHeader call as "The header map is cleared when 2xx-5xx headers are
+	// sent".
+	if w.w == nil {
+		if w.ResponseWriter.Header().Get("Content-Encoding") == "" {
```
bradfitz, line 49: skip image/ responses too?

DentonGentry:
The depaware CI failure is real, not a flake.

bradfitz:
Also, what's the context (bug/motivation) for this change?

mihaip:
> Also, what's the context (bug/motivation) for this change?

AFAIK it's about wanting compression for the large assets in the blog post discussed at https://tailscale.slack.com/archives/CQNV2U73N/p1669845175407059?thread_ts=1669060695.400279&cid=CQNV2U73N, though based on later findings in the thread we're already getting compression and this may not be necessary.

raggi:
Yup. I was also somewhat motivated to solve this now as we seem to be somewhat afraid of it, but in the interest of time I think I'm going to put it on the backburner. I do think we should take the time to document precisely where the issues came from.

Things I would want to do with this patch before landing it:

- [ ] do not turn it on for all stdhandlers, make it opt-in instead
- [ ] add a construct that accepts options that enables control over mime types to be compressed, matching against mime types in the response headers
- [ ] explicitly close the writer on hijack, in case the protocol being hijacked has an http response preamble with data still to flush
- [ ] think some more about what common usage would look like with request.context.done, and ensure that the close call requirement is not too burdensome (or improve it).
- [ ] if time allowed, merge zstd support into the path as well, so this could replace other instances where we're doing that, too.

PR #6580, opened 2022-11-30
types/key,cmd/tailscale/cli: support tlpub prefix for tailnet-lock keys

twitchyliquid64:
/cc @mayakacz as discussed on meet

### DIFF
--- a/types/key/nl.go
+++ b/types/key/nl.go
@@ -15,12 +15,19 @@ import (
 
 const (
 	// nlPrivateHexPrefix is the prefix used to identify a
-	// hex-encoded network-lock key.
+	// hex-encoded tailnet-lock key.
 	nlPrivateHexPrefix = "nlpriv:"
 
 	// nlPublicHexPrefix is the prefix used to identify the public
-	// side of a hex-encoded network-lock key.
+	// side of a hex-encoded tailnet-lock key.
 	nlPublicHexPrefix = "nlpub:"
+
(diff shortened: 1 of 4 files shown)


types/key/nl.go
```diff
@@ -15,12 +15,19 @@ import (
...
 	nlPrivateHexPrefix = "nlpriv:"
 
 	// nlPublicHexPrefix is the prefix used to identify the public
-	// side of a hex-encoded network-lock key.
+	// side of a hex-encoded tailnet-lock key.
 	nlPublicHexPrefix = "nlpub:"
+
+	// nlPublicHexPrefixCLI is the prefix used for tailnet-lock keys
+	// when shown on the CLI.
+	// Its not practical for us to change the prefix everywhere due to
```
bradfitz, line 27: It's

types/key/nl.go
```diff
@@ -126,6 +137,12 @@ func (k NLPublic) MarshalText() ([]byte, error) {
 	return toHex(k.k[:], nlPublicHexPrefix), nil
 }
 
+// CLIString returns a marshalled representation suitable for use
```
bradfitz, line 140: also spell out what this means concretely (which prefix it uses instead of which other prefix)

and update docs on the other string method

bradfitz (approved):

types/key/nl.go
twitchyliquid64: Done.

PR #6588, opened 2022-12-01
net/dns: [linux] log and add metric for dnsMode

maisem:
I couldn't find any logs that indicated which mode it was running in so adding that. Also added a gauge metric for dnsMode.

### DIFF
--- a/net/dns/manager_linux.go
+++ b/net/dns/manager_linux.go
@@ -10,12 +10,15 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
+	"sync"
 	"time"
 
 	"github.com/godbus/dbus/v5"
 	"tailscale.com/health"
 	"tailscale.com/net/netaddr"
 	"tailscale.com/types/logger"
+	"tailscale.com/util/clientmetric"
 	"tailscale.com/util/cmpver"
 )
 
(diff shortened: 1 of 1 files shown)


net/dns/manager_linux.go
```diff
@@ -40,6 +44,11 @@ func NewOSConfigurator(logf logger.Logf, interfaceName string) (ret OSConfigurat
 	if err != nil {
 		return nil, err
 	}
+	publishOnce.Do(func() {
+		m := clientmetric.NewGauge(fmt.Sprintf("linux_dns_mode_%s", mode))
```
knyar, line 48: Perhaps `dns_manager_linux_mode_` to keep this in the dns_* namespace?

knyar (approved):

net/dns/manager_linux.go
```diff
@@ -40,6 +44,11 @@ func NewOSConfigurator(logf logger.Logf, interfaceName string) (ret OSConfigurat
...
+		m.Set(1)
+	})
+	logf("dns: using %q mode", mode)
```
DentonGentry, line 51: There is a log of how it decided what mode to operate in:
`dns: [resolved-ping=yes rc=resolved nm=no ret=systemd-resolved]`
However it seems fine to log it here in a more clear way.

DentonGentry (approved):

net/dns/manager_linux.go
maisem: Done.

PR #6592, opened 2022-12-01
ipn/{ipnserver,localapi}: fix InUseOtherUser handling with WatchIPNBus

bradfitz:
Updates tailscale/corp#8222

### DIFF
--- a/ipn/ipnserver/server.go
+++ b/ipn/ipnserver/server.go
@@ -52,7 +52,8 @@ type Server struct {
 	mu            sync.Mutex
 	lastUserID    ipn.WindowsUserID // tracks last userid; on change, Reset state for paranoia
 	activeReqs    map[*http.Request]*ipnauth.ConnIdentity
-	backendWaiter set.HandleSet[context.CancelFunc] // values are wake-up funcs of lb waiters
+	backendWaiter waiterSet // of LocalBackend waiters
+	zeroReqWaiter waiterSet // of blockUntilZeroConnections waiters
 }
 
 func (s *Server) mustBackend() *ipnlocal.LocalBackend {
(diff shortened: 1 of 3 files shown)


bradfitz:
(not done testing this)

bradfitz:
Okay, I think this is good now.

ipn/ipnserver/server.go
```diff
@@ -63,22 +64,47 @@ func (s *Server) mustBackend() *ipnlocal.LocalBackend {
...
 }
 
+// waiterSet is a set of callers waiting on something. Each item (map value) in
+// the set is a func that wakes up that waiter's context. The waiter is responsible
+// for removing itself from the set when woken up. The (*waiterSet).add method
+// returns a cleanup method which does that removal. The caller than defers that
+// cleanup.
+//
+// TODO(bradfitz): this is a generally useful pattern. Move elsewhere?
+type waiterSet set.HandleSet[context.CancelFunc]
```
maisem, line 74: if the goal is to wake up everyone on an event, you could also do something like this and not have to worry about all the accounting and cleanup.

```
type waiterSet struct {
  ch syncs.AtomicValue[chan struct{}]
}
func (s *waiterSet) get() (ready <-chan struct{}) {
  return s.ch.Load()
}
func (s *waiterSet) wakeAll() {
  close(s.ch.Swap(make(chan struct{})))
}
```

maisem: you will need to initialize it once

ipn/localapi/localapi.go
```diff
@@ -590,6 +591,32 @@ func (h *Handler) serveStatus(w http.ResponseWriter, r *http.Request) {
 	e.Encode(st)
 }
 
+// InUseOtherUserIPNStream reports whether r is a request for the watch-ipn-bus
+// handler. If so, it writes an ipn.Notify InUseOtherUser message to the user
+// returns true. Otherwise it returns false, in which case it doesn't write to
```
maisem, line 596: And returns true?

maisem (approved):

ipn/ipnserver/server.go
creachadair: ^ I've likewise had good luck using this simpler pattern for the broadcast case (though I did it with explicit locks, https://pkg.go.dev/github.com/creachadair/msync#Trigger).

bradfitz: Per chat, we're going to do this later in its own PR adding the type/docs/tests and convert multiple callers.

ipn/localapi/localapi.go
bradfitz: Done.

PR #6593, opened 2022-12-01
api.md: change admin panel to admin console

JuliaTailscale:
We've settled on "console" as a clearer term.

Signed-off-by: Julia Stein Randall <julia@tailscale.com>


bradfitz:
(new version of #6545)

I assume the new PR is just getting up to speed with git? Normally during the course of a review you re-use the same PR and just edit it until it's ready.

Looks like DCO is still not happy here. I bet you need a blank line before the "Signed-off-by" line so it's considered to be in the "footer" of the commit message.

PR #6599, opened 2022-12-02
paths: set QNAP socket to /tmp.

DentonGentry:

### DIFF
--- a/paths/paths.go
+++ b/paths/paths.go
@@ -37,6 +37,8 @@ func DefaultTailscaledSocket() string {
 		return "/var/packages/Tailscale/var/tailscaled.sock"
 	case distro.Gokrazy:
 		return "/perm/tailscaled/tailscaled.sock"
+	case distro.QNAP:
+		return "/tmp/tailscale/tailscaled.sock"
 	}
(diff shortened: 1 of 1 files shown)


bradfitz (approved):

bradfitz (commented):
Such a weird path but 🤷

DentonGentry:
We could move the socket to the package install directory instead of /tmp, but QNAP has a much wider variety of paths where packages might be installed. We'd need to have tailscaled fork `/sbin/getcfg Tailscale Install_Path -f /etc/config/qpkg.conf`

```
[admin@QNAP ~]$ /sbin/getcfg Tailscale Install_Path -f /etc/config/qpkg.conf
/share/CACHEDEV1_DATA/.qpkg/Tailscale
```

bradfitz:
Can you strace that `getcfg` command? What's it do? Something we could do ourselves without the fork optimistically, only falling back to calling it if we fail to find it?

bradfitz:
In any case, this is fine to submit for now. We can get fancier later.

DentonGentry:
We'd have to get an strace binary onto the QNAP, it doesn't have one.
I'll submit this as it is for now.

PR #6604, opened 2022-12-02
cmd/tailscaled: rename variables to be more descriptive

maisem:
renamed from `useNetstack` to `onlyNetstack` which is 1 letter more but
more descriptive because we always have netstack enabled and `useNetstack`
doesn't convey what it is supposed to be used for. e.g. we always use
netstack for Tailscale SSH.

Also renamed shouldWrapNetstack to handleSubnetsInNetstack as it was only used
to configure subnet routing via netstack.

Updates tailscale/corp#8020

### DIFF
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -462,7 +462,7 @@ func getLocalBackend(ctx context.Context, logf logger.Logf, logid string) (_ *ip
 	socksListener, httpProxyListener := mustStartProxyListeners(args.socksAddr, args.httpProxyAddr)
 
 	dialer := &tsdial.Dialer{Logf: logf} // mutated below (before used)
-	e, useNetstack, err := createEngine(logf, linkMon, dialer)
+	e, onlyNetstack, err := createEngine(logf, linkMon, dialer)
 	if err != nil {
 		return nil, fmt.Errorf("createEngine: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


cmd/tailscaled/tailscaled.go
```diff
@@ -540,24 +540,24 @@ func getLocalBackend(ctx context.Context, logf logger.Logf, logid string) (_ *ip
...
 		}
 		logf("wgengine.NewUserspaceEngine(tun %q) error: %v", name, err)
 		errs = append(errs, err)
 	}
 	return nil, false, multierr.New(errs...)
 }
 
-func shouldWrapNetstack() bool {
+func handleSubnetsInNetstack() bool {
 	if v, ok := envknob.LookupBool("TS_DEBUG_WRAP_NETSTACK"); ok {
```
bradfitz, line 561: this env too then?

cmd/tailscaled/tailscaled.go
```diff
@@ -540,24 +540,24 @@ func getLocalBackend(ctx context.Context, logf logger.Logf, logid string) (_ *ip
...
+			return e, userspaceNetworking, nil
 		}
 		logf("wgengine.NewUserspaceEngine(tun %q) error: %v", name, err)
 		errs = append(errs, err)
 	}
 	return nil, false, multierr.New(errs...)
 }
 
-func shouldWrapNetstack() bool {
+func handleSubnetsInNetstack() bool {
```
bradfitz, line 560: docs then while you're here and apparently caring about such details?

bradfitz (approved):
Not sure this is more than a neutral change (longer words but not much more descriptive/disambiguating?) but sure, if you care strongly?

cmd/tailscaled/tailscaled.go
maisem: Done

cmd/tailscaled/tailscaled.go
maisem: Done.

maisem:
> Not sure this is more than a neutral change (longer words but not much more descriptive/disambiguating?) but sure, if you care strongly?

Renamed to `onlyNetstack`, I think that is at least more descriptive in comparison to `useNetstack` because `use` doesn't clarify use for what. `only` at least implies that we only have `netstack` and nothing else (i.e. no tun).

PR #6616, opened 2022-12-05
ipn/ipnlocal: fix use of stale profile while processing netmap

twitchyliquid64:


DentonGentry:
Do we want this in 1.34.0? There isn't much time to re-test.

twitchyliquid64:
I'm fine with this not being in `1.34.0`, on the basis that the only-known encounter is when logging out + logging in again with tailnet lock enabled, and the failure mode is quite reasonable.

maisem (approved):

DentonGentry:
Based on discussion in Slack: it is fine to submit this in time for 1.34.0.

PR #6617, opened 2022-12-05
ipn/ipnlocal,cmd/tailscale: minor improvements to lock modify command

twitchyliquid64:
* Do not print the status at the end of a successful operation
 * Ensure the key of the current node is actually trusted to make these changes
 * `go fmt` on `network-lock.go`

### DIFF
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -1254,14 +1254,7 @@ func (h *Handler) serveTKAModify(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "network-lock modify failed: "+err.Error(), http.StatusInternalServerError)
 		return
 	}
-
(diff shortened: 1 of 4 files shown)


ipn/localapi/localapi.go
```diff
@@ -1254,14 +1254,7 @@ func (h *Handler) serveTKAModify(w http.ResponseWriter, r *http.Request) {
...
 	}
-
-	j, err := json.MarshalIndent(h.b.NetworkLockStatus(), "", "\t")
-	if err != nil {
-		http.Error(w, "JSON encoding error", 500)
-		return
-	}
-	w.Header().Set("Content-Type", "application/json")
-	w.Write(j)
+	w.WriteHeader(200)
```
sailorfrag, line 1257: Technically I think this warrants a 204?

sailorfrag (approved):

PR #6630, opened 2022-12-06
net/tsaddr: remove ContainsFunc helpers (they're now in x/exp/slices)

bradfitz:
x/exp/slices now has ContainsFunc (golang/go#53983) so we can delete our versions.


bradfitz:
This deletes net code so is permissible on vacation. Sorry, I didn't make the rules.

(Actually I've just been waiting for that ContainsFunc thing to land and just got the email that it finally did.)

DentonGentry (approved):

PR #6642, opened 2022-12-06
ipn/ipnlocal: improve redactErr to handle more cases

andrew-d:
This handles the case where the inner *os.PathError is wrapped in another error type, and additionally will redact errors of type *os.LinkError. Finally, add tests to verify that redaction works.

Change-Id: Ie83424ff6c85cdb29fb48b641330c495107aab7c


ipn/ipnlocal/peerapi.go
```diff
@@ -341,11 +341,71 @@ func (s *peerAPIServer) DeleteFile(baseName string) error {
...
+	return re.msg
+}
+
+func (re *redactedErr) Unwrap() error {
+	return re.inner
+}
+
 func redactErr(err error) error {
-	if pe, ok := err.(*os.PathError); ok {
+	var redactStrings []string
```
dsnet, line 358: golang/go#53435 recently added the concept of a multi-error. Effectively, errors in Go are no longer a chain, but rather a tree. The use of `errors.As` unfortunately will not catch all possible cases.

In that proposal, there was some discussion about whether you can iterate or split apart a multi-error, but it was deemed both esoteric and/or unclear what the use-case would be.

This PR presents a use-case for such a feature.

ipn/ipnlocal/peerapi.go
```diff
@@ -341,11 +341,71 @@ func (s *peerAPIServer) DeleteFile(baseName string) error {
...
+		redactStrings = append(redactStrings, le.New, le.Old)
+		le.New = redacted
+		le.Old = redacted
+	}
+
+	if len(redactStrings) == 0 {
+		return err
+	}
+
+	s := err.Error()
```
dsnet, line 400: Hopefully `len(redactStrings)` isn't large, otherwise this is expensive.

ipn/ipnlocal/peerapi.go
```diff
@@ -341,11 +341,71 @@ func (s *peerAPIServer) DeleteFile(baseName string) error {
...
+
 func redactErr(err error) error {
-	if pe, ok := err.(*os.PathError); ok {
+	var redactStrings []string
+
+	var pe *os.PathError
+	if errors.As(err, &pe) {
+		// If this is the root error, then we can redact it directly.
+		if err == pe {
+			pe.Path = redacted
```
dsnet, line 364: Would it be worth appending some hash of the original value so that we can distinguish between them?

ipn/ipnlocal/peerapi.go
```diff
@@ -341,11 +341,71 @@ func (s *peerAPIServer) DeleteFile(baseName string) error {
...
+			return pe
+		}
+
+		// Otherwise, we have a *PathError somewhere in the error
+		// chain, and we can't redact it because something later in the
+		// chain may have cached the Error() return already (as
+		// fmt.Errorf does).
+		//
+		// Add this path to the set of paths that we will redact, below.
+		redactStrings = append(redactStrings, pe.Path)
```
dsnet, line 374: Yuck, gross. It's unfortunate that `fmt.Errorf` caches the error string.

dsnet (approved):

ipn/ipnlocal/peerapi.go
andrew-d: I wasn't able to find any case where we'd have more than two entries in the list, since I don't see any cases where an `*os.PathError` or `*os.LinkError` wrap either of those types. In the multi-error case, it's possible that this could get expensive with multiple distinct `*os.PathError`s wrapped into a single overarching error, at which point we would probably want to revisit this.

ipn/ipnlocal/peerapi.go
andrew-d: Yeah, I was puzzled and then 😵 when I figured out why the tests were failing despite redacting the underlying `PathError` 😢

dsnet: I feel bad since I reviewed that code when it was added to `fmt`. I should have thought about this use-case.

ipn/ipnlocal/peerapi.go
andrew-d: Yeah, good idea; perhaps something like `hash/maphash.Hash.Sum64()` with a constant per-process `Seed`? Thoughts?

ipn/ipnlocal/peerapi.go
andrew-d: All good 😃 This is frankly a pretty niche use-case, TBQH.

ipn/ipnlocal/peerapi.go
dsnet: I was thinking something simpler like just `adler32.Checksum` so that it's deterministic across nodes.

ipn/ipnlocal/peerapi.go
dsnet: I sent out #6643, which could augment this PR.

andrew-d:
CIFuzz test failure is unrelated, so going to merge

PR #6643, opened 2022-12-06
util/multierr: add Range

dsnet:
Errors in Go are no longer viewed as a linear chain, but a tree.
See golang/go#53435.

Add a Range function that iterates through an error in a pre-order, depth-first order.
This matches the iteration order of errors.As in Go 1.20.

### DIFF
--- a/util/multierr/multierr.go
+++ b/util/multierr/multierr.go
@@ -9,6 +9,8 @@ package multierr
 import (
 	"errors"
 	"strings"
+
+	"golang.org/x/exp/slices"
 )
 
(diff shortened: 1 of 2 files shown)


dsnet:
Hmmm... bleh. The Go 1.19 vet checker is complaining about the "wrong" signature of `Unwrap`.

andrew-d (approved):
LGTM! Approving now; I guess we can merge after the Go 1.20 upgrade, since I don't see a good way to get `go vet` to ignore this specific error?

andrew-d (approved):

Issue #6433, opened 2022-11-21
Tailscale doesn't remove its Ethernet Adapter and Windows cannot remove it either


MickPBduece:
### What is the issue?

Uninstalled Tailscale app and experience network issues on device software was removed from,  Tried to remove adaper and was not able to errir 0x00071779.  Tried reinstalling to remove the connection but no  joy,  I was able to disable it from powershell but still unable to remove it. 

Windows 10  
Reference "Tailscale installer not installing interface once interface got uninstalled manually (on Windows 10) #1327"

### OS

Windows

### Tailscale version

The leaky one

bradfitz:
Can you include a version or bug report or Tailscale IP? And this is windows 10? Which exact version?

Cc @dblohm7

bradfitz:
Dup of #3834, #4944, etc?

MickPBduece:
Def not like 4944.  Maybe 3834 but a lot more like  6433.

I've moved on for now.  I'll be back when the clients are a bit more sorted
so no screaming will happen if it's closed - Just want to note that there
was a gap in the removal process.

MickPBduece:
Win 10.

Tailscale v1.32.1 to the best of my knowledge. I removed it before sending
the email and removed it from Tailscale.com as well.

PR #6645, opened 2022-12-06
cmd/tailscaled: add a special command to tailscaled's Windows service…

dblohm7:
… for removing WinTun

WinTun is installed lazily by tailscaled while it is running as LocalSystem. Based upon what we're seeing in bug reports and support requests, removing WinTun as a lesser user may fail under certain Windows versions, even when that user is an Administrator.

By adding a user-defined command code to tailscaled, we can ask the service to do the removal on our behalf while it is still running as LocalSystem.

* The uninstall code is basically the same as it is in corp;
* The command code will be sent as a service control request and is protected by the SERVICE_USER_DEFINED_CONTROL access right, which requires Administrator.

I'll be adding follow-up patches in corp to engage this functionality.

Updates https://github.com/tailscale/tailscale/issues/6433


cmd/tailscaled/tailscaled_windows.go
```diff
@@ -184,6 +191,17 @@ func (service *ipnService) Execute(args []string, r <-chan svc.ChangeRequest, ch
 				syslogf("Service session change notification")
 				handleSessionChange(cmd)
 				changes <- cmd.CurrentStatus
+			case cmdUninstallWinTun:
+				syslogf("Stopping tailscaled child process and uninstalling WinTun")
+				childDoneCh := doneCh
+				doneCh = make(chan struct{})
```
maisem, line 197: This channel seems unused, except the close in the go routine below

sailorfrag: I think this needs some more comments to explain the flow, but from what I can tell this is what's happening:
1. The existing `doneCh` that will be closed when the subprocess exits is saved in `childDoneCh`.
2. `doneCh` is replaced with a new channel so this event loop will continue even after the subprocess is gone.
3. `cancel()` triggers shutting down the subprocess
4. A new goroutine finishes the wintun uninstall process, the current goroutine goes back to the service event loop.
5. The new goroutine waits for the subprocess to finish exiting, then it uninstalls wintun, followed by closing `doneCh` so the service event loop exits.

I think this looks correct, although maybe there should be a service status update indicating that it's in the process of shutting down?

dblohm7: Will add comments. I was a bit iffy on whether to change the status to StopPending when receiving the event, but I think you've convinced me.

Issue #6629, opened 2022-12-06
Tailscale kubernetes deployment is broken in prod with latest change (k8s)


opqpop:
### What is the issue?

Our Tailscale Kubernetes deployment has been working for the past few days, but has suddenly broken on its own w/o us having done anything. 

Error:

<img width="915" alt="image" src="https://user-images.githubusercontent.com/5566843/205791835-0394874d-b20f-4d66-a152-28e3b0ee8727.png">

"storing device ID in kube secret", "got non-200 status code 403"

There are a couple recent changes that happened ~28d ago from @danderson:
- https://github.com/tailscale/tailscale/blame/4b34c88426de324255f0906db8ae6ecb6da5d107/cmd/containerboot/kube.go#L72
- https://github.com/tailscale/tailscale/commit/bf2d3cd07471cb0501343f694594b3e0b0e3db18

Do you know if these might be related? 

Our k8 container (deployed on Google Kubernetes Engine):

```
        - name: ts-sidecar
          imagePullPolicy: Always
          image: "ghcr.io/tailscale/tailscale:latest"
          env:
            # Store the state in a k8s secret
            - name: TS_KUBE_SECRET
              value: "tailscale-auth-dracula"
            - name: TS_USERSPACE
              value: "false"
            - name: TS_AUTH_KEY
              valueFrom:
                secretKeyRef:
                  name: tailscale-auth-dracula
                  key: TS_AUTH_KEY
                  optional: true
          securityContext:  
            capabilities:
              add:
                - NET_ADMIN
```

What we've tried
- tried using `tailscale/tailscale:stable` docker image but still failed

Any thoughts on how we can further debug this? Ideally we can figure out which of the lines is raising this error, and even better, what `kubectl` commands we can run to repro the actual error to debug.

```

// storeDeviceID writes deviceID into the "device_id" data field of
// the kube secret secretName.
func storeDeviceID(ctx context.Context, secretName, deviceID string) error {
	// First check if the secret exists at all. Even if running on
	// kubernetes, we do not necessarily store state in a k8s secret.
	req, err := http.NewRequest("GET", fmt.Sprintf("/api/v1/namespaces/%s/secrets/%s", kubeNamespace, secretName), nil)
	if err != nil {
		return err
	}
	resp, err := doKubeRequest(ctx, req)
	if err != nil {
		if resp != nil && resp.StatusCode >= 400 && resp.StatusCode <= 499 {
			// Assume the secret doesn't exist, or we don't have
[...]

PR #6646, opened 2022-12-06
cmd/containerboot: check that k8s secret permissions are correct.

danderson:
Updates #6629.

### DIFF
--- a/cmd/containerboot/kube.go
+++ b/cmd/containerboot/kube.go
@@ -23,6 +23,54 @@ import (
 	"time"
 )
 
+// checkSecretPermissions checks that the current pod has permission to read,
+// write and patch secretName in the cluster.
+func checkSecretPermissions(ctx context.Context, secretName string) error {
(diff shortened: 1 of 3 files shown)


cmd/containerboot/kube.go
```diff
@@ -23,6 +23,54 @@ import (
...
+		var res struct {
+			Status struct {
+				Allowed bool `json:"allowed"`
+			} `json:"status"`
+		}
+		if err := json.Unmarshal(bs, &res); err != nil {
+			return err
+		}
+		if !res.Status.Allowed {
+			return fmt.Errorf("missing permission: cannot %s secret %q", "patch", secretName)
```
DentonGentry, line 68: I suspect the "patch" constant is supposed to be `verb`

danderson: Oops, the sedimentary layers of iteration are poking through. Fixed.

DentonGentry (approved):

PR #6653, opened 2022-12-07
syncs,ipn/ipnserver: add and use WakeGroup

maisem:


bradfitz (commented):
You're ignoring the caller's wait context. It can block forever when the caller passed a 1 sec timeout.

bradfitz:
Closing for now. Reopen if you choose to resurrect this?

PR #6656, opened 2022-12-07
Backport cmd/containerboot fixes.

danderson:
Also includes some refactors that happened in 1.35, which are semantic no-ops but cause merge sadness with the bugfixes.


DentonGentry (approved):
I appreciate the amount of tests in this PR.

PR #6658, opened 2022-12-07
cmd/containerboot: switch to IPN bus monitoring instead of polling.

### LINKED ISSUE #6657: Teach cmd/containerboot to record node FQDN in kubernetes.

To enable some new k8s use cases, cmd/containerboot optionally records the ID of the authed tailscale device in tailscale's k8s state Secret. We now also need it to track and record the node's FQDN, and update it as it changes (e.g. when an admin renames the node in the admin panel).

To do this, containerboot's implementation needs to change from a startup polling loop, into continuous monitoring of the daemon's state.

### How should we solve this?

_No response_

_No response_

### Anything else?

_No response_

danderson:
We still have to shell out to `tailscale up` because the container image's API includes "arbitrary flags to tailscale up", unfortunately. But this should still speed up startup a little, and also enables k8s-bound containers to update their device information as new netmap updates come in.

Fixes #6657

### DIFF
--- a/cmd/containerboot/main.go
+++ b/cmd/containerboot/main.go
@@ -60,7 +60,8 @@ import (
 
 	"golang.org/x/sys/unix"
 	"tailscale.com/client/tailscale"
-	"tailscale.com/ipn/ipnstate"
+	"tailscale.com/ipn"
+	"tailscale.com/util/deephash"
 )
 
 func main() {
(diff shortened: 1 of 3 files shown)


cmd/containerboot/main.go
```diff
@@ -152,46 +153,148 @@ func main() {
...
+					log.Fatalf("failed to auth tailscale: %v", err)
+				}
+			case ipn.NeedsMachineAuth:
+				log.Printf("machine authorization required, please visit the admin panel")
+			case ipn.Running:
+				// Technically, all we want is to keep monitoring the bus for
+				// netmap updates. However, in order to make the container crash
+				// if tailscale doesn't initially come up, the watch has a
+				// startup deadline on it. So, we have to break out of this
+				// watch loop, cancel the watch, and watch again with no
```
maisem, line 211: Incomplete sentence

danderson:
CIFuzz failure is some runtime error in the fuzz infrastructure, not in the change.

cmd/containerboot/main.go
```diff
@@ -152,46 +153,148 @@ func main() {
...
+		if n.State != nil {
+			switch *n.State {
+			case ipn.NeedsLogin:
+				if err := authTailscale(); err != nil {
+					log.Fatalf("failed to auth tailscale: %v", err)
+				}
+			case ipn.NeedsMachineAuth:
+				log.Printf("machine authorization required, please visit the admin panel")
+			case ipn.Running:
+				// Technically, all we want is to keep monitoring the bus for
```
maisem, line 207: Would it instead make sense to run this in a goroutine and have that push the relevant notifications to the main goroutine? That way you can keep the cancel the watcher goroutine if something fails or times out?

maisem (approved):

cmd/containerboot/main.go
danderson: Possibly. I was trying to avoid channels and goroutines, because it's already quite tricky to sequence things when it's just ~1 linear state, and as long as the IPN watch is active you have to consume the updates without blocking. Splitting the logic would force me to reimplement a chunk of ipn/ipnlocal to coalesce changes and track overall state.

I'll give it a try in a followup change, getting this one to pass tests was tricky enough that I don't want to lose it.

maisem: Makes sense, thanks!

PR #6660, opened 2022-12-07
tsnet: allow users to expose HTTPS port over Funnel

Xe:
Signed-off-by: Xe Iaso <xe@tailscale.com>


tsnet/example/tsnet-funnel/tsnet-funnel.go
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// The tshello server demonstrates how to use Tailscale as a library.
```
bradfitz, line 5: wrong date, wrong package doc

tsnet/tsnet.go
```diff
@@ -504,6 +504,91 @@ func (s *Server) APIClient() (*tailscale.Client, error) {
...
 }
 
+// ExposeHTTPS returns a HTTPS listener that is exposed over Funnel.
+// It will start the server if it has not been started yet.
+func (s *Server) ExposeHTTPS() (net.Listener, error) {
+	if err := s.Start(); err != nil {
+		return nil, err
+	}
+
+	ln, err := net.Listen("tcp", "[::1]:42069")
```
bradfitz, line 514: ??

tsnet/tsnet.go
```diff
@@ -504,6 +504,91 @@ func (s *Server) APIClient() (*tailscale.Client, error) {
 	return c, nil
 }
 
+// ExposeHTTPS returns a HTTPS listener that is exposed over Funnel.
+// It will start the server if it has not been started yet.
+func (s *Server) ExposeHTTPS() (net.Listener, error) {
```
bradfitz, line 509: shouldn't it return the public URL or something?

tsnet/tsnet.go
```diff
@@ -504,6 +504,91 @@ func (s *Server) APIClient() (*tailscale.Client, error) {
...
+
+	lc, err := flw.s.LocalClient()
+	if err != nil {
+		return err
+	}
+
+	if err := lc.SetServeConfig(context.Background(), &ipn.ServeConfig{
+		TCP:         map[uint16]*ipn.TCPPortHandler{},
+		Web:         map[ipn.HostPort]*ipn.WebServerConfig{},
+		AllowFunnel: map[ipn.HostPort]bool{},
```
bradfitz, line 584: let's wait and do the ephemeral mapping thing first.

tsnet/tsnet.go
shayne: Please excuse us as this PR is very draft. I asked @Xe to post the PR as a draft so I could check it out and look into a peerapi issue. It's borked when using tsnet.

shayne:
OK, I got it working. We have some work to do to fix it, but it's no longer a (complete) mystery. I'm including _temporary workarounds_ so you can move forward while we discuss long-term solutions.

### First issue

In `tsnet.go`, we set `ForwardTCPIn` on the netstack to handle forwarding inbound TCP connections. The implementation in `tsnet.go` closes the connection for anything that's not in the `tsnet.Server` listeners. In this example, it's ":443". This caused the peerapi traffic to fall on the floor.

**[Workaround]** Comment out the line below:
https://github.com/tailscale/tailscale/blob/9d335aabb28477f3f97adfebde7c37e9bfd8cdcf/tsnet/tsnet.go#L319
Here's the implementation that closes the connections:
https://github.com/tailscale/tailscale/blob/9d335aabb28477f3f97adfebde7c37e9bfd8cdcf/tsnet/tsnet.go#L443-L458

### Second issue

Now that the funnel-ingress-node can talk to the peerapi, we hit a wall. `peerAPIHandler.canIngress()` will return false. The "peer" reportedly doesn't have the capability.

What's interesting here is that the `h.remoteAddr.Addr()` in `peerAPIHAndler.peerHasCap()` is `127.0.0.1` ?! I'm afraid that's not right. The remoteAddr should be the remote src IP of the funnel-ingress-node. Are we losing the src IP somewhere in the netstack?

https://github.com/tailscale/tailscale/blob/9d335aabb28477f3f97adfebde7c37e9bfd8cdcf/ipn/ipnlocal/peerapi.go#L918-L930

**[Workaround]** Comment out the lines below:
https://github.com/tailscale/tailscale/blob/9d335aabb28477f3f97adfebde7c37e9bfd8cdcf/ipn/ipnlocal/peerapi.go#L741-L745

### Third issue

This one is just how the `ipn.ServeConfig` was set up. There was a TCPForward and HTTPS simultaneously in the `ipn.TCPPortHandler`.

https://github.com/tailscale/tailscale/blob/cf0f980e7b8c7f5f989c4c2b8b1edd9d229c62d4/tsnet/tsnet.go#L534-L545

**Fix** Simply remove the `TCPForward` field on line 537.

There may have been confusion around how Funnel works and where port 443 comes into play. Here's a sequence diagram:

```mermaid
sequenceDiagram
    Note left of ingress-node: https://<public-ip>:443/ (TLS+SNI)
    ingress-node->>tsnet-funnel.go: peerapi v0/ingress
    tsnet-funnel.go->>tsnet-funnel.go: can peer ingress?
    ingress-node-->tsnet-funnel.go: hijack peerapi request => bi-directional communication
    Note right of tsnet-funnel.go: TLS termination happens in-process, + proxy HTTP
```

---

### Last but not least 🪄, here's Funnel traffic being served to an ephemeral node running on my tailnet accessed over the scary Internet:

image

jordan-rash:
This work is super exciting.  I have created a a [wasmcloud](https://wasmcloud.com/docs/app-dev/create-provider/) capability provider that exposes an HTTP Server over tailnet and the power is just amazing.  Looking forward for this work to land!!! :-)

PR #6662, opened 2022-12-07
fix todo: eval all 2xx status codes as success in client

ajmarcus:
Signed-off-by: Ariel Marcus <ajmarcus@gmail.com>

Not sure if this was still something worth addressing but noticed the following `TODO` in the client code:

```
// TODO: Change the check for the StatusCode to include other 2XX success codes.
```

Did some research that led me to this golang [proposal](https://github.com/golang/go/issues/29652) and old-ish [repo](https://github.com/bradfitz/exp-httpclient/blob/master/problems.md#too-easy-to-not-check-return-status-codes) that explores some ergonomic fixes to golang's http client.

Totally fine taking a different approach if this isn't a good solution. :)

### DIFF
--- a/client/tailscale/acl.go
+++ b/client/tailscale/acl.go
@@ -13,6 +13,8 @@ import (
 	"fmt"
 	"net/http"
 	"net/netip"
+
+	"tailscale.com/util/httpclient"
 )
 
 // ACLRow defines a rule that grants access by a set of users or groups to a set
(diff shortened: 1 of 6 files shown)


client/tailscale/acl.go
```diff
@@ -13,6 +13,8 @@ import (
 	"fmt"
 	"net/http"
 	"net/netip"
+
+	"tailscale.com/util/httpclient"
```
bradfitz, line 17: I don't think we need a whole new package for this. `httpclient` as a new package is quite the ambitious name for a single func. :)

Can you just add a unexported "func is2xx" func?

```go
func is2xx(r *http.Response) bool { return 200 <= r.StatusCode && r.StatusCode <= 299 }
```

ajmarcus: Yep! And agreed. 😊

Issue #4311, opened 2022-03-28
tsnet: make it easier / more obvious to get your own IP address


josharian:
### What are you trying to do?

I started a tsnet instance, and wanted to get my own IP address. I started by looking at the *tsnet.Server, and didn't see any relevant methods. I then tried using tailscale.StatusWithoutPeers, but that didn't give me an IP address, for no clear reason; tailscale.Status did, although in some cases it required looping.

### How should we solve this?

I'd like a tsnet.Server method that takes a context and then blocks until it returns either IP addresses or an error.

### What is the impact of not solving this?

Bit harder to get up and running.

PR #6670, opened 2022-12-08
Get own IP when there are no peers

clairew:
Fixes [#4311](https://github.com/tailscale/tailscale/issues/4311)

When getting status without peers, the tailscale ips key should still be populated with the tsnet server's self ip

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -570,6 +570,27 @@ func (b *LocalBackend) UpdateStatus(sb *ipnstate.StatusBuilder) {
 func (b *LocalBackend) updateStatus(sb *ipnstate.StatusBuilder, extraLocked func(*ipnstate.StatusBuilder)) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
+	sb.MutateSelfStatus(func(ss *ipnstate.PeerStatus) {
+		ss.Online = health.GetInPollNetMap()
+		if b.netMap != nil {
+			ss.InNetworkMap = true
+			ss.HostName = b.netMap.Hostinfo.Hostname
+			ss.DNSName = b.netMap.Name
+			ss.UserID = b.netMap.User
+			if sn := b.netMap.SelfNode; sn != nil {
+				peerStatusFromNode(ss, sn)
+				if c := sn.Capabilities; len(c) > 0 {
+					ss.Capabilities = append([]string(nil), c...)
+				}
+			}
+		} else {
+			ss.HostName, _ = os.Hostname()
+		}
+		for _, pln := range b.peerAPIListeners {
+			ss.PeerAPIURL = append(ss.PeerAPIURL, pln.urlStr)
+		}
+	})
+	// MutateSelfStatus is called before MutateStatus, so that way the self node's tailscale ips can be assigned to ipn state's tailscale ips.
 	sb.MutateStatus(func(s *ipnstate.Status) {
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/state_test.go
```diff
@@ -291,6 +292,47 @@ func (cc *mockControl) UpdateEndpoints(endpoints []tailcfg.Endpoint) {
 // network delays) are just ignored for now, which makes the test
 // predictable, but maybe a bit less thorough. This is more of an overall
 // state machine test than a test of the wgengine+magicsock integration.
+func TestStatusWithoutPeers(t *testing.T) {
```
clairew, line 295: I added the test here to make use of the mock control server helper methods - but we may want to refactor and put these methods in a helper library.

maisem: You don't need to be in the same file for the helpers to be accessible, so you can put the test in local_test.go and still use these.

ipn/ipnlocal/local.go
```diff
@@ -660,16 +661,10 @@ func (b *LocalBackend) populatePeerStatusLocked(sb *ipnstate.StatusBuilder) {
...
 		}
-		var tailscaleIPs = make([]netip.Addr, 0, len(p.Addresses))
-		for _, addr := range p.Addresses {
-			if addr.IsSingleIP() && tsaddr.IsTailscaleIP(addr.Addr()) {
-				tailscaleIPs = append(tailscaleIPs, addr.Addr())
-			}
-		}
 		ps := &ipnstate.PeerStatus{
 			InNetworkMap: true,
+			TailscaleIPs: nil,
```
maisem, line 666: remove the line

ipn/ipnlocal/state_test.go
```diff
@@ -291,6 +292,47 @@ func (cc *mockControl) UpdateEndpoints(endpoints []tailcfg.Endpoint) {
...
+	b.Login(nil)
+	cc.send(nil, "", false, &netmap.NetworkMap{
+		MachineStatus: tailcfg.MachineAuthorized,
+		Addresses:     ipps("100.101.101.101"),
+		SelfNode: &tailcfg.Node{
+			Addresses: ipps("100.101.101.101"),
+		},
+	})
+	{
+	}
```
maisem, line 326: remove

maisem (commented):
We use `<project>: <description>` as the commit title to provide more info on the packages being modified.

Something like
`ipn/ipnlocal: fill out self.Addresses in StatusWithoutPeers

bradfitz (commented):
I'm working on trying to understand this and how the bug came to be.

bradfitz:
Okay, I think I understand now.

The idea with the UpdateStatus stuff is that there are a bunch of subsystems that can all participate in building it out: LocalBackend, the wgengine.Engine (userspace engine), magicsock.Conn. But with the localapi StatusWithoutPeers path, the magicsock.Conn never gets a chance to make its edit. That's the real bug. I don't think we should try to make LocalBackend paper over that by fixing it up.

Rather than changing the Engine interface's UpdateStatus method, how about we add a new method like `WantPeers() bool` on `StatusBuilder`, and then `magicsock.Conn` can check that before doing its `c.peerMap.forEachEndpoint` call.

Then LocalBackend's Status-vs-StatusWithoutPeers can be made more consistent to go down similar paths. The only difference would be when they make their StatusBuilder whether they set that bool.

Issue #6639, opened 2022-12-06
SSH client not working on Windows due to invalid pipe path


Mansarde:
### What is the issue?

On Windows, connecting to one of my Tailnet machines via an SSH client (e.g. PuTTY) works without any problems.

But when I try to connect to that machine via the `tailscale ssh` command though, then I get the following error message:

> failed to connect to local tailscaled (which appears to be running as tailscaled.exe, pid 21916). Got error: Get "http://local-tailscaled.sock/localapi/v0/status": open \\\\\\\\.\\\\pipe\\\\ProtectedPrefix\\\\Administrators\\\\Tailscale\\\\tailscaled: **The specified path is invalid.**

&nbsp;
It can't be the case that the path doesn't exist, because the error message for a non-existing path would be `The system cannot find the file specified.`.
I confirmed this by verifying that the pipe does indeed exist on my system (via the PowerShell command `(Get-ChildItem \\.\pipe\).FullName`).

&nbsp;
The only thing I can imagine is that the escaped backslashes of the pipe path (as shown in the error message) aren't just in the error message, but are exactly how Tailscale tried to access the pipe.

So I googled around until I found someone who had a similar problem with pipes that were escaped incorrectly:
https://github.com/rlivings39/vscode-fzf-quick-open/issues/23

&nbsp;
Could this be a similar problem here?
Perhaps `tailscale ssh` on Windows unnecessarily escapes the pipe (i.e. turning \\'s to \\\\'s) when trying to communicate with it?

### Steps to reproduce

1. Create a tailnet with one Windows machine and one Linux machine.
2. Enable/advertise SSH on the Linux machine via `tailscale up --ssh`.
3. From the windows machine, confirm that you can connect to the Linux machine via an SSH client (like PuTTY).
4. Then from a console on the Windows machine, try to connect via the `tailscale ssh` command (e.g.: `tailscale ssh root@linux-box`).

Result:
It will fail with the error message as mentioned in the isse description above.

### OS

Linux, Windows

### OS version

Windows 10 Pro 22H2 19045.2311, Ubuntu 22.04 LTS

### Tailscale version

1.34.0

### Bug report

BUG-6adf97a70915dec675da100fb304a9b92866b88058952bb3deac5c87fc1296c2-20221206182828Z-3123bcfffdccfc00

floogulinc:
This appears to have also broken Caddy TLS integration:
```json
[...]

PR #6673, opened 2022-12-08
cmd/tailscale/cli: add workaround for improper named socket quoting

raggi:
This avoids the issue in the common case where the socket path is the default path, avoiding the immediate need for a Windows shell quote implementation.

Updates #6639


bradfitz (approved):

bradfitz:
Get [ssh] into the git commit message subject?

PR #6682, opened 2022-12-09
cmd/tailscaled: change Windows implementation to shut down subprocess…

dblohm7:
… via closing its stdin

We've been doing a hard kill of the subprocess, which is only safe as long as both the cli and gui are not running and the subprocess has had the opportunity to clean up DNS settings etc. If unattended mode is turned on, this is definitely unsafe.

I changed babysitProc to close the subprocess's stdin to make it shut down, and then I plumbed a cancel function into the stdin reader on the subprocess side.

Fixes https://github.com/tailscale/corp/issues/5621

### DIFF
--- a/cmd/tailscaled/tailscaled_windows.go
+++ b/cmd/tailscaled/tailscaled_windows.go
@@ -25,6 +25,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"log"
 	"net/netip"
 	"os"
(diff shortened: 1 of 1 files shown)


bradfitz:
This seems fine, but is this the whole fix?

If we e.g. crash or shut down hard (poweroff), then those rules persist, no?

Shouldn't we also clear any NRPT rules pointing to 100.100.100.100 on tailscaled service start-up at least? (that is, clean up things from a previous run of ourselves, like the systemd `--cleanup` flag it passes to tailscaled?)

bradfitz (approved):

PR #6688, opened 2022-12-09
Avoid subshell for variable outputs in docs/k8 Makefile

VeryStrongFingers:
docs/k8s Makefile has been updated to output variables.

Prior to this change bash-like shells would interpret `$(...)` as a subshell and essentially yield an empty string. I'm presuming this was originally intended to be curly braces?
```
zsh: command not found: TS_KUBE_SECRET
zsh: command not found: TS_DEST_IP
zsh: command not found: SA_NAME
```


bradfitz:
cc @maisem @danderson

PR #6690, opened 2022-12-09
net/tlsdial,tstest,version: use go command from $PATH

### LINKED ISSUE #6689: Simplify tests use of GOROOT
Alpine has GOROOT test changes:
https://git.alpinelinux.org/aports/tree/community/tailscale/findgo.patch

But in Go 1.19:

https://tip.golang.org/doc/go1.19#go-command

> go test and go generate now place GOROOT/bin at the beginning of the PATH used for the subprocess, so tests and generators that execute the go command will resolve it to same GOROOT.

So we can probably simplify our tests.

cc @raggi

raggi:
Go now includes the GOROOT bin directory in $PATH while running tests and generate, so it is no longer necessary to construct a path using runtime.GOROOT().

Fixes #6689


bradfitz (commented):
No need for must. Just "go" no?

raggi:
> No need for must. Just "go" no?

Yup

bradfitz (approved):

PR #6692, opened 2022-12-09
all: update golang.zx2c4.com/wireguard to github.com/tailscale/wiregu…

jwhited:
…ard-go

This is temporary while we work to upstream performance work in https://github.com/WireGuard/wireguard-go/pull/64. A replace directive is less ideal as it breaks dependent code without duplication of the directive.


raggi (approved):
You'll need to run `go generate`, but otherwise LGTM

PR #6700, opened 2022-12-10
go.mod: run "go mod tidy" to add missing entries

bradfitz:
Upstream wireguard wasn't entirely removed from the rename in ea5ee6f87c15fa6c069c5cb942adb6af0798e62b; Windows still uses it for winipcfg.


DentonGentry (approved):
I don't understand why these aren't labeled "indirect" if the dependency is via winipcfg.

bradfitz:
ah, sorry, dirty local tree false alarm. I had some debug tools sitting in my cmd/ directory that used wireguard.

bradfitz:
package golang.zx2c4.com/wireguard/windows/tunnel/winipcfg is from a separate module, golang.zx2c4.com/wireguard/windows.

Issue #502, opened 2020-06-23
Feature Request: Kubernetes Operator


toonsevrin:
I may implement this myself, so consider this a tracker.

Essentially I want to be able to locally access kubernetes services (in an already secured cluster) within the security of Tailscale. We have a ton of corporate services that are currently acessed with "kubectl port-forward", using tailscale would streamline this a lot (dns and no command line).

My preference goes to having a tailscale identity (IP) per services you wish to expose to the network. This can be achieved by having an operator that watches LoadBalancer Services. If they are labelled with "tailscale: true", the operator creates an identity and a proxy deployment that terminates the tailscale traffic for that identity, routing the traffic to the actual service.

What would be the best way to handle identity in this setup?

danderson:
We're not going to work on this just yet, blocked on #504. The questions you're asking are exactly the ones we need to answer before building a Kubernetes operator, and we haven't thought about it at all yet. If you're interested in building this yourself, you're welcome to experiment with the tradeoffs and see what works best in the k8s ecosystem.

toonsevrin:
Once I do need this, I'll implement it myself :)

Can you perhaps provide a bit more detail in #504 at the issues that you are facing with containerization, the potential solutions and why they are not immediately solveable?

Thank you very much!

zfogg:
i really want to see this. please do it

bithavoc:
This feature would be absolutely amazing, imagine using Tailscale for cross-cluster resource access 🥇 

There is money to be made here Tailscale, please take my money.

niels-s:
We recently deployed Tailscale on our k8s cluster, we had to tweak our CoreDNS config a bit, but right now, when DNS is enabled, you can do something like `curl http://<svc name.<namespace>.svc.osig.lan`. 
It's pretty minimal and doesn't come with cross-cluster connection or per-service settings like requested above. My colleague @alex-richman-onesignal  is writing a blog post on it, pinging him to share the story once published.

enigmatix:
I'd be very interested to learn exactly how you did it and if there were any issues.  I'm currently working through an implementation myself @niels-s @alex-richman-onesignal

bithavoc:
@niels-s how did you get your pods to talk to the MagicDNS Server in first place?

bithavoc:
[...]

PR #6713, opened 2022-12-12
cmd/k8s-operator: add a kubernetes operator.

danderson:
This was initially developed in a separate repo, but for build/release reasons and because go module management limits the damage of importing k8s things now, moving it into this repo.

At time of commit, the operator enables exposing services over tailscale, with the 'tailscale' loadBalancerClass. It also currently requires an unreleased feature to access the Tailscale API, so is not usable yet.

Updates #502 


bradfitz (approved):
Pretty bland cmd binary name "k8s-operator". No "ts" in there?

PR #6715, opened 2022-12-12
client/tailscale: add APIs for auth key management.

danderson:
Updates #502.

### DIFF
--- a/client/tailscale/authkey.go
+++ b/client/tailscale/authkey.go
@@ -1,0 +1,142 @@
+// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package tailscale
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+)
+
+// Key represents a Tailscale auth key.
+type Key struct {
+	ID           string          `json:"id"`
+	Created      time.Time       `json:"created"`
+	Expires      time.Time       `json:"expires"`
+	Capabilities KeyCapabilities `json:"capabilities"`
+}
+
+// KeyCapabilities are the capabilities of an auth key.
+type KeyCapabilities struct {
+	Devices KeyDeviceCapabilities `json:"devices,omitempty"`
+}
+
+// KeyDeviceCapabilities are the device-related capabilities of an auth key.
+type KeyDeviceCapabilities struct {
+	Create KeyDeviceCreateCapabilities `json:"create"`
+}
+
+// KeyDeviceCreateCapabilities are the device creation capabilities of an auth key.
+type KeyDeviceCreateCapabilities struct {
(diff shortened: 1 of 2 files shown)


client/tailscale/authkey.go
```diff
@@ -0,0 +1,142 @@
...
+
+// KeyDeviceCreateCapabilities are the device creation capabilities of an auth key.
+type KeyDeviceCreateCapabilities struct {
+	Reusable      bool     `json:"reusable"`
+	Ephemeral     bool     `json:"ephemeral"`
+	Preauthorized bool     `json:"preauthorized"`
+	Tags          []string `json:"tags,omitempty"`
+}
+
+// Keys returns the list of auth keys for the current user.
```
willnorris, line 42: This endpoint will return both api keys and auth keys.  I think the only endpoint here that works exclusively with auth keys is key creation.  So we probably want to update documentation throughout (and maybe rename the file to keys.go or similar?)

client/tailscale/authkey.go
```diff
@@ -0,0 +1,142 @@
...
+	ret := make([]string, 0, len(keys))
+	for _, k := range keys {
+		ret = append(ret, k.ID)
+	}
+	return ret, nil
+}
+
+// CreateKey creates a new auth key for the current user. Returns the auth key
+// itself, which cannot be retrieved again later, and the key metadata.
+func (c *Client) CreateKey(ctx context.Context, caps KeyCapabilities) (string, *Key, error) {
```
willnorris, line 73: maybe rename to CreateAuthKey, since this really does only work for auth keys (as opposed to all of the other methods)

client/tailscale/authkey.go
danderson: Hmm, from the docs and server-side code, it looked like everything only returned authkeys, but I now see that it also supports API keys, albeit not ideally right now (e.g. returns no metadata about them). I'll update the docs and code to be about all kinds of keys.

client/tailscale/authkey.go
danderson: Could we go the other way instead and make CreateKey the universal key creation endpoint? It feels reasonable that this API could request the creation of other API keys, even if the response is "no". Right now the data structures have no way of encoding API-related capabilities in the creation request, but presumably if we ever allowed it, that's where we'd put it?

willnorris: sure, that's fine.  And maybe just document that it only works for auth keys currently?

client/tailscale/authkey.go
danderson: Adjusted to reflect that the API can (mostly) handle all key types, PTAL.

client/tailscale/authkey.go
danderson: PTAL. I've implemented what I'm proposing, which is to keep CreateKey as a generic key creation API. That way it'll keep working as we add new capabilities not related to device creation (i.e. more API-ish keys than auth-ish keys).

willnorris (approved):

PR #6716, opened 2022-12-13
ipn/ipnlocal: simplify redactErr

dsnet:
Use multierr.Range to iterate through an error tree instead of multiple invocations of errors.As.
This scales better as we add more Go error types to the switch.


andrew-d (approved):
Thanks! Much cleaner 😃

PR #6717, opened 2022-12-13
util/endian: delete package; use updated josharian/native instead

bradfitz:
See josharian/native#3

Updates golang/go#57237


DentonGentry (approved):

josharian (approved):
I wonder whether I still have the “Approve” bit)…

bradfitz:
@josharian, apparently yes! But a different color?

PR #6718, opened 2022-12-13
cmd/k8s-operator: add a first unit test.

danderson:
Testing k8s continues to be hateful, the least bad option in terms of hermeticity is to use a fake client and manually write objects, step the operator forward by one, and read+compare objects to check what the operator did.

More tests to follow to exercise other codepaths, but just the groundwork to get any test is verbose enough to merit a break and a PR.


maisem (approved):

cmd/k8s-operator/main.go
```diff
@@ -7,14 +7,10 @@
 package main
 
 import (
```
maisem, line 9: can we rename this file to `operator.go` and the test to `operator_test.go`?

danderson: Done.

PR #6728, opened 2022-12-13
net/dns: don't send on closed channel when message too large

### LINKED ISSUE #6725: Panic in `dnsTCPSession` when the request is too large
e.g.
```
dns: resolver: forward: no upstream resolvers set, returning SERVFAIL
dns: resolver: forward: no upstream resolvers set, returning SERVFAIL
dns: tcp request too large (10500 > 2048)
dns: resolver: parseQuery(00): unpacking header: id: insufficient data for base length type
panic: send on closed channel

goroutine 48924 [running]:
tailscale.com/net/dns.(*dnsTCPSession).handleQuery(0xc00088da40, {0x1522e18?, 0xc000a131d0?, 0x1?})
tailscale.com@v1.34.0/net/dns/manager.go:479 +0x1ed
created by tailscale.com/net/dns.(*dnsTCPSession).handleReads
tailscale.com@v1.34.0/net/dns/manager.go:518 +0x85
```

andrew-d:
Previously, if a DNS-over-TCP message was received while there were existing queries in-flight, and it was over the size limit, we'd close the 'responses' channel. This would cause those in-flight queries to send on the closed channel and panic.

Instead, don't close the channel at all and rely on s.ctx being canceled, which will ensure that in-flight queries don't hang.

Fixes #6725

Change-Id: I8267728ac37ed7ae38ddd09ce2633a5824320097

### DIFF
--- a/net/dns/manager.go
+++ b/net/dns/manager.go
@@ -447,8 +447,16 @@ type dnsTCPSession struct {
 
 func (s *dnsTCPSession) handleWrites() {
 	defer s.conn.Close()
-	defer close(s.responses)
 	defer s.closeCtx()
+
+	// NOTE(andrew): we explicitly do not close the 'responses' channel
+	// when this function exits. If we hit an error and return, we could
+	// still have outstanding 'handleQuery' goroutines running, and if we
+	// closed this channel they'd end up trying to send on a closed channel
+	// when they finish.
+	//
+	// Because we call closeCtx, those goroutines will not hang since they
+	// select on <-s.ctx.Done() as well as s.responses.
 
 	for {
 		select {
@@ -476,6 +484,7 @@ func (s *dnsTCPSession) handleQuery(q []byte) {
 		return
 	}
 
+	// See note in handleWrites (above) regarding this select{}
 	select {
 	case <-s.ctx.Done():
 	case s.responses <- resp:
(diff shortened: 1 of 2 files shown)


net/dns/manager.go
```diff
@@ -476,13 +484,15 @@ func (s *dnsTCPSession) handleQuery(q []byte) {
...
 
+	// See note in handleWrites (above) regarding this select{}
 	select {
 	case <-s.ctx.Done():
 	case s.responses <- resp:
 	}
 }
 
 func (s *dnsTCPSession) handleReads() {
+	defer s.conn.Close()
```
bradfitz, line 495: is this an unrelated fix?

bradfitz (approved):

net/dns/manager.go
bradfitz: Gotcha. Do we need both conn.Close boths? Looks like it's done in both read and write exit paths?

andrew-d: I think so? The same issue occurs there; if we're trying to write to a full `conn` (or a `net.Pipe` in tests), then `handleReads` can read something, bail due the size limit, and any in-progress writes can block if the client isn't reading. I guess it will only block for `idleTimeoutTCP` time, though, since we call `SetWriteDeadline`, but I'd prefer not to rely on that :smiley:

PR #6732, opened 2022-12-13
cmd/k8s-operator: sprinkle debug/info logging throughout.

danderson:
As is convention in the k8s world, use zap for structured logging. For development, OPERATOR_LOGGING=dev switches to a more human-readable output than JSON.

Updates #502

### DIFF
--- a/cmd/k8s-operator/operator.go
+++ b/cmd/k8s-operator/operator.go
@@ -380,7 +380,7 @@ func (a *ServiceReconciler) Reconcile(ctx context.Context, req reconcile.Request
 		if err := a.Status().Update(ctx, svc); err != nil {
 			return reconcile.Result{}, fmt.Errorf("failed to update service status: %w", err)
 		}
-		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		return reconcile.Result{}, nil
 	}
 
 	logger.Debugf("setting ingress hostname to %q", tsHost)
--- a/cmd/k8s-operator/operator.go
+++ b/cmd/k8s-operator/operator.go
@@ -10,11 +10,13 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
-	"log"
 	"os"
 	"strings"
 	"time"
 
+	"github.com/go-logr/zapr"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 	"golang.org/x/exp/slices"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


cmd/k8s-operator/operator.go
```diff
@@ -347,14 +374,16 @@ func (a *ServiceReconciler) Reconcile(ctx context.Context, req reconcile.Request
 		return reconcile.Result{}, fmt.Errorf("failed to get device ID: %w", err)
 	}
 	if tsHost == "" {
+		logger.Debugf("no Tailscale hostname known yet, waiting for proxy pod to finish auth")
 		// No hostname yet. Wait for the proxy pod to auth.
 		svc.Status.LoadBalancer.Ingress = nil
 		if err := a.Status().Update(ctx, svc); err != nil {
 			return reconcile.Result{}, fmt.Errorf("failed to update service status: %w", err)
 		}
-		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
skriptble, line 355: Is this cleanup from the previous PR?

cmd/k8s-operator/operator_test.go
```diff
@@ -54,7 +60,7 @@ func TestLoadBalancerClass(t *testing.T) {
 		},
 	})
 
-	expectRequeue(t, sr, "default", "test")
+	expectReconciled(t, sr, "default", "test")
```
skriptble, line 63: Same as above, was is this related to the logging or just a missed change from the previous PR?

skriptble (approved):
It looks like there are two changes unrelated to the logging, but there's only 1 commit this time. Not sure if it's worth a separate commit for those two changes though.

cmd/k8s-operator/operator.go
danderson: Cleanup found while working through the code and adding logging. The previous code was mostly harmless, just queued some pointless no-op work for later.

cmd/k8s-operator/operator_test.go
danderson: Yeah just a missed reconcile change.

cmd/k8s-operator/operator.go
skriptble: 👍🏾

cmd/k8s-operator/operator_test.go
skriptble: 👍🏾

danderson:
It was easy enough to split up, so I made this into 2 changes: the headline logging change, and a small followup for the timed reconcile thing.

PR #6745, opened 2022-12-14
net/tstun: don't return early from a partial tun.Read()

### LINKED ISSUE #6730: MTU under 518 causes a stall
### What is the issue?

An unstable build `iperf3 -M 517` will cause a stall in the send side.

This was originally reported by Ryan Huber on Twitter - thanks Ryan!

### Steps to reproduce

Tested against 44be59c15a2841231eeb7d4a5c50929b4f0afb49

`iperf3 -M 517 -c host` observe stall shortly into flow, and inability to establish new flows.

Likely 76389d8baf942b10a8f0f4201b7c4b0737a0172c but not confirmed.

### OS

Linux

### OS version

_No response_

### Tailscale version

44be59c15a2841231eeb7d4a5c50929b4f0afb49

### Bug report

_No response_

jwhited:
Fixes #6730

### DIFF
--- a/net/tstun/wrap.go
+++ b/net/tstun/wrap.go
@@ -187,8 +187,9 @@ type tunInjectedRead struct {
 // tunVectorReadResult is the result of a tun.Read(), or an injected packet
 // pretending to be a tun.Read().
 type tunVectorReadResult struct {
-	// Only one of err, data, or injected should be set, and are read in that
-	// order of precedence.
+	// When err AND data are nil, injected will be set with meaningful data
+	// (injected packet). If either err OR data is non-nil, injected should be
+	// ignored (a "real" tun.Read).
 	err      error
 	data     [][]byte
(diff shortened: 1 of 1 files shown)


sailorfrag (approved):

net/tstun/wrap.go
```diff
@@ -543,7 +544,7 @@ func (t *Wrapper) Read(buffs [][]byte, sizes []int, offset int) (int, error) {
 	if !ok {
 		return 0, io.EOF
 	}
-	if res.err != nil {
+	if res.err != nil && len(res.data) == 0 {
```
bradfitz, line 547: the next line has `res.data == nil` ... does len 0 vs nil mean different things?

jwhited: clarified in the docs on `type tunVectorReadResult`

PR #6746, opened 2022-12-14
cmd/k8s-operator: use oauth credentials for API access.

danderson:
This automates both the operator's initial login, and provisioning/deprovisioning of proxies.


willnorris (approved):

danderson:
Annoyingly this broke tsnet bringup somehow, so holding off while I debug that.

danderson:
Fixed, was missing a state machine transition that becomes needed when you want to give an authkey to tsnet after it's already started up.

PR #6748, opened 2022-12-14
ssh/tailssh: set default Tailscale SSH $PATH for non-interactive commands

### LINKED ISSUE #5285: Tailscale ssh lacks $PATH for non-interactive sessions (VSCode, Xpra, etc)
### What is the issue?

When I connect to a remote machine using VSCode Remote SSH, my $PATH gets stomped with a VSCode path which is supposed to be appended, not overwrite my entire path.

This may be a VSCode bug ([here](https://github.com/microsoft/vscode-remote-release/issues/6975)) but I'm posting here because there's definitely some interaction with tailscale. This _only_ occurs via tailscale ssh.

Works:
- SSH in terminal not via tailscale
- SSH in terminal via tailscale (tailscale up --ssh)
- SSH via VSCode Remote SSH not via tailscale

Stomps $PATH:
- SSH via VSCode Remote SSH via tailscale

### Steps to reproduce

- `tailscale up --ssh`
- In VSCode, connect to that remote host (disconnect and reconnect several times)

_No response_

### OS

Linux

### OS version

Ubuntu 20.04

### Tailscale version

1.28.0

### Bug report

_No response_

bradfitz:
Fixes #5285

Co-authored-by: @andrew-d

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -31,11 +31,16 @@ import (
 	"github.com/creack/pty"
 	"github.com/pkg/sftp"
 	"github.com/u-root/u-root/pkg/termios"
+	"go4.org/mem"
 	gossh "golang.org/x/crypto/ssh"
 	"golang.org/x/sys/unix"
 	"tailscale.com/cmd/tailscaled/childproc"
+	"tailscale.com/hostinfo"
 	"tailscale.com/tempfork/gliderlabs/ssh"
 	"tailscale.com/types/logger"
+	"tailscale.com/util/lineread"
+	"tailscale.com/util/strs"
+	"tailscale.com/version/distro"
 )
 
 func init() {
@@ -59,7 +64,14 @@ var maybeStartLoginSession = func(logf logger.Logf, ia incubatorArgs) (close fun
 //
 // If ss.srv.tailscaledPath is empty, this method is equivalent to
 // exec.CommandContext.
-func (ss *sshSession) newIncubatorCommand() *exec.Cmd {
+//
+// The returned Cmd.Env is guaranteed to be nil; the caller populates it.
+func (ss *sshSession) newIncubatorCommand() (cmd *exec.Cmd) {
+	defer func() {
+		if cmd.Env != nil {
+			panic("internal error")
+		}
+	}()
 	var (
 		name    string
 		args    []string
@@ -292,7 +304,7 @@ func (ss *sshSession) launchProcess() error {
 	} else {
 		return err
 	}
-	cmd.Env = append(cmd.Env, envForUser(ss.conn.localUser)...)
+	cmd.Env = envForUser(ss.conn.localUser)
 	for _, kv := range ss.Environ() {
 		if acceptEnvPair(kv) {
 			cmd.Env = append(cmd.Env, kv)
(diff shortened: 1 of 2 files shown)


andrew-d (approved):
LGTM; we should probably write tests for this if we make it more complicated in the future, but this is easy enough for now.

ssh/tailssh/tailssh_test.go
```diff
@@ -755,3 +755,27 @@ func TestAcceptEnvPair(t *testing.T) {
...
+			u, ""},
+	}
+
+	for i, tt := range tests {
+		got := pathFromPAMEnvLine([]byte(tt.line), tt.u)
+		if got != tt.want {
+			t.Errorf("%d. got %q; want %q", i, got, tt.want)
+		}
+	}
+
```
andrew-d, line 780: Extra newline

ssh/tailssh/tailssh_test.go
```diff
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-//go:build linux || darwin
+//go:build linux || darwin || freebsd
```
andrew-d, line 5: ???

ssh/tailssh/incubator.go
```diff
@@ -567,7 +579,60 @@ func envForUser(u *user.User) []string {
...
+func pathFromPAMEnvLine(line []byte, u *user.User) (path string) {
+	if !mem.HasPrefix(mem.B(line), mem.S("PATH")) {
+		return ""
+	}
+	rest := strings.TrimSpace(strings.TrimPrefix(string(line), "PATH"))
+	if quoted, ok := strs.CutPrefix(rest, "DEFAULT="); ok {
+		if path, err := strconv.Unquote(quoted); err == nil {
+			path = strings.NewReplacer(
+				"@{HOME}", u.HomeDir,
+				"@{PAM_USER}", u.Username,
```
andrew-d, line 627: *Technically* these can be escaped; from [here](https://man7.org/linux/man-pages/man5/pam_env.conf.5.html):

> Both the $ and @ characters can be backslash escaped to be used as literal values values

Probably fine to ignore for now, though.

andrew-d (approved):

ssh/tailssh/incubator.go
bradfitz: Happy to ignore :)

ssh/tailssh/tailssh_test.go
bradfitz: Oh, Tailscale SSH FreeBSD support was added in Tailscale 1.34 so figured it should start running tests too. I'll remove for now and add in its own commit.

ssh/tailssh/tailssh_test.go
bradfitz: done

ssh/tailssh/tailssh_test.go
```diff
@@ -755,3 +756,43 @@ func TestAcceptEnvPair(t *testing.T) {
...
+	}
+	for i, tt := range tests {
+		got := pathFromPAMEnvLine([]byte(tt.line), tt.u)
+		if got != tt.want {
+			t.Errorf("%d. got %q; want %q", i, got, tt.want)
+		}
+	}
+}
+
+func TestPathFromPAMEnvLineOnNixOS(t *testing.T) {
```
bradfitz, line 781: added this new test too to give us an early warning if NixOS changes

andrew-d: Nice; I like it!

PR #6749, opened 2022-12-15
cmd/k8s-operator: move the operator into its own namespace.

danderson:
The operator creates a fair bit of internal cluster state to manage proxying, dumping it all in the default namespace is handy for development but rude for production.

Updates #502

### DIFF
--- a/cmd/k8s-operator/manifests/operator.yaml
+++ b/cmd/k8s-operator/manifests/operator.yaml
@@ -3,16 +3,23 @@
 # license that can be found in the LICENSE file.
 
 apiVersion: v1
+kind: Namespace
+metadata:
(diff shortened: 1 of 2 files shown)


skriptble (approved):
Any particular reason you choose `tailscale-operator` instead of just `tailscale`?

danderson:
Not really. My vague thought was that k8s users already doing tailscale by hand might have put stuff in the `tailscale` namespace, but I'm planning an installer mode that lets you customize the namespace, so defaulting to `tailscale` sgtm. I'll make the change.

danderson:
Updated to make the default namespace `tailscale`, ptal?

maisem (approved):

PR #6753, opened 2022-12-15
ipn/ipnserver: propagate http.Serve error

knyar:
This ensures that we capture error returned by `Serve` and exit with a non-zero exit code if it happens.


bradfitz (approved):

andrew-d (approved):
Nice catch!

PR #6759, opened 2022-12-15
ssh/tailssh: add envknob for default PATH

bradfitz:
As backup plan, just in case the earlier fix's logic wasn't correct and we want to experiment in the field or have users have a quicker fix.

Updates #5285

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -35,6 +35,7 @@ import (
 	gossh "golang.org/x/crypto/ssh"
 	"golang.org/x/sys/unix"
 	"tailscale.com/cmd/tailscaled/childproc"
+	"tailscale.com/envknob"
 	"tailscale.com/hostinfo"
 	"tailscale.com/tempfork/gliderlabs/ssh"
 	"tailscale.com/types/logger"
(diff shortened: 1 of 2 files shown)


ssh/tailssh/incubator.go
```diff
@@ -626,19 +641,24 @@ func pathFromPAMEnvLine(line []byte, u *user.User) (path string) {
...
 }
 
+func expandDefaultPathTmpl(t string, u *user.User) string {
+	p := strings.NewReplacer(
+		"@{HOME}", u.HomeDir,
+		"@{PAM_USER}", u.Username,
+	).Replace(t)
+	if strings.Contains(p, "@{") {
+		// If there are unknown expansions, conservatively fail closed.
+		return ""
```
maisem, line 657: Return t?

maisem (approved):

ssh/tailssh/incubator.go
bradfitz: If p contains it, so will t. So they screwed up the template so we ignore it.

maisem: I was thinking that having it show up in the PATH unexpanded would be better.
However corrupting the PATH sounds worse, so this lgtm

ShadowJonathan:
Thank you, I think this'll suffice in case there's a problem

PR #6760, opened 2022-12-16
net/connstats: enforce maximum number of connections

dsnet:
The Tailscale logging service has a hard limit on the maximum log message size that can be accepted.
We want to ensure that network logging messages
never exceed this limit otherwise a client cannot transmit logs.

Move the goroutine for periodically dumping netlog messages from wgengine/netlog to net/connstats.
This allows net/connstats to manage when it dumps messages, either based on time or by size.

Updates tailscale/corp#8427

### DIFF
--- a/net/connstats/stats.go
+++ b/net/connstats/stats.go
@@ -7,9 +7,12 @@
 package connstats
 
 import (
+	"context"
 	"net/netip"
 	"sync"
+	"time"
 
+	"golang.org/x/sync/errgroup"
 	"tailscale.com/net/packet"
 	"tailscale.com/types/netlogtype"
 )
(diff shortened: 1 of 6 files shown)


net/connstats/stats.go
```diff
@@ -18,11 +21,64 @@ import (
...
+
+type connCnts struct {
+	start    time.Time
+	end      time.Time
 	virtual  map[netlogtype.Connection]netlogtype.Counts
 	physical map[netlogtype.Connection]netlogtype.Counts
 }
 
+// NewStatistics creates a data structure for tracking connection statistics
+// and periodically dumps the virtual and physical connetion counts
```
dblohm7, line 43: s/connetion/connection/

dblohm7 (approved):

PR #6767, opened 2022-12-16
cmd/containerboot: allow disabling secret storage in k8s.

### LINKED ISSUE #6704: FR:  TS_KUBE_SECRET cannot be unset. As it defaults to `tailscale` if it's `""`

Run ts in a kube environment, without the use of TS_KUBE_SECRET

### How should we solve this?

In the case that someone runs tailscale in a pod with other container(s), you can't give access to the secret to a single container on that pod, service account binds to the whole pod.

Please allow having empty TS_KUBE_SECRET, so people can set a TS_STATE_DIR of their liking and also pass TS_AUTH_KEY as they please too.

Thanks.

https://github.com/tailscale/tailscale/blob/7b65b7f85c6484cc852c201423ac577d9a4ceebf/cmd/containerboot/main.go#L40-L41

When using it in a pod with other containers, you are force to bind a Service account to tall the containers inside that pod.

### Anything else?

_No response_

danderson:
In some configurations, user explicitly do not want to store tailscale state in k8s secrets, because doing that leads to some annoying permission issues with sidecar containers. With this change, TS_KUBE_SECRET="" and TS_STATE_DIR=/foo will force storage to file when running in kubernetes.

Fixes #6704.

### DIFF
--- a/cmd/containerboot/main_test.go
+++ b/cmd/containerboot/main_test.go
@@ -347,6 +347,31 @@ func TestContainerBoot(t *testing.T) {
 			},
 		},
 		{
+			Name: "kube_disk_storage",
+			Env: map[string]string{
+				"KUBERNETES_SERVICE_HOST":       kube.Host,
+				"KUBERNETES_SERVICE_PORT_HTTPS": kube.Port,
+				// Explicitly set to an empty value, to override the default of "tailscale".
+				"TS_KUBE_SECRET": "",
+				"TS_STATE_DIR":   filepath.Join(d, "tmp"),
+				"TS_AUTH_KEY":    "tskey-key",
+			},
+			KubeSecret: map[string]string{},
+			Phases: []phase{
+				{
+					WantCmds: []string{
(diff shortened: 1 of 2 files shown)


cmd/containerboot/main_test.go
```diff
@@ -346,6 +346,30 @@ func TestContainerBoot(t *testing.T) {
 				},
 			},
 		},
+		{
+			Name: "kube_disk_storage",
+			Env: map[string]string{
+				"KUBERNETES_SERVICE_HOST":       kube.Host,
+				"KUBERNETES_SERVICE_PORT_HTTPS": kube.Port,
+				"TS_KUBE_SECRET":                "",
```
bradfitz, line 354: add comment like `// testing that it's explicitly set to an empty value` ?

bradfitz (approved):

cmd/containerboot/main_test.go
danderson: Done.

Issue #6768, opened 2022-12-16
Make logs flush delay dynamic


mihaip:
On mobile devices we default to a 2 minute delay, but it'd be nice to be more aggressive if we can (e.g. while on WiFi and charging).

PR #6769, opened 2022-12-16
logtail: make logs flush delay dynamic

mihaip:
Instead of a static FlushDelay configuration value, use a FlushDelayFn function that we invoke every time we decide send logs. This will allow mobile clients to be more dynamic about when to send logs.

Updates #6768

### DIFF
--- a/logtail/logtail.go
+++ b/logtail/logtail.go
@@ -66,13 +66,12 @@ type Config struct {
 	// that's safe to embed in a JSON string literal without further escaping.
 	MetricsDelta func() string
 
-	// FlushDelay is how long to wait to accumulate logs before
-	// uploading them.
+	// FlushDelayFn, if non-nil is a func that returns how long to wait to
+	// accumulate logs before uploading them. 0 or negative means to upload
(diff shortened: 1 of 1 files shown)


dsnet (approved):

logtail/logtail.go
```diff
@@ -179,8 +178,8 @@ type Logger struct {
 	skipClientTime bool
 	linkMonitor    *monitor.Mon
 	buffer         Buffer
-	drainWake      chan struct{} // signal to speed up drain
-	flushDelay     time.Duration // negative or zero to upload agressively, or >0 to batch at this delay
+	drainWake      chan struct{}        // signal to speed up drain
+	flushDelayFn   func() time.Duration // negative or zero return value to upload agressively, or >0 to batch at this delay
```
creachadair, line 182:
```suggestion
	flushDelayFn   func() time.Duration // negative or zero return value to upload aggressively, or >0 to batch at this delay
```

creachadair (commented):
Just a minor typo. 🙂

PR #6770, opened 2022-12-17
cmd/containerboot: allow disabling secret storage in k8s.

### LINKED ISSUE #6704: FR:  TS_KUBE_SECRET cannot be unset. As it defaults to `tailscale` if it's `""`

danderson:
In some configurations, user explicitly do not want to store tailscale state in k8s secrets, because doing that leads to some annoying permission issues with sidecar containers. With this change, TS_KUBE_SECRET="" and TS_STATE_DIR=/foo will force storage to file when running in kubernetes.

Fixes #6704.

(cherry picked from commit af3127711a6df3a920204623bb24e1d74df97f1b)

---

The issue was classed as a feature request, because this has never worked before, even pre-containerboot. However, I'm thinking of it as a bug in the sense that the containerboot documentation (in the form of the binary's doc comments) implied that this configuration _should_ work and the code was written assuming that it would.

### DIFF
--- a/cmd/containerboot/main.go
+++ b/cmd/containerboot/main.go
@@ -4,13 +4,13 @@
 
 //go:build linux
 
-// The containerboot binary is a wrapper for starting tailscaled in a
-// container. It handles reading the desired mode of operation out of
-// environment variables, bringing up and authenticating Tailscale,
-// and any other kubernetes-specific side jobs.
+// The containerboot binary is a wrapper for starting tailscaled in a container.
+// It handles reading the desired mode of operation out of environment
+// variables, bringing up and authenticating Tailscale, and any other
+// kubernetes-specific side jobs.
 //
-// As with most container things, configuration is passed through
-// environment variables. All configuration is optional.
+// As with most container things, configuration is passed through environment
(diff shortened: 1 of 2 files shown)


danderson:
Bah, other containerboot changes at head make the test patch not apply. I'll revisit this Monday and fix up the backport.

DentonGentry (approved):

danderson:
1.36 released with this patch.

Issue #6364, opened 2022-11-17
ipn/ipnlocal: local machine serve listener fails on mac GUI sandbox


bradfitz:
```
2022-11-17 07:05:33.189454 +0000 UTC: serve failed to listen on 100.126.185.78:443, backing off: listen tcp 100.126.185.78:443: bind: permission denied
2022-11-17 07:05:54.822975 +0000 UTC: serve failed to listen on 100.126.185.78:443, backing off: listen tcp 100.126.185.78:443: bind: permission denied
2022-11-17 07:06:22.08098 +0000 UTC: serve failed to listen on 100.126.185.78:443, backing off: listen tcp 100.126.185.78:443: bind: permission denied
2022-11-17 07:06:48.602011 +0000 UTC: serve failed to listen on 100.126.185.78:443, backing off: listen tcp 100.126.185.78:443: bind: permission denied
```

cc @shayne @maisem

shayne:
Doesn't look like Mac App's can listen on privileged ports. I can change the default to 8443 and error out in the for 443.

bradfitz:
@nickoneill @mihaip is there some entitlement for this?

mihaip:
As far as I can tell this should work -- we have the `com.apple.security.network.server` entitlement and as Shayne observed starting with 10.14 [you no longer need to be root](https://developer.apple.com/forums/thread/674179) to bind to privileged ports.

bradfitz:
Oh I bet we're missing the ioctl on the listener to get outside the sandbox. We do that for peerapiserver etc.

mihaip:
You can check to see if it's sandboxing-related by looking for `Sandbox``deny` messages in Console.app (https://github.com/tailscale/tailscale/issues/4518#issuecomment-1170656652 has an example of what it looks like for `ioctl`, https://developer.apple.com/forums/thread/705461?answerId=712659022#712659022 has an example for port binding.

shayne:
In researching this a bit more, I'm currently challenging the assumption that interface specific binding of privileged is still borked in macOS:
https://developer.apple.com/forums/thread/674179

I was able to hard-code `net.Listen(":443")` and alas, it works. 
![image](https://user-images.githubusercontent.com/79330/202779172-9ca50f8f-54b2-403c-b403-ee76b2149ad8.png)

shayne:
For reference, here's the relevant lines of peerAPIServer.listen:
[...]

PR #6771, opened 2022-12-17
ipn/ipnlocal: [serve] listen on all-interfaces for macOS

shayne:
On macOS (AppStore and macsys), we need to bind to ""/all-interfaces due to the network sandbox. Ideally we would only bind to the Tailscale interface, but that's not possible, macOS fails to listen on privileged ports when binding to a specific interface.

We also implement the lc.Control hook, same as we do for peerapi. It doesn't solve our problem but it's better that we do and would likely be required when Apple gets around to fixing per-interface priviliged port binding.

Fixes: #6364

### DIFF
--- a/ipn/ipnlocal/serve.go
+++ b/ipn/ipnlocal/serve.go
@@ -33,6 +33,7 @@ import (
 	"tailscale.com/types/logger"
 	"tailscale.com/util/mak"
 	"tailscale.com/util/strs"
+	"tailscale.com/version"
 )
 
 // serveHTTPContextKey is the context.Value key for a *serveHTTPContext.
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/serve.go
```diff
@@ -91,7 +92,36 @@ func (s *serveListener) Close() error {
...
+			// On iOS/macOS, this sets the lc.Control hook to
+			// setsockopt the interface index to bind to, to get
+			// out of the network sandbox.
+			if err := initListenConfig(&lc, ip, s.b.prevIfState, s.b.dialer.TUNName()); err != nil {
+				s.logf("serve failed to init listen config %v, backing off: %v", s.ap, err)
+				s.bo.BackOff(s.ctx, err)
+				continue
+			}
+			// On macOS and if we're binding to a privileged port,
+			if runtime.GOOS == "darwin" && s.ap.Port() < 1024 {
```
bradfitz, line 109: there are three mac variants. you only want to do this for 2 of the 3 probably. see the IsSandboxed funcs in the version package.

shayne: Done. Using `version.IsSandboxedMacOS()` which is the 2 of the 3 (AppStore and macsys).

ipn/ipnlocal/serve.go
```diff
@@ -91,7 +92,36 @@ func (s *serveListener) Close() error {
...
 	for {
-		ln, err := net.Listen("tcp", s.ap.String())
+		ip := s.ap.Addr()
+		ipStr := ip.String()
+
+		var lc net.ListenConfig
+		if initListenConfig != nil {
+			// On iOS/macOS, this sets the lc.Control hook to
+			// setsockopt the interface index to bind to, to get
+			// out of the network sandbox.
```
bradfitz, line 102: can you reword this "to get out of the network sandbox" to explain more what's going on and what the change in behavior is?

ipn/ipnlocal/serve.go
```diff
@@ -91,7 +92,36 @@ func (s *serveListener) Close() error {
 // Listen is retried until the context is canceled.
 func (s *serveListener) Run() {
 	for {
-		ln, err := net.Listen("tcp", s.ap.String())
+		ip := s.ap.Addr()
+		ipStr := ip.String()
+
+		var lc net.ListenConfig
+		if initListenConfig != nil {
+			// On iOS/macOS, this sets the lc.Control hook to
```
bradfitz, line 100: macOS sandbox only?

ipn/ipnlocal/serve.go
```diff
@@ -91,7 +92,36 @@ func (s *serveListener) Close() error {
...
+				s.bo.BackOff(s.ctx, err)
+				continue
+			}
+			// On macOS (AppStore or macsys) and if we're binding to a privileged port,
+			if version.IsSandboxedMacOS() && s.ap.Port() < 1024 {
+				// On macOS, we need to bind to ""/all-interfaces due to
+				// the network sandbox. Ideally we would only bind to the
+				// Tailscale interface, but macOS errors out if we try to
+				// to listen on privileged ports binding only to a specific
+				// interface. (6364)
```
bradfitz, line 114: `#6364`

bradfitz (approved):

Issue #6671, opened 2022-12-08
Android Client change server functionality broken after v1.34.0


rbangert:
### What is the issue?

Verified issue with 2 others running headscale servers. 

v1.33.97 is the last version that we have found this function to not be broken. 

Starting with v1.34.0, we are able to tap the hamburger icon several times to add the change server option, select it, enter the new address, and press restart the app, at which point it closes and never reopens, all of which mirrors the previous version's behavior.

After reopening, the change server option is still available, and the address to the new server is still in the input box, but pressing the button loads the tailscale login, and no connection attempts show on the server. 

### Steps to reproduce

Upgrade to v1.34.0 or above.

### OS

Android

### OS version

Andriod 13

### Tailscale version

1.34.0

rbangert:
It appears to have happened in this commit. 

https://github.com/tailscale/tailscale-android/blob/7cdd9abc964b1bbecf1f2134387b6416b8c15b6b/cmd/tailscale/main.go#L441-L446

mihaip:
@maisem I think this is due to the fast user switching refactoring. When the app is in the initial state (with a new profile) we don't persist any data until there's a login. The `os.Exit` at https://github.com/tailscale/tailscale-android/blob/7f11150cb6a3df6a1c36125f187b6892024fe46f/cmd/tailscale/main.go is resulting in the `ControlURL` URL pref being lost.

I think the fix here would be to do something like that what we do for the Mac app with tailscale/corp#8137 where we do the `EditPrefs` before we call `StartLoginInteractive`. Or possibly it's closer to what we do in the CLI at https://github.com/tailscale/tailscale/blob/c084c7d7edf425fa661e7543298c14a93f867c81/cmd/tailscale/cli/up.go#L657-L659, where we can provide the ControlURL as part of the `UpdatePrefs` when starting.

rbangert:
This is the block of code in main.go you are speaking of. 
https://github.com/tailscale/tailscale-android/blob/7cdd9abc964b1bbecf1f2134387b6416b8c15b6b/cmd/tailscale/main.go#L441-L446

It was added with this commit to add this functionality, and had been functioning until v1.34.0.
https://github.com/tailscale/tailscale-android/commit/d8aedf721ae17326561fa12b2699bdb19cc677f9

So the new ControlURL was getting picked up by main.go in after reopening, until these two commits: 

[...]

PR #6803, opened 2022-12-20
ipn/profiles: set default prefs based on Windows registry

kradalby:
paired with @maisem.
fixes: https://github.com/tailscale/corp/issues/8417
updates: #6671

Signed-off-by: Kristoffer Dalby <kristoffer@tailscale.com>

### DIFF
--- a/ipn/ipnlocal/profiles.go
+++ b/ipn/ipnlocal/profiles.go
@@ -324,7 +324,7 @@ func (pm *profileManager) setAsUserSelectedProfileLocked() error {
 func (pm *profileManager) loadSavedPrefs(key ipn.StateKey) (ipn.PrefsView, error) {
 	bs, err := pm.store.ReadState(key)
 	if err == ipn.ErrStateNotExist || len(bs) == 0 {
-		return emptyPrefs, nil
+		return defaultPrefs, nil
 	}
 	if err != nil {
 		return ipn.PrefsView{}, err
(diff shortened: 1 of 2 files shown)


DentonGentry:
If possible we'd like to fix this in a 1.34.2 release today.

bradfitz (approved):

ipn/ipnlocal/profiles.go
```diff
@@ -394,15 +396,29 @@ func (pm *profileManager) writeKnownProfiles() error {
...
 }
 
-// emptyPrefs is the default prefs for a new profile.
-var emptyPrefs = func() ipn.PrefsView {
+// defaultPrefs is the default prefs for a new profile.
+var defaultPrefs = func() ipn.PrefsView {
 	prefs := ipn.NewPrefs()
 	prefs.WantRunning = false
+
+	prefs.ControlURL = winutil.GetPolicyString("LoginURL", "")
```
mihaip, line 409: It seems unfortunate to have knowledge/application of policies split between the OSS and corp repos. There's some very similar code for this at https://github.com/tailscale/corp/blob/33299d68f14826899283270da8455013d56b8014/win/ui/ui.go#L387-L406 (refactored from https://github.com/tailscale/corp/commit/35735329841815a892220aec59d9c91f97c51fc4).

One alternate implementation would be to not add Windows-specific logic here, and instead have a `defaultPrefsMutatorFn` hook (a `func(prefs *ipn.Prefs)`) that can be set by the clients that use it. The Windows client could then set it with a function that does the policy (and we could also consider migrating the Mac and iOS ones to use it to respect the `LoginURL` pref key).

PR #6804, opened 2022-12-20
ipn/ipnlocal: maintain a proxy handler per backend

knyar:
By default, `http.Transport` keeps idle connections open hoping to re-use them in the future. Combined with a separate transport per request in HTTP proxy this results in idle connection leak.

I initially added simple transport caching per backend (see first commit) but then decided to keep proxy handlers per backend, which is a bit more complex given its interactions with prefs.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/http/httputil"
 	"net/netip"
 	"net/url"
 	"os"
@@ -206,8 +207,8 @@ type LocalBackend struct {
 	lastServeConfJSON mem.RO              // last JSON that was parsed into serveConfig
 	serveConfig       ipn.ServeConfigView // or !Valid if none
 
-	serveListeners       map[netip.AddrPort]*serveListener // addrPort => serveListener
-	serveProxyTransports map[string]*http.Transport        // proxy backend (HTTPHandler.Proxy) => http.Transport
+	serveListeners     map[netip.AddrPort]*serveListener // addrPort => serveListener
+	serveProxyHandlers sync.Map                          // string (HTTPHandler.Proxy) => httputil.ReverseProxy
 
 	// statusLock must be held before calling statusChanged.Wait() or
 	// statusChanged.Broadcast().
@@ -3774,6 +3775,9 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked(prefs ipn.
 			return true
 		})
 		handlePorts = append(handlePorts, servePorts...)
+
+		b.setServeProxyHandlersLocked()
+
 		// don't listen on netmap addresses if we're in userspace mode
 		if !wgengine.IsNetstack(b.e) {
 			b.updateServeTCPPortNetMapAddrListenersLocked(servePorts)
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/local.go
```diff
@@ -3788,6 +3793,46 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked(prefs ipn.
...
 }
 
+// setServeProxyHandlersLocked ensures there is an http proxy handler for each
+// backend specified in serveConfig. It expects serveConfig to be valid and
+// up-to-date, so should be called after reloadServeConfigLocked.
+func (b *LocalBackend) setServeProxyHandlersLocked() {
+	if !b.serveConfig.Valid() {
+		return
+	}
+	config := b.serveConfig.AsStruct()
```
bradfitz, line 3803: You could use the view and not do this allocation probably.

ipn/ipnlocal/local.go
```diff
@@ -3788,6 +3793,46 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked(prefs ipn.
...
+	backends := make(map[string]bool)
```
bradfitz, line 3804: You could avoid this allocation in the case of no proxy handlers by using mak.Set below. But I suppose that'll be rare-ish.

ipn/ipnlocal/local.go
```diff
@@ -3788,6 +3793,46 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked(prefs ipn.
...
+			}
+			b.serveProxyHandlers.Store(h.Proxy, p)
+		}
+	}
+	// Clean up handlers for proxy backends that are no longer present
+	// in configuration.
+	b.serveProxyHandlers.Range(func(key, value any) bool {
+		backend := key.(string)
+		if !backends[backend] {
+			b.logf("closing idle connections to %s", backend)
```
bradfitz, line 3828: "serve: closing idle ...

ipn/ipnlocal/local.go
```diff
@@ -206,7 +207,8 @@ type LocalBackend struct {
 	lastServeConfJSON mem.RO              // last JSON that was parsed into serveConfig
 	serveConfig       ipn.ServeConfigView // or !Valid if none
 
-	serveListeners map[netip.AddrPort]*serveListener // addrPort => serveListener
+	serveListeners     map[netip.AddrPort]*serveListener // addrPort => serveListener
+	serveProxyHandlers sync.Map                          // string (HTTPHandler.Proxy) => httputil.ReverseProxy
```
bradfitz, line 211: *httputil.ReverseProxy

ipn/ipnlocal/serve.go
```diff
@@ -389,6 +388,24 @@ func (b *LocalBackend) getServeHandler(r *http.Request) (_ ipn.HTTPHandlerView,
...
+// proxyHandlerForBackend creates a new HTTP reverse proxy for a particular backend that
+// we serve requests for. `backend` is a HTTPHandler.Proxy string (url, hostport or just port).
+func (b *LocalBackend) proxyHandlerForBackend(backend string) (*httputil.ReverseProxy, error) {
+	targetURL, insecure := expandProxyArg(backend)
+	u, err := url.Parse(targetURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid url %s: %w", targetURL, err)
+	}
+	rp := httputil.NewSingleHostReverseProxy(u)
+	rp.Transport = &http.Transport{
```
bradfitz, line 400: probably more we want to set, like IdleConnTimeout.

See DefaultTransport

bradfitz (approved):

ipn/ipnlocal/serve.go
knyar: Thanks, I've copied all values from `DefaultTransport` except for `ProxyFromEnvironment`. Or should we make this respect `HTTP_PROXY` et al?

shayne: The CLI only allows backends of `localhost/127.0.0.1`. While it's possible to `set-raw` a backend pointing elsewhere, it's not supported. Because of this, I don't think we want to support `HTTP_PROXY`.

shayne (approved):

PR #6808, opened 2022-12-20
wgengine/magicsock: avoid ReadBatch/WriteBatch on old Linux kernels

### LINKED ISSUE #6807: wgengine/magic: broken on older Linux kernels?
Tailscale is broken on older Synology NAS due to recvmmsg:

```
admin@ds212j:~$ tailscale  version
1.35.85
  tailscale commit: 98daf99775ea7e61ab103bb42e777ac5e80ea422
  go version: go1.19.4-tsdc0ce6324d

admin@ds212j:~$ uname -a
Linux ds212j 2.6.32.12 #25556 Thu Jul 1 14:25:48 CST 2021 armv5tel GNU/Linux synology_88f6281_212j

admin@ds212j:~$ tailscale  status
...
# Health check:
#     - ReceiveIPv4 is not running
#     - ReceiveIPv6 is not running
#     - This is an unstable (development) version of Tailscale; frequent updates and bugs are likely
```
...
```
2022-12-20 17:47:02.74448869 +0000 UTC: Routine: receive incoming v4 - started
2022-12-20 17:47:02.82383548 +0000 UTC: Routine: receive incoming v6 - started
2022-12-20 17:47:03.095672665 +0000 UTC: Failed to receive v4 packet: read udp 0.0.0.0:41641: recvmmsg: function not implemented
2022-12-20 17:47:03.096176723 +0000 UTC: Routine: receive incoming v4 - stopped
2022-12-20 17:47:03.097173345 +0000 UTC: Failed to receive v6 packet: read udp [::]:41641: recvmmsg: function not implemented
2022-12-20 17:47:03.097574948 +0000 UTC: Routine: receive incoming v6 - stopped
2022-12-20 17:47:03.174790769 +0000 UTC: Clearing router settings...
2022-12-20 17:47:03.175046079 +0000 UTC: Starting link monitor...
2022-12-20 17:47:03.175475542 +0000 UTC: Engine created.
```

/cc @jwhited @raggi @DentonGentry

bradfitz:
Fixes #6807

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -37,6 +37,7 @@ import (
 	"tailscale.com/derp/derphttp"
 	"tailscale.com/disco"
 	"tailscale.com/health"
+	"tailscale.com/hostinfo"
 	"tailscale.com/ipn/ipnstate"
 	"tailscale.com/logtail/backoff"
 	"tailscale.com/net/connstats"
@@ -3208,6 +3209,14 @@ type RebindingUDPConn struct {
 func upgradePacketConn(p nettype.PacketConn, network string) nettype.PacketConn {
 	uc, ok := p.(*net.UDPConn)
 	if ok && runtime.GOOS == "linux" && (network == "udp4" || network == "udp6") {
+		// recvmmsg/sendmmsg were added in 2.6.33 but we support down to 2.6.32
+		// for old NAS devices. See https://github.com/tailscale/tailscale/issues/6807.
+		// As a cheap heuristic: if the Linux kernel starts with "2", just consider
+		// it too old for the fast paths. Nobody who cares about performance runs such
+		// ancient kernels.
+		if strings.HasPrefix(hostinfo.GetOSVersion(), "2") {
+			return p
+		}
 		// Non-Linux does not support batch operations. x/net will fall back to
 		// recv/sendmsg, but not all platforms have recv/sendmsg support. Keep
 		// this simple for now.


bradfitz:
I'm thinking something like this?

wgengine/magicsock/magicsock.go
```diff
@@ -3310,6 +3312,10 @@ func (c *RebindingUDPConn) WriteBatch(msgs []ipv6.Message, flags int) (int, erro
 
 		n, err = bw.WriteBatch(msgs[start:], flags)
 		if err != nil {
+			if errors.Is(err, errNoSys) && !oldKernel {
```
jwhited, line 3315: it would be nice if we could detect this in `func upgradePacketConn()` somehow.

bradfitz: We could try to parse the kernel version like I just did in https://github.com/golang/sys/commit/2204b6615fb84fc69a9e5d37b2d508ddf544dcea but I kinda hate that... vendor kernels are weird cherry-pick messes of different things. I prefer to just feature test at runtime and see what we get.

But that means we'd need to make some dummy system calls on upgrade.

Maybe just parse very broadly: "Kernel version start with '2'? Sorry, no fast mode for you!"

bradfitz: e.g.
```patch
diff --git a/wgengine/magicsock/magicsock.go b/wgengine/magicsock/magicsock.go
index 5c42867d..415726d7 100644
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -37,6 +37,7 @@
        "tailscale.com/derp/derphttp"
        "tailscale.com/disco"
        "tailscale.com/health"
+       "tailscale.com/hostinfo"
        "tailscale.com/ipn/ipnstate"
        "tailscale.com/logtail/backoff"
        "tailscale.com/net/connstats"
@@ -3208,6 +3209,14 @@ type RebindingUDPConn struct {
 func upgradePacketConn(p nettype.PacketConn, network string) nettype.PacketConn {
        uc, ok := p.(*net.UDPConn)
        if ok && runtime.GOOS == "linux" && (network == "udp4" || network == "udp6") {
+               // recvmmsg/sendmmsg were added in 2.6.33 but we support down to 2.6.32
+               // for old NAS devices. See https://github.com/tailscale/tailscale/issues/6807.
+               // As a cheap heuristic: if the Linux kernel starts with "2", just consider
+               // it too old for the fast paths. Nobody who cares about performance runs such
+               // ancient kernels.
+               if strings.HasPrefix(hostinfo.GetOSVersion(), "2") {
+                       return p
+               }
                // Non-Linux does not support batch operations. x/net will fall back to
                // recv/sendmsg, but not all platforms have recv/sendmsg support. Keep
                // this simple for now.
```

jwhited: I wonder if a zero len `msgs` will get down into the syscall. Though, that may be fragile/break later if it does. Alternatively, we could move `oldKernel` up into `RebindingUDPConn` so we don't have to evaluate on every call to Read/WriteBatch.

bradfitz: Just tested this `"2"` variant.  Both versions work, but the `"2"` variant is less intrusive. PTAL.

wgengine/magicsock/magicsock.go
```diff
@@ -3208,6 +3209,14 @@ type RebindingUDPConn struct {
 func upgradePacketConn(p nettype.PacketConn, network string) nettype.PacketConn {
 	uc, ok := p.(*net.UDPConn)
 	if ok && runtime.GOOS == "linux" && (network == "udp4" || network == "udp6") {
+		// recvmmsg/sendmmsg were added in 2.6.33 but we support down to 2.6.32
+		// for old NAS devices. See https://github.com/tailscale/tailscale/issues/6807.
+		// As a cheap heuristic: if the Linux kernel starts with "2", just consider
+		// it too old for the fast paths. Nobody who cares about performance runs such
+		// ancient kernels.
+		if strings.HasPrefix(hostinfo.GetOSVersion(), "2") {
```
bradfitz, line 3217: fyi: this doesn't allocate or do a system call except on the first call

jwhited (approved):

PR #6814, opened 2022-12-20
tsnet: fix handling of non-empty hostnames in Server.Listen

### LINKED ISSUE #6815: tsnet: connections fail to listener with non-empty host field
If the `tsnet.Server.Listen` method has a non-empty hostname in its address, e.g., `hostname:80` instead of `:80`, connections accepted by that listener may not be forwarded correctly.

Repro:

- Set up a `tsnet.Server` instance, `s := &tsnet.Server{Hostname: "foo"}`.
- Call `s.Listen("tcp", "foo:8080")`
- After starting up and authenticating the node, connect to `foo:8080` (e.g., via `nc foo 8080`).

The connection is accepted (and logged as accepted), but its address is [not found in the forwarding map](https://github.com/tailscale/tailscale/blob/main/tsnet/tsnet.go#L445) and the connection will be closed.

Calling `Listen` without the hostname works fine. At minimum, we should document this constraint, and perhaps also make it report an error.

creachadair:
tsnet: fix handling of non-empty hostname in Server.Listen

If Listen is called with a non-empty hostname, packets will not correctly be
forwarded. Document that the hostname must be empty, and report an error if it
is not.

Since the hostname is now definitely empty, omit it from the listener's lookup
plumbing.

Moreover, since we only support TCP protocols, that does not need to be part of
the lookup key either.

Fixes #6815.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -102,7 +102,7 @@ type Server struct {
 	logtail          *logtail.Logger
 
 	mu        sync.Mutex
-	listeners map[listenKey]*listener
+	listeners map[tcpListenKey]*listener
 	dialer    *tsdial.Dialer
 }
 
@@ -441,7 +441,7 @@ func (s *Server) printAuthURLLoop() {
 
 func (s *Server) forwardTCP(c net.Conn, port uint16) {
 	s.mu.Lock()
-	ln, ok := s.listeners[listenKey{"tcp", "", port}]
+	ln, ok := s.listeners[tcpListenKey{port: port}]
 	s.mu.Unlock()
 	if !ok {
 		c.Close()
@@ -517,6 +517,7 @@ func (s *Server) APIClient() (*tailscale.Client, error) {
 
 // Listen announces only on the Tailscale network.
 // It will start the server if it has not been started yet.
+// The addr must not include a hostname, only a port, for example ":80".
 func (s *Server) Listen(network, addr string) (net.Listener, error) {
 	switch network {
 	case "", "tcp", "tcp4", "tcp6":
@@ -527,6 +528,9 @@ func (s *Server) Listen(network, addr string) (net.Listener, error) {
 	if err != nil {
 		return nil, fmt.Errorf("tsnet: %w", err)
 	}
+	if host != "" {
+		return nil, errors.New("tsnet: a non-empty hostname is not supported")
+	}
 	port, err := net.LookupPort(network, portStr)
 	if err != nil || port < 0 || port > math.MaxUint16 {
 		return nil, fmt.Errorf("invalid port: %w", err)
@@ -535,10 +539,11 @@ func (s *Server) Listen(network, addr string) (net.Listener, error) {
 		return nil, err
 	}
 
-	key := listenKey{network, host, uint16(port)}
+	key := tcpListenKey{port: uint16(port)}
 	ln := &listener{
 		s:    s,
 		key:  key,
+		net:  network,
 		addr: addr,
 
 		conn: make(chan net.Conn),
@@ -553,15 +558,14 @@ func (s *Server) Listen(network, addr string) (net.Listener, error) {
 	return ln, nil
 }
 
-type listenKey struct {
-	network string
-	host    string
-	port    uint16
+type tcpListenKey struct {
+	port uint16
 }
 
 type listener struct {
 	s    *Server
-	key  listenKey
+	key  tcpListenKey
+	net  string
 	addr string
 	conn chan net.Conn
 }
@@ -590,5 +594,5 @@ func (ln *listener) Server() *Server { return ln.s }
 
 type addr struct{ ln *listener }
 
-func (a addr) Network() string { return a.ln.key.network }
+func (a addr) Network() string { return a.ln.net }
 func (a addr) String() string  { return a.ln.addr }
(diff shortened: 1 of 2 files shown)


tsnet/tsnet.go
```diff
@@ -528,6 +529,9 @@ func (s *Server) Listen(network, addr string) (net.Listener, error) {
 	if err != nil {
 		return nil, fmt.Errorf("tsnet: %w", err)
 	}
+	if host != "" && host != s.Hostname {
+		return nil, fmt.Errorf("tsnet: hostname %q is not supported", host)
+	}
```
bradfitz, line 534: this kinda implies a hostname means something, though?

It doesn't yet, because we don't know the hostname-to-IP(s) mapping at the time of Listen... we might not yet be in control.

And if addr is an IP address, it should be possible to listen on both the Tailscale IP and IPv6 separately too.

Also: if addr is an IP, its adddress family should be checked against tcp4 or tcp6 if those are specified.

tsnet/tsnet.go
```diff
@@ -536,7 +540,7 @@ func (s *Server) Listen(network, addr string) (net.Listener, error) {
 		return nil, err
 	}
 
-	key := listenKey{network, host, uint16(port)}
+	key := listenKey{network, "", uint16(port)}
```
bradfitz, line 543: how about we keep it the host but try more matches on the lookup side?

tsnet/tsnet.go
creachadair: > this kinda implies a hostname means something, though?

Yes, and I'm not sure whether it does. In `start()` the `s.Hostname` gets plumbed into the preferences, so I guess it does affect what we ask the coordination server to authenticate for? But the "host" part of the address doesn't appear to be used again until we're routing TCP sessions later on.

So maybe a better solution would be to reject _any_ non-empty "host" field here? We don't seem to make any other use of it.

tsnet/tsnet.go
creachadair: That makes sense—but I'm not sure what keys we would use. I guess we could use `s.hostname`, since at this point it's been filled in by `start()`. Are there other possible values we need to consider?

creachadair: I ask because it seems like an incoming connection could have lots of different spellings: The bare hostname, the fully-qualified hostname, some IP address, perhaps aliases?

tsnet/tsnet.go
```diff
@@ -518,6 +518,7 @@ func (s *Server) APIClient() (*tailscale.Client, error) {
 
 // Listen announces only on the Tailscale network.
 // It will start the server if it has not been started yet.
+// The addr should not include a hostname, only a port, for example ":80".
 func (s *Server) Listen(network, addr string) (net.Listener, error) {
 	switch network {
 	case "", "tcp", "tcp4", "tcp6":
```
bradfitz, line 524: will 3 of these 4 be a problem too?

tsnet/tsnet.go
bradfitz: From most specific to least specific:

* "IP:port" (or "[ip]:port" for IPv6)
* "hostname:port"
* ":port"

Then could return an error if it's not one of those?

But then you'd expect an error if they listen on "tcp6" but give an IPv4 or vice versa.

But I don't think anything but "tcp" even works today.

tsnet/tsnet.go
creachadair: That's a good question. The [docs for net.Listen](https://pkg.go.dev/net#Listen) say that you can use a hostname with "tcp4" and "tcp6", so I do not think these values make the situation any worse. The real problem is that if we set the hostname portion of the address to anything other than "", the routing won't work.

But then, if we want to check for "valid" values at the point where we're routing, we'd have to re-parse out the address. Mechanically that's fine, but do we want to spend the resources to do that on each connection? I'm happy to do the work if you think it's worthwhile.

tsnet/tsnet.go
creachadair: To get that, I think we have to call `conn.LocalAddr()` and re-parse the string, and then hold that up against what all those addresses are supposed to be. Presumably those are assigned by the coordination server—but they don't seem to be recorded in the Server type so I guess we'd have to either look them up, or update the setup to save them. 

It seems like a simpler solution is to simply disallow (and report an error for) non-empty hostnames in Listen: They already don't work now, so we wouldn't be breaking any working software—and getting an error would be an improvement over the silent failure we have with the current situation.

tsnet/tsnet.go
creachadair: Am I correct in thinking that this listener won't even receive a connection if it didn't already arrive with an address wireguard thinks corresponds to this TS instance?

creachadair: I guess as it's written now, we do need them to match because we include it in the key.

It doesn't look like we necessarily need the network type in the key, though—for forwarding purposes, I think we've already asked netstack to only do this for TCP (cf. "ForwardTCPIn, if non-nil, handles forwarding an inbound TCP connection"). 

Could we do the check in Listen to give a sensible response (as now) but then omit the network from the listenKey entirely?

bradfitz:
I'm not reviewing this while it's in draft & continually changing. I saw a bunch of emails about it over the holidays but haven't looked yet. (just closing tabs and saw it again)

Let me know when it's ready.

creachadair:
> I'm not reviewing this while it's in draft & continually changing. I saw a bunch of emails about it over the holidays but haven't looked yet. (just closing tabs and saw it again)
> 
> 
> 
> Let me know when it's ready.
> 
> 

I'm sorry about the spam. I haven't made any other changes, it was probably mostly my script updating the branch, I should have disabled it. 

If you think the general approach is reasonable I'll clean it up and make it worth reviewing; otherwise I'm open to suggestions.

PR #6819, opened 2022-12-21
envknob, magicsock: add TS_DISCO_PONG_IPV4_DELAY knob to bias IPv6 paths

### LINKED ISSUE #6818: wgengine/magicsock: add envknob to conditionally bias IPv6
My server has IPv4 & IPv6 transit. Except my IPv6 is free and my IPv4 is not, so I'd much prefer bulk multi-terabyte copies from my home to server go over IPv6. But of my two home ISPs (both of which have IPv6), the one with higher bandwidth has non-native IPv6 and requires a 6rd tunnel that adds a millisecond of latency. So magicsock usually picks IPv4 for that pair of nodes, despite 2ms vs 3ms being immaterial.

We already bias IPv6 a bit elsewhere, but I want a knob to bias it very definitively from my server's side and say "no, prefer IPv6 even if IPv4 is X duration faster."

bradfitz:
Fixes #6818

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -36,6 +36,7 @@ import (
 	"tailscale.com/derp"
 	"tailscale.com/derp/derphttp"
 	"tailscale.com/disco"
+	"tailscale.com/envknob"
 	"tailscale.com/health"
 	"tailscale.com/hostinfo"
 	"tailscale.com/ipn/ipnstate"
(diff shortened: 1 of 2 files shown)


wgengine/magicsock/magicsock.go
```diff
@@ -1946,13 +1947,24 @@ const (
...
 // sendDiscoMessage sends discovery message m to dstDisco at dst.
 //
 // If dst is a DERP IP:port, then dstKey must be non-zero.
 //
 // The dstKey should only be non-zero if the dstDisco key
 // unambiguously maps to exactly one peer.
 func (c *Conn) sendDiscoMessage(dst netip.AddrPort, dstKey key.NodePublic, dstDisco key.DiscoPublic, m disco.Message, logLevel discoLogLevel) (sent bool, err error) {
+	isDERP := dst.Addr() == derpMagicIPAddr
+	if _, isPong := m.(*disco.Pong); isPong && !isDERP && dst.Addr().Is4() {
+		time.Sleep(debugIPv4DiscoPingPenalty())
```
andrew-d, line 1965: I don't super love `time.Sleep` here, but after looking through the code there doesn't appear to be a cleaner simple way to do this.

andrew-d (approved):

wgengine/magicsock/magicsock.go
bradfitz: No locks are held. It holds a goroutine but that's just some memory. An AfterFunc would be cheaper in that regard but doesn't seem worth it. It a peer is DoSing you with pings you shouldn't be replying anyway.

PR #6825, opened 2022-12-21
cmd/tailscale/cli: redact private key in netmap output by default.

danderson:
This makes `tailscale debug watch-ipn` safe to use for troubleshooting user issues, in addition to local debugging during development.

Updates tailscale/corp#8489


bradfitz (approved):
Sure, but little weird to have some filtering on the client side and some filtering (prefs) in the server side.

Maybe instead pass a new watchIPNArgs bit mask flag to the backend to say which behavior we want?

But we can do that later too without a UX change. This is fine for now.

danderson:
Was easy enough to move into tailscaled, PTAL?

ipn/ipnlocal/local.go
```diff
@@ -1742,6 +1742,21 @@ func (b *LocalBackend) readPoller() {
 func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWatchOpt, fn func(roNotify *ipn.Notify) (keepGoing bool)) {
 	ch := make(chan *ipn.Notify, 128)
 
+	redact := func(n *ipn.Notify) bool {
```
bradfitz, line 1745: This works, but the "if not redact" below reads weirdly. "You don't want me to redact? Did the redaction fail?"

Maybe instead here do something like:

```go
    origFn := fn
    if mask&ipn.NotifyNoPrivateKeys != 0 {
          fn = func(...) {
                 ... use origFn conditionally ...
          }
     }
```

... and then keep the callers below just calling `fn` ?

bradfitz (approved):

ipn/ipnlocal/local.go
danderson: Done.

ipn/ipnlocal/local.go
```diff
@@ -1742,6 +1742,24 @@ func (b *LocalBackend) readPoller() {
...
+				return origFn(n)
+			}
+
+			// The netmap in n is shared across all watchers, so to mutate it for a
+			// single watcher we have to clone the notify and the netmap. We can
+			// make shallow clones, at least.
+			nm := *n.NetMap
+			r := *n
+			r.NetMap = &nm
+			r.NetMap.PrivateKey = key.NodePrivate{}
```
bradfitz, line 1758: Maybe variable names `nm2` (netmap 2) and `n2` (notify 2) ?

bradfitz (approved):

bradfitz:
Nit: no period in commit message title at end.

ipn/ipnlocal/local.go
danderson: Done. And one more round through the dog-slow CI, yay!

PR #6831, opened 2022-12-22
wgengine/netstack: delete some dead code, old comment, use atomic int types

bradfitz:
Noticed while looking at something else; #cleanup.

Change-Id: Icde7749363014eab9bebe1dd80708f5491f933d1
Signed-off-by: Brad Fitzpatrick <bradfitz@tailscale.com>

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -107,8 +107,8 @@ type Impl struct {
 	lb        *ipnlocal.LocalBackend // or nil
 	dns       *dns.Manager
 
-	peerapiPort4Atomic uint32 // uint16 port number for IPv4 peerapi
-	peerapiPort6Atomic uint32 // uint16 port number for IPv6 peerapi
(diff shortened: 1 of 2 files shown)


wgengine/netstack/netstack.go
```diff
@@ -559,11 +559,6 @@ func (ns *Impl) shouldProcessInbound(p *packet.Parsed, t *tstun.Wrapper) bool {
 	if p.IPVersion == 6 && !isLocal && viaRange.Contains(dstIP) {
 		return ns.lb != nil && ns.lb.ShouldHandleViaIP(dstIP)
 	}
-	if !ns.ProcessLocalIPs && !ns.ProcessSubnets {
-		// Fast path for common case (e.g. Linux server in TUN mode) where
-		// netstack isn't used at all; don't even do an isLocalIP lookup.
-		return false
-	}
```
bradfitz, line 566: (this is all redundant with the following lines and the comment is no longer accurate anyway: we always do the isLocalIP lookup earlier anyway)

DentonGentry (approved):

Issue #6054, opened 2022-10-25
SSH and SCP commands to tailscale ssh endpoints do not return correct error / success codes


SebastianZaha:
### What is the issue?

Issue found while testing the [mutagen](https://github.com/mutagen-io/mutagen) file synchronization tool with a server where I use `tailscale --ssh`.

That program attempts to run a remote agent via a ssh command `ssh user@tailscalehost agent` and handles the 127 POSIX error code for when the command does not exist. tailscale ssh just returns an exit code of 1, although I think it writes the correct exit code in stderr. (see below).

Aside from that, a normal (and **successful**) scp command to a tailscale --ssh host also returns EOF with exit code 1 on my machine.

Local ssh version on arch linux: `core/openssh 9.1p1-1 [installed]`.

### Steps to reproduce

```
sebi@archdesktop ~/devel/mutagen $ ssh deckbox@a1 boguscommand
/bin/bash: boguscommand: command not found
exit status 127
sebi@archdesktop ~/devel/mutagen $ echo $?
1
sebi@archdesktop ~/devel/mutagen $ ssh deckbox@a1.bifolia.net boguscommand
bash: boguscommand: command not found
sebi@archdesktop ~/devel/mutagen $ echo $?
127
```

```
sebi@archdesktop ~/devel/mutagen $ scp README.md deckbox@a1.bifolia.net:
README.md                                                                     100% 5058   189.3KB/s   00:00    
sebi@archdesktop ~/devel/mutagen $ echo $?
0
sebi@archdesktop ~/devel/mutagen $ scp README.md deckbox@a1:
README.md                                                                     100% 5058   158.6KB/s   00:00    
EOF
sebi@archdesktop ~/devel/mutagen $ echo $?
1
```

### OS

Linux

### OS version

Ubuntu 18.04

### Tailscale version

1.32.1

### Bug report

BUG-b56db22c2e3ba12f87fd30ea839658e217bf7311f497485a617add42aac24bdb-20221025090330Z-0104a27f39e6c969

DentonGentry:
I cannot reproduce the scp part of the issue:
```
% scp ./README.md ubuntu:
README.md                                                                             100% 8834     1.2MB/s   00:00
% echo $?
0
```

jfbus:
I have the same issue with scp

```
% scp /tmp/foo rousseau-1:.
foo                                                                                                                                            100% 8576KB 531.9KB/s   00:16
EOF
% echo $?
1
```

Server:
```
[...]

PR #6837, opened 2022-12-23
ssh/tailssh: don't swallow process exit code in be-child

bradfitz:
Thanks to @nshalman and @Soypete for debugging!

Updates #6054

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -283,7 +283,20 @@ func beIncubator(args []string) error {
 			Foreground: true,
 		}
 	}
-	return cmd.Run()
+	err = cmd.Run()
+	if ee, ok := err.(*exec.ExitError); ok {
+		ps := ee.ProcessState
+		code := ps.ExitCode()
+		if code < 0 {
+			// TODO(bradfitz): do we need to also check the syscall.WaitStatus
+			// and make our process look like it also died by signal/same signal
+			// as our child process? For now we just do the exit code.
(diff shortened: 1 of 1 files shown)


ssh/tailssh/incubator.go
```diff
@@ -283,7 +283,20 @@ func beIncubator(args []string) error {
...
-	return cmd.Run()
+	err = cmd.Run()
+	if ee, ok := err.(*exec.ExitError); ok {
+		ps := ee.ProcessState
+		code := ps.ExitCode()
+		if code < 0 {
+			// TODO(bradfitz): do we need to also check the syscall.WaitStatus
+			// and make our process look like it also died by signal/same signal
+			// as our child process? For now we just do the exit code.
+			fmt.Fprintf(os.Stderr, "[tailscale-ssh: process died: %v]\n", ps.String())
```
nshalman, line 294: Should this be a more formal log line with `logf`? (I can't quite tell if you have access to that here...)

Soypete (approved):
LGTM! checking the ExitError should help debugging future issues as well

nshalman (approved):
Worked for my test case!

ssh/tailssh/incubator.go
nshalman: Looks like `logf := logger.Discard` so I guess not.

maisem (approved):

ssh/tailssh/incubator.go
bradfitz: logf at this point is a weirdo thing that writes via a high fd up to the child process, I believe.

This is definitely a weird environment.

PR #6844, opened 2022-12-24
flake.nix: make a flake that gives access to unstable tailscale builds.

danderson:
This lets me (and other brave souls) pull tailscale/tailscale into my NixOS configs as a source, and use the `tailscale-unstable` package to run dev builds on my home tailnet.

Because of the forced vendorSha256 semantics for go module builds, this flake is unfortunately going to be annoyingly brittle as we update go.mod during development. I don't want to force everyone to have a nix dep and fix things up every time they need to change go.mod, so this flake is expected to be brittle right now, with frequently broken builds that I'll need to fix up manually. I'm okay with this.

Longer term, I think we can reuse the NAR hash computation thing @bradfitz made to make it possible to update the vendor NAR hash without needing a full nix install, at which point we can add automation and testing to keep the flake working. But it's christmas eve and all I want for christmas is a tailscale-unstable package that works well enough for just me to use it :)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,154 @@
+# flake.nix describes a Nix source repository that provides
+# development builds of Tailscale and the fork of the Go compiler
+# toolchain that Tailscale maintains. It also provides a development
+# environment for working on tailscale, for use with "nix develop".
+#
+# For more information about this and why this file is useful, see:
+# https://nixos.wiki/wiki/Flakes
+#
+# Also look into direnv: https://direnv.net/, this can make it so that you can
+# automatically get your environment set up when you change folders into the
+# project.
+#
+# WARNING: currently, the packages provided by this flake are brittle,
+# and importing this flake into your own Nix configs is likely to
+# leave you with broken builds periodically.
+#
+# The issue is that building Tailscale binaries uses the buildGoModule
+# helper from nixpkgs. This helper demands to know the content hash of
+# all of the Go dependencies of this repo, in the form of a Nix SRI
+# hash. This hash isn't automatically kept in sync with changes made
+# to go.mod yet, and so every time we update go.mod while hacking on
+# Tailscale, this flake ends up with a broken build due to hash
+# mismatches.
+#
+# Right now, this flake is intended for use by Tailscale developers,
+# who are aware of this mismatch and willing to live with it. At some
+# point, we'll add automation to keep the hashes more in sync, at
+# which point this caveat should go away.
+#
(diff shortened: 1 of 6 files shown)


danderson:
With a few more words: the vendor hash in go module builds is just the NAR hash of `go mod download`, effectively. So, we can take the NAR hashing thing we already have, teach it how to make a NAR hash given a go.mod/go.sum, and that should all work out. I'm playing with it as time/energy permits.

flake.nix
```diff
@@ -0,0 +1,112 @@
...
+    # but alas.
+    #
+    # So really, this flake is for tailscale devs to dogfood with, if
+    # you're an end user you should be prepared for this flake to not
+    # build periodically.
+    tailscaleUnstable = pkgs: pkgs.buildGo119Module rec {
+      name = "tailscale-unstable";
+
+      src = ./.;
+      vendorSha256 = "sha256-+5icFKDHXt3JMbUjLQGes4R+GeUi48xRgGd0yPKVrw0=";
```
bradfitz, line 76: can you at least pull this/these out to their own files like we have in the other repo? That'll make GitHub Actions automations easier later.

version/version.go
```diff
@@ -61,6 +58,9 @@ func init() {
 			}
 		}
 	}
+	if commitHashAbbrev == "" && len(GitCommit) >= 9 {
```
bradfitz, line 61: how `commitHashAbbrev` _not_ be the empty string at this point?  Remove that first part of the test? And move the decl down?

flake.nix
```diff
@@ -0,0 +1,112 @@
+{
```
bradfitz, line 1: Add a comment to the top of this file (like shell.nix has), explaining to non-Nix people what this file even is? And maybe a URL to a tailscale/tailscale issue about this flake in particular and about why it's brittle for now?

bradfitz (approved):

flake.nix
bradfitz: Oh, and make a new issue I guess. I assumed you had one already. Then reference it from the commit message too.

flake.nix
danderson: Done, pulled out into .rev and .sri files as appropriate.

version/version.go
danderson: Good catch, fixed up.

flake.nix
danderson: Done, and done.

Issue #6845, opened 2022-12-24
Ship a Nix flake for unstable builds


danderson:
I use NixOS everywhere. I would also like to dogfood unstable builds.

The modern nix-ish way to do this is to turn our source repository into a Nix flake, by adding flake.nix at the repo root. This lets us provide some package build instructions and other niceties like configuration for a development shell, in a form that nix flake users can import in a nice versioned way (e.g. `nix flake lock --update-input tailscale` does the equivalent of a git pull of our repo, and rebuilds tailscale with the updated source code and build scripts).

Using flakes to distribute Go code currently has an irritating sharp edge: because Nix wants to be hermetic, and because of an obscure ecosystem limitation ("IFD" or "import from derivation" being a widely discouraged thing, for complicated reasons), the Nix build instructions for Tailscale need to know ahead of time the NAR hash of all of our go.mod dependencies. Roughly, a hash of the output of `go mod vendor`. This means that whenever we update go.mod as part of normal development, the flake.nix build will break, because suddenly the Nix-universe hash of dependencies no longer matches the dependencies that go.mod specifies. This sucks and I wish Nix were instead capable of integrating more gracefully with go.mod and go.sum, but for now this is the situation.

So, initially, our flake.nix is going to be very appropriately named, in that it'll be flaky and break periodically as people update go.mod. This is okay to get started, because I and other nix-enthusiast tailscalars are the primary consumers of this flake, and I/we are willing to fix the breakages asynchronously.

Medium-term, we can fix this the way we fixed Go toolchain updates in our other repo: @bradfitz made a Go NAR writer library, which can take an input file tree and plop out the Nix SRI hash for it. With that, we can build some tooling to automate updating the SRI hashes that flake.nix depends on, and end up with a less broken flake.

danderson:
Flake is implemented, along with some tooling to keep some of the SRI hashes naturally in sync during development. The final stage that isn't implemented yet, is a CI step that asks people to `./update-flake.sh` when necessary. Once that's done, the flake should be reasonably robust for general use.

DentonGentry:
> The final stage that isn't implemented yet, is a CI step that asks people to ./update-flake.sh when necessary. Once that's done, the flake should be reasonably robust for general use.

https://github.com/tailscale/tailscale/issues/7151 appears to cover that step specifically.

Barring further comment, we'll expect to close this during the next regular bug scrub.

PR #6847, opened 2022-12-24
cmd/printdep: support printing the toolchain SRI hash.

danderson:
Updates #6845.


danderson:
(in case not obvious: in prep for syncing up our corp repo's build system, to keep everything in sync from a single source)

bradfitz (approved):

Issue #6865, opened 2023-01-02
Copyright year in generated files


DentonGentry:
### What is the issue?

We have three decisions which individually are reasonable, but taken together result in an impossible situation:
1. The [go-generate CI test](https://github.com/tailscale/tailscale/blob/main/.github/workflows/go_generate.yml) checks that `go generate` is up to date.
1. [util/codegen](https://github.com/tailscale/tailscale/blob/692eac23adf62f979dddb5b9694d084142036f10/util/codegen/codegen.go#L108) uses time.Now().Year()
1. [Software copyright headers](https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code) should not routinely update the year every year.

Right now the CI test fails because it wants all generated files to be updated for 2023.

### Proposal
Modify util/codegen to not include the year in the copyright header at all. It is [the practice at Microsoft now](https://github.com/dsccommunity/DFSDsc/issues/79) that software copyright headers do not include the year, so there is some precedent for it.

The change in the code generator will need to go in along with an update to all generated files to remove the year from their header.

Manually produced software can — and should — continue to include the year of creation in the header. This proposal is just pragmatic, about bots producing software.

[PR#6866](https://github.com/tailscale/tailscale/pull/6866) implements this proposal.

bradfitz:
@DentonGentry, leaving this open for now for, @willnorris, et al to figure out what we want to do here long-term (if anything).

willnorris:
Copyrights should just include the first year of publication and that's it.  No need to include a range that includes the current year.

willnorris:
(or optionally no year at all.  That's fine too, but less common)

willnorris:
for reference, here are the docs we published at Google years ago: https://opensource.google/documentation/reference/copyright.  Notably, current year is marked as optional... if I remember correctly, mainly so that teams that were already using current year didn't think that they **needed** to go remove it from all of their files.  Though if I remember correctly, we mostly recommended that the next time they would normally go bump the current year, to then take that opportunity to remove it.

DentonGentry:
> the first year of publication

I think we could get this by changing:
https://github.com/tailscale/tailscale/blob/91e64ca74ffb22cf5a32f5eeee60196b12aa1636/util/codegen/codegen.go#L305

from `y > year` to `y != 0 && y < year`.

This will result in one last regeneration of the existing generated files, but should not update them again.

willnorris:
[...]

PR #6866, opened 2023-01-02
util/codegen: Remove year from copyright header.

DentonGentry:
Copyright notices in software are not supposed to update the year in the header.

Because we have a CI check for `go generate`, we're failing CI until we go update all of the copyright headers in generated files to say 2023.

Fixes https://github.com/tailscale/tailscale/issues/6865

### DIFF
--- a/cmd/viewer/tests/tests_clone.go
+++ b/cmd/viewer/tests/tests_clone.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 Tailscale Inc & AUTHORS All rights reserved.
+// Copyright (c) Tailscale Inc & AUTHORS All rights reserved.
 // Use of this source code is governed by a BSD-style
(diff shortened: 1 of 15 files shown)


creachadair:
I am not a lawyer, but my understanding is that for legal purposes in the U.S., the copyright notice is expected to contain the year of first publication. But copyright notices are not required for protection under the Berne Convention, so I think we could also safely omit them entirely.

(The folk convention of "updating" the copyright year on existing works strikes me as odd; it's common but doesn't make sense: From the perspective of protection, you want the earliest date possible to be indicated)

I'm not sure whether, by removing the dates entirely, we're disrupting the legal form of the notice. Could we maybe drop the headers entirely?

DentonGentry:
We'll use https://github.com/tailscale/tailscale/pull/6877 instead.

PR #6876, opened 2023-01-03
net/dns/resolvconffile: fix handling of multiple search domains

### LINKED ISSUE #6875: /etc/resolv.conf parsing limits entire search domain list to 253 bytes
### What is the issue?

When reading in a resolv.conf file the entire `search` line is passed to ToFQDN:
https://github.com/tailscale/tailscale/blob/692eac23adf62f979dddb5b9694d084142036f10/net/dns/resolvconffile/resolvconffile.go#L91

ToFQDN compares the length of the string to 253, the maximum length of a domain name: https://github.com/tailscale/tailscale/blob/692eac23adf62f979dddb5b9694d084142036f10/util/dnsname/dnsname.go#L38

However there can be many search domains on the one line. Each one should be allowed to be 253 bytes, not the sum of all of them.

HS #8274

### Steps to reproduce

_No response_

_No response_

### OS

_No response_

### OS version

_No response_

### Tailscale version

1.34.1

### Bug report

_No response_

bradfitz:
Fixes #6875

### DIFF
--- a/net/dns/resolvconffile/resolvconffile.go
+++ b/net/dns/resolvconffile/resolvconffile.go
@@ -83,17 +83,26 @@ func Parse(r io.Reader) (*Config, error) {
 		}
 
 		if s, ok := strs.CutPrefix(line, "search"); ok {
-			domain := strings.TrimSpace(s)
-			if len(domain) == len(s) {
+			domains := strings.TrimSpace(s)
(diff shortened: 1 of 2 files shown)


bradfitz:
Kids are asleep so I guess vacation's over.

DentonGentry (approved):

PR #6877, opened 2023-01-03
util/codegen, all: use latest year, not time.Now, in generated files

### LINKED ISSUE #6865: Copyright year in generated files

bradfitz:
Updates #6865

### DIFF
--- a/util/codegen/codegen.go
+++ b/util/codegen/codegen.go
@@ -6,6 +6,7 @@
 package codegen
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"go/ast"
@@ -13,13 +14,16 @@ import (
 	"go/types"
 	"io"
 	"os"
+	"path/filepath"
 	"reflect"
+	"regexp"
+	"strconv"
 	"strings"
-	"time"
 
 	"golang.org/x/tools/go/packages"
 	"golang.org/x/tools/imports"
 	"tailscale.com/util/mak"
+	"tailscale.com/util/must"
 )
 
 // LoadTypes returns all named types in pkgName, keyed by their type name.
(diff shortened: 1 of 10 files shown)


bradfitz:
Alternative (or interim?) proposal instead of #6866?

util/codegen/codegen.go
```diff
@@ -263,3 +267,49 @@ func IsViewType(typ types.Type) bool {
 	}
 	return t.Field(0).Name() == "ж"
 }
+
+// MaxCopyrightYear reports the greated copyright year in non-generated *.go
```
DentonGentry, line 271: greated == greatest

util/codegen/codegen.go
```diff
@@ -263,3 +267,49 @@ func IsViewType(typ types.Type) bool {
...
+		}
+		src, err := os.ReadFile(filepath.Join(dir, name))
+		if err != nil {
+			panic(err)
+		}
+		bs := bufio.NewScanner(bytes.NewReader(src))
+		for bs.Scan() {
+			line := bs.Bytes()
+			if m := rxYear.FindSubmatch(line); m != nil {
+				if y := must.Get(strconv.Atoi(string(m[1]))); y > year {
```
DentonGentry, line 301: I think the usual guidance is the year of creation, and then not updating date in the future even as changes are made in the code. That would suggest the earliest non-zero year, instead of the latest.

However anything would be better than the tree being broken, we can discuss it more this coming week.

DentonGentry (approved):

util/codegen/codegen.go
bradfitz: Renamed the func to "CopyrightYear" instead of "MaxCopyrightYear" and added a TODO and changed the commit to not "Fixes" the issue but rather only "Updates".

PR #6882, opened 2023-01-03
TKA tweaks, fixes, error handling

twitchyliquid64:
* `tka.Key.KeyID()` now returns an error instead of panicking.
 * Unknown AUM types are now correctly handled.
 * Minor UX tweaks and fixes falling out of the alpha.

The former two changes won't matter now (no new key types or AUM types) but we'll need to lockstep `1.34` in the future :grimacing:

### DIFF
--- a/cmd/tailscale/cli/network-lock.go
+++ b/cmd/tailscale/cli/network-lock.go
@@ -5,6 +5,7 @@
 package cli
 
 import (
+	"bytes"
 	"context"
 	"crypto/rand"
 	"encoding/hex"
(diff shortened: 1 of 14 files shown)


bradfitz:
> cmd/tailscale,tka: KeyID returns an error instead of panicking

Wrong verb tense for a commit (that's how you'd write a bug report). Is this "make KeyID return error instead of panicking"?

twitchyliquid64:
Yes it is! willfix.jpg

sailorfrag:
> Yes it is! willfix.jpg

s/returns/return/ 😄 

The other one doesn't start with a verb, FWIW, I suggest "add" or "implement"

cmd/tailscale/cli/network-lock.go
```diff
@@ -99,6 +100,18 @@ func runNetworkLockInit(ctx context.Context, args []string) error {
 		return err
 	}
 
+	// Common mistake: Not specifying the current node's key as one of the trusted keys.
```
sailorfrag, line 103: optional: I think this could be done more succinctly using `x/exp/slices` `slices.ContainsFunc` but I'm not sure if it's worth changing that now.

sailorfrag (approved):
Other than commit message tweaks this seems fine

PR #6892, opened 2023-01-04
health,ipn/ipnlocal: report the node being locked out as a health issue

twitchyliquid64:
Looked like this during testing:

```
# Health check:
#     - dns-os: getting OS base config is not supported
#     - dns: getting OS base config is not supported
#     - tailnet-lock: this node is locked out; it will not have connectivity till it is signed
#     - This is an unstable (development) version of Tailscale; frequent updates and bugs are likely
```

### DIFF
--- a/ipn/ipnlocal/network-lock.go
+++ b/ipn/ipnlocal/network-lock.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"tailscale.com/envknob"
+	"tailscale.com/health"
 	"tailscale.com/ipn"
 	"tailscale.com/ipn/ipnstate"
 	"tailscale.com/net/tsaddr"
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/network-lock.go
```diff
@@ -111,6 +114,13 @@ func (b *LocalBackend) tkaFilterNetmapLocked(nm *netmap.NetworkMap) {
 	} else {
 		b.tka.filtered = nil
 	}
+
+	// Check that we ourselves are not locked out, report a health issue if so.
+	if nm.SelfNode != nil && b.tka.authority.NodeKeyAuthorized(nm.SelfNode.Key, nm.SelfNode.KeySignature) != nil {
+		health.SetTKAHealth(errors.New("this node is locked out; it will not have connectivity till it is signed"))
```
bradfitz, line 120: until

twitchyliquid64: Oooooh, is that a thing we can do / do we do that for anything else? I like the idea of a specialized "uwu oopsie daisies y'all locked out, do the thing" page.

bradfitz: yeah `git grep tailscale.com/s/`

sailorfrag (approved):

Issue #755, opened 2020-09-11
FR: add auto-update mechanism


bradfitz:
We need a Windows auto-updater. Windows updates the slowest of all our platforms:

<img width="1071" alt="Screen Shot 2020-09-11 at 12 19 00 PM" src="https://user-images.githubusercontent.com/2621/92964761-80cd5c00-f429-11ea-83c6-73e17c2a26e6.png">

Misc notes copied from internal bug:

* It might be opt-in or might be opt-out, and should probably have a registry-settable override for org admins to prevent their managed machines from auto-updating outside the org's schedules. But we should support an auto-update mechanism in any case, even if not everybody wants to use it.
* Checkbox in systray menu, like: `[X] Auto-update software`
* Windows service uses BITS via the bitsadmin CLI tool to download the new installer. It'd be easy in Go too, of course, but too easy to do poorly. By using BITS, we avoid being rude to their network (using too much bandwidth or money) and can instead let Windows decide how quickly and when to download, based on which wi-fi networks they've classified in which ways (marking an LTE hotspot as expensive, for example).
* Need to verify the download once it's downloaded from BITS: https://twitter.com/bradfitz/status/1289300026200481792
* When the download is complete, we simply run the NSIS installer in headless mode, the same way that Chocolatey does it, with /S: https://github.com/zombiezen/tailscale-chocolatey/blob/master/tools/chocolateyinstall.ps1#L20
* The install stops & re-starts everything.
* We can have a intent state file that makes sure we don't try repeatedly in a loop, lest an update go poorly
* we could have the service send a Notify to the GUI to schedule itself to start again when it notices the mtime/inode/size of the GUI is different. So then the GUI can fork off a new process (a copy of the exe to a temp file?) that's polling the file every second for up to 30 seconds and doing the RunFromProcess.exe trick when it notices a new binary is in place? But then it's the GUI restarting itself (upon command) from its context, rather than the admin context.

jucor:
Naive related question: it seems the [official windows installer](https://tailscale.com/download) is stuck at 1.0.4, whereas 1.0.5 is available on Linux and MacOS. Is that intended?
Would *love* an auto-update :)

bradfitz:
There were only two changes in 1.0.5: one was Linux only, and one was a client bug for all platforms that we ended up also working around server-side. So there was no reason to do 1.0.5 on Windows.

bradfitz:
@alexbrainman, this something you'd be interested in working on?

[...]

PR #6894, opened 2023-01-05
cmd/tailscale: add start of "tailscale update" command

bradfitz:
Goal: one way for users to update Tailscale, downgrade, switch tracks, regardless of platform (Windows, most Linux distros, macOS, Synology).

This is a start.

Updates #755, etc


cmd/tailscale/cli/update.go
```diff
@@ -0,0 +1,199 @@
...
+	if up.currentOrDryRun(ver) {
+		return nil
+	}
+	url := fmt.Sprintf("https://pkgs.tailscale.com/%s/debian/pool/tailscale_%s_%s.deb", up.track, ver, runtime.GOARCH)
+	// TODO(bradfitz): require root/sudo
+	return errors.New("TODO: Debian/Ubuntu deb download of " + url)
+}
+
+func (up *updater) updateMacSys() error {
+	// use sparkle? do we have permissions from this context? does sudo help?
```
mihaip, line 159: Sparkle seems best (as far as handling sandboxed installs). We could perhaps set a "check for updates" flag in the IPN state and have the UI process check for that and call `updater.checkForUpdates(nil)` on it.

bradfitz: If I can LocalAPI to the backend and say "yo, please ask Sparkle to do X", can X include asking Sparkle to try switching to a specific version? (for switching tracks, downgrading, etc)

mihaip: I think we can override the feed URL that Sparkle checks (https://github.com/tailscale/corp/blob/34b0d183fb5ad08cfa2227a6678e05309411a6b2/xcode/IPN/UI/macOS/Sparkle/SparkleMacsys.swift#L45-L48) to get it to switch tracks.

bradfitz: But can we change it at runtime?

mihaip: Based on https://sparkle-project.org/documentation/publishing/#setting-the-feed-programmatically it sounds like we can, but we could also switch to a single feed and use Sparkle's native support for channels.

cmd/tailscale/cli/update.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/peterbourgon/ff/v3/ffcli"
+	"tailscale.com/util/winutil"
+	"tailscale.com/version"
+	"tailscale.com/version/distro"
+)
+
+var updateCmd = &ffcli.Command{
+	Name:       "update",
+	ShortUsage: "update",
+	ShortHelp:  "Update Tailscale to the latest/different version",
```
mayakacz, line 27: I believe(?) you also need to explicitly add long help for it to show up.

cmd/tailscale/cli/update.go
```diff
@@ -0,0 +1,199 @@
...
+	Exec:       runUpdate,
+	FlagSet: (func() *flag.FlagSet {
+		fs := newFlagSet("update")
+		fs.BoolVar(&updateArgs.yes, "yes", false, "avoid interactive prompts and just update")
```
mayakacz, line 31:
```suggestion
		fs.BoolVar(&updateArgs.yes, "yes", false, "Update without interactive prompts")
```

cmd/tailscale/cli/update.go
```diff
@@ -0,0 +1,199 @@
...
+		fs.BoolVar(&updateArgs.dryRun, "dry-run", false, "print what would be done without doing anything or prompting")
```
mayakacz, line 32:
```suggestion
		fs.BoolVar(&updateArgs.dryRun, "dry-run", false, "Print what update would be done without prompting")
```

cmd/tailscale/cli/update.go
```diff
@@ -0,0 +1,199 @@
...
+		fs.StringVar(&updateArgs.track, "track", "", `which track to check for updates: "stable" or "unstable" (dev); empty means same as current`)
```
mayakacz, line 33: Is this based on current version then? Default/empty is same as current, not stable?

mayakacz (approved):

cmd/tailscale/cli/update.go
```diff
@@ -0,0 +1,199 @@
...
+		ver = latest.Version
+		if ver == "" {
+			return errors.New("no version found")
+		}
+	}
+	arch := runtime.GOARCH
+	if arch == "386" {
+		arch = "x86"
+	}
+	url := fmt.Sprintf("https://pkgs.tailscale.com/%s/tailscale-setup-%s-%s.msi", up.track, ver, arch)
```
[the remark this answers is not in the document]
bradfitz: Not exactly, because this tool lets you downgrade to specific versions and that API endpoint doesn't let us ask "do you have this version number?"

cmd/tailscale/cli/update.go
bradfitz: Yes, current version they're already running. Default is no change of track.

PR #6896, opened 2023-01-05
ipn/ipnlocal: add c2n handler to flush logtail for support debugging

bradfitz:
Updates tailscale/corp#8564

### DIFF
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -292,6 +292,7 @@ func (h *Handler) serveBugReport(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "only POST allowed", http.StatusMethodNotAllowed)
 		return
 	}
+	defer h.b.TryFlushLogs() // kick off upload after bugreport's done logging
 
 	logMarker := func() string {
 		return fmt.Sprintf("BUG-%v-%v-%v", h.backendLogID, time.Now().UTC().Format("20060102150405Z"), randHex(8))
--- a/logtail/logtail.go
+++ b/logtail/logtail.go
@@ -461,10 +461,22 @@ func (l *Logger) upload(ctx context.Context, body []byte, origlen int) (uploaded
 	return true, nil
 }
 
-// Flush uploads all logs to the server.
-// It blocks until complete or there is an unrecoverable error.
+// Flush uploads all logs to the server. It blocks until complete or there is an
+// unrecoverable error.
+//
+// TODO(bradfitz): this apparently just returns nil, as of tailscale/corp@9c2ec35.
+// Finish cleaning this up.
 func (l *Logger) Flush() error {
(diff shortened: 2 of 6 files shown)


ipn/localapi/localapi.go
```diff
@@ -914,6 +915,23 @@ func (h *Handler) serveFileTargets(w http.ResponseWriter, r *http.Request) {
...
+	if !h.PermitWrite {
+		http.Error(w, "access denied", http.StatusForbidden)
+		return
+	}
+	if r.Method != "POST" {
+		http.Error(w, "want POST", http.StatusMethodNotAllowed)
+		return
+	}
+	if h.b.TryFlushLogs() {
+		w.Header().Set("Content-Type", "application/json")
```
mihaip, line 928: `w.WriteHeader(http.StatusNoContent)` seems to be the more common way in which LocalAPI handlers indicate success with no body.

logtail/logtail.go
```diff
@@ -461,12 +461,25 @@ func (l *Logger) upload(ctx context.Context, body []byte, origlen int) (uploaded
 	return true, nil
 }
 
-// Flush uploads all logs to the server.
-// It blocks until complete or there is an unrecoverable error.
+// Flush uploads all logs to the server. It blocks until complete or there is an
+// unrecoverable error.
+//
+// TODO(bradfitz): this apparently just returns nil. Not sure the history of it
```
mihaip, line 467: https://github.com/tailscale/corp/commit/9c2ec3565bb94f1d78de18988eb6b6b793209ed6 appears to be when it became a no-op

ipn/localapi/localapi.go
```diff
@@ -73,6 +73,7 @@ var handler = map[string]localAPIHandler{
 	"dev-set-state-store":         (*Handler).serveDevSetStateStore,
 	"dial":                        (*Handler).serveDial,
 	"file-targets":                (*Handler).serveFileTargets,
+	"flush-logtail":               (*Handler).serveFlushLogtail,
```
mihaip, line 76: Do we need this LocalAPI method? As far as I can tell it's unused. Or were you thinking of add a `tailscale debug` command to also allow log flushing?

mihaip (approved):

ipn/localapi/localapi.go
bradfitz: in this case I'm imagining including content here later, like some stats on what was buffered

ipn/localapi/localapi.go
bradfitz: Yeah, this is mostly for debugging. But I'm also going to make "tailscale bugreport" etc do it.

I suppose I could remove this one and instead make this PR do it on bugreport.

bradfitz: Done

PR #6897, opened 2023-01-05
net/dns,userspace: remove unused DNS paths, increase query limit on iOS

twitchyliquid64:
With https://github.com/tailscale/corp/commit/a42a594bb311686197b64d6389524920a2b07b2a, iOS uses netstack and hence there are no longer any platforms which use the legacy MagicDNS path. As such, we remove it.

We also boop the limit for max in-flight DNS queries on iOS. On all other platforms its 256, so we set it to that and see if we cause memory pressure.


net/dns/manager.go
```diff
@@ -46,10 +43,11 @@ var (
 // the request will be dropped to avoid blocking the caller.
 func maxActiveQueries() int32 {
 	if runtime.GOOS == "ios" {
-		// For memory paranoia reasons on iOS, match the
-		// historical Tailscale 1.x..1.8 behavior for now
-		// (just before the 1.10 release).
-		return 64
+		// For memory paranoia reasons, we have a smaller limit on iOS.
```
danderson, line 46: No need to have the history of changes in the comment, git history will do that.

96 seems fine, although I'd be tempted to remove the special case entirely and see what happens.

danderson (approved):

bradfitz (approved):

net/dns/manager.go
DentonGentry: > Given I've seen this limit hit a few times, it seems prudent to boop it slightly to 96.

Since that limit of 64 was added, we dropped support for iOS versions older than iOS 15. We now have 50 MBytes available on all iOS devices instead of 15 MBytes (https://tailscale.com/blog/go-linker/)

I think we can drop special cases about iOS memory at this point.

twitchyliquid64: I'll remove the special-case.

net/dns/manager.go
```diff
@@ -316,89 +294,6 @@ func toIPsOnly(resolvers []*dnstype.Resolver) (ret []netip.Addr) {
...
-	// Unused space is needed further down the stack. To avoid extra
-	// allocations/copying later on, we allocate such space here.
-	const offset = tstun.PacketStartOffset
-
-	var buf []byte
-	switch {
-	case resp.to.Addr().Is4():
-		h := packet.UDP4Header{
-			IP4Header: packet.IP4Header{
-				Src: magicDNSIP,
```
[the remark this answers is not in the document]
twitchyliquid64: Love that for me! Done :grinning:

PR #6914, opened 2023-01-08
tsnet: add ipn notification callback

morgangallant:
Hey y'all, using `tsnet` for service discovery (i.e. other nodes on the network), figured I'd PR a small change that really helped with this. Previously, if you wanted to keep up to date with peers on the network, you'd have to poll the LocalAPI every few seconds, not super fun. I think `WatchNotifications` itself polls every few seconds under the hood (saw a @bradfitz [TODO](https://github.com/tailscale/tailscale/blob/be67b8e75b9bedb28fc8502c9295fb5f4b3fbb5f/ipn/ipnlocal/local.go#L1809) related to this), though this is likely more efficient than the LocalAPI polling approach mentioned above.

As a (not so great) example, `tsnet` users can now do something like this (to keep track of any peers with matching tags):
```
&tsnet.Server{
	Hostname:  hostname,
	Dir:       dir,
	AuthKey:   authKey,
	Ephemeral: true,
	Logf:      func(format string, args ...any) {},
	NotifyFunc: func(n *ipn.Notify) (keepGoing bool) {
		if n.NetMap != nil {
			ic.mu.Lock()
			existing := make(map[string]struct{}, len(ic.peers))
			for id := range ic.peers {
				existing[id] = struct{}{}
			}
			ic.peers = make(map[string]*tailcfg.Node, len(n.NetMap.Peers))
			for _, p := range n.NetMap.Peers {
				if p.Online == nil || !*p.Online {
					continue
				}
				cloned := p.Clone()
				slices.Sort(cloned.Tags)
				if !slices.Equal(cloned.Tags, ic.selfTags) {
					continue
				}
				ic.peers[cloned.ID.String()] = cloned
				if _, ok := existing[cloned.ID.String()]; ok {
					delete(existing, cloned.ID.String())
				} else {
					log.Printf("found matching peer %q (%v)", cloned.ID, cloned.Addresses)
				}
			}
			for id := range existing {
				log.Printf("peer %q disconnected", id)
			}
			ic.mu.Unlock()
		}
		return true
	},
}
```

Obv. very niche use case, figured I'd open a PR and start a conversation about it. Definitely open to ideas here, and if there's a better approach to solving the same problem, would love to hear it. Thanks!

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -88,6 +88,11 @@ type Server struct {
 	// If empty, the Tailscale default is used.
 	ControlURL string
 
+	// NotifyFunc is an optional callback function for engine updates from
+	// the LocalBackend. This can be used as an alternative to polling the
+	// LocalAPI in client applications.
+	NotifyFunc func(*ipn.Notify) (keepGoing bool)
+
 	initOnce         sync.Once
 	initErr          error
 	lb               *ipnlocal.LocalBackend
@@ -342,6 +347,15 @@ func (s *Server) start() (reterr error) {
 	lb, err := ipnlocal.NewLocalBackend(logf, logid, s.Store, "", s.dialer, eng, loginFlags)
 	if err != nil {
 		return fmt.Errorf("NewLocalBackend: %v", err)
+	}
+	if s.NotifyFunc != nil {
+		notifCtx, cancel := context.WithCancel(context.Background())
+		closePool.addFunc(cancel)
+		go lb.WatchNotifications(
+			notifCtx,
+			ipn.NotifyWatchEngineUpdates,
+			s.NotifyFunc,
+		)
 	}
 	lb.SetVarRoot(s.rootPath)
 	logf("tsnet starting with hostname %q, varRoot %q", s.hostname, s.rootPath)


tsnet/tsnet.go
```diff
@@ -343,6 +348,15 @@ func (s *Server) start() (reterr error) {
 	if err != nil {
 		return fmt.Errorf("NewLocalBackend: %v", err)
 	}
+	if s.NotifyFunc != nil {
+		notifCtx, cancel := context.WithCancel(context.Background())
+		closePool.addFunc(cancel)
```
bradfitz, line 353: closePool only closes on error, not when the tsnet.Server is closed.

tsnet/tsnet.go
```diff
@@ -88,6 +88,11 @@ type Server struct {
 	// If empty, the Tailscale default is used.
 	ControlURL string
 
+	// NotifyFunc is an optional callback function for engine updates from
+	// the LocalBackend. This can be used as an alternative to polling the
+	// LocalAPI in client applications.
+	NotifyFunc func(*ipn.Notify) (keepGoing bool)
```
bradfitz, line 94: I think we need to make this more fine-grained for a stable API we can support long-term. This is kinda exposing some guts.

Also, WatchNotifications without letting the user specify what they want to watch isn't enough.

I'd rather put N different callbacks on here to let tsnet.Server users subscribe to specific changes and then we can map that to whatever the WatchNotification subscription bits are needed over time.

And then we can also shut down all the listeners on Close, and then not expose this keepGoing bool.

bradfitz (changes requested):

morgangallant:
Appreciate the review @bradfitz :) Agree that there's likely a cleaner API here, few options:

1. Could convert this into an issue and use that to have a larger discussion re. the proper design / whether this is the right change to go in & be maintained long-term (apologies for not doing this beforehand; didn't read the contributing guide).
2. Could fix up the shutdown stuff + put in the first of the N callbacks, i.e. `OnNetMapUpdate` or something like that (which'd lay the groundwork for any future callbacks, i.e. engine status, ipn state, prefs, etc).

Happy to do either — lmk!

Issue #6903, opened 2023-01-06
QNAP: cannot authenticate if "Force HTTPS" enabled in General Settings


DentonGentry:
### What is the issue?

Added as a [comment on a pull request](https://github.com/tailscale/tailscale/commit/e32d5af73f5f164722630c23642b5ae3a3ba242c#commitcomment-95298827):
_I'd like to mention that if a user has the setting "Force secure connection (HTTPS) only" enabled under QNAPs General Settings, the above fix will not work and the user will still get the error message indicated._

Indeed, if in the QNAP Control Panel > General Settings one has enabled "Force secure connection (HTTPS) only", the Tailscale app cannot be reached: `Get "https://127.0.0.1:443/cgi-bin/authLogin.cgi?sid=okwrzdwp": x509: cannot validate certificate for 127.0.0.1 because it doesn't contain any IP SANs`
The certificate presented by the QNAP is self-signed and does not validate.

To make this work I think we'd need to add a Transport in https://github.com/tailscale/tailscale/blob/ebbf5c57b32d60a6261aee1a3fe93bed11d884ae/cmd/tailscale/cli/web.go#L235 which disables certificate validation. That isn't very satisfying as it is homeopathic security, but is the only way I can see.

The certificate presented is:
```
Common Name (CN)	QNAP NAS
Organization (O)	QNAP Systems, Inc.
Organizational Unit (OU)	QTS
Common Name (CN)	QNAP NAS
Organization (O)	QNAP Systems, Inc.
Organizational Unit (OU)	QTS
Issued On	Friday, July 22, 2022 at 10:28:46 PM
Expires On	Monday, July 19, 2032 at 10:28:46 PM
SHA-256 Fingerprint	72 EC E6 F8 05 98 F5 31 30 41 F6 CF D3 EC 58 F8
1A 30 14 A9 9E 43 72 39 C4 5D 04 71 F6 B4 38 59
SHA-1 Fingerprint	68 46 D3 D7 62 ED 46 24 73 F7 97 2C 92 9F 6E 7F
F5 CF 69 01
```

### OS

Other

### OS version

QTS 5.0.1.2248

### Tailscale version

1.34.2

PR #6919, opened 2023-01-09
cmd/tailscale: disable HTTPS verification for QNAP auth.

DentonGentry:
QNAP's "Force HTTPS" mode redirects even localhost HTTP to HTTPS, but uses a self-signed certificate which fails verification. We accommodate this by disabling checking of the cert.

Fixes https://github.com/tailscale/tailscale/issues/6903

### DIFF
--- a/cmd/tailscale/cli/web.go
+++ b/cmd/tailscale/cli/web.go
@@ -247,7 +247,14 @@ func qnapAuthnSid(r *http.Request, user, sid string) (string, *qnapAuthResponse,
 }
 
 func qnapAuthnFinish(user, url string) (string, *qnapAuthResponse, error) {
-	resp, err := http.Get(url)
+	// QNAP Force HTTPS mode uses a self-signed certificate. Even importing
+	// the QNAP root CA isn't enough, the cert doesn't have a usable CN nor
+	// SAN. See https://github.com/tailscale/tailscale/issues/6903
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	}
+	client := &http.Client{Transport: tr}
+	resp, err := client.Get(url)
 	if err != nil {
 		return "", nil, err
 	}


cmd/tailscale/cli/web.go
```diff
@@ -247,7 +247,13 @@
 }
 
 func qnapAuthnFinish(user, url string) (string, *qnapAuthResponse, error) {
-	resp, err := http.Get(url)
+	// QNAP Force HTTPS mode uses a self-signed certificate.
+	// https://github.com/tailscale/tailscale/issues/6903
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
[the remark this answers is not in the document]
DentonGentry: It is a well-meaning bot. The point of this PR is that QNAP's feature forces us to use HTTPS and then ignore the certificate it presents.

I'm looking for a way to add a comment or otherwise suppress this finding, we don't want it firing every time someone modifies this file.

cmd/tailscale/cli/web.go
```diff
@@ -247,7 +247,13 @@ func qnapAuthnSid(r *http.Request, user, sid string) (string, *qnapAuthResponse,
 }
 
 func qnapAuthnFinish(user, url string) (string, *qnapAuthResponse, error) {
-	resp, err := http.Get(url)
+	// QNAP Force HTTPS mode uses a self-signed certificate.
+	// https://github.com/tailscale/tailscale/issues/6903
```
maisem, line 251: Can we somehow discover the cert and use that here?

maisem (approved):

cmd/tailscale/cli/web.go
DentonGentry: It currently lives in /etc/ssl/certs/myhost.crt (and another copy of itself in /etc/ssl/certs/myrootca.crt to sign itself with).

I was concerned about the fragility in case QNAP ever moves it, but I suppose it isn't likely to move outside of /etc/ssl.

DentonGentry: I tried to do this, by importing the QNAP root certificate:
```
+func qnapGetInsecureTLSConfig() *tls.Config {
+       return &tls.Config{InsecureSkipVerify: true}
+}
+
+func qnapGetTLSConfig() *tls.Config {
+       rootCAs, err := x509.SystemCertPool()
+       if err != nil {
+               log.Printf("qnapGetTLSConfig SystemCertPool failed: %v, falling back to no verification", err)
+               return qnapGetInsecureTLSConfig()
+       }
+
+       pem, err := os.ReadFile("/etc/ssl/certs/myrootca.crt")
+       if err != nil {
+               log.Printf("qnapGetTLSConfig read myrootca.crt: %v, falling back to no verification", err)
+               return qnapGetInsecureTLSConfig()
+       }
+
+       if !rootCAs.AppendCertsFromPEM(pem) {
+               log.Println("qnapGetTLSConfig AppendCertsFromPEM failed, falling back to no verification", err)
+               return qnapGetInsecureTLSConfig()
+       }
+
+       return &tls.Config{RootCAs: rootCAs}
+}
+
 func qnapAuthnFinish(user, url string) (string, *qnapAuthResponse, error) {
        // QNAP Force HTTPS mode uses a self-signed certificate.
        // https://github.com/tailscale/tailscale/issues/6903
        tr := &http.Transport{
-               TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+               TLSClientConfig: qnapGetTLSConfig(),
        }
        client := &http.Client{Transport: tr}
        resp, err := client.Get(url)
```

However this still doesn't work because though it is now signed by a valid root certificate, none of the CNs match and there are no alternative names:

`Get "https://127.0.0.1:443/cgi-bin/authLogin.cgi?sid=jvys7c89": x509: cannot validate certificate for 127.0.0.1 because it doesn't contain any IP SANs`

The fields in the certificate are:
```
emailAddress = support@qnap.com
CN = QNAP NAS
OU = QTS
O = QNAP Systems, Inc.
L = Taipei
ST = Taipei
C = TW
```

The CN is not a valid hostname, and there are no alternate names.
I don't think this certificate can pass any validation.

maisem: Thanks for testing. Can you leave a comment describing this in the code?

maisem (approved):

DentonGentry:
This will be delivered in a future 1.36.0 release for QNAP.

PR #6927, opened 2023-01-10
go.mod: bump golang.org/x/net and dependencies

willnorris:
I don't think CVE-2022-41717 necessarily impacts us (maybe as part of funnel?), but it came up in a recent security scan so worth updating anyway.


bradfitz:
`go mod tidy` first

PR #6929, opened 2023-01-10
control/controlclient, tailcfg: add Node.Expired field, set for expired nodes

andrew-d:
Change-Id: Ia2bd6b56064416feee28aef5699ca7090940662a


control/controlclient/map.go
```diff
@@ -343,6 +345,40 @@ func undeltaPeers(mapRes *tailcfg.MapResponse, prev []*tailcfg.Node) {
...
+// removeExpiredPeers updates mapRes.Peers to remove all peers that have
+// expired per the ControlTime field in the MapResponse.
+//
+// This is a defense-in-depth against something going wrong with control such
+// that we start seeing expired peers; we'll drop them here, if so.
+func removeExpiredPeers(mapRes *tailcfg.MapResponse) {
+	if mapRes.ControlTime == nil {
+		log.Printf("netmap: removeExpiredPeers: MapResponse.ControlTime not present; skipping")
+		return
+	}
```
bradfitz, line 363: this is going to be spammy and also insufficient: most map responses don't have ControlTime set.

only the rare full path adds that.

and we don't really want those extra bytes all the time for every little update. ControlTime is supposed to be an occasional adjustment thing so the client can learn its offset.

sailorfrag: Storing the difference between control time and local time whenever the field _is_ set may be a close enough approximation, especially if we arrange for it to be sent at least once a day or something. If the difference is small enough (e.g. less than a minute), we could also simply say the local clock is fine and use local time.

sailorfrag: If we never get a ControlTime (does headscale send it?) then I'd recommend just assuming the local time is good.

andrew-d: Okay, I reworked this; we now store a delta on the `mapSession` and use that to calculate the adjusted current time before checking expiry. Thoughts?

Also: @maisem thinks that we shouldn't drop peers here, and should instead mutate their keys to be invalid; if nobody has any objections I can do that next.

bradfitz: I view mutating keys to be invalid as mostly a server-side hack for old clients.

You can _also_ do that for defense in depth, but I'd like to set a "expired bool" on the peer in the netmap so things like "tailscale ping" and magicsock and debug can say/log explicit things about peers being expired rather than obscure wireguard key mismatch errors.

maisem: dropping peers from the netmap would mean that "tailscale status" and the likes would just omit them. That's probably going to result in more support noise. It also has weird implications in terms of subnet routers.

I'd very much prefer if we just somehow blocked traffic to expired peers instead of completely dropping them.

andrew-d: Okay, we now replicate what control does and clear `Endpoints` and `DERP`, clear the node's `Key`, and add an `Expired` boolean and set it to `true` in this codepath.

control/controlclient/map.go
```diff
@@ -343,6 +352,73 @@ func undeltaPeers(mapRes *tailcfg.MapResponse, prev []*tailcfg.Node) {
...
+	if mapRes.ControlTime != nil && !mapRes.ControlTime.IsZero() {
+		delta := mapRes.ControlTime.Sub(now)
+		if delta.Abs() > minClockDelta {
+			log.Printf("[v1] netmap: removeExpiredPeers: setting clock delta to %v", delta)
+			ms.clockDelta = delta
+		}
+	}
+
+	// Adjust our current time by any saved delta to adjust for clock skew.
+	now = now.Add(ms.clockDelta)
```
maisem, line 384: maybe `controlNow := localNow.Add(ms.clockDelta)`?

control/controlclient/map.go
```diff
@@ -343,6 +352,73 @@ func undeltaPeers(mapRes *tailcfg.MapResponse, prev []*tailcfg.Node) {
...
+			continue
+		}
+
+		log.Printf("[v1] netmap: removeExpiredPeers: clearing expired peer %v", peer.StableID)
+		numExpired++
+
+		// Actually mark the node as expired
+		peer.Expired = true
+
+		// Control clears the Endpoints and DERP fields of expired
```
bradfitz, line 403: add another sentence explaining that this is insufficient, though, due to endpoints-over-DERP and DERP discovery via DRPO?

control/controlclient/map.go
```diff
@@ -343,6 +352,73 @@ func undeltaPeers(mapRes *tailcfg.MapResponse, prev []*tailcfg.Node) {
...
+// larger than this, we store an offset in our mapSession to adjust future
+// clock timings.
+const minClockDelta = 1 * time.Minute
+
+// removeExpiredPeers updates mapRes.Peers to remove all peers that have
+// expired per the ControlTime field in the MapResponse.
+//
+// This is a defense-in-depth against something going wrong with control such
+// that we start seeing expired peers; we'll drop them here, if so.
+func (ms *mapSession) removeExpiredPeers(mapRes *tailcfg.MapResponse) {
```
maisem, line 371: old docs

control/controlclient/map.go
```diff
@@ -343,6 +352,73 @@ func undeltaPeers(mapRes *tailcfg.MapResponse, prev []*tailcfg.Node) {
 	mapRes.PeersRemoved = nil
 }
 
+// For extra defense-in-depth, when we're testing expired nodes we check
+// ControlTime against this 'epoch' (set to the approximate time that this code
+// was written) such that if control (or Headscale, etc.) sends a ControlTime
+// that's sufficiently far in the past, we can safely ignore it.
+var removeExpiredPeersEpoch = time.Unix(1673373066, 0)
```
bradfitz, line 359: rename "remove expired peers" here and line 366/371 to something like "breakExpiredPeers" or "flagExpiredPeers" or "noteExpiredPeers" ?

control/controlclient/map_test.go
```diff
@@ -308,6 +308,112 @@ func TestUndeltaPeers(t *testing.T) {
 	}
 }
 
+func TestRemoveExpiredPeers(t *testing.T) {
```
bradfitz, line 311: and here

tailcfg/tailcfg.go
```diff
@@ -255,6 +255,10 @@ type Node struct {
 
 	// DataPlaneAuditLogID is the per-node logtail ID used for data plane audit logging.
 	DataPlaneAuditLogID string `json:",omitempty"`
+
+	// Expired is whether this node's key has expired. This is calculated
+	// based on a timestamp sent from control, to avoid clock skew issues.
+	Expired bool `json:",omitempty"`
```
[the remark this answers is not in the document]
andrew-d: Done

control/controlclient/map.go
andrew-d: Done

control/controlclient/map.go
```diff
@@ -343,6 +350,75 @@ func undeltaPeers(mapRes *tailcfg.MapResponse, prev []*tailcfg.Node) {
...
+// DERP field.
+func (ms *mapSession) flagExpiredPeers(mapRes *tailcfg.MapResponse) {
+	localNow := clockNow()
+
+	// If we have a ControlTime field, update our delta.
+	if mapRes.ControlTime != nil && !mapRes.ControlTime.IsZero() {
+		delta := mapRes.ControlTime.Sub(localNow)
+		if delta.Abs() > minClockDelta {
+			log.Printf("[v1] netmap: flagExpiredPeers: setting clock delta to %v", delta)
+			ms.clockDelta = delta
```
maisem, line 382: do we have to worry about resetting this to 0?

andrew-d:
Also per @maisem's request, we now mutate the node's key instead of replacing it with `1`, `2`, `3`, etc.

control/controlclient/map.go
andrew-d: Hm, for safety's sake, yes, added.

types/key/node.go
```diff
@@ -193,6 +193,23 @@ func NodePublicFromRaw32(raw mem.RO) NodePublic {
 	return ret
 }
 
+// badOldPrefix is a nodekey/discokey prefix that, when base64'd, serializes
+// with a "bad01" ("bad ol'", ~"bad old") prefix. It's used for expired node
+// keys so when we debug a customer issue, the "bad01" can jump out to us. See:
+//
+//	https://github.com/tailscale/corp/issues/8613
```
bradfitz, line 200: links to corp issues aren't really appropriate in oss.

control/controlclient/map.go
```diff
@@ -85,6 +91,7 @@ func (ms *mapSession) addUserProfile(userID tailcfg.UserID) {
 // information from prior MapResponse values.
 func (ms *mapSession) netmapForResponse(resp *tailcfg.MapResponse) *netmap.NetworkMap {
 	undeltaPeers(resp, ms.previousPeers)
+	ms.flagExpiredPeers(resp)
```
maisem, line 94: we probably need to call this with a timer for cases where control doesn't send netmap updates for a long time (or is unreachable).

control/controlclient/map.go
```diff
@@ -343,6 +350,77 @@ func undeltaPeers(mapRes *tailcfg.MapResponse, prev []*tailcfg.Node) {
...
+		// Nodes that don't expire have KeyExpiry set to the zero time
+		if peer.KeyExpiry.IsZero() || peer.KeyExpiry.After(controlNow) {
+			continue
+		}
+
+		log.Printf("[v1] netmap: flagExpiredPeers: clearing expired peer %v", peer.StableID)
+		numExpired++
+
+		// Actually mark the node as expired
+		peer.Expired = true
```
bradfitz, line 406: if it's already true, avoid some of this? at least avoid the log.Printf above?

also, I'm concerned that log print will be very noisy, even at v1. can you only log on transitions, keeping state on mapSession?

control/controlclient/map.go
andrew-d: Yep, planning on following up with that in another PR; it's a bit tricky since I think we need to create a new `NetMap`, since we can't mutate the existing one

types/key/node.go
andrew-d: Fixed

control/controlclient/map.go
andrew-d: Added a tracking `map` to only print on transition now

control/controlclient/map.go
```diff
@@ -60,11 +67,12 @@ type mapSession struct {
...
-		logf:            logger.Discard,
-		vlogf:           logger.Discard,
-		lastDNSConfig:   new(tailcfg.DNSConfig),
-		lastUserProfile: map[tailcfg.UserID]tailcfg.UserProfile{},
+		privateNodeKey:    privateNodeKey,
+		logf:              logger.Discard,
+		vlogf:             logger.Discard,
+		lastDNSConfig:     new(tailcfg.DNSConfig),
+		lastUserProfile:   map[tailcfg.UserID]tailcfg.UserProfile{},
+		previouslyExpired: make(map[tailcfg.StableNodeID]bool),
```
bradfitz, line 75: the previous line uses a different init style. match, especially since so close.

control/controlclient/map.go
```diff
@@ -343,6 +352,81 @@ func undeltaPeers(mapRes *tailcfg.MapResponse, prev []*tailcfg.Node) {
...
+	if controlNow.Before(flagExpiredPeersEpoch) {
+		log.Printf("netmap: flagExpiredPeers: [unexpected] delta-adjusted current time is before hardcoded epoch; skipping")
+		return
+	}
+
+	for _, peer := range mapRes.Peers {
+		// Nodes that don't expire have KeyExpiry set to the zero time;
+		// skip those and peers that are already marked as expired
+		// (e.g. from control).
+		if peer.KeyExpiry.IsZero() || peer.KeyExpiry.After(controlNow) || peer.Expired {
```
bradfitz, line 401: if server says peer.Expired is true, you don't want to do the map delete on line 402 I don't think.

add a new if peer.Expired { continue } after line 404?

control/controlclient/map.go
```diff
@@ -343,6 +352,81 @@ func undeltaPeers(mapRes *tailcfg.MapResponse, prev []*tailcfg.Node) {
...
+			delete(ms.previouslyExpired, peer.StableID)
+			continue
+		}
+
+		if !ms.previouslyExpired[peer.StableID] {
+			log.Printf("[v1] netmap: flagExpiredPeers: clearing expired peer %v", peer.StableID)
```
bradfitz, line 407: ms.logf, not log.Printf.

here and line 393 etc too

control/controlclient/map_test.go
```diff
@@ -321,6 +427,12 @@ func formatNodes(nodes []*tailcfg.Node) string {
 		if n.LastSeen != nil {
 			extra += fmt.Sprintf(", lastSeen=%v", n.LastSeen.Unix())
 		}
+		if n.Key != (key.NodePublic{}) {
+			extra += fmt.Sprintf(", key=%v", n.Key.String())
```
bradfitz, line 431: phew, this += string stuff is only in tests :)

extra silly that it's being done in a func that's already using strings.Builder 😂

bradfitz (approved):
LGTM otherwise

control/controlclient/map.go
andrew-d: Done

control/controlclient/map.go
andrew-d: Done; also updated the test helper to log properly

control/controlclient/map_test.go
andrew-d: Heh, fixed that up while I was here 😃

andrew-d:
Okay, looked at this with fresh eyes this morning ✅

PR #6933, opened 2023-01-10
ipn/ipnlocal: disallow unsigned peers from WoL

raggi:
Unsigned peers should not be allowed to generate Wake-on-Lan packets,
only access Funnel.

Updates #6934
Updates #7515
Updates #6475


danderson (approved):
Mildly confused about what signed means here, is it signed as in tailnet lock? From the linked bug it sounds like it's related to funnel's special locked-down status, but that's orthogonal I thought... Anyway, change itself LGTM, just a little unfamiliar with the terminology in use.

twitchyliquid64 (approved):

PR #6936, opened 2023-01-11
cmd/tailscale: fix race in tailscale up's state transition handling

saljmz:
This change delays the first flush in the /watch-ipn-bus/ handler
until after the watcher has been successfully installed on the IPN
bus. It does this by adding a new onWatchAdded callback to
LocalBackend.WatchNotifications().

Without this, the endpoint returns a 200 almost immediatly, and
only then installs a watcher for IPN events.  This means there's a
small window where events could be missed by clients after calling
WatchIPNBus().

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -1734,13 +1734,17 @@ func (b *LocalBackend) readPoller() {
 //
 // WatchNotifications blocks until ctx is done.
 //
-// The provided fn will only be called with non-nil pointers. The caller must
-// not modify roNotify. If fn returns false, the watch also stops.
+// The provided onWatchAdded, if non-nil, will be called once the watcher
+// is installed.
+//
+// The provided fn will be called for each notification. It will only be
+// called with non-nil pointers. The caller must not modify roNotify. If
+// fn returns false, the watch also stops.
 //
 // Failure to consume many notifications in a row will result in dropped
 // notifications. There is currently (2022-11-22) no mechanism provided to
 // detect when a message has been dropped.
-func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWatchOpt, fn func(roNotify *ipn.Notify) (keepGoing bool)) {
+func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWatchOpt, onWatchAdded func(), fn func(roNotify *ipn.Notify) (keepGoing bool)) {
 	ch := make(chan *ipn.Notify, 128)
 
 	origFn := fn
@@ -1789,6 +1793,10 @@ func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWa
 		delete(b.notifyWatchers, handle)
 		b.mu.Unlock()
 	}()
+
+	if onWatchAdded != nil {
+		onWatchAdded()
+	}
 
 	if ini != nil {
 		if !fn(ini) {
--- a/ipn/ipnlocal/local_test.go
+++ b/ipn/ipnlocal/local_test.go
@@ -5,6 +5,7 @@
 package ipnlocal
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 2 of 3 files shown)


bradfitz (commented):
I might not get to this today but I want to review it carefully and understand it.

ipn/ipnlocal/local.go
```diff
@@ -1732,22 +1732,21 @@ func (b *LocalBackend) readPoller() {
...
-// The provided fn will only be called with non-nil pointers. The caller must
-// not modify roNotify. If fn returns false, the watch also stops.
+// The returned channel will only receive non-nil pointers. The caller must
+// not modify the values read from the channel.
 //
 // Failure to consume many notifications in a row will result in dropped
 // notifications. There is currently (2022-11-22) no mechanism provided to
 // detect when a message has been dropped.
-func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWatchOpt, fn func(roNotify *ipn.Notify) (keepGoing bool)) {
+func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWatchOpt) <-chan *ipn.Notify {
```
bradfitz, line 1741: Channels in APIs are very rarely the right path. I'd really rather not.

cmd/tailscale/cli/up.go
```diff
@@ -543,7 +543,12 @@ func runUp(ctx context.Context, cmd string, args []string, upArgs upArgsT) (retE
 
 	watchCtx, cancelWatch := context.WithCancel(ctx)
 	defer cancelWatch()
-	watcher, err := localClient.WatchIPNBus(watchCtx, 0)
+	watcher, err := localClient.WatchIPNBus(watchCtx, ipn.NotifyInitialState)
+	if err != nil {
+		return err
+	}
+	// Read the initial state and throw it away, to make sure the watcher is watching before we call Start().
+	_, err = watcher.Next()
```
bradfitz, line 551: This is at least a succinct definition of the problem, but the intent of the API was that this was _already_ the case, so I'd prefer to fix the API somehow than require all callers to cargo-cult this.

ipn/ipnlocal/local.go
bradfitz: Per chat, let's go with (3) and then:

```patch
diff --git a/ipn/localapi/localapi.go b/ipn/localapi/localapi.go
index 9cefb172..97a3aa0e 100644
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -721,7 +721,6 @@ func (h *Handler) serveWatchIPNBus(w http.ResponseWriter, r *http.Request) {
                return
        }
-       f.Flush()
 
        var mask ipn.NotifyWatchOpt
        if s := r.FormValue("mask"); s != "" {
@@ -733,7 +732,8 @@ func (h *Handler) serveWatchIPNBus(w http.ResponseWriter, r *http.Request) {
                mask = ipn.NotifyWatchOpt(v)
        }
        ctx := r.Context()
-       h.b.WatchNotifications(ctx, mask, func(roNotify *ipn.Notify) (keepGoing bool) {
+
+       h.b.WatchNotifications(ctx, mask, f.Flush, func(roNotify *ipn.Notify) (keepGoing bool) {
                js, err := json.Marshal(roNotify)
                if err != nil {
                        h.logf("json.Marshal: %v", err)
```

ipn/ipnlocal/local.go
```diff
@@ -1734,13 +1734,16 @@ func (b *LocalBackend) readPoller() {
 //
 // WatchNotifications blocks until ctx is done.
 //
-// The provided fn will only be called with non-nil pointers. The caller must
-// not modify roNotify. If fn returns false, the watch also stops.
+// The provided onWatchAdded will be called once the watched is installed.
```
bradfitz, line 1737: The provided onWatchAdded, if non-nil, will ...

ipn/ipnlocal/local_test.go
```diff
@@ -820,3 +821,29 @@ type legacyBackend interface {
 // Verify that LocalBackend still implements the legacyBackend interface
 // for now, at least until the macOS and iOS clients move off of it.
 var _ legacyBackend = (*LocalBackend)(nil)
+
+func TestWatchNotificationsCallbacks(t *testing.T) {
+	b := new(LocalBackend)
+	n := new(ipn.Notify)
+	b.WatchNotifications(context.Background(), 0, func() {
+		// Ensure a watcher has been installed.
+		if len(b.notifyWatchers) != 1 {
```
bradfitz, line 830: this is a potential data race I think? This callback is called with the lock not held, so you should probably grab it here. 

And everything else in this test: line 835, and 846?

bradfitz (approved):

bradfitz:
Added a `Fixes #nnnn` line to the commit message.

ipn/ipnlocal/local_test.go
```diff
@@ -820,3 +821,32 @@ type legacyBackend interface {
...
+		defer b.mu.Unlock()
+
+		// Ensure a watcher has been installed.
+		if len(b.notifyWatchers) != 1 {
+			t.Fatalf("unexpected number of watchers in new LocalBackend, want: 1 got: %v", len(b.notifyWatchers))
+		}
+		// Send a notification. Range over notifyWatchers to get the channel
+		// because WatchNotifications doesn't expose the handle for it.
+		for _, c := range b.notifyWatchers {
+			c <- n
```
bradfitz, line 839: can this send block?

maybe put it in a select with a default case that does a t.Fatal?

ipn/ipnlocal/local_test.go
```diff
@@ -820,3 +821,32 @@ type legacyBackend interface {
...
+		}
+	}, func(roNotify *ipn.Notify) bool {
+		if roNotify != n {
+			t.Fatalf("unexpected notification received. want: %v got: %v", n, roNotify)
+		}
+		return false
+	})
+
+	// Ensure watchers have been cleaned up.
+	if len(b.notifyWatchers) != 0 {
```
bradfitz, line 849: shouldn't be concurrent access here, but might want to do the lock/defer unlock here too just for consistency?

bradfitz (approved):
🚢

PR #6938, opened 2023-01-11
net/flowtrack,wgengine/filter: refactor Cache to use generics

twitchyliquid64:
Now that we have generics.

### DIFF
--- a/wgengine/filter/filter.go
+++ b/wgengine/filter/filter.go
@@ -59,7 +59,7 @@ type Filter struct {
 // filterState is a state cache of past seen packets.
 type filterState struct {
 	mu  sync.Mutex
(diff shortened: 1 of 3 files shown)


wgengine/filter/filter.go
```diff
@@ -59,7 +59,7 @@ type Filter struct {
 // filterState is a state cache of past seen packets.
 type filterState struct {
 	mu  sync.Mutex
-	lru *flowtrack.Cache // from flowtrack.Tuple -> nil
+	lru *flowtrack.Cache[struct{}] // from flowtrack.Tuple -> nil
```
bradfitz, line 62: the nil part isn't right anymore

bradfitz (approved):

PR #6939, opened 2023-01-11
types/views: add SliceEqualAnyOrder func

willnorris:
This is a copy of the tagsEqual func from corp/control/control.go, moved here so that it can be reused in other places.

I'm not sure if there is a better name for this, since these slices aren't truly equal, since they can be in different order.  But it's also not just a "ContainsAll" function.  Open to suggestions.

### DIFF
--- a/types/views/views.go
+++ b/types/views/views.go
@@ -170,6 +170,37 @@ func SliceContains[T comparable](v Slice[T], e T) bool {
 	return false
 }
 
+// SliceEqualAnyOrder reports whether a and b contain the same elements, regardless of order.
+// The underlying slices for a and b can be nil.
+func SliceEqualAnyOrder[T comparable](a, b Slice[T]) bool {
+	if a.Len() != b.Len() {
+		return false
+	}
+
+	var diffStart int // beginning index where a and b differ
+	for n := a.Len(); diffStart < n; diffStart++ {
+		if a.At(diffStart) != b.At(diffStart) {
+			break
+		}
+	}
+	if diffStart == a.Len() {
+		return true
+	}
+
+	// count the occurrences of remaining values and compare
+	valueCount := make(map[T]int)
+	for i, n := diffStart, a.Len(); i < n; i++ {
+		valueCount[a.At(i)]++
+		valueCount[b.At(i)]--
+	}
+	for _, count := range valueCount {
+		if count != 0 {
+			return false
+		}
+	}
+	return true
+}
+
 // IPPrefixSlice is a read-only accessor for a slice of netip.Prefix.
 type IPPrefixSlice struct {
 	ж Slice[netip.Prefix]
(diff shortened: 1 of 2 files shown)


types/views/views.go
```diff
@@ -170,6 +170,23 @@ func SliceContains[T comparable](v Slice[T], e T) bool {
 	return false
 }
 
+// SliceEquals reports whether a and b contain the same elements, regardless of order.
+func SliceEquals[T comparable](a, b Slice[T]) bool {
```
bradfitz, line 174: this sounds too much like https://pkg.go.dev/golang.org/x/exp/slices#Equal which is NOT this function.

Let's name this something uglier like `SliceEqualAnyOrder`

Then we can reserve the name `SliceEqual` for one that just forwards to x/exp/slices.Equal.

Clarify the nilness of the slice doesn't matter too? (unlike reflect.DeepEqual)

types/views/views.go
```diff
@@ -170,6 +170,23 @@ func SliceContains[T comparable](v Slice[T], e T) bool {
...
+	if a.Len() != b.Len() {
+		return false
+	}
+	aHasValue := make(map[T]bool)
+	for i, n := 0, a.Len(); i < n; i++ {
+		aHasValue[a.At(i)] = true
```
bradfitz, line 180: are [A, B, B, C, C, C]
and [A, B, B, B, C, C] equal?

Under this implementation, they are. But that's kinda weird.

We really want to count the number of occurrences of each.

types/views/views.go
```diff
@@ -170,6 +170,23 @@ func SliceContains[T comparable](v Slice[T], e T) bool {
 	return false
 }
 
+// SliceEquals reports whether a and b contain the same elements, regardless of order.
+func SliceEquals[T comparable](a, b Slice[T]) bool {
+	if a.Len() != b.Len() {
+		return false
+	}
+	aHasValue := make(map[T]bool)
```
bradfitz, line 178: seems like we should try our hardest to match index-for-index between the two slices first before allocating and see how far we get. Then we only need to this allocating slow path for the remainder bit that wasn't in order.

that is, for two slices a and b both of length 10, maybe the first 5 elements of each are equal, but the last 5 are mixed up.  then we only need to do this slow path for the last 5.

willnorris:
Done. PTAL

types/views/views.go
```diff
@@ -170,6 +170,37 @@ func SliceContains[T comparable](v Slice[T], e T) bool {
...
+	for n := a.Len(); diffStart < n; diffStart++ {
+		if a.At(diffStart) != b.At(diffStart) {
+			break
+		}
+	}
+	if diffStart == a.Len() {
+		return true
+	}
+
+	// count the occurances of remaining values and compare
```
bradfitz, line 190: typo occurrences

bradfitz (approved):

PR #6942, opened 2023-01-12
tailcfg: document zero value for KeyExpiry

andrew-d:
Change-Id: I50889f0205ecf66c415f50f9019c190448c991fc

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -188,7 +188,7 @@ type Node struct {
 	Sharer UserID `json:",omitempty"`
 
 	Key          key.NodePublic
-	KeyExpiry    time.Time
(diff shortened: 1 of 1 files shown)


tailcfg/tailcfg.go
```diff
@@ -188,7 +188,7 @@ type Node struct {
 	Sharer UserID `json:",omitempty"`
 
 	Key          key.NodePublic
-	KeyExpiry    time.Time
+	KeyExpiry    time.Time                  // zero if this node does not expire (e.g. tagged nodes)
```
bradfitz, line 191: I'd remove the "tagged nodes" bit. That's a bit of a distraction.

And maybe say "zero value" to help people avoid thinking 0 unix epoch time?  "zero value means no expiry" ?

bradfitz (approved):

PR #6946, opened 2023-01-12
net/portmapper: relax handling of UPnP resp

### LINKED ISSUE #5502: NAT-PMP will only accept default gateway IP as IGD
### What is the issue?

Tailscale clients could not detect the UPnP nor NAT-PMP capabilities of my OPNSense failover pair of routers. UPnP & NAT-PMP were enabled, and other devices could discover and use these services.

I eventually discovered that the IGD response from the currently-primary router was coming from its LAN address instead of the gateway IP, which is a virtual (floating) IP that can failover to the secondary router. Because of this, tailscale client seemed unable to discover UPnP or NAT-PMP.

I successfully worked around this issue by forwarding port 5351/udp from the gateway IP to the primary router's LAN address, which allowed the tailscale client to discover NAT-PMP (but not UPnP).

I expected tailscale to detect these capabilities, or expose configuration to override its (arguably reasonable) default behavior to only accept the system's default gateway as IGD.

### Steps to reproduce

1. Use OPNSense in failover pair with virtual IP as default gateway.
2. Enable UPnP & NAT-PMP in OPNSense.

Unclear if this is recent behavior, though PR #3202 seems relevant.

### OS

Linux, macOS

### OS version

Ubuntu 22.04, macOS 12.5

### Tailscale version

1.28.0

### Bug report

_No response_

jwhited:
Gateway devices operating as an HA pair w/VRRP or CARP may send UPnP replies from static addresses rather than the floating gateway address. This commit relaxes our source address verification such that we parse responses from non-gateway IPs, and re-point the UPnP root desc URL to the gateway IP. This ensures we are still interfacing with the gateway device (assuming L2 security intact), even though we got a root desc from a non-gateway address.

This relaxed handling is required for ANY port mapping to work on certain OPNsense/pfsense distributions using CARP at the time of writing, as miniupnpd may only listen on the static, non-gateway interface address for PCP and PMP.

Fixes #5502

### DIFF
--- a/net/portmapper/portmapper.go
+++ b/net/portmapper/portmapper.go
@@ -100,11 +100,11 @@ type mapping interface {
 	// but can be called asynchronously. Release should be idempotent, and thus even if called
 	// multiple times should not cause additional side-effects.
 	Release(context.Context)
-	// goodUntil will return the lease time that the mapping is valid for.
+	// GoodUntil will return the lease time that the mapping is valid for.
 	GoodUntil() time.Time
-	// renewAfter returns the earliest time that the mapping should be renewed.
+	// RenewAfter returns the earliest time that the mapping should be renewed.
 	RenewAfter() time.Time
-	// externalIPPort indicates what port the mapping can be reached from on the outside.
+	// External indicates what port the mapping can be reached from on the outside.
 	External() netip.AddrPort
 }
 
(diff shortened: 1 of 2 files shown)


StraightfaceStudios:
UPnP is fixed but NAT-PMP still fails with CARP floating IPs.

PR #6948, opened 2023-01-13
HTTP headers are supposed to be handled case insensitively

gabm:
Ran into a problem with apache that sends "WebSocket" to backend servers when reverse proxying a websocket. Tailscale does not accept that string.

According to [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-upgrade) names should be handled case-insensitively


bradfitz:
Per https://github.com/tailscale/tailscale#contributing, please adjust the commit message style to be idiomatic for this repo.

And add the `"strings"` import so it compiles.

Otherwise looks good, thanks!

gabm:
thanks for the feedback, should be fixed

bradfitz:
Fixed in #6955

PR #6951, opened 2023-01-13
cmd/tailscale: correct order for -terminate-tls flag in serve tcp usage

saljmz:
The -terminate-tls flag is for the tcp subsubcommand, not the serve subcommand like the usage example suggests.


bradfitz (approved):

shayne:
We should also update https://github.com/tailscale/tailscale/blob/25a0091f69880169ef8359f6b59bf2c5d737f03d/cmd/tailscale/cli/serve.go#L588

shayne (approved):
Thanks!

PR #6957, opened 2023-01-13
cmd/tailscale/cli: make web show/link Synology outgoing connection mode/docs

### LINKED ISSUE #4015: Synology: make the web UI show which tun mode it's in
We should make the Synology web UI show whether it's using netstack or kernel mode for tun and have a link to the KB on how to upgrade it to kernel mode.

/cc @maisem @crawshaw

bradfitz:
Fixes #4015


bradfitz:
Not the prettiest, but something:

Screenshot 2023-01-13 at 1 27 59 PM

Somebody on @tailscale/frontend want to suggest a better style?

Issue #2163, opened 2021-06-17
GRE L2/L3 tunneling protocol support


tinoue:
I made PR to support GRE tunneling protocol over tailscale. This supports L2/L3 tunneling over IPv4 (GRE over IPv6 is not functional because tailscale MTU is 1280 which is minimum value for IPv6).

https://github.com/tailscale/tailscale/pull/2162

bradfitz:
Thanks! We're about to cut a 1.10 release, so we can merge this after 1.10 is out for the 1.12 release. (But there will be unstable builds out earlier)

DentonGentry:
The 1.10 release branch was created yesterday, this feature can go into the tree (after PR review) for the 1.12 release.

DentonGentry:
Fixed in 1.20 by https://github.com/tailscale/tailscale/commit/69de3bf7bfddb37b4c0e076c93115f82a51ec407

PR #6965, opened 2023-01-14
net/{packet,tstun},wgengine/filter: fix unknown IP protocol handling

### LINKED ISSUE #6423: FR: Unable to run EOIP (Protocol 97) between two nodes

Building an EOIP tunnel between two nodes on two subnet routers, with snat disabled.

TCP and UDP pass in both directions, however, protocol 97, EoIP does not traverse the tailscale tunnel as expected.

### How should we solve this?

Native wireguard allows EoIP. Tailscale should as well.

bradfitz:
01b90df2fa4f9101e4f0ae8334b00dd9c3ccc148 added SCTP support before (with explicit parsing for ports) and
69de3bf7bfddb37b4c0e076c93115f82a51ec407 tried to add support for arbitrary IP protocols (as long as the ACL permited a port of "*", since we might not know how to find ports from an arbitrary IP protocol, if it even has such a concept). But apparently that latter commit wasn't tested end-to-end enough. It had a lot of tests, but the tests made assumptions about layering that either weren't true, or regressed since 1.20. Notably, it didn't remove the (*Filter).pre bidirectional filter that dropped all "unknown" protocol packets both leaving and entering, even if there were explicit protocol matches allowing them in.

Also, don't map all unknown protocols to 0. Keep their IP protocol number parsed so it's matchable by later layers. Only reject illegal things.

Fixes #6423
Updates #2162
Updates #2163

### DIFF
--- a/net/packet/packet_test.go
+++ b/net/packet/packet_test.go
@@ -6,12 +6,16 @@ package packet
 
 import (
 	"bytes"
+	"encoding/hex"
 	"net/netip"
 	"reflect"
+	"strings"
 	"testing"
+	"unicode"
 
 	"tailscale.com/tstest"
 	"tailscale.com/types/ipproto"
+	"tailscale.com/util/must"
 )
 
 const (
@@ -440,6 +444,17 @@ func TestParsedString(t *testing.T) {
 	}
 }
 
+// mustHexDecode is like hex.DecodeString, but panics on error
+// and ignores whitespcae in s.
+func mustHexDecode(s string) []byte {
+	return must.Get(hex.DecodeString(strings.Map(func(r rune) rune {
+		if unicode.IsSpace(r) {
+			return -1
+		}
+		return r
+	}, s)))
+}
+
 func TestDecode(t *testing.T) {
 	tests := []struct {
 		name string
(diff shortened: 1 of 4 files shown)


net/tstun/wrap_test.go
```diff
@@ -310,8 +312,9 @@ func TestFilter(t *testing.T) {
 		drop bool
 		data []byte
 	}{
-		{"junk_in", in, true, []byte("\x45not a valid IPv4 packet")},
-		{"junk_out", out, true, []byte("\x45not a valid IPv4 packet")},
+		{"short_in", in, true, []byte("\x45xxx")},
+		{"short_out", out, true, []byte("\x45xxx")},
+		{"ip97_out", out, false, must.Get(hex.DecodeString(strings.ReplaceAll("4500 0019 d186 4000 4061 751d 644a 4603 6449 e549 6865 6c6c 6f", " ", "")))},
```
DentonGentry, line 317: I note that packet_test.go defined a noWS() function to do this for all unicode whitespace runes, though I don't see why it was needed there.

DentonGentry (approved):

net/tstun/wrap_test.go
bradfitz: made them both consistent with a new mustDecodeHex func. maybe I'll move that to some packet test helper package later.

DentonGentry (approved):

net/packet/packet_test.go
```diff
@@ -440,6 +444,17 @@ func TestParsedString(t *testing.T) {
 	}
 }
 
+// mustHexDecode is like hex.DecodeString, but panics on error
+// and ignores whitespcae in s.
```
andrew-d, line 448:
```suggestion
// and ignores whitespace in s.
```

net/tstun/wrap_test.go
```diff
@@ -293,6 +296,17 @@ func TestWriteAndInject(t *testing.T) {
 	}
 }
 
+// mustHexDecode is like hex.DecodeString, but panics on error
+// and ignores whitespcae in s.
```
andrew-d, line 300:
```suggestion
// and ignores whitespace in s.
```

net/packet/packet_test.go
bradfitz: https://github.com/tailscale/tailscale/pull/6968

PR #6971, opened 2023-01-15
cmd/tailscale, logtail: add 'tailscale debug daemon-logs' logtail mechanism

### LINKED ISSUE #6836: cmd/tailscale{,d}: add daemon log tail mechanism
Getting at the tailscaled logs varies by platform and is particularly annoying on some.

Let's add a portable log viewing mechanism:

* add LocalAPI method to tap/stream logs, hooking into logtail
* add CLI `tailscale debug daemon-logs` or something (akin to `debug daemon-goroutines`) to use that LocalAPI method and stream
* by default only follow and don't keep any history
* optional, later: add in a small window of history server-side for client to optionally request (e.g. 1MB). so then you can restart tailscaled and get the early stuff from the CLI.

bradfitz:
Fixes #6836

### DIFF
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -34,6 +34,7 @@ import (
 	"tailscale.com/ipn"
 	"tailscale.com/ipn/ipnlocal"
 	"tailscale.com/ipn/ipnstate"
+	"tailscale.com/logtail"
 	"tailscale.com/net/netutil"
 	"tailscale.com/safesocket"
 	"tailscale.com/tailcfg"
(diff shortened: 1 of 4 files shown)


ipn/localapi/localapi.go
```diff
@@ -421,6 +423,45 @@ func (h *Handler) serveGoroutines(w http.ResponseWriter, r *http.Request) {
...
 
+// serveLogTap taps into the tailscaled/logtail server output and streams
+// it to the client.
+func (h *Handler) serveLogTap(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	// Require write access (~root) as the logs could contain something
+	// sensitive.
+	if !h.PermitWrite {
+		http.Error(w, "goroutine dump access denied", http.StatusForbidden)
```
DentonGentry, line 434: :s/goroutine dump/logtap/

DentonGentry (approved):

Issue #6995, opened 2023-01-18
cmd/tailscale: add "update" command


bradfitz:
We'd like to have a portable CLI command to do upgrades, downgrades, track switches, etc across a bunch of platforms. At least Windows, common Linux distros (Debian, Ubuntu) on the host (not containers), Synology, tailscaled-on-macOS, maybe the macOS GUI builds too (after figuring out some Sparkle stuff).

Earlier commit starting to add some of this was d9144c73a8ed520c8b59b604323acca7d879a568.

bradfitz:
Related, but distinct: #755 (windows auto-update). This bug isn't about auto-updating necessarily, except in that people could set this to run as a scheduled task. But there will be real Windows auto-update support later. This bug's about the CLI.

Foxboron:
Code like this should not exist on rolling release distros like Arch. Running `tailscale update` on Arch *is* going to break peoples systems sooner or later.

I need a way to disable this in the Arch package.

Foxboron:
Void Linux has patched it out; https://github.com/void-linux/void-packages/commit/60610c8b1081d300e1358a0626a94d13931021dc

classabbyamp:
This should only be enabled in versions of tailscale that come from and are built by tailscale directly, IMO. When tailscale (or any package) comes from a central source like a linux distro's package manager, that should be the only method of doing updates, and users know this. There is no need to add a "special" method of updating tailscale, even if it's just invoking `apt` or `apk` or `pacman` behind the scenes.

I suggest putting this behind a [build tag](https://pkg.go.dev/go/build#hdr-Build_Constraints), like `go build -tags selfupdate ...`, so that it is disabled by default.

awly:
@Foxboron removed Arch support from `tailscale update` in https://github.com/tailscale/tailscale/pull/9081
The command is still there, but it always returns an error on Arch now.

Foxboron:
@awly thanks!

PR #6996, opened 2023-01-18
cmd/tailscale/cli: make "update" work on Windows

bradfitz:
cmd/tailscale/cli: make "update" work on Windows

Updates #6995

```
Co-authored-by: Aaron Klotz 
```


cmd/tailscale/cli/update.go
```diff
@@ -43,7 +55,22 @@ var updateArgs struct {
 	version string // explicit version; empty means auto
 }
 
+// winMSIEnv is the environment variable that, if set, contains makes the update
```
DentonGentry, line 58: "contains makes the update command"
I think "contains" was intended to be removed from the sentence.

cmd/tailscale/cli/update.go
```diff
@@ -200,6 +271,169 @@ func (up *updater) updateWindows() error {
...
+
+	log.Printf("copying tailscaled.exe to copy...")
+	selfCopy, err := makeSelfCopy()
+	if err != nil {
+		return err
+	}
+	defer os.Remove(selfCopy)
+	log.Printf("running copy of self...")
+
+	cmd := exec.Command(selfCopy, "update")
```
DentonGentry, line 295: msiexec handles signature checking? I suppose it would.

cmd/tailscale/cli/update.go
```diff
@@ -200,6 +271,169 @@ func (up *updater) updateWindows() error {
...
+	selfExe, err := os.Executable()
+	if err != nil {
+		return "", err
+	}
+	f, err := os.Open(selfExe)
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+	f2, err := os.CreateTemp("", "tailscale-updater-*.exe")
```
DentonGentry, line 329: This seems like this might trigger Windows Defender, writing out an executable and running it.

cmd/tailscale/cli/update.go
```diff
@@ -200,6 +271,169 @@ func (up *updater) updateWindows() error {
...
+	}
+	if err := of.Close(); err != nil {
+		return err
+	}
+	pw.print()
+
+	if !bytes.Equal(hash.Sum(nil), wantHash) {
+		return fmt.Errorf("SHA-256 of downloaded MSI didn't match expected value")
+	}
+	log.Printf("hash matched")
```
DentonGentry, line 417: In https://github.com/tailscale/tailscale/pull/6370 we added a way to set the Zone.Identifier of a downloaded file. Should we do that here?

cmd/tailscale/cli/update.go
bradfitz: Presumably!

The wireguard-windows MSI updater additionally does:

```go
func verifyAuthenticode(path string) bool {
	path16, err := windows.UTF16PtrFromString(path)
	if err != nil {
		return false
	}
	data := &windows.WinTrustData{
		Size:             uint32(unsafe.Sizeof(windows.WinTrustData{})),
		UIChoice:         windows.WTD_UI_NONE,
		RevocationChecks: windows.WTD_REVOKE_WHOLECHAIN, // Full revocation checking, as this is called with network connectivity.
		UnionChoice:      windows.WTD_CHOICE_FILE,
		StateAction:      windows.WTD_STATEACTION_VERIFY,
		FileOrCatalogOrBlobOrSgnrOrCert: unsafe.Pointer(&windows.WinTrustFileInfo{
			Size:     uint32(unsafe.Sizeof(windows.WinTrustFileInfo{})),
			FilePath: path16,
		}),
	}
	verified := windows.WinVerifyTrustEx(windows.InvalidHWND, &windows.WINTRUST_ACTION_GENERIC_VERIFY_V2, data) == nil
	data.StateAction = windows.WTD_STATEACTION_CLOSE
	windows.WinVerifyTrustEx(windows.InvalidHWND, &windows.WINTRUST_ACTION_GENERIC_VERIFY_V2, data)
	return verified
}
```

I suppose that can't hurt to do early, but I think it's still going to do it itself.

@dblohm7?

cmd/tailscale/cli/update.go
bradfitz: Why? We don't want a warning dialog popping up. And if msiexec won't have a warning dialog pop up anyway, then ... why? For people double-clicking it by hand in the MSICache dir?

cmd/tailscale/cli/update.go
bradfitz: Dunno?

It didn't for me. I don't think I have any of the relevant things turned off. Doesn't look like it.

@dblohm7?

cmd/tailscale/cli/update.go
bradfitz: reworded

cmd/tailscale/cli/update.go
bradfitz: @andrew-d and I chatted about this and we think we should copy that wireguard-windows code.

cmd/tailscale/cli/update.go
dblohm7: Windows Defender is kind of a black box wrt this, hard to say. This is a very common thing for updaters to do, though.

cmd/tailscale/cli/update.go
dblohm7: I think we should only do that for arbitrary executables. At this point we know it's ours.

cmd/tailscale/cli/update.go
andrew-d: (for the record: I'm fairly certain that `msiexec` doesn't do full revocation checking, so to be safe and prevent someone from being able to force a downgrade to an older/vulnerable and revoked version, we run `WinVerifyTrustEx` with the `WTD_REVOKE_WHOLECHAIN` flag, as above, to protect against that)

DentonGentry (commented):
This LGTM to me. This is an area where the rest of y'all know way more than I do, I'll leave it to you.

bradfitz:
Okay, waiting on @dblohm7 to do another pass over it. I'll resume work on Debian support.

cmd/tailscale/cli/update.go
```diff
@@ -24,7 +37,7 @@ import (
 var updateCmd = &ffcli.Command{
 	Name:       "update",
 	ShortUsage: "update",
-	ShortHelp:  "Update Tailscale to the latest/different version",
+	ShortHelp:  "[ALPHA] Update Tailscale to the latest/different version",
```
mayakacz, line 40: Have we used "Alpha" or other warnings in other parts of the CLI? Not afaik?
If not, then not sure we should include, unless we're going to add throughout the CLI.

mayakacz (approved):

cmd/tailscale/cli/update.go
```diff
@@ -200,6 +277,211 @@ func (up *updater) updateWindows() error {
...
+		if err == nil {
+			break
+		}
+		uninstallVersion := version.Short
+		if v := os.Getenv("TS_DEBUG_UNINSTALL_VERSION"); v != "" {
+			uninstallVersion = v
+		}
+		// Assume it's a downgrade, which msiexec won't permit. Uninstall our current version first.
+		log.Printf("Uninstalling current version %q for downgrade...", uninstallVersion)
+		cmd = exec.Command("msiexec.exe", "/x", msiUUIDForVersion(uninstallVersion), "/promptrestart", "/qn")
```
andrew-d, line 339: Do we want `/promptrestart` here (and above)? Will the user be able to click yes on the prompt if this is being run?

andrew-d (approved):

cmd/tailscale/cli/update.go
bradfitz: Well, the one above we can keep, assuming it ends up doing a GUI thing. But this uninstall one is redundant and we should remove. If a reboot is required, the subsequent install step seconds later will prompt. Better than two pop-ups.

cmd/tailscale/cli/update.go
bradfitz: Yes, we have. I made it match the others:

```
bradfitz@tsdev:~/src/tailscale.com$ git grep ALPHA
cmd/tailscale/cli/login.go:     ShortUsage: "[ALPHA] login [flags]",
cmd/tailscale/cli/serve.go:             ShortHelp: "[ALPHA] Serve from your Tailscale node",
cmd/tailscale/cli/serve.go:*** ALPHA; all of this is subject to change ***
cmd/tailscale/cli/switch.go:  [ALPHA] switch <name>
cmd/tailscale/cli/switch.go:  [ALPHA] switch --list
cmd/tailscale/cli/update.go:    ShortHelp:  "[ALPHA] Update Tailscale to the latest/different version",
```

cmd/tailscale/cli/update.go
```diff
@@ -200,6 +277,211 @@ func (up *updater) updateWindows() error {
...
 		return errors.New("must be run as Administrator")
 	}
-	// TODO(bradfitz): require elevated mode
-	return errors.New("TODO: download + msiexec /i /quiet " + url)
+	if err := up.confirm(ver); err != nil {
+		return err
+	}
+
+	targetDir := filepath.Join(os.Getenv("ProgramData"), "Tailscale", "MSICache")
+	if err := os.MkdirAll(targetDir, 0700); err != nil {
```
dblohm7, line 285: We need to be careful that the ACL is set correctly on the Tailscale directory. See `paths.MkStateDir`.

cmd/tailscale/cli/update.go
```diff
@@ -200,6 +277,211 @@ func (up *updater) updateWindows() error {
...
+	pw.done += int64(len(p))
+	if time.Since(pw.lastPrint) > 2*time.Second {
+		pw.print()
+	}
+	return len(p), nil
+}
+
+func (pw *progressWriter) print() {
+	pw.lastPrint = time.Now()
+	log.Printf("Downloaded %v/%v (%.1f%%)", pw.done, pw.total, float64(pw.done)/float64(pw.total)*100)
```
dblohm7, line 486: Wait, no emoji? ;-)

(Kidding)

cmd/tailscale/cli/update.go
bradfitz: Done.

cmd/tailscale/cli/update.go
bradfitz: I'm sure cmd.exe would be just great at that.

dblohm7 (approved):
SHIP IT

PR #7012, opened 2023-01-19
cmd/tailscale/cli: un-alpha login+switch in ShortUsage docs

bradfitz:


bradfitz:
cc @mayakacz @WalterHub

maisem (approved):

PR #7014, opened 2023-01-19
cmd/k8s-operator: Finalizer should requeue immediately

vincepri:
Added details in the comment, long story short is that by not reconciling immediately there could be a race condition between the update, deletion, and next reconciliation; especially if there are long(ish) running operations.

Signed-off-by: Vince Prignano <vince@prigna.com>

### DIFF
--- a/cmd/k8s-operator/operator.go
+++ b/cmd/k8s-operator/operator.go
@@ -26,10 +26,12 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	kzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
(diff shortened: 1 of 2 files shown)


cmd/k8s-operator/operator.go
```diff
@@ -268,25 +270,45 @@ func childResourceLabels(parent *corev1.Service) map[string]string {
...
+		// because once the finalizer is in place this block gets skipped. So,
+		// this is a nice place to tell the operator that the high level,
+		// multi-reconcile operation is underway.
+		logger.Infof("exposing service over tailscale")
+		controllerutil.AddFinalizer(svc, FinalizerName)
+		if err := a.Update(ctx, svc); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to add finalizer: %w", err)
+		}
+
+		// If we just added the finalizer, we need to requeue the service to
```
danderson, line 301: Incomplete sentence?

cmd/k8s-operator/operator.go
```diff
@@ -268,25 +270,45 @@ func childResourceLabels(parent *corev1.Service) map[string]string {
...
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	// Handle deletion.
 	if !svc.DeletionTimestamp.IsZero() || !a.shouldExpose(svc) {
 		logger.Debugf("service is being deleted or should not be exposed, cleaning up")
 		return reconcile.Result{}, a.maybeCleanup(ctx, logger, svc)
 	}
```
vincepri, line 309: While looking through the code, noticed in `maybeCleanup` we're checking if the finalizer doesn't exist, to skip cleanup. Is that because the cleanup logic is not idempotent? Or could we potentially remove the check and cleanup regardless of the finalizer, and always removing it at the end? The latter is usually the pattern I've seen the most, but no strong opinions

danderson (commented):
Thanks for the PR! I don't think I understand the sequence of reconcile executions that cause the race you describe. Is the case you're worried about something like reconcile gets called only once, but DeletionTimestamp is already non-zero? Or something else?

Unless maybeProvision has executed at least once, there is nothing for us to clean up, so I'm trying to figure out why it's necessary to move the finalizer install sooner, and why the explicit requeue is needed. Help? :)

cmd/k8s-operator/operator.go
vincepri: Yeah, leftover before I moved the details on the outer block, will fix

vincepri: ok to close? :)

Issue #7015, opened 2023-01-19
tailcfg: use capver for peerapi, localapi


bradfitz:
Per discussion with @mihaip, for http://go/2mac work (macsys GUI connecting to oss `tailscaled`), and for peerapi, we want to use tailcfg.CapabilityVersion more. (similar to how we re-used it for RegisterRequest.Version and SetDNSRequest.Version, etc)

For LocalAPI (GUI/CLI &lt;=&gt; tailscaled), we'll add it to HTTP headers in/out, and later plumb via contexts in localapi HTTP handlers, so new tailscaled can fake old results to old clients if/when needed. For now just informational.

For peerapi it'll let us avoid things like the dumb `tailcfg.PeerAPIDNS` hack, which we can then start phasing out.

PR #7016, opened 2023-01-19
all: start groundwork for using capver for localapi & peerapi

bradfitz:
Updates #7015


client/tailscale/localclient.go
```diff
@@ -114,6 +114,7 @@ func (lc *LocalClient) defaultDialer(ctx context.Context, network, addr string)
 //
 // DoLocalRequest may mutate the request to add Authorization headers.
 func (lc *LocalClient) DoLocalRequest(req *http.Request) (*http.Response, error) {
+	req.Header.Set("Tailscale-Cap", fmt.Sprint(int(tailcfg.CurrentCapabilityVersion)))
```
maisem, line 117: would it be worth storing the string version to avoid stringification at every call?

tailcfg/tailcfg.go
```diff
@@ -199,6 +200,7 @@ type Node struct {
 	DERP         string         `json:",omitempty"` // DERP-in-IP:port ("127.3.3.40:N") endpoint
 	Hostinfo     HostinfoView
 	Created      time.Time
+	Cap          CapabilityVersion `json:",omitempty"` // if non-zero, the node's capability version
```
maisem, line 203: nit: CapVer?

maisem (approved):

client/tailscale/localclient.go
bradfitz: drop in the bucket I think. also IIRC all two byte int->string don't allocate

tailcfg/tailcfg.go
bradfitz: (discussed on chat)

client/tailscale/localclient.go
mihaip: Out of curiosity, is `fmt.Sprint` preferable to `strconv.Itoa`?

mihaip (approved):

tailcfg/tailcfg.go
```diff
@@ -90,7 +90,8 @@ type CapabilityVersion int
 //   - 51: 2022-11-30: Client understands CapabilityTailnetLockAlpha
 //   - 52: 2023-01-05: client can handle c2n POST /logtail/flush
 //   - 53: 2023-01-18: client respects explicit Node.Expired + auto-sets based on Node.KeyExpiry
-const CurrentCapabilityVersion CapabilityVersion = 53
+//   - 54: 2023-01-19: Node.Cap added, PeersChangedPatch.Cap, uses Node.Cap for ExitDNS before Hostinfo.Services fallback
+const CurrentCapabilityVersion CapabilityVersion = 54
```
mihaip, line 94: May want to also change the comment for `CapabilityVersion` to indicate that it's used for more than just client <-> control negotiation. I had it as:

```
-// server (control plane), bump this number and document what's new.
+// server (control plane), peers (over PeerAPI) or frontend (over LocalAPI),
+// bump this number and document what's new.
```

client/tailscale/localclient.go
bradfitz: strconv.Itoa would be a bit faster, actually. I can change.

I often use `fmt.Sprint` when I don't want to think about what the argument is.

tailcfg/tailcfg.go
bradfitz: Good idea, done.

PR #7017, opened 2023-01-19
cmd/k8s-operator: remove use of InjectClient (deprecated)

vincepri:
The dependency injection functionality has been deprecated a while back and it'll be removed in the 0.15 release of Controller Runtime. This changeset sets the Client after creating the Manager, instead of using InjectClient.

Signed-off-by: Vince Prignano <vince@prigna.com>


danderson (approved):
Thanks!

PR #7018, opened 2023-01-19
all: implement pcap streaming for datapath debugging

twitchyliquid64:
image

image

Done:
 - Hooks on the 4 main paths
 - Generating pcap streams
 - Custom dissector so wireshark can show Tailscale metadata (I've only implemented which path it traversed for now)

Future:
 - Show the outcome of ACL eval
 - Encode the target node in the pcap stream maybe?
 - When we have cgnat maybe we show the original / translated addresses as well?


ipn/ipnlocal/local.go
```diff
@@ -4700,3 +4702,23 @@ func (b *LocalBackend) ListProfiles() []ipn.LoginProfile {
...
 	return b.pm.Profiles()
 }
+
+func (b *LocalBackend) StartDebugCapture(addr string) error {
+	s, err := capture.TCPSink(addr)
+	if err != nil {
+		return err
+	}
+
+	b.debugSink = s
```
twitchyliquid64, line 4712: Probs needs locking aye

bradfitz:
This is tailscale/corp#8470

Add `Updates tailscale/corp#8470` in the commit message.

client/tailscale/localclient.go
```diff
@@ -1019,6 +1019,23 @@ func (lc *LocalClient) DebugDERPRegion(ctx context.Context, regionIDOrCode strin
 	return decodeJSON[*ipnstate.DebugDERPRegionReport](body)
 }
 
+func (lc *LocalClient) StartDebugCapture(ctx context.Context, addr string) error {
```
bradfitz, line 1022: godoc. and what is "addr" ?

I'd like to be able to attach to multiple point(s) inside the data plane, like all-in, post-filter-in, out, etc.

client/tailscale/localclient.go
```diff
@@ -1019,6 +1019,23 @@ func (lc *LocalClient) DebugDERPRegion(ctx context.Context, regionIDOrCode strin
...
+	v := url.Values{"addr": {addr}}
+	body, err := lc.send(ctx, "POST", "/localapi/v0/start-debug-capture?"+v.Encode(), 200, nil)
+	if err != nil {
+		return fmt.Errorf("error %w: %s", err, body)
+	}
+	return nil
+}
+
+func (lc *LocalClient) StopDebugCapture(ctx context.Context) error {
```
bradfitz, line 1031: stop what? there can only be 1 capture globally?

then why doesn't it end when the HTTP request above ends? the one above can stream a pcap?

client/tailscale/localclient.go
twitchyliquid64: Why attach separately? I've always just fed wireshark the firehose and used its filtering.

In this case, our custom dissector exposes a field `tsdebug.path` which we could use to zoom in on only one of the points.

bradfitz: how does wireshark know what we rejected?

client/tailscale/localclient.go
twitchyliquid64: Currently the way it works is you install a hook, which is an atomic func callback, and that gets wired to all sockets spawned from a TCP listener. I'm not crashhot on trying to wire it from the localAPI socket but I guess we could try that.

I lean towards TCP with a default localhost address but the ability to open it up via the command line, so then you can capture on one machine and monitor on another.

wgengine/capture/ts-dissector.lua
```diff
@@ -0,0 +1,27 @@
...
+function tsdebug_ll.dissector(buffer, pinfo, tree)
+    pinfo.cols.protocol = tsdebug_ll.name
+    packet_length = buffer:len()
+    local offset = 0
+    local subtree = tree:add(tsdebug_ll, buffer(), "Tailscale packet")
+
+    -- -- Get path UINT16
+    local path_id = buffer:range(offset, 2):le_uint()
+    if     path_id == 0 then subtree:add(PATH, "FromLocal")
+    elseif path_id == 1 then subtree:add(PATH, "FromPeer")
```
bradfitz, line 14: can there be a separate path for "FromPeerDropped" vs "FromPeerAccepted" or something?

client/tailscale/localclient.go
twitchyliquid64: Wireshark knows whatever we tell it. For something like filter-dropped packets, we could call the callback after the filter and add a new metadata field `dropped` or something like that.

Currently the cb is wired before ACL eval, cuz WIP.

net/tstun/wrap.go
```diff
@@ -69,6 +70,8 @@ var parsedPacketPool = sync.Pool{New: func() any { return new(packet.Parsed) }}
 // It must not hold onto the packet struct, as its backing storage will be reused.
 type FilterFunc func(*packet.Parsed, *Wrapper) filter.Response
 
+type CaptureFunc func(capture.Path, time.Time, []byte)
```
bradfitz, line 73: docs, discuss ownership/lifetime of that []byte

wgengine/capture/capture.go
```diff
@@ -0,0 +1,138 @@
+// Copyright (c) 2023 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package capture
```
bradfitz, line 5: package doc

wgengine/capture/capture.go
```diff
@@ -0,0 +1,138 @@
...
+
+// Path describes where in the data path the packet was captured.
+type Path uint8
+
+// Valid Path values.
+const (
+	FromLocal Path = iota
+	FromPeer
+	SynthesizedToLocal
+	SynthesizedToPeer
```
bradfitz, line 45: describe these

wgengine/capture/capture.go
```diff
@@ -0,0 +1,138 @@
...
+	binary.Write(w, binary.LittleEndian, uint32(length)) // length present
+	binary.Write(w, binary.LittleEndian, uint32(length)) // total length
+}
+
+// Path describes where in the data path the packet was captured.
+type Path uint8
+
+// Valid Path values.
+const (
+	FromLocal Path = iota
```
bradfitz, line 42: I wouldn't use iota here. Only use iota when the value doesn't leave the program to network/disk.

But we need to keep it in sync with lua, so spell out numbers.

wgengine/capture/capture.go
```diff
@@ -0,0 +1,138 @@
...
+	SynthesizedToPeer
+)
+
+func TCPSink(addr string) (*Sink, error) {
+	a, err := net.ResolveTCPAddr("tcp", addr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid address: %w", err)
+	}
+
+	l, err := net.ListenTCP("tcp", a)
```
bradfitz, line 54: why ListenTCP + ResolveTCPAddr as opposed to just "net.Listen" ?

but now I get why there's an addr... you're doing a separate listener.

note that this won't work well on macOS in the network extension sandbox.

That's why I was thinking this would instead streap the pcap out the HTTP long poll call

wgengine/capture/ts-dissector.lua
twitchyliquid64: I'd rather add additional metadata (maybe an enum `acl_verdict` with pre-intercept, accept, drop) than overload `path`. We have total flexibility here, Wireshark is really nice like that.

bradfitz: Sure, that works.

wgengine/capture/capture.go
twitchyliquid64: Hmmm, do we have anywhere where we plumb streaming back through localAPI like that already?

bradfitz: multiple things: watch-ipn-bus and the new recent logtap

nyghtowl:
Issue: #8470

twitchyliquid64:
Main changes since last revision: pcaps are streamed back over localAPI to stdout, in lieu of all that TCP listener nonsense

twitchyliquid64:
Should be good for another round @bradfitz. In particular I'm a little iffy on the lifetime of the (localAPI HTTP) request goroutine, blocking on a channel till close feels a little weird.

client/tailscale/localclient.go
```diff
@@ -1028,6 +1028,22 @@ func (lc *LocalClient) DebugSetExpireIn(ctx context.Context, d time.Duration) er
...
+		return nil, err
+	}
+	res, err := lc.doLocalRequestNiceError(req)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode != 200 {
+		return nil, errors.New(res.Status)
+	}
+	return res.Body, nil
```
bradfitz, line 1044: this is a ReadCloser so you should probably return one and document that the ctx is only used for the initial return and not the lifetime of the reader content.

cmd/tailscale/cli/debug.go
```diff
@@ -190,6 +190,11 @@ var debugCmd = &ffcli.Command{
 			Exec:      runDebugDERP,
 			ShortHelp: "test a DERP configuration",
 		},
+		{
+			Name:      "capture",
+			Exec:      runCapture,
+			ShortHelp: "streams pcaps to stdout for debugging",
```
bradfitz, line 196: "to stdout" ?  make that a flag, with "-" meaning stdout?  if you do a filename, you need explicit "-" support for macOS which can't write to disk when sandboxed.

cmd/tailscale/cli/debug.go
```diff
@@ -734,3 +739,18 @@ func runSetExpire(ctx context.Context, args []string) error {
...
 }
+
+func runCapture(ctx context.Context, args []string) error {
+	stream, err := localClient.StreamDebugCapture(ctx)
+	if err != nil {
+		return err
+	}
+	fmt.Fprintln(os.Stderr, "To examine in wireshark, run:")
+	fmt.Fprintf(os.Stderr, "\t%s | wireshark -X lua_script:<path to ts-dissector.lua> -k -i -\n", strings.Join(os.Args, " "))
+	fmt.Fprintln(os.Stderr, "Dont forget to download the Tailscale debug dissector! You can get it from the Tailscale source tree in wgengine/capture.")
```
bradfitz, line 750: why not go:embed it and let this tool write it out as needed? It could even start wireshark.

ipn/ipnlocal/local.go
```diff
@@ -4791,3 +4798,30 @@ func (b *LocalBackend) ResetAuth() error {
...
+	var (
+		s   *capture.Sink
+		err error
+	)
+
+	b.mu.Lock()
+	if b.debugSink == nil {
+		if s, err = capture.New(); err != nil {
+			b.mu.Unlock()
+			return err
```
bradfitz, line 4814: have this do the 200 write so the error path can return a nice 500 + error text?

also do a http Flush after the 200 so the HTTP response headers get sent immediately even if no packets are coming?

wgengine/capture/capture.go
```diff
@@ -0,0 +1,150 @@
...
+	// SynthesizedToLocal indicates the packet was generated from within tailscaled,
+	// and is being routed to the local machine's network stack.
+	SynthesizedToLocal Path = 2
+	// SynthesizedToPeer indicates the packet was generated from within tailscaled,
+	// and is being routed to a remote Wireguard peer.
+	SynthesizedToPeer Path = 3
+)
+
+// New creates a new capture sink.
+func New() (*Sink, error) {
```
bradfitz, line 55: when do you plan to make this return an error?

wgengine/capture/capture.go
```diff
@@ -0,0 +1,150 @@
...
+		close: make(chan struct{}),
+	}
+	return s, nil
+}
+
+// Type Sink handles callbacks with packets to be logged,
+// formatting them into a pcap stream which is mirrored to
+// all registered outputs.
+type Sink struct {
+	close chan struct{}
```
bradfitz, line 66: this is a clone-once channel?

rather than documenting, more clear to use a context.WithCancel() and store the context.CancelFunc and context.Context in here. then callers know it's safe to dup call the CancelFunc and it's a no-op.

wgengine/capture/capture.go
```diff
@@ -0,0 +1,150 @@
...
+	mu      sync.Mutex
+	outputs []io.Writer
+}
+
+// RegisterOutput connects an output to this sink, which
+// will be written to with a pcap stream as packets are logged.
+//
+// If w implements io.Closer, it will be closed upon error
+// or when the sink is closed. If w implements http.Flusher,
+// it will be flushed after each packet.
```
bradfitz, line 77: after each packet is a little aggressive during heavy flows. maybe each packet starts a 200ms flush timer (if a timer doesn't already exist)?

wgengine/capture/capture.go
```diff
@@ -0,0 +1,150 @@
...
+func (s *Sink) RegisterOutput(w io.Writer) {
```
bradfitz, line 78: normally a RegisterFoo returns an unregister func.  how does the caller unregister this?

wgengine/capture/capture.go
```diff
@@ -0,0 +1,150 @@
...
+//
+// This function does not take ownership of the provided data slice.
+func (s *Sink) LogPacket(path Path, when time.Time, data []byte) {
+	select {
+	case <-s.close:
+		return
+	default:
+	}
+
+	b := bytes.NewBuffer(make([]byte, 0, 16+2+len(data)))
```
bradfitz, line 123: this seems like it'll be pretty garbage-y (lot of GC activity). sync.Pool these buffers?

wgengine/wgengine.go
```diff
@@ -171,4 +178,6 @@ type Engine interface {
 	// WhoIsIPPort looks up an IP:port in the temporary registrations,
 	// and returns a matching Tailscale IP, if it exists.
 	WhoIsIPPort(netip.AddrPort) (netip.Addr, bool)
+
+	InstallCaptureHook(CaptureCallback)
```
bradfitz, line 182: godoc.

and how does caller unregister?

cmd/tailscale/cli/debug.go
twitchyliquid64: Can't someone just pipe it to a file if they want a stored capture? That way we don't have to deal with invariants like MacOS.

cmd/tailscale/cli/debug.go
twitchyliquid64: Ooooo I like this. Can a sandboxed MacOS binary fork out like that, and can the exec'd program access tempfiles of its parent?

wgengine/capture/capture.go
twitchyliquid64: Callers don't atm: if/when the connection is closed, the Write() starts to error and it is removed as an output. Is that bad?

wgengine/capture/capture.go
twitchyliquid64: Done, but PTAL because Ive never used Pool before so not sure if I'm doing it correctly.

client/tailscale/localclient.go
```diff
@@ -1028,6 +1028,25 @@ func (lc *LocalClient) DebugSetExpireIn(ctx context.Context, d time.Duration) er
...
+	req, err := http.NewRequestWithContext(ctx, "POST", "http://"+apitype.LocalAPIHost+"/localapi/v0/debug-capture", nil)
+	if err != nil {
+		return nil, err
+	}
+	res, err := lc.doLocalRequestNiceError(req)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode != 200 {
+		return nil, errors.New(res.Status)
```
bradfitz, line 1045: missing res.Body.Close

but isn't there an unexported LocalClient method that does this already? or does doLocalRequestNiceError do that?

cmd/tailscale/cli/debug.go
```diff
@@ -190,6 +192,16 @@ var debugCmd = &ffcli.Command{
 			Exec:      runDebugDERP,
 			ShortHelp: "test a DERP configuration",
 		},
+		{
+			Name:      "capture",
+			Exec:      runCapture,
+			ShortHelp: "streams pcaps for debugging",
+			FlagSet: (func() *flag.FlagSet {
+				fs := newFlagSet("capture")
+				fs.StringVar(&captureArgs.outFile, "o", "<wireshark>", "where to stream the pcap, or <wireshark> to start wireshark")
```
bradfitz, line 201: that's a kinda hard sentinel value to document due to the different quoting rules between operating systems & shells. and it'll render in flag help doc as if it's a parameter. 

also, what happened to "-" for stdout?

twitchyliquid64: "-" for stdout works! I guess we should fill in `LongHelp` with all the details

twitchyliquid64: As for triggering the wireshark flow, I'll move that to a positional parameter

bradfitz: I mean you didn't document "-" for stdout.

wgengine/capture/capture.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// Type Sink handles callbacks with packets to be logged,
+// formatting them into a pcap stream which is mirrored to
+// all registered outputs.
+type Sink struct {
+	ctx       context.Context
+	ctxCancel context.CancelFunc
+
+	bufferPool sync.Pool
```
bradfitz, line 83: sync.Pools should be global variables, somewhat counter-intuitively. (there's a huge history there you can go read about but it's exhausting)

ipn/ipnlocal/local.go
```diff
@@ -4791,3 +4798,23 @@ func (b *LocalBackend) ResetAuth() error {
 	}
 	return b.resetForProfileChangeLockedOnEntry()
 }
+
+// StreamDebugCapture writes a pcap stream of packets traversing
+// tailscaled to the provided response writer.
+func (b *LocalBackend) StreamDebugCapture(w http.ResponseWriter) error {
```
bradfitz, line 4804: having a ResponseWriter deep in here feels a bit strange. Could this just be an io.Writer?

ipn/ipnlocal/local.go
```diff
@@ -4791,3 +4798,23 @@ func (b *LocalBackend) ResetAuth() error {
...
+	var s *capture.Sink
+
+	b.mu.Lock()
+	if b.debugSink == nil {
+		s = capture.New()
```
bradfitz, line 4809: why doesn't capture.New take the writer? what's up with the RegisterOutput step being separate?

ipn/ipnlocal/local.go
```diff
@@ -4791,3 +4798,23 @@ func (b *LocalBackend) ResetAuth() error {
...
+		b.debugSink = s
+		b.e.InstallCaptureHook(s.LogPacket)
+	} else {
+		s = b.debugSink
+	}
+	b.mu.Unlock()
+
+	s.RegisterOutput(w)
+	s.Wait()
```
bradfitz, line 4818: you should take a context instead, then this becomes a select on both the context and the LocalBackend's shutdownCtx

bradfitz: And after the context(s) are done, I'd prefer to see an explicit unregister call to stop the sink.

And why can't there be multiple going on at once?

ipn/ipnlocal/local.go
twitchyliquid64: Because a sink can have multiple outputs. Its the sink's responsibility to fan out from the `LogPacket` call to each output.

client/tailscale/localclient.go
twitchyliquid64: Done. Poking around, `doLocalRequestNiceError` only handles 403, bad connections, and version mismatches, so we still need this.

wgengine/capture/capture.go
```diff
@@ -0,0 +1,208 @@
...
+// Type Sink handles callbacks with packets to be logged,
+// formatting them into a pcap stream which is mirrored to
+// all registered outputs.
+type Sink struct {
+	ctx       context.Context
+	ctxCancel context.CancelFunc
+
+	mu         sync.Mutex
+	outputs    []io.Writer
+	flushTimer *time.Timer
```
bradfitz, line 86: // or nil if none running

wgengine/capture/capture.go
```diff
@@ -0,0 +1,208 @@
...
+
+// RegisterOutput connects an output to this sink, which
+// will be written to with a pcap stream as packets are logged.
+// A function is returned which unregisters the output when
+// called.
+//
+// If w implements io.Closer, it will be closed upon error
+// or when the sink is closed. If w implements http.Flusher,
+// it will be flushed periodically.
+func (s *Sink) RegisterOutput(w io.Writer) func() {
```
bradfitz, line 97: name the result `(unregister func()) {

wgengine/capture/capture.go
```diff
@@ -0,0 +1,208 @@
...
+	s.mu.Lock()
+	s.outputs = append(s.outputs, w)
+	s.mu.Unlock()
+
+	return func() {
+		s.mu.Lock()
+		defer s.mu.Unlock()
+		for i := range s.outputs {
+			if s.outputs[i] == w {
+				s.outputs = append(s.outputs[:i], s.outputs[i+1:]...)
```
bradfitz, line 114: as long as you're requiring that the io.Writer type be `comparable` (you use == here), why not just use a `map` for a set? Then you don't need to worry about retaining the old io.Writers here, which you are... you're not zeroing the value between the len and cap when slide things down and thus you're still retaining things GC-wise

bradfitz: Or better: use `set.HandleSet` and then you don't need to worry about documenting that the caller supply a comparable io.Writer.

wgengine/capture/capture.go
```diff
@@ -0,0 +1,208 @@
...
+	for _, o := range s.outputs {
+		if o, ok := o.(io.Closer); ok {
+			o.Close()
+		}
+	}
+	s.outputs = nil
+	return nil
+}
+
+// WaitCh returns a channel which blocks till
```
bradfitz, line 142: until

wgengine/capture/capture.go
```diff
@@ -0,0 +1,208 @@
...
+	if s.flushTimer == nil {
+		t := time.NewTimer(flushPeriod)
+		s.flushTimer = t
+		go func() {
+			// We do not check ctx here as Close() will
+			// stop the timer and unblock us.
+			_, ok := <-t.C
+			if !ok {
+				return
+			}
```
bradfitz, line 196: seems like you actually want a time.AfterFunc?

wgengine/wgengine.go
```diff
@@ -42,6 +43,12 @@ type NetInfoCallback func(*tailcfg.NetInfo)
 // into network map updates.
 type NetworkMapCallback func(*netmap.NetworkMap)
 
+// CaptureCallback is the type uses to record packets when
```
bradfitz, line 46: used

ipn/ipnlocal/local.go
```diff
@@ -516,6 +518,11 @@ func (b *LocalBackend) Shutdown() {
 		b.sshServer = nil
 	}
 	b.closePeerAPIListenersLocked()
+	if b.debugSink != nil {
+		b.e.InstallCaptureHook(nil)
```
bradfitz, line 522: are you missing a call to InstallCaptureHook(nil) when the Sink's number of Outputs transitions from 1 to 0? I'm not seeing it.

twitchyliquid64: The sink and the hook stays installed until the `LocalBackend` shuts down (the logic to close the sink is in Close). Are you saying you want to attempt teardown when there are no outputs?

Seems doable but might be a lil messy trying to manage concurrency with other `StreamDebugCapture` calls

twitchyliquid64: I've gone ahead and made it shut down the sink if there are no more outputs.

bradfitz (approved):

PR #7019, opened 2023-01-20
Update ipn/store/awsstore/store_aws.go to wrap boolean values in aws.Bool

thisisaaronland:
`ssm.GetParameterInput` now requires that `WithDecryption` be passed by reference:

https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ssm#GetParameterInput

### DIFF
--- a/ipn/store/awsstore/store_aws.go
+++ b/ipn/store/awsstore/store_aws.go
@@ -105,7 +105,7 @@ func (s *awsStore) LoadState() error {
 		context.TODO(),
 		&ssm.GetParameterInput{
 			Name:           aws.String(s.ParameterName()),
-			WithDecryption: true,
+			WithDecryption: aws.Bool(true),
 		},
 	)
 
@@ -167,7 +167,7 @@ func (s *awsStore) persistState() error {
 		&ssm.PutParameterInput{
 			Name:      aws.String(s.ParameterName()),
 			Value:     aws.String(string(bs)),
-			Overwrite: true,
+			Overwrite: aws.Bool(true),
 			Tier:      ssmTypes.ParameterTierStandard,
 			Type:      ssmTypes.ParameterTypeSecureString,
 		},


bradfitz:
I assume this doesn't compile without go.mod updates?

bradfitz:
We'll handle this next time we bump deps. Unless there's some reason to bump this sooner?

thisisaaronland:
No rush. I was able to work around it with a fork and `replace` statements in go.mod files.

Thanks,

bradfitz:
> I assume this doesn't compile without go.mod updates?

This assumption was correct. This broke all our CI builds, which didn't leave a comment, but you can see at https://github.com/tailscale/tailscale/pull/7019/checks

> No rush. I was able to work around it with a fork and replace statements in go.mod files.

Can you back up and explain the problem?  I'm now assuming you were using tailscale as a library in your own project and your project's go.mod was using a newer version of the AWS sdk?

Pretty lame of them to break API like that, if so.

If so, that seems worthy of us updating to get past that breakage.

But you didn't lead with that problem statement and instead sent a patch which didn't compile (due to lack of the go.mod change) so I closed it for lack of context & lack of compiling.

bradfitz:
Now that I understand the problem I've sent #7025 to reapply this change with go.mod and depaware updates to pass CI.

thisisaaronland:
First, sorry about not updating the `go.mod` stuff. Everything built correctly on my machine but presumably that's because I had the relevant packages... somewhere in the Go-package-cache-kingdom. But, yes, it stands to reason that the tests would break without that update. My bad.

I am using the Tailscale `tsnet` stuff here:

* https://github.com/aaronland/go-http-server-tsnet

Which is then imported here:

* https://github.com/whosonfirst/go-whosonfirst-browser

Specifically: 

* https://github.com/whosonfirst/go-whosonfirst-browser#tailscale

Which is a bit confusing because everything compiled fine when the "v6" release was published last week:

* https://github.com/whosonfirst/go-whosonfirst-browser/blob/main/go.mod#L75

But then broke yesterday when compiling a new branch that, mysteriously, requires the same version of `aws-sdk-go-v2/service/ssm` (v1.33.1)

* https://github.com/whosonfirst/go-whosonfirst-browser/blob/edit-geom/go.mod#L93

So, I confess I'm not really sure _why_ I suddenly started getting this error yesterday and not sooner:

```
GOOS=linux go build -o main cmd/browser/main.go
# tailscale.com/ipn/store/awsstore
../../go/pkg/mod/tailscale.com@v1.34.2/ipn/store/awsstore/store_aws.go:108:20: cannot use true (untyped bool constant) as *bool value in struct literal
../../go/pkg/mod/tailscale.com@v1.34.2/ipn/store/awsstore/store_aws.go:170:15: cannot use true (untyped bool constant) as *bool value in struct literal
make: *** [lambda] Error 2
```

bradfitz:
https://research.swtch.com/vgo-mvs has details.

It broke for Tailscale because within the `tailscale.com` module (its `go.mod` universe) the minimum was one thing and worked for the code that the `tailscale.com` module was using. But in your other modules, the minimum for the AWS SDK across all modules was now higher, so the tailscale.com code built with that higher version and no longer worked.

This is AWS's fault for violating the rule of no API changes for a given import path in a non-v0 major version ("semantic import versioning" ... https://research.swtch.com/vgo-module)

thisisaaronland:
Right, but what's confusing to me is that when `go-whosonfirst-browser` was bumped to "v6" last week it replaced v1.27.6 of the `ssm` package (with v1.33.1 ) and required tailscale 1.34.2 which, in turn, required v1.17.1 of the `ssm` package. So I am not clear why the problem didn't manifest itself then.

https://github.com/whosonfirst/go-whosonfirst-browser/commit/ca8986cb60ec299396d74039f6747c3ece9c049d#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R75

Issue #7020, opened 2023-01-20
client.Keys() incorrectly un-marshals API response


hbowron:
### What is the issue?

Per the [Tailscale API documentation](https://github.com/tailscale/tailscale/blob/main/api.md#tailnet-keys-get) (and some validation curls) `GET /api/v2/tailnet/:tailnet/keys ` return this format:
```
{"keys": [
	{"id": "kYKVU14CNTRL"},
	{"id": "k68VdZ3CNTRL"},
	{"id": "kJ9nq43CNTRL"},
	{"id": "kkThgj1CNTRL"}
]}
```

However the `client.Keys(ctx context.Context)` function is attempting to unmarshal the response into this incorrect struct:
https://github.com/tailscale/tailscale/blob/6edf357b96b28ee1be659a70232c0135b2ffedfd/client/tailscale/keys.go#L58-L60

### Steps to reproduce

This code:
```go
func main() {
	tailscale.I_Acknowledge_This_API_Is_Unstable = true

	client := tailscale.NewClient(os.Getenv("TAILSCALE_TAILNET"), tailscale.APIKey(os.Getenv("TAILSCALE_API_KEY")))
	keys, err := client.Keys(context.Background())
	if err != nil {
		log.Fatalf("failed to get keys: %v", err)
	}
}
```

returns the following error:
```
2023/01/20 04:51:18 failed to get keys: json: cannot unmarshal object into Go value of type []struct { ID string "json:\"id\"" }
exit status 1
```

### Are there any recent changes that introduced the issue?

This logic is unchanged from the original introduction of this function https://github.com/tailscale/tailscale/pull/6715, possible that the API return changed since then.

### Bug report

BUG-a9573e0e4ef0f542794101a105bce7fa0e00ef67fe9b11168448597cd391734d-20230120095521Z-7e36622137432ae1

PR #7021, opened 2023-01-20
client/tailscale/keys: fix client.Keys unmarshalling

hbowron:
Updates https://github.com/tailscale/tailscale/issues/7020

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203810144085942

### DIFF
--- a/client/tailscale/keys.go
+++ b/client/tailscale/keys.go
@@ -55,14 +55,14 @@ func (c *Client) Keys(ctx context.Context) ([]string, error) {
 		return nil, handleErrorResponse(b, resp)
 	}
 
-	var keys []struct {
-		ID string `json:"id"`
+	var keys struct {
+		Keys []*Key `json:"keys"`
 	}
 	if err := json.Unmarshal(b, &keys); err != nil {
 		return nil, err
 	}
-	ret := make([]string, 0, len(keys))
-	for _, k := range keys {
+	ret := make([]string, 0, len(keys.Keys))
+	for _, k := range keys.Keys {
 		ret = append(ret, k.ID)
 	}
 	return ret, nil


client/tailscale/keys.go
```diff
@@ -13,6 +13,10 @@ import (
 	"time"
 )
 
+type GetKeysResponse struct {
```
knyar, line 16: Could you please make this unexported (`getKeysResponse`) and move it closer to the `Keys` function where it's used (perhaps even inline like the `var keys` was before)?

knyar (commented):
Thank you for investigating and fixing this! Looks good, I just have one minor comment.

client/tailscale/keys.go
hbowron: Happy to!  My intention was to keep it consistent with [devices.go](https://github.com/tailscale/tailscale/blob/6edf357b96b28ee1be659a70232c0135b2ffedfd/client/tailscale/devices.go#L21-L23)

hbowron: ✅

knyar:
Sorry, I've just noticed that the [DCO check is failing](https://github.com/tailscale/tailscale/pull/7021/checks?check_run_id=10775554935). Could you please add the `Signed-off-by` line to the commit message? It's also probably worth squashing commits into one. Thank you!

hbowron:
> Sorry, I've just noticed that the [DCO check is failing](https://github.com/tailscale/tailscale/pull/7021/checks?check_run_id=10775554935). Could you please add the `Signed-off-by` line to the commit message? It's also probably worth squashing commits into one. Thank you!

@knyar added!  Thanks for all the help btw

PR #7028, opened 2023-01-20
cmd/tailscale/cli: make 'tailscale update' support Debian/Ubuntu apt

bradfitz:
Updates #6995

### DIFF
--- a/cmd/tailscale/cli/update.go
+++ b/cmd/tailscale/cli/update.go
@@ -5,6 +5,7 @@
 package cli
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"crypto/sha256"
@@ -20,6 +21,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


cmd/tailscale/cli/update.go
```diff
@@ -222,13 +224,92 @@ func (up *updater) updateDebLike() error {
...
+		}
+		buf.Write(line)
+		buf.WriteByte('\n')
+	}
+	if hadCorrect || bytes.Equal(bytes.TrimSpace(was), bytes.TrimSpace(buf.Bytes())) {
+		// Unchanged or close enough.
+		return false, nil
+	}
+	if changes != 1 {
+		// No changes, but also no correct line. Add it.
```
maisem, line 309: `Add it` indicates you are trying to add it, but you aren't

cmd/tailscale/cli/update.go
```diff
@@ -222,13 +224,92 @@ func (up *updater) updateDebLike() error {
...
+			})
+		}
+		buf.Write(line)
+		buf.WriteByte('\n')
+	}
+	if hadCorrect || bytes.Equal(bytes.TrimSpace(was), bytes.TrimSpace(buf.Bytes())) {
+		// Unchanged or close enough.
+		return false, nil
+	}
+	if changes != 1 {
```
maisem, line 308: can there be more than 1 change?

bradfitz: not in the happy path

but if they did something unexpected (this error path is unexpected stuff) then maybe

cmd/tailscale/cli/update.go
bradfitz: whoops. deleted.

cmd/tailscale/cli/update.go
bradfitz: clarified

maisem (approved):

Issue #6907, opened 2023-01-07
FR: Updating clients remotely


LunchBoxC:
### What are you trying to do?

I'm using Tailscale to connect to clients remotely using remote desktop software.  

When an update for the Tailscale software exists I must manually visit these clients to update them as installing the new package over remote desktop disconnects the client and it can't be reconnected to until the setup process is fully completed and the installer exits.

### How should we solve this?

It could be solved by the following:

1) Allow clients to be updates from the admin console

OR

2) allow for automatic updating of the clients

OR

3) Change the installer so that it briefly disconnects to do the install and then reconnects all clients before the installer exits as there is no way to click on "Finish" when tailscale is now in the disconnected state and preventing the remote desktop software from making a connection.

### What is the impact of not solving this?

Waste of resources driving to clients to update their Tailscale application version.  

It also means that clients run on outdated versions for longer.

### Anything else?

It would also be great if we could pre-configure the installer and send it out with all options like "run unattended" and with the credentials already installed so the installer would make the connection after the install.  

This would make sending it to someone to install a snap, rather than needing to configure it thereafter with novice users that don't understand what to click on.

Again to aid remote connection environments.

bradfitz:
Much of this is currently underway.

SilverBut:
Great feature I'm waiting for. 

Also, it would be great if the remote updating will failsafe, so that if a server is lockdown (for example, only allowing login from other devices in the `tailnet`) and client update failed, the old client process can still working.

cc @bradfitz

DentonGentry:
The initial step in this is `tailscale update`

noncombatant:
News on this bug: Proposed solution (2) is done; see [the Beta release of auto-updates](https://tailscale.com/blog/auto-update-beta/) if you haven't already.

Proposed solution (1) is in development now. We provide a guarantee of when it will land, but it is a key part of our plan for taking auto-updates out of Beta, and we are working on it now.

PR #7029, opened 2023-01-21
version, cmd/tailscale: add version.Meta, tailscale version --json

bradfitz:
Add `tailscale version --json` JSON output mode. This will be used later for a double-opt-in (per node consent like Tailscale SSH + control config) to let admins do remote upgrades via `tailscale update` via a c2n call, which would then need to verify the cmd/tailscale found on disk for running tailscale update corresponds to the running tailscaled, refusing if anything looks amiss.

Plus JSON output modes are just nice to have, rather than parsing unstable/fragile/obscure text formats.

Updates #6995
Updates #6907

### DIFF
--- a/version/prop.go
+++ b/version/prop.go
@@ -12,7 +12,9 @@ import (
 	"strings"
 	"sync"
 
+	tailscaleroot "tailscale.com"
 	"tailscale.com/syncs"
+	"tailscale.com/tailcfg"
 )
 
 // IsMobile reports whether this is a mobile client build.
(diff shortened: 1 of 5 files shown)


version/prop.go
```diff
@@ -104,3 +106,76 @@ func initUnstable() {
...
+	// branch. That is, it reports whether the minor version is odd.
+	UnstableBranch bool `json:"unstableBranch,omitempty"`
+
+	// GitCommit, if non-empty, is the git commit of the
+	// github.com/tailscale/tailscale repository at which Tailscale was
+	// built. Its format is the one returned by `git describe --always
+	// --exclude "*" --dirty --abbrev=200`.
+	GitCommit string `json:"gitCommit,omitempty"`
+
+	// GitDirty is whether Go stamped the binary has having dirty version
```
DentonGentry, line 141: I think "has having" should be "as having"

DentonGentry (approved):

version/prop.go
bradfitz: Hah, I copy/pasted this from elsewhere that was wrong. Fixed in both places.

PR #7031, opened 2023-01-21
envknob, hostinfo, ipn/ipnlocal: add start of opt-in remote update support

bradfitz:
Updates #6907


bradfitz:
Rebased. PTAL.

ipn/ipnlocal/c2n.go
```diff
@@ -77,3 +88,92 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
...
+		res.Err = "cmd/tailscale version mismatch"
+		return
+	}
+	cmd := exec.Command(cmdTS, "update", "--yes")
+	if err := cmd.Start(); err != nil {
+		res.Err = fmt.Sprintf("failed to start cmd/tailscale update: %v", err)
+		return
+	}
+	res.Started = true
+	defer cmd.Wait()
```
andrew-d, line 155: There might be a race condition here on Windows:
1. We start the update process
2. `tailscale.exe` copies itself and kicks off the update process
3. `msiexec` stops this process during the update before the `selfCopy` exits(?)
4. This doesn't return because the process is dead

This seems fairly unlikely, but worth flagging?

ipn/ipnlocal/c2n.go
```diff
@@ -77,3 +88,92 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
...
+	return
+}
+
+func findCmdTailscale() (string, error) {
```
andrew-d, line 159: This is the safe way to do this; not relying on `$PATH` but checking the location where everything is expected to be 👍

ipn/ipnlocal/c2n.go
```diff
@@ -77,3 +88,92 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "unknown c2n path", http.StatusBadRequest)
 	}
 }
+
+func (b *LocalBackend) handleC2NUpdate(w http.ResponseWriter, r *http.Request) {
```
andrew-d, line 92: Thought for the future: we may want some sort of semaphore that prevents two concurrent updates, or if one happened in the past 5 minutes, or something.

andrew-d (approved):
LGTM! The double opt-in is great from a security perspective 👍

ipn/ipnlocal/c2n.go
bradfitz: Left a comment to that effect lest somebody add a PATH check later.

ipn/ipnlocal/c2n.go
bradfitz: Added TODO with that text

ipn/ipnlocal/c2n.go
```diff
@@ -77,3 +88,108 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
...
+	// * this process during the update before the selfCopy exits(?) This doesn't
+	//   return because the process is dead.
+	//
+	// This seems fairly unlikely, but worth checking.
+	defer cmd.Wait()
+	return
+}
+
+// findCmdTailscale looks for the cmd/tailscale that corresponds to the
+// currently running cmd/tailcsaled. It's up to the caller to verify that the
```
andrew-d, line 172:
```suggestion
// currently running cmd/tailscaled. It's up to the caller to verify that the
```

ipn/ipnlocal/c2n.go
```diff
@@ -77,3 +88,108 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
...
+		res.Err = fmt.Sprintf("failed to start cmd/tailscale update: %v", err)
+		return
+	}
+	res.Started = true
+
+	// TODO(bradfitz,andrew): There might be a race condition here on Windows:
+	// * We start the update process
+	// * tailscale.exe copies itself and kicks off the update process msiexec stops
+	// * this process during the update before the selfCopy exits(?) This doesn't
+	//   return because the process is dead.
```
andrew-d, line 164: Bad formatting

PR #7035, opened 2023-01-22
tailcfg,hostinfo: add Hostinfo.Machine and Hostinfo.GoArchVar

bradfitz:
For detecting a non-ideal binary running on the current CPU.

And for helping detect the best Synology package to update to.

Updates #6995


tailcfg/tailcfg.go
```diff
@@ -528,7 +528,9 @@ type Hostinfo struct {
 	ShareeNode      bool           `json:",omitempty"` // indicates this node exists in netmap because it's owned by a shared-to user
 	NoLogsNoSupport bool           `json:",omitempty"` // indicates that the user has opted out of sending logs and support
 	WireIngress     bool           `json:",omitempty"` // indicates that the node wants the option to receive ingress connections
+	Machine         string         `json:",omitempty"` // the host's machine type (uname -m)
 	GoArch          string         `json:",omitempty"` // the host's GOARCH value (of the running binary)
+	GoArchVar       string         `json:",omitempty"` // GOARM, GOAMD64, etc
```
DentonGentry, line 533: The previous comment says it is the running binary, I think this one is the compiled binary?

hostinfo/hostinfo.go
```diff
@@ -321,6 +326,40 @@ func inDockerDesktop() bool {
 	return false
 }
 
+// goArchVar returns the GOARM or GOAMD64 etc value that the binary was built
+// with.
+func goArchVar() string {
```
DentonGentry, line 331: If a GOARM64 or GORISCV is ever added I don't think we have a way to automatically detect it, I don't see a way to check https://go.dev/src/internal/buildcfg/cfg.go in a CI test or something.

DentonGentry (approved):

tailcfg/tailcfg.go
bradfitz: clarified all

DentonGentry (approved):

hostinfo/hostinfo.go
bradfitz: Okay, I rewrote it to be future-proof, as long as the future follows the conventions thus far. Not that we really care about anything except GOARM, which is how this all started 😅

PR #7040, opened 2023-01-23
ipn/ipnstate: add PeerStatus.KeyExpiry for tailscale status --json

### LINKED ISSUE #6712: FR: Add key expiry time to status

I would like a mechanism to find the key expiry time from the CLI

### How should we solve this?

add a line to `tailscale status --json` with the expiry time of the current key.

The information is only available in the API currently.

### Anything else?

_No response_

bradfitz:
Fixes #6712


andrew-d (approved):
Tested locally with my tailnet with a mix of expired/non-expired, and it LGTM

PR #7043, opened 2023-01-23
cmd/tailscale/cli: add debug set-expire command for testing

bradfitz:
Updates tailscale/corp#8811
Updates tailscale/corp#8613

### DIFF
--- a/client/tailscale/localclient.go
+++ b/client/tailscale/localclient.go
@@ -1020,6 +1020,15 @@ func (lc *LocalClient) DebugDERPRegion(ctx context.Context, regionIDOrCode strin
 	return decodeJSON[*ipnstate.DebugDERPRegionReport](body)
 }
 
+// DebugSetExpireIn marks the current node key to expire in d.
+//
+// This is meant primarily for debug and testing.
+func (lc *LocalClient) DebugSetExpireIn(ctx context.Context, d time.Duration) error {
+	v := url.Values{"expiry": {fmt.Sprint(time.Now().Add(d).Unix())}}
+	_, err := lc.send(ctx, "POST", "/localapi/v0/set-expiry-sooner?"+v.Encode(), 200, nil)
+	return err
+}
+
 // WatchIPNBus subscribes to the IPN notification bus. It returns a watcher
 // once the bus is connected successfully.
 //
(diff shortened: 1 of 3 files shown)


cmd/tailscale/cli/debug.go
```diff
@@ -166,6 +166,16 @@ var debugCmd = &ffcli.Command{
 				return fs
 			})(),
 		},
+		{
+			Name:      "set-expire",
+			Exec:      runSetExpire,
+			ShortHelp: "manipulate node key expiry for testing",
+			FlagSet: (func() *flag.FlagSet {
+				fs := newFlagSet("set-expire")
+				fs.DurationVar(&setExpireArgs.in, "in", 0, "if non-zero, set node key to expire this duration from now")
```
nickoneill, line 175: Can we indicate that duration is seconds and not ms/h/d?

bradfitz: those all work. It's a duration flag var, so it requires a unit suffix.

nickoneill: 👍

client/tailscale/localclient.go
```diff
@@ -1020,6 +1020,15 @@ func (lc *LocalClient) DebugDERPRegion(ctx context.Context, regionIDOrCode strin
 	return decodeJSON[*ipnstate.DebugDERPRegionReport](body)
 }
 
+// DebugSetExpireIn marks the current node key to expire in d.
+//
+// This is meant primarily for debug and testing.
+func (lc *LocalClient) DebugSetExpireIn(ctx context.Context, d time.Duration) error {
+	v := url.Values{"expiry": {fmt.Sprint(time.Now().Add(d).Unix())}}
+	_, err := lc.send(ctx, "POST", "/localapi/v0/set-expiry-sooner?"+v.Encode(), 200, nil)
```
maisem, line 1028: I don't think this endpoint does any auth, can you pls fix?

bradfitz: Discussed on chat. Also that endpoint isn't new.

bradfitz: Done, though. serveLogout required Write access as-is, so now they match.

maisem (approved):

Issue #6932, opened 2023-01-10
Improve the handling of expired nodes


andrew-d:
This is a general tracking issue for better handling on the client of nodes that have expired–e.g. providing a more useful error message when running `tailscale ping`.

PR #7044, opened 2023-01-23
ipn/ipnlocal: fire expiry timer when the current node expires

andrew-d:
The current node isn't in NetMap.Peers, so without this we would not have fired this timer on self expiry.

Updates #6932

Change-Id: Id57f96985397e372f9226802d63b42ff92c95093

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -852,6 +852,13 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 			}
 		}
 
+		// Ensure that we also fire this timer if our own node key expires.
+		if st.NetMap.SelfNode != nil {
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/local.go
```diff
@@ -852,6 +852,11 @@ func (b *LocalBackend) setClientStatus(st controlclient.Status) {
 			}
 		}
 
+		// Ensure that we also fire this timer if our own node key expires.
+		if selfExpiry := st.NetMap.SelfNode.KeyExpiry; !selfExpiry.IsZero() && selfExpiry.Before(nextExpiry) {
```
bradfitz, line 856: SelfNode can't be nil at this point? I thought there were cases it could be but forget where.

bradfitz (approved):

ipn/ipnlocal/local.go
andrew-d: At least in the tests, it appears so!

PR #7052, opened 2023-01-24
scripts: explicitly install tailscale-archive-keyring

### LINKED ISSUE #3151: Ship a keyring package for debian
### Tell us about your idea!

pkgs.tailscale.com should serve a tailscale-archive-keyring package

New style key management for 3rd party repos on Debian derivatives requires a separate keyring from the primary Debian archive keyring. We now install one of those during the initial setup process, but we have no way to roll keys over time.

### How should we solve this?

Debian's policies say that we should ship an additional `tailscale-archive-keyring` package that owns and manages /usr/share/keyrings/tailscale-archive-keyring.gpg and keeps it up to date.

In some amount of years when the current key expires, people will be unable to download updates from pkgs.tailscale.com. We also can't do basic hygiene key rotations, which would be nice.

### Anything else?

_No response_

knyar:
This will ensure that the `tailscale-archive-keyring` Debian package gets installed by the installer script.

Fixes #3151

### DIFF
--- a/scripts/installer.sh
+++ b/scripts/installer.sh
@@ -487,7 +487,7 @@ main() {
 				;;
 			esac
 			$SUDO apt-get update
-			$SUDO apt-get install -y tailscale
+			$SUDO apt-get install -y tailscale tailscale-archive-keyring
 			if [ "$APT_SYSTEMCTL_START" = "true" ]; then
 				$SUDO systemctl enable --now tailscaled
 				$SUDO systemctl start tailscaled


bradfitz:
Is that new package in the stable and unstable pkgs.tailscale.com repos yet? If not, this'll cause errors for users. Don't merge this until that package exists.

bradfitz:
(... because this goes live ~immediately once merged, not on next release)

knyar:
> Is that new package in the stable and unstable pkgs.tailscale.com repos yet? If not, this'll cause errors for users. Don't merge this until that package exists.

Yes, it has already been published to both, and I have verified that the script installs both packages now ([example](https://gist.github.com/knyar/68a4f865e529107881d2ff8c32d08589)).

bradfitz (approved):

danderson (approved):

PR #7064, opened 2023-01-25
cmd/tailscale/cli: fix false positive on cli page

### LINKED ISSUE #7059: Synology 1.36.0 claims outgoing connections when I’ve not enabled them
### What is the issue?

I updated both my real Synology and the Virtual DSM Synology to 1.36.0, and on both of them Tailscale claims “Outgoing access enabled” even though on neither of them I’ve done anything to enable it. If I ssh in, no 100.* addresses work, so I suspect the detection of this state is unreliable.

### Steps to reproduce

Install Tailscale 1.36.0 on a Synology DSM.

_No response_

### OS

Synology

### OS version

Synology DSM 7.1.1-42962 Update 2

### Tailscale version

1.36.0

### Bug report

BUG-58e464b3f7e4c645ab93703678cb504b35e27890719d5038d7186929129b905f-20230125034526Z-371a751eeff7ed68

phirework:
Fixes #7059

### DIFF
--- a/cmd/tailscale/cli/web.html
+++ b/cmd/tailscale/cli/web.html
@@ -100,7 +100,7 @@
 	</div>
 	{{ if .IsSynology }}
 		<div class="border border-gray-200 bg-orange-0 rounded-lg p-2 pl-3 pr-3 mb-8 width-full text-orange-800">
-			Outgoing access {{ if true }}enabled{{ else }}not configured{{ end }}.
+			Outgoing access {{ if .TUNMode }}enabled{{ else }}not configured{{ end }}.
 			<nobr><a href="https://tailscale.com/kb/1152/synology-outbound/"
 				class="font-medium link"
 				target="_blank"


cmd/tailscale/cli/web.html
```diff
@@ -99,13 +99,13 @@ <h3 class="text-3xl font-semibold mb-3">Log in</h3>
 		<a href="#" class="mb-4 link font-medium js-loginButton" target="_blank">Reauthenticate</a>
 	</div>
 	{{ if .IsSynology }}
+		{{ if not .TUNMode }}
```
bradfitz, line 102: and if it is? where'd that text go?

also, please adjust the git title to say something about Synology and web

cmd/tailscale/cli/web.html
```diff
@@ -99,13 +99,13 @@ <h3 class="text-3xl font-semibold mb-3">Log in</h3>
...
 		<div class="border border-gray-200 bg-orange-0 rounded-lg p-2 pl-3 pr-3 mb-8 width-full text-orange-800">
-			Outgoing access {{ if true }}enabled{{ else }}not configured{{ end }}.
-			<nobr><a href="https://tailscale.com/kb/1152/synology-outbound/"
-				class="font-medium link"
-				target="_blank"
-				rel="noopener noreferrer">Learn more &rarr;</a></nobr>
+			Outgoing traffic of services is not configured.
```
bradfitz, line 104: Can we do the wording change in its own PR?

I'm not a huge fan of this but it's fine if we're consistent, but https://tailscale.com/kb/1131/synology/#enabling-synology-outbound-connections doesn't match this

cmd/tailscale/cli/web.html
phirework: Discussed with @mayakacz offline and she suggested removing entirely, because calling it out in an alert box didn't make sense when it's a success case, and leaving it up by itself underneath reauth/advertise links looked (imo) messy and didn't follow a clear content hierarchy 

Screen Shot 2023-01-24 at 17 55 02

bradfitz: I want something there for debugging with users.

If nothing's there then we don't know whether they're just using an old version that didn't say anything.

phirework: But yes I can leave that to a separate PR

bradfitz (approved):

bradfitz:
maybe "fix TUN mode display" instead of "fix false positive" ?

Issue #5719, opened 2022-09-22
Mac userspace subnet router binding to wrong interface


bradfitz:
On macOS, subnet routers forwarding connections on with netstack can bind to the wrong device when the machine has multiple NICs.

Situation from debug call:

* macOS (one of the GUI builds)
* three network devices (wired ethernet (primary), wifi, USB ethernet dongle with a /30 to a raspberry pi)
* `tailscale up --advertise-routes=192.168.1.0/30`

The netstack forwarding code gets the packet and does an outbound TCP dial but netns binds the socket to the primary wired ethernet's NIC instead of the USB ethernet dongle.

We should look at the routing table and pick the right device.

/cc @andrew-d @danderson @crawshaw @maisem @quinnypig

andrew-d:
I'm pretty sure this is because this code here picks the default route (i.e. "to the internet") to bind to, rather than the route for a specific destination:
https://github.com/tailscale/tailscale/blob/0607832397046fd0acb73daf8e00ef17b171a5c6/net/netns/netns_darwin_tailscaled.go#L35

andrew-d:
@bradfitz - Any chance you can get an (appropriately-redacted, if necessary) dump of `netstat -nr` from the system in question?

bradfitz:
The line you link to is for tailscaled, not the GUI version. But similar.

I don't have the routing table output but it was as described above: a /30 on a separate link that wasn't the default route's.

QuinnyPig:
I do; it's my system in question. Here's a partial dump just because it's wordy as hell, particularly on the ipv6 section; en0 is wired, en1 is wireless (same network), en8 is the other interface with a small network attached, utun5 is Tailscale. 
```
Internet:
Destination        Gateway            Flags           Netif Expire
default            192.168.1.1        UGScg             en0       
default            192.168.1.1        UGScIg            en1       
default            link#34            UCSIg           utun5       
10.98.32/28        link#22            UCS               en8      !
10.98.32.11/32     link#22            UCS               en8      !
10.98.32.15        ff:ff:ff:ff:ff:ff  UHLWbI            en8      !
100.64/10          link#34            UCS             utun5       
100.93.10.83       link#34            UHWIi           utun5       
100.99.181.129     link#34            UHWIi           utun5       
100.100.100.100/32 link#34            UCS             utun5       
100.100.100.100    link#34            UHWIi           utun5       
100.126.71.95      100.126.71.95      UH              utun5       
[...]

PR #7065, opened 2023-01-25
net/netns: add functionality to bind outgoing sockets based on route table

andrew-d:
When turned on via environment variable (off by default), this will use the BSD routing APIs to query what interface index a socket should be bound to, rather than binding to the default interface in all cases.

Updates https://github.com/tailscale/tailscale/issues/5719
Updates https://github.com/tailscale/tailscale/pull/5940

Change-Id: Ib4c919471f377b7a08cd3413f8e8caacb29fee0b

### DIFF
--- a/net/netns/netns_darwin.go
+++ b/net/netns/netns_darwin.go
@@ -11,10 +11,14 @@ import (
 	"fmt"
 	"log"
 	"net"
+	"net/netip"
+	"os"
 	"strings"
 	"syscall"
 
+	"golang.org/x/net/route"
 	"golang.org/x/sys/unix"
+	"tailscale.com/envknob"
 	"tailscale.com/net/interfaces"
 	"tailscale.com/types/logger"
 )
@@ -25,6 +29,10 @@ func control(logf logger.Logf) func(network, address string, c syscall.RawConn) 
 	}
 }
 
+var bindToInterfaceByRouteEnv = envknob.RegisterBool("TS_BIND_TO_INTERFACE_BY_ROUTE")
+
+var errInterfaceIndexInvalid = errors.New("interface index invalid")
+
 // controlLogf marks c as necessary to dial in a separate network namespace.
 //
 // It's intentionally the same signature as net.Dialer.Control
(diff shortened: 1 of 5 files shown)


bradfitz:
This is the sort of commit that _really_ needs to reference an issue :)

net/netns/netns_darwin.go
```diff
@@ -25,6 +29,10 @@ func control(logf logger.Logf) func(network, address string, c syscall.RawConn)
 	}
 }
 
+var bindToDeviceByRoute = envknob.RegisterBool("TS_BIND_TO_DEVICE_BY_ROUTE")
```
mihaip, line 32: Environment variables are annoying to set on macOS and ~impossible on iOS. Should we also add a node attribute that can be used to enable this?

net/netns/netns_darwin.go
```diff
@@ -34,15 +42,109 @@ func controlLogf(logf logger.Logf, network, address string, c syscall.RawConn) e
...
 		return nil
 	}
 
+	var idx int
+	if bindToDeviceByRoute() {
+		idx, err = interfaceIndexFor(addr)
+
+		// Verify that we didn't just choose the Tailscale interface;
+		// if so, we fall back to binding from the default.
+		if err == nil {
```
mihaip, line 58: Add logging if we get an error from `interfaceIndexFor`?

net/netns/netns_darwin.go
```diff
@@ -34,15 +42,109 @@ func controlLogf(logf logger.Logf, network, address string, c syscall.RawConn) e
...
+			_, tsif, err2 := interfaces.Tailscale()
+			if err2 == nil && tsif.Index == idx {
+				err = errInterfaceIndexInvalid
```
mihaip, line 61: Logging for this error state too?

net/netns/netns_darwin.go
```diff
@@ -34,15 +42,109 @@ func controlLogf(logf logger.Logf, network, address string, c syscall.RawConn) e
...
+			return nil
+		}
+	}
+
 	return bindConnToInterface(c, network, address, idx, logf)
 }
 
+// interfaceIndexFor returns the interface index that we should bind to in
+// order to send traffic to the provided address.
+func interfaceIndexFor(addr netip.Addr) (int, error) {
```
mihaip, line 80: Depending on how expensive this is, we could add a cache for this (and invalidate it via ` LinkMonitor` `RegisterChangeCallback` callback).

net/netns/netns_darwin.go
andrew-d: Added 👍

net/netns/netns_darwin.go
andrew-d: Added 👍

net/netns/netns_darwin.go
andrew-d: This is pretty expensive, yeah, which is why I'm not planning on enabling this by default. In the long term, I want to move ~all of this to the `LinkMonitor` by having it maintain a copy of the system's route table so we don't need to re-parse everything.

In the mean time, happy to do this; any suggestions or thoughts on where the callback should be registered, since we don't really have persistent state conveniently available?

net/netns/netns_darwin.go
andrew-d: Yep, good idea; will add.

andrew-d: Added 👍

net/netns/netns_darwin.go
```diff
@@ -34,15 +42,118 @@ func controlLogf(logf logger.Logf, network, address string, c syscall.RawConn) e
 		// Don't bind to an interface for localhost connections.
 		return nil
 	}
-	idx, err := interfaces.DefaultRouteInterfaceIndex()
+
+	host, _, err := net.SplitHostPort(address)
```
mihaip, line 46: The splitting/parsing/validation wasn't being done before. For both performance and correctness/paranoia reasons you may want to keep this only inside the branch for the new code path.

net/netns/netns_darwin.go
mihaip: I think for now we don't have to do any of it, was just speculating.

When we do start to do it, triggering the cache invalidation from `LocalBackend`'s `linkChange` seems like a reasonable place (as far as `LocalBackend` also enabling this code path to start with).

mihaip (approved):

andrew-d:
@mihaip - I wrote a test and refactored this a fair bit; during my testing, I found that it's sometimes possible that the first call to the route APIs returns the gateway IP rather than the interface, and we need to recursively call ourselves (exactly once) to get the interface for a given gateway IP. Mind re-reviewing this?

net/netns/netns_darwin_test.go
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
```
mihaip, line 1: 2023?

mihaip (approved):

net/netns/netns_darwin.go
bradfitz: Filed https://github.com/tailscale/tailscale/issues/7077 too now that you reminded me :)

net/netns/netns_darwin.go
```diff
@@ -34,15 +42,145 @@ func controlLogf(logf logger.Logf, network, address string, c syscall.RawConn) e
...
+
+	var routeAddr route.Addr
+	if addr.Is4() {
+		routeAddr = &route.Inet4Addr{IP: addr.As4()}
+	} else {
+		routeAddr = &route.Inet6Addr{IP: addr.As16()}
+	}
+
+	rm := route.RouteMessage{
+		Version: unix.RTM_VERSION,
```
bradfitz, line 118: pin this to a fixed version with comment?

net/netns/netns_darwin.go
```diff
@@ -34,15 +42,145 @@ func controlLogf(logf logger.Logf, network, address string, c syscall.RawConn) e
...
+	b, err := rm.Marshal()
+	if err != nil {
+		return 0, fmt.Errorf("marshaling RouteMessage: %w", err)
+	}
+	_, err = unix.Write(fd, b)
+	if err != nil {
+		return 0, fmt.Errorf("writing message: %w", err)
+	}
+	var buf [2048]byte
+	n, err := unix.Read(fd, buf[:])
```
bradfitz, line 136: // safe to do single read because X and smaller than 4KB?

bradfitz (commented):
LGTM

tunix:
Hi @bradfitz , I've installed tailscale on my Mac via homebrew. I want to enable this feature but couldn't figure out whether I have this feature on my computer (I'm on version 1.72.2) or what to write into nodeAttrs. What's the value to use inside attr part to enable this?

PR #7067, opened 2023-01-25
tailcfg: add caps and SSH action for SSH hauling

skriptble:
Add capabilities and an SSH action for SSH session hauling.

Signed-off-by: Kris Brandow <kris@tailscale.com>

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -1747,6 +1747,9 @@ const (
 	CapabilityWakeOnLAN = "https://tailscale.com/cap/wake-on-lan"
 	// CapabilityIngress grants the ability for a peer to send ingress traffic.
 	CapabilityIngress = "https://tailscale.com/cap/ingress"
+	// CapabilitySSHSessionHaul grants the ability to receive SSH session logs
+	// from a peer.
+	CapabilitySSHSessionHaul = "https://tailscale.com/cap/ssh-session-haul"
 
 	// Funnel warning capabilities used for reporting errors to the user.
 
@@ -1767,6 +1770,8 @@ const (
 const (
 	// NodeAttrFunnel grants the ability for a node to host ingress traffic.
 	NodeAttrFunnel = "funnel"
+	// NodeAttrSSHAggregator grants the ability for a node to collect SSH sessions.
+	NodeAttrSSHAggregator = "ssh-aggregator"
 )
 
 // SetDNSRequest is a request to add a DNS record.
@@ -1940,6 +1945,10 @@ type SSHAction struct {
 	// AllowLocalPortForwarding, if true, allows accepted connections
 	// to use local port forwarding if requested.
 	AllowLocalPortForwarding bool `json:"allowLocalPortForwarding,omitempty"`
+
+	// SessionHaulTargetNode, if non-empty, is the Stable ID of a peer to
+	// stream this SSH session's logs to.
+	SessionHaulTargetNode StableNodeID `json:"sessionHaulTargetNode,omitempty"`
 }
 
 // OverTLSPublicKeyResponse is the JSON response to /key?v=<n>


tailcfg/tailcfg.go
```diff
@@ -1725,6 +1725,7 @@ const (
 	CapabilitySSHRuleIn          = "https://tailscale.com/cap/ssh-rule-in"           // some SSH rule reach this node
 	CapabilityDataPlaneAuditLogs = "https://tailscale.com/cap/data-plane-audit-logs" // feature enabled
 	CapabilityDebug              = "https://tailscale.com/cap/debug"                 // exposes debug endpoints over the PeerAPI
+	CapabilitySSHAggregator      = "https://tailscale.com/cap/ssh-aggregator"        // enable SSH session aggregation
```
maisem, line 1728: I think this should be `ssh-aggregator` based on https://tailscale.slack.com/archives/CQQ1AUCCE/p1674669132275479?thread_ts=1674662145.743249&cid=CQQ1AUCCE

cc @bradfitz

tailcfg/tailcfg.go
```diff
@@ -1940,6 +1943,10 @@ type SSHAction struct {
 	// AllowLocalPortForwarding, if true, allows accepted connections
 	// to use local port forwarding if requested.
 	AllowLocalPortForwarding bool `json:"allowLocalPortForwarding,omitempty"`
+
+	// SessionHauling, if non-empty, is the IP address of a peer to stream this
+	// SSH session's logs to.
+	SessionHauling string `json:"sessionHauling,omitempty"`
```
maisem, line 1949: should this be the peers NodeID instead of the IP?
If it has to be an IP then it should likely be `netip.Addr`

bradfitz, line 1949: this name sounds like a bool. How about `SessionHaulDest`?

Also, I think we should use a node stable ID here instead of IP address, lest the source not have that address family available. See `func peerAPIBase`.

bradfitz: Or `SessionHaulTargetNode`.

tailcfg/tailcfg.go
```diff
@@ -1747,6 +1748,8 @@ const (
 	CapabilityWakeOnLAN = "https://tailscale.com/cap/wake-on-lan"
 	// CapabilityIngress grants the ability for a peer to send ingress traffic.
 	CapabilityIngress = "https://tailscale.com/cap/ingress"
+	// CapabilitySSHSessionHaul grants the ability to send SSH session logs.
+	CapabilitySSHSessionHaul = "https://tailscale.com/cap/ssh-session-haul"
```
bradfitz, line 1752: why is this a cap at all, as opposed to just its presence in the SSHConfig? If control can send one it can send both, but why two?

bradfitz: oh, this is in the inter-node section! carry on. Maybe reword the comment to make that more clear. I see some of them are more clear about that but then the last few got shorter.

skriptble: This is checked on the aggregator to know if the SSH server can send logs to it.

tailcfg/tailcfg.go
skriptble: @twitchyliquid64 pointed out that we might want to run the aggregator standalone in the future (maybe behind a subnet router), so being able to select a specific IP:Port that isn't a Tailscale node could be useful for that.

But we could also just change this setting (or add a new one) in the future to handle that, so I'm fine with changing both the name and the type here.

bradfitz: We'd need a new field to handle a `:port` later anyway, so let's add a new field later if/when we do something else.

bradfitz:
Open a bug too (I don't think we have one right now) and reference it in the commit message (and future commits)

tailcfg/tailcfg.go
skriptble: I'll also rename it to `NodeAttrSSHAggregator` since we've started prefixing them with `NodeAttr` instead of `Capability`.

tailcfg/tailcfg.go
skriptble: Updated to use a `tailcfg.StableNodeID` and updated name.

bradfitz (approved):

PR #7083, opened 2023-01-27
cmd/derpprobe: migrate to the prober framework

knyar:
`prober.DERP` was created in #5988 based on derpprobe. Having used it instead of derpprobe for a few months, I think we have enough confidence that it works and can now migrate derpprobe to use the prober framework and get rid of code duplication.

A few notable changes in behaviour:
- results of STUN probes over IPv4 and IPv6 are now reported separately;
- TLS probing now includes OCSP verification;
- probe names in the output have changed slightly;
- ability to send Slack notification from the prober has been removed. Instead, the prober now exports metrics in Expvar (/debug/vars) and Prometheus (/debug/varz) formats.

Fixes https://github.com/tailscale/corp/issues/8497

### DIFF
--- a/cmd/derpprobe/derpprobe.go
+++ b/cmd/derpprobe/derpprobe.go
@@ -3,72 +3,45 @@
 // license that can be found in the LICENSE file.
 
 // The derpprobe binary probes derpers.
-package main // import "tailscale.com/cmd/derper/derpprobe"
+package main
 
 import (
-	"bytes"
-	"context"
-	crand "crypto/rand"
-	"crypto/x509"
-	"encoding/json"
-	"errors"
+	"expvar"
 	"flag"
 	"fmt"
 	"html"
 	"io"
 	"log"
-	"net"
 	"net/http"
-	"os"
 	"sort"
-	"strings"
-	"sync"
 	"time"
 
-	"tailscale.com/derp"
-	"tailscale.com/derp/derphttp"
-	"tailscale.com/net/stun"
-	"tailscale.com/tailcfg"
-	"tailscale.com/types/key"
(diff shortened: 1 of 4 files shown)


cmd/derpprobe/derpprobe.go
```diff
@@ -101,471 +70,43 @@ func (st *overallStatus) addGoodf(format string, a ...any) {
...
-	st := getOverallStatus()
-	summary := "All good"
-	if (float64(len(st.bad)) / float64(len(st.bad)+len(st.good))) > 0.25 {
-		// This will generate an alert and page a human.
-		// It also ends up in Slack, but as part of the alert handling pipeline not
-		// because we generated a Slack notification from here.
-		w.WriteHeader(500)
-		summary = fmt.Sprintf("%d problems", len(st.bad))
-	}
-
```
DentonGentry, line 173: There is pleasingly less code.

cmd/derpprobe/derpprobe.go
```diff
@@ -101,471 +70,43 @@ func (st *overallStatus) addGoodf(format string, a ...any) {
...
-	var subjs []string
-	for k := range certs {
-		subjs = append(subjs, k)
-	}
-	sort.Strings(subjs)
-
-	soon := time.Now().Add(14 * 24 * time.Hour) // in 2 weeks; autocert does 30 days by default
-	for _, s := range subjs {
-		cert := certs[s]
-		if cert.NotBefore.Before(certReissueAfter) {
```
DentonGentry, line 149: I guess we don't need this check any more since no new DERPs would be created with a cert this old.

DentonGentry (approved):

PR #7085, opened 2023-01-27
net/netstat: add nil checks to Windows OSMetadata implementation

dblohm7:
The API documentation does claim to output empty strings under certain conditions, but we're sometimes seeing nil pointers in the wild, not empty strings.

Fixes https://github.com/tailscale/corp/issues/8878

### DIFF
--- a/net/netstat/netstat_windows.go
+++ b/net/netstat/netstat_windows.go
@@ -224,6 +224,7 @@ type moduleInfoConstraint interface {
 	_MIB_TCPROW_OWNER_MODULE | _MIB_TCP6ROW_OWNER_MODULE
 }
 
+// moduleInfo may return "", nil indicating a successful call but with empty data
 func moduleInfo[entryType moduleInfoConstraint](entry *entryType, proc *windows.LazyProc) (string, error) {
 	var buf []byte
 	var desiredLen uint32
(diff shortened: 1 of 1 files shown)


bradfitz (approved):
Add little doc on func moduleInfo that it can return "", nil? I assume callers are good?

dblohm7:
> Add little doc on func moduleInfo that it can return "", nil? I assume callers are good?

Will do. Callers are good.

mihaip:
This should most likely get cherrypicked into https://github.com/tailscale/tailscale/tree/release-branch/1.36

net/netstat/netstat_windows.go
```diff
@@ -249,6 +250,12 @@ func moduleInfo[entryType moduleInfoConstraint](entry *entryType, proc *windows.
 	}
 
 	basicInfo := (*_TCPIP_OWNER_MODULE_BASIC_INFO)(addr)
+	// GetOwnerModuleFromTcp*Entry is apparently using nil as an empty result
+	// under certain circumstances, so we check all the things.
+	if basicInfo == nil || basicInfo.moduleName == nil {
```
bradfitz, line 255: oh, I just noticed this latter check is unnecessary, as UTF16PtrToString is documented to handle nil:

```go
// UTF16PtrToString takes a pointer to a UTF-16 sequence and returns the corresponding UTF-8 encoded string.
// If the pointer is nil, it returns the empty string. It assumes that the UTF-16 sequence is terminated
// at a zero word; if the zero word is not present, the program may crash.
func UTF16PtrToString(p *uint16) string {
	if p == nil {
		return ""
	}
	if *p == 0 {
		return ""
	}

	// Find NUL terminator.
	n := 0
	for ptr := unsafe.Pointer(p); *(*uint16)(ptr) != 0; n++ {
		ptr = unsafe.Pointer(uintptr(ptr) + unsafe.Sizeof(*p))
	}

	return string(utf16.Decode(unsafe.Slice(p, n)))
}
```

PR #7091, opened 2023-01-27
all: update copyright and license header in source files

willnorris:
This implements the changes discussed in https://github.com/tailscale/tailscale/issues/6865#issuecomment-1406947636.

This has two commits: the first is purely mechanical and updates the existing file headers.  The second updates the tooling that manages those headers.


DentonGentry (approved):
I will admit to not having scrutinized all 800 files.

willnorris:
I did :)

PR #7093, opened 2023-01-27
wgengine/magicsock: de-dup disco pings

### LINKED ISSUE #7078: wgengine/magicsock: disco packets lost under high throughput load on Linux
Disco packets may not arrive on the bpf-filtered raw sockets under high throughput load. The buffers for these sockets are currently sized to system defaults.

```
# ss -w -e -m
Recv-Q     Send-Q         Local Address:Port                Peer Address:Port     Process
0          0                    0.0.0.0:udp                      0.0.0.0:*         ino:121415 sk:4004 cgroup:/system.slice/tailscaled.service <->
	 skmem:(r0,rb425984,t0,tb212992,f0,w0,o592,bl0,d0)
0          0                          *:udp                            *:*         ino:121420 sk:4005 cgroup:/system.slice/tailscaled.service v6only:0 <->
	 skmem:(r0,rb425984,t0,tb212992,f0,w0,o400,bl0,d0)
```

If disco ping/pong packets are lost, valid UDP path heartbeating (every 3s) may exceed `trustUDPAddrDuration` (6.5s) causing magicsock to transmit data via both valid UDP and DERP paths until the next disco pong is received for the endpoint in question.

cc @raggi

jwhited:
Fixes #7078

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -172,9 +172,10 @@ func (m *peerMap) forEachEndpoint(f func(ep *endpoint)) {
 	}
 }
 
-// forEachEndpointWithDiscoKey invokes f on every endpoint in m
-// that has the provided DiscoKey.
-func (m *peerMap) forEachEndpointWithDiscoKey(dk key.DiscoPublic, f func(ep *endpoint)) {
+// forEachEndpointWithDiscoKey invokes f on every endpoint in m that has the
+// provided DiscoKey until f returns false or there are no endpoints left to
+// iterate.
+func (m *peerMap) forEachEndpointWithDiscoKey(dk key.DiscoPublic, f func(*endpoint) (keepGoing bool)) {
 	for nk := range m.nodesOfDisco[dk] {
 		pi, ok := m.byNodeKey[nk]
 		if !ok {
@@ -185,7 +186,9 @@ func (m *peerMap) forEachEndpointWithDiscoKey(dk key.DiscoPublic, f func(ep *end
 			// into Conn.
 			continue
 		}
-		f(pi.ep)
+		if !f(pi.ep) {
+			return
+		}
 	}
 }
 
@@ -1782,7 +1785,7 @@ func (c *Conn) receiveIPv6(buffs [][]byte, sizes []int, eps []conn.Endpoint) (in
 		reportToCaller := false
 		for i, msg := range batch.msgs[:numMsgs] {
 			ipp := msg.Addr.(*net.UDPAddr).AddrPort()
-			if ep, ok := c.receiveIP(msg.Buffers[0][:msg.N], ipp, &c.ippEndpoint6, c.closeDisco6 == nil); ok {
+			if ep, ok := c.receiveIP(msg.Buffers[0][:msg.N], ipp, &c.ippEndpoint6); ok {
 				metricRecvDataIPv6.Add(1)
 				eps[i] = ep
 				sizes[i] = msg.N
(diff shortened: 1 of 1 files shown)


wgengine/magicsock/magicsock.go
```diff
@@ -2233,11 +2227,17 @@ func (c *Conn) handlePingLocked(dm *disco.Ping, src netip.AddrPort, di *discoInf
 	var numNodes int
 	if isDerp {
 		if ep, ok := c.peerMap.endpointForNodeKey(derpNodeSrc); ok {
+			if ep.isDuplicatePing(dm.TxID) {
+				return
+			}
 			ep.addCandidateEndpoint(src)
```
bradfitz, line 2233: this added an extra Lock/Unlock. maybe that's a drop in the bucket, but you could pass dm.TxID to `addCandidateEndpoint` (or some new name or new variant) that does the same work as `addCandidateEndpoint` but returns some value (new bool or sentinel error) that says it was a dup and not change the number of lock operations.

wgengine/magicsock/magicsock.go
```diff
@@ -4195,6 +4196,21 @@ func (de *endpoint) updateFromNode(n *tailcfg.Node, heartbeatDisabled bool) {
 	de.sendFunc.Store(nil)
 }
 
+// isDuplicatePing records the txID for the endpoint. It returns true if
```
bradfitz, line 4199: Go style nit: s/returns true if/reports whether/

Also: this name sounds like it wouldn't have a side effect of recording the last value. Not that I have a better suggestion.

bradfitz (approved):

wgengine/magicsock/magicsock.go
jwhited: I ended up plumbing the TxID down into `addCandidateEndpoint()` and tracking it on `endpointState` where there are other ping-related states.

wgengine/magicsock/magicsock.go
jwhited: This function was removed in favor of plumbing into `addCandidateEndpoint()`, which uses `reports` instead of `returns` in its doc.

wgengine/magicsock/magicsock.go
```diff
@@ -172,9 +172,10 @@ func (m *peerMap) forEachEndpoint(f func(ep *endpoint)) {
 	}
 }
 
-// forEachEndpointWithDiscoKey invokes f on every endpoint in m
-// that has the provided DiscoKey.
-func (m *peerMap) forEachEndpointWithDiscoKey(dk key.DiscoPublic, f func(ep *endpoint)) {
+// forEachEndpointWithDiscoKey invokes f on every endpoint in m that has the
+// provided DiscoKey until f returns true or there are no endpoints left to
+// iterate.
+func (m *peerMap) forEachEndpointWithDiscoKey(dk key.DiscoPublic, f func(ep *endpoint) bool) {
```
bradfitz, line 178: I'd write the signature with named result like `f func(*endpoint) (stop bool)` and omitting the `ep` name.

But I think that's a little backwards than typical. Normally I think it's `(keepGoing bool)`.

But as long as it's named.

wgengine/magicsock/magicsock.go
```diff
@@ -2231,19 +2225,29 @@ func (c *Conn) handlePingLocked(dm *disco.Ping, src netip.AddrPort, di *discoInf
...
-			ep.addCandidateEndpoint(src)
+			if ep.addCandidateEndpoint(src, dm.TxID) {
+				return
+			}
 			numNodes = 1
 		}
 	} else {
-		c.peerMap.forEachEndpointWithDiscoKey(di.discoKey, func(ep *endpoint) {
-			ep.addCandidateEndpoint(src)
+		c.peerMap.forEachEndpointWithDiscoKey(di.discoKey, func(ep *endpoint) bool {
```
bradfitz, line 2237: I'd even name it here.

bradfitz (approved):

wgengine/magicsock/magicsock.go
jwhited: changed return to `(keepGoing bool)`, updated doc, and omitted `ep`

wgengine/magicsock/magicsock.go
jwhited: updated

PR #7094, opened 2023-01-27
ipn/ipnlocal: add /reset-auth LocalAPI endpoint

mihaip:
The iOS has a command to reset the persisted state of the app, but it was doing its own direct keychain manipulation. This proved to be brittle (since we changed how preferences are stored with #6022), so we instead add a LocalAPI endpoint to do do this, which can be updated in tandem.

This clears the same state as the iOS implementation (tailscale/corp#3186), that is the machine key and preferences (which includes the node key). Notably this does not clear the logtail ID, so that logs from the device still end up in the same place.

Updates tailscale/corp#8923

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -2148,6 +2148,20 @@ func (b *LocalBackend) initMachineKeyLocked() (err error) {
 	return nil
 }
 
+// clearMachineKeyLocked is called to clear the persisted and in-memory
+// machine key, so that initMachineKeyLocked (called as part of starting)
+// generates a new machine key.
+//
+// b.mu must be held.
+func (b *LocalBackend) clearMachineKeyLocked() error {
+	if err := b.store.WriteState(ipn.MachineKeyStateKey, nil); err != nil {
+		return err
+	}
+	b.machinePrivKey = key.MachinePrivate{}
+	b.logf("machine key cleared")
+	return nil
+}
+
 // migrateStateLocked migrates state from the frontend to the backend.
 // It is a no-op if prefs is nil
 // b.mu must be held.
@@ -4758,4 +4772,22 @@ func (b *LocalBackend) ListProfiles() []ipn.LoginProfile {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 	return b.pm.Profiles()
+}
+
+// ResetAuth resets the authentication state, including persisted keys. Also
+// has the side effect of removing all profiles and reseting preferences. The
+// backend is left with a new profile, ready for StartLoginInterative to be
+// called to register it as new node.
+func (b *LocalBackend) ResetAuth() error {
+	b.mu.Lock()
+	b.resetControlClientLockedAsync()
+	if err := b.clearMachineKeyLocked(); err != nil {
+		b.mu.Unlock()
+		return err
+	}
+	if err := b.pm.DeleteAllProfiles(); err != nil {
+		b.mu.Unlock()
+		return err
+	}
+	return b.resetForProfileChangeLockedOnEntry()
 }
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/local.go
```diff
@@ -4759,3 +4773,19 @@ func (b *LocalBackend) ListProfiles() []ipn.LoginProfile {
 	defer b.mu.Unlock()
 	return b.pm.Profiles()
 }
+
+// ResetAuth resets the authentication state, including persisted keys. The
```
bradfitz, line 4777: mention that it's for all profiles?

ipn/ipnlocal/profiles.go
```diff
@@ -382,6 +383,24 @@ func (pm *profileManager) DeleteProfile(id ipn.ProfileID) error {
...
+// DeleteAllProfiles removes all known profiles and switches to a new empty
+// profile.
+func (pm *profileManager) DeleteAllProfiles() error {
+	metricDeleteAllProfile.Add(1)
+
+	for _, kp := range pm.knownProfiles {
+		if err := pm.store.WriteState(kp.Key, nil); err != nil {
+			// Write profiles we've already deleted, but return the original
+			// error.
+			pm.writeKnownProfiles()
```
maisem, line 395: why? this wouldn't have changed yet. If you want that, you need to call `delete(pm.knownProfiles, k)`
You also probably want to do pm.NewProfile first

ipn/ipnlocal/local.go
```diff
@@ -4759,3 +4773,19 @@ func (b *LocalBackend) ListProfiles() []ipn.LoginProfile {
...
+// backend is left with a new profile, ready for StartLoginInterative to be
+// called to register it as new node.
+func (b *LocalBackend) ResetAuth() error {
+	b.mu.Lock()
+	if err := b.clearMachineKeyLocked(); err != nil {
```
bradfitz, line 4782: don't we have to stop some things before we do this?

like stop the control client if it's connected?

it works for your use case (already broken), but I don't want call this on a running state to crash the backend and I'm not convinced the machine key won't be accessed while it's running. not a data race, but an assumption that it can access the machine key under a lock and then transform it to a public key and have that operation crash.

ipn/ipnlocal/profiles.go
mihaip: Was implementing feedback from a previous iteration of this code (https://github.com/tailscale/tailscale/pull/6434#discussion_r1028674287). I can just do the return, I don't think we can really leave things in a consistent state either way.

maisem: Oh, yeah not leaving them in an inconsistent state sgtm.
But you still need to call `delete(pm.knownProfiles, k)` otherwise you will end up writing the same contents.

ipn/ipnlocal/local.go
maisem: I think you also need to update `createGetMachinePrivateKeyFunc`

mihaip: `resetForProfileChangeLockedOnEntry()` ends up doing this shortly after (it calls `Start`, which calls `resetControlClientLockedAsync`). I've added an explicit call before we clear the machine key.

mihaip: > I think you also need to update `createGetMachinePrivateKeyFunc`

I don't think that's necessary (empirically a new machine key is generated). We call it again when creating the new control client, thus we end up capturing a new `cache` variable in the closure.

bradfitz (approved):
But LGTM, misc comments aside.

ipn/ipnlocal/profiles.go
mihaip: I added something like this (it's a bit more awkward, because we want to `delete()` all of the profile entries that we are able to successfully remove from the state store, thus we need to keep track of that).

maisem (approved):

PR #7095, opened 2023-01-28
ipn/ipnlocal: remove unused TestOnlyPublicKeys function

maisem:
#cleanup

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -3994,24 +3994,6 @@ func (b *LocalBackend) OperatorUserID() string {
 	return u.Uid
 }
 
-// TestOnlyPublicKeys returns the current machine and node public
-// keys. Used in tests only to facilitate automated node authorization
-// in the test harness.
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/local.go
```diff
@@ -3994,24 +3994,6 @@ func (b *LocalBackend) OperatorUserID() string {
 	return u.Uid
 }
 
-// TestOnlyPublicKeys returns the current machine and node public
-// keys. Used in tests only to facilitate automated node authorization
-// in the test harness.
-func (b *LocalBackend) TestOnlyPublicKeys() (machineKey key.MachinePublic, nodeKey key.NodePublic) {
```
mihaip, line 4000: I believe this is used in corp test code: https://github.com/tailscale/corp/blob/41e9f6e1b55fd1836acea1036f4a9c4f2416dab8/ipn/e2e_test.go#L528-L532

maisem: so it is 🤦🏼‍♂️

PR #7101, opened 2023-01-28
cmd/k8s-operator: use the unstable operator image

danderson:
There is no stable release yet, and for alpha we want people on the unstable build while we iterate.

Updates #502


danderson:
Er, oops, no, not this.

Issue #1534, opened 2021-03-17
Support iOS on demand connections


mattradford:
**Is your feature request related to a problem? Please describe.**

Submitting in response to this Twitter thread with @crawshaw: https://twitter.com/davidcrawshaw/status/1372299409451413507.

The official [Wireguard iOS app](https://apps.apple.com/us/app/wireguard/id1441195209) allows for [on-demand VPN connections](https://developer.apple.com/documentation/networkextension/personal_vpn/vpn_on_demand_rules) - auto-connecting to a VPN when switching networks. I currently use this to connect to my home Wireguard VPN whenever I leave the house, i.e. not on my home wifi.

With the release of exit node functionallity in Tailscale 1.6.0, it becomes possible to use Tailscale as a replacement for this Wireguard VPN, by routing all traffic back to an exit node at my home.

I cannot currently have this connection be automatically enabled when I leave the house.

**Describe the solution you'd like**

Quite a few other iOS VPN apps allow for on demand connections. The Tailscale iOS app does not.

In the Tailscale iOS app, allow for:

-  on demand connections
-  the ability to include or exclude cellular and specified wifi networks

If users have multiple exit nodes available, Tailscale should connect to the last one chosen.

This is how it looks in the Wireguard iOS app:

![IMG_06BDA28D8CD0-1](https://user-images.githubusercontent.com/2804540/111551904-29fa6380-8779-11eb-912f-b4cc15458fea.jpeg)

**Describe alternatives you've considered**

An alternative would be to support Shortcuts, such as Tracker VPN does: https://medium.com/@equinux/thats-what-you-ve-been-waiting-for-siri-shortcuts-in-vpn-tracker-world-connect-1-1-2eb5b56ebba0.

This means people have to set up their own shortcuts to toggle VPN. It's less discoverable, but does allow for more conditions than just changing networks. Hey, why not both? ;)

**Additional context**

I suspect this would allow some people to ditch commercial VPN providers, especially if we can share exit nodes.

mcompton13:
I would also like a feature like this, but with a slight twist. I'd like to keep tailscale always running on my phone (or laptop) all the time so I can always access servers/services on running in my personal tailscale network, but it would be awesome if it could conditionally use an exit node when I join an untrusted wifi networks. That way if I'm on my home network wifi or using LTE talescale does not need to worry about routing traffic through my exit node, but as soon as I join and use the wifi at the airport all of my traffic is encrypted and routed over my tailscale network through my exit node.

[...]

PR #7104, opened 2023-01-29
tailcfg, etc: plumb OnDemandDomains from the server

crawshaw:
This lets the server tell macOS/iOS that tailscale should be started to handle any "ts.net" domains.

Updates #1534


crawshaw:
It's not clear to me I understand what dns.OSConfig is all about, but it is ultimately the struct that is passed to the relevant code on macOS/iOS that configures DNS.

I would be open to putting this somewhere else if you have suggestions.

crawshaw:
Turns out I made a critical mistake in some of my plumbing and I can't (yet) get these values where I need to get them. Closing for now, will reopen if I can figure it out.

Issue #7108, opened 2023-01-29
QNAP authlogin fails if WebUI port number has been changed


DentonGentry:
### What is the issue?

In the QNAP Control Panel > System > General is a setting to change the port number for the Web UI from 8080 to some other port. We currently use http://localhost:8080/ to reach authLogin.cgi, which fails if the user has changed the port number.

### Steps to reproduce

Change the port number in the Control Panel.

### OS

Other

### OS version

QTS 5.0.1.2248

### Tailscale version

1.36.0

PR #7109, opened 2023-01-29
cmd/tailscale: use request Schema+Host for QNAP authLogin.cgi

DentonGentry:
QNAP allows users to set the port number for the management WebUI, which includes authLogin.cgi. If they do, then connecting to localhost:8080 fails.

https://github.com/tailscale/tailscale-qpkg/issues/74#issuecomment-1407486911

We have to use the Scheme+Path from the request but disable any TLS verification for https:
- The UI can force https, but presents a completely invalid certificate
- The UI can change the port number, and though https://download.qnap.com/dev/API_QNAP_QTS_Authentication.pdf documents the use of localhost:8080 for authentication this fails if the user changed the port number.

This reverts commit 467ace7d0c0655e69cc219ec380c0119d909f36b.

Fixes https://github.com/tailscale/tailscale/issues/7108

### DIFF
--- a/cmd/tailscale/cli/web.go
+++ b/cmd/tailscale/cli/web.go
@@ -228,33 +228,48 @@ func qnapAuthn(r *http.Request) (string, *qnapAuthResponse, error) {
 	return "", nil, fmt.Errorf("not authenticated by any mechanism")
 }
 
+// qnapAuthnURL returns the auth URL to use by inferring where the UI is
+// running based on the request URL. This is necessary because QNAP has so
+// many options, see https://github.com/tailscale/tailscale/issues/7108
+// and https://github.com/tailscale/tailscale/issues/6903
+func qnapAuthnURL(requestUrl string, query url.Values) string {
+	in, err := url.Parse(requestUrl)
+	scheme := ""
+	host := ""
+	if err != nil || in.Scheme == "" {
+		log.Printf("Cannot parse QNAP login URL %v", err)
+
+		// try localhost and hope for the best
+		scheme = "http"
+		host = "localhost"
+	} else {
+		scheme = in.Scheme
+		host = in.Host
+	}
+
+	u := url.URL{
+		Scheme:   scheme,
+		Host:     host,
+		Path:     "/cgi-bin/authLogin.cgi",
+		RawQuery: query.Encode(),
+	}
+
+	return u.String()
+}
+
 func qnapAuthnQtoken(r *http.Request, user, token string) (string, *qnapAuthResponse, error) {
 	query := url.Values{
 		"qtoken": []string{token},
 		"user":   []string{user},
 	}
-	u := url.URL{
-		Scheme:   "http",
-		Host:     "127.0.0.1:8080",
-		Path:     "/cgi-bin/authLogin.cgi",
-		RawQuery: query.Encode(),
-	}
-
-	return qnapAuthnFinish(user, u.String())
+	return qnapAuthnFinish(user, qnapAuthnURL(r.URL.String(), query))
 }
 
 func qnapAuthnSid(r *http.Request, user, sid string) (string, *qnapAuthResponse, error) {
 	query := url.Values{
 		"sid": []string{sid},
 	}
-	u := url.URL{
-		Scheme:   "http",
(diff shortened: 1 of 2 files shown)


bradfitz (commented):
That Scheme is not populated on incoming requests iirc. It's for outgoing requests. See http docs on that type.

DentonGentry:
Parsed the request URL separately.

cmd/tailscale/cli/web.go
```diff
@@ -228,33 +228,49 @@
...
+func qnapAuthnURL(requestUrl string, query url.Values) string {
+	in, err := url.Parse(requestUrl)
+	scheme := in.Scheme
+	host := in.Host
+	if err != nil || scheme == "" {
+		errUrl := requestUrl
+		if len(errUrl) > 40 {
+			errUrl = requestUrl[0:39]
+		}
+		log.Printf("Cannot parse QNAP login URL %v: %v", errUrl, err)
```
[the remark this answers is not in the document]
DentonGentry: I'd truncated it for that reason, but I don't really want CI failures in this file all the time.

cmd/tailscale/cli/web.go
```diff
@@ -228,33 +228,45 @@ func qnapAuthn(r *http.Request) (string, *qnapAuthResponse, error) {
...
-		"qtoken": []string{token},
-		"user":   []string{user},
+// qnapAuthnURL returns the auth URL to use by inferring where the UI is
+// running based on the request URL. This is necessary because QNAP has so
+// many options, see https://github.com/tailscale/tailscale/issues/7108
+// and https://github.com/tailscale/tailscale/issues/6903
+func qnapAuthnURL(requestUrl string, query url.Values) string {
+	in, err := url.Parse(requestUrl)
+	scheme := in.Scheme
+	host := in.Host
```
maisem, line 238: this will panic if `err != nil`

cmd/tailscale/cli/web_test.go
```diff
@@ -34,9 +37,59 @@ func TestUrlOfListenAddr(t *testing.T) {
...
+			want: "https://[ff7d:0:1:2::1]/cgi-bin/authLogin.cgi?qtoken=token",
+		},
+		{
+			name: "hostname https",
+			in:   "https://qnap.example.com/",
+			want: "https://qnap.example.com/cgi-bin/authLogin.cgi?qtoken=token",
+		},
+		{
+			name: "invalid URL",
+			in:   "This is not a URL, it is a really really really really really really really really really really really really long string to exercise the URL truncation code in the error path.",
```
maisem, line 84: ... why doesn't this panic

cmd/tailscale/cli/web.go
DentonGentry: I never found a condition where url.Parse returns an actual error, it instead returns err=nil and a struct with no fields populated. That is why the conditional includes `scheme == ""`, because that is what it returns when it cannot parse anything out of the URL.

Nonetheless it is a good point that tailscaled would crash if we ever did find a case where url.Parse returns an err. I think the updated code works: if err != nil the conditional will be short-circuited and in.Scheme won't be evaluated.

cmd/tailscale/cli/web_test.go
maisem: I got it to panic with `in: ":"`

maisem:
```
+		{
+			name: "invalidURL",
+			in:   ":",
+			want: "http://localhost/cgi-bin/authLogin.cgi?qtoken=token",
+		},
```

DentonGentry: I'm unconvinced that there is any set of inputs for which url.Parse returns an err.

maisem (approved):

cmd/tailscale/cli/web_test.go
DentonGentry: Looked at url.Parse's own tests: https://cs.opensource.google/go/go/+/master:src/net/url/url_test.go;l=697

I verified that if I remove the `err == nil` conditional, the code panics. So the last test case does exercise the err != nil path.

maisem (approved):

PR #7111, opened 2023-01-29
tailcfg, localapi: plumb device token to server

crawshaw:
questions in the form of a PR:

- I need to get this value out of the iOS app, not the networkextension, and localapi was handy. Is this a reasonable use of localAPI?
- should we send new stuff to control in Hostinfo? is there somewhere else this belongs? (it has the awkward property that on start the device token will be empty which I'm not a fan of, but otherwise looks like hostinfo stuff.)

### DIFF
--- a/client/tailscale/apitype/apitype.go
+++ b/client/tailscale/apitype/apitype.go
@@ -32,3 +32,9 @@ type WaitingFile struct {
 	Name string
 	Size int64
 }
+
+// SetPushDeviceTokenRequest is the body POSTed to the LocalAPI endpoint /set-device-token.
+type SetPushDeviceTokenRequest struct {
+	// PushDeviceToken is the iOS/macOS APNs device token (and any future Android equivalent).
+	PushDeviceToken string
+}
--- a/hostinfo/hostinfo.go
+++ b/hostinfo/hostinfo.go
@@ -50,6 +50,7 @@ func New() *tailcfg.Hostinfo {
 		GoVersion:       runtime.Version(),
 		Machine:         condCall(unameMachine),
 		DeviceModel:     deviceModel(),
+		PushDeviceToken: pushDeviceToken(),
 		Cloud:           string(cloudenv.Get()),
 		NoLogsNoSupport: envknob.NoLogsNoSupport(),
 		AllowsUpdate:    envknob.AllowsRemoteUpdate(),
@@ -153,11 +154,15 @@ func GetEnvType() EnvType {
 }
 
 var (
-	deviceModelAtomic atomic.Value // of string
-	osVersionAtomic   atomic.Value // of string
-	desktopAtomic     atomic.Value // of opt.Bool
-	packagingType     atomic.Value // of string
+	pushDeviceTokenAtomic atomic.Value // of string
+	deviceModelAtomic     atomic.Value // of string
+	osVersionAtomic       atomic.Value // of string
+	desktopAtomic         atomic.Value // of opt.Bool
+	packagingType         atomic.Value // of string
 )
+
+// SetPushDeviceToken sets the device token for use in Hostinfo updates.
+func SetPushDeviceToken(token string) { pushDeviceTokenAtomic.Store(token) }
 
 // SetDeviceModel sets the device model for use in Hostinfo updates.
 func SetDeviceModel(model string) { deviceModelAtomic.Store(model) }
@@ -173,6 +178,11 @@ func SetPackage(v string) { packagingType.Store(v) }
 
 func deviceModel() string {
 	s, _ := deviceModelAtomic.Load().(string)
+	return s
+}
+
+func pushDeviceToken() string {
+	s, _ := pushDeviceTokenAtomic.Load().(string)
 	return s
 }
 
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -1811,6 +1811,21 @@ func (b *LocalBackend) readPoller() {
 	}
 }
 
+// ResendHostinfoIfNeeded is called to recompute the Hostinfo and send
+// the new version to the control server.
+func (b *LocalBackend) ResendHostinfoIfNeeded() {
+	hi := hostinfo.New()
+
+	b.mu.Lock()
+	if b.hostinfo != nil {
+		hi.Services = b.hostinfo.Services
+	}
+	b.hostinfo = hi
+	b.mu.Unlock()
+
+	b.doSetHostinfoFilterServices(hi)
+}
+
 // WatchNotifications subscribes to the ipn.Notify message bus notification
 // messages.
 //
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -70,6 +70,7 @@ var handler = map[string]localAPIHandler{
 	"debug-packet-filter-rules":   (*Handler).serveDebugPacketFilterRules,
 	"derpmap":                     (*Handler).serveDERPMap,
 	"dev-set-state-store":         (*Handler).serveDevSetStateStore,
+	"set-push-device-token":       (*Handler).serveSetPushDeviceToken,
 	"dial":                        (*Handler).serveDial,
 	"file-targets":                (*Handler).serveFileTargets,
 	"goroutines":                  (*Handler).serveGoroutines,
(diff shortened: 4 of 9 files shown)


bradfitz:
Context? Link to bug?

crawshaw:
https://github.com/tailscale/corp/issues/8940

ipn/localapi/localapi.go
```diff
@@ -1200,6 +1201,24 @@ func (h *Handler) serveDial(w http.ResponseWriter, r *http.Request) {
 	<-errc
 }
 
+func (h *Handler) serveDeviceToken(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "POST" {
+		http.Error(w, "unsupported method", http.StatusMethodNotAllowed)
+		return
+	}
```
bradfitz, line 1208: need permission checks here. See PermitWrite elsewhere in this file.

Is this only for setting the token? "set-device-token" and "serveSetDeviceToken"? Or will you add GET later so the frontend can decide whether it needs to do the init.

ipn/localapi/localapi.go
```diff
@@ -1200,6 +1201,24 @@ func (h *Handler) serveDial(w http.ResponseWriter, r *http.Request) {
...
+	var params struct {
+		deviceToken string `json:"deviceToken"`
```
bradfitz, line 1210: capital DeviceToken

tailcfg/tailcfg.go
```diff
@@ -525,6 +525,7 @@ type Hostinfo struct {
 	Desktop         opt.Bool       `json:",omitempty"` // if a desktop was detected on Linux
 	Package         string         `json:",omitempty"` // Tailscale package to disambiguate ("choco", "appstore", etc; "" for unknown)
 	DeviceModel     string         `json:",omitempty"` // mobile phone model ("Pixel 3a", "iPhone12,3")
+	DeviceToken     string         `json:",omitempty"` // macOS/iOS APNs device token for notifications
```
bradfitz, line 528: could be android too in the future

bradfitz (commented):
Seems fine overall

ipn/localapi/localapi.go
crawshaw: just the field or also the json parameter name? we seem to have both kinds of parameter names?

bradfitz: You can't json.Unmarshal (due to `reflect` rules) into a field that's not exported. JSON parameter name doesn't matter.

But really this type shouldn't be defined here like this. It should be in a package where the client side can use it, even if the sole client currently in Swift that doesn't need it.

crawshaw: Yeah I changed it last minute after testing all the code.

Does the localapi have tests I should be extending for a new endpoint?

What package would the type go in?

crawshaw: (the question about the parameter name was a style question, I've no opinion other than I couldn't figure out what the standard was here.)

bradfitz:
```go
// Package apitype contains types for the Tailscale LocalAPI and control plane API.
package apitype
```

client/tailscale/apitype/apitype.go
```diff
@@ -32,3 +32,9 @@ type WaitingFile struct {
 	Name string
 	Size int64
 }
+
+// SetDeviceToken is the body POSTed to /set-device-token.
```
bradfitz, line 36: SetDeviceTokenRequest ?

mention LocalAPI in the docs too, as this package is a weird mix.

client/tailscale/apitype/apitype.go
```diff
@@ -32,3 +32,9 @@ type WaitingFile struct {
...
+// It contains the iOS/macOS APNs device token (and any future Android equivalent).
```
bradfitz, line 37: move this line down a line to be a doc on the field? "DeviceToken is ...."

hostinfo/hostinfo.go
```diff
@@ -153,12 +154,16 @@ func GetEnvType() EnvType {
 }
 
 var (
+	deviceTokenAtomic atomic.Value // of string (base64)
```
bradfitz, line 157: what's this "base64" doing here? it's just a string, no?

base64 is a detail for another layer?

tailcfg/tailcfg.go
```diff
@@ -525,6 +525,7 @@ type Hostinfo struct {
 	Desktop         opt.Bool       `json:",omitempty"` // if a desktop was detected on Linux
 	Package         string         `json:",omitempty"` // Tailscale package to disambiguate ("choco", "appstore", etc; "" for unknown)
 	DeviceModel     string         `json:",omitempty"` // mobile phone model ("Pixel 3a", "iPhone12,3")
+	DeviceToken     string         `json:",omitempty"` // macOS/iOS APNs device token for notifications (and maybe Android in the future)
```
bradfitz, line 528: should we call this NotifyToken? MobileNotifyToken?

bradfitz (approved):

tailcfg/tailcfg.go
crawshaw: it's called DeviceToken in APNs. I'm not super keen on inventing a new name.

crawshaw:
I wrote a test!

...which required fixing a bug in the host validation logic. 💀  PTAL

tailcfg/tailcfg.go
bradfitz: Well I was thinking that if we use it for Android too it'll be wrong for one platform at least so we should make up a common term for both.

But it seems like Android also uses "Device Token" so sure! Good as is.

ipn/localapi/localapi.go
```diff
@@ -202,7 +203,7 @@ func validHost(h string) bool {
 	if host == "localhost" {
 		return true
 	}
-	addr, err := netip.ParseAddr(h)
+	addr, err := netip.ParseAddr(host)
```
bradfitz, line 206: whoa, what? this should be its own commit

crawshaw: no good deed, etc etc

crawshaw: https://github.com/tailscale/tailscale/pull/7114

crawshaw:
PTAL

tailcfg/tailcfg.go
mihaip: Still think that having something about "push" or "notification" in the name would be useful, since it's next to a bunch of other generic fields (`DeviceModel`, `Package`) and it's not clear what kind of token we're talking about.

crawshaw: ok done!

ipn/localapi/localapi.go
```diff
@@ -1198,6 +1199,26 @@ func (h *Handler) serveDial(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	var params apitype.SetPushDeviceTokenRequest
+	if err := json.NewDecoder(r.Body).Decode(&params); err != nil {
+		http.Error(w, "invalid JSON body", 400)
+		return
+	}
+	if params.PushDeviceToken != "" {
+		hostinfo.SetPushDeviceToken(params.PushDeviceToken)
+	}
```
bradfitz, line 1218: don't want to support any sort of clearing operation here?

tailcfg/tailcfg.go
```diff
@@ -525,6 +525,7 @@ type Hostinfo struct {
 	Desktop         opt.Bool       `json:",omitempty"` // if a desktop was detected on Linux
 	Package         string         `json:",omitempty"` // Tailscale package to disambiguate ("choco", "appstore", etc; "" for unknown)
 	DeviceModel     string         `json:",omitempty"` // mobile phone model ("Pixel 3a", "iPhone12,3")
+	PushDeviceToken string         `json:",omitempty"` // macOS/iOS APNs device token for notifications (and maybe Android in the future)
```
bradfitz, line 528: s/maybe //

bradfitz (approved):

mihaip (commented):
New name LGTM

Out of curiosity, do we have any kind of mechanism to force the controlclient to send an updated hostinfo when the token is set?

bradfitz:
Yeah we do. This should be connected to that. Can look later when off phone.

crawshaw:
I couldn't find an obvious thing for pushing a hostinfo update so I added one. PTAL?

ipn/ipnlocal/local.go
```diff
@@ -1811,6 +1811,21 @@ func (b *LocalBackend) readPoller() {
 	}
 }
 
+// HostinfoUpdated is called to recompute the Hostinfo and send
+// the new version to the control server.
+func (b *LocalBackend) HostinfoUpdated() {
```
bradfitz, line 1816: this name sounds like a accessor that returns a bool

make it sound imperative? `ResendHostinfoIfNeeded` maybe? Otherwise LGTM.

Wow, I forgot how horrible that `doSetHostinfoFilterServices` is. TODOs about in that call chain. (cc @andrew-d)

bradfitz (approved):

ipn/ipnlocal/local.go
crawshaw: done.

some of the hostinfo call chain looks like it wouldn't be too hard to clean up.

PR #7114, opened 2023-01-30
ipn/localapi: fix validHost parsing logic

crawshaw:
It was more restricted than it was intended to be. But now it does what it says it does: is this what we want?

Added a test.


ipn/localapi/localapi_test.go
```diff
@@ -0,0 +1,42 @@
...
+		if !validHost(h) {
+			t.Errorf("validHost(%q)=false, want true", h)
+		}
+	}
+
+	invalid := []string{
+		"100.100.100.100:41112",
+		"10.0.0.1:41112",
+		"37.16.9.210:41112",
+	}
```
bradfitz, line 35: very JS style style test with two sections of good and bad :)

ipn/localapi/localapi_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"tailscale.com/client/tailscale/apitype"
+)
+
+func TestValidHost(t *testing.T) {
+	valid := []string{
+		"",
+		apitype.LocalAPIHost,
+		"localhost:41112", // port is the safesocket port on Windows
+		"127.0.0.1:41112",
+		"[::1]:41112",
```
bradfitz, line 19: But we don't use localhost TCP on Windows anymore. Do we even want to allow these now? cc @maisem

ipn/localapi/localapi_test.go
crawshaw: I thought about defining types to parameterize the boolean but... is that really more readable? (I'm honestly not sure? I think not?) Parameterizing scales to more complex scenarios, but... tailscale.

bradfitz: It's pretty idiomatic Go and we do it a 190 other places in the tree:

```
$ git grep -F '[]struct {' | wc -l
190
```

twitchyliquid64: +1 to table-driven

```go
for tc in ([]struct{host string, err error}{ .. }) {
  t.Run(tc.host, func(t *testing.T){ 
    ...
  })
}
```

ipn/localapi/localapi_test.go
maisem: yeah we don't need these anymore. Sent https://github.com/tailscale/tailscale/pull/7115 for cleanup

crawshaw: so, on windows the validHost logic should be changed to reject anything other than the socket.

but we continue to use TCP elsewhere? this function is otherwise correct?

crawshaw: i'll update this pr once https://github.com/tailscale/tailscale/pull/7115 is in

crawshaw: PTAL

ipn/localapi/localapi.go
```diff
@@ -179,22 +179,28 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// validLocalHost allows either localhost or loopback IP hosts on platforms
+// that use token security.
+var validLocalHost = runtime.GOOS == "darwin" || runtime.GOOS == "android"
```
bradfitz, line 184: @mihaip, do we need `"ios"` too?

Actually, I don't think any of those three platforms use the ipnserver code, so is this code even exercised? This whole const seems misleading then.

mihaip: Yeah, we'd need `ios`. Both it and macOS/`darwin` end up making requests to `localhost` (https://github.com/tailscale/corp/blob/24f5a2d230a0d27b9665e0d5c33eeabbcdd11dbf/xcode/Shared/LocalAPIClient.swift#L320), they should be hitting this code path.

bradfitz: Yeah, okay, I see the `localapi.NewHandler` call site on the macOS/iOS side now. I was thinking this check was only for ipnserver but that's wrong... it's just checking the http.Request.Host which'll still be checked for macOS/iOS.

crawshaw: added ios

bradfitz (approved):

PR #7115, opened 2023-01-30
safesocket: remove the now unused WindowsLocalPort

maisem:
Also drop the port param from safesocket.Listen. #cleanup


crawshaw (approved):

bradfitz (approved):
yay

PR #7120, opened 2023-01-30
util/vizerror: add new package for visible errors

willnorris:


willnorris:
(missed testing it with tsweb, so moving back to draft for now.  But go ahead and see if this is what you had in mind)

util/vizerror/vizerror.go
```diff
@@ -0,0 +1,24 @@
...
+
+// Error is an error that is safe to display to end users.
+type Error error
+
+// New returns an error that formats as the given text.
+func New(text string) error {
+	return Error(errors.New(text))
+}
+
+// Errorf returns an error with the specified format and values.
```
bradfitz, line 21: Maybe on both of these document that the returned type is guaranteed to be of type Error?

util/vizerror/vizerror.go
```diff
@@ -0,0 +1,24 @@
...
+func Errorf(format string, a ...any) error {
+	return Error(fmt.Errorf(format, a...))
+}
```
bradfitz, line 24: maybe add the `func FromError(err error) (_ Error, ok bool)` too that does the errors.As for caller?

maybe `FromError` is too long. Maybe `Get` or `As` or `From`? :shrug:

willnorris: maybe I just haven't used `errors.As` enough, but why would we use that over simply casting:`vizerror.Error(existingErr)` ?

bradfitz: You mean type assert and not cast. But it doesn't work in any case: https://go.dev/play/p/fPJ_y_F_0q1

You don't want the error as the underlying type. Do `type Error struct { error }`

But you can't just type assert because of wrapping. If somebody does `fmt.Errorf("bad policy: %w", err)` you want to be able to get the user visible error out of that %w-wrapped inner `err` too.

bradfitz: See https://go.dev/blog/go1.13-errors

willnorris: I added a simple `Wrap(err) Error` func.  But I'm not if that's what you were getting and, since I'm not sure what you mean by doing the errors.As for the caller.  Isn't that going the other direction... unwrapping and getting the `vizerror.Error` back out?

util/vizerror/vizerror.go
willnorris: Any reason not to just return an `Error` to make that really explicit?  I know it's recommended to return concrete types from functions rather than interfaces, but I'm not sure if errors are treated differently.

willnorris:
Okay, see if this is closer to what you were imagining.

util/vizerror/vizerror.go
bradfitz: Because that's the convention (e.g. look at all the errors returned by the `os` package) Otherwise people fall into https://go.dev/doc/faq#nil_error ... 

Or do stuff like:

```go
     err := vizerror.New("foo")
     if thingsGood {
          err = nil
      }
```

And are surprised it doesn't compile. Or the type is a pointer, find it compiles but hit that FAQ problem

willnorris: done

tsweb/tsweb_test.go
```diff
@@ -151,6 +152,23 @@ func TestStdHandler(t *testing.T) {
 			},
 		},
 
+		{
+			name:     "handler returns user visible error",
+			rh:       handlerErr(0, vizerror.New("visible error")),
```
bradfitz, line 157: add another test case with `fmt.Errorf("some internal error oh but user can see: %w", vizerror.New("bad field foo"))`

util/vizerror/vizerror.go
```diff
@@ -0,0 +1,44 @@
...
+func Errorf(format string, a ...any) error {
+	return Error{fmt.Errorf(format, a...)}
+}
+
+// Unwrap returns the underlying error.
+func (e Error) Unwrap() error {
+	return e.err
+}
+
+// Wrap err with a vizerror.Error.
```
bradfitz, line 38: wrong verb tense for go docs

(this is imperative)

util/vizerror/vizerror.go
```diff
@@ -0,0 +1,44 @@
...
+type Error struct {
+	err error
+}
+
+// Error implements the error interface.
+func (e Error) Error() string {
+	return e.err.Error()
+}
+
+// New returns an error that formats as the given text. Always returns a vizerror.Error.
```
bradfitz, line 23: s/Always/\nIt always/g

tsweb/tsweb_test.go
willnorris: done

util/vizerror/vizerror.go
willnorris: done

util/vizerror/vizerror.go
willnorris: done

bradfitz (approved):

bradfitz:
Both CI failures look real.

willnorris:
oh, yep.  fixed.

PR #7121, opened 2023-01-30
cmd/tsconnect: use empty string as the default state store key

mihaip:
Makes the Wasm client more similar to the others, and allows the default profile to be correctly picked up when restarting the client in dev mode (where we persist the state in sessionStorage).

Also update README to reflect that Go Wasm changes can be picked up with just a reload (as of #5383).

### DIFF
--- a/cmd/tsconnect/wasm/wasm_js.go
+++ b/cmd/tsconnect/wasm/wasm_js.go
@@ -124,7 +124,7 @@ func newIPN(jsConfig js.Value) map[string]any {
 
 	logid := lpc.PublicID.String()
 	srv := ipnserver.New(logf, logid)
-	lb, err := ipnlocal.NewLocalBackend(logf, logid, store, "wasm", dialer, eng, controlclient.LoginEphemeral)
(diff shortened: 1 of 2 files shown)


cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -124,7 +124,7 @@ func newIPN(jsConfig js.Value) map[string]any {
 
 	logid := lpc.PublicID.String()
 	srv := ipnserver.New(logf, logid)
-	lb, err := ipnlocal.NewLocalBackend(logf, logid, store, "wasm", dialer, eng, controlclient.LoginEphemeral)
+	lb, err := ipnlocal.NewLocalBackend(logf, logid, store, "", dialer, eng, controlclient.LoginEphemeral)
```
maisem, line 127: I think this might be the last user of this field, we can probably just drop it from `NewLocalBackend`

maisem (approved):

cmd/tsconnect/wasm/wasm_js.go
mihaip: Good point, added to my Fixit Friday cleanup list.

PR #7127, opened 2023-01-31
scripts: use pkg server to determine supported deb/rpm distros

knyar:
Fixes https://github.com/tailscale/corp/issues/8952

### DIFF
--- a/scripts/installer.sh
+++ b/scripts/installer.sh
@@ -293,83 +293,28 @@ main() {
 		fi
 	fi
 
+	# Ideally we want to use curl, but on some installs we
+	# only have wget. Detect and use what's available.
+	CURL=
+	if type curl >/dev/null; then
+		CURL="curl -fsSL"
+	elif type wget >/dev/null; then
+		CURL="wget -q -O-"
+	fi
+	if [ -z "$CURL" ]; then
+		echo "The installer needs either curl or wget to download files."
+		echo "Please install either curl or wget to proceed."
+		exit 1
+	fi
+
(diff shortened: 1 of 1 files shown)


scripts/installer.sh
```diff
@@ -7,6 +7,23 @@
...
+	# Ideally we want to use curl, but on some installs we
+	# only have wget. Detect and use what's available.
+	CURL=
+	if type curl >/dev/null; then
+		CURL="curl -fsSL"
+	elif type wget >/dev/null; then
+		CURL="wget -q -O-"
+	fi
+	if [ -z "$CURL" ]; then
+		echo "The installer needs either curl or wget to download files."
```
danderson, line 20: Now that this error path is in a function, I think `exit 1` will just exit from the function, and dump the error message into whatever do_curl is being piped into (apt-key, sudo tee, ...).

I think this should stay as top-level code in main, rather than be a helper func. Then exit will work normally, and the rest of the code can use $CURL.

knyar: Oh thanks, good point – I was thinking about redirecting these errors to stderr but forgot to actually do it.

The reason for moving this into a separate function was to only require `curl` or `wget` when they are needed (on rpm & apr distros). Having this in `main` (which is what I've done now) would require these dependencies even if the script does not use them (e.g. on Gentoo, Arch, FreeBSD). I suspect that's fine – realistically most users will use `curl` to download the script in the first place. But I am also happy to make this a bit more complex and only require `curl` or `wget` when we are using them.

danderson (approved):

PR #7128, opened 2023-01-31
wgengine/magicsock: add envknob to print information about port selection

andrew-d:
To aid in debugging where a customer has static port-forwards set up and there are issues establishing a connection through that port.

Change-Id: Ic5558bcdb40c9119b83f79dcacf2233b07777f2a

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -3010,6 +3010,8 @@ func (c *Conn) listenPacket(network string, port uint16) (nettype.PacketConn, er
 	return nettype.MakePacketListenerWithNetIP(netns.Listener(c.logf)).ListenPacket(ctx, network, addr)
 }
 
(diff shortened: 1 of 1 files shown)


wgengine/magicsock/magicsock.go
```diff
@@ -3027,7 +3033,7 @@ func (c *Conn) bindSocket(ruc *RebindingUDPConn, network string, curPortFate cur
 		return nil
 	}
 
-	if debugAlwaysDERP() {
+	if debugAlwaysDERP() || debugBindSocket() {
```
bradfitz, line 3036: this seems like it's changing behavior rather than just adding debug prints

andrew-d: Blergh, good catch; fixed.

bradfitz (approved):

PR #7129, opened 2023-01-31
ipn/ipnlocal: rate-limit diagnose logs in bugreport

andrew-d:
We can log too quickly for logtail to catch up, even when we opt out of log rate-limiting. When the user passes the `--diagnose` flag to bugreport, we use a token bucket to control how many logs per second are printed and sleep until we're able to write more.

Change-Id: If27672d66b621b589280bd0fe228de367ffcbd8f

### DIFF
--- a/types/logger/logger.go
+++ b/types/logger/logger.go
@@ -228,6 +228,53 @@ func RateLimitedFnWithClock(logf Logf, f time.Duration, burst int, maxCache int,
 	}
 }
 
+// SlowLoggerWithClock is a logger that applies rate limits similar to
+// RateLimitedFnWithClock, but instead of dropping logs will sleep until they
+// can be written. This should only be used for debug logs, and not in a hot path.
+//
+// The provided context, if canceled, will cause all logs to be dropped and
+// prevent any sleeps.
+func SlowLoggerWithClock(ctx context.Context, logf Logf, f time.Duration, burst int, timeNow func() time.Time) Logf {
+	var (
+		mu sync.Mutex
+		tb = newTokenBucket(f, burst, timeNow())
+	)
+	return func(format string, args ...any) {
+		if ctx.Err() != nil {
+			return
+		}
+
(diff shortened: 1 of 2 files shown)


types/logger/logger.go
```diff
@@ -228,6 +228,51 @@ func RateLimitedFnWithClock(logf Logf, f time.Duration, burst int, maxCache int,
...
+//
+// The provided context, if canceled, will cause all logs to be dropped and
+// prevent any sleeps.
+func SlowLoggerWithClock(ctx context.Context, logf Logf, f time.Duration, burst int, timeNow func() time.Time) Logf {
+	var (
+		mu sync.Mutex
+		tb = newTokenBucket(f, burst, timeNow())
+	)
+	return func(format string, args ...any) {
+		if err := ctx.Err(); err != nil {
```
dsnet, line 243: Simpler as just `ctx.Err() != nil`?

types/logger/logger.go
```diff
@@ -228,6 +228,51 @@ func RateLimitedFnWithClock(logf Logf, f time.Duration, burst int, maxCache int,
...
+		// If we can get a token, then do that and return.
+		if tb.Get() {
+			logf(format, args...)
+			return
+		}
+
+		// Otherwise, sleep for 2x the duration so that we don't
+		// immediately sleep again on the next call.
+		select {
+		case curr := <-time.After(f):
```
dsnet, line 263: `2*f` to match the comment?

In the event that `ctx.Done` triggers, the underlying timer can't be GCd until it triggers. Do we care? I assume this is only for debugging so the presumably short-lived timer won't will eat much resources.

dsnet (approved):

types/logger/logger.go
andrew-d: Good catch, updated 🤦 

It's not a huge issue, but I switched to `time.NewTimer` anyway since it's not any more complex.

Issue #6468, opened 2022-11-23
FR: Support for funnel in `containerboot`


garrett-ts:
### What are you trying to do?

This request originated from a ticket (#7128)

We should consider updating cmd/containerboot to support the funnel feature. It may be valuable to add supporting help documentation as well.

danderson:
Rough outline: I believe Funnel can be programmatically activated through localAPI, which containerboot already wields to achieve other things. We'd want yet another envvar to allow operators to express that the container should funnel itself, as well as additional logic to enforce valid flag combos (e.g. maybe don't allow funnel combined with ephemeral storage, since that'll lead to spamming LE with cert requests?).

Before doing that we should also change containerboot to use the new IPN bus subcribe mechanism that @bradfitz just shipped, so we can stop shelling out to `tailscale up ` and get tighter coordination with the daemon.

DentonGentry:
https://github.com/tailscale/tailscale/pull/7136 is a work in progress toward this.

PR #7136, opened 2023-02-01
cmd/containerboot: add support for setting funnel TCP portforward

maisem:
WIP

### DIFF
--- a/cmd/containerboot/main.go
+++ b/cmd/containerboot/main.go
@@ -90,8 +90,31 @@ func main() {
 		AuthOnce:        defaultBool("TS_AUTH_ONCE", false),
 		Root:            defaultEnv("TS_TEST_ONLY_ROOT", "/"),
 	}
+	funnelForwardPorts := strings.Split(defaultEnv("TS_FUNNEL_TCP_PORTFORWARD", ""), ",")
+	if len(funnelForwardPorts) > 0 {
+		ffp := make(map[uint16]uint16)
+		for _, p := range funnelForwardPorts {
+			if p == "" {
+				continue
+			}
+			from, to, ok := strings.Cut(p, ":")
+			if !ok {
+				log.Fatalf("TS_FUNNEL_TCP_PORTFORWARD: %q is not a valid port pair", p)
+			}
+			fp, err := strconv.ParseUint(from, 10, 16)
+			if err != nil {
+				log.Fatalf("TS_FUNNEL_TCP_PORTFORWARD: %v", err)
+			}
+			tp, err := strconv.ParseUint(to, 10, 16)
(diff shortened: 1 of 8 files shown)


cmd/containerboot/main.go
```diff
@@ -90,8 +90,31 @@ func main() {
 		AuthOnce:        defaultBool("TS_AUTH_ONCE", false),
 		Root:            defaultEnv("TS_TEST_ONLY_ROOT", "/"),
 	}
+	funnelForwardPorts := strings.Split(defaultEnv("TS_FUNNEL_TCP_PORTFORWARD", ""), ",")
```
danderson, line 93: I'd probably call this just "TS_FUNNEL_TCP", to reduce typing. Also add the envvar to the big comment at the top that lists all the available options.

cmd/containerboot/main.go
```diff
@@ -305,6 +324,35 @@ authLoop:
...
+		TCP:         make(map[uint16]*ipn.TCPPortHandler),
+	}
+	for f, t := range cfg.FunnelTCPPorts {
+		sc.TCP[f] = &ipn.TCPPortHandler{
+			TCPForward:   fmt.Sprintf("%s:%d", cfg.ProxyTo, t),
+			TerminateTLS: cd,
+		}
+		sc.AllowFunnel[ipn.HostPort(fmt.Sprintf("%s:%d", cd, f))] = true
+	}
+	return client.SetServeConfig(ctx, sc)
```
danderson, line 353: How do we know when funnel is working? Ideally the operator would like some signal that pod startup has completed correctly, so it doesn't advertise the ingress hostname until it's actually serving.

cmd/containerboot/main.go
```diff
@@ -305,6 +324,35 @@ authLoop:
 	}
 }
 
+func configureForwarding(ctx context.Context, client *tailscale.LocalClient, cfg *settings) error {
+	if cfg.ProxyTo == "" {
```
danderson, line 328: I don't think this is right for all scenarios, it's valid to configure funnel without ProxyTo if you're running the container as a sidecar. I think funnel needs to work in either case?

danderson (commented):
I don't think I like this as-is, especially the various "if kubernetes" branches in ipnlocal. We already have a bunch of these branches throughout our code and the magically different behavior makes it hard to maintain stuff :(

Maybe it's time to graduate the store API from plain key/value to structured Get/Set for specific data, so that the underlying implementations cna handle all this weird differential storage we're doing.

DentonGentry:
Recommend the commit message state that it Updates https://github.com/tailscale/tailscale/issues/6468
(or Fixes, as the case may be)

PR #7142, opened 2023-02-01
ipn/ipnlocal: add support to store certs in k8s secrets

### LINKED ISSUE #5676: FR: Cache tailscale cert and key data to kube state secret

- When obtaining a cert via socket integration (E.g. Caddy) certs are stored in the state dir. Based on the default run.sh this is /tmp.
- When the pod is recycled the state dir is emptied and cert is requested again.
- If this happens too often it hits the LetsEncrypt limit for that name.

### How should we solve this?

- It would be great if the certificate and key data was cached in the kube state/secret similar to the node key so it can write the latest cert data here and check and use this before it goes off and requests again.

- Having to change the machine name.
- Creating a custom caching mechanism outside of tailscaled.

### Anything else?

_No response_

maisem:
Fixes #5676

### DIFF
--- a/ipn/ipnlocal/cert.go
+++ b/ipn/ipnlocal/cert.go
@@ -32,6 +32,8 @@ import (
 
 	"golang.org/x/crypto/acme"
 	"tailscale.com/envknob"
+	"tailscale.com/hostinfo"
+	"tailscale.com/ipn"
 	"tailscale.com/ipn/ipnstate"
 	"tailscale.com/types/logger"
 	"tailscale.com/version"
@@ -94,9 +96,9 @@ func (b *LocalBackend) GetCertPEM(ctx context.Context, domain string) (*TLSCertK
 		log.Printf("acme %T: %s", v, j)
 	}
 
-	if pair, ok := getCertPEMCached(dir, domain, now); ok {
+	if pair, err := b.getCertPEMCached(dir, domain, now); err == nil {
 		future := now.AddDate(0, 0, 14)
-		if shouldStartDomainRenewal(dir, domain, future) {
+		if b.shouldStartDomainRenewal(dir, domain, future) {
 			logf("starting async renewal")
 			// Start renewal in the background.
 			go b.getCertPEM(context.Background(), logf, traceACME, dir, domain, future)
@@ -112,7 +114,7 @@ func (b *LocalBackend) GetCertPEM(ctx context.Context, domain string) (*TLSCertK
 	return pair, nil
 }
 
-func shouldStartDomainRenewal(dir, domain string, future time.Time) bool {
+func (b *LocalBackend) shouldStartDomainRenewal(dir, domain string, future time.Time) bool {
 	renewMu.Lock()
 	defer renewMu.Unlock()
 	now := time.Now()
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/cert.go
```diff
@@ -122,10 +124,78 @@ func shouldStartDomainRenewal(dir, domain string, future time.Time) bool {
...
+	WriteCert(domain string, cert []byte) error
+	WriteKey(domain string, key []byte) error
+}
+
+func (b *LocalBackend) getCertStore(dir string) certStore {
+	if hostinfo.GetEnvType() == hostinfo.Kubernetes {
+		return certStateStore{b.store}
+	}
+	return certFileStore(dir)
+}
```
bradfitz, line 145: Not a big fan of this. Too magic and limiting... now people can't store certs on disk (some persistent volume?) like they could've before if so configured.

I'd rather have the wiring be explicit.

maisem: We ignore the `StateDir` if a `KubeSecret` is specified. Added a check to make sure the underlying store is k8s secrets.
```
	args := []string{"--socket=" + cfg.Socket}
	switch {
	case cfg.InKubernetes && cfg.KubeSecret != "":
		args = append(args, "--state=kube:"+cfg.KubeSecret, "--statedir=/tmp")
	case cfg.StateDir != "":
		args = append(args, "--statedir="+cfg.StateDir)
	default:
		args = append(args, "--state=mem:", "--statedir=/tmp")
	}
```

maisem: would you prefer an envknob? or another flag to tailscaled?

bradfitz: I also don't think we should do this:

```go
	case cfg.InKubernetes && cfg.KubeSecret != "":
		args = append(args, "--state=kube:"+cfg.KubeSecret, "--statedir=/tmp")
```

... if StateDir is set to something else. We shouldn't magically overwrite/ignore their old value.

A default is fine.

maisem: Fixed the StateDir handling in containerboot and added another check to use the kubestore only if the statedir is `/tmp`

ipn/ipnlocal/cert.go
```diff
@@ -122,10 +125,80 @@ func shouldStartDomainRenewal(dir, domain string, future time.Time) bool {
...
-	_, ok := getCertPEMCached(dir, domain, future)
+	_, ok := b.getCertPEMCached(dir, domain, future)
 	return !ok
 }
 
+// certStore provides a way to perist and retrieve TLS certificates.
+// As of 2023-02-01, we use store certs in directories on disk everywhere
+// except on Kubernetes, where we use the state store.
+type certStore interface {
+	Read(domain string, now time.Time) (*TLSCertKeyPair, error)
```
bradfitz, line 136: is the error anything special? any particular types? is it an error if now is greater than the cert's expiry? does this auto-fetch a new one?

looks like caller ignores it entirely and just treated it as a boolean?

ipn/ipnlocal/cert.go
```diff
@@ -122,10 +125,80 @@ func shouldStartDomainRenewal(dir, domain string, future time.Time) bool {
...
+func (b *LocalBackend) getCertStore(dir string) certStore {
+	if hostinfo.GetEnvType() == hostinfo.Kubernetes {
+		if _, ok := b.store.(*kubestore.Store); ok && dir == "/tmp" {
+			return certStateStore{b.store}
+		}
+	}
+	return certFileStore(dir)
+}
+
+type certFileStore string // dir
```
bradfitz, line 150: // certFileStore implements certStore by storing the cert & key files in the named directory.

ipn/ipnlocal/cert.go
```diff
@@ -122,10 +125,80 @@ func shouldStartDomainRenewal(dir, domain string, future time.Time) bool {
...
+
+func (f certFileStore) WriteCert(domain string, cert []byte) error {
+	return os.WriteFile(keyFile(string(f), domain), cert, 0644)
+}
+
+func (f certFileStore) WriteKey(domain string, key []byte) error {
+	return os.WriteFile(keyFile(string(f), domain), key, 0600)
+}
+
+type certStateStore struct {
```
bradfitz, line 175: // certStateStore implements certStore by storing the cert & key files in an ipn.StateStore.

ipn/ipnlocal/cert.go
bradfitz: shouldn't we at least define what the error for "Not found" is to prevent log spam later? currently the two impls have two different errors.

ipn/ipnlocal/cert.go
maisem: done

ipn/ipnlocal/cert.go
maisem: made the errors be consistent, and plumbed it through the caller.

ipn/ipnlocal/cert.go
maisem: done

ipn/ipnlocal/cert.go
maisem: PTAL

bradfitz (approved):

PR #7143, opened 2023-02-01
Allow the expiry time to be specified for new keys

### LINKED ISSUE #7965: FR: Allow Key Duration To Be Passed As Parameter in Tailscale Client

Provide a duration until expiry for a new key generated through the API.

### How should we solve this?

We should add this as a parameter to the `CreateKey` function.

_No response_

### Anything else?

I've already created a PR for this issue: https://github.com/tailscale/tailscale/pull/7143

matty234:
Adds a parameter for create key that allows a number of seconds (less than 90) to be specified for new keys. 

Closes #7965 

Signed-off-by: Matthew Brown <matthew@bargrove.com>

### DIFF
--- a/client/tailscale/keys.go
+++ b/client/tailscale/keys.go
@@ -70,12 +70,12 @@ func (c *Client) Keys(ctx context.Context) ([]string, error) {
 // CreateKey creates a new key for the current user. Currently, only auth keys
 // can be created. Returns the key itself, which cannot be retrieved again
 // later, and the key metadata.
-func (c *Client) CreateKey(ctx context.Context, caps KeyCapabilities, expirySeconds int64) (string, *Key, error) {
+func (c *Client) CreateKey(ctx context.Context, caps KeyCapabilities, expiry time.Duration) (string, *Key, error) {
 
-	ninetydays := int64(90 * 24 * 60 * 60)
-
-	if expirySeconds > ninetydays {
-		return "", nil, fmt.Errorf("expiry must be less than 90 days")
+	// convert expirySeconds to an int64 (seconds)
+	expirySeconds := int64(expiry.Seconds())
+	if expirySeconds < 0 {
+		return "", nil, fmt.Errorf("expiry must be positive")
 	}
 
 	keyRequest := struct {
--- a/client/tailscale/keys.go
+++ b/client/tailscale/keys.go
@@ -70,10 +70,18 @@ func (c *Client) Keys(ctx context.Context) ([]string, error) {
 // CreateKey creates a new key for the current user. Currently, only auth keys
 // can be created. Returns the key itself, which cannot be retrieved again
 // later, and the key metadata.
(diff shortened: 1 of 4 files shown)


client/tailscale/keys.go
```diff
@@ -71,10 +71,18 @@ func (c *Client) Keys(ctx context.Context) ([]string, error) {
 // CreateKey creates a new key for the current user. Currently, only auth keys
 // can be created. Returns the key itself, which cannot be retrieved again
 // later, and the key metadata.
-func (c *Client) CreateKey(ctx context.Context, caps KeyCapabilities) (string, *Key, error) {
+func (c *Client) CreateKey(ctx context.Context, caps KeyCapabilities, expirySeconds int64) (string, *Key, error) {
+
+	ninetydays := int64(90 * 24 * 60 * 60)
+
+	if expirySeconds > ninetydays {
```
DentonGentry, line 78: Recommend not checking the validity of the expiration parameter in the client: maybe someday support will be added for longer expirations. The API call will fail if the expiration is out of range.

client/tailscale/keys.go
```diff
@@ -71,10 +71,18 @@ func (c *Client) Keys(ctx context.Context) ([]string, error) {
 // CreateKey creates a new key for the current user. Currently, only auth keys
 // can be created. Returns the key itself, which cannot be retrieved again
 // later, and the key metadata.
-func (c *Client) CreateKey(ctx context.Context, caps KeyCapabilities) (string, *Key, error) {
+func (c *Client) CreateKey(ctx context.Context, caps KeyCapabilities, expirySeconds int64) (string, *Key, error) {
```
DentonGentry, line 74: Recommend that the argument be a time.Duration instead of an int64.

cmd/k8s-operator/operator.go
```diff
@@ -547,7 +549,8 @@ func (a *ServiceReconciler) newAuthKey(ctx context.Context, tags []string) (stri
 			},
 		},
 	}
-	key, _, err := a.tsClient.CreateKey(ctx, caps)
+	ninetyDays := 90 * 24 * time.Hour
```
DentonGentry, line 552: Recommend passing in zero, which will cause the `omitempty` of the keyRequest to leave ExpirySeconds unpopulated, which will inherit the server default.

DentonGentry:
Another caller has been added since this PR was first opened, which is causing the CI failure. Could you:
1. Rebase to HEAD
2. fix cmd/get-authkey/main.go to pass in a 0 for the expiration.

```
cmd/get-authkey/main.go:70:45: not enough arguments in call to tsClient.CreateKey
	have (context.Context, tailscale.KeyCapabilities)
	want (context.Context, tailscale.KeyCapabilities, time.Duration)
```

matty234:
@DentonGentry  - apologies for the delay here (and the commit mess) not sure how to make issuebot happy

Issue #3198, opened 2021-10-27
FR: Flag health errors for self-signed or otherwise invalid certificates


DentonGentry:
### Tell us about your idea!

Provide a health signal when connectivity is blocked by a TLS MITM self-signed certificate

### What are you trying to do?

We have had a few instances where connectivity has been interrupted due to a security appliance deciding to MITM tailscaled's connections to Control or DERP servers. The most recent incident was when a Fortigate config changed, and TLS connections to derp4 and log.tailscale.io failed because the certificate was self-signed.

A self-signed certificate is never right when speaking to production, we should flag that as a health check error.

<img src="https://frontapp.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_3wtht)

danderson:
Such a cert is likely to also lead to the control connection failing, so it's unclear to me how useful this'll be. We can certainly log it, but a health status seems likely to not propagate for the very reason we're logging about.

DentonGentry:
If we can build it in a way to signal to the user that they're offline because they are experiencing some kind of MITM, that would also be good.

In the case which prompted filing this, we were not getting control or logs connections. We got onto a videoconference to look at local logs, and they were full of complaints about a self-signed certificate that had "Fortigate" in its name.

mihaip:
My local library appears to have introduced ContentKeeper on its WiFi, which is attempting to do some kind of SSL MITM. Even with #7214 (running 1.37.135) I'm not able to have Tailscale successfully establish a connection with control, it fails with a `x509: certificate signed by unknown authority` error. Bug report at BUG-2170128b2685f0d55872528a2732875013d69c81f7eaaeb98554137d4adf722e-20230218221708Z-c941b8e967746073

```
$ /Applications/Tailscale.app/Contents/MacOS/Tailscale status
# Health check:
#     - not logged in, last login error=fetch control key: Get "https://controlplane.tailscale.com/key?v=57": x509: certificate signed by unknown authority
#     - This is an unstable (development) version of Tailscale; frequent updates and bugs are likely

unexpected state: NoState
```

This is what Chrome sees as the cert for controlplane.tailscale.com:

<img width="604" alt="image" src="https://user-images.githubusercontent.com/513813/219902141-a6750c37-aba6-4aaf-a867-2e7ef5bdcbfb.png">

andrew-d:
Looks like #7149 works correctly in this case:
[...]

PR #7149, opened 2023-02-01
health, net/tlsdial: add healthcheck for self-signed cert

andrew-d:
When we make a connection to a server, we previously would verify with the system roots, and then fall back to verifying with our baked-in Let's Encrypt root if the system root cert verification failed.

We now explicitly check for, and log a health error on, self-signed certificates. Additionally, we now always verify against our baked-in Let's Encrypt root certificate and log an error if that isn't successful. We don't consider this a health failure, since if we ever change our server certificate issuer in the future older non-updated versions of Tailscale will no longer be healthy despite being able to connect.

Updates #3198

Change-Id: I00be5ceb8afee544ee795e3c7a2815476abc4abf


health/health.go
```diff
@@ -476,6 +489,9 @@ func overallErrorLocked() error {
 	if err := envknob.ApplyDiskConfigError(); err != nil {
 		errs = append(errs, err)
 	}
+	for serverName, err := range tlsConnectionErrors {
+		errs = append(errs, fmt.Errorf("TLS connection error for %q: %v", serverName, err))
```
bradfitz, line 493: %w instead of %v?

net/tlsdial/tlsdial.go
```diff
@@ -99,6 +115,12 @@ func Config(host string, base *tls.Config) *tls.Config {
 	return conf
 }
 
+func certIsSelfSigned(cert *x509.Certificate) bool {
```
bradfitz, line 118: oh, this is tighter than I imagined it'd be at first reading of the call site.

or does this also cover signatures with internal corp CAs?

net/tlsdial/tlsdial.go
```diff
@@ -79,18 +92,21 @@ func Config(host string, base *tls.Config) *tls.Config {
...
+		// so we can log an informational message. This is useful for
+		// detecting SSL MiTM.
 		opts.Roots = bakedInRoots()
-		_, err := cs.PeerCertificates[0].Verify(opts)
+		_, bakedErr := cs.PeerCertificates[0].Verify(opts)
 		if debug() {
-			log.Printf("tlsdial(bake %q): %v", host, err)
+			log.Printf("tlsdial(bake %q): %v", host, bakedErr)
+		} else if bakedErr != nil {
+			log.Printf("tlsdial: server cert for %q is not a Let's Encrypt cert", host)
```
bradfitz, line 104: this'll be spammy for e.g. all the headscale users in China or anybody running a custom derper with a non-LE cert

andrew-d: Good point; let me rate-limit this.

net/tlsdial/tlsdial.go
andrew-d: It doesn't, for now; as a follow-up I want to import all public root CAs and verify against that set so we can detect non-public CA certs.

net/tlsdial/tlsdial.go
andrew-d: We now log only once per `host` per process invocation; PTAL? I could switch this to "every hour/day/etc." instead if you'd rather.

net/tlsdial/tlsdial.go
```diff
@@ -79,18 +97,23 @@ func Config(host string, base *tls.Config) *tls.Config {
...
+		// detecting SSL MiTM.
 		opts.Roots = bakedInRoots()
-		_, err := cs.PeerCertificates[0].Verify(opts)
+		_, bakedErr := cs.PeerCertificates[0].Verify(opts)
 		if debug() {
-			log.Printf("tlsdial(bake %q): %v", host, err)
+			log.Printf("tlsdial(bake %q): %v", host, bakedErr)
+		} else if bakedErr != nil {
+			if _, loaded := tlsdialWarningPrinted.LoadOrStore(host, true); !loaded {
+				log.Printf("tlsdial: server cert for %q is not a Let's Encrypt cert", host)
```
bradfitz, line 110: get the word "error" or "warn" in this string for grepping ease?

bradfitz: I guess it's an error if errSys is non-nil and warn otherwise?

net/tlsdial/tlsdial.go
```diff
@@ -33,6 +36,11 @@ var sslKeyLogFile = os.Getenv("SSLKEYLOGFILE")
 
 var debug = envknob.RegisterBool("TS_DEBUG_TLS_DIAL")
 
+// tlsdialWarningPrinted tracks whether we've printed a warning about a given
+// hostname already, to avoid log spam for users with custom DERP servers,
+// Headscale, etc.
+var tlsdialWarningPrinted sync.Map
```
bradfitz, line 42: for untyped collections, always document the keys + value types in a little same-line comment

bradfitz (approved):

net/tlsdial/tlsdial.go
andrew-d: Added the word `warning` or `error` depending on that case.

PR #7153, opened 2023-02-01
util/vizerror: add As function to get wrapped Error

willnorris:
okay, NOW I get what you were suggesting [in this comment](https://github.com/tailscale/tailscale/pull/7120#discussion_r1091180336).

### DIFF
--- a/util/vizerror/vizerror.go
+++ b/util/vizerror/vizerror.go
@@ -42,3 +42,9 @@ func Wrap(err error) error {
 	}
 	return Error{err}
 }
+
+// As returns the first vizerror.Error in err's chain.
+func As(err error) (e Error, ok bool) {
+	ok = errors.As(err, &e)
+	return
+}
--- a/tsweb/tsweb.go
+++ b/tsweb/tsweb.go
@@ -266,11 +266,10 @@ func (h retHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	err := h.rh.ServeHTTPReturn(lw, r)
 
 	var hErr HTTPError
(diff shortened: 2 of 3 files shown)


util/vizerror/vizerror.go
```diff
@@ -42,3 +42,9 @@ func Wrap(err error) error {
 	}
 	return Error{err}
 }
+
+// As returns the first vizerror.Error in err's chain.
+func As(err error) (_ Error, ok bool) {
+	var e Error
+	return e, errors.As(err, &e)
```
bradfitz, line 49: I don't immediately understand why this works. My intuition from the spec was that this shouldn't work, so I'm surprised. Or maybe this is undefined. I need to go read the spec more.

I thought the return statement assigned all its results in order, so it effectively compiled to:

```
      res[0] = e
      res[1] = errors.As(&e)
      return
```

In that case, the first result should always be the zero Error value. Yet it's not. Now I need to go read.

Maybe rewrite this to be more explicit and less tricky:

```go
	var e Error
        ok = errors.As(&e)
	return e, ok
```

Or name the result, and then it's also safe:

```go
func As(err error) (e Error, ok bool) {
	ok = errors.As(err, &e)
        return
```

cc @dsnet @creachadair for Go spec stuff

bradfitz: e.g. https://go.dev/play/p/UHESK3ILu3j

bradfitz (approved):

util/vizerror/vizerror.go
willnorris: done.  (no more fixup! commits for me!)

Issue #7123, opened 2023-01-31
Upgrade to Go 1.20


bradfitz:
Tracking bug for updating to Go 1.20.

Notes:

https://tip.golang.org/doc/go1.20

Tree is rebased at https://github.com/tailscale/go/commits/tailscale.go1.20 ... even smaller than our Go 1.19 delta. (net/http stuff is in, mutex contention counters are in)

Going to switch our tree to https://github.com/tailscale/go/commit/6a17f14c058e230cc620b72cf80099c2aa7f7335 (Go 1.20rc3)

Then apply the https://github.com/tailscale/tailscale/issues?q=is%3Aissue+is%3Aopen+Go+1.20+label%3Afuture-go changes.

bradfitz:
(We won't depend on any Go 1.20 features in the open source tree until Go 1.20 is officially out)

bradfitz:
Instead of the metrics we'd previously added to Go, we need to update our internal mutex contention metrics to use this: https://github.com/golang/go/issues/49881

bradfitz:
tailscale/tailscale is updated, but I hit problems in the corp repo: https://github.com/golang/go/issues/58159

bradfitz:
Okay, we're now using Go 1.20 in both repos.

We'll hold on using Go 1.20 API until it's officially out.

PR #7155, opened 2023-02-02
all: use Go 1.20's bytes.Clone

bradfitz:
Updates #7123
Updates #6257 (more to do in other repos)


bradfitz:
```
perl -i -npe 's/append...byte.nil., (.+?)\.\.\../bytes.Clone($1)/' $( git grep -l 'append.*byte.*nil')
goimports -w $(git grep -l bytes.Clone)
```

mihaip (approved):

PR #7156, opened 2023-02-02
.github/workflows: delete CIFuzz job

bradfitz:
It doesn't yet support Go 1.20. We can bring it back later.

Updates #7123


creachadair (approved):

.github/workflows/cifuzz.yml
```diff
@@ -1,31 +0,0 @@
-name: CIFuzz
-on: [pull_request]
```
mihaip, line 2: Could also just disable the `on` line for now.

bradfitz: Done, with comment.

PR #7158, opened 2023-02-02
net/{netns,netstat}: use new x/sys/cpu.IsBigEndian

bradfitz:
See golang/go#57237


maisem (approved):

go.mod
```diff
@@ -36,7 +36,7 @@ require (
 	github.com/iancoleman/strcase v0.2.0
 	github.com/illarion/gonotify v1.0.1
 	github.com/insomniacslk/dhcp v0.0.0-20221215072855-de60144f33f8
-	github.com/josharian/native v1.0.1-0.20221213033349-c1e37c09b531
```
maisem, line 39: We still have usages? Or indirect?

bradfitz: Indirect. So I want to make sure that if any of our indirect ones use it, at least they use the version I bumped to use x/sys/cpu

PR #7164, opened 2023-02-02
net/dnsfallback: update `dns-fallback-servers.json`

coadler:
I was getting some test errors and noticed the default derpmap changed. Not sure what's correct to do here, so feel free to close if this isn't intended to be updated.

Signed-Off-By: Colin Adler <colin1adler@gmail.com>


DentonGentry:
No, we definitely do not want to update this list. I think we should remove the CI check which suggests doing so.

PR #7169, opened 2023-02-03
patch go-iptables to allow dummy netfilter

oskarirauta:
nftables is here to stay and legacy iptables conversion tool does exist- but everyone does not want to use it, or have it at their systems at all; openwrt routers have gone to nftables and deprecated iptables some time ago, and some other systems also have gone that way- then there is problems with tailscaling... Until now..

There is a way, use --netfilter-mode=off to disable all netfilter related features, except it is not enough-
when you start server, tailscaled - it won't start if iptables is not found from path - as it constructs ipt4 and init executes iptables command with argument --version. Ofcourse this fails in a system without iptables, as executable is not available.

There is a workaround though; you could make a shell script to /usr/bin with name iptables, and another possibly with ip6tables, if you are looking forward to have ipv6 support as well, then make it regocnize --version argument and echo a "fake" iptables version string. But if your system runs other software that is capable of using either, nftables or iptables, and prefers iptables, well- rules are made by dummy iptables command. Instead we would need native nftables support for tailscale, but most likely, it will take some more time until development reaches that point. For now, we can use dummy iptables as netfilter, but it would be less hacky for end-user, if it would be possible to use dummy internally instead of creating scripts that fake ipv6.

We could use patched go-iptables. I provided it as a local replacement. All it does, is that it adds few new functions to go-iptables that allow us to "fake" ipt4 and ipt6 that construct IPTables struct with faked values, allowing us to use current tailscale code pretty much as it is and jump over the problem.

I would had created ipt4 and ipt6 structs in linux_router.go but it was not possible as their members were not public, they were unexported, so I was not able to construct struct outside go-iptables module.

What it then does, well- it creates new IPTables struct with hardcoded values for:
```
path to iptables binary: /bin/false
proto: ipv4
hascheck, haswait, waitsupportsecond, hasrandomfully: all false
v1, v2, v3: this is iptables version, result is version 0.0.0
mode: "legacy"
timeout: 0
```
or values can be overriden, for example to support ipv6. Functions work exactly like their original versions, except there's no any checking for this to actually work.
Functions are named NewFake and NewFakeWithProtocol and they are located at iptables/iptables.go

Then I modified router_linux.go to support these new functions, in case - where iptables is not found, new fake ipt is constructed. In case where iptables indeed is a fake, and ip6tables also isn't found, we fake it aswell- IF system supports ipv6.

When iptables is faked, we log it and suggest user to use argument --netfilter-mode off when executing
tailscale up

With this patch, we can ignore iptables and go with netfilter=off, without problems when starting tailscaled, until nft support actually is ready for tailscale. Or if one has iptables, this patch won't render it useless- with iptables in system, it is used unless user sets netfilter off.

Patching go-iptables is made by adding replace directive to pre-patched go-iptables version 0.6.0, to avoid change to system's shared version.

Signed-off-by: Oskari Rauta <oskari.rauta@gmail.com>

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -4626,6 +4626,32 @@ func (b *LocalBackend) Doctor(ctx context.Context, logf logger.Logf) {
 	var checks []doctor.Check
 	checks = append(checks, routetable.Check{})
 
+	// Print a log message if any of the global DNS resolvers are Tailscale
+	// IPs; this can interfere with our ability to connect to the Tailscale
+	// controlplane.
+	checks = append(checks, doctor.CheckFunc("dns-resolvers", func(_ context.Context, logf logger.Logf) error {
+		b.mu.Lock()
+		nm := b.netMap
+		b.mu.Unlock()
+		if nm == nil {
+			return nil
+		}
+
+		for i, resolver := range nm.DNS.Resolvers {
+			ipp, ok := resolver.IPPort()
+			if ok && tsaddr.IsTailscaleIP(ipp.Addr()) {
+				logf("resolver %d is a Tailscale address: %v", i, resolver)
+			}
+		}
+		for i, resolver := range nm.DNS.FallbackResolvers {
+			ipp, ok := resolver.IPPort()
+			if ok && tsaddr.IsTailscaleIP(ipp.Addr()) {
+				logf("fallback resolver %d is a Tailscale address: %v", i, resolver)
+			}
+		}
+		return nil
+	}))
+
 	// TODO(andrew): more
 
 	numChecks := len(checks)
--- a/wgengine/router/router_linux.go
+++ b/wgengine/router/router_linux.go
@@ -111,6 +111,7 @@ type linuxRouter struct {
 	v6Available     bool
 	v6NATAvailable  bool
 	fwmaskWorks     bool // whether we can use 'ip rule...fwmark <mark>/<mask>'
+	iptFaked	bool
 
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -2,6 +2,8 @@ module tailscale.com
 
 go 1.20
 
+replace github.com/coreos/go-iptables => ./go-iptables
```
DentonGentry, line 5: Replace directives break any repo which includes the tailscale repo as a dependency. If you really want this functionality, it needs to go upstream o go-iptables not make a copy of it here.

DentonGentry (changes requested):

go.mod
oskarirauta: I doubt that they need such feature as their purpose is to provide support for iptables instead of nothing. Most likely they propose making changes here instead to not use go-iptables when it doesn't suite. It's just that patch to avoid it, would need a lot more work, I assume.
Is there repos that actually include tailscale as dependency by the way, didn't come to think of that at all..

I am not resolving this conversation, as I didn't actually resolve the issue here. You could though feed my curiosity if you know a repo that has tailscale as dependency, I'd like to see how it's utilized.

oskarirauta: I'll provide another solution that does not require patching go-iptables. Hold on..

oskarirauta: @DentonGentry

I provided another kind of solution.

oskarirauta: For some reason it says that I changed file ipn/ipnlocal/local.go - never touched it... What's with that?

oskarirauta:
Replaced with #7187 - closed

PR #7177, opened 2023-02-03
sockstats: instrument networking code paths

mihaip:
Uses the hooks added by tailscale/go#45 to instrument the reads and writes on the major code paths that do network I/O in the client. The convention is to use `<package>.<type>:<label>` as the annotation for the responsible code path.

Enabled on iOS, macOS and Android only, since mobile platforms are the ones we're most interested in, and we are less sensitive to any throughput degradation due to the per-I/O callback overhead (macOS is also enabled for ease of testing during development).

We also keep track of the current interface so that we can break out the stats by interface.

For now just exposed as counters on a `/v0/sockstats` PeerAPI endpoint.

Updates tailscale/corp#9230
Updates #3363


willnorris (approved):
I did a pretty careful read through and LGTM

ipn/ipnlocal/peerapi.go
```diff
@@ -850,6 +853,54 @@ func (h *peerAPIHandler) handleServeDoctor(w http.ResponseWriter, r *http.Reques
...
+	w.Header().Set("Content-Type", "text/html; charset=utf-8")
+	fmt.Fprintln(w, "<!DOCTYPE html><h1>Socket Stats</h1>")
+
+	stats := sockstats.Get()
+	if stats == nil {
+		fmt.Fprintln(w, "No socket stats available")
+		return
+	}
+
+	fmt.Fprintln(w, "<table border='1' cellspacing='0' style='border-collapse: collapse;'>")
```
crawshaw, line 870: I usually get in trouble for `<table>-in-a-printf` programming, but I see it's the norm in this file! LGTM!

net/sockstats/sockstats.go
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+// Package sockstats collects statistics about network sockets used by
+// the Tailscale client. The context where sockets are used must be
+// instrumented with the WithSockStats() function.
+//
+// Only available on Posix platforms when built with Tailscale's fork of Go.
```
crawshaw, line 8: POSIX

net/sockstats/sockstats_tsgo.go
```diff
@@ -0,0 +1,138 @@
...
+	"tailscale.com/net/interfaces"
+	"tailscale.com/wgengine/monitor"
+)
+
+type sockStatCounters struct {
+	txBytes, rxBytes                           atomic.Uint64
+	rxByBytesByInterface, txByBytesByInterface map[string]*atomic.Uint64
+}
+
+var (
```
crawshaw, line 24: This is fine, but FYI, sometimes to make mutex-guarded values clearer I package these up in an anonymous type:

```
var counters struct {
    mu      sync.Mutex
    byLabel map[string]*sockStatCounters
    curInterface string
    ...
}
```

Then all the access sites are `counters.mu.Lock()`, `counters.byLabel`, etc.

crawshaw (approved):

net/sockstats/sockstats_tsgo.go
mihaip: Added a `sockStats` anonymous struct (didn't call it `counters` because it has interfaces in there too)

mihaip:
I've changed the build tags to only enable this for iOS, Android and macOS, to minimize any performance concerns on other platforms.

Issue #5055, opened 2022-07-14
FR: kubectl support for zero trust


johnathon-b:
### What are you trying to do?

Use `kubectl` across all different clusters with a similar user experience as Tailscale SSH.

### What is the impact of not solving this?

I'm still managing contexts across different clouds via VPN and creds.

### Anything else?

It would be the bee's knees if this were to be implemented.

maisem:
Are you managing your own k8s clusters or are you using a managed service like GKE?

johnathon-b:
I'm using managed services such as:
- Azure Kubernetes Services
- Digital Ocean Managed Kubernetes

As well as a self-hosted cluster.

axsgitty:
This would be a significant improvement to our workflow. We mostly manage remote clusters on bare metal in colocations. We are also doing some testing with Platform9.

johnathon-b:
> This would be a significant improvement to our workflow. We mostly manage remote clusters on bare metal in colocations. We are also doing some testing with Platform9.

I agree. It would allow us to offload most our RBAC. 🤤

guitmz:
this is a killer feature to have :)

rohannair:
Would love this. We have multi-cloud K8S setups, mostly on GKE.

mayakacz:
We have a **Tailscale Kubernetes operator with an auth proxy** now available in [alpha](https://tailscale.com/kb/1167/release-stages/#alpha) for testing. 🎉

The Tailscale Kubernetes operator allows you to, among other things, access the Kubernetes control plane (kube-apiserver) over Tailscale. When a user tries to access the Kubernetes control plane, they will access it with the same user identity that they have in Tailscale.
To set this up, you need to deploy the Tailscale Kubernetes operator, update your kubeconfig locally, and ensure that the user has a cluster role that grants access to the Kubernetes control plane.

See the [documentation on the Kubernetes operator to getting this set up and running](https://tailscale.com/kb/1236/kubernetes-operator/#enabling-an-auth-proxy).

Specific areas we’d like feedback on:
* There is no way to define Tailscale ACLs that restrict Kubernetes RBAC, or map Tailscale ACLs to RBAC. How do you expect to handle user permissions in Kubernetes RBAC if accessing over Tailscale?
* Do you need the ability to record kubectl sessions?
* This operator only allows user@example.com to authenticate as themselves, not impersonate another user, like sre-service-account@example.com or just `sre`. Is impersonation something you need?

[...]

PR #7178, opened 2023-02-03
cmd/k8s-operator: add support for running an auth proxy

maisem:
Updates #5055

### DIFF
--- a/cmd/k8s-operator/operator.go
+++ b/cmd/k8s-operator/operator.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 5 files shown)


cmd/k8s-operator/operator.go
```diff
@@ -222,11 +229,44 @@ waitOnline:
...
+	}
 	if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
 		startlog.Fatalf("could not start manager: %v", err)
 	}
 }
 
+func runAuthProxy(lc *tailscale.LocalClient, ls net.Listener) {
+	u, err := url.Parse(fmt.Sprintf("https://%s:%s", os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT_HTTPS")))
+	if err != nil {
+		panic(err)
```
danderson, line 247: log.fatal instead of panics?

Same below

cmd/k8s-operator/operator.go
```diff
@@ -222,11 +229,44 @@ waitOnline:
...
+	}
+	kc, err := kube.New()
+	if err != nil {
+		panic(err)
+	}
+	p := &httputil.ReverseProxy{
+		Director: func(r *http.Request) {
+			who, err := lc.WhoIs(r.Context(), r.RemoteAddr)
+			if err != nil {
+				panic(err)
```
danderson, line 257: Wrap the reverse proxy in another handler that does the WhoIs lookup and passes it in request context? That way if the whois fails you can respond with a 403 or 500 before ReverseProxy is invoked instead of panicking.

danderson (approved):

cmd/k8s-operator/operator.go
maisem: Good idea, done.

cmd/k8s-operator/operator.go
maisem: Done

maisem:
I was also able to reuse the rest.Client from the operator and added logic to drop any of the other known `Impersonate-*` headers. PTAL

cmd/k8s-operator/proxy.go
```diff
@@ -0,0 +1,76 @@
...
+// LocalAPI and then proxies them to the Kubernetes API.
+type authProxy struct {
+	lc *tailscale.LocalClient
+	rp *httputil.ReverseProxy
+}
+
+func (h *authProxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	who, err := h.lc.WhoIs(r.Context(), r.RemoteAddr)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
danderson, line 34: Log the error internally and return a generic error to the client, we don't know what info might be lurking in the err.

cmd/k8s-operator/proxy.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	ap := &authProxy{
+		lc: lc,
+		rp: &httputil.ReverseProxy{
+			Director: func(r *http.Request) {
+				// Replace the request with the user's identity.
+				who := r.Context().Value(whoIsKey{}).(*apitype.WhoIsResponse)
+				r.Header.Set("Impersonate-User", who.UserProfile.LoginName)
+
+				// Remove all authentication headers.
```
danderson, line 54: May want to wipe all headers and retain only an allowlist, using a blocklist of headers for security purposes is risky if k8s adds any new ones.

danderson (approved):

cmd/k8s-operator/proxy.go
maisem: Spoke on chat, gonna keep a blocklist for now and revisit later.

cmd/k8s-operator/proxy.go
maisem: Good point, done

PR #7180, opened 2023-02-04
net/dns/resolver: remove maxDoHInFlight

mihaip:
It was originally added to control memory use on iOS (#2490), but then was relaxed conditionally when running on iOS 15 (#3098). Now that we require iOS 15, there's no need for the limit at all, so simplify back to the original state.


twitchyliquid64 (approved):

maisem (approved):

maisem:
Thanks!

PR #7187, opened 2023-02-05
disable netfilter if iptables is not present

oskarirauta:
on nftables or any other system using other firewall
than iptables, iptables command is not available
(unless nft-compat/iptables-nft is used)

On these systems, tailscaled cannot be used, as even with
tailscale up -netfilter-mode off, tailscaled daemon constructs
ipt4 and ipt6 - and if ipt4 fails, daemon won't start.

As a workaround, user can create /usr/bin/iptables and
/usr/bin/ip6tables shell scripts responding to --version
and if that argument is used, script returns faked
version string. Downside of this is that if system runs
other software capable of using iptables and prefers
to iptables, rules are made with faking shell script.

With this commit, netfilter mode is set off, in case
that iptables is not present in system and all functions
that handle rules are ignored, allowing users without
iptables to use tailscale. Also, if iptables is missing
SNAT rules are also ignored.

If user has iptables- it is used as this commit would
not exist at all, so it does not affect iptables users
in anyway.

Replaces #7169


bradfitz (changes requested):
I don't think we want to just silently change behavior and defaults based on $PATH or what packages may or may not be installed.

This needs to be more explicit from the user (e.g. some new `envknob` for `tailscaled`) to go into this mode.

oskarirauta:
@bradfitz
it would be much better - if argument `--netfilter-mode off` would work from tailscaled command-line, as it does from tailclient. I actually wanted to do that instead, but it is a bit difficult, because preferences from tailscaled, are not carried to router setup, and even if it works, there is Reconfig- to archieve this solution, there would be needed to be a globally accessible variable that would be set to true -

Do you have any ideas on how to archieve that?

bradfitz:
I forget. I've asked others about that history.

Meanwhile, this needs to be explicit. I still suggest an `envknob`.

PR #7194, opened 2023-02-06
ipn/ipnlocal: handle more edge cases in netmap expiry timer

### LINKED ISSUE #7193: Infinite netmap reload loop due to client-side expiry handling
There's CPU profile from a user ([ticket](http://go/hs/9322)) that shows a lot of time spent doing netmap updates:

image

Based on logs, it appears to be related to client-side expiration handling (#6937), e.g. there so many logs that they're getting throttled:

```
2023-01-31 23:38:38.729611 +0000 UTC: [RATELIMIT] format("setClientStatus: netmap expiry timer triggered after %v") (48618 dropped)
```

andrew-d:
We now handle the case where the NetworkMap.SelfNode has already expired and do not return an expiry time in the past (which causes an ~infinite loop of timers to fire).

Additionally, we now add an explicit check to ensure that the next expiry time is never before the current local-to-the-system time, to ensure that we don't end up in a similar situation due to clock skew.

Finally, we add more tests for this logic to ensure that we don't regress on these edge cases.

Fixes #7193

Change-Id: Iaf8e3d83be1d133a7aab7f8d62939e508cc53f9c

### DIFF
--- a/ipn/ipnlocal/expiry.go
+++ b/ipn/ipnlocal/expiry.go
@@ -69,15 +69,16 @@ func (em *expiryManager) onControlTime(t time.Time) {
 // Node.Expired so other parts of the codebase can provide more clear error
 // messages when attempting to e.g. ping an expired node.
 //
+// The localNow time should be the output of time.Now for the local system; it
+// will be adjusted by any stored clock skew from ControlTime.
+//
 // This is additionally a defense-in-depth against something going wrong with
 // control such that we start seeing expired peers with a valid Endpoints or
 // DERP field.
 //
 // This function is safe to call concurrently with onControlTime but not
 // concurrently with any other call to flagExpiredPeers.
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/expiry.go
```diff
@@ -119,3 +120,94 @@ func (em *expiryManager) flagExpiredPeers(netmap *netmap.NetworkMap) {
 		peer.Key = key.NodePublicWithBadOldPrefix(peer.Key)
 	}
 }
+
+// nextPeerExpiry returns the time that the next node in the netmap expires
```
mihaip, line 124: May want to document that the zero value for `time.Time` is a possible return value.

mihaip (approved):

ipn/ipnlocal/expiry.go
andrew-d: Good idea; added

PR #7197, opened 2023-02-07
version: unify and optimize the various not-version funcs

danderson:


maisem (approved):

bradfitz (commented):
I'm not a fan of this. It's extra init work that can't be deadcoded anymore.

PR #7198, opened 2023-02-07
version: return correct Meta.MajorMinorPatch in non-dev builds

danderson:

### DIFF
--- a/version-embed.go
+++ b/version-embed.go
@@ -6,8 +6,13 @@ package tailscaleroot
 
 import _ "embed"
 
+// VersionDotTxt is the contents of VERSION.txt. Despite the tempting filename,
+// this does not necessarily contain the accurate version number of the build, which
+// depends on the branch type and how it was built. To get version information, use
+// the version package instead.
+//
 //go:embed VERSION.txt
-var Version string
+var VersionDotTxt string
 
 //go:embed ALPINE.txt
 var AlpineDockerTag string
--- a/version/version.go
+++ b/version/version.go
@@ -80,6 +80,9 @@ var (
 
 	// isWindowsGUI is whether the current binary is the Windows GUI binary.
 	isWindowsGUI bool
+
+	// majorMinorPatch is the major.minor.patch portion of Short.
+	majorMinorPatch string
 )
 
 func init() {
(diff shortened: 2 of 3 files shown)


version/version.go
```diff
@@ -126,10 +130,18 @@ func initVersion() {
 	}
 
 	// Backup path, using Go 1.18's built-in git stamping.
-	Short = strings.TrimSpace(tailscaleroot.Version) + "-dev" + commitDate
+	Short = strings.TrimSpace(tailscaleroot.VersionFileDoNotUseDirectly) + "-dev" + commitDate
 	Long = Short + "-t" + commitHashAbbrev + dirty
 }
 
+func initMajorMinorPatch() {
+	if i := strings.Index(Short, "-"); i != -1 {
```
bradfitz, line 138: strings.Cut

version-embed.go
```diff
@@ -6,8 +6,12 @@ package tailscaleroot
 
 import _ "embed"
 
+// VersionFileDoNotUseDirectly is the contents of VERSION.txt. Do not use this
+// directly to figure out the version of the binary, use the version package
+// instead.
+//
 //go:embed VERSION.txt
-var Version string
+var VersionFileDoNotUseDirectly string
```
bradfitz, line 14: kinda gross name. I don't like prescriptive stuff like this. I'd rather name it what it is. What is it?

danderson: The bug was that unstable builds, even built correctly with version embedding, would report 1.37.0 as their MajorMinorPatch despite the correct version information existing in the binary. Easy mistake to make because that top level package has a juicy sounding "Version" that must surely be the version, only it's not, it's the literal contents of VERSION.txt, which is only loosely related to the actual build version.

The more literal name would be something like `VersionFileContents`, but then I still have to use the comment to say that the version file contents is almost certainly not what you want to use if you're hoping to get version information. Do you prefer that?

bradfitz: Or `VersionTxtBranchOrReleaseVersionFileContents` if you want something long and descriptive rather than prescriptive.  Because it's either a branch (1.35.0) or an explicit stable release version number.

danderson: How's this?

version-embed.go
```diff
@@ -6,8 +6,12 @@ package tailscaleroot
 
 import _ "embed"
 
+// VersionDotTxt is the contents of VERSION.txt. Do not use this
+// directly as the binary's version, use the version package
+// instead.
```
bradfitz, line 11: If you're going to say no, say a bit about why.

```
// VersionDotTxt is the contents of VERSION.txt. Despite the tempting filename,
// this does not necessarily contain the version number of the build, which depends
// on the branch type and how it was built. To get version information, use the
// version package instead.
```

danderson: Done.

bradfitz (approved):

PR #7203, opened 2023-02-07
ipn/ipnlocal: fix the path for writing cert files

### LINKED ISSUE #7202: ipn/ipnlocal: certificates and keys are both written to the same file
See: https://github.com/tailscale/tailscale/blob/main/ipn/ipnlocal/cert.go#L179

This should be using `certFile` instead (was modified in #7142). I believe this is the underlying cause of internal issue https://github.com/tailscale/corp/issues/9090)

creachadair:
I'd welcome suggestions for a good way to add a test for this.

Fixes #7202.

Change-Id: I1f8e9c59d5e42e7df7a3fbbd82ae2b4293845916
Signed-off-by: M. J. Fromberger <fromberger@tailscale.com>


maisem:
Oops! We can write a test where we instantiate the two cert stores and write some data and try to read it back
we can use a temp dir (like `t.TempDir()`) and use the a `mem.Store` for the state store

creachadair:
> Oops! We can write a test where we instantiate the two cert stores and write some data and try to read it back we can use a temp dir (like `t.TempDir()`) and use the a `mem.Store` for the state store

Good idea, I'll do that.

creachadair:
@maisem I added a test, but it's not entirely satisfactory. The x509 verifier won't accept a self-signed certificate, and the way the plumbing is set up we don't get the specific error back when validation fails. For now I just tested that the bits go to the right place (and in doing so, found another path issue).  I'm happy to come back for another pass if we can figure out a better solution.

creachadair:
Converted this back to a draft while I follow up on some ideas @maisem suggested, I'll ping back when it's ready for another round.

creachadair:
I think this is ready for another look. Following @maisem's suggestion I plumbed a hook so the test can inject a CA root, and used `mkcert` to generate a root CA and site cert to complete the round trip.

ipn/ipnlocal/cert.go
```diff
@@ -458,14 +468,26 @@ func acmeKey(dir string) (crypto.Signer, error) {
...
+		log.Printf("cert check: failed to get system roots: %v", err)
+		return false
+	}
+	if len(extraRoots) != 0 {
+		if !roots.AppendCertsFromPEM(extraRoots) {
+			log.Print("cert check: failed append certificates")
+			return false
+		}
+	}
+
```
maisem, line 490: instead of appending to the system `*x509.CertPool` here, create an empty root for in the test code and pass that directly

ipn/ipnlocal/cert_test.go
```diff
@@ -26,3 +33,64 @@ func TestValidLookingCertDomain(t *testing.T) {
...
+	// use:
+	//
+	//   	go run filippo.io/mkcert@latest example.com
+	//
+	// The content is not important except to be structurally valid so we can be
+	// sure the round-trip succeeds.
+	testRoot, err := certTestFS.ReadFile("testdata/rootCA.pem")
+	if err != nil {
+		t.Fatal(err)
+	}
```
maisem, line 58: so you can do
```
roots := x509.NewCertPool
if !roots.AppendCertsFromPEM(extraRoots) {
  t.Fatal("cert check: failed append certificates")
  return false
}
```

maisem (approved):

bradfitz:
Didn't review but be sure to squash.

ipn/ipnlocal/cert.go
```diff
@@ -458,14 +468,26 @@ func acmeKey(dir string) (crypto.Signer, error) {
...
 	}
 	tlsCert, err := tls.X509KeyPair(certPEM, keyPEM)
 	if err != nil {
 		return false
 	}
+	roots, err := x509.SystemCertPool()
+	if err != nil {
+		log.Printf("cert check: failed to get system roots: %v", err)
+		return false
+	}
```
bradfitz, line 483: Why is this a failure for a new path that should only be able tests?

bradfitz: Oh, this was implicit before by being nil?

ipn/ipnlocal/cert.go
```diff
@@ -146,46 +146,56 @@ var errCertExpired = errors.New("cert expired")
...
+	return certFileStore{dir: dir}
 }
 
 // certFileStore implements certStore by storing the cert & key files in the named directory.
-type certFileStore string // dir
+type certFileStore struct {
+	dir string
+
+	// This field allows a test to inject additional CA root(s) for certificate
+	// verification. If nil or empty, the default system pool is used.
```
bradfitz, line 159: // testRoots, if non-empty for tests, is the PEM-formatted blah blah.

(is it PEM? I don't know!)

ipn/ipnlocal/cert.go
creachadair: Yes, the default if the option is nil is to use the system cert pool.

ipn/ipnlocal/cert.go
creachadair: It is. Updated.

ipn/ipnlocal/cert_test.go
creachadair: Almost—but we have to leave `roots == nil` for the production case, because otherwise we won't get the system cert pool (which is the default).

ipn/ipnlocal/cert.go
creachadair: Oh, I see what you are saying now. Will update again.

creachadair: Done, you're right this is better.

ipn/ipnlocal/cert_test.go
creachadair: (see above, I misunderstood what you were asking for here, but I get it now 🙂)

Issue #7055, opened 2023-01-24
cmd/tailscale: automatically set unattended on Windows from CLI?


bradfitz:
Should we make `tailscale up --unattended` the default on Windows, at least when the GUI's not running?

Otherwise if the CLI's the only client, the connection count goes from 0 to 1 (for the up) to 0 and then tailscaled stops, as it's not in server mode on Windows by default. Weird experience. 

If the GUI's running, though, the connection count is 1 (from the GUI) so it goes 1 to 2 to 1 and the GUI keeps it running.

So maybe we should only default to `--unattended` when it's the first connection. But we have no LocalAPI to expose that. I guess we could.

/cc @dblohm7 @maisem @kradalby

bradfitz:
Or, if we add a LocalAPI to ask whether a GUI's connected (or at least any long-lived connection) we could then have the CLI `up` command make the lack of `--unattended` be an error, making users spell it out, rather than be surprised by the disconnect a half second later.

DentonGentry:
Related: https://github.com/tailscale/tailscale/issues/6611 proposes to add an `--unattended` flag to the Tailscale service, so `tailscale.exe` won't need to remember to set it.

bradfitz:
I just verified this is still a problem in 1.98.2. And it's even kinda worse than it used to be with various state machines and netmap caching fighting each other. It's a super confusing experience on Windows.

PR #7205, opened 2023-02-07
ipn/ipnlocal: return error on missing --unattended on CLI up on Windows

bradfitz:
Maybe not the best long-term fix, but better than letting the user start the server for a millisecond just to have it all torn down as their CLI command finishes running.

Updates #7055


bradfitz:
(not yet tested)

Thoughts?

DentonGentry (approved):

dblohm7 (approved):
I've been burned by this before!

maisem (approved):

bradfitz:
If anybody can test this easily that'd be great. My Windows dev environment is still down until I repair my server rack power.

bradfitz:
Ah, test failure looks real, though:

```
    local.go:2603: EditPrefs check error: Use 'tailscale up --unattended'
```

Guess we can just fix the test, though.

PR #7214, opened 2023-02-08
control/controlhttp: don't require valid TLS cert for Noise connection

bradfitz:
We don't require any cert at all for Noise-over-plaintext-port-80-HTTP, so why require a valid cert chain for Noise-over-HTTPS? The reason we use HTTPS at all is to get through firewalls that allow tcp/443 but not tcp/80, not because we need the security properties of TLS.

Updates #3198

### DIFF
--- a/control/controlhttp/client.go
+++ b/control/controlhttp/client.go
@@ -410,10 +410,24 @@ func (a *Dialer) tryURLUpgrade(ctx context.Context, u *url.URL, addr netip.Addr,
 	tr.TLSClientConfig.NextProtos = []string{}
 	tr.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{}
 	tr.TLSClientConfig = tlsdial.Config(a.Hostname, tr.TLSClientConfig)
-	if a.insecureTLS {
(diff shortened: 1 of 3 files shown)


DentonGentry (approved):

andrew-d (approved):
I confirmed that this is only (currently) called via the Noise codepath, so it seems fine, but is there any way we can be extra-sure that we're not accidentally disabling this for non-Noise dials? e.g. any future endpoints we might add, if we refactor our log dialer, etc.?

bradfitz:
@andrew-d, this code is only modifying a Noise file?

andrew-d:
> @andrew-d, this code is only modifying a Noise file?

Right, this is 100% fine today... but `controlhttp.Dialer` does not sound like a noise-only type and I'm slightly worried that this totally-fine code will migrate over time to somewhere that's less fine.

Also fine to say "nope, nothing easy to do here" and merge.

bradfitz:
Well, that whole package and directory hierarchy (e.g. control/controlbase) are only about noise so I'm not super worried. I don't know what else we could do short of renaming the whole directly more loudly.

Issue #7188, opened 2023-02-05
iOS 16.3 tailscale 1.36.0 cannot reauthenticate


digaround:
### What is the issue?

Can not reauthenticate on ios 16.3 and tailscaled 1.36.0, on the screen loading icon, after click on "Reauthenticate" button nothing happen:

<img width="529" alt="image" src="https://user-images.githubusercontent.com/62701121/216832358-c98a0012-c2c2-47fa-8eb9-591bb8b43ca0.png">

Rebooted phone, reinstalled tailscaled app, of course network is working, other sites/application loads network data correclty. Attaching bug report.

### Are there any recent changes that introduced the issue?

Just upgrade ios and tailscaled

### OS

iOS

### OS version

iOS 16.3

### Tailscale version

1.36.0

### Bug report

BUG-5f67cb5a09456f762532941be8d9c6754b6c1f6542b17603f81c6b1d3e17c988-20230205112024Z-7a32b25eccbeb6c0

mihaip:
Based on the logs the Tailscale extension is not able to reach the control plane server -- they're filled with lines like:

```
13:51:14 | ^38.5M control: TryLogin: register request: Post "https://controlplane.tailscale.com/machine/register": connection attempts aborted by context: context deadline exceeded
13:51:14 | ^38.4M Received error: register request: Post "https://controlplane.tailscale.com/machine/register": connection attempts aborted by context: context deadline exceeded
13:51:14 | ^38.5M control: PollNetMap: Post "https://controlplane.tailscale.com/machine/map": connection attempts aborted by context: context deadline exceeded
13:51:14 | ^38.4M Received error: PollNetMap: Post "https://controlplane.tailscale.com/machine/map": connection attempts aborted by context: context deadline exceeded
```

Are you running any other VPN or traffic blocking software on your phone (e.g. iCloud Private Relay)? Does this persist if you toggle Tailscale from the "VPN" section in the iOS settings app?

iksaku:
@mihaip I've got the exact same issues and I did try "disconnecting" from the settings app as you suggest, but it doesn't work, and if you go back again to settings app, the VPN toggle is back on.

I've got no other VPN profiles/apps installed nor iCloud Relay enabled.

quartermarsh:
[...]

Issue #7184, opened 2023-02-04
Tailscale on = no cellular Internet


antgiant:
Over the past few days I’ve been having an issue where turning on tailscale on my iPhone results in no cellular Internet or tailscale network. I am not aware of any changes to my config, and this bug is not happening when I use wireless on my phone. I’ve tried turning the active switch off and on, force restarting the app, cycling airplane mode, and completely rebooting my phone. None made any difference.
If it helps here is a bug id.

BUG-fc43f800d65de659824423a8c60156ce018ce2c91077b49c757807cd6ecd353f-20230204224633Z-6ff0a265b04d66d9

mihaip:
Do you happen to have multiple eSIM cards in your phone or multiple cell data plans? We made a change in Tailscale v1.36 (to fix #6565) where we more specifically request Wi-Fi or cell data networking from iOS. Looking at the logs for your phone, with v1.34 it would pick the `pdp_ip0` when using cell data, but it now picks `pdp_ip1`, and so perhaps it's trying to use the wrong cell plan.

antgiant:
No, I only have one cell plan and one sim. However, I do have an Apple Watch with a cell plan. Could that perhaps be related?

mihaip:
I don't believe the Apple Watch cell plan is visible to the phone. Based on some more digging, `pdp_ip0` should be the standard cell interface, and `pdp_ip1` is the one used by VoLTE (voice-over-LTE), which is not something that Tailscale should be trying to do.

A long shot, but are you aware of any non-standard cell data configurations on your phone (either done by you or your carrier)?

mihaip:
We released a build to TestFlight that we believe should fix this problem. If you'd like to try it out, you can install version 1.37.101 using the instructions at https://tailscale.com/kb/1083/install-unstable/.

antgiant:
Hey that fixed it for me! Thank you so much! So, nice to have it working in both Cell and WiFi again!

mihaip:
Thank you for conforming, the fix was picked up in the recently-released 1.36.2 stable release.

henrikstroem:
I have problems with Tailscale on iOS since last update: when activated, Apple Car Play no longer has internet access. Disabling Tailscale solves the problem.
I don't know if this is important, but reading the posts above, I just want to add that I have two SIMs in my phone: a physical card and an eSIM. Both have a data plan, and iOS is set up to use both.

PR #7215, opened 2023-02-08
net/interfaces, net/netns: add node attributes to control default interface getting and binding

mihaip:
With #6566 we started to more aggressively bind to the default interface on Darwin. We are seeing some reports of the wrong cellular interface being chosen on iOS. To help with the investigation, this adds to knobs to control the behavior changes:
- `CapabilityDebugDisableAlternateDefaultRouteInterface` disables the alternate function that we use to get the default interface on macOS and iOS (implemented in tailscale/corp#8201). We still log what it would have returned so we can see if it gets things wrong.
- `CapabilityDebugDisableBindConnToInterface` is a bigger hammer that disables binding of connections to the default interface altogether.

Updates #7184
Updates #7188

### DIFF
--- a/net/interfaces/interfaces_bsd.go
+++ b/net/interfaces/interfaces_bsd.go
@@ -40,9 +40,15 @@ func defaultRoute() (d DefaultRouteDetails, err error) {
 // owns the default route. It returns the first IPv4 or IPv6 default route it
 // finds (it does not prefer one or the other).
 func DefaultRouteInterfaceIndex() (int, error) {
+	disabledAlternateDefaultRouteInterface := false
 	if f := defaultRouteInterfaceIndexFunc.Load(); f != nil {
 		if ifIndex := f(); ifIndex != 0 {
-			return ifIndex, nil
+			if !disableAlternateDefaultRouteInterface.Load() {
+				return ifIndex, nil
+			} else {
+				disabledAlternateDefaultRouteInterface = true
+				log.Printf("interfaces_bsd: alternate default route interface function disabled, would have returned interface %d", ifIndex)
+			}
 		}
 		// Fallthrough if we can't use the alternate implementation.
 	}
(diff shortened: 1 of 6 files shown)


net/interfaces/interfaces_bsd.go
```diff
@@ -40,9 +40,15 @@ func defaultRoute() (d DefaultRouteDetails, err error) {
 // owns the default route. It returns the first IPv4 or IPv6 default route it
 // finds (it does not prefer one or the other).
 func DefaultRouteInterfaceIndex() (int, error) {
+	disabledAlternateDefaultRouteInterface := false
```
andrew-d, line 43: Plausibly not, but do we want to add an `envknob` as well as a nodeAttr? Specifically for `utun`-mode macOS and/or Headscale?

tailcfg/tailcfg.go
```diff
@@ -94,7 +94,8 @@ type CapabilityVersion int
 //   - 55: 2023-01-23: start of c2n GET+POST /update handler
 //   - 56: 2023-01-24: Client understands CapabilityDebugTSDNSResolution
 //   - 57: 2023-01-25: Client understands CapabilityBindToInterfaceByRoute
-const CurrentCapabilityVersion CapabilityVersion = 57
+//   - 58: 2023-02-08: Client understands CapabilityDebugDisableAlternateDefaultRouteInterface and CapabilityDebugDisableBindConnToInterface
+const CurrentCapabilityVersion CapabilityVersion = 58
```
bradfitz, line 98: Let's not bump this. Then we can cherry-pick it back. We can't cherry-pick capver bumps.

andrew-d (approved):

tailcfg/tailcfg.go
mihaip: Removed.

net/interfaces/interfaces_bsd.go
mihaip: I believe #6566 did not affect those clients -- all OSS versions don't provide an alternate interface function, and we were binding to the default interface for them (see the implementation in `net/netns/netns_darwin_tailscaled.go` from before the change).

Also, I'm hoping these knobs are temporary, so don't necessarily want to add more surface area for them.

Also I was lazy :)

bradfitz (approved):

PR #7218, opened 2023-02-08
.github/workflows: re-enable CIFuzz job

mihaip:
Having an empty `on` spec results in the job still running, but it immediately fails with a "No jobs were run" message.

Go back to the original `on: [pull_request]` spec, and disable the workflow in the GitHub UI instead.

This reverts commit f7b3156f1635d539bc59e5015661b235f80f2b90.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,5 +1,5 @@
 name: CIFuzz
-on: [] # was: [pull_request], but disabled in https://github.com/tailscale/tailscale/pull/7156
+on: [pull_request]
 
 concurrency:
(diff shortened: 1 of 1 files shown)


mihaip:
https://github.com/tailscale/tailscale/actions/workflows/cifuzz.yml says
image

bradfitz:
Do they support Go 1.20 yet?

bradfitz (approved):
Or just delete it all like I wanted to originally :)

mihaip:
https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/install_go.sh is still hardcoding 1.19

Issue #7220, opened 2023-02-08
FR: add "tailscale configure" subcommand


maisem:
### What are you trying to do?

This would ideally be the place where we add things that configure the environment to enable more features without having the user manually do them.

Ideally things like #6665 and #2928 would go there.

PR #7222, opened 2023-02-08
cmd/tailscale/cli: add "configure kubeconfig" subcommand 

maisem:
It takes in a node hostname and configures the local kubeconfig file to
point to that.

Updates #7220


maisem:
the k8s dependency adds 5M to the binary size, going to do something else

cmd/tailscale/cli/configure.go
```diff
@@ -0,0 +1,32 @@
...
+	"tailscale.com/version/distro"
+)
+
+var configureCmd = &ffcli.Command{
+	Name: "configure",
+	FlagSet: (func() *flag.FlagSet {
+		fs := newFlagSet("configure")
+		return fs
+	})(),
+	Subcommands: configureSubcommands(),
```
bradfitz, line 21: no help text?

cmd/tailscale/cli/configure-kube.go
```diff
@@ -0,0 +1,132 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+//go:build !ts_omit_kube
```
bradfitz, line 3: Let's also add `ts_small` too. I filed https://github.com/tailscale/tailscale/issues/7230 

So this is then

`!ts_omit_kube && !ts_small`

cmd/tailscale/cli/configure-kube.go
```diff
@@ -0,0 +1,132 @@
...
+		return errors.New("unknown arguments")
+	}
+
+	st, err := localClient.Status(ctx)
+	if err != nil {
+		return err
+	}
+
+	confPath := filepath.Join(homedir.HomeDir(), ".kube", "config")
+	targetFQDN := args[0] + "." + st.CurrentTailnet.MagicDNSSuffix + "."
```
bradfitz, line 49: check status before using this. Is CurrentTailnet non-nil? Is there a suffix?

cmd/tailscale/cli/configure-kube.go
maisem: I'll add another commit that does it so that it's a single change that spans the repo as opposed to being introduced in one commit in one place.

cmd/tailscale/cli/configure-kube.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+func init() {
+	configureCmd.Subcommands = append(configureCmd.Subcommands, configureKubeconfigCmd)
+}
+
+var configureKubeconfigCmd = &ffcli.Command{
+	Name:       "kubeconfig",
+	ShortHelp:  "Configure kubeconfig to use Tailscale",
+	ShortUsage: "kubeconfig [hostname]",
```
bradfitz, line 30: hostname of what?

cmd/tailscale/cli/configure-synology.go
```diff
@@ -18,9 +18,23 @@ import (
 	"tailscale.com/version/distro"
 )
 
+// configureHostCmd is the "tailscale configure-host" command which
+// was once used to configure Synology devices, but is now a redirect
+// to "tailscale configure synology".
```
mayakacz, line 23: sgtm

mayakacz (approved):

cmd/tailscale/cli/configure-kube.go
maisem: Done and added more checks, also allowed inputing the FQDN so that users can target shared nodes.

cmd/tailscale/cli/configure-kube.go
maisem: Added usage docs, ptal

cmd/tailscale/cli/configure.go
maisem: Added

cmd/tailscale/cli/configure-kube.go
maisem: Done. PTAL

maisem: Realized that we already have `./build_dist.sh --extra-small` so not going to add `ts_small` as a tag and instead make that the supported way.

cmd/tailscale/cli/configure-kube.go
```diff
@@ -0,0 +1,156 @@
...
+		if st.CurrentTailnet.MagicDNSSuffix == "" {
+			return errors.New("no magic DNS suffix")
+		}
+		if !st.CurrentTailnet.MagicDNSEnabled {
+			return errors.New("magic DNS is not enabled")
+		}
+		targetFQDN = fmt.Sprintf("%s.%s.", targetFQDN, st.CurrentTailnet.MagicDNSSuffix)
+	} else if !strings.HasSuffix(targetFQDN, ".") {
+		targetFQDN = targetFQDN + "."
+	}
```
bradfitz, line 71: IIRC, doesn't the cli package already have some func that maps from a rando string to a peer node? pity to see another implementation of this.

cmd/tailscale/cli/configure-kube.go
```diff
@@ -0,0 +1,156 @@
...
+
+	confPath := filepath.Join(homedir.HomeDir(), ".kube", "config")
```
bradfitz, line 73: you want to stat it first to see if it exists? Or do you expect it to not exist commonly?

cmd/tailscale/cli/configure-kube.go
maisem: let me see if i can find it/reuse.

cmd/tailscale/cli/configure-kube.go
```diff
@@ -0,0 +1,156 @@
...
+	for _, peer := range st.Peer {
+		if peer.DNSName == targetFQDN {
+			if err := setKubeconfigForPeer(peer, confPath); err != nil {
+				return err
+			}
+			printf("kubeconfig configured for %q\n", args[0])
```
bradfitz, line 79: no args[0] here and L83. make a var so this has a good name.

cmd/tailscale/cli/configure-kube.go
maisem: crap, I was relying on the old behavior of the k8s library. Will fix

cmd/tailscale/cli/configure-kube.go
```diff
@@ -0,0 +1,156 @@
...
+	var clusters []any
+	if cm, ok := cfg["clusters"]; ok {
+		clusters = cm.([]any)
+	}
+	cfg["clusters"] = appendOrSetNamed(clusters, fqdn, map[string]any{
+		"name": fqdn,
+		"cluster": map[string]string{
+			"server": "https://" + fqdn,
+		},
+	})
```
bradfitz, line 122: let's not mix the I/O with the pure func part of this. Move the YAML manipulation out to its own func and then its testable. then write some tests, since we're messing with user's configs.

cmd/tailscale/cli/configure-kube.go
maisem: Done

cmd/tailscale/cli/configure-kube.go
maisem: Done

cmd/tailscale/cli/configure-kube.go
maisem: Done

cmd/tailscale/cli/configure-kube.go
maisem: Done

cmd/tailscale/cli/configure-synology.go
```diff
@@ -18,9 +18,23 @@ import (
 	"tailscale.com/version/distro"
 )
 
+// configureHostCmd is the "tailscale configure-host" command which
+// was once used to configure Synology devices, but is now a redirect
```
bradfitz, line 22: now a compatibility alias for "tailscale configure synology"

bradfitz (approved):

cmd/tailscale/cli/configure-synology.go
maisem: Done

PR #7223, opened 2023-02-08
wgengine: start logging DISCO frames to pcap stream

twitchyliquid64:


twitchyliquid64:
image

wgengine/capture/capture.go
```diff
@@ -65,6 +65,9 @@ const (
 	// SynthesizedToPeer indicates the packet was generated from within tailscaled,
 	// and is being routed to a remote Wireguard peer.
 	SynthesizedToPeer Path = 3
+
+	// PathDisco indicates the packet is metadata about a valid disco frame.
```
bradfitz, line 69: drop "valid"? invites more questions. (like: is there a path for invalid disco frames?)

And is it just "metadata"? It also contains the payload. Maybe you want something like "a decrypted disco message"?

wgengine/magicsock/magicsock.go
```diff
@@ -80,6 +82,12 @@ const (
 	socketBufferSize = 7 << 20
 )
 
+// CaptureFunc describes a callback to record packets when
+// debugging packet-capture. Such callbacks must not take
+// ownership of the provided data slice: it may only copy
+// out of it within the lifetime of the function.
+type CaptureFunc func(capture.Path, time.Time, []byte)
```
bradfitz, line 89: why's this here in magicsock? if you're already importing the capture package?

wgengine/magicsock/magicsock.go
```diff
@@ -348,6 +356,9 @@ type Conn struct {
 	// stats maintains per-connection counters.
 	stats atomic.Pointer[connstats.Statistics]
 
+	// pcap logging callback - non-nil when capturing.
```
bradfitz, line 359: // captureHook, if non-nil, is the pcap logging callback when capturing.

wgengine/magicsock/magicsock.go
```diff
@@ -664,6 +675,10 @@ func NewConn(opts Options) (*Conn, error) {
 	return c, nil
 }
 
+func (c *Conn) InstallCaptureHook(cb CaptureFunc) {
```
bradfitz, line 678: docs

wgengine/magicsock/magicsock.go
```diff
@@ -2099,6 +2114,29 @@ func (c *Conn) handleDiscoMessage(msg []byte, src netip.AddrPort, derpNodeSrc ke
 		return
 	}
 
+	// Emit information about the disco frame into the pcap stream
+	// if a capture hook is installed.
+	if cb := c.captureHook.Load(); cb != nil {
+		var (
+			l    bytes.Buffer
```
bradfitz, line 2121: never use `l` as a variable name. Looks terrible in too many fonts. :)

wgengine/magicsock/magicsock.go
```diff
@@ -2099,6 +2114,29 @@ func (c *Conn) handleDiscoMessage(msg []byte, src netip.AddrPort, derpNodeSrc ke
 		return
 	}
 
+	// Emit information about the disco frame into the pcap stream
+	// if a capture hook is installed.
+	if cb := c.captureHook.Load(); cb != nil {
+		var (
```
bradfitz, line 2120: pull this all out into a func? handleDiscoMessage is long enough already.

wgengine/magicsock/magicsock.go
```diff
@@ -2099,6 +2114,29 @@ func (c *Conn) handleDiscoMessage(msg []byte, src netip.AddrPort, derpNodeSrc ke
...
+			l    bytes.Buffer
+			flag uint8
+		)
+		if src.Addr() == derpMagicIPAddr {
+			flag |= 0x01
+		}
+		l.Write([]byte{flag}) // 1b: flag
```
bradfitz, line 2127: WriteByte?

wgengine/magicsock/magicsock.go
bradfitz, line 2121: call Grow on it first, before all the little writes?

but this is garbage per disco message. probably okay. Maybe just add a comment like "// Warning: alloc garbage. Acceptable when capturing."

bradfitz (commented):
Exciting!

wgengine/magicsock/magicsock.go
twitchyliquid64: Just keeping the type where its used. We already have equivalents in tstun and wgengine, if we want to centralize it maybe we should rip the same out of there too?

bradfitz: yeah, one place would be ideal

bradfitz (approved):

PR #7225, opened 2023-02-09
net/connstats: fix ticker in `NewStatistics`

coadler:
`:=` was accidentally used, so `maxPeriod` never worked.


bradfitz:
cc @dsnet

dsnet:
Ick! Thank you for the bug fix.

PR #7244, opened 2023-02-10
doctor, ipnlocal: add scutil check and tweak logging

andrew-d:
This prints the slightly-parsed output of "scutil --dns" on macOS when running a bugreport with the diagnose flag. Additionally, this reworks how the doctor package prints logs to ensure that we don't interleave log output between checks while still allowing checks to execute concurrently.

Change-Id: If3049f8c4e1b2cf624ed8670883ad6166b15bf7f


doctor/scutil/scutil_darwin.go
```diff
@@ -0,0 +1,156 @@
...
+	"os/exec"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"tailscale.com/types/logger"
+)
+
+func (Check) Run(ctx context.Context, logf logger.Logf) error {
+	cmd := exec.CommandContext(ctx, "scutil", "--dns")
```
mihaip, line 17: Somewhat surprised this works from the macOS sandbox (but it does). However, we won't be able to use it on iOS. 

Though more work (as far as having to rely on CGo), you could use the dnsinfo private API (see tailscale/corp#7680) which lets you get at this information more directly (it's what scutil uses in its implementation -- https://github.com/apple-oss-distributions/configd/blob/ebd2cef166d947659a001b076ea919351f7baf5d/scutil.tproj/tests.c#L703-L716).

doctor/scutil/scutil_darwin.go
```diff
@@ -0,0 +1,156 @@
...
+	Config     map[string]string
+	ListConfig map[string][]string
+}
+
+var (
+	reSpacePrefix = regexp.MustCompile(`\A\s+[^\s]`)
+	reNumSuffix   = regexp.MustCompile(`\A(.+)\[(\d+)\]\z`)
+)
+
+func parseScutilDns(logf logger.Logf, data string) (*dnsInfo, error) {
```
mihaip, line 67: Nit: I think it's common Go style for acronyms to remain all caps, and not be treated as words (https://github.com/golang/go/wiki/CodeReviewComments#initialisms), so this should be called `parseScutilDNS`

mihaip (approved):

doctor/scutil/scutil_darwin.go
andrew-d: Would you prefer that I update this to rely on CGo, or are you okay to merge this as-is and follow up with the dnsinfo private API? I'm a bit leery of relying on that, though given that this is debug-only and how long it appears to have been around for it's probably fine?

mihaip (approved):

PR #7245, opened 2023-02-10
cmd/tailscale/cli: add logout and debug info to web

### LINKED ISSUE #7238: FR: Add logout function to web interface

Allow someone to log out and in through `tailscale web` 

### How should we solve this?

_No response_

non-root users can't log out then back in using the --operator function. `tailscale web` can be run with sufficient privileges to do this. 

### Anything else?

Internal ref HS9534

phirework:
Fixes #7238

Also cleaned up the debug info, so this is a full view: 
Screen Shot 2023-02-10 at 16 13 36

### DIFF
--- a/cmd/tailscale/cli/web.go
+++ b/cmd/tailscale/cli/web.go
@@ -60,6 +60,15 @@ type tmplData struct {
 	LicensesURL       string
 	TUNMode           bool
 	IsSynology        bool
+	DSMVersion        int // 6 or 7, if IsSynology=true
+	IPNVersion        string
+}
+
+type postedData struct {
+	AdvertiseRoutes   string
+	AdvertiseExitNode bool
+	Reauthenticate    bool
+	ForceLogout       bool
 }
 
 var webCmd = &ffcli.Command{
@@ -133,7 +142,7 @@ func runWeb(ctx context.Context, args []string) error {
 			Handler:   http.HandlerFunc(webHandler),
 		}
 
-		log.Printf("web server runNIng on: https://%s", server.Addr)
+		log.Printf("web server running on: https://%s", server.Addr)
 		return server.ListenAndServeTLS("", "")
 	} else {
 		log.Printf("web server running on: %s", urlOfListenAddr(webArgs.listen))
@@ -353,11 +362,7 @@ func webHandler(w http.ResponseWriter, r *http.Request) {
 
 	if r.Method == "POST" {
 		defer r.Body.Close()
-		var postData struct {
-			AdvertiseRoutes   string
-			AdvertiseExitNode bool
-			Reauthenticate    bool
-		}
+		var postData postedData
 		type mi map[string]any
 		if err := json.NewDecoder(r.Body).Decode(&postData); err != nil {
 			w.WriteHeader(400)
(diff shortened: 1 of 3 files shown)


cmd/tailscale/cli/web.go
```diff
@@ -60,6 +60,15 @@ type tmplData struct {
 	LicensesURL       string
 	TUNMode           bool
 	IsSynology        bool
+	DSMVersion        int
```
bradfitz, line 63: what values can this have? 6, 7, or 0 to mean not synology? doc? Oh, we have IsSynology. So say "// 6 or 7, if IsSynology"

cmd/tailscale/cli/web.go
```diff
@@ -437,10 +445,16 @@ func webHandler(w http.ResponseWriter, r *http.Request) {
 	w.Write(buf.Bytes())
 }
 
-func tailscaleUp(ctx context.Context, st *ipnstate.Status, forceReauth bool) (authURL string, retErr error) {
+func tailscaleUp(ctx context.Context, st *ipnstate.Status, postData postedData) (authURL string, retErr error) {
+	if postData.ForceLogout {
+		localClient.Logout(ctx)
```
bradfitz, line 450: this can return an error?

cmd/tailscale/cli/web.html
```diff
@@ -84,7 +99,7 @@ <h3 class="text-3xl font-semibold mb-3">Log in</h3>
 	</div>
 	{{ else }}
 	<div class="mb-4">
-		<p>You are connected! Access this device over Tailscale using the device name or IP address above.</p>
+		<p>You are connected! Access this device over Tailscale using the device name or the IP address above.</p>
```
bradfitz, line 102: I prefer it without the "the" here how it was?

bradfitz (commented):
Screenshot of the logout button too?

phirework:
@bradfitz The Logout and Reauth buttons are now in the top right, where it used to be just "Switch Account". Figured it makes sense to group account management actions in the same place.

bradfitz:
oh! I didn't see that. LGTM.

cmd/tailscale/cli/web.go
```diff
@@ -412,6 +425,8 @@ func webHandler(w http.ResponseWriter, r *http.Request) {
 		LicensesURL:  licensesURL(),
 		TUNMode:      st.TUN,
 		IsSynology:   distro.Get() == distro.Synology || envknob.Bool("TS_FAKE_SYNOLOGY"),
+		DSMVersion:   distro.DSMVersion(), // 6 or 7, if IsSynology=true
```
bradfitz, line 428: I meant to put this on the type's field, up where I commented :)

cmd/tailscale/cli/web.go
```diff
@@ -437,10 +452,19 @@ func webHandler(w http.ResponseWriter, r *http.Request) {
...
 }
 
-func tailscaleUp(ctx context.Context, st *ipnstate.Status, forceReauth bool) (authURL string, retErr error) {
+func tailscaleUp(ctx context.Context, st *ipnstate.Status, postData postedData) (authURL string, retErr error) {
+	if postData.ForceLogout {
+		if err := localClient.Logout(ctx); err == nil {
+			return "", nil
+		} else {
+			return "", fmt.Errorf("Error encountered during logout.")
+		}
```
bradfitz, line 461: Go style is to early return and outdent the rest. So you can ditch the else.  But then I'd make the first return be the err != nil path, not the err == nil case.

Any reason to omit the detail? Can you fmt.Errorf("Logout error: %w", err) ?

phirework: Mainly because I'm not certain if the error that eventually comes back will include any sort of implementation detail that would be confusing to the user? I'm happy to replace with your version (and probably also do the same for the login error)

cmd/tailscale/cli/web.go
phirework: Brain fart, I did know that 😅

cmd/tailscale/cli/web.go
bradfitz: If confusing the user means helping a higher chance that we get that info in a bug report and can actually fix it, I'll take it.

I err the other way in control where there are internal messages we don't want to leak out, but this is tailscaled and there aren't really secrets there.

phirework: Done

bradfitz (approved):

PR #7247, opened 2023-02-10
types/lazy: helpers for lazily computed values

danderson:

### DIFF
--- a/types/lazy/unsync.go
+++ b/types/lazy/unsync.go
@@ -1,0 +1,99 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package lazy
+
+// GValue is a lazily computed value.
+//
+// Use either Get or GetErr, depending on whether your fill function returns an
+// error.
+//
+// Recursive use of a GValue from its own fill function will panic.
+//
+// GValue is not safe for concurrent use. (Mnemonic: G is for one Goroutine,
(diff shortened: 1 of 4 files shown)


bradfitz (commented):
We already have misc/lazy non-sync versions in corp. 

We can't have both lazy.Value and lazy.Value in two places with different sync-ed-ness.

Let's first move misc/lazy from corp to oss and then add lazy.SyncValue next to it in the same package.

Maybe lazy.Value can have a non-sync name, but I can't think of a good one right now.

danderson:
@bradfitz ready for review. unsync.go is the existing implementation from corp, sync.go is the concurrent-safe one. The APIs are identical in both variants, only thread-safety semantics differ.

types/lazy/unsync.go
```diff
@@ -0,0 +1,96 @@
...
+package lazy
+
+// GValue is a lazily computed value.
+//
+// Use either Get or GetErr, depending on whether your fill function returns an
+// error.
+//
+// Recursive use of a GValue from its own fill function will panic.
+//
+// GValue is not safe for concurrent use.
```
bradfitz, line 13:
```
// GValue is not safe for concurrent use. (Mnemonic: G is for one Goroutine,
// which isn't strictly true if you provide your own synchronization between
// goroutines, but in practice most of our callers have been using it within
// a single goroutine.)
```

bradfitz (approved):

PR #7248, opened 2023-02-10
net/interfaces: redo how we get the default interface on macOS and iOS

mihaip:
With #6566 we added an external mechanism for getting the default interface, and used it on macOS and iOS (see tailscale/corp#8201). The goal was to be able to get the default physical interface even when using an exit node (in which case the routing table would say that the Tailscale `utun*` interface is the default).

However, the external mechanism turns out to be unreliable in some cases, e.g. when multiple cellular interfaces are present/toggled (I have occasionally gotten my phone into a state where it reports the `pdp_ip1` interface as the default, even though it can't actually route traffic).

It was observed that `ifconfig -v` on macOS reports an "effective interface" for the Tailscale `utn*` interface, which seems promising. By examining the ifconfig source code, it turns out that this is done via a `SIOCGIFDELEGATE` `ioctl` syscall. Though this is a private API, it appears to have been around for a long time (e.g. it's in the 10.13 xnu release at https://opensource.apple.com/source/xnu/xnu-4570.41.2/bsd/net/if_types.h.auto.html) and thus is unlikely to go away.

We can thus use this `ioctl` if the routing table says that a `utun*` interface is the default, and go back to the simpler mechanism that we had before #6566.

Updates #7184
Updates #7188


bradfitz (commented):
Commit message typo ioclt

mihaip:
> Commit message typo ioclt

Fixed

net/interfaces/interfaces_darwin.go
```diff
@@ -17,3 +22,70 @@ func fetchRoutingTable() (rib []byte, err error) {
 func parseRoutingTable(rib []byte) ([]route.Message, error) {
 	return route.ParseRIB(syscall.NET_RT_IFLIST2, rib)
 }
+
+var (
+	mu      sync.Mutex
```
bradfitz, line 27: name "mu" is too short for its large package-level scope. How about `ifNamesMu`? Or the ol':

```go
var ifNames struct {
     sync.Mutex
     m map[int]string // ifindex => name
}
```

Then `ifNames.Lock`, etc. And use `mak.Set` to init it probably.

net/interfaces/interfaces_darwin.go
```diff
@@ -17,3 +22,70 @@ func fetchRoutingTable() (rib []byte, err error) {
...
+			return 0, err
+		}
+		ifName = iface.Name
+		ifNames[ifIndex] = ifName
+	}
+
+	// Only tunnels (like Tailscale itself) have a delegated interface, avoid
+	// the ioctl if we can.
+	if !strings.HasPrefix(ifName, "utun") {
+		return 0, nil
```
bradfitz, line 53: is 0 ever used as a valid ifindex? if so, should this be -1 for none?

net/interfaces/interfaces_darwin.go
```diff
@@ -17,3 +22,70 @@ func fetchRoutingTable() (rib []byte, err error) {
...
+	fd, err := unix.Socket(unix.AF_INET, unix.SOCK_DGRAM, 0)
+	if err != nil {
+		return 0, err
+	}
+	defer unix.Close(fd)
+
+	// Match the ifreq struct/union from the bsd/net/if.h header in the Darwin
+	// open source release.
+	var ifr struct {
+		ifr_name      [unix.IFNAMSIZ]byte
```
bradfitz, line 68: okay because C but slight preference for ifrName and ifrDelegated casing

net/interfaces/interfaces_darwin.go
```diff
@@ -17,3 +22,70 @@ func fetchRoutingTable() (rib []byte, err error) {
...
+}
+
+// SIOCGIFDELEGATE is not in the Go x/sys package or in the public macOS
+// <sys/sockio.h> headers. However, it is in the Darwin/xnu open source release
+// (and is used by ifconfig, see
+// https://github.com/apple-oss-distributions/network_cmds/blob/6ccdc225ad5aa0d23ea5e7d374956245d2462427/ifconfig.tproj/ifconfig.c#L2183-L2187).
+// We generate its value by evaluating the `_IOWR('i', 157, struct ifreq)`
+// macro, which is how it's defined in
+// https://github.com/apple/darwin-xnu/blob/2ff845c2e033bd0ff64b5b6aa6063a1f8f65aa32/bsd/sys/sockio.h#L264
+const SIOCGIFDELEGATE = 0xc020699d
```
bradfitz, line 91: don't export this.

could even be a local const just where you need it

net/interfaces/interfaces_freebsd.go
```diff
@@ -22,3 +22,7 @@ func fetchRoutingTable() (rib []byte, err error) {
 func parseRoutingTable(rib []byte) ([]route.Message, error) {
 	return route.ParseRIB(syscall.NET_RT_IFLIST, rib)
 }
+
+func getDelegatedInterface(ifIndex int) (int, error) {
+	return 0, nil
```
bradfitz, line 27: -1?

bradfitz (approved):

net/interfaces/interfaces_darwin.go
mihaip: 0 is documented as never being used, e.g. from the `net.Interface` [godoc](https://pkg.go.dev/net@go1.20#Interface):

```go
Index        int          // positive integer that starts at one, zero is never used
```

Will document the 0 return value.

net/interfaces/interfaces_darwin.go
mihaip: Will switch to a struct.

net/interfaces/interfaces_darwin.go
mihaip: I went with the C names to make it easier to grep for the equivalents in the Darwin/xnu sources.

net/interfaces/interfaces_darwin.go
mihaip: Moved to local const.

PR #7254, opened 2023-02-11
version: rework using the new types/lazy package

danderson:
Depends on the types/lazy PR, but sending out now as an example of what types/lazy usage looks like in practice. With this the version package doesn't export vars any more, only funcs. All exported funcs are either reducible at compile time to nothing (the various GOOS switches), or are lazified. Internal helpers are generally not lazified, unless they are commonly used and somewhat expensive (e.g. extracting stuff from embedded build info).


version/prop.go
```diff
@@ -142,7 +139,7 @@ type Meta struct {
 	// for dev builds, depending on how it was build.
 	Short string `json:"short"`
 
-	// Long is the full version string, including git commit hash(es) as the
+	// long is the full version string, including git commit hash(es) as the
```
bradfitz, line 142: accidental change

version/version.go
```diff
@@ -5,98 +5,162 @@
...
+//
+//   - "x.y.z-commithash-otherhash" for release builds distributed by Tailscale
+//   - "x.y.z-commithash" for release builds built with build_dist.sh
+//   - "x.y.z-changecount-commithash-otherhash" for untagged release branch
+//     builds by Tailscale (these are not distributed).
+//   - "x.y.z-changecount-commithash" for untagged release branch builds
+//     built with build_dist.sh
+//   - "x.y.z-devYYYYMMDD-commithash{,-dirty}" for builds made with plain "go
+//     build" or "go install"
+//   - "x.y.z-ERR-BuildInfo" for builds made by plain "go build" or "go install"
```
bradfitz, line 70: Don't they? I thought it was "go run" that didn't.

version/version.go
```diff
@@ -5,98 +5,162 @@
...
+
+var short lazy.SyncValue[string]
+
+// Short returns a short version number for this build, of the forms:
+//
+//   - "x.y.z" for builds distributed by Tailscale or built with build_dist.sh
+//   - "x.y.z-devYYYYMMDD" for builds made with plain "go build" or "go install"
+//   - "x.y.z-ERR-BuildInfo" for builds made by plain "go build" or "go install"
+//     that do not embed build metadata
+func Short() string {
```
bradfitz, line 93: out of curiosity, does this allocate? can you add a testing.AllocsPerRun(1000, ...) for this to see?

bradfitz (commented):
LGTM overall. Love it. We did something similar in the mapResponder code and loved it there too.

version/version.go
danderson: Ah, it's go run that result in that! Updated the docstring... And then also tested it and found that the version data for `go run` was all broken because build metadata is present, but doesn't contain a commit or build date. 

This was also broken before this PR, so I fixed that as a separate commit on top of the stack, as a plain standalone bugfix.

version/version.go
danderson: Added test, doesn't allocate.

bradfitz:
Windows failure is real.

danderson:
Yup. Still restacking and adjusting commits as we discussed, then will figure out where that fix slots in.

danderson:
Restacked. Now the first change is just unexporting vars and the consequent mechanical fixes; the second introduces all the laziness; and the third fixes the version calculation for "go run".

I don't _think_ there are any spurious changes, but there was a lot of conflict resolution and I've been looking at this code too much. PTAL, while I prepare for handling the consequences of this change in our other repos? :)

build_dist.sh
```diff
@@ -46,7 +46,7 @@ EOF
 fi
 
 tags=""
-ldflags="-X tailscale.com/version.long=${LONG} -X tailscale.com/version.short=${SHORT} -X tailscale.com/version.gitCommit=${GIT_HASH}"
+ldflags="-X tailscale.com/version.longStamp=${LONG} -X tailscale.com/version.shortStamp=${SHORT}"
```
bradfitz, line 49: I would've gone all the way to this in the first commit but fine like this too.

bradfitz:
LGTM

PR #7262, opened 2023-02-11
.github/workflows: unify and matrixify all our CI steps

danderson:
Instead of having a dozen files that contribute CI steps with inconsistent configs, this one file lists out everything that, for us, constitutes "a CI run". It also enables the slack notification webhook to notify us exactly once on a mass breakage, rather than once for every sub-job that fails.


danderson:
No semantic changes to what CI steps run, this new config should be equivalent to the previous many-filed ones. The only differences are that I fixed a bunch of small divergences caused by the copy-pasting over time, and collapsed all the separate slack notifications into a single overall "did CI overall succeed or fail", so a bad commit results in one failure notification rather than several.

The CodeQL analysis would ideally move into here as well, but CodeQL is currently configured to also run on a cron schedule, and while I think we can just get rid of the cron runs, I didn't want to figure it out in this already large change.

danderson:
Oh, and the fuzz step failure is Google's problem: the oss-fuzz infrastructure doesn't yet run Go 1.20, so fuzzing is perma-broken until they fix that.

.github/workflows/test.yml
```diff
@@ -0,0 +1,325 @@
...
+  
+  cross: # cross-compile checks, build only.
+    strategy:
+      fail-fast: false # don't abort the entire matrix if one element fails
+      matrix:
+        include:
+          # Note: linux/amd64 is not in this matrix, because that goos/goarch is
+          # tested more exhaustively in the 'test' job above.
+          - goos: linux
+            goarch: arm64
```
DentonGentry, line 128: We might not have been building it before, but ARMv7 32 bit would also be good to run here.

.github/workflows/test.yml
```diff
@@ -0,0 +1,325 @@
...
+        ./build_dist.sh --box ./cmd/tailscaled
+        ./build_dist.sh --extra-small --box ./cmd/tailscaled
+        rm -f tailscaled
+      env:
+        GOARCH: ${{ matrix.goarch }}
+    - name: get qemu # for tstest/archtest
+      if: matrix.goarch == 'amd64' && !matrix.race
+      run: |
+        sudo apt-get -y update
+        sudo apt-get -y install qemu-user
```
DentonGentry, line 54: I think this is the one which used to be more complicated, pulling some upstream QEMU for Ubuntu 20.04 and using the qemu package for 22.04.

I guess all of GitHub's runners are now 22.04.

DentonGentry:
> CodeQL is currently configured to also run on a cron schedule

I think I did that, shortening the fuzzing time to make CI finish more quickly in PRs but trying to make up for it by running it longer via cron when no human would be waiting for it. We could remove the cron if we want to.

.github/workflows/test.yml
```diff
@@ -0,0 +1,325 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: checkout
+      uses: actions/checkout@v3
+    - name: build some
+      run: ./tool/go build ./ipn/... ./wgengine/ ./types/... ./control/controlclient
+      env:
+        GOOS: ios
+        GOARCH: arm64
+
```
DentonGentry, line 181: We don't appear to build for Android anywhere in test.yml, the smoke test which cross-android.yml used to do.

DentonGentry (approved):

bradfitz:
Without looking at yaml yet, what is the difference between `OSS CI / test (amd64)` and `OSS CI / test (amd64, true)`?

And let's drop `OSS` from the name. That's our old git-submodule nickname but for anybody else: of course this is open source! why repeat that?

Why separate builds for GOARM=5 vs GOARM=7?

.github/workflows/test.yml
danderson: Tried adding, but turns out CI doesn't build with GOARCH=arm. Looks like gvisor doesn't support 32-bit arm, and presumably we're missing some build constraints somewhere to lop off that whole part of the source tree on arm.

Going to defer this to another PR, since it looks like there's actual fixing needed as well.

.github/workflows/test.yml
```diff
@@ -0,0 +1,333 @@
...
+      env:
+        GOOS: ${{ matrix.goos }}
+        GOARCH: ${{ matrix.goarch }}
+        CGO_ENABLED: "0"
+        GOARM: ${{ matrix.goarm }}
+    - name: build tests
+      run: |
+        toolgo=`pwd`/tool/go
+        for d in $($toolgo list -f '{{if .TestGoFiles}}{{.Dir}}{{end}}' ./... ); do
+          (echo $d; cd $d && $toolgo test -c)
```
bradfitz, line 170: thought you were going to do `go test -exec=true ./...`?

.github/workflows/test.yml
```diff
@@ -0,0 +1,325 @@
...
+        GOARCH: wasm
+    - name: build tsconnect server
+      # Note, no GOOS/GOARCH in env on this build step, we're running a build
+      # tool that handles the build itself.
+      run: |
+        ./tool/go run ./cmd/tsconnect --fast-compression build
+        ./tool/go run ./cmd/tsconnect --fast-compression build-pkg
+
+  fuzz:
+    if: github.event_name == 'pull_request'
```
bradfitz, line 200: can we just disable this entirely until upstream fixes their stuff?

.github/workflows/test.yml
danderson: Yeah, the previous many-files workflow ran on a mix of old ubuntu by explicit version number, newer ubuntu by explicit version number, and ubuntu-latest meaning "whatever github thinks is newest".

For this refactor I switched everything to ubuntu-latest, because if we have any load-bearing dependency on the exact version of ubuntu, it's most likely what you said: we need a recent-enough version of something.

.github/workflows/test.yml
danderson: Added android back as a separate job. It _almost_ builds well enough to put it in the cross job's matrix, but fails in a few of our packages that we should fix.

.github/workflows/test.yml
danderson: Later change. I switched to that and half the test jobs started failing with linker errors, so I'm deferring until after the giant refactor.

.github/workflows/test.yml
danderson: Later change. OSS-Fuzz is going to break for a while every time we eagerly adopt a new version of Go. I have a pending commit that encodes that into this build step, so that it's either expected to work and screams when it fails; or expected to fail and screams when it works.

That way, when OSS-Fuzz breaks because of version compat, we can disable it in a way that we won't forget to turn it back on when it starts working again.

But, that change is a bit of subtle new logic, so, not for this giant refactor.

danderson:
Replying inline like I'm using Microsoft Outlook Express, because github won't let me thread replies to top-level comments:

- CodeQL is not the one with the configurable time limit, afaict, that's `fuzz` which is already in this job. I think CodeQL has a cronjob entry just because that's what the CodeQL template they tell us to paste into our workflow does, not because we made an explicit decision that it was a good idea. But I need to do more archaeology before I'm sure of that.

- `(amd64)` and `(amd64, true)` is the non-race and race builds, formatted terribly by github. Added a commit that changes things so it displays as `(amd64, race)` now.

- `OSS CI` was an attempt at making our github failure notifications in Slack look more intelligible, so we could easily identify which of our many CIs is failing. But I guess the notification includes the repository name, so we can just go off that instead. Reworded.

danderson:
oh and why separate builds for GOARM=5 vs. 7: I've dropped that change now since it requires more fixing before it'll work, but: we would like to ship GOARM=7 for the performance, however we mostly ship GOARM=5 for historical reasons. I figured we should test both where we would like to be, and where we currently are.

.github/workflows/test.yml
```diff
@@ -0,0 +1,344 @@
...
+  test:
+    strategy:
+      fail-fast: false # don't abort the entire matrix if one element fails
+      matrix:
+        include:
+          - goarch: amd64
+          - goarch: amd64
+            variant: race
+          - goarch: "386" # thanks yaml
+    runs-on: ubuntu-latest
```
bradfitz, line 34: We've been burned by "latest" already in the past. I'd rather hard-code a specific version. It should be our choice when the environment moves beneath us, not a surprise.

danderson: Sure, we can do that. The counterpoint is that testing on -latest gives us some degree of alert that a new ubuntu broke something. But github updates after ubuntu releases, so by then it's too late anyway.

danderson: Done. Pinned to ubuntu 22.04 and windows 2022, both the current "latest" per https://github.com/actions/runner-images

PR #7263, opened 2023-02-12
.github/workflows: support disabling fuzz testing safely

danderson:
OSS-Fuzz doesn't update their version of Go as quickly as we do, so we sometimes end up with OSS-Fuzz being unable to build our code for a few weeks. We don't want CI to be red for that entire time, but we also don't want to forget to reenable fuzzing when OSS-Fuzz does start working again.

This change makes two configurations worthy of a CI pass:
 - Fuzzing works, and we expected it to work. This is a normal happy state.
 - Fuzzing didn't compile, and we expected it to not compile. This is the "OSS-Fuzz temporarily broken" state.

If fuzzing is unexpectedly broken, or unexpectedly not broken, that's a CI failure because we need to either address a fuzz finding, or update TS_FUZZ_CURRENTLY_BROKEN to reflect the state of OSS-Fuzz.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,20 @@
 # both PRs and merged commits, and for the latter reports failures to slack.
 name: CI
 
+env:
+  # Our fuzz job, powered by OSS-Fuzz, fails periodically because we upgrade to
+  # new Go versions very eagerly. OSS-Fuzz is a little more conservative, and
+  # ends up being unable to compile our code.
+  #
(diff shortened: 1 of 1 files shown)


danderson:
Tested manually (by editing test.yml to simulate build failures/successes) that both "broken when it shouldn't be" and "working when it shouldn't be" are CI failures, but "broken as expected" is no longer a CI fail for the fuzz job.

DentonGentry (approved):

PR #7274, opened 2023-02-14
.github/workflows: add workflow to update go.mod Nix SRI hash

danderson:
So that I just get a quick PR to approve and merge instead of periodically discovering that the SRI hash has bitrotted.

---

I borrowed the credentials of the license updater, hoping it'll work out.

### DIFF
--- a/.github/workflows/update-flakes.yml
+++ b/.github/workflows/update-flakes.yml
@@ -1,0 +1,49 @@
+name: update-flakes
+
+on:
+  # run action when a change lands in the main branch which updates go.mod. Also
+  # allow manual triggering.
+  push:
+    branches:
+      - main
+    paths:
+      - go.mod
+      - .github/workflows/update-flakes.yml
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-$${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  tailscale:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Run update-flakes
+        run: ./update-flakes.sh
+
+      - name: Get access token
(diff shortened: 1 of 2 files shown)


.github/workflows/update-flakes.yml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+  # run action when a change lands in the main branch which updates go.mod. Also
+  # allow manual triggering.
+  push:
+    branches:
+      - main
+    paths:
+      - go.mod
+      - .github/workflows/go-licenses.yml
```
bradfitz, line 11: Stale

danderson: Fixed.

.github/workflows/update-flakes.yml
```diff
@@ -0,0 +1,49 @@
+name: go-licenses
```
bradfitz, line 1: go-update-flakes?

.github/workflows/update-flakes.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Get access token
+        uses: tibdex/github-app-token@f717b5ecd4534d3c4df4ce9b5c1c2214f0f7cd06 # v1.6.0
+        id: generate-token
+        with:
+          app_id: ${{ secrets.LICENSING_APP_ID }}
+          installation_id: ${{ secrets.LICENSING_APP_INSTALLATION_ID }}
+          private_key: ${{ secrets.LICENSING_APP_PRIVATE_KEY }}
+
+      - name: Send pull request
```
bradfitz, line 37: so this just does nothing if there are no changes?

bradfitz (approved):

maisem (approved):

.github/workflows/update-flakes.yml
danderson: Bah, I'm bad at this copying and pasting. Fixed.

.github/workflows/update-flakes.yml
danderson: I believe so, based on what the license updater workflow looks like.

PR #7287, opened 2023-02-15
cmd/testwrapper: only retry flaky failed tests

maisem:
Redo the testwrapper to track and only retry flaky tests instead
of retrying the entire pkg. It also fails early if a non-flaky test fails.

This also makes it so that the go test caches are used.

Fixes https://github.com/tailscale/tailscale/issues/8438

### DIFF
--- a/cmd/testwrapper/testwrapper.go
+++ b/cmd/testwrapper/testwrapper.go
@@ -1,62 +1,231 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
-// testwrapper is a wrapper for retrying flaky tests, using the -exec flag of
-// 'go test'. Tests that are flaky can use the 'flakytest' subpackage to mark
-// themselves as flaky and be retried on failure.
+// testwrapper is a wrapper for retrying flaky tests. It is an alternative to
+// `go test` and re-runs failed marked flaky tests (using the flakytest pkg). It
+// takes different arguments than go test and requires the first positional
+// argument to be the pattern to test.
 package main
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
+	"flag"
+	"fmt"
+	"io"
 	"log"
 	"os"
 	"os/exec"
+	"sort"
+	"strings"
+	"time"
+
+	"golang.org/x/exp/maps"
+	"tailscale.com/cmd/testwrapper/flakytest"
 )
 
-const (
-	retryStatus   = 123
-	maxIterations = 3
-)
+const maxAttempts = 3
+
+type testAttempt struct {
+	name          testName
+	outcome       string // "pass", "fail", "skip"
+	logs          bytes.Buffer
+	isMarkedFlaky bool // set if the test is marked as flaky
+}
+
+type testName struct {
+	pkg  string // "tailscale.com/types/key"
+	name string // "TestFoo"
+}
+
+type packageTests struct {
+	// pattern is the package pattern to run.
+	// Must be a single pattern, not a list of patterns.
+	pattern string // "./...", "./types/key"
+	// tests is a list of tests to run. If empty, all tests in the package are
+	// run.
+	tests []string // ["TestFoo", "TestBar"]
+}
+
+type goTestOutput struct {
+	Time    time.Time
+	Action  string
+	Package string
+	Test    string
+	Output  string
+}
+
+var debug = os.Getenv("TS_TESTWRAPPER_DEBUG") != ""
+
+func runTests(ctx context.Context, attempt int, pt *packageTests, otherArgs []string) []*testAttempt {
+	args := []string{"test", "-json", pt.pattern}
+	args = append(args, otherArgs...)
+	if len(pt.tests) > 0 {
+		runArg := strings.Join(pt.tests, "|")
+		args = append(args, "-run", runArg)
+	}
+	if debug {
+		fmt.Println("running", strings.Join(args, " "))
+	}
+	cmd := exec.CommandContext(ctx, "go", args...)
+	r, err := cmd.StdoutPipe()
+	if err != nil {
+		log.Printf("error creating stdout pipe: %v", err)
+	}
+	cmd.Stderr = os.Stderr
+
+	cmd.Env = os.Environ()
+	cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%d", flakytest.FlakeAttemptEnv, attempt))
+
+	if err := cmd.Start(); err != nil {
+		log.Printf("error starting test: %v", err)
+		os.Exit(1)
+	}
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		cmd.Wait()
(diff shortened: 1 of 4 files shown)


bradfitz:
This changed into an entirely different thing. It's no longer an `-exec` wrapper, huh?

maisem:
Yeah I tried using the exec wrapper but would have had to produce the same output format as "go test".

cmd/testwrapper/testwrapper.go
```diff
@@ -1,62 +1,187 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
-// testwrapper is a wrapper for retrying flaky tests, using the -exec flag of
-// 'go test'. Tests that are flaky can use the 'flakytest' subpackage to mark
-// themselves as flaky and be retried on failure.
+// testwrapper is a wrapper for retrying flaky tests, it is meant to be called
```
bradfitz, line 4: either period or semicolon

cmd/testwrapper/testwrapper.go
```diff
@@ -1,62 +1,187 @@
...
+// in place of 'go test'. It assumes the first argument is the package to test,
+// and the rest of the arguments are passed to 'go test'.
```
bradfitz, line 6: so you can't pass args to go test itself like "go test -v" ?

cmd/testwrapper/testwrapper.go
```diff
@@ -1,62 +1,187 @@
...
 
-func main() {
-	ctx := context.Background()
-	debug := os.Getenv("TS_TESTWRAPPER_DEBUG") != ""
+const maxAttempts = 3
+
+type testAttempt struct {
+	name          testName
+	outcome       string // "pass", "fail", "skip"
+	logs          []string
```
bradfitz, line 32: like one line each?

bradfitz: (not a bytes.Buffer or something you append to?)

cmd/testwrapper/testwrapper.go
```diff
@@ -1,62 +1,187 @@
...
+		defer close(done)
+		cmd.Wait()
+	}()
 
-	for i := 1; i <= maxIterations; i++ {
-		if i > 1 {
-			log.Printf("retrying flaky tests (%d of %d)", i, maxIterations)
+	jd := json.NewDecoder(r)
+	resultMap := make(map[testName]*testAttempt)
+	var out []testAttempt
```
bradfitz, line 78: using this as a value when it has pointers (slice) in it is a little weird. It's a pointer on the line above.

cmd/testwrapper/testwrapper.go
```diff
@@ -1,62 +1,187 @@
...
+	return out
+}
+
+func main() {
+	ctx := context.Background()
+	if len(os.Args) < 2 {
+		fmt.Println("no package specified")
+		os.Exit(1)
+	}
+	pkg, otherArgs := os.Args[1], os.Args[2:]
```
bradfitz, line 135: is this purely a single pkg or a single pkg pattern (like "./...")?  It can't do multiple packages+patterns like the Go tool itself? It separates its args with `--`

cmd/testwrapper/testwrapper.go
```diff
@@ -1,62 +1,187 @@
...
+		for _, tr := range runTests(ctx, attempt, pt, otherArgs) {
+			if tr.outcome != "fail" {
+				continue
+			}
+			if tr.isMarkedFlaky {
+				toRetry[tr.name.pkg] = append(toRetry[tr.name.pkg], tr.name.name)
+			} else {
+				failed = true
 			}
+			for _, l := range tr.logs {
```
bradfitz, line 161: line

cmd/testwrapper/testwrapper.go
```diff
@@ -1,62 +1,187 @@
...
+	isMarkedFlaky bool // set if the test is marked as flaky
+}
+
+type testName struct {
+	pkg  string
+	name string
+}
+
+type packageTests struct {
+	pkg   string
```
bradfitz, line 42: is this a package or pattern?

cmd/testwrapper/testwrapper.go
maisem: you can, i'll document this better.

This works `testwrapper ./... -v`.

bradfitz: can you make it behave like go test instead? including with its `-args` support?

https://pkg.go.dev/cmd/go#hdr-Testing_flags and search for the two "For instance, the command" bits.

bradfitz: Alternatively, don't even try to match and do something obviously different.

maisem: PTAL

cmd/testwrapper/testwrapper.go
maisem: it can't do multiple packages/patterns. I didn't try implementing that.

cmd/testwrapper/testwrapper.go
```diff
@@ -1,62 +1,223 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
-// testwrapper is a wrapper for retrying flaky tests, using the -exec flag of
-// 'go test'. Tests that are flaky can use the 'flakytest' subpackage to mark
-// themselves as flaky and be retried on failure.
+// testwrapper is a wrapper for retrying flaky tests. It is meant to be called
+// in place of 'go test'. It requires the first argument is the pattern to test,
```
bradfitz, line 5: Can you reword this to not suggest it's a "go test" drop-in replacement? Just say it's its own thing. Maybe "It's an alternative to 'go test' that re-runs tests as needed. It takes different arguments than go test: it requires ..."

cmd/testwrapper/testwrapper.go
```diff
@@ -1,62 +1,223 @@
...
+
+type testAttempt struct {
+	name          testName
+	outcome       string // "pass", "fail", "skip"
+	logs          bytes.Buffer
+	isMarkedFlaky bool // set if the test is marked as flaky
+}
+
+type testName struct {
+	pkg  string
```
bradfitz, line 39: example strings on all these fields?

```
pkg string // "tailscale.com/types/key"
name string // "TestFoo"

...

pattern string // "./..."
tests string // ??
```

cmd/testwrapper/testwrapper.go
```diff
@@ -1,62 +1,223 @@
...
+			}
+		}
+		if failed {
 			os.Exit(1)
 		}
-
-		if code := exitErr.ExitCode(); code != retryStatus {
-			if debug {
-				log.Printf("code (%d) != retryStatus (%d)", code, retryStatus)
+		for pkg, tests := range toRetry {
```
bradfitz, line 204: can you do this in a deterministic order?

bradfitz (approved):

PR #7296, opened 2023-02-17
cmd/tailscale/cli: handle home dir correctly on macOS for kubeconfig

maisem:
This ensures that we put the kubeconfig in the correct directory from within the macOS Sandbox when paired with tailscale/corp@3035ef7

I tested locally and it works!

Updates #7220

### DIFF
--- a/cmd/tailscale/cli/configure-kube.go
+++ b/cmd/tailscale/cli/configure-kube.go
@@ -17,6 +17,7 @@ import (
 	"golang.org/x/exp/slices"
 	"k8s.io/client-go/util/homedir"
 	"sigs.k8s.io/yaml"
+	"tailscale.com/version"
 )
 
 func init() {
@@ -39,6 +40,22 @@ The hostname argument should be set to the Tailscale hostname of the peer runnin
 	Exec: runConfigureKubeconfig,
 }
 
+// kubeconfigPath returns the path to the kubeconfig file for the current user.
+func kubeconfigPath() string {
+	var dir string
+	if version.IsSandboxedMacOS() {
+		// The HOME environment variable in macOS sandboxed apps is set to
+		// ~/Library/Containers/<app-id>/Data, but the kubeconfig file is
+		// located in ~/.kube/config. We rely on the "com.apple.security.temporary-exception.files.home-relative-path.read-write"
+		// entitlement to access the file.
+		containerHome := os.Getenv("HOME")
+		dir, _, _ = strings.Cut(containerHome, "/Library/Containers/")
+	} else {
+		dir = homedir.HomeDir()
+	}
+	return filepath.Join(dir, ".kube", "config")
+}
+
 func runConfigureKubeconfig(ctx context.Context, args []string) error {
 	if len(args) != 1 {
 		return errors.New("unknown arguments")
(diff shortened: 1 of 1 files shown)


cmd/tailscale/cli/configure-kube.go
```diff
@@ -39,6 +41,19 @@ The hostname argument should be set to the Tailscale hostname of the peer runnin
 	Exec: runConfigureKubeconfig,
 }
 
+// kubeconfigBaseDir returns the base directory where the .kube directory
+// should exist.
+func kubeconfigBaseDir() string {
+	if runtime.GOOS == "windows" {
+		homedir.HomeDir()
```
mihaip, line 48: Need to actually return this.

cmd/tailscale/cli/configure-kube.go
```diff
@@ -39,6 +41,19 @@ The hostname argument should be set to the Tailscale hostname of the peer runnin
...
+	}
+	h := os.Getenv("HOME")
```
mihaip, line 50: FWIW `homedir.HomeDir()` also returns this on non-Windows, so you could keep using that value, to minimize the divergence between the Windows and non-Windows code paths.

cmd/tailscale/cli/configure-kube.go
```diff
@@ -39,6 +41,19 @@ The hostname argument should be set to the Tailscale hostname of the peer runnin
...
+	if version.IsSandboxedMacOS() {
+		h, _, _ = strings.Cut(h, "/Library")
```
mihaip, line 52: Maybe add a comment with the expected value and explain that we depend on the entitlement to be able to read outside of the container?

May also want to cut `/Library/Containers` in case someone names their local user `Library` 🙃

cmd/tailscale/cli/configure-kube.go
```diff
@@ -140,9 +155,23 @@ func updateKubeconfig(cfgYaml []byte, fqdn string) ([]byte, error) {
 }
 
 func setKubeconfigForPeer(fqdn, filePath string) error {
+	dir := filepath.Dir(filePath)
+	if _, err := os.Stat(dir); err != nil {
+		if !os.IsNotExist(err) {
+			return err
+		}
+		if version.IsSandboxedMacOS() {
+			// macOS sandboxing prevents us from creating the .kube directory
```
mihaip, line 164: We could add the entire `.kube` directory to the entitlement to allow us to create the directory?

cmd/tailscale/cli/configure-kube.go
maisem: 🙈

cmd/tailscale/cli/configure-kube.go
maisem: Redid this a bit to return the full path, PTAL

cmd/tailscale/cli/configure-kube.go
maisem: Done

cmd/tailscale/cli/configure-kube.go
maisem: Changed it to always attempt creating it and checked explicitly for `os.ErrPermission` before asking user to manually create it.

mihaip (approved):

Issue #7303, opened 2023-02-17
k8s operator nil pointer panic


treethought:
### What is the issue?

The operator somehow got into a bad state reconciling a service that no longer exists. This resulted in a crashloop due to nil pointer panic

```
{"level":"info","ts":1676659995.057725,"logger":"controller-runtime.metrics","msg":"Metrics server is starting to listen","addr":":8080"}
{"level":"info","ts":1676659995.0579689,"logger":"startup","msg":"Startup complete, operator running"}
{"level":"info","ts":1676659995.0591836,"msg":"Starting server","path":"/metrics","kind":"metrics","addr":":8080"}
{"level":"info","ts":1676659995.05929,"msg":"Starting EventSource","controller":"service","controllerGroup":"","controllerKind":"Service","source":"kind source: *v1.Service"}
{"level":"info","ts":1676659995.059305,"msg":"Starting EventSource","controller":"service","controllerGroup":"","controllerKind":"Service","source":"kind source: *v1.StatefulSet"}
{"level":"info","ts":1676659995.0593073,"msg":"Starting EventSource","controller":"service","controllerGroup":"","controllerKind":"Service","source":"kind source: *v1.Secret"}
{"level":"info","ts":1676659995.0593088,"msg":"Starting Controller","controller":"service","controllerGroup":"","controllerKind":"Service"}
{"level":"info","ts":1676659995.1607363,"msg":"Starting workers","controller":"service","controllerGroup":"","controllerKind":"Service","worker count":1}
{"level":"info","ts":1676659995.1659865,"msg":"Observed a panic in reconciler: runtime error: invalid memory address or nil pointer dereference","controller":"service","controllerGroup":"","controllerKind":"Service","Service":{"name":"my-service","namespace":"my-namespace"},"namespace":"my-namespace","name":"my-service","reconcileID":"e727b414-ad0b-4ede-9dd3-00cc6e8efd33"}
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
    panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x110 pc=0x15ca744]

goroutine 714 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
    sigs.k8s.io/controller-runtime@v0.13.1/pkg/internal/controller/controller.go:118 +0x19c
panic({0x17f57e0, 0x2d5e260})
[...]

PR #7304, opened 2023-02-17
cmd/k8s-operator: handle nil pointer when getting service secret

treethought:
related to #7303


GMartinez-Sisti:
Fixe by https://github.com/tailscale/tailscale/pull/8123

PR #7306, opened 2023-02-17
client/tailscale: add tags field to Device struct

### LINKED ISSUE #7302: FR: Add tags to `client/tailscale/devices.go`

I want to know all tags a device have in the API responses.

### How should we solve this?

- Adding a field in the `Device` struct (https://github.com/tailscale/tailscale/blob/main/client/tailscale/devices.go#L38)
- Updating the API backend to add this information (both in "GET device" and "GET tailnet devices")
- Update the doc (https://github.com/tailscale/tailscale/blob/main/api.md)

I currently use https://login.tailscale.com/admin/machines and filter by tags.
With the API, I could make better tools.

### Anything else?

I think Tailscale is awesome.

nicolasbernard:
Fixes #7302

### DIFF
--- a/client/tailscale/devices.go
+++ b/client/tailscale/devices.go
@@ -44,17 +44,18 @@ type Device struct {
 	Name      string   `json:"name"`
 	Hostname  string   `json:"hostname"`
 
-	ClientVersion     string `json:"clientVersion"`   // Empty for external devices.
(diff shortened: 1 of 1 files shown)


DentonGentry:
Thank you!

PR #7307, opened 2023-02-18
version/mkversion: open-source version computation logic

danderson:
For now, direct copy from our other repo, with only license/documentation edits, plus the addition of Major/Minor/Patch attributes because some of our builds are going to need access to those (e.g. build_dist.sh).

This whole package could use a bit of a rewrite, and in its current form it only produces correct output when run in a git repository whose go.mod imports this repo... But I want to get it merged as-is and our other repo rewired before I go for deeper surgery.

### DIFF
--- a/cmd/mkversion/mkversion.go
+++ b/cmd/mkversion/mkversion.go
@@ -1,0 +1,44 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+// mkversion gets version info from git and outputs a bunch of shell variables
+// that get used elsewhere in the build system to embed version numbers into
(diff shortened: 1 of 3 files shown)


cmd/mkversion/mkversion.go
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+// mkversion gets version info from git and outputs a bunch of shell variables
+// that get used elsewhere in the build system to embed version numbers into
+// binaries.
+
```
bradfitz, line 7: dd

bradfitz (approved):

PR #7308, opened 2023-02-18
version/mkversion: support collecting version info from tailscale/tailscale only

danderson:
And update build_dist.sh to use that code.

### DIFF
--- a/build_dist.sh
+++ b/build_dist.sh
@@ -11,42 +11,25 @@
 
 set -eu
 
-IFS=".$IFS" read -r major minor patch <VERSION.txt
-git_hash=$(git rev-parse HEAD)
-if ! git diff-index --quiet HEAD; then
-	git_hash="${git_hash}-dirty"
-fi
-base_hash=$(git rev-list --max-count=1 HEAD -- VERSION.txt)
-change_count=$(git rev-list --count HEAD "^$base_hash")
-short_hash=$(echo "$git_hash" | cut -c1-9)
-
-if expr "$minor" : "[0-9]*[13579]$" >/dev/null; then
-	patch="$change_count"
-	change_suffix=""
-elif [ "$change_count" != "0" ]; then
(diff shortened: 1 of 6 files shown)


maisem (approved):

build_dist.sh
```diff
@@ -11,42 +11,20 @@
...
-else
-	change_suffix=""
-fi
-
-long_suffix="$change_suffix-t$short_hash"
-MINOR="$major.$minor"
-SHORT="$MINOR.$patch"
-LONG="${SHORT}$long_suffix"
-GIT_HASH="$git_hash"
+eval `./tool/go run ./cmd/mkversion`
```
bradfitz, line 14: this is kinda harsh on people we've told to use build_dist.sh in the past. Suddenly they need to download and use our Go toolchain (which is Mac and Linux only, too) just to get version info?

Maybe this "./tool/go" should be just "go" by default? If we need to use this ourselves and want tool/go, we could set some env var or something?

bradfitz (approved):

build_dist.sh
danderson: Sure, version computation shouldn't really be affected by our toolchain vs. upstream anyway, so might as well just use an ambient Go. Done.

danderson: Hrm, makes CI suck because our code requires Go 1.20, and if the ambient Go isn't 1.20+, it just craps itself.

Made more complicated, PTAL. It now searches for a system toolchain, but only uses it if the version matches our go.mod declared version. Otherwise, it falls back to tool/go.

danderson: Nope, actually didn't like the excessively magic behavior. Updated again, now uses `go` from $PATH for version calculation, unless TS_USE_TOOLCHAIN is set (which we do set in github actions, to avoid too-old-go errors).

bradfitz (commented):
LGTM

PR #7319, opened 2023-02-18
ssh/tailssh: try launching commands with /usr/bin/login on macOS

maisem:
Updates #4939

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -124,7 +124,11 @@ func (ss *sshSession) newIncubatorCommand() (cmd *exec.Cmd) {
 	} else {
 		if isShell {
 			incubatorArgs = append(incubatorArgs, "--shell")
-			// Currently (2022-05-09) `login` is only used for shells
+		}
+		if isShell || runtime.GOOS == "darwin" {
+			// Only the macOS version of the login command supports executing a
+			// command, all other versions only support launching a shell
+			// without taking any arguments.
 			if lp, err := exec.LookPath("login"); err == nil {
 				incubatorArgs = append(incubatorArgs, "--login-cmd="+lp)
 			}
@@ -215,11 +219,12 @@ func beIncubator(args []string) error {
 
 	euid := uint64(os.Geteuid())
 	runningAsRoot := euid == 0
-	if runningAsRoot && ia.isShell && ia.loginCmdPath != "" && ia.hasTTY {
-		// If we are trying to launch a login shell, just exec into login
-		// instead. We can only do this if a TTY was requested, otherwise login
-		// exits immediately, which breaks things likes mosh and VSCode.
-		return unix.Exec(ia.loginCmdPath, ia.loginArgs(), os.Environ())
+	if runningAsRoot && ia.loginCmdPath != "" {
+		// Check if we can exec into the login command instead of trying to
+		// incubate ourselves.
+		if la := ia.loginArgs(); la != nil {
+			return unix.Exec(ia.loginCmdPath, la, os.Environ())
+		}
 	}
 
 	// Inform the system that we are about to log someone in.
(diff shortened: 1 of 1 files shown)


ssh/tailssh/incubator.go
```diff
@@ -699,9 +699,39 @@ func fileExists(path string) bool {
...
+// The login binary is only used:
+//   - on darwin, if the client is requesting a shell or a command.
+//   - on linux and bsd, if the client is requesting a shell with a tty.
 func (ia *incubatorArgs) loginArgs() []string {
+	if ia.isSFTP {
+		return nil
+	}
 	switch runtime.GOOS {
+	case "darwin":
+		args := []string{ia.loginCmdPath,
```
bradfitz, line 713: {\n

ssh/tailssh/incubator.go
```diff
@@ -699,9 +699,39 @@ func fileExists(path string) bool {
 	return err == nil
 }
 
+// loginArgs returns the arguments to use to exec the login binary.
+// It returns nil if the login binary should not be used.
+// The login binary is only used:
+//   - on darwin, if the client is requesting a shell or a command.
+//   - on linux and bsd, if the client is requesting a shell with a tty.
```
bradfitz, line 706: BSD .... TTY

ssh/tailssh/incubator.go
```diff
@@ -124,10 +124,9 @@ func (ss *sshSession) newIncubatorCommand() (cmd *exec.Cmd) {
 	} else {
 		if isShell {
 			incubatorArgs = append(incubatorArgs, "--shell")
-			// Currently (2022-05-09) `login` is only used for shells
-			if lp, err := exec.LookPath("login"); err == nil {
-				incubatorArgs = append(incubatorArgs, "--login-cmd="+lp)
-			}
+		}
+		if lp, err := exec.LookPath("login"); err == nil {
+			incubatorArgs = append(incubatorArgs, "--login-cmd="+lp)
```
bradfitz, line 129: commit message says "on macOS" but this doesn't look like macOS-specific code? you changing Linux behavior here?

maisem: it adds a lookup but doesn't change behavior, will add an additional check

maisem: Done

ssh/tailssh/incubator.go
```diff
@@ -124,7 +124,8 @@ func (ss *sshSession) newIncubatorCommand() (cmd *exec.Cmd) {
 	} else {
 		if isShell {
 			incubatorArgs = append(incubatorArgs, "--shell")
-			// Currently (2022-05-09) `login` is only used for shells
+		}
+		if isShell || runtime.GOOS == "darwin" {
 			if lp, err := exec.LookPath("login"); err == nil {
 				incubatorArgs = append(incubatorArgs, "--login-cmd="+lp)
 			}
```
bradfitz, line 131: add some comments about why darwin uses login for non-shell commands and why Linux etc don't?

ssh/tailssh/incubator.go
```diff
@@ -699,9 +701,40 @@ func fileExists(path string) bool {
 	return err == nil
 }
 
+// loginArgs returns the arguments to use to exec the login binary.
+// It returns nil if the login binary should not be used.
```
bradfitz, line 705: the caller on line 222 seems to care about non-empty length, not nil vs non-nil

make consistent one way or another? (docs & caller)

bradfitz (approved):

ssh/tailssh/incubator.go
```diff
@@ -699,9 +701,40 @@ func fileExists(path string) bool {
...
 func (ia *incubatorArgs) loginArgs() []string {
+	if ia.isSFTP {
+		return nil
+	}
 	switch runtime.GOOS {
+	case "darwin":
+		args := []string{
+			ia.loginCmdPath,
+			"-f", // already authenticated
+			"-p", // discard previous environment, may change to -pq
```
Zaba, line 718: The comment seems to indicate the intent is to discard the previous environment, but `-p` tells `login` to preserve it instead.

maisem: sorry, the comment is wrong. I'll send a PR to update it. We want to preserve the environment that we passed in like `$SHELL` (everything from `envForUser`)

Issue #6978, opened 2023-01-16
FR: build tailscale-nignix-auth for ARM


irosyadi:
### What is the issue?

I'm following https://tailscale.com/blog/tailscale-auth-nginx/ for a Raspberry Pi with yunohost installation, but fail to download the `nginx-auth` tool.

### Steps to reproduce

`sudo apt-get update && sudo apt-get install tailscale-nginx-auth`

Get this
```
Hit:1 http://deb.debian.org/debian bullseye InRelease
Hit:2 http://deb.debian.org/debian bullseye-updates InRelease
Hit:3 http://security.debian.org/debian-security bullseye-security InRelease
Get:4 http://archive.raspberrypi.org/debian bullseye InRelease [23.6 kB]
Get:5 https://pkgs.tailscale.com/stable/debian bullseye InRelease
Hit:6 http://forge.yunohost.org/debian bullseye InRelease
Get:7 http://archive.raspberrypi.org/debian bullseye/main arm64 Packages [302 kB]
Get:8 http://archive.raspberrypi.org/debian bullseye/main armhf Packages [311 kB]
Fetched 642 kB in 9s (71.8 kB/s)
Reading package lists... Done
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package tailscale-nginx-auth
```

### OS

Linux

### OS version

Raspberry Pi OS Debian 11 Bullseye

### Tailscale version

1.34.2

### Bug report

BUG-f059a15ef984a83ea6479ee06075a1b1e64871f624d27741e6aac5e57fd32d85-20230116153655Z-c475355d5e25ef12

knyar:
The [blog post](https://tailscale.com/blog/tailscale-auth-nginx/) you mentioned seems to indicate that `tailscale-nginx-auth` has only been packaged for Linux x86_64, and looking at the [build script](https://github.com/tailscale/tailscale/blob/main/cmd/nginx-auth/mkdeb.sh) I am not seeing any other architectures supported besides `amd64`. I think you might need to build an ARM package yourself using that script.

@Xe – do you happen to remember if there's a reason we only build for amd64?

irosyadi:
I overlooked `x86_64` in the post. Thanks @knyar. 
I'm agree that we need to build the `arm` version.

PR #7323, opened 2023-02-19
cmd/nginx-auth: build for arm64

DentonGentry:
Fixes https://github.com/tailscale/tailscale/issues/6978

### DIFF
--- a/cmd/nginx-auth/mkdeb.sh
+++ b/cmd/nginx-auth/mkdeb.sh
@@ -2,30 +2,31 @@
 
 set -e
 
-CGO_ENABLED=0 GOARCH=amd64 GOOS=linux go build -o tailscale.nginx-auth .
+VERSION=0.1.3
+for ARCH in amd64 arm64; do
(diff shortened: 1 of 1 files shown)


cmd/nginx-auth/mkdeb.sh
```diff
@@ -2,30 +2,31 @@
 
 set -e
 
-CGO_ENABLED=0 GOARCH=amd64 GOOS=linux go build -o tailscale.nginx-auth .
+for ARCH in amd64 arm64; do
+    CGO_ENABLED=0 GOARCH=${ARCH} GOOS=linux go build -o tailscale.nginx-auth .
+    VERSION=0.1.3
```
creachadair, line 7: I suppose this could remain at the top level since it's shared. YMMV.

creachadair (approved):

cmd/nginx-auth/mkdeb.sh
DentonGentry: Done.

PR #7326, opened 2023-02-19
syncs: add Waiter as a way to wakeup worker goroutines

maisem:


syncs/syncs.go
```diff
@@ -12,6 +12,42 @@ import (
...
+	return &Waiter{ch: make(chan struct{}, 1)}
+}
+
+// Wake wakes up a goroutine waiting on Wait. It returns true if it managed to
+// mark the waiter as woken up. If it returns false, a Wake was already pending.
+// If there are multiple goroutines waiting, only one will wake up.
+// If there are no goroutines waiting, the next call to Wait will return
+// immediately. Multiple calls to Wake without a call to Wait in between will
+// only wake up one goroutine.
+func (t *Waiter) Wake() (ok bool) {
```
bradfitz, line 31: If the method is called "Wake" why not make the type be "Waker" or "Sleeper"?

syncs/syncs.go
```diff
@@ -12,6 +12,42 @@ import (
...
+type Waiter struct {
+	ch chan struct{} // buffered chan of size 1
+}
+
+// NewWaiter returns a new Waiter.
+func NewWaiter() *Waiter {
+	return &Waiter{ch: make(chan struct{}, 1)}
+}
+
+// Wake wakes up a goroutine waiting on Wait. It returns true if it managed to
```
bradfitz, line 25: s/returns true if/reports whether/

bradfitz (approved):

PR #7329, opened 2023-02-20
tool/gocross: a tool for building Tailscale binaries

danderson:
First commit is a straight copy over, except for adding license headers and fixing up the the paths from misc/gocross to tool/gocross. Second commit makes tool/go use gocross unconditionally.

I'm keeping gocross-wrapper.sh around standalone for now rather than moving it into tool/go, because I'm not yet sure how this is going to get wired into our other repos, and having the standalone wrapper still available might help.


danderson:
wasm build is broken and I don't understand why yet. Converting to draft while I figure that out.

danderson:
Okay, it's because gocross does -trimpath, and cmd/tsconnect assumes that it can locate directories on disk based on the embedded package info. Pondering how to fix, but my current thinking is to change cmd/tsconnect and not gocross.

danderson:
Okay, after some digging: bunch of stuff to change, in gocross and tsconnect's builder, but it's doable.

@bradfitz @DentonGentry  how'd you prefer to proceed? I can special-case tsconnect to bypass gocross and let it keep doing its normal thing while I work on the fixes, or figure out the fixups in this PR and only submit gocross as a whole when that's worked out. I'm happy either way, it's just a question of which happens-before and if you prefer a hack + multiple smaller PRs, or a big PR with a bunch of commits here.

danderson:
(implemented the "bypass for now" option with a CI change just to make CI clean)

danderson:
I've implemented the various fixes to make tsconnect (and generally a bunch of other things like invoking runtime.GOROOT+"bin/go" from tests) work correctly. 

These changes all need to be squashed and merged as one, since the intermediate states are various kinds of broken.

First two changes are just mechanically moving gocross from the other repo and wiring it in. Subsequent 4 commits are the variety of fixes needed to make builds in this repo work happily. PTAL?

tool/gocross/autoflags.go
```diff
@@ -0,0 +1,183 @@
...
+		failReflect = true
+		fallthrough
+	case "darwin":
+		cgo = nativeGOOS == "darwin"
+		tags = append(tags, "omitidna", "omitpemdecrypt")
+		if env.IsSet("XCODE_VERSION_ACTUAL") {
+			var xcodeFlags []string
+			// Minimum OS version being targeted, results in
+			// e.g. -mmacosx-version-min=11.3
+			minOSKey := env.Get("DEPLOYMENT_TARGET_CLANG_FLAG_NAME", "")
```
DentonGentry, line 108: Having the default be "" means xcode will be passed an argument of `-=`, which seems like it cannot work. Should we hard-code a minimum iOS and macOS version here?

tool/gocross/autoflags.go
```diff
@@ -0,0 +1,183 @@
...
+		newArgv = append(newArgv, fmt.Sprintf("-tags=%s", strings.Join(tags, ",")))
+	}
+	if len(ldflags) > 0 {
+		newArgv = append(newArgv, "-ldflags", strings.Join(ldflags, " "))
+	}
+	newArgv = append(newArgv, argv[2:]...)
+
+	env.Set("GOOS", targetOS)
+	env.Set("GOARCH", targetArch)
+	env.Set("GOARM", "5") // TODO: fix, see go/internal-bug/3092
```
DentonGentry, line 147: Since no-one outside will be able to resolve go/internal-bug/3092, we might consider putting the note on that bug instead that if we change the GOARM default that gocross needs to be updated.

tool/gocross/gocross.go
```diff
@@ -0,0 +1,132 @@
...
+func debug(format string, args ...interface{}) {
+	debug := os.Getenv("GOCROSS_DEBUG")
+	var (
+		out *os.File
+		err error
+	)
+	switch debug {
+	case "0", "":
+		return
+	case "1":
```
DentonGentry, line 106: `GOCROSS_DEBUG=stderr` might also be reasonable.

tool/gocross/autoflags.go
danderson: AFAICT, Xcode can't not set these envvars. And if it does stop setting them, the compile here will fail because `-mmacosx-version-min=""` isn't valid. I think I'd rather thave that failure mode, rather than silently paper over Xcode behavioral change and not figure out mysterious problems for a long time.

tool/gocross/autoflags.go
danderson: My plan once gocross is merged into this repo, is to move all the remaining bugs over. I can't do that easily however (can't move issues from private to public repos, I have to re-transcribe them), and I'm trying to keep the set of work required for initial import down. So, I would prefer to update this in a followup PR, once I've refiled the relevant bugs.

tool/gocross/gocross.go
danderson: Sure, why not, added as one more commit to this chain.

danderson:
Given the size of the PR seems to be turning off reviewers, going to close this, submit all the fixups as PRs in the other repo, and then reopen a PR that is purely a `cp` into this repo with no non-mechanical changes.

Issue #6571, opened 2022-11-30
FR: Support stripping HTTP request paths from funnel proxy routes


mvdan:
### What are you trying to do?

I'm using Funnel to replace ngrok with tailscale. So far it's been working well!

Right now, I have set it up as follows:

```
# tailscale serve status
https://p14s.blue-toad.ts.net (Funnel on)
|-- / proxy http://127.0.0.1:10080
```

Right now I am using this to test HTTP webhooks from a GitHub App, where the local Go server simply listens on http at port 10080. GitHub sends the events to https://p14s.blue-toad.ts.net/. So far so good, and it works perfectly.

However, this particular GitHub app is only one of the pieces of software that I want to expose via Funnel for end-to-end testing. I _could_ swap which program listens on port 10080 each time, but that's not great. For example, if I stop my Go backend for that GitHub App and work on something else serving HTTP over funnel, it will _still_ receive the GitHub App webhook events from the previous app, because they're still wired up to the same place.

I _could_ disable the webhook events from the GitHub App every time I stop testing it, but that's cumbersome. The problem here is that I'm sharing the same host (https://p14s.blue-toad.ts.net/) to test multiple apps at different times.

One decent option is to use different ports:

```
  --serve-port uint
    	port to serve on (443, 8443 or 10000) (default 443)
```

However, that's only three ports, so I could only test three different pieces of software, and then we're back to the same problem of sharing the same host and port. It would also be confusing for future me: I'll be sure to forget what I used each port for, even if I had unlimited ports. Potentially, attaching some form of label or name to each route or port, but that wouldn't be visible when looking at the HTTPS URLs alone.

I tried to solve this by using different paths, like so:

```
# tailscale serve status
https://p14s.blue-toad.ts.net (Funnel on)
|-- /                     text  "p14s is online"
|-- /cue-unity-github-app proxy http://127.0.0.1:10080
```

This way, the GitHub App was configured with the URL https://p14s.blue-toad.ts.net/cue-unity-github-app, which is not going to be shared with any other funnel route. The path also helps me remember what I exposed this particular route for. However, when doing a GET on that URL, an HTTP server listening on `10080` sees:

```
$ nc -l 127.0.0.1 10080
GET /cue-unity-github-app HTTP/1.1
Host: p14s.blue-toad.ts.net
[...]
X-Forwarded-For: 100.93.97.75
```

[...]

Issue #6930, opened 2023-01-10
Funnel TCP doesn't work locally from macOS


FiloSottile:
### What is the issue?

I have a shiny TCP Funnel that's reachable from the outside world, but if I try to connect to it from the machine where it's served, or from iOS with Tailscale on, it doesn't work.

Arguably, I can just bind to 0.0.0.0 and connect to the local port instead, but that's marginally less magical and requires switching addresses between environments, which is not necessarily trivial with some protocols.

### Steps to reproduce

* `tailscale serve funnel on`
* start a server at `localhost:8443`
* `tailscale serve tcp 8443`
* `curl https://$NODE.$TSNET.ts.net/` from the local machine fails
* loading `https://$NODE.$TSNET.ts.net/` from iOS with Tailscale on fails
* loading `https://$NODE.$TSNET.ts.net/` from iOS with Tailscale off works

### OS

macOS

### OS version

macOS Ventura 13.1

### Tailscale version

1.34.2

### Bug report

BUG-78902bdfaf724fdf4c1dd793ee86605b16b8d9226361f5cceff79dd02ede7d38-20230110191137Z-7f81aa22b0fb97a4

bradfitz:
This is what the "works best on Linux" comment at https://tailscale.com/kb/1223/tailscale-funnel/#limitations is referring to.

We try to run an extra listener for local traffic but the macOS sandbox is annoying in this regard. @shayne was fighting it a bit and trying to convince macOS to do what we want but we were getting permission errors, even with our various network server entitlements enabled.

We need to try fighting it more.

shayne:
I have https://github.com/tailscale/tailscale/pull/6771, which would bring the magic back by working around platform limitations. It's silly, but we can listen on all interfaces, but strangely the system fails on a specific interface. This PR makes an exception for privileged ports when running in the macOS sandbox and listens on all interfaces.

shayne:
@FiloSottile https://github.com/tailscale/tailscale/pull/6771 was merged! This should make it into the v1.36 release.

shayne:
@FiloSottile v1.36 was released, and you should be able to access your Funnel URL locally now. LMK

If you have yet to receive the update, check the [App Store](https://apps.apple.com/us/app/tailscale/id1475387142?mt=12).

FiloSottile:
It works! Thank you @shayne.

Multiply:
[...]

PR #7334, opened 2023-02-21
ipn/ipnlocal: [serve] Trim mountPoint prefix from proxy path

shayne:
This change trims the mountPoint from the request URL path before
sending the request to the reverse proxy.

Today if you mount a proxy at `/foo` and request to
`/foo/bar/baz`, we leak the `mountPoint` `/foo` as part of the request
URL's path.

This fix makes removed the `mountPoint` prefix from the path so
proxied services receive requests as if they were running at the root
(`/`) path.

This could be an issue if the app generates URLs (in HTML or otherwise)
and assumes `/path`. In this case, those URLs will 404.

With that, I still think we should trim by default and not leak the
`mountPoint` (specific to Tailscale) into whatever app is hosted.
If it causes an issue with URL generation, I'd suggest looking at configuring
an app-specific path prefix or running Caddy as a more advanced
solution.

Fixes: #6571

### DIFF
--- a/ipn/ipnlocal/serve.go
+++ b/ipn/ipnlocal/serve.go
@@ -476,7 +476,12 @@ func (b *LocalBackend) serveWebHandler(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, "unknown proxy destination", http.StatusInternalServerError)
 			return
 		}
-		p.(http.Handler).ServeHTTP(w, r)
+		h := p.(http.Handler)
+		// Trim the mount point from the URL path before proxying. (#6571)
+		if r.URL.Path != "/" {
+			h = http.StripPrefix(strings.TrimSuffix(mountPoint, "/"), h)
+		}
+		h.ServeHTTP(w, r)
 		return
 	}
 


maisem (approved):

shayne:
Sorry, you're having trouble. It's a tricky decision, but we decided it's better not to leak the mount-point to the backend. Imagine the case where you mount the same app at two different mount points or if you're using a project that doesn't support the path prefix.

There are a couple of options. You can use `/` as your mount point. This setup will forward all paths to your reverse proxy. 

You could also set up Caddy (or similar) and use our TCP forwarding features to have complete control over web requests.

Multiply:
Could you add support for adding an optional path to the `<source>`?

That way I could do something similar to the following, and still support it:
```
tailscale serve https:443 / http://127.0.0.1:4000
tailscale serve https:443 /api/v1/graphql http://127.0.0.1:4000/api/v1/graphql
```

It's explicit that I want this behaviour for the second backend, as it's expecting that path to be hit.

shayne:
The second backend also listens on `127.0.0.1:4000`, in which case the first serve command would satisfy all paths.

Multiply:
My bad, I meant to write `:3000` for the first one. Bad copypaste.

shayne:
OK, thanks. I'll take a look at what it would look like and take it into consideration. Thanks for reaching out!

Multiply:
Thanks a lot @shayne - It would be super beneficial to our team at least.

Do let me know if you want an alpha tester or 7! 🚀

shayne:
Will do. In the meantime definitely consider wiring up Caddy via:
```console
$ tailscale serve tls-terminated-tcp:443 tcp://localhost:80
```

Multiply:
Slightly off-topic, hope it's okay.

Do you know why my colleague can access his own funnel from his own machine (tls on port 443), and I can't access my own?
It worked at some point prior to rebooting my own machine, but no longer. Nothing binds to 443 on my machine, but IPNExtension does on his.

shayne:
I worked on this at https://github.com/tailscale/tailscale/issues/6930

macOS sandbox can be annoying. We should move specifics on to that issue. I'm curious if macOS versions are the same and if you can bind any other process to 443, even netcat.

shayne:
@Multiply see #7743

Multiply:
@shayne thanks for working on this so quickly! 🚀 

Just waiting for my app store to show the update, and I'll try it out. ⌛

shayne:
Sure thing! We also have builds available on our packages server:
https://pkgs.tailscale.com/stable/#macos

larryqiann:
Sorry to reopen this, but I'm wondering whether this works the way I think it is working.

I have a service running at localhost:3080. When I do sudo tailscale serve --bg --set-path /path 3080, I get 404 errors when I go to my-tailnet.ts.net/path, but when i do sudo tailscale serve --bg 3080 and go to my-tailnet.ts.net, everything works fine.

My understanding is that after this change, the /path part is removed silently from the URL when being passed to the app. Is this a misunderstanding of how this works? How can I achieve this functionality?

Thanks!

PR #7338, opened 2023-02-21
wgengine/magicsock: track per-endpoint changes in ringbuffer

andrew-d:
This change adds a ringbuffer to each magicsock endpoint that keeps a fixed set of "changes"–debug information about what updates have been made to that endpoint.

Additionally, this adds a LocalAPI endpoint and associated "debug peer-status" CLI subcommand to fetch the set of changes for a given IP or hostname.

Change-Id: I34f726a71bddd0dfa36ec05ebafffb24f6e0516a


andrew-d:
I'm happy to take suggestions on the naming and return types here; this is very much a proof-of-concept.

andrew-d:
Example output for my ARM dev VM, slightly redacted:
```
$ ./tailscale --socket /tmp/tailscaled.usermode.sock debug peer-status andrew-arm64-dev
2023/02/21 15:56:08 lookup "andrew-arm64-dev" => "100.64.1.2"
[
 {
  "From": "",
  "To": "127.3.3.40:21",
  "What": "updateFromNode-DERP"
 },
 {
  "To": [
   "18.217.1.1:53442",
   "172.31.16.1:53442"
  ],
  "What": "updateFromNode-new-Endpoints"
 }
]
```

ipn/localapi/localapi.go
```diff
@@ -718,6 +719,34 @@ func (h *Handler) serveStatus(w http.ResponseWriter, r *http.Request) {
 	e.Encode(st)
 }
 
+func (h *Handler) servePeerStatus(w http.ResponseWriter, r *http.Request) {
```
maisem, line 722: peerStatus seems like a bad name here, lets be super specific `serveDebugPeerEndpointChanges`
same in the url

client/tailscale/localclient.go
```diff
@@ -430,6 +430,19 @@ func (lc *LocalClient) status(ctx context.Context, queryString string) (*ipnstat
 	return decodeJSON[*ipnstate.Status](body)
 }
 
+// PeerStatus returns debug information about the provided peer.
+//
+// TODO(andrew): better return type
+func (lc *LocalClient) PeerStatus(ctx context.Context, ip netip.Addr) ([]any, error) {
```
maisem, line 436: lets at least do `[]magicsock.EndpointChange` for now

maisem: we can also skip `LocalClient` for now and only have the API ¯\_(ツ)_/¯

wgengine/magicsock/magicsock.go
```diff
@@ -3679,6 +3700,15 @@ func ippDebugString(ua netip.AddrPort) string {
 	return ua.String()
 }
 
+// EndpointChange is a structure containing information about changes made to a
+// particular endpoint. This is not a stable interface and could change at any
+// time.
+type EndpointChange struct {
+	What string // what this change is
```
maisem, line 3707: Store time too?

wgengine/magicsock/magicsock.go
```diff
@@ -2598,6 +2618,7 @@ func (c *Conn) SetNetworkMap(nm *netmap.NetworkMap) {
 
 		ep := &endpoint{
 			c:                 c,
+			debugUpdates:      ringbuffer.New[EndpointChange](100),
```
maisem, line 2621: could this be too much for a large tailnet?

ipn/localapi/localapi.go
andrew-d: Done

client/tailscale/localclient.go
andrew-d: Removed the `LocalClient` for now and just had the `debug` subcommand make the HTTP request.

wgengine/magicsock/magicsock.go
andrew-d: Done

maisem (approved):

wgengine/magicsock/magicsock.go
andrew-d: I just pushed a change that limits the total number of ring buffer entries, counted by dividing `len(nm.Peers)`. I added a new envknob type (`RegisterInt`) for this.

cmd/tailscale/cli/debug.go
```diff
@@ -789,3 +795,52 @@ func runCapture(ctx context.Context, args []string) error {
...
+	}
+
+	resp, err := localClient.DoLocalRequest(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	var changes []magicsocktypes.EndpointChange
+	if err := json.NewDecoder(resp.Body).Decode(&changes); err != nil {
```
maisem, line 841: Lol, why decode to marshal again?

Just print it out? Or json.Indent

client/tailscale/localclient.go
andrew-d: Pulled this out into a subpackage so that we don't pull in a dependency on the `magicsock` package in `cmd/tailscale` here

cmd/tailscale/cli/debug.go
andrew-d: Heh, brain fart. Will just do that.

andrew-d: (I copied from the `localClient` method which decoded JSON... but just printing is easier and allows removing the new package that I just added 😅 )

andrew-d: Okay, removed the decode/marshal dance entirely

andrew-d:
If you could double-check my logic when picking the size of the ring buffers, I'd appreciate it; otherwise I think I addressed all remaining comments.

maisem:
> If you could double-check my logic when picking the size of the ring buffers, I'd appreciate it; otherwise I think I addressed all remaining comments.

LGTM

PR #7350, opened 2023-02-23
tsnet: add data transfer test

crawshaw:
It works! Well, sort-of.

```
$ go test -v -run=TestConn ./tsnet
=== RUN   TestConn
    tsnet_test.go:63: DERP httpsrv listener: 127.0.0.1:59689
    tsnet_test.go:73: testcontrol listening on http://127.0.0.1:59691
    tsnet_test.go:128: ping success: &ipnstate.PingResult{IP:"100.64.0.1", NodeIP:"100.64.0.1", NodeName:"s1", Err:"", LatencySeconds:5.219534209, Endpoint:"", DERPRegionID:0, DERPRegionCode:"", PeerAPIPort:0x0, PeerAPIURL:"", IsLocalIP:false}
    tsnet_test.go:149: got: "hello"
    stuntest.go:63: STUN server shutdown
--- PASS: TestConn (5.60s)
PASS
```

The ping takes 5 seconds: `LatencySeconds:5.219534209` That's a very familiar number. It suggests... something isn't working? The local DERP? But I can't quite figure it out.

Also if I remove the ping and just attempt the TCP connection immediately, it times out after 15s. So it's a load bearing ping. :-/ Any ideas?

### DIFF
--- a/tsnet/tsnet_test.go
+++ b/tsnet/tsnet_test.go
@@ -4,8 +4,23 @@
 package tsnet
 
 import (
+	"context"
 	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"path/filepath"
+	"os"
+	"net/http/httptest"
 	"testing"
+	"time"
+
+	"tailscale.com/ipn/store/mem"
+	"tailscale.com/tailcfg"
+	"tailscale.com/tstest/integration"
+	"tailscale.com/net/netns"
+	"tailscale.com/tstest/integration/testcontrol"
+	"tailscale.com/types/logger"
 )
 
 // TestListener_Server ensures that the listener type always keeps the Server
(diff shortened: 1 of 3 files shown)


tsnet/tsnet_test.go
```diff
@@ -44,3 +56,98 @@ func TestListenerPort(t *testing.T) {
...
+	want := "hello"
+	if _, err := io.WriteString(w, want); err != nil {
+		t.Fatal(err)
+	}
+
+	got := make([]byte, len(want))
+	if _, err := io.ReadAtLeast(r, got, len(got)); err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("got: %q", string(got))
```
bradfitz, line 149: btw %q works on []byte the same as string so this string(got) conversion is unnecessary

bradfitz (approved):
Phew, looks like I expected! :)

crawshaw:
the 5s is a complete mystery to me, looks like some sort of client reconfig after 5 seconds kicks the whole system into gear. going to move forward with it as is, because its existence doesn't stop me generalizing this test to things built on tsnet.

crawshaw:
and the test works entirely differently on linux and macos. great.

crawshaw:
OK I fixed this the usual way, by making the test spin. :( Hopefully future controlclient work makes that unnecessary.

PTAL @bradfitz @maisem 

I also fixed some distracting Close/Shutdown panics that were appearing when the test was failing.

PR #7355, opened 2023-02-23
tka: add public API on NodeKeySignature for key information

twitchyliquid64:
This is needed in the coordination server.

### DIFF
--- a/tka/sig.go
+++ b/tka/sig.go
@@ -96,6 +96,18 @@ type NodeKeySignature struct {
 	WrappingPubkey []byte `cbor:"6,keyasint,omitempty"`
 }
 
+// UnverifiedWrappingPublic returns the public key which must sign a
+// signature which embeds this one, if any.
+//
+// See docs on NodeKeySignature.WrappingPubkey & SigRotation for documentation
+// about wrapping public keys.
+//
+// SAFETY: The caller MUST verify the signature using
+// Authority.NodeKeyAuthorized if treating this as authentic information.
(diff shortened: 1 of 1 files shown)


tka/sig.go
```diff
@@ -96,6 +96,15 @@ type NodeKeySignature struct {
 	WrappingPubkey []byte `cbor:"6,keyasint,omitempty"`
 }
 
+// UnverifiedWrappingPublic returns the public key which must sign a
+// signature which embeds this one, if any.
```
maisem, line 100: this needs more explanation

twitchyliquid64: Ive added a line pointing the reader towards documentation about wrapping keys.

tka/sig.go
```diff
@@ -96,6 +96,18 @@ type NodeKeySignature struct {
...
+//
+// See docs on NodeKeySignature.WrappingPubkey & SigRotation for documentation
+// about wrapping public keys.
+//
+// SAFETY: The caller MUST verify the signature using
+// Authority.NodeKeyAuthorized if treating this as authentic information.
```
maisem, line 106: why is it `Unverified`?

twitchyliquid64: Because its possible for the caller to get at this information without verifying the signature.

maisem (approved):

tka/sig.go
maisem: thanks!

PR #7357, opened 2023-02-23
ipn/ipnlocal: add delegated interface information to /interfaces PeerAPI handler

mihaip:
Exposes the delegated interface data added by #7248 in the debug endpoint. I would have found it useful when working on that PR, and it may be handy in the future as well.

Also makes the interfaces table slightly easier to parse by adding borders to it. To make then nicer-looking, the CSP was relaxed to allow inline styles.

### DIFF
--- a/ipn/ipnlocal/peerapi.go
+++ b/ipn/ipnlocal/peerapi.go
@@ -670,7 +670,7 @@ func (h *peerAPIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if peerAPIRequestShouldGetSecurityHeaders(r) {
-		w.Header().Set("Content-Security-Policy", `default-src 'none'; frame-ancestors 'none'; script-src 'none'; script-src-elem 'none'; script-src-attr 'none'`)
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/peerapi.go
```diff
@@ -799,15 +799,21 @@ func (h *peerAPIHandler) handleServeInterfaces(w http.ResponseWriter, r *http.Re
...
+	}
+
 	i, err := interfaces.GetList()
 	if err != nil {
 		fmt.Fprintf(w, "Could not get interfaces: %s\n", html.EscapeString(err.Error()))
 		return
 	}
 
-	fmt.Fprintln(w, "<table>")
+	fmt.Fprintln(w, "<table style='border-collapse: collapse' border=1 cellspacing=0 cellpadding=2>")
```
mihaip, line 814: This `style` attribute gets rejected because of the CSP added in #6408. What do you think about allowing inline styles?

bradfitz: i'm fine with inline styles

mihaip: OK, added `style-src 'unsafe-inline'` to the policy.

bradfitz (approved):

PR #7366, opened 2023-02-24
tsnet: add Up method to block until ready

crawshaw:
cc @bradfitz

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -28,6 +28,7 @@ import (
 	"tailscale.com/hostinfo"
 	"tailscale.com/ipn"
 	"tailscale.com/ipn/ipnlocal"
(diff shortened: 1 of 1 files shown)


tsnet/tsnet.go
```diff
@@ -146,6 +147,48 @@ func (s *Server) Start() error {
 	return s.initErr
 }
 
+// Up connects the server to the tailnet and waits until it is running.
+// On success it returns the current status, including a Tailscale IP address.
+func (s *Server) Up(ctx context.Context) (*ipnstate.Status, error) {
```
maisem, line 152: probably should also call `Start`.

crawshaw: added a comment that this is done by LocalClient

maisem (approved):

PR #7368, opened 2023-02-24
release: open-source release build logic for unix packages

danderson:
Mechanical copy from other repo. Only changes are package paths and paths to the deb/rpm helper scripts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ cmd/tailscaled/tailscaled
 .direnv/
 
 /gocross
+/dist
--- a/cmd/dist/dist.go
+++ b/cmd/dist/dist.go
@@ -1,0 +1,134 @@
+// The dist command builds Tailscale release packages for distribution.
+package main
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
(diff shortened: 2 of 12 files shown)


cmd/dist/README.md
```diff
@@ -0,0 +1,11 @@
+<!-- README.md auto-generated by misc/genreadme; DO NOT EDIT. (or remove this line) -->
+
+# dist
+
+The dist command builds Tailscale release packages for distribution.
+
+To build signed packages for distribution (which only the build servers should
+be doing), pass --win-signer, or set the SIGNER environment variable.
```
bradfitz, line 8: --win-signer is only for Windows or just a bad flag name?

cmd/dist/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+The value of the flag/envvar can be a path to a local signing script, or a URL
+(typically: http://builds.corp.ts.net/api/sign).
```
bradfitz, line 11: we want this URL in public docs?

cmd/dist/README.md
danderson: Deleted the readme, since it's not really contributing anything (and was outdated even before I copied it, now I look at what it said)

cmd/dist/README.md
danderson: Same, deleted file.

bradfitz (approved):

PR #7371, opened 2023-02-24
.github/workflows: add a pass/fail verdict job to the test workflow

danderson:
Github requires explicitly listing every single job within a workflow that is required for status checks, instead of letting you list entire workflows. This is ludicrous, and apparently this nonsense is the workaround.


crawshaw (approved):
sure looks like yaml

PR #7374, opened 2023-02-24
cmd/tailscale/cli: fix TestUpdatePrefs over Tailscale SSH

### LINKED ISSUE #7373: cmd/tailscale/cli: TestUpdatePrefs/enable_ssh fails when test is run over tailscale SSH

jwhited:
Fixes #7373


valscale:
This works when run both from tailscale ssh login and normal ssh login.

danderson (approved):

PR #7385, opened 2023-02-26
net/pidlisten: new package that restricts dials to the current process

crawshaw:
To be used in the C library wrapping tsnet to provide LocalAPI access.

This commit contains a linux implementation.
More operating systems to follow.

### DIFF
--- a/net/pidlisten/pidlisten.go
+++ b/net/pidlisten/pidlisten.go
@@ -1,0 +1,42 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+// Package pidlisten implements a TCP listener that only
+// accepts connections from the current process.
+package pidlisten
+
+import (
+	"fmt"
+	"net"
+)
+
+type listener struct {
+	ln net.Listener
+}
+
+func (pln *listener) Accept() (net.Conn, error) {
+	for {
+		conn, err := pln.ln.Accept()
+		if err != nil {
+			return nil, err
+		}
+		ok, err := checkPIDLocal(conn)
+		if err != nil {
+			conn.Close()
+			return nil, fmt.Errorf("pidlisten: %w", err)
+		}
+		if !ok {
+			conn.Close()
+			continue
+		}
+		return conn, nil
+	}
+}
+
+func (pln *listener) Close() error {
+	return pln.ln.Close()
+}
+
+func (pln *listener) Addr() net.Addr {
+	return pln.ln.Addr()
+}
--- a/net/pidlisten/pidlisten_linux.go
+++ b/net/pidlisten/pidlisten_linux.go
@@ -1,0 +1,63 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package pidlisten
+
+import (
+	"errors"
+	"fmt"
+	"go4.org/mem"
+	"io/fs"
+	"net"
+	"os"
+	"path/filepath"
+	"tailscale.com/util/dirwalk"
+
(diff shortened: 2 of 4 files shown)


net/pidlisten/pidlisten_linux.go
```diff
@@ -0,0 +1,63 @@
...
+	"path/filepath"
+	"tailscale.com/util/dirwalk"
+
+	"github.com/vishvananda/netlink"
+)
+
+// NewPIDListener wraps a net.Listener so that it only accepts connections from the current process.
+func NewPIDListener(ln net.Listener) net.Listener {
+	return &listener{ln: ln}
+}
```
bradfitz, line 22: I prefer the style of moving this to the main portable file (pidlisten.go) so the docs are in one place, and then having a `var newPIDListener func(net.Listener) net.Listener` that OS-specific files can init in their `init`.

Also: can this just be called `New`, given the package name?

net/pidlisten/pidlisten_linux.go
```diff
@@ -0,0 +1,63 @@
...
+	// that its inode belongs to the current PID.
+	s, err := netlink.SocketGet(conn.RemoteAddr(), conn.LocalAddr())
+	if err != nil {
+		return false, err
+	}
+
+	want := fmt.Sprintf("socket:[%d]", s.INode)
+	dir := fmt.Sprintf("/proc/%d/fd", os.Getpid())
+	err = dirwalk.WalkShallow(mem.S(dir), func(name mem.RO, de fs.DirEntry) error {
+		n, err := os.Readlink(filepath.Join(dir, name.StringCopy()))
```
bradfitz, line 53: add TODO to remove this garbage like how portlist_linux.go does it with its readlink?  (maybe we export that somewhere)

net/pidlisten/pidlisten.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+}
+
+func (pln *listener) Close() error {
+	return pln.ln.Close()
+}
+
+func (pln *listener) Addr() net.Addr {
+	return pln.ln.Addr()
+}
```
bradfitz, line 42: or embed `ln` instead and omit these. I don't care either way.

net/pidlisten/pidlisten_linux.go
bradfitz: Oh, this part is portable. You meant to move this to the other file.

net/pidlisten/pidlisten_noimpl.go
```diff
@@ -0,0 +1,13 @@
...
+
+//go:build !linux
+// +build !linux
+
+package pidlisten
+
+import "net"
+
+func checkPIDLocal(conn net.Conn) (bool, error) {
+	panic("not implemented")
```
bradfitz, line 12: just return an error for now? The signature permits it.

net/pidlisten/pidlisten_noimpl.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+//go:build !linux
+// +build !linux
```
bradfitz, line 5: we removed +build from the tree already

net/pidlisten/pidlisten_test.go
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+//go:build linux
+// +build linux
```
bradfitz, line 5: delete +build

crawshaw:
Planning to switch to a consistent http header approach that will work on iOS.

PR #7389, opened 2023-02-27
tsnet: add a LocalAPI listener on loopback, with required header

crawshaw:
This is for use by LocalAPI clients written in other languages that don't appear to be able to talk HTTP over a socket (e.g. java.net.http.HttpClient).

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -8,6 +8,8 @@ package tsnet
 
 import (
 	"context"
+	crand "crypto/rand"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


crawshaw:
cc @twitchyliquid64 who loves things that listen on tcp loopback addresses

crawshaw:
Ugh I see after getting this working that 1. there's something different about linux, and 2. we already have `localapi.Handler.RequirePassword`, which I may as well use instead. Will rework.

tsnet/tsnet_test.go
```diff
@@ -167,3 +173,59 @@ func TestConn(t *testing.T) {
...
+	res.Body.Close()
+	if res.StatusCode != 403 {
+		t.Errorf("GET %s returned %d, want 403 permission denied without header", url, res.StatusCode)
+	}
+
+	req, err = http.NewRequestWithContext(ctx, "GET", url, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	req.Header.Set("Sec-Tailscale-LocalAPI", cred)
```
creachadair, line 222: Can we please also have a test with the header but an incorrect value for it?

tsnet/tsnet.go
```diff
@@ -139,6 +144,63 @@ func (s *Server) LocalClient() (*tailscale.LocalClient, error) {
...
+//
+// As the LocalAPI is powerful, access to endpoints requires passing a
+// "Sec-Tailscale-LocalAPI" HTTP header containing the value of cred.
+//
+// It will start the server and the local client listener if they have not
+// been started yet.
+//
+// If you only need to use the LocalAPI from Go, then prefer LocalClient
+// as it does not require communication via TCP.
+func (s *Server) LocalAPI() (addr string, cred string, err error) {
```
creachadair, line 157: "LocalAPI" seems too similar to "LocalClient" without an obvious explanatory role. Maybe something that describes the return value, e.g., "LocalAPIAddress"?
```suggestion
func (s *Server) LocalAPI() (addr, cred string, err error) {
```

twitchyliquid64:
I don't love localhost listeners for us, but given its non-default, on a non-stable port, and requires a bearer value in a header browsers cant set, I think this is okay from a security standpoint.

crawshaw:
reworked to use existing RequiredPassword machinery instead of a header

crawshaw:
oh but maybe @twitchyliquid64 this isn't great because now in theory a browser could do it. hmmm.

maybe a `Sec-` header in addition to basic auth then.

tsnet/tsnet_test.go
crawshaw: I switched the cred over to the RequiredPassword machinery that's tested elsewhere, so I believe "" and the correct value are enough now.

crawshaw:
OK this is now basic auth + a Sec- header.

tsnet/tsnet.go
```diff
@@ -88,19 +90,22 @@ type Server struct {
...
+	initOnce            sync.Once
+	initErr             error
+	lb                  *ipnlocal.LocalBackend
+	netstack            *netstack.Impl
+	linkMon             *monitor.Mon
+	rootPath            string // the state directory
+	hostname            string
+	shutdownCtx         context.Context
+	shutdownCancel      context.CancelFunc
+	localAPICred        string                 // Sec-Tailscale-LocalAPI value for localAPITCPListener
```
twitchyliquid64, line 102: Comment needs updating that this is now basic auth

twitchyliquid64 (approved):

PR #7393, opened 2023-02-27
Add support for OAuth tokens

farcaller:
Add support for OAuth tokens for the gitops-pusher.


DentonGentry (approved):
Thank you!

DentonGentry:
This will definitely be in a Tailscale 1.38.0 release.
If we build a 1.36.3 before then, it should be in that release.

We can update https://github.com/tailscale/gitops-acl-action/blob/main/action.yml once the updated cmd/gitops-pusher/gitops-pusher.go makes it into a release build.

DentonGentry:
The PR doesn't cleanly cherry-pick into the 1.36 releaser branch, please assume it will first appear in 1.38.0.

PR #7396, opened 2023-02-28
Julia updateintro1

JuliaTailscale:


api.md
```diff
@@ -1,199 +1,465 @@
...
-  * `enabledRoutes`
-  * `advertisedRoutes`
-  * `clientConnectivity` (which contains the following fields: `mappingVariesByDestIP`, `derp`, `endpoints`, `latency`, and `clientSupports`)
+## Endpoint URL 
+The Tailscale API URL path begins with `https://api.tailscale.com/api/v2/`. 
+
+Examples in this document begin with `api/v2/...`.
+
+## Authentication 
+TailScale uses several types of keys. The type of key is identified in the key prefix. All keys are structured as follows: `tskey-{key type}-{unique key value}`. The `key type` indicates the type of Tailscale key.
```
bradfitz, line 13: Wrong case on Tailscale

willnorris:
This is great, Julia!  I started to add inline comments the other day, but realized that it was going to take just as long to comment as it would be to fix it.  So I went ahead and made changes locally which I'm going to push to your branch here.  It looks like a lot, but most of it is either stylistic, updating some language from the existing docs, or removing a couple of sections that I don't think are quite as necessary (at least for now).

At this point I think everything is in a good state, so I'm going to go ahead and merge this in as well.  I'll still come back to make other changes we discussed later (like OAuth scopes, etc), but I think this is at a good point to merge.  🎉

PR #7400, opened 2023-02-28
ipn/ipnlocal: remove windows exception from profile migration

### LINKED ISSUE #7398: Profile migration does not work on Windows
STR:

1. Install 1.22 on a Windows box with a clean `%ProgramData%\Tailscale` directory;
2. Log the machine into a tailnet, approve it, etc;
3. Upgrade to 1.36.2

Expected:

Still logged into tailnet

Actual:

Prompted to login again. `_migrated-to-profiles.conf` exists in `%LocalAppData%\Tailscale` but the state in `%ProgramData%\Tailscale` has not been updated.

Holding up upgrades in https://github.com/tailscale/coral/issues/78#issuecomment-1438799286 because they want to be able to do so without forcing users to log in again.

I've found the problem. PR forthcoming after I'm done with a meeting.

dblohm7:
The check in question results in profiles never being migrated to backend prefs on Windows clients. We should be doing that on Windows too.

This should be save vis-a-vis unattended mode since we won't see the unmigrated prefs until the GUI signs in.

Fixes #7398

### DIFF
--- a/ipn/ipnlocal/profiles.go
+++ b/ipn/ipnlocal/profiles.go
@@ -534,7 +534,7 @@ func newProfileManagerWithGOOS(store ipn.StateStore, logf logger.Logf, goos stri
 		if err := pm.setPrefsLocked(prefs); err != nil {
 			return nil, err
 		}
-	} else if len(knownProfiles) == 0 && goos != "windows" {
+	} else if len(knownProfiles) == 0 {
 		// No known profiles, try a migration.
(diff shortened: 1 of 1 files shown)


dblohm7:
@maisem and I paired a bit on this and its looking like the problem isn't in here, but rather a race somewhere else. (He was able to successfully upgrade from 1.22 to an unaltered 1.36 in a VM of his, even though I was able to reproduce on my laptop)

dblohm7:
Not the right thing to do. Closing.

PR #7401, opened 2023-02-28
derp/derphttp: pass `*tls.Config` to WebSocket connections

kylecarbs:


derp/derphttp/websocket.go
```diff
@@ -18,8 +20,13 @@ func init() {
 	dialWebsocketFunc = dialWebsocket
 }
 
-func dialWebsocket(ctx context.Context, urlStr string) (net.Conn, error) {
+func dialWebsocket(ctx context.Context, urlStr string, tlsConfig *tls.Config) (net.Conn, error) {
 	c, res, err := websocket.Dial(ctx, urlStr, &websocket.DialOptions{
+		HTTPClient: &http.Client{
```
mihaip, line 25: This breaks the wasm client (see the "CI / wasm" CI failure):

```
derp/derphttp/websocket.go:25:3: unknown field HTTPClient in struct literal of type websocket.DialOptions
```
The WebSocket library defines a different `DialOptions` struct for non-JS vs JS targets:

https://github.com/nhooyr/websocket/blob/14fb98eba64eeb5e9d06a88b98c47ae924ac82b4/dial.go#L23-L47
https://github.com/nhooyr/websocket/blob/14fb98eba64eeb5e9d06a88b98c47ae924ac82b4/ws_js.go#L236-L240

Easiest way may be to fork a `websocket_js.go` file that preserves the current code (and ignores the `tlsConfig` param), but open to other ideas if you have them.

derp/derphttp/websocket.go
```diff
@@ -1,14 +1,16 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
-//go:build linux || js
+//go:build linux || windows || darwin || js
```
mihaip, line 4: Would prefer to still exclude this on iOS (by using `(darwin && !ios)`), where we're still somewhat memory constrained.

derp/derphttp/websocket.go
kylecarbs: Yup, I'll adjust!

kylecarbs: I actually changed this in our code upstream to the exact same, just forgot to mirror!

ThomasK33:
Hey @kylecarbs, can I implement the requested changes and put you as a co-author in the new PR?

kylecarbs:
Of course! Feel free.

PR #7417, opened 2023-03-01
ipn,cmd/tailscale: implement resigning nodes on tka key removal

twitchyliquid64:
When `tailscale lock remove <keys>` is used, nodes which are signed by a valid to-be-removed signature are re-signed with the node's own key prior to being removed.

### DIFF
--- a/cmd/tailscale/cli/network-lock.go
+++ b/cmd/tailscale/cli/network-lock.go
@@ -267,14 +267,78 @@ var nlAddCmd = &ffcli.Command{
 	},
 }
 
+var nlRemoveArgs struct {
+	resign bool
+}
+
 var nlRemoveCmd = &ffcli.Command{
 	Name:       "remove",
-	ShortUsage: "remove <public-key>...",
+	ShortUsage: "remove [--re-sign=false] <public-key>...",
 	ShortHelp:  "Removes one or more trusted signing keys from tailnet lock",
 	LongHelp:   "Removes one or more trusted signing keys from tailnet lock",
-	Exec: func(ctx context.Context, args []string) error {
-		return runNetworkLockModify(ctx, nil, args)
-	},
+	Exec:       runNetworkLockRemove,
+	FlagSet: (func() *flag.FlagSet {
+		fs := newFlagSet("lock remove")
+		fs.BoolVar(&nlRemoveArgs.resign, "re-sign", true, "resign signatures which would be invalidated by removal of trusted signing keys")
+		return fs
+	})(),
+}
+
+func runNetworkLockRemove(ctx context.Context, args []string) error {
+	removeKeys, _, err := parseNLArgs(args, true, false)
+	if err != nil {
+		return err
+	}
+	st, err := localClient.NetworkLockStatus(ctx)
+	if err != nil {
+		return fixTailscaledConnectError(err)
+	}
+	if !st.Enabled {
+		return errors.New("tailnet lock is not enabled")
+	}
+
+	if nlRemoveArgs.resign {
+		// Validate we are not removing trust in ourselves while resigning. This is because
+		// we resign with our own key, so the signatures would be immediately invalid.
+		for _, k := range removeKeys {
+			kID, err := k.ID()
+			if err != nil {
+				return fmt.Errorf("computing KeyID for key %v: %w", k, err)
+			}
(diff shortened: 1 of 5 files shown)


ipn/localapi/localapi.go
```diff
@@ -1469,6 +1470,32 @@ func (h *Handler) serveTKALog(w http.ResponseWriter, r *http.Request) {
 	w.Write(j)
 }
 
+func (h *Handler) serveTKAAffectedSigs(w http.ResponseWriter, r *http.Request) {
+	if r.Method != httpm.POST {
+		http.Error(w, "use POST", http.StatusMethodNotAllowed)
+		return
+	}
+	keyID, err := ioutil.ReadAll(r.Body)
```
sailorfrag, line 1478: Wrap in a `http.MaxBytesReader`?

ipn/ipnlocal/network-lock_test.go
```diff
@@ -877,3 +878,137 @@ func TestTKAForceDisable(t *testing.T) {
 		t.Fatal("tka was re-initalized")
 	}
 }
+
+func TestTKAAffectedSigs(t *testing.T) {
+	envknob.Setenv("TAILSCALE_USE_WIP_CODE", "1")
```
sailorfrag, line 883: is this still needed?

ipn/ipnlocal/network-lock.go
```diff
@@ -1110,3 +1168,39 @@ func (b *LocalBackend) tkaSubmitSignature(ourNodeKey key.NodePublic, sig tkatype
...
+		NodeKey: ourNodeKey,
+		KeyID:   key,
+	}); err != nil {
+		return nil, fmt.Errorf("encoding request: %v", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	req2, err := http.NewRequestWithContext(ctx, "GET", "https://unused/machine/tka/affected-sigs", &req)
```
sailorfrag, line 1185: I feel like we can probably use better names than req/req2

cmd/tailscale/cli/network-lock.go
```diff
@@ -267,14 +267,78 @@ var nlAddCmd = &ffcli.Command{
...
 	ShortUsage: "remove <public-key>...",
 	ShortHelp:  "Removes one or more trusted signing keys from tailnet lock",
 	LongHelp:   "Removes one or more trusted signing keys from tailnet lock",
-	Exec: func(ctx context.Context, args []string) error {
-		return runNetworkLockModify(ctx, nil, args)
-	},
+	Exec:       runNetworkLockRemove,
+	FlagSet: (func() *flag.FlagSet {
+		fs := newFlagSet("lock remove")
+		fs.BoolVar(&nlRemoveArgs.noResigning, "no-resigning", false, "do not resign signature which would be invalidated by removal of trusted signing keys")
```
sailorfrag, line 282: Would it be better to have a `-re-sign` option that defaults to true rather than having an inverted meaning boolean flag?

cmd/tailscale/cli/network-lock.go
```diff
@@ -267,14 +267,78 @@ var nlAddCmd = &ffcli.Command{
...
+	if err != nil {
+		return fixTailscaledConnectError(err)
+	}
+	if !st.Enabled {
+		return errors.New("tailnet lock is not enabled")
+	}
+
+	if !nlRemoveArgs.noResigning {
+		// Validate we are not removing trust in ourselves while resigning. This is because
+		// we resign with our own key, so the signatures would be immediately invalid.
```
sailorfrag, line 302: It would probably be best to use "re-sign" rather than "resign" because that could mean resign as in resignation.

sailorfrag (approved):

cmd/tailscale/cli/network-lock.go
twitchyliquid64: Yeah that makes sense, but I had to lookup how to unset a default true flag in Go. I think it ends up being `--re-sign=false`

cmd/tailscale/cli/network-lock.go
twitchyliquid64: I think I gave Denton a heart attack on exactly this in #standup :grin: :sweat_smile:

PR #7418, opened 2023-03-01
tka: compact TKA storage on startup

twitchyliquid64:


ipn/ipnlocal/network-lock.go
```diff
@@ -37,6 +37,11 @@ import (
 var (
 	errMissingNetmap        = errors.New("missing netmap: verify that you are logged in")
 	errNetworkLockNotActive = errors.New("network-lock is not active")
+
+	tkaCompactionDefaults = tka.CompactionOptions{
+		MinChain: 24,                  // Keep at minimum 24 AUMs since head.
```
sailorfrag, line 42: Should we rate limit how fast AUMs can be issued (like a real blockchain!) to ensure that this minimum covers a reasonable length of time even if the device clock is wrong?

sailorfrag (approved):

ipn/ipnlocal/network-lock.go
twitchyliquid64: I'd rather not because a node can get a flurry of updates if its catching up after some period of time. I would also put such a ratelimit in Control, rather than on the client as for now at least all AUMs flow from control.

PR #7429, opened 2023-03-02
cmd/containerboot,kube: consolidate the two kube clients

maisem:
We had two implemenetations of the kube client, merge them.

### DIFF
--- a/kube/client.go
+++ b/kube/client.go
@@ -14,11 +14,15 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"sync"
 	"time"
+
+	"tailscale.com/util/multierr"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


kube/client.go
```diff
@@ -14,19 +14,35 @@ import (
...
 )
 
 const (
 	saPath     = "/var/run/secrets/kubernetes.io/serviceaccount"
 	defaultURL = "https://kubernetes.default.svc"
 )
 
+// rootPathForTests is set by tests to override the root path to the
+// service account directory.
+var rootPathForTests string
```
danderson, line 35: Package global means we can never parallelize any tests that use kube. It would be nice to avoid that, but given the current semantics of this package, I guess this is fine.

kube/client.go
```diff
@@ -14,19 +14,35 @@ import (
...
+
+// SetRootPathForTesting sets the path to the service account directory.
+func SetRootPathForTesting(p string) {
+	rootPathForTests = p
+}
+
 func readFile(n string) ([]byte, error) {
+	if rootPathForTests != "" {
+		return os.ReadFile(filepath.Join(rootPathForTests, saPath, n))
```
danderson, line 44: Does this actually work? I thought filepath.Join with a string that starts with `/` was equivalent to "remove all the parents and start again at `/`, but I've been programming in two other languages recently so maybe I'm confusing myself.

danderson (approved):

kube/client.go
maisem: I checked
https://go.dev/play/p/oMFIx5iG5M_C

kube/client.go
maisem: yeah, i plan on fixing this ~soon.

PR #7431, opened 2023-03-02
all: implement preauth-key support with tailnet lock

twitchyliquid64:
Design (internal doc): http://go/tailnet-lock-preauth-keys

The basic idea:
1. Admin goes to admin panel to generate a preauth key
2. Admin runs `tailscale lock tskey-wrap tskey-auth-XXXXXXX` to get a longer form of the same thing
3. Admin can now `tailscale up --authkey tskey-auth-XXXXXXX--NNNNNNNNNNNNN` to bring up nodes in a locked tailnet unattended.

Implementation details:
1. A new tailnet-lock key is generated for every wrapped key, to permit revocation on a per-authkey basis
2. The new tailnet-lock key has a smol amount of metadata that gets stamped into it: (that it backed a preauth key, when it was created, and the stable ID of the node that created it). EG: `map[purpose:pre-auth key wrapper_createtime:1677789199 wrapper_stableid:nwkXEu1CNTRL]`
3. There are two pieces to the wrapping information: the SigCredential signature (a `NodeKeySignature` that endorses a single-use public key for signing a node key), and the single-use private key (ed25519 private).
4. When the `controlclient.Direct` starts and sees the authkey, it uses the embedded SigCredential + private key to generate a SigRotation signature and sign itself.

Thoughts?

### DIFF
--- a/cmd/tailscale/cli/network-lock.go
+++ b/cmd/tailscale/cli/network-lock.go
@@ -15,6 +15,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/mattn/go-colorable"
 	"github.com/mattn/go-isatty"
@@ -39,6 +40,7 @@ var netlockCmd = &ffcli.Command{
 		nlDisablementKDFCmd,
 		nlLogCmd,
 		nlLocalDisableCmd,
+		nlTskeyWrapCmd,
 	},
 	Exec: runNetworkLockStatus,
 }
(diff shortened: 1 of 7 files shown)


twitchyliquid64:
For scalars: http://go/tailnet-lock-preauth-keys

willnorris:
Generally looks good to me.  We'll just need to make sure we never accidentally create a key with "--" (because some dash-delimited value is empty or something... it won't happen now, I'm just thinking about future changes), or maybe we use something a little more explicit here like "-NL-" ?

twitchyliquid64:
Sounds good, I've updated the suffix delimiter from `--` to `--TL`? That should:

1. Give us room to change the format in the future by changing the delimiter
2. Give us room to have different kinds of suffixes (idk why but lets imagine we did) and have them error out early

cmd/tailscale/cli/network-lock.go
```diff
@@ -622,3 +624,48 @@ func runNetworkLockLog(ctx context.Context, args []string) error {
...
+	// Generate a separate tailnet-lock key just for the credential signature.
+	// We use the free-form meta strings to mark a little bit of metadata about this
+	// key.
+	priv := key.NewNLPrivate()
+	k := tka.Key{
+		Kind:   tka.Key25519,
+		Public: priv.Public().Verifier(),
+		Votes:  1,
+		Meta: map[string]string{
+			"purpose":            "pre-auth key",
```
sailorfrag, line 655: might want to include the non-secret part (stableid) of the key

control/controlclient/direct.go
```diff
@@ -1713,6 +1731,50 @@ func (c *Direct) ReportHealthChange(sys health.Subsystem, sysErr error) {
...
+// If the authkey is wrapped, isWrapped returns true, along with the wrapping signature
+// and private key.
+func decodeWrappedAuthkey(key string, logf logger.Logf) (authKey string, isWrapped bool, sig *tka.NodeKeySignature, priv ed25519.PrivateKey) {
+	suffixIdx := strings.Index(key, "--TL")
+	if suffixIdx < 0 {
+		return key, false, nil, nil
+	}
+
+	suffix := key[suffixIdx+len("--TL"):]
+	sigIdx := strings.Index(suffix, "-")
```
sailorfrag, line 1746: also a good candidate for `strings.Cut()`

sailorfrag (approved):

cmd/tailscale/cli/web_test.go
```diff
@@ -86,10 +86,9 @@ func TestQnapAuthnURL(t *testing.T) {
 		},
 		{
 			name: "err != nil",
-                        in:   "http://192.168.0.%31/",
+			in:   "http://192.168.0.%31/",
```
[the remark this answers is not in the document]
twitchyliquid64: Yeah, I think someone forgot to `go fmt` a change their merged.

cmd/tailscale/cli/network-lock.go
```diff
@@ -622,3 +624,48 @@ func runNetworkLockLog(ctx context.Context, args []string) error {
...
+			"wrapper_stableid":   string(st.Self.ID),
+			"wrapper_createtime": fmt.Sprint(time.Now().Unix()),
+		},
+	}
+
+	wrapped, err := localClient.NetworkLockWrapPreauthKey(ctx, args[0], priv)
```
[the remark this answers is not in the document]
twitchyliquid64: I'm hesitant to having an implementation in the client that expects any particular format because that locks us in in the future.

But I guess we could implement a `/machine` endpoint that returns 200 OK if a preauth key is 1) valid and 2) for the same tailnet?

cmd/tailscale/cli/network-lock.go
twitchyliquid64: I had this originally, but this then means we no longer treat the 'authkey' part of the wrapped authkey as an opaque string, which limits its format for compatibility for future changes.

Maybe we can embed a cryptographic hash of the authkey? thoughts?

twitchyliquid64: @willnorris Do you think its likely we'll majorly revamp the authkey format in the near future?

Maybe if I:
1. Check that it has a `tskey-auth-` prefix
2. Check that it has another `-` and the stableID bit ends in `CNTRL`

its fairly safe to embed the stableID bit into the generated tailnet lock key?

sailorfrag: > @willnorris Do you think its likely we'll majorly revamp the authkey format in the near future?
> 
> Maybe if I:
> 
> 1. Check that it has a `tskey-auth-` prefix
> 2. Check that it has another `-` and the stableID bit ends in `CNTRL`
> 
> its fairly safe to embed the stableID bit into the generated tailnet lock key?

Don't assume a particular suffix on the stableID, that will break for compliance-US and onprem.

willnorris: I suspect that tskey-{type}-{stableid} is unlikely to change anytime soon, so that should be a safe way to extract the stable id (until of course, it isn't).  But as Adrian said, certainly don't assume CTRL.

sailorfrag: Maybe verifying the prefix is enough, I think catching obvious mistakes has value but I agree that expecting a particular format is problematic and I don't like the idea of asking control about an auth key unless it's being used to auth. I guess having an auth key validity oracle API wouldn't be too bad but we'd need to be careful with the secret part and definitely never send the wrapper.

Oh, one thing it should reject is if the key is already wrapped.

twitchyliquid64: Done & done!

Issue #1748, opened 2021-04-20
FR: Automatically forward particular domains through subnet relay


sailorfrag:
**Is your feature request related to a problem? Please describe.**

A couple customers have recently requested a way to route connections to a 3rd party SaaS product via a subnet router. This is needed because the SaaS product is restricted to only allow incoming connections from a particular IP range (typically that of the office and/or a subnet router). That's not too hard if the SaaS product uses a small range of IPs that don't change but is difficult in the presence of location-aware DNS load balancers, CDNs, etc.

Configuring an exit node can achieve this goal but captures far more traffic than required.

**Describe the solution you'd like**

A nicely packaged feature that will automatically redirect traffic going to a particular domain through an exit node or subnet relay. Maybe it could look at the DNS responses within that domain and automatically add them as /32 routes (caution: with CDNs this could grab unrelated traffic), or maybe the request could be modified to return a special IP address that the relay can DNAT (caution: complex).

This could be a magical experience if done well.

**Describe alternatives you've considered**

It's possible to get this result through a combination of
1. Configure a proxy server on the subnet router (or something reachable via the subnet router) that will redirect requests to the SaaS product by DNS name.
2. Use split DNS to generate DNS responses that send all traffic for the SaaS product's domain to the proxy server. Or, I suppose, a PAC file could accomplish that too.

But it's clunky

<img src="https://frontapp.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_31sgh)

Silvian:
This is exactly what we all need!!!

bradfitz:
A number of customers have asked for this in the past and what we've told them as a workaround is to run their own DNS server with records for the service they want to proxy, pointing to a self-hosted rinetd or https://github.com/inetaf/tcpproxy proxying to the origin. It's involved, but works. Making that all automatic would be awesome.

I thought we had some related bugs in this space, but maybe not. #1543 is about adding custom DNS records but that's all I found.

sailorfrag:
I'm thinking about this some more and one of the cleaner ways to implement this might actually be to implement a "Magic HTTP(S) proxy" that would remove the need for DNS to be involved. Maybe we auto generate a proxy config for the browser to evaluate, maybe we just proxy all HTTP(S), not sure. But it sure makes it easy to control which subnet relay or exit node handles the traffic based on hostname.

jack1902:
[...]

PR #7432, opened 2023-03-02
cmd/sniproxy: add start of a tsnet-based SNI proxy

bradfitz:
$ curl https://canhazip.com/
    170.173.0.21
    $ curl --resolve canhazip.com:443:100.85.165.81 https://canhazip.com/
    34.223.127.151

Updates #1748

### DIFF
--- a/cmd/sniproxy/snipproxy.go
+++ b/cmd/sniproxy/snipproxy.go
@@ -1,0 +1,90 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+// The sniproxy is an outbound SNI proxy. It receives TLS connections over
+// Tailscale on one or more TCP ports and sends them out to the same SNI
+// hostname & port on the internet. It only does TCP.
+package main
+
+import (
+	"context"
+	"flag"
+	"log"
+	"net"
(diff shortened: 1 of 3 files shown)


cmd/sniproxy/snipproxy.go
```diff
@@ -0,0 +1,90 @@
...
+		c, err := ln.Accept()
+		if err != nil {
+			log.Fatal(err)
+		}
+		go s.serveConn(c)
+	}
+}
+
+func (s *server) serveConn(c net.Conn) {
+	addrPortStr := c.LocalAddr().String()
```
maisem, line 67: Probably could take this as an input from serve.

maisem (approved):

cmd/sniproxy/snipproxy.go
bradfitz: I feared there'd be start-up races. I wanted to wire up all that paths before our Tailscale IP is known.

maisem: I meant just the port

bradfitz: It can be N ports.

maisem: this is fine :)

Issue #7377, opened 2023-02-25
UPnP not working with TP-Link ER605 router


future10se:
### What is the issue?

I have a [TP-Link ER605 v2](https://www.tp-link.com/us/business-networking/omada-sdn-router/er605/) router running [firmware 2.1.2](https://www.tp-link.com/us/support/download/er605/v2/#Firmware). UPnP is enabled in the settings, and other devices/apps on my network [are able to add mappings](https://user-images.githubusercontent.com/109644501/221343880-1b394886-97b4-4d7e-8ed7-9a5273dc8035.png). It does not support NAT-PMP.

The Tailscale WebUI reports that [UPnP is unavailable](https://user-images.githubusercontent.com/109644501/221344387-924d47dd-ef36-4515-80bd-b9b6ddec2b94.png), and `tailscale netcheck` [shows the same](https://user-images.githubusercontent.com/109644501/221345181-44b56af5-f089-41af-acb0-9ef16bfabf27.png). I have tested this on multiple machines (macOS, Windows, Linux) running tailscale on this network.

As with #6833 and #6320, I have made a wireshark packet capture. I used `( (udp contains "HTTP/1.1") and (udp contains 0a:53:54:3a) ) or (http contains 55:50:6e) or (http contains 75:70:6e)` for the filter. During the capture, I relaunched Tailscale, and also used [Port Map](https://www.codingmonkeys.de/portmap/) to query the mappings on my router. (It uses `miniupnpc` under the hood.)

It might be worth noting that the ER605 is a multi-WAN load balancing router, but I currently have it set to route over one WAN interface (also, regardless of whether or not load balancing is active, my other devices are able to map over UPnP). It also seems that the ER605 [does not report an external IP](https://user-images.githubusercontent.com/109644501/221344708-e9c33563-1417-4bac-b814-500e37072a8b.png). I don't know if this is an issue as per #682.

Thank you!

### Steps to reproduce

1. Enable/toggle UPnP on the router settings.
2. Relaunch Tailscale.
3. Check UPnP availability.

### Are there any recent changes that introduced the issue?

N/A

### OS

Linux, macOS, Windows

### OS version

macOS Monterey 12.6.3, Windows 11, Debian 11 (bullseye)

### Tailscale version

1.36.2, 1.36.1

### Bug report

BUG-e4511fa4255388b7817aed70c08a0e8a2d39a1e5f87052bae3bab118ba464d44-20230225073056Z-f3095245493bc672

[...]

PR #7435, opened 2023-03-02
cmd/tailscale, cmd/tailscaled: move portmapper debugging into tailscale CLI

andrew-d:
The debug flag on tailscaled isn't available in the macOS App Store build, since we don't have a tailscaled binary; move it to the 'tailscale debug' CLI that is available on all platforms instead.

Updates #7377

Change-Id: I47bffe4461e036fab577c2e51e173f4003592ff7

### DIFF
--- a/cmd/tailscale/cli/debug.go
+++ b/cmd/tailscale/cli/debug.go
@@ -201,6 +201,18 @@ var debugCmd = &ffcli.Command{
 				return fs
 			})(),
 		},
+		{
+			Name:      "portmap",
+			Exec:      debugPortmap,
+			ShortHelp: "run portmap debugging debugging",
+			FlagSet: (func() *flag.FlagSet {
+				fs := newFlagSet("portmap")
+				fs.DurationVar(&debugPortmapArgs.duration, "duration", 5*time.Second, "timeout for port mapping")
+				fs.StringVar(&debugPortmapArgs.ty, "type", "", `portmap debug type (one of "", "pmp", "pcp", or "upnp")`)
+				fs.StringVar(&debugPortmapArgs.gwSelf, "gw-self", "", `override gateway and self IP (format: "gatewayIP/selfIP")`)
+				return fs
+			})(),
+		},
 	},
 }
 
@@ -787,5 +799,26 @@ func runCapture(ctx context.Context, args []string) error {
 	defer f.Close()
 	fmt.Fprintln(os.Stderr, "Press Ctrl-C to stop the capture.")
 	_, err = io.Copy(f, stream)
+	return err
+}
+
+var debugPortmapArgs struct {
+	duration time.Duration
+	gwSelf   string
+	ty       string
+}
+
+func debugPortmap(ctx context.Context, args []string) error {
+	rc, err := localClient.DebugPortmap(ctx,
+		debugPortmapArgs.duration,
+		debugPortmapArgs.ty,
+		debugPortmapArgs.gwSelf,
+	)
+	if err != nil {
+		return err
+	}
+	defer rc.Close()
+
+	_, err = io.Copy(os.Stdout, rc)
 	return err
 }
--- a/net/portmapper/portmapper.go
+++ b/net/portmapper/portmapper.go
@@ -28,16 +28,18 @@ import (
 	"tailscale.com/util/clientmetric"
 )
 
-// Debug knobs for "tailscaled debug --portmap".
-var (
+// DebugKnobs contains debug configuration that can be provided when creating a
+// Client. The zero value is valid for use.
+type DebugKnobs struct {
+	// VerboseLogs tells the Client to print additional debug information
+	// to its logger.
 	VerboseLogs bool
 
 	// Disable* disables a specific service from mapping.
-
 	DisableUPnP bool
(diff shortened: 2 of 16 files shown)


cmd/tailscale/cli/debug.go
```diff
@@ -789,3 +802,102 @@ func runCapture(ctx context.Context, args []string) error {
...
+	duration time.Duration
+	ty       string
+}
+
+func debugPortmap(ctx context.Context, args []string) error {
+	ctx, cancel := context.WithTimeout(ctx, debugPortmapArgs.duration)
+	defer cancel()
+
+	portmapper.VerboseLogs = true
+	switch debugPortmapArgs.ty {
```
andrew-d, line 816: This is the only substantive change to this function: using a command-line argument instead of an envknob.

cmd/tailscale/cli/debug.go
```diff
@@ -41,6 +42,7 @@ import (
 	"tailscale.com/types/logger"
 	"tailscale.com/util/must"
 	"tailscale.com/wgengine/capture"
+	"tailscale.com/wgengine/monitor"
```
mihaip, line 45: Was going to say that perhaps the debugging should still live in `tailscaled` (and be accessed via a LocalAPI endpoint) to avoid adding extra dependencies to `tailscale`, but my #7177 already brought this package in 🙈

mihaip (approved):

cmd/tailscale/cli/debug.go
bradfitz: Oh no! Let's keep the CLI small. It's been a constant battle. Hence depaware.

mihaip: I just need to use the type, I'll see about moving it behind an interface.

mihaip: Sent out #7438

andrew-d: Okay, I'll have to rewrite a bunch of the portmap code in order to move this to LocalAPI since the various `Disable*` options and the `VerboseLogs` flag are unsynchronized global variables, so we can't modify them in `tailscaled` concurrently between the LocalAPI endpoint and the actual for-real portmapper 😬

mihaip: Is it required? Seems like you could have a mutex that you hold while you invoke the portmapper (which sets the globals), so that you don't have to worry about concurrent invocations.

andrew-d: Okay, switched the debug preferences to an `atomic.Pointer` and then used that to override options while doing the debug run, and then moved the code to LocalAPI. PTAL?

cmd/tailscale/cli/debug.go
```diff
@@ -789,3 +801,31 @@ func runCapture(ctx context.Context, args []string) error {
...
+
+func debugPortmap(ctx context.Context, args []string) error {
+	vals := make(url.Values)
+	vals.Set("duration", debugPortmapArgs.duration.String())
+	vals.Set("type", debugPortmapArgs.ty)
+	if debugPortmapArgs.gwSelf != "" {
+		vals.Set("gateway_and_self", debugPortmapArgs.gwSelf)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "GET", "http://local-tailscaled.sock/localapi/v0/debug-portmap?"+vals.Encode(), nil)
```
mihaip, line 819: Can this be a `LocalClient` method (so that callers don't have to be aware of query argument mapping)? That's how all other CLI functionality that uses LocalAPI endpoints is implemented.

net/portmapper/portmapper.go
```diff
@@ -154,6 +157,9 @@ func (m *pmpMapping) Release(ctx context.Context) {
 // goroutine whenever the port mapping status has changed. If nil,
 // it doesn't make a callback.
 func NewClient(logf logger.Logf, onChange func()) *Client {
+	// Initialize the debug knobs if they're not set yet.
```
mihaip, line 160: Any chance we make the debug knobs be `NewClient` parameters instead (and store them on the `Client` instance instead of in globals). It seems like in all the places they're accessed we have either direct or indirect access to the `Client` to read them from.

andrew-d: Updated 👍

cmd/tailscale/cli/debug.go
andrew-d: Pushed; I had the method return an `io.ReadCloser` so that we can still stream logs from the server.

net/portmapper/portmapper.go
```diff
@@ -28,16 +28,15 @@ import (
 	"tailscale.com/util/clientmetric"
 )
 
-// Debug knobs for "tailscaled debug --portmap".
-var (
+type DebugKnobs struct {
```
mihaip, line 31: Brad would say that this needs docs since it's public

mihaip (approved):
Thanks for the cleanup!

net/portmapper/portmapper.go
andrew-d: We do really need a 🤖-Brad... added some 👍

PR #7438, opened 2023-03-03
net/sockstats: remove explicit dependency on wgengine/monitor

mihaip:
Followup to #7177 to avoid adding extra dependencies to the CLI. We instead declare an interface for the link monitor.


wgengine/monitor/monitor.go
```diff
@@ -52,7 +52,7 @@ type osMon interface {
 // changed. The changed parameter is whether the network changed
 // enough for interfaces.State to have changed since the last
 // callback.
-type ChangeFunc func(changed bool, state *interfaces.State)
+type ChangeFunc = func(changed bool, state *interfaces.State)
```
mihaip, line 55: Had to change this to a type alias since otherwise I was getting complaints that `Mon` was not satisfying the `LinkMonitor` interface.

bradfitz: Type aliases are kinda gross. They should only be temporary during multistep migrations.

bradfitz: You can put the ChangeFunc type in the interfaces package by type State.

mihaip: Sounds good, moved it to `interfaces.go`.

andrew-d (approved):

PR #7440, opened 2023-03-03
cmd/derper: randomize IPs on refreshBootstrapDNS

shayne:
This is to address a possible DNS failure on startup. Prior to this change IPv6 addresses would be listed first and the client dialer would fail for hosts without IPv6 connectivity.

### DIFF
--- a/cmd/derper/bootstrap_dns.go
+++ b/cmd/derper/bootstrap_dns.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"expvar"
 	"log"
+	"math/rand"
 	"net"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmd/derper/bootstrap_dns.go
```diff
@@ -52,6 +53,11 @@ func refreshBootstrapDNS() {
 	ctx, cancel := context.WithTimeout(context.Background(), refreshTimeout)
 	defer cancel()
 	dnsEntries := resolveList(ctx, strings.Split(*bootstrapDNS, ","))
+	// Randomize the order of the IPs for each name to avoid biasing the client
+	// to a particular IP (also v4 vs. v6).
```
maisem, line 57: clients bias to v6 IIRC

maisem (approved):

Issue #7346, opened 2023-02-22
Windows: [WSL2] Address MTU issue when running tailscaled under WSL2


shayne:
> **Note**: A meta issue to track efforts to improve the user experience of running `tailscaled` under WSL2.

Packet sizes and the MTU in WSL2 are still causing persistent issues that, ideally, are resolved, and at minimum, we address them with an in-process workaround.

As of late 2022, the state of MTU in WSL2 is: `eth0: ... mtu 1280 ...`. Please look over #4833 for full context.

The user experience is pretty bad and could be better as everything appears to work, and Tailscale can connect. Sending pings to nodes receive replies, but larger-packet-leaning protocols will hang, like SSH and TLS.

At a minimum, we should add a warning and suggestion to increase MTU (to what?) to the `$ tailscale status` output. I also have tailscale/tailscale-www#2443 open to add WSL2 install instructions and known issues to our docs.

At best, we should devise a workaround that makes it "just work," and we continue to track the progress of WSL2 and changes to the network stack.

cc/ @raggi @dblohm7

DentonGentry:
@s-aia as I think PMTU discovery is one of the potential ways to make this more smooth.

raggi:
The broader/general recommendation would be to run Tailscale on the Windows side by preference. WSL is changing in undocumented ways at a pace that means we'll always have trouble making sure it's ok. Users also don't really have control over what version of WSL they're running.

Reading up on various related bugs in the WSL bug tracker, it's become a real wild west. It seems that some other VPN software now reaches around and changes the (windows side) MTU of the interfaces too in order to (mistakenly) attempt to fix the problem. This causes more issues because there's no synchronization between the host side interface MTU and the guest side interface MTU - the current defaults (at time of writing) are 1500 host side and 1280 guest side.

I think the most safe thing we can probably do, but it still might break users arbitrarily is to have a check in tailscaled for:

- $WSL_INTEROP is set
- eth0 is the default interface
- eth0 has an MTU of exactly 1280

If all of the above conditions are met, then we raise eth0 MTU to 1360. This is a conservative strategy that will work in most cases, even if the user has secondary VPN software on the Windows host side that can not transport nearer ethernet frame sized packets, while providing exactly enough space for tailscaled to function correctly when the user sends 1280 byte packets on tailscale0.

[...]

Issue #4833, opened 2022-06-11
Cannot receive large packets from webserver running inside wsl2


mhlakhani:
### What is the issue?

I'm having connection issues talking to HTTP servers running inside WSL2 on a windows machine. The connection issues show up regardless of the host I use to connect (I tried connecting from a windows machine and an iphone).

According to the webserver logs, it's successfully returning a 200, but the client never receives it.

Small packets (< MTU) seem to work, anything larger breaks. I have a pretty clear repro for this but no idea for a fix.

Output of `ip -a`:

```
➜  ~ ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: bond0: <BROADCAST,MULTICAST,MASTER> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 3e:6e:7f:b4:09:10 brd ff:ff:ff:ff:ff:ff
3: dummy0: <BROADCAST,NOARP> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 92:05:fb:ae:bf:a0 brd ff:ff:ff:ff:ff:ff
4: tunl0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN group default qlen 1000
    link/ipip 0.0.0.0 brd 0.0.0.0
5: sit0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN group default qlen 1000
    link/sit 0.0.0.0 brd 0.0.0.0
6: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1280 qdisc mq state UP group default qlen 1000
    link/ether 00:15:5d:13:43:7f brd ff:ff:ff:ff:ff:ff
    inet 172.19.6.223/20 brd 172.19.15.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::215:5dff:fe13:437f/64 scope link
       valid_lft forever preferred_lft forever
7: tailscale0: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1280 qdisc fq_codel state UNKNOWN group default qlen 500
    link/none
[...]

PR #7441, opened 2023-03-03
version/distro,wgengine/router: raise WSL eth0 MTU when too low

raggi:
WSL has started to set the eth0 default route interface default to 1280 MTU, which is too low to carry 1280 byte packets from tailscale0 once wrapped in WireGuard. The change down to 1280 is very likely smaller than necessary for almost all users. We can not easily determine the ideal MTU, but if all the preconditions match, we raise the MTU to 1360, which is just enough for Tailscale traffic to work.

Updates #4833
Updates #7346

### DIFF
--- a/version/distro/distro.go
+++ b/version/distro/distro.go
@@ -32,6 +32,7 @@ const (
 )
 
 var distro lazy.SyncValue[Distro]
+var isWSL lazy.SyncValue[bool]
 
 // Get returns the current distro, or the empty string if unknown.
 func Get() Distro {
@@ -44,6 +45,15 @@ func Get() Distro {
 		default:
 			return Distro("")
 		}
+	})
+}
+
+// IsWSL reports whether we're running in the Windows Subsystem for Linux.
+func IsWSL() bool {
+	return runtime.GOOS == "linux" && isWSL.Get(func() bool {
+		// We could look for $WSL_INTEROP instead, however that may be missing if
+		// the user has started to use systemd in WSL2.
+		return have("/proc/sys/fs/binfmt_misc/WSLInterop") || have("/mnt/wsl")
 	})
 }
 
--- a/wgengine/router/router_linux.go
+++ b/wgengine/router/router_linux.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"net"
 	"net/netip"
 	"os"
 	"os/exec"
@@ -223,6 +224,8 @@ func newUserspaceRouterAdvanced(logf logger.Logf, tunname string, linkMon *monit
 		r.ipPolicyPrefBase = 1300
 		r.logf("mwan3 on openWRT detected, switching policy base priority to 1300")
 	}
+
+	r.fixupWSLMTU()
 
 	return r, nil
 }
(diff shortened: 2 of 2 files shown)


bradfitz:
typo "rrouter"

version/distro/distro.go
```diff
@@ -47,6 +48,14 @@ func Get() Distro {
 	})
 }
 
+func IsWSL() bool {
+	return isWSL.Get(func() bool {
+		// We could look for $WSL_INTEROP instead, however that may be missing if
+		// the user has started to use systemd in WSL2.
+		return runtime.GOOS == "linux" && have("/proc/sys/fs/binfmt_misc/WSLInterop") || have("/mnt/wsl")
```
bradfitz, line 55: The runtime.GOOS == "linux" part can go before line 52 (before the Get call) to deadcode more even if the IsWSL symbol makes it into code.

version/distro/distro.go
```diff
@@ -47,6 +48,14 @@ func Get() Distro {
 	})
 }
 
+func IsWSL() bool {
```
bradfitz, line 51: // IsWSL reports whether we're running in the Windows Subsystem For Linux.

version/distro/distro.go
shayne: FWIW I still see `WSL_INTEROP` with a vanilla Ubuntu distro with `systemd` enabled. 

Also:
`uname -r` => `5.15.90.1-microsoft-standard-WSL2`

raggi: Good call!

wgengine/router/router_linux.go
```diff
@@ -891,6 +894,46 @@ func (r *linuxRouter) downInterface() error {
 	return netlink.LinkSetDown(link)
 }
 
+// fixupWSLMTU fixes the MTU on the WSL eth0 interface to be at least 1360 bytes
+// iff the running system appears to be in a common/well-known WSL
```
bradfitz, line 898: I know what "iff" means but from 10 years of Go reviews, empirically nobody else does and we get "typo" fixes forever about them. So just spell it out. :(

bradfitz (approved):
Looks plausible but I don't really know what the WSL MTU situation is really.

shayne:
Tested on my WSL and works

raggi:
```
~ts/tailscale fixup-wsl-mtu % sudo ip link set eth0 mtu 1280                                     
~ts/tailscale fixup-wsl-mtu % ip link show eth0                                                  
6: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1280 qdisc mq state UP mode DEFAULT group default qlen 1000
    link/ether 00:15:5d:be:1f:8c brd ff:ff:ff:ff:ff:ff
~ts/tailscale fixup-wsl-mtu % go run -exec sudo ./cmd/tailscaled 2>/dev/null & sleep 10 && kill %1
[1] 59274
[1]  + terminated  go run -exec sudo ./cmd/tailscaled 2> /dev/null                                        
~ts/tailscale fixup-wsl-mtu % ip link show eth0                                                   
6: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1360 qdisc mq state UP mode DEFAULT group default qlen 1000
    link/ether 00:15:5d:be:1f:8c brd ff:ff:ff:ff:ff:ff
```

PR #7444, opened 2023-03-03
wgengine/router: drop non-Tailscale IPv6 traffic from ULA range

andrew-d:
This replicates the same behaviour in our IPv4 rules by dropping traffic from the IPv6 ULA range that Tailscale uses that does not come from Tailscale itself (i.e. does not have the fwmark).

Change-Id: I75bc09fab73b5171de09a6828549644637c8a495

### DIFF
--- a/wgengine/router/router_linux.go
+++ b/wgengine/router/router_linux.go
@@ -1300,7 +1300,7 @@ func (r *linuxRouter) addNetfilterBase4() error {
 	return nil
 }
 
-// addNetfilterBase4 adds some basic IPv6 processing rules to be
+// addNetfilterBase6 adds some basic IPv6 processing rules to be
 // supplemented by later calls to other helpers.
 func (r *linuxRouter) addNetfilterBase6() error {
 	// TODO: only allow traffic from Tailscale's ULA range to come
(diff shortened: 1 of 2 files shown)


danderson (approved):
Plausible, but yeah worth testing all our feature combinations.

wgengine/router/router_linux.go
```diff
@@ -1314,8 +1314,13 @@ func (r *linuxRouter) addNetfilterBase6() error {
 	if err := r.ipt6.Append("filter", "ts-forward", args...); err != nil {
 		return fmt.Errorf("adding %v in v6/filter/ts-forward: %w", args, err)
 	}
-	// TODO: drop forwarded traffic to tailscale0 from tailscale's ULA
-	// (see corresponding IPv4 CGNAT rule).
+
+	// Drop traffic from the Tailscale ULA range; by this rule, we know that
+	// it doesn't have our fwmark and is thus not from us and safe to drop.
+	args = []string{"-o", r.tunname, "-s", tsaddr.TailscaleULARange().String(), "-j", "DROP"}
```
twitchyliquid64, line 1320: I'm not sure I understand: what does this have to do with the fwmark? Should we be checking it in this rule?

twitchyliquid64: Answering my own question: an earlier rule matches on the fwmark so this rule will only get traffic not having the fwmark.

twitchyliquid64 (approved):

PR #7445, opened 2023-03-03
util/slicesx: add package for generic slice functions, use

andrew-d:
Now that we're using rand.Shuffle in a few locations, create a generic shuffle function and use it instead. While we're at it, move the interleaveSlices function to the same package for use.

Change-Id: I0b00920e5b3eea846b6cedc30bd34d978a049fd3


util/slicesx/slicesx.go
```diff
@@ -0,0 +1,44 @@
...
+	ret = append(ret, a[i:]...)
+	ret = append(ret, b[i:]...)
+	return ret
+}
+
+// Shuffle randomly shuffles a slice in-place, similar to rand.Shuffle.
+func Shuffle[S ~[]T, T any](s S) {
+	// TODO(andrew): use a pooled Rand?
+
+	// This is the same Fisher-Yates shuffle implementation as rand.Shuffle
```
maisem, line 33: Can we just call that instead of reimplementing?

andrew-d: Per https://godbolt.org/, the `swap` function doesn't get inlined, so you pay a function call penalty per element in the array; I figured that this was (a) straightforward enough, and (b) a well-known algorithm that's unlikely to change, so it was worth copying here to ensure that everything is inlined.

maisem (approved):

PR #7447, opened 2023-03-03
net/interfaces: always return an IPv4 LikelyHomeRouterIP

andrew-d:
We weren't filtering out IPv6 addresses from this function, so we could be returning an IPv4 gateway IP and an IPv6 self IP. Per the function comments, only return IPv4 addresses for the self IP.

Change-Id: If19a4aadc343fbd4383fc5290befa0eff006799e


bradfitz (approved):

crawshaw (approved):

crawshaw:
Future work: how could we write a test for this?

maisem (approved):

PR #7449, opened 2023-03-04
net/interfaces: add better test for LikelyHomeRouterIP

andrew-d:
Return a mock set of interfaces and a mock gateway during this test and verify that LikelyHomeRouterIP returns the outcome we expect. Also verify that we return an error if there are no IPv4 addresses available.

Follow-up to #7447

Change-Id: I8f06989e7f1f0bebd108861cbff17b820ed2e6e4

### DIFF
--- a/net/interfaces/interfaces_test.go
+++ b/net/interfaces/interfaces_test.go
@@ -8,6 +8,8 @@ import (
 	"net"
 	"net/netip"
 	"testing"
+
+	"tailscale.com/tstest"
 )
 
 func TestGetState(t *testing.T) {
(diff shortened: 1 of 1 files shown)


net/interfaces/interfaces_test.go
```diff
@@ -37,12 +37,107 @@ func TestGetState(t *testing.T) {
...
 	t.Logf("myIP = %v; gw = %v", my, gw)
+
+	if want := netip.MustParseAddr("192.168.7.1"); gw != want {
+		t.Errorf("got gateway %v; want %v", gw, want)
+	}
+	if want := netip.MustParseAddr("192.168.7.100"); my != want {
+		t.Errorf("got self IP %v; want %v", my, want)
+	}
+
+	t.Run("Invalid", func(t *testing.T) {
```
maisem, line 118: add comment what this is supposed to test.

net/interfaces/interfaces_test.go
```diff
@@ -37,12 +37,107 @@ func TestGetState(t *testing.T) {
...
+	altNetInterfaces = func() ([]Interface, error) {
+		return mockInterfaces, nil
+	}
+
+	// Mock out the likelyHomeRouterIP to return a known gateway.
+	oldGw := likelyHomeRouterIP
+	t.Cleanup(func() { likelyHomeRouterIP = oldGw })
+	likelyHomeRouterIP = func() (netip.Addr, bool) {
+		return netip.MustParseAddr("192.168.7.1"), true
+	}
```
maisem, line 103: we have this in so many places, maybe time for tstest.Replace or something.

net/interfaces/interfaces_test.go
andrew-d: Done, also clarified test name

crawshaw (approved):
Thanks!

PR #7451, opened 2023-03-04
tstest: add method to Replace values for tests

maisem:
We have many function pointers that we replace for the duration of test and restore it on test completion, add method to do that.

### DIFF
--- a/tstest/tstest.go
+++ b/tstest/tstest.go
@@ -6,11 +6,28 @@ package tstest
 
 import (
 	"context"
+	"testing"
 	"time"
 
 	"tailscale.com/logtail/backoff"
 	"tailscale.com/types/logger"
 )
+
+// Replace replaces the value of target with val.
+// The old value is restored when the test ends.
+func Replace[T any](t *testing.T, target *T, val T) {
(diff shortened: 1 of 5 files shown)


tstest/tstest.go
```diff
@@ -6,12 +6,31 @@ package tstest
...
 	"time"
 
 	"tailscale.com/logtail/backoff"
 	"tailscale.com/types/logger"
 )
 
+// Replace replaces the value of target with val.
+// It returns a function that restores target to its original value.
+// After test t completes, the restore function is called automatically.
+func Replace[T any](t *testing.T, target *T, val T) (restore func()) {
```
crawshaw, line 19: You're not calling restore yet anywhere directly, are you planning to soon? If not I'd leave it out, slightly simpler function signature makes this all easier to read and reason about.

andrew-d (approved):
Nice–this is much much cleaner 🎉

crawshaw (approved):

tstest/tstest.go
maisem: good point, removed and added a test.

PR #7452, opened 2023-03-04
ipn/ipnlocal: reject unmasked routes

andrew-d:
Change-Id: Ic804efd24f5f536de1f2c910de3a24372d48d54d


danderson (approved):

maisem (approved):
We should also reject advertising them if we don't already

PR #7455, opened 2023-03-04
all: use tstest.Replace more

maisem:


andrew-d (approved):

tstest/tstest.go
```diff
@@ -15,7 +15,7 @@ import (
 
 // Replace replaces the value of target with val.
 // The old value is restored when the test ends.
-func Replace[T any](t *testing.T, target *T, val T) {
+func Replace[T any](t testing.TB, target *T, val T) {
 	t.Helper()
 	if target == nil {
 		t.Fatalf("Replace: nil pointer")
```
andrew-d, line 21: I think the change to `testing.TB` means that static analysis can't prove that this never returns; you could add a `panic("unreachable")` after this if you wanted?

maisem: i added a `return` but `panic("unreachable")` is a better idea.

andrew-d (approved):

PR #7456, opened 2023-03-04
ipn/localapi: fix panic after handler returns

andrew-d:
Change-Id: I612f9ebf78d962e094bff908670b3ffd89f756e5

### DIFF
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -637,7 +637,10 @@ func (h *Handler) serveDebugPortmap(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var logLock sync.Mutex
+	var (
+		logLock     sync.Mutex
+		handlerDone bool
+	)
 	logf := func(format string, args ...any) {
 		if !strings.HasSuffix(format, "\n") {
 			format = format + "\n"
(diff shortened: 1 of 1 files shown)


ipn/localapi/localapi.go
```diff
@@ -637,21 +637,32 @@ func (h *Handler) serveDebugPortmap(w http.ResponseWriter, r *http.Request) {
...
+		handlerDone bool
+	)
 	logf := func(format string, args ...any) {
 		if !strings.HasSuffix(format, "\n") {
 			format = format + "\n"
 		}
 
 		logLock.Lock()
 		defer logLock.Unlock()
+		if handlerDone {
```
creachadair, line 651: I feel like something near here deserves a comment, it took me a while to figure out how we could be calling this from another goroutine (which I guess is from the callback to portmapper.NewClient, but that was not at all obvious).

creachadair (approved):

ipn/localapi/localapi.go
creachadair: Something along the lines of: "The callback may try to log after the request returns, and at that point w is invalid." or words to that effect.

andrew-d: Good call; added a comment with an explanation.

PR #7457, opened 2023-03-04
net/portmapper: don't pick external ports below 1024

andrew-d:
Some devices don't let you UPnP portmap a port below 1024, so let's just avoid that range of ports entirely.

Updates #7377

Change-Id: Ib7603b1c9a019162cdc4fa21744a2cae48bb1d86

### DIFF
--- a/net/portmapper/upnp.go
+++ b/net/portmapper/upnp.go
@@ -124,8 +124,15 @@ func addAnyPortMapping(
 			uint32(leaseDuration.Seconds()),
 		)
 	}
-	for externalPort == 0 {
-		externalPort = uint16(rand.Intn(65535))
+
+	// Some devices don't let clients add a port mapping for privileged
+	// ports (ports below 1024).
+	//
+	// Pick an external port that's greater than 1024 by getting a random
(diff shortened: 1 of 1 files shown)


net/portmapper/upnp.go
```diff
@@ -124,8 +124,15 @@ func addAnyPortMapping(
...
-		externalPort = uint16(rand.Intn(65535))
+
+	// Some devices don't let clients add a port mapping for privileged
+	// ports (ports below 1024).
+	//
+	// Pick an external port that's greater than 1024 by getting a random
+	// number in [0, 65535 - 1024] and then adding 1024 to it, shifting the
+	// range to [1024, 65535].
+	if externalPort < 1024 {
+		externalPort = uint16(rand.Intn(65535-1024) + 1024)
```
creachadair, line 135: Maybe this should be obvious, but I'm not sure how we avoid picking a port that's already in use. If we do will the mapping fail and retry?

andrew-d: Yep, there's no consistently-available method that can tell us if there's a conflicting external port; we'll pick a new port on the next iteration and hopefully eventually hit one that's not in use.

creachadair (approved):

PR #7458, opened 2023-03-04
net/portmapper: send UPnP protocol in upper-case

andrew-d:
We were previously sending a lower-case "udp" protocol, whereas other implementations like miniupnp send an upper-case "UDP" protocol. For compatibility, use an upper-case protocol instead.

Updates #7377

Change-Id: I4aed204f94e4d51b7a256d29917af1536cb1b70f

### DIFF
--- a/net/portmapper/upnp.go
+++ b/net/portmapper/upnp.go
@@ -44,11 +44,21 @@ type upnpMapping struct {
 	client upnpClient
 }
 
+// upnpProtocolUDP represents the protocol name for UDP, to be used in the UPnP
+// <AddPortMapping> message in the <NewProtocol> field.
+//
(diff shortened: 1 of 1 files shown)


bradfitz (approved):
"as well" can read like "in addition" instead of "instead"

net/portmapper/upnp.go
```diff
@@ -44,11 +44,18 @@ type upnpMapping struct {
 	client upnpClient
 }
 
+// NOTE: this must be an upper-case string, or certain routers will reject the
```
bradfitz, line 47: If already writing such a big doc, start with normal form saying its name and what it is before getting into warnings?

Issue #5871, opened 2022-10-07
FR: tsnet: ListenPacket


Merovius:
### What are you trying to do?

I'm trying to use `tsnet` and listen for UDP packets. Specifically, I'd like to run a DNS server on my tailnet. I want to use `tsnet` because it wouldn't require a privileged user and wouldn't conflict with `resolved` and running on gokrazy needs a self-contained binary.

If I try `(*Server).Listen("udp", "100.x.y.z:1234")`, the `net.Listener` is not a `net.PacketConn`, so it can't be used with packages expecting that interface.

I can use `(*Server).Dial(ctx, "udp", "100.x.y.z:1234")` and type-assert to get a `net.PacketConn`, but the local port of that socket is apparently randomly chosen. I would like to be able to listen on a known port.

### How should we solve this?

I'd suggest adding a method `ListenPacket(net, addr string) (net.PacketConn, error)`, which returns a `net.PacketConn`. Alternatively, the `net.Listener` returned by `Listen` could implement `net.PacketConn` if the network is `"udp"`, so that it can be type-asserted.

### What is the impact of not solving this?

I'm not sure I can do what I want without this.

DentonGentry:
https://github.com/tailscale/tailscale/pull/6293 made Listen() reject "udp", since it didn't work.

I think we'd add the handling for UDP here: https://github.com/tailscale/tailscale/blob/1410682fb68e650786940811784c835ddb786e5a/tsnet/tsnet.go#L646

DentonGentry:
In the same way that netstack.go provides a `ForwardTCPIn` field to override `forwardTCP`'s default handling, we'll need to add a ForwardUDPIn and hook it up in `acceptUDP`.

https://github.com/tailscale/tailscale/blob/a9c17dbf939049ba7fc8f0be0bc1cb3510753287/wgengine/netstack/netstack.go#L80

https://github.com/tailscale/tailscale/blob/a9c17dbf939049ba7fc8f0be0bc1cb3510753287/wgengine/netstack/netstack.go#L989

wizzard0:
Loosely related: exit node support with tsnet? (this way, I could make #7766 and #4073 in userland via LD_PRELOAD/DYLD_INSERT_LIBRARIES/Detours)

DentonGentry:
I don't really see how exit node support could work with tsnet: it provides a socket abstraction. The application code uses the tsnet socket to listen for connections from the tailnet or make connections to the tailnet.

[...]

PR #7465, opened 2023-03-05
wgengine/netstack: add support for custom UDP flow handlers

bradfitz:
To be used by tsnet and sniproxy later.

Updates #5871
Updates #1748

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -41,6 +41,7 @@ import (
 	"tailscale.com/net/tsdial"
 	"tailscale.com/smallzstd"
 	"tailscale.com/types/logger"
+	"tailscale.com/types/nettype"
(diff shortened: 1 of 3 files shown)


bradfitz:
Sorry @DentonGentry, after I wrote this and was looking up the bug number to reference in the commit message, I saw you'd also started working on this. I don't want to copy the ForwardTCPIn signature, though, as it was kinda broken to begin with. (It doesn't provide a way to reject incoming flows)

DentonGentry:
Mine doesn't actually work anyway: it wants to return a new PacketConn for every packet, which isn't right.

I'll focus on the DNS server (last commit in https://github.com/tailscale/tailscale/compare/main...dgentry/sniproxy-dns) and rebase it atop whatever develops here.

DentonGentry (approved):

PR #7467, opened 2023-03-05
tsnet: add UDP support to Server.Listen

bradfitz:
No ListenPacket support yet, but Listen with a udp network type fit easier into netstack's model to start.

Then added an example of using it to cmd/sniproxy with a little udp :53 handler.

No tests in tsnet yet because we don't have support for dialing over UDP in tsnet yet. When that's done, a new test can test both sides.

Updates #5871
Updates #1748

### DIFF
--- a/cmd/sniproxy/snipproxy.go
+++ b/cmd/sniproxy/snipproxy.go
@@ -18,6 +18,7 @@ import (
 	"tailscale.com/client/tailscale"
 	"tailscale.com/net/netutil"
 	"tailscale.com/tsnet"
+	"tailscale.com/types/nettype"
 )
 
 var ports = flag.String("ports", "443", "comma-separated list of ports to proxy")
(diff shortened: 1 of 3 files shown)


cmd/sniproxy/snipproxy.go
```diff
@@ -63,6 +71,25 @@ func (s *server) serve(ln net.Listener) {
...
+	}
+}
+
+func (s *server) serveDNSConn(c nettype.ConnPacketConn) {
+	defer c.Close()
+	c.SetReadDeadline(time.Now().Add(5 * time.Second))
+	buf := make([]byte, 1500)
+	n, err := c.Read(buf)
+	log.Printf("got DNS packet: %q, %v", buf[:n], err)
+	// TODO: rest of the owl
```
crawshaw, line 90: 😂

crawshaw (approved):
Woo!

PR #7468, opened 2023-03-05
net/socks5: add password auth support

crawshaw:
Conforms to RFC 1929.

To support Java HTTP clients via libtailscale, who offer no other reliable hooks into their sockets.

### DIFF
--- a/net/socks5/socks5.go
+++ b/net/socks5/socks5.go
@@ -25,14 +25,19 @@ import (
 	"tailscale.com/types/logger"
 )
 
+// Authentication METHODs described in RFC 1928, section 3.
 const (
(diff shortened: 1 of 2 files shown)


net/socks5/socks5.go
```diff
@@ -83,6 +88,10 @@ type Server struct {
 	// Dialer optionally specifies the dialer to use for outgoing connections.
 	// If nil, the net package's standard dialer is used.
 	Dialer func(ctx context.Context, network, addr string) (net.Conn, error)
+
```
bradfitz, line 91: Typo

bradfitz (approved):

PR #7469, opened 2023-03-05
tsnet: generalize loopback listener to include SOCKS5

crawshaw:
Some languages do not give you any useful access to the sockets underlying their networking packages. E.g. java.net.http.HttpClient provides no official access to its dialing logic.

...but everyone supports proxies. So add a SOCKS5 proxy on the listener we are already running.

(The function being revamped is very new, I only added it in the last week and it wasn't part of any release, so I believe it is fine to redo its function signature.)

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -38,6 +38,8 @@ import (
 	"tailscale.com/logtail"
 	"tailscale.com/logtail/filch"
 	"tailscale.com/net/memnet"
+	"tailscale.com/net/proxymux"
+	"tailscale.com/net/socks5"
 	"tailscale.com/net/tsdial"
 	"tailscale.com/smallzstd"
 	"tailscale.com/types/logger"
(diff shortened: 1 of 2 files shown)


tsnet/tsnet.go
```diff
@@ -91,22 +93,23 @@ type Server struct {
...
+	lb               *ipnlocal.LocalBackend
+	netstack         *netstack.Impl
+	linkMon          *monitor.Mon
+	rootPath         string // the state directory
+	hostname         string
+	shutdownCtx      context.Context
+	shutdownCancel   context.CancelFunc
+	proxyCred        string                 // SOCKS5 proxy auth for loopbackListener
+	localAPICred     string                 // basic auth password for loopbackListener
+	loopbackListener net.Listener           // optional loopback, restricted to PID
```
maisem, line 107: say that its used for localapi and proxies

maisem (approved):

Issue #3847, opened 2022-01-30
FR: Multiple services on a single host using separate IPs, SNI proxy, or other


stefangrosaru:
### What are you trying to do?

I'm trying to access multiple services on the same machine over http. This machine runs traefik as a reverse proxy which exposes multiple web services over HTTP (port 80). 

### How should we solve this?

It would be useful if the MagicDNS entries could extend and cover all subdomains on a machine.

Example, I have a linux host called server, which exposes two services called service1 and service2, a reverse proxy sits in front of these services and forwards HTTP requests based on hostname so that service1.server goes to service1 and service2.server goes to server2.

### What is the impact of not solving this?

The only workaround at the moment is to use different ports for these services and this makes it difficult to work with reverse proxies. Also remembering subdomains is easier than remembering specific port numbers.

DentonGentry:
To set expectations: we're more likely to allow multiple certificates to be issues for a node, and allow different services to be muxed based on SNI.

AlexPygin:
Is this expected to be completed soon? Today, I realized how challenging it is to do basically the same thing called out in the original issue.

I want subdomains so 1Password will suggest a single login per port/subdomain, I can generate a cert so I don't always see "Not Secure", and so I can reverse proxy my http services.

DentonGentry:
I promise that we will update the bug when work begins in earnest. This is not currently at the top of the list to work on.

mholt:
Hi there, we have Caddy users asking for this feature as well: https://caddy.community/t/reverse-proxy-and-https-for-multiple-services-over-tailscale/16287

Basically they're running multiple services on a single host and would like to use subdomains to address them.

I also would like to personally +100 the usecase of 1Password so that it can more easily distinguish which service is being accessed.

I realize from #1196 that you want to reserve the next subdomain label in case it's needed for future features, but in that case could I suggest a single subdomain label for custom subdomains, and then yet another label for the actual subdomains, e.g. `service.svc.host.random.ts.net`?

That's a bit long of course, but at least it's organized. :man_shrugging:  Just an idea.

06kellyjac:
I've also been running into this with tailscale magicdns + https. It's surprising how many programs/services dont support subpaths e.g. `host.random.ts.net/subpath`

riekusr:
having this issue now as well.

Joshuapwilley:
Also having this issue. Just chiming in so hopefully this issue gets bumped closer to the top of the list.

[...]

PR #7470, opened 2023-03-06
cmd/sniproxy: draw the rest of the DNS owl.

DentonGentry:
Add a DNS server which always responds as its own IP addresses.

Additionally add a tsnet TailscaleIPs() function to return the IP addresses, both IPv4 and IPv6.

### DIFF
--- a/cmd/sniproxy/snipproxy.go
+++ b/cmd/sniproxy/snipproxy.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/net/dns/dnsmessage"
 	"inet.af/tcpproxy"
 	"tailscale.com/client/tailscale"
 	"tailscale.com/net/netutil"
@@ -22,6 +23,8 @@ import (
 )
 
 var ports = flag.String("ports", "443", "comma-separated list of ports to proxy")
+
+var tsMBox = dnsmessage.MustNewName("support.tailscale.com.")
 
 func main() {
 	flag.Parse()
@@ -86,8 +89,29 @@ func (s *server) serveDNSConn(c nettype.ConnPacketConn) {
 	c.SetReadDeadline(time.Now().Add(5 * time.Second))
 	buf := make([]byte, 1500)
 	n, err := c.Read(buf)
-	log.Printf("got DNS packet: %q, %v", buf[:n], err)
-	// TODO: rest of the owl
+	if err != nil {
+		log.Printf("c.Read failed: %v\n ", err)
+		return
+	}
+
+	var msg dnsmessage.Message
+	err = msg.Unpack(buf[:n])
+	if err != nil {
+		log.Printf("dnsmessage unpack failed: %v\n ", err)
+		return
+	}
+
+	buf, err = s.dnsResponse(&msg)
+	if err != nil {
+		log.Printf("s.dnsResponse failed: %v\n", err)
+		return
+	}
+
+	_, err = c.Write(buf)
+	if err != nil {
+		log.Printf("c.Write failed: %v\n", err)
+		return
+	}
 }
 
 func (s *server) serveConn(c net.Conn) {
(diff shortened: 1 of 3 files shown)


tsnet/tsnet.go
```diff
@@ -323,6 +323,17 @@ func (s *Server) doInit() {
 	}
 }
 
+func (s *Server) TailscaleIPs() []netip.Addr {
```
maisem, line 326: @crawshaw wanted to do something similar but I convinced him not to. If we really want this API to exist, it should take a context, have docs and return the error.

cmd/sniproxy/snipproxy.go
```diff
@@ -85,9 +90,30 @@ func (s *server) serveDNSConn(c nettype.ConnPacketConn) {
 	defer c.Close()
 	c.SetReadDeadline(time.Now().Add(5 * time.Second))
 	buf := make([]byte, 1500)
-	n, err := c.Read(buf)
-	log.Printf("got DNS packet: %q, %v", buf[:n], err)
-	// TODO: rest of the owl
+	_, err := c.Read(buf)
```
maisem, line 93: you need to get `n` and then use it below in `msg.Unpack(buf[:n])`

tsnet/tsnet.go
bradfitz, line 326: Needs docs

tsnet/tsnet.go
```diff
@@ -323,6 +323,17 @@ func (s *Server) doInit() {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer cancel()
+
+	st, err := s.localClient.Status(ctx)
```
bradfitz, line 330: This is much heavier than you need. StatusWithoutPeers is cheaper but still not as cheap as calling NetMap method on s.lb

cmd/sniproxy/snipproxy.go
```diff
@@ -107,11 +133,88 @@ func (s *server) serveConn(c net.Conn) {
...
+	if err != nil {
+		return
+	}
+
+	buf, err = resp.Finish()
+	if err != nil {
+		return
+	}
+
+	return
```
maisem, line 219: return resp.Finish()

tsnet/tsnet.go
maisem: Oh If we go the `s.lb.Netmap` route then we can skip the context and the error. Instead just document that it returns `nil` if the server isn't running or IPs are unavailable.

cmd/sniproxy/snipproxy.go
```diff
@@ -107,11 +133,88 @@ func (s *server) serveConn(c net.Conn) {
...
+	resp := dnsmessage.NewBuilder(buf,
+		dnsmessage.Header{
+			ID:            req.Header.ID,
+			Response:      true,
+			Authoritative: true,
+		})
+	resp.EnableCompression()
+
+	if len(req.Questions) == 0 {
+		buf, _ = resp.Finish()
```
maisem, line 168: return resp.Finish()

cmd/sniproxy/snipproxy.go
```diff
@@ -107,11 +133,88 @@ func (s *server) serveConn(c net.Conn) {
...
 		return &tcpproxy.DialProxy{
 			Addr:        net.JoinHostPort(sniName, port),
 			DialContext: dialer.DialContext,
 		}, true
 	})
 	p.Start()
 }
+
+// getAddresses returns the tsnet IP addresses of this process
+func (s *server) getAddresses() (ip4, ip6 netip.Addr) {
```
maisem, line 145: we can probably cache these, they won't change during the duration of the process.

tsnet/tsnet.go
bradfitz: Yeah. Seems good API to have and don't need context.

cmd/sniproxy/snipproxy.go
DentonGentry: The case where the IPs do change is the very first time the process starts, if authentication doesn't happen right away. Either they don't supply an authkey and they eventually get around to visiting the LoginURL, or the authkey takes some time to reach the coordination server.

Added caching but only if the addresses are valid.

cmd/sniproxy/snipproxy.go
DentonGentry: Done.

cmd/sniproxy/snipproxy.go
DentonGentry: Done.

cmd/sniproxy/snipproxy.go
DentonGentry: Done.

tsnet/tsnet.go
DentonGentry: Done.

tsnet/tsnet.go
DentonGentry: Done.

tsnet/tsnet.go
DentonGentry: Added docs.

cmd/sniproxy/snipproxy.go
```diff
@@ -11,17 +11,22 @@ import (
...
 	"tailscale.com/client/tailscale"
 	"tailscale.com/net/netutil"
 	"tailscale.com/tsnet"
 	"tailscale.com/types/nettype"
 )
 
-var ports = flag.String("ports", "443", "comma-separated list of ports to proxy")
+var (
+	ports  = flag.String("ports", "443", "comma-separated list of ports to proxy")
+	tsMBox = dnsmessage.MustNewName("support.tailscale.com.")
```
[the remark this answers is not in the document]
DentonGentry: Done.

cmd/sniproxy/snipproxy.go
```diff
@@ -23,6 +25,11 @@ import (
 
 var ports = flag.String("ports", "443", "comma-separated list of ports to proxy")
 
+var (
+	tsMBox   = dnsmessage.MustNewName("support.tailscale.com.")
+	ip4, ip6 netip.Addr
```
bradfitz, line 30: The use of these look like a data race. Let's just remove the caching.

tsnet/tsnet.go
```diff
@@ -323,6 +323,17 @@ func (s *Server) doInit() {
 	}
 }
 
+// TailscaleIPs returns IPv4 and IPv6 addresses for this node. If the node
+// has not yet joined a tailnet or is otherwise unaware of its own IP addrsses,
+// an empty slice will be returned.
+func (s *Server) TailscaleIPs() (addrs []netip.Addr) {
```
bradfitz, line 329: If this had the signature of returning v4 and v6 then it'd both not allocate and not risk aliasing the NetMap memory.

DentonGentry: Is it likely that we'll have multiple IP addresses for specific purposes? https://github.com/tailscale/tailscale/issues/3847

Maybe that would just mean changing the API (sorry, folks).

crawshaw: I came here to suggest the same as Brad, not for performance but because it would be much easier to reason about. I suspect users, like me, come looking for the single Tailscale IP, and learn it has some arbitrary number of IPs in a sort order I don't understand. v4/v6 would make it clear and what I should do.

cmd/sniproxy/snipproxy.go
DentonGentry: Removed getAddresses, now calls `s.ts.TailscaleIPs()`

cmd/sniproxy/snipproxy.go
DentonGentry: Removed caching, calls s.ts.TailscaleIPs()

tsnet/tsnet.go
DentonGentry: Done.

tsnet/tsnet.go
```diff
@@ -323,6 +323,24 @@ func (s *Server) doInit() {
 	}
 }
 
+// TailscaleIPs returns IPv4 and IPv6 addresses for this node. If the node
+// has not yet joined a tailnet or is otherwise unaware of its own IP addrsses,
+// netip.IsUnspecified() addresses will be returned.
```
maisem, line 328: Unspecified is for `0.0.0.0` or `::0`

You want `!IsValid`

https://pkg.go.dev/net/netip#Addr.IsUnspecified

maisem (approved):

tsnet/tsnet.go
DentonGentry: Updated description.

PR #7473, opened 2023-03-06
ipn/ipnlocal: stop netmap expiry timer when resetting control client

### LINKED ISSUE #7392: nil pointer dereference panic in ipn/ipnlocal/local.go
```
07:27:21 | setClientStatus: netmap expiry timer triggered after 45h45m26.64464318s
[...]
07:27:23 | panic: runtime error: invalid memory address or nil pointer dereference
07:27:23 | [signal 0xc0000005 code=0x0 addr=0x48 pc=0xb5d833]
07:27:23 |  
07:27:23 | goroutine 641623 [running]:
07:27:23 | tailscale.com/ipn/ipnlocal.(*LocalBackend).setClientStatus(0xc00011c380,  {{}, 0x0, 0x0, {0x0, 0x0}, {0x0, 0x0}, 0xc0007d3b00, 0x5, ...})
07:27:23 | tailscale.com@v1.1.1-0.20230217011852-d811c5a7f099/ipn/ipnlocal/local.go:985 +0x17f3
07:27:23 | tailscale.com/ipn/ipnlocal.(*LocalBackend).setClientStatus.func1()
07:27:23 | tailscale.com@v1.1.1-0.20230217011852-d811c5a7f099/ipn/ipnlocal/local.go:866 +0xf0
07:27:23 | created by time.goFunc
07:27:23 | time/sleep.go:176 +0x32
```

https://github.com/tailscale/tailscale/blob/d811c5a7f099cadb050f800ca1362766cf6ab02f/ipn/ipnlocal/local.go#L985
The control client is nil when `setClientStatus()` is being called recursively to expire nodes in the future via `time.AfterFunc()`

https://github.com/tailscale/tailscale/blob/d811c5a7f099cadb050f800ca1362766cf6ab02f/ipn/ipnlocal/local.go#L866

andrew-d:
This prevents a panic where we synthesize a new netmap in setClientStatus after we've shut down and nil'd out the controlclient, since that function expects to be called while connected to control.

Fixes #7392

Change-Id: Ib631eb90f34f6afa008d69bbb386f70da145e102

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -3682,6 +3682,21 @@ func (b *LocalBackend) resetControlClientLockedAsync() {
 	if b.cc == nil {
 		return
 	}
+
+	// When we clear the control client, stop any outstanding netmap expiry
+	// timer; synthesizing a new netmap while we don't have a control
+	// client will break things.
+	//
+	// See https://github.com/tailscale/tailscale/issues/7392
+	if b.nmExpiryTimer != nil {
+		b.nmExpiryTimer.Stop()
+		b.nmExpiryTimer = nil
+
+		// Also bump the epoch to ensure that if the timer started, it
+		// will abort.
+		b.numClientStatusCalls.Add(1)
+	}
+
 	go b.cc.Shutdown()
 	b.cc = nil
 	b.ccAuto = nil


ipn/ipnlocal/local.go
```diff
@@ -3682,6 +3682,21 @@ func (b *LocalBackend) resetControlClientLockedAsync() {
...
+	// client will break things.
+	//
+	// See https://github.com/tailscale/tailscale/issues/7392
+	if b.nmExpiryTimer != nil {
+		b.nmExpiryTimer.Stop()
+		b.nmExpiryTimer = nil
+
+		// Also bump the epoch to ensure that if the timer started, it
+		// will abort.
+		b.numClientStatusCalls.Add(1)
```
maisem, line 3697: Maybe reset to 0 instead?

maisem (approved):

ipn/ipnlocal/local.go
andrew-d: I thought about it, but I wasn't convinced that there wasn't some sort of race there. It's an `atomic.Uint32`, so it'll wrap around on overflow anyway, so we don't really need to worry about that particular problem.

PR #7482, opened 2023-03-07
tstime: rely on stdlib parse functionality

dsnet:
The time.Parse function has been optimized to the point where it is faster than our custom implementation. See upstream changes in:

* https://go.dev/cl/429862
* https://go.dev/cl/425197
* https://go.dev/cl/425116

Performance:

	BenchmarkGoParse3339/Z     38.75 ns/op    0 B/op    0 allocs/op
	BenchmarkGoParse3339/TZ    54.02 ns/op    0 B/op    0 allocs/op
	BenchmarkParse3339/Z       40.17 ns/op    0 B/op    0 allocs/op
	BenchmarkParse3339/TZ      87.06 ns/op    0 B/op    0 allocs/op

We can see that the stdlib implementation is now faster.


maisem (approved):

creachadair (approved):
Most excellent 😄

dsnet:
Wow. You're too fast. Approving before I can post in the #lgtm channel.

PR #7483, opened 2023-03-07
derp/derphttp: cleanup WebSocket connection on close

kylecarbs:
This was causing a leak in our CI!

### DIFF
--- a/derp/derphttp/derphttp_client.go
+++ b/derp/derphttp/derphttp_client.go
@@ -320,7 +320,7 @@ func (c *Client) connect(ctx context.Context, caller string) (client *derp.Clien
 		}
 		c.serverPubKey = derpClient.ServerPublicKey()
 		c.client = derpClient
-		c.netConn = tcpConn
+		c.netConn = conn
 		c.connGen++
 		return c.client, c.connGen, nil
 	case c.url != nil:


mihaip:
Thanks!

bradfitz:
This doesn't seem right to me. The `netConn` field is the lower-level TCP net.Conn elsewhere. It's just meant to forcible close the lower level. We don't need to send a high-level WebSocket (which ends up writing + waiting).

What was the "leak" you were seeing? Memory? Goroutine?

bradfitz:
@mihaip, does this change seem right to you?

I'd like to understand the leak. And ideally see a test.

mihaip:
> @mihaip, does this change seem right to you?

Well, `tcpConn` is nil in the websocket code path, so the code was definitely doing nothing before.

> I'd like to understand the leak. And ideally see a test.

I don't see any other code path for the websocket connection to be closed, so if you `Close()` the `derphttp.Client` it would leave it open (this is based on reading the code, @kylecarbs I would appreciate any real-world examples you have).

kylecarbs:
@bradfitz I believe the WebSocket was never being closed prior to this change. It's a goroutine leak, but a very small impact one. If we don't send a high-level close which incurs waiting then the DERP client `Close()` func won't fully clean up before returning, which seems odd. I feel it's idiomatic because the DERP client dials the WebSocket itself.

@mihaip I don't have any code samples that readily cause this, but if you're running on Linux then the `TS_DEBUG_DERP_WS_CLIENT` environment variable will force DERP over a WebSocket and you can see that post-close and pre this change, the socket won't be closed.

It's possible I'm misunderstanding something here, just let me know.

PR #7484, opened 2023-03-07
words: add word we forgot

bradfitz:
I explained this tails/scales list to my 5yo and he looked at me like it was the most obvious idea ever. Of course we'd make such lists at work!  What else do grown-ups do all day? And then he wouldn't stop talking about coelacanths and I had no clue what he was saying or how to spell it until I asked my phone and the phone apparently understood me and I realized it was a fish and he was helping me? I think?

### DIFF
--- a/words/scales.txt
+++ b/words/scales.txt
@@ -318,3 +318,4 @@ uaru
 vimba
 wahoo
 zebra
+coelacanth
--- a/words/tails.txt
+++ b/words/tails.txt
@@ -540,3 +540,4 @@ tailor
 uaru
 vimba
 wahoo
+coelacanth


words/scales.txt
```diff
@@ -318,3 +318,4 @@ uaru
 vimba
 wahoo
 zebra
+coelacanth
```
bradfitz, line 321: or do fish go in both files, @catzkorn?

catzkorn: If it has a tail and it has scales, it goes in both!

catzkorn (approved):
This has made my day ♥️

words/scales.txt
bradfitz: Done.

PR #7491, opened 2023-03-07
tstime/rate: add Value

dsnet:
Add Value, which measures the rate at which an event occurs, exponentially weighted towards recent activity.
It is guaranteed to occupy O(1) memory and operate in O(1) runtime.

### DIFF
--- a/tstime/rate/value.go
+++ b/tstime/rate/value.go
@@ -1,0 +1,183 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package rate
+
+import (
+	"fmt"
+	"math"
+	"sync"
+	"time"
+
+	"tailscale.com/tstime/mono"
+)
+
+// Value measures the rate at which events occur,
+// exponentially weighted towards recent activity.
+// It is guaranteed to occupy O(1) memory, operate in O(1) runtime,
+// and is safe for concurrent use.
+// The zero value is safe for immediate use.
+//
+// The algorithm is based on and semantically equivalent to
+// [exponentially weighted moving averages (EWMAs)],
+// but modified to avoid assuming that event samples are gathered
+// at fixed and discrete time-step intervals.
+//
+// In EWMA literature, the average is typically tuned with a λ parameter
+// that determines how much weight to give to recent event samples.
+// A high λ value reacts quickly to new events favoring recent history,
+// while a low λ value reacts more slowly to new events.
+// The EWMA is computed as:
+//
+//	zᵢ = λxᵢ + (1-λ)zᵢ₋₁
+//
+// where:
+//   - λ is the weight parameter, where 0 ≤ λ ≤ 1
+//   - xᵢ is the number of events that has since occurred
+//   - zᵢ is the newly computed moving average
+//   - zᵢ₋₁ is the previous moving average one time-step ago
+//
+// As mentioned, this implementation does not assume that the average
+// is updated periodically on a fixed time-step interval,
+// but allows the application to indicate that events occurred
+// at any point in time by simply calling Value.Add.
+// Thus, for every time Value.Add is called, it takes into consideration
+// the amount of time elapsed since the last call to Value.Add as
+// opposed to assuming that every call to Value.Add is evenly spaced
+// some fixed time-step interval apart.
+//
+// Since time is critical to this measurement, we tune the metric not
+// with the weight parameter λ (a unit-less constant between 0 and 1),
+// but rather as a half-life period t½. The half-life period is
+// mathematically equivalent but easier for humans to reason about.
+// The parameters λ and t½ and directly related in the following way:
+//
+//	t½ = -(ln(2) · ΔT) / ln(1 - λ)
+//
+//	λ = 1 - 2^-(ΔT / t½)
+//
+// where:
+//   - t½ is the half-life commonly used with exponential decay
+//   - λ is the unit-less weight parameter commonly used with EWMAs
+//   - ΔT is the discrete time-step interval used with EWMAs
+//
+// The internal algorithm does not use the EWMA formula,
+// but is rather based on [half-life decay].
+// The formula for half-life decay is mathematically related
+// to the formula for computing the EWMA.
+// The calculation of an EWMA is a geometric progression [[1]] and
+// is essentially a discrete version of an exponential function [[2]],
+// for which half-life decay is one particular expression.
+// Given sufficiently small time-steps, the EWMA and half-life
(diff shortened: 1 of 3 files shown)


tstime/rate/value.go
```diff
@@ -0,0 +1,112 @@
...
+	"math"
+	"time"
+
+	"tailscale.com/tstime/mono"
+)
+
+// Value measures the rate at which an event occurs,
+// exponentially weighted towards recent activity.
+// It is guaranteed to occupy O(1) memory and operate in O(1) runtime.
+type Value struct {
```
maisem, line 17: add a note that it is not threadsafe

maisem: same, but a mutex might be too expensive for my application.

dsnet: I went just a trivial `sync.Mutex` for now. The single-threaded performance degrades from 62ns to 70ns. Around 25ns of that is just calling `mono.Now()`.

We can optimize the concurrency aspect of this in a future PR. Any ideas that I can come up greatly complicates the code, so we should review the math portion of this by itself. That itself is a large tricky enough.

tstime/rate/value.go
```diff
@@ -0,0 +1,121 @@
...
+	//
+	// Since our events are decaying, we can divide the number of events
+	// by the total possible accumulated value, which we determine
+	// by integrating the half-life formula from t=0 until t=∞,
+	// assuming that N₀ is 1:
+	//	∫ N(t) dt = t½ / ln(2)
+	//
+	// Recall that the integral of a curve is the area under a curve,
+	// which carries the units of the X-axis multiplied by the Y-axis.
+	// In our case this would be the units "events · seconds".
```
crawshaw, line 109: well that's just sneaky

crawshaw (commented):
there's something a bit tricky going on here I don't quite understand. do you have a reference for this algorithm I could read? N₀ is both your initial quantity when calculating valueNow, but you update it whenever Add is called and I can't convince myself that the rate computation makes sense on top of the updated value.

dsnet:
> do you have a reference for this algorithm I could read?

Other than general resources on "half-life decay" and "exponential decay", not really. This algorithm came about because I was trying to [do some proportional decay](https://go.dev/play/p/ThceKBgvvME), and then @maisem suggested maybe instead doing something like `val*e^(-age)`, which reminded me that the suggested formula looked like atomic decay, which is what eventually led to what we got now.

There has got to be some formal computer science literature on solving this problem, but searches for things like "measure event rate algorithm" or "rate algorithm in O(1) memory and runtime" didn't really yield anything useful. You find [Stack Overflow posts](https://stackoverflow.com/questions/11731735/how-to-calculate-continuous-smooth-event-rate-based-on-event-times) that recommend decaying by 1/60th of the value, but such a solution assumes that the sample rate is fixed.

> N₀ is both your initial quantity when calculating valueNow, but you update it whenever Add is called and I can't convince myself that the rate computation makes sense on top of the updated value.

The way I reason about it is that `value` carries the units "events" and is an approximation on the total number of events still alive (i.e., has not decayed away). `r.valueNow` computes the total number of events alive right now after decaying away everything between `now` and `updated`. We add `n` since those events just happens right now and have not decayed away.

tstime/rate/value.go
```diff
@@ -0,0 +1,121 @@
...
+	HalfLife time.Duration
+
+	mu      sync.Mutex
+	updated mono.Time
+	value   float64 // adjusted count of events
+}
+
+// halfLife returns the half-life period in seconds.
+func (r *Value) halfLife() float64 {
+	halfLife := atomic.LoadInt64((*int64)(&r.HalfLife))
```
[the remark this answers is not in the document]
dsnet: Ooops. This is leftover from an attempt to make this thing fully depend on atomics, but then I reduced it back to just using a `sync.Mutex` and leave optimizations for the future.

crawshaw:
A common academic term for algorithms that process an event stream in some fixed memory window is "online." Searching for that didn't reveal anything exactly, but it did point to a popular finance technique called EWMA (exponential weighted moving average) for time series data.

https://www.itl.nist.gov/div898/handbook/pmc/section4/pmc431.htm

My first read-through of it gives me a better intuition for the updating of `value` you're doing, but it's been a long day. I'll think about it again in the morning and if it still makes sense will wave this PR on. :)

dsnet:
Interesting. I think @andrew-d pointed at something with EWMA, but I had no idea what the acronym meant (sometimes acronyms really get in the way of understanding something).

It does appear that EWMA and the algorithm here share similarities:
* EWMA appear to be defined according to a [decay constant](https://en.wikipedia.org/wiki/Exponential_smoothing#Time_constant), which is semantically equivalent to a half-life according to the [last half-life formula variant](https://en.wikipedia.org/wiki/Half-life#Formulas_for_half-life_in_exponential_decay). I find the "half-life" representation much more intuitive as a half-life of 1s tells me that our rate will be 50% accurate within 1s, 75% accurate within 2s, 87.5% accurate within 3s, and so on. The equivalent decay constant of 0.6931 tells me nothing unless I plot it out.
* A significant difference between our `rate.Value` and an EWMA is that our `rate.Value` takes variable time into consideration, while EWMA seems to heavily assume a fixed sampling period (e.g., the daily stock price). Our rate needs to be sensible even if there is a long period of silence between events or the occurrence of events is not uniformly spread out. For example, if we want to compute the rate of how often a node disconnects, we are not in control of when that event occurs. It is simpler and more efficient to call `rate.Value.Add(1)` whenever the node disconnects, than it is to have some background ticker that checks every T period whether the node is connected or not.

dsnet:
> Giving it a name other than `rate.Value`.

From a Go package perspective, this would introduce a naming inconsistency. At present, the package's naming convention is based on what they do (i.e., `Limiter` limits a rate, while `Value` measures a rate), rather than how they do it. Both fundamentally choose some algorithm (from among many) to do what it does, but documents which algorithm. For example, `Limiter` clearly documents that it uses the "Token bucket" algorithm.

What are your thoughts on keeping it simple like `rate.Value`, but writing a blog post and then pointing the documentation of `Value` at the blog post along with a reasonable algorithm name that we come up with?

Alternatively, we can just put `rate.Value` in a new package and thus absolve ourselves of any naming precedence.

crawshaw:
Documenting a Google-able algorithm name in the type comments is probably good enough. The problem I'm trying to solve is "oh a rate.Value, how do I interpret this data? let me search..."

maisem (approved):

crawshaw (approved):

dsnet:
Thanks for the reviews! I think this is the first Go package I wrote with more comments than code 😆

PR #7493, opened 2023-03-07
control/controlclient: cache control key

andrew-d:
This can allow us to continue to communicate with control when moving to a network that does TLS MiTM, which would otherwise prevent us from being able to fetch the Noise key and establish a connection.

Updates #3198 (sorta)

Change-Id: I52caf5079de744874a2bdd0c9ffb9e8f087ff8e0


control/controlclient/direct.go
```diff
@@ -1225,39 +1240,47 @@ func encode(v any, serverKey, serverNoiseKey key.MachinePublic, mkey key.Machine
...
-	}
-	if res.StatusCode != 200 {
-		return nil, fmt.Errorf("fetch control key: %d", res.StatusCode)
-	}
-	var out tailcfg.OverTLSPublicKeyResponse
-	jsonErr := json.Unmarshal(b, &out)
-	if jsonErr == nil {
-		return &out, nil
-	}
+func (c *Direct) loadServerPubKeys(ctx context.Context, httpc *http.Client, serverURL string) (*tailcfg.OverTLSPublicKeyResponse, error) {
```
maisem, line 1243: if moving this to the be on `*Direct`, reduce parameters that it can get access to directly. currently the `serverURL` parameter suggests that this can end up caching the key for different servers, but really there is only one server per `*Direct` client and it never changes

util/cache/disk.go
```diff
@@ -0,0 +1,109 @@
...
+
+package cache
+
+import (
+	"encoding/json"
+	"os"
+	"time"
+)
+
+// Disk is a cache that stores data in a file on-disk. It also supports
```
DentonGentry, line 12: Is there a concern about lateral movement, that anything which can write to this file can fool us into using a key which will match a hostile control server?

andrew-d: Yeah, right now this is using the memory-only cache for that reason, along with the more general question of cache expiry/etc. If we were going to switch to the Disk cache, we'd probably want to add some mitigations, like trying to load the key from the server on startup in all cases, ignoring the presence of the cache and the timestamp in it, and only fallback to the cached value if that fails.

andrew-d:
Also: I'm intending to merge this after the 1.38 branch-off, just to be safe and ensure we have lots of testing.

control/controlclient/direct.go
```diff
@@ -143,6 +147,10 @@ type Options struct {
 	// If we receive a new DialPlan from the server, this value will be
 	// updated.
 	DialPlan ControlDialPlanner
+
+	// ControlKeyCache caches Noise keys returned from control; if nil, no
+	// cache will be used.
+	ControlKeyCache cache.Cache[*tailcfg.OverTLSPublicKeyResponse]
```
maisem, line 153: Oh I just realized how this work, I had the wrong view initially

This won't work with FUS when working with profiles across say Tailscale and Headscale. It will cache the first entry and reuse for both instances

andrew-d: Hm, it looks like this worked largely due to an implementation detail; `resetForProfileChangeLockedOnEntry` calls `b.Start`, which calls `resetControlClientLockedAsync` and then re-creates the control client with a new in-memory-only cache. This won't work if we switch to a disk cache, though.

Pushed a change that keys based on the actual URL being fetched, instead.

util/cache/disk.go
andrew-d: Removed the disk cache entirely in the other PR

DentonGentry:
> Also: I'm intending to merge this after the 1.38 branch-off, just to be safe and ensure we have lots of testing.

Just to be clear: this should be submitted after the 1.56 branch this week, to go into 1.58.

bradfitz:
Closing, but will probably incorporate this or similar for #12639.

PR #7494, opened 2023-03-07
control/controlclient: improve handling of concurrent lite map requests

andrew-d:
Prior to this change, if we were in the middle of a lite map update we'd tear down the entire map session and restart it. With this change, we'll cancel an in-flight lite map request up to 10 times and restart before we tear down the streaming map request. We tear down everything after 10 retries to ensure that a steady stream of calls to sendNewMapRequest doesn't fail to make progress by repeatedly canceling and restarting.

Change-Id: I9392bf8cf674e7a58ccd1e476039300a359ef3b1

### DIFF
--- a/control/controlclient/auto.go
+++ b/control/controlclient/auto.go
@@ -59,15 +59,17 @@ type Auto struct {
 
 	mu sync.Mutex // mutex guards the following fields
 
-	paused          bool // whether we should stop making HTTP requests
-	unpauseWaiters  []chan struct{}
-	loggedIn        bool       // true if currently logged in
-	loginGoal       *LoginGoal // non-nil if some login activity is desired
-	synced          bool       // true if our netmap is up-to-date
-	inPollNetMap    bool       // true if currently running a PollNetMap
-	inLiteMapUpdate bool       // true if a lite (non-streaming) map request is outstanding
-	inSendStatus    int        // number of sendStatus calls currently in progress
-	state           State
+	paused               bool // whether we should stop making HTTP requests
+	unpauseWaiters       []chan struct{}
+	loggedIn             bool       // true if currently logged in
+	loginGoal            *LoginGoal // non-nil if some login activity is desired
+	synced               bool       // true if our netmap is up-to-date
+	inPollNetMap         bool       // true if currently running a PollNetMap
+	inLiteMapUpdate      bool       // true if a lite (non-streaming) map request is outstanding
+	liteMapUpdateCancel  func()     // cancels a lite map update
+	liteMapUpdateCancels int        // how many times we've canceled a lite map update
+	inSendStatus         int        // number of sendStatus calls currently in progress
+	state                State
 
 	authCtx    context.Context // context used for auth requests
 	mapCtx     context.Context // context used for netmap requests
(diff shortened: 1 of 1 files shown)


control/controlclient/auto.go
```diff
@@ -171,27 +173,45 @@ func (c *Auto) sendNewMapRequest() {
...
 		c.mu.Unlock()
 		c.cancelMapSafely()
 		return
 	}
+	if c.inLiteMapUpdate {
+		if c.liteMapUpdateCancels > 10 {
+			// We've canceled 10 of these, and thus aren't making
+			// progress; fall back to tearing everything down.
+			c.mu.Unlock()
+			c.cancelMapSafely()
```
maisem, line 186: You probably want to clean up after yourself here like cancel the inprogress lite map update and reset the counter to 0

control/controlclient/auto.go
```diff
@@ -168,39 +170,64 @@ func (c *Auto) Start() {
...
 		t0 := time.Now()
 		err := c.direct.SendLiteMapUpdate(ctx)
 		d := time.Since(t0).Round(time.Millisecond)
 		c.mu.Lock()
 		c.inLiteMapUpdate = false
+		c.liteMapUpdateCancel = nil
 		c.mu.Unlock()
 		if err == nil {
 			c.logf("[v1] successful lite map update in %v", d)
+			c.liteMapUpdateCancels = 0
```
maisem, line 218: this happens outside of a lock

andrew-d: Note to self: figure out how to add a test for this path, since ideally the race detector would have caught this

control/controlclient/auto.go
```diff
@@ -168,39 +170,67 @@ func (c *Auto) Start() {
...
 	defer c.mu.Unlock()
 	c.inLiteMapUpdate = true
 	ctx, cancel := context.WithTimeout(c.mapCtx, 10*time.Second)
+	c.liteMapUpdateCancel = cancel
 	go func() {
 		defer cancel()
 		t0 := time.Now()
 		err := c.direct.SendLiteMapUpdate(ctx)
 		d := time.Since(t0).Round(time.Millisecond)
 		c.mu.Lock()
```
maisem, line 212: add a new line here to clarify the block that has the lock

maisem (approved):

control/controlclient/auto.go
```diff
@@ -168,39 +170,67 @@ func (c *Auto) Start() {
...
-	if !c.inPollNetMap || c.inLiteMapUpdate || !c.loggedIn {
+	// If we're not already streaming a netmap, then tear down everything
+	// and start a new stream (which starts by sending a new map request)
+	if !c.inPollNetMap || !c.loggedIn {
 		c.mu.Unlock()
 		c.cancelMapSafely()
 		return
 	}
 
+	if c.inLiteMapUpdate {
```
maisem, line 181: maybe a comment describing why

maisem: something like
```
// If we are already in process of doing a LiteMapUpdate, cancel it and try a new one.
// If this is the 10th time we have done this cancelation, tear down everything and start again
```

control/controlclient/auto.go
```diff
@@ -168,39 +170,67 @@ func (c *Auto) Start() {
 func (c *Auto) sendNewMapRequest() {
 	c.mu.Lock()
 
-	// If we're not already streaming a netmap, or if we're already stuck
-	// in a lite update, then tear down everything and start a new stream
-	// (which starts by sending a new map request)
-	if !c.inPollNetMap || c.inLiteMapUpdate || !c.loggedIn {
+	// If we're not already streaming a netmap, then tear down everything
+	// and start a new stream (which starts by sending a new map request)
+	if !c.inPollNetMap || !c.loggedIn {
```
maisem, line 175: reset `liteMapUpdateCancels` here too?

andrew-d: I actually moved the `liteMapUpdateCancels` reset into `cancelMapSafely`, since I can't think of a reason not to reset it there and (IMO) it's less likely to be forgotten. Thoughts on that?

maisem: yeah I was going to suggest something similar, sgtm

maisem (approved):

PR #7496, opened 2023-03-07
tsnet: only intercept TCP flows that have listeners

maisem:
Previously, it would accept all TCP connections and then close the ones it did not care about. Make it only ever accept the connections that it cares about.

### DIFF
--- a/tsnet/tsnet_test.go
+++ b/tsnet/tsnet_test.go
@@ -175,6 +175,11 @@ func TestConn(t *testing.T) {
 	if string(got) != want {
 		t.Errorf("got %q, want %q", got, want)
 	}
+
+	_, err = s2.Dial(ctx, "tcp", fmt.Sprintf("%s:8082", s1ip)) // some random port
+	if err == nil {
+		t.Fatalf("unexpected success; should have seen a connection refused error")
+	}
 }
 
 func TestLoopbackLocalAPI(t *testing.T) {
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -79,12 +79,18 @@ func init() {
 // and implements wgengine.FakeImpl to act as a userspace network
 // stack when Tailscale is running in fake mode.
 type Impl struct {
-	// ForwardTCPIn, if non-nil, handles forwarding an inbound TCP connection.
+	// GetTCPHandlerForFlow conditionally handles an incoming TCP flow for the
+	// provided (src/port, dst/port) 4-tuple.
 	//
-	// TODO(bradfitz): convert this to the GetUDPHandlerForFlow pattern below to
-	// provide mechanism for tsnet to reject a port other than accepting it and
-	// closing it.
-	ForwardTCPIn func(c net.Conn, port uint16)
+	// A nil value is equivalent to a func returning (nil, false).
+	//
+	// If func returns intercept=false, the default forwarding behavior (if
+	// ProcessLocalIPs and/or ProcesssSubnetIPs) takes place.
+	//
+	// When intercept=true, the behavior depends on whether the returned handler
+	// is non-nil: if nil, the connection is rejected. If non-nil, handler takes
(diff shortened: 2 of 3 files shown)


wgengine/netstack/netstack.go
```diff
@@ -913,13 +921,19 @@ func (ns *Impl) acceptTCP(r *tcp.ForwarderRequest) {
...
 
-	if ns.ForwardTCPIn != nil {
-		c := createConn()
-		if c == nil {
-			return
+	if ns.GetTCPHandlerForFlow != nil {
+		handler, ok := ns.GetTCPHandlerForFlow(clientRemoteAddrPort, dstAddrPort)
+		if ok {
+			if handler == nil {
+				r.Complete(true)
```
danderson, line 928: You complete(true) to refuse a connection? Weird :)

danderson (approved):

wgengine/netstack/netstack.go
andrew-d: This is extremely confusing 😓 I remember tripping over this when working on netstack previously 😵‍💫 The boolean parameter is at least named `sendReset`, I guess?

wgengine/netstack/netstack.go
```diff
@@ -913,13 +921,19 @@ func (ns *Impl) acceptTCP(r *tcp.ForwarderRequest) {
...
+				return
+			}
+			c := createConn()
+			if c == nil {
+				return
```
bradfitz, line 933: will createConn on failure r.Complete(true) to send the RST too?

tsnet/tsnet_test.go
```diff
@@ -175,6 +176,19 @@ func TestConn(t *testing.T) {
...
 		t.Errorf("got %q, want %q", got, want)
 	}
+
+	_, err = s2.Dial(ctx, "tcp", fmt.Sprintf("%s:8082", s1ip)) // some random port
+	if err != nil {
+		var oe *net.OpError
+		if !errors.As(err, &oe) {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		if oe.Err.Error() != "connection was refused" {
```
bradfitz, line 186: string compare is kinda gross. Windows might have a different message? Compare against the underlying syscall err numbers if possible?

bradfitz (approved):
Thanks for finishing!

wgengine/netstack/netstack.go
```diff
@@ -913,13 +921,19 @@ func (ns *Impl) acceptTCP(r *tcp.ForwarderRequest) {
...
+			}
+			handler(c)
```
andrew-d, line 935: I think you're missing a `return` after this?

wgengine/netstack/netstack.go
andrew-d: (I wrote this) Yep, it should:
https://github.com/tailscale/tailscale/blob/c09578d060e001453c43c78d492891ad92bc862e/wgengine/netstack/netstack.go#L817-L823

maisem: yeah, added a comment at the call site

wgengine/netstack/netstack.go
maisem: yep
```
func (r *ForwarderRequest) Complete(sendReset bool) {
```

tsnet/tsnet_test.go
maisem: sadly the underlying error is just a `*errors.errorString`

```
oe.Err is of type *errors.errorString: "connection was refused"; is syscall.ECONNREFUSED = false
```

maisem: Dropped the string comparison and just checking for errors (verified test fails w/o the change)

wgengine/netstack/netstack.go
maisem: Yep! fixed

andrew-d (approved):

PR #7498, opened 2023-03-08
ipn,log: add logger for sockstat deltas

willnorris:
Add logging of changes in sockstat values. Events are logged as JSON which adds a bit of overhead, but with the sparse logging, it's not too bad... here's a sample:

```json
{"t":1678487681796,"d":98,"s":{"1":[1064,934],"6":[113,0],"7":[365,0],"8":[960,220]}}
{"t":1678487681895,"d":100,"s":{"8":[600,440]}}
{"t":1678487681995,"d":99,"s":{"1":[1746,35],"6":[73,0],"7":[351,0],"8":[320,528]}}
{"t":1678487682095,"d":100,"s":{"2":[629,4688],"8":[0,572]}}
{"t":1678487682196,"d":100,"s":{"1":[0,88],"8":[0,220]}}
{"t":1678487682296,"d":99,"s":{"3":[3295,0],"8":[0,88]}}
{"t":1678487682396,"d":99,"s":{"3":[449,97]}}
```

This uses filch for now just to get something working, and that seems to work fine so far.  When we have a better idea of storage requirements, we can bump up the max filesize.  Logs are stored in the normal log directory, named `sockstats.log{1,2}.txt`

/cc @mihaip @dsnet


dsnet:
JSON isn't the most efficient format, but I'd recommend against doing something custom. It's worth considering switching our logs telemetry to something like CBOR, but that can be future work that spans a large number of components.

log/sockstatlog/logger.go
```diff
@@ -0,0 +1,124 @@
...
+// This method does not return.
+func (l *Logger) poll() {
+	enc := json.NewEncoder(l.logbuffer)
+	for t := range l.ticker.C {
+		stats := sockstats.Get()
+		diffstats, changed := delta(l.last, stats.Stats)
+		if changed {
+			entry := logEntry{
+				Start: l.lastTime.UnixMilli(),
+				End:   t.UnixMilli(),
```
mihaip, line 75: May want just a duration, so that we don't end up with two very long timestamps.

log/sockstatlog/logger.go
```diff
@@ -0,0 +1,124 @@
...
+// Zero values are omitted from the returned map, and changed indicates if any stats have changed during the interval.
+func delta(a, b statMap) (diffstats statMap, changed bool) {
+	diffstats = make(statMap, len(b))
+	for label, bs := range b {
+		var labelChanged bool
+		as := a[label]
+		stat := sockstats.SockStat{
+			TxBytes:            bs.TxBytes - as.TxBytes,
+			RxBytes:            bs.RxBytes - as.RxBytes,
+			TxBytesByInterface: make(map[string]int64, len(bs.TxBytesByInterface)),
```
mihaip, line 102: Storing the per-interface information seems like it'll be pretty bulky. Could we instead return the current interface in sockstats.SockStat, and include it in the delta if it changes? We can assume that all transfers in a delta were on the same interface, which is basically true at the 10hz sampling rate.

log/sockstatlog/logger.go
```diff
@@ -0,0 +1,124 @@
...
+// delta calculates the delta stats between two snapshots.
+// a and b are maps of SockStat values containing total bytes transferred.
+// The returned map contains the bytes transferred between the two snapshots.
+// Zero values are omitted from the returned map, and changed indicates if any stats have changed during the interval.
+func delta(a, b statMap) (diffstats statMap, changed bool) {
+	diffstats = make(statMap, len(b))
+	for label, bs := range b {
+		var labelChanged bool
+		as := a[label]
+		stat := sockstats.SockStat{
```
mihaip, line 99: I assume this always does an alloc to create the struct, even if we end up throwing it away because nothing changed.

May want to have a fast check of old/new TxBytes/RxBytes instead.

mihaip (commented):
My data estimate was assuming an efficient binary encoding: 12 counters * 5 bytes/counter (ID + value) * 2 (for tx/rx) * 10 hz * 86400 = 104MB/day of data. This is in some ways better (because it's deltas, and we log nothing if not actively used), but in some ways worse (because of the JSON overhead).

It's probably OK for experimenting in unstable builds though.

log/sockstatlog/logger.go
```diff
@@ -0,0 +1,124 @@
...
+	go l.poll()
+}
+
+// poll fetches the current socket stats at the configured time interval,
+// calculates the delta since the last poll, and logs any non-zero values.
+// This method does not return.
+func (l *Logger) poll() {
+	enc := json.NewEncoder(l.logbuffer)
+	for t := range l.ticker.C {
+		stats := sockstats.Get()
```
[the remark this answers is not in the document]
willnorris: marking as resolved since you're handling this in #7517

log/sockstatlog/logger.go
willnorris: This is particularly bad when we have multiple labels logged in a given time period, since each label repeats the interface.  However, I worry about dropping the interface, since it means you can't necessarily pick a random entry and know what interface it was on.  What about just logging it once per entry?

```
{"t":1678419626985,"d":99,"i":"en0","s":{"0":{"tx":55,"rx":710}}}
{"t":1678419627185,"d":100,"i":"en0","s":{"1":{"tx":332}}}
{"t":1678419627285,"d":99,"i":"en0","s":{"1":{"tx":1170,"rx":364}}}
{"t":1678419627484,"d":100,"i":"en0","s":{"1":{"tx":876,"rx":721}}}
{"t":1678419627685,"d":100,"i":"en0","s":{"1":{"tx":45,"rx":12453}}}
{"t":1678419627885,"d":99,"i":"en0","s":{"1":{"tx":1032,"rx":1018},"6":{"tx":113},"7":{"tx":365},"8":{"tx":960,"rx":352}}}
{"t":1678419627984,"d":99,"i":"en0","s":{"1":{"rx":58},"8":{"tx":600,"rx":528}}}
{"t":1678419628084,"d":99,"i":"en0","s":{"1":{"tx":1670},"2":{"tx":268},"6":{"tx":73},"7":{"tx":351},"8":{"tx":320,"rx":704}}}
```

This new update I'm about to push is also very close to allowing us to start coalescing some log records into short events which will save us on storage.  We had talked about doing that event aggregation only on the c2n call, but we could also do smaller interval joining for storage.  So for example, we could still grab stats at 10hz, write to disk in 300ms sessions, then respond to c2n at a longer control-defined session interval.  That's not what this code is doing yet, but I mention it here because it makes me feel a little more comfortable including the interface in every log entry.

log/sockstatlog/logger.go
willnorris: done

log/sockstatlog/logger.go
willnorris: done

log/sockstatlog/logger.go
```diff
@@ -0,0 +1,135 @@
...
+
+// delta calculates the delta stats between two SockStats snapshots.
+// b is assumed to have occurred after a.
+// Zero values are omitted from the returned map, and an empty map is returned if no bytes were transferred.
+// iface is the first network interface found that bytes were transferred on.
+func delta(a, b *sockstats.SockStats) (stats map[sockstats.Label]deltaStat, iface string) {
+	if a == nil || b == nil {
+		return nil, ""
+	}
+	stats = make(map[sockstats.Label]deltaStat, len(b.Stats))
```
mihaip, line 111: Can use `mak.Set` to avoid allocating the map if it ends up being empty.

log/sockstatlog/logger_noop.go
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+//go:build !tailscale_go || !(darwin || ios || android)
+
+// Keep build tags in sync with net/sockstats/sockstats_noop.go
```
mihaip, line 6: Rather than doing this (which requires manual upkeep and an extra file), I think it's OK to include the logger code regardless of build tags (since it will build with the standard Go too). 

We could instead have a `sockstats.IsAvailable` const that is set to true in `sockstats_tsgo.go` and false in `sockstats_noop.go`.

log/sockstatlog/logger.go
```diff
@@ -0,0 +1,135 @@
...
+	ticker time.Ticker
+
+	// last is the last set of socket stats we saw.
+	lastStats *sockstats.SockStats
+	lastTime  time.Time
+
+	logbuffer *filch.Filch
+}
+
+// deltaStat represents the bytes transmitted and received during a time period.
```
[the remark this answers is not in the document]
willnorris: done

log/sockstatlog/logger.go
```diff
@@ -0,0 +1,135 @@
...
+// event represents the socket stats on a specific interface during a time period.
+type event struct {
+	// Time is when the event started as a Unix timestamp in milliseconds.
+	Time int64 `json:"t"`
+
+	// Duration is the duration of this event in milliseconds.
+	Duration int64 `json:"d"`
+
+	// Interface is the network interface that the event occurred on.
+	Interface string `json:"i"`
```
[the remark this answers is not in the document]
willnorris: done

log/sockstatlog/logger.go
willnorris: done

log/sockstatlog/logger_noop.go
willnorris: done

log/sockstatlog/logger.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+// pollPeriod specifies how often to poll for socket stats.
+const pollPeriod = time.Second / 10
+
+// Logger logs statistics about network sockets.
+type Logger struct {
+	ticker time.Ticker
+
+	// last is the last set of socket stats we saw.
```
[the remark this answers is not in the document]
willnorris: done

ipn/ipnlocal/local.go
```diff
@@ -304,6 +308,14 @@ func NewLocalBackend(logf logger.Logf, logid string, store ipn.StateStore, diale
 		loginFlags:     loginFlags,
 	}
 
+	// for now, only log sockstats on unstable builds
+	if version.IsUnstableBuild() {
```
willnorris, line 312: given how much we've trimmed down the log messages, I'm now actually feeling a little more comfortable going ahead and turning this on for all devices.  What do you think @mihaip ?

dsnet: hahaha, wow... you went aggressive with single character names:
```
{"t":1678419626985,"d":99,"i":"en0","s":{"0":{"tx":55,"rx":710}}}
{"t":1678419627185,"d":100,"i":"en0","s":{"1":{"tx":332}}}
{"t":1678419627285,"d":99,"i":"en0","s":{"1":{"tx":1170,"rx":364}}}
{"t":1678419627484,"d":100,"i":"en0","s":{"1":{"tx":876,"rx":721}}}
{"t":1678419627685,"d":100,"i":"en0","s":{"1":{"tx":45,"rx":12453}}}
{"t":1678419627885,"d":99,"i":"en0","s":{"1":{"tx":1032,"rx":1018},"6":{"tx":113},"7":{"tx":365},"8":{"tx":960,"rx":352}}}
{"t":1678419627984,"d":99,"i":"en0","s":{"1":{"rx":58},"8":{"tx":600,"rx":528}}}
{"t":1678419628084,"d":99,"i":"en0","s":{"1":{"tx":1670},"2":{"tx":268},"6":{"tx":73},"7":{"tx":351},"8":{"tx":320,"rx":704}}}
```

What's the compression ratio of this now? This is now dominated by double-quotes, the repeated prefix for a timestamp, and the interface name, all of which are highly compressible.

willnorris: It's even more compact than that now, since @mihaip suggested I remove the "tx" and "rx".  It now looks like:

```json
{"t":1678487681796,"d":98,"s":{"1":[1064,934],"6":[113,0],"7":[365,0],"8":[960,220]}}
{"t":1678487681895,"d":100,"s":{"8":[600,440]}}
{"t":1678487681995,"d":99,"s":{"1":[1746,35],"6":[73,0],"7":[351,0],"8":[320,528]}}
{"t":1678487682095,"d":100,"s":{"2":[629,4688],"8":[0,572]}}
{"t":1678487682196,"d":100,"s":{"1":[0,88],"8":[0,220]}}
{"t":1678487682296,"d":99,"s":{"3":[3295,0],"8":[0,88]}}
```

I haven't actually tried compressing it yet, but yeah it would probably be pretty good.

creachadair: I suppose if you really want to squeeze bytes, you could turn the pairs into triples, and move the key inside, e.g.,

```
{"1":[1064,934],"6":[113,0],"7":[365,0],"8":[960,220]}
[[1,1064,934],[6,113,0],[7,365,0],[8,960,220]]
```

but that may be too obfuscated to be worthwhile.

log/sockstatlog/logger.go
```diff
@@ -0,0 +1,131 @@
...
+// b is assumed to have occurred after a.
+// Zero values are omitted from the returned map, and an empty map is returned if no bytes were transferred.
+// iface is the first network interface found that bytes were transferred on.
+func delta(a, b *sockstats.SockStats) (stats map[sockstats.Label]deltaStat, iface string) {
+	if a == nil || b == nil {
+		return nil, ""
+	}
+	for label, bs := range b.Stats {
+		as := a.Stats[label]
+		if as.TxBytes == bs.TxBytes && as.RxBytes == bs.RxBytes {
```
dsnet, line 113: Does this conditional somehow imply that `sockstats.SockStat..XXXBytesByInterface` are nil?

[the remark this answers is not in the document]
dsnet: Disregard.

ipn/ipnlocal/local.go
willnorris: closing the loop on the original comment here... discussed in team meeting this morning, and we'll keep it only for unstable builds right now.

log/sockstatlog/logger.go
mihaip: They're not ever `nil`, but it's an optimization to only compare those two fields, instead of doing a deeper comparison of the maps too.

log/sockstatlog/logger.go
```diff
@@ -0,0 +1,131 @@
...
+			// fast path for unchanged stats
+			continue
+		}
+		mak.Set(&stats, label, deltaStat{bs.TxBytes - as.TxBytes, bs.RxBytes - as.RxBytes})
+		if iface == "" {
+			// Figure out which interface was used.
```
mihaip, line 119: We could have sockstats also return the current active interface, to make this cheaper

net/sockstats/sockstats.go
```diff
@@ -99,3 +100,11 @@ type LinkMonitor interface {
 func SetLinkMonitor(lm LinkMonitor) {
 	setLinkMonitor(lm)
 }
+
+// IsLikelyCellularInterface identifies whether a network interface is likely
```
mihaip, line 104: `net/interfaces.go` is likely a better home for this, if it's going to be public.

Could also not expose this function at all, and instead have the `SockStats` struct have a `CurrentInterfaceIsLikelyCellular` boolean field, which would also address the comment above about not having to loop over interfaces for every change.

mihaip (commented):
Feels very close.

net/sockstats/sockstats.go
willnorris: Done.  Added `SockStats.CurrentInterfaceCellular` (mirroring name on sockStats struct, but I can use the longer name if you prefer)

mihaip (approved):
LGTM for an unstable build

mihaip:
(though tests need to be updated)

PR #7499, opened 2023-03-08
sockstats: add validation for TCP socket stats

mihaip:
We can use the `TCP_CONNECTION_INFO` `getsockopt()` on Darwin to get OS-collected tx/rx bytes for TCP sockets. Since this API is not available for UDP sockets (or on Linux/Android -- the `tcpinfo` struct is not the same), we can't rely on it for actual stats gathering.

However, we can use it to validate the stats that we collect ourselves using read/write hooks, so that we can be more confident in them. We do need additional hooks from the Go standard library (added in tailscale/go#59) to be able to collect them.

Updates tailscale/corp#9230
Updates #3363


net/sockstats/sockstats_tsgo_darwin.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"golang.org/x/sys/unix"
+)
+
+func init() {
+	tcpConnStats = darwinTcpConnStats
+}
+
+func darwinTcpConnStats(c syscall.RawConn) (tx, rx uint64) {
+	c.Control(func(fd uintptr) {
```
crawshaw, line 19: this might allocate the closure on the heap. but I guess that's ok because it's called once per connection?

crawshaw (approved):

willnorris (approved):

net/sockstats/sockstats_tsgo_darwin.go
raggi: Fair, it might. We'll be allocating in the maps during that time too, so it probably falls out in the wash, but if we wanted to eradciate it completely we could capture our own fd table earlier in the process, or add a more direct API to get at the fd from a netconn - the rules for not breaking the reactor are straightforward so I'd be fine with us exposing the raw fd's if necessary.

raggi (approved):

net/sockstats/sockstats_tsgo_darwin.go
mihaip: Since this is all for validation, I can make the entire code path opt-in, triggered by the `https://tailscale.com/cap/debug` node attr (or a more specific one)

raggi:
It looks like Android is failing, some fields have different names, but also x/unix has various fields missing I think, in particular many of the metrics added for SNMP: https://github.com/torvalds/linux/blob/6a98c9cae232800c319ed69e1063480d31430887/include/uapi/linux/tcp.h#L241

mihaip:
> It looks like Android is failing, some fields have different names, but also x/unix has various fields missing I think, in particular many of the metrics added for SNMP: https://github.com/torvalds/linux/blob/6a98c9cae232800c319ed69e1063480d31430887/include/uapi/linux/tcp.h#L241

Yeah, not sure why I thought Linux/Android had similar fields. I've removed the Android implementation for now.

net/sockstats/sockstats_tsgo_darwin.go
mihaip: Actually, now that this is Darwin and TCP-only I think it's infrequent enough that it's not worth the extra knob.

PR #7503, opened 2023-03-08
ipn/ipnlocal: add support for funnel in tsnet

maisem:
Previously the part that handled Funnel connections was not
aware of any listeners that tsnet.Servers might have had open
so it would check against the ServeConfig and fail.

Adding a ServeConfig for a TCP proxy was also not suitable in this
scenario as that would mean creating two different listeners and have
one forward to the other, which really meant that you could not have
funnel and tailnet-only listeners on the same port.

This also introduces the ipn.FunnelConn as a way for users to identify
whether the call is coming over funnel or not. Currently it only holds
the underlying conn and the target as presented in the "Tailscale-Ingress-Target"
header.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -150,6 +150,22 @@ type LocalBackend struct {
 	shutdownCalled        bool // if Shutdown has been called
 	debugSink             *capture.Sink
 
+	// getTCPHandlerForFunnelFlow returns a handler for an incoming TCP flow for
+	// the provided srcAddr and dstPort if one exists.
+	//
+	// srcAddr is the source address of the flow, not the address of the Funnel
+	// node relaying the flow.
+	// dstPort is the destination port of the flow.
+	//
+	// It returns nil if there is no known handler for this flow.
+	//
+	// This is specifically used to handle TCP flows for Funnel connections to tsnet
+	// servers.
+	//
+	// It is set once during initialization, and can be nil if SetTCPHandlerForFunnelFlow
+	// is never called.
+	getTCPHandlerForFunnelFlow func(srcAddr netip.AddrPort, dstPort uint16) (handler func(net.Conn))
+
 	// lastProfileID tracks the last profile we've seen from the ProfileManager.
 	// It's used to detect when the user has changed their profile.
 	lastProfileID ipn.ProfileID
@@ -3115,6 +3131,12 @@ func dnsConfigForNetmap(nm *netmap.NetworkMap, prefs ipn.PrefsView, logf logger.
 	}
 
 	return dcfg
+}
+
+// SetTCPHandlerForFunnelFlow sets the TCP handler for Funnel flows.
+// It should only be called before the LocalBackend is used.
+func (b *LocalBackend) SetTCPHandlerForFunnelFlow(h func(src netip.AddrPort, dstPort uint16) (handler func(net.Conn))) {
+	b.getTCPHandlerForFunnelFlow = h
 }
 
 // SetVarRoot sets the root directory of Tailscale's writable
--- a/ipn/serve.go
+++ b/ipn/serve.go
@@ -3,6 +3,11 @@
 
 package ipn
 
+import (
+	"net"
+	"net/netip"
+)
+
 // ServeConfigKey returns a StateKey that stores the
 // JSON-encoded ServeConfig for a config profile.
 func ServeConfigKey(profileID ProfileID) StateKey {
@@ -28,6 +33,26 @@ type ServeConfig struct {
 // HostPort is an SNI name and port number, joined by a colon.
 // There is no implicit port 443. It must contain a colon.
 type HostPort string
+
+// A FunnelConn wraps a net.Conn that is coming over a
+// Funnel connection. It can be used to determine further
+// information about the connection, like the source address
+// and the target SNI name.
+type FunnelConn struct {
+	// Conn is the underlying connection.
+	net.Conn
+
+	// Target is what was presented in the "Tailscale-Ingress-Target"
+	// HTTP header.
+	Target HostPort
+
+	// Src is the source address of the connection.
+	// This is the address of the client that initiated the
+	// connection, not the address of the Tailscale Funnel
+	// node which is relaying the connection. That address
+	// can be found in Conn.RemoteAddr.
+	Src netip.AddrPort
+}
 
 // WebServerConfig describes a web server's configuration.
 type WebServerConfig struct {
--- a/tsnet/example/tsnet-funnel/tsnet-funnel.go
+++ b/tsnet/example/tsnet-funnel/tsnet-funnel.go
@@ -1,0 +1,131 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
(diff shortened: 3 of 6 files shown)


ipn/ipnlocal/local.go
```diff
@@ -219,6 +219,14 @@ type LocalBackend struct {
 	serveListeners     map[netip.AddrPort]*serveListener // addrPort => serveListener
 	serveProxyHandlers sync.Map                          // string (HTTPHandler.Proxy) => *httputil.ReverseProxy
 
+	// getTCPHandlerForFunnelFlow conditionally handles an incoming TCP flow for
```
crawshaw, line 222: looks like this is protected by the mutex on a paragraph of fields above. does that mutex protect all these other fields too (like `serveProxyHandlers`)? that would be a little surprising. assuming it doesn't, I think this field should be mashed into the very large paragraph with mu. (might be time for a new private type or something for those grouped fields just for readability)

ipn/serve.go
```diff
@@ -29,6 +31,16 @@ type ServeConfig struct {
 // There is no implicit port 443. It must contain a colon.
 type HostPort string
 
+// FunnelConn is a net.Conn that also has an SNI name.
```
crawshaw, line 34: I feel like you're underselling it?

tsnet/example/tsnet-funnel/tsnet-funnel.go
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+// The tsnet-funnel server demonstrates how to use tsnet with funnel.
```
crawshaw, line 4: There is some ideal world where you write a tsnet test that starts a testcontrol and then builds+executes this example against it. I added testcontrol-based tests of tsnet. Might be worth a try.

crawshaw (approved):

ipn/ipnlocal/local.go
```diff
@@ -3117,6 +3125,12 @@ func dnsConfigForNetmap(nm *netmap.NetworkMap, prefs ipn.PrefsView, logf logger.
 	return dcfg
 }
 
+// SetTCPHandlerForFunnelFlow sets the TCP handler for funnel flows.
+// It should only be called before the LocalBackend is used.
+func (b *LocalBackend) SetTCPHandlerForFunnelFlow(h func(netip.AddrPort, uint16) func(net.Conn)) {
```
bradfitz, line 3130: It's non-obvious enough that I'd name src and dstPort here.

ipn/ipnlocal/local.go
```diff
@@ -219,6 +219,14 @@ type LocalBackend struct {
 	serveListeners     map[netip.AddrPort]*serveListener // addrPort => serveListener
 	serveProxyHandlers sync.Map                          // string (HTTPHandler.Proxy) => *httputil.ReverseProxy
 
+	// getTCPHandlerForFunnelFlow conditionally handles an incoming TCP flow for
+	// the provided src and dport. It returns a handler function to handle the
+	// flow, and returns nil if the flow should be handled by serve listeners.
```
bradfitz, line 224: I don't quite follow this bit about nil.

tsnet/example/tsnet-funnel/tsnet-funnel.go
```diff
@@ -0,0 +1,103 @@
...
+	srvConfig := &ipn.ServeConfig{
+		AllowFunnel: map[ipn.HostPort]bool{
+			hp: true,
+		},
+	}
+	lc, err := s.LocalClient()
+	if err != nil {
+		return err
+	}
+	return lc.SetServeConfig(ctx, srvConfig)
```
shayne, line 46: What do you think about rolling this up into `tsnet.Server` functions like `FunnelOn(hp ipn.HostPort) error` and `FunnelOff(hp ipnHostPort) error`.

tsnet/example/tsnet-funnel/tsnet-funnel.go
maisem: I couldn't figure out a nice way to do this for the funnel bit, I'll add it as a follow up.

tsnet/example/tsnet-funnel/tsnet-funnel.go
maisem: yeah, something like that would be nice! Can do it in a follow up change so that its isolated and we can figure out the appropriate API.

tsnet/example/tsnet-funnel/tsnet-funnel.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	log.Fatal(httpServer.Serve(ln))
+}
+
+type funnelKey struct{}
+
+func isFunnel(ctx context.Context) bool {
+	v, _ := ctx.Value(funnelKey{}).(bool)
+	return v
+}
```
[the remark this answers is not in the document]
maisem: Added docs

ipn/ipnlocal/local.go
maisem: Reworded, ptal

ipn/ipnlocal/local.go
maisem: done.

ipn/serve.go
maisem: Reworded

ipn/ipnlocal/local.go
maisem: It didn't need a mutex, I think I just had it a remnant of a previous iteration of the logic. Moved and updated docs

ipn/ipnlocal/local.go
```diff
@@ -150,6 +150,17 @@ type LocalBackend struct {
 	shutdownCalled        bool // if Shutdown has been called
 	debugSink             *capture.Sink
 
+	// getTCPHandlerForFunnelFlow returns a handler for an incoming TCP flow for
+	// the provided srcAddr and dstPort if one exists.
+	// It returns nil if there is no known handler for this flow.
+	//
+	// This is specifically used to handle TCP flows for funnel connections to tsnet
```
bradfitz, line 157: Funnel capitalized

ipn/ipnlocal/local.go
```diff
@@ -150,6 +150,17 @@ type LocalBackend struct {
 	shutdownCalled        bool // if Shutdown has been called
 	debugSink             *capture.Sink
 
+	// getTCPHandlerForFunnelFlow returns a handler for an incoming TCP flow for
+	// the provided srcAddr and dstPort if one exists.
```
bradfitz, line 154: Clarify that srcAddr is from the public internet?

ipn/ipnlocal/local.go
```diff
@@ -3117,6 +3128,12 @@ func dnsConfigForNetmap(nm *netmap.NetworkMap, prefs ipn.PrefsView, logf logger.
 	return dcfg
 }
 
+// SetTCPHandlerForFunnelFlow sets the TCP handler for funnel flows.
```
bradfitz, line 3131: Funnel

bradfitz (approved):

tsnet/example/tsnet-funnel/tsnet-funnel.go
maisem: https://github.com/tailscale/tailscale/pull/7519

PR #7504, opened 2023-03-08
net/httpconnect: HTTP CONNECT implementation

crawshaw:
Turns out we had two of these.
Factor them out, and add auth support.


crawshaw:
will write tests if the idea is reasonable.

i'm factoring this out because i want to include it in tsnet

PR #251, opened 2020-04-04
tsweb: don't flush, treat no-op Handler as 200, like Go

bradfitz:
This change is [Reviewable](https://reviewable.io/reviews/tailscale/tailscale/251)

PR #7508, opened 2023-03-08
logtail: remove unncessary response read

mihaip:
Effectively reverts #249, since the server side was fixed (with #251?) to send a 200 OK/content-length 0 response.


mihaip:
Empirically verified that responses that we get look like:
```
HTTP/1.1 200 OK
Content-Length: 0
Date: Wed, 08 Mar 2023 23:09:23 GMT
```

danderson (approved):

danderson:
3 years is "a few days", right? :P

dsnet (approved):

dsnet:
> 3 years is "a few days", right? :P

Ya never know...

mihaip:
> > 3 years is "a few days", right? :P
> 
> Ya never know...

Since it's March 1103rd, 2020, I guess we're still the bounds of the original comment.

PR #7509, opened 2023-03-09
control/controlclient: improve handling of concurrent lite map requests

maisem:
This reverts commit 6eca47b16c37dac2984de23d5af6ecb15c54bb97 and fixes forward.

Previously the first ever streaming MapRequest that a client sent would also set ReadOnly to true as it didn't have any endpoints and expected/relied on the map poll to restart as soon as it got endpoints. However with 48f6c1eba4e29fdac9b0f807ee50dcefa387471d, we would no longer restart MapRequests as frequently as we used to, so control would only ever get the first streaming MapRequest which had ReadOnly=true.

Control would treat this as an uninteresting request and would not send it any further netmaps, while the client would happily stay in the map poll forever while litemap updates happened in parallel.

This makes it so that we never set `ReadOnly=true` when we are doing a streaming MapRequest. This is no longer necessary either as most endpoint discovery happens over disco anyway.


twitchyliquid64 (approved):

control/controlclient/auto.go
```diff
@@ -59,15 +59,17 @@ type Auto struct {
...
-	inSendStatus    int        // number of sendStatus calls currently in progress
-	state           State
+	paused               bool // whether we should stop making HTTP requests
+	unpauseWaiters       []chan struct{}
+	loggedIn             bool       // true if currently logged in
+	loginGoal            *LoginGoal // non-nil if some login activity is desired
+	synced               bool       // true if our netmap is up-to-date
+	inPollNetMap         bool       // true if currently running a PollNetMap
+	inLiteMapUpdate      bool       // true if a lite (non-streaming) map request is outstanding
+	liteMapUpdateCancel  func()     // cancels a lite map update
```
bradfitz, line 69: can it be nil? why not an context.CancelFunc? If it's a CancelFunc, then I know it's harmless to call multiple times.

control/controlclient/auto.go
```diff
@@ -168,39 +170,69 @@ func (c *Auto) Start() {
...
+	// and start a new stream (which starts by sending a new map request)
+	if !c.inPollNetMap || !c.loggedIn {
 		c.mu.Unlock()
 		c.cancelMapSafely()
 		return
 	}
 
+	// If we are already in process of doing a LiteMapUpdate, cancel it and
+	// try a new one. If this is the 10th time we have done this
+	// cancelation, tear down everything and start again
```
bradfitz, line 183: period

control/controlclient/auto.go
```diff
@@ -168,39 +170,69 @@ func (c *Auto) Start() {
...
+	if c.inLiteMapUpdate {
+		// Always cancel the in-flight lite map update, regardless of
+		// whether we cancel the streaming map request or not.
+		c.liteMapUpdateCancel()
+		c.inLiteMapUpdate = false
+
+		if c.liteMapUpdateCancels > 10 {
```
bradfitz, line 190: 10th or 11th time? (per Line 182 docs)

control/controlclient/auto.go
```diff
@@ -168,39 +170,69 @@ func (c *Auto) Start() {
...
+	c.liteMapUpdateCancel = cancel
 	go func() {
 		defer cancel()
 		t0 := time.Now()
 		err := c.direct.SendLiteMapUpdate(ctx)
 		d := time.Since(t0).Round(time.Millisecond)
+
 		c.mu.Lock()
 		c.inLiteMapUpdate = false
+		c.liteMapUpdateCancel = nil
```
bradfitz, line 216: ah, yes, it can be nil

maisem: Updated docs

control/controlclient/auto.go
maisem: 10th, fixed check and made a const.

control/controlclient/auto.go
maisem: done

control/controlclient/auto.go
maisem: updated docs.

bradfitz (approved):

PR #7517, opened 2023-03-09
sockstats: refactor validation to be opt-in

mihaip:
Followup to #7499 to make validation a separate function (`GetWithValidation` vs. `Get`). This way callers that don't need it don't pay the cost of a syscall per active TCP socket.

Also clears the conn on close, so that we don't double-count the stats.

Also more consistently uses Go doc comments for the exported API of the sockstats package.

Updates tailscale/corp#9230
Updates #3363

### DIFF
--- a/net/sockstats/sockstats.go
+++ b/net/sockstats/sockstats.go
@@ -14,9 +14,23 @@ import (
 	"tailscale.com/net/interfaces"
 )
 
+// SockStats contains statistics for sockets instrumented with the
+// WithSockStats() function, along with the interfaces that we have
+// per-interface statistics for.
 type SockStats struct {
 	Stats      map[Label]SockStat
 	Interfaces []string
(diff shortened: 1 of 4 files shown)


net/sockstats/sockstats.go
```diff
@@ -41,32 +55,47 @@ const (
...
+}
+
 // LinkMonitor is the interface for the parts of wgengine/mointor's Mon that we
 // need, to avoid the dependency.
 type LinkMonitor interface {
 	InterfaceState() *interfaces.State
 	RegisterChangeCallback(interfaces.ChangeFunc) (unregister func())
 }
 
+// SetLinkMonitor configurs the sockstats package to monitor the active
```
willnorris, line 97: configures

willnorris (approved):

PR #7525, opened 2023-03-10
tailcfg: bump capver for 1.38

maisem:


DentonGentry (approved):

andrew-d (approved):

tailcfg/tailcfg.go
```diff
@@ -94,7 +94,8 @@ type CapabilityVersion int
 //   - 55: 2023-01-23: start of c2n GET+POST /update handler
 //   - 56: 2023-01-24: Client understands CapabilityDebugTSDNSResolution
 //   - 57: 2023-01-25: Client understands CapabilityBindToInterfaceByRoute
-const CurrentCapabilityVersion CapabilityVersion = 57
+//   - 58: 2023-03-10: Client retries lite map updates before restarting map poll.
```
bradfitz, line 97: This isn't really a typical capver bump description. The client can't do anything new. Its behavior changed, sure, but will the server do anything differently as a result of this?

maisem: I added some tracking metrics, and wanted to correlate them with the change. I can change the description to only be `nothing, bump for 1.38`

andrew-d: If something goes wrong with these changes, I like having this as a distinct capver so we can have the server ignore the `ReadOnly` flag or force-restart map polls or whatever until we fix the issue.

bradfitz: sure, s'fine

bradfitz: But remember we never bump capver on branches. So if we have to fix the issue in the client, we can't distinguish 1.38.0 from 1.38.1 using capver alone.

PR #7526, opened 2023-03-10
ipn: add c2n endpoint for sockstats logs

willnorris:
Add c2n endpoint for collecting sockstats logs from a device.

### DIFF
--- a/ipn/ipnlocal/c2n.go
+++ b/ipn/ipnlocal/c2n.go
@@ -83,6 +83,9 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		writeJSON(res)
+	case "/sockstats":
+		w.Header().Set("Content-Type", "text/plain")
+		b.sockstatLogger.WriteLogs(w)
 	default:
 		http.Error(w, "unknown c2n path", http.StatusBadRequest)
 	}
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -310,7 +310,7 @@ func NewLocalBackend(logf logger.Logf, logid string, store ipn.StateStore, diale
 
 	// for now, only log sockstats on unstable builds
 	if version.IsUnstableBuild() {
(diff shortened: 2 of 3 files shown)


willnorris:
This might still be more granular than we were imagining reporting to control, I'm not sure.  This also has the side effect (because of how filch works) that logs are removed from disk when they are read.  So if we were planning to allow local browsing of the on-device logs also, then we need to adjust this approach (likely to not use filch as-is).

mihaip (commented):
High-level thoughts, we should probably talk through them at 1pm:
- Since this is an on-demand/debugging endpoint, it seems preferable to do any kind of coalescing/analysis on the control side, where we have more flexibility to change it without having to deploy new clients/get them updated.
- Having this be a lossy process seems not great, as far as a seemingly-read endpoint having side effects. Perhaps one alternate design to consider is that the c2n API does a flush to the logcatcher instead (in a new bucket), and then visualization tooling can read from there. That way we can repeatedly query the data without having to build a bespoke storage system for it. I assume we can configure the bucket with a different (shorter) retention period since it's more sensitive and higher-volume. @dsnet probably has opinions here.

willnorris:
okay, simplified this to just write all current sockstat logs in the c2n response with no event coalescing.  We'll look at adjusting this to use logtail in the future, putting logs into a separate bucket.

ipn/ipnlocal/c2n.go
```diff
@@ -83,6 +83,8 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		writeJSON(res)
+	case "/sockstats":
+		b.sockstatLogger.WriteLogs(w)
```
mihaip, line 87: Should probably do a `w.Header().Set("Content-Type", "text/plain")` to ensure the browser doesn't try to interpret the output as HTML.

mihaip (approved):

ipn/ipnlocal/c2n.go
willnorris: done

PR #7528, opened 2023-03-10
tsnet: always attempt to create storage directory

crawshaw:
This lets users start a server with:

	Dir: "./my-demo"

easily.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -415,9 +415,9 @@ func (s *Server) start() (reterr error) {
 		if err != nil {
 			return err
 		}
-		if err := os.MkdirAll(s.rootPath, 0700); err != nil {
-			return err
-		}
+	}
+	if err := os.MkdirAll(s.rootPath, 0700); err != nil {
+		return err
 	}
 	if fi, err := os.Stat(s.rootPath); err != nil {
 		return err


maisem (approved):

bradfitz (commented):
Seems kinda risky. I'd rather caller do that if they assume that risk.

crawshaw:
The problem is our examples right now are loaded up with all sorts of config gunk for starting a tsnet.Server. I'm looking for layers of gunk I can remove from the examples.

I tried not mentioning Dir/Store at all, but it goes terribly wrong and I picked up an old store saved in my user dir from a tsnet server months ago, and was undebuggable. So one of them should be set. But we only allow store.Mem for ephemeral nodes, which has naming problems. So where's the Dir? Now it's the user's job to make it. (And I tried to use cwd, but we spew a stack of log files into the state directory.) So what's the easy thing to do?

crawshaw:
See this example here: https://github.com/tailscale/tailscale/pull/7529/files#diff-154a80f096dcf09396422e2089bcd9e3de4be9eb792b051c9f2763de2f6bfe33

That's already a lot of setup for the simple act of dropping a server on funnel. I'd like to do even less. Ideas?

uhthomas:
Would this have a negative affect on rootless containers?

PR #7529, opened 2023-03-10
tsnet: add ListenFunnel

crawshaw:
This lets a tsnet binary share a server out over Tailscale Funnel.

### DIFF
--- a/tsnet/example/tsnet-funnel/tsnet-funnel.go
+++ b/tsnet/example/tsnet-funnel/tsnet-funnel.go
@@ -2,130 +2,40 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 // The tsnet-funnel server demonstrates how to use tsnet with Funnel.
+//
+// To use it, generate an auth key from the Tailscale admin panel and
+// run the demo with the key:
+//
+//	TS_AUTHKEY=<yourkey> go run tsnet-funnel.go
 package main
 
 import (
-	"context"
-	"crypto/tls"
-	"errors"
 	"flag"
 	"fmt"
 	"log"
-	"net"
 	"net/http"
-	"net/netip"
 
-	"tailscale.com/ipn"
 	"tailscale.com/tsnet"
+	"tailscale.com/types/logger"
 )
-
-var (
-	addr = flag.String("addr", ":443", "address to listen on")
-)
-
-func enableFunnel(ctx context.Context, s *tsnet.Server) error {
-	st, err := s.Up(ctx)
-	if err != nil {
-		return err
-	}
-	if len(st.CertDomains) == 0 {
-		return errors.New("tsnet: you must enable HTTPS in the admin panel to proceed")
-	}
-	domain := st.CertDomains[0]
-
-	hp := ipn.HostPort(net.JoinHostPort(domain, "443"))
-
-	srvConfig := &ipn.ServeConfig{
-		AllowFunnel: map[ipn.HostPort]bool{
-			hp: true,
-		},
-	}
-	lc, err := s.LocalClient()
-	if err != nil {
-		return err
-	}
-	return lc.SetServeConfig(ctx, srvConfig)
-}
 
 func main() {
 	flag.Parse()
-	s := new(tsnet.Server)
-	defer s.Close()
-	ctx := context.Background()
-	if err := enableFunnel(ctx, s); err != nil {
-		log.Fatal(err)
+	s := &tsnet.Server{
+		Dir:      "./funnel-demo-config.state",
+		Logf:     logger.Discard,
+		Hostname: "fun",
 	}
+	defer s.Close()
 
-	ln, err := s.Listen("tcp", *addr)
+	ln, err := s.ListenFunnel("tcp", ":443")
 	if err != nil {
 		log.Fatal(err)
 	}
 	defer ln.Close()
 
-	lc, err := s.LocalClient()
-	if err != nil {
-		log.Fatal(err)
-	}
-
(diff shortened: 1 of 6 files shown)


tsnet/example/tsnet-funnel/tsnet-funnel.go
```diff
@@ -2,130 +2,40 @@
...
-
 func main() {
 	flag.Parse()
-	s := new(tsnet.Server)
-	defer s.Close()
-	ctx := context.Background()
-	if err := enableFunnel(ctx, s); err != nil {
-		log.Fatal(err)
+	s := &tsnet.Server{
+		Dir:      "./funnel-demo-config.state",
```
bradfitz, line 25: That looks like a file, not a dir

crawshaw:
Shayne wants to make ListenFunnel do both funnel and tailnet listening, which this PR doesn't do. Investigating.

tsnet/tsnet.go
```diff
@@ -760,6 +761,74 @@ func (s *Server) APIClient() (*tailscale.Client, error) {
 // Listen announces only on the Tailscale network.
 // It will start the server if it has not been started yet.
 func (s *Server) Listen(network, addr string) (net.Listener, error) {
+	return s.listen(network, addr, false)
+}
+
+// ListenFunnel announces on the public internet using Tailscale Funnel.
+//
+// Currently, Funnel only supports TCP on ports 443, 8443, and 10000.
```
bradfitz, line 769: Temporal references should include a date in parens

tsnet/tsnet.go
```diff
@@ -760,6 +761,74 @@ func (s *Server) APIClient() (*tailscale.Client, error) {
...
+// The support host name is limited to that configured for the tsnet.Server.
```
bradfitz, line 770: Supported

tsnet/tsnet.go
```diff
@@ -760,6 +761,74 @@ func (s *Server) APIClient() (*tailscale.Client, error) {
...
+//
+// and the only other supported addrs currently are ":8443" and ":10000".
+//
+// It will start the server if it has not been started yet.
+func (s *Server) ListenFunnel(network string, addr string) (net.Listener, error) {
+	if network != "tcp" {
+		return nil, fmt.Errorf("ListenFunnel(%q, %q): only tcp is supported", network, addr)
+	}
+	switch addr {
+	case ":443", ":8443", ":10000":
```
bradfitz, line 783: Don't hard code this. Boxes us in later

tsnet/tsnet.go
```diff
@@ -760,6 +761,74 @@ func (s *Server) APIClient() (*tailscale.Client, error) {
...
+		return nil, errors.New("tsnet: you must enable HTTPS in the admin panel to proceed")
+	}
+
+	lc, err := s.LocalClient() // do local client first before listening.
+	if err != nil {
+		return nil, err
+	}
+
+	// May not have funnel enabled. Enable it.
+	srvConfig, err := lc.GetServeConfig(ctx)
```
maisem, line 802: Add a check for the funnel nodeAttr too

crawshaw:
I still have to address all the comments, but I believe ListenFunnel now listens on both by default.

tsnet/tsnet.go
```diff
@@ -760,6 +760,95 @@ func (s *Server) APIClient() (*tailscale.Client, error) {
 // Listen announces only on the Tailscale network.
 // It will start the server if it has not been started yet.
 func (s *Server) Listen(network, addr string) (net.Listener, error) {
+	return s.listen(network, addr, listenOnTailnet)
+}
+
+// FunnelOption is an option passed to ListenFunnel to configure the listener.
+type FunnelOption string
```
maisem, line 767: I'm concerned that `string` is going to corner us in the future.

tsnet/tsnet.go
```diff
@@ -760,6 +761,96 @@ func (s *Server) APIClient() (*tailscale.Client, error) {
...
+	}
+	domain := st.CertDomains[0]
+	hp := ipn.HostPort(domain + addr) // valid only because of the strong restrictions on addr above
+	if !srvConfig.AllowFunnel[hp] {
+		mak.Set(&srvConfig.AllowFunnel, hp, true)
+		srvConfig.AllowFunnel[hp] = true
+		if err := lc.SetServeConfig(ctx, srvConfig); err != nil {
+			return nil, err
+		}
+	}
```
shayne, line 827: We should clear the `serveConfig` when a tsnet app starts. However, we're persisting the `serveConfig`, so if the user turns on Funnel with `ListenFunnel` and then changes it to `Listen`, well, it is still on. There are other oddities too.

I would suggest just clobbering it and not looking up the existing config, but a complex tsnet app could be running multiple listeners.

crawshaw: OK, we could clobber it in Up.

tsnet/tsnet.go
crawshaw: I don't really see how, but what alternative would you prefer?

maisem: My concern only concern is that we won't be able to pass parameters.

We can do the usual modifier function option thing.

Like

`type ListenOption func(*listenConfig)`

crawshaw: I went with the most generic of the option type variants, figuring it gives us the flexibility to convert later.

maisem: Thanks!

maisem (approved):

tsnet/tsnet.go
maisem: I'm working on a separate change for this.

PR #7532, opened 2023-03-11
ipn: add Funnel port check from nodeAttr

maisem:


tailcfg/tailcfg.go
```diff
@@ -1058,6 +1058,11 @@ type PortRange struct {
 	Last  uint16
 }
 
+// Contains returns whether port is in pr.
```
bradfitz, line 1061: Reports

bradfitz (approved):

tailcfg/tailcfg.go
maisem: not applicable here, not a bool return.

bradfitz: Yeah it is

maisem: oops, i thought this was in a different place. my bad

crawshaw (approved):

tailcfg/tailcfg.go
maisem: fixed

PR #7533, opened 2023-03-11
tsnet: add CertDomains helper

maisem:


shayne (approved):

tsnet/example/tsnet-funnel/tsnet-funnel.go
```diff
@@ -34,6 +34,8 @@ func main() {
 	}
 	defer ln.Close()
 
+	fmt.Printf("Listening on https://%v\n", s.CertDomains()[0])
```
bradfitz, line 37: Won't this crash in a common case?

maisem: No, ListenFunnel does all the necessary checks

PR #7534, opened 2023-03-11
tsnet/tsnet: clear ipn.ServeConfig on Up for tsnet apps

shayne:
We persist the ServeConfig, even for tsnet apps. It's quite possible for
the ServeConfig to be out of step with the code. Example: If you run
`ListenFunnel` then later turn it off, the ServeConfig will still show
it enabled, the admin console will show it enabled, but the packet
handler will reject the packets.

Workaround by clearing the ServeConfig in `tsnet.Up`


maisem (approved):

bradfitz:
In future just one tsnet before colon, the package name.

Issue #7537, opened 2023-03-12
tsnet (and everything else) doesn't call dnsfallback.SetLogger()


eric:
### What is the issue?

Continuation of #5743

It appears that `tsnet` doesn't call `dnsfallback.SetLogger()`, which is required to handle those messages.

Found a few more messages:

```
bootstrapDNS("derp4c.tailscale.com", "134.122.77.138") for "controlplane.tailscale.com" error: Get "https://derp4c.tailscale.com/bootstrap-dns?q=controlplane.tailscale.com": dial tcp 134.122.77.138:443: connect: network is unreachable
```

```
trying bootstrapDNS("derp9c.tailscale.com", "2001:19f0:6401:fe7:5400:3ff:fe8d:6d9c") for "controlplane.tailscale.com" ...
```

```
[unexpected] netns: DefaultRouteInterfaceIndex: no gateway index found
```

PR #7539, opened 2023-03-12
various: pass logger.Logf through to more places

andrew-d:
Updates #7537

Change-Id: Id89acab70ea678c8c7ff0f44792d54c7223337c6

### DIFF
--- a/control/controlhttp/client.go
+++ b/control/controlhttp/client.go
@@ -388,12 +388,14 @@ func (a *Dialer) tryURLUpgrade(ctx context.Context, u *url.URL, addr netip.Addr,
 		dns = &dnscache.Resolver{
 			SingleHostStaticResult: []netip.Addr{addr},
 			SingleHost:             u.Hostname(),
+			Logf:                   a.Logf, // not a.logf method; we want to propagate nil-ness
 		}
 	} else {
 		dns = &dnscache.Resolver{
 			Forward:          dnscache.Get().Forward,
 			LookupIPFallback: dnsfallback.Lookup,
 			UseLastGood:      true,
+			Logf:             a.Logf, // not a.logf method; we want to propagate nil-ness
 		}
 	}
 
--- a/net/dnsfallback/dnsfallback.go
+++ b/net/dnsfallback/dnsfallback.go
@@ -250,10 +250,13 @@ func SetCachePath(path string) {
 // logfunc stores the logging function to use for this package.
 var logfunc syncs.AtomicValue[logger.Logf]
 
-// SetLogger sets the logging function that this package will use. The default
-// logger if this function is not called is 'log.Printf'.
-func SetLogger(log logger.Logf) {
-	logfunc.Store(log)
+// SetLogger sets the logging function that this package will use, and returns
+// the old value (which may be nil).
+//
+// If this function is never called, or if this function is called with a nil
+// value, 'log.Printf' will be used to print logs.
+func SetLogger(log logger.Logf) (old logger.Logf) {
+	return logfunc.Swap(log)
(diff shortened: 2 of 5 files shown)


control/controlhttp/client.go
```diff
@@ -382,18 +383,27 @@ func (a *Dialer) dialURL(ctx context.Context, u *url.URL, addr netip.Addr) (*Cli
 func (a *Dialer) tryURLUpgrade(ctx context.Context, u *url.URL, addr netip.Addr, init []byte) (net.Conn, error) {
 	var dns *dnscache.Resolver
 
+	// Maintain prior behaviour where passing no logger to
```
bradfitz, line 386: logf method instead?

tsnet/tsnet.go
```diff
@@ -619,6 +620,25 @@ func (s *Server) logf(format string, a ...interface{}) {
 	log.Printf(format, a...)
 }
 
+// ReplaceGlobalLoggers will replace any Tailscale-specific package-global
+// loggers with this Server's logger. It returns a function that, when called,
+// will undo any changes made.
+//
+// Note that calling this function from multiple Servers will result in the
+// last call taking all logs; logs are not duplicated.
+func (s *Server) ReplaceGlobalLoggers() func() {
```
bradfitz, line 629: Name result?

bradfitz (approved):

control/controlhttp/client.go
maisem, line 386: Maintain prior behavior from when?

net/dnsfallback/dnsfallback.go
```diff
@@ -250,10 +250,11 @@ func SetCachePath(path string) {
 // logfunc stores the logging function to use for this package.
 var logfunc syncs.AtomicValue[logger.Logf]
 
-// SetLogger sets the logging function that this package will use. The default
-// logger if this function is not called is 'log.Printf'.
-func SetLogger(log logger.Logf) {
-	logfunc.Store(log)
+// SetLogger sets the logging function that this package will use, and returns
+// the old value (which may be nil). The default logger if this function is not
```
maisem, line 254: The default value of the logger is log.Printf

control/controlhttp/client.go
andrew-d: I intentionally didn't use that; the `a.logf` method will drop logs if no `a.Logf` option is specified, whereas the previous behaviour of `dnscache.Resolver` always logs with `log.Printf` if there's no logger specified.

maisem (approved):

control/controlhttp/client.go
bradfitz: logfOrDefault method then? I just don't want to see a big comment and var decl/conditional in the middle of unrelated code

net/dnsfallback/dnsfallback.go
andrew-d: Yep, but the default value of the `logfunc` is nil; the top-level `logf` function will fall back to `log.Printf` if so.

maisem: The value defaults to log.Printf if not explicitly set?

control/controlhttp/client.go
andrew-d: I actually just changed this to use `a.Logf` directly in the argument; this propagates nil-ness and retains the default "fall back to log.Printf" behaviour that `dnscache.Resolver` has. PTAL?

andrew-d: Changed the approach here to pass along `a.Logf` directly

maisem (approved):

PR #7542, opened 2023-03-13
tsnet: avoid deadlock on close

crawshaw:
tsnet.Server.Close was calling listener.Close with the server mutex held, but the listener close method tries to grab that mutex, resulting in a deadlock.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -349,12 +349,17 @@ func (s *Server) Close() error {
 		s.loopbackListener.Close()
 	}
 
+	var lns []*listener
+
 	s.mu.Lock()
-	defer s.mu.Unlock()
 	for _, ln := range s.listeners {
+		lns = append(lns, ln)
+	}
+	s.mu.Unlock()
+
+	for _, ln := range lns {
 		ln.Close()
 	}
-	s.listeners = nil
 
 	wg.Wait()
 	return nil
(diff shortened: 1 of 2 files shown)


bradfitz (approved):
Or add a closeLocked method saying s.mu must be held? And then call that from Close and avoid the slice and extra locking and unlocking?

tsnet/tsnet.go
```diff
@@ -349,12 +349,17 @@ func (s *Server) Close() error {
...
-	defer s.mu.Unlock()
 	for _, ln := range s.listeners {
+		lns = append(lns, ln)
+	}
+	s.mu.Unlock()
+
+	for _, ln := range lns {
 		ln.Close()
 	}
-	s.listeners = nil
```
maisem, line 357: Don't you still need this?

maisem: You could do

```
s.mu.Lock()
lns := s.listeners
s.listeners = nil
s.mu.Unlock()
```

maisem (approved):

tsnet/tsnet.go
maisem: Or brads suggestion

maisem: sent https://github.com/tailscale/tailscale/pull/7559

Issue #7365, opened 2023-02-24
net/netcheck: determine region from cloud VM's metadata server?


bradfitz:
For users running tailscaled on cloud VMs with metadata servers, we could avoid some netcheck probes. e.g. if your VM's metadata server says you're in oregon-2, then Bangalore or Singapore isn't worth even trying.

/cc @DentonGentry @andrew-d

maisem:
We should be careful to not strictly rely on location when doing the netcheck though as that obvious answers don't work in all countries.

andrew-d:
> We should be careful to not strictly rely on location when doing the netcheck though as that obvious answers don't work in all countries.

My current theory is that we will publish an approximate lat/lon pair for each DERP region, and have a mapping of cloud provider region to approximate location baked into `tailscaled` (`us-west-2` = 45.841111, -119.291667, etc.). We can then sort regions by their "closeness", at the least.

Still TBD on exactly what we do, but... if we want to skip netcheck probes, we can pick from the sorted-by-distance list until we have at least 5 DERP regions, and after that skip anything that's more than 1/3rd the circumference of the earth away (i.e. 40,075km / 3 ≈ 13,350km). This means that for nodes near a lot of DERP regions, we only probe stuff that's likely to matter, and for nodes farther away from DERP regions, we'll at least probe 5 of them, regardless of how far away things are. I want to do some math to verify that the 1/3rd heuristic above won't break cases that are far from _everything_, or be too permissive as to not matter, so that heuristic is subject to change 😃

PR #7544, opened 2023-03-13
tailcfg: add optional Location field to DERPRegion

andrew-d:
This allows future optimizations for finding the closest DERP region without needing to send a netcheck probe to everywhere in the world.

Updates #7365

Change-Id: I9c0a235283d6a46959c68f87d4aa2fc1cb86ba19

### DIFF
--- a/tailcfg/derpmap.go
+++ b/tailcfg/derpmap.go
@@ -67,6 +67,17 @@ type DERPRegion struct {
 	// away to a new region without Avoid set.
 	Avoid bool `json:",omitempty"`
 
+	// Location is an (optional) geographic location for this region; it is
+	// used in conjunction with ambient location information like the
+	// current cloud provider's region to try optimal regions first and
+	// exclude DERP regions on the other side of the world from being
+	// checked.
+	//
+	// If this region has no location, it should always be considered a
+	// candidate as maybe being the closest. That is, it should not be
+	// ruled out as a candidate if the node knows its own actual location.
+	Location *DERPLocation `json:",omitempty"`
+
 	// Nodes are the DERP nodes running in this region, in
 	// priority order for the current client. Client TLS
 	// connections should ideally only go to the first entry
@@ -142,6 +153,15 @@ type DERPNode struct {
 	// STUNTestIP is used in tests to override the STUN server's IP.
 	// If empty, it's assumed to be the same as the DERP server.
 	STUNTestIP string `json:",omitempty"`
+}
+
+// DERPLocation contains information about a DERP region's physical location.
+type DERPLocation struct {
+	// Latitude contains the latitude component of this location.
+	Latitude float64
+
+	// Longitude contains the longitude component of this location.
+	Longitude float64
 }
 
 // DotInvalid is a fake DNS TLD used in tests for an invalid hostname.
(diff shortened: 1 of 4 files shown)


tailcfg/derpmap.go
```diff
@@ -144,5 +155,14 @@ type DERPNode struct {
 	STUNTestIP string `json:",omitempty"`
 }
 
+// DERPLocation contains information about a DERP region's physical location.
+type DERPLocation struct {
```
DentonGentry, line 159: The geohash locations we use in https://mon.corp.ts.net/d/lBFZ0P_Wk/derp?orgId=1 are the center of the city where the datacenter is located, which will probably be fine.

Those geohashes are supplied as metadata in the prometheus.yml.

tailcfg/derpmap.go
```diff
@@ -67,6 +67,17 @@ type DERPRegion struct {
...
 
+	// Location is an (optional) geographic location for this region; it is
+	// used in conjunction with ambient location information like the
+	// current cloud provider's region to try optimal regions first and
+	// exclude DERP regions on the other side of the world from being
+	// checked.
+	//
+	// If no location is provided, then any location-based checks will not
+	// exclude this region from consideration; i.e. the region will be
+	// considered "sufficiently close" to wherever the node is.
```
bradfitz, line 78: I don't follow this paragraph.

andrew-d: Just re-worded it. PTAL?

tailcfg/derpmap.go
```diff
@@ -67,6 +67,17 @@ type DERPRegion struct {
...
 
+	// Location is an (optional) geographic location for this region; it is
+	// used in conjunction with ambient location information like the
+	// current cloud provider's region to try optimal regions first and
+	// exclude DERP regions on the other side of the world from being
+	// checked.
+	//
+	// If this region has no location, it should be treated as always
+	// "close" to the current node–e.g. not excluded from netcheck probes
+	// by a location-based heuristic.
```
bradfitz, line 78: I would say "If this region has no location, it should always be considered a candidate as maybe being the closest. That is, it should not be ruled out as a candidate if the node knows its own actual location."

BTW, Go standard library doc style is to avoid all Latin abbreviations. I'm not super picky about it, but I've come to kinda like it.

tailcfg/derpmap.go
bradfitz: Ooooh, yeah, let's use a Geohash string omitempty instead of a new type.

Plus geohash was made by somebody pretty active in the Go community :)

tailcfg/derpmap.go
andrew-d: Yeah, I like that wording; pushed 👍

tailcfg/derpmap.go
andrew-d: Is there any reason to use a geohash on the wire? It's one thing to show that to a user, but you'd have to turn them into a lat/lon pair to do distance calculations anyway, so I don't think it really gains us anything. Lat/lon is pretty unambiguous, IMO.

raggi:
Implemented slightly differently in https://github.com/tailscale/tailscale/commit/352c1ac96c363d5618f042e72ca05ad955c91bbb / #11162 

This can be closed now.

PR #7548, opened 2023-03-13
cmd/k8s-operator: require HTTPS to be enabled for AuthProxy

maisem:
Updates #5055

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -822,7 +822,7 @@ func (s *Server) ListenTLS(network, addr string) (net.Listener, error) {
 		return nil, err
 	}
 	if len(st.CertDomains) == 0 {
-		return nil, errors.New("tsnet: you must enable HTTPS in the admin panel to proceed")
+		return nil, errors.New("tsnet: you must enable HTTPS in the admin panel to proceed. See https://tailscale.com/kb/1153/enabling-https/")
 	}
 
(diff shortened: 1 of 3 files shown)


tsnet/tsnet.go
```diff
@@ -822,7 +822,7 @@ func (s *Server) ListenTLS(network, addr string) (net.Listener, error) {
 		return nil, err
 	}
 	if len(st.CertDomains) == 0 {
-		return nil, errors.New("tsnet: you must enable HTTPS in the admin panel to proceed")
+		return nil, errors.New("tsnet: you must enable HTTPS in the admin panel to proceed. See https://tailscale.com/kb/1153/enabling-https/")
```
danderson, line 825: You may want to set up a tailscale.com/s/... shortlink, so we can point the link elsewhere later if we refactor the KB.

danderson (approved):

tsnet/tsnet.go
maisem: I agree, I am going to send a clean up PR as we have many KBs linked directly from the CLI.

maisem: (this one is already linked in the CLI which is why I used it, working on the relevant PRs now)

maisem: https://github.com/tailscale/tailscale/pull/7549

PR #7552, opened 2023-03-13
derp, derphttp, magicsock: send new unknown peer frame when destination is unknown

### LINKED ISSUE #4326: debug: make DERP send frame (framePeerGone) as limited reply as "nobody's here"

valscale:
If node A sends a CallMeMaybe packet to DERP for node B, and DERP has no record of node B, send an "unknown peer" frame back to node A.

Uses the existing framePeerGone and adds a reason byte at the end in what is intended to be a backwards-compatible manner.

Fixes #4326


derp/derp_server.go
```diff
@@ -960,7 +967,7 @@ type dropReason int
 const (
 	dropReasonUnknownDest      dropReason = iota // unknown destination pubkey
 	dropReasonUnknownDestOnFwd                   // unknown destination pubkey on a derp-forwarded packet
-	dropReasonGone                               // destination tailscaled disconnected before we could send
+	dropReasonGoneDisconnected                   // destination tailscaled disconnected before we could send
```
DentonGentry, line 970: There is no dropReasonGoneUnknown here because we don't actually have any such cases in the current code?

(Note: if this comment seems nonsensical, it probably is. I don't understand this part of the code very well)

valscale: I *think* that the already existing "dropReasonUnknownDest" and "dropReasonUnknownDestOnFwd" are the variables that were already being incremented for this case (the ones described in the original issue text) so all we need to do is actually send a peerGone packet back to the client.

derp/derp_client.go
```diff
@@ -524,7 +532,16 @@ func (c *Client) recvTimeout(timeout time.Duration) (m ReceivedMessage, err erro
...
+			// Backward compatibility for the older peerGone version without reason byte
+			var reason byte = peerGoneReasonDisconnected
+			if n != keyLen {
+				reason = b[keyLen]
+			}
+			if reason == peerGoneReasonUnknown {
+				pg := PeerGoneMessageUnknown(key.NodePublicFromRaw32(mem.B(b[:keyLen])))
+				return pg, nil
+			}
+			pg := PeerGoneMessageDisconnected(key.NodePublicFromRaw32(mem.B(b[:keyLen])))
```
jwhited, line 544: If we add more reason codes later they will end up classified here until the clients update.

valscale: Thank you, this is one of the places where I am not sure what is the most forward/backward compatible way. Fundamentally, it comes down to this question:

Should future new peerGone reasons cause the client to log an informative error message or not?

The current code preserves the old behavior (no log message) for new reason codes. The alternative would cause the new behavior (log a message) for all future new codes.

I am not convinced that either is correct so happy to hear other people's thoughts!

derp/derp_client.go
```diff
@@ -347,12 +347,20 @@ type ReceivedPacket struct {
...
+// server.
+type PeerGoneMessageDisconnected key.NodePublic
 
-func (PeerGoneMessage) msg() {}
+func (PeerGoneMessageDisconnected) msg() {}
+
+// PeerGoneMessageUnknown is a ReceivedMessage that indicates that the
+// client identified by the underlying public key is not connected to
+// this server and we don't know why.
+type PeerGoneMessageUnknown key.NodePublic
```
jwhited, line 361: Here `Unknown` means the server doesn't know the reason, but also *could* be used to mean the reason code was a value that the client doesn't know. Maybe we can disambiguate those cases more explicitly in the naming of things. Relates to classification of unknown reason code values below.

derp/derp.go
```diff
@@ -116,6 +116,14 @@ const (
 	frameRestarting = frameType(0x15)
 )
 
+type peerGoneType byte
```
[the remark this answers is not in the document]
bradfitz: Never use iota for values that are part of network protocols or file formats. When the value matters, spell it out. (style/safety thing)

Only for local process only enum values.

That said, I haven't looked at this PR at all yet.

jwhited:
I haven't gotten very deeply into DERP yet, so my comments come with a grain of salt.

derp/derp.go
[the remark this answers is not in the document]
valscale: Ha, I got so excited about using go style I forgot about network compatibility. :) Fixing.

valscale:
Upon reflection, I'm rewriting this to push the reason back up the call chain instead of creating a new chan for each different peerGone reason.

derp/derp_client.go
valscale: I ended up pushing the gone reason out to the client so it could decide what to do. Currently the clients log all the reasons other than disconnected.

derp/derp_client.go
valscale: Thank you, this is now one PeerGoneMessage and the reasons are named more clearly.

derp/derp.go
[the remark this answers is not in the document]
valscale: New version actually uses this type in addition to fixing the stable network protocol issue.

derp/derp_server.go
```diff
@@ -878,6 +880,9 @@ func (c *sclient) handleFrameForwardPacket(ft frameType, fl uint32) error {
 		reason := dropReasonUnknownDestOnFwd
 		if dstLen > 1 {
 			reason = dropReasonDupClient
+		} else {
+			// TODO: rate-limit
+			go c.requestPeerGoneWrite(dstKey, PeerGoneReasonNotHere)
```
bradfitz, line 885: I think we should probably only send this type of DERP frame on disco packets (CallMeMaybe messages), not on WireGuard packets.

derp/derp_server.go
```diff
@@ -1301,6 +1313,12 @@ type pkt struct {
 	bs []byte
 }
 
+// peerGoneMsg is a request to write a peerGone frame to an sclient
```
bradfitz, line 1316: trailing period

derp/derp_server.go
```diff
@@ -1452,13 +1470,21 @@ func (c *sclient) sendPong(data [8]byte) error {
...
+func (c *sclient) sendPeerGone(peer key.NodePublic, reason PeerGoneReasonType) error {
+	switch reason {
+	case PeerGoneReasonDisconnected:
+		c.s.peerGoneDisconnectedFrames.Add(1)
+	case PeerGoneReasonNotHere:
+		c.s.peerGoneNotHereFrames.Add(1)
+	}
 	c.setWriteDeadline()
-	if err := writeFrameHeader(c.bw.bw(), framePeerGone, keyLen); err != nil {
+	data := peer.AppendTo(nil)
```
danderson, line 1481: Very slight optimization, but might be worth it in the derp sending path:

```
data := make([]byte, 0, 33) // 32 for key + 1 for reason, there's probably a constant kicking around for this value
data = peer.AppendTo(bs)
data = append(bs, byte(reason))
```

That'll make this send be one alloc instead of two (one for the peer.AppendTo that allocates a 32b slice, then one more to grow the slice for the reason).

Although really this specific send path is infrequently used, so it probably doesn't matter. But in our network-ish code, keeping awareness of allocs-per-packet is a useful background thread to cultivate.

wgengine/magicsock/magicsock.go
```diff
@@ -1727,7 +1727,16 @@ func (c *Conn) runDerpReader(ctx context.Context, derpFakeAddr netip.AddrPort, d
 		case derp.HealthMessage:
 			health.SetDERPRegionHealth(regionID, m.Problem)
 		case derp.PeerGoneMessage:
-			c.removeDerpPeerRoute(key.NodePublic(m), regionID, dc)
+			switch m.Reason {
+			case derp.PeerGoneReasonDisconnected:
+			case derp.PeerGoneReasonNotHere:
+				c.logf("magicsock: derp-%d does not know about peer %s, removing route",
```
danderson, line 1733: A convention in the Tailscale codebase: start the log message string with `[unexpected]` to signal "this shouldn't be happening. i.e. `c.logf("[unexpected] magicsock: derp-%s ...")`

That way we can look for unexpected events in debug logs and keep track of how much unexpected stuff is going on.

danderson (commented):
Looks right to me. Some nitpicky comments since you asked for Go style feedback, but aside from the missing rate limiter I'd say this is good to go.

derp/derp_client.go
```diff
@@ -524,7 +527,15 @@ func (c *Client) recvTimeout(timeout time.Duration) (m ReceivedMessage, err erro
 				c.logf("[unexpected] dropping short peerGone frame from DERP server")
 				continue
 			}
-			pg := PeerGoneMessage(key.NodePublicFromRaw32(mem.B(b[:keyLen])))
+			// Backward compatibility for the older peerGone without reason byte
```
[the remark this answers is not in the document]
danderson: And actually, I recall bradfitz explicitly made the derp protocol such that clients are tolerant of receiving extra bytes at the end of frames that they don't know about, to allow for exactly this kind of change without having to bump protocol versions. So, as long as the added bytes are understanding-optional, the way you did it here is fine.

derp/derp_server.go
valscale: So always test the packet to see if it is CallMeMaybe and only send if it is? Any thoughts on rate-limiting, should that be per server, per client, per false destination, something else?

derp/derp_server.go
```diff
@@ -931,14 +936,18 @@ func (c *sclient) handleFrameSendPacket(ft frameType, fl uint32) error {
 		if fwd != nil {
 			s.packetsForwardedOut.Add(1)
 			if err := fwd.ForwardPacket(c.key, dstKey, contents); err != nil {
-				// TODO:
+				// TODO: rate-limit
```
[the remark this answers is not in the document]
valscale: Add in the request to only send for CallMeMaybe packets and this definitely needs a helper method.

derp/derp_server.go
valscale: It now checks for CallMeMaybe with LooksLikeDiscoWrapper() and sends only 3 PeerGone packets per second.

wgengine/magicsock/magicsock.go
```diff
@@ -1727,7 +1727,16 @@ func (c *Conn) runDerpReader(ctx context.Context, derpFakeAddr netip.AddrPort, d
 		case derp.HealthMessage:
 			health.SetDERPRegionHealth(regionID, m.Problem)
 		case derp.PeerGoneMessage:
-			c.removeDerpPeerRoute(key.NodePublic(m), regionID, dc)
+			switch m.Reason {
+			case derp.PeerGoneReasonDisconnected:
+			case derp.PeerGoneReasonNotHere:
```
bradfitz, line 1732: you know this isn't like C, right? (There's no implicit `fallthrough`)

It's not clear this is intentional. Maybe add a comment like `// Do nothing.` to the first case.

bradfitz: Or if you want to do it on both cases, do "case A, B:"

bradfitz:
Nit: lowercase verb after colon in `derp: Remove dead sclient struct member replaceLimiter`

wgengine/magicsock/magicsock.go
valscale: Thanks! Comment added.

valscale:
Differences from first version:

* Tested end-to-end with all combos of new/old DERP and clients
* Only sent on bad disco messages
* Debug knob added to force bad CallMeMaybe for testing
* Rate limited to 3 per second per client
* Actually uses PeerGoneReasonType
* Unused ratelimiter removal separated into its own commit

I'd be happy to merge this one but still appreciate nitpicky feedback.

wgengine/magicsock/magicsock.go
```diff
@@ -1727,7 +1727,17 @@ func (c *Conn) runDerpReader(ctx context.Context, derpFakeAddr netip.AddrPort, d
 		case derp.HealthMessage:
 			health.SetDERPRegionHealth(regionID, m.Problem)
 		case derp.PeerGoneMessage:
-			c.removeDerpPeerRoute(key.NodePublic(m), regionID, dc)
+			switch m.Reason {
+			case derp.PeerGoneReasonDisconnected:
+				// Do nothing.
+			case derp.PeerGoneReasonNotHere:
+				c.logf("[unexpected] magicsock: derp-%d does not know about peer %s, removing route",
```
mihaip, line 1734: May want to increment a client metric if this is an unexpected/interesting thing to happen (there's already a few DERP-related metrics in this file).

mihaip (approved):

wgengine/magicsock/magicsock.go
valscale: Good idea! I added two new metrics, noticed two arguments were swapped for the log message for the unknown reason case and fixed that, and tested the not here path. If you can double check the logf() after metricRecvDiscoDERPPeerGoneUnknown.Add(1), I would appreciate it.

mihaip (approved):
Metrics and logf LGTM

PR #7553, opened 2023-03-13
cmd/k8s-operator: drop trailing dot in tagged node name

maisem:
Also update tailcfg docs.

Updates #5055

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -183,7 +183,12 @@ func (emptyStructJSONSlice) UnmarshalJSON([]byte) error { return nil }
 type Node struct {
 	ID       NodeID
 	StableID StableNodeID
-	Name     string // DNS
+
+	// Name is the FQDN of the node.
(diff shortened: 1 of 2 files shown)


danderson (approved):

tailcfg/tailcfg.go
```diff
@@ -183,7 +183,9 @@ func (emptyStructJSONSlice) UnmarshalJSON([]byte) error { return nil }
 type Node struct {
 	ID       NodeID
 	StableID StableNodeID
-	Name     string // DNS
+
+	// Name is the FQDN of the node. It has a trailing dot.
```
bradfitz, line 187: you dropped the word DNS in here. Add it back, and maybe give a full example value in quotes while here?

bradfitz (approved):

tailcfg/tailcfg.go
maisem: Done

PR #7559, opened 2023-03-14
tsnet: avoid deadlock on close

maisem:
tsnet.Server.Close was calling listener.Close with the server mutex held, but the listener close method tries to grab that mutex, resulting in a deadlock.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -117,6 +117,7 @@ type Server struct {
 	mu        sync.Mutex
 	listeners map[listenKey]*listener
 	dialer    *tsdial.Dialer
+	closed    bool
 }
 
 // Dial connects to the address on the tailnet.
@@ -302,6 +303,11 @@ func (s *Server) Up(ctx context.Context) (*ipnstate.Status, error) {
 //
 // It must not be called before or concurrently with Start.
 func (s *Server) Close() error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.closed {
+		return fmt.Errorf("tsnet: %w", net.ErrClosed)
+	}
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 	var wg sync.WaitGroup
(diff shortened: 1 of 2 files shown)


maisem:
Taking over for https://github.com/tailscale/tailscale/pull/7542/files

tsnet/tsnet.go
```diff
@@ -1012,15 +1017,25 @@ func (ln *listener) Accept() (net.Conn, error) {
...
 
 func (ln *listener) Addr() net.Addr { return addr{ln} }
+
 func (ln *listener) Close() error {
 	ln.s.mu.Lock()
 	defer ln.s.mu.Unlock()
+	return ln.closeLocked()
+}
+
+// closeLocked closes the listener. It must be called with s.mu held.
```
bradfitz, line 1027: // It must ...

on its own line

bradfitz: And "with ln.s.mu held".

tsnet/tsnet_test.go
```diff
@@ -344,3 +345,24 @@ func TestTailscaleIPs(t *testing.T) {
 			sIp4, upIp4, sIp6, upIp6)
 	}
 }
+
+func TestListenerCleanup(t *testing.T) {
```
bradfitz, line 349: comment somewhere in here that this used to deadlock or `// verify this doesn't deadlock [anymore]` ?

bradfitz (approved):

tsnet/tsnet_test.go
maisem: Done

tsnet/tsnet.go
maisem: Done

tsnet/tsnet.go
```diff
@@ -997,10 +1001,11 @@ type listenKey struct {
...
 type listener struct {
-	s    *Server
-	keys []listenKey
-	addr string
-	conn chan net.Conn
+	s      *Server
+	keys   []listenKey
+	addr   string
+	conn   chan net.Conn
+	closed bool // guarded by s.mu
```
DentonGentry, line 1008: I think all of the fields are guarded by s.mu, it isn't safe to manipulate s.listeners without holding s.mu.

DentonGentry (approved):

tsnet/tsnet.go
maisem: These are immutable after creation, so they can be read without holding a lock.

rejaulkarimm299-code:
1. 
```html
Close 
```

PR #7566, opened 2023-03-14
version/mkversion: don't break on tagged go.mod entries

danderson:
I thought our versioning scheme would make go.mod include a commit hash even on stable builds. I was wrong. Fortunately, the rest of this code wants anything that 'git rev-parse' understands (to convert it into a full git hash), and tags qualify.

### DIFF
--- a/version/mkversion/mkversion.go
+++ b/version/mkversion/mkversion.go
@@ -150,14 +150,14 @@ func InfoFrom(dir string) (VersionInfo, error) {
 	}
 
 	// Note, this mechanism doesn't correctly support go.mod replacements,
-	// or go workdirs. We only parse out the commit hash from go.mod's
+	// or go workdirs. We only parse out the commit ref from go.mod's
 	// "require" line, nothing else.
-	tailscaleHash, err := tailscaleModuleHash(modBs)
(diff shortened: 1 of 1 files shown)


version/mkversion/mkversion.go
```diff
@@ -187,7 +187,8 @@ func tailscaleModuleHash(modBs []byte) (string, error) {
 		if i := strings.LastIndexByte(req.Mod.Version, '-'); i != -1 {
 			return req.Mod.Version[i+1:], nil
 		}
-		return "", fmt.Errorf("couldn't parse git hash from tailscale.com version %q", req.Mod.Version)
+		// If there are no dashes, the version is a tag.
```
maisem, line 190: update comment above too? it says `git hash`

maisem (approved):

version/mkversion/mkversion.go
maisem: oh even the func says `tailscaleModuleHash` maybe `tailscaleModuleHashOrTag`

danderson: Fixed, and also renamed variables and functions to say "ref" instead of "hash" for this value.

PR #7575, opened 2023-03-15
tailscale/cmd: Warn for up --force-reauth over SSH without accepting risk

### LINKED ISSUE #6377: tailscale up --force-reauth does not require --accept-risk=lose-ssh
### What is the issue?

I received a report from a friend that "`--accept-risk=lose-ssh` does not kick in if you do `--force-reauth`, and then you lose connectivity to the host".

We should handle this case.

### Steps to reproduce

_No response_

_No response_

### OS

_No response_

### OS version

_No response_

### Tailscale version

_No response_

### Bug report

_No response_

zofrex:
Give a warning that issuing `tailscale up --force-reauth` when you're logged in via SSH over Tailscale will disconnect you, and give a 5 second countdown, overridable with`--accept-risk=lose-ssh`. Similar to the existing behaviour with `tailscale down` and running `up` to change SSH configuration.

Fixes #6377

### DIFF
--- a/cmd/tailscale/cli/cli_test.go
+++ b/cmd/tailscale/cli/cli_test.go
@@ -1071,6 +1071,32 @@ func TestUpdatePrefs(t *testing.T) {
 			},
 			env: upCheckEnv{backendState: "Running"},
 		},
+		{
+			name:             "force_reauth_over_ssh_no_risk",
+			flags:            []string{"--force-reauth"},
+			sshOverTailscale: true,
+			curPrefs: &ipn.Prefs{
+				ControlURL:       "https://login.tailscale.com",
+				AllowSingleHosts: true,
+				CorpDNS:          true,
+				NetfilterMode:    preftype.NetfilterOn,
+			},
+			env:          upCheckEnv{backendState: "Running"},
+			wantErrSubtr: "aborted, no changes made",
+		},
+		{
+			name:             "force_reauth_over_ssh",
+			flags:            []string{"--force-reauth", "--accept-risk=lose-ssh"},
+			sshOverTailscale: true,
+			curPrefs: &ipn.Prefs{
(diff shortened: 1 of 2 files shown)


cmd/tailscale/cli/up.go
```diff
@@ -409,6 +409,12 @@ func updatePrefs(prefs, curPrefs *ipn.Prefs, env upCheckEnv) (simpleUp bool, jus
 		return false, nil, err
```
zofrex, line 409: Is this worth folding into `present SSHToggleRisk` above here?

cmd/tailscale/cli/cli_test.go
```diff
@@ -1071,6 +1071,32 @@ func TestUpdatePrefs(t *testing.T) {
...
+				ControlURL:       "https://login.tailscale.com",
+				AllowSingleHosts: true,
+				CorpDNS:          true,
+				NetfilterMode:    preftype.NetfilterOn,
+			},
+			env:          upCheckEnv{backendState: "Running"},
+			wantErrSubtr: "aborted, no changes made",
+		},
+		{
+			name:             "force_reauth_over_ssh",
```
maisem, line 1088: Does this cause a 5s wait in the test?

zofrex: No, it doesn't. I don't really understand how this is the case, but I've cribbed this test from the tests like `disable_ssh_over_ssh_no_risk` (line 983 in this file) which is testing similar functionality (that also has a 5s wait when run interactively).

zofrex: Ah, it's because of this:

```
func presentRiskToUser(riskType, riskMessage, acceptedRisks string) error {
	// ...
	if inTest() {
		return errAborted
	}
```

maisem (approved):

PR #7577, opened 2023-03-15
ipn: json encode PeerStatus.UserID as string

### LINKED ISSUE #7576: large numbers in `tailscale status -json` output can't be parsed in some languages
### What is the issue?

`tailscale status -json` returns UserID as a JSON number in the Self and Peer objects.  These are 64 bit integers which can't be parsed in some languages (javascript, lua, et al).  The User object however uses strings as its keys.

```
$ tailscale status -json

{
  "Version": "1.37.278-t6ef2105a8-g486270c2c",
  ...
  "Self": {
    "UserID": 39003421729112490,
    ...
  },
  "User": {
    "39003421729112490": {
      "ID": 39003421729112490,
      "DisplayName": "Will Norris",
    },
  }
}
```

We should probably encode the UserID as a string so that it can be properly parsed, and then matched to entries in the "User" object.

### Steps to reproduce

_No response_

_No response_

### OS

_No response_

### OS version

_No response_

### Tailscale version

_No response_

### Other software

_No response_

### Bug report

_No response_

willnorris:
Fixes #7576

### DIFF
--- a/ipn/ipnstate/ipnstate.go
+++ b/ipn/ipnstate/ipnstate.go
@@ -191,8 +191,8 @@ type PeerStatus struct {
 	// DNSName is the Peer's FQDN. It ends with a dot.
 	// It has the form "host.<MagicDNSSuffix>."
 	DNSName string
-	OS      string // HostInfo.OS
-	UserID  tailcfg.UserID
+	OS      string         // HostInfo.OS
+	UserID  tailcfg.UserID `json:",string"`
 
 	// TailscaleIPs are the IP addresses assigned to the node.
 	TailscaleIPs []netip.Addr


ipn/ipnstate/ipnstate.go
```diff
@@ -183,6 +183,13 @@ type PeerStatusLite struct {
 	NodeKey key.NodePublic
 }
 
+type UserID tailcfg.UserID
+
+// MarshalJSON marshals a UserID as a JSON string.
```
maisem, line 188: you probably want an unmarshal method too

ipn/ipnstate/ipnstate.go
```diff
@@ -183,6 +183,13 @@ type PeerStatusLite struct {
 	NodeKey key.NodePublic
 }
 
+type UserID tailcfg.UserID
```
maisem, line 186: add a comment why this exists

ipn/ipnstate/ipnstate.go
```diff
@@ -522,7 +529,7 @@ table tbody tr:nth-child(even) td { background-color: #f5f5f5; }
 			}
 		}
 		var owner string
-		if up, ok := st.User[ps.UserID]; ok {
+		if up, ok := st.User[tailcfg.UserID(ps.UserID)]; ok {
```
maisem, line 532: i think you should also make this map be keyed of `UserID` instead of `tailcfg.UserID`

mihaip (commented):
Note that Swift also parses this JSON, and it expects user IDs to be int64s: https://github.com/tailscale/corp/blob/5322306defd9c8f1433b571fc180acede85e1163/xcode/Shared/Go.swift#L429

Used at: https://github.com/tailscale/corp/blob/5322306defd9c8f1433b571fc180acede85e1163/xcode/Shared/Go.swift#L276

ipn/ipnstate/ipnstate.go
```diff
@@ -183,6 +185,26 @@ type PeerStatusLite struct {
...
+type UserID tailcfg.UserID
+
+// MarshalJSON marshals a UserID as a JSON string.
+func (u UserID) MarshalJSON() ([]byte, error) {
+	return fmt.Appendf(nil, `"%d"`, u), nil
+}
+
+// UnmarshalJSON unmarshals a UserID from a JSON string.
+func (u *UserID) UnmarshalJSON(b []byte) error {
+	b = bytes.Trim(b, `"`)
```
creachadair, line 198: Unlikely to occur in practice, but this would successfully unmarshal `"12` and `""456""`. (We use it inside a bigger object so we get the outside validity check so it isn't a huge risk)

willnorris: true.  Unlikely, but I've gone ahead and updated to verify JSON string syntax.

ipn/ipnstate/ipnstate.go
```diff
@@ -183,6 +185,29 @@ type PeerStatusLite struct {
...
+type UserID tailcfg.UserID
+
+// MarshalJSON marshals a UserID as a JSON string.
+func (u UserID) MarshalJSON() ([]byte, error) {
+	return fmt.Appendf(nil, `"%d"`, u), nil
+}
+
+// UnmarshalJSON unmarshals a UserID from a JSON string.
+func (u *UserID) UnmarshalJSON(b []byte) error {
+	if b[0] != '"' || b[len(b)-1] != '"' {
```
bradfitz, line 198: verify it has len(b) >= 2 first

willnorris: oops, yeah.

ipn/ipnstate/ipnstate.go
```diff
@@ -183,6 +185,29 @@ type PeerStatusLite struct {
...
+type UserID tailcfg.UserID
+
+// MarshalJSON marshals a UserID as a JSON string.
+func (u UserID) MarshalJSON() ([]byte, error) {
+	return fmt.Appendf(nil, `"%d"`, u), nil
+}
+
+// UnmarshalJSON unmarshals a UserID from a JSON string.
+func (u *UserID) UnmarshalJSON(b []byte) error {
+	if len(b) < 2 || b[0] != '"' || b[len(b)-1] != '"' {
```
creachadair, line 198: I thought part of the objective was to handle either numeric of string encoding, though. Otherwise isn't this a breaking change? 

Maybe:

```go
if len(b) >= 2 && b[0] == '"' && b[len(b)-1] != '"' {
  fail
} else v, err := strconv.Atoi(string(b[1:len(b)-1])); err != nil {
  fail
}
return stconv.Atoi(string(b))
```

or something like that?

willnorris: We could have it support either, but I don't think it's necessary.  I believe this is only used locally as part of LocalAPI, so it's encoded and decoded as the same string type on both sides.

creachadair: I see—in that case, we could do this more easily by implementing `text.Unmarshaler` instead. `json.Marshal` will call it with an unmarshaled string value already resolved.

See https://pkg.go.dev/encoding/json#Unmarshal amid the comments.

willnorris: TIL!  yeah, that's a lot simpler. Thanks!

willnorris: hahaha, I'm a dork.  This could be handled simply be adding a tag, there's no need for a new type:

``` go
type PeerStatus struct {
    UserID  tailcfg.UserID `json:",string"`
}
```

creachadair: Good point—I almost suggested that earlier, but then I got rabbitholed thinking we wanted to support both formats 😆

willnorris:
abandoning this PR for now while we discuss future of LocalAPI changes

PR #7583, opened 2023-03-16
build_dist.sh: make cross-compilation friendly

nshalman:
For the sunos port I do cross compilation on GitHub to publish release binaries for anyone not using distro packages.

This is how I fixed `build_dist.sh` to work for me, but ultimately I have no strong feelings on how this is accomplished.

I do think it's worth keeping that script cross-compilation friendly.


bradfitz:
This looks like your department, @danderson.

danderson (approved):
Yup, works for me. Thanks!

PR #7593, opened 2023-03-17
cmd/tailscale/cli: fix inconsistency between serve text and example command

mihaip:
Use the same local port number in both.

### DIFF
--- a/cmd/tailscale/cli/serve.go
+++ b/cmd/tailscale/cli/serve.go
@@ -45,7 +45,7 @@ serve https:<port> <mount-point> <source> [off]
 
 The 'tailscale serve' set of commands allows you to serve
 content and local servers from your Tailscale node to
-your tailnet. 
+your tailnet.
 
 You can also choose to enable the Tailscale Funnel with:
 'tailscale funnel on'. Funnel allows you to publish
@@ -66,10 +66,12 @@ EXAMPLES
   - To serve simple static text:
     $ tailscale serve https:8080 / text:"Hello, world!"
 
-  - To forward raw TCP packets to a local TCP server on port 5432:
+  - To forward incoming TCP connections on port 2222 to a local TCP server on
+    port 22 (e.g. to run OpenSSH in parallel with Tailscale SSH):
     $ tailscale serve tcp:2222 tcp://localhost:22
 
-  - To forward raw, TLS-terminated TCP packets to a local TCP server on port 80:
+  - To accept TCP TLS connections (terminated within tailscaled) proxied to a
+    local plaintext server on port 80:
     $ tailscale serve tls-terminated-tcp:443 tcp://localhost:80
 `),
 		Exec:      e.runServe,


cmd/tailscale/cli/serve.go
```diff
@@ -67,7 +67,7 @@ EXAMPLES
     $ tailscale serve https:8080 / text:"Hello, world!"
 
   - To forward raw TCP packets to a local TCP server on port 5432:
-    $ tailscale serve tcp:2222 tcp://localhost:22
+    $ tailscale serve tcp:2222 tcp://localhost:5432
```
mihaip, line 70: Not sure if the 2222 should also be called out in the text as the incoming port.

bradfitz: The line before is also slightly misleadingly wrong. It's not forwarding "raw TCP packets"; it's reconstruction a new TCP stream and proxying bytes between them.

"To forward incoming TCP connections to a local TCP server on ..."

mihaip: Should the `tls-terminated-tcp` option below get a similar change?

bradfitz: yup

mihaip: Done.

cmd/tailscale/cli/serve.go
```diff
@@ -66,10 +66,10 @@ EXAMPLES
   - To serve simple static text:
     $ tailscale serve https:8080 / text:"Hello, world!"
 
-  - To forward raw TCP packets to a local TCP server on port 5432:
-    $ tailscale serve tcp:2222 tcp://localhost:22
+  - To forward incoming TCP connections on port 2222 to a local TCP server on port 5432:
```
bradfitz, line 69: this line's now 88 chars wide. From lines above, it looks like we're trying to be 80 char friendly. (some people configure their tiling window managers to have NxM terminals where each is 80 cols wide)

cmd/tailscale/cli/serve.go
```diff
@@ -66,10 +66,10 @@ EXAMPLES
...
+    $ tailscale serve tcp:2222 tcp://localhost:5432
 
-  - To forward raw, TLS-terminated TCP packets to a local TCP server on port 80:
+  - To forward incoming TLS-terminated TCP connections to a local TCP server on port 80:
```
bradfitz, line 72: I think the word order is wrong now for what this mode does.

Maybe: "To accept TCP TLS connections, terminate their TLS inside tailscaled, and then proxy to a local plaintext server on port 80, wrapping the plaintext response in TLS to the client:"

Or something. @shayne can help wordsmith.

shayne:
```suggestion
  - Accept TCP TLS connections (terminated within tailscaled) proxied to a local plaintext server on port 80. 
  ```

shayne:
Regarding the `$ tailscale serve tcp:2222 tcp://localhost:22`, this was a use case/customer question that popped up a couple of times. It allows you to rebind OpenSSH to port 2222 on your tailnet, allowing access to the real SSH server if you're running Tailscale SSH.

Access isn't allowed via Funnel (non-TLS and the port is not allowed), but for serving only on your tailnet, you can proxy non-TLS-wrapped TCP connections on any valid port. Before the recent CLI updates, `serve` restricted `--serve-port` to the three we allowed on Funnel (`443,8443,10000`).

shayne (approved):
Thanks for the updates! I added a comment and suggestion—otherwise, LGTM.

shayne:
Typo in commit message:
> fix inconsistency **between** serve text and ...

cmd/tailscale/cli/serve.go
mihaip: Incorporated Shayne's suggestion.

mihaip:
> Regarding the `$ tailscale serve tcp:2222 tcp://localhost:22`, this was a use case/customer question that popped up a couple of times. It allows you to rebind OpenSSH to port 2222 on your tailnet, allowing access to the real SSH server if you're running Tailscale SSH.

Mentioned this more explicitly, I think it's a good use case that we should document.

PR #7609, opened 2023-03-19
ssh/tailssh: fix privilege dropping on FreeBSD

andrew-d:
On FreeBSD, changing a process's supplementary groups with setgroups(2) will also change the egid of the process, setting it to the first entry in the provided list. This is distinct from the behaviour on other platforms (and possibly a violation of the POSIX standard).

Because of this, our incubator code wouldn't change the process's gid, because it would read the newly-changed egid, compare it against the expected egid, and since they matched, not change the gid.

This could be observed by running "id -p" in two contexts. The expected output, and the output returned when running from a SSH shell, is:

    andrew@freebsd:~ $ id -p
    uid         andrew
    groups      andrew

However, when run via "ssh andrew@freebsd id -p", the output would be:

    $ ssh andrew@freebsd id -p
    login       root
    uid         andrew
    rgid        wheel
    groups      andrew

(this could also be observed via "id -g -r" to print just the gid)

We fix this by ensuring that the setGroups call behaves the same on FreeBSD as it does on other platforms by adding the current egid as the first entry in the list, and thus not changing it unexpectedly.

More information can be found in the following article:
    https://www.usenix.org/system/files/login/articles/325-tsafrir.pdf

Also, while we're here... defensively call runtime.LockOSThread to make it less likely we trip over bugs like golang/go#1435

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -204,6 +204,12 @@ func parseIncubatorArgs(args []string) (a incubatorArgs) {
 // OS, sets its UID and groups to the specified `--uid`, `--gid` and
 // `--groups` and then launches the requested `--cmd`.
 func beIncubator(args []string) error {
+	// To defend against issues like https://golang.org/issue/1435,
+	// defensively lock our current goroutine's thread to the current
+	// system thread before we start making any UID/GID/group changes.
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
 	ia := parseIncubatorArgs(args)
 	if ia.isSFTP && ia.isShell {
 		return fmt.Errorf("--sftp and --shell are mutually exclusive")
@@ -771,6 +777,20 @@ func (ia *incubatorArgs) loginArgs() []string {
 }
 
 func setGroups(groupIDs []int) error {
+	if runtime.GOOS == "freebsd" || runtime.GOOS == "darwin" {
+		// On FreeBSD and Darwin, the first entry returned from the
+		// getgroups(2) syscall is the egid, and changing it with
+		// setgroups(2) changes the egid of the process. This is
+		// technically a violation of the POSIX standard; see the
+		// following article for more detail:
+		//    https://www.usenix.org/system/files/login/articles/325-tsafrir.pdf
+		//
+		// In this case, we add an entry at the beginning of the
+		// groupIDs list containing the current egid, which replicates
+		// the behaviour on every other platform of just changing the
+		// supplementary groups and not the egid.
+		groupIDs = append([]int{os.Getegid()}, groupIDs...)
+	}
 	if runtime.GOOS == "darwin" && len(groupIDs) > 16 {
 		// darwin returns "invalid argument" if more than 16 groups are passed to syscall.Setgroups
 		// some info can be found here:


ssh/tailssh/incubator.go
```diff
@@ -204,6 +204,11 @@ func parseIncubatorArgs(args []string) (a incubatorArgs) {
 // OS, sets its UID and groups to the specified `--uid`, `--gid` and
 // `--groups` and then launches the requested `--cmd`.
 func beIncubator(args []string) error {
+	// To defend against issues like https://golang.org/issue/1435,
+	// defensively lock our current goroutine's thread to the current
+	// system thread before we start making any UID/GID/group changes.
+	runtime.LockOSThread()
```
twitchyliquid64, line 210: Do we need to do this in an `init()`? And if this method is hit by tests, do we need to do a `defer runtime.UnlockOSThread()` to be good thread denizens?

twitchyliquid64 (approved):

ssh/tailssh/incubator.go
```diff
@@ -780,6 +785,20 @@ func setGroups(groupIDs []int) error {
...
+		// changes the egid of the process. This is technically a
+		// violation of the POSIX standard; see the following article
+		// for more detail:
+		//    https://www.usenix.org/system/files/login/articles/325-tsafrir.pdf
+		//
+		// In this case, we add an entry at the beginning of the
+		// groupIDs list containing the current egid, which replicates
+		// the behaviour on every other platform of just changing the
+		// supplementary groups and not the egid.
+		groupIDs = append([]int{os.Getegid()}, groupIDs...)
```
bradfitz, line 800: Do you want to filter out egid from the groupIDs?

crawshaw (commented):
LockOSThread is like grabbing a mutex, it should be unlocked carefully.

Is FreeBSD the only BSD with this behavior or does it apply to all of them?

ssh/tailssh/incubator.go
andrew-d: I don't think we can do this in an `init`, due to the way that the childproc stuff works in Tailscale. I did add a defer unlock though; even though this isn't called in tests, no reason not to 👍

ssh/tailssh/incubator.go
andrew-d: Surprisingly not; FreeBSD can legitimately return an array of `[]int{0, 0, 1234}` from `getgroups`, and happily accepts an array with duplicates passed to `setgroups`. The linked article also does not filter out egid in their example code.

andrew-d:
> LockOSThread is like grabbing a mutex, it should be unlocked carefully.

Done.

> Is FreeBSD the only BSD with this behavior or does it apply to all of them?

I don't believe this applies to OpenBSD, but @twitchyliquid64 found https://github.com/samba-team/samba/blob/master/source3/lib/system.c#L824-L837 which indicates that it applies to Darwin, and I've confirmed with a test program. I've pushed a change that applies the same fix there.

As for others ... both the [OpenBSD manpage](https://man.openbsd.org/setgid.2) and [NetBSD manpage](https://man.netbsd.org/setgid.2) for `setgid` have the following claim, which FreeBSD's manpage does not:
> Supplementary group IDs remain unchanged.

This implies to me that the egid isn't returned in the list from `getgroups`, though I haven't tested this myself.

andrew-d:
Did this in #7615 instead

PR #7613, opened 2023-03-20
funnel: change references from alpha to beta

shayne:
Updates CLI and docs to reference Funnel as beta


creachadair (commented):
LGTM (re: https://github.com/tailscale/corp/issues/10077)

maisem (commented):
Lgtm

PR #7615, opened 2023-03-20
ssh/tailssh: fix privilege dropping on FreeBSD; add tests

andrew-d:
On FreeBSD and Darwin, changing a process's supplementary groups with setgroups(2) will also change the egid of the process, setting it to the first entry in the provided list. This is distinct from the behaviour on other platforms (and possibly a violation of the POSIX standard).

Because of this, on FreeBSD with no TTY, our incubator code would previously not change the process's gid, because it would read the newly-changed egid, compare it against the expected egid, and since they matched, not change the gid. Because we didn't use the 'login' program on FreeBSD without a TTY, this would propagate to a child process.

This could be observed by running "id -p" in two contexts. The expected output, and the output returned when running from a SSH shell, is:

    andrew@freebsd:~ $ id -p
    uid         andrew
    groups      andrew

However, when run via "ssh andrew@freebsd id -p", the output would be:

    $ ssh andrew@freebsd id -p
    login       root
    uid         andrew
    rgid        wheel
    groups      andrew

(this could also be observed via "id -g -r" to print just the gid)

We fix this by pulling the details of privilege dropping out into their own function and prepending the expected gid to the start of the list on Darwin and FreeBSD.

Finally, we add some tests that run a child process, drop privileges, and assert that the final UID/GID/additional groups are what we expect.

More information can be found in the following article:
    https://www.usenix.org/system/files/login/articles/325-tsafrir.pdf

Updates #7616
Alternative to #7609 

Change-Id: I0e6513c31b121108b50fe561c89e5816d84a45b9

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -235,6 +235,7 @@ func beIncubator(args []string) error {
 	if err == nil && sessionCloser != nil {
 		defer sessionCloser()
 	}
+
 	var groupIDs []int
 	for _, g := range strings.Split(ia.groups, ",") {
 		gid, err := strconv.ParseInt(g, 10, 32)
@@ -244,22 +245,10 @@ func beIncubator(args []string) error {
 		groupIDs = append(groupIDs, int(gid))
 	}
 
-	if err := setGroups(groupIDs); err != nil {
+	if err := dropPrivileges(logf, int(ia.uid), ia.gid, groupIDs); err != nil {
 		return err
 	}
-	if egid := os.Getegid(); egid != ia.gid {
-		if err := syscall.Setgid(int(ia.gid)); err != nil {
-			logf(err.Error())
-			os.Exit(1)
-		}
-	}
-	if euid != ia.uid {
-		// Switch users if required before starting the desired process.
-		if err := syscall.Setuid(int(ia.uid)); err != nil {
-			logf(err.Error())
-			os.Exit(1)
-		}
-	}
+
 	if ia.isSFTP {
 		logf("handling sftp")
 
@@ -302,6 +291,108 @@ func beIncubator(args []string) error {
 		os.Exit(code)
 	}
 	return err
+}
+
+// TODO(andrew-d): verify that this works in more configurations before
+// enabling by default.
+const assertDropPrivileges = false
+
+// dropPrivileges contains all the logic for dropping privileges to a different
+// UID, GID, and set of supplementary groups. This function is
+// security-sensitive and ordering-dependent; please be very cautious if/when
+// refactoring.
+//
+// WARNING: if you change this function, you *MUST* run the TestDropPrivileges
+// test in this package as root on at least Linux, FreeBSD and Darwin. This can
+// be done by running:
+//
+//	go test -c ./ssh/tailssh/ && sudo ./tailssh.test -test.v -test.run TestDropPrivileges
+func dropPrivileges(logf logger.Logf, wantUid, wantGid int, supplementaryGroups []int) error {
+	fatalf := func(format string, args ...any) {
+		logf(format, args...)
+		os.Exit(1)
+	}
+
+	euid := os.Geteuid()
+	egid := os.Getegid()
+
+	if runtime.GOOS == "darwin" || runtime.GOOS == "freebsd" {
+		// On FreeBSD and Darwin, the first entry returned from the
+		// getgroups(2) syscall is the egid, and changing it with
+		// setgroups(2) changes the egid of the process. This is
+		// technically a violation of the POSIX standard; see the
+		// following article for more detail:
+		//    https://www.usenix.org/system/files/login/articles/325-tsafrir.pdf
+		//
+		// In this case, we add an entry at the beginning of the
+		// groupIDs list containing the expected gid if it's not
+		// already there, which modifies the egid and additional groups
+		// as one unit.
+		if len(supplementaryGroups) == 0 || supplementaryGroups[0] != wantGid {
+			supplementaryGroups = append([]int{wantGid}, supplementaryGroups...)
+		}
+	}
+
+	if err := setGroups(supplementaryGroups); err != nil {
+		return err
+	}
+	if egid != wantGid {
+		// On FreeBSD and Darwin, we may have already called the
+		// equivalent of setegid(wantGid) via the call to setGroups,
+		// above. However, per the manpage, setgid(getegid()) is an
+		// allowed operation regardless of privilege level.
+		//
+		// FreeBSD:
+		//	The setgid() system call is permitted if the specified ID
+		//	is equal to the real group ID or the effective group ID
+		//	of the process, or if the effective user ID is that of
+		//	the super user.
+		//
+		// Darwin:
+		//	The setgid() function is permitted if the effective
+		//	user ID is that of the super user, or if the specified
+		//	group ID is the same as the effective group ID.  If
+		//	not, but the specified group ID is the same as the real
+		//	group ID, setgid() will set the effective group ID to
+		//	the real group ID.
+		if err := syscall.Setgid(wantGid); err != nil {
+			fatalf("Setgid(%d): %v", wantGid, err)
+		}
+	}
+	if euid != wantUid {
+		// Switch users if required before starting the desired process.
+		if err := syscall.Setuid(wantUid); err != nil {
+			fatalf("Setuid(%d): %v", wantUid, err)
+		}
+	}
+
+	// If we changed either the UID or GID, defensively assert that we
+	// cannot reset the it back to our original values, and that the
+	// current egid/euid are the expected values after we change
+	// everything; if not, we exit the process.
+	if assertDropPrivileges {
+		if egid != wantGid {
+			if err := syscall.Setegid(egid); err == nil {
+				fatalf("unexpectedly able to set egid back to %d", egid)
+			}
+		}
+		if euid != wantUid {
+			if err := syscall.Seteuid(euid); err == nil {
+				fatalf("unexpectedly able to set euid back to %d", euid)
+			}
+		}
+
+		if got := os.Getegid(); got != wantGid {
+			fatalf("got egid=%d, want %d", got, wantGid)
+		}
+		if got := os.Geteuid(); got != wantUid {
+			fatalf("got euid=%d, want %d", got, wantUid)
+		}
+
+		// TODO(andrew-d): assert that our supplementary groups are correct
+	}
+
+	return nil
 }
 
 // launchProcess launches an incubator process for the provided session.
--- a/ssh/tailssh/privs_test.go
+++ b/ssh/tailssh/privs_test.go
@@ -1,0 +1,295 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+//go:build linux || darwin || freebsd || openbsd || netbsd || dragonfly
+
+package tailssh
+
+import (
+	"encoding/json"
+	"errors"
+	"os"
+	"os/exec"
+	"os/user"
+	"path/filepath"
+	"reflect"
+	"regexp"
+	"runtime"
+	"strconv"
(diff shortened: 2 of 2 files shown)


bradfitz:
(in general, anything security-related or anything that'll be cherry-picked _definitely_ needs a public bug)

ssh/tailssh/incubator.go
```diff
@@ -304,6 +293,82 @@ func beIncubator(args []string) error {
...
+		if err := syscall.Setuid(wantUid); err != nil {
+			fatalf("Setuid(%d): %v", wantUid, err)
+		}
+	}
+
+	// Defensively assert that we cannot reset the uid/gid back to our
+	// original values, and that the current egid/euid are the expected
+	// values after we change everything; if not, we exit the process.
+	if assertDropPrivileges {
+		if err := syscall.Setegid(egid); err == nil {
```
twitchyliquid64, line 355: If we are SSH'ing into a root account, would we expect this to work?

Or maybe `dropPrivileges()` is not called in that case?

andrew-d:
> This needs a public bug filed & referenced from the commit message.

Done; #7616

ssh/tailssh/incubator.go
andrew-d: Hm, I think it might; good thing this isn't turned on by default 🙈 Lemme update this to only assert if the UID/GID changed.

andrew-d: (done)

ssh/tailssh/incubator.go
```diff
@@ -304,6 +293,87 @@ func beIncubator(args []string) error {
...
+		}
+	}
+	if euid != wantUid {
+		// Switch users if required before starting the desired process.
+		if err := syscall.Setuid(wantUid); err != nil {
+			fatalf("Setuid(%d): %v", wantUid, err)
+		}
+	}
+
+	// If we changed either the UID or GID, efensively assert that we
```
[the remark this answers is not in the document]
andrew-d: This one is more complicated since we can technically fail to call `setgroups` but succeed otherwise, since `setgroups` requires root whereas `{set,get}uid` can work as a non-root user. I'll add a TODO for myself to come back and add something here, though.

crawshaw (approved):

ssh/tailssh/incubator.go
```diff
@@ -304,6 +293,87 @@ func beIncubator(args []string) error {
...
+	// cannot reset the it back to our original values, and that the
+	// current egid/euid are the expected values after we change
+	// everything; if not, we exit the process.
+	if assertDropPrivileges {
+		if egid != wantGid {
+			if err := syscall.Setegid(egid); err == nil {
```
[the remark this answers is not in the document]
andrew-d: This `setegid` (and the following `seteuid`) should fail, since we've dropped privileges in a way that shouldn't allow us to regain them. This verifies it by actually trying it, and failing if it succeeds (which would indicate that we didn't drop permissions correctly). This would have caught the FreeBSD bug because we would have been able to regain root permissions here unexpectedly.

ssh/tailssh/incubator.go
```diff
@@ -244,22 +245,10 @@ func beIncubator(args []string) error {
 		groupIDs = append(groupIDs, int(gid))
 	}
 
-	if err := setGroups(groupIDs); err != nil {
+	if err := dropPrivileges(logf, int(ia.uid), ia.gid, groupIDs); err != nil {
```
danderson, line 248: Hmm, why is ia.uid an int64 but gids are a regular int? Whenever I see a cast to int it sets off my integer overflow spidey sense. I wonder if we can standardize on whichever type the syscalls want, so that the parsing funcs take care of overflow handling? (_do_ they take care of overflow handling?)

ssh/tailssh/incubator.go
```diff
@@ -304,6 +293,89 @@ func beIncubator(args []string) error {
...
+		// On FreeBSD and Darwin, the first entry returned from the
+		// getgroups(2) syscall is the egid, and changing it with
+		// setgroups(2) changes the egid of the process. This is
+		// technically a violation of the POSIX standard; see the
+		// following article for more detail:
+		//    https://www.usenix.org/system/files/login/articles/325-tsafrir.pdf
+		//
+		// In this case, we add an entry at the beginning of the
+		// groupIDs list containing the expected gid if it's not
+		// already there, which modifies the egid and additional groups
```
danderson, line 329: "If it's not already there" - on compliant platforms, is it permitted to repeat the egid in the supplemental groups list? If not, I would defensively make that a fatal error, just to reduce the universe of possible states we have to deal with (and thus the universe of possible bugs stemming from different OSes's interpretation/violation of posix)

ssh/tailssh/incubator.go
```diff
@@ -304,6 +293,89 @@ func beIncubator(args []string) error {
...
+		// as one unit.
+		if len(supplementaryGroups) == 0 || supplementaryGroups[0] != wantGid {
+			supplementaryGroups = append([]int{wantGid}, supplementaryGroups...)
+		}
+	}
+
+	if err := setGroups(supplementaryGroups); err != nil {
+		return err
+	}
+	if egid != wantGid {
```
danderson, line 339: Won't this branch fail on darwin/freebsd after the setGroups call mutates the process's egid?

Oh, or is `Setgid(myCurrentEgid)` specified to be a no-op success? Add comment if so, given the large comment about freebsd/darwin's behavior above, by the time I get to here in the code I'm hyper-aware of how setGroups has mutated the egid and I would want to know that this branch has considered and validated that case.

ssh/tailssh/incubator.go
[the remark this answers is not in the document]
danderson: Noting that on linux at least, capabilites can be used to make this check misfire: CAP_SETUID and CAP_SETGID allow a process to change uid/gid in arbitrary ways regardless of the current uid/gid (in effect, it grants root's "setuid/gid to whatever you want" powers to non-root users).

However, I don't think we would ever want to run the incubator in a non-root setting with CAP_SETUID or CAP_SETGID, and this check will misfire-safe. IMO, that's fine.

ssh/tailssh/incubator.go
```diff
@@ -304,6 +293,89 @@ func beIncubator(args []string) error {
...
+	// If we changed either the UID or GID, defensively assert that we
+	// cannot reset the it back to our original values, and that the
+	// current egid/euid are the expected values after we change
+	// everything; if not, we exit the process.
+	if assertDropPrivileges {
+		if egid != wantGid {
+			if err := syscall.Setegid(egid); err == nil {
+				fatalf("unexpectedly able to set egid back to %d", egid)
+			}
+			if got := os.Getegid(); got != wantGid {
```
danderson, line 360: I think this check should be outside the `if egid != wantGid`. That is, unconditionally check that euid, egid and supplemental groups are exactly as we expect them to be, regardless of which branches we went down previously in the code.

Similarly, assuming the set calls are specified to no-op succeed if you set to the existing value, I would run all those sanity checks unconditionally. That way the checks are robust against a bug where euid == wantUid but the setting code above took the wrong path and did something wrong. At this point in the code, we just want to assert strongly "I think I've done everyting I need, does the uid/gid state _exactly_ match the intent I was given?"

ssh/tailssh/privs_test.go
```diff
@@ -0,0 +1,294 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		inputb, err := json.Marshal(input)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		cmd := exec.Command(os.Args[0], "-test.v", "-test.run", "^"+regexp.QuoteMeta(t.Name())+"$")
```
danderson, line 90: Alternatively: you could move the envvar check to `func TestMain()` and switch into the child logic before any of Go's test framework runs at all. No strong feelings from me, that may not be cleaner given other test files in this package.

ssh/tailssh/privs_test.go
```diff
@@ -0,0 +1,294 @@
...
+			gid:              gid1,
+			additionalGroups: []int{gid2, gid3},
+		},
+		{
+			name:             "no_additional_groups",
+			uid:              uid1,
+			gid:              gid1,
+			additionalGroups: []int{},
+		},
+		// Note: the below test case triggers the FreeBSD "setgroups
```
danderson, line 139: Link to the bug here, e.g. `Regression test for <bug link>`, that way you don't have to re-mention "the freebsd setgroups thing" because people can visit the bug for the gory details.

danderson (approved):

ssh/tailssh/incubator.go
andrew-d: @maisem - do you remember why? On every platform that I can find (Linux, Darwin, various BSDs), `uid_t` is 32-bit, and the Go stdlib takes the `int` type for both `os.Getuid` and the `syscall.{Get,Set}uid` functions.

ssh/tailssh/incubator.go
andrew-d: I tested both Darwin and FreeBSD, and it appears to be permitted to repeat the egid in the set of supplemental groups.  This behaviour also matches what Samba does [here](https://github.com/samba-team/samba/blob/master/source3/lib/system.c#L851-L863), which makes me a bit more confident that it's correct.

ssh/tailssh/incubator.go
andrew-d: Yep, that's permitted; I left a giant comment with manpage citations.

ssh/tailssh/incubator.go
andrew-d: Moved the checks outside the `if egid != wantGid`; checking unconditionally seems sensible.

The set calls are different, though; we're checking for failure, and setting them to the current value would succeed (and thus result in an abort); I think we need to guard these with the "only if changed" if statement.

andrew-d:
@danderson - Mind taking another look at my most recent changes?

ssh/tailssh/privs_test.go
andrew-d: Yeah, I didn't want to hijack the other behaviour in this file, especially since in the common case this won't run at all since people aren't running `go test` as root 🙈

ssh/tailssh/incubator.go
danderson: Okay, fair enough.

Incidentally, that Samba code might point to why this behaves differently on Darwin: the comment suggests that darwin ignores the first supplemental gid entirely. So, it's possible on darwin prior to this change we were actually dropping one of the supplemental gids?

ssh/tailssh/incubator.go
danderson: Yup, makes sense.

ssh/tailssh/incubator.go
```diff
@@ -304,6 +293,108 @@ func beIncubator(args []string) error {
...
+		}
+
+		if got := os.Getegid(); got != wantGid {
+			fatalf("got egid=%d, want %d", got, wantGid)
+		}
+		if got := os.Geteuid(); got != wantUid {
+			fatalf("got euid=%d, want %d", got, wantUid)
+		}
+
+		// TODO(andrew-d): assert that our supplementary groups are correct
```
danderson, line 392: Addressing this todo before submission? Or in a followup?

danderson (approved):
Looks good, thank you! Some minor thoughts but I don't need another round unless you explicitly would like one.

ssh/tailssh/incubator.go
andrew-d: Yeah, also that 😬 Yay for fixing bugs as well, I guess?

ssh/tailssh/incubator.go
andrew-d: Will address in a follow-up; I didn't want to make this too complicated to start with, and I need to test more (see other comment).

ssh/tailssh/incubator.go
maisem: I went back to find the history and couldn't this looks like an oversight.

PR #7618, opened 2023-03-20
tool/gocross: support running from outside the repo dir

danderson:


danderson:
Darn, CI is correct that my fix doesn't work in all cases. Back to the drawing board...

catzkorn (approved):

PR #7647, opened 2023-03-22
ssh/tailssh: use background context for uploading recordings

maisem:
Otherwise we see errors like
```
ssh-session(sess-20230322T005655-5562985593): recording: error sending recording to <addr>:80: Post "http://<addr>:80/record": context canceled
```

The ss.ctx is closed when the session closes, but we don't want to break the upload at that time. Instead we want to wait for the session to close the writer when it finishes, which it is already doing.

Updates tailscale/corp#9967

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -1343,7 +1343,12 @@ func (ss *sshSession) startNewRecording() (_ *recording, err error) {
 	}
 
 	pr, pw := io.Pipe()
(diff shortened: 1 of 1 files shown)


crawshaw (approved):

crawshaw:
You might want to turn your commit messages into a code comment

maisem:
> The ss.ctx is closed when the session closes, but we don't want to break the upload at that time. Instead we want to wait for the session to close the writer when it finishes, which it is already doing.

Good idea :) Done.

twitchyliquid64 (approved):

PR #7654, opened 2023-03-22
log: use logtail to log and upload sockstat logs

willnorris:
Switch to using logtail for logging sockstat logs. Always log locally (on supported platforms), but disable automatic uploading.  Change existing c2n sockstats request to trigger upload to log server and return log ID.

The main thing left to do here is figure out which logid to use.  @dsnet and I talked about this briefly the other day, but didn't arrive at a conclusive answer.  We can't use the main logid for the device because we're logging into a different collection, and that can cause issues.  We could derive the ID from the main logid (maybe `sha256(logid + "sockstats")` or something?)  It needs to be easily reproducible so we can get the sockstats for given node ID, but I don't know that it's worth the effort to plumb it all the way through control like we do for other IDs.  wdyt?

### DIFF
--- a/ipn/ipnlocal/c2n.go
+++ b/ipn/ipnlocal/c2n.go
@@ -84,8 +84,13 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
 		}
 		writeJSON(res)
 	case "/sockstats":
+		if r.Method != "POST" {
+			http.Error(w, "bad method", http.StatusMethodNotAllowed)
+			return
+		}
 		w.Header().Set("Content-Type", "text/plain")
-		b.sockstatLogger.WriteLogs(w)
+		b.sockstatLogger.Flush()
+		fmt.Fprintln(w, b.sockstatLogger.LogID())
 	default:
 		http.Error(w, "unknown c2n path", http.StatusBadRequest)
 	}
--- a/log/sockstatlog/logger.go
+++ b/log/sockstatlog/logger.go
@@ -6,15 +6,21 @@ package sockstatlog
 
 import (
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"io"
+	"net/http"
 	"os"
 	"path/filepath"
 	"time"
 
+	"tailscale.com/logpolicy"
+	"tailscale.com/logtail"
 	"tailscale.com/logtail/filch"
 	"tailscale.com/net/sockstats"
+	"tailscale.com/smallzstd"
 	"tailscale.com/types/logger"
+	"tailscale.com/types/logid"
 	"tailscale.com/util/mak"
 )
 
@@ -26,9 +32,10 @@ type Logger struct {
 	ctx      context.Context
 	cancelFn context.CancelFunc
 
-	ticker    *time.Ticker
-	logf      logger.Logf
-	logbuffer *filch.Filch
+	ticker *time.Ticker
+	logf   logger.Logf
+
+	logger *logtail.Logger
 }
 
 // deltaStat represents the bytes transferred during a time period.
@@ -50,10 +57,18 @@ type event struct {
 	Stats map[sockstats.Label]deltaStat `json:"s"`
 }
 
+// SockstatLogID reproducibly derives a new logid.PrivateID for sockstat logging from a node's public backend log ID.
+// The returned PrivateID is the sha256 sum of id + "sockstat".
+// If a node's public log ID becomes known, it is trivial to spoof sockstat logs for that node.
+// Given the this is just for debugging, we're not too concerned about that.
+func SockstatLogID(id string) logid.PrivateID {
+	return logid.PrivateID(sha256.Sum256([]byte(id + "sockstat")))
+}
+
 // NewLogger returns a new Logger that will store stats in logdir.
 // On platforms that do not support sockstat logging, a nil Logger will be returned.
 // The returned Logger must be shut down with Shutdown when it is no longer needed.
-func NewLogger(logdir string, logf logger.Logf) (*Logger, error) {
+func NewLogger(logdir string, logf logger.Logf, backendLogID string) (*Logger, error) {
 	if !sockstats.IsAvailable {
 		return nil, nil
 	}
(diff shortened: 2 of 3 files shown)


dsnet:
If the private log ID is:
* `sha256(logid.PrivateID + "sockstats")` then we won't know what the public log ID is until the node tells control (perhaps through a `tailcfg.Hostinfo` update?
* `sha256(logid.PublicID + "sockstats")` then we can derive the public log ID from our side without the node telling us anything. However, this means that it is trivial for anyone to spoof statistics for this given log ID so long as they know the public log ID for a particular node. Since this is for statistics gathering and performance testing, perhaps we don't care about spoofing.

willnorris:
oh, are the private log IDs completely private to the client?  I sort of assumed (I'm not sure why) that we knew them on control as well.  But I see now in logstore where we are switching them to the public ID.

dsnet:
Yea, private log IDs are sent to us, but we never store it. We use it to derive the public log ID, which is the ID used to read a logstream.

One way of thinking about it is that:
* the private log ID is used for writing to a log stream, while
* the public log ID is used for reading the log stream.

willnorris:
okay, so it sounds like options are:

1. derive ID on control and hand it to the client (what netlog does)
2. derive ID on client and inform control via hostinfo update
3. derive ID from public logID so it can be done on either side, and accept that someone could theoretically spoof stats for a different node (if they knew there public ID).

I think the practical risk of #3 is sufficiently low that I'm not terribly concerned about it and am leaning toward that.  I worry a little more about setting that precedent that adding other logs of this type in the future begins to litter HostInfo or netmap.SelfNode with a ton of separate log IDs.


@mihaip wdyt?

dsnet:
Approach 4:
* Hard-code a universal private log that all nodes use for this purpose. Log messages from each node self-identifies what node it came from. Thus, reading the sock stats for all nodes only requires reading from a single public log ID.  Similar to approach 3, messages can be trivially spoofed.

willnorris:
yeah, that could work too.  Though it makes it a lot harder to see stats for a single node over time since you have to parse them out of all the other logs.  I assume the log server can do this much more efficiently for a unique logID?  I've implemented number 3, so will go ahead and push that for folks to take a look at.

log/sockstatlog/logger.go
```diff
@@ -50,10 +57,16 @@ type event struct {
 	Stats map[sockstats.Label]deltaStat `json:"s"`
 }
 
+// sockstatLogID reproducibly derives a new logid.PrivateID to be used for sockstat logging.
+// The returned PrivateID is the sha256 sum of id + "sockstat".
+func sockstatLogID(id string) logid.PrivateID {
```
dsnet, line 62: It's kinda gross that `id` is not type-safe, when I would expect this to be `logid.PublicID`. Unfortunately, the callers above this carry this variable around as a `string`. Perhaps, we should switch them all to specify `logid.PublicID`? Up to you whether to clean this up.

log/sockstatlog/logger.go
```diff
@@ -50,10 +57,16 @@ type event struct {
 	Stats map[sockstats.Label]deltaStat `json:"s"`
 }
 
+// sockstatLogID reproducibly derives a new logid.PrivateID to be used for sockstat logging.
```
dsnet, line 60: I would place the caveats about spoofing somewhere and why we're not worried about it. Perhaps here?

log/sockstatlog/logger.go
```diff
@@ -69,12 +82,31 @@ func NewLogger(logdir string, logf logger.Logf) (*Logger, error) {
...
+		NewZstdEncoder: func() logtail.Encoder {
+			w, err := smallzstd.NewEncoder(nil)
+			if err != nil {
+				panic(err)
+			}
+			return w
+		},
+		FlushDelayFn: func() time.Duration {
+			// set flush delay to 100 years so it never flushes automatically
+			return 100 * 365 * 24 * time.Hour
```
dsnet, line 104: 100 years ought to be long enough 😆

dsnet (approved):

dsnet:
I'm not in a position to adequately evaluate the risks of spoofing for what you are you trying to accomplish. So long as you are okay with the possibility of spoofed log messages, then the approach LGTM.

log/sockstatlog/logger.go
dsnet: Also, should this be exported so that we can call this logic from elsewhere?

ipn/ipnlocal/c2n.go
```diff
@@ -85,7 +85,8 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
 		writeJSON(res)
 	case "/sockstats":
 		w.Header().Set("Content-Type", "text/plain")
-		b.sockstatLogger.WriteLogs(w)
+		b.sockstatLogger.Flush()
```
mihaip, line 88: Since this has a side effects it should probably be a POST-only method, similar to `/logtail/flush`

mihaip (approved):
I don't have big concerns about spoofing - this is all for debugging use and is not load bearing. The worst case is that somehow a malicious actor gets a public log ID for one of our devices and uses it to spam its logs, but in that case we can reset the client's log ID (by removing the keychain entry).

As a variant of 3, you may want to consider deriving this log ID from the node ID or other field that already exists, to avoid the confusion of the co-mingling of "public log ID is used to generate a private one".

willnorris:
> As a variant of 3, you may want to consider deriving this log ID from the node ID or other field that already exists, to avoid the confusion of the co-mingling of "public log ID is used to generate a private one".

Yeah, I believe that is basically what we do for netlog, though it happens control-side.  Sounds good.

(for the moment, I'm tracking down these test failures which seem legitimate, but indicate some possible shenanigans happening in our integration tests)

log/sockstatlog/logger.go
willnorris: switched this to StableNodeID at Mihai's suggestion.  And yes, making it exported.  I'll come back and look at cleaning up the type safety on these other log IDs after.

dsnet: There is minor benefit of using public log ID over StableNodeID. The StableNodeID is broadcasted to other nodes in the tailnet, while the public log ID (to my recollection) is not as readily known by other nodes. This only changes the degree of spoofability (again, if we care).

willnorris: yeah, I suppose it's also a lot less disruptive to roll a new log ID if someone does start spoofing traffic than it would be to roll a new stable ID.  I think that's good enough reason to switch it back.

Issue #7616, opened 2023-03-20
Tailscale SSH doesn't drop privileges correctly on FreeBSD when running a non-shell command


andrew-d:
### What is the issue?

On FreeBSD, we don't use the `login` command when running a command without a shell/TTY:
https://github.com/tailscale/tailscale/blob/654b5a0616f235eb41ccb31fc0c4499b1d952a26/ssh/tailssh/incubator.go#L762-L767

Additionally, FreeBSD (and Darwin) exhibit non-POSIX-compliant behaviour in their `getgroups()` and `setgroups()` system calls: the first entry in the list is returned and used to set the process's egid, as if it were returned by `getegid` and changed by `setegid`.

As a result, our existing behaviour for dropping privileges has a bug:
https://github.com/tailscale/tailscale/blob/654b5a0616f235eb41ccb31fc0c4499b1d952a26/ssh/tailssh/incubator.go#L247-L262

For example, if we call our incubator with `--uid=1002 --gid=1002 --groups=1002`, it will:
* Set the new supplemental groups to `[]int{1002}` with `setGroups`, which changes the egid to 1002
* Read the current now-changed egid, which is 1002, and compare against the expected egid of 1002; nothing changes, so it does not call `setgid`.
* Check the current UID and call `setuid(1002)`.

As a result, the egid has changed to 1002, but *the gid remains 0* (assuming Tailscale is running as root).

This doesn't affect other OSes because:
* On Darwin, we always use the `login` command, which resets groups for us
* On non-FreeBSD platforms, the `setgroups` command doesn't change the egid, which means that the comparison against egid succeeds and we call `setgid` to reset the GID.

See also:
* Samba's workaround for this problem: https://github.com/samba-team/samba/blob/master/source3/lib/system.c#L824-L837
* An article containing more details: https://www.usenix.org/system/files/login/articles/325-tsafrir.pdf

### Steps to reproduce

This could be observed by running "id -p" in two contexts. The expected output, and the output returned when running from a Tailscale SSH shell, is:
```
andrew@freebsd:~ $ id -p
uid         andrew
groups      andrew
```

However, when run via "ssh andrew@freebsd id -p", the output would be:
```
$ ssh andrew@freebsd id -p
login       root
uid         andrew
rgid        wheel
groups      andrew
```

### Are there any recent changes that introduced the issue?

n/a

### OS

Other

### OS version

FreeBSD

### Tailscale version

1.38.1

DentonGentry:
[...]

PR #7659, opened 2023-03-23
ssh/tailssh: always assert our final uid/gid

andrew-d:
Move the assertions about our post-privilege-drop UID/GID out of the conditional if statement and always run them; I haven't been able to find a case where this would fail. Defensively add an envknob to disable this feature, however, which we can remove after the 1.40 release.

Updates #7616

Change-Id: Iaec3dba9248131920204bd6c6d34bbc57a148185

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -293,9 +293,19 @@ func beIncubator(args []string) error {
 	return err
 }
 
-// TODO(andrew-d): verify that this works in more configurations before
-// enabling by default.
-const assertDropPrivileges = false
+const (
+	// This controls whether we assert that our privileges were dropped
+	// using geteuid/getegid; it's a const and not an envknob because the
+	// incubator doesn't see the parent's environment.
+	//
+	// TODO(andrew): remove this const and always do this after sufficient
+	// testing, e.g. the 1.40 release
+	assertPrivilegesWereDropped = true
+
+	// TODO(andrew-d): verify that this works in more configurations before
+	// enabling by default.
+	assertPrivilegesWereDroppedByAttemptingToUnDrop = false
+)
 
 // dropPrivileges contains all the logic for dropping privileges to a different
 // UID, GID, and set of supplementary groups. This function is
@@ -309,7 +319,7 @@ const assertDropPrivileges = false
 //	go test -c ./ssh/tailssh/ && sudo ./tailssh.test -test.v -test.run TestDropPrivileges
 func dropPrivileges(logf logger.Logf, wantUid, wantGid int, supplementaryGroups []int) error {
 	fatalf := func(format string, args ...any) {
-		logf(format, args...)
+		logf("[unexpected] error dropping privileges: "+format, args...)
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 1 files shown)


ssh/tailssh/incubator.go
```diff
@@ -295,7 +295,10 @@ func beIncubator(args []string) error {
 
 // TODO(andrew-d): verify that this works in more configurations before
 // enabling by default.
-const assertDropPrivileges = false
+const checkUnDropPrivileges = false
```
maisem, line 298: `assertPrivilegesWereDroppedByAttemptingToUnDrop`?

ssh/tailssh/incubator.go
```diff
@@ -370,26 +373,26 @@ func dropPrivileges(logf logger.Logf, wantUid, wantGid int, supplementaryGroups
...
 		if euid != wantUid {
 			if err := syscall.Seteuid(euid); err == nil {
-				fatalf("unexpectedly able to set euid back to %d", euid)
+				fatalf("able to set euid back to %d", euid)
 			}
 		}
-
+	}
+	// TODO(andrew): remove this envknob after sufficient testing, e.g. the 1.40 release
+	if !envknob.BoolDefaultTrue("TS_DEBUG_DISABLE_ASSERT_PRIVILEGES") {
```
maisem, line 389: there are too many negations here. Whats the intent?

andrew-d: Heh, enough that I actually screwed it up 🤦 Lemme fix.

andrew-d: Okay, removed a bunch of negations; PTAL?

ssh/tailssh/incubator.go
```diff
@@ -370,26 +373,26 @@ func dropPrivileges(logf logger.Logf, wantUid, wantGid int, supplementaryGroups
...
+	// TODO(andrew): remove this envknob after sufficient testing, e.g. the 1.40 release
+	if envknob.BoolDefaultTrue("TS_DEBUG_INCUBATOR_ASSERT_PRIVILEGES") {
 		if got := os.Getegid(); got != wantGid {
 			fatalf("got egid=%d, want %d", got, wantGid)
 		}
 		if got := os.Geteuid(); got != wantUid {
 			fatalf("got euid=%d, want %d", got, wantUid)
 		}
-
-		// TODO(andrew-d): assert that our supplementary groups are correct
```
maisem, line 392: we don't wanna do this anymore?

maisem (approved):

ssh/tailssh/incubator.go
```diff
@@ -370,26 +373,26 @@ func dropPrivileges(logf logger.Logf, wantUid, wantGid int, supplementaryGroups
...
 		if euid != wantUid {
 			if err := syscall.Seteuid(euid); err == nil {
-				fatalf("unexpectedly able to set euid back to %d", euid)
+				fatalf("able to set euid back to %d", euid)
 			}
 		}
-
+	}
+	// TODO(andrew): remove this envknob after sufficient testing, e.g. the 1.40 release
+	if envknob.BoolDefaultTrue("TS_DEBUG_INCUBATOR_ASSERT_PRIVILEGES") {
```
maisem, line 389: can you make sure that this is never forwarded from the SSH session?
Actually, maybe lets not have an envknob here at all.

andrew-d: I could just unset the variable unconditionally after this `if`?

maisem: the incubator doesn't get any of the `os.Environ`. So you'll have to build a custom binary anyways, so a const should work?

andrew-d: Hm, yeah, a `const` is probably fine then

andrew-d: Pushed; PTAL?

maisem (approved):
Thanks!

danderson (approved):

PR #7660, opened 2023-03-23
ssh/tailssh: lock OS thread during incubator

andrew-d:
This makes it less likely that we trip over bugs like golang/go#1435.

Updates #7616

Change-Id: Ic28c03c3ad8ed5274a795c766b767fa876029f0e

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -204,6 +204,16 @@ func parseIncubatorArgs(args []string) (a incubatorArgs) {
 // OS, sets its UID and groups to the specified `--uid`, `--gid` and
 // `--groups` and then launches the requested `--cmd`.
 func beIncubator(args []string) error {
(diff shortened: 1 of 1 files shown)


ssh/tailssh/incubator.go
```diff
@@ -204,6 +204,12 @@ func parseIncubatorArgs(args []string) (a incubatorArgs) {
 // OS, sets its UID and groups to the specified `--uid`, `--gid` and
 // `--groups` and then launches the requested `--cmd`.
 func beIncubator(args []string) error {
+	// To defend against issues like https://golang.org/issue/1435,
```
bradfitz, line 207: Maybe say how this shouldn't matter on Linux but might help other platforms?

andrew-d: @bradfitz - added a bit more comment; PTAL?

maisem (approved):

bradfitz (approved):

PR #7661, opened 2023-03-23
all: pass log IDs as the proper type rather than strings

willnorris:
This change focuses on the backend log ID, which is the mostly commonly used in the client.  Tests which don't seem to make use of the log ID just use the zero value.


dsnet (approved):
Nice. Code-health++

PR #7663, opened 2023-03-23
ipn/store/kubestore: handle "/" in ipn.StateKeys

### LINKED ISSUE #7662: tsnet: ListenTLS/ListenFunnel/Up fail on Kubernetes
This is because we are trying to `nil` out the existing ServeConfig in `(*tsnet.Server).Up`. This inadvertently tries to write to the StateStore with a key like `_serve/<profileID>` however k8s does not allow slashes in the key

maisem:
Kubernetes doesn't allow slashes as keys in secrets, replace them with "__".

This shows up in the kubernetes-operator now that tsnet sets resets the ServeConfig at startup.

Fixes #7662

### DIFF
--- a/ipn/store/kubestore/store_kube.go
+++ b/ipn/store/kubestore/store_kube.go
@@ -7,6 +7,7 @@ package kubestore
 
 import (
 	"context"
+	"strings"
 	"time"
 
 	"tailscale.com/ipn"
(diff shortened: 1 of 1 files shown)


ipn/store/kubestore/store_kube.go
```diff
@@ -46,13 +47,21 @@ func (s *Store) ReadState(id ipn.StateKey) ([]byte, error) {
...
+	b, ok := secret.Data[sanitizeKey(id)]
 	if !ok {
 		return nil, ipn.ErrStateNotExist
 	}
 	return b, nil
 }
 
+func sanitizeKey(k ipn.StateKey) string {
+	id := string(k)
+	if strings.Contains(id, "/") {
```
crawshaw, line 59: Code is easier to read without the if statement, do you need it for performance?

why __? Changing the length is unfortunate and doesn’t actually escape to remove conflicts.

maisem: picked `__` arbitrarily, went with `_` and more stricter parsing.

crawshaw (approved):

shayne:
A similar escape was needed for macsys, addressed in tailscale/corp#8142. Slashes preexisted `serve` [here](https://github.com/tailscale/tailscale/blob/57a008a1e17438ffd7b30cd8d8f16e4097ab131f/ipn/store.go#L55-L64) but from when I looked at it back then, we don't hit the `/` path as it never came up before `serve`.

PR #7666, opened 2023-03-23
tka: implement filesystem chonk garbage collection

twitchyliquid64:


tka/tailchonk.go
```diff
@@ -380,6 +380,104 @@ func (c *FS) AllAUMs() ([]AUMHash, error) {
...
 }
 
+// CollectGarbage frees up disk space by removing purged AUMs
+// and files which contain no data.
+func (c *FS) CollectGarbage(maxAge time.Duration) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// Collect the list of all stored hashes which are marked
+	// for deletion & old enough to delete.
```
crawshaw, line 390: not quite? you collect candidates with `info.AUM == nil && len(info.Children) == 0` regardless of their age.

did you mean to only ever consider older hashes?

crawshaw (approved):
This feels like a very aggressive algorithm, specifically cleaning up brand new hashes with no AUM or children. I think that makes sense, but it's also not something I would expect to happen?

PR #7669, opened 2023-03-23
ipn/ipnlocal: fix cert storage in Kubernetes

### LINKED ISSUE #7588: Tailscale 1.38.1 not caching certificate in Kubernetes state secret
### What is the issue?

Tailscale isn't storing certificate in Kubernetes state following upgrade to 1.38.1. I understand this feature is now in there. Original issue: https://github.com/tailscale/tailscale/issues/5676

I removed my homebrew caching mechanism and triggered a cert request via the Caddy sidecar and went to check the state and I can't see anything that looks like a certificate. I can see a new cert in /tmp/ though. Maybe I've missed a parameter?

Secret:

```
 kubectl describe -n mynamespace secret tailscale-state
Name:         tailscale-state
Namespace:    mynamespace
Labels:       <none>
Annotations:  <none>

Type:  Opaque

Data
====
_current-profile:  12 bytes
_daemon:           817 bytes
_machinekey:       72 bytes
_nl-node-key:      135 bytes
_profiles:         331 bytes
device_fqdn:       35 bytes
device_id:         12 bytes
profile-577c:      1176 bytes
```

Sidecar params: 

```
  ts-sidecar:
    Container ID:   containerd://e9f802263677a529d918073f900d3d53a3c7551eeb6ce293086ebe7a1e848502
    Image:          ghcr.io/tailscale/tailscale:v1.38.1
    Image ID:       ghcr.io/tailscale/tailscale@sha256:f2be570c3a890c4f70bcf4dcbba2c92e1b9ea38bbfb453f0b0113f9f102b1b78
 ...
    Environment:
      TS_KUBE_SECRET:  tailscale-state
      TS_USERSPACE:    true
      TS_EXTRA_ARGS:   --hostname myhostname
      TS_AUTH_KEY:     <set to the key 'TS_AUTH_KEY' in secret 'tailscale-auth'>  Optional: false
```

### Steps to reproduce

Install Tailscale into Kubernetes and trigger a certificate request.

Updating to version 1.38.1

### OS

Linux

### OS version

-

### Tailscale version

1.38.1

### Other software

Kubernetes

### Bug report

_No response_

### LINKED ISSUE #7665: cmd/k8s-operator: store certs in k8s state store
Currently tsnet does not cache certs in the Kubernetes Secret. In a crashlooping pod, this can result in LetsEncrypt backoff periods where we can't get a new certificate for 168 hours. This specifically impacts the AuthProxy.

maisem:
We were checking against the wrong directory, instead if we
have a custom store configured just use that.

Fixes #7588
Fixes #7665


dcarrion87:
I'm not sure as to all the times it needs to go off to the kube store to check certificate but is there any performance impact on just relying on kube secret directly? If there is some level of caching happening in memory I guess that's fine.

maisem:
> I'm not sure as to all the times it needs to go off to the kube store to check certificate but is there any performance impact on just relying on kube secret directly? If there is some level of caching happening in memory I guess that's fine.

Thats a good point in general for the kubestore, I think it should do some caching currently it doesn't. I'll track that separately though.

ipn/ipnlocal/cert.go
```diff
@@ -140,15 +142,32 @@ type certStore interface {
...
+	WriteACMEKey([]byte) error
 }
 
 var errCertExpired = errors.New("cert expired")
 
-func (b *LocalBackend) getCertStore(dir string) certStore {
-	if hostinfo.GetEnvType() == hostinfo.Kubernetes && dir == "/tmp" {
-		return certStateStore{StateStore: b.store}
+func (b *LocalBackend) getCertStore() (certStore, error) {
+	switch (b.store).(type) {
```
creachadair, line 155: I don't think the first set of parenthesis is necessary here.

ipn/ipnlocal/cert.go
```diff
@@ -160,6 +179,25 @@ type certFileStore struct {
...
+			return nil, ipn.ErrStateNotExist
+		}
+		return nil, err
+	}
+	return v, nil
+}
+
+func (f certFileStore) WriteACMEKey(b []byte) error {
+	pemName := filepath.Join(f.dir, acmePEMName)
+	return os.WriteFile(pemName, b, 0600)
```
creachadair, line 198: Should we use `atomicfile` here?

creachadair (approved):

crawshaw (approved):

ipn/ipnlocal/cert.go
maisem: Updated all three usages of `os.WriteFile` to use `atomicfile` as its own commit. PTAL.

ipn/ipnlocal/cert.go
maisem: removed

creachadair (approved):

dcarrion87:
Hi team, is this likely to make it into a 1.38.X release in the next week?

DentonGentry:
I expect to build a 1.38.3 in the next week.
I've cherry-picked the three commits in this PR to the release branch in preparation.

PR #7674, opened 2023-03-24
log/sockstatlog: make shutdown close idle connections

maisem:
Updates tailscale/corp#10030


mihaip (approved):
LGTM

Should probably find a way to test this directly in OSS instead of relying on corp pulls.

maisem:
Sent https://github.com/tailscale/tailscale/pull/7675

PR #7678, opened 2023-03-24
cmd/tailscale: extend hostname length checks

knyar:
In addition to checking the total length, enforce the 63-char limit on the first label. Use limits documented in RFC 1035.

Updates https://github.com/tailscale/corp/issues/10012

### DIFF
--- a/cmd/tailscale/cli/cli_test.go
+++ b/cmd/tailscale/cli/cli_test.go
@@ -621,9 +621,16 @@ func TestPrefsFromUpArgs(t *testing.T) {
 		{
 			name: "error_long_hostname",
 			args: upArgsT{
-				hostname: strings.Repeat("a", 300),
+				hostname: strings.Repeat(strings.Repeat("a", 63)+".", 5),
 			},
-			wantErr: `hostname too long: 300 bytes (max 256)`,
+			wantErr: `hostname too long: 320 bytes (max 255)`,
+		},
+		{
+			name: "error_long_label",
+			args: upArgsT{
+				hostname: strings.Repeat("a", 64) + ".example.com",
+			},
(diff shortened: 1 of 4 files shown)


cmd/tailscale/cli/up.go
```diff
@@ -320,8 +321,13 @@ func prefsFromUpArgs(upArgs upArgsT, warnf logger.Logf, st *ipnstate.Status, goo
 		}
 	}
 
-	if len(upArgs.hostname) > 256 {
-		return nil, fmt.Errorf("hostname too long: %d bytes (max 256)", len(upArgs.hostname))
+	// RFC 1035 says the length of a DNS name is restricted to 255 octets or less.
+	if len(upArgs.hostname) > 255 {
```
sailorfrag, line 325: I might suggest moving the length checks to `dnsname` so that it can use the existing constants there.

cmd/tailscale/cli/cli_test.go
```diff
@@ -621,9 +621,16 @@ func TestPrefsFromUpArgs(t *testing.T) {
 		{
 			name: "error_long_hostname",
 			args: upArgsT{
-				hostname: strings.Repeat("a", 300),
+				hostname: strings.Repeat(strings.Repeat("a", 63)+".", 5),
 			},
-			wantErr: `hostname too long: 300 bytes (max 256)`,
+			wantErr: `hostname too long: 320 bytes (max 255)`,
```
sailorfrag, line 626: additional test cases for the boundary may be good to verify no off-by-one errors.

sailorfrag (approved):

sailorfrag (approved):
Seems good.

It's going to reject the `SRV` record format of `_service._proto.name.` but that's probably fine since this is intended for hostnames not service records.

PR #7693, opened 2023-03-25
ssh/tailssh: add tests for recording failure

maisem:
Updates tailscale/corp#9967


ssh/tailssh/tailssh.go
```diff
@@ -1425,6 +1425,7 @@ func (ss *sshSession) startNewRecording() (_ *recording, err error) {
 		pw.Close()
 		return nil, err
 	}
+	req.Header.Set("Expect", "100-continue")
```
crawshaw, line 1428: why? please add comment

maisem: added a comment

crawshaw (approved):

Issue #3560, opened 2021-12-13
derp: track flows internally


bradfitz:
The derp server should internally track flows (tuple `(pubKeySrc, pubKeyDst)`) with stats on some new flow type.

@JulianKnodt did most of this in https://github.com/tailscale/tailscale/pull/2531 but his co-op term ended before it got finished, and then it's bitrot in the meantime.

Tracking bug to get that finished up.

PR #7697, opened 2023-03-27
derp: Track flows internally

valscale:
The derp server should track stats on flows between src/dst public keys and make them available in some way. This is a bare bones initial pass that does the absolute minimum and only for locally connected destinations, intended for discussion only.

Updates #3560

### DIFF
--- a/derp/derp_server.go
+++ b/derp/derp_server.go
@@ -26,6 +26,7 @@ import (
 	"net/netip"
 	"os/exec"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -87,6 +88,26 @@ const (
 	// then disables all of them.
 	disableFighters
 )
+
+// flowStats is a data structure containing stats for a client to
+// client data flow through this server. The counters are written
+// solely by run(), read by the debug server, and do not need to be
+// thread-safe.
+type flowStats struct {
+	bytes   int64
+	packets int64
+	// TODO: replace with a server thread that sweeps every N time
+	// units
+	lastActive time.Time
+}
+
+// uniqueConn is used to uniquely index flows by source, since the
+// source public key can be duplicated by multiple connected
+// clients. The connection number is used to disambiguate them.
+type uniqueConn struct {
+	key     key.NodePublic
+	connNum int64
+}
 
 type align64 [0]atomic.Int64 // for side effect of its 64-bit alignment
 
@@ -162,6 +183,19 @@ type Server struct {
 	// src.
 	sentTo map[key.NodePublic]map[key.NodePublic]int64 // src => dst => dst's latest sclient.connNum
 
+	// flows tracks statistics for the various client flows
+	// through this server. It includes flows that have a src or
+	// dst that is not a local client. Since client keys may be
+	// duplicated, we distinguish between flows to different
+	// duplicate clients using the connection number, like sentTo.
+	//
+	// The src sclient has a map with references to all the flows
+	// for which it is the source (indexed by dst key), so our
+	// first level map index is by dst key to be able to quickly
+	// delete the flows for which this sclient is the destination.
+	//
+	// TODO: this seems like it should obviously be merged with sentTo[][]
+	flows map[key.NodePublic]map[uniqueConn]*flowStats // dst => src,connNum => flow stats
 	// maps from netip.AddrPort to a client's public key
 	keyOfAddr map[netip.AddrPort]key.NodePublic
 }
@@ -318,6 +352,7 @@ func NewServer(privateKey key.NodePrivate, logf logger.Logf) *Server {
 		avgQueueDuration:     new(uint64),
 		tcpRtt:               metrics.LabelMap{Label: "le"},
 		keyOfAddr:            map[netip.AddrPort]key.NodePublic{},
+		flows:                map[key.NodePublic]map[uniqueConn]*flowStats{},
 	}
 	s.initMetacert()
 	s.packetsRecvDisco = s.packetsRecvByKind.Get("disco")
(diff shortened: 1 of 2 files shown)


derp/derp_server.go
```diff
@@ -157,15 +158,35 @@ type Server struct {
...
 	clientsMesh map[key.NodePublic]PacketForwarder
 	// sentTo tracks which peers have sent to which other peers,
-	// and at which connection number. This isn't on sclient
-	// because it includes intra-region forwarded packets as the
-	// src.
-	sentTo map[key.NodePublic]map[key.NodePublic]int64 // src => dst => dst's latest sclient.connNum
+	// at which connection number, and how many packets/bytes were
+	// sent per connection. This isn't on sclient because it
+	// includes intra-region forwarded packets as the src.
+	sentTo map[key.NodePublic]map[key.NodePublic]flowData
```
bradfitz, line 164: I would expect the map key to be a comparable struct value type being a 2-tuple of two NodePublics. Then make the value be a pointer type so you can update it cheaply without doing map operations all the time.

```go
type FlowKey struct {
    Src, Dst key.NodePublic
}

...

    flows map[FlowKey]*flowState
```

derp/derp_server.go
```diff
@@ -157,15 +158,35 @@ type Server struct {
...
 
 	// maps from netip.AddrPort to a client's public key
 	keyOfAddr map[netip.AddrPort]key.NodePublic
 }
 
+// Per connection statistics for observability of who is using the
+// server and why, including intra-region forwarded packets, kept in
+// the sentTo map in the server. Accessed only under s.mu.
+type flowData struct {
```
bradfitz, line 173: Go doc style starts with the subject (https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences)

So:

// flowData contains the ...

or

// flowData is the ...

derp/derp_server.go
```diff
@@ -157,15 +158,35 @@ type Server struct {
...
+	// TODO: why is connNum signed?
+	connNum int64  // unique number per src->dest connection
+	pkts    uint64 // total packets sent on this connection
+	bytes   uint64 // total bytes
+}
+
+// updateFlowStats tracks how much data is flowing between a src and
```
bradfitz, line 180: missing "Locked"

derp/derp_server.go
```diff
@@ -157,15 +158,35 @@ type Server struct {
...
+// dst client pair. Currently it only records for local destination
+// clients. Only call when holding S.mu and the key always exists.
```
bradfitz, line 182: I'd do the invariants on its own line:

```
// Caller must hold s.mu.
```
or
```
// s.mu must be held.
```

derp/derp_server.go
```diff
@@ -1931,6 +1958,43 @@ func (s *Server) ServeDebugTraffic(w http.ResponseWriter, r *http.Request) {
...
+// communicating directly but still using this DERP server as their
+// home.
+//
+// TODO: What else do we want to know? Time connection was
+// established? Time since last packet? Average bandwidth?
+//
+// TODO: Convert into Prometheus histograms
+func (s *Server) ServeDebugFlows(w http.ResponseWriter, r *http.Request) {
+	// Lock, then copy out the current map, then unlock
+	fmt.Fprintf(w, "src dst bytes packets\n")
```
bradfitz, line 1972: I'd just go straight to JSON

derp/derp_server.go
```diff
@@ -1931,6 +1958,43 @@ func (s *Server) ServeDebugTraffic(w http.ResponseWriter, r *http.Request) {
...
+	s.mu.Unlock()
+
+	// Sort by largest flow
+	sort.Slice(flows, func(i, j int) bool { return flows[i].flw.bytes < flows[j].flw.bytes })
+
+	for _, f := range flows {
+		fmt.Fprintf(w, "%v %v %v %v\n", f.src.ShortString(), f.dst.ShortString(), f.flw.bytes, f.flw.pkts)
+	}
+
+	w.(http.Flusher).Flush()
```
bradfitz, line 1995: not needed. It happens on exit. Plus you're assuming the ResponseWriter is a Flusher here.

derp/derp_server.go
```diff
@@ -157,15 +158,35 @@ type Server struct {
...
 
+// Per connection statistics for observability of who is using the
+// server and why, including intra-region forwarded packets, kept in
+// the sentTo map in the server. Accessed only under s.mu.
+type flowData struct {
+	// TODO: why is connNum signed?
+	connNum int64  // unique number per src->dest connection
+	pkts    uint64 // total packets sent on this connection
+	bytes   uint64 // total bytes
+}
```
bradfitz, line 178: you'll also want to track a last seen time, so you can clean up old flows not seen in N minutes

bradfitz:
Nit: lowercase verb after colon in commit message

valscale:
This is an extremely minimal proof of concept to have a concrete discussion on things like:

* What stats do we eventually want to export in what form? The plan is to start using Prometheus sparse histograms to track overall distributions of things, but do we also want to keep extremely detailed stats on every single flow exposed via /debug? Do we want stats on a sliding time window or just since the beginning of the connection? How does this interface with exporting to Prometheus?

* There are a number of data structures kept by the server that are partial maps of src, dst tuples with some info in them, should we mash them in to one full map with all per-tuple info in it? E.g., this patch extends the sentTo[src key][dst key] structure, created to track when we need to send peer gone notices, to track bytes and packets as well. However it only tracks flows to local destination clients. We could create an entirely separate map of all (src, dst) tuples and keep our stats in there but it's a lot of duplication. And any per-server thing has to be updated under s.mu().

PR #7699, opened 2023-03-27
tsnet,wgenegine/netstack: add test and fix resource leaks

maisem:
We were not closing the http.Server and were also not waiting for netstack to fully close.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -117,6 +117,7 @@ type Server struct {
 	loopbackListener net.Listener           // optional loopback for localapi and proxies
 	localAPIListener net.Listener           // in-memory, used by localClient
 	localClient      *tailscale.LocalClient // in-memory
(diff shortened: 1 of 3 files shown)


tsnet/tsnet.go
```diff
@@ -566,8 +574,9 @@ func (s *Server) start() (reterr error) {
 	lal := memnet.Listen("local-tailscaled.sock:80")
```
crawshaw, line 574: does anything here need to be cleaned up? I would've expected closing this to shut down the server

crawshaw (approved):

tsnet/tsnet.go
maisem: Reading the docs on `(net.Listener).Close` suggests that closing existing conns is not a part of it. It only stops new connections from being established. So if there is a persistent albeit idle connection open, we need to call `server.Shutdown` to close them.

PR #7704, opened 2023-03-27
log: allow toggling sockstat logs via c2n component logging

willnorris:
Once we're confident about this, we can change ipnlocal to construct the sockstat logger on all clients, but then only enable it by default on unstable builds.

Updates tailscale/corp#9230

### DIFF
--- a/log/sockstatlog/logger.go
+++ b/log/sockstatlog/logger.go
@@ -12,6 +12,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"sync/atomic"
 	"time"
 
 	"tailscale.com/logpolicy"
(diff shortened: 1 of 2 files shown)


log/sockstatlog/logger.go
```diff
@@ -114,11 +114,24 @@ func NewLogger(logdir string, logf logger.Logf, logID logid.PublicID) (*Logger,
...
-
 	return logger, nil
 }
 
+// SetLoggingEnabled enables or disables logging.
+// When disabled, socket stats are not polled and no new logs are written to disk.
+// Existing logs can still be fetched via the C2N API.
+func (l *Logger) SetLoggingEnabled(v bool) {
+	old := l.enabled.Load()
+	if l.enabled.CompareAndSwap(old, v) {
```
mihaip, line 125: Seems like you should also check that `v != old`, otherwise you'll start another poll if someone calls `SetLoggingEnabled` twice.

willnorris: oops, never mind :)  forget to reset old on each of those tests

willnorris: done

mihaip (approved):

PR #7705, opened 2023-03-27
util/codegen: add -copyright to control presence of copyright headers

### LINKED ISSUE #7702: FR: add a no-copyright flag to cloner

I'd like to use cloner in other projects. It currently inserts a header asserting copyright of the generated code by Tailscale. That's not ideal.

Mind if I add a no-copyright flag that suppresses that header?

### How should we solve this?

_No response_

_No response_

### Anything else?

_No response_

josharian:
I took the simple approach of declaring a flag directly in the codegen package, rather than threading it through (as a bool param, yuck) in two places.

I also opted for a bool rather than anything more complicated: The header contains a copyright holder and a license indicator, and trying to model all of that in flags was going to get ugly. (Multiple flags? Read from a file?)

I'm happy to alter either of those choices if you prefer, though.

Fixes #7702

Oh and btw, looks like the fallback dns server file could use regeneration.

### DIFF
--- a/util/codegen/codegen.go
+++ b/util/codegen/codegen.go
@@ -6,6 +6,7 @@ package codegen
 
 import (
 	"bytes"
+	"flag"
 	"fmt"
 	"go/ast"
 	"go/token"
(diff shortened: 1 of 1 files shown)


crawshaw (approved):

josharian:
(An alternative, flag-free approach is to toggle the copyright header based on the package path; if it starts with tailscale.com / tailscale.io, then leave in the header.)

willnorris (approved):

crawshaw:
filed the CI flake here: https://github.com/tailscale/tailscale/issues/7707

PR #7711, opened 2023-03-28
cmd/tsconnect: export IPNNetMap types in ts package

twitchyliquid64:
No idea if I'm doing this right?

### DIFF
--- a/cmd/tsconnect/src/pkg/pkg.ts
+++ b/cmd/tsconnect/src/pkg/pkg.ts
@@ -38,3 +38,5 @@ export async function createIPN(config: IPNPackageConfig): Promise<IPN> {
 }
 
 export { runSSHSession } from "../lib/ssh"
+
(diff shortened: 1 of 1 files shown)


twitchyliquid64:
Sweet - think that worked

mihaip:
CI passed, but did it actually export the type in the `pkg.d.ts` file? I don't see it when I patch in your change.

Will play around with this, but I think implementing the alternate suggestion from https://github.com/tailscale/corp/pull/10110#discussion_r1151047339 might be the better overall solution.

mihaip:
Actually, I don't think you need this change at all. The type definition already says:

```typescript
declare global {
  ...
  type IPNNetMap = { ... }
}
```

So it's already known in the global scope. You should be able to use it without any `import` on the corp repo side.

twitchyliquid64:
Gotcha - will try that in the other PR!

PR #7716, opened 2023-03-29
go.mod: bump gvisor to 20230320 for dispatcher locking

### LINKED ISSUE #7715: gvisor shutdown sometimes causes a panic due to race on endpoint dispatcher
### What is the issue?

See panic caught in CI:

https://github.com/tailscale/tailscale/actions/runs/4548299564/jobs/8019187385#step:7:843

Related updates in upstream: google/gvisor@13d7bf69d8068a0b1939e94bdb9ff08b1cdad913

We must avoid calling into `ep.Inject*` after we have started to shut down. The change in 920ec69241930c352818b0bf3eab20e62df93ed1 fixed an issue where we were not waiting for netstack to complete shutdown, however this made it likely that we will encounter this race, as netstack asynchronously sets the endpoint dispatcher to nil, at which point subsequent packets that get sent to the endpoint via ep.Inject* will trigger an NPE.

raggi:
Upstream improved code around an issue showing up in CI, where sometimes
shutdown will race on endpoint.dispatcher being nil'd, causing a panic
down stack of injectInbound. The upstream patch makes some usage more
safe, but it does not itself fix the local issue.

See panic in https://github.com/tailscale/tailscale/actions/runs/4548299564/jobs/8019187385#step:7:843

See fix in https://github.com/google/gvisor/commit/13d7bf69d8068a0b1939e94bdb9ff08b1cdad913

Updates https://github.com/tailscale/tailscale/issues/7715


maisem:
The panic is still showing up.

raggi:
> The panic is still showing up.

yup, looking, will likely reword and otherwise look at a fix

raggi:
@maisem I'd still like to land this, so we avoid the correctness problem showing up somewhere else.

maisem (approved):
Change commit message to not say this is the fix.

But yeah lgtm on merging this

raggi:
> Change commit message to not say this is the fix.

Sorry yeah, it was just the PR UI not being in sync.

PR #7720, opened 2023-03-29
wgengine/netstack: do not send packets to netstack after close

### LINKED ISSUE #7715: gvisor shutdown sometimes causes a panic due to race on endpoint dispatcher

raggi:
Use the local context on Impl to check for shut down state in order to drop rather than inject packets after close has begun.

Netstack sets endpoint.dispatcher to nil during shutdown. After the recent adjustment in 920ec69241930c352818b0bf3eab20e62df93ed1 we now wait for netstack to fully shutdown before we release tests. This means that we may continue to accept packets and attempt to inject them, which we must prevent in order to avoid nil pointer panic.

References google/gvisor#8765
Fixes #7715

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -394,6 +394,10 @@ func (ns *Impl) updateIPs(nm *netmap.NetworkMap) {
 // the host and arriving at tailscaled. This method returns filter.DropSilently
 // to intercept a packet for handling, for instance traffic to quad-100.
 func (ns *Impl) handleLocalPackets(p *packet.Parsed, t *tstun.Wrapper) filter.Response {
+	if ns.ctx.Err() != nil {
+		return filter.DropSilently
+	}
+
 	// If it's not traffic to the service IP (i.e. magicDNS) we don't
 	// care; resume processing.
 	if dst := p.Dst.Addr(); dst != magicDNSIP && dst != magicDNSIPv6 {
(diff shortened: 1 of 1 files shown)


raggi:
I'm not super keen on hitting context here. It is currently relatively cheap as there is no context chain, the cancellation context is created locally, but it would be easy to accidentally regress on this and start doing a walk.

maisem (approved):

bradfitz:
> I'm not super keen on hitting context here. It is currently relatively cheap as there is no context chain

Relative to having a context chain, I guess.

But it's still two atomic write operations: the Mutex.Lock and the Unlock.

We could have it just be an atomic read to check a "closing atomic.Bool" on the netstack.Impl?

PR #7723, opened 2023-03-29
log/sockstatlog: add delay before writing logs to disk

### LINKED ISSUE #7719: uploading sockstat logs results in infinite upload loop
Whenever triggering upload of sockstat logs, uploads continue seemingly forever.  This can be seen at http://go/logs/BUG-276722f7c2d07cfe13681d9d8181db6b0a6149efa4acc70337e6b8ee8de3dbd7-20230329014701Z-adfa92486dd47764

```
01:44:44 | {"c":1,"d":99,"s":{"3":[672,97]},"t":1680054284508}
01:44:44 | {"c":1,"d":100,"s":{"3":[289,194]},"t":1680054284608}
01:44:44 | {"c":1,"d":100,"s":{"3":[291,97]},"t":1680054284808}
01:44:45 | {"c":1,"d":100,"s":{"3":[290,97]},"t":1680054284908}
01:44:45 | {"c":1,"d":99,"s":{"3":[290,97]},"t":1680054285108}
01:44:45 | {"c":1,"d":100,"s":{"3":[289,97]},"t":1680054285208}
01:44:45 | {"c":1,"d":100,"s":{"3":[289,97]},"t":1680054285308}
01:44:45 | {"c":1,"d":99,"s":{"3":[290,97]},"t":1680054285608}
01:44:45 | {"c":1,"d":100,"s":{"3":[288,97]},"t":1680054285708}
01:44:45 | {"c":1,"d":99,"s":{"3":[290,97]},"t":1680054285808}
```

The constant stream of these small events with label 3 ([LogtailLogger](https://github.com/tailscale/tailscale/blob/840d69e1ebfc39c6a58ae38a7bf4aae321e7720a/net/sockstats/sockstats.go#L44)) indicates the problem. Logtail reads from the local logs one line at a time until it can't read anymore, uploading them in batches to logcatcher.  This upload itself results in network traffic which is itself captured by sockstat logs.  The events are written to disk more frequently (10hz) than they can be uploaded to logcatcher, so logtail never finishes reading from the log file.

This problem wasn't observed during development because I was uploading to a dev logcatcher on localhost, and it WAS able to upload fast enough that it got ahead of the sockstat poll.

One simple solution (which I've coded up, and will submit shortly) is to delay writing events to disk, giving the logtail uploader time to upload and finish before the sockstats for that upload are captured and written back out to disk.

willnorris:
Split apart polling of sockstats and logging them to disk.  Add a 3 second delay before writing logs to disk to prevent an infinite upload loop when uploading stats to logcatcher.  3 seconds is admittedly somewhat arbitrary, but seems to be long enough to address the problem, but not too long to worry about the memory usage of the buffered channel.

Fixes #7719

### DIFF
--- a/log/sockstatlog/logger.go
+++ b/log/sockstatlog/logger.go
@@ -1,7 +1,9 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
-// Package sockstatlog provides a logger for capturing and storing network socket stats.
+// Package sockstatlog provides a logger for capturing network socket stats for debugging.
+// Stats are collected at a frequency of 10 Hz and logged to disk.
+// Stats are only uploaded to the log server on demand.
 package sockstatlog
 
 import (
@@ -25,19 +27,29 @@ import (
 	"tailscale.com/util/mak"
 )
 
-// pollPeriod specifies how often to poll for socket stats.
-const pollPeriod = time.Second / 10
+// pollInterval specifies how often to poll for socket stats.
+const pollInterval = time.Second / 10
+
+// logInterval specifies how often to log sockstat events to disk.
+// This delay is added to prevent an infinite loop when logs are uploaded,
+// which itself creates additional sockstat events.
+const logInterval = 3 * time.Second
 
 // Logger logs statistics about network sockets.
 type Logger struct {
+	// enabled identifies whether the logger is enabled.
 	enabled atomic.Bool
 
 	ctx      context.Context
 	cancelFn context.CancelFunc
 
-	ticker *time.Ticker
-	logf   logger.Logf
+	// eventCh is used to pass events from the poller to the logger.
+	eventCh chan event
 
+	logf logger.Logf
+
+	// logger is the underlying logtail logger than manages log files on disk
+	// and uploading to the log server.
 	logger *logtail.Logger
 	filch  *filch.Filch
 	tr     http.RoundTripper
@@ -73,6 +85,7 @@ func SockstatLogID(logID logid.PublicID) logid.PrivateID {
 // NewLogger returns a new Logger that will store stats in logdir.
 // On platforms that do not support sockstat logging, a nil Logger will be returned.
 // The returned Logger is not yet enabled, and must be shut down with Shutdown when it is no longer needed.
+// Logs will be uploaded to the log server using a new log ID derived from the provided backend logID.
 func NewLogger(logdir string, logf logger.Logf, logID logid.PublicID) (*Logger, error) {
 	if !sockstats.IsAvailable {
 		return nil, nil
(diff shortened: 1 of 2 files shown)


log/sockstatlog/logger.go
```diff
@@ -137,19 +155,19 @@ func (l *Logger) Write(p []byte) (int, error) {
...
 	// last is the last set of socket stats we saw.
 	var lastStats *sockstats.SockStats
 	var lastTime time.Time
 
-	enc := json.NewEncoder(l)
+	ticker := time.NewTicker(pollInterval)
 	for {
 		select {
 		case <-l.ctx.Done():
 			return
```
mihaip, line 169: Should we call `ticker.Stop()` here (since we don't have a `Stop()` call in `Shutdown()` anymore)? The docs for `NewTicker` says "Stop the ticker to release associated resources."

log/sockstatlog/logger.go
```diff
@@ -173,6 +189,33 @@ func (l *Logger) poll() {
 	}
 }
 
+// logEvents reads events from the event channel at logInterval and logs them to disk.
+// This method does not return.
+func (l *Logger) logEvents() {
+	enc := json.NewEncoder(l)
+	flush := func() {
```
mihaip, line 196: Any reason why is in a separate function? The `poll()` ticker handling is inline in the `select`.

log/sockstatlog/logger.go
```diff
@@ -137,19 +155,19 @@ func (l *Logger) Write(p []byte) (int, error) {
 }
 
 // poll fetches the current socket stats at the configured time interval,
-// calculates the delta since the last poll, and logs any non-zero values.
+// calculates the delta since the last poll, and writes any non-zero values to the logger event channel.
```
mihaip, line 158: Nit: seems like this should wrap at ~80 characters, like the previous line break.

mihaip (approved):
May want to make the `sockstats.Label` used by logtail configurable, so that we can have a separate label for these logs. It might make it easier to identify such infinite loops in the future.

log/sockstatlog/logger.go
willnorris: It was just for readability.  I originally had it inline, and it had nested for loops that I need to break out of when the event channel was empty.  I thought this was a little easier to reason about what was happening (and less error prone).

log/sockstatlog/logger.go
willnorris: yeah, good call.  Added to both poll and logEvent, since they each maintain their own tickers.

logtail/logtail.go
```diff
@@ -220,6 +222,11 @@ func (l *Logger) SetLinkMonitor(lm *monitor.Mon) {
 	l.linkMonitor = lm
 }
 
+// SetSockstatsLabel sets the label used in sockstat logs to identify network traffic from this logger.
```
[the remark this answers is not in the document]
willnorris: I originally put it in the Config struct, but the zero value of `sockstats.Label` is meaningful, so then I had to make it a pointer, and then you can't take the address of a constant for the default value, and then I just stopped fighting it and added the method.  🤷🏻 

And no, it's not currently being called.  I was sure if you were suggesting that we should go ahead and add a new label for sockstat logs now, or just have this available as a debugging tool in the future.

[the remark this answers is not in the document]
willnorris: If you don't mind bumping `sockstats.Label` so that it has a real zero value, that would certainly make this easier.  I just wasn't sure how you felt about changing that at this point.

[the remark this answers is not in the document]
mihaip: Renumbering seems awkward given that there already are clients that are uploading metrics with this -- sorry for not leaving 0 from the initial set of values.

I see what you mean about the pointer value not working  (would it be too gross to wrap it in an array?), so this function seems fine.

I still think we should use it in `sockstatlog` (and `netlog`) now, as opposed to wishing we had it when debugging in the future.

PR #7724, opened 2023-03-29
doctor/permissions: add new check to print process permissions

andrew-d:
Since users can run tailscaled in a variety of ways (root, non-root, non-root with process capabilities on Linux), this check will print the current process permissions to the log to aid in debugging.

Change-Id: Ida93a206123f98271a0c664775d0baba98b330c7


doctor/permissions/permissions.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"golang.org/x/exp/constraints"
+	"tailscale.com/types/logger"
+)
+
+// Check implements the doctor.Check interface.
+type Check struct{}
+
+func (Check) Name() string {
+	return "routetable"
```
maisem, line 22: permissions?

doctor/permissions/permissions.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (Check) Run(_ context.Context, logf logger.Logf) error {
+	return permissionsImpl(logf)
+}
+
+func formatUserId[T constraints.Integer](id T) string {
```
maisem, line 29: `ID` here and below

maisem (approved):

PR #7727, opened 2023-03-29
net/tstun,wgengine/*: add support for NAT to routes

maisem:
This adds support to make exit nodes and subnet routers work
when in scenarios where NAT is required.

It also updates the NATConfig to be generated from a `wgcfg.Config` as
that handles merging prefs with the netmap, so it has the required information
about whether an exit node is already configured and whether routes are accepted.

Updates tailscale/corp#8020


net/tstun/table/table.go
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package table
```
crawshaw, line 4: package docs

crawshaw (approved):

net/tstun/table/table.go
```diff
@@ -0,0 +1,58 @@
...
+	"tailscale.com/util/mak"
+)
+
+// RoutingTable provides a mapping from IP addresses to peers identified by
+// their public node key. It is used to find the peer that should be used to
+// route a given IP address.
+//
+// It is not safe for concurrent use.
+type RoutingTable struct {
+	peers      map[key.NodePublic]*device.Peer
```
twitchyliquid64, line 22: Maybe add a comment about why we need peers and that its not the actual routing table?

net/tstun/table/table.go
```diff
@@ -0,0 +1,58 @@
...
+	allowedIPs device.AllowedIPs
```
twitchyliquid64, line 23: Better name plz, took a few takes to figure out what this was meant to be

net/tstun/wrap.go
```diff
@@ -522,7 +523,9 @@ type natV4Config struct {
 	// dstMasqAddrs is map of dst addresses to their respective MasqueradeAsIP
 	// addresses. The MasqueradeAsIP address is the address that should be used
 	// as the source address for packets to dst.
-	dstMasqAddrs views.Map[netip.Addr, netip.Addr] // dst -> masqAddr
+	dstMasqAddrs views.Map[key.NodePublic, netip.Addr] // dst -> masqAddr
+
+	routingTable *table.RoutingTable
```
twitchyliquid64, line 528: Add a comment describing the key and value, as well as saying which ip (native ip, masq ip etc) it stores as value

twitchyliquid64, line 528: Could this be named more functionally than routing table? Which direction is this storing the routes for? (in vs out?) etc

twitchyliquid64: Maybe `nodeNativeIPTable` or something

net/tstun/table/table.go
maisem: I split it into a RoutingTableBuilder and a RoutingTable, so that its clear that we don't use this field at runtime. Also added docs.

net/tstun/table/table.go
maisem: went with prefixTrie and added docs

net/tstun/wrap.go
maisem: renamed to `dstAddrToPeerKeyMapper`

twitchyliquid64 (approved):

Issue #7626, opened 2023-03-20
Potential spurious login needed notifications when Windows GUI connects to tailscaled


dblohm7:
In the fast user switching world, then tailscaled receives its initial connection from the Windows GUI, it sets the SID of the Windows user and ends up doing a `(*LocalBackend).Start({})`. The state machine ends up in `ipn.NeedsLogin` even if there are no prefs. At the very least, this may trigger spurious notifications to the GUI.

I traced the state change to [this line](https://github.com/tailscale/tailscale/blob/e7a78bc28f93eec24d92c7ad7acee2b111b71cad/ipn/ipnlocal/local.go#L3625).

I don't think that predicate makes sense unless `wantRunning` is true. We probably got away with it before, but now we're resetting the client in ways that have us reaching that state even when nothing is supposed to even be attempting to log in to control.

dblohm7:
EOD for me, but tomorrow I'm going to validate the fix in the frontend to backend migration case and review unit tests.

dblohm7:
I've got a better way of dealing with this in the short term that I am hoping will solve Glencore as well:

There's no reason why `tailscaled` cannot migrate Windows profiles like it does on other platforms, albeit differently.

The profile manager already receives notice of the currently connected user. `tailscaled` is already running as LocalSystem, which for all intents and purposes is "God mode." Rather than relying on the client to send us the source prefs to migrate, the profile manager should try to do so in `tailscaled` as soon as it receives a `SetCurrentUserID` call. This will eliminate the boundary case where `LocalClient` thinks it's starting on a new machine because it hasn't yet received migration prefs via a the options passed into a `Start` call.

PR #7734, opened 2023-03-29
ipn/ipnlocal, util/winutil/policy: modify Windows profile migration t…

dblohm7:
…o load legacy prefs from within tailscaled

I realized that a lot of the problems that we're seeing around migration and LocalBackend state can be avoided if we drive Windows pref migration entirely from within tailscaled. By doing it this way, tailscaled can automatically perform the migration as soon as the connection with the client frontend is established.

Since tailscaled is already running as LocalSystem, it already has access to the user's local AppData directory. The profile manager already knows which user is connected, so we simply need to resolve the user's prefs file and read it from there.

Of course, to properly migrate this information we need to also check system policies. I moved a bunch of policy resolution code out of the GUI and into a new package in util/winutil/policy.

Updates #7626


ipn/ipnlocal/profiles.go
```diff
@@ -66,7 +65,13 @@ func (pm *profileManager) SetCurrentUserID(uid ipn.WindowsUserID) error {
 	// the selected profile for the current user.
 	b, err := pm.store.ReadState(ipn.CurrentProfileKey(string(uid)))
 	if err == ipn.ErrStateNotExist || len(b) == 0 {
-		pm.NewProfile()
+		if runtime.GOOS == "windows" {
+			if err := pm.migrateFromLegacyPrefs(); err != nil {
```
maisem, line 69: I think you need to handle `errAlreadyMigrated`?

ipn/ipnlocal/profiles_windows.go
```diff
@@ -0,0 +1,89 @@
...
+		return "", err
+	}
+	if usr.HomeDir == "" {
+		return "", fmt.Errorf("user %q does not have a home directory", uid)
+	}
+	userLegacyPrefsDir := filepath.Join(usr.HomeDir, "AppData", "Local", "Tailscale")
+	return userLegacyPrefsDir, nil
+}
+
+type ctxKeyMigrationSentinelPath struct{}
```
maisem, line 42: I don't understand why we need a context here. just return the path the sentinel file instead?

dblohm7: Sure.

ipn/ipnlocal/profiles.go
dblohm7: @maisem I guess the question is, what can we actually do to handle it at this point?

By reaching this point, we've already established that the current profile key is not yielding any profile info, and yet the migration code is telling us that we've already migrated. Given the inconsistency of that state, I'm not sure what else we can do other than propagate that message up the stack. Thoughts?

ipn/ipnlocal/profiles_windows.go
```diff
@@ -0,0 +1,84 @@
...
+		return "", err
+	}
+	if usr.HomeDir == "" {
+		return "", fmt.Errorf("user %q does not have a home directory", uid)
+	}
+	userLegacyPrefsDir := filepath.Join(usr.HomeDir, "AppData", "Local", "Tailscale")
+	return userLegacyPrefsDir, nil
+}
+
+func (pm *profileManager) loadLegacyPrefs() (string, ipn.PrefsView, error) {
```
maisem, line 41: document that the returned string is the `migrationSentinel` file name

ipn/ipnlocal/profiles.go
maisem: good point, i guess it should error out

maisem (approved):

Issue #7695, opened 2023-03-26
tailscale exit nodes do not work in Kubernetes


uhthomas:
### What is the issue?

It looks like Tailscale forgets that it needs to use the local cluster DNS resolver to communicate with the Kubernetes API and then times out and crashes.

```
2023/03/26 16:12:38 [RATELIMIT] format("open-conn-track: timeout opening %v to node %v; online=%v, lastRecv=%v")
2023/03/26 16:12:39 open-conn-track: flow TCP (TCP 100.97.169.251:52157 => 149.57.16.58:4934) got RST by peer
53579 => 95.102.251.18:41837) to node [iabZZ]; online=yes, lastRecv=0s
2023/03/26 16:12:48 open-conn-track: timeout opening (TCP 100.97.169.251:51851 => 82.155.153.203:48400) to node [iabZZ]; online=yes, lastRecv=0s
2023/03/26 16:12:48 [RATELIMIT] format("open-conn-track: timeout opening %v to node %v; online=%v, lastRecv=%v")
2023/03/26 16:12:49 [RATELIMIT] format("open-conn-track: flow TCP %v got RST by peer") (47 dropped)
2023/03/26 16:12:49 open-conn-track: flow TCP (TCP 100.97.169.251:36395 => 5.225.83.238:1900) got RST by peer
2023/03/26 16:12:49 open-conn-track: flow TCP (TCP 100.97.169.251:47147 => 156.146.49.174:40608) got RST by peer
2023/03/26 16:12:49 [RATELIMIT] format("open-conn-track: flow TCP %v got RST by peer")
2023/03/26 16:12:58 [RATELIMIT] format("open-conn-track: timeout opening %v to node %v; online=%v, lastRecv=%v") (206 dropped)
2023/03/26 16:12:58 open-conn-track: timeout opening (TCP 100.97.169.251:59757 => 185.229.59.115:12658) to node [iabZZ]; online=yes, lastRecv=0s
2023/03/26 16:12:58 open-conn-track: timeout opening (TCP 100.97.169.251:35647 => 41.216.203.148:47278) to node [iabZZ]; online=yes, lastRecv=0s
2023/03/26 16:12:58 [RATELIMIT] format("open-conn-track: timeout opening %v to node %v; online=%v, lastRecv=%v")
2023/03/26 16:12:59 [RATELIMIT] format("open-conn-track: flow TCP %v got RST by peer") (21 dropped)
[...]

PR #7735, opened 2023-03-29
ipn/store: add support for stores to hook into a custom dialer

maisem:
For stores like k8s secrets we need to dial out to the k8s API as though Tailscale wasn't running. The issue currently only manifests when you try to use an exit node while running inside a k8s cluster and are trying to use Kubernetes secrets as the backing store.

This doesn't address cmd/containerboot, which I'll do in a follow up.

Updates #7695

cc @uhthomas


uhthomas (approved):
Looks good, thank you! :smiley:

andrew-d (approved):
✅

Issue #3707, opened 2022-01-11
add mechanism for macOS users to set IPNExtension environment variables


bradfitz:
Like b03170b901b1bb89f815e589750a76ea4a2254bf did for  #3688, we should have the same for macOS GUI users.

DentonGentry:
Progress in https://github.com/tailscale/tailscale/pull/5665

reteps:
waiting on this! https://github.com/tailscale/tailscale/blob/main/envknob/envknob.go#L398

PR #7744, opened 2023-03-31
derphttp_client: add debug envknob to use HTTP client connections

valscale:
This is a POC for making DERP development easier by allowing clients to connect with HTTP, thereby avoiding the need for acquiring a valid TLS cert. This version uses an envknob and hardcodes in the port number the DERP server uses in "-dev" mode, while acquiring the DERP server hostname from the control server. To use this:

1. run derper with -dev flag

2. edit the ACL in the admin console to add a custom DERP map with your dev server's hostname

3. run tailscaled with TS_DEBUG_DERP_USE_HTTP=1

I would prefer to have something akin to "-login-server" for local control server development, which would instead look like:

1. run derper with -dev flag

2. run tailscaled with -derp-server=localhost

And default to the 3340 port.

Thoughts?

### DIFF
--- a/derp/derphttp/derphttp_client.go
+++ b/derp/derphttp/derphttp_client.go
@@ -185,6 +185,18 @@ func (c *Client) useHTTPS() bool {
 	if c.url != nil && c.url.Scheme == "http" {
 		return false
 	}
+	// Turn on development mode and connect to DERP using plain HTTP.
+	// Questions:
+	// * Should this be a command line option or an envknob?
+	// * If envknob, should it be stubbed out for iOS, etc.?
+	// * Should it be a port number instead of a boolean?
+	// * Should we be able to set the DERP hostname via envknob
+	//   instead of creating a custom derp map in the admin console?
+	// * Is there a better name?
+	if envknob.Bool("TS_DEBUG_DERP_USE_HTTP") {
+		return false
+	}
+
 	return true
 }
 
@@ -200,7 +212,11 @@ func (c *Client) urlString(node *tailcfg.DERPNode) string {
 	if c.url != nil {
 		return c.url.String()
 	}
-	return fmt.Sprintf("https://%s/derp", node.HostName)
+	proto := "https"
+	if envknob.Bool("TS_DEBUG_DERP_USE_HTTP") {
+		proto = "http"
+	}
+	return fmt.Sprintf("%s://%s/derp", proto, node.HostName)
 }
 
 // AddressFamilySelector decides whether IPv6 is preferred for
@@ -644,7 +660,18 @@ func (c *Client) dialNode(ctx context.Context, n *tailcfg.DERPNode) (net.Conn, e
 			}
 			port := "443"
 			if n.DERPPort != 0 {
+				// Question: should the port set in
+				// the tailcfg take precedence or
+				// should the debug envknob? Currently
+				// the tailcfg takes precedence but
+				// that's probably backwards.
 				port = fmt.Sprint(n.DERPPort)
+			} else if envknob.Bool("TS_DEBUG_DERP_USE_HTTP") {
+				// DERP listens on this port with -dev flag
+				// Questions:
+				// * Should this be a shared constant?
+				// * Should the vale of the envknob be the port number?
+				port = "3340"
 			}
 			c, err := c.dialContext(ctx, proto, net.JoinHostPort(dst, port))
 			select {
@@ -725,6 +752,7 @@ func (c *Client) dialNodeUsingProxy(ctx context.Context, n *tailcfg.DERPNode, pr
 		}
 	}()
 
+	// TODO: test TS_DEBUG_DERP_USE_HTTP with http proxies
 	target := net.JoinHostPort(n.HostName, "443")
 
 	var authHeader string


derp/derphttp/derphttp_client.go
```diff
@@ -185,6 +185,18 @@ func (c *Client) useHTTPS() bool {
 	if c.url != nil && c.url.Scheme == "http" {
 		return false
 	}
+	// Turn on development mode and connect to DERP using plain HTTP.
+	// Questions:
+	// * Should this be a command line option or an envknob?
```
mihaip, line 190: I think envknob is a better fit, since this is for development/testing only, and not something that we'd want users to use (both because it's less secure and because we want to be able to change it at any time). It matches the intent of the package's Go doc: https://github.com/tailscale/tailscale/blob/02582083d54f20e63796a518888aad68ace20b52/envknob/envknob.go#L4-L10

derp/derphttp/derphttp_client.go
```diff
@@ -185,6 +185,18 @@ func (c *Client) useHTTPS() bool {
...
+	// * If envknob, should it be stubbed out for iOS, etc.?
```
mihaip, line 191: Not sure we need to worry about iOS on a per-knob basis. #3707 is open about allowing envknobs to be configured on platforms where it's harder to set environment variables, when we get around to that this knob will also become settable there.

derp/derphttp/derphttp_client.go
```diff
@@ -185,6 +185,18 @@ func (c *Client) useHTTPS() bool {
...
+	// * Should it be a port number instead of a boolean?
```
mihaip, line 192: My preference would be to have a simpler setting so that there's less chances for a developer to get into a broken state, but I haven't done enough DERP-related development to have an opinion as to whether more flexibility is needed (would you want to run multiple derpers locally?).

derp/derphttp/derphttp_client.go
```diff
@@ -644,7 +660,18 @@ func (c *Client) dialNode(ctx context.Context, n *tailcfg.DERPNode) (net.Conn, e
 			}
 			port := "443"
 			if n.DERPPort != 0 {
+				// Question: should the port set in
+				// the tailcfg take precedence or
+				// should the debug envknob? Currently
+				// the tailcfg takes precedence but
+				// that's probably backwards.
 				port = fmt.Sprint(n.DERPPort)
+			} else if envknob.Bool("TS_DEBUG_DERP_USE_HTTP") {
```
mihaip, line 669: May want to use envknob.RegisterBool both because it's more efficient to read and because you avoid having to repeat the knob name 3 times.

derp/derphttp/derphttp_client.go
valscale: Thanks, that helps. What do you think about having an additional envknob to say "ignore the DERP map, just connect to the DERP at this hostname?" I think the infra is all there for that. I'd like to take out the "edit the derp map" step if possible.

derp/derphttp/derphttp_client.go
valscale: Thanks! I was wondering in particular if this should be added to magicsock/debugknobs{,stubs}.go since there is another envknob in this file that isn't in there, and I'm not clear on how we decide what goes in there and what doesn't?

derp/derphttp/derphttp_client.go
valscale: Ooh, good point. With the mesh code I can see wanting to run multiple derps locally for local development. But let's leave that for when we need it.

derp/derphttp/derphttp_client.go
valscale: Great, will do.

derp/derphttp/derphttp_client.go
mihaip: @bradfitz can say for sure, but I think that file is for magicsock-specific knobs, and this is operating at a slightly different level. 

We also used to be a lot more memory constrained on iOS (before iOS 15, https://tailscale.com/blog/go-linker/ has the full story), which would perhaps explain why we didn't want to include `envknob` on iOS, but we're OK with it nowadays.

derp/derphttp/derphttp_client.go
mihaip: That additional knob sounds good to me, I would have found it useful.

bradfitz:
nit: `derp/derphttp: ` in the commit message (package name as the prefix before the colon)

but once a bunch are involved, you can abbreviate, like `derp,magicsock: ....`

But no underscores. The prefix isn't about filenames.

derp/derphttp/derphttp_client.go
valscale: I added a TS_DEBUG_USE_DERP_ADDR which fakes up a derp map/region/node and it's really great.

derp/derphttp/derphttp_client.go
valscale: It's now down to two instances so I left these as-is, but I registered the other two envknobs.

wgengine/magicsock/debugknobs.go
```diff
@@ -33,13 +33,19 @@ var (
...
+	debugUseDERPHTTP = envknob.RegisterBool("TS_DEBUG_USE_DERP_HTTP")
 	// debugEnableSilentDisco disables the use of heartbeatTimer on the endpoint struct
 	// and attempts to handle disco silently. See issue #540 for details.
 	debugEnableSilentDisco = envknob.RegisterBool("TS_DEBUG_ENABLE_SILENT_DISCO")
-	// DebugSendCallMeUnknownPeer sends a CallMeMaybe to a
+	// debugSendCallMeUnknownPeer sends a CallMeMaybe to a
 	// non-existent destination every time we send a real
 	// CallMeMaybe to test the PeerGoneNotHere logic.
 	debugSendCallMeUnknownPeer = envknob.RegisterBool("TS_DEBUG_SEND_CALLME_UNKNOWN_PEER")
+	// Hey you! Adding a new debugknob? Make sure to stub it out in the debugknob_stubs.go
```
mihaip, line 47: CI will catch this (it builds the Wasm/JS target), so not strictly required to add the comment.

wgengine/magicsock/magicsock.go
```diff
@@ -2533,6 +2533,32 @@ func (c *Conn) SetDERPMap(dm *tailcfg.DERPMap) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	if debugUseDERPAddr() != "" {
```
mihaip, line 2536: Nit: I believe it's more idiomatic to save the result in a variable, so that you can use it below without repeating the function call. Something like `if derpAddr := debugUseDERPAddr(); derpAddr != "" {`

wgengine/magicsock/magicsock.go
```diff
@@ -2533,6 +2533,32 @@ func (c *Conn) SetDERPMap(dm *tailcfg.DERPMap) {
...
+		// We only need to set the DERPMap once
```
mihaip, line 2537: There's already an early exit below, this one may not be necessary. It's possible for envknob values to be changed at runtime (via `envknob.Setenv`), so `debugUseDERPAddr()` may return a different value across calls.

mihaip (approved):

wgengine/magicsock/debugknobs.go
valscale: It does catch it but I just hate every time the CI goes red on something and I have to do another force push. :)

wgengine/magicsock/magicsock.go
```diff
@@ -2533,6 +2533,29 @@ func (c *Conn) SetDERPMap(dm *tailcfg.DERPMap) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	var derpAddr = debugUseDERPAddr()
+	if derpAddr != "" {
```
mihaip, line 2537: May want to consider putting the assignment in the if statement, that way `derpAddr` is scoped to it and is easier to reason about (the reader doesn't have to wonder if it's used later in the function).

```suggestion
	if derpAddr := debugUseDERPAddr(); derpAddr != "" {
```

mihaip (approved):

PR #7748, opened 2023-03-31
cmd/tailscale/cli: fix fixTailscaledConnectError for macsys version

jilyaluk:
Currently, on macOS Standonlone version if local client returns any errors while tailscaled is running, cli would not print the error at all, showing only that Tailscale might not be running.

Add tailscaled process name of macsys version, so that fixTailscaledConnectError reports error correctly.

Contains workaround for go-ps bug which trims long process names (>16 characters): mitchellh/go-ps#15


jilyaluk:
Hey @bradfitz, could you please take a look? I can't add reviewers to the PR.

PR #7767, opened 2023-04-03
tsweb: expose native Prometheus metrics in /debug/varz

knyar:
The handler will expose built-in process and Go metrics by default, which currently duplicate some of the expvar-proxied metrics (`goroutines` vs `go_goroutines`, `memstats` vs `go_memstats`), but as long as their names are different, Prometheus server will just scrape both.

This will change /debug/varz behaviour for most tsweb binaries, but notably not for control, which configures a `tsweb.VarzHandler` [explicitly](https://github.com/tailscale/corp/blob/a5b5d5167f1a2b780d313ec6a30e7b96dfa25b2d/cmd/tailcontrol/tailcontrol.go#L779).

In a second commit, migrated the prober library to native Prometheus metrics.


knyar:
Hm, this increased `tailscaled` binary size from 21MB to 27.4MB, perhaps I will need to rethink this in a way that does not add additional dependencies to the client.

cmd/derper/depaware.txt
```diff
@@ -8,21 +8,63 @@ tailscale.com/cmd/derper dependencies: (generated by github.com/tailscale/depawa
...
+        github.com/prometheus/common/internal/bitbucket.org/ww/goautoneg from github.com/prometheus/common/expfmt
+        github.com/prometheus/common/model                           from github.com/prometheus/client_golang/prometheus+
+  LD    github.com/prometheus/procfs                                 from github.com/prometheus/client_golang/prometheus
+  LD    github.com/prometheus/procfs/internal/fs                     from github.com/prometheus/procfs
+  LD    github.com/prometheus/procfs/internal/util                   from github.com/prometheus/procfs
         github.com/x448/float16                                      from github.com/fxamacker/cbor/v2
      💣 go4.org/mem                                                  from tailscale.com/client/tailscale+
         go4.org/netipx                                               from tailscale.com/wgengine/filter
    W 💣 golang.zx2c4.com/wireguard/windows/tunnel/winipcfg           from tailscale.com/net/interfaces+
+        google.golang.org/protobuf/encoding/prototext                from github.com/golang/protobuf/proto+
```
danderson, line 40: We should try really hard to not pull protobuf into our client builds. It adds a ton of bloat to binaries. If there isn't a way to use that prometheus client lib without protobuf, that might be a deal breaker :/

danderson (changes requested):
In theory, looks good. In practice, that's a large increase to client binary size, especially on iOS where we have a hard memory limit (binaries contribute to that limit). I'm guessing almost all of hte increase comes from the inclusion of protobuf :/

cmd/derper/depaware.txt
knyar: For now, I believe, native histograms can only be exposed in protobuf format rather than text format (text format is [blocked on openmetrics](https://github.com/prometheus/prometheus/issues/11265)), so I think we'll actually need protobuf as a dependency to export native histograms.

I understand the constraints around client size. What I will try to do is restructure the modules in a way to make the official Prometheus client (and protobuf) an optional dependency for `tsweb` users, which we'll use for `derper` and other servers. The client will continue to only use our hand-rolled expvar converter, at least for now.

Please let me know if you have any other thoughts.

knyar:
I have extracted the expvar-to-prometheus converter into `tsweb/varz` and isolated the new dependency on the Go prometheus library in a new package `tsweb/promvarz`. This allows us to use native Prometheus metrics in server side software (like the `prober` binary in this PR), keeping `tailscaled` with only a dependency on `tsweb/varz`.

These new changes are in a separate, third commit.

This is ready for another review.

danderson (approved):

PR #7770, opened 2023-04-03
cmd/tailscale/cli: do not allow turning Funnel on while shields-up

shayne:
Fixes: tailscale/corp#9027

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -2544,6 +2544,9 @@ func (b *LocalBackend) checkPrefsLocked(p *ipn.Prefs) error {
 	if err := b.checkExitNodePrefsLocked(p); err != nil {
 		errs = append(errs, err)
 	}
+	if err := b.checkFunnelEnabledLocked(p); err != nil {
+		errs = append(errs, err)
+	}
 	return multierr.New(errs...)
 }
 
@@ -2624,6 +2627,13 @@ func (b *LocalBackend) isDefaultServerLocked() bool {
 func (b *LocalBackend) checkExitNodePrefsLocked(p *ipn.Prefs) error {
 	if (p.ExitNodeIP.IsValid() || p.ExitNodeID != "") && p.AdvertisesExitNode() {
 		return errors.New("Cannot advertise an exit node and use an exit node at the same time.")
+	}
+	return nil
+}
+
+func (b *LocalBackend) checkFunnelEnabledLocked(p *ipn.Prefs) error {
+	if p.ShieldsUp && b.serveConfig.IsFunnelOn() {
+		return errors.New("Cannot enable shields-up when Funnel is enabled.")
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


cmd/tailscale/cli/funnel.go
```diff
@@ -73,6 +73,18 @@ func (e *serveEnv) runFunnel(ctx context.Context, args []string) error {
...
+	prefs, err := e.lc.GetPrefs(ctx)
+	if err != nil {
+		return fmt.Errorf("getting prefs: %w", err)
+	}
+
+	if on && prefs.ShieldsUp {
+		return fmt.Errorf("%s\n%s",
+			"Unable to turn on Funnel while shields-up=true.",
+			"Run `tailscale set --shields-up=false` to turn off Shields.")
+	}
```
maisem, line 86: I think you want to do this in local.go
https://github.com/tailscale/tailscale/blob/f4da995940ea13bda69979ec1a07251c6b757c5d/ipn/ipnlocal/local.go#L2532

ipn/ipnlocal/serve.go
```diff
@@ -218,6 +218,11 @@ func (b *LocalBackend) SetServeConfig(config *ipn.ServeConfig) error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 
+	prefs := b.sanitizedPrefsLocked()
```
maisem, line 221: you can do `prefs := b.pm.CurrentPrefs()`

ipn/ipnlocal/serve.go
```diff
@@ -218,6 +218,11 @@ func (b *LocalBackend) SetServeConfig(config *ipn.ServeConfig) error {
...
+	if config.IsFunnelOn() && prefs.ShieldsUp() {
+		return errors.New("Unable to turn on Funnel while shields-up is enabled")
```
maisem, line 223: this is fine, but you also need the opposite side

maisem (approved):

ipn/ipnlocal/local.go
```diff
@@ -2628,6 +2631,18 @@ func (b *LocalBackend) checkExitNodePrefsLocked(p *ipn.Prefs) error {
 	return nil
 }
 
+func (b *LocalBackend) checkFunnelEnabledLocked(p *ipn.Prefs) error {
+	sc0 := b.serveConfig
+	if !sc0.Valid() {
+		return nil
+	}
+	sc1 := sc0.AsStruct()
```
maisem, line 2639: Oh thats an expensive call just to check if funnel is enabled. You can selectively expose methods from structs to their corresponding views, so we should add a method "IsFunnelOn" to the view and use it here.

See tailcfg.Node.IsTagged as an example.

maisem (approved):

ipn/ipnlocal/local.go
shayne: Updated. PTAL. IsTagged looked pretty standard, but I think I did what we wanted based on other view methods.

shayne: [posterity] Example @ `dbx.Node.IsTagged`

Updated. PTAL

maisem (approved):

PR #7776, opened 2023-04-03
util/sysresources, magicsock: scale DERP buffer based on system memory

andrew-d:
This adds the util/sysresources package, which currently only contains a function to return the total memory size of the current system.

Then, we modify magicsock to scale the number of buffered DERP messages based on the system's available memory, ensuring that we never use a value lower than the previous constant of 32.

Change-Id: Ib763c877de4d0d4ee88869078e7d512f6a3a148d


wgengine/magicsock/magicsock.go
```diff
@@ -1418,12 +1419,54 @@ func (c *Conn) sendAddr(addr netip.AddrPort, pubKey key.NodePublic, b []byte) (s
...
+		// is about 0.04% of the total system memory.
+		//
+		// For sake of this calculation, then, let's double that memory
+		// usage to 0.08% and scale based on total system memory.
+		//
+		// For a 16GiB Linux box, this should buffer just over 256
+		// messages.
+		systemMemory := sysresources.TotalMemory()
+		memoryUsable := float64(systemMemory) * 0.0008
+		bufferedDerpWrites = int(memoryUsable / (100 * 512))
```
maisem, line 1461: why are you dividing by 100 again?

maisem: or does that not represent a percentage?

andrew-d: Theoretical upper bound for the number of DERP regions we have; will constant-ify

andrew-d: Changed the numbers to constants to clarify what the calculation is doing; PTAL?

maisem (approved):

PR #7777, opened 2023-04-03
sockstats: add client metrics for radio power state

willnorris:
power state is very roughly approximated based on observed network activity and AT&T's state transition timings for a typical 3G radio.

Updates tailscale/corp#9230
Updates #3363

### DIFF
--- a/net/sockstats/sockstats_tsgo.go
+++ b/net/sockstats/sockstats_tsgo.go
@@ -9,11 +9,13 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"math"
 	"net"
 	"strings"
 	"sync"
 	"sync/atomic"
 	"syscall"
+	"time"
 
 	"tailscale.com/net/interfaces"
 	"tailscale.com/util/clientmetric"
@@ -49,6 +51,7 @@ var sockStats = struct {
 	currentInterfaceCellular atomic.Bool
 
 	txBytesMetric, rxBytesMetric, txBytesCellularMetric, rxBytesCellularMetric *clientmetric.Metric
+	radioHighMetric                                                            *clientmetric.Metric
 }{
 	countersByLabel:       make(map[Label]*sockStatCounters),
 	knownInterfaces:       make(map[int]string),
@@ -57,6 +60,7 @@ var sockStats = struct {
 	rxBytesMetric:         clientmetric.NewCounter("sockstats_rx_bytes"),
 	txBytesCellularMetric: clientmetric.NewCounter("sockstats_tx_bytes_cellular"),
 	rxBytesCellularMetric: clientmetric.NewCounter("sockstats_rx_bytes_cellular"),
+	radioHighMetric:       clientmetric.NewGaugeFunc("sockstats_cellular_radio_high_fraction", radio.radioHighPercent),
 }
 
 func withSockStats(ctx context.Context, label Label) context.Context {
@@ -122,6 +126,9 @@ func withSockStats(ctx context.Context, label Label) context.Context {
 		if sockStats.currentInterfaceCellular.Load() {
 			sockStats.rxBytesCellularMetric.Add(int64(n))
 			counters.rxBytesCellularMetric.Add(int64(n))
+			if n > 0 {
+				radio.active()
+			}
 		}
 	}
 	didWrite := func(n int) {
@@ -136,6 +143,9 @@ func withSockStats(ctx context.Context, label Label) context.Context {
 		if sockStats.currentInterfaceCellular.Load() {
 			sockStats.txBytesCellularMetric.Add(int64(n))
 			counters.txBytesCellularMetric.Add(int64(n))
+			if n > 0 {
+				radio.active()
+			}
 		}
 	}
 	willOverwrite := func(trace *net.SockTrace) {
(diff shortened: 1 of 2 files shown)


net/sockstats/sockstats_tsgo.go
```diff
@@ -122,6 +123,9 @@ func withSockStats(ctx context.Context, label Label) context.Context {
 		if sockStats.currentInterfaceCellular.Load() {
 			sockStats.rxBytesCellularMetric.Add(int64(n))
 			counters.rxBytesCellularMetric.Add(int64(n))
+			if n > 0 {
```
mihaip, line 126: I'm somewhat concerned about this adding a per-packet mutex lock + timer creation/reset overhead. An alternative design to consider is to just store an atomic int of the time of last activity, and then to have a low-frequency timer (1 Hz) that does the state transitions based on a time comparison between that int and the current time.

@raggi / @jwhited do you all have a preference from a perf perspective?

net/sockstats/sockstats_tsgo.go
```diff
@@ -276,3 +283,71 @@ func isLikelyCellularInterface(ifName string) bool {
...
+// Timings for radio power state transitions taken from
+// https://developer.android.com/training/connectivity/network-access-optimization#radio-state
+// Even though that documents a typical 3G radio and newer radios are much more efficient,
+// it provides worst-case timings to use for analysis.
+const (
+	radioHighIdle = 5 * time.Second  // time radio idles in high power state before transitioning to low
+	radioLowIdle  = 12 * time.Second // time radio idles in low power state before transitioning to off
+)
+
+func setRadioState(newState radioState) {
```
mihaip, line 325: This seems pretty unit-testable if you were so inclined.

net/sockstats/sockstats_tsgo.go
willnorris: yeah, I had the same concern.  Fortunately, there's very little in (the current) setRadioState that would take any time.  I did have an atomic int in a previous iteration, though without the timer for handling state transitions... I ended up replacing it with the mutex, but I could certainly go back to that if we're particularly concerned.  This is certainly at a low enough point in the stack that we need to be really careful.

net/sockstats/sockstats_tsgo.go
```diff
@@ -276,3 +283,71 @@ func isLikelyCellularInterface(ifName string) bool {
...
+	radio.mu.Lock()
+	defer radio.mu.Unlock()
+
+	switch newState {
+	case radioHighPower:
+		if radio.state != newState {
+			radio.wakeTime = time.Now()
+			if radio.timerLow != nil {
+				radio.timerLow.Stop()
```
mihaip, line 334: I think you'd want to increment `lowMetric` in this case too, for cases where the radio gets turned on again without making it all the way to an off state.

willnorris: ah yeah, you're right.  Went too far with trying to simplify this.

net/sockstats/sockstats_tsgo.go
raggi: Is it my understanding that we're doing this for iOS only right now?

Per the brief discussion we had early on that was the case, and eschewed worrying much about performance.

If we wanted to do this for all platforms then there's a good number of considerations to expand into as the impact even on x64/arm64 at >10gbps is relevant. It is also relevant at much lower rates for our 386 and arm32 builds as the implementations don't optimize on sub-architecture features today (sse for intel, neon for arm, etc).

At the highest level, and as a gross generalization, a pointer load is the cheapest of the ops here. The Add/CAS/etc start to get into cross-domain instructions that may have a much higher cost and vary by architecture and topology. A Mutex fast path is just a CAS (when uncontended).

The timers do allocate on the heap (more than 64 bytes if I eyeballed correctly), so I share the concern about their cost. The only data needed from the data path to generate this information is a single atomic write of a boottime monotonic (that is, monotonic that ignores sleep) and everything else, current state, etc could be handled by the outside thread.

So:
- Where do we want to run (platforms, users, etc)?
- How much precision do we need?

willnorris: I've updated the implementation, but to answer your questions:

 - platforms / users: this is only included in macOS, iOS, and and android builds.  iOS and android are the ones we really care about.  This particular client metric will be on for all users.
 - precision: we can be somewhat loose, since our timings for transition states are very approximate anyway.  The new implementation only uses second precision, since we're now looking at percentage of time spent in different power states over the last hour, and transition periods are measured in whole seconds.

New implementation is doing a `time.Now()` call per packet and storing that in a fixed-size array.  All other computation is handled when reporting client metrics.

willnorris:
@mihaip PTAL.  this is still only reporting the single metric of time spent in high power state.  I think it should be relatively easy to also record time spent in low power state, with the caveat that state transitions that occur over the hour boundary would be a little off.  But even the 12 second low power transition period over a 1 hour window is 0.3%, so not enough to skew our numbers in a meaningful way.

If you think this approach makes sense, I start on adding low power state as well (and probably change the metric names)

net/sockstats/sockstats_tsgo.go
raggi: time.Now is exceptionally slow on Xen hypervisors (about 10% or so of the Amazon fleet), as it's a work-stealing hypercall (no kvm-clock style setup, and no vdso fast-path). It's best avoided in the datapath if at all possible.

Another general issue is that CLOCK_MONOTONIC particularly for these purposes in general, may not provide the desired accuracy across sleeps (it stops on some platforms). Jason was telling me recently that he's moving WireGuard over to CLOCK_BOOTTIME.

willnorris: since we're only building this for the above platforms (enforced by build tags), that shouldn't matter in this particular case should it?

willnorris: well, put another way... how bad is time.Now() on macOS, iOS, and android?

willnorris: I guess if we're rounding to second precision anyway, we could just have a one second ticker that looks for a change in the total Tx/Rx counters to detect activity.  Then we stay out of the datapath entirely.

raggi: I assume it's ok for those more limited, but this is one of those areas where people don't look often enough, and they get broken by sandbox changes and so on. Our using Go in a performance critical setting in these platforms is rare, so we might be the only ones who _would be affected_ - this is all to say we're going to need to test.

We can look at the implementations too, for example see `nanotime` in https://tip.golang.org/src/runtime/sys_darwin_amd64.s as you can see in the amd64 case here it varies depending on libc state, so starting to look raises more questions.

I just realized I misread "This particular client metric will be on for all users." to mean more platforms, but on re-reading I don't think that's what it meant.

As this has come up before, what I've been wanting to do, though Brad expressed an initial distaste for doing so, is to have an explicit single global clock that we step for each packet that we process, and all consumers read from that clock, with one or two places where we explicitly drive it forward. This is what netstack does, for the same reasons.

raggi: To be clear: don't let me block you, we don't have data here. It shouldn't be catastrophic, but it may contribute to reduced performance and we'll have to come back and have a look when we study performance again. I'm more strongly concerned about macOS here than the mobiles, but there are use cases such as moving videos around from peoples primary video camera where it does matter.

net/sockstats/sockstats_tsgo.go
```diff
@@ -276,3 +286,73 @@ func isLikelyCellularInterface(ifName string) bool {
 		strings.HasPrefix(ifName, "ww") || // systemd naming scheme for WWAN
 		strings.HasPrefix(ifName, "pdp") // iOS
 }
+
+// radioMonitor tracks usage of the cellular radio, approximates the power state transitions,
+// and reports the percentage of time the radio was on.
+type radioMonitor struct {
+	// usage tracks the last time (as unix timestamp) the radio was used over the last hour.
+	// Values are indexed by the number of seconds since the beginning of the current hour.
+	usage [3600]int64
```
mihaip, line 295: May want to add a `const` for the 3600 since it shows up in a bunch of places.

mihaip (approved):

net/sockstats/sockstats_tsgo.go
willnorris: done

PR #7796, opened 2023-04-05
ssh/tailssh: make Tailscale SSH work on gokrazy

bradfitz:


creachadair (approved):

ipn/ipnlocal/local.go
```diff
@@ -3220,6 +3220,10 @@ func (b *LocalBackend) TailscaleVarRoot() string {
 	switch runtime.GOOS {
 	case "ios", "android", "darwin":
 		return paths.AppSharedDir.Load()
+	case "linux":
+		if distro.Get() == distro.Gokrazy {
+			return "/perm/tailscaled"
```
stapelberg, line 3225: Why is hard-coding this path necessary here? Since https://github.com/gokrazy/gokrazy/commit/e2c5a62f677847d491591be77950f1dba084eda5, we use `/perm/home/tailscaled` for new installations. I wonder if we should update the path here to match the new default path? Or avoid it entirely somehow?

ssh/tailssh/incubator.go
```diff
@@ -691,6 +691,9 @@ func (ss *sshSession) startWithStdPipes() (err error) {
 func loginShell(u *user.User) string {
 	switch runtime.GOOS {
 	case "linux":
+		if distro.Get() == distro.Gokrazy {
+			return "/tmp/serial-busybox/ash"
```
stapelberg, line 695: Likewise, I’m a little concerned about hard-coding the serial-busybox path here.

(Maybe we’ll eventually follow the suggestion to use u-root instead of busybox, for example.)

Maybe we should create /bin/sh on gokrazy when gokrazy finds ash? So that we only need to hard-code it in one place, not in downstream applications like tailscale?

stapelberg (commented):
Thanks @bradfitz :)

PR #7805, opened 2023-04-05
util/clientmetric: allow client metric values to be provided by a function

mihaip:
Adds `NewGaugeFunc` and `NewCounterFunc` (inspired by `expvar.Func`), which allow the current value to be reported by a function. This allows some client metric values to be computed on-demand during uploading (at most every 15 seconds), instead of being continuously updated.

clientmetric uploading had a bunch of micro-optimizations for memory access (#3331) which are not possible with this approach. However, any performance hit from function-based metrics is contained to those metrics only, and we expect to have very few.

Also adds a `DisableDeltas()` option for client metrics, so that absolute values are always reported. This makes server-side processing of some metrics easier to reason about.

Updates tailscale/corp#9230

### DIFF
--- a/util/clientmetric/clientmetric.go
+++ b/util/clientmetric/clientmetric.go
@@ -39,8 +39,9 @@ var (
 // scanEntry contains the minimal data needed for quickly scanning
 // memory for changed values. It's small to reduce memory pressure.
 type scanEntry struct {
-	v          *int64 // Metric.v
-	lastLogged int64  // last logged value
+	v          *int64       // Metric.v
+	f          func() int64 // Metric.f
+	lastLogged int64        // last logged value
 }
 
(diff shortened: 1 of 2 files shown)


util/clientmetric/clientmetric.go
```diff
@@ -55,10 +56,12 @@ const (
 //
 // It's safe for concurrent use.
 type Metric struct {
-	v      *int64 // atomic; the metric value
-	regIdx int    // index into lastLogVal and unsorted
-	name   string
-	typ    Type
+	v              *int64       // atomic; the metric value
+	f              func() int64 // alternate value function to use (instead of v)
```
willnorris, line 60: perhaps make clear that f takes precedence over v?

willnorris (approved):

Issue #7813, opened 2023-04-06
FR: Kubernetes operator, print out version in log


rodrigc:
### What are you trying to do?

When debugging the Tailscale Kubernetes operator, this is displayed:

```
kubectl -n tailscale logs -l app=operator
```

```
{"level":"info","ts":1680644846.8081307,"logger":"startup","msg":"Startup complete, operator running"}
```

It would help if the version of the operator was displayed also.

### How should we solve this?

This would help in debugging and root causing issues with the Tailscale Kubernetes operator

PR #7814, opened 2023-04-06
cmd/k8s-operator: print version in startup logs

rodrigc:
Fixes: #7813


rodrigc:
Since the k8s-operator is quite new and undergoing active development, I thought that it might
be helpful to print out the version of the operator on startup, to help with debugging.

danderson (approved):

danderson:
Works for me, thanks!

PR #7819, opened 2023-04-07
net/sockstats: return early if no radio period length

willnorris:


net/sockstats/sockstats_tsgo.go
```diff
@@ -336,6 +336,10 @@ func (rm *radioMonitor) radioHighPercent() int64 {
 	}
 	periodStart := now - periodLength // start of current reporting period
 
+	if periodLength <= 0 {
```
mihaip, line 339: Nit: could move this to be inside the if block at line 335 (and check `t <= 0`), since that's the only place where it can actually be 0.

mihaip (approved):

net/sockstats/sockstats_tsgo.go
willnorris: done

PR #7821, opened 2023-04-08
tailcfg,wgengine: add initial support for WireGuard only peers

raggi:
A peer can have IsWireGuardOnly, which means it will not support DERP or Disco, and it must have Endpoints filled in order to be usable.

In the present implementation only the first Endpoint will be used as the bestAddr.

Updates tailscale/corp#10351


wgengine/magicsock/magicsock.go
```diff
@@ -4656,6 +4688,11 @@ func (de *endpoint) updateFromNode(n *tailcfg.Node, heartbeatDisabled bool) {
 	de.sendFunc.Store(nil)
 }
 
+// supportsDisco reports whether or not the endpoint supports disco.
+func (de *endpoint) supportsDisco() bool {
+	return !de.discoKey.IsZero()
```
maisem, line 4693: Can we use IsWireGuardOnly?

tailcfg/tailcfg.go
```diff
@@ -289,6 +289,11 @@ type Node struct {
 	// peer or any of its subnets. Traffic originating from subnet routes will
 	// not be masqueraded (e.g. in case of --snat-subnet-routes).
 	SelfNodeV4MasqAddrForThisPeer netip.Addr `json:",omitempty"`
+
+	// IsWireGuardOnly indicates that this peer is a pure WireGuard peer, it is
+	// not expected to speak Disco or DERP, and it must have Endpoints in order
+	// to be reachable.
+	IsWireGuardOnly bool
```
maisem, line 296: Bump capver too

wgengine/magicsock/magicsock.go
raggi: pre 0.100 nodes match this, but do not match IsWireGuardOnly. While that shouldn't be a problem due to guards elsewhere in the code, should those guards change, this check is the safer one, as it guards against entering a code path that will panic the client (the noise shared key production from discoKey). As such I think being specific here is more robust to future change.

tailcfg/tailcfg.go
```diff
@@ -289,6 +290,11 @@ type Node struct {
...
```
bradfitz, line 297: omitempty

raggi: Thanks for taking a look!

wgengine/magicsock/magicsock.go
raggi: I looked at this again briefly. I can see a few issues here:

- If the compiler fails to optimize zero checks well, the 32 byte zero check could be expensive, though I beleive this isn't in particularly hot paths. I swapped out for an empty string check against discoShort, which assuming the commentary on the struct remains correct should be functionally equivalent, but faster.
- There's poor locking behavior around upsertEndpoint and elsewhere. I improved this now for the case of _this change_, but there are further cases that now I need to look into - including at least one I want to look at for this change.
- As above, I'm not sure plumbing through IsWireGuardOnly to the endpoint is necessarily a better change here. That would cover fewer cases, and I would end up wanting to do something like short-circuit on IsWireGuardOnly, but then fall down to checking discoShort or discoKey to correctly cover the case of older nodes without disco keys. Swapping the method out to be ep.isWireGuardOnly has more subtle usage throughout, which I don't think is a good approach - call sites need more expert knowledge about the difference between IsWireGuardOnly and pre-0.100 nodes. This method instead is specific and general around the state that needs protection.

If you have more specific concerns you can articulate, can you expand those?

maisem: thanks for the info, no concerns was just asking for context

tailcfg/tailcfg.go
```diff
@@ -289,6 +290,11 @@ type Node struct {
 	// peer or any of its subnets. Traffic originating from subnet routes will
 	// not be masqueraded (e.g. in case of --snat-subnet-routes).
 	SelfNodeV4MasqAddrForThisPeer netip.Addr `json:",omitempty"`
+
+	// IsWireGuardOnly indicates that this peer is a pure WireGuard peer, it is
```
maisem, line 294: i would say `non-Tailscale WireGuard peer`

wgengine/magicsock/magicsock.go
raggi: After fixing up the synchronization, this method disappeared.

wgengine/magicsock/magicsock.go
```diff
@@ -4455,8 +4471,9 @@ func (de *endpoint) send(buffs [][]byte) error {
 		return errNoUDPOrDERP
 	}
 	var err error
+
 	if udpAddr.IsValid() {
-		_, err = de.c.sendUDPBatch(udpAddr, buffs)
+		_, err := de.c.sendUDPBatch(udpAddr, buffs)
```
maisem, line 4476: this looks wrong, it was previously updating the err outside of this block

tailcfg/tailcfg.go
```diff
@@ -289,6 +290,11 @@ type Node struct {
 	// peer or any of its subnets. Traffic originating from subnet routes will
 	// not be masqueraded (e.g. in case of --snat-subnet-routes).
 	SelfNodeV4MasqAddrForThisPeer netip.Addr `json:",omitempty"`
+
+	// IsWireGuardOnly indicates that this is a non-Tailscale WireGuard peer, it
+	// is not expected to speak Disco or DERP, and it must have Endpoints in
+	// order to be reachable.
+	IsWireGuardOnly bool `json:",omitempty"`
```
maisem, line 297: Maybe leave a note saying that as of `2023-04-06 only the first parseable Endpoint is used`

maisem (approved):

wgengine/magicsock/magicsock.go
raggi: :O good catch, thank you.

PR #7825, opened 2023-04-08
wgengine/magicsock: fix synchronization of endpoint disco fields

raggi:
Identified in review in https://github.com/tailscale/tailscale/pull/7821 endpoint.discoKey and endpoint.discoShort
are often accessed without first taking endpoint.mu. The arrangement
with endpoint.mu is inconvenient for a good number of those call-sites,
so it is instead replaced with an atomic pointer to carry both pieces of
disco info. This will also help with https://github.com/tailscale/tailscale/pull/7821 that wants to add explicit
checks/guards to disable disco behaviors when disco keys are missing
which is necessarily implicitly mostly covered by this change.

Updates #7821

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -203,13 +203,17 @@ func (m *peerMap) upsertEndpoint(ep *endpoint, oldDiscoKey key.DiscoPublic) {
 	if m.byNodeKey[ep.publicKey] == nil {
 		m.byNodeKey[ep.publicKey] = newPeerInfo(ep)
 	}
-	if oldDiscoKey != ep.discoKey {
+	epDisco := ep.disco.Load()
+	if epDisco == nil || oldDiscoKey != epDisco.key {
 		delete(m.nodesOfDisco[oldDiscoKey], ep.publicKey)
 	}
-	set := m.nodesOfDisco[ep.discoKey]
+	if epDisco == nil {
+		return
+	}
+	set := m.nodesOfDisco[epDisco.key]
 	if set == nil {
 		set = map[key.NodePublic]bool{}
-		m.nodesOfDisco[ep.discoKey] = set
+		m.nodesOfDisco[epDisco.key] = set
 	}
 	set[ep.publicKey] = true
 }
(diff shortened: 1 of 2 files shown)


maisem (approved):
🤦🏼

raggi:
@maisem as discussed, after looking further, there were more problems, so approach changed to cover more cases. PTAL.

maisem (approved):

wgengine/magicsock/magicsock.go
```diff
@@ -4040,6 +4069,11 @@ func ippDebugString(ua netip.AddrPort) string {
 // recalculated.
 type endpointSendFunc func([][]byte) error
 
+type endpointDisco struct {
+	key   key.DiscoPublic // for discovery messages.
+	short string          // ShortString of discoKey.
+}
```
bradfitz, line 4075: document that this is immutable? I hope/assume.

wgengine/magicsock/magicsock.go
```diff
@@ -4057,12 +4091,11 @@ type endpoint struct {
 	fakeWGAddr   netip.AddrPort // the UDP address we tell wireguard-go we're using
 	nodeAddr     netip.Addr     // the node's first tailscale address; used for logging & wireguard rate-limiting (Issue 6686)
 
+	disco atomic.Pointer[endpointDisco] // if the peer supports disco, the key and short string
+
 	// mu protects all following fields.
 	mu sync.Mutex // Lock ordering: Conn.mu, then endpoint.mu
```
bradfitz, line 4097: ugh, so some fields are behind a mutex and some are an atomic?

raggi: I know. I looked at the mutex strategy and due to the way these two fields are accessed moving over to using the mutex properly would have been painful and require restructuring more implementations. The mutex is used inconsistently today.

I'm not particularly happy with it either, but the root cause is at the higher level organization - that requires us to sit down and review the design in order to address it.

wgengine/magicsock/magicsock.go
raggi: It is. Will do.

raggi: https://github.com/tailscale/tailscale/pull/7833

wgengine/magicsock/magicsock.go
bradfitz: Maybe I'm missing it, but I can't see from the comments on #7821 what the actual data race was.

Which code path was using these fields without the proper locks being held?

raggi: Conn.upsertEndpoint, Conn.deleteEndpoint, Conn.handleDiscoMessage, Conn.unambiguousNodeKeyOfPingLocked, usage of ep.shortString in a number of logf calls, and endpoint.String.

Of all of these the one most tricky to address is endpoint.String has it has no clear lock bounds and is a generic interface that escapes into a foreign code path, it may not be safe to introduce a lock there.

I could write a version of this that uses the mutex everywhere, but drops the disco key from endpoint.String?

PR #7834, opened 2023-04-10
wgengine/netstack: fix data-race on startup

raggi:
Running tailscaled with the race detector enabled immediately fires on this field, as it is updated after first read.

Updates #cleanup


wgengine/netstack/netstack.go
```diff
@@ -215,6 +215,8 @@ func Create(logf logger.Logf, tundev *tstun.Wrapper, e wgengine.Engine, mc *magi
 	}
 	ns.ctx, ns.ctxCancel = context.WithCancel(context.Background())
 	ns.atomicIsLocalIPFunc.Store(tsaddr.NewContainsIPFunc(nil))
+	ns.tundev.PostFilterPacketInboundFromWireGaurd = ns.injectInbound
```
creachadair, line 218: I think I've convinced myself these values don't depend on the stuff before their old location (notwithstanding the goroutine). 👀

creachadair (approved):

Issue #7844, opened 2023-04-11
FR: Add `funnel https ...` support for single-command serve + Funnel support


shayne:
Based on user feedback, we should consider a single-command for serving and turning Funnel on.

PR #7845, opened 2023-04-11
cmd/tailscale/cli: [funnel] add https:<port> ... ability

shayne:
Adds all-in-one ability to start a serve + Funnel with one command.

Closes: #7844


bradfitz:
Nit: in the commit, instead of `Closes: #7844` can you do `Fixes #7844`? I know GitHub supports both (and a hundred other variants), but we try to mostly stick with "Fixes" or "Updates" as the two closing and non-closing verbs, respectively. And without a colon.

cmd/tailscale/cli/funnel.go
```diff
@@ -33,14 +33,32 @@ func newFunnelCommand(e *serveEnv) *ffcli.Command {
 		ShortUsage: strings.TrimSpace(`
 funnel <serve-port> {on|off}
   funnel status [--json]
+  funnel https:<port> <mount-point> <source> [off]
```
bradfitz, line 36: why do you need the source for `off` mode?

shayne: We don't; you can give just the mount point.

cmd/tailscale/cli/funnel.go
```diff
@@ -33,14 +33,32 @@ func newFunnelCommand(e *serveEnv) *ffcli.Command {
...
+`),
+		LongHelp: strings.TrimSpace(`
+*** BETA; all of this is subject to change ***
+
+Funnel allows you to publish a Tailscale Serve
+server publicly, open to the entire internet.
+
+EXAMPLES
+  - To toggle Funnel on port 443 (default):
```
bradfitz, line 45: mention somehow that 443 is https? People might not know?

cmd/tailscale/cli/funnel.go
shayne: But in the terminal if you do `<up> off` it'll work

cmd/tailscale/cli/funnel.go
```diff
@@ -33,14 +33,32 @@ func newFunnelCommand(e *serveEnv) *ffcli.Command {
...
+    $ tailscale funnel 443 on
+    $ tailscale funnel 443 off
+
+    Turning off Funnel only turns off serving to the internet.
+    It does not affect serving to your tailnet.
+
+  - To proxy requests to a web server at 127.0.0.1:3000:
+    $ tailscale funnel https:443 / http://127.0.0.1:3000
```
bradfitz, line 53: Oh, I didn't think we were going to go this direction from that thread. Now there are two ways to configure web mounts. Are they separate namespaces or one? Both way are confusing for different reasons.

Did you not like a `--funnel` flag to `serve`?

Or you just really like positional arguments?

shayne: We had a call earlier today where we discussed a lot of the feedback. I'm happy to chat with you if you want to catch up on that conversation.

There's a simple use case of enabling Funnel to point to some source. That involves two commands today, the `serve` command and the `funnel` command. The first sets up the local server, and the latter toggles Funnel.

The thinking for folks who are just looking to answer the question, "How can I enable a Funnel to my local server?" we have the short-cut command:

`$ tailscale funnel https / localhost:3000`

This sets up the server and enables the funnel in one shot. Further, if you want to turn it off, you just add `off` to the end, which turns the local server and Funnel off.

If we went attach it to `serve` it might look like this:

`$ tailscale serve --funnel https / localhost:3000`

I'm not a fan of the flag, especially if I'm thinking about a simpler use case. I also don't want to bury the Funnel command under a flag in the `serve` subcommand.

There's also the issue with the position of the flag. I'm a little tired and foggy on it, but I'm pretty sure it has to follow the `serve` sub-command, so if folks try to add it on at the end, it won't work, which is arbitrary and frustrating. But I haven't looked at flags in a while. I think that's how it was though.

Lastly, there's also the case of turning it off. If you just add `off` to the end of the command, does that turn off Funnel? Probably not, unless you have the `--funnel` flag on that command. 

What I'd like to do is really elevate `serve` to capital Serve. And promote it as a swiss-army knife built-in server capability we added to `tailscaled`. The additions to the `funnel` command in this PR are just a macro to the `serve` + `funnel` command hoop folks jump through now. It simplifies the common case of both turning on and off a Serve with Funnel enabled.

bradfitz: I'm not opposed to making a way to enable them both in one shot but I want to make sure we're not introducing two ways to do the same thing and potential ambiguity about which world you're operating in: e.g. is `tailscale funnel https / localhost:3000` operating in a different world (public) than `tailscale serve https / localhost:3000`? Like, can you configure different HTTP routing as a function of whether the traffic came from the tailnet or the internet?

> What I'd like to do is really elevate serve to capital Serve. And promote it as a swiss-army knife built-in server capability we added to tailscaled

It being a swiss-army knife of serving was the goal of it. It doing 3 things already (TCP forwarding, TLS termination of TCP, and HTTPS) is much more than 1. I'm sure we can add more. (like plain HTTP) What else do you want to add? You'd mentioned RDP and VNC. Are those more than named ports?

Issue #7769, opened 2023-04-03
DNS resolution failing on macsys after upgrading to 1.39.x


lincolnmantracer:
### What is the issue?

I recently upgraded Tailscale macsys from 1.38.3 to 1.39.x (.2 I think?). DNS resolution immediately started failing despite several restarts and "reinstalling" the system extension. Errors from a go program looked like this:

```
lookup foo.example.com on 100.100.100.100:53: read udp 100.124.137.53:63265->100.100.100.100:53: i/o timeout
```

We use split dns for the domain in question, with "override local DNS" checked. Downgrading to 1.38 immediately fixed the issue.

### OS

macOS

### OS version

macOS Monterey 12.3.1

### Tailscale version

1.39.2

### Bug report

BUG-beeb61f2c7675c37eed3fc1620c04be9cbcfc4a05ea7d8675daad3110d57edd0-20230403123654Z-b7b311863d49a81e

mihaip:
Looking at our [internal logs](http://logz.ts.tailscale.com:9431/logs?start=2023-03-10T16%3A58%3A20Z&end=2023-04-10T16%3A58%3A20Z&instances=beeb61f2c7675c37eed3fc1620c04be9cbcfc4a05ea7d8675daad3110d57edd0&match=goSetupLogs), these are the versions that were installed on that machine:
- 1.39.55 (installed on March 27)
- 1.39.92 (installed on April 2)
- 1.38.3 (installed on April 3)

Based on how broken things are, I'm guessing that this started to happen with the 1.39.92 release rather than the 1.39.55 one, which narrows down the regression range somewhat.

Changes between 1.39.55 and 1.39.92: [open source](https://github.com/tailscale/tailscale/compare/4a89642f7...7b4e85aa7), [closed source](https://github.com/tailscale/corp/compare/52f5a6d7f...b07177ba9). Nothing immediately jumps out.

mihaip:
@knyar is also experiencing this, and has narrowed it down a bit: 1.39.73 is the first version it’s broken in; 1.39.55 works fine ([open source changes](https://github.com/tailscale/tailscale/compare/4a89642f7...920ec6924), [closed source changes](https://github.com/tailscale/corp/compare/52f5a6d7f...d55f94b15)).

[...]

PR #7849, opened 2023-04-11
tool/gocross: add ts_macext build tag for Xcode builds

mihaip:
It's used to control various opt-in functionality for the macOS and iOS apps, and was lost in the migration to gocross.

Updates tailscale/tailscale#7769


mihaip:
Would appreciate tips for how to try this out (as far as what I have to tweak on the corp side to build with this change). Relatedly, do we need to do anything beyond a normal OSS update to pick up the change once it's merged?

tool/gocross/autoflags.go
```diff
@@ -102,6 +102,10 @@ func autoflagsForTest(argv []string, env *Environment, goroot, nativeGOOS, nativ
 		cgo = nativeGOOS == "darwin"
 		tags = append(tags, "omitidna", "omitpemdecrypt")
 		if env.IsSet("XCODE_VERSION_ACTUAL") {
+			// If we're building via Xcode, we must be making the extension
+			// version (as opposed to tailscaled on Mac).
+			tags = append(tags, "ts_macext")
```
danderson, line 107: Does ts_macext cover the App Store build as well?

danderson (approved):
Oops, sorry for missing that!

tool/gocross/autoflags.go
mihaip: As far as I can tell it was specified for all the Xcode builds (iOS, macOS, macsys): https://github.com/tailscale/corp/blob/72401f9d3bd6eecfde5f5a2cb70a5469fb012bf0/xcode/ipn-go-bridge/default.a.do#L43-L51.

PR #7854, opened 2023-04-12
.github/workflows: run one set of tests with the tailscale_go build tag

mihaip:
We use it to gate code that depends on custom Go toolchain, but it's currently only passed in the corp runners. Add a set on OSS so that we can catch regressions earlier.
    
To specifically test `sockstats` this required adding a build tag to explicitly enable them -- they're normally on for iOS, macOS and Android only, and we don't run tests on those platforms normally.


mihaip:
(verified via a revert of 22680a11ae2a3061c57123bc6afac2f11b7b51a7 that this repros the CI failure we saw in corp)

willnorris (approved):

PR #7857, opened 2023-04-13
tailcfg: make SelfNodeV4MasqAddrForThisPeer a pointer

maisem:
This makes `omitempty` actually work, and saves bytes in each map response.

Updates tailscale/corp#8020


raggi (approved):
Thanks!

PR #7864, opened 2023-04-14
wgengine/magicsock: factor out receiveIPv4 & receiveIPv6 common code

bradfitz:
Updates #2331

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -322,11 +322,6 @@ type Conn struct {
 	// bind is the wireguard-go conn.Bind for Conn.
 	bind *connBind
 
-	// ippEndpoint4 and ippEndpoint6 are owned by receiveIPv4 and
-	// receiveIPv6, respectively, to cache an IPPort->endpoint for
-	// hot flows.
-	ippEndpoint4, ippEndpoint6 ippEndpointCache
-
 	// ============================================================
 	// Fields that must be accessed via atomic load/stores.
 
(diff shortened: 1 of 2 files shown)


maisem (approved):

wgengine/magicsock/magicsock.go
```diff
@@ -1850,80 +1845,65 @@ func (c *Conn) putReceiveBatch(batch *receiveBatch) {
...
-			if neterror.PacketWasTruncated(err) {
-				// TODO(raggi): discuss whether to log?
-				continue
+		batch := c.getReceiveBatchForBuffs(buffs)
+		defer c.putReceiveBatch(batch)
+		for {
+			numMsgs, err := ruc.ReadBatch(batch.msgs[:len(buffs)], 0)
+			if err != nil {
+				if neterror.PacketWasTruncated(err) {
+					// TODO(raggi): discuss whether to log?
```
raggi, line 1879: this TODO can go, we don't need to log this.

raggi (approved):
Thanks for coming and cleaning this up, I've been considering doing so.

I'd love it if we could avoid the closure, but I'm not sure that'd be easy to do.

bradfitz:
FWIW, the closure is allocated exactly once. Performance-wise it's no different than a method on a type. If it gets any more complicated I'd move it to a type with methods but this is still within my threshold of it being easier to use a closure to close over a couple variables.

jwhited (approved):

PR #7865, opened 2023-04-14
net/netcheck: use dnscache.Resolver when resolving DERP IPs

andrew-d:
This also adds a bunch of tests for this function to ensure that we're returning the proper IP(s) in all cases.

Change-Id: I0d9d57170dbab5f2bf07abdf78ecd17e0e635399

### DIFF
--- a/net/netcheck/netcheck.go
+++ b/net/netcheck/netcheck.go
@@ -25,6 +25,7 @@ import (
 	"github.com/tcnksm/go-httpstat"
 	"tailscale.com/derp/derphttp"
 	"tailscale.com/envknob"
+	"tailscale.com/net/dnscache"
 	"tailscale.com/net/interfaces"
 	"tailscale.com/net/netaddr"
 	"tailscale.com/net/neterror"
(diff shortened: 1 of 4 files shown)


net/netcheck/netcheck.go
```diff
@@ -1597,12 +1605,46 @@ func (c *Client) nodeAddr(ctx context.Context, n *tailcfg.DERPNode, proto probeP
 		return
 	}
 
-	// TODO(bradfitz): add singleflight+dnscache here.
-	addrs, _ := net.DefaultResolver.LookupIPAddr(ctx, n.HostName)
+	c.mu.Lock()
```
bradfitz, line 1608: lot of lines between this Lock and the Unlock in this otherwise unrelated method.  can any/all of this be pulled out into a new method?

bradfitz:
What's the bug? None's linked. We don't resolve DERP IPs. We push down DERP IPs in the DERPMap.

andrew-d:
> What's the bug? None's linked. We don't resolve DERP IPs. We push down DERP IPs in the DERPMap.

I wanted to add tests to this function (looking at it for a Coral thing) and figured I'd resolve the TODO; happy to remove the `dnscache` changes and the TODO if you don't think we should do this.

bradfitz:
Naah, this is fine. I guess it'll help users with custom DERPs who don't specify IPs.

net/netcheck/netcheck.go
andrew-d: Moved the default resolver function out of the lock section; PTAL?

net/netcheck/netcheck.go
```diff
@@ -181,6 +182,11 @@ type Client struct {
 	// If nil, portmap discovery is not done.
 	PortMapper *portmapper.Client // lazily initialized on first use
 
+	// UseDNSCache controls whether this client should use a
+	// *dnscache.Resolver to resolve host names. If false, the default
+	// net.Resolver will be used, with no caching.
```
bradfitz, line 187: clarify which hostnames? If only DERP hostnames, then also clarify that DERP hostnames aren't resolved if the IP(s) are provided in the DERP map, as they typically are.

So this then is only used for people with custom DERPs who don't spell out the IP addresses?

bradfitz (approved):

bradfitz:
🚢

PR #7877, opened 2023-04-14
wgengine/magicsock: keep advertising endpoints after we stop discovering them

andrew-d:
Previously, when updating endpoints we would immediately stop advertising any endpoint that wasn't discovered during `determineEndpoints`. This could result in, for example, a case where we performed an incremental netcheck, didn't get any of our three STUN packets back, and then dropped our STUN endpoint from the set of advertised endpoints... which would result in clients falling back to a DERP connection until the next call to `determineEndpoints`.

Instead, let's cache endpoints that we've discovered and continue reporting them to clients until a timeout expires. In the above case where we temporarily don't have a discovered STUN endpoint, we would continue reporting the old value, then re-discover the STUN endpoint again and continue reporting it as normal, so clients never see a withdrawal.

Updates tailscale/coral#108

Change-Id: I42de72e7418ab328a6c732bdefc74549708cf8b9

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -63,6 +63,7 @@ import (
 	"tailscale.com/util/clientmetric"
 	"tailscale.com/util/mak"
 	"tailscale.com/util/ringbuffer"
+	"tailscale.com/util/set"
 	"tailscale.com/util/sysresources"
 	"tailscale.com/util/uniq"
 	"tailscale.com/version"
@@ -418,6 +419,10 @@ type Conn struct {
 	// onEndpointRefreshed are funcs to run (in their own goroutines)
 	// when endpoints are refreshed.
 	onEndpointRefreshed map[*endpoint]func()
+
+	// endpointTracker tracks the set of cached endpoints that we advertise
+	// for a period of time before withdrawing them.
+	endpointTracker endpointTracker
 
 	// peerSet is the set of peers that are currently configured in
 	// WireGuard. These are not used to filter inbound or outbound
@@ -1195,6 +1200,22 @@ func (c *Conn) determineEndpoints(ctx context.Context) ([]tailcfg.Endpoint, erro
 	}
 
 	c.ignoreSTUNPackets()
+
+	// Update our set of endpoints by adding any endpoints that we
+	// previously found but haven't expired yet. This also updates the
+	// cache with the set of endpoints discovered in this function.
+	//
+	// NOTE: we do this here and not below so that we don't cache local
+	// endpoints; we know that the local endpoints we discover are all
+	// possible local endpoints since we determine them by looking at the
+	// set of addresses on our local interfaces.
+	//
+	// TODO(andrew): If we pull in any cached endpoints, we should probably
+	// do something to ensure we're propagating the removal of those cached
+	// endpoints if they do actually time out without being rediscovered.
+	// For now, though, rely on a minor LinkChange event causing this to
+	// re-run.
+	eps = c.endpointTracker.update(time.Now(), eps)
 
 	if localAddr := c.pconn4.LocalAddr(); localAddr.IP.IsUnspecified() {
 		ips, loopback, err := interfaces.LocalAddresses()
@@ -4148,6 +4169,11 @@ const (
 	// STUN-derived endpoint valid for. UDP NAT mappings typically
 	// expire at 30 seconds, so this is a few seconds shy of that.
 	endpointsFreshEnoughDuration = 27 * time.Second
+
+	// endpointTrackerLifetime is how long we continue advertising an
+	// endpoint after we last see it. This is intentionally chosen to be
+	// slightly longer than a full netcheck period.
+	endpointTrackerLifetime = 5*time.Minute + 10*time.Second
 )
 
 // Constants that are variable for testing.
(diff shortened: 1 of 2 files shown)


wgengine/magicsock/magicsock.go
```diff
@@ -5105,6 +5125,88 @@ func (s derpAddrFamSelector) PreferIPv6() bool {
...
+	mu    sync.Mutex
+	cache map[netip.AddrPort]endpointTrackerEntry
+}
+
+const endpointTrackerLifetime = 5*time.Minute + 10*time.Second // slightly longer than a full netcheck period
+
+func (et *endpointTracker) update(now time.Time, eps []tailcfg.Endpoint) (ret []tailcfg.Endpoint) {
+	ret = eps
+
+	// We can have a very large set of endpoints for machines with many
```
danderson, line 5143: Making a couple dicts to diff and update a list is a fairly common pattern in Go, so IMO there's no need to spell it out. It made me double-take a bit and hunt for what might be different, when nothing was.

wgengine/magicsock/magicsock_test.go
```diff
@@ -2280,3 +2283,113 @@ func TestIsWireGuardOnlyPeerWithMasquerade(t *testing.T) {
...
+			name: "after timeout still caches",
+			now:  start.Add(4*time.Minute + endpointTrackerLifetime + time.Minute),
+			eps:  []tailcfg.Endpoint{local},
+			want: []tailcfg.Endpoint{local, stun4_2, stun6_2},
+		},
+	}
+
+	var et endpointTracker
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
```
danderson, line 2379: If the tests are interdependent, don't t.Run them. t.Run allows me to -run=Some/subset/of/tests, which will result in failures in this case. Since the cases depend on each other, this is a single big test, not a collection of small ones.

danderson (approved):
The data structures in endpointTracker feel like they could be more efficient, but I couldn't work out the exact shape of them during code review, and that's easy to improve later on. LGTM.

wgengine/magicsock/magicsock.go
```diff
@@ -5105,6 +5125,88 @@ func (s derpAddrFamSelector) PreferIPv6() bool {
...
+}
+
+type endpointTracker struct {
+	mu    sync.Mutex
+	cache map[netip.AddrPort]endpointTrackerEntry
+}
+
+const endpointTrackerLifetime = 5*time.Minute + 10*time.Second // slightly longer than a full netcheck period
+
+func (et *endpointTracker) update(now time.Time, eps []tailcfg.Endpoint) (ret []tailcfg.Endpoint) {
```
bradfitz, line 5140: document what it returns? Or change `ret` to something like `stillAlive` or `newlyAdded` or `nowDeleted` or whatever it is

wgengine/magicsock/magicsock.go
```diff
@@ -5105,6 +5125,88 @@ func (s derpAddrFamSelector) PreferIPv6() bool {
...
+	// We can have a very large set of endpoints for machines with many
+	// interfaces; rather than a O(n²)-ish loop over every entry in eps for
+	// each entry in the cache, create a map so that lookups are cheap.
+	//
+	// TODO(andrew): we could probably have a fast path for when we have a
+	// small number of endpoints.
+	inputEps := make(map[netip.AddrPort]bool, len(eps))
+	for _, ep := range eps {
+		inputEps[ep.Addr] = true
+	}
```
bradfitz, line 5152: yeah, I guess the only operation you care about is "is it in the set?" Our corp `misc/set.Slice` basically does this with its:

```go
// Slice is a set of elements tracked in a slice of unique elements.
type Slice[T comparable] struct {
	slice []T
	set   map[T]bool // nil until/unless slice is large enough
}

// Contains reports whether v is in the set.
// The amortized cost is O(1).
func (ss *Slice[T]) Contains(v T) bool {
	if ss.set != nil {
		return ss.set[v]
	}
	return slices.Index(ss.slice, v) != -1
}
```

wgengine/magicsock/magicsock.go
andrew-d: Renamed to `epsPlusCached`

wgengine/magicsock/magicsock.go
andrew-d: Moved that type here to oss and pushed an update to use it 👍

wgengine/magicsock/magicsock.go
andrew-d: Removed this comment now that we're using a `set.Slice` 👍

wgengine/magicsock/magicsock.go
```diff
@@ -5105,6 +5126,82 @@ func (s derpAddrFamSelector) PreferIPv6() bool {
...
+	endpoint tailcfg.Endpoint
+	until    time.Time
+}
+
+type endpointTracker struct {
+	mu    sync.Mutex
+	cache map[netip.AddrPort]endpointTrackerEntry
+}
+
+const endpointTrackerLifetime = 5*time.Minute + 10*time.Second // slightly longer than a full netcheck period
```
bradfitz, line 5139: move this to the big const(...) block of magic time durations elsewhere in this file/package?

bradfitz (approved):

PR #7879, opened 2023-04-14
util/set: move Slice type from corp to oss

andrew-d:
This is an exact copy of the files misc/set/set{,_test}.go from https://github.com/tailscale/corp/commit/a5415daa9ca80ce9b798cd80148e4bc8818f8e15, plus the license headers.

For use in #7877

Change-Id: I712d09c6d1a180c6633abe3acf8feb59b27e2866

### DIFF
--- a/util/set/slice.go
+++ b/util/set/slice.go
@@ -1,0 +1,69 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package set
+
+import (
+	"golang.org/x/exp/slices"
+	"tailscale.com/types/views"
+)
+
+// Slice is a set of elements tracked in a slice of unique elements.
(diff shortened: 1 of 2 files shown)


util/set/slice.go
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package set
```
bradfitz, line 4: needs package doc. maybe in a new file `set.go` if we foresee more types of sets. (I think we have other types in corp already?)

andrew-d: Package doc already exists in `set.go`, albeit not the most verbose:
https://github.com/tailscale/tailscale/blob/main/util/set/set.go#L4

bradfitz (approved):

Issue #7876, opened 2023-04-14
Tracking bug: restoring Windows CI


raggi:
### What is the issue?

We regressed on windows CI in https://github.com/tailscale/tailscale/commit/4022796484db7f44f0a8598ed9a5d880d1a5b6ae and since then have not been running any Windows tests.

This lead to a bunch of test regressions, where tests aren't passing on Windows.

In order to get CI back soon, I'm adding test skips on the failing tests referencing this bug, and once there are no more TODO's referencing this bug we're done.

--

Left over:

- [x] net/netcheck/netcheck_test.go:          t.Skip("TODO(#7876): test regressed on windows while CI was broken")
- [x] tsnet/tsnet_test.go:            t.Skip("TODO(#7876): test regressed on windows while CI was broken")

raggi:
All done!

PR #7880, opened 2023-04-14
ipn/ipnlocal: reenable profile tests on Windows

raggi:
This fix does not seem ideal, but the test infrastructure using a local goos doesn't seem to avoid all of the associated challenges, but is somewhat deeply tied to the setup.

The core issue this addresses for now is that when run on Windows there can be no code paths that attempt to use an invalid UID string, which on Windows is described in [1].

For the goos="linux" tests, we now explicitly skip the affected migration code if runtime.GOOS=="windows", and for the Windows test we explicitly use the running users uid, rather than just the string "user1". We also now make the case where a profile exists and has already been migrated a non-error condition toward the outer API.

Updates #7876

[1] https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers

### DIFF
--- a/ipn/ipnlocal/profiles.go
+++ b/ipn/ipnlocal/profiles.go
@@ -22,6 +22,8 @@ import (
 	"tailscale.com/util/winutil"
 )
 
+var errAlreadyMigrated = errors.New("profile migration already completed")
+
 // profileManager is a wrapper around a StateStore that manages
 // multiple profiles and the current profile.
 type profileManager struct {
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/profiles.go
```diff
@@ -544,7 +546,7 @@ func newProfileManagerWithGOOS(store ipn.StateStore, logf logger.Logf, goos stri
 		if err := pm.setPrefsLocked(prefs); err != nil {
 			return nil, err
 		}
-	} else if len(knownProfiles) == 0 && goos != "windows" {
+	} else if len(knownProfiles) == 0 && goos != "windows" && runtime.GOOS != "windows" {
```
creachadair, line 549: I don't follow why we need to check both of these. If that's intentional (which I assume it is) a comment might be worthwhile.

creachadair (approved):

ipn/ipnlocal/profiles.go
raggi: It is intentional, I'll add a comment. I think this needs some extra cleanup - I think the intent was for most behavior to be driven by goos parameter, but that isn't consistently the case, and we execute tests with `goos == 'linux'` in the Windows pipeline which implies behavior by build tag which otherwise fails.

dblohm7 (approved):

PR #7881, opened 2023-04-14
safesocket: enable test to run on Windows unpriviliged

raggi:
I manually tested that the code path that relaxes pipe permissions is not executed when run with elevated priviliges, and the test also passes in that case.

Updates #7876

### DIFF
--- a/safesocket/basic_test.go
+++ b/safesocket/basic_test.go
@@ -10,10 +10,10 @@ import (
 	"testing"
 )
 
+// downgradeSDDL is a no-op test helper on non-Windows systems.
+var downgradeSDDL = func() func() { return func() {} }
+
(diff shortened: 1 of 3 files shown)


safesocket/basic_test.go
```diff
@@ -10,10 +10,10 @@ import (
 	"testing"
 )
 
+// downgradeSDDL is a no-op test helper on non-Windows systems.
+var downgradeSDDL = func() func() { return func() {} }
```
creachadair, line 14: This signature tho 🤣

creachadair (approved):

safesocket/basic_test.go
raggi: i know right, super funcy

PR #7882, opened 2023-04-15
syncs: relax TestWatchMultipleValues timing on Windows

raggi:
The test is re-enabled for Windows with a relaxed time assertion.

On Windows the runtime poller currently does not have sufficient
resolution to meet the normal requirements for this test.

See https://github.com/golang/go/issues/44343 for background.

Updates https://github.com/tailscale/tailscale/issues/7876

### DIFF
--- a/syncs/watchdog_test.go
+++ b/syncs/watchdog_test.go
@@ -46,9 +46,6 @@ func TestWatchContended(t *testing.T) {
 }
 
 func TestWatchMultipleValues(t *testing.T) {
-	if runtime.GOOS == "windows" {
-		t.Skip("TODO(#7876): test regressed on windows while CI was broken")
-	}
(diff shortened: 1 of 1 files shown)


raggi:
Sorry to send this over to you brad, but I need eyes on this in case you happen to know of usage somewhere else I wasn't able to find. From my grep it only showed up in one place internally atm.

bradfitz:
There's nothing Linux-specific about this. I'd rather just keep the t.Skip on windows with a note or bug reference.

raggi:
> There's nothing Linux-specific about this. I'd rather just keep the t.Skip on windows with a note or bug reference.

I looked up the root cause and referenced that bug, and set the timing so as to avoid that issue.

creachadair (approved):

PR #7883, opened 2023-04-15
wgengine/magicsock: reenable magicsock tests on Windows

raggi:
These tests are passing locally and on CI. They had failed earlier in
the day when first fixing up CI, and it is not immediately clear why. I
have cycled IPv6 support locally, but this should not have a substantial
effect.

Updates #7876


raggi:
Oh, they started passing locally because I'd disabled IPv6 when testing the DNSCache test.

crawshaw (approved):

PR #7892, opened 2023-04-16
wgengine/magicsock: update comments on what implements conn.Bind

bradfitz:
The comment still said *magicsock.Conn implemented wireguard-go conn.Bind. That wasn't accurate anymore.

A doc #cleanup.

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -268,7 +268,6 @@ func (m *peerMap) deleteEndpoint(ep *endpoint) {
 }
 
 // A Conn routes UDP packets and actively manages a list of its endpoints.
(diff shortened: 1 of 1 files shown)


wgengine/magicsock/magicsock.go
```diff
@@ -3063,8 +3079,8 @@ func (c *connBind) Close() error {
 	return nil
 }
 
-// Closed reports whether c is closed.
-func (c *connBind) Closed() bool {
+// isClosed reports whether c is closed.
+func (c *connBind) isClosed() bool {
```
bradfitz, line 3083: (unexported so it doesn't look like it's part of the interface)

maisem (approved):

Issue #7621, opened 2023-03-20
High CPU usage in the face of routing table churn (Tailscale 1.38.1, Ubuntu 22.04)


davidbalbert:
### What is the issue?

I've been replacing the previous iteration of the VPN we use for managing our network equipment with Tailscale and ran into a snag.

Last night I put Tailscale on our BGP router. It's a [Supermicro 5018D-FN8T](https://www.supermicro.com/products/system/1u/5018/SYS-5018D-FN8T.cfm) with a 4 core, 8 thread Broadwell-based Xeon D-1518 running Ubuntu 22.04 and FRR. We're getting full IPv4+IPv6 tables – 1.07 million routes – from a single upstream provider (our backup is just sending a default route).

Without Tailscale, our CPU usage is basically zero – htop reports a load average of between 0.1 and 0.2 out of 8, or between 1.25 and 2.5% total utilization.

With Tailscale, Load average is between 1.6 and 1.8, and htop reports `tailscaled` using between 150% and 300% CPU. I took this measurement while SSH'd in via a public IP address, so there was no traffic going over `tailscale0`.

After running `tailscale down` and `tailscale logout`, `tailscaled` uses less CPU, but it still runs hot – Between 20% and 170% CPU with a load average of ~0.5.

Here's the output of the `top` command in `go tool pprof` on the results of `tailscale debug --cpu-profile` while Tailscale was logged in and running:

```
(pprof) top 20
Showing nodes accounting for 20.07s, 73.03% of 27.48s total
Dropped 218 nodes (cum <= 0.14s)
Showing top 20 nodes out of 125
      flat  flat%   sum%        cum   cum%
     7.62s 27.73% 27.73%      7.62s 27.73%  runtime/internal/syscall.Syscall6
     1.69s  6.15% 33.88%      7.02s 25.55%  runtime.mallocgc
     1.65s  6.00% 39.88%      1.65s  6.00%  runtime.memclrNoHeapPointers
     1.50s  5.46% 45.34%      5.61s 20.41%  runtime.scanobject
     0.95s  3.46% 48.80%      1.56s  5.68%  runtime.findObject
     0.85s  3.09% 51.89%      1.91s  6.95%  runtime.greyobject
     0.84s  3.06% 54.95%      0.88s  3.20%  runtime.pageIndexOf (inline)
     0.62s  2.26% 57.21%      0.62s  2.26%  runtime.memmove
[...]

PR #7900, opened 2023-04-17
net/interfaces: cache "home" router lookup on big Linux routers

bradfitz:
This is a continuation of the earlier 2a67beaacfbeee8f9055c2e5d9200a1fa81dc94a but more aggressive; this now remembers that we failed to find the "home" router IP so we don't try again later on the next call.

Updates #7621


andrew-d (approved):
👍

Issue #7850, opened 2023-04-11
Very frequent "no gateway index found" log messages when there's no cell reception


mihaip:
I was in a place with low connectivity for a few hours, and Tailscale was responsible for a lot of battery use.

| 2PM | 3PM | 4PM |
|-|-|-|
| ![Screenshot 2023-04-09 at 5 07 44 PM](https://user-images.githubusercontent.com/513813/231306228-ce9132e8-ec3b-4701-89e9-7ab2eb2af99f.png) | ![Screenshot 2023-04-09 at 4 57 08 PM](https://user-images.githubusercontent.com/513813/231306233-463d7682-8d3b-4115-888a-73381d910fea.png) | ![Screenshot 2023-04-09 at 5 07 09 PM](https://user-images.githubusercontent.com/513813/231306230-d2294448-b5c0-45cd-9865-2e68680d46cb.png) |

At least part of that was due to log uploads ([sockstats](http://logz.ts.tailscale.com:9431/sockstats?start=2023-04-09T21%3A00%3A00Z&end=2023-04-10T00%3A04%3A42Z&instances=eb5cafba4ece2a3a99c7563d0a89b796107161c359f028891ad0cd7962ce189a)) -- there were some very large requests that were retried and eventually went though.
![image](https://user-images.githubusercontent.com/513813/231306789-fd9a759a-b073-4a76-b7bd-0247ed3a1ccf.png)

Looking at my [logs](http://logz.ts.tailscale.com:9431/logs?start=2023-04-09T21%3A00%3A00Z&end=2023-04-10T00%3A04%3A42Z&instances=eb5cafba4ece2a3a99c7563d0a89b796107161c359f028891ad0cd7962ce189a) for that period, there's a lot of `DefaultRouteInterfaceIndex: no gateway index found` errors being logged (about 10 lines/second). This is both spammy and means that we're not using the cached response from `monitor.mon.InterfaceState`

PR #7902, opened 2023-04-17
net/dnsfallback: more explicitly pass through logf function

mihaip:
Redoes the approach from #5550 and #7539 to explicitly pass in the `logf` function, instead of having global state that can be overridden.


mihaip:
@andrew-d somewhat of a drive-by cleanup (found while I was investigating #7850, which requires passing around a `wgengine/monitor` `Mon` in most places where we also pass in a `logf`).

andrew-d (approved):
Thanks! This seems like an improvement ✨

PR #7904, opened 2023-04-17
net/connstats: exclude traffic with internal Tailscale service

dsnet:
Exclude traffic with 100.100.100.100 (for IPv4) and with fd7a:115c:a1e0::53 (for IPv6) since this traffic with the Tailscale service running locally on the node.
This traffic never left the node.

It also happens to be a high volume amount of traffic since DNS requests occur over UDP with each request coming from a unique port, thus resulting in many discrete traffic flows.

Fixes tailscale/corp#10554

### DIFF
--- a/net/connstats/stats.go
+++ b/net/connstats/stats.go
@@ -13,6 +13,7 @@ import (
 
 	"golang.org/x/sync/errgroup"
 	"tailscale.com/net/packet"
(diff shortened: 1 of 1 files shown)


ramyfication (approved):

mihaip (approved):

dsnet:
Verified locally that it works. Submitting.

PR #7905, opened 2023-04-17
ci: add golangci-lint configuration and CI action

andrew-d:
This PR first adds configuration for the [`golangci-lint`](https://github.com/golangci/golangci-lint) linter with some initial (and hopefully non-controversial) rules, along with one commit per linter fixing the issues found. Finally, the last commit in this series adds the linter to our CI, where it will currently run only on PRs (i.e. not the `main` branch).

I intend to squash these into a single commit on merge.


danderson (approved):
I'm fine with it, but advance warning that some tailscalars very dislike linter nags that aren't strictly correctness errors, and may end up turning it all off :)

maisem (approved):
LGTM, but tests so red!

andrew-d:
> I'm fine with it, but advance warning that some tailscalars very dislike linter nags that aren't strictly correctness errors, and may end up turning it all off :)

Yeah, totally understood. My goal here is to start with only the things that are currently being caught by a human reviewer during style checks... and I was half-tempted to name this "robo-Brad" 😆🙈

creachadair:
FWIW I have generally found golangci-lint quite annoying, but that was on projects that started from the default grab-bag of every random opinionated thing in the universe. Starting from a basic subset and growing out feels like a good compromise.

A good rule of thumb, IMO, is if we find ourselves having to restrict its checks to "only code in this PR", we have probably crossed onto the bad side of the line.

andrew-d:
> FWIW I have generally found golangci-lint quite annoying, but that was on projects that started from the default grab-bag of every random opinionated thing in the universe. Starting from a basic subset and growing out feels like a good compromise.

Same here; I think we need to be pretty opinionated about what we choose to lint, which is why this is opt-in-only, and intentionally minimal to start.

> A good rule of thumb, IMO, is if we find ourselves having to restrict its checks to "only code in this PR", we have probably crossed onto the bad side of the line.

FWIW, this is currently in "report only new issues" mode just to be safe, but after the commits in this PR there are no issues across the entire codebase when I run it locally. 👍

PR #7908, opened 2023-04-17
all: avoid repeated default interface lookups

### LINKED ISSUE #7850: Very frequent "no gateway index found" log messages when there's no cell reception

mihaip:
On some platforms (notably macOS and iOS) we look up the default interface to bind outgoing connections to. This is both duplicated work and results in logspam when the default interface is not available (i.e. when a phone has no connectivity, we log an error and thus cause more things that we will try to upload and fail).

Fixed by passing around a `netmon.Monitor` to more places, so that we can use its cached interface state.

Fixes #7850
Updates #7621


net/netcheck/netcheck.go
```diff
@@ -864,22 +870,29 @@ func (c *Client) GetReport(ctx context.Context, dm *tailcfg.DERPMap) (_ *Report,
 		return c.finishAndStoreReport(rs, dm), nil
 	}
 
-	ifState, err := interfaces.GetState()
-	if err != nil {
-		c.logf("[v1] interfaces: %v", err)
-		return nil, err
+	var ifState *interfaces.State
+	if c.LinkMon == nil || reflect.ValueOf(c.LinkMon).Kind() == reflect.Ptr && reflect.ValueOf(c.LinkMon).IsNil() {
```
bradfitz, line 874: That's not what I would've expected from the docs on the Client.LinkMon type.

Also it means you can't implement `interfaces.LinkMonitor` with a nil pointer receiver. I don't see that being a documentation restriction. That's a valid thing to do in Go.

net/portmapper/portmapper_test.go
```diff
@@ -16,7 +16,7 @@ func TestCreateOrGetMapping(t *testing.T) {
 	if v, _ := strconv.ParseBool(os.Getenv("HIT_NETWORK")); !v {
 		t.Skip("skipping test without HIT_NETWORK=1")
 	}
-	c := NewClient(t.Logf, nil, nil)
+	c := NewClient(t.Logf, nil, nil, nil)
```
bradfitz, line 19: good ol' nil, nil, nil

control/controlclient/noise.go
```diff
@@ -167,6 +169,9 @@ type NoiseClient struct {
 	// be nil.
 	dialPlan func() *tailcfg.ControlDialPlan
 
+	logf    logger.Logf
+	linkMon interfaces.LinkMonitor
```
bradfitz, line 173: document on both of these either `// non-nil` or `// or nil; nil means XXXX`

net/netcheck/netcheck.go
mihaip: This was mostly a bandaid because I was getting panics -- we must have a `var linkMon *monitor.Mon` somewhere or something like it. Is there a preferred pattern I should use instead?

net/portmapper/portmapper_test.go
mihaip: What's you threshold for when I should switch this to a `Config` struct? 4 params?

bradfitz: I kinda like the pattern that constructors only take the required stuff and then the caller can assign to exported fields of whatever the constructor returns. With docs like "you can only set this once before use", etc.

But don't need to do that cleanup now.

net/netcheck/netcheck.go
bradfitz: I'm forgetting the current package dep graph today vs what you're proposing vs what I was toying with the other day, but maybe something like adding a method on `monitor.(*Mon)` that returns a nil interface or a non-nil interface of itself?

```go
// Interface returns m as an interface, if m is non-nil.
// If nil, it returns a nil interface.
func (m *Mon) Interface() LinkMonitor {
    if m == nil {
        return nil
    }
    return m
}
```

(You'd have to duplicate the LinkMonitor definition in the `wgengine/monitor` package but that's okay... docs can cross-reference each other)

mihaip: It seems even nicer to not have the interface at all, and just have a pointer to a `Mon` directly. I sent out #7917 to make it easier to do that without the import path making it look like we're pulling in other wireguard bits.

derp/derphttp/derphttp_client.go
```diff
@@ -88,11 +90,12 @@ func (c *Client) String() string {
 
 // NewRegionClient returns a new DERP-over-HTTP client. It connects lazily.
 // To trigger a connection, use Connect.
-func NewRegionClient(privateKey key.NodePrivate, logf logger.Logf, getRegion func() *tailcfg.DERPRegion) *Client {
+func NewRegionClient(privateKey key.NodePrivate, logf logger.Logf, mon *netmon.Monitor, getRegion func() *tailcfg.DERPRegion) *Client {
 	ctx, cancel := context.WithCancel(context.Background())
```
bradfitz, line 94: document that this can be nil (apparently) and what nil means

Maybe:

    // The mon parameter is optional; if non-nil it's used to do faster interface lookups.

ipn/ipnserver/server.go
```diff
@@ -413,10 +415,11 @@ func (s *Server) addActiveHTTPRequest(req *http.Request, ci *ipnauth.ConnIdentit
 //
 // At some point, either before or after Run, the Server's SetLocalBackend
 // method must also be called before Server can do anything useful.
-func New(logf logger.Logf, logID logid.PublicID) *Server {
+func New(logf logger.Logf, logID logid.PublicID, netMon *netmon.Monitor) *Server {
```
bradfitz, line 418: likewise docs on whether nil means something.

you use "netMon" as the var name here but "mon" earlier. be consistent?

ipn/localapi/localapi.go
```diff
@@ -125,8 +125,8 @@ var (
 	metrics   = map[string]*clientmetric.Metric{}
 )
 
-func NewHandler(b *ipnlocal.LocalBackend, logf logger.Logf, logID logid.PublicID) *Handler {
-	return &Handler{b: b, logf: logf, backendLogID: logID}
+func NewHandler(b *ipnlocal.LocalBackend, logf logger.Logf, netMon *netmon.Monitor, logID logid.PublicID) *Handler {
```
bradfitz, line 128: docs on nil-ness.

(sorry, no docs previously anyway)

I assume b is required, maybe logf too?, and netMon apparently not?

logpolicy/logpolicy.go
```diff
@@ -450,14 +451,14 @@ func tryFixLogStateLocation(dir, cmdname string) {
 
 // New returns a new log policy (a logger and its instance ID) for a
 // given collection name.
-func New(collection string) *Policy {
-	return NewWithConfigPath(collection, "", "")
+func New(collection string, netMon *netmon.Monitor) *Policy {
```
bradfitz, line 454: can it be nil?

if the answer is always yes, then maybe the type itself needs docs saying a nil Monitor is always valid (not sure whether that's true) and what that means.

logpolicy/logpolicy.go
```diff
@@ -675,8 +676,14 @@ func (p *Policy) Shutdown(ctx context.Context) error {
 //   - If DNS lookup fails, consults the bootstrap DNS list of Tailscale hostnames.
 //   - If TLS connection fails, try again using LetsEncrypt's built-in root certificate,
 //     for the benefit of older OS platforms which might not include it.
-func DialContext(ctx context.Context, netw, addr string) (net.Conn, error) {
-	nd := netns.FromDialer(log.Printf, &net.Dialer{
+func CreateDialContext(netMon *netmon.Monitor) func(ctx context.Context, netw, addr string) (net.Conn, error) {
```
bradfitz, line 679: docs on line 674 don't match the new name.

also it needs to be reworded.

bradfitz: probably sounds better as `MakeDialFunc` ... the `Context` bit is a legacy wart we needed to add because the good name was taken, but context is otherwise implicit.

net/dnscache/dnscache.go
```diff
@@ -91,6 +92,11 @@ type Resolver struct {
 	// be added to all log messages printed with this logger.
 	Logf logger.Logf
 
+	// NetMon optionally provides a netmon.Monitor to use to get the current
+	// (cached) network interface. If nil, the interface will be looked up
```
bradfitz, line 96: \nIf nil, ...

net/dnsfallback/dnsfallback.go
```diff
@@ -31,13 +32,13 @@ import (
 	"tailscale.com/util/slicesx"
 )
 
-func Lookup(logf logger.Logf) func(ctx context.Context, host string) ([]netip.Addr, error) {
+func Lookup(logf logger.Logf, netMon *netmon.Monitor) func(ctx context.Context, host string) ([]netip.Addr, error) {
```
bradfitz, line 35: nil docs

net/netns/netns.go
```diff
@@ -55,19 +56,19 @@ func SetDisableBindConnToInterface(v bool) {
 // Listener returns a new net.Listener with its Control hook func
 // initialized as necessary to run in logical network namespace that
 // doesn't route back into Tailscale.
-func Listener(logf logger.Logf) *net.ListenConfig {
+func Listener(logf logger.Logf, mon *netmon.Monitor) *net.ListenConfig {
```
bradfitz, line 59: nil docs

net/netns/netns.go
```diff
@@ -55,19 +56,19 @@ func SetDisableBindConnToInterface(v bool) {
...
+	return &net.ListenConfig{Control: control(logf, mon)}
 }
 
 // NewDialer returns a new Dialer using a net.Dialer with its Control
 // hook func initialized as necessary to run in a logical network
 // namespace that doesn't route back into Tailscale. It also handles
 // using a SOCKS if configured in the environment with ALL_PROXY.
-func NewDialer(logf logger.Logf) Dialer {
-	return FromDialer(logf, &net.Dialer{
+func NewDialer(logf logger.Logf, mon *netmon.Monitor) Dialer {
```
bradfitz, line 70: nil docs & `netMon` instead of `mon` to match most the other code?

net/netns/netns.go
```diff
@@ -76,11 +77,11 @@ func NewDialer(logf logger.Logf) Dialer {
 // network namespace that doesn't route back into Tailscale. It also
 // handles using a SOCKS if configured in the environment with
 // ALL_PROXY.
-func FromDialer(logf logger.Logf, d *net.Dialer) Dialer {
+func FromDialer(logf logger.Logf, mon *netmon.Monitor, d *net.Dialer) Dialer {
```
bradfitz, line 80: same

wgengine/netlog/logger.go
```diff
@@ -91,15 +92,15 @@ var testClient *http.Client
 // is a non-tailscale IP address to contact for that particular tailscale node.
 // The IP protocol and source port are always zero.
 // The sock is used to populated the PhysicalTraffic field in Message.
-func (nl *Logger) Startup(nodeID tailcfg.StableNodeID, nodeLogID, domainLogID logid.PrivateID, tun, sock Device) error {
+func (nl *Logger) Startup(nodeID tailcfg.StableNodeID, nodeLogID, domainLogID logid.PrivateID, tun, sock Device, netMon *netmon.Monitor) error {
```
bradfitz, line 95: nil docs

bradfitz (approved):

ipn/ipnserver/server.go
mihaip: Went with `netMon` everywhere.

PR #7917, opened 2023-04-18
all: move network monitoring from wgengine/monitor to net/netmon

mihaip:
We're using it in more and more places, and it's not really specific to our use of Wireguard.

Also removes the separate interface we had for it in `sockstats` -- it's a small enough package (we already pull in all of its dependencies via other paths) that it's not worth the extra complexity.

Updates #7621
Updates #7850


bradfitz:
I'm happy to move/rename it, but while we're doing so, many it's time to rename the "monitor" to something more clear.

I've been increasingly calling the variable for it `linkMon` ... maybe the package could be `linkmon`? `net/linkmon.(*Monitor)`?  It's not just interfaces. Also routes.

raggi:
I don't know if we'll need to distinguish between interface and link monitoring, but there could be a difference if we decided to implement both. Is this code going to monitor one, or the other or both?

bradfitz:
The point of this package is to watch the system network state (interfaces, their names, up/down, states, their IP addresses, network routes) in the OS/platform-specific ways and then provide portable Go APIs for querying/watching.

For macOS/iOS/BSD, it's via AF_ROUTE sucks and RTM_ RTAX_ etc messages. For Linux, netlink. For Android, various Java callback APIs. etc.

Maybe `net/netmon.(*Monitor)`?  Then we can rename `linkMon` to `netMon`?

raggi:
netMon makes sense to me, it is probably most useful to combine some of the factors into a simpler interface for callers to consume.

e.g. We probably don't care much about interface down, carrier up, but we probably do care about interface up, carrier down.

mihaip:
OK, went with `netmon.Monitor` and renamed all of the variables and fields that I found find. @bradfitz if you want to take another look.

raggi (approved):
LGTM, thanks!

PR #7933, opened 2023-04-20
ssh/tailssh: handle dialing multiple recorders and failing open 

maisem:
This adds support to try dialing out to multiple recorders each
with a 5s timeout and an overall 30s timeout. It also starts respecting
the actions `OnRecordingFailure` field if set, if it is not set
it fails open.

Updates https://github.com/tailscale/corp/issues/9967

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -17,6 +17,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/http/httptrace"
 	"net/netip"
 	"net/url"
 	"os"
@@ -42,6 +43,7 @@ import (
 	"tailscale.com/types/netmap"
 	"tailscale.com/util/clientmetric"
 	"tailscale.com/util/mak"
+	"tailscale.com/util/multierr"
 	"tailscale.com/version/distro"
 )
 
@@ -79,31 +81,9 @@ type server struct {
 
 	// mu protects the following
 	mu                   sync.Mutex
-	httpc                *http.Client                // for calling out to peers.
 	activeConns          map[*conn]bool              // set; value is always true
 	fetchPublicKeysCache map[string]pubKeyCacheEntry // by https URL
 	shutdownCalled       bool
-}
-
-// sessionRecordingClient returns an http.Client that uses srv.lb.Dialer() to
-// dial connections. This is used to make requests to the session recording
-// server to upload session recordings.
-func (srv *server) sessionRecordingClient() *http.Client {
-	srv.mu.Lock()
-	defer srv.mu.Unlock()
-	if srv.httpc != nil {
-		return srv.httpc
-	}
-	tr := http.DefaultTransport.(*http.Transport).Clone()
-	tr.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
-		ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
-		defer cancel()
-		return srv.lb.Dialer().UserDial(ctx, network, addr)
-	}
-	srv.httpc = &http.Client{
-		Transport: tr,
-	}
-	return srv.httpc
 }
 
 func (srv *server) now() time.Time {
@@ -1078,7 +1058,7 @@ func (ss *sshSession) run() {
 			if err != nil {
 				var uve userVisibleError
 				if errors.As(err, &uve) {
-					fmt.Fprintf(ss, "%s\r\n", uve)
+					fmt.Fprintf(ss, "%s\r\n", uve.SSHTerminationMessage())
 				} else {
 					fmt.Fprintf(ss, "can't start new recording\r\n")
 				}
@@ -1086,7 +1066,9 @@ func (ss *sshSession) run() {
 				ss.Exit(1)
 				return
 			}
-			defer rec.Close()
+			if rec != nil {
+				defer rec.Close()
+			}
 		}
 	}
 
@@ -1169,15 +1151,16 @@ func (ss *sshSession) run() {
 // If the final action has a non-empty list of recorders, that list is
 // returned. Otherwise, the list of recorders from the initial action
 // is returned.
-func (ss *sshSession) recorders() []netip.AddrPort {
+func (ss *sshSession) recorders() ([]netip.AddrPort, *tailcfg.SSHRecorderFailureAction) {
 	if len(ss.conn.finalAction.Recorders) > 0 {
-		return ss.conn.finalAction.Recorders
+		return ss.conn.finalAction.Recorders, ss.conn.finalAction.OnRecordingFailure
 	}
-	return ss.conn.action0.Recorders
+	return ss.conn.action0.Recorders, ss.conn.action0.OnRecordingFailure
 }
 
 func (ss *sshSession) shouldRecord() bool {
-	return len(ss.recorders()) > 0
+	recs, _ := ss.recorders()
+	return len(recs) > 0
 }
 
 type sshConnInfo struct {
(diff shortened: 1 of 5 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -1409,16 +1392,103 @@ type CastHeader struct {
 	LocalUser string `json:"localUser"`
 }
 
+// sessionRecordingClient returns an http.Client that uses srv.lb.Dialer() to
+// dial connections. This is used to make requests to the session recording
+// server to upload session recordings.
+// It uses the provided dialCtx to dial connections, and limits a single dial
+// to 5 seconds.
+func (ss *sshSession) sessionRecordingClient(dialCtx context.Context) (hc *http.Client, _ error) {
```
andrew-d, line 1400: Do you need the named return `hc` here?

ssh/tailssh/tailssh.go
```diff
@@ -1409,16 +1392,103 @@ type CastHeader struct {
...
+				errChan <- fmt.Errorf("recording: unexpected status: %v", resp.Status)
+				return
+			}
+			errChan <- nil
+		}()
+		select {
+		case <-got100:
+		case err := <-errChan:
+			errs = append(errs, err)
+			continue
```
andrew-d, line 1478: You could get a nil error here, since if both `got100` and `errChan` have values it's nondeterministic which one you'll receive.

ssh/tailssh/tailssh.go
```diff
@@ -1409,16 +1392,103 @@ type CastHeader struct {
...
+		return nil, errors.New("no peer API transport")
+	}
+	tr := dialer.PeerAPITransport().Clone()
+	dialContextFn := tr.DialContext
+
+	tr.DialContext = func(_ context.Context, network, addr string) (net.Conn, error) {
+		perAttemptCtx, cancel := context.WithTimeout(dialCtx, 5*time.Second)
+		defer cancel()
+		return dialContextFn(perAttemptCtx, network, addr)
+	}
```
andrew-d, line 1412: I have a mild preference for joining the two contexts, in case there's a context key that matters on the input. Something like:

```go
	tr.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
		perAttemptCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
		defer cancel()
		go func() {
			select {
			case <-perAttemptCtx.Done():
			case <-dialCtx.Done():
				cancel()
			}
		}()
		return dialContextFn(perAttemptCtx, network, addr)
	}
```

andrew-d (commented):
Only reviewed the dialing parts; can review the rest if you like?

ssh/tailssh/tailssh.go
maisem: good call, done

ssh/tailssh/tailssh.go
maisem: added an explicit check for `err == nil` here, and added another bad server test

ssh/tailssh/tailssh.go
```diff
@@ -1409,16 +1392,113 @@ type CastHeader struct {
...
+		// got100 is closed when we receive the 100-continue response.
+		got100 := make(chan struct{})
+		ctx = httptrace.WithClientTrace(ctx, &httptrace.ClientTrace{
+			Got100Continue: func() {
+				close(got100)
+			},
+		})
+
+		pr, pw := io.Pipe()
+		req, err := http.NewRequestWithContext(ctx, "POST", fmt.Sprintf("http://%s:%d/record", ap.Addr(), ap.Port()), pr)
```
catzkorn, line 1460: Shouldn't we use the URL package here to set up the URL?

ssh/tailssh/tailssh.go
```diff
@@ -1409,16 +1392,113 @@ type CastHeader struct {
...
+		if err != nil {
+			errs = append(errs, fmt.Errorf("recording: error starting recording: %w", err))
+			continue
+		}
+		req.Header.Set("Expect", "100-continue")
```
catzkorn, line 1465: Is it worth adding a link here to the expect header type, it may be unfamiliar to some users. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect

ssh/tailssh/tailssh.go
```diff
@@ -1409,16 +1392,113 @@ type CastHeader struct {
...
+			if resp.StatusCode != 200 {
+				errChan <- fmt.Errorf("recording: unexpected status: %v", resp.Status)
+				return
+			}
+			errChan <- nil
+		}()
+		select {
+		case <-got100:
+		case err := <-errChan:
+			if err == nil {
```
catzkorn, line 1484: Can we document why this is a failure case? It looks like we expect the request to block and if we get a nil it's because the request succeeded when it shouldn't, but that wasn't immediately obvious to me.

tailcfg/tailcfg.go
```diff
@@ -2048,7 +2049,59 @@ type SSHAction struct {
...
+
+// SSHRecorderFailureAction is the action to take if recording fails.
+type SSHRecorderFailureAction struct {
+	// RejectSession, if true, specifies that the session should
+	// be rejected if the recording fails to start.
+	RejectSession bool `json:",omitempty"`
+
+	// TerminateSession, if true, specifies that the session should
+	// be terminated if the recording fails after it has started.
+	TerminateSession bool `json:",omitempty"`
```
catzkorn, line 2067: I think these two fields could be more explicitly named, but I don't have better ideas that aren't super long in the few minutes of noodling i've been doing.

ssh/tailssh/tailssh.go
```diff
@@ -1409,16 +1392,113 @@ type CastHeader struct {
...
+			}
+			if resp.StatusCode != 200 {
+				errChan <- fmt.Errorf("recording: unexpected status: %v", resp.Status)
+				return
+			}
+			errChan <- nil
+		}()
+		select {
+		case <-got100:
+		case err := <-errChan:
```
twitchyliquid64, line 1483: Im having a bit of trouble understanding `errChan`, maybe we can add comments?

It seems to be doing two roles:
1. Detect when the `Do()` fails and we need to quickly try another recorder candidate, and
2. Propergate that the streaming has suddenly failed and recording should stahp

maisem: there are comments on the func, added more in the `select`

ssh/tailssh/tailssh.go
```diff
@@ -1086,7 +1066,9 @@ func (ss *sshSession) run() {
 				ss.Exit(1)
 				return
 			}
-			defer rec.Close()
+			if rec != nil {
+				defer rec.Close()
```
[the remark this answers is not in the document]
maisem: this is not an error case

ssh/tailssh/tailssh.go
```diff
@@ -1409,16 +1392,113 @@ type CastHeader struct {
...
 }
 
+// sessionRecordingClient returns an http.Client that uses srv.lb.Dialer() to
+// dial connections. This is used to make requests to the session recording
+// server to upload session recordings.
+// It uses the provided dialCtx to dial connections, and limits a single dial
+// to 5 seconds.
+func (ss *sshSession) sessionRecordingClient(dialCtx context.Context) (*http.Client, error) {
+	dialer := ss.conn.srv.lb.Dialer()
+	if dialer == nil {
```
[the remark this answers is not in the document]
maisem: it can't be nil today, and if the assumption changes tests will blow up.

ssh/tailssh/tailssh.go
maisem: don't need to in simple cases

ssh/tailssh/tailssh.go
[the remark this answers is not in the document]
maisem: the case is documented on `startNewRecording`

ssh/tailssh/tailssh.go
maisem: done

ssh/tailssh/tailssh.go
```diff
@@ -1436,51 +1516,43 @@ func (ss *sshSession) startNewRecording() (_ *recording, err error) {
...
-		hc := ss.conn.srv.sessionRecordingClient()
-		resp, err := hc.Do(req)
-		if err != nil {
-			err := fmt.Errorf("recording: error sending recording: %w", err)
-			ss.logf("%v", err)
-			ss.cancelCtx(userVisibleError{
-				msg:   "recording: error sending recording",
-				error: err,
-			})
+		err := <-errChan
```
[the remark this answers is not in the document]
maisem: the nil doesn't get here unless its a success

tailcfg/tailcfg.go
maisem: eh, docs are clear no?

tailcfg/tailcfg.go
```diff
@@ -2048,7 +2049,59 @@ type SSHAction struct {
...
+
+	// OnRecorderFailure is the action to take if recording fails.
+	// If nil, the default action is to fail open.
+	OnRecordingFailure *SSHRecorderFailureAction `json:"onRecordingFailure,omitempty"`
+}
+
+// SSHRecorderFailureAction is the action to take if recording fails.
+type SSHRecorderFailureAction struct {
+	// RejectSession, if not empty, specifies that the session should
+	// be rejected if the recording fails to start.
```
twitchyliquid64, line 2062: It may be obvious but specify when this message will be sent, and to where

tailcfg/tailcfg.go
```diff
@@ -2048,7 +2049,59 @@ type SSHAction struct {
...
+	RejectSessionWithMessage string `json:",omitempty"`
+
+	// TerminateSession, if not empty, specifies that the session should
+	// be terminated if the recording fails after it has started.
+	TerminateSessionWithMessage string `json:",omitempty"`
```
twitchyliquid64, line 2067: ditto

twitchyliquid64 (approved):

PR #7940, opened 2023-04-21
wgengine/magicsock: add hysteresis to endpoint selection

andrew-d:
Avoid selecting an endpoint as "better" than the current endpoint if the total latency improvement is less than 1%. This adds some hysteresis to avoid flapping between endpoints for a minimal improvement in latency.

Change-Id: If8312e1768ea65c4b4d4e13d8de284b3825d7a73

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -4921,6 +4921,28 @@ func betterAddr(a, b addrLatency) bool {
 			return false
 		}
 	}
+
+	// If we get here, then both addresses are the same IP type (i.e. both
+	// IPv4 or both IPv6). All decisions below are made solely on latency.
+	//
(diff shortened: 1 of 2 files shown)


wgengine/magicsock/magicsock.go
```diff
@@ -4920,7 +4920,25 @@ func betterAddr(a, b addrLatency) bool {
 			return false
 		}
 	}
-	return a.latency < b.latency
+
+	// If a is higher-latency than b, then it's not better.
+	if a.latency >= b.latency {
```
maisem, line 4925: This assumes `a` is the current selected best?
What if this difference is less than 1%?

maisem (approved):

maisem (approved):

Issue #7750, opened 2023-03-31
Update `tailscale up --help` to recommend --authkey instead of --auth-key


samlinville:
When you look at `tailscale up --help`, we say that the proper auth key flag is `--auth-key`. However, all of our docs specify `--authkey`, and the CLI will take either.

Based on https://github.com/tailscale/tailscale/issues/6321, it seems the direction we're moving is to use `$TS_AUTHKEY` and `--authkey` over `$TS_AUTH_KEY` and `--auth-key`.

Thus, we should update the output of `tailscale up --help` to use `--authkey`

DentonGentry:
This is harder than it looks. The compatibility handling to accept either `--authkey` or `--auth-key` is https://github.com/tailscale/tailscale/blob/38bbb30aaf8fbf2452cda2054bfff45b2d1cb418/cmd/tailscale/cli/cli.go#L61

It rewrites an `--authkey` argument to be `--auth-key`, meaning that the actual argument is `--authkey`. It isn't that the helptext is wrong: the helptext correctly describes the actual argument as being `--auth-key`.

bradfitz:
The intention was to move to `--auth-key` as canonical and make `--authkey` be the legacy way.

uhthomas:
> The intention was to move to `--auth-key` as canonical and make `--authkey` be the legacy way.

Ah, I see. So should this issue be closed then as the current behaviour of `--help` is correct?

bradfitz:
I don't know what's up with this issue. We originally had only `--authkey` and changed it to `--auth-key` on purpose to migrate to `--auth-key` as canonical and keeping the old alias but apparently there's now discussion of reverting that decision? I'm not current on whatever's being discussed.

uhthomas:
I agree, there seems to be conflicting information on what the intention here is. Though I do recognise the environment variable `TS_AUTH_KEY` has been replaced with the canonical form `TS_AUTHKEY` which does conflict with the `--auth-key` flag semantically.

samlinville:
@uhthomas Let me get some alignment internally on what we want to move forward with and I'll post an update here.

samlinville:
@uhthomas I apologize for the churn on this, but we're going to go with --auth-key and TS_AUTH_KEY.

uhthomas:
> @uhthomas I apologize for the churn on this, but we're going to go with --auth-key and TS_AUTH_KEY.

[...]

PR #7941, opened 2023-04-21
use canonical authkey flag

uhthomas:
The flag "--auth-key" is changed to the canonical form "--authkey", and the compatibility code has been flipped to rewrite usages of "--auth-key" to "--authkey".

Fixes: https://github.com/tailscale/tailscale/issues/7750


DentonGentry:
It is not at all clear to me that this is the right thing to do, given saved configs out there.

uhthomas:
> It is not at all clear to me that this is the right thing to do, given saved configs out there.

I think that's a fair comment, though given both `authkey` and `auth-key` are still accepted this shouldn't have any effect on existing configs, right?

bradfitz:
I don't think we want this. See this comment: https://github.com/tailscale/tailscale/issues/7750#issuecomment-1518260341

PR #7942, opened 2023-04-21
net/dnscache: use Go DNS resolver on Windows

uhthomas:
Go 1.19 introduced Resolver.PreferGo which allows the Go resolver to be used on Windows.

See https://github.com/golang/go/issues/33097.

Fixes https://github.com/tailscale/tailscale/issues/5161

### DIFF
--- a/net/dnscache/dnscache.go
+++ b/net/dnscache/dnscache.go
@@ -31,19 +31,19 @@ import (
 var zaddr netip.Addr
 
 var single = &Resolver{
-	Forward: &net.Resolver{PreferGo: preferGoResolver()},
+	Forward: &net.Resolver{PreferGo: preferGoResolver(runtime.GOOS)},
 }
 
-func preferGoResolver() bool {
+func preferGoResolver(goos string) bool {
 	// There does not appear to be a local resolver running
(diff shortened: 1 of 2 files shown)


net/dnscache/dnscache.go
```diff
@@ -31,19 +31,19 @@ import (
 var zaddr netip.Addr
 
 var single = &Resolver{
-	Forward: &net.Resolver{PreferGo: preferGoResolver()},
+	Forward: &net.Resolver{PreferGo: preferGoResolver(runtime.GOOS)},
```
bradfitz, line 34: We can't change this one. Setting PreferGoResolver to true on Windows isn't good unless you're also overriding the dialer which this isn't.

bradfitz (changes requested):

net/dnscache/dnscache.go
uhthomas: The behaviour here shouldn't have changed. Was this comment intended for a different line?

bradfitz: The bug you referenced (#5161) is a big project. It's not as simple as flipping a bool. This change doesn't look correct. Did you even test it?

PR #7950, opened 2023-04-23
net/netcheck: fix crash when IPv6 kinda but not really works

### LINKED ISSUE #7949: Running netcheck results in a runtime error: invalid memory address or nil pointer dereference
### What is the issue?

When running tailscale netcheck in cmd, the following error occurs:
```
[signal 0xc0000005 code=0x0 addr=0x20 pc=0x93c5b3]

goroutine 50 [running]:
tailscale.com/net/netcheck.(*reportState).runProbe(0xc00027c000, {0xbc9690, 0xc000273810}, 0x0?, {0x0, {0xc00020a570, 0x2}, 0x1, 0x0}, 0xc000228420)
        tailscale.com@v1.1.1-0.20230422043621-e3b2250e2654/net/netcheck/netcheck.go:1569 +0x653
created by tailscale.com/net/netcheck.(*Client).GetReport.func6
        tailscale.com@v1.1.1-0.20230422043621-e3b2250e2654/net/netcheck/netcheck.go:998 +0x76
```

### Steps to reproduce

Open cmd.exe, run the command ``tailscale netcheck``, it fails.

This is a fairly new install, I didn't make any changes after install as far as I recall.

### OS

Windows

### OS version

Win10 22H2

### Tailscale version

1.39.194

### Other software

I'm running NordVPN which might potentially be causing the issue. I added tailscaled.exe, tailscale.exe and the Tailscale GUI client to the split tunneling list to try and avoid issues.

### Bug report

BUG-4fab5752eefc5f1b4358829341a442f10e86e1519ee9cca7f78d398eb07048ca-20230422212303Z-48106b53f4cbf11b

bradfitz:
Looks like on some systems there's an IPv6 address, but then opening a IPv6 UDP socket fails later. Probably some firewall. Tolerate it better and don't crash.

To repro: check the "udp6" to something like "udp7" (something that'll fail) and run "go run ./cmd/tailscale netcheck" on a machine with active IPv6. It used to crash and now it doesn't.

Fixes #7949

### DIFF
--- a/net/netcheck/netcheck.go
+++ b/net/netcheck/netcheck.go
@@ -40,6 +40,7 @@ import (
 	"tailscale.com/types/logger"
 	"tailscale.com/types/nettype"
 	"tailscale.com/types/opt"
+	"tailscale.com/types/ptr"
 	"tailscale.com/util/clientmetric"
 	"tailscale.com/util/mak"
 )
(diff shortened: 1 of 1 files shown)


bradfitz:
cc @andrew-d @maisem

maisem (approved):

DentonGentry (approved):

PR #7960, opened 2023-04-24
net/packet: add NAT support for DCCP and GRE

maisem:
Updates tailscale/corp#8020


net/packet/packet_test.go
```diff
@@ -92,6 +106,9 @@ func TestHeaderChecksums(t *testing.T) {
 			if got != want {
 				t.Fatalf("got %x want %x", got, want)
 			}
+			if tt.name == "SCTP" {
+				fmt.Printf("%x\n", tt.packet)
```
maisem, line 110: Ah debug output. Will remove

nyghtowl (approved):

jwhited (approved):

PR #7969, opened 2023-04-26
ssh/tailssh: also handle recording upload failure during writes

maisem:
Previously we would error out when the recording server disappeared after the in memory buffer filled up for the io.Copy. This makes it so that we handle failing open correctly in that path.

Updates tailscale/corp#9967

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -1519,8 +1519,9 @@ func (ss *sshSession) startNewRecording() (_ *recording, err error) {
 
 	now := time.Now()
 	rec := &recording{
-		ss:    ss,
-		start: now,
+		ss:       ss,
+		start:    now,
+		failOpen: onFailure == nil || onFailure.TerminateSessionWithMessage == "",
 	}
 
 	// We want to use a background context for uploading and not ss.ctx.
@@ -1611,6 +1612,10 @@ type recording struct {
 	ss    *sshSession
 	start time.Time
 
+	// failOpen specifies whether the session should be allowed to
+	// continue if writing to the recording fails.
+	failOpen bool
+
 	mu  sync.Mutex // guards writes to, close of out
 	out io.WriteCloser
 }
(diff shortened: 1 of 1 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -1519,8 +1519,9 @@ func (ss *sshSession) startNewRecording() (_ *recording, err error) {
 
 	now := time.Now()
 	rec := &recording{
-		ss:    ss,
-		start: now,
+		ss:       ss,
+		start:    now,
+		failOpen: onFailure == nil || onFailure.TerminateSessionWithMessage == "",
```
twitchyliquid64, line 1524: Why this second conditional? Can we make a nil callback be how fail open is communicated?

maisem: Nil callback is useful for when the recording fails to start. This is for when it starts and then fails.

ssh/tailssh/tailssh.go
```diff
@@ -1642,7 +1647,7 @@ func (r *recording) writer(dir string, w io.Writer) io.Writer {
 		// passwords.
 		return w
 	}
-	return &loggingWriter{r, dir, w}
+	return &loggingWriter{r, dir, w, false}
```
catzkorn, line 1650: Can you name these 😅

catzkorn (approved):

ssh/tailssh/tailssh.go

ssh/tailssh/tailssh.go
```diff
@@ -1651,20 +1656,29 @@ type loggingWriter struct {
 	r   *recording
 	dir string    // "i" or "o" (input or output)
 	w   io.Writer // underlying Writer, after writing to r.out
+
+	// recordingFailed is whether we should skip recording.
+	// It is set to true if we fail to write to r.out and r.failOpen is set.
+	recordingFailed bool
```
creachadair, line 1662: Minor thing: Maybe `recordingFailedOpen` just to reinforce that this means not only the failure but also that we checked openness?

maisem: good point. done

PR #7971, opened 2023-04-26
net/ping,netcheck: add v6 pinging capabilities to pinger

catzkorn:
This change adds a v6conn to the pinger to enable sending pings to v6 addrs and handling the responses.


net/ping/ping_test.go
```diff
@@ -58,7 +62,62 @@ func TestPinger(t *testing.T) {
...
+		}
+	case <-ctx.Done():
+		t.Fatal("did not get response by timeout")
+	}
+}
+
+func TestV6Pinger(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skipf("IPv6")
+	}
```
catzkorn, line 81: Is this the correct thing to do?

net/netcheck/netcheck.go
```diff
@@ -1321,9 +1321,9 @@ func (c *Client) measureAllICMPLatency(ctx context.Context, rs *reportState, nee
 	ctx, done := context.WithTimeout(ctx, icmpProbeTimeout)
 	defer done()
 
-	p, err := ping.New(ctx, c.logf, c.NetMon)
-	if err != nil {
-		return err
+	p := ping.New(ctx, c.logf, c.NetMon)
+	if p == nil {
```
raggi, line 1325: you can drop the nil check if you want, a panic is fine in this exceptional case

net/ping/ping.go
```diff
@@ -38,7 +42,15 @@ type outstanding struct {
 // A new instance should be created for each concurrent set of ping requests;
 // this type should not be reused.
 type Pinger struct {
-	c       net.PacketConn
+	// ListenPacketForTesting is used to create the underlying net.PacketConn.
+	// If nil, netns.Listener is used.
+	// This only exists for testing.
+	ListenPacketForTesting func(ctx context.Context, typ, addr string) (net.PacketConn, error)
```
raggi, line 48: We could decouple this package from `netns` and from `netmon` if we do this:

```
type ListenPacketer interface {
  ListenPacket(context.Context, string, string) (PacketConn, error)
}

func New(ctx context.Context, logf logger.Logf, lp ListenPacketer) ...
```

It moves a little burden in to the caller, but now they can pick from `net.ListenConfig`, `netns.Listener` and so on, and unit tests can likely avoid both netns and netmon entirely.

net/ping/ping_test.go
raggi: It's an ok match for our CI, but I think we could do a quick probe instead such as:

```
if c, err := net.ListenPacket("udp6", "::1"); err != nil {
  t.Skipf("IPv6 not supported: %s", err)
} else {
  c.Close()
}
```

raggi (changes requested):
Looks good, the two things are:

- There's a race in getv4conn and getv6conn, it looks like those can be moved under the mutex, and the atomics then dropped, to solve the race there.
- Consider adjusting New to take a ListenPacketer to decouple the package from netmon and netns. Feel free to push back on this if you feel the burden on the callers isn't worthwhile.

raggi (approved):

PR #7974, opened 2023-04-26
.github/workflows: split tests and benchmarks for caching

raggi:
Benchmark flags prevent test caching, so benchmarks are now executed
independently of tests.

Fixes https://github.com/tailscale/tailscale/issues/7975

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,7 +89,11 @@ jobs:
     - name: build test wrapper
       run: ./tool/go build -o /tmp/testwrapper ./cmd/testwrapper
     - name: test all
-      run: ./tool/go test ${{matrix.buildflags}} -exec=/tmp/testwrapper -bench=. -benchtime=1x ./...
+      run: ./tool/go test ${{matrix.buildflags}} -exec=/tmp/testwrapper
+      env:
+        GOARCH: ${{ matrix.goarch }}
+    - name: bench all
+      run: ./tool/go test ${{matrix.buildflags}} -exec=/tmp/testwrapper -test.bench=. -test.benchtime=1x -test.run=^$
       env:
         GOARCH: ${{ matrix.goarch }}
     - name: check that no tracked files changed


.github/workflows/test.yml
```diff
@@ -89,7 +90,11 @@ jobs:
     - name: build test wrapper
       run: ./tool/go build -o /tmp/testwrapper ./cmd/testwrapper
     - name: test all
-      run: ./tool/go test ${{matrix.buildflags}} -exec=/tmp/testwrapper -bench=. -benchtime=1x ./...
+      run: ./tool/go test ${{matrix.buildflags}} -exec=/tmp/testwrapper
```
bradfitz, line 93: I wonder if `-exec` also disables test caching?

raggi: good question, I'll put debug back and check!

bradfitz:
I would approve the benchmark split but I don't know what's going on with the rest of your change. Is it related?

raggi:
I've split the cache key out into a separate commit so we can do that in another PR

.github/workflows/test.yml
```diff
@@ -88,8 +88,15 @@ jobs:
         sudo apt-get -y install qemu-user
     - name: build test wrapper
       run: ./tool/go build -o /tmp/testwrapper ./cmd/testwrapper
+    - name: build test wrapper
+      run: ./tool/go build -o /tmp/testwrapper ./cmd/testwrapper
```
bradfitz, line 92: I think building it once is enough :)

.github/workflows/test.yml
```diff
@@ -88,8 +88,15 @@ jobs:
...
     - name: test all
-      run: ./tool/go test ${{matrix.buildflags}} -exec=/tmp/testwrapper -bench=. -benchtime=1x ./...
+      run: ./tool/go test ${{matrix.buildflags}} -exec=/tmp/testwrapper
+      env:
+        GOARCH: ${{ matrix.goarch }}
+        GODEBUG: gocachetest=1
```
bradfitz, line 97: leftover debug?

.github/workflows/test.yml
raggi: Whoops, yeah, copypasta fail.

.github/workflows/test.yml
raggi: yeah, I was just investigating from the question above, but now did that locally. It seems exec does affect test caching, so that'll be another step to look into.

Issue #7982, opened 2023-04-26
cmd/tailscale: let tailscale up accept oauth creds


bradfitz:
It should be possible to use `tailscale up --auth-key=` with oauth creds.

shubb30:
Looks like this has already done in #8403.

DentonGentry:
Duplicate of https://github.com/tailscale/tailscale/issues/8403
Fixed in 1.46

PR #7983, opened 2023-04-26
cmd/tailscale/cli: [up] add experimental oauth2 authkey support

bradfitz:
Updates #7982


cmd/tailscale/cli/up.go
```diff
@@ -1102,3 +1111,96 @@ func anyPeerAdvertisingRoutes(st *ipnstate.Status) bool {
...
+		if v == "" {
+			return def, nil
+		}
+		ret, err := strconv.ParseBool(v)
+		if err != nil {
+			return false, fmt.Errorf("invalid attribute boolean attribute %s value %q", name, v)
+		}
+		return ret, nil
+	}
+	ephemeral, err := getBool("ephemeral", false)
```
bradfitz, line 1156: maybe this should default to true, @DentonGentry?

cmd/tailscale/cli/up.go
```diff
@@ -1102,3 +1111,96 @@ func anyPeerAdvertisingRoutes(st *ipnstate.Status) bool {
...
+	if err != nil {
+		return "", err
+	}
+	preauth, err := getBool("preauthorized", false)
```
bradfitz, line 1160: same?

cmd/tailscale/cli/up.go
```diff
@@ -1102,3 +1111,96 @@ func anyPeerAdvertisingRoutes(st *ipnstate.Status) bool {
...
+//
+// If the BASE_URL argument is not provided, it defaults to https://api.tailscale.com.
+func resolveAuthKey(ctx context.Context, v string) (string, error) {
+	suff, ok := strings.CutPrefix(v, "oauth:")
+	if !ok {
+		return v, nil
+	}
+	if !envknob.Bool("TS_EXPERIMENT_OAUTH_AUTHKEY") {
+		return "", errors.New("oauth authkeys are in experimental status")
+	}
```
bradfitz, line 1131: I figured this lets us get this into 1.40.0 with minimal testing. We can use it ourselves in this release and remove this limitation later.

cmd/tailscale/cli/up.go
DentonGentry: That seems like a safer default if one is automatically generating authkeys. Otherwise people spin up a couple thousand nodes and only figure it out later, then ask us for help in deleting them.

cmd/tailscale/cli/up.go
DentonGentry: What do you think about a `tailscale debug get-authkey` to get an authkey in one part of the script and pass it to `tailscale up --authkey=...` later?
 
I bring it up because one is likely to run `tailscale up` in the same Actions step where `tailscaled` is running, which means the client-id and secret are sitting in environment variables through the whole run.

With a `tailscale debug get-authkey`, the secrets would only be passed to an Action step which outputs the authkey string, they wouldn't be present in the step where tailscaled runs.

cmd/tailscale/cli/up.go
DentonGentry: Personally, I'd leave this one false. If people have device authorization turned on, they can make the choice whether to preauthorize.

For github runners they clearly would preauthorize as it would work badly without it, but other uses might want the device to sit there until someone decides it can proceed.

cmd/tailscale/cli/up.go
```diff
@@ -663,6 +668,10 @@ func runUp(ctx context.Context, cmd string, args []string, upArgs upArgsT) (retE
 		if err != nil {
 			return err
 		}
+		authKey, err = resolveAuthKey(ctx, authKey)
+		if err != nil {
+			return err
+		}
```
maisem, line 674: this should also work for `tailscale login` not just for `tailscale up`

cmd/tailscale/cli/up.go
bradfitz: Hm. But `tailscale debug get-authkey` isn't where we want to end up CLI-wise, though.

cmd/tailscale/cli/up.go
maisem: oh does it already?

cmd/tailscale/cli/up.go
bradfitz: Why do they need to be in the same `run:` step?

bradfitz: I'm looking at https://github.com/tailscale/github-action/blob/main/action.yml#L57-L68 and can't that be two steps?

cmd/tailscale/cli/up.go
bradfitz: the GitHub action only uses `up` and it's experimental for now, aimed at the GitHub Action primarily.

But sure.

cmd/tailscale/cli/up.go
DentonGentry: We could definitely make our Action do so in two steps.

I'm more thinking of how other people would use it, where it is a very common pattern to:
```
tailscaled &
tailscale up --authkey=...
```

They'd end up with the client-id and secret in environment variables in tailscaled.
Which isn't immediately a problem, just less safe than them not being present.

bradfitz: We could document best practice somewhere?

I'd prefer to keep it part of `up --authkey=` rather than adding a `debug` thing that isn't.

maisem (approved):

PR #7985, opened 2023-04-26
cmd/tailscale/cli: [up] change oauth authkey format

bradfitz:
Updates #7982


willnorris (approved):

maisem:
Tested locally, and it works!

PR #7991, opened 2023-04-27
build_dist.sh: make cross-compilation friendly for env CC specified

ayanamist:
I'm using tailscale on an openwrt device. Openwrt has its own toolchains so i have to specify `CC` env to openwrt toolchains gcc instead of the one of my host os.
However, `build_dist.sh` runs a `$go run ./cmd/mkversion` and it will fail when `CC` is specified.
I review the logic of mkversion and found it can be run using `CGO_ENABLED=0` to avoid `CC` modification.

### DIFF
--- a/build_dist.sh
+++ b/build_dist.sh
@@ -16,7 +16,7 @@ if [ -n "${TS_USE_TOOLCHAIN:-}" ]; then
 	go="./tool/go"
 fi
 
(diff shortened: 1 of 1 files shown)


DentonGentry:
I doubt very much that this is going to be acceptable.

danderson:
Yup, sorry. Was waiting for CI to finish and got distracted. Thanks for the patch!

PR #8001, opened 2023-04-28
.github: test installer script in CI in docker

knyar:
Every time we change `installer.sh`, run it in a few docker containers based on different Linux distros, just as a simple test.

Also includes a few changes to the installer script itself to make installation work in Docker:
- install `dnf config-manager` command before running it
- run `zypper` in non-interactive mode
- update `pacman` indexes before installing packages

Updates https://github.com/tailscale/corp/issues/8952

### DIFF
--- a/.github/workflows/installer.yml
+++ b/.github/workflows/installer.yml
@@ -1,0 +1,102 @@
+name: test installer.sh
+
+on:
+  push:
+    branches:
+      - "main"
+    paths:
+      - scripts/installer.sh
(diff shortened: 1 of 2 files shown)


DentonGentry:
Recommend that the commit message reference https://github.com/tailscale/corp/issues/8952, as github will linkify #8952 as https://github.com/tailscale/tailscale/issues/8952

danderson (approved):

PR #8002, opened 2023-04-28
scripts/installer.sh: check connectivity with pkgs.tailscale.com

knyar:
Installer script relies on pkgs.tailscale.com being reachable, both for checking what Linux distros are supported, but also for actually downloading repo configuration files, gpg keys and packages themselves.

This change adds a simple reachability check which will print an error message when pkgs.tailscale.com is not reachable.

Fixes https://github.com/tailscale/corp/issues/8952

### DIFF
--- a/scripts/installer.sh
+++ b/scripts/installer.sh
@@ -321,6 +321,17 @@ main() {
 		exit 1
 	fi
 
+	TEST_URL="https://pkgs.tailscale.com/"
+	RC=0
+	TEST_OUT=$($CURL "$TEST_URL" 2>&1) || RC=$?
+	if [ $RC != 0 ]; then
+		echo "The installer cannot reach $TEST_URL"
+		echo "Please make sure that your machine has internet access."
+		echo "Test output:"
+		echo $TEST_OUT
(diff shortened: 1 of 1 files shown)


danderson (approved):

scripts/installer.sh
```diff
@@ -321,6 +321,18 @@ main() {
 		exit 1
 	fi
 
+	TEST_URL="https://pkgs.tailscale.com/"
+	RC=0
+	TEST_OUT=$($CURL "$TEST_URL" 2>&1) || RC=$?
+	if [ $RC != 0 ]; then
+		echo "The installer needs to be able to reach Tailscale package server"
```
DentonGentry, line 328: I'd recommend simplifying the structure of this sentence, a number of our users in China rely on machine translation. pkgs.tailscale.com is blocked from China.

Suggest: "The installer cannot reach $TEST_URL, please make sure that your machine has internet access."

knyar: Thanks, I've rephrased the message.

DentonGentry (approved):

Issue #6110, opened 2022-10-28
Tailscale hangs indefinitely on misbehaving DNS


siikanen:
### What is the issue?

Sometimes when I try to go log into tailscale, I get this weird behaviour where running `sudo tailscale up` freezes. Seems like the login flow goes ok until `control: creating new noise client` comes up and then fails with `Received error: register request: Post "https://controlplane.tailscale.com/machine/register": connection attempts aborted by context: context deadline exceeded`

See logs below

```Oct 28 13:51:27 carbon tailscaled[7554]: ipnserver: conn2: connection from userid 0; root has access
Oct 28 13:51:27 carbon tailscaled[7554]: Start
Oct 28 13:51:27 carbon tailscaled[7554]: control: client.Shutdown()
Oct 28 13:51:27 carbon tailscaled[7554]: control: client.Shutdown: inSendStatus=0
Oct 28 13:51:27 carbon tailscaled[7554]: control: mapRoutine: quit
Oct 28 13:51:27 carbon tailscaled[7554]: control: Client.Shutdown done.
Oct 28 13:51:27 carbon tailscaled[7554]: using backend prefs for "_daemon": Prefs{ra=false dns=true want=true routes=[] nf=on Persist=nil}
Oct 28 13:51:27 carbon tailscaled[7554]: Backend: logs: be:495a217121ba33adf5c90d17c7138fb21263e9d8d9900b113a00357307310882 fe:
Oct 28 13:51:27 carbon tailscaled[7554]: Switching ipn state NoState -> NeedsLogin (WantRunning=true, nm=false)
Oct 28 13:51:27 carbon tailscaled[7554]: blockEngineUpdates(true)
Oct 28 13:51:27 carbon tailscaled[7554]: Reconfig(down): no changes made to Engine config
Oct 28 13:51:27 carbon tailscaled[7554]: StartLoginInteractive: url=false
Oct 28 13:51:27 carbon tailscaled[7554]: control: client.Login(false, 2)
Oct 28 13:51:27 carbon tailscaled[7554]: control: LoginInteractive -> regen=true
Oct 28 13:51:27 carbon tailscaled[7554]: control: doLogin(regen=true, hasUrl=false)
Oct 28 13:51:37 carbon tailscaled[7554]: trying bootstrapDNS("derp12c.tailscale.com", "149.28.119.105") for "controlplane.tailscale.com" ...
[...]

Issue #4845, opened 2022-06-14
control/controlclient: noiseClient dial doesn't use DNS bootstrap (dnsfallback) dialer


bradfitz:
When we added the Noise dialer in 1.24.0 we lost the DNS bootstrap (dnsfallback) dialer support.  (found via debugging ticket http://go/hs/2307; details in https://github.com/tailscale/corp/issues/5702)

The Noise client now dow does:

```go
   conn, err := controlhttp.Dial(ctx, nc.serverHost, nc.privKey, nc.serverPubKey, uint16(tailcfg.CurrentCapabilityVersion), nc.dialer.SystemDial)
```

Where `nc.dialer.SystemDial` doesn't have any of the DNS fallback.

That means when somebody has a broken DNS config on their machine we can hit https://controlplane.tailscale.com/key?v=32 over TLS to get the key(s) but then the Noise dial to the same hostname fails, unable to find its IP address(es).

/cc @maisem @danderson

mihaip:
As far as I can tell, this is already happening. The `nc.dialer.SystemDial` dialer is wrapped with a `dnscache.Resolver` that uses `dnsfallback.Lookup`:

https://github.com/tailscale/tailscale/blob/d6b8a18b09d2c24111777d51f79fd63680023438/control/controlhttp/client.go#L196-L206

Empirically it also appears to work (at least if I mess with Mac's DNS settings, the client is still able to connect).

Perhaps the issue that the user was facing was a slow DNS server, and we need to tweak the timeouts? Or we should reuse the `dnscache.Resolver` across calls, so that even if the initial call fails, the next call doesn't have to start from scratch in terms of looking up fallbacks?

DentonGentry:
Barring further comment as it appears this is already implemented, we'll expect to close this during the next regular bug scrub.

DentonGentry:
Believed fixed by https://github.com/tailscale/tailscale/commit/42fd96409023f2d2d8ec0111c18e6be0c4791bad

PR #8008, opened 2023-04-30
control/controlclient: use dnscache.Resolver for Noise client

andrew-d:
This passes the *dnscache.Resolver down from the Direct client into the Noise client and from there into the controlhttp client. This retains the Resolver so that it can share state across calls instead of creating a new resolver.

Updates https://github.com/tailscale/tailscale/issues/4845
Updates https://github.com/tailscale/tailscale/issues/6110

Change-Id: Ia5d6af1870f3b5b5d7dd5685d775dcf300aec7af

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -61,6 +61,7 @@ import (
 type Direct struct {
 	httpc                  *http.Client // HTTP client used to talk to tailcontrol
 	dialer                 *tsdial.Dialer
+	dnsCache               *dnscache.Resolver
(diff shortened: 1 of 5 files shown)


maisem (approved):
Nice cleanup!

andrew-d:
Confirmed on a DigitalOcean droplet that this successfully establishes a connection in [the case reported in #6110](https://github.com/tailscale/tailscale/issues/6110#issuecomment-1528859574), whereas the current `main` build does not.

PR #8014, opened 2023-05-01
ssh/tailssh: improve debug logging around revoked sessions

twitchyliquid64:
Updates https://github.com/tailscale/corp/issues/10943

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -256,6 +256,12 @@ func (c *conn) logf(format string, args ...any) {
 	c.srv.logf(format, args...)
 }
 
+func (c *conn) vlogf(format string, args ...any) {
+	if sshVerboseLogging() {
+		c.logf(format, args...)
+	}
+}
+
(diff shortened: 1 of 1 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -861,6 +861,9 @@ func (c *conn) newSSHSession(s ssh.Session) *sshSession {
 // isStillValid reports whether the conn is still valid.
 func (c *conn) isStillValid() bool {
 	a, localUser, err := c.evaluatePolicy(c.pubKey)
+	if sshVerboseLogging() {
```
maisem, line 864: use `ss.vlogf`

twitchyliquid64: This is on the conn, there can be multiple sessions and those structures are all guarded by Mu

twitchyliquid64: :+1:

maisem (approved):

PR #8023, opened 2023-05-02
ssh/tailssh: restore support for recording locally

maisem:
Updates tailscale/corp#9967

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -67,6 +67,7 @@ type ipnLocalBackend interface {
 	WhoIs(ipp netip.AddrPort) (n *tailcfg.Node, u tailcfg.UserProfile, ok bool)
 	DoNoiseRequest(req *http.Request) (*http.Response, error)
 	Dialer() *tsdial.Dialer
+	TailscaleVarRoot() string
 }
 
 type server struct {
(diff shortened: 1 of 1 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -1154,6 +1155,10 @@ func (ss *sshSession) run() {
 	return
 }
 
+// recordSSHToLocalDisk is a temporary dev knob to test the SSH recording
+// functionality and support off-node streaming.
```
twitchyliquid64, line 1159: > and support off-node streaming

This knob is just for enabling local storage, no?

ssh/tailssh/tailssh.go
```diff
@@ -1510,12 +1515,33 @@ func (ss *sshSession) connectToRecorder(ctx context.Context, recs []netip.AddrPo
...
+}
+
 // startNewRecording starts a new SSH session recording.
 // It may return a nil recording if recording is not available.
 func (ss *sshSession) startNewRecording() (_ *recording, err error) {
 	recorders, onFailure := ss.recorders()
+	var localRecording bool
 	if len(recorders) == 0 {
-		return nil, errors.New("no recorders configured")
+		if recordSSHToLocalDisk() {
```
twitchyliquid64, line 1540: I guess this implies that local recording is only done if remote recording wasnt requested? we should document that somewhere.

ssh/tailssh/tailssh.go
maisem: yep, updated docs

ssh/tailssh/tailssh.go
maisem: updated docs

twitchyliquid64 (approved):

PR #8029, opened 2023-05-03
various: add Unraid detection and direct Taildrop support

### LINKED ISSUE #8025: FR: Unraid: Detection and Taildrop Behavior

Tailscale should use the "NAS behavior" when running on Unraid, placing files into a designated folder instead of being held for pickup with the tailscale CLI.

### How should we solve this?

- Add ability for Tailscale to detect Unraid as a platform (/etc/unraid-version)
- Add configuration option to select Taildrop destination when running on Unraid
- Disable Taildrop on Unraid if a valid destination is not set.

The community Unraid plugin currently has to "simulate" being a QNAP NAS in order for Taildrop to use the NAS-style functionality (i.e., placing taildropped files directly into a folder on the device). This limits further development of the plugin, however. since other QNAP-specific behavior does not work properly with Unraid (e.g., `tailscale web` cannot be used since it invokes QNAP authentication). 

### Anything else?

The community plugin is located at https://github.com/dkaser/unraid-tailscale

Adding detection of Unraid to Tailscale will permit further integration between Tailscale and the Unraid web UI. Once Unraid can be detected, additional FRs/PRs can be created for functionality that cannot currently be implemented within the plugin.

dkaser:
The community Unraid plugin currently has to "simulate" being a QNAP NAS in order for Taildrop to use the NAS-style functionality (i.e., placing taildropped files directly into a folder on the device). This change adds Unraid as a detected distro and enables direct file mode.

Closes #8025

### DIFF
--- a/cmd/tailscaled/taildrop.go
+++ b/cmd/tailscaled/taildrop.go
@@ -18,7 +18,7 @@ import (
 func configureTaildrop(logf logger.Logf, lb *ipnlocal.LocalBackend) {
 	dg := distro.Get()
 	switch dg {
-	case distro.Synology, distro.TrueNAS, distro.QNAP:
+	case distro.Synology, distro.TrueNAS, distro.QNAP, distro.Unraid:
 		// See if they have a "Taildrop" share.
 		// See https://github.com/tailscale/tailscale/issues/2179#issuecomment-982821319
 		path, err := findTaildropDir(dg)
@@ -42,6 +42,8 @@ func findTaildropDir(dg distro.Distro) (string, error) {
 		return findTrueNASTaildropDir(name)
 	case distro.QNAP:
 		return findQnapTaildropDir(name)
+	case distro.Unraid:
+		return findUnraidTaildropDir(name)
 	}
 	return "", fmt.Errorf("%s is an unsupported distro for Taildrop dir", dg)
 }
@@ -100,6 +102,28 @@ func findQnapTaildropDir(name string) (string, error) {
 	}
 	if fi, err = os.Stat(fullpath); err == nil && fi.IsDir() {
 		return dir, nil // return the symlink, how QNAP set it up
+	}
+	return "", fmt.Errorf("shared folder %q not found", name)
+}
+
+// findUnraidTaildropDir looks for a directory linked at
+// /var/lib/tailscale/Taildrop. This is a symlink to the
+// path specified by the user in the Unraid Web UI
+func findUnraidTaildropDir(name string) (string, error) {
+	dir := fmt.Sprintf("/var/lib/tailscale/%s", name)
+	_, err := os.Stat(dir)
+	if err != nil {
+		return "", fmt.Errorf("symlink %q not found", name)
+	}
+
+	fullpath, err := filepath.EvalSymlinks(dir)
+	if err != nil {
+		return "", fmt.Errorf("symlink %q to shared folder not valid", name)
+	}
+
+	fi, err := os.Stat(fullpath)
+	if err == nil && fi.IsDir() {
+		return dir, nil // return the symlink
 	}
 	return "", fmt.Errorf("shared folder %q not found", name)
 }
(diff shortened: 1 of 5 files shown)


cmd/tailscaled/taildrop.go
```diff
@@ -103,3 +105,23 @@ func findQnapTaildropDir(name string) (string, error) {
 	}
 	return "", fmt.Errorf("shared folder %q not found", name)
 }
+
+// findUnraidTaildropDir looks for a directory linked at
+// /var/lib/tailscale/Taildrop. This is a symlink to the
+// patch specified by the user in the Unraid Web UI
```
DentonGentry, line 111: :s/patch/path/

ipn/ipnlocal/peerapi.go
```diff
@@ -1090,6 +1091,10 @@ func (h *peerAPIHandler) handlePeerPut(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, errNoTaildrop.Error(), http.StatusInternalServerError)
 		return
 	}
+	if ((distro.Get() == distro.Unraid) && (!h.ps.directFileMode)) {
```
DentonGentry, line 1094: I don't think this should be here: if they haven't configured a shared Taildrop folder, that shouldn't turn into a mysterious internal server error.

dkaser: I'm not sure what the best behavior here is. The fundamental problem is that Unraid boots from a flash drive and runs from an in-memory image. There just isn't a good place to cache files for retrieval if a folder isn't configured, so the best option (to me) seems to be that any attempt to drop files should fail until a path is configured/available.

cmd/tailscaled/taildrop.go
dkaser: Fixed.

ipn/ipnlocal/peerapi.go
```diff
@@ -1090,6 +1091,10 @@ func (h *peerAPIHandler) handlePeerPut(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, errNoTaildrop.Error(), http.StatusInternalServerError)
 		return
 	}
+	if (distro.Get() == distro.Unraid) && (!h.ps.directFileMode) {
```
bradfitz, line 1094: no need for the parens

cmd/tailscaled/taildrop.go
```diff
@@ -103,3 +105,25 @@ func findQnapTaildropDir(name string) (string, error) {
...
+	fullpath, err := filepath.EvalSymlinks(dir)
+	if err != nil {
+		return "", fmt.Errorf("symlink to shared folder %q not found", name)
+	}
+
+	fi, err := os.Stat(fullpath)
+	if err == nil && fi.IsDir() {
+		return dir, nil // return the symlink
+	}
+	return "", fmt.Errorf("shared folder %q not found", name)
```
bradfitz, line 128: make this error message text different from 116's somewhat so we can tell them apart when debugging?

dkaser: I adjusted the messages.

ipn/ipnlocal/peerapi.go
dkaser: I updated the error message for this... attempting to Taildrop when the folder is not configured will still result in an internal server error, but the error message should be less mysterious now.

dkaser:
Changes incorporated and squashed.

ipn/ipnlocal/peerapi.go
DentonGentry: Mainly: I would like to not get support tickets that Unraid taildrop isn't working.

In tailscale/health is a health check package, which is output in `tailscale status` and surfaced to the support team. If you can add a health check warning that taildrop won't work because no shared volume has been set up, that will give us something to work on tickets involving Unraid.

dkaser: That makes sense.

The logical place to put the health check is right next to this log:

https://github.com/dkaser/tailscale/blob/687f3872ff4820f4d8907600063d78137012ca79/cmd/tailscaled/taildrop.go#L26

With that being the case, a question -- I can limit the warnable to Unraid, but it seems like it would be beneficial to set for the other NAS platforms too. Which would you prefer?

bradfitz:
Thanks!

Issue #6102, opened 2022-10-28
FR: VNC via Tailscale in the browser


avidrissman:
### What are you trying to do?

I ran across #3157 via a mention in a podcast, where Brad was talking about the possibility of doing VNC via Tailscale in the browser. #3157 was just closed with you shipping SSH in the browser (congrats!) but a version of Tailscale where you can one-click VNC like you one-click SSH would be a dream.

### How should we solve this?

Integrate https://github.com/novnc/noVNC on top of the existing Tailscale-in-the-browser code you implemented for SSH?

### What is the impact of not solving this?

I will have to continue to pine for this feature.

Ivan-Firefly:
Great request, also waiting for VNC integration.

conblem:
I am currently implementing this and already have a working prototype stay tuned.

conblem:
@avidrissman @Ivan-Firefly I create a pull request which will allow for discussion about adding tcp support to the tsconnect package. This is needed for VNC support. In the meantime I'll try to upload my local vnc prototype in the next few days.

conblem:
If anyone is interested here is the prototype which supports vnc in the browser: [tailvnc](https://github.com/conblem/tailvnc). I hope the readme is useful enough, have a great weekend everybody!

nblythe:
Cloudflare supports in-browser VNC. It would be great if Tailscale supported it, too, so we could use Tailscale on certain projects where clientless VNC is required, instead of using Cloudflare.

PR #8047, opened 2023-05-04
Add direct tcp support to tsconnect wasm

conblem:
To support VNC #6102 I would like to add direct tcp support to the tsconnect package. Feedback would be greatly appreciated! This is just a prototype for the moment as i'd like some feedback first. Most importantly are you willing to add direct tcp support to the package at all? Cheers thanks for the great product!


bradfitz:
I think this is a primitive that's needed for VNC support later to use.

> a browser generally doesn't allow JavaScript to open raw TCP sessions.

Well, we open a TCP connection to SSH servers for Tailscale SSH. I think this just exposes that step as its own explicit step.

I haven't read this closely yet.

conblem:
Correct but because this uses the tailscale derp tunnel it makes it possible to open a raw tcp connection. It's pretty similiar to the SSH feature of the tsconnect package. Obviously browsers don't allow creating SSH connections directly but the derp tunnel makes it possible. This allows the same for raw tcp connections. I used this to implement VNC directly in the browser using the novnc package.

conblem:
@bradfitz yes that's exactly it

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -323,6 +343,120 @@ func (i *jsIPN) logout() {
...
+}
+
+type jsTCPSession struct {
+	jsIPN       *jsIPN
+	conn        net.Conn
+	writeBuffer []byte
+	readBuffer  []byte
+}
+
+func (i *jsIPN) NewJsTCPSession(addr string, config js.Value) (*jsTCPSession, error) {
```
bradfitz, line 373: NewJSTCP ...

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -323,6 +343,120 @@ func (i *jsIPN) logout() {
 	go i.lb.Logout()
 }
 
+func (i *jsIPN) tcp(host string, port int, config js.Value) interface{} {
```
bradfitz, line 346: we should probably expose the timeout in this signature.

also: document what "config" is.

bradfitz, line 346: "any" instead of "interface{}"

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -323,6 +343,120 @@ func (i *jsIPN) logout() {
...
+	writeBuffer []byte
+	readBuffer  []byte
+}
+
+func (i *jsIPN) NewJsTCPSession(addr string, config js.Value) (*jsTCPSession, error) {
+	if config.Type() != js.TypeFunction {
+		return nil, errors.New("Invalid callback received")
+	}
+
+	const timeoutSeconds time.Duration = 5
```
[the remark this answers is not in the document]
conblem: Ah my bad wrong variable name but I do multiply it later at line 379 will fix this. The code atm tho is not really polished overall as I wanted some feedback first, if this would even be wanted in the package.

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -323,6 +343,120 @@ func (i *jsIPN) logout() {
...
+	s := &jsTCPSession{
+		jsIPN:       i,
+		conn:        conn,
+		writeBuffer: make([]byte, 1024),
+		readBuffer:  make([]byte, 4*1024*1024),
+	}
+
+	go func() {
+		s.readLoop(config)
+	}()
```
bradfitz, line 396: you can write just `go s.readLoop(config)`

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -323,6 +343,120 @@ func (i *jsIPN) logout() {
...
+
+	// we always reuse the same dst but make use of subarrays
+	subarray := dst.Call("subarray", 0, len(s.readBuffer))
+	subarrayLength := len(s.readBuffer)
+	for {
+		n, err := s.conn.Read(s.readBuffer[:])
+		if err != nil {
+			if err == io.EOF {
+				// TODO: print probably not needed
+				log.Printf("eof")
```
bradfitz, line 413: probably not

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -323,6 +343,120 @@ func (i *jsIPN) logout() {
...
+			subarray = dst.Call("subarray", 0, n)
+		}
+
+		js.CopyBytesToJS(subarray, s.readBuffer[:n])
+
+		readCallback.Invoke(subarray)
+	}
+}
+
+func (s *jsTCPSession) jsWrapper() interface{} {
```
bradfitz, line 432: any

conblem:
So I implemented the review findings and signed off my commits.

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -324,6 +349,117 @@ func (i *jsIPN) logout() {
...
+	conn, err := i.dialer.UserDial(ctx, "tcp", addr)
+	if err != nil {
+		return nil, err
+	}
+
+	s := &jsTCPSession{
+		jsIPN:       i,
+		conn:        conn,
+		writeBuffer: make([]byte, 1024),
+		readBuffer:  make([]byte, 4*1024*1024),
```
conblem, line 398: maybe the size should be configurable same for the writeBuffer

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -324,6 +349,117 @@ func (i *jsIPN) logout() {
...
+					return s.write(dst, byteLength.Int())
+				}
+				return nil, errors.New("Invalid type has no byteLength")
+			})
+		}),
+	}
+}
+
+func (s *jsTCPSession) write(dst js.Value, length int) (int, error) {
+	if len(s.writeBuffer) < length {
```
conblem, line 456: Not sure If this behaviour is the best way. At the moment we are just increasing the size of the buffer if the js caller provides a bigger ArrayBuffer

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -290,7 +315,7 @@ func (i *jsIPN) run(jsCallbacks js.Value) {
 		err := i.lb.Start(ipn.Options{
 			UpdatePrefs: &ipn.Prefs{
 				ControlURL:       i.controlURL,
-				RouteAll:         false,
+				RouteAll:         i.routeAll,
```
conblem, line 318: Not needed for this PR but I think this should still be configurable

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -121,11 +129,10 @@ func newIPN(jsConfig js.Value) map[string]any {
 	dialer.NetstackDialTCP = func(ctx context.Context, dst netip.AddrPort) (net.Conn, error) {
 		return ns.DialContextTCP(ctx, dst)
 	}
-	sys.NetstackRouter.Set(true)
 
 	logid := lpc.PublicID
 	srv := ipnserver.New(logf, logid, nil /* no netMon */)
-	lb, err := ipnlocal.NewLocalBackend(logf, logid, sys, controlclient.LoginEphemeral)
+	lb, err := ipnlocal.NewLocalBackend(logf, logid, store, dialer, eng, controlclient.LoginEphemeral)
```
conblem, line 135: I will have to look into these changes don't think they are intended probably something todo with merging in main

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -195,6 +204,22 @@ func newIPN(jsConfig js.Value) map[string]any {
...
 		}),
+		"tcp": js.FuncOf(func(this js.Value, args []js.Value) any {
+			if len(args) != 3 && len(args) != 4 {
+				log.Printf("Usage: tcp(hostname, port, readCallback, connectTimeoutSeconds)")
+				return nil
+			}
+			connectTimeoutSeconds := 0
+			if len(args) == 4 && args[3].Type() == js.TypeNumber {
+				connectTimeoutSeconds = args[3].Int()
+			}
```
bradfitz, line 215: let's use float64 (a javascript number) instead of int. Then we can express "0.3" for 300 ms, etc.

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -324,6 +350,117 @@ func (i *jsIPN) logout() {
...
+}
+
+type jsTCPSession struct {
+	jsIPN       *jsIPN
+	conn        net.Conn
+	writeBuffer []byte
+	readBuffer  []byte
+}
+
+const defaultTCPConnectTimeout time.Duration = 5 * time.Second
```
bradfitz, line 381: you can drop "time.Duration" here and let it be implicit.

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -324,6 +350,117 @@ func (i *jsIPN) logout() {
...
+
+func (i *jsIPN) NewJSTCPSession(addr string, readCallback js.Value, connectTimeout time.Duration) (*jsTCPSession, error) {
+	if connectTimeout == 0 {
+		connectTimeout = defaultTCPConnectTimeout
+	}
```
bradfitz, line 386: I'd prefer the caller do this, at the first point it goes from optional float64 seconds to a time.Duration.

So we never have a magic meaning of a time.Duration value (0).

cmd/tsconnect/wasm/wasm_js.go
bradfitz: that's quite the read buffer.  BTW, another Go idiomatic way to write 4 MB would be "4 << 20". But I think 1<<20 is probably enough?

and why are write and read so different?

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -324,6 +350,117 @@ func (i *jsIPN) logout() {
...
+	return map[string]any{
+		"close": js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+			return makePromise(func() (any, error) {
+				err := s.conn.Close()
+				return nil, err
+			})
+		}),
+		"write": js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+			return makePromise(func() (any, error) {
+				dst := args[0]
```
bradfitz, line 446: validate args and return an error if bogus?

cmd/tsconnect/wasm/wasm_js.go
```diff
@@ -324,6 +350,117 @@ func (i *jsIPN) logout() {
...
+
+		js.CopyBytesToJS(subarray, s.readBuffer[:n])
+
+		readCallback.Invoke(subarray)
+	}
+}
+
+func (s *jsTCPSession) jsWrapper() any {
+	return map[string]any{
+		"close": js.FuncOf(func(this js.Value, args []js.Value) interface{} {
```
bradfitz, line 438: use "any" for all "interface{}".

Also, can you add little comments briefly describing the signatures of these JS funcs (whatever args is supposed to be). It can just be in TypeScript syntax if you prefer.

bradfitz (commented):
Lookin' good. Few little comments.

cmd/tsconnect/wasm/wasm_js.go
conblem: Honestly at first i just chose a kb for each but for vnc a big buffer just worked way better. [novnc even uses 4 mb internally](https://github.com/novnc/noVNC/blob/cbbd9ab069e384d230c032f51bf4e713ed02766e/core/websock.js#L56) so I picked that for the moment. I think this should be configurable with the defaults you specified as for the VNC usecase 4 megs just work way better.

To answer your second question the behaviour is different because this way the writer can implicitly choose the buffer size using the size property on the JS ArrayBuffer. Thinking about the read side tho, in hindsight my code does not make alot of sense. So I will simplify that part and just use the size of the go buffer (which is the case anyhow atm just with some useless code).

conblem:
I have implemented your review findings.

To clean the code up a little I created a jsTCPSessionBuilder as an abstraction, lmk what you think about that.

I also took another look at the write and read logic and I think they make sense. 
The write logic uses a predefined buffer size and if the caller provides a bigger buffer from JS we increase the buffer size on go side of things.
For the read logic we try to read the read buffer size. If the size returned from the network is smaller we create a subarray view and provide that to the JS caller. This way the JS caller always gets a subarray with the exact length of the received bytes.

Cheers and thanks for the feedback

Cediddi:
It's been 2.5 years since this MR was open. Conblem did an amazing work, is it possible to re-visit this MR? I'm no Go developer but I'd love to help any way possible.

PR #8061, opened 2023-05-05
 ssh/tailssh: send ssh event notifications on recording failures 

catzkorn:
This change sends an SSHEventNotificationRequest over noise when a
SSH session is set to fail closed and the session is unable to start
because a recorder is not available or a session is terminated because
connection to the recorder is ended. Each of these scenarios have their
own event type.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -3627,6 +3627,19 @@ func (b *LocalBackend) hasNodeKey() bool {
 	return p.Valid() && p.Persist().Valid() && !p.Persist().PrivateNodeKey().IsZero()
 }
 
+// NodeKey returns the public node key.
+func (b *LocalBackend) NodeKey() key.NodePublic {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	p := b.pm.CurrentPrefs()
+	if !p.Valid() || !p.Persist().Valid() || p.Persist().PrivateNodeKey().IsZero() {
+		return key.NodePublic{}
+	}
+
+	return p.Persist().PublicNodeKey()
+}
+
 // nextState returns the state the backend seems to be in, based on
 // its internal state.
 func (b *LocalBackend) nextState() ipn.State {
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -39,6 +39,7 @@ import (
 	"tailscale.com/net/tsdial"
 	"tailscale.com/tailcfg"
 	"tailscale.com/tempfork/gliderlabs/ssh"
+	"tailscale.com/types/key"
 	"tailscale.com/types/logger"
 	"tailscale.com/types/netmap"
 	"tailscale.com/util/clientmetric"
@@ -68,6 +69,7 @@ type ipnLocalBackend interface {
 	DoNoiseRequest(req *http.Request) (*http.Response, error)
 	Dialer() *tsdial.Dialer
 	TailscaleVarRoot() string
+	NodeKey() key.NodePublic
 }
 
 type server struct {
@@ -105,6 +107,7 @@ func init() {
 			logf:           logf,
 			tailscaledPath: tsd,
 		}
+
 		return srv, nil
 	})
 }
@@ -1445,11 +1448,17 @@ func (ss *sshSession) sessionRecordingClient(dialCtx context.Context) (*http.Cli
 // On success, it returns a WriteCloser that can be used to upload the
 // recording, and a channel that will be sent an error (or nil) when the upload
 // fails or completes.
-func (ss *sshSession) connectToRecorder(ctx context.Context, recs []netip.AddrPort) (io.WriteCloser, <-chan error, error) {
+//
+// In both cases, a slice of SSHRecordingAttempts is returned which detail the
+// attempted recorder IP and the error message, if the attempt failed. The
+// attempts are in order the recorder(s) was attempted. If successful a
+// successful connection is made, the last attempt in the slice is the
+// attempt for connected recorder.
+func (ss *sshSession) connectToRecorder(ctx context.Context, recs []netip.AddrPort) (io.WriteCloser, []*tailcfg.SSHRecordingAttempt, <-chan error, error) {
 	if len(recs) == 0 {
-		return nil, nil, errors.New("no recorders configured")
+		return nil, nil, nil, errors.New("no recorders configured")
 	}
-
+	var attempts []*tailcfg.SSHRecordingAttempt
 	// We use a special context for dialing the recorder, so that we can
 	// limit the time we spend dialing to 30 seconds and still have an
 	// unbounded context for the upload.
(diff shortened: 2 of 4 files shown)


ipn/ipnlocal/local.go
```diff
@@ -3627,6 +3627,19 @@ func (b *LocalBackend) hasNodeKey() bool {
 	return p.Valid() && p.Persist().Valid() && !p.Persist().PrivateNodeKey().IsZero()
 }
 
+// NodeKey returns the public node key.
```
maisem, line 3630: If available

ssh/tailssh/tailssh.go
```diff
@@ -105,6 +108,13 @@ func init() {
 			logf:           logf,
 			tailscaledPath: tsd,
 		}
+
```
maisem, line 111: Add comment saying that we want to grab it early in case of FUS

ssh/tailssh/tailssh.go
```diff
@@ -1445,22 +1455,30 @@ func (ss *sshSession) sessionRecordingClient(dialCtx context.Context) (*http.Cli
 // On success, it returns a WriteCloser that can be used to upload the
 // recording, and a channel that will be sent an error (or nil) when the upload
 // fails or completes.
-func (ss *sshSession) connectToRecorder(ctx context.Context, recs []netip.AddrPort) (io.WriteCloser, <-chan error, error) {
+//
+// In both cases, a slice of SSHRecordingAttempts is returned which detail the
+// attempted recorder IP and the error message, if the attempt failed.
```
maisem, line 1460: The attempts are in order the recorder was attempted. If successful, the last attempt in the slice is the attempt for connected recorder.

ssh/tailssh/tailssh.go
```diff
@@ -1476,7 +1494,9 @@ func (ss *sshSession) connectToRecorder(ctx context.Context, recs []netip.AddrPo
 		pr, pw := io.Pipe()
 		req, err := http.NewRequestWithContext(ctx, "POST", fmt.Sprintf("http://%s:%d/record", ap.Addr(), ap.Port()), pr)
 		if err != nil {
-			errs = append(errs, fmt.Errorf("recording: error starting recording: %w", err))
+			errMsg := fmt.Errorf("recording: error starting recording: %w", err)
```
maisem, line 1497: Just err

ssh/tailssh/tailssh.go
```diff
@@ -1445,22 +1455,30 @@ func (ss *sshSession) sessionRecordingClient(dialCtx context.Context) (*http.Cli
...
-
 	// We use a special context for dialing the recorder, so that we can
 	// limit the time we spend dialing to 30 seconds and still have an
 	// unbounded context for the upload.
 	dialCtx, dialCancel := context.WithTimeout(ctx, 30*time.Second)
 	defer dialCancel()
 	hc, err := ss.sessionRecordingClient(dialCtx)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
```
maisem, line 1472: You don't return an attempt here so it won't get notified

tailcfg/tailcfg.go
```diff
@@ -2093,10 +2095,18 @@ type SSHRecordingFailureNotifyRequest struct {
...
-	Attempts []SSHRecordingAttempt
+	// RecordingAttempts is the list of recorders that were attempted, in order.
+	RecordingAttempts []*SSHRecordingAttempt
 }
 
+// SSHEventType defines the event type linked to a SSH action or state.
+type SSHEventType string
+
+const (
+	SessionRecordingRejected   SSHEventType = "recording rejected"
```
maisem, line 2106: You don't need these to be strings. They can be ints

tailcfg/tailcfg.go
```diff
@@ -2093,10 +2095,18 @@ type SSHRecordingFailureNotifyRequest struct {
...
+	SessionRecordingTerminated SSHEventType = "recording terminated"
```
maisem, line 2107: SSHSession...

twitchyliquid64 (approved):
Broad LGTM.

mine thought: So we are getting the nodeKey when the server is initialized, that seems to happen on the first SSH connection? If FUS user switch occurs, is the next ssh connection control send thingy then going to use the node key of the previous node/login ?

sorry if this is smoothbrained, reviewing on mobile sucks lol

perhaps the easiest way to confirm is to test this:
1. Login as user A, accept an ssh connection
2. FUS to user B, accept an ssh connection then cause a termination 
3. See if control got the notification with nodeKey of user a or b

ssh/tailssh/tailssh.go
```diff
@@ -1535,6 +1555,15 @@ func (ss *sshSession) openFileForRecording(now time.Time) (_ io.WriteCloser, err
...
 // It may return a nil recording if recording is not available.
 func (ss *sshSession) startNewRecording() (_ *recording, err error) {
+	// We store the node key as soon as possible when creating
+	// a new recording incase of FUS.
+	var nodeKey key.NodePublic
+	if nk := ss.conn.srv.lb.NodeKey(); nk.IsZero() {
+		return nil, errors.New("ssh server is unavailable: no node key")
+	} else {
+		nodeKey = nk
+	}
```
maisem, line 1570: you can just do
```
nodeKey := ss.conn.srv.lb.NodeKey()
if nodeKey.IsZero() {
  return nil, errors.New("ssh server is unavailable: no node key")
}
```

ssh/tailssh/tailssh.go
```diff
@@ -1407,6 +1410,11 @@ type CastHeader struct {
 
 	// LocalUser is the effective username on the server.
 	LocalUser string `json:"localUser"`
+
+	// ConnectionID uniquely identifies a connection made to the SSH server.
+	// It may be shared across multiple sessions over the same connection in
+	// case a single connection creates multiple sessions.
```
maisem, line 1416: how about

```
// It may be shared across multiple sessions over the same connection in
// case of SSH multiplexing.
```

here and below

ssh/tailssh/tailssh.go
```diff
@@ -1445,22 +1453,36 @@ func (ss *sshSession) sessionRecordingClient(dialCtx context.Context) (*http.Cli
...
 	// We use a special context for dialing the recorder, so that we can
 	// limit the time we spend dialing to 30 seconds and still have an
 	// unbounded context for the upload.
 	dialCtx, dialCancel := context.WithTimeout(ctx, 30*time.Second)
 	defer dialCancel()
 	hc, err := ss.sessionRecordingClient(dialCtx)
 	if err != nil {
-		return nil, nil, err
+		attempts = append(attempts, &tailcfg.SSHRecordingAttempt{
+			FailureMessage: err.Error(),
```
maisem, line 1475: oh i guess this only errors out in case of internal errors, so maybe skip here?

maisem (approved):
LGTM after nits

tailcfg/tailcfg.go
```diff
@@ -2075,9 +2075,17 @@ type SSHRecorderFailureAction struct {
 	NotifyURL string `json:",omitempty"`
 }
 
-// SSHRecordingFailureNotifyRequest is the JSON payload sent to the NotifyURL
-// when a recording fails.
-type SSHRecordingFailureNotifyRequest struct {
+// SSHEventNotifyRequest is the JSON payload sent to the NotifyURL
+// for an SSH event.
+type SSHEventNotifyRequest struct {
```
maisem, line 2080: You should bump the CapVer

PR #8063, opened 2023-05-05
cmd/equaler: add command to generate Equal() methods

### LINKED ISSUE #8077: cmd/viewer,cmd/cloner: add a new command to generate Equal() methods
### What is the issue?

similar to how cmd/cloner generates `Clone()` methods, it would be nice to have a command to generate `Equal()` methods to avoid hand rolling them or relying on `reflect.DeepEqual()`.

for example, the `tailcfg` package, which already uses veiwer/cloner, has a few unwieldy `Equal()`s and  comments warning not to forget to keep them updated with struct changes.

### Steps to reproduce

_No response_

_No response_

### OS

_No response_

### OS version

_No response_

### Tailscale version

_No response_

### Other software

_No response_

### Bug report

_No response_

saljmz:
The generator is still crude and do not cover most types, but it covers all the ones needed by the tailcfg package. It's a start.

Fixes #8077.


tailcfg/tailcfg_test.go
```diff
@@ -449,7 +449,7 @@ func TestNodeEqual(t *testing.T) {
 		{
 			&Node{Addresses: []netip.Prefix{}},
 			&Node{Addresses: nil},
-			false,
+			true,
```
bradfitz, line 452: uhhh

util/codegen/codegen.go
```diff
@@ -55,6 +55,17 @@ func HasNoClone(structTag string) bool {
...
 
+// HasNoEqual reports whether the provided tag has `codegen:noequal`.
+func HasNoEqual(structTag string) bool {
+	val := reflect.StructTag(structTag).Get("codegen")
+	for _, v := range strings.Split(val, ",") {
+		if v == "noequal" {
+			return true
+		}
+	}
+	return false
```
bradfitz, line 66: https://pkg.go.dev/golang.org/x/exp/slices#Contains

tailcfg/tailcfg.go
```diff
@@ -582,9 +583,6 @@ type Hostinfo struct {
 	Cloud           string         `json:",omitempty"`
 	Userspace       opt.Bool       `json:",omitempty"` // if the client is running in userspace (netstack) mode
 	UserspaceRouter opt.Bool       `json:",omitempty"` // if the client's subnet router is running in userspace (netstack) mode
-
-	// NOTE: any new fields containing pointers in this type
-	//       require changes to Hostinfo.Equal.
 }
```
bradfitz, line 586: how do we get a test failure if you forget to re-gen?

PR #8064, opened 2023-05-05
net/netns: fix segv when no tailscale interface is found

raggi:
Observed via isolated test failure.

Updates #cleanup


bradfitz:
Code seems fine but this needs more info, not just #cleanup. Why'd it crash? What was the stack? What was the test?

bradfitz (approved):

Issue #8067, opened 2023-05-05
ipn/ipnlocal: cert fetching sometimes hangs+fails on macsys


bradfitz:
@marwan-at-work reports that LetsEncrypt cert fetching sometimes hangs+fails on macsys and debugged it down to LookupTXT hanging.

Marwan can elaborate.

Tracking bug.

bradfitz:
To @marwan-at-work to verify the fix. We'll need an unstable build first, or you'll need to build the mac client yourself.

marwan-at-work:
I can confirm it works (I deleted the certs to verify a few times).

However, a couple of times I got: 

```
cert("marwan-tsmac.corp.ts.net"): getCertPEM: SetDNS "_acme-challenge.marwan-tsmac.corp.ts.net" => "xxx-xxx": read tcp [2603:7000:2103:2d11:d171:dd23:19a3:80f2]:57904->[2a05:d014:386:202:24dd:46aa:f98e:9997]:80: read: connection reset by peer
```

bradfitz:
We'll call it good then. :) 

If we see that more we can open a new bug.

Issue #3273, opened 2021-11-06
MacOS cert fails and takes down Tailscale


metcalfc:
### What is the issue?

``` 
/Applications/Tailscale.app/Contents/MacOS/Tailscale cert macbook-pro.cute-animal.ts.net
2021/11/06 14:45:07 Get "http://local-tailscaled.sock/localapi/v0/cert/macbook-pro.cute-animal.ts.net?type=pair": EOF
```
After the command fails tailscale drops. You have to log in again.

I would have tried sudo but researching this bug I found #3063 and figured out that wouldn't help. I checked the console and didn't see anything.

### Steps to reproduce

Login
Try to get a cert. 
Tailscale stops.
Rinse and repeat.

### Are there any recent changes that introduced the issue?

This is a fresh install of tailscale of a clean install of Montereyy 12.01.

### OS

macOS

### OS version

12.01

### Tailscale version

1.16.2

### Bug report

BUG-2a47ed8230543b75fa33001d82ebc1d7e963b868c5cd51bc0fbd39afe9459267-20211106215621Z-e3c3c32cb8fc4c66

bradfitz:
Unfortunately the bugreport/logs aren't capturing the crash, if one occured.

I see:

```
2021-11-06 14:45:03.532443 -0700 PDT: cert("xxx-ostrich.ts.net"): already had ACME account.
2021-11-06 14:45:07.856475 -0700 PDT: [D2lkC] - Sending keepalive packet
2021-11-06 14:45:29.363765 -0700 PDT: logtail started
2021-11-06 14:45:29.364188 -0700 PDT: goSetupLogs: success (started with NaNM); version 1.16.2-tf44911179-g3853a3b92, go1.17.1-ts097d1284f9
```

(that is, the "already had ACME account" line, followed by it starting again some time later...)

If you watch the logs in Console.app you might be able to see the actual crash. (Search for "IPNExtension" and start recording, then repro it)

But we can try again here and see if we get lucky and can repro it too.

metcalfc:
Thanks @bradfitz its different today. 500 error and the daemon lives.

```
/Applications/Tailscale.app/Contents/MacOS/Tailscale cert xxx-ostrich.ts.net
2021/11/07 20:04:56 HTTP 500 Internal Server Error: timeout waiting for order to not be invalid
 (expected 200)
```

```
[...]

PR #8068, opened 2023-05-05
ipn/ipnlocal: bound how long cert fetchher checks for existing DNS records

bradfitz:
It was supposed to be best effort but in some cases (macsys at least, per @marwan-at-work) it hangs and exhausts the whole context.Context deadline so we fail to make the SetDNS call to the server.

Updates #8067
Updates #3273 etc


marwan-at-work (approved):
👍🏼

PR #8070, opened 2023-05-05
Add environment flags for custom TLS root CA and SNI

### LINKED ISSUE #8069: FR: Flag to override server TLS root certificates

Use tailscale client with a control server which is hosted with a TLS server certificate signed by a trusted root that does not chain up to the normal system CA. For example, a corporate CA that isn't signed by digikey, etc.

### How should we solve this?

Implement environment variable flags to (optionally) set the root CAs used by tailscale cleint to verify the control server. The default will remain to use the "normal" trust root of the system.

All users of tailscale must use a control server with a TLS certificate signed by something like digikey / letsencrypt. In some cases, users don't wish to trust these major trust root organizations (which have historically had leaks and vulnerabilities) and would prefer to issue their own CA. 

### Anything else?

Another flag to set the ServerName is also desrireable, so the control server can be verified against a different SNI the the one connecting to.

kev-the-dev:
Add optional environment variables which can modify the way tailscaled verifies the TLS certificate of the control server. If unset (default), there is no change to current functionality

`TS_TLS_DIAL_ROOT_CA` specifies a file to treat as the root certificate authority when verifying the control server, rather then the system default. This is useful when connecting to a control server signed by a trusted CA which is not part of the normal system trust root (digicert, let's encrypt, etc) - for example a trusted corporate CA. 

`TS_TLS_DIAL_CONNECT_TO` sets the server name (SNI) that the control server certificate is verified against, as well as the SNI given to the server when connecting. This is useful for verifying the server against a different domain then the one used to connect with, for example using `--login-server https://mycompany.net`, but verifying against `server-1.special-sni`. It can also be used by reverse proxies to serve a different TLS certificate based on SNI.

Closes #8069

Signed-off-by: Kevin Allen kallen@bostondynamics.com

### DIFF
--- a/net/tlsdial/tlsdial.go
+++ b/net/tlsdial/tlsdial.go
@@ -41,6 +41,17 @@ var debug = envknob.RegisterBool("TS_DEBUG_TLS_DIAL")
 // Headscale, etc.
 var tlsdialWarningPrinted sync.Map // map[string]bool
 
+var (
+	// rootCAOverride creates environment variable config TS_TLS_DIAL_ROOT_CA which
+	// will override the certificate authority used to verify the server instead
+	// of the system default
+	rootCAOverride = envknob.RegisterString("TS_TLS_DIAL_ROOT_CA")
+	// serverHostOverride creates environment variable TS_TLS_DIAL_CONNECT_TO which
+	// will override the server name the certificate is validated against AND the SNI
+	// name presented to the server, which may affect virtual hosts
+	serverHostOverride = envknob.RegisterString("TS_TLS_DIAL_CONNECT_TO")
+)
+
 // Config returns a tls.Config for connecting to a server.
 // If base is non-nil, it's cloned as the base config before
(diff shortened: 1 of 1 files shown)


DentonGentry (changes requested):
Marking this as request changes to make it clear that it is not to go in yet. It is not at all clear to me that this is the right thing to do.

net/tlsdial/tlsdial.go
```diff
@@ -52,6 +63,9 @@ func Config(host string, base *tls.Config) *tls.Config {
 		conf = base.Clone()
 	}
 	conf.ServerName = host
+	if len(serverHostOverride()) != 0 {
+		conf.ServerName = serverHostOverride()
+	}
```
bradfitz, line 68: But this tlsdial.Config func is used for more than just dialing the control plane. There are ~4 different types of services it's used for, so 1 env var can't work for all of them.

PR #8073, opened 2023-05-06
fix sys.Set(router) issue will crash the daemon in some OSs

gps949:
Signed-off-by: Chenyang Gao <gps949@outlook.com>

in commit 6e96744, the tsd system type has been added. Which will cause the daemon will crash on some  OSs (Windows, darwin and so on). The root cause is that on those OSs, handleSubnetsInNetstack() will return true and set the conf.Router with a wrapper.  Later in NewUserspaceEngine() it will do subsystem set and found that early set router mismatch to current value, then panic.


bradfitz:
Thanks!

PR #8078, opened 2023-05-06
cmd/tsmultiserve: tsnet-based tcp proxy

saljmz:
Like `tailscale serve` but can listen on multiple names at the same time.

I use this to publish things on my tailnet using canonical ports with memorable names instead of whatever they're listening on. Something I wish I could do from the "services" tab on login.tailscale.com.

### DIFF
--- a/cmd/tsmultiserve/main.go
+++ b/cmd/tsmultiserve/main.go
@@ -1,0 +1,176 @@
+// Server tsmultiserve is a TCP proxy that can register and listen on multiple tailscale addresses
+// using tsnet.
+//
+// The main motivation for this is to run multiple services on the same host, but give
+// them memorable names and use canonical ports.
+//
+// Usage:
+//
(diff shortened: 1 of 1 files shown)


bradfitz:
I'd prefer "tsnet" not be in the name. That's an implementation detail and not what it _is_.

cmd/tsnetd/main.go
```diff
@@ -0,0 +1,158 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"tailscale.com/tsnet"
+)
+
+var logf = func(string, ...any) {}
+
+func proxy(s *tsnet.Server, a net.Conn, dst string) {
```
maisem, line 41: can we do this using serve now?

maisem: ah so you're basically running multiple tsnet instances here, so you could still use serve on a per tsnet.Server basis.

PR #8084, opened 2023-05-07
fix: to get the dirty field of version from vcs.modified value

gps949:
Signed-off-by: Chenyang Gao <gps949@outlook.com>

Current code will set the "dirty" field of VersionInfo always "true" if vcs.modified flag is there. No matter whether the flag is "true" or "false".     
It will make sense to set this field due to vcs.modified value, not only the existence of the key.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -131,7 +131,7 @@ var getEmbeddedInfo = lazy.SyncFunc(func() embeddedInfo {
 				ret.commitDate = strings.ReplaceAll(ret.commitDate, "-", "")
 			}
 		case "vcs.modified":
-			ret.dirty = true
+			ret.dirty = s.Value == "true"
 		}
 	}
 	if ret.commit == "" || ret.commitDate == "" {


bradfitz:
I verified Go does:

```
src/cmd/go/internal/load/pkg.go:                appendSetting("vcs.modified", strconv.FormatBool(st.Uncommitted))
```

So it's only ever "true" or "false" and we don't need strconv.ParseBool, etc.

@gps949, can you amend & force push your commit message to be in our conventional format?

The commit message should be like:

```
version: properly handle vcs.modified when it's "false"

Current code will set the "dirty" field of VersionInfo always "true"
if vcs.modified flag is there. No matter whether the flag is "true" or "false".    
It will make sense to set this field due to vcs.modified value,
not only the existence of the key.

Signed-off-by: Chenyang Gao <gps949@outlook.com>
```

bradfitz:
Actually, nevermind. I reworded it for you and merged it in #8087.

gps949:
> Actually, nevermind. I reworded it for you and merged it in #8087.
> 
> 
You are so nice. Thanks!

PR #8087, opened 2023-05-07
version: properly handle vcs.modified when it's "false"

bradfitz:
Current code will set the "dirty" field of VersionInfo always "true" if vcs.modified flag is there. No matter whether the flag is "true" or "false".  It will make sense to set this field due to vcs.modified value, not only the existence of the key.


bradfitz:
(This is #8084, but with the commit message re-worded.)

PR #8091, opened 2023-05-08
metrics: add histogram support

maisem:
Add initial histogram support.

Updates tailscale/corp#8641


metrics/metrics.go
```diff
@@ -58,3 +63,55 @@ func (m *LabelMap) GetFloat(key string) *expvar.Float {
...
+// NewHistogram returns a new histogram that reports to the given
+// expvar map under the given name.
+//
+// The buckets are the boundaries of the histogram buckets, in
+// increasing order. The last bucket is +Inf.
+func NewHistogram(name string, buckets []float64) *histogram {
+	if !slices.IsSorted(buckets) {
+		panic("buckets must be sorted")
+	}
+	lm := &LabelMap{
```
knyar, line 88: LabelMap will return keys sorted alphabetically, while prometheus metric format [requires](https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#histograms-and-summaries) buckets to appear in increasing numeric order.

metrics/metrics.go
```diff
@@ -58,3 +63,55 @@ func (m *LabelMap) GetFloat(key string) *expvar.Float {
...
+// Observe records a new observation in the histogram.
+func (h *histogram) Observe(v float64) {
+	h.sum.Add(v)
+	h.count.Add(1)
+	for i, b := range h.buckets {
+		if v <= b {
+			str := h.bucketStrings[i]
+			if str == "" {
+				str = fmt.Sprintf("%.2f", b)
+				h.bucketStrings[i] = str
```
knyar, line 110: I think this might crash if multiple goroutines `Observe` concurrently for the first time and attempt concurrent writes to the map. Might be safer to populate `bucketStrings` in `NewHistogram`?

metrics/metrics.go
```diff
@@ -58,3 +63,55 @@ func (m *LabelMap) GetFloat(key string) *expvar.Float {
...
+			}
+			h.lm.Add(str, 1)
+			return
```
knyar, line 113: Buckets are cumulative, so you need to continue incrementing all of them (incl. `+Inf`).

knyar (commented):
You can use prometheus `testutil.ScrapeAndCompare` to test this ([example](https://github.com/tailscale/tailscale/blob/main/tsweb/promvarz/promvarz_test.go)), but it will probably require a separate test package to avoid pulling in `prometheus` as a build dependency.

metrics/metrics.go
maisem: Done

metrics/metrics.go
maisem: Done.

metrics/metrics.go
```diff
@@ -58,3 +63,55 @@ func (m *LabelMap) GetFloat(key string) *expvar.Float {
...
+
+// Observe records a new observation in the histogram.
+func (h *histogram) Observe(v float64) {
+	h.sum.Add(v)
+	h.count.Add(1)
+	for i, b := range h.buckets {
+		if v <= b {
+			str := h.bucketStrings[i]
+			if str == "" {
+				str = fmt.Sprintf("%.2f", b)
```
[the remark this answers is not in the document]
maisem: I used `%v` should I use something else here?

metrics/metrics.go
```diff
@@ -58,3 +65,86 @@ func (m *LabelMap) GetFloat(key string) *expvar.Float {
...
+	h.sum.Add(v)
+	h.count.Add(1)
+	for i, b := range h.buckets {
+		if v <= b {
+			h.bucketVars[i].Add(1)
+		}
+	}
+}
+
+func (h *Histogram) String() string {
```
knyar, line 114: Perhaps it's worth adding a comment clarifying that this method is here to satisfy `expvar.Var`? Also, do we want to export `count` as well?

metrics/metrics.go
```diff
@@ -58,3 +65,86 @@ func (m *LabelMap) GetFloat(key string) *expvar.Float {
 func CurrentFDs() int {
 	return currentFDs()
 }
+
+type Histogram struct {
+	name string
```
knyar, line 70: Seems like `name` is no longer used, and `Histogram` needs to be explicitly published?

metrics/metrics.go
```diff
@@ -58,3 +65,86 @@ func (m *LabelMap) GetFloat(key string) *expvar.Float {
...
+	// buckets is a list of bucket boundaries, in increasing order.
+	buckets []float64
+	// buckStrings is a list of the same buckets, but as strings.
+	// To avoid allocations, this is only populated on first use.
```
knyar, line 74: The second line of this comment is no longer true.

knyar (approved):

knyar (approved):

PR #8095, opened 2023-05-08
wgengine/magicsock: prefer private endpoints to public ones

andrew-d:
Switch our best address selection to use a scoring-based approach, where we boost each address based on whether it's a private IP or IPv6.

For users in cloud environments, this biases endpoint selection towards using an endpoint that is less likely to cost the user money, and should be less surprising to users.

This also involves updating the tests to not use private IPv4 addresses; other than that change, the behaviour should be identical for existing endpoints.

Updates #8097

Change-Id: I069e3b399daea28be66b81f7e44fc27b2943d8af

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -5090,40 +5090,58 @@ func betterAddr(a, b addrLatency) bool {
 	if !a.IsValid() {
 		return false
 	}
-	if a.Addr().Is6() && b.Addr().Is4() {
-		// Prefer IPv6 for being a bit more robust, as long as
-		// the latencies are roughly equivalent.
-		if a.latency/10*9 < b.latency {
-			return true
-		}
-	} else if a.Addr().Is4() && b.Addr().Is6() {
-		if betterAddr(b, a) {
-			return false
-		}
+
+	// Each address starts with a set of points (from 0 to 100) that
+	// represents how much faster they are than the highest-latency
+	// endpoint. For example, if a has latency 200ms and b has latency
+	// 190ms, then a starts with 0 points and b starts with 5 points since
+	// it's 5% faster.
+	var aPoints, bPoints int
+	if a.latency > b.latency && a.latency > 0 {
+		bPoints = int(100 - ((b.latency * 100) / a.latency))
+	} else if b.latency > 0 {
+		aPoints = int(100 - ((a.latency * 100) / b.latency))
 	}
 
-	// If we get here, then both addresses are the same IP type (i.e. both
-	// IPv4 or both IPv6). All decisions below are made solely on latency.
+	// Prefer private IPs over public IPs as long as the latencies are
+	// roughly equivalent, since it's less likely that a user will have to
+	// pay for the bandwidth in a cloud environment.
 	//
-	// Determine how much the latencies differ; we ensure the larger
-	// latency is the denominator, so this fraction will always be <= 1.0.
-	var latencyFraction float64
-	if a.latency >= b.latency {
-		latencyFraction = float64(b.latency) / float64(a.latency)
-	} else {
-		latencyFraction = float64(a.latency) / float64(b.latency)
+	// Additionally, prefer any loopback address strongly over non-loopback
+	// addresses.
+	if a.Addr().IsLoopback() {
+		aPoints += 50
+	} else if a.Addr().IsPrivate() {
+		aPoints += 20
+	}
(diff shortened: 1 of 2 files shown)


andrew-d:
Do we already have a bug for "improve how we pick endpoints"? I couldn't find one with some initial searches. If not, happy to file one, but I wanted to check first 😃

sailorfrag:
> Do we already have a bug for "improve how we pick endpoints"? I couldn't find one with some initial searches. If not, happy to file one, but I wanted to check first smiley

This _might_ improve the scenario outlined in #2331 but I don't think it directly covers what's being done here.

andrew-d:
Filed #8097 to track this kind of change

wgengine/magicsock/magicsock.go
```diff
@@ -5090,40 +5090,61 @@ func betterAddr(a, b addrLatency) bool {
...
+	// 190ms, then a starts with 0 points and b starts with 5 points since
+	// it's 5% faster.
+	highestLatency := a.latency
+	if b.latency > highestLatency {
+		highestLatency = b.latency
+	}
+	var aPoints, bPoints int
+	if highestLatency > 0 {
+		aPoints = int(100 - ((a.latency * 100) / highestLatency))
+		bPoints = int(100 - ((b.latency * 100) / highestLatency))
```
sailorfrag, line 5106: One of these will always be 0, the other will be between 0 and 100. It's probably going to be simpler just to do
```golang
if a.latency > b.latency && a.latency > 0 {
   bPoints = int(100 - ((b.latency * 100) / a.latency))
} else if b.latency > 0 {
   aPoints = int(100 - ((a.latency * 100) / b.latency))
}
```

wgengine/magicsock/magicsock.go
```diff
@@ -5090,40 +5090,61 @@ func betterAddr(a, b addrLatency) bool {
...
 	// Don't change anything if the latency improvement is less than 1%; we
 	// want a bit of "stickiness" (a.k.a. hysteresis) to avoid flapping if
 	// there's two roughly-equivalent endpoints.
-	if latencyFraction >= 0.99 {
+	//
+	// The points added by our latency scoring, above, are already as a
+	// fraction of the fastest latency. Thus, we can return false where the
+	// aPoints value is <= 1 (since we also want to ignore any other
+	// boosts for private IPs/IPv6/etc.).
+	if aPoints > bPoints && aPoints <= 1 {
```
sailorfrag, line 5143: I think the `aPoints > bPoints` is unnecessary. Currently, if `aPoints <= 1` and `bPoints` is 0, it'll return false here. If `bPoints > 0`, it'll return false at the end anyway.

wgengine/magicsock/magicsock.go
```diff
@@ -5090,40 +5090,61 @@ func betterAddr(a, b addrLatency) bool {
...
 	}
 
 	// Don't change anything if the latency improvement is less than 1%; we
 	// want a bit of "stickiness" (a.k.a. hysteresis) to avoid flapping if
 	// there's two roughly-equivalent endpoints.
-	if latencyFraction >= 0.99 {
+	//
+	// The points added by our latency scoring, above, are already as a
+	// fraction of the fastest latency. Thus, we can return false where the
+	// aPoints value is <= 1 (since we also want to ignore any other
```
sailorfrag, line 5141: It took a few reads through to convince myself that `<= 1` is correct. It might be good to restate a summary of the formula ("points are the percentage improvement in latency vs the slower endpoint").

sailorfrag (approved):

wgengine/magicsock/magicsock.go
andrew-d: I tried this but found it more confusing; changed it to check if bPoints == 0 (since that's the only way aPoints > bPoints and aPoints <= 1 here).

wgengine/magicsock/magicsock_test.go
```diff
@@ -1631,52 +1631,101 @@ func TestBetterAddr(t *testing.T) {
...
+		publicV4   = "1.2.3.4:555"
+		publicV4_2 = "5.6.7.8:999"
+		publicV6   = "[2001::5]:123"
+
+		privateV4 = "10.0.0.2:123"
+	)
+
 	tests := []struct {
 		a, b addrLatency
 		want bool
```
bradfitz, line 1645: want bool // whether a is better

bradfitz (approved):

PR #8101, opened 2023-05-08
syncs: add Map.WithLock to allow mutations to the underlying map

dsnet:
Some operations cannot be implemented with the prior API:
* Iterating over the map and deleting keys
* Iterating over the map and replacing items
* Calling APIs that expect a native Go map

Add a Map.WithLock method that acquires a write-lock on the map
and then calls a user-provided closure with the underlying Go map.
This allows users to interact with the Map as a regular Go map,
but with the gaurantees that it is concurrent safe.

Updates tailscale/corp#9115

### DIFF
--- a/syncs/syncs.go
+++ b/syncs/syncs.go
@@ -252,8 +252,10 @@ func (m *Map[K, V]) Delete(key K) {
 	delete(m.m, key)
 }
 
-// Range iterates over the map in undefined order calling f for each entry.
+// Range iterates over the map in an undefined order calling f for each entry.
 // Iteration stops if f returns false. Map changes are blocked during iteration.
+// A read lock is held for the entire duration of the iteration.
+// Use the [WithLock] method instead to mutate the map during iteration.
 func (m *Map[K, V]) Range(f func(key K, value V) bool) {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
@@ -262,6 +264,15 @@ func (m *Map[K, V]) Range(f func(key K, value V) bool) {
 			return
 		}
 	}
+}
+
+// WithLock calls f with the underlying map.
+// Use of m2 must not escape the duration of this call.
+// The write-lock is held for the entire duration of this call.
+func (m *Map[K, V]) WithLock(f func(m2 map[K]V)) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	f(m.m)
 }
 
 // Len returns the length of the map.
--- a/syncs/syncs_test.go
+++ b/syncs/syncs_test.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"io"
 	"os"
-	"sync"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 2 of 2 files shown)


syncs/syncs.go
```diff
@@ -160,20 +160,29 @@ func (s Semaphore) Release() {
 // at a relatively high frequency.
 // This must not be shallow copied.
 type Map[K comparable, V any] struct {
+	// alreadyLocked reports whether a write-lock is already held.
+	// If so, we avoid acquiring any locks on mu.
+	// This is used by RangeMutable to allow mutations during iteration.
+	alreadyLocked bool
```
maisem, line 166: I can't convince myself this is safe. cc @bradfitz

raggi:
Why not:

```go
func RangeReplace(f func(newM map[K]V, k K, v V) {
  ...
  
  newM[k] = v
})
```

i.e. the caller is _required_ to write into the new map.

You could also draw from the functional model and this becomes a specialization of `Reduce` that returns an accumulator along with each iteration, so then it's `ReduceReplace` (or InjectReplace in some parlance) (e.g. https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.reduce).

dsnet:
@raggi 

I'm not sure how your proposed API works. I see no return variables, nor any persisted state that it would be mutating. Did you intend for that to be a method on `Map`?

syncs/syncs.go
dsnet: This field is only ever set in a new `Map`, so there should be raciness with regard to this field itself.

With regard to the data protected by `Map.mu`, the `RangeMutable` method

raggi:
> I'm not sure how your proposed API works. I see no return variables, nor any persisted state that it would be mutating. Did you intend for that to be a method on `Map`?

Yes, sorry, that would be a method on map

raggi:
I think the underlying question is: why not just give the user a regular map, rather than a Map?

dsnet:
> I think the underlying question is: why not just give the user a regular map, rather than a Map?

That's reasonable too, we could just have an API like:
```go
// WithLock calls f with the underlying map.
// The write-lock is held for the entire duration of this call. 
func (m *Map[K,V]) WithLock(f func(m map[K]V)) {
    m.mu.Lock()
    defer m.mu.Unlock()
    f(m.m)
}
```
I don't have a strong opinion to go with `Map.WithLock` or `Map.RangeMutable`.

dsnet:
Any thoughts on whether to add `RangeMutable` or `WithLock`?

raggi:
I'd probably go for `WithLock` as it potentially covers more cases, such as amortizing the cost of mutating a batch of entries, perhaps in a non-enumeration fashion, as well as the enumeration case. I've seen some evidence of dislike for closures following this kind of pattern, I don't know if there are related Go culture preferences that push back on that aspect?

dsnet:
> I've seen some evidence of dislike for closures following this kind of pattern

We already use a closure for `Map.Range`. At least `WithLock` will be consistent.

bradfitz (approved):

PR #8103, opened 2023-05-09
derp,derper: add flow tracking

### LINKED ISSUE #3560: derp: track flows internally

valscale:
Track all data flows between clients of a derp server, both local and remote, and display them in /debug/flows/.

Fixes #3560


derp/derp_server.go
```diff
@@ -944,6 +1028,48 @@ func (s *Server) notePeerSendLocked(src key.NodePublic, dst *sclient) {
...
+	return &f
+}
+
+// updateFlowStats tracks every byte that enters the derp for a
+// src,dst,connNum tuple. f must be the valid flow, either looked up from the src
+// client's flows or recently allocated and stored only in the server
+// flow map.
+func (c *sclient) updateFlowStats(f *flowStats, dstKey, srcOrFwd key.NodePublic, contents []byte) {
+	if _, ok := c.flows[dstKey]; !ok {
+		c.flows[dstKey] = make(map[key.NodePublic]*flowStats)
```
raggi, line 1062: https://pkg.go.dev/tailscale.com/util/mak#Set may be helpful here and in a few other places

derp/derp_server.go
```diff
@@ -1942,6 +2101,111 @@ func (s *Server) ServeDebugTraffic(w http.ResponseWriter, r *http.Request) {
...
+	sort.Slice(lines, func(i, j int) bool { return lines[i].bps > lines[j].bps })
+
+	headStr := "%-10v %-10v %-10v %-10v %-10v %10v %10v %10v %10v %-12v\n"
+	bodyStr := "%-10v %-10v %-10v %-10v %-10v %10v %10v %10v %10v %5v secs\n"
+	fmt.Fprintf(w, headStr, "conn#", "src", "dst", "recv from", "fwded to", "bytes", "packets", "bps", "pps", "last active")
+	for _, l := range lines {
+		fmt.Fprintf(w, bodyStr, l.connNum, l.src.ShortString(), l.dst.ShortString(), l.recvFrom.ShortString(), l.fwdTo.ShortString(), l.bytes, l.packets, l.bps, l.pps, int(l.lastActive.Seconds()))
+	}
+
+	w.(http.Flusher).Flush()
```
raggi, line 2206: Is this flush needed? Normally the http library will flush after a synchronous http handler returns. If it is it would be great to have a comment, as this function otherwise looks like a normal synchronous handler.

derp/derp_server.go
```diff
@@ -1942,6 +2101,111 @@ func (s *Server) ServeDebugTraffic(w http.ResponseWriter, r *http.Request) {
...
+					if _, ok2 := lastSnap[c][dstKey]; ok2 {
+						last = lastSnap[c][dstKey][srcOrFwd]
+					}
+				}
+				bytes := l.bytes - last.bytes
+				packets := l.packets - last.packets
+				l.bps = int64(float64(bytes) / dur.Seconds())
+				l.pps = int64(float64(packets) / dur.Seconds())
+				// src and mesh take a lil calculation
+				if srcOrFwd == (key.NodePublic{}) {
```
raggi, line 2175: nit: `srcOrFwd.IsZero()`

derp/derp_server.go
```diff
@@ -795,6 +824,49 @@ func (c *sclient) run(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+		c.cleanFlows()
+	}
+}
+
+// flowsTimerFunc tells all the sclients to scan for inactive flows
+// and free them.
+func (s *Server) flowsTimerFunc() {
```
raggi, line 833: It would be nice to be able to have this shutdown when the derp server shuts down so that integration/e2e tests don't leak a periodic task.

The shortest path may be to check s.closed before releasing the lock.

raggi (approved):
Looks good, a few comments inline - stopping the timer on close being the most important change.

derp/derp_server.go
```diff
@@ -88,6 +93,15 @@ const (
 	disableFighters
 )
 
+// flowStats tracks statitics for a client to client data flow through
```
andrew-d, line 96:
```suggestion
// flowStats tracks statistics for a client to client data flow through
```

derp/derp_server.go
```diff
@@ -162,6 +176,14 @@ type Server struct {
 	// src.
 	sentTo map[key.NodePublic]map[key.NodePublic]int64 // src => dst => dst's latest sclient.connNum
 
+	// flows is a map of all active flowStats on the server,
+	// protected by s.mu(). After allocation by the server,
```
andrew-d, line 180:
```suggestion
	// protected by s.mu. After allocation by the server,
```

(since `s.mu` isn't a function)

derp/derp_server.go
```diff
@@ -944,6 +1028,48 @@ func (s *Server) notePeerSendLocked(src key.NodePublic, dst *sclient) {
...
+	}
+
+	if _, ok := s.flows[c][dstKey][srcOrFwd]; ok {
+		s.logf("[unexpected] flowStat already allocated for s.flows[%v(%v)][%v][%v]",
+			c.key.ShortString(), c.connNum, dstKey.ShortString(), srcOrFwd.ShortString())
+	}
+	f := flowStats{}
+	s.flows[c][dstKey][srcOrFwd] = &f
+	// Populate the sclient's flow cache outside the lock
+	return &f
```
andrew-d, line 1053: You could declare `f` as `&flowStats{}` and then avoid taking the address twice below, if you wanted.

derp/derp_server.go
```diff
@@ -1942,6 +2101,111 @@ func (s *Server) ServeDebugTraffic(w http.ResponseWriter, r *http.Request) {
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	for c := range s.flows {
+		snap[c] = make(map[key.NodePublic]map[key.NodePublic]flowStats)
+		for dstKey := range s.flows[c] {
+			snap[c][dstKey] = make(map[key.NodePublic]flowStats)
+			for srcOrFwd := range s.flows[c][dstKey] {
+				snap[c][dstKey][srcOrFwd] = *s.flows[c][dstKey][srcOrFwd]
+			}
```
andrew-d, line 2120: Using the [`maps.Clone`](https://pkg.go.dev/golang.org/x/exp/maps#Clone) function might simplify some of this, if you want.

derp/derp_server.go
```diff
@@ -1942,6 +2101,111 @@ func (s *Server) ServeDebugTraffic(w http.ResponseWriter, r *http.Request) {
...
+// second.
+func (s *Server) ServeFlows(w http.ResponseWriter, r *http.Request) {
+	fmt.Fprintf(w, "current time: %s\n", time.Now().Format("2006-01-02 15:04:05"))
+
+	lastTime := time.Now()
+	lastSnap := s.takeFlowsSnapshot()
+	time.Sleep(1 * time.Second)
+	currTime := time.Now()
+	currSnap := s.takeFlowsSnapshot()
+	dur := currTime.Sub(lastTime)
```
andrew-d, line 2137: A comment explaining why we're taking two snapshots and sleeping for 1 second would help here; I think it's because we're using the `lastSnap` to calculate bandwidth over the 1s interval, but I'm not sure I'm reading this code right? Does the fact that it's 1s matter, other than making the "per-second" column calculations easier?

derp/derp_server.go
valscale: You are correct, it doesn't need a flush.

derp/derp_server.go
valscale: Good point. I read a lot of documentation on the difficulty of using timer.Stop() correctly and I think the new version does it properly, using s.mu to synchronize the restart/stop.

derp/derp_server.go
valscale: Thanks, done!

derp/derp_server.go
valscale: Thanks! I was able to use mak.Set() in the initial allocation but it didn't make sense here where the common case is lookup.

derp/derp_server.go
valscale: Thanks, comment added! 1 second is an arbitrary "enough" to get some kind of reasonable bandwidth estimate while not delaying too long in response.

knyar (approved):

derp/derp_server.go
valscale: I gave it a go! Multi-level maps aren't a great match for existing libraries.

derp/derp_server.go
```diff
@@ -88,6 +93,15 @@ const (
 	disableFighters
 )
 
+// flowStats tracks statistics for a client to client data flow through
+// this server. The counters are written solely by run(), read by the
+// debug server, and do not need to be thread-safe.
```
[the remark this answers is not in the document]
valscale: Looking forward to working through the potential concurrency problems. :)

derp/derp_server.go
```diff
@@ -1942,6 +2085,113 @@ func (s *Server) ServeDebugTraffic(w http.ResponseWriter, r *http.Request) {
...
+// ServeFlows prints a list of recently active data flows through
+// this server, including the bandwidth used by each over the last
+// second.
+func (s *Server) ServeFlows(w http.ResponseWriter, r *http.Request) {
+	fmt.Fprintf(w, "current time: %s\n", time.Now().Format("2006-01-02 15:04:05"))
+
+	lastTime := time.Now()
+	lastSnap := s.takeFlowsSnapshot()
+	// Sleep long enough to get some kind of useful bandwidth estimate
+	time.Sleep(1 * time.Second)
```
bradfitz, line 2121: Maybe make this a URL parameter? It can default to 1 second, if that's good. It feels low, though.

Better would probably be a background snapshotter every 30 seconds or something, then this handler doesn't need to block at all.

derp/derp_server.go
```diff
@@ -1942,6 +2085,113 @@ func (s *Server) ServeDebugTraffic(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
+
+	fmt.Fprintf(w, "number flows: %v\n", len(lines))
+	// Sort by highest bandwidth flow
+	sort.Slice(lines, func(i, j int) bool { return lines[i].bps > lines[j].bps })
+
+	headStr := "%-10v %-10v %-10v %-10v %-10v %10v %10v %10v %10v %-12v\n"
+	bodyStr := "%-10v %-10v %-10v %-10v %-10v %10v %10v %10v %10v %5v secs\n"
+	fmt.Fprintf(w, headStr, "conn#", "src", "dst", "recv from", "fwded to", "bytes", "packets", "bps", "pps", "last active")
```
bradfitz, line 2189: add a TODO at least to add a machine-readable format? JSON something?

derp/derp_server.go
```diff
@@ -795,6 +825,54 @@ func (c *sclient) run(ctx context.Context) error {
...
+					delete(c.flows, dstKey)
+					s.mu.Lock()
+					delete(s.flows[c], dstKey)
+					s.mu.Unlock()
+				} else {
+					delete(c.flows[dstKey], srcOrFwd)
+					s.mu.Lock()
+					delete(s.flows[c][dstKey], srcOrFwd)
+					s.mu.Unlock()
+				}
```
bradfitz, line 873: It's really hard for me to convince myself that this locking is correct. It looks kinda bizarre.

derp/derp_server.go
```diff
@@ -162,6 +177,13 @@ type Server struct {
 	// src.
 	sentTo map[key.NodePublic]map[key.NodePublic]int64 // src => dst => dst's latest sclient.connNum
 
+	// flows is a map of all active flowStats on the server,
+	// protected by s.mu. After allocation by the server,
```
bradfitz, line 181: "by mu".  (there's no "s" in scope here)

derp/derp_server.go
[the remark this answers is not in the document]
bradfitz: Document why it's thread-safe & what protects what, not that it doesn't need to be. (it does need to be)

derp/derp_server.go
```diff
@@ -795,6 +825,54 @@ func (c *sclient) run(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+		c.cleanFlows()
+	}
+}
+
+// flowsTimerFunc tells all the sclients to scan for inactive flows
```
bradfitz, line 832: name this based on what it does, not the mechanism that invokes it.

derp/derp_server.go
```diff
@@ -1338,6 +1466,17 @@ type sclient struct {
...
+	// flows is a cache of the subset of flows statistics for
+	// which this client is either the src or the mesh peer
+	// forwarding for the src. The flowStats fields are changed by
+	// only the sclient read loop, but the flowStats must be
+	// allocated by the server.
+	flows map[key.NodePublic]map[key.NodePublic]*flowStats
+
+	// flowsCleanNow is set by the server to indicate it is time
+	// for this sclient to look for and free inactive flows.
+	flowsCleanNow bool
```
bradfitz, line 1479: this is guarded by what? s.mu also? It doesn't look consistently guarded elsewhere. flowsTimerFunc holds s.mu but cleanFlows reads it without any locking.

derp/derp_server.go
```diff
@@ -1338,6 +1466,17 @@ type sclient struct {
 	// client that it's trying to establish a direct connection
 	// through us with a peer we have no record of.
 	peerGoneLim *rate.Limiter
+
+	// flows is a cache of the subset of flows statistics for
+	// which this client is either the src or the mesh peer
+	// forwarding for the src. The flowStats fields are changed by
+	// only the sclient read loop, but the flowStats must be
+	// allocated by the server.
+	flows map[key.NodePublic]map[key.NodePublic]*flowStats
```
bradfitz, line 1475: what are the two maps keyed by?

derp/derp_server.go
```diff
@@ -944,6 +1024,48 @@ func (s *Server) notePeerSendLocked(src key.NodePublic, dst *sclient) {
...
+// forwarding the data from another mesh peer, then srcOrFwd is the
+// source key. If this sclient is forwarding the data to another mesh
+// peer, then srcOrFwd is the key of the mesh peer we are sending it
+// to. Otherwise, srcOrFwd is the zero value.
+func (c *sclient) updateFlowStats(dstKey, srcOrFwd key.NodePublic, contents []byte) {
+	var f *flowStats
+	// The common case is that the flowStat already exists
+	if _, ok := c.flows[dstKey]; ok {
+		f = c.flows[dstKey][srcOrFwd]
+	}
```
bradfitz, line 1060: It's well-defined in Go to read from a nil map: it returns the zero value for the map value type.

So this can be just `f := c.flows[dstKey][srcOrFwd]`

And then you also avoid the double map lookup.

derp/derp_server.go
```diff
@@ -944,6 +1024,48 @@ func (s *Server) notePeerSendLocked(src key.NodePublic, dst *sclient) {
...
+	if f == nil {
+		f = c.allocFlowStats(dstKey, srcOrFwd)
+	}
+	f.packets += 1
+	f.bytes += int64(len(contents))
```
bradfitz, line 1065: so these are incremented without holding any locks, but then you read/clone them in the snapshot without any locks too? that looks racy.

If you don't have tests that are failing in race mode, there probably isn't sufficient new test coverage.

(or I'm not understanding the locking yet)

bradfitz (changes requested):

Issue #8005, opened 2023-04-29
Debian/RPM packages for MIPS stopped after 1.36


kevinisageek:
### What is the issue?

Hey folks,

I know that EdgeOS isn't officially supported, however it has worked fine up until recently, following [jamesog's guide](https://github.com/jamesog/tailscale-edgeos). The machines page says there is an update available to 1.40.0 but SSH'ing into the edgerouter and upgrading says everything is up to date. I'm wondering if the Stretch repository needs the new version added to it? 

### Steps to reproduce

Standard `sudo apt update` and `sudo apt upgrade`

### Are there any recent changes that introduced the issue?

New tailscale version released

### OS

Linux

### OS version

EdgeOS v2.0.9-hotfix.6

### Tailscale version

1.36.2

### Other software

N/A

### Bug report

BUG-d817bfcb0a57e8b22b3df24fdd8799d186220f46b3924273e1f41e1a2c305f2c-20230429063528Z-def693c9d64b89fe

DentonGentry:
Was `apt install tailscale=1.36.2` ever done, locking it to a particular version?

The stretch repository does get updated for each release.

tgies:
Can confirm that the stretch repo, at least for mipsel, only has up to 1.36.2. This can be seen by examining the Packages manifest directly: https://pkgs.tailscale.com/stable/debian/dists/stretch/main/binary-mipsel/Packages

DentonGentry:
I think the MIPS support didn't get properly converted when the release/dist tooling was open sourced.
I believe the support needs to go here: https://github.com/tailscale/tailscale/blob/979d29b5f59fcf67d96d2dfd3bcc91c6fd912dc7/release/dist/unixpkgs/pkgs.go#L344

DentonGentry:
We are currently at the "how did this ever work?" stage of debugging. Even the prior distribution builder, which was in a private repo, doesn't look like it should have built deb/RPM packages for MIPS. Yet, apparently, it did.

danderson:
It ever worked because we accidentally started shipping MIPS debian packages at some point. When building cmd/dist, I stopped building them because I was pretty sure they weren't being built right (e.g. wrong architecture constants) : https://github.com/tailscale/tailscale/blob/979d29b5f59fcf67d96d2dfd3bcc91c6fd912dc7/release/dist/unixpkgs/targets.go#L90

[...]

PR #8104, opened 2023-05-09
release/dist: add deb/rpm arch mappings for mipses

saljmz:
According to https://wiki.debian.org/SupportedArchitectures Debian does not support big-endian mips64, so that one remains disabled.

According to https://fedoraproject.org/wiki/Architectures Fedora only supports little-endian mips, so leaving both big-endian ones out too.

Updates #8005.


DentonGentry:
> According to https://wiki.debian.org/SupportedArchitectures Debian does not support big-endian mips64, so that one remains disabled.

Many of the people reporting trouble on https://github.com/tailscale/tailscale/issues/8005 use an EdgeRouter, which is a big-endian mips64 (a Cavium CPU). Maybe Ubiquiti added big-endian support in their tree?

danderson (approved):
Thanks!

PR #8105, opened 2023-05-09
net/sockstats: wait before reporting battery usage

willnorris:
Wait 2 minutes before we start reporting battery usage. There is always radio activity on initial startup, which gets reported as 100% high power usage.  Let that settle before we report usage data.

Update tailscale/corp#9230

### DIFF
--- a/net/sockstats/sockstats_tsgo.go
+++ b/net/sockstats/sockstats_tsgo.go
@@ -325,6 +325,10 @@ type radioMonitor struct {
 // Usage is measured once per second, so this is the number of seconds of history to track.
 const radioSampleSize = 3600 // 1 hour
 
+// initStallPeriod is the minimum amount of time in seconds to collect data before reporting.
+// Otherwise, all clients will report 100% radio usage on startup.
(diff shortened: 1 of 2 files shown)


net/sockstats/sockstats_tsgo.go
```diff
@@ -321,9 +321,15 @@ type radioMonitor struct {
...
-// Usage is measured once per second, so this is the number of seconds of history to track.
-const radioSampleSize = 3600 // 1 hour
+const (
+	// radioSampleSize is the number of samples to store and report for cellular radio usage.
+	// Usage is measured once per second, so this is the number of seconds of history to track.
+	radioSampleSize = 3600 // 1 hour
+
+	// minPeriodLength is the minimum amount of time to collect data before reporting.
+	// Otherwise, all clients will report 100% radio usage on startup.
+	minPeriodLength = 120 // 2 minutes
```
dsnet, line 331: Optional: feels like a unspecific variable name. How about `initStallPeriod`?

dsnet (approved):

Issue #7560, opened 2023-03-14
The subnet router on Windows is so slow or unreliable that it is useless


strich:
### What is the issue?

I setup a simple Tailscale network between my machine and a remote server in Hetzner with a static IP. One of the reasons for doing this is because I need the static IP as an exit node for a specific subnet of public IPs as the remote company requires static allowed IPs. I thought Tailscale to be a perfect solution for this as I can reuse an existing server and help build up a secure internal network.

Unfortunately the performance and reliability of the subnet router for Windows is so bad that I'm unable to do the modest work I need to do with it. I thought maybe it is the remote company with the issue initially but after testing independent raw file downloads ([like this one](http://testfiles.hostnetworks.com.au/1000MB.iso)) I get speeds of 10+MB per sec going direct and 250KB per sec with the subnet router.

What's worse is that within a few minutes the transfer will just _stop_. I can see in TCPView is doesn't even disconnect. It just stops transferring data.

I know that on Windows Tailscale doesn't have very good performance currently, and I would have accepted modest speeds. But the feature is unusable at the moment.

Exit Node: Windows 10 latest
Client: Windows 11 latest

### OS

Windows

### OS version

Windows 10/11

### Tailscale version

1.36.2

DentonGentry:
The template asks for a bugreport to let us look at telemetry around the time of the problem. In this instance, a bugreport shortly after noting 250 KBps would be useful.

My suspicion is that the subnet router is not able to establish direct connections so is going through DERP relays https://tailscale.com/kb/1232/derp-servers/

You can verify this in two ways:
1. `tailscale ping` will say whether it found a direct path
2. `tailscale status` will show whether an existing connection is direct or relayed

```
$ tailscale status
100.100.1.2  dgentry-sfo1         dgentry@       linux   active; offers exit node; direct 1.2.3.4:60350
100.100.1.3  dgentry-tor1         dgentry@       linux   active; offers exit node; relay "tor"
```

strich:
My apologies I'm not sure why I didn't see the template requirements.

I did already check ping and status in my troubleshooting and it is a direct connection - I have updated the OP with it.

How do I get the telemetry?

strich:
I've just recorded a bug report on both nodes and included them in OP. Though after I did so I tried looking at the log files locally and they were empty? 

[...]

PR #8106, opened 2023-05-09
wgengine/netstack: switch to cubic congestion control

kylecarbs:
A retransmission issue surfaces in gVisor under Reno congestion control when dispatching a rapid sequence of small packets. This bug can be reliably replicated using tsnet with the patch delineated below:

```diff
diff --git a/tsnet/tsnet_test.go b/tsnet/tsnet_test.go
index 5ab27099..d9c2f51b 100644
--- a/tsnet/tsnet_test.go
+++ b/tsnet/tsnet_test.go
@@ -261,24 +261,20 @@ func TestConn(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-
-	want := "hello"
-	if _, err := io.WriteString(w, want); err != nil {
-		t.Fatal(err)
-	}
-
-	got := make([]byte, len(want))
-	if _, err := io.ReadAtLeast(r, got, len(got)); err != nil {
-		t.Fatal(err)
-	}
-	t.Logf("got: %q", got)
-	if string(got) != want {
-		t.Errorf("got %q, want %q", got, want)
-	}
-
-	_, err = s2.Dial(ctx, "tcp", fmt.Sprintf("%s:8082", s1ip)) // some random port
-	if err == nil {
-		t.Fatalf("unexpected success; should have seen a connection refused error")
+	defer r.Close()
+	go io.Copy(io.Discard, r)
+
+	t.Log("writing")
+
+	payload := []byte(strings.Repeat("hello world\n", 65536/12))
+	size := 0
+	for i := 0; i < 1024*2; i++ {
+		t.Logf("write payload %d: %d bytes", i, size/1024)
+		n, err := w.Write(payload)
+		if err != nil {
+			t.Fatal(err)
+		}
+		size += n
 	}
 }
```

Execute the given test multiple iterations, and you'll observe it becomes unresponsive during payload writing. Under the hood, gVisor is invoking the following function:

https://github.com/google/gvisor/blob/0b76fe6c0038e71a3f9dd458193e6bcf92430f13/pkg/tcpip/transport/tcp/snd.go#L425-L435

After changing the congestion control to cubic (as this PR does), the behavior cannot be replicated. I'm unfamiliar with the implications of adjusting the congestion control to cubic, so guidance would be helpful here.

Related: https://github.com/tailscale/tailscale/issues/7560

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -178,6 +178,11 @@ func Create(logf logger.Logf, tundev *tstun.Wrapper, e wgengine.Engine, mc *magi
 	if tcpipErr != nil {
 		return nil, fmt.Errorf("could not enable TCP SACK: %v", tcpipErr)
 	}
+	congestionOpt := tcpip.CongestionControlOption("cubic") // Reno is used by default
+	tcpipErr = ipstack.SetTransportProtocolOption(tcp.ProtocolNumber, &congestionOpt)
+	if tcpipErr != nil {
+		return nil, fmt.Errorf("could not set TCP congestion control: %v", tcpipErr)
+	}
 	linkEP := channel.New(512, tstun.DefaultMTU(), "")
 	if tcpipProblem := ipstack.CreateNIC(nicID, linkEP); tcpipProblem != nil {
 		return nil, fmt.Errorf("could not create netstack NIC: %v", tcpipProblem)


jwhited:
> Execute the given test multiple iterations, and you'll observe it becomes unresponsive during payload writing. Under the hood, gVisor is invoking the following function:

Do you have a goroutine stack from when it stalls?

> After changing the congestion control to cubic (as this PR does), the behavior cannot be replicated. I'm unfamiliar with the implications of adjusting the congestion control to cubic, so guidance would be helpful here.

We've done some preliminary testing with gVisor cubic in a performance context. The results did not make it clear we should make the switch. The motivation in this PR is a bit different.

kdrag0n:
> We've done some preliminary testing with gVisor cubic in a performance context. The results did not make it clear we should make the switch. The motivation in this PR is a bit different.

@jwhited Sorry for the slight off-topic, but would you be willing to share more details about cubic performance results by any chance? I agree that it's not a solution to this issue, but it'd be really helpful to narrow down some other gVisor TCP behavior I've been investigating recently.

kylecarbs:
Feel free to follow along here: https://github.com/google/gvisor/issues/6113

There is a real bug here, but it seems to exist in gVisor with `TCPRACKLossDetection`.

PR #8108, opened 2023-05-09
tool/gocross: fix incorrect relpath usage in bootstrap script

danderson:
The subshell in which gocross gets built cd's to the corp checkout dir near the top, so all future references to corp repository files should be simple relative paths, and not reference $repo_root. When $repo_root is an absolute path, it doesn't matter and everything works out, but on some OSes and shells and invocations, $repo_root is a completely relative path that is invalidated by the "cd".

Fixes tailscale/corp#11183

---

Woops, fixed this bug in the wrong repo, so our repo sync logic reverted it. Fixing in the right place this time.


creachadair (approved):

creachadair:
But note the typo in the description 🙂

danderson:
Fixed, thanks!

Issue #8036, opened 2023-05-03
clean up tailscaled initialization, subsystems


bradfitz:
The relationship and setup between all the tailscaled subsystems is very hairy.

There's a lot of shared and nearly duplicate code between tailscaled, tailscaled as a Windows service, tsnet, and the macOS GUI. Not to mention all the tests.

And then we have a bunch of weird optional interfaces to expose internals from things like wgengine.Engine that really don't belong there or make any sense design-wise.

I tried to start cleaning this up in #4754 but it was a bad time of the release cycle and we hated the name and it's since bit rot.

I'm going to try again.

The first goal is to just remove the optional interfaces and add some strictness (can only set something once, require things to be non-nil, etc). Future changes can then start unifying some of the setup between platforms.

/cc @maisem

PR #8109, opened 2023-05-10
cmd/tsconnect: fix forgotten API change for wasm

bradfitz:
Fix regression from 6e967446e41

Updates #8036


maisem (approved):

maisem:
Can we add some test?

bradfitz:
> Can we add some test?

I'd love that but don't know how. Wasm's no longer really staffed :)

Issue #5853, opened 2022-10-05
Replace fallback DNS with recursive DNS resolver


bradfitz:
We've discussed adding a recursive DNS resolver to tailscaled for various things. Originally @danderson proposed it as a means to do DNS resolution when we're in a sandbox and don't have permission to read the host operating system's DNS settings and we don't want to hard-code any public well-known recursive resolver.

One thing we could use a recursive DNS resolver for is replacing our fallbackdns (bootstrap DNS) so we don't need use our existing DERP servers.

/cc @DentonGentry @maisem @andrew-d @crawshaw

danderson:
The only scenario I can think of where that won't work out, is on networks that block udp/53 traffic to everything except the local resolver. Right now in those networks, with a misconfigured client, we're still able to get out to the internet over HTTPS and get a good resolution to fix up the local config. A recursive resolver would fail and leave the machine stranded.

OTOH, it would save us from the headaches of having hardcoded bootstrap maps in old binaries, and having to maintain moribund servers for ever to keep serving those clients.

bradfitz:
Another thing to consider: we have IP space we own these days and could start anycasting it for bootstrap DNS. Then at least we're only baking in one/two IPs. It doesn't have to even be particularly good anycast if it's only for bootstrap DNS.

danderson:
Yeah, anycasting to our derps (or something else) would help future-proof things. It means we're committed to owning our AS and IP space "for ever" but that's significantly easier to do.

DentonGentry:
Related: https://github.com/tailscale/tailscale/issues/3507

DentonGentry:
(we used to have an entirely separate concept of Fallback Resolvers and it keeps weirding me out)

PR #8115, opened 2023-05-10
net/dns/recursive: add initial implementation of recursive DNS resolver

andrew-d:
We've talked in the past about reworking how bootstrap DNS works to instead do recursive DNS resolution from the root; this would better support on-prem customers and Headscale users where the DERP servers don't currently resolve their DNS name. This package is an initial implementation of recursive resolution for A and AAAA records.

Updates #5853

Change-Id: Ibe974d78709b4b03674b47c4ef61f9a00addf8b4

### DIFF
--- a/net/dns/recursive/recursive.go
+++ b/net/dns/recursive/recursive.go
@@ -1,0 +1,640 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+// Package recursive implements a simple recursive DNS resolver.
+package recursive
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"net/netip"
+	"strings"
+	"time"
+
+	"github.com/miekg/dns"
+	"golang.org/x/exp/constraints"
+	"golang.org/x/exp/slices"
+	"tailscale.com/envknob"
+	"tailscale.com/net/netns"
+	"tailscale.com/types/logger"
+	"tailscale.com/util/dnsname"
+	"tailscale.com/util/mak"
+	"tailscale.com/util/multierr"
+	"tailscale.com/util/slicesx"
+)
+
+const (
+	// maxDepth is how deep from the root nameservers we'll recurse when
+	// resolving; passing this limit will instead return an error.
+	//
+	// maxDepth must be at least 20 to resolve "console.aws.amazon.com",
+	// which is a domain with a moderately complicated DNS setup. The
+	// current value of 30 was chosen semi-arbitrarily to ensure that we
+	// have about 50% headroom.
+	maxDepth = 30
+	// numStartingServers is the number of root nameservers that we use as
+	// initial candidates for our recursion.
+	numStartingServers = 3
+	// udpQueryTimeout is the amount of time we wait for a UDP response
+	// from a nameserver before falling back to a TCP connection.
+	udpQueryTimeout = 5 * time.Second
+
+	// These constants aren't typed in the DNS package, so we create typed
+	// versions here to avoid having to do repeated type casts.
+	qtypeA    dns.Type = dns.Type(dns.TypeA)
+	qtypeAAAA dns.Type = dns.Type(dns.TypeAAAA)
+)
+
+var (
+	// ErrMaxDepth is returned when recursive resolving exceeds the maximum
+	// depth limit for this package.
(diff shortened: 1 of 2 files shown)


net/dns/recursive/recursive_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+	"math/rand"
+	"testing"
+	"time"
+
+	"tailscale.com/envknob"
+)
+
+func init() {
+	rand.Seed(time.Now().UnixMicro())
```
raggi, line 16: We had a test recently that was dependent upon random seeding and it was flaking for some cases, with no control over it. I'm thinking about adding a package to make this easier, but in the meantime could you just avoid the seed here?

andrew-d: Huh! Sure, happy to do that; this test won't fail no matter what the random seed is, since it's only used to shuffle lists, but happy to remove it since it's also not necessary.

andrew-d:
Moving to draft until I get CNAME support properly working

andrew-d:
Okay, this can now recursively resolve all of the hostnames I've tried it with, and we have unit tests that cover all the major behaviour, and an integration test that makes real queries to real nameservers successfully. There's a couple of TODOs, but this should be ready for an initial review.

net/dns/recursive/recursive.go
```diff
@@ -0,0 +1,633 @@
...
+	for _, rr := range resp.Extra {
+		minTTL = min(minTTL, int(rr.Header().Ttl))
+	}
+
+	mak.Set(&r.queryCache, cacheKey, dnsMsgWithExpiry{
+		Msg:       resp,
+		expiresAt: now.Add(time.Duration(minTTL) * time.Second),
+	})
+
+	//r.depthlogf(depth,"response from %s: %+v", nameserverStr, resp)
```
bradfitz, line 631: dd

net/dns/recursive/recursive.go
```diff
@@ -0,0 +1,633 @@
...
+	return &queryState{
+		rootServers: rootServers,
+	}
+}
+
+// Resolve will perform a recursive DNS resolution for the provided name,
+// starting at a randomly-chosen root DNS server, and return the A and AAAA
+// responses as a slice of netip.Addrs along with the minimum TTL for the
+// returned records.
+func (r *Resolver) Resolve(ctx context.Context, name string) ([]netip.Addr, time.Duration, error) {
```
bradfitz, line 224: name at least Duration

net/dns/recursive/recursive.go
```diff
@@ -0,0 +1,633 @@
...
+
+	slicesx.Shuffle(addrs)
+	return addrs, minTTL, nil
+}
+
+func (r *Resolver) resolveRecursiveFromRoot(
+	ctx context.Context,
+	qstate *queryState,
+	depth int,
+	name string, // what we're querying
```
bradfitz, line 271: // what we're querying ("controlplane.tailscale.com.")

(e.g. with trailing dot?)

net/dns/recursive/recursive.go
```diff
@@ -0,0 +1,633 @@
...
+		switch v := rr.(type) {
+		case *dns.NS:
+			authorities = append(authorities, v.Ns)
+		}
+	}
+
+	// Also check for "glue" records, which are IP addresses provided by
+	// the DNS server for authority responses; these are required when the
+	// authority server is a subdomain of what's being resolved.
+	glueRecords := make(map[string][]netip.Addr)
```
bradfitz, line 411: give example of key

net/dns/recursive/recursive.go
```diff
@@ -0,0 +1,633 @@
...
+			ip, ok := netip.AddrFromSlice(v.AAAA)
+			if !ok || !ip.Is6() {
+				r.logf("unexpected bad Extra IPv6 addr")
+				continue
+			}
+			glueRecords[v.Hdr.Name] = append(glueRecords[v.Hdr.Name], ip)
+		default:
+			continue
+		}
+	}
```
bradfitz, line 431: pull out this, other stuff into little helper funcs/methods?

bradfitz (approved):

net/dns/recursive/recursive.go
andrew-d: Converted to `dnsname.FQDN` for lack of ambiguity

PR #8117, opened 2023-05-10
logtail: be less aggressive about re-uploads

dsnet:
The retry logic was pathological in the following ways:

* If we restarted the logging service, any pending uploads
would be placed in a retry-loop where it depended on backoff.Backoff,
which was too aggresive. It would retry failures within milliseconds,
taking at least 10 retries to hit a delay of 1 second.

* In the event where a logstream was rate limited,
the aggressive retry logic would severely exacerbate the problem
since each retry would also log an error message.
It is by chance that the rate of log error spam
does not happen to exceed the rate limit itself.

We modify the retry logic in the following ways:

* We now respect the "Retry-After" header sent by the logging service.

* Lacking a "Retry-After" header, we retry after a hard-coded period of
30 to 60 seconds. This avoids the thundering-herd effect when all nodes
try reconnecting to the logging service at the same time after a restart.

* We do not treat a status 400 as having been uploaded.
This is simply not the behavior of the logging service.

Updates #https://github.com/tailscale/corp/pull/11213


logtail/logtail.go
```diff
@@ -433,7 +441,7 @@ func (l *Logger) awaitInternetUp(ctx context.Context) {
 // upload uploads body to the log server.
 // origlen indicates the pre-compression body length.
 // origlen of -1 indicates that the body is not compressed.
-func (l *Logger) upload(ctx context.Context, body []byte, origlen int) (uploaded bool, err error) {
+func (l *Logger) upload(ctx context.Context, body []byte, origlen int) (retryAfter int, err error) {
```
willnorris, line 444: totally up to you, but maybe return a `time.Duration` so the units are abundantly clear?

willnorris (approved):

logtail/logtail.go
dsnet: Done.

logtail/logtail.go
```diff
@@ -373,23 +368,36 @@ func (l *Logger) uploading(ctx context.Context) {
...
+					fmt.Fprintf(l.stderr, "logtail: upload: %v\n", err)
+					lastError = currError
+				}
+
+				// Sleep for the specified retryAfter period.
+				// If unspecified, default to some random value.
+				if retryAfter <= 0 {
+					retryAfter = time.Duration(30+mrand.Intn(30)) * time.Second
+				}
+				tstime.Sleep(ctx, time.Duration(retryAfter)*time.Second)
```
willnorris, line 395: don't need this time.Duration anymore then

dsnet: Ooopfh. Good catch.

Issue #4945, opened 2022-06-27
ssh/tailssh: Cannot login as LDAP user on Linux host


DilumAluthge:
### What is the issue?

I've set up Tailscale SSH on a Linux server where most of the users are from LDAP. For example:

```
# cat /etc/passwd | grep jsmith
# cat /etc/shadow | grep jsmith
# getent passwd | grep jsmith
jsmith:x:123456789:123456789:Smith, John:/home/jsmith:/bin/bash
```

When I try to SSH in as an LDAP user, I am denied with the following message:

```
$ ssh jsmith@myserver
failed to lookup jsmith
jsmith@myserver: Permission denied (tailscale).
```

I'm guessing that this corresponds to an error being returned by this call to `user.Lookup`:

https://github.com/tailscale/tailscale/blob/1d04e01d1e62e0f83a24c0e2963c1b74db33c00d/ssh/tailssh/tailssh.go#L208-L215

If I try to SSH in as a local (non-LDAP) user, it works successfully.

### Steps to reproduce

To reproduce this issue, first make sure that `LDAPUSER` is an LDAP user on your server. Then run:

```
ssh LDAPUSER@myserver
```

### Are there any recent changes that introduced the issue?

I recently did the following:
1. Updated Tailscale (by doing `sudo yum update tailscale`).
2. Ran `sudo tailscale up --ssh` to start the Tailscale SSH server.

### OS

Linux

### OS version

CentOS Linux release 7.9.2009 (Core)

### Tailscale version

1.26.1

### Bug report

BUG-53b053b3be60de1d1ace2edd177cb07c87b1f3b1a4c49b3b307d74b854fe7801-20220627044844Z-95747bf68628c167

### Other information

I installed Tailscale by doing adding the Tailscale repository and then doing `sudo yum install tailscale`.

```
# tailscale --version
1.26.1
  tailscale commit: b3687b70758d29f5fd6021c34d58eba7381812c0
  other commit: bcd70ce328082f464f89735d9adfa5bcd90ccf7e
  go version: go1.18.3-ts04d67b90d8
```

```
# cat /etc/system-release
CentOS Linux release 7.9.2009 (Core)
# uname -a
Linux myserver 3.10.0-1160.59.1.el7.x86_64 #1 SMP Wed Feb 23 16:47:03 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

bradfitz:
[...]

PR #8120, opened 2023-05-11
Do not set cmd.Dir when spawning tailscaled ssh command

### LINKED ISSUE #8118: SSH fails with `${HOME}` permission `drwxr-s---`
### What is the issue?

When I ssh to server over tailnet I receive Connection to 100.106.100.25 closed immediately.

I expect authentication to succeed like it does for normal sshd.

Error is the following:
```
May 10 20:36:04 qqqqq tailscaled[10697]: ssh-session(sess-20230511T033604-7bc50db3b2): access granted to user@user.com as ssh-user "ldaptestuser"
May 10 20:36:04 qqqqq tailscaled[10697]: ssh-session(sess-20230511T033604-7bc50db3b2): starting pty command: [/usr/sbin/tailscaled be-child ssh --uid=4007 --gid=2564 --groups=2564 --local-user=ldaptestuser --remote-user=user@user.com>
May 10 20:36:04 qqqqq tailscaled[10697]: ssh-session(sess-20230511T033604-7bc50db3b2): start failed: fork/exec /usr/sbin/tailscaled: permission denied
```

### Steps to reproduce

System is configured with Redhat authselect, configured against LDAP.

I am using latest source code which solves user not found issue from #8098. Next issue is permission related.
Git commit cb2fd5be922eb048c2f4f85a529082b1913a9dbd

Built using command:
```bash
TS_USE_TOOLCHAIN=1 ./build_dist.sh tailscale.com/cmd/tailscaled
```

### OS

Linux

### OS version

Linux 5.14.0-162.6.1.el9_1.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Sep 30 07:36:03 EDT 2022 x86_64 x86_64 x86_64 GNU/Linux

### Tailscale version

1.41.55

### Other software

Redhat SSSD and Authselect
SELINUX is disabled

### Bug report

BUG-6bbc532ccfa255db2925d630b88107fec4525ee345569ffe7197be1f3431325c-20230511033358Z-9019d864fc06f0be

derek-burdick:
Leave cmd.Dir empty, which should configure spawned process to use working directory of existing process.

When a user logs in, their directory is set based on shell semantics and is not related to the working directory of 'tailscaled ssh' command, based on my testing.

Fixes #8118
Related to pull #4945

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -113,6 +113,7 @@ func (ss *sshSession) newIncubatorCommand() (cmd *exec.Cmd) {
 		"--remote-ip=" + ci.src.Addr().String(),
 		"--has-tty=false", // updated in-place by startWithPTY
 		"--tty-name=",     // updated in-place by startWithPTY
+		"--pwd=" + ss.conn.localUser.HomeDir,
 	}
 
 	if isSFTP {
@@ -177,6 +178,7 @@ type incubatorArgs struct {
 	isShell      bool
 	loginCmdPath string
 	cmdArgs      []string
+	pwd          string
 }
 
 func parseIncubatorArgs(args []string) (a incubatorArgs) {
@@ -193,6 +195,7 @@ func parseIncubatorArgs(args []string) (a incubatorArgs) {
 	flags.BoolVar(&a.isShell, "shell", false, "is launching a shell (with no cmds)")
 	flags.BoolVar(&a.isSFTP, "sftp", false, "run sftp server (cmd is ignored)")
 	flags.StringVar(&a.loginCmdPath, "login-cmd", "", "the path to `login` cmd")
+	flags.StringVar(&a.pwd, "pwd", "/", "process initial working directory, if possible. else / is used")
 	flags.Parse(args)
 	a.cmdArgs = flags.Args()
 	return a
@@ -278,6 +281,12 @@ func beIncubator(args []string) error {
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	cmd.Env = os.Environ()
+
+	if _, err := os.Stat(ia.pwd); err != nil && os.IsNotExist(err) {
+		cmd.Dir = "/"
+	} else {
+		cmd.Dir = ia.pwd
+	}
 
 	if ia.hasTTY {
 		// If we were launched with a tty then we should
@@ -428,16 +437,7 @@ func (ss *sshSession) launchProcess() error {
 	ss.cmd = ss.newIncubatorCommand()
 
 	cmd := ss.cmd
-	homeDir := ss.conn.localUser.HomeDir
-	if _, err := os.Stat(homeDir); err == nil {
-		cmd.Dir = homeDir
-	} else if os.IsNotExist(err) {
-		// If the home directory doesn't exist, we can't chdir to it.
-		// Instead, we'll chdir to the root directory.
-		cmd.Dir = "/"
-	} else {
-		return err
-	}
+
 	cmd.Env = envForUser(ss.conn.localUser)
 	for _, kv := range ss.Environ() {
 		if acceptEnvPair(kv) {


ssh/tailssh/incubator.go
```diff
@@ -222,13 +222,23 @@ func beIncubator(args []string) error {
 		}
 	}
 
+	homeDir := "/"
+	if userMeta, err := userLookupGetent(ia.localUser); err == nil {
```
bradfitz, line 226: This file is cross-platform (see the line near the top: `//go:build linux || (darwin && !ios) || freebsd || openbsd`) but this userLookupGetEnt should only be used on Linux.

ssh/tailssh/incubator.go
```diff
@@ -222,13 +222,23 @@ func beIncubator(args []string) error {
...
+	}
+
 	euid := os.Geteuid()
 	runningAsRoot := euid == 0
 	if runningAsRoot && ia.loginCmdPath != "" {
 		// Check if we can exec into the login command instead of trying to
 		// incubate ourselves.
 		if la := ia.loginArgs(); la != nil {
-			return unix.Exec(ia.loginCmdPath, la, os.Environ())
+			// PWD is used by unix.Exec to set the directory
```
bradfitz, line 239: That's not true. Not in Go (see the code in x/sys/unix and syscall) nor documented in the exec manpages.

bradfitz: I guess you mean by the thing that was execed (the login cmd).

ssh/tailssh/user.go
```diff
@@ -74,11 +74,18 @@ func userLookup(uid string) (*userMeta, error) {
 }
 
 func validUsername(uid string) bool {
-	if len(uid) > 32 || len(uid) == 0 {
+	maxUid := 32
+	if runtime.GOOS == "linux" {
+		maxUid = 256
+	}
```
bradfitz, line 80: this part seems like a fine change but unrelated? send as its own PR?

ssh/tailssh/incubator.go
```diff
@@ -222,13 +222,23 @@ func beIncubator(args []string) error {
 		}
 	}
 
+	homeDir := "/"
```
maisem, line 225: I would be more comfortable if we passed down the desired `homeDir` from the parent process as an argument.

ssh/tailssh/user.go
derek-burdick: Will revert and open separate PR.

derek-burdick: https://github.com/tailscale/tailscale/pull/8278

ssh/tailssh/incubator.go
derek-burdick: Yes, was referring to login cmd. Based on interactive and non-interactive commands it worked. I can test SSHD as well.

ssh/tailssh/incubator.go
derek-burdick: Got it. I will refactor.

ssh/tailssh/incubator.go
```diff
@@ -184,6 +186,7 @@ func parseIncubatorArgs(args []string) (a incubatorArgs) {
 	flags.BoolVar(&a.isShell, "shell", false, "is launching a shell (with no cmds)")
 	flags.BoolVar(&a.isSFTP, "sftp", false, "run sftp server (cmd is ignored)")
 	flags.StringVar(&a.loginCmdPath, "login-cmd", "", "the path to `login` cmd")
+	flags.StringVar(&a.homeDir, "home-dir", "/", "the path to user `$HOME` directory")
```
bradfitz, line 189: there's nothing really "home" about this from the incubator's perspective. This is just the dir it uses as its pwd. The fact that it might've come from $HOME originally is just a detail for the incubator. How about we just rename it pwd? I can do that for you.

bradfitz (approved):

derek-burdick:
@bradfitz  This may also break SFTP default directory. I should have a fix for tests and will validate SFTP by end of week.

PR #8122, opened 2023-05-11
util/set: add a basic map-based Set type

bradfitz:
We have two other types of Sets here. Add the basic obvious one too.

Needed for a change elsewhere.

Updates #cleanup

### DIFF
--- a/util/set/set.go
+++ b/util/set/set.go
@@ -4,6 +4,21 @@
 // Package set contains set types.
 package set
 
+// Set is a set of T.
+type Set[T comparable] map[T]struct{}
+
+// Add adds e to the set.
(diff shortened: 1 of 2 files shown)


util/set/set.go
```diff
@@ -4,6 +4,21 @@
 // Package set contains set types.
 package set
 
+// Set is a set of T.
+type Set[T comparable] map[T]struct{}
```
maisem, line 8: make this a struct?

maisem (approved):

util/set/set.go
bradfitz: naah. I want people to be able to range over it still.

I just want it for better declarations so I know people's intent from the signature/fields

PR #8124, opened 2023-05-12
cmd/tailscale/cli: [up] fix CreateKey missing argument

shayne:

### DIFF
--- a/cmd/tailscale/cli/up.go
+++ b/cmd/tailscale/cli/up.go
@@ -1198,7 +1198,8 @@ func resolveAuthKey(ctx context.Context, v, tags string) (string, error) {
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/tailscale/cli/up.go
```diff
@@ -1198,7 +1198,7 @@ func resolveAuthKey(ctx context.Context, v, tags string) (string, error) {
 		},
 	}
 
-	authkey, _, err := tsClient.CreateKey(ctx, caps)
+	authkey, _, err := tsClient.CreateKey(ctx, caps, time.Duration(0))
```
maisem, line 1201: I think you can just do 0. Maybe you can declare a const to make it clear?

```
const defaultExpiry = 0
```

maisem (approved):

maisem (approved):

PR #8129, opened 2023-05-13
client/tailscale: revert CreateKey API change, add Client.CreateKeyWithExpiry

bradfitz:
The client/tailscale is a stable-ish API we try not to break. Revert the Client.CreateKey method as it was and add a new CreateKeyWithExpiry method to do the new thing. And document the expiry field and enforce that the time.Duration can't be between in range greater than 0 and less than a second.

Updates #7143
Updates #8124 (reverts it, effectively)

### DIFF
--- a/client/tailscale/keys.go
+++ b/client/tailscale/keys.go
@@ -68,14 +68,26 @@ func (c *Client) Keys(ctx context.Context) ([]string, error) {
 }
 
 // CreateKey creates a new key for the current user. Currently, only auth keys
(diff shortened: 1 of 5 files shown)


DentonGentry (approved):

DentonGentry (approved):

maisem (approved):

danderson (approved):
cmd/k8s-operator needs fixups as it was already patched to fix the API change breakage.

Issue #8119, opened 2023-05-11
Each ssh connection creates a new PTS


fritterhoff:
### What is the issue?

Using tailscale ssh with an linux target creates a new TTY/PTS session even after disconnect. 
In case of a long runtime this causes some sort of exhausting/overflow of sessions and a new login is not possible any more. 

I'm not sure which logs are helpful.  

### Steps to reproduce

- Connect to host using Tailscale SSH (pts/n)
- Disconnect
- Connect to same host using Tailscale SSH 
- New Session with new/next pts (pts/n+1)

### OS

Linux

### OS version

Debian 10 & 11

### Tailscale version

1.40.0

### Bug report

BUG-e8cd3deb76c9799e8f24f364b1415730a05d3dcb7472bc901d652139ee8853c4-20230511101031Z-a7692b90016dfd95

DentonGentry:
Is this new behavior in 1.40.1? Did it not happen in 1.40.0?

fritterhoff:
Hi,
this bug exists since I am using tailscale. At the moment the most servers are using 1.14.0. From my point of view it exists at least since version 1.36.0 until 1.14.1.

Until now it only was some sort of optical issue. Now I am using tailscale in combination with Ansible and CI/CD and create several new ssh session in a few sessions so I exceed the default limit of 4096 sessions after a few days.

fritterhoff:
P.S. Since this sort of breaks the access after 4096 connections I don't think this is only a usability issue.

fritterhoff:
So I started digging a little bit into the code of the tailscale ssh server. From my point of view this is caused due to open file handles.

fritterhoff:
To illustrate it a little bit a just added some logging information:
https://github.com/fritterhoff/tailscale/commit/b041ea2c3e5b685a72579994571f1c4fcdbef497

After exiting the ssh session i would expect both lines to appear. 

In reality I only get the log line for stdin. Since stdin and stdout both point to `/dev/ptmux` this could be okay and is maybe only an minor "bug"?  

On the other side the stdin and stdout of the cmd itself seem to be kept open even after exiting the ssh session so `/dev/pts/X` stays alive. Obviously this is caused due to the missing `Close()` call on the attached TTY? 

```
root@pve01:~# lsof -a -p 1876627
COMMAND       PID USER   FD      TYPE             DEVICE SIZE/OFF       NODE NAME
tailscale 1876627 root  cwd       DIR              253,0     4096          2 /
[...]

PR #8133, opened 2023-05-14
ssh/tailssh: close tty after process has finished

fritterhoff:
More or less a workaround for https://github.com/tailscale/tailscale/issues/8119

Probably there is a better solution instead of catching the error inside the goroutine and ignoring it.


bradfitz:
Looking briefly on my phone, seems okay in principle but I think there might be a data race on that new bool. Will look more later.

fritterhoff:
Ah... That's a good point. Maybe there is a better solution 🤔

bradfitz:
@maisem said he'd do this review.

maisem:
This is what I had in mind. https://github.com/tailscale/tailscale/pull/8157

Issue #6096, opened 2022-10-27
FR:  reduce QR code size 


orent:
### What are you trying to do?

Terminals are sometimes small. The 80x25 console of a VM, for example.  The 29x29 QR code with its generous margins can easily overflow such small displays, making it hard to scan. 

### How should we solve this?

HTTPS://XYZ.TLD/12+3:AB.C

This URL encodes to a 21x21 code using the the QR alphanumeric character set at 4.5 bits per char. The 48 bit login code is reduced from 12 hex digits to 9 base41 symbols (A-Z0-9 and 5 more chars from the QR alphanumeric set such as +-*.: )

The domain can be up to 7 chars (ABCD.CC or XYZ.TLD). It should not be expensive or hard to get (as I imagine ts.net probably was). 

### Anything else?

Uppercase url schema and dns names are RFC compliant and I am not aware of any device that has problems with opening these urls. I have encountered uppercase URLs like these “in the wild”

DentonGentry:
To make sure I understand: you are suggesting that Tailscale purchase a short domain name in order to make the QR code smaller?

bradfitz:
FWIW we could make it a bit smaller if we wanted: https://go.dev/play/p/trWH82MUykI

orent:
Not a bit smaller - a minimum size 21x21 code. You probably haven’t seen many of those.

If you increase the length by just 1 or use any char outside the limited QR alphanumeric charset it will switch to the next larger size (25x25).

A domain like A.TS.NET is 8 chars. You need 7 or less. Alternatively, you can reduce the login token size a bit below 48 bits.

orent:
Proposed domain: TAIL.SC

bradfitz:
Gotcha. Or just TS.NET. Updated example: https://go.dev/play/p/ugA6TB5uM0u

bradfitz:
We're about to launch a public landing page at https://ts.net/ that'll mostly be a redirect anyway, so paths starting with `/A` could just be another type of redirect.

orent:
Alternatively, if you are not into such extreme compression, uppercase, etc the QR code can be rendered at half size using the following characters:

* U+2580 | ▀ | Upper half block
* U+2584 | ▄ | Lower half block
* U+2588 | █ | Full block

You can also do both, of course.

sfllaw:
> Alternatively, if you are not into such extreme compression, uppercase, etc the QR code can be rendered at half size using the following characters:

@orent, that is #17084 will switch to.

PR #8136, opened 2023-05-15
cmd/tailscale/cli: [up] compact QR code

orent:
Reduce QR code from 35 to 15 lines by using half-block characters, disabling border and switching to negative.

See #6096 for potentially of reducing it further (11 lines)


bradfitz:
Does this reduce compatibility on lesser capable terminals, such as cmd.exe on Windows?

orent:
> Does this reduce compatibility on lesser capable terminals, such as cmd.exe on Windows?

If a terminal shows the big representation correctly this means the terminal already supports utf8 and the unicode full block character and should also support the adjacent half-block characters.

It’s conceivable that in some cases this was shown incorrectly as a unicode question mark replacement character but was still somehow scannable. In that case the compact version would fail.

But if the code just does not fit in a 80x25 console or in a tiny web shell window it is guaranteed to fail scanning and I encountered this several times

PR #8140, opened 2023-05-15
cmd/tailscaled, util/winutil: log our registry keys during tailscaled…

dblohm7:
… startup

In order to improve our ability to understand the state of policies and registry settings when troubleshooting, we enumerate all values in all subkeys. x/sys/windows does not already offer this, so we need to call RegEnumValue directly.

For now we're just logging this during startup, however in a future PR I plan to also trigger this code during a bugreport. I also want to log more than just registry.


util/winutil/winutil_windows.go
```diff
@@ -551,3 +555,164 @@ func findHomeDirInRegistry(uid string) (dir string, err error) {
...
+	initialValueBufLen = 80    // large enough to contain a stringified GUID encoded as UTF-16
+)
+
+const (
+	supportInfoKeyRegistry = "Registry"
+)
+
+// LogSupportInfo obtains information useful for troubleshooting and support,
+// and writes it to the log as a JSON-encoded object.
+func LogSupportInfo() {
```
andrew-d, line 571: If you make this take a `logf logger.Logf` parameter, it'd be pretty trivial to integrate into the bugreport machinery in a follow-up PR.

util/winutil/winutil_windows.go
```diff
@@ -551,3 +555,164 @@ func findHomeDirInRegistry(uid string) (dir string, err error) {
...
+		switch valueType {
+		case registry.SZ, registry.EXPAND_SZ:
+			value = windows.UTF16PtrToString((*uint16)(unsafe.Pointer(&vbuf[0])))
+		case registry.BINARY:
+			if valueLen > maxBinaryValueLen {
+				valueLen = maxBinaryValueLen
+			}
+			value = append([]byte{}, vbuf[:valueLen]...)
+		case registry.DWORD:
+			value = binary.LittleEndian.Uint32(vbuf[:4])
```
andrew-d, line 679: Are we guaranteed to get at least 4 bytes here (and thus not panic)?

util/winutil/winutil_windows.go
```diff
@@ -551,3 +555,164 @@ func findHomeDirInRegistry(uid string) (dir string, err error) {
...
+				for i, c := range p {
+					if c == 0 {
+						strs = append(strs, string(utf16.Decode(p[from:i])))
+						from = i + 1
+					}
+				}
+			}
+			value = strs
+		case registry.QWORD:
+			value = binary.LittleEndian.Uint64(vbuf[:8])
```
andrew-d, line 699: Ditto above?

util/winutil/winutil_windows.go
```diff
@@ -551,3 +555,164 @@ func findHomeDirInRegistry(uid string) (dir string, err error) {
...
+			value = fmt.Sprintf("<unsupported value type %d>", valueType)
+		}
+
+		kv[windows.UTF16PtrToString(&nbuf[0])] = value
+		index++
+	}
+
+	if sks, err := k.ReadSubKeyNames(0); err == nil {
+		for _, sk := range sks {
+			if err := getRegSubKey(k, sk, bufs, kv); err != nil {
```
andrew-d, line 710: It's a bit of a rare case, but... Windows does support symbolic links in registry keys (`REG_OPTION_CREATE_LINK`) so we might want to have a recursion limit?

util/winutil/winutil_windows_test.go
```diff
@@ -28,3 +33,117 @@ func TestLookupPseudoUser(t *testing.T) {
...
+}
+
+func TestRegistrySupportInfo(t *testing.T) {
+	// Make sure the key doesn't exist yet
+	k, err := registry.OpenKey(registry.CURRENT_USER, keyNameTest, registry.READ)
+	switch {
+	case err == nil:
+		k.Close()
+		t.Fatalf("Test key already exists")
+	case err != registry.ErrNotExist:
```
andrew-d, line 100:
```suggestion
	case !errors.Is(err, registry.ErrNotExist):
```

util/winutil/winutil_windows_test.go
```diff
@@ -28,3 +33,117 @@ func TestLookupPseudoUser(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("Error setting %q: %v", vk, err)
+		}
+	}
+}
+
+func TestRegistrySupportInfo(t *testing.T) {
+	// Make sure the key doesn't exist yet
+	k, err := registry.OpenKey(registry.CURRENT_USER, keyNameTest, registry.READ)
+	switch {
```
andrew-d, line 96: Do you need a `switch` here or would an `if` / `else` work?

util/winutil/winutil_windows_test.go
```diff
@@ -28,3 +33,117 @@ func TestLookupPseudoUser(t *testing.T) {
...
+			t.Fatalf("Unknown type")
+		}
+
+		if err != nil {
+			t.Fatalf("Error setting %q: %v", vk, err)
+		}
+	}
+}
+
+func TestRegistrySupportInfo(t *testing.T) {
```
andrew-d, line 93: Love the detailed test btw; thank you for this 🙏

andrew-d (commented):
Generally LGTM but I want to give this another read-through later today to make sure I'm not missing anything; it's been a while since I worked with the registry 😅

util/winutil/winutil_windows_test.go
dblohm7: `switch` felt a bit cleaner to me.

util/winutil/winutil_windows.go
dblohm7: Ditto response :-)

util/winutil/winutil_windows.go
dblohm7: Sure.

andrew-d (approved):

PR #8148, opened 2023-05-16
cmd/tailscale/cli: [serve] add reset flag

### LINKED ISSUE #8139: FR: Add `--reset` to tailscale serve / funnel
We should have an easier way to clear the serve config other than `$ echo {} | sudo tailscale serve set-raw`

soniaappasamy:
Usage: `tailscale serve reset`

Fixes #8139


bradfitz:
Not at a computer for 30 but IIRC these commands are easy to write tests for. There's an existing table you can just add to?

shayne (approved):
Thanks for doing this!

PR #8157, opened 2023-05-17
ssh/tailssh: close tty on session close

### LINKED ISSUE #8119: Each ssh connection creates a new PTS

maisem:
We were only closing on side of the pty/tty pair.
Close the other side too.

Fixes #8119

cc @fritterhoff

### DIFF
--- a/ssh/tailssh/incubator.go
+++ b/ssh/tailssh/incubator.go
@@ -452,7 +452,7 @@ func (ss *sshSession) launchProcess() error {
 		return ss.startWithStdPipes()
 	}
 	ss.ptyReq = &ptyReq
-	pty, err := ss.startWithPTY()
+	pty, tty, err := ss.startWithPTY()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


bradfitz (approved):

maisem:
@fritterhoff if you can confirm this works, I'll can merge it.

maisem:
this looks like an incomplete fix, looking deeper.

fritterhoff:
Yeah. Thats why I added some (hacky) error handling and the mutex in my PR.

maisem:
>Yeah. Thats why I added some (hacky) error handling and the mutex in my PR.

The mutex doesn't help much sadly, as it is still a race between which copy routine acquires it first. I think what I have no works.

fritterhoff:
Since the std out was never closed before closing the tty it looked fine for me, but prop. your solution is better 😉

bradfitz (approved):

PR #8163, opened 2023-05-18
derp: remove default logging of disconnecting clients

### LINKED ISSUE #8024: derp logs filled up with normal client disconnect messages
### What is the issue?

Whenever a client disconnects from derp, derp logs the following messages:

2023/05/02 12:52:08 derp client 127.0.0.1:50529[UY2Jk]: read EOF
2023/05/02 12:52:08 derp client 127.0.0.1:50529[UY2Jk]: removed connection

This fills up the logs with what seems like a pretty normal non-error thing, making it hard to find actual errors. I'd rather see this only get logged if the error is something other than EOF.

### Steps to reproduce

_No response_

_No response_

### OS

_No response_

### OS version

_No response_

### Tailscale version

_No response_

### Other software

_No response_

### Bug report

_No response_

valscale:
~97% of the log messages derper outputs are related to the normal non-error state of a client disconnecting in some manner. Add a verbose logging feature that only logs these messages when enabled.

Fixes #8024

### DIFF
--- a/derp/derp_server.go
+++ b/derp/derp_server.go
@@ -498,7 +498,7 @@ func (s *Server) registerClient(c *sclient) {
 	switch set := set.(type) {
 	case nil:
 		s.clients[c.key] = singleClient{c}
-		c.debug("register single client")
+		c.debugLogf("register single client")
 	case singleClient:
 		s.dupClientKeys.Add(1)
 		s.dupClientConns.Add(2) // both old and new count
@@ -514,7 +514,7 @@ func (s *Server) registerClient(c *sclient) {
 			},
 			sendHistory: []*sclient{old},
 		}
-		c.debug("register duplicate client")
+		c.debugLogf("register duplicate client")
 	case *dupClientSet:
 		s.dupClientConns.Add(1)     // the gauge
 		s.dupClientConnTotal.Add(1) // the counter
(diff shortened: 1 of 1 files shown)


derp/derp_server.go
```diff
@@ -1024,9 +1037,7 @@ func (s *Server) recordDrop(packetBytes []byte, srcKey, dstKey key.NodePublic, r
 		msg := fmt.Sprintf("drop (%s) %s -> %s", srcKey.ShortString(), reason, dstKey.ShortString())
 		s.limitedLogf(msg)
 	}
-	if s.debug {
-		s.logf("dropping packet reason=%s dst=%s disco=%v", reason, dstKey, disco.LooksLikeDiscoWrapper(packetBytes))
-	}
+	s.debugLogf("dropping packet reason=%s dst=%s disco=%v", reason, dstKey, disco.LooksLikeDiscoWrapper(packetBytes))
```
knyar, line 1040: This will call `LooksLikeDiscoWrapper` on every dropped packet even when debug logging is not enabled - perhaps it's worth keeping the result of a [previous call](https://github.com/tailscale/tailscale/blob/f0ee03dfafda9c5a8804b25f20ea881090bd4ab5/derp/derp_server.go#L1014) in a variable and use it here?

derp/derp_server.go
```diff
@@ -744,16 +753,20 @@ func (c *sclient) run(ctx context.Context) error {
 	defer func() {
 		cancelSender()
 		if err := grp.Wait(); err != nil && !c.s.isClosed() {
-			c.logf("sender failed: %v", err)
+			if errors.Is(err, context.Canceled) {
```
knyar, line 756: Nice!

knyar (approved):
Thank you!

derp/derp_server.go
valscale: Good catch! Thank you, fixed as you describe.

fylie:
thanks for doing this, @valscale 💜

Issue #8171, opened 2023-05-19
portlist: allow to receive updates for loopback addresses


marwan-at-work:
After an internal discussion, there are use cases (such as using `tailscale serve`) where loopback addresses are of interest from within the portlist package. 

Ideally we can use a param to switch the behavior on off based on the caller.

bradfitz:
I'd like to see this fleshed out more, as I don't remember or missed the "internal discussion".

Can you elaborate here?

marwan-at-work:
Re-opening as the last PR did not account for the constructor

PR #6202, opened 2022-11-04
cmd/tswrap: command to run a child process and make it accessible over Tailscale.

danderson:
Still rough, but basic functionality works.

Main wart right now is shutdown doesn't do a clean logout for ephemeral mode, so you end up with a proliferation of machines when things restart. Also missing TLS/HTTP/etc. support, everything but the bare minimum.

PR #8172, opened 2023-05-19
portlist: add Poller.IncludeLocalhost option

marwan-at-work:
This PR parameterizes receiving loopback updates from the portlist package.
Note that the functional-options pattern was used to avoid breaking changes.

Fixes https://github.com/tailscale/tailscale/issues/8171

Note: tested on mac/linux but have not had the chance to work on the windows part yet. Happy to do so in this PR or as a follow up.

### DIFF
--- a/portlist/netstat_test.go
+++ b/portlist/netstat_test.go
@@ -8,6 +8,7 @@ package portlist
 import (
 	"bufio"
 	"encoding/json"
+	"fmt"
 	"strings"
 	"testing"
 
@@ -52,30 +53,40 @@ udp46      0      0  *.146                 *.*
 `
 
 func TestParsePortsNetstat(t *testing.T) {
-	want := List{
-		Port{"tcp", 23, ""},
-		Port{"tcp", 24, ""},
-		Port{"udp", 104, ""},
-		Port{"udp", 106, ""},
-		Port{"udp", 146, ""},
-		Port{"tcp", 8185, ""}, // but not 8186, 8187, 8188 on localhost
-	}
-
-	pl, err := appendParsePortsNetstat(nil, bufio.NewReader(strings.NewReader(netstatOutput)))
-	if err != nil {
-		t.Fatal(err)
-	}
-	pl = sortAndDedup(pl)
-	jgot, _ := json.MarshalIndent(pl, "", "\t")
-	jwant, _ := json.MarshalIndent(want, "", "\t")
-	if len(pl) != len(want) {
-		t.Fatalf("Got:\n%s\n\nWant:\n%s\n", jgot, jwant)
-	}
-	for i := range pl {
-		if pl[i] != want[i] {
-			t.Errorf("row#%d\n got: %+v\n\nwant: %+v\n",
-				i, pl[i], want[i])
-			t.Fatalf("Got:\n%s\n\nWant:\n%s\n", jgot, jwant)
-		}
+	for _, loopBack := range [...]bool{false, true} {
+		t.Run(fmt.Sprintf("loopback_%v", loopBack), func(t *testing.T) {
+			want := List{
+				{"tcp", 23, "", 0},
+				{"tcp", 24, "", 0},
+				{"udp", 104, "", 0},
+				{"udp", 106, "", 0},
+				{"udp", 146, "", 0},
+				{"tcp", 8185, "", 0}, // but not 8186, 8187, 8188 on localhost, when loopback is false
+			}
+			if loopBack {
+				want = append(want,
+					Port{"tcp", 8186, "", 0},
+					Port{"tcp", 8187, "", 0},
+					Port{"tcp", 8188, "", 0},
+				)
+			}
+			pl, err := appendParsePortsNetstat(nil, bufio.NewReader(strings.NewReader(netstatOutput)), loopBack)
+			if err != nil {
+				t.Fatal(err)
+			}
+			pl = sortAndDedup(pl)
+			jgot, _ := json.MarshalIndent(pl, "", "\t")
+			jwant, _ := json.MarshalIndent(want, "", "\t")
+			if len(pl) != len(want) {
+				t.Fatalf("Got:\n%s\n\nWant:\n%s\n", jgot, jwant)
+			}
+			for i := range pl {
+				if pl[i] != want[i] {
+					t.Errorf("row#%d\n got: %+v\n\nwant: %+v\n",
+						i, pl[i], want[i])
+					t.Fatalf("Got:\n%s\n\nWant:\n%s\n", jgot, jwant)
+				}
+			}
+		})
 	}
 }
--- a/portlist/poller.go
+++ b/portlist/poller.go
@@ -24,6 +24,11 @@ var debugDisablePortlist = envknob.RegisterBool("TS_DEBUG_DISABLE_PORTLIST")
 // Poller scans the systems for listening ports periodically and sends
 // the results to C.
 type Poller struct {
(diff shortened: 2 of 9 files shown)


portlist/portlist_macos.go
```diff
@@ -184,6 +188,10 @@ func (im *macOSImpl) addProcesses() error {
 			// starting a new process
 			cmd = ""
 			proto = ""
+			pid = 0
+			if p, err := strconv.Atoi(string(val)); err == nil {
```
bradfitz, line 192: this is an allocation which the rest of this code is trying to avoid. You could use https://pkg.go.dev/go4.org/mem#ParseInt

portlist/portlist_windows.go
```diff
@@ -33,7 +33,7 @@ type portMeta struct {
 	keep bool
 }
 
-func newWindowsImpl() osImpl {
+func newWindowsImpl(bool) osImpl {
```
bradfitz, line 36: name this bool so it's less mysterious

portlist/poller.go
```diff
@@ -62,20 +64,34 @@ type osImpl interface {
 }
 
 // newOSImpl, if non-nil, constructs a new osImpl.
-var newOSImpl func() osImpl
+var newOSImpl func(withLoopbackUpdates bool) osImpl
 
 var errUnimplemented = errors.New("portlist poller not implemented on " + runtime.GOOS)
 
+type Option func(p *Poller)
```
bradfitz, line 71: docs. and you can unname p

portlist/poller.go
```diff
@@ -62,20 +64,34 @@ type osImpl interface {
...
+
+// WithLoopBackUpdates tells the Poller to send updates for new ports
+// that are opened on a loopback address. The default is that loopback
+// addresses are skipped.
+func WithLoopbackUpdates(b bool) Option {
+	return func(p *Poller) {
+		p.withLoopbackUpdates = b
+	}
+}
```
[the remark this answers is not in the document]
marwan-at-work: @bradfitz yeah it's def not my favorite either, but otherwise changing `NewPoller` would be a breaking change. Is it okay to make a breaking change here?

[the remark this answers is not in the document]
marwan-at-work: Some options can be: 

1.
```golang
type PollerOptions struct { WithLoopback bool }
func NewPoller(opts PollerOptions) ...
```

2.
```golang
func NewPoller(withLoopback bool) ...
```

```golang
func NewPoller() ... // deprecated.
func NewPollerWithOptions(...) ... // non breaking change, but now we have two constructors.
```

portlist/netstat_test.go
```diff
@@ -52,30 +53,40 @@ udp46      0      0  *.146                 *.*
...
-		t.Fatalf("Got:\n%s\n\nWant:\n%s\n", jgot, jwant)
-	}
-	for i := range pl {
-		if pl[i] != want[i] {
-			t.Errorf("row#%d\n got: %+v\n\nwant: %+v\n",
-				i, pl[i], want[i])
-			t.Fatalf("Got:\n%s\n\nWant:\n%s\n", jgot, jwant)
-		}
+	for _, loopBack := range [...]bool{false, true} {
+		t.Run(fmt.Sprintf("loopback_%t", loopBack), func(t *testing.T) {
```
bradfitz, line 57: just use %v; it's super rare to see anybody use %t :)

portlist/poller.go
```diff
@@ -45,6 +45,10 @@ type Poller struct {
 	scratch []Port
 
 	prev List // most recent data, not aliasing scratch
+
+	// Include updates for loopback addresses.
+	// This field should only be changed before calling Run.
+	IncludeLocalhost bool
```
bradfitz, line 51: put the knobs at the top of the struct.

also, use conventional go doc style: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

// IncludeLocalhost controls whether services bound to localhost are included.
//
// This field should only be set before calling Run.

portlist/portlist_windows.go
```diff
@@ -33,7 +33,7 @@ type portMeta struct {
 	keep bool
 }
 
-func newWindowsImpl() osImpl {
+func newWindowsImpl(includeLocalhost bool) osImpl {
 	return &windowsImpl{
 		known: map[famPort]*portMeta{},
 	}
```
bradfitz, line 39: not doing windows? and openbsd/freebsd/etc?

kinda a bummer that any non-portable code needs to know about this.

seems like it'd be nice if the OS-specific files just included all ports (even localhost) internally and then the portable layer did the localhost filtering before the process name lookup

marwan-at-work: @bradfitz I just missed it between rebases, pushing an update in the next commit.

marwan-at-work:
@bradfitz thanks for the second round of review 👍🏼 pushed the updates you suggested.

portlist/poller.go
```diff
@@ -24,6 +24,11 @@ var debugDisablePortlist = envknob.RegisterBool("TS_DEBUG_DISABLE_PORTLIST")
 // Poller scans the systems for listening ports periodically and sends
 // the results to C.
 type Poller struct {
+	// IncludeLocalhost controls whether services bound to localhost are included
```
bradfitz, line 27: Missing period

portlist/portlist.go
```diff
@@ -18,6 +18,7 @@ type Port struct {
 	Proto   string // "tcp" or "udp"
 	Port    uint16 // port number
 	Process string // optional process name, if found
+	Pid     int    // process id, if known
```
bradfitz, line 21: Thought you were planning on adding a bool for localhost on ports? Guess not needed?

marwan-at-work: @bradfitz yeah it's not needed for now since we changed how we do the filtering, but I'm happy to add it back in a follow up PR.

portlist/portlist_macos.go
```diff
@@ -202,6 +209,7 @@ func (im *macOSImpl) addProcesses() error {
 			switch {
 			case m != nil:
 				m.port.Process = cmd
+				m.port.Pid = pid
```
bradfitz, line 212: why is this only populated on mac?

marwan-at-work: @bradfitz great catch actually, I initially rebased the pid part from https://github.com/tailscale/tailscale/pull/6202, but now I realize there are a few missing areas and not just linux/mac_os. For example [this line](https://github.com/tailscale/tailscale/blob/4e86857313f3acb370393aacbbbd4d47cba105c5/portlist/portlist.go#L72) should also make sure to copy over the pid

marwan-at-work:
@bradfitz 

> No pid on Windows?


I believe it's there: https://github.com/tailscale/tailscale/pull/8172/files#diff-abc054a2c3ff8a317910bf1f6f27bf7cb380a6a7b0a5e7c584540b1b62ac932bR88

portlist/portlist_macos.go
```diff
@@ -184,6 +187,10 @@ func (im *macOSImpl) addProcesses() error {
 			// starting a new process
 			cmd = ""
 			proto = ""
+			pid = 0
+			if p, err := mem.ParseInt(mem.B(val), 10, 0); err == nil {
+				pid = int(p)
+			}
```
bradfitz, line 193: or just `pid, _ = mem.ParseInt(mem.B(val), 10, 0)`

bradfitz (approved):
But you might want to amend the commit to change from your personal to work email in the git author + DCO

PR #8182, opened 2023-05-21
ssh/tailssh: fix regression after LDAP support

### LINKED ISSUE #8180: tailssh 1.40.1 needs getent command, not available on openwrt
### What is the issue?

hello again! since 1.40.1, `tailscale ssh` tries & fails to use `getent` command, which stops progressing ssh access. on debian it's part of libc packaging. with busybox/uclibc on openwrt, i couldn't find a equivalent command.

downgrading to 1.40.0 lets ssh work.

1.40.1

BUG-146bdbf1075209acb686a6cb6b377fbd8d807118b3b6c75946863236f51358e2-20230521100239Z-91a99b038f7edeba
BUG-146bdbf1075209acb686a6cb6b377fbd8d807118b3b6c75946863236f51358e2-20230521100247Z-1c02730609a67d1d

1.41.69

BUG-146bdbf1075209acb686a6cb6b377fbd8d807118b3b6c75946863236f51358e2-20230521100611Z-a8dfbabbd91ae70d
BUG-146bdbf1075209acb686a6cb6b377fbd8d807118b3b6c75946863236f51358e2-20230521100615Z-c6b9cd3ee9a55cd3

keep on doing great!

### Steps to reproduce

server 1.40.1 up to 1.41.69:

```console
ssh-conn-20230521T100546-1ee5495256: handling conn: 100.93.a.b:52218->root@100.117.c.d:22
error calling getent for user "root": exec: "getent": executable file not found in $PATH
ssh-conn-20230521T100546-1ee5495256: failed to look up root: user: invalid userid root
```

client sees:

```console
ssh root@node
failed to look up root
root@node: Permission denied (tailscale).
```

BUG for 1.40.1

LDAP support reorganized these part a bit

### OS

Other

### OS version

openwrt 22.03.5

### Tailscale version

1.40.1

### Other software

_No response_

### Bug report

BUG-146bdbf1075209acb686a6cb6b377fbd8d807118b3b6c75946863236f51358e2-20230521100611Z-a8dfbabbd91ae70d

bradfitz:
58ab66ec51f1963fbee302c75ad0017d81d37884 added LDAP support for #4945 by shelling out to getdent.

It was supposed to fall back to the old method when getdent wasn't found, but some variable name confusion (uid vs username) meant the old path wasn't calling the right lookup function (user.LookupId instead of user.Lookup).

Which meant that changed probably also broke FreeBSD and macOS SSH support in addition to the reported OpenWRT regression.

The gokrazy support didn't look right either.

Fixes #8180

### DIFF
--- a/ssh/tailssh/user.go
+++ b/ssh/tailssh/user.go
@@ -48,19 +48,26 @@ func (u *userMeta) GroupIds() ([]string, error) {
 	return u.User.GroupIds()
 }
 
-// userLookup is like os/user.LookupId but it returns a *userMeta wrapper
+// userLookup is like os/user.Lookup but it returns a *userMeta wrapper
 // around a *user.User with extra fields.
-func userLookup(uid string) (*userMeta, error) {
+func userLookup(username string) (*userMeta, error) {
 	if runtime.GOOS != "linux" {
-		return userLookupStd(uid)
+		return userLookupStd(username)
 	}
 
 	// No getent on Gokrazy. So hard-code the login shell.
 	if distro.Get() == distro.Gokrazy {
-		um, err := userLookupStd(uid)
-		if err == nil {
-			um.loginShellCached = "/tmp/serial-busybox/ash"
+		um, err := userLookupStd(username)
+		if err != nil {
+			um.User = user.User{
+				Uid:      "0",
+				Gid:      "0",
+				Username: "root",
+				Name:     "Gokrazy",
+				HomeDir:  "/",
+			}
 		}
+		um.loginShellCached = "/tmp/serial-busybox/ash"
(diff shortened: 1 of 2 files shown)


maisem (approved):

DentonGentry (approved):

ssh/tailssh/user.go
```diff
@@ -50,17 +50,24 @@ func (u *userMeta) GroupIds() ([]string, error) {
 
 // userLookup is like os/user.LookupId but it returns a *userMeta wrapper
```
bradfitz, line 51: missed another one: this should be Lookup in the docs here

PR #8187, opened 2023-05-22
Create devcontainer.json

mmfiv:


DentonGentry (changes requested):
Did you intend to send this PR to your own fork upstream? It is not at all clear to me that we want it in the main repository.

PR #8200, opened 2023-05-24
tka: provide authority StateID in NetworkLockStatus response

agottardo:


twitchyliquid64 (approved):
Issue ID + signed-off-by using a DCO-signed or Tailscale email and you should be g2g :)

PR #8203, opened 2023-05-24
portlist: remove NewPoller constructor

### LINKED ISSUE #8171: portlist: allow to receive updates for loopback addresses

marwan-at-work:
This is a follow up on PR #8172 and a breaking change that removes NewPoller. The issue with the previous PR was that NewPoller immediately initializes the underlying os implementation and therefore setting IncludeLocalhost as an exported field happened too late and cannot happen early enough. Using the zero value of Poller was also not an option from outside of the package because we need to set initial private fields

Fixes #8171

### DIFF
--- a/portlist/poller.go
+++ b/portlist/poller.go
@@ -31,14 +31,15 @@ type Poller struct {
 
 	c chan List // unbuffered
 
+	initOnce sync.Once // guards init of private fields
+
 	// os, if non-nil, is an OS-specific implementation of the portlist getting
 	// code. When non-nil, it's responsible for getting the complete list of
 	// cached ports complete with the process name. That is, when set,
 	// addProcesses is not used.
 	// A nil values means we don't have code for getting the list on the current
 	// operating system.
-	os     osImpl
-	osOnce sync.Once // guards init of os
+	os osImpl
 
 	// closeCtx is the context that's canceled on Close.
 	closeCtx       context.Context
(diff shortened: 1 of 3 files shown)


portlist/poller.go
```diff
@@ -31,14 +31,15 @@ type Poller struct {
 
 	c chan List // unbuffered
 
+	initOnce sync.Once // guards init of private fields
```
bradfitz, line 34: move this up a couple lines before `c`.  (normally mutexes and sync.once etc come before the stuff they guard)

portlist/poller.go
```diff
@@ -71,54 +72,44 @@ var newOSImpl func(includeLocalhost bool) osImpl
...
+// all private fields. All exported methods must call this in a
+// Once, otherwise they may panic.
+func (p *Poller) init() {
+	if debugDisablePortlist() {
+		return
+	}
 	if newOSImpl != nil {
 		p.os = newOSImpl(p.IncludeLocalhost)
 	}
+	p.closeCtx, p.closeCtxCancel = context.WithCancel(context.Background())
```
bradfitz, line 91: minor, but you don't want to call WithCancel if it'll never be called. Imagine you're LocalBackend and Check returns an error and you then give up on it.

So Check should probably call Close on error

marwan-at-work: @bradfitz done

marwan-at-work:
Closing in favor of https://github.com/tailscale/tailscale/pull/8275

PR #8210, opened 2023-05-25
Add a knob for maxInFlightConnectionAttempts on TCP connections

rowillia:
We were having major stability problems with the Tailscale Subnet router - we'd see the router cycle between stability and dropping all new connections.  Upon further investigation (thanks @nelhage - https://github.com/nelhage/tailscale/commit/94633bfe0e82fe5bcadf788908f1a7b1d4f3758c ! ) we discovered Tailscale sets maxInFlightConnectionAttempts to 16.  Bumping this to 1024 dramatically improved reliability!

### DIFF
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -246,6 +246,15 @@ func (ns *Impl) wrapProtoHandler(h func(stack.TransportEndpointID, stack.PacketB
 	}
 }
 
+var tcpMaxInFlightConnectionAttemptsVar = envknob.RegisterInt("TS_TCP_MAX_IN_FLIGHT_CONNETION_ATTEMPTS")
+func () tcpMaxInFlightConnectionAttempts int {
+	if x := tcpMaxInFlightConnectionAttemptsVar(); x != 0 {
+		return x
+	}
+	return 16
+}
+
+
 // Start sets up all the handlers so netstack can start working. Implements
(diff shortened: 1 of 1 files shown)


bradfitz:
In general we eschew "make-it-work" knobs. Especially ones with typos in them :)

I'd rather we just do The Right Thing automatically, once we figure out what that is.

wgengine/netstack/netstack.go
```diff
@@ -246,6 +246,15 @@ func (ns *Impl) wrapProtoHandler(h func(stack.TransportEndpointID, stack.PacketB
 	}
 }
 
+var tcpMaxInFlightConnectionAttemptsVar = envknob.RegisterInt("TS_TCP_MAX_IN_FLIGHT_CONNETION_ATTEMPTS")
+func () tcpMaxInFlightConnectionAttempts int {
```
bradfitz, line 250: this also doesn't look like it compiles

rowillia:
@bradfitz makes sense!  We met w/ @JayWStapleton to debug this - it's been unfortunately plaguing us for some time.  Happy to abandon this change but we'd love to be able to get a change to this setting in ASAP so we don't have to run a fork for long.

What would your preferred path forward be?  Just change the value to 1024?  I'm unsure as to what the downsides would be to bumping this too high.

bradfitz:
@andrew-d is currently researching this, I believe.

Issue #8217, opened 2023-05-26
Make cmd/dist build synology packages


danderson:
Now that we have cmd/dist, it can build synology packages as well as other things, rather than having a separate build process for those.

DentonGentry:
Note one subtlety: Synology requires that version numbers be in a specific format.
https://github.com/tailscale/tailscale-synology/commit/250e8cfd471d1111003ef3237897e12007f257c7

Builds with version numbers which do not start with "60" or "70" will be rejected by the Package Center.

--------

Implementing Synology support in cmd/dist may also fix https://github.com/tailscale/corp/issues/7959 (which most of y'all won't be able to see, sorry).

danderson:
I missed @DentonGentry's comment originally, but for the record: I ported synology builds to cmd/dist, and verified that the new build produces bit-identical outputs to the old, except for some minor diffs in what version information gets embedded into binaries, and slightly different file sizes due to different compression settings. But all the important package metadata, binaries and version numbers behave the same as they did before, including the `60` and `70` prefixes.

danderson:
Also for the record: I verified that www.tailscale.com doesn't need any updates. The one pointer to the tailscale-synology repo is appropriate (crediting the original creator of the package, prior to us taking on its maintenance), everything else points to either the official Synology software center, or pkgs.tailscale.com for sideloading.

PR #8218, opened 2023-05-26
release/dist/synology: build synology packages with cmd/dist

danderson:
Updates #8217

---

Because of cmd/dist's memoizing and build cache sharing, the build is faster. Before:

```
$ time make spkall
...
make spkall  34.45s user 7.29s system 229% cpu 18.183 total
```

After:

```
$ go run ./cmd/dist build synology
...
go run ./cmd/dist build synology  27.50s user 5.43s system 957% cpu 3.437 total
```

I compared the outputs of the old and new implementation. They are bit for bit identical, except that cmd/dist built binaries are a bit larger due to extra version info (a dozen bytes or so), and cmd/dist compresses the packages slightly better (~1k smaller to distribute).


release/dist/synology/files/config
```diff
@@ -0,0 +1,12 @@
+{
+	".url": {
+		"SYNO.SDS.Tailscale": {
+			"type": "url",
+			"version": "1.8.3",
```
sailorfrag, line 5: Does this need to be generated with the current version number?

release/dist/synology/files/Tailscale.sc
sailorfrag, line 1: Is missing trailing EOL required for this file?

There are several other files showing this too. I recognize that these are identical to what was in the `tailscale/tailscale-synology` repo and thus it's not a new concern.

cmd/dist/dist.go
```diff
@@ -13,15 +13,28 @@ import (
...
 )
 
+var synologyPackageCenter bool
+
 func getTargets() ([]dist.Target, error) {
-	return unixpkgs.Targets(), nil
+	var ret []dist.Target
+
+	ret = append(ret, unixpkgs.Targets()...)
+	ret = append(ret, synology.Targets(synologyPackageCenter)...)
```
sailorfrag, line 26: Is this supposed to be conditional on `synologyPackageCenter`?

release/dist/synology/pkgs.go
```diff
@@ -0,0 +1,306 @@
...
+	f := func(k, v string) {
+		fmt.Fprintf(&ret, "%s=%q\n", k, v)
+	}
+	f("package", "Tailscale")
+	f("version", fmt.Sprintf("%s-%d", b.Version.Short, b.Version.Synology[t.dsmMajorVersion]))
+	f("arch", t.filenameArch)
+	f("description", "Connect all your devices using WireGuard, without the hassle.")
+	f("displayname", "Tailscale")
+	f("maintainer", "Tailscale, Inc.")
+	f("maintainer_url", "https://github.com/tailscale/tailscale-synology")
```
sailorfrag, line 105: Is this still correct?

sailorfrag (approved):
Seems mostly fine, likely some trivial changes needed

release/dist/synology/files/config
danderson: It's a TODO in the code I ported over from tailscale-synology. I decided to not get to the bottom of it in this change, so I left it as-is and ported the TODO to figure it out.

release/dist/synology/files/Tailscale.sc
danderson: I have no idea offhand. I can't imagine that synology software would care... But I'm also afraid to find out now :)

cmd/dist/dist.go
danderson: No, the bool is specifying which variant of the packages to build: standalone for sideloading, or the ones for distribution in the synology package center. The latter can request permission to use tuntap devices and such, whereas the sideloaded ones can't. Effectively, we're the only consumers of this flag, and we'll set it when generating packages for Synology to distribute, as opposed to the ones we put on pkgs.tailscale.com so people can sideload.

danderson: Added a comment to expand on this at the call site.

release/dist/synology/pkgs.go
danderson: Woops, no it's not. Fixed.

PR #8224, opened 2023-05-27
ipn/ipnlocal: [serve/funnel] add forwarded host and proto header

### LINKED ISSUE #7061: FR: serve: Add the X-Forwarded-Proto header

I've tried to use `tailscale serve` to simplify my setup since I don't have any complex reverse proxying requirements, however I found that my backend really wants to have the `X-Forwarded-Proto` header set to `https` if the proxy terminated TLS on its behalf.

### How should we solve this?

Since the Tailscale server knows whether it had terminated TLS, it seems like it should be fairly easy for it to insert the `X-Forwarded-Proto` header with the appropriate value when reverse-proxying the requests.

Some picky backends want to see the `X-Forwarded-Proto: https` header and will display warnings, etc., if it's missing.

### Anything else?

_No response_

mKeRix:
This replicates the headers also sent by the golang reverse proxy by default (see the [recent commit](https://github.com/golang/go/commit/1513e57b704056b794f0706362fa3c949f2972a4)). Some applications depend on these headers to generate redirects or links correctly.

I verified this change by building the tool locally on my Mac, running an echo server and serving that via the locally built tailscale CLI. I then used my phone in the same tailnet to access this, verifying that the returned headers included both X-Forwarded-Host and X-Forwarded-Proto.

I need this change to continue with my personal Tailscale use case, so I'd be happy to see this get merged. Happy to adjust as needed, just let me know.

Fixes #7061.

### DIFF
--- a/ipn/ipnlocal/serve.go
+++ b/ipn/ipnlocal/serve.go
@@ -447,8 +447,14 @@ func (b *LocalBackend) proxyHandlerForBackend(backend string) (*httputil.Reverse
 		Rewrite: func(r *httputil.ProxyRequest) {
 			r.SetURL(u)
 			r.Out.Host = r.In.Host
+			r.Out.Header.Set("X-Forwarded-Host", r.In.Host)
 			if c, ok := r.Out.Context().Value(serveHTTPContextKey{}).(*serveHTTPContext); ok {
 				r.Out.Header.Set("X-Forwarded-For", c.SrcAddr.Addr().String())
+			}
+			if r.In.TLS == nil {
+				r.Out.Header.Set("X-Forwarded-Proto", "http")
+			} else {
+				r.Out.Header.Set("X-Forwarded-Proto", "https")
 			}
 		},
 		Transport: &http.Transport{
--- a/ipn/ipnlocal/serve.go
+++ b/ipn/ipnlocal/serve.go
@@ -448,13 +448,9 @@ func (b *LocalBackend) proxyHandlerForBackend(backend string) (*httputil.Reverse
 			r.SetURL(u)
 			r.Out.Host = r.In.Host
 			r.Out.Header.Set("X-Forwarded-Host", r.In.Host)
+			r.Out.Header.Set("X-Forwarded-Proto", "https")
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/serve.go
```diff
@@ -447,9 +447,15 @@ func (b *LocalBackend) proxyHandlerForBackend(backend string) (*httputil.Reverse
...
 			r.SetURL(u)
 			r.Out.Host = r.In.Host
+			r.Out.Header.Set("X-Forwarded-Host", r.In.Host)
 			if c, ok := r.Out.Context().Value(serveHTTPContextKey{}).(*serveHTTPContext); ok {
 				r.Out.Header.Set("X-Forwarded-For", c.SrcAddr.Addr().String())
 			}
+			if r.In.TLS == nil {
+				r.Out.Header.Set("X-Forwarded-Proto", "http")
+			} else {
+				r.Out.Header.Set("X-Forwarded-Proto", "https")
```
shayne, line 457: I believe the `X-Forwarded-Proto` will always be `https`. We currently only support an HTTPS listener with `serve`/`funnel`.

mKeRix: Thanks for the hint! I added another commit that simplifies the code change by always adding https as the value (= dropping the if).

shayne (approved):

rocktavious:
@DentonGentry - we are trying to use tailscale funnel on mac with a rails monolith and we think we are running into the fact that our app needs `X-FORWARDED-PROTO` - how long until this makes it way into the Mac OS app?  Is there a way for us to use a built binary for MacOS in the interim?

DentonGentry:
It will appear in the unstable channel first, which for macOS is the TestFlight build: https://testflight.apple.com/join/7ExrFq00

This fix is NOT in 1.43.15, the current macOS TestFlight build at the time of this writing. This fix will appear in the next TestFlight build to appear. That will probably be tomorrow, but please monitor for it.

This fix will then appear in the 1.44.0 stable release.

Issue #8232, opened 2023-05-30
Clean up release/dist/synology/files/config


danderson:
`release/dist/synology/files/config` is the file that becomes `ui/config` in our Synology packages, and whose content describes our icon in the program list on the DSM UI, afaict.

This file is partly documented in the developer guide (https://help.synology.com/developer-guide/integrate_dsm/config.html), but we're setting some fields that don't seem to exist, and some undocumented fields also seem to exist:

 - We set `version` to 1.8.3, the version of Tailscale at which we added UI configuration to the app. AFAICT, this field doesn't exist, in either the developer docs or the empirically known undocumented fields used in the wild. I think we can just remove it?
 - We set `urlTarget`, which is undocumented but is understood (via the synology community forum) to behave similarly to `<a target=...>`: it lets you specify a symbolic name for the DSM web GUI window that the app should open in, and specifically setting it ensures that trying to open the app multiple times just refreshes the same window rather than create more windows. This seems reasonable to keep.

I think I would like to convert this config file to hujson in git, and have cmd/dist translate it back to plain JSON. In particular, I want to be able to add comments that link to the developer guide, and explain the undocumented field we're using.

Additionally, I think we should remove the `version` field unless we have evidence that it's used. Empirically we've not updated it for several years and nobody's noticed or complained, which strongly suggests the field is not read at all by DSM.

cc @DentonGentry who I think currently knows the most about the synology app, and cc @crawshaw who added the config originally and may have found something I didn't re: `version`.

DentonGentry:
> cc @DentonGentry who I think currently knows the most about the synology app

I believe you're correct about the version field not being used. It was never in our build instructions to increment it, so we never did, and nobody has noticed.

> I think I would like to convert this config file to hujson in git, and have cmd/dist translate it back to plain JSON

Sure, that would be fine.

PR #8233, opened 2023-05-30
release/dist/synology: remove 'version' field from ui/config

danderson:
As far as I can tell from the DSM documentation and known undocumented fields, there is no 'version' field in this config file that DSM cares about.

Updates #8232


DentonGentry (approved):

danderson:
This isn't urgent to submit so going to wait an amount of days in case @crawshaw has a distinct memory of this being load-bearing in a surprising way. Absent that, I'll merge the cleanup.

Issue #4908, opened 2022-06-23
ssh: handle SELinux somehow?


bradfitz:
https://twitter.com/dstaley/status/1539773980692733952 has tips about SELinux.

As somebody who knows ~nothing about SELinux, I'm not sure what it's saying. (That's some sort of exception policy you put somewhere?)

/cc @maisem @danderson @raggi

danderson:
If I'm reading the policy violation right, executing a shell seems to be a restricted action for the default "be a daemon" policy on this system - possibly as a defense against RCE trying to spawn a shell?

I don't know enough about SELinux to construct, distribute, install and apply a better policy though.

dstaley:
Yeah, that matches my (also limited!) understanding. I was able to work around it with the following policy:

```
module tailscale 1.0;

require {
        type unconfined_service_t;
        type unconfined_t;
        class process transition;
}

#============= unconfined_service_t ==============
allow unconfined_service_t unconfined_t:process transition;
```

This is likely disabling the specific feature of SELinux that's mitigating against RCE vulnerabilities, so definitely only use this if you know what you're doing!

I have come across [some documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/using_selinux/writing-a-custom-selinux-policy_using-selinux) on how to generate a SELinux policy for a given executable, so it's at least theoretically possible to distribute a `tailscale.pp` file that users can compile and install on their own.

bradfitz:
#4914 is a report that this breaks Fedora 36.

It works if you run `setenforce 0` (which obviously isn't an ideal fix).

Will-W:
The cheat here is presumably going to be to copy the SELinux policy chunks that sshd has, given that you have exactly the same security constraints.

Will-W:
I had a little dig into the Fedora OpenSSH package, out of curiosity, and it is of course complicated. It's not just a nasty set of policies, but the code itself has awareness of SELinux - presumably because it has to do some fairly dynamic things based on who you are logging into SSH as etc.

What I might look at if I have time this weekend is a slightly better workaround policy than the one above. Rather than saying "all unconfined daemons can execute arbitrary stuff" we can label tailscaled appropriately and then say "tailscaled can execute arbitrary stuff". That at least narrows the hole a bit, and can be a workaround while a "proper" implementation happens.

SebDanielsson:
[...]

PR #8251, opened 2023-06-01
ssh/tailssh: work around lack of scontext in SELinux

maisem:
Trying to SSH when SELinux is enforced results in errors like:

```
➜  ~ ssh ec2-user@<ip>
Last login: Thu Jun  1 22:51:44 from <ip2>
ec2-user: no shell: Permission denied
Connection to <ip> closed.
```

while the `/var/log/audit/audit.log` has
```
type=AVC msg=audit(1685661291.067:465): avc:  denied  { transition } for  pid=5296 comm="login" path="/usr/bin/bash" dev="nvme0n1p1" ino=2564 scontext=system_u:system_r:unconfined_service_t:s0 tcontext=unconfined_u:unconfined_r:unconfined_t:s0 tclass=process permissive=0
```

The right fix here would be to somehow install the appropriate context when tailscale is installed on host, but until we figure out a way to do that stop using the `login` cmd in these situations.

Updates #4908

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -4,7 +4,6 @@
 package ipnlocal
 
 import (
-	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/json"
@@ -18,7 +17,6 @@ import (
 	"net/netip"
 	"net/url"
 	"os"
-	"os/exec"
 	"os/user"
 	"path/filepath"
 	"runtime"
@@ -2583,7 +2581,7 @@ func (b *LocalBackend) checkSSHPrefsLocked(p *ipn.Prefs) error {
 		if distro.Get() == distro.QNAP && !envknob.UseWIPCode() {
 			return errors.New("The Tailscale SSH server does not run on QNAP.")
 		}
-		checkSELinux()
+		b.updateSELinuxHealthWarning()
 		// otherwise okay
 	case "darwin":
 		// okay only in tailscaled mode for now.
@@ -4705,12 +4703,8 @@ func (b *LocalBackend) sshServerOrInit() (_ SSHServer, err error) {
 
 var warnSSHSELinux = health.NewWarnable()
 
-func checkSELinux() {
-	if runtime.GOOS != "linux" {
-		return
-	}
-	out, _ := exec.Command("getenforce").Output()
-	if string(bytes.TrimSpace(out)) == "Enforcing" {
+func (b *LocalBackend) updateSELinuxHealthWarning() {
+	if hostinfo.IsSELinuxEnforcing() {
 		warnSSHSELinux.Set(errors.New("SELinux is enabled; Tailscale SSH may not work. See https://tailscale.com/s/ssh-selinux"))
 	} else {
 		warnSSHSELinux.Set(nil)
@@ -4722,7 +4716,7 @@ func (b *LocalBackend) handleSSHConn(c net.Conn) (err error) {
 	if err != nil {
 		return err
 	}
-	checkSELinux()
+	b.updateSELinuxHealthWarning()
 	return s.HandleSSHConn(c)
 }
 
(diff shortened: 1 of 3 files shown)


ipn/ipnlocal/local.go
```diff
@@ -4649,24 +4649,27 @@ func (b *LocalBackend) sshServerOrInit() (_ SSHServer, err error) {
 
 var warnSSHSELinux = health.NewWarnable()
 
-func checkSELinux() {
+// CheckSELinux reports whether SELinux is in "Enforcing" mode.
+// If it is, it warns the user that Tailscale SSH may not work.
+func CheckSELinux() bool {
```
bradfitz, line 4654: weird func. It both reports a value and has some side effects. Document more?

maisem: Thus a draft

ssh/tailssh/incubator.go
```diff
@@ -120,10 +121,18 @@ func (ss *sshSession) newIncubatorCommand() (cmd *exec.Cmd) {
...
 		}
-		if isShell || runtime.GOOS == "darwin" {
-			// Only the macOS version of the login command supports executing a
-			// command, all other versions only support launching a shell
-			// without taking any arguments.
+		// Only the macOS version of the login command supports executing a
+		// command, all other versions only support launching a shell
+		// without taking any arguments.
+		shouldUseLoginCmd := isShell || runtime.GOOS == "darwin"
+		if ipnlocal.CheckSELinux() {
```
bradfitz, line 128: `if shouldUseLoginCmd && ipnlocal.CheckSELinux() {` ?

"Check" doesn't say much to me, though, bool-wise. Check what? I guess "whether we're using".

Maybe "UsingSELinux" ? Or "IsSELinuxEnforcing" ?

ipn/ipnlocal/local.go
maisem: split in two

ipn/ipnlocal/local.go
```diff
@@ -4662,12 +4662,17 @@ func (b *LocalBackend) sshServerOrInit() (_ SSHServer, err error) {
...
+// IsSELinuxEnforcing reports whether SELinux is in "Enforcing" mode.
+func IsSELinuxEnforcing() bool {
 	if runtime.GOOS != "linux" {
-		return
+		return false
 	}
 	out, _ := exec.Command("getenforce").Output()
-	if string(bytes.TrimSpace(out)) == "Enforcing" {
+	return string(bytes.TrimSpace(out)) == "Enforcing"
+}
```
bradfitz, line 4672: ipnlocal is kinda a bizarre place for this. Maybe "hostinfo" ?

ssh/tailssh/incubator.go
```diff
@@ -120,10 +121,18 @@ func (ss *sshSession) newIncubatorCommand() (cmd *exec.Cmd) {
...
-			// command, all other versions only support launching a shell
-			// without taking any arguments.
+		// Only the macOS version of the login command supports executing a
+		// command, all other versions only support launching a shell
+		// without taking any arguments.
+		shouldUseLoginCmd := isShell || runtime.GOOS == "darwin"
+		if ipnlocal.IsSELinuxEnforcing() {
+			// If we're running on a SELinux-enabled system, the login
+			// command will be unable to set the correct context for the
+			// shell. Fallback to using the incubator to launch the shell.
```
bradfitz, line 131: Fall back (verb) instead of fallback (noun)

ssh/tailssh/incubator.go
```diff
@@ -120,10 +121,18 @@ func (ss *sshSession) newIncubatorCommand() (cmd *exec.Cmd) {
...
+			// See #4908.
```
bradfitz, line 132: clickable URL

bradfitz (approved):

mochaaP:
There should be an option for this, otherwise on correctly configured systems this will bypass the audit rules.

PR #8263, opened 2023-06-03
net/dnscache: use PreferGo on Windows.

DentonGentry:
Updates https://github.com/tailscale/tailscale/issues/5161


bradfitz (approved):
LGTM but drop the ".go" in the commit message title. The part before the colon is conventionally a package name (~directory). Sometimes we write "go.mod" as a filename for the top level, but `net/dnscache.go` is also not a filename that exists.

PR #8280, opened 2023-06-06
portlist: remove async functionality

### LINKED ISSUE #8171: portlist: allow to receive updates for loopback addresses

marwan-at-work:
This PR removes all async functionality from the portlist package which may be a breaking change for non-tailscale importers. The only importer within this codebase (LocalBackend) is already using the synchronous API so no further action needed.

Fixes #8171

### DIFF
--- a/portlist/portlist_test.go
+++ b/portlist/portlist_test.go
@@ -4,11 +4,8 @@
 package portlist
 
 import (
-	"context"
 	"net"
-	"sync"
 	"testing"
-	"time"
 
 	"tailscale.com/tstest"
 )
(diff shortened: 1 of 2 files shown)


portlist/portlist_test.go
```diff
@@ -271,6 +200,11 @@ func BenchmarkGetListIncremental(b *testing.B) {
 func benchmarkGetList(b *testing.B, incremental bool) {
 	b.ReportAllocs()
 	var p Poller
+	p.init()
+	if p.initErr != nil {
+		b.Fatal(p.initErr)
```
bradfitz, line 205: skip, no?

otherwise this won't pass `-benchtime=1x` on CI on all platforms.

marwan-at-work: @bradfitz would love to learn but at this point I'm not familiar with how this repo's CI runs its benchmarks and how this change could affect it, I updated the PR nonetheless 👍🏼

bradfitz (approved):

PR #8285, opened 2023-06-06
cmd/tailscale/cli, util/winutil/authenticode: flesh out authenticode …

### LINKED ISSUE #8284: Authenticode improvements
* MSI support;
* Check that the cert subject is who we expect to be;
* Add code to extract the cert subject (without verification) for diagnostic logging purposes.

dblohm7:
…support

Previously, tailscale upgrade was doing the bare minimum for checking authenticode signatures via `WinVerifyTrustEx`. This is fine, but we can do better:

* WinVerifyTrustEx verifies that the binary's signature is valid, but it doesn't determine *whose* signature is valid; tailscale upgrade should also ensure that the binary is actually signed *by us*.
* I added the ability to check the signatures of MSI files.
* In future PRs I will be adding diagnostic logging that lists details about every module (ie, DLL) loaded into our process. As part of that metadata, I want to be able to extract information about who signed the binaries.

This code is modelled on some C++ I wrote for Firefox back in the day. See https://searchfox.org/mozilla-central/rev/27e4816536c891d85d63695025f2549fd7976392/toolkit/xre/dllservices/mozglue/Authenticode.cpp for reference.

Fixes #8284


util/winutil/authenticode/authenticode_windows.go
```diff
@@ -0,0 +1,425 @@
...
+		memberTag:       memberTag,
+		memberFilePath:  utf16Path,
+		memberFile:      memberFile,
+		catAdmin:        catAdmin,
+	})
+	err = verifyTrust(windows.WTD_CHOICE_CATALOG, wintrustArg)
+	// Note that we intentionally return certSubj even when err != nil; the idea
+	// is that we might still want to know who the cert subject claims to be,
+	// even if the validation has failed (eg for troubleshooting purposes).
+	return certSubj, err
```
andrew-d, line 407: No strong opinions, but if we wanted to reduce the likelihood that someone accidentally uses it, we could also store the `certSubj` in a field on the error return, similar to e.g. [`net.AddrError.Addr`](https://pkg.go.dev/net#AddrError).

util/winutil/authenticode/mksyscall.go
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package authenticode
+
+//go:generate go run golang.org/x/sys/windows/mkwinsyscall -output zsyscall_windows.go mksyscall.go
+//go:generate go run golang.org/x/tools/cmd/goimports -w zsyscall_windows.go
+
+// The following two declarations use the '?' suffix because Windows 7.
```
andrew-d, line 9: Because Windows 7 what?

util/winutil/authenticode/authenticode_windows.go
```diff
@@ -0,0 +1,425 @@
...
+		return "", err
+	}
+
+	if !verify {
+		return certSubj, nil
+	}
+
+	wintrustArg := unsafe.Pointer(&windows.WinTrustFileInfo{
+		Size:     uint32(unsafe.Sizeof(windows.WinTrustFileInfo{})),
+		FilePath: utf16Path,
```
andrew-d, line 258: Technically this is vulnerable to a time-of-check time-of-use attack, where an attacker could swap out the file at this path between `newObjectQuery`, above, and replace it with a non-Tailscale-but-verified binary before `WinVerifyTrust` is called. I think this is pretty low-risk since it still needs a valid binary, but if we really wanted to defend against that we could open/read/query it ourselves with `CERT_QUERY_OBJECT_BLOB` and use the `hFile` field of `WINTRUST_FILE_INFO`.

util/winutil/authenticode/authenticode_windows.go
```diff
@@ -0,0 +1,425 @@
...
+	err = verifyTrust(windows.WTD_CHOICE_FILE, wintrustArg)
+	// Note that we intentionally return certSubj even when err != nil; the idea
+	// is that we might still want to know who the cert subject claims to be,
+	// even if the validation has failed (eg for troubleshooting purposes).
+	return certSubj, err
+}
+
+var (
+	_BCRYPT_SHA256_ALGORITHM   = &([]uint16{'S', 'H', 'A', '2', '5', '6', 0})[0]
+	_OID_CERT_STRONG_SIGN_OS_1 = &([]byte("1.3.6.1.4.1.311.72.1.1\x00"))[0]
```
andrew-d, line 269: For future reference, this is defined [here](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_strong_sign_para#:~:text=szOID_CERT_STRONG_SIGN_OS_1) and means:

> The SHA2 hash algorithm is supported. MD2, MD4, MD5, and SSHA1 are not supported.
> The signing and public key algorithms can be RSA or ECDSA. The DSA algorithm is not supported. The key size for the RSA algorithm must equal or be greater than 2047 bits. The key size for the ECDSA algorithm must equal or be greater than 256 bits.
> 
> Strong signing of CRLs and OCSP responses are enabled.

util/winutil/authenticode/authenticode_windows.go
```diff
@@ -0,0 +1,425 @@
...
+
+	wintrustArg := unsafe.Pointer(&windows.WinTrustFileInfo{
+		Size:     uint32(unsafe.Sizeof(windows.WinTrustFileInfo{})),
+		FilePath: utf16Path,
+	})
+	err = verifyTrust(windows.WTD_CHOICE_FILE, wintrustArg)
+	// Note that we intentionally return certSubj even when err != nil; the idea
+	// is that we might still want to know who the cert subject claims to be,
+	// even if the validation has failed (eg for troubleshooting purposes).
+	return certSubj, err
```
andrew-d, line 264: Ditto previous comment about field on an error type.

andrew-d (approved):
Broadly LGTM, with a few specific comments around error handling 👍

util/winutil/authenticode/authenticode_windows.go
```diff
@@ -0,0 +1,425 @@
...
+	_CMSG_SIGNER_CERT_INFO_PARAM      = 7
+	_MSI_INVALID_HASH_IS_FATAL        = 1
+	_TRUST_E_NOSIGNATURE              = wingoes.HRESULT(-((0x800B0100 ^ 0xFFFFFFFF) + 1))
+)
+
+// Verify performs authenticode verification on the file at path, and also
+// ensures that expectedCertSubject was the entity who signed it. path may point
+// to either a PE binary or an MSI package. ErrSigNotFound is returned if no
+// signature is found.
+func Verify(path string, expectedCertSubject string) error {
```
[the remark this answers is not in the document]
dblohm7: This is a valid point. I definitely want to be able to extract cert subject for diagnostic reasons, however for verification it would be nice to use something better. OTOH, this code will be used by an older version desiring to upgrade to a new version, so it won't have awareness of new serials.

Acknowledging that my understanding of concrete implementations of PKI is somewhat limited, any suggestions/advice offered here would be useful.

util/winutil/authenticode/authenticode_windows.go
dblohm7: I like it! Will do.

util/winutil/authenticode/mksyscall.go
dblohm7: LOL! Actually I can remove that because we've now dropped Win7 support.

util/winutil/authenticode/authenticode_windows.go
dblohm7: Good point. I'll see what I can do here.

PR #8289, opened 2023-06-06
tsnet: add examples

### LINKED ISSUE #8288: Copy tsnet examples from the KB into the tsnet package docs
Ref #8287

Improving the package docs to make it easier to discover tsnet functionality.

Xe:
Closes #8288

Follows the examples from the KB[1].

[1]: https://tailscale.com/kb/1244/tsnet/

### DIFF
--- a/tsnet/example_tsnet_test.go
+++ b/tsnet/example_tsnet_test.go
@@ -1,0 +1,212 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package tsnet_test
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"tailscale.com/tsnet"
+)
+
+// ExampleServer shows you how to construct a ready-to-use tsnet instance.
+func ExampleServer() {
+	srv := new(tsnet.Server)
+	if err := srv.Start(); err != nil {
+		log.Fatalf("can't start tsnet server: %v", err)
+	}
+	defer srv.Close()
+}
+
+// ExampleServer_hostname shows you how to set a tsnet server's hostname.
+//
+// This setting lets you control the host name of your program on your
+// tailnet. By default this will be the name of your program (such as foo
+// for a program stored at /usr/local/bin/foo). You can also override this
+// by setting the Hostname field.
+func ExampleServer_hostname() {
+	srv := &tsnet.Server{
+		Hostname: "kirito",
+	}
+
+	// do something with srv
+	_ = srv
+}
+
+// ExampleServer_dir shows you how to configure the persistent directory for
+// a tsnet application. This is where the Tailscale node information is stored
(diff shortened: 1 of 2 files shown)


tsnet/example_tsnet_test.go
```diff
@@ -0,0 +1,201 @@
...
+		log.Fatalf("can't start tsnet server: %v", err)
+	}
+	defer srv.Close()
+}
+
+// ExampleServer_hostname shows you how to set a tsnet server's hostname.
+//
+// This setting lets you control the host name of your program on your
+// tailnet. By default, this will be the name of your program, such as foo
+// for a program stored at /usr/local/bin/foo). You can also override this
```
WalterHub, line 30: I see the closing parenthesis after "foo", but not an opening parenthesis.

tsnet/example_tsnet_test.go
```diff
@@ -0,0 +1,201 @@
...
+	resp, err := cli.Get("https://hello.ts.net")
+	if resp == nil {
+		log.Fatal(err)
+	}
+	// do something with resp
+	_ = resp
+}
+
+// ExampleServer_Start demonstrates the Start method, which should be called if
+// you need to explicitly start it. Please note that the Start method is implicitly
```
WalterHub, line 127:
```suggestion
// you need to explicitly start it. Note that the Start method is implicitly
```

tsnet/example_tsnet_test.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Fatal(http.Serve(ln, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, "Hi there! Welcome to the tailnet!")
+	})))
+}
+
+// ExampleServer_ListenFunnel shows you how to create a HTTPS service on both your tailnet
```
WalterHub, line 173:
```suggestion
// ExampleServer_ListenFunnel shows you how to create an HTTPS service on both your tailnet
```

creachadair (approved):
Probably ignore my comments below, I see now that the code was already like this in the KB, and we should probably fix it there if we're going to fix it anywhere. That is not my task for this day.

tsnet/example_tsnet_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"tailscale.com/tsnet"
+)
+
+// ExampleServer shows you how to construct a ready-to-use tsnet instance.
+func ExampleServer() {
+	srv := new(tsnet.Server)
```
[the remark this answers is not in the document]
Xe: Server needs to be a pointer receiver.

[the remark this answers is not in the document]
creachadair: Yes, that's fine, the compiler will take the address in the call. But as noted, this is the way the KB has it, so we might as well keep them in sync.

Issue #6954, opened 2023-01-13
serve: add user identity header when used as an HTTP server


shayne:
When running using `serve` as an HTTP server we should add a header (name/value TBD) that passes along user identity.

This was discussed internally (https://tailscale.slack.com/archives/CQ5SPRTV4/p1672864924006679?thread_ts=1672855409.699419&cid=CQ5SPRTV4).

Xe:
Bikeshed:

```http
Server: tailscaled
```

soniaappasamy:
[e1cdcf7](https://github.com/tailscale/tailscale/commit/e1cdcf770803090d1dd451e4c748792082eb1fec) added two new headers
* `Tailscale-User-Login`: Filled with requester's login name.
* `Tailscale-User-Name`: Filled with requester's display name.

Follow-up coming to include a `Tailscale-Headers-Info` disclaimer header with link to docs explaining these headers, their values, and security considerations around usage.

soniaappasamy:
Closing this out, new headers released in 1.44. https://tailscale.com/s/serve-headers

PR #8297, opened 2023-06-07
ipn/ipnlocal: add identity headers to HTTP serve proxy

soniaappasamy:
Adds two new headers to HTTP serve proxy:
- `Tailscale-User-Login`: Filled with requester's login name.
- `Tailscale-User-Name`: Filled with requester's display name.

These headers only get filled when the SrcAddr is associated with a non-tagged (i.e. user-owned) node within the client's Tailnet.

The headers are passed through empty when the request originated from another tailnet, or the public internet (via funnel).
    
Updates https://github.com/tailscale/tailscale/issues/6954

### DIFF
--- a/ipn/ipnlocal/serve.go
+++ b/ipn/ipnlocal/serve.go
@@ -398,6 +398,11 @@ func (b *LocalBackend) tcpHandlerForServe(dport uint16, srcAddr netip.AddrPort) 
 	return nil
 }
 
+func getServeHTTPContext(r *http.Request) (c *serveHTTPContext, ok bool) {
+	c, ok = r.Context().Value(serveHTTPContextKey{}).(*serveHTTPContext)
+	return c, ok
+}
+
 func (b *LocalBackend) getServeHandler(r *http.Request) (_ ipn.HTTPHandlerView, at string, ok bool) {
 	var z ipn.HTTPHandlerView // zero value
 
@@ -405,7 +410,7 @@ func (b *LocalBackend) getServeHandler(r *http.Request) (_ ipn.HTTPHandlerView, 
 		return z, "", false
 	}
 
-	sctx, ok := r.Context().Value(serveHTTPContextKey{}).(*serveHTTPContext)
+	sctx, ok := getServeHTTPContext(r)
 	if !ok {
 		b.logf("[unexpected] localbackend: no serveHTTPContext in request")
 		return z, "", false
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/serve.go
```diff
@@ -469,6 +466,37 @@ func (b *LocalBackend) proxyHandlerForBackend(backend string) (*httputil.Reverse
...
+		// No identity headers to add.
+		return
+	}
+	if !node.IsTagged() {
+		r.Out.Header.Set("X-Tailscale-User-Login", user.LoginName)
+		r.Out.Header.Set("X-Tailscale-User-Name", user.DisplayName)
+	} else {
+		r.Out.Header.Set("X-Tailscale-Tags", strings.Join(node.Tags, ","))
+	}
+	r.Out.Header.Set("X-Tailscale-Node", node.ComputedName)
```
bradfitz, line 497: We should drop the "X-": https://www.rfc-editor.org/rfc/rfc6648

Nobody will conflict with "Tailscale-" prefixed headers. :)

soniaappasamy: Updated to `X-Webauth-User` and `X-Webauth-Name` as more standard headers used by other services, such as grafana and gitea.

soniaappasamy: Caught up with @bradfitz on Slack. I'm opting to go with `Tailscale-User-Login` and `Tailscale-User-Name` headers, and will add a followup commit making these easily configurable against other services, such as grafana. This should still keep things easily plug-and-play with common services, but lets us avoid being tied to other service's concepts of "users"/"logins". Also will allow us to keep consistent `Tailscale-` namings for tagged identity/other fields in the future that may not have matching equivalents in other services.

cc @willnorris who had some thoughts on header names.

ipn/ipnlocal/serve.go
```diff
@@ -398,14 +398,19 @@ func (b *LocalBackend) tcpHandlerForServe(dport uint16, srcAddr netip.AddrPort)
 	return nil
 }
 
+func getServeHTTPContext(r *http.Request) (c *serveHTTPContext, ok bool) {
+	c, ok = r.Context().Value(serveHTTPContextKey{}).(*serveHTTPContext)
+	return c, ok
```
bradfitz, line 403: two-three line functions is one of the few times naked returns are permissible if you prefer. either way.

bradfitz (approved):

PR #8301, opened 2023-06-07
wgengine: do not terminate the process when in a debugger breakpoint

raggi:
The watchdog attempts to check if an operation has locked up, but this happens as a natural course of running under a debugger. The output is noisy, and the final operation is fatal. Replace that with a short output instead when $TS_DEBUGGER is set.

### DIFF
--- a/wgengine/watchdog.go
+++ b/wgengine/watchdog.go
@@ -60,6 +60,11 @@ type watchdogEngine struct {
 	inFlightCtr uint64
 }
 
+// inDebugger returns true if the process has $TS_DEBUGGER set. This should be
+// used to avoid terminating the process when operations take an unusual amount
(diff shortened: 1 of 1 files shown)


wgengine/watchdog.go
```diff
@@ -60,6 +60,11 @@ type watchdogEngine struct {
 	inFlightCtr uint64
 }
 
+// inDebugger returns true if the process has $TS_DEBUGGER set. This should be
+// used to avoid terminating the process when operations take an unusual amount
+// of time, as would be caused by a debugger breakpoint.
+var inDebugger = envknob.RegisterBool("TS_DEBUGGER")
```
bradfitz, line 66: we already have `TS_DEBUG_DISABLE_WATCHDOG`. Just use that instead? Or is there a difference?

PR #8303, opened 2023-06-07
tka: provide verify-deeplink local API endpoint

agottardo:
Provides a local API endpoint to verify a `sign-device` deeplink, which can be scanned using the mobile/desktop apps to sign a device for Tailnet lock.

Fixes https://github.com/tailscale/tailscale/issues/8302

### DIFF
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -104,6 +104,7 @@ var handler = map[string]localAPIHandler{
 	"tka/force-local-disable":     (*Handler).serveTKALocalDisable,
 	"tka/affected-sigs":           (*Handler).serveTKAAffectedSigs,
 	"tka/wrap-preauth-key":        (*Handler).serveTKAWrapPreauthKey,
+	"tka/verify-deeplink":         (*Handler).serveTKAVerifySigningDeeplink,
 	"upload-client-metrics":       (*Handler).serveUploadClientMetrics,
 	"watch-ipn-bus":               (*Handler).serveWatchIPNBus,
 	"whois":                       (*Handler).serveWhoIs,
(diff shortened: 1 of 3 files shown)


ipn/localapi/localapi.go
```diff
@@ -1610,6 +1611,41 @@ func (h *Handler) serveTKAWrapPreauthKey(w http.ResponseWriter, r *http.Request)
 	w.Write([]byte(wrappedKey))
 }
 
+func (h *Handler) serveTKAVerifyDeeplink(w http.ResponseWriter, r *http.Request) {
```
twitchyliquid64, line 1614: Is the plan for this to verify other deeplinks in the future? if not, maybe we should specialize the method name to be about the sign device deeplink?

tka/deeplink.go
```diff
@@ -0,0 +1,161 @@
...
+// tailscale://sign-device/v1/?nk=xxx&tp=xxx&dn=xxx&os=xxx&em=xxx&hm=xxx
+//
+// where:
+// - "nk" is the nodekey of the node being signed
+// - "tp" is the tailnet lock public key
+// - "dn" is the name of the node
+// - "os" is the operating system of the node
+// - "em" is the email address associated with the node
+// - "hm" is a SHA-256 HMAC computed over the concatenation of the above fields
+func ValidateDeeplink(urlString string, stateID uint64) DeeplinkValidationResult {
```
twitchyliquid64, line 56: Maybe make this a method on a `*tka.Authority`, so it can just get the stateID automatically?

ipn/localapi/localapi.go
agottardo: done

tka/deeplink.go
agottardo: done

tka/deeplink.go
```diff
@@ -0,0 +1,162 @@
...
+			IsValid: false,
+			Error:   fmt.Errorf("missing hm (HMAC) query parameter"),
+		}
+	}
+
+	components := []string{nodeKey, tlPub, deviceName, osName, emailAddress}
+	stateID1, _ := a.StateIDs()
+	computedHMAC := generateHMAC(stateID1, components)
+
+	if computedHMAC != hmacString {
```
rosszurowski, line 147: Not sure how relevant it is here, but the `hmac` package has an explicit `Equals` function which doesn't leak timing information. Might be a better practice to have `generateHMAC` return a `[]byte` and then use:

```go
if !hmac.Equals(computedHMAC, hmacString) {
  return DeeplinkValidationResult{...}
}
```

rosszurowski (approved):

PR #8309, opened 2023-06-08
util/slicesx: add Partition function

andrew-d:
Change-Id: If97995ca9ee9fad40f327420dcb1857dd7ea2315

### DIFF
--- a/util/slicesx/slicesx.go
+++ b/util/slicesx/slicesx.go
@@ -42,3 +42,18 @@ func Shuffle[S ~[]T, T any](s S) {
 		s[i], s[j] = s[j], s[i]
 	}
 }
+
+// Partition returns two slices, the first containing the elements of the input
+// slice for which the callback evaluates to true, the second containing the rest.
+//
+// This function does not mutate s.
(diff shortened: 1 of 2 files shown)


util/slicesx/slicesx.go
```diff
@@ -42,3 +42,16 @@ func Shuffle[S ~[]T, T any](s S) {
 		s[i], s[j] = s[j], s[i]
 	}
 }
+
+// Partition returns two slices, the first containing the elements of the input
+// slice for which the callback evaluates to true, the second containing the rest.
+func Partition[S ~[]T, T any](s S, cb func(T) bool) (a, b S) {
```
creachadair, line 48: Side note: In the event you want a version that doesn't allocate (by permuting the slice in place), I wrote https://pkg.go.dev/github.com/creachadair/mds/slice#Partition for that.

andrew-d: Oh neat! That's super handy to know about 👍

bradfitz, line 48: maybe call the results `trues, falses S` ?

and document that it doesn't mutate s?

PR #8311, opened 2023-06-09
ipn/ipnlocal,wgengine/netstack: move LocalBackend specifc serving logic to LocalBackend

maisem:
The netstack code had a bunch of logic to figure out if the LocalBackend should handle an incoming connection and then would call the function directly on LocalBackend. Move that logic to LocalBackend and refactor the methods to return conn handlers.

Updates #cleanup


andrew-d (approved):

ipn/ipnlocal/serve.go
```diff
@@ -298,35 +299,40 @@ func (b *LocalBackend) HandleIngressTCPConn(ingressPeer *tailcfg.Node, target ip
 			return
 		}
 	}
-	// TODO(bradfitz): pass ingressPeer etc in context to HandleInterceptedTCPConn,
+	// TODO(bradfitz): pass ingressPeer etc in context to TCPHandlerForServe,
```
bradfitz, line 302: tcpHandlerForServe ?

ipn/ipnlocal/serve.go
```diff
@@ -298,35 +299,40 @@ func (b *LocalBackend) HandleIngressTCPConn(ingressPeer *tailcfg.Node, target ip
...
 	// extend serveHTTPContext or similar.
-	b.HandleInterceptedTCPConn(dport, srcAddr, getConn, sendRST)
+	handler := b.tcpHandlerForServe(dport, srcAddr)
+	if handler == nil {
+		sendRST()
+		return
+	}
+	c, ok := getConn()
+	if !ok {
+		b.logf("localbackend: getConn didn't complete from %v to port %v", srcAddr, dport)
```
bradfitz, line 311: we don't want to `sendRST` here too?

bradfitz (commented):
I didn't review super carefully but no objections. I'm fine with @andrew-d's review.

ipn/ipnlocal/serve.go
maisem: `getConn` already does that. I renamed it to `getConnOrReset` to be clearer.

Issue #8320, opened 2023-06-11
port to WASI


bradfitz:
The upcoming Go 1.21 supports WASI: https://tip.golang.org/doc/go1.21#wasip1

Tracking bug to add Go support.

This will permit running on e.g. Fastly Compute@Edge (https://developer.fastly.com/learning/compute/) and other such things.

PR #8321, opened 2023-06-11
all: adjust some build tags for wasi

bradfitz:
A start.

Updates #8320


cmd/derper/depaware.txt
```diff
@@ -180,6 +180,7 @@ tailscale.com/cmd/derper dependencies: (generated by github.com/tailscale/depawa
         bytes                                                        from bufio+
         compress/flate                                               from compress/gzip+
         compress/gzip                                                from internal/profile+
+   L    compress/zlib                                                from debug/elf
```
bradfitz, line 183: These are due to a change in x/sys/unix (it depending on debug/elf now) and we need the newer x/sys/unix for wasip1 consts for josharian/native, etc.

bradfitz: I sent https://go-review.googlesource.com/c/sys/+/502356 to fix upstream.

types/logger/rusage_stub.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
-//go:build windows || js
+//go:build windows || js || wasip1
```
DentonGentry, line 4: While harmless to leave here, the `js` build tag support in the filch files is being removed.

DentonGentry (approved):

types/logger/rusage_stub.go
bradfitz: It's one of:

GOARCH=wasm GOOS=js
GOARCH=wasm GOOS=wasip1

So using build tag or filename with `wasm` effectively gets both, as those are the only two geese (GOOSes) with GOARCH=wasm.

PR #8337, opened 2023-06-13
prober: make DERP probes fail when stun requests fail

raggi:
The client treats a single lost stun request as a failure in netcheck and will bounce regions as a result. The prober results should represent this if it is happening regularly in a region we need to know about that. The fist step is to report the data with more precision; the likely immediate next step is to adjust alerting to soften the blow.

Updates tailscale/corp#11492

### DIFF
--- a/prober/derp.go
+++ b/prober/derp.go
@@ -13,6 +13,7 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"net/netip"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


prober/derp.go
```diff
@@ -209,47 +210,44 @@ func derpProbeUDP(ctx context.Context, ipStr string, port int) (latency time.Dur
...
-		}
-		if txBack != tx {
-			return 0, fmt.Errorf("read wrong tx back from %v", ip)
-		}
-		if latency == 0 || d < latency {
-			latency = d
+	addr := netip.AddrPortFrom(ip, uint16(port))
+
+	// Binding requests and responses are fairly small (~40 bytes),
+	// but in practice a STUN response can be up to the size of the
```
danderson, line 223: Why can it be that? At least from our stun servers the size should be static and small, shouldn't it?

danderson (approved):

prober/derp.go
```diff
@@ -209,47 +210,44 @@ func derpProbeUDP(ctx context.Context, ipStr string, port int) (latency time.Dur
...
+	_, err = uc.WriteToUDPAddrPort(req, addr)
+	if err != nil {
+		return 0, err
+	}
+	t0 := time.Now()
+	n, _, err := uc.ReadFromUDP(buf)
+	d := time.Since(t0)
+	if err != nil {
+		if ctx.Err() != nil || err == context.DeadlineExceeded || d > time.Second {
+			return 0, fmt.Errorf("timeout reading from %s: (%s) %w", addr, d, err)
```
valscale, line 238: Tiny nit: it seems like "d > time.Second" could mask a different error, the one recorded in err.

prober/derp.go
```diff
@@ -209,47 +210,44 @@ func derpProbeUDP(ctx context.Context, ipStr string, port int) (latency time.Dur
...
+		return 0, fmt.Errorf("error reading from %s: (%s) %w", addr, d, err)
+	}
+	txBack, _, err := stun.ParseResponse(buf[:n])
+	if err != nil {
+		return 0, fmt.Errorf("parsing STUN response from %s: %v", addr, err)
+	}
+	if txBack != tx {
+		return 0, fmt.Errorf("read wrong tx back from %s", addr)
+	}
+	if latency == 0 || d < latency {
```
valscale, line 249: I think this can now be just 'latency = d'?

prober/derp.go
```diff
@@ -209,47 +210,44 @@ func derpProbeUDP(ctx context.Context, ipStr string, port int) (latency time.Dur
...
-		}
-		if latency == 0 || d < latency {
-			latency = d
+	addr := netip.AddrPortFrom(ip, uint16(port))
+
+	// Binding requests and responses are fairly small (~40 bytes),
+	// but in practice a STUN response can be up to the size of the
+	// path MTU, so we use a jumbo frame size buffer here.
+	buf := make([]byte, 9000)
+
```
valscale, line 226: I notice this removes the 'uc.SetReadDeadline(time.Now().Add(2 * time.Second))' from the previous code - just checking if there is a default deadline/timeout?

valscale (commented):
Thanks for fixing this!

PR #8346, opened 2023-06-14
wgengine/magicsock: ensure that an existing timer is stopped

raggi:
There presently is not a bug here, but this is a time-bomb waiting for an innocent incremental change to accidentally adjust the parent call chain resulting in a timer leak.

Updates #cleanup


valscale (approved):

wgengine/magicsock/magicsock.go
```diff
@@ -4441,7 +4441,10 @@ func (de *endpoint) heartbeat() {
 	de.mu.Lock()
 	defer de.mu.Unlock()
 
-	de.heartBeatTimer = nil
+	if de.heartBeatTimer != nil {
+		de.heartBeatTimer.Stop()
```
bradfitz, line 4445: if you want, you could even check that this returns false and panic or at least log [unexpected]. The invariant is that `heartBeatTimer` of nil means the timer isn't armed. And the only way into this method is via the timer firing. (which is why the first thing this does is nil out the non-dead timer, to preserve the invariant that nil means not armed, before any of the exit paths below)

at least add a comment like `// but this should never be armed`, otherwise it's misleading, making readers think it might still be firing here, in which case Stop could legit return true (stopped) or false (missed a race and another callback into heartbeat is right behind us, even though that never happens by design).

bradfitz (approved):

PR #8348, opened 2023-06-14
ipn/ipnlocal: add docs header to serve HTTP proxy

soniaappasamy:
Adds a `Tailscale-Headers-Info` header whenever the `Tailscale-User-` headers are filled from the HTTP proxy handler.

Planning on hooking this shorturl up to KB docs about the header values (i.e. what's a login name vs. display name) and security considerations to keep in mind while using these headers - notibly that they can also be filled from external requests that do not hit tailscaled.

Updates https://github.com/tailscale/tailscale/issues/6954

### DIFF
--- a/ipn/ipnlocal/serve.go
+++ b/ipn/ipnlocal/serve.go
@@ -482,6 +482,7 @@ func (b *LocalBackend) addTailscaleIdentityHeaders(r *httputil.ProxyRequest) {
 	// Clear any incoming values squatting in the headers.
 	r.Out.Header.Del("Tailscale-User-Login")
 	r.Out.Header.Del("Tailscale-User-Name")
+	r.Out.Header.Del("Tailscale-Headers-Info")
 
 	c, ok := getServeHTTPContext(r.Out)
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/serve.go
```diff
@@ -498,6 +499,7 @@ func (b *LocalBackend) addTailscaleIdentityHeaders(r *httputil.ProxyRequest) {
 	}
 	r.Out.Header.Set("Tailscale-User-Login", user.LoginName)
 	r.Out.Header.Set("Tailscale-User-Name", user.DisplayName)
+	r.Out.Header.Set("Tailscale-Headers-Info", "https://tailscale.com/serve-headers")
```
willnorris, line 502: should this be a /s/ URL?  I'm assuming this will redirect somewhere?

soniaappasamy: Oh yes! Didn't realize we had that convention, updating

shayne (approved):

PR #8355, opened 2023-06-15
tool/gocross: properly set simulator deployment target

nickoneill:
Fix from @agottardo, I am just a video-chat-to-keyboard interface


agottardo (approved):
dead 💀

Issue #8357, opened 2023-06-16
FR: Add support for HTTP to `tailscale serve`


shayne:
It has come up a few times that adding HTTP support for `tailscale serve` (tailnet only) would be useful. This allows you to access services over the tailnet using the MagicDNS name without the DNS Suffix (Tailnet DNS name).

PR #8358, opened 2023-06-16
{cmd/tailscale/cli,ipn}: add http support to tailscale serve

shayne:
Updates #8357

### DIFF
--- a/cmd/tailscale/cli/serve.go
+++ b/cmd/tailscale/cli/serve.go
@@ -35,13 +35,14 @@ func newServeCommand(e *serveEnv) *ffcli.Command {
 	return &ffcli.Command{
 		Name:      "serve",
 		ShortHelp: "Serve content and local servers",
-		ShortUsage: strings.TrimSpace(`
-serve https:<port> <mount-point> <source> [off]
-  serve tcp:<port> tcp://localhost:<local-port> [off]
-  serve tls-terminated-tcp:<port> tcp://localhost:<local-port> [off]
-  serve status [--json]
-  serve reset
-`),
+		ShortUsage: strings.Join([]string{
+			"serve http:<port> <mount-point> <source> [off]",
+			"serve https:<port> <mount-point> <source> [off]",
+			"serve tcp:<port> tcp://localhost:<local-port> [off]",
+			"serve tls-terminated-tcp:<port> tcp://localhost:<local-port> [off]",
+			"serve status [--json]",
+			"serve reset",
+		}, "\n  "),
 		LongHelp: strings.TrimSpace(`
 *** BETA; all of this is subject to change ***
 
@@ -58,8 +59,8 @@ EXAMPLES
   - To proxy requests to a web server at 127.0.0.1:3000:
     $ tailscale serve https:443 / http://127.0.0.1:3000
 
-	Or, using the default port:
-	$ tailscale serve https / http://127.0.0.1:3000
+    Or, using the default port (443):
+    $ tailscale serve https / http://127.0.0.1:3000
 
   - To serve a single file or a directory of files:
     $ tailscale serve https / /home/alice/blog/index.html
@@ -67,6 +68,12 @@ EXAMPLES
 
   - To serve simple static text:
     $ tailscale serve https:8080 / text:"Hello, world!"
+
+  - To serve over HTTP (tailnet only):
+    $ tailscale serve http:80 / http://127.0.0.1:3000
+
+    Or, using the default port (80):
+    $ tailscale serve http / http://127.0.0.1:3000
 
   - To forward incoming TCP connections on port 2222 to a local TCP server on
     port 22 (e.g. to run OpenSSH in parallel with Tailscale SSH):
@@ -175,6 +182,7 @@ func (e *serveEnv) getLocalClientStatus(ctx context.Context) (*ipnstate.Status, 
 // serve config types like proxy, path, and text.
 //
 // Examples:
+// - tailscale serve http / http://localhost:3000
 // - tailscale serve https / http://localhost:3000
 // - tailscale serve https /images/ /var/www/images/
 // - tailscale serve https:10000 /motd.txt text:"Hello, world!"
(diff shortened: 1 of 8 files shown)


cmd/tailscale/cli/serve.go
```diff
@@ -171,6 +171,15 @@ func (e *serveEnv) getLocalClientStatus(ctx context.Context) (*ipnstate.Status,
 	return st, nil
 }
 
+// parsePort parses a port string into a uint16.
+func (e *serveEnv) parsePort(portStr string) (uint16, error) {
```
bradfitz, line 175: the receiver is never used. Do you plan to make it used? For what?

ipn/serve.go
```diff
@@ -150,19 +156,36 @@ func (sc *ServeConfig) IsTCPForwardingOnPort(port uint16) bool {
...
 
 // IsServingWeb checks if ServeConfig is currently serving
-// Web/HTTPS on the given port.
+// Web (HTTP/HTTPS) on the given port.
 // This is exclusive of TCPForwarding.
 func (sc *ServeConfig) IsServingWeb(port uint16) bool {
+	return sc.IsServingHTTP(port) || sc.IsServingHTTPS(port)
+}
+
+// IsServingHTTPS checks if ServeConfig is currently serving
```
bradfitz, line 169: for bool methods, the conventional wording is "reports whether"

ipn/serve.go
```diff
@@ -150,19 +156,36 @@ func (sc *ServeConfig) IsTCPForwardingOnPort(port uint16) bool {
...
+// HTTP on the given port.
```
bradfitz, line 170: HTTPS

cmd/tailscale/cli/serve.go
shayne: https://github.com/tailscale/tailscale/blob/a82c3cae977a3e9f86ab3ddcb85198491d052b1a/cmd/tailscale/cli/serve.go#L231

https://github.com/tailscale/tailscale/blob/a82c3cae977a3e9f86ab3ddcb85198491d052b1a/cmd/tailscale/cli/serve.go#L686

I am using it here and there.

bradfitz: 🤦‍♂️

bradfitz: Wait, no, I mean in this method. Why is this a method and not a function?

shayne: You're right! 🤦‍♂️ You said receiver, and I 🧠💨 . I moved it to a function and gave it a more specific name since we're in the somewhat sprawling `cli` package.

ipn/serve.go
shayne: Updated these and all the others in this file. Spring cleaning.

cmd/tailscale/cli/serve.go
```diff
@@ -725,3 +751,16 @@ func (e *serveEnv) runServeReset(ctx context.Context, args []string) error {
 	sc := new(ipn.ServeConfig)
 	return e.lc.SetServeConfig(ctx, sc)
 }
+
+// parseValidServePort parses a port number from a string and returns it as a
+// uint16. It returns an error if the port number is invalid or zero.
+func parseValidServePort(s string) (uint16, error) {
```
bradfitz, line 757: I think the word "Valid" is kinda redundant here. It returns an error; that basically already says that it returns an error if the input is not valid. parseServePort works just as well.

ipn/ipnlocal/local.go
```diff
@@ -4131,6 +4131,10 @@ func (b *LocalBackend) setServeProxyHandlersLocked() {
 	b.serveConfig.Web().Range(func(_ ipn.HostPort, conf ipn.WebServerConfigView) (cont bool) {
 		conf.Handlers().Range(func(_ string, h ipn.HTTPHandlerView) (cont bool) {
 			backend := h.Proxy()
+			if backend == "" {
+				// Skip if not a proxy handler
+				return true
+			}
```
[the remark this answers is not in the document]
bradfitz: add a comment then that we shouldn't get here?

ipn/serve.go
```diff
@@ -128,9 +134,8 @@ func (sc *ServeConfig) GetTCPPortHandler(port uint16) *TCPPortHandler {
 	return sc.TCP[port]
 }
 
-// IsTCPForwardingAny checks if ServeConfig is currently forwarding
-// in TCPForward mode on any port.
-// This is exclusive of Web/HTTPS serving.
+// IsTCPForwardingAny reports whether if ServeConfig is currently forwarding in
```
bradfitz, line 137: "whether if" => whether

here and elsewhere

bradfitz (approved):

cmd/tailscale/cli/serve.go
```diff
@@ -36,7 +36,8 @@ func newServeCommand(e *serveEnv) *ffcli.Command {
 		Name:      "serve",
 		ShortHelp: "Serve content and local servers",
 		ShortUsage: strings.TrimSpace(`
-serve https:<port> <mount-point> <source> [off]
+serve http:<port> <mount-point> <source> [off]
+  serve https:<port> <mount-point> <source> [off]
```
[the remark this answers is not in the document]
shayne: Switched to `strings.Join()` and updated the `funnel` cmd too.

cmd/tailscale/cli/serve.go
```diff
@@ -665,20 +672,37 @@ func printTCPStatusTree(ctx context.Context, sc *ipn.ServeConfig, st *ipnstate.S
...
+		scheme = "http"
 	}
+
+	portPart := ":" + portStr
+	if scheme == "http" && portStr == "80" ||
+		scheme == "https" && portStr == "443" {
+		portPart = ""
+	}
+	if scheme == "http" {
+		hostname := strings.SplitN(host, ".", 2)[0]
```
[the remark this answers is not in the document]
shayne: Done.

cmd/tailscale/cli/serve.go
shayne: Removed!

ipn/ipnlocal/local.go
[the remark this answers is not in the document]
shayne: Updated the comment. We range over the handlers but only want to create a serve proxy handler for the `Proxy()` type serves.

ipn/serve.go
shayne: Thx, done!

PR #8360, opened 2023-06-16
tailcfg: add location field to hostinfo

catzkorn:
This change adds Location and ExitNodePriority fields to HostInfo. Location contains the option for a Country and a City. Neither of these fields are populated by default.

The ExitNodePriority field is used to determine the priority an exit node should be given for use, if the field is set. The higher the value set, the higher priority the node should be given for use.

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -3,7 +3,7 @@
 
 package tailcfg
 
-//go:generate go run tailscale.com/cmd/viewer --type=User,Node,Hostinfo,NetInfo,Login,DNSConfig,RegisterResponse,DERPRegion,DERPMap,DERPNode,SSHRule,SSHAction,SSHPrincipal,ControlDialPlan --clonefunc
+//go:generate go run tailscale.com/cmd/viewer --type=User,Node,Hostinfo,NetInfo,Login,DNSConfig,RegisterResponse,DERPRegion,DERPMap,DERPNode,SSHRule,SSHAction,SSHPrincipal,ControlDialPlan,Location --clonefunc
 
 import (
 	"bytes"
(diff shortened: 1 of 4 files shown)


tailcfg/tailcfg.go
```diff
@@ -584,6 +591,12 @@ type Hostinfo struct {
 	Cloud           string         `json:",omitempty"`
 	Userspace       opt.Bool       `json:",omitempty"` // if the client is running in userspace (netstack) mode
 	UserspaceRouter opt.Bool       `json:",omitempty"` // if the client's subnet router is running in userspace (netstack) mode
+	Location        *Location      `json:",omitempty"`
```
bradfitz, line 594: docs

tailcfg/tailcfg.go
```diff
@@ -584,6 +591,12 @@ type Hostinfo struct {
...
+
+	// ExitNodePriority indicates the priority an exit node is given.
+	// A higher value indicates that the exit node is more preferable
+	// for use.
+	ExitNodePriority int `json:",omitempty"`
```
bradfitz, line 599: is zero special at all? (any default value?)

is negative allowed?

tailcfg/tailcfg.go
raggi: Is it necessary to have fields be exclusively one or the other?

In the initial use case, control provides these fields, but there are other fairly obvious use cases where users may want to set them to make use of the same client capabilities.

That is, if the client supports "pick by location, biased by priority", it's reasonable to enable users who manage their own nodes be able to make use of that feature set, in order for them to do so it's reasonable for them to use these fields - unless we have or need a separation based on where data can originate, in which case we need two sets of fields for the data, and to rationalize how they're used.

bradfitz: Sure. here's fine if you think we might send them up later.

tailcfg/tailcfg.go
```diff
@@ -585,6 +594,18 @@ type Hostinfo struct {
...
+	// Tailscale host. Location is optional and only set if
+	// explicitly declared by a node.
+	Location *Location `json:",omitempty"`
+
+	// ExitNodePriority indicates the priority an exit node is given.
+	// A higher value indicates that the exit node is more preferable
+	// for use.
+	// A value of 0 means the exit node does not have a priority
+	// preference. A negative int is not allowed.
+	ExitNodePriority int `json:",omitempty"`
```
maisem, line 607: I don't think HostInfo is the right place for this field. This should be on the node object. It is not declared by a node, but is instead declared by the coordination server

catzkorn:
Discussed with @maisem in slack. Switching the fields over to the `Node` struct and Maisem is going to follow up with a PR to update docs to `HostInfo` and add docs to `Node` to detail the nuance between the two/when to choose one over the other.

tailcfg/tailcfg.go
```diff
@@ -531,6 +531,24 @@ type Service struct {
 	// TODO(apenwarr): add "tags" here for each service?
 }
 
+// Location represents geographical location data about a
+// Tailscale host. Location is optional and only set if
+// explicitly declared by a node.
+type Location struct {
+	Country     string `json:",omitempty"` // e.g "Canada"
```
raggi, line 538:
```suggestion
	Country     string `json:",omitempty"` // User friendly country name, with proper capitalization, e.g "Canada"
```

tailcfg/tailcfg.go
```diff
@@ -531,6 +531,24 @@ type Service struct {
...
+	CountryCode string `json:",omitempty"` // ISO 3166-1 alpha-2, e.g "ca"
+	City        string `json:",omitempty"`
```
raggi, line 540:
```suggestion
	CountryCode string `json:",omitempty"` // ISO 3166-1 alpha-2 in lower case, e.g "ca"
	City        string `json:",omitempty"` // User friendly city name, with proper capitalization, e.g. "Squamish"
```

raggi (approved):

PR #8362, opened 2023-06-16
util/linuxfw: decoupling IPTables logic from linux router

KevinLiang10:
### Summary:

This change is introducing new `netfilterRunner` interface and moving iptables manipulation to a lower leveled iptables runner to decouple router_linux from IPtables APIs. This will allow manipulating Netfilter rules via other tools.

### Issue related

Updates: #391 

### Changes made

1. Added `iptableRunner` and move iptables related task from linuxRouter (in router_linux.go) to `iptableRunner` in (linuxfw/iptables.go)
2. Added `iptablesInterface` to abstract `iptables.IPTable` like currently we do in netfilterRunner interface, on "operation" basis (like Insert, Append, Delete).
3. netfilterRunner interface now defines methods on "task" level (like `AddNetfilterHooks`), instead of "operation" level that was strongly coupled with iptables.
4. Added unit test for new class iptableRunner by reusing our current mock to `iptables.IPTable` in router_linux_test.go. These tests assumes the correctness of `iptablesInterface`'s implementation, and focuses on testing `iptableRunner`'s behaviour.
5. Updated unit test for linuxRouter, by adding a mock to `netfilterRunner` assuming the correctness of `netfilterRunner`'s implementation, and focuses on testing `linuxRouter`'s behaviour.

### Thought Process

1. For change number 5 above the mock always have `V6Available` and `V6NATAvailable` set to true for coverage of test. These values don't affect router's logic, but just asks router to do extra task like "do the same thing for V6"
2. Also for change number 5 I used IPtables rules to mock `netfilterRunner`, but this is irrelevant because this is a mock anyways. Correctness of which ever `netfilterRunner` we use will be proven in their own unit tests.
3. In change number 4 I thought about adding some failing tests like test AddLoopbackRule with malformed address, but `netip.Addr` is always correctly formed. Another test I thought of was testing for failure on we add rules when chain doesn't exist, but again that's a correctness test to `iptablesInterface`, which we don't really care in this unit test. Any suggestion on how additional tests may help would be appreciated.

### Expected Result

Tailscale and Tailscaled should be behaving the exact same way as it used to be. In particular, when tailscale up is called, the tailscale firewall rules should be injected. like

Screen Shot 2023-06-16 at 3 42 51 PM

As an example for IPv4 rules.

### How to test

On linux environment do the following:

1. Run a local control server and login admin
2. Bring up tailscaled by `sudo ./tool/go run ./cmd/tailscaled --socket=/tmp/ts/ts1.sock --state=/tmp/ts/ts1.state --tun=ts1`
3. Run `sudo ./tool/go run ./cmd/tailscale --socket=/tmp/ts/ts1.sock up -login-server=http://localhost:31544`
4. authenticate
5. Run `sudo iptables --list` and `sudo ip6tables --list` to view the base rules added, and do whatever will modify rules (like change `NetfilterMode`) to test further.


util/linuxfw/iptables.go
```diff
@@ -823,3 +825,479 @@ type markTarget struct {
...
+	Insert(table, chain string, pos int, args ...string) error
+	Append(table, chain string, args ...string) error
+	Exists(table, chain string, args ...string) (bool, error)
+	Delete(table, chain string, args ...string) error
+	ClearChain(table, chain string) error
+	NewChain(table, chain string) error
+	DeleteChain(table, chain string) error
+}
+
+type iptableRunner struct {
```
twitchyliquid64, line 841: s/iptableRunner/iptablesRunner

util/linuxfw/iptables.go
```diff
@@ -823,3 +825,479 @@ type markTarget struct {
...
+	ipt4 iptablesInterface
+	ipt6 iptablesInterface
+
+	v6Available    bool
+	v6NATAvailable bool
+}
+
+func NewIptableRunner(logf logger.Logf) (*iptableRunner, error) {
```
twitchyliquid64, line 849: Plz add a doc-string describing any new exported function: https://tip.golang.org/doc/comment#Funcs

util/linuxfw/iptables.go
```diff
@@ -823,3 +825,479 @@ type markTarget struct {
...
+	ipt4, err := iptables.NewWithProtocol(iptables.ProtocolIPv4)
+	if err != nil {
+		return nil, err
+	}
+
+	v6err := checkIPv6(logf)
+	if v6err != nil {
```
twitchyliquid64, line 856: Some of this logic could be simplified if you nest the if statements:

```go
if v6err != nil {
  log(...)
} else {
  supportsV6 = true
  ...
}
```

util/linuxfw/iptables.go
```diff
@@ -823,3 +825,479 @@ type markTarget struct {
...
+	}
+	return newIptableRunnerAdvanced(ipt4, ipt6, supportsV6, supportsV6NAT), nil
+}
+
+// newIptableRunnerAdvanced is a helper function for testing purposes
+func newIptableRunnerAdvanced(ipt4, ipt6 iptablesInterface, supV6, supV6NAT bool) *iptableRunner {
+	return &iptableRunner{ipt4, ipt6, supV6, supV6NAT}
+}
+
+func (i *iptableRunner) GetV6Available() bool {
```
twitchyliquid64, line 880: Ditto on documenting exported functions, here and below

util/linuxfw/iptables.go
```diff
@@ -823,3 +825,479 @@ type markTarget struct {
...
+	return nil
+}
+
+// tsChain returns the name of the tailscale sub-chain corresponding
+// to the given "parent" chain (e.g. INPUT, FORWARD, ...).
+func tsChain(chain string) string {
+	return "ts-" + strings.ToLower(chain)
+}
+
+// getIptablebyProto returns the iptablesInterface in iptables runner for the given protocol.
```
twitchyliquid64, line 909: Nit: Maybe more functional? i.e.: `getIptablebyProto returns an object you can use to manipulate IP tables for the given protocol.`

util/linuxfw/iptables.go
```diff
@@ -823,3 +825,479 @@ type markTarget struct {
...
+		return []iptablesInterface{i.ipt4, i.ipt6}
+	}
+	return []iptablesInterface{i.ipt4}
+}
+
+// addNetfilterHooks inserts calls to tailscale's netfilter chains in
+// the relevant main netfilter chains. The tailscale chains must
+// already exist.
+func (i *iptableRunner) AddNetfilterHooks() error {
+	divert := func(ipt iptablesInterface, table, chain string) error {
```
twitchyliquid64, line 948: Whats the function of `divert()`? What is it diverting?

util/linuxfw/iptables.go
```diff
@@ -823,3 +825,479 @@ type markTarget struct {
...
+		}
+	}
+	return nil
+}
+
+// AddNetfilterChains creates custom Tailscale chains in netfilter via iptables.
+func (i *iptableRunner) AddNetfilterChains() error {
+	create := func(ipt iptablesInterface, table, chain string) error {
+		err := ipt.ClearChain(table, chain)
+		if errCode(err) == 1 {
```
twitchyliquid64, line 988: Optional: consider making `errCode(err) == 1` be a function `isErrChainNotExist` or something. But if this is the only place we do this, its probably not worth it.

util/linuxfw/iptables.go
```diff
@@ -823,3 +825,479 @@ type markTarget struct {
...
+	}
+	if i.v6NATAvailable {
+		if err := create(i.ipt6, "nat", "ts-postrouting"); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// addNetfilterBase4 adds some basic IPv4 processing rules to be
```
twitchyliquid64, line 1017: s/addNetfilterBase4/AddNetfilterBase4

wgengine/router/router_linux.go
```diff
@@ -579,11 +529,19 @@ func (r *linuxRouter) setNetfilterMode(mode preftype.NetfilterMode) error {
 	return nil
 }
 
+func (r *linuxRouter) getV6Available() bool {
+	return r.nfr.GetV6Available()
```
twitchyliquid64, line 533: Why is this its own method? could the caller just do `r.nfr.GetV6Available()` themselves?

twitchyliquid64 (commented):
Looking good! I ran this locally as well and didn't run into any issues.

util/linuxfw/iptables.go
KevinLiang10: It was named divert before I changed it so just kept the name. Is there a better name you would suggest?

util/linuxfw/iptables.go
KevinLiang10: Sorry are you suggesting we should go with iptablesRunner?

util/linuxfw/iptables.go
KevinLiang10: Yeah that would make the code self explanatory, thank for pointing out

util/linuxfw/iptables.go
KevinLiang10: Updated

util/linuxfw/iptable_runner.go
```diff
@@ -0,0 +1,506 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// addSNATRule adds a netfilter rule to SNAT traffic destined for
+// local subnets.
+func (i *iptablesRunner) AddSNATRule() error {
+
```
catzkorn, line 405: NIT: Remove extra line

util/linuxfw/iptable_runner.go
```diff
@@ -0,0 +1,506 @@
...
+			return fmt.Errorf("adding %v in v6/nat/ts-postrouting: %w", args, err)
+		}
+	}
+	return nil
+}
+
+// delSNATRule removes the netfilter rule to SNAT traffic destined for
+// local subnets. Fails if the rule does not exist.
+func (i *iptablesRunner) DelSNATRule() error {
+
```
catzkorn, line 421: NIT: Remove extra line

util/linuxfw/iptable_runner.go
```diff
@@ -0,0 +1,506 @@
...
+		logf("ERROR: clear iptables: %w", err)
+	}
+
+	// Clear ip6tables rules
+	err = clearRules(iptables.ProtocolIPv6, logf)
+	if err != nil {
+		// If the system doesn't support ipv6, we'll just log the error and continue
+		logf("ERROR: clear ip6tables: %w", err)
+	}
+
```
catzkorn, line 448: NIT: Remove extra line

util/linuxfw/iptable_runner.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+	if i.v6NATAvailable {
+		if err := i.ipt6.Delete("nat", "ts-postrouting", args...); err != nil {
+			return fmt.Errorf("deleting %v in v6/nat/ts-postrouting: %w", args, err)
+		}
+	}
+	return nil
+}
+
+// Cleans the iptables rules created by Tailscale.
```
catzkorn, line 434: We should start docs with the function name, so in this case

```suggestion
// IptablesCleanUp clears the iptables rules for IPv4 and IPv6. If an error occurs,
// it is logged.
```

util/linuxfw/iptable_runner.go
```diff
@@ -0,0 +1,506 @@
...
+	if i.v6NATAvailable {
+		if err := i.ipt6.Append("nat", "ts-postrouting", args...); err != nil {
+			return fmt.Errorf("adding %v in v6/nat/ts-postrouting: %w", args, err)
+		}
+	}
+	return nil
+}
+
+// delSNATRule removes the netfilter rule to SNAT traffic destined for
+// local subnets. Fails if the rule does not exist.
```
catzkorn, line 419: Capitalize the function name to match, and a little NIT on the second sentence. 

```suggestion
// DelSNATRule removes the netfilter rule to SNAT traffic destined for
// local subnets. An error is returned if the rule does not exist.
```

util/linuxfw/iptable_runner.go
catzkorn: With this change, I think you could also remove the 

```
// Clear iptables rules
and
// Clear ip6tables rules
and 
// If the system doesn't support ipv6, we'll just log the error and continue
```

comments

util/linuxfw/iptable_runner.go
```diff
@@ -0,0 +1,506 @@
...
+				// nonexistent chain. nothing to do.
+				return nil
+			}
+			return fmt.Errorf("flushing %s/%s: %w", table, chain, err)
+		}
+		if err := ipt.DeleteChain(table, chain); err != nil {
+			return fmt.Errorf("deleting %s/%s: %v", table, chain, err)
+		}
+		return nil
+	}
```
catzkorn, line 491: I would move the two functions from this function into their own functions instead.

util/linuxfw/iptable_runner.go
```diff
@@ -0,0 +1,506 @@
...
+	// Clear ip6tables rules
+	err = clearRules(iptables.ProtocolIPv6, logf)
+	if err != nil {
+		// If the system doesn't support ipv6, we'll just log the error and continue
+		logf("ERROR: clear ip6tables: %w", err)
+	}
+
+}
+
+// clearRules clears all the iptables rules created by Tailscale for the given protocol.
```
catzkorn, line 451: I would add a comment about how we handle errors in here too.

catzkorn (commented):
Started going through but hopping offline for a bit - will go through the other files either tonight or tomorrow!

catzkorn (commented):
couple more

util/linuxfw/iptable_runner.go
```diff
@@ -0,0 +1,506 @@
...
+	if supportsV6 {
+		ipt6, err = iptables.NewWithProtocol(iptables.ProtocolIPv6)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return newIptablesRunnerAdvanced(ipt4, ipt6, supportsV6, supportsV6NAT), nil
+}
+
+// newIptablesRunnerAdvanced is a helper function for testing purposes
```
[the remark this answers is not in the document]
KevinLiang10: It actually exist for testing purpose but not only used for testing purpose, but I get the idea. What about 
"newIptablesRunnerAdvanced returns an iptablesRunner. It is a wrapper for iptablesRunner init, which exists for testing purpose."

util/linuxfw/iptable_runner.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+	if i.v6NATAvailable {
+		if err := divert(i.ipt6, "nat", "POSTROUTING"); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// AddNetfilterChains creates custom Tailscale chains in netfilter via iptables.
```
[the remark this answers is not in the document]
KevinLiang10: AddNetfilterChains creates custom Tailscale chains in netfilter via iptables if the ts-chain doesn't already exist.

util/linuxfw/iptable_runner.go
```diff
@@ -0,0 +1,506 @@
...
+		err := ipt.ClearChain(table, chain)
+		if isErrChainNotExist(err) {
+			// nonexistent chain. let's create it!
+			return ipt.NewChain(table, chain)
+		}
+		if err != nil {
+			return fmt.Errorf("setting up %s/%s: %w", table, chain, err)
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
KevinLiang10: I actually took this implementation from the original router_linux file. I'm not sure why it was implemented this way, I guess this implementation was trying to keep the global namespace clean. If we want to move it out we probably need better name that create/divert which would not make too much sense outside of the function. 

Note: These functions are embedded with args so technically not designed to be reusable by other function, therefore I believe should not be globally scoped

util/linuxfw/iptable_runner.go
KevinLiang10: I think this one would probably make sense since they are actually reused and can have meaningful names like delChain/clearChain.

util/linuxfw/linuxfw.go
```diff
@@ -2,10 +2,250 @@
...
+	}
+	s := err.Error()
+	if strings.HasPrefix(s, "exitcode:") {
+		code, err := strconv.Atoi(s[9:])
+		if err == nil {
+			return code
+		}
+	}
+	return -42
+}
```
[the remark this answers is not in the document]
KevinLiang10: 1. I think in linuxfw package we only care about if it's 1 (which maps to chain doesn't exist in our use cases)
2. This I believe is a tool that's used in router, to get errorCode from different types of errors . I doubt it would mean a lot for us to doc these package dependent error codes. We look for a errorCode == 2 from `osCommandRunner` in `router_linux.go`, errorCode == 1 from `nftables.ClearChain` and some other code in runGroup. These values doesn't seem necessary for us to record, and they aren't really related either. We might have another 2 from another package one day.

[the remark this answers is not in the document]
KevinLiang10: Also I copied rather than moved the errorCode implementation from `wgengine/router/runner.go` to linuxfw package and use it from import. It doesn't make sense to import this util function from another package. It's not a linuxfw specific thing. Adding this to a third util package would make some sense to me if you have a good place in mind,

KevinLiang10:
> IpTables

KevinLiang10:
> A minor style note: Go convention is to [capitalize initialisms consistently](https://github.com/golang/go/wiki/CodeReviewComments#initialisms), so, e.g., "IPTables" or "ipTables" rather than "IpTables" in this case.

@creachadair Thanks for the insight, but I failed to find a place where I use `IpTables`. Could you point the direction for me please?

creachadair:
> @creachadair Thanks for the insight, but I failed to find a place where I use `IpTables`. Could you point the direction for me please?

I had seen a couple in previous iterations but it looks like maybe they're gone now. Please ignore me, in that case 😆

wgengine/router/router_linux.go
twitchyliquid64: Thanks for the explanation!

wgengine/router/router_linux.go
```diff
@@ -1619,118 +1301,14 @@ func cidrDiff(kind string, old map[netip.Prefix]bool, new []netip.Prefix, add, d
...
-	rule.Table = tailscaleRouteTable.num
-	rule.Family = netlink.FAMILY_V6
-	// First delete the rule unconditionally, and don't check for
-	// errors. This is just cleaning up anything that might be already
-	// there.
-	netlink.RuleDel(rule)
-	// And clean up on exit.
-	defer netlink.RuleDel(rule)
-	return netlink.RuleAdd(rule)
+	linuxfw.IptablesCleanUp(logf)
```
twitchyliquid64, line 1311: Maybe only do this logic if `interfaceName != "userspace-networking"` ? In that mode, we dont mess with any IPTables rules or anything, so there isnt anything for the router to clean up.

util/linuxfw/linuxfw.go
```diff
@@ -2,10 +2,254 @@
...
+	TailscaleSubnetRouteMark    = "0x40000"
+	TailscaleSubnetRouteMarkNum = 0x40000
+	// This one is same value but padded to even number of digit, so
+	// hex decoding can work correctly.
+	TailscaleSubnetRouteMarkHexStr = "0x040000"
+
+	// Packet was originated by tailscaled itself, and must not be
+	// routed over the Tailscale network.
+	//
+	// Keep this in sync with tailscaleBypassMark in
```
raggi, line 75: We no longer need to keep this in sync, as there's only one.

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+
+	if err := nf.Delete("filter", "ts-input", "-i", "lo", "-s", addr.String(), "-j", "ACCEPT"); err != nil {
+		return fmt.Errorf("deleting loopback allow rule for %q: %w", addr, err)
+	}
+
+	return nil
+}
+
+// getTables gets the available iptablesInterface in iptables runner.
+func (i *iptablesRunner) getTables() []iptablesInterface {
```
raggi, line 139: If we had a sibling getNATTables() then we could apply this pattern to a lot of cases below that are conditional on NAT, too.

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+		if err := i.ipt6.Delete("nat", "ts-postrouting", args...); err != nil {
+			return fmt.Errorf("deleting %v in v6/nat/ts-postrouting: %w", args, err)
+		}
+	}
+	return nil
+}
+
+// IptablesCleanUp clears the netfilter rules for IPv4 and IPv6.
+// If error occurs, it's logged but not returned.
+func IptablesCleanUp(logf logger.Logf) {
```
raggi, line 408: `IPTablesCleanup`

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+	if i.v6NATAvailable {
+		if err := i.ipt6.Delete("nat", "ts-postrouting", args...); err != nil {
+			return fmt.Errorf("deleting %v in v6/nat/ts-postrouting: %w", args, err)
+		}
+	}
+	return nil
+}
+
+// IptablesCleanUp clears the netfilter rules for IPv4 and IPv6.
+// If error occurs, it's logged but not returned.
```
raggi, line 407:
```suggestion
// IPTablesCleanup removes all Tailscale added iptables rules. Any errors that occur are logged to the provided logf.
```

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+func IptablesCleanUp(logf logger.Logf) {
+	err := clearRules(iptables.ProtocolIPv4, logf)
+	if err != nil {
+		logf("ERROR: clear iptables: %w", err)
```
raggi, line 411: "linuxfw: clear iptables: %v"

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+func delChain(ipt iptablesInterface, table, chain string) error {
+	if err := ipt.ClearChain(table, chain); err != nil {
+		if isErrChainNotExist(err) {
+			// nonexistent chain. nothing to do.
+			return nil
+		}
+		return fmt.Errorf("flushing %s/%s: %w", table, chain, err)
+	}
+	if err := ipt.DeleteChain(table, chain); err != nil {
+		return fmt.Errorf("deleting %s/%s: %v", table, chain, err)
```
raggi, line 449: %w

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+func delTSHook(ipt iptablesInterface, table, chain string, logf logger.Logf) error {
+	tsChain := tsChain(chain)
+	args := []string{"-j", tsChain}
+	if err := ipt.Delete(table, chain, args...); err != nil {
+		// TODO(apenwarr): check for errCode(1) here.
+		// Unfortunately the error code from the iptables
+		// module resists unwrapping, unlike with other
+		// calls. So we have to assume if Delete fails,
+		// it's because there is no such rule.
+		logf("deleting %v in %s/%s: %w", args, table, chain, err)
```
raggi, line 432: %s or %v for errors in logf, %w for fmt.Errorf

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+// clearRules clears all the iptables rules created by Tailscale
+// for the given protocol. If error occurs, it's logged but not returned.
+func clearRules(proto iptables.Protocol, logf logger.Logf) error {
+	ipt, err := iptables.NewWithProtocol(proto)
+	if err != nil {
+		return err
+	}
+
+	if err := delTSHook(ipt, "filter", "INPUT", logf); err != nil {
+		logf("ERROR: delete hook: %w", err)
```
raggi, line 463: See format from above comment, here and below

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+}
+
+// clearRules clears all the iptables rules created by Tailscale
+// for the given protocol. If error occurs, it's logged but not returned.
+func clearRules(proto iptables.Protocol, logf logger.Logf) error {
+	ipt, err := iptables.NewWithProtocol(proto)
+	if err != nil {
+		return err
+	}
+
```
raggi, line 461: consider using https://pkg.go.dev/tailscale.com/util/multierr here, and letting the parent log once, rather than producing many individual logs.

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+	}
+	if i.v6NATAvailable {
+		if err := delTSHook(i.ipt6, "nat", "POSTROUTING", logf); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// addSNATRule adds a netfilter rule to SNAT traffic destined for
```
raggi, line 376: AddSNATRule

util/linuxfw/linuxfw.go
```diff
@@ -2,10 +2,254 @@
...
+	"github.com/tailscale/netlink"
+	"tailscale.com/types/logger"
+)
+
+type ProtoVersion int
+
+const (
+	IPv4 ProtoVersion = iota
+	IPv6
+)
```
raggi, line 28: Is this still used, it looks like it is used by methods that are unused, once we remove all of them, can these also go away?

util/linuxfw/linuxfw.go
```diff
@@ -2,10 +2,254 @@
...
+
+type MatchDecision int
+
+const (
+	Accept MatchDecision = iota
+	Masq
+)
```
raggi, line 35: Is this unused?

util/linuxfw/iptable_runner.go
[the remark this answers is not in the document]
catzkorn: Sounds good!

util/linuxfw/linuxfw.go
```diff
@@ -2,10 +2,254 @@
...
+
+// IpCmdArg returns the string form of the table to pass to the "ip" command.
+func (rt RouteTable) IpCmdArg() string {
+	if rt.Num >= 253 {
+		return rt.Name
+	}
+	return strconv.Itoa(rt.Num)
+}
+
+var routeTableByNumber = map[int]RouteTable{}
```
raggi, line 96: Are you certain that this is only ever accessed by a single goroutine? If not, it will need synchronization.

Do we need to use route table names, or can we exclusively use numbers? If we can exclusively use numbers, can we remove the need for this entirely?

util/linuxfw/linuxfw_unsupported.go
```diff
@@ -9,9 +9,15 @@
...
 
 import (
+	"errors"
+
 	"tailscale.com/types/logger"
 )
 
+// ErrUnsupported is the error returned from all functions on non-Linux
+// platforms.
+var ErrUnsupported = errors.New("unsupported")
```
raggi, line 19: linuxfw:unsupported, just in case we actually end up logging this somewhere, it'll be much easier to indentify where it came from.

wgengine/router/router_linux.go
```diff
@@ -1,10 +1,11 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
+//go:build linux
```
raggi, line 4: this is implied by the _linux suffix on the file name, and doesn't need to be added

wgengine/router/router_linux.go
```diff
@@ -38,56 +38,35 @@ const (
...
-	Insert(table, chain string, pos int, args ...string) error
-	Append(table, chain string, args ...string) error
-	Exists(table, chain string, args ...string) (bool, error)
-	Delete(table, chain string, args ...string) error
-	ClearChain(table, chain string) error
-	NewChain(table, chain string) error
-	DeleteChain(table, chain string) error
+	AddLoopbackRule(addr netip.Addr) error
+	DelLoopbackRule(addr netip.Addr) error
+	AddNetfilterHooks() error
```
raggi, line 47: We don't use "netfilter" in the name for AddLoopbackRule, or for AddSNATRule. For consistency and simplicity we can remove it from these other names as well.

wgengine/router/router_linux.go
```diff
@@ -1285,7 +1148,7 @@ func (r *linuxRouter) addNetfilterBase() error {
 	if err := r.addNetfilterBase4(); err != nil {
```
raggi, line 1148: In order rules, the difference between 4 and 6 is handled inside the netfilter runner. Can the runner interface drop AddNetfilterBase4 and AddNetfilterBase6, and just have AddNetfilterBase?

wgengine/router/router_linux.go
```diff
@@ -1296,146 +1159,35 @@ func (r *linuxRouter) addNetfilterBase() error {
...
-	// (see corresponding IPv4 CGNAT rule).
-	args = []string{"-o", r.tunname, "-j", "ACCEPT"}
-	if err := r.ipt6.Append("filter", "ts-forward", args...); err != nil {
-		return fmt.Errorf("adding %v in v6/filter/ts-forward: %w", args, err)
+	if err := r.nfr.AddNetfilterBase6(r.tunname); err != nil {
+		return err
 	}
-
 	return nil
 }
```
raggi, line 1175: these two wrapper functions addNetfilterBase4 and addNetfilterBase6 can be removed, as they're only called from above which is already performing error checking

raggi: same with most of these methods below, which likely no longer need to be present.

wgengine/router/router_linux.go
```diff
@@ -1239,43 +1127,18 @@ func (r *linuxRouter) delIPRulesWithIPCommand() error {
 	return rg.ErrAcc
 }
 
-func (r *linuxRouter) netfilterFamilies() []netfilterRunner {
-	if r.v6Available {
-		return []netfilterRunner{r.ipt4, r.ipt6}
+func (r *linuxRouter) netfilterFamilies() []linuxfw.ProtoVersion {
```
raggi, line 1130: This appears to be unused

raggi (commented):
Looking great overall.

If you start to remove the unused code and flatten out the usage fo the netfilter runner in-place, removing the extra wrappers still present in router_linux, I think it may reveal some more things that can be removed, and a couple of places where we can simplify the interface a little more.

raggi (commented):
Looking great overall.

If you start to remove the unused code and flatten out the usage fo the netfilter runner in-place, removing the extra wrappers still present in router_linux, I think it may reveal some more things that can be removed, and a couple of places where we can simplify the interface a little more.

util/linuxfw/linuxfw.go
```diff
@@ -2,10 +2,254 @@
...
+func errCode(err error) int {
+	if err == nil {
+		return 0
+	}
+	var e *exec.ExitError
+	if ok := errors.As(err, &e); ok {
+		return e.ExitCode()
+	}
+	s := err.Error()
+	if strings.HasPrefix(s, "exitcode:") {
```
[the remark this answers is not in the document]
KevinLiang10: Btw in iptables package the error we need to handle are falling under this category, tho we are not expecting these errors to actually occur.

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+	v6NATAvailable bool
+}
+
+// NewIptablesRunner returns a new iptablesRunner which implements
+// NetfilterRunner interface. It returns an error if the iptables library
+// fails to initialize. The iptablesRunner checks for IPv6 support and
+// IPv6 NAT support on initialization and stores the results. Also stores
+// iptablesInterface for IPv4 and IPv6. In production this interface
+// is a wrapper to *iptables.IPTables.
+func NewIptablesRunner(logf logger.Logf) (*iptablesRunner, error) {
```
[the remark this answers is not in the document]
KevinLiang10: Thanks for pointing this out. Tho I'm wondering would nftables be Nftables or NFTables since I'm not sure if nf is an abbreviation. Maybe it stands for netfilter?

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	return newIptablesRunnerAdvanced(ipt4, ipt6, supportsV6, supportsV6NAT), nil
+}
+
+// newIptablesRunnerAdvanced returns an iptablesRunner.
+// It is a wrapper for iptablesRunner init, which exists for testing purpose."
+func newIptablesRunnerAdvanced(ipt4, ipt6 iptablesInterface, supV6, supV6NAT bool) *iptablesRunner {
```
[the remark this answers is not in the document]
KevinLiang10: Sure. I was following the pattern we used to have in router_linux.go but I guess the point we did that with inner was for further configuring the linux router.

wgengine/router/router_linux.go
```diff
@@ -1285,7 +1148,7 @@ func (r *linuxRouter) addNetfilterBase() error {
 	if err := r.addNetfilterBase4(); err != nil {
 		return err
 	}
-	if r.v6Available {
+	if r.getV6Available() {
```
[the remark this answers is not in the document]
KevinLiang10: I guess I will just go with James's [suggestion](https://github.com/tailscale/tailscale/pull/8362#discussion_r1237608564)

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+	return errCode(err) == 1
+}
+
+// AddLoopbackRule adds a iptables rule to permit loopback traffic to
+// a local Tailscale IP.
+func (i *iptablesRunner) AddLoopbackRule(addr netip.Addr) error {
+	nf := i.ipt4
+	if addr.Is6() {
+		nf = i.ipt6
+	}
```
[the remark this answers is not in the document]
KevinLiang10: I guess we can have it, it's just used twice but I guess this change would make the code cleaner.

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,485 @@
...
+}
+
+// tsChain returns the name of the tailscale sub-chain corresponding
+// to the given "parent" chain (e.g. INPUT, FORWARD, ...).
+func tsChain(chain string) string {
+	return "ts-" + strings.ToLower(chain)
+}
+
+// getIptablebyProto returns an object you can use to manipulate IP tables for the given protocol.
+func (i *iptablesRunner) getIptablebyProto(proto ProtoVersion) iptablesInterface {
```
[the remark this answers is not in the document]
KevinLiang10: Yeah it seems like I lost the my reason to keep the type ProtoVersion as the version iterating

util/linuxfw/iptables_runner.go
KevinLiang10: Sounds good

util/linuxfw/linuxfw.go
KevinLiang10: This is used in nftables thing, I'll remove this from this pr.

util/linuxfw/linuxfw.go
KevinLiang10: This I what I took from router_linux, but it appears to me that it's no longer needed in linuxfw package. Also it seems like it's more reasonable to be in router_linux just looking at it's names and references, so I will just return this to router_linux.go. 

I'm not sure if there will be more goroutine accessing but I didn't really change the logic so I guess @bradfitz  was certain about what this meant. I can do a further investigation

wgengine/router/router_linux.go
KevinLiang10: Sure, I think being a method of netfilterRunner would already contain the information about where the hook is going to.

wgengine/router/router_linux.go
KevinLiang10: Yeah we totally can.

wgengine/router/router_linux.go
```diff
@@ -1239,43 +1127,18 @@ func (r *linuxRouter) delIPRulesWithIPCommand() error {
...
+func (r *linuxRouter) netfilterFamilies() []linuxfw.ProtoVersion {
+	if r.getV6Available() {
+		return []linuxfw.ProtoVersion{linuxfw.IPv4, linuxfw.IPv6}
 	}
-	return []netfilterRunner{r.ipt4}
+	return []linuxfw.ProtoVersion{linuxfw.IPv4}
 }
 
 // addNetfilterChains creates custom Tailscale chains in netfilter.
 func (r *linuxRouter) addNetfilterChains() error {
```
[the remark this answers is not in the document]
KevinLiang10: Will just remove the wrappers and call the method from nfr directly

wgengine/router/router_linux.go
KevinLiang10: Now we just call these methods from nfr without the wrapper

util/linuxfw/linuxfw.go
bradfitz: At least in the old code in router_linux.go that's a global map that's only written to during `init`, not at runtime.

util/linuxfw/iptable_runner.go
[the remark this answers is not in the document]
KevinLiang10: Now it's removed

util/linuxfw/__debug_bin
raggi, line 1: Make sure you remove this file, you should probably add it to your global .gitignore

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+	if i.HasIPV6() {
+		if err := i.addBase6(tunname); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// AddBase4 adds some basic IPv6 processing rules to be
```
raggi, line 226: match the case of the name with the method `addBase4`

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+	args = []string{"-o", tunname, "-j", "ACCEPT"}
+	if err := i.ipt4.Append("filter", "ts-forward", args...); err != nil {
+		return fmt.Errorf("adding %v in v4/filter/ts-forward: %w", args, err)
+	}
+
+	return nil
+}
+
+// AddBase6 adds some basic IPv4 processing rules to be
```
raggi, line 275:
```suggestion
// addBase6 adds some basic IPv4 processing rules to be
```

util/linuxfw/iptables_runner.go
```diff
@@ -0,0 +1,480 @@
...
+
+	if err := delChain(ipt, "nat", "ts-postrouting"); err != nil {
+		errs = append(errs, err)
+	}
+
+	if len(errs) > 0 {
+		return multierr.New(errs...)
+	}
+
+	return nil
```
raggi, line 478: this can just be `return multierr.New(errs...)` if it finds no errors in the slice, it returns nil.

raggi (commented):
Looking good, just a couple of nits and please remove that binary!

DentonGentry:
> please remove that binary

It will be important to squash the commits, otherwise the binary will go into git with the commit which added it, and then a subsequent commit removes it. Every `git clone` will pull a copy of that binary in the history.

raggi (approved):
Looking good, thank you Kevin!

catzkorn (approved):
Lets gooooo! 🎉

PR #8391, opened 2023-06-20
prober: allow monitoring of nodes marked as STUN only in default derpmap

valscale:
prober uses NewRegionClient() to connect to a derper using a faked up single-node region, but NewRegionClient() fails to connect if there is no non-STUN only client in the region. Set the STUN only flag to false before we call NewRegionClient() so we can monitor nodes marked as STUN only in the default derpmap.

Updates #11492

### DIFF
--- a/prober/derp.go
+++ b/prober/derp.go
@@ -348,6 +348,8 @@ func newConn(ctx context.Context, dm *tailcfg.DERPMap, n *tailcfg.DERPNode) (*de
 	priv := key.NewNode()
 	dc := derphttp.NewRegionClient(priv, l, nil /* no netMon */, func() *tailcfg.DERPRegion {
 		rid := n.RegionID
+		// Allow the prober to monitor nodes marked as STUN only in the default map
+		n.STUNOnly = false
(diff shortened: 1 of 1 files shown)


prober/derp.go
```diff
@@ -348,6 +348,8 @@ func newConn(ctx context.Context, dm *tailcfg.DERPMap, n *tailcfg.DERPNode) (*de
 	priv := key.NewNode()
 	dc := derphttp.NewRegionClient(priv, l, nil /* no netMon */, func() *tailcfg.DERPRegion {
 		rid := n.RegionID
+		// Allow the prober to monitor nodes marked as STUN only in the default map
+		n.STUNOnly = false
```
raggi, line 352: Do the stun only derps all respond to disco?

If they don't this may raise some alerts for probes failing with meshing.

valscale: Right now this code only runs on things in the superset derpmap, which only has one server marked STUN only. Other derps are monitored with different code.

raggi: 👍🏼

raggi (approved):

PR #8399, opened 2023-06-21
Revert "ssh/tailssh: change to user directory when running login/command"

bradfitz:
This reverts commit dc5bc32d8fbacd7e1dd34d138de542d401a11ebd.

It broke tests. (sadly, ones which we have disabled on CI, but go test ./ssh/tailssh broke)


maisem (approved):

bradfitz:
cc @derek-burdick

PR #8405, opened 2023-06-22
ssh/tailssh: fix double race condition with non-pty command

### LINKED ISSUE #7601: Tailscale SSH truncates streams at multiples of 32768
### What is the issue?

When transferring data over stdout on an SSH connection using Tailscale SSH, the stream can be incorrectly truncated.

### Steps to reproduce

On a server:
```
[root@leon:~]# cat /dev/random | head -c 4000000 > /run/rand

[root@leon:~]# wc -c /run/rand
4000000 /run/rand

[root@leon:~]# sha256sum rand
0394c6312d9ea5ebb10db46b190616fd67a27f153d1cd6f42faa5a9f13f00e75  rand

[root@leon:~]# head -c 3964928 rand > rand-trunc

[root@leon:~]# sha256sum rand-trunc
6d0b715d29c8067cd3d038fb0b3d553f87c804a34c8430de804e5f892e13ac94  rand-trunc
```

On a client:
```
[randomcat@groves:~]$ ssh -vvv root@leon 'cat /root/rand' > rand 2> ssh_out.txt; wc -c rand
3964928 rand

[randomcat@groves:~]$ sha256sum rand
6d0b715d29c8067cd3d038fb0b3d553f87c804a34c8430de804e5f892e13ac94  rand
```

[ssh_out.txt](https://github.com/tailscale/tailscale/files/11010469/ssh_out.txt)

The exact number of bytes that are properly transferred varies from run to run, but (at least with this file size) is always a multiple of 32768.

I am unable to reproduce this using normal SSH over the pubilc internet or by using an SSH (but not Tailscale SSH) connection with a normal keypair over Tailscale.

_No response_

### OS

Linux

### OS version

NixOS 23.05.20230312.5a1dc8a (Stoat)

### Tailscale version

1.36.2

### Other software

None that should be relevant.

### Bug report

BUG-a5a749adeecdbeebd3da59f158fcdf69ec0fde2f96161a43f4b57db26d1094c4-20230319075559Z-498daae6a7fdea28

dsnet:
There are two race conditions in output handling.

The first race condition is due to a misuse of exec.Cmd.StdoutPipe. The documentation explicitly forbids concurrent use of StdoutPipe with exec.Cmd.Wait (see golang/go#60908) because Wait will close both sides of the pipe once the process ends without any guarantees that all data has been read from the pipe. To fix this, we allocate the os.Pipes ourselves and manage cleanup ourselves when the process has ended.

The second race condition is because sshSession.run waits upon exec.Cmd to finish and then immediately proceeds to call ss.Exit, which will close all output streams going to the SSH client. This may interrupt any asynchronous io.Copy still copying data. To fix this, we close the write-side of the os.Pipes after the process has finished (and before calling ss.Exit) and synchronously wait for the io.Copy routines to finish.

Fixes #7601

### DIFF
--- a/ssh/tailssh/tailssh.go
+++ b/ssh/tailssh/tailssh.go
@@ -823,12 +823,16 @@ type sshSession struct {
 	agentListener net.Listener // non-nil if agent-forwarding requested+allowed
 
 	// initialized by launchProcess:
-	cmd    *exec.Cmd
-	stdin  io.WriteCloser
-	stdout io.ReadCloser
-	stderr io.Reader // nil for pty sessions
-	ptyReq *ssh.Pty  // non-nil for pty sessions
-	tty    *os.File  // non-nil for pty sessions, must be closed after process exits
+	cmd      *exec.Cmd
+	wrStdin  io.WriteCloser
+	rdStdout io.ReadCloser
+	rdStderr io.ReadCloser // rdStderr is nil for pty sessions
+	ptyReq   *ssh.Pty      // non-nil for pty sessions
+
+	// childPipes is a list of pipes that need to be closed when the process exits.
+	// For pty sessions, this is the tty fd.
+	// For non-pty sessions, this is the stdin, stdout, stderr fds.
+	childPipes []io.Closer
 
 	// We use this sync.Once to ensure that we only terminate the process once,
 	// either it exits itself or is terminated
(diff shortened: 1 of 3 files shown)


ssh/tailssh/tailssh.go
```diff
@@ -1131,32 +1131,41 @@ func (ss *sshSession) run() {
...
+
 	// This will either make the SSH Termination goroutine be a no-op,
 	// or itself will be a no-op because the process was killed by the
 	// aforementioned goroutine.
 	ss.exitOnce.Do(func() {})
 
+	// Close the process-side of all pipes to signal the asynchronous
+	// io.Copy routines reading/writing from the pipes to terminate.
+	// Block for the io.Copy to finish before calling ss.Exit below.
+	closeAll(ss.rdStdin, ss.wrStdout, ss.wrStderr, ss.tty)
```
maisem, line 1163: Create a slice `childFDs` and initialize closer to where these are created so that it's easier to reason about here too?

maisem (approved):
Thanks!

ssh/tailssh/tailssh.go
dsnet: Done.

maisem (approved):

Issue #8413, opened 2023-06-22
Add platform-independent abstraction for fetching TCP information


andrew-d:
So that we can fetch TCP RTT, MSS, etc. in a platform-independent way instead of e.g. [the platform-specific code in `derp/derp_server_linux.go`](https://github.com/tailscale/tailscale/blob/main/derp/derp_server_linux.go).

PR #8414, opened 2023-06-22
net/tcpinfo: add package to allow fetching TCP information

andrew-d:
This package contains platform-independent abstractions for fetching information about an open TCP connection.

Updates https://github.com/tailscale/tailscale/issues/8413


net/tcpinfo/tcpinfo.go
```diff
@@ -0,0 +1,51 @@
...
+
+// netConner is implemented by crypto/tls.Conn to unwrap into an underlying
+// net.Conn.
+type netConner interface {
+	NetConn() net.Conn
+}
+
+// unwrap attempts to unwrap a net.Conn into an underlying *net.TCPConn
+func unwrap(nc net.Conn) (*net.TCPConn, error) {
+	for {
```
valscale, line 41: Nit: Maybe I am not go-y enough, but this would definitely be easier for me to read as something like:

if nc.(type) == netConner {
    nc = v.NetConn()
}

Then the rest of the switch (or else another if statement). The initial "for {" kind of visually disappears into the function declaration.

valscale (approved):
Nice work!

net/tcpinfo/tcpinfo.go
andrew-d: This is intentionally a loop, in case we want to unwrap multiple layers of `netConner`. If we have `&wrapper{inner: &wrappper{inner: *net.TCPConn}}`, then the `for` loop will correctly unwrap to the inner `*net.TCPConn`, whereas a single `if v, ok := nc.(netConner); ok {` would not.

valscale: Oh no! Multiple layers didn't even occur to me. :) Thanks for explaining to me!

Issue #7946, opened 2023-04-22
FR: Add Control D as a DNS option


Glenta:
### What are you trying to do?

Incorporate Control D as an available DNS option.

### How should we solve this?

Similar to the inclusion of NextDNS, I propose the addition of Control D as an alternative option.

### What is the impact of not solving this?

Users are limited in their ability to customize Tailscale, as certain features, such as using Control D with the platform, require workarounds. For example, users must manually input four DNS addresses in the Tailscale console and register their IP address with Control D to authenticate themselves. This can be particularly onerous for users with dynamic IP addresses, which are common in mobile networks. Although secure DNS options are available that do not require this workaround, Tailscale does not currently support features such as DNS over HTTPS/TLS/DoQ.

### Anything else?

[Control D's API documentation](https://docs.controld.com/reference/get-started)
[Why you need to register your IP with Control D for the service to work](https://docs.controld.com/docs/ip-not-authorized)

trollybusman:
+1 I would love this as I recently switched from NextDNS and prefer Control D.

yegors:
One of the devs here, would love to get this integration done.

Aureyl:
Would love to Tailscale ;-)

Ginder-Singh:
It would be great to see Control D as option.

bradfitz:
Please use GitHub reactions instead of duplicate comments unless you have something unique to say.

bradfitz:
@yegors, here's what we'd want: a way to send an HTTP header or URL parameter to your DoH endpoint to specify the profile ID to use.

And your DoH endpoint should have a well-known anycast IP (v4 and v6) and not require bootstrap DNS, ideally.

yegors:
Hey @bradfitz, 

The DNS-over-HTTPS endpoint is indeed anycast and supports IPv4/IPv6 is: https://dns.controld.com/xxxxxxxxx

- All 3 versions of HTTP are supported, including H3 / QUIC
- `xxxxxxxxx` is where the Resolver ID goes, which is unique to every "Device": https://docs.controld.com/docs/devices

Control D is different from NextDNS in a sense that Resolvers (Devices) enforce Policies (Profiles), as opposed to being "Profiles" themselves. This allows for a Profile to be swapped (or rotated on a schedule) on an existing Device. 

Additionally, for a more useful integration the following DOH headers can be supplied along with every DNS query. 
- `x-cd-mac` - MAC address of the client
[...]

PR #8417, opened 2023-06-23
Add Control D support, along with DoH client metadata

alexelisenko:
This adds support for Control D in a similar fashion to how NextDNS was integrated via [this merge](https://github.com/tailscale/tailscale/pull/5556).

Resolves: https://github.com/tailscale/tailscale/issues/7946

Control D allows for many advanced use cases, and many customers of both Control D and Tailscale have requested this integration.

This also adds support for providing the client hostname, mac address and tailscale local IP to the DNS server to facilitate client specific DNS configurations and data analysis via the Control D analytics feature. While I understand that adding Control D specific fields to the forwarder may not be ideal, I do think that these fields may become useful for other providers over time. If this is something that the tailscale team is very opposed to, I can remove this as it's not a requirement for integration, but does add value for clients that want to use a single DoH endpoint for multiple devices and manage them under a single scope. 

@bradfitz Please take a look 🙏


net/dns/resolver/forwarder.go
```diff
@@ -419,6 +448,20 @@ func (f *forwarder) sendDoH(ctx context.Context, urlBase string, c *http.Client,
...
+	// this allows for additional DNS analytics and client specific DNS policies
+	if strings.HasPrefix(urlBase, publicdns.ControlDBase) {
+		if f.dohClientHostname != "" {
+			req.Header.Set("X-Cd-Host", f.dohClientHostname)
+		}
+		if f.dohClientMac != "" {
+			req.Header.Set("X-Cd-Mac", f.dohClientMac)
+		}
+		if f.dohClientIP != "" {
+			req.Header.Set("X-Cd-IP", f.dohClientIP)
```
bradfitz, line 461: lowercase "p" in "Ip" (otherwise Set will just do it for you, allocating)

net/dns/resolver/forwarder.go
```diff
@@ -200,18 +202,45 @@ type forwarder struct {
...
+		tailscaleIP = tsIPs[0].String()
+	}
+
+	// TODO: is this the best way to determine the MAC address of this node?
+	if idx, err := interfaces.DefaultRouteInterfaceIndex(); err == nil {
+		iface, err := net.InterfaceByIndex(idx)
+		if err == nil {
+			clientMac = iface.HardwareAddr.String()
+		}
+	}
```
bradfitz, line 233: Remove all this. We'd push these parameters to the node from the control plane, like we do with NextDNS.

net/dns/publicdns/publicdns.go
```diff
@@ -220,9 +279,22 @@ func nextDNSv6Gen(ip netip.Addr, id []byte) netip.Addr {
 	return netip.AddrFrom16(a)
 }
 
+// controlDv6Gen generates a Control D IPv6 address from provided ip and id.
```
bradfitz, line 282: comment on what "id" is more?

net/dns/publicdns/publicdns.go
```diff
@@ -185,6 +206,37 @@ func populate() {
...
+	addDoH("2606:1a40:1::2", "https://freedns.controld.com/p2")
+
+	// Control D -Malware + Ads + Social
+	addDoH("76.76.2.3", "https://freedns.controld.com/p3")
+	addDoH("76.76.10.3", "https://freedns.controld.com/p3")
+	addDoH("2606:1a40::3", "https://freedns.controld.com/p3")
+	addDoH("2606:1a40:1::3", "https://freedns.controld.com/p3")
+
+	// Control D -Malware + Ads + Adult
+	addDoH("76.76.2.4", "https://freedns.controld.com/family")
```
bradfitz, line 235: `freedns.controld.com` just resolved to `76.76.2.11` for me, not `76.76.2.4` as you have listed here.

Which means these entries might not work. (unless you also support all these DoH SNI/HTTP host names to the IPs you've listed)

See the commented out OpenDNS code & comment above.

net/dns/publicdns/publicdns.go
```diff
@@ -23,6 +26,9 @@ var dohOfIP = map[netip.Addr]string{} // 8.8.8.8 => "https://..."
 var dohIPsOfBase = map[string][]netip.Addr{}
 var populateOnce sync.Once
 
+const NextDNSBase = "https://dns.nextdns.io/"
+const ControlDBase = "https://dns.controld.com/"
```
bradfitz, line 30: if these need to be exported, they need comments.

bradfitz (commented):
Please adjust the commit message too to be in our conventional format (look at git log for examples... it's mostly the same as https://github.com/golang/go/wiki/CommitMessage)

And for the bug reference line uses `Updates #nnnn` on its own line (We only use verbs "Updates" and "Fixes", but this won't be the only PR for the issue, so Updates is most appropriate)

bradfitz:
@alexelisenko, @yegors, ping. See comments/questions above. I just saw this again when cleaning up PRs and noticed it had stallled.

net/dns/publicdns/publicdns.go
alexelisenko: So the DoH endpoint listed there work on the IPs provided, as well as on the IPs the hostnames resolve to, This is because those special resolvers, like "family" or "Malware +ads" can be reached via the DoH endpoint or via legacy DNS. The ips provided are the legacy DNS IPs.

net/dns/resolver/forwarder.go
alexelisenko: Can you please link me to where this control plane is doing this for NextDNS?

Also, we currently only support passing full metadata (local IP, MAC and hostname) about the client via DoH headers, but we do support adding just the hostname via the DoH URI, i.e:

https://dns.controld.com/<resolver_id>/<hostname>

Would DoH headers with full metadata be completely out of the question? If so, would love to lear n more about what we can do with the control plane.

net/dns/resolver/forwarder.go
alexelisenko: fixed

net/dns/publicdns/publicdns.go
alexelisenko: updated comment

net/dns/publicdns/publicdns.go
alexelisenko: added comments

sebastian-tischer:
Hey team, is there any update on this? I would love to see the integration

tmchow:
What’s happening with this? Stalled since October?

Coder84619:
I'm a new Control D user (loving the service) and really need Tailscale integration.

devowloper:
At the risk aggravating you, @bradfitz can you cast your eye on this PR, please? I see you're on parental leave, yet I've spotted you've approved a few PRs (though you should enjoy your time off!). Is this held up due to the commit message check failing, or something else?

bradfitz:
This PR is lacking DCO (`git commit -s`) but even if that's fixed and the rest of the PR is fine, there are server-side bits that would also need to be done & prioritized before this PR would be usable.

/cc @kabirsikand to think about that.

yegors:
@bradfitz / @kabirsikand  Hate to be annoying but any chance to get this moving along anytime soon? Anything we can do on our end?

The associated issue is now 3rd most upvoted one here.

bradfitz:
I could merge at least this client-side portion of things but what's going on with this PR's commits? It has 744 commits in it. Looks like the rebase went poorly. There's a merge commit but then also 744 commits from main that are all re-signed by a `Signed-off-by: Alex Paguis <alex@windscribe.com>` (is that the same Alex as @alexelisenko?).

To answer the questions in the TODOs in the code: that's not how we should get the hostname, Tailscale IP, and MAC. Let's omit that part for now. I assume it's optional. We can add it back in a subsequent commit, plumbed the right way.

bradfitz:
BTW, `ip.String() == controlDv4One.String() || ip.String() == controlDv4Two.String()` is an inefficient (+ allocating) way to compare to netip.Addrs. You can just `==` on them directly.

bradfitz:
The newly added tests also fail?

```
--- FAIL: TestDoHIPsOfBase (0.00s)
    publicdns_test.go:141: DoHIPsOfBase("https://dns.controld.com/hyq3ipr2ct") = [76.76.2.22 76.76.10.22 2a07:a8c0:0:6:7b5b:5949:35ad:0 2a07:a8c1:0:6:7b5b:5949:35ad:0]; want [76.76.2.22 76.76.10.22 2606:1a40:0:6:7b5b:5949:35ad:0 2606:1a40:1:6:7b5b:5949:35ad:0]
    publicdns_test.go:141: DoHIPsOfBase("https://dns.controld.com/112233445566778899aabbcc") = [76.76.2.22 76.76.10.22 2a07:a8c0:0:ffff:ffff:ffff:ffff:0 2a07:a8c1:0:ffff:ffff:ffff:ffff:0]; want [76.76.2.22 76.76.10.22 2606:1a40:0:ffff:ffff:ffff:ffff:0 2606:1a40:1:ffff:ffff:ffff:ffff:0]
FAIL
FAIL    tailscale.com/net/dns/publicdns 0.124s
FAIL
```

bradfitz:
You didn't check that box that maintainers could make changes, so I ended up forking this off into #11659 to fix the git mistakes and commit message and remove the extra headers for now.

bradfitz:
This is now mostly merged (except the parts I removed). If you want to send followup changes, please check the git result first.

Issue #8434, opened 2023-06-25
cmd/derper: serve STUN out of separate process


crawshaw:
Ideally, on a CPU-bound DERP, CPU priority would go to serving STUN packets.

PR #8435, opened 2023-06-25
cmd/derper: serve STUN responses out of a separate process

crawshaw:
Updates #8434

Two things to consider before merging this PR:
- [ ] the effect on monitoring. I believe monitoring derper CPU will now only report the STUN CPU use. What is the best way to tell our monitoring about the subprocess? (Can monitoring monitor a process group?)
- [ ] now that STUN is higher CPU priority it may be worth adding a per-remoteaddr rate limiter

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -19,6 +19,7 @@ import (
 	"net/http"
 	"net/netip"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


bradfitz:
I thought the plan was to run STUN on its own server in a region?

crawshaw:
I'm not up to date on plans (if there was a meeting, I haven't found the notes to read). This seemed really easy until I was done and got to thinking about monitoring.

bradfitz:
cc @valscale @raggi

cmd/derper/derper.go
```diff
@@ -317,14 +325,45 @@ func main() {
...
+// This means that STUN packets get answered by a separate, higher-priority
+// process.
+func runNiceDERPChild() {
+	// Build an arg list, filtering out the --stun flag.
+	var args []string
+	flag.Visit(func(f *flag.Flag) {
+		if f.Name == "stun" {
+			return
+		}
+		args = append(args, "-"+f.Name, f.Value.String())
```
creachadair, line 348: If we have (or add) any Boolean flags, I think we will want `-name=value` format as the blanks will not be read properly and will terminate option parsing.

PR #8439, opened 2023-06-25
syncs: add ShardedMap type

bradfitz:
Updates tailscale/corp#7354

### DIFF
--- a/syncs/shardedmap.go
+++ b/syncs/shardedmap.go
@@ -1,0 +1,111 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package syncs
+
+import (
+	"sync"
+
+	"golang.org/x/sys/cpu"
+)
+
+// ShardedMap is a synchronized map[K]V, internally sharded by a user-defined
+// K-sharding function.
+//
+// The zero value is not safe for use; use NewShardedMap.
+type ShardedMap[K comparable, V any] struct {
+	shardFunc func(K) int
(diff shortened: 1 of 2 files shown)


syncs/shardedmap.go
```diff
@@ -0,0 +1,108 @@
...
+// The zero value is not safe for use; use NewShardedMap.
+type ShardedMap[K comparable, V any] struct {
+	shardFunc func(K) int
+	shards    []mapShard[K, V]
+}
+
+type mapShard[K comparable, V any] struct {
+	mu sync.Mutex
+	m  map[K]V
+	_  cpu.CacheLinePad // avoid false sharing of neighboring shards' mutexes
```
maisem, line 24: Link to docs or some other reference?

syncs/shardedmap.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// NewShardedMap returns a new ShardedMap with the given number of shards
+// and sharding function.
```
maisem, line 28: The sharding function should have more docs I think. Like be strict about what values it can return and panic with a better message if it returns outside of 0-(N-1)

maisem (approved):

syncs/shardedmap.go
bradfitz: that type itself has docs. click it in VS code :)

syncs/shardedmap.go
bradfitz: oh, it'll panic.

bradfitz: added docs.

PR #8440, opened 2023-06-25
.github: actually run tests in CI

maisem:
Updates #cleanup


bradfitz (approved):

util/deephash/deephash_test.go
```diff
@@ -139,7 +139,7 @@ func TestHash(t *testing.T) {
 		{in: tuple{netip.AddrPortFrom(netip.AddrFrom4([4]byte{192, 168, 0, 1}), 1234), netip.AddrPortFrom(netip.AddrFrom4([4]byte{192, 168, 0, 1}), 1235)}, wantEq: false},
 		{in: tuple{netip.AddrPortFrom(netip.AddrFrom4([4]byte{192, 168, 0, 1}), 1234), netip.AddrPortFrom(netip.AddrFrom4([4]byte{192, 168, 0, 2}), 1234)}, wantEq: false},
 		{in: tuple{netip.Prefix{}, netip.Prefix{}}, wantEq: true},
-		{in: tuple{netip.Prefix{}, netip.PrefixFrom(netip.Addr{}, 1)}, wantEq: false},
+		{in: tuple{netip.Prefix{}, netip.PrefixFrom(netip.Addr{}, 1)}, wantEq: true},
```
maisem, line 142: New behavior in go1.21
cc @dsnet @bradfitz

maisem: https://cs.opensource.google/go/go/+/3693fd255f44acc24efa2b45317473d4d1e1abde

bradfitz: this needs to pass with both Go 1.20 and Go 1.21 for a few months. So maybe just remove this test case for now. It's not really material to the deephash package.

maisem:
The regression was introduced in 43819309e18ac7d496643fcd86489a2ee0927964

bradfitz (approved):

util/deephash/deephash_test.go
dsnet: Agreed with removing this case. I made the upstream change to have one consistent invalid representation.

PR #8443, opened 2023-06-26
types/key: add test for NodePublic.Shard

bradfitz:
Updates #cleanup

### DIFF
--- a/types/key/node_test.go
+++ b/types/key/node_test.go
@@ -157,3 +157,26 @@ func TestChallenge(t *testing.T) {
 		t.Errorf("didn't round trip: %v != %v", back, pub)
 	}
 }
+
+// Test that NodePublic.Shard is uniformly distributed.
+func TestShard(t *testing.T) {
+	const N = 1_000
+	var shardCount [256]int
+	for i := 0; i < N; i++ {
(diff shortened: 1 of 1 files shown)


bradfitz:
@twitchyliquid64, how did you come up with:

```go
	s := uint8(p.k[31]) + uint8(p.k[30]) + uint8(p.k[20])
	return s ^ uint8(p.k[2]+p.k[12])
```

Because just `return p.k[0]` also passes this chi-squared test. (but not, say, `return p.k[31]`)

twitchyliquid64:
I yeeted it to mix in bits from not the start and end of the key.

I know the key isnt uniformly random, so I wanted to grab more entropy from a random dabble of other bits.

This was for when we were experimenting with sharded locks, I don't think we ended up using this. So maybe we can delete this method entirely?

bradfitz:
We are using it now. Hence my new interest in it. :)

twitchyliquid64:
Ah gotcha. Well to answer your question there was no grand design to which bytes I am mashing with which other bytes - just a general desire to mix stuff from across the array because entropy is not uniform.

twitchyliquid64 (approved):

PR #8455, opened 2023-06-27
net/netutil: parse IP forwarding val as int, not bool

### LINKED ISSUE #8375: net.ipv6.conf.all.forwarding=2 does not parse correctly leading to inability to accept routes
### What is the issue?

`net.ipv6.conf.all.forwarding=2` in `sysctl` causes `tailscale up` command to error out due to Bool parsing of non bool number.v

### Likely Culprit
https://github.com/tailscale/tailscale/blob/13022952993c1d662dfed5326951fe5e15b0fc26/net/netutil/ip_forward.go#L192-L220

```
	on, err := strconv.ParseBool(string(bytes.TrimSpace(bs)))
```

Have double-checked and the likely culprit `ParseBool` is still on `main` so this issue might affect the `main` branch still.

I can see that documentation refers to this entry as a boolean so I can see why it might be tempting to parse it as one:

https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt

```
forwarding - BOOLEAN
	Enable IP forwarding on this interface.  This controls whether packets
	received _on_ this interface can be forwarded.
```

However I also dug out the (hopefully correct) source for the forwarding var:
https://github.com/torvalds/linux/blob/45a3e24f65e90a047bef86f927ebdc4c710edaa1/include/linux/ipv6.h#L13

Here it is in the wild:
https://github.com/search?q=repo%3Atorvalds%2Flinux%20%20-%3Eforwarding&type=code

It seems to be recorded as an integer, rather than a bool. So despite documentation suggesting that it is supposed to be a bool nothing precludes it from being set to any integer value.

I have no clear idea what setting forwarding to 2 actually does outside of this blog post:
> What's this special value "2"? Originally the value was "1", but this disabled autoconfiguration on all interfaces. That is, you couldn't appear to be a router on some interfaces and appear to be a host on other interfaces. But that's exactly the mental model of a ADSL router. (This is the trouble you make when you ignore the UNIX tenet that the system administrator knows what they are doing.)
> — https://vk5tu.livejournal.com/37206.html

If you google "net.ipv6.conf.all.forwarding=2" there are certainly quite a few results popping up, with very few explaining the reasoning behind it.

I also asked ChatGPT since googling did not get me very far... 

> By default, the value of "net.ipv6.conf.all.forwarding" is usually set to "0" or "1". A value of "0" means forwarding is disabled, while "1" means forwarding is enabled. Setting it to "2" explicitly enables forwarding and allows the system to act as a router for IPv6 traffic.

Not sure where this is all coming from since this is absent in Kernel Documentation. Perhaps that is some deprecated behavior, or folks have just been using the `2` value because it seemed to work for others before them.

### OS Version OpenWRT Details:

DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='21.02-SNAPSHOT'
DISTRIB_REVISION='r16399+159-c67509efd7'
DISTRIB_TARGET='ipq807x/ipq60xx'
DISTRIB_ARCH='arm_cortex-a7'
DISTRIB_DESCRIPTION='OpenWrt 21.02-SNAPSHOT r16399+159-c67509efd7'
DISTRIB_TAINTS='no-all busybox override'
DISTRIB_TIP='OpenWrt 21.02-SNAPSHOT r16399+159-c67509efd7 / TIP-devel-cd69974d'
DISTRIB_TIP_VERSION='devel'

### Steps to reproduce

```
# Set up ipv6 config (copied from my router)
$ cat <<EOS > /etc/sysctl.d/99-ipv6.conf
root@NLWarp:~# cat /etc/sysctl.d/99-disable-ipv6.conf
net.ipv6.conf.all.disable_ipv6=0
net.ipv6.conf.lo.disable_ipv6=0
net.ipv6.conf.default.disable_ipv6=0
net.ipv6.conf.default.forwarding=2
net.ipv6.conf.all.forwarding=2
net.ipv6.conf.all.accept_ra=2
net.ipv6.conf.bond0.accept_ra=2
net.ipv6.conf.br-guest.accept_ra=2
net.ipv6.conf.br-lan.accept_ra=2
net.ipv6.conf.br-lan.5.accept_ra=2
net.ipv6.conf.default.accept_ra=2
net.ipv6.conf.eth0.accept_ra=2
net.ipv6.conf.eth1.accept_ra=2
net.ipv6.conf.eth2.accept_ra=2
net.ipv6.conf.lo.accept_ra=2
net.ipv6.conf.tailscale0.accept_ra=2
net.ipv6.conf.wlan0.accept_ra=2
EOS

# Run tailscale up
$ tailscale up --accept-routes --advertise-routes=<routes>
Warning: Couldn't check system's IP forwarding configuration, subnet routing/exit nodes may not work: couldn't parse net/ipv6/conf/all/forwarding (strconv.ParseBool: parsing "2": invalid syntax)
See https://tailscale.com/kb/1104/enable-ip-forwarding/
```

N/A

### OS

Linux

### OS version

OpenWRT 21.02-SNAPSHOT

### Tailscale version

1.32.2-dev; go version: go1.19.3

### Other software

_No response_

### Bug report

BUG-0fc74ff8b2436071663900a36ba7368d5003e6927c2fdcccc4ab261d509cfd96-20230619001353Z-c4aad93674920261

rosszurowski:
This commit updates our IP forwarding parsing logic to allow the less common but still valid value of `2` to be parsed as `true`, which fixes an error some users encountered.

Fixes #8375

### DIFF
--- a/net/netutil/ip_forward.go
+++ b/net/netutil/ip_forward.go
@@ -212,9 +212,16 @@ func ipForwardingEnabledLinux(p protocol, iface string) (bool, error) {
 		}
 		return false, err
 	}
-	on, err := strconv.ParseBool(string(bytes.TrimSpace(bs)))
+
+	val, err := strconv.ParseInt(string(bytes.TrimSpace(bs)), 10, 32)
 	if err != nil {
 		return false, fmt.Errorf("couldn't parse %s: %w", k, err)
 	}
+	// 0 = disabled, 1 = enabled, 2 = enabled (but uncommon)
+	// https://github.com/tailscale/tailscale/issues/8375
+	if val < 0 || val > 2 {
+		return false, fmt.Errorf("unexpected value %d for %s", val, k)
+	}
+	on := val == 1 || val == 2
 	return on, nil
 }


net/netutil/ip_forward.go
```diff
@@ -212,9 +212,11 @@ func ipForwardingEnabledLinux(p protocol, iface string) (bool, error) {
 		}
 		return false, err
 	}
-	on, err := strconv.ParseBool(string(bytes.TrimSpace(bs)))
+
+	val, err := strconv.ParseInt(string(bytes.TrimSpace(bs)), 10, 32)
 	if err != nil {
 		return false, fmt.Errorf("couldn't parse %s: %w", k, err)
 	}
+	on := val == 1 || val == 2
```
bradfitz, line 220: should you return an error on negative values or val greater than 2?

rosszurowski: Updated, PTAL!

bradfitz (approved):

PR #8460, opened 2023-06-27
words: scaly fruit salad? go fly a kite

thisisparker:
Some tails, some scales.

This is my first commit in this package so please let me know if I'm doing something wrong!

### DIFF
--- a/words/scales.txt
+++ b/words/scales.txt
@@ -388,3 +388,4 @@ becrux
 godzilla
 sirius
 vector
+cherimoya


bradfitz:
The commits should start with "words: "

catzkorn:
We already have salak! :D

thisisparker:
> We already have salak! :D

so we do! I saw it referenced in an article about cherimoya and I guess I was too excited to check?!

clairew:
squash to 1 commit that says something like "add fruit to scales add kite to tails"

catzkorn (approved):

PR #8464, opened 2023-06-27
tstest, tstime: mockable timers and tickers

### LINKED ISSUE #8463: FR: Test tooling for timers and tickers

I'm trying to test some code that uses timers and tickers but the standard library has no way to fake them for use in tests. `tstest.Clock` plays this sort of role for `time.Now()` but not others.

### How should we solve this?

Extend `tstest.Clock` so that it can manage timers and tickers.

https://github.com/benbjohnson/clock does approximately what I'm looking for but it is now unmaintained and I'm reluctant to bring in an unmaintained 3rd party dependency.

It is much harder or impossible to test things that rely on timers and tickers (e.g. keepalive code).

### Anything else?

_No response_

sailorfrag:
This change introduces tstime.Clock which is the start of a mockable interface for use with testing other upcoming code changes.

Fixes #8463

cc @bradfitz and @maisem in case you have opinions

This should not be merged until a corresponding update to `corp` is ready to go because it changes how a `tstest.Clock` is created.

### DIFF
--- a/tstest/clock.go
+++ b/tstest/clock.go
@@ -4,8 +4,12 @@
 package tstest
 
 import (
+	"container/heap"
 	"sync"
 	"time"
+
+	"tailscale.com/tstime"
+	"tailscale.com/util/mak"
 )
 
 // ClockOpts is used to configure the initial settings for a Clock. Once the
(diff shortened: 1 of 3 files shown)


tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+}
+
+func NewClockBuilder() *ClockBuilder {
+	cb := &ClockBuilder{
+		c: new(Clock),
+	}
+	return cb
+}
+
+func (cb *ClockBuilder) Start(t time.Time) *ClockBuilder {
```
danderson, line 29: Docstrings please, for all methods of ClockBuilder

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+
+func (cb *ClockBuilder) Finish() *Clock {
+	c := cb.c
+	cb.c = nil
+	c.init()
+	return c
+}
+
+// NewClock returns a Clock initialized with default settings (equivalent to
+// NewClockBuilder().Finish()).
```
danderson, line 52: Make the implementation literally NewClockBuilder().Finish() ? That way if we ever change defaults we don't need to remember to update two places

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+	step time.Duration
+	// present is the last value returned by Now (and will be returned again by
+	// ConstantNow).
+	present time.Time
+	// skipStep indicates that the next call to Now should not add step to
+	// present. This occurs after initialization and after Advance.
+	skipStep bool
+	// timerChannelSize is the buffer size to use for channels created by
+	// NewTimer and NewTicker.
+	timerChannelSize int
```
danderson, line 80: Is it important for the caller to care about this? Since this is for tests only, any downside to making the channel absurdly large (e.g. 4k slots) ? Or is it important that callers can set up loss of notifications?

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+	defer c.mu.Unlock()
+	if !c.skipStep && c.step > 0 {
+		c.present = c.present.Add(c.step)
+		c.events.AdvanceTo(c.present)
+	}
+	c.skipStep = false
+	return c.present
+}
+
+// ConstantNow returns the virtual clock's current time but does not advance it
```
danderson, line 110: bikeshed: maybe PeekNow() ? But maybe that's my unsavory network upbringing speaking where "peek" is a thing that makes sense.

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+// if a step is configured.
+func (c *Clock) ConstantNow() time.Time {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.present
+}
+
+func (c *Clock) Advance(d time.Duration) time.Time {
```
danderson, line 118: Docstrings for all exported methods pls

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+
+	if d > 0 {
+		c.present = c.present.Add(d)
+		c.events.AdvanceTo(c.present)
+	}
+
+	return c.present
+}
+
+func (c *Clock) Start() time.Time {
```
danderson, line 131: Hmm, Start() sounds like a command to start the clock running, rather than an accessor. Maybe StartTime() ? Or maybe with a docstring and then used in context it doesn't matter. Not a strong opinion.

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.start
+}
+
+func (c *Clock) Step() time.Duration {
```
danderson, line 137: Same, maybe StepTime?

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+	defer c.mu.Unlock()
+
+	t := &Ticker{
+		nextTrigger: c.present.Add(d),
+		period:      d,
+		em:          &c.events,
 	}
-	if c.Present.Before(c.Start) {
-		c.Present = c.Start
+	t.init(c.timerChannelSize)
```
danderson, line 166: hmm, so that's what timerChannelSize is for. Isn't this channel always size 1 in the real ticker/timer implementations? With specified semantics for what happens if events occur faster than they're processed?

I lack context here, so maybe I'm way off base. At a glance, this feels like it shouldn't be configurable, because it lets you create tickers and timers with subtly different semantics to the production ones?

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+
+	t := &Timer{
+		nextTrigger: c.present.Add(d),
+		em:          &c.events,
 	}
+	t.init(c.timerChannelSize)
+	return t, t.C
+}
+
+type eventHandler interface {
```
danderson, line 184: docstrings for important unexported structs and interfaces. I loosely understand what eventHandler and event are doing here, but some additional docs (even if a bit boilerplatey) would be welcome to reduce uncertainty.

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+	Fire(time.Time) time.Time
 }
 
-// Reset rewinds the virtual clock to its start time.
-func (c *Clock) Reset() {
-	c.Lock()
-	defer c.Unlock()
-	c.Present = c.Start
+type event struct {
+	position int       // The current index in the heap.
```
danderson, line 193: It took me a bit to figure out why we track position here. A few more words on the comment would be nice.

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+// The next trigger time for the ticker is updated to the last computed trigger
+// time + the ticker period (set at creation or using Reset). The next trigger
+// time is computed this way to match standard time.Ticker behavior, which
+// prevents accumulation of long term drift caused by delays in event execution.
+func (t *Ticker) Fire(curTime time.Time) time.Time {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	if t.nextTrigger.IsZero() {
+		return t.nextTrigger
```
danderson, line 354: for parity with Timer.Fire below, I'd make this return time.time{}, which makes it more explicit at a glance that you're returning a zero time.

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+func (em *eventManager) Now() time.Time {
+	em.mu.Lock()
+	defer em.mu.Unlock()
+	return em.now
+}
+
+// Ticker is a time.Ticker lookalike for use in tests that need to control when
+// events fire. Ticker could be made standalone in future but for now is
+// expected to be paired with a Clock and created by Clock.NewTicker.
+type Ticker struct {
```
danderson, line 312: Sort of general comment: do we need a time.AfterFunc equivalent as well? Easy to assemble from a Timer and some glue, just curious.

tstime/tstime.go
```diff
@@ -59,3 +59,35 @@ func Sleep(ctx context.Context, d time.Duration) bool {
 		return true
 	}
 }
+
+type Clock interface {
```
danderson, line 63: Docstrings throughout please, on: interfaces, interface methods, structs and struct methods.

danderson (approved):
Looks reasonable to me. A few comments here and there, the main question I have is around the timer/ticker channels, the rest is cosmetic.

tstime/tstime.go
```diff
@@ -59,3 +59,35 @@ func Sleep(ctx context.Context, d time.Duration) bool {
...
+
+type TickerController interface {
+	Reset(d time.Duration)
+	Stop()
+}
+
+type TimerController interface {
+	Reset(d time.Duration) bool
+	Stop() bool
+}
```
bradfitz, line 77: why separate types? call this "ResetStopper" (like io.ReadWriter, io.ReadSeeker, etc)?

tstime/tstime.go
```diff
@@ -59,3 +59,35 @@ func Sleep(ctx context.Context, d time.Duration) bool {
...
+
+type StdClock struct{}
+
+func (*StdClock) Now() time.Time {
```
bradfitz, line 81: no need for the pointer receivers on all these. A zero-width struct doesn't really have an address even :)

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
 	"time"
+
+	"tailscale.com/tstime"
+	"tailscale.com/util/mak"
 )
 
+// ClockBuilder is used to configure the initial settings for a Clock. Once
+// the settings are configured as desired, call ClockBuilder.Finish to get the
+// resulting Clock. ClockBuilder can only be used to create a single Clock.
+type ClockBuilder struct {
```
bradfitz, line 18: the Java Builder pattern isn't commonly seen in Go. You trying to avoid an Options struct or the Functional Options Pattern?

tstest/clock.go
```diff
@@ -4,57 +4,495 @@
...
+	t.nextTrigger = time.Time{}
+	select {
+	case t.c <- curTime:
+	default:
+	}
+	return time.Time{}
+}
+
+// Reset reschedules the next fire time to the current simulated time + d.
+// Reset returns true if the timer was still active before the reset.
```
bradfitz, line 464: s/returns true if/reports whether/g

bradfitz (commented):
quick skim. curious higher level where this is all going... for testing stuff?

tstest/clock.go
sailorfrag: For some reason I forgot about the Options struct, which works fine here.

tstest/clock.go
sailorfrag: I don't know, I feel like loss by default seems good but sometimes loss will be bad. I think we kind of need to see what the tests end up looking like that need protection from loss in order to figure it out.

tstest/clock.go
sailorfrag: Yeah that makes sense to me. Although Peek usually means "what's about to be read", whereas this is "repeat the last thing you told me".

tstest/clock.go
sailorfrag: I feel like StartTime() risks being stuttery? I made it `GetStart` to be consistent with the mutation accessors (e.g. `SetStep`).

tstest/clock.go
sailorfrag: I went with `GetStep`, does that seem better?

tstest/clock.go
sailorfrag: Yes it's always 1 in real timers/tickers. The main place it comes up is if you have a 1 second ticker and jump time forward by a minute, now there are 60 ticks that appear all at once with no chance to have a draining goroutine get scheduled and do something meaningful. On the other hand, in other situations maybe you want to test what happens if a tick is lost (which is annoyingly difficult to create a test for without being able to arrange for that) so I feel like it needs to be an option.

tstest/clock.go
sailorfrag: Yeah, I really needed to do another pass to docstring all the things.

Now the only things without docstrings are pretty much a few unexported init receivers which should be obvious enough.

tstest/clock.go
sailorfrag: Fair.

tstest/clock.go
sailorfrag: Yeah I ended up needing it anyway to implement the new `FollowsRealTime` option that replicates what `tstester` in corp does.

sailorfrag:
Besides minor review comments / adding the missing docstrings (sorry), the main change from before is adding a `FollowRealTime` option when creating `tstest.Clock` which makes time progress at the same rate as the system clock (or, for internal testing purposes, a second `tstest.Clock`), so when 1 second elapses in the real world, the simulated clock will progress 1 second + the sum of any advances. To simplify testing, I disallowed the use of `Step` with `FollowRealTime` since they accomplish the same goal different ways.

We already have some internal tests that use a simulated clock which moves with real world time, so this allows all of our simulated clock needs to be served by a single package.

sailorfrag:
I split a subset of these changes into #8542 to avoid breaking internal builds until the non-OSS updates are merged.

PR #8465, opened 2023-06-27
cmd/tailscale: improve error message when signing without a tailnet lock key

twitchyliquid64:
Not sure exactly what help / suggestion text we should have, but coming up with a consise way to say use a device with a trusted tailnet lock key is hard.

Also @agottardo should we add something like this in MacOS/iOS?

### DIFF
--- a/cmd/tailscale/cli/network-lock.go
+++ b/cmd/tailscale/cli/network-lock.go
@@ -465,7 +465,16 @@ func runNetworkLockSign(ctx context.Context, args []string) error {
 		}
 	}
 
-	return localClient.NetworkLockSign(ctx, nodeKey, []byte(rotationKey.Verifier()))
+	err := localClient.NetworkLockSign(ctx, nodeKey, []byte(rotationKey.Verifier()))
+	// Provide a better help message for when someone clicks through the signing flow
+	// on the wrong device.
+	if err != nil && strings.Contains(err.Error(), "this node is not trusted by network lock") {
+		fmt.Fprintln(os.Stderr, "Error: Signing is not available on this device because it does not have a trusted tailnet lock key.")
+		fmt.Fprintln(os.Stderr)
+		fmt.Fprintln(os.Stderr, "Try again on a signing device instead. Tailnet admins can see signing devices on the admin panel.")
+		fmt.Fprintln(os.Stderr)
+	}
+	return err
 }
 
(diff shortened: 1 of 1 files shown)


twitchyliquid64:
Looking at figma it seems we've landed on 'signing device', so maybe we just say that, possibly even link to it via a slug URL?

agottardo:
@twitchyliquid64 Cloning to issue in corp for iOS/macOS implementation 👍

cmd/tailscale/cli/network-lock.go
```diff
@@ -465,7 +465,16 @@ func runNetworkLockSign(ctx context.Context, args []string) error {
...
 	}
 
-	return localClient.NetworkLockSign(ctx, nodeKey, []byte(rotationKey.Verifier()))
+	err := localClient.NetworkLockSign(ctx, nodeKey, []byte(rotationKey.Verifier()))
+	// Provide a better help message for when someone clicks through the signing flow
+	// on the wrong device.
+	if err != nil && strings.Contains(err.Error(), "this node is not trusted by network lock") {
+		fmt.Fprintln(os.Stderr, "Error: Signing is not available on this device because it does not have a trusted tailnet lock key.")
+		fmt.Fprintln(os.Stderr)
+		fmt.Fprintln(os.Stderr, "Perform signing on another device instead.")
```
rosszurowski, line 474: We could link to the admin console view which shows all signing devices, but that assumes they have admin access.

```
https://login.tailscale.com/admin/machines?q=attribute%3Atailnet-lock
```

rosszurowski (approved):
Thanks!

rosszurowski:
> Looking at figma it seems we've landed on 'signing device', so maybe we just say that, possibly even link to it via a slug URL?

It's a bit messy. We're using device on the native client stuff, but we're saying "signing node" in the admin console because we don't currently use the word "device" on the admin console. I'd stick with device for now to be consistent within the clients. The admin console is easier to change later.

Unless you're saying "signing device" instead of "trusted tailnet lock key" which is different.

twitchyliquid64:
I'm a little weary about having a URL in the error in case our attribute names or paths change. Any objections to `Tailnet admins can see signing devices on the admin panel.` ?

rosszurowski:
Works for me! At least it points folks in the right direction.

PR #8471, opened 2023-06-28
syncs: add ShardedMap.Mutate

bradfitz:
To let callers do atomic/CAS-like operations.

### DIFF
--- a/syncs/shardedmap.go
+++ b/syncs/shardedmap.go
@@ -59,9 +59,36 @@ func (m *ShardedMap[K, V]) Get(key K) (value V) {
 	return
 }
 
+// Mutate atomically mutates m[k] by calling mutator.
+//
+// The mutator function is called with the old value (or its zero value) and
+// whether it existed in the map and it returns the new value and whether it
+// should be set in the map (true) or deleted from the map (false).
+//
+// It returns the change in size of the map as a result of the mutation, one of
+// -1 (delete), 0 (change), or 1 (addition).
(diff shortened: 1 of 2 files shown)


syncs/shardedmap.go
```diff
@@ -59,9 +59,36 @@ func (m *ShardedMap[K, V]) Get(key K) (value V) {
 	return
 }
 
+// Mutate atomically mutates m[k] by calling mutator.
+//
+// The mutator function is called with the old value (or its zero value) and
+// whether it existed in the map and it returns the new value and whether it
+// should be added vs deleted from the map.
```
creachadair, line 66: maybe: added/kept or added/retained ?

syncs/shardedmap_test.go
```diff
@@ -41,4 +41,41 @@ func TestShardedMap(t *testing.T) {
 	if g, w := m.Len(), 0; g != w {
 		t.Errorf("got Len %v; want %v", g, w)
 	}
+
+	// Mutation adding an entry.
+	if v := m.Mutate(1, func(was string, ok bool) (string, bool) {
```
creachadair, line 46: Is it worthwhile also verifying that the implementation made the desired change in the map?

creachadair (approved):

syncs/shardedmap.go
bradfitz: clarified

syncs/shardedmap_test.go
creachadair: Ah yeah, so it does.

dsnet (approved):
I like it.

PR #8473, opened 2023-06-28
util/lru: add a package for a typed LRU cache

bradfitz:
Updates tailscale/corp#7355

### DIFF
--- a/util/lru/lru.go
+++ b/util/lru/lru.go
@@ -1,0 +1,110 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+// Package lru contains a typed Least-Recently-Used cache.
+package lru
+
+import (
+	"container/list"
+)
+
+// Cache is container type keyed by K, storing V, optionally evicting the least
+// recently used items if a maximum size is exceeded.
+//
+// The zero value is valid to use.
+//
+// It is not safe for concurrent access.
+//
+// The current implementation is just the traditional LRU linked list; a future
+// implementation may be more advanced to avoid pathological cases.
+type Cache[K comparable, V any] struct {
+	// MaxEntries is the maximum number of cache entries before
+	// an item is evicted. Zero means no limit.
+	MaxEntries int
+
+	ll *list.List
+	m  map[K]*list.Element // of *entry[K,V]
+}
+
+// entry is the element type for the container/list.Element.
+type entry[K comparable, V any] struct {
+	key   K
+	value V
+}
+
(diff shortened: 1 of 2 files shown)


types/lru/lru.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Cache is an LRU cache keyed by K, storing V.
+//
+// The zero value is valid to use.
+//
+// It is not safe for concurrent access.
+type Cache[K comparable, V any] struct {
+	// MaxEntries is the maximum number of cache entries before
+	// an item is evicted. Zero means no limit.
+	MaxEntries int
```
dsnet, line 19: I've had need for caches that we're evicted by time rather than MaxEntries, but that can be a future consideration.

types/lru/lru.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Contains reports whether c contains key.
+func (c *Cache[K, V]) Contains(key K) bool {
+	_, ok := c.GetOk(key)
+	return ok
+}
+
+// GetOk looks up a key's value from the cache, also reporting
+// whether it was present.
+func (c *Cache[K, V]) GetOk(key K) (value V, ok bool) {
```
dsnet, line 68: Comment that this "refreshes" the age.

types/lru/lru.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Get looks up a key's value from the cache, returning either
+// the value or the zero value if it not present.
+func (c *Cache[K, V]) Get(key K) V {
+	v, _ := c.GetOk(key)
+	return v
+}
+
+// Contains reports whether c contains key.
+func (c *Cache[K, V]) Contains(key K) bool {
```
dsnet, line 61: Comment that this "refreshes" the age.

dsnet (approved):
"types" feels like a weird home for it. I think of "types" as lightweight type declaration shared by packages with minimal functionality.

dsnet:
Also, do you think LRU is carrying too much implementation details in the name? One could imagine this using a more advanced algorithm like [ARC](https://en.wikipedia.org/wiki/Adaptive_replacement_cache) or [CLOCK](https://en.wikipedia.org/wiki/Page_replacement_algorithm#Variants_of_clock), both of which are resistant to linear scanning destroying the cache contents.

bradfitz:
Moved to util, alongside util/set, etc.

bradfitz:
> Also, do you think "lru" in the package name is carrying too much implementation details in the name? 

Clarified docs per chat after being unable to find a better name.

PR #8477, opened 2023-06-28
logpolicy,ipn,cmd/tailscaled: make log files unique

### LINKED ISSUE #8476: tailscaled sometimes uses the same log files as other tailscaled instances on the same file system
### What is the issue?

It's possible to run multiple tailscaled clients on the same machine but if they are invoked in a similar manner, they will use the same log files as other clients, resulting in corrupting each other's logs.

A few useful cases to think about:

1. Common case: a single tailscaled process running in kernel mode. Users will expect the logs to be in the usual system log location.
2. Two kernel mode tailscaled processes. Each will be passed its own different state/statedir but currently they will open the same log files.
3. Two userspace-networking mode tailscaled processes. Again, will have different state/statedir but will open the same log files.

I propose that we add some unique element to the log file pathname and store it in the client configuration in state/statedir. I tried to think of some usefully human-readable string, but the only thing that is both unique and stable to the client is the full pathname to the statedir. Everything else can change: port, machine name, tun name, tailscale IP, etc. So probably the string should be randomly generated.

### Steps to reproduce

_No response_

_No response_

### OS

_No response_

### OS version

_No response_

### Tailscale version

_No response_

### Other software

_No response_

### Bug report

_No response_

valscale:
Proof of concept to generate unique log files and store them in the client's persistent config so that multiple clients can be run on the same machine without stomping on each other's log files. This uses os.MkdirTemp() to generate uniquely named subdirectories within the existing suite of potential log directories.

Fixes #8476

### DIFF
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -32,6 +32,7 @@ import (
 	"tailscale.com/cmd/tailscaled/childproc"
 	"tailscale.com/control/controlclient"
 	"tailscale.com/envknob"
+	"tailscale.com/ipn"
 	"tailscale.com/ipn/ipnlocal"
 	"tailscale.com/ipn/ipnserver"
 	"tailscale.com/ipn/store"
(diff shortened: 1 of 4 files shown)


crawshaw (commented):
I’m struggling to page the client back into my memory, but it looks like you’re creating a new temp directory on each run? If so that means tailscaled won’t be able to upload pending logs from a previous run.

valscale:
Thanks! I'm only creating the tmp dir once, then storing it in the state file and reading it back out after that. But I really like your idea of using the logtail public id and will do that! It will still need to be stored in the state file because currently I believe it is only stored in the conf file in the log dir itself, which we don't know until we read it. :)

PR #8483, opened 2023-06-28
cmd/tailscale: warn if node is locked out on bringup

twitchyliquid64:
Updates https://github.com/tailscale/corp/issues/12718


twitchyliquid64:
Friendly boop!

twitchyliquid64:
Thats a good point - what I did in this PR was wire the locked out health-check status to be allowed to be displayed at the end of `tailscale up`. This is the same place as other configuration errors are warned, such as not accepting routes.

I'm happy to change this to use the more helpful error message wording however.

rosszurowski (approved):
Approving so we can get this into 1.46. Sounds like a data race causes it to now show 100% of the time, which is what I ran into. Better to have the message there for this release and we can fix the race in the future.

rosszurowski:
Flagging for Linear:
Fixes ENG-1871

PR #8488, opened 2023-06-29
types/views: add JSON marshal/unmarshal and AsMap to Map

maisem:
This allows cloning a Map as well as marshaling the Map as JSON.

Updates tailscale/corp#12754


types/views/views.go
```diff
@@ -332,6 +333,26 @@ func (m Map[K, V]) GetOk(k K) (V, bool) {
 	return v, ok
 }
 
+// MarshalJSON implements json.Marshaler.
+func (m Map[K, V]) MarshalJSON() ([]byte, error) {
+	return json.Marshal(m.ж)
+}
+
+// UnmarshalJSON implements json.Unmarshaler.
+func (m *Map[K, V]) UnmarshalJSON(b []byte) error {
```
dsnet, line 342: The whole point of a view is to provide immutable access to the underlying type. Wouldn't `UnmarshalJSON` violate this goal?

bradfitz: See Maisem's latest push.

maisem: Fixed with an additional check. This is how we handle it in other places.
```
	if v.ж != nil {
		return errors.New("already initialized")
	}
```

bradfitz:
If you need to clone, add a clone?

If you need to a view of some JSON, unmarshal it into T and then make a View[T] ?

types/views/views.go
dsnet: Should this use the `unmarshalJSON` helper method like most other code here?

types/views/views.go
```diff
@@ -332,6 +333,29 @@ func (m Map[K, V]) GetOk(k K) (V, bool) {
 	return v, ok
 }
 
+// MarshalJSON implements json.Marshaler.
+func (m Map[K, V]) MarshalJSON() ([]byte, error) {
+	return json.Marshal(m.ж)
+}
+
+// UnmarshalJSON implements json.Unmarshaler.
```
bradfitz, line 341: say that it only works on the zero value?

types/views/views.go
```diff
@@ -332,6 +333,29 @@ func (m Map[K, V]) GetOk(k K) (V, bool) {
...
+func (m *Map[K, V]) UnmarshalJSON(b []byte) error {
+	if m.ж != nil {
+		return errors.New("already initialized")
+	}
+	return json.Unmarshal(b, &m.ж)
+}
+
+// AsMap returns a deep-clone of the underlying map.
```
bradfitz, line 349: shallow clone, no?

types/views/views.go
maisem: >Should this use the unmarshalJSON helper method like most other code here?

that only works for slices, I didn't spend too much time make it work on both slices and maps

bradfitz (approved):

dsnet (approved):

Issue #8489, opened 2023-06-29
Improve Funnel UX with new foreground mode and improved CLI


soniaappasamy:
This is a parent issue.

We're looking at improving the CLI to conform to this space's existing mental models and established norms. We have also received feedback from users and internally regarding confusion and friction in 1) establishing a Funnel, 2) understanding the differences between Serve and Funnel, and 3) understanding when a Funnel is running and remembering to turn it off.

The CLI work aims to create a more straightforward set of commands for the typical case, starting an internet Funnel targeting a local web server.

In addition, we will introduce the concept of a foreground mode with two primary goals: 1) provide the user with feedback when the Funnel is established; 2) provide a log or TUI to view requests in real-time.

The foreground mode also establishes a typical pattern of closing the tunnel (Funnel or Serve) when the command exits, which addresses feedback and concerns that servers are left running in the background or general confusion about the state of Serve/Funnel.

CLI change design doc: https://docs.google.com/document/d/180hvH5D1l0381X4SdQYUKneRGMvfGP3irzb0Vuowgxk/edit
Linear: https://linear.app/tailscale/issue/ENG-2054/funnel-cli-evolution

```[tasklist]
- [ ] https://github.com/tailscale/tailscale/issues/9292
```

tylersmalley:
This has shipped

PR #8491, opened 2023-06-29
cmd/tailscale/cli, ipn/ipnlocal: [funnel] add stream mode

soniaappasamy:
Adds ability to start Funnel in the foreground and stream incoming connections. When foreground process is stopped, Funnel is turned back off for the port.

Exampe usage:
```
TAILSCALE_FUNNEL_DEV=on tailscale funnel 443 stream
```

Updates #8489


cmd/tailscale/cli/funnel.go
```diff
@@ -117,6 +130,33 @@ func (e *serveEnv) runFunnel(ctx context.Context, args []string) error {
...
+	sc, err := e.lc.GetServeConfig(ctx)
+	if err != nil {
+		return err
+	}
+	if sc == nil {
+		sc = new(ipn.ServeConfig)
+	}
+	if sc.AllowFunnel[hp] {
+		return fmt.Errorf("Funnel already running on %q", hp)
+	}
```
shayne, line 143: `if sc == nil`, then we can return early as there's no active `ServeConfig`

ipn/ipnlocal/serve.go
```diff
@@ -257,6 +257,96 @@ func (b *LocalBackend) ServeConfig() ipn.ServeConfigView {
 	return b.serveConfig
 }
 
+// StreamFunnel turns on Funnel for the given HostPort, and keeps it on
```
shayne, line 260: I suggest being explicit about "enables Funnel for the ServeConfig"

ipn/localapi/localapi.go
```diff
@@ -842,6 +843,30 @@ func (h *Handler) serveServeConfig(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (h *Handler) serveStreamFunnel(w http.ResponseWriter, r *http.Request) {
+	if !h.PermitWrite {
```
shayne, line 847: Comment why it's write? Because we implicitly enable funnel.

ipn/ipnlocal/serve.go
```diff
@@ -257,6 +257,96 @@ func (b *LocalBackend) ServeConfig() ipn.ServeConfigView {
...
+
+	return errors.Join(writeErrs...)
+}
+
+// maybeLogFunnelConnection logs out an ipn.FunnelRequestLog for the
+// connection if a streamer is currently active for the given destPort.
+func (b *LocalBackend) maybeLogFunnelConnection(destPort uint16, srcAddr netip.AddrPort) {
+	b.mu.Lock()
+	stream := b.funnelStreamers[destPort]
+	b.mu.Unlock()
```
shayne, line 328: Should we use an RWMutex for the funnelStreamers?

ipn/ipnlocal/serve.go
```diff
@@ -257,6 +257,96 @@ func (b *LocalBackend) ServeConfig() ipn.ServeConfigView {
...
+		f.Flush()
+	}
+
+	// Hook up connections stream.
+	b.mu.Lock()
+	mak.Set(&b.funnelStreamers, port, &writeToStream)
+	b.mu.Unlock()
+
+	select {
+	case <-ctx.Done():
```
shayne, line 313: Comment on what should trigger `ctx.Done()`? Does this handle if the client process was killed (SIGKILL)?

ipn/ipnlocal/serve.go
```diff
@@ -257,6 +257,96 @@ func (b *LocalBackend) ServeConfig() ipn.ServeConfigView {
...
+	if err != nil {
+		return err
+	}
+
+	// Turn on Funnel for the given HostPort.
+	sc := b.ServeConfig().AsStruct()
+	mak.Set(&sc.AllowFunnel, hp, true)
+	if err := b.SetServeConfig(sc); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
soniaappasamy: Yeah good question - I guess we could still show the stream, that would be cool. Although then maybe we actually need to support multiple streamers at once?

ipn/ipnlocal/serve.go
soniaappasamy: > Does this handle if the client process was killed (SIGKILL)?

Yeah it does, adding comment

cmd/tailscale/cli/funnel.go
soniaappasamy: I think it's ok to treat this the same as if we use `stream` but don't have a serve config for the hostport (even if `sc != nil`). In that case we show the warning message, but still start things up:
```
Warning: funnel=on for tardos-2.sando.ts.net:443, but no serve config
         run: `tailscale serve --help` to see how to configure handlers
Funnel started on "https://tardos-2.sando.ts.net:443".
Press Ctrl-C to stop Funnel.
```
(same thing that happens when you are just turning funnel on for a port that lacks config - warns but doesn't fail)

ipn/ipnlocal/serve.go
soniaappasamy: I'm using the general `sync.Mutex` from the LocalBackend - you suggesting one for particularly the streamers? Also interested in your thoughts on the new typing, which now supports multiple streamers watching one hostport at the same time.

```
funnelStreamers    map[uint16]map[uint32]func(ipn.FunnelRequestLog) // serve port => map of stream loggers (key is UUID)
```

ipn/ipnlocal/serve.go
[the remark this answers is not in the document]
soniaappasamy: Updated to support streaming even when already on! LMK what you think

cmd/tailscale/cli/funnel.go
```diff
@@ -66,20 +67,20 @@ func (e *serveEnv) runFunnel(ctx context.Context, args []string) error {
 		return flag.ErrHelp
 	}
 
-	var on bool
+	var stream, on bool
 	switch args[1] {
+	case "stream":
```
shayne, line 72: Can we change the arg to `"debug-stream"`?

cmd/tailscale/cli/funnel.go
```diff
@@ -117,6 +137,29 @@ func (e *serveEnv) runFunnel(ctx context.Context, args []string) error {
...
+func (e *serveEnv) streamFunnel(ctx context.Context, hp ipn.HostPort) error {
+	sc, err := e.lc.GetServeConfig(ctx)
+	if err != nil {
+		return err
+	}
+	if sc == nil {
+		sc = new(ipn.ServeConfig)
+	}
+	mak.Set(&sc.AllowFunnel, hp, true)
+	printFunnelWarning(sc)
```
shayne, line 149: Since we're going to merge this in a debug state, let's remove this bit that enables Funnel as a side-effect of running stream.

ipn/ipnlocal/serve.go
```diff
@@ -257,6 +258,111 @@ func (b *LocalBackend) ServeConfig() ipn.ServeConfigView {
...
 }
 
+// StreamFunnel opens a stream to write any incoming connections made
+// to the given HostPort out to the listening io.Writer.
+//
+// If Funnel was not already enabled for the HostPort, it is enabled
+// in the ServeConfig for the duration of the context's lifespan and
+// then turned off once the context is done.
+// If Funnel was already enabled, the ServeConfig is untouched.
+func (b *LocalBackend) StreamFunnel(ctx context.Context, w io.Writer, hp ipn.HostPort) (err error) {
```
shayne, line 268: I wonder if we should associate this with Funnel vs. Serve or something else. WDYT?

client/tailscale/localclient.go
```diff
@@ -1057,6 +1057,29 @@ func (lc *LocalClient) NetworkLockDisable(ctx context.Context, secret []byte) er
...
 }
 
+// StreamFunnel returns an io.ReadCloser that streams serve/Funnel
+// connections made to the provided HostPort.
+//
+// If Funnel was not already enabled for the HostPort in the ServeConfig,
+// the backend enables it for the duration of the context's lifespan and
+// then turns it back off once the context is closed.
+// If Funnel was already enabled, the ServeConfig is untouched.
+func (lc *LocalClient) StreamFunnel(ctx context.Context, hp ipn.FunnelStreamRequest) (io.ReadCloser, error) {
```
soniaappasamy, line 1067: Maybe now that this is anchored to serve we rename this func? And the others with "Funnel" naming

And then I think some of the docs may be outdated

cmd/tailscale/cli/funnel.go
```diff
@@ -20,7 +20,13 @@ import (
 	"tailscale.com/util/mak"
 )
 
-var funnelCmd = newFunnelCommand(&serveEnv{lc: &localClient})
+var funnelCmd = func() *ffcli.Command {
+	se := &serveEnv{lc: &localClient}
+	if os.Getenv("TAILSCALE_FUNNEL_V2") == "on" {
```
soniaappasamy, line 25: Nit: TAILSCALE_FUNNEL_DEV

A little clearer that this is still in development

soniaappasamy, line 25: Also add a little doc here about what this flag is/that it's currently unsupported. Could even link the GH ticket

cmd/tailscale/cli/funnel_v2.go
```diff
@@ -0,0 +1,112 @@
...
+		},
+	}
+}
+
+// runFunnel is the entry point for the "tailscale funnel" subcommand and
+// manages turning on/off funnel. Funnel is off by default.
+//
+// Note: funnel is only supported on single DNS name for now. (2022-11-15)
+func (e *serveEnv) runFunnel2(ctx context.Context, args []string) error {
+	// TODO: allow length of args to be 2 when adding "on|off"
```
soniaappasamy, line 60: Maybe just remove the on/off docs above for now if not supported? Then we can just get rid of this TODO too

cmd/tailscale/cli/funnel_v2.go
```diff
@@ -0,0 +1,112 @@
...
+	if len(args) != 1 {
+		return flag.ErrHelp
+	}
+	var source string
+	port64, err := strconv.ParseUint(args[0], 10, 16)
+	if err == nil {
+		source = fmt.Sprintf("http://127.0.0.1:%d", port64)
```
soniaappasamy, line 67: Okay so this is specifically for a web server proxy - not for TCP or serving text? I think that's okay at this stage, but worth documenting. And thinking about what would be required to potentially support the others

cmd/tailscale/cli/funnel_v2.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		return fmt.Errorf("getting client status: %w", err)
+	}
+
+	if err := e.verifyFunnelEnabled(ctx, st, 443); err != nil {
+		return err
+	}
+
+	dnsName := strings.TrimSuffix(st.Self.DNSName, ".")
+	hp := ipn.HostPort(dnsName + ":443") // TODO(marwan-at-work): support the 2 other ports
```
soniaappasamy, line 85: Nit: maybe warn if we're overriding an existing config

cmd/tailscale/cli/serve.go
```diff
@@ -130,11 +130,13 @@ func (e *serveEnv) newFlags(name string, setup func(fs *flag.FlagSet)) *flag.Fla
 // The purpose of this interface is to allow tests to provide a mock.
 type localServeClient interface {
 	Status(context.Context) (*ipnstate.Status, error)
+	StatusWithoutPeers(context.Context) (*ipnstate.Status, error)
```
soniaappasamy, line 133: Rebase off main?

cmd/tailscale/cli/funnel_v2.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// newFunnelCommand returns a new "funnel" subcommand using e as its environment.
+// The funnel subcommand is used to turn on/off the Funnel service.
+// Funnel is off by default.
+// Funnel allows you to publish a 'tailscale serve' server publicly, open to the
+// entire internet.
+// newFunnelCommand shares the same serveEnv as the "serve" subcommand. See
+// newServeCommand and serve.go for more details.
+func newFunnelCommand2(e *serveEnv) *ffcli.Command {
```
soniaappasamy, line 26: Nit: newFunnelDevCommand

ipn/serve.go
```diff
@@ -62,6 +78,33 @@ type FunnelConn struct {
 	Src netip.AddrPort
 }
 
+// FunnelStreamRequest defines the json request body
+// for the funnel stream endpoint
+type FunnelStreamRequest struct {
+	HostPort   HostPort `json:",omitempty"`
```
soniaappasamy, line 84: Docs on this and the other fields

cmd/tailscale/cli/funnel.go
tylersmalley: Let's rename the associated files/methods to match if we do that.

cmd/tailscale/cli/funnel_dev.go
```diff
@@ -0,0 +1,110 @@
...
+		return flag.ErrHelp
+	}
+	var source string
+	port64, err := strconv.ParseUint(args[0], 10, 16)
+	if err == nil {
+		source = fmt.Sprintf("http://127.0.0.1:%d", port64)
+	} else {
+		source, err = expandProxyTarget(args[0])
+	}
+	if err != nil {
```
tylersmalley, line 69: Did you intend for this check to be within the else after `expandProxyTarget`?

cmd/tailscale/cli/funnel_dev.go
```diff
@@ -0,0 +1,110 @@
...
+// Funnel allows you to publish a 'tailscale serve' server publicly, open to the
+// entire internet.
+// newFunnelCommand shares the same serveEnv as the "serve" subcommand. See
+// newServeCommand and serve.go for more details.
+func newFunnelDevCommand(e *serveEnv) *ffcli.Command {
+	return &ffcli.Command{
+		Name:      "funnel",
+		ShortHelp: "Turn on/off Funnel service",
+		ShortUsage: strings.Join([]string{
+			"funnel <serve-port>",
```
tylersmalley, line 31:
```suggestion
			"funnel <port>",
```

tylersmalley:
@marwan-at-work - while the functionality is behind the `TAILSCALE_FUNNEL_DEV` environment variable, it still introduces concepts in code outside the CLI we know we need to change immediately. For example, right now StreamFunnel in LocalBackend should not be specific to funnel, and probably should not mutate the serve config. Of course, we can iterate on these after it's merged, but it will still create hesitancy to change the interfaces in the coming days. We could iterate quicker with more confidence if we took an iterative approach.

* Fork `tailscale funnel` and expose with the `TAILSCALE_FUNNEL_DEV` environment variable
* Change to accept the local port as an argument instead of the serve port, while updating  serve config
* Introduce foreground mode while making it the default

cmd/tailscale/cli/funnel_v2.go
marwan-at-work: @soniaappasamy this actually brings up a question of where the config manipulation happens: on the client (cli) or on the backend (tailscaled): right now it's on the backend which has the benefit of "ensured cleanup" via the "defer" that when the connection is severed (gracefully or otherwise) we know the serve is going to be cleaned up. 

If the CLI, on the other hand, gets/sets/defers config manipulation, then it has the added benefit of being able to check/warn on pre-existing serves. But that also means a "defer" may actually not end up running if the client connection is killed ungracefully and the backend right now is not smart enough to clean only a certain part of the whole config.

So with the current implementation (config manipulation on the backend), we can either: 

1. Duplicate some of the logic and let the CLI get the serve config and check/warn. Or,
2. The backend can actually push some warnings and not just request-logs that we can show the user. 

I'm leaning towards option 2 but as a follow up. 

In the future though, it might be nice to follow the docker model: create a single serve as a separate end point, and then send another request to attach yourself to its logs. 

I'm not sure if this is something we should do soon, or if it requires some designs first.

tylersmalley:
@soniaappasamy, as a follow-up to my last message -  @marwan-at-work and I discussed and if you're OK with the recent changes here we would like to move forward and will continue to iterate with everything behind the environment flag.

cmd/tailscale/cli/funnel_v2.go
marwan-at-work: All of the above sounds good! ✌🏼 

> But then does add back an extra step that would be nice to eliminate in the "start up a quick proxy so I can show someone something and then tear it down" situation.

FWIW the end user would still only run a single command, so that detail/complexity would be hidden from them. Under the hood, the CLI would have to make a a couple of calls to the backend instead of one as you mentioned.

soniaappasamy (commented):
Left a couple more nits/suggestions, but looks good! Github won't let me click "Approve" because I was the original author of the pull request, but LGTM. ✅

ipn/ipnlocal/serve.go
```diff
@@ -257,6 +258,166 @@ func (b *LocalBackend) ServeConfig() ipn.ServeConfigView {
...
+	select {
+	case <-ctx.Done():
+		// Triggered by foreground `tailscale funnel` process
+		// (the streamer) getting closed, or by turning off Tailscale.
+	}
+
+	return errors.Join(writeErrs...)
+}
+
+func setHandler(sc *ipn.ServeConfig, req ipn.FunnelStreamRequest) {
```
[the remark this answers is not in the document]
marwan-at-work: For sure, though the signature ideally would be more generic than taking a stream request. I'll leave this for now so that it stays a "private function" and not increase the public API surface.

tylersmalley (approved):

ipn/ipnlocal/serve.go
```diff
@@ -257,6 +258,165 @@ func (b *LocalBackend) ServeConfig() ipn.ServeConfigView {
...
+
+	// Turn on Funnel for the given HostPort.
+	sc := b.ServeConfig().AsStruct()
+	if sc == nil {
+		sc = &ipn.ServeConfig{}
+	}
+	setHandler(sc, req)
+	if err := b.SetServeConfig(sc); err != nil {
+		return fmt.Errorf("errro setting serve config: %w", err)
+	}
```
shayne, line 288: We shouldn't use `ServeConfig` for the foreground feature as it's persisted. If the process crashes, something goes wrong or the cleanup doesn't happen we have a foot gun where the next time Tailscale comes up the Funnel/Serve is running in the background. We should have an in-memory-only way of configuring a foreground Funnel/Serve that can be included in a new abstraction above `ServeConfig`. This can be addressed in a follow-up.

ipn/ipnlocal/serve.go
```diff
@@ -257,6 +258,165 @@ func (b *LocalBackend) ServeConfig() ipn.ServeConfigView {
...
+// If Serve and Funnel were not already enabled for the HostPort in the ServeConfig,
+// the backend enables it for the duration of the context's lifespan and
+// then turns it back off once the context is closed. If either are already enabled,
+// then they remain that way but logs are still streamed
+func (b *LocalBackend) StreamServe(ctx context.Context, w io.Writer, req ipn.ServeStreamRequest) (err error) {
+	f, ok := w.(http.Flusher)
+	if !ok {
+		return errors.New("writer not a flusher")
+	}
+	f.Flush()
```
bradfitz, line 273: If this is required, put it in the signature instead of `io.Writer`?

ipn/ipnlocal/serve.go
```diff
@@ -257,6 +258,165 @@ func (b *LocalBackend) ServeConfig() ipn.ServeConfigView {
...
+	}
+	setHandler(sc, req)
+	if err := b.SetServeConfig(sc); err != nil {
+		return fmt.Errorf("errro setting serve config: %w", err)
+	}
+	// Defer turning off Funnel once stream ends.
+	defer func() {
+		sc := b.ServeConfig().AsStruct()
+		deleteHandler(sc, req, port)
+		err = errors.Join(err, b.SetServeConfig(sc))
```
bradfitz, line 293: this is hiding other errors

ipn/ipnlocal/serve.go
```diff
@@ -257,6 +258,165 @@ func (b *LocalBackend) ServeConfig() ipn.ServeConfigView {
...
+	b.mu.Lock()
+	streamers := b.serveStreamers[destPort]
+	b.mu.Unlock()
+	if len(streamers) == 0 {
+		return
+	}
+
+	var log ipn.FunnelRequestLog
+	log.SrcAddr = srcAddr
+	log.Time = time.Now() // TODO: use a different clock somewhere?
```
bradfitz, line 403: b.clock exists.  So b.clock.Now()

ipn/serve.go
```diff
@@ -62,6 +78,41 @@ type FunnelConn struct {
 	Src netip.AddrPort
 }
 
+// ServeStreamRequest defines the json request body
```
bradfitz, line 81: JSON

ipn/serve.go
```diff
@@ -62,6 +78,41 @@ type FunnelConn struct {
...
+// for the serve stream endpoint
+type ServeStreamRequest struct {
+	// HostPort is the DNS and port of the tailscale
+	// URL.
+	HostPort HostPort `json:",omitempty"`
+
+	// Source is the user's serve destination
```
bradfitz, line 88: Why's it called "Source" if the documentation calls it a "destination"?

ipn/serve.go
```diff
@@ -62,6 +78,41 @@ type FunnelConn struct {
...
+	// such as their localhost server.
```
bradfitz, line 89: can it have a port? a URL scheme? give some example values.

ipn/serve.go
```diff
@@ -62,6 +78,41 @@ type FunnelConn struct {
...
+	Source string `json:",omitempty"`
+
+	// MountPoint is the path prefix for
+	// the given HostPort.
+	MountPoint string `json:",omitempty"`
```
bradfitz, line 94: "omitempty" implies this can be empty? What's its zero value mean then?

cmd/tailscaled/depaware.txt
```diff
@@ -93,6 +93,7 @@ tailscale.com/cmd/tailscaled dependencies: (generated by github.com/tailscale/de
    L    github.com/google/nftables/expr                              from github.com/google/nftables+
    L    github.com/google/nftables/internal/parseexprfunc            from github.com/google/nftables+
    L    github.com/google/nftables/xt                                from github.com/google/nftables/expr+
+        github.com/google/uuid                                       from tailscale.com/ipn/ipnlocal
```
bradfitz, line 96: please remove this dep. UUIDs are overused for some random hex bytes and that package isn't worth bringing in for just that. We don't actually need proper UUIDs here.

ipn/ipnlocal/serve.go
marwan-at-work: @bradfitz the function needs _both_ `io.Writer` and `http.Flusher` -- there isn't an interface that combines them. Should we define: 

```
type WriteFlusher interface {
  io.Writer
  http.Flusher
}
```

?

Note it would need to be exported given the package structure.

ipn/ipnlocal/serve.go
marwan-at-work: I think it only hides it if callers used `errors.Is/As`. In any case, we can make SetServeConfig only log the error and not concatenate it to the function's error?

cmd/tailscale/cli/funnel.go
```diff
@@ -20,7 +20,16 @@ import (
 	"tailscale.com/util/mak"
 )
 
-var funnelCmd = newFunnelCommand(&serveEnv{lc: &localClient})
+var funnelCmd = func() *ffcli.Command {
+	se := &serveEnv{lc: &localClient}
+	// This flag is used to switch to an in-development
+	// implementation of the tailscale funnel command.
+	// See https://github.com/tailscale/tailscale/issues/7844
+	if os.Getenv("TAILSCALE_FUNNEL_DEV") == "on" {
```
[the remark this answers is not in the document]
marwan-at-work: @maisem sounds good, how should the CLI then register the new funnel codepath? A new command name? Like `tailscale funnel_dev 4545` instead of `tailscale funnel 4545`? Or shuold the env var happen on _both_ tailscaled and cli?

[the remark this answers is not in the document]
maisem: either both, or a hidden new command

ipn/serve.go
marwan-at-work: Will change the doc

ipn/serve.go
marwan-at-work: @bradfitz this is the argument as is that is called source and defined with examples here: https://github.com/tailscale/tailscale/blob/50b558de745582873dd89f39ecc14a0d19401c76/cmd/tailscale/cli/serve.go#L43

I can mention that and copy the examples.

ipn/serve.go
marwan-at-work: We can treat that as an error or default to `/`? Either way is fine with me

cmd/tailscale/cli/funnel.go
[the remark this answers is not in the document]
tylersmalley: @maisem, is the concern that someone finds and uses the environment variable to expose the work under development?

ipn/serve.go
marwan-at-work: Addressed in https://github.com/tailscale/tailscale/pull/9014

ipn/ipnlocal/serve.go
marwan-at-work: Addressed in https://github.com/tailscale/tailscale/pull/9014

ipn/ipnlocal/serve.go
```diff
@@ -257,6 +258,165 @@ func (b *LocalBackend) ServeConfig() ipn.ServeConfigView {
...
+		b.serveStreamers[port] = make(map[uint32]func(ipn.FunnelRequestLog))
+	}
+	id := uuid.New().ID()
+	b.serveStreamers[port][id] = writeToStream
+	b.mu.Unlock()
+
+	// Clean up streamer when done.
+	defer func() {
+		b.mu.Lock()
+		mak.NonNilMapForJSON(&b.serveStreamers)
```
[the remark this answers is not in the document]
marwan-at-work: Addressed in https://github.com/tailscale/tailscale/pull/9014

cmd/tailscale/cli/funnel.go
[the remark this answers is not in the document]
maisem: > is the concern that someone finds and uses the environment variable to expose the work under development?
yes, or uses the LocalAPI directly without ever seeing the env var.

Experimental features that aren't fully baked yet or might have bugs should be disabled by default in tailscaled.

PR #8492, opened 2023-06-29
tsnet: add Server.AdvertisedTags option

valentinalexeev:
expose ipn.Prefs AdvertiseTags to library consumers

As part of my work on the [Tailscale-based Ingress Controller](https://github.com/valentinalexeev/tailscale-ingress-controller), I would like to be able to advertise additional tags on the tailing for HTTP and TCP services.

To achieve this, I need the ``tsnet`` library to accept the ``AdvertiseTags`` as an option when Server starts.

### DIFF
--- a/tsnet/tsnet.go
+++ b/tsnet/tsnet.go
@@ -113,7 +113,11 @@ type Server struct {
 	// field at zero unless you know what you are doing.
 	Port uint16
 
-	// Additional tags to advertise to Tailnet.
+	// AdvertiseTags specifies groups that this embedded server wants to join, for
+	// purposes of ACL enforcement. These can be referenced from the ACL
+	// security policy. Note that advertising a tag doesn't guarantee that
+	// the control server will allow you to take on the rights for that
+	// tag.
 	AdvertiseTags []string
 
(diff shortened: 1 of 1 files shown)


tsnet/tsnet.go
```diff
@@ -113,6 +113,9 @@ type Server struct {
 	// field at zero unless you know what you are doing.
 	Port uint16
 
+	// Additional tags to advertise to Tailnet.
```
bradfitz, line 116: Please write the docs in the same style as the ones above, starting with the subject being the name of the field.

Also, please match your commit message style to our project's style. We don't use Semantic Commit Messages, nor Slack emoji. Just "tsnet: add Server.AdvertisedTags option" as the title is good.

Thanks!

valentinalexeev: Thanks for the feedback, @bradfitz! Updated the pull request as suggested.

I cannot change the signoff state on the commit as per the instructions. Is there still something I can/need to do to pass that check?

bradfitz: `git commit --amend -s` to "sign" the commit for that check to be happy.

Can you squash your three commits down into one with the earlier mentioned commit message title?

Issue #8502, opened 2023-06-30
util/linuxfw: investigate package availability and support more archs.


KevinLiang10:
### What is the issue?

There are five files including `iptables.go`,  `nftables.go`, `nftables_types.go`, `linuxfw_struct_linux_test.go` and `linuxfw_unsupported.go` that currently needs extra build tags to avoid building on some architectures due to lack of support in upstream dependencies. We should investigate to remove these constrains on architectures.

### Steps to reproduce

The files would not compile on architectures including: 386,loong64,arm,armbe,mips,mips64,mips64le,mipsle,riscv64.

### OS

Linux

PR #8503, opened 2023-06-30
util/linuxfw: add new arch build constraints

KevinLiang10:
Exclide GOARCHs including: mips, mips64, mips64le, mipsle, riscv64. These archs are not supported by gvisor.dev/gvisor/pkg/hostarch.

Updates: #391, #8502

The `gvisor` dependency only supports arm64 and x86 architectures. So we should not include relative files in build when we are on other archs, and include the linuxfw_unsupported file in the build. 

Notice that only `iptables.go` is referencing `gvisor` so technically we only need to exclude `iptables.go` from build. Tho all these files are not really used to we can exclude them just to be safe.

### DIFF
--- a/util/linuxfw/iptables.go
+++ b/util/linuxfw/iptables.go
@@ -1,7 +1,8 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
-//go:build linux && !(386 || loong64 || arm || armbe)
+// TODO(#8502): add support for more architectures
+//go:build linux && (arm64 || amd64)
 
 package linuxfw
 
--- a/util/linuxfw/linuxfw_unsupported.go
+++ b/util/linuxfw/linuxfw_unsupported.go
@@ -1,10 +1,11 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
-// NOTE: linux_{386,loong64,arm,armbe} are currently unsupported due to missing
+// NOTE: linux_{arm64, x86} are the only two currently supported archs due to missing
 // support in upstream dependencies.
 
-//go:build !linux || (linux && (386 || loong64 || arm || armbe))
+// TODO(#8502): add support for more architectures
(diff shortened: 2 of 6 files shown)


util/linuxfw/iptables.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
```
raggi, line 3: Can you add `// TODO(#8502): add support for more architectures`

util/linuxfw/linuxfw_unsupported.go
```diff
@@ -1,10 +1,10 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
-// NOTE: linux_{386,loong64,arm,armbe} are currently unsupported due to missing
+// NOTE: linux_{arm64, x86} are the only two currently supported archs due to missing
```
raggi, line 4: Add a reference to #8502 here too

util/linuxfw/nftables.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
-//go:build linux && !(386 || loong64 || arm || armbe)
+//go:build linux && (arm64 || amd64)
```
raggi, line 4: #8502 here too

util/linuxfw/nftables_types.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
-//go:build linux && !(386 || loong64 || arm || armbe)
+//go:build linux && (arm64 || amd64)
```
raggi, line 4: #8502 here too

raggi (approved):
Looks good, please add the #8502 references :)

raggi:
would you mind squashing it down to two commits:

- one to mark the flaky test
- one to set the correct build tags

Issue #8511, opened 2023-07-01
k8s Makefile applies directly to kube cluster without opportunity to inspect changes


wolever:
### What is the issue?

Following the k8s setup instructions, changes are applied directly to whichever kube cluster is currently default for `kubectl`. This doesn't play well with infrastructure-as-code tools, and opens the possibility of accidentally deploying to a production environment.

Ideally the Makefile would either write to stdout or create a yaml file, allowing the caller to review/save/etc the generated yaml as appropriate for their environment and processes.

### Steps to reproduce

Notice that `tailscale/docs/k8s/Makefile` calls `kubectl apply -f -` directly.

PR #8512, opened 2023-07-01
docs/k8s: don't call kubectl directly from Makefile

wolever:
Instead of calling kubectl directly in k8s Makefile, write the yaml to stdout so it can be reviewed/edited/etc before manually applying with kubectl.

Docs can be updated to, ex, `make subnet-router | kubectl apply -f -` so they can continue to work with "one click".

Fixes: https://github.com/tailscale/tailscale/issues/8511

Ping: @maisem

### DIFF
--- a/docs/k8s/README.md
+++ b/docs/k8s/README.md
@@ -26,7 +26,7 @@ There are quite a few ways of running Tailscale inside a Kubernetes Cluster, som
    ```bash
    export SA_NAME=tailscale
    export TS_KUBE_SECRET=tailscale-auth
-   make rbac
+   make rbac | kubectl apply -f-
    ```
 
 ### Sample Sidecar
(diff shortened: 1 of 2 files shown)


docs/k8s/README.md
```diff
@@ -133,7 +133,7 @@ the entire Kubernetes cluster network (assuming NetworkPolicies allow) over Tail
 1. Deploy the subnet-router pod.
 
    ```bash
-   make subnet-router
+   make subnet-router | kubectl apply -f -
```
maisem, line 136: nit: `-f-`
here and above

maisem:
@wolever instead of `Issue: #8511` can you change the commit message to have `Fixes #8511` 
and change the commit title to `docs/k8s: don't call kubectl directly from Makefile`

Otherwise LGTM

docs/k8s/README.md
wolever: Fixed!

wolever:
Great, thanks @maisem! I've fixed up a couple other little things and tested to confirm that each target applies cleanly with `kubectl -f-` (nb: have not checked functionality, but that should be unaffected by this change).

Also note: because `kubectl apply` has been used (instead of `create`), the `kubectl delete` that was in the previous Makefile is unnecessary.

maisem:
Thanks for the change! will merge concurrently with doc changes on https://tailscale.com/kb/1185/kubernetes/

PR #8513, opened 2023-07-01
release/dist: add helper to build Go binaries with custom tags

danderson:
Updates tailscale/corp#12861


creachadair (approved):
It seems like the windows CI failure is unrelated, although if it's a flake it's not one I've observed before.

danderson:
It's a flake, seen by someone else today.

PR #8514, opened 2023-07-01
metrics: add LabelMap.GetIncrFunc

bradfitz:
Updates tailscale/corp#7354

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -45,6 +45,14 @@ func (m *LabelMap) Get(key string) *expvar.Int {
 	return m.Map.Get(key).(*expvar.Int)
 }
 
+// GetIncrFunc returns a function that increments the expvar.Int named by key.
+//
+// Most callers should not need this; it exists to satisfy an
+// interface elsewhere.
+func (m *LabelMap) GetIncrFunc(key string) func(delta int64) {
+	return m.Get(key).Add
+}
+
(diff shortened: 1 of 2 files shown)


metrics/metrics.go
```diff
@@ -45,6 +45,12 @@ func (m *LabelMap) Get(key string) *expvar.Int {
 	return m.Map.Get(key).(*expvar.Int)
 }
 
+// GetIncrFunc returns a function that increments the expvar.Int named by key.
+func (m *LabelMap) GetIncrFunc(key string) func(delta int64) {
```
bradfitz, line 49: This is so it can implement the same interface as DomainVar which increments two parallel metrics.

metrics/metrics.go
```diff
@@ -45,6 +45,12 @@ func (m *LabelMap) Get(key string) *expvar.Int {
...
+	v := m.Get(key)
+	return func(delta int64) { v.Add(delta) }
```
creachadair, line 51: Would it suffice to return v.Add directly rather than a separate closure?

creachadair (approved):

metrics/metrics.go
bradfitz: It would indeed and I'm surprised I didn't! Probably because I'd just written the other implemention which couldn't.

maisem (approved):

Issue #8493, opened 2023-06-29
use testwrapper json output for ci workflow 


clairew:
In ci workflow, write the testwrapper json output to a file. add step in github workflow and write file after running testwrapper

PR #8520, opened 2023-07-02
cmd/testwrapper: handle build failures

maisem:
`go test -json` outputs invalid JSON when a build fails. Handle that case by reseting the json.Decode and continuing to read.

Updates #8493


bradfitz (approved):
This testable?

maisem:
Created a tracking bug (https://github.com/tailscale/tailscale/issues/8586), will add tests

PR #8537, opened 2023-07-06
tempfork/heap: add copy of Go's container/heap but using generics

bradfitz:
Updates tailscale/corp#7354


bradfitz:
(This is probably going to complain about license headers, isn't it @willnorris?)

bradfitz:
> (This is probably going to complain about license headers, isn't it @willnorris?)

Ah, right, the answer is our `tempfork/` directory. 😍

maisem (approved):

PR #8541, opened 2023-07-06
wgengine/magicsock: remove noV4/noV6 check in addrForSendWireGuardLocked

catzkorn:
This change removes the noV4/noV6 check from addrForSendWireGuardLocked.

On Android, the client panics when reaching	`rand.Intn()`, likely due to the candidates list being containing no candidates. The suspicion is that the `noV4` and the `noV6` are both being triggered causing the loop to continue.

Updates tailscale/corp#12938

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/tailscale/wireguard-go/conn"
 	"go4.org/mem"
+	"golang.org/x/exp/maps"
 	"golang.org/x/net/ipv4"
 	"golang.org/x/net/ipv6"
 	"tailscale.com/control/controlclient"
@@ -4409,16 +4410,12 @@ func (de *endpoint) addrForWireGuardSendLocked(now mono.Time) (udpAddr netip.Add
 		return udpAddr, false
 	}
 
-	candidates := make([]netip.AddrPort, 0, len(de.endpointState))
-	for ipp := range de.endpointState {
-		if ipp.Addr().Is4() && de.c.noV4.Load() {
-			continue
-		}
-		if ipp.Addr().Is6() && de.c.noV6.Load() {
-			continue
-		}
-		candidates = append(candidates, ipp)
+	candidates := maps.Keys(de.endpointState)
+	if len(candidates) == 0 {
+		de.c.logf("magicsock: addrForSendWireguardLocked: [unexpected] no candidates available for endpoint")
+		return udpAddr, false
 	}
+
 	// Randomly select an address to use until we retrieve latency information
 	// and give it a short trustBestAddrUntil time so we avoid flapping between
 	// addresses while waiting on latency information to be populated.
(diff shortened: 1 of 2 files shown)


raggi (approved):

bradfitz:
Wasn't there a reason that feature existed before? Isn't this regressing something else to fix another thing?

wgengine/magicsock/magicsock.go
```diff
@@ -4411,14 +4411,14 @@ func (de *endpoint) addrForWireGuardSendLocked(now mono.Time) (udpAddr netip.Add
...
 	candidates := make([]netip.AddrPort, 0, len(de.endpointState))
 	for ipp := range de.endpointState {
-		if ipp.Addr().Is4() && de.c.noV4.Load() {
-			continue
-		}
-		if ipp.Addr().Is6() && de.c.noV6.Load() {
-			continue
-		}
 		candidates = append(candidates, ipp)
 	}
```
bradfitz, line 4415: if you want, at this point these 4 lines can use https://pkg.go.dev/golang.org/x/exp/maps#Keys and be:

```go
     candidates := maps.Keys(de.endpointState)
```

bradfitz:
> Wasn't there a reason that feature existed before? Isn't this regressing something else to fix another thing?

Oh, this is only for WG-only endpoints from ddb4040aa0cd7cdb. But what now prevents a user without IPv6 from randomly getting an IPv6 endpoint 50% of the time and then it not working?

Maybe only remove the broken IPv4 check?

wgengine/magicsock/magicsock.go
```diff
@@ -4411,14 +4411,18 @@ func (de *endpoint) addrForWireGuardSendLocked(now mono.Time) (udpAddr netip.Add
 
 	candidates := make([]netip.AddrPort, 0, len(de.endpointState))
 	for ipp := range de.endpointState {
-		if ipp.Addr().Is4() && de.c.noV4.Load() {
```
raggi, line 4414: This might be fixing part of an issue right now, but I think it's fixing a symptom of a different problem.

There are several things at play:
- the comment over nov4/nov6 says they're only for logging and only at startup. If they're only for logging we shouldn't be using them here.
- the log data we've inspected suggests neither of the nov* fields should be set, so there's some other bug (perhaps conn.mu is being held too long at startup preventing the netcheck from applying in a timely fashion).
- per brads comment it is possible for a user to get stuck on a particular protocol version here, but it could happen in both directions, stuck on v4 but needing v6 or vice versa. I'm not sure we should be treating either of v4 or v6 specially here, so I'd prefer to see us remove both or neither of these checks.

wgengine/magicsock/magicsock.go
```diff
@@ -4411,14 +4411,18 @@ func (de *endpoint) addrForWireGuardSendLocked(now mono.Time) (udpAddr netip.Add
...
-			continue
-		}
 		if ipp.Addr().Is6() && de.c.noV6.Load() {
 			continue
 		}
+
 		candidates = append(candidates, ipp)
 	}
+
+	if len(candidates) == 0 {
```
raggi, line 4421: This is the part that's urgent, we could land with just this.

PR #8552, opened 2023-07-07
control/controlclient: use ctx passed down to NoiseClient.getConn

maisem:
Without this, the client would just get stuck dialing even if the context was canceled.

Updates tailscale/corp#12590

### DIFF
--- a/control/controlclient/noise.go
+++ b/control/controlclient/noise.go
@@ -287,6 +287,25 @@ func (nc *NoiseClient) GetSingleUseRoundTripper(ctx context.Context) (http.Round
 	return nil, nil, errors.New("[unexpected] failed to reserve a request on a connection")
 }
 
+// contextErr is an error that wraps another error and is used to indicate that
+// the error was because a context expired.
+type contextErr struct {
+	err error
+}
+
+func (e contextErr) Error() string {
+	return e.err.Error()
+}
+
+func (e contextErr) Unwrap() error {
+	return e.err
+}
+
+// getConn returns a noiseConn that can be used to make requests to the
+// coordination server. It may return a cached connection or create a new one.
+// Dials are singleflighted, so concurrent calls to getConn may only dial once.
+// As such, context values may not be respected as there are no guarantees that
+// the context passed to getConn is the same as the context passed to dial.
 func (nc *NoiseClient) getConn(ctx context.Context) (*noiseConn, error) {
 	nc.mu.Lock()
 	if last := nc.last; last != nil && last.canTakeNewRequest() {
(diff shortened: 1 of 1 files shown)


control/controlclient/noise.go
```diff
@@ -295,7 +295,9 @@ func (nc *NoiseClient) getConn(ctx context.Context) (*noiseConn, error) {
 	}
 	nc.mu.Unlock()
 
-	conn, err, _ := nc.sfDial.Do(struct{}{}, nc.dial)
+	conn, err, _ := nc.sfDial.Do(struct{}{}, func() (*noiseConn, error) {
+		return nc.dial(ctx)
+	})
```
bradfitz, line 300: this doesn't seem right. If N dials are coalesced together with single flight, N-1 shouldn't be canceled just because 1 of them's context was canceled?

maisem: fair point, any suggestion on how to handle this then? i.e. how do i signal shutdown?

control/controlclient/noise.go
```diff
@@ -295,11 +295,27 @@ func (nc *NoiseClient) getConn(ctx context.Context) (*noiseConn, error) {
...
+		// We singeflight the dial to avoid making multiple connections, however
+		// that means that we can't simply cancel the dial if the context is
+		// canceled. Instead, we have to additionally check that the context
+		// which was canceled is our context and retry if our context is still
+		// valid.
+		conn, err, _ := nc.sfDial.Do(struct{}{}, func() (*noiseConn, error) {
+			return nc.dial(ctx)
+		})
+		if err != nil {
+			if errors.Is(err, context.Canceled) {
```
bradfitz, line 308: not just Canceled. Could also be deadline exceeded.

bradfitz: maybe don't look at this at all and only look at ctx.Err() being non-nil?

control/controlclient/noise.go
```diff
@@ -344,7 +360,7 @@ func (nc *NoiseClient) Close() error {
 
 // dial opens a new connection to tailcontrol, fetching the server noise key
 // if not cached.
-func (nc *NoiseClient) dial() (*noiseConn, error) {
+func (nc *NoiseClient) dial(ctx context.Context) (*noiseConn, error) {
```
bradfitz, line 363: document that this ctx is 1 of N waiting dialers and thus shouldn't assume any context values.

bradfitz (approved):
LGTM overall

control/controlclient/noise.go
maisem: Redid this, PTAL

control/controlclient/noise.go
```diff
@@ -295,11 +314,35 @@ func (nc *NoiseClient) getConn(ctx context.Context) (*noiseConn, error) {
...
+		// We singeflight the dial to avoid making multiple connections, however
+		// that means that we can't simply cancel the dial if the context is
+		// canceled. Instead, we have to additionally check that the context
+		// which was canceled is our context and retry if our context is still
+		// valid.
+		conn, err, _ := nc.sfDial.Do(struct{}{}, func() (*noiseConn, error) {
+			c, err := nc.dial(ctx)
+			if err != nil {
+				if ctx.Err() != nil {
+					return nil, contextErr{err}
```
bradfitz, line 327: there's a race here: maybe the dial failed due to getting a RST from the peer. But then independently, by chance, the context was canceled just before line 326's check. Now you're wrapping a "connection refused" error in a "contextErr{...}" wrapper.

That's harmless for what you need, but doesn't match what the docs on L290 say.

bradfitz: Just do `contextErr{ctx.Err()}` instead?

maisem: yeah, i wasn't sure how to handle the race. `contextErr{ctx.Err()}` makes sense. Done.

bradfitz (approved):

Issue #8249, opened 2023-06-01
logpolicy uses an explicit log.Printf for dnsfallback instead of allowing it to be overridden


eric:
### What is the issue?

Continuation of #7537, #5743.

When using `tsnet` and setting the `Server.Logf` these log messages still appear in the main logs:

```
2023/06/01 12:00:02.348945 logtail: dial "log.tailscale.io:443" failed: dial tcp: lookup log.tailscale.io: no such host (in 3ms), trying bootstrap...
2023/06/01 12:00:02.351004 trying bootstrapDNS("derp5d.tailscale.com", "43.245.48.250") for "log.tailscale.io" ...
2023/06/01 12:00:02.351972 bootstrapDNS("derp5d.tailscale.com", "43.245.48.250") for "log.tailscale.io" error: Get "https://derp5d.tailscale.com/bootstrap-dns?q=log.tailscale.io": dial tcp 43.245.48.250:443: connect: network is unreachable
```

It appears that #7902 updated `logpolicy` to use `log.Printf`:

https://github.com/tailscale/tailscale/blob/2aa8299c375b7f105e9e599501c4607b6dc22421/logpolicy/logpolicy.go#L721-L727

eric:
Thanks!

PR #8574, opened 2023-07-10
logpolicy, various: allow overriding log function

andrew-d:
This allows sending logs from the "logpolicy" package (and associated callees) to something other than the log package. The behaviour for tailscaled remains the same, passing in log.Printf

Updates #8249

Change-Id: Ie1d43b75fa7281933d9225bffd388462c08a5f31

### DIFF
--- a/logpolicy/logpolicy.go
+++ b/logpolicy/logpolicy.go
@@ -110,6 +110,8 @@ type Policy struct {
 	Logtail *logtail.Logger
 	// PublicID is the logger's instance identifier.
 	PublicID logid.PublicID
+	// Logf is where to write informational messages about this Logger.
+	Logf logger.Logf
 }
 
 // NewConfig creates a Config with collection and a newly generated PrivateID.
@@ -310,7 +312,7 @@ func winProgramDataAccessible(dir string) bool {
 // log state for that command exists in dir, then the log state is
 // moved from wherever it does exist, into dir. Leftover logs state
 // in / and $CACHE_DIRECTORY is deleted.
-func tryFixLogStateLocation(dir, cmdname string) {
+func tryFixLogStateLocation(dir, cmdname string, logf logger.Logf) {
 	switch runtime.GOOS {
 	case "linux", "freebsd", "openbsd":
 		// These are the OSes where we might have written stuff into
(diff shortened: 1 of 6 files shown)


logpolicy/logpolicy.go
```diff
@@ -452,14 +452,14 @@ func tryFixLogStateLocation(dir, cmdname string) {
...
-	return NewWithConfigPath(collection, "", "", netMon)
+func New(collection string, netMon *netmon.Monitor, logf logger.Logf) *Policy {
+	return NewWithConfigPath(collection, "", "", netMon, logf)
 }
 
 // NewWithConfigPath is identical to New,
 // but uses the specified directory and command name.
 // If either is empty, it derives them automatically.
-func NewWithConfigPath(collection, dir, cmdName string, netMon *netmon.Monitor) *Policy {
+func NewWithConfigPath(collection, dir, cmdName string, netMon *netmon.Monitor, logf logger.Logf) *Policy {
```
dsnet, line 462: If logf is nil, should we just default to log.Printf?

logpolicy/logpolicy.go
```diff
@@ -615,16 +615,16 @@ func NewWithConfigPath(collection, dir, cmdName string, netMon *netmon.Monitor)
 	log.SetFlags(0) // other log flags are set on console, not here
 	log.SetOutput(logOutput)
 
-	log.Printf("Program starting: v%v, Go %v: %#v",
+	logf("Program starting: v%v, Go %v: %#v",
```
dsnet, line 618: On L615 and L616, we call mutating functions in the "log" package. This seems really weird if we're trying to circumvent the "log" package.

In general, it seems like there should be no calls to the "log" package if a custom `logf` function is provided.

andrew-d: That's a good point. Since `func` isn't comparable to anything other than `nil`, I think this means we'll need to special-case `nil` and have that mean to use `log.Printf` and also change global settings. Does that sound right?

dsnet: I think that's fairly normal to check if it's nil and do some default as a result.

bradfitz (commented):
Add some docs to the constructors about what the mystery parameter is.

logpolicy/logpolicy.go
andrew-d: Okay, did this; PTAL?

bradfitz (approved):

logpolicy/logpolicy.go
```diff
@@ -488,7 +495,12 @@ func NewWithConfigPath(collection, dir, cmdName string, netMon *netmon.Monitor)
 	if cmdName == "" {
 		cmdName = version.CmdName()
 	}
-	tryFixLogStateLocation(dir, cmdName)
+
+	haveCustomLogf := logf == nil
```
dsnet, line 499: Shouldn't this be `logf != nil`?

dsnet: Perhaps just invert this variable as "useStdLogger"? That way the conditionals below read nicer.

dsnet (approved):

logpolicy/logpolicy.go
andrew-d: Gooooood catch 🤦 The rename makes this much clearer.

andrew-d:
Verified that the logs printed by `tailscaled` are the same (excepting build info, times, etc.) between this branch and `main`.

PR #8579, opened 2023-07-11
cmd/derper: add X-Content-Type-Options: nosniff

awly:
This is one of generic HTTP hardening headers (for XSS) that has no impact on real usage. Doesn't do much for derper, but makes security scanners happy.

### DIFF
--- a/cmd/derper/derper.go
+++ b/cmd/derper/derper.go
@@ -185,6 +185,23 @@ func main() {
 	mux.HandleFunc("/bootstrap-dns", handleBootstrapDNS)
 	mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "text/html; charset=utf-8")
+		// Set HTTP headers to appease automated security scanners.
+		//
+		// Security automation gets cranky when HTTPS sites don't
+		// set HSTS, and when they don't specify a content
+		// security policy for XSS mitigation.
+		//
+		// DERP's HTTP interface is only ever used for debug
+		// access (for which trivial safe policies work just
+		// fine), and by DERP clients which don't obey any of
+		// these browser-centric headers anyway.
+		//
+		// Note: if we ever add a real UI component to DERP, these headers
+		// should move to a middleware wrapper that applies to more request
+		// paths.
+		w.Header().Set("Strict-Transport-Security", "max-age=63072000; includeSubDomains")
+		w.Header().Set("Content-Security-Policy", "default-src 'none'; frame-ancestors 'none'; form-action 'none'; base-uri 'self'; block-all-mixed-content; plugin-types 'none'")
+		w.Header().Set("X-Content-Type-Options", "nosniff")
 		w.WriteHeader(200)
 		io.WriteString(w, `<html><body>
 <h1>DERP</h1>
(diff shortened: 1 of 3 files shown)


bradfitz (commented):
Can we only send these on the / path and not to tailscaled clients?

awly:
Sure, looking through the other endpoints that seems innocent enough. Saves a tiny bit of bandwidth /latency for tailscaled.
Updated the comment to call out that we might need to move it, if derper ever gets more of a UI.

cmd/derper/derper.go
```diff
@@ -185,6 +185,23 @@ func main() {
 	mux.HandleFunc("/bootstrap-dns", handleBootstrapDNS)
 	mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
bradfitz, line 186: This "/" here doesn't mean only /. It means anything not covered by a more specific handler (like a mount point)

By moving it here, this is no longer on the actual HTML pages we serve (over Tailscale) via the line 226 (tsweb.Debugger) registration.

The two endpoints I want to not send these headers on are /derp and /generate_204 (the two that tailscaled hit).

So you can move it back where you add it as long as you omit those path prefixes.

Or add all this stuff to some common place (new func in tsweb, like `tsweb.AddVariousHeadersToAppeaseSecurityScanners(w http.ResponseWriter)`?) and then use it here and in the tsweb debug handlers too? Your call.

But by making the func in tailscale/tailscale, we can probably remove duplication in our other repos too.

awly: I was thinking of creating a helper function in other repos, but I suppose we import tsweb everywhere so this makes more sense.
Added some helpers in tsweb and sprinkled them around derper mux.

tsweb/tsweb.go
```diff
@@ -423,3 +423,37 @@ func Error(code int, msg string, err error) HTTPError {
 func VarzHandler(w http.ResponseWriter, r *http.Request) {
 	varz.Handler(w, r)
 }
+
+// AddBrowserHeaders sets various HTTP security headers for browser-facing endpoints.
+//
+// The specific headers:
+// - require HTTPS access (HSTS)
+// - disallow iframe embedding
+// - mitigate MIME confusion attacks
```
bradfitz, line 432: Does this follow the new formatting in https://go.dev/doc/go1.19#go-doc .... https://go.dev/doc/comment#lists ?

My editor always rewrites my comments on save into the new style but this doesn't look like it. (needs a blank line before the first bullet and them all to be indented?)

cmd/derper/derper.go
```diff
@@ -182,8 +182,9 @@ func main() {
 	}
 	mux.HandleFunc("/derp/probe", probeHandler)
 	go refreshBootstrapDNSLoop()
-	mux.HandleFunc("/bootstrap-dns", handleBootstrapDNS)
+	mux.HandleFunc("/bootstrap-dns", tsweb.BrowserHeaderHandlerFunc(handleBootstrapDNS))
```
bradfitz, line 185: This one's primarily hit by tailscaled, but only when things have gone very wrong anyway, so I don't care about the extra bytes here.

bradfitz (approved):

tsweb/tsweb.go
awly: So, the blank line is not required, but two spaces before a dash are required. Updated

bradfitz:
When you merge, be sure to squash all your commits down to one commit with a nice, conventional commit message.

Issue #8587, opened 2023-07-11
Refactor in tstime.Clock


clairew:
After adding tstime.Clock, we should replace all time package clock related usage with tstime.Clock.

PR #8595, opened 2023-07-12
control: use tstime instead of time

clairew:
Updates #8587
Signed-off-by: Claire Wang <claire@tailscale.com>


control/controlhttp/constants.go
```diff
@@ -89,6 +90,8 @@ type Dialer struct {
 	drainFinished        chan struct{}
 	omitCertErrorLogging bool
 	testFallbackDelay    time.Duration
+
+	Clock tstime.Clock
```
bradfitz, line 94: all the other exported fields are up above the unexported ones and have docs including the behavior on a zero value. Can this be nil? What does that mean? It should probably mean stdclock.

control/controlclient/direct.go
```diff
@@ -191,8 +192,8 @@ func NewDirect(opts Options) (*Direct, error) {
 	if err != nil {
 		return nil, err
 	}
-	if opts.TimeNow == nil {
-		opts.TimeNow = time.Now
+	if opts.Clock == nil {
+		opts.Clock = &tstime.StdClock{}
```
sailorfrag, line 196:
```suggestion
		opts.Clock = tstime.StdClock{}
```

control/controlclient/direct.go
```diff
@@ -1401,7 +1403,7 @@ func answerHeadPing(logf logger.Logf, c *http.Client, pr *tailcfg.PingRequest) {
 	if pr.Log {
 		logf("answerHeadPing: sending HEAD ping to %v ...", pr.URL)
 	}
-	t0 := time.Now()
+	t0 := clockNow()
 	_, err = c.Do(req)
 	d := time.Since(t0).Round(time.Millisecond)
```
sailorfrag, line 1408: these `time.Since` calls need to be converted for it to make sense

This is a pattern that appears often enough that I wonder if we should have something like `tstime.Stopwatch` that can handle this in a way that requires less plumbing and remains easy to use. This is mostly an observation, not a recommendation to make any changes, unless you have any good ideas for how to implement such a thing easily. Maybe one approach would be to not bother simulating time for "action took X ms" use cases if they're just used for logging purposes. Going through a `tstime` wrapper would still keep the rules about stdlib `time` usage clear and make static checks straightfoward.

control/controlclient/direct.go
```diff
@@ -1297,7 +1298,8 @@ func initDevKnob() devKnobs {
 	}
 }
 
-var clockNow = time.Now
+var clock = &tstime.StdClock{}
+var clockNow = clock.Now // used instead of time.Now()
```
sailorfrag, line 1302: Looking at the places that use `clockNow`, I think it would be preferable to pass a `tstime.Clock` to those `func`s as a parameter.

sailorfrag (approved):

PR #8596, opened 2023-07-12
net/netcheck, tailcfg: add DERPHomeParams and use it

andrew-d:
This allows providing additional information to the client about how to
select a home DERP region, such as preferring a given DERP region over
all others.

Updates https://github.com/tailscale/tailscale/issues/8603

Change-Id: Ic6bfe1122ef41fffbe96aae4be3a8f9e7cd9d514

### DIFF
--- a/net/netcheck/netcheck.go
+++ b/net/netcheck/netcheck.go
@@ -41,6 +41,7 @@ import (
 	"tailscale.com/types/nettype"
 	"tailscale.com/types/opt"
 	"tailscale.com/types/ptr"
+	"tailscale.com/types/views"
 	"tailscale.com/util/clientmetric"
 	"tailscale.com/util/cmpx"
 	"tailscale.com/util/mak"
@@ -1110,7 +1111,7 @@ func (c *Client) finishAndStoreReport(rs *reportState, dm *tailcfg.DERPMap) *Rep
 	report := rs.report.Clone()
 	rs.mu.Unlock()
 
-	c.addReportHistoryAndSetPreferredDERP(report)
+	c.addReportHistoryAndSetPreferredDERP(report, dm.View())
 	c.logConciseReport(report, dm)
 
 	return report
@@ -1444,7 +1445,7 @@ func (c *Client) timeNow() time.Time {
 
 // addReportHistoryAndSetPreferredDERP adds r to the set of recent Reports
 // and mutates r.PreferredDERP to contain the best recent one.
-func (c *Client) addReportHistoryAndSetPreferredDERP(r *Report) {
+func (c *Client) addReportHistoryAndSetPreferredDERP(r *Report, dm tailcfg.DERPMapView) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
@@ -1476,11 +1477,33 @@ func (c *Client) addReportHistoryAndSetPreferredDERP(r *Report) {
 		}
 	}
 
+	// Scale each region's best latency by any provided scores from the
+	// DERPMap, for use in comparison below.
+	var scores views.Map[int, float64]
+	if hp := dm.HomeParams(); hp.Valid() {
+		scores = hp.RegionScore()
+	}
+	for regionID, d := range bestRecent {
+		if score := scores.Get(regionID); score > 0 {
+			bestRecent[regionID] = time.Duration(float64(d) * score)
+		}
+	}
+
 	// Then, pick which currently-alive DERP server from the
 	// current report has the best latency over the past maxAge.
-	var bestAny time.Duration
-	var oldRegionCurLatency time.Duration
+	var (
+		bestAny             time.Duration // global minimum
+		oldRegionCurLatency time.Duration // latency of old PreferredDERP
+	)
 	for regionID, d := range r.RegionLatency {
+		// Scale this report's latency by any scores provided by the
+		// server; we did this for the bestRecent map above, but we
+		// don't mutate the actual reports in-place (in case scores
+		// change), so we need to do it here as well.
+		if score := scores.Get(regionID); score > 0 {
+			d = time.Duration(float64(d) * score)
+		}
+
 		if regionID == prevDERP {
 			oldRegionCurLatency = d
 		}
@@ -1491,13 +1514,11 @@ func (c *Client) addReportHistoryAndSetPreferredDERP(r *Report) {
 		}
 	}
 
-	// If we're changing our preferred DERP but the old one's still
-	// accessible and the new one's not much better, just stick with
-	// where we are.
-	if prevDERP != 0 &&
-		r.PreferredDERP != prevDERP &&
-		oldRegionCurLatency != 0 &&
-		bestAny > oldRegionCurLatency/3*2 {
+	// If we're changing our preferred DERP, the old one's still
+	// accessible, and the new one's not much better, just stick
+	// with where we are.
+	changingPreferred := prevDERP != 0 && r.PreferredDERP != prevDERP
+	if changingPreferred && oldRegionCurLatency != 0 && bestAny > oldRegionCurLatency/3*2 {
 		r.PreferredDERP = prevDERP
 	}
 }
--- a/tailcfg/derpmap.go
+++ b/tailcfg/derpmap.go
@@ -7,6 +7,11 @@ import "sort"
 
 // DERPMap describes the set of DERP packet relay servers that are available.
 type DERPMap struct {
+	// HomeParams, if non-nil, is a change in home parameters.
+	//
+	// The rest of the DEPRMap fields, if zero, means unchanged.
+	HomeParams *DERPHomeParams `json:",omitempty"`
+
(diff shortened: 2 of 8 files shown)


andrew-d:
I'm adding tests now, but wanted to get this up for review while I did so

tailcfg/derpmap.go
```diff
@@ -14,6 +14,10 @@ type DERPMap struct {
 	// The numbers are not necessarily contiguous.
 	Regions map[int]*DERPRegion
 
+	// DERPHomeHint is a hint to this client to use the provided DERP
+	// region ID as the home region with higher affinity than normal.
+	DERPHomeHint int `json:",omitempty"`
```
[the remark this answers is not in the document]
andrew-d: We should pick a method that allows us to send it to the node as part of the initial connection, to avoid flaps where the node starts, picks a close DERP, and then gets hinted to one that's also close but not the fastest. How about putting this in `MapResponse`?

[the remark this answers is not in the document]
bradfitz: I want to making MapResponse less of a dumping ground than it already is. It's making control too hard to work on.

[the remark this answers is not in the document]
crawshaw: I think the hint should be part of the initial response for the /derpmap because we can generate it there.

But maybe there's a way to do that efficiently? Worst case an HTTP response header.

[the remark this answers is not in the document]
bradfitz: Maybe we define at this capver that DERPMap can be more incremental, with just setting hints (a pointer, not an int) to a new DERPHomeHints type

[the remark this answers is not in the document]
crawshaw: We can deliver a good hint before you have a derp home connection and we should do that. Maybe another endpoint? Or is that what you mean?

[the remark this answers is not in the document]
bradfitz: Concrete proposal below

tailcfg/derpmap.go
```diff
@@ -14,6 +14,10 @@ type DERPMap struct {
 	// The numbers are not necessarily contiguous.
 	Regions map[int]*DERPRegion
 
+	// DERPHomeHint is a hint to this client to use the provided DERP
+	// region ID as the home region with higher affinity than normal.
```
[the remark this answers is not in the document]
andrew-d: For `DERPHomeHints`: presumably we'd use this in `addReportHistoryAndSetPreferredDERP` to weight `Report.RegionLatency` (if provided in `RegionWeight`) when calculating `bestAny` and thus `PreferredDERP`?

For the `DERPMap`... we'd allow the control server to send a `&MapResponse{DERPMap: &DERPMap{Hints: foo}}` and then have the `mapSession` properly merge the previous value (`ms.lastDERPMap`) into the provided one?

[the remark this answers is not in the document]
andrew-d: Pushed another commit with this idea; thoughts?

tailcfg/derpmap.go
```diff
@@ -7,6 +7,11 @@ import "sort"
 
 // DERPMap describes the set of DERP packet relay servers that are available.
 type DERPMap struct {
+	// Hints, if non-nil, is a change in hints.
+	//
+	// The rest of the DEPRMap fields, if zero, means unchanged.
```
bradfitz, line 12: maybe clarify that OmitDefaultRegions is only applicable when Regions is non-nil?

tailcfg/derpmap.go
```diff
@@ -29,6 +34,11 @@ func (m *DERPMap) RegionIDs() []int {
 	return ret
 }
 
+// DERPHomeHints contains hints from the server about which DERP servers to use as the home server.
+type DERPHomeHints struct {
+	RegionWeight map[int]float64 // Region ID to its latency scaling factor
```
bradfitz, line 39: specify what nil means. Probably it should mean unchanged, but an empty non-nil maps means clear all scaling factors (reset everything back to 1)

And specify that if a region doesn't have a scaling factor in the map, that means 1?

net/netcheck/netcheck.go
```diff
@@ -1476,11 +1476,33 @@ func (c *Client) addReportHistoryAndSetPreferredDERP(r *Report) {
...
 	}
 
+	// Scale each region's best latency by any provided weights from the
+	// DERPMap, for use in comparison below.
+	var weights map[int]float64
+	if dm.Hints != nil {
+		weights = dm.Hints.RegionWeight
+	}
+	for regionID, d := range bestRecent {
+		if weight := weights[regionID]; weight != 0 {
```
bradfitz, line 1486: `weight != 0` or `ok` ? Your call. But an explicit map entry with a value of 0 should be have its behavior documented, whatever it is.

bradfitz:
This needs an issue and a squash to one commit and maybe a better name for the field/type but otherwise LGTM

net/netcheck/netcheck.go
andrew-d: IMO, weights should not be 0 (or negative); I'll document this and update it to check `weight > 0` to ignore such weights.

andrew-d:
> This needs an issue and a squash to one commit and maybe a better name for the field/type but otherwise LGTM

Do we have an existing issue for the general "DERP home changes"?

net/netcheck/netcheck.go
```diff
@@ -1491,13 +1513,11 @@ func (c *Client) addReportHistoryAndSetPreferredDERP(r *Report) {
...
-	// where we are.
-	if prevDERP != 0 &&
-		r.PreferredDERP != prevDERP &&
-		oldRegionCurLatency != 0 &&
-		bestAny > oldRegionCurLatency/3*2 {
+	// If we're changing our preferred DERP, the old one's still
+	// accessible, and the new one's not much better, just stick
+	// with where we are.
+	changingPreferred := prevDERP != 0 && r.PreferredDERP != prevDERP
+	if changingPreferred && oldRegionCurLatency != 0 && bestAny > oldRegionCurLatency/3*2 {
```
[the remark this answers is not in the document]
bradfitz: that works as well as anything I can think of right now

bradfitz:
I filed https://github.com/tailscale/tailscale/issues/8603 that we can use for this.

net/netcheck/netcheck.go
```diff
@@ -1444,7 +1444,7 @@ func (c *Client) timeNow() time.Time {
 
 // addReportHistoryAndSetPreferredDERP adds r to the set of recent Reports
 // and mutates r.PreferredDERP to contain the best recent one.
-func (c *Client) addReportHistoryAndSetPreferredDERP(r *Report) {
+func (c *Client) addReportHistoryAndSetPreferredDERP(r *Report, dm *tailcfg.DERPMap) {
```
maisem, line 1447: consider taking in a View

andrew-d: Good idea; updated. PTAL?

tailcfg/derpmap.go
```diff
@@ -29,6 +36,23 @@ func (m *DERPMap) RegionIDs() []int {
...
+	// factor when determining which region to use as the home
+	// ("preferred") DERP.
+	//
+	// If a region is not present in this map, it is treated as having a
+	// weight of 1.0.
+	//
+	// Weights should not be 0 or negative; such weights will be ignored.
+	//
+	// A nil map means no change from the previous value (if any); an empty
+	// non-nil map can be sent to reset all weights back to 1.0.
```
maisem, line 52: i got confused a bit reading this and the code, code looks fine but weight implies that if the value increases, the preference increases. but its the opposite here.
What about `RegionLatencyMultiplier` or something that implies that lower values should be assigned to regions that are preferred

bradfitz: kinda a mouthful, that name

is `RegionScale` enough to not sound bigger-is-better to you?

bradfitz (approved):

tailcfg/derpmap.go
maisem: `RegionScore` with docs that call out 0<x<1 makes it preferred and >1 penalizes?

maisem (approved):

tailcfg/derpmap.go
andrew-d: Renamed to `RegionScore` and added docs on the semantics.

tailcfg/derpmap.go
```diff
@@ -29,6 +36,25 @@ func (m *DERPMap) RegionIDs() []int {
...
+	// RegionScore scales latencies of DERP regions by a given scaling
+	// factor when determining which region to use as the home
+	// ("preferred") DERP. Scores in the range (0, 1) will cause this
+	// region to be proportionally more preferred, and scores in the range
+	// (1, ∞) will penalize a region.
+	//
+	// If a region is not present in this map, it is treated as having a
+	// weight of 1.0.
+	//
+	// Weights should not be 0 or negative; such weights will be ignored.
```
maisem, line 51: s/Weight/Score

PR #8598, opened 2023-07-12
wgengine/netlog: fix nil pointer dereference in logtail

awly:
When `Logger.Startup` is called and a new `http.Client` is initialized for logtail, `nl.logger` is nil. It's initialized right after creating the `http.Client`. But because we're passing a method value, it actually works and passes the value with a `nil` receiver.
Later, when `logtail.Logger.Logf` is called from the `http.Client`, `tailscaled` panics when trying to access the fields of that nil logger.

### DIFF
--- a/wgengine/netlog/logger.go
+++ b/wgengine/netlog/logger.go
@@ -101,7 +101,7 @@ func (nl *Logger) Startup(nodeID tailcfg.StableNodeID, nodeLogID, domainLogID lo
 	}
 
 	// Startup a log stream to Tailscale's logging service.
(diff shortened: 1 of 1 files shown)


andrew-d (approved):
Very good catch, thank you 🤦  Optional, but if you feel like it... we could pull out the `log.Printf` from the call to `logtail.NewLogger` into a shared variable between that function and `logpolicy.NewLogtailTransport`. We can always do that later after this fix though 😄

awly:
@andrew-d updated, is this what you meant?

andrew-d (approved):
> @andrew-d updated, is this what you meant?

Yeah, exactly; thanks! Please make sure to squash this into the first commit, but otherwise LGTM 🎉

PR #8599, opened 2023-07-12
ipn/ipnlocal: add optional support for ACME Renewal Info (ARI)

awly:
Add optional support for the RFC draft of ARI:
https://datatracker.ietf.org/doc/draft-ietf-acme-ari/
This is supported in LetsEncrypt today.

Since the RFC is still in draft state, gate this code on
`TS_USE_ACME_ARI` env var. By default, use the old expiry-based logic
for deciding when to rotate.


ipn/ipnlocal/cert.go
```diff
@@ -157,6 +162,44 @@ func shouldStartDomainRenewal(domain string, now time.Time, pair *TLSCertKeyPair
...
+	defer cancel()
+	d, err := ac.Discover(ctx)
+	if err != nil {
+		return false, err
+	}
+	ri, err := ac.FetchRenewalInfo(ctx, blocks[0].Bytes, blocks[len(blocks)-1].Bytes)
+	if err != nil {
+		return false, fmt.Errorf("failed to fetch renewal info from ACME server: %w", err)
+	}
+	fmt.Printf("==== ARI: %+v\n", ri)
```
bradfitz, line 193: debug

awly: left it intentionally just to sanity-check that it's hitting the right code path :sweat_smile: 
will retest and remove this line after we get @noncombatant's change to x/crypto/acme cherry-picked into our fork

ipn/ipnlocal/cert.go
```diff
@@ -157,6 +170,42 @@ func shouldStartDomainRenewal(domain string, now time.Time, pair *TLSCertKeyPair
...
+	if len(blocks) < 2 {
+		return false, fmt.Errorf("could not parse certificate chain from certStore, got %d PEM block(s)", len(blocks))
+	}
+	ac, err := acmeClient(cs)
+	if err != nil {
+		return false, err
+	}
+	ctx, cancel := context.WithTimeout(b.ctx, 5*time.Second)
+	defer cancel()
+	ri, err := ac.FetchRenewalInfo(ctx, blocks[0].Bytes, blocks[len(blocks)-1].Bytes)
```
noncombatant, line 193: The last element of the slice is most likely to be the issuer when the length of the slice is 2, but if we found ourselves for some reason looking at a slice of 3+, I'd still think/guess that the 1th element is most likely to be the issuer of the 0th. (There's no particular guarantee of this — the only real check is to actually build the chain from the assumption that the slice is unordered.)

If we see a significant number of failures from `FetchRenewalInfo`, picking the wrong cert as the issuer might turn out to be the/a reason. In that case, we might to really build the chain.

For now, as a first version, I'd pass `blocks[1].Bytes` as the issuer and keep an eye on the failure log.

ipn/ipnlocal/cert.go
```diff
@@ -540,6 +585,20 @@ func acmeKey(cs certStore) (crypto.Signer, error) {
 	return privKey, nil
 }
 
+func acmeClient(cs certStore) (*acme.Client, error) {
```
noncombatant, line 588: Style pico-nit: Personally I like for function names to always be verb phrases (e.g. `getACMEClient`), but that might just be me. Either way is fine.

ipn/ipnlocal/cert.go
```diff
@@ -101,7 +102,7 @@ func (b *LocalBackend) GetCertPEM(ctx context.Context, domain string) (*TLSCertK
 	}
 
 	if pair, err := getCertPEMCached(cs, domain, now); err == nil {
-		shouldRenew, err := shouldStartDomainRenewal(domain, now, pair)
+		shouldRenew, err := b.shouldStartDomainRenewal(cs, domain, now, pair)
 		if err != nil {
 			logf("error checking for certificate renewal: %v", err)
```
noncombatant, line 107: %w here too? (I don't know; legitimately wondering)

ipn/ipnlocal/cert.go
```diff
@@ -130,6 +131,18 @@ func shouldStartDomainRenewal(domain string, now time.Time, pair *TLSCertKeyPair
 	}
 	lastRenewCheck[domain] = now
 
+	renew, err := b.shouldStartDomainRenewalByARI(cs, now, pair)
+	if err != nil {
+		// Log any ARI failure and fall back to checking for renewal by expiry.
+		b.logf("acme: ARI check failed: %v; falling back to expiry-based check", err)
```
noncombatant, line 137: %w?

noncombatant (approved):
LGTM with tiny questions!

crawshaw (approved):
LGTM with Chris's comments!

ipn/ipnlocal/cert.go
awly: Good point. I was specifically trying to get the root CA here, but an intermediate is the actual "issuer" in this case.
And yeah, the proper way would be to  parse everything and figure out which one is leaf and which one is issuer.

Updated to use `blocks[1]`

ipn/ipnlocal/cert.go
awly: I always thought that the Go style was generally to avoid verbs unless they provide more clarity, e.g. https://pkg.go.dev/net/http#Request.Cookie vs `GetCookie`.
But this may just be my Java scars, with all those getters and setters :scream:

ipn/ipnlocal/cert.go
awly: Unfortunately `%w` is only supported in `fmt.Errorf` and not in logging packages.
It would be nice if it treated `%w` like `%v`, but it does not :)

Issue #8219, opened 2023-05-27
Unable to lower the MTU


cespare:
### What is the issue?

I mostly cannot connect to other devices over Tailscale while my internet connection is via a wifi hotspot on my phone.

I say mostly because some requests sometimes work (such as a very tiny HTTP request). But most requests time out and SSH doesn't work.

I'm 99% certain that the problem is that the MTU on the hotspot or the cell network is lower than what Tailscale is using. I've previously debugged this issue when using plain wireguard and I was able to work around it by setting a lower MTU in wireguard. In Wireguard that was trivial: its config file has a setting named `MTU`; I lowered it, restarted Wireguard, and everything worked.

In Tailscale, I've been debugging for an hour or so and I can't get it working.

The [troubleshooting guide](https://tailscale.com/kb/1023/troubleshooting/?q=mtu#unable-to-make-a-tcp-connection-between-two-nodes) says:

> In order to solve this, we can set the MTU at the LAN level to a lower value, or use MSS (Maxiumum Segment Size) clamping..

but it does not say how to do these things.

I found some various iptables rules online and tried them but none seemed to work.

I learned about the `TS_DEBUG_MTU` setting (mentioned [here](https://github.com/tailscale/tailscale/wiki/Tailscale-1.10)). I set that in my `/etc/default/tailscaled`. I ran `tailscale down` and then `tailscale up --accept routes` to reload the config. There was no apparent effect, even at very low MTU (800). But also, I don't see anything about MTU in the logs, so it's really hard to tell if it's having an effect. I even tried introducing syntax errors in that config and got no warnings, so it's really unclear that it's doing anything whatsoever.

BTW that file did exist before I edited it. It had these contents:

```
# Set the port to listen on for incoming VPN packets.
# Remote nodes will automatically be informed about the new port number,
# but you might want to configure this in order to set external firewall
# settings.
PORT="41641"

# Extra flags you might want to pass to tailscaled.
FLAGS=""
```

(Also, this is a bit off-topic, but Tailscale's "sticky flags with hidden state" method of config is already pretty surprising and unlike all the other software I use, but adding a "normal" /etc config file *on top* is extra confusing.)

Questions:

* Is `TS_DEBUG_MTU` the best way to lower the MTU?
* How do I know whether my config file is being used?
* After I change my config, what is the easiest way to get tailscale to load it?
* How can I see what MTU tailscale is using?

### Steps to reproduce

[...]

PR #8605, opened 2023-07-13
wgengine/router,ipn/ipnlocal: add MTU field to router config

valscale:
The MacOS client can't set the MTU when creating the tun due to lack of permissions, so add it to the router config and have MacOS set it in the callback using a method that it does have permissions for.

Updates #8219

### DIFF
--- a/wgengine/router/callback.go
+++ b/wgengine/router/callback.go
@@ -24,9 +24,16 @@ type CallbackRouter struct {
 	// will return ErrGetBaseConfigNotSupported.
 	GetBaseConfigFunc func() (dns.OSConfig, error)
 
-	mu   sync.Mutex    // protects all the following
-	rcfg *Config       // last applied router config
-	dcfg *dns.OSConfig // last applied DNS config
+	// InitialMTU is the MTU the tun should be initialized with.
+	// Zero means don't change the MTU from the default. This MTU
+	// is applied only once, shortly after the TUN is created, and
+	// ignored thereafter.
+	InitialMTU uint32
+
+	mu        sync.Mutex    // protects all the following
+	didSetMTU bool          // if we set the MTU already
+	rcfg      *Config       // last applied router config
+	dcfg      *dns.OSConfig // last applied DNS config
 }
 
 // Up implements Router.
@@ -40,6 +47,10 @@ func (r *CallbackRouter) Set(rcfg *Config) error {
 	defer r.mu.Unlock()
 	if r.rcfg.Equal(rcfg) {
 		return nil
+	}
+	if r.didSetMTU == false {
+		r.didSetMTU = true
+		rcfg.NewMTU = int(r.InitialMTU)
 	}
 	r.rcfg = rcfg
 	return r.SetBoth(r.rcfg, r.dcfg)
--- a/wgengine/router/router.go
+++ b/wgengine/router/router.go
@@ -67,6 +67,11 @@ type Config struct {
 	// routing rules apply.
 	LocalRoutes []netip.Prefix
 
+	// NewMTU is currently only used by the MacOS network extension
+	// app to set the MTU of the tun in the router configuration
+	// callback. If zero, the MTU is unchanged.
+	NewMTU int
+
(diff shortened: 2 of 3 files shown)


wgengine/router/router.go
```diff
@@ -71,6 +71,7 @@ type Config struct {
 	SubnetRoutes     []netip.Prefix         // subnets being advertised to other Tailscale nodes
 	SNATSubnetRoutes bool                   // SNAT traffic to local subnets
 	NetfilterMode    preftype.NetfilterMode // how much to manage netfilter rules
+	MTU              int                    // MTU
```
bradfitz, line 74: document what zero means? unchanged? some default?

valscale: Thanks, fixed.

valscale:
I changed the behavior so that it only sets the MTU on the first call to setRoutes after the tun is created, so that you can still set the MTU by ifconfig and have it stick, since this is useful in a bunch of different scenarios.

wgengine/router/callback.go
```diff
@@ -24,9 +24,17 @@ type CallbackRouter struct {
...
-	dcfg *dns.OSConfig // last applied DNS config
+	mu sync.Mutex // protects all the following
+
+	// InitialMTU is the MTU the tun should be initialized with.
+	// Zero means don't change the MTU. Once set, this field is
+	// zeroed out. This allows the MTU to be changed after
+	// initialization by other means (such as ifconfig). This
+	// field should be set only on creation and zeroed after the
+	// first use while holding mu.
+	initialMTU uint32
```
andrew-d, line 35: If this isn't a public field, I think that the macOS/iOS code can't set this, so it'll always be zero (below)?

wgengine/router/callback.go
```diff
@@ -41,6 +49,10 @@ func (r *CallbackRouter) Set(rcfg *Config) error {
 	if r.rcfg.Equal(rcfg) {
 		return nil
 	}
+	if r.initialMTU != 0 {
```
andrew-d, line 52: (this will always be `0` if `r.initialMTU` is never initialized)

wgengine/router/callback.go
valscale: Oh right, the user is in corp... reverting this unexport and adding a comment so no one else gets confused.

wgengine/router/callback.go
valscale: Fixed.

wgengine/router/callback.go
```diff
@@ -41,6 +50,10 @@ func (r *CallbackRouter) Set(rcfg *Config) error {
 	if r.rcfg.Equal(rcfg) {
 		return nil
 	}
+	if r.InitialMTU != 0 {
+		rcfg.NewMTU = int(r.InitialMTU)
+		r.InitialMTU = 0
+	}
```
andrew-d, line 56: `r.InitialMTU` is mutated but not guarded by `r.mu` here. And `Set` can be called concurrently, so we do need to consider this.

Also now that I think about it, probably worth mentioning in the comment on the `InitialMTU` field that the caller should never mutate it after creating the `CallbackRouter` struct.

andrew-d (commented):
I think this is one of the last things that I can see for this review; I'm going to re-read the corp code shortly to confirm that I understand what it's doing, just to be sure.

wgengine/router/callback.go
andrew-d: I guess the other option (up to you, no preference here) is to have `r.InitialMTU` be configuration that's only set at creation and never modified, and add a new `r.didSetMTU` `bool` that is protected by the mutex and you use to avoid setting twice.  That way it's ~impossible for a caller to accidentally cause a race condition.

valscale: Huh, I think line 48 is r.mu.Lock() so r.InitialMTU is mutated only under the lock. But the flag version is better anyway. I didn't like exporting a thing protected by an unexported lock.

andrew-d (approved):

PR #8607, opened 2023-07-13
logtail: use tstime

clairew:
Updates #8587
Signed-off-by: Claire Wang <claire@tailscale.com>

### DIFF
--- a/logtail/logtail_test.go
+++ b/logtail/logtail_test.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"tailscale.com/tstest"
+	"tailscale.com/tstime"
 )
 
 func TestFastShutdown(t *testing.T) {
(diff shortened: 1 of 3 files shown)


dsnet (approved):
My apologies for the delay, I was OOO end of last week.

logtail/logtail_test.go
```diff
@@ -363,7 +363,7 @@ func TestEncode(t *testing.T) {
 	for _, tt := range tests {
 		buf := new(simpleMemBuf)
 		lg := &Logger{
-			timeNow:      func() time.Time { return time.Unix(123, 456).UTC() },
+			clock:        tstest.NewClock(tstest.ClockOpts{Start: time.Unix(123, 456).UTC()}),
```
sailorfrag, line 366: In the spirit of not updating tests that don't actually care about controlling time: keep this one, the others can be left as `tstime.StdClock{}`.

Or go ahead unchanged. _Something_ needed to be specified in the other tests so I guess installing a fake clock is as good a plan as any.

sailorfrag (approved):

logtail/logtail_test.go
clairew: the clock is used in this test in the encodeText method in logtail

PR #8611, opened 2023-07-13
net/art: implement the path compression optimization

danderson:
Several individual changes. Commits 2 and 3 are the meat of the optimization's implementation, the rest are small fixups and lots of tests.

This implementation currently passes all tests with -count=1000 reliably, so this _appears_ to be correct code. I still need to run benchmarks again, I'll post those in a comment later.

### DIFF
--- a/net/art/stride_table.go
+++ b/net/art/stride_table.go
@@ -85,16 +85,12 @@ func (t *strideTable[T]) deleteChild(idx int) {
 
 // setChild replaces the child strideTable for addr (if any) with child.
 func (t *strideTable[T]) setChild(addr uint8, child *strideTable[T]) {
-	idx := hostIndex(addr)
-	if t.entries[idx].child == nil {
-		t.childRefs++
-	}
-	t.entries[idx].child = child
+	t.setChildByIndex(hostIndex(addr), child)
 }
 
-// setChildByIdx replaces the child strideTable at idx (if any) with
+// setChildByIndex replaces the child strideTable at idx (if any) with
 // child. idx should be obtained via a call to getChild.
-func (t *strideTable[T]) setChildByIdx(idx int, child *strideTable[T]) {
+func (t *strideTable[T]) setChildByIndex(idx int, child *strideTable[T]) {
 	if t.entries[idx].child == nil {
 		t.childRefs++
 	}
--- a/net/art/stride_table.go
+++ b/net/art/stride_table.go
@@ -175,7 +175,9 @@ func (t *strideTable[T]) insert(addr uint8, prefixLen int, val *T) {
 	return
 }
 
-// delete removes the route addr/prefixLen from t.
+// delete removes the route addr/prefixLen from t. Returns the value
+// that was associated with the deleted prefix, or nil if the prefix
+// wasn't in the strideTable.
 func (t *strideTable[T]) delete(addr uint8, prefixLen int) *T {
 	idx := prefixIndex(addr, prefixLen)
 	recordedIdx := t.entries[idx].prefixIndex
--- a/net/art/stride_table.go
+++ b/net/art/stride_table.go
@@ -13,6 +13,11 @@ import (
 	"strings"
 )
 
+const (
+	debugStrideInsert = false
+	debugStrideDelete = false
+)
+
 // strideEntry is a strideTable entry.
 type strideEntry[T any] struct {
 	// prefixIndex is the prefixIndex(...) value that caused this stride entry's
--- a/net/art/stride_table.go
+++ b/net/art/stride_table.go
@@ -227,7 +227,7 @@ func (t *strideTable[T]) treeDebugString() string {
 func (t *strideTable[T]) treeDebugStringRec(w io.Writer, idx, indent int) {
 	addr, len := inversePrefixIndex(idx)
 	if t.entries[idx].prefixIndex != 0 && t.entries[idx].prefixIndex == idx {
-		fmt.Fprintf(w, "%s%d/%d (%d/%d) = %v\n", strings.Repeat(" ", indent), addr, len, addr, len, *t.entries[idx].value)
+		fmt.Fprintf(w, "%s%d/%d (%02x/%d) = %v\n", strings.Repeat(" ", indent), addr, len, addr, len, *t.entries[idx].value)
 		indent += 2
 	}
 	if idx >= firstHostIndex {
--- a/net/art/stride_table.go
+++ b/net/art/stride_table.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"math/bits"
+	"net/netip"
 	"strconv"
 	"strings"
 )
@@ -32,6 +33,9 @@ type strideEntry[T any] struct {
 // The leaves of the binary tree are host routes (/8s). Each parent is a
 // successively larger prefix that encompasses its children (/7 through /0).
 type strideTable[T any] struct {
+	// prefix is the prefix represented by the 0/0 route of this strideTable. It
+	// is used in multi-level tables to support path compression.
+	prefix netip.Prefix
 	// entries is the nodes of the binary tree, laid out in a flattened array.
 	//
 	// The array indices are arranged by the prefixIndex function, such that the
@@ -76,7 +80,9 @@ func (t *strideTable[T]) deleteChild(idx int) {
 func (t *strideTable[T]) getOrCreateChild(addr uint8) *strideTable[T] {
 	idx := hostIndex(addr)
 	if t.entries[idx].child == nil {
-		t.entries[idx].child = new(strideTable[T])
+		t.entries[idx].child = &strideTable[T]{
+			prefix: childPrefixOf(t.prefix, addr),
+		}
 		t.refs++
 	}
 	return t.entries[idx].child
@@ -228,4 +234,30 @@ func formatPrefixTable(addr uint8, len int) string {
 		return "<nil>"
 	}
 	return fmt.Sprintf("%3d/%d", addr, len)
+}
+
+// childPrefixOf returns the child prefix of parent whose final byte
+// is stride. The parent prefix must be byte-aligned
+// (i.e. parent.Bits() must be a multiple of 8), and be no more
+// specific than /24 for IPv4 or /120 for IPv6.
+//
+// For example, childPrefixOf("192.168.0.0/16", 8) == "192.168.8.0/24".
+func childPrefixOf(parent netip.Prefix, stride uint8) netip.Prefix {
+	l := parent.Bits()
+	if l%8 != 0 {
+		panic("parent prefix is not 8-bit aligned")
+	}
+	if l >= parent.Addr().BitLen() {
+		panic("parent prefix cannot be extended further")
+	}
+	off := l / 8
+	if parent.Addr().Is4() {
+		bs := parent.Addr().As4()
+		bs[off] = stride
+		return netip.PrefixFrom(netip.AddrFrom4(bs), l+8)
+	} else {
+		bs := parent.Addr().As16()
+		bs[off] = stride
+		return netip.PrefixFrom(netip.AddrFrom16(bs), l+8)
+	}
 }
--- a/net/art/stride_table.go
+++ b/net/art/stride_table.go
@@ -119,8 +119,8 @@ func (t *strideTable[T]) getValAndChild(addr uint8) (*T, *strideTable[T]) {
 	return t.entries[idx].value, t.entries[idx].child
 }
 
-// findFirstChild returns the first non-nil child strideTable in t, or
-// nil if t has no children.
+// findFirstChild returns the first child strideTable in t, or nil if
+// t has no children.
 func (t *strideTable[T]) findFirstChild() *strideTable[T] {
 	for i := firstHostIndex; i <= lastHostIndex; i++ {
 		if child := t.entries[i].child; child != nil {
--- a/net/art/stride_table.go
+++ b/net/art/stride_table.go
@@ -38,8 +38,11 @@ type strideEntry[T any] struct {
 // The leaves of the binary tree are host routes (/8s). Each parent is a
 // successively larger prefix that encompasses its children (/7 through /0).
 type strideTable[T any] struct {
-	// prefix is the prefix represented by the 0/0 route of this strideTable. It
-	// is used in multi-level tables to support path compression.
+	// prefix is the prefix represented by the 0/0 route of this
+	// strideTable. It is used in multi-level tables to support path
+	// compression. All strideTables must have a valid prefix
+	// (non-zero value, passes IsValid()) whose length is a multiple
+	// of 8 (e.g. /8, /16, but not /15).
 	prefix netip.Prefix
 	// entries is the nodes of the binary tree, laid out in a flattened array.
 	//
--- a/net/art/stride_table.go
+++ b/net/art/stride_table.go
@@ -48,10 +48,10 @@ type strideTable[T any] struct {
 	// memory trickery to store the refcount, but this is Go, where we don't
 	// store random bits in pointers lest we confuse the GC)
 	entries [lastHostIndex + 1]strideEntry[T]
-	// refs is the number of route entries and child strideTables referenced by
-	// this table. It is used in the multi-layered logic to determine when this
-	// table is empty and can be deleted.
-	refs int
+	// routeRefs is the number of route entries in this table.
+	routeRefs uint16
+	// childRefs is the number of child strideTables referenced by this table.
+	childRefs uint16
 }
 
 const (
@@ -72,25 +72,58 @@ func (t *strideTable[T]) getChild(addr uint8) (child *strideTable[T], idx int) {
 // obtained via a call to getChild.
 func (t *strideTable[T]) deleteChild(idx int) {
 	t.entries[idx].child = nil
-	t.refs--
+	t.childRefs--
+}
+
+// setChild replaces the child strideTable for addr (if any) with child.
+func (t *strideTable[T]) setChild(addr uint8, child *strideTable[T]) {
+	idx := hostIndex(addr)
+	if t.entries[idx].child == nil {
+		t.childRefs++
+	}
+	t.entries[idx].child = child
+}
+
+// setChildByIdx replaces the child strideTable at idx (if any) with
+// child. idx should be obtained via a call to getChild.
+func (t *strideTable[T]) setChildByIdx(idx int, child *strideTable[T]) {
+	if t.entries[idx].child == nil {
+		t.childRefs++
+	}
+	t.entries[idx].child = child
 }
 
 // getOrCreateChild returns the child strideTable for addr, creating it if
 // necessary.
-func (t *strideTable[T]) getOrCreateChild(addr uint8) *strideTable[T] {
+func (t *strideTable[T]) getOrCreateChild(addr uint8) (child *strideTable[T], created bool) {
 	idx := hostIndex(addr)
 	if t.entries[idx].child == nil {
 		t.entries[idx].child = &strideTable[T]{
 			prefix: childPrefixOf(t.prefix, addr),
 		}
-		t.refs++
+		t.childRefs++
+		return t.entries[idx].child, true
 	}
-	return t.entries[idx].child
+	return t.entries[idx].child, false
 }
 
+// getValAndChild returns both the prefix and child strideTable for
+// addr. Both returned values can be nil if no entry of that type
+// exists for addr.
 func (t *strideTable[T]) getValAndChild(addr uint8) (*T, *strideTable[T]) {
 	idx := hostIndex(addr)
 	return t.entries[idx].value, t.entries[idx].child
+}
+
+// findFirstChild returns the first non-nil child strideTable in t, or
+// nil if t has no children.
+func (t *strideTable[T]) findFirstChild() *strideTable[T] {
+	for i := firstHostIndex; i <= lastHostIndex; i++ {
+		if child := t.entries[i].child; child != nil {
+			return child
+		}
+	}
+	return nil
 }
 
 // allot updates entries whose stored prefixIndex matches oldPrefixIndex, in the
@@ -133,7 +166,7 @@ func (t *strideTable[T]) insert(addr uint8, prefixLen int, val *T) {
 	if oldIdx != idx {
 		// This route entry was freshly created (not just updated), that's a new
 		// reference.
-		t.refs++
+		t.routeRefs++
 	}
 	return
 }
@@ -150,7 +183,7 @@ func (t *strideTable[T]) delete(addr uint8, prefixLen int) *T {
 
 	parentIdx := idx >> 1
 	t.allot(idx, idx, t.entries[parentIdx].prefixIndex, t.entries[parentIdx].value)
-	t.refs--
+	t.routeRefs--
 	return val
 }
 
--- a/net/art/table.go
+++ b/net/art/table.go
@@ -468,7 +468,7 @@ func (t *Table[T]) Delete(pfx netip.Prefix) {
 			if debugDelete {
 				fmt.Printf("delete: compact parent.prefix=%s st.prefix=%s child.prefix=%s\n", parent.prefix, cur.prefix, child.prefix)
 			}
-			strideTables[strideIdx-1].setChildByIdx(strideIndexes[strideIdx-1], child)
+			strideTables[strideIdx-1].setChildByIndex(strideIndexes[strideIdx-1], child)
 			return
 		default:
 			// This table has two or more children, so it's acting as a "fork in
--- a/net/art/table.go
+++ b/net/art/table.go
@@ -45,13 +45,25 @@ func (t *Table[T]) init() {
 // no route matched.
 func (t *Table[T]) Get(addr netip.Addr) *T {
 	t.init()
-	st := &t.v4
+
+	// Ideally we would use addr.AsSlice here, but AsSlice is just
+	// barely complex enough that it can't be inlined, and that in
+	// turn causes the slice to escape to the heap. Using As16 and
+	// manual slicing here helps the compiler keep Get alloc-free.
+	rawAddr := addr.As16()
+	var (
+		st *strideTable[T]
+		bs []byte
+	)
 	if addr.Is6() {
 		st = &t.v6
+		bs = rawAddr[:]
+	} else {
+		st = &t.v4
+		bs = rawAddr[12:]
 	}
 
 	i := 0
-	bs := addr.AsSlice()
 	// With path compression, we might skip over some address bits while walking
 	// to a strideTable leaf. This means the leaf answer we find might not be
 	// correct, because path compression took us down the wrong subtree. When
--- a/net/art/table.go
+++ b/net/art/table.go
@@ -14,6 +14,7 @@ package art
 
 import (
 	"bytes"
+	"encoding/binary"
 	"fmt"
 	"io"
 	"math/bits"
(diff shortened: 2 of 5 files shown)


net/art/table.go
```diff
@@ -18,17 +18,27 @@ import (
...
+	v4       strideTable[T]
+	v6       strideTable[T]
+	initOnce sync.Once
+}
+
+func (t *Table[T]) init() {
+	t.initOnce.Do(func() {
+		t.v4.prefix = netip.PrefixFrom(netip.IPv4Unspecified(), 0)
+		t.v6.prefix = netip.PrefixFrom(netip.IPv6Unspecified(), 0)
+	})
```
bradfitz, line 41: I'm pretty sure this allocates on on 2nd+ calls since the func closes over the receiver t.

using a method value wouldn't:

```go
func (t *Table[T]) init() {  t.initOnce.Do(t.doInit) }
func (t *Table[T]) doInit() {
  t.v4.prefix = netip.PrefixFrom(netip.IPv4Unspecified(), 0)
  t.v6.prefix = netip.PrefixFrom(netip.IPv6Unspecified(), 0)
}
```

net/art/table.go
```diff
@@ -156,20 +488,152 @@ func (t *Table[T]) debugSummary() string {
...
+func ipv4AsUint(ip netip.Addr) uint32 {
+	bs := ip.As4()
+	return uint32(bs[0])<<24 | uint32(bs[1])<<16 | uint32(bs[2])<<8 | uint32(bs[3])
+}
+
+// ipv6AsUint returns ip as a pair of uint64s.
+func ipv6AsUint(ip netip.Addr) (uint64, uint64) {
+	bs := ip.As16()
+	hi := uint64(bs[0])<<56 | uint64(bs[1])<<48 | uint64(bs[2])<<40 | uint64(bs[3])<<32 | uint64(bs[4])<<24 | uint64(bs[5])<<16 | uint64(bs[6])<<8 | uint64(bs[7])
+	lo := uint64(bs[8])<<56 | uint64(bs[9])<<48 | uint64(bs[10])<<40 | uint64(bs[11])<<32 | uint64(bs[12])<<24 | uint64(bs[13])<<16 | uint64(bs[14])<<8 | uint64(bs[15])
```
bradfitz, line 637: if you want, encoding/binary.BigEndian.Uint64(bs[:8]) etc should compile down to exactly the same I believe (`binary.BigEndian` is a concrete unexported type specifically so the compiler can inline and not worry about seeing through an interface)

bradfitz (commented):
Wow, that's a lot! Nice tests.

I don't remember any of the path compression bits of the paper anymore (will have to go re-read) but here are some superficial comments meanwhile.

danderson:
Benchmarks are below. They construct random route tables of various sizes, and measure performance. Short 
 - IPv4 (IPv6) route insertion is ~30% (~80%) faster, uses ~30% (~80%) less memory.
 - IPv4 (IPv6) route deletion is ~20% slower (~30-60% faster), no change in memory.
 - Route lookup is ~25% slower across the board, although still very fast (12-25 million lookups/sec, depending on test conditions). Memory behavior unchanged.

As expected, the major gain is that IPv6 routes now cost approximately as much IPv4 routes (~10-15kiB per route), instead of consuming 6x more memory for the same amount of information.

There's still opportunities to optimize further (removing the one alloc in the route lookup path is trivial, another more complex optimization could halve the memory cost per route across the board). But assuming this code is correct, it no longer has a completely pathological performance envelope.

```
goos: linux
goarch: amd64
pkg: tailscale.com/net/art
cpu: AMD Ryzen 9 5950X 16-Core Processor
                          │  before2.txt  │              after2.txt              │
                          │    sec/op     │    sec/op     vs base                │
TableInsertion/ipv4/10      1091.5n ±  1%   787.2n ±  1%  -27.88% (p=0.000 n=10)
TableInsertion/ipv4/100     2701.5n ±  4%   923.6n ±  0%  -65.81% (p=0.000 n=10)
TableInsertion/ipv4/1000     2.095µ ±  2%   1.678µ ±  2%  -19.93% (p=0.000 n=10)
TableInsertion/ipv4/10000    2.830µ ±  7%   2.083µ ±  5%  -26.40% (p=0.000 n=10)
TableInsertion/ipv6/10       4.266µ ±  6%   1.112µ ±  1%  -73.93% (p=0.000 n=10)
TableInsertion/ipv6/100     11.154µ ±  2%   1.307µ ±  9%  -88.28% (p=0.000 n=10)
TableInsertion/ipv6/1000    15.263µ ±  4%   2.304µ ±  4%  -84.90% (p=0.000 n=10)
TableInsertion/ipv6/10000   15.166µ ±  5%   2.947µ ±  5%  -80.57% (p=0.000 n=10)
TableDelete/ipv4/10          218.8n ±  2%   292.3n ±  6%  +33.59% (p=0.000 n=10)
TableDelete/ipv4/100         367.3n ±  1%   275.2n ±  2%  -25.07% (p=0.000 n=10)
TableDelete/ipv4/1000        349.5n ±  3%   406.7n ±  2%  +16.35% (p=0.000 n=10)
TableDelete/ipv4/10000       602.8n ±  7%   591.7n ±  5%        ~ (p=0.796 n=10)
TableDelete/ipv6/10          666.8n ±  5%   450.9n ±  5%  -32.37% (p=0.000 n=10)
TableDelete/ipv6/100        1008.5n ±  5%   174.0n ±  2%  -82.75% (p=0.000 n=10)
TableDelete/ipv6/1000       1436.5n ±  5%   507.7n ±  3%  -64.66% (p=0.000 n=10)
TableDelete/ipv6/10000      1700.0n ±  7%   730.6n ± 13%  -57.02% (p=0.000 n=10)
TableGet/ipv4/10             33.63n ±  3%   40.85n ±  2%  +21.47% (p=0.000 n=10)
TableGet/ipv4/100            38.01n ±  2%   50.99n ±  2%  +34.15% (p=0.000 n=10)
TableGet/ipv4/1000           45.09n ±  2%   61.57n ±  2%  +36.56% (p=0.000 n=10)
TableGet/ipv4/10000          50.44n ± 12%   68.38n ± 11%  +35.54% (p=0.000 n=10)
TableGet/ipv6/10             41.80n ±  1%   45.82n ±  1%   +9.62% (p=0.000 n=10)
TableGet/ipv6/100            44.28n ±  1%   57.59n ±  1%  +30.04% (p=0.000 n=10)
TableGet/ipv6/1000           48.82n ±  6%   65.79n ±  3%  +34.76% (p=0.000 n=10)
TableGet/ipv6/10000          67.26n ± 12%   83.48n ±  2%  +24.12% (p=0.000 n=10)
geomean                      509.7n         324.0n        -36.43%

                          │  before2.txt  │               after2.txt               │
                          │   avg-B/op    │   avg-B/op    vs base                  │
TableInsertion/ipv4/10       18.55Ki ± 0%   10.62Ki ± 0%  -42.78% (p=0.000 n=10)
TableInsertion/ipv4/100      19.88Ki ± 0%   11.00Ki ± 0%  -44.65% (p=0.000 n=10)
TableInsertion/ipv4/1000     14.04Ki ± 0%   10.90Ki ± 0%  -22.38% (p=0.000 n=10)
TableInsertion/ipv4/10000   12.904Ki ± 0%   8.995Ki ± 0%  -30.29% (p=0.000 n=10)
TableInsertion/ipv6/10       70.24Ki ± 0%   15.93Ki ± 0%  -77.32% (p=0.000 n=10)
TableInsertion/ipv6/100      94.49Ki ± 0%   13.93Ki ± 0%  -85.26% (p=0.000 n=10)
TableInsertion/ipv6/1000     93.19Ki ± 0%   15.04Ki ± 0%  -83.86% (p=0.000 n=10)
TableInsertion/ipv6/10000    90.96Ki ± 0%   13.19Ki ± 0%  -85.50% (p=0.000 n=10)
TableDelete/ipv4/10            3.200 ± 0%     3.200 ± 0%        ~ (p=1.000 n=10)
TableDelete/ipv4/100           4.030 ± 0%     3.930 ± 0%   -2.48% (p=0.000 n=10)
TableDelete/ipv4/1000          3.990 ± 0%     3.990 ± 0%        ~ (p=1.000 n=10) ¹
TableDelete/ipv4/10000         4.000 ± 0%     4.000 ± 0%        ~ (p=1.000 n=10) ¹
TableDelete/ipv6/10            16.00 ± 0%     16.00 ± 0%        ~ (p=1.000 n=10) ¹
TableDelete/ipv6/100           16.00 ± 0%     15.84 ± 0%   -1.00% (p=0.000 n=10)
TableDelete/ipv6/1000          16.00 ± 0%     15.98 ± 0%   -0.12% (p=0.000 n=10)
TableDelete/ipv6/10000         16.00 ± 0%     16.00 ± 0%        ~ (p=1.000 n=10) ¹
geomean                        545.4          311.7       -42.86%
¹ all samples are equal

                          │  before2.txt  │               after2.txt                │
                          │ avg-allocs/op │ avg-allocs/op  vs base                  │
TableInsertion/ipv4/10         2.400 ± 0%      1.700 ± 0%  -29.17% (p=0.000 n=10)
TableInsertion/ipv4/100        2.500 ± 0%      1.810 ± 0%  -27.60% (p=0.000 n=10)
TableInsertion/ipv4/1000       2.060 ± 0%      1.820 ± 0%  -11.65% (p=0.000 n=10)
TableInsertion/ipv4/10000      1.970 ± 0%      1.680 ± 0%  -14.72% (p=0.000 n=10)
TableInsertion/ipv6/10         6.300 ± 0%      2.100 ± 0%  -66.67% (p=0.000 n=10)
TableInsertion/ipv6/100        8.130 ± 0%      2.040 ± 0%  -74.91% (p=0.000 n=10)
TableInsertion/ipv6/1000       8.030 ± 0%      2.130 ± 0%  -73.47% (p=0.000 n=10)
TableInsertion/ipv6/10000      7.860 ± 0%      1.990 ± 0%  -74.68% (p=0.000 n=10)
TableDelete/ipv4/10            1.000 ± 0%      1.000 ± 0%        ~ (p=1.000 n=10) ¹
TableDelete/ipv4/100         1000.0m ± 0%     990.0m ± 0%   -1.00% (p=0.000 n=10)
TableDelete/ipv4/1000          1.000 ± 0%      1.000 ± 0%        ~ (p=1.000 n=10) ¹
TableDelete/ipv4/10000         1.000 ± 0%      1.000 ± 0%        ~ (p=1.000 n=10) ¹
TableDelete/ipv6/10            1.000 ± 0%      1.000 ± 0%        ~ (p=1.000 n=10) ¹
TableDelete/ipv6/100         1000.0m ± 0%     990.0m ± 0%   -1.00% (p=0.000 n=10)
TableDelete/ipv6/1000          1.000 ± 0%      1.000 ± 0%        ~ (p=1.000 n=10) ¹
TableDelete/ipv6/10000         1.000 ± 0%      1.000 ± 0%        ~ (p=1.000 n=10) ¹
geomean                        2.023           1.377       -31.92%
¹ all samples are equal

                          │ before2.txt │               after2.txt               │
                          │  routes/s   │   routes/s     vs base                 │
TableInsertion/ipv4/10      916.3k ± 1%   1270.3k ±  1%   +38.64% (p=0.000 n=10)
TableInsertion/ipv4/100     370.2k ± 4%   1082.8k ±  0%  +192.49% (p=0.000 n=10)
TableInsertion/ipv4/1000    477.4k ± 2%    596.1k ±  2%   +24.86% (p=0.000 n=10)
TableInsertion/ipv4/10000   353.3k ± 7%    480.1k ±  5%   +35.87% (p=0.000 n=10)
TableInsertion/ipv6/10      234.4k ± 6%    899.0k ±  1%  +283.45% (p=0.000 n=10)
TableInsertion/ipv6/100     89.66k ± 2%   765.16k ±  8%  +753.41% (p=0.000 n=10)
TableInsertion/ipv6/1000    65.52k ± 4%   434.02k ±  4%  +562.45% (p=0.000 n=10)
TableInsertion/ipv6/10000   65.94k ± 5%   339.30k ±  5%  +414.59% (p=0.000 n=10)
TableDelete/ipv4/10         4.570M ± 2%    3.421M ±  6%   -25.15% (p=0.000 n=10)
TableDelete/ipv4/100        2.723M ± 1%    3.634M ±  3%   +33.46% (p=0.000 n=10)
TableDelete/ipv4/1000       2.861M ± 3%    2.459M ±  2%   -14.05% (p=0.000 n=10)
TableDelete/ipv4/10000      1.659M ± 7%    1.690M ±  5%         ~ (p=0.796 n=10)
TableDelete/ipv6/10         1.500M ± 6%    2.218M ±  4%   +47.85% (p=0.000 n=10)
TableDelete/ipv6/100        991.6k ± 6%   5746.5k ±  2%  +479.49% (p=0.000 n=10)
TableDelete/ipv6/1000       696.2k ± 5%   1969.7k ±  3%  +182.91% (p=0.000 n=10)
TableDelete/ipv6/10000      588.4k ± 8%   1368.8k ± 12%  +132.61% (p=0.000 n=10)
geomean                     584.8k         1.305M        +123.17%

                    │ before2.txt  │              after2.txt              │
                    │   addrs/s    │   addrs/s     vs base                │
TableGet/ipv4/10      29.74M ±  3%   24.48M ±  2%  -17.68% (p=0.000 n=10)
TableGet/ipv4/100     26.31M ±  2%   19.61M ±  2%  -25.46% (p=0.000 n=10)
TableGet/ipv4/1000    22.18M ±  2%   16.24M ±  2%  -26.77% (p=0.000 n=10)
TableGet/ipv4/10000   19.82M ± 10%   14.63M ± 10%  -26.22% (p=0.000 n=10)
TableGet/ipv6/10      23.92M ±  1%   21.82M ±  1%   -8.78% (p=0.000 n=10)
TableGet/ipv6/100     22.58M ±  1%   17.36M ±  1%  -23.11% (p=0.000 n=10)
TableGet/ipv6/1000    20.48M ±  6%   15.20M ±  3%  -25.80% (p=0.000 n=10)
TableGet/ipv6/10000   14.90M ± 11%   11.98M ±  2%  -19.58% (p=0.000 n=10)
geomean               22.09M         17.26M        -21.88%

                    │ before2.txt │             after2.txt              │
                    │    B/op     │    B/op     vs base                 │
TableGet/ipv4/10       4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv4/100      4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv4/1000     4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv4/10000    4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv6/10       16.00 ± 0%   16.00 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv6/100      16.00 ± 0%   16.00 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv6/1000     16.00 ± 0%   16.00 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv6/10000    16.00 ± 0%   16.00 ± 0%       ~ (p=1.000 n=10) ¹
geomean                8.000        8.000       +0.00%
¹ all samples are equal

                    │ before2.txt │             after2.txt              │
                    │  allocs/op  │ allocs/op   vs base                 │
TableGet/ipv4/10       1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv4/100      1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv4/1000     1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv4/10000    1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv6/10       1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv6/100      1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv6/1000     1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=10) ¹
TableGet/ipv6/10000    1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                1.000        1.000       +0.00%
¹ all samples are equal
```

net/art/table.go
danderson: It doesn't _appear_ to allocate, benchmarks say there's 1 alloc per Get, and I believe it's the `bs := addr.AsSlice()` a few lines below there. But I'll refactor and verify.

net/art/table.go
danderson: I thought from past optimization work that binary.BigEndian specifically did _not_ trigger a bunch of optimizations that were possible (broke inlining or simplification or something?). I remember either Josh or @dsnet filing Go compiler bugs about it?

Also easy to A/B test if I can't find the upstream bug.

net/art/table.go
danderson: Did this refactor. Benchmarks show no change in allocs/op for any of insert/delete/get (still ~2 allocs/insert, 1 alloc/delete, 1 alloc/get). I don't mind keeping or discarding it, it seems to have no significant impact.

net/art/table.go
danderson: Couldn't find the bug I was thinking of. Benchmarks inconclusive, some shapes of route table got a bit faster, some got a bit slower, allocs moved around by 0-1% in either direction. I'm not sure how to interpret that, but I'm thinking that means the change is in the noise compared to the variance of my machine while it's running the bench.

sailorfrag: Having recently investigated `binary.BigEndian` and `binary.LittleEndian`, the code behind `binary.BigEndian` looks basically the same as what you have written there, and then the Go compiler optimizes it into efficient instructions if they exist. It takes a slice instead of an array and then has some magic syntax to ensure that only one length check occurs; here you have an array so can bypass that. I wouldn't be surprised if the resulting assembly is pretty similar with both implementations.

bradfitz:
I don't know why, but here's a test that sometimes fails, assuming the test is valid:

```go
func TestStress(t *testing.T) {
	var tab Table[string]

	rand.Seed(time.Now().UnixNano())
	const N = 10
	var prefixes []netip.Prefix
	for i := 0; i < N; i++ {
		prefixes = append(prefixes, netip.PrefixFrom(
			netip.AddrFrom4([4]byte{
				byte(rand.Intn(256)),
				byte(rand.Intn(256)),
				byte(rand.Intn(256)),
				byte(rand.Intn(256)),
			}), rand.Intn(31)+1).Masked())
	}
	var want []string
	for i, p := range prefixes {
		want = append(want, tab.debugSummary())
		tab.Insert(p, ptr.To(fmt.Sprintf("at-%d", i)))
	}

	for i := len(prefixes) - 1; i >= 0; i-- {
		tab.Delete(prefixes[i])
		if got := tab.debugSummary(); got != want[i] {
			t.Fatalf("after delete %d, mismatch:\n\n got: %s\n\nwant: %s", i, got, want[i])
		}
	}
}
```

That is, add a bunch of random routes, remembering the table's internal state at each point, then remove them in order and make sure the table still has the same state at each point as it had as it was growing. Is that a valid assumption?

sailorfrag (approved):
OK I admit to not having gone to the level of explicitly understanding the full process but I read through the paper and the comment narration and everything seemed to make sense along the way.

It seems to me like there may be some ability to get better compiler optimizations by splitting the IPv4 and IPv6 paths more but that would result in some duplication and probably harder to read code, so that seems like something to only do if profiling suggests it's needed to achieve some performance goal.

danderson:
Hmm. Assuming the test does what you describe, that is a valid assumption. It's possible debugSummary's output is not quite stable, but I would have expected that test to pass. I'll add it to the suite and investigate.

net/art/table.go
bradfitz: Huh, cool. I'm curious now when this improved. I'd stick with your prettier code if it doesn't allocate.

danderson:
TestStress failures are real. Digging.

danderson:
Ah, never mind, phew. @bradfitz you have to dedup the prefixes you add, otherwise very infrequently you end up inserting the same prefix twice. The second insert replaces the first insert's state, and so when you delete that second insert's state, you end up effectively deleting both the inserts - which is correct semantically, but the test as written fails since the table state no longer shows the between-dupe-inserts state.

bradfitz:
Ah, whoops. Phew!

danderson:
Switching TestStress to use the `randomPrefixes4(N)` helper, which returns N distinct prefixes, TestStress passes with -count=10000. That test looks reasonable to have, I'll add another commit to the PR to include it (and make it run IPv6 as well).

crawshaw:
As discussed, I had a go at fuzzing. You can see it on the branch [crawshaw/art-table](https://github.com/tailscale/tailscale/tree/crawshaw/art-table). It almost instantly produced:

```
$ go test -fuzz=FuzzInsertDelete ./net/art
fuzz: elapsed: 0s, gathering baseline coverage: 0/27 completed
fuzz: elapsed: 0s, gathering baseline coverage: 27/27 completed, now fuzzing with 10 workers
fuzz: elapsed: 2s, execs: 115133 (58452/sec), new interesting: 12 (total: 39)
--- FAIL: FuzzInsertDelete (1.97s)
    --- FAIL: FuzzInsertDelete (0.00s)
        table_test.go:701: sequence (3 steps):
        table_test.go:703:      insert(48.48.48.48/0, 3472328296227680304)
        table_test.go:703:      delete(48.48.48.49/0)
        table_test.go:703:      get(48.48.48.48)
        table_test.go:732: get("48.48.48.48") = 0x0, want 0x14007df8898
    
    Failing input written to testdata/fuzz/FuzzInsertDelete/c3bf922077abcc0d
    To re-run:
    go test -run=FuzzInsertDelete/c3bf922077abcc0d
FAIL
exit status 1
FAIL    tailscale.com/net/art   4.764s
```

This appears to be a legitimate semantic difference between `Table` and `slowPrefixTable` but I don't think it's a bug in Table? In fact, I think you could claim it's a bug in slowPrefixTable?

But either way, it found something very quickly, and that's with very weak input test cases. So it seems worth exploring.

Let me know how you want to go about dealing with this error.

crawshaw:
There is a more interesting variation of this with `/3`:

```
insert(48.48.48.48/3, &val)
delete(48.48.48.49/3)
get(48.48.48.48)  // mismatch
```

crawshaw:
For those interested in the fuzzing machinery on that branch: I manually wrote a lot of code that can be entirely machine-generated. Tracing tests to collect instruction streams, designing a wire encoding of that instruction stream, and plugging that into a direct fuzzer is entirely automatable (and has been talked about in the fuzzing world for a long time). Looking at the comments on the `Fuzz` method suggests the Go team are aware of this and might one day expand it to complex types, which is a part of that. But for now fuzzing requires a lot of boilerplate.

crawshaw:
Another bug in slowPrefixTable the fuzzer found, updating the value in a copy of the prefix not the underlying value:

```
@@ -1097,9 +1304,10 @@ func (t *slowPrefixTable[T]) delete(pfx netip.Prefix) {
 }
 
 func (t *slowPrefixTable[T]) insert(pfx netip.Prefix, val *T) {
-       for _, ent := range t.prefixes {
+       pfx = pfx.Masked()
+       for i, ent := range t.prefixes {
                if ent.pfx == pfx {
-                       ent.val = val
+                       t.prefixes[i].val = val
                        return
                }
        }
```

danderson:
The issues with slowPrefixTable make sense. In early implementation, net/art implicitly required well-formed masked prefixes, and so the reference code is implemented to that assumption (including the test data generators). Later on I made the main implementation mask its inputs, since failing to do so invalidates the algorithms in use. Guess I forgot to do that to the test implementation as well.

I'll take a closer look at the fuzzing code tomorrow. As you say, there's going to be duplication between "classical" tests and fuzz tests. Not sure how to reconcile that, but failing a better plan I would err on the side of running duplicate tests and probing this implementation from multiple different angles, given the importance of its correctness.

danderson:
oh, and: at a glance, I agree that it's slowPrefixTable that needs fixing (i.e. the main implementation doesn't seem to need changing), and your fixes seem like the correct ones to make it match our desired API semantics.

crawshaw:
I gave the fuzzer about 16 hours on a powerful machine, and it hasn't found anything else. Most of it on one beefy apple studio:

```
fuzz: elapsed: 12h59m13s, execs: 13386876614 (270281/sec), new interesting: 2 (total: 317)
```

(The rest on my old linux workstation, which appears to have been killed by this fuzzing attempt. Will not power on.)

danderson:
I added the test @bradfitz wrote, as well as the fixes @crawshaw found in fuzzing.

@crawshaw want me to import the fuzzing commits as well? Or do you want to send that as a separate PR?

net/art/table.go
danderson: Ack, keeping original implementation.

crawshaw:
There’s a bit of cleaning up I can do, separate PR?

danderson:
Separate PR works for me. So, I think this is ready to go, and I have more faith in its correctness than two days ago given extra stress testing and early fuzzing results. if someone's had a chance to ponder the change in depth (so far all the reviews emphasized they did a quick pass but no more).

net/art/table.go
```diff
@@ -16,147 +16,624 @@ import (
...
+			if remain := pfx.Bits() - intermediate.prefix.Bits(); remain <= 8 {
+				// pfx lives in intermediate.
+				if debugInsert {
+					fmt.Printf("insert: into intermediate intermediate.prefix=%s addr=%d/%d\n", intermediate.prefix, bs[finalByteIdx], finalBits)
+				}
+				intermediate.insert(bs[finalByteIdx], finalBits, val)
+			} else {
+				// pfx lives in a different child subtree of
+				// intermediate. By definition this subtree doesn't
+				// exist at all, otherwise we'd never have entereed
```
maisem, line 279: entered

net/art/table.go
```diff
@@ -16,147 +16,624 @@ import (
...
+	// the strideTable prefix on each table as we descend, but that means we
+	// have to pay N prefix.Contains checks on every route lookup (where N is
+	// the number of strideTables in the path), rather than only paying M prefix
+	// comparisons in the edge case (where M is the number of strideTables in
+	// the path with a non-nil route of their own).
+	strideIdx := 0
+	stridePrefixes := [16]netip.Prefix{}
+	strideRoutes := [16]*T{}
+findLeaf:
+	for {
```
maisem, line 74: This helps clarify scope
```
	for i, bs := 0, addr.AsSlice(); ; {
```

net/art/table.go
```diff
@@ -16,147 +16,624 @@ import (
...
+	// cases.
+	if pfx.Bits() == 0 {
+		if debugInsert {
+			fmt.Printf("insert: default route\n")
+		}
+		st.insert(0, 0, val)
+		return
+	}
+
 	bs := pfx.Addr().AsSlice()
```
maisem, line 159: can you move this down near the for loop?

net/art/stride_table.go
```diff
@@ -144,7 +188,7 @@ func (t *strideTable[T]) delete(addr uint8, prefixLen int) *T {
 
 	parentIdx := idx >> 1
 	t.allot(idx, idx, t.entries[parentIdx].prefixIndex, t.entries[parentIdx].value)
-	t.refs--
```
maisem, line 147: document the return value

maisem (approved):

net/art/stride_table.go
```diff
@@ -68,25 +77,60 @@ func (t *strideTable[T]) getChild(addr uint8) (child *strideTable[T], idx int) {
...
+	idx := hostIndex(addr)
+	if t.entries[idx].child == nil {
+		t.childRefs++
+	}
+	t.entries[idx].child = child
+}
+
+// setChildByIdx replaces the child strideTable at idx (if any) with
+// child. idx should be obtained via a call to getChild.
+func (t *strideTable[T]) setChildByIdx(idx int, child *strideTable[T]) {
```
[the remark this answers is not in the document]
bradfitz: (but "idx" or "i" in local variable names are fine... it's just funcs and types/fields where it's generally spelled out more, the further away it is, the longer of a name it gets)

net/art/table.go
```diff
@@ -16,147 +16,624 @@ import (
...
+		defer func() {
+			fmt.Printf("%s", t.debugSummary())
+		}()
+		fmt.Printf("\ninsert: start pfx=%s\n", pfx)
+	}
+
 	st := &t.v4
 	if pfx.Addr().Is6() {
 		st = &t.v6
 	}
```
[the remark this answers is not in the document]
danderson: Done.

net/art/table.go
danderson: Done.

net/art/table.go
danderson: Fixed

net/art/table.go
danderson: Going to switch to encoding/binary. The benchmarks are still kinda inconclusive, with some table shapes getting inexplicably faster/slower under this change, but it's not definitely better or definitely worse, and the readability benefit is probaby worth it.

net/art/table.go
```diff
@@ -16,147 +16,624 @@ import (
...
+	//
+	// We could also deal with this edge case of path compression by checking
+	// the strideTable prefix on each table as we descend, but that means we
+	// have to pay N prefix.Contains checks on every route lookup (where N is
+	// the number of strideTables in the path), rather than only paying M prefix
+	// comparisons in the edge case (where M is the number of strideTables in
+	// the path with a non-nil route of their own).
+	strideIdx := 0
+	stridePrefixes := [16]netip.Prefix{}
+	strideRoutes := [16]*T{}
```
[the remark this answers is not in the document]
danderson: Discussion offline: going with the more readable version for now, the perf loss is roughly what was gained with making Get alloc-free, so it's still quite fast indeed, and complex code should be readable foremost.

PR #8617, opened 2023-07-14
cli: introduce exit-node subcommand to list and filter exit nodes

catzkorn:
This change introduces a new subcommand, `exit-node`, along with a subsubcommand of `list` and a `--filter` flag.

Exit nodes without location data will continue to be displayed when `status` is used. Exit nodes with location data will only be displayed behind `exit-node list`, and in status if they are the active exit node.

The `filter` flag can be used to filter exit nodes with location data by country.

Exit nodes with Location.Priority data will have only the highest priority option for each country and city listed. For countries with multiple cities, a <Country> <Any> option will be displayed, indicating the highest priority node within that country.

Updates tailscale/corp#13025


cmd/tailscale/cli/exitnode.go
```diff
@@ -0,0 +1,245 @@
...
+	fmt.Fprintln(w)
+	fmt.Fprintln(w)
+	fmt.Fprintln(w, "# To use an exit node, use `tailscale set --exit-node=` followed by the hostname or IP")
+
+	return nil
+}
+
+// peerStatus returns a string representing the current state of
+// a peer. If there is no notable state, a - is returned.
+func peerStatus(peer *ipnstate.PeerStatus) string {
```
raggi, line 105: I think this is missing the "offers exit node" distinction, and also the case where an exit node is in use, but not "active", so e.g.

idle; offers exit node
active; offers exit node

raggi: Perhaps we can share printPS from status.go?

raggi (approved):
I think we should share the node status construction, but other than that looks good!

PR #8622, opened 2023-07-14
tstime: add Since method

clairew:
Updates #8463

### DIFF
--- a/tstest/clock.go
+++ b/tstest/clock.go
@@ -303,6 +303,11 @@ func (c *Clock) AfterFunc(d time.Duration, f func()) tstime.TimerController {
 	return t
 }
 
+// Since subtracts specified duration from Now().
+func (c *Clock) Since(t time.Time) time.Duration {
+	return c.Now().Sub(t)
+}
+
 // eventHandler offers a common interface for Timer and Ticker events to avoid
 // code duplication in eventManager.
 type eventHandler interface {
--- a/tstest/clock_test.go
+++ b/tstest/clock_test.go
@@ -2437,3 +2437,47 @@ func TestAfterFunc(t *testing.T) {
 		})
 	}
 }
+
+func TestSince(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name  string
+		start time.Time
+		since time.Time
+		want  time.Duration
+	}{
+		{
+			name:  "positive",
+			start: time.Unix(12345, 1000),
+			since: time.Unix(11111, 1000),
+			want:  1234 * time.Second,
+		},
+		{
+			name:  "negative",
+			start: time.Unix(12345, 1000),
+			since: time.Unix(15436, 1000),
+			want:  -3091 * time.Second,
+		},
+		{
+			name:  "zero",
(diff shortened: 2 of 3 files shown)


tstest/clock_test.go
```diff
@@ -2437,3 +2437,47 @@ func TestAfterFunc(t *testing.T) {
...
+		name  string
+		start time.Time
+		since time.Time
+		want  time.Duration
+	}{
+		{
+			name:  "positive since duration",
+			start: time.Unix(12345, 1000),
+			since: time.Unix(15436, 1000),
+			want:  time.Unix(12345, 1000).Sub(time.Unix(15436, 1000)),
```
sailorfrag, line 2454:
```suggestion
			name:  "negative",
			start: time.Unix(12345, 1000),
			since: time.Unix(15436, 1000),
			want:  -3091 * time.Second,
```

tstest/clock_test.go
```diff
@@ -2437,3 +2437,47 @@ func TestAfterFunc(t *testing.T) {
...
+		},
+		{
+			name:  "negative since duration",
+			start: time.Unix(12345, 1000),
+			since: time.Unix(11111, 1000),
+			want:  time.Unix(12345, 1000).Sub(time.Unix(11111, 1000)),
```
sailorfrag, line 2460:
```suggestion
			name:  "positive",
			start: time.Unix(12345, 1000),
			since: time.Unix(11111, 1000),
			want:  1234 * time.Second,
```

tstest/clock_test.go
```diff
@@ -2437,3 +2437,47 @@ func TestAfterFunc(t *testing.T) {
...
+		},
+		{
+			name:  "no since duration",
+			start: time.Unix(12345, 1000),
+			since: time.Unix(12345, 1000),
+			want:  time.Unix(12345, 1000).Sub(time.Unix(12345, 1000)),
```
sailorfrag, line 2466:
```suggestion
			name:  "zero",
			start: time.Unix(12345, 1000),
			since: time.Unix(12345, 1000),
			want:  0,
```

tstest/clock.go
```diff
@@ -303,6 +303,12 @@ func (c *Clock) AfterFunc(d time.Duration, f func()) tstime.TimerController {
 	return t
 }
 
+// Since subtracts specified duration from Now().
+func (c *Clock) Since(t time.Time) time.Duration {
+	c.init()
```
sailorfrag, line 308: Hmm. `c.Now()` will init so it's probably not necessary, but I guess leaving it here maintains the pattern.

sailorfrag (approved):

tstest/clock.go
```diff
@@ -303,6 +303,11 @@ func (c *Clock) AfterFunc(d time.Duration, f func()) tstime.TimerController {
 	return t
 }
 
+// Since subtracts specified duration from Now().
```
bradfitz, line 306: there is no duration specified here.

How about we copy the docs from the standard library, which says:

> Since returns the time elapsed since t. It is shorthand for time.Now().Sub(t).

PR #8624, opened 2023-07-15
controlclient: make logout a no-op if already logged out

### LINKED ISSUE #3833: "tailscale logout" twice in a row leads to tailscaled "no nodekey to log out" loop spam
(at f5ec916214e21bc0af0bdb7d48041d02e7703288) 

Run "tailscale logout" twice and watch tailscaled loop...

```
control: authRoutine: state:not-authenticated; wantLoggedIn=false
control: direct.TryLogout()
control: logging out...
control: doLogin(regen=false, hasUrl=false)
control: TryLogout control response: mustRegen=false, newURL=, err=no nodekey to log out
control: TryLogout: no nodekey to log out
control: sendStatus: authRoutine-report: state:not-authenticated
Received error: no nodekey to log out
control: authRoutine: backoff: 1662 msec
control: authRoutine: state:not-authenticated; wantLoggedIn=false
control: direct.TryLogout()
control: logging out...
control: doLogin(regen=false, hasUrl=false)
control: TryLogout control response: mustRegen=false, newURL=, err=no nodekey to log out
control: TryLogout: no nodekey to log out
control: sendStatus: authRoutine-report: state:not-authenticated
Received error: no nodekey to log out
control: authRoutine: backoff: 1856 msec
control: authRoutine: state:not-authenticated; wantLoggedIn=false
control: direct.TryLogout()
control: logging out...
control: doLogin(regen=false, hasUrl=false)
control: TryLogout control response: mustRegen=false, newURL=, err=no nodekey to log out
control: TryLogout: no nodekey to log out
control: sendStatus: authRoutine-report: state:not-authenticated
Received error: no nodekey to log out
control: authRoutine: backoff: 1497 msec
control: authRoutine: state:not-authenticated; wantLoggedIn=false
control: direct.TryLogout()
control: logging out...
control: doLogin(regen=false, hasUrl=false)
control: TryLogout control response: mustRegen=false, newURL=, err=no nodekey to log out
control: TryLogout: no nodekey to log out
control: sendStatus: authRoutine-report: state:not-authenticated
Received error: no nodekey to log out
control: authRoutine: backoff: 1411 msec
control: authRoutine: state:not-authenticated; wantLoggedIn=false
control: direct.TryLogout()
control: logging out...
control: doLogin(regen=false, hasUrl=false)
control: TryLogout control response: mustRegen=false, newURL=, err=no nodekey to log out
control: TryLogout: no nodekey to log out
control: sendStatus: authRoutine-report: state:not-authenticated
Received error: no nodekey to log out
control: authRoutine: backoff: 3748 msec
```

Maybe related: https://twitter.com/dorkmatt/status/1486923525336141827

> $ sudo tailscale logout
> HTTP 500 Internal Server Error: no nodekey to log out
>  (expected 204)

irbekrm:
See https://github.com/tailscale/tailscale/issues/3833#issuecomment-1636514610 for context- if a user attempts to log out when there is no node key (because they are already in logged out state), logout will fail and the [auth routine](https://github.com/tailscale/tailscale/blob/v1.44.0/control/controlclient/auto.go#L311) will keep attempting to log out and fail in a loop.

This PR fixes that by making any logout when the client is not logged in a no-op from the auth client 
It uses the [`Auto.loggedIn`](https://github.com/tailscale/tailscale/blob/v1.44.0/control/controlclient/auto.go#L64) field to determine if the client is logged in.
This field could be false:
- after a fully successful logout [here](https://github.com/tailscale/tailscale/blob/v1.44.0/control/controlclient/auto.go#L362)
- after logout from control server succeeded, but the state machine transition failed [here-ish](https://github.com/tailscale/tailscale/blob/v1.44.0/ipn/ipnlocal/local.go#L3916). The state machine transition will still run on subsequent logouts after this PR, but will also be a no-op (`NeedsLogin` -> `NeedsLogin` in success cases) and it does not seem to depend on anything that the now-skipped parts of code do.
- after a logout from control server succeeded, but [client state update](https://github.com/tailscale/tailscale/blob/v1.44.0/control/controlclient/auto.go#L409) failed (i.e updating the `tailscaled.state` file on disk). This is already never retried on subsequent logouts as the auth routine will fail before that and this PR does not change that
- after failure of the [login flow](https://github.com/tailscale/tailscale/blob/v1.44.0/control/controlclient/direct.go#L426). That should hopefully never require logging out.

An alternative fix to #3833 could have been to run the logout routine as before and only prevent the auth routine from looping by setting the `loginGoal` to nil [here-ish](https://github.com/tailscale/tailscale/blob/v1.44.0/control/controlclient/auto.go#L354-L358) if `loggedIn` is already set to false- but that kind of assumes that users having to log out multiple times might be a required flow in some cases, which it probably shouldn't be.

Fixes #3833

### DIFF
--- a/control/controlclient/auto.go
+++ b/control/controlclient/auto.go
@@ -694,9 +694,18 @@ func (c *Auto) StartLogout() {
 func (c *Auto) Logout(ctx context.Context) error {
 	c.logf("client.Logout()")
 
+	c.mu.Lock()
+	// This happens if a user logs out when tailscale client is already logged
+	// out either because the user logged out previously or hasn't yet logged
+	// in.
+	if !c.loggedIn {
+		c.logf("client.Logout(): no action taken, client is already logged out.")
+		c.mu.Unlock()
+		return nil
+	}
+
 	errc := make(chan error, 1)
 
-	c.mu.Lock()
 	c.loginGoal = &LoginGoal{
 		wantLoggedIn:    false,
 		loggedOutResult: errc,


maisem:
Can you update the commit message to match PR title?

control/controlclient/auto.go
```diff
@@ -694,6 +694,14 @@ func (c *Auto) StartLogout() {
 func (c *Auto) Logout(ctx context.Context) error {
 	c.logf("client.Logout()")
 
+	// This happens if a user logs out when tailscale client is alreay logged
+	// out either because the user logged out previously or hasn't yet logged
+	// in.
+	if !c.loggedIn {
```
maisem, line 700: Do you need to hold c.mu there?

irbekrm: Thank you, yes I guess the auth routine that sets that field can be running concurrently. Updated.

control/controlclient/auto.go
```diff
@@ -694,6 +694,17 @@ func (c *Auto) StartLogout() {
 func (c *Auto) Logout(ctx context.Context) error {
 	c.logf("client.Logout()")
 
+	c.mu.Lock()
+	loggedIn := c.loggedIn
+	c.mu.Unlock()
```
maisem, line 699: if you combine this block with the one below you will only need to acquire the lock once. something like

```
c.mu.Lock()
if !c.loggedIn {
  c.mu.Unlock()
  return nil
}
...
c.mu.Unlock()
```

irbekrm: 👍🏼 nice, that's a lot cleaner

control/controlclient/auto.go
```diff
@@ -694,9 +694,18 @@ func (c *Auto) StartLogout() {
 func (c *Auto) Logout(ctx context.Context) error {
 	c.logf("client.Logout()")
 
+	c.mu.Lock()
+	// This happens if a user logs out when tailscale client is alreay logged
```
bradfitz, line 698: typo: already

control/controlclient/auto.go
```diff
@@ -694,9 +694,18 @@ func (c *Auto) StartLogout() {
...
+	// out either because the user logged out previously or hasn't yet logged
+	// in.
+	if !c.loggedIn {
```
bradfitz, line 701: Today, Logout seems like it also aborts an in-flight login attempt (one that was started but hasn't completed). Now it doesn't? Is that a problem?

Would it mean a user couldn't transition from the logging-in to to Stopped state in the state machine?

In any case, let's not land this right before the upcoming release per Denton's earlier announcement, lest it destabilize the various clients.

irbekrm: > Today, Logout seems like it also aborts an in-flight login attempt (one that was started but hasn't completed). Now it doesn't? Is that a problem?

It doesn't seem to abort it at the moment.
If I run the cli login command from https://github.com/tailscale/tailscale/commit/c19b5bfbc391637b11c2acb3c725909a0046d849, wait for it to display the login URL, leave it hanging and run logout in parallel, the login remains hanging (I tried starting from both logged in and logged out state) and the logout results in the 'no nodekey to log out' error loop in both cases.
From this PR, the login remains hanging too.
In all cases the state ends up being `NeedsLogin` which as I understand it should be post-logout.
I might take a look at why the login is hanging and create a separate issue.

> In any case, let's not land this right before the upcoming release per Denton's earlier announcement, lest it destabilize the various clients.

👍🏼 I'll add a hold

irbekrm:
/hold

till 1.46.0 release as per https://github.com/tailscale/tailscale/pull/8624#discussion_r1264311096

irbekrm:
/on-hold

till 1.46.0 release as per https://github.com/tailscale/tailscale/pull/8624#discussion_r1264311096

control/controlclient/auto.go
irbekrm: > Would it mean a user couldn't transition from the logging-in to to Stopped state in the state machine?

Thanks for the comment - that transition does not currently work, but seems like it would be more intuitive if that's how logout worked, so I've changed this in https://github.com/tailscale/tailscale/pull/8757

Issue #8632, opened 2023-07-17
Update to Go 1.21


bradfitz:
Tracking bug for updating to Go 1.21 (which isn't entirely out yet, but rc3 is now out)

bradfitz:
Dup of #8419

PR #8633, opened 2023-07-17
go.toolchain.rev: bump Go to Go1.21rc3

bradfitz:
Updates #8632


go.toolchain.rev
```diff
@@ -1 +1 @@
-492f6d9d792fa6e4caa388e4d7bab46b48d07ad5
+41b84e37e1a195a6ac86876f2ba39e0840bd9621
```
bradfitz, line 1: https://github.com/tailscale/go/commits/41b84e37e1a195a6ac86876f2ba39e0840bd9621

phirework: Marwan flagged this on on Slack - do we also need to update `go.mod` because of this?

> A module’s go line must declare a version greater than or equal to the go version declared by each of the modules listed in require statements. A workspace’s go line must declare a version greater than or equal to the go version declared by each of the modules listed in use statements.
> 
> Before Go 1.21, toolchains did not require a module or workspace to have a go line greater than or equal to the go version required by each of its dependency modules.

bradfitz: Not yet. We're limiting ourselves to Go 1.20 features on purpose until Go 1.21 is actually out.

phirework (approved):

PR #8634, opened 2023-07-17
derp: use tstime

clairew:
Updates #8587


valscale (commented):
Nice work! I am wondering, does it make any sense to have everything use the one single s.clock instance of the interface rather than creating one for each set of client threads? There are several tens of thousands of clients connected to an average derper so if we can avoid creating a c.clock for each one, that would be nice. But there may be reasons to have one per client connection.

derp/derp_server.go
```diff
@@ -49,6 +50,8 @@ import (
 // verbosely log whenever DERP drops a packet.
 var verboseDropKeys = map[key.NodePublic]bool{}
 
+var clock = tstime.StdClock{}
```
bradfitz, line 53: this `clock` is of static type `tstime.StdClock` which means no other clock type can be assigned here.

I assume this should be:

```go
   var clock tstime.Clock = tstime.StdClock{}
```

derp/derp_server.go
```diff
@@ -65,7 +68,7 @@ func init() {
 }
 
 func init() {
-	rand.Seed(time.Now().UnixNano())
+	rand.Seed(clock.Now().UnixNano())
```
bradfitz, line 71: what's the goal of changing this one? who can change this clock?

clairew: I think this is unused so maybe I should leave it as is?

bradfitz: Yeah, let's keep in as-is. Any test wanting a deterministic seed can call `rand.Seed` again.

derp/derp_server.go
```diff
@@ -318,6 +323,7 @@ func NewServer(privateKey key.NodePrivate, logf logger.Logf) *Server {
 		avgQueueDuration:     new(uint64),
 		tcpRtt:               metrics.LabelMap{Label: "le"},
 		keyOfAddr:            map[netip.AddrPort]key.NodePublic{},
+		clock:                clock,
```
bradfitz, line 326: this is the only use of the global? I'd get rid of the global and set this to StdClock explicitly here and let tests change it as needed.

valscale (approved):
I agree with Brad's suggestions, otherwise looks good!

derp/derp_client.go
```diff
@@ -457,9 +461,8 @@ func (c *Client) recvTimeout(timeout time.Duration) (m ReceivedMessage, err erro
 			c.readErr.Store(err)
 		}
 	}()
-
 	for {
-		c.nc.SetReadDeadline(time.Now().Add(timeout))
+		c.nc.SetReadDeadline(c.clock.Now().Add(timeout))
```
sailorfrag, line 465: I recommend leaving this as `time.Now` because the deadline will be evaluated based on the real world clock. We'll have to do more work before we can simulate deadlines expiring.

derp/derp_server.go
```diff
@@ -664,12 +668,12 @@ func (s *Server) addWatcher(c *sclient) {
 
 func (s *Server) accept(ctx context.Context, nc Conn, brw *bufio.ReadWriter, remoteAddr string, connNum int64) error {
 	br := brw.Reader
-	nc.SetDeadline(time.Now().Add(10 * time.Second))
+	nc.SetDeadline(s.clock.Now().Add(10 * time.Second))
```
sailorfrag, line 671: Same with these deadlines.

derp/derp_server.go
```diff
@@ -1490,14 +1494,14 @@ func (c *sclient) sendLoop(ctx context.Context) error {
...
-		case <-keepAliveTick.C:
+		case <-keepAliveTickChannel:
 			werr = c.sendKeepAlive()
 		}
 	}
 }
 
 func (c *sclient) setWriteDeadline() {
-	c.nc.SetWriteDeadline(time.Now().Add(writeTimeout))
+	c.nc.SetWriteDeadline(c.s.clock.Now().Add(writeTimeout))
```
sailorfrag, line 1504: Also a deadline that we should leave with `time.Now`

sailorfrag:
> Nice work! I am wondering, does it make any sense to have everything in the derp server use the one single s.clock instance of the interface rather than creating one for each set of client threads (in the sclient struct)? There are several tens of thousands of clients connected to an average derper so if we can avoid creating a c.clock for each one on the server side, that would be nice. But there may be reasons to have one per client connection. (edited to clarify sclient in the server, not client in the client :) )

For `tstime.StdClock` it's an empty struct so there is only really one instance. My understanding is it'll show up in the interface as having a type pointer set but with a nil instance pointer. This does still have the interface overhead of a pair of pointers in each client/server struct, but that should be it.

Having all clients and servers share a clock for tests makes a certain amount of sense but then the plumbing is more difficult unless using globals, but that makes parallel tests impossible and is kind of icky.

Maybe there are some better ideas that I haven't thought of, but adding the clock to the client/server struct matches the general pattern I'd recommended and has worked well enough elsewhere.

derp/derp_server.go
```diff
@@ -669,7 +673,7 @@ func (s *Server) accept(ctx context.Context, nc Conn, brw *bufio.ReadWriter, rem
 	if err := s.sendServerKey(bw); err != nil {
 		return fmt.Errorf("send server key: %v", err)
 	}
-	nc.SetDeadline(time.Now().Add(10 * time.Second))
+	nc.SetDeadline(s.clock.Now().Add(10 * time.Second))
```
sailorfrag, line 676:
```suggestion
	nc.SetDeadline(time.Now().Add(10 * time.Second))
```

sailorfrag (approved):

PR #8641, opened 2023-07-18
cmd/tailscale/cli: make `tailscale update` query `softwareupdate`

noncombatant:
Even on macOS when Tailscale was installed via the App Store, we can check for
and even install new versions if people ask explicitly. Also, warn if App Store
AutoUpdate is not turned on.

Updates #6995

Signed-off-by: Chris Palmer <cpalmer@tailscale.com>

### DIFF
--- a/cmd/tailscale/cli/update.go
+++ b/cmd/tailscale/cli/update.go
@@ -356,20 +356,14 @@ func (up *updater) updateMacAppStore() error {
 		return nil
 	}
 
-	// parseSoftwareupdateList guarantees the format of the version string, but
-	// check and document it here, too. A case could be made for `!= 2`, not just
-	// `< 2`, as well.
-	newTailscaleParts := strings.Split(newTailscale, "-")
-	if len(newTailscaleParts) < 2 {
-		return fmt.Errorf(`Tailscale version from the App Store has unexpected format: %q, expect "Tailscale-x.y.z"`, newTailscale)
-	}
-	if err := up.confirm(newTailscaleParts[1]); err != nil {
+	newTailscaleVer := strings.TrimPrefix(newTailscale, "Tailscale-")
+	if err := up.confirm(newTailscaleVer); err != nil {
 		return err
 	}
 
 	cmd := exec.Command("sudo", "softwareupdate", "--install", newTailscale)
 	if err := cmd.Run(); err != nil {
-		return fmt.Errorf("can't install App Store update for Tailscale: %v", err)
+		return fmt.Errorf("can't install App Store update for Tailscale: %w", err)
 	}
 	return nil
 }
--- a/cmd/tailscale/cli/update.go
+++ b/cmd/tailscale/cli/update.go
@@ -357,11 +357,16 @@ func (up *updater) updateMacAppStore() error {
 	}
 
 	newTailscaleVer := strings.TrimPrefix(newTailscale, "Tailscale-")
+	if up.currentOrDryRun(newTailscaleVer) {
+		return nil
+	}
 	if err := up.confirm(newTailscaleVer); err != nil {
 		return err
 	}
 
 	cmd := exec.Command("sudo", "softwareupdate", "--install", newTailscale)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
 	if err := cmd.Run(); err != nil {
 		return fmt.Errorf("can't install App Store update for Tailscale: %w", err)
 	}
--- a/cmd/tailscale/cli/update.go
+++ b/cmd/tailscale/cli/update.go
@@ -44,6 +44,7 @@ var updateCmd = &ffcli.Command{
 		fs := newFlagSet("update")
 		fs.BoolVar(&updateArgs.yes, "yes", false, "update without interactive prompts")
 		fs.BoolVar(&updateArgs.dryRun, "dry-run", false, "print what update would do without doing it, or prompts")
+		fs.BoolVar(&updateArgs.appStore, "app-store", false, "check the App Store for updates, even if this is not an App Store install (for testing only!)")
 		fs.StringVar(&updateArgs.track, "track", "", `which track to check for updates: "stable" or "unstable" (dev); empty means same as current`)
 		fs.StringVar(&updateArgs.version, "version", "", `explicit version to update/downgrade to`)
 		return fs
@@ -51,10 +52,11 @@ var updateCmd = &ffcli.Command{
 }
 
 var updateArgs struct {
-	yes     bool
-	dryRun  bool
-	track   string // explicit track; empty means same as current
-	version string // explicit version; empty means auto
+	yes      bool
+	dryRun   bool
+	appStore bool
+	track    string // explicit track; empty means same as current
+	version  string // explicit version; empty means auto
 }
 
 // winMSIEnv is the environment variable that, if set, is the MSI file for the
@@ -140,12 +142,12 @@ func newUpdater() (*updater, error) {
 		}
 	case "darwin":
 		switch {
-		case !version.IsSandboxedMacOS():
+		case !updateArgs.appStore && !version.IsSandboxedMacOS():
 			return nil, errors.New("The 'update' command is not yet supported on this platform; see https://github.com/tailscale/tailscale/wiki/Tailscaled-on-macOS/ for now")
-		case strings.HasSuffix(os.Getenv("HOME"), "/io.tailscale.ipn.macsys/Data"):
+		case !updateArgs.appStore && strings.HasSuffix(os.Getenv("HOME"), "/io.tailscale.ipn.macsys/Data"):
 			up.update = up.updateMacSys
 		default:
-			return nil, errors.New("This is the macOS App Store version of Tailscale; update in the App Store, or see https://tailscale.com/s/unstable-clients to use TestFlight or to install the non-App Store version")
+			up.update = up.updateMacAppStore
 		}
 	}
 	if up.update == nil {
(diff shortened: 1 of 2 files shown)


cmd/tailscale/cli/update.go
```diff
@@ -333,6 +333,73 @@ func (up *updater) updateMacSys() error {
 	return errors.New("The 'update' command is not yet implemented on macOS.")
 }
 
+func (up *updater) updateMacAppStore() error {
+	fmt.Println(`This is the macOS App Store version of Tailscale. Updates are managed via the
+App Store, and the program 'softwareupdate'.
+
+See https://tailscale.com/s/unstable-clients to use TestFlight or to install
+the non-App Store version.`)
```
awly, line 341: nit: I wonder if this disclaimer is useful to the user, now that the command will deal with running the update. Maybe we can just omit it?

cmd/tailscale/cli/update.go
```diff
@@ -333,6 +333,73 @@ func (up *updater) updateMacSys() error {
...
+
+	cmd := exec.Command("softwareupdate", "--schedule")
+	stdout := bytes.NewBuffer(nil)
+	stderr := bytes.NewBuffer(nil)
+	cmd.Stdout = stdout
+	cmd.Stdout = stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("can't check App Store update schedule: %v", err)
```
awly, line 349:
```suggestion
		return fmt.Errorf("can't check App Store update schedule: %w, output: %q", err, stderr.String())
```

cmd/tailscale/cli/update.go
```diff
@@ -333,6 +333,73 @@ func (up *updater) updateMacSys() error {
...
+	}
+	fmt.Println(stdout.String())
```
awly, line 351: Can we parse the output and suggest turning on auto-updates if they are off?

cmd/tailscale/cli/update.go
```diff
@@ -333,6 +333,73 @@ func (up *updater) updateMacSys() error {
...
+
+	cmd = exec.Command("softwareupdate", "--list")
+	stdout.Reset()
+	stderr.Reset()
+	cmd.Stdout = stdout
+	cmd.Stdout = stderr
+	if err := cmd.Run(); err != nil {
+		fmt.Fprintln(os.Stderr, stdout.String())
+		return fmt.Errorf("can't check App Store for available updates: %v", err)
```
awly, line 360:
```suggestion
		return fmt.Errorf("can't check App Store for available updates: %w, output: %q", err, stderr.String())
```

cmd/tailscale/cli/update.go
```diff
@@ -333,6 +333,73 @@ func (up *updater) updateMacSys() error {
...
+
+See https://tailscale.com/s/unstable-clients to use TestFlight or to install
+the non-App Store version.`)
+
+	cmd := exec.Command("softwareupdate", "--schedule")
+	stdout := bytes.NewBuffer(nil)
+	stderr := bytes.NewBuffer(nil)
+	cmd.Stdout = stdout
+	cmd.Stdout = stderr
+	if err := cmd.Run(); err != nil {
```
awly, line 348: optional: you can also use `cmd.CombinedOutput()` to run the command and grab stdout+stderr

cmd/tailscale/cli/update.go
```diff
@@ -333,6 +333,73 @@ func (up *updater) updateMacSys() error {
...
+
+	if err := up.confirm(strings.Split(newTailscale, "-")[1]); err != nil {
+		return err
+	}
+
+	cmd = exec.Command("sudo", "softwareupdate", "--install", newTailscale)
+	stdout.Reset()
+	stderr.Reset()
+	cmd.Stdout = stdout
+	cmd.Stdout = stderr
```
awly, line 377: if we run with sudo, perhaps keep stdout/stderr piped into the usual os.Stdout/Stderr so that the password prompt is visible to the user?

cmd/tailscale/cli/update.go
```diff
@@ -333,6 +333,73 @@ func (up *updater) updateMacSys() error {
...
+// Darwin and returns the matching Tailscale package label. If there is none,
+// returns the empty string.
+//
+// See TestParseSoftwareupdateList for example inputs.
+func parseSoftwareupdateList(stdout []byte) string {
+	re, err := regexp.Compile(`(?m)^\s+\*\s+Label:\s*(Tailscale-\d[\d\.]+)`)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Could not compile regexp: %v", err)
+		return ""
+	}
```
awly, line 395: you can create a package-level var via `regexp.MustCompile` and catch regexp syntax errors early

cmd/tailscale/cli/update.go
```diff
@@ -333,6 +333,73 @@ func (up *updater) updateMacSys() error {
...
+		return fmt.Errorf("can't check App Store for available updates: %v", err)
+	}
+
+	newTailscale := parseSoftwareupdateList(stdout.Bytes())
+	if newTailscale == "" {
+		fmt.Println("no Tailscale update available")
+		return nil
+	}
+
+	if err := up.confirm(strings.Split(newTailscale, "-")[1]); err != nil {
```
awly, line 369:
```suggestion

	newTailscaleParts := strings.Split(newTailscale, "-")
	if len(newTailscaleParts) < 2 {
		return fmt.Errorf(`Tailscale version from the App Store has unexpected format: %q, expect "Tailscale-x.y.z"`, newTailscale)
	}
	if err := up.confirm(newTailscaleParts[1]); err != nil {
```

cmd/tailscale/cli/update_test.go
```diff
@@ -73,3 +73,74 @@ func TestUpdateDebianAptSourcesListBytes(t *testing.T) {
...
+			* Label: MacBookAirEFIUpdate2.4-2.4
+					 Title: MacBook Air EFI Firmware Update, Version: 2.4, Size: 3817K, Recommended: YES, Action: restart,
+			* Label: Malware-1.0
+					 Title: * Label: Tailscale-0.99.0, Version: 1.0, Size: 968K, Recommended: NOT REALLY TBH,
+`),
+			want: "",
+		},
+	}
+
+	for _, test := range tests {
```
awly, line 140: Consider using subtests (`t.Run`) with human-friendly descriptions to distinguish between the different test cases

awly:
This PR can be linked to https://github.com/tailscale/tailscale/issues/6995 to make issuebot happy

cmd/tailscale/cli/update.go
```diff
@@ -333,6 +333,65 @@ func (up *updater) updateMacSys() error {
...
+		return nil
+	}
+
+	// parseSoftwareupdateList guarantees the format of the version string, but
+	// check and document it here, too. A case could be made for `!= 2`, not just
+	// `< 2`, as well.
+	newTailscaleParts := strings.Split(newTailscale, "-")
+	if len(newTailscaleParts) < 2 {
+		return fmt.Errorf(`Tailscale version from the App Store has unexpected format: %q, expect "Tailscale-x.y.z"`, newTailscale)
+	}
```
awly, line 368: Thinking about it more, this could just be `newTailscaleVer := strings.TrimPrefix(newTailscale, "Tailscale-")` :sweat_smile:

cmd/tailscale/cli/update.go
```diff
@@ -333,6 +333,65 @@ func (up *updater) updateMacSys() error {
...
+	if err := up.confirm(newTailscaleParts[1]); err != nil {
+		return err
+	}
+
+	cmd := exec.Command("sudo", "softwareupdate", "--install", newTailscale)
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("can't install App Store update for Tailscale: %v", err)
```
awly, line 375:
```suggestion
		return fmt.Errorf("can't install App Store update for Tailscale: %w", err)
```

awly (approved):

cmd/tailscale/cli/update.go
```diff
@@ -44,17 +44,19 @@ var updateCmd = &ffcli.Command{
 		fs := newFlagSet("update")
 		fs.BoolVar(&updateArgs.yes, "yes", false, "update without interactive prompts")
 		fs.BoolVar(&updateArgs.dryRun, "dry-run", false, "print what update would do without doing it, or prompts")
+		fs.BoolVar(&updateArgs.appStore, "app-store", false, "check the App Store for updates, even if this is not an App Store install (for testing only!)")
```
awly, line 47:
```suggestion
		fs.BoolVar(&updateArgs.appStore, "app-store", false, "check the Mac App Store for updates, even if this is not an App Store install (for testing only!)")
```

cmd/tailscale/cli/update.go
```diff
@@ -44,17 +44,19 @@ var updateCmd = &ffcli.Command{
...
 }
 
 var updateArgs struct {
-	yes     bool
-	dryRun  bool
-	track   string // explicit track; empty means same as current
-	version string // explicit version; empty means auto
+	yes      bool
+	dryRun   bool
+	appStore bool
```
awly, line 57: just to disambiguate from other app stores
```suggestion
	macAppStore bool
```

cmd/tailscale/cli/update.go
noncombatant: Thanks! :)

cmd/tailscale/cli/update.go
noncombatant: I imagine we might want to overload this for testing on Windows, too. Having separate options for Mac and Win, but each only used on 1 platform, would be overly verbose.

cmd/tailscale/cli/update.go
noncombatant: same as above

cmd/tailscale/cli/update.go
```diff
@@ -333,6 +335,54 @@ func (up *updater) updateMacSys() error {
...
+		return fmt.Errorf("can't check App Store for available updates: %w, output: %q", err, string(out))
+	}
+
+	newTailscale := parseSoftwareupdateList(out)
+	if newTailscale == "" {
+		fmt.Println("no Tailscale update available")
+		return nil
+	}
+
+	newTailscaleVer := strings.TrimPrefix(newTailscale, "Tailscale-")
```
awly, line 359:
```suggestion
	newTailscaleVer := strings.TrimPrefix(newTailscale, "Tailscale-")
	if up.currentOrDryRun(newTailscaleVer) {
		return nil
	}
```

cmd/tailscale/cli/update.go
```diff
@@ -333,6 +335,54 @@ func (up *updater) updateMacSys() error {
...
+	if err := up.confirm(newTailscaleVer); err != nil {
+		return err
+	}
+
+	cmd := exec.Command("sudo", "softwareupdate", "--install", newTailscale)
```
awly, line 364: not sure how noisy it is, but you could set `cmd.Stdout = os.Stdout` and `cmd.Stderr = os.Stderr`
as I just learned, these are not the default :suspect: 

this would match the current deb behavior

awly (approved):

PR #8644, opened 2023-07-18
all: implement lock revoke-keys command

twitchyliquid64:
The recover-compromised-key command allows nodes with tailnet lock keys
to collaborate to erase the use of a compromised key, and remove trust
in it.

I could use some help with the help text for sure.

```
[nix-shell:~/tailscale]$ go build -o tailscale ./cmd/tailscale && sudo ./tailscale --socket=/tmp/ts/ts1.sock lock recover-compromised-key --help
USAGE
  recover-compromised-key <tailnet-lock-key>... or recover-compromised-key [--cosign] [--finish] <recovery-blob>

Retroactively revokes trust in the specified keys, erasing any trust in those keys and preventing them from being used in the future.

Start the process by running: tailscale recover-compromised-key <keys>, where keys is the list of compromised tailnet-lock keys.

The outputted command will then need to be run on other devices with trusted tailnet-lock keys to co-sign the recovery blob.
The latest-outputted from each cosign command should be used each time.

Once the command has been run on more devices than keys were compromised, the final outputted command can be run again with the --finish flag.

FLAGS
  -cosign=false  continue generating the recovery using the tailnet lock key on this device and the provided recovery blob
  -finish=false  finish the recovery process by transmitting the revocation
  ```

Signed-off-by: Tom DNetto <tom@tailscale.com>
Updates ENG-1848

### DIFF
--- a/cmd/tailscale/cli/network-lock.go
+++ b/cmd/tailscale/cli/network-lock.go
@@ -23,6 +23,7 @@ import (
 	"tailscale.com/ipn/ipnstate"
 	"tailscale.com/tka"
 	"tailscale.com/types/key"
+	"tailscale.com/types/tkatype"
 )
 
 var netlockCmd = &ffcli.Command{
@@ -40,6 +41,7 @@ var netlockCmd = &ffcli.Command{
 		nlDisablementKDFCmd,
 		nlLogCmd,
 		nlLocalDisableCmd,
+		nlRevokeKeysCmd,
 	},
 	Exec: runNetworkLockNoSubcommand,
 }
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by rosszurowski** [comment]
I've got some feedback. Take what makes sense, but feel to disagree with what doesn't.

- Why is the command “recover” compromised key and not "revoke" compromised key? The description makes it sound like revocation is what’s happening here.
- Why `compromised-key` specifically? Could we just say `revoke-key`? That sounds scary but also accurate.
- “the list of compromised tailnet-lock keys” might be nice to clarify if this is the tlpub: keys or nodekey: keys, or another type.
- Rather than abstract statements like "erasing any trust" maybe we can talk about specific impacts, like "revoking any nodes signed by that key"
- Using a numbered list to display the process might make it easier to grok.

Here's an edit that tries to capture some of these changes.

```
$ tailscale lock revoke-key --help
USAGE
  revoke-keys <tlpub-keys>
  revoke-keys [--cosign] [--finish] <recovery-blob>

Retroactively revoke the specified tailnet lock keys (tlpub:abc).

Revoked keys are prevented from being used in the future. Any nodes previously signed
by revoked keys lose their authorization and must be signed again.

Revocation is a multi-step process that requires several signing nodes to `--cosign` the revocation.

1. To start, run `tailscale revoke-keys <tlpub-keys>` with the tailnet lock keys to revoke.
2. Re-run the `--cosign` command output by `revoke-keys` on other signing nodes. Use the
   most recent command output on the next signing node in sequence.
3. Once the number of `--cosign`s is greater than the number of keys being revoked,
   the revocation will automatically complete.

FLAGS
  --cosign=false  cosign the revocation using the tailnet lock key on this device and the provided recovery blob
  --finish=false  finish the recovery process by transmitting the revocation
```

If it's possible to automatically detect when `--finish` should be called, it'd be great to eliminate that flag and just rely on the one `--cosign` variant. But I suspect that's not possible / too hard, otherwise you would have done it.

**Comment by sailorfrag** [comment]
I like the idea of combining the "revoke a key without having to rewrite the past" flow with the "revoke a key retroactively and rewrite the past" flow. I'm still working on reviewing the implementation.

**Comment by mayakacz** [comment]
> Why is the command “recover” compromised key and not "revoke" compromised key? The description makes it sound like revocation is what’s happening here.
Why compromised-key specifically? Could we just say revoke-key? That sounds scary but also accurate.
Rather than abstract statements like "erasing any trust" maybe we can talk about specific impacts, like "revoking any nodes signed by that key"

Agreed with @rosszurowski 's suggestions

> Here's an edit that tries to capture some of these changes.

👌 👨‍🍳 💋

File: ipn/ipnlocal/network-lock_test.go [hunk 1]
```diff
@@ -994,3 +994,129 @@ func TestTKAAffectedSigs(t *testing.T) {
...
+				if err := toApply[i].Unserialize(a); err != nil {
+					t.Fatalf("decoding missingAUM[%d]: %v", i, err)
+				}
+			}
+
+			// Apply the recovery AUM to an authority to make sure it works.
+			if err := authority.Inform(chonk, toApply); err != nil {
+				t.Errorf("recovery AUM could not be applied: %v", err)
+			}
+			// And make sure the key we removed isnt trusted.
```
sailorfrag, line 1058:
```suggestion
			// Make sure the key we removed isn't trusted.
```

ipn/ipnlocal/network-lock.go
```diff
@@ -845,6 +845,96 @@ func (b *LocalBackend) NetworkLockAffectedSigs(keyID tkatype.KeyID) ([]tkatype.M
...
+		return nil, errNetworkLockNotActive
+	}
+	var nlPriv key.NLPrivate
+	if p := b.pm.CurrentPrefs(); p.Valid() && p.Persist().Valid() {
+		nlPriv = p.Persist().NetworkLockKey()
+	}
+	if nlPriv.IsZero() {
+		return nil, errMissingNetmap
+	}
+	if !b.tka.authority.KeyTrusted(nlPriv.KeyID()) {
```
sailorfrag, line 899: I think this check is good if you initialize the tka to the parent of the recovery AUM? But otherwise it's probably not going to be accurate.

tka/tka.go
```diff
@@ -721,3 +721,86 @@ func (a *Authority) Compact(storage CompactableChonk, o CompactionOptions) error
...
+			return AUMHash{}, fmt.Errorf("computing state for %v: %w", cursor, err)
+		}
+		keyTrusted := false
+		for _, key := range removeKeys {
+			if _, err := state.GetKey(key); err == nil {
+				keyTrusted = true
+			}
+		}
+		if !keyTrusted {
+			return cursor, nil // niether key was trusted! Huzzah!
```
sailorfrag, line 754:
```suggestion
			return cursor, nil // Success: the keys are not trusted!
```

tka/tka_test.go
```diff
@@ -524,3 +524,125 @@ func TestAuthorityCompact(t *testing.T) {
...
+		optTemplate("add2", AUM{MessageKind: AUMAddKey, Key: &k2}),
+		optTemplate("add3", AUM{MessageKind: AUMAddKey, Key: &k3}),
+		optTemplate("remove2", AUM{MessageKind: AUMRemoveKey, KeyID: k2ID}))
+
+	a, err := Open(c.Chonk())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// k1 was trusted at genesis, so theres no better rewrite parent
```
sailorfrag, line 558:
```suggestion
	// k1 was trusted at genesis, so there's no better rewrite parent
```

ipn/ipnlocal/network-lock.go
```diff
@@ -845,6 +845,96 @@ func (b *LocalBackend) NetworkLockAffectedSigs(keyID tkatype.KeyID) ([]tkatype.M
...
```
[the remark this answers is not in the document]
twitchyliquid64: Yepp good point - ill move this check into tka so we can check against the earlier state.

tka/tka.go
```diff
@@ -721,3 +721,86 @@ func (a *Authority) Compact(storage CompactableChonk, o CompactionOptions) error
...
+				return nil, fmt.Errorf("computing keyID: %v", err)
+			}
+			if bytes.Equal(keyToRevoke, keyID) {
+				idx = i
+				break
+			}
+		}
+		if idx >= 0 {
+			state.Keys = append(state.Keys[:idx], state.Keys[idx+1:]...)
+		}
```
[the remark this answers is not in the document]
twitchyliquid64: Its safe to do nothing in this case, as the revoked key we are looking to remove already isnt part of the existing state.

twitchyliquid64:
Updated based on feedback:

1. Named `revoke-keys`
2. Updated help text per suggestion
3. Removed check that cosign/start tailnet lock key is currently trusted
4. Added check that the starting signing key is trusted when selecting a non-genesis AUM to fork from

twitchyliquid64:
```shell
[nix-shell:~/tailscale]$ go build -o tailscale ./cmd/tailscale && sudo ./tailscale --socket=/tmp/ts/ts1.sock lock revoke-keys --help
USAGE
  revoke-keys <tailnet-lock-key>...
  revoke-keys [--cosign] [--finish] <recovery-blob>

Retroactively revoke the specified tailnet lock keys (tlpub:abc).

Revoked keys are prevented from being used in the future. Any nodes previously signed
by revoked keys lose their authorization and must be signed again.

Revocation is a multi-step process that requires several signing nodes to `--cosign` the revocation. Use `tailscale lock remove` instead if the key has not been compromised.

1. To start, run `tailscale revoke-keys <tlpub-keys>` with the tailnet lock keys to revoke.
2. Re-run the `--cosign` command output by `revoke-keys` on other signing nodes. Use the
   most recent command output on the next signing node in sequence.
3. Once the number of `--cosign`s is greater than the number of keys being revoked,
   run the command one final time with `--finish` instead of `--cosign`.

FLAGS
  -cosign=false  continue generating the recovery using the tailnet lock key on this device and the provided recovery blob
  -finish=false  finish the recovery process by transmitting the revocation
```

rosszurowski:
Nice! Those updates look good to me! I'll leave it to Adrian to review the changes to the command itself.

cmd/tailscale/cli/network-lock.go
```diff
@@ -711,3 +713,96 @@ func wrapAuthKey(ctx context.Context, keyStr string, status *ipnstate.Status) er
...
+Revoked keys are prevented from being used in the future. Any nodes previously signed
+by revoked keys lose their authorization and must be signed again.
+
+Revocation is a multi-step process that requires several signing nodes to ` + "`--cosign`" + ` the revocation. Use ` + "`tailscale lock remove`" + ` instead if the key has not been compromised.
+
+1. To start, run ` + "`tailscale revoke-keys <tlpub-keys>`" + ` with the tailnet lock keys to revoke.
+2. Re-run the ` + "`--cosign`" + ` command output by ` + "`revoke-keys`" + ` on other signing nodes. Use the
+   most recent command output on the next signing node in sequence.
+3. Once the number of ` + "`--cosign`" + `s is greater than the number of keys being revoked,
+   run the command one final time with ` + "`--finish`" + ` instead of ` + "`--cosign`" + `.`,
```
sailorfrag, line 737: Is this technically "once the number of votes are greater than the votes of the keys being revoked"?

sailorfrag (approved):

cmd/tailscale/cli/network-lock.go
twitchyliquid64: Correct, but there is no documented way to create a key with a voting power different to 1.

PR #8651, opened 2023-07-19
cmd/gitops-pusher: re-use existing types from acl package

phirework:
This changes the ACLTestError type to reuse the existing/identical types from the ACL implementation, to avoid issues in the future if the two types fall out of sync.

### DIFF
--- a/cmd/gitops-pusher/gitops-pusher.go
+++ b/cmd/gitops-pusher/gitops-pusher.go
@@ -23,6 +23,7 @@ import (
 	"github.com/peterbourgon/ff/v3/ffcli"
 	"github.com/tailscale/hujson"
 	"golang.org/x/oauth2/clientcredentials"
+	"tailscale.com/client/tailscale"
 	"tailscale.com/util/httpm"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/gitops-pusher/gitops-pusher.go
```diff
@@ -327,12 +328,12 @@ func testNewACLs(ctx context.Context, client *http.Client, tailnet, apiKey, poli
 
 var lineColMessageSplit = regexp.MustCompile(`line ([0-9]+), column ([0-9]+): (.*)$`)
 
-type ACLTestError struct {
-	Message string               `json:"message"`
-	Data    []ACLTestErrorDetail `json:"data"`
+// ACLGitopsTestError is redefined here so we can add a custom .Error() response
+type ACLGitopsTestError struct {
```
maisem, line 332: does this unmarshal correctly?

maisem: looks like it does
https://go.dev/play/p/kYqClLeSNdw

maisem (approved):

maisem (approved):

PR #8655, opened 2023-07-19
cmd/tailscale/cli: implement update for arch-based distros

awly:
Arch version of tailscale is not maintained by us, but is generally up-to-date with our releases. Therefore "tailscale update" is just a thin wrapper around "pacman -Sy tailscale" with different flags.

Updates #6995

### DIFF
--- a/cmd/tailscale/cli/update.go
+++ b/cmd/tailscale/cli/update.go
@@ -45,8 +45,12 @@ var updateCmd = &ffcli.Command{
 		fs.BoolVar(&updateArgs.yes, "yes", false, "update without interactive prompts")
 		fs.BoolVar(&updateArgs.dryRun, "dry-run", false, "print what update would do without doing it, or prompts")
 		fs.BoolVar(&updateArgs.appStore, "app-store", false, "check the App Store for updates, even if this is not an App Store install (for testing only!)")
-		fs.StringVar(&updateArgs.track, "track", "", `which track to check for updates: "stable" or "unstable" (dev); empty means same as current`)
-		fs.StringVar(&updateArgs.version, "version", "", `explicit version to update/downgrade to`)
+		// These flags are not supported on Arch-based installs. Arch only
+		// offers one variant of tailscale and it's always the latest version.
+		if distro.Get() != distro.Arch {
+			fs.StringVar(&updateArgs.track, "track", "", `which track to check for updates: "stable" or "unstable" (dev); empty means same as current`)
+			fs.StringVar(&updateArgs.version, "version", "", `explicit version to update/downgrade to`)
+		}
 		return fs
 	})(),
 }
@@ -139,6 +143,8 @@ func newUpdater() (*updater, error) {
 			up.update = up.updateSynology
 		case distro.Debian: // includes Ubuntu
 			up.update = up.updateDebLike
+		case distro.Arch:
+			up.update = up.updateArch
 		}
 	case "darwin":
 		switch {
@@ -173,6 +179,8 @@ func (up *updater) currentOrDryRun(ver string) bool {
 	return false
 }
 
+var errUserAborted = errors.New("aborting update")
+
 func (up *updater) confirm(ver string) error {
 	if updateArgs.yes {
 		log.Printf("Updating Tailscale from %v to %v; --yes given, continuing without prompts.\n", version.Short(), ver)
@@ -187,7 +195,7 @@ func (up *updater) confirm(ver string) error {
 	case "y", "yes", "sure":
 		return nil
 	}
-	return errors.New("aborting update")
+	return errUserAborted
 }
 
 func (up *updater) updateSynology() error {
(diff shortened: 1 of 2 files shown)


cmd/tailscale/cli/update.go
```diff
@@ -44,8 +44,8 @@ var updateCmd = &ffcli.Command{
 		fs := newFlagSet("update")
 		fs.BoolVar(&updateArgs.yes, "yes", false, "update without interactive prompts")
 		fs.BoolVar(&updateArgs.dryRun, "dry-run", false, "print what update would do without doing it, or prompts")
-		fs.StringVar(&updateArgs.track, "track", "", `which track to check for updates: "stable" or "unstable" (dev); empty means same as current`)
-		fs.StringVar(&updateArgs.version, "version", "", `explicit version to update/downgrade to`)
+		fs.StringVar(&updateArgs.track, "track", "", `which track to check for updates: "stable" or "unstable" (dev); empty means same as current; ignored on Arch-based distros`)
```
noncombatant, line 47: I wonder if we can gate printing this additional caveat on GOOS=linux? It won't mean anything to Windows or macOS people, and might even confuse them.

cmd/tailscale/cli/update.go
```diff
@@ -324,6 +328,63 @@ func updateDebianAptSourcesListBytes(was []byte, dstTrack string) (newContent []
...
 }
 
+func (up *updater) updateArch() (err error) {
+	if os.Geteuid() != 0 {
+		return errors.New("must be root; use sudo")
+	}
+
+	defer func() {
+		if err != nil && !errors.Is(err, errUserAborted) {
+			err = fmt.Errorf(`%w; you can try updating using "pacman -Sy tailscale"`, err)
```
noncombatant, line 338: Might be nice to recommend the exact same command line that we run below? I like the long form of options since it's documentation.

noncombatant (approved):

cmd/tailscale/cli/update.go
awly: Added a condition for these flags to only register on non-Arch hosts :+1:

cmd/tailscale/cli/update.go
```diff
@@ -326,6 +334,63 @@ func updateDebianAptSourcesListBytes(was []byte, dstTrack string) (newContent []
...
+		return errors.New("must be root; use sudo")
+	}
+
+	defer func() {
+		if err != nil && !errors.Is(err, errUserAborted) {
+			err = fmt.Errorf(`%w; you can try updating using "pacman --sync --refresh tailscale"`, err)
+		}
+	}()
+
+	out, err := exec.Command("pacman", "--sync", "--refresh", "--info", "tailscale").CombinedOutput()
```
Foxboron, line 348: This is going to break peoples systems as partial updates are not supported on Arch, it's everything or nothing. So this will at some point lead to someone installing incompatible so-name updates on their system and hose it.

You have access to a reasonably responsive Arch maintainer, please reach out before implementing things like this.

cmd/tailscale/cli/update.go
```diff
@@ -326,6 +334,63 @@ func updateDebianAptSourcesListBytes(was []byte, dstTrack string) (newContent []
...
+		if len(parts) != 2 {
+			return "", fmt.Errorf("version output from pacman is malformed: %q, cannot determine upgrade version", line)
+		}
+		ver := strings.TrimSpace(parts[1])
+		// Trim the Arch patch version.
+		ver = strings.Split(ver, "-")[0]
+		if ver == "" {
+			return "", fmt.Errorf("version output from pacman is malformed: %q, cannot determine upgrade version", line)
+		}
+		return ver, nil
```
Foxboron, line 389: This misses versions like `1:1.4.2-1` which would be a valid version if we where to preform a downgrade on the package.

cmd/tailscale/cli/update.go
christian-heusel: As this command in short it is the prime example for a partial update `pacman -Syi tailscale`, more info on that here: https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported

jelly: Alternatively you don't have to parse the whole information snippet but instead use `--print-format`.

```
pacman -S tailscale --print-format "%v"
```

awly: Thanks for your comments!
https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported talks mostly about libraries. AFAIK the tailscale package itself doesn't contain any shared libraries and only depends on libc.
Is it still risky to update this kind of "leaf" package?

As for the general topic of auto-updates, is there any "standard" way to apply updates in the background for a rolling release like Arch? Other than writing a cronjob, of course.

We would really like to keep all tailscale clients on recent-ish versions, both for general bug/security fixes and for reducing the backwards-compatibility kludges in APIs that tailscale clients talk to.
But we also want to play nice with the package ecosystem of the distro that people use. So if we can't help users keep tailscale up-to-date on Arch without breaking stuff, then we'll have to resort to docs and hope people read them.

Foxboron: > https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported talks mostly about libraries. AFAIK the tailscale package itself doesn't contain any shared libraries and only depends on libc.
Is it still risky to update this kind of "leaf" package?

The command is not acting on the package at all.

`pacman -Sy tailscale` is the same as `pacman -Sy; pacman -S tailscale`. You are refreshing the databases and will pull inn updates relevant for the `tailscale` package. Currently `tailscale` depends on `glibc` so if you are so unlucky there is a new glibc version (which your system does not yet support), you are pulling inn both.

This will break your system.

> As for the general topic of auto-updates, is there any "standard" way to apply updates in the background for a rolling release like Arch? Other than writing a cronjob, of course.

You just don't do this at all on Arch. Arch is a rolling release distro where users needs to resolve the result of any updates. Stuffing this into a cronjob is just a disaster waiting to happen. Unattended upgrades has the same issue.

>We would really like to keep all tailscale clients on recent-ish versions, both for general bug/security fixes and for reducing the backwards-compatibility kludges in APIs that tailscale clients talk to.
But we also want to play nice with the package ecosystem of the distro that people use.

On Linux distros, how would implementing a completely new command encourage people to update more frequently? People are lazy when things work and updating unattended is bad behavior. Telling users to update more often is not really going to make them do it either.


> . So if we can't help users keep tailscale up-to-date on Arch without breaking stuff, then we'll have to resort to docs and hope people read them.

Arch assumes competent users, so this is completely fine to expect of them.

Foxboron: To be clear, `pacman -Sy tailscale` is going to make all the next `pacman -S` operations dangerous until a full upgrade , `pacman -Syu`,  is completed.

PR #8657, opened 2023-07-20
cmd/tailscale/cli: limit Darwin-only option to Darwin

noncombatant:
Hide the `-app-store` option with `HIDDEN: `.

Updates #6995

### DIFF
--- a/cmd/tailscale/cli/update.go
+++ b/cmd/tailscale/cli/update.go
@@ -44,7 +44,10 @@ var updateCmd = &ffcli.Command{
 		fs := newFlagSet("update")
 		fs.BoolVar(&updateArgs.yes, "yes", false, "update without interactive prompts")
 		fs.BoolVar(&updateArgs.dryRun, "dry-run", false, "print what update would do without doing it, or prompts")
-		fs.BoolVar(&updateArgs.appStore, "app-store", false, "check the App Store for updates, even if this is not an App Store install (for testing only!)")
+		if runtime.GOOS == "darwin" {
+			// TODO(cpalmer): Maybe Windows someday as well.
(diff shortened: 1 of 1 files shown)


cmd/tailscale/cli/update.go
```diff
@@ -44,7 +44,10 @@ var updateCmd = &ffcli.Command{
 		fs := newFlagSet("update")
 		fs.BoolVar(&updateArgs.yes, "yes", false, "update without interactive prompts")
 		fs.BoolVar(&updateArgs.dryRun, "dry-run", false, "print what update would do without doing it, or prompts")
-		fs.BoolVar(&updateArgs.appStore, "app-store", false, "check the App Store for updates, even if this is not an App Store install (for testing only!)")
+		if runtime.GOOS == "darwin" {
+			// TODO(cpalmer): Maybe Windows someday as well.
+			fs.BoolVar(&updateArgs.appStore, "app-store", false, "check the App Store for updates, even if this is not an App Store install (for testing only!)")
```
bradfitz, line 49: drop the exclamation mark. they're always more fun to write than they are to read.

bradfitz (approved):

cmd/tailscale/cli/update.go
bradfitz: But what does "for testing only" mean? Testing by whom? Us? Should this be behind an envknob?

Or! Better! (if you want this to be available only if you know about it)

Remove the darwin conditional and instead make the description start with the magic phrase `"HIDDEN: "` and then it'll be omitted by our flag help printing code but the flag will still exist, if you know about it.

awly (approved):
TIL about `HIDDEN: ` flag prefix!

cmd/tailscale/cli/update.go
noncombatant: Yes, testing by us. `HIDDEN: ` sounds like the way to go; new version coming up.

PR #8660, opened 2023-07-20
tstest/linuxdeps: add test to reject gvisor.dev/gvisor/pkg/hostarch

raggi:
This needs a little finishing, an example for @KevinLiang10

### DIFF
--- a/tstest/linuxdeps/linuxdeps.go
+++ b/tstest/linuxdeps/linuxdeps.go
@@ -1,0 +1,5 @@
+package linuxdeps
+
+import (
+	_ "tailscale.com/util/linuxfw"
+)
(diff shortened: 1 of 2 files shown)


bradfitz:
Whoops, just back to a computer and sent Kevin https://github.com/tailscale/tailscale/pull/8667 before seeing this one.

bradfitz:
Notably, #8667 removes all the copy/paste of those tests rather than adding another paste of it. And the test is in `cmd/tailscaled` so it can't go stale over time. We only do the `tstest/iotest` and `tstest/jstest` in oss as a smoke test because the actual source is in another repo. But for `cmd/tailscaled`, we can put the dep test directly in the `cmd/tailscaled` package's tests and not have to mirror its deps manually.

raggi:
Yeah, that's a much better plan, thanks!

PR #8661, opened 2023-07-20
net: use tstime

clairew:
Updates #8587


bradfitz (commented):
I'd like to put a pause on all these clock plumbing changes and not submit this change.

It's fine that we have this clock mechanism now, but the point of all the clock work was to make things more testable (and less flakily testable).

Going forward I only want to see further clock plumbing done to support test improvements.

If this change is deflaking a test or adding test coverage, please note that in the commit message.

PR #8662, opened 2023-07-20
cmd/tailscale/cli: make update visible in list

noncombatant:
This also makes "HIDDEN: " work (requires the custom UsageFunc).

Updates #6995

### DIFF
--- a/cmd/tailscale/cli/cli.go
+++ b/cmd/tailscale/cli/cli.go
@@ -130,6 +130,7 @@ change in the future.
 			netlockCmd,
 			licensesCmd,
 			exitNodeCmd,
+			updateCmd,
 		},
 		FlagSet:   rootfs,
 		Exec:      func(context.Context, []string) error { return flag.ErrHelp },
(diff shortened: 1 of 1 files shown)


awly (commented):
How about we also move https://github.com/tailscale/tailscale/pull/8662/files#diff-5d05c094ca6c20d2455074acd16f67c05ac944caab86622408ff06502eb3aa70R138-R142 until after all the subcommands are populated?

bradfitz:
FWIW it was hidden on purpose because it wasn't quite done. I'm fine unhiding it, as long as we're doing it intentionally and not just to like 'fix a bug' that it wasn't showing up or something?

bradfitz:
Oh, there's even a comment above that section of the code saying that basically :)

bradfitz (approved):

noncombatant:
Yes, @awly and I are working on a project to make `update` explicitly available everywhere.

bradfitz:
> Yes, @awly and I are working on a project to make update explicitly available everywhere.

I guess the question is whether we unhide it before or after the project is complete. Because arguably I was also working on a project to make it available everywhere, but it was just very low priority and I didn't know when it'd be done so I kept it hidden... 😂

bradfitz:
Maybe just hide it again before the next release if you're not happy about the state of things.

noncombatant:
Let's keep it hidden until we're done. We'll just keep this PR un-landed until we're ready.

In the meantime, this PR makes the custom UsageFunc always set for all subcommands, as Andrew suggested: https://github.com/tailscale/tailscale/pull/8665

noncombatant:
I think we can land this one now. What do you say?

awly (approved):

PR #8664, opened 2023-07-20
cmd/stunc: add features for imperative debugging of stun services

raggi:
Add various features more useful for confirmation/diagnosis work.

- Probes repeat forever, reporting results one per line
- Repetition is approximately isochronous at the given rate
- Either a host name or an address can be given
- The timeout is tuneable
- Keep going after a timeout occurs

Updates tailscale/corp#13378

### DIFF
--- a/cmd/stunc/stunc.go
+++ b/cmd/stunc/stunc.go
@@ -5,53 +5,121 @@
 package main
 
 import (
+	"encoding/json"
+	"flag"
+	"fmt"
 	"log"
 	"net"
+	"net/netip"
 	"os"
+	"time"
 
 	"tailscale.com/net/stun"
 )
 
+var (
+	rate    = flag.Duration("rate", time.Second, "rate at which to send probes (0 means as fast as possible)")
+	timeout = flag.Duration("timeout", time.Second, "time to wait for a response")
+	reuse   = flag.Bool("reuse", true, "reuse the same UDP socket for each probe")
+	jsonout = flag.Bool("json", false, "output in JSON format (default is human-readable)")
+)
+
 func main() {
-	log.SetFlags(0)
-
-	if len(os.Args) != 2 {
-		log.Fatalf("usage: %s <hostname>", os.Args[0])
-	}
-	host := os.Args[1]
-
-	uaddr, err := net.ResolveUDPAddr("udp", host+":3478")
-	if err != nil {
-		log.Fatal(err)
-	}
-	c, err := net.ListenUDP("udp", nil)
-	if err != nil {
-		log.Fatal(err)
+	flag.Usage = func() {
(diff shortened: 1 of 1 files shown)


cmd/stunc/stunc.go
```diff
@@ -5,53 +5,121 @@
...
+	"time"
 
 	"tailscale.com/net/stun"
 )
 
-func main() {
-	log.SetFlags(0)
+var (
+	rate    = flag.Duration("rate", time.Second, "rate at which to send probes (0 means as fast as possible)")
+	timeout = flag.Duration("timeout", time.Second, "time to wait for a response")
```
maisem, line 22: add v6 mode too for resolution?

cmd/stunc/stunc.go
```diff
@@ -5,53 +5,121 @@
...
+		}
 	}
 
-	log.Printf("sent addr: %v", uaddr)
-	log.Printf("from addr: %v", raddr)
-	log.Printf("stun addr: %v", saddr)
+	for {
+		if c == nil || !*reuse {
+			if c != nil {
+				c.Close()
```
maisem, line 74: nit: add this outside of the loop
```
defer func() {
  if c != nil {
    c.Close()
  }
}()
```

cmd/stunc/stunc.go
```diff
@@ -5,53 +5,121 @@
...
+		if result["status"] == "ok" {
+			r = fmt.Sprintf("%s; %s < %s in %s", result["status"], result["from"], result["stun"], result["dur"])
+		}
+		fmt.Printf("%s > %s; %s\n", result["local"], result["to"], r)
 	}
-	if tid != txID {
-		log.Fatalf("txid mismatch: got %v, want %v", tid, txID)
+	if *jsonout {
+		j := json.NewEncoder(os.Stdout)
+		print = func(result map[string]string) {
```
maisem, line 64: add SetIndent?

cmd/stunc/stunc.go
```diff
@@ -5,53 +5,121 @@
...
+			}
+
+			c, err = net.ListenUDP("udp", nil)
+			if err != nil {
+				log.Fatal(err)
+			}
+		}
+
+		result := map[string]string{}
+		result["to"] = uaddr.String()
```
maisem, line 84: may want to use a defined struct so that its easier to figure out what fields are available.

maisem (approved):

cmd/stunc/stunc.go
raggi: It's producing a stream of single line json in the current version, which is approximately the most convenient if you're scripting something with a less smart parser in the mix, but it could be an option. I think for now people can pipe to jq though

maisem: Yeah piping to jq is fine

PR #8666, opened 2023-07-20
version/distro: fallback to /etc/os-release on Linux

awly:
On most Linux distros, you can read the distro ID from `/etc/os-release`. In fact, we do that in
https://github.com/tailscale/tailscale/blob/main/scripts/installer.sh#L38

Do this as a fallback to avoid accidentally breaking some logic elsewhere. Also add a few known distros to the const list.

### DIFF
--- a/version/distro/distro.go
+++ b/version/distro/distro.go
@@ -5,21 +5,34 @@
 package distro
 
 import (
+	"bufio"
 	"bytes"
 	"io"
 	"os"
 	"runtime"
 	"strconv"
+	"strings"
 
+	"go4.org/mem"
 	"tailscale.com/types/lazy"
 	"tailscale.com/util/lineread"
 )
 
 type Distro string
 
+// List of some common Distro names.
+//
+// This list is not exhaustive, Get may return other values based on
+// /etc/os-release on Linux.
+//
+// Distro values returned by Get represent the oldest Distro in the family of
+// Linux distributions. For example, Debian is returned for Ubuntu and any
+// other Debian-derived distributions. Please do not add constants for Ubuntu
+// or other derived distributions.
 const (
-	Debian    = Distro("debian")
-	Arch      = Distro("arch")
+	Debian    = Distro("debian") // includes Ubuntu, Linux Mint, etc.
+	Arch      = Distro("arch")   // includes EndeavourOS, SteamOS, etc.
+	Fedora    = Distro("fedora") // includes CentOS, RHEL, Oracle Linux, etc.
 	Synology  = Distro("synology")
 	OpenWrt   = Distro("openwrt")
 	NixOS     = Distro("nixos")
@@ -30,12 +43,20 @@ const (
 	Gokrazy   = Distro("gokrazy")
 	WDMyCloud = Distro("wdmycloud")
 	Unraid    = Distro("unraid")
+	Alpine    = Distro("alpine")
+	Void      = Distro("void")
+	Gentoo    = Distro("gentoo")
+	Unknown   = Distro("")
 )
 
 var distro lazy.SyncValue[Distro]
 var isWSL lazy.SyncValue[bool]
 
-// Get returns the current distro, or the empty string if unknown.
+// Get returns the current distro, or Unknown.
+//
+// Distro values returned by Get represent the oldest Distro in the family of
+// Linux distributions. For example, Debian is returned for Ubuntu and any
+// other Debian-derived distributions.
 func Get() Distro {
 	return distro.Get(func() Distro {
 		switch runtime.GOOS {
@@ -44,7 +65,7 @@ func Get() Distro {
 		case "freebsd":
 			return freebsdDistro()
 		default:
-			return Distro("")
+			return Unknown
 		}
 	})
 }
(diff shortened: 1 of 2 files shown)


bradfitz (commented):
Warning: I tried to do this once and it got hairy and I aborted for lack of time. We have some code that assumes the Debian const means Debian or Ubuntu or Debian-like. I then started adding a Distro.Like method to make a family tree of distros for code making decisions based on distros and it got gross.

But just adding Ubuntu seems almost certain to break things.

version/distro/distro_test.go
```diff
@@ -13,3 +17,58 @@ func BenchmarkGet(b *testing.B) {
...
+NAME=Debian
+ID='debian'
+`,
+			want: Debian,
+		},
+		{
+			desc: "double quoted",
+			osRelease: `
+NAME=Debian
+ID="debian"
```
noncombatant, line 47: Not sure if it would ever happen, and I think your `Trim` call would handle it, but maybe test unclosed and un-opened quotes?

"Debian
Debian'

version/distro/distro_test.go
```diff
@@ -13,3 +17,58 @@ func BenchmarkGet(b *testing.B) {
...
+`,
+			want: Debian,
+		},
+		{
+			desc: "missing ID",
+			osRelease: `
+NAME=Debian
+`,
```
noncombatant, line 55: Maybe also a totally empty file, and maybe

NAME=
ID=

?

noncombatant (approved):
LGTM; trivially pathological test cases optional.

version/distro/distro.go
```diff
@@ -93,8 +109,32 @@ func linuxDistro() Distro {
...
+	if err != nil {
+		return Unknown, err
+	}
+	defer f.Close()
+	s := bufio.NewScanner(f)
+	for s.Scan() {
+		if !strings.HasPrefix(s.Text(), "ID=") {
+			continue
+		}
+		d := strings.TrimPrefix(s.Text(), "ID=")
```
bradfitz, line 132: this looks like two allocs on match

it could be one with CutPrefix.

but then it'd still allocate on each line that didn't match. but you could use `!mem.HasPrefix(s.Bytes(), mem.S("ID="))` and not allocate on the non-matching lines

bradfitz:
For example, code like this is really asking "Is this a Debian or Debian-derived distro?"

```go
	switch distro.Get() {
	case distro.Debian:
		dpkgOut, err := exec.Command("dpkg", "-S", "kernel/drivers/net/tun.ko").CombinedOutput()
```

So that'd be better like:

```go
    d := distro.Get()
    switch {
     case d.HasDpkg():
            ....
```

(with some new methods on `Distro string` answering specific questions)

awly:
@bradfitz good point. There's also an `ID_LIKE` in `/etc/os-release` that _should_ include `debian` on that family of systems. I'll take another stab at this, giving preference to `ID_LIKE` over `ID`

version/distro/distro.go
awly: Done.
I've never used that mem package, please correct if my usage is wrong

awly:
@bradfitz added handling or `ID_LIKE` values. Attached below a list of `os-release` IDs across most of the linuxes we support (at least their official docker images). Specifically for Ubuntu, this code should identify it as `distro.Debian`.
Also, the `os-release` parsing is only added as fallback. It will only change cases where existing `distro.Get` would return an empty value.

```
$ for img in (cat images); echo --- image: $img --- && docker run -it --rm -v .:/mnt $img cat /etc/os-release | ag '^ID'; end
--- image: ubuntu:16.04 ---
ID=ubuntu
ID_LIKE=debian
--- image: ubuntu:18.04 ---
ID=ubuntu
ID_LIKE=debian
--- image: ubuntu:19.10 ---
ID=ubuntu
ID_LIKE=debian
--- image: ubuntu:20.04 ---
ID=ubuntu
ID_LIKE=debian
--- image: ubuntu:20.10 ---
ID=ubuntu
ID_LIKE=debian
--- image: ubuntu:21.04 ---
ID=ubuntu
ID_LIKE=debian
--- image: ubuntu:21.10 ---
ID=ubuntu
ID_LIKE=debian
--- image: ubuntu:22.04 ---
ID=ubuntu
ID_LIKE=debian
--- image: ubuntu:22.10 ---
ID=ubuntu
ID_LIKE=debian
--- image: ubuntu:23.04 ---
ID=ubuntu
ID_LIKE=debian
--- image: ubuntu:23.10 ---
ID=ubuntu
ID_LIKE=debian
--- image: debian:bookworm ---
ID=debian
--- image: debian:bullseye ---
ID=debian
--- image: debian:buster ---
ID=debian
--- image: debian:sid ---
ID=debian
--- image: debian:trixie ---
ID=debian
--- image: fedora:33 ---
ID=fedora
--- image: fedora:34 ---
ID=fedora
--- image: fedora:35 ---
ID=fedora
--- image: fedora:36 ---
ID=fedora
--- image: fedora:37 ---
ID=fedora
--- image: redhat/ubi8 ---
ID="rhel"
ID_LIKE="fedora"
--- image: redhat/ubi9 ---
ID="rhel"
ID_LIKE="fedora"
--- image: centos:7 ---
ID="centos"
ID_LIKE="rhel fedora"
--- image: centos:8 ---
ID="centos"
ID_LIKE="rhel fedora"
--- image: alpine ---
ID=alpine
--- image: amazonlinux:2 ---
ID="amzn"
ID_LIKE="centos rhel fedora"
--- image: amazonlinux:2022 ---
ID="amzn"
ID_LIKE="fedora"
--- image: amazonlinux:2023 ---
ID="amzn"
ID_LIKE="fedora"
--- image: oraclelinux:7 ---
ID="ol"
ID_LIKE="fedora"
--- image: oraclelinux:8 ---
ID="ol"
ID_LIKE="fedora"
--- image: oraclelinux:9 ---
ID="ol"
ID_LIKE="fedora"
--- image: manjarolinux/base ---
ID=manjaro
ID_LIKE=arch
```

version/distro/distro.go
```diff
@@ -93,8 +101,56 @@ func linuxDistro() Distro {
...
+	defer f.Close()
+	s := bufio.NewScanner(f)
+	var id, idLike string
+	for s.Scan() {
+		line := mem.S(s.Text())
+		if mem.HasPrefix(line, mem.S("ID=")) {
+			id = mem.TrimPrefix(line, mem.S("ID=")).StringCopy()
+			id = strings.Trim(id, `"'`)
+		}
+		if mem.HasPrefix(line, mem.S("ID_LIKE=")) {
```
bradfitz, line 131: I should add CutPrefix to go4.org/mem :)

version/distro/distro.go
```diff
@@ -30,6 +33,11 @@ const (
 	Gokrazy   = Distro("gokrazy")
 	WDMyCloud = Distro("wdmycloud")
 	Unraid    = Distro("unraid")
+	Fedora    = Distro("fedora")
+	Alpine    = Distro("alpine")
+	Void      = Distro("void")
+	Gentoo    = Distro("gentoo")
+	Unknown   = Distro("")
 )
```
bradfitz, line 41: Can you update (er, add) the docs on the Distro type while you're here, warning that these consts aren't an exhaustive set, and warning about adding ubuntu or others?

And add an explicit line like:

```
Debian    = Distro("debian") // includes ubuntu, etc
```

version/distro/distro.go
```diff
@@ -93,8 +101,56 @@ func linuxDistro() Distro {
...
+			if len(idLikeItems) == 0 {
+				continue
+			}
+			idLike = idLikeItems[len(idLikeItems)-1]
+		}
+	}
+	if err := s.Err(); err != nil {
+		return Unknown, err
+	}
+	if idLike != "" {
```
bradfitz, line 147: should we have a denylist of values we never want an idLike value of? such as "ubuntu" for now?

awly: Added a denylist on the return value of `parseLinuxOSRelease`, whether it's coming from `ID` or `ID_LIKE`.

awly:
Closing for now, as `tailscale update` doesn't need this anymore and this change risks making `distro.Get()` a bit more non-deterministic.
Note to self: if we need this change again, maybe put it in https://github.com/tailscale/tailscale/blob/main/hostinfo/hostinfo_linux.go ?

Issue #8658, opened 2023-07-20
tailscale-1.46.0 does not work on ARM64 systems with page size != 4k


krsuhjunho:
### What is the issue?

1.46.0-1
panic: Only 4K page size is supported on arm64!

goroutine 1 [running]:
gvisor.dev/gvisor/pkg/hostarch.init.0()
        gvisor.dev/gvisor@v0.0.0-20230504175454-7b0a1988a28f/pkg/hostarch/hostarch_arm64.go:59 +0x44

1.44.0 

No Error

Thank you

### Steps to reproduce

tailscale in ssh 

### OS

Linux

### OS version

AlmaLinux release 8.8 (Sapphire Caracal)

### Tailscale version

1.46.0-1

MDrollette:
I ran into this this morning as well. It had been working fine and suddenly started getting this error during the middle of the night. I'm not sure if it was a tailscale auto-update or an Oracle update that triggered it.

DentonGentry:
We are looking into this.

bradfitz:
@DentonGentry, debugging notes. We didn't bump our gvisor dep from 1.44.2 to 1.46.0:

```patch
$ git diff -r v1.44.2..v1.46.0 go.mod
diff --git a/go.mod b/go.mod
index 92613bc6..97314f32 100644
--- a/go.mod
+++ b/go.mod
@@ -59,28 +59,29 @@ require (
 	github.com/tailscale/certstore v0.1.1-0.20220316223106-78d6e1c49d8d
 	github.com/tailscale/depaware v0.0.0-20210622194025-720c4b409502
 	github.com/tailscale/goexpect v0.0.0-20210902213824-6e8c725cea41
-	github.com/tailscale/golang-x-crypto v0.0.0-20221115211329-17a3db2c30d2
+	github.com/tailscale/golang-x-crypto v0.0.0-20230713185742-f0b76a10a08e
 	github.com/tailscale/goupnp v1.0.1-0.20210804011211-c64d0f06ea05
 	github.com/tailscale/hujson v0.0.0-20221223112325-20486734a56a
 	github.com/tailscale/mkctr v0.0.0-20220601142259-c0b937af2e89
 	github.com/tailscale/netlink v1.1.1-0.20211101221916-cabfb018fe85
-	github.com/tailscale/wireguard-go v0.0.0-20230410165232-af172621b4dd
+	github.com/tailscale/wireguard-go v0.0.0-20230710185534-bb2c8f22eccf
[...]

PR #8667, opened 2023-07-21
tstest/deptest: add test-only package to unify negative dep tests

bradfitz:
Updates #8658

### DIFF
--- a/cmd/tailscaled/tailscaled_test.go
+++ b/cmd/tailscaled/tailscaled_test.go
@@ -3,10 +3,34 @@
 
 package main // import "tailscale.com/cmd/tailscaled"
 
-import "testing"
+import (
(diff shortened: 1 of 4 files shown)


KevinLiang10:
We had the conversation today about what we are going to do with gvisor, there are two things we need to think of.
1. We will still need functions in there later, if we want to replace gvisor we have to implement our own.
2. There aren't quick replacements for gvisor

So question is shall we comment the function out (or even delete the function), or take some time to do the replacement?

bradfitz:
Well, let's have that discussion elsewhere. This PR is just removing some copy/paste and centralizing the helper code to write such dep tests.

Even if you keep end up fixing + keeping gvisor, we'll always have negative dep tests.

PR #8668, opened 2023-07-21
util/linuxfw: remove yet-unused code to fix linux/arm64 crash

### LINKED ISSUE #8658: tailscale-1.46.0 does not work on ARM64 systems with page size != 4k

bradfitz:
The util/linuxfw/iptables.go had a bunch of code that wasn't yet used
(in prep for future work) but because of its imports, ended up
initializing code deep within gvisor that panicked on init on arm64
systems not using 4KB pages.

This deletes the unused code to delete the imports and remove the
panic. We can then cherry-pick this back to the branch and restore it
later in a different way.

A new test makes sure we don't regress in the future by depending on
the panicking package in question.

Fixes #8658


KevinLiang10:
Went through the tests and ran the code, I think these achieve our purpose of forbidding `hostarch`. Thanks for the code reworking!

KevinLiang10 (approved):
Went through the tests and ran the code, I think these achieve our purpose of forbidding hostarch. Thanks for the code reworking!

PR #8674, opened 2023-07-21
1.46.1 cherry picks

bradfitz:
(3 commits: 1 for tests only + arm64 crash + ssh scp fix)


bradfitz:
cc @KevinLiang10 @dsnet

PR #8677, opened 2023-07-21
*: use tstime

clairew:
Updates #8587

Grouped together the smaller package refactors together


hostinfo/hostinfo.go
```diff
@@ -29,7 +30,8 @@ import (
 	"tailscale.com/version"
 )
 
-var started = time.Now()
+var clock = tstime.StdClock{}
```
bradfitz, line 33: What's the point of all these? You can't override them in init because their static type is `tstime.StdClock`. Did you mean to make the variables of the interface type?

PR #8678, opened 2023-07-21
cmd/tailscale/cli: implement update for dnf/yum-based distros

awly:
This is the Fedora family of distros, including CentOS, RHEL and others. Tested in `fedora:latest` and `centos:7` containers.

### DIFF
--- a/cmd/tailscale/cli/update.go
+++ b/cmd/tailscale/cli/update.go
@@ -144,7 +144,20 @@ func newUpdater() (*updater, error) {
 		case distro.Debian: // includes Ubuntu
 			up.update = up.updateDebLike
 		case distro.Arch:
-			up.update = up.updateArch
+			up.update = up.updateArchLike
+		}
+		// TODO(awly): add support for Alpine
+		switch {
+		case haveExecutable("pacman"):
+			up.update = up.updateArchLike
+		case haveExecutable("apt-get"): // TODO(awly): add support for "apt"
+			// The distro.Debian switch case above should catch most apt-based
+			// systems, but add this fallback just in case.
+			up.update = up.updateDebLike
+		case haveExecutable("dnf"):
+			up.update = up.updateFedoraLike("dnf")
+		case haveExecutable("yum"):
+			up.update = up.updateFedoraLike("yum")
 		}
 	case "darwin":
 		switch {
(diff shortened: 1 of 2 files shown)


cmd/tailscale/cli/update.go
```diff
@@ -209,46 +221,24 @@ func (up *updater) updateSynology() error {
...
 	if up.currentOrDryRun(ver) {
 		return nil
 	}
 
-	track := "unstable"
-	if stable, ok := versionIsStable(ver); !ok {
-		return fmt.Errorf("malformed version %q", ver)
-	} else if stable {
-		track = "stable"
-	}
```
awly, line 242: this is already handled here: https://github.com/tailscale/tailscale/blob/cb0e3b4497426d21a4d705a01ee0b6f140aa674f/cmd/tailscale/cli/update.go#L118-L132

cmd/tailscale/cli/update.go
```diff
@@ -209,46 +221,24 @@ func (up *updater) updateSynology() error {
...
-
 	if os.Geteuid() != 0 {
 		return errors.New("must be root; use sudo")
```
noncombatant, line 235: I keep thinking about this. It is bad to simply re-invoke ourselves with sudo?

And/or, should we re-tool the error message to talk about `doas` (OBSD at least, maybe others?) and `runas.exe` (Windows, when we get to that point)?

cmd/tailscale/cli/update.go
```diff
@@ -391,6 +381,93 @@ func parsePacmanVersion(out []byte) (string, error) {
...
+		}
+		defer func() {
+			if err != nil && !errors.Is(err, errUserAborted) {
+				err = fmt.Errorf(`%w; you can try updating using "%s upgrade tailscale"`, err, packageManager)
+			}
+		}()
+
+		ver := updateArgs.version
+		if ver == "" {
+			ver, err = latestTailscaleVersion(up.track)
```
noncombatant, line 402: This code is repeated now. Time to abstract it out into a helper function?

cmd/tailscale/cli/update.go
```diff
@@ -705,3 +782,32 @@ func (pw *progressWriter) print() {
...
+	return err == nil && path != ""
+}
+
+func latestTailscaleVersion(track string) (string, error) {
+	res, err := http.Get("https://pkgs.tailscale.com/" + track + "/?mode=json")
+	if err != nil {
+		return "", fmt.Errorf("fetching latest tailscale version: %w", err)
+	}
+	var latest struct {
+		Tarballs map[string]string // ~goarch (ignoring "geode") => "tailscale_1.34.2_mips.tgz"
```
noncombatant, line 797: This is now duplicated, too.

awly: Updated this to get the `Version` field from the JSON response (after tweaking pkgserve to always return it) and deduped everywhere

cmd/tailscale/cli/update.go
awly: Added a helper to check running as root and print an OS-appropriate message.
It doesn't look like `updateWindows` requires admin privileges, at least not explicitly in the code today.

cmd/tailscale/cli/update.go
```diff
@@ -144,7 +144,19 @@ func newUpdater() (*updater, error) {
 		case distro.Debian: // includes Ubuntu
 			up.update = up.updateDebLike
 		case distro.Arch:
-			up.update = up.updateArch
+			up.update = up.updateArchLike
+		}
+		switch {
+		case haveExecutable("pacman"):
+			up.update = up.updateArchLike
+		case haveExecutable("apt-get"):
```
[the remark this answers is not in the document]
awly: Left a TODO, since `updateDebLike` assumes `apt-get` and will fail if that's missing, even if `apt` is present.
Will add `apt` support in a separate PR.

noncombatant (approved):

Issue #7594, opened 2023-03-18
Kubernetes routing loop involving calico-node


tasuren:
### What is the issue?

I tried to use a private network created by Tailscale to build a cluster of k8s (Kubernetes).
Then, Tailscale running on two servers running worker nodes joined to the cluster started to consume an unnatural amount of network (tx 600Mbps) and CPU (80%, total 100%).
Why is this?

Log of tailscaled:
```shell
$ journalctl -fu tailscaled
 3月 18 12:26:21 ... tailscaled[60693]: magicsock: disco: failed to send *disco.CallMeMaybe to 127.3.3.40:7: too many DERP packets queued; dropping
 3月 18 12:26:27 ... tailscaled[60693]: Accept: UDP{100.97.160.100:32927 > 100.112.81.21:4789} 638 ok
 3月 18 12:26:34 ... tailscaled[60693]: monitor: RTM_DELROUTE: src=, dst=fe80::/64, gw=, outif=59, table=254
 3月 18 12:26:34 ... tailscaled[60693]: monitor: RTM_DELROUTE: src=, dst=fe80::/128, gw=, outif=59, table=255
 3月 18 12:26:34 ... tailscaled[60693]: monitor: RTM_DELROUTE: src=, dst=fe80::ecee:eeff:feee:eeee/128, gw=, outif=59, table=255
 3月 18 12:26:34 ... tailscaled[60693]: monitor: RTM_DELROUTE: src=, dst=ff00::/8, gw=, outif=59, table=255
 3月 18 12:26:34 ... tailscaled[60693]: magicsock: endpoints changed: <CP Node IP>:53641 (stun), <CP Node IP>:41641 (stun4localport), 10.96.2.0:41641 (local), 10.244.55.193:41641 (local), 192.168.3.22:41641 (local), [<Some IPv6>]:41641 (local)
 3月 18 12:26:34 ... tailscaled[60693]: magicsock: endpoints changed: <CP Node IP>:35358 (stun), <CP Node IP>:41641 (stun4localport), 10.96.2.0:41641 (local), 10.244.55.193:41641 (local), 192.168.3.22:41641 (local), [<Some IPv6>]:41641 (local)
 3月 18 12:26:37 ... tailscaled[60693]: Accept: UDP{100.112.81.21:47872 > 100.97.160.100:4789} 638 ok out
[...]

Issue #1552, opened 2021-03-21
Provide mechanism for users to exclude interfaces from disco + endpoint advertisements


bradfitz:
User wishes to exclude an interface from being used for disco + endpoint advertisement:

https://forum.tailscale.com/t/limit-the-network-interfaces-that-are-sent-as-endpoints-for-a-device/88/4

Perhaps we could add `TS_AVOID_PREFIX=10.5.4.0/24`, which people could add to their /etc/default/tailscale. And then we could check that IPSet for discoverEndpoints and sending disco messages.

An IPSet is probably easier than an interface name. But I suppose we could also accept an interface name and map it to a prefix on start-up (or on a timer or major link change?).

@danderson, thoughts?

danderson:
We already have a hack to skip ZeroTier interfaces - we should add Nebula interfaces to that as well.

More flags, more options: not a huge fan, but if we're getting to a world with a nicer CLI that can blend with daemon flags, I guess it's okay.

bradfitz:
It could be just an environment variable for now. But that's hard for Windows users and doesn't work at all currently with netext/sysext macOS.

bradfitz:
But I agree there needs to be a plan for CLI and config overall. I guess ignored interfaces really is account specific. I could imagine difficult tailnets wanting that to differ.

danderson:
I think we can get decent mileage out of a global option, although there's probably edge cases like "only use my static wireguard VPN on my personal account" or somesuch. But no way -> some way is an okay start.

DentonGentry:
Some overlap with feature request in https://github.com/tailscale/tailscale/issues/1066 (where the discussion is mostly about multilink and/or bonding)

timwhite:
If we aren't adding a flag, just adding hacks to ignore things like Zerotier, can we ignore existing wireguard tunnels as well? I have a P2P Wireguard link that shouldn't be used for tailscale to also shuffle traffic, double encryption isn't very nice on the CPU.

parasiteoflife:
In the meantime, is there any way to make Tailscale use a specific interface? I'm currently having this problem, Tailscale is trying to use an interface that has no Internet access.

DentonGentry:
Are you using 1.32.2? https://github.com/tailscale/tailscale/pull/5940 improved detection of which interface to use, and went into 1.32.2.

parasiteoflife:
[...]

PR #8682, opened 2023-07-22
Ignore Kubernetes network interfaces

oscrx:
I use Tailscale for the control plane node communication of my Kubernetes cluster.

A routing loop had been happening every so often. Resulting in major cpu spikes all over the cluster.

After some research I found out magicsock was preferring internal routes provided by Cilium.
I was following this issue in the hope it would solve my problem: https://github.com/tailscale/tailscale/issues/1552

That would be the ultimate solution, a flag where specific interfaces could be excluded.

Today I noticed this PR however.
https://github.com/tailscale/tailscale/pull/1210/files

And even with my little go experience I knew I could extend this exclusion of unwanted interfaces a little bit more.
I also found other users having the same issue with Calico: https://github.com/tailscale/tailscale/issues/7594

This would probably solve that issue as well.

I'm currently running a custom build of it and it runs silky smooth.

I hope this will be sufficient and am open for any feedback!


bradfitz:
While this fixes you, it likely breaks somebody else where this was their only or best path.

The real problem is the amplification. Why's that happening? Let's fix that instead.

oscrx:
The nodes in my cluster are spread out geografically and are not in the same networks. With Tailscale they can communicate with each other (with stable ip addresses). I use Cilium for internal cluster communication.

When tailscale decides the 'local' kubernetes interface is preffered it is actually using it's own tunnel causing the cpu to go to 100% utilisation.

 Functionality to exclude certain interfaces would be the better option. But the implementation of such functionality is not trivial (for me)

bradfitz:
> using its own tunnel causing the cpu to go to 100% utilisation.

But why?

What's the loop?

oscrx:
Sorry, I missed your reply.

I don't know what causes the loop.
I'm not even 100% sure if it is a loop.

I will try to describe the pattern that I see:

1. Tailscale sets up the network between the nodes.
2. Cilium sets up some tunnel interfaces for containers to communicate over the cluster.
3. Tailscale keeps trying to find the best path.
4. At some point it thinks that the Cilium tunnel interface is the most optimal connection.
5. The CPU usage explodes.

I know that the code doesn't feel optimal, but the 'skip Zerotier' code that is above it is basically the same.
Until the real cause is found this seemed like the best solution.

This is what the `cilium_host` interface looks like:

```shell
ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
    inet6 ::1/128 scope host 
2: ens18: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 1e:81:e9:0f:f5:07 brd ff:ff:ff:ff:ff:ff
    altname enp0s18
    inet 192.168.123.222/23 metric 100 brd 192.168.123.255 scope global dynamic ens18
       valid_lft 79705sec preferred_lft 79705sec
    inet6 2a02:a470:edcd:0:1c81:e9ff:fe0f:f507/64 scope global dynamic mngtmpaddr noprefixroute 
       valid_lft 86269sec preferred_lft 14269sec
    inet6 fe80::1c81:e9ff:fe0f:f507/64 scope link 
3: tailscale0: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1280 qdisc fq_codel state UNKNOWN group default qlen 500
    link/none 
    inet 100.64.0.7/32 scope global tailscale0
    inet6 fdd3:a7d6:38c5::7/128 scope global 
    inet6 fe80::fa47:cd49:67f:dad5/64 scope link stable-privacy 
4: cilium_net@cilium_host: <BROADCAST,MULTICAST,NOARP,UP,LOWER_UP> mtu 1280 qdisc noqueue state UP group default qlen 1000
    link/ether 56:2d:30:d9:40:34 brd ff:ff:ff:ff:ff:ff
    inet6 fe80::542d:30ff:fed9:4034/64 scope link 
5: cilium_host@cilium_net: <BROADCAST,MULTICAST,NOARP,UP,LOWER_UP> mtu 1280 qdisc noqueue state UP group default qlen 1000
    link/ether 9a:3f:b7:88:1b:b6 brd ff:ff:ff:ff:ff:ff
    inet 10.0.1.193/32 scope global cilium_host
    inet6 fdd3:a7d6:38c5::7/128 scope global 
6: cilium_vxlan: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1280 qdisc noqueue state UNKNOWN group default qlen 1000
    link/ether b2:36:7a:fc:fb:83 brd ff:ff:ff:ff:ff:ff
    inet6 fe80::b036:7aff:fefc:fb83/64 scope link 
1554: lxc307eef7f217a@if1553: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1280 qdisc noqueue state UP group default qlen 1000
    link/ether f2:32:7d:71:15:a5 brd ff:ff:ff:ff:ff:ff link-netns 1a8faef5-4ca0-48d3-831a-c8aecc29a0c0
    inet6 fe80::f032:7dff:fe71:15a5/64 scope link 
...
```

Unfortunately I don't seem to have the logs of the issue anymore, but I posted a screenshot of the problem a few months back: https://github.com/tailscale/tailscale/issues/1552#issuecomment-1368864133 (10.240.0.0/16 was my cilium range at the time)
Notice the huge traffic volume just after restart of the Tailscale daemons.

net/interfaces/interfaces.go
```diff
@@ -79,6 +79,14 @@ func isProblematicInterface(nif *net.Interface) bool {
 	if strings.HasPrefix(name, "zt") || (runtime.GOOS == "windows" && strings.Contains(name, "ZeroTier")) {
 		return true
 	}
+	// Don't try to send disco/etc packets over Cilium or Calico; they effectively
+	// DoS each other by doing traffic amplification, both of them
+	// preferring/trying to use each other for transport. See:
+	// https://github.com/tailscale/tailscale/issues/7594
+	// https://github.com/tailscale/tailscale/issues/1552
+	if strings.HasPrefix(name, "cilium") || strings.HasPrefix(name, "cali") {
```
unixfox, line 87: Do note that if you truly want to exclude all the network interfaces created by a Kubernetes CNI, the list is going to be huge.

For instance, for flannel it's `flannel.X`.

I feel like there should be a better way than excluding predefined interface names.

oscrx: Maybe a config option to prevent this behavior would be better.
Like the one discussed here: https://github.com/tailscale/tailscale/issues/1552

justyns:
FWIW this PR or something similar is still desired.  I am running into something similar where tailscale occasionally tries to route over cilium (which is already using the tailscale interface) and causes high cpu usage and connectivity blips.

PR #8688, opened 2023-07-23
net/dns: overwrite /tmp/resolv.conf on gokrazy

stapelberg:
Appliances built using https://gokrazy.org/ have a read-only root file system, including /etc/resolv.conf, which is a symlink to /tmp/resolv.conf.

The system’s dhcp client overwrites /tmp/resolv.conf instead, so we need to use this path in Tailscale, too.

An alternative would have been to make the code follow symlinks when handling /etc/resolv.conf, but this fix feels a little simpler — less potential to break something and less code complexity.

related to https://github.com/gokrazy/gokrazy/issues/209

### DIFF
--- a/net/dns/resolvconfpath_default.go
+++ b/net/dns/resolvconfpath_default.go
@@ -1,0 +1,11 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+//go:build !gokrazy
+
+package dns
+
+const (
+	resolvConf = "/etc/resolv.conf"
+	backupConf = "/etc/resolv.pre-tailscale-backup.conf"
+)
--- a/net/dns/direct.go
+++ b/net/dns/direct.go
@@ -27,11 +27,6 @@ import (
 	"tailscale.com/version/distro"
 )
 
-const (
-	backupConf = "/etc/resolv.pre-tailscale-backup.conf"
-	resolvConf = "/etc/resolv.conf"
-)
-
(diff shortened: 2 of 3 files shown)


net/dns/resolvconfpath_default.go
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+//go:build !gokrazy
```
bradfitz, line 4: We don't use this build tag anywhere else and this isn't a GOOS.

Let's do this at runtime with a comparison to distro.Get() == distro.Gokrazy instead.

bradfitz: If you want to minimize how often/ugly that looks doing it at runtime, you can do it just in the DNS manager constructor and add new fields for these two strings.

bradfitz: Concrete counter-proposal: #8690

bradfitz: I guess the gokrazy build tool always sets the `gokrazy` tag, huh? So maybe this PR is fine.

I think I have a slight preference for not using build tags if we can avoid it, though.

bradfitz: But let's get your change in first. LGTM if it works for gokr.

bradfitz (approved):

crawshaw:
FYI following symlinks would cause a lot of trouble with systemd which uses a symlink to a file it controls as a signal of ownership. So this strategy seems a lot safer

PR #8690, opened 2023-07-23
net/dns: make directManager use /tmp/resolv.conf on gokrazy

### LINKED ISSUE #8689: Magic DNS does not work on gokrazy devices
### What is the issue?

On gokrazy, Magic DNS does not work because /etc/resolv.conf cannot be updated (the root file system is read-only in gokrazy).

### Steps to reproduce

Follow https://gokrazy.org/packages/tailscale/

Notice that Magic DNS does not work.

n/a

### OS

Linux

### OS version

gokrazy

### Tailscale version

1.46.0

### Other software

_No response_

### Bug report

_No response_

bradfitz:
Appliances built using https://gokrazy.org/ have a read-only root file system, including /etc/resolv.conf, which is a symlink to /tmp/resolv.conf.

The system’s dhcp client overwrites /tmp/resolv.conf instead, so we need to use this path in Tailscale, too.

Fixes #8689
Updates gokrazy/gokrazy#209

### DIFF
--- a/net/dns/direct.go
+++ b/net/dns/direct.go
@@ -28,8 +28,8 @@ import (
 )
 
 const (
-	backupConf = "/etc/resolv.pre-tailscale-backup.conf"
-	resolvConf = "/etc/resolv.conf"
+	defaultBackupConf = "/etc/resolv.pre-tailscale-backup.conf"
+	defaultResolvConf = "/etc/resolv.conf"
 )
 
 // writeResolvConf writes DNS configuration in resolv.conf format to the given writer.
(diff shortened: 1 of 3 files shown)


bradfitz:
cc @stapelberg

bradfitz:
I could go either way on this; #8688 is probably fine. We could always do this in the future if/when we need it for other distros or for testing or something.

stapelberg:
Thanks for the PR! I successfully verified that this approach works just as well. Either way is fine with me — this approach might be a little easier to work with in terms of cross-referencing (always a bit awkward with multiple build tags), so feel free to merge :)

stapelberg (approved):

PR #8693, opened 2023-07-23
tailcfg: add UserProfile.Groups

bradfitz:
Updates tailscale/corp#13375

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -3,7 +3,7 @@
 
 package tailcfg
 
-//go:generate go run tailscale.com/cmd/viewer --type=User,Node,Hostinfo,NetInfo,Login,DNSConfig,RegisterResponse,DERPHomeParams,DERPRegion,DERPMap,DERPNode,SSHRule,SSHAction,SSHPrincipal,ControlDialPlan,Location --clonefunc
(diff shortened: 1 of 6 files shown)


tailcfg/tailcfg.go
```diff
@@ -175,6 +175,21 @@ type UserProfile struct {
 	// Roles exists for legacy reasons, to keep old macOS clients
 	// happy. It JSON marshals as [].
 	Roles emptyStructJSONSlice
+
+	// Groups contains group identifiers for any group that this user is
+	// a part of and that the coordination server is configured to tell
+	// your nodes about. (Thus, it may be empty or incomplete.)
+	// There's no difference between a nil vs an empty list.
+	// The list is always be sorted.
```
maisem, line 183: Grammar

maisem (approved):

tailcfg/tailcfg.go
bradfitz: words

tailcfg/tailcfg.go
```diff
@@ -175,6 +175,21 @@ type UserProfile struct {
 	// Roles exists for legacy reasons, to keep old macOS clients
 	// happy. It JSON marshals as [].
 	Roles emptyStructJSONSlice
+
+	// Groups contains group identifiers for any group that this user is
```
maisem, line 179: Could also bump the cap ver so that we don't send it down to clients who don't understand it

bradfitz: done

PR #8700, opened 2023-07-24
cmd/tailscale/cli: add latest version output to "tailscale version"

### LINKED ISSUE #8669: FR: tailscale version in machine readable/json format

Trying to write some automation scripts, it would be great to be able to provide tailscale version installed locally, and version latest version available in json format.

### How should we solve this?

either under tailscale version --json provide local and latest keys

alternatively under tailscale update --json

harder to write automation scripts and deployments.

### Anything else?

_No response_

awly:
Add optional `--with-latest` flag to `tailscale version` to fetch the latest release version from `pkgs.tailscale.com`. This is useful to diagnose `tailscale update` behavior or write other tooling.

Example output:
```sh
$ tailscale version --with-latest --json
{
	"majorMinorPatch": "1.47.35",
	"short": "1.47.35",
	"long": "1.47.35-t6afffece8",
	"unstableBranch": true,
	"gitCommit": "6afffece8a32509aa7a4dc2972415ec58d8316de",
	"cap": 66,
	"latest": "1.45.61"
}
```

Fixes #8669

### DIFF
--- a/cmd/tailscale/cli/version.go
+++ b/cmd/tailscale/cli/version.go
@@ -23,14 +23,16 @@ var versionCmd = &ffcli.Command{
 		fs := newFlagSet("version")
 		fs.BoolVar(&versionArgs.daemon, "daemon", false, "also print local node's daemon version")
 		fs.BoolVar(&versionArgs.json, "json", false, "output in JSON format")
+		fs.BoolVar(&versionArgs.upstream, "upstream", false, "fetch and print the latest upstream release version from pkgs.tailscale.com")
 		return fs
 	})(),
 	Exec: runVersion,
 }
 
 var versionArgs struct {
-	daemon bool // also check local node's daemon version
-	json   bool
+	daemon   bool // also check local node's daemon version
+	json     bool
+	upstream bool
 }
 
 func runVersion(ctx context.Context, args []string) error {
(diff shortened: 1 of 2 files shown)


cmd/tailscale/cli/version.go
```diff
@@ -23,14 +23,16 @@ var versionCmd = &ffcli.Command{
 		fs := newFlagSet("version")
 		fs.BoolVar(&versionArgs.daemon, "daemon", false, "also print local node's daemon version")
 		fs.BoolVar(&versionArgs.json, "json", false, "output in JSON format")
+		fs.BoolVar(&versionArgs.withLatest, "with-latest", false, "include latest released version output")
```
bradfitz, line 26: say something about how it hits the network?

also, the "with" here doesn't seem consistent: "daemon" isn't "with-daemon". Not that I have other suggestions at the moment.

awly: Changed flag to `--upstream` and updated comment to say it's fetching the version from `pkgs.tailscale.com`

bradfitz (approved):

bradfitz:
btw, if you add a line to the commit message like:

```
RELNOTE=adds tailscale version --upstream
```

Then we'll find it for the next release notes.

PR #8709, opened 2023-07-25
tailcfg,ipn/ipnlocal,wgengine: add values to PeerCapabilities

maisem:
Define PeerCapabilty and PeerCapMap as the new way of sending down inter-peer capability information.

Previously, this was unstructured and you could only send down strings which got too limiting for certain usecases. Instead add the ability to send down raw JSON messages that are opaque to Tailscale but provide the applications to define them however they wish.

Also update accessors to use the new values.

Updates #4217

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -8,6 +8,7 @@ package tailcfg
 import (
 	"bytes"
 	"encoding/hex"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/netip"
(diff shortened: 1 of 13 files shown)


tailcfg/tailcfg.go
```diff
@@ -1182,7 +1183,48 @@ type CapGrant struct {
...
+)
+
+// PeerCapMap is a map of capabilities to their optional values. It is valid for a
+// capability to have no values (nil slice).
+//
+// The values are opaque to Tailscale, but are passed through from the ACLs to
+// the application via the WhoIs API.
+type PeerCapMap map[PeerCapability][]json.RawMessage
+
+// HasCapability reports whether c has the capability cap.
```
bradfitz, line 1224: and say this is meant for capabilities which don't have values.

tailcfg/tailcfg.go
```diff
@@ -1182,7 +1183,48 @@ type CapGrant struct {
...
+	// goroutines, metrics, magicsock internal state, etc.
+	PeerCapabilityDebugPeer PeerCapability = "https://tailscale.com/cap/debug-peer"
+	// PeerCapabilityWakeOnLAN grants the ability to send a Wake-On-LAN packet.
+	PeerCapabilityWakeOnLAN PeerCapability = "https://tailscale.com/cap/wake-on-lan"
+	// PeerCapabilityIngress grants the ability for a peer to send ingress traffic.
+	PeerCapabilityIngress PeerCapability = "https://tailscale.com/cap/ingress"
+)
+
+// PeerCapMap is a map of capabilities to their optional values. It is valid for a
+// capability to have no values (nil slice).
```
bradfitz, line 1218: (nil slice); such capabilities can be tested for by using the HasCapability method.

bradfitz (approved):
and bump capver

PR #8710, opened 2023-07-25
cmd/tailscale/cli: implement update on FreeBSD

awly:
Implement `tailscale update` on FreeBSD. This is much simpler than other platforms because `pkg rquery` lets us get the version in their repos without any extra parsing.

Updates #6995


cmd/tailscale/cli/update.go
```diff
@@ -45,10 +45,13 @@ var updateCmd = &ffcli.Command{
 		fs.BoolVar(&updateArgs.yes, "yes", false, "update without interactive prompts")
 		fs.BoolVar(&updateArgs.dryRun, "dry-run", false, "print what update would do without doing it, or prompts")
 		fs.BoolVar(&updateArgs.appStore, "app-store", false, "HIDDEN: check the App Store for updates, even if this is not an App Store install (for testing only)")
-		// These flags are not supported on Arch or Alpine-based installs.
-		// Package repos on these distros only offer one variant of tailscale
-		// and it's always the latest version.
-		if distro.Get() != distro.Arch && distro.Get() != distro.Alpine {
+		// These flags are not supported on several distros that only provide
```
noncombatant, line 48: Minor: Maybe "systems" instead of "distros" now, since it's not just Linux.

noncombatant (approved):

PR #8715, opened 2023-07-26
magicsock: set the don't fragment sockopt

saljam:
This sets the Don't Fragment flag, for now behind the TS_DEBUG_PMTUD envknob.

This is a prerequisite for PMTUD.

Updates #311.

### DIFF
--- a/wgengine/magicsock/debugknobs.go
+++ b/wgengine/magicsock/debugknobs.go
@@ -47,6 +47,8 @@ var (
 	// debugRingBufferMaxSizeBytes overrides the default size of the endpoint
 	// history ringbuffer.
 	debugRingBufferMaxSizeBytes = envknob.RegisterInt("TS_DEBUG_MAGICSOCK_RING_BUFFER_MAX_SIZE_BYTES")
+	// debugPMTUD enables path MTU discovery. Currently only sets the Don't Fragment sockopt.
+	debugPMTUD = envknob.RegisterBool("TS_DEBUG_ENABLE_PMTUD")
 	// Hey you! Adding a new debugknob? Make sure to stub it out in the debugknob_stubs.go
 	// file too.
 )
--- a/wgengine/magicsock/dontfrag_darwin.go
+++ b/wgengine/magicsock/dontfrag_darwin.go
@@ -1,0 +1,35 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+//go:build darwin && !ios
+
+package magicsock
+
+import (
+	"net"
+	"syscall"
+
+	"golang.org/x/sys/unix"
+	"tailscale.com/types/nettype"
+)
+
+func setDontFragment(pconn nettype.PacketConn, network string) (err error) {
+	if c, ok := pconn.(*net.UDPConn); ok {
+		rc, err := c.SyscallConn()
+		if err == nil {
+			rc.Control(func(fd uintptr) {
+				if network == "udp4" {
+					err = syscall.SetsockoptInt(int(fd), syscall.IPPROTO_IP, unix.IP_DONTFRAG, 1)
+				}
+				if network == "udp6" {
+					err = syscall.SetsockoptInt(int(fd), syscall.IPPROTO_IPV6, unix.IPV6_DONTFRAG, 1)
+				}
+			})
+		}
+	}
+	return err
+}
+
+func CanPMTUD() bool {
+	return debugPMTUD() // only if the envknob is for now.
+}
--- a/wgengine/magicsock/dontfrag_default.go
+++ b/wgengine/magicsock/dontfrag_default.go
@@ -1,0 +1,24 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+//go:build (!linux && !darwin) || android || ios
+
+package magicsock
+
+import (
+	"errors"
+
+	"tailscale.com/types/nettype"
+)
+
+// setDontFragment sets the dontfragment sockopt on pconn on the platforms that support it,
+// for both IPv4 and IPv6.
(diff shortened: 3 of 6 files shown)


wgengine/magicsock/dontfrag_darwin.go
```diff
@@ -0,0 +1,34 @@
...
+	"net"
+	"syscall"
+
+	"tailscale.com/types/nettype"
+)
+
+const (
+	// From https://opensource.apple.com/source/xnu/xnu-6153.141.1/bsd/netinet6/in6.h.auto.html
+	socketOptionIPDontFrag   = 28
+	socketOptionIPv6DontFrag = 62
```
valscale, line 16: So I finally looked up why we have to do this and IPv6 still automatically fragments packets sent on UDP and raw sockets! I guess it's just TCP packets that don't get fragmented on IPv6. So maybe expand this comment to add the below link and the above fact. And maybe a TODO on getting these into the golang sockopt library?

https://datatracker.ietf.org/doc/html/rfc3542#section-11.2

wgengine/magicsock/magicsock.go
```diff
@@ -2238,6 +2238,14 @@ func (c *Conn) bindSocket(ruc *RebindingUDPConn, network string, curPortFate cur
 			continue
 		}
 		trySetSocketBuffer(pconn, c.logf)
+
+		if debugPMTUD() {
+			err = setDontFragment(pconn, network)
+			if err != nil {
+				c.logf("magicsock: unable disable fragmentation for %v port %d: %v", network, port, err)
```
valscale, line 2245: Tiny typo: "unable disable" -> "unable to disable"

wgengine/magicsock/debugknobs.go
```diff
@@ -47,6 +47,8 @@ var (
 	// debugRingBufferMaxSizeBytes overrides the default size of the endpoint
 	// history ringbuffer.
 	debugRingBufferMaxSizeBytes = envknob.RegisterInt("TS_DEBUG_MAGICSOCK_RING_BUFFER_MAX_SIZE_BYTES")
+	// debugPMTUD enables path MTU discovery. Currently only sets the Don't Fragment sockopt.
+	debugPMTUD = envknob.RegisterBool("TS_DEBUG_PMTUD")
```
valscale, line 51: Maybe TS_DEBUG_PMTUD -> TS_DEBUG_ENABLE_PMTUD ? Matches some other debugknobs. But if you do you will want to update the commit message.

valscale (approved):
Thanks for making this so much more readable!

wgengine/magicsock/dontfrag_darwin.go
raggi: Yeah, the subtlety you probably are aware of is that UDP datagrams never had a traditional association with MTU, unlike the MSS exchange with TCP, so fragmentation is/was a necessary evil to avoid changing in-flow semantics in the userspace APIs.

I'm not fully up to date on the IPv6 RFC's but I'd hazard a guess that fragmentation is coming back for v6 TCP as well as part of the MPTCP work - just a guess though.

wgengine/magicsock/dontfrag_default.go
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+//go:build !linux && !darwin
```
raggi, line 4: Given this platform constraint, are you going to introduce an exported function that other code will use to determine if PMTU can/can't be turned on for a particular platform?

wgengine/magicsock/magicsock.go
raggi: perhaps something like: `magicsock: set dontfragment failed: %v...`

raggi (commented):
Mostly LGTM. I think it might be good to plumb something like:

```
magicsock.CanPMTU()
```

which can return false if the platform is unsupported (or other feature detections fail) and then use that to avoid printing an error, so the setup condition being:

```
CanPMTU() && ShouldPMTU()
```

or the like

wgengine/magicsock/dontfrag_darwin.go
raggi: There's a subtle drop-down on pkg.go.dev that adds ?GOOS=darwin to the url and renders that goos

valscale (approved):

wgengine/magicsock/magicsock.go
```diff
@@ -2233,6 +2233,16 @@ func (c *Conn) bindSocket(ruc *RebindingUDPConn, network string, curPortFate cur
 			continue
 		}
 		trySetSocketBuffer(pconn, c.logf)
+
+		if CanPMTUD() {
+			err = setDontFragment(pconn, network)
+			if err != nil {
+				c.logf("magicsock: set dontfragment failed for %v port %d: %v", network, port, err)
+				// TODO disable PMTUD in this case. We don't expect the setsockopt to fail on
```
raggi, line 2241: What happens on Android? Have they allowlisted the sockopt?

raggi (approved):
Assuming you have a mitigation solution for Android, LGTM.

wgengine/magicsock/magicsock.go
raggi: That's a good question. I think the unfortunately even more subtle answer is that we want to disable PMTU when we're on battery and on some LTE or lesser protocol radios. Those are definitely harder but also quite interesting conditions to determine so I suspect we'll want to defer them. Upcoming pervasive 5G deployments, and wifi could benefit, and so it is worth considering in time.

raggi: FWIW, this is where I would introduce a separation between "can" and "should", and "can" would be set here on failure to to "can not"

valscale: That makes sense. I think we should change the build to exclude android for now and then get fancy with battery, LTE, wifi later on.

PR #8719, opened 2023-07-26
net/netcheck, wgengine/magicsock: bound DERP latency by TCP RTT

andrew-d:
For discussion; will add tests and flesh out further if this is something we're intending to merge.

Change-Id: Ie554364227ce7973644025129b8046fc7df8fb35

### DIFF
--- a/derp/derp_client.go
+++ b/derp/derp_client.go
@@ -10,12 +10,14 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/netip"
 	"sync"
 	"time"
 
 	"go4.org/mem"
 	"golang.org/x/time/rate"
+	"tailscale.com/net/tcpinfo"
 	"tailscale.com/syncs"
 	"tailscale.com/types/key"
 	"tailscale.com/types/logger"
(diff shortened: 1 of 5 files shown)


valscale:
I like relying on the RTT when we have it. I am wondering what happens when latency to a derp actually does increase, or we lose connectivity entirely? This code seems like it can only adjust the latency down, which makes me wonder why we aren't using the RTT directly when it is available. Do we have a sense of how quickly the RTT changes when the actual connectivity to the derp changes? Basically I'm trying to figure out how to best combine the RTT info with our most recent latency check rather than throwing away most of the info from the latency check.

raggi:
Have you sampled accuracy in the relevant challenging conditions? This is the RTT estimate from the congestion control system, it's not sampled by performing active round trips.

Example pulled from a derper: rtt:311.541/53.151 / rtt:233.216/37.389 /  rtt:245.848/110.86 / rtt:19.503/9.587 / rtt:1.435/0.464

valscale (commented):
Github is bugging me for a review because I put my comment in a comment, so here is a review. :)

PR #8731, opened 2023-07-27
ipn/{ipnlocal,localapi}: actually renew certs before expiry

### LINKED ISSUE #8725: TLS certificate issued with tailscale cert does not renew before expiry
### What is the issue?

Requesting a TLS certificate with `tailscale cert` that is due for renewal (whether by the original 'expiry-within-14-days' logic, the new '2/3 of lifetime' logic, or using ARI) results in the `starting async renewal` message being logged, but the cert is not actually renewed, and subsequent calls to `tailscale cert` return the same certificate, _until_ the certificate has actually expired.

### Steps to reproduce

1. Observe that a TLS certificate issued with `tailscale cert` is expiring (according to the logic relevant to that version of the Tailscale client).
2. Call `tailscale cert`.
3. Observe the `starting async renewal` log entry.
4. Wait a suitable period of time for the async renewal process that we presume is happening to complete.
5. Call `tailscale cert` again and observe that the certificate has not, in fact, been renewed. Further observe that there are no relevant log entries following `starting async renewal`, implying that the async renewal process did not, in fact, happen.

This _doesn't_ seem to have been triggered by #8204, although that issue appears to have shed light on this problem (there are several comments there reporting this behavior as well).  My amateur diagnosis, after a few minutes perusing `cert.go`, is that the problem lies [in the `getCertPEM()` function](https://github.com/tailscale/tailscale/blob/b3138a71ad0b2fb7d99f0dcecc5e4bc85483bb4f/ipn/ipnlocal/cert.go#L321) (which is what is called in a goroutine immediately after the `starting async renewal` message), which in turn calls `getCertPEMCached()` early in its execution, _passing the current time as the time for the validity test_ (which winds up coming from [here](https://github.com/tailscale/tailscale/blob/35bdbeda9fa149bd6caf93b30d71dd767d16d4b6/ipn/ipnlocal/cert.go#L90C3-L90C3)).  Thus even if `getCertPEM()` has been called explicitly for the purpose of renewing the certificate, if that certificate is still valid, it will return the cached certificate and do no more.

Still in the realm of amateur diagnosis, it seems like this could be addressed by adding a flag to `getCertPEM()` to bypass the cache when it has been called for the explicit purpose of performing a renewal, or, passing a manifestly invalid (e.g. far future or past) `now` time, so that the cache test cannot succeed in those cases and a new certificate will always be fetched.

(Arguably `getCertPEM()` should _never_ look for a cached certificate, since `GetCertPEM()` (the public API) already [looks for a cached certificate](https://github.com/tailscale/tailscale/blob/b3138a71ad0b2fb7d99f0dcecc5e4bc85483bb4f/ipn/ipnlocal/cert.go#L103C22-L103C22) first.)

### OS

Linux, Synology

### OS version

Debian 11

### Tailscale version

1.46.0

### Other software

_No response_

### Bug report

BUG-6af6aba81615e09513ecb877765b58b71235444a83b4614a5fc5f63525bdfdfd-20230727033840Z-c13db33870051c71

awly:
While our `shouldStartDomainRenewal` check is correct, `getCertPEM` would always bail if the existing cert is not expired. Add the same `shouldStartDomainRenewal` check to `getCertPEM` to make it proceed with renewal when existing certs are still valid but should be renewed.

The extra check is expensive (ARI request towards LetsEncrypt), so cache the last check result to not degrade `tailscale serve` performance.

Also, asynchronous renewal is great for `tailscale serve` but confusing for `tailscale cert`. Add an explicit flag to `GetCertPEM` to force a synchronous renewal for `tailscale cert`.

Tested this manually by stubbing out `shouldStartDomainRenewal` and running `tailscale cert` multiple times. Also tested renewal check caching via `tailscale serve` and looking through logs to make sure we don't hit ARI every time.

Fixes #8725

### DIFF
--- a/ipn/ipnlocal/cert.go
+++ b/ipn/ipnlocal/cert.go
@@ -53,8 +53,8 @@ var (
 	// populate the on-disk cache and the rest should use that.
 	acmeMu sync.Mutex
 
-	renewMu        sync.Mutex // lock order: don't hold acmeMu and renewMu at the same time
-	lastRenewCheck = map[string]time.Time{}
+	renewMu     sync.Mutex // lock order: acmeMu before renewMu
+	renewCertAt = map[string]time.Time{}
 )
 
 // certDir returns (creating if needed) the directory in which cached
@@ -80,9 +80,15 @@ func (b *LocalBackend) certDir() (string, error) {
 
 var acmeDebug = envknob.RegisterBool("TS_DEBUG_ACME")
 
-// getCertPEM gets the KeyPair for domain, either from cache, via the ACME
-// process, or from cache and kicking off an async ACME renewal.
-func (b *LocalBackend) GetCertPEM(ctx context.Context, domain string) (*TLSCertKeyPair, error) {
+// GetCertPEM gets the TLSCertKeyPair for domain, either from cache or via the
+// ACME process. ACME process is used for new domain certs, existing expired
+// certs or existing certs that should get renewed due to upcoming expiry.
+//
+// syncRenewal changes renewal behavior for existing certs that are still valid
+// but need renewal. When syncRenewal is set, the method blocks until a new
+// cert is issued. When syncRenewal is not set, existing cert is returned right
+// away and renewal is kicked off in a background goroutine.
+func (b *LocalBackend) GetCertPEM(ctx context.Context, domain string, syncRenewal bool) (*TLSCertKeyPair, error) {
 	if !validLookingCertDomain(domain) {
 		return nil, errors.New("invalid domain")
 	}
@@ -105,12 +111,15 @@ func (b *LocalBackend) GetCertPEM(ctx context.Context, domain string) (*TLSCertK
 		shouldRenew, err := b.shouldStartDomainRenewal(cs, domain, now, pair)
 		if err != nil {
 			logf("error checking for certificate renewal: %v", err)
-		} else if shouldRenew {
+		} else if !shouldRenew {
+			return pair, nil
+		}
+		if !syncRenewal {
 			logf("starting async renewal")
 			// Start renewal in the background.
 			go b.getCertPEM(context.Background(), cs, logf, traceACME, domain, now)
 		}
-		return pair, nil
+		// Synchronous renewal happens below.
 	}
 
 	pair, err := b.getCertPEM(ctx, cs, logf, traceACME, domain, now)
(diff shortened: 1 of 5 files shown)


ipn/ipnlocal/cert.go
```diff
@@ -121,17 +136,21 @@ func (b *LocalBackend) GetCertPEM(ctx context.Context, domain string) (*TLSCertK
...
+func (b *LocalBackend) shouldStartDomainRenewal(cs certStore, domain string, now time.Time, pair *TLSCertKeyPair) (renew bool, err error) {
 	renewMu.Lock()
 	defer renewMu.Unlock()
-	if last, ok := lastRenewCheck[domain]; ok && now.Sub(last) < time.Minute {
-		// We checked very recently. Don't bother reparsing &
-		// validating the x509 cert.
-		return false, nil
-	}
-	lastRenewCheck[domain] = now
+	if last, ok := lastRenewCheck[domain]; ok && now.Sub(last.time) < time.Hour {
```
maisem, line 142: i think this is still too frequent, we get the SuggestedWindow from LE. We should cache until then IMO.

ipn/ipnlocal/cert.go
```diff
@@ -53,10 +54,15 @@ var (
...
 	acmeMu sync.Mutex
 
-	renewMu        sync.Mutex // lock order: don't hold acmeMu and renewMu at the same time
-	lastRenewCheck = map[string]time.Time{}
+	renewMu        sync.Mutex // lock order: don't lock acmeMu while renewMu is locked
+	lastRenewCheck = map[string]renewCheckResult{}
 )
 
+type renewCheckResult struct {
+	cacheUntil time.Time
```
maisem, line 62: maybe call this `renewAt` and `needsRenewal`?

maisem: actually you could probably only cache `renewAt`

ipn/ipnlocal/cert.go
```diff
@@ -124,37 +139,40 @@ func (b *LocalBackend) GetCertPEM(ctx context.Context, domain string) (*TLSCertK
...
 	renewMu.Lock()
 	defer renewMu.Unlock()
-	if last, ok := lastRenewCheck[domain]; ok && now.Sub(last) < time.Minute {
-		// We checked very recently. Don't bother reparsing &
-		// validating the x509 cert.
-		return false, nil
+	if last, ok := lastRenewCheck[domain]; ok && now.Before(last.cacheUntil) {
+		// Don't bother reparsing & validating the x509 cert until we know it's
+		// needed.
+		return last.res, nil
```
maisem, line 145: yeah you're only using it here, and this an be replaced by `now.Before(renewAt)`

ipn/ipnlocal/cert.go
```diff
@@ -53,8 +53,8 @@ var (
 	// populate the on-disk cache and the rest should use that.
 	acmeMu sync.Mutex
 
-	renewMu        sync.Mutex // lock order: don't hold acmeMu and renewMu at the same time
-	lastRenewCheck = map[string]time.Time{}
+	renewMu     sync.Mutex // lock order: don't lock acmeMu while renewMu is locked
```
maisem, line 56: ```// lock order: acmeMu before renewMu```

maisem: Convention is to say the order in which they should be acquired if both need to be acquired, if any one of them needs to be acquired independently, that is allowed.

maisem (approved):

PR #8739, opened 2023-07-28
wgengine,ipn,cmd/tailscale: add size option to ping

saljam:
This adds the capability to pad disco ping message payloads to reach a specified size. It also plumbs it through to the tailscale ping -size flag.

Disco pings used for actual endpoint discovery do not use this yet.

Updates #311.

Older endpoints respond to larger pings as normal:
```
% go run ./cmd/tailscale ping -size 500 saldev # new client, pong has same size as ping
pong from saldev (100.88.76.100) via [redacted]:64638 in 34ms

% go run ./cmd/tailscale ping -size 500 xn--mgbx5ceg # old client, pong has the usual size
pong from xn--mgbx5ceg (100.119.252.109) via [redacted]:51213 in 22ms
```

### DIFF
--- a/client/tailscale/localclient.go
+++ b/client/tailscale/localclient.go
@@ -807,17 +807,37 @@ func (lc *LocalClient) ExpandSNIName(ctx context.Context, name string) (fqdn str
 	return "", false
 }
 
+// PingOpts contains options for the ping request.
+//
+// The zero value is valid, which means to use defaults.
+type PingOpts struct {
+	// Size is the length of the ping message in bytes. It's ignored if it's
+	// smaller than the minimum message size.
+	//
+	// For disco pings, it specifies the length of the packet's payload. That
+	// is, it includes the disco headers and message, but not the IP and UDP
+	// headers.
+	Size int
+}
+
 // Ping sends a ping of the provided type to the provided IP and waits
-// for its response.
-func (lc *LocalClient) Ping(ctx context.Context, ip netip.Addr, pingtype tailcfg.PingType) (*ipnstate.PingResult, error) {
+// for its response. The opts type specifies additional options.
+func (lc *LocalClient) PingWithOpts(ctx context.Context, ip netip.Addr, pingtype tailcfg.PingType, opts PingOpts) (*ipnstate.PingResult, error) {
 	v := url.Values{}
 	v.Set("ip", ip.String())
+	v.Set("size", strconv.Itoa(opts.Size))
 	v.Set("type", string(pingtype))
 	body, err := lc.send(ctx, "POST", "/localapi/v0/ping?"+v.Encode(), 200, nil)
 	if err != nil {
 		return nil, fmt.Errorf("error %w: %s", err, body)
 	}
 	return decodeJSON[*ipnstate.PingResult](body)
+}
+
+// Ping sends a ping of the provided type to the provided IP and waits
+// for its response.
+func (lc *LocalClient) Ping(ctx context.Context, ip netip.Addr, pingtype tailcfg.PingType) (*ipnstate.PingResult, error) {
+	return lc.PingWithOpts(ctx, ip, pingtype, PingOpts{})
 }
 
 // NetworkLockStatus fetches information about the tailnet key authority, if one is configured.
--- a/cmd/tailscale/cli/ping.go
+++ b/cmd/tailscale/cli/ping.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/peterbourgon/ff/v3/ffcli"
+	"tailscale.com/client/tailscale"
 	"tailscale.com/ipn/ipnstate"
 	"tailscale.com/tailcfg"
 )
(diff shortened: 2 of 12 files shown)


client/tailscale/localclient.go
```diff
@@ -807,11 +807,12 @@ func (lc *LocalClient) ExpandSNIName(ctx context.Context, name string) (fqdn str
 	return "", false
 }
 
-// Ping sends a ping of the provided type to the provided IP and waits
+// Ping sends a ping of the provided type and size to the provided IP and waits
```
bradfitz, line 810: document more what size is. The size of what part?

bradfitz: And I assume zero is safe here too?

wgengine/userspace.go
```diff
@@ -1231,7 +1231,7 @@ func (e *userspaceEngine) Ping(ip netip.Addr, pingType tailcfg.PingType, cb func
 	e.logf("ping(%v): sending %v ping to %v %v ...", ip, pingType, peer.Key.ShortString(), peer.ComputedName)
```
valscale, line 1231: Maybe add the ping payload size in this log message? I was running this code without the DF bit patch and had to add that in to help debug.

cmd/tailscale/cli/ping.go
```diff
@@ -53,12 +53,14 @@ relay node.
 		fs.BoolVar(&pingArgs.peerAPI, "peerapi", false, "try hitting the peer's peerapi HTTP server")
 		fs.IntVar(&pingArgs.num, "c", 10, "max number of pings to send. 0 for infinity.")
 		fs.DurationVar(&pingArgs.timeout, "timeout", 5*time.Second, "timeout before giving up on a ping")
+		fs.IntVar(&pingArgs.size, "size", 0, "send a packet with this many bytes in the payload (disco pings only). 0 for minimum size.")
```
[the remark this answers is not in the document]
valscale: And should we add support for the size argument to all the types of pings?

valscale (commented):
Looks really clean and idiomatic, thank you for fixing so many problems and porting to the new magicsock code structure!

I thought really hard about whether we can ignore the IP headers because the pinger and the ponger may have different IPv4/IPv6 sockets, and decided that my original code had a serious conceptual error: we shouldn't be trying to send back a pong of the same size at all!

I was thinking about the path between the nodes as a symmetric two way connection when it's actually two collections of UDP sockets with the current chosen endpoint changing constantly. Node A can send the ping to an IPv4 endpoint of Node B, and Node B can send the pong to an IPv6 endpoint of Node A, or to a DERP endpoint, and the MTU can be different on each path and the headers will be different too. And even when there is a two-way TCP connection on top, TCP can easily have a different MSS in each direction.

So I think we should send back a bare pong reply because we do only want to know what the MTU is in one direction to one endpoint, and not in one direction to one endpoint plus whatever path it happened to use to come back this time. This is a difference in behavior from ICMP pings which do send a reply with the same total packet length, but this makes more sense for what we are doing.

I put together a patch to send back a bare ping in valscale/pmtud2 if you want it!

cmd/tailscale/cli/ping.go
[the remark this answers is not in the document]
valscale: Agreed on all points, thanks!

wgengine/wgengine.go
```diff
@@ -152,7 +152,7 @@ type Engine interface {
 
 	// Ping is a request to start a ping with the peer handling the given IP and
 	// then call cb with its ping latency & method.
-	Ping(ip netip.Addr, pingType tailcfg.PingType, cb func(*ipnstate.PingResult))
+	Ping(ip netip.Addr, pingType tailcfg.PingType, size int, cb func(*ipnstate.PingResult))
```
bradfitz, line 155: size docs here?

client/tailscale/localclient.go
```diff
@@ -807,11 +807,16 @@ func (lc *LocalClient) ExpandSNIName(ctx context.Context, name string) (fqdn str
...
 // for its response.
-func (lc *LocalClient) Ping(ctx context.Context, ip netip.Addr, pingtype tailcfg.PingType) (*ipnstate.PingResult, error) {
+//
+// For disco pings, the size argument specifies the length of the packet's payload, that
+// is, including the disco headers and message, but not including the IP and UDP headers.
+// If size is smaller than the minimum message size it's ignored.
+func (lc *LocalClient) Ping(ctx context.Context, ip netip.Addr, pingtype tailcfg.PingType, size int) (*ipnstate.PingResult, error) {
```
bradfitz, line 816: we kinda sorta try to keep API stability in this package (one of the few in oss, as it's part of tsnet)

But OTOH this is weird enough that I don't think many people use this. You could GitHub/Google code search and see I suppose.

We might want a new method like we did for `StatusWithoutPeers`/`Status` and `BugReport`/`BugReportWithOpts`, etc.

valscale: Yeah I wasn't sure how stable the API was myself. I think this is relatively easy to do, we can just make the old Ping() a wrapper around the new Ping()? Salman, what do you think?

valscale (approved):
Looks great!

wgengine/wgengine.go
```diff
@@ -150,9 +150,9 @@ type Engine interface {
 	// status builder.
 	UpdateStatus(*ipnstate.StatusBuilder)
 
-	// Ping is a request to start a ping with the peer handling the given IP and
-	// then call cb with its ping latency & method.
-	Ping(ip netip.Addr, pingType tailcfg.PingType, cb func(*ipnstate.PingResult))
+	// Ping is a request to start a ping of the given message size to the peer
+	// handling the given IP, then call cb with its ping latency & method.
+	Ping(ip netip.Addr, pingType tailcfg.PingType, size int, cb func(*ipnstate.PingResult))
```
bradfitz, line 155: // If size is zero too small, it is ignored. See tailscale.PingOpts for details.

bradfitz (approved):

PR #8740, opened 2023-07-28
all: update exp/slices and fix call sites

danderson:
slices.SortFunc suffered a late-in-cycle API breakage.

Updates #cleanup

---

I had to import cmp.Compare into cmpx, since we don't yet want to depend on Go 1.21.


net/dns/manager.go
```diff
@@ -139,14 +140,14 @@ func compileHostEntries(cfg Config) (hosts []*HostEntry) {
...
-	slices.SortFunc(hosts, func(a, b *HostEntry) bool {
+	slices.SortFunc(hosts, func(a, b *HostEntry) int {
 		if len(a.Hosts) == 0 {
-			return false
+			return -1
 		}
 		if len(b.Hosts) == 0 {
-			return true
+			return 1
 		}
```
bradfitz, line 149: what if they're both len 0?

net/dnsfallback/dnsfallback.go
```diff
@@ -76,11 +76,27 @@ func MakeLookupFunc(logf logger.Logf, netMon *netmon.Monitor) func(ctx context.C
...
 			// Wait for a response from the main function
 			oldAddrs := <-addrsCh
-			slices.SortFunc(oldAddrs, func(a, b netip.Addr) bool { return a.Less(b) })
+			slices.SortFunc(oldAddrs, func(a, b netip.Addr) int {
+				if a.Less(b) {
+					return -1
+				} else if b.Less(a) {
+					return 1
+				} else {
+					return 0
```
bradfitz, line 97: amusingly we used to have Compare and as part of putting it in std, they asked us to remove it: https://pkg.go.dev/inet.af/netaddr#IP.Compare

Now we have evidence we need it back :)

bradfitz: Should we add AddrCompare to https://pkg.go.dev/go4.org/netipx ?

net/dns/manager.go
danderson: Added a case for that.

net/dnsfallback/dnsfallback.go
bradfitz: https://github.com/go4org/netipx/pull/14

cmd/tailscale/cli/exitnode.go
```diff
@@ -227,19 +228,21 @@ func filterFormatAndSortExitNodes(peers []*ipnstate.PeerStatus, filterBy string)
 // sortPeersByPriority sorts a slice of PeerStatus
 // by location.Priority, in order of highest priority.
 func sortPeersByPriority(peers []*ipnstate.PeerStatus) {
-	slices.SortFunc(peers, func(a, b *ipnstate.PeerStatus) bool { return a.Location.Priority > b.Location.Priority })
+	slices.SortFunc(peers, func(a, b *ipnstate.PeerStatus) int {
+		return cmpx.Compare(b.Location.Priority, a.Location.Priority)
+	})
```
bradfitz, line 233: unrelated but this should probably be SortStableFunc if there are a bunch of priorities that are the same and the input was already in some reasonable order (like sorted by name)

but can do later

bradfitz (approved):

cmd/tailscale/cli/exitnode.go
danderson: Switched to SortStableFunc.

PR #8743, opened 2023-07-28
Noncombatant/large int string

noncombatant:


bradfitz:
(missed the Comment button by a few pixels)

ipn/ipnstate/ipnstate.go
```diff
@@ -197,7 +197,7 @@ type PeerStatus struct {
 	// It has the form "host.<MagicDNSSuffix>."
 	DNSName string
 	OS      string // HostInfo.OS
-	UserID  tailcfg.UserID
+	UserID  tailcfg.UserID `json:",string"`
```
maisem, line 200: This may break clients like iOS and macOS

bradfitz: Yeah. We'll need to do this gradually with both sides agreeing.

noncombatant: Yes, I talked to Will about this, and he had already written the exact same patch. (https://github.com/tailscale/tailscale/commit/352f38feccc4b741c83c91f3382e6a494bb843d9) He also pointed out to me the "both sides need to agree" problem. (And, any other consumer of this data, not necessarily authored by us.)

PR #8745, opened 2023-07-29
wgengine/magicsock: remove dead code (i think)

saljam:
The nonce value is not read by anything, and `di.sharedKey.Seal()` a few lines below generates its own (and also panics on `crand.Read()` errors).

I really really hope this isn't somehow magically load-bearing. 😅

### DIFF
--- a/wgengine/magicsock/magicsock.go
+++ b/wgengine/magicsock/magicsock.go
@@ -8,7 +8,6 @@ package magicsock
 import (
 	"bufio"
 	"context"
-	crand "crypto/rand"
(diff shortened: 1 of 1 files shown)


bradfitz (approved):

danderson (approved):
Ah yeah, this dates back to before we had the types/key package, and this code was doing all the cryptography itself. Now, `di.sharedKey.Seal()` handles nonce generation as part of constructing the ciphertext, but I forgot to remove the nonce generation here. Thanks!

PR #8747, opened 2023-07-30
tailcfg: add helper to unmarshal PeerCap values

maisem:
Updates #4217

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -1223,6 +1223,25 @@ const (
 // the application via the WhoIs API.
 type PeerCapMap map[PeerCapability][]json.RawMessage
 
+// UnmarshalCapJSON unmarshals each JSON value in cm[cap] as T.
+// If cap does not exist in cm, it returns (nil, nil).
+// It returns an error if the values cannot be unmarshaled into the provided type.
+func UnmarshalCapJSON[T any](cm PeerCapMap, cap PeerCapability) ([]T, error) {
+	vals, ok := cm[cap]
+	if !ok {
+		return nil, nil
+	}
+	out := make([]T, 0, len(vals))
+	for _, v := range vals {
+		var t T
+		if err := json.Unmarshal(v, &t); err != nil {
+			return nil, err
+		}
+		out = append(out, t)
+	}
+	return out, nil
+}
+
 // HasCapability reports whether c has the capability cap.
 // This is used to test for the existence of a capability, especially
 // when the capability has no values.


danderson (approved):

tailcfg/tailcfg.go
```diff
@@ -1223,6 +1223,26 @@ const (
 // the application via the WhoIs API.
 type PeerCapMap map[PeerCapability][]json.RawMessage
 
+// GetCap returns the values of the given capability JSON unmarshaled into the
+// provided type. It returns nil, nil if the capability is not present. It
+// returns an error if the capability is present but the value is not of the
+// provided type.
+func GetCap[T any](cm PeerCapMap, cap PeerCapability) ([]*T, error) {
```
bradfitz, line 1230: Instead of GetCap how about UnmarshalCapJSON? Gets JSON in there and unmarshal so you know the error is related to that. Other Get returning no error on miss is kinda weird?

maisem: `UnmarshalCapJSON` sgtm

bradfitz: Does it need to be in tailcfg? This tries to be a leaf data type package without much code. This is fine but worried about additions later with new deps

maisem: I wanted to keep it closer to `PeerCapMap`, I agree with concern about this turning into a data package though `encoding/json` is not new so I thought it was ok.

maisem: Done, PTAL at docs

tailcfg/tailcfg.go
```diff
@@ -1223,6 +1223,26 @@ const (
 // the application via the WhoIs API.
 type PeerCapMap map[PeerCapability][]json.RawMessage
 
+// UnmarshalCapJSON unmarshals the cap values into the provided type.
+// It returns nil, nil if the capability is not present.
+// It returns an error if the values cannot be unmarshaled into the provided
+// type.
+func UnmarshalCapJSON[T any](cm PeerCapMap, cap PeerCapability) ([]*T, error) {
```
bradfitz, line 1230: this is to assume structs (JSON objects) too much. didn't we design the ACLs such that the args can be JSON objects, strings, numbers, or anything except lists? for strings, a slice of pointers to strings would be weird.

Return `[]T` instead and let caller choose to pass `*T` if they want that?

tailcfg/tailcfg.go
```diff
@@ -1223,6 +1223,26 @@ const (
 // the application via the WhoIs API.
 type PeerCapMap map[PeerCapability][]json.RawMessage
 
+// UnmarshalCapJSON unmarshals the cap values into the provided type.
```
bradfitz, line 1226:
```
// UnmarshalCapJSON unmarshals each JSON value in cm[cap].
// If cap does not exist in cm, it returns (nil, nil).
// It returns an error if the values cannot be unmarshaled into the provided type.
```

?

tailcfg/tailcfg.go
```diff
@@ -1223,6 +1223,26 @@ const (
...
+// It returns an error if the values cannot be unmarshaled into the provided
+// type.
+func UnmarshalCapJSON[T any](cm PeerCapMap, cap PeerCapability) ([]*T, error) {
+	vals, ok := cm[cap]
+	if !ok {
+		return nil, nil
+	}
+	out := make([]*T, 0, len(vals))
+	for _, v := range vals {
+		t := new(T)
```
bradfitz, line 1237:
```
var t T
json.Unmarshal(v, &T)
```
?

e.g. https://go.dev/play/p/LP0tZ1qdzL7

If they pass in a `*T` pointer (struct type) it'll work, but so will `string`.

tailcfg/tailcfg.go
maisem: We restricted a single grant to not be a JSON array, but each rule can have multiple grants and there are multiple rules. So at this point we were going to get a JSON array. Our reasoning to disallow JSON arrays in the ACL was specifically to avoid the case of `[[..], [.], [...]]` as thats a bit confusing.

return `[]T` sgtm, done.

bradfitz (approved):

PR #8752, opened 2023-07-31
derp/derphttp: add optional Client.BaseContext hook

bradfitz:
Like net/http.Server.BaseContext, this lets callers specify a base context for dials.

Updates tailscale/corp#12702

### DIFF
--- a/derp/derphttp/derphttp_client.go
+++ b/derp/derphttp/derphttp_client.go
@@ -56,6 +56,11 @@ type Client struct {
 	MeshKey   string             // optional; for trusted clients
 	IsProber  bool               // optional; for probers to optional declare themselves as such
 
+	// BaseContext, if non-nil, returns the base context to use for dialing a
+	// new derp server. If nil, context.Background is used.
+	// In either case, additional timeouts may be added to the base context.
+	BaseContext func() context.Context
+
 	privateKey key.NodePrivate
 	logf       logger.Logf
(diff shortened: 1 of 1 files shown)


derp/derphttp/derphttp_client.go
```diff
@@ -56,6 +56,11 @@ type Client struct {
 	MeshKey   string             // optional; for trusted clients
 	IsProber  bool               // optional; for probers to optional declare themselves as such
 
+	// BaseContext, if non-nil, returns the base context to use for dialing a
```
creachadair, line 59: Non-blocking idea, maybe out of scope here.

One thing I dislike about the http.Server option is that it doesn't let you get hold of the CancelFunc, if the context has one.

What do you think about maybe:

```go
BaseContext func() (context.Context, context.CancelFunc)
```

and then have newContext just return `func() {}` along with background?

creachadair (approved):

derp/derphttp/derphttp_client.go
bradfitz: Why does derphttp want the CancelFunc?

The goal of this is entirely context values, not deadlines/timeouts.

Also: it's sounding like Go 1.22 will have no cost for un-canceled timers and they'll just be GC'ed away whenever possible.

creachadair: Ah, I was thinking of default dial timeouts. But that's cool about the runtime handling it better soon!

PR #8758, opened 2023-07-31
client/tailscale/apitype: document never-nil property of WhoIsResponse

danderson:
Every time I use WhoIsResponse I end up writing mildly irritating nil-checking for both Node and UserProfile, but it turns out our code guarantees that both are non-nil in successful whois responses.

Updates #cleanup

### DIFF
--- a/client/tailscale/apitype/apitype.go
+++ b/client/tailscale/apitype/apitype.go
@@ -10,6 +10,7 @@ import "tailscale.com/tailcfg"
 const LocalAPIHost = "local-tailscaled.sock"
 
 // WhoIsResponse is the JSON type returned by tailscaled debug server's /whois?ip=$IP handler.
+// In successful whois responses, Node and UserProfile are never nil.
 type WhoIsResponse struct {
 	Node        *tailcfg.Node
 	UserProfile *tailcfg.UserProfile
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -427,8 +427,8 @@ func (h *Handler) serveWhoIs(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	res := &apitype.WhoIsResponse{
-		Node:        n,
-		UserProfile: &u,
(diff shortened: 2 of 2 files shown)


danderson:
wasm CI failure is an npm outage, it seems.

bradfitz (approved):

ipn/localapi/localapi.go
```diff
@@ -426,6 +426,10 @@ func (h *Handler) serveWhoIs(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "no match for IP:port", 404)
 		return
 	}
+	// Note, the API contract specifies that on successful response,
+	// Node and UserProfile are both non-nil. Dear future reader, if
+	// you're planning to change that, please consider this to be an
+	// API break.
 	res := &apitype.WhoIsResponse{
 		Node:        n,
 		UserProfile: &u,
```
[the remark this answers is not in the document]
danderson: done

client/tailscale/apitype/apitype.go
```diff
@@ -10,6 +10,7 @@ import "tailscale.com/tailcfg"
 const LocalAPIHost = "local-tailscaled.sock"
 
 // WhoIsResponse is the JSON type returned by tailscaled debug server's /whois?ip=$IP handler.
+// In successful whois responses, Node and UserProfile are never nil.
 type WhoIsResponse struct {
 	Node        *tailcfg.Node
```
maisem, line 15: While here might as well add docs about which Node and which UserProfile this refers to. And can then say always non-nil in their respective doc strings

maisem (approved):

client/tailscale/apitype/apitype.go
danderson: I don't know how to describe those fields, in particular I don't know what UserProfile represents for tagged devices. Sounds like you do though, so please send a followup?

maisem: Will do

PR #8759, opened 2023-07-31
cmd/dist,release/dist: sign release tarballs with an ECDSA key

awly:
Pass an optional PEM-encoded ECDSA key to `cmd/dist` to sign all built tarballs. The signature is stored next to the tarball with a `.sig` extension.

Tested this with an `openssl`-generated key pair and verified the resulting signature.

### DIFF
--- a/release/dist/cli/cli.go
+++ b/release/dist/cli/cli.go
@@ -6,6 +6,9 @@ package cli
 
 import (
 	"context"
+	"crypto"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"flag"
 	"fmt"
@@ -23,7 +26,7 @@ import (
 // getTargets is a function that gets run in the Exec function of commands that
 // need to know the target list. Its execution is deferred in this way to allow
 // customization of command FlagSets with flags that influence the target list.
-func CLI(getTargets func() ([]dist.Target, error)) *ffcli.Command {
+func CLI(getTargets func(tgzSigner crypto.Signer) ([]dist.Target, error)) *ffcli.Command {
 	return &ffcli.Command{
 		Name:       "dist",
 		ShortUsage: "dist [flags] <command> [command flags]",
@@ -33,7 +36,7 @@ func CLI(getTargets func() ([]dist.Target, error)) *ffcli.Command {
 			{
 				Name: "list",
 				Exec: func(ctx context.Context, args []string) error {
-					targets, err := getTargets()
+					targets, err := getTargets(nil)
 					if err != nil {
 						return err
 					}
@@ -49,7 +52,11 @@ func CLI(getTargets func() ([]dist.Target, error)) *ffcli.Command {
 			{
 				Name: "build",
 				Exec: func(ctx context.Context, args []string) error {
-					targets, err := getTargets()
+					tgzSigner, err := parseSigningKey(buildArgs.tgzSigningKey)
+					if err != nil {
+						return err
+					}
+					targets, err := getTargets(tgzSigner)
 					if err != nil {
 						return err
 					}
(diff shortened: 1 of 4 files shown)


release/dist/cli/cli.go
```diff
@@ -61,6 +68,7 @@ func CLI(getTargets func() ([]dist.Target, error)) *ffcli.Command {
 					fs := flag.NewFlagSet("build", flag.ExitOnError)
 					fs.StringVar(&buildArgs.manifest, "manifest", "", "manifest file to write")
 					fs.BoolVar(&buildArgs.verbose, "verbose", false, "verbose logging")
+					fs.StringVar(&buildArgs.tgzSigningKey, "tgz-signing-key", "", "private signing key for release tarballs")
```
noncombatant, line 71: Documentation nit: Maybe "path to private signing key..." ?

release/dist/cli/cli.go
```diff
@@ -142,3 +151,18 @@ func runBuild(ctx context.Context, filters []string, targets []dist.Target) erro
...
+
+func parseSigningKey(path string) (crypto.Signer, error) {
+	if path == "" {
+		return nil, nil
+	}
+	raw, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	b, _ := pem.Decode(raw)
```
noncombatant, line 163: In the interest of eye-watering tightness, maybe check that the `[]rest` return value is 0 bytes long? If there was extra stuff at the end of the PEM file, it's possible something weird or bad is happening.

release/dist/unixpkgs/pkgs.go
```diff
@@ -65,7 +69,9 @@ func (t *tgzTarget) Build(b *dist.Build) ([]string, error) {
 		return nil, err
 	}
 	defer f.Close()
-	gw := gzip.NewWriter(f)
+	h := sha512.New()
+	hw := io.MultiWriter(f, h)
+	gw := gzip.NewWriter(hw)
```
noncombatant, line 74: As constructed, we'll be hashing the bytes after tar'ing and gzip'ing, right? It looks like that to me, and that's what we want, but it took me a minute to figure that out. It might be good for future readers to put a comment here saying that we intentionally hash the result of gzip (and not, say, the file contents that go into tar or gzip).

release/dist/unixpkgs/pkgs.go
```diff
@@ -146,7 +152,21 @@ func (t *tgzTarget) Build(b *dist.Build) ([]string, error) {
 		return nil, err
 	}
 
-	return []string{filename}, nil
+	files := []string{filename}
+
+	if t.signer != nil {
+		sig, err := t.signer.Sign(rand.Reader, h.Sum(nil), crypto.SHA512)
```
noncombatant, line 158: I'm not 100% sure, but it might be the case that `Close()` causes some last bits of buffered I/O to be written out. Since we `defer` `Close`, we actually might (non-deterministically) sign an incomplete file? Does that sound possible? It might be best to explicitly, deterministically `Close`.

(I'm guessing on the basis of libc `fclose`, which is obviously not the same as `os.File`, but the basic concept might cross languages.)

release/dist/unixpkgs/pkgs.go
awly: Good point, left a comment :+1:

release/dist/unixpkgs/pkgs.go
awly: All the writers are explicitly closed above: https://github.com/tailscale/tailscale/pull/8759/files#diff-c313c5cc009cf463fc06aae4a619fbd78ec928df4bd80e7917145610ecee694eL139-L147
The deferred `Close` calls are just fail-safe noops in this case.

noncombatant (approved):

PR #8761, opened 2023-07-31
client/tailscale/apitype,tailcfg: document WhoIsResponse fields

maisem:
Updates #cleanup

### DIFF
--- a/client/tailscale/apitype/apitype.go
+++ b/client/tailscale/apitype/apitype.go
@@ -10,9 +10,16 @@ import "tailscale.com/tailcfg"
 const LocalAPIHost = "local-tailscaled.sock"
 
 // WhoIsResponse is the JSON type returned by tailscaled debug server's /whois?ip=$IP handler.
-// In successful whois responses, Node and UserProfile are never nil.
 type WhoIsResponse struct {
-	Node        *tailcfg.Node
+	// Node is the node that the IP address belongs to.
+	//
+	// It is always non-nil in successful responses.
+	Node *tailcfg.Node
+
+	// UserProfile is the profile of the Node.User if the node is not tagged.
+	// For tagged nodes, the profile should be ignored.
+	//
+	// It is always non-nil in successful responses.
 	UserProfile *tailcfg.UserProfile
 
 	// CapMap is a map of capabilities to their values.
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -221,9 +221,12 @@ type Node struct {
 	// e.g. "host.tail-scale.ts.net."
 	Name string
 
-	// User is the user who created the node. If ACL tags are in
-	// use for the node then it doesn't reflect the ACL identity
-	// that the node is running as.
+	// User represents the owner of an untagged node. This field should be
+	// ignored for tagged nodes, as the value is undefined and may change in the
+	// future.
+	//
+	// To support old clients, the value is used to group all tagged nodes
+	// together for UI purposes and to circumvent isSelf checks for the PeerAPI.
 	User UserID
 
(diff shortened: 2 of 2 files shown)


danderson (approved):

client/tailscale/apitype/apitype.go
```diff
@@ -10,9 +10,16 @@ import "tailscale.com/tailcfg"
...
 // WhoIsResponse is the JSON type returned by tailscaled debug server's /whois?ip=$IP handler.
-// In successful whois responses, Node and UserProfile are never nil.
 type WhoIsResponse struct {
-	Node        *tailcfg.Node
+	// Node is the node that the IP address belongs to.
+	//
+	// It is always non-nil in successful responses.
+	Node *tailcfg.Node
+
+	// UserProfile is the profile of the Node.User if the node is not tagged.
```
creachadair, line 19: Just for my understanding: If the node _is_ tagged, it sounds like this field is still set (per the "always non-nil" below), but then what _is_ the value of this field in that case?

maisem: It basically says "tagged-devices"

client/tailscale/apitype/apitype.go
```diff
@@ -10,9 +10,16 @@ import "tailscale.com/tailcfg"
...
+	// For tagged nodes, the profile should be ignored.
+	//
+	// It is always non-nil in successful responses.
```
bradfitz, line 22: these last two sentences invite some questions you could get in front of.

```
// UserProfile is the profile of the Node.User.
//
// While this field is always non-nil, it's only meant to be used for human peers.
// For tagged users it is populated with fake data and should be ignored.
```

client/tailscale/apitype/apitype.go
```diff
@@ -10,9 +10,16 @@ import "tailscale.com/tailcfg"
 const LocalAPIHost = "local-tailscaled.sock"
 
 // WhoIsResponse is the JSON type returned by tailscaled debug server's /whois?ip=$IP handler.
-// In successful whois responses, Node and UserProfile are never nil.
 type WhoIsResponse struct {
-	Node        *tailcfg.Node
+	// Node is the node that the IP address belongs to.
+	//
+	// It is always non-nil in successful responses.
```
bradfitz, line 16: in a failure response we return a nil `*WhoIsResponse` anyway, so you can just say "It is always non-nil." if you wanted. Also true as written, though.

tailcfg/tailcfg.go
```diff
@@ -221,9 +221,12 @@ type Node struct {
 	// e.g. "host.tail-scale.ts.net."
 	Name string
 
-	// User is the user who created the node. If ACL tags are in
-	// use for the node then it doesn't reflect the ACL identity
-	// that the node is running as.
+	// User represents the owner of an untagged node. This field should be
```
bradfitz, line 224: "untagged (human) node" ?

tailcfg/tailcfg.go
```diff
@@ -221,9 +221,12 @@ type Node struct {
...
+	// ignored for tagged nodes, as the value is undefined and may change in the
+	// future.
+	//
+	// To support old clients, the value is used to group all tagged nodes
+	// together for UI purposes and to circumvent isSelf checks for the PeerAPI.
```
bradfitz, line 229: To support old clients, the same fake UserID is used for all tagged nodes.
That made them them grouped together in the UI. Also, because the fake UserID for
tagged nodes doesn't correspond to any real human's UserID, ...

wait, the peerapi check is:

```
		isSelf:     nm.SelfNode.User == peerNode.User,
```

Which means this Node.User is correct (a human) for the self node but a fake non-user UserID for all the peers' Nodes.  So the definition of this field changes depending on not only the taggedness but also its parent container?

Issue #8111, opened 2023-05-10
tailscale-operator not working on AKS


tiesmaster:
### What is the issue?

I want to expose my AKS kubernetes cluster via Tailscale, and tried to do that via the [Tailscale proxy](https://tailscale.com/kb/1185/kubernetes/#sample-proxy), however, that doesn't seems to proxy the traffic to the destination IP. As [suggested here](https://github.com/tailscale/tailscale/issues/7322#issuecomment-1435826933), I tried out the [Tailscale Operator](https://tailscale.com/kb/1236/kubernetes-operator/), hoping that that might work, however, I'm running into the same issue here.

To troubleshoot the issue, I've ramped up a secondary cluster, using microk8s, and I've got the Tailscale Operator working there, as you can see in the logs of the proxy pod:

**Proxy pod on AKS**
```
$ kubectl logs -n tailscale pod/ts-sample-workload-one-bjvhs-0
...
2023/05/10 08:27:55 Accept: TCP{100.115.48.77:64882 > 100.79.209.22:80} 52 tcp ok
2023/05/10 08:27:56 Accept: TCP{100.115.48.77:64882 > 100.79.209.22:80} 52 tcp ok
```

**Proxy pod on microk8s**
```
$ kubectl logs -n tailscale pod/ts-sample-workload-one-7nst7-0
...
2023/05/10 08:27:40 Accept: TCP{100.115.48.77:64880 > 100.67.186.56:80} 52 tcp ok
2023/05/10 08:27:40 Accept: TCP{100.67.186.56:80 > 100.115.48.77:64880} 52 ok out
2023/05/10 08:27:40 Accept: TCP{100.115.48.77:64880 > 100.67.186.56:80} 40 tcp non-syn
```

This is what I get when I `curl` the Tailscale IP of the sample workload (which is [this Deployment, and Service](https://gist.github.com/tiesmaster/d7b397f19015514451fd0cd58b37fb06) that I've used for that). When I hit the AKS cluster, nothing happens, when I hit the microk8s cluster, I get the [expected output](https://gist.github.com/tiesmaster/d7b397f19015514451fd0cd58b37fb06?permalink_comment_id=4563570#gistcomment-4563570).

[...]

Issue #8733, opened 2023-07-27
Tailscale-operator not working on k3s cluster


clarkezone:
`###` What is the issue?

I’m attempting a simple helloworld level scenario for tailscale-operator on k3s on ubnuntu 22.04 and the scenario isn't working correctly.

### Steps to reproduce

1. backup kubeconfig:
```bash
mv ~/.kube/config ~/.kube/backup
```
2. install k3s and confirm running:
```bash
curl -sfL https://get.k3s.io | sh -
sudo mv /etc/rancher/k3s/k3s.yaml ~/.kube/config
sudo chmod 777 ~/.kube/config
➜  tailscaleoperator git:(c4updates) ✗ k get nodes
NAME                   STATUS   ROLES                  AGE   VERSION
clarkezonedevbox5-tr   Ready    control-plane,master   94s   v1.27.3+k3s1
```

3. install tailscale operator from instructions here: https://tailscale.com/kb/1236/kubernetes-operator/ 
```bash


➜  tailscaleoperator git:(c4updates) ✗ k apply -f operator.yaml
namespace/tailscale created
serviceaccount/proxies created
role.rbac.authorization.k8s.io/proxies created
rolebinding.rbac.authorization.k8s.io/proxies created
serviceaccount/operator created
clusterrole.rbac.authorization.k8s.io/tailscale-operator created
clusterrolebinding.rbac.authorization.k8s.io/tailscale-operator created
role.rbac.authorization.k8s.io/operator created
rolebinding.rbac.authorization.k8s.io/operator created
secret/operator-oauth created
deployment.apps/operator created
➜  tailscaleoperator git:(c4updates) ✗ k get pods -n tailscale
NAME                       READY   STATUS    RESTARTS   AGE
operator-74cdfb6f5-gj7dw   1/1     Running   0          8s

# verify operator showing in tailnet
➜  tailscaleoperator git:(c4updates) ✗ tailscale status | grep operator
100.82.108.42   tailscale-operator   tagged-devices linux   -
```

4. install test workload:
```bash
k apply -f https://gist.github.com/clarkezone/b22a5851f2e4229f5fd29f1115ddee32/raw/766708eee8f614d846dc12afe4dfaa819a678ee9/tailscaletest.yaml

➜  tailscaleoperatortest git:(master) k get pods -n tailscaletest
NAME                              READY   STATUS    RESTARTS   AGE
nginx-tailscale-7bbbb87bf-2f7lc   1/1     Running   0          30s
[...]

PR #8762, opened 2023-08-01
wgengine/router: add auto selection heuristic for iptables/nftables

KevinLiang10:
### Summary:

This change implements `chooseFireWallMode` that decides which one of `iptables`/`nftables` will be used.

### Issue related

Updates: #391 
Fix: #8111 #8733

### Changes made

1. Added `chooseFireWallMode` that automatically decides which one of `iptables` or`nftables` should be used.
2. Replaced the `TS_DEBUG_USE_NETLINK_NFTABLES` envknob with `TS_DEBUG_FIREWALL_MODE` that should be set to either 'iptables' or 'nftables' to select firewall mode manually.
3. Reimplemented `DetectIptables` that detects `iptables` binary and returns number of rules in `iptables` in current network namespace.

### Thought Process
1. It is a very edge case when iptables binary is not available, nft is also not available and iptables is in use. So we are not supporting this case for now.
2. It is a very edge case when nftables is only available throught `nft` binary but not netlinkAPI, so we will skip this case for now also. This is usually due to system configuration.
3. We always prioritize nftables ahead of iptables due to it's excellence in performance, and less likely to fail in containerized environment.
4. When In a container, if there was other app in the same container using `iptables`, we use iptables since if iptables wouldn't work the container would fail anyways. Otherwise we tend to use `nftables` since netlinkAPI is less likely to run into incompatibility issues.

### Expected Result

1. If either `'iptables'` or `'nftables'` is set for `TS_DEBUG_FIREWALL_MODE` use the tool respectively.
5. If `nftables` is in use already use nftables.
6. if `iptables` is already in use user iptables, notice that since iptables-nft would write to nftables subsystem, iptables here refer to iptables-legacy.
7. Otherwise if `nftables` is available throught `netlink API`, use nftables. This case includes both ipt, nft are in use, neither is used or iptables-nft was used.
8. If only `iptables` binary is available, use iptables.
9. If neither iptables or nftables is available, return error.

### How to test

On linux environment do the following:

1. Bring up tailscaled by running `sudo TS_DEBUG_FIREWALL_MODE=iptables ./tool/go run ./cmd/tailscaled`
2. Start tailscale by running `sudo ./tool/go run ./cmd/tailscale up`.
3. Run `sudo iptables -L` to view the rules added. In log of tailscaled, there should be a line `router: router: envknob TS_DEBUG_NETFILTER_MODE=iptables set`.
4. Turn everything down.
5. Bring up tailscaled by running `sudo TS_DEBUG_FIREWALL_MODE=nftables ./tool/go run ./cmd/tailscaled`
6. Start tailscale by running `sudo ./tool/go run ./cmd/tailscale up`.
7. Run `sudo nft list ruleset` to view the rules added. In the output log of tailscaled, it should say `router: envknob TS_DEBUG_NETFILTER_MODE=nftables set`.
8. Turn everything down.
9.  Buring up tailscaled by running `sudo ./tool/go run ./cmd/tailscaled`
10. Start tailscale
11. You should see the count of iptables rules and nftables rule currently in system from the log in the form 
```
router: router: iptables rule count: {number}
router: router: nftables rule count: {number}
```
if iptables is associated with iptables-nft, or nftables is used, it's likely that bother number are non zero depending on system implementation.
12. Your tailscale should be working, with which ever mode is selected.

### DIFF
--- a/util/linuxfw/iptables.go
+++ b/util/linuxfw/iptables.go
@@ -7,7 +7,13 @@
 package linuxfw
 
 import (
+	"fmt"
+	"os/exec"
+	"strings"
+	"unicode"
+
 	"tailscale.com/types/logger"
+	"tailscale.com/util/multierr"
 )
 
 // DebugNetfilter prints debug information about iptables rules to the
@@ -21,9 +27,44 @@ func DebugIptables(logf logger.Logf) error {
 // system, ignoring the default "ACCEPT" rule present in the standard iptables
 // chains.
 //
-// It only returns an error when the kernel returns an error (i.e. when a
-// syscall fails); when there are no iptables rules, it is valid for this
-// function to return 0, nil.
+// It only returns an error when there is no iptables binary, or when iptables -S
+// fails. In all other cases, it returns the number of non-default rules.
 func DetectIptables() (int, error) {
-	panic("unused")
+	// run "iptables -S" to get the list of rules using iptables
+	// exec.Command returns an error if the binary is not found
+	cmd := exec.Command("iptables", "-S")
+	output, err := cmd.Output()
+	ip6cmd := exec.Command("ip6tables", "-S")
+	ip6output, ip6err := ip6cmd.Output()
+	var allLines []string
+	outputStr := string(output)
+	lines := strings.Split(outputStr, "\n")
+	ip6outputStr := string(ip6output)
+	ip6lines := strings.Split(ip6outputStr, "\n")
+	switch {
+	case err == nil && ip6err == nil:
+		allLines = append(lines, ip6lines...)
+	case err == nil && ip6err != nil:
+		allLines = lines
+	case err != nil && ip6err == nil:
+		allLines = ip6lines
+	default:
+		return 0, ErrorFWModeNotSupported{
+			Mode: FirewallModeIPTables,
+			Err:  fmt.Errorf("iptables command run fail: %w", multierr.New(err, ip6err)),
+		}
+	}
+
+	// count the number of non-default rules
+	count := 0
+	for _, line := range allLines {
+		trimmedLine := strings.TrimLeftFunc(line, unicode.IsSpace)
+		if line != "" && strings.HasPrefix(trimmedLine, "-A") {
+			// if the line is not empty and starts with "-A", it is a rule appended not default
+			count++
+		}
+	}
+
+	// return the count of non-default rules
+	return count, nil
 }
--- a/wgengine/router/router_linux.go
+++ b/wgengine/router/router_linux.go
@@ -54,24 +54,95 @@ type netfilterRunner interface {
 	HasIPV6NAT() bool
 }
 
+// tableDetector abstracts helpers to detect the firewall mode.
+// It is implemented for testing purposes.
+type tableDetector interface {
+	iptDetect() (int, error)
+	nftDetect() (int, error)
+}
+
+type linuxFWDetector struct{}
+
+// iptDetect returns the number of iptables rules in the current namespace.
+func (l *linuxFWDetector) iptDetect() (int, error) {
+	return linuxfw.DetectIptables()
+}
+
+// nftDetect returns the number of nftables rules in the current namespace.
+func (l *linuxFWDetector) nftDetect() (int, error) {
+	return linuxfw.DetectNetfilter()
+}
+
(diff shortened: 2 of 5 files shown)


wgengine/router/router_linux.go
```diff
@@ -54,24 +54,97 @@ type netfilterRunner interface {
...
+func chooseFireWallMode(logf logger.Logf, det tableDetector) (linuxfw.FirewallMode, error) {
+	iptAva, nftAva := true, true
+	iptRuleCount, err := det.iptDetect()
+	logf("router: iptables rule count: %d", iptRuleCount)
+	if err != nil {
+		logf("roter: %v", err)
+		iptAva = false
+	}
+	nftRuleCount, err := det.nftDetect()
+	logf("router: nftables rule count: %d", nftRuleCount)
```
andrew-d, line 87: We should consider logging both the `iptables` and `nftables` rule counts in the same log line, so that if we want to debug a system we can look at a single log line instead of two different ones that might be separated by some logs from other goroutines.

andrew-d (commented):
Looking good! I did a preliminary review, and I'd appreciate if @raggi and/or @catzkorn could do one as well since they're more familiar with this than I am.

util/linuxfw/iptables.go
```diff
@@ -17,13 +22,48 @@ func DebugIptables(logf logger.Logf) error {
...
 // chains.
 //
-// It only returns an error when the kernel returns an error (i.e. when a
-// syscall fails); when there are no iptables rules, it is valid for this
-// function to return 0, nil.
+// It only returns an error when there is no iptables binary, or when iptables -S
+// fails. In all other cases, it returns the number of non-default rules.
 func DetectIptables() (int, error) {
-	panic("unused")
+	if !checkBinaryExists("iptables") {
```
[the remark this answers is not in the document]
KevinLiang10: Actually it seems like user were able to use ip6tables if it exists, so I should probably add that in this pr.

util/linuxfw/iptables.go
```diff
@@ -17,13 +23,61 @@ func DebugIptables(logf logger.Logf) error {
...
 //
-// It only returns an error when the kernel returns an error (i.e. when a
-// syscall fails); when there are no iptables rules, it is valid for this
-// function to return 0, nil.
+// It only returns an error when there is no iptables binary, or when iptables -S
+// fails. In all other cases, it returns the number of non-default rules.
 func DetectIptables() (int, error) {
-	panic("unused")
+	// TODO(kevinliang10): Check ip6tables in case someone is running IPv6 only.
+	if !checkBinaryExists("iptables") && !checkBinaryExists("ip6tables") {
```
raggi, line 40: The exec.Command invocations below will implicitly be doing `exec.LookPath`, do we need both?

raggi (approved):
Looks good overall, but please adjust the logic so that we can still detect and function correctly on systems that only have iptables and are missing ip6tables, as we are able to function on those configurations today.

util/linuxfw/iptables.go
KevinLiang10: I guess we don't need both. I think it's easier to backtrace the error, but the error message wouldn't make too much of a change.

joesankey:
I was getting the following error while trying these changes with the Kubernetes operator to expose a service on a tailnet:
```
│ modprobe: can't change directory to '/lib/modules': No such file or directory                                                                                               │
│ modprobe: can't change directory to '/lib/modules': No such file or directory                                                                                               │
│ modprobe: can't change directory to '/lib/modules': No such file or directory                                                                                               │
│ iptables v1.8.8 (legacy): can't initialize iptables table `nat': Table does not exist (do you need to insmod?)                                                              │
│ Perhaps iptables or your kernel needs to be upgraded.                                                                                                                       │
│ boot: 2023/08/11 01:42:37 installing proxy rules: executing iptables failed: exit status 3
```

Looks like `containerboot` needs to be updated to use iptables/nftables autoselection this PR introduced. When in nftable mode, the nat table no longer exists for container boot to add rules to.

I added this [commit](https://github.com/tailscale/tailscale/commit/b852f9dde6f3dc96e9a862a90253c066c9f81d55#diff-700830ac79a41da42cd3a82065f0fee3003a4390eead4db1de4bf30557a06bd6L521) to my fork to fix the issue, at least for nftables.

PR #8767, opened 2023-08-02
tailcfg: add CapabilityHTTPS const

soniaappasamy:
A #cleanup

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -1940,6 +1940,7 @@ const (
 	CapabilitySSHRuleIn          = "https://tailscale.com/cap/ssh-rule-in"           // some SSH rule reach this node
 	CapabilityDataPlaneAuditLogs = "https://tailscale.com/cap/data-plane-audit-logs" // feature enabled
 	CapabilityDebug              = "https://tailscale.com/cap/debug"                 // exposes debug endpoints over the PeerAPI
+	CapabilityHTTPS              = "https"                                           // https cert provisioning enabled on tailnet
 
 	// CapabilityBindToInterfaceByRoute changes how Darwin nodes create
 	// sockets (in the net/netns package). See that package for more


tailcfg/tailcfg.go
```diff
@@ -1940,6 +1940,7 @@ const (
 	CapabilitySSHRuleIn          = "https://tailscale.com/cap/ssh-rule-in"           // some SSH rule reach this node
 	CapabilityDataPlaneAuditLogs = "https://tailscale.com/cap/data-plane-audit-logs" // feature enabled
 	CapabilityDebug              = "https://tailscale.com/cap/debug"                 // exposes debug endpoints over the PeerAPI
+	CapabilityHTTPS              = "https"                                           // https enabled on tailnet
```
bradfitz, line 1943: `// https cert provisioning enabled on tailnet` ?

bradfitz (approved):

tailcfg/tailcfg.go
creachadair: I see why this might be a logical exception to the format illustrated above, but should we use a URL format anyway? (I don't have a strong opinion, it just stood out)

soniaappasamy: URL is actually the old format, sounds like the new style is just the cap string itself. https://github.com/tailscale/corp/pull/13496#discussion_r1281119497

creachadair: Ah, TIL, thanks for the clarification!

bradfitz: So really we wanted URLs on the wire everywhere and have since screwed that up. The wrong turn we took was when we started using URLs as nodeAttrs in policy JSON and some people objected to how ugly it was. Rather than saying the policy JSON "foo" expands to `"tailscale.com/cap/foo"`, we instead started sending bare caps without URLs on the wire. Oh well.

But that's the history. They should have URLs, but bare ones are a special case now that means tailscale.com.

PR #8768, opened 2023-08-02
go.toolchain.rev: bump to ~go1.21rc4

bradfitz:
Updates tailscale/go#69


bradfitz:
(This will fail CI first because it's still building in GitHub Actions for another 10 minutes)

soniaappasamy (approved):

PR #8774, opened 2023-08-02
tailcfg,ipn/localapi,client/tailscale: add QueryFeature endpoint

soniaappasamy:
Updates tailscale/corp#10577

### DIFF
--- a/client/tailscale/localclient.go
+++ b/client/tailscale/localclient.go
@@ -1124,6 +1124,27 @@ func (lc *LocalClient) DeleteProfile(ctx context.Context, profile ipn.ProfileID)
 	return err
 }
 
+// QueryFeature makes a request for instructions on how to enable a
+// feature, such as Funnel, for the node's tailnet.
+//
+// This request itself does not directly enable the feature on behalf
+// of the node, but rather returns information that can be presented
+// to the acting user about where/how to enable the feature.
+//
+// If relevant, this includes a control URL the user can visit to
+// explicitly consent to using the feature. LocalClient.WatchIPNBus
+// can be used to block on the feature being enabled.
+//
+// 2023-08-02: Valid feature values are "serve" and "funnel".
+func (lc *LocalClient) QueryFeature(ctx context.Context, feature string) (*tailcfg.QueryFeatureResponse, error) {
+	v := url.Values{"feature": {feature}}
+	body, err := lc.send(ctx, "POST", "/localapi/v0/query-feature?"+v.Encode(), 200, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error %w: %s", err, body)
+	}
+	return decodeJSON[*tailcfg.QueryFeatureResponse](body)
+}
+
 func (lc *LocalClient) DebugDERPRegion(ctx context.Context, regionIDOrCode string) (*ipnstate.DebugDERPRegionReport, error) {
 	v := url.Values{"region": {regionIDOrCode}}
 	body, err := lc.send(ctx, "POST", "/localapi/v0/debug-derp-region?"+v.Encode(), 200, nil)
--- a/ipn/localapi/localapi.go
+++ b/ipn/localapi/localapi.go
@@ -113,6 +113,7 @@ var handler = map[string]localAPIHandler{
 	"upload-client-metrics":       (*Handler).serveUploadClientMetrics,
 	"watch-ipn-bus":               (*Handler).serveWatchIPNBus,
 	"whois":                       (*Handler).serveWhoIs,
+	"query-feature":               (*Handler).serveQueryFeature,
 }
 
 func randHex(n int) string {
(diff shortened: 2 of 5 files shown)


client/tailscale/localclient.go
```diff
@@ -1124,6 +1124,25 @@ func (lc *LocalClient) DeleteProfile(ctx context.Context, profile ipn.ProfileID)
 	return err
 }
 
+// EnableFeature makes a request for instructions on how to enable a
```
bradfitz, line 1127: Kinda a weird name in that it doesn't actually enable it.

It's more like "QueryFeatureAvailability" or something, which is kinda a mouthful. Maybe "QueryFeature" or "CheckFeature"?

tailcfg/tailcfg.go
```diff
@@ -2257,6 +2257,46 @@ type SSHRecordingAttempt struct {
...
+// to enable a feature, such as Funnel, for the node's tailnet.
+//
+// This request itself DOES NOT modify tailnet configuration to enable the
+// feature, but returns information that can be presented to the user about
+// where/how to enable the feature. See EnableFeatureResponse for response
+// structure.
+type EnableFeatureRequest struct {
+	// NodeKey is the client's current node key.
+	NodeKey key.NodePublic `json:",omitempty"`
+}
```
bradfitz, line 2270: now that I see this written out, it's kinda weird that the feature string isn't in the request but lives outside in the URL.

I don't think REST-y purity should be our goal here: this is a weirdo API over noise only that end users can't really hit.  So the server endpoint could just be `/machine/feature` or something?

tailcfg/tailcfg.go
```diff
@@ -2257,6 +2257,46 @@ type SSHRecordingAttempt struct {
...
+// EnableFeatureResponse is the response to an EnableFeatureRequest.
+type EnableFeatureResponse struct {
+	// Complete is true when the feature is already enabled.
+	Complete bool `json:",omitempty"`
+
+	// Text holds lines to display in the CLI with information
+	// about the feature and how to enable it.
+	//
+	// Lines are separated by newline characters. No trailing
+	// newline.
```
bradfitz, line 2281: last sentence isn't a full sentence. But maybe leave more wiggle worm. "The final newline may be omitted." ?

tailcfg/tailcfg.go
```diff
@@ -2257,6 +2257,46 @@ type SSHRecordingAttempt struct {
...
+	Text string `json:",omitempty"`
+
+	// URL is filled with the link for the user to visit to take
```
bradfitz, line 2284: can drop the passive voice. "URL is the link for the user to visit to..."

client/tailscale/localclient.go
soniaappasamy: Yea I was going back and forth on naming too - I like `QueryFeature`, was also considering `RequestFeature` but then that's a lot of "request request" repetition

tailcfg/tailcfg.go
soniaappasamy: Yeah - can do `/machine/feature` and move feature to the request

client/tailscale/localclient.go
soniaappasamy: Okay, updated to `QueryFeature`. I'm still not totally in love with the name, but can start here and update if I come up with a better one.

client/tailscale/localclient.go
```diff
@@ -1124,6 +1124,25 @@ func (lc *LocalClient) DeleteProfile(ctx context.Context, profile ipn.ProfileID)
 	return err
 }
 
+// QueryFeature makes a request for instructions on how to enable a
+// feature, such as Funnel, for the node's tailnet.
+//
+// This request itself DOES NOT directly enable the feature on behalf
```
bradfitz, line 1130: I think this shouting is probably no longer necessary given the new name of the method.

client/tailscale/localclient.go
```diff
@@ -1124,6 +1124,25 @@ func (lc *LocalClient) DeleteProfile(ctx context.Context, profile ipn.ProfileID)
...
+// of the node, but rather returns information that can be presented
+// to the acting user about where/how to enable the feature.
+//
+// If relevant, this includes a control URL the user can visit to
+// explicitly consent to using the feature. LocalClient.WatchIPNBus
+// can be used to block on the feature being enabled.
+func (lc *LocalClient) QueryFeature(ctx context.Context, feature string) (*tailcfg.QueryFeatureResponse, error) {
```
bradfitz, line 1137: any doc guidance on what strings this feature might be?

ipn/localapi/localapi.go
```diff
@@ -1932,6 +1933,64 @@ func (h *Handler) serveProfiles(w http.ResponseWriter, r *http.Request) {
...
+	b, err := json.Marshal(&tailcfg.QueryFeatureRequest{
+		NodeKey: nm.NodeKey,
+		Feature: feature,
+	})
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	req, err := http.NewRequest("POST", "https://unused/machine/feature/enable", bytes.NewReader(b))
```
bradfitz, line 1975: NewRequestWithContext(r.Context(), "POST" ...

ipn/localapi/localapi.go
```diff
@@ -1932,6 +1933,64 @@ func (h *Handler) serveProfiles(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	resp, err := h.b.DoNoiseRequest(req)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	defer resp.Body.Close()
+	w.WriteHeader(resp.StatusCode)
```
bradfitz, line 1987: you want to copy the Content-Type header at least here before the WriteHeader?

tailcfg/tailcfg.go
```diff
@@ -2257,6 +2257,48 @@ type SSHRecordingAttempt struct {
 	FailureMessage string
 }
 
+// QueryFeatureRequest is a request sent to "/machine/feature/enable"
```
bradfitz, line 2260: do you want "enable" here? not "query" now?

bradfitz (approved):

client/tailscale/localclient.go
soniaappasamy: added

Issue #8419, opened 2023-06-23
all: switch to Go 1.21


bradfitz:
Tracking bug for switching to Go 1.21.

First bit was: https://github.com/tailscale/go/pull/66

/cc @phirework

maisem:
Had to turn off a couple of tests in #8440, which should be turned back on when we enable this everywhere.

PR #8775, opened 2023-08-02
go.toolchain.rev: go1.21rc4 (now that VERSION file is updated upstream)

bradfitz:
Updates #8419


go.toolchain.rev
```diff
@@ -1 +1 @@
-16af3c2f4711e2bb6f662b2bc8d5cc6eb4c572cc
+d149af282305d5365d5a4fb576d9fa81247eb6da
```
bradfitz, line 1: Entire diff is the VERSION file: https://github.com/tailscale/go/compare/16af3c2f4711e2bb6f662b2bc8d5cc6eb4c572cc..d149af282305d5365d5a4fb576d9fa81247eb6da

PR #8782, opened 2023-08-03
tailscale/cli: add interactive flow for enabling Funnel

soniaappasamy:
Updates tailscale/corp#10577

### DIFF
--- a/client/tailscale/localclient.go
+++ b/client/tailscale/localclient.go
@@ -1144,18 +1144,18 @@ func (lc *LocalClient) DeleteProfile(ctx context.Context, profile ipn.ProfileID)
 	return err
 }
 
-// QueryFeature makes a request for instructions on how to enable a
-// feature, such as Funnel, for the node's tailnet.
+// QueryFeature makes a request for instructions on how to enable
+// a feature, such as Funnel, for the node's tailnet. If relevant,
+// this includes a control server URL the user can visit to enable
+// the feature.
 //
-// This request itself does not directly enable the feature on behalf
-// of the node, but rather returns information that can be presented
-// to the acting user about where/how to enable the feature.
+// If you are looking to use QueryFeature, you'll likely want to
+// use cli.enableFeatureInteractive instead, which handles the logic
+// of wraping QueryFeature and translating its response into an
+// interactive flow for the user, including using the IPN notify bus
+// to block until the feature has been enabled.
 //
-// If relevant, this includes a control URL the user can visit to
-// explicitly consent to using the feature. LocalClient.WatchIPNBus
-// can be used to block on the feature being enabled.
-//
-// 2023-08-02: Valid feature values are "serve" and "funnel".
+// 2023-08-09: Valid feature values are "serve" and "funnel".
 func (lc *LocalClient) QueryFeature(ctx context.Context, feature string) (*tailcfg.QueryFeatureResponse, error) {
 	v := url.Values{"feature": {feature}}
 	body, err := lc.send(ctx, "POST", "/localapi/v0/query-feature?"+v.Encode(), 200, nil)
--- a/cmd/tailscale/cli/funnel.go
+++ b/cmd/tailscale/cli/funnel.go
@@ -13,7 +13,10 @@ import (
 	"strings"
 
 	"github.com/peterbourgon/ff/v3/ffcli"
+	"golang.org/x/exp/slices"
 	"tailscale.com/ipn"
+	"tailscale.com/ipn/ipnstate"
+	"tailscale.com/tailcfg"
 	"tailscale.com/util/mak"
 )
 
@@ -91,9 +94,10 @@ func (e *serveEnv) runFunnel(ctx context.Context, args []string) error {
 	}
 	port := uint16(port64)
 
-	if err := ipn.CheckFunnelAccess(port, st.Self.Capabilities); err != nil {
+	if err := e.verifyFunnelEnabled(ctx, st, port); err != nil {
 		return err
 	}
+
 	dnsName := strings.TrimSuffix(st.Self.DNSName, ".")
 	hp := ipn.HostPort(dnsName + ":" + strconv.Itoa(int(port)))
 	if on == sc.AllowFunnel[hp] {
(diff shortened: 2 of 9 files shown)


willnorris (approved):
test failure looks legit, but new interactive flow lgtm

client/tailscale/localclient.go
```diff
@@ -1124,19 +1125,81 @@ func (lc *LocalClient) DeleteProfile(ctx context.Context, profile ipn.ProfileID)
...
+// If err is returned non-empty, the client failed to query the control
+// server for information about how to enable the feature.
+//
+// If ok is false and err is nil, the feature could not be enabled
+// interactively (maybe this user doesn't have permission to enable it,
+// or they're on the wrong billing plan).
+// In this case, the CLI should exit.
+//
+// 2023-08-03: Valid feature values are "serve" and "funnel".
+func (lc *LocalClient) EnableFeatureInteractive(ctx context.Context, feature string, hasRequiredCapabilities func(caps []string) bool) (ok bool, err error) {
```
bradfitz, line 1146: This seems pretty hyper-specialized for being in the tailscale client package where we kinda have an API promise. Plus that func callback is pretty specific.

This seems like it could live elsewhere, like in the CLI package even? Does anything else need to call this besides the CLI?

cmd/tailscale/cli/serve.go
```diff
@@ -129,7 +128,7 @@ type localServeClient interface {
 	Status(context.Context) (*ipnstate.Status, error)
 	GetServeConfig(context.Context) (*ipn.ServeConfig, error)
 	SetServeConfig(context.Context, *ipn.ServeConfig) error
-	QueryFeature(context.Context, string) (*tailcfg.QueryFeatureResponse, error)
+	EnableFeatureInteractive(ctx context.Context, feature string, hasRequiredAttrs func(attrs []string) bool) (ok bool, err error)
```
bradfitz, line 131: if this is the only reason you want to put the code on LocalClient, maybe you can rework the tests a bit instead?

tsnet/tsnet.go
```diff
@@ -930,6 +930,9 @@ func (s *Server) ListenFunnel(network, addr string, opts ...FunnelOption) (net.L
 	if err != nil {
 		return nil, err
 	}
+	// TODO(sonia,tailscale/corp#10577): We may want to use the interactive enable
+	// flow here instead of CheckFunnelAccess to allow the user to turn on funnel
+	// if not already on. See cli.serveEnv.verifyFunnelEnabled.
```
bradfitz, line 935: Only if we're super sure they're running it from a terminal. It's better to either continue on in degraded more or crash hard if they're running it under systemd/etc.

bradfitz (commented):
I didn't review this super carefully but seems right at first glance. I defer to other reviewers. My only preference is if we can maybe move that interactive checking code out of the LocalClient.

tsnet/tsnet.go
soniaappasamy: Yea figured tsnet is probably a different case. I'll add a little more to this comment.

cmd/tailscale/cli/serve.go
soniaappasamy: Updated

client/tailscale/localclient.go
soniaappasamy: Okay, I moved this to a new `interactive.go` file in the `cli` package. Was gonna just put it in `serve.go` but then we'll probably end up moving it in the future (there's been some chatter about using the interactive consent flow for SSH too). Happy to move it to `serve.go` for now though if you've got strong opinions on that!

soniaappasamy: Actually, I ended up changing my mind and keeping it in `serve.go` for now to keep things simple. Take a quick look whenever you get a chance.

cmd/tailscale/cli/funnel.go
```diff
@@ -117,6 +120,50 @@ func (e *serveEnv) runFunnel(ctx context.Context, args []string) error {
...
+		if err := ipn.CheckFunnelAccess(port, st.Self.Capabilities); err != nil {
+			return err
+		}
+	default:
+		// Done with enablement, make sure the requested port is allowed.
+		if err := ipn.CheckFunnelPort(port, st.Self.Capabilities); err != nil {
+			return err
+		}
+	}
+	fmt.Fprintln(os.Stdout, "\nSuccess.")
```
bradfitz, line 163: this is surprising. Have the caller write this instead?

cmd/tailscale/cli/serve.go
```diff
@@ -766,3 +768,67 @@ func parseServePort(s string) (uint16, error) {
...
+		os.Exit(0)
+	}
+	defer watcher.Close()
+	for {
+		n, err := watcher.Next()
+		if err != nil {
+			// Stop blocking if we error.
+			// Let the user finish enablement then rerun their
+			// command themselves.
+			os.Exit(0)
```
bradfitz, line 826: why 0? isn't this bad? log.Fatalf("lost connection to tailscaled: %v", err) ?

tailcfg/tailcfg.go
```diff
@@ -1964,10 +1964,11 @@ const (
 	// Funnel warning capabilities used for reporting errors to the user.
 
 	// CapabilityWarnFunnelNoInvite indicates whether Funnel is enabled for the tailnet.
-	// NOTE: In transition from Alpha to Beta, this capability is being reused as the enablement.
+	// This cap is no longer used in client versions released after 2023-08-03.
```
bradfitz, line 1967: date is also fine, but we prefer to mention the capVer. Do you need to bump the capVer for this, if this is the commit that stopped using this? Or did the earlier commit do it?

tailcfg/tailcfg.go
```diff
@@ -1964,10 +1964,11 @@ const (
...
 	CapabilityWarnFunnelNoInvite = "https://tailscale.com/cap/warn-funnel-no-invite"
 
 	// CapabilityWarnFunnelNoHTTPS indicates HTTPS has not been enabled for the tailnet.
+	// This cap is no longer used in client versions released after 2023-08-03.
```
bradfitz, line 1971: same

bradfitz (approved):

tailcfg/tailcfg.go
soniaappasamy: So for this one, the control server actually stopped sending it a while back, but not totally sure when. But we were still checking it in the client code until now. Anyways, yeah let me bump the capVer.

PR #8787, opened 2023-08-03
tool/gocross: ensure there is no mid-build toolchain change

raggi:
Set GOTOOLCHAIN=local as per https://go.dev/doc/toolchain to prevent dynamic switching of the toolchain during a build.

I haven't reproduced directly that this contributes to symptoms we are presently diagnosing, but I am confident this is the behavior that we want.

Updates #cleanup

### DIFF
--- a/tool/gocross/autoflags.go
+++ b/tool/gocross/autoflags.go
@@ -169,6 +169,9 @@ func autoflagsForTest(argv []string, env *Environment, goroot, nativeGOOS, nativ
 	env.Set("CC", "cc")
 	env.Set("TS_LINK_FAIL_REFLECT", boolStr(failReflect))
 	env.Set("GOROOT", goroot)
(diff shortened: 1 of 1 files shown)


tool/gocross/autoflags.go
```diff
@@ -169,6 +169,7 @@ func autoflagsForTest(argv []string, env *Environment, goroot, nativeGOOS, nativ
 	env.Set("CC", "cc")
 	env.Set("TS_LINK_FAIL_REFLECT", boolStr(failReflect))
 	env.Set("GOROOT", goroot)
+	env.Set("GOTOOLCHAIN", "local")
```
danderson, line 172: Add a comment pointing at the docs for GOTOOLCHAIN? I feel like I'll forget what this is almost immediately without a pointer.

danderson (approved):

PR #8801, opened 2023-08-06
client/tailscale: include other 2XX success codes in response check

beyazit:


bradfitz:
We don't use Semantic Commit Messages prefixes and this also doesn't sound like a feature.

Is this a fix? Do we ever respond with a status code that the client doesn't currently handle correctly? Which?

0bl:
> We don't use Semantic Commit Messages prefixes and this also doesn't sound like a feature.
> 
> Is this a fix? Do we ever respond with a status code that the client doesn't currently handle correctly? Which?

no, i just saw there was a TODO while reviewing the codebase and made a pull request

PR #8812, opened 2023-08-08
ipn/ipnlocal: make tests pass when offline

maisem:
Updates #cleanup

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -8,6 +8,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"flag"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/local.go
```diff
@@ -91,6 +93,14 @@ import (
 	"tailscale.com/wgengine/wgcfg/nmcfg"
 )
 
+var lazyInTest lazy.SyncValue[bool]
+
+func inTest() bool {
+	return lazyInTest.Get(func() bool {
+		return flag.Lookup("test.v") != nil
+	})
+}
```
bradfitz, line 102: sure, but this should live elsewhere like we have in corp. Then corp can use the oss copy too.

bradfitz (approved):

ipn/ipnlocal/local.go
maisem: yeah, we have a few copies of this in oss. Any package preferences?

bradfitz: We have `util/cibuild` for just `func On() bool`.

Something beside that? `util/testenv`? `testenv.InTest()` ?

maisem: Sent #8815

PR #8815, opened 2023-08-08
util/testenv: add new package to hold InTest

maisem:
Removes duplicated code.

Updates #cleanup


util/testenv/testenv.go
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+// Package testenv provides utility functions for tests.
```
bradfitz, line 4: Mention that this package specifically does not depend on the testing package? Might even be worth locking that in with a test.

We have a package to do negative tests now! Search for `deptest.DepChecker` and look at existing ones.

bradfitz (approved):

util/testenv/testenv.go
maisem: Done, PTAL

bradfitz (approved):

PR #8818, opened 2023-08-08
cmd/tailscale: refactor shared utility methods

willnorris:
Refactor two shared functions used by the tailscale cli, `calcAdvertiseRoutes` and `licensesURL`. These are used by the web client as well as other tailscale subcommands. The web client is being moved out of the cli package, so move these two functions to new locations.

Updates tailscale/corp#13775

PR #8819, opened 2023-08-09
client/web: extract web client from cli package 

willnorris:
move the tailscale web client out of the cmd/tailscale/cli package, into a new client/web package.  The remaining cli/web.go file is still responsible for parsing CLI flags and such, and then calls into client/web. This will allow the web client to be hooked into from other contexts (for example, from a tsnet server), and provide a dedicated space to add more functionality to this client.

This PR is purely refactoring; no other changes.

Updates https://github.com/tailscale/corp/issues/13775

(This PR builds on #8818, so ignore the first commit)


soniaappasamy (approved):
Refactor LGTM

PR #8826, opened 2023-08-09
client/web: move api handler into web.go

soniaappasamy:
Also uses `http.HandlerFunc` to pass the handler into `csrfProtect` so we can get rid of the extraneous `api` struct.


client/web/web.go
```diff
@@ -280,6 +293,15 @@ func Handle(w http.ResponseWriter, r *http.Request) {
...
 	}
+	switch {
+	case strings.HasPrefix(r.URL.Path, "/localapi/"):
+		if c.APIHander == nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			json.NewEncoder(w).Encode("missing api handler")
+			return
+		}
+		c.APIHander.proxyRequestToLocalAPI(w, r)
+	}
```
soniaappasamy, line 304: May not want this piece quite yet.

client/web/api.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"io"
+	"net/http"
+	"strings"
+
+	"tailscale.com/client/tailscale"
+	"tailscale.com/client/tailscale/apitype"
+)
+
+type apiHandler struct {
```
willnorris, line 15: do we need the extra apiHandler type?  What if we just put the *tailscale.LocalClient directly on the main WebClient type (or whatever we call it)?

soniaappasamy: yeah fine with that, probably don't need the extra layer of abstraction

client/web/web.go
```diff
@@ -280,6 +293,15 @@ func Handle(w http.ResponseWriter, r *http.Request) {
 		io.WriteString(w, authenticationRedirectHTML)
 		return
 	}
+	switch {
+	case strings.HasPrefix(r.URL.Path, "/localapi/"):
+		if c.APIHander == nil {
```
willnorris, line 298: if we remove the apihandler type, then even a zero value tailscale.LocalClient should function, so I don't think we'd need any special error handling here?

client/web/api.go
```diff
@@ -0,0 +1,58 @@
...
+func newAPIHandler() *apiHandler {
+	return &apiHandler{lc: &localClient}
+}
+
+// proxyRequestToLocalAPI proxies the web API request to the localapi.
+//
+// The web API request path is expected to exactly match a localapi path,
+// with prefix /localapi/.
+func (h *apiHandler) proxyRequestToLocalAPI(w http.ResponseWriter, r *http.Request) {
+	if !strings.HasPrefix(r.URL.Path, "/localapi/") {
```
willnorris, line 28: big TODO here about securing this endpoint before merging this 😄

soniaappasamy:
Repurposed this PR to _just_ include the initial move of api.go into web.go. Will follow-up with the localapi proxy changes.

willnorris (approved):

PR #8827, opened 2023-08-09
cmd/tailscale/cli,clientupdate: extract new clientupdate package

awly:
Extract the self-update logic from cmd/tailscale/cli into a standalone package that could be used from tailscaled later.
The update setup code was refactored a bit compared to original.
The actual `updater.update*` methods are largely the same, with some extra validation and calls to `log.Print*`/`fmt.Print*` replaced with the pluggable logger.

Updates #6995


clientupdate/clientupdate.go
```diff
@@ -0,0 +1,888 @@
...
+	// Track is the release track to get updated version for. Defaults to
+	// Current, which means using the same track as the current running binary.
+	Track Track
+	// Version is the specific version to update to. This field may not be
+	// supported on all platforms (like Arch and Alpine Linuxes or FreeBSD).
+	// Mutually exclusive with Track.
+	Version string
+	// AppStore forces a local app store check, even if the current binary was
+	// not installed via an app store.
+	AppStore bool
```
noncombatant, line 83: Leave it for now, but there is a possible future in which I remove it, FYI.

clientupdate/clientupdate.go
```diff
@@ -0,0 +1,888 @@
...
+}
+
+// UpdateArgs contains arguments needed to run an update.
+type UpdateArgs struct {
+	// Track is the release track to get updated version for. Defaults to
+	// Current, which means using the same track as the current running binary.
+	Track Track
+	// Version is the specific version to update to. This field may not be
+	// supported on all platforms (like Arch and Alpine Linuxes or FreeBSD).
+	// Mutually exclusive with Track.
```
noncombatant, line 79: Call me an absolute maniac if you must — and indeed you must: Since Go doesn't have a way to express sum types, maybe this means we should just give in to the Dark Side and combine Track and Version into a single stringly typed field, with special constant values Current, Stable, and Unstable that we treat magically.

I can't believe I thought that thought and then typed it in. My Rust license is going to be revoked

noncombatant (approved):
LGTM either way; your call on the Track + Version issue.

clientupdate/clientupdate.go
awly: we still have separate flags on the CLI, but it does simplify the API of the library. I'll give it a shot and report back!

awly: updated, WDYT?

noncombatant (approved):

PR #8828, opened 2023-08-09
client/web: add web client Server struct

soniaappasamy:
Updates tailscale/corp#13775

### DIFF
--- a/client/web/web.go
+++ b/client/web/web.go
@@ -44,7 +44,25 @@ var authenticationRedirectHTML string
 
 var tmpl *template.Template
 
-var localClient tailscale.LocalClient
+// Server is the backend server for a Tailscale web client.
+type Server struct {
+	devMode bool
+	lc      *tailscale.LocalClient
+}
+
+// NewServer constructs a new Tailscale web client server.
+//
+// lc is an optional parameter. When not filled, NewServer
+// initializes its own tailscale.LocalClient.
+func NewServer(devMode bool, lc *tailscale.LocalClient) *Server {
+	if lc == nil {
+		lc = &tailscale.LocalClient{}
+	}
+	return &Server{
+		devMode: devMode,
+		lc:      lc,
+	}
+}
 
 func init() {
 	tmpl = template.Must(template.New("web.html").Parse(webHTML))
@@ -264,8 +282,8 @@ req.send(null);
 </body></html>
 `
 
-// Handle processes all requests for the Tailscale web client.
-func Handle(w http.ResponseWriter, r *http.Request) {
+// ServeHTTP processes all requests for the Tailscale web client.
+func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	if authRedirect(w, r) {
 		return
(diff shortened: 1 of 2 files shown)


client/web/web.go
```diff
@@ -46,6 +46,20 @@ var tmpl *template.Template
 
 var localClient tailscale.LocalClient
```
willnorris, line 47: I think we can remove this localClient now?

willnorris (approved):

client/web/web.go
```diff
@@ -44,7 +44,19 @@ var authenticationRedirectHTML string
 
 var tmpl *template.Template
 
-var localClient tailscale.LocalClient
+// Server is the backend server for a Tailscale web client.
+type Server struct {
+	devMode bool
```
willnorris, line 49: What about exporting these fields, and just letting callers set them directly?  For example, in the tsnet use case, the caller is going to need to set the `*tailscale.LocalClient` to the one associated with the tsnet server.  So something like:

``` go
s := new(tsnet.Server)
lc, _ := s.LocalClient()
ws := &web.Server{
    LocalClient: lc,
}
```

client/web/web.go
```diff
@@ -265,7 +277,7 @@ req.send(null);
 `
 
 // Handle processes all requests for the Tailscale web client.
-func Handle(w http.ResponseWriter, r *http.Request) {
+func (s *Server) Handle(w http.ResponseWriter, r *http.Request) {
```
willnorris, line 280: oh, if we're making this a method on the Server type, then we should rename it to `ServeHTTP` so then it implements the `http.Handler` interface directly.  So then callers could just do:

```go
ws := new(web.Server)
http.ListenAndServe(":8080", ws)
```

soniaappasamy: nice okay!

client/web/web.go
soniaappasamy: thoughts on having `NewServer` accept a `*tailscale.LocalClient` instead?

willnorris: yeah, that could certainly work too.  I suspect we have prior art for both throughout the codebase.  I'm just not sure how long this list of method params would grow.  Maybe not too bad.

willnorris: I was trying to think about the the zero value `web.Server` being usable.  But if LocalClient is a pointer, then that doesn't really work.  So in that case, I guess a New method is clearer in how it should be used.

soniaappasamy: That way if we ever have special stuff in there that callers don't need to provide then we don't have to switch stuff over. Could even make the local client arg optional...

cmd/tailscale/cli/web.go
```diff
@@ -76,7 +78,8 @@ func runWeb(ctx context.Context, args []string) error {
 		return fmt.Errorf("too many non-flag arguments: %q", args)
 	}
 
-	webHandler := http.HandlerFunc(web.Handle)
+	webServer := web.NewServer(webArgs.dev, nil)
+	webHandler := http.HandlerFunc(webServer.ServeHTTP)
```
willnorris, line 82: there's no need for `webHandler` at all now.  You can just pass webServer directly wherever you need an `http.Handler` below.  e.g.

```go
cgi.Serve(webServer)
```

willnorris (approved):
one final simplification, but LGTM

PR #8830, opened 2023-08-09
control/controlhttp: remove tstest.Clock from tests

awly:
These specific tests rely on some timers in the controlhttp code. Without time moving forward and timers triggering, the tests fail.

Updates #8587

### DIFF
--- a/control/controlhttp/http_test.go
+++ b/control/controlhttp/http_test.go
@@ -26,6 +26,7 @@ import (
 	"tailscale.com/net/tsdial"
 	"tailscale.com/tailcfg"
 	"tailscale.com/tstest"
+	"tailscale.com/tstime"
 	"tailscale.com/types/key"
 	"tailscale.com/types/logger"
 )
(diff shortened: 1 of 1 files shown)


control/controlhttp/http_test.go
```diff
@@ -205,7 +205,7 @@ func testControlHTTP(t *testing.T, param httpTestParam) {
 		Logf:                 t.Logf,
 		omitCertErrorLogging: true,
 		testFallbackDelay:    50 * time.Millisecond,
-		Clock:                &tstest.Clock{},
```
andrew-d, line 208: Worth noting that `tstest.Clock` supports fake "advances", which cause simulated time to move forward; you can create a clock with `NewClock` and the `Step` option:
https://pkg.go.dev/tailscale.com/tstest#ClockOpts

awly: I've spent some time trying to get this to work, but only got one of the tests passing with `tstest.Clock`. I'm not familiar with this package enough to get the advance hooks in all the right places.
OK if we merge this PR like this to unbreak CI for all other PRs, and update the other test later?

andrew-d (approved):
👍 to merging to fix, but would you mind filing a GitHub issue and mentioning that so we don't forget it?

PR #8836, opened 2023-08-10
control/controlclient: add Auto.informRoutine

maisem:
Instead of having updates replace the map polls, create a third goroutine which is solely responsible for making sure that control is aware of the latest client state.

This also makes it so that the streaming map polls are only broken when there are auth changes, or the client is paused.

Updates https://github.com/tailscale/corp/issues/5761

### DIFF
--- a/control/controlclient/auto.go
+++ b/control/controlclient/auto.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/http"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"tailscale.com/health"
@@ -46,6 +47,91 @@ func (g *LoginGoal) sendLogoutError(err error) {
 
 var _ Client = (*Auto)(nil)
 
+// waitUnpause waits until the client is unpaused then returns. It only
+// returns an error if the client is closed.
+func (c *Auto) waitUnpause(routineLogName string) error {
+	c.mu.Lock()
+	if !c.paused {
+		c.mu.Unlock()
+		return nil
+	}
+	unpaused := c.unpausedChanLocked()
+	c.mu.Unlock()
+	c.logf("%s: awaiting unpause", routineLogName)
+	select {
+	case <-unpaused:
+		c.logf("%s: unpaused", routineLogName)
+		return nil
+	case <-c.quit:
+		return errors.New("quit")
+	}
+}
+
+// updateRoutine is responsible for informing the server of worthy changes to
+// our local state. It runs in its own goroutine.
+func (c *Auto) updateRoutine() {
+	defer close(c.updateDone)
+	bo := backoff.NewBackoff("updateRoutine", c.logf, 30*time.Second)
+	for {
+		if err := c.waitUnpause("updateRoutine"); err != nil {
+			c.logf("updateRoutine: exiting")
+			return
+		}
+		c.mu.Lock()
+		gen := c.lastUpdateGen
+		ctx := c.mapCtx
+		needUpdate := gen > 0 && gen != c.lastUpdateGenInformed && c.loggedIn
+		c.mu.Unlock()
+
+		if needUpdate {
+			select {
+			case <-c.quit:
+				c.logf("updateRoutine: exiting")
+				return
+			default:
+			}
+		} else {
+			// Nothing to do, wait for a signal.
+			select {
+			case <-c.quit:
+				c.logf("updateRoutine: exiting")
+				return
+			case <-c.updateCh:
+				continue
+			}
+		}
+
+		t0 := c.clock.Now()
+		err := c.direct.SendUpdate(ctx)
+		d := time.Since(t0).Round(time.Millisecond)
+		if err != nil {
+			if ctx.Err() == nil {
+				c.direct.logf("lite map update error after %v: %v", d, err)
+			}
+			bo.BackOff(ctx, err)
+			continue
+		}
+		bo.BackOff(ctx, nil)
+		c.direct.logf("[v1] successful lite map update in %v", d)
+
+		c.mu.Lock()
+		c.lastUpdateGenInformed = gen
+		c.mu.Unlock()
+	}
+}
+
+// atomicGen is an atomic int64 generator. It is used to generate monotonically
+// increasing numbers for updateGen.
+var atomicGen atomic.Int64
+
+func nextUpdateGen() updateGen {
+	return updateGen(atomicGen.Add(1))
+}
+
+// updateGen is a monotonically increasing number that represents a particular
+// update to the local state.
+type updateGen int64
+
 // Auto connects to a tailcontrol server for a node.
 // It's a concrete implementation of the Client interface.
 type Auto struct {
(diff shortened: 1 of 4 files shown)


tailcfg/tailcfg.go
```diff
@@ -1082,8 +1083,18 @@ type MapRequest struct {
...
-	Stream      bool // if true, multiple MapResponse objects are returned
-	Hostinfo    *Hostinfo
+
+	// Stream is whether the client wants to receive multiple MapResponses over
+	// the same HTTP connection.
+	//
+	// If false, the server will send a single MapResponse and then close the
+	// connection.
+	//
+	// If true and Version >= 68, the server can treat this as a read-only
```
bradfitz, line 1093: s/can/will/ ? Or "should", probably?

tailcfg/tailcfg.go
```diff
@@ -1082,8 +1083,18 @@ type MapRequest struct {
...
+	// request and ignore any Hostinfo or other fields that might be set.
+	Stream bool
+
+	Hostinfo *Hostinfo
```
bradfitz, line 1097: docs while you're at it? repeating a bit when it's applicable and what nil means.

tailcfg/tailcfg.go
```diff
@@ -1114,13 +1125,12 @@ type MapRequest struct {
 	// It is encoded as tka.AUMHash.MarshalText.
 	TKAHead string `json:",omitempty"`
 
-	// ReadOnly is whether the client just wants to fetch the
-	// MapResponse, without updating their Endpoints. The
-	// Endpoints field will be ignored and LastSeen will not be
-	// updated and peers will not be notified of changes.
+	// ReadOnly was set when client just wanted to fetch the MapResponse,
```
bradfitz, line 1128: out of curiosity, when did we stop using this?

tailcfg/tailcfg.go
```diff
@@ -105,7 +105,8 @@ type CapabilityVersion int
 //   - 65: 2023-07-12: Client understands DERPMap.HomeParams + incremental DERPMap updates with params
 //   - 66: 2023-07-23: UserProfile.Groups added (available via WhoIs)
 //   - 67: 2023-07-25: Client understands PeerCapMap
-const CurrentCapabilityVersion CapabilityVersion = 67
+//   - 68: 2023-08-09: Client has dedicated informRoutine
```
bradfitz, line 108: Client has dedicated informRoutine; MapRequest.Stream true means ignore Hostinfo+Endpoints

control/controlclient/direct.go
```diff
@@ -771,13 +771,13 @@ func (c *Direct) SetEndpoints(endpoints []tailcfg.Endpoint) (changed bool) {
 // It always returns a non-nil error describing the reason for the failure
 // or why the request ended.
 func (c *Direct) PollNetMap(ctx context.Context, cb func(*netmap.NetworkMap)) error {
-	return c.sendMapRequest(ctx, -1, false, cb)
+	return c.sendMapRequest(ctx, true, cb)
 }
 
 // FetchNetMap fetches the netmap once.
 func (c *Direct) FetchNetMap(ctx context.Context) (*netmap.NetworkMap, error) {
```
bradfitz, line 778: this is only used by tests? Rename to add ForTest suffix maybe?

control/controlclient/auto.go
```diff
@@ -46,6 +47,79 @@ func (g *LoginGoal) sendLogoutError(err error) {
...
+				c.logf("informRoutine: exiting")
+				return
+			case <-c.informCh:
+				continue
+			}
+		}
+
+		t0 := c.clock.Now()
+		err := c.direct.SendLiteMapUpdate(ctx)
+		d := time.Since(t0).Round(time.Millisecond)
```
bradfitz, line 106: time.Since is mixing clocks :(

control/controlclient/auto.go
```diff
@@ -46,6 +47,79 @@ func (g *LoginGoal) sendLogoutError(err error) {
...
+		if err := c.waitUnpause("informRoutine"); err != nil {
+			c.logf("informRoutine: exiting")
+			return
+		}
+		c.mu.Lock()
+		at := c.lastUpdateAt
+		ctx := c.mapCtx
+		needUpdate := at != c.lastUpdateInformed && !at.IsZero()
+		c.mu.Unlock()
+
```
bradfitz, line 85: how about a new type for this informRoutine and breaking up this up into methods?

control/controlclient/auto.go
```diff
@@ -46,6 +47,79 @@ func (g *LoginGoal) sendLogoutError(err error) {
...
+				continue
+			}
+		}
+
+		t0 := c.clock.Now()
+		err := c.direct.SendLiteMapUpdate(ctx)
+		d := time.Since(t0).Round(time.Millisecond)
+		if err != nil {
+			if ctx.Err() == nil {
+				c.direct.logf("lite map update after %v: %v", d, err)
```
bradfitz, line 109: get the word "error" in here

control/controlclient/auto.go
```diff
@@ -46,6 +47,79 @@ func (g *LoginGoal) sendLogoutError(err error) {
...
+			}
+			bo.BackOff(ctx, err)
```
bradfitz, line 111: so I worry this can wait up to 30 seconds, even if something else network connectivity-wise has improved/changed

what wakes this up if it turns out the network is back?

control/controlclient/auto.go
```diff
@@ -54,32 +128,38 @@ type Auto struct {
 	logf       logger.Logf
 	expiry     *time.Time
 	closed     bool
+	informCh   chan struct{} // readable when we should inform the server of a change
```
bradfitz, line 131: oh, sad, I was hoping it'd include the actual change and the informRoutine would track what it's sent

I see now how scary this is

control/controlclient/auto.go
```diff
@@ -170,73 +253,20 @@ func (c *Auto) Start() {
 //
 // It should be called whenever there's something new to tell the server.
 func (c *Auto) sendNewMapRequest() {
+	at := mono.Now()
```
bradfitz, line 256: isn't this only like 16ms resolution on Windows?

You'd be better off with an atomic integer sequence number.

control/controlclient/auto.go
maisem: This is sadly true for the mapRoutine too. There isn't a way to get out of the backoff without canceling the ctx. I can take a stab at that in a follow up change

control/controlclient/auto.go
```diff
@@ -162,152 +255,78 @@ func (c *Auto) SetPaused(paused bool) {
...
 	}
 	if !c.closed {
 		c.mapCtx, c.mapCancel = context.WithCancel(context.Background())
 		c.mapCtx = sockstats.WithSockStats(c.mapCtx, sockstats.LabelControlClientAuto, c.logf)
-
 	}
 }
 
-func (c *Auto) cancelMapUnsafely() {
+// cancelMap cancels the existing mapPoll and liteUpdates.
```
bradfitz, line 306: liteUpdates == inform?

maisem: yeah

bradfitz (approved):
LGTM after misc nitpick things as discussed here & slack

control/controlclient/auto.go
maisem: :(

PR #8842, opened 2023-08-10
tool/{node,yarn}: update node and yarn tools

soniaappasamy:
Syncing these up with what we've got in corp.

Updates tailscale/corp#13775


willnorris (approved):

willnorris:
wasm failure looks legit, caused by the yarn changes it looks like

PR #8845, opened 2023-08-10
VERSION.txt: this is v1.49.0

awly:


DentonGentry:
The intent is to merge this next Wednesday?

DentonGentry (changes requested):
Sorry, but I'm marking this Request Changes to make sure we don't submit this today.

awly:
My intent was to merge this today, since 1.48 will be today's `main` plus some cherry-picks. So I assumed that 1.49 should be today's `main` too.

awly:
Discussed offline, we will cut a release branch later and keep the current version on `main` as is.

Issue #8142, opened 2023-05-15
Add winsock layered service providers to Windows troubleshooting log


dblohm7:
LSPs are deprecated but still supported. Lots of software out there still uses them. They might interfere with Tailscale. We should log them.

PR #8854, opened 2023-08-11
util/osdiag: add logging for winsock layered service providers to Win…

dblohm7:
…dows bugreports

The Layered Service Provider (LSP) is a deprecated (but still supported) mechanism for inserting user-mode DLLs into a filter chain between the Winsock API surface (ie, ws2_32.dll) and the internal user-mode interface to the networking stack.

While their use is becoming more rare due to the aforementioned deprecation, it is still possible for third-party software to install their DLLs into this filter chain and interfere with Winsock API calls. Knowing whether this is happening is useful for troubleshooting.

Fixes https://github.com/tailscale/tailscale/issues/8142

### DIFF
--- a/util/osdiag/mksyscall.go
+++ b/util/osdiag/mksyscall.go
@@ -7,3 +7,6 @@ package osdiag
 //go:generate go run golang.org/x/tools/cmd/goimports -w zsyscall_windows.go
 
 //sys regEnumValue(key registry.Key, index uint32, valueName *uint16, valueNameLen *uint32, reserved *uint32, valueType *uint32, pData *byte, cbData *uint32) (ret error) [failretval!=0] = advapi32.RegEnumValueW
+//sys wscEnumProtocols(iProtocols *int32, protocolBuffer *wsaProtocolInfo, bufLen *uint32, errno *int32) (ret int32) = ws2_32.WSCEnumProtocols
+//sys wscGetProviderInfo(providerId *windows.GUID, infoType _WSC_PROVIDER_INFO_TYPE, info unsafe.Pointer, infoSize *uintptr, flags uint32, errno *int32) (ret int32) = ws2_32.WSCGetProviderInfo
+//sys wscGetProviderPath(providerId *windows.GUID, providerDllPath *uint16, providerDllPathLen *int32, errno *int32) (ret int32) = ws2_32.WSCGetProviderPath
--- a/util/osdiag/osdiag_windows.go
+++ b/util/osdiag/osdiag_windows.go
@@ -22,6 +22,10 @@ import (
 	"tailscale.com/util/winutil/authenticode"
 )
 
+var (
+	errUnexpectedResult = errors.New("API call returned an unexpected value")
+)
+
 const (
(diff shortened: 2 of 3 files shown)


util/osdiag/osdiag_windows.go
```diff
@@ -328,3 +340,144 @@ func getModuleInfo() (map[string]moduleInfo, error) {
...
+	if ret != _SOCKET_ERROR {
+		return nil, errUnexpectedResult
+	}
+	if e := windows.Errno(errno); e != windows.WSAENOBUFS {
+		return nil, e
+	}
+
+	// Allocate
+	szEntry := uint32(unsafe.Sizeof(wsaProtocolInfo{}))
+	buf := make([]wsaProtocolInfo, bytesReqd/szEntry)
```
sailorfrag, line 473: Should there be a check that `bytesReqd` is a multiple of `szEntry`?

util/osdiag/mksyscall.go
```diff
@@ -7,3 +7,6 @@ package osdiag
 //go:generate go run golang.org/x/tools/cmd/goimports -w zsyscall_windows.go
 
 //sys regEnumValue(key registry.Key, index uint32, valueName *uint16, valueNameLen *uint32, reserved *uint32, valueType *uint32, pData *byte, cbData *uint32) (ret error) [failretval!=0] = advapi32.RegEnumValueW
+//sys wscEnumProtocols(iProtocols *int32, protocolBuffer *wsaProtocolInfo, bufLen *uint32, errno *int32) (ret int32) = ws2_32.WSCEnumProtocols
```
sailorfrag, line 10: `protocolBuffer` should probably be `*windows.WSAProtocolInfo` I think?

sailorfrag (approved):
This seems fine at present, but I'm worried about `wsaProtocolInfo` being a footgun.

util/osdiag/osdiag_windows.go
dblohm7: We actually deal with that by computing `bufLen` independently of  `bytesReqd` in the following line to account for any potential rounding down during the integer division.

util/osdiag/osdiag_windows.go
```diff
@@ -328,3 +340,143 @@ func getModuleInfo() (map[string]moduleInfo, error) {
...
+	providerInfoLspCategories _WSC_PROVIDER_INFO_TYPE = 0
+)
+
+const (
+	_SOCKET_ERROR = -1
+)
+
+type wsaProtocolInfo struct {
+	windows.WSAProtocolInfo
+}
```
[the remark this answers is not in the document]
dblohm7: sgtm

util/osdiag/mksyscall.go
dblohm7: Strictly speaking, yes, but since the ABI of `wsaProtocolInfo` is expected to be identical to `windows.WSAProtocolInfo` (I'm just using the new type so that I can use it as a receiver in some of the funcs), it seemed a lot cleaner to do it this way. I'll add a comment.

PR #8859, opened 2023-08-11
tsweb: add request ID for errors

sailorfrag:
If an optional request ID generating func is supplied to StdHandler, then requests that return an error will be logged with a request ID that is also shown as part of the response.

Updates tailscale/corp#2549

Change-Id: Ic7499706df42f95b6878d44d4aab253e2fc6a69b

### DIFF
--- a/tsweb/log.go
+++ b/tsweb/log.go
@@ -45,6 +45,12 @@ type AccessLogRecord struct {
 	Bytes int `json:"bytes,omitempty"`
 	// Error encountered during request processing.
 	Err string `json:"err,omitempty"`
+	// RequestID is a unique ID for this request. When a request fails due to an
+	// error, the ID is generated and displayed to the client immediately after
+	// the error text, as well as logged here. This makes it easier to correlate
+	// support requests with server logs. If a RequestID generator is not
+	// configured, RequestID will be empty.
+	RequestID RequestID `json:"request_id,omitempty"`
 }
 
 // String returns m as a JSON string.
--- a/tsweb/tsweb.go
+++ b/tsweb/tsweb.go
@@ -169,7 +169,8 @@ type ReturnHandler interface {
 type HandlerOptions struct {
 	QuietLoggingIfSuccessful bool // if set, do not log successfully handled HTTP requests (200 and 304 status codes)
 	Logf                     logger.Logf
-	Now                      func() time.Time // if nil, defaults to time.Now
+	Now                      func() time.Time              // if nil, defaults to time.Now
+	GenerateRequestID        func(*http.Request) RequestID // if nil, no request IDs are generated
 
 	// If non-nil, StatusCodeCounters maintains counters
 	// of status codes for handled responses.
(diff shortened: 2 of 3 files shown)


tsweb/log.go
```diff
@@ -45,6 +45,10 @@ type AccessLogRecord struct {
 	Bytes int `json:"bytes,omitempty"`
 	// Error encountered during request processing.
 	Err string `json:"err,omitempty"`
+	// The unique ID for this request. When a request fails due to an error, the
```
bradfitz, line 48: I was going to say these docs aren't in the usual style (complete sentences starting with the name of the field) but then I noticed none of them are. At least fix this one and we can clean the others later?

tsweb/log.go
```diff
@@ -45,6 +45,10 @@ type AccessLogRecord struct {
...
+	// ID is generated and displayed as part of the error message for the
+	// client.
```
bradfitz, line 50: Any guidance here or elsewhere on its format? I guess more than digits. Whitespace? What's the allowed alphabet of characters?

tsweb/tsweb.go
```diff
@@ -169,7 +169,8 @@ type ReturnHandler interface {
 type HandlerOptions struct {
 	QuietLoggingIfSuccessful bool // if set, do not log successfully handled HTTP requests (200 and 304 status codes)
 	Logf                     logger.Logf
-	Now                      func() time.Time // if nil, defaults to time.Now
+	Now                      func() time.Time           // if nil, defaults to time.Now
+	GenerateRequestID        func(*http.Request) string // if nil, no request IDs are generated
```
bradfitz, line 173: Maybe if we use a `type RequestID string` type we can put the docs about recommended format in one place

tsweb/tsweb.go
```diff
@@ -296,14 +302,23 @@ func (h retHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			if hErr.RequestID != "" {
+				fmt.Fprintln(lw, hErr.RequestID)
+			}
 		}
 	case err != nil:
+		errorMessage := "internal server error"
+
+		if h.opts.GenerateRequestID != nil {
+			msg.RequestID = h.opts.GenerateRequestID(r)
+			errorMessage = errorMessage + "\n" + msg.RequestID
```
bradfitz, line 314: Yeah, the docs on the type might mention it's appended after "internal server error\n" so should have a prefix that sounds like a useful identifier for humans reporting bugs. (Having just a number or just hex would look weird.)

tsweb/log.go
sailorfrag: Yeah it seemed weird but I went with matching the surrounding comment format.

I've fixed this one.

tsweb/log.go
sailorfrag: I didn't want `tsweb` to be particularly opinionated. Anything printable is fine. But I took your suggestion regarding adding a type and documenting the format there.

tsweb/tsweb.go
sailorfrag: Done

tsweb/tsweb.go
sailorfrag: Included in the docs.

tsweb/log.go
```diff
@@ -45,6 +45,11 @@ type AccessLogRecord struct {
 	Bytes int `json:"bytes,omitempty"`
 	// Error encountered during request processing.
 	Err string `json:"err,omitempty"`
+	// RequestID is a unique ID for this request. When a request fails due to an
+	// error, the ID is generated and displayed to the client immediately after
+	// the error text, as well as logged here. This makes it easier to correlate
+	// support requests with server logs.
```
bradfitz, line 51: mention that it may be empty if not configured

bradfitz (approved):

sailorfrag:
Alright, I'm going to hold off on merging this until the stable release is cut unless I hear a strong opinion that I should merge right away. I don't expect it to affect the release, but there's minimal cost in delaying for a few days.

PR #8860, opened 2023-08-11
net/netns: indicate when android protect func is no longer nil

catzkorn:
We have been getting into routing loops due to the timing of when we bind sockets on starting the tailscale android app. At this point, we do not have access to `VpnService.protect()` and are unable to protect the magicsock sockets, which causes a routing loop issue until we forcibly rebind about 10 minutes into the service being started.

This change returns a bool from SetAndroidProtectFunc we get access to a func.

Updates tailscale/corp#13814

### DIFF
--- a/net/netns/netns_android.go
+++ b/net/netns/netns_android.go
@@ -44,10 +44,23 @@ func UseSocketMark() bool {
 // fwmark return errors on Android. The actual implementation of
 // VpnService.protect ends up doing an IPC to another process on
 // Android, asking for the fwmark to be set.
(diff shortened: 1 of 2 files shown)


bradfitz (commented):
Docs aren't super clear on what the bool result means

bradfitz:
Actually I think it's worded backwards?

catzkorn:
> Actually I think it's worded backwards?

Yeah, fixing now. Bools suffer the same fate as left and right for me. 50 percent chance I get it correct 50 percent of the time.

PR #8861, opened 2023-08-11
cmd/tailscale,ipn: add auto-update flags and prefs

awly:
The flags are hidden for now. Adding propagation to tailscaled and persistence only.

Updates #6907

### DIFF
--- a/cmd/tailscale/cli/set.go
+++ b/cmd/tailscale/cli/set.go
@@ -11,6 +11,7 @@ import (
 	"net/netip"
 
 	"github.com/peterbourgon/ff/v3/ffcli"
+	"tailscale.com/clientupdate"
 	"tailscale.com/ipn"
 	"tailscale.com/net/netutil"
 	"tailscale.com/net/tsaddr"
@@ -46,6 +47,8 @@ type setArgsT struct {
 	acceptedRisks          string
 	profileName            string
 	forceDaemon            bool
+	updateCheck            bool
+	updateApply            bool
 }
 
 func newSetFlagSet(goos string, setArgs *setArgsT) *flag.FlagSet {
@@ -61,6 +64,8 @@ func newSetFlagSet(goos string, setArgs *setArgsT) *flag.FlagSet {
 	setf.StringVar(&setArgs.hostname, "hostname", "", "hostname to use instead of the one provided by the OS")
 	setf.StringVar(&setArgs.advertiseRoutes, "advertise-routes", "", "routes to advertise to other nodes (comma-separated, e.g. \"10.0.0.0/8,192.168.0.0/24\") or empty string to not advertise routes")
 	setf.BoolVar(&setArgs.advertiseDefaultRoute, "advertise-exit-node", false, "offer to be an exit node for internet traffic for the tailnet")
+	setf.BoolVar(&setArgs.updateCheck, "update-check", true, "HIDDEN: notify about available Tailscale updates")
+	setf.BoolVar(&setArgs.updateApply, "auto-update", false, "HIDDEN: automatically update to the latest available version")
 	if safesocket.GOOSUsesPeerCreds(goos) {
 		setf.StringVar(&setArgs.opUser, "operator", "", "Unix username to allow to operate on tailscaled without sudo")
 	}
@@ -99,6 +104,10 @@ func runSet(ctx context.Context, args []string) (retErr error) {
 			Hostname:               setArgs.hostname,
 			OperatorUser:           setArgs.opUser,
 			ForceDaemon:            setArgs.forceDaemon,
+			AutoUpdate: ipn.AutoUpdatePrefs{
+				Check: setArgs.updateCheck,
+				Apply: setArgs.updateApply,
+			},
 		},
 	}
 
@@ -141,6 +150,12 @@ func runSet(ctx context.Context, args []string) (retErr error) {
 		wantSSH, haveSSH := maskedPrefs.RunSSH, curPrefs.RunSSH
 		if err := presentSSHToggleRisk(wantSSH, haveSSH, setArgs.acceptedRisks); err != nil {
 			return err
+		}
+	}
+	if maskedPrefs.AutoUpdateSet {
+		_, err := clientupdate.NewUpdater(clientupdate.Arguments{})
+		if errors.Is(err, errors.ErrUnsupported) {
+			return errors.New("automatic updates are not supported on this platform")
 		}
 	}
 	checkPrefs := curPrefs.Clone()
--- a/cmd/tailscale/cli/up.go
+++ b/cmd/tailscale/cli/up.go
@@ -97,6 +97,8 @@ func newUpFlagSet(goos string, upArgs *upArgsT, cmd string) *flag.FlagSet {
 	}
 	upf := newFlagSet(cmd)
 
+	// When adding new flags, prefer to put them under "tailscale set" instead
+	// of here. Setting preferences via "tailscale up" is deprecated.
 	upf.BoolVar(&upArgs.qr, "qr", false, "show QR code for login URLs")
 	upf.StringVar(&upArgs.authKeyOrFile, "auth-key", "", `node authorization key; if it begins with "file:", then it's a path to a file containing the authkey`)
 
@@ -712,6 +714,8 @@ func init() {
 	addPrefFlagMapping("operator", "OperatorUser")
 	addPrefFlagMapping("ssh", "RunSSH")
 	addPrefFlagMapping("nickname", "ProfileName")
+	addPrefFlagMapping("update-check", "AutoUpdate")
+	addPrefFlagMapping("auto-update", "AutoUpdate")
 }
 
 func addPrefFlagMapping(flagName string, prefNames ...string) {
--- a/ipn/ipnlocal/c2n.go
+++ b/ipn/ipnlocal/c2n.go
@@ -127,9 +127,10 @@ func (b *LocalBackend) handleC2NUpdate(w http.ResponseWriter, r *http.Request) {
 	//
 	// Note that we create the Updater solely to check for errors; we do not
 	// invoke it here. For this purpose, it is ok to pass it a zero Arguments.
+	prefs := b.Prefs().AutoUpdate()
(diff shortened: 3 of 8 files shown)


noncombatant (approved):

cmd/tailscale/cli/up.go
```diff
@@ -112,6 +112,7 @@ func newUpFlagSet(goos string, upArgs *upArgsT, cmd string) *flag.FlagSet {
 	upf.StringVar(&upArgs.hostname, "hostname", "", "hostname to use instead of the one provided by the OS")
 	upf.StringVar(&upArgs.advertiseRoutes, "advertise-routes", "", "routes to advertise to other nodes (comma-separated, e.g. \"10.0.0.0/8,192.168.0.0/24\") or empty string to not advertise routes")
 	upf.BoolVar(&upArgs.advertiseDefaultRoute, "advertise-exit-node", false, "offer to be an exit node for internet traffic for the tailnet")
+	upf.BoolVar(&upArgs.autoUpdates, "auto-updates", false, "HIDDEN: automatically update to the latest available version in the background")
```
bradfitz, line 115: let's not make this available via deprecated "up"; only "set".

Also, I think it should be "update" (verb) not "updates" (noun)

bradfitz: we might want more than a bool. Security only? What time schedule? What's the plan for configuring stuff like that?

awly: Moved to `set` and added a comment in `newUpFlagSet` for future readers to avoid adding flags there.
Also made the `ipn.Prefs` field a struct instead of bool to account for other features in the future.
I'd like to start simple (client polling daily with some jitter) and add customization later on. First goal is to get updates working in general :sweat_smile:

ipn/prefs.go
```diff
@@ -195,6 +195,9 @@ type Prefs struct {
 	// and CLI.
 	ProfileName string `json:",omitempty"`
 
+	// AutoUpdate settings for the node agent.
```
bradfitz, line 198: complete sentence. there's no verb here.

and what does nil mean?

awly: Updated the comment. Nil means auto-updates are disabled (until we're ready to make this opt-out).

ipn/prefs.go
```diff
@@ -203,6 +208,14 @@ type Prefs struct {
 	Persist *persist.Persist `json:"Config"`
 }
 
+// AutoUpdatePrefs are the auto update settings for the node agent.
+type AutoUpdatePrefs struct {
+	// Enable specifies whether background auto-updates are enabled. When
+	// enabled, tailscaled will periodically check for available updates and
+	// apply them.
+	Enable bool
```
bradfitz, line 216: I still think this is too coarse of a setting.

It's really like four things:

* check / notify
* auto-download the updates (and verify them) but don't install/restart.
* allow remote updates (c2n triggered) based on smart schedule, admin clicks
* do it locally

This lumps it all together into one bool and I'm afraid we're going to end up in a state state (even temporarily) where tailscale nodes are randomly rebooting on users at inopportune times and people will turn it off.

awly: Split the flag into separate `Check` and `Apply` flags. We can add a flag for accepting remote updates when that's actually implemented (and default it to "true" then).  For the initial Beta we're making auto-updates opt-in and checks/notifications opt-out.

For schedule, should we do a time window? A daily check time? Some other spec?
What would you want for your nodes?

When would users want to auto-download updates but not apply them?

awly: @bradfitz as discussed, we will do most of the update triggering via c2n messages from coordination server.
I'd still like these flags, to later hook them up to https://github.com/tailscale/tailscale/blob/9c4364e0b764e39084f207347b5cfd0dfba5a59d/ipn/ipnlocal/c2n.go#L110
The `--update-check` will let the coordination send update c2n messages to the client.
The `--update-apply` will gate whether client actually applies them.
WDYT?

awly:
Made some tweaks, but overall I'd still like to start with 2 flags: `--update-check` and `--auto-update`.
Smart update scheduling can happen on the control plane side, or on the client once it receives the c2n update.
@bradfitz PTAL

noncombatant (approved):
Still LGTM. :)

cmd/tailscale/cli/set.go
```diff
@@ -61,6 +64,8 @@ func newSetFlagSet(goos string, setArgs *setArgsT) *flag.FlagSet {
 	setf.StringVar(&setArgs.hostname, "hostname", "", "hostname to use instead of the one provided by the OS")
 	setf.StringVar(&setArgs.advertiseRoutes, "advertise-routes", "", "routes to advertise to other nodes (comma-separated, e.g. \"10.0.0.0/8,192.168.0.0/24\") or empty string to not advertise routes")
 	setf.BoolVar(&setArgs.advertiseDefaultRoute, "advertise-exit-node", false, "offer to be an exit node for internet traffic for the tailnet")
+	setf.BoolVar(&setArgs.updateCheck, "update-check", true, "HIDDEN: automatically check for available Tailscale updates")
```
bradfitz, line 67: do we still need this one if we can just push down this in MapResponse.ClientVersion? What's to "check"? Or is this really about whether the user wants to be notified about the availability of updates?

cmd/tailscale/cli/set.go
```diff
@@ -61,6 +64,8 @@ func newSetFlagSet(goos string, setArgs *setArgsT) *flag.FlagSet {
...
+	setf.BoolVar(&setArgs.updateApply, "auto-update", false, "HIDDEN: automatically update to the latest available version as soon as it's available")
```
bradfitz, line 68: "as soon as it's available" might be over-promising. we might want to roll it out in waves.

cmd/tailscale/cli/set.go
awly: Yeah, it's whether they will see GUI or CLI notifications. The assumption is that dinner users will want to run older versions and get annoyed at CLI messages telling them about updates

awly: Updated the flag description to better describe it

ipn/ipnlocal/c2n.go
```diff
@@ -127,9 +127,10 @@ func (b *LocalBackend) handleC2NUpdate(w http.ResponseWriter, r *http.Request) {
 	//
 	// Note that we create the Updater solely to check for errors; we do not
 	// invoke it here. For this purpose, it is ok to pass it a zero Arguments.
+	prefs := b.Prefs().AutoUpdate()
 	_, err := clientupdate.NewUpdater(clientupdate.Arguments{})
 	res := tailcfg.C2NUpdateResponse{
-		Enabled:   envknob.AllowsRemoteUpdate(),
+		Enabled:   envknob.AllowsRemoteUpdate() || prefs.Apply || prefs.Check,
```
bradfitz, line 133:
```go
// Enabled indicates whether the user has opted in to updates triggered from
// control.
Enabled bool
```

I don't think `|| prefs.Check` is right here.

ipn/ipnlocal/c2n.go
```diff
@@ -149,6 +150,10 @@ func (b *LocalBackend) handleC2NUpdate(w http.ResponseWriter, r *http.Request) {
 		res.Err = "not supported"
 		return
 	}
+	if prefs.Check && !prefs.Apply {
+		// Client opted in to update notifications but not automatic updates.
+		return
```
bradfitz, line 155: note line 145-147 above. It's already bailing out based on the prefs you set on line 133.

ipn/prefs.go
```diff
@@ -204,6 +208,18 @@ type Prefs struct {
 	Persist *persist.Persist `json:"Config"`
 }
 
+// AutoUpdatePrefs are the auto update settings for the node agent.
+type AutoUpdatePrefs struct {
+	// Check specifies whether background checks for updates are enabled. When
+	// enabled, tailscaled will periodically check for available and notify the
```
bradfitz, line 214: check for available .... updates? (missing word)

ipn/prefs.go
```diff
@@ -204,6 +208,18 @@ type Prefs struct {
...
+	// user about them.
+	Check bool
+	// Apply specifies whether background auto-updates are enabled. When
+	// enabled, tailscaled will apply available updates in the background as
+	// soon as they are available. Check must also be set when Apply is set.
```
bradfitz, line 219: another "as soon as" over promise

ipn/prefs.go
```diff
@@ -414,7 +438,16 @@ func (p *Prefs) Equals(p2 *Prefs) bool {
 		compareIPNets(p.AdvertiseRoutes, p2.AdvertiseRoutes) &&
 		compareStrings(p.AdvertiseTags, p2.AdvertiseTags) &&
 		p.Persist.Equals(p2.Persist) &&
-		p.ProfileName == p2.ProfileName
+		p.ProfileName == p2.ProfileName &&
+		p.AutoUpdate.Equals(p2.AutoUpdate)
+}
+
+func (au AutoUpdatePrefs) Equals(au2 AutoUpdatePrefs) bool {
```
bradfitz, line 445: do you need an Equals method for some reason? it's fine. just curious.

ipn/prefs_test.go
```diff
@@ -372,30 +388,30 @@ func TestPrefsPretty(t *testing.T) {
 		{
 			Prefs{},
 			"linux",
-			"Prefs{ra=false mesh=false dns=false want=false routes=[] nf=off Persist=nil}",
+			"Prefs{ra=false mesh=false dns=false want=false routes=[] nf=off AutoUpdatePrefs{check=false apply=false} Persist=nil}",
```
bradfitz, line 391: can we make these less verbose? we abbreviate the other options a lot.

e.g. `up=no`, `up=notify`, `up=yes`?

ipn/prefs.go
awly: Nope, was just mimicking other code. Removed

ipn/prefs_test.go
bradfitz: you concerned at all that your chosen `upd` looks a lot like `udp=` which we use elsewhere in the netcheck report lines? your call.

bradfitz (approved):

Issue #1909, opened 2021-05-11
Pass through control netmap deltas to engine/GUI/etc


bradfitz:
We have many layers of netmap flattening & delta-ifying going on:

* In control, a single node changes (e.g. a peer changes its Hostinfo or endpoints)
* Control then generates a full netmap to send to the client
* For modern clients (almost all at this point), control minimizes the netmap update down to a delta from the last sent netmap
* The controlclient code then inflates that back to a full netmap and propages it within tailscaled
* Various parts of tailscaled then diff it against the old netmap to compute what's changed and make minimal edits to e.g. wireguard or DNS or routing tables, or wglog.Logger.SetPeers, etc.

It'd be nicer if at least within tailscaled we kept things as mutations and passed those around, rather than recomputing the mutations from the full netmaps in N places. And when we got such mutations from control directly, we could keep them as-is and not even have to "compute" them.

/cc @josharian @danderson @crawshaw

bradfitz:
(The GUIs too, with WatchIPNBus: https://github.com/tailscale/tailscale/issues/13390)

bradfitz:
This increasingly shows up as a problem in big churny tailnets. e.g. this node was doing lots of GC CPU, in part due to all these allocs:

```
(pprof) list maybeReconfig
Total: 1.43TB
ROUTINE ======================== tailscale.com/wgengine.(*userspaceEngine).maybeReconfigWireguardLocked in tailscale.com/wgengine/userspace.go
   95.51GB   192.35GB (flat, cum) 13.18% of Total
         .          .    682:func (e *userspaceEngine) maybeReconfigWireguardLocked(discoChanged map[key.NodePublic]bool) error {
         .          .    683:   if hook := e.testMaybeReconfigHook; hook != nil {
         .          .    684:           hook()
         .          .    685:           return nil
         .          .    686:   }
         .          .    687:
         .          .    688:   full := e.lastCfgFull
         .    53.64GB    689:   e.wgLogger.SetPeers(full.Peers)
         .          .    690:
         .          .    691:   // Compute a minimal config to pass to wireguard-go
         .          .    692:   // based on the full config. Prune off all the peers
         .          .    693:   // and only add the active ones back.
   68.01MB    68.01MB    694:   min := full
   19.46GB    19.46GB    695:   min.Peers = make([]wgcfg.Peer, 0, e.lastNMinPeers)
         .          .    696:
         .          .    697:   // We'll only keep a peer around if it's been active in
[...]

PR #8865, opened 2023-08-12
control/controlclient: refactor in prep for optimized delta handling

bradfitz:
See issue. This is a baby step towards passing through deltas
end-to-end from node to control back to node and down to the various
engine subsystems, not computing diffs from two full netmaps at
various levels. This will then let us support larger netmaps without
burning CPU.

But this change itself changes no behavior. It just changes a func
type to an interface with one method. That paves the way for future
changes to then add new NetmapUpdater methods that do more
fine-grained work than updating the whole world.

Updates #1909

### DIFF
--- a/control/controlclient/auto.go
+++ b/control/controlclient/auto.go
@@ -480,11 +480,42 @@ func (c *Auto) unpausedChanLocked() <-chan struct{} {
 	return unpaused
 }
 
+// mapRoutineState is the state of Auto.mapRoutine while it's running.
+type mapRoutineState struct {
+	c  *Auto
+	bo *backoff.Backoff
+}
+
+func (mrs mapRoutineState) UpdateFullNetmap(nm *netmap.NetworkMap) {
+	c := mrs.c
+	health.SetInPollNetMap(true)
+
+	c.mu.Lock()
+	ctx := c.mapCtx
+	c.synced = true
+	if c.loggedIn {
+		c.state = StateSynchronized
+	}
+	c.expiry = ptr.To(nm.Expiry)
+	stillAuthed := c.loggedIn
+	c.logf("[v1] mapRoutine: netmap received: %s", c.state)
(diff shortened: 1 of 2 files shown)


control/controlclient/direct.go
```diff
@@ -1135,7 +1152,7 @@ func (c *Direct) sendMapRequest(ctx context.Context, isStreaming bool, cb func(*
 		c.expiry = &nm.Expiry
 		c.mu.Unlock()
 
-		cb(nm)
+		nu.UpdateFullNetmap(nm)
```
bradfitz, line 1155: @maisem, this is the core of the change: changing a func callback to an interface method call.

control/controlclient/auto.go
```diff
@@ -480,11 +480,42 @@ func (c *Auto) unpausedChanLocked() <-chan struct{} {
...
+	stillAuthed := c.loggedIn
+	c.logf("[v1] mapRoutine: netmap received: %s", c.state)
+	c.mu.Unlock()
+
+	if stillAuthed {
+		c.sendStatus("mapRoutine-got-netmap", nil, "", nm)
+	}
+	// Reset the backoff timer if we got a netmap.
+	mrs.bo.BackOff(ctx, nil)
+}
```
bradfitz, line 509: This is purely movement, out of the func below.

It was closing over `backoff`, though, hence the new type.

control/controlclient/auto.go
```diff
@@ -567,10 +580,10 @@ func (c *Auto) mapRoutine() {
 
 			if err != nil {
```
maisem, line 581: Err is always non-nil at this point.

maisem (approved):

control/controlclient/auto.go
bradfitz: And it was before too-- glad we can actually see that easily now! 😂

PR #8869, opened 2023-08-13
control/controlclient: move lastUpdateGenInformed to tighter scope

bradfitz:
No need to have it on Auto or be behind a mutex; it's only read/written
from a single goroutine. Move it there.

Updates tailscale/corp#5761


bradfitz:
Or did you have plans for that struct field, @maisem? Like metrics or debug handler or ... ?

maisem (approved):

maisem:
Likely lost in refactors

PR #8872, opened 2023-08-14
wgengine/magicsock: only accept pong sent by CLI ping

valscale:
Looking for suggestions on how to best restructure the code to deal with this. Currently the ping code has a number of special cases to handle discovery ping, CLI ping, size ping, and heartbeat ping that end up having the wrong behavior. It looks like we may need to do a little code reorganization to make this work

What this code does is fix a bug where CLI ping would accept any pong - including heartbeat pongs - as a reply to its ping. This is a problem because CLI ping cannot be used to probe MTU, since it will accept any heartbeat pong as a reply. It does it by pushing the code that fills out the pending CLI ping list to the point that we specify the tx id for the ping so that we can record it in the pending CLI ping entry.

The problem is: The code currently switches from DERP over to UDP-only once it gets any direct UDP connection, even if the connection can't handle the size of ping. This is due to a special case in the code to prevent flip-flopping between endpoints on CLI ping. This results in three successful pings from DERP, followed by timeouts for subsequent pings in the case that the path can't handle this size of packet. Ideally it would instead continue to ping all paths until it got a direct connection that can handle this size of ping, just like CLI pings with no size attached.

### DIFF
--- a/wgengine/magicsock/endpoint.go
+++ b/wgengine/magicsock/endpoint.go
@@ -83,6 +83,7 @@ type endpoint struct {
 }
 
 type pendingCLIPing struct {
+	tx  stun.TxID
 	res *ipnstate.PingResult
 	cb  func(*ipnstate.PingResult)
 }
@@ -363,7 +364,7 @@ func (de *endpoint) heartbeat() {
 	udpAddr, _, _ := de.addrForSendLocked(now)
 	if udpAddr.IsValid() {
 		// We have a preferred path. Ping that every 2 seconds.
-		de.startDiscoPingLocked(udpAddr, now, pingHeartbeat, 0)
+		de.startDiscoPingLocked(udpAddr, now, pingHeartbeat, 0, nil, nil)
 	}
 
 	if de.wantFullPingLocked(now) {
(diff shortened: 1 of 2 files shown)


valscale:
It looks like one approach would be to expand the sentPing struct to include all the state related to a sent ping of any type, including the CLI ping callback and result, as well as the size of the ping and the type.

valscale:
The behavior change would be that we no longer accept any pong whatsoever as a reply to an ordinary (no size) CLI ping.

wgengine/magicsock/endpoint.go
```diff
@@ -415,22 +416,28 @@ func (de *endpoint) cliPing(res *ipnstate.PingResult, size int, cb func(*ipnstat
 		return
 	}
 
-	de.pendingCLIPings = append(de.pendingCLIPings, pendingCLIPing{res, cb})
```
knyar, line 418: Instead of moving this to startDiscoPingLocked, perhaps startDiscoPingLocked should just return txid that this function will put into pendingCLIPing?

wgengine/magicsock/endpoint.go
```diff
@@ -914,9 +925,16 @@ func (de *endpoint) handlePongConnLocked(m *disco.Pong, di *discoInfo, src netip
...
 	for _, pp := range de.pendingCLIPings {
-		de.c.populateCLIPingResponseLocked(pp.res, latency, sp.to)
-		go pp.cb(pp.res)
+		if m.TxID == pp.tx {
+			de.c.logf("\n\n\nGOT CLI PONG sent tx %v received tx %v\n\n\n", pp.tx, m.TxID)
+			de.c.populateCLIPingResponseLocked(pp.res, latency, sp.to)
+			go pp.cb(pp.res)
+		}
 	}
 	de.pendingCLIPings = nil
```
knyar, line 939: Now when you're not fully draining `pendingCLIPings` perhaps you have to keep the unprocessed items around for the next time?

wgengine/magicsock/endpoint.go
```diff
@@ -591,15 +607,10 @@ func (de *endpoint) startDiscoPingLocked(ep netip.AddrPort, now mono.Time, purpo
 			return
```
bradfitz, line 607: this exit path means there's a stranded entry in the sentPing map that was never sent and will never be cleaned up

PR #8877, opened 2023-08-14
Draft ping size fixes

valscale:
@knyar, as we discussed, these are the three commits necessary to make ping -s work like regular ping: continue to ping until it finds a direct path for a ping of this size.

The first commit folds CLI ping tracking into normal ping tracking and matches the pongs with the pings using tx ids.

The second commit splits up the concept of MTU into wire and TUN MTU and introduces several utility functions to convert between them. It also adds a list of MTUs to probe (but does not probe them yet).

The third commit records the MTU information from a CLI ping and continues pinging on all endpoints if it has not found a non-DERP path that can accommodate the whole ping.

I'd appreciate your suggestions at any level from architectural redesign to typos. :)


wgengine/magicsock/endpoint.go
```diff
@@ -951,19 +988,23 @@ func (de *endpoint) handlePongConnLocked(m *disco.Pong, di *discoInfo, src netip
...
+type addrQuality struct {
 	netip.AddrPort
 	latency time.Duration
+	wireMTU tstun.WireMTU
 }
 
-func (a addrLatency) String() string {
-	return a.AddrPort.String() + "@" + a.latency.String()
+func (a addrQuality) String() string {
+	return a.AddrPort.String() + "@" + a.latency.String() + "+" + strconv.Itoa(int(a.wireMTU))
```
knyar, line 999: I wonder if `fmt.Sprintf` would make this more readable?

wgengine/netstack/netstack.go
```diff
@@ -147,10 +147,10 @@ type Impl struct {
...
 const nicID = 1
 
-// maxUDPPacketSize is the maximum size of a UDP packet we copy in startPacketCopy
-// when relaying UDP packets. We don't use the 'mtu' const in anticipation of
-// one day making the MTU more dynamic.
-const maxUDPPacketSize = 1500
+// maxUDPPacketSize is the maximum size of a UDP packet we copy in
+// startPacketCopy when relaying UDP packets. We support arbitrarily
+// large MTUs so this is now the maximum IP packet size.
+const maxUDPPacketSize = packet.MaxIPPacketSize
```
knyar, line 153: I don't fully understand how hot `startPacketCopy` is, but it allocates a buffer of this size, which is now 43 times larger. It seems like we can safely use `MaxProbedWireMTU` instead?

wgengine/magicsock/endpoint.go
```diff
@@ -907,25 +947,24 @@ func (de *endpoint) handlePongConnLocked(m *disco.Pong, di *discoInfo, src netip
 	}
 
 	if sp.purpose != pingHeartbeat {
-		de.c.dlogf("[v1] magicsock: disco: %v<-%v (%v, %v)  got pong tx=%x latency=%v pong.src=%v%v", de.c.discoShort, de.discoShort(), de.publicKey.ShortString(), src, m.TxID[:6], latency.Round(time.Millisecond), m.Src, logger.ArgWriter(func(bw *bufio.Writer) {
+		de.c.dlogf("[v1] magicsock: disco: %v<-%v (%v, %v)  got pong tx=%x latency=%v mtu=%v pong.src=%v%v", de.c.discoShort, de.discoShort(), de.publicKey.ShortString(), src, m.TxID[:6], latency.Round(time.Millisecond), pingSizeToWireMTU(sp.size, sp.to.Addr()), m.Src, logger.ArgWriter(func(bw *bufio.Writer) {
```
knyar, line 950: Seems a bit weird to log the MTU computed from the ping size rather than the ping size itself. We know that a ping of that size got through, but we don't necessarily know that a larger packet would be dropped, right? I'd suggest to log the size itself.

wgengine/magicsock/endpoint.go
```diff
@@ -415,22 +427,30 @@ func (de *endpoint) cliPing(res *ipnstate.PingResult, size int, cb func(*ipnstat
...
+		return
+	}
 
 	now := mono.Now()
-	udpAddr, derpAddr, _ := de.addrForSendLocked(now)
+	udpAddr, derpAddr, _ := de.addrForSendLocked(now, size)
+
 	if derpAddr.IsValid() {
-		de.startDiscoPingLocked(derpAddr, now, pingCLI, size)
+		de.c.logf("\nDERP ping\n")
```
knyar, line 440: I assume you intend to remove these messages eventually, but if you think they are going to be useful longer term, perhaps prefix them with `[v2]` and add a consistent prefix (e.g. `cliPing: DERP`, `cliPing: single UDP`, etc) that could be grepped for in the future?

wgengine/magicsock/endpoint.go
```diff
@@ -914,11 +911,10 @@ func (de *endpoint) handlePongConnLocked(m *disco.Pong, di *discoInfo, src netip
 		}))
 	}
 
-	for _, pp := range de.pendingCLIPings {
-		de.c.populateCLIPingResponseLocked(pp.res, latency, sp.to)
-		go pp.cb(pp.res)
+	if sp.purpose == pingCLI {
```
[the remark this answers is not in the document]
valscale: Yes, I like that idea, will do.

net/tstun/mtu.go
```diff
@@ -1,33 +1,161 @@
...
-	if mtu, ok := envknob.LookupUintSized("TS_DEBUG_MTU", 10, 32); ok {
-		mtu := uint32(mtu)
-		if mtu > maxMTU {
-			mtu = maxMTU
+// WireMTUsToProbe are the MTUs we are most likely to see in real world
+// networks. These are used to send discovery pings in packets padded
+// to exactly match the goal MTU.
+var WireMTUsToProbe = [...]WireMTU{
+	576,  // Smallest MTU for IPv4, probably useless?
+	1124, // An observed max MTU in the wild, maybe 1100 instead?
```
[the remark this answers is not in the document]
valscale: ha ha terminology! the proper term is just "MTU" as in Maximum Transmission Unit. I will fix.

wgengine/magicsock/endpoint.go
```diff
@@ -250,15 +255,27 @@ func (de *endpoint) DstToBytes() []byte  { return packIPPort(de.fakeWGAddr) }
...
 
-	if udpAddr.IsValid() && !now.After(de.trustBestAddrUntil) {
+	mtuOk := true
+	if size != 0 {
+		pathMTU := de.bestAddr.wireMTU
+		requestedMTU := pingSizeToWireMTU(size, udpAddr.Addr())
+		if requestedMTU > pathMTU {
+			mtuOk = false
+		}
+		de.c.logf("pathMTU %v requestedMTU %v", pathMTU, requestedMTU)
```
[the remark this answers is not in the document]
valscale: Probably yes! Let me look at turning my temporary debug logging into things that will be helpful long term.

net/tstun/mtu.go
```diff
@@ -1,33 +1,161 @@
...
 // SPDX-License-Identifier: BSD-3-Clause
+
 package tstun
 
-import "tailscale.com/envknob"
+import (
+	"tailscale.com/envknob"
+)
+
+// We now have several different kinds of MTU:
```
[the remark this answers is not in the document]
valscale: I would appreciate more feedback on this. There are actually three kinds of MTUs: tailscale TUN MTU, underlying network interface MTU (which could be physical or virtual), and peer MTU, and I am not sure how to explain or name them. Clearly it needs another pass!

disco/disco.go
```diff
@@ -46,6 +47,10 @@ const (
 	TypeCallMeMaybe = MessageType(0x03)
 )
 
+// MaxDiscoPingSize is the largest useful ping message size that we
+// can send - a single IP payload.
+const MaxDiscoPingSize = packet.MaxIPPacketSize
```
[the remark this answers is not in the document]
valscale: Correct, will fix!

wgengine/magicsock/endpoint.go
valscale: Will do.

wgengine/netstack/netstack.go
valscale: Thanks, I didn't like this either. I think what we actually want is the maximum of MaxProbedWireMTU and DefaultTunMTU + wg overhead (due to user setting of MTU via ifconfig or TS_DEBUG_MTU).

wgengine/magicsock/endpoint.go
valscale: Yeah, that shouldn't be "mtu=" it should be something like "pkt len=". We do want to know the size of the packet that traversed the path - the ping size is just the ping message itself, and reading the log message we don't know whether it had IPv6 or IPv4 headers.

wgengine/magicsock/endpoint.go
valscale: Yeah they are just debugging, will remove! Thanks.

wgengine/magicsock/endpoint.go
```diff
@@ -250,15 +255,27 @@ func (de *endpoint) DstToBytes() []byte  { return packIPPort(de.fakeWGAddr) }
...
 // addrForSendLocked returns the address(es) that should be used for
 // sending the next packet. Zero, one, or both of UDP address and DERP
-// addr may be non-zero. If the endpoint is WireGuard only and does not have
-// latency information, a bool is returned to indiciate that the
-// WireGuard latency discovery pings should be sent.
+// addr may be non-zero. If the endpoint is WireGuard only and does
+// not have latency information, a bool is returned to indiciate that
+// the WireGuard latency discovery pings should be sent. If size is
+// non-zero, then it must return at least one addr with a large enough
+// path MTU to permit a ping payload of size bytes to be delivered.
```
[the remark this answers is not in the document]
valscale: I added a new function, addrForPingLocked(), following the example of addrForWireGuardSendLocked(), and also added a bunch of tests for it.

net/packet/packet.go
```diff
@@ -34,6 +34,12 @@ const (
 	TCPECNBits TCPFlag = TCPECNEcho | TCPCWR
 )
 
+// Since we support arbitrarily large MTUs, export a constant for use
+// by any code that is handling IP packets of unknown size. See
+// https://pkg.go.dev/gvisor.dev/gvisor/pkg/tcpip/header#pkg-constants
+// for discussion of why this is 65536 and not some smaller value.
+const MaxIPPacketSize = 65536
```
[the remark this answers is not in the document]
valscale: Turns out MaxPacketSize is defined properly in net/tstun/wrap.go, using that.

net/tstun/mtu.go
```diff
@@ -1,33 +1,161 @@
...
+//
+// Current MTU: This the MTU of the tailscale TUN at any given moment
+// after TUN creation. In order of priority, it is:
+//
+// 1. The MTU set by the user via the OS, if it has ever been set
+// 2. If TS_DEBUG_PMTUD is set, the maximum size MTU we probe, minus wg overhead
+// 4. If TS_DEBUG_PMTUD is not set, the Safe MTU
+
+type TunMTU uint32  // MTU for a Tailscale TUN
+type WireMTU uint32 // MTU for underlying network devices
```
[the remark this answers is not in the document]
valscale: Fixed, thanks!

wgengine/netstack/netstack.go
valscale: Oof. I still don't fully understand this code, but I do believe that that the user can configure the tailscale MTU up to tstun.MaxPacketSize and we have to be able to handle that.

Issue #8878, opened 2023-08-14
fallback to the iptables runner when iptables is present but can't be queried, and nftables is unavailable


raggi:
The code path prior to the firewall heuristics would fall back to iptables if the iptables binary is present and responds to `--version` even if it can't program rules. We need to preserve this behavior for now, as it is presently depended upon by some distributions.

PR #8879, opened 2023-08-14
wgengine/router: use iptablesRunner when no firewall tool is available:

KevinLiang10:
### Change:
The current router errors out when neither iptables nor nftables support is present. We should fall back to the previous behaviour which we creates a dummy iptablesRunner. Creating iptablesRunner errors only if the iptables command is missing / doesn’t support --version, as long as it can determine a version (it doesn’t have to work) then it’ll carry on.

Fixes: #8878


wgengine/router/router_linux.go
```diff
@@ -92,40 +92,38 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) (linuxfw.FirewallMo
...
 		// neither/both are currently used, use nftables.
 		logf("router: nftables is available")
-		return linuxfw.FirewallModeNfTables, nil
+		return linuxfw.FirewallModeNfTables
 	case iptAva:
 		logf("router: iptables is available")
-		return linuxfw.FirewallModeIPTables, nil
+		return linuxfw.FirewallModeIPTables
 	default:
 		// if neither iptables nor nftables are available,
```
raggi, line 114: Please update this comment to describe what this behavior is and why we have this behavior (essentially what you have in the commit message)

raggi (approved):
LGTM. Please update the doc string, it'll help the next reader to spot what we missed!

PR #8880, opened 2023-08-14
client/tailscale: add LocalClient.IncrementMetric func

soniaappasamy:
A #cleanup to add a func to utilize the already-present "/localapi/v0/upload-client-metrics" localapi endpoint.

### DIFF
--- a/client/tailscale/localclient.go
+++ b/client/tailscale/localclient.go
@@ -259,6 +259,29 @@ func (lc *LocalClient) DaemonMetrics(ctx context.Context) ([]byte, error) {
 	return lc.get200(ctx, "/localapi/v0/metrics")
 }
 
+// IncrementMetric increments the value of a Tailscale daemon's metric by
+// the given delta. If the metric has yet to exist, a new counter metric is
+// created and initialized to delta.
+//
+// IncrementMetric only supports counter metrics and non-negative delta values.
+// Gauge metrics are unsupported.
+func (lc *LocalClient) IncrementMetric(ctx context.Context, name string, delta int) error {
+	type metricUpdate struct {
+		Name  string `json:"name"`
+		Type  string `json:"type"`
+		Value int    `json:"value"` // amount to increment by
+	}
+	if delta < 0 {
+		return errors.New("negative delta not allowed")
+	}
+	_, err := lc.send(ctx, "POST", "/localapi/v0/upload-client-metrics", 200, jsonBody([]metricUpdate{{
+		Name:  name,
+		Type:  "counter",
+		Value: delta,
+	}}))
+	return err
+}
+
 // TailDaemonLogs returns a stream the Tailscale daemon's logs as they arrive.
 // Close the context to stop the stream.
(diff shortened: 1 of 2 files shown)


client/tailscale/localclient.go
```diff
@@ -259,6 +259,22 @@ func (lc *LocalClient) DaemonMetrics(ctx context.Context) ([]byte, error) {
 	return lc.get200(ctx, "/localapi/v0/metrics")
 }
 
+// ClientMetric is the request data for UploadClientMetric.
+type ClientMetric struct {
+	Name string `json:"name"`
+	// One of "counter" or "gauge"
```
bradfitz, line 265: // Type is ...

And say that it's only used if the metric named by Name doesn't already exist?

And document the others.

client/tailscale/localclient.go
```diff
@@ -259,6 +259,22 @@ func (lc *LocalClient) DaemonMetrics(ctx context.Context) ([]byte, error) {
...
+	Type  string `json:"type"`
+	Value int    `json:"value"`
```
bradfitz, line 267: Is this int or int64 internally?

client/tailscale/localclient.go
```diff
@@ -259,6 +259,22 @@ func (lc *LocalClient) DaemonMetrics(ctx context.Context) ([]byte, error) {
...
+}
+
+// UploadClientMetrics posts metric to the Tailscale daemon.
+// If any metric doesn't exist yet, the daemon creates a new one with the given
+// name and initial value.
+func (lc *LocalClient) UploadClientMetrics(ctx context.Context, metrics []ClientMetric) error {
```
bradfitz, line 273: The slice should only contain one item per Name, right?

soniaappasamy: Based on the localapi implementation, it doesn't really have to - if included multiple times, the metric value will be increased both of those times.

client/tailscale/localclient.go
```diff
@@ -259,6 +259,23 @@ func (lc *LocalClient) DaemonMetrics(ctx context.Context) ([]byte, error) {
...
+// IncrementMetric increments the value of a Tailscale daemon's metric by
+// the given delta. If the metric has yet to exist, a new counter metric is
+// created and initialized to delta.
+func (lc *LocalClient) IncrementMetric(ctx context.Context, name string, delta int) error {
+	type metricUpdate struct {
+		Name  string `json:"name"`
+		Type  string `json:"type"`
+		Value int    `json:"value"` // amount to increment by
+	}
+	_, err := lc.send(ctx, "POST", "/localapi/v0/upload-client-metrics", 200, jsonBody([]metricUpdate{{
```
bradfitz, line 271: return an error if delta is negative? Since we document this is only for counters.

(If we need gauges later, we can add a new method)

bradfitz (approved):

client/tailscale/localclient.go
bradfitz: I suppose could also rename this `IncrementCounter` if preferred, to leave open the obvious name for `IncrementGauge` later.

PR #8891, opened 2023-08-15
wgengine/magicsock: only cache N most recent endpoints per-Addr

andrew-d:
If a node is flapping or otherwise generating lots of STUN endpoints, we can end up caching a ton of useless values and sending them to peers. Instead, let's apply a fixed per-Addr limit of endpoints that we cache, so that we're only sending peers up to the N most recent.

Updates tailscale/corp#13890

Change-Id: I8079a05b44220c46da55016c0e5fc96dd2135ef8


andrew-d:
This turned out to be a bit more of a rewrite than expected, so feedback appreciated. There's also one TODO in the code about efficiency that I'd appreciate 👀 on.

wgengine/magicsock/magicsock.go
```diff
@@ -2686,16 +2694,50 @@ type discoInfo struct {
...
+	*eh = old[0 : n-1]
+	return item
 }
 
 type endpointTracker struct {
-	mu    sync.Mutex
-	cache map[netip.AddrPort]endpointTrackerEntry
+	mu        sync.Mutex
+	endpoints map[netip.Addr]endpointHeap
 }
```
bradfitz, line 2738: while you're rewriting this anyway, now would be a good time to move this all to its own file in this directory.

also, can you get some more (any) docs on these three types?

Looks good at first glance but I'm reviewing it more closely now.

wgengine/magicsock/magicsock.go
```diff
@@ -2704,55 +2746,126 @@ func (et *endpointTracker) update(now time.Time, eps []tailcfg.Endpoint) (epsPlu
...
+			return v.endpoint == ep
+		})
+		if !found {
+			heap.Push(&epHeap, &endpointTrackerEntry{endpoint: ep, until: until})
+		}
 	} else {
-		entry = endpointTrackerEntry{ep, until}
+		// Create a new heap with this exact address
+		heap.Push(&epHeap, &endpointTrackerEntry{endpoint: ep, until: until})
 	}
```
bradfitz, line 2840: not sure `found` + separate code paths really matters here. `ContainsFunc` on an empty slice will quickly report false. Could collapse this all a bit

wgengine/magicsock/magicsock.go
```diff
@@ -2704,55 +2746,126 @@ func (et *endpointTracker) update(now time.Time, eps []tailcfg.Endpoint) (epsPlu
...
+		// in the past.
+		for len(epHeap) > 0 {
+			minElem := epHeap[0]
+			if now.After(minElem.until) {
+				heap.Pop(&epHeap)
+			} else {
+				break
+			}
 		}
+		et.endpoints[k] = epHeap
```
bradfitz, line 2868: if it's now len 0, do you want to delete from the map?

wgengine/magicsock/magicsock.go
andrew-d: Added docs, moved to a new file, moved tests to a new file as well

wgengine/magicsock/magicsock.go
andrew-d: Yep, that seems to work 👍

wgengine/magicsock/magicsock.go
andrew-d: Side benefit of pulling this out to a new file: I confirmed that there's 100% test coverage, other than the "defense in depth" / "should never happen" comment 🎉

wgengine/magicsock/endpoint_tracker.go
```diff
@@ -0,0 +1,226 @@
...
+	// endpointTrackerLifetime is how long we continue advertising an
+	// endpoint after we last see it. This is intentionally chosen to be
+	// slightly longer than a full netcheck period.
+	endpointTrackerLifetime = 5*time.Minute + 10*time.Second
+
+	// endpointTrackerMaxPerAddr is how many cached addresses we track for
+	// a given netip.Addr. This allows e.g. restricting the number of STUN
+	// endpoints we cache (which usually have the same netip.Addr but
+	// different ports).
+	endpointTrackerMaxPerAddr = 10
```
bradfitz, line 28: seems pretty overkill. why not like ... 2?

wgengine/magicsock/endpoint_tracker.go
```diff
@@ -0,0 +1,226 @@
...
+)
+
+// endpointTrackerEntry is an entry in an endpointHeap that stores the state of
+// a given cached endpoint.
+type endpointTrackerEntry struct {
+	// endpoint is the cached endpoint
```
bradfitz, line 34: periods on all three of these sentences.

wgengine/magicsock/endpoint_tracker.go
```diff
@@ -0,0 +1,226 @@
...
+	endpoint tailcfg.Endpoint
+	// until is the time until which this endpoint is being cached
+	until time.Time
+	// index is the index within the containing endpointHeap
+	index int
+}
+
+// endpointHeap is an ordered heap of endpointTrackerEntry structs, ordered in
+// ascending order by the expiry time (i.e. oldest first).
```
bradfitz, line 43: by the 'until' expiry time

wgengine/magicsock/endpoint_tracker.go
```diff
@@ -0,0 +1,226 @@
...
+	old := *eh
+	n := len(old)
+	item := old[n-1]
+	old[n-1] = nil  // avoid memory leak
+	item.index = -1 // for safety
+	*eh = old[0 : n-1]
+	return item
+}
+
+// endpointTracker caches endpoints that are advertised to peers
```
bradfitz, line 84: periods on all these.

Maybe include a link to more context? Is there an older bug explaining why we do this?

Why is any caching even needed?

wgengine/magicsock/endpoint_tracker.go
```diff
@@ -0,0 +1,226 @@
...
+				// it anyway.
+				continue
+			}
+
+			// We haven't seen this endpoint; add to the return array
+			epsPlusCached = append(epsPlusCached, ep.endpoint)
+		}
+	}
+
+	// Remove the oldest items from the heap
```
bradfitz, line 140: old comment?

wgengine/magicsock/endpoint_tracker.go
andrew-d: Added a bunch of comments; PTAL?

wgengine/magicsock/endpoint_tracker.go
```diff
@@ -0,0 +1,231 @@
...
+	// bound the number of items in the heap.
+	//
+	// TODO(andrew): this means we iterate over the entire heap once per
+	// endpoint; even if the heap is small, if we have a lot of input
+	// endpoints this can be expensive?
+	for i, entry := range epHeap {
+		if entry.endpoint == ep {
+			entry.until = until
+			heap.Fix(&epHeap, i)
+			mak.Set(&et.endpoints, key, epHeap)
```
bradfitz, line 170: did you consider storing a pointer to the heap slice in the map to avoid having to re-assign the slice each time?

actually, in this case this line shouldn't matter: the heap isn't changing in size so the slice is identical. you just rearranged its memory.

wgengine/magicsock/endpoint_tracker.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// Now that we've added everything, pop from our heap until we're below
+	// the limit. This is a min-heap, so popping removes the lowest (and
+	// thus oldest) endpoint.
+	for epHeap.Len() > endpointTrackerMaxPerAddr {
+		heap.Pop(&epHeap)
+	}
+
+	// Store back in map
+	mak.Set(&et.endpoints, key, epHeap)
```
bradfitz, line 205: but in this case you'd benefit from storing the the pointer in the map value

wgengine/magicsock/endpoint_tracker.go
andrew-d: Dropped to 6 and left a comment explaining why; PTAL?

wgengine/magicsock/endpoint_tracker.go
```diff
@@ -0,0 +1,240 @@
...
+	//
+	// The value of 6 is chosen because we can advertise up to 3 endpoints
+	// based on the STUN IP:
+	//    1. The STUN endpoint itself (EndpointSTUN)
+	//    2. The STUN IP with the local Tailscale port (EndpointSTUN4LocalPort)
+	//    3. The STUN IP with a portmapped port (EndpointPortmapped)
+	//
+	// Storing 6 endpoints in the cache means we can store up to 2 previous
+	// sets of endpoints.
+	endpointTrackerMaxPerAddr = 6
```
bradfitz, line 37: thanks. great comment.

wgengine/magicsock/endpoint_tracker.go
andrew-d: I took a look at this and, while not really changing LoC, it means it'll be harder to forget to assign `epHeap` back to the map in the future 👍 Mind taking a quick look one more time?

bradfitz (approved):

wgengine/magicsock/endpoint_tracker.go
```diff
@@ -0,0 +1,246 @@
...
+	old[n-1] = nil  // avoid memory leak
+	item.index = -1 // for safety
+	*eh = old[0 : n-1]
+	return item
+}
+
+// Min returns a pointer to the minimum element in the heap, without removing
+// it. Since this is a min-heap ordered by the 'until' field, this returns the
+// chronologically "earliest" element in the heap.
+func (eh endpointHeap) Min() *endpointTrackerEntry {
```
bradfitz, line 96: Maybe pointer receiver for consistency? And document the invariant that Len must be non-zero?

PR #8895, opened 2023-08-15
wgengine/router: always use nftables mode on gokrazy

bradfitz:
Updates gokrazy/gokrazy#209


KevinLiang10 (approved):
I'm not familiar with GoKrazy, so I took a look at it and it seems like there is Netlink API supported so I think using nftables would make sense in this case.

PR #8901, opened 2023-08-15
client/web: hook up data fetching to fill --dev React UI

soniaappasamy:
This hooks up the GET endpoint. POST to come in a follow-up.

Updates tailscale/corp#13775


client/web/src/hooks/node-data.ts
```diff
@@ -20,6 +16,12 @@ export type NodeData = {
...
 
+export type UserProfile = {
+  LoginName: string
+  DisplayName: string
+  ProfilePicURL: string
+}
+
 // testData is static set of nodedata used during development.
 // This can be removed once we have a real node data API.
 const testData: NodeData = {
```
willnorris, line 27: I think we can probably remove the testData?

client/web/web.go
```diff
@@ -411,6 +374,16 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		data.IP = st.TailscaleIPs[0].String()
 	}
 
+	if asJSON {
+		if err := json.NewEncoder(w).Encode(data); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
```
willnorris, line 383: this is getting logged in output:

```
superfluous response.WriteHeader call from tailscale.com/client/web.(*Server).serveGetNodeData (web.go:383)
```

I think at this point, a 200 response header has already been written

client/web/web.go
```diff
@@ -309,80 +300,52 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	TUNMode           bool
+	IsSynology        bool
+	DSMVersion        int // 6 or 7, if IsSynology=true
+	IsUnraid          bool
+	UnraidToken       string
+	IPNVersion        string
+}
 
-	st, err := s.lc.Status(ctx)
+func (s *Server) serveGetNodeData(w http.ResponseWriter, r *http.Request, user string, asJSON bool) {
```
willnorris, line 333: I think overloading this method to handle both the old static index page, as well as the JSON raw data output for the new page is somewhat confusing.  

Maybe there is a single `getNodeData()` method that returns a `nodeData` struct.  And then separate handler methods for the old go-template index page and the new `/api/data` JSON endpoint, each of which call `getNodeData()` ?

willnorris (approved):
a couple of inline comments, otherwise lgtm

client/web/src/hooks/node-data.ts
soniaappasamy: whoop! yes!

client/web/web.go
soniaappasamy: Figured we'd be getting rid of it once we cut out the old html templates, but I like your suggestion - a little cleaner in the interim.

PR #8902, opened 2023-08-15
tailcfg: Add FirewallMode to NetInfo to provide metrics.

KevinLiang10:
To record wether user is using iptables or nftables after we add support to nftables on linux, we are adding a field FirewallMode to NetInfo in HostInfo to reflect what firewall mode the host is running, and form metrics. The information is gained from a global constant in hostinfo.go. We set it when selection heuristic made the decision, and magicsock reports this to control.

Updates: tailscale/corp#13943

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -734,6 +734,11 @@ type NetInfo struct {
 	// the control plane.
 	DERPLatency map[string]float64 `json:",omitempty"`
 
+	// FirewallMode is the current firewall utility in use by router (iptables, nftables).
+	// FirewallMode ipt means iptables, nft means nftables. When it's empty user is not using
+	// our netfilter runners to manage firewall rules.
+	FirewallMode string `json:",omitempty"`
+
 	// Update BasicallyEqual when adding fields.
 }
 
@@ -741,10 +746,10 @@ func (ni *NetInfo) String() string {
 	if ni == nil {
 		return "NetInfo(nil)"
 	}
-	return fmt.Sprintf("NetInfo{varies=%v hairpin=%v ipv6=%v ipv6os=%v udp=%v icmpv4=%v derp=#%v portmap=%v link=%q}",
+	return fmt.Sprintf("NetInfo{varies=%v hairpin=%v ipv6=%v ipv6os=%v udp=%v icmpv4=%v derp=#%v portmap=%v link=%q firewallmode=%q}",
 		ni.MappingVariesByDestIP, ni.HairPinning, ni.WorkingIPv6,
 		ni.OSHasIPv6, ni.WorkingUDP, ni.WorkingICMPv4,
-		ni.PreferredDERP, ni.portMapSummary(), ni.LinkType)
+		ni.PreferredDERP, ni.portMapSummary(), ni.LinkType, ni.FirewallMode)
 }
 
 func (ni *NetInfo) portMapSummary() string {
@@ -792,7 +797,8 @@ func (ni *NetInfo) BasicallyEqual(ni2 *NetInfo) bool {
 		ni.PMP == ni2.PMP &&
 		ni.PCP == ni2.PCP &&
 		ni.PreferredDERP == ni2.PreferredDERP &&
-		ni.LinkType == ni2.LinkType
+		ni.LinkType == ni2.LinkType &&
+		ni.FirewallMode == ni2.FirewallMode
 }
 
 // Equal reports whether h and h2 are equal.
--- a/wgengine/router/router_linux.go
+++ b/wgengine/router/router_linux.go
@@ -22,6 +22,7 @@ import (
 	"golang.org/x/sys/unix"
 	"golang.org/x/time/rate"
 	"tailscale.com/envknob"
+	"tailscale.com/hostinfo"
 	"tailscale.com/net/netmon"
 	"tailscale.com/types/logger"
 	"tailscale.com/types/preftype"
@@ -97,29 +98,36 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
 	case envknob.String("TS_DEBUG_FIREWALL_MODE") == "nftables":
 		// TODO(KevinLiang10): Updates to a flag
 		logf("router: envknob TS_DEBUG_FIREWALL_MODE=nftables set")
+		hostinfo.SetFirewallMode("nft-forced")
 		return linuxfw.FirewallModeNfTables
 	case envknob.String("TS_DEBUG_FIREWALL_MODE") == "iptables":
 		logf("router: envknob TS_DEBUG_FIREWALL_MODE=iptables set")
+		hostinfo.SetFirewallMode("ipt-forced")
 		return linuxfw.FirewallModeIPTables
 	case nftRuleCount > 0 && iptRuleCount == 0:
 		logf("router: nftables is currently in use")
+		hostinfo.SetFirewallMode("nft-inuse")
 		return linuxfw.FirewallModeNfTables
 	case iptRuleCount > 0 && nftRuleCount == 0:
 		logf("router: iptables is currently in use")
+		hostinfo.SetFirewallMode("ipt-inuse")
 		return linuxfw.FirewallModeIPTables
 	case nftAva:
 		// if both iptables and nftables are available but
 		// neither/both are currently used, use nftables.
 		logf("router: nftables is available")
+		hostinfo.SetFirewallMode("nft")
 		return linuxfw.FirewallModeNfTables
 	case iptAva:
 		logf("router: iptables is available")
+		hostinfo.SetFirewallMode("ipt")
 		return linuxfw.FirewallModeIPTables
 	default:
 		// if neither iptables nor nftables are available, use iptablesRunner as a dummy
 		// runner which exists but won't do anything. Creating iptablesRunner errors only
 		// if the iptables command is missing or doesn’t support "--version", as long as it
 		// can determine a version then it’ll carry on.
+		hostinfo.SetFirewallMode("ipt-fb")
 		return linuxfw.FirewallModeIPTables
 	}
 }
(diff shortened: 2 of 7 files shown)


wgengine/router/router_linux.go
```diff
@@ -129,6 +130,12 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
 func newNetfilterRunner(logf logger.Logf) (netfilterRunner, error) {
 	tableDetector := &linuxFWDetector{}
 	mode := chooseFireWallMode(logf, tableDetector)
+	if mode == linuxfw.FirewallModeIPTables {
```
maisem, line 133: Did you mean nftables?

KevinLiang10: I thought nftables user would be more than iptables due to our heuristic preferring nftables. Do you believe otherwise?

maisem: Right, let's save bytes in the common case

tailcfg/tailcfg.go
```diff
@@ -734,17 +734,20 @@ type NetInfo struct {
 	// the control plane.
 	DERPLatency map[string]float64 `json:",omitempty"`
 
+	// FirewallMode is the current firewall utility in use by router (iptables, nftables)
```
catzkorn, line 737:
```suggestion
	// FirewallMode is the current firewall utility in use by router (iptables, nftables).
	// If FirewallMode is empty, nftables is being used.
```

maisem: This makes it clearer

tailcfg/tailcfg.go
```diff
@@ -734,17 +734,21 @@ type NetInfo struct {
 	// the control plane.
 	DERPLatency map[string]float64 `json:",omitempty"`
 
+	// FirewallMode is the current firewall utility in use by router (iptables, nftables).
+	// If FirewallMode is empty, nftables is being used.
```
bradfitz, line 738: "nftables" is given as an example, but then says empty means nftables. So both mean nftables? I assume only on Linux.

What about neither? (userspace mode)

tailcfg/tailcfg.go
```diff
@@ -734,17 +734,21 @@ type NetInfo struct {
...
+
 	// Update BasicallyEqual when adding fields.
 }
 
 func (ni *NetInfo) String() string {
 	if ni == nil {
 		return "NetInfo(nil)"
 	}
-	return fmt.Sprintf("NetInfo{varies=%v hairpin=%v ipv6=%v ipv6os=%v udp=%v icmpv4=%v derp=#%v portmap=%v link=%q}",
+	return fmt.Sprintf("NetInfo{varies=%v hairpin=%v ipv6=%v ipv6os=%v udp=%v icmpv4=%v derp=#%v portmap=%v link=%q firewallmode=%q}",
```
bradfitz, line 748: maybe just fw=%q ?

But does empty string here also mean nftables? Or is the empty string meaning nftables only an optimization for the wire?

wgengine/router/router_linux.go
```diff
@@ -129,6 +130,12 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
...
+		hostinfo.SetFirewallMode("iptables")
+	} else {
+		hostinfo.SetFirewallMode("")
```
bradfitz, line 136: let's remove the empty-means-nftables bit. If we want to save some bytes, we can up "ipt" and "nft" here.

bradfitz: The more interesting bit for stats is _why_ we chose whatever path. Maybe we should move these SetFirewallMode calls to chooseFireWallMode before each return statement.

e.g.

     hostinfo.SetFirewallMode("nft-inuse")

tailcfg/tailcfg.go
KevinLiang10: It's the later

wgengine/router/router_linux.go
KevinLiang10: I don't think we are interested in the reason we choose ipt or nft. The reasons is we are always prioritizing nft. In what circumstances would we usually want that bit? Like what's the background we might use that stat?

KevinLiang10: This would also be hard for user or support team to understand what's going on since they don't necessarily understand our prioritization? I don't know but It just seems complicated comparing to other fields in NetInfo

bradfitz: Support team can look at the prefix if they only want the current state. I'm interesting in learning how we transition this over time and what the state of the world looks like (like: how many people are forcing this? what are we failing on?) Also, support could be interested in knowing why a path was taken. Logs can answer that, but so could this.

Or we could add clientmetrics too.

In any case, at least remove the zero value meaning nftables. That's too confusing as implemented with showing the empty string in some log statements to mean nftables when the empty string might also mean neither.

KevinLiang10: Updated, thoughts on the latest?

tailcfg/tailcfg.go
maisem: change to `fw=`?

wgengine/router/router_linux.go
```diff
@@ -97,23 +98,29 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
...
 		return linuxfw.FirewallModeIPTables
 	case nftAva:
 		// if both iptables and nftables are available but
 		// neither/both are currently used, use nftables.
 		logf("router: nftables is available")
+		hostinfo.SetFirewallMode("nft-ava")
 		return linuxfw.FirewallModeNfTables
 	case iptAva:
 		logf("router: iptables is available")
+		hostinfo.SetFirewallMode("ipt-ava")
```
maisem, line 123: ipt-only

wgengine/router/router_linux.go
```diff
@@ -97,23 +98,29 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
...
 		return linuxfw.FirewallModeNfTables
 	case iptRuleCount > 0 && nftRuleCount == 0:
 		logf("router: iptables is currently in use")
+		hostinfo.SetFirewallMode("ipt-inuse")
 		return linuxfw.FirewallModeIPTables
 	case nftAva:
 		// if both iptables and nftables are available but
 		// neither/both are currently used, use nftables.
 		logf("router: nftables is available")
+		hostinfo.SetFirewallMode("nft-ava")
```
maisem, line 119: nft-pref
for preferred

wgengine/router/router_linux.go
```diff
@@ -97,23 +98,29 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
 	case envknob.String("TS_DEBUG_FIREWALL_MODE") == "nftables":
 		// TODO(KevinLiang10): Updates to a flag
 		logf("router: envknob TS_DEBUG_FIREWALL_MODE=nftables set")
+		hostinfo.SetFirewallMode("nft-chosen")
```
maisem, line 101: nft-forced

wgengine/router/router_linux.go
```diff
@@ -97,23 +98,29 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
...
 		return linuxfw.FirewallModeNfTables
 	case envknob.String("TS_DEBUG_FIREWALL_MODE") == "iptables":
 		logf("router: envknob TS_DEBUG_FIREWALL_MODE=iptables set")
+		hostinfo.SetFirewallMode("ipt-chosen")
```
maisem, line 105: ipt-forced

wgengine/router/router_linux.go
```diff
@@ -97,23 +98,29 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
...
 		// neither/both are currently used, use nftables.
 		logf("router: nftables is available")
+		hostinfo.SetFirewallMode("nft-ava")
 		return linuxfw.FirewallModeNfTables
 	case iptAva:
 		logf("router: iptables is available")
+		hostinfo.SetFirewallMode("ipt-ava")
 		return linuxfw.FirewallModeIPTables
 	default:
 		// if neither iptables nor nftables are available, use iptablesRunner as a dummy
```
maisem, line 126: ipt-fb // fallback

wgengine/router/router_linux.go
KevinLiang10: Well we can't tell cuz when we made this decision we don't know if ipt is available right? I think pref is like having the other one but choosing this one?

wgengine/router/router_linux.go
KevinLiang10: Sure

wgengine/router/router_linux.go
maisem: oh i see

maisem: then just say `nft`

wgengine/router/router_linux.go
maisem: and this `ipt`

KevinLiang10: Good Idea

wgengine/router/router_linux.go
KevinLiang10: Yeah makes sense

wgengine/router/router_linux.go
```diff
@@ -97,29 +98,36 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
...
 	case iptAva:
 		logf("router: iptables is available")
+		hostinfo.SetFirewallMode("ipt")
 		return linuxfw.FirewallModeIPTables
 	default:
 		// if neither iptables nor nftables are available, use iptablesRunner as a dummy
 		// runner which exists but won't do anything. Creating iptablesRunner errors only
 		// if the iptables command is missing or doesn’t support "--version", as long as it
 		// can determine a version then it’ll carry on.
+		hostinfo.SetFirewallMode("ipt-fallback")
```
maisem, line 130: ipt-fb?

catzkorn (approved):
Nice! I like the change to show more detail about what the actual state of the world is.

maisem (approved):

wgengine/router/router_linux.go
KevinLiang10: Sure, why not save a few bytes

PR #8906, opened 2023-08-16
util/cache: add package for general-purpose caching

andrew-d:
This package allows caching arbitrary key/value pairs in-memory, along with an interface implemented by the cache types.

Extracted from #7493

Change-Id: Ic8ca820927c456721cf324a0c8f3882a57752cc9

### DIFF
--- a/util/cache/cache_test.go
+++ b/util/cache/cache_test.go
@@ -1,0 +1,199 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package cache
+
+import (
+	"errors"
+	"testing"
+	"time"
+)
+
+var startTime = time.Date(2023, time.March, 1, 0, 0, 0, 0, time.UTC)
+
+func TestSingleCache(t *testing.T) {
+	testTime := startTime
+	timeNow := func() time.Time { return testTime }
+	c := &Single[string, int]{
+		timeNow: timeNow,
+	}
+
+	t.Run("NoServeExpired", func(t *testing.T) {
+		testCacheImpl(t, c, &testTime, false)
+	})
+
+	t.Run("ServeExpired", func(t *testing.T) {
+		c.Empty()
+		c.ServeExpired = true
+		testTime = startTime
+		testCacheImpl(t, c, &testTime, true)
+	})
+}
+
+func TestLocking(t *testing.T) {
+	testTime := startTime
+	timeNow := func() time.Time { return testTime }
+	c := NewLocking(&Single[string, int]{
+		timeNow: timeNow,
+	})
+
+	// Just verify that the inner cache's behaviour hasn't changed.
+	testCacheImpl(t, c, &testTime, false)
+}
+
+func testCacheImpl(t *testing.T, c Cache[string, int], testTime *time.Time, serveExpired bool) {
+	var fillTime time.Time
+	t.Run("InitialFill", func(t *testing.T) {
+		fillTime = testTime.Add(time.Hour)
+		val, err := c.Get("key", func() (int, time.Time, error) {
+			return 123, fillTime, nil
+		})
+		if err != nil {
+			t.Fatal(err)
+		}
+		if val != 123 {
+			t.Fatalf("got val=%d; want 123", val)
+		}
+	})
+
+	// Fetching again won't call our fill function
+	t.Run("SecondFetch", func(t *testing.T) {
+		*testTime = fillTime.Add(-1 * time.Second)
+		called := false
+		val, err := c.Get("key", func() (int, time.Time, error) {
+			called = true
+			return -1, fillTime, nil
+		})
+		if called {
+			t.Fatal("wanted no call to fill function")
+		}
+		if err != nil {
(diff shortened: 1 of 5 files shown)


util/cache/disk.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"encoding/json"
+	"os"
+	"time"
+)
+
+// Disk is a cache that stores data in a file on-disk. It also supports
+// returning a previously-expired value if refreshing the value in the cache
+// fails.
+type Disk[K comparable, V any] struct {
```
bradfitz, line 15: mention V needs to be JSON-able?

bradfitz: and I guess K

util/cache/interface.go
```diff
@@ -0,0 +1,26 @@
...
+type Cache[K comparable, V any] interface {
+	// Get should return a previously-cached value or call the provided
+	// FillFunc to obtain a new one. The provided key can be used either to
+	// allow multiple cached values, or to drop the cache if the key
+	// changes; either is valid.
+	Get(K, FillFunc[V]) (V, error)
+
+	// Forget should empty the cache such that the next call to Get should
+	// call the provided FillFunc.
+	Forget()
```
bradfitz, line 20: Forget the whole cache? Not just a single key?

util/cache/interface.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+// FillFunc is the signature of a function for filling a cache. It should
+// return the value to be cached, the time that the cached value is valid
+// until, or an error
```
bradfitz, line 25: trailing period.

util/cache/interface.go
andrew-d: Split `Forget(K)` from `Empty` and added a test 👍

andrew-d:
(force-pushed to remove the disk cache entirely now since I don't need it for anything yet; it's in Git history if we need it)

util/cache/cache_test.go
```diff
@@ -0,0 +1,188 @@
+// Copyright (c) Tailscale Inc & AUTHORS
```
noncombatant, line 1: 2023

util/cache/interface.go
```diff
@@ -0,0 +1,39 @@
...
+// Cache is the interface for the cache types in this package.
+//
+// Functions in this interface take a key parameter, but it is valid for a
+// cache type to hold a single value associated with a key, and simply drop the
+// cached value if provided with a different key.
+//
+// It is valid for Cache implementations to be concurrency-safe or not, and
+// each implementation should document this.
+//
+// K and V should be types that can be successfully passed to json.Marshal.
```
noncombatant, line 19: Although encoding/json does not (sigh), maybe we can create a `JSONMarshaler` interface (or similar name) interface, and use `V JSONMarshaler | TextMarshaler` instead of `V any`? Basically, it'd be nice if this comment were enforced at compile time.

Might also be good to report the lack of `JSONMarshaler` and the lack of its use upstream and provide a patch?

util/cache/none.go
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package cache
+
+// None provides no caching and always calls the provided FillFunc.
+//
+// It is safe for concurrent use.
```
noncombatant, line 8: Does this depend technically on whether the FillFunc is?

util/cache/single.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"time"
+)
+
+// Single is a simple in-memory cache that stores a single value until a
+// defined time before it is re-fetched. It also supports returning a
+// previously-expired value if refreshing the value in the cache fails.
+//
+// Single is not safe for concurrent use.
```
noncombatant, line 14: Not sure if this is a good idea; see what you think: It might be nice to provide a `type LockingCache[C Cache, ...] struct { ... }` that can lock any cache, and forwards method calls to the underlying `C`, holding a lock as necessary.

Then you could document just once to the effect of, "If you need a concurrency-safe cache, use `LockingCache`..."

util/cache/interface.go
creachadair: Are you referring to something besides https://pkg.go.dev/encoding/json#Marshaler ?

It is not possible to write a static type that describes "marshalability" in general, since the marshaler walks arbitrary structures with reflection. One could eliminate certain obvious classes (e.g., functions, channels), but even then you'd have to be careful not to rule out type definitions with those underlying types.

util/cache/none.go
andrew-d: Yes, good point; will update the comment 👍

util/cache/single.go
andrew-d: Good idea; just added that as an implementation.

util/cache/interface.go
andrew-d: Yeah, my understanding is that the there's no way to express "is safe to pass to `json.Marshal`" as part of the type system, unfortunately. I wish there was!

Issue #7781, opened 2023-04-03
Implement the ART routing table


danderson:
We need a routing table for some upcoming things. This is the tracking bug for implementing ART (https://cseweb.ucsd.edu//~varghese/TEACH/cs228/artlookup.pdf)

PR #8910, opened 2023-08-16
net/art: move child table pointers out of strideEntry

danderson:
In preparation for a different refactor, but incidentally also saves 10-25% memory on overall table size in benchmarks.

Updates #7781

### DIFF
--- a/net/art/stride_table.go
+++ b/net/art/stride_table.go
@@ -29,8 +29,6 @@ type strideEntry[T any] struct {
 	prefixIndex int
 	// value is the value associated with the strideEntry, if any.
 	value *T
-	// child is the child strideTable associated with the strideEntry, if any.
-	child *strideTable[T]
 }
 
 // strideTable is a binary tree that implements an 8-bit routing table.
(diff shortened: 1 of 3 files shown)


net/art/stride_table.go
```diff
@@ -69,61 +69,50 @@ const (
 	lastHostIndex = 0b1_1111_1111
 )
 
-// getChild returns the child strideTable pointer for addr (if any), and an
-// internal array index that can be used with deleteChild.
-func (t *strideTable[T]) getChild(addr uint8) (child *strideTable[T], idx int) {
-	idx = hostIndex(addr)
-	return t.entries[idx].child, idx
+// getChild returns the child strideTable pointer for addr (if any).
```
bradfitz, line 72: (or nil if none)

?

net/art/stride_table.go
```diff
@@ -69,61 +69,50 @@ const (
...
-	idx := hostIndex(addr)
-	if t.entries[idx].child == nil {
-		t.entries[idx].child = &strideTable[T]{
+	if t.children[addr] == nil {
+		t.children[addr] = &strideTable[T]{
 			prefix: childPrefixOf(t.prefix, addr),
 		}
 		t.childRefs++
-		return t.entries[idx].child, true
+		return t.children[addr], true
```
bradfitz, line 99: could use a local variable to avoid the map operation on this return, since you know the value from line 95

not sure if the compiler gets this right these days. at least in the past it did not.

net/art/stride_table.go
```diff
@@ -69,61 +69,50 @@ const (
...
+	return t.entries[hostIndex(addr)].value, t.children[addr]
 }
 
 // findFirstChild returns the first child strideTable in t, or nil if
 // t has no children.
 func (t *strideTable[T]) findFirstChild() *strideTable[T] {
-	for i := firstHostIndex; i <= lastHostIndex; i++ {
-		if child := t.entries[i].child; child != nil {
+	for i := 0; i < len(t.children); i++ {
+		if child := t.children[i]; child != nil {
```
bradfitz, line 115: `for _, child := range t.children {` ?

bradfitz (approved):

net/art/stride_table.go
danderson: Replaced the explicit 256-wise iterations with array ranges.

net/art/stride_table.go
danderson: I've forgotten how to get the compiler to show its workings, so I made the lookup use a local.

net/art/stride_table.go
danderson: Replaced the various (if any)s with better words.

bradfitz (approved):

PR #8913, opened 2023-08-16
release/dist: run yarn build before building CLI

willnorris:
This builds the assets for the new (still in development) web client. These assets will soon be embedded into the cmd/tailscale binary, but are not actually done so yet.

Updates tailscale/corp#13775


release/dist/dist.go
```diff
@@ -315,6 +340,18 @@ func findGo(path string) (string, error) {
 	return toolGo, nil
 }
 
+func findYarn(path string) (string, error) {
```
soniaappasamy, line 343: Pretty similar to `findGo` above. I guess they could both call some shared `findTool` function. No strong opinion.

release/dist/dist.go
```diff
@@ -43,6 +43,8 @@ type Build struct {
 	Tmp string
 	// Go is the path to the Go binary to use for building.
 	Go string
+	// Yarn is the path to the yarn binary to use for building.
+	Yarn string
```
soniaappasamy, line 47: Maybe a small doc that this is only used to build the web client.

soniaappasamy (approved):

release/dist/dist.go
willnorris: done

release/dist/dist.go
willnorris: done

PR #8920, opened 2023-08-16
client/web: add csrf protection to web client api

soniaappasamy:
Adds csrf protection and hooks up an initial POST request from the React web client.

Updates tailscale/corp#13775


client/web/api.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/gorilla/csrf"
+	"tailscale.com/util/httpm"
+)
+
+type api struct {
+	s *Server
+}
+
+// ServeHTTP serves requests for the web client api without CSRF protection.
```
soniaappasamy, line 17:
```suggestion
// ServeHTTP serves requests for the web client api.
```

client/web/web.go
```diff
@@ -272,18 +280,8 @@ req.send(null);
 func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if s.devMode {
 		if r.URL.Path == "/api/data" {
```
soniaappasamy, line 282:
```suggestion
		if r.URL.Path == "/api/" {
```

client/web/web.go
```diff
@@ -272,18 +280,8 @@ req.send(null);
...
-			}
-			switch r.Method {
-			case httpm.GET:
-				s.serveGetNodeDataJSON(w, r, user)
-			case httpm.POST:
-				s.servePostNodeUpdate(w, r)
-			default:
-				http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
-			}
+			// Pass through to other handlers via CSRF protection.
```
willnorris, line 283: same... this `/api/data` can probably be exposed outside of devmode.  Nothing in the current static web client is calling it, but there's really no harm in exposing it... it's the same data we put in the go template anyway.

go.mod
```diff
@@ -208,6 +210,7 @@ require (
 	github.com/gordonklaus/ineffassign v0.0.0-20230107090616-13ace0543b28 // indirect
 	github.com/goreleaser/chglog v0.5.0 // indirect
 	github.com/goreleaser/fileglob v1.3.0 // indirect
+	github.com/gorilla/csrf v1.7.1
```
willnorris, line 213: does this need `go mod tidy` ?  Typically, direct package dependencies are groups separately from the indirect packages.

client/web/src/api.ts
```diff
@@ -0,0 +1,27 @@
...
+// with csrf header handling.
+export function apiFetch(
+  input: RequestInfo | URL,
+  init?: RequestInit | undefined
+): Promise<Response> {
+  return fetch(input, {
+    ...init,
+    headers: withCsrfToken(init?.headers),
+  }).then((r) => {
+    updateCsrfToken(r)
```
willnorris, line 13: if we're only updating the csrf token on api responses, then how is the token set on the initial api request?  I think we need the `csrf.Protect` middleware below on more than just the `/api/` handler... that would allow bootstrapping the csrf token for the first api request.  I'll need to look at the gorilla docs a little more closely to confirm that's how we're supposed to be holding it.

client/web/web.go
```diff
@@ -70,6 +73,11 @@ func NewServer(devMode bool, lc *tailscale.LocalClient) (s *Server, cleanup func
 	if s.devMode {
 		cleanup = s.startDevServer()
 		s.addProxyToDevServer()
+
+		// Create new handler for "/api" requests.
+		// And protect with gorilla csrf.
+		csrfProtect := csrf.Protect(csrfKey())
+		s.apiHandler = csrfProtect(&api{s: s})
```
willnorris, line 80: as noted above, we probably want this csrfProtect on more than just the API handler.  But I also think we can safely pull it out of the `devMode` check at this point.  It should basically be a noop on the existing static web client, and we're past the 1.48 release, so we'll be switching things over relatively soon anyway.

client/web/web.go
soniaappasamy: Talked live, leaving this behind `--dev` for now because we've got the `POST` endpoint hooked up already. Will move it out when we hook up posting from the browser side.

client/web/web.go
soniaappasamy: Yeah we probably actually could widen this to protect basically the whole Server handler. By just wrapping the whole thing. Once we get rid of the existing static site, though, the only piece that won't be wrapped will be the dev server proxy. So maybe okay as-is.

go.mod
soniaappasamy: Running `go mod tidy` didn't seem to change anything - It actually looks like there are no groupings in our current mod file?

client/web/src/hooks/node-data.ts
```diff
@@ -22,16 +23,104 @@ export type UserProfile = {
...
+        body = JSON.stringify(update)
+        contentType = "application/json"
+      }
+
+      apiFetch(url, {
+        method: "POST",
+        headers: { Accept: "application/json", "Content-Type": contentType },
+        body: body,
+      })
+        .then((r) => r.json())
```
willnorris, line 97: if you force the csrf check to fail, then the `r.json()` call fails on the response not being JSON, which gives a not-great error message:

Screenshot 2023-08-17 at 3 15 52 PM

client/web/src/hooks/node-data.ts
```diff
@@ -22,16 +23,104 @@ export type UserProfile = {
...
+      }
+
+      const urlParams = new URLSearchParams(window.location.search)
+      const nextParams = new URLSearchParams({ up: "true" })
+      const token = urlParams.get("SynoToken")
+      if (token) {
+        nextParams.set("SynoToken", token)
+      }
+      const search = nextParams.toString()
+      const url = `/api/data${search ? `?${search}` : ""}`
```
willnorris, line 77: wow... we pass these as URL query parameters on a POST that includes additional data in the POST body?  I wonder if synology requires that the `SynoToken` always be in the URL?  In any event, it's consistent with what we're doing today, so 🤷🏻

willnorris (approved):

client/web/src/hooks/node-data.ts
soniaappasamy: Yeah it's pretty strange. Kept consistent for now, but we should probably clean it up (or at least figure out why and document!) as we start to break things out of the single GET/POST.

PR #8922, opened 2023-08-17
cmd/k8s-operator: support being the default loadbalancer controller

michaelbeaumont:
It'd be great to support this use case where Tailscale is the only controller for LoadBalancer Services.

Updates https://github.com/tailscale/tailscale/issues/502

### DIFF
--- a/cmd/k8s-operator/operator.go
+++ b/cmd/k8s-operator/operator.go
@@ -183,6 +183,9 @@ waitOnline:
 // startReconcilers starts the controller-runtime manager and registers the
 // ServiceReconciler.
 func startReconcilers(zlog *zap.SugaredLogger, tsNamespace string, restConfig *rest.Config, tsClient *tailscale.Client, image, priorityClassName, tags string) {
+	var (
+		isDefaultLoadBalancer = defaultBool("OPERATOR_DEFAULT_LOAD_BALANCER", false)
+	)
 	startlog := zlog.Named("startReconcilers")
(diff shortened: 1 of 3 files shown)


DentonGentry:
`issuebot` complains because the text "Updates https://github.com/tailscale/tailscale/issues/502" was only added to the PR description, which doesn't make it into the git tree after submission. We want to issue number to be in the actual git commit message.

michaelbeaumont:
> We want to issue number to be in the actual git commit message.

Ah right, I somehow overlooked "commit" in the error message.

maisem:
> > We want to issue number to be in the actual git commit message.
> 
> Ah right, I somehow overlooked "commit" in the error message. Is there a specific keyword `issuebot` expects? Does it have to close the issue?

Change the commits to have `Updates #502` instead of `Part of #502`
Can you also squash the commits into one commit that adds the test. Title it `cmd/k8s-operator: support being the default loadbalancer controller`

rhealitycheck (approved):

maisem:
Rebase here too

maisem:
@michaelbeaumont I'd like to get this and #8817 in

michaelbeaumont:
@maisem got it, I'll be able to rebase in about 24 hours

michaelbeaumont:
@maisem done, I left out checking for an IngressClass and default annotation in this PR since that's a little more involved

maisem (approved):

Issue #8923, opened 2023-08-17
tailcfg: clean up Debug


bradfitz:
tailcfg.Debug is an old mechanism to send random debug/tuning/imperative commands to nodes.

It predated the effort to make all fields of MapResponse have a zero value mean unchanged, and it predates c2n (the control-to-node RPC mechanism). As a result, it has a mix of three styles of fields in it: imperative things (which should be c2n), things that need to always be set and don't have a zero value that's sticky, and modern-ish things.

Tracking bug to clean up the fields.
 
/cc @maisem

PR #8924, opened 2023-08-17
tailcfg: move LogHeapPprof from Debug to c2n [capver 69]

bradfitz:
Updates #8923


tailcfg/tailcfg.go
```diff
@@ -1749,15 +1750,6 @@ type ControlIPCandidate struct {
 //
 // TODO(bradfitz): start migrating the imperative ones to c2n requests.
 type Debug struct {
-	// LogHeapPprof controls whether the client should log
-	// its heap pprof data. Each true value sent from the server
-	// means that client should do one more log.
-	LogHeapPprof bool `json:",omitempty"`
```
maisem, line 1755: Kinda annoying to use these fields for old clients if we don't have them in tailcfg. I guess we could do tailcfgout. Also debug, so don't really care if we lose the ability for old clients

maisem (approved):

tailcfg/tailcfg.go
bradfitz: We don't use these old fields. This was here for debugging iOS memory usage. I think I was the only one to ever use it, on my own phone only.

Because there was no other way to get the pprof from the phone.

ipn/ipnlocal/c2n.go
```diff
@@ -25,6 +25,8 @@ import (
 	"tailscale.com/version/distro"
 )
 
+var c2nLogHeap func(http.ResponseWriter, *http.Request) // non-nil on most platforms (c2n_pprof.go)
```
maisem, line 28: Fine, but I hate that click through doesn't work for these in vscode

maisem (approved):

ipn/ipnlocal/c2n.go
bradfitz: does for me?

bradfitz: oh, that it jumps to this line

maisem (approved):

PR #8926, opened 2023-08-17
all: use Go 1.21 slices, maps instead of x/exp/{slices,maps}

bradfitz:
Updates #8419


ipn/ipnlocal/local.go
```diff
@@ -27,9 +27,10 @@ import (
 	"sync/atomic"
 	"time"
 
+	"slices"
```
creachadair, line 30: Group with the rest of the stdlib? (Usually fmt is better about this)

ipn/serve.go
```diff
@@ -12,7 +12,8 @@ import (
 	"strconv"
 	"strings"
 
-	"golang.org/x/exp/slices"
+	"slices"
```
creachadair, line 15: Here too.

creachadair (approved):
There's a handful of places where fmt missed grouping but otherwise LGTM.

ipn/ipnlocal/local.go
bradfitz: Whoops. I thought I'd gotten them all already.

portlist/poller.go
```diff
@@ -13,7 +13,8 @@ import (
 	"sync"
 	"time"
```
valscale, line 15: Rogue newline :)

ssh/tailssh/incubator.go
```diff
@@ -27,11 +27,12 @@ import (
 	"strings"
 	"syscall"
```
valscale, line 29: I think this is another rogue newline, c'mon fmt.

types/views/views.go
```diff
@@ -10,8 +10,9 @@ import (
 	"errors"
 	"net/netip"
```
valscale, line 12: Rogue newline

valscale (approved):
I'm assuming that the flaky test that failed on windows was a flake. Thanks for this!

PR #8931, opened 2023-08-17
cmd/derper: count bootstrap dns unique lookups.

DentonGentry:
Updates https://github.com/tailscale/corp/issues/13979


cmd/derper/bootstrap_dns.go
```diff
@@ -33,8 +34,15 @@ var (
 	publishedDNSMisses   = expvar.NewInt("counter_bootstrap_dns_published_misses")
 	unpublishedDNSHits   = expvar.NewInt("counter_bootstrap_dns_unpublished_hits")
 	unpublishedDNSMisses = expvar.NewInt("counter_bootstrap_dns_unpublished_misses")
+	bootstrapLookupMap   = make(map[string]bool)
```
creachadair, line 37: I believe this is racy, since the metrics scrape is handled in a different goroutine from where these are added. 

Maybe use an expvar.Map just for its baked in synchronization?

syncs/syncs_test.go
```diff
@@ -137,4 +137,23 @@ func TestMap(t *testing.T) {
...
+		if m.Len() != 5 {
+			t.Errorf("Len after loading want=5 got=%d", m.Len())
+		}
+
+		m.Clear()
+		if m.Len() != 0 {
+			t.Errorf("Len after Clear want=0 got=%d", m.Len())
+		}
+	})
+
```
andrew-d, line 158: Trailing newline

syncs/syncs.go
```diff
@@ -227,6 +228,13 @@ func (m *Map[K, V]) Len() int {
 	return len(m.m)
 }
 
+// Clear removes all entries from the map.
+func (m *Map[K, V]) Clear() {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
```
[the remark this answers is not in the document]
DentonGentry: Done.

syncs/syncs_test.go
DentonGentry: Done. However lint is still unhappy, doesn't say why, and doesn't run locally.

```
dgentry@ubuntu:tailscale/tailscale/syncs$ golangci-lint run --new=false --new-from-rev=
ERRO Running error: unknown linters: 'bidichk,revive', run 'golangci-lint linters' to see the list of supported linters
```

DentonGentry: Done.

cmd/derper/bootstrap_dns.go
```diff
@@ -35,6 +37,12 @@ var (
 	unpublishedDNSMisses = expvar.NewInt("counter_bootstrap_dns_unpublished_misses")
 )
 
+func init() {
+	expvar.Publish("counter_bootstrap_dns_queried_domains", expvar.Func(func() any {
+		return fmt.Sprintf("%d", bootstrapLookupMap.Len())
```
andrew-d, line 42: Optional, but... `expvar.Func` works by formatting the returned value using JSON, so you should be able to return `bootstrapLookupMap.Len()` directly here, without the `fmt.Sprintf`.

andrew-d (approved):

cmd/derper/bootstrap_dns.go
DentonGentry: Done.

cmd/derper/bootstrap_dns.go
creachadair: It would. I was imagining you wouldn't publish it directly, but just use it as a repository for the data to publish via the Func.

But that would be a kluge, and your solution is better. 😄

creachadair (approved):

PR #8932, opened 2023-08-17
wgengine/magicsock: only accept pong sent by CLI ping

valscale:
When sending a ping from the CLI, only accept a pong that is in reply to the specific CLI ping we sent. This matters now that not all pings are equal - some of them are different sizes and we want to know which one got through, specifically. Otherwise we get a heartbeat pong and think that the larger sized ping got through when it did not.

Updates #311

### DIFF
--- a/wgengine/magicsock/endpoint.go
+++ b/wgengine/magicsock/endpoint.go
@@ -71,8 +71,6 @@ type endpoint struct {
 	endpointState      map[netip.AddrPort]*endpointState
 	isCallMeMaybeEP    map[netip.AddrPort]bool
 
-	pendingCLIPings []pendingCLIPing // any outstanding "tailscale ping" commands running
-
 	// The following fields are related to the new "silent disco"
 	// implementation that's a WIP as of 2022-10-20.
 	// See #540 for background.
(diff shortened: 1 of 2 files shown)


danderson (approved):

wgengine/magicsock/endpoint.go
```diff
@@ -99,6 +92,8 @@ type sentPing struct {
 	at      mono.Time
 	timer   *time.Timer // timeout timer
 	purpose discoPingPurpose
+	res     *ipnstate.PingResult
+	cb      func(*ipnstate.PingResult)
```
bradfitz, line 96: looks like this can be nil? document a bit here

```
    cb func(*ipnstate.PingResult) // or nil if .....
```

wgengine/magicsock/endpoint.go
```diff
@@ -574,7 +568,7 @@ const (
 )
 
 // startDiscoPingLocked sends a disco ping to ep in a separate goroutine.
```
bradfitz, line 570: which of res/cb can be nil here? res is the result to fill in?

wgengine/magicsock/endpoint.go
valscale: Thanks! Fixed.

bradfitz (approved):

PR #8934, opened 2023-08-17
wgengine,net,ipn,disco: split up and define different types of MTU

valscale:
Prepare for path MTU discovery by splitting up the concept of DefaultMTU() into the concepts of the Tailscale TUN MTU, MTUs of underlying network interfaces, minimum "safe" TUN MTU, user configured TUN MTU, probed path MTU to a peer, and maximum probed MTU.

When reading, keep in mind that the user can at any point unilaterally reconfigure the Tailscale TUN MTU to nearly 64K so the code needs to handle packets up to that size one way or another.

I split the MTUs to probe off into the next set of commits.


net/tstun/mtu.go
```diff
@@ -1,33 +1,168 @@
...
+//
+// 1. The MTU set by the user via the OS, if it has ever been set
+// 2. If TS_DEBUG_ENABLE_PMTUD is set, the maximum size MTU we probe, minus wg overhead
+// 4. If TS_DEBUG_ENABLE_PMTUD is not set, the Safe MTU
+
+// TunMTU is used to describe MTUs for the tailscale TUN, while
+// WireMTU is used to describe MTUs for the underlying network
+// devices.
+type TunMTU uint32
+type WireMTU uint32
```
andrew-d, line 70: Go documentation is only applied to the variable directly below a comment, so the docs look like this:
```
$ go doc ./net/tstun TunMTU
package tstun // import "tailscale.com/net/tstun"

type TunMTU uint32
    TunMTU is used to describe MTUs for the tailscale TUN, while WireMTU is used
    to describe MTUs for the underlying network devices.

func DefaultTunMTU() TunMTU
func WireToTunMTU(w WireMTU) TunMTU

$ go doc ./net/tstun WireMTU
package tstun // import "tailscale.com/net/tstun"

type WireMTU uint32

var MaxProbedWireMTU WireMTU
func DefaultWireMTU() WireMTU
func TunToWireMTU(t TunMTU) WireMTU
```

You may want to move the comment, so that documentation is applied correctly:
```suggestion
// TunMTU is used to describe MTUs for the tailscale TUN.
type TunMTU uint32
// WireMTU is used to describe MTUs for the underlying network
// devices.
type WireMTU uint32
```

net/tstun/mtu.go
```diff
@@ -1,33 +1,168 @@
...
-	tunMTU := defaultMTU
-	if mtu, ok := envknob.LookupUintSized("TS_DEBUG_MTU", 10, 32); ok {
-		mtu := uint32(mtu)
-		if mtu > maxMTU {
-			mtu = maxMTU
+// WireMTUsToProbe are the MTUs we are most likely to see in real world
+// networks. These are used to send discovery pings in packets padded
+// to exactly match the goal MTU.
+//
+// XXX Review these before committing
```
andrew-d, line 84: PR is still a draft, but FYI that this is still here

net/tstun/mtu.go
```diff
@@ -1,33 +1,168 @@
...
+	if m, ok := envknob.LookupUintSized("TS_DEBUG_MTU", 10, 32); ok {
+		debugMTU := TunMTU(m)
+		if debugMTU > maxTunMTU {
+			debugMTU = maxTunMTU
+		}
+		return debugMTU
+	}
+
+	debugPMTUD, _ := envknob.LookupBool("TS_DEBUG_ENABLE_PMTUD")
+	if debugPMTUD == true {
```
andrew-d, line 157: Can simplify to `if debugPMTUD {` if you want

andrew-d (commented):
Generally LGTM, with a few minor comments. That comment block in `mtu.go` is a thing of beauty 🤩

net/tstun/mtu.go
valscale: Thanks! done.

net/tstun/mtu.go
valscale: Thanks! Let me know if you have any thoughts on which sizes to test

net/tstun/mtu.go
valscale: Thanks, fixed!

wgengine/netstack/netstack.go
```diff
@@ -1041,17 +1042,17 @@ func (ns *Impl) acceptUDP(r *udp.ForwarderRequest) {
...
 	// bound on resource usage, while hopefully still being long enough for
 	// a heavily loaded system.
 	const readDeadline = 150 * time.Millisecond
 
 	defer c.Close()
-	q := make([]byte, maxUDPReqSize)
+	// Implementations are advised not to exceed 512 bytes per DNS
+	// request due to fragmenting but in reality can and do send
+	// much larger packets.
+	q := make([]byte, maxUDPPacketSize)
```
bradfitz, line 1055: So instead of allocating ~1280 bytes here we're allocating ~64K now for DNS queries? The comment says it may exceed 512 but do we really expect DNS queries up to 64KB in a UDP packet?

(Bigger allocs here means more garbage and thus more GCs and thus more CPU usage.)

MagicDNS gets enough traffic to be worth paying attention to performance here probably.

net/tstun/mtu.go
```diff
@@ -1,33 +1,171 @@
...
-	tunMTU := defaultMTU
-	if mtu, ok := envknob.LookupUintSized("TS_DEBUG_MTU", 10, 32); ok {
-		mtu := uint32(mtu)
-		if mtu > maxMTU {
-			mtu = maxMTU
+// WireMTUsToProbe are the MTUs we are most likely to see in real world
+// networks. These are used to send discovery pings in packets padded
+// to exactly match the goal MTU.
+//
+// XXX These need adjusting before merging to main. What do you think?
```
bradfitz, line 82: reminder to remove all "XXX" before merging. (we check in TODO but never "XXX" so you can always depend on XXX being greppable to find your own temporary comments)

wgengine/netstack/netstack.go
valscale: Yeah, this is a bummer and I would appreciate suggestions for making it go fast given that in reality, TS_DEBUG_MTU means a UDP packet can be up to ~64K, and when PeerMTU is on, we will regularly see 9000 byte MTU and potentially DNS packets as large as either. Googling a bit suggests that 8K is the real world max of a DNS query but I don't know if that's a hard limit. Thoughts?

net/tstun/mtu.go
valscale: The XXX is here precisely so I don't merge it. :) Any thoughts on which MTU size to probe?

net/tstun/mtu_test.go
```diff
@@ -4,25 +4,81 @@ package tstun
...
+	os.Setenv("TS_DEBUG_ENABLE_PMTUD", "")
+
+	// With no MTU envknobs, we should get the conservative MTU.
+	if DefaultTunMTU() != safeTunMTU {
+		t.Errorf("DefaultTunMTU() = %d, want %d", DefaultTunMTU(), safeTunMTU)
+	}
+
+	// TS_DEBUG_MTU should take precedence over every other setting.
+	mtu := maxTunMTU - 1
+	os.Setenv("TS_DEBUG_MTU", strconv.Itoa(int(mtu)))
```
bradfitz, line 31: in general you want `envknob.Setenv` because it handles registered (the more efficient) envknobs, which you might want to switch to later if the envknob lookups are in your hot paths

net/tstun/mtu_test.go
```diff
@@ -4,25 +4,81 @@ package tstun
...
-	if DefaultMTU() != maxMTU {
-		t.Errorf("DefaultMTU() = %d, want %d", DefaultMTU(), maxMTU)
+	for _, tt := range tests {
+		m := WireToTunMTU(tt.w)
+		if m != tt.t {
+			t.Errorf("conversion of wire MTU %v to TUN MTU = %v, want %v", tt.w, m, tt.t)
+		}
+	}
+	for _, tt := range tests {
+		m := WireToTunMTU(tt.w)
```
bradfitz, line 79: this block is identical to line 72's block? did you mean to test the other direction?

wgengine/netstack/netstack.go
bradfitz: we could use a sync.Pool of 64KB ones to at least not make garbage in the common/hot cases

sailorfrag: I was about to suggest `sync.Pool` but also we limit forwarded DNS responses to 4KB
https://github.com/tailscale/tailscale/blob/main/net/dns/resolver/tsdns.go#L42-L45

We can also pick an arbitrary lower limit and respond with the error that tells the client to switch to TCP.

bradfitz: e.g.

```patch
diff --git a/wgengine/netstack/netstack.go b/wgengine/netstack/netstack.go
index 54e99180..0217f5e1 100644
--- a/wgengine/netstack/netstack.go
+++ b/wgengine/netstack/netstack.go
@@ -1040,10 +1040,14 @@ func (ns *Impl) acceptUDP(r *udp.ForwarderRequest) {
        go ns.forwardUDP(c, srcAddr, dstAddr)
 }
 
+var udpBufPool = &sync.Pool{
+       New: func() any {
+               b := make([]byte, maxUDPPacketSize)
+               return &b
+       },
+}
+
 func (ns *Impl) handleMagicDNSUDP(srcAddr netip.AddrPort, c *gonet.UDPConn) {
-       // In practice, implementations are advised not to exceed 512 bytes
-       // due to fragmenting. Just to be sure, we bump all the way to the MTU.
-       var maxUDPReqSize = tstun.DefaultMTU()
        // Packets are being generated by the local host, so there should be
        // very, very little latency. 150ms was chosen as something of an upper
        // bound on resource usage, while hopefully still being long enough for
@@ -1051,7 +1055,10 @@ func (ns *Impl) handleMagicDNSUDP(srcAddr netip.AddrPort, c *gonet.UDPConn) {
        const readDeadline = 150 * time.Millisecond
 
        defer c.Close()
-       q := make([]byte, maxUDPReqSize)
+
+       bufp := udpBufPool.Get().(*[]byte)
+       defer udpBufPool.Put(bufp)
+       q := *bufp
 
        // libresolv from glibc is quite adamant that transmitting multiple DNS
        // requests down the same UDP socket is valid. To support this, we read
@@ -1164,7 +1171,11 @@ func startPacketCopy(ctx context.Context, cancel context.CancelFunc, dst net.Pac
        }
        go func() {
                defer cancel() // tear down the other direction's copy
-               pkt := make([]byte, maxUDPPacketSize)
+
+               bufp := udpBufPool.Get().(*[]byte)
+               defer udpBufPool.Put(bufp)
+               pkt := *bufp
+
                for {
                        select {
                        case <-ctx.Done():
```

net/tstun/mtu.go
bradfitz: No clue. I had always assumed the plan was to binary search it rather than try some specific values.

No opinions, as I don't really know what the design is.

Are these some sizes to spray out all at once to get a starting point to find the specific number? How often?

wgengine/netstack/netstack.go
valscale: To be honest I don't understand the big picture of how this code is used. Why are we forwarding DNS packets? Who do they come from? Who will be sad if we truncate or drop a packet?

net/tstun/mtu.go
valscale: Yeah, the observation is that in reality, there are only a small number of MTU values. Rather than binary search for them, just try them all at the same time and take the largest. The plan is to limit MTU discovery to once every 10 minutes, which is the standard for normal PMTUD cache expiration.

sailorfrag: There's a study on this subject, but it doesn't include jumbo frames since those can't usually transit the Internet
https://aura.abdn.ac.uk/bitstream/handle/2164/11693/tma2018_paper57.pdf
It... is surprisingly lacking in helpful answers for this question though.

This would be my list to guess from:

- Interface MTU if >9000
- 9000 (common jumbo frame choice)
- Some intermediate jumbo sizes. I've seen guides that recommend 8192 because it's a multiple of page size but that doesn't include the Ethernet header so I'm unclear it actually does what they expect. Perhaps 2048 - 14, 4096 - 14, 8192 - 14?
- 1500
- 1492 (PPPoE)
- 1476 (GRE)
- 1436? (IPsec ESP with GRE says maximum of 64 bytes overhead [e.g. Zscaler guide](https://help.zscaler.com/zia/determining-the-optimal-mtu-for-gre-or-ipsec-tunnels))
- 1400 (Zscaler guide recommends this as a pretty safe option for most cases)
- 1280 (IPv6 minimum, current Tailscale value)

If you want a super short list, I guess 9000, 1500, 1492, 1400, 1280 seems good?

Converging quickly on one of those and then probing a bit further to find the exact value later seems like an ideal approach, but is a little annoying because I don't think TCP is willing to increase the MSS after it's initially established. I believe it can be reduced later. It might be worth accepting the cost of some fragmentation at first before the value has been nailed down if TCP can be coaxed into decreasing it later.

RFC 1191 says that the next-hop MTU must be populated in the ICMP error, but RFC 792 does not, so fast probing a few options and then using that to guide finding an exact value (binary search between options otherwise) might work well. It's going to be impossible to get this perfect because the correct answer can change as peers switch between different networks from time to time (e.g. a laptop that moves around).

net/tstun/mtu.go
```diff
@@ -1,33 +1,171 @@
...
+var WireMTUsToProbe = [...]WireMTU{
+	576,  // Smallest MTU for IPv4, probably useless?
```
sailorfrag, line 84: `576 - 80` is going to be too small, so I'd say `656` is the lowest worth probing.

net/tstun/mtu.go
```diff
@@ -1,33 +1,171 @@
...
+	1124, // An observed MTU in the wild, maybe 1100 instead?
+	1280, // Smallest MTU allowed for IPv6, current default
```
sailorfrag, line 86: I believe the current default is `1280` TUN MTU, so `1360` wire MTU? If we can't manage a payload of the IPv6 minimum packet size, we're probably going to have a bad time.

net/tstun/mtu.go
```diff
@@ -1,33 +1,171 @@
...
+}
+
+// WireToTunMTU converts the on-the-wire MTU to the Tailscale MTU
+// necessary to transmit a packet of TUN MTU bytes plus the
+// Tailscale/Wireguard overhead.
+func WireToTunMTU(w WireMTU) TunMTU {
+	if w < wgHeaderLen {
+		return 0
+	}
+	return TunMTU(w - wgHeaderLen)
```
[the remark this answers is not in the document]
valscale: Thanks! This specific case was too hard to read using min() because the args are unsigned. So we can't do the straightforward max(w-wgHeaderLen, 0) and instead have to do w - min(w, wgHeaderLen) which ??? no thanks. But I used it in a bunch of other places.

net/tstun/mtu.go
valscale: Thanks! So the standard PMTUD algorithm actually recommends stopping binary search when it gets "close enough" rather than getting it to the exact bytes, and it discards the cached PMTU entry every 10 minutes and restarts discovery, so that's what people are living with right now. I like the idea of adding an underlying interface MTU as one of the probe values.

net/tstun/mtu.go
```diff
@@ -1,33 +1,171 @@
...
+// TunMTU is used to describe MTUs for the tailscale TUN.
+type TunMTU uint32
+
+// WireMTU is used to describe MTUs for the underlying network devices.
+type WireMTU uint32
 
 const (
-	maxMTU     uint32 = 65536
-	defaultMTU uint32 = 1280
+	// Largest MTU we will consider for the Tailscale TUN. This is inherited
```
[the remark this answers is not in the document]
valscale: Fixed, thanks!

net/tstun/mtu_test.go
```diff
@@ -4,25 +4,81 @@ package tstun
...
 )
 
-func TestDefaultMTU(t *testing.T) {
-	orig := os.Getenv("TS_DEBUG_MTU")
-	defer os.Setenv("TS_DEBUG_MTU", orig)
+func TestDefaultTunMTU(t *testing.T) {
+	// Save and restore the envknobs we will be changing
 
+	// The user can specify the MTU with this envknob.
+	ts_debug_mtu := os.Getenv("TS_DEBUG_MTU")
```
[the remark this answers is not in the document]
valscale: Fixed, thanks! Slowly working the K&R C out of my system.

net/tstun/mtu_test.go
valscale: Yes, thanks! Fixed.

net/tstun/mtu.go
valscale: Thanks, fixed also. :)

wgengine/netstack/netstack.go
raggi: Regarding large DNS packets, we have had some problems with responses over 1280 bytes with specific queries and resolvers. About a year ago I diagnosed an issue with pihole and the DNS resolver it was using that was producing abnormally large responses when it had DNSSEC validation enabled. A more recent version stopped doing that, but as an anecdote, this was an in-the-wild case of DNS responses being larger than we could (at the time) handle, and it caused user visible issues (~they couldn't resolve DNSSEC enabled domains with IPv6 addresses).

We have some large buffer pools in wireguard-go already, which we needed to add as part of the packet coalescing work. I wouldn't be too concerned about adding another, though at some point we may need to have bounded sized pools. wireguard-go has an implementation of such a thing here: https://github.com/tailscale/wireguard-go/blob/master/device/pools.go#L13 i don't remember seeing one of our own, though perhaps we have one somewhere.

net/tstun/mtu.go
raggi: How close is "close enough", without doing the math it surely doesn't save much more than two packets?

I think what I'd be tempted to do is to _not discard_ and to instead use the already discovered MTU for a lightweight retest, so test [cur, cur+1]. If cur+1 succeeds, then initiate a new "full probe" from [cur+1, max]. If both fail, initiate full probe with the default range.

Another value we might be able to gather on _some platforms_ as a starting point is to request the MSS from getsockopt TCP_INFO on platforms where it/equivalents are available, for any socket that's going through `netns`. This would give us a solid starting point.

net/tstun/mtu.go
```diff
@@ -1,33 +1,171 @@
...
+import (
+	"tailscale.com/envknob"
+)
+
+// The MTU (Maximum Transmission Unit) of a network interface is the largest
+// packet that can be sent or received through that interface, including all
+// headers above the link layer (e.g. IP headers, UDP headers, Wireguard
+// headers, etc.). We have to think about several different values of MTU:
+//
+// Wire MTU: The MTU of an interface underneath the tailscale TUN, e.g. an
```
raggi, line 15: I feel like we still need to discuss the possibility of multiple wire paths to a peer somewhere.

net/tstun/mtu.go
raggi, line 84: I don't think we should probe down to 576, we definitely don't support it today, and we would definitely need 576 + tunnel overhead to fit valid IPv4, which is our practical lower bound. I don't want to go down to 576+overhead either, though.

I want to suggest we don't probe below 1280, however, I also want to suggest:

- We have had a handful of WSL users trying to use Tailscale on Windows and WSL at the same time, then turn on an exit node on the Windows side and break themselves, as Tailscale can't carry Tailscale traffic today.
- Similar story, but for less easy to describe VM configurations.

So I'd like to suggest we go down to a minimum of 1280-tunnel overhead. This would allow us to fit exactly one layer of Tailscale over Tailscale. If someone sets up more layers in nested virt, I'm 100% ready to let that fail.

I would also be entirely comfortable with saying we only go as low as we do today, 1280.

raggi (commented):
LGTM, modulo the need for pooling the DNS buffer. As commented I think we could safely trim the probed range a little, cutting off the bottom.

net/tstun/mtu.go
valscale: These are all great ideas! I am going to remove the probed MTUs array part of the patch and hard code the max probed MTU for the purposes of this PR for splitting up MTU concepts, which shouldn't include that stuff anyway, and then incorporate these ideas in the PR that actually probes the MTU. Thank you!

net/tstun/mtu.go
valscale: The MTU to a peer is stored per endpoint, so if I understand your comment correctly, that should allow us to have multiple different MTUs for different wire paths to a peer as long as they have different endpoints. We can still have multiple wire paths to the same IP/port but the standard TCP PMTUD implementation has to deal with that too, and I think we will handle it the same way: reprobe when we have too much packet loss and/or every 10 minutes. I think you might be suggesting that we could have multiple paths such that, e.g., every other packet goes over a path with one MTU and the in between packets go over a different path with another MTU? Which would definitely be bad for this algorithm. Let me know if I'm not getting it!

wgengine/netstack/netstack.go
valscale: I agree with you both. I added a separate commit co-authored by Brad that adds the buffer pools, then the following commit raises the buffer size.

net/tstun/mtu.go
raggi: Endpoint is an overloaded term in our system, but magicsock has an `endpoint` type that in our system is 1-1 with a peer, and internally carries IP:port tuples. When WireGuard sends a packet, it is directed to `(*endpoint).send` which then picks an address based on a latency sample and time based heuristic, falling back to DERP if there is no other good option.

We've observed that the selected IP:port in this selection can change regularly if there are influences to latency (for example, running a performance test that saturates bloated buffers along the link can trigger a switch in IP:port tuple regularly during the run). Most often for dual homed users this is a switch between IPv4 and IPv6, which would in fact alter the peer path MTU by the difference in header size between IPv4 and IPv6. This switching does not* cause an MTU change today because our MTU is fixed at 1280.

*Technically it does actually produce an MTU change, as it alters the minimum MTU space we need on the host link in order to successfully communicate packets, which if you happen to be sitting between 1300 and 1360 bytes MTU can be a problem, but we see this approximately never in practice. There appears to be a bias of either low MTUs (very close to or exactly 1280, such as the current WSL default), or higher MTUs (1400+).

valscale: Notes from our discussion: first, yes, endpoint is overloaded :) I should have said "path MTU is stored per address" and there are multiple of those per endpoint/peer. Currently the MTU calculations assume worst case overhead (IPv6) so flipping between v4/v6 shouldn't break anything. We think that the worst case is we end up with the lower path MTU for 10 minutes, which is a good reason to avoid switching between best addresses unless there's a really obvious improvement. We think it's likely that if there's a link with a significantly larger MTU that it is also the lowest latency/highest bandwidth/most reliable path. I'll address this more in later commits as it doesn't need to be resolved in this one.

valscale:
@raggi @sailorfrag Somehow this PR included an unrelated commit; I rebased and now it is just the two relevant commits. Thanks for your patience!

net/tstun/mtu.go
```diff
@@ -1,33 +1,145 @@
...
+
+// WireMTU is the MTU for the underlying network devices.
+type WireMTU uint32
 
 const (
-	maxMTU     uint32 = 65536
-	defaultMTU uint32 = 1280
+	// maxTunMTU is the largest MTU we will consider for the Tailscale
+	// TUN. This is inherited from wireguard-go and can be surprisingly
+	// small; on Windows it is currently 2048 - 32 bytes.
```
raggi, line 71: We don't currently use the rio Bind on Windows, we use our own conn.Bind in https://github.com/tailscale/tailscale/blob/main/wgengine/magicsock/magicsock.go#L2018

raggi: That said, the value below bottoms out in: 

```
device/queueconstants_android.go:       MaxSegmentSize             = 2200
device/queueconstants_default.go:       MaxSegmentSize             = (1 << 16) - 1 // largest possible UDP datagram
device/queueconstants_ios.go:const MaxSegmentSize = 1700
device/queueconstants_windows.go:       MaxSegmentSize             = 2048 - 32 // largest possible UDP datagram
```

So iOS is also rather low and I do not know/remember where the 1700 value came from. For one thing we need to remain somewhat conservative with allocation on iOS and we have a desire to adjust the buffer pooling strategy so we can eventually increase these, but without substantially increasing our active pool watermark. The last time we were thinking about this substantially, we were considering a dependency on gvisors buffer infrastructure (which is skbuff like in capabilities), though there are necessary discussions about that and it's relationship to pooling on the wireguard-go side.

net/tstun/mtu.go
```diff
@@ -1,33 +1,145 @@
...
+//
+// A Wireguard header includes:
+//
+// - 20-byte IPv4 header or 40-byte IPv6 header
+// - 8-byte UDP header
+// - 4-byte type
+// - 4-byte key index
+// - 8-byte nonce
+// - 16-byte authentication tag
+
```
raggi, line 101: the extra line will cause the comment parser to not associate this with the symbol below

net/tstun/mtu.go
```diff
@@ -1,33 +1,145 @@
...
+const wgHeaderLen = 40 + 8 + 4 + 4 + 8 + 16
+
+// TunToWireMTU takes the MTU that the Tailscale TUN presents to the user and
+// returns the on-the-wire MTU necessary to transmit the largest packet that
+// will fit through the TUN, given that we have to add wireguard headers.
+func TunToWireMTU(t TunMTU) WireMTU {
```
raggi, line 107: Although I'm not keen on how it reads, the convention for acronym naming would be `TUNToWireMTU`, `WireToTUNMTU` and `DefaultTUNMTU`, same actually for the TunMTU constant would be `TUNMTU`

raggi (approved):
Looks good to me, just a convention items

sailorfrag (approved):

net/tstun/mtu.go
valscale: 😭 Snake case works great until you have two acronyms together... I did the rename but I cried the whole time. 😆

net/tstun/mtu.go
sailorfrag: Yeah, Windows uses it in the rio Bind that doesn't affect us, but the value also propagated elsewhere in Windows code that does affect us. I think that plan sounds good.

PR #8939, opened 2023-08-17
tailcfg: add nodeAttrs for forcing OneCGNAT on/off [capver 71]

bradfitz:
Updates #8923


tailcfg/tailcfg.go
```diff
@@ -1754,26 +1755,35 @@ type Debug struct {
...
 	// re-enabled for the lifetime of the process.
+	//
+	// This is primarily used by Headscale.
 	DisableLogTail bool `json:",omitempty"`
 
 	// Exit optionally specifies that the client should os.Exit
-	// with this code.
+	// with this code. This is a safety measure in case a client is crash
+	// looping or in an unsafe state and we need to remotely shut it down.
 	Exit *int `json:",omitempty"`
```
maisem, line 1787: Could be c2n

maisem (approved):

tailcfg/tailcfg.go
bradfitz: Could. But then if we ever need this there are conditionals in control by capver. We've never needed it. Just one scare IIRC.

Issue #8927, opened 2023-08-17
"cannot determine CPU architecture for Synology model "DS213v10-j" (Go arch "arm")"


Smultie:
### What is the issue?

Decided to try out the new 'tailscale update' feature on my Synology NAS. 
Not sure if it's supposed to work on the device, either

### Steps to reproduce

Log into Synology with SSH.
'tailscale update'

### OS

Synology

### OS version

Synology DSM 6.2.4

### Tailscale version

1.48.0

snachx:
same problem here DS216play-j

awly:
Those device models need a [CPU-specific SPK package](https://pkgs.tailscale.com/stable/#spks). Without knowing the exact CPU architecture, `tailscale update` can't figure out which SPK to download.
Some of the Synology CPU architectures (anything other than i686, x86_64 and armv8) are armv5 or armv7 from Go's perspective, but they won't accept a generic armv5/v7 SPK. So we currently map certain device models to CPU architectures they have per Synology docs.  

We detect the device model in `tailscale update` by reading `/proc/sys/kernel/syno_hw_version`, which may or may not be the right place. DSM doesn't seem to ship any other tools to detect this programmatically.

@Smultie @snachx do you know which exact model your devices are, according to the list at https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures or https://kb.synology.com/en-us/DSM/tutorial/What_kind_of_CPU_does_my_NAS_have?

Is `DS213v10-j` is `DS213` or `DS213j`?
Is `DS216play-j` the same as `DS216play` or `DS216j`?

One other useful piece of data is `/etc/synoinfo.conf`. Could you paste the output of `grep platform_name /etc/synoinfo.conf`?
Perhaps we can detect CPU platforms that way, instead of by device model name.

Smultie:
grep platform_name /etc/synoinfo.conf showed no information 

However, when scrolling through /etc/synoinfo.conf I noticed the following field:
unique="synology_88f6282_213"

Not sure what that reads for other devices (@snachx ??), but I'm pretty sure the '88f6282' portion refers to the CPU.

snachx:
@awly @Smultie 

My device model is DS216play, the package arch is monaco, DS216play and DS216j are different devices with different package arch.

the output of grep platform_name /etc/synoinfo.conf is empty.
grep unique /etc/synoinfo.conf shows:
unique="synology_monaco_ds216play"

[...]

PR #8940, opened 2023-08-17
clientupdate: parse /etc/synoinfo.conf to get CPU arch

awly:
The hardware version in `/proc/sys/kernel/syno_hw_version` does not map exactly to versions in
https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures. It contains some slightly different version formats.

Instead, `/etc/synoinfo.conf` exists and contains a `unique` line with the CPU architecture encoded. Parse that out and filter through the list of architectures that we have SPKs for.

Tested on DS218 and DS413j.

Updates #8927

### DIFF
--- a/clientupdate/clientupdate_test.go
+++ b/clientupdate/clientupdate_test.go
@@ -8,8 +8,6 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
-
-	"tailscale.com/tailcfg"
 )
 
 func TestUpdateDebianAptSourcesListBytes(t *testing.T) {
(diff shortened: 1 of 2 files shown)


clientupdate/clientupdate_test.go
```diff
@@ -446,29 +444,108 @@ tailscale installed size:
...
+			content: `
+company_title="Synology"
+`,
+			wantErr: true,
+		},
+		{
+			desc: "empty unique",
+			content: `
+company_title="Synology"
+unique=""
```
noncombatant, line 529: Maybe also try blank and without quotes. (Should we handle single quotes, too?)

clientupdate/clientupdate_test.go
```diff
@@ -446,29 +444,108 @@ tailscale installed size:
...
-		{goarch: "arm", model: "DS115", want: "armada375"},
-		{goarch: "arm", model: "DS419slim", want: "armada38x"},
-		{goarch: "arm", model: "RS815", want: "armadaxp"},
-		{goarch: "arm", model: "DS414j", want: "comcerto2k"},
-		{goarch: "arm", model: "DS216play", want: "monaco"},
-		{goarch: "riscv64", model: "DS999", wantErr: true},
+		{
+			desc: "normal",
+			content: `
+company_title="Synology"
```
noncombatant, line 505: Do we have an example of a complete file? (Or is this it?)

clientupdate/clientupdate_test.go
awly: Added more test cases!

clientupdate/clientupdate_test.go
awly: The complete file is like 380 lines long.
It's a kitchen sink of undocumented parameters :(

noncombatant: Wow! Alas.

noncombatant (approved):

PR #8941, opened 2023-08-17
util/winutil: add RegisterForRestart, allowing programs to indicate t…

dblohm7:
…heir preferences to the Windows restart manager

In order for the installer to restart the GUI correctly post-upgrade, we need the GUI to be able to register its restart preferences.

This PR adds API support for doing so. I'm adding it to OSS so that it is available should we need to do any such registrations on OSS binaries in the future.

Updates https://github.com/tailscale/corp/issues/13998


util/winutil/winutil_windows.go
```diff
@@ -551,3 +553,58 @@ func findHomeDirInRegistry(uid string) (dir string, err error) {
...
+	}
+	if !opts.RestartOnReboot {
+		flags |= _RESTART_NO_REBOOT
+	}
+
+	var cmdLine *uint16
+	if opts.UseCmdLineArgs {
+		if len(opts.CmdLineArgs) == 0 {
+			// re-use our current args, excluding the exe name itself
+			opts.CmdLineArgs = os.Args[1:]
```
sailorfrag, line 584: What's `os.Args[0]`? Isn't that the exe name?

sailorfrag (approved):

util/winutil/winutil_windows.go
dblohm7: Yeah. This API explicitly does *not* want the exe name included in the command line (it prepends that itself).

sailorfrag: Weird, I swear that said _including_ so that's a reading comprehension fail, I guess.

Issue #8760, opened 2023-07-31
FR: sign release tarballs for Linux


awly:
### What are you trying to do?

Prove that release tarballs served from pkgs.tailscale.com have actually been built by Tailscale to prevent pkgs server compromise.

### How should we solve this?

Sign tarballs with a key owned by Tailscale and distribute signatures along with the tarballs.
This key should support some form of rotation.

### What is the impact of not solving this?

Trust in WebPKI and pkgs server not being owned serving malicious tarballs.

awly:
Unstable tarballs are now signed and can be downloaded/verified using https://github.com/tailscale/tailscale/tree/main/clientupdate/distsign.
Using these signed tarballs for `tailscale update` is tracked in https://github.com/tailscale/tailscale/issues/6995

PR #8943, opened 2023-08-17
clientupdate/distsign: add new library for package signing/verification

awly:
This library is intended for use during release (`cmd/dist`) to sign packages which are then served from pkgs.tailscale.com.
The library is also then used by clients downloading packages for `tailscale update` where OS package managers / app stores aren't used.

In a separate PR I will make a CLI around `distsign.Signer` for common operations like "generate new keys", "sign signing keys", "re-sign stuff with a new signing key". This PR is too big to include that :) 

Updates https://github.com/tailscale/tailscale/issues/8760 
Updates https://github.com/tailscale/tailscale/issues/6995

### DIFF
--- a/clientupdate/distsign/distsign.go
+++ b/clientupdate/distsign/distsign.go
@@ -1,0 +1,338 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+// Package distsign implements signature and validation of arbitrary
+// distributable files.
+//
+// There are 3 parties in this exchange:
+//   - builder, which creates files, signs them with signing keys and publishes
+//     to server
+//   - server, which distributes public signing keys, files and signatures
+//   - client, which downloads files and signatures from server, and validates
+//     the signatures
+//
+// There are 2 types of keys:
+//   - signing keys, that sign individual distributable files on the builder
+//   - root keys, that sign signing keys and are kept offline
+//
+// root keys -(sign)-> signing keys -(sign)-> files
+//
+// All keys are asymmetric Ed25519 key pairs.
+//
+// The server serves static files under some known prefix. The kinds of files are:
+//   - distsign.pub - bundle of PEM-encoded public signing keys
+//   - distsign.pub.sig - signature of distsign.pub using one of the root keys
+//   - $file - any distributable file
+//   - $file.sig - signature of $file using any of the signing keys
+//
+// The root public keys are baked into the client software at compile time.
+// These keys are long-lived and prove the validity of current signing keys
+// from distsign.pub. To rotate root keys, a new client release must be
+// published, they are not rotated dynamically. There are multiple root keys in
+// different locations specifically to allow this rotation without using the
+// discarded root key for any new signatures.
+//
+// The signing public keys are fetched by the client dynamically before every
+// download and can be rotated more readily, assuming that most deployed
+// clients trust the root keys used to issue fresh signing keys.
+package distsign
+
+import (
+	"crypto"
+	"crypto/ed25519"
+	"crypto/rand"
+	"encoding/binary"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"hash"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+
+	"github.com/hdevalence/ed25519consensus"
+	"golang.org/x/crypto/blake2s"
+)
+
+const (
+	pemTypePrivate = "PRIVATE KEY"
+	pemTypePublic  = "PUBLIC KEY"
+
+	downloadSizeLimit    = 1 << 29 // 512MB
+	signingKeysSizeLimit = 1 << 20 // 1MB
+	signatureSizeLimit   = ed25519.SignatureSize
+)
+
+// GenerateKey generates a new key pair and encodes it as PEM.
+func GenerateKey() (priv, pub []byte, err error) {
+	pub, priv, err = ed25519.GenerateKey(rand.Reader)
+	if err != nil {
+		return nil, nil, err
+	}
+	return pem.EncodeToMemory(&pem.Block{
+			Type:  pemTypePrivate,
+			Bytes: []byte(priv),
+		}), pem.EncodeToMemory(&pem.Block{
+			Type:  pemTypePublic,
+			Bytes: []byte(pub),
+		}), nil
+}
+
+// RootKey is a root key Signer used to sign signing keys.
+type RootKey Signer
+
+// SignSigningKeys signs the bundle of public signing keys. The bundle must be
+// a sequence of PEM blocks joined with newlines.
+func (s *RootKey) SignSigningKeys(pubBundle []byte) ([]byte, error) {
+	return s.Sign(nil, pubBundle, crypto.Hash(0))
+}
+
+// SigningKey is a signing key Signer used to sign packages.
+type SigningKey Signer
+
+// SignPackageHash signs the hash and the length of a package. Use PackageHash
+// to compute the inputs.
+func (s SigningKey) SignPackageHash(hash []byte, len int64) ([]byte, error) {
+	if len <= 0 {
+		return nil, fmt.Errorf("package length must be positive, got %d", len)
+	}
+	msg := binary.LittleEndian.AppendUint64(hash, uint64(len))
+	return s.Sign(nil, msg, crypto.Hash(0))
+}
+
+// PackageHash is a hash.Hash that counts the number of bytes written. Use it
+// to get the hash and length inputs to SigningKey.SignPackageHash.
+type PackageHash struct {
+	hash.Hash
+	len int64
+}
+
+// NewPackageHash returns an initialized PackageHash using BLAKE2s.
+func NewPackageHash() *PackageHash {
+	h, err := blake2s.New256(nil)
+	if err != nil {
+		// Should never happen with a nil key passed to blake2s.
+		panic(err)
+	}
+	return &PackageHash{Hash: h}
+}
+
+func (ph *PackageHash) Write(b []byte) (int, error) {
+	ph.len += int64(len(b))
+	return ph.Hash.Write(b)
+}
+
+// Reset the PackageHash to its initial state.
+func (ph *PackageHash) Reset() {
+	ph.len = 0
+	ph.Hash.Reset()
+}
+
+// Len returns the total number of bytes written.
+func (ph *PackageHash) Len() int64 { return ph.len }
+
+// Signer is crypto.Signer using a single key (root or signing).
+type Signer struct {
+	crypto.Signer
+}
+
+// NewSigner parses the PEM-encoded private key stored in the file named
+// privKeyPath and creates a Signer for it. The key is expected to be in the
+// same format as returned by GenerateKey.
+func NewSigner(privKeyPath string) (Signer, error) {
+	raw, err := os.ReadFile(privKeyPath)
+	if err != nil {
+		return Signer{}, err
+	}
+	k, err := parsePrivateKey(raw)
+	if err != nil {
+		return Signer{}, fmt.Errorf("failed to parse %q: %w", privKeyPath, err)
+	}
+	return Signer{Signer: k}, nil
+}
+
+// Client downloads and validates files from a distribution server.
+type Client struct {
+	roots    []ed25519.PublicKey
+	pkgsAddr *url.URL
+}
+
+// NewClient returns a new client for distribution server located at pkgsAddr,
+// and uses embedded root keys from the roots/ subdirectory of this package.
+func NewClient(pkgsAddr string) (*Client, error) {
+	u, err := url.Parse(pkgsAddr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid pkgsAddr %q: %w", pkgsAddr, err)
+	}
+	return &Client{roots: roots(), pkgsAddr: u}, nil
+}
+
+func (c *Client) url(path string) string {
+	return c.pkgsAddr.JoinPath(path).String()
+}
+
+// Download fetches a file at path srcPath from pkgsAddr passed in NewClient.
+// The file is downloaded to dstPath and its signature is validated using the
+// embedded root keys. Download returns an error if anything goes wrong with
+// the actual file download or with signature validation.
+func (c *Client) Download(srcPath, dstPath string) error {
+	// Always fetch a fresh signing key.
+	sigPub, err := c.signingKeys()
+	if err != nil {
+		return err
+	}
+
+	srcURL := c.url(srcPath)
+	sigURL := srcURL + ".sig"
+
+	dstPathUnverified := dstPath + ".unverified"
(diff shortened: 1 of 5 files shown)


clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,228 @@
+// Package distsign implements signature and validation of arbitrary
+// distributable files.
+//
+// There are 3 parties in this exchange:
+//   - builder, which creates files, signs them and publishes to server
+//   - server, which distributes signing keys, files and signatures
```
noncombatant, line 9: Maybe specify public keys (to distinguish it from the secrets server)

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,228 @@
...
+// The root public keys are baked into the client software at compile time.
+// These keys are long-lived and prove the validity of current signing keys
+// from distsign.pub. To rotate root keys, a new client release must be
+// published, they are not rotated dynamically. There are multiple root keys in
+// different locations specifically to allow this rotation without using the
+// discarded root key for any new signatures.
+//
+// The signing public keys are fetched by the client dynamically before every
+// download and can be rotated more readily, assuming that most deployed
+// clients trust the root keys used to new fresh signing keys.
```
noncombatant, line 36: s/new/issue/ or similar?

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,228 @@
...
+	"os"
+)
+
+const (
+	pemTypePrivate = "PRIVATE KEY"
+	pemTypePublic  = "PUBLIC KEY"
+)
+
+// GenerateKey generates a new key pair and encodes it as PEM.
+func GenerateKey() (priv, pub []byte, retErr error) {
```
noncombatant, line 56: Random thought: You could s/retErr/err/ (which requires you to use `=` instead of `:=` on line 54).

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,228 @@
...
+			Bytes: []byte(pub),
+		}), nil
+}
+
+// Signer is a single key (root or signing) that can sign arbitrary blobs.
+type Signer struct {
+	pk ed25519.PrivateKey
+}
+
+// NewSigner parses the PEM-encoded private key and creates a Signer for it.
```
noncombatant, line 75: Nit: "...the PEM-encoded private key stored in the file named privKeyPath..." (or some other phrasing)

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,228 @@
...
+
+// Sign signs msg and returns the signature.
+func (s *Signer) Sign(msg []byte) []byte {
+	return ed25519.Sign(s.pk, msg)
+}
+
+// Client downloads and validates files from a distribution server.
+type Client struct {
+	roots    []ed25519.PublicKey
+	pkgsAddr string
```
noncombatant, line 97: Should this be `URL`?

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,228 @@
...
+	// Always fetch a fresh signing key.
+	sigPub, err := c.signingKeys()
+	if err != nil {
+		return err
+	}
+
+	srcURL := fmt.Sprintf("%s/%s", c.pkgsAddr, srcPath)
+	sigURL := srcURL + ".sig"
+
+	data, err := download(srcURL, 1<<30) // limit to 1Gb.
```
noncombatant, line 122: Can we safely set an even lower limit? Maybe even as low as 128 MiB?

noncombatant: I might also give that magic number a name, like `const ReasonablePackageSizeLimit = ...`.

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	sig, err := download(sigURL, ed25519.SignatureSize)
+	if err != nil {
+		return err
+	}
+	if !verifyAny(sigPub, data, sig) {
+		return fmt.Errorf("signature %q for key %q does not validate with the current release signing key; either you are under attack, or attempting to download an old version of Tailscale which was signed with an older signing key", srcURL, sigURL)
+	}
+
+	return os.WriteFile(dstPath, data, 0600)
```
noncombatant, line 134: Arguably could use 0644 here, since it's already public.

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,231 @@
...
+	data, err := download(srcURL, 1<<30) // limit to 1Gb.
+	if err != nil {
+		return err
+	}
+	sig, err := download(sigURL, ed25519.SignatureSize)
+	if err != nil {
+		return err
+	}
+	if !verifyAny(sigPub, data, sig) {
+		return fmt.Errorf("signature %q for key %q does not validate with the current release signing key; either you are under attack, or attempting to download an old version of Tailscale which was signed with an older signing key", srcURL, sigURL)
```
noncombatant, line 131: If I'm reading this right (might not be), this should be:

```
  return fmt.Errorf("signature %q for package %q does not...", sigURL, srcURL)
```

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,231 @@
...
+	return os.WriteFile(dstPath, data, 0600)
+}
+
+// signingKeys fetches current signing keys from the server and validates them
+// against the roots. Should be called before validation of any downloaded file
+// to get the fresh keys.
+func (c *Client) signingKeys() ([]ed25519.PublicKey, error) {
+	keyURL := c.pkgsAddr + "/distsign.pub"
+	sigURL := keyURL + ".sig"
+	raw, err := download(keyURL, 1<<20) // up to 1MB
```
noncombatant, line 143: I'd name this constant, too

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return nil, err
+	}
+	sig, err := download(sigURL, ed25519.SignatureSize)
+	if err != nil {
+		return nil, err
+	}
+	if !verifyAny(c.roots, raw, sig) {
+		return nil, fmt.Errorf("signature %q for key %q does not validate with any known root key; either you are under attack, or running a very old version of Tailscale with outdated root keys", keyURL, sigURL)
```
noncombatant, line 152: Last 2 arguments are reversed, I think.

clientupdate/distsign/distsign.go
awly: done

clientupdate/distsign/distsign.go
awly: oops, fixed

clientupdate/distsign/distsign.go
awly: done

clientupdate/distsign/distsign.go
awly: done

clientupdate/distsign/distsign.go
awly: updated to `url.URL` as type. is that what you meant?

clientupdate/distsign/distsign.go
awly: Set to 512Mb, to somewhat future-proof it for packages other than our current release builds.
This is intended as a fail-safe to avoid OOM from a malicious download, so I kept it generous.

clientupdate/distsign/distsign.go
awly: oops, fixed!

clientupdate/distsign/distsign.go
awly: updated

clientupdate/distsign/distsign.go
awly: done

clientupdate/distsign/distsign.go
awly: done

awly:
Updated PR to use `Ed25519ph` (pre-hashed version) for actual packages so that we don't have to buffer them in memory. The signing key bundle is still signed with regular `Ed25519`, it should be small enough to buffer in memory.

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,286 @@
...
+// a sequence of PEM blocks joined with newlines.
+func (s *RootKey) SignSigningKeys(pubBundle []byte) ([]byte, error) {
+	return s.Sign(nil, pubBundle, crypto.Hash(0))
+}
+
+// SigningKey is a signing key Signer used to sign packages.
+type SigningKey Signer
+
+// SignPackageHash signs a SHA-512 hash of a package.
+func (s SigningKey) SignPackageHash(sha512 []byte) ([]byte, error) {
```
noncombatant, line 91: I had a tiny thought: Perhaps we should sign the combination of the length of the package in bytes as well as the package's SHA-512. Or, I think equivalently, `SHA512(file size || file contents)` (where `||` is concatenation).

I don't expect that SHA-512 will be meaningfully weakened any time soon, but it would give us just a bit more assurance that we are signing exactly and only the right package. In my limited knowledge, attacks on hash functions (collision or pre-image) typically (but not necessarily always) require finding an input with a different length than the original input has. (For example the MD5 collisions I've seen involve creating colliding inputs by adding additional junk bytes to given files.)

And then we need to specify `binary.LittleEndian` or `BigEndian`. Little-endian seems to have won, so.

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,286 @@
...
+	"net/http"
+	"net/url"
+	"os"
+)
+
+const (
+	pemTypePrivate = "PRIVATE KEY"
+	pemTypePublic  = "PUBLIC KEY"
+
+	downloadSizeLimit    = 1 << 29 // 512Mb
```
noncombatant, line 58: Capital B to make clear(er) it's bytes, not bits

(why am I like this. I'm sorry)

clientupdate/distsign/distsign.go
noncombatant: As for how to decide on signing length || hash vs. hashing length || contents: whatever is easier, I think?

When you use `io.TeeReader` to hash the HTTP response body, you could get the expected length from the Content-Length response header. (We must make sure that's really how many bytes we read and hashed, of course.)

clientupdate/distsign/distsign_test.go
```diff
@@ -0,0 +1,346 @@
...
+	if err != nil {
+		t.Fatalf("parsePrivateKey: %v", err)
+	}
+	return signingKeyPair{
+		SigningKey: &SigningKey{Signer: priv},
+		keyPair:    kp,
+	}
+}
+
+func (s signingKeyPair) sign(blob string) []byte {
```
noncombatant, line 323: `[]byte` feels better here than `string`. I don't trust that a conversion from `[]byte` to `string` and then again into `[]byte` will be lossless, but I do trust that `string` to `[]byte` will be.

(Maybe I'm wrong? But I think `ed25519.Sign` and `rootKeyPair.sign` take `[]byte` for good reason and we should, if for no other reason, be consistent.)

awly: It was just easier to write test cases this way, but changed to `[]byte` for consistency anyway.
AFAIK in Go converting from `[]byte` to `string` and back is lossless. Is that not so in C++?

clientupdate/distsign/distsign.go
awly: Cool, is that a common technique?
We don't want to append the length because most collision attacks append garbage padding?
Appending the length is less error-prone to implement, but I understand if it's less robust.

clientupdate/distsign/distsign_test.go
noncombatant: It does seem to be lossless in Go (tested with random bytes). TIL! (I asked because Rust is, naturally, more cranky about it.)

clientupdate/distsign/distsign.go
noncombatant: Yeah, I was thinking about appending rather than pre-pending, on the grounds of error-prone-age, and I'm not sure — but appending garbage was indeed the main thing I was afraid of.

awly: Added length suffix to signed message. This meant switching from Ed25519ph to regular Ed25519 for package signatures, so we're effectively double (triple?) hashing the input.
Technically for signing keys (where we sign the entire blob instead of a hash) we don't need it, as regular Ed25519 is collision-resistant per [the RFC](https://www.rfc-editor.org/rfc/rfc8032#section-4). But I added the length there too for consistency.

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,320 @@
...
+	// Always fetch a fresh signing key.
+	sigPub, err := c.signingKeys()
+	if err != nil {
+		return err
+	}
+
+	srcURL := c.url(srcPath)
+	sigURL := srcURL + ".sig"
+
+	hash, len, err := download(srcURL, dstPath, downloadSizeLimit)
```
twitchyliquid64, line 182: If the program crashes after the download but before the signature verification, will this result in the destination file remaining on disk?

If so, maybe we want to download to dest + `.unverified` and then atomic-rename after the file is verified?

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,320 @@
...
+	if b.Type != pemTypePublic {
+		return nil, nil, fmt.Errorf("PEM type is %q, want %q", b.Type, pemTypePublic)
+	}
+	if len(b.Bytes) != ed25519.PublicKeySize {
+		return nil, nil, errors.New("public key has incorrect length for an Ed25519 public key")
+	}
+	return ed25519.PublicKey(b.Bytes), rest, nil
+}
+
+func verifyAny(keys []ed25519.PublicKey, msg, sig []byte) bool {
```
twitchyliquid64, line 313: Add a comment that verifyAny will panic if any public key has the wrong len.

I see you are already checking the len elsewhere :+1:

clientupdate/distsign/distsign.go
```diff
@@ -0,0 +1,320 @@
...
+	for _, k := range keys {
+		if ed25519.Verify(k, msg, sig) {
```
twitchyliquid64, line 315: golang signatures+Verify() arent malleable, but I think its worth using a more-defined verification algorithm if we ever expect to have non-go code verifying the signature. For tailnet lock we used ZIP15 / ed25519consensus package.

clientupdate/distsign/distsign.go
twitchyliquid64: If we are worried about SHA-512 we could use blake2 for some diversity. We already use it for tailnet lock so it wont add to the binary size, and ed25519 already hashes the input for sign/verify using sha-512 so we end up with both being used.

clientupdate/distsign/distsign.go
awly: TIL!

twitchyliquid64 (approved):

PR #8945, opened 2023-08-18
all: remove some Debug fields, NetworkMap.Debug, Reconfig Debug arg

bradfitz:
Updates #8923

### DIFF
--- a/wgengine/userspace.go
+++ b/wgengine/userspace.go
@@ -766,7 +766,9 @@ func hasOverlap(aips, rips []netip.Prefix) bool {
 	return false
 }
 
-func (e *userspaceEngine) Reconfig(cfg *wgcfg.Config, routerCfg *router.Config, dnsCfg *dns.Config, debug *tailcfg.Debug) error {
+var randomizeClientPort = controlclient.RandomizeClientPort
+
+func (e *userspaceEngine) Reconfig(cfg *wgcfg.Config, routerCfg *router.Config, dnsCfg *dns.Config) error {
 	if routerCfg == nil {
 		panic("routerCfg must not be nil")
 	}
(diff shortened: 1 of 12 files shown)


bradfitz:
This neuters a test, though, so I kinda want to figure out how to restore it.

wgengine/userspace.go
```diff
@@ -792,8 +792,7 @@ func (e *userspaceEngine) Reconfig(cfg *wgcfg.Config, routerCfg *router.Config,
 	e.mu.Unlock()
 
 	listenPort := e.confListenPort
-	if controlclient.RandomizeClientPort() || // new way (capver 70)
-		debug != nil && debug.RandomizeClientPort { // old way which server might still send (as of 2023-08-16)
+	if controlclient.RandomizeClientPort() {
```
maisem, line 795: could add a var fn in userspace like

```
var randomizeClientPort = controlclient.RandomizeClientPort
```

and in test
```
tstest.Replace(t, &randomizeClientPort, func() bool {return true})
```

wgengine/userspace_test.go
```diff
@@ -181,21 +183,24 @@ func TestUserspaceEnginePortReconfig(t *testing.T) {
...
+	if err := ue.Reconfig(cfg, routerCfg, &dns.Config{}); err != nil {
 		t.Fatal(err)
 	}
 	if got := ue.magicConn.LocalPort(); got == startingPort {
 		t.Errorf("debug setting did not change local port from %d", startingPort)
 	}
 
 	lastPort := ue.magicConn.LocalPort()
-	if err := ue.Reconfig(cfg, routerCfg, &dns.Config{}, nil); err != nil {
+	randomizeClientPort = func() bool { return false }
```
maisem, line 202: use tstest.Replace so that it gets unset after the test is finished. Otherwise we will spend hours debugging a test failure only to discover this :P

maisem (approved):

wgengine/userspace_test.go
bradfitz: I did at the top already.

wgengine/userspace_test.go
```diff
@@ -143,6 +143,8 @@ func TestUserspaceEngineReconfig(t *testing.T) {
 }
 
 func TestUserspaceEnginePortReconfig(t *testing.T) {
+	tstest.Replace(t, &randomizeClientPort, func() bool { return false })
```
bradfitz, line 146: yo @maisem up here

wgengine/userspace_test.go
maisem: oh i see, sure

PR #8946, opened 2023-08-18
net/interfaces: fix android "route ip+net: netlinkrib: permission denied"

Asutorufa:


DentonGentry (commented):
To be clear: you are suggesting this to replace https://github.com/tailscale/tailscale-android/pull/21, which we implemented two years ago?

You're suggesting this be used until https://github.com/golang/go/issues/40569 is fixed?

PR #8947, opened 2023-08-18
wgengine/router: fall back and set iptables as default again

KevinLiang10:
Due to the conflict between our nftables implementation and ufw, which is a common utility used on linux. We now want to take a step back to prevent regression. This will give us more chance to let users to test our nftables support and heuristic.

Updates: #391

### DIFF
--- a/wgengine/router/router_linux.go
+++ b/wgengine/router/router_linux.go
@@ -85,41 +85,32 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
 	iptAva, nftAva := true, true
 	iptRuleCount, err := det.iptDetect()
 	if err != nil {
-		logf("router: detect iptables rule: %v", err)
+		logf("detect iptables rule: %v", err)
 		iptAva = false
 	}
 	nftRuleCount, err := det.nftDetect()
 	if err != nil {
-		logf("router: detect nftables rule: %v", err)
+		logf("detect nftables rule: %v", err)
 		nftAva = false
 	}
-	logf("router: nftables rule count: %d, iptables rule count: %d", nftRuleCount, iptRuleCount)
+	logf("nftables rule count: %d, iptables rule count: %d", nftRuleCount, iptRuleCount)
 	switch {
-	case envknob.String("TS_DEBUG_FIREWALL_MODE") == "nftables":
-		// TODO(KevinLiang10): Updates to a flag
-		logf("router: envknob TS_DEBUG_FIREWALL_MODE=nftables set")
-		hostinfo.SetFirewallMode("nft-forced")
-		return linuxfw.FirewallModeNfTables
-	case envknob.String("TS_DEBUG_FIREWALL_MODE") == "iptables":
-		logf("router: envknob TS_DEBUG_FIREWALL_MODE=iptables set")
-		hostinfo.SetFirewallMode("ipt-forced")
-		return linuxfw.FirewallModeIPTables
 	case nftRuleCount > 0 && iptRuleCount == 0:
-		logf("router: nftables is currently in use")
+		logf("nftables is currently in use")
 		hostinfo.SetFirewallMode("nft-inuse")
 		return linuxfw.FirewallModeNfTables
 	case iptRuleCount > 0 && nftRuleCount == 0:
-		logf("router: iptables is currently in use")
+		logf("iptables is currently in use")
 		hostinfo.SetFirewallMode("ipt-inuse")
 		return linuxfw.FirewallModeIPTables
 	case nftAva:
 		// if both iptables and nftables are available but
 		// neither/both are currently used, use nftables.
-		logf("router: nftables is available")
+		logf("nftables is available")
 		hostinfo.SetFirewallMode("nft")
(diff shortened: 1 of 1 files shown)


wgengine/router/router_linux.go
```diff
@@ -136,7 +127,24 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
...
+	switch {
+	case envknob.String("TS_DEBUG_FIREWALL_MODE") == "nftables":
+		logf("router: envknob TS_DEBUG_FIREWALL_MODE=nftables set")
+		hostinfo.SetFirewallMode("nft-forced")
+		mode = linuxfw.FirewallModeNfTables
+
+	case envknob.String("TS_DEBUG_FIREWALL_MODE") == "auto":
+		mode = chooseFireWallMode(logf, tableDetector)
+	default:
+		logf("router: envknob TS_DEBUG_FIREWALL_MODE=iptables set")
```
bradfitz, line 143: It's not necessarily set to this. This is the default case.

wgengine/router/router_linux.go
```diff
@@ -136,7 +127,24 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
...
+		hostinfo.SetFirewallMode("ipt-forced")
```
bradfitz, line 144: How about "fallback" instead of "forced"?

wgengine/router/router_linux.go
KevinLiang10: Oh yeah need to change the log.

wgengine/router/router_linux.go
KevinLiang10: Yeah updating these, was just trying to look at the pipelines

KevinLiang10: I would suggest using "ipt-default"? We are trying to use iptables when nothing is done by user, "fallback" sounds like we tried to do something else but failed so we fallback?

bradfitz: default is fine too

bradfitz:
LGTM at least but I'll let others approve.

wgengine/router/router_linux.go
```diff
@@ -95,15 +95,6 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
 	}
 	logf("router: nftables rule count: %d, iptables rule count: %d", nftRuleCount, iptRuleCount)
 	switch {
-	case envknob.String("TS_DEBUG_FIREWALL_MODE") == "nftables":
-		// TODO(KevinLiang10): Updates to a flag
-		logf("router: envknob TS_DEBUG_FIREWALL_MODE=nftables set")
-		hostinfo.SetFirewallMode("nft-forced")
-		return linuxfw.FirewallModeNfTables
-	case envknob.String("TS_DEBUG_FIREWALL_MODE") == "iptables":
```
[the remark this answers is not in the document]
catzkorn: I agree with this, since we are making a change we still want to make sure we are aware of the configurations that were set for 1.48. It will help support out.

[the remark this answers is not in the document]
catzkorn: Oh it got moved further down, ignore me!

wgengine/router/router_linux.go
```diff
@@ -136,7 +127,28 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
...
+	var mode linuxfw.FirewallMode
+
+	// We now use iptables as default and have "auto" and "nftables" as
+	// options for people to test further.
+	switch {
+	case envknob.String("TS_DEBUG_FIREWALL_MODE") == "nftables":
+		logf("router: envknob TS_DEBUG_FIREWALL_MODE=nftables set")
+		hostinfo.SetFirewallMode("nft-forced")
+		mode = linuxfw.FirewallModeNfTables
+
```
catzkorn, line 139: NIT: remove empty line

wgengine/router/router_linux.go
raggi: You could use "set" rather than "forced", but I don't feel strongly about it.

raggi (commented):
There was a distro exception added for Gokrazy, that per the comment we should probably carry forward.

raggi (approved):

wgengine/router/router_linux.go
```diff
@@ -136,7 +127,33 @@ func chooseFireWallMode(logf logger.Logf, det tableDetector) linuxfw.FirewallMod
...
-	mode := chooseFireWallMode(logf, tableDetector)
+	var mode linuxfw.FirewallMode
+
+	// We now use iptables as default and have "auto" and "nftables" as
+	// options for people to test further.
+	switch {
+	case distro.Get() == distro.Gokrazy:
+		// Reduce startup logging on gokrazy. There's no way to do iptables on
+		// gokrazy anyway.
+		logf("router: GoKrazy should use nftables.")
```
raggi, line 138: here and elsewhere, drop the router prefix, it's added in the logf in router.New(), so having it here creates `router: router: ` prefixes.

Issue #8948, opened 2023-08-18
all: use NodeView inside tailscaled


bradfitz:
As part of working on #1909, we want to use views more inside tailscaled. In particular, the `NetworkMap.Peers []*tailcfg.Node` is expensive to defensively clone all the time.

Fortunately we already generate NodeView in the tailcfg package. It just isn't widely used yet. Tracking bug to do so.

/cc @maisem

PR #8949, opened 2023-08-18
types/views: add LenIter method to slice view types

bradfitz:
This is basically https://github.com/bradfitz/iter which was
a joke but now that Go's adding range over int soonish, might
as well. It simplies our code elsewher that uses slice views.

Updates #8948


maisem (approved):
Why doesn't it allocate?

bradfitz:
> Why doesn't it allocate?

Because the value type `struct{}` is zero-length. No need to allocate a backing array. All the info (len + cap) fits in the slice header.

See https://github.com/bradfitz/iter/blob/master/iter_test.go

PR #8961, opened 2023-08-19
types/netmap, all: use read-only tailcfg.NodeView in NetworkMap

bradfitz:
Updates #8948

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -204,7 +204,7 @@ type LocalBackend struct {
 	// netMap is not mutated in-place once set.
 	netMap           *netmap.NetworkMap
 	nmExpiryTimer    tstime.TimerController // for updating netMap on node expiry; can be nil
-	nodeByAddr       map[netip.Addr]*tailcfg.Node
+	nodeByAddr       map[netip.Addr]tailcfg.NodeView
 	activeLogin      string // last logged LoginName from netMap
 	engineStatus     ipn.EngineStatus
 	endpoints        []tailcfg.Endpoint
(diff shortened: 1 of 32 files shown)


ipn/ipnlocal/local.go
```diff
@@ -2438,36 +2442,37 @@ func (b *LocalBackend) Ping(ctx context.Context, ip netip.Addr, pingType tailcfg
 	}
 }
 
-func (b *LocalBackend) pingPeerAPI(ctx context.Context, ip netip.Addr) (peer *tailcfg.Node, peerBase string, err error) {
+func (b *LocalBackend) pingPeerAPI(ctx context.Context, ip netip.Addr) (peer tailcfg.NodeView, peerBase string, err error) {
+	var zero tailcfg.NodeView
```
maisem, line 2446: I've been calling this noNode/noUser in a few places.

net/tsdial/dnsmap_test.go
```diff
@@ -12,6 +12,14 @@ import (
 	"tailscale.com/types/netmap"
 )
 
+func nodeViews(v []*tailcfg.Node) []tailcfg.NodeView {
+	nv := make([]tailcfg.NodeView, len(v))
```
maisem, line 16: Could also do
views.SliceOfViews(v).AsSlice()

maisem: Considering it's repeated a few times

maisem (approved):

ipn/ipnlocal/local.go
bradfitz: I'm getting ready for https://github.com/golang/go/issues/61372 in which cas we just need to delete this declaration line and it'll keep working.

net/tsdial/dnsmap_test.go
bradfitz: I'm too dumb for that.

I can't follow these docs:

```go
// SliceOfViews returns a ViewSlice for x.
func SliceOfViews[T ViewCloner[T, V], V StructView[T]](x []T) SliceView[T, V] {
	return SliceView[T, V]{x}
}

// SliceView is a read-only wrapper around a struct which should only be exposed
// as a View.
type SliceView[T ViewCloner[T, V], V StructView[T]] struct {
```

"a struct which should only be exposed as a View".

I have no structs here?

PR #8967, opened 2023-08-19
ipn/ipnstate: add PeerStatus.AltSharerUserID, stop mangling Node.User

bradfitz:
In b987b2ab18ff48 (2021-01-12) when we introduced sharing we mapped
the sharer to the userid at a low layer, mostly to fix the display of
"tailscale status" and the client UIs, but also some tests.

The commit earlier today, 7dec09d1693, removed the 2.5yo option
to let clients disable that automatic mapping, as clearly we were never
getting around to it.

This plumbs the Sharer UserID all the way to ipnstatus so the CLI
itself can choose to print out the Sharer's identity over the node's
original owner.

Then we stop mangling Node.User and let clients decide how they want
to render things.

To ease the migration for the Windows GUI (which currently operates on
tailcfg.Node via the NetMap from WatchIPNBus, instead of PeerStatus),
a new method Node.SharerOrUser is added to do the mapping of
Sharer-else-User.

Updates #1909
Updates tailscale/corp#1183

### DIFF
--- a/ipn/ipnstate/ipnstate.go
+++ b/ipn/ipnstate/ipnstate.go
@@ -199,6 +199,10 @@ type PeerStatus struct {
 	OS      string // HostInfo.OS
 	UserID  tailcfg.UserID
 
+	// AltSharerUserID is the user who shared this node
+	// if it's different than UserID. Otherwise it's zero.
+	AltSharerUserID tailcfg.UserID `json:",omitempty"`
+
 	// TailscaleIPs are the IP addresses assigned to the node.
 	TailscaleIPs []netip.Addr
 
@@ -386,6 +390,9 @@ func (sb *StatusBuilder) AddPeer(peer key.NodePublic, st *PeerStatus) {
 	}
 	if v := st.UserID; v != 0 {
 		e.UserID = v
+	}
+	if v := st.AltSharerUserID; v != 0 {
+		e.AltSharerUserID = v
 	}
 	if v := st.TailscaleIPs; v != nil {
 		e.TailscaleIPs = v
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -21,6 +21,7 @@ import (
 	"tailscale.com/types/opt"
 	"tailscale.com/types/structs"
 	"tailscale.com/types/tkatype"
+	"tailscale.com/util/cmpx"
 	"tailscale.com/util/dnsname"
 )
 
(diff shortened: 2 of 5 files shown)


tailcfg/tailcfg.go
```diff
@@ -223,9 +224,18 @@ type Node struct {
...
+	// as.
+	//
+	// As of 2021-01-13 (c1dabd94), the control client rewrites this field to be
+	// Sharer for clients & tests that didn't yet support Sharer as a concept.
+	// We want to move towards this always meaning User, but in the meantime,
+	// the ActualUser field (if set) is the unmodified User value. The plan (as
+	// of 2023-08-19) is for users to move to calling GetUser which returns
+	// either ActualUser if set, else User. Once we've validated all the clients
+	// and tests work with the Sharer-to-User rewrite removed, then we can
+	// remove ActualUser and GetUser to just access User.
```
maisem, line 238: These don't exist

ipn/ipnstate/ipnstate.go
```diff
@@ -199,6 +199,10 @@ type PeerStatus struct {
 	OS      string // HostInfo.OS
 	UserID  tailcfg.UserID
 
+	// AltSharerUserID is the user who shared this node
+	// if it's different than UserID. Otherwise it's zero.
+	AltSharerUserID tailcfg.UserID `json:",omitempty"`
```
maisem, line 204: Why Alt? Why not just SharerUserID

tailcfg/tailcfg.go
bradfitz: Whoops, deleted. Old approach I abandoned.

ipn/ipnstate/ipnstate.go
bradfitz: Because the name "SharerUserID" (which I started with) is super misleading however it's documented. It makes it sound like it's always non-zero if the node was shared, which it's not.

The whole zero-means-same-as-UserID thing is a network optimization at best but sucks as an API.

Because PeerState is more of the API that end users will see (since tailcfg.Node is increasingly internal, as we remove it from IPNBus), I wanted to be super explicit in its name, crying out that it's weird.

bradfitz: If we knew for sure the node was shared to us, then I could always set SharerUserID (and name it as such) but there's no bit we have on the wire that says that the node was shared.

maisem (approved):

PR #8972, opened 2023-08-20
types/tkatype: change MarshaledSignature from []byte to string

bradfitz:
The MarshaledSignature's underlying type of []byte didn't play nicely with cmd/viewer, which mapped the []byte to a mem.RO handle which was then clumsy to work with.

Instead, change it to a string. (But keep its JSON encoding for compatibility)

Updates #1909

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -6,7 +6,6 @@ package tailcfg
 //go:generate go run tailscale.com/cmd/viewer --type=User,Node,Hostinfo,NetInfo,Login,DNSConfig,RegisterResponse,RegisterResponseAuth,RegisterRequest,DERPHomeParams,DERPRegion,DERPMap,DERPNode,SSHRule,SSHAction,SSHPrincipal,ControlDialPlan,Location,UserProfile --clonefunc
 
 import (
-	"bytes"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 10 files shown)


bradfitz:
Thoughts, @twitchyliquid64?

Note that #8968 was already submitted (which doesn't hurt regardless of whether this change goes in) and that test continues to pass with this change.

The motivation for this is that we're converting tailscaled to use views internally, but TKA then gets super clunky to deal with this type.

tailcfg/tailcfg.go
```diff
@@ -1018,7 +1017,7 @@ func (req *RegisterRequest) Clone() *RegisterRequest {
 	}
 	res.DeviceCert = append(res.DeviceCert[:0:0], res.DeviceCert...)
 	res.Signature = append(res.Signature[:0:0], res.Signature...)
-	res.NodeKeySignature = append(res.NodeKeySignature[:0:0], res.NodeKeySignature...)
+	res.NodeKeySignature = res.NodeKeySignature
```
maisem, line 1020: hah, this is still manually done. sent https://github.com/tailscale/tailscale/pull/8973

maisem (approved):

twitchyliquid64:
I think this will work fine, bit I'll take a true-blue look on Monday.

bradfitz:
The third option is we make cmd/viewer smarter on `[]byte` types and instead of returning `mem.RO` (kinda a cop-out), we make it return a new FooView type (like it does for everything else) that feels like a `mem.RO` but has extra type-specific useful methods so it's not clunky to use.

I'll see how painful that is instead. Because while I don't mind this change, it'd be a pity to have to always do this if we grow more `[]byte` types in the future.

bradfitz:
Obsoleted by #8976

PR #8976, opened 2023-08-20
types/views,cmd/viewer: add ByteSlice[T] to replace mem.RO

maisem:
Add a new views.ByteSlice[T ~[]byte] to provide a better API to use with views.

Updates #cleanup

### DIFF
--- a/types/views/views.go
+++ b/types/views/views.go
@@ -6,9 +6,12 @@
 package views
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"maps"
+
+	"go4.org/mem"
 )
 
 func unmarshalSliceFromJSON[T any](b []byte, x *[]T) error {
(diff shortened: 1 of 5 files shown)


types/views/views.go
```diff
@@ -21,6 +24,80 @@ func unmarshalSliceFromJSON[T any](b []byte, x *[]T) error {
...
+func (v ByteSlice[T]) Equal(b T) bool {
+	return bytes.Equal(v.ж, b)
+}
+
+// EqualView reports whether the underlying slice is equal to b.
+func (v ByteSlice[T]) EqualView(b ByteSlice[T]) bool {
+	return bytes.Equal(v.ж, b.ж)
+}
+
+// AsSlice returns a deep-copy of the underlying slice.
```
bradfitz, line 66: "deep"? It's a byte slice by definition.

maisem: Copy paste fail, will fix

types/views/views.go
```diff
@@ -21,6 +24,80 @@ func unmarshalSliceFromJSON[T any](b []byte, x *[]T) error {
...
+func (v ByteSlice[T]) AsSlice() T {
+	return v.AppendTo(v.ж[:0:0])
+}
+
+// AppendTo appends the underlying slice values to dst.
```
bradfitz, line 71: You have all of Mem.RO almost but not At. Add At too?

bradfitz (commented):
Nice.

bradfitz (approved):

twitchyliquid64:
Yeah I like this better, cheers

PR #8978, opened 2023-08-20
types/netmap: move some mutations earlier, remove, document some fields

bradfitz:
And optimize the Persist setting a bit, allocating later and only mutating
fields when there's been a Node change.

Updates #1909

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -54,6 +54,7 @@ import (
 	"tailscale.com/types/ptr"
 	"tailscale.com/types/tkatype"
 	"tailscale.com/util/clientmetric"
+	"tailscale.com/util/cmpx"
 	"tailscale.com/util/multierr"
 	"tailscale.com/util/singleflight"
 	"tailscale.com/util/systemd"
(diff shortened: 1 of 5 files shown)


control/controlclient/direct.go
```diff
@@ -1144,6 +1150,23 @@ func (c *Direct) sendMapRequest(ctx context.Context, isStreaming bool, nu Netmap
...
+// The magicDNSSuffix used is based on selfNode.
+func initDisplayNames(selfNode tailcfg.NodeView, resp *tailcfg.MapResponse) {
+	if resp.Node == nil && len(resp.PeersChanged) == 0 {
+		return
+	}
+	magicDNSSuffix := netmap.MagicDNSSuffixOfNodeName(selfNode.Name())
+	if resp.Node != nil {
+		resp.Node.InitDisplayNames(magicDNSSuffix)
+	}
+	for _, n := range resp.PeersChanged {
```
maisem, line 1165: you need to do this on resp.Peers too

bradfitz: done

types/netmap/netmap.go
```diff
@@ -66,10 +76,6 @@ type NetworkMap struct {
 	// hash of the latest update message to tick through TKA).
 	TKAHead tka.AUMHash
 
-	// ACLs
-
-	User tailcfg.UserID
```
maisem, line 71: would this break GUIs?

bradfitz: It broke Windows but I adjusted it in https://github.com/tailscale/corp/pull/14065

bradfitz: All xcode builds seem happy locally.

maisem (approved):

PR #8979, opened 2023-08-20
cmd/sniproxy: add port forwarding and prometheus metrics

DentonGentry:
1. Add TCP port forwarding. For example: ./sniproxy -forwards=tcp/22/github.com will forward SSH to github.

   % ssh -i ~/.ssh/id_ecdsa.pem -T git@github.com
   Hi GitHubUser! You've successfully authenticated, but GitHub does not provide shell access.

   % ssh -i ~/.ssh/id_ecdsa.pem -T git@100.65.x.y
   Hi GitHubUser! You've successfully authenticated, but GitHub does not provide shell access.

2. Additionally export clientmetrics as prometheus metrics for local scraping over the tailnet: http://sniproxy-hostname:8080/debug/varz

### DIFF
--- a/cmd/sniproxy/sniproxy.go
+++ b/cmd/sniproxy/sniproxy.go
@@ -3,15 +3,20 @@
 
 // The sniproxy is an outbound SNI proxy. It receives TLS connections over
 // Tailscale on one or more TCP ports and sends them out to the same SNI
-// hostname & port on the internet. It only does TCP.
+// hostname & port on the internet. It can optionally forward one or more
+// TCP ports to a specific destination. It only does TCP.
 package main
 
 import (
 	"context"
+	"errors"
+	"expvar"
 	"flag"
+	"fmt"
 	"log"
 	"net"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -19,27 +24,54 @@ import (
 	"inet.af/tcpproxy"
 	"tailscale.com/client/tailscale"
 	"tailscale.com/hostinfo"
+	"tailscale.com/metrics"
 	"tailscale.com/net/netutil"
 	"tailscale.com/tsnet"
+	"tailscale.com/tsweb"
 	"tailscale.com/types/nettype"
 	"tailscale.com/util/clientmetric"
 )
 
 var (
 	ports        = flag.String("ports", "443", "comma-separated list of ports to proxy")
+	forwards     = flag.String("forwards", "", "comma-separated list of ports to transparently forward, protocol/number/destination. For example, --forwards=tcp/22/github.com,tcp/5432/sql.example.com")
 	wgPort       = flag.Int("wg-listen-port", 0, "UDP port to listen on for WireGuard and peer-to-peer traffic; 0 means automatically select")
 	promoteHTTPS = flag.Bool("promote-https", true, "promote HTTP to HTTPS")
+	debugPort    = flag.Int("debug-port", 8080, "Listening port for debug/metrics endpoint")
 )
 
 var tsMBox = dnsmessage.MustNewName("support.tailscale.com.")
 
-var (
-	numSessions    = clientmetric.NewCounter("sniproxy_sessions")
-	numBadAddrPort = clientmetric.NewCounter("sniproxy_bad_addrport")
-	dnsResponses   = clientmetric.NewCounter("sniproxy_dns_responses")
-	dnsFailures    = clientmetric.NewCounter("sniproxy_dns_failed")
-	httpPromoted   = clientmetric.NewCounter("sniproxy_http_promoted")
-)
+// portForward is the state for a single port forwarding entry, as passed to the --forward flag.
+type portForward struct {
+	Port        int
+	Proto       string
+	Destination string
+}
+
+// parseForward takes a proto/port/destination tuple as an input, as would be passed
+// to the --forward command line flag, and returns a *portForward struct of those parameters.
+func parseForward(value string) (*portForward, error) {
+	parts := strings.Split(value, "/")
+	if len(parts) != 3 {
+		return nil, errors.New("cannot parse: " + value)
+	}
+
+	proto := parts[0]
+	if proto != "tcp" {
+		return nil, errors.New("unsupported forwarding protocol: " + proto)
+	}
+	port, err := strconv.ParseUint(parts[1], 10, 16)
+	if err != nil {
+		return nil, errors.New("bad forwarding port: " + parts[1])
+	}
+	host := parts[2]
+	if host == "" {
+		return nil, errors.New("bad destination: " + value)
+	}
+
+	return &portForward{Port: int(port), Proto: proto, Destination: host}, nil
+}
 
 func main() {
 	flag.Parse()
@@ -58,6 +90,7 @@ func main() {
 		log.Fatal(err)
 	}
 	s.lc = lc
+	s.initMetrics()
 
 	for _, portStr := range strings.Split(*ports, ",") {
 		ln, err := s.ts.Listen("tcp", ":"+portStr)
@@ -66,6 +99,34 @@ func main() {
 		}
 		log.Printf("Serving on port %v ...", portStr)
 		go s.serve(ln)
+	}
+
+	for _, forwStr := range strings.Split(*forwards, ",") {
+		if forwStr == "" {
+			continue
+		}
+		forw, err := parseForward(forwStr)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		ln, err := s.ts.Listen("tcp", ":"+strconv.Itoa(forw.Port))
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Printf("Serving on port %d to %s...", forw.Port, forw.Destination)
+
+		// Add an entry to the expvar LabelMap for Prometheus metrics,
+		// and create a clientmetric to report that same value.
+		service := portNumberToName(forw)
+		s.numTCPsessions.SetInt64(service, 0)
+		metric := fmt.Sprintf("sniproxy_tcp_sessions_%s", service)
+		clientmetric.NewCounterFunc(metric, func() int64 {
+			return s.numTCPsessions.Get(service).Value()
+		})
+
+		go s.forward(ln, forw)
+
 	}
 
 	ln, err := s.ts.Listen("udp", ":53")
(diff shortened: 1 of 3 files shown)


cmd/sniproxy/sniproxy.go
```diff
@@ -68,6 +95,34 @@ func main() {
 		go s.serve(ln)
 	}
 
+	for _, forwStr := range strings.Split(*forwards, ",") {
+		if forwStr == "" {
+			break
```
creachadair, line 100: continue? (I suspect this is here to handle `*forwards == ""`, otherwise I'd suggest it should be an error)

cmd/sniproxy/sniproxy.go
```diff
@@ -3,43 +3,69 @@
...
 var tsMBox = dnsmessage.MustNewName("support.tailscale.com.")
 
-var (
-	numSessions    = clientmetric.NewCounter("sniproxy_sessions")
-	numBadAddrPort = clientmetric.NewCounter("sniproxy_bad_addrport")
-	dnsResponses   = clientmetric.NewCounter("sniproxy_dns_responses")
-	dnsFailures    = clientmetric.NewCounter("sniproxy_dns_failed")
-	httpPromoted   = clientmetric.NewCounter("sniproxy_http_promoted")
-)
+type PortForward struct {
```
creachadair, line 45: Nit: PortForward and ParseForward can be unexported.

But even if not, they deserve some docs IMO 🙂

cmd/sniproxy/sniproxy.go
```diff
@@ -83,12 +138,35 @@ func main() {
...
+		mux := http.NewServeMux()
+		tsweb.Debugger(mux)
+		srv := &http.Server{
+			Handler: mux,
+		}
+		dln, err := s.ts.Listen("tcp", fmt.Sprintf(":%d", *debugPort))
+		if err != nil {
+			log.Fatal(err)
+		}
+		go func() {
```
creachadair, line 151: I don't think we need an explicit `http.Server` here:
```go
go http.Serve(dln, mux)
```

(we also don't really care about the exit status, since it's just serving debug endpoints and it doesn't control the process lifecycle)

cmd/sniproxy/sniproxy_test.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func TestPortForwardingArguments(t *testing.T) {
+	tests := []struct {
+		in      string
+		wanterr string
+		want    *PortForward
+	}{
+		{"", "", nil},
+		{"bad port specifier", "cannot parse", nil},
```
creachadair, line 19: Some other cases to consider: `"tcp/pdq/example.com"` (non-numeric port); `"tcp/2112/"` (empty host).

cmd/sniproxy/sniproxy_test.go
```diff
@@ -0,0 +1,33 @@
...
+		{"tcp/22/github.com", "", &PortForward{Proto: "tcp", Port: 22, Destination: "github.com"}},
+	}
+	for _, tt := range tests {
+		got, goterr := ParseForward(tt.in)
+		if tt.wanterr != "" {
+			if !strings.Contains(goterr.Error(), tt.wanterr) {
+				t.Errorf("f(%q).err = %v; want %v", tt.in, goterr, tt.wanterr)
+			}
+		}
+		if !reflect.DeepEqual(got, tt.want) {
```
creachadair, line 29: Consider instead:
```go
if diff := cmp.Diff(got, tt.want); diff != "" {
   t.Errorf("Parsed forward (-got, +want):\n%s", diff)
}
```

or words to that effect.

cmd/sniproxy/sniproxy.go
```diff
@@ -83,12 +138,35 @@ func main() {
...
+	}
+
 	select {}
 }
 
 type server struct {
 	ts tsnet.Server
 	lc *tailscale.LocalClient
+
+	stats          *metrics.Set
```
creachadair, line 163: This does not seem to be used outside `initMetrics`, it can be a local variable there.

cmd/sniproxy/sniproxy.go
```diff
@@ -83,12 +138,35 @@ func main() {
...
+	numTLSsessions *expvar.Int
```
creachadair, line 164: (optional) Since we pass the server around by pointer anyway, we could also store these by value and just take their address in `initMetrics` (then we don't need to initialize them explicitly)

cmd/sniproxy/sniproxy.go
```diff
@@ -235,8 +366,41 @@ func (s *server) dnsResponse(req *dnsmessage.Message) (buf []byte, err error) {
...
+	clientmetric.NewCounterFunc("sniproxy_tls_sessions", func() int64 { return s.numTLSsessions.Value() })
+
+	s.numTCPsessions = &metrics.LabelMap{Label: "proto"}
+	s.stats.Set("tcp_sessions", s.numTCPsessions)
+	// clientmetric doesn't have a good way to implement a Map type.
+	// We create clientmetrics dynamically when parsing the --forwards argument
+
+	s.numBadAddrPort = &expvar.Int{}
+	s.stats.Set("bad_addrport", s.numBadAddrPort)
+	clientmetric.NewCounterFunc("sniproxy_bad_addrport", func() int64 { return s.numBadAddrPort.Value() })
```
creachadair, line 391: I think this could more simply be:
```suggestion
	clientmetric.NewCounterFunc("sniproxy_bad_addrport", s.numBadAddrPort.Value)
```
and so on below (the `Value` method already has this signature)

cmd/sniproxy/sniproxy.go
```diff
@@ -235,8 +366,41 @@ func (s *server) dnsResponse(req *dnsmessage.Message) (buf []byte, err error) {
...
+		s.httpPromoted.Add(1)
 		http.Redirect(w, r, "https://"+r.Host+r.RequestURI, http.StatusFound)
 	}))
 	log.Fatalf("promoteHTTPS http.Serve: %v", err)
 }
+
+// initMetrics sets up local prometheus metrics, and creates clientmetrics to report those
+// same counters.
+func (s *server) initMetrics() {
+	s.stats = new(metrics.Set)
```
creachadair, line 378:
```go
stats := …
// ...
expvar.Publish("sniproxy", stats)
```

cmd/sniproxy/sniproxy.go
```diff
@@ -174,6 +262,49 @@ func (s *server) serveConn(c net.Conn) {
 	p.Start()
 }
 
+// portNumberToName returns a human-readable name for several port numbers commonly forwarded,
+// and "tcp###" for everything else. It is used for metric label names.
+func (s *server) portNumberToName(forw *PortForward) string {
```
creachadair, line 267: This could be a plain top-level function, it doesn't use any state of `*server`.

cmd/sniproxy/sniproxy.go
```diff
@@ -174,6 +262,49 @@ func (s *server) serveConn(c net.Conn) {
...
+	case 3389:
+		return "rdp"
+	case 5432:
+		return "postgres"
+	default:
+		return fmt.Sprintf("%s%d", forw.Proto, forw.Port)
+	}
+}
+
+// forwardConn sets up a forwarder for a TCP connection. It does not inspection of the data
```
creachadair, line 284: …does not inspect…

cmd/sniproxy/sniproxy.go
```diff
@@ -174,6 +262,49 @@ func (s *server) serveConn(c net.Conn) {
...
+// like the SNI forwarding does, merely forwards all data to the destination specified in the
```
creachadair, line 285: …it merely…

creachadair (commented):
I'm not entirely clear on why this is part of the SNI proxy, since as far as I can tell we aren't restricting the target of forwards to the SNI (? or are we), but mechanically this seems fine modulo some small things noted in my comments.

cmd/sniproxy/sniproxy.go
```diff
@@ -3,43 +3,69 @@
...
+	parts := strings.Split(value, "/")
+	if len(parts) != 3 {
+		return nil, errors.New("cannot parse: " + value)
+	}
+
+	proto := parts[0]
+	if proto != "tcp" {
+		return nil, errors.New("unsupported forwarding protocol: " + proto)
+	}
+	port, err := strconv.Atoi(parts[1])
```
bradfitz, line 61: ParseUint(parts[1], 10, 16) ?

cmd/sniproxy/sniproxy.go
DentonGentry: Done

cmd/sniproxy/sniproxy.go
DentonGentry: Done

cmd/sniproxy/sniproxy.go
DentonGentry: Done.

Left the exit status because this is intended to be run within a user tailnet, and it will be the user who is scraping the metrics for their own monitoring. If the metrics export fails I think we do want to fail the service.

cmd/sniproxy/sniproxy_test.go
DentonGentry: Added cases.

cmd/sniproxy/sniproxy.go
DentonGentry: Done

cmd/sniproxy/sniproxy.go
DentonGentry: Done, except for the LabelMap which takes an argument in its initialization.

cmd/sniproxy/sniproxy.go
DentonGentry: Done

cmd/sniproxy/sniproxy.go
DentonGentry: Done

cmd/sniproxy/sniproxy.go
DentonGentry: Done

cmd/sniproxy/sniproxy.go
DentonGentry: Done

cmd/sniproxy/sniproxy.go
DentonGentry: Done

cmd/sniproxy/sniproxy.go
DentonGentry: Done

DentonGentry:
> I'm not entirely clear on why this is part of the SNI proxy, since as far as I can tell we aren't restricting the target of forwards to the SNI (? or are we), but mechanically this seems fine modulo some small things noted in my comments.

It is true that it is becoming increasingly inappropriately named. We should change its name to the name of the feature, which shall remain nameless here because it is a public PR.

DentonGentry:
I think all comments are addressed. No-one has approved yet, can't submit.

cmd/sniproxy/sniproxy.go
```diff
@@ -174,6 +264,49 @@ func (s *server) serveConn(c net.Conn) {
...
+}
+
+// forwardConn sets up a forwarder for a TCP connection. It does not inspect of the data
+// like the SNI forwarding does, it merely forwards all data to the destination specified
+// in the --forward=tcp/22/github.com argument.
+func (s *server) forwardConn(c net.Conn, forw *portForward) {
+	addrPortStr := c.LocalAddr().String()
+
+	var dialer net.Dialer
+	dialer.Timeout = 5 * time.Second
```
creachadair, line 293: Is this long enough? If the network is coming up it could be a bit before the netmap settles. I'm assuming this is a safety check rather than a hard timeout?

creachadair (approved):

cmd/sniproxy/sniproxy.go
DentonGentry: Intended just as a safety check, made it be 30 seconds instead.

PR #8980, opened 2023-08-21
control/controlclient: flesh out mapSession to break up gigantic method

bradfitz:
Now mapSession has a bunch more fields and methods, rather than being
just one massive func with a ton of local variables.

So far there are no major new optimizations, though. It should behave
the same as before.

This has been done with an eye towards testability (so tests can set
all the callback funcs as needed, or not, without a huge Direct client
or long-running HTTP requests), but this change doesn't add new tests
yet. That will follow in the changes which flesh out the NetmapUpdater
interface.

Updates #1909

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -54,7 +54,6 @@ import (
 	"tailscale.com/types/ptr"
 	"tailscale.com/types/tkatype"
 	"tailscale.com/util/clientmetric"
-	"tailscale.com/util/cmpx"
 	"tailscale.com/util/multierr"
 	"tailscale.com/util/singleflight"
 	"tailscale.com/util/systemd"
@@ -806,10 +805,10 @@ func (c *Direct) SendUpdate(ctx context.Context) error {
 	return c.sendMapRequest(ctx, false, nil)
 }
 
-// If we go more than pollTimeout without hearing from the server,
+// If we go more than watchdogTimeout without hearing from the server,
 // end the long poll. We should be receiving a keep alive ping
 // every minute.
-const pollTimeout = 120 * time.Second
+const watchdogTimeout = 120 * time.Second
 
 // sendMapRequest makes a /map request to download the network map, calling cb
 // with each new netmap. If isStreaming, it will poll forever and only returns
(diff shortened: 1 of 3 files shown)


control/controlclient/direct.go
```diff
@@ -1539,6 +1503,9 @@ func answerC2NPing(logf logger.Logf, c2nHandler http.Handler, c *http.Client, pr
 }
 
 func sleepAsRequested(ctx context.Context, logf logger.Logf, timeoutReset chan<- struct{}, d time.Duration, clock tstime.Clock) error {
+	if timeoutReset == nil {
+		timeoutReset = make(chan struct{})
+	}
```
maisem, line 1508: this is confusing, whats the point of making a channel here that no one can read from and always blocks anyway?

control/controlclient/map.go
```diff
@@ -49,23 +77,121 @@ type mapSession struct {
...
+	lastNetmapSummary      string // from NetworkMap.VeryConcise
 
 	// netMapBuilding is non-nil during a netmapForResponse call,
 	// containing the value to be returned, once fully populated.
 	netMapBuilding *netmap.NetworkMap
 }
 
-func newMapSession(privateNodeKey key.NodePrivate) *mapSession {
+// newMapSession returns a new mapSession.
+//
```
maisem, line 88: say that fields like cancel, onConciseNetMapSummary are stubs and should be replaced?

control/controlclient/direct.go
bradfitz: Oh, I was wrong about nil writable channels in select. Will remove. (They're fine: https://go.dev/play/p/po5IcntgwFU)

maisem: yeah, they just block forever

control/controlclient/direct.go
```diff
@@ -806,10 +805,10 @@ func (c *Direct) SendUpdate(ctx context.Context) error {
 	return c.sendMapRequest(ctx, false, nil)
 }
 
-// If we go more than pollTimeout without hearing from the server,
+// If we go more than watchdogTimeout without hearing from the server,
 // end the long poll. We should be receiving a keep alive ping
 // every minute.
-const pollTimeout = 120 * time.Second
+const watchdogTimeout = 120 * time.Second
```
maisem, line 811: oh this timeout lol

control/controlclient/direct.go
```diff
@@ -1074,82 +1078,42 @@ func (c *Direct) sendMapRequest(ctx context.Context, isStreaming bool, nu Netmap
...
-		if nm.SelfNode == nil {
-			c.logf("MapResponse lacked node")
-			return errors.New("MapResponse lacked node")
-		}
-
-		if DevKnob.StripEndpoints() {
-			for _, p := range resp.Peers {
-				p.Endpoints = nil
+		} else {
+			if resp.Node == nil {
```
maisem, line 1084: if the first message you get is a keep alive, followed up by a map response which doesn't have the SelfNode this check won't fire

bradfitz: I don't think that's legal. The very first MapResponse must be a full one.

control/controlclient/direct.go
```diff
@@ -1074,82 +1082,42 @@ func (c *Direct) sendMapRequest(ctx context.Context, isStreaming bool, nu Netmap
 		}
 
 		metricMapResponseMap.Add(1)
-		if i > 0 {
+		if mapResIdx > 0 {
 			metricMapResponseMapDelta.Add(1)
```
maisem, line 1086: this isn't necessarily delta until after `gotNonKeepAliveMessage = true`

maisem (approved):

PR #8982, opened 2023-08-21
cmd/tailscale: support write files to stdout for file get

### LINKED ISSUE #8981: Taildrop: support write file to stdout for file get

Currently Taildrop supports reading file from stdin (`tailscale file cp - ...`). Could you consider adding the counterpart for `file get`? That would allow for easier integration with other CLI tools.

### How should we solve this?

For example, `tailscale file get -` writes files to stdout. I will provide an example implementation.

_No response_

### Anything else?

_No response_

Contextualist:
This resolves #8981.

Caveat: Tailscale might write log messages to stdout. Normally `file get` does not print any message.


DentonGentry (changes requested):
It is not clear to me that this is the right thing to do. The discussion should happen on the issue, marking the PR as requesting changes to ensure it does not get submitted.

PR #8983, opened 2023-08-21
tool: force HTTP/1.1 in curl to prevent hang behind load balancer

valscale:
When running in our github CI environment, curl sometimes hangs while closing the download from the nodejs.org server and fails with INTERNAL_ERROR. This is likely caused by CI running behind some kind of load balancer or proxy that handles HTTP/2 incorrectly in some minor way.

This PR is a draft while I rerun the test enough times to see if this fixes it.

### DIFF
--- a/tool/node
+++ b/tool/node
@@ -49,7 +49,12 @@ fi
                 ARCH="arm64"
             fi
             mkdir -p "$cachedir"
-            curl -f -L -o "$tarball" "https://nodejs.org/dist/v${want_rev}/node-v${want_rev}-${OS}-${ARCH}.tar.gz"
+	    # When running on GitHub in CI, the below curl sometimes fails with
+	    # INTERNAL_ERROR after finishing the download. The most common cause
+	    # of INTERNAL_ERROR is glitches in intermediate hosts handling of
+	    # HTTP/2 forwarding, so forcing HTTP 1.1 often fixes the issue. See
+	    # https://github.com/tailscale/tailscale/issues/8988
+            curl -f -L --http1.1 -o "$tarball" "https://nodejs.org/dist/v${want_rev}/node-v${want_rev}-${OS}-${ARCH}.tar.gz"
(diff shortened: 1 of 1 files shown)


DentonGentry (approved):

soniaappasamy (approved):

tool/node
```diff
@@ -49,7 +49,8 @@ fi
                 ARCH="arm64"
             fi
             mkdir -p "$cachedir"
-            curl -f -L -o "$tarball" "https://nodejs.org/dist/v${want_rev}/node-v${want_rev}-${OS}-${ARCH}.tar.gz"
+	    # Running behind a load balancer can cause problems with HTTP/2
```
bradfitz, line 52: Can you elaborate? Or say that this is an experiment and speculative?

valscale: I googled the error message and "load balancers don't handle HTTP/2 quite right" was the most common answer for what causes INTERNAL_ERROR messages and switching to HTTP 1.1 was the most common solution. It seems to work and I wanted to spend an appropriate amount of time on this issue so I didn't try to understand in more detail. What do you think is an appropriate level of documentation here?

bradfitz: This comment as-is doesn't give any information about when we can remove this hack or any clues to figure that out.

Also, it's more confident sounding that the PR description itself which says that this might be the answer.

Ideally there would be a bug showing the exact error we got without this, and in which environment, and which curl version, and pointing the finger at a specific load balancer/host/vendor that we can then work with to get fixed (I worked with a handful of vendors when I wrote Go's HTTP/2 implementation). Then people reading this comment can look at the issue and see if it's been resolved and know they can then remove the flag.

bradfitz: (and then once you have a bug open, replace `#cleanup` in the commit message with the number)

valscale: Okay! Here you go: https://github.com/tailscale/tailscale/issues/8988

Let me know what edits you have. Thanks!

bradfitz (approved):

PR #8991, opened 2023-08-21
types/netmap, all: make NetworkMap.SelfNode a tailcfg.NodeView

bradfitz:
Updates #1909


envknob/logknob/logknob.go
```diff
@@ -68,7 +68,7 @@ func (lk *LogKnob) UpdateFromNetMap(nm NetMap) {
 		return
 	}
 
-	lk.cap.Store(slices.Contains(nm.SelfCapabilities(), lk.capName))
+	lk.cap.Store(views.SliceContains(nm.SelfCapabilities(), lk.capName))
```
maisem, line 71: why didn't we add `Contains` on `views.Slice` 🤦🏼

maisem: oh... it needs `T comparable` 😞

maisem (approved):

ssh/tailssh/tailssh.go
```diff
@@ -787,7 +787,7 @@ func (c *conn) expandDelegateURLLocked(actionURL string) string {
 	lu := c.localUser
 	var dstNodeID string
 	if nm != nil {
-		dstNodeID = fmt.Sprint(int64(nm.SelfNode.ID))
+		dstNodeID = fmt.Sprint(int64(nm.SelfNode.ID()))
```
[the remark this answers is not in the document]
bradfitz: Don't think so? Its String method is used otherwise which this was wanting to avoid.

maisem (approved):

Issue #8992, opened 2023-08-21
Tracking issues for UPnP improvements


andrew-d:
This is a tracking issue for improvements to UPnP portmapping (i.e. not NAT-PMP, PCP, etc.)

PR #8998, opened 2023-08-21
cmd/tailscale, net/portmapper: add --log-http option to "debug portmap"

andrew-d:
This option allows logging the raw HTTP requests and responses that the portmapper Client makes when using UPnP. This can be extremely helpful when debugging strange UPnP issues with users' devices, and might allow us to avoid having to instruct users to perform a packet capture.

Updates #8992

Change-Id: I2c3cf6930b09717028deaff31738484cc9b008e4

### DIFF
--- a/client/tailscale/localclient.go
+++ b/client/tailscale/localclient.go
@@ -37,6 +37,7 @@ import (
 	"tailscale.com/tka"
 	"tailscale.com/types/key"
 	"tailscale.com/types/tkatype"
+	"tailscale.com/util/cmpx"
 )
 
 // defaultLocalClient is the default LocalClient when using the legacy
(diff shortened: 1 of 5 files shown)


client/tailscale/localclient.go
```diff
@@ -391,16 +391,26 @@ func (lc *LocalClient) DebugAction(ctx context.Context, action string) error {
...
+	// LogHTTP instructs the debug-portmap endpoint to print all HTTP
+	// requests and responses made to the logs.
+	LogHTTP bool
+}
+
 // DebugPortmap invokes the debug-portmap endpoint, and returns an
 // io.ReadCloser that can be used to read the logs that are printed during this
 // process.
-func (lc *LocalClient) DebugPortmap(ctx context.Context, duration time.Duration, ty, gwSelf string) (io.ReadCloser, error) {
+func (lc *LocalClient) DebugPortmap(ctx context.Context, duration time.Duration, ty, gwSelf string, opts *DebugPortmapOpts) (io.ReadCloser, error) {
```
bradfitz, line 404: normally we don't change the signature of things in this package because it's kinda a stable API, but this is named "Debug" so I'll let it slide :)

client/tailscale/localclient.go
```diff
@@ -391,16 +391,26 @@ func (lc *LocalClient) DebugAction(ctx context.Context, action string) error {
...
 }
 
+// DebugPortmapOpts contains options for the DebugPortmap command.
+type DebugPortmapOpts struct {
+	// LogHTTP instructs the debug-portmap endpoint to print all HTTP
+	// requests and responses made to the logs.
+	LogHTTP bool
+}
+
 // DebugPortmap invokes the debug-portmap endpoint, and returns an
```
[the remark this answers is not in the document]
bradfitz: And the IP should be a netip.Addr probably and required.

client/tailscale/localclient.go
andrew-d: Yeah, I figured that things named `Debug` are not in our stability guarantee 😅

client/tailscale/localclient.go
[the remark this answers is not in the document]
andrew-d: > what is duration used for? I guess how long to make the mapping?

Yep!

> Maybe we should just change this signature entirely to only take big opts, or make everything plausibly optional into opts, like duration (default to 1 minute or something).

I could do that, yeah; I like it better than "a bucket of strings".

> What is "ty"? Can it have a default?

From reading the code... it's whether we want to perform a specific kind of portmap test; empty means "all", otherwise you can specify "pmp", "pcp" or "upnp". Will document.

> And the IP should be a netip.Addr probably and required.

`gwSelf` isn't required because it will be determined via `netMon.GatewayAndSelfIP` if not provided; will document.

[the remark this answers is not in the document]
andrew-d: @bradfitz - rewrote this method to take a (documented) struct type. I intentionally didn't change the format of the requests to `tailscaled` so as not to completely break backwards compatibility, but at least all the fields in the struct are documented and properly typed (i.e. not `gatewayAndSelf string // format "gatewayAddr/selfAddr"`)

client/tailscale/localclient.go
```diff
@@ -391,15 +391,53 @@ func (lc *LocalClient) DebugAction(ctx context.Context, action string) error {
...
+	SelfAddr netip.Addr
+
+	// LogHTTP instructs the debug-portmap endpoint to print all HTTP
+	// requests and responses made to the logs.
+	LogHTTP bool
+}
+
 // DebugPortmap invokes the debug-portmap endpoint, and returns an
 // io.ReadCloser that can be used to read the logs that are printed during this
 // process.
```
bradfitz, line 422: document that nil is a valid opts value

client/tailscale/localclient.go
```diff
@@ -391,15 +391,53 @@ func (lc *LocalClient) DebugAction(ctx context.Context, action string) error {
...
-		vals.Set("gateway_and_self", gwSelf)
+	if opts == nil {
+		opts = &DebugPortmapOpts{}
+	}
+
+	if opts.Duration != 0 {
+		vals.Set("duration", opts.Duration.String())
+	} else {
+		vals.Set("duration", "5s")
+	}
```
bradfitz, line 433: if you want:

```go
    vals.Set("duration", cmpx.Or(opts.Duration, 5 * time.Second).String())
```

(`cmp.Or` is coming to Go; we brought it in early as `cmpx.Or` for now)

cmd/tailscale/cli/debug.go
```diff
@@ -818,17 +821,34 @@ func runCapture(ctx context.Context, args []string) error {
...
+		opts.GatewayAddr, err = netip.ParseAddr(debugPortmapArgs.gatewayAddr)
+		if err != nil {
+			return fmt.Errorf("invalid --gateway-addr: %w", err)
+		}
+		opts.SelfAddr, err = netip.ParseAddr(debugPortmapArgs.selfAddr)
+		if err != nil {
+			return fmt.Errorf("invalid --self-addr: %w", err)
+		}
+	}
+	rc, err := localClient.DebugPortmap(ctx, &opts)
```
bradfitz, line 851: nit: kinda prefer this & moves to line 832 so this doesn't look like an out parameter.

bradfitz (approved):

Issue #8999, opened 2023-08-21
addrForWireGuardSendLocked will stick with an endpoint for 1hr even if a network change made that endpoint not vialbe


raggi:
addrForWireGuardSendLocked relies on the length of the pong state to determine if it should persist an endpoint for 1hr.

If a device has successfully had ICMP ping/pong for a while, but its endpoints then change, those pongs are no longer applicable, and could lead to incorrect endpoint fixation for an extended period of time.

PR #9004, opened 2023-08-21
wgengine/magicsock: clear out endpoint statistics when it becomes bad

raggi:
There are cases where we do not detect the non-viability of a route, but we will instead observe a failure to send. In a Disco path this would normally be handled as a side effect of Disco, which is not available to non-Disco WireGuard nodes. In both cases, recognizing the failure as such will result in faster convergence.

Updates #8999

### DIFF
--- a/wgengine/magicsock/endpoint.go
+++ b/wgengine/magicsock/endpoint.go
@@ -469,6 +469,14 @@ func (de *endpoint) send(buffs [][]byte) error {
 	var err error
 	if udpAddr.IsValid() {
 		_, err = de.c.sendUDPBatch(udpAddr, buffs)
+
+		// If the error is known to indicate that the endpoint is no longer
+		// usable, clear the endpoint statistics so that the next send will
+		// re-evaluate the best endpoint.
+		if err != nil && isBadEndpointErr(err) {
+			de.noteBadEndpoint(udpAddr)
+		}
+
 		// TODO(raggi): needs updating for accuracy, as in error conditions we may have partial sends.
 		if stats := de.c.stats.Load(); err == nil && stats != nil {
 			var txBytes int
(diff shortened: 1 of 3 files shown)


wgengine/magicsock/endpoint.go
```diff
@@ -475,6 +476,14 @@ func (de *endpoint) send(buffs [][]byte) error {
 	var err error
 	if udpAddr.IsValid() {
 		_, err = de.c.sendUDPBatch(udpAddr, buffs)
+
+		// If the OS reports the address as unreachable, clear out any potential
+		// reasons that we decided to prefer this address, as they're no longer
+		// valid.
+		if errors.Is(err, unix.EHOSTUNREACH) {
```
bradfitz, line 483: this allocs per call because `unix.EHOSTUNREACH` is a uintptr (not a pointer) so it can't go in an interface value.

You'll want to do `var errHOSTUNREACH error = unix.EHOSTUNREACH` as a global to do it once and then use that err. And maybe `if err != nil && errors.Is` too IIRC, if that still matters.

wgengine/magicsock/endpoint.go
```diff
@@ -866,6 +875,22 @@ func (de *endpoint) addCandidateEndpoint(ep netip.AddrPort, forRxPingTxID stun.T
...
+func (de *endpoint) noteBadEndpoint(ipp netip.AddrPort) {
+	de.mu.Lock()
+	defer de.mu.Unlock()
+
+	if de.bestAddr.AddrPort == ipp {
+		de.bestAddr = addrLatency{}
+	}
+
+	if _, ok := de.endpointState[ipp]; ok {
+		de.endpointState[ipp].clear()
```
bradfitz, line 890: why do the map lookup twice? It was right there in _

wgengine/magicsock/endpoint.go
```diff
@@ -858,14 +879,35 @@ func (de *endpoint) addCandidateEndpoint(ep netip.AddrPort, forRxPingTxID stun.T
...
 
+// noteBadEndpoint marks ipp as a bad endpoint that would need to be
+// re-evaluated before future use, this should be called for example if a send
+// to ipp fails due to a host unreachable error or similar.
+func (de *endpoint) noteBadEndpoint(ipp netip.AddrPort) {
+	de.mu.Lock()
+	defer de.mu.Unlock()
+
+	if de.bestAddr.AddrPort == ipp {
+		de.bestAddr = addrLatency{}
```
catzkorn, line 890: I think we should do `de.trustBestAddrUntil = 0` here too

wgengine/magicsock/endpoint.go
```diff
@@ -866,15 +879,36 @@ func (de *endpoint) addCandidateEndpoint(ep netip.AddrPort, forRxPingTxID stun.T
 	return false
 }
 
+// clearBestAddrLocked clears the bestAddr and related fields such that future
+// packets will re-evaluate the best address to send to next.
```
catzkorn, line 883:
```suggestion
// clearBestAddrLocked clears the bestAddr and related fields such that future
// packets will re-evaluate the best address to send to next.
//
// b.mu must be held.
```
I believe we follow the convention that we add this type of comment for functions like this.

PR #9006, opened 2023-08-22
wgengine/magicsock: improve endpoint selection for WireGuard peers with rx time

raggi:
If we don't have the ICMP hint available, such as on Android, we can use the signal of rx traffic to bias toward a particular endpoint.

We don't want to stick to a particular endpoint for a very long time without any signals, so the sticky time is reduced to 1 second, which is large enough to avoid excessive packet reordering in the common case, but should be small enough that either rx provides a strong signal, or we rotate in a user-interactive schedule to another endpoint, improving the feel of failover to other endpoints.

Updates #8999


catzkorn (commented):
Reviewed in VC + viewed here

PR #9009, opened 2023-08-22
.github: add flakehub-publish-tagged.yml

### LINKED ISSUE #9008: Publish releases to FlakeHub
FlakeHub is a Nix flake management platform built by [Determinate Systems](https://determinate.systems/).

"Find, install, and publish Nix Flakes with FlakeHub.
Predictable and safe flake updates for users, and the freedom for developers to iterate on their API."

https://flakehub.com/

shayne:
This workflow will publish a flake to flakehub when a tag is pushed to the repository. It will only publish tags that match the pattern `v*.*.*`.

Fixes #9008

### DIFF
--- a/.github/workflows/flakehub-publish-tagged.yml
+++ b/.github/workflows/flakehub-publish-tagged.yml
@@ -1,0 +1,17 @@
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - '^v[0-9]+\.[0-9]*[02468]+\.[0-9]+$'
+jobs:
+  publish:
+    runs-on: "ubuntu-latest"
+    permissions:
+      id-token: "write"
+      contents: "read"
+    steps:
+      - uses: "actions/checkout@v3"
+      - uses: "DeterminateSystems/nix-installer-action@main"
+      - uses: "DeterminateSystems/flakehub-push@main"
+        with:
+          visibility: "public"


shayne:
Added ` workflow_dispatch:` to publish the latest tag on demand. Will PR its removal after the first publish.

.github/workflows/flakehub-publish-tagged.yml
```diff
@@ -0,0 +1,17 @@
+on:
+  workflow_dispatch:
+  push:
```
DentonGentry, line 3: Can we make this only run for branches named `release-branch/*`?

At some point I expect to start tagging `main` with unstable builds 1.49.X, because tracking what is in each unstable has been a pain point. We wouldn't want to accidentally push those to Nix.

DentonGentry (approved):

.github/workflows/flakehub-publish-tagged.yml
shayne: Unfortunately, the semantics require a tag. What if we matched against even minor versions?

```
'^v\d+\.\d*[02468]+\.\d+$'
```

DentonGentry: Triggering only on even numbered branches would solve the issue, if that is the only way to make it work.

This doesn't work?
```
on:
  push:
    branches:
      - "release-branch/*"
    tags:
      - "v*.*.*"
```

shayne: I believe this will run either branches OR tags.

shayne: Changed it to `^v[0-9]+\.[0-9]*[02468]+\.[0-9]+$` because `\d` might be too cute for GitHub

DentonGentry (approved):

PR #9012, opened 2023-08-22
client/web: hook up remaining legacy POST requests

soniaappasamy:
Hooks up remaining legacy POST request from the React side in --dev.

Updates tailscale/corp#13775

### DIFF
--- a/client/web/src/components/legacy.tsx
+++ b/client/web/src/components/legacy.tsx
@@ -7,9 +7,13 @@ import { NodeData, NodeUpdate } from "src/hooks/node-data"
 // purely to ease migration to the new React-based web client, and will
 // eventually be completely removed.
 
-export function Header(props: { data: NodeData }) {
-  const { data } = props
-
+export function Header({
+  data,
+  updateNode,
+}: {
+  data: NodeData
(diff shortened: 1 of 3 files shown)


client/web/src/components/legacy.tsx
```diff
@@ -61,41 +65,52 @@ export function Header(props: { data: NodeData }) {
...
-                <a href="#" className="hover:text-gray-700 js-loginButton">
-                  Reauthenticate
-                </a>{" "}
-                |{" "}
-                <a href="#" className="hover:text-gray-700 js-logoutButton">
-                  Logout
-                </a>
+        {data.Profile &&
+          data.Status !== "NoState" &&
+          data.Status !== "NeedsLogin" && (
```
soniaappasamy, line 70: The top bar was showing up even when they couldn't use it. Updated to hide when in an invalid state. The actions for those two cases are handled in the `State` component below. Ideally this top bar should probably be rendered from the `State` component that looks at these values already, but w/o changing around a bunch of the legacy code, I think just hiding here is okay.

willnorris (approved):

PR #2093, opened 2021-06-10
cmd/tailscale/web: add support for QNAP

maisem:
Updates #1346

PR #1621, opened 2021-03-30
cmd/tailscale: add initial web UI

rosszurowski:
This is a draft of a some UI improvements for the web frontend introduced in #1605. It's pretty basic at the moment, showing device and user information, and any urgent actions that need to be taken.

### Todo

- [ ] Investigate randomly getting kicked back to "NeedsLogin"
- [ ] Reduce unused CSS rules
- [ ] Include a "log out" action?

PR #9024, opened 2023-08-22
client/web: remove old /redirect handler

willnorris:
I thought this had something to do with Synology or QNAP support, since they both have specific authentication logic.  But it turns out this was part of the original web client added in #1621, and then refactored as part of #2093.  But with how we handle logging in now, it's never called.

Updates tailscale/corp#13775


soniaappasamy (approved):
Nice!

PR #9026, opened 2023-08-23
tsweb: clean up pprof handler registration, document why it's there

danderson:
Updates #cleanup

### DIFF
--- a/tsweb/debug.go
+++ b/tsweb/debug.go
@@ -48,16 +48,19 @@ func Debugger(mux *http.ServeMux) *DebugHandler {
 	}
 	mux.Handle("/debug/", ret)
 
-	// Register this one directly on mux, rather than using
-	// ret.URL/etc, as we don't need another line of output on the
-	// index page. The /pprof/ index already covers it.
-	mux.Handle("/debug/pprof/profile", BrowserHeaderHandler(http.HandlerFunc(pprof.Profile)))
-
 	ret.KVFunc("Uptime", func() any { return varz.Uptime() })
 	ret.KV("Version", version.Long())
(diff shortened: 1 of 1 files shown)


tsweb/debug.go
```diff
@@ -48,16 +48,11 @@ func Debugger(mux *http.ServeMux) *DebugHandler {
 	}
 	mux.Handle("/debug/", ret)
 
-	// Register this one directly on mux, rather than using
-	// ret.URL/etc, as we don't need another line of output on the
-	// index page. The /pprof/ index already covers it.
-	mux.Handle("/debug/pprof/profile", BrowserHeaderHandler(http.HandlerFunc(pprof.Profile)))
```
bradfitz, line 54: was that BrowserHeaderHandler important?

bradfitz (approved):

tsweb/debug.go
danderson: Yes, but the debughandler's ServerHTTP does it inline, so no need for a dedicated extra wrapping.

danderson:
Strange, tests are failing now. From the pprof package's documentation, pprof.Index _should_ serve the individual profiles as well as the index of available profiles, but doesn't seem to. I'll have to look more tomorrow.

danderson:
Found it. /debug/pprof/profile is special and not like all the other profile types, in that pprof.Index doesn't know about it. So, we _do_ have to register that handler separately from pprof.Index. I've changed the PR to make that registration happen inline with other pprof things, and document why pprof.Index is not sufficient.

creachadair (approved):

Issue #8184, opened 2023-05-22
FR: Kubernetes operator to provide egress proxy


parasol-me:
### What are you trying to do?

Our stack consists of two Kubernetes clusters, each with an API service.
The backing of the API service is a deployment to scale the pod replicas.
We want to provide two-way connection between these services over our tailnet in order to avoid public ingress and authentication.

We have been using the tailscale operator to provide inbound connectivity to the services, but in order to provide outbound connectivity to the API pods we have the following options:

- Manually create a tailscale proxy service in each cluster that points to the IP address of the service exposed by operator from the other cluster
- Manually add tailscale sidecar to the api pods

We are deploying many of these two cluster setups, so managing the hostnames for these services is important.

The first option is a bit limited because it requires providing the destination IP address and port.

For the second option, the sidecar already provides two-way communication, but we don't actually want each pod in the deployment to be exposed as a host. It should only provide outbound connection to the service exposed by operator.

### How should we solve this?

It would be good if the operator could provide both outbound and inbound connections via either an additional proxy (ideally only hostname required) or by attaching sidecars to pods but with only an outbound connection (not exposing each pod as a tailnet host).

Maybe there is a better way to achieve it, we are also reaching out for suggestions.

Basically, we would like to use only the operator to provide bi-directional communication between services where the backing is a scalable deployment.

### What is the impact of not solving this?

Currently, we are either manually creating outbound proxies to specific IP addresses or manually attaching a sidecar to all pods but only for outbound connections.

In combination with using the operator to tag services for inbound connection.

### Anything else?

N/A

maisem:
We have been thinking about this problem a bit, would you need Tailscale provisioned TLS certs in this setup? I ask because there are a few ways to expose the service. Either we expose the full MagicDNS name `api.jay-snake.ts.net` or something that is only cluster resolvable like `api.tailscale.svc`.

mayakacz:
This is now supported in the K8s operator: https://tailscale.com/kb/1236/kubernetes-operator/#cluster-egress

> We want to provide two-way connection between these services over our tailnet in order to avoid public ingress and authentication.

This is now documented as well: https://tailscale.com/kb/1236/kubernetes-operator/#exposing-a-service-in-one-cluster-to-another-cluster-cross-cluster-connectivity

parasol-me:
@mayakacz Thank you for letting me know, we will definitely give this a try. It should simplify our deployments a lot!

PR #9031, opened 2023-08-23
cmd/k8s-operator,cmd/containerboot: add kube egress proxy

irbekrm:
First part of work for the functionality that allows users to create an egress proxy to access Tailnet services from within Kubernetes cluster workloads. This PR allows creating an egress proxy that can access Tailscale services over HTTP only.

This PR allows users to specify a Tailscale service that should be made accessible to workloads in Kubernetes cluster by adding a `tailscale.com/target-ip` annotation to a Kubernetes `Service` with the value being the Tailscale IP of the service they wish to access.
Tailscale operator then does work in the background to create an egress proxy.

Tailscale operator updates `spec.externalName` of the annotated Service with a DNS name that workloads in cluster can use to access the Tailscale service. This currently gets set to the Kuberentes DNS name of the proxy service, but this will be changed to Tailscale service MagicDNS name when we implement support for accessing services over TLS.

Updates https://github.com/tailscale/tailscale/issues/8184

### DIFF
--- a/cmd/containerboot/main.go
+++ b/cmd/containerboot/main.go
@@ -16,6 +16,8 @@
 //   - TS_ROUTES: subnet routes to advertise.
 //   - TS_DEST_IP: proxy all incoming Tailscale traffic to the given
 //     destination.
+//   - TS_TAILNET_TARGET_IP: proxy all incoming non-Tailscale traffic to the given
+//     destination.
 //   - TS_TAILSCALED_EXTRA_ARGS: extra arguments to 'tailscaled'.
 //   - TS_EXTRA_ARGS: extra arguments to 'tailscale login', these are not
 //     reset on restart.
@@ -88,8 +90,9 @@ func main() {
 		AuthKey:         defaultEnvs([]string{"TS_AUTHKEY", "TS_AUTH_KEY"}, ""),
 		Hostname:        defaultEnv("TS_HOSTNAME", ""),
 		Routes:          defaultEnv("TS_ROUTES", ""),
-		ProxyTo:         defaultEnv("TS_DEST_IP", ""),
 		ServeConfigPath: defaultEnv("TS_SERVE_CONFIG", ""),
+		ProxyTo:         defaultEnv("TS_DEST_IP", ""),
+		TailnetTargetIP: defaultEnv("TS_TAILNET_TARGET_IP", ""),
 		DaemonExtraArgs: defaultEnv("TS_TAILSCALED_EXTRA_ARGS", ""),
 		ExtraArgs:       defaultEnv("TS_EXTRA_ARGS", ""),
 		InKubernetes:    os.Getenv("KUBERNETES_SERVICE_HOST") != "",
@@ -107,16 +110,17 @@ func main() {
 	if cfg.ProxyTo != "" && cfg.UserspaceMode {
 		log.Fatal("TS_DEST_IP is not supported with TS_USERSPACE")
 	}
-	if cfg.ProxyTo != "" && cfg.ServeConfigPath != "" {
-		log.Fatal("TS_DEST_IP is not supported with TS_SERVE_CONFIG")
+
+	if cfg.TailnetTargetIP != "" && cfg.UserspaceMode {
+		log.Fatal("TS_TAILNET_TARGET_IP is not supported with TS_USERSPACE")
 	}
 
 	if !cfg.UserspaceMode {
 		if err := ensureTunFile(cfg.Root); err != nil {
 			log.Fatalf("Unable to create tuntap device file: %v", err)
 		}
-		if cfg.ProxyTo != "" || cfg.Routes != "" {
-			if err := ensureIPForwarding(cfg.Root, cfg.ProxyTo, cfg.Routes); err != nil {
+		if cfg.ProxyTo != "" || cfg.Routes != "" || cfg.TailnetTargetIP != "" {
+			if err := ensureIPForwarding(cfg.Root, cfg.ProxyTo, cfg.TailnetTargetIP, cfg.Routes); err != nil {
 				log.Printf("Failed to enable IP forwarding: %v", err)
 				log.Printf("To run tailscale as a proxy or router container, IP forwarding must be enabled.")
 				if cfg.InKubernetes {
@@ -270,7 +274,7 @@ authLoop:
 	}
 
 	var (
-		wantProxy         = cfg.ProxyTo != ""
+		wantProxy         = cfg.ProxyTo != "" || cfg.TailnetTargetIP != ""
 		wantDeviceInfo    = cfg.InKubernetes && cfg.KubeSecret != "" && cfg.KubernetesCanPatch
 		startupTasksDone  = false
 		currentIPs        deephash.Sum // tailscale IPs assigned to device
@@ -298,10 +302,12 @@ authLoop:
 		}
 		if n.NetMap != nil {
 			addrs := n.NetMap.SelfNode.Addresses().AsSlice()
-			if cfg.ProxyTo != "" && len(addrs) > 0 && deephash.Update(&currentIPs, &addrs) {
+			newCurrentIPs := deephash.Hash(&addrs)
+			ipsHaveChanged := newCurrentIPs != currentIPs
+			if cfg.ProxyTo != "" && len(addrs) > 0 && ipsHaveChanged {
 				log.Printf("Installing proxy rules")
-				if err := installIPTablesRule(ctx, cfg.ProxyTo, addrs); err != nil {
-					log.Fatalf("installing proxy rules: %v", err)
+				if err := installIngressForwardingRule(ctx, cfg.ProxyTo, addrs); err != nil {
+					log.Fatalf("installing ingress proxy rules: %v", err)
 				}
 			}
 			if cfg.ServeConfigPath != "" && len(n.NetMap.DNS.CertDomains) > 0 {
@@ -314,6 +320,13 @@ authLoop:
 					}
 				}
 			}
+			if cfg.TailnetTargetIP != "" && ipsHaveChanged && len(addrs) > 0 {
+				if err := installEgressForwardingRule(ctx, cfg.TailnetTargetIP, addrs); err != nil {
+					log.Fatalf("installing egress proxy rules: %v", err)
+				}
+			}
+			currentIPs = newCurrentIPs
+
 			deviceInfo := []any{n.NetMap.SelfNode.StableID(), n.NetMap.SelfNode.Name()}
 			if cfg.InKubernetes && cfg.KubernetesCanPatch && cfg.KubeSecret != "" && deephash.Update(&currentDeviceInfo, &deviceInfo) {
 				if err := storeDeviceInfo(ctx, cfg.KubeSecret, n.NetMap.SelfNode.StableID(), n.NetMap.SelfNode.Name(), n.NetMap.SelfNode.Addresses().AsSlice()); err != nil {
@@ -572,14 +585,25 @@ func ensureTunFile(root string) error {
 }
 
 // ensureIPForwarding enables IPv4/IPv6 forwarding for the container.
-func ensureIPForwarding(root, proxyTo, routes string) error {
+func ensureIPForwarding(root, clusterProxyTarget, tailnetTargetiP, routes string) error {
 	var (
 		v4Forwarding, v6Forwarding bool
 	)
-	if proxyTo != "" {
-		proxyIP, err := netip.ParseAddr(proxyTo)
+	if clusterProxyTarget != "" {
+		proxyIP, err := netip.ParseAddr(clusterProxyTarget)
 		if err != nil {
-			return fmt.Errorf("invalid proxy destination IP: %v", err)
+			return fmt.Errorf("invalid cluster destination IP: %v", err)
+		}
+		if proxyIP.Is4() {
+			v4Forwarding = true
+		} else {
+			v6Forwarding = true
+		}
+	}
+	if tailnetTargetiP != "" {
+		proxyIP, err := netip.ParseAddr(tailnetTargetiP)
+		if err != nil {
+			return fmt.Errorf("invalid tailnet destination IP: %v", err)
 		}
 		if proxyIP.Is4() {
 			v4Forwarding = true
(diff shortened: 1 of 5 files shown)


irbekrm:
@maisem this has been rebased on the latest main, PTAL

cmd/containerboot/main.go
```diff
@@ -102,19 +105,16 @@ func main() {
 		Root:            defaultEnv("TS_TEST_ONLY_ROOT", "/"),
 	}
 
-	if cfg.ProxyTo != "" && cfg.UserspaceMode {
+	if cfg.IngressProxyTo != "" && cfg.UserspaceMode {
```
maisem, line 108: You need to add a similar check for egress

cmd/containerboot/main.go
```diff
@@ -295,9 +297,11 @@ authLoop:
 			log.Fatalf("tailscaled left running state (now in state %q), exiting", *n.State)
 		}
 		if n.NetMap != nil {
-			if cfg.ProxyTo != "" && len(n.NetMap.Addresses) > 0 && deephash.Update(&currentIPs, &n.NetMap.Addresses) {
-				if err := installIPTablesRule(ctx, cfg.ProxyTo, n.NetMap.Addresses); err != nil {
-					log.Fatalf("installing proxy rules: %v", err)
+			newCurrentIPs := deephash.Hash(&n.NetMap.Addresses)
+			ipsHaveChanged := newCurrentIPs != currentIPs
```
maisem, line 301: You never set currentIPs

cmd/containerboot/main.go
```diff
@@ -310,6 +314,15 @@ authLoop:
 					}
 				}
 			}
+			// TODO (irbekrm): check if destination IP changes (i.e user made a typo & later corrected)
```
maisem, line 317: That will require a container restart to register atm

cmd/k8s-operator/sts.go
```diff
@@ -346,8 +357,8 @@ func (a *tailscaleSTSReconciler) reconcileSTS(ctx context.Context, logger *zap.S
 	ss.Spec.Template.Annotations = map[string]string{
 		"tailscale.com/operator-last-set-hostname": sts.Hostname,
 	}
-	if sts.TargetIP != "" {
-		ss.Spec.Template.Annotations["tailscale.com/operator-last-set-ip"] = sts.TargetIP
+	if sts.ClusterTargetIP != "" {
+		ss.Spec.Template.Annotations["tailscale.com/operator-last-set-ip"] = sts.ClusterTargetIP
```
maisem, line 361: You want another similar annotation for egress

cmd/k8s-operator/svc.go
```diff
@@ -181,7 +196,10 @@ func (a *ServiceReconciler) hasLoadBalancerClass(svc *corev1.Service) bool {
 		*svc.Spec.LoadBalancerClass == "tailscale"
 }
 
-func (a *ServiceReconciler) hasAnnotation(svc *corev1.Service) bool {
-	return svc != nil &&
-		svc.Annotations[AnnotationExpose] == "true"
+func (a *ServiceReconciler) hasExposeAnnotation(svc *corev1.Service) bool {
```
maisem, line 199: Needs docs now

cmd/containerboot/main.go
```diff
@@ -102,19 +105,16 @@ func main() {
 		Root:            defaultEnv("TS_TEST_ONLY_ROOT", "/"),
 	}
 
-	if cfg.ProxyTo != "" && cfg.UserspaceMode {
+	if cfg.IngressProxyTo != "" && cfg.UserspaceMode {
 		log.Fatal("TS_DEST_IP is not supported with TS_USERSPACE")
 	}
-	if cfg.ProxyTo != "" && cfg.ServeConfigPath != "" {
```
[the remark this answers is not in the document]
maisem: Also check for egress

cmd/containerboot/main.go
irbekrm: We actually already do it in an init container as well https://github.com/tailscale/tailscale/blob/main/cmd/k8s-operator/manifests/proxy.yaml#L13-L21. I'm guessing we can then remove the init container - the check in the code is smarter as it allows for cases where it is not allowed to run sysctl commands

maisem: the init container isn't for checking, its to ensure we have ip forwarding enabled.

cmd/containerboot/main.go
irbekrm: They get set here https://github.com/tailscale/tailscale/pull/9031/commits/4fe28d58331cd365ba884e48714d9d6d801820a9#diff-700830ac79a41da42cd3a82065f0fee3003a4390eead4db1de4bf30557a06bd6R324
It's the same logic where they are set after the loop has ran once, but it's moved lower so that we use the same currentIPs for egress proxy

cmd/containerboot/main.go
irbekrm: I'll think about that separately

cmd/k8s-operator/sts.go
irbekrm: I added this field to the `tailscaleSTSConfig` that will be populated with the value of the `tailscale.com/target-ip` annotation value

cmd/containerboot/main.go
irbekrm: the code also enables, it writes `1` here if not already https://github.com/tailscale/tailscale/blob/main/cmd/containerboot/main.go#L607

irbekrm: (I did add the check to code though)

cmd/k8s-operator/svc.go
maisem: oh i meant the function needs a docstring

cmd/k8s-operator/operator.go
```diff
@@ -264,5 +264,5 @@ func startReconcilers(zlog *zap.SugaredLogger, tsNamespace string, restConfig *r
 
 type tsClient interface {
 	CreateKey(ctx context.Context, caps tailscale.KeyCapabilities) (string, *tailscale.Key, error)
-	DeleteDevice(ctx context.Context, nodeStableID string) error
```
maisem, line 267: revert

cmd/k8s-operator/sts.go
maisem: >"tailscale.com/operator-last-set-ip"

is an annotation added by the controller so that the stateful set notices it and rolls out a new version of the proxy in case the value changes. You will need another annotation that matches the desired value of `TailscaleTargetIP`

cmd/k8s-operator/svc.go
```diff
@@ -181,7 +196,14 @@ func (a *ServiceReconciler) hasLoadBalancerClass(svc *corev1.Service) bool {
 		*svc.Spec.LoadBalancerClass == "tailscale"
 }
 
-func (a *ServiceReconciler) hasAnnotation(svc *corev1.Service) bool {
-	return svc != nil &&
-		svc.Annotations[AnnotationExpose] == "true"
+// hasExposeAnnotation returns true if Service has tailscale.com/expose
```
maisem, line 199: go style is to say `reports whether` for bool returns

cmd/k8s-operator/sts.go
irbekrm: Ahh I misunderstood you- I thought you meant I was overriding an existing annotation. Added a `tailscale.com/operator-last-set-tailnet-ip` that will force the pod to restart if the target changes.

cmd/containerboot/main.go
irbekrm: This is now solved by the last-target-ip annotation that forces pod restart, so I've removed the comment

knyar (commented):
I am a bit surprised that we call this a "proxy" even though it does NAT. I don't feel strongly about this, but I thought I would mention it in case other people also think it's a bit weird.

cmd/containerboot/main.go
```diff
@@ -612,7 +639,53 @@ func ensureIPForwarding(root, proxyTo, routes string) error {
...
+		}
+		local = pfx.Addr().String()
+		break
+	}
+	if local == "" {
+		return fmt.Errorf("no tailscale IP matching family of %s found in %v", dstStr, tsIPs)
+	}
+	// Technically, if the control server ever changes the IPs assigned to this
+	// node, we'll slowly accumulate iptables rules. This shouldn't happen, so
+	// for now we'll live with it.
```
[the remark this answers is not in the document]
irbekrm: they will get installed more than once during proxy pod's lifetime if the Tailscale IP of the pod changes (which is of course unlikely)

cmd/containerboot/main.go
```diff
@@ -310,6 +318,14 @@ authLoop:
 					}
 				}
 			}
+			if cfg.EgressProxyTo != "" && !installedEgressProxy && len(n.NetMap.Addresses) > 0 && ipsHaveChanged {
```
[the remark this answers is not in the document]
irbekrm: Thanks for spotting that, should have been fixed now https://github.com/tailscale/tailscale/pull/9031/commits/24207a94cfb6f59eee54858fa4327bd98f47d5db#diff-700830ac79a41da42cd3a82065f0fee3003a4390eead4db1de4bf30557a06bd6R323

cmd/containerboot/main.go
```diff
@@ -16,6 +16,8 @@
 //   - TS_ROUTES: subnet routes to advertise.
 //   - TS_DEST_IP: proxy all incoming Tailscale traffic to the given
 //     destination.
+//   - TS_EGRESS_IP: proxy all incoming non-Tailscale traffic to the given
```
maisem, line 19: This has five different names. 

TS_EGRESS_IP
EgressProxyTo
TailnetProxyTarget
TailscaleProxyTarget
TargetIP


Let's pick one and use that everywhere

cmd/containerboot/main.go
```diff
@@ -312,6 +320,14 @@ authLoop:
 					}
 				}
 			}
+			if cfg.EgressProxyTo != "" && !installedEgressProxy && len(n.NetMap.Addresses) > 0 && ipsHaveChanged {
+				if err := installEgressForwardingRule(ctx, cfg.EgressProxyTo, n.NetMap.Addresses); err != nil {
+					log.Fatalf("installing egress proxy rules: %v", err)
+				}
+				installedEgressProxy = true
```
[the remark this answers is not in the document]
irbekrm: Not anymore https://github.com/tailscale/tailscale/commit/24207a94cfb6f59eee54858fa4327bd98f47d5db#diff-700830ac79a41da42cd3a82065f0fee3003a4390eead4db1de4bf30557a06bd6R323 they will also get installed if the proxy Tailnet IP has changed thanks to Anton also noticing this

cmd/containerboot/main.go
irbekrm: I've changed all to some version of 'TS_EGRESS_IP'

cmd/containerboot/main.go
```diff
@@ -88,8 +90,9 @@ func main() {
 		AuthKey:         defaultEnvs([]string{"TS_AUTHKEY", "TS_AUTH_KEY"}, ""),
 		Hostname:        defaultEnv("TS_HOSTNAME", ""),
 		Routes:          defaultEnv("TS_ROUTES", ""),
-		ProxyTo:         defaultEnv("TS_DEST_IP", ""),
 		ServeConfigPath: defaultEnv("TS_SERVE_CONFIG", ""),
+		IngressProxyTo:  defaultEnv("TS_DEST_IP", ""),
```
maisem, line 94: nit: leave as `ProxyTo`

cmd/containerboot/main.go
```diff
@@ -312,6 +320,14 @@ authLoop:
 					}
 				}
 			}
+			if cfg.TSEgressIP != "" && (!installedEgressProxy || ipsHaveChanged) && len(n.NetMap.Addresses) > 0 {
+				if err := installEgressForwardingRule(ctx, cfg.TSEgressIP, n.NetMap.Addresses); err != nil {
+					log.Fatalf("installing egress proxy rules: %v", err)
+				}
+				installedEgressProxy = true
```
maisem, line 327: i think the variable isn't really useful anymore
you can make this check be identical to the one above

cmd/containerboot/main.go
maisem: I think `TS_TAILNET_TARGET_IP` is more descript

cmd/containerboot/main.go
irbekrm: true

cmd/k8s-operator/svc.go
```diff
@@ -122,24 +122,34 @@ func (a *ServiceReconciler) maybeProvision(ctx context.Context, logger *zap.Suga
...
-	clusterIPAddr, err := netip.ParseAddr(svc.Spec.ClusterIP)
-	if err != nil {
-		return fmt.Errorf("failed to parse cluster IP: %w", err)
-	}
-
 	sts := &tailscaleSTSConfig{
 		ParentResourceName:  svc.Name,
 		ParentResourceUID:   string(svc.UID),
-		TargetIP:            svc.Spec.ClusterIP,
+		ClusterTargetIP:     svc.Spec.ClusterIP,
```
maisem, line 128: hmm, i think we want to not set this when `TailnetTargetIP` is also set.
We probably want to prefer `AnnotationTailnetTargetIP`, or even bail early/raise events if `AnnotationTailnetTargetIP` is set and there is a selector defined.

Can fix in a follow up though.

maisem:
> I am a bit surprised that we call this a "proxy" even though it does NAT. I don't feel strongly about this, but I thought I would mention it in case other people also think it's a bit weird.

Its technically a L3 proxy

cmd/k8s-operator/operator_test.go
```diff
@@ -153,6 +154,111 @@ func TestLoadBalancerClass(t *testing.T) {
...
+			// on it being set.
+			UID: types.UID("1234-UID"),
+			Annotations: map[string]string{
+				AnnotationTailnetTargetIP: tailnetTargetIP,
+			},
+		},
+		Spec: corev1.ServiceSpec{
+			Type: corev1.ServiceTypeClusterIP,
+			Selector: map[string]string{
+				"foo": "bar",
```
maisem, line 194: i think we want to error out if a selector is defined and the service has `AnnotationTailnetTargetIP`

maisem (approved):

cmd/k8s-operator/operator_test.go
irbekrm: We could ask that users explicitly create a Service with `spec.type` set to `ExternalName` (they will also have to set `spec.externalName` to some fake value) and error out/throw an event on the Service if there are any other fields set such as ports etc. Is it worth doing? It doesn't really matter for us if the selector or ports or different service type are set and it's not really a good user experience on kube if it allows to apply an invalid resource, but then errors out later.

irbekrm:
> [maisem](https://github.com/maisem) approved these changes [10 hours ago](https://github.com/tailscale/tailscale/pull/9031#pullrequestreview-1601292325)

Merged after rebase

cmd/k8s-operator/svc.go
irbekrm: https://github.com/tailscale/tailscale/pull/9155

irbekrm: I've done the easiest thing in https://github.com/tailscale/tailscale/pull/9155

We could exit early in a reconciler if we see a Service that is invalid, but I am not sure about throwing an event against the Service- that will again trigger our own reconciler and we'll be just spamming kube apiserver and blocking our own reconcile loop. Maybe we could set our condition on the Service status- I'll try to do that in a separate PR

irbekrm: @maisem we could set our own status condition on Services like so https://github.com/tailscale/tailscale/commit/fc7267fddb113ecfc9efb9f81af69b7a7aedcb41

```
apiVersion: v1
kind: Service
metadata:
  annotations:
    tailscale.com/expose: "true"
    tailscale.com/ts-tailnet-target-ip: 100.68.29.93
  finalizers:
  - tailscale.com/finalizer
  name: invalid
  namespace: default
spec:
  externalName: placeholder
  type: ExternalName
status:
  conditions:
  - lastTransitionTime: "2023-08-30T10:50:43Z"
    message: Service has both tailscale.com/tailnet-target-ip and tailscale.com/expose
      annotation set.
    reason: InvalidTailscaleService
    status: "False"
    type: TailscaleStatus
```

Issue #5210, opened 2022-07-28
Switch to Go 1.19


bradfitz:
Tracking bug for moving from Go 1.18 to Go 1.19.

bradfitz:
Go 1.19 rebased at https://github.com/tailscale/go/commits/tailscale.go1.19 .... it was the easiest rebase ever. Our 5 tiny commits all cherry-picked cleanly. https://github.com/tailscale/go/commit/4cbd2683f8f31b9498dec50d0180106bb2d8a875 is a git merge -s ours of the tailscale.go1.18 tree for history only but with a git tree of the current golang/go master branch (between Go 1.19rc2 and final on Tuesday)
I plan to switch to this later today for oss+corp, but without permitting any Go 1.19 API usage until ~Tuesday when the final release is out.

bradfitz:
This is done.

PR #9035, opened 2023-08-23
all: replace deprecated ioutil references

### LINKED ISSUE #9034: cleanup: remove deprecated ioutil package
We reference the ioutil package in a few places which is deprecated. We should replace them with the new imports so VSCode stops yelling at me.

marwan-at-work:
This PR removes calls to ioutil library and replaces them with their new locations in the io and os packages.

Fixes #9034

### DIFF
--- a/tstest/iosdeps/iosdeps.go
+++ b/tstest/iosdeps/iosdeps.go
@@ -17,7 +17,6 @@ import (
 	_ "fmt"
 	_ "io"
 	_ "io/fs"
-	_ "io/ioutil"
 	_ "log"
 	_ "math"
 	_ "net"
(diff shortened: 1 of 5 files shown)


tstest/iosdeps/iosdeps.go
```diff
@@ -17,7 +17,6 @@ import (
 	_ "fmt"
 	_ "io"
 	_ "io/fs"
-	_ "io/ioutil"
```
marwan-at-work, line 20: is this ok to remove?

bradfitz: Yes.

bradfitz:
In commit message, change "misc" to "all".

And add `Updates #5210`.

marwan-at-work:
@bradfitz done, thanks!

bradfitz (approved):

Issue #9040, opened 2023-08-23
net/netmon: make ChangeFunc richer, remove "Major" bool eventually


bradfitz:
We just debugging an issue where one of my machines was dropping its map poll a half dozen times over the night due to IPv6 addresses (link local? privacy?) coming & going or something. It got classified as a "major" link change which then made tsdial drop the control connection because it might be bad now.

We should flesh out ChangeFunc's signature so watchers can get more info than just one "major" bit on a change.

/cc @maisem @andrew-d

PR #9043, opened 2023-08-23
net/interfaces: ignore non-Tailscale ULA address changes for state equality

bradfitz:
Updates #9040

### DIFF
--- a/net/interfaces/interfaces.go
+++ b/net/interfaces/interfaces.go
@@ -456,7 +456,8 @@ func UseInterestingInterfaces(i Interface, ips []netip.Prefix) bool {
 }
 
 // UseInterestingIPs is an IPFilter that reports whether ip is an interesting IP address.
-// An IP address is interesting if it is neither a loopback nor a link local unicast IP address.
+// An IP address is interesting if it is not a loopback, link local unicast IP address,
+// or non-Tailscale Unique Local Address.
 func UseInterestingIPs(ip netip.Addr) bool {
 	return isInterestingIP(ip)
 }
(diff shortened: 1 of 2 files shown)


raggi (commented):
- IPv6 PD environments: most likely these addresses are uninteresting.
- IPv6 NAT, second address added: most likely uninteresting.
- IPv6 NAT, address going away: not urgently important, active conns would break. We might need to restun though if we were using v6 endpoints.
- IPv6 NAT, first address added: might be interesting, could be a route to the internet.

How would this affect our state judgements in IPv6 NAT environments?

bradfitz:
Specifically IPv6 NAT with ULA addresses. Do we ever see that?

maisem (approved):

raggi:
> Specifically IPv6 NAT with ULA addresses. Do we ever see that?

Per slack, I found some, it's not easy for us to quantify though.

net/interfaces/interfaces.go
```diff
@@ -456,7 +456,8 @@ func UseInterestingInterfaces(i Interface, ips []netip.Prefix) bool {
 }
 
 // UseInterestingIPs is an IPFilter that reports whether ip is an interesting IP address.
-// An IP address is interesting if it is neither a loopback nor a link local unicast IP address.
+// An IP address is interesting if it is not a loopback, link local unicast IP address,
```
crawshaw, line 459: This isn't a particularly useful definition of interesting. This is saying what the code is doing (and is likely to go out of date). What would be useful here is a description of why it's being done. Why are some IP addresses interesting?

I ask, because my default assumption is "interesting" means "there might be a tailscale node reachable on this." But the problem is, an IPv6 ULA absolutely could have an interesting tailscale node on it. So I do not understand the PR.

bradfitz: I don't like most of this either.

raggi: A rewording of some of the referenced bug and some extra details: we've seen some nodes rebinding more often than we want as a result of ULA range addresses getting added & removed quite regularly, but those additions and removals are not affecting / siblings of changes to a prefix delegated address or link local address. In searching around to answer the review question whether we see ULA-only nodes with working IPv6 NAT I also turned up an example node that seemed to be following the pattern Brad saw directly, which is it had over 20(!!!) ULA addresses which were identical except for the second address byte (inside the ULA prefix).

bradfitz:
This PR's on hold until cleanup https://github.com/tailscale/tailscale/pull/9100 is in. That moves a bunch of stuff around to where it should be, finishing some migrations. (netmon is supposed to be the center of the knowing-the-network-state universe)

PR #9045, opened 2023-08-23
clientupdate/distsign: use distinct PEM types for root/signing keys

awly:
To make key management less error-prone, use different PEM block types for root and signing keys. As a result, separate out most of the Go code between root/signing keys too.

Updates #8760


clientupdate/distsign/distsign.go
```diff
@@ -56,49 +55,90 @@ import (
...
+// be in the same format as returned by GenerateRootKey.
+func ParseRootKey(privKey []byte) (*RootKey, error) {
+	k, err := parsePrivateKey(privKey, pemTypeRootPrivate)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse root key: %w", err)
+	}
+	return &RootKey{k: k}, nil
+}
 
 // SignSigningKeys signs the bundle of public signing keys. The bundle must be
```
noncombatant, line 98: Should we enforce that, by actually calling `ParseSigningKey` on it? (Can you tell I'm getting increasingly paranoid)

clientupdate/distsign/distsign.go
```diff
@@ -56,49 +55,90 @@ import (
...
+	return pem.EncodeToMemory(&pem.Block{
+			Type:  pemTypeSigningPrivate,
+			Bytes: []byte(priv),
+		}), pem.EncodeToMemory(&pem.Block{
+			Type:  pemTypeSigningPublic,
+			Bytes: []byte(pub),
+		}), nil
+}
+
+// ParseSigningKey parses the PEM-encoded private signing key. The key is
```
noncombatant, line 124: Same thing here, re: "must"

clientupdate/distsign/distsign.go
```diff
@@ -284,25 +304,25 @@ func download(url, dst string, limit int64) ([]byte, int64, error) {
...
 	b, rest := pem.Decode(data)
 	if b == nil {
 		return nil, errors.New("failed to decode PEM data")
 	}
 	if len(rest) > 0 {
 		return nil, errors.New("trailing PEM data")
 	}
-	if b.Type != pemTypePrivate {
-		return nil, fmt.Errorf("PEM type is %q, want %q", b.Type, pemTypePrivate)
+	if b.Type != typ {
```
noncombatant, line 315: Aesthetic suggestion, not functional:

```
if b.TypeTag != typeTag {
```

noncombatant: (and below, if you agree)

clientupdate/distsign/distsign_test.go
```diff
@@ -285,13 +285,20 @@ type rootKeyPair struct {
...
-	priv, err := parsePrivateKey(kp.privRaw)
+	privRaw, pubRaw, err := GenerateRootKey()
+	if err != nil {
+		t.Fatalf("GenerateRootKey: %v", err)
+	}
+	kp := keyPair{
+		privRaw: privRaw,
+		pubRaw:  pubRaw,
+	}
+	priv, err := parsePrivateKey(kp.privRaw, pemTypeRootPrivate)
```
noncombatant, line 296: Might be good to write some negative test cases for parsing roots as signing keys and vice versa, and parsing keys with invalid type tags like meaningless strings, near-miss typos, et c.

release/dist/cli/cli.go
```diff
@@ -190,12 +192,25 @@ func parseSigningKey(path string) (crypto.Signer, error) {
 }
 
 var genKeyArgs struct {
+	root     bool
```
noncombatant, line 195: This leaves open the (unlikely) possibility that both might be true, or both false. Arguable, but what do you think of:

```
type KeyType int

const (
  KeyTypeInvalid KeyType = iota
  KeyTypeRoot
  KeyTypeSigning
)

var genKeyArgs struct {
  keyType KeyType
```

Maybe I'm just having `enum` feelings today.

release/dist/cli/cli.go
```diff
@@ -190,12 +192,25 @@ func parseSigningKey(path string) (crypto.Signer, error) {
...
 	privPath string
 	pubPath  string
 }
 
 func runGenKey(ctx context.Context) error {
-	priv, pub, err := distsign.GenerateKey()
+	var pub, priv []byte
+	var err error
+	switch {
+	case genKeyArgs.root && genKeyArgs.signing:
```
noncombatant, line 205: Despite my enum-y feelings, we still need this, of course.

release/dist/cli/cli.go
awly: I was considering doing the enum in the distsign package, but decided to just use different func/type names as a stronger compile-time check.
If we want to keep the flags as `--root/--signing` and not `--type=root|signing`, I don't think an enum here would do much. We'd basically convert the flags to enum, and then convert the enum to function call right after. The intermediate steps seems unnecessary :thinking:

noncombatant (approved):

Issue #7895, opened 2023-04-17
FR: support funnel with k8s operator


asg0451:
### What is the issue?

i started a funnel on a tailscale kubernetes sidecar, but nothing will connect to it. looking at the logs:

```

ts-sidecar boot: 2023/04/17 03:13:07 Startup complete, waiting for shutdown signal
ts-sidecar 2023/04/17 03:13:07 magicsock: derp-1 connected; connGen=1
ts-sidecar 2023/04/17 03:13:07 health("overall"): ok
ts-sidecar 2023/04/17 03:13:12 wgengine: idle peer [Zhyr6] now active, reconfiguring WireGuard
ts-sidecar 2023/04/17 03:13:12 wgengine: Reconfig: configuring userspace WireGuard config (with 1/16 peers)
ts-sidecar 2023/04/17 03:13:12 Accept: TCP{100.113.125.113:35272 > 100.117.132.167:443} 52 tcp non-syn
ts-sidecar 2023/04/17 03:14:31 cert("sproter-wtf.beagle-chickadee.ts.net"): already had ACME account.
ts-sidecar 2023/04/17 03:14:31 cert("sproter-wtf.beagle-chickadee.ts.net"): starting SetDNS call...
ts-sidecar 2023/04/17 03:14:42 cert("sproter-wtf.beagle-chickadee.ts.net"): did SetDNS
ts-sidecar 2023/04/17 03:15:01 cert("sproter-wtf.beagle-chickadee.ts.net"): getCertPEM: context canceled
```

i tried with different host names and nothing works. maybe something is messed up with my account..

### OS

Linux

### OS version

Kubernetes/Ubuntu

### Tailscale version

1.38.4   tailscale commit: 043a34500dd2bb07c34e3b28a56cdbc8b5434454-dirty   go version: go1.20.3-tsddff070

### Bug report

BUG-18ca3f34a86d4f9a0fbda73c544c06739a75d4bed51de23dac4f2dff27e796c3-20230417031900Z-c851ed6f55a07ae5

DentonGentry:
I don't think Funnel works with the k8s operator yet. https://github.com/tailscale/tailscale/issues/502#issuecomment-1414522795

maisem:
Closed accidentally

rodrigc:
@maisem in addition to mentioning latest `unstable`, can you also refer to an actual SHA or 
a label for the operator image with this new funnel feature?
[...]

PR #9048, opened 2023-08-23
cmd/k8s-operator: add initial support for k8s ingress

maisem:
Previously, the operator would only monitor Services and create
a Tailscale StatefulSet which acted as a L3 proxy which proxied
traffic inbound to the Tailscale IP onto the services ClusterIP.

This extends that functionality to also monitor Ingress resources
where the `ingressClassName=tailscale` and similarly creates a
Tailscale StatefulSet, acting as a L7 proxy instead.

Users can override the desired hostname by setting:

```
- tls
  hosts:
  - "foo"
```

Hostnames specified under `rules` are ignored as we only create a single
host. This is emitted as an event for users to see.

Depends on #9074

Updates #7895

### DIFF
--- a/cmd/k8s-operator/ingress.go
+++ b/cmd/k8s-operator/ingress.go
@@ -1,0 +1,233 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+//go:build !plan9
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"go.uber.org/zap"
+	"golang.org/x/exp/slices"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"tailscale.com/ipn"
+	"tailscale.com/types/opt"
+)
+
+type IngressReconciler struct {
+	client.Client
+
+	recorder record.EventRecorder
+	ssr      *tailscaleSTSReconciler
+	logger   *zap.SugaredLogger
+}
+
+func (a *IngressReconciler) Reconcile(ctx context.Context, req reconcile.Request) (_ reconcile.Result, err error) {
+	logger := a.logger.With("ingress-ns", req.Namespace, "ingress-name", req.Name)
+	logger.Debugf("starting reconcile")
+	defer logger.Debugf("reconcile finished")
+
+	ing := new(networkingv1.Ingress)
+	err = a.Get(ctx, req.NamespacedName, ing)
+	if apierrors.IsNotFound(err) {
+		// Request object not found, could have been deleted after reconcile request.
+		logger.Debugf("ingress not found, assuming it was deleted")
+		return reconcile.Result{}, nil
+	} else if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get ing: %w", err)
+	}
+	if !ing.DeletionTimestamp.IsZero() || !a.shouldExpose(ing) {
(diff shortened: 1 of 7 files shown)


cmd/k8s-operator/ingress.go
```diff
@@ -0,0 +1,233 @@
...
+	defer logger.Debugf("reconcile finished")
+
+	ing := new(networkingv1.Ingress)
+	err = a.Get(ctx, req.NamespacedName, ing)
+	if apierrors.IsNotFound(err) {
+		// Request object not found, could have been deleted after reconcile request.
+		logger.Debugf("ingress not found, assuming it was deleted")
+		return reconcile.Result{}, nil
+	} else if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get svc: %w", err)
```
danderson, line 44: failed to get ingress

cmd/k8s-operator/ingress.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	if !ing.DeletionTimestamp.IsZero() || !a.shouldExpose(ing) {
+		logger.Debugf("service is being deleted or should not be exposed, cleaning up")
```
danderson, line 47: ingress is being deleted

cmd/k8s-operator/ingress.go
```diff
@@ -0,0 +1,233 @@
...
+	ing.Finalizers = append(ing.Finalizers[:ix], ing.Finalizers[ix+1:]...)
+	if err := a.Update(ctx, ing); err != nil {
+		return fmt.Errorf("failed to remove finalizer: %w", err)
+	}
+
+	// Unlike most log entries in the reconcile loop, this will get printed
+	// exactly once at the very end of cleanup, because the final step of
+	// cleanup removes the tailscale finalizer, which will make all future
+	// reconciles exit early.
+	logger.Infof("unexposed service from tailnet")
```
danderson, line 77: ingress

cmd/k8s-operator/ingress.go
```diff
@@ -0,0 +1,233 @@
...
+	return nil
+}
+
+// maybeProvision ensures that svc is exposed over tailscale, taking any actions
```
danderson, line 81: ing

cmd/k8s-operator/ingress.go
```diff
@@ -0,0 +1,233 @@
...
+// necessary to reach that state.
+//
+// This function adds a finalizer to svc, ensuring that we can handle orderly
```
danderson, line 84: ing

cmd/k8s-operator/manifests/proxy.yaml
```diff
@@ -2,7 +2,7 @@
 # at build time and then uses to construct Tailscale proxy pods.
 apiVersion: apps/v1
 kind: StatefulSet
-metadata:
+metadata: {}
```
danderson, line 5: ? I guess a linter did this?

cmd/k8s-operator/operator_test.go
```diff
@@ -748,9 +748,9 @@ func expectedSTS(stsName, secretName, hostname, priorityClassName string) *appsv
 							Env: []corev1.EnvVar{
 								{Name: "TS_USERSPACE", Value: "false"},
 								{Name: "TS_AUTH_ONCE", Value: "true"},
-								{Name: "TS_DEST_IP", Value: "10.20.30.40"},
 								{Name: "TS_KUBE_SECRET", Value: secretName},
 								{Name: "TS_HOSTNAME", Value: hostname},
+								{Name: "TS_DEST_IP", Value: "10.20.30.40"},
```
danderson, line 753: Why the reorder? Fine either way.

danderson (approved):

cmd/k8s-operator/operator_test.go
maisem: because its conditionally added now at the end, the differ didn't like it

cmd/k8s-operator/manifests/proxy.yaml
maisem: it seemed like it was missing, and i wasn't sure if yaml just allowed this or not. thought being explicit was nicer

PR #9049, opened 2023-08-23
ipn/ipnauth, ipn/ipnlocal, ipn/ipnserver, ipn/localapi, safesocket: m…

dblohm7:
…ake Windows builds use impersonation to obtain the named pipe's access token and use that for localapi auth

On Windows, the idiomatic way to check access on a named pipe is for the server to impersonate the client on its current OS thread, perform access checks using the client's access token, and then revert the OS thread's access token back to its true self.

The access token is a better representation of the client's rights than just a username/userid check, as it represents the client's effective rights at connection time, which might differ from their normal rights.

This patch updates safesocket to do the aforementioned impersonation, extract the token handle, and then revert the impersonation. We retain the token handle for the remaining duration of the connection (the token continues to be valid even after we have reverted back to self).

Since the token is a property of the connection, I changed ipnauth to wrap the concrete net.Conn to include the token. I then plumbed that change through ipnlocal, ipnserver, and localapi as necessary.

I also added a PermitLocalAdmin flag to the localapi Handler which I intend to use for controlling access to a few new localapi endpoints intended for configuring auto-update.


go.mod
```diff
@@ -362,3 +362,5 @@ require (
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
+
+replace github.com/Microsoft/go-winio => github.com/tailscale/go-winio v0.0.0-20230929170319-cac332cedf3a
```
bradfitz, line 366: we can't use replace in oss go.mod. only in corp.

bradfitz: Well, what's the extent of the fork changes? If it changes the API shape at all, we definitely can't.

If it only fixes some bugs with the existing API we _could_, but then people running e.g. tsnet servers on Windows wouldn't get the fixes. (people using this module from their own go module)

So sadly I think we need to rename in our fork.

dblohm7: Yeah it's an API change. Sadly, the upstream PR was actually accepted, but whoever is in charge of merging it has not done so.

dblohm7: OK, I've renamed our fork.

ipn/ipnserver/server.go
```diff
@@ -202,6 +202,7 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) {
 		lah := localapi.NewHandler(lb, s.logf, s.netMon, s.backendLogID)
 		lah.PermitRead, lah.PermitWrite = s.localAPIPermissions(ci)
 		lah.PermitCert = s.connCanFetchCerts(ci)
+		lah.PermitLocalAdmin = s.connIsLocalAdmin(ci)
```
tylersmalley, line 205: Brad suggested `CallerIsLocalAdmin`, even if inconsistent per https://github.com/tailscale/tailscale/pull/9969#discussion_r1372278915

dblohm7: Okay, updating.

ipn/ipnserver/server.go
```diff
@@ -341,6 +360,26 @@ func (s *Server) connCanFetchCerts(ci *ipnauth.ConnIdentity) bool {
...
+// connIsLocalAdmin reports whether ci has administrative access to the local
+// machine, for whatever that means with respect to the current OS.
+//
+// This returns true only on Windows machines when the client user is a
+// member of the built-in Administrators group (but not necessarily elevated).
+// This is useful because, on Windows, tailscaled itself always runs with
+// elevated rights: we want to avoid privilege escalation for certain mutative operations.
+func (s *Server) connIsLocalAdmin(ci *ipnauth.ConnIdentity) bool {
+	tok, err := ci.WindowsToken()
+	if err != nil {
```
tylersmalley, line 372: We should consider logging this error, and the IsAdministrator check following.

ipn/localapi/localapi.go
```diff
@@ -157,6 +157,12 @@ type Handler struct {
 	// cert fetching access.
 	PermitCert bool
 
+	// CallerIsLocalAdmin is whether the client may write machine-wide state to
+	// the local machine. Currently relevant to Windows where tailscaled runs as
+	// LocalSystem (ie, root) but we want to restrict some localapi operations
```
bradfitz, line 162: (i.e., root)
or
(~root)

bradfitz, line 162: LocalAPI

safesocket/pipe_windows.go
```diff
@@ -3,16 +3,27 @@
...
+	"time"
 
-	"github.com/Microsoft/go-winio"
+	"github.com/tailscale/go-winio"
+	"golang.org/x/sys/windows"
 )
 
 func connect(s *ConnectionStrategy) (net.Conn, error) {
-	return winio.DialPipe(s.path, nil)
+	dl := time.Now().Add(2 * time.Second)
```
bradfitz, line 21: this enough? Our GitHub Actions CI has shown me that even huge timeouts like this regularly trip 😭

safesocket/pipe_windows.go
```diff
@@ -39,5 +50,100 @@ func listen(path string) (net.Listener, error) {
...
+// the Windows access token associated with the connection's client.
+type ConnWithToken struct {
+	net.Conn
+	token windows.Token
+}
+
+func (conn *ConnWithToken) handle() windows.Handle {
+	fd := conn.Conn.(interface {
+		Fd() uintptr
+	})
```
bradfitz, line 66: so this just panics if you use the wrong embedded Conn?

use a comma-ok type assertion to not panic?

pull this interface out into an unexported named type with some comments about who implements it?

safesocket/pipe_windows.go
```diff
@@ -39,5 +50,100 @@ func listen(path string) (net.Listener, error) {
...
 		return nil, fmt.Errorf("namedpipe.Listen: %w", err)
 	}
-	return lc, nil
+	return &listenerWrap{Listener: lc}, nil
+}
+
+// ConnWithToken is an implementation of net.Conn that permits retrieval of
+// the Windows access token associated with the connection's client.
+type ConnWithToken struct {
+	net.Conn
```
bradfitz, line 59: document restrictions on this Conn type? It needs an Fd() uintptr method?

ipn/ipnauth/ipnauth_windows.go
```diff
@@ -6,53 +6,157 @@ package ipnauth
...
+		&h,
+		0,
+		false,
+		windows.DUPLICATE_SAME_ACCESS,
+	); err != nil {
+		return nil, err
+	}
+
+	result := &token{t: windows.Token(h)}
+	runtime.SetFinalizer(result, func(t *token) { t.Close() })
```
bradfitz, line 160: Finalizer's a bit sad (it may not run for hours even) but probably okay.

Is it too expensive to call the various methods to get the properties we have ConnIdentity accessors for and stash away their results in ConnIdentity for later, and then close the token immediately here before returning?

ipn/ipnauth/ipnauth.go
```diff
@@ -25,6 +26,32 @@ import (
 	"tailscale.com/version/distro"
 )
 
+var ErrNotImplemented = errors.New("not implemented for GOOS=" + runtime.GOOS)
+
+type WindowsToken interface {
```
bradfitz, line 31: Needs docs, especially as LocalBackend references this.

// WindowsToken represents a Windows user and their current security
// context.

?

safesocket/pipe_windows.go
dblohm7: It's what the previous `go-winio` API did under the hood. Happy to make this longer!

bradfitz: Naah, this is fine if it's not a change and hasn't caused us CI flakes this far.

dblohm7: I bumped to 20 for now.

ipn/ipnserver/server.go
```diff
@@ -341,6 +360,30 @@ func (s *Server) connCanFetchCerts(ci *ipnauth.ConnIdentity) bool {
...
 }
 
+// connIsLocalAdmin reports whether ci has administrative access to the local
+// machine, for whatever that means with respect to the current OS.
+//
+// This returns true only on Windows machines when the client user is a
+// member of the built-in Administrators group (but not necessarily elevated).
+// This is useful because, on Windows, tailscaled itself always runs with
+// elevated rights: we want to avoid privilege escalation for certain mutative operations.
+func (s *Server) connIsLocalAdmin(ci *ipnauth.ConnIdentity) bool {
```
tylersmalley, line 370: I might be incorrect here, but should we just put the inverse logic of this into the windows section of the `localAPIPermission` func and get rid of `CallerIsLocalAdmin`?

tylersmalley: So if they are not a local admin, they are read-only?

tylersmalley: @bradfitz curious about your thoughts here, I might be missing some nuance.

ipn/ipnauth/ipnauth_windows.go
dblohm7: My concern is that there's a level of indirection here: we have an open handle to the token, but the underlying token itself could be modified by something else while we're running; caching this data could result in incorrect outcomes in the (admittedly slim) chance that the cached data goes stale.

I did implement a `Close` method, so I'll ensure that `WindowsToken` is also an `io.Closer` and see if I can `Close()` all the things, with the finalizer serving only as a backstop.

ipn/ipnserver/server.go
bradfitz: Extending `localAPIPermission` does seem like the right direction. Its signature (returning two bools) is already kinda borderline gross. Adding a third bool would probably be over the line. We could add a new struct (unexported ideally but not required) as the result type.

dblohm7: My memory is really fuzzy on this because I wrote this code so long ago. IIRC I initially did write it as you described, but yanked it out for some reason that now escapes me. I'll take another look to see whether I can remember why I believed this was warranted.

dblohm7: @tylersmalley I think this is minor enough that, given my comment above, I'd prefer dealing with as a follow-up issue. Does that work for you?

tylersmalley: Works for me

bradfitz (commented):
commit title, maybe just: `ipn, safesocket: use Windows token in LocalAPI` ?

ipn/localapi/localapi.go
```diff
@@ -157,6 +157,12 @@ type Handler struct {
 	// cert fetching access.
 	PermitCert bool
 
+	// CallerIsLocalAdmin is whether the client may write machine-wide state to
```
bradfitz, line 160: Start with saying what it actually means before getting too prescriptive about how people might use it. Maybe:

```go
// CallerIsLocalAdmin is whether this handler is being invoked as a result
// of a LocalAPI call from a user who is a local admin of the current machine.
//
// As of 2023-10-25 it is only populated on Windows.
//
// It can be used to restrict LocalAPI operations which should only be run
// by an admin and not e.g. an unprivileged user using Tailscale as
// as configured by by their employer's IT admins. 
```

bradfitz (approved):

safesocket/pipe_windows.go
```diff
@@ -39,5 +50,109 @@ func listen(path string) (net.Listener, error) {
...
+
+func (conn *ConnWithToken) Close() error {
+	if conn.token != 0 {
+		conn.token.Close()
+		conn.token = 0
+	}
+	return conn.Conn.Close()
+}
+
+type listenerWrap struct {
```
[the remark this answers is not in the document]
dblohm7: I can do the former. I don't have access to the concrete type though.

safesocket/pipe_windows.go
```diff
@@ -39,5 +50,109 @@ func listen(path string) (net.Listener, error) {
 	if err != nil {
 		return nil, fmt.Errorf("namedpipe.Listen: %w", err)
 	}
-	return lc, nil
+	return &listenerWrap{Listener: lc}, nil
+}
+
+// ConnWithToken is an implementation of net.Conn that permits retrieval of
+// the Windows access token associated with the connection's client. The
+// embedded net.Conn must be a go-winio PipeConn.
```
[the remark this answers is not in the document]
dblohm7: Not really. `PipeConn` doesn't actually include `Fd()` as part of its interface.

ipn/ipnserver/server.go
```diff
@@ -372,14 +420,28 @@ func (s *Server) addActiveHTTPRequest(req *http.Request, ci *ipnauth.ConnIdentit
...
+			if !errors.Is(err, ipnauth.ErrNotImplemented) {
+				s.logf("error obtaining access token: %v", err)
+			}
+		} else {
+			// Tell the LocalBackend about the identity we're now running as.
+			if err := lb.SetCurrentUser(token); err != nil {
+				token.Close()
+				return nil, err
+			}
+			uid, err := token.UID()
```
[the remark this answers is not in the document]
dblohm7: `SetCurrentUser` needs the full token, but I updated it to return the resolved uid upon success.

PR #9050, opened 2023-08-23
tailcfg: bump capver to 72 to restore UPnP

bradfitz:
Actually fixed in 77ff705545 but that was cherry-picked to a branch
and we don't bump capver in branches.

This tells the control plane that UPnP should be re-enabled going
forward.

Updates #8992


danderson (approved):
Lint failure looks like transient CI issue, change LGTM

creachadair (approved):

creachadair:
Yeah, it looks like GitHub is having a hairball.

PR #9051, opened 2023-08-23
client/web: don't require secure cookies for csrf

willnorris:
Under normal circumstances, you would typically want to keep the default behavior of requiring secure cookies.  In the case of the Tailscale web client, we are regularly serving on localhost (where secure cookies don't really matter), and/or we are behind a reverse proxy running on a network appliance like a NAS or Home Assistant. In those cases, those devices are regularly accessed over local IP addresses without https configured, so would not work with secure cookies.

Updates tailscale/corp#13775

cc @noncombatant @awly

### DIFF
--- a/client/web/web.go
+++ b/client/web/web.go
@@ -77,9 +77,12 @@ func NewServer(devMode bool, lc *tailscale.LocalClient) (s *Server, cleanup func
 		cleanup = s.startDevServer()
 		s.addProxyToDevServer()
 
-		// Create new handler for "/api" requests.
-		// And protect with gorilla csrf.
-		csrfProtect := csrf.Protect(csrfKey())
+		// Create handler for "/api" requests with CSRF protection.
+		// We don't require secure cookies, since the web client is regularly used
+		// on network appliances that are served on local non-https URLs.
+		// The client is secured by limiting the interface it listens on,
+		// or by authenticating requests before they reach the web client.
+		csrfProtect := csrf.Protect(csrfKey(), csrf.Secure(false))
 		s.apiHandler = csrfProtect(&api{s: s})
 	}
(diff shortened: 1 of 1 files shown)


client/web/web.go
```diff
@@ -77,9 +77,12 @@ func NewServer(devMode bool, lc *tailscale.LocalClient) (s *Server, cleanup func
 		cleanup = s.startDevServer()
 		s.addProxyToDevServer()
 
-		// Create new handler for "/api" requests.
-		// And protect with gorilla csrf.
-		csrfProtect := csrf.Protect(csrfKey())
+		// Create handler for "/api" requests with CSRF protection.
+		// We don't require secure cookies, since the web client is regularly used
+		// on network appliances that are served on local non-https URLs.
+		// The client is secured by limiting the interface it listens on,
```
noncombatant, line 83: Is there any way to check or enforce this comment? Something like

```
isOnLocalInterface := CheckForThat();
csrfProtect := csrf.Protect(csrfKey(), csrf.Secure(!isOnLocalInterface))
```

willnorris: not really.  The client/web package just exposes an http.Handler, but it's the caller that connects that to a listener.  The `tailscale web` command listens on localhost by default, but a user could always override that with the `listen` flag.  We could look at the http request, but that tells us how the client was actually accessed, not how it was intended to be accessed.

We could expose a new flag for secure cookies, but it would almost certainly need to default to false to prevent breaking existing users.

noncombatant (approved):

client/web/web.go
willnorris: discussed offline and agreed to move forward with this, but revisit later (tracking in #9057)

PR #9053, opened 2023-08-23
net/netmon: factor out debounce loop, simplify polling impl

bradfitz:
This simplifies some netmon code in prep for other changes.

It breaks up Monitor.debounce into a helper method so locking is easier to read and things unindent, and then it simplifies the polling netmon implementation to remove the redundant stuff that the caller (the Monitor.debounce loop) was already basically doing.

Updates #9040

### DIFF
--- a/net/netmon/polling.go
+++ b/net/netmon/polling.go
@@ -13,7 +13,6 @@ import (
 	"sync"
 	"time"
 
-	"tailscale.com/net/interfaces"
 	"tailscale.com/types/logger"
 )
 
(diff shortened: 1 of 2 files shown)


net/netmon/polling.go
```diff
@@ -68,22 +67,18 @@ func (pm *pollingMon) Receive() (message, error) {
...
 		d = 24 * time.Hour
 	}
-	ticker := time.NewTicker(d)
-	defer ticker.Stop()
-	base := pm.m.InterfaceState()
+	timer := time.NewTimer(d)
+	defer timer.Stop()
 	for {
-		if cur, err := pm.m.interfaceStateUncached(); err == nil && !cur.EqualFiltered(base, interfaces.UseInterestingInterfaces, interfaces.UseInterestingIPs) {
-			return unspecifiedMessage{}, nil
```
bradfitz, line 83: this check now moves to the netmon loop, which already did it anyway

raggi:
This is a good step, but looking at the other Receive() implementations still feels like they end up doing a lot of loopy, concurrency bits for not that much benefit. Perhaps they should move to pushing latest state or something, but clearly that's a larger change.

raggi (approved):

bradfitz:
Yeah, some implementations know exactly what to mutate but mostly they parse only enough to wake up the poller or to know to skip that netlink/etc message.

PR #9055, opened 2023-08-23
client/web: cache csrf key when running in CGI mode

willnorris:
Indicate to the web client when it is running in CGI mode, and cache the csrf key in that case.  Otherwise, continue to create a new key on startup that is used until the service is restarted.

Updates tailscale/corp#13775


client/web/web.go
```diff
@@ -57,20 +58,28 @@ type Server struct {
 	devMode  bool
 	devProxy *httputil.ReverseProxy // only filled when devMode is on
 
+	cgiMode    bool
 	apiHandler http.Handler // csrf-protected api handler
 }
 
+type ServerOpts struct {
```
soniaappasamy, line 65: tiny docs since exported

client/web/web.go
```diff
@@ -57,20 +58,28 @@ type Server struct {
...
+	DevMode     bool
+	CGIMode     bool
+	LocalClient *tailscale.LocalClient
+}
+
 // NewServer constructs a new Tailscale web client server.
 //
 // lc is an optional parameter. When not filled, NewServer
```
soniaappasamy, line 73: Maybe move this doc to be on the `LocalClient` field in `ServerOpts`

client/web/web.go
```diff
@@ -530,13 +539,42 @@ func (s *Server) tailscaleUp(ctx context.Context, st *ipnstate.Status, postData
...
 
-// csrfKey creates a new random csrf token.
-// If an error surfaces during key creation,
-// the error is logged and the active process terminated.
-func csrfKey() []byte {
+// csrfKey returns a key that can be used for CSRF protection.
+// If an error occurs during key creation, the error is logged and the active process terminated.
+// If the server is running in CGI mode, the key is cached to disk and reused between requests.
+// If an error occurs during key storage, the error is logged and the active process terminated.
+func csrfKey(cgiMode bool) []byte {
```
soniaappasamy, line 546: Maybe switch this to `func (s *Server) csrfKey()`? Then could just check `s.cgiMode`

Feels like it can be a receiver now that the key logic is a little more specific to the web client setup.

soniaappasamy (approved):

client/web/web.go
soniaappasamy: (specifically the "lc is an optional..." part)

client/web/web.go

client/web/web.go

PR #9056, opened 2023-08-23
ipn, ipn/ipnlocal: add an in memory serve config

marwan-at-work:
Note: this PR probably breaks a lot of code paths, but it's a super early draft to see what it looks like to have an in memory serve config.

This PR adds a parallel in-memory ServeConfig so that foreground funnels are guaranteed to go away in case of unexpected shutdown

Updates #8489


marwan-at-work:
Closing in favor of https://github.com/tailscale/tailscale/pull/9071 because it's a smaller change. Happy to re-open per the note in the 9071 PR description.

PR #9069, opened 2023-08-24
client/web: refresh on tab focus

soniaappasamy:
Refresh node data when user switches to the web client browser tab. This helps clean up the auth flow where they're sent to another tab to authenticate then return to the original tab, where the data should be refreshed to pick up the login updates.

Updates tailscale/corp#13775

### DIFF
--- a/client/web/src/hooks/node-data.ts
+++ b/client/web/src/hooks/node-data.ts
@@ -38,7 +38,7 @@ export default function useNodeData() {
   const fetchNodeData = useCallback(() => {
     apiFetch("/api/data")
       .then((r) => r.json())
-      .then((data) => setData(data))
+      .then((d) => setData(d))
       .catch((error) => console.error(error))
   }, [setData])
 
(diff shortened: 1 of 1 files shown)


client/web/src/hooks/node-data.ts
```diff
@@ -114,8 +114,21 @@ export default function useNodeData() {
...
   useEffect(
-    fetchNodeData,
-    // Initial data load.
+    () => {
+      // Initial data load.
+      fetchNodeData()
+
+      // Refresh on browser tab focus.
+      const onVisibilityChange = () => {
+        document.hidden && fetchNodeData()
```
willnorris, line 123: MDN [mentions](https://developer.mozilla.org/en-US/docs/Web/API/Document/hidden) that `document.hidden` is historical, and we should use `document.visibilityState` instead.  Also, isn't this logic backwards?  We want to fetch node data when the document *isn't* hidden?  Or now, according to [these samples](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event#examples), when `document.visibilityState === "visible"` ?

soniaappasamy: whoops yep! good catch

PR #9071, opened 2023-08-24
ipn, ipn/ipnlocal: use Foreground field in ServeConfig

marwan-at-work:
ipn, ipn/ipnlocal: add Foreground field for ServeConfig
    
This PR adds a new field to the serve config that can be used to identify which serves are in "foreground mode" and then can also be used to ensure they do not get persisted to disk so that if Tailscaled gets ungracefully shutdown, the reloaded ServeConfig will not have those ports opened.
    
Updates #8489

### DIFF
--- a/cmd/tailscale/cli/serve_dev.go
+++ b/cmd/tailscale/cli/serve_dev.go
@@ -5,9 +5,9 @@ package cli
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
-	"io"
 	"log"
 	"os"
 	"os/signal"
@@ -16,6 +16,7 @@ import (
 
 	"github.com/peterbourgon/ff/v3/ffcli"
 	"tailscale.com/ipn"
+	"tailscale.com/util/mak"
 )
 
 type execFunc func(ctx context.Context, args []string) error
@@ -30,14 +31,14 @@ var infoMap = map[string]commandInfo{
 		ShortHelp: "Serve content and local servers on your tailnet",
 		LongHelp: strings.Join([]string{
 			"Serve lets you  share a local server securely within your tailnet.",
-			"To share a local server on the internet, use \"tailscale funnel\"",
+			`To share a local server on the internet, use "tailscale funnel"`,
 		}, "\n"),
 	},
 	"funnel": {
 		ShortHelp: "Serve content and local servers on the internet",
 		LongHelp: strings.Join([]string{
 			"Funnel lets you share a local server on the internet using Tailscale.",
-			"To share only within your tailnet, use \"tailscale serve\"",
+			`To share only within your tailnet, use "tailscale serve"`,
 		}, "\n"),
 	},
 }
@@ -134,14 +135,56 @@ func (e *serveEnv) runServeDev(funnel bool) execFunc {
 }
 
 func (e *serveEnv) streamServe(ctx context.Context, req ipn.ServeStreamRequest) error {
-	stream, err := e.lc.StreamServe(ctx, req)
+	watcher, err := e.lc.WatchIPNBus(ctx, ipn.NotifyInitialState)
 	if err != nil {
 		return err
 	}
-	defer stream.Close()
+	defer watcher.Close()
+	n, err := watcher.Next()
+	if err != nil {
+		return err
+	}
+	sessionID := n.SessionID
+	if sessionID == "" {
+		return errors.New("missing SessionID")
+	}
+	sc, err := e.lc.GetServeConfig(ctx)
+	if err != nil {
+		return fmt.Errorf("error getting serve config: %w", err)
+	}
+	if sc == nil {
+		sc = &ipn.ServeConfig{}
+	}
+	setHandler(sc, req, sessionID)
+	err = e.lc.SetServeConfig(ctx, sc)
+	if err != nil {
+		return fmt.Errorf("error setting serve config: %w", err)
+	}
 
-	fmt.Fprintf(os.Stderr, "Serve started on \"https://%s\".\n", strings.TrimSuffix(string(req.HostPort), ":443"))
-	fmt.Fprintf(os.Stderr, "Press Ctrl-C to stop.\n\n")
-	_, err = io.Copy(os.Stdout, stream)
-	return err
+	fmt.Fprintf(os.Stderr, "Funnel started on \"https://%s\".\n", strings.TrimSuffix(string(req.HostPort), ":443"))
+	fmt.Fprintf(os.Stderr, "Press Ctrl-C to stop Funnel.\n\n")
+
+	for {
+		_, err = watcher.Next()
+		if err != nil {
+			if errors.Is(err, context.Canceled) {
+				return nil
+			}
+			return err
+		}
+	}
+}
+
+// setHandler modifies sc to add a Foreground config (described by req) with the given sessionID.
+func setHandler(sc *ipn.ServeConfig, req ipn.ServeStreamRequest, sessionID string) {
+	fconf := &ipn.ServeConfig{}
+	mak.Set(&sc.Foreground, sessionID, fconf)
+	mak.Set(&fconf.TCP, 443, &ipn.TCPPortHandler{HTTPS: true})
+
+	wsc := &ipn.WebServerConfig{}
+	mak.Set(&fconf.Web, req.HostPort, wsc)
+	mak.Set(&wsc.Handlers, req.MountPoint, &ipn.HTTPHandler{
+		Proxy: req.Source,
+	})
+	mak.Set(&fconf.AllowFunnel, req.HostPort, true)
 }
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"maps"
 	"net"
 	"net/http"
 	"net/http/httputil"
@@ -241,8 +242,9 @@ type LocalBackend struct {
 	componentLogUntil       map[string]componentLogState
 
 	// ServeConfig fields. (also guarded by mu)
-	lastServeConfJSON mem.RO              // last JSON that was parsed into serveConfig
-	serveConfig       ipn.ServeConfigView // or !Valid if none
+	lastServeConfJSON   mem.RO              // last JSON that was parsed into serveConfig
+	serveConfig         ipn.ServeConfigView // or !Valid if none
+	activeWatchSessions set.Set[string]     // of WatchIPN SessionID
 
 	serveListeners     map[netip.AddrPort]*serveListener // addrPort => serveListener
 	serveProxyHandlers sync.Map                          // string (HTTPHandler.Proxy) => *httputil.ReverseProxy
@@ -301,23 +303,24 @@ func NewLocalBackend(logf logger.Logf, logID logid.PublicID, sys *tsd.System, lo
 	clock := tstime.StdClock{}
 
 	b := &LocalBackend{
-		ctx:            ctx,
-		ctxCancel:      cancel,
-		logf:           logf,
-		keyLogf:        logger.LogOnChange(logf, 5*time.Minute, clock.Now),
-		statsLogf:      logger.LogOnChange(logf, 5*time.Minute, clock.Now),
-		sys:            sys,
-		e:              e,
-		dialer:         dialer,
-		store:          store,
-		pm:             pm,
-		backendLogID:   logID,
-		state:          ipn.NoState,
-		portpoll:       portpoll,
-		em:             newExpiryManager(logf),
-		gotPortPollRes: make(chan struct{}),
-		loginFlags:     loginFlags,
-		clock:          clock,
+		ctx:                 ctx,
+		ctxCancel:           cancel,
+		logf:                logf,
+		keyLogf:             logger.LogOnChange(logf, 5*time.Minute, clock.Now),
+		statsLogf:           logger.LogOnChange(logf, 5*time.Minute, clock.Now),
+		sys:                 sys,
+		e:                   e,
+		dialer:              dialer,
+		store:               store,
+		pm:                  pm,
+		backendLogID:        logID,
+		state:               ipn.NoState,
+		portpoll:            portpoll,
+		em:                  newExpiryManager(logf),
+		gotPortPollRes:      make(chan struct{}),
+		loginFlags:          loginFlags,
+		clock:               clock,
+		activeWatchSessions: make(set.Set[string]),
 	}
 
 	netMon := sys.NetMon.Get()
@@ -1956,6 +1959,7 @@ func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWa
 	var ini *ipn.Notify
 
 	b.mu.Lock()
+	b.activeWatchSessions.Add(sessionID)
 
 	const initialBits = ipn.NotifyInitialState | ipn.NotifyInitialPrefs | ipn.NotifyInitialNetMap
 	if mask&initialBits != 0 {
@@ -1981,6 +1985,7 @@ func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWa
 	defer func() {
 		b.mu.Lock()
 		delete(b.notifyWatchers, handle)
+		delete(b.activeWatchSessions, sessionID)
 		b.mu.Unlock()
 	}()
 
@@ -2011,7 +2016,9 @@ func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWa
 		go b.pollRequestEngineStatus(ctx)
 	}
 
-	defer b.DeleteForegroundSession(sessionID) // TODO(marwan-at-work): check err
+	// TODO(marwan-at-work): check err
+	// TODO(marwan-at-work): streaming background logs?
+	defer b.DeleteForegroundSession(sessionID)
 
 	for {
 		select {
@@ -2776,7 +2783,7 @@ func (b *LocalBackend) SetPrefs(newp *ipn.Prefs) {
 // doesn't affect security or correctness. And we also don't expect people to
 // modify their ServeConfig in raw mode.
 func (b *LocalBackend) wantIngressLocked() bool {
-	return b.serveConfig.Valid() && b.serveConfig.AllowFunnel().Len() > 0
+	return b.serveConfig.Valid() && b.serveConfig.HasAllowFunnel()
 }
 
 // setPrefsLockedOnEntry requires b.mu be held to call it, but it
@@ -4092,6 +4099,10 @@ func (b *LocalBackend) setDebugLogsByCapabilityLocked(nm *netmap.NetworkMap) {
 	}
 }
 
+// reloadServeConfigLocked reloads the serve config from the store or resets the
+// serve config to nil if not logged in. The "changed" parameter, when false, instructs
+// the method to only run the reset-logic and not reload the store from memory to ensure
+// foreground sessions are not removed if they are not saved on disk.
 func (b *LocalBackend) reloadServeConfigLocked(prefs ipn.PrefsView) {
 	if b.netMap == nil || !b.netMap.SelfNode.Valid() || !prefs.Valid() || b.pm.CurrentProfile().ID == "" {
 		// We're not logged in, so we don't have a profile.
@@ -4100,6 +4111,7 @@ func (b *LocalBackend) reloadServeConfigLocked(prefs ipn.PrefsView) {
 		b.serveConfig = ipn.ServeConfigView{}
 		return
 	}
+
 	confKey := ipn.ServeConfigKey(b.pm.CurrentProfile().ID)
 	// TODO(maisem,bradfitz): prevent reading the config from disk
 	// if the profile has not changed.
@@ -4119,6 +4131,12 @@ func (b *LocalBackend) reloadServeConfigLocked(prefs ipn.PrefsView) {
 		b.serveConfig = ipn.ServeConfigView{}
 		return
 	}
+
+	// remove inactive sessions
+	maps.DeleteFunc(conf.Foreground, func(s string, sc *ipn.ServeConfig) bool {
+		return !b.activeWatchSessions.Contains(s)
+	})
+
 	b.serveConfig = conf.View()
 }
 
@@ -4136,7 +4154,7 @@ func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked(prefs ipn.
 	b.reloadServeConfigLocked(prefs)
 	if b.serveConfig.Valid() {
 		servePorts := make([]uint16, 0, 3)
-		b.serveConfig.TCP().Range(func(port uint16, _ ipn.TCPPortHandlerView) bool {
+		b.serveConfig.RangeOverTCPs(func(port uint16, _ ipn.TCPPortHandlerView) bool {
 			if port > 0 {
 				servePorts = append(servePorts, uint16(port))
 			}
@@ -4169,7 +4187,7 @@ func (b *LocalBackend) setServeProxyHandlersLocked() {
 		return
 	}
 	var backends map[string]bool
-	b.serveConfig.Web().Range(func(_ ipn.HostPort, conf ipn.WebServerConfigView) (cont bool) {
+	b.serveConfig.RangeOverWebs(func(_ ipn.HostPort, conf ipn.WebServerConfigView) (cont bool) {
 		conf.Handlers().Range(func(_ string, h ipn.HTTPHandlerView) (cont bool) {
 			backend := h.Proxy()
 			if backend == "" {
--- a/ipn/ipnlocal/serve.go
+++ b/ipn/ipnlocal/serve.go
@@ -274,111 +274,6 @@ func (b *LocalBackend) DeleteForegroundSession(sessionID string) error {
 	return b.setServeConfigLocked(sc)
 }
 
-// StreamServe opens a stream to write any incoming connections made
-// to the given HostPort out to the listening io.Writer.
-//
-// If Serve and Funnel were not already enabled for the HostPort in the ServeConfig,
-// the backend enables it for the duration of the context's lifespan and
-// then turns it back off once the context is closed. If either are already enabled,
-// then they remain that way but logs are still streamed
-//
-// TODO(marwan-at-work): this whole endpoint will be
-// deleted in a follow up PR in favor of WatchIPNBus
-func (b *LocalBackend) StreamServe(ctx context.Context, w io.Writer, req ipn.ServeStreamRequest) (err error) {
-	f, ok := w.(http.Flusher)
-	if !ok {
-		return errors.New("writer not a flusher")
-	}
-	f.Flush()
-
-	port, err := req.HostPort.Port()
-	if err != nil {
-		return err
-	}
-
-	// Turn on Funnel for the given HostPort.
-	sc := b.ServeConfig().AsStruct()
-	if sc == nil {
-		sc = &ipn.ServeConfig{}
-	}
-	setHandler(sc, req)
-	if err := b.SetServeConfig(sc); err != nil {
-		return fmt.Errorf("errro setting serve config: %w", err)
-	}
-	// Defer turning off Funnel once stream ends.
-	defer func() {
-		sc := b.ServeConfig().AsStruct()
-		deleteHandler(sc, req, port)
-		err = errors.Join(err, b.SetServeConfig(sc))
-	}()
-
-	select {
-	case <-ctx.Done():
-		// Triggered by foreground `tailscale funnel` process
-		// (the streamer) getting closed, or by turning off Tailscale.
-	}
-
-	return nil
-}
-
-func setHandler(sc *ipn.ServeConfig, req ipn.ServeStreamRequest) {
-	if sc.TCP == nil {
-		sc.TCP = make(map[uint16]*ipn.TCPPortHandler)
-	}
-	if _, ok := sc.TCP[443]; !ok {
-		sc.TCP[443] = &ipn.TCPPortHandler{
-			HTTPS: true,
-		}
-	}
-	if sc.Web == nil {
-		sc.Web = make(map[ipn.HostPort]*ipn.WebServerConfig)
-	}
-	wsc, ok := sc.Web[req.HostPort]
-	if !ok {
-		wsc = &ipn.WebServerConfig{}
-		sc.Web[req.HostPort] = wsc
-	}
-	if wsc.Handlers == nil {
-		wsc.Handlers = make(map[string]*ipn.HTTPHandler)
-	}
-	wsc.Handlers[req.MountPoint] = &ipn.HTTPHandler{
-		Proxy: req.Source,
-	}
-	if req.Funnel {
-		if sc.AllowFunnel == nil {
-			sc.AllowFunnel = make(map[ipn.HostPort]bool)
-		}
-		sc.AllowFunnel[req.HostPort] = true
-	}
-}
-
-func deleteHandler(sc *ipn.ServeConfig, req ipn.ServeStreamRequest, port uint16) {
-	delete(sc.AllowFunnel, req.HostPort)
-	if sc.TCP != nil {
-		delete(sc.TCP, port)
-	}
-	if sc.Web == nil {
-		return
-	}
-	if sc.Web[req.HostPort] == nil {
-		return
-	}
-	wsc, ok := sc.Web[req.HostPort]
-	if !ok {
-		return
-	}
-	if wsc.Handlers == nil {
-		return
-	}
-	if _, ok := wsc.Handlers[req.MountPoint]; !ok {
-		return
-	}
-	delete(wsc.Handlers, req.MountPoint)
-	if len(wsc.Handlers) == 0 {
-		delete(sc.Web, req.HostPort)
-	}
-}
-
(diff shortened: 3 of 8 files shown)


ipn/serve.go
```diff
@@ -39,6 +40,27 @@ type ServeConfig struct {
 	AllowFunnel map[HostPort]bool `json:",omitempty"`
 }
 
+// StripEphemeral ensures that ephemeral tcp connections
```
bradfitz, line 43: TCP

ipn/serve.go
```diff
@@ -39,6 +40,27 @@ type ServeConfig struct {
...
+// stay in memeory only by removing both tcp and web handlers
```
bradfitz, line 44: typo: memory

TCP

ipn/serve.go
```diff
@@ -39,6 +40,27 @@ type ServeConfig struct {
...
+// where the Ephemeral field is set to true.
+func (s *ServeConfig) StripEphemeral() *ServeConfig {
+	conf := s.Clone()
```
bradfitz, line 47: oh, you return a copy. document that?

ipn/ipnlocal/local.go
```diff
@@ -4067,14 +4067,17 @@ func (b *LocalBackend) setDebugLogsByCapabilityLocked(nm *netmap.NetworkMap) {
 	}
 }
 
-func (b *LocalBackend) reloadServeConfigLocked(prefs ipn.PrefsView) {
+func (b *LocalBackend) reloadServeConfigLocked(prefs ipn.PrefsView, changed bool) {
```
bradfitz, line 4070: what is changed?

bradfitz (commented):
What about having a second LocalBackend serveConfig field (serveConfigEphemeral) that's tried first and if that doesn't match anything, then serveConfig is used?

marwan-at-work:
> What about having a second LocalBackend serveConfig field (serveConfigEphemeral) that's tried first and if that doesn't match anything, then serveConfig is used?

@bradfitz That was my approach in https://github.com/tailscale/tailscale/pull/9056 per the PR description above. Happy to switch back to it but there are more open questions like how to resolve conflicts and show a merged view back to the user.

ipn/ipnlocal/local.go
marwan-at-work: @bradfitz I noticed that `SetServeConfig` saves to disk, then calls `setTCPPortsInterceptedFromNetmapAndPrefsLocked` which reloads the entire config back from disk. Which has two issues: 
1. Feels inefficient, because we have the config in memory, why read it back from disk again? Not sure, but maybe [this comment](https://github.com/tailscale/tailscale/blob/94304819263b0553d7a93d2c061e8f824233456d/ipn/ipnlocal/local.go#L4079) is related here. 
2. More importantly, this would end up removing the in-memory config so the `changed` bool says "don't actually reload the config from disk" but run the rest of the code, like validation checks or w/e it's doing. Definitely happy to hear better suggestions...or to actually be educated on what this code is doing cuz im flying a lil blind :)

ipn/serve.go
```diff
@@ -38,27 +37,11 @@ type ServeConfig struct {
...
-		tcpHandler := conf.TCP[p]
-		return tcpHandler != nil && tcpHandler.Ephemeral
-	})
-	maps.DeleteFunc(conf.TCP, func(u uint16, th *TCPPortHandler) bool {
-		return th.Ephemeral
-	})
-	return conf
+	// Foreground is a map of an IPN Bus session id to a
+	// foreground serve config. Note that only TCP and Web
+	// are used inside the Foreground map.
```
bradfitz, line 43: we'll want to flesh out this comment more :)

cmd/tailscale/cli/serve_dev.go
```diff
@@ -29,15 +30,15 @@ var infoMap = map[string]commandInfo{
 	"serve": {
 		ShortHelp: "Serve content and local servers on your tailnet",
 		LongHelp: strings.Join([]string{
-			"Serve lets you  share a local server securely within your tailnet.",
-			"To share a local server on the internet, use \"tailscale funnel\"",
+			`Serve lets you  share a local server securely within your tailnet.`,
```
bradfitz, line 33: accidental double space. (no real need for raw quotes)

ipn/serve.go
```diff
@@ -37,6 +37,11 @@ type ServeConfig struct {
 	// AllowFunnel is the set of SNI:port values for which funnel
 	// traffic is allowed, from trusted ingress peers.
 	AllowFunnel map[HostPort]bool `json:",omitempty"`
+
+	// Foreground is a map of an IPN Bus session id to a
```
bradfitz, line 41: ID

ipn/ipnlocal/local.go
```diff
@@ -2014,6 +2011,16 @@ func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWa
 		go b.pollRequestEngineStatus(ctx)
 	}
 
+	if mask&ipn.NotifyServeRequest != 0 {
+		defer func() {
+			sc := b.ServeConfig().AsStruct()
```
bradfitz, line 2016: avoid this AsStruct call (a clone) until needed.

ipn/ipnlocal/local.go
```diff
@@ -2014,6 +2011,16 @@ func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWa
...
+			if sc != nil {
+				delete(sc.Foreground, sessionID)
+				b.SetServeConfig(sc) // TODO(marwan-at-work): check err
```
bradfitz, line 2019: this seems racy

instead add a method next to:

```
func (b *LocalBackend) SetServeConfig(config *ipn.ServeConfig) error {
	b.mu.Lock()
	defer b.mu.Unlock()
```

like `(*LocalBackend) DeleteForegroundServeConfig(sessionID string)`

That does the whole delete & conditional write under the same lock.

ipn/backend.go
```diff
@@ -122,6 +124,10 @@ type Notify struct {
 	ClientVersion *tailcfg.ClientVersion `json:",omitempty"`
 
 	// type is mirrored in xcode/Shared/IPN.swift
+
+	// FunnelRequestLog is a notification that a request
+	// has been sent via the serve config.
+	FunnelRequestLog *RequestAccessLog `json:",omitempty"`
```
bradfitz, line 130: not "funnel" here per chat

cmd/tailscale/cli/serve_dev.go
```diff
@@ -134,14 +135,104 @@ func (e *serveEnv) runServeDev(funnel bool) execFunc {
...
+		}
+		if n.FunnelRequestLog == nil || n.FunnelRequestLog.SessionID != sessionID {
+			continue
+		}
+		bts, _ := json.Marshal(n.FunnelRequestLog)
+		fmt.Printf("%s\n", bts)
+	}
+}
+
+func setHandler2(sc *ipn.ServeConfig, req ipn.ServeStreamRequest, sessionID string) {
```
bradfitz, line 180: 2? what's 1? docs?

marwan-at-work: @bradfitz that was me debugging stuff. deleting.

ipn/ipnlocal/local.go
```diff
@@ -2014,7 +2011,18 @@ func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWa
...
 
-	defer b.DeleteForegroundSession(sessionID) // TODO(marwan-at-work): check err
+	// TODO(marwan-at-work): check err
+	// TODO(marwan-at-work): streaming background logs?
+	defer b.DeleteForegroundSession(sessionID)
+
+	if mask&ipn.NotifyServeRequest == 0 {
+		fn = func(roNotify *ipn.Notify) (keepGoing bool) {
+			if roNotify.RequestAccessLog != nil {
+				return true
```
bradfitz, line 2021: so your strategy here is to try to filter out non-access log things but otherwise send them all to bus?

that's slightly inefficient when there are no interested listeners at all. it'd be nice to keep a reference count of sessionIDs that care and when it's at zero, don't publish an access log record at all.

ipn/serve.go
```diff
@@ -107,16 +107,20 @@ type ServeStreamRequest struct {
...
 // This structure is in development and subject to change.
-type FunnelRequestLog struct {
+type RequestAccessLog struct {
 	Time time.Time `json:",omitempty"` // time of request forwarding
 
-	// SrcAddr is the address that initiated the Funnel request.
+	// SrcAddr is the address that initiated the serve or funnel request.
 	SrcAddr netip.AddrPort `json:",omitempty"`
 
+	// SessionID if not empty means this request was
```
bradfitz, line 120: , if not empty,

... then end with a period.

ipn/ipnlocal/serve.go
```diff
@@ -271,154 +276,63 @@ func (b *LocalBackend) DeleteForegroundSession(sessionID string) error {
...
-	if err != nil {
-		return err
-	}
-
-	// Turn on Funnel for the given HostPort.
-	sc := b.ServeConfig().AsStruct()
-	if sc == nil {
-		sc = &ipn.ServeConfig{}
+// shouldDeleteFunnel returns the port and true if:
+// 1. This foregrounc has a TCP port, that
```
bradfitz, line 290: typo foreground

but what's "this foreground"?

ipn/ipnlocal/serve.go
```diff
@@ -430,9 +344,9 @@ func (b *LocalBackend) maybeLogServeConnection(destPort uint16, srcAddr netip.Ad
 		}
 	}
 
-	for _, stream := range streamers {
-		stream(log)
-	}
+	b.send(ipn.Notify{
+		RequestAccessLog: &log,
```
bradfitz, line 348: move this `&` to line 331

ipn/ipnlocal/serve.go
```diff
@@ -271,154 +276,63 @@ func (b *LocalBackend) DeleteForegroundSession(sessionID string) error {
...
-	if sc.Web[req.HostPort] == nil {
-		return
-	}
-	wsc, ok := sc.Web[req.HostPort]
-	if !ok {
-		return
-	}
-	if wsc.Handlers == nil {
-		return
+func (b *LocalBackend) maybeLogServeConnection(destPort uint16, srcAddr netip.AddrPort, sessionID string) {
```
bradfitz, line 330: document this? can sessionID be empty?

is this only called if the sessionID is alive? Is the session necessarily subscribing to access log traffic?

ipn/ipnlocal/serve.go
```diff
@@ -504,83 +418,93 @@ func (b *LocalBackend) tcpHandlerForServe(dport uint16, srcAddr netip.AddrPort)
...
-				return context.WithValue(context.Background(), serveHTTPContextKey{}, &serveHTTPContext{
-					SrcAddr:  srcAddr,
-					DestPort: dport,
-				})
-			},
-		}
-		if tcph.HTTPS() {
-			hs.TLSConfig = &tls.Config{
-				GetCertificate: b.getTLSServeCertForPort(dport),
+	f := func(tcpCfg views.MapFn[uint16, *ipn.TCPPortHandler, ipn.TCPPortHandlerView], sessionID string) (handler func(net.Conn) error) {
```
bradfitz, line 421: "f" is indeed a function, but what's it do? lot of code & far away from its use. could use a better name and/or comment.

ipn/ipnlocal/serve.go
```diff
@@ -700,40 +624,42 @@ func (b *LocalBackend) addTailscaleIdentityHeaders(r *httputil.ProxyRequest) {
...
-		io.WriteString(w, s)
-		return
-	}
-	if v := h.Path(); v != "" {
-		b.serveFileOrDirectory(w, r, v, mountPoint)
-		return
-	}
-	if v := h.Proxy(); v != "" {
-		p, ok := b.serveProxyHandlers.Load(v)
+func (b *LocalBackend) serveWebHandler(sessionID string) http.HandlerFunc {
```
bradfitz, line 627: docs + rename to newServeWebHandlerForSession? and document that sessionID can be empty and what tha means
```suggestion
func (b *LocalBackend) serveWebHandler(sessionID string) http.HandlerFunc {
```
```suggestion
func (b *LocalBackend) serveWebHandler(sessionID string) http.HandlerFunc {
```

ipn/ipnlocal/serve.go
bradfitz: and is this sessionID also able to be empty?

ipn/backend.go
```diff
@@ -66,6 +66,8 @@ const (
 	NotifyInitialNetMap // if set, the first Notify message (sent immediately) will contain the current NetMap
 
 	NotifyNoPrivateKeys // if set, private keys that would normally be sent in updates are zeroed out
+
+	NotifyServeRequest // if set, Serve requests will be sent to the watcher
```
bradfitz, line 70: if set, RequestAccessLog messages from serve will be sent to the watcher

ipn/ipnlocal/serve.go
marwan-at-work: the foreground whose sessionID is being passed to this function

ipn/ipnlocal/serve.go
marwan-at-work: Adding the following docs: 

```
// maybeLogServeConnection creates a RequestAccessLog and sends it to any watchers
// who are subscribed to ipn.NotifyServeRequest. SessionID, when not empty, indicates
// that this log is intended for a specific foreground serve session, otherwise it may be
// a background serve whose logs are being followed.
```

That said, I'm not sure how WatchIPNBus will work for when you want to stream the request logs of a background session?

ipn/ipnlocal/local.go
marwan-at-work: I thought this would never send messages to uninterested listeners (we can also be more efficient to inverse this and never send anything to interested listeners that don't have a non-nil *RequestAccessLog.) 

In any case, not sure how keeping a reference count would filter out sending those messages to GUIs/uninterested clients?

ipn/ipnlocal/local.go
bradfitz: Do we still need changed? It's still not documented here.

ipn/ipnlocal/local.go
```diff
@@ -4129,20 +4140,27 @@ func (b *LocalBackend) reloadServeConfigLocked(prefs ipn.PrefsView) {
 // the ports that tailscaled should handle as a function of b.netMap and b.prefs.
 //
 // b.mu must be held.
-func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked(prefs ipn.PrefsView) {
+func (b *LocalBackend) setTCPPortsInterceptedFromNetmapAndPrefsLocked(prefs ipn.PrefsView, changed bool) {
```
bradfitz, line 4143: changed isn't documented. do we need it?

ipn/ipnlocal/local.go
```diff
@@ -4172,29 +4190,36 @@ func (b *LocalBackend) setServeProxyHandlersLocked() {
...
-			backend := h.Proxy()
-			if backend == "" {
-				// Only create proxy handlers for servers with a proxy backend.
-				return true
-			}
-			mak.Set(&backends, backend, true)
-			if _, ok := b.serveProxyHandlers.Load(backend); ok {
-				return true
-			}
+	f := func(w views.MapFn[ipn.HostPort, *ipn.WebServerConfig, ipn.WebServerConfigView]) {
```
bradfitz, line 4193: what is f? rename/document.

ipn/ipnlocal/serve.go
```diff
@@ -239,17 +240,21 @@ func (b *LocalBackend) setServeConfigLocked(config *ipn.ServeConfig) error {
...
-		bs = j
+		b.serveConfig = config.View()
+		b.lastServeConfJSON = mem.B(bs)
 	}
 	if err := b.store.WriteState(confKey, bs); err != nil {
 		return fmt.Errorf("writing ServeConfig to StateStore: %w", err)
 	}
 
-	b.setTCPPortsInterceptedFromNetmapAndPrefsLocked(b.pm.CurrentPrefs())
+	b.setTCPPortsInterceptedFromNetmapAndPrefsLocked(b.pm.CurrentPrefs(), false)
```
bradfitz, line 257: boolean literals are hard to read. false what?

ipn/ipnlocal/serve.go
bradfitz: > how WatchIPNBus will work for when you want to stream the request logs of a background session?

In what way? You can not filter the access log stream if the CLI client wants all of them.

ipn/ipnlocal/local.go
marwan-at-work: it depends on whether we remove foreground on save, or on tailscaled-startup. Neither of which are happening, but if on startup, then we need to make sure this happens top level and not in this method. Any preference? 

Either way, I'll push the doc update.

ipn/ipnlocal/local.go
marwan-at-work: @bradfitz see my comment above

ipn/ipnlocal/serve.go
marwan-at-work: @bradfitz see my comment above, we can remove but clean up on tailscaled-startup, but I could use some help locating where that needs to happen?

ipn/ipnlocal/serve.go
```diff
@@ -504,83 +420,93 @@ func (b *LocalBackend) tcpHandlerForServe(dport uint16, srcAddr netip.AddrPort)
...
-				return context.WithValue(context.Background(), serveHTTPContextKey{}, &serveHTTPContext{
-					SrcAddr:  srcAddr,
-					DestPort: dport,
-				})
-			},
-		}
-		if tcph.HTTPS() {
-			hs.TLSConfig = &tls.Config{
-				GetCertificate: b.getTLSServeCertForPort(dport),
+	findHandler := func(tcpCfg views.MapFn[uint16, *ipn.TCPPortHandler, ipn.TCPPortHandlerView], sessionID string) (handler func(net.Conn) error) {
```
bradfitz, line 423: either document it can return nil or make it a comma-ok result to make that clear?

the type of `tcpCfg` is pretty unfortunate. Maybe we can pull that out to a local type alias, like:

```
    type ServeConfigTCPView = views.MapFn[uint16, *ipn.TCPPortHandler, ipn.TCPPortHandlerView]
``

Or maybe even move that type alias into the `ipn` package exported?

File: ipn/ipnlocal/serve.go [hunk 24]
```diff
@@ -271,154 +276,67 @@ func (b *LocalBackend) DeleteForegroundSession(sessionID string) error {
...
-	// Turn on Funnel for the given HostPort.
-	sc := b.ServeConfig().AsStruct()
-	if sc == nil {
-		sc = &ipn.ServeConfig{}
+// shouldDeleteFunnel returns the port and true if:
+// 1. This foreground has a TCP port, that
+// 2. Funnel has it referenced and allowed, and
+// 3. No other foreground or background session has the port
+// Ambiguity: what if another background had funnel on?
+func shouldDeleteFunnel(sc *ipn.ServeConfig, sessionID string) (hp ipn.HostPort, ok bool) {
```
bradfitz, line 294: this func might be a good candidate for a table of tests. It's a pure function of its inputs and it's kinda non-trivial.

ipn/ipnlocal/serve.go
```diff
@@ -430,9 +348,7 @@ func (b *LocalBackend) maybeLogServeConnection(destPort uint16, srcAddr netip.Ad
 		}
 	}
 
-	for _, stream := range streamers {
-		stream(log)
-	}
+	b.send(ipn.Notify{RequestAccessLog: log})
```
bradfitz, line 351: so this whole Line 335-351 is wasteful in the case that nobody is connected to the bus and wanting streaming logs.

that's why I was talking about keeping a refcount of log watchers. You can then skip these allocs & WhoIs lookup & b.send if nobody's streaming logs.

ipn/ipnlocal/serve.go
```diff
@@ -271,154 +276,67 @@ func (b *LocalBackend) DeleteForegroundSession(sessionID string) error {
...
-	if err != nil {
-		return err
-	}
-
-	// Turn on Funnel for the given HostPort.
-	sc := b.ServeConfig().AsStruct()
-	if sc == nil {
-		sc = &ipn.ServeConfig{}
+// shouldDeleteFunnel returns the port and true if:
+// 1. This foreground has a TCP port, that
```
bradfitz, line 290: Document what "this foreground" means. Or change it to say "The sessionID" or "The foreground sessionID" ?

ipn/serve.go
```diff
@@ -38,7 +38,7 @@ type ServeConfig struct {
 	// traffic is allowed, from trusted ingress peers.
 	AllowFunnel map[HostPort]bool `json:",omitempty"`
 
-	// Foreground is a map of an IPN Bus session id to a
+	// Foreground is a map of an IPN Bus session ID to a
 	// foreground serve config. Note that only TCP and Web
 	// are used inside the Foreground map.
```
bradfitz, line 43: why not AllowFunnel too? Doesn't that simplify your whole `shouldDeleteFunnel` code?

ipn/ipnlocal/serve.go
```diff
@@ -844,6 +775,13 @@ func (b *LocalBackend) webServerConfig(hostname string, port uint16) (c ipn.WebS
 	if !b.serveConfig.Valid() {
 		return c, false
 	}
+	b.serveConfig.Foreground().Range(func(k string, v ipn.ServeConfigView) (cont bool) {
+		c, ok = v.Web().GetOk(key)
+		return !ok
+	})
+	if ok {
+		return c, ok
+	}
```
bradfitz, line 784: let's move add a few methods to ipn.ServeConfig to do these lookups that prefer the Foreground and then do the base one?

marwan-at-work: Done

ipn/serve.go
marwan-at-work: Per our chat, we decided to just nest AllowFunnel in the foreground config

ipn/ipnlocal/serve.go
marwan-at-work: Per our chat, we decided to nest AllowFunnel in foreground configs.

ipn/ipnlocal/serve.go
marwan-at-work: Per our chat, we're punting on logs in this PR

ipn/ipnlocal/serve.go
marwan-at-work: Per our chat, we removed this function since we're nested AllowFunnel in the foreground configs.

ipn/ipnlocal/serve.go
marwan-at-work: Ended up having helper methods instead of inlining the iterations.

ipn/ipnlocal/serve.go
```diff
@@ -700,40 +626,45 @@ func (b *LocalBackend) addTailscaleIdentityHeaders(r *httputil.ProxyRequest) {
...
-	if v := h.Path(); v != "" {
-		b.serveFileOrDirectory(w, r, v, mountPoint)
-		return
-	}
-	if v := h.Proxy(); v != "" {
-		p, ok := b.serveProxyHandlers.Load(v)
+// newServeWebHandlerForSession returns an http.HandlerFunc that matches incoming
+// HTTP requests to its corresponding Web Handler in the ServeConfig. The handler
+// also logs requests to the IPN Bus which will include the given sessionID if not empty.
+func (b *LocalBackend) newServeWebHandlerForSession(sessionID string) http.HandlerFunc {
```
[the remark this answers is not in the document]
marwan-at-work: This was reverted back to the old code as we are not passing sessionIDs for request logs in this pr

ipn/ipnlocal/serve.go
marwan-at-work: This was reverted as we now store active sessions and clean them up on reload

ipn/ipnlocal/local.go
marwan-at-work: Moved to iterator methods

ipn/ipnlocal/local.go
marwan-at-work: Removed in favor of active sessions map.

ipn/ipnlocal/serve.go
marwan-at-work: We're punting on streaming request logs, but we should still chat about how to stream logs for both foreground and background sessions.

ipn/ipnlocal/local.go
marwan-at-work: removed sending request logs in the context of this pr

ipn/serve.go
marwan-at-work: Done, added more words. But happy to change/add if needed

cmd/tailscale/cli/serve_dev.go
```diff
@@ -134,14 +134,70 @@ func (e *serveEnv) runServeDev(funnel bool) execFunc {
...
 
-	fmt.Fprintf(os.Stderr, "Serve started on \"https://%s\".\n", strings.TrimSuffix(string(req.HostPort), ":443"))
-	fmt.Fprintf(os.Stderr, "Press Ctrl-C to stop.\n\n")
-	_, err = io.Copy(os.Stdout, stream)
-	return err
+	fmt.Fprintf(os.Stderr, "Funnel started on \"https://%s\".\n", strings.TrimSuffix(string(req.HostPort), ":443"))
+	fmt.Fprintf(os.Stderr, "Press Ctrl-C to stop Funnel.\n\n")
+
+	// TODO(marwan-at-work): watcher closes but the CLI doesn't realize it.
+	// Should we just keep calling Next and ignoring the results?
```
bradfitz, line 167: yes. Otherwise they'll get clogged up server-side and you might get killed for dropping too many.

ipn/ipn_view_helpers.go
```diff
@@ -0,0 +1,101 @@
...
+		var exists bool
+		v.Foreground().Range(func(k string, v ServeConfigView) (cont bool) {
+			exists = v.AllowFunnel().Len() > 0
+			return !exists
+		})
+		return exists
+	}()
+}
+
+// FindFunnel returns true if target exists in at least the background AllowFunnel
```
bradfitz, line 87: Go style is to say "reports whether" instead of "returns true if"

ipn/ipnlocal/local.go
```diff
@@ -241,8 +242,9 @@ type LocalBackend struct {
 	componentLogUntil       map[string]componentLogState
 
 	// ServeConfig fields. (also guarded by mu)
-	lastServeConfJSON mem.RO              // last JSON that was parsed into serveConfig
-	serveConfig       ipn.ServeConfigView // or !Valid if none
+	lastServeConfJSON            mem.RO              // last JSON that was parsed into serveConfig
+	serveConfig                  ipn.ServeConfigView // or !Valid if none
+	activeServeConfigForegrounds map[string]struct{} // active foreground sessions to ensure cleanup on reload
```
bradfitz, line 247: `activeServeForegroundSessions set.Set[string] // of WatchIPN SessionID`

ipn/ipn_view_helpers.go
```diff
@@ -0,0 +1,101 @@
...
+		v.TCP().Range(func(k uint16, v TCPPortHandlerView) (cont bool) {
+			parentCont = f(k, v)
+			return parentCont
+		})
+		return parentCont
+	})
+}
+
+// RangeOverWebs ranges over both background and foreground Webs.
+// If "cont" returns false, it stops iterating immediately and does not check
```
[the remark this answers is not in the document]
marwan-at-work: This is from the underlying `f`, will clarify

ipn/ipn_view_helpers.go
```diff
@@ -0,0 +1,101 @@
...
+// other foreground configs.
+func (v ServeConfigView) RangeOverWebs(f func(_ HostPort, conf WebServerConfigView) bool) {
+	parentCont := true
+	v.Web().Range(func(k HostPort, v WebServerConfigView) (cont bool) {
+		parentCont = f(k, v)
+		return parentCont
+	})
+	v.Foreground().Range(func(k string, v ServeConfigView) (cont bool) {
```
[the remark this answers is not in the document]
marwan-at-work: Will switch, but just FYI the default behavior will be that there should only ever be 1 matching Web (we still need to add validation logic to enforce this). So in that sense, it doesn't matter where we look first. That is until/if we end up with a --shadow flag.

ipn/ipnlocal/serve.go
```diff
@@ -247,7 +247,9 @@ func (b *LocalBackend) setServeConfigLocked(config *ipn.ServeConfig) error {
 	if err := b.store.WriteState(confKey, bs); err != nil {
 		return fmt.Errorf("writing ServeConfig to StateStore: %w", err)
 	}
-
+	for key := range config.Foreground {
+		mak.Set(&b.activeServeConfigForegrounds, key, struct{}{})
```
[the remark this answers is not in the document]
bradfitz: wait, isn't this backwards? I thought the plan was to delete from this map for sessions that are no longer alive, where active is defined in terms of who's connected to the bus, not what's in the config.

As is, you're saying anything saved is considered alive and anything read stays undeleted if it is considered alive... which I guess kinda works out due to ordering: the first load is before the first save and nothing has been added to this set yet. But after that this set is kinda weird and makes me think this could just be a bool.

I'd prefer if we didn't mutate this here and instead kept this activeServeConfigForegrounds map set in sync based on connects and disconnects of bus watchers instead.

ipn/ipnlocal/serve.go
```diff
@@ -643,6 +540,9 @@ func (b *LocalBackend) addTailscaleIdentityHeaders(r *httputil.ProxyRequest) {
 	r.Out.Header.Set("Tailscale-Headers-Info", "https://tailscale.com/s/serve-headers")
 }
 
+// newServeWebHandlerForSession returns an http.HandlerFunc that matches incoming
+// HTTP requests to its corresponding Web Handler in the ServeConfig. The handler
+// also logs requests to the IPN Bus which will include the given sessionID if not empty.
```
bradfitz, line 545: outdated comment

ipn/serve.go
```diff
@@ -37,12 +37,11 @@ type ServeConfig struct {
...
 	AllowFunnel map[HostPort]bool `json:",omitempty"`
 
-	// Foreground is a map of an IPN Bus session id to a
-	// foreground serve config. Note that only TCP and Web
-	// are used inside the Foreground map.
-	//
-	// TODO(marwan-at-work): this is not currently
-	// used. Remove the TODO in the follow up PR.
+	// Foreground is a map of an IPN Bus session ID to a
+	// foreground serve config. This allows the config to specify
```
bradfitz, line 41: ... to an alternate foreground serve config that's valid for the life of that WatchIPNBus session ID. This allows ...

ipn/ipn_view_helpers.go
```diff
@@ -0,0 +1,103 @@
...
+		var exists bool
+		v.Foreground().Range(func(k string, v ServeConfigView) (cont bool) {
+			exists = v.AllowFunnel().Len() > 0
+			return !exists
+		})
+		return exists
+	}()
+}
+
+// FindFunnel reports whether target exists in at least the background AllowFunnel
```
bradfitz, line 89: in either?
in at least one of?

ipn/ipn_view_helpers.go
```diff
@@ -0,0 +1,103 @@
...
+// or any of the foregrounds.
```
bradfitz, line 90: s/foregrounds/foreground configs/ to match line 77 etc.

ipn/ipn_view_helpers.go
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) Tailscale Inc & AUTHORS
```
bradfitz, line 1: instead of filename `ipn_view_helpers.go`, let's just add it to the bottom of ipn/serve.go.

cmd/tailscale/cli/serve_dev.go
```diff
@@ -134,14 +135,56 @@ func (e *serveEnv) runServeDev(funnel bool) execFunc {
...
+	fconf := &ipn.ServeConfig{}
+	mak.Set(&sc.Foreground, sessionID, fconf)
+	mak.Set(&fconf.TCP, 443, &ipn.TCPPortHandler{HTTPS: true})
+
+	wsc := &ipn.WebServerConfig{}
+	mak.Set(&fconf.Web, req.HostPort, wsc)
+	mak.Set(&wsc.Handlers, req.MountPoint, &ipn.HTTPHandler{
+		Proxy: req.Source,
+	})
+	mak.Set(&fconf.AllowFunnel, req.HostPort, true)
```
bradfitz, line 189: so much shorter! ✨

bradfitz (approved):
LGTM, but fix at least the activeServeConfigForegrounds accounting before submitting.

ipn/ipnlocal/serve.go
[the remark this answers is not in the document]
marwan-at-work: Per our chat, we'll set/unset entries in this map from WatchIPNBus

ipn/ipnlocal/serve.go
marwan-at-work: reverting, it never had docs, but i added a little bit

bradfitz (approved):

PR #9073, opened 2023-08-24
cmd/containerboot: reapply known args on restart

maisem:
Previously we would not reapply changes to TS_HOSTNAME etc when then the container restarted and TS_AUTH_ONCE was enabled.

This splits those into two steps login and set, allowing us to only rerun the set step on restarts.

Updates #502

### DIFF
--- a/cmd/containerboot/main.go
+++ b/cmd/containerboot/main.go
@@ -17,7 +17,8 @@
 //   - TS_DEST_IP: proxy all incoming Tailscale traffic to the given
 //     destination.
 //   - TS_TAILSCALED_EXTRA_ARGS: extra arguments to 'tailscaled'.
-//   - TS_EXTRA_ARGS: extra arguments to 'tailscale up'.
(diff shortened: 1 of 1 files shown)


cmd/containerboot/main.go
```diff
@@ -385,32 +392,44 @@ func tailscaledArgs(cfg *settings) []string {
...
+	cmd := exec.CommandContext(ctx, "tailscale", args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("tailscale login failed: %v", err)
+	}
+	return nil
+}
+
+func tailscaleSet(ctx context.Context, cfg *settings) error {
```
danderson, line 414: docstring?

danderson (approved):

cmd/containerboot/main.go
maisem: done

PR #9075, opened 2023-08-24
client/web: refactor and improve synology and qnap request handling

willnorris:
This should be one of the final sets of changes before we can migrate over to the new javascript web client.  This updates how we run the client in CGI environments (specifically on Synology and QNAP) so that they are able to serve the different paths appropriately.

This PR is three commits that should probably be reviewed separately in order.

Manually tested on Synology DSM7 and QNAP devices.

### DIFF
--- a/client/web/api.go
+++ b/client/web/api.go
@@ -20,16 +20,12 @@ type api struct {
 // which protects the handler using gorilla csrf.
 func (a *api) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("X-CSRF-Token", csrf.Token(r))
-	user, err := authorize(w, r)
-	if err != nil {
-		return
-	}
 	path := strings.TrimPrefix(r.URL.Path, "/api")
 	switch path {
 	case "/data":
 		switch r.Method {
 		case httpm.GET:
-			a.s.serveGetNodeDataJSON(w, r, user)
+			a.s.serveGetNodeDataJSON(w, r)
 		case httpm.POST:
 			a.s.servePostNodeUpdate(w, r)
 		default:
--- a/client/web/synology.go
+++ b/client/web/synology.go
@@ -15,6 +15,37 @@ import (
 	"tailscale.com/util/groupmember"
 )
 
+// authorizeSynology authenticates the logged-in Synology user and verifies
+// that they are authorized to use the web client.  It returns true if the
+// request was handled and no further processing is required.
+func authorizeSynology(w http.ResponseWriter, r *http.Request) (handled bool) {
+	if synoTokenRedirect(w, r) {
+		return true
+	}
+
+	// authenticate the Synology user
+	cmd := exec.Command("/usr/syno/synoman/webman/modules/authenticate.cgi")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		http.Error(w, fmt.Sprintf("auth: %v: %s", err, out), http.StatusUnauthorized)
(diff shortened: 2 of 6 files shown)


soniaappasamy (commented):
Going commit-by-commit. First commit with refactoring looks good

client/web/synology.go
```diff
@@ -31,21 +50,17 @@ func synoTokenRedirect(w http.ResponseWriter, r *http.Request) bool {
...
-var req = new XMLHttpRequest();
-req.overrideMimeType("application/json");
-req.open("GET", serverURL + "/webman/login.cgi", true);
-req.onload = function() {
-	var jsonResponse = JSON.parse(req.responseText);
-	var token = jsonResponse["SynoToken"];
-	document.location.href = serverURL + "/webman/3rdparty/Tailscale/?SynoToken=" + token;
-};
-req.send(null);
+  fetch("/webman/login.cgi")
```
soniaappasamy, line 58: This only ever runs once, right?

client/web/web.go
```diff
@@ -100,48 +100,21 @@ func init() {
...
-	return "", nil
-}
-
-func authRedirect(w http.ResponseWriter, r *http.Request) bool {
-	if distro.Get() == distro.Synology {
-		return synoTokenRedirect(w, r)
+		s.serveQNAP(w, r)
+		return
+	default:
+		s.serve(w, r)
```
soniaappasamy: Since the above two eventually land at `s.serve`, maybe we refactor those to be `authorizeQNAP` and `authorizeSynology` and then remove the default here and have everything fall to `s.serve`?

```
var err error
switch distro.Get() {
	case distro.Synology:
		err = s.authorizeSynology(w, r)
        case distro.QNAP:
                err = s.authorizeQnap(w, r)
}
if err != nil {
        http.Error(w, err.Error(), http.StatusForbidden)
	return
}
s.serve(w, r)
```

That way it's very obvious that they all use `serve`. At a quick glance without knowing that, it might seem like more logic differs than just the authorization. Might also be worth a tiny comment about why we only have auth cases for Synology and QNAP.

client/web/api.go
```diff
@@ -20,16 +20,12 @@ type api struct {
...
-	if err != nil {
-		return
-	}
 	path := strings.TrimPrefix(r.URL.Path, "/api")
 	switch path {
 	case "/data":
 		switch r.Method {
 		case httpm.GET:
-			a.s.serveGetNodeDataJSON(w, r, user)
+			a.s.serveGetNodeDataJSON(w, r)
```
soniaappasamy, line 28: Oh! I forgot to add the metrics counter here, mind adding?
```
s.lc.IncrementCounter(context.Background(), "web_client_page_load", 1)
```

soniaappasamy (commented):
Couple small thoughts on the second commit

soniaappasamy (approved):

soniaappasamy:
Third commit LGTM

client/web/synology.go
willnorris: correct

client/web/web.go

client/web/api.go
willnorris: is page_load the right counter to increment here?  That would cause two counts, one for the HTML page, and then one for the data, wouldn't it?

willnorris: (going ahead and submitting without... we can add it later if this is indeed the right counter)

PR #9077, opened 2023-08-24
client/web: add self node cache

soniaappasamy:
Adds a cached self node to the web client Server struct, which will be used from the web client api to verify that request came from the node's own machine (i.e. came from the web client frontend). We'll be using when we switch the web client api over to acting as a proxy to the localapi, to protect against DNS rebinding attacks.

Updates tailscale/corp#13775

### DIFF
--- a/client/web/web.go
+++ b/client/web/web.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/gorilla/csrf"
 	"tailscale.com/client/tailscale"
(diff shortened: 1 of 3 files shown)


willnorris (approved):

soniaappasamy:
Fixed merge conflicts

client/web/web.go
```diff
@@ -74,7 +87,7 @@ type ServerOpts struct {
 }
 
 // NewServer constructs a new Tailscale web client server.
-func NewServer(opts ServerOpts) (s *Server, cleanup func()) {
+func NewServer(ctx context.Context, opts ServerOpts) (s *Server, cleanup func()) {
```
bradfitz, line 90: document the lifetime/use of the context. Is it only for the duration of the NewServer call, or does is it for the lifetime of the server?

client/web/web.go
```diff
@@ -105,6 +127,58 @@ func init() {
...
+
+	watcher, err := s.lc.WatchIPNBus(watchCtx, ipn.NotifyInitialNetMap|ipn.NotifyNoPrivateKeys)
+	if err != nil {
+		log.Fatalf("lost connection to tailscaled: %v", err)
+	}
+	defer watcher.Close()
+
+	// updateSelf grabs the lock and updates s.self.
+	// Then logs if anything changed.
+	updateSelf := func(self tailcfg.NodeView) {
```
bradfitz, line 144: might as well make it a method?

bradfitz (commented):
LGTM

client/web/web.go
soniaappasamy: Should live for the lifetime of the server, will add a doc

PR #9079, opened 2023-08-24
control,tailcfg,wgengine/magicsock: add nodeAttr to enable/disable pe…

valscale:
…er MTU

Add a nodeAttr to enable/disable path MTU discovery to peers and the supporting functions to actually do the enable/disable.

Updates #311


tailcfg/tailcfg.go
```diff
@@ -1984,6 +1985,16 @@ const (
 	// rather than one big /10 CGNAT route. At most one of this or
 	// NodeAttrOneCGNATEnable may be set; if neither are, it's automatic.
 	NodeAttrOneCGNATDisable = "one-cgnat?v=false"
+
+	// NodeAttrPeerMTUEnable makes the client do path MTU discovery to its
```
maisem, line 1989: these aren't imperative commands, only add a single attr `NodeAttrPeerMTUEnable = "peer-mtu"` and use its presence to check if it should be enabled/disabled

ipn/ipnlocal/local.go
```diff
@@ -3081,6 +3081,8 @@ func (b *LocalBackend) authReconfig() {
 //
 // The versionOS is a Tailscale-style version ("iOS", "macOS") and not
 // a runtime.GOOS.
+//
+// XXX doesn't seem to use nm arg?
```
maisem, line 3085: no XXX in submitted code

valscale: You're just the person I wanted to ask! Does it make sense to remove this arg?

maisem: dropping the param sgtm, though do it in its own commit

tailcfg/tailcfg.go
valscale: Here are the stages:

1. PeerMTU disabled on client by default, but turned on by control per tailnet policy.
2. PeerMTU enabled on client by default, but turned off by per-tailnet policy. <- enable flag removed here
3. If there is a bug, PeerMTU disabled by control regardless of per-tailnet policy.

So I think I want two nodeAttrs until step 2, at which point we can remove PeerMTUEnable and have PeerMTUDisable only for catastrophic bugs.

maisem: add the first nodeAttr to enable in step 1
add the second nodeAttr to disable, remove first nodeAttr in step 2
As each step would require a capVer bump, adding those values then makes sense.

If we add two now, ordering of the nodeAttrs delivered to the node ends up mattering which is not desired.

valscale: Okay, that makes sense! Will do when I next have the chance.

wgengine/magicsock/magicsock.go
```diff
@@ -2339,6 +2336,30 @@ func (c *Conn) bindSocket(ruc *RebindingUDPConn, network string, curPortFate cur
 	return fmt.Errorf("failed to bind any ports (tried %v)", ports)
 }
 
+// getSyscallConn retrieves a raw connection out of Conn corresponding to the
+// string network, which must be "udp4" or "udp6". If it cannot get an
+// underlying UDPConn, it returns nil and errNoActiveUDP. Otherwise it returns
+// the result of SyscallConn() on the UDPConn.
+func (c *Conn) getSyscallConn(network string) (syscall.RawConn, error) {
```
raggi, line 2343: We might want to make sure that we enforce that c.pconn4 and c.pconn6 always implement syscall.Conn, and then implement syscall.Conn as needed on our custom conn types.

If we do that we could make this method `connControl(network string, fn func(fd int) error) error` and have it do along the lines of:

```
func (c *Conn) connControl(network string, fn func(fd int) error) error {
  pconn := c.pconn4
  if network == "udp6" {
    pconn = c.pconn6
  }
  sc, ok := pconn.(syscall.Conn)
  if !ok {
    return errors.New("unsupported connection type")
  }
  rc, err := sc.SyscallConn()
  if err != nil {
    return err
  }
  return rc.Control(fn)
}
```

wgengine/magicsock/peermtu.go
```diff
@@ -0,0 +1,89 @@
...
+func (c *Conn) DontFragSetting() bool {
+	return c.getDontFragment("udp4") || c.getDontFragment("udp6")
+}
+
+// ShouldPMTUD returns true if this client should try to enable peer MTU
+// discovery, false otherwise.
+func (c *Conn) ShouldPMTUD() bool {
+	if v, ok := debugEnablePMTUD().Get(); ok {
+		if debugPMTUD() {
+			c.logf("magicsock: peer path MTU discovery set via envknob to %v", v)
```
raggi, line 22: you might want to ensure that all of the pmtu logs share a common phrase, such as consistently doing:

```
c.logf("magicsock: pmtu: ...")
```

Not directly relevant to your PR, but we should also change c.logf sometime to logger.WithPrefix then we can remove our need to manual prefix `magicsock: `

wgengine/magicsock/peermtu.go
```diff
@@ -0,0 +1,89 @@
...
+	newStatus := enable
+	err4 := c.setDontFragment("udp4", enable)
+	err6 := c.setDontFragment("udp6", enable)
+	anySuccess := err4 == nil || err6 == nil
+	noFailures := (err4 == nil || err4 == errNoActiveUDP) && (err6 == nil || err6 == errNoActiveUDP)
+
+	if anySuccess && noFailures {
+		c.logf("magicsock: peer MTU status updated to %v", newStatus)
+	} else {
+		c.logf("magicsock: updating peer MTU status to %v failed (v4: %v, v6: %v), disabling", enable, err4, err6)
```
raggi, line 83: Is it possible that we take this branch because we happen to be mid rebind? If so do we end up permanently applying this state?

wgengine/magicsock/peermtu_darwin.go
```diff
@@ -0,0 +1,52 @@
...
+	optArg := 1
+	if enable == false {
+		optArg = 0
+	}
+	rc, err := c.getSyscallConn(network)
+	if err != nil {
+		return err
+	}
+
+	rc.Control(func(fd uintptr) {
```
raggi, line 31: Control itself might error, in which case the inner doesn't run, and we'd report nil implying success, though we were not successful

wgengine/magicsock/peermtu.go
valscale: Good question! The answer is no since this is only called in two places: initialization, and in reconfig() under wg lock, and rebind has just happened (if it is going to happen at all). I was wondering how to document this since the rest of the things that happen in reconfig() don't have any obvious concurrency-related documentation that I can see?

The actual theory for failure of the setsockopt is ???. I don't know a thing that will cause it to fail that is not a bug in our code (which I did find one already).

On state: we will keep trying to implement the result of ShouldPMTUD() every time we do a reconfig. That's part of why I took the setsockopt out of the rebind() code, since if there was some persistent error we would rebind on every reconfig.

I would love any thoughts on how to best document this stuff!

wgengine/magicsock/peermtu_darwin.go
valscale: Ooh nice catch, will fix.

wgengine/magicsock/peermtu.go
valscale: Done! Agreed on the log prefix change.

wgengine/magicsock/magicsock.go
valscale: Thank you, I wanted to do that but didn't know how! You're now a co-author on this patch. :)

control/controlknobs/controlknobs.go
```diff
@@ -79,6 +83,8 @@ func (k *Knobs) UpdateFromNodeAttributes(selfNodeAttrs []string) {
 			forceBackgroundSTUN = true
 		case tailcfg.NodeAttrDisableDeltaUpdates:
 			disableDeltaUpdates = true
+		case tailcfg.NodeAttrPeerMTUEnable:
+			peerMTUEnable.Set(false)
```
valscale, line 87: Just a quick thank-you to @noncombatant for noticing that this should be true and not false.

wgengine/magicsock/magicsock.go
valscale: I wasn't sure what you meant by "enforce that c.pconn4 and c.pconn6 always implement syscall.Conn". I implemented it as a cast of the underlying pconn to syscall.Conn as above in batchingUDPConn and RebindingUDPConn and returning errUnsupportedConnType if it doesn't succeed, but I could also have extended net.Packetconn to add a Syscall() method I think which would do the check at compile time? Lemme know what you had in mind.

control/controlknobs/controlknobs.go
```diff
@@ -44,6 +44,9 @@ type Knobs struct {
 	// incremental (delta) netmap updates and should treat all netmap
 	// changes as "full" ones as tailscaled did in 1.48.x and earlier.
 	DisableDeltaUpdates atomic.Bool
+
+	// PeerMTUEnable is whether the node should do peer path MTU discovery.
+	PeerMTUEnable syncs.AtomicValue[opt.Bool]
```
twitchyliquid64, line 49: Can I ask about using `syncs.AtomicValue[opt.Bool]` over `atomic.Bool`? It seems like we only ever set it to true and we write it back to `*Knobs` unconditionally, so I'm not sure what the false represents (as opposed to unset or true?)

valscale: Good catch, that's left over from a previous implementation! Thank you. FYI the plan is to replace this with a corresponding PeerMTUDisable when we start enabling it by default.

wgengine/userspace_test.go
```diff
@@ -227,6 +231,86 @@ func TestUserspaceEnginePortReconfig(t *testing.T) {
...
+	}{
+		{wantP: false, wantDF: osDefaultDF, shouldP: "", desc: "after first reconfig"},
+		{wantP: true, wantDF: true, shouldP: "true", desc: "enabling PMTUD, first time"},
+		{wantP: false, wantDF: false, shouldP: "false", desc: "disabling PMTUD"},
+		{wantP: true, wantDF: true, shouldP: "true", desc: "enabling PMTUD, second time"},
+		{wantP: false, wantDF: false, shouldP: "", desc: "returning to default PMTUD"},
+	}
+
+	for _, tt := range tests {
+		t.Logf("--- testing %s", tt.desc)
```
raggi, line 295: You might like `t.Run` here, though you'll want to swap spaces for underscores or similar in the test name, as go replaces space with underscore in the output

raggi (approved):

creachadair (approved):

wgengine/userspace_test.go
valscale: Thanks so much for teaching me that! Fixed.

PR #9080, opened 2023-08-24
clientupdate/distsign/roots: add temporary dev root key

awly:
Adding a root key that signs the current signing key on pkgs.tailscale.com. This key is here purely for development and should be replaced before 1.50 release.

Updates #8760


noncombatant:
Can we embed this key in the code, and write a check like `if weAreRunningInProd && parsedKey == thisDeveloperKey { panic ...` in the code signing path? Or is that too fiddly?

awly:
We embed it via `go:embed`. I hope to completely remove this key once all the signing machinery is working e2e.
Both stable and unstable releases will be signed by the same keys (although we can have separate ones if needed).
So that check, if we add it, will become dead code rather soon :crossed_fingers:

noncombatant (approved):

Issue #5794, opened 2022-09-30
FR: Plan9 support


extrowerk:
### What are you trying to do?

Millions of Plan9 users waiting to be able to enjoy TS on their favorite system.

### How should we solve this?

Port TS to Plan9

### What is the impact of not solving this?

Millions of Plan9 users won't be able to enjoy TS.

### Anything else?

Plan9 have Go ports, so it should be "easy".

DentonGentry:
To set expectations: Tailscale is not likely to undertake a Plan9 port, but would link to a third-party port from the newsletter.

bradfitz:
I'll go a step further: I don't want to maintain a Plan 9 port.

I already was a maintainer of Go for 10 years and had to deal with its Plan 9 support and making its build system test Plan 9 and it was super annoying relative to its benefits. I appreciated that it, along with OSes like Windows, was different enough to challenge any Unix-specific assumptions we were making designing Go APIs but I don't see Tailscale getting that benefit.

I also doubt there are millions of Plan 9 users. I always got the impression it was closer to a dozen or so.

If Tailscale ever hires a full-time engineer who's a Plan 9 aficionado and wants to maintain the Plan 9 port & its CI, then we can reopen this, but for now I'm going to close this as a declined proposal.

9nut:
as the saying goes: "There are dozens of us!! DOZENS!!!"

There is a workaround: I use this regularly to access my home Plan 9 environment from outside. I have a Linux box at home with IP forwarding enabled that runs a tailscale agent configured to be a subnet router (see docs). From my laptop, I boot a 9legacy term vm under qemu that is configured to access the auth and the fs by their IP addresses.

bradfitz:
In a moment of insanity, I made the Tailscale CLI & daemon compile on plan9/amd64. They run, but don't necessarily work. Plenty isn't wired up or defined, like `safesocket` and `paths`. But somebody that knows Plan 9 can probably do that more easily now.

bradfitz:
I talked to @9nut the other day and got him up to speed on what needs to be done here for tailscaled to start to come alive. (mostly safesocket package's connect+listen implementations)

Reopening this since there's some interest and it wasn't much to make it compile at least. I don't intend to run real CI for it. But we can probably keep it compiling.

danderson:
[...]

PR #9082, opened 2023-08-24
all: adjust some build tags for plan9

bradfitz:
I'm not saying it works, but it compiles.

Updates #5794

### DIFF
--- a/cmd/k8s-operator/operator.go
+++ b/cmd/k8s-operator/operator.go
@@ -1,6 +1,8 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
+//go:build !plan9
+
 // tailscale-operator provides a way to expose services running in a Kubernetes
 // cluster to your Tailnet.
 package main
--- a/cmd/tailscaled/tailscaled.go
+++ b/cmd/tailscaled/tailscaled.go
@@ -1,7 +1,7 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
-//go:build go1.19
+//go:build go1.21
 
 // The tailscaled program is the Tailscale client daemon. It's configured
 // and controlled via the tailscale CLI program.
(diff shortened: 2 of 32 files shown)


cmd/k8s-operator/operator.go
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Tailscale Inc & AUTHORS
 // SPDX-License-Identifier: BSD-3-Clause
 
+//go:build !plan9
```
raggi, line 4: This feels burdensome if it grows, I assume this is to make ./... functional?

Really my question becomes a bit different: we have a community member who's been diligently working on making illumos work, and I wonder if we'd accept the same for that?

cmd/tailscaled/tailscaled.go
```diff
@@ -409,7 +411,9 @@ func startIPNServer(ctx context.Context, logf logger.Logf, logID logid.PublicID,
 	// SIGPIPE sometimes gets generated when CLIs disconnect from
 	// tailscaled. The default action is to terminate the process, we
 	// want to keep running.
-	signal.Ignore(syscall.SIGPIPE)
+	if sigPipe != nil {
+		signal.Ignore(sigPipe)
```
raggi, line 415: signal.Ignore takes an argv of signals to ignore, if argv is empty all incoming signals are ignored. The docs don't actually say what nil means. Looks like the meaning varies by platform, on plan9 it ends up signal_ignore'ing `-1`, which I stopped there, but I assume gets rejected somewhere along the way to talking to whoever deals with signal ignores (tty fd?), unless plan9's interface has negative signal numbers?

raggi (approved):

cmd/k8s-operator/operator.go
bradfitz: Yes, this is for `./...` (which CI does).

I'm fine with adding `illumos` here. But that might not need it, being more Unix-y. (actually Unix?)

raggi: Yeah, it may not raise the same thing. I need to follow up with that person sometime and see how their patches are. I know they've sent some and periodically the mention wanting to send more, but have some questions.

cmd/tailscaled/tailscaled.go
bradfitz: signal.Ignore takes `...os.Signal`. If I passed a nil here, it'd be like passing `[]os.Signal{nil}`, a 1-length slice with a nil interface value.

raggi: yeah, a nil interface value gets turned into a -1 by signum(), I stopped there, as I remembered that the place that this num is going to be sent isn't as simple as a syscall interface, and I didn't want to go dig. I assume it's fine, but I have no idea

PR #9083, opened 2023-08-24
client/web: always use new web client; remove old client 

willnorris:
This switches us over to fully using the new react-based web client.

The first commit just allows serving static assets on synology and qnap without authentication.  The second commit does the big switch.  There is a little bit of duplicated code in node-data.ts for attaching the SynoToken on `/api/data` requests.  I looked at moving that into api.ts, but got a little caught up on the various different types that `apiFetch` can accept.


client/web/build/index.html
```diff
@@ -0,0 +1,40 @@
...
+    <title>Tailscale</title>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link rel="shortcut icon" href="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAACAAAAAgCAQAAADZc7J/AAAABGdBTUEAALGPC/xhBQAAACBjSFJNAAB6JgAAgIQAAPoAAACA6AAAdTAAAOpgAAA6mAAAF3CculE8AAAAAmJLR0QA/4ePzL8AAAAHdElNRQflAx4QGA4EvmzDAAAA30lEQVRIx2NgGAWMCKa8JKM4A8Ovt88ekyLCDGOoyDBJMjExMbFy8zF8/EKsCAMDE8yAPyIwFps48SJIBpAL4AZwvoSx/r0lXgQpDN58EWL5x/7/H+vL20+JFxluQKVe5b3Ke5V+0kQQCamfoYKBg4GDwUKI8d0BYkWQkrLKewYBKPPDHUFiRaiZkBgmwhj/F5IgggyUJ6i8V3mv0kCayDAAeEsklXqGAgYGhgV3CnGrwVciYSYk0kokhgS44/JxqqFpiYSZbEgskd4dEBRk1GD4wdB5twKXmlHAwMDAAACdEZau06NQUwAAACV0RVh0ZGF0ZTpjcmVhdGUAMjAyMC0wNy0xNVQxNTo1Mzo0MCswMDowMCVXsDIAAAAldEVYdGRhdGU6bW9kaWZ5ADIwMjAtMDctMTVUMTU6NTM6NDArMDA6MDBUCgiOAAAAAElFTkSuQmCC" />
+    
+    <script type="module" crossorigin src="./assets/index-7005f6ca.js"></script>
+    <link rel="stylesheet" href="./assets/index-d517315e.css">
+  </head>
+  <body>
+    <div class="min-h-screen py-10 flex justify-center items-center" style="display: none">
```
soniaappasamy, line 13: Just realized this has `display: none` on it and we never set it to anything else haha. Maybe we just remove this block for now. We might have accidentally gotten rid of the logic that toggled this before, but IIRC it was only ever doing it for certain versions of IE.

client/web/src/hooks/node-data.ts
```diff
@@ -36,7 +36,16 @@ export default function useNodeData() {
...
   const fetchNodeData = useCallback(() => {
-    apiFetch("api/data")
+    const urlParams = new URLSearchParams(window.location.search)
+    const nextParams = new URLSearchParams()
+    const token = urlParams.get("SynoToken")
+    if (token) {
+      nextParams.set("SynoToken", token)
+    }
+    const search = nextParams.toString()
+    const url = `api/data${search ? `?${search}` : ""}`
```
soniaappasamy, line 46: This kind-of mirrors the logic down in the post function below. Could share some of it. But that can be cleaned up in a follow-up, I'm happy to do that with the localapi proxy work.

client/web/src/index.tsx
```diff
@@ -2,6 +2,9 @@ import React from "react"
 import { createRoot } from "react-dom/client"
 import App from "src/components/app"
 
+declare var window: any;
+window.Tailscale = true
```
soniaappasamy, line 6: Add a little doc that this is used to determine if the react client is built?

client/web/web.go
```diff
@@ -216,25 +205,41 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	case authorize(w, r):
+		// Authenticate and authorize the request for platforms that support it.
+		// Return if the request was processed.
+		return
+	case strings.HasPrefix(r.URL.Path, "/api/"):
+		// Pass API requests through to the API handler.
+		s.apiHandler.ServeHTTP(w, r)
+		return
+	case s.devMode:
+		// When in dev mode, proxy all requests to the Vite dev server.
```
soniaappasamy, line 236: nit
```suggestion
		// When in dev mode, proxy non-api requests to the Vite dev server.
```

soniaappasamy (approved):
LGTM!

client/web/src/hooks/node-data.ts
willnorris: that was what I was referring to in my comment at the top of the PR 😛

client/web/src/index.tsx
willnorris: will do.  Also, is there a better way to do this?  I couldn't come up with another way to detect if we have the client assets.

client/web/build/index.html
willnorris: done

client/web/src/index.tsx
willnorris: done

client/web/web.go

client/web/src/hooks/node-data.ts
soniaappasamy: Oh whoops, I didn't process that! Okay sgtm!

Issue #9084, opened 2023-08-25
netfilter: add missing input accepts for both tailscale0 and the bound port


raggi:
A common user issue is if they have a user configured firewall such as `ufw` running, they have to add explicit rules for:

```
ufw allow in tailscale0
ufw allow in 41644/udp
```

This would address #7332 and improve integration for #391, and avoid repetitions of problems such as #8533.

The first step is easier than the second, as the second requires plumbing rebind outcome back to the current netfilter runner.

```[tasklist]
### Tasks
- [ ] Add a firewall rule that explicitly accepts traffic inbound to tailscale0
- [ ] Add a firewall rule that explicitly accepts traffic to the tailscale UDP sockets
```

raggi:
Analysis from #7332 copying here, as it pertains to both issues and documents the diagnosis:

ufw programs rules via iptables, which may be either iptables-legacy or iptables-nft.

Somewhat common hand rolled ufw rules will block:

- UDP traffic to tailscale itself that does not match an existing flow
- Subnet router traffic to local ip addresses (non-local addresses work fine)
- Exit node traffic to a local ip address (non-local addresses work fine)

This is typically configured with sudo ufw enable when the distro has no other standard rules applied.

One option would be to use an application profile, unfortunately application profiles only appear to be able to solve for port unblocking, not unblocking tun traffic flows. Furthermore it becomes necessary to call the ufw CLI each time we update the application profile, so there's no advantage to this strategy.

## What we should do:

Explicitly add a rule to the firewall that enables $currentPort/udp then that would address the UDP issue, and if we insert the rule as we do our jump rules for routing, then they will take precedence over other rules, and users would no longer need to do ufw allow 41641/udp and similar, as seen in https://tailscale.com/kb/1082/firewall-ports/#my-devices-are-using-a-relay-what-can-i-do-to-help-them-connect-peer-to-peer

Explicitly add a rule to the firewall that allows "in" traffic on tailscale0, again inserted at the top of the chains, then users will no longer need to do ufw allow in on tailscale0 as in https://tailscale.com/kb/1077/secure-server-ubuntu-18-04/?q=sudo%20ufw#step-4-allow-access-over-tailscale

Finally cleanup our KB articles on the topic, as referenced above, none of the locations describing interactions with ufw contain complete guidance for the necessary changes. Once the aforementioned changes are performed as part of our firewall configuration, no manual steps would be required by users in normal operation anymore.

garry-t:
[...]

PR #9085, opened 2023-08-25
util/linuxfw: add missing input rule to the tailscale tun

### LINKED ISSUE #7332: FR: Detect and report when ufw blocks `tailscale0`

https://github.com/tailscale/tailscale/issues/4917#issuecomment-1437684843 reports an exit node not working because ufw blocked it. `tailscaled` could detect when ufw is blocking, and:
- report it as a health check in `tailscale status`
- report it to the coordination server to turn into a warning

### How should we solve this?

_No response_

_No response_

### Anything else?

_No response_

raggi:
Add an explicit accept rule for input to the tun interface, as a mirror to the explicit rule to accept output from the tun interface.

The rule matches any packet in to our tun interface and accepts it, and the rule is positioned and prioritized such that it should be evaluated prior to conventional ufw/iptables/nft rules.

Updates #391
Fixes #7332
Updates #9084


bradfitz (commented):
I'll defer to @danderson on all this.

util/linuxfw/nftables_runner.go
```diff
@@ -877,6 +877,38 @@ func addAcceptOutgoingPacketRule(conn *nftables.Conn, table *nftables.Table, cha
...
 }
 
+// createAcceptIncomingPacketRule creates a rule to accept incoming packets to
+// the given interface.
+func createAcceptIncomingPacketRule(table *nftables.Table, chain *nftables.Chain, tunname string) *nftables.Rule {
+	return &nftables.Rule{
+		Table: table,
+		Chain: chain,
+		Exprs: []expr.Any{
+			&expr.Meta{Key: expr.MetaKeyIIFNAME, Register: 1},
```
danderson, line 887: huh TIL nftables rules are effectively a little BPF-ish virtual machine!

danderson (approved):

PR #9097, opened 2023-08-25
client/web: add localapi proxy

soniaappasamy:
Adds proxy to the localapi from /api/local/ web client endpoint. The localapi proxy is restricted to an allowlist of those actually used by the web client frontend.

Updates tailscale/corp#13775

### DIFF
--- a/client/web/web.go
+++ b/client/web/web.go
@@ -18,11 +18,13 @@ import (
 	"net/netip"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"sync"
 
 	"github.com/gorilla/csrf"
 	"tailscale.com/client/tailscale"
+	"tailscale.com/client/tailscale/apitype"
 	"tailscale.com/envknob"
 	"tailscale.com/ipn"
 	"tailscale.com/ipn/ipnstate"
@@ -251,8 +253,8 @@ func (s *Server) serve(w http.ResponseWriter, r *http.Request) {
 func (s *Server) serveAPI(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("X-CSRF-Token", csrf.Token(r))
 	path := strings.TrimPrefix(r.URL.Path, "/api")
-	switch path {
-	case "/data":
+	switch {
+	case path == "/data":
 		switch r.Method {
 		case httpm.GET:
 			s.serveGetNodeDataJSON(w, r)
@@ -261,6 +263,9 @@ func (s *Server) serveAPI(w http.ResponseWriter, r *http.Request) {
 		default:
 			http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
 		}
+		return
+	case strings.HasPrefix(path, "/local/"):
+		s.proxyRequestToLocalAPI(w, r)
 		return
 	}
 	http.Error(w, "invalid endpoint", http.StatusNotFound)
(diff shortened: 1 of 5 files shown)


soniaappasamy:
Opening in draft for now, still needs tests and has a couple TODOs to clean up. But proxy works checking the host. Hooked up the `/logout` endpoint as a tester.

client/web/web.go
```diff
@@ -262,10 +270,52 @@ func (s *Server) serveAPI(w http.ResponseWriter, r *http.Request) {
...
 		return
+	case strings.HasPrefix(path, "/local/"):
+		s.proxyRequestToLocalAPI(w, r)
+		return
 	}
 	http.Error(w, "invalid endpoint", http.StatusNotFound)
 }
 
+// verifyRequestHost checks the request header to verify that the request
+// came from the node's web client frontend, and not some other source.
```
awly, line 281: this is not exactly true - the request could've come from any web client that can reach this web UI, not just our official frontend.
maybe update to something like `verifyRequestHost checks that the request was intended for a known web client address. This is used to prevent DNS rebinding attacks, where the browser is tricked into opening the web client by some other name.`

client/web/web_test.go
```diff
@@ -62,3 +72,110 @@ func TestQnapAuthnURL(t *testing.T) {
...
+		Hostname: "node.local",
+	}
+	dummyNode := tailcfg.Node{
+		Name:      "node.tailnet.ts.net.",
+		Hostinfo:  dummyNodeHostInfo.View(),
+		Addresses: []netip.Prefix{netip.MustParsePrefix("100.101.102.103/32")},
+	}
+	attachDummyNode := func() { s.updateSelf(dummyNode.View()) }
+
+	tests := []struct {
```
awly, line 106: a couple more test suggestions:
* empty Host header
* malformed IP address like `1.2.3:100`
* IPv6 address
* "garbage string that's neither IP nor hostname"
* node's public IP? that should probably never get exposed, right?
* `localhost` without port number

client/web/web.go
```diff
@@ -262,10 +270,52 @@ func (s *Server) serveAPI(w http.ResponseWriter, r *http.Request) {
...
+// verifyRequestHost reports false if the request's host does not match
+// a known host of the web client.
+func (s *Server) verifyRequestHost(r *http.Request) bool {
```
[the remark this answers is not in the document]
soniaappasamy: Okay Will and I are pivoting a bit away from matching hosts and towards a more solidified auth solution for the web client. I've removed the host verification here and we're restricting this _only_ to logout (i.e. no new localapi access). We've got https://github.com/tailscale/corp/issues/14335 open to start tracking auth work.

client/web/web_test.go
```diff
@@ -62,3 +70,59 @@ func TestQnapAuthnURL(t *testing.T) {
...
+		if err := localapi.Serve(lal); err != nil {
+			t.Error(err)
+		}
+	}()
+	s := &Server{lc: &tailscale.LocalClient{Dial: lal.Dial}}
+
+	tests := []struct {
+		name     string
+		reqPath  string
+		wantResp string
```
noncombatant, line 96: I'd check the `http.Response.StatusCode` too (or instead).

client/web/web.go
```diff
@@ -464,6 +469,61 @@ func (s *Server) tailscaleUp(ctx context.Context, st *ipnstate.Status, postData
...
+// If the localapi path is not included in localapiAllowlist,
+// the request is rejected.
+func (s *Server) proxyRequestToLocalAPI(w http.ResponseWriter, r *http.Request) {
+	path := strings.TrimPrefix(r.URL.Path, "/api/local")
+	if r.URL.Path == path { // missing prefix
+		http.Error(w, "invalid request", http.StatusBadRequest)
+		return
+	}
+	if !slices.Contains(localapiAllowlist, path) {
+		http.Error(w, fmt.Sprintf("%s not allowed from webapi proxy", path), http.StatusForbidden)
```
willnorris, line 486: nit: "webapi" isn't a term we use anywhere.  Maybe "localapi proxy" ? I think the "web" part is implied, given that to get this error you'd be calling the proxy in the web client.

client/web/web_test.go
```diff
@@ -62,3 +70,59 @@ func TestQnapAuthnURL(t *testing.T) {
...
+	}{{
+		name:     "invalid_endpoint",
+		reqPath:  "/not-an-endpoint",
+		wantResp: "invalid endpoint",
+	}, {
+		name:     "allowlist_check",
+		reqPath:  "/local/v0/not-allowlisted",
+		wantResp: "/v0/not-allowlisted not allowed from webapi proxy",
+	}, {
+		name:     "localhost_works",
```
willnorris, line 106: maybe "logout_works" ?  We're not actually testing hostnames anymore.

willnorris (approved):

client/web/web.go
```diff
@@ -464,6 +469,61 @@ func (s *Server) tailscaleUp(ctx context.Context, st *ipnstate.Status, postData
...
+	resp, err := s.lc.DoLocalRequest(req)
+	if err != nil {
+		http.Error(w, err.Error(), resp.StatusCode)
+		return
+	}
+	defer resp.Body.Close()
+
+	// Send response back to web frontend.
+	w.Header().Set("Content-Type", resp.Header.Get("Content-Type"))
+	w.WriteHeader(resp.StatusCode)
```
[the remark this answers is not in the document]
soniaappasamy: From what I can tell, no... But I only have logout running right now, so we might eventually hit some? Since we wrap with the localclient, I think we might be okay, but will revisit when we add more endpoints.

client/web/web.go
soniaappasamy: yea good idea!

client/web/web_test.go
soniaappasamy: nice catch, ty!

soniaappasamy: Will make this one `in_localapi_allowlist` and above `not_in_localapi_allowlist`.

client/web/web.go
```diff
@@ -464,6 +469,61 @@ func (s *Server) tailscaleUp(ctx context.Context, st *ipnstate.Status, postData
...
+	}
+	if !slices.Contains(localapiAllowlist, path) {
+		http.Error(w, fmt.Sprintf("%s not allowed from webapi proxy", path), http.StatusForbidden)
+		return
+	}
+
+	localAPIURL := "http://" + apitype.LocalAPIHost + "/localapi" + path
+	req, err := http.NewRequestWithContext(r.Context(), r.Method, localAPIURL, r.Body)
+	if err != nil {
+		http.Error(w, "failed to construct request", http.StatusInternalServerError)
```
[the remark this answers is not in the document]
soniaappasamy: Yeah good idea - I'll add a clientmetric to track it back. That way we can watch for errors on this. Will add in a follow-up.

client/web/web_test.go
soniaappasamy: done!

client/web/web.go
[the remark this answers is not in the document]
willnorris: also just to clarify, one of the big reasons we can't do hostname checking is that it is very common to have custom local hostnames for NAS devices and such, which is the most common usage of the web client.  Adding proper auth with handle the DNS rebinding concern and a number of other things.

PR #9098, opened 2023-08-25
cmd/tailscale,ipn,types/preftype,wgengine/router: introduce netfilter modes representing new capabilities

raggi:
These are not yet plumbed to work, and are not yet documented in the CLI help output. If accepted they would take the place of the debug flags, and provide a persistent way for users to explicitly switch firewall modes between the integrations that we support (currently iptables and nftables).


raggi:
Split into a new pref & flag, perhaps `netfilter-backend`

PR #9099, opened 2023-08-25
wgengine/router: create netfilter runner in setNetfilterMode

raggi:
This will enable the runner to be replaced as a configuration side effect in a later change.

Updates tailscale/corp#14029

### DIFF
--- a/wgengine/router/router_linux.go
+++ b/wgengine/router/router_linux.go
@@ -200,8 +200,8 @@ type linuxRouter struct {
 	// ipPolicyPrefBase is the base priority at which ip rules are installed.
 	ipPolicyPrefBase int
 
-	nfr netfilterRunner
 	cmd commandRunner
+	nfr netfilterRunner
 }
 
 func newUserspaceRouter(logf logger.Logf, tunDev tun.Device, netMon *netmon.Monitor) (Router, error) {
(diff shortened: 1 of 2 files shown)


raggi:
I'm not 100% keen on this, as it doesn't feel like there's enough of a concurrency guard on `Up()` and particularly `Set()`, but that is a pre-existing condition that needs to be handled separately.

wgengine/router/router_linux.go
```diff
@@ -200,8 +200,8 @@ type linuxRouter struct {
 	// ipPolicyPrefBase is the base priority at which ip rules are installed.
 	ipPolicyPrefBase int
 
-	nfr netfilterRunner
 	cmd commandRunner
+	nfr netfilterRunner
```
danderson, line 204: Why the move? I don't particularly care, just curious.

danderson (approved):

wgengine/router/router_linux.go
raggi: Oh, this went through a few changes, I was originally going to make it atomic / locked, but undid all that as I need to spend far more time scrutinizing our surrounding lock strategy. Right now the new behavior will at least remain consistent with the reconfig path.

PR #9104, opened 2023-08-27
util/deephash: add IncludeFields, ExcludeFields HasherForType Options

bradfitz:
Updates tailscale/corp#6198

### DIFF
--- a/util/deephash/deephash.go
+++ b/util/deephash/deephash.go
@@ -23,11 +23,13 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"encoding/hex"
+	"fmt"
 	"reflect"
 	"sync"
 	"time"
 
 	"tailscale.com/util/hashx"
+	"tailscale.com/util/set"
 )
 
 // There is much overlap between the theory of serialization and hashing.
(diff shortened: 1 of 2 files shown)


util/deephash/deephash.go
```diff
@@ -157,7 +157,25 @@ func HasherForType[T any]() func(*T) Sum {
...
-	hash := lookupTypeHasher(t)
+	return hasherFunc[T](t, lookupTypeHasher(t))
+}
+
+// HasherForStructTypeWithFieldFilter is like HasherForType but requires that T
+// be a struct and accepts a filter function to choose which structs get hashed.
+//
+// This API only permits selecting the top layer of filtering. You can't choose
+// filter paths of fields down through structs of structs.
+func HasherForStructTypeWithFieldFilter[T any](filter func(reflect.StructField) bool) func(*T) Sum {
```
bradfitz, line 168: open for alternative names if this one is too long 😂

maisem: Could do an opt parameter on HasherForType instead

maisem: Like 

`HasherForType[T any](opts ...HasherOpt)`

bradfitz: Could. It does add 3 API symbols rather than 1, though. Do you foresee any more HasherOpts in the future?

maisem: Nope, I'm fine with this too. Just giving an alternative

bradfitz: Done.

bradfitz:
@dsnet, preference? I can revert back to the earlier commit (https://github.com/tailscale/tailscale/commit/20b7237379226b0896bbf9747df88c6391b53400) if you prefer a separate top-level constructor and no options.

dsnet:
I like the variadic arguments a lot more. It's flexible for future filter options like `FilterType`.

Minor comments:
* Maybe just name it `Option`? Isn't `HasherOpt` redundant with being in the `deephash` package?
* Having `FilterFields` operate on `reflect.StructField` is more flexible, but it's kinda annoying to work with. What are your thoughts about the following:
```go
// FieldFields filters out fields of T of the provided names.
// T must be a struct type. 
func FilterFields[T any](names ...string) Option
```
Thus usage would be:
```go
deephash.FilterFields[tar.Header]("Uid", "Gid")
```

Overall, the API LGTM, so consider these comments to be just suggestions.

dsnet (approved):

bradfitz:
> Thus usage would be

Now you have two type parameters, one to HasherForType and one for the option? 😬

bradfitz:
That is, callers would look like:

```go
    var hashFoo = deephash.HasherForType[tailcfg.Foo](deephash.FilterFields[tailcfg.Foo]("ID", "StableID"))
```

Pretty verbose.

Also, that makes me declare up front the fields I _want_ to keep. I want to also be able to specify "everything but X", to let it just work as fields get added over time, since I'm effectively specifying the negative set.

dsnet:
(In the future) I would not be surprised if we need to support ignoring fields that are not on the top-level type. The moment we accept that reality, we must support specifying multiple types.

It doesn't seem all that unreadable on multiple lines:
```go
var hashFoo = deephash.HasherForType[tailcfg.Foo](
    deephash.FilterFields[tailcfg.Foo]("ID", "StableID"),
)
```

bradfitz:
Gotcha. I see your goal now. Okay, that works for me. I don't feel like implementing that yet (probably means reworking the type hasher cache) but I'll document the one filter restriction for now.

bradfitz:
I still don't like the ...string allowlist though, as I want a blocklist of fields and magic like minus prefixes seems kinda gross?

dsnet:
> I still don't like the ...string allowlist though, as I want a blocklist of fields and magic like minus prefixes seems kinda gross?

Hahaha, that's the same problem that the `cmp` package runs into. We provide the ability to ignore certain fields of a struct, but not an easy way to say "ignore everything except x, y, z".

It's not hard to add an option that's `FieldFieldsExcept`, but then you run into gnarly questions like:
* what happens if you specify `FilterFields[T]("foo")` and `FilterFieldsExcept[T]("foo")` at the same time? Do you panic?

Speaking of which, I'm not fond of the name "filter" since I can never remember whether it is a filter to include or filter to exclude? Perhaps just use the words "Include" and "Exclude" in the option name?

maisem (approved):

util/deephash/deephash.go
```diff
@@ -152,12 +153,51 @@ func Hash[T any](v *T) Sum {
...
+func (hashOptStructFieldFilter) isOption() {}
+
+// FilterFields returns a HasherOpt that only hashes the struct fields for which
+// filter reports true.
+//
+// Use of this option is only valid for struct types.
+//
+// This API only permits selecting the top layer of filtering. You can't choose
+// filter paths of fields down through structs of structs.
+func FilterFields(filter func(reflect.StructField) bool) Option {
```
maisem, line 172: Could name it MatchFields instead of FilterFields

bradfitz:
@dsnet, PTAL.

dsnet (approved):

PR #9105, opened 2023-08-27
types/views: add SliceEqual, like std slices.Equal

bradfitz:
Updates tailscale/corp#6198

### DIFF
--- a/types/views/views.go
+++ b/types/views/views.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"errors"
 	"maps"
+	"slices"
 
 	"go4.org/mem"
 )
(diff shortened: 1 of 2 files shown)


maisem (approved):

types/views/views.go
```diff
@@ -275,6 +275,19 @@ func SliceContains[T comparable](v Slice[T], e T) bool {
 	return false
 }
 
+// SliceEqual is like the standard library's slice.Equal, but for two views.
```
crawshaw, line 278: slices?

crawshaw (approved):

types/views/views.go
bradfitz: whoops. I had it right in the commit message.

bradfitz:
also changed to just `return slices.Equal(a.ж, b.ж)` which gives better bounds checks:

```
             │    before    │                after                │
             │    sec/op    │   sec/op     vs base                │
SliceEqual-8   1243.5n ± 1%   992.9n ± 4%  -20.15% (p=0.000 n=10)
```

PR #9109, opened 2023-08-28
client/web: fix lock/unlock order when updating s.self

soniaappasamy:
Updates tailscale/corp#13775


soniaappasamy:
whoops! had the order wrong locally, not here! closing

PR #9114, opened 2023-08-28
ipn/ipnlocal: support most Linuxes in handleC2NUpdate

noncombatant:
Updates #6995


ipn/ipnlocal/c2n.go
```diff
@@ -115,17 +115,21 @@ func (b *LocalBackend) handleC2NUpdate(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
 
+	var res tailcfg.C2NUpdateResponse
+	res.Enabled = envknob.AllowsRemoteUpdate()
+
+	// We support `tailscale update` on most Linux distributions that we create
+	// packages for. Arch is a special case (see
+	// https://github.com/tailscale/tailscale/issues/6995).
+	res.Supported = runtime.GOOS == "windows" || (runtime.GOOS == "linux" && distro.Get() != distro.Arch)
```
awly, line 131: There's also Synology, macOS, FreeBSD.
I think we can invert this to only list unsupported platforms (Arch, iOS, Android?)
Or better - make a helper function in `clientupdate` (maybe identical to `func Update` but without executing the actual update)

noncombatant: macOS macsys may require some fanciness to get working. I intend to update this for macOS when that's ready.

I'll do FreeBSD and Synology in this PR; coming up.

awly: We also have macOS app store version support here: https://github.com/tailscale/tailscale/blob/main/clientupdate/clientupdate.go#L556

ipn/ipnlocal/c2n.go
```diff
@@ -115,17 +115,26 @@ func (b *LocalBackend) handleC2NUpdate(w http.ResponseWriter, r *http.Request) {
...
-	case "GET", "POST":
-	default:
+	// We interpret GET as a request for current status, and POST as a request to
+	// actually begin an update.
+	if r.Method != "GET" && r.Method != "POST" {
 		http.Error(w, "bad method", http.StatusMethodNotAllowed)
 		return
 	}
 
+	var res tailcfg.C2NUpdateResponse
```
awly, line 125: optional: define this later, as a struct literal
```go
res := tailcfg.C2NUpdateResponse{
  Enabled: ...,
  Supported: ...,
}
```

clientupdate/clientupdate.go
```diff
@@ -89,94 +83,114 @@ type UpdateArgs struct {
...
+			return up.updateMacSys
 		default:
-			up.update = up.updateMacAppStore
+			return up.updateMacAppStore
 		}
 	case "freebsd":
-		up.update = up.updateFreeBSD
+		return up.updateFreeBSD
+	}
+	return nil
```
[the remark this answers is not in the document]
noncombatant: Picking (1), because then we can do `_, err := NewUpdater...` and the code will be more obvious. The current code requires that explanatory comment which is a bad smell. By contrast, the `_, err` convention is attested in the wild. :)

clientupdate/clientupdate.go
```diff
@@ -89,94 +83,114 @@ type UpdateArgs struct {
...
+// On Windows, this copies the calling binary and re-executes it to apply the
+// update. The calling binary should handle an "update" subcommand and call
+// this function again for the re-executed binary to proceed.
+func Update(args Arguments) error {
+	if err := args.validate(); err != nil {
+		return err
+	}
+	if args.PkgsAddr == "" {
+		args.PkgsAddr = "https://pkgs.tailscale.com"
 	}
```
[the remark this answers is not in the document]
noncombatant: Got closer to that; see what you think.

clientupdate/clientupdate.go
```diff
@@ -127,56 +122,78 @@ func Update(args UpdateArgs) error {
...
+// On Windows, this copies the calling binary and re-executes it to apply the
+// update. The calling binary should handle an "update" subcommand and call
+// this function again for the re-executed binary to proceed.
+func Update(args Arguments) error {
+	if err := args.validate(); err != nil {
+		return err
+	}
+	if args.PkgsAddr == "" {
+		args.PkgsAddr = "https://pkgs.tailscale.com"
+	}
```
awly, line 188: any reason not to move these to `NewUpdater` too?

noncombatant: `validate` requires `args.Confirm` and `args.Logf`, which are extra work to construct when you just need to see if you can get an `Updater` that *would* work, and which are only necessary here inside `Update`.

We could move the code to check and set `PkgsAddr` to `NewUpdater`, though.

noncombatant: (Done; see what you think)

ipn/ipnlocal/c2n.go
```diff
@@ -115,17 +115,24 @@ func (b *LocalBackend) handleC2NUpdate(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
 
+	// If NewUpdater does not return an error, we can update the installation.
+	// Exception: When version.IsMacSysExt returns true, we don't support that
+	// yet. TODO(cpalmer, #6995): Implement it.
+	//
+	// Note that we create the Updater solely to check for errors; we do not
+	// invoke it here. For this purpose, it is ok to pass it a zero Arguments.
+	_, err := clientupdate.NewUpdater(&clientupdate.Arguments{})
```
awly, line 130:
```suggestion
	_, err := clientupdate.NewUpdater(clientupdate.Arguments{})
```

awly (approved):
I would still prefer to do validation somewhere within the `Updater` to make it safer to use.
With the current callers we have, it's ok to keep it in `Update` only, we can revisit this later if needed.

ipn/ipnlocal/c2n.go
```diff
@@ -115,17 +115,24 @@ func (b *LocalBackend) handleC2NUpdate(w http.ResponseWriter, r *http.Request) {
...
 
+	// If NewUpdater does not return an error, we can update the installation.
+	// Exception: When version.IsMacSysExt returns true, we don't support that
+	// yet. TODO(cpalmer, #6995): Implement it.
+	//
+	// Note that we create the Updater solely to check for errors; we do not
+	// invoke it here. For this purpose, it is ok to pass it a zero Arguments.
+	_, err := clientupdate.NewUpdater(clientupdate.Arguments{})
+	res := tailcfg.C2NUpdateResponse {
+		Enabled: envknob.AllowsRemoteUpdate(),
```
twitchyliquid64, line 132: Confirming we can't set envknobs remotely? I remember something like that, but my memory could be faulty

awly: There's https://github.com/tailscale/tailscale/blob/main/control/controlknobs/controlknobs.go but it's not automatically wired to envknobs (which is good).
We're adding a `tailscale set` flag in https://github.com/tailscale/tailscale/pull/8861 for user opt-in.

PR #9117, opened 2023-08-28
cmd/k8s-operator: add matching family addresses to status

maisem:
This was added in 3451b89e5f01fd6c4845bcbf06f1bece9f39cdfc, but resulted in the v6 Tailscale address being added to status when when the forwarding only happened on the v4 address.

Updates #502


maisem:
cc @michaelbeaumont

cmd/k8s-operator/svc.go
```diff
@@ -158,7 +164,13 @@ func (a *ServiceReconciler) maybeProvision(ctx context.Context, logger *zap.Suga
 		{Hostname: tsHost},
 	}
 	for _, ip := range tsIPs {
-		ingress = append(ingress, corev1.LoadBalancerIngress{IP: ip})
+		addr, err := netip.ParseAddr(ip)
+		if err != nil {
+			continue
+		}
+		if (addr.Is4() && clusterIPAddr.Is4()) || (addr.Is6() && clusterIPAddr.Is6()) {
```
danderson, line 171: If both are valid you should be able to `addr.Is4() == other.Is4()` here as a simplification

danderson (approved):

cmd/k8s-operator/svc.go
maisem: fair, done

PR #9118, opened 2023-08-28
clientupdate: use SPKsVersion instead of Version

awly:
Top-level Version in pkgs response is not always in sync with SPK versions, especially on unstable track. It's very confusing when the confirmation prompt asks you "update to 1.49.x?" and you end up updating to 1.49.y.
Instead, grab the SPK-specific version field.

Updates #cleanup.

### DIFF
--- a/clientupdate/clientupdate.go
+++ b/clientupdate/clientupdate.go
@@ -202,15 +202,12 @@ func (up *updater) updateSynology() error {
 	if err != nil {
 		return err
 	}
-	if latest.Version == "" {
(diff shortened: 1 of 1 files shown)


danderson (approved):
Could we fix the json format server-side instead? I'm worried that we keep hardcoding assumptions about release filenames all over the place, every time we do this it becomes harder to change anything in future :/

awly:
Hmm... we could make the version field per-package-type to remove any ambiguity and the often-confusing top-level field., I'll give that a try!

danderson (approved):
<3 thanks!

PR #9134, opened 2023-08-29
cmd/tailscale: combine funnel and serve under dev flag

marwan-at-work:
This PR combines the funnel and serve code under the same path. However, it is using the new code which means features being added to the funnel command will automatically be added to serve but also things that are missing are missing from both.

Updates #8489


tylersmalley:
Opened an alternative implementation at https://github.com/tailscale/tailscale/pull/9169 where the CLI definition is also combined.

PR #9135, opened 2023-08-29
cli/serve: shorten help text on error

### LINKED ISSUE #14274: FR: create an android build for AAOS (Android Auto Operating System) plus Google Automotive Services (GAS)

I'd like to install Tailscale onto my soon to be owned Polestar 3 running AAOS.  I currently have a loaner Polestar 3 running AAOS, but can't install Tailscale.

Note this isn't a request for support on Android Auto - where the car mirrors what appears on a phone screen, but for a car running Android on the onboard device.

### How should we solve this?

Build the android client for AAOS - see https://developers.google.com/cars/design/create-apps and https://built-in.google/cars/ for cars (scroll to the bottom) for cars with this OS.  This will enable these vehicles to join a tailscale network and connect to home networks (and thus to apps like home assistant) over tailnets rather than public internet).

Expsosing apps like Home Assistant to the public internet instead of remaining on home networks and tailscale tailnets increases the attack surface unnecessarily.  Inability to stream media from home NAS storage over tailnets impacts ability to access home media (again without exposing these devices to public internet).

### Anything else?

I can attempt to fork tailscale-android to see how difficult it is to support this app profile, but I'm not a go developer or an android developer, but your minimum android supported version (8) is well below the AAOS version (12), so hopefully this isn't impossible, but I get how things like VPNs might well be completely out of reach on AAOS.

soniaappasamy:
Our BETA serve help text is long and often hides the actual error in the user's usage. Instead of printing the full text, prompt users to use `serve --help` if they want the help info.

Fixes #14274

```
$ tailscale serve https / 127.0.0.1:
error: path must be absolute

try tailscale serve --help for usage info
```


cmd/tailscale/cli/serve.go
```diff
@@ -861,3 +861,10 @@ func (e *serveEnv) enableFeatureInteractive(ctx context.Context, feature string,
 		}
 	}
 }
+
+// printHelp prints standard error text that prompts users
+// to use `serve --help` for information on how to use serve.
+func printHelp() error {
+	fmt.Fprintln(os.Stderr, "try tailscale serve --help for usage info")
+	return nil // exit cleanly
```
bradfitz, line 869: We don't want a 0 exit status from cmd/tailscale on an error.

marwan-at-work (approved):
👌🏼

PR #9144, opened 2023-08-29
clientupdate: add linux tarball updates

awly:
As a fallback to package managers, allow updating tailscale that was self-installed in some way. There are some tricky bits around updating the systemd unit (should we stick to local binary paths or to the ones in tailscaled.service?), so leaving that out for now.

Updates #6995


clientupdate/clientupdate.go
```diff
@@ -822,7 +825,179 @@ func (up *updater) updateFreeBSD() (err error) {
...
+		case "tailscale":
+			files["tailscale"]++
+			if err := writeFile(tr, tailscale+".new", 0755); err != nil {
+				return fmt.Errorf("failed extracting the new tailscale binary from %q: %w", path, err)
+			}
+		case "tailscaled":
+			files["tailscaled"]++
+			if err := writeFile(tr, tailscaled+".new", 0755); err != nil {
+				return fmt.Errorf("failed extracting the new tailscaled binary from %q: %w", path, err)
+			}
```
noncombatant, line 924: should we have a `default:` to check for weird junk? If we did that, we'd need a case to account for tailscaled.service, even if just to say "Yes, we got it."

clientupdate/clientupdate.go
```diff
@@ -822,7 +825,179 @@ func (up *updater) updateFreeBSD() (err error) {
...
+	if err := os.Rename(tailscaled+".new", tailscaled); err != nil {
+		return err
+	}
+	up.Logf("Updated %s", tailscaled)
+	return nil
+}
+
+func writeFile(r io.Reader, path string, perm os.FileMode) error {
+	f, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, perm)
+	if err != nil {
```
noncombatant, line 945: what if `errors.Is(err, fs.ErrExist)`? Do we still want to fail in that case?

clientupdate/clientupdate.go
```diff
@@ -822,7 +825,179 @@ func (up *updater) updateFreeBSD() (err error) {
...
+	up.Logf("Updated %s", tailscale)
+	if err := os.Rename(tailscaled+".new", tailscaled); err != nil {
+		return err
+	}
+	up.Logf("Updated %s", tailscaled)
+	return nil
+}
+
+func writeFile(r io.Reader, path string, perm os.FileMode) error {
+	f, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, perm)
```
noncombatant, line 944: Could use `O_WRONLY` instead of `O_RDWR`. Also `syscall.O_NOFOLLOW` might be a good idea for belt-and-suspenders paranoia? (Maybe not, but maybe.)

clientupdate/clientupdate.go
```diff
@@ -822,7 +825,179 @@ func (up *updater) updateFreeBSD() (err error) {
...
+	defer f.Close()
+	if _, err := io.Copy(f, r); err != nil {
+		return err
+	}
+	return f.Close()
+}
+
+// Var allows overriding this in tests.
+var binaryPaths = func() (tailscale, tailscaled string, err error) {
+	// this can be either tailscale or tailscaled.
```
noncombatant, line 957: Capital T

clientupdate/clientupdate.go
```diff
@@ -822,7 +825,179 @@ func (up *updater) updateFreeBSD() (err error) {
...
+		return "", "", fmt.Errorf("cannot find %q in neither %q nor $PATH: %w", otherName, filepath.Dir(this), err)
+	}
+	if otherName == "tailscaled" {
+		return this, other, nil
+	} else {
+		return other, this, nil
+	}
+}
+
+func restartSystemdUnit() error {
```
noncombatant, line 983: Is this useful, or more trouble than it's worth? : https://github.com/coreos/go-systemd

clientupdate/clientupdate.go
awly: That would cause updates to fail if we add a new file to the tarball in a new release.
Without `default` we have some forward-compatibility :)

clientupdate/clientupdate.go
awly: Added `O_WRONLY`.
`O_NOFOLLOW` is not portable (i.e. doesn't compile on Windows), so I'd rather leave it out.

clientupdate/clientupdate.go
awly: `os.OpenFile` with the above flags will not return `ErrExist` if the file already exists.
Added a unit test to ensure that

clientupdate/clientupdate.go
awly: It's not a new dependency, so might be worth it: https://pkg.go.dev/github.com/coreos/go-systemd/v22/dbus
I'll give it a shot!

clientupdate/clientupdate.go
noncombatant: I can't quite shake the fear of a symlink attack when writing as root. `O_NOFOLLOW` is present on macOS and FreeBSD as well as Linux, leaving only Windows, suggesting that a build tag approach wouldn't get too out of hand.

According to MSDN, https://learn.microsoft.com/en-us/windows/win32/fileio/symbolic-link-effects-on-file-systems-functions#createfile-and-createfiletransacted, `FILE_FLAG_OPEN_REPARSE_POINT` should work (in combination with `CREATE_ALWAYS` and/or `TRUNCATE_EXISTING`. All are present in the `syscall` package.

Alas, in my testing just now, passing `FILE_FLAG_OPEN_REPARSE_POINT | CREATE_ALWAYS | TRUNCATE_EXISTING` results in silent failure. `os.OpenFile` does not return an `error`. I think we'd have to use the actual `syscall.CreateFile` (which does exist, if we ever need to go that far).

awly: Would it be easier to just `os.Remove` the target first and then write a fresh file?

clientupdate/clientupdate.go
awly: Works like a charm :ok_hand:

clientupdate/clientupdate.go
awly: Updated to do a delete before writing the file. In case there is a random symlink at the destination, it will just get replaced with a regular file instead of failing the update.
PTAL

noncombatant: That would allow a local attacker who can place a symlink to cause root to remove any file, right? (Instead of overwriting any file with a tailscale binary.)

Alternate strategies:

* Tag this function as building only on Linux, and add a comment to the effect that if people want to build and use it on Windows, they should come talk to us about the bug
* Create a new directory and ensure that only root can write to it, thus ensuring that local attackers can't create symlinks in it (we may already be creating a new directory? if so adding the 1 check is easy)
* Document it as a TODO (possibly with a bug link) and come back to it later

awly: > That would allow a local attacker who can place a symlink to cause root to remove any file, right? (Instead of overwriting any file with a tailscale binary.)

I added a test specifically for this ;)
`os.Remove` unlinks a file and does not follow symlinks https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/os/file_unix.go;l=364

PR #9146, opened 2023-08-29
net/wsconn: accept a remote addr string and plumb it through

danderson:
This makes wsconn.Conns somewhat present reasonably when they are the client of an http.Request, rather than just put a placeholder in that field.

Updates tailscale/corp#13777

---

Sadly I can't make this a properly typed net.Addr, since the call sites where wsconns get established only have strings of undefined content as remote addresses. But we can at least pass through that unspecified string to whoever cares.

### DIFF
--- a/net/wsconn/wsconn.go
+++ b/net/wsconn/wsconn.go
@@ -48,10 +48,18 @@ import (
 //
 // A received StatusNormalClosure or StatusGoingAway close frame will be translated to
 // io.EOF when reading.
-func NetConn(ctx context.Context, c *websocket.Conn, msgType websocket.MessageType) net.Conn {
+//
+// The given remoteAddr will be the value of the returned conn's
+// RemoteAddr().String(). For best compatibility with consumers of
+// conns, the string should be an ip:port if available, but in the
+// absence of that it can be any string that describes the remote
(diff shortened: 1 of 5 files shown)


net/wsconn/wsconn.go
```diff
@@ -48,10 +48,11 @@ import (
 //
 // A received StatusNormalClosure or StatusGoingAway close frame will be translated to
 // io.EOF when reading.
-func NetConn(ctx context.Context, c *websocket.Conn, msgType websocket.MessageType) net.Conn {
+func NetConn(ctx context.Context, c *websocket.Conn, msgType websocket.MessageType, remoteAddr string) net.Conn {
```
bradfitz, line 51: Docs on this? Is it an IP:port always or can it be a URL sometimes too it looks like?

bradfitz (approved):

net/wsconn/wsconn.go
danderson: Docced. It can be anything at all, because it's the r.RemoteAddr of the incoming request, which can be anything at all. For client connections, I arbitrarily decided to make it be the Websocket URL being connected to, but it wasn't deeply thought out. Just something that represents the remote endpoint, and seems somewhat sensible.

PR #9148, opened 2023-08-30
cmd/sniproxy: switch to peterbourgon/ff for flags

DentonGentry:
Add support for TS_APPC_* variables to supply arguments by
switching to https://github.com/peterbourgon/ff for CLI
flag parsing. For example:
TS_APPC_FORWARDS=tcp/22/github.com ./sniproxy

### DIFF
--- a/cmd/sniproxy/sniproxy.go
+++ b/cmd/sniproxy/sniproxy.go
@@ -16,10 +16,12 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"os"
 	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


raggi:
You might want to considering taking a dependency on `ff` (https://github.com/peterbourgon/ff), which we already use parts of elsewhere. It has a generalized solution that looks like this:

```
	err := ff.Parse(flag.CommandLine, os.Args[1:],
		ff.WithEnvVarPrefix("TS_APPC"),
	)
```

DentonGentry:
Today I learned that peterbourgon/ff is awesome. Done.

maisem (approved):

Issue #9141, opened 2023-08-29
tailscale-operator ingress can't hit intra-cluster service


jtschelling:
### What is the issue?

i'm running the tailscale-operator on a kubernetes cluster, and attempting to use the new ingressclass introduced recently. here's my example manifest i'm using to test:

```
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: nginx-ingress
  namespace: tailscale-test
spec:
  ingressClassName: tailscale
  rules:
    - http:
        paths:
          - path: /
            pathType: Exact
            backend:
              service:
                name: nginx-service
                port:
                  name: web-service
---
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
  namespace: tailscale-test
spec:
  type: ClusterIP
  selector:
    app.kubernetes.io/name: nginx
  ports:
    - name: web-service
      protocol: TCP
      port: 80
      targetPort: web
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  namespace: tailscale-test
spec:
  replicas: 2
  selector:
    matchLabels:
      app.kubernetes.io/name: nginx
  template:
    metadata:
      labels:
        app.kubernetes.io/name: nginx
    spec:
      containers:
      - name: web
        image: nginx
        ports:
        - containerPort: 80
          name: web
```

the ingress comes up fine and i get `tailscale-test-nginx-ingress-ingress` machine added to my tailnet in the admin console. I can run `tailscale ping $INGRESSTAILNETIP` fine, here's the output

```
pong from tailscale-test-nginx-ingress-ingress ($INGRESSTAILNETIP) via DERP(ord) in 12ms
pong from tailscale-test-nginx-ingress-ingress ($INGRESSTAILNETIP) via DERP(ord) in 16ms
pong from tailscale-test-nginx-ingress-ingress ($INGRESSTAILNETIP) via DERP(ord) in 27ms
pong from tailscale-test-nginx-ingress-ingress ($INGRESSTAILNETIP) via $LANIP:58208 in 9ms
```

when i try to `curl http://$INGRESSTAILNETIP` i get:

```
curl: (7) Failed to connect to $INGRESSTAILNETIP port 80 after 37 ms: Couldn't connect to server
```

there's a pod `ts-nginx-ingress-58pz9-0` that gets created in the `tailscale` namespace that the operator is running in. when i look at the logs in that pod after running the curl command i see:

```
[...]

PR #9165, opened 2023-08-30
cmd/k8s-operator: emit event if HTTPS is disabled on Tailnet

maisem:
Instead of confusing users, emit an event that explicitly tells the user that HTTPS is disabled on the tailnet and that ingress may not work until they enable it.

Updates #9141

### DIFF
--- a/cmd/k8s-operator/ingress.go
+++ b/cmd/k8s-operator/ingress.go
@@ -125,6 +125,10 @@ func (a *IngressReconciler) maybeProvision(ctx context.Context, logger *zap.Suga
 	gaugeIngressResources.Set(int64(a.managedIngresses.Len()))
 	a.mu.Unlock()
 
+	if !a.ssr.IsHTTPSEnabledOnTailnet() {
+		a.recorder.Event(ing, corev1.EventTypeWarning, "HTTPSNotEnabled", "HTTPS is not enabled on the tailnet; ingress may not work")
+	}
+
 	// magic443 is a fake hostname that we can use to tell containerboot to swap
 	// out with the real hostname once it's known.
 	const magic443 = "${TS_CERT_DOMAIN}:443"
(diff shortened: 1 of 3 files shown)


cmd/k8s-operator/ingress.go
```diff
@@ -125,6 +125,10 @@ func (a *IngressReconciler) maybeProvision(ctx context.Context, logger *zap.Suga
 	gaugeIngressResources.Set(int64(a.managedIngresses.Len()))
 	a.mu.Unlock()
 
+	if !a.ssr.IsHTTPSEnabledOnTailnet(ctx) {
+		a.recorder.Event(ing, corev1.EventTypeWarning, "HTTPSNotEnabled", "HTTPS is not enabled on the tailnet, ingress may not work")
```
bradfitz, line 129: semicolon?

cmd/k8s-operator/sts.go
```diff
@@ -71,13 +72,19 @@ type tailscaleSTSConfig struct {
...
+	tsnetServer            *tsnet.Server
 	tsClient               tsClient
 	defaultTags            []string
 	operatorNamespace      string
 	proxyImage             string
 	proxyPriorityClassName string
 }
 
+// IsHTTPSEnabledOnTailnet reports whether HTTPS is enabled on the tailnet.
+func (a *tailscaleSTSReconciler) IsHTTPSEnabledOnTailnet(ctx context.Context) bool {
```
bradfitz, line 84: what's the context for?

bradfitz (approved):

cmd/k8s-operator/sts.go
maisem: oh i was using LocalAPI previously, then realized i could just get `CertDomains` will remove

PR #9167, opened 2023-08-30
client/web: simply csrf key caching in cgi mode

willnorris:
Instead of trying to use the user config dir, and then fail back to the OS temp dir, just always use the temp dir. Also use a filename that is less likely to cause collisions.

This addresses an issue on a test synology instance that was mysteriously failing because there was a file at /tmp/tailscale. We could still technically run into this issue if a
/tmp/tailscale-web-csrf.key file exists, but that seems far less likely.

Updates tailscale/corp#13775

### DIFF
--- a/client/web/web.go
+++ b/client/web/web.go
@@ -456,16 +456,10 @@ var localapiAllowlist = []string{
 // If the server is running in CGI mode, the key is cached to disk and reused between requests.
 // If an error occurs during key storage, the error is logged and the active process terminated.
 func (s *Server) csrfKey() []byte {
-	var csrfFile string
+	csrfFile := filepath.Join(os.TempDir(), "tailscale-web-csrf.key")
 
 	// if running in CGI mode, try to read from disk, but ignore errors
 	if s.cgiMode {
-		confdir, err := os.UserConfigDir()
-		if err != nil {
-			confdir = os.TempDir()
-		}
-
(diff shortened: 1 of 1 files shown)


client/web/web.go
```diff
@@ -456,16 +456,10 @@ var localapiAllowlist = []string{
 // If the server is running in CGI mode, the key is cached to disk and reused between requests.
 // If an error occurs during key storage, the error is logged and the active process terminated.
 func (s *Server) csrfKey() []byte {
-	var csrfFile string
+	csrfFile := filepath.Join(os.TempDir(), "tailscale-web-csrf.key")
```
clairew, line 459: can you easily check if if a /tmp/tailscale-web-csrf.key exists already before creating it?

creachadair: Alternatively I suppose we could use `os.MkdirTemp` to ensure we don't have a collision (though if the idea is to reuse the value over restarts, perhaps not).

willnorris: yeah, exactly... in a CGI environment, every request is a run in a separate process, so we have to have a stable file between runs.

willnorris: outside of CGI, when we actually start up a server that servers requests, we don't cache the csrf key at all... we just hold it in memory.  This is only a workaround for CGI (which proves to be super common with NAS devices... synology, qnap, and unraid all use CGI for plugins)

PR #9174, opened 2023-08-30
tailcfg,ipn/ipnlocal: add C2NUpdateRequest to c2n /update handler

awly:
An optional (for now) request body for POST /update messages to carry the suggested version. This removes the need from clients to figure out what the latest available version is.

Updates #6907


tailcfg/c2ntypes.go
```diff
@@ -34,6 +34,15 @@ type C2NSSHUsernamesResponse struct {
 	Usernames []string
 }
 
+// C2NUpdateRequest is the request (from control to node) to the /update
+// handler. It tells the node what the latest version is. The node may choose
+// to ignore this update, notify the user, or apply it automatically (for POST
+// requests only).
+type C2NUpdateRequest struct {
+	// LatestVer is the latest released version.
+	LatestVer string
```
bradfitz, line 43: "Latest" or "LatestVersion". Abbreviating "Ver" here is a bit weird. (in tailcfg.go we already have `IPNVersion`, `OSVersion`, `DistroVersion`, `GoVersion`, `Version`, etc)

Also: what format is this in? Give an example quoted string. The bare "1.50.2" or the full version with all the hash suffixes?

tailcfg/c2ntypes.go
```diff
@@ -34,6 +34,15 @@ type C2NSSHUsernamesResponse struct {
 	Usernames []string
 }
 
+// C2NUpdateRequest is the request (from control to node) to the /update
+// handler. It tells the node what the latest version is. The node may choose
```
bradfitz, line 38: // C2NUpdateRequest is the request body for POST requests to the c2n /update handler.
// It tells ...

?

tailcfg/c2ntypes.go
```diff
@@ -34,6 +34,15 @@ type C2NSSHUsernamesResponse struct {
...
+// to ignore this update, notify the user, or apply it automatically (for POST
```
bradfitz, line 39: Based on what? Configuration?

Should the node double check that 1.47.9384932849238493 isn't some bespoke version meant just for that node and not listed in the usual indexes? I'd call out how trustworthy this is.

Can this be used to downgrade? In the future? Or will that be a separate bool/field?

ipn/ipnlocal/c2n.go
```diff
@@ -121,6 +122,16 @@ func (b *LocalBackend) handleC2NUpdate(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	var req tailcfg.C2NUpdateRequest
+	if r.ContentLength > 0 {
```
bradfitz, line 126: what about -1 for chunked? Maybe != 0 instead?

tailcfg/c2ntypes.go
awly: Updated, it should be the bare `x.y.z`, which is what we serve from pkgs.

tailcfg/c2ntypes.go
awly: Based on configuration, yes (see https://github.com/tailscale/tailscale/pull/8861).
Good point on validation. I've added some more of it in `handleC2NUpdate` and documented here that the value should not be blindly trusted.

> Should the node double check that 1.47.9384932849238493 isn't some bespoke version

Hmm, yes. We could check it against the `pkgs` JSON response or just avoid passing the version to `tailscale update` (let it figure out the latest version). I'm not sure if there's a better validation strategy, do you have any suggestions?

> Can this be used to downgrade?

I'd really really like to not support downgrades.
But if there's a good reason for it, we can add a bool to this request object and corresponding client validation changes.

tailcfg/c2ntypes.go
awly: We want to support both notification-only (e.g. show an upgrade icon in the tray) and automatic update modes.

I'm torn between two API options:
1. send request in both GET (as an indication of latest available version) and POST (as a command to apply the update); this informs the client about available updates, for GUI clients to display it but not apply
2. send request only in POST; if the client has not opted in to automatic updates, it can return some new response flag `Notified` instead of `Started`

This PR does # 1, always getting the version in GET requests. But I'm open to # 2 if that's a better API design.

bradfitz: So c2n is the imperative API ("do $X now").

The declarative API is MapResponse, which already has `MapResponse.ClientVersion` to just advertise an "FYI, there's a 1.x.y available now and FYI it's a security update if you care".

awly:
Closing because `MapResponse.ClientVersion` already exists

PR #9181, opened 2023-08-31
ipn/ipnlocal: deflake some tests

bradfitz:
* don't try to re-Start (and thus create a new client) during Shutdown
* in tests, wait for controlclient to fully shut down when replacing it
* log a bit more

Updates tailscale/corp#14139
Updates tailscale/corp#13175 etc
Updates #9178 and its flakes.

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -565,7 +565,14 @@ func (b *LocalBackend) Shutdown() {
 		b.mu.Unlock()
 		ctx, cancel := context.WithTimeout(b.ctx, 5*time.Second)
 		defer cancel()
-		b.LogoutSync(ctx) // best effort
+		t0 := time.Now()
+		err := b.LogoutSync(ctx) // best effort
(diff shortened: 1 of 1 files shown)


ipn/ipnlocal/local.go
```diff
@@ -3872,7 +3884,18 @@ func (b *LocalBackend) resetControlClientLockedAsync() {
...
-	go b.cc.Shutdown()
+	if testenv.InTest() {
+		// From 2021-04-20 to 2023-08-31 we always did this shutdown
+		// asynchronously. But tests flaked on it sometimes, as our
+		// tests often do resource leak checks at the end to make sure
+		// everything's shut down. So do this synchronously in tests
+		// to deflake tests.
+		//
+		// TODO(bradfitz,maisem): do this always?
+		b.cc.Shutdown()
```
maisem, line 3895: i think we should just do synchronously always and deal with the fallout.

maisem: I think previously we were doing this async in fears that it might block, but i don't think it can.

PR #9183, opened 2023-08-31
ipn, ipn/ipnlocal: add session identifier for WatchIPNBus

marwan-at-work:
This PR adds a SessionID field to the ipn.Notify struct so that ipn buses can identify a session and register deferred clean up code in the future. The first use case this is for is to be able to tie foreground serve configs to a specific watch session and ensure its clean up when a connection is closed.

Updates #8489

### DIFF
--- a/ipn/backend.go
+++ b/ipn/backend.go
@@ -74,8 +74,9 @@ const (
 // that they have not changed.
 // They are JSON-encoded on the wire, despite the lack of struct tags.
 type Notify struct {
-	_       structs.Incomparable
-	Version string // version number of IPN backend
+	_         structs.Incomparable
+	Version   string // version number of IPN backend
+	SessionID string // identifies the unique WatchIPNBus session.
 
 	// ErrMessage, if non-nil, contains a critical error message.
 	// For State InUseOtherUser, ErrMessage is not critical and just contains the details.
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -5,7 +5,9 @@ package ipnlocal
 
 import (
 	"context"
+	"crypto/rand"
 	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -1939,6 +1941,10 @@ func (b *LocalBackend) ResendHostinfoIfNeeded() {
 func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWatchOpt, onWatchAdded func(), fn func(roNotify *ipn.Notify) (keepGoing bool)) {
 	ch := make(chan *ipn.Notify, 128)
 
+	randomBytes := make([]byte, 16)
+	rand.Read(randomBytes)
+	sessionID := hex.EncodeToString(randomBytes)
+
 	origFn := fn
 	if mask&ipn.NotifyNoPrivateKeys != 0 {
 		fn = func(n *ipn.Notify) bool {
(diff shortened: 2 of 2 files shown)


ipn/backend.go
```diff
@@ -74,8 +74,9 @@ const (
 // that they have not changed.
 // They are JSON-encoded on the wire, despite the lack of struct tags.
 type Notify struct {
-	_       structs.Incomparable
-	Version string // version number of IPN backend
+	_         structs.Incomparable
+	Version   string // version number of IPN backend
+	SessionID string // identifies the unique WatchIPNBus session.
```
bradfitz, line 79: you'll want to omitempty this

but now I'm confused why the others don't have it. I think it's a Swift problem (cc @agottardo) as that also has to decode these types.

ipn/ipnlocal/local.go
```diff
@@ -1962,7 +1968,7 @@ func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWa
 	b.mu.Lock()
 	const initialBits = ipn.NotifyInitialState | ipn.NotifyInitialPrefs | ipn.NotifyInitialNetMap
 	if mask&initialBits != 0 {
-		ini = &ipn.Notify{Version: version.Long()}
+		ini = &ipn.Notify{Version: version.Long(), SessionID: sessionID}
```
bradfitz, line 1971: I think this'll break swift unless you make the session ID opt-in with a new bitmask flag.

also, you'll want to register this somewhere on LocalBackend and call LocalBackend when this Watch call dies.

agottardo: New fields as I see in this PR are fine, Swift will simply ignore them. The problem is when a field is non-optional in Swift, and the server stops sending it: that will cause a failure to decode.

ipn/ipnlocal/local.go
```diff
@@ -1939,6 +1941,10 @@ func (b *LocalBackend) ResendHostinfoIfNeeded() {
 func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWatchOpt, onWatchAdded func(), fn func(roNotify *ipn.Notify) (keepGoing bool)) {
 	ch := make(chan *ipn.Notify, 128)
 
+	randomBytes := make([]byte, 16)
+	rand.Read(randomBytes)
+	sessionID := hex.EncodeToString(randomBytes)
```
bradfitz, line 1946: sent https://github.com/tailscale/tailscale/pull/9237 which this could use

ipn/ipnlocal/local.go
```diff
@@ -1960,9 +1966,10 @@ func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWa
 	var ini *ipn.Notify
 
 	b.mu.Lock()
+
 	const initialBits = ipn.NotifyInitialState | ipn.NotifyInitialPrefs | ipn.NotifyInitialNetMap
 	if mask&initialBits != 0 {
-		ini = &ipn.Notify{Version: version.Long()}
+		ini = &ipn.Notify{Version: version.Long(), SessionID: sessionID}
```
bradfitz, line 1972: let's break this up to one field per line now (so four lines total out of this line)

bradfitz: maybe we need a new NotifyWatchOpt bit to request a session ID and send it at start so they don't need InitialState? But I guess InitialState is cheap enough. Just document on the NotifyWatchOpt that setting InitialState also gets you a SessionID?

ipn/ipnlocal/local.go
```diff
@@ -2018,6 +2025,7 @@ func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWa
 		case <-ctx.Done():
 			return
 		case n := <-ch:
+			n.SessionID = sessionID
```
bradfitz, line 2028: I think only the first one is fine. We can remove this line.

ipn/backend.go
marwan-at-work: Resolved per slack chat and @agottardo 's comment below

ipn/ipnlocal/local.go
marwan-at-work: Approved, will use that if it goes in first. Otherwise, I'll send a follow up. Thanks!

bradfitz (approved):

PR #9184, opened 2023-08-31
clientupdate/distsign: add ability to validate a binary that is alrea…

dblohm7:
…dy located on disk

Our build system caches files locally and only updates them when something changes. Since I need to integrate some distsign stuff into the build system to validate our Windows 7 MSIs, I want to be able to check the cached copy before downloading another copy from pkgs.

Updates https://github.com/tailscale/corp/issues/14334

### DIFF
--- a/clientupdate/distsign/distsign.go
+++ b/clientupdate/distsign/distsign.go
@@ -247,6 +247,48 @@ func (c *Client) Download(ctx context.Context, srcPath, dstPath string) error {
 	return nil
 }
 
+// ValidateLocalBinary fetches the latest signature associated with the binary
+// at srcURLPath and uses it to validate the file located on disk via
+// localFilePath. ValidateLocalBinary returns an error if anything goes wrong
+// with the signature download or with signature validation.
+func (c *Client) ValidateLocalBinary(srcURLPath, localFilePath string) error {
+	// Always fetch a fresh signing key.
+	sigPub, err := c.signingKeys()
+	if err != nil {
+		return err
+	}
+
+	srcURL := c.url(srcURLPath)
+	sigURL := srcURL + ".sig"
+
+	localFile, err := os.Open(localFilePath)
+	if err != nil {
+		return err
+	}
+	defer localFile.Close()
+
+	h := NewPackageHash()
+	_, err = io.Copy(h, localFile)
+	if err != nil {
+		return err
+	}
+	hash, hashLen := h.Sum(nil), h.Len()
+
+	c.logf("Downloading %q", sigURL)
+	sig, err := fetch(sigURL, signatureSizeLimit)
+	if err != nil {
+		return err
+	}
+
+	msg := binary.LittleEndian.AppendUint64(hash, uint64(hashLen))
+	if !VerifyAny(sigPub, msg, sig) {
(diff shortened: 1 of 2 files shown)


clientupdate/distsign/distsign.go
```diff
@@ -247,6 +247,44 @@ func (c *Client) Download(ctx context.Context, srcPath, dstPath string) error {
 	return nil
 }
 
+func (c *Client) ValidateLocalBinary(srcURLPath, localFilePath string) error {
```
awly, line 250: Please add godoc

clientupdate/distsign/distsign.go
```diff
@@ -247,6 +247,44 @@ func (c *Client) Download(ctx context.Context, srcPath, dstPath string) error {
...
+	if err != nil {
+		return err
+	}
+	hash, hashLen := h.Sum(nil), h.Len()
+
+	c.logf("Downloading %q", sigURL)
+	sig, err := fetch(sigURL, signatureSizeLimit)
+	if err != nil {
+		return err
+	}
```
awly, line 277: can the signature file also be local?
it's available on the build server during new builds

clientupdate/distsign/distsign.go
```diff
@@ -247,6 +247,44 @@ func (c *Client) Download(ctx context.Context, srcPath, dstPath string) error {
...
+
+	msg := binary.LittleEndian.AppendUint64(hash, uint64(hashLen))
+	if !VerifyAny(sigPub, msg, sig) {
+		return fmt.Errorf("signature %q for file %q does not validate with the current release signing key; either you are under attack, or attempting to download an old version of Tailscale which was signed with an older signing key", sigURL, srcURL)
```
awly, line 281:
```suggestion
		return fmt.Errorf("signature %q for file %q does not validate with the current release signing key; either you are under attack, or attempting to download an old version of Tailscale which was signed with an older signing key", sigURL, localFilePath)
```

awly (commented):
Please also add some unit tests

clientupdate/distsign/distsign.go
dblohm7: @awly I'd say that the main reason I don't make it local is because we want to know if our local copy of the package is out of date or not; I'm essentially using the signature as an indicator of this, in addition to its typical use case.

clientupdate/distsign/distsign.go
dblohm7: doh! Thanks for catching that!

clientupdate/distsign/distsign_test.go
```diff
@@ -119,6 +119,124 @@ func TestDownload(t *testing.T) {
...
+		t.Run(tt.desc, func(t *testing.T) {
+			srv.reset()
+
+			// First just do a successful Download.
+			want := []byte("world")
+			srv.addSigned("hello", want)
+			dst := filepath.Join(t.TempDir(), tt.src)
+			t.Cleanup(func() {
+				os.Remove(dst)
+			})
```
awly, line 209: `t.TempDir()` should get removed once the test is finished.
```suggestion
```

awly (approved):

PR #9185, opened 2023-08-31
.github/workflows: add checklocks

raggi:
Currently the checklocks step is not configured to fail, as we do not yet have the appropriate annotations.

Updates tailscale/corp#14381

### DIFF
--- a/.github/workflows/checklocks.yml
+++ b/.github/workflows/checklocks.yml
@@ -1,0 +1,28 @@
+name: checklocks
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths:
+      - '**/*.go'
+      - '.github/workflows/checklocks.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-$${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  checklocks:
+    runs-on: [ ubuntu-latest ]
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Build checklocks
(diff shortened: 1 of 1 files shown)


.github/workflows/vet.yml
```diff
@@ -0,0 +1,31 @@
...
+
+  steps:
+    - name: Check out code
+      uses: actions/checkout@v3
+
+    - name: Run go vet
+      run: ./tool/go vet ./...
+
+    - name: Build checklocks
+      run: ./tool/go build -o /tmp/checklocks gvisor.dev/gvisor/tools/checklocks/cmd/checklocks
```
andrew-d, line 27: If you wanted: you can build a custom `vettool` that bundles the `checklocks` Analyzer along with the rest of the standard `go vet` analyzers, and can filter failures to be non-fatal. See `cmd/vet` in the corp repo for more details (`ignoreTests`, etc.).

raggi: I looked at that. Is there a reason to add that cost of maintenance right now?

andrew-d: Nope, feel free to ignore; just wanted to flag if you haven't seen it 👍

raggi: Thanks yeah, I'm just not sure I want to track upstream vet changes to keep the lint set up to date. I can imagine sometime there may be a performance advantage to folding into a single pass, or if there are checks we want to disable. Right now vet passes, so that seems like a good place to maintain for now!

raggi:
This took 10m to run the pipeline, which may be more than we want to add. One part was the gocross build which could probably be avoided, but the checklocks itself took 6m11s.

andrew-d (approved):
This seems reasonable to try out as a non-blocking check 👍

PR #9187, opened 2023-08-31
control/controlclient: rename Auto cancel methods, add missing Lock variant

bradfitz:
Then use the Locked variants in Shutdown while we already hold the lock.

Updates #cleanup

### DIFF
--- a/control/controlclient/auto.go
+++ b/control/controlclient/auto.go
@@ -231,7 +231,7 @@ func NewNoStart(opts Options) (_ *Auto, err error) {
 func (c *Auto) SetPaused(paused bool) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	if paused == c.paused {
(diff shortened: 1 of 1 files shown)


maisem (approved):

control/controlclient/auto.go
```diff
@@ -279,9 +279,17 @@ func (c *Auto) updateControl() {
...
+// & creates a new one, causing it to restart.
+func (c *Auto) cancelAuthCtx() {
 	c.mu.Lock()
 	defer c.mu.Unlock()
+	c.cancelAuthCtxLocked()
+}
+
+// cancelAuthCtxLocked is like cancelAuthCtx, but assumes the caller holds c.mu.
+func (c *Auto) cancelAuthCtxLocked() {
+	c.cancelMapCtxLocked()
```
bradfitz, line 292: whoops :P

PR #9188, opened 2023-08-31
control/controlclient: run SetControlClientStatus in goroutine

maisem:
We have cases where the SetControlClientStatus would result in a Shutdown call back into the auto client that would block forever. The right thing to do here is to fix the LocalBackend state machine but thats a different dumpster fire that we are slowly making progress towards.

This makes it so that the SetControlClientStatus happens in a different goroutine so that calls back into the auto client do not block.

Also add a few missing mu.Unlocks in LocalBackend.Start.

Updates #9181

### DIFF
--- a/control/controlclient/auto.go
+++ b/control/controlclient/auto.go
@@ -159,7 +159,6 @@ type Auto struct {
 	loggedIn       bool       // true if currently logged in
 	loginGoal      *LoginGoal // non-nil if some login activity is desired
 	synced         bool       // true if our netmap is up-to-date
-	inSendStatus   int        // number of sendStatus calls currently in progress
 	state          State
 
 	authCtx    context.Context // context used for auth requests
@@ -638,7 +637,6 @@ func (c *Auto) sendStatus(who string, err error, url string, nm *netmap.NetworkM
 	state := c.state
 	loggedIn := c.loggedIn
 	synced := c.synced
-	c.inSendStatus++
 	c.mu.Unlock()
 
 	c.logf("[v1] sendStatus: %s: %v", who, state)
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/local.go
```diff
@@ -1387,6 +1387,7 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 		newPrefs.Persist = oldPrefs.Persist().AsStruct()
 		pv := newPrefs.View()
 		if err := b.pm.SetPrefs(pv); err != nil {
+			b.mu.Unlock()
```
bradfitz, line 1390: this isn't an exit path?

maisem: huh 🤦🏼 , fixed

control/controlclient/auto.go
```diff
@@ -159,7 +159,6 @@ type Auto struct {
 	loggedIn       bool       // true if currently logged in
 	loginGoal      *LoginGoal // non-nil if some login activity is desired
 	synced         bool       // true if our netmap is up-to-date
-	inSendStatus   int        // number of sendStatus calls currently in progress
```
maisem, line 162: this was only ever used for logs

ipn/ipnlocal/local.go
```diff
@@ -1396,6 +1396,7 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 	wantRunning := prefs.WantRunning()
 	if wantRunning {
 		if err := b.initMachineKeyLocked(); err != nil {
+			b.mu.Unlock()
```
bradfitz, line 1399: also:

```
		// Don't warn about broken Linux IP forwarding when
		// netstack is being used.
		SkipIPForwardingCheck: isNetstack,
	})
	if err != nil {
		return err
	}
```

bradfitz: oh nevermind

maisem: Unlock happens earlier than this. This is fine.

bradfitz (approved):

control/controlclient/auto.go
bradfitz: yeah

PR #9189, opened 2023-08-31
client/web,cmd/tailscale: add prefix flag for web command

willnorris:
We already had a path on the web client server struct, but hadn't plumbed it through to the CLI. Add that now and use it for Synology and QNAP instead of hard-coding the path. (Adding flag for QNAP is https://github.com/tailscale/tailscale-qpkg/pull/112) This will allow supporting other environments (like unraid) without additional changes to the client/web package.

Also fix a small bug in unraid handling to only include the csrf token on POST requests.

Updates tailscale/corp#13775

### DIFF
--- a/client/web/web.go
+++ b/client/web/web.go
@@ -58,7 +58,7 @@ type Server struct {
 	devProxy *httputil.ReverseProxy // only filled when devMode is on
 
 	cgiMode    bool
-	cgiPath    string
+	pathPrefix string
 	apiHandler http.Handler // csrf-protected api handler
 }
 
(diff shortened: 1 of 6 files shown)


WalterHub (approved):

client/web/web.go
```diff
@@ -487,6 +475,19 @@ func (s *Server) csrfKey() []byte {
 // Unlike http.StripPrefix, it does not return a 404 if the prefix is not present.
 // Instead, it returns a redirect to the prefix path.
 func enforcePrefix(prefix string, h http.HandlerFunc) http.HandlerFunc {
+	if prefix == "" {
+		return h
+	}
+
+	// ensure that prefix always has both a leading and trailing slash so
+	// that relative links for JS and CSS assets work correctly.
+	if prefix[0:1] != "/" {
```
maisem, line 484: reads better if you use `strings.HasPrefix` and `strings.HasSuffix`

maisem (approved):

client/web/web.go
willnorris: 🤦🏻 No idea why I didn't do that

Issue #9182, opened 2023-08-31
CI: linux-amd64 "bench all" takes longer than tests


bradfitz:
Our CI does:

```yaml
    - name: bench all
      run: PATH=$PWD/tool:$PATH /tmp/testwrapper ./... ${{matrix.buildflags}} -bench=. -benchtime=1x -run=^$ 
      env:
        GOARCH: ${{ matrix.goarch }}
```

But on `test (amd64, -race)` that ends up taking longer than the "test all" step.

The point of that "bench all" with "1x" is basically to make sure things compile & barely run.

It shouldn't be running any tests or taking barely any time at all.

I think something is broken. Is `testwrapper` not passing `-run=^$` down?

/cc @maisem @raggi

bradfitz:
3 min locally for me on an M1, vs 20s without `-race`.

But trimming it to just the 34 packages with benchmarks drops it to 44 seconds:

```sh
% time ./tool/go test -race -bench=. -benchtime=1x '-run=^$' $(for x in $(git grep -l "^func Benchmark" | xargs dirname | sort | uniq); do echo "./$x"; done)
...
./tool/go test -race -bench=. -benchtime=1x '-run=^$'   12.80s user 3.28s system 36% cpu 44.282 total
```

PR #9190, opened 2023-08-31
.github: do not use testwrapper for benchmarks

maisem:
Updates #9182

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -94,7 +94,7 @@ jobs:
       env:
         GOARCH: ${{ matrix.goarch }}
     - name: bench all
-      run: PATH=$PWD/tool:$PATH /tmp/testwrapper ./... ${{matrix.buildflags}} -bench=. -benchtime=1x -run=^$ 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -141,10 +141,12 @@ jobs:
           ${{ github.job }}-${{ runner.os }}-go-2-${{ hashFiles('**/go.sum') }}
           ${{ github.job }}-${{ runner.os }}-go-2-
     - name: test
+      run: go run ./cmd/testwrapper ./...
+    - name: bench all
       # Don't use -bench=. -benchtime=1x.
       # Somewhere in the layers (powershell?)
       # the equals signs cause great confusion.
-      run: go run ./cmd/testwrapper ./... -bench . -benchtime 1x
+      run: ./tool/go test ./... -bench . -benchtime 1x -run=^$
```
raggi, line 149: Per the comment above, the `-run=^$` may not work, we might need `-run '^$'` instead

raggi (approved):

.github/workflows/test.yml
maisem: oops, good catch!

bradfitz (approved):

PR #9191, opened 2023-09-01
ipn/ipnlocal,control/controlclient: make Logout more sync

maisem:
We already removed the async API, make it more sync and remove
the FinishLogout state too.
    
This also makes the callback be synchronous again as the previous
attempt was trying to work around the logout callback resulting
in a client shutdown getting blocked forever.
    
Updates #3833
    
### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -143,9 +143,8 @@ type LocalBackend struct {
 	statsLogf             logger.Logf        // for printing peers stats on change
 	sys                   *tsd.System
 	e                     wgengine.Engine // non-nil; TODO(bradfitz): remove; use sys
-	pm                    *profileManager
-	store                 ipn.StateStore // non-nil; TODO(bradfitz): remove; use sys
-	dialer                *tsdial.Dialer // non-nil; TODO(bradfitz): remove; use sys
+	store                 ipn.StateStore  // non-nil; TODO(bradfitz): remove; use sys
+	dialer                *tsdial.Dialer  // non-nil; TODO(bradfitz): remove; use sys
 	backendLogID          logid.PublicID
 	unregisterNetMon      func()
 	unregisterHealthWatch func()
@@ -188,6 +187,7 @@ type LocalBackend struct {
 
 	// The mutex protects the following elements.
 	mu             sync.Mutex
+	pm             *profileManager // mu guards access
 	filterHash     deephash.Sum
 	httpTestClient *http.Client // for controlclient. nil by default, used by tests.
 	ccGen          clientGen    // function for producing controlclient; lazily populated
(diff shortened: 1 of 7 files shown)


ipn/ipnlocal/local.go
```diff
@@ -3959,9 +3939,15 @@ func (b *LocalBackend) LogoutSync(ctx context.Context) error {
...
 
-	err := cc.Logout(ctx)
-	b.stateMachine()
-	return err
+	if err := cc.Logout(ctx); err != nil {
+		return err
+	}
+	b.mu.Lock()
+	if err := b.pm.DeleteProfile(profile.ID); err != nil {
+		b.logf("error deleting profile: %v", err)
```
bradfitz, line 3947: Missing unlock?

ipn/ipnlocal/local.go
```diff
@@ -188,6 +187,7 @@ type LocalBackend struct {
 
 	// The mutex protects the following elements.
 	mu             sync.Mutex
+	pm             *profileManager
```
bradfitz, line 190: // mu guards use of it

Right? Because the field itself never changes?

bradfitz (approved):

ipn/ipnlocal/local.go
```diff
@@ -3936,7 +3907,16 @@ func (b *LocalBackend) ShouldHandleViaIP(ip netip.Addr) bool {
 
 func (b *LocalBackend) LogoutSync(ctx context.Context) error {
```
bradfitz, line 3908: time to rename this to just "Logout"?

and maybe some docs while we're here?

ipn/ipnlocal/local.go
```diff
@@ -3936,7 +3907,16 @@ func (b *LocalBackend) ShouldHandleViaIP(ip netip.Addr) bool {
...
 	b.mu.Lock()
+	if b.state == ipn.NeedsLogin {
```
bradfitz, line 3910: what about NoState? :)

ipn/ipnlocal/local.go
```diff
@@ -3959,9 +3939,16 @@ func (b *LocalBackend) LogoutSync(ctx context.Context) error {
...
 	}
 
-	err := cc.Logout(ctx)
-	b.stateMachine()
-	return err
+	if err := cc.Logout(ctx); err != nil {
+		return err
+	}
+	b.mu.Lock()
+	if err := b.pm.DeleteProfile(profile.ID); err != nil {
```
bradfitz, line 3946: there's still a way to force delete the profile if the cc Logout fails right?

ipn/ipnlocal/local.go
maisem: No can't do it on NoState, only on NeedsLogin. NeedsLogin is set when control has said we are logged out. NoState is when we don't know the state

ipn/ipnlocal/local.go
maisem: I don't think there is today, we need to add one.

ipn/ipnlocal/local.go
maisem: reworked to use presence of NodeKey instead

bradfitz (approved):

PR #9193, opened 2023-09-01
control/controlclient: clean up various things in prep for state overhaul

bradfitz:
We want the overall state (used only for tests) to be computed from the individual states of each component, rather than moving the state around by hand in dozens of places.

In working towards that, we found a lot of things to clean up.

### DIFF
--- a/control/controlclient/auto.go
+++ b/control/controlclient/auto.go
@@ -25,44 +25,28 @@ import (
 )
 
 type LoginGoal struct {
-	_               structs.Incomparable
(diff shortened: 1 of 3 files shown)


bradfitz:
(still with a bunch of gunk in it. slowly picking out the pieces into separate PRs.)

bradfitz:
This has been mostly done already in other PRs, except the `func (c *Auto) computeStateLocked() State {` part.

/cc @maisem

PR #9196, opened 2023-09-01
control/controlclient: remove some channels (quit, updateCh)

bradfitz:
Updates #cleanup


bradfitz:
(We're doing this in #9195 instead)

PR #9198, opened 2023-09-01
net/tsdial: be smarter about when to close SystemDial conns

bradfitz:
It was too aggressive before, as it only had the ill-defined "Major"
bool to work with. Now it can check more precisely.

Updates #9040

### DIFF
--- a/net/tsdial/tsdial.go
+++ b/net/tsdial/tsdial.go
@@ -139,15 +139,40 @@ func (d *Dialer) SetNetMon(netMon *netmon.Monitor) {
 }
 
 func (d *Dialer) linkChanged(delta *netmon.ChangeDelta) {
-	if !delta.Major {
-		return
-	}
 	d.mu.Lock()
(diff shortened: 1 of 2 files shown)


net/tsdial/tsdial.go
```diff
@@ -139,15 +139,39 @@ func (d *Dialer) SetNetMon(netMon *netmon.Monitor) {
...
+	if delta.Old.DefaultRouteInterface != delta.New.DefaultRouteInterface ||
+		delta.Old.HTTPProxy != delta.New.HTTPProxy {
+		return true
+	}
+	if !delta.New.HasIP(lip) && delta.Old.HasIP(lip) {
+		// Our interface with this source IP went away.
+		return true
 	}
+	_ = rip // TODO(bradfitz): use the remote IP?
+	return false
```
maisem, line 174: Could check if the remote addr uses a non-default interface, like if a longer prefix matches or something. That may already be handled by the local ip check though.

PR #9201, opened 2023-09-01
cmd/k8s-operator/manifests,cmd/k8s-nameserver: add a simple nameserver, makes operator populate its config

irbekrm:
This PR is the first part of the work to make it possible for the kube egress proxy to be used to access HTTPS services on tailnet.

- adds an implementation of a very simple nameserver and basic scripts to deploy it on Kubernetes (a manifest and image build target/script update)
- adds a new `--enable-dns` flag to the k8s-operator (false by default)
- adds functionality to the operator that makes it update the nameserver's config as users create/delete egress proxies

Users will then have to manually update their cluster config to add a ts.net stub nameserver pointing at the new nameserver.
Next steps for this will be to add functionality to attempt to automatically update cluster DNS config (behind another 'experimental'  flag)

Flow:
-  cluster admin deploys the operator and sets `--enable-dns=true` flag
- cluster admin deploys the nameserver i.e `kubectl apply -f ./cmd/k8s-operator/manifests/nameserver.yaml` (we still need to push image for this)
- cluster admin modifies cluster DNS config to add [ts.net stub-domain nameserver](https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/#configuration-of-stub-domain-and-upstream-nameserver-using-coredns)
- application developer creates a kube Service that specifies a tailscale service to proxy to via `tailscale.com/ts-tailnet-target-ip`
- k8s-operator sets up a proxy (statefulset with iptables rules that forward packets to the tailscale service etc)
- k8s-operator updates nameserver config so Tailscale services's MagicDNS name is resolved to proxy's kube Service name
- application developer can use Tailscale MagicDNS name to access the Tailscale service from workloads running in cluster

This PR also bumps controller-runtime dependency as it uses the [new namespaced cache functionality from c/r v0.16.0](https://github.com/kubernetes-sigs/controller-runtime/pull/2421) and this caused a couple more dep updates (i.e kubernetes libraries being bumped to v0.28)

Updates https://github.com/tailscale/tailscale/issues/502


cmd/k8s-operator/hostscm.go
```diff
@@ -0,0 +1,158 @@
...
+
+type localClient interface {
+	WhoIs(ctx context.Context, remoteAddr string) (*apitype.WhoIsResponse, error)
+}
+
+// Provision ensures that ts.net nameserver config has been updated with a
+// mapping of Tailscale service to its egress proxy kube Service IP. If
+// successful, returns the Tailscale service FQDN
+func (h *hostsCMProvisioner) Provision(ctx context.Context, logger *zap.SugaredLogger, hcc *hostsCMConfig) (string, error) {
+	logger.Info("IN CM PROV")
```
knyar, line 55: Do you want to merge this eventually? If yes, perhaps it's worth logging the contents of `hcc` as well?

cmd/k8s-operator/hostscm.go
```diff
@@ -0,0 +1,158 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"tailscale.com/client/tailscale/apitype"
+)
+
+const dnsConfigKey = "dns.json"
+
+// hostsCMConfig contains the config needed to update ts.net config for a
+// particular egress proxy
+type hostsCMConfig struct {
+	// IP of the Tailscale service that we are setting up egress for
```
knyar, line 29: Perhaps "Tailscale node" rather than "service"?

cmd/k8s-operator/hostscm.go
```diff
@@ -0,0 +1,158 @@
...
+		// event and the hosts config will get updated then
+		logger.Debugf("proxy Service not yet created waiting...")
+		return "", nil
+	}
+	if err != nil {
+		logger.Errorf("error retrieving proxy Service: %v", err)
+		return "", err
+	}
+
+	if proxySvc == nil || proxySvc.Spec.ClusterIP == "" || proxySvc.Spec.ClusterIP == "None" {
```
knyar, line 77: Do we want to look at [Spec.ClusterIPs](https://github.com/kubernetes/api/blob/7f63b7b88cb6833c4eae362e1fa4de9e06b1ba12/core/v1/types.go#L4838) as well?

cmd/k8s-operator/hostscm.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		logger.Errorf("error marshaling hosts config %v", err)
+		return "", err
+	}
+
+	cm.Data[dnsConfigKey] = string(hostsBytes)
+
+	// TODO (irbekrm): probably better to SSA here
+	// TODO (irbekrm): check diff only apply update if needed
+	if err := h.Update(ctx, cm); err != nil {
```
knyar, line 111: Just curious, what would happen if the ConfigMap was modified between our Get and this Update?

cmd/k8s-operator/hostscm.go
```diff
@@ -0,0 +1,158 @@
...
+		logger.Errorf("failed to update ts.net DNS config: %v", err)
+		return "", err
+	}
+
+	return fqdn, nil
+}
+
+func (h *hostsCMProvisioner) Cleanup(ctx context.Context, logger *zap.SugaredLogger, hcc *hostsCMConfig) error {
```
knyar, line 119: This function looks very similar to `Provision` above, perhaps it's worth having a separate helper function that will return a ConfigMap to be updated, which can be used both by Cleanup and Provision?

Makefile
```diff
@@ -77,6 +77,14 @@ publishdevoperator: ## Build and publish k8s-operator image to location specifie
 	@test "${REPO}" != "ghcr.io/tailscale/k8s-operator" || (echo "REPO=... must not be ghcr.io/tailscale/k8s-operator" && exit 1)
 	TAGS=latest REPOS=${REPO} PUSH=true TARGET=operator ./build_docker.sh
 
+publishdevk8snameserver: ## Build and publish k8s-nameserver image to location specified by ${REPO}
+	@test -n "${REPO}" || (echo "REPO=... required; e.g. REPO=ghcr.io/${USER}/tailscale" && exit 1)
+	@test "${REPO}" != "tailscale/tailscale" || (echo "REPO=... must not be tailscale/tailscale" && exit 1)
+	@test "${REPO}" != "ghcr.io/tailscale/tailscale" || (echo "REPO=... must not be ghcr.io/tailscale/tailscale" && exit 1)
+	@test "${REPO}" != "tailscale/k8s-nameserver" || (echo "REPO=... must not be tailscale/k8s-nameserver" && exit 1)
+	@test "${REPO}" != "ghcr.io/tailscale/k8s-nameserver" || (echo "REPO=... must not be ghcr.io/k8s-tailscale/nameserver" && exit 1)
```
knyar, line 85: I don't fully understand why we are blocking some values of REPO here, but I suspect it should read "tailscale/k8s-nameserver" rather than "k8s-tailscale/nameserver"

cmd/k8s-nameserver/main.go
```diff
@@ -0,0 +1,211 @@
...
+		panic(fmt.Sprintf("error opening udp connection: %v", err))
+	}
+	go func() {
+		<-ctx.Done()
+		conn.Close()
+	}()
+
+	logger("k8s-nameserver listening on: %v", addr)
+
+	for {
```
knyar, line 122: Do we really need to implement our own UDP server here and proxy requests to `tailscale.com/net/dns/resolver`? If we only wanrt to respond to a few DNS names based on a config file, it might be easier to just use miekg's dns.Server (https://jameshfisher.com/2017/08/04/golang-dns-server/)

What tailscale.com/net/dns/resolver also gives us is [automatic forwarding](https://github.com/tailscale/tailscale/blob/51d3220153e25f8144ca7186a0a5d851ffedd32b/net/dns/resolver/tsdns.go#L284) of DNS requests for names it does not keep track of, but we don't actually configure upstream resolvers to forward to, and I am not sure that forwarding unknown requests to another DNS server is actually desired here.

knyar (commented):
I am not qualified to review the operator part, so I mostly have high level comments and a few DNS-related ones.

cmd/k8s-operator/hostscm.go
irbekrm: The Update would fail because the `metadata.resourceVersion` would have changed in kube-apiserver see [docs](https://kubernetes.io/docs/reference/using-api/api-concepts/#:~:text=You%20can%20overwrite,Conflict%20error%20response.) which would cause this reconcile to exit with an error and be retried with backoff till it succeeds. We watch this configmap so we will eventually have the up to date version of the configmap in cache. In practice, it would not take more than one extra reconcile and actually we no-one else is expected to update this configmap.

cmd/k8s-nameserver/main.go
```diff
@@ -0,0 +1,211 @@
...
+func main() {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	logger := log.Printf
+
+	res := resolver.New(logger, nil, nil, &tsdial.Dialer{Logf: logger})
+
+	var configReader configReaderFunc = func() ([]byte, error) {
+		return os.ReadFile(fmt.Sprintf("%s/%s", defaultDNSConfigDir, defaultDNSFile))
```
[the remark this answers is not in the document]
creachadair: Though given it's a file system path, probably `filepath.Join`

cmd/k8s-operator/hostscm.go
irbekrm: Thanks, I was trying to think of a better name, as we already have a Kubernetes service!

cmd/k8s-operator/hostscm.go
irbekrm: Thanks, this needs to be removed

Makefile
irbekrm: >  why we are blocking some values of REPO here

to avoid accidentally using this make target to push to 'prod' if you have creds configured I guess - I just copy pasted it from the other dev image build targets

cmd/k8s-nameserver/main.go
```diff
@@ -0,0 +1,211 @@
...
+				}
+				// it seems like an update to the file has a
+				// potential to produce different types of
+				// events, but only one per configmap update.
+				// fsnotify docs suggest to only react to Write
+				// events, however we should be safe to react to
+				// other events too as we run in a container and
+				// there shouldn't be random changes to file
+				// metadata
+				msg := fmt.Sprintf("new file event: %v", event)
```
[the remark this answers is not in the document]
irbekrm: hmm, I stand corrected, looks like kubelet puts some temp files there during update that we get events on, so filtering on name will help!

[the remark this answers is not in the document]
irbekrm: .. or not as simple https://ahmet.im/blog/kubernetes-inotify/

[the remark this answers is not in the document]
irbekrm: well that would be a potentially sensible, but unsatisfying solution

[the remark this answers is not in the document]
irbekrm: I looked at the kubelet code that does the actual mounting and found a suggestion what clients that want to watch mounted files should do https://github.com/kubernetes/kubernetes/blob/v1.28.1/pkg/volume/util/atomic_writer.go#L39-L61
That filter seems to work nicely, we get a single event when a file is created/a record added/a record removed

cmd/k8s-operator/hostscm.go
irbekrm: No, we want to know that the `spec.clusterIP` field is populated because that's where we read the value from. AFAIU Kubernetes derives that from `spec.clusterIPs` so we aren't missing anything by only looking at that and because we don't create a dual-stack Service there will only ever be one IP address.

rhealitycheck (approved):

cmd/k8s-nameserver/main.go
irbekrm: I am not very strongly opinionated on this.
This uses the internal DNS resolver framework that we have to avoid pulling in another implementation - I originally wanted to use CoreDNS, but we had a chat with @maisem and came to an agreement that we should use one of the implementations we already have in codebase

> What tailscale.com/net/dns/resolver also gives us is [automatic forwarding](https://github.com/tailscale/tailscale/blob/51d3220153e25f8144ca7186a0a5d851ffedd32b/net/dns/resolver/tsdns.go#L284) of DNS requests for names it does not keep track of, but we don't actually configure upstream resolvers to forward to, and I am not sure that forwarding unknown requests to another DNS server is actually desired here.

In practice this should not be a problem- this nameserver will only be used as a [stub nameserver for ts.net subdomains]() and [I have added `ts.net`](https://github.com/tailscale/tailscale/pull/9201/commits/7ed55c26a2e52bb7e8cc41efa2b302817b33334d#diff-e0c39cc52dbd9188fc43ba19c1955045c74511bdcbbd9978230fdeb42714a712R212-R214) to [local domains](https://github.com/tailscale/tailscale/blob/v1.48.2/net/dns/resolver/tsdns.go#L74-L76) of the resolver config, so this should never receive any other requests than for ts.net and if it does not find a record for a ts.net subdomain, it should not attempt to forward to upstream resolver even if one was configured

irbekrm:
Thank you very much for the reviews everyone, I am going to close this PR.
We revisited this recently and as a couple new requirements have come up we agreed on a slightly different design:
- operator creates nameserver resources
- MagicDNS names can be resolved for ingress proxies too (with some caveats)

The latest working implementation is here https://github.com/tailscale/tailscale/tree/irbekrm/clustermagicdns, I will be adding some tests to that and splitting it into a couple different PRs.

PR #9202, opened 2023-09-01
cmd/tailscale: add background mode to serve/funnel wip

tylersmalley:
> **Note**
> Behind the `TAILSCALE_USE_WIP_CODE` flag

In preparing for incoming CLI changes, this PR adds a background mode to both serve and funnel.

See the [parent issue](https://github.com/tailscale/tailscale/issues/8489) for more context.

```
./tool/go run -exec "sudo env TAILSCALE_USE_WIP_CODE=1" ./cmd/tailscaled --socket=/tmp/ts/ts1.sock --state=/tmp/ts/ts1.state --statedir=/Library/Tailscale

TAILSCALE_USE_WIP_CODE=1 ./tool/go run ./cmd/tailscale --socket=/tmp/ts/ts1.sock funnel --bg localhost:3000
TAILSCALE_USE_WIP_CODE=1 ./tool/go run ./cmd/tailscale --socket=/tmp/ts/ts1.sock funnel --set-path=/bar localhost:4000
```

Replaces #9134
Updates #8489

### DIFF
--- a/cmd/tailscale/cli/serve.go
+++ b/cmd/tailscale/cli/serve.go
@@ -39,7 +39,7 @@ var serveCmd = func() *ffcli.Command {
 	// implementation of the tailscale funnel command.
 	// See https://github.com/tailscale/tailscale/issues/7844
 	if envknob.UseWIPCode() {
-		return newServeDevCommand(se, "serve")
+		return newServeDevCommand(se, serve)
 	}
 	return newServeCommand(se)
 }
@@ -158,8 +158,17 @@ type localServeClient interface {
 //
 // It also contains the flags, as registered with newServeCommand.
 type serveEnv struct {
-	// flags
+	// v1 flags
 	json bool // output JSON (status only for now)
+
+	// v2 specific flags
+	bg               bool      // background mode
+	setPath          string    // serve path
+	https            string    // HTTP port
+	http             string    // HTTP port
+	tcp              string    // TCP port
+	tlsTerminatedTcp string    // a TLS terminated TCP port
+	subcmd           serveMode // subcommand
 
 	lc localServeClient // localClient interface, specific to serve
 
--- a/cmd/tailscale/cli/serve_dev.go
+++ b/cmd/tailscale/cli/serve_dev.go
@@ -9,61 +9,110 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"net"
+	"net/url"
 	"os"
 	"os/signal"
+	"path/filepath"
+	"slices"
+	"sort"
 	"strconv"
 	"strings"
 
 	"github.com/peterbourgon/ff/v3/ffcli"
 	"tailscale.com/ipn"
+	"tailscale.com/ipn/ipnstate"
+	"tailscale.com/tailcfg"
 	"tailscale.com/util/mak"
+	"tailscale.com/version"
 )
 
 type execFunc func(ctx context.Context, args []string) error
 
 type commandInfo struct {
+	Name      string
 	ShortHelp string
 	LongHelp  string
 }
 
-var infoMap = map[string]commandInfo{
-	"serve": {
+var serveHelpCommon = strings.TrimSpace(`
+<target> can be a port number (e.g., 3000), a partial URL (e.g., localhost:3000), or a
+full URL including a path (e.g., http://localhost:3000/foo, https+insecure://localhost:3000/foo).
+
+EXAMPLES
+  - Mount a local web server at 127.0.0.1:3000 in the foreground:
+    $ tailscale %s localhost:3000
+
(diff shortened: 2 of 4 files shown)


cmd/tailscale/cli/serve.go
```diff
@@ -160,7 +160,14 @@ type localServeClient interface {
 // It also contains the flags, as registered with newServeCommand.
 type serveEnv struct {
 	// flags
-	json bool // output JSON (status only for now)
+	json             bool   // output JSON (status only for now)
```
marwan-at-work, line 163: Might be a little ugly but we can potentially best this under another struct just distinguish v1 params from v2 ones. Or add a comment at the top

marwan-at-work (commented):
First pass, but will take a closer look after

cmd/tailscale/cli/serve_dev.go
```diff
@@ -83,34 +122,84 @@ func newServeDevCommand(e *serveEnv, subcmd string) *ffcli.Command {
...
 		st, err := e.getLocalClientStatusWithoutPeers(ctx)
 		if err != nil {
 			return fmt.Errorf("getting client status: %w", err)
 		}
 
 		if funnel {
+			// verify node has funnel capabilities
 			if err := e.verifyFunnelEnabled(ctx, st, 443); err != nil {
-				return err
+				return fmt.Errorf("error: %w:", err)
```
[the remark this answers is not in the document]
tylersmalley: I left this one but improved the rest. The underlying code for the funnel verification provides the context which I don't think we should change as part of this.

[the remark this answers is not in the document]
marwan-at-work: @tylersmalley in this case just `return err` -- returning `fmt.Errorf("error: %w")` basically does nothing.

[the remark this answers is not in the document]
tylersmalley: Good point, done.

cmd/tailscale/cli/serve_dev.go
```diff
@@ -5,64 +5,113 @@ package cli
...
 		}, "\n  "),
-		LongHelp:  info.LongHelp,
-		Exec:      e.runServeDev(subcmd == "funnel"),
+		LongHelp: info.LongHelp + fmt.Sprintf(strings.TrimSpace(serveHelpCommon), subcmd, subcmd),
+		Exec:     e.runServeCombined(subcmd),
+
+		FlagSet: e.newFlags("serve-set", func(fs *flag.FlagSet) {
+			fs.BoolVar(&e.bg, "bg", false, "run the command in the background")
+			fs.StringVar(&e.setPath, "set-path", "", "set a path for a specific target")
+			fs.StringVar(&e.https, "https", "", "default; HTTPS listener")
```
marwan-at-work, line 107: You can put the 443 default here?

cmd/tailscale/cli/serve_dev.go
```diff
@@ -5,64 +5,113 @@ package cli
...
 	LongHelp  string
 }
 
-var infoMap = map[string]commandInfo{
-	"serve": {
+var serveHelpCommon = strings.TrimSpace(`
+<target> can be a port number (e.g., 3000), a partial URL (e.g., localhost:3000), or a
+full URL including a path (e.g., http://localhost:3000/foo, https+insecure://localhost:3000/foo).
+
+EXAMPLES
```
marwan-at-work, line 42: Let's match the older long help with examples of background, serving files, etc?

cmd/tailscale/cli/serve_dev.go
```diff
@@ -83,36 +132,86 @@ func newServeDevCommand(e *serveEnv, subcmd string) *ffcli.Command {
...
 		}
-		if err != nil {
-			return err
+
+		// always run in the background when using --set-path
+		if e.setPath != "" {
+			e.bg = true
 		}
 
+		turnOff := "off" == args[len(args)-1]
```
marwan-at-work, line 160: I ran `tailscale funnel --bg 4545` and worked nicely
But then I just put an "off" at the end and got the "legacy" message.

marwan-at-work (commented):
2nd round, will look at tests next ✌🏼

cmd/tailscale/cli/serve_dev.go
tylersmalley: I don't believe there is a way to differentiate between a default and an input. The only way I see us doing this is to check if the value is 443, but that still doesn't tell us if they specified it.

```
// this should fail
$ tailscale serve --https=443 --tcp=80 3000
```

marwan-at-work (approved):

marwan-at-work:
Follow up issue; https://github.com/tailscale/tailscale/issues/9292

PR #9207, opened 2023-09-01
all: implement AppendText alongside MarshalText

dsnet:
This eventually allows encoding packages that may respect the proposed encoding.TextAppender interface.
The performance gains from this is between 10-30%.

Updates tailscale/corp#14379

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -7,7 +7,6 @@ package tailcfg
 
 import (
 	"bytes"
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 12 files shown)


tailcfg/tailcfg.go
```diff
@@ -1765,18 +1772,6 @@ type Debug struct {
 	Exit *int `json:",omitempty"`
 }
 
-func appendKey(base []byte, prefix string, k [32]byte) []byte {
```
dsnet, line 1768: This is dead code. I suspect the key types in `tailscale.com/tailcfg` were moved to `tailscale.com/types/key` at some point?

bradfitz: yup

bradfitz (approved):
Test corp's tests with this go.mod replaced/go.worked in too?

creachadair (approved):

dsnet:
> Test corp's tests with this go.mod replaced/go.worked in too?

Good call. Found a small bug in `key.NLPublic.CLIString`, which used a different prefix than the `MarshalText` method.

Issue #9213, opened 2023-09-02
In Go code, do not use %w format modifier when logging errors, only use it for wrapping errors with fmt.Errorf()


rodrigc:
### What is the issue?

In a few places in the Go code, it looks like `%w` is used when logging errors.

I don't think you want to do this, because in the log output, it leads to funny looking errors like:

```
router: disabling tunneled IPv6 due to system IPv6 config: %!w(*errors.errorString=&{disable_ipv6 is set})
```

I think that:

1.  `OK`:  use %w in `fmt.Errorf()` to wrap errors
2.  `OK`: use %v in `logf()` when printing out errors

### OS

Linux

### Tailscale version

1.49

### Bug report

BUG-3f7fc157b0949c712ce2ee441e83d2869cdf7282bc253adc827ec04eefc02a02-20230902180819Z-edb7f998896e74da

PR #9214, opened 2023-09-02
client/web, clientupdate, util/linuxfw, wgengine/magicsock: Use %v verb for errors

rodrigc:
Replace %w verb with %v verb when logging errors.
Use %w only for wrapping errors with fmt.Errorf()

Fixes: #9213

Signed-off-by: Craig Rodrigues <rodrigc@crodrigues.org>

### DIFF
--- a/client/web/web.go
+++ b/client/web/web.go
@@ -457,7 +457,7 @@ func (s *Server) csrfKey() []byte {
 	// create a new key
 	key := make([]byte, 32)
 	if _, err := rand.Read(key); err != nil {
-		log.Fatal("error generating CSRF key: %w", err)
+		log.Fatalf("error generating CSRF key: %v", err)
 	}
 
 	// if running in CGI mode, try to write the newly created key to disk, and exit if it fails.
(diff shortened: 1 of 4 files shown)


rodrigc:
WIthout this fix, the logs have funny looking lines like this:

```
router: disabling tunneled IPv6 due to system IPv6 config: %!w(*errors.errorString=&{disable_ipv6 is set})
```

client/web/web.go
```diff
@@ -457,7 +458,7 @@ func (s *Server) csrfKey() []byte {
 	// create a new key
 	key := make([]byte, 32)
 	if _, err := rand.Read(key); err != nil {
-		log.Fatal("error generating CSRF key: %w", err)
+		log.Fatal("error generating CSRF key: %v", err)
```
bradfitz, line 461: Fatalf, not log.Fatal

client/web/web.go
```diff
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"github.com/gorilla/csrf"
+
```
bradfitz, line 25: remove this extra newline

bradfitz (commented):
Thanks! Few comments...

client/web/web.go
rodrigc: Fixed

client/web/web.go
rodrigc: Fixed

bradfitz (approved):

PR #9215, opened 2023-09-02
ipn/ipnlocal: fix deadlock in resetControlClientLocked

maisem:
resetControlClientLocked is called while b.mu was held and would call cc.Shutdown which would wait for the observer queue to drain.
However, there may be active callbacks from cc already waiting for b.mu resulting in a deadlock.

This makes it so that resetControlClientLocked does not call Shutdown, and instead just returns the value.
It also makes it so that any status received from previous cc are ignored.

Updates tailscale/corp#12827

### DIFF
--- a/control/controlclient/direct.go
+++ b/control/controlclient/direct.go
@@ -104,7 +104,11 @@ type Direct struct {
 // Observer is implemented by users of the control client (such as LocalBackend)
 // to get notified of changes in the control client's status.
 type Observer interface {
-	SetControlClientStatus(Status)
+	// SetControlClientStatus is called when the client has a new status to
+	// report. The Client is provided to allow the Observer to track which
+	// Client is reporting the status, allowing it to ignore stale status
+	// reports from previous Clients.
+	SetControlClientStatus(Client, Status)
 }
 
 type Options struct {
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -897,9 +897,16 @@ func (b *LocalBackend) SetDecompressor(fn func() (controlclient.Decompressor, er
 
 // SetControlClientStatus is the callback invoked by the control client whenever it posts a new status.
 // Among other things, this is where we update the netmap, packet filters, DNS and DERP maps.
-func (b *LocalBackend) SetControlClientStatus(st controlclient.Status) {
+func (b *LocalBackend) SetControlClientStatus(c controlclient.Client, st controlclient.Status) {
+	b.mu.Lock()
+	if b.cc != c {
+		b.logf("Ignoring SetControlClientStatus from old client")
+		b.mu.Unlock()
+		return
+	}
 	// The following do not depend on any data for which we need to lock b.
 	if st.Err != nil {
+		b.mu.Unlock()
 		if errors.Is(st.Err, io.EOF) {
 			b.logf("[v1] Received error: EOF")
 			return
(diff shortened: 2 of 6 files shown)


ipn/ipnlocal/local.go
```diff
@@ -897,9 +897,16 @@ func (b *LocalBackend) SetDecompressor(fn func() (controlclient.Decompressor, er
 
 // SetControlClientStatus is the callback invoked by the control client whenever it posts a new status.
 // Among other things, this is where we update the netmap, packet filters, DNS and DERP maps.
-func (b *LocalBackend) SetControlClientStatus(st controlclient.Status) {
+func (b *LocalBackend) SetControlClientStatus(c controlclient.Client, st controlclient.Status) {
+	b.mu.Lock()
+	if b.cc != c {
```
bradfitz, line 902: new requirement to document on Client: they must be comparable.

control/controlclient/direct.go
```diff
@@ -104,7 +104,7 @@ type Direct struct {
 // Observer is implemented by users of the control client (such as LocalBackend)
 // to get notified of changes in the control client's status.
 type Observer interface {
-	SetControlClientStatus(Status)
+	SetControlClientStatus(Client, Status)
```
bradfitz, line 107: kinda gross, but understandable. gross enough to need docs now, though. it's not entirely obvious.

ipn/ipnlocal/local.go
```diff
@@ -3847,12 +3854,12 @@ func (b *LocalBackend) requestEngineStatusAndWait() {
 	b.statusLock.Unlock()
 }
 
-// resetControlClientLockedAsync sets b.cc to nil, and starts a
-// goroutine to Shutdown the old client. It does not wait for the
-// shutdown to complete.
-func (b *LocalBackend) resetControlClientLockedAsync() {
+// resetControlClientLocked sets b.cc to nil and returns the old value. If the
```
bradfitz, line 3857: returns the old value, which might be nil.

ipn/ipnlocal/local.go
```diff
@@ -4997,7 +5011,10 @@ func (b *LocalBackend) ListProfiles() []ipn.LoginProfile {
 // called to register it as new node.
 func (b *LocalBackend) ResetAuth() error {
 	b.mu.Lock()
-	b.resetControlClientLockedAsync()
+	prevCC := b.resetControlClientLocked()
+	if prevCC != nil {
+		defer prevCC.Shutdown()
```
bradfitz, line 5016: add same comment above: `// call must happen after release b.mu`

bradfitz (commented):
So we're going to need this on all the Observer methods in the future?

Can we instead pass an Observer implementation to the Auto client that does this automatically, checking the current one and otherwise bailing out? Seems like this could be a wrapper type implementing Observer.

bradfitz (approved):

bradfitz:
> Seems like this could be a wrapper type implementing Observer.

... but I think that has its own complications.

This is fine for now, sufficiently documented. I'd later like to nuke the Auto state machine in the future and only have one state machine overall.

ipn/ipnlocal/local.go
maisem: thats covered in the second sentence

ipn/ipnlocal/local.go
maisem: done

control/controlclient/direct.go
maisem: done

ipn/ipnlocal/local.go
maisem: done

PR #9217, opened 2023-09-02
cmd/tailscale: Prefer socket on Darwin.

### LINKED ISSUE #5761: macOS default socket handling of tailscale and tailscaled are different
### What is the issue?

`tailscale` claims it's default socket is `/var/run/tailscaled.socket`, same as `tailscaled`'s, but in fact it dials `[::1]:64511`.

### Steps to reproduce

```
❯ tailscale help
USAGE
  tailscale [flags] <subcommand> [command flags]

For help on subcommands, add --help after: "tailscale status --help".

This CLI is still under active development. Commands and flags will
change in the future.

SUBCOMMANDS
  up         Connect to Tailscale, logging in if needed
  down       Disconnect from Tailscale
  logout     Disconnect from Tailscale and expire current node key
  netcheck   Print an analysis of local network conditions
  ip         Show Tailscale IP addresses
  status     Show state of tailscaled and its connections
  ping       Ping a host at the Tailscale layer, see how it routed
  nc         Connect to a port on a host, connected to stdin/stdout
  ssh        SSH to a Tailscale machine
  version    Print Tailscale version
  web        Run a web server for controlling Tailscale
  file       Send or receive files
  bugreport  Print a shareable identifier to help diagnose issues
  cert       get TLS certs
  lock       Manipulate the tailnet key authority
  licenses   Get open source license information

FLAGS
  --socket string
    	path to tailscaled's unix socket (default /var/run/tailscaled.socket)
```
```
❯ tailscale status
failed to connect to local tailscaled (which appears to be running as tailscaled, pid 22000). Got error: Failed to connect to local Tailscale daemon for /localapi/v0/status; not running? Error: dial tcp [::1]:64511: connect: connection refused
❯ 
❯ tailscale --socket /var/run/tailscaled.socket status
100.105.126.126 monson-mbp-cli       holymonson@  macOS   -
100.94.240.71   dskt                 holymonson@  linux   -
100.95.87.78    zenobia-mba          holymonson@  macOS   -
```

Seems it's introduced  by version `1.30.2`.

### OS

macOS

### OS version

macos 12.6

### Tailscale version

1.30.2

### Bug report

BUG-9d2329911aa0d09a17272bf7484567e965d2956691dcab7aee5a67ecdd98385d-20220927091756Z-0faaddd812048e06

icio:
If Tailscale has been installed and run from the App Store then the `tailscale` CLI only ever attempts to connect to it, unless you explicitly run `tailscale --socket /var/run/tailscale.socket`. This is because the /Library/Tailscale/ipnport symlink stays even after you quit the App Store app, so localTCPPortAndTokenMacsys returns port+token successfully:

https://github.com/tailscale/tailscale/blob/c6fadd6d71ac884e74bcedd3942576c397108093/safesocket/safesocket_darwin.go#L23-L50

and then in localclient we never try the socket after that fails to connect:

https://github.com/tailscale/tailscale/blob/c6fadd6d71ac884e74bcedd3942576c397108093/client/tailscale/localclient.go#L91-L107

With this change, localclient's defaultDialer first tries the socket then falls back to checking the App Store instance as suggested in https://github.com/tailscale/tailscale/issues/5761#issuecomment-1259724660.

I tested the following scenarios:

* Only OSS tailscaled running, `tailscale status` connects via socket.
* Only App Store tailscaled running, `tailscale status` connects via /Library/Tailscale/ipnport TCP port.
* Both running: `tailscale status` connects now via socket.

❓ Previously, if both tailscaled's were running we'd only attempt to connect to the App Store tailscaled and it would succeed. Now we will only connect to the socket. Is this a breaking change? The `--socket` flag lets us specify when we want to connect to the socket, but if that's now the default option, do users need a way to specify that they instead want to connect to the App Store `tailscaled`? An alternative fix might be to keep prefering the App Store connection and if it fails actually fall back to the socket. Then when running both you'd still have the option of connecting to either.

Fixes #5761

### DIFF
--- a/client/tailscale/localclient.go
+++ b/client/tailscale/localclient.go
@@ -92,6 +92,14 @@ func (lc *LocalClient) defaultDialer(ctx context.Context, network, addr string) 
 	if addr != "local-tailscaled.sock:80" {
 		return nil, fmt.Errorf("unexpected URL address %q", addr)
 	}
+
+	// Prefer the socket (typically /var/run/tailscale.socket).
+	s := safesocket.DefaultConnectionStrategy(lc.socket())
+	conn, sockErr := safesocket.Connect(s)
+	if sockErr == nil {
+		return conn, nil
+	}
+
 	if !lc.UseSocketOnly {
 		// On macOS, when dialing from non-sandboxed program to sandboxed GUI running
 		// a TCP server on a random port, find the random port. For HTTP connections,
@@ -99,11 +107,15 @@ func (lc *LocalClient) defaultDialer(ctx context.Context, network, addr string) 
 		if port, _, err := safesocket.LocalTCPPortAndToken(); err == nil {
 			// We use 127.0.0.1 and not "localhost" (issue 7851).
 			var d net.Dialer
-			return d.DialContext(ctx, "tcp", "127.0.0.1:"+strconv.Itoa(port))
+			conn, dialErr := d.DialContext(ctx, "tcp", "127.0.0.1:"+strconv.Itoa(port))
+			if dialErr != nil {
+				return nil, errors.Join(sockErr, dialErr)
+			}
+			return conn, nil
 		}
 	}
-	s := safesocket.DefaultConnectionStrategy(lc.socket())
-	return safesocket.Connect(s)
+
+	return nil, sockErr
 }
 
 // DoLocalRequest makes an HTTP request to the local machine's Tailscale daemon.


bradfitz:
LGTM but we do the Linux-style DCO thing (in lieu of a CLA).

`git commit --amend -s --no-edit` to add it to the commit message, if you agree to https://developercertificate.org/, and then force push it back to this branch.

DentonGentry:
The DCO error is because we want commits to contain a Signed-Off-By line, using `git commit --signoff`
The `issuebot` failure is because non-trivial commits need to link to an issue number in the commit message.

To resolve these you can:
1. `git commit --amend --signoff`
1. Add a lined at the bottom: "Fixes https://github.com/tailscale/tailscale/issues/5761"
1. `git push -f`

bradfitz:
> ❓ Previously, if both tailscaled's were running we'd only attempt to connect to the App Store tailscaled and it would succeed. Now we will only connect to the socket. Is this a breaking change? The --socket flag lets us specify when we want to connect to the socket, but if that's now the default option, do users need a way to specify that they instead want to connect to the App Store tailscaled? An alternative fix might be to keep prefering the App Store connection and if it fails actually fall back to the socket. Then when running both you'd still have the option of connecting to either.

I guess they can force the App Store variant by saying `--socket=`, right?

icio:
The App Store tailscaled is listening on a TCP port so I'm not sure they could use `--socket` to connect to it.

```
$ readlink /Library/Tailscale/ipnport
57338

$ lsof -i :57338
COMMAND     PID USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
Tailscale 44077 icio    7u  IPv4 0x366f221c133bc407      0t0  TCP localhost:57343->localhost:57338 (ESTABLISHED)
Tailscale 44077 icio    9u  IPv4 0x366f221c150e0cb7      0t0  TCP localhost:57400->localhost:57338 (ESTABLISHED)
```

icio:
This is the alternative: https://github.com/tailscale/tailscale/compare/main...icio:tailscale:icio/fallback-socket

It works well, except that you have to properly quit the App Store instance or you get a "Tailscale is stopped" error instead of trying the socket.

icio:
#9217 might be a better option.

icio:
#9234 was merged instead of this one.

Issue #9062, opened 2023-08-24
FR: Specifying node selector for k8s operator proxy pods


zlepper:
### What are you trying to do?

When I tag a service to be available in tailscale it creates a stateful set to handle the proxying. Sadly this proxy doesn't come with any node selectors available so AKS tries to start it on the windows node I have in my cluster. This ofcause doesn't work. 

I can't manually add in the node selector due to how Tailscale operates with the kubernetes api, since it overwrites any changes done manually. 

### How should we solve this?

At the very least adding a default node selector of `kubernetes.io/os=linux` would at least allow tailscale to start. But even better i would like to be able to actually control what the node selector actually is, so i can ensure that tailscale is at least running somewhat close to the service it's going to be talking to. 

### What is the impact of not solving this?

I think i can workaround it by applying a taint to the windows node, but that means i have to modify all my windows workloads to have a toleration to that taint. 

### Anything else?

I wasn't entirely sure if this should be considered a bug.

jakedgy:
What if we just remove the [embedded proxy.yaml](https://github.com/tailscale/tailscale/blob/main/cmd/k8s-operator/manifests/proxy.yaml) file from the operator binary and include it in the container at a path that users can override via ConfigMap?

maisem:
>What if we just remove the [embedded proxy.yaml](https://github.com/tailscale/tailscale/blob/main/cmd/k8s-operator/manifests/proxy.yaml) file from the operator binary and include it in the container at a path that users can override via ConfigMap?

that sgtm, we would need to move some of the fields that are set there (like initContainers) to be set by the operator

zlepper:
> What if we just remove the [embedded proxy.yaml](https://github.com/tailscale/tailscale/blob/main/cmd/k8s-operator/manifests/proxy.yaml) file from the operator binary and include it in the container at a path that users can override via ConfigMap?
> 
> I've started working on a Helm chart for the operator -- it could reuse some of the relevant `podSpecTemplate` fragments to simplify tweaking things on the proxy StatefulSet like annotations, nodeSelectors, tolerations, etc.

That would certainly solve the issue I have :)

jakedgy:
> 

@maisem I propose we just use a ConfigMap with the proxy skeleton as data:

```yaml
kind: ConfigMap
metadata:
  name: <release-name>-tailscale-operator-proxy
data:
  kind: StatefulSet
  ...
```

[...]

PR #9224, opened 2023-09-04
charts/tailscale-operator: WIP add a helm chart for cmd/k8s-operator

jakedgy:
> **Note**
> This is a Work In Progress!

Create a Helm chart to deploy the k8s-operator.

Install using:

```
export TAILSCALE_CLIENT_ID="<client-id>"
export TAILSCALE_CLIENT_SECRET="<client-secret>"
helm -n tailscale install --set "secret.create=true" --set "secret.clientId=$TAILSCALE_CLIENT_ID" --set "secret.clientSecret=$TAILSCALE_CLIENT_SECRET" tailscale-operator .
```

### Remaining Tasks

- [ ] [Load proxy StatefulSet template from ConfigMap](https://github.com/tailscale/tailscale/issues/9062#issuecomment-1703089626)
  - [ ] Make sure the proxy serviceAccountName is not static
- [ ] README generated using helm-docs
- [ ] [Chart tests](https://helm.sh/docs/topics/chart_tests/)
- [ ] CI (readme, linting, etc.)
- [ ] Release management (maybe [chart-releaser-action](https://github.com/helm/chart-releaser-action)?)
- [ ] Clean up `helm create` boilerplate comments

### DIFF
--- a/charts/tailscale-operator/templates/operator-cluster-role.yaml
+++ b/charts/tailscale-operator/templates/operator-cluster-role.yaml
@@ -1,0 +1,19 @@
+{{- if .Values.rbac.create -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    {{- include "tailscale-operator.labels" . | nindent 4 }}
+    app.kubernetes.io/component: operator
+    {{- with .Values.operator.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  name: {{ include "tailscale-operator.fullname" . }}
+rules:
+  - apiGroups: [""]
+    resources: ["events", "services", "services/status"]
+    verbs: ["*"]
+  - apiGroups: ["networking.k8s.io"]
+    resources: ["ingresses", "ingresses/status"]
+    verbs: ["*"]
+{{- end }}
--- a/charts/tailscale-operator/templates/operator-role.yaml
+++ b/charts/tailscale-operator/templates/operator-role.yaml
@@ -1,0 +1,19 @@
+{{- if .Values.rbac.create -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels:
+    {{- include "tailscale-operator.labels" . | nindent 4 }}
+    app.kubernetes.io/component: operator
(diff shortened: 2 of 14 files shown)


jakedgy:
We may want to consider restricting RBAC by watching a set of namespaces, or the entire cluster.  If we're only watching a set of namespaces, we should just create a ClusterRole and RoleBindings in the watched namespaces.

charts/tailscale-operator/templates/operator-cluster-role.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.rbac.create -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
jakedgy, line 3: The ClusterRole defines the necessary access for watching resources (Services, Ingresses) and could be bound either with a ClusterRole if the operator is watching the entire cluster, or by targeted RoleBindings if it is restricted to specific namespaces.

charts/tailscale-operator/templates/operator-role.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.rbac.create -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
jakedgy, line 3: The operator Role defines the necessary access for controlling resources related to the proxy StatefulSet and backing Secrets.  This will only ever be scoped to the namespace in which the operator is deployed, at least as it is designed today.  That could change if the operator adds support for placing proxy pods in the same namespaces as the Service/Ingress that it targets.

jakedgy:
We may want to consider allowing users to scope the operator to specific namespaces that match a labelSelector.  That would allow for deploying multiple operators to expose resources in different namespaces on different tailnets.

mayakacz:
I think you already saw, but similar work is started in https://github.com/tailscale/tailscale/pull/9920. Suggest closing this.

GMartinez-Sisti:
> I think you already saw, but similar work is started in #9920. Suggest closing this.

The above has now been merged so this should be closed.

irbekrm:
Thank you for working on this @jakedgy !

We implemented a chart in a separate PR.

Release is still in progress.
Similarly, we have some work in progress to make proxies more configurable.

PR #9229, opened 2023-09-04
ipn/ipnlocal: prevent cc leaks on multiple Start calls

maisem:


ipn/ipnlocal/local.go
```diff
@@ -1324,7 +1324,7 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 	// but meanwhile we can make Start cheaper here for such a
 	// case and not restart the world (which takes a few seconds).
 	// Instead, just send a notify with the state that iOS needs.
-	if b.startIsNoopLocked(opts) && profileID == b.lastProfileID {
+	if b.startIsNoopLocked(opts) && profileID == b.lastProfileID && profileID != "" {
```
bradfitz, line 1327: we should also fix the iOS client too so we can remove this special case hack.

cc @agottardo

bradfitz (approved):

Issue #9231, opened 2023-09-04
cmd/testwrapper: does not handle test timeouts well


maisem:
It just says the pkg failed but doesn't print the logs.
```
FAIL	tailscale.com/tsnet
```

icio:
Looks like #9232 fixed it. I've added a test in https://github.com/tailscale/tailscale/pull/9276/commits/50befede518f5a300e56c8bd43c2dc5d3302beda

PR #9232, opened 2023-09-04
cmd/testwrapper: emit logs of failed tests on timeout

maisem:
It would just fail the entire pkg, but would not print any logs. It was already tracking all the logs, so have it emit them when the pkg fails/times out.

Updates #9231

### DIFF
--- a/cmd/testwrapper/testwrapper.go
+++ b/cmd/testwrapper/testwrapper.go
@@ -29,17 +29,13 @@ import (
 const maxAttempts = 3
 
 type testAttempt struct {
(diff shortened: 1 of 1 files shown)


crawshaw (approved):
Not clear the ordering is deterministic.

maisem:
> Not clear the ordering is deterministic.

From slack, you're talking about non-deterministic map order in test output which doesn't really end up mattering in this case because the receiving code is written in a way where the tests can run/finish in an arbitrary order especially because of t.Parallel

Issue #7851, opened 2023-04-12
Use matching address families for LocalAPI localhost dials on macOS


bradfitz:
Summarizing an issue that @nickoneill was debugging elsewhere.

LocalAPI on macOS (from the GUI builds) does a `net.Listen("tcp", "127.0.0.1:0")` to run its HTTP server on (which then requires an HTTP basic auth token distributed such that only the GUI process's user can read it)

But the client side (in two places) does a dial to that listen just using the name "localhost", once in Go, and once in Swift. In some cases (which we don't really understand yet) the client can resolve "localhost" to the IPv6 localhost `::1` and then it fails.

We can't use the `IPV6_V6ONLY` sockopt on the listener because that only works on the wildcard address, which we can't use for firewall pop-up warning reasons. We could have two listeners, but then we'd need a little retry loop to get the ephemeral port numbers lined up, making sure the v6 port is free that corresponds to the ephemeral v4 (or complicate the port number distribution to the GUI and have per-family port numbers).

So easiest is probably just to hard-code "127.0.0.1" on the dialing side in both Swift and Go.

nickoneill:
released in 1.40

PR #9234, opened 2023-09-05
cmd/tailscale: Check App Store tailscaled dialable before selecting.

### LINKED ISSUE #5761: macOS default socket handling of tailscale and tailscaled are different

icio:
My earlier PR #9217 attempted to fix the same issue, but suffered from not letting the user connect to non-oss tailscaled if oss tailscaled was listening on the socket. The --socket flag doesn't let you select the mac apps.

Rather than leave the user unable to choose which tailscaled to connect to, this PR keeps the mac/socket preference order the same and checks a bit harder whether the macsys version really is running. Now, we still prefer the App Store Tailscale (even if it's listening with BackendState=Stopped) and you can use --socket to switch. But now if you quit the App Store Tailscale, we'll try the socket without needing the flag.

Fixes #5761

### DIFF
--- a/safesocket/safesocket_darwin.go
+++ b/safesocket/safesocket_darwin.go
@@ -15,6 +15,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 )
 
 func init() {
@@ -48,9 +49,11 @@ func localTCPPortAndTokenMacsys() (port int, token string, err error) {
 		return 0, "", errors.New("empty auth token in sameuserproof file")
 	}
 
-	// The above files exist forever after the first run. Check that the server
-	// is actually running (though it may still be in a Stopped state).
-	conn, err := net.Dial("tcp", "127.0.0.1:"+portStr)
+	// The above files exist forever after the first run of
+	// /Applications/Tailscale.app, so check we can connect to avoid returning a
+	// port nothing is listening on. Connect to "127.0.0.1" rather than
+	// "localhost" due to #7851.
+	conn, err := net.DialTimeout("tcp", "127.0.0.1:"+portStr, time.Second)
 	if err != nil {
 		return 0, "", err
 	}
--- a/safesocket/safesocket_darwin.go
+++ b/safesocket/safesocket_darwin.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -46,6 +47,15 @@ func localTCPPortAndTokenMacsys() (port int, token string, err error) {
 	if auth == "" {
 		return 0, "", errors.New("empty auth token in sameuserproof file")
 	}
+
+	// The above files exist forever after the first run. Check that the server
+	// is actually running (though it may still be in a Stopped state).
+	conn, err := net.Dial("tcp", "127.0.0.1:"+portStr)
+	if err != nil {
+		return 0, "", err
+	}
+	conn.Close()
+
 	return port, auth, nil
 }
 


safesocket/safesocket_darwin.go
```diff
@@ -46,6 +47,15 @@ func localTCPPortAndTokenMacsys() (port int, token string, err error) {
 	if auth == "" {
 		return 0, "", errors.New("empty auth token in sameuserproof file")
 	}
+
+	// The above files exist forever after the first run. Check that the server
+	// is actually running (though it may still be in a Stopped state).
+	conn, err := net.Dial("tcp", "127.0.0.1:"+portStr)
```
icio, line 53: We could try even harder than this by making a HTTP request and looking for Tailscale headers:

```
$ curl -si http://127.0.0.1:61115/ | grep ^Tailscale
Tailscale-Cap: 68
Tailscale-Version: 1.48.1-t528f95da6-gc4e4acad7
```

or checking that it's a process called Tailscale doing the listening:

```
$ lsof -i:61115 -F | grep ^cTailscale
cTailscale
```

safesocket/safesocket_darwin.go
```diff
@@ -46,6 +47,15 @@ func localTCPPortAndTokenMacsys() (port int, token string, err error) {
 	if auth == "" {
 		return 0, "", errors.New("empty auth token in sameuserproof file")
 	}
+
+	// The above files exist forever after the first run. Check that the server
+	// is actually running (though it may still be in a Stopped state).
```
knyar, line 52: I'd suggest rephrasing the comment a bit to reflect that we're checking that _some_ server is listening on the port, not that something is running. Perhaps "avoid returning a port that nothing is listening on"?

Also, I was surprised to see `127.0.0.1` here rather than `localhost` and tracked it down to #7851. Could you please also add a short comment referencing that issue?

safesocket/safesocket_darwin.go
knyar, line 53: Would it make sense to use `DialTimeout` here with a fairly short timeout (<1s)?

knyar: If we're doing an L7 check, we could also confirm validity of the auth token :) But realistically it's pretty unlikely that something else will be listening on a randomly chosen port, so I think a simple TCP check is fine.

icio: Switched to DialTimeout in [safesocket: Timeout macsys test dial after 1s, clarify comments.](https://github.com/tailscale/tailscale/pull/9234/commits/c27afd755fcd81345844082427410c62b9f9607d).

safesocket/safesocket_darwin.go
icio: Clarified both points in https://github.com/tailscale/tailscale/pull/9234/commits/c27afd755fcd81345844082427410c62b9f9607d.

knyar (approved):

PR #9235, opened 2023-09-05
tailcfg: Support JSON-unmarshalling UserID from string.

icio:
If we merge this, eventually our clients will support receiving UserID strings, and we can later have the server return them by adding a <code>\`json:",string"\`</code> field tag to ipnstate.PeerStatus.UserID as in #8743.

(An alternative approach might be https://github.com/tailscale/tailscale/issues/7576#issuecomment-1706498888.)

Updates #7576

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -122,6 +122,27 @@ func (u UserID) IsZero() bool {
 	return u == 0
 }
 
+func (u *UserID) UnmarshalJSON(b []byte) error {
+	return json.Unmarshal(unquoteInt(b), (*int64)(u))
+}
+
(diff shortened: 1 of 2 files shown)


icio:
Should this get a new CapabilityVersion?

bradfitz:
@icio, I don't think we can do this without breaking people who depend on the current format from `tailscale status --json`. (think people using custom programs they wrote that are unaware of CapabilityVersion/etc)

I left a comment at https://github.com/tailscale/tailscale/issues/7576#issuecomment-1706985986 with an alternate fix that's purely server-side.

icio:
> people who depend on the current format from `tailscale status --json`. (think people using custom programs they wrote that are unaware of CapabilityVersion/etc)

Ah, gotcha. Thanks for the suggestion 👍

PR #9252, opened 2023-09-06
update gvisor

abarisani:
This PR updates gvisor to a more recent version. See https://github.com/tailscale/tailscale/issues/9253


abarisani:
I forced pushed issue reference in the commit name as requested by the issuebot, so this seems solved but issuebot probably has not been re-triggered by the force pushe.

bradfitz:
It wants your commit message to be formatted like:

```
go.mod: bump gvisor

Updates #9253 
```

wgengine/netstack/netstack.go
```diff
@@ -190,8 +189,8 @@ func Create(logf logger.Logf, tundev *tstun.Wrapper, e wgengine.Engine, mc *magi
 	ipstack.SetPromiscuousMode(nicID, true)
 	// Add IPv4 and IPv6 default routes, so all incoming packets from the Tailscale side
 	// are handled by the one fake NIC we use.
-	ipv4Subnet, _ := tcpip.NewSubnet(tcpip.Address(strings.Repeat("\x00", 4)), tcpip.AddressMask(strings.Repeat("\x00", 4)))
-	ipv6Subnet, _ := tcpip.NewSubnet(tcpip.Address(strings.Repeat("\x00", 16)), tcpip.AddressMask(strings.Repeat("\x00", 16)))
+	ipv4Subnet, _ := tcpip.NewSubnet(tcpip.AddrFromSlice(bytes.Repeat([]byte{0x00}, 4)), tcpip.MaskFromBytes(bytes.Repeat([]byte{0x00}, 4)))
```
bradfitz, line 192: might be easier to not use bytes.Repeat now and just write []byte{0,0,0,0} in both places

wgengine/netstack/netstack.go
```diff
@@ -373,7 +372,7 @@ func (ns *Impl) updateIPs(nm *netmap.NetworkMap) {
 		pa := tcpip.ProtocolAddress{
 			AddressWithPrefix: ipp,
 		}
-		if ipp.Address.To4() == "" {
+		if ipp.Address.String() == "" {
```
bradfitz, line 375: I'd prefer to not allocate here (by calling String) just to test if it's unset

wgengine/netstack/netstack.go
```diff
@@ -789,12 +788,13 @@ func (ns *Impl) shouldHandlePing(p *packet.Parsed) (_ netip.Addr, ok bool) {
...
 func netaddrIPFromNetstackIP(s tcpip.Address) netip.Addr {
-	switch len(s) {
+	switch s.Len() {
 	case 4:
+		s := s.As4()
 		return netaddr.IPv4(s[0], s[1], s[2], s[3])
 	case 16:
 		var a [16]byte
-		copy(a[:], s)
+		copy(a[:], s.AsSlice())
```
bradfitz, line 797: use As16 instead. That's exactly what we need, rather than copying to a slice of length 16.

wgengine/netstack/netstack.go
```diff
@@ -1203,8 +1203,8 @@ func stringifyTEI(tei stack.TransportEndpointID) string {
 
 func ipPortOfNetstackAddr(a tcpip.Address, port uint16) (ipp netip.AddrPort, ok bool) {
 	var a16 [16]byte
-	copy(a16[:], a)
-	switch len(a) {
+	copy(a16[:], a.AsSlice())
```
bradfitz, line 1206: same. use As16 inside the "case 16" block

wgengine/netstack/netstack.go
abarisani: I am not sure `[]byte{0,0,0,0}` would also work for `ipv6Subnet`? Somewhat `[]byte{0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0}` feels worse than bytes.Repeat.

wgengine/netstack/netstack.go
abarisani: Would `ipp.Address.Unspecified()` work?

wgengine/netstack/netstack.go
abarisani: Done in https://github.com/tailscale/tailscale/pull/9252/commits/8368ba975c2e4b9a00c7a3971f7982f3d35c7f02

wgengine/netstack/netstack.go
abarisani: ping?

abarisani:
ping?

bradfitz:
I've been on vacation and now we're in a release freeze for a few days.

You need to add the DCO Signed-Off-By line to the commit for this first too.

abarisani:
> I've been on vacation and now we're in a release freeze for a few days.
> 
> You need to add the DCO Signed-Off-By line to the commit for this first too.

Sorry for that. Done.

bradfitz:
@abarisani, compilation failures are real.

wgengine/netstack/netstack.go
```diff
@@ -233,7 +232,7 @@ func (ns *Impl) Close() error {
 func (ns *Impl) wrapProtoHandler(h func(stack.TransportEndpointID, stack.PacketBufferPtr) bool) func(stack.TransportEndpointID, stack.PacketBufferPtr) bool {
 	return func(tei stack.TransportEndpointID, pb stack.PacketBufferPtr) bool {
 		addr := tei.LocalAddress
-		ip, ok := netip.AddrFromSlice(net.IP(addr))
+		ip, ok := netip.AddrFromSlice(net.IP(addr.AsSlice()))
```
raggi, line 235: The `net.IP()` conversion is no longer needed here

raggi (approved):
Thank you, just one more small thing and I'll be happy to merge with fixes for the failures.

wgengine/netstack/netstack.go
abarisani: Resolved [here](https://github.com/tailscale/tailscale/pull/9252/commits/a7f8c6848096917d266aa24f2f73b9f506fcec26). Thanks.

bradfitz:
Still not compiling.

wgengine/netstack/netstack.go
```diff
@@ -373,7 +372,7 @@ func (ns *Impl) updateIPs(nm *netmap.NetworkMap) {
 		pa := tcpip.ProtocolAddress{
 			AddressWithPrefix: ipp,
 		}
-		if ipp.Address.To4() == "" {
+		if ipp.Address.Unspecified() {
```
raggi, line 375: this isn't equivalent to the prior path, the protocol must be v6 if the address is a specified v6 address as well

raggi (changes requested):

wgengine/netstack/netstack.go
raggi: the commit below fixes this case

raggi (approved):
We'll need to squash, as not all of the commits are stable.

wgengine/netstack/netstack.go
```diff
@@ -190,8 +189,8 @@ func Create(logf logger.Logf, tundev *tstun.Wrapper, e wgengine.Engine, mc *magi
 	ipstack.SetPromiscuousMode(nicID, true)
 	// Add IPv4 and IPv6 default routes, so all incoming packets from the Tailscale side
 	// are handled by the one fake NIC we use.
-	ipv4Subnet, _ := tcpip.NewSubnet(tcpip.Address(strings.Repeat("\x00", 4)), tcpip.AddressMask(strings.Repeat("\x00", 4)))
-	ipv6Subnet, _ := tcpip.NewSubnet(tcpip.Address(strings.Repeat("\x00", 16)), tcpip.AddressMask(strings.Repeat("\x00", 16)))
+	ipv4Subnet, _ := tcpip.NewSubnet(tcpip.AddrFromSlice([]byte{0, 0, 0, 0}), tcpip.MaskFromBytes([]byte{0, 0, 0, 0}))
+	ipv6Subnet, _ := tcpip.NewSubnet(tcpip.AddrFromSlice(bytes.Repeat([]byte{0x00}, 16)), tcpip.MaskFromBytes(bytes.Repeat([]byte{0x00}, 16)))
```
bradfitz, line 193: or `make([]byte, 16)` ?

wgengine/netstack/netstack.go
```diff
@@ -1202,16 +1191,14 @@ func stringifyTEI(tei stack.TransportEndpointID) string {
...
+	switch a.Len() {
 	case 4:
 		return netip.AddrPortFrom(
-			netip.AddrFrom4(*(*[4]byte)(a16[:4])).Unmap(),
+			netip.AddrFrom4(a.As4()).Unmap(),
 			port,
 		), true
 	case 16:
-		return netip.AddrPortFrom(netip.AddrFrom16(a16).Unmap(), port), true
+		return netip.AddrPortFrom(netip.AddrFrom16(a.As16()).Unmap(), port), true
```
bradfitz, line 1201: Does `tcpip.Address.As16` not panic and do the wrapping thing for IPv4 addresses?

If so, you can just use only this line and ditch the switch on Len?

bradfitz (commented):
Thanks!

wgengine/netstack/netstack.go
raggi: Yes, in fact there's a much shorter version, though it drops unmap. I have mixed feelings about that, netstack is _specific_ about mapping, so I believe this to be correct, but I'm wary of it being not directly related. It's not a bad time to land those changes though.

raggi:
Ok, I've squashed this down to two commits, one stable that performs the gvisor bump with minimal other changes, and one that applies the requested refactors which include a small change in semantics around unmapping that appears to be more correct. Both pass tests, so can be committed as individual patches.

raggi:
Thank you @abarisani !

tmm1:
Thanks @raggi!

FYI, updating to a newer gvisor surfaces some issues:

```
$ go get -u gvisor.dev/gvisor@latest
go: upgraded gvisor.dev/gvisor v0.0.0-20230928000133-4fe30062272c => v0.0.0-20231006231926-addac5f2487d

$ go build
~/go/pkg/mod/tailscale.com@v1.1.1-0.20231007034813-6f36f8842c3e/wgengine/netstack/netstack.go:25:2: found packages refs (refcounter.go) and refs_template (refs_template.go) in ~/go/pkg/mod/gvisor.dev/gvisor@v0.0.0-20231006231926-addac5f2487d/pkg/refs
```

bradfitz:
@tmm1, you can't use "latest". Their default branch isn't in normal Go layout. They use Bazel. You need to use their "go" branch which they translate to Go layout.

raggi:
> Thanks @raggi!

Hey @tmm1 long time no see! I hope you're well 😄 

as @bradfitz said, you have to use their go branch, in literal terms that's:

`go get -u gvisor.dev/gvisor@go`

tmm1:
Aha, that did it. Thanks guys!

PR #9261, opened 2023-09-06
tailcfg: define a type for NodeCapability

maisem:
Instead of untyped string, add a type to identify these.

Updates #cleanup

### DIFF
--- a/ipn/serve_test.go
+++ b/ipn/serve_test.go
@@ -9,20 +9,21 @@ import (
 )
 
 func TestCheckFunnelAccess(t *testing.T) {
-	portAttr := "https://tailscale.com/cap/funnel-ports?ports=443,8080-8090,8443,"
+	caps := func(c ...tailcfg.NodeCapability) []tailcfg.NodeCapability { return c }
+	const portAttr tailcfg.NodeCapability = "https://tailscale.com/cap/funnel-ports?ports=443,8080-8090,8443,"
 	tests := []struct {
 		port    uint16
-		caps    []string
+		caps    []tailcfg.NodeCapability
 		wantErr bool
 	}{
-		{443, []string{portAttr}, true}, // No "funnel" attribute
-		{443, []string{portAttr, tailcfg.NodeAttrFunnel}, true},
-		{443, []string{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, false},
-		{8443, []string{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, false},
-		{8321, []string{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, true},
-		{8083, []string{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, false},
-		{8091, []string{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, true},
-		{3000, []string{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, true},
+		{443, caps(portAttr), true}, // No "funnel" attribute
+		{443, caps(portAttr, tailcfg.NodeAttrFunnel), true},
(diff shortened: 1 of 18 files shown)


tailcfg/tailcfg.go
```diff
@@ -1221,11 +1222,22 @@ type CapGrant struct {
...
-// It's a string, but its meaning is application-defined.
-// It must be a URL, like "https://tailscale.com/cap/file-sharing-target" or
-// "https://example.com/cap/read-access".
-type PeerCapability string
+// CapabilityName is a string used to identify a specific capability. Its
+// meaning is application-defined.
+//
+// It must be a URL like "https://tailscale.com/cap/file-send", or a well-known
+// capability name like 'funnel'.
+type CapabilityName string
```
bradfitz, line 1230: Is there any reason for this? why not just make the other two be of type string. there's no inheritance going on here.

what would use this type?

tailcfg/tailcfg.go
```diff
@@ -1221,11 +1222,22 @@ type CapGrant struct {
...
+
+// NodeCapability is a capability granted to a node.
+//
+// Unlike PeerCapability, NodeCapability is not in context of a peer and is
+// granted to the node itself.
+type NodeCapability CapabilityName
```
bradfitz, line 1236: trying to avoid "node attribute"? don't want to use that name or at least mention it as another name for this?

creachadair (approved):

ipn/serve_test.go
```diff
@@ -9,20 +9,20 @@ import (
...
-		{8091, []string{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, true},
-		{3000, []string{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, true},
+		{443, []tailcfg.NodeCapability{portAttr}, true}, // No "funnel" attribute
+		{443, []tailcfg.NodeCapability{portAttr, tailcfg.NodeAttrFunnel}, true},
+		{443, []tailcfg.NodeCapability{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, false},
+		{8443, []tailcfg.NodeCapability{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, false},
+		{8321, []tailcfg.NodeCapability{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, true},
+		{8083, []tailcfg.NodeCapability{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, false},
+		{8091, []tailcfg.NodeCapability{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, true},
+		{3000, []tailcfg.NodeCapability{portAttr, tailcfg.CapabilityHTTPS, tailcfg.NodeAttrFunnel}, true},
```
bradfitz, line 25: make a local `type cap = tailcfg.NodeCapability` to make these less visually noisy?

or `caps := func(x ....tailcfg.NodeCapability) []tailcfg.NodeCapability { return x }` ?

tailcfg/tailcfg.go
maisem: this was only for documentation purposes, can remove

tailcfg/tailcfg.go
```diff
@@ -1911,112 +1913,116 @@ type Oauth2Token struct {
...
-	// These are the capabilities that the self node has as listed in
-	// MapResponse.Node.Capabilities.
-	//
-	// We've since started referring to these as "Node Attributes" ("nodeAttrs"
-	// in the ACL policy file).
+// NodeCapability represents a capability granted to the self node as listed in
+// MapResponse.Node.Capabilities.
+//
+// It must be a URL like "https://tailscale.com/cap/file-sharing", or a
+// well-known capability name like 'funnel'.
```
bradfitz, line 1920: double quotes here like you did in other string

bradfitz, line 1920: mention that this well-known short form is restricted to Tailscale-official things and aren't available for end users?

bradfitz (approved):

PR #9267, opened 2023-09-06
util/limiter: add a keyed token bucket rate limiter

danderson:
Updates tailscale/corp#3601

### DIFF
--- a/util/limiter/limiter.go
+++ b/util/limiter/limiter.go
@@ -1,0 +1,149 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package limiter
+
+import (
+	"sync"
+	"time"
+
+	"tailscale.com/util/lru"
+)
+
+// Limiter is a keyed token bucket rate limiter.
+//
+// Each key gets its own separate token bucket to pull from, enabling
+// enforcement on things like "requests per IP address". To avoid
+// unbounded memory growth, Limiter actually only tracks limits
+// precisely for the N most recently seen keys, and assumes that
+// untracked keys are well-behaved. This trades off absolute precision
+// for bounded memory use, while still enforcing well for outlier
+// keys.
+//
+// As such, Limiter should only be used in situations where "rough"
+// enforcement of outliers only is sufficient, such as throttling
+// egregious outlier keys (e.g. something sending 100 queries per
+// second, where everyone else is sending at most 5).
+//
+// Each key's token bucket behaves like a regular token bucket, with
+// the added feature that a bucket's token count can optionally go
+// negative. This implements a form of "cooldown" for keys that exceed
+// the rate limit: once a key starts getting denied, it must stop
+// requesting tokens long enough for the bucket to return to a
+// positive balance. If the key keeps hammering the limiter in excess
+// of the rate limit, the token count will remain negative, and the
+// key will not be allowed to proceed at all. This is in contrast to
+// the classic token bucket, where a key trying to use more than the
+// rate limit will get capped at the limit, but can still occasionally
+// consume a token as one becomes available.
+//
+// The zero value is a valid limiter that rejects all requests. A
+// useful limiter must specify a Size, Max and RefillInterval.
+type Limiter[K comparable] struct {
+	// Size is the number of keys to track. Only the Size most
+	// recently seen keys have their limits enforced precisely, older
+	// keys are assumed to not be querying frequently enough to bother
+	// tracking.
+	Size int
+
+	// Max is the number of tokens available for a key to consume
+	// before time-based rate limiting kicks in. An unused limiter
+	// regains available tokens over time, up to Max tokens. A newly
+	// tracked key initially receives Max tokens.
+	Max int64
+
+	// RefillInterval is the interval at which a key regains tokens for
+	// use, up to Max tokens.
+	RefillInterval time.Duration
+
+	// Overdraft is the amount of additional tokens a key can be
+	// charged for when it exceeds its rate limit. Each additional
+	// request issued for the key charges one unit of overdraft, up to
+	// this limit. Overdraft tokens are refilled at the normal rate,
+	// and must be fully repaid before any tokens become available for
+	// requests.
+	//
+	// A non-zero Overdraft results in "cooldown" behavior: with a
+	// normal token bucket that bottoms out at zero tokens, an abusive
+	// key can still consume one token every RefillInterval. With a
+	// non-zero overdraft, a throttled key must stop requesting tokens
+	// entirely for a cooldown period, otherwise they remain
+	// perpetually in debt and cannot proceed at all.
+	Overdraft int64
+
+	mu    sync.Mutex
+	cache *lru.Cache[K, *bucket]
+}
+
+// QPSInterval returns the interval between events corresponding to
+// the given queries/second rate.
+//
+// This is a helper to be used when populating Limiter.RefillInterval.
+func QPSInterval(qps float64) time.Duration {
+	return time.Duration(float64(time.Second) / qps)
+}
+
+type bucket struct {
+	cur        int64     // current available tokens
+	lastUpdate time.Time // last timestamp at which cur was updated
+}
+
+// Allow charges the key one token (up to the overdraft limit), and
+// reports whether the key can perform an action.
+func (l *Limiter[K]) Allow(key K) bool {
+	return l.allow(key, time.Now())
+}
+
+func (l *Limiter[K]) allow(key K, now time.Time) bool {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	return l.allowBucketLocked(l.getBucketLocked(key, now), now)
+}
+
+func (l *Limiter[K]) getBucketLocked(key K, now time.Time) *bucket {
+	if l.cache == nil {
+		l.cache = &lru.Cache[K, *bucket]{MaxEntries: l.Size}
+	} else if b := l.cache.Get(key); b != nil {
+		return b
+	}
+	b := &bucket{
+		cur:        l.Max,
+		lastUpdate: now.Truncate(l.RefillInterval),
+	}
+	l.cache.Set(key, b)
+	return b
+}
+
+func (l *Limiter[K]) allowBucketLocked(b *bucket, now time.Time) bool {
+	// Only update the bucket quota if needed to process request.
+	if b.cur <= 0 {
+		l.updateBucketLocked(b, now)
+	}
+	ret := b.cur > 0
+	if b.cur > -l.Overdraft {
+		b.cur--
+	}
+	return ret
+}
+
+func (l *Limiter[K]) updateBucketLocked(b *bucket, now time.Time) {
+	now = now.Truncate(l.RefillInterval)
+	if now.Before(b.lastUpdate) {
+		return
+	}
+	timeDelta := max(now.Sub(b.lastUpdate), 0)
+	tokenDelta := int64(timeDelta / l.RefillInterval)
+	b.cur = min(b.cur+tokenDelta, l.Max)
+	b.lastUpdate = now
+}
+
+// peekForTest returns the number of tokens for key, also reporting
+// whether key was present.
+func (l *Limiter[K]) tokensForTest(key K) (int64, bool) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	if b, ok := l.cache.PeekOk(key); ok {
+		return b.cur, true
+	}
+	return 0, false
+}
--- a/util/limiter/limiter_test.go
+++ b/util/limiter/limiter_test.go
@@ -1,0 +1,151 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package limiter
+
+import (
+	"testing"
+	"time"
+)
+
+const testRefillInterval = time.Second
+
+func TestLimiter(t *testing.T) {
+	// 1qps, burst of 10, 2 keys tracked
+	l := &Limiter[string]{
+		Size:           2,
+		Max:            10,
+		RefillInterval: testRefillInterval,
+	}
+
+	// Consume entire burst
+	now := time.Now().Truncate(testRefillInterval)
+	allowed(t, l, "foo", 10, now)
+	denied(t, l, "foo", 1, now)
+	hasTokens(t, l, "foo", 0)
+
+	allowed(t, l, "bar", 10, now)
+	denied(t, l, "bar", 1, now)
+	hasTokens(t, l, "bar", 0)
+
+	// Refill 1 token for both foo and bar
+	now = now.Add(time.Second + time.Millisecond)
+	allowed(t, l, "foo", 1, now)
+	denied(t, l, "foo", 1, now)
+	hasTokens(t, l, "foo", 0)
+
+	allowed(t, l, "bar", 1, now)
+	denied(t, l, "bar", 1, now)
+	hasTokens(t, l, "bar", 0)
+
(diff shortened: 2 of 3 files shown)


util/limiter/limiter.go
```diff
@@ -0,0 +1,124 @@
...
+// the classic token bucket, where a key trying to use more than the
+// rate limit will get capped at the limit, but can still occasionally
+// consume a token as one becomes available.
+type Limiter[K comparable] struct {
+	// Size is the number of keys to track. Only the Size most
+	// recently seen keys have their limits enforced precisely, older
+	// keys are assumed to not be querying frequently enough to bother
+	// tracking.
+	Size int
+	// TTL is an optional timeout for tracked keys. Keys not used for
```
creachadair, line 42: I'm not sure I understand how this works—is the idea that any key that was registered longer ago than this is dropped, and that the "recency" computation only applies to keys younger than that?

util/limiter/limiter.go
```diff
@@ -0,0 +1,124 @@
...
+	// A non-zero overdraft results in "cooldown" behavior: with a
+	// normal token bucket that bottoms out at zero tokens, an abusive
+	// key can still consume one token every RefillEvery
+	// interval. With a non-zero overdraft, a throttled key must stop
+	// requesting tokens entirely for a cooldown period, otherwise
+	// they remain perpetually in debt and cannot proceed at all.
+	Overdraft int64
+	// RefillEvery is the interval at which a key regains tokens for
+	// use, up to Max tokens.
+	RefillEvery time.Duration
```
creachadair, line 68: Maybe RefillInterval?

util/limiter/limiter.go
```diff
@@ -0,0 +1,124 @@
...
+	// keys are assumed to not be querying frequently enough to bother
+	// tracking.
+	Size int
+	// TTL is an optional timeout for tracked keys. Keys not used for
+	// TTL are removed from the limiter's state, even if they fall
+	// within the Size most recently used. If TTL is zero, up to Size
+	// keys are tracked regardless of how long they've sat unused.
+	TTL time.Duration
+
+	// Burst is the number of tokens available for a key to consume
```
creachadair, line 48: Max

util/limiter/limiter.go
```diff
@@ -0,0 +1,124 @@
...
+	// before time-based rate limiting kicks in. An unused limiter
+	// regains available tokens over time, up to Max tokens. A newly
+	// tracked key initially receives Max tokens.
+	Max int64
+	// Overdraft is the amount of additional tokens a key can consume
```
creachadair, line 53: It might be worth distinguishing "consume" here, which implies we are giving them a lease to query. Perhaps: "Overdraft is a number of additional tokens a key may be charged for when it exceeds its rate limit. Each additional request issued for the key charges one unit of overdraft, up to this limit. Overdraft tokens are refilled at the normal rate, and must be fully repaid before any tokens will be made available for requests." or words to that effect.

util/limiter/limiter.go
danderson: Yes. In implementation terms: the LRU is a regular LRU of size N, but it periodically looks at the tail (least recently used) entries and evicts anything that's not been touched for longer than TTL.

So, if I set up a limiter tracking 1000 keys and with a 10-minute TTL, then hit it once with 1000 different keys, 10 minutes later the LRU will empty itself out because even though no other keys have showed up, all 1000 entries have exceeded their TTL without being touched.

danderson: TBH this may be over-optimizing, the goal is to further limit memory overhead for a limiter that ends up only tracking a keyset that's smaller than Limiter.Size... But each entry only costs an extra 16 bytes, and the expected uses are for hundreds of keys not hundreds of thousands... So sticking to just regular LRU semantics is probably fine, on reflection. I'll switch to a non-expiring LRU and remove this knob.

util/limiter/limiter.go
danderson: Done

util/limiter/limiter.go
danderson: Fixed

util/limiter/limiter.go
danderson: Done

util/limiter/limiter.go
```diff
@@ -0,0 +1,126 @@
...
+	// they remain perpetually in debt and cannot proceed at all.
+	Overdraft int64
+	// RefillEvery is the interval at which a key regains tokens for
+	// use, up to Max tokens.
+	RefillEvery time.Duration
+
+	// even though expirable.LRU is thread-safe, it cannot atomically
+	// upsert, so we have to wrap it in our own mutex.
+	mu    sync.Mutex
+	cache *lru.Cache[K, *bucket]
```
bradfitz, line 69: Is that BSL? We have an implementation in tailscale.com/util/lru that should work instead?

danderson: It's still MPL, but TIL we have one. I'll switch to that.

danderson:
Okay, tests implemented and ready for real now.

util/limiter/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+// of the rate limit, the token count will remain negative, and the
+// key will not be allowed to proceed at all. This is in contrast to
+// the classic token bucket, where a key trying to use more than the
+// rate limit will get capped at the limit, but can still occasionally
+// consume a token as one becomes available.
+type Limiter[K comparable] struct {
+	// Size is the number of keys to track. Only the Size most
+	// recently seen keys have their limits enforced precisely, older
+	// keys are assumed to not be querying frequently enough to bother
+	// tracking.
```
creachadair, line 43: Somewhere we should define the required fields. I think that Size, Max, and RefillInterval must all currently be positive (unless we want to specify some default values so a zero is usable).

util/limiter/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+	Size int
+
+	// Max is the number of tokens available for a key to consume
+	// before time-based rate limiting kicks in. An unused limiter
+	// regains available tokens over time, up to Max tokens. A newly
+	// tracked key initially receives Max tokens.
+	Max int64
+	// Overdraft is the amount of additional tokens a key can be
```
creachadair, line 51: trivial aesthetic request: please leave a blank between the item and the next block comment?

util/limiter/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+	// charged for when it exceeds its rate limit. Each additional
+	// request issued for the key charges one unit of overdraft, up to
+	// this limit. Overdraft tokens are refilled at the normal rate,
+	// and must be fully repaid before any tokens become available for
+	// requests.
+	//
+	// A non-zero Overdraft results in "cooldown" behavior: with a
+	// normal token bucket that bottoms out at zero tokens, an abusive
+	// key can still consume one token every RefillInterval
+	// interval. With a non-zero overdraft, a throttled key must stop
```
creachadair, line 61: I'd drop the second "interval" here, it's not wrong but it feels weird. I think "every RefillInterval" is OK.

util/limiter/limiter.go
creachadair: Separately, it might be nice for us to provide a convenience constructor that takes a size and a rate limit and computes Max and RefillInterval. The calculation isn't hard, but it'd read nicely to get `limiter.New(25, qps)`. YMMV.

util/limiter/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+	mu    sync.Mutex
+	cache *lru.Cache[K, *bucket]
+}
+
+type bucket struct {
+	cur        int64     // current available tokens
+	lastUpdate time.Time // last timestamp at which cur was updated
+}
+
+// Allow reports whether the key can perform an action.
```
creachadair, line 78: Perhaps also: "…and if so charges the key one token."

util/limiter/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+func (l *Limiter[K]) Allow(key K) bool {
+	return l.allow(key, time.Now())
+}
+
+func (l *Limiter[K]) allow(key K, now time.Time) bool {
+	defer func() {
```
creachadair, line 84: Do we really want this here? (It seems like we might want an allowForTest rather than having this on the production path, but maybe I missed something)

util/limiter/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+		log.Printf("after allow: %d tokens, incache=%v", toks, ok)
+	}()
+
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	return l.allowBucketLocked(l.getBucketLocked(key, now), now)
+}
+
+func (l *Limiter[K]) getBucketLocked(key K, now time.Time) *bucket {
+	defer log.Printf("bucket %v", key)
```
creachadair: debugging?

util/limiter/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+	if l.cache == nil {
+		l.cache = &lru.Cache[K, *bucket]{MaxEntries: l.Size}
+	}
+	if b := l.cache.Get(key); b != nil {
```
creachadair, line 97: This could be `else if`, since this Get will always fail if the previous test was true.

util/limiter/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+
+// Allow reports whether the key can perform an action.
+func (l *Limiter[K]) Allow(key K) bool {
+	return l.allow(key, time.Now())
+}
+
+func (l *Limiter[K]) allow(key K, now time.Time) bool {
+	defer func() {
+		toks, ok := l.tokensForTest(key)
+		log.Printf("after allow: %d tokens, incache=%v", toks, ok)
```
andrew-d, line 88: Leftover test code?

util/limiter/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (l *Limiter[K]) getBucketLocked(key K, now time.Time) *bucket {
+	defer log.Printf("bucket %v", key)
+	if l.cache == nil {
+		l.cache = &lru.Cache[K, *bucket]{MaxEntries: l.Size}
+	}
+	if b := l.cache.Get(key); b != nil {
+		return b
+	}
+	log.Printf("new bucket %v", key)
```
andrew-d, line 104: (also here and a few other places)

util/limiter/limiter.go
```diff
@@ -0,0 +1,146 @@
+// Copyright (c) Tailscale Inc & AUTHORS
+// SPDX-License-Identifier: BSD-3-Clause
+
+package limiter
+
+import (
+	_ "embed"
```
andrew-d, line 7: Needed for anything?

util/limiter/limiter.go
danderson: Wow, re-read that code before sending and just glossed over those entirely. Removed throughout, thanks.

util/limiter/limiter.go
danderson: Nope, was an extra thing I decided to do separately for reasons. Removed.

util/limiter/limiter.go
```diff
@@ -0,0 +1,138 @@
...
+	return b
+}
+
+func (l *Limiter[K]) allowBucketLocked(b *bucket, now time.Time) bool {
+	// Only update the bucket quota if needed to process request.
+	if b.cur <= 0 {
+		l.updateBucketLocked(b, now)
+	}
+	ret := b.cur > 0
+	b.cur = max(b.cur-1, -l.Overdraft)
```
creachadair, line 114: It might be worth a comment here that this makes sense because we let cur go negative (that is the consequence of the docs above, but I had to read this a couple times to see that we really do want `max` rather than `min` here).

Not picky about wording, but maybe:

```go
// We allow b.cur to go negative up to the overdraft limit.
```

Or maybe it'd be simpler to read as

```go
if b.cur > -l.Overdraft {
  b.cur--
}
```

?

util/limiter/limiter_test.go
```diff
@@ -0,0 +1,147 @@
...
+	allowed(t, l, "qux", 1, now)
+	hasTokens(t, l, "qux", 9)
+	notInLimiter(t, l, "foo")
+	allowed(t, l, "foo", 10, now)
+	denied(t, l, "foo", 1, now)
+	hasTokens(t, l, "foo", -1)
+}
+
+func allowed(t *testing.T, l *Limiter[string], key string, count int, now time.Time) {
+	for i := 0; i < count; i++ {
```
creachadair, line 117: Probably `t.Helper()` in these.

util/limiter/limiter.go
danderson: Hmm, the issue is that Max doesn't have a universally sensible value for consumer who want "just X QPS". The mathematically correct value would be Max=1, but that's almost always too stuttery in real use, and real use wants to allow some amount of bursting before strict enforcement kicks in... But how much? Only the caller knows.

How's this: I added a `QPSInterval(qps)` helper that computes the value of RefillInterval from the given qps target, so an initialization now looks like:

```
l := limiter.Limiter{
  Size: 1234,
  Max: myWellChosenBurstForThisApplication,
  RefillInterval: limiter.QPSInterval(25.0),
}
```

I also find this pleasant because it doesn't give any particular limiter flavor precedence over another, i.e. overdrafting limiters gets built exactly the same way.

WDYT?

(my side rant is that I think x/time/rate's API of asking for a rate and burst ends up being harder to reason about than just telling the user "look, it's a token bucket. It depletes like this, it refills like this" and letting them reason about the underlying implementation, not just some abstract mathematical construct; hence my mild resistance to presenting a rate+burst constructor rather than helpers for configuring a token bucket specifically)

util/limiter/limiter.go
danderson: Done.

util/limiter/limiter.go
danderson: Done.

util/limiter/limiter.go
danderson: That's not correct though, in the presence of a non-zero Overdraft a denial also consumes a token (albeit an overdraft one). Given the entire purpose of the limiter is to do the accounting on whether or not actions are allowed, "... and also does bookkeeping" seems like the default behavior that isn't worth mentioning? If an exported API allowed querying _without_ doing bookkeeping, IMO that would be noteworthy, but not the reverse? Dunno, thoughts?

util/limiter/limiter.go
danderson: Oo nice, done.

util/limiter/limiter.go
danderson: Yeah, I got carried away with my new min/max toys. Changed to your suggested if.

util/limiter/limiter_test.go
danderson: Done.

util/limiter/limiter.go
danderson: This was leftover debug garbage, removed.

util/limiter/limiter.go
danderson: Removed.

util/limiter/limiter.go
creachadair: I like it.

util/limiter/limiter.go
creachadair: Good point, re: correctness. Mostly what I was trying to get at was the idea that calling this also charges against the rate. So perhaps instead:

```go
// Allow charges the key one token (up to the overdraft limit) and reports whether
// that key can perform an action.
```

?

creachadair (approved):
Just little stuff at this point.

util/limiter/limiter.go
danderson: sg, done.

PR #9271, opened 2023-09-07
util/lru: replace container/list with a custom ring implementation

danderson:
pre-generics container/list is quite unpleasant to use, and the pointer manipulation operations for an LRU are simple enough to implement directly now that we have generic types.

With this change, the LRU uses a ring (aka circularly linked list) rather than a simple doubly-linked list as its internals, because the ring makes list manipulation edge cases more regular: the only remaining edge case is the transition between 0 and 1 elements, rather than also having to deal specially with manipulating the first and last members of the list.

Updates #cleanup

### DIFF
--- a/util/lru/lru.go
+++ b/util/lru/lru.go
@@ -4,10 +4,6 @@
 // Package lru contains a typed Least-Recently-Used cache.
 package lru
 
-import (
-	"container/list"
-)
-
 // Cache is container type keyed by K, storing V, optionally evicting the least
 // recently used items if a maximum size is exceeded.
 //
@@ -22,14 +18,27 @@ type Cache[K comparable, V any] struct {
 	// an item is evicted. Zero means no limit.
 	MaxEntries int
 
-	ll *list.List
-	m  map[K]*list.Element // of *entry[K,V]
+	// head is a ring of LRU values. head points to the most recently
+	// used element, head.prev is the least recently used.
+	//
+	// An LRU is technically a simple list rather than a ring, but
+	// implementing it as a ring makes the list manipulation
+	// operations more regular, because the first/last positions in
+	// the list stop being special.
+	//
+	// head is nil when the LRU is empty.
+	head *entry[K, V]
+	// lookup is a map of all the LRU entries contained in
+	// head. lookup and head always contain exactly the same elements;
+	// lookup is just there to allow O(1) lookups of keys.
+	lookup map[K]*entry[K, V]
 }
 
-// entry is the element type for the container/list.Element.
+// entry is an entry of Cache.
 type entry[K comparable, V any] struct {
-	key   K
-	value V
+	prev, next *entry[K, V]
+	key        K
+	value      V
 }
 
 // Set adds or replaces a value to the cache, set or updating its associated
@@ -38,19 +47,18 @@ type entry[K comparable, V any] struct {
 // If MaxEntries is non-zero and the length of the cache is greater
 // after any addition, the least recently used value is evicted.
 func (c *Cache[K, V]) Set(key K, value V) {
-	if c.m == nil {
-		c.m = make(map[K]*list.Element)
-		c.ll = list.New()
+	if c.lookup == nil {
+		c.lookup = make(map[K]*entry[K, V])
 	}
-	if ee, ok := c.m[key]; ok {
-		c.ll.MoveToFront(ee)
-		ee.Value.(*entry[K, V]).value = value
+	if ent, ok := c.lookup[key]; ok {
+		c.moveToFront(ent)
+		ent.value = value
 		return
 	}
-	ele := c.ll.PushFront(&entry[K, V]{key, value})
-	c.m[key] = ele
+	ent := c.newAtFront(key, value)
+	c.lookup[key] = ent
 	if c.MaxEntries != 0 && c.Len() > c.MaxEntries {
-		c.DeleteOldest()
+		c.deleteOldest()
 	}
 }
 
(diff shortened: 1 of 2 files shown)


danderson:
Extremely confused by the mass test failure, but I pull --rebased and it went away. I think something went wrong in my prior rebase when I peeled this change out of the util/limiter branch. Hopefully okay now :crossed_fingers:

bradfitz:
Missing from this change description is the motivation. What does this improve/hurt? Any new restrictions or properties?

danderson:
Mostly readability. I was trying to add debugging tools to inspect LRU contents, and I was going cross-eyed at the amount of `foo.Value.(*something).otherValue` necessary to do a simple traversal. Memory utilization is a little better too from not having to indirect from list.Element through an interface cast to get at the LRU entries.

If that's unconvincing, I can abandon this and live with the icky internals though. Like the commit says, #cleanup imo.

creachadair (approved):

util/lru/lru.go
```diff
@@ -91,34 +94,75 @@ func (c *Cache[K, V]) GetOk(key K) (value V, ok bool) {
...
+	ent.prev.next = ent.next
+	ent.next.prev = ent.prev
+	ent.prev = c.l.prev
+	ent.next = c.l
+	c.l.prev.next = ent
+	c.l.prev = ent
+	c.l = ent
+}
+
+// deleteOldest removes the oldest entry in the cache. Panics if there
```
bradfitz, line 154: It panic if ...

util/lru/lru.go
```diff
@@ -91,34 +94,75 @@ func (c *Cache[K, V]) GetOk(key K) (value V, ok bool) {
...
+// are no entries in the cache.
+func (c *Cache[K, V]) deleteOldest() { c.deleteElement(c.l.prev) }
+
+// deleteElement removes ent, which must be an element of c.l, from
```
bradfitz, line 158: same: c.l or c.m?

I guess it's saying "already in the ring pointed to by c.l"? Maybe just say "which must already be in the LRU"?

util/lru/lru.go
```diff
@@ -22,14 +18,22 @@ type Cache[K comparable, V any] struct {
 	// an item is evicted. Zero means no limit.
 	MaxEntries int
 
-	ll *list.List
-	m  map[K]*list.Element // of *entry[K,V]
+	// l is a ring of LRU values. l points to the most recently used
```
bradfitz, line 21: I try to avoid `l` as a variable name because in many fonts it's hard to distinguish between `1`, `l`, and `I`.

Maybe `r` for ring?

bradfitz: or `h` for head?

bradfitz (approved):

util/lru/lru.go
danderson: Renamed to `c.head`, and `c.m` to `c.lookup`. While really short var names are nice, in this case the savings on typing seem less important than readability.

util/lru/lru.go
```diff
@@ -91,34 +94,75 @@ func (c *Cache[K, V]) GetOk(key K) (value V, ok bool) {
...
+		ret.next = c.l
+		ret.prev = c.l.prev
+		c.l.prev.next = ret
+		c.l.prev = ret
+	}
+	c.l = ret
+	return ret
+}
+
+// moveToFront moves ent, which must be an element of c.l, to the
```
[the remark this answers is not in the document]
danderson: Reworded. The ring and map are always 1:1 the same, all entries exist in both.

util/lru/lru.go
danderson: Done.

util/lru/lru.go
danderson: Yeah, good call, updated.

danderson:
With bradfitz's nudging (thanks!), I implemented a quick benchmark for the package, to verify what impact this change has on performance. Turns out, it's a bit faster than the container/list version, which wasn't my initial intent, but I'll take it!

util/lru/lru.go
```diff
@@ -22,14 +18,25 @@ type Cache[K comparable, V any] struct {
...
 
-	ll *list.List
-	m  map[K]*list.Element // of *entry[K,V]
+	// head is a ring of LRU values. head points to the most recently
+	// used element, head.prev is the least recently used.
+	//
+	// An LRU is technically a simple list rather than a ring, but
+	// implementing it as a ring makes the list manipulation
+	// operations more regular, because the first/last positions in
+	// the list stop being special.
```
bradfitz, line 27: say it's nil when empty? (kinda implicit, but...)

util/lru/lru.go
```diff
@@ -22,14 +18,25 @@ type Cache[K comparable, V any] struct {
...
+	head *entry[K, V]
+	// lookup is a map of all the LRU entries contained in
+	// head. lookup and head always contain exactly the same elements,
```
bradfitz, line 30: s/,/;/ ?

util/lru/lru_test.go
```diff
@@ -40,3 +43,16 @@ func TestLRU(t *testing.T) {
 		t.Errorf("contains 3; should not")
 	}
 }
+
+func BenchmarkLRU(b *testing.B) {
+	const lruSize = 10
```
bradfitz, line 48: b.ReportAllocs() ?

bradfitz (approved):

util/lru/lru.go
danderson: Done.

util/lru/lru.go
danderson: Done

util/lru/lru_test.go
danderson: Done:

```
       │ before.txt  │              after.txt              │
       │   sec/op    │   sec/op     vs base                │
LRU-32   67.05n ± 2%   59.73n ± 2%  -10.90% (p=0.000 n=20)

       │ before.txt │             after.txt              │
       │    B/op    │    B/op     vs base                │
LRU-32   21.00 ± 0%   10.00 ± 0%  -52.38% (p=0.000 n=20)

       │ before.txt │           after.txt            │
       │ allocs/op  │ allocs/op   vs base            │
LRU-32   0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=20) ¹
```

I don't really understand how we're allocating 10B/op on average, but doing that in 0.000 allocs/op... I guess the way the benchmark is constructed, allocs are infrequent enough that go test rounds it to zero, and so benchstat has no decimals to work with...

But hey, apparently allocating 50% less now!

PR #9272, opened 2023-09-07
ipn, ipn/ipnlocal: remove log streaming for StreamServe

marwan-at-work:
This PR removes the per request logging to the CLI as the CLI will not be displaying those logs initially.

Updates #8489


ipn/ipnlocal/local.go
```diff
@@ -248,7 +248,7 @@ type LocalBackend struct {
 	serveProxyHandlers sync.Map                          // string (HTTPHandler.Proxy) => *httputil.ReverseProxy
 	// serveStreamers is a map for those running Funnel in the foreground
 	// and streaming incoming requests.
-	serveStreamers map[uint16]map[uint32]func(ipn.FunnelRequestLog) // serve port => map of stream loggers (key is UUID)
+	serveStreamers map[uint16]map[uint32]func() // serve port => map of stream loggers (key is UUID)
```
bradfitz, line 251: Is this field even needed?

bradfitz, line 251: Document what the func is

marwan-at-work: not needed 👍🏼

bradfitz (approved):

Issue #9264, opened 2023-09-06
DNS-over-TCP replies using the Tailscale DNS server are not complete


timtailscale:
### What is the issue?

Long DNS records are not resolved through the 100.100.100.100 DNS server if they are using DNS-over-TCP. DNS-over-HTTPS (1.1.1.1, 8.8.8.8) backend servers work, but when the backend DNS servers don't support DoH, truncated UDP responses are correctly retried over TCP to the quad100 DNS server but the forwarded queries from the quad100 server to the backend server are not made over TCP and therefor the query fails or is returned empty.

### Steps to reproduce

Query a large DNS record through the quad100 DNS resolver with backend servers set to a DNS server that does not support DNS-over-HTTPS.

### OS

Linux

### OS version

Ubuntu 22.04.3 LTS

### Tailscale version

1.48.1

DentonGentry:
Assigning because Andrew appears to be working on this.

PR #9288, opened 2023-09-07
net/dns: retry forwarder requests over TCP

andrew-d:
We weren't correctly retrying truncated requests to an upstream DNS server with TCP. Instead, we'd return a truncated request to the user, even if the user was querying us over TCP and thus able to handle a large response.

Updates #9264

Change-Id: Ifb04b563839a9614c0ba03e9c564e8924c1a2bfd

### DIFF
--- a/net/dns/resolver/forwarder.go
+++ b/net/dns/resolver/forwarder.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	dns "golang.org/x/net/dns/dnsmessage"
+	"tailscale.com/control/controlknobs"
 	"tailscale.com/envknob"
 	"tailscale.com/net/dns/publicdns"
 	"tailscale.com/net/dnscache"
@@ -68,6 +69,10 @@ const (
 	// DNS queries to the "fallback" DNS server IP for a known provider
 	// (e.g. how long to wait to query Google's 8.8.4.4 after 8.8.8.8).
 	wellKnownHostBackupDelay = 200 * time.Millisecond
+
+	// tcpQueryTimeout is the timeout for a DNS query performed over TCP.
+	// It matches the default 5sec timeout of the 'dig' utility.
+	tcpQueryTimeout = 5 * time.Second
 )
 
 // txid identifies a DNS transaction.
@@ -180,6 +185,8 @@ type forwarder struct {
 	linkSel ForwardLinkSelector // TODO(bradfitz): remove this when tsdial.Dialer absorbs it
 	dialer  *tsdial.Dialer
 
+	controlKnobs *controlknobs.Knobs // or nil
+
 	ctx       context.Context    // good until Close
 	ctxCancel context.CancelFunc // closes ctx
 
@@ -206,12 +213,13 @@ func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
-func newForwarder(logf logger.Logf, netMon *netmon.Monitor, linkSel ForwardLinkSelector, dialer *tsdial.Dialer) *forwarder {
+func newForwarder(logf logger.Logf, netMon *netmon.Monitor, linkSel ForwardLinkSelector, dialer *tsdial.Dialer, knobs *controlknobs.Knobs) *forwarder {
 	f := &forwarder{
-		logf:    logger.WithPrefix(logf, "forward: "),
-		netMon:  netMon,
-		linkSel: linkSel,
-		dialer:  dialer,
+		logf:         logger.WithPrefix(logf, "forward: "),
+		netMon:       netMon,
+		linkSel:      linkSel,
+		dialer:       dialer,
+		controlKnobs: knobs,
 	}
 	f.ctx, f.ctxCancel = context.WithCancel(context.Background())
 	return f
@@ -443,7 +451,10 @@ func (f *forwarder) sendDoH(ctx context.Context, urlBase string, c *http.Client,
 	return res, err
 }
 
-var verboseDNSForward = envknob.RegisterBool("TS_DEBUG_DNS_FORWARD_SEND")
+var (
+	verboseDNSForward = envknob.RegisterBool("TS_DEBUG_DNS_FORWARD_SEND")
+	skipTCPRetry      = envknob.RegisterBool("TS_DNS_FORWARD_SKIP_TCP_RETRY")
+)
 
 // send sends packet to dst. It is best effort.
 //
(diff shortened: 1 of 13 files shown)


andrew-d:
Marking as draft since I think there's one edge case this doesn't handle: the incoming request is UDP, we retry an upstream query over TCP and get a response, but then need to send it back over UDP. Do we need to truncate the response at that point?

twitchyliquid64:
Im ... surprised we didnt implement looking up DNS over TCP lol.

WRT the edge case: yes. If the request we are forwarding is over UDP, we need to truncate the request, clamp EDNS, and return it back. We don't for TCP.

Maybe we can avoid forwarding over TCP entirely if the request was from UDP? we just return the truncated response and the client can retry over TCP.

net/dns/resolver/forwarder.go
```diff
@@ -577,6 +601,89 @@ func (f *forwarder) sendUDP(ctx context.Context, fq *forwardQuery, rr resolverAn
...
+	metricDNSFwdTCP.Add(1)
+	ctx = sockstats.WithSockStats(ctx, sockstats.LabelDNSForwarderTCP, f.logf)
+
+	// Specify the exact family to work around https://github.com/golang/go/issues/52264
+	tcpFam := "tcp4"
+	if ipp.Addr().Is6() {
+		tcpFam = "tcp6"
+	}
+
+	conn, err := f.dialer.SystemDial(ctx, tcpFam, ipp.String())
```
twitchyliquid64, line 619: I'm not familiar with `SystemDial` - is it already context-aware or should we set some deadlines on the conn?

andrew-d:
> Maybe we can avoid forwarding over TCP entirely if the request was from UDP? we just return the truncated response and the client can retry over TCP.

That's a good point, actually; I just plumbed the "family" through to this function so that we can distinguish whether it's a UDP or TCP query, and then avoid falling back to TCP for an input UDP query (i.e. return the truncated query). Thoughts?

net/dns/resolver/forwarder.go
andrew-d: Yeah, it's context-aware and calls `DialContext`, which properly aborts if the input context is canceled/times out 👍

twitchyliquid64 (approved):
One last thing, might be good to have an envknob for this so we can quickly turn it off if we find issues?

And then turn down the envknob once its stable? yeah/nah?

andrew-d:
Good call on that; I added an envknob that can disable this, along with a controlknob (cc @bradfitz for review since we changed that recently). I figure we may want the ability to turn this off if it turns out to be buggy in some configurations; we can remove this knob after we're confident that things work properly.

andrew-d:
(rebased on top of the merge conflict in #9288)

twitchyliquid64 (approved):
LGTM.

@maisem as we need a stamp for the new controlknob, and Brad is out.

net/dns/resolver/forwarder.go
```diff
@@ -601,6 +730,7 @@ func (f *forwarder) resolvers(domain dnsname.FQDN) []resolverAndDelay {
 type forwardQuery struct {
 	txid   txid
 	packet []byte
+	family string // "tcp" or "udp"
```
maisem, line 733: protocol

maisem: family is also fine

net/dns/resolver/forwarder_test.go
```diff
@@ -240,3 +250,203 @@ func FuzzClampEDNSSize(f *testing.F) {
...
+			Port: tcpLn.Addr().(*net.TCPAddr).Port,
+		})
+		if err == nil {
+			break
+		}
+
+	}
+	if tcpLn == nil || udpLn == nil {
+		tb.Fatal("failed to listen on same port for TCP/UDP")
+	}
```
maisem, line 289: this seems like it'll be super flaky

maisem: you also leak one listener

net/dns/resolver/forwarder_test.go
```diff
@@ -240,3 +250,203 @@ func FuzzClampEDNSSize(f *testing.F) {
...
+	}()
+
+	handleUDP := func(addr netip.AddrPort, req []byte) {
+		onRequest(false, req)
+		if _, err := udpLn.WriteToUDPAddrPort(response, addr); err != nil {
+			tb.Logf("error writing response: %v", err)
+		}
+	}
+
+	go func() {
```
maisem, line 338: track these with a waitgroup and wait on it in cleanup?

net/dns/resolver/forwarder_test.go
andrew-d: FWIW, I ran this test with `gostress -count 5000` and it didn't flake once, so I can't say for sure that it's perfect, but it's hopefully not too bad? This just tries to get the same port, not any specific port, and lets the system assign a random one for TCP (which I'm assuming will be more contended).

andrew-d:
@maisem PTAL? Rebased on top of the `NodeCapMap` changes.

net/dns/resolver/forwarder_test.go
maisem: I think it depends on how happy the CI machines :)

net/dns/resolver/forwarder.go
```diff
@@ -577,6 +623,89 @@ func (f *forwarder) sendUDP(ctx context.Context, fq *forwardQuery, rr resolverAn
...
+	metricDNSFwdTCP.Add(1)
+	ctx = sockstats.WithSockStats(ctx, sockstats.LabelDNSForwarderTCP, f.logf)
+
+	// Specify the exact family to work around https://github.com/golang/go/issues/52264
+	tcpFam := "tcp4"
+	if ipp.Addr().Is6() {
+		tcpFam = "tcp6"
+	}
+
+	conn, err := f.dialer.SystemDial(ctx, tcpFam, ipp.String())
```
maisem, line 641: should we attempt to keep these around or are they a use once connection?

andrew-d: They can be re-used, but some DNS servers will close the connection after a single query so I didn't really want to deal with detecting that, handling cache sizing/eviction, etc.

andrew-d: How so? If we hit this codepath, we just sent a UDP packet and got a truncated response so the device radio should already be active. I'm not super familiar with how battery management works; is a TCP connection different than UDP?

maisem: just thinking about the tcp handshake + retry loops

net/dns/resolver/forwarder_test.go
andrew-d: Done

andrew-d:
@maisem Okay, made the changes to de-flake the "listen on the same port" part of the test; PTAL?

twitchyliquid64 (approved):

net/dns/resolver/forwarder.go
```diff
@@ -477,10 +484,49 @@ func (f *forwarder) send(ctx context.Context, fq *forwardQuery, rr resolverAndDe
...
+		// Successful, non-truncated response; return it.
+		return ret, nil
+	} else if fq.family == "udp" {
+		// If this is a UDP query, return it regardless of whether the
+		// response is truncated or not; the client can retry
+		// communicating with tailscaled over TCP. There's no point
+		// falling back to TCP for a truncated query if we can't return
+		// the results to the client.
+		return ret, nil
+	} else if skipTCPRetry() || (f.controlKnobs != nil && f.controlKnobs.DisableDNSForwarderTCPRetries.Load()) {
```
maisem, line 502: you can drop the elses

net/dns/resolver/forwarder.go
```diff
@@ -477,10 +484,49 @@ func (f *forwarder) send(ctx context.Context, fq *forwardQuery, rr resolverAndDe
...
+
+	// Don't retry if our context is done.
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
+	// Retry over TCP, best-effort; return the truncated UDP response if we
+	// cannot query via TCP.
+	ret2, err2 := f.sendTCP(ctx, fq, rr)
+	if err2 == nil {
```
maisem, line 516:
```
if ret2, err2 := f.sendTCP(ctx, fq, rr); err2 == nil {
}
```

net/dns/resolver/forwarder.go
maisem, line 641: should add a shorter timeout on the context

net/dns/resolver/forwarder_test.go
```diff
@@ -240,3 +251,221 @@ func FuzzClampEDNSSize(f *testing.F) {
...
+		tcpLn.Close()
+		udpLn.Close()
+		tb.Logf("waiting for listeners to finish...")
+		wg.Wait()
+	})
+	return
+}
+
+func TestForwarderTCPFallback(t *testing.T) {
+	envknob.Setenv("TS_DEBUG_DNS_FORWARD_SEND", "true")
```
maisem, line 372: unset after test?

net/dns/resolver/forwarder.go
andrew-d: Added

net/dns/resolver/forwarder_test.go
andrew-d: Good catch; had it read and restore the prior value after the test 👍

andrew-d:
@maisem - addressed all changes; PTAL?

maisem (approved):

PR #9290, opened 2023-09-07
ipn/ipnlocal: add logging and locking to c2n /update

awly:
Log some progress info to make updates more debuggable. Also, track whether an active update is already started and return an error if a concurrent update is attempted.

Some planned future PRs:
* add JSON output to `tailscale update`
* use JSON output from `tailscale update` to provide a more detailed status of in-progress update (stage, download progress, etc)

Updates #6907

### DIFF
--- a/ipn/ipnlocal/c2n.go
+++ b/ipn/ipnlocal/c2n.go
@@ -4,6 +4,7 @@
 package ipnlocal
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -14,6 +15,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
+	"strings"
 	"time"
 
 	"tailscale.com/clientupdate"
@@ -38,7 +40,15 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
 		body, _ := io.ReadAll(r.Body)
 		w.Write(body)
 	case "/update":
-		b.handleC2NUpdate(w, r)
+		switch r.Method {
+		case http.MethodGet:
+			b.handleC2NUpdateGet(w, r)
+		case http.MethodPost:
+			b.handleC2NUpdatePost(w, r)
+		default:
+			http.Error(w, "bad method", http.StatusMethodNotAllowed)
+			return
+		}
 	case "/logtail/flush":
 		if r.Method != "POST" {
 			http.Error(w, "bad method", http.StatusMethodNotAllowed)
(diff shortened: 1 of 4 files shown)


noncombatant (approved):

noncombatant (approved):

ipn/ipnlocal/local.go
```diff
@@ -266,6 +266,15 @@ type LocalBackend struct {
 	// at the moment that tkaSyncLock is taken).
 	tkaSyncLock sync.Mutex
 	clock       tstime.Clock
+
+	// c2nUpdateStatus is the status of c2n-triggered client update, protected
+	// by c2nUpdateStatusMu.
+	c2nUpdateStatusMu sync.Mutex
```
bradfitz, line 272: Ideally we wouldn't have more than one Mutex in this type. I think statusLock and tkaSyncLock shouldn't even exist.

Can you use the `mu` lock instead?

Otherwise you need to document the locking order between b.mu and this one.

ipn/ipnlocal/c2n.go
```diff
@@ -129,19 +130,31 @@ func (b *LocalBackend) handleC2NUpdate(w http.ResponseWriter, r *http.Request) {
...
 		w.Header().Set("Content-Type", "application/json")
-		json.NewEncoder(w).Encode(res)
+		w.Write(resb)
 	}()
 
 	if r.Method == "GET" {
 		return
 	}
+	if b.c2nUpdateStatus.started {
+		res.Err = "update already in progress"
```
bradfitz, line 156: missing return?

ipn/ipnlocal/local.go
bradfitz: Oh, you run `exec.Command(cmdTS, "version", "--json").Output()` while holding the mutex. That's not ideal to do while holding LocalBackend.mu. So you'd need to release the mutex earlier, without `defer b.c2nUpdateStatusMu.Unlock()` , and then be sure to clear it back to false in any error path, like:

```go
   exec.Command(cmdTS, "version", "--json").Output()
  ....
   defer func() {
         if ret.Err != nil { 
               b.setC2NUpdateRunning(false) // (and this would do the lines 207-208)
        }
   }()
```

awly: Ok, switched to `mu` and reorganized things a bit to avoid races.
Out of curiosity: we don't like multiple mutexes for independent data because the locking order can be confusing?

ipn/ipnlocal/c2n.go
```diff
@@ -38,7 +40,14 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
...
 		w.Write(body)
 	case "/update":
-		b.handleC2NUpdate(w, r)
+		switch r.Method {
+		case http.MethodGet:
+			b.handleC2NUpdateGet(w, r)
+		case http.MethodPost:
+			b.handleC2NUpdatePost(w, r)
+		default:
+			http.Error(w, "bad method", http.StatusMethodNotAllowed)
```
creachadair, line 49: `return`

tailcfg/c2ntypes.go
```diff
@@ -50,5 +50,5 @@ type C2NUpdateResponse struct {
 	Supported bool
 
 	// Started indicates whether the update has started.
-	Started bool
+	Started bool `json:",omitempty"`
```
creachadair, line 53: Do we really want this to be omitted? It seems like `false` is an interesting value, not just an absence. (I realize Go doesn't care, but as a matter of protocol it seems useful to have explicit).

ipn/ipnlocal/c2n.go
```diff
@@ -151,6 +159,22 @@ func (b *LocalBackend) handleC2NUpdate(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Check if update was already started, and mark as started.
+	b.mu.Lock()
```
creachadair, line 163: Consider maybe pulling this into a method, e.g., 

```go
if !testAndSetC2NUPdateStarted(true) {
   res.Err = "not supported"
   return
}
defer func() { /* … */ }()
```

where the method checks whether it's already set while holding the lock and reports back.

ipn/ipnlocal/c2n.go
```diff
@@ -111,37 +120,36 @@ func (b *LocalBackend) handleC2N(w http.ResponseWriter, r *http.Request) {
...
+	b.logf("c2n: GET /update response: %s", resb)
+	w.Header().Set("Content-Type", "application/json")
+	w.Write(resb)
+}
 
+func (b *LocalBackend) handleC2NUpdatePost(w http.ResponseWriter, r *http.Request) {
+	b.logf("c2n: POST /update received")
+	res := b.newC2NUpdateResponse()
 	defer func() {
+		resb, err := json.Marshal(res)
```
creachadair, line 143: I think this (and the above) is probably a bit excessive; we control the contents of this type fully within the package. I'd probably simplify it to

```go
json.NewEncoder(w).Encode(res)
```

YMMV.

creachadair (commented):
A few small things.

ipn/ipnlocal/c2n.go
awly: I changed it be able to log the response body.
It's very handy to see what was returned in the local tailscaled logs :)

creachadair: That means we'll be shipping the full response to logtail, though—do we really want that in production? If you think so I don't object, but our logs are already pretty full of spam that was only useful during development. 🙂

awly: Hmm, good point, I didn't consider log shipping.
Even though the response bodies are small, removed them since other logs should give us enough info :+1:

creachadair (approved):

ipn/ipnlocal/local.go
bradfitz: Because we've had too many deadlocks in the past with multiple mutexes. It not only requires a lot more documentation but it also requires more brainpower to follow the rules which empirically we're not smart enough about doing. And/or we don't have good static analysis tooling around it. So in lieu of tooling or brain, we prefer to keep it simple so it's less likely to be screwed up. It also means we can name methods "doFooLocked" with that "Locked" suffix and it's obvious which lock you mean: the only one.

PR #9293, opened 2023-09-07
cmd/tailscale: combine foreground and background serve logic

marwan-at-work:
Previously, foreground mode only worked in the simple case of `tailscale funnel <port>`. This PR ensures that whatever you can do in the background can also be done in the foreground such as setting mount paths or tcp forwarding.

Updates #8489

### DIFF
--- a/cmd/tailscale/cli/serve_dev.go
+++ b/cmd/tailscale/cli/serve_dev.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/peterbourgon/ff/v3/ffcli"
+	"tailscale.com/client/tailscale"
 	"tailscale.com/ipn"
 	"tailscale.com/ipn/ipnstate"
 	"tailscale.com/tailcfg"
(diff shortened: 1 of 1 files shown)


cmd/tailscale/cli/serve_dev.go
```diff
@@ -305,14 +244,42 @@ func (e *serveEnv) setServe(ctx context.Context, st *ipnstate.Status, srvType st
...
+	// nil if no config
+	if sc == nil {
+		sc = new(ipn.ServeConfig)
+	}
+
+	// set parent serve config to always be persisted
+	// at the top level, but a nested config might be
+	// the one that gets manipulated depending on
+	// foreground or background.
+	parnetSC := sc
```
tylersmalley, line 256:
```suggestion
	parentSC := sc
```

marwan-at-work: @tylersmalley fixed

tylersmalley (approved):
LGTM

PR #9324, opened 2023-09-09
net/ping: fix ICMP echo code field to 0

raggi:
The code was trying to pass the ICMP protocol number here (1), which is not a valid code. Many servers will not respond to echo messages with codes other than 0.

### DIFF
--- a/net/ping/ping.go
+++ b/net/ping/ping.go
@@ -303,7 +303,7 @@ func (p *Pinger) Send(ctx context.Context, dest net.Addr, data []byte) (time.Dur
 
 	m := icmp.Message{
 		Type: icmpType,
-		Code: icmpType.Protocol(),
+		Code: 0,
 		Body: &icmp.Echo{
 			ID:   int(p.id),
 			Seq:  int(seq),


crawshaw:
rfc792 page 13 says type 8 for an echo request, type 0 for an echo reply

crawshaw:
oh sorry, you're setting code. yes the rfc says 0 for both.

raggi:
> rfc792 page 13 says type 8 for an echo request, type 0 for an echo reply

Type and code are different fields, for both echo request and echo reply there is only one code, 0.

https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml#icmp-parameters-codes-0
https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml#icmp-parameters-codes-8

crawshaw (approved):
Needs to reference an issue, and perhaps worth seeing if you can check in a test, e.g. the test could not respond to a non-zero code.

raggi:
> Needs to reference an issue, and perhaps worth seeing if you can check in a test, e.g. the test could not respond to a non-zero code.

Yeah, I'm looking at doing this. I actually build a program in cmd/ that's not quite ready for submission yet to diagnose some of the problem.

I want to do a few things here in the near future:

1. Land a tool that can be used to actually test this.
2. Land a way to run privileged tests in CI (simple model would be a CI pipeline that runs specific tests, with `-exec sudo` on specific (safe) hosts).
3. Implement datagram socket based ICMP sends, which requires moving away from net.PacketConn to an API that supports sendmsg, as this will work on many more platforms.

PR #9325, opened 2023-09-09
tailcfg: flesh out some docs on MapResponse, clarify slices w/ omitempty

bradfitz:
Updates #cleanup

### DIFF
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -1077,7 +1077,9 @@ type Endpoint struct {
 	Type EndpointType
 }
 
-// MapRequest is sent by a client to start a long-poll network map updates.
+// MapRequest is sent by a client to either update the control plane
+// about its current state, or to start a long-poll of network map updates.
+//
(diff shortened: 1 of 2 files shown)


tailcfg/tailcfg.go
```diff
@@ -1539,6 +1539,26 @@ type PingResponse struct {
 	IsLocalIP bool `json:",omitempty"`
 }
 
+// MapResponse is the value that's sent by the control plane's coordination
+// server whenever there's a update to tell the client. It's the response to a
+// MapRequest.
+//
+// A client can either send a polling or non-polling MapRequest. If the request
+// isn't a long poll, the corresponding MapResponse will be complete. For
```
maisem, line 1547: It doesn't talk about the read-only streaming vs write-only updates which don't result in a mapResponse

bradfitz: Fleshed out a bit more.

maisem (approved):

Issue #9273, opened 2023-09-07
localapi: add support for conditional writes on SetServeConfig, with "etag" from GetServeConfig


bradfitz:
It should be possible to do a read-modify-write on ServeConfig safely.

We could add a new `ServeConfig` etag response on Get (e.g. the sha256 of the JSON of the ServeConfig) and then let callers do a SetServeConfig with HTTP If-Match semantics, failing if the hash has changed. Then callers can loop on a race. Or just fail and the user and re-run the command. But at least we wouldn't silently blow something else away.

/cc @marwan-at-work @maisem

marwan-at-work:
@bradfitz without breaking the Go client, I see three ways of implementing this: 

1. Adding an ETag field to ipn.ServeConfig, maybe with `json:"-"` and just letting the Go client get/set it via HTTP Headers. 
2. Passing the ETag via `context.Context` and adding some APIs to get/set the if-match headers on the context.
3. Adding a new GetServeConfig SetServeConfig APIs, which might not be worth just for the sake of a single http header, but maybe worth it if we added new APIs for setting a partial config so the client wouldn't need to get the whole config and manipulate it, but just tells the API to set what it wants to set. 

Either of the 3 works for me, or others if I missed another way.

marwan-at-work:
Fixed by #9327

PR #9327, opened 2023-09-09
ipn/localapi, ipn/ipnlocal: add etag support for SetServeConfig

marwan-at-work:
This PR adds optimistic concurrency control in the local client and api in order to ensure multiple writes of the ServeConfig do not conflict with each other.

Updates #9273

### DIFF
--- a/ipn/ipnlocal/serve.go
+++ b/ipn/ipnlocal/serve.go
@@ -5,7 +5,9 @@ package ipnlocal
 
 import (
 	"context"
+	"crypto/sha256"
 	"crypto/tls"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -32,6 +34,11 @@ import (
 	"tailscale.com/util/mak"
 	"tailscale.com/version"
 )
+
+// ErrETagMismatch signals that the given
+// If-Match header does not match with the
+// current etag of a resource.
+var ErrETagMismatch = errors.New("etag mismatch")
 
 // serveHTTPContextKey is the context.Value key for a *serveHTTPContext.
 type serveHTTPContextKey struct{}
@@ -214,13 +221,15 @@ func (b *LocalBackend) updateServeTCPPortNetMapAddrListenersLocked(ports []uint1
 }
 
 // SetServeConfig establishes or replaces the current serve config.
-func (b *LocalBackend) SetServeConfig(config *ipn.ServeConfig) error {
+// ETag is an optional parameter to enforce Optimistic Concurrency Control.
+// If it is an empty string, then the config will be overwritten.
+func (b *LocalBackend) SetServeConfig(config *ipn.ServeConfig, etag string) error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	return b.setServeConfigLocked(config)
+	return b.setServeConfigLocked(config, etag)
 }
 
-func (b *LocalBackend) setServeConfigLocked(config *ipn.ServeConfig) error {
+func (b *LocalBackend) setServeConfigLocked(config *ipn.ServeConfig, etag string) error {
 	prefs := b.pm.CurrentPrefs()
 	if config.IsFunnelOn() && prefs.ShieldsUp() {
 		return errors.New("Unable to turn on Funnel while shields-up is enabled")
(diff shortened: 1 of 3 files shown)


ipn/localapi/localapi.go
```diff
@@ -830,8 +830,9 @@ func (h *Handler) serveServeConfig(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, "serve config denied", http.StatusForbidden)
 			return
 		}
+		config, etag := h.b.ServeConfig()
+		w.Header().Set("ETag", etag)
```
bradfitz, line 834: "Etag" with lowercase t. That's what Set will do anyway.

client/tailscale/localclient.go
```diff
@@ -81,6 +89,31 @@ func (lc *LocalClient) socket() string {
...
+func (lc *LocalClient) WithResponseCallback(f func(*http.Response) error) *LocalClient {
+	return &LocalClient{
+		Dial:             lc.Dial,
+		Socket:           lc.Socket,
+		UseSocketOnly:    lc.UseSocketOnly,
+		tsClient:         lc.tsClient,
+		requestCallback:  lc.requestCallback,
+		responseCallback: f,
+	}
+}
```
bradfitz, line 115: this seems like a lot of mechanism and API to expose in the package to users. I'd rather not show all this.

client/tailscale/localclient.go
```diff
@@ -81,6 +89,31 @@ func (lc *LocalClient) socket() string {
 	return paths.DefaultTailscaledSocket()
 }
 
+// WithRequestHeaders returns a copy of the *LocalClient that will send
+// the given http headers on every request.
```
bradfitz, line 93: HTTP

client/tailscale/localclient.go
```diff
@@ -170,6 +208,22 @@ func IsAccessDeniedError(err error) bool {
 	return errors.As(err, &ae)
 }
 
+// UnprocessableEntityError is an error due to permissions.
```
bradfitz, line 211: permissions? or etag?

ipn/ipnlocal/serve.go
```diff
@@ -214,13 +219,17 @@ func (b *LocalBackend) updateServeTCPPortNetMapAddrListenersLocked(ports []uint1
 }
 
 // SetServeConfig establishes or replaces the current serve config.
-func (b *LocalBackend) SetServeConfig(config *ipn.ServeConfig) error {
+// ETag is an optional parameter to enforce Optimistic Concurrency Control.
+// It is a pointer to a string to ensure backwards compatibility with clients
```
bradfitz, line 223: Pointer to string is pretty weird. Are you implicitly saying here that nil means don't use an etag? Then why can't an empty string mean the same?

ipn/ipnlocal/serve.go
```diff
@@ -254,10 +268,10 @@ func (b *LocalBackend) setServeConfigLocked(config *ipn.ServeConfig) error {
 
 // ServeConfig provides a view of the current serve mappings.
 // If serving is not configured, the returned view is not Valid.
-func (b *LocalBackend) ServeConfig() ipn.ServeConfigView {
+func (b *LocalBackend) ServeConfig() (ipn.ServeConfigView, string) {
```
bradfitz, line 271: undocumented string result that's not obvious from its type

ipn/ipnlocal/serve.go
```diff
@@ -254,10 +268,10 @@ func (b *LocalBackend) setServeConfigLocked(config *ipn.ServeConfig) error {
...
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	return b.serveConfig
+	return b.serveConfig, b.serveConfigETag
```
bradfitz, line 274: how about we just define that the etag is the sha256 of the JSON-ified result? So we don't need to store it and return it, etc?

marwan-at-work: @bradfitz sounds good, do you mean to have clients calculate the etag on their side, or that the LocalBackend still sets the etag header but just on the fly by the handler and not as a persisted field? 

I'm going with the latter, but double checking.

ipn/ipnlocal/serve.go
marwan-at-work: Removing per comment below

ipn/ipnlocal/serve.go
marwan-at-work: Given the changes below, this is no longer needed. Further explanation below for brevity:

Initial intention was to distinguish between clients who don't send etags at all, or might send an empty string one (the initial state of the `serveConfigETag` field. But with your suggestion, `etag` is never an empty string, because even if the serve config is `nil`, then json-ifying that gives you `"null"` which gives a non-empty string sha256.

client/tailscale/localclient.go
marwan-at-work: Removing the whole thing per your comment above ✌🏼

ipn/localapi/localapi.go
```diff
@@ -843,7 +853,12 @@ func (h *Handler) serveServeConfig(w http.ResponseWriter, r *http.Request) {
 			writeErrorJSON(w, fmt.Errorf("decoding config: %w", err))
 			return
 		}
-		if err := h.b.SetServeConfig(configIn); err != nil {
+		etag := r.Header.Get("If-Match")
+		if err := h.b.SetServeConfig(configIn, etag); err != nil {
+			if errors.Is(err, ipnlocal.ErrETagMismatch) {
+				http.Error(w, err.Error(), http.StatusUnprocessableEntity)
```
bradfitz, line 859: per https://http.dev/if-match it looks like you want StatusPreconditionFailed (412), not 422.

bradfitz (commented):
Commit message says "client/tailscale" but this doesn't touch client/tailscale.

Note that client/tailscale is under stricter (non-zero) API stability rules, so don't change signatures there. You can add a new method, though.

ipn/ipnlocal/serve.go
bradfitz: Yes, the latter. The client should still remember it from a get and send it along explicitly in a conditional put later.

bradfitz (approved):

PR #9328, opened 2023-09-10
cmd/tailscale: reduce duplicate calls to LocalBackend in funnel WIP

marwan-at-work:
This PR ensures calls to the LocalBackend are not happening multiples times and ensures the set/unset methods are only manipulating the serve config

Updates #8489

> **Note**
> Behind the `TAILSCALE_USE_WIP_CODE` flag

### DIFF
--- a/cmd/tailscale/cli/serve_dev.go
+++ b/cmd/tailscale/cli/serve_dev.go
@@ -203,85 +203,97 @@ func (e *serveEnv) runServeCombined(subcmd serveMode) execFunc {
 			return errHelp
 		}
 
-		if turnOff {
-			err := e.unsetServe(ctx, srvType, srvPort, mount)
-			if err != nil {
-				fmt.Fprintf(os.Stderr, "error: %v\n\n", err)
-				return errHelp
-			}
-			return nil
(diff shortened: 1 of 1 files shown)


cmd/tailscale/cli/serve_dev.go
```diff
@@ -203,85 +203,97 @@ func (e *serveEnv) runServeCombined(subcmd serveMode) execFunc {
...
+		sc, err := e.lc.GetServeConfig(ctx)
+		if err != nil {
+			return fmt.Errorf("error getting serve config: %w", err)
+		}
+
+		// nil if no config
+		if sc == nil {
+			sc = new(ipn.ServeConfig)
+		}
+		dnsName := strings.TrimSuffix(st.Self.DNSName, ".")
```
tylersmalley, line 215: nit: could move assignment closer to its usage starting on line 263

tylersmalley (approved):

cmd/tailscale/cli/serve_dev.go
marwan-at-work: Sounds good, let's do that in the follow up ✌🏼

PR #9347, opened 2023-09-11
wgengine/magicsock: make peerMap also keyed by NodeID

bradfitz:
In prep for incremental netmap update plumbing (#1909), make peerMap
also keyed by NodeID, as all the netmap node mutations passed around
later will be keyed by NodeID.

In the process, also:

* add envknob.InDevMode, as a signal that we can panic more aggressively
  in unexpected cases.
* pull two moderately large blocks of code in Conn.SetNetworkMap out
  into their own methods
* convert a few more sets from maps to set.Set

Updates #1909

### DIFF
--- a/envknob/envknob.go
+++ b/envknob/envknob.go
@@ -389,11 +389,23 @@ func CanTaildrop() bool { return !Bool("TS_DISABLE_TAILDROP") }
 // SSHPolicyFile returns the path, if any, to the SSHPolicy JSON file for development.
 func SSHPolicyFile() string { return String("TS_DEBUG_SSH_POLICY_FILE") }
 
-// SSHIgnoreTailnetPolicy is whether to ignore the Tailnet SSH policy for development.
+// SSHIgnoreTailnetPolicy reports whether to ignore the Tailnet SSH policy for development.
 func SSHIgnoreTailnetPolicy() bool { return Bool("TS_DEBUG_SSH_IGNORE_TAILNET_POLICY") }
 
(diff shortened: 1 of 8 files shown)


bradfitz:
Test flake is unrelated and fixed by pending https://github.com/tailscale/tailscale/pull/9342

envknob/envknob.go
```diff
@@ -389,12 +389,18 @@ func CanTaildrop() bool { return !Bool("TS_DISABLE_TAILDROP") }
...
 
-// TKASkipSignatureCheck is whether to skip node-key signature checking for development.
+// TKASkipSignatureCheck reports whether to skip node-key signature checking for development.
 func TKASkipSignatureCheck() bool { return Bool("TS_UNSAFE_SKIP_NKS_VERIFICATION") }
 
+// InDevMode reports whether the Tailscale client is running in development mode
+// and should panic more aggressively on errors, log more, etc.
+func InDevMode() bool { return devMode() }
+
+var devMode = RegisterBool("TS_DEBUG_DEV")
```
danderson, line 402: Not sure how to feel about this vs. more specific debug knobs, because I fear it'll end up with being forced to turn this on for a specific purpose and being drowned in unrelated logs and CHECK-fails. It's kinda the same failure mode as having an undifferentiated --verbose-all-the-things, which we've mostly managed to avoid so far.

danderson (approved):

envknob/envknob.go
bradfitz: Replaced with more explicit knob per chat.

Issue #9351, opened 2023-09-11
control/controlknob: two tsnet nodes overwrite each other's control knobs


bradfitz:
Currently controlknob values are stored in package globals.

This means two tsnet nodes in the same process can't have different settings from the control plane. It also makes testing a bit gross, having to tstest.Replace things.

It's time to finally get rid of the globals. (especially as a few more are on their way)

PR #9352, opened 2023-09-11
control/controlknobs, all: add plumbed Knobs type, not global variables

bradfitz:
Previously two tsnet nodes in the same process couldn't have disjoint
sets of controlknob settings from control as both would overwrite each
other's global variables.

This plumbs a new controlknobs.Knobs type around everywhere and hangs
the knobs sent by control on that instead.

Updates #9351

### DIFF
--- a/control/controlclient/map.go
+++ b/control/controlclient/map.go
@@ -15,6 +15,7 @@ import (
 	"strconv"
 	"sync"
 
+	"tailscale.com/control/controlknobs"
 	"tailscale.com/envknob"
 	"tailscale.com/tailcfg"
 	"tailscale.com/tstime"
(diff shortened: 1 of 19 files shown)


crawshaw (approved):

control/controlclient/map.go
```diff
@@ -38,7 +39,8 @@ import (
 // one MapRequest).
 type mapSession struct {
 	// Immutable fields.
-	nu             NetmapUpdater // called on changes (in addition to the optional hooks below)
+	nu             NetmapUpdater       // called on changes (in addition to the optional hooks below)
```
crawshaw, line 42: this is what I was talking about in slack. unrelated to this PR

Issue #9343, opened 2023-09-11
FR: Support UPnP gateways without lease duration


mcheshkov:
### What are you trying to do?

I'm trying to setup direct connections to Tailscale on my laptop. 
My home router does not support UPnP with lease duration.
Because of this tailscale could not setup port forwarding, did not tell me about that, and every connection to laptop is DERPed, when other UPnP apps are working just fine.

More details in extra section

### How should we solve this?

One approach could be to add fallback to permanent lease, similar to Syncthing: if port mapping failed with specific error, retry with adjusted lease duration.
Another is to allow user to specify lease time explicitly (either number of seconds, boolean with "permanent lease" semantic, or even boolean with "MikroTik quirks mode" semantic), to be able to "opt-out" from RFC recommendation.
Both of approaches would require a bit of hacking around `GoodUntil` and `RenewAfter` for `upnpMapping`.

Also it would be a bit helpful if `tailscale netcheck` would report not only presence of UPnP gateway, but status of port forwarding attempts.

### What is the impact of not solving this?

For now I just don't use direct connections, since most of the time both machines are either in home network, or have persistent IPs.
I could setup static port forwarding on router, but since it's a laptop I would rather not.
Main issue for me is when I'm trying to access my laptop sitting at home from my phone from external network.

### Anything else?

Laptop is running tailscale 1.48.1 on Fedora Silverblue 38.
My home router is MikroTik hAP AC3 (model RBD53iG-5HacD2HnD), running RouterOS 7.11.2.
Router has UPnP enabled, and some of apps working just fine with it, e.g. Synthing and Transmission.
`tailscale netcheck` shows UPnP replies, and reports `PortMapping: UPnP`
There are `portmapper: saw UPnP type WANIPConnection1 at http://192.168.88.1:2828/gateway.xml; MikroTik Router (MikroTik)` in system journal.
But all connections to laptop ends up DERPed, and there's no NAT rules in router for tailscale.

After a bit of debugging I've discovered that MikroTik didn't implement support for lease duration; only unlimited permanent leases are supported.

Here's part of UPnP exchange for tailscale just after `systemctl restart tailscaled.service`, observed with `tcpdump -A port 2828`  on laptop:

Request
```
POST /upnp/control/mbmmiboxom/wanipconn-3 HTTP/1.1
Host: 192.168.88.1:2828
User-Agent: Go-http-client/1.1
Content-Length: 620
CONTENT-TYPE: text/xml; charset="utf-8"
[...]

PR #9354, opened 2023-09-11
net/portmapper: fall back to permanent UPnP leases if necessary

andrew-d:
Some routers don't support lease times for UPnP portmapping; let's fall back to adding a permanent lease in these cases. Additionally, add a proper end-to-end test case for the UPnP portmapping behaviour.

Updates #9343

Change-Id: I17dec600b0595a5bfc9b4d530aff6ee3109a8b12


bradfitz (approved):
Nice!

bradfitz:
I assume that merge conflict is my fault. Sorry!

PR #9360, opened 2023-09-12
ipn/ipnlocal: fix missing controlknobs.Knobs plumbing

bradfitz:
I missed connecting some controlknobs.Knobs pieces in 4e91cf20a854 resulting in that breaking control knobs entirely.

Whoops.

The fix in ipn/ipnlocal (where it makes a new controlclient) but to atone, I also added integration tests. Those integration tests use a new "tailscale debug control-knobs" which by itself might be useful for future debugging.

Updates #9351

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -1462,6 +1462,7 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 		Observer:             b,
 		C2NHandler:           http.HandlerFunc(b.handleC2N),
 		DialPlan:             &b.dialPlan, // pointer because it can't be copied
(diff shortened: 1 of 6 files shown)


ipn/ipnlocal/local.go
```diff
@@ -1462,6 +1462,7 @@ func (b *LocalBackend) Start(opts ipn.Options) error {
 		Observer:             b,
 		C2NHandler:           http.HandlerFunc(b.handleC2N),
 		DialPlan:             &b.dialPlan, // pointer because it can't be copied
+		ControlKnobs:         b.sys.ControlKnobs(),
```
bradfitz, line 1465: (the actual bug fix)

The rest is tests.

maisem (approved):

PR #9365, opened 2023-09-12
tstest/integration: add start of integration tests for incremental map updates

bradfitz:
This adds a new integration test with two nodes where the first gets a
incremental MapResponse (with only PeersRemoved set) saying that the
second node disappeared.

This extends the testcontrol package to support sending raw
MapResponses to nodes.

Updates #1909

### DIFF
--- a/tstest/integration/integration_test.go
+++ b/tstest/integration/integration_test.go
@@ -328,6 +328,92 @@ func TestTwoNodes(t *testing.T) {
 	d2.MustCleanShutdown(t)
 }
 
+// tests two nodes where the first gets a incremental MapResponse (with only
+// PeersRemoved set) saying that the second node disappeared.
+func TestIncrementalMapUpdatePeersRemoved(t *testing.T) {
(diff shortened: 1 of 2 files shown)


tstest/integration/integration_test.go
```diff
@@ -300,8 +300,8 @@ func TestTwoNodes(t *testing.T) {
 	t.Logf("node1 SOCKS5 addr: %v", n1Socks)
 	t.Logf("node2 SOCKS5 addr: %v", n2Socks)
 
-	n1.AwaitListening()
 	n2.AwaitListening()
+	n1.AwaitListening()
```
maisem, line 304: Why?

maisem (approved):

tstest/integration/integration_test.go
bradfitz: oh, whoops. at one point I started editing the wrong test and thought I'd restored it. guess not. will revert.

PR #9371, opened 2023-09-12
net/netmon: log when the gateway/self IP changes

andrew-d:
This logs that the gateway/self IP address has changed if one of the new values differs.

Updates #8992

Change-Id: I0919424b68ad97fbe1204dd36317ed6f5915411f

### DIFF
--- a/net/netmon/netmon.go
+++ b/net/netmon/netmon.go
@@ -174,8 +174,14 @@ func (m *Monitor) GatewayAndSelfIP() (gw, myIP netip.Addr, ok bool) {
 		return m.gw, m.gwSelfIP, true
 	}
 	gw, myIP, ok = interfaces.LikelyHomeRouterIP()
+	changed := false
 	if ok {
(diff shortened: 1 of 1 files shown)


net/netmon/netmon.go
```diff
@@ -401,7 +409,7 @@ func (m *Monitor) handlePotentialChange(newState *interfaces.State, forceCallbac
 
 	delta.Major = m.IsMajorChangeFrom(oldState, newState)
 	if delta.Major {
```
bradfitz, line 411: I'm trying to get rid of this "Major" bool (it's so undefined).

Does this need to depend on "Major"? Can it not? :)

andrew-d: @bradfitz - I was going off `Major` because that's what the `gwValid` boolean does. I can have it re-log on every link change, but that seems super noisy ... maybe just log when it changes, in all cases?

andrew-d: Updated; PTAL?

bradfitz (approved):

PR #9379, opened 2023-09-13
wgengine: remove SetDERPMap method from Engine interface

bradfitz:
(continuing the mission of removing rando methods from the Engine
 interface that we don't need anymore)

Updates #cleanup


bradfitz:
But in the process I found https://github.com/tailscale/tailscale/pull/9380 ... so I can merge that one first and then rebase this one against that before submitting.

maisem (approved):

PR #9390, opened 2023-09-13
.github/workflows: fix slack-action format in govulncheck.yml

awly:
Currently slack messages for errors fail:
https://github.com/tailscale/tailscale/actions/runs/6159104272/job/16713248204

```
Error: Unexpected token
 in JSON at position 151
```

This is likely due to the line break in the text. Restructure the message to use separate title/text and fix the slack webhook body.

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -27,8 +27,9 @@ jobs:
           payload: >
             {
               "attachments": [{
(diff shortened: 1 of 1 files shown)


andrew-d (approved):
This drops `by ${{ github.event.head_commit.committer.name }}` which is probably fine but I wanted to flag 😃

awly:
Yeah, that value is always empty on `schedule` triggers and I can't find in github docs how to find the committer name some other way :/

PR #9396, opened 2023-09-13
cmd/tailscale/cli: add set serve validations

marwan-at-work:
This PR adds validations for the new new funnel/serve commands under the following rules:
1. There is always a single config for one port (bg or fg).
2. Foreground configs under the same port cannot co-exists (for now).
3. Background configs can change as long as the serve type is the same.

Updates #8489

### DIFF
--- a/cmd/tailscale/cli/serve_dev.go
+++ b/cmd/tailscale/cli/serve_dev.go
@@ -266,6 +266,9 @@ func (e *serveEnv) runServeCombined(subcmd serveMode) execFunc {
 		if turnOff {
 			err = e.unsetServe(sc, dnsName, srvType, srvPort, mount)
 		} else {
+			if err := validateConfig(sc, srvPort, srvType); err != nil {
+				return err
+			}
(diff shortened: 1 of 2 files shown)


cmd/tailscale/cli/serve_dev.go
```diff
@@ -266,6 +266,9 @@ func (e *serveEnv) runServeCombined(subcmd serveMode) execFunc {
 		if turnOff {
 			err = e.unsetServe(sc, dnsName, srvType, srvPort, mount)
 		} else {
+			if err := validateConfig(sc, srvPort, srvType); err != nil {
```
tylersmalley, line 269: We should move this check to immediately after we fetch the serve config to avoid mutating between fetching and doing this check, which I believe is happening [here](https://github.com/tailscale/tailscale/pull/9396/files#diff-d8fbd814bb4009e6070207323c70fcd9840a03f7a173d0aea0cec79ff943506dR261).

tylersmalley (approved):

Issue #8317, opened 2023-06-11
FR: Allow `Authorization` header passthrough for kubernetes operator `auth-proxy`


samdyzon:
### What are you trying to do?

We are really keen to use the Kubernetes operator to allow access to internal services (non-public facing) in our cluster via Tailscale. We are also really excited about the prospect of using the operator to enable proxying requests to the k8s API (which are all explicitly not publicly accessible) with the same mechanism. However, we do not use Tailscale for managing AuthN/AuthZ to our k8s clusters - this is handled by an external OIDC provider that K8s can communicate with. The current implementation of the operator auth-proxy explicitly removes all Auth headers applied to the request by the client.

We are humbly requesting the ability to disable the stripping of client-provided auth headers during the request proxy, perhaps via an environment variable. This variable could be provided such that the default behaviour is the current behaviour, and the operator must explicitly be configured to allow it. This would allow us to utilise existing techniques for handling AuthN + AuthZ to k8s clusters, but also provide private connectivity to the k8s APIs.

### How should we solve this?

The logic for implementing this is trivial, a simple conditional check within the request handler [(tailscale/cmd/k8s-operator/proxy.go)](https://github.com/tailscale/tailscale/blob/f077b672e4a399f4dcafdfe4f15c75d5a580c7cb/cmd/k8s-operator/proxy.go#L77) to determine if request headers should be replaced. Passing the configuration down to the server is less trivial - I do not know enough about golang to propose how to get the required context into the request/response cycle.

### What is the impact of not solving this?

We, and anyone who has an established OIDC provider for their K8s cluster, would be unable to use the auth-proxy component of the k8s operator with their auth provider. This would mean that the use-case we have in mind for the k8s operator would no longer be viable.

The solution we would continue using is a "bastion host" on the same network as the cluster (not publicly accessible), running Tailscale and a reverse-proxy (Caddy) which forwards requests to our private services (private ingress controller) and the k8s API. This solution is less than ideal due to the fact that Caddy does not support SPDY, so we cannot run `kubectl exec ...` commands using the reverse-proxy (we have to SSH into the node and run these commands, which defeats the whole purpose of this design).

### Anything else?

[...]

PR #9405, opened 2023-09-14
cmd/k8s-operator: add APISERVER_PROXY env

maisem:
The kube-apiserver proxy in the operator would only run in auth proxy mode but thats not always desirable. There are situations where the proxy should just be a transparent proxy and not inject auth headers, so do that using a new env var APISERVER_PROXY and deprecate the AUTH_PROXY env.

THe new env var has three options `false`, `true` and `noauth`.

Updates #8317

### DIFF
--- a/cmd/k8s-operator/proxy.go
+++ b/cmd/k8s-operator/proxy.go
@@ -45,12 +45,52 @@ func addWhoIsToRequest(r *http.Request, who *apitype.WhoIsResponse) *http.Reques
 
 var counterNumRequestsProxied = clientmetric.NewCounter("k8s_auth_proxy_requests_proxied")
 
-// launchAuthProxy launches the auth proxy, which is a small HTTP server that
-// authenticates requests using the Tailscale LocalAPI and then proxies them to
-// the kube-apiserver.
-func launchAuthProxy(zlog *zap.SugaredLogger, restConfig *rest.Config, s *tsnet.Server) {
+type apiServerProxyMode int
+
+const (
+	apiserverProxyModeDisabled apiServerProxyMode = iota
+	apiserverProxyModeEnabled
+	apiserverProxyModeNoAuth
+)
+
+func parseAPIProxyMode() apiServerProxyMode {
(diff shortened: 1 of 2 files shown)


cmd/k8s-operator/proxy.go
```diff
@@ -91,35 +131,41 @@ func (h *authProxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-// runAuthProxy runs an HTTP server that authenticates requests using the
+// runAPIServerProxy runs an HTTP server that authenticates requests using the
 // Tailscale LocalAPI and then proxies them to the Kubernetes API.
 // It listens on :443 and uses the Tailscale HTTPS certificate.
 // s will be started if it is not already running.
 // rt is used to proxy requests to the Kubernetes API.
 //
 // It never returns.
-func runAuthProxy(s *tsnet.Server, rt http.RoundTripper, logf logger.Logf) {
+func runAPIServerProxy(s *tsnet.Server, rt http.RoundTripper, logf logger.Logf, mode apiserverProxyMode) {
```
shayne, line 141: apiServerProxyMode

cmd/k8s-operator/proxy.go
```diff
@@ -91,35 +131,41 @@ func (h *authProxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	h.rp.ServeHTTP(w, addWhoIsToRequest(r, who))
 }
 
-// runAuthProxy runs an HTTP server that authenticates requests using the
+// runAPIServerProxy runs an HTTP server that authenticates requests using the
 // Tailscale LocalAPI and then proxies them to the Kubernetes API.
 // It listens on :443 and uses the Tailscale HTTPS certificate.
 // s will be started if it is not already running.
 // rt is used to proxy requests to the Kubernetes API.
 //
```
shayne, line 139: Add comment for mode

shayne (approved):

shayne (approved):

shayne (approved):

Issue #9377, opened 2023-09-13
Mullvad: default to the Mullvad DNS resolver when no override is configured


raggi:
If there is no global nameserver configured in the admin panel with "Override local DNS" enabled, and "Allow Local Network Access" is disabled, and the user has a LAN DNS server address configured, then DNS resolutions will fail. In some sense this is good: we explicitly have not leaked DNS, which is why it is failing, but in another it is very bad: the user effectively lost internet access.

In this configuration we should default to the in-network `10.64.0.1` address, or one of the DOH addresses (https://mullvad.net/en/help/dns-over-https-and-dns-over-tls/). This behavior would be directly equivalent to the regular exit node behavior that resolves via the peer API of the selected Tailscale exit node.

The current recommended workaround for this issue is documented here: https://tailscale.com/kb/1258/mullvad-exit-nodes/#important-dns-considerations

alexl4321:
important indeed. in a mixed server/desktop environment simply starting to force a DNS as in the workaround descr might create a lot of troubles - especially as you cannot set / force a DNS per single tailscale client (urgently needed). guess you could set individual problematic nodes e.g. servers to accept-dns false but still kinda of not a solution.

PR #9408, opened 2023-09-14
tailcfg: add DNS address list for IsWireGuardOnly nodes

raggi:
Tailscale exit nodes provide DNS service over the peer API, however IsWireGuardOnly nodes do not have a peer API, and instead need client DNS parameters passed in their node description.

For Mullvad nodes this will contain the in network 10.64.0.1 address.

Updates #9377

### DIFF
--- a/control/controlclient/map.go
+++ b/control/controlclient/map.go
@@ -693,6 +693,19 @@ func peerChangeDiff(was tailcfg.NodeView, n *tailcfg.Node) (_ *tailcfg.PeerChang
 			if va == nil || vb == nil || *va != *vb {
 				return nil, false
 			}
+		case "ExitNodeDNSResolvers":
+			va, vb := was.ExitNodeDNSResolvers(), views.SliceOfViews(n.ExitNodeDNSResolvers)
+
+			if va.Len() != vb.Len() {
+				return nil, false
+			}
+
+			for i := range va.LenIter() {
+				if !va.At(i).Equal(vb.At(i)) {
+					return nil, false
+				}
+			}
+
 		}
 	}
 	if ret != nil {
--- a/tailcfg/tailcfg.go
+++ b/tailcfg/tailcfg.go
@@ -336,6 +336,10 @@ type Node struct {
 	// is not expected to speak Disco or DERP, and it must have Endpoints in
 	// order to be reachable.
 	IsWireGuardOnly bool `json:",omitempty"`
+
+	// ExitNodeDNSResolvers is the list of DNS servers that should be used when this
+	// node is marked IsWireGuardOnly and being used as an exit node.
+	ExitNodeDNSResolvers []*dnstype.Resolver `json:",omitempty"`
 }
 
 // DisplayName returns the user-facing name for a node which should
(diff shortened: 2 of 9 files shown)


tailcfg/tailcfg.go
```diff
@@ -336,6 +336,10 @@ type Node struct {
 	// is not expected to speak Disco or DERP, and it must have Endpoints in
 	// order to be reachable.
 	IsWireGuardOnly bool `json:",omitempty"`
+
+	// DNSServers is the list of DNS servers that should be used when this
```
bradfitz, line 340: can this be defined more? How does this interact with DNSConfig split DNS magic DNS etc? Is this just a fallback?

Cc @danderson

raggi: It should have the same semantics as the the peer API path.

There are other issues we need to look into with the peer API path, which are entirely separate, if we address those issues, some specific semantics may change, regardless, this particular field should mimic the semantics of the peer API behavior for exit node DNS.

danderson: This seems straightforward to me as described. Our normal exit node behavior with DNS is to forward all resolution traffic except MagicDNS to the exit node, via peerapi. For wireguard-only peers, this breaks because those peers don't have a peerapi. But we still need to forward DNS somewhere, because in the default config exit nodes prevent LAN traffic and so break local DNS. So, this field says "if this is a wireguard-only peer being used as an exit node, here's the DNS config to use to emulate peerapi forwarding behavior".

If we try to describe all the possible interactions of different DNS configs here, the comment will become obsolete almost immedaitely, so I'd prefer to not try and write down a spec for our full DNS semantics in this comment. The comment as-is seems as descriptive to me as the rest of our DNS config structs in tailcfg, which all describe a data plane mechanism without specifying 100% of every interaction with every other feature. Those decisions should live in the DNS manager code, and its tests.

raggi: If SplitDNS works with the peer API configuration, it should work with the configuration produced as a result of consuming this field.

maisem: This should be named appropriately  like `ExitNodeDNSServers`

control/controlclient/map.go
```diff
@@ -693,6 +693,21 @@ func peerChangeDiff(was tailcfg.NodeView, n *tailcfg.Node) (_ *tailcfg.PeerChang
...
+				return nil, false
+			}
+
+			for i := range va.LenIter() {
+				for j := range vb.LenIter() {
+					if !va.At(i).Equal(vb.At(j)) {
+						return nil, false
+					}
+				}
+			}
```
andrew-d, line 709: I think this results in calling the inner `for j := range vb.LenIter()` loop once per entry in `va`? I think this does the same, since you check the length above:
```suggestion
			for i := range va.LenIter() {
				if !va.At(i).Equal(vb.At(j)) {
					return nil, false
				}
			}
```

andrew-d (approved):

types/dnstype/dnstype.go
```diff
@@ -51,3 +52,15 @@ func (r *Resolver) IPPort() (ipp netip.AddrPort, ok bool) {
 	}
 	return
 }
+
+// Equal returns true if r and other are equal.
```
bradfitz, line 56: the usual Go wording on bool funcs is s/returns true if/reports whether/

bradfitz (approved):

PR #9410, opened 2023-09-14
cmd/cloner: do not allocate slices when the source is nil

raggi:
tailcfg.Node zero-value clone equality checks failed when I added a []*foo to the structure, as the zero value and it's clone contained a different slice header.

Updates #9377
Updates #9408

### DIFF
--- a/cmd/cloner/cloner.go
+++ b/cmd/cloner/cloner.go
@@ -122,6 +122,7 @@ func gen(buf *bytes.Buffer, it *codegen.ImportTracker, typ *types.Named) {
 		case *types.Slice:
 			if codegen.ContainsPointers(ft.Elem()) {
 				n := it.QualifiedName(ft.Elem())
+				writef("if src.%s != nil {", fname)
 				writef("dst.%s = make([]%s, len(src.%s))", fname, n, fname)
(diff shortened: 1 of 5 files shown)


cmd/cloner/cloner.go
```diff
@@ -122,6 +122,7 @@ func gen(buf *bytes.Buffer, it *codegen.ImportTracker, typ *types.Named) {
 		case *types.Slice:
 			if codegen.ContainsPointers(ft.Elem()) {
 				n := it.QualifiedName(ft.Elem())
+				writef("if src != nil {")
```
maisem, line 125: needs `src.%s`

maisem (approved):
Thanks!

josharian:
FWIW, if you'd like to get the "round up to malloc size class" behavior here (and less indentation, to boot), you could also use:

```go
writef("dst.%s = append([]%s(nil), make([]%s, len(src.%s))...)", fname, n, n, fname)
```

The generated code would look like:

```go
dst.Peers = append([]Peer(nil), make([]Peer, len(src.Peers))...)
for i := range dst.Peers {
	dst.Peers[i] = *src.Peers[i].Clone()
}
```

That should do the right thing with nils.

PR #9416, opened 2023-09-14
ipn/ipnlocal: plumb ExitNodeDNSResolvers for IsWireGuardOnly exit nodes

raggi:
This enables installing default resolvers specified by tailcfg.Node.ExitNodeDNSResolvers when the exit node is selected.

Updates #9377

### DIFF
--- a/ipn/ipnlocal/local_test.go
+++ b/ipn/ipnlocal/local_test.go
@@ -22,6 +22,7 @@ import (
 	"tailscale.com/tailcfg"
 	"tailscale.com/tsd"
 	"tailscale.com/tstest"
+	"tailscale.com/types/dnstype"
 	"tailscale.com/types/key"
 	"tailscale.com/types/logger"
 	"tailscale.com/types/logid"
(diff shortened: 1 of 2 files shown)


ipn/ipnlocal/local_test.go
```diff
@@ -952,3 +954,112 @@ func TestUpdateNetmapDelta(t *testing.T) {
...
+		CorpDNS:    true,
+	}
+
+	got := dnsConfigForNetmap(nm, prefs.View(), t.Logf, "")
+	if !resolversEqual(got.DefaultResolvers, resolvers) {
+		t.Errorf("got %v, want %v", got.DefaultResolvers, resolvers)
+	}
+}
+
+func resolversEqual(a, b []*dnstype.Resolver) bool {
```
danderson, line 1052: Ideally, would add a .Equal on dnstype.Resolver, with tests that verify equality behavior and do that trick we do to make a test fail if someone adds more fields without updating the equality check.

danderson (approved):

ipn/ipnlocal/local_test.go
raggi: Oh, you reminded me, I did add the equal function in the prior PR, I omitted tests as there were no siblings, but I'll go back and do that.

PR #9429, opened 2023-09-17
ipn/ipnlocal: handle all peer mutations in LocalBackend

maisem:
Consumers of ipn.Notify (WatchIPNBus) need to be able to reconstruct the state of the world by just subscribing to changes after they ask for the InitialNetmap. This behavior was regressed when we made LocalBackend only handle a few types of mutations and not all; as the InitialNetmap sent down the WatchIPNBus may no longer contain all the up to date information and there was no way to recover the previously unhandled information.

This makes it so that the LocalBackend handles all mutations and panics if it sees an unknown type of mutation, and it also sends the mutations over the ipn.Notify bus.

Updates tailscale/corp#12990

### DIFF
--- a/ipn/backend.go
+++ b/ipn/backend.go
@@ -87,13 +87,14 @@ type Notify struct {
 	// For State InUseOtherUser, ErrMessage is not critical and just contains the details.
 	ErrMessage *string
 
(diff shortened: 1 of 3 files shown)


bradfitz:
No longer applicable. We went a different way in chat.

PR #9430, opened 2023-09-17
ipn/ipnlocal: keep internal map updated of latest Nodes post mutations

bradfitz:
We have some flaky integration tests elsewhere that have no one place
to ask about the state of the world. This makes LocalBackend be that
place (as it's basically there anyway) but doesn't yet add the ForTest
accessor method.

This adds a LocalBackend.peers map[NodeID]NodeView that is
incrementally updated as mutations arrive. And then we start moving
away from using NetMap.Peers at runtime (UpdateStatus no longer uses
it now). And remove another copy of NodeView in the LocalBackend
nodeByAddr map. Change that to point into b.peers instead.

Future changes will then start streaming whole-node-granularity peer
change updates to WatchIPNBus clients, tracking statefully per client
what each has seen. This will get the GUI clients from receiving less
of a JSON storm of updates all the time.

Updates #1909


ipn/ipnlocal/local.go
```diff
@@ -204,10 +204,11 @@ type LocalBackend struct {
 	capTailnetLock bool // whether netMap contains the tailnet lock capability
 	// hostinfo is mutated in-place while mu is held.
 	hostinfo         *tailcfg.Hostinfo
-	netMap           *netmap.NetworkMap     // not mutated in place once set (except for Peers slice)
+	netMap           *netmap.NetworkMap                  // not mutated in place once set (except for Peers slice)
+	peers            map[tailcfg.NodeID]tailcfg.NodeView // mutated in place while mu is held; map values may escape LocalBackend, map itself may not
```
maisem, line 209: you now have two different lists of Peers in LocalBackend which go out of sync when you receive the first delta netmap

ipn/ipnlocal/local.go
```diff
@@ -4140,12 +4142,33 @@ func (b *LocalBackend) setNetMapLocked(nm *netmap.NetworkMap) {
...
 }
 
+func (b *LocalBackend) updatePeersFromNetmapLocked(nm *netmap.NetworkMap) {
+	if nm == nil {
+		b.peers = nil
+		return
+	}
+	// First pass, mark everything unwanted.
+	for k := range b.peers {
+		b.peers[k] = tailcfg.NodeView{}
```
maisem, line 4191: maybe have an explicit var

```
var invalidNode tailcfg.NodeView
for k := range b.peers {
  b.peers[k] = invalidNode
}
```

types/netmap/nodemut.go
```diff
@@ -4,20 +4,23 @@
...
 	"tailscale.com/tailcfg"
+	"tailscale.com/types/ptr"
 	"tailscale.com/util/cmpx"
 )
 
 // NodeMutation is the common interface for types that describe
 // the change of a node's state.
 type NodeMutation interface {
 	NodeIDBeingMutated() tailcfg.NodeID
+	ApplyMutation(*tailcfg.Node)
```
maisem, line 23: maybe just `Apply`?

maisem (approved):

ipn/ipnlocal/local.go
bradfitz: The current way is consistent with what we do elsewhere. We can change it to `zero` if/when that happens.

types/netmap/nodemut.go
bradfitz: done

PR #9440, opened 2023-09-18
tailcfg: add NodeCapMap

maisem:
Like PeerCapMap, add a field to `tailcfg.Node` which provides a map of Capability to raw JSON messages which are deferred to be parsed later by the application code which cares about the specific capabilities. This effectively allows us to prototype new behavior without having to commit to a schema in tailcfg, and it also opens up the possibilities to develop custom behavior in tsnet applications w/o having to plumb through application specific data in the MapResponse.

Updates #4217


tailcfg/tailcfg.go
```diff
@@ -292,6 +317,9 @@ type Node struct {
 	//    "https://tailscale.com/cap/file-sharing"
 	Capabilities []NodeCapability `json:",omitempty"`
 
+	// CapMap is a map of capabilities to their raw values.
+	CapMap NodeCapMap
```
bradfitz, line 321: omitempty?

tailcfg/tailcfg.go
```diff
@@ -343,6 +371,16 @@ type Node struct {
 	ExitNodeDNSResolvers []*dnstype.Resolver `json:",omitempty"`
 }
 
+// HasCap reports whether the node has the given capability.
+func (v NodeView) HasCap(cap NodeCapability) bool {
+	return v.ж.HasCap(cap)
+}
+
+// HasCap reports whether the node has the given capability.
+func (v *Node) HasCap(cap NodeCapability) bool {
```
bradfitz, line 380: HasCap here vs HasCapability on the other type is weird. pick a name?

bradfitz: Alternative, HasCap here but just "Contains" on the other one, since it's only about capabilities it doesn't need "Cap" or "Capabilities" in its name and "Contains" seems like the common name.

tailcfg/tailcfg.go
```diff
@@ -1250,13 +1288,64 @@ const (
...
+	if len(c) != len(c2) {
+		return false
+	}
+	for k, v := range c {
+		v2, ok := c2[k]
+		if !ok {
+			return false
+		}
+		if !slices.EqualFunc(v, v2, func(e1, e2 RawMessage) bool {
+			return bytes.Equal([]byte(e1), []byte(e2))
```
bradfitz, line 1307: why not e1 == e2?

maisem: Oh yeah that'll work too

tailcfg/tailcfg.go
```diff
@@ -1250,13 +1288,64 @@ const (
...
+			return nil, err
+		}
+		out = append(out, t)
+	}
+	return out, nil
+}
+
+// HasCapability reports whether c has the capability cap.
+// This is used to test for the existence of a capability, especially
+// when the capability has no values.
```
bradfitz, line 1336: maybe reword "has no values" somehow to be more clear?

"has no associated argument/data values"
"has no ancillary values"

We should come up with a good name for that optional value bit and use its name consistently.

ipn/ipnlocal/peerapi.go
```diff
@@ -1035,7 +1035,7 @@ func (h *peerAPIHandler) canPutFile() bool {
 // canDebug reports whether h can debug this node (goroutines, metrics,
 // magicsock internal state, etc).
 func (h *peerAPIHandler) canDebug() bool {
-	if !views.SliceContains(h.selfNode.Capabilities(), tailcfg.CapabilityDebug) {
+	if !h.selfNode.HasCap(tailcfg.CapabilityDebug) {
```
bradfitz, line 1038: all these changes could be in their own commit if you wanted to make this smaller.

control/controlknobs/controlknobs.go
```diff
@@ -48,39 +49,30 @@ type Knobs struct {
...
-		case tailcfg.NodeAttrDebugForceBackgroundSTUN:
-			forceBackgroundSTUN = true
-		case tailcfg.NodeAttrDisableDeltaUpdates:
-			disableDeltaUpdates = true
-		}
+
+	if has(tailcfg.NodeAttrOneCGNATEnable) {
+		oneCGNAT.Set(true)
+	} else if has(tailcfg.NodeAttrOneCGNATDisable) {
+		oneCGNAT.Set(false)
```
bradfitz, line 73: this is a bit worse performance-wise now. Now it does 8 passes over the slice.

maisem: I expect we will just start sending down the CapMap from control instead.

tailcfg/tailcfg.go
```diff
@@ -368,6 +368,16 @@ type Node struct {
 	ExitNodeDNSResolvers []*dnstype.Resolver `json:",omitempty"`
 }
 
+// HasCap reports whether the node has the given capability.
+func (v NodeView) HasCap(cap NodeCapability) bool {
+	return v.ж.HasCap(cap)
+}
+
+// HasCap reports whether the node has the given capability.
+func (v *Node) HasCap(cap NodeCapability) bool {
```
bradfitz, line 381: document that both of these are safe to call on a nil/invalid value?

tailcfg/tailcfg.go
```diff
@@ -317,6 +318,9 @@ type Node struct {
 	//    "https://tailscale.com/cap/file-sharing"
 	Capabilities []NodeCapability `json:",omitempty"`
 
+	// CapMap is a map of capabilities to their raw values.
```
bradfitz, line 321: say more about the values and how they can be nil? and the relation between this and the Capabilities slice?

tailcfg/tailcfg.go
```diff
@@ -1285,6 +1289,57 @@ const (
...
+	for k, v := range c {
+		v2, ok := c2[k]
+		if !ok {
+			return false
+		}
+		if !slices.EqualFunc(v, v2, func(e1, e2 RawMessage) bool {
+			return e1 == e2
+		}) {
+			return false
+		}
```
bradfitz, line 1311: Is this different from https://pkg.go.dev/slices#Equal ?

bradfitz (approved):

PR #9447, opened 2023-09-18
ipn/ipnlocal: close foreground sessions on SetServeConfig

marwan-at-work:
This PR ensures zombie foregrounds are shutdown if a new ServeConfig is created that wipes the ongoing foreground ones. For example, "tailscale serve|funnel reset|off" should close all open sessions.

Updates #8489

### DIFF
--- a/ipn/ipnlocal/local.go
+++ b/ipn/ipnlocal/local.go
@@ -128,6 +128,13 @@ func RegisterNewSSHServer(fn newSSHServerFunc) {
 	newSSHServer = fn
 }
 
+// watchSession represents a WatchNotifications channel
+// and sessionID as required to close targeted buses.
+type watchSession struct {
+	ch        chan *ipn.Notify
+	sessionID string
+}
+
 // LocalBackend is the glue between the major pieces of the Tailscale
 // network software: the cloud control plane (via controlclient), the
 // network data plane (via wgengine), and the user-facing UIs and CLIs
@@ -233,7 +240,7 @@ type LocalBackend struct {
 	loginFlags       controlclient.LoginFlags
 	incomingFiles    map[*incomingFile]bool
 	fileWaiters      set.HandleSet[context.CancelFunc] // of wake-up funcs
-	notifyWatchers   set.HandleSet[chan *ipn.Notify]
+	notifyWatchers   set.HandleSet[*watchSession]
 	lastStatusTime   time.Time // status.AsOf value of the last processed status update
 	// directFileRoot, if non-empty, means to write received files
 	// directly to this directory, without staging them in an
@@ -2044,7 +2051,7 @@ func (b *LocalBackend) WatchNotifications(ctx context.Context, mask ipn.NotifyWa
 		}
 	}
 
-	handle := b.notifyWatchers.Add(ch)
+	handle := b.notifyWatchers.Add(&watchSession{ch, sessionID})
 	b.mu.Unlock()
 
 	defer func() {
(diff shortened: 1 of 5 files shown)


ipn/ipnlocal/local.go
```diff
@@ -128,6 +128,14 @@ func RegisterNewSSHServer(fn newSSHServerFunc) {
 	newSSHServer = fn
 }
 
+// watchSession represents a WatchNotifications
+// channel along with its sessionID in order
+// notifiers to close targeted buses.
```
tylersmalley, line 133: nit: could improve the wording here. 

```suggestion
// watchSession represents a WatchNotifications channel
// and sessionID as required to close targeted buses.
```

ipn/ipnlocal/local.go
```diff
@@ -233,7 +241,7 @@ type LocalBackend struct {
 	loginFlags       controlclient.LoginFlags
 	incomingFiles    map[*incomingFile]bool
 	fileWaiters      set.HandleSet[context.CancelFunc] // of wake-up funcs
-	notifyWatchers   set.HandleSet[chan *ipn.Notify]
+	notifyWatchers   set.HandleSet[*watchSession]
```
tylersmalley, line 244: @maisem, can you confirm we're OK to change this?

ipn/ipnlocal/serve.go
```diff
@@ -279,6 +281,24 @@ func (b *LocalBackend) setServeConfigLocked(config *ipn.ServeConfig, etag string
 	}
 
 	b.setTCPPortsInterceptedFromNetmapAndPrefsLocked(b.pm.CurrentPrefs())
+
+	if prevConfig.Valid() {
```
tylersmalley, line 285: This could benefit from a comment

tylersmalley (approved):
After the small nits this LGTM, though, I would like to see if @maisem could review the ipnlocal changes in the absence of Brad to ensure we're not messing with any existing usage.

ipn/ipnlocal/local.go
marwan-at-work: One other alternative here is to make notfiyWatchers a `map[string]chan *ipn.Notify` where string is sessionID.

ipn/ipnlocal/local.go
```diff
@@ -2143,9 +2150,9 @@ func (b *LocalBackend) send(n ipn.Notify) {
 		n.FilesWaiting = &empty.Message{}
 	}
 
-	for _, ch := range b.notifyWatchers {
+	for _, sesh := range b.notifyWatchers {
```
bradfitz, line 2153: sess? (here and below)

ipn/ipnlocal/serve.go
```diff
@@ -279,6 +281,26 @@ func (b *LocalBackend) setServeConfigLocked(config *ipn.ServeConfig, etag string
 	}
 
 	b.setTCPPortsInterceptedFromNetmapAndPrefsLocked(b.pm.CurrentPrefs())
+
+	// clean up and close all previously open foreground sessions
+	// if the current ServeConfig has overwritten them.
+	if prevConfig.Valid() {
+		var newFg views.MapFn[string, *ipn.ServeConfig, ipn.ServeConfigView]
```
bradfitz, line 288: this type is pretty ugly. maybe a bit prettier if we did like:

```go
    has := func(string) bool { return false }
    if b.serveConfig.Valid() {
           has = b.serveConfig.Foreground().Has
     }
....
     if !has(k) {
```

bradfitz (approved):

PR #9451, opened 2023-09-18
client/web: split build into separate module

willnorris:
Define `tailscale.com/client/web/build` as a separate Go module and move vite output to an "out" subdirectory. This allows client builds in corp to first build the web client assets into a local writable directory, then go.mod replace the `client/web/build` module with that local directory.

Change how we serve the default index page when assets haven't been built. This prevents "dirty" builds in corp when assets ARE built.

Updates tailscale/corp#14733

### DIFF
--- a/client/web/assets.go
+++ b/client/web/assets.go
@@ -5,7 +5,6 @@ package web
 
 import (
 	"embed"
-	"io/fs"
 	"log"
 	"net/http"
 	"net/http/httputil"
@@ -15,7 +14,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	"tailscale.com/util/must"
+	"tailscale.com/client/web/build"
 )
 
 // This contains all files needed to build the frontend assets.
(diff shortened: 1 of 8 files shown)


willnorris:
(converting this back to draft for a bit.  I think this is going to end up causing other problems.  Feel free to hold off reviewing for now)

client/web/assets.go
```diff
@@ -27,24 +27,31 @@ import (
...
-// staticfiles serves static files from the build directory.
-var staticfiles http.Handler
-
-func init() {
-	buildFiles := must.Get(fs.Sub(embeddedFS, "build"))
-	staticfiles = http.FileServer(http.FS(buildFiles))
-}
+// defaultIndex is the default index page to serve if web client assets haven't been built.
+// We embed this separately rather than checking in a default build/index.html file
+// to prevent "dirty" builds when build/index.html is overwritten by vite.
```
soniaappasamy, line 32:
```suggestion
// We embed this separately rather than checking a default build/index.html file
// into git to prevent "dirty" builds when build/index.html is overwritten by Vite.
```

client/web/assets.go
```diff
@@ -27,24 +27,31 @@ import (
...
+//
+//go:embed index.html
+var defaultIndex []byte
```
soniaappasamy, line 35: nit: fallbackIndex?

soniaappasamy (approved):
Two minor nits, lgtm

willnorris:
Closing in favor of #9463

PR #9467, opened 2023-09-19
util/lru: roll back the new implementation

danderson:
We have a memory leak that is attributed to this LRU, and after much digging and staring we're not seeing the bug yet. Revert to the previous implementation to A/B verify that the leak goes away.


danderson:
Never mind, @bradfitz found the bug as I was sending this PR. Fixing in another PR.

